PR #7997, opened 2018-03-13
Add destroy clean up functions to accessibility

cosycode:
Fixes some missing destroy functions in accessibility

### DIFF
--- a/js/modules/keyboard-navigation.src.js
+++ b/js/modules/keyboard-navigation.src.js
@@ -6,7 +6,7 @@
  *
  * License: www.highcharts.com/license
  */
-    
+
 'use strict';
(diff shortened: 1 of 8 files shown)


cosycode:
I reverted some whitespacing changes

TorsteinHonsi:
The blame will still be affected, because there is now the "Revert some spacing..." commit that is the last action on the lines in question.

I think the quickest way to omit this is to make a new PR with only the [actual code changes](https://github.com/highcharts/highcharts/pull/7997/files?w=1) (trick: add `?w=1` to ignore whitespace in the diff - unfortunately this doesn't work on the blame page.)

What do you think @oysteinmoseng ?

Issue #7996, opened 2018-03-13
Spiderweb Chart: Changing startAngle to particular angle, one data point on X-axis disappears


tanmayshah95:
#### Expected behaviour
On changing Start angle of polar X axis, all X-Axis data points should get plotted from given angle.
There should not be any data loss.

#### Actual behaviour
On changing startAngle to some particular angle, one data point on X-Axis disappears.
some angles on which issue is reproduced are: 15, 120, 135, 225, 285, 300.

#### Live demo with steps to reproduce
startAngle: 0   -->  http://jsfiddle.net/dhr5dxxa/
startAngle: 15 -->  http://jsfiddle.net/dhr5dxxa/3/

Verify that last data point is not plotted, when startAngle = 15.

pawelfus:
Hi @tanmayshah95 

Thank you for reporting this bug.


Note:
It looks like last point is rendered at the same tick as the first one and `autoConnect` is wrong.

PR #8001, opened 2018-03-13
bugfix/7996-polar-startangle

pawelfus:


pawelfus:
Oops, it looks like commit-message has unescaped parts, it should be:
> Fixed #7996, polar chart with particular `startAngle` values was rendered wrong.

Issue #7991, opened 2018-03-12
Treemap: wrong level on exporting


pawelfus:
#### Expected behaviour
Current level should be exported.

#### Actual behaviour
Always root level is exported.

#### Live demo with steps to reproduce
http://jsfiddle.net/BlackLabel/4uuuxp8q/ 

Drilldown to any country, then export chart. 

#### Product version
4.2.2+

It's a regression after a85f5767f (@jon-a-nygaard)

PR #8009, opened 2018-03-14
bugfix/7991-treemap-wrong-level-exported

jon-a-nygaard:
# Description
When drilling down in a treemap chart and then exporting, the exported image would be of the top node in the tree, in stead of the current root node.

# Issue(s)
- #7991

### DIFF
--- a/js/modules/treemap.src.js
+++ b/js/modules/treemap.src.js
@@ -1011,13 +1011,16 @@ seriesType('treemap', 'scatter', {
 	translate: function () {
 		var series = this,
 			options = series.options,
-			rootId = series.rootNode =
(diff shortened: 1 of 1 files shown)


TorsteinHonsi:
I merged this PR too early before and tried to revert it, but apparently I can't reopen the PR. So I'm leaving the branch, probably we can create a new PR.

Since treemaps and sunburst are so closely related in structure, I would expect that this work equally on both. By changing the series type to `sunburst`, we should be able to drill down and export only the drilled node: http://jsfiddle.net/highcharts/4uuuxp8q/33/.

Maybe we need to go further in refactoring out the shared concepts between these two series types?

PR #8020, opened 2018-03-15
Fixed #7991, treemap had wrong root on export.

### LINKED ISSUE #7991: Treemap: wrong level on exporting

TorsteinHonsi:


jon-a-nygaard:
That's a good point.
> Maybe we need to go further in refactoring out the shared concepts between these two series types?

In this case, there is not much refactoring needed for sunburst and treemap to share the logic for drilling, it is covered by https://github.com/highcharts/highcharts/pull/8020/commits/c03b5bcb7c954726d31595962e6d075e55de088e

There is however some logic in `modules/treemap.src.js` which is better suited to be moved into `mixins/tree-series.js`

jon-a-nygaard:
Since you created this PR I can't add you as a reviewer @TorsteinHonsi, so I am adding a mention in stead.

PR #8025, opened 2018-03-15
test/clean_unit-tests.js

cosycode:
Moved all unit-tests.js into samples/unit-tests and improved some regarding failure handling with help of test-controller.js


samples/unit-tests/chart/getselectedseries/demo.html
```diff
@@ -0,0 +1,9 @@
+<script src="https://code.jquery.com/jquery-3.1.1.min.js"></script>
```
TorsteinHonsi: Are we actually using jQuery here? This also applies to many other of the new `demo.html` files. I think in the original visual samples we only used jQuery to add button click listeners.

samples/unit-tests/exporting/backgroundcolor/demo.js
```diff
@@ -0,0 +1,91 @@
...
+            }],
+
+            navigation: {
+                buttonOptions: {
+                    enabled: false
+                }
+            }
+        });
+
+    // the button handler
```
TorsteinHonsi: Do we have a button?

samples/unit-tests/series/update/demo.html
```diff
@@ -1,4 +1,4 @@
-<script src="https://code.highcharts.com/stock/highstock.js"></script>
+<script src="https://code.highcharts.com/highchart.js"></script>
```
TorsteinHonsi: Wrong file name => test fails in utils.

samples/unit-tests/loading/hideduration/demo.js
```diff
@@ -0,0 +1,100 @@
+
+QUnit.test('Show and hide duration', function (assert) {
```
TorsteinHonsi: This test fails unreliably in the browser (using utils), due to float errors. We should use `QUnit.close` or `toPrecision` on all asserts.

samples/unit-tests/chart/reflow/demo.js
```diff
@@ -1,19 +1,140 @@
 /* eslint func-style:0 */
 
+QUnit.test('Set reflow to false', function (assert) {
```
TorsteinHonsi: This test fails for me in Chrome (using utils).

samples/unit-tests/series/hide-show-highstock/demo.html
```diff
@@ -0,0 +1,6 @@
+<script src="https://code.highcharts.com/stock/highstock.js"></script>
```
TorsteinHonsi: Fails in the browser (using utils). Add `usdeur.js`, or use local data instead.

samples/unit-tests/series/update/demo.html
cosycode: Test does not fails for me. I also do not find any use of Highstock in the demo.js.

samples/unit-tests/loading/hideduration/demo.js
cosycode: I will restore the original timing, that has no problem with the float precision.

PR #8061, opened 2018-03-21
bugfix/8031-column-overlap-xaxis

ppotaczek:


ppotaczek:
Yes, you are right, sorry. I have already created a new PR with the right changes (bugfix/8031-columns-overlap-xaxis).

Issue #8031, opened 2018-03-16
Chart columns overlap X axis


proteamer:
Hi, I've faced with issue when tried set custom [xAxis.lineWidth](https://api.highcharts.com/highstock/xAxis.lineWidth) in **Highstock** chart. 

#### Actual behaviour
When I set `xAxis.lineWidth` to 10 I see that chart columns overlap `X` axis. See image bellow: 

<img width="398" alt="screen shot 2018-03-16 at 17 50 08" src="https://user-images.githubusercontent.com/1376014/37527242-a05ae6aa-2942-11e8-9f04-3d95678b6be7.png">

#### Expected behaviour
I tried to set `xAxis.lineWidth` to 10  in **Highcharts** chart. It works as expected:

<img width="397" alt="screen shot 2018-03-16 at 17 55 27" src="https://user-images.githubusercontent.com/1376014/37527518-43311660-2943-11e8-9f8c-b1bc360ea639.png">

Also I figured out that in Highstock v5.0.4 the same example work as expected.

#### Live demo with steps to reproduce

Highstock v6.0.7 example: https://jsfiddle.net/unlimit/yjjqL44x/ (works with issue)
Highcharts v6.0.7 example: https://jsfiddle.net/unlimit/ycyy7k5m/ (works as expected)

#### Product version

Highstock v6.0.7

#### Affected browser(s)
Google Chrome v65, Safari v11.0.2, etc.

pawelfus:
Hi @proteamer 

Thank  your for reporting this issue. 

*Internal note:*
Regression from 0a9220256 @KacperMadej

ppotaczek:
The workaround is to set zIndex for xAxis:

`
  xAxis: {
    zIndex: 999
  }
`

Live demo: https://jsfiddle.net/BlackLabel/jjk39ass/

PR #8066, opened 2018-03-22
bugfix/8031-columns-overlap-xaxis

ppotaczek:
This PR  resolves only: https://github.com/highcharts/highcharts/issues/8031


ppotaczek:
I did not anticipate such a situation. I wanted to fix an animation that currently does not look right, even in Highcharts: http://jsfiddle.net/BlackLabel/vw1a28rp/
I understand that the next PR should only include fix without animation?

TorsteinHonsi:
Probably the animation needs to be considered too, but maybe you should look at the clipping logic instead of the actual column shapes.

Issue #4960, opened 2016-01-26
Flags are not being show near to the x axis with inverted chart


pghilardi:
Flags on x axis works for normal charts, but not for inverted charts. The flag is placed far from the x axis.

Example: http://jsfiddle.net/sH777/323/

oysteinmoseng:
The reason this is not supported currently is that Flags series are a Highstock feature, and Highstock charts are never inverted. We see the issue, though, and obviously there are use cases for this in normal Highcharts as well.

It will technically require quite some work to make these flags function correctly on inverted charts, as the whole SVG group is being flipped and mirrored, leading not only to wrong positioning, but also mirrored text inside the flags.

pghilardi:
I think that inverted charts should be better supported by Highstock. Exists a plan to proper support of inverted charts on Highstock? Actually I am using Highcharts with the inverted option enabled and importing highstock.js to enable data grouping.

We have a use case that has time on x axis and multiple y axis with one series by y axis.

TorsteinHonsi:
I gave it another try and see that we can simply disable the invertion of the marker group by adding this override to the FlagSeries prototype:

```js
	/**
	 * Don't invert the flag marker group (#4960)
	 */
	invertGroups: noop,
```

And since flags already misbehave in inverted charts, we're not breaking anything. @oysteinmoseng now you should be able to fix this by just getting the coordinates right.

PR #8069, opened 2018-03-22
feature/4960-onseries-mixin-inverted

pawelfus:
Note:

It resolves issue #4960, but since windbarb type is also based on onseries-mixin I decided to add this feature there too. I don't think we want additionally rotated arrows by groups. 

New option added for windbarb: `xOffset` - works the same as `yOffset`, but useful for rotated chart when placed before the xAxis.

Demos for windbarb:
- before: http://jsfiddle.net/BlackLabel/yp9bc46z/27/
- after: http://jsfiddle.net/BlackLabel/yp9bc46z/25/

Maybe default value for xOffset/yOffset should set according to inverted option? I would leave this as is.

It's not a perfect solution: tooltip's position (and anchor?) for non-rotated series could be improved.


pawelfus:
PS: Testing resolve conflicts from github directly, I hope it will be fine :)

PR #8145, opened 2018-04-06
test/introducing-templates

cosycode:
Introducing chart templates for faster unit-tests


cosycode:
Implementing wrap for Chart.update and adding additional template tests

cosycode:
`./node_modules/.bin/gulp test --tests unit-tests/performance/*`

Performance benchmarks (MacBook Pro):
performance/chart-template: 100 tests, 350ms - 457ms
performance/highcharts-chart: 100 tests, 954ms - 1261ms

test/chart-template.js
```diff
@@ -0,0 +1,392 @@
...
+     */
+    function treeCopy(source, propertiesTree) {
+        if (!source) {
+            return source;
+        }
+        switch (typeof source) {
+            default:
+                return source;
+            case 'array':
+                return JSON.parse(JSON.stringify(source));
```
pawelfus: Won't that be a problem with dataset like this? 

```javascript
{
  series: [{
    data: [{
      y: 10,
      events: { click: function () { } }
    }]
  }]
}
```

`stringify()` will remove callback. Similar for array of series, where a series has events.

test/chart-template.js
```diff
@@ -0,0 +1,392 @@
...
+         *  Instance Properties
+         *
+         * */
+
+        /**
+         * The name of the template as a flag in the chart
+         *
+         * @type {string}
+         */
+        Object.defineProperty(chart, 'template', {
```
pawelfus: Will we run tests in IE < 9 ? `defineProperty` is supported since ~IE9 (https://caniuse.com/#search=defineProperty)

(Below we have more of them)

test/chart-template.js
```diff
@@ -0,0 +1,392 @@
...
+    });
+
+    /* *
+     *
+     *  Global
+     *
+     * */
+
+    // Prevent changes to ChartTemplate properties
+    Object.freeze(ChartTemplate);
```
pawelfus: Similar to `defineProperty`: https://caniuse.com/#search=freeze (no support for IE < 9)

test/karma-setup.js
```diff
@@ -85,14 +85,25 @@ QUnit.module('Highcharts', {
...
-        Highcharts.charts.length = 0;
+            if (currentChart.template) {
+                templateCharts.push(currentChart);
+            } else if (currentChart.destroy && currentChart.renderer) {
+                currentChart.destroy();
+            }
+        }
 
+        Highcharts.charts.length = 0;
+        Array.prototype.push.call(Highcharts.charts, templateCharts);
```
pawelfus: Shouldn't be `apply` instead of `call`? 

```
var a = [], b = [1, 2, 3];
Array.prototype.push.call(a, b);
```
> a = [[1, 2, 3]]


```
var a = [], b = [1, 2, 3];
Array.prototype.push.apply(a, b);
```
> a = [1, 2, 3]

test/chart-template.js
cosycode: This would be only a problem, if you like to create templates with events. In the test you can create event functions and they will get removed in the end of the test case.

test/chart-template.js
cosycode: The class does not depend on it, so I could change it for now.

test/karma-setup.js
cosycode: Ooops. Thanks! I will fix this.

pawelfus:
Can we somehow access the test in browser? For example: http://utils.highcharts.local/samples/#test/unit-tests/series-line/type - throws error

cosycode:
Chart templates are not available in the browser and visual tests. See test/readme.md
This will requires changes to highcharts-utils and how it is working right now.

test/chart-template.js
pawelfus: Sorry, let me explain, for example I want to test multiple yAxes with events:

```javascript
    ChartTemplate.test('line-simple', {
        yAxis: [{
            events: {
                afterSetExtremes: function () {
                    this.chart.firedCallback1 = true;
                }
            }
        }, {
            events: {
                afterSetExtremes: function () {
                    this.chart.firedCallback2 = true;
                }
            }
        }],
        series: [{
            data: [1, 2]
        }]

    }, function (template) {

        var chart = template.chart;

        chart.yAxis[0].setExtremes(0, 10000, false);
        chart.yAxis[1].setExtremes(0, 10000);

        assert.strictEqual(
            chart.firedCallback1 && chart.firedCallback2,
            true,
            'Events are fine'
        );
    });
```

Demo is fine: http://jsfiddle.net/BlackLabel/o5san9w6/3/

Test failed:
> HeadlessChrome 0.0.0 (Mac OS X 10.13.3) Highcharts series-line/type FAILED
>	Events are fine
>	Expected: true
>	Actual: undefined

pawelfus:
Thanks! Missed this info

test/chart-template.js
cosycode: If I apply the same update functionality of the template to your fiddle, it fails too: http://jsfiddle.net/o5san9w6/7/
I am not sure, why it breaks - especially with the first y-axis. Chart.update should support axis updates and fire events.

pawelfus: Ha! Looks like I found a bug in attaching/removing events for the yAxis..

cosycode:
Issue #8161 is related to this branch.

test/chart-template.js
cosycode: I have integrated the bugfix #8161. Please retry your tests.

PR #8201, opened 2018-04-18
test/test-controller

cosycode:
Add additional OnElement functions to test/test-controller.js


cosycode:
There is a problem with the event chain.

PR #8278, opened 2018-05-03
Add H.browser string to js/parts/Utilities.js

cosycode:
Provide `Highcharts.browser` property with short name of current browser context.


TorsteinHonsi:
Sorry, I think we shouldn't bloat Highcharts with this. If we need it in the tests, we should keep this code in the `highcharts-utils` repo instead.

Issue #8342, opened 2018-05-15
[BUG] Loading options documentation is incorrect regarding styled mode classes


marques-work:
For the lines below, the classes should read `highcharts-loading-inner` and `highcharts-loading`, respectively:

https://github.com/highcharts/highcharts/blob/88e8c4884cae0884fc8dc90a80b2c4d80f793ceb/js/parts/Options.js#L2338

https://github.com/highcharts/highcharts/blob/88e8c4884cae0884fc8dc90a80b2c4d80f793ceb/js/parts/Options.js#L2354

KacperMadej:
Hi @marques-work 

Thank you for finding this one.

marques-work:
@KacperMadej I submitted a small [PR to fix this](https://github.com/highcharts/highcharts/pull/8345)

PR #8345, opened 2018-05-16
Fix the documented classnames for styled mode chart loading message

marques-work:
Fixes https://github.com/highcharts/highcharts/issues/8342

Should be documented as `highcharts-loading` and `highcharts-loading-inner`.


TorsteinHonsi:
Thanks!

Issue #7067, opened 2017-08-16
chart.update() doesn't enable the navigator under certain conditions


ransoing:
#### Expected behaviour

Calling `chart.update()` to toggle `navigator.enabled` should always work properly.

#### Actual behaviour

If a plot is initialized with both `scrollbar.enabled` and `navigator.enabled` set to `false`, then the navigator will not appear when `chart.update( {navigator: {enabled:true}} )` is called.

If the plot is initialized with either `scrollbar.enabled` or `navigator.enabled` set to `true`, then it will work properly.

#### Live demo with steps to reproduce
Press the "Toggle Navigators" button. Note how the second plot does not show a navigator.
http://jsfiddle.net/x40me94t/1/

KacperMadej:
Hi,

Thank you for reporting about the bug.

_Internalnote:_ Navigator is added only on `beforeRender` event and cannot be added through `chart.update`.

sebastianbochan:
_Internal note_
It seems that options are not "updated" in update function. However when we edit it, it is still wrong.

Demo:
- http://jsfiddle.net/rrLzatcv/

KacperMadej:
**Workaround:** 
Set options for navigator manually, fire event that adds navigator/scrollbar, redraw the chart to see the changes. 
Demo: http://jsfiddle.net/x40me94t/2/

cosycode:
_Internal note_
Requires unit test (scheduled for test/move_issues_tests_hs4.0.1)

PR #8397, opened 2018-05-29
#7067 Chart.update enables navigator

cosycode:
- enables Navigator after Chart.update, if not previously initiated
- corresponding unit test will be part of test/move_issues_tests_hs4.0.1
- original issue as test: http://jsfiddle.net/j47frz8j/

### DIFF
--- a/js/parts/Dynamics.js
+++ b/js/parts/Dynamics.js
@@ -522,6 +522,9 @@ extend(Chart.prototype, /** @lends Highcharts.Chart.prototype */ {
         } else if (pick(redraw, true)) {
             chart.redraw(animation);
         }
+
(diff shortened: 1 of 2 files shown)


TorsteinHonsi:
There is already logic dealing with `chart.update` on the Navigator object:

https://github.com/highcharts/highcharts/blob/291f69b500dab1cf7b7dc6ce6911f416b340f036/js/parts/Navigator.js#L804-L820

By default, if there's a `chart.navigator` object, calling `chart.update` with a `navigator` member automatically relay the call to `chart.navigator.update()`. Has this been accounted for? I assume the fix is for the case when `chart.navigator` does _not_ exist from before, in which case the existing function will fail?

How does this work with the new logic in the PR?

PR #8405, opened 2018-05-30
[closes #8404] Add symbolUnicode property to Series after getSymbol() is called

### LINKED ISSUE #8404: Add series.symbolUnicode property for easy tooltip formatting
PR https://github.com/highcharts/highcharts/pull/8405

#### Expected behaviour

As per [this ](https://stackoverflow.com/questions/25973920/how-to-display-highchart-series-line-marker-symbol-from-tooltip-formatter) popular SO question, it would be nice if the tooltip symbol matched the actual chart symbol, and if this was available as a Unicode character, this could be achieved using `pointFormat`

#### Actual behaviour

    pointFormat = '{series.symbol} {point.y}';

Using the above, you will get `circle 123` or `diamond 123` in the tooltip, not very useful

    pointFormat = '{series.symbolUnicode} {point.y}';

Using the above, you could get `● 123` or `♦123` in the tooltip, which you can color appropriately

current `symbol` example http://jsfiddle.net/w6yvxpzh/

Proposed `symbolUnicode` example http://jsfiddle.net/pjqz79pv/ (will work when GA)

#### Product version
Highcharts v6.1.0

#### Affected browser(s)
All

ciarancourtney:
Linked issue https://github.com/highcharts/highcharts/issues/8404

This is just a draft idea, please let me know if its an acceptable addition and any changes or documentation required. Feel free to make any changes yourself.

Thanks

### DIFF
--- a/js/parts/Series.js
+++ b/js/parts/Series.js
@@ -2408,6 +2408,27 @@ H.Series = H.seriesType('line', null, { // base series options
         series.getColor();
         series.getSymbol();
 
+        switch (series.symbol) {
+            case 'circle':
+                series.symbolUnicode = '●';
+                break;
+            case 'diamond':
+                series.symbolUnicode = '♦';
+                break;
+            case 'square':
+                series.symbolUnicode = '■';
+                break;
+            case 'triangle':
+                series.symbolUnicode = '▲';
+                break;
+            case 'triangle-down':
+                series.symbolUnicode = '▼';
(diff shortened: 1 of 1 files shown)


TorsteinHonsi:
Thanks for the proposal, I suggested a an event-based solution at https://github.com/highcharts/highcharts/issues/8404#issuecomment-393162016

PR #8413, opened 2018-05-31
Fixed issue where image patterns could not have translucent opacity w…

oysteinmoseng:
…ith default animation.

We animate image patterns by opacity by default, to avoid choppy loading of images. We should obviously take opacity setting into consideration when deciding what opacity to animate to.


oysteinmoseng:
Actually we should add a unit test for this too

oysteinmoseng:
Unit test added, extra bugs fixed, but fails in karma for unknown reasons. Sophie investigating.

oysteinmoseng:
This is now considered ready for merge.

PR #8448, opened 2018-06-11
hc7/columnpyramid

sebastianbochan:


sebastianbochan:
HC7

sebastianbochan:
1) I created the demo with "the highest pyramids in the world."
2) Reorganized demos
3) Added

One additional questions, by "Review the options, for example it now has references to a column series which is not used." do you mean that in some descriptions of parameters, do not include information that option is available for `columnpyramid` like

```
borderColor: Color
The color of the border surrounding each column or bar.
```

Should I update desc in `ColumnSeries.js` file or overwrite each option with desc in my file?

TorsteinHonsi:
I fixed the column reference, thanks!

PR #8463, opened 2018-06-13
Improve description

sophiebremer:
Improve description for Highstock option `series.indicator.params.period`


sebastianbochan:
Small misspelling in the last word.

PR #8575, opened 2018-07-05
Feature/multiple datalabels

oysteinmoseng:
Adds functionality for multiple datalabels. This functionality is added in core, and no additional module is required. It allows users to specify dataLabel options as an array, drawing multiple dataLabels per point.

Existing tests are passing, but there is a todo list for after Gantt Beta:

- Improve support for arearange/errorbar etc series, the current solution is very inefficient
- Add full support for other series types, particularly pie charts
- Add tests


js/modules/overlapping-datalabels.src.js
```diff
@@ -42,25 +42,29 @@ addEvent(Chart, 'render', function collectAndHide() {
...
+                        label.labelrank = pick(
                             point.labelrank,
                             point.shapeArgs && point.shapeArgs.height
                         ); // #4118
-                        labels.push(point[coll]);
-                    }
-                });
+                        labels.push(label);
+                    });
+                } else if (point.dataLabel) {
```
jon-a-nygaard: Is it possible for the future to use only `point.dataLabels` and remove logic for `point.dataLabel`?

oysteinmoseng: Ideally, yes, but it would potentially break customer code (plugins etc), so we'd have to deprecate it first at least

PR #8593, opened 2018-07-09
bugfix/2165-legend-labelformatter-visiblity-change

pawelfus:


pawelfus:
Yes, I was thinking about performance and decided that legend's redraw is quite fast. You are right, we should limit this case to `labelFormatter` (how about `labelFormat`? I think we can't use there `if-else`, right?).

TorsteinHonsi:
Yes I believe you can check directly on the options, but it needs to be verified of course.

js/parts/Chart.js
```diff
@@ -446,16 +446,16 @@ extend(Chart.prototype, /** @lends Highcharts.Chart.prototype */ {
                     if (serie.updateTotals) {
                         serie.updateTotals();
                     }
-                    redrawLegend = true;
                 }
+                redrawLegend = true; // #2165
```
TorsteinHonsi: Additional tests to redraw legend only when a formatter is defined.

pawelfus:
**Note**: condition checks for `labelFormat` too, for example: http://jsfiddle.net/BlackLabel/s7pxqt8v/ for dynamic formats:

```javascript
    legend: {
      labelFormat: '{name} (click to {options.nextStatus})'
    },

    series: [{
      ...
      nextStatus: 'hide',
      events: {
        legendItemClick: function () {
          this.options.nextStatus = this.visible ? 'show' : 'hide';
        }
      }
    }]
```

js/parts/Chart.js
pawelfus: Done

pawelfus:
Wait, it doesn't make sense what I did, I will revert and commit correct solution.

Issue #2405, opened 2013-10-28
Tooltip ignores transform: scale


kzoon:
When I use transform: scale, the tooltips ignore this. 
See http://jsfiddle.net/kzoon/v3aa8/
The mouseover on december points display the august tooltips

Thirteen09:
I've run into this problem too. Do you have some kind of solution? 
Can you fasten fixing this bug? We use scale all over our application and have tool-tips show in wrong places all over it. The discussion at #1753 does not help in our case.

kzoon:
I use this workaround:

    HighCharts.wrap(HighCharts.Pointer.prototype, 'normalize', function (proceed, event, chartPosition) {
        var e = proceed.call(this, event, chartPosition);
  
        var element = this.chart.container;
        if (element && element.offsetWidth && element.offsetHeight) {

            var scaleX = element.getBoundingClientRect().width / element.offsetWidth;
            var scaleY = element.getBoundingClientRect().height / element.offsetHeight;
            if (scaleX !== 1) {
                e.chartX = parseInt(e.chartX / scaleX, 10);
            }
            if (scaleY !== 1) {
                e.chartY = parseInt(e.chartY / scaleY, 10);
            }

        }
        return e;
    });

oriko:
Hi 
How do you use this workaround? 
Where do I need copy this code to? I am using Highcharts JS v4.1.8 with angular (highcharts-ng).

Thanks,

kzoon:
No. This fiddle  http://jsfiddle.net/0qppo7yr/ implements my workaround. There it works to prevent problems due to transform: scale(0.7) in the css

dmitrygorbenko:
Hi guys,

any update on this thread ?
Can the patch provided by user kzoon be applied ?

thanks,
Dima

TorsteinHonsi:
Yes, the drop-in fix seems to work well with the current version, but it will most likely fail with old browsers that don't implement the `getBoundingClientRect` API.

dmitrygorbenko:
yes, you are right !

kzoon:
Not solved in Highcharts 5. My workaround still works.

KonstantinJahnel:
@kzoon your fix seems to work perfectly fine in Firefox and IE but i can't get it running in Chrome? 
Any idea why?

starquake:
For me the fix works perfectly fine. Does the fiddle work?

What issues do you have?

starquake:
[...]

PR #8634, opened 2018-07-17
Scaling mouse cursor fix (PoC)

Llorx:
I've created this pull request **not to be merged**, but to show and discuss a fix for this issue: https://github.com/highcharts/highcharts/issues/2405. Will create the appropriate pull request when we reach an agreement. **The only important commit is [this one](https://github.com/highcharts/highcharts/pull/8634/commits/3cb94c7bad5c4acc89eeecfd0d7caed4262fc367)**

PoC: http://jsfiddle.net/tnprxch5/8/

What it does is compare the actual container size with the configured container size, to get the scaling ratio applied to it.

It will detect the ratio when the chart is created, but because of the inability to detect when a parent has changed the scaling, the user needs to call `updateScaling()` each time the scaling changes.

I don't like the `updateScaling()` thingy, but at least we can have the bug fixed internally. Way better than applying custom fixes externally.

Another way can be to detect the ratio each time the mouse moves over the container. Is not that bad as `getBoundingClientRect()` is mid performant. We can add a option to "enable scaling detection" (disabled by default) and document it.

With this we can have both options, to call `updateScaling()` manually for the ones that want to optimize performance and to "enable scaling detection" for the ones that don't mind the small performance hit.

By the way, the "detect scaling change" is only when the chart is inserted inside dynamic libraries. For the 99% of users, by detecting it on init and window resize it will work right away without any extra work.


TorsteinHonsi:
Closing due to inactivity

Llorx:
Aren't we going to discuss if this can be applied?

PR #8663, opened 2018-07-23
Current price indicator.

sebastianbochan:


samples/highcharts/studies/stocktools-gui/demo.js
```diff
@@ -1,5 +1,17 @@
...
+        // line
+        showPrice: {
+            enabled: true,
+            color: 'red'
+        },
+        // label
+        priceIndicator: {
+            enabled: true,
+            label: {
+                enabled: true
```
pawelfus: Can this be set by default? When priceIndicator is enabled, label is enabled too.

sebastianbochan:
TODO:
- styles, when whole layout will be available.

PR #8671, opened 2018-07-24
hc7/stock-tools: Fullscreen module.

sebastianbochan:


TorsteinHonsi:
@KacperMadej  There are conflicting files now, can you resolve those?

TorsteinHonsi:
@sebastianbochan When I click the weather icon I get a JS error that `H.FullScreen` is not a constructor...

sebastianbochan:
Fixed missing reference to the file. 

Please note that our iframes in the unit tests, blocks full screen (missing allowfullscreen parameters), so nothing will happend after click.

TorsteinHonsi:
Thanks, it works now:
- In the utils, cmd-click the _studies/stocktools-gui_ demo to open it in a separate frame.
- Click the weather icon.

However I requested that the chart height was updated to use the full screen when opened in full screen. It seems to me that most of the screen are is wasted and the chart is still too small. Shouldn't this be the default behaviour?

sebastianbochan:
It was a problem with css styles, but now works.

One extra note, the stockToolbar is not resized, but its related with chart.setSize() bug reported internally.

PR #8679, opened 2018-07-25
Indicators: Support for OHLC data with line type of series.

sebastianbochan:


js/indicators/indicators.src.js
```diff
@@ -4,12 +4,39 @@ import '../parts/Utilities.js';
...
+            dataGrouping: {
+                approximation: 'ohlc'
+            }
+        });
+        this.pointValKey = 'close';
+        this.keys = ['x', 'open', 'high', 'low', 'close'];
+        this.pointArrayMap = ['open', 'high', 'low', 'close'];
+        this.toYData = function (point) {
+            return [point.open, point.high, point.low, point.close];
+        };
```
pawelfus: Can we extend `this` with properties from OHLC? Something like this:

```
extend(this, {
  keys: ohlcPrototype.keys,
  pointValKey: ohlcPrototype.pointValKay,
  ...
});
```

js/indicators/indicators.src.js
```diff
@@ -4,12 +4,39 @@ import '../parts/Utilities.js';
...
     error = H.error,
     Series = H.Series,
     isArray = H.isArray,
     addEvent = H.addEvent,
     seriesType = H.seriesType;
 
+/*
+ * Support for OHLC data with line type of series.
+ */
+wrap(H.Series.prototype, 'init', function (proceed, chart, options) {
```
pawelfus: Can we add `init` event in the core?

sebastianbochan: No reference to current series

pawelfus: I mean add in the core `fireEvent(series, 'init')` then use `addEvent(H.Series, 'init', function () { .. });` instead of `wrap()`.

PR #8723, opened 2018-08-02
 hs7/tooltip-split-positioner

pawelfus:
PR adds two features
- `tooltip.headerShape` to control shape of header when `tooltip.split = true`. For example we want rectangle boxes, but header should be rendered as anchored label (callout)
- `tooltip.positioner` is now called even when `split` is enabled. For each box, `tooltip.positioner` is called separately

Discussion points:
- at this moment, `tooltip.positioner` is called for `header` too. That may be confusing for users, and might be a problem for wrappers (to be confirmed - different arguments in the same callback for strong typing). Maybe `tooltip.headerPositioner` is a better idea? We already have `tooltip.headerFormat`, now added `tooltip.headerShape`, then `tooltip.headerPositioner` makes sense.

Demo: 
- [Tooltip box rendered in the top left corner of the pane](http://jsfiddle.net/BlackLabel/841k7wjn/43/)

Note:
- Even though docs says it's ready since v7, it's safe to merge it with v6 and release.

### DIFF
--- a/js/parts/Options.js
+++ b/js/parts/Options.js
@@ -2575,6 +2575,13 @@ H.defaultOptions = {
          * where the reference point is in the plot area. Add `chart.plotLeft`
          * and `chart.plotTop` to get the full coordinates.
          *
+         * Since v7, when [tooltip.split](#tooltip.split) option is enabled,
+         * positioner is called for each of the boxes separately, including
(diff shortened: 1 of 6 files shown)


sebastianbochan:
IMHO, headerPositioner is definately better, but maybe we should also consider to add tooltips per pane, by single param: `showTooltipPerPane` without positioner too?

PR #8728, opened 2018-08-03
doc/types-chart

sophiebremer:
TypeScript-compatible doclets


js/parts/Chart.js
```diff
@@ -1242,24 +1361,34 @@ extend(Chart.prototype, /** @lends Highcharts.Chart.prototype */ {
      * Resize the chart to a given width and height. In order to set the width
      * only, the height argument may be skipped. To set the height only, pass
      * `undefined` for the width.
-     * @param  {Number|undefined|null} [width]
+     *
+     * @function Highcharts.Chart#setSize
+     *
+     * @param  {number|null} [width]
```
pawelfus: We don't define `undefiend` anymore?

sophiebremer: In TypeScript every optional parameter implies support of `undefined`. So if you like to skip an optional parameter to get to an other optional parameter, you are always allowed to set the one to skip to `undefined`.

sophiebremer: Do you think, we should add to every optional parameter the `undefined` type, to make this clear?

pawelfus: I'm just worried it might be confusing for users. Like "is undefined deprecated?"

sophiebremer: I see your point. I will add `undefined` to optionals, to make it clear for everyone.

js/parts/Chart.js
```diff
@@ -4,6 +4,27 @@
  * License: www.highcharts.com/license
  */
 
+/**
+ * The chart title. The title has an `update` method that allows modifying the
+ * options directly or indirectly via `chart.update`.
+ *
+ * @typedef {Highcharts.SVGElement} Highcharts.TitleObject
+ *
+ * @property {Function} update
```
TorsteinHonsi: The parameters are the same as `TitleOptions`.

js/parts/Chart.js
```diff
@@ -4,6 +4,27 @@
...
+ */
+
+/**
+ * The chart subtitle. The subtitle has an `update` method that
+ * allows modifying the options directly or indirectly via
+ * `chart.update`.
+ *
+ * @typedef {Highcharts.SVGElement} Highcharts.SubtitleObject
+ *
+ * @property {Function} update
```
TorsteinHonsi: The parameters are the same as `SubtitleOptions`.

PR #8730, opened 2018-08-03
hs7/natr-indicator

PaulDalek:
Added Normalized ATR indicator.


js/indicators/natr.src.js
```diff
@@ -0,0 +1,77 @@
...
+
+            for (; i < atrLength; i++) {
+                atrData.yData[i] = atrData.values[i][1] / yVal[period][3] * 100;
+                atrData.values[i][1] = atrData.yData[i];
+                period++;
+            }
+
+            NATR = atrData.values;
+            xData = atrData.xData;
+            yData = atrData.yData;
```
pawelfus: We can simply return `atrData`, destructuring not needed.

PR #8780, opened 2018-08-14
bugfix/8769-rangeselector-ignores-marginLeft-update

sebastianbochan:


sebastianbochan:
Will test and let you know

sebastianbochan:
I created simple test: http://jsfiddle.net/BlackLabel/uywep8o5/ and seems to not working.

I need to destroy / create new Rangeselector, cause plotLeft is not updated yet (chart.update), when buttons are rendered.

PR #8807, opened 2018-08-19
Mention the .change field seen in the sample

dandv:
[The sample](http://jsfiddle.net/46no1q9m/1/) shows a `.change` field being used:

```js
pointFormat: '<span style="color:{series.color}">{series.name}</span>: <b>{point.y}</b> ({point.change}%)<br/>',
```


TorsteinHonsi:
Thanks!

PR #8810, opened 2018-08-20
Document the relationship between ordinal:false and gapSize

dandv:
This is my understanding; feel free to change.

Also, the documentation for gapSize should be updated to reflect the relationship with `ordinal: false`.


pawelfus:
Hi @dandv 

Disabling `xAxis.ordinal` does not ignore `series.gapSize`, please compare demos:

- [gapSize: 2 + ordinal: true](http://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/stock/plotoptions/series-gapsize/)
- [gapSize: 2 + ordinal: false](http://jsfiddle.net/BlackLabel/b8ekp41j/)

PR #8825, opened 2018-08-22
hc7/bubble-legend

ppotaczek:
New feature with unit tests and documentation added to bubble series.


KacperMadej:
@TorsteinHonsi Is everything fine now a could be pulled or do you have any more comments?

Issue #7186, opened 2017-09-25
Publish native ES6 modules


dandv:
With [native ES6 imports support having landed in Chrome and Safari](https://jakearchibald.com/2017/es-modules-in-browsers/) a few months ago, and [Node v8.5.0 adding experimental support for ES6 modules](http://2ality.com/2017/09/native-esm-node.html) last week, it would be great to publish native ES6 modules for HighCharts and HighStock.

Publishing them at https://code.highcharts.com/ would make it very easy to use from browsers without any bundlers, solving (at least for compatible browsers) long-standing issues like #4994.

jon-a-nygaard:
@TorsteinHonsi This has been technically ready for a while through our build script. What we need to discuss first is a good location for the files. E.g code.highcharts.com/es6 and code.highcharts.com/js/es6, could be an option.

Also quality check that the final structure works for all our distribution platforms, code.highcharts.com, zip folders, and highcharts-dist.

jon-a-nygaard:
*Internal Note: Landed on folder structure `code.highcharts.com/es-modules` and `code.highcharts.com/js/es-modules`*

jon-a-nygaard:
The ES modules will be published in the next maintenance release v6.0.8.

jon-a-nygaard:
Hi @netaisllc,

Apologies for the late reply.

The ES-modules is released since v6.1.0, release v6.0.8 was bumped to v6.1.0.
You can find the es-modules located at `/es-modules` for classic mode, and the styled mode under `/js/es-modules`.
For instance the highcharts master file can be found on our DNS at https://code.highcharts.com/es-modules/masters/highcharts.src.js

The publish of our ES modules was not mentioned in the release article, only briefly in the [changelog](https://www.highcharts.com/documentation/changelog) of v6.1.0.

Best regards
Jon

dandv:
Thanks @jon-a-nygaard. I've submitted a [PR to document how to load Highcharts from CDN as an ES module](https://github.com/highcharts/highcharts/pull/8827). I was going to also add how to dynamically import it, but ran into [this issue](https://github.com/basarat/typescript-book/issues/422). Any chance to avoid `export default`? More reasons:

1. https://basarat.gitbooks.io/typescript/docs/tips/defaultIsBad.html
2. https://blog.neufund.org/why-we-have-banned-default-exports-and-you-should-do-the-same-d51fdc2cf2ad
3. https://github.com/airbnb/javascript/issues/1365

[...]

PR #8827, opened 2018-08-22
Explain how to load as ECMAScript module from CDN

dandv:
See #7186

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -8,8 +8,34 @@ Highcharts JS is a JavaScript charting library based on SVG, with fallbacks to V
 
 ## Download and install Highcharts
 This is the *working repo* for Highcharts. If you simply want to include Highcharts into a project, use the [distribution package](https://www.npmjs.com/package/highcharts) instead, or read the [download page](http://www.highcharts.com/download). Please note that there are several ways to use Highcharts. For general installation instructions, see [the docs](http://www.highcharts.com/docs/getting-started/installation).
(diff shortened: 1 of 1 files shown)


TorsteinHonsi:
Thanks, looks good to me. What do you think @jon-a-nygaard ?

Also, we should make the demo our own so we don't depend on jsFiddle accounts that we don't manage.

dandv:
> Also, we should make the demo our own so we don't depend on jsFiddle accounts that we don't manage.

Good point. Feel free to change my PR.

Also a Dynamic import example would be great because it's the most interesting use of ES module imports.

TorsteinHonsi:
@jon-a-nygaard Awaiting your review.

TorsteinHonsi:
@jon-a-nygaard Doesn't really fit in anywhere - I suppose putting it under `highcharts/chart/....` is least counterintuitive. But it doesn't matter much as long as it is there.

dandv:
The URL that ended up in the [ES CDN loading example](https://github.com/highcharts/highcharts#load-highcharts-from-the-cdn-as-ecmascript-modules) doesn't work:

    https://code.highcharts.com/es-modules/masters/highcharts.js

Is this intentional, so the dev would select the appropriate version, or should a working URL be used in the `import` statement?

TorsteinHonsi:
Sorry about that, it was not intentional.

PR #8835, opened 2018-08-23
Added support for AMD module loading.

oysteinmoseng:
We now support AMD loading, e.g. RequireJS.

We also need to update github.highcharts.com.


jon-a-nygaard:
`highcharts-assembler#v1.0.27` is now deployed to github.highcharts.com

PR #8870, opened 2018-08-31
Remove old build files of v4

sophiebremer:


jon-a-nygaard:
The download builder still uses `parts.js`, but it is using one that is hosted on www.highcharts.com
So it should be safe to remove it from this repository

TorsteinHonsi:
Okay, but I think it is copied over from this repo to the www as part of the release process. I'll revert that particular file before merging.

TorsteinHonsi:
Deleting this, awaiting a clean PR with only the intro files deleted.

jon-a-nygaard:
@TorsteinHonsi I don't think the `parts.js` in this repository is used at all. There's no reference to it in the highcharts repository to it, and the last time the file in www.highcharts.com was updated, it was done manually.

PR #8875, opened 2018-09-03
hs7/regressions

kamilkulig:
Added Linear Regression, Linear Regression Slope, Linear Regression Intercept and Linear Regression Angle indicators.

### DIFF
--- a/js/indicators/regressions.src.js
+++ b/js/indicators/regressions.src.js
@@ -203,9 +203,9 @@ seriesType('linearRegression', 'sma',
                 xAxisUnit = this.options.params.xAxisUnit ||
                 this.findClosestDistance(xData);
 
-          // Iteration logic: x value of the last point within the period
-          // (end point) is used to represent the y value (regression)
-          // of the entire period.
+            // Iteration logic: x value of the last point within the period
+            // (end point) is used to represent the y value (regression)
+            // of the entire period.
 
             for (i = period - 1; i <= xData.length - 1; i++) {
                 periodStart = i - period + 1; // adjusted for slice() function
@@ -217,8 +217,8 @@ seriesType('linearRegression', 'sma',
                   xAxisUnit);
 
                 lineParameters = this.getRegressionLineParameters(
-                periodTransformedXData, periodYData
-              );
+                  periodTransformedXData, periodYData
+                );
 
                 endPointY = this.getEndPointY(lineParameters,
                 periodTransformedXData[periodTransformedXData.length - 1]);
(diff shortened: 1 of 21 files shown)


js/indicators/regressions.src.js
```diff
@@ -0,0 +1,387 @@
...
+     * @since 7.0.0
+     * @optionparent plotOptions.linearRegression
+     */
+    {
+        name: 'Linear Regression Indicator',
+        marker: {
+            enabled: false
+        },
+        params: {
+            period: 2,
```
pawelfus: Is it common to use only two points for a linear regression indicator?

js/indicators/regressions.src.js
```diff
@@ -0,0 +1,387 @@
...
+     *
+     * @extends plotOptions.sma
+     * @product highstock
+     * @sample {highstock} stock/indicators/linear-regression
+     *                     Linear regression indicator
+     * @since 7.0.0
+     * @optionparent plotOptions.linearRegression
+     */
+    {
+        name: 'Linear Regression Indicator',
```
pawelfus: By default, shouldn't LRI include `period` in the name? For example `Linear Regression Indicator (2)` ? Check ZigZag indicator for example (`nameBase` should be enough, I think).

js/indicators/regressions.src.js
```diff
@@ -32,12 +32,8 @@ seriesType('linearRegression', 'sma',
      * @optionparent plotOptions.linearRegression
      */
     {
-        name: 'Linear Regression Indicator',
-        marker: {
-            enabled: false
-        },
         params: {
-            period: 2,
+            period: 14,
```
pawelfus: No need to set it here, this can be inherited from `SMA.params.period`.

js/indicators/regressions.src.js
```diff
@@ -32,12 +32,8 @@ seriesType('linearRegression', 'sma',
...
             index: 2,
```
pawelfus: Just realized: why `index: 2` ?

js/indicators/regressions.src.js
pawelfus: No need to set it here, this can be inherited from `SMA.params.period`.

js/indicators/regressions.src.js
```diff
@@ -218,10 +185,57 @@ seriesType('linearRegression', 'sma',
...
+                endPointY,
+                periodXData,
+                periodYData,
+                periodTransformedXData,
+                xAxisUnit = this.options.params.xAxisUnit ||
+                this.findClosestDistance(xData);
+
+          // Iteration logic: x value of the last point within the period
+          // (end point) is used to represent the y value (regression)
+          // of the entire period.
```
pawelfus: Indentation again - doesn't `gulp lint` show errors here?

PR #8877, opened 2018-09-03
Fixed #8440, keyboard navigation in stock charts throwing errors.

### LINKED ISSUE #8440: Keyboard navigation of chart throws errors instead of looping
#### Expected behaviour

Using the left/right arrow keys I should be able to loop through the points in the chart series

#### Actual behaviour

When hitting the last visible points and then using left/right arrows instead of looping an error is thrown

Using keyboard navigation upon reaching the graph points, use the right key to navigate through each of the points instead of looping an error is thrown. Sometimes it will loop through and then only error after doing a few loops.

#### Product version
Highstock 6.1.0

#### Affected browser(s)
Chrome, Firefox and IE

oysteinmoseng:
Keyboard navigation on stock charts failed with errors in some cases. Caused by series with `enableMouseInteraction: false` (such as navigator series).

Fixing this bug immediately unearthed several others, so I added better handling of series that should be ignored by the module, and fixed issues with hiding the export menu when leaving it using the keyboard.

On the topic of hiding the export menu: It seems we currently store a reference to the context menu `div` on `chart['cache-' + className]`, where className can be configured by the user (https://github.com/highcharts/highcharts/compare/bugfix/8440-keyboard-navigation-stock#diff-8c6bf8acef37c90d2cc3aab9b3bf8258L1260). I didn't find any references to this outside of the export module. Any particular reason for the naming? I now added it to `chart.exportContextMenu` as well, in order to pick it up in the keyboard navigation module.

### DIFF
--- a/js/modules/exporting.src.js
+++ b/js/modules/exporting.src.js
@@ -1250,14 +1250,14 @@ extend(Chart.prototype, /** @lends Highcharts.Chart.prototype */ {
             menu = chart[cacheName],
             menuPadding = Math.max(width, height), // for mouse leave detection
             innerMenu,
-            hide,
             menuStyle;
 
         // create the menu only the first time
         if (!menu) {
 
             // create a HTML element above the SVG
-            chart[cacheName] = menu = createElement('div', {
+            chart.exportContextMenu = chart[cacheName] = menu =
+            createElement('div', {
                 className: className
             }, {
                 position: 'absolute',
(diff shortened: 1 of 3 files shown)


js/modules/keyboard-navigation.src.js
```diff
@@ -86,6 +86,29 @@ function stripTags(s) {
...
+ * grouping.
+ *
+ * @param {Point} point The point to find index of.
+ * @return {number} The index in the series.points array of the point.
+ */
+function getPointIndex(point) {
+    var index = point.index,
+        points = point.series.points,
+        i = points.length;
+    if (points[index] !== point) {
```
jon-a-nygaard: Just out of curiousity, in what cases could the `point.index` be wrong?

js/modules/exporting.src.js
```diff
@@ -1283,18 +1283,19 @@ extend(Chart.prototype, /** @lends Highcharts.Chart.prototype */ {
             /*= } =*/
 
             // hide on mouse out
-            hide = function () {
+            menu.hideMenu = function () {
```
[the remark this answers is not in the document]
oysteinmoseng: Just so that we can use it in the keyboard navigation module

js/modules/keyboard-navigation.src.js
oysteinmoseng: `point.index` basically points into `xData`/`yData`, not `series.points`. `series.points` only includes the visible points: https://jsfiddle.net/07hc8jts/

TorsteinHonsi:
Thanks @oysteinmoseng!

Looks like the `cacheName` thing goes all the way back to the beginning, but if it is not used anymore, in the exporting module or any other module files, it should be safe to remove it.

PR #8878, opened 2018-09-03
hs7/stock-tools: Price channel

Denyllon:


Denyllon:
Mistake with merging branches.

PR #8894, opened 2018-09-05
hs7/stock-tools: Price channel.

Denyllon:


samples/stock/indicators/price-channel/demo.html
```diff
@@ -0,0 +1,8 @@
+<script src="https://code.jquery.com/jquery-3.1.1.min.js"></script>
+<script src="https://code.highcharts.com/stock/highstock.js"></script>
+<script src="https://code.highcharts.com/stock/modules/exporting.js"></script>
+<script src="https://code.highcharts.com/stock/indicators/indicators.js"></script>
+<script src="http://code.highcharts.local/stock/indicators/bollinger-bands.js"></script>
```
pawelfus: Wrong URL

js/indicators/price-channel.src.js
```diff
@@ -0,0 +1,125 @@
...
+                ctx = this;
+
+            parentLoaded(
+                BB,
+                'bollinger-bands',
+                ctx.type,
+                function (indicator) {
+                    indicator.prototype.init.apply(ctx, args);
+                }
+            );
```
pawelfus: Can we define this in `indicators.js` to extend this to all indicators that base on others? Then we could define dependancies in `prototype` e.g. `proto.indicatorDependancies = [{ name: 'bollinger-bands', indicator: BB }]` and check this in the core (if array is defined of course).

js/indicators/price-channel.src.js
```diff
@@ -0,0 +1,125 @@
...
+
+            if (xVal.length < period) {
+                return false;
+            }
+
+            for (i = period; i <= yValLen; i++) {
+                date = xVal[i - 1];
+                slicedY = yVal.slice(i - period, i);
+                TL = maxInArray(slicedY, 1);
+                BL = minInArray(slicedY, 2);
```
pawelfus: Can we speed this up? For example just one method call with iterations?

js/indicators/price-channel.src.js
```diff
@@ -0,0 +1,125 @@
...
+     * @sample {highstock} stock/indicators/price-channel Price Channel
+     * @excluding
+     *        allAreas, colorAxis, compare, compareBase, joinBy, keys, stacking,
+     *        showInNavigator, navigatorOptions, pointInterval,
+     *        pointIntervalUnit, pointPlacement, pointRange, pointStart, joinBy
+     * @since 7.0.0
+     * @optionparent plotOptions.pc
+     */
+    {
+        name: 'Price Channel (20)',
```
pawelfus: Name should be build from `nameBase` and `nameComponents`, see for example `ZigZag` props.

js/indicators/price-channel.src.js
```diff
@@ -0,0 +1,125 @@
...
+        /**
+         * @excluding
+         *    index, standardDeviation
+         */
+        params: {},
+        lineWidth: 1,
```
pawelfus: Is that different from default option?

js/indicators/price-channel.src.js
```diff
@@ -0,0 +1,125 @@
...
+                yValLen = yVal ? yVal.length : 0,
+                PC = [], // 0- date, 1-top line, 2-middle line, 3-bottom line
+                ML, TL, BL, // middle line, top line and bottom line
+                date,
+                xData = [],
+                yData = [],
+                slicedY,
+                i;
+
+            if (xVal.length < period) {
```
pawelfus: Can we use `yValLen` here?

js/indicators/price-channel.src.js
```diff
@@ -0,0 +1,125 @@
...
+        name: 'Price Channel (20)',
+        /**
+         * @excluding
+         *    index, standardDeviation
+         */
+        params: {},
+        lineWidth: 1,
+        topLine: {
+            styles: {
+                lineColor: '${palette.colors}'.split(' ')[2]
```
pawelfus: Missing doclet - otherwise it will show something about BB ;) From:

```
                /**
                 * Color of the line. If not set, it's inherited from
                 * [plotOptions.bb.color](#plotOptions.bb.color).
                 *
                 * @type {String}
                 * @since 6.0.0
                 * @product highstock
                 */
                lineColor: undefined
```

samples/stock/indicators/price-channel/demo.html
Denyllon: Fixed.

js/indicators/price-channel.src.js
Denyllon: Yes, it is. By default it would inherit `2`.

js/indicators/price-channel.src.js
Denyllon: Of course. Used `yValLen`.

js/indicators/price-channel.src.js
Denyllon: Doclets added.

js/indicators/price-channel.src.js
Denyllon: Fixed.

js/mixins/reduce-array.js
```diff
@@ -11,15 +11,42 @@ import '../parts/Utilities.js';
...
+     * @param {string} maxIndex Index of "high" value in point array.
+     * @returns {array} Returns array with min and max value.
+     */
+    getArrayExtremes: function (arr, minIndex, maxIndex) {
+        return reduce(arr, function (prev, target) {
+            return [
+                Math.min(prev[0], target[minIndex]),
+                Math.max(prev[1], target[maxIndex])
+            ];
+        }, [Infinity, 0]);
```
pawelfus: Should be `Number.MAX_VALUE` and `-Number.MAX_VALUE`.

js/indicators/price-channel.src.js
```diff
@@ -36,15 +35,32 @@ H.seriesType('pc', 'bb',
         lineWidth: 1,
         topLine: {
             styles: {
+                /**
+                 * Color of the top line. If not set, it's inherited from
+                 * [plotOptions.pc.color](#plotOptions.pc.color).
+                 *
+                 * @type {String}
+                 * @since 7.0.0
```
pawelfus: Do we need to redefine `@since`?

js/indicators/price-channel.src.js
```diff
@@ -72,7 +88,7 @@ H.seriesType('pc', 'bb',
                 slicedY,
                 i;
 
-            if (xVal.length < period) {
+            if (yValLen.length < period) {
```
pawelfus: Are you sure this is correct? Is `yValLen` an array?

js/mixins/reduce-array.js
Denyllon: Fixed.

js/indicators/price-channel.src.js
Denyllon: Removed from both lines definitions.

PR #8905, opened 2018-09-07
Bugfix: Add 'allow' label overflow option in addition to 'justify'.

sophiebremer:
This includes in combination with other branches the following changes:

- annotations.labelOptions.overflow: "allow" | "justify" (previously: "justify" | "none")
- series.dataLabels.overflow: "allow" | "justify" (previously: "justify" | "none")
- xAxis.labels.overflow: "allow" | "justify" (previously: "justify" | false)

In all cases the default behaviour continues to be "justify", and "allow" requires to disable crop.

### DIFF
--- a/js/modules/annotations.src.js
+++ b/js/modules/annotations.src.js
@@ -681,7 +681,7 @@ Annotation.prototype = /** @lends Highcharts.Annotation# */ {
              * How to handle the annotation's label that flow outside the plot
              * area. The justify option aligns the label inside the plot area.
              *
-             * @validvalue ["none", "justify"]
+             * @validvalue ["allow", "justify"]
              * @sample highcharts/annotations/label-crop-overflow/
              *         Crop or justify labels
              **/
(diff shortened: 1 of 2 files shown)


pawelfus:
Shouldn't we change description in docs? For example:
https://github.com/highcharts/highcharts/blob/c670516ce/js/parts/Axis.js#L895-#L908

And here:
https://github.com/highcharts/highcharts/blob/c670516ce/js/parts/Series.js#L1972-#L1984

Is that a breaking change? Should this be included in `"Release notes"` too? For example, if I'm using now `"none"` and upgrade to v7, will my chart's design change?

sophiebremer:
These doclets will be changed in combination with the type changes for TypeScript.

In general the change is backwards compatible, because it only tests agains `"justify"` in the code. I included you for review, because I have no overview, if in some wraps or custom modules you might test the affected options again "none" or false. This would break with the change.

PR #8907, opened 2018-09-07
bugfix/7051-datagrouping-week

pawelfus:
I did it 😄 

The problem is for all our build-in data grouping for weeks (under specific conditions, first grouped point is never calculated&rendered). What happens:

When calculating `groupPositions` for weeks, we start on Monday (`startOfWeek = 1`), so any datapoint (on the beginning of the dataset) that is placed in a previous day will not create extra group, because we translate this min-date back in `getTimeTicks` to Monday here:

https://github.com/highcharts/highcharts/blob/5980b76465ea293a88ccf982fe42fff4aff8c536/js/parts/Time.js#L701-L712

For example: `minDay` is Sunday, so translation goes back to Monday, always:
```
         ( 
             time.get('Date', minDate) -  // any Sunday date
             time.get('Day', minDate) + // returns 0 !
             pick(startOfWeek, 1)  // 1
         ) 
```

Known visual tests that fail (and should fail):
- http://utils.highcharts.local/samples/#test/stock/issues/2696 - chart is no longer disconnected
- http://utils.highcharts.local/samples/#test/stock/issues/datagrouping-series - first point in dataset is actually `new Date(1388538000000) -> Wed Jan 01 2014 02:00:00 GMT+0100 (Central European Standard Time)` and was never rendered 😄

### DIFF
--- a/js/parts/DataGrouping.js
+++ b/js/parts/DataGrouping.js
@@ -505,7 +505,20 @@ seriesProto.groupData = function (xData, yData, groupPositions, approximation) {
 
     // Start with the first point within the X axis range (#2696)
     for (i = 0; i <= dataLength; i++) {
-        if (xData[i] >= groupPositions[0]) {
+        var diff = xData[i] - groupPositions[0];
+        if (diff >= 0) {
+            break;
+        } else if (
+            groupPositions.info.unitName === 'week' &&
+            diff > -groupPositions.info.unitRange / 7
+        ) {
+            // Add an extra position before first position, becasue
+            // getTimeTicks trims first position if it starts on Sunday (#7051)
+            groupPositions.splice(
+                0,
+                0,
+                groupPositions[0] - groupPositions.info.unitRange
+            );
             break;
         }
     }
--- a/samples/unit-tests/series/datagrouping/demo.js
+++ b/samples/unit-tests/series/datagrouping/demo.js
@@ -296,3 +296,37 @@ QUnit.test('Switch from grouped to non-grouped', function (assert) {
     );
 
 });
+
+QUnit.test('Weekly grouped data should not trim Sunday points (#7051)', function (assert) {
+    var chart = Highcharts.stockChart('container', {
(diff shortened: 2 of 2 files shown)


js/parts/DataGrouping.js
```diff
@@ -505,7 +505,20 @@ seriesProto.groupData = function (xData, yData, groupPositions, approximation) {
...
-        if (xData[i] >= groupPositions[0]) {
+        var diff = xData[i] - groupPositions[0];
+        if (diff >= 0) {
+            break;
+        } else if (
+            groupPositions.info.unitName === 'week' &&
+            diff > -groupPositions.info.unitRange / 7
+        ) {
+            // Add an extra position before first position, becasue
+            // getTimeTicks trims first position if it starts on Sunday (#7051)
```
TorsteinHonsi: Isn't this a bug in getTimeTicks? Shouldn't it extend the tick positions to previous Monday instead of trimming the first position? In that case we must patch `getTimeTicks` instead of creating a workaround in DataGrouping.js.

Proposal for a unit test: http://jsfiddle.net/highcharts/1v9kbpy7/

In this case we pass Sunday as the min date, yet the first tick is placed on the next Monday.

pawelfus: Yes, this is actually bug in getTimeTicks, but it only applies to dataGrouping so I decided for a workaround in dataGrouping. I couldn't imagine a case where we need this fixed for time ticks.

TorsteinHonsi: I think this is the case: http://jsfiddle.net/highcharts/zsywLdug/ . The starting point is on Sunday 12:00.

If we change the starting point to Saturday 12:00, the axis will start on the Monday before, which is the correct behaviour: http://jsfiddle.net/highcharts/zsywLdug/2/

pawelfus: Ah, `startOnTick`. Let me create a separate branch and a fix for it.

PR #8914, opened 2018-09-10
Docs: Made legend class official.

sophiebremer:
We might need to hide the readonly flag for the properties in the navigation sidebar of the api documentation. It looks a bit overloaded.


TorsteinHonsi:
> We might need to hide the readonly flag for the properties in the navigation sidebar of the api documentation. It looks a bit overloaded.

Or we can style it our way.

PR #8959, opened 2018-09-14
SvgRenderer.label missing useHTML in JSDoc.

jon-a-nygaard:
# Description
Found that [Highcharts.SVGRenderer#label](https://api.highcharts.com/class-reference/Highcharts.SVGRenderer#label) was missing the parameter `useHTML`.
Copied the description of `useHTML` from [xAxis.labels.useHTML](https://api.highcharts.com/highcharts/xAxis.labels.useHTML).

Feel free to edit the PR if necessary, or request changes.

### DIFF
--- a/js/parts/SvgRenderer.js
+++ b/js/parts/SvgRenderer.js
@@ -4559,28 +4559,28 @@ extend(SVGRenderer.prototype, /** @lends Highcharts.SVGRenderer.prototype */ {
      *         The y position of the label's top side or baseline, depending on
      *         the `baseline` parameter.
      *
-     * @param  {string} shape
+     * @param  {string} [shape='rect']
      *         The shape of the label's border/background, if any. Defaults to
      *         `rect`. Other possible values are `callout` or other shapes
      *         defined in {@link Highcharts.SVGRenderer#symbols}.
      *
-     * @param  {number} anchorX
+     * @param  {number} [anchorX]
      *         In case the `shape` has a pointer, like a flag, this is the
      *         coordinates it should be pinned to.
      *
-     * @param  {number} anchorY
+     * @param  {number} [anchorY]
      *         In case the `shape` has a pointer, like a flag, this is the
      *         coordinates it should be pinned to.
      *
-     * @param  {boolean} useHTML
+     * @param  {boolean} [useHTML=false]
      *         Wether to use HTML to render the label.
      *
-     * @param  {boolean} baseline
+     * @param  {boolean} [baseline=false]
      *         Whether to position the label relative to the text baseline,
      *         like {@link Highcharts.SVGRenderer#text|renderer.text}, or to the
      *         upper border of the rectangle.
      *
-     * @param  {string} className
+     * @param  {string} [className]
      *         Class name for the group.
      *
      * @return {Highcharts.SVGElement}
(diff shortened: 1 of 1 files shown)


jon-a-nygaard:
@sophiebremer Makes sense, will add another commit which solves that

jon-a-nygaard:
@sophiebremer This is not important for this PR, but more as a food for thought.
I came to think that using the `{*|undefined}` pattern for parameters where we have defined a default value, it does not make much sense that users should be able to pass in `undefined`. In cases such as [`useHTML`](https://github.com/highcharts/highcharts/pull/8959/files#diff-65bc0ad38de3454c8d9cf0313d7bb35eR4575) they should rather pass in the default value `false` in stead of `undefined`.
It still makes sense for cases like [`anchorX`](https://github.com/highcharts/highcharts/pull/8959/files#diff-65bc0ad38de3454c8d9cf0313d7bb35eR4567) where I did not specify a default value.

sophiebremer:
@jon-a-nygaard
This represents the behaviour of JavaScript and TypeScript. They actually do not care about the default value. You can always set in this case the parameter to `undefined`, to let the function use the default value (that can change in the future). It is a better practice than using the current valid default.

jon-a-nygaard:
But in cases where we do type checking, the user can pass in all kinds of invalid types as a parameter and it would still fall back to the default value.
Take the following example:
```javascript
/**
 * Multiplies a number by 2.
 * @param {number|*} [a=0] The number to be multiplied by 2.
 * @return {number} Returns the multiplied number.
 */
function timesTwo(a) {
    var x = typeof a === 'number' ? a : 0;
    return x * 2;
}
```
Because of the type checking, any value is "legal" for `a`.
But I still feel the `@param {number|*} [a=0]` is incorrect, since `a` should be a `number`.

TorsteinHonsi:
JSDoc has their own standard for optional parameters, and optional parameters with a default. See http://usejsdoc.org/tags-param.html.

Shouldn't we stick with the standard instead of creating our own? The syntax `{string|undefined}` seems like bloat to me, especially since most parameters are optional in Highcharts.

jon-a-nygaard:
I've had an internal discussion with @sophiebremer about this and I've been somehow convinced that this `{string|undefined}` is correct. And to help on that view it is good to have look at how the new [default parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters#Passing_undefined_vs._other_falsy_values) will behave in ES6 when passing in `undefined` as a value for a parameter.

@TorsteinHonsi I agree that the `{string|undefined}` still looks like bloat, and it would be great if there is a standard in the JS community that we don't have to specify `undefined` for optional parameters, since that is always a valid input for optional parameters.

TorsteinHonsi:
> it would be great if there is a standard in the JS community

According to http://usejsdoc.org/tags-param.html, Closure Compiler uses `{string=}` for optional parameters.

sophiebremer:
Because optional implies undefined as a valid type, we could just mention this in the online documentation (with a change in docstrap). Then we do not need to write it every time in the doclet itself. For beginners it is a helpful information.

sophiebremer:
@TorsteinHonsi
I find the Closure Compiler syntax misleading, because it is mixing type and value.

PR #9002, opened 2018-09-19
feature/data-grouping-options

### LINKED ISSUE #8999: Highstock: in styled mode, column charts lose color if too many bars are displayed
Use case: display a volume pane (like in [this demo](https://www.highcharts.com/stock/demo/candlestick-and-volume)), but add color for the volume (based on the price going up/down), as in many professional charting applications (see [TradingView](https://www.tradingview.com/chart/?symbol=NASDAQ:AAPL)).

I've been trying to do this in styled mode for [too long, IMO](https://github.com/highcharts/highcharts/issues/8941#issuecomment-421101229) due to an undocumented behavior: when Highstock considers the chart is too "narrow", `colorIndex` information is lost and all column bars take the same color. Given that I was using real data, all volume bars appeared black, and I couldn't figure out why, until it occurred to me to artificially limit the number of data points.

http://jsfiddle.net/dandv/rt716yf0/4/

#### Expected behaviour
HighStock should try to leave color information alone for colored bars (e.g. the volume bars in the example). If this isn't possible past a number of bars due to CSS limitations, the behavior should be documented, in at least these places:
* https://www.highcharts.com/docs/chart-design-and-style/style-by-css
* https://api.highcharts.com/highcharts/plotOptions.series.colorIndex
* https://api.highcharts.com/highcharts/series.column.colorByPoint

#### Actual behaviour
The extra confusing bit is that this unexpected behavior only occurs with `colorByPoint: false` (or omitted), which suggests the entire series should have one color. But that's not the case, as can be seen in the beginning of the demo, when only 4 points are defined, or if 200 is changed to 50 in the fiddle. Setting `colorByPoint: true` *does* display the colors per bar, which suggests there's no CSS limitation (or the colors are displayed incorrectly, similarly to setting `crisp: false`. Either way, this should be documented.

#### Other notes
As I suggested in https://github.com/highcharts/highcharts/issues/8941#issuecomment-421101229, given that colored volume charts are a very common use case for stock charts, an official demo would be great.

TorsteinHonsi:
The options from the first point are copied over to the data group. Closes #8999.

### DIFF
--- a/js/parts/DataGrouping.js
+++ b/js/parts/DataGrouping.js
@@ -40,7 +40,11 @@ var addEvent = H.addEvent,
  *
  * If data grouping is applied, the grouping information of grouped
  * points can be read from the [Point.dataGroup](
- * /class-reference/Highcharts.Point#.dataGroup).
+ * /class-reference/Highcharts.Point#.dataGroup). If point options other than
+ * the data itself are set, for example `name` or `color` or custom properties,
+ * the grouping logic doesn't know how to group it. In this case the options of
+ * the first point instance are copied over to the group point. This can be
+ * altered through a custom `approximation` callback function.
  *
  * @product highstock
  * @apioption plotOptions.series.dataGrouping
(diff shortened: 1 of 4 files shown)


TorsteinHonsi:
> Note: We are slowly spamming generatePoints with dataGrouping logic in Highcharts.

Yes... At some point we should consider firing and event that Highstock can pick up.

PR #9006, opened 2018-09-19
bugfix/8995-set-data-duplicated-x

KacperMadej:


js/parts/Series.js
```diff
@@ -3112,10 +3114,10 @@ H.Series = H.seriesType('line', null, { // base series options
                     // are not touched.
                     oldData[pointIndex].touched = true;
 
-                    // Speed optimize by only searching from last known index.
+                    // Speed optimize by only searching after last known index.
                     // Performs ~20% bettor on large data sets.
                     if (requireSorting) {
-                        lastIndex = pointIndex;
+                        lastIndex = pointIndex + 1;
```
KacperMadej: @TorsteinHonsi 
This should resolve the problem for sorted data - checking from next point to avoid using the same point again.

js/parts/Series.js
```diff
@@ -3095,8 +3095,10 @@ H.Series = H.seriesType('line', null, { // base series options
                 // Search for the same X in the existing data set
                 pointIndex = H.inArray(x, this.xData, lastIndex);
 
-                // Matching X not found, add point (but later)
-                if (pointIndex === -1) {
+                // Matching X not found
+                // or used already due to ununique x values (#8995),
+                // add point (but later)
+                if (pointIndex === -1 || oldData[pointIndex].touched) {
```
KacperMadej: @TorsteinHonsi 
`|| oldData[pointIndex].touched` is for series that do not have data sorted - it avoids using the same point again, so a new point is added. This probably could be improved by keep looking for next point with the same x value that was not updated in current `setData` run, but I was unable to find a solution that would be light and fast.

TorsteinHonsi:
Looks good, thanks!

PR #9014, opened 2018-09-20
doc/types-datalabels

sophiebremer:
TypeScript-compatible doclets for internal usage

### DIFF
--- a/js/parts/DataLabels.js
+++ b/js/parts/DataLabels.js
@@ -75,13 +75,11 @@ var addEvent = H.addEvent,
  * @private
  * @function Highcharts.distribute
  *
- * @param  {Array<Highcharts.DataLabelBoxObject>} boxes
+ * @param {Array<Highcharts.DataLabelBoxObject>} boxes
  *
- * @param  {number} len
+ * @param {number} len
  *
- * @param  {number} maxDistance
- *
- * @return {void}
+ * @param {number} maxDistance
  */
 H.distribute = function (boxes, len, maxDistance) {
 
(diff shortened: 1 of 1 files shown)


js/parts/DataLabels.js
```diff
@@ -60,23 +65,21 @@ var addEvent = H.addEvent,
...
  * along a confined length in two dimensions. The algorithm takes an array of
  * objects containing a size, a target and a rank. It will place the labels as
  * close as possible to their targets, skipping the lowest ranked labels if
  * necessary.
  *
+ * @private
  * @function Highcharts.distribute
  *
- * @param  {Array<Highcharts.DataLabelBoxObject>} boxes
+ * @param {Array<Highcharts.DataLabelBoxObject>} boxes
```
TorsteinHonsi: Too specific... It can be any box. Why do we need typing for this internal function in the first place?

sophiebremer: What does any box mean? This type only describes the minimum of required properties and corresponding types. This does not mean, that the provided objects may not contain other properties.

I type everything, to make sure, that the behavior of public functions using internal function is consistent with their doclet. Otherwise I would not know after several function calls what types are passing through the code and in the end are possible parameter types or return types.

TorsteinHonsi: The `distribute` function, in addition to data labels, is used for legend items, split tooltips and flag series. So `DataLabelBoxObject` is too specific.

sophiebremer: Okay, I will reduce the doclets.

PR #9017, opened 2018-09-20
doc/types-html

sophiebremer:
TypeScript-compatible doclets for internal usage


js/parts/Html.js
```diff
@@ -24,11 +26,19 @@ var attr = H.attr,
...
-// Extend SvgElement for useHTML option
+// Extend SvgElement for useHTML option.
 extend(SVGElement.prototype, /** @lends SVGElement.prototype */ {
+
     /**
      * Apply CSS to HTML elements. This is used in text within SVG rendering and
      * by the VML renderer
+     *
+     * @private
+     * @function Highcharts.SVGElement#htmlCss
```
TorsteinHonsi: Is it necessary to explicitly write the function name when we have the `@lends` statement plus the function name in the code?

sophiebremer: I find it necessary to improving the understanding of the code structure and also solve some JSDoc errors. For consistency I do it everywhere, even when JSDoc is getting it right.

You can of course just parse code and try to extract the meaning behind it, but JSDoc is failing especially determing static and instance functions, and sometimes even placing it at the correct class/namespace.

So in explicitly describing the function relation there is no room left for JSDoc to guess.

PR #9038, opened 2018-09-20
feature/sankey-min-node-height

calsupik:
**Adds minNodeHeight option and functionality for sankey chart type.**

minNodeHeight added to options object in sankey chart type.
getSum() node function will now return the max of either the sum
of the links to, sum of the links from, or the minNodeHeight
option set. minNodeHeight defaults to 0.


calsupik:
@TorsteinHonsi Can I get any more detail as to why it may not be working for you? Any errors? I have it working as expected in my local environment. Anything will help me troubleshoot.

As for the translation from weight to pixels, I can look into that. Depending on how this is desired to be implemented, we could also just change the option name/key to minNodeWeight, which would be an accurate representation of the current functionality in this PR.  Again, this is more of a code standard decision for the Highcharts team and if the option minNodeWeight name/key is a viable option.

Either way works for me as I have updated our app specific sankey.src.js file with these changes until they can be merged into master.

TorsteinHonsi:
> Can I get any more detail as to why it may not be working for you?

No errors or nothing, it just didn't adjust the height of the node. Can you set up a live JSFiddle demo where it works?

> As for the translation from weight to pixels, I can look into that.

Thanks. In my opinion, pixels make more sense here, because users don't always have control over the values if they come from a dynamic source. I assume the use case here is similar to [minPointLength](https://api.highcharts.com/highcharts/series.column.minPointLength).

calsupik:
> No errors or nothing, it just didn't adjust the height of the node. Can you set up a live JSFiddle demo where it works?
Here is a live JSFiddle demo that I will be using to show the changes I have made. I have copied the file at code/modules/sankey.src.js in my forked repo into a script tag on the demo.
> Thanks. In my opinion, pixels make more sense here, because users don't always have control over the values if they come from a dynamic source. I assume the use case here is similar to minPointLength.
I agree, pixels will make more sense. I will update the above JSFiddle demo as I make the changes in the code base.

calsupik:
@TorsteinHonsi I have updated the code to allow minNodeHeight be passed in as a pixel value (instead of weight) as mentioned above. I have also updated the JSFiddle demo with the new build file for your review. Please let me know if any additional changes are required.

calsupik:
@TorsteinHonsi Just following up if anything else needs to be adjusted or tested for this PR. Thanks!

joshyudeep:
@calsupik you have some conflicts

PR #9063, opened 2018-09-26
Tools: Add tree.json and tree-namespace.json for wrappers.

sophiebremer:
Make life easier for the developers of the wrappers.


cvasseng:
I feel very strongly that files generated based on the state of the source tree at the time of generation should not be committed.

In this particular case it IMO makes more sense to add `tree_namespace.json` to api.highcharts.com (same as `tree.json`), and have the wrapper devs pull the files from there. This also makes the latest files available for others to use as well (lots of people using `tree.json` and the old `dump.json` for various things).

PR #9068, opened 2018-09-27
Samples blog demo

mekhatria:


samples/highcharts/blog/annotations-aapl-iphone/demo.js
```diff
@@ -0,0 +1,146 @@
...
+                align: 'top'
+            }, {
+                point: {
+                    xAxis: 0,
+                    yAxis: 0,
+                    x: Date.UTC(2016, 9),
+                    y: 113.54
+                },
+                text: 'iPhone 7'
+            }]
```
TorsteinHonsi: Do we also have information about 2017's models?

mekhatria: The new models are added.
The data is also updated.

Issue #8525, opened 2018-06-26
Gantt: Inverted chart fails


oysteinmoseng:
We have use cases for inverted gantt charts. Currently this fails.

#### Live demo with steps to reproduce
http://jsfiddle.net/m3u1bghy/1/

jon-a-nygaard:
**Internal Note:** The inverted yAxis (`type: ['treegrid', 'category]`) actually works as "expected", but it looks off due to `yAxis.staticScale: 50`

jon-a-nygaard:
After a series of fixes I believe we can close this issue.

The following has been fixed:
- Label positioning in grid axis
- borderWidth, lineWidth, and tickWidth behaves the same accross horizontal and vertical axes.
- tickSize could sometimes return wrong length of tick line, causing a gap in the grid cell.

The reproduction demo still looks a bit off, but that is due to `yAxis.staticScale` is defaulting to `50`.

PR #9096, opened 2018-10-01
Bugfix/grid axis ticksize

jon-a-nygaard:
# Description
`Axis.tickSize` returned the wrong length of ticks when the axis is on the top or bottom side. This caused the grid cells to have gaps.
This PR also includes correcting of the vertical alignment of labels with multiple lines within a grid axis, and a unit tests which test vertical align middle, and align center for both svg and html labels.

# Examples
- [Example of behaviour before](http://jsfiddle.net/jon_a_nygaard/gueqL5hd/)
- [Example of behaviour after](http://jsfiddle.net/jon_a_nygaard/nmxsw0ur/)

# Related issue(s)
- #8525


TorsteinHonsi:
I had to revert this because the vertical-alignment tests failed on the Mac. Pixel-testing for text is very inconsistent and must be tested across browsers.

PR #9103, opened 2018-10-02
Highstock: Fixed #9102, hardcoded stochastic series name.

### LINKED ISSUE #9102: Highstock - Stochastic indicator name shows wrong periods (always default)
#### Expected behaviour
The name should show actual periods.

#### Actual behaviour
The name shows always default periods.

http://jsfiddle.net/wchmiel/to3jns9d/1/

#### Product version
Highstock

#### Affected browser(s)
all

wchmiel:


pawelfus:
As long as we don't define `series.name` it should be `Stochastic (a, b)`. Accidentally I have left `series.name` for Stochastic when refactoring names for indicators (all series use `getName()` method when `series.name` is not defined).

PR #9107, opened 2018-10-02
bugfix/9075-addpoint-navigator-sticktomin

### LINKED ISSUE #9075: Highstock multiple series zoom reset on adding point
When zooming into a HighStock multi-series line chart with navigator, the zoom resets or jumps to start of series when adding a new point but only if current zoomed area is missing full data from that series.

What should happen is the same behavior as when all series have data in current zoomed area; the zoomed area stays the same.

Example:
http://jsfiddle.net/7qtbfwu4/10/
To trigger issue: select zoom range lower than all, then drag navigator off secondary series (or until zoomed area contains secondary series start point).

Issue exists up to HighStock v6.1.1 (latest tried)

pawelfus:
Highstock: Fixed #9075, adding a point used to reset zoom when series had different ranges.

Previously, minimum extreme was taken always from first series, now it takes min from all series in navigator.

### DIFF
--- a/js/parts/Navigator.js
+++ b/js/parts/Navigator.js
@@ -2065,6 +2065,24 @@ Navigator.prototype = {
     },
 
     /**
+     * Get minimum from all base series connected to the navigator
+     *
+     * @param  {number} currentSeriesMin
+     *         Minium from the current series
+     *
+     * @return {number} Minimum from all series
+     */
+    getBaseSeriesMin: function (currentSeriesMin) {
+        return H.reduce(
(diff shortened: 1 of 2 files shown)


pawelfus:
Not really, but I was not sure about the second case, where "default" value should is used on (`-Number.MAX_VAUE`) here:

https://github.com/highcharts/highcharts/pull/9107/commits/7bf433607a12d7884a35a26d8a7241931f78e460#diff-22ae6f4417509a4b0157d4e068ccb8fcL2143

PR #9137, opened 2018-10-08
hc7/test-labelposition-failed

KacperMadej:


KacperMadej:
@TorsteinHonsi This only fixes a test, so it should be fine to merge.

PR #9154, opened 2018-10-10
hs7/aroon

wchmiel:
Added new indicators:
- Aroon indicator
- Aroon Oscillator indicator


js/indicators/aroon.src.js
```diff
@@ -0,0 +1,181 @@
...
+    {
+        /**
+         * Paramters used in calculation of aroon series points.
+         * @excluding periods, index
+         */
+        params: {
+            /**
+             * Period for Aroon indicator
+             *
+             * @default 14
```
pawelfus: No need to define default, when it's set below. Similar to type.

js/indicators/aroon.src.js
```diff
@@ -0,0 +1,181 @@
...
+             * @type {number}
+             * @since 7.0.0
+             * @product highstock
+             */
+            period: 14
+        },
+        marker: {
+            enabled: false
```
pawelfus: Any particular reason to disable markers?

js/indicators/aroon.src.js
pawelfus: Similar with `since` - when defined on the parent, no need to define for all children (needs verification).

js/indicators/aroon.src.js
```diff
@@ -0,0 +1,181 @@
...
+                 * @product highstock
+                 */
+                lineColor: undefined
+            }
+        },
+        dataGrouping: {
+            approximation: 'averages'
+        }
+    }, /** @lends Highcharts.Series.prototype */ H.merge(multipleLinesMixin, {
+        nameComponents: ['period'],
```
pawelfus: This is already defined in SMA

js/indicators/aroon.src.js
```diff
@@ -0,0 +1,181 @@
...
+        getValues: function (series, params) {
+            var period = params.period,
+                xVal = series.xData,
+                yVal = series.yData,
+                yValLen = yVal ? yVal.length : 0,
+                AR = [], // 0- date, 1- Aroon Up, 2- Aroon Down
+                xData = [],
+                yData = [],
+                slicedY,
+                low = 2,
```
pawelfus: I assume low and high indexes are fixed in algorithm. In that case we should check if `yData` is an array of numbers (for reference see RSI).

samples/stock/indicators/aroon/demo.details
```diff
@@ -0,0 +1,6 @@
+---
+ name: Highstock Demo
+ authors:
+   - Pawel fus
```
pawelfus: Thanks! ;)

samples/stock/indicators/aroon/demo.js
```diff
@@ -0,0 +1,51 @@
...
+            linkedTo: 'aapl',
+            color: 'green',
+            lineWidth: 1,
+            aroonDown: {
+                styles: {
+                    lineColor: 'red'
+                }
+            },
+            params: {
+                period: 25
```
pawelfus: If period=25 is a common way to use in this indicator, then please set this in default options.

test/karma-files.json
```diff
@@ -66,6 +66,7 @@
 	"code/indicators/roc.src.js",
 	"code/indicators/rsi.src.js",
 	"code/indicators/stochastic.src.js",
+	"code/indicators/aroon.src.js",
```
pawelfus: Not a big deal, but could you (unless this should be here) place this in alphabetical order?

js/indicators/aroon.src.js
```diff
@@ -0,0 +1,181 @@
...
+                    return elem;
+                }));
+
+                xHigh = maxIndexInArray(slicedY.map(function (elem) {
+                    if (elem[high]) {
+                        return elem[high];
+                    }
+                    return elem;
+                }));
+
```
sebastianbochan: Consider to optymize this both functions which seems to me more or less the same.

js/indicators/aroon.src.js
wchmiel: When markers are enabled, only one line has it.

js/indicators/aroon.src.js
wchmiel: Aroon indicator is a difference between low or high value and period. So it can be calculated also for a simple line chart. Low or high indexes are not required.

pawelfus: Thanks!

js/indicators/aroon.src.js
```diff
@@ -4,18 +4,24 @@ import multipleLinesMixin from '../mixins/multipe-lines.js';
...
-// Index of element with maximal value from array
-function maxIndexInArray(arr) {
-    var maxValue = Math.max.apply(null, arr);
-    return arr.lastIndexOf(maxValue);
-}
+    for (i = 1; i < arr.length; i++) {
+        if (
+            extreme === 'max' && arr[i] >= extremeValue ||
+            extreme === 'min' && arr[i] <= extremeValue
+        ) {
```
sebastianbochan: Powinno zadzialac: `arr[i] >= extremeValue && (extreme === 'max' || extreme === 'min')`

PR #9172, opened 2018-10-12
hs7/stock-tools-keltner-channels

Denyllon:


js/indicators/keltner-channels.src.js
```diff
@@ -0,0 +1,196 @@
+'use strict';
+
+import H from '../parts/Globals.js';
+import '../parts/Utilities.js';
+import './ema.src.js';
+import './atr.src.js';
```
pawelfus: We don't load indicators anymore - user should do this + we have an error when indicator is not found.

sebastianbochan: In MACD we loading it -> https://github.com/highcharts/highcharts/blob/master/js/indicators/macd.src.js#L4

pawelfus: Yes, MACD needs this change in v7 too - we will stop doing this. Now multiple indicators use EMA, so we don't want to load more than needed. It will artificially increase packages/bundles sizes.

PR #9194, opened 2018-10-16
doc/types-modules-offline-exporting

bre1470:
TypeScript-compatible doclets.


js/modules/offline-exporting.src.js
```diff
@@ -6,11 +6,54 @@
  * License: www.highcharts.com/license
  */
 
+/**
+ * Download options for offline exporting.
+ *
+ * @typedef Highcharts.OfflineExportingDownloadOptionsObject
```
TorsteinHonsi: I believe these are the same options that are available for the online exporting, so they should probably use the same typedef.

js/modules/offline-exporting.src.js
```diff
@@ -6,11 +6,54 @@
...
+ *
+ * @property {number} [scale]
+ *           Scaling factor of downloaded image compared to source
+ *
+ * @property {string} [type]
+ *           File type of resulting download
+ */
+
+/**
+ * @typedef Highcharts.OfflineExportingOptionsObject
```
TorsteinHonsi: I believe these are the same options that are available for the online exporting, so they should probably use the same typedef.

sophiebremer:
I changed the dependencies.

js/modules/offline-exporting.src.js
```diff
@@ -6,11 +6,34 @@
  * License: www.highcharts.com/license
  */
 
+/**
+ * Download options for offline exporting.
+ *
+ * @interface Highcharts.OfflineExportingOptionsObject
```
TorsteinHonsi: I don't think we need a specific options object for offline exporting. The `libURL` option is the only thing missing in online, while `scale` already exists in online. We should rather add these to the `Highcharts.ExportingOptionsObject`.

bre1470:
I removed the interfaces.

PR #9219, opened 2018-10-19
StockTools: Absolute Price Oscillator.

wchmiel:


js/indicators/apo.src.js
```diff
@@ -0,0 +1,137 @@
...
+         * @excluding period
+         */
+        params: {
+            /**
+             * Periods for Absolute Price Oscillator calculations.
+             * @default [10, 20]
+             * @type {Array}
+             * @since 7.0.0
+             */
+            periods: [10, 20]
```
sebastianbochan: Should we use array of periods instead of two separate periods like in MACD -> https://github.com/highcharts/highcharts/blob/master/js/indicators/macd.src.js#L40 ?

In this case it will be periodSPE and periodLPE.

Add short info in docs, which periods are.

@pawelfus what do you think?

pawelfus: For example Stochastic has `periods`, `macd` has multiple periods, so we don't have strict rules here..

sebastianbochan: Ok clear!

PR #9241, opened 2018-10-24
hc7/3d cylinder

KacperMadej:


KacperMadej:
@TorsteinHonsi When this PR will be merged into hc7/master then we'll know for sure the code is accepted and I could start working on pyramid3d and funnel3d modules based on this code. 3rd step could be optimizing code - moving shared Renderer functions to a mixin.

tl;dr This PR should be resolved before pyramid3d and funnel3d

PR #9256, opened 2018-10-27
fixes linebreak in documentation link

dimitropoulos:
otherwise the docs (now) read

```
For solid gauges, the Y axis also inherits the concept of [data classes](http://api. highcharts.com/highmaps#colorAxis.dataClasses) from the Highmaps color axis
```

### DIFF
--- a/js/parts/Axis.js
+++ b/js/parts/Axis.js
@@ -2408,9 +2408,9 @@ H.extend(Axis.prototype, /** @lends Highcharts.Axis.prototype */{
          * item is a float between 0 and 1 assigning the relative position in
          * the gradient, and the second item is the color.
          *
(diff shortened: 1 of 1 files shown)


TorsteinHonsi:
Thanks!

dimitropoulos:
@TorsteinHonsi and, as an aside, this seems to be a systemic problem in the docs (I just made PRs for the ones I saw).  I'll leave it to you to decide how to proceed, but you may want to make a github issue to try to somehow grep through the comments somehow and see where else this is problem is occurring.  You guys have a lot of docs, heh.

TorsteinHonsi:
Actually we do have a [lint rule intended for this](https://github.com/highcharts/highcharts/commit/37d461bd11e6e9867a82f35bbffd002f2241335f), but apparently it doesn't catch all cases.

PR #9140, opened 2018-10-08
Feature/gantt dragdrop. Closes #8584.

oysteinmoseng:
Adds drag/drop functionality to Highcharts. The old `draggable-gantt` module is now deleted and obsolete.

Options:

```javascript

series: {
    dragDrop: {
        draggableX: true, // Enable dragging in the x-dimension
        draggableY: true, // Enable dragging in the y-dimension
        draggableEnd: true, // Also other options, such as draggableStart/draggableHigh/draggableQ1 etc. Enable dragging a specific data property.
        liveRedraw: true, // Update points as they are dragged. If false, a guide box is drawn to illustrate the drag.
        
        // Options for the guide box
        guideBox: {
            // Guide box default state. This is the only state we currently support
            default: {
                className: '', // The class name of the guide box
                lineColor: '', // Color of the border around the guide box
                lineWidth: 1, // Width of the border around the guide box
                color: '', // Fill color of the guide box
                cursor: 'move', // Cursor of the guide box
                zIndex: 900 // zIndex of the guide box
            }
        },

        dragMinX: 0, // Minimum x value we can drag to
        dragMaxX: 0, // Maximum x value we can drag to
        dragMinY: 0, // Minimum y value we can drag to
        dragMaxY: 0, // Maximum y value we can drag to
        dragPrecisionX: 0, // Precision when dragging along the x axis
        dragPrecisionY: 0, // Precision when dragging along the y axis
        dragSensitiviy: 1, // Number of pixels you need to drag before it registers
        groupBy: '', // Property to group points together by
        zoomKey: '', // Key to hold down to allow zooming without triggering drag events

        // Options for the drag handles
        dragHandle: {
            className: '', // Class name for the drag handles
            pathFormatter: function () {}, // Function that defines SVG path to use for drag handles
            color: '', // Fill color for the drag handles
            lineColor: '', // Line color for the drag handles
            lineWidth: 1, // Width of the drag handle lines
            cursor: '' // Cursor for the drag handles        
        }
    }
}
```
In addition, there are events for drag and drop, allowing override of the functionality.

PR #9263, opened 2018-10-29
Fixed #9258, remove use of id in treemap data labels.

### LINKED ISSUE #9258: treemap render emtpy-string label wrong
#### Actual behaviour
https://jsfiddle.net/gc3rx9fo/

#### Product version

6.2.0

jon-a-nygaard:
# Description
Currently `point.id` is used as a fallback when `point.name` is not defined, which caused unexpected issues after [point.id started to get assigned a value by default](https://github.com/highcharts/highcharts/pull/9140). This PR solves the issue by removing the use of `point.id` in the data labels.

Additionally, it also adds a unit test which checks that default options for dataLabels in treemap is assigned the correct values.

# Example(s)
- [current behaviour](https://jsfiddle.net/gc3rx9fo/)
- [new behaviour](https://jsfiddle.net/jon_a_nygaard/w0t9mnzs/)

# Related issue(s)
- Closes #9258

### DIFF
--- a/js/modules/treemap.src.js
+++ b/js/modules/treemap.src.js
@@ -170,9 +170,7 @@ seriesType('treemap', 'scatter', {
         enabled: true,
         defer: false,
         verticalAlign: 'middle',
-        formatter: function () { // #2945
-            return this.point.name || this.point.id;
-        },
+        format: '{point.name}',
         inside: true
     },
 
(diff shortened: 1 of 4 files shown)


jon-a-nygaard:
@TorsteinHonsi Please see my additions to the PR.

PR #9306, opened 2018-11-07
Fixed #9298, zooming with dragdrop.

### LINKED ISSUE #9298: Zooming is not possible when some series is draggable
#### Expected behaviour
Zooming is possible even though draggableX/Y is true for some series in the graph

#### Actual behaviour
Zooming is not possible (the selection event is not even firing) when some series has draggableX/Y true

Here you see a presumably zoomable chart, but the draggableY is set to true, so the selection event doesn't even fire: http://jsfiddle.net/sxk32aoL/

Here is the exact same thing, but the draggableY is set to false, so the zooming is possible (and naturally the selection even fires upon zooming and resetting zoom): http://jsfiddle.net/ko1jLtfp/1

#### Product version
Highcharts 6.2.0

#### Affected browser(s)
Tested only on chrome

oysteinmoseng:
Zoom was prevented even when we didn't have a draggable point on mousedown.

Added zoom to one of the visual tests, and made note to remember to test this.

### DIFF
--- a/samples/highcharts/dragdrop/drag-xrange/demo.js
+++ b/samples/highcharts/dragdrop/drag-xrange/demo.js
@@ -14,10 +14,6 @@ Highcharts.chart('container', {
         text: 'Highcharts draggable xrange demo'
     },
 
-    subtitle: {
-        text: 'Test zoom and dragging/resizing of points'
-    },
-
     tooltip: {
         headerFormat: '<span style="font-size: 10px">{point.yCategory}</span><br/>',
         pointFormat: '{point.name}'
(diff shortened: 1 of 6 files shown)


samples/highcharts/dragdrop/drag-xrange/demo.js
```diff
@@ -6,13 +6,18 @@ var setDragStatus = function (status) {
...
     },
 
     title: {
         text: 'Highcharts draggable xrange demo'
     },
 
+    subtitle: {
+        text: 'Test zoom and dragging/resizing of points'
+    },
+
```
TorsteinHonsi: Internal testing notes shouldn't bloat the demo (this is available from the API). Instead of this, it would be better if we add the instructions to a `test-notes.md` file.

PR #9331, opened 2018-11-07
hs7/chaikin-oscillator

wchmiel:


js/indicators/chaikin.src.js
```diff
@@ -0,0 +1,141 @@
...
+            SPE = EMA.prototype.getValues.call(this, ADL, {
+                period: periods[0]
+            });
+
+            LPE = EMA.prototype.getValues.call(this, ADL, {
+                period: periods[1]
+            });
+
+            // Check if ema is calculated properly, if not skip
+            if (!SPE || !LPE) {
```
sebastianbochan: How about verifing ADL values?

js/indicators/chaikin.src.js
```diff
@@ -0,0 +1,141 @@
...
+                return false;
+            }
+
+            periodsOffset = periods[1] - periods[0];
```
sebastianbochan: How about situation when user set periods: [10,3] ?

samples/unit-tests/indicator-chaikin/recalculations/demo.js
```diff
@@ -0,0 +1,174 @@
+
```
sebastianbochan: In a few places, you use chart.series[0], chart.series[1], chart.series[2]. Consider to use variables.

js/indicators/chaikin.src.js
wchmiel: There is an if statement above that check it: _line 70_ `if (periods.length !== 2 || periods[1] <= periods[0])`.

sebastianbochan:
Now is fine, so we can merge.

PR #9373, opened 2018-11-09
hc7/debugger

ppotaczek:
Displaying errors on the chart.


bre1470:
Yes, JSON and JS are generated. We need to commit them because of the source dependency that would otherwise break the download builder: https://www.highcharts.com/download

bre1470:
I do not think that we need JSON though. It will be also available under api.highcharts.com.

bre1470:
A notice: I did not add the markdown files to the watch list of the default gulp task, so if you make changes to the text, you have to restart the task.

PR #9440, opened 2018-11-19
doc/types-indicators-bb

bre1470:
TypeScript-compatible doclets.


js/indicators/bollinger-bands.src.js
```diff
@@ -26,16 +27,26 @@ function getStandardDeviation(arr, index, isOHLC, mean) {
     return std;
 }
 
+/**
+ * Bowlinger Bands series type.
```
TorsteinHonsi: Bollinger

PR #9507, opened 2018-11-26
Hs7/macd-refactoring

sebastianbochan:


sebastianbochan:
Too many commits inside PR.

PR #9510, opened 2018-11-26
hc7/packed-bubbles-series

sebastianbochan:
To remove branches:
- packed-bubble
- packed-bubbles


js/parts-more/PackedBubbleSeries.js
```diff
@@ -516,6 +501,11 @@ H.addEvent(H.Chart, 'beforeRedraw', function () {
     }
 });
 
+// Hide all axes, we cannot use isCartesian, because it breaks all updates.
```
TorsteinHonsi: Can't do that, it will make every axis invisible on all charts in the page. It's better to use `isCartesian` and find out why the updates break.

PR #9530, opened 2018-11-28
hc7/series type venn diagram

### LINKED ISSUE #6875: Implement Venn diagram series type
Implement Venn diagram series type.

Read for examples and Uservoice's FR here: https://highcharts.uservoice.com/forums/55896-highcharts-javascript-api/suggestions/878283-venn-diagram

### LINKED ISSUE #9358: Center venn labels in set
The labels in a venn diagram should be centered in the non-overlapping area, and not in their shape. Following example below shows that the labels "A" and "B" are shifted towards the sides and away from the union.

Child of #6875

### LINKED ISSUE #9359: Implement venn diagram positioning algorithm
Calculate the positions of sets in a venn diagram.
[Ben Frederickson](https://www.benfrederickson.com) has done some great work here which may be of inspiration.

Child of #6875

### LINKED ISSUE #9360: Calculate the overlapping shapes in venn diagram
Calulate the overlapping shapes in a venn diagram to support additional interaction.

### LINKED ISSUE #9362: Add venn diagram API docs
Implement all the JSDoc doclets required for the venn series type.

Child of #6875

jon-a-nygaard:
# Description
Adds a new series type named `venn`.

The layout uses a greedy algorithm which sort the sets by the amount of overlap, before positioning them. The overlap is only calculated from relations including two sets.

The series type can also be used to visualize Euler diagrams.

## Data structure
```javascript
Highcharts.chart(..., {
...
series: [{
  data: [{
    // Set the area of A to a size of 12
    sets: ['A'],
    value: 12
  }, {
    // Set the area of B to a size of 12
    sets: ['B'],
    value: 12
  }, {
    // Set the area of overlap between A and B to be 6
    sets: ['A', 'B'],
    value: 6
  },]
}],
...
});
```

## Supported features
The following features has been adapted to the venn diagram series.
- Data labels
- Interaction
- Tooltip

# Example(s)
- [Euler Diagram](http://utils.highcharts.local/samples/#view/highcharts/demo/euler-diagram)
- [Venn Diagram](http://utils.highcharts.local/samples/#view/highcharts/demo/venn-diagram)

# Related issue(s)
- Closes #9359 
- Closes #9358 
- Closes #9360 
- Closes #9362 
- Closes #9362 
- Closes #6875


js/mixins/geometry.js
```diff
@@ -0,0 +1,17 @@
+/**
```
TorsteinHonsi: Is it necessary to place this code in a separate file, used only once? It's basically just one line of code. We're doing the same operation inline in for example tooltip placement, click handling and series label.

js/mixins/geometry-circles.js
```diff
@@ -0,0 +1,47 @@
+/**
```
TorsteinHonsi: This mixin is used only once, is it necessary to add complexity by putting it in a separate file? If we want to reuse it in another place in the future, we can separate it out then.

js/modules/venn.src.js
```diff
@@ -0,0 +1,742 @@
...
+import geometry from '../mixins/geometry.js';
+import geometryCircles from '../mixins/geometry-circles.js';
+import H from '../parts/Globals.js';
+import '../parts/Series.js';
+var color = H.Color,
+    extend = H.extend,
+    getDistanceBetweenPoints = geometry.getDistanceBetweenPoints,
+    getOverlapBetweenCirclesByDistance =
+        geometryCircles.getOverlapBetweenCircles,
+    isArray = H.isArray,
```
TorsteinHonsi: Deprecated, use `Array.indexOf` instead.

js/modules/venn.src.js
```diff
@@ -0,0 +1,742 @@
...
+    isBoolean = function (x) {
+        return typeof x === 'boolean';
+    },
+    isNumber = H.isNumber,
+    isObject = H.isObject,
+    isString = H.isString,
+    keys = H.keys,
```
TorsteinHonsi: Deprecated, use `Object.keys` instead.

js/modules/venn.src.js
```diff
@@ -0,0 +1,742 @@
...
+            params.optionsSeries.dataLabels :
+            {}
+        ),
+        options = merge({
+            style: {}
+        }, optionsSeries, optionsPoint);
+    return options;
+};
+
+var vennOptions = {
```
TorsteinHonsi: Where are the doclets? Test with `gulp jsdoc`.

TorsteinHonsi:
In addition to the inline comments. I know some of these are planned, but I write them here anyway as part of a full review:

 - [ ] In the basic demo, the circles aren't centered in the plot area. Shouldn't that be covered by the `getScale` function?
 - [ ] Data labels are currently centered in the circle, but should avoid other circles.
 - [ ] The intersections should be hoverable.
 - [ ] `chart.inverted` looks weird. Probably shouldn't affect this chart type at all.
 - [ ] Select state not working. (Set `series.allowPointSelect` to test).
 - [ ] Doesn't animate on redraws. (Test by running `point.update`).
 - [ ] Styled mode (`chart.styledMode = true`): The circles have no color.
 - [ ] Fails with JS error on pure individual sets, no overlap.
 - [ ] Initial animation?
 - [ ] `chart.zoomType` does something, but not what I expected... Does the venn series need axes at all? Removing axes would also fix the `chart.inverted` issue.

js/modules/venn.src.js
jon-a-nygaard: Doclets are added to the branch, and api of venn series type has been reviewed.

js/mixins/geometry-circles.js
jon-a-nygaard: I guess this depends on preferences, but I in my opinion it is less complex to have this logic seperated, easier to maintain, and that it makes it more likely to be reused if it is already seperated.

PR #9536, opened 2018-11-28
hc7/merge

bre1470:
I had to re-apply three bugfixes. There might be more bugs in the visual tests, which I did not checked.


bre1470:
Includes now working JSDoc for debugger module.

PR #9548, opened 2018-11-29
hs7/detrended-price-oscillator

wchmiel:
Added detrended price oscillator (DPO).


js/indicators/dpo.src.js
```diff
@@ -0,0 +1,126 @@
...
+            for (j = 0; j <= yValLen - range; j++) {
+                periodIndex = j + period - 1;
+                rangeIndex = j + range - 1;
+
+                lastPeriodPrice =
+                    pick(yVal[periodIndex][index], yVal[periodIndex]);
+                currentPrice =
+                    pick(yVal[rangeIndex][index], yVal[rangeIndex]);
+
+                sum = correctFloat(sum + lastPeriodPrice);
```
sebastianbochan: Line 79 / 82 are the same like 66/67, consider to optimize it by i.e function

PR #9575, opened 2018-12-01
hs7/required-indicators-mixin-improvement

Denyllon:
- Checking indicators dependencies added by `requiredIndicators` parameter.
- From now on, user should care about loading all required indicators manually. If some indicator is missing, there is error thrown in browsers console. This type of approach prevents loading the same indicator's code many times.
- All existing indicators are adjusted to this approach.


Denyllon:
Right, we're adjusting it now. Should be fixed after moment. Thank you!

Denyllon:
@TorsteinHonsi it's fixed now. Please merge #9582.

PR #9608, opened 2018-12-05
Fixed issues in sonification with hidden series.

oysteinmoseng:
Hiding a series no longer crashes sonification, but mutes the series being hidden.

If all series are hidden, the sonification stops.

Ideally we would want this to be somewhat smarter and also mute Earcons on the series that is being hidden. We will also want to handle sequentially sonified series smarter, as you would probably expect the sonification to jump to the next series if you hid a series in that case. Starting sonification with a hidden series also skips this series completely, and it would be nice to be able to show it mid-sonification and hear it (for simultaneous series). For now, the current behavior should be acceptable. I will add a bug report for the enhancements mentioned.


oysteinmoseng:
#9609

PR #9624, opened 2018-12-07
StockTools: fixed classList access.

sebastianbochan:


TorsteinHonsi:
Already merged, thanks!

PR #9655, opened 2018-12-12
Stocktools/fixed-ignoring-padding-by-toolbar.

sebastianbochan:


TorsteinHonsi:
@sebastianbochan I did some refactoring in https://github.com/highcharts/highcharts/pull/9655/commits/558a6ca704e861734216117a20dc04127d20d7ba because there were conflicts with the `chart.margin` and `chart.spacing` setttings. They can be numbers or arrays, plus they can be set individually like `chart.marginLeft` or `chart.spacingLeft`.

Can you test and approve these changes? There's now a small initial animation of the range selector buttons, but I don't think that matters.

sebastianbochan:
Will test

sebastianbochan:
I tested it and if the animation is not big deal, we can merge with master.

PR #9661, opened 2018-12-13
feature/new-waterfall-stacking

PaulDalek:
Modified existing waterfall stacking feature by adding new types of stacking (normal and overlap).


js/parts/Series.js
```diff
@@ -960,8 +960,9 @@ null
      * Whether to stack the values of each series on top of each other. Possible
      * values are `undefined` to disable, `"normal"` to stack by value or
      * `"percent"`. When stacking is enabled, data must be sorted in ascending
-     * X order. A special stacking option is with the streamgraph series type,
-     * where the stacking option is set to `"stream"`.
+     * X order. There are also two special stacking options. The first one is
+     * reserved for streamgraph series type, where the stacking option is
+     * set to `"stream"` and `"overlap"` prepared for waterfall series type.
```
TorsteinHonsi: Better sentence:
> ... where the stacking option is set to `"stream"`. The second one is `"overlap"`, which only applies to waterfall series.

js/parts/Series.js
```diff
@@ -960,8 +960,9 @@ null
...
      *
```
TorsteinHonsi: Add `@sample` tags for the two new waterfall stacking samples.

js/parts/Series.js
```diff
@@ -982,6 +982,10 @@ null
      *         Bar
      * @sample {highcharts} highcharts/plotoptions/series-stacking-percent-area/
      *         Area
+     * @sample {highcharts} highcharts/plotoptions/series-stacking-waterfall-normal
+     *         Column
```
TorsteinHonsi: Wrong text. Should be Waterfall with normal stacking

js/parts/Series.js
```diff
@@ -982,6 +982,10 @@ null
...
+     * @sample {highcharts} highcharts/plotoptions/series-stacking-waterfall-overlap
+     *         Column
```
TorsteinHonsi: Wrong text. Should be Waterfall with overlap stacking

js/parts/Series.js
TorsteinHonsi: The _name_ still needs to be changed. "Column" is not the correct name.

js/parts/Series.js
TorsteinHonsi: The _name_ still needs to be changed. "Column" is not the correct name.

js/parts/Series.js
PaulDalek: Ok, sorry for the misunderstanding. Corrected.

PR #9672, opened 2018-12-14
docs/3590/added-chart-argument

ppotaczek:
Added chart argument description in Highcharts.error method.

### DIFF
--- a/js/error-messages.js
+++ b/js/error-messages.js
@@ -123,7 +123,6 @@ H.errorMessages = {
             "errors/28/readme.md",
             "errors/29/readme.md",
             "errors/30/readme.md"
-        ],
-        "version": "7.0.0"
+        ]
     }
 };
--- a/js/error-messages.js
+++ b/js/error-messages.js
@@ -124,6 +124,6 @@ H.errorMessages = {
             "errors/29/readme.md",
             "errors/30/readme.md"
         ],
-        "version": "6.2.0"
+        "version": "7.0.0"
     }
 };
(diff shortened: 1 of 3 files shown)


js/error-messages.js
```diff
@@ -124,6 +124,6 @@ H.errorMessages = {
             "errors/29/readme.md",
             "errors/30/readme.md"
         ],
-        "version": "6.2.0"
```
KacperMadej: This change is against the "DO NOT EDIT!" rule placed in this file.
https://github.com/highcharts/highcharts/blob/ec1b1667a800dcf9eead053b89debc21cdeb772c/js/error-messages.js#L8

@bre1470 btw. Could you take a look? Running `gulp` updates this file and the file's current content on the remote seems not up-to-date. How and when will be this file updated?

bre1470:
This comes from `./tools/error-messages.js`
```js
            let parsedErrors = {
                meta: {
                    files: [],
                    version: require(rootPath  + '/package.json').version
                }
            };
```
Is this wrong, @TorsteinHonsi ?

bre1470:
Oh, I realize now the problem. ErrorMessages rewrites the JSON files only, if the makrdown files change. I could add, that changes the JSON file also when the package version has changed. Should I add this? Or should version get removed?

TorsteinHonsi:
Maybe just remove the version? I don't think we need it - if someone requires a specific version it could be loaded directly from a specific GitHub tag or from `code.highcharts.com/{version}`?

bre1470:
If this is fine with Highcharts Cloud, I will just remove it. Objections, @cvasseng and @kai2468 ?

PR #9695, opened 2018-12-17
bugfix/8160-treemap-datalabels-overflow

kamilkulig:


TorsteinHonsi:
Thanks for the suggestion, I did some  changes and reapplied this fix to the master due to major white-space changes there.

PR #9737, opened 2018-12-20
bugfix/Ichimoku-fixed-missing-points-in-the-Chikou-line.

sebastianbochan:


sebastianbochan:
I noticed that missed "Fixed 9736," in the commit's title.

PR #9783, opened 2019-01-03
bugfix/9235-liveredraw-scrollbar

### LINKED ISSUE #9235: Issue with ScrollBar in HighStocks
#### Expected behaviour
Highstocks trigger the redraw or render function after the scroll bar moved.

#### Actual behaviour
**But I don't want to trigger the redraw or render function after the scroll bar moved**. Because I setting axis maximum value inside the render function.

http://jsfiddle.net/qfhc09z7/

#### Product version

Highstock 6.2.0
#### Affected browser(s)
All browsers

pawelfus:
Highstock: Fixed #9235, option `Axis.scrollbar.liveRedraw` did not block events and redraw.

ESLint has thrown errors like this:

>   57:17  error  Expected property shorthand  object-shorthand
>   95:21  error  Expected method shorthand    object-shorthand

So functions are now defined as:
```javascript
             chart: {
                 events: {
                     redraw() {
                         iterator++;
                     }
                 }
             },
```
AFAIK this won't work in old IE. Is that expected @TorsteinHonsi ?

### DIFF
--- a/js/parts/Scrollbar.js
+++ b/js/parts/Scrollbar.js
@@ -1008,7 +1008,28 @@ addEvent(Axis, 'afterInit', function () {
                 from = unitedMin + range * (1 - this.to);
             }
 
-            axis.setExtremes(from, to, true, false, e);
+            if (
+                pick(
+                    this.options.liveRedraw,
+                    H.svg && !H.isTouchDevice && !this.chart.isBoosting
+                ) ||
+                // Mouseup always should change extremes
+                e.DOMType === 'mouseup' ||
+                // Internal events
+                !defined(e.DOMType)
+            ) {
+                axis.setExtremes(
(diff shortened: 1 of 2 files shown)


TorsteinHonsi:
> AFAIK this won't work in old IE. Is that expected @TorsteinHonsi ?

No, we can't use this yet. It is a side-effect of us updating ESLint right before Christmas. I did a commit yesterday that takes care of it: https://github.com/highcharts/highcharts/commit/6d9b6de1de98821baaa17b8f01658372e9184d35. I'll do the merge and fix it in my end.

PS: ESLint is now set up to auto-update, so we may have situations in the future where a new or changed rule suddenly changes stuff for us. Then we should do a separate lint commit before committing the actual change.

PR #9789, opened 2019-01-03
feature/Add the series index to classname of series labels

jostermanAtPEW:
Adding the series index to the classname of series labels allows for more targeted CSS styling.


TorsteinHonsi:
Thank you for the PR! I'm committing a slightly modified change that also adds a general class name plus the series' `className` option.

jostermanAtPEW:
Great, thanks!

PR #9802, opened 2019-01-04
Fixed #9801, networkgraph not compatible with no-data module.

### LINKED ISSUE #9801: No Data module not working correctly with Network Graph
#### Expected behaviour

When rendering a `networkgraph` with the `no-data-to-display` module loaded, the "No data to display" message should not show if the graph has data.

#### Actual behaviour

The `networkgraph` always displays the "No data to display" message even if it has data showing.

I assume this behavior is occurring because the `hasData` function needs a different implementation for `networkgraph`s.

trentmwillis:
This PR fixes the issue described in #9801. The solution was to add `networkgraph` to the list of series types that use a `hasData` function that simply checks if any `points` exist in the series.

### DIFF
--- a/js/modules/no-data-to-display.src.js
+++ b/js/modules/no-data-to-display.src.js
@@ -136,6 +136,7 @@ defaultOptions.noData = {
     'bubble',
     'gauge',
     'heatmap',
+    'networkgraph',
     'pie',
     'sankey',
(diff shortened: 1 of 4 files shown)


TorsteinHonsi:
Thanks for your fix!

PR #9811, opened 2019-01-07
Fixed incorrect calculations in getOverlapBetweenCircles

### LINKED ISSUE #9718: Venn chart displayed incorrectly
#### Expected behaviour
With 2 sets:
- A: 20
- B: 30
- And section relation: 15
I would like to see chart displayed like in the picture:
venn

#### Actual behaviour
Venn chart is displayed incorrectly: [jsFiddle](https://jsfiddle.net/BlackLabel/akcyLohm/)

Note that exported chart looks different.

#### Product version
Highcharts 7.0.0

#### Affected browser(s)
All

### LINKED ISSUE #9768: Venn chart displayed incorrectly for 3 overlapping sets
#### Expected behaviour
image

#### Actual behaviour
image

https://jsfiddle.net/k2xvL7y6/

#### Product version
Highmaps 7.0.1

#### Affected browser(s)
Latest google chrome.

jon-a-nygaard:
# Description
In certain cases the previous formula of `getOverlapBetweenCircles` was returning incorrect values causing the positioning of the circles to fault. This is solved by modifying `getOverlapBetweenCircles` to calculate the overlap by first calculate the area of each circular segment and adding these together to get a correct result.

# Example(s)
- [Demo with fix applied](https://jsfiddle.net/jon_a_nygaard/3o798y2v/)
- [Demo of the issue](https://jsfiddle.net/k2xvL7y6/)

# Related issue(s)
- Closes #9718 
- Closes #9768

### DIFF
--- a/js/mixins/geometry-circles.js
+++ b/js/mixins/geometry-circles.js
@@ -9,6 +9,19 @@ var round = function round(x, decimals) {
 };
 
 /**
+ * Calculates the area of a circular segment based on the radius of the circle
+ * and the height of the segment.
+ * See http://mathworld.wolfram.com/CircularSegment.html
+ *
+ * @param {number} r The radius of the circle.
+ * @param {number} h The height of the circular segment.
+ * @returns {number} Returns the area of the circular segment.
+ */
+var getCircularSegmentArea = function getCircularSegmentArea(r, h) {
+    return r * r * Math.acos(1 - h / r) - (r - h) * Math.sqrt(h * (2 * r - h));
+};
+
+/**
  * Calculates the area of overlap between two circles based on their radiuses
  * and the distance between them.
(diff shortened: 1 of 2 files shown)


jon-a-nygaard:
@gtanczyk Thanks for reporting.

I found that the interval for the range used in the binary search in `getDistanceBetweenCirclesByOverlap` is not precise enough, and that it is difficult to predict when it is. So at the moment  I am working on switching from using binary search to using the bisection method which avoids the issue.

Will come back to you when there is an update on the progress.

jon-a-nygaard:
@gtanczyk A PR #9844 which solves your reported issue is currently in review. It is expected to make it into the next maintenance release.

Thanks again for reporting the issue and should there be anything else please let us know.

PR #9844, opened 2019-01-09
Replace binary search with bisect in getDistanceBetweenCirclesByOverlap

### LINKED ISSUE #9769: Sets on venn diagram are visually overlapping while there is no overlap in data
#### Expected behaviour
Circles should not overlap

#### Actual behaviour
image
(error was marked in red color using Gimp)

https://jsfiddle.net/q8ucmd0a/

#### Product version
Highcharts 7.0.1

#### Affected browser(s)
Latest google chrome

jon-a-nygaard:
# Description
The interval for the range used in the binary search in `getDistanceBetweenCirclesByOverlap` is sometimes not precise enough, and it is difficult to predict when it is. To avoid the issue the binary search is replaced with bisection method which does not use use static intervals.

This edit also takes #9769 into account and makes sure that when overlap equals 0 then the distance between two circles are greater or equal to the sum of their radiuses. 

# Example(s)
- [Demo of issue in #9769](https://jsfiddle.net/q8ucmd0a/)
- [Demo of issue in #9769 with fix applied](https://jsfiddle.net/jon_a_nygaard/8jv2uqog/)
- [Demo of issue in #9811](https://jsfiddle.net/10yvLwuq/)
- [Demo of issue in #9811 with fix applied](https://jsfiddle.net/jon_a_nygaard/ha421jxg/1/)

# Related issue(s)
- Addition to #9811 
- Closes #9769

### DIFF
--- a/js/modules/venn.src.js
+++ b/js/modules/venn.src.js
@@ -35,7 +35,6 @@ var color = H.Color,
     isPointOutsideAllCircles = geometryCircles.isPointOutsideAllCircles,
     isString = H.isString,
     merge = H.merge,
-    round = geometryCircles.round,
     seriesType = H.seriesType;
 
 var objectValues = function objectValues(obj) {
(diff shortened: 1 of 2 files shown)


gtanczyk:
@jon-a-nygaard  I can see different problem now in [Demo of issue in #9811 with fix applied](https://jsfiddle.net/jon_a_nygaard/ha421jxg/1/):
image

`A` shouldn't include entire `C`

jon-a-nygaard:
Thanks @gtanczyk 👍 I will look into the issue as soon as possible.

jon-a-nygaard:
Thanks again @gtanczyk for reporting the issue. There is now a pull request https://github.com/highcharts/highcharts/pull/9987 which solves this issue. It is currently in review, and if accepted then it will be included in the next maintenance release.

PR #9846, opened 2019-01-09
Updated highcharts-assembler to v1.1.0.

jon-a-nygaard:
# Description
Updated dev dependency `highcharts-assembler` to `v1.1.0`. 
The update includes a change which introduces the property `default` on the factory function that is returned to `module.exports`. This allows for TypeScript users to import the Highcharts from default as following `import Highcharts from 'highcharts';` instead of using import whole module as following `import * as Highcharts from 'highcharts';`.
Both of these mentioned cases will now supported in TypeScript.

# Related issue(s)
- https://github.com/highcharts/highcharts-angular/issues/101

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,7 @@
     "gulp-util": "^3.0.8",
     "gzip-size": "^3.0.0",
     "highcharts-api-docs": "github:highcharts/api-docs",
-    "highcharts-assembler": "github:highcharts/highcharts-assembler#v1.0.28",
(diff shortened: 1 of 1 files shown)


bre1470:
> This allows for TypeScript users to import the Highcharts from default as following import Highcharts from 'highcharts'; instead of using import whole module as following import * as Highcharts from 'highcharts';.
I like to point out, that this is only working, if `"allowSyntheticDefaultImports": true` is set in `tsconfig.json`.

PR #9848, opened 2019-01-09
feature/9798-networkgraph-particular-link-options-enhancement

raf18seb:
Added ability to change options for a particular link and improved nodes docs for networkgraph series.

### DIFF
--- a/js/modules/networkgraph/networkgraph.src.js
+++ b/js/modules/networkgraph/networkgraph.src.js
@@ -437,12 +437,13 @@ seriesType('networkgraph', 'line', {
     },
     // Links:
     getLinkAttribues: function () {
-        var linkOptions = this.series.options.link;
+        var linkOptions = this.series.options.link,
+            pointOptions = this.options;
 
         return {
-            'stroke-width': linkOptions.width,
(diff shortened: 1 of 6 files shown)


js/modules/networkgraph/networkgraph.src.js
```diff
@@ -437,12 +437,13 @@ seriesType('networkgraph', 'line', {
...
+        var linkOptions = this.series.options.link,
+            pointOptions = this.options;
 
         return {
-            'stroke-width': linkOptions.width,
-            stroke: linkOptions.color,
-            dashstyle: linkOptions.dashStyle
+            'stroke-width': pick(pointOptions.width, linkOptions.width),
+            stroke: pointOptions.color || linkOptions.color,
+            dashstyle: pointOptions.dashStyle || linkOptions.dashStyle
```
sebastianbochan: What is the reason of using the `pick()` for `stroke-width` but not for others?

raf18seb: Link's width can be set to 0, but color and dashStyle can not.

PR #9856, opened 2019-01-10
bugfix/9854-wrong-compare-value

raf18seb:


raf18seb:
You are right, it can be simplified.

Also, is condition ` && keyIndex === -1` required?
It is always true because keyIndex is defined as `var keyIndex = -1;`

PR #9919, opened 2019-01-16
tsc

bre1470:
Initial TypeScript migration configuration


bre1470:
To actually build something from the `./ts` folder into the existing building system, you have to manually do `npx gulp tsc && npx gulp scripts`.

bre1470:
Done.

PR #9922, opened 2019-01-17
Feature/a11y improvements 2

oysteinmoseng:
Improvements to a11y module as response to user feedback.

Made reset zoom button keyboard navigable, added better legend accessibility, improved chart description handling, and added the series name to the point labels.

### DIFF
--- a/js/modules/a11y-i18n.src.js
+++ b/js/modules/a11y-i18n.src.js
@@ -297,24 +297,26 @@ H.Chart.prototype.langFormat = function (langKey, context, time) {
 H.setOptions({
     lang: {
 
+        /* eslint-disable max-len */
         /**
-     * Configure the accessibility strings in the chart. Requires the
-     * [accessibility module](//code.highcharts.com/modules/accessibility.js)
-     * to be loaded. For a description of the module and information on its
-     * features, see [Highcharts Accessibility](
-     * http://www.highcharts.com/docs/chart-concepts/accessibility).
-     *
-     * For more dynamic control over the accessibility functionality, see
-     * [accessibility.pointDescriptionFormatter](
-     * accessibility.pointDescriptionFormatter),
-     * [accessibility.seriesDescriptionFormatter](
-     * accessibility.seriesDescriptionFormatter), and
-     * [accessibility.screenReaderSectionFormatter](
-     * accessibility.screenReaderSectionFormatter).
-     *
-     * @since        6.0.6
-     * @optionparent lang.accessibility
-     */
+         * Configure the accessibility strings in the chart. Requires the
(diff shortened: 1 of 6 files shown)


js/modules/screen-reader.src.js
```diff
@@ -772,6 +773,60 @@ H.Chart.prototype.addScreenReaderRegion = function (id, tableId) {
...
+                    itemGroup.attr('aria-label', label);
+                }
+                text.attr('aria-hidden', 'false');
+            }
+        });
+    }
+});
+
+
+// Handle show/hide series
```
TorsteinHonsi: Legend items can also belong to Points, like the case is with the pie series. We have a legend event, `afterColorizeItem`, maybe it's more appropriate to use that? One of its parameters is `visible`.

js/modules/keyboard-navigation.src.js
```diff
@@ -1117,6 +1102,31 @@ H.Chart.prototype.addKeyboardNavigationModules = function () {
             }
         }),
 
+        // Reset zoom
+        navModuleFactory('resetZoom', [
+            // Tab/Up/Down/Left/Right - just move
+            [[9, 37, 38, 39, 40], function (keyCode, e) {
```
jon-a-nygaard: Will suggest using a map for improving readability.
E.g:
```javascript
var keyCodes: {
  tab: 9,
  up: 37.
  // ...
};
Object.keys(keyCodes).forEach(keyCode, e) {
  // ...
    keyCode === keyCodes.tab && e.shiftKey ||
  // ...
});
// ....
```

oysteinmoseng: Agree, but that is a bigger refactoring and should be put off to a later commit when we do the options refactoring etc for 7.1. Note that the above is not a forEach, but an array of numbers that get the same handler function in the navModule, so it is not a straight refactor.

js/modules/a11y-i18n.src.js
```diff
@@ -297,24 +297,26 @@ H.Chart.prototype.langFormat = function (langKey, context, time) {
 H.setOptions({
     lang: {
 
+        /* eslint-disable max-len */
```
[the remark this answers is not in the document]
oysteinmoseng: We could do line breaks with \ or + for the actual code (not sure if JSDoc handles line breaks in links), but that is a mess and easily breaks if you add whitespace, plus adds overhead to the code, so I suggest we keep the exception since the `max-len` rule is not making the code more readable here. But that is up for discussion. To me it seemed easier to do the exception for the whole block in case we want to make changes to the texts later, but we still do it block by block and not for the whole option set.

js/modules/screen-reader.src.js
oysteinmoseng: Added

PR #9951, opened 2019-01-21
docs/dateformats-type

### LINKED ISSUE #9920: Highcharts.dateFormats is readonly
#### Expected behaviour
I expect to be able to set custom date formatters through Highcharts.dateFormats like so:
`Highcharts.dateFormats = {
    Q: function(timestamp) {
        return timestamp + '';
    },
};`

#### Actual behaviour
Highcharts.dateFormats is readonly in the typescript module, and so an error is raised by my IDE when I try to assign this value.

I don't see the typescript module on the CDN, so not sure I can reproduce in a fiddle.

#### Product version
Highcharts 7.0.1

#### Affected browser(s)
I'm working with Jetbrains Webstorm

bre1470:
Fixed #9920 (Possible types of Highcharts.dateFormats)

### DIFF
--- a/js/parts/Time.js
+++ b/js/parts/Time.js
@@ -596,7 +596,7 @@ Highcharts.Time.prototype = {
                  *         Adding support for week number
                  *
                  * @name Highcharts.dateFormats
-                 * @type {Highcharts.Dictionary<Highcharts.TimeFormatCallbackFunction>}
+                 * @type {Highcharts.Dictionary<Highcharts.TimeFormatCallbackFunction>|undefined}
                  */
                 H.dateFormats
             );


Timppe:
This will not remove the type error "Cannot assign to 'dateFormats' because it is a constant or a read-only property" when assigning custom formatters like in the official jsfiddle: http://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/global/dateformats/ as we do in our codebase like this:
```
Highcharts.dateFormats = {
  W: weekFormatter,
  Y: yearFormatter,
  Q: quarterFormatter,
  M: monthFormatter
};
```

PR #9953, opened 2019-01-21
Fixed draggable points capturing click events

jon-a-nygaard:
# Description
The drag guide box would in most cases capture the click event and prevent the click event from being fired on the point. This is solved by setting the style property `pointer-events` to `none` on the drag guide box element.


jon-a-nygaard:
@oysteinmoseng Agreed, please have a look at recent changes.

PR #9956, opened 2019-01-21
bugfix/7872-column-animate

sebastianbochan:


sebastianbochan:
@TorsteinHonsi 
The bug is related with run animation in the hidden tab, but Im not sure how we can prepare a unit-test for it.

I verified something like below, without success.
- https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/move but  

The current Chrome does not support some references from security reasons.

TorsteinHonsi:
I'm closing this, fixed it on a lower level.

Regarding tests we can probably set `document.hidden` temporarily in the test runner, otherwise we can open up the `animate` function to override the check.  I'll leave it without a test for now, but if more cases like this come along we should set up something.

PR #9974, opened 2019-01-23
bugfix/9964-legend-title-margin-top

sebastianbochan:


sebastianbochan:
Well its strange cause I merge chartOptions with titleOptions (used in setTitle). Will check, fix and add test to it too.

TorsteinHonsi:
An alternative approach could be to check `chart.titleOffset` in these two locations:

https://github.com/highcharts/highcharts/blob/eb55a7478e5d61b6e80a25de4b30fad224bcf1f3/js/parts/Legend.js#L759

https://github.com/highcharts/highcharts/blob/eb55a7478e5d61b6e80a25de4b30fad224bcf1f3/js/parts/Legend.js#L978

So if `chart.titleOffset` is falsy, the `title.margin` is not added. It may be a cleaner approach, as we don't have to modify options.

sebastianbochan:
When I was testing it, the "higher" spacing has already been added. Manipulation in these lines, only reposition legend, but there was still too much space for the legend.

PR #9976, opened 2019-01-23
Enhancement/9812 treemap events

### LINKED ISSUE #9812: Treemap chart drillup button click event
I want to do a modification to a treemap chart after click on the drill up button (back button). I am using the allowDrillToNode: true property, but there should be a way to catch the back button's click event.

jon-a-nygaard:
# Description
## setRootNode
Adds the option `plotOptions.treemap.events.setRootNode`, this event is called whenever the root node changes and contains information about the what triggered the event, e.g. `click` or `traverseUpButton`.

In addition the option `plotOptions.treemap.allowDrillToNode` has been renamed to `allowTraversingTree`, and `plotOptions.treemap.drillUpButton` has been renamed to `traverseUpButton`.
This is to avoid confusion with the drilldown module.

The old option names has been deprecated and will be removed in the future. Any usage of these options will warn about the deprecation.

I suggest using `traverse` as a term for the options here as it is more flexible IMHO than `step`, which I believe would indicate that you only travel one level up or down, while traverse does not indicate as strongly either direction or number of levels.

Adding Sebastian and Kacper as reviewers since you guys were involved in the discussion in #9812 

## Series.setOptions
Additionally the `Series.setOptions` now fires a new event `setOptions`. This comes in handy when checking if the user has set a deprecated option, to then warn about it and copy it to the correct option. See the following example: https://github.com/highcharts/highcharts/blob/e9e1328cadb948a7eaa0e31533de094525ebe26d/js/modules/treemap.src.js#L658-L692

# Related issue(s)
- Fixed #9812


js/modules/treemap.src.js
```diff
@@ -246,20 +279,67 @@ seriesType(
...
+            position: {
+
+                /**
+                 * Vertical alignment of the button.
+                 *
+                 * @default    top
+                 * @validvalue ["top", "middle", "bottom"]
+                 * @product    highcharts
+                 * @apioption  plotOptions.treemap.traverseUpButton.position.verticalAlign
+                 */
```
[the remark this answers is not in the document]
jon-a-nygaard: It is actually not quite a duplicate, it is the old option which has been renamed and is now deprecated. If I remove it then it will not appear in the api, which I would like so that I can inform of the deprecation.
The best thing would be to make a alias/copy in the options and just set one of them to deprecated, but I do not believe this is possible with current system.

jon-a-nygaard:
Thanks @KacperMadej 👍 I will fix the Sunburst options and add it to the PR

js/modules/treemap.src.js
```diff
@@ -609,27 +657,46 @@ seriesType(
...
+                var options = event.userOptions;
 
-            options = series.options;
+                // Test if deprecated option is set.
+                if (defined(options.allowDrillToNode)) {
+                    error(
+                        'WARNING: plotOptions.treemap.allowDrillToNode has ' +
+                        'been renamed to plotOptions.treemap.' +
+                        'allowTraversingTree, and will be removed in the ' +
+                        'next major version.'
```
sebastianbochan: 1. Export text of error to number of error (like error 19/25).
2. Add chart context (third param of error).

Example:
```
H.error(18, true, chart);
```

It will allow us to keep full compatibility with `debug module`.

js/modules/treemap.src.js
```diff
@@ -609,27 +657,46 @@ seriesType(
...
+                    }
+                    delete options.allowDrillToNode;
+                }
 
-                // Copy option if not already defined with correct name.
-                if (!defined(options.allowTraversingTree)) {
-                    options.allowTraversingTree =
-                        options.allowDrillToNode;
+                if (defined(options.drillUpButton)) {
+                    error(
```
sebastianbochan: 1. Export text of error to number of error (like error 19/25).
2. Add chart context (third param of error).

Example:
```
H.error(18, true, chart);
```

It will allow us to keep full compatibility with `debug module`.

sebastianbochan: 1. Export text of error to number of error (like error 19/25).
2. Add chart context (third param of error).

Example:
```
H.error(18, true, chart);
```

It will allow us to keep full compatibility with `debug module`.

js/modules/treemap.src.js
```diff
@@ -609,27 +657,46 @@ seriesType(
...
             }
 
-            Series.prototype.init.call(series, chart, options);
+            addEvent(series, 'setOptions', function (event) {
+                var options = event.userOptions;
 
-            options = series.options;
+                // Test if deprecated option is set.
+                if (defined(options.allowDrillToNode)) {
+                    error(
```
sebastianbochan: 1. Export text of error to number of error (like error 19/25).
2. Add chart context (third param of error).

Example:
```
H.error(18, true, chart);
```

It will allow us to keep full compatibility with `debug module`.

js/parts/Series.js
```diff
@@ -3009,13 +3009,16 @@ H.Series = H.seriesType(
                 userOptions = chart.userOptions || {},
                 userPlotOptions = userOptions.plotOptions || {},
                 typeOptions = plotOptions[this.type],
+                seriesUserOptions = merge(itemOptions),
                 options,
                 zones,
                 zone,
                 styledMode = chart.styledMode;
 
+            fireEvent(this, 'setOptions', { userOptions: seriesUserOptions });
```
sebastianbochan: What is the reason of calling setOptions?

js/modules/treemap.src.js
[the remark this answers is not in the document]
KacperMadej: Isn't the `apioption` the same (311 vs 354), so both will create the same API doc?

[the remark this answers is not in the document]
jon-a-nygaard: You are correct, these options were the same ones, but they should not be. I corrected this in dfab9a4 and deprecated the one in `drillUpButton`.

js/parts/Series.js
jon-a-nygaard: The `setOptions` event was added to do what is described in https://github.com/highcharts/highcharts/pull/9976#issue-247001601 at the section Series.setOptions. The event is used at [modules/treemap.src.js#L658](https://github.com/highcharts/highcharts/blob/e9e1328cadb948a7eaa0e31533de094525ebe26d/js/modules/treemap.src.js#L658)

js/modules/treemap.src.js
jon-a-nygaard: I had a chat with Torstein about this, and we do not have a practice of warning of deprecations in the console. So ended up removing the warning.

If we should use the error system, then it would not only have to work with the debug module, but also it has to be flexible enough to specify which deprecated option is used, which is not possible when using error numbers.
Something to keep in mind for the future.

js/modules/treemap.src.js
jon-a-nygaard: See https://github.com/highcharts/highcharts/pull/9976#discussion_r250987773

jon-a-nygaard: https://github.com/highcharts/highcharts/pull/9976#discussion_r250987773

js/modules/treemap.src.js
jon-a-nygaard: See https://github.com/highcharts/highcharts/pull/9976#discussion_r250987773

jon-a-nygaard:
@sebastianbochan 
1. Agreed, alert is not ideal, but console.log is not as easily visible for a user, so I changed it to update a `p` tag in the demo. See changes at https://github.com/highcharts/highcharts/pull/9976/commits/4623b834af99d2b0db3e254fbfe1ebe79f690d95
2. Unit tests can be found at [samples/unit-tests/series-treemap/members/demo.js#L152-L252](https://github.com/highcharts/highcharts/blob/dfab9a461d3e0c639ad444b13547bae72eef5112/samples/unit-tests/series-treemap/members/demo.js#L152-L252)

PR #10008, opened 2019-01-28
bugfix/9988-stocktools-popup-scroll

sebastianbochan:


sebastianbochan:
Alternatively we can enable scrollbar on the whole popup by CSS:

```
.highcharts-popup {
  overflow-x: hidden;
  overflow-y: scroll;
}
```

Is it better solution?

TorsteinHonsi:
Yes if that gives the desired result it would be better.

TorsteinHonsi:
What's the status here @sebastianbochan ?

sebastianbochan:
Will do commit today.

PR #10034, opened 2019-01-31
bugfix/9209-unexpected-zoom-extremes-after-update

wchmiel:
`rangeSelector.selected` and `rangeSelector.options.selected` were not cleared when all buttons were unchecked. After chart update, this option was used as a pre-selected button index what should be performed only on chart initialization.

### DIFF
--- a/js/parts/RangeSelector.js
+++ b/js/parts/RangeSelector.js
@@ -962,8 +962,7 @@ RangeSelector.prototype = {
         // should be cleared (#9209)
         if (
             btnSelected === undefined &&
-            selected !== null &&
-            selected !== undefined
+            defined(selected)
         ) {
             rangeSelector.setSelected(null);
         } else if (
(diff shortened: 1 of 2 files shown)


js/parts/RangeSelector.js
```diff
@@ -948,13 +949,29 @@ RangeSelector.prototype = {
...
                 button.setState(state);
             }
         });
+
+        // When all buttons are unchecked selected property
+        // should be cleared (#9209)
+        if (
+            btnSelected === undefined &&
+            selected !== null &&
+            selected !== undefined
```
sebastianbochan: `H.defined` verifies null/undefined.

js/parts/RangeSelector.js
```diff
@@ -948,13 +949,29 @@ RangeSelector.prototype = {
...
+        ) {
+            rangeSelector.setSelected(null);
+        } else if (
+            btnSelected !== undefined &&
+            btnSelected !== selected
+        ) {
+            rangeSelector.setSelected(btnSelected);
+        }
```
sebastianbochan: Consider to replace the whole condition with:
```
 if (btnSelected === undefined) {
           if(H.defined(selected) {
                 rangeSelector.setSelected(null);
           } else if (btnSelected !== selected) {
                 rangeSelector.setSelected(btnSelected);
           }
}
```

Why? Because in two IFs you check the same part `btnSelected !== undefined`.

PR #10045, opened 2019-02-04
Fixed #10043, corrected setRootNode doclets

### LINKED ISSUE #10043: Docs: Treemap setRootNode function must be removed from global scope and needs param interface
https://github.com/highcharts/highcharts/blob/c9f93f5acb3b5b16a18ce7098b9e51f6e61e242f/js/modules/treemap.src.js#L1525-L1543

- [ ] `setRootNode` needs clarification were it should be attached. Series probably.
- [ ] Parameter needs interface.
- [ ] Internal `defaultFn` needs a name if it should be public otherwise, the `@private` tag is required.
- [ ] If `defaultFn` is public, an interface for param is needed

jon-a-nygaard:
# Description
Fixed incorrect doclets for `setRootNode` in treemap.

# Related issue(s)
- Closes #10043

### DIFF
--- a/js/modules/treemap.src.js
+++ b/js/modules/treemap.src.js
@@ -1525,20 +1525,13 @@ seriesType(
         /**
          * Sets a new root node for the series.
          *
+         * @private
+         * @function Highcharts.Series#setRootNode
+         *
          * @param {string} id The id of the new root node.
          * @param {boolean} [redraw=true] Wether to redraw the chart or not.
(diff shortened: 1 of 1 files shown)


bre1470:
If it is set to private you can use nested params, if you like, so it is not really necessary to remove them.

jon-a-nygaard:
Aha, will add them back again to improve the clarity of the params 👍

PR #10048, opened 2019-02-04
feature/networkgraph-hover-opacity

pawelfus:
Networkgraph: Added transparency to nodes that are not hovered. 

By highlighting only hovered nodes we greatly improve readability of the graph:
screen shot 2019-02-04 at 12 32 43
screen shot 2019-02-04 at 12 32 34

Demo: http://jsfiddle.net/BlackLabel/k4bs02gh/690/show

Issue: Where should we place an option (options?) to control this from API? Sounds like a `series/marker.states.hover`, but actually we change opacity of inactive nodes (and links). I'm open for suggestions here.

### DIFF
--- a/js/modules/networkgraph/networkgraph.src.js
+++ b/js/modules/networkgraph/networkgraph.src.js
@@ -497,25 +497,60 @@ seriesType('networkgraph', 'line', {
         return Series.prototype.destroy.apply(this, arguments);
     }
 }, {
-    init: function () {
-        Point.prototype.init.apply(this, arguments);
+    onMouseOver: function () {
+        if (this.series.draggable) {
+            H.css(this.series.chart.container, { cursor: 'move' });
+        }
 
-        addEvent(
-            this,
-            'mouseOver',
-            function () {
-                H.css(this.series.chart.container, { cursor: 'move' });
-            }
-        );
-        addEvent(
-            this,
-            'mouseOut',
-            function () {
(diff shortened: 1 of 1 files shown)


TorsteinHonsi:
> Where should we place an option (options?) to control this from API? 

We have the same situation in the [item chart POC](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/studies/item-rows/), where I also haven't come up with a name. Plus this implementation is so simple and useful that we should apply it to all series.

What about `states.hover.otherOpacity`? It will signal that it belongs to the state, and that the opacity applies to the other items. I'm also open to other suggestions to find the most intuitive name.

pawelfus:
How about a separate state, e.g. `inactive`? We have limited usage of `otherOpacity`. For example what if users would prefer to change dashstyle and color instead of opacity? We would need to create more and more options under `hover` state. Alternatively `hover.others = { .. }` could be acceptable? But we create another nested level.. I think a separate state (which I think this is what it is) is a nice solution.

I agree, that would be useful for all series, something similar in styled-mode is quite easy if I remember correctly. And the name was `inactive`, I think?

TorsteinHonsi:
What's the status on this PR @pawelfus ? We decided to add an `inactive` state for general series types.

pawelfus:
Separate PR and branch created for `inactive` state: #10298 - thus closing this one.

PR #10053, opened 2019-02-04
bugfix/7773-compare-settings

raf18seb:
Added new property _plotOptions.sma.compareToMain_ (boolean) which allows the user to choose whether to compare indicator to the main series values or indicator's values (default).

### DIFF
--- a/js/indicators/indicators.src.js
+++ b/js/indicators/indicators.src.js
@@ -83,10 +83,10 @@ seriesType(
      *
      * @extends      plotOptions.line
      * @since        6.0.0
-     * @excluding    allAreas, colorAxis, compare, compareBase, joinBy, keys,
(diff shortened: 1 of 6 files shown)


js/indicators/indicators.src.js
```diff
@@ -112,6 +112,16 @@ seriesType(
          * @type {string}
          */
         linkedTo: undefined,
+        /**
+         * Whether to compare indicator to the main series values
+         * or indicator values.
+         *
+         * @sample {highstock} stock/plotoptions/series-comparetomain/
+         *         compareToMain
```
pawelfus: Could you change the name of the demo to be more user friendly? @raf18seb Also, it would be a good idea to provide a demo that shows two different settings at the same time to highlight the difference.

raf18seb: Sure, I'll do it soon.

PR #10072, opened 2019-02-06
feature/networkgraph-link-labels

pawelfus:
Refactored networkgraph to support labels for links. ~~PR does not support text following a path~~. It does now.

Demo: http://jsfiddle.net/BlackLabel/0n731wb8/29/

Screenshot:
screen shot 2019-02-06 at 16 12 32

Problem with text following path is that `drawDataLabels` supports only `text()` and `label()` and we don't have `textPath()`. Recreating whole logic, similar to `text()` sounds troubling. @TorsteinHonsi - any suggestions here? 

Maybe something like this:
Add an option to let series choose constructor for creating dataLabel here:
https://github.com/highcharts/highcharts/blob/2c52e0a6d172f67f2b9068b176de73e17b96cbf1/js/parts/DataLabels.js#L439-L446

Then wrap `text()` (similar to `label()`?) to create `textPath()` in renderer?

Alternatively:
Extend `text()` with new argument `textPath`/`usePath` (which will be actual path, or SVG element) which will create one more SVG level: 
`<g> <text> <textPath> <tspan> </tspan> <tspan> </tspan> </textPath> </text> </g>`
Instead of the current one:
`<g> <text> <tspan> </tspan> <tspan> </tspan> </text> </g>`


pawelfus:
Note: This feature could be used not only in networkgraph, but also in pie chart, sankey chart, series.label plugin, xAxis labels (categories) on radial axes (gauge/polar) or `mapline` in Highmaps.

TorsteinHonsi:
I think the alternative approach of allowing a pseudo-attribute, `textPath`, sounds like a good solution. It should be a quite lightweight implementation (starting with `SVGElement.prototype.textPathSetter`?), and would allow us to start using it wherever we have texts today by simply adding an attribute. 

And yes, I've been missing this for both series labels and sankey links.

pawelfus:
Thanks!

### Current status (PoC):

- http://jsfiddle.net/mq9j0krc/1/ (just links)
- http://jsfiddle.net/1fvp795a/5/ (just dataLabels)
- http://jsfiddle.net/1fvp795a/6/ (bubble series, actually it works for all series)

### API proposal:
Simple `attr({ textPath: element })` is a bit problematic, as a lot of options can change how textPath is rendered. Instead I used: `attr({ element: SVGElement, options: { ... }})` but it seems too complicated. Maybe `attr({ textPath: element })` should stay, and we would need to manually change all other attributes? That would require more changes in `dataLabels.js` file to support additional attrs for `textPath` only.

```javascript
{
  dataLabels: {
    // all series
    textPath: { // maybe `useTextPath` is a better name?
      enabled: boolean,
      attributes: {
         startOffset:  string | number,
         // Probably text-anchor should be replaced with current `alignSetter` logic
         'text-anchor': 'middle' | 'start' | 'end', 
         dy: number,
         dx: number
      }
    },
    linkTextPath: { // networkgraph only
      enabled: boolean,
      attributes: { ... }
    }
  }
}
```

### Known issues:
- [`dataLabels.style.textOutline`](https://api.highcharts.com/highcharts/series.line.dataLabels.style) breaks this. It looks strange, as DOM structure is exactly the same, must be related to `dy`/`dx` or `x`/`y` and how these are calculated when text is following a path, demo: http://jsfiddle.net/mq9j0krc/3/ **Note that `textOutline` greatly decreases performance in networkgraph**
- it may be smart to change how circle marker is defined, compare http://jsfiddle.net/1fvp795a/6/ and http://jsfiddle.net/1fvp795a/7/
- allowOverlap has strange behaviour ( http://jsfiddle.net/1fvp795a/8/ )
- `textPath` shouldn't use `translateX` and `translateY`, I'm not sure what is the best way to deal with it. For now, I simply removed function that translates elements: https://github.com/highcharts/highcharts/pull/10072/commits/351ca36444aac9bd69e2f4c9615429bd3111a0f3#diff-65bc0ad38de3454c8d9cf0313d7bb35eR2369 (`updateTransform = H.noop`)

### Screenshots:
screen shot 2019-02-11 at 18 46 31
screen shot 2019-02-11 at 18 30 55
screen shot 2019-02-11 at 18 30 50

TorsteinHonsi:
### API proposal
Since the `textPath` is not just a simple string setting, I think we should leave the idea of performing it as a custom attribute. So I can imagine a couple of solutions:

* Add a `textPath()` function to the `SVGElement` that can be called after contructing a `text` item. That would be equivalent to functions like `SVGElement.shadow` or `SVGElement.clip`.
* As suggeted earlier, add a new element type next to `text` and `label`. But this may be more involving, plus it makes it harder (or at least counterintuitive) to switch the text path on and off for a text element.

### Other comments
* I support the proposal for using text anchor directly, though the option should be camel case `textAnchor`.
* As for `textOutline`, it should probably just be disabled when doing `textPath`.
* >  it may be smart to change how circle marker is defined
  Yes, we can adapt it so it starts in the right place, right?

pawelfus:
Thanks! I like the idea with `shadow()` similarity. Sounds like we can keep separated SVG logic from DataLabels -  let me try that.

TorsteinHonsi:
> Sounds like we can keep separated SVG logic from DataLabels
Yes that's very important

pawelfus:
Done, the only important thing about `setTextPath()` is order: should be called after creating `text()` element. Otherwise we won't be able to disable `textOutline`, for example:

```javascript
var text = renderer.text(...);

text
  .setTextPath(textPathOptions, path)
  .attr(attribs)
  .css(styles)
  .shadow(shadowOptions)
  .add(parent);
```

For now, default attributes are hardcoded, so we simple can set `dataLabels.textPath.enabled = true`, instead of defining all options:
https://github.com/highcharts/highcharts/blob/feature/networkgraph-link-labels/js/parts/SvgRenderer.js#L2324-L2328

pawelfus:
Added samples. I couldn't find a right place for http://utils.highcharts.local/samples/#view/highcharts/studies/renderer-textpath - I don't see any directory for `renderer` or something similar.

TorsteinHonsi:
> I don't see any directory for `renderer` or something similar.

Similar samples are currently placed under `highcharts/members` (for example [text](http://utils.highcharts.local/samples/#view/highcharts/members/renderer-rect)), so you can use that placement for now.

TorsteinHonsi:
@pawelfus All changes approved 👍 . Can you resolve the conflicts before I do the merge?

pawelfus:
Please check 60b19fe before merging

Issue #9836, opened 2019-01-08
TypeScript: Change naming of literal type definitions


bre1470:
Instead of `Type` suffix literal type definitions should have a `Value` suffix.

PR #10085, opened 2019-02-07
docs/9836-literal-type-names

bre1470:
Changed naming of literal definitions. #9836


bre1470:
This will break existing TypeScript code from our customers and should therefor be merged with v7.1.

PR #10089, opened 2019-02-07
bugfix/9962-boost-outside-points

### LINKED ISSUE #9962: Boost - line series is invisible if the points are outside the visible part of the plot area
#### Expected behaviour
The chart should look like without the boost module:
http://jsfiddle.net/BlackLabel/ryLqx7ku/

#### Actual behaviour
The line series is invisible which causes inconvenience when zooming.

http://jsfiddle.net/BlackLabel/L73x6f5e/

#### Product version
Highcharts v7.0.2

#### Affected browser(s)
All

pawelfus:
Fixed #9962, line series was not rendered in the boost mode, when all points were outside the extremes.

Will that be fine? In general, data is sorted, but I'm not sure if this is the case for boost mode too.

### DIFF
--- a/js/modules/boost.src.js
+++ b/js/modules/boost.src.js
@@ -1585,6 +1585,7 @@ function GLRenderer(postRenderCallback) {
             color,
             scolor,
             sdata = isStacked ? series.data : (xData || rawData),
+            sdataLength = sdata.length,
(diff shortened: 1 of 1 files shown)


pawelfus:
I think there is a better solution for that.

PR #10118, opened 2019-02-12
bugfix/10061-multiple-loaded-annotations-module

### LINKED ISSUE #10061: Annotations module is rendered more than once.
After resizing the window when the annotations module is added more than once chart renders more than one annotation. This problem was noticed and reported by Angular user. 

[DEMO](https://jsfiddle.net/BlackLabel/k3140d2u/)

#### Product version

Highcharts 7.0.2

#### Affected browser(s)

All

ppotaczek:
Fixed #10061, Annotations were duplicated if the module was loaded more than once.

### DIFF
--- a/js/annotations/annotations.src.js
+++ b/js/annotations/annotations.src.js
@@ -47,1185 +47,1193 @@ var merge = H.merge,
 /**
  * @typedef {Annotation.ControllableLabel} Annotation.Label
  */
-
-/**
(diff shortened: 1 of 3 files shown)


TorsteinHonsi:
This PR is now redundant, as our updated module bundler avoids loading the same module twice. Tested the original demo locally.

PR #10123, opened 2019-02-13
bugfix/9233-heatmap-null-point-label-and-tooltip

kamilkulig:


TorsteinHonsi:
Thanks for the PR! Unfortunately (or luckily), we already have logic in place that picks up `nullFormat` and `nullFormatter`, so we can do this much simpler by only changing a few lines of code.

1. In `Point.applyOptions`, set `point.formatPrefix` to `null` conditionally.
2. The map series has a property called `dataLabelOnNull`. We want to move that to the general `colorPointMixin`.

After applying that, [this fiddle](https://jsfiddle.net/highcharts/hxrbtd8f/) should provide both a tooltip and a data label for the null point.

Can you do that patch and prepare a PR including the API docs, tests and samples you already created? Though the new options are called `nullFormat` rather than `nullPointFormat`.

Here's the patch to make it work:
```diff
diff --git a/js/parts-map/ColorSeriesMixin.js b/js/parts-map/ColorSeriesMixin.js
index 1050af4b2..1a3bdc75d 100644
--- a/js/parts-map/ColorSeriesMixin.js
+++ b/js/parts-map/ColorSeriesMixin.js
@@ -20,6 +20,9 @@ var defined = H.defined,
  * @mixin Highcharts.colorPointMixin
  */
 H.colorPointMixin = {
+
+    dataLabelOnNull: true,
+
     /**
      * Color points have a value option that determines whether or not it is
      * a null point
diff --git a/js/parts-map/MapSeries.js b/js/parts-map/MapSeries.js
index 810b5514d..60a7700ae 100644
--- a/js/parts-map/MapSeries.js
+++ b/js/parts-map/MapSeries.js
@@ -1036,8 +1036,6 @@ seriesType(
     // Point class
     }), extend({
 
-        dataLabelOnNull: true,
-
         // Extend the Point object to split paths
         applyOptions: function (options, x) {
 
diff --git a/js/parts/Point.js b/js/parts/Point.js
index 3cdf640d7..7149b313f 100644
--- a/js/parts/Point.js
+++ b/js/parts/Point.js
@@ -220,6 +220,9 @@ Highcharts.Point.prototype = {
             point.isValid && !point.isValid(),
             point.x === null || !isNumber(point.y, true)
         ); // #3571, check for NaN
+        if (point.isNull) {
+            point.formatPrefix = 'null';
+        }
 
         // The point is initially selected by options (#5777)
         if (point.selected) {

```

kamilkulig:
Sure, I'll take care of this.

TorsteinHonsi:
Applied in another PR

PR #10141, opened 2019-02-14
Added scripts get-filesizes and compare-filesizes

jon-a-nygaard:
# Description
Added two new utility scripts to gulp, `get-filesizes` and `compare-filesizes`.

**Note:** These scripts are similar to the `gulp filesize` script, and can achieve the same result.

## Arguments
`get-filesizes` has support for the following arguments:
- `--file`, specifyes which files get filesizes of. If not specified it will default to all distributed files.
- `--out`, specify where to store the result. Defaults `filesizes.json`

`compare-filesizes` has support for the following arguments:
- `--old`, specify path to the "oldest" filesizes to compare. Required.
- `--new`, specify path to the "newest" filesizes to compare. Required. 
- `--out`, specify where to store the resulting information. If not specifyed then the information will be outputted to the console.
 
## Example use case
```
npx gulp get-filesizes --out old.json
# Do wanted modifications to files before running the next commands
npx gulp get-filesizes --out new.json
npx gulp compare-filesizes --old old.json --new new.json
# Will output information about the filesizes for new vs old
```

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1312,6 +1312,14 @@ const uploadAPIDocs = () => {
 };
 gulp.task('upload-api', uploadAPIDocs);
 
+/**
+ * Usage: npx gulp get-filesizes --out old.json
+ *
+ * Options:
+ *   --file     Specify which files to get filesizes of. If not specified it
+ *              will default to all distributed files.
+ *   --out      Specify where to store the result. Defaults "./filesizes.json"
+ */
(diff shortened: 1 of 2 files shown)


TorsteinHonsi:
@jon-a-nygaard See above

jon-a-nygaard:
@TorsteinHonsi Please see latest changes.

PR #10143, opened 2019-02-14
bugfix/1289/multiple-panes-zoom

wchmiel:
Wrong extremes on second pane yAxis when zoomed the first one. The solution is to zoom only the pane where an area was selected.

### DIFF
--- a/js/parts/Interaction.js
+++ b/js/parts/Interaction.js
@@ -449,7 +449,8 @@ extend(Chart.prototype, /** @lends Chart.prototype */ {
             hasZoomed,
             pointer = chart.pointer,
             displayButton = false,
-            mouseDownY = pointer.mouseDownY,
+            mouseDownPos =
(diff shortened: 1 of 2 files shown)


js/parts/Interaction.js
```diff
@@ -461,10 +462,25 @@ extend(Chart.prototype, /** @lends Chart.prototype */ {
         } else { // else, zoom in on all axes
             event.xAxis.concat(event.yAxis).forEach(function (axisData) {
                 var axis = axisData.axis,
-                    isXAxis = axis.isXAxis;
+                    axisTop = axis.top,
+                    axisBottom = axisTop + axis.height,
+                    isXAxis = axis.isXAxis,
```
TorsteinHonsi: Looking for whether is it an X axis or Y axis may fail if the chart is inverted. We should probably check `axis.horiz` instead. Check with inverted chart: http://jsfiddle.net/highcharts/mLfucwgd/5/

PR #10153, opened 2019-02-15
docs/10149-series-legend-click

### LINKED ISSUE #10149: TypeScript typing for SeriesLegendItemClickCallbackFunction is missing some properties
according to the definition 
```
export type SeriesLegendItemClickCallbackFunction = (this: Series, event: SeriesLegendItemClickEventObject) => void;
```

the `this` value for `plotOptions.series.events.legendItemClick` is of Chart instance, however it seems like it is actually some sort of "enriched" instance. For example it has properties like `index`, `color`, `symbol` and a lot of other calculated values like `closestPointRange`.

If needed I can provide a complete list with values that are missing.

bre1470:
- Completed properties of series legend click event. Fixed #10149


bre1470:
The `legendItemClick` option is excluded for the pie series and gets also not fired. If there is something point related, I recommend an additional set of interfaces, so that series and point context are not mixed up. Otherwise there is no type benefit for users and we could simplify it down to a generic function callback for this event.

bre1470:
Please review ones more.

PR #10182, opened 2019-02-20
hc71/documentation-generators

bre1470:
*Note:* `hc71/value-types` has to be merged first into `hc71/master`
- Switched to centralized documentation generators
- We could move `tools/jsdoc` too.


bre1470:
After merge into the root master, we have to make the `highcharts/highcharts-documentation-generators` repository public for customers.

js/parts/Options.js
```diff
@@ -4,6 +4,14 @@
  * License: www.highcharts.com/license
  */
 
+/**
+ * @typedef {"plot"|"chart"} Highcharts.ButtonAlignValue
+ */
+
+/**
+ * @typedef {"plotBox"|"spacingBox"} Highcharts.ButtonAlign2Value
```
TorsteinHonsi: Shouldn't it be `ButtonAlignToValue`? And should we try to combine `ButtonAlignValue` with `ButtonAlign2Value`?

bre1470: I fixed it with the merge.

PR #10197, opened 2019-02-22
bugfix/10163-networkgraph-setdata

### LINKED ISSUE #10163: Network graph:  series.setData does not remove existing nodes.
I am trying update network graph data. But existing nodes are not getting removed.

#### Expected behavior
Existing nodes should be removed and replaced with new nodes as per data.

#### Actual behaviour
Existing nodes are not getting removed.

[https://jsfiddle.net/px964goa/1/](https://jsfiddle.net/px964goa/1/)

Click on any node. I've added `setData` function on the same.

pawelfus:
Fixed #10163, `series.setData()` did not remove old nodes for a networkgraph.

I know, this fix will be delayed until v7.1, but we will have conflicts otherwise.

### DIFF
--- a/js/modules/networkgraph/networkgraph.src.js
+++ b/js/modules/networkgraph/networkgraph.src.js
@@ -238,6 +238,8 @@ seriesType('networkgraph', 'line', {
      * links.
      */
     createNode: H.NodesMixin.createNode,
+    setData: H.NodesMixin.setData,
+    destroy: H.NodesMixin.destroy,
 
     /**
      * Extend init with base event, which should stop simulation during update.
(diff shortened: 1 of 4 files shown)


js/modules/networkgraph/networkgraph.src.js
```diff
@@ -754,14 +757,18 @@ addEvent(
                                 chart.container,
                                 'mousemove',
                                 function (e) {
-                                    return point.series.onMouseMove(point, e);
+                                    return point &&
+                                        point.series &&
+                                        point.series.onMouseMove(point, e);
```
AyanGhatak: Since point.series is being used multiple times, should we consider creating a variable instead and access it from there ?

pawelfus: In general - yes. However, `point` can be undefined in a first place, so this would throw an error.

js/modules/networkgraph/networkgraph.src.js
```diff
@@ -531,13 +531,26 @@ seriesType('networkgraph', 'line', {
...
+    setData: function () {
+        if (this.nodes) {
+            this.nodes.forEach(function (node) {
+                node.destroy();
+            });
+            this.nodes.length = 0;
+        }
+
+        Series.prototype.setData.apply(this, arguments);
+    },
```
TorsteinHonsi: We have this code in the sankey series too: https://github.com/highcharts/highcharts/blob/b5cdcdf80bdd7692afdf60ec3ecf3346ea73b649/js/modules/sankey.src.js#L443-L452

Shouldn't we move it to `mixins/node.js`?

pawelfus: Thanks! Moved actually more methods:
- `generatePoints` (part of the code was the same)
- `setData` 
- `destroy`

PR #10205, opened 2019-02-25
Fixed #10176, scrollbar error when collapsed treegrid

### LINKED ISSUE #9547: Scroll breaks when i collapse elements
#### Expected behaviour
I build a Gantt chart using the highcharts API.
 I added a scrollbar on the yAxis, it should works when i collapse to show or hide sub elements.
#### Actual behaviour
When i collapse elements to hide sub-Lines of a category the length of yAxis is not updated.

https://jsfiddle.net/lodou/6up1gezf/

#### Product version
6.2.0

#### Affected browser(s)
ALL

### LINKED ISSUE #10176: Script error (length is NaN) when scrollbar from HighStock is used w/ Gannt chart and collapsing
#### Expected behaviour
Gannt chart vertical scrolling (from HighStock) should work even when all series' are collapsed.

#### Actual behaviour
On supplied JSFiddle, if the scrollbar is scrolled after collapsing "New Offices" we see script errors

`Error: <rect> attribute y: Expected length, "NaN".`

https://jsfiddle.net/kr4c0wz3/

#### Product version
v7.0.3

#### Affected browser(s)
Chrome

jon-a-nygaard:
# Description
Setting extremes on the `yAxis` while it is broken will cause an error. This is due to the use of `closestPointRange` when modifying the `newMin` and `newMax`. `closestPointRange` is not set on `yAxis`.

Solved by looking at `from` and `to` on the axis breaks to modify the extremes instead.

This PR is not doing much to improve the usability of scrollbar with Gantt, but it solves the script error at least.

# Example(s)
Open a demo, click on "New Offices" to collapse it, then drag the scrollbar handle.
- [Demo of issue](https://jsfiddle.net/kr4c0wz3/)
- [Demo with fix applied](https://jsfiddle.net/jon_a_nygaard/zyw1r8Lj/)

# Related issue(s)
- Closes #9547 
- Closes #10176

### DIFF
--- a/js/modules/broken-axis.src.js
+++ b/js/modules/broken-axis.src.js
@@ -205,10 +205,10 @@ Axis.prototype.setBreaks = function (breaks, redraw) {
                     breaks = this.options.breaks;
 
                 while ((axisBreak = findBreakAt(newMin, breaks))) {
-                    newMin = axisBreak.from;
+                    newMin = axisBreak.to;
                 }
                 while ((axisBreak = findBreakAt(newMax, breaks))) {
-                    newMax = axisBreak.to;
+                    newMax = axisBreak.from;
                 }
 
                 // If both min and max is within the same break.
--- a/js/modules/broken-axis.src.js
+++ b/js/modules/broken-axis.src.js
@@ -15,6 +15,7 @@ var addEvent = H.addEvent,
     pick = H.pick,
     extend = H.extend,
     isArray = H.isArray,
+    find = H.find,
     fireEvent = H.fireEvent,
     Axis = H.Axis,
     Series = H.Series;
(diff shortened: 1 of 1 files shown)


js/modules/broken-axis.src.js
```diff
@@ -19,6 +19,21 @@ var addEvent = H.addEvent,
...
+ * Returns the first break found where the x is larger then break.from and
+ * smaller then break.to.
+ *
+ * @param {number} x The number which should be within a break.
+ * @param {array} breaks The array of breaks to search within.
+ * @return {object|boolean} Returns the first break found that matches, returns
+ * false if no break is found.
+ */
+var findBreakAt = function (x, breaks) {
+    return breaks.find(function (b) {
```
TorsteinHonsi: Unfortunately `Array.prototype.find` is off limits for now, IE11 doesn't support it. Use `H.find`.

jon-a-nygaard: Good catch.
I often forget which Array methods that are off limits, is it possible to add an eslint rule or plugin that checks browser compatibility?

jon-a-nygaard:
@TorsteinHonsi I have removed the use of `Array.prototype.find`

PR #10206, opened 2019-02-25
feature/10036-dataLabels-inside-pyramid

wchmiel:
Added `dataLabels.inside` property in funnel and pyramid series types.

Now labels can be aligned inside points:

- vertically (top, center, bottom)

- horizontally (left, center, right)

### DIFF
--- a/js/parts/DataLabels.js
+++ b/js/parts/DataLabels.js
@@ -1264,6 +1264,9 @@ if (seriesTypes.pie) {
             seriesCenter = series.center,
             radius = seriesCenter[2] / 2,
             centerY = seriesCenter[1],
+            reversed = series.options.reversed,
+            dataLabelsInsideFunnel = options.inside &&
+                (series.type === 'funnel' || series.type === 'pyramid'),
             dataLabel,
             dataLabelWidth,
             // labelPos,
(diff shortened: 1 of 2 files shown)


TorsteinHonsi:
Thanks for the PR!

I'm rejecting it because I think this can be done less intrusively. When writing logic for a relatively marginal plugin, it is important that the impact on the core files is as small as possible, so we don't want a large block of extra logic inside `highcharts.js` in order to support a feature in `modules/funnel.js`.

Additionally, we already have logic for putting data labels inside a square, so we should try to leverage that. 

I've created a new feature branch that supports alignment inside the boxes, see https://jsfiddle.net/1jqe804v/1/. In short, it does two things: 1) Adds a `point.dlBox` to each point and 2) borrow the column series' implementation of `alignDataLabel` to pick up the `dlBox` and align the label inside it.

The only potential downside I can see is that the dlBox is a square, so when aligning text horizontally it may not be able to align to the funnel shape. I've set the square to the width of the neck for now, but maybe we can improve that.

Can you apply your unit tests to that branch and see if they all pass? And do changes if it breaks.

wchmiel:
Ok, I understand. Thanks for the feedback;) Sure, I will test it and try to improve.

Issue #10189, opened 2019-02-21
highcharts.d.ts autoRotation of XAxisLabelsOptions needs an updated type to include false.


justinmwarner:
#### Expected behaviour
Per the docs:

> Set it to `false` to disable rotation, which will cause the labels to word-wrap if possible.

Meaning this property can be false.

#### Actual behaviour
Can't give it false.

#### Live demo with steps to reproduce
None, not needed.

#### Product version
Latest, 7.0.3

#### Affected browser(s)
Not a browser issue

#### Fix
Change the line to be the following:

`autoRotation?: Array<number> | false;`

This should probably be updated in all cases, not just XAxisLabelsOptions

justinmwarner:
If this is something I can make a PR for, please let me know how. I saw the definition file was in the release repo for this package, and am unsure how I would update it.

sebastianbochan:
FYI @bre1470

PR #10212, opened 2019-02-26
Add false as a possible type to autoRotation

justinmwarner:
For #10189


TorsteinHonsi:
Thanks, I already merged @bre1470 's PR with the same content.

PR #10223, opened 2019-02-27
bugfix/10214-stocktools-class-bindings

sebastianbochan:


pawelfus:
Issue was caused by decoupling bindings and stock-tools. This used to be just one class name to control both (bindings events and CSS), but now we can not use the same class since styling and events come from different modules.

PR #10233, opened 2019-02-28
hc71/timeline-series-enhancement

Denyllon:


js/modules/timeline.src.js
```diff
@@ -124,19 +124,19 @@ undocumentedSeriesType('timeline', 'line',
...
-                    (this.point.label || '');
+                        '">● </span><span style="" > ' +
+                        (this.key || '') + '</span><br/>' +
+                        (this.point.label || '');
                 } else {
                     format = '<span>● </span>' +
-                    '<span>' + (this.point.name || '') +
-                    '</span><br/>' + (this.point.label || '');
+                        '<span>' + (this.key || '') +
+                        '</span><br/>' + (this.point.label || '');
```
sebastianbochan: Duplicate of code.

**Suggestion:**

```
 if (!this.series.chart.styledMode) {
   format = '<span style="color:' + this.point.color +
                    '">● </span>';
} else {
    format = '<span>● </span>'
}

format += '<span>' + (this.key || '') +'</span><br/>' + (this.point.label || '');

```

js/modules/timeline.src.js
```diff
@@ -145,7 +145,7 @@ undocumentedSeriesType('timeline', 'line',
              * @default 100
              */
             distance: 100,
-            /* *
+            /**
              * Whether to position data labels alternately. For example, if
              * [distance](#plotOptions.timeline.dataLabels.distance) is set
              * equal to `100`, then the first data label 's distance will be
```
sebastianbochan: Im not sure if it will be clear description for end-user.

js/modules/timeline.src.js
```diff
@@ -181,6 +180,31 @@ undocumentedSeriesType('timeline', 'line',
...
+                series.points.forEach(function (point) {
+                    // Set the isInside parameter basing on the real point
+                    // visibility, in order to avoid showing hidden points
+                    // in drawPoints method.
+                    point.isInside = point.visible;
+
+                    // New way of calculating closestPointRangePx value, which
+                    // respects the real point visibility is needed.
+                    if (point.visible) {
+                        if (defined(lastPlotX)) {
```
sebastianbochan: Can we use?

```
if (point.visible && defined(lastPlotX)) {
   ...
}
```

js/modules/timeline.src.js
```diff
@@ -181,6 +180,31 @@ undocumentedSeriesType('timeline', 'line',
...
+                        }
+                        lastPlotX = point.plotX;
+                    }
+                });
+                series.closestPointRangePx = closestPointRangePx;
+            });
+
             // Distribute data labels before rendering them. Distribution is
             // based on the 'dataLabels.distance' and 'dataLabels.alternate'
             // property.
```
sebastianbochan: Replace with block comment,

js/modules/timeline.src.js
```diff
@@ -472,32 +506,37 @@ undocumentedSeriesType('timeline', 'line',
...
-                    lineWidth: 0,
-                    title: null,
-                    tickPositions: []
-                },
-                timelineYAxis = {
-                    gridLineWidth: 0,
-                    title: null,
-                    labels: {
-                        enabled: false
+                conf = {
```
sebastianbochan: maybe better is use OPTIONS or PARAMS name, conf is a little bit confusing.

js/modules/timeline.src.js
```diff
@@ -100,19 +100,19 @@ undocumentedSeriesType('timeline', 'line',
              */
             connectorColor: '${palette.neutralColor100}',
```
TorsteinHonsi: The connectorColor should default to the point.color || series.color like in pie charts. https://api.highcharts.com/highcharts/series.pie.dataLabels.connectorColor

js/modules/timeline.src.js
KacperMadej: An unwanted space: 
> label 's

js/modules/timeline.src.js
```diff
@@ -447,18 +478,25 @@ undocumentedSeriesType('timeline', 'line',
...
+                    pointMarkerOptions.states[state] || {};
 
                 radius = pick(
                     pointStateOptions.radius,
                     seriesStateOptions.radius,
                     radius + (
                         seriesStateOptions.radiusPlus ||
-                    0
+                        0
                     )
```
KacperMadej: The above 4 lines fits well under 80 characters line limit -> `radius + (seriesStateOptions.radiusPlus || 0)`

js/modules/timeline.src.js
Denyllon: Suggested description:
Whether to position data labels alternately. For example, if is set equal to `100`, then data labels will be positioned alternately (on both sides of the point) at a distance of 100px.

js/modules/timeline.src.js
Denyllon: It's not good way out, because `lastPlotX` always would be `undefined`, and `closestPointRange` will never be set.

Denyllon: It's not good way out, because `lastPlotX` would always be `undefined`, and `closestPointRange` will never be set.

js/modules/timeline.src.js
```diff
@@ -145,7 +145,7 @@ undocumentedSeriesType('timeline', 'line',
              * @default 100
```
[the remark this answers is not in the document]
Denyllon: Thanks for information. Fixed also for `alternate` and `connectorWidth` parameters.

js/modules/timeline.src.js
sebastianbochan: Ok, makes sense for me!

js/modules/timeline.src.js
sebastianbochan: Talked about it on priv. Ignore this point.

js/modules/timeline.src.js
Denyllon: > There's a code block where oldTextWidth is deleted. This looks like a workaround. Check if it is still necessary after 6d6773e.

Good point. It's currently not necessary.
> In bindAxes, why force axis options? Maybe it’s better (more predictable to users) to just set xAxis.visible: false and yAxis.visible: false in the demos?

Agree, it would be more predictable to hide axes on demos.
> We have many cases where we borrow functions from the core series types, so I think we should continue doing that. In the (near) future we may have a build system that allows us to borrow functions without importing the whole parent series, but then we should change all cases consistently. Alternatively, we can create a mixin now.

Those two functions (fro Pie and Timeline) are not identical, because it's necessary to call `processData()` before redrawing the chart. We can say, that function is taken from Pie, but it's a bit extended and it's not possible to wrap it. My proposition is to do something like below:

```js
        setVisible: function (vis, redraw) {
            var point = this,
                series = point.series;

            redraw = pick(redraw, series.options.ignoreHiddenPoint);

            seriesTypes.pie.prototype.pointClass.prototype
                .setVisible.call(point, vis, false);
            // Process new data
            series.processData();

            if (redraw) {
                series.chart.redraw();
            }
        },
```
Regarding connector groups, currently they're moved to a specific data label's group, and their default colors are taken from the point options.

PR #10271, opened 2019-03-06
tools/tsdoc

bre1470:
- Added TypeDoc gulp tasks
- Updated TypeScript sources for debugging


bre1470:
@TorsteinHonsi I will use this branch as the starting point for the gulp task splits. So it should be merged soon to avoid conflicts.

PR #10275, opened 2019-03-07
 bugfix/networkgraph-dynamics

pawelfus:
I forget that now we support multiple labels per point, updated code to remove all connected dataLabels.

### DIFF
--- a/js/modules/networkgraph/networkgraph.src.js
+++ b/js/modules/networkgraph/networkgraph.src.js
@@ -794,17 +794,7 @@ seriesType('networkgraph', 'line', {
         if (this.isNode) {
             this.linksFrom.forEach(
                 function (linkFrom) {
-                    if (linkFrom.graphic) {
-                        linkFrom.graphic = linkFrom.graphic.destroy();
-                    }
-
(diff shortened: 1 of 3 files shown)


js/modules/networkgraph/networkgraph.src.js
```diff
@@ -798,8 +798,12 @@ seriesType('networkgraph', 'line', {
...
                     }
 
-                    if (linkFrom.dataLabel) {
-                        linkFrom.dataLabel = linkFrom.dataLabel.destroy();
+                    if (linkFrom.dataLabels) {
+                        linkFrom.dataLabels.forEach(function (dataLabel) {
+                            dataLabel.destroy();
+                        });
+                        linkFrom.dataLabels.length = 0;
+                        delete linkFrom.dataLabel;
```
TorsteinHonsi: I think you can do all this with one call to `Point.destroyElements`. It destroys graphics and multiple data labels.

pawelfus: Nice, thanks!

PR #10280, opened 2019-03-08
bugfix/10193-wrong-navigator-series-indexes

wchmiel:
The bug was caused mainly by the wrong navigator series indexes (regression:  https://github.com/highcharts/highcharts/commit/1ae61e6d0).

After the fix `unit-tests/stock-tools/bindings` "Flag: no duplicated flags." test failed because of the wrong series used and had to be improved.


js/parts/Dynamics.js
```diff
@@ -1103,7 +1103,10 @@ extend(Series.prototype, /** @lends Series.prototype */ {
 
             // Do the merge, with some forced options
             options = merge(oldOptions, animation, {
-                index: series.index,
+                // When oldOptions.index is null it should't be cleared.
+                // Otherwise navigator series will have wrong indexes (#10193).
+                index: oldOptions.index === undefined ?
+                    series.index : oldOptions.index,
```
TorsteinHonsi: Use the shorter `pick(oldOptions.index, series.index)`

TorsteinHonsi:
Also, please resolve the conflict in Dynamics.js.

js/parts/Dynamics.js
wchmiel: I haven't used `pick` because it return the first value that is not `null` or `undefined`. Here I want only `undefined` to be true.

TorsteinHonsi: Oh yes, it's not the same! Sorry.

Issue #9367, opened 2018-11-08
3D Funnel Series


cvasseng:

KacperMadej:
Minor features still needs to be resolved - including neck in a point feature available in funnel, but not in pyramid and 3d rendering with correct z indexing. Base 3d features are extended from 3d column series to avoid code duplication - this might result in same bugs (not crucial) as in 3d column, but it's better to later fix all than only some IMHO.

KacperMadej:
Initial commit pushed: https://github.com/highcharts/highcharts/commit/8ccc78d212c84e19a19ee38423158034c4f26253

Done:
- [ ] funnel3d module
- [ ] API docs
- [ ] demo

To do:
- [ ] article for general docs

KacperMadej:
The general docs article draft: https://docs.google.com/document/d/1qnysmpCwLR68qaYR8cdh8vZt3PhEebcfHRJZaTqdido/edit#heading=h.nc0dzws9revo

TorsteinHonsi:
Thanks, great work on the geometry! Two questions:

- Have you tried applying gradient colors to the 3D shapes? Without it, they don't really appear like 3D objects in space.
- The data label placement is a bit off in both the funnel and pyramid sample. We now have the `inside` option working on funnels, that could work good if the position is translated in 3D space. Otherwise, if not using `inside`,  the labels should have connectors pointing to the shapes.

KacperMadej:
@TorsteinHonsi 
1. Gradients were postponed (discussed with Havard), but I can take care of them now.
2. I'll check the labels placement.

Those two issue are rather minor, so maybe to avoid any complications with later code merging I would prepare a PR? It the fixed `inside` dataLabels logic committed on the hc71/master branch?

TorsteinHonsi:
> Gradients were postponed (discussed with Havard), but I can take care of them now.

Yes, please. At the very least we should have it in the demos.

> It the fixed inside dataLabels logic committed on the hc71/master branch?

It's currently on https://github.com/highcharts/highcharts/tree/feature/10036-datalabels-inside-funnel-2, awaiting some tests and checking from @wchmiel .

KacperMadej:
@TorsteinHonsi 
Linear gradient is now working fine. Radial works too, but I'm not sure where would be a good center for it - based on a shape? based on point? based on chart? I'm already using pie's `setRadialReference`.

This doesn't look good:
[...]

Issue #9366, opened 2018-11-08
3D Pyramid series


cvasseng:

KacperMadej:
Pyramid series, but in 3D
Related to 3D Funnel (#9367)

KacperMadej:
Code should be prepared for easy extend for Funnel features in mind, to avoid redundant code for any of the series. Base 3d features are extended from 3d column series to avoid code duplication - this might result in same bugs (not crucial) as in 3d column, but it's better to later fix all than only some IMHO.

KacperMadej:
General documentation article draft: https://docs.google.com/document/d/1SSbttSrNB2vejN3uIQsVJT-2ewEUl82nhDOce3-dH6k/edit#heading=h.nc0dzws9revo

KacperMadej:
Initial commit for the pyramid3d module: https://github.com/highcharts/highcharts/commit/703b8bdea5e6fa4caf25de2320fd89bc51a09c03

PR #10292, opened 2019-03-11
hc71/funnel3d

KacperMadej:
Initial commits for Funnel3d and Pyramid3d

More info and related issues:
https://github.com/highcharts/highcharts/issues/9367
https://github.com/highcharts/highcharts/issues/9366

### DIFF
--- a/js/masters/modules/funnel3d.src.js
+++ b/js/masters/modules/funnel3d.src.js
@@ -1,0 +1,10 @@
+/**
+ * @license @product.name@ JS v@product.version@ (@product.date@)
+ * Highcharts funnel module
+ *
+ * (c) 2010-2019 Kacper Madej
+ *
+ * License: www.highcharts.com/license
+ */
+'use strict';
+import '../../modules/funnel3d.src.js';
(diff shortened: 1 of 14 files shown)


KacperMadej:
* edgeWidth: 0, resolved
* gradient, resolved

-----------------

* inside dataLabels - this will be tricky in 3d, there's a lot of wrapped functions that are in use for dataLabels in 3d, so it might take more time or might require some rewrite of the related code (clean up). I'll check this tomorrow.
* translate function similarities with funnel - the code is similar, because I started with that code, but now there are some major differences and to get common code I would have to rewrite some of the funnel code, but it's doable - I just wonder if this might be a lot of work with a small gain. I'll check this after the dataLabel issue will be solved, to better find common code parts.

KacperMadej:
* dataLabels work for 3d pyramid and 3d funnel with align, verticalAlign and inside options tested.

* I'm unable to find similar code from funnel and funnel 3d to be worth a mixin.

It seems to be all for now.
@TorsteinHonsi is all fine now?

TorsteinHonsi:
All fine. Thanks for the great work!

PR #10298, opened 2019-03-11
feature/inactive-state

pawelfus:
This PR provides `inactive` state for `point` only. Does not solve the `inactive` series (e.g. hovering over a legend). Additionally, PR adds `opacity` option to the states.

### API:
(note that `linkOpacity` is set on a `series.states` level, not `series.marker.states` for Sankey and Dependency wheel)
```
  plotOptions: {
    sankey: {
      states: {
        inactive: {
          enabled: boolean
          linkOpacity: number
        }
      }
    },

    dependencywheel: {
      states: {
        inactive: {
          enabled: boolean,
          linkOpacity: number
        }
      }
    },

    networkgraph: {
      states: {
        inactive: {
          enabled: boolean,
          linkOpacity: number
        }
      },
      marker: {
        states: {
          inactive: {
            enabled: bolean,
            linkColor: Color
            dashstyle: string
            opacity: number,
            linkWidth: number
          }
        }
      }
    }
  }
```

### Demos:

- Sankey: https://highcharts.github.io/highcharts-utils/samples/#gh/35423a9fc7/sample/highcharts/plotoptions/sankey-inverted
- Dependency wheel: https://highcharts.github.io/highcharts-utils/samples/#gh/35423a9fc7/sample/highcharts/demo/dependency-wheel
- Networkgraph: https://highcharts.github.io/highcharts-utils/samples/#gh/35423a9fc7/sample/highcharts/series-networkgraph/integration-comparison

### Screenshots:
Screen Shot 2019-03-12 at 11 40 45

Screen Shot 2019-03-12 at 11 41 29
Screen Shot 2019-03-12 at 11 40 57

### Observations:
- our `setState` is quite tricky, called here and there (e.g. on `chart.destroy`) so a few sanity checks is required 
- all of the series above can share the same logic for `point.inactive` state, but Networkgraph is a bit more complicated: in networkgraph we have both, `series.marker.states` and `series.states`, while others have only `series.states`. That's why I ended up with crazy settings (see networkgraph demo changes)
- I'm not sure about the performance for huge datasets.. on mouse over we need to loop over all existing points in every series..
- It might be a good idea to call `graphic.toFront()` for elements on hover, otherwise we can accidentally hover wrong links/nodes 
- I decided to add `inherit` option to `series.setState(state, inherit)` -> when set to try, it will delegate the state to every point in that series
- IMHO, with disabled animation it looks better..

PS: Dependency wheel screenshot reminds me of VFD logo from Series of unfortunate events :D 
vfd logo


pawelfus:
Thanks! 

> Line and line-like series also have that duality.

It depends how we define it. For basic series we have states that affects line/area, and for point it affects markers.
In sankey, we have `series.states` that uses `linkOpacity` to change points opacity (nodes and links are Points)
In networkgraph we have now `series.states` to change `linkOpacity` and `series.marker.states` to change nodes.
I think it's confusing. I just realized (literally just now) that I have `series.link` config, which should be the same as `marker` (so `states` should be defined as `link.states`). Doesn't this make sense for dependancy wheel and sankey (instead `link`, should be `node`) ? 

> `states.enabled` issue

Right! I forget that it changes the marker, not the state itself.

> Missing opacity for pie series

We can add this to all series, sure, it's just PoC

> `series.state` in the basic line demo with multiple lines

Well, implementation is quite simple - just `onmouseover` for legend items calling `series.setState(..)` but later some magic happens - especially when we decide to dim other series on hover. I decided to start with `point.state` before digging further, to see the direction we want to go. Since the solution is acceptable, then I will check this.

pawelfus:
Resolved all the issues you mentioned. 

Note: If you find some issues with opacity and 3D, let me know, I needed to remove a line there, visual tests looks good, but I did not run all of visual tests.

Some demos (for now, inactive is enabled for all series): 
- https://highcharts.github.io/highcharts-utils/samples/#gh/35848f0db1/sample/highcharts/plotoptions/series-allowpointselect-column
- https://highcharts.github.io/highcharts-utils/samples/#gh/35848f0db1/sample/highcharts/demo/pie-basic

There is a few other series that don't have `opacity` implemented (those series that have overwritten `pointAttribs`):
- vector
- venn
- windbarb
- bubble (this one uses `fillOpacity` already)
- flag

Do you think we need `opacity` support for the series above?

Next week I will work on `inactive` state for legend hover and series hover.

TorsteinHonsi:
To do
 - [ ] Remove the custom inactive state in `item-series.src.js` and replace it with the built-in inactive state. It should be enabled by default and work from both series hover and legend hover.

pawelfus:
Some more examples, controlled by `series.inactiveOtherPoints` (we need a better name for this) option: 
- Line series (default: false) - https://highcharts.github.io/highcharts-utils/samples/#gh/f86f50f801/sample/highcharts/demo/line-basic
- Column series (default: false) - https://highcharts.github.io/highcharts-utils/samples/#gh/f86f50f801/sample/highcharts/demo/column-negative
- Pie series (default: true) - https://highcharts.github.io/highcharts-utils/samples/#gh/f86f50f801/sample/highcharts/demo/pie-legend
- Dependency wheel (default: true) - https://highcharts.github.io/highcharts-utils/samples/#gh/f86f50f801/sample/highcharts/demo/dependency-wheel

Option decides if `inactive` state should be applied for all other points in the same series.

pawelfus:
- Item series (default: true, inherited from Pie) - https://highcharts.github.io/highcharts-utils/samples/#gh/fe4c863152/sample/highcharts/demo/parliament-chart

pawelfus:
TO DO:
- [ ] unit-tests
- [ ] change defaults

PR #10315, opened 2019-03-14
hc71/bugfix-timeline-discrepancy-after-merge

Denyllon:
- The `alignConnector` function restored (it's better leave it defined on the prototype, instead of merging bit with `drawConnector`, because it could be useful in future).
- Additionally, zooming feature is fixed.


TorsteinHonsi:
I removed the `alignConnector` function because it is not only the alignment that should be updated on redraw, it is also other things like stroke and stroke-width.

Regarding the zoom issue - it seems odd to check for `if (!series.chart.resetZoomButton) {` - what matters is the xAxis.min and xAxis.max. What happens if someone sets the extremes using `setExtremes` instead of zooming`. Can you create another PR for the zoom issue only?

TorsteinHonsi:
I see you had some other unrelated changes too, so I'll merge this then check it after.

TorsteinHonsi:
Continuing on loose threads in https://github.com/highcharts/highcharts/issues/10323

PR #10327, opened 2019-03-17
hc71/build-tasks

bre1470:
- detail output for default task
- build scripts recognize created and deleted files and folders


bre1470:
wrong branch

PR #10338, opened 2019-03-18
networkgraph/defaults

pawelfus:
I think we can now enable the simulation by default. 

Additionally, for me `Euler` gives better results in most cases than `Verlet` + we keep similar layout between v7.0 and v7.1, so I suggest to stick to `Euler`.

### DIFF
--- a/js/modules/networkgraph/networkgraph.src.js
+++ b/js/modules/networkgraph/networkgraph.src.js
@@ -294,7 +294,7 @@ seriesType('networkgraph', 'line', {
          * @sample       highcharts/demo/network-graph/
          *               Live simulation enabled
          */
-        enableSimulation: false,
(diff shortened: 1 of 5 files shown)


pawelfus:
> I think the simulation in the main demo runs too slowly now - can we speed it up? Ideally the nodes should be within their final positions within about a second.

I think that would be way too fast (1sec ~ 60frames => 60 iterations). Changed friction, now it's faster, like in v7.0.

> It seems that setting linkFormat: '' doesn't work. It should be used instead of linkFormatter, as we should prefer configuration over callbacks in the demos whenever we can.

Done, thanks! I assume that in default options I should use `formatter` and `linkFormatter` (as it's now)? As per docs: "`format` takes precedence over `formatter`", so it's fine?

Issue #10360, opened 2019-03-19
Highcharts.SVGRenderer#button cannot use HTML


BigDrums00:
#### Expected behaviour
Pass HTML into text argument and have it render in button.
(documentation states "The text or HTML to draw.")

#### Actual behaviour
HTML is replaced with svg elements

#### Live demo with steps to reproduce

http://jsfiddle.net/9fmv1p7s/1/

#### Product version

Highcharts 7.0.3

#### Affected browser(s)
All

If the documentation is correct, the issue looks to be where label is generated in button method, useHTML is null and cant be set.

sebastianbochan:
Hi @BigDrums00,
Thank you for reporting about the problem.

As a workaround you can use `renderer.label`

```
chart.renderer.label(
            "<div class='blue'>Hello <span class='red'>People</span></div>",
            point.plotX,
            point.plotY,
            'button',
            null,
            null,
            true
        ).add();
```

Demo:
- http://jsfiddle.net/BlackLabel/c8dopybv/

**Internal note**
The `renderer.button` uses `renderer.label` but ignores the `useHTML` flag.

I will fix it as soon as possible.

PR #10366, opened 2019-03-20
bugfix/10360-button-missing-usehtml

sebastianbochan:


sebastianbochan:
Github cut the part of commit title.

Should be: "Fixed #10360, `renderer.button` ignored html tags."

PR #10369, opened 2019-03-20
hc71/test-speaking

bre1470:
Optional speaking test results helps reducing delays while waiting for results and meanwhile working on a side project.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,8 +25,8 @@
     "lib"
   ],
   "scripts": {
-    "test": "gulp test --all",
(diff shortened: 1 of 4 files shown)


TorsteinHonsi:
Nice :D

My system reads the English text with a Norwegian speech synthesis, which results in Norwenglish pronounciation. Is it possible to detect the system language and set the text accordingly?

bre1470:
Well, you can set a voice - so we can set it to Alex for proper English.

PR #10375, opened 2019-03-21
AMD dependencies

### LINKED ISSUE #8890: Loading dependent Highcharts modules or highcharts-more via RequireJS fails from ver 6.1.2.
I'm using RequireJS to dynamically load Highcharts to specific pages on my site. I have configured it with shim to load the highcharts-more and additionals modules that I'm using. Starting from few days ago with the version 6.1.2 update in Highchart CDN, it now fails to correctly load the highcharts-more and the modules.

#### Expected behaviour

I expect to be able to use the highcharts-more and other modules' features.

#### Actual behaviour

The files are fetched, but initializing Highcharts fails with error code 17 unknown chart type for `arearange` and other weird behaviour. For example polar area charts are not polar etc.
It does fetch the actual files, but for some reason does not account for the additional features anymore.

Working demo with Highcharts 6.1.1: http://jsfiddle.net/3kpazfw4/
Failing demo with CDN: http://jsfiddle.net/6zypqedn/

#### Product version
Highcharts 6.1.2

#### Affected browser(s)
Chrome 68.

jon-a-nygaard:
# Description
Loading our modules in AMD is currently a bit cumbersome as it requires the user to load all modules and its dependencies and apply them to the Highcharts instance. It also requires the user to have knowledge of said dependencies and which order to apply them.

This could be made a lot easier by defining the dependencies when we declare a AMD module in our UMD pattern, and let RequireJS or similar libraries handle the order of the loading.

Solved by updating the `highcharts-assembler` to `v1.3.2` to have support for `@module` and `@requires` tags in our files, and then add these necessary tags to all our master files which we build our distributed files from.

# Related issue(s)
- Closes #8890

### DIFF
--- a/js/masters/highcharts-3d.src.js
+++ b/js/masters/highcharts-3d.src.js
@@ -1,5 +1,7 @@
 /**
  * @license @product.name@ JS v@product.version@ (@product.date@)
+ * @module highcharts/highcharts-3d
+ * @requires highcharts
  *
  * 3D features for Highcharts JS
  *
--- a/js/masters/highcharts-gantt.src.js
+++ b/js/masters/highcharts-gantt.src.js
@@ -1,5 +1,6 @@
 /**
  * @license @product.name@ JS v@product.version@ (@product.date@)
+ * @module highcharts/highcharts-gantt
  *
  * (c) 2017-2018 Lars Cabrera, Torstein Honsi, Jon Arild Nygard & Oystein Moseng
  *
--- a/js/masters/highcharts-more.src.js
+++ b/js/masters/highcharts-more.src.js
@@ -1,5 +1,7 @@
 /**
  * @license @product.name@ JS v@product.version@ (@product.date@)
+ * @module highcharts/highcharts-more
+ * @requires highcharts
  *
  * (c) 2009-2018 Torstein Honsi
  *
--- a/js/masters/highcharts.src.js
+++ b/js/masters/highcharts.src.js
@@ -1,5 +1,6 @@
 /**
  * @license @product.name@ JS v@product.version@ (@product.date@)
+ * @module highcharts/highcharts
  *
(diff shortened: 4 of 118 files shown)


bre1470:
I had no chance to test it in a TypeScript project yet, because JSDoc plugin breaks on the tags in the code folder. Please remove the tags during processing, so that in code folder the tags are gone.

jon-a-nygaard:
Thanks @bre1470  👍 I have now updated the assembler to a new version which removes these tags.
Please checkout the latest.

bre1470:
The AMD example in the readme should be updated. Suggestion:

## Load Highcharts from the CDN as an AMD module
Highcharts is compatible with AMD module loaders (such as RequireJS). The following example demonstrates loading Highcharts along with two modules from our CDN using RequireJS.
```html
<html>
    <head>
        <script src="require.js"></script>
        <script>
            require.config({
                packages: [{
                    name: 'highcharts',
                    main: 'highcharts'
                }],
                paths: {
                    // Change this to your server if you do not wish to use our CDN.
                    'highcharts': 'https://code.highcharts.com'
                }
            });
        </script>
    </head>
    <body>
        <div id="container"></div>
        <script>
            require([
                'highcharts/modules/exporting',
                'highcharts/modules/accessibility'
                // This function runs when the above files have been loaded.
            ], function (ExportingModule, AccessibilityModule) {
                const Highcharts = ExportingModule.Highcharts;
                // Create a test chart.
                    series: [{
                        data: [1,2,3,4,5]
                    }]
                });
            });
        </script>
    </body>
</html>
```

bre1470:
I updated my readme suggestion. Great work btw, @jon-a-nygaard ! 👍

TorsteinHonsi:
@bre1470 Will you create a PR for the readme?

bre1470:
Yes, I will do.

PR #10380, opened 2019-03-21
hc71/6870-axis-tickpositions

### LINKED ISSUE #6870: Axis breaks can cause axis labels to be hidden
Although I expected axis labels to display regardless of the presence of axis breaks, when the breaks span the time that is automatically calculated for the axis ticks, no axis labels are displayed.  This can be seen in the following fiddle: http://jsfiddle.net/moses_seeq/k75j6a3f/.  This is a contrived example to display the issue, but it is expected that when tick marks are dynamically selected that ticks outside of the breaks would be selected and displayed.

bre1470:
- Made Axis.tickPositions public. Not readonly because it is already optional. Fix #6870
- Fixed some commit issues.

### DIFF
--- a/js/parts/Axis.js
+++ b/js/parts/Axis.js
@@ -4632,7 +4632,6 @@ H.extend(Axis.prototype, /** @lends Highcharts.Axis.prototype */{
          *
          * @name Highcharts.Axis#tickPositions
          * @type {Array<number>|undefined}
-         * @readonly
          */
         this.tickPositions =
             // Find the tick positions. Work on a copy (#1565)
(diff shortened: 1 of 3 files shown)


js/parts/Axis.js
```diff
@@ -4627,9 +4627,15 @@ H.extend(Axis.prototype, /** @lends Highcharts.Axis.prototype */{
                 options.allowDecimals !== false
             );
 
-        // Find the tick positions. Work on a copy (#1565)
-        this.tickPositions = tickPositions =
-            tickPositionsOption && tickPositionsOption.slice();
+        /**
+         * Contains all ticks that are laid out on the axis.
+         *
```
TorsteinHonsi: Note to self: Write a more concise explanation and warn that it should not be changed.

PR #10386, opened 2019-03-21
bugfix/10106-showempty-and-hasdata-fix

### LINKED ISSUE #10106: axis.showEmpty does not hide title in heatmap
#### Expected behaviour

Axis titles should be hidden, similar to the behaviour for a line series: https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/yaxis/showempty/

#### Actual behaviour

Axis titles are still displayed, even if there is no data being shown.

https://jsfiddle.net/monicajianu/64owL3df/

#### Product version
Highcharts v7.0.2

#### Affected browser(s)
All

Basts:
Fixed #10106, moving hasData from no-data-modules to Axis.js and Series.js to repair showEmpty feature.

### DIFF
--- a/js/parts-map/ColorAxis.js
+++ b/js/parts-map/ColorAxis.js
@@ -558,6 +558,12 @@ extend(ColorAxis.prototype, {
         });
     },
 
+    // Define hasData function for non-cartesian seris.
+    // Returns true if the series has points at all.
+    hasData: function () {
+        return !!(this.tickPositions && this.tickPositions.length);
+    },
+
     /**
      * Override so that ticks are not added in data class axes (#6914)
      *
--- a/js/parts-map/ColorAxis.js
+++ b/js/parts-map/ColorAxis.js
@@ -558,7 +558,7 @@ extend(ColorAxis.prototype, {
         });
     },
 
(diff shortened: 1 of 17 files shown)


js/parts-map/ColorAxis.js
```diff
@@ -558,6 +558,12 @@ extend(ColorAxis.prototype, {
         });
     },
 
+    // Define hasData function for non-cartesian seris.
```
KacperMadej: The comment is about series, but this is a colorAxis.
Also, there's another problem with comments - a repeated typo: `seris`

js/parts/Axis.js
```diff
@@ -5591,15 +5591,8 @@ H.extend(Axis.prototype, /** @lends Highcharts.Axis.prototype */{
...
-        return (
-            this.hasVisibleSeries ||
-            (
-                defined(this.min) &&
-                defined(this.max) &&
-                this.tickPositions &&
-                this.tickPositions.length > 0
-            )
-        );
+        return this.series.some(s => s.hasData()) ||
```
TorsteinHonsi: Arrow functions can't be used in source code yet.

Issue #29, opened 2010-07-20
Pie chart tooltips not hiding on mouse out


TorsteinHonsi:
Demo at http://jsfiddle.net/mqk8S/

Reported at http://highslide.com/forum/viewtopic.php?f=9&t=6944&view=unread#unread

PR #10393, opened 2019-03-22
hc71/fix-sonification-ie

### LINKED ISSUE #10391: Sonification reports error #29 in IE tests
- [x] Sonification should handle unsupported browsers.
- [x] We have to make sure, that sonification tests get skipped in IE too.

bre1470:
- Solved sonification error #29 in IE.  Fix #10391
- Skipping sonification tests in IE
- Linting
- Karma polyfill turned off

### DIFF
--- a/js/modules/sonification/Instrument.js
+++ b/js/modules/sonification/Instrument.js
@@ -144,7 +144,9 @@ function Instrument(options) {
 }
 Instrument.prototype.init = function (options) {
     if (!this.initAudioContext()) {
-        H.error(29);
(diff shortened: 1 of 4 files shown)


bre1470:
@oysteinmoseng 
I try now to get rid of the error in karma. There are more unit tests triggering it than the two specifically for sonification. Do you have an idea where I have to look for the other ones?

oysteinmoseng:
@bre1470 Probably the sonification demo, you can try searching for Sonification, and it shouldn't be many

PR #10396, opened 2019-03-22
bugfix/10392-annotations-each-msie

sebastianbochan:


sebastianbochan:
1. Thank you for the suggestion, I updated it.
2. I do not have any error, what kind of you have?

PR #10397, opened 2019-03-22
bugfix/10323-timeline-zooming-and-connectors

Denyllon:
- Setting extremes and zooming fixed.
- Connectors appears correctly.


TorsteinHonsi:
Thanks 👍

PR #10424, opened 2019-03-26
feature/inactive-series-group

pawelfus:
Inactive state: applied series.opacity for groups instead of path elements.

### DIFF
--- a/js/parts/Interaction.js
+++ b/js/parts/Interaction.js
@@ -825,6 +825,8 @@ extend(Point.prototype, /** @lends Highcharts.Point.prototype */ {
             halo = series.halo,
             haloOptions,
             markerAttribs,
+            pointAttribs,
+            pointAttribsAnimation,
             hasMarkers = markerOptions && series.markerAttribs,
             newSymbol;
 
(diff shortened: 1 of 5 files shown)


js/parts/Interaction.js
```diff
@@ -1186,6 +1205,23 @@ extend(Series.prototype, /** @lends Highcharts.Series.prototype */ {
                         i = i + 1;
                     }
                 }
+
+                // For some types (pie, networkgraph, sankey) opacity is
+                // resolved on a point level
+                if (!series.options.inactiveOtherPoints) {
+                    [
+                        series.group,
+                        series.dataLabelsGroup
```
TorsteinHonsi: What about the `series.markerGroup`? Are we still looping over every `point.graphic` and setting its opacity? In that case, wouldn't it be more effective to set the series opacity on the marker group?

pawelfus: Yes, we still have `marker.states.inactive.something` option. Should we remove that?

TorsteinHonsi: We must make sure that when we set a whole _series_ to inactive (like a line series), we don't loop over the points and animate the opacity of each. In this case, setting the opacity on the group should be enough.

js/parts/Interaction.js
```diff
@@ -1186,6 +1205,23 @@ extend(Series.prototype, /** @lends Highcharts.Series.prototype */ {
...
+                    ].forEach(
+                        function (group) {
+                            if (group) {
+                                group.attr({
```
TorsteinHonsi: Should be `.animate`. Both this problem and the above (individual handling of markers), can be seen in http://utils.highcharts.local/samples/#view/highcharts/demo/windbarb-series when hovering the wind barbs.

js/parts/Interaction.js
pawelfus: Ok 👍

js/parts/Interaction.js
pawelfus: Ok 👍

js/parts/Interaction.js
pawelfus: Done

js/parts/Interaction.js
pawelfus: Done

TorsteinHonsi:
Works very well now, I've manually tested all the samples in the online demos. Found a couple minor issues, but I will merge this and handle them from the master.

Issue #9124, opened 2018-10-04
Zoom in/out when using zones not working with boost


Abrissirba:
#### Expected behaviour
When zooming in and out I expect to only have one line per serie

#### Actual behaviour
If I zoom in and out multiple times I sometimes get two lines for a serie. This happens when I use zones together with the boost module.

#### Live demo with steps to reproduce
http://jsfiddle.net/81vhr764/5/

The fiddle shows a use case where I want multiple axes with one serie per axis stacked in the same graph. I use a zone for each graph to hide any points that lays below the axis min value. If I zoom in and out multiple times, sometimes the bottom serie get multiple lines visible in the graph. 

#### Product version
Highcharts 6.1.1

#### Affected browser(s)
Tested in chrome

pawelfus:
Hi @Abrissirba 

Thank you for reporting the issue! You are right, it looks broken.

**Workaround:**
Use only boosted rendering engine by setting [`boost.seriesThreshold=1`](https://api.highcharts.com/highstock/boost.seriesThreshold).

Demo: http://jsfiddle.net/BlackLabel/ks9zqc07/2/

**Simplified demo:**
http://jsfiddle.net/BlackLabel/ks9zqc07/

*Internal note:*
It's probably caused by generating separate shapes for each zone and boost is not clearing them all.

yoderexploder:
I'm interested in this as well, and I see you committed changes referencing this issue.  Is there anything that needs to be done to get zones working with Boost?  Thanks so much.

pawelfus:
Hi @yoderexploder 

My demo above seems to be fine ( http://jsfiddle.net/BlackLabel/ks9zqc07/ ) - could you recreate issue in jsFiddle? In case something is still broken, please consider using my workaround posted above, it should help.

yoderexploder:
Thanks @pawelfus, for your response.  

I think I've just figured out why my zones were not working the Boost module.  It looks as if you must define a 'value' for every zone.  This is in contrast to not using Boost where the value for the final zone does not need to be defined.  I've worked around this by just setting the value of the last zone to Infinity.  Example here: [https://jsfiddle.net/09cxhn7f/](https://jsfiddle.net/09cxhn7f/)

I am still having a problem after resetting the zoom.  The zoomed image remains after zooming back out.  The same fiddle demonstrates that problem.  I think you've got an issue logged for this already, i'll see if I can find it and participate there.

yoderexploder:
[...]

Issue #8288, opened 2018-05-07
Boost module does not render areaspline in canvas


ghost:
The boost module does not boost areaspline, although it is supporting it internally.

Example: https://jsfiddle.net/7ss4um2q/1/

Jaibharath:
Is this resolved. I am using the areaspline type with (zones, dash styles)

Jaibharath:
Please let me know, as I need to show more then 1000 points in areaspline chart. rendering is too slow even after using the boost module. I tried checking with type 'line' with boost module, it worked, but it is not working for areasplin type.

sebastianbochan:
Hi @Jaibharath,
At this moment areasplines are still not supported, but could you simplify your demo and post here? It will allow me to verify if any workaround or different solution is possible (like dataGrouping, lazy loading etc).

Jaibharath:
Ok, We are using angular 6 with highcharts, I conffigured the datagrouping with different zones for each point. Once the series of data is ready through the scripting then I configured the config options with type of chart, series and axis, zoom options, We are using the areaspline type relating which related to our design, It worked until like 300 - 400  xaxis label, Once the labels increasing, chart loading taking more memory. Then I tried using the  Boost module which did not work. I tried with type 'line' boost module worked. But with line also once I zoomed in chart is not removing after I zoomed out.

Jaibharath:
Also when trying with spline+boost module not working. with line+boost module. on reset zoom the chart is not hiding.  Its showing the dense line chart and also the zoomed chart on reset zoom

pawelfus:
At this moment, `spline`-type is not supported (list of supported types is available in our docs: https://www.highcharts.com/docs/advanced-chart-features/boost-module ). If series on reset zoom does not hide, please recreate issue in jsFiddle (without angular). Thanks!

Jaibharath:
Hi @pawelfus I dont have the jsfiddle, please replace js below code in below jsfiddel link(https://jsfiddle.net/gh/get/jquery/1.7.2/highslide-software/highcharts.com/tree/master/samples/highcharts/boost/line/)




function getData(n) {
var dataObj = {};
dataObj.arr = [];
dataObj.zones = [];
    
        var i;
        var a;
       	var b;
        var c;
        var color;
        var spike;
    for (i = 0; i < n; i = i + 1) {
        if (i % 100 === 0) {
            a = 2 * Math.random();
        }
        if (i % 1000 === 0) {
            b = 2 * Math.random();
[...]

PR #10439, opened 2019-03-28
feature/boost-zones

cvasseng:
Feature: Added zone support in boost. See #8288 and #9124.

### DIFF
--- a/js/modules/boost/wgl-renderer.js
+++ b/js/modules/boost/wgl-renderer.js
@@ -1,12 +1,12 @@
-/* *
+/**
  *
- *  Copyright (c) 2019-2019 Highsoft AS
+ * Copyright (c) 2019-2019 Highsoft AS
  *
- *  Boost module: stripped-down renderer for higher performance
+ * Boost module: stripped-down renderer for higher performance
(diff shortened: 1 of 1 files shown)


js/modules/boost/wgl-renderer.js
```diff
@@ -602,6 +620,28 @@ function GLRenderer(postRenderCallback) {
                 continue;
             }
 
+            // Note: Boost requires that zones are sorted!
+            if (zones) {
+                pcolor = zoneDefColor.rgba;
```
tzaitoun, line 625: @cvasseng
I am getting this bug with this line: [example](https://jsfiddle.net/8wujdtx0/)
The curve should show purple up to a y value of 3, red up to a value of 5, and **zoneDefColor** till infinity (Line 285).
But I think this happens because you mutate the zoneDefColor.rgba when looping through each point, so the rgba is normalized every time and rgb becomes [0, 0, 0] which is black.

Assuming no performance impact, we can simply clone the rgb for each point e.g.
Line 625:  `pcolor = zoneDefColor.rgba.slice()`

PR #10444, opened 2019-03-28
bugfix/3d-columns-inactive

pawelfus:
Regression: Inactive state didn't work correctly for grouped columns in 3D.

I couldn't find any reason for `opacitySetter` in 3D columns, tested all visual 3D column samples and it works fine. I don't like the solution below (random `preventGroupOpacity` property) but it's the easiest way to change how opacity works in 3D columns.


pawelfus:
I think it's wrong solution after all, I will PR something else shortly.

PR #10448, opened 2019-03-29
docs/typescript-fixes

### LINKED ISSUE #10178: Docs: Exporting module pollutes the global scope.
See https://api.highcharts.com/class-reference/global

There should be only GlobalSVGElement visible.

bre1470:
- Fixed #10178 - unintended option spawning in global scope.
- Docs: Fixed option types.
- Tools: Added organization declaration and more tests.
- Tools: Improved options parsing.

Note: PackedBubbleSeries has somehow options excluded that were available in version 7.0 like `maxSize` and `minSize`. Please double check if it is intended or a mistake.


sebastianbochan:
The `minSize` and `maxSize` are still part of API and should not be removed.

PR #10455, opened 2019-03-29
bugfix/10451-failing-tests-in-firefox

wchmiel:
This issue was probably not related to a browser but to a width of the screen. When the chart was truncated `controller.pan()` method was not able to select the chart area outside of the browser window. I changed the approach of testing it and now pass a fake event directly to `chart.zoom()` method. I tested it on my device and it works correctly on all browsers I have. Could you test it on yours?


TorsteinHonsi:
Thanks, but this was already fixed on master.

PR #10473, opened 2019-04-02
bugfix/10013-boost-hide-show-series

### LINKED ISSUE #10013: Series don't hide when expected with boost
Sorry if this is a duplicate, there seems to be a tonne of issues around boost and hiding series, but I couldn't find a similar one to this!

#### Expected behaviour

Hiding a series hides the series.

#### Actual behaviour

Hiding a series doesn't hide the series.

https://jsfiddle.net/p2kxm7rv/

Sorry, that's a tonne of data. Couldn't replicate the bug after removing some data.

- Click GoGo Zooom to hide
- Click Konex to hide
- Click MayZing to hide

MayZing doesn't hide.

Clicking randomly on different series does all sorts of weird stuff.

Here it is in gif form:

ezgif com-optimize 8

#### Product version

Highcharts 6.11 and Highcharts 7

#### Affected browser(s)

All.

pawelfus:
Fixed #10013, series did not hide markers when other series were in boosted mode.

Checked also boost visual tests.

Maybe I misunderstand [`boost.allowForce`](https://api.highcharts.com/highcharts/boost.allowForce) option:

> allowForce: boolean Since 1.0.0
> If set to true, the whole chart will be boosted if one of the series crosses its threshold, and all the series can be boosted.

 Description is a bit vague: **"and all the series can be boosted"** - suggests support for boosting, but from the source code it seems `series.boostThreshold` is taken into account too. What is the point of that option if only one crosses it's `boostThreshold` and other don't? Do I miss something?

### DIFF
--- a/js/modules/boost/boost-init.js
+++ b/js/modules/boost/boost-init.js
@@ -136,6 +136,12 @@ function init() {
             // If we're rendering per. series we should create the marker groups
             // as usual.
             if (!chart.isChartSeriesBoosting()) {
+                // If all series were boosting, but are not anymore
+                // restore private markerGroup
+                if (this.markerGroup === chart.markerGroup) {
+                    this.markerGroup = undefined;
+                }
+
                 this.markerGroup = series.plotGroup(
                     'markerGroup',
                     'markers',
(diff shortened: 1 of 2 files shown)


TorsteinHonsi:
@cvasseng Any input on Pawel's comments?

cvasseng:
@pawelfus It takes `series.boostThreshold` into account if it's `0` because if it's `0` boosting for that particular series is supposed to be unconditionally disabled,  so it's treated the same as a series that can't be boosted. The reason in general is that we can't do a chart-wide force (i.e. shared canvas for all boosted series) if there are series that can't be boosted due to z-index woes (example: if there are three series, [a, b, c] and a and c is boosted in a shared context, they will both end up on the same z-level, which would be unexpected in terms of "regular" behavior).

PR #10476, opened 2019-04-02
bugfix/10413-tweak-addPoint-function

### LINKED ISSUE #10413: addPoint() does not return point
The one-liner `return point;` is missing from addPoint() so it returns nothing. Please hear out my use-case:

Since there is no way to toggle visibility of categories, I must remove the points manually to "hide" them. If storing the added points were possible:

#### Expected behaviour
```
// Adding data
let points = [];

for (let value of category1.data) {
   // Store the added point!!
   let point = series.addPoint({
      name: 'Category 1',
      y: value
   });

   points.push(point);
}
```

Then I would be able to do this:

```
// Hiding a category
for (let point of points) {
   point.remove();
}
```

#### Actual behaviour
But instead I have to iterate through the entire series to find the points to remove for one category:
```
for (let points of series.points) {
   if (point.name === 'Category1') {
      point.remove();
   }
}
```

https://jsfiddle.net/lanimelrok/1b6z8vqk/

Looking at the source code, it looks like just a one-liner to return the point. Could we please return the added point so that I won't have to iterate through possibly hundreds of points in a series just to delete a few? Thank you so much for your understanding. 

https://github.com/highcharts/highcharts/blob/891c59ae75d8118a6da7bd2e0efb90d16c599c03/js/parts/Dynamics.js#L825-L897

Basts:
Fixed #10413, tweak addPoint function to returning added point.

### DIFF
--- a/js/parts/Dynamics.js
+++ b/js/parts/Dynamics.js
@@ -845,6 +845,9 @@ extend(Series.prototype, /** @lends Series.prototype */ {
      * @param {boolean} [withEvent=true]
      *        Used internally, whether to fire the series `addPoint` event.
      *
+     * @return {Highcharts.Point}
+     *         The newly generated point object.
+     *
      * @fires Highcharts.Series#event:addPoint
      */
     addPoint: function (options, redraw, shift, animation, withEvent) {
@@ -856,8 +859,8 @@ extend(Series.prototype, /** @lends Series.prototype */ {
             names = xAxis && xAxis.hasNames && xAxis.names,
             dataOptions = seriesOptions.data,
             point,
-            isInTheMiddle,
             xData = series.xData,
+            PointClass = series.pointClass,
             i,
             x;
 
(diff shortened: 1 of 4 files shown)


js/parts/Dynamics.js
```diff
@@ -924,6 +924,7 @@ extend(Series.prototype, /** @lends Series.prototype */ {
         if (redraw) {
             chart.redraw(animation); // Animation is set anyway on redraw, #5665
         }
+        return point;
```
TorsteinHonsi: We also need to update the doclet so that JSDoc is aware of the correct return value.

samples/unit-tests/series/addpoint/demo.js
```diff
@@ -0,0 +1,15 @@
...
+    var chart = Highcharts.chart('container', {
+        series: [{
+            data: [3, 4, 5]
+        }]
+    });
+
+    var point = chart.series[0].addPoint(8);
+
+    assert.strictEqual(
+        typeof (point),
```
TorsteinHonsi: Consider a stricter test, `point instanceof Highcharts.Point`.

PR #10486, opened 2019-04-03
bugfix/10471-fixing-pareto-baseSeries-0

### LINKED ISSUE #10471: Pareto chart not working on a baseindex of 0
#### Expected behaviour
When the baseindex is set to 0 for a pareto chart the line should be displayed

#### Actual behaviour
When the baseindex is set to 0 for a pareto chart the pareto line is not displayed

https://jsfiddle.net/zc8nyhwq/

the chart should display the pareto line for both Complaints0 and Complaints1 but it is only displayed for Complaints1

#### Product version
6.2, 7.1

Basts:
Fixed #10471, Pareto series wasn't rendering when baseSeries was set to the first series by 0 number.

### DIFF
--- a/js/mixins/derived-series.js
+++ b/js/mixins/derived-series.js
@@ -64,7 +64,7 @@ var derivedSeriesMixin = {
         var chart = this.chart,
             baseSeriesOptions = this.options.baseSeries,
             baseSeries =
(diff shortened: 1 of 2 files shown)


sebastianbochan:
Good point, thanks!

PR #10512, opened 2019-04-08
Feature/circleci integration

activist:


activist:
Note that karmaconfig has changed to read browserstack credentials from env variables rather than from a file.

PR #10517, opened 2019-04-09
enhancement/8854-pie-datalabels-distance-percentage

kamilkulig:


js/parts/PieSeries.js
```diff
@@ -765,6 +750,28 @@ seriesType('pie', 'line',
...
+                    ),
+                    labelDistance
+                );
+
+                // Compute point.labelDistance if it's defined as percentage
+                // of slice radius (#8854)
+                if (H.isString(point.labelDistance)) {
+                    point.labelDistance = point.shapeArgs.r *
+                      H.relativeLength(point.labelDistance, 1);
+                }
```
TorsteinHonsi: The `relativeLength` function has a built-in check for whether the parameter is a percentage string. If it is a number it simply returns the number, ignoring the percentage base. So these four lines can simply be reduced to:
`point.labelDistance = H.relativeLength(point.labelDistance, point.shapeArgs.r)`

PR #10548, opened 2019-04-12
bugfix/10450-init-animate-clip-series

KacperMadej:


KacperMadej:
@TorsteinHonsi Please check the addition. I'm not sure if this is what you have in mind, so please let me know if and how could I improve the code further.

js/parts/Series.js
```diff
@@ -4084,22 +4125,35 @@ H.Series = H.seriesType(
...
+                    } : {
+                        width: chart.chartWidth,
+                        x: -xAxis.pos
+                    };
+                } else {
+                    finalBox = {
                         width: chart.plotSizeX,
                         x: 0
-                    }, animation);
+                    };
```
TorsteinHonsi: Can we use the new `getClipBox` for this finalBox as well? In the doclet it says it can be called twice, but it is actually only called once.

KacperMadej: Yes, done in the commit below. The doclet description is fixed too.

Issue #10569, opened 2019-04-16
Series line disappears when you remove zones


f1ghtingfalcons:
#### Expected behaviour
Adding color zones to a series should color the line based on what zone its in. Removing the zones from the series should return the line to normal colors.

#### Actual behaviour
Removing zones from a series removes the line ( but the line points stay and are colored correctly ).

#### Live demo with steps to reproduce
https://jsfiddle.net/f1ghtingfalcons/8sekb5qw/9/

1) click the `remove zones` button to set zones to []. You'll see that the points themselves return to normal color, but the line disappears.

#### Product version
Works correctly with version 7.0 of highcharts, but broken in version 7.1

#### Affected browser(s)
Chrome, Firefox, Edge

sebastianbochan:
Hi @f1ghtingfalcons,
Thank you for reporting about the problem.

_Internal note:_
The last working version: 
- v6.2 - https://jsfiddle.net/BlackLabel/zk83dg4e/

Bisect leads me to https://github.com/highcharts/highcharts/commit/4dc2b38

pawelfus:
**Workaround (without styled mode)**
Demo: https://jsfiddle.net/BlackLabel/hpz5yv13/

Snippet:
```js
Highcharts.wrap(Highcharts.Series.prototype, 'applyZones', function(proceed) {

  proceed.apply(this, Array.prototype.slice.call(arguments, 1));

  if (this.zones.length === 0) {
    if (this.graph) {
      this.graph.show();
    }
    if (this.area) {
      this.area.show();
    }
  }
});
```

Fredo70:
I try the workaround. Without zones, the line is not hidden on click on the legend to hide the series.

Basts:
As a @Fredo70 noticed - after removing zones legend doesn't work properly (the line is not hidden). 

[Demo with master source code](https://jsfiddle.net/BlackLabel/amo23fcu/)

FYI @pawelfus

pawelfus:
Thanks!

PR #10662, opened 2019-05-06
bugfix/regression-10569

pawelfus:
Fixed regression after #10569, which forced graph's visibility.

### DIFF
--- a/js/parts/Series.js
+++ b/js/parts/Series.js
@@ -4893,10 +4893,10 @@ H.Series = H.seriesType(
             } else if (series.visible) {
                 // If zones were removed, restore graph and area
                 if (graph) {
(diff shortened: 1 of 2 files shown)


js/parts/Series.js
```diff
@@ -4893,10 +4893,10 @@ H.Series = H.seriesType(
             } else if (series.visible) {
                 // If zones were removed, restore graph and area
                 if (graph) {
-                    graph.show();
+                    graph.attr({ visibility: 'inherit' });
```
TorsteinHonsi: Use `graph.show(true)` instead.

pawelfus: Nice, thanks!

PR #10674, opened 2019-05-07
bugfix/10450-addition-test

KacperMadej:
Tested on Win10 / Chrome, FF, IE11, Edge - all passed.


KacperMadej:
...and yet it fails on headless FF. I'll try to debug this.

PR #10680, opened 2019-05-07
bugfix/10618-scrollbar-overlapping

### LINKED ISSUE #10618: Scrollbar: mixed opposite with multiple vertical axes
#### Behaviour
yAxis with mixed opposite options are causing some scrollbar to be overlapping, so only one is visible.

https://jsfiddle.net/BlackLabel/4ckmpL29/1/

Both yAxes should have scrollbars.

#### Product version
7.1.1

#### Workaround
Adjust position of the scrollbar using [`margin`](https://api.highcharts.com/highstock/yAxis.scrollbar.margin) option.
Demo: https://jsfiddle.net/BlackLabel/4ckmpL29/2/

pawelfus:
Highstock: Fixed #10618, different `yAxis.opposite` settings caused scrollbars to overlap.

### DIFF
--- a/js/parts/Scrollbar.js
+++ b/js/parts/Scrollbar.js
@@ -1053,7 +1053,9 @@ if (!H.Scrollbar) {
                 pick(axis.dataMax, axis.max) // #6930
             ),
             scrollbar = axis.scrollbar,
-            titleOffset = axis.titleOffset || 0,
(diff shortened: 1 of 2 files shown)


pawelfus:
Thanks! I thought the same, but I decided to repeat myself in favor of readability. Why? Because it's still not a perfect solution, so I expect more use-cases where it's broken. Unless we have more tests I would wait with refactoring.

PR #10710, opened 2019-05-13
Bugfix/addevent order

jon-a-nygaard:
Fixed issue with sorting of events when addEvent is called with parameter options.order value of undefined.

---
# Description
Sorting of event orders went wrong when either options or options.order was undefined.


jon-a-nygaard:
Closing this PR, as I would like to suggest a different approach.

PR #10716, opened 2019-05-13
bugfix/6644-plotline-panes-stock

### LINKED ISSUE #6644: bug: highstock plotbands for one x-axis copied to all other x axes
see example :

http://jsfiddle.net/s01ew96z/1/

each series has their own x-axis
plotbands are only defined for the first x-axis
but show on the others as well

pawelfus:
Added new options `plotBands.acrossPanes` and `plotLines.acrossPanes` to control rendering lines across all panes. See #6644.
___
Note: I decided to refactor a bit `Axis.getPlotLinePath()` method instead of adding n-th argument. Let me know if this is fine. Alternatively we can just add another param to `getPlotLinePath()`.

@bre1470 - could you check `AxisPlotLinePathOptionsObject` interface?

### DIFF
--- a/js/parts-3d/Axis.js
+++ b/js/parts-3d/Axis.js
@@ -238,8 +238,8 @@ wrap(Axis.prototype, 'getPlotBandPath', function (proceed) {
         from = args[1],
         to = args[2],
         path = [],
-        fromPath = this.getPlotLinePath(from),
-        toPath = this.getPlotLinePath(to);
+        fromPath = this.getPlotLinePath({ value: from }),
+        toPath = this.getPlotLinePath({ value: to });
 
     if (fromPath && toPath) {
         for (var i = 0; i < fromPath.length; i += 6) {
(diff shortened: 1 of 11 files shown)


pawelfus:
I'm not sure which label will suit here better: Bugfix or Feature - up to you @TorsteinHonsi

TorsteinHonsi:
> I'm not sure which label will suit here better: Bugfix or Feature

Since it's a new option, it belongs in the "Feature" section at the top of the changelog.

I also changed the description to "Added new options plotBands.acrossPanes and plotLines.acrossPanes to control rendering lines across all panes. See #6644.". Since it's a feature we don't need to start with "Fixed ...", (and we don't need to use the auto-close keywords. And by spelling out the full option paths they should be automagically linked to the API in the website changelog.

pawelfus:
> Thanks for the PR!
> 
> If we watn to change the `getPlotLinePath` arguments to an options object, which I think is a good idea at this point, we should probably go all the way. So that the `old` and `pos` arguments in `ColorAxis`, and `reverse` in `RadialAxis` (which looks like the same as `lineWidth`?, and `translatedValue` in Axis.js, are also set at object properties. What do you think?

Sure, I like that idea too 👍

PR #10722, opened 2019-05-14
Tools: Fixed issue with disconnected Edge when testing on BrowserStack

TorsteinHonsi:
A loop was running over numerous DOM nodes for each step of `TestController.moveTo`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -185,11 +185,6 @@ workflows:
           requires:
             - build_dependencies
           browsers: "Mac.Firefox"
-      - test_browsers:
-          name: "test-Win.Edge"
-          requires:
-            - build_dependencies
-          browsers: "Win.Edge --split 2"
       - hold_for_approval:
           type: approval
           filters:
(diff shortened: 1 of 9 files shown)


TorsteinHonsi:
@activist The tests are now passing when I run `gulp test --browsers Win.Edge --force` from the command line. 

Is there anything different in the setup when it runs from CircleCI? Memory/CPU or other things related to performance? The tests are running almost twice as fast when ran from the command line, also indicating some difference in the resource assignment.

activist:
@TorsteinHonsi No, I don't see any particular reasons as to why it should work locally, but not on CI. The Circle CI CPU might be slightly slower than your own device, also it will wait for other browserstack tests to complete before starting. The available node memory is slightly increased compared to the defaults, so it depends if you have changed this on your own device?

Also CI runs the tests outside of gulp as we had different disconnect issue earlier when running via gulp. CI command is currently: `npx cross-env NODE_OPTIONS=--max_old_space_size=3072 karma start test/karma-conf.js --single-run --browsers << parameters.browsers >>`

PR #10732, opened 2019-05-15
Improved performance for exporting packedbubble series.

Izothep:
Improved performance for exporting packedbubble series.
___
With current packed bubbles series the process of exporting an image may take more than 6500ms. 

It was needed to add independent **seriesBbox** method, calculating bBox for bubble series without rendered elements. 

With new approach it was possible to reduce exporting time to 2100ms.

### DIFF
--- a/js/parts-more/PackedBubbleSeries.js
+++ b/js/parts-more/PackedBubbleSeries.js
@@ -715,15 +715,21 @@ seriesType(
          */
         seriesBox: function () {
             var series = this,
+                chart = series.chart,
                 data = series.data,
                 max = Math.max,
                 min = Math.min,
                 radius,
                 // bBox = [xMin, xMax, yMin, yMax]
-                bBox = [Infinity, -Infinity, Infinity, -Infinity];
+                bBox = [
+                    chart.plotLeft,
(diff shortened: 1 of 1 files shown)


js/parts-more/PackedBubbleSeries.js
```diff
@@ -705,6 +707,35 @@ seriesType(
...
+         * @private
+         */
+        seriesBox: function () {
+            var series = this,
+                data = series.data,
+                max = Math.max,
+                min = Math.min,
+                radius,
+                // bBox = [xMin, xMax, yMin, yMax]
+                bBox = [Infinity, -Infinity, Infinity, -Infinity];
```
pawelfus: How about we start from `[plotLeft, plotLeft + plotWidth, plotTop, plotTop + plotHeight]` ?

Izothep: good idea, much more clear approach

pawelfus:
6.5s -> 2.1s ? With just a few lines of code? Great job!

js/parts-more/PackedBubbleSeries.js
Izothep: Done with e25a2ab

PR #10739, opened 2019-05-15
bugfix/9198-negative-color-ignored-setData

### LINKED ISSUE #9198: Bug: Negative border on area with incorrect color due chart height
#### Expected behaviour
Negative border on area chart should show in red color (defined by negativeColor prop)

#### Actual behaviour
Negative border on area chart shows in green color, inherited from color for positive area.

http://jsfiddle.net/0srqapLz/

#### Product version
Highstock JS v6.1.4 (2018-09-25)

#### Affected browser(s)
Chrome latest version

sebastianbochan:
Fixed #9198, zones and negative color were not applied after `series.setData`.

### DIFF
--- a/js/parts/Series.js
+++ b/js/parts/Series.js
@@ -4928,15 +4928,19 @@ H.Series = H.seriesType(
                         clips[i].animate(clipAttr);
                     } else {
                         clips[i] = renderer.clipRect(clipAttr);
-
-                        if (graph) {
-                            series['zone-graph-' + i].clip(clips[i]);
-                        }
-
-                        if (area) {
-                            series['zone-area-' + i].clip(clips[i]);
-                        }
                     }
+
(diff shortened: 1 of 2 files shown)


TorsteinHonsi:
Thanks for the PR!

A little comment for the PR description:

> Changelog: Fixed #9198, zones and negative color were not applied after setData.

See [writing content for the changelog](https://github.com/highcharts/highcharts/blob/master/changelog/readme.md#writing-content-for-the-changelog). I accidentally used the term "tag", but I meant "label". I've corrected that. So, the text should not start with "Changelog: ", but rather be _labeled_ with `Changelog: Fixed`. We should also use backticks for code in the description.

No further action required on your part.

sebastianbochan:
Thank you for your suggestions.

PR #10740, opened 2019-05-15
bugfix/10713-respo-annotations-init

### LINKED ISSUE #10713: Initializing annotations in responsive rules doesn't work
Reopening https://github.com/highcharts/highcharts/issues/8680

Annotations don't work with responsive rules

https://jsfiddle.net/BlackLabel/sxw9eb7j
Click **Small** button - annotations are not added.

#### Product version
7.1.1

#### Affected browser(s)
All

pawelfus:
Fixed #10713, initializing annotations in responsive rules didn't work.
___
It is a general issue, where responsive rules don't set `oneToOne` argument in `chart.update()`.

A few things to consider:

- `['series', 'xAxis', 'yAxis', 'annotations'].indexOf(key) > -1` - we may want to replace this array with `chart.collectionsWithUpdate`. In theory, we don't have logic for adding/removing `zAxis`, `colorAxis` and `pane` but one day we may want to add this.

- Maybe this if-else tree 
https://github.com/highcharts/highcharts/blob/f3f7a6e42b1614d19f7b56ec5c4704f8c8b563f5/js/parts/Dynamics.js#L569-L580
could be replaced with (a): 
```
Chart.prototype.initWith = {
  // collectionName: [ 'initializer', [extraArgs]]
  xAxis: ['addAxis', [true]],
  yAxis: [ 'addAxis', [false]],
  series: ['addSeries', []]
}
```
Or (b):
```
Chart.prototype.initWith = {
  // collectionName: [ 'initializer', [extraArgs]]
  xAxis: [Chart.prototype.addAxis, [true]],
  yAxis: [Chart.prototype.addAxis, [false]],
  series: [Chart.prototype.addSeries, []]
}
```
So it will be easily extendable and we could avoid polluting Highcharts core with modules definitions here. We would get simple (for a)):
```
chart[chart.initWith[coll][0]](newOptions, ...chart.initWith[coll][1], false)
```
(I know, rest param syntax can't be used here, but it's just an idea).

### DIFF
--- a/js/annotations/annotations.src.js
+++ b/js/annotations/annotations.src.js
@@ -986,10 +986,10 @@ merge(
         },
 
         /**
-         * See {@link Highcharts.Annotation#destroy}.
+         * See {@link Highcharts.Chart#removeAnnotation}.
          */
         remove: function () {
-            return this.destroy();
+            return this.chart.removeAnnotation(this);
         },
 
         update: function (userOptions) {
--- a/js/parts/Dynamics.js
+++ b/js/parts/Dynamics.js
@@ -386,16 +386,16 @@ extend(Chart.prototype, /** @lends Highcharts.Chart.prototype */ {
      *        Whether to redraw the chart.
      *
      * @param {boolean} [oneToOne=false]
-     *        When `true`, the `series`, `xAxis` and `yAxis` collections will
-     *        be updated one to one, and items will be either added or removed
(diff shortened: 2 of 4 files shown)


TorsteinHonsi:
Thanks for the PR!

I like your two refactoring suggestions @pawelfus . As you say it keeps module-specific code out of the core. So if a PR for that lands in my inbox it will be used.

Issue #5526, opened 2016-07-25
Pie chart looks not good for all zero data


ipip2005:
Hi, we levarage pie chart with authoring experience. So if user make all the values empty, it will looks weird. I mean, not so pro.
#### Expected behaviour

We haven't figured out yet. But probably rendering all zero data like they are equal positive like this:
![image](https://cloud.githubusercontent.com/assets/7711735/17092073/7a476586-5272-11e6-8b8f-21e96ddcd6f5.png)
or an empty circle, means there is no usable data.
It's just weird for me to have no circle in pie chart.
#### Actual behaviour

![image](https://cloud.githubusercontent.com/assets/7711735/17092049/48f95d54-5272-11e6-8459-f5f9670c9c04.png)
#### Live demo with steps to reproduce

http://jsfiddle.net/highcharts/Qgn59/
#### Affected browser(s)

all

TorsteinHonsi:
If we changed it it would have to be optional at least. Many existing users would regard the equal positives a bug. The current display is correct in the sense that the data labels and connectors are there, since 0 is actually a value. In contrast to a pie with [all nulls](http://jsfiddle.net/highcharts/Qgn59/9/). 

Another way of dealing with this could be to add a circle to indicate the outline of the pie when empty: http://jsfiddle.net/highcharts/Qgn59/10/

jon-a-nygaard:
Closing this issue since we consider it not a bug, and it is pending a reply. Should anyone have further information on the issue, please give us a comment.

ipip2005:
It's cool we have such workaround. But it needs overwriting the prototype function,
To not confuse in-production users, maybe we can add this as an option in pie chart.
For example: `showPlaceholderForEmptyPie`? Than the default value will be `false`. In my case, I will set it to true.

dhruv-m-patel:
Maybe not a really great solution, but I am using a donut chart and did this trick:

```
let nonzeroDataPoints = dataPoints;
const dataPointsSum = _.sumBy(dataPoints, data => data.y); // data is an array of datapoints here with y values maybe set to all zeros
if (dataPointsSum === 0) {
   nonzeroDataPoints = [{ y: 1, color: '#DDDDDD', name: 'foo' }];
   tooltip.pointFormat = '<tr><td> 0 </td></tr>';
}

Highcharts.chart('container', {
  chart: {
    ...
    type: 'pie',
    ...
  },
  series: {
    ...
    data: nonzeroDataPoints
    ...
  }
[...]

PR #10757, opened 2019-05-17
enhancement/5526-pie-zero-data-refactored

kamilkulig:
Added empty pie graphic: if total sum of all visible points is 0 then the chart is represented as an empty circle (#5526).


TorsteinHonsi:
Looking good now!

After a quick discussion, we decided to simplify the API a bit. We can get rid of the `emptySeries` options structure be using existing `color` and `fillColor` properties for the circle. So the remaining work is:

* Apply the `color`, simply use `series.color` from the instance. This would be a good default and respond to theming. Usually it will be the same color as the first slice.
* Apply the `fillColor` option. This is currently available in area series, now it gets a new meaning in empty pie charts. It should be `undefined` by default in pies.
* Add doclets for both, explaining when they apply.

TorsteinHonsi:
@kamilkulig Can you make the final changes?

bre1470:
@kamilkulig This should get updated and merged asap, because I start now migrating series to TypeScript.

bre1470:
@TorsteinHonsi Please review changes of handling series options `colors`, `color`, and new `fillColor`.

bre1470:
@TorsteinHonsi `series.supportingColor` is also something new, but seems to be an internal option only.

PR #10759, opened 2019-05-17
bugfix/10669-sunburst-add-ids-verification

### LINKED ISSUE #10669: Sunburst - chart is changed after update with initial options
#### Expected behaviour

The chart should look the same as before the update. Highcharts v 7.0.3: http://jsfiddle.net/BlackLabel/v8nht4c7/

#### Actual behaviour

The chart is changed: http://jsfiddle.net/BlackLabel/gLs4y39u/

#### Product version
Highcharts v 7.1.1

#### Affected browser(s)
All

sebastianbochan:
Fixed #10669, sunburst didn´t show error when duplicate point id's were used.

### DIFF
--- a/js/modules/sunburst.src.js
+++ b/js/modules/sunburst.src.js
@@ -887,7 +887,7 @@ var sunburstSeries = {
             nodeTop,
             tree,
             values,
-            nodeIds = [];
+            nodeIds = {};
 
         series.shapeRoot = nodeRoot && nodeRoot.shapeArgs;
         // Call prototype function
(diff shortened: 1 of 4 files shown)


sebastianbochan:
Thank you Torstein, now it looks like ready to merge, or should I change something more?

js/modules/sunburst.src.js
```diff
@@ -941,6 +942,18 @@ var sunburstSeries = {
         this.setShapeArgs(nodeTop, values, mapOptionsToLevel);
         // Set mapOptionsToLevel on series for use in drawPoints.
         series.mapOptionsToLevel = mapOptionsToLevel;
+
+        // #10669 - verify if all nodes have unique ids
+        series.data.forEach(function (child) {
+            if (nodeIds.indexOf(child.id) > -1) {
```
jon-a-nygaard: Would it be faster or maybe simpler to use a map instead of the array and indexOf? E.g:
```javascript
if (nodeIds[child.id]) {
// throw error
}
nodeIds[child.id] = true;
```

Issue #8912, opened 2018-09-10
Stacked labels of first and last column are getting out of chart or truncated


ivan-nejezchleb:
#### Expected behaviour
There should be same crop+overflow settings for stacked labels same as for data labels and stack labels should behave according it

#### Actual behaviour
Stack labels cant be positioned and are getting out of chart. Crop and overflow options don't work

#### Live demo with steps to reproduce
http://jsfiddle.net/gq4fahnk/2/

#### Product version
latest

pawelfus:
Hi @ivan-nejezchleb 

Thank you for your suggestion! I think you are right, this would be useful to share crop&overflow feature between dataLabels and stackLabels.

alexeykuznetsov:
When?

sebastianbochan:
Hi @alexeykuznetsov,
Unfortunately, at this moment we do not have ETA for this case, yet. But Im preparing you a really simple snippet, which will overflow the first and the last label. Will post today or tomorrow morning.

balkarov:
Hello. When you fixed the bug?

balkarov:
@sebastianbochan hello!
I tried your workaround, but it did not help me.
https://jsfiddle.net/nt5oh2pw/

![image](https://user-images.githubusercontent.com/6030335/56951379-59c9ff00-6b40-11e9-84d7-5671fd10db0e.png)

pawelfus:
The workaround above is designed to work with `column` series, for `bar` series it's a bit different: https://jsfiddle.net/BlackLabel/yocf798u/1/

balkarov:
@pawelfus but why do you send some workaround? Why can't you add this fix to the library?
The bug from 2018 September.
We pay for the library and now our project has 7 workarounds for Highcharts already. It is not normally.

balkarov:
@pawelfus when you fix this problem in library?

pawelfus:
Hi @balkarov 

No ETA for this feature at this moment, but we will take a look this week.

balkarov:
@pawelfus hello. What about ETA?

Basts:
@balkarov I am going fix it this week.

sebastianbochan:
Hi @balkarov,
Me and @Basts just prepared a solution which solve the issue. All our tests look like promising, but we would like to ensure that all will work also in your environment. 

**Demo:**
- http://jsfiddle.net/BlackLabel/q8ocrw10/5/

Any feedback from you will be really helpful.

[...]

PR #10844, opened 2019-05-29
enhancement/8912-stacked-labels-crop-overflow

Basts:
Added `overflow` and `crop` options to stack labels. See #8912.


Basts:
@TorsteinHonsi 

I considered using **justifyDataLabel**, but **dataLabels** are rendered in a different way than **stackLabels** 
- main differences: **dataLabels** has additional features, which has an impact for the position, like rotation and **padding**, **stackLabels** "finds" their positions via **getStackBox** function, while **dataLabels** does it via **alignTo** property. 

Using a justifyDataLabel function for both cases requires:
1. adding padding feature into **stackLabels**,
2. **justifyDataLabel** function is justified dataLabel via: 
`    if (justified) {
        dataLabel.placed = !isNew;
        dataLabel.align(options, null, alignTo);
    }
 `
I haven't tested how it will be aligining for **getStackBox** rather than **alignTo**, but as I did investigation their **alignAttr.x** is totally different, so I assumed that some additional refactor will be needed.

TorsteinHonsi:
Hmm.. I think it's unfortunate that we have two different code paths for so similar features. I imagine that the positioning logic should be identical for a _stack_ and a _column_ in a column series. The `getStackBox` would be equivalent to the column. Can you do some more research and see if we can piggy-back the column datalabel positioning logic more?

Basts:
@TorsteinHonsi 
Ok, then I will try to refactor `stackLabels` to match them to `justifyDataLabel` function.

What about padding feature? Should it be implemented into stackLabels or set to 0 by default? This value is needed to find a proper position via `justifyDataLabel` function.

TorsteinHonsi:
> What about padding feature? 

As a minimum viable solution, it is probably ok to set it to 0 for now.

bre1470:
@basts You have to resolve the conflict by applying the changes to the `ts/parts/Stacking.ts` file.

TorsteinHonsi:
Fixed in another PR.

PR #10853, opened 2019-05-31
Tools: Added --splitbrowsers as argument when running karma

activist:
in split/sharded mode. Using --browsers as cli argument will otherwise override the config.

Rename --split argument to --browsercount.
Enabled Win.Edge on CI again.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -194,7 +194,7 @@ workflows:
           requires:
             - build_dependencies
           browsers: "Win.Edge"
-          browsercount: 2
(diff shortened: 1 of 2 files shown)


activist:
@TorsteinHonsi  This fails on Edge due to a failing test now, so that is progress. 

At the same time the disconnect issue was not solved by changing to Edge insider preview (Chromium) alone. Using polling as transport policy when running tests for Edge seems to also help although I guess we need to run more builds to know if this is a permanent solution or not.

PR #10860, opened 2019-06-03
Allow bottom alignment of subtitle

TorsteinHonsi:
Currently the chart doesn't reserve space for the subtitle if it is bottom aligned. If this is allowed, it can be used for general chart descriptions/captions/narratives.

Corresponding Cloud issue https://github.com/highcharts/highcharts-cloud/issues/485

#### Work in progress
http://jsfiddle.net/highcharts/0qkudhow/

#### To do
 - [x] Align the legend above the subtitle.
 - [x] Adjust the vertical alignment of the subtitle, it looks like it is one line-height too high. Probably currently aligned to the first text baseline.
 - [ ] Investigate possible refactoring of `chart.titleOffset` and `chart.titleOffsetBottom` into `chart.titleOffset = [number, undefined, number]`, so that we can reference it by `axis.side`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -187,7 +187,7 @@ workflows:
       - test_browsers:
           name: "test-Mac.Firefox"
           requires:
-            - build_dependencies
+            - "test-Mac.Safari"
           browsers: "Mac.Firefox"
       - hold_for_approval:
           type: approval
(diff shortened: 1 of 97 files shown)


TorsteinHonsi:
Closing due to messed up commit history. Creating new PR.

PR #10885, opened 2019-06-05
Fixed #10708, remove use of findIndex

### LINKED ISSUE #10708: Use of Array.prototype.findIndex breaks on IE
[`Array.prototype.findIndex`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex) is used at https://github.com/highcharts/highcharts/blob/de7aff6930f4f58bf543db476863649e4aa5c899/js/modules/sankey.src.js#L472
According to MDN this is not supported in any IE browser and will break.

Can be replaced by `Array.prototype.find` and a local index variable.
#### Affected browser(s)
Internet Explorer

jon-a-nygaard:
Fixed #10708, removed use of `Array.prototype.findIndex` which would cause Sankey series to break in IE.

---
# Description
Unfortunately this replacement does not look pretty code-wise, but I am unable to see an elegant solution for this issue.

# Related issue(s)
- Closes #10708

### DIFF
--- a/js/modules/sankey.src.js
+++ b/js/modules/sankey.src.js
@@ -469,11 +469,17 @@ seriesType('sankey', 'column',
                         // Hanging layout for organization chart
                         if (fromNode.options.layout === 'hanging') {
                             node.hangsFrom = fromNode;
-                            node.column += fromNode.linksFrom.findIndex(
-                                function (link) {
-                                    return link.toNode === node;
+                            i = -1; // Reuse existing variable i
+                            fromNode.linksFrom.find(
+                                function (link, index) {
+                                    var found = link.toNode === node;
+                                    if (found) {
+                                        i = index;
+                                    }
+                                    return found;
                                 }
                             );
+                            node.column += i;
                         }
                     }
                 }
--- a/js/modules/sankey.src.js
+++ b/js/modules/sankey.src.js
@@ -114,6 +114,7 @@ var getLevelOptions = mixinTreeSeries.getLevelOptions;
 
 
 var defined = H.defined,
+    find = H.find,
     isObject = H.isObject,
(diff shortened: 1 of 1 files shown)


js/modules/sankey.src.js
```diff
@@ -469,11 +469,17 @@ seriesType('sankey', 'column',
                         // Hanging layout for organization chart
                         if (fromNode.options.layout === 'hanging') {
                             node.hangsFrom = fromNode;
-                            node.column += fromNode.linksFrom.findIndex(
-                                function (link) {
-                                    return link.toNode === node;
+                            i = -1; // Reuse existing variable i
+                            fromNode.linksFrom.find(
```
TorsteinHonsi: Unfortunately `Array.prototype.find` is also [not available in IE11](https://caniuse.com/#feat=array-find) :(

The fastest code would be a good old `for` loop with a `break`, so we don't have to continue iterating once the item is found. Or a `while`.

jon-a-nygaard: > Unfortunately Array.prototype.find is also not available in IE11 :(

Nooo, I did not think of that. Haha, this was ironic 😆 

> The fastest code would be a good old for loop with a break, so we don't have to continue iterating once the item is found. Or a while.

`find` stops iterating once it matches a value. Could I replace `Array.prototype.find` with `Highcharts.find`, or is still a `for` loop preferred?

TorsteinHonsi: Ok, yes of course `find` is okay.

jon-a-nygaard: I updated the branch with this change now, just waiting for CI to make final confirmation.

PR #10895, opened 2019-06-05
Fixed #10788, wrong docs for stock tools.

### LINKED ISSUE #10788: Stock Tools Custom GUI Demo not showing custom Dialogs/Popups
#### Expected behaviour

I'm not quite sure if this expectation does actually reflect the intention of the demo but when I started diving into the demo's codes I struggled to understand what's happening:
The [Stock Tools Custom GUI Demo](https://www.highcharts.com/stock/demo/stock-tools-custom-gui) should show a popup containing the two fields "Color" and "Width" as defined in the demo source codes.

At the end I need an example which shows how to integrate some sort of form where I can add some additional annotation metadata for an annotation plus some way to export the defined markers/annotations details - I thought the mentioned demo would be a good starting point - but as I said I'm struggling to get my head around what happens there.

#### Actual behaviour

The [Stock Tools Custom GUI Demo](https://www.highcharts.com/stock/demo/stock-tools-custom-gui) has defined a custom Popup in the Markup which also reflects the pattern described in the stock tools docs - but when invoking the popup it's not showing the custom popup but some default thing from which I can't figure where it's coming from.

https://github.com/highcharts/highcharts/blob/76b92282ff80a08e355ad4f532b88de94fee5508/samples/stock/demo/stock-tools-custom-gui/demo.html#L33-L42

* Go to the Demo page https://www.highcharts.com/stock/demo/stock-tools-custom-gui
* Add a Text annotation
* Click on that annotation and hit the edit icon
* Review the form

You can even monitor the DOM to see that it's not the element `.highcharts-popup.highcharts-popup-annotations` where the `display` is toggled.

RkypGZ6cK4

#### Product version

highcharts 7.1.1 plus highstock

#### Affected browser(s)

Chrome 73.0.3683.103
Firefox 66.0.3

pawelfus:
Fixed #10788, samples in stock-tools documentation had minor bugs.
___

In general, a few minor fixes:
- reviewed docs, fixed minor things there too
- moved demos from Gert's jsfiddle account to official samples + added description for manual testing
- added docs file for stock tools. Created `stock/` dir for that. I'm not sure about convention here (e.g. all series are placed in the main dir), but I guess something similar will be implemented anyway

Let me know if something else is needed too. It's my first PR for website docs.

### DIFF
--- a/docs/stock/stock-tools.md
+++ b/docs/stock/stock-tools.md
@@ -1,0 +1,203 @@
+# Stock Tools
+
+Stock Tools is a Highstock module for building a GUI that enables user interaction with the chart, such as adding annotations, technical indicators or just for zooming in or out. The module is released with Highstock 7 and won't work with previous versions.
+
+Building the GUI is facilitated by the Stock Tools module, where your HTML elements for user interaction are automatically bound to predefined events in this module. Full list of the available bindings can be found in the API [navigation.bindings](https://api.highcharts.com/highstock/navigation.bindings).
+
+## QUICK START
+To get started quickly it's recommended to use the default toolbar of the Stock Tools module. We need to load the JavaScript files in the following order.
+
+1. First load Highstock and the bundle with all the technical indicators for convenience.
+```
+<script src="http://code.highcharts.com/indicators/indicators-all.js"></script>
+```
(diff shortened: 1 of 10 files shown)


pawelfus:
PS:
Shared samples ( e.g. https://highcharts.github.io/highcharts-utils/samples/#gh/3e96bd230b/sample/stock/stocktools/basic-gui ) don't work because of CORS: 

> Access to XMLHttpRequest at 'https://www.highcharts.com/samples/data/aapl-c.json' from origin 'https://highcharts.github.io' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.

TorsteinHonsi:
FYI @pawelfus Updated samples/docs don't go in the changelog, only actual product/API changes do.

FYI @cvasseng This PR adds content to the pre-production `/docs` folder.

PR #10906, opened 2019-06-06
enhancement/plot-lines-radial-axis

wchmiel:
Added feature to draw plot lines on radial axes from the inner to the outer radius of the main pane.


js/parts-more/RadialAxis.js
```diff
@@ -415,20 +415,44 @@ radialAxisMixin = {
...
+            a,
+            b,
             xAxis,
             xy,
             tickPositions,
             ret;
 
         // Spokes
         if (axis.isCircular) {
+            a = (typeof innerRadius === 'string') ?
```
TorsteinHonsi: Use `H.relativeLength` to handle the percentages.

PR #10975, opened 2019-06-12
bugfix/10930-datalabels-visibility-toggle

### LINKED ISSUE #10930: data labels are missing when you zoom
#### Expected behaviour
data labels are shown as expected when you zoom

#### Actual behaviour
data labels are missing when you zoom, even though there is plenty of room for them

[live demo of issue](https://jsfiddle.net/kdragon/gd38rf7c/11/)

hit "ig", then hit "hg", u will notice that several points are not labeled that should be 

if u zoom to "hg" originally, it will show all the data labels, as it should

#### Product version
highcharts 7.1.2

#### Affected browser(s)
all

pawelfus:
Fixed #10930, sometimes data labels were not visible after zooming.
___

TS note (@bre):
I might be missing something, I have this error:

> ts/parts/DataLabels.ts(1356,24): error TS2554: Expected 0 arguments, but got 1.

But I have added this: https://github.com/highcharts/highcharts/compare/bugfix/10930-datalabels-visibility-toggle?expand=1#diff-eaca87a6f0fc86b8d82591529cab6728R2452

Should I run some other task, not only `npx gulp scripts-ts` ? Did I make some silly mistake perhaps? :)

@TorsteinHonsi 
One of the issues that can be resolved in many ways. Alternative solutions (and why I decided to choose this one):

- overwrite `dataLabels.hide = function () { ... };` instead of polluting `hide(..args)`, but that would require creating function for every label (prototyped function vs assigned function) - performance issue (I don't think we want to decrease performance for dataLabels any bit more). Additionally this may cause issues with extending library/workarounding issues (no way to use `wrap()` on SVGElement, as later it will be overwritten).

- replace `label.hide()`/`label.show()` in overlapping's algorithm with `attr({ y: -9999 })`, again no polluting `hide(..args)` but it's easy to forget that `dataLabel.show()`/`dataLabel.hide()` shouldn't be used - maintenance issue.

Pros of the PR'ed solution - I have fixed some minor issues with `hide(true)` that were called with an argument indirectly, e.g. `axisLine[showAxis ? 'show' : 'hide'](true);` - `hide(true)` used to make no sense. Note that TS didn't comply about extra arg in this particular case.

### DIFF
--- a/js/parts/SvgRenderer.js
+++ b/js/parts/SvgRenderer.js
@@ -1571,16 +1571,26 @@ extend(SVGElement.prototype, /** @lends Highcharts.SVGElement.prototype */ {
         return this.attr({ visibility: inherit ? 'inherit' : 'visible' });
     },
     /**
-     * Hide the element, equivalent to setting the `visibility` attribute to
+     * Hide the element, similar to setting the `visibility` attribute to
      * `hidden`.
      *
      * @function Highcharts.SVGElement#hide
      *
+     * @param {boolean} [hideByTranslation=false]
+     *        The flag to determine if element should be hidden by moving out
+     *        of the viewport. Used for example for dataLabels.
+     *
      * @return {Highcharts.SVGElement}
      *         Returns the SVGElement for chaining.
      */
-    hide: function () {
-        return this.attr({ visibility: 'hidden' });
+    hide: function (hideByTranslation) {
+        if (hideByTranslation) {
+            this.attr({ y: -9999 });
+        }
+        else {
+            this.attr({ visibility: 'hidden' });
+        }
+        return this;
     },
     /**
      * Fade out an element by animating its opacity down to 0, and hide it on
--- a/ts/parts/SvgRenderer.ts
+++ b/ts/parts/SvgRenderer.ts
@@ -200,7 +200,7 @@ declare global {
             public getBBox(reload?: boolean, rot?: number): BBoxObject;
             public getStyle(prop: string): string;
             public hasClass(className: string): boolean;
-            public hide(): SVGElement;
+            public hide(hideByTranslation?: boolean): SVGElement;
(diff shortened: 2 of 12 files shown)


js/parts/SvgRenderer.js
```diff
@@ -1571,16 +1571,26 @@ extend(SVGElement.prototype, /** @lends Highcharts.SVGElement.prototype */ {
...
+     * @param {boolean} [hideByTranslation=false]
+     *        The flag to determine if element should be hidden by moving out
+     *        of the viewport. Used for example for dataLabels.
+     *
      * @return {Highcharts.SVGElement}
      *         Returns the SVGElement for chaining.
      */
-    hide: function () {
-        return this.attr({ visibility: 'hidden' });
+    hide: function (hideByTranslation) {
```
bre1470: You have to add the new parameter at the top of the file to the internal types as well. Otherwise TypeScript will not successful compile. This is necessary during the migration process.

ts/parts/SvgRenderer.ts
```diff
@@ -2435,16 +2435,30 @@ extend((
...
+     *        of the viewport. Used for example for dataLabels.
+     *
      * @return {Highcharts.SVGElement}
      *         Returns the SVGElement for chaining.
      */
-    hide: function (this: Highcharts.SVGElement): Highcharts.SVGElement {
-        return this.attr({ visibility: 'hidden' }) as any;
+    hide: function (
+        this: Highcharts.SVGElement,
+        hideByTranslation?: boolean
```
bre1470: You have to add the new parameter at the top of the file to the internal types as well. Otherwise TypeScript will not successful compile. This is necessary during the migration process.

pawelfus:
Changes to TS - done. Thank you @bre1470 !

Issue #10510, opened 2019-04-08
Charts with pattern fill create elements with duplicate ids


bvancea:
#### Expected behavior
In a page, render multiple charts, each with pattern fill. The HTML document should not contain elements with duplicate id attributes, in order to comply with [WCAG 4.1.1](https://www.w3.org/WAI/WCAG21/Understanding/parsing.html). The duplicate ids could cause issues with assisstive technologies.

#### Actual behavior
Highcharts creates multiple `<pattern>` elements with the same id.

#### Live demo with steps to reproduce
1. Access https://codepen.io/anon/pen/zXKQrV
2. Execute the following command in the browser console: `document.querySelectorAll("#highcharts-default-pattern-0")`

#### Product version
Highcharts 7.1, but the issue appears in older versions as well.

#### Affected browser(s)
All browsers.

pawelfus:
Hi @bvancea 

Thanks for reporting this bug! 

FYI @oysteinmoseng

Issue #10594, opened 2019-04-22
Cannot read property 'defIds' of undefined


brandynbennett:
#### Expected behaviour
If the chart is destroyed mid-animation an error will not be thrown from [pattern-fill](https://github.com/highcharts/highcharts/blob/master/js/modules/pattern-fill.src.js#L512)
Error `Uncaught TypeError: Cannot read property 'defIds' of undefined` is not thrown from [pattern-fill](https://github.com/highcharts/highcharts/blob/master/js/modules/pattern-fill.src.js#L512)

#### Actual behaviour
`Uncaught TypeError: Cannot read property 'defIds' of undefined` is thrown after chart is destroyed mid-animation.

#### Live demo with steps to reproduce

I'm struggling to reproduce in isolated environment. Somehow you have to destroy the chart while the animation is running.
Likely doing a check to see if `this.renderer` exists would fix it

#### Product version
Highcharts 7.0.3 && 7.1.1

#### Affected browser(s)
Evergreen

sebastianbochan:
Hi @brandynbennett,
Thank you for reporting about the problem.

FYI @oysteinmoseng

Blapi:
Hi, I'm experiencing the same issue, I've got multiple times this error while using highcharts

```
Uncaught TypeError: Cannot read property 'defIds' of undefined
    at a.Chart.<anonymous> (pattern-fill.js:164)
    at highcharts.js:645
    at a.fireEvent (highcharts.js:647)
    at highcharts.js:6119
```

highcharts-react-official version "2.1.3" (using highcharts version "7.1.2") on Chrome 74.0

Any updates/hints ?

oysteinmoseng:
I will investigate this week, it is probably an easy fix.

PR #10997, opened 2019-06-14
Bugfix/pattern fill issues

oysteinmoseng:
Fixed issues #10510 and #10594 with pattern fills.
___
Added chart index to generated pattern fill IDs, and tested for renderer existing in endResize to avoid crash.

### DIFF
--- a/js/modules/pattern-fill.src.js
+++ b/js/modules/pattern-fill.src.js
@@ -510,7 +510,7 @@ H.addEvent(H.SVGRenderer, 'complexColor', function (args) {
 // resize, as the bounding boxes are not available until then.
 H.addEvent(H.Chart, 'endResize', function () {
     if (
-        (this.renderer.defIds || []).filter(function (id) {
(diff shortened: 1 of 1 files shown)


js/modules/pattern-fill.src.js
```diff
@@ -597,7 +599,7 @@ H.Chart.prototype.callbacks.push(function (chart) {
         'M 0 0 L 5 10 L 10 0'
     ].forEach(function (pattern, i) {
         chart.renderer.addPattern({
-            id: 'highcharts-default-pattern-' + i,
+            id: 'highcharts-default-pattern-' + (index ? index + '-' : '') + i,
```
jon-a-nygaard: Can index equal `0` in this case? If so then you should probably test for `isNumber(index) ? : ;` instead.

oysteinmoseng: It can, but on purpose we don't want to append the index to the ID in those cases. This way we won't break backwards compatibility

PR #10998, opened 2019-06-14
Tools: Migrated S3 upload and compression from ant 

activist:
Migration of s3 upload and compression from ant to js.

Currently for testing.


activist:
Deploy / upload to s3 (currently a staging bucket) will now be triggered when a vX.X.X tag is pushed.

tools/gulptasks/dist-upload.js
```diff
@@ -0,0 +1,53 @@
...
+ * @param {object} productProps containing product name and version (build.properties).
+ * @return {Promise<any> | Promise | Promise} Promise to keep
+ */
+function uploadPackage(productName, productProps) {
+    const { name, version, cdnpath } = productProps;
+    return uploadProductPackage(productName, cdnpath, name, version);
+}
+
+
+const distUpload = products => {
```
bre1470: I favor real functions with own scope, because then call & order come not into play, as functions get always parsed before everything else.

tools/gulptasks/unsorted/uploadS3.js
```diff
@@ -0,0 +1,240 @@
+const semver = require('semver');
```
bre1470: If it is used only as a lib, please move it to `tools/gulptasks/lib`.

tools/gulptasks/dist-upload.js
```diff
@@ -0,0 +1,53 @@
...
+    promises.push(uploadFiles({ files: productJs, name: 'products.js' }));
+    return Promise.all(promises);
+
+    // TODO: ant push-all-versions-to-S3: push whole versions structure from local machine to Amazon S3
+    // TODO: ant copy-to-site: Copy files from this repo to repository www.highcharts.com (joomla?)
+};
+
+
+gulp.task('dist-upload', () => distUpload(['highcharts', 'highstock', 'highmaps', 'gantt']));
+gulp.task('dist-upload-highcharts', () => distUpload(['highcharts']));
```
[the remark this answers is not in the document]
activist: Fixed as you specified :)

tools/gulptasks/dist-upload.js
activist: Fixed. I wasn't aware of this. Thanks :+1:

tools/gulptasks/unsorted/uploadS3.js
activist: Fixed.

tools/gulptasks/dist-upload.js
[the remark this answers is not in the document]
activist: @bre1470 Regarding testing the bucket can be specified by using the `git-ignore-me.properties`file as used by ant earlier or you can specify config in env vars:
```
export HIGHCHARTS_S3_BUCKET="staging-code.highcharts.com"
export AWS_REGION="eu-central-1"
export AWS_ACCESS_KEY_ID="<key>"
export AWS_SECRET_ACCESS_KEY="<secret>"
```
The credentials can also be set by installing & configuring aws-cli locally, but if both are in use env vars will have precedence.

tools/gulptasks/dist-compress.js
```diff
@@ -0,0 +1,103 @@
...
+    });
+
+    return Promise.all(zipTasks);
+}
+
+/**
+ * Creates gzipped version of files in build/dist/{productname}/code and stores them
+ * in build/dist/{productname}/js-gzip. E.g. build/dist/highcharts/code --> gzip and store in build/dist/highcharts/code/js-gzip
+ *
+ * @return {Promise<any> | Promise | Promise} Promise to keep
```
bre1470: I do not understand the type here. If it gets always solved to the last `then` without returning something inside `then`, it should be a simple `Promise<void>`.

tools/gulptasks/dist-compress.js
```diff
@@ -0,0 +1,103 @@
...
+        return new Promise((resolve, reject) => {
+            gulp.src([dirToZip + '/**', excludedDir])
+                .pipe(zip(zipFileName))
+                .pipe(gulp.dest(DIST_DIR))
+                .on('error', reject)
+                .on('end', resolve);
+        });
+    });
+
+    return Promise.all(zipTasks);
```
bre1470: If you add here a `.then(() => undefined);`, you can change the return type to `Promise<void>`.

tools/gulptasks/dist-compress.js
```diff
@@ -0,0 +1,103 @@
...
+const util = require('util');
+
+
+const DIST_DIR = 'build/dist';
+
+
+/**
+ * Creates zip files for highcharts, highstock, highmaps, highcharts-gantt.
+ *
+ * @return {Promise<any> | Promise | Promise} Promise to keep
```
bre1470: Confusing type

tools/gulptasks/dist-upload.js
```diff
@@ -0,0 +1,52 @@
...
+function uploadPackage(productName, productProps) {
+    const { name, version, cdnpath } = productProps;
+    return uploadProductPackage(productName, cdnpath, name, version);
+}
+
+/**
+ * Upload a distribution version for all products.
+ *
+ * @param {Array<string>} products (product names) to initiate an upload for.
+ * @return {Promise<any> | Promise | Promise} Promise to keep
```
bre1470: Here you could do the same approach and add a `.then(() => undefined)` to the returning `Promise.all` to cast it to a simple `Promise<void>`. Gulp does not care about resolve results - only reject errors are handled (for now).

bre1470:
If you are not sure about a type, you can use `Promise<*>` for now. We then would later cover it, when migrating gulp tasks to TypeScript (if so).

activist:
I took the easy way out and did as you suggested in your last comment. If there is more to comment, please do :)

tools/gulptasks/dist-compress.js
activist: Fixed by removing the last then() as it is not really needed. And changed return type to Promise<*> for now.

tools/gulptasks/dist-compress.js
activist: Changed return type to Promise<*> for now.

tools/gulptasks/dist-upload.js
activist: Changed return type to Promise<*> for now.

PR #11139, opened 2019-06-24
bugfix/11124-networkgraph-draggable-delay

### LINKED ISSUE #11124: Force-Directed  Network Graph in Highcharts: Each time a node is clicked, its position moves.
#### Expected behaviour
After loading, there is no longer any position movement of any element in the graph. When clicking any node (regardless of whether the node is bound with the click event or not), the node position is not moved.

#### Actual behaviour
Change the color of the node by clicking on the event, and use chart.update () to update. We have tried, no matter whether the click event is added here or whether there is Chart. Update () in the click event, as long as the node is clicked, the node will move its position.

https://jshare.com.cn/temp/hOtWfs

#### Product version
"highcharts": "^7.1.1"

#### Affected browser(s)
Chrome,
Not tried in other browsers.

pawelfus:
Fixed #11124, click on a node in networkgraph caused all nodes to reposition.

### DIFF
--- a/js/modules/networkgraph/draggable-nodes.js
+++ b/js/modules/networkgraph/draggable-nodes.js
@@ -62,6 +62,7 @@ H.dragNodesMixin = {
                 if (chart.isInsidePlot(newPlotX, newPlotY)) {
                     point.plotX = newPlotX;
                     point.plotY = newPlotY;
+                    point.hasDragged = true;
 
                     this.redrawHalo(point);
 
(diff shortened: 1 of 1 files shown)


leostar-eva:
@pawelfus Thanks for the answer to my question, but at the same time, I found another problem: when 'chart.update()' exists in the click event of the node to update the node, the node will still move. For example, we need to change the node's 'lineColor', 'lineWidth', 'radius', and update the node through' chart.update() '. Like this:

`point: {`
          ` events: {`
            `click: function (e) {`
             ` this.update({`
				  `marker: {`
					`...this.marker,`
					`...{`
					  `lineColor: this.commonColor // '#f73cb2'`
					`}`
				  `}`
				`})`
            `}`
         ` }`
       ` }`

So I wonder if I can add some judgment in 'chart.update ()', if the node only changes some appearance properties, and does not change the number of nodes and the relationship between nodes, there is no need to recalculate the position. Or you can update the appearance properties of the node in some other way. Very urgent. Thank you.

PR #11151, opened 2019-06-24
bugfix/6657-contrast-column-datalabels

### LINKED ISSUE #6657: Wrong color of first and last datalabel [5.0.10]
#### Expected behaviour
Display all datalabels in dark colour.
screen shot 2017-04-28 at 15 36 50

#### Actual behaviour
First and last datalabels are coulored with white
screen shot 2017-04-28 at 15 37 40

https://jsfiddle.net/c1rcxn0r/ [5.0.10] - Issue
https://jsfiddle.net/c1rcxn0r/1 [5.0.9] - Works OK

pawelfus:
Fixed #6657, inappropriate contrast color for first and last dataLabel in column/bar series.
___
Simply moved `intersectRect` to Utils, I think this function is quite handy.
Let me know if TS definitions should look different (personally, I think it's more readable this way).

### DIFF
--- a/js/parts/Utilities.js
+++ b/js/parts/Utilities.js
@@ -1355,6 +1355,37 @@ H.getMagnitude = function (num) {
     return Math.pow(10, Math.floor(Math.log(num) / Math.LN10));
 };
 /**
+ * Get the magnitude of a number.
+ *
+ * @function Highcharts.intersectRect
+ *
+ * @param {number} x1
+ *        x-position of the first box
+ * @param {number} y1
+ *        y-position of the first box
+ * @param {number} w1
+ *        width of the first box
+ * @param {number} h1
+ *        height of the first box
+ * @param {number} x2
+ *        x-position of the second box
+ * @param {number} y2
+ *        y-position of the second box
+ * @param {number} w2
+ *        width of the second box
+ * @param {number} h2
+ *        height of the second box
+ *
+ * @return {boolean}
+ *         Boolean wheater rects are intersecting.
+ */
+H.intersectRect = function (x1, y1, w1, h1, x2, y2, w2, h2) {
+    return !(x2 > x1 + w1 ||
+        x2 + w2 < x1 ||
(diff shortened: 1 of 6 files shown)


js/parts/Utilities.js
```diff
@@ -1354,6 +1354,37 @@ H.format = function (str, ctx, time) {
 H.getMagnitude = function (num) {
     return Math.pow(10, Math.floor(Math.log(num) / Math.LN10));
 };
+/**
+ * Get the magnitude of a number.
```
TorsteinHonsi: Description is non-descriptive

js/parts/Utilities.js
```diff
@@ -1354,6 +1354,37 @@ H.format = function (str, ctx, time) {
...
+ *        x-position of the second box
+ * @param {number} y2
+ *        y-position of the second box
+ * @param {number} w2
+ *        width of the second box
+ * @param {number} h2
+ *        height of the second box
+ *
+ * @return {boolean}
+ *         Boolean wheater rects are intersecting.
```
TorsteinHonsi: Typo: whether

ts/parts/DataLabels.ts
```diff
@@ -2180,7 +2181,20 @@ if (seriesTypes.column) {
         );
 
         // If label was justified and we have contrast, set it:
-        if (point.isLabelJustified && point.contrastColor) {
+        if (
+            point.contrastColor &&
+            point.shapeArgs &&
+            intersectRect(
```
TorsteinHonsi: It's probably better to refactor the function to take two objects, each having `{ x, y, width, height }`, so we don't have to pass over eight nested parameters.

bre1470: Then it is actually better to have two parameter, because we have interfaces like `PositionObject` and `DimensionObject`.

bre1470: Even better, we have the `BBoxObject` interface, that takes all four properties.

PR #11152, opened 2019-06-24
bugfix/11146-ts-TooltipFormatterCallbackFunction

### LINKED ISSUE #11146: TooltipFormatterCallbackFunction is incompletely typed
#### Expected behaviour
`defaultFormatter` should be a property of `Tooltip`.
`tooltip` should be a parameter in `TooltipFormatterCallbackFunction`.

#### Actual behaviour
Example [Extending default formatter](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/tooltip/formatter-conditional-default/) will cause type errors in typescript.

#### Product version
7.1.2

pawelfus:
Fixed #11146, added Tooltip argument to TooltipFormatterCallbackFunction interface.

PS: Let me know if this should be somehow labeled (added just product: TS). Should this be placed in changelog too?

### DIFF
--- a/ts/parts/Tooltip.ts
+++ b/ts/parts/Tooltip.ts
@@ -143,6 +143,9 @@ declare global {
  *
  * @param {Highcharts.TooltipFormatterContextObject} this
  *        Context to format
+
+ * @param {Highcharts.Tooltip} tooltip
+ *        The tooltip instance
  *
  * @return {false|string|Array<string>}
  *         Formatted text or false
--- a/ts/parts/Tooltip.ts
+++ b/ts/parts/Tooltip.ts
@@ -1383,7 +1383,7 @@ H.Tooltip.prototype = {
         // Set the renderer size dynamically to prevent document size to change
         if (this.outside) {
             pad = (this.options.borderWidth || 0) + 2 * this.distance;
-            this.renderer.setSize(
+            (this.renderer as any).setSize(
                 label.width + pad,
                 label.height + pad,
                 false
--- a/ts/parts/Tooltip.ts
+++ b/ts/parts/Tooltip.ts
@@ -25,7 +25,7 @@ declare global {
         interface TooltipFormatterCallbackFunction {
             (
                 this: TooltipFormatterContextObject,
-                tooltip?: Tooltip
+                tooltip: Tooltip
             ): (false|string|Array<string>);
         }
         interface TooltipFormatterContextObject {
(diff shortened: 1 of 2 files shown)


ts/parts/Tooltip.ts
```diff
@@ -23,7 +23,10 @@ declare global {
             tt?: SVGElement;
         }
         interface TooltipFormatterCallbackFunction {
-            (this: TooltipFormatterContextObject): (false|string|Array<string>);
+            (
+                this: TooltipFormatterContextObject,
+                tooltip?: Tooltip
```
bre1470: This new parameter requires a change in the corresponding doclet for `Highcharts.TooltipFormatterCallbackFunction` as well.

pawelfus: Thanks! Added, let me know if anything else

pawelfus: Now I think.. `tooltip?: tooltip` or perhaps `tooltip: tooltip` ?

bre1470: I am unsure (would tend to optional) and actually I see a broader problem. We do not have Tooltip as an official API class. So you are referencing something, that is not documented.

TorsteinHonsi:
Awaiting @bre1470 's approval

TorsteinHonsi:
I suggest the following `Tooltip` properties are exposed:

* chart
* container
* options
* outside (readonly)
* renderer
* shared (readonly)
* split (readonly)
* destroy
* getLabel
* hide
* refresh
* update

Those should do for a start.

pawelfus:
I would add `defaultFormatter` too (as reported in #11146).

pawelfus:
> The correct solution would be to actually to check before this, if the renderer property is set. Then you do not have to cast to any.

I agree. However, in theory `this.outside` is responsible for creating `this.renderer` (and `this.container`). If `this.outside` happens without `this.renderer` then something went wrong somewhere else and should be fixed, not worked around.

bre1470:
> I agree. However, in theory `this.outside` is responsible for creating `this.renderer` (and `this.container`). If `this.outside` happens without `this.renderer` then something went wrong somewhere else and should be fixed, not worked around.

TypeScript does not look that deep into the code and instead is expecting every function call to be async, what in practically happens all the time in the browser thanks to the DOM event system. So if you check, you make the function atomic solid.

Issue #5075, opened 2016-02-24
Data labels with useHTML - display is not deferred


KacperMadej:
**Problem:** dataLabels are visible from before initial animation and then they blink.
This is visible when `dataLabels` have `useHTML: true` and pie has set `size`

**Example:** http://jsfiddle.net/ayw4ajzg/

TorsteinHonsi:
It's not only pies: http://jsfiddle.net/highcharts/ayw4ajzg/1/

video-internal:
Should I just use highcharts.js and exporting.js latest versions and this issue will be solved. Am I right?

TorsteinHonsi:
​No, the fix hasn't been published yet, just pushed to the repo. See
https://github.com/highcharts/highcharts/blob/master/repo-guidelines.md#apply-a-fix
.

video-internal:
I'm not able to open http://i-tools.org/diff 

can anyone help me?

TorsteinHonsi:
Okay, the easy way is: 
1. Open your copy of highcharts.src.js v4.2.3 in a text editor.
2. Open https://github.com/highcharts/highcharts/commit/81e9bbfcec8e52e49910e679497b1de6666b49d2, and see what has changed in highcharts.src.js. 
3. Perform those changes to your copy of highcharts.src.js.

video-internal:
@TorsteinHonsi  - I'm using highcharts.js so I can't see anything like htmlGroup

I have used this version - https://code.highcharts.com/highcharts.js

TorsteinHonsi:
Then you must use https://code.highcharts.com/highcharts.src.js and optionally go to the [Closure Compiler web service](https://closure-compiler.appspot.com/) to compile it.

video-internal:
@TorsteinHonsi - I used that version of js fie and found where I have to enter that patch, but I'm unable to compile the file using the web service you provided. I just get something went wrong page.

I just copied and pasted whole js file after editing there

TorsteinHonsi:
I think you need to host it online and point the closure compiler to the location - it's all explained on the compiler web service.

video-internal:
@TorsteinHonsi - thank you for help

Issue #5010, opened 2016-02-06
 error when stacking Area charts 


mattwang75:
I created an example to reproduce it:  http://jsfiddle.net/a0yxuwx1/10/

Series are not rendered.  In Chrome, the console shows this error: "Uncaught TypeError: Cannot read property 'points' of undefined".

If you change the chart type from 'area' to some other type (such as 'line' or 'column'), then my example works fine.  What's so special about 'area' chart type that can't be stacked?

TorsteinHonsi:
The problem here is that `new Date` returns a string, while [point.x expects a number](http://api.highcharts.com/highcharts#series<area>.data.x). It works by accident with line series, but you'll see that the X values are not what you expect.

Here's a sample where I call `.getTime()` on your date objects: http://jsfiddle.net/highcharts/a0yxuwx1/11/

Issue #4856, opened 2015-12-16
zooming doesn't work in Tree Map with Color Axis


dmitrygorbenko:
Hi,

I created my demo based on your demo of tree map with color axis where I enabled zooming:
https://jsfiddle.net/DimaGorbenko/mrjnLxao/
it has version 4.2.0.

Zooming seems like is broken since 4.1.9.
Here is a working version with 4.1.8: http://jsfiddle.net/DimaGorbenko/mrjnLxao/2/
Here is a not working version with 4.1.9: http://jsfiddle.net/DimaGorbenko/mrjnLxao/3/

( I tired to search for similar bug but didn't find it, so I created this new one )

Thanks
Dima

pawelfus:
Fails since https://github.com/highcharts/highcharts/commit/53f13b3

fclmman:
hi! Are there any news about this issue. We are looking for a tree map solution and highcharts were the bust, but zoom issue stops us.

jon-a-nygaard:
@fclmman For now there is unfortunately no update on this issue. We have an option [allowDrillToNode ](http://api.highcharts.com/highcharts#plotOptions.treemap.allowDrillToNode) which might suit your needs.

fclmman:
@jon-a-nygaard thank you for responding, we will look into allowDrillToNode option, but what we really want to know is will this functionality, zooming i mean, be available in the future, and how soon, if it will?
Thank you in advance.

jon-a-nygaard:
@fclmman This issue will be fixed in the upcoming maintenance release. If you are in need of a workaround before the release, you can try out the one below.

__Workaround__
Open the module `treemap.src.js` and look for the following code in the translate function around line 591.
```javascript
// Replace this code
val = this.nodeMap[this.rootNode].pointValues;
this.xAxis.setExtremes(val.x, val.x + val.width, false);
this.yAxis.setExtremes(val.y, val.y + val.height, false);
this.xAxis.setScale();
this.yAxis.setScale();

// With this code
if (this.options.allowDrillToNode) {
	val = this.nodeMap[this.rootNode].pointValues;
	this.xAxis.setExtremes(val.x, val.x + val.width, false);
	this.yAxis.setExtremes(val.y, val.y + val.height, false);
	this.xAxis.setScale();
	this.yAxis.setScale();
}
```

Issue #4813, opened 2015-12-08
Mapline is filled after hover/mouseover. 


Izothep:
In Highmaps mapline is filled after hover. When mouse leaves the mapline it still stays filled.

example:
http://jsfiddle.net/izothep/2yunqjey/9/
To recreate the issue hover the line.

The same problem exists in Highcharts demos:
http://www.highcharts.com/maps/demo/mapline-mappoint

This issue is not present when tested with versions of Highmaps older then 1.1.9: 

working example with 1.1.8. version:
http://jsfiddle.net/2yunqjey/11/

TorsteinHonsi:
Fails since c1aaaed

pawelfus:
Right. I have question before fixing/reverting: do we have any specific reason for setting `fill` for mapline series? See: https://github.com/highcharts/highcharts/blob/0e3824372b2a82c70dd081157de84dceef9013b3/js/parts-map/MapLineSeries.js#L14

TorsteinHonsi:
It is probably because in a mapline, the color value sets the stroke, so the stroke reflects the point value. That allows us to set a static fill. And [it's documented](http://api.highcharts.com/highmaps#plotOptions.mapline.fillColor), so I guess we must keep backwards compatibility.

pawelfus:
Ok, so by default, on mouseOver should we fill the path or it suppose to work as in 1.1.8 version (on hover, only stroke get highlighted)?

TorsteinHonsi:
The fill color is static, so it shouldn't change on mouse over/out.

jbkunst:
Hi everyone! Thank so much for highcharts!

I use the highstock.src with the fix to avoid the fill in maplines, but breaks when I want to make the combo chart (pie chart into a barchart) the error in the console is:

```
Uncaught TypeError: Cannot read property 'val2lin' of undefined
```

Hope it helps.

TorsteinHonsi:
@jbkunst It is probably unrelated to this fix. [This fiddle](http://jsfiddle.net/highcharts/12o8f4dp/) shows the combo chart with `highstock.js` pulled from the master branch, where this fix is present.

jbkunst:
@TorsteinHonsi  Yep, you're right. I guess I miss something.

Sorry and thanks again!

TorsteinHonsi:
Reverted the fix, as it broke the test for #3636.

pawelfus:
[...]

PR #11181, opened 2019-06-26
Tests/moving v.4.2.3 to unit tests

sigrunnu:
Moved issue test to unit test, v.4.2.3:
#4679
#4700
#4813
#4856
#5010
#5075


samples/unit-tests/pointer/click/demo.js
```diff
@@ -1,3 +1,69 @@
+function test(type, tooltipEnabled) {
```
TorsteinHonsi: This is polluting the global scope, which may come back and haunt us later. Wrap the `test` function and its calls (only the changes in this diff) in an iife.

samples/unit-tests/series-bubble/minmaxsize/demo.js
```diff
@@ -1,3 +1,52 @@
+QUnit.test('Setting X axis extremes on bubble (#5167)', function (assert) {
```
TorsteinHonsi: Not related to `minSize` or `maxSize`, so a better location would be `unit-tests/series-bubble/extremes`.

samples/unit-tests/series-column/zero-column-visible/demo.js
```diff
@@ -0,0 +1,16 @@
+QUnit.test('Zero column visible (#5146)', function (assert) {
```
TorsteinHonsi: The folder name is a bit too specific, rename it to `series-column/data`.

samples/unit-tests/tooltip/zindex/demo.js
```diff
@@ -0,0 +1,83 @@
+function test(key, options) {
```
TorsteinHonsi: Polluting the global namespace, wrap the whole content of this file in an iife or a QUnit.module.

sigrunnu:
The two commits "renamed file" is one renaming and one deleting of the old. Both belongs to issue test #5146.

The "Wrapped #5146 in iife" is for issue test #4679 and not #5146.

Issue #5261, opened 2016-04-28
Tooltip height incorrect if content starts with line break


nipehe:
#### Expected behaviour
Formatter either strips line breaks from the start of the string OR displays tooltips with enough height.

#### Actual behaviour
Line breaks at the start of a tooltip formatter cause the displayed text to have a line break at the beginning (as specified) but tooltip bubble height is too low, causing overflow.

#### Live demo of the issue
http://jsfiddle.net/j840waqa/

#### Steps to reproduce the behaviour
1. Add a tooltip formatter with at least one `<br>` tag at the beginning

#### Affected browser(s)
Chrome 49, Firefox 38, IE11

TorsteinHonsi:
Thanks for reporting! 

#### Minimal demo
The bug applies to all labels: http://jsfiddle.net/highcharts/ery8vzfa/

Issue #5254, opened 2016-04-27
VerticalAlign works incorrect in the heatmap


sebastianbochan:
#### Expected behaviour
Datalabels should be vertically aligned.

#### Actual behaviour
Labels oscillate about middle point

#### Live demo of the issue
http://jsfiddle.net/L0brzonp/

TorsteinHonsi:
Thanks for reporting! The fix will be part of the next maintenance release.

### Workaround
Until the fixed release, a simple fix is to borrow the `alignDataLabel` function from the column prototype:
```js
	Highcharts.seriesTypes.heatmap.prototype.alignDataLabel = Highcharts.seriesTypes.column.prototype.alignDataLabel;
```

[View it live on jsFiddle](http://jsfiddle.net/highcharts/L0brzonp/1/).

Issue #5236, opened 2016-04-21
Last visible point not always hoverable


sirxemic:
#### Expected behaviour
All visible points should be hoverable. E.g. I want the marker line and tooltip to update when I move close to the last visible point.

#### Actual behaviour
In the test case it doesn't happen when the chart is wide. However, when it's less wide, it works fine.

#### Live demo of the issue
http://jsfiddle.net/qxpn4kbs/

#### Affected browser(s)
Tested in Chrome 49.0.2623.112m and Firefox 45.0.2

TorsteinHonsi:
Thanks for reporting! An interesting observation is that if you make the chart 999px or 1001px wide the last point is included, but with 1000px is it not. This indicates rounding issues.

PR #11187, opened 2019-06-27
Tests/moving v.4.2.4 to unit tests

sigrunnu:
Moved issue tests v.4.2.4 to unit-tests:
#5236
#5254
#5261


samples/unit-tests/axis/labels/demo.js
```diff
@@ -1,3 +1,49 @@
+QUnit.test('Left trim (#5261)', function (assert) {
```
TorsteinHonsi: Move this to the existing `unit-tests/svgrenderer/label/demo.js`

samples/unit-tests/series-spline/last-point-clipped/demo.js
```diff
@@ -0,0 +1,41 @@
+QUnit.test('Check last point visible (#5236)', function (assert) {
```
TorsteinHonsi: Sample folder name is a bit too specific, rather call it `series-spline/data/`.

Issue #5721, opened 2016-09-22
Tooltip / Point selection issue in v4.2.7


nvindice:
#### Expected behaviour
With "sticky" but without "shared tooltip" option the tooltip of the nearest data point should be displayed. This worked fine until version 4.2.6.

#### Actual behaviour
Since version 4.2.7, the x-nearest point (and tooltip) is being selected, not the (x,y)-nearest point.

#### Live demo with steps to reproduce
http://jsfiddle.net/29ho1gd8/1/
Place the cursor around 6:01 am, it will always display the "Sleep"-Tooltip, even though the cursor is at the very bottom of the chart.

#### Possible reason
In [line 10424](https://github.com/highcharts/highcharts/blob/master/js/highcharts.src.js#L10424), only the x-distance is being taken into account, the code in line 10429 (desired behaviour) won't be reached normally. The comment in L10423 says "...and shared tooltip", which is not being checked.

Thank you for this great product!

mhwlng:
f.y.i.
I had the same problem with tooltips in stockchart. Reverting to 4.2.6 fixed it.

oysteinmoseng:
Thank you for reporting. I'm however seeing the same behavior also in 4.2.6 and 4.2.5?

nvindice:
Interestingly, you have a point. At least for this live demo - my real world example works fine with 4.2.6, can't see the difference right now.

However, modifying [L10424 ](https://github.com/highcharts/highcharts/blob/master/js/highcharts.src.js#L10424) to check for the "shared" property fixes the problem in my real world use case as well as in the live demo: [http://jsfiddle.net/29ho1gd8/2/](http://jsfiddle.net/29ho1gd8/2/)

`#10424: if (shared && isCloserX !== 0) {`

TorsteinHonsi:
@pawelfus Any insight?

pawelfus:
I think there should be `&& shared`, just like OP said. Let me run tests and we will see.

nvindice:
Thank you very much for your fast and helpful response. Is there a planned date for the next (bugfixed) release? Unfortenately, I cannot use my own (fixed but unofficial) version.

mhwlng:
![2016-09-30_10-36-53](https://cloud.githubusercontent.com/assets/1388886/18985778/f56240ae-86f9-11e6-959b-5c2420cc47b0.png)

I have the same problem again with stockchart in v 5.0.0.0

pawelfus:
[...]

PR #11190, opened 2019-06-27
tests/moving-v.2.4.7-to-unit-tests

sigrunnu:
Tests: Moved issue-test #5721 to unit-tests.


samples/unit-tests/tooltip/non-shared/demo.js
```diff
@@ -0,0 +1,34 @@
+QUnit.test('Non-shared tooltip should highlight the clostest point (#5721)', function (assert) {
```
TorsteinHonsi: Delete this test, we have good coverage with the test for `findNearestPointBy`.

PR #11201, opened 2019-06-27
bugfix/11095-update-switchrowsandcolumn

### LINKED ISSUE #11095: After using method Highcharts.chart.update property switchRowsAndColumns not saved
#### Expected behaviour
save property switchRowsAndColumns in Highchart.chart.options and in Highchart.chart.options.data.

#### Actual behaviour
Property not saved. Other property, like csv and delimiters saved correctly after update.
If save chart.options object and try to render a new chart with this object, chart will be not correct, if we switch rows and columns in previous version.

#### Product version
highcharts 7.03

pawelfus:
Fixed #11095, updating `switchRowsAndColumns` did work only for the first time.

### DIFF
--- a/js/modules/data.src.js
+++ b/js/modules/data.src.js
@@ -2199,8 +2199,8 @@ Highcharts.extend(Data.prototype, {
                 }
             };
             // Apply it
-            merge(true, this.options, options);
-            this.init(this.options);
+            merge(true, chart.options.data, options);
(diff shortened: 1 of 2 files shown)


TorsteinHonsi:
Thanks for the PR, remember to add the `Changelog: Bugfix` label where appropriate.

PR #11207, opened 2019-06-28
tests/moving-issue-test-compare-negative

GoodDD:
Tests/moving issue test compare negative


samples/unit-tests/series/data/demo.js
```diff
@@ -1,3 +1,23 @@
+QUnit.test('Compare to negative (#4661)', function (assert) {
```
TorsteinHonsi: Move this test to the existing `unit-tests/series/compare/demo.js` file.

PR #11208, opened 2019-06-28
bugfix/11109-oldie-polyfills

### LINKED ISSUE #11109: Internal ES5 target conflicts with Array.find usage
Actually everyone should have problems with using `Array.find` because `tsconfig.compiler.target` is set to `es5`, but somehow except for @pawelfus everyone has the es2015 declaration in use. Maybe the module `es6` is the reason behind it. We need to solve it properly. The oldie module adds a find function on the Array.prototype, so it should be solved there (if not in tsconfig).

bre1470:
Fixed #11109 - extended oldie polyfill with internal type declarations.

### DIFF
--- a/ts/modules/oldie-polyfills.src.ts
+++ b/ts/modules/oldie-polyfills.src.ts
@@ -15,33 +15,49 @@
 
 'use strict';
 
+import H from '../parts/Globals';
+
 /**
  * Internal types
  * @private
  */
-declare interface ArrayFilterCallbackFunction<T> {
-    (value: T, index: number, array: Array<T>): boolean;
-}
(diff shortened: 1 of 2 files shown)


pawelfus:
Looks good! Is that safe? Won't we accidentally use `array.find(..)` instead of `Highcharts.find(array, ..)` ?

bre1470:
> Looks good! Is that safe? Won't we accidentally use `array.find(..)` instead of `Highcharts.find(array, ..)` ?

It will not change a thing like it is at the moment. These function are just internal for the compiler to not complain about missing functions for the target.

PR #11216, opened 2019-06-28
docs/jsdoc-internal-options

bre1470:
Introduced JSDoc tag `@internal` for options. Similar to TypeScript you can tag options doclets, that should not go into the TypeScript declarations. The result is, that they will show up only in the API documentation.


pawelfus:
Could you remind me difference between `@ignore-option` and `@internal`? I think we discussed this a while ago..

bre1470:
For options doclets the following behaviour is expected:
`@deprecated` will not show up in declarations.
`@ignore-option` will not show up in options and declarations.
`@internal` will not show up in declarations.
`@private` will not show up in class-reference (happens sometimes because of some doclet glitch).

PR #11217, opened 2019-06-28
Added sensible defaults to plot lines and bands

TorsteinHonsi:
Added sensible defaults to plot lines and bands. A plot band can now be initialized with only a `value`, and a plot band with either `from`, `to`, or both.

### DIFF
--- a/ts/parts/PlotLineOrBand.ts
+++ b/ts/parts/PlotLineOrBand.ts
@@ -82,7 +82,7 @@ declare global {
             events?: any;
             id?: string;
             label?: AxisPlotLinesLabelOptions;
-            value?: number;
+            value: number;
             width?: number;
             zIndex?: number;
         }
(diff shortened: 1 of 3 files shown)


ts/parts/PlotLineOrBand.ts
```diff
@@ -82,7 +82,7 @@ declare global {
             events?: any;
             id?: string;
             label?: AxisPlotLinesLabelOptions;
-            value?: number;
+            value: number;
```
bre1470: This has to be kept optional, because it is not describing a code-side interface, but an options interface from our API options doclets. These properties are always optional by nature, because a user can choose, which should be set and which not.

ts/parts/PlotLineOrBand.ts
```diff
@@ -229,18 +229,17 @@ H.PlotLineOrBand.prototype = {
         // Set the presentational attributes
         if (!axis.chart.styledMode) {
             if (isLine) {
-                attribs.stroke = color as any;
+                attribs.stroke = (color as any) || '${palette.neutralColor40}';
                 attribs['stroke-width'] =
-                    (options as Highcharts.AxisPlotLinesOptions).width;
+                    (options as Highcharts.AxisPlotLinesOptions).width ||
```
pawelfus: Just for a bullet proof solution, I would go with `pick()` instead of `||`. Setting `width=0` is the easiest way to hide/show plotLine

bre1470: `pick()` is not type safe (at the moment).

ts/parts/PlotLineOrBand.ts
TorsteinHonsi: 👍

TorsteinHonsi:
Closing this, unable to resolve conflicts with master

PR #11220, opened 2019-06-28
bugfix/11212-networkgraph-selected-state

### LINKED ISSUE #11212: Networkgraph: selected state pollutes other nodes
#### Expected behaviour
Only clicked node should be highlighted.

#### Actual behaviour
All connected nodes are highlighted.

Demo: https://jsfiddle.net/BlackLabel/bs4p0dmk/

**Workaround:**
```js
Highcharts.seriesTypes.networkgraph.prototype.pointClass.prototype.setState = Highcharts.seriesTypes.line.prototype.pointClass.prototype.setState;
```

And disable inactive state for subnodes: `series.inactiveOtherPoints = false`.

Demo: https://jsfiddle.net/BlackLabel/t2uy159p/

pawelfus:
Fixed #11212, selected state for a node in network graph was applied to all connected nodes.
___
It's a temporary solution, but makes presentational attributes of a node consistent with `point.selected` flag. If more requests like will show up, we should go back to the discussion about neighbourhood for a points/nodes/cells in series and states (ref heatmap: https://jsfiddle.net/BlackLabel/x71wvh39/show )

### DIFF
--- a/samples/unit-tests/series-networkgraph/networkgraph/demo.js
+++ b/samples/unit-tests/series-networkgraph/networkgraph/demo.js
@@ -2,6 +2,17 @@ QUnit.test('Network Graph', function (assert) {
     var chart = Highcharts.chart('container', {
         chart: {
             type: 'networkgraph'
+        },
+        plotOptions: {
+            series: {
+                marker: {
+                    states: {
+                        select: {
+                            fillColor: 'orange'
+                        }
+                    }
+                }
+            }
         }
     });
     var point;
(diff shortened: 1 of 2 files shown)


samples/unit-tests/series-networkgraph/networkgraph/demo.js
```diff
@@ -88,6 +99,16 @@ QUnit.test('Network Graph', function (assert) {
...
     );
 
+    chart.series[0].nodes[0].setState('select');
+
+    assert.strictEqual(
+        chart.series[0].nodes.filter(
+            node => node.graphic.element.getAttribute('fill') === 'orange'
+        ).length,
+        1,
+        'Only one node has selected attributes (#11212)'
```
TorsteinHonsi: Better (but not a blocker): `'Only one node should have selected attributes (#11212)'`.

PR #11222, opened 2019-06-28
samples/periodic-table

pawelfus:
Samples: Created periodic table (using heatmap and bubble series). 

Available only locally: http://utils.highcharts.local/samples/#view/highcharts/series-heatmap/periodic-table

"Share" doesn't work - I don't know the correct URL for data sample. Right now I'm pointing `url: 'http://utils.highcharts.local/samples/data/periodic-table.json',` - [here]( https://github.com/highcharts/highcharts/compare/samples/periodic-table?expand=1#diff-95996bba86e84e9ba76be4934bb67ea5R2) - which is definitively wrong.

Demo probably requires a bit more polishing (e.g. fixed size? maybe colors for color axis).

**Known issues:**
- Calling `setData(points)` animates nicely bubbles, but `chart.update({ series: [{ points }] })` doesn't.
- I wasn't able to animate colors for heatmap. Probably caused by extremes that change on update too.

**Ideas:**
There's more properties for an element we can show, but I decided to use only the most common (and numerical ;) ) ones. 
Tooltip can show more information (e.g. who discovered and when).


pawelfus:
Issue with color animation reported: #11239

pawelfus:
>  For more space economics, the title can be moved into the white notch on top of the periodic table. Verify in mobile view.

Can this be done without custom code? I was thinking about moving there tooltip (which has `positioner`). Title can be a little tricky because of the cell size after window resize..

TorsteinHonsi:
> Title can be a little tricky because of the variable cell size after window resize

If it works with mobile and desktop widths it is okay, just use `x` for an acceptible placement. But there's also a problem that the Y axis leaves one category of empty space at the top. Maybe we need to set a `min`.

pawelfus:
It doesn't leave that extra space for me: 

Screen Shot 2019-07-03 at 16 52 15


Could share a screenshot?

pawelfus:
Current design (need to fix the second one dropdown):

Screen Shot 2019-07-03 at 17 07 53
Screen Shot 2019-07-03 at 17 08 09

PR #11223, opened 2019-06-28
enhancement/shadows-inside-series

wchmiel:
Added shadows inside series.

### DIFF
--- a/js/parts/SvgRenderer.js
+++ b/js/parts/SvgRenderer.js
@@ -1808,14 +1808,12 @@ extend(SVGElement.prototype, /** @lends Highcharts.SVGElement.prototype */ {
                     'stroke-opacity': shadowOpacity,
                     'stroke-width': strokeWidth,
                     transform: translate,
-                    fill: 'none'
+                    fill: 'none',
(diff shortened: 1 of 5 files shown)


TorsteinHonsi:
What does this PR do? Is it an answer to an issue? Or is it a new option? Where can we see the effect, should it have a visual sample?

TorsteinHonsi:
Demo: http://jsfiddle.net/highcharts/zsy5ojnq/

Personally I'm not sure this feature is worth the bloat... Maybe it's better off as a plugin.

wchmiel:
Oh sorry, I forgot to add an additional comment.

It was a ticket that I've made on support. The client was asking how to add shadows inside a bar series something like CSS box-shadow inset. I thought it will be a good idea to add this feature. Of course, I can make it work better and add a visual sample, etc. but first I wanted you to see it and decide if it makes any sense for you. A plugin can be a good idea.

TorsteinHonsi:
Thanks! Yes I think we go for a plugin in this case, as it is not a commonly requested feature.

PR #11237, opened 2019-07-01
enhancement/8912-stackLabels-crop-overflow

Basts:
Refactoring stack-labels: adding crop and overflow features. See #8912.


TorsteinHonsi:
Rotation fails with the new logic, please fix: http://utils.highcharts.local/samples/#test/highcharts/yaxis/stacklabels-rotation

Basts:
@TorsteinHonsi Yeah, right! Thank you for pointing out! 

Fixed here #11525

PR #11240, opened 2019-07-01
Fixed #11229, avoid negative width and height in heatmap shapes

### LINKED ISSUE #11229: A negative width/height of heatmap cells is not valid
#### Expected behaviour
Heatmap is rendered without Js error

#### Actual behaviour
If there are many categories on xAxis or yAxis then the original width/height of heatmap cells may be less than `pointPadding * 2`. So after applying pointPadding, the cells have negative width or height.
example: 
`{Height: 1px; pointPadding: 2px}`
=> after apply pointPadding, new height is `-3` and heatmap is rendered with js error
`Error: <rect> attribute height: A negative value is not valid. ("-3")`

https://jsfiddle.net/corundhope/nzphexyv/10/

#### Product version

Highcharts 7.1.1

#### Affected browser(s)
All browsers

jon-a-nygaard:
Fixed #11229, avoid negative width and height in Heatmap shapes.

### DIFF
--- a/js/parts-map/HeatmapSeries.js
+++ b/js/parts-map/HeatmapSeries.js
@@ -276,8 +276,8 @@ seriesType(
                 point.shapeArgs = {
                     x: Math.min(x1, x2) + pointPadding,
                     y: Math.min(y1, y2) + pointPadding,
-                    width: Math.abs(x2 - x1) - pointPadding * 2,
-                    height: Math.abs(y2 - y1) - pointPadding * 2
+                    width: Math.max(Math.abs(x2 - x1) - pointPadding * 2, 0),
+                    height: Math.max(Math.abs(y2 - y1) - pointPadding * 2, 0)
                 };
             });
 


js/parts-map/HeatmapSeries.js
```diff
@@ -276,8 +276,8 @@ seriesType(
                 point.shapeArgs = {
                     x: Math.min(x1, x2) + pointPadding,
                     y: Math.min(y1, y2) + pointPadding,
-                    width: Math.abs(x2 - x1) - pointPadding * 2,
-                    height: Math.abs(y2 - y1) - pointPadding * 2
+                    width: Math.max(Math.abs(x2 - x1) - pointPadding * 2, 0),
+                    height: Math.max(Math.abs(y2 - y1) - pointPadding * 2, 0)
```
lanhq:
```
if (
    Math.abs(x2 - x1) - pointPadding * 2 < 0 ||
    Math.abs(y2 - y1) - pointPadding * 2 < 0
) {
     pointPadding = 0
}
```
Hi @jon-a-nygaard ,
I thought that if we reset `pointPadding` before set width or height is better. It mean that pointPadding will be ignored if heatmap shapes is too small. what do you think?

jon-a-nygaard: Thanks for your input @lanhq, but I am worried that removing the `pointPadding` altogether in these cases would be an unpredictable behavior, especially when `pointPadding` is a larger value.

PR #11254, opened 2019-07-02
docs/11198-fix-datetimelabelformats-type

### LINKED ISSUE #11198: Docs are confusing/incomplete for dateTimeLabelFormats
#### Expected behaviour
As described on the docs page (https://api.highcharts.com/highstock/xAxis.dateTimeLabelFormats) I expect that I can write the configuration of the dateTimeLabelFormats section as described in the default values. 

#### Actual behaviour
Instead of the values accepting a string, like in the defaults, it requires an object - some of which have both a `main` and a `range` variable, some of which only have a `main` value. It is not described what those variables are for, albeit that it's obvious that the `main` value is for the formatting string.

n/a

#### Product version
Highcharts 7.1.2, typescript 3.4.5.

#### Affected browser(s)
n/a

bre1470:
Fixed #11198 - Incomplete type for dateTimeLabelFormats option.


ts/parts/Axis.ts
```diff
@@ -149,14 +149,14 @@ declare global {
...
+            day?: (string|XAxisDateTimeLabelFormatsSubOptions);
+            hour?: (string|XAxisDateTimeLabelFormatsSubOptions);
+            millisecond?: (string|XAxisDateTimeLabelFormatsSubOptions);
+            minute?: (string|XAxisDateTimeLabelFormatsSubOptions);
+            month?: (string|XAxisDateTimeLabelFormatsSubOptions);
+            second?: (string|XAxisDateTimeLabelFormatsSubOptions);
+            week?: (string|XAxisDateTimeLabelFormatsSubOptions);
+            year?: (string| XAxisDateTimeLabelFormatsSubOptions);
         }
         interface XAxisDateTimeLabelFormatsSubOptions {
```
TorsteinHonsi: I see this is an existing interface, but I think the name is too specific. It could be generalized to `DateTimeLabelFormatOptions`

bre1470: Renamed and moved.

Issue #4202, opened 2015-05-19
Higstock scrollbar is not well placed with empty charts in 2.1.5


sebastianbochan:
Run below examples, click resize and let's compare:
- http://jsfiddle.net/mtze4j3m/6/ HS 2.0.5
- http://jsfiddle.net/mtze4j3m/5/ HS 2.1.5

The scrollbar is placed in the top of chart (2.1.5).

Issue #4196, opened 2015-05-18
Highstock v2.1.5 axis can't handle the option "ordinal" change properly


dreamershl:
In the beforeSetTickPositions(), it will calculate the "ordinalPositions" if the axis.options.ordinal is true. But if the application change the axis.options.ordinal from "true" to "false", the function doesn't set the flag "axis.doPostTranslate" properly. So in the translate(), the axis will still believe the tick position should still rely on the outdated "ordinalPositions" array.

highslide-software:
Demo: http://jsfiddle.net/highcharts/knvrz35z/

When starting with `ordinal:false` it correctly updates to `ordinal:true`.

dreamershl:
This issue is more on the "ordinalPositions" re-calculation after the ordinal flag is changed. The demo is using the same data. While my case is I change the data array as well as the ordinal flag is changed. Later will change your example to demo this issue. Thx

dreamershl:
I have modified the example to demo the issue. In the modified example, if I change the seriers data without redraw flag to true, the axis ticks are wrong even I put the "chart.redraw();" to redraw the chart later.

http://jsfiddle.net/pdu7Lp7f/3/

          chart.series[0].setData(!xAxis.options.ordinal ? data: data2, false);

highslide-software:
Seems to work as expected with the fix applied: http://jsfiddle.net/highcharts/pdu7Lp7f/4/

dreamershl:
Yes.. thx for the fast fix. now I can remove my patch in my app. Thx again

PR #11256, opened 2019-07-02
tests/moving-v2.1.5-to-unit-tests

GoodDD:
Moved issue test v2.1.5 to unit-tests.
#4196
#4202


samples/unit-tests/chart/chart-update/demo.js
```diff
@@ -376,4 +376,87 @@
             'Options should not be mutated after chart.update'
         );
     });
+
```
TorsteinHonsi: Move to `unit-tests/axis/update/demo.js`

Issue #8834, opened 2018-08-23
Multiple exception in log console Error: <text> attribute y: Expected length, "NaN"


ivanmjartan:
We have all negative series in Stacked column chart when we try set yAxis min value for value grater than 0 than graph looks rendered correctly (any items in view) but in console is flooded by exceptions. 

I guess thats it looks  like bug i not pls how set chart to avoid given ex.  

#### Expected behaviour
No exception in log console 

#### Actual behaviour
Several exceptions in log console

highcharts.js:83 Error: <text> attribute y: Expected length, "NaN".
a.xSetter @ highcharts.js:83
(anonymous) @ highcharts.js:43
a.objectEach @ highcharts.js:28
attr @ highcharts.js:42
align @ highcharts.js:50
setOffset @ highcharts.js:322
translate @ highcharts.js:301
translate @ highcharts.js:354
(anonymous) @ highcharts.js:265
a.each @ highcharts.js:28
renderSeries @ highcharts.js:265
render @ highcharts.js:267
firstRender @ highcharts.js:270
(anonymous) @ highcharts.js:245
a.fireEvent @ highcharts.js:31
init @ highcharts.js:244
getArgs @ highcharts.js:244
a.Chart @ highcharts.js:243
a.chart @ highcharts.js:243
(anonymous) @ (index):39
2highcharts.js:83 Error: <tspan> attribute y: Expected length, "NaNNaN".
2highcharts.js:83 Error: <tspan> attribute y: Expected length, "NaNNaNNaN". 
..

#### Live demo with steps to reproduce
https://jsfiddle.net/m7k54px8/31/

#### Product version
Current

#### Affected browser(s)
Tested in Chrome

pawelfus:
Hi @ivanmjartan 

Thank you for reporting this issue. Issue comes from `stackLabels` - when disabled, it works fine: https://jsfiddle.net/BlackLabel/m7k54px8/37/

Minified demo: https://jsfiddle.net/BlackLabel/1c8fht32/

dandv:
I see the same error with a stockChart, and no `stackLabels`. No [Gantt](https://github.com/highcharts/highcharts/issues/12088) either. I'll try to make a reproducible sample.

PR #11263, opened 2019-07-02
samples/stacklabels-demo-improvement

Basts:
Fixing stack labels demo, touch #8912

### DIFF
--- a/samples/unit-tests/axis/stacklabels/demo.js
+++ b/samples/unit-tests/axis/stacklabels/demo.js
@@ -37,11 +37,11 @@ QUnit.test('Stack labels on non-data axis', function (assert) {
     );
 });
 
-QUnit.skip('Stack labels crop and overflow features #8912', function (assert) {
+QUnit.test('Stack labels crop and overflow features #8912', function (assert) {
     var chart = Highcharts.chart('container', {
         chart: {
             type: 'column',
-            width: 300,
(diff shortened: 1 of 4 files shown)


samples/unit-tests/axis/stacklabels/demo.js
```diff
@@ -21,27 +21,26 @@ QUnit.test('Stack labels on non-data axis', function (assert) {
...
 
-    assert.close(
+    assert.strictEqual(
         chart.container.querySelector('.highcharts-stack-labels text')
-            .getBBox().y,
-        0,
-        1,
-        'The y attribute is around to 0.'
+            .getAttribute('y') !== null,
+        true,
```
TorsteinHonsi: Instead of `strictEqual` with an !== expression in the first parameter, use `notStrictEqual`:

samples/unit-tests/axis/stacklabels/demo.js
```diff
@@ -21,27 +21,26 @@ QUnit.test('Stack labels on non-data axis', function (assert) {
...
+        'Y attribute is set (#8834)'
```
TorsteinHonsi: Instead of stating what it is (which is ambiguous when the test fails), say what it _should_ be. ("The Y attribute should be set (#8834)").

PR #11264, opened 2019-07-02
ts/areaseries

### LINKED ISSUE #11021: Migrate AreaSeries to TypeScript

bre1470:
Fixed #11021 - Migrated AreaSeries to TypeScript.

### DIFF
--- a/ts/.eslintrc
+++ b/ts/.eslintrc
@@ -37,6 +37,7 @@
         "no-undefined": 0,
         "no-underscore-dangle": 0,
         "no-unused-expressions": 0,
+        "no-use-before-define": 0,
         "no-useless-constructor": 0,
         "object-curly-spacing": [2, "always"],
         "object-shorthand": 0,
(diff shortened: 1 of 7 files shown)


ts/.eslintrc
```diff
@@ -37,6 +37,7 @@
         "no-undefined": 0,
         "no-underscore-dangle": 0,
         "no-unused-expressions": 0,
+        "no-use-before-define": 0,
```
jon-a-nygaard: @bre1470 Do we need to disable this? I am rather in favor of keeping this rule and instead move the `isNumber` to before it is used.

jon-a-nygaard: @bre1470 Do we need to disable this? I am rather in favor of keeping this rule and instead move the isNumber to before it is used.

bre1470: The rule seems to be quite buggy, because it does not realize, that the function is already available at that line of the code where it complains. I see no point in having faulty rules turned on.

bre1470: And if we have a situation where it is really a problem, then our unit tests will right away fail.

bre1470: Actually TypeScript would even before that complain about it and not transpile the code at all.

pawelfus: Hmm.. maybe `"no-use-before-define": ["error", { "functions": false, "classes": true, "variables": true }]`?

TorsteinHonsi: We decided to disable the rule for now

PR #11310, opened 2019-07-04
Tools: Added config for enabling builds from vacation branch

activist:
This should only be merged if we are to enable CI runs from vacation2019 branch.


activist:
Removed as we instead enabled building from branches directly in CI.

PR #11318, opened 2019-07-04
bugfix/8608-bubble-zparam-hidden

### LINKED ISSUE #8608: Bubble chart not seen properly when z param is not set in datapoint
#### Expected behaviour
In earlier Highchart versions, bubble chart was plotted correctly even when the z param was not set in datapoint. Backward compatibility should be supported.

#### Actual behaviour
Consider a case of plotting bubble chart where z param is not set in datapoints.
Example: https://jsfiddle.net/amrutaJgtp/hmvfrwa9/
Above JS fiddle shows that bubbles are not seen.

However, in Highcharts v6.0.6 - z param in bubble chart datapoint was optional.
Example: https://jsfiddle.net/amrutaJgtp/hmvfrwa9/1/

#### Product version
v6.1.1

#### Affected browser(s)
All

sebastianbochan:
Fixed #8608, bubble markers were hidden when z param missing.

### DIFF
--- a/js/parts-more/BubbleSeries.js
+++ b/js/parts-more/BubbleSeries.js
@@ -316,6 +316,7 @@ seriesType('bubble', 'scatter', {
         var len,
             i,
             zData = this.zData,
+            yData = this.yData,
             minSize = series.minPxSize,
             maxSize = series.maxPxSize,
             radii = [],
@@ -325,13 +326,22 @@ seriesType('bubble', 'scatter', {
         for (i = 0, len = zData.length; i < len; i++) {
             value = zData[i];
             // Separate method to get individual radius for bubbleLegend
-            radii.push(this.getRadius(zMin, zMax, minSize, maxSize, value));
+            radii.push(
+                this.getRadius(
+                    zMin,
+                    zMax,
+                    minSize,
+                    maxSize,
+                    value,
+                    yData[i]
+                )
+            );
         }
         this.radii = radii;
     },
 
     // Get the individual radius for one point.
-    getRadius: function (zMin, zMax, minSize, maxSize, value) {
+    getRadius: function (zMin, zMax, minSize, maxSize, value, yValue) {
         var options = this.options,
             sizeByArea = options.sizeBy !== 'width',
             zThreshold = options.zThreshold,
@@ -350,8 +360,8 @@ seriesType('bubble', 'scatter', {
             zMin = 0;
         }
 
-        if (value === null) {
-        //if (!isNumber(value)) {
+        // #8608 - bubble should be visible when z is undefined
+        if (yValue === null || value === null) {
             radius = null;
         // Issue #4419 - if value is less than zMin, push a radius that's
         // always smaller than the minimum size
--- a/js/parts-more/BubbleSeries.js
+++ b/js/parts-more/BubbleSeries.js
@@ -350,7 +350,8 @@ seriesType('bubble', 'scatter', {
             zMin = 0;
         }
 
-        if (!isNumber(value)) {
+        if (value === null) {
+        //if (!isNumber(value)) {
             radius = null;
         // Issue #4419 - if value is less than zMin, push a radius that's
         // always smaller than the minimum size
--- a/samples/unit-tests/series-bubble/marker/demo.js
+++ b/samples/unit-tests/series-bubble/marker/demo.js
@@ -87,3 +87,27 @@ QUnit.test('Clicking marker (#6705)', function (assert) {
         'Click event fired'
     );
 });
+
+QUnit.test('Bubble data points without z-param.(#8608)', function (assert) {
+    var chart = Highcharts.chart('container', {
+        chart: {
+            type: 'bubble',
+        },
+        series: [{
+            data: [{
+                x: 95,
+                y: 95
+            },
+            {
+                x: 86.5,
+                y: 102.9
+            }]
+        }]
+    });
+
+    assert.strictEqual(
+        typeof chart.series[0].points[0].graphic,
+        'object',
+        'Has marker'
+    );
+});


pawelfus:
How about implementing `isValid()` method in BubbleSeries' Point class (just like other, new series have) and then just checking if `point.isNull` ?

jon-a-nygaard:
I like your idea here @pawelfus, but I am struggling to see how this should be implemented.

For instance, where would the check for `point.isNull ?` be made, I can't see that we have access to `point` in `getRadius` or `getRadii`?

Could you elaborate a bit on the implementation?

sebastianbochan:
We cannot use isNumber because of boost issues. 

`if (value === null) {` is reverting, I mean that it was in the past and worked. So at this moment I think that we should leave it in this form.

jon-a-nygaard:
> if (value === null) { is reverting, I mean that it was in the past and worked.

For me it seems that we should not use this either, because `value = undefined` would result in `radius = NaN`.

Is `if (typeof value === 'number')` a possibility, even if `isNumber` can not be used?

Also out of curiosity, why is `isNumber` causing issues with boost?

pawelfus:
> For instance, where would the check for point.isNull ? be made, I can't see that we have access to point in getRadius or getRadii?
>
> Could you elaborate a bit on the implementation?

Good point, `getRadii` is called in `beforePadding`, so I'm not sure if points are already initialized.. Amnyway, I was thinking about this:
https://github.com/highcharts/highcharts/blob/a5ed2cec199d367c6eb64cfa12b32eefb913a10f/js/parts-more/BubbleSeries.js#L325-L329

Where we have access to `series`, so could be simple:
```js
 for (i = 0, len = zData.length; i < len; i++) { 
     // Separate method to get individual radius for bubbleLegend 
     radii.push(
         series.points[i].isNull ?
             null :
             this.getRadius(zMin, zMax, minSize, maxSize, zData[i])
         ); 
 } 
```
(ternary operator can be replaced with simple `if(!point.isNull) { .. }` if indexes for `radii` and `points` don't need to match)

oysteinmoseng:
Question: Is there any reason why we can't test against both `null` and `undefined` in `getRadius`? That would seem to be the simplest solution, instead of having to pull in logic from points?

And as Jon says, it would be interesting to know more about the issue with `isNumber` and boost. Is this something that could cause problems elsewhere when we use `isNumber`?

sebastianbochan:
Let me explain all aspects.

The `value` is extracted from `zData`, so when point has x/y values but z is missing, then value is `undefined`. As a result point should be visible, because has x/y values.

The point should be hidden only when y is null.

@pawelfus  in the `getRadii` function we do not have access to `series.points` or `series.data`, only `yData`. But it was good clue to modify this solution and use `yData`.

My idea is to check the y value instead of z value and then run drawing process.

PoC which works at this moment (not tested with all examples yet).

```
getRadii: function (zMin, zMax, series) {
        var len,
            i,
            zData = this.zData,
            yData = this.yData,
            minSize = series.minPxSize,
            maxSize = series.maxPxSize,
            radii = [],
            value;

        // Set the shape type and arguments to be picked up in drawPoints
        for (i = 0, len = zData.length; i < len; i++) {
            value = zData[i];
            // Separate method to get individual radius for bubbleLegend
            radii.push(
                this.getRadius(
                    zMin,
                    zMax,
                    minSize,
                    maxSize,
                    value, 
                    yData[i]
                )
            );
        }
        this.radii = radii;
    },
```

```
getRadius: function (zMin, zMax, minSize, maxSize, value, yValue) {
        var options = this.options,
            sizeByArea = options.sizeBy !== 'width',
            zThreshold = options.zThreshold,
            pos,
            zRange = zMax - zMin,
            radius;

        // When sizing by threshold, the absolute value of z determines
        // the size of the bubble.
        if (options.sizeByAbsoluteValue && value !== null) {
            value = Math.abs(value - zThreshold);
            zMax = zRange = Math.max(
                zMax - zThreshold,
                Math.abs(zMin - zThreshold)
            );
            zMin = 0;
        }

        if (!isNumber(yValue)) {
//...
```

What do  you think @oysteinmoseng @jon-a-nygaard @pawelfus

oysteinmoseng:
Yes, if the point should be hidden only when y is null I think it makes sense to test for y. This might mean that we don't need to get the radius then if the y value is null? Also presumably we still need to test against null/undefined for the z value in getRadius, in case it is not provided?

sebastianbochan:
@oysteinmoseng 
Thank you, will test it manually again ;)

sebastianbochan:
1. I cannot imagine the scenario when y value is null and bubble should be visible. Correct me if Im wrong.
2. I agree that it does not make sense to call `getRadius` function when y is null.

Updated code of `getRadii`
```
    getRadii: function (zMin, zMax, series) {
        var len,
            i,
            zData = this.zData,
            yData = this.yData,
            minSize = series.minPxSize,
            maxSize = series.maxPxSize,
            radii = [],
            value;

        // Set the shape type and arguments to be picked up in drawPoints
        for (i = 0, len = zData.length; i < len; i++) {
            value = zData[i];
            // Separate method to get individual radius for bubbleLegend
            radii.push(
                yData[i] === null ? null // skip calculation when y is null
                    this.getRadius(
                        zMin,
                        zMax,
                        minSize,
                        maxSize,
                        value, 
                        yData[i]
                    )
            );
        }
        this.radii = radii;
    },
```

The solution passes all our unit-tests.

@oysteinmoseng @jon-a-nygaard do you have any extra inputs about this solution?

oysteinmoseng:
@sebastianbochan Looks good to me! Assuming it does not break in getRadius if you have a y value, but z value is undefined?

sebastianbochan:
In case of undefined `z param`, bubble is draw, when null is not.

sebastianbochan:
It turned out that `getRadius` calling is requires (bubble Legend needs to radius information), but I mixed two solution (first and last) and now all works like a charm.

sebastianbochan:
It turned out that  the `getRadius` calling is required and we cannot skip taht. Why? Because `radius` variable is used by Bubble Legend. 

I combined two solutions (first and last) and now all works like a charm. However, the fix requires to be applied in TS (`BubbleSeries.ts` was created later than this PR).

New PR https://github.com/highcharts/highcharts/pull/11400

PR #11327, opened 2019-07-05
vacation2019/ts/mixins-derived-series

### LINKED ISSUE #11269: Migrate `mixins/derived-series` to TypeScript

bre1470:
Fixed #11269 - Migrated derived-series mixin to TypeScript.


ts/parts/Navigator.ts
```diff
@@ -56,7 +56,6 @@ declare global {
             navigator?: NavigatorOptions;
         }
         interface Series {
-            baseSeries?: Series;
```
jon-a-nygaard: What is the reason behind the removal of the `baseSeries` property?

bre1470: It is actually part of the mixins DerivedSeries class and is only used in the Navigator.

jon-a-nygaard: Conclusion after internal discussion with @bre1470, @oysteinmoseng, and @jon-a-nygaard is that the `baseSeries` property in Navigator is a different mechanism than the one in DerivedSeries, which means that we would like to still keep it in Navigator.ts.

PR #11341, opened 2019-07-08
 bugfix/8735-boost-gapsize

### LINKED ISSUE #8735: gapSize doesn't work in boost module
#### Expected behaviour
Gaps should be draw the same with and without boost module.

#### Actual behaviour
Gaps don't work in boost mode.

Works fine (no boost): http://jsfiddle.net/BlackLabel/8bect5rh/
Doesn't work (with boost): http://jsfiddle.net/BlackLabel/78yb3596/

#### Product version
6.1.1

pawelfus:
Fixed #8735, `gapSize` did not work in boost mode.


oysteinmoseng:
Await review from @cvasseng

PR #11352, opened 2019-07-09
vacation2019/ts/gaugeseries

### LINKED ISSUE #11072: Migrate `parts-more/GaugeSeries` to TypeScript

bre1470:
Fixed #11072 - Migrated GaugeSeries to TypeScript.

### DIFF
--- a/ts/parts-more/GaugeSeries.ts
+++ b/ts/parts-more/GaugeSeries.ts
@@ -1,8 +1,12 @@
 /* *
- * (c) 2010-2019 Torstein Honsi
  *
- * License: www.highcharts.com/license
- */
+ *  (c) 2010-2019 Torstein Honsi
+ *
+ *  License: www.highcharts.com/license
+ *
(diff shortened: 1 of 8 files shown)


ts/parts-more/GaugeSeries.ts
```diff
@@ -0,0 +1,682 @@
...
+    // and this will be used on the axes
+    angular: true,
+    directTouch: true, // #5063
+    drawGraph: noop,
+    fixedBox: true,
+    forceDL: true,
+    noSharedTooltip: true,
+    trackerGroups: ['group', 'dataLabelsGroup'],
+
+    /* eslint-disable valid-jsdoc */
```
jon-a-nygaard: `/* eslint-disable-next-line valid jsdoc */` may be a safer alternative for this.

bre1470: It is needed for nearly every function doclet, because the `this` parameter has to be defined for TypeScript but not for our documentation. Midterm we will switch from series factory to regular class and then it is not necessary anymore to turn it off.

PR #11354, opened 2019-07-09
 bugfix/11344-stock-icons-iconsURL

### LINKED ISSUE #11344: Some stock-icons always served from CDN
#### Expected behaviour
Setting the option `stockTools.gui.iconsURL`, should cause all icons to be served from my local server.

#### Actual behaviour
A few background-images in `popup.css` and `gui.css` are hardcoded to be served from `https://code.highcharts.com/7.1.2/gfx/stock-icons/`

#### Product version
Highstock 7.1.2

Perhaps these background images can use relative URLs?  Or be overlayed with JS instead of CSS `background`?

sebastianbochan:
Fixed #11344, some icons were not loaded from custom url.

### DIFF
--- a/js/annotations/popup.js
+++ b/js/annotations/popup.js
@@ -49,8 +49,8 @@ wrap(H.Pointer.prototype, 'onContainerMouseDown', function (proceed, e) {
     }
 });
 
-H.Popup = function (parentDiv) {
-    this.init(parentDiv);
+H.Popup = function (chart) {
+    this.init(chart);
 };
 
 H.Popup.prototype = {
@@ -60,14 +60,15 @@ H.Popup.prototype = {
      * @param {HTMLDOMElement} - container where popup should be placed
      * @return {HTMLDOMElement} - return created popup's div
      */
-    init: function (parentDiv) {
+    init: function (chart) {
 
         // create popup div
         this.container = createElement(DIV, {
             className: PREFIX + 'popup'
-        }, null, parentDiv);
+        }, null, chart.container);
 
         this.lang = this.getLangpack();
+        this.iconsURL = chart.options.navigation.annotationsPopup.iconsURL;
 
         // add close button
         this.addCloseBtn();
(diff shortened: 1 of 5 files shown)


js/annotations/popup.js
```diff
@@ -60,14 +60,15 @@ H.Popup.prototype = {
...
+    init: function (chart) {
 
         // create popup div
         this.container = createElement(DIV, {
             className: PREFIX + 'popup'
-        }, null, parentDiv);
+        }, null, chart.container);
 
         this.lang = this.getLangpack();
+        this.iconsURL = chart.options.navigation.annotationsPopup.iconsURL;
```
jon-a-nygaard: Could the `iconsURL` maybe be an option for the `Popup` constructor?
Then `Popup` would be more flexible in my opinion as it does not require a whole chart object, but just the `parentDiv` and the `iconsURL`. E.g `new Popup({ parentDiv, iconsURL })` .

pawelfus: I think just config object would be nice. At this moment we have only `iconsURL` but in future, we may expect more options to pop up. Like this `new Popup({ parentDiv, { iconsURL: ..., optionA: ..., optionB: ... } })`. And user config would be: `chart.options.navigation.annotationsPopup = { iconsURL, optionA, optionB }`.

jon-a-nygaard: @oysteinmoseng had a good point that keeping `parentDiv` as first parameter would make it consistent with our chart constructors. Then it would look like `new Popup(parentDiv, options)`.
It has an additional parameter, but I agree that this looks like a good solution.

What do you guys think @pawelfus and @sebastianbochan?

pawelfus: Yes, of course! Just realized that I accidentally copied `new Popup({ ... , ... })`, instead of `new Popup ( ..., ... )` :)

jon-a-nygaard: Great 👍 If @sebastianbochan also agrees, then we have a plan.

sebastianbochan: Make sense for me.

sebastianbochan: Ah, I commited changes before your discussion :) 
Will change, because config sounds really better.

PR #11357, opened 2019-07-09
bugfix/11246-crosshair-stroke-width

### LINKED ISSUE #11246: Styled mode: Invalid xAxis crosshair's width when xAxis has categories
#### Expected behaviour

If the xAxis has categories, and `crosshair` is set to `true`, then the crosshair's width should be one category, as specified in the [API](https://api.highcharts.com/highcharts/xAxis.crosshair.width).

That doesn't works as expected when in styled mode (but it works when NOT in styled mode).

#### Actual behaviour

When in styled mode, the crosshair's width is 1px instead of one category.

- 6.2.0 version: crosshair width is one category: http://jsfiddle.net/Metoule/2wp19qda/2/
- 7.1.2 version: crosshair width is 1px: http://jsfiddle.net/Metoule/2wp19qda/3/

#### Product version

Highcharts >= 7.0.0

#### Affected browser(s)

All

Basts:
Fixed #11246, setting crosshair stroke-width in styledMode. Additionally repairing windbarb and vector series in styledMode.

### DIFF
--- a/css/highcharts.scss
+++ b/css/highcharts.scss
@@ -266,8 +266,9 @@ g.highcharts-series,
     stroke-width: 0;
 }
 .highcharts-markers {
-    stroke-width: 1px;
-    stroke: $background-color;
+    stroke-width: 2px;
+    stroke: #000000;
+    fill: #cccccc;
 }
 .highcharts-point {
     stroke-width: 1px;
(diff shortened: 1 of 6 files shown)


css/highcharts.scss
```diff
@@ -266,8 +266,9 @@ g.highcharts-series,
     stroke-width: 0;
 }
 .highcharts-markers {
-    stroke-width: 1px;
-    stroke: $background-color;
+    stroke-width: 2px;
+    stroke: #000000;
+    fill: #cccccc;
```
pawelfus: Why do we need to change default styles for all markers?

PR #11385, opened 2019-07-12
bugfix/10980-stocktools-single-item

### LINKED ISSUE #10980: Unable to define single items in stock tools gui definitions
#### Actual behaviour

Error:

`Uncaught TypeError: Cannot read property 'addEventListener' of undefined`

https://jsfiddle.net/BlackLabel/gp49smqc/

#### Product version
Highstock v7.1.2

#### Affected browser(s)
All

pawelfus:
Fixed #10980, defining just one item in `stockTools.gui.definitions[name]` threw errors.
___
I went the simpler route - I let render just one submenu button so errors are not thrown. Alternative approach is to replace button+submenu with that item (of course when exactly one item is set). What do you think @sebastianbochan ?

### DIFF
--- a/js/modules/stock-tools-gui.js
+++ b/js/modules/stock-tools-gui.js
@@ -1082,7 +1082,7 @@ H.Toolbar.prototype = {
 
 
         // submenu
-        if (items && items.length > 1) {
+        if (items && items.length) {
 
             // arrow is a hook to show / hide submenu
             submenuArrow = createElement(SPAN, {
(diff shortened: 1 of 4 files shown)


oysteinmoseng:
Awaiting review from @sebastianbochan

sebastianbochan:
The items array is declared for submenu, so if someone degines a single element, it general, it does not make sense for submenus. 

TD;LR The solution is simple and make sense for me.

PR #11395, opened 2019-07-15
bugfix/11389-inactive-styled-mode-nodes

### LINKED ISSUE #11389: Dependency wheel styles are missing from the styled mode version
#### Expected behaviour

Using dependency wheel with styled mode on an off results in exactly the same behaviour on hover.

#### Actual behaviour

If styled mode is enabled, the chords are not dimmed and are rendered on top of each other at the same opacity. Upon opening the Dev Tools, and hovering on each of them, we can see that the classes `highcharts-point-hover` and `highcharts-point-inactive` are applied, but to no effect.

Since I use other charts with styled mode enabled in the same page, disabling it for just the wheel results in broken behaviour.

These codepens demo the issue: [styled mode off](https://codepen.io/anon/pen/WqWYOZ), [styled mode on](https://codepen.io/anon/pen/jjRQzz), and [styled mode off but with the css loaded](https://codepen.io/anon/pen/JQVeZP).

In the codepen for the styled mode, I added four CSS classes to demo a possible fix.

#### Product version

Highcharts 7.1.2

#### Affected browser(s)

Tested in Firefox 70.0a1 and Vivaldi 2.6.1566.44 (Chrome/75.0.3770.102).

pawelfus:
Fixed #11389, inactive styles for dependency wheel series in styled mode were missing.
___
Fixed styles and refactored a bit (DRY). ~Is that safe to use `getComputedStyle` in tests? Some tests use that (which is not supported in IE8).~ (nvm, removed test due to Karma issues)

Edit: Added dependency wheel to karma-files config.


ts/parts/Tooltip.ts
```diff
@@ -1049,19 +1049,7 @@ H.Tooltip.prototype = {
...
-                inactiveSeries: Highcharts.Series
-            ): void {
-                if (
-                    inactiveSeries.options.inactiveOtherPoints ||
-                    activeSeries.indexOf(inactiveSeries) === -1
-                ) {
-                    inactiveSeries.setState('inactive', true);
-                }
-            });
+            chart.pointer.setInactiveSeries(point as Array<Highcharts.Point>);
```
bre1470: The cast should not be necessary if the outer conditional is type complete. If it is incomplete please cast to any, so that we can pick it up later.

pawelfus: Yup, I was really curious why casting to `any` when I explicite know the type?

bre1470: Cast to any will be later in the migration process forbidden and is kind of our todo type, so that the compiler gets quiet.

pawelfus: Okay! Let me change then, thanks :)

pawelfus:
It's strange, tests are fine on my local version, though here fails.. will try tomorrow with a separate branch.

PR #11398, opened 2019-07-15
enhancement/8356-plotlines-plotbands-label-formatter

kamilkulig:
Added support for `label.formatter` callback for plot lines and plot bands.


TorsteinHonsi:
@bre1470 Any idea why the TS declarations fail in this PR?

PR #11400, opened 2019-07-15
bugfix/8608-missing-bubble-zparam

### LINKED ISSUE #8608: Bubble chart not seen properly when z param is not set in datapoint

sebastianbochan:
Fixed #8608, bubble markers were hidden when z param missing.

### DIFF
--- a/ts/parts-more/BubbleSeries.ts
+++ b/ts/parts-more/BubbleSeries.ts
@@ -78,8 +78,8 @@ declare global {
                 zMax: number,
                 minSize: number,
                 maxSize: number,
-                value: number,
-                yValue?: number
+                value: (number|null|undefined),
+                yValue?: (number|null|undefined)
             ): (number|null);
             public hasData(): boolean;
             public pointAttribs(
@@ -436,8 +436,8 @@ seriesType<Highcharts.BubbleSeriesOptions>('bubble', 'scatter', {
                 zMax,
                 minSize as any,
                 maxSize as any,
-                value as any,
-                yData && yData[i] as any
+                value,
+                yData[i]
             ));
         }
         this.radii = radii;
@@ -453,19 +453,19 @@ seriesType<Highcharts.BubbleSeriesOptions>('bubble', 'scatter', {
         zMax: number,
         minSize: number,
         maxSize: number,
-        value: number,
-        yValue?: number
+        value: (number|null|undefined),
+        yValue?: (number|null|undefined)
     ): (number|null) {
         var options = this.options,
             sizeByArea = options.sizeBy !== 'width',
             zThreshold = options.zThreshold,
-            pos,
             zRange = zMax - zMin,
+            pos = 0.5,
             radius;
 
         // When sizing by threshold, the absolute value of z determines
         // the size of the bubble.
-        if (options.sizeByAbsoluteValue && value !== null) {
+        /*if (options.sizeByAbsoluteValue && isNumber(value)) {
             value = Math.abs(value - (zThreshold as any));
             zMax = zRange = Math.max(
                 zMax - (zThreshold as any),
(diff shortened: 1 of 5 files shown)


ts/parts-more/BubbleSeries.ts
```diff
@@ -468,7 +474,8 @@ seriesType<Highcharts.BubbleSeriesOptions>('bubble', 'scatter', {
             zMin = 0;
         }
 
-        if (!isNumber(value)) {
+        // #8608 - bubble should be visible when z is undefined
+        if (yValue === null || value === null) {
```
jon-a-nygaard: Now that BubbleSeries is migrated to TypeScript this condition here becomes redundant in my understanding.

Since `value` is required and can only be a `number` and thus never `null`.
Since `yValue` is optional it can only be `number` or `undefined` and thus never `null`

Because of this TypeScript should alert us when we try to pass in `yValue` or `value` with the type of `null`, and we could in my opinion remove this condition.

sebastianbochan: IMHO its safer to leave this condition.

What do you think @oysteinmoseng @pawelfus ?

pawelfus: TS defs are wrong:
> public yData: Array<number>;

Should be:
> public yData: Array<(number|undefined|null)>;

The same for `zData`.

Demo: https://jsfiddle.net/BlackLabel/ek83tch9/

ts/parts-more/BubbleSeries.ts
```diff
@@ -41,6 +41,7 @@ declare global {
             minPxSize?: BubbleSeries['minPxSize'];
             specialGroup?: BubbleSeries['specialGroup'];
             zData?: BubbleSeries['zData'];
+            yData?: BubbleSeries['yData'];
```
bre1470: You should defined yData as a permanent property in the BubbleSeries class (like zData), otherwise you are here actually referencing to the base Series class, because there is no further clarification in BubbleSeries.

ts/parts-more/BubbleSeries.ts
jon-a-nygaard: Also, the TypeScript defs for `getRadius` has to be changed to `value: number|null|undefined` and `yValue?: number|null` for this to be resolved. When this is changed the casting of `value` and `yValue` to `any` in `getRadii` can be removed again.

sebastianbochan:
@oysteinmoseng @bre1470 could you verify if now all is correct?

ts/parts-more/BubbleSeries.ts
bre1470: Looks good!

ts/parts-more/BubbleSeries.ts
```diff
@@ -448,7 +453,8 @@ seriesType<Highcharts.BubbleSeriesOptions>('bubble', 'scatter', {
         zMax: number,
         minSize: number,
         maxSize: number,
-        value: number
+        value: number,
```
jon-a-nygaard: @sebastianbochan Please see https://github.com/highcharts/highcharts/pull/11400#discussion_r304253652

If yData and zData can be `Array<number|null|undefined>`, then `getRadius` has to support `number|null|undefined` types for `value` and `yValue` as well.

We should avoid casting to `any` as we at some point will have to remove all `any`'s when we switch to strict mode in TypeScript.

bre1470: I am not sure if changes to `yData` imply a change for `getRadius`, but I agree, that we should check for defined types as much as possible to avoid `any` casts.

sebastianbochan: Ok but when I use value as <number|null|undefined> then the problem is here:

```
if (yValue === null || value === null) {
            radius = null;
        // Issue #4419 - if value is less than zMin, push a radius that's
        // always smaller than the minimum size
        } else if (value < zMin) {
            radius = minSize / 2 - 1;
        } else {
            // Relative size, a number between 0 and 1
            pos = zRange > 0 ? (value - zMin) / zRange : 0.5;

            if (sizeByArea && pos >= 0) {
                pos = Math.sqrt(pos);
            }
            radius = Math.ceil(minSize + pos * (maxSize - minSize)) / 2;
        }
```

Lets analyze the second IF and ELSE. Linter blocks me because of "value can be undefined", but in JS it works properly. 

Morever other parametrs like `minSize` and `maxSize` can be declared as `any`. 

@bre1470 what do you think? Should we fix that or leave like it is?

jon-a-nygaard: @sebastianbochan  What should `getRadius` return if `value=undefined`?
My first thought was that it could either return `null` or maybe `0`.

If `null` is the correct return value, then this is solved by doing:
```
if (yValue === null || value === null || value === undefined) {
  radius = null;
} else if ...
```

sebastianbochan: Completely no.

If `value` is `undefined` it means that only z param is not defined (by user), but bubble should be visible because y value is declared. This is the title of bug.

jon-a-nygaard: So what value should it return in this case? 🤔 Currently it will return `NaN` which although works somehow, I believe it is not a good return value.

bre1470: Maybe the condition has to be more specific. Something like `(value === void 0 /* undefined */ && !U.defined(yValue))`?

sebastianbochan: @jon-a-nygaard 
Lets analyze all step by step.

When `value` is undefined the `zRange` is -Infinity. In this condition: ` pos = zRange > 0 ? (value - zMin) / zRange : 0.5;` value is not used and `pos = 0.5`. As a result we do not have NaN anywhere and radius is calculated correctly.

Do you agree with me?

jon-a-nygaard: Sorry @sebastianbochan, but I do not. It may end up well in the case you are testing, but `getRadius` does not cover all possible situations at the moment.

Please see the following example of how `radius` could end up as `NaN`, open the console to view steps and values: https://jsfiddle.net/jon_a_nygaard/wut3moxz/

This is also the reason why TypeScript complains, because we are doing math operations on variables that do not have the type `number`.

But if I understand your last comment correctly, then maybe we could say that if `value===undefined` then `pos=0.5` to achieve the desired result?

jon-a-nygaard:
Removed a forgotten variable in `getRadius`, otherwise I think it looks good 👍 

Will do the merge when tests have completed.

PR #11410, opened 2019-07-16
bugfix/11389-inactive-styledmode-nodes

### LINKED ISSUE #11389: Dependency wheel styles are missing from the styled mode version

pawelfus:
Fixed #11389, inactive styles for dependency wheel series in styled mode were missing.
___
Refactored a bit + added dependency wheel to karma-files config. Let's hope tests will pass in CircleCI..

### DIFF
--- a/ts/parts/Interaction.ts
+++ b/ts/parts/Interaction.ts
@@ -77,8 +77,8 @@ declare global {
             onMouseOut(): void;
             onMouseOver(): void;
             select(selected?: boolean): void;
+            setAllPointsToState(state?: string): void;
             setState(state?: string, inherit?: boolean): void;
-            setAllPointsState(state?: string): void;
             setVisible(visible?: boolean, redraw?: boolean): void;
             show(): void;
         }
@@ -1526,21 +1526,21 @@ extend(Series.prototype, /** @lends Highcharts.Series.prototype */ {
         // Don't loop over points on a series that doesn't apply inactive state
         // to siblings markers (e.g. line, column)
         if (inherit && inactiveOtherPoints && series.points) {
-            series.setAllPointsState(state);
+            series.setAllPointsToState(state);
         }
     },
 
     /**
      * Set the state for all points in the series.
      *
-     * @function Highcharts.Series#setAllPointsState
+     * @function Highcharts.Series#setAllPointsToState
      *
      * @private
      *
      * @param {string} [state]
      *        Can be either `hover` or undefined to set to normal state.
      */
-    setAllPointsState: function (
+    setAllPointsToState: function (
         this: Highcharts.Series,
         state?: string
     ): void {
--- a/ts/parts/Interaction.ts
+++ b/ts/parts/Interaction.ts
@@ -78,6 +78,7 @@ declare global {
             onMouseOver(): void;
             select(selected?: boolean): void;
             setState(state?: string, inherit?: boolean): void;
+            setAllPointsState(state?: string): void;
             setVisible(visible?: boolean, redraw?: boolean): void;
             show(): void;
         }
@@ -1525,12 +1526,29 @@ extend(Series.prototype, /** @lends Highcharts.Series.prototype */ {
         // Don't loop over points on a series that doesn't apply inactive state
         // to siblings markers (e.g. line, column)
         if (inherit && inactiveOtherPoints && series.points) {
-            series.points.forEach(function (point: Highcharts.Point): void {
-                if (point.setState) {
-                    point.setState(state);
-                }
-            });
+            series.setAllPointsState(state);
         }
+    },
+
+    /**
+     * Set the state for all points in the series.
+     *
+     * @function Highcharts.Series#setAllPointsState
+     *
+     * @private
+     *
+     * @param {string} [state]
+     *        Can be either `hover` or undefined to set to normal state.
+     */
+    setAllPointsState: function (
+        this: Highcharts.Series,
+        state?: string
+    ): void {
+        this.points.forEach(function (point: Highcharts.Point): void {
+            if (point.setState) {
+                point.setState(state);
+            }
+        });
     },
 
     /**
--- a/ts/parts/Pointer.ts
+++ b/ts/parts/Pointer.ts
@@ -69,6 +69,7 @@ declare global {
             public options: Options;
             public pinchDown: Array<any>;
             public runChartClick: boolean;
+            public applyInactiveState(points: Array<Point>): void;
             public destroy(): void;
             public drag(e: PointerEventObject): void;
             public dragStart(e: PointerEventObject): void;
@@ -78,7 +79,6 @@ declare global {
                 shared: (boolean|undefined),
                 e: PointerEventObject
             ): (Point|undefined);
-            public getActiveSeries(points: Array<Point>): Array<Series>;
             public getChartCoordinatesFromPoint(
                 point: Point,
                 inverted?: boolean
@@ -118,7 +118,6 @@ declare global {
                 clip?: boolean
             ): void;
             public setDOMEvents(): void;
-            public setInactiveSeries(points: Array<Point>): void;
             public zoomOption(e: Event): void;
         }
         let chartCount: number;
@@ -758,7 +757,7 @@ Highcharts.Pointer.prototype = {
                 hoverSeries.onMouseOver();
             }
 
-            pointer.setInactiveSeries(points);
+            pointer.applyInactiveState(points);
 
             // Do mouseover on all points (#3919, #3985, #4410, #5622)
             (points || []).forEach(function (p: Highcharts.Point): void {
@@ -848,7 +847,7 @@ Highcharts.Pointer.prototype = {
      * or, if `inactiveOtherPoints` is set to true, set inactive state to
      * all points within that series.
      *
-     * @function Highcharts.Pointer#setInactiveSeries
+     * @function Highcharts.Pointer#applyInactiveState
      *
      * @private
      *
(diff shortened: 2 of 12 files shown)


pawelfus:
Ok, no tests then, I'll remove them in the next commit.

ts/parts/Pointer.ts
```diff
@@ -854,6 +843,39 @@ Highcharts.Pointer.prototype = {
...
+     *
+     * @private
+     *
+     * @param {Array<Highcharts.Point>} points
+     *        Currently hovered points
+     *
+     */
+    setInactiveSeries: function (
+        this: Highcharts.Pointer,
+        points: Array<Highcharts.Point>
```
bre1470: As indicated in the corresponding description, this should also support Series beside Point. It is a bit strange to have a function `setInactiveSeries` that takes points only.

bre1470:
> We should rename the `setInactiveSeries` function or allow array of series as the argument.

That is my humbled opinion.

ts/parts/Interaction.ts
```diff
@@ -1525,14 +1526,31 @@ extend(Series.prototype, /** @lends Highcharts.Series.prototype */ {
...
+     * Set the state for all points in the series.
+     *
+     * @function Highcharts.Series#setAllPointsState
+     *
+     * @private
+     *
+     * @param {string} [state]
+     *        Can be either `hover` or undefined to set to normal state.
+     */
+    setAllPointsState: function (
```
jon-a-nygaard: Do we really need to add this function to the series prototype? I feel that it is already quite bloated with utility functions.
Since this is used only in one place, at least for now, it could be an internal utility function in the same file.

ts/parts/Pointer.ts
```diff
@@ -854,6 +843,39 @@ Highcharts.Pointer.prototype = {
...
+     *
+     * @function Highcharts.Pointer#setInactiveSeries
+     *
+     * @private
+     *
+     * @param {Array<Highcharts.Point>} points
+     *        Currently hovered points
+     *
+     */
+    setInactiveSeries: function (
```
jon-a-nygaard: Same concern with prototype as for `Highcharts.Series#setAllPointsState`

ts/parts/Pointer.ts
pawelfus: There's no need for `Series` array. From currently hovered points, retrieve all inactive series and set them inactive. You are right, we could use a better name - what do you suggest? `setInactiveSeriesFromPoints` is just too long. Description of the function is:
>     * Set inactive state to all series that are not currently hovered,
>     * or, if `inactiveOtherPoints` is set to true, set inactive state to
>     * all points within that series.

ts/parts/Interaction.ts
jon-a-nygaard: Maybe a bit verbose, but I would prefer it to be named `setAllPointsToState`.

ts/parts/Pointer.ts
bre1470: How about `applyInactiveState`?

jon-a-nygaard:
> Ok, no tests then, I'll remove them in the next commit.

Btw, do you @pawelfus know what is causing the test to fail on CircleCI? It looks good to me 🤔

ts/parts/Interaction.ts
pawelfus: Yes, we need this here. It's used in two files (`Pointer.js` and `Series.js`) and it's a very common task for series with `inactiveOtherPoints=true` (pie, sankey, networkgraph etc.) - it's called on every point/node/link/slice mouseOver/Out.

IMHO private functions are bad in a library like Highcharts - no way to wrap them, making customizations, plugins and test hard/impossible to write. There's no reason for protected/private functions in our case - we gain nothing, giving up flexibility.

Alternatively, we can create `Series.prototype.utils` object, like in `NavigationBindings` class. Maybe we should make this a common practice?

pawelfus:
> Btw, do you @pawelfus know what is causing the test to fail on CircleCI? It looks good to me 🤔

Hmm.. this is what I see:
Screen Shot 2019-07-17 at 10 40 28

Test that is failing is the one I added.

ts/parts/Pointer.ts
pawelfus: Doesn't `applyInactiveState` miss `Series` keyword then? Maybe something like hmm.. `inactivateSeries`? I don't know.. what do you think @jon-a-nygaard , @oysteinmoseng , @sebastianbochan ?

ts/parts/Interaction.ts
jon-a-nygaard: > Yes, we need this here. It's used in two files (Pointer.js and Series.js) and it's a very common task for series

Okay, then it makes sense to leave it as is.

jon-a-nygaard:
>Test that is failing is the one I added.

Yes, I noticed, but I had a look at the code for the test, and I do not understand why it would fail on CircleCI, as I believe we have similar async tests already. Would be nice if we are able to keep it somehow.

pawelfus:
> Would be nice if we are able to keep it somehow.

Any ideas how to debug it? Locally it works fine (both, karma and browser). In CircleCI it seems that JS files are not built or hovering a non existing point..

ts/parts/Pointer.ts
pawelfus: The same as `Highcharts.Series#setAllPointsState` - different files, affection for all series with `inactiveOtherPoints=true`.

ts/parts/Pointer.ts
jon-a-nygaard: I do not think `Series` should be included in the name since it may also set the state on points. I think I prefer @bre1470 suggestion of `applyInactiveState` or plural `applyInactiveStates`.

Alternatively `setInactiveStates`, but I do not like that much since it is not setting the state on the points we pass in, and also it does not seem much like a regular setter function.

jon-a-nygaard:
> Any ideas how to debug it?

I have no clue unfortunately, and Jan Martin is currently on vacation. Maybe you could just add `QUnit.skip` and an explanation to the test for now? Then it is at least registered as a test that needs our attention in the future.

bre1470:
> > Would be nice if we are able to keep it somehow.
> 
> Any ideas how to debug it? Locally it works fine (both, karma and browser). In CircleCI it seems that JS files are not built or hovering a non existing point..

I made a fresh clone with checkout and it fails for me local in headless Chrome.

bre1470:
and in utils it fails too. So make everything new and you can probably debug it.

ts/parts/Pointer.ts
pawelfus: You are right @bre1470 and @jon-a-nygaard - my mistake, where were my eyes.. 👀 Fix coming soon!

pawelfus:
@jon-a-nygaard - refactored as you and @bre1470 suggested. Thanks!

Found the root of the failing commit:
In Karma tests we run a few additional tests: `templates`. Those tests create charts that exist through the entire testing process. In my test I used `document.querySelectorAll('.highcharts-point-inactive');` - which returned more points than expected (some pie chart was interfering). I simply added a class name to the chart to avoid this.

@bre1470 - any thoughts about this problem? Can we somehow resolve this, so other devs won't fail into this trap? :)

bre1470:
@pawelfus  Actually, that is the purpose of the templates to not destroy charts between tests. I wonder, why you do not do a `querySelectorAll` on the chart container?

pawelfus:
> Actually, that is the purpose of the templates to not destroy charts between tests.

Yes, I realized that :) 
> I wonder, why you do not querySelectorAll on the chart container?

It seems that charts were created in the same container (or on top of each other). I didn't dig deeper, but in general TestController fired events for unrelated charts.

Still, can we do something about this? Perhaps adding info in docs? Maybe here:
- https://github.com/highcharts/highcharts/tree/master/test#features
- https://github.com/highcharts/highcharts/tree/master/samples#tests

bre1470:
The template containers are a fixed stack, but they only lay over your template-less chart, if they get used (positive z-index). When to many events are firing, maybe something in Highcharts is going wrong.

The same situation would be in RL like having a webpage with several charts and a popup where another chart is available. Now when you interact with the chart in the popup, there should be no events firing on the chart underneath imho.

pawelfus:
> Now when you interact with the chart in the popup, there should be no events firing on the chart underneath imho.

That's how it works, see: https://jsfiddle.net/BlackLabel/zv75g6wm/
I'm not sure what do you meant with to many events - my test has just one event `controller.mouseOver(x, y);`, could you elaborate?

bre1470:
TestController is working by getting the element on top and firing there your requested event. As I understood you, the problem for you is, that events from other charts are interfering your test and your solution was to check for the SVG output. My point was, that something is wrong with Highcharts, if unexpected events on other charts get fired.

PR #11411, opened 2019-07-16
bugfix/11388-histogram-error-nodata

### LINKED ISSUE #11388: Error when try to generate histogram with empty array passed as data.
#### Expected behaviour
Chart should generate regardless the data array is empty or not.

#### Actual behaviour
There is error thrown in console, when trying to create the chart with histogram series, and an empty data array is passed through `baseSeries`'s configuration.

```
VM693 highcharts.js:23 Uncaught TypeError: Cannot read property 'toPrecision' of undefined
    at a.correctFloat (VM693 highcharts.js:23)
    at f.derivedData (VM694 histogram-bellcurve.js:13)
    at f.setDerivedData (VM694 histogram-bellcurve.js:12)
    at a.Chart.<anonymous> (VM694 histogram-bellcurve.js:10)
    at VM693 highcharts.js:31
    at Object.a.each (VM693 highcharts.js:28)
    at a.fireEvent (VM693 highcharts.js:31)
    at a.Chart.linkSeries (VM693 highcharts.js:267)
    at a.Chart.firstRender (VM693 highcharts.js:272)
```

https://jsfiddle.net/BlackLabel/58fpota6/

#### Product version
v6.1.2 and above
Regression after: https://github.com/highcharts/highcharts/commit/8abd3065a

#### Suggested workaround
Just need to check whether baseSeries data is not empty:
```js
(function(H) {
    H.wrap(H.seriesTypes.histogram.prototype, 'setDerivedData', function(proceed) {
        if(this.baseSeries.yData.length !== 0) {
            proceed.apply(this)
        }
    })
})(Highcharts)
```
Live example with fix applied: https://jsfiddle.net/BlackLabel/u9p6fk5m/

sebastianbochan:
Fixed #11388, error was return when empty histogram data.

### DIFF
--- a/js/modules/histogram.src.js
+++ b/js/modules/histogram.src.js
@@ -125,8 +125,14 @@ seriesType(
     },
     merge(derivedSeriesMixin, {
         setDerivedData: function () {
+            var yData = this.baseSeries.yData;
+
+            if (!yData.length) {
+                return;
+            }
+
             var data = this.derivedData(
-                this.baseSeries.yData,
+                yData,
                 this.binsNumber(),
                 this.options.binWidth
             );
--- a/samples/unit-tests/series-histogram/members/demo.details
+++ b/samples/unit-tests/series-histogram/members/demo.details
@@ -1,0 +1,6 @@
+---
+ resources:
+   - https://code.jquery.com/qunit/qunit-2.0.1.js
+   - https://code.jquery.com/qunit/qunit-2.0.1.css
+ js_wrap: b
+...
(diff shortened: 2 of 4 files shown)


jon-a-nygaard:
I realized that this is a case where we could add a simple unit test to avoid regressions in the future.

So before I merge it in I will quickly add a unit test that calls `Highcharts.seriesTypes.histogram.prototype.setDerivedData` with a context that has an empty `baseSeries.yData`.

sebastianbochan:
Thank you @jon-a-nygaard.

Issue #11366, opened 2019-07-10
Stock-tools horizontal-line cursor / annotationOptions


3Ue:
#### Expected behaviour
Cursor for horizontal line from stock-tools should be set by default to 'ns-resize' and for vertical line to 'ew-resize'.

Manually setting `navigation.bindings.horizontalLine.annotationOptions.draggable` to 'y' should work.

#### Actual behaviour
Cursor for horizontal- and vertical line from stock-tools are set by default to 'move'.

Manually setting `navigation.bindings.horizontalLine.annotationOptions.draggable` to 'y' does not work. Instead `navigation.bindings.infinityLine.annotationsOptions.draggable` is used also for horizontalLine (note the difference between the documented `annotationOptions` and the actual implementation `annotationsOptions`).

#### Live demo with steps to reproduce
https://jsfiddle.net/geshnqbv/

1. Add horizontal line over stock-tools
2. Hover mouse over the line

#### Product version
Highstock 7.1.2

#### Affected browser(s)
FF, probably all

pawelfus:
Thanks for reporting the issue @3Ue 

It seems you found a few issues in one. Bug in docs (should be `annotationsOptions`), wrong type config merge (`horizontalLine` vs `infinityLine`) and lastly missing defaults for `draggable` vertical/horizontal lines.

PR #11416, opened 2019-07-17
feature/11175-rect-annotation-value-coords

### LINKED ISSUE #11175: Stock tools - problem with responsive of area elements
#### Expected behaviour

All annotations should be related to an axis scale.

#### Actual behaviour

Zrzut ekranu 2019-06-26 o 10 59 50

Zrzut ekranu 2019-06-26 o 11 00 30

In the above screenshots we can see that 'Line' and 'Label' annotations, as opposed to 'Circle' and 'Reactangle', are located correctly and keep the scale of the axes.

https://jsfiddle.net/BlackLabel/7zfjy0sc/

#### Product version
Highstock v7.1.2

#### Affected browser(s)
All

pawelfus:
Fixed #11175, `navigation.bindings.rectangleAnnotation` did not resize it's width/height according to xAxis/yAxis extremes.
___
Note that these two commits are expected here. I realized that #11366 was not fixed for basic annotations (but Stock only) so I added here since it is for the fix.

Issue related to Cloud.

I'm not sure if this is a feature or a bugfix - please mark this as you prefer. From my perspective it's a feature (so rectangle coordinates are bound to x/y values), but for a user it may be a bug.

### DIFF
--- a/js/annotations/navigationBindings.js
+++ b/js/annotations/navigationBindings.js
@@ -993,7 +993,7 @@ H.setOptions({
              * `steps` array.
              *
              * @type    {Highcharts.StockToolsBindingsObject}
-             * @default {"className": "highcharts-circle-annotation", "start": function() {}, "steps": [function() {}], "annotationOptions": {}}
+             * @default {"className": "highcharts-circle-annotation", "start": function() {}, "steps": [function() {}], "annotationsOptions": {}}
              */
             circleAnnotation: {
                 /** @ignore */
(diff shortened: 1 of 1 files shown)


js/annotations/navigationBindings.js
```diff
@@ -1098,101 +1102,111 @@ H.setOptions({
...
 
                         annotation.update({
                             shapes: [{
-                                width: width,
-                                height: height,
-                                point: {
-                                    x: point.x,
-                                    y: point.y
-                                }
+                                points: points
```
jon-a-nygaard: Will the change from multiple properties to the property `points` cause a breaking change in the `Annotation` API?
If so is this something that can cause issues for our users?

pawelfus: No, not at all. This change is done only for bindings - not for Annotations API. Bindings: [open demo](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/annotations/gui-buttons/) and add a rectangle - all new created annotations via this button will have different config. Old config is still valid in terms of Annotations API (so loading it from backend will render an annotation anyway).

jon-a-nygaard: Thanks for the clarification 👍

PR #11421, opened 2019-07-17
docs/fix-formatter

bre1470:
Fixed formatter types and doclets. Tooltip types should show up in class references.

### DIFF
--- a/ts/parts/Tooltip.ts
+++ b/ts/parts/Tooltip.ts
@@ -158,11 +158,11 @@ declare global {
  *
  * @param {Highcharts.TooltipFormatterContextObject} this
  *        Context to format
-
+ *
  * @param {Highcharts.Tooltip} tooltip
  *        The tooltip instance
  *
- * @return {false|string|Array<string>}
+ * @return {false|string|Array<(string|null|undefined)>|null|undefined}
  *         Formatted text or false
  */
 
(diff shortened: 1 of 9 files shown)


ts/parts/Tooltip.ts
```diff
@@ -247,6 +247,8 @@ declare global {
  * @typedef {"callout"|"circle"|"square"} Highcharts.TooltipShapeValue
  */
 
+''; // separates doclets above from variables below
```
pawelfus: ?

bre1470: Yeah. That was interesting. All the doclets above where attached to the variable below, which was internal and filtered out. Therefor no interface was going into the class reference. :)

jon-a-nygaard: As I understand it is a hack to make sure that the doclets is picked up by JSDoc without the variables below interfering with the result.

Maybe you can confirm or correct me on this @bre1470?

pawelfus: I guess simple:

// ---- doclets separator ----

Won't work? :) Maybe it's better to place doclets at the end of the file? And make this a general rule? I'm worried that in future someone else will spot the same issue, wasting time to find answer "whyyyyy?"

PR #11422, opened 2019-07-17
vacation2019/ts/mapbubbleseries

### LINKED ISSUE #11057: Migrate `parts-map/MapBubbleSeries` to TypeScript

bre1470:
Fixed #11057 - Migrated MapBubbleSeries to TypeScript.


ts/parts-map/MapSeries.ts
```diff
@@ -1317,7 +1317,7 @@ seriesType<Highcharts.MapSeriesOptions>(
                     (series.mapMap as any)[(point as any)[joinBy[1]]];
                 if (mapPoint) {
                     // This applies only to bubbles
-                    if ((series as Highcharts.MapBubbleSeries).xyFromShape) {
+                    if ((series as any).xyFromShape) {
```
jon-a-nygaard: Why is it necessary to cast to `any` in this case?

bre1470: Because `MapSeries` can never become `MapBubbleSeries`, because `MapBubbleSeries` is based on `BubbleSeries`. I wonder, if this gets actually ever triggered.

jon-a-nygaard: Before I merge the PR, can you confirm that this is correct @oysteinmoseng?

jon-a-nygaard: @bre1470 I had a chat with @oysteinmoseng and we found that this condition is triggered if the following gets called: https://github.com/highcharts/highcharts/blob/2a5bfe812ffbdd2e1f808bdfe67cfac6a6cb558a/ts/parts-map/MapBubbleSeries.ts#L249-L252

I am not sure how to solve the types to handle this case.

bre1470: So it is misuse of mixin from another series prototype. We will have to solve this when removing the any cast.

jon-a-nygaard: Okay, then we'll leave the `any` for now.

PR #11435, opened 2019-07-18
vacation2019/ts/networkgraph-quadtree

### LINKED ISSUE #11430: Migrate `modules/networkgraph/QuadTree` to TypeScript

bre1470:
Fixed #11430 - Migrated QuadTree to TypeScript.


ts/modules/networkgraph/QuadTree.ts
```diff
@@ -0,0 +1,517 @@
...
+            public visitNodeRecursive(
+                node: (QuadTreeNode|null),
+                beforeCallback: (Function|null),
+                afterCallback: (Function|null),
+                chart?: Chart,
+                clear?: boolean
+            ): void;
+        }
+        class QuadTreeNode {
+            public constructor(box: RectangleObject);
```
pawelfus: Why it's `box: RectangleObject` but a few lines below it's: `public box: Dictionary<number>;` ?

ts/modules/networkgraph/QuadTree.ts
```diff
@@ -0,0 +1,517 @@
...
+            public body: (boolean|Point);
+            public box: Dictionary<number>;
```
pawelfus: 1.1)

ts/modules/networkgraph/QuadTree.ts
```diff
@@ -0,0 +1,517 @@
...
+ *
+ * @private
+ * @class
+ * @name Highcharts.QuadTreeNode
+ *
+ * @param {Highcharts.Dictionary<number>} box Available space for the node
+ */
+var QuadTreeNode = H.QuadTreeNode = function (
+    this: Highcharts.QuadTreeNode,
+    box: Highcharts.Dictionary<number>
```
pawelfus: 1.2)

ts/modules/networkgraph/QuadTree.ts
bre1470: Originally you had set it to this type, but it is not working, because RectangleObject does have the `left` and `top` properties, that you are using. So I switch it everywhere to a simple number dictionary.

bre1470:
I will merge vacation2019 to check for conflicts.

ts/modules/networkgraph/QuadTree.ts
pawelfus: Ah, thanks for the explanation! Does that mean if I switch from `left`/`top` to `x`/`y` it will be fine?

bre1470: Yes, it would be fine, but you differentiate like this in other networkgraph components, so I actually believe that this is done by intention.

pawelfus: It's just my personal preference `left` over `x` and `top` over `y` (it's easy to mistake `x` with `y`, but not `left` with `top` ;) ).

PS: See my edit above

bre1470: Here is the other place, where you use left/top: https://github.com/highcharts/highcharts/blob/vacation2019/ts/modules/networkgraph/integrations.ts#L120-L122

The type has been changed. The code is simply outdated.

PR #11436, opened 2019-07-18
vacation2019/ts/networkgraph-draggable-nodes

### LINKED ISSUE #11431: Migrate `modules/networkgraph/draggable-nodes` to TypeScript

bre1470:
Fixed #11431 - Migrated DragNodesMixin to TypeScript.

### DIFF
--- a/js/modules/networkgraph/draggable-nodes.js
+++ b/js/modules/networkgraph/draggable-nodes.js
@@ -1,37 +1,33 @@
 /* *
- * Networkgraph series
  *
- * (c) 2010-2019 Paweł Fus
+ *  Networkgraph series
  *
- * License: www.highcharts.com/license
- */
-
+ *  (c) 2010-2019 Paweł Fus
+ *
+ *  License: www.highcharts.com/license
+ *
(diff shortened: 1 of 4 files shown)


js/modules/networkgraph/draggable-nodes.js
```diff
@@ -1,37 +1,33 @@
 /* *
- * Networkgraph series
  *
- * (c) 2010-2019 Paweł Fus
+ *  Networkgraph series
```
pawelfus: Actually, PackedBubble series also uses this module. Should move this to mixins or modules?

ts/modules/networkgraph/draggable-nodes.ts
```diff
@@ -0,0 +1,253 @@
+/* *
+ *
+ *  Networkgraph series
```
pawelfus: Actually, PackedBubble series also uses this module. Should move this to mixins or modules?

bre1470: Can we delay this until after this migration phase?

pawelfus: Yes, of course! I'm just thinking out loud :)

PR #11444, opened 2019-07-18
Expose the ESM entry point

TehShrike:
Exposes the standard `module` entry point for ESM-aware bundlers, utilities like @pika/web, and helper sites like unpkg.com and bundle.run

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "7.1.2-modified",
   "description": "JavaScript charting framework",
   "main": "lib/highcharts",
+  "module": "es-modules/masters/highcharts.src.js",
   "author": "Highsoft AS <support@highcharts.com> (http://www.highcharts.com/about)",
   "repository": {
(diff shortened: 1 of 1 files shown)


TehShrike:
Sadly, the exports design goals included [having no overlap with the current ecosystem usage](https://github.com/jkrems/proposal-pkg-exports/issues/5#issue-378356390).

I suspect that we're entering a world where "module" will mean "ESM entry point for browser bundlers" and "type: module" means "entry points are all ESM for node.js", at least for the foreseeable future, which seems confusing and unnecessary.

Oh well

https://xkcd.com/927/

jkrems:
For context: The issue with `module` is that it's using CommonJS semantics but ESM syntax which is something that works only in bundlers like webpack. I don't think there's many packages using `module` that would be compatible with native ESM (browser or node). So we were unable to use that name because everything would've broken immediately. :(

PR #11446, opened 2019-07-19
bugfix/svgrenderer-symbol-demo

KacperMadej:
Improved tests for svgrenderer/symbol to prevent errors on different browsers - Chrome on Windows and Safari on Mac.

### DIFF
--- a/samples/unit-tests/svgrenderer/symbol/demo.js
+++ b/samples/unit-tests/svgrenderer/symbol/demo.js
@@ -4,17 +4,14 @@ QUnit.test('Symbol tests', function (assert) {
         h = 400,
         done = assert.async(),
         total = 0,
-        count = 0;
-
-    var ren = new Highcharts.Renderer(
-        document.getElementById('container'),
-        w,
-        h
-    );
-
-    var url = location.host.substr(0, 12) === 'localhost:98' ?
-        'url(base/test/testimage.png)' : // karma
-        'url(testimage.png)'; // utils
+        count = 0,
+        symbol1,
+        symbol2,
+        symbol3,
+        label,
+        url = location.host.substr(0, 12) === 'localhost:98' ?
+            'url(base/test/testimage.png)' : // karma
+            'url(testimage.png)'; // utils
 
     function ifDone() {
         count++;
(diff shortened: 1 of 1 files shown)


samples/unit-tests/svgrenderer/symbol/demo.js
```diff
@@ -23,117 +20,128 @@ QUnit.test('Symbol tests', function (assert) {
...
+                        .symbol('krakow', 100, 200, 10, 10)
+                        .attr({
+                            fill: 'red'
+                        })
+                        .add();
+                    total++;
+
+                },
+                // images are loaded
+                load: function () {
```
jon-a-nygaard: I understand why you use the `Highcharts.load` event here, but I would prefer if we could avoid constructing a chart to test the the SvgRenderer.

Is it possible to solve this using `Highcharts.addEvent` and by listening for the load event on the individual images? E.g:
```javascript
const onLoad = () => {
    console.log('loaded')
};
const waitForLoad = (function (onAllElementsLoaded) {
    const { addEvent } = Highcharts;
    let loadElements = [];
    return el => {
        loadElements.push(el);
        addEvent(el, 'load', () => {
            loadElements = loadElements.filter(x => x !== el);
            if (loadElements !== undefined && loadElements.length === 0) {
                onAllElementsLoaded();
            }
        });
    };
}(onLoad));
```

KacperMadej: @jon-a-nygaard 
There's a problem with that in Safari. If you are about to use `onload` event then you need to set it before `src` is set for the image. Other browsers don't care about the order (as long as the image is not yet loaded before the `onload` is set).

This is resolved in core Highcharts. A dummy image with the same source is create that has the correct order (event [first](https://github.com/highcharts/highcharts/blob/6310ae47e67906f02a7ae5a29d51fadc53d20b73/ts/parts/SvgRenderer.ts#L5322), and src [next](https://github.com/highcharts/highcharts/blob/6310ae47e67906f02a7ae5a29d51fadc53d20b73/ts/parts/SvgRenderer.ts#L5363)), so it works well when it comes to detection of loaded image size and this is the main point of tests here. I could write a similar or even the same logic, but since we have it in code already it seemed like a faster solution to just use it.

What you are suggesting is doable (it's already done in [here](https://github.com/highcharts/highcharts/blob/6310ae47e67906f02a7ae5a29d51fadc53d20b73/ts/parts/SvgRenderer.ts#L5320-L5362)), but won't work directly on adding a symbol as the relevant code is inaccessible without a chart (as it fires `chart.load` event), so I don't see any other option then to either:
* write a wrapper (bad idea in a test as far as it's avoidable),
* or use a different functions (bad idea as the test will no longer serve the purpose of symbol testing),
* or fake a chart in `Highcharts.charts` as well as `chartIndex` on the used Renderer (this is the least ugly solution I can come up with).

tl;dr: I don't like using it either, but alternatives are even worse, so maybe there's another way?

jon-a-nygaard: Hmm, I see.

I believe an `onload` event for the symbol would be ideal here, but as far as I can see it is not available at the moment.

Would it make sense to add an `onload` property to the `options` parameter in `SVGREnderer.symbol`, so that we could add an event listener for the symbol, or is it better to use the existing solution for this test? @KacperMadej @TorsteinHonsi

KacperMadej: IMHO adding a functionality to the core just to run some tests is not optimal, but I could work with any decision @TorsteinHonsi @jon-a-nygaard

TorsteinHonsi: Agreed, we can keep it as it is for now to avoid tempering with the core.

Issue #9036, opened 2018-09-20
Sankey minimum node height enhancement request


calsupik:
#### Expected behaviour
Add an option to set a minimun node height (similar to nodeLength) to prevent nodes from becoming too small.

#### Actual behaviour
Node height is determined by the larger of either the links to or the links from the node.

#### Live demo with steps to reproduce
In my below demo, the node 'No Next Step' has a very small weight, and no paths after it, therefore the node is extremely small (I understand it is by design). It would be beneficial in my specific use case to have a minimum height for that node so that the label fits fully inside of the node.

https://jsfiddle.net/tvwr8fd3/193/

#### Product version
Highcharts 6.1.0

#### Affected browser(s)
N/A

calsupik:
Support added in PR #9038.

mkpasala:
http://jsfiddle.net/Lmn6gjby/24/ I had a similar issue here. When the values are 0 then the labels are overlapped one on another.

baahuabhi:
hi , i too am facing same issue with google react charts for sankey. can anyone suggest better library?

PR #11454, opened 2019-07-22
enhancement/7318-minLinkHeight-sankey

### LINKED ISSUE #7318: Sankey / minPointLength not working
If I use links with big differences in weight, it happens, that the lower ones are shown with a linkHeight lower than 1px and so they become invisible. Unfortunately the parameter **minPointLength** isn't working, respectively it is not implemented.

As a workaround I've added following code to the source:

`linkHeight = (point.weight != null && linkHeight < 1) ? 1 : linkHeight;	`

With this hack, all valid links will be shown with a minimal linkHeight of 1px.

It would be nice, if I could use the parameter **minPointLength** for that instead of the hack!

sebastianbochan:
Fixed #7318, added new minLinkWidth option to sankey series.


pawelfus:
There's also another request for `minPointLength` which would set.. min node height (width in inverted), see #9036.

I guess one should be `minPointLength` and the other `minLinkLength`/`minNodeLength`. Since `series.data` creates links, then maybe `minPointLength` should define height/width for links and `minNodeLength` for nodes?

oysteinmoseng:
Ah, I see. Your suggestion makes sense to me. There is a `nodeWidth` option, so maybe `minNodeWidth` for nodes and reuse `minPointLength` for links?

That being said, `minPointLength` is probably not a very intuitive name for the user (the length of the link to me would be the length between the nodes - not the width/height), so it might be better in the long run to actually add `minLinkWidth` and remove `minPointLength` from the API. Thoughts?

sebastianbochan:
Make sense for me, will change that today.

sebastianbochan:
@oysteinmoseng I replaced parameter with the suggested name. 

Let me know if you have any additional questions.

oysteinmoseng:
Mention @TorsteinHonsi in case he has additional comments since we are adding API option (`minLinkWidth`)

PR #11456, opened 2019-07-22
bugfix/4024-stacking-xaxis

### LINKED ISSUE #4024: column stacking applied on seperate axes
Columns appear to be stacked even if they are on different xAxes: http://jsfiddle.net/0d5mysLy/

pawelfus:
Fixed #4024, stacking columns on different xAxis translated columns as in stacking on the same xAxis.
___
Note: We can not use `series.options.xAxis` option for generating key, because two axes with the same position (e.g. duplicated xAxis on top) should apply stacking naturally.

Additionally replaced hardcoded stack keys names in tests with dynamic values, generated in core.

### DIFF
--- a/ts/parts/Stacking.ts
+++ b/ts/parts/Stacking.ts
@@ -499,11 +499,15 @@ Chart.prototype.getStacks = function (this: Highcharts.Chart): void {
     });
 
     chart.series.forEach(function (series: Highcharts.Series): void {
+        var xAxis = series.xAxis;
+
         if (series.options.stacking &&
             (series.visible === true ||
             (chart.options.chart as any).ignoreHiddenSeries === false)
         ) {
-            series.stackKey = series.type + pick(series.options.stack, '');
+            series.stackKey = series.type +
+                pick(series.options.stack, '') + ',' +
+                xAxis.options.left + ',' + xAxis.options.width;
         }
     });
 };
--- a/ts/parts/Stacking.ts
+++ b/ts/parts/Stacking.ts
@@ -489,7 +489,8 @@ H.StackItem.prototype = {
  * @return {void}
  */
 Chart.prototype.getStacks = function (this: Highcharts.Chart): void {
-    var chart = this;
+    var chart = this,
+        inverted = chart.inverted;
 
     // reset stacks for each yAxis
     chart.yAxis.forEach(function (axis: Highcharts.Axis): void {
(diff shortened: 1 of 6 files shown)


ts/parts/Stacking.ts
```diff
@@ -499,11 +499,15 @@ Chart.prototype.getStacks = function (this: Highcharts.Chart): void {
...
+        var xAxis = series.xAxis;
+
         if (series.options.stacking &&
             (series.visible === true ||
             (chart.options.chart as any).ignoreHiddenSeries === false)
         ) {
-            series.stackKey = series.type + pick(series.options.stack, '');
+            series.stackKey = series.type +
+                pick(series.options.stack, '') + ',' +
+                xAxis.options.left + ',' + xAxis.options.width;
```
jon-a-nygaard: In theory `xAxis.options.left` and `xAxis.options.width` could be equal for multiple axes, is there a unique identifier for the axis that could be used instead?

pawelfus: That's correct - and that's why I've picked these options. See my comment above:
> Note: We can not use series.options.xAxis option for generating key, because two axes with the same position (e.g. duplicated xAxis on top) should apply stacking naturally.

For example: https://jsfiddle.net/BlackLabel/qeny2h9b/1/ - left/width are the same, stacking should be applied. I simply don't want to introduce any regression here, so decided to use left+width instead of `series.xAxis`.

jon-a-nygaard: Aha, I understand now.

>  two axes with the same position (e.g. duplicated xAxis on top) should apply stacking naturally.

Is this documented behavior, and a common use case?
Reason why I am asking is because I find this to be a bit curious behavior, and would have expected that this should rather be supported by using `xAxis.linkedTo`: https://jsfiddle.net/jon_a_nygaard/gnehrow6/3/

pawelfus: It works unless you use categorized axes: https://jsfiddle.net/BlackLabel/9pLv5rtq/3/ - of course, adding `xAxis: 1` does work, but them `linkedTo` is unnecessary: https://jsfiddle.net/BlackLabel/9pLv5rtq/5/

> Is this documented behavior,

I don't recall. 
> and a common use case?

I don't know :) I simply don't want to break backward compatibility. We can change this if you want and see if users will come back with regressions

jon-a-nygaard: > We can change this if you want and see if users will come back with regressions

I will prefer to have this part changed. I will add Torstein in a comment so that he can have a look and make final decision when he is back from vacation.

pawelfus: Just realized that solution should also support inverted charts. Added in the commit below.

jon-a-nygaard: @pawelfus Is the `top`, `left`, `height` and `width` if you add an unique identifier as previously discussed?

pawelfus: No, no, it's the same - just previous solution (left+width) didn't work for inverted charts, where top+height should be used to generate a key. It works the same way, so series on two different xAxes, but the same position, will stack.

Issue #11316, opened 2019-07-04
Adding Trendline to official Highcharts indicators package


Izothep:
Adding Trendline to official Highcharts indicators group. 

Currently Trendline is described in official HighStock documentation:
https://www.highcharts.com/docs/advanced-chart-features/custom-technical-indicators

PR #11463, opened 2019-07-23
enhancement/11316-trendline-indicator

sebastianbochan:
Enhancement #11316, trendline indicator.

### DIFF
--- a/js/indicators/trendline.src.js
+++ b/js/indicators/trendline.src.js
@@ -39,13 +39,16 @@ seriesType(
      * @optionparent plotOptions.trendline
      */
     {
+        /**
+         * @excluding period
+         */
         params: {
             /**
              * The point index which indicator calculations will base. For
              * example using OHLC data, index=2 means the indicator will be
              * calculated using Low values.
              *
-             * By default index value used to be set to 3.
+             * @default 3
              */
             index: 3
         }
--- a/js/indicators/trendline.src.js
+++ b/js/indicators/trendline.src.js
@@ -1,0 +1,126 @@
+/* *
+ *
+ *  License: www.highcharts.com/license
+ *
+ * */
+
+'use strict';
+
+import H from '../parts/Globals.js';
+
+import U from '../parts/Utilities.js';
+var isArray = U.isArray;
+
+var seriesType = H.seriesType;
+
+/**
+ * The Trend line series type.
+ *
+ * @private
+ * @class
+ * @name Highcharts.seriesTypes.trendline
+ *
+ * @augments Highcharts.Series
+ */
+seriesType(
+    'trendline',
+    'sma',
+    /**
+     * Trendline (linear regression) fits a straight line to the selected data
+     * using a method called the Sum Of Least Squares. This series requires the
+     * `linkedTo` option to be set.
+     *
+     * @sample stock/indicators/trendline
+     *         Exponential moving average indicator
+     *
+     * @extends      plotOptions.sma
+     * @since        7.1.3
+     * @product      highstock
+     * @optionparent plotOptions.trendline
+     */
+    {
+        params: {
+            /**
+             * The point index which indicator calculations will base. For
+             * example using OHLC data, index=2 means the indicator will be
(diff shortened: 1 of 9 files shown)


js/indicators/trendline.src.js
```diff
@@ -0,0 +1,128 @@
...
+     * @sample stock/indicators/trendline
+     *         Exponential moving average indicator
+     *
+     * @extends      plotOptions.sma
+     * @since        7.1.3
+     * @product      highstock
+     * @optionparent plotOptions.trendline
+     */
+    {
+        name: 'TRENDLINE',
```
pawelfus: Instead of fixed `name`, use `proto.nameBase`. I think capitals are not needed too, simple `Trendline` should be fine.

js/indicators/trendline.src.js
```diff
@@ -0,0 +1,128 @@
...
+seriesType(
+    'trendline',
+    'sma',
+    /**
+     * Trendline (linear regression) fits a straight line to the selected data
+     * using a method called the Sum Of Least Squares. This series requires the
+     * `linkedTo` option to be set.
+     *
+     * @sample stock/indicators/trendline
+     *         Exponential moving average indicator
```
pawelfus: Wrong name

js/indicators/trendline.src.js
```diff
@@ -0,0 +1,128 @@
...
+             * The point index which indicator calculations will base. For
+             * example using OHLC data, index=2 means the indicator will be
+             * calculated using Low values.
+             *
+             * By default index value used to be set to 0. Since Highstock 7
+             * by default index is set to 3 which means that the trendline
+             * indicator will be calculated using Close values.
+             */
+            index: 3,
+            period: 0
```
pawelfus: Is period available in this indicator? Shouldn't be excluded?

js/indicators/trendline.src.js
```diff
@@ -0,0 +1,128 @@
...
+            /**
+             * The point index which indicator calculations will base. For
+             * example using OHLC data, index=2 means the indicator will be
+             * calculated using Low values.
+             *
+             * By default index value used to be set to 0. Since Highstock 7
+             * by default index is set to 3 which means that the trendline
+             * indicator will be calculated using Close values.
+             */
+            index: 3,
```
pawelfus: I think description is a bit vague - indicator will be available since v7.1.3, but description mentions older versions? I think second paragraph can be simply removed.

js/indicators/trendline.src.js
```diff
@@ -0,0 +1,128 @@
...
+                sumX = 0,
+                sumY = 0,
+                sumXY = 0,
+                sumX2 = 0,
+                xValLength = xVal.length,
+                index, alpha, beta, i, x, y;
+
+            // Switch index for OHLC / Candlestick / Arearange
+            if (isArray(yVal[0])) {
+                index = params.index ? params.index : 3;
```
pawelfus: Replace with `pick(params.index, 3)` - otherwise `index=0` won't work. Anyway, do we need this? `params.index` should be always defined, right?

js/indicators/trendline.src.js
```diff
@@ -0,0 +1,128 @@
...
+        }
+    }
+);
+
+/**
+ * A `TrendLine` series. If the [type](#series.trendline.type) option is not
+ * specified, it is inherited from [chart.type](#chart.type).
+ *
+ * @extends   series,plotOptions.trendline
+ * @since     6.0.0
```
pawelfus: Wrong version

sebastianbochan: Its already fixed in last commit.

js/indicators/trendline.src.js
sebastianbochan: Makes sense for me.

TorsteinHonsi:
@sebastianbochan Can you give a more descriptive description for the changelog?

sebastianbochan:
@TorsteinHonsi I updated the description, let me know if now all is clear.

PR #11465, opened 2019-07-23
ts/11438-quad-node-clean

### LINKED ISSUE #11438: QuadTree handles two nodes with same box not type safe
#### Expected behaviour
```
var newNode = new QuadTreeNode({});

newNode.body = point;
newNode.isInternal = false;

this.nodes.push(newNode);
```
on line https://github.com/highcharts/highcharts/blob/29640f7f501afcdb9ef7342d3cb9f38cbe6f70e7/ts/modules/networkgraph/QuadTree.ts#L173

#### Actual behaviour

https://github.com/highcharts/highcharts/blob/29640f7f501afcdb9ef7342d3cb9f38cbe6f70e7/ts/modules/networkgraph/QuadTree.ts#L173

#### Internal background

This is kind of a safeguard when two nodes fall exactly into the same position/box.

In this case, we will never be able to place these two nodes in two separate leaves with separate boxes.

Then I simply break the rule of quadtree - and that branch has more leaves (5 instead of 4)

pawelfus:
Fixed #11438 - wrong types in QuadTree nodes. Additionally cleaned up the QuadNode - there was post dev code that was not necessary anymore.

Just for future reference, it helps debugging BH-approx: https://jsfiddle.net/BlackLabel/m3tr104z/

### DIFF
--- a/ts/modules/networkgraph/QuadTree.ts
+++ b/ts/modules/networkgraph/QuadTree.ts
@@ -25,14 +25,7 @@ declare global {
             public maxDepth: number;
             public root: QuadTreeNode;
             public calculateMassAndCenter(): void;
-            public clear(chart?: Chart): void;
             public insertNodes(nodes: Array<NetworkgraphPoint>): void;
-            public render(chart?: Chart, clear?: boolean): void;
-            public renderBox(
-                qtNode: QuadTreeNode,
-                chart: Chart,
-                clear?: boolean
-            ): void;
             public visitNodeRecursive(
                 node: (QuadTreeNode|null),
                 beforeCallback: (Function|null),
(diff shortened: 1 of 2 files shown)


ts/modules/networkgraph/QuadTree.ts
```diff
@@ -151,6 +139,8 @@ H.extend(
             point: Highcharts.NetworkgraphPoint,
             depth: number
         ): void {
+            var newQuadTreeNode: Highcharts.QuadTreeNode;
```
jon-a-nygaard: FYI: `const` and `let` can be used in our TypeScript files and is transpiled nicely to `var` in our JS files. Block scoping is also handled nicely, which is relevant for this example where `newQuadTreeNode` is used only within the `else` block.

pawelfus: Thanks for the info!

PR #11466, opened 2019-07-23
attach accessibility feature callbacks via local binding

### LINKED ISSUE #11457: Clicking Exporting Context Button opens wrong menu
#### Expected behaviour

When I have multiple charts on a single web page, and I click the exporting module's context button, the context menu for that chart should become visible.

#### Actual behaviour

When there are multiple charts on the page, clicking the context button reveals and shifts focus to the context menu for the previous chart.

#### Product version
 
7.1.2

#### Affected browser(s)

Firefox 67.0, Chromium

sent1nel:
fixes #11457


sent1nel:
This PR won't work against master and is version-specific to v7.1.2.

PR #11472, opened 2019-07-24
Exported isObject in parts/Utilities.js

jon-a-nygaard:
Title says it all.


jon-a-nygaard:
> Type looks good. I wish we could make it an intelligent type.

I had a look at making `isObject` a type guard, but I was not able to do it.

PR #11473, opened 2019-07-24
Export isDOMElement in parts/Utilities.js

jon-a-nygaard:
Title says it all.
**NB!:** Dependent upon #11472, which should be merged first.

### DIFF
--- a/ts/parts/Utilities.ts
+++ b/ts/parts/Utilities.ts
@@ -194,7 +194,7 @@ declare global {
         ): number;
         function isArray(obj: unknown): obj is Array<unknown>;
         function isClass(obj: any): boolean;
-        function isDOMElement(obj: any): boolean;
+        function isDOMElement(obj: unknown): obj is HTMLElement;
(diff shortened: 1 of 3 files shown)


ts/parts/Utilities.ts
```diff
@@ -1283,7 +1283,7 @@ function isObject(obj: any, strict?: boolean): boolean {
  * @return {boolean}
  *         True if the argument is a HTML Element.
  */
-H.isDOMElement = function (obj: any): boolean {
+function isDOMElement(obj: any): boolean {
```
bre1470: Did you tried `function isDOMElement(obj: unknown): obj is HTMLElement`?

jon-a-nygaard: No, will check that out 👍

jon-a-nygaard: @bre1470 Please see latest changes

bre1470: Looks great! Thanks. 👍

PR #11474, opened 2019-07-24
vacation2019/ts/networkgraph

### LINKED ISSUE #11434: Migrate `modules/networkgraph/networkgraph` to TypeScript

bre1470:
Fixed #11434 - Migrated NetworkgraphSeries to TypeScript.

### DIFF
--- a/ts/modules/networkgraph/networkgraph.src.ts
+++ b/ts/modules/networkgraph/networkgraph.src.ts
@@ -79,7 +79,6 @@ declare global {
             draggable?: boolean;
             link?: SVGAttributes;
             marker?: NetworkgraphPointMarkerOptionsObject;
-            mass?: NodesSeriesOptions['mass'];
             nodes?: Array<NetworkgraphPointOptions>;
             states?: NetworkgraphSeriesStatesOptions;
         }
--- a/ts/modules/networkgraph/networkgraph.src.ts
+++ b/ts/modules/networkgraph/networkgraph.src.ts
@@ -1,0 +1,1471 @@
+/* *
+ *
+ *  Networkgraph series
+ *
+ *  (c) 2010-2019 Paweł Fus
+ *
+ *  License: www.highcharts.com/license
+ *
+ *  !!!!!!! SOURCE GETS TRANSPILED BY TYPESCRIPT. EDIT TS FILE ONLY. !!!!!!!
+ *
+ * */
+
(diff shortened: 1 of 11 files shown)


ts/modules/networkgraph/networkgraph.src.ts
```diff
@@ -0,0 +1,1471 @@
...
+            inactive?: NetworkgraphPointStatesInactiveOptionsObject;
+        }
+        interface NetworkgraphSeriesOptions
+            extends LineSeriesOptions, NodesSeriesOptions
+        {
+            dataLabels?: NetworkgraphDataLabelsOptionsObject;
+            draggable?: boolean;
+            link?: SVGAttributes;
+            marker?: NetworkgraphPointMarkerOptionsObject;
+            mass?: NodesSeriesOptions['mass'];
```
pawelfus: There shouldn't be `series.mass` option. Only `series.nodes[index].mass`

ts/modules/networkgraph/networkgraph.src.ts
```diff
@@ -0,0 +1,1471 @@
...
+        {
+            public chart: NetworkgraphChart;
+            public createNode: NodesMixin['createNode'];
+            public data: Array<NetworkgraphPoint>;
+            public destroy: NodesMixin['destroy'];
+            public directTouch: boolean;
+            public drawTracker: TrackerMixin['drawTrackerPoint'];
+            public forces: Array<string>;
+            public hasDraggableNodes: boolean;
+            public isCartesian: boolean;
```
pawelfus: Isn't this defined in `Series` class? The same for `noSharedTooltip`, `pointArrayMap` and `requireSorting`?

bre1470: Yes, but it gets here overwritten. So if you change the type in the parent, it will bring up a conflict, so that you have to update the overwrite here as well.

pawelfus: Nice, thanks!

ts/modules/networkgraph/networkgraph.src.ts
bre1470: I removed the mass property. Please review.

pawelfus: Thanks 👍

PR #11490, opened 2019-07-25
Export erase in parts/Utilities.js

jon-a-nygaard:
Title says it all.

### DIFF
--- a/ts/parts/Utilities.ts
+++ b/ts/parts/Utilities.ts
@@ -1337,7 +1337,7 @@ function isNumber(n: unknown): n is number {
  *
  * @return {void}
  */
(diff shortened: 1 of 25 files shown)


ts/parts/Utilities.ts
```diff
@@ -1337,7 +1337,7 @@ function isNumber(n: unknown): n is number {
  *
  * @return {void}
  */
-H.erase = function (arr: Array<any>, item: any): void {
+function erase(arr: Array<any>, item: any): void {
```
bre1470: Maybe it is better to replace the `any` types with `unknown`, which should be similar for TypeScript.

jon-a-nygaard: I agree, please see latest changes.

PR #11508, opened 2019-07-30
Add missing TS changes in JS

jon-a-nygaard:
Added the missing changes in JS files compiled from TS.

### DIFF
--- a/js/modules/networkgraph/networkgraph.src.js
+++ b/js/modules/networkgraph/networkgraph.src.js
@@ -1060,7 +1060,7 @@ seriesType('networkgraph', 'line',
  * Individual data label for each node. The options are the same as
  * the ones for [series.networkgraph.dataLabels](#series.networkgraph.dataLabels).
  *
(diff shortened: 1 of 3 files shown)


js/modules/networkgraph/networkgraph.src.js
```diff
@@ -1056,12 +1056,4 @@ seriesType('networkgraph', 'line',
  * @product   highcharts
  * @apioption series.networkgraph.nodes.mass
  */
-/**
```
bre1470: This is result of an older merge conflict and has to be part of the TS file. Additionally the type naming is not correct instead of `...Network...` it hast to be `...Networkgraph...`. Can you fix it, please? Then I do not have to mix it with my migration PR.

jon-a-nygaard: Thanks @bre1470 👍 Please see latest changes for fix.

bre1470: 👍 Looks good. Thank you!

PR #11520, opened 2019-07-31
ts/fix-polyfill-type

bre1470:
Improved type for `Array.some` in polyfill.


bre1470:
Yes. Please review once more.

Issue #7711, opened 2018-01-23
yAxis display out of pane after setExtremes


zhy1stgg:
#### Live demo with steps to reproduce
http://jsfiddle.net/b62tdy3w/

click the button (set yAxis[0]'s extremes)

![snipaste_2018-01-23_17-38-56](https://user-images.githubusercontent.com/6829290/35268512-48cd3002-0064-11e8-8b93-bc0343164bab.png)

#### Product version
Highstock 6.0 +

sebastianbochan:
Hi @zhy1stgg 
Thank you for the reporting.

_Internal note_
It looks like an issue with candlestick and OHLC, cause line /area work fine.

zhy1stgg:
Hi @sebastianbochan line /area also have this issue, 
![snipaste_2018-01-24_15-32-58](https://user-images.githubusercontent.com/6829290/35319677-db2b3b62-011b-11e8-8d70-a9c923f97830.png)

KacperMadej:
Minimal demo for line: http://jsfiddle.net/BlackLabel/b62tdy3w/4/
and for column (wrong initial animation): http://jsfiddle.net/BlackLabel/b62tdy3w/5/

sebastianbochan:
@zhy1stgg problem with halo is related with known bug  https://github.com/highcharts/highcharts/issues/7650

pawelfus:
Related to #7711

jon-a-nygaard:
Could you @KacperMadej and @sebastianbochan confirm if this issue is fixed by #11531?

KacperMadej:
@jon-a-nygaard 
The good news is that the issue is resolved.
A bad news is below:

Assuming that current master branch code behavior is caused by the fix for the #11531 the column issue seems to be worse - no tooltip at all for the visible points. Demo: http://jsfiddle.net/BlackLabel/ke7nphqu/

For the line chart it also doesn't look perfect - demo: http://jsfiddle.net/BlackLabel/4auLorb7/
When you hover over a plot series are getting inactive state after hovering a pane with visible tooltip, single tooltip is frozen, crosshair is out of sync with tooltip.

jon-a-nygaard:
> no tooltip at all for the visible points

> For the line chart it also doesn't look perfect

Hmm, this is strange. Will look into what is causing these regressions.

PR #11531, opened 2019-08-01
Bugfix/11493 split tooltip

### LINKED ISSUE #7650: Hovered points outside plotting area
#### Expected behaviour
Points which are outside the plotting area should not be visible/hovered.

#### Actual behaviour
See the attached screenshot:
image

http://jsfiddle.net/efujfuat/1/
1. Move the mouse cursor to point {Series 1, x: 2, y: 3} 
-> point {Series 2, x: 2, y: 1} will be visible outside the plotting area.

Removing the chart setting `{ tooltip: {shared: true} }` fixes the problem. But I need this shared tooltip.

Similar issue: #3916

#### Product version
Highcharts: latest version

#### Affected browser(s)
Chrome v63

### LINKED ISSUE #11493: Y Axis scaled using set Extremes Is inappropriate
#### Expected behaviour

#### Actual behaviour

5. Hover over the Chart

Observation

AfterScaling

#### Product version

Highstock - 7.1.2

#### Affected browser(s)

jon-a-nygaard:
Fixed #7650 and #11493, points outside plot area were included as hover points, and were displayed in shared and split tooltip.

---
# Description
- Filter `hoverPoints` by `point.isInside` in `getHoverData` 

# Example(s)
- [Demo of issue #7650](https://jsfiddle.net/jon_a_nygaard/z0o94uwm/)
- [Demo of issue #7650 with fix applied](https://jsfiddle.net/jon_a_nygaard/nv6tq2kh/)
- [Demo of issue #11493](https://jsfiddle.net/jon_a_nygaard/nhudqsfa/)
- [Demo of issue #11493 with fix applied](https://jsfiddle.net/jon_a_nygaard/L68gt5pu/)

# Related issue(s)
- Closes #7650
- Closes #11493 
- Possibly related to #7711

### DIFF
--- a/js/parts/Pointer.js
+++ b/js/parts/Pointer.js
@@ -401,7 +401,9 @@ Highcharts.Pointer.prototype = {
                 // Get all points with the same x value as the hoverPoint
                 searchSeries.forEach(function (s) {
                     var point = find(s.points, function (p) {
-                        return p.x === hoverPoint.x && !p.isNull;
+                        return (p.x === hoverPoint.x &&
+                            !p.isNull &&
+                            p.isInside !== false);
                     });
                     if (isObject(point)) {
                         /*
--- a/samples/unit-tests/pointer/members/demo.js
+++ b/samples/unit-tests/pointer/members/demo.js
@@ -864,3 +864,42 @@ QUnit.test('Pointer.getHoverData', function (assert) {
         'Combination chart: Only one point hovered when hovered series has noSharedTooltip'
     );
 });
+
+QUnit.test('Hover points', assert => {
+    const chart = Highcharts.chart('container', {
+        yAxis: {
+            min: 2,
+            max: 5
+        },
+        series: [{
+            data: [1, 2, 3, 4]
+        }, {
+            data: [3, 2, 1, 4]
+        }],
+        tooltip: {
+            shared: true
+        }
+    });
+    const { series: [series1] } = chart;
+    const getSeriesNameAndPointIndex = point => `${point.series.name}.${point.index}`;
+
+    assert.strictEqual(
+        chart.hoverPoints,
+        undefined,
+        'Should have no hover points before any interaction'
+    );
+
+    series1.points[3].onMouseOver();
+    assert.deepEqual(
+        chart.hoverPoints.map(getSeriesNameAndPointIndex),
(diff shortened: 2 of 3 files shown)


pawelfus:
According to #11493 - tooltip should not display values that are outside the plotting area - shouldn't we filter them?

Maybe not that important, but solution doesn't work for xAxis: https://jsfiddle.net/BlackLabel/y8p47k2d/

TorsteinHonsi:
> shouldn't we filter them

Agreed, I have the feeling it would be a much simpler and cleaner solution just to filter the `hoverPoints` by which points are inside the plot area. And it would not differentiate between `split` and `shared`.

This seems to fix both cases, with the consequence that points outside the plot area are not rendered in the shared tooltip, which I think is correct:

TorsteinHonsi:
> shouldn't we filter them

Agreed, I have the feeling it would be a much simpler and cleaner solution just to filter the `hoverPoints` by which points are inside the plot area. And it would not differentiate between `split` and `shared`.

Checking for `p.isInside` seems to fix both cases, with the consequence that points outside the plot area are not rendered in the shared tooltip, which I think is correct:

```js

**Comment by TorsteinHonsi** [comment]
> shouldn't we filter them

Agreed, I have the feeling it would be a much simpler and cleaner solution just to filter the `hoverPoints` by which points are inside the plot area. And it would not differentiate between `split` and `shared`. This is also in accordance with how it behaved in v5.0.12, prior to the regression.

Checking for `p.isInside` seems to fix both cases:

```ts
                    var point = find(s.points, function (
                        p: Highcharts.Point
                    ): boolean {
                        return p.isInside === true && p.x === hoverPoint.x && !p.isNull;
                    });
```

jon-a-nygaard:
> shouldn't we filter them

I have a feeling that the behavior was like that once, but it was complained about in an issue that points was missing from the shared tooltip. I will see if I can find this issue.

jon-a-nygaard:
I am unable to find any issue about what was mentioned in my previous comment, so I think we then all agree then that only points within the plot area should be regarded as hover points, and included in the shared tooltip.

I will update the branch accordingly.

jon-a-nygaard:
@pawelfus and @TorsteinHonsi please see latest changes

pawelfus:
Looks good, thanks!

PR #11548, opened 2019-08-02
`inactive.except`: implemented PoC for heatmap.

pawelfus:
Added `series.state.inactive.except` option to the heatmap series.
___
### Demos:
- ~[Demo](https://highcharts.github.io/highcharts-utils/samples/#gh/4b3df845ee/sample/highcharts/series-heatmap/inactive-except) for API~ <- broken, [another demo](https://jsfiddle.net/BlackLabel/amx6sdtj/2/)

### API proposal:
I don't see a perfect name for this option. I suggest to use `inactive.except: string`:

> When hovering over a cell, decide which **other** cells
> should not apply an inactive state.
> For example `'column-row'` will apply inactive state to all
> other cells except cells within the same row or column.
>
> Possible options:
> - `'all'`- all cells (aka disabled)
> - `'column'` - all cells within the same row
> - `'row'` - all cells within the same column
> - `'column-row'` - all cells within the same row or column
> - `'self'` - only current cell
> Point padding for a single point.

```
/**
 * @sample         highcharts/series-heatmap/inactive-except/
 *                 Compare different `except` options
 *
 * @validvalue     ["all", "column", "column-row", "row",  "self"]
 * @since          7.1.4
 * @product        highcharts highmaps
 */
```

Why not simply `filter`? While logic in heatmap looks exactly like filters, but for node-based series (networkgraph, sankey, dependency wheel, organization) it's not: in tree-structure, we don't loop over all points (performance), but only from the currently hovered point we change state for it's `from`/`to`/`siblings`/`parents`/`children`. So in this case, options for other series could be:

1. Dependencywheel/Sankey/Networkgraph/:
`@validvalue     ["all", "from", "from-to", "to", "self"]` - currently "from-to" (highlights all incoming/outgoing links with nodes) - we are good here, no need to implement "from" and "to" now. We could consider implementing simple "all"/"self" options, just to improve performance (instead of setting `inactive.opacity = 1` which still loops over all points, `all` could be simply be a switch on/off). Unless we would prefer "all" to be replaced with `inactive.enabled: boolean`.

2. Organization chart:
`@validvalue     ["all", "ancestors", "descendants-ancestors",  "descendants",  "self"]` - quite similar to the above. PoC:
- ["ancestors"](https://jsfiddle.net/BlackLabel/trb8y6fm/)
- ["descendants-ancestors"](https://jsfiddle.net/BlackLabel/rs41uk2a/)
- ["descendants"](https://jsfiddle.net/BlackLabel/k9nz0L8w/)

### Discussion points/issues:
- disabled in boosted mode - it seems that opacity is not supported and iterating over all cells takes too much time (see changes in `boost-overrides.js` file). I have a dilemma here: add a  wrapper (see PR) or add a condition in heatmap.js (`this.series.isSeriesBoosting`) to prevent inactive state.
- disabled animation - when animation is enabled ([demo](https://jsfiddle.net/BlackLabel/amx6sdtj/)) - the experience is.. no the best, feels like performance is poor

### To do:
- unit-tests

### DIFF
--- a/js/modules/boost/boost-overrides.js
+++ b/js/modules/boost/boost-overrides.js
@@ -331,6 +331,21 @@ wrap(Series.prototype, 'processData', function (proceed) {
     }
 });
 
+
+// In heatmap, inactive state tries to modify opacity for other points.
+// Disable this for canvas (where opacity is disabled) to improve performance.
+if (seriesTypes.heatmap) {
+    wrap(
+        seriesTypes.heatmap.prototype.pointClass.prototype,
+        'setState',
+        function (proceed) {
+            if (!this.series.isSeriesBoosting) {
+                proceed.apply(this, Array.prototype.slice.call(arguments, 1));
+            }
+        }
+    );
+}
+
 addEvent(Series, 'hide', function () {
     if (this.canvas && this.renderTarget) {
         if (this.ogl) {
--- a/js/parts-map/HeatmapSeries.js
+++ b/js/parts-map/HeatmapSeries.js
@@ -155,6 +155,30 @@ seriesType('heatmap', 'scatter',
              * rule.
              */
             brightness: 0.2
+        },
+        inactive: {
+            /**
+             * When hovering over a cell, decide which **other** cells
+             * should not apply an inactive state.
+             * For example `'column-row'` will apply inactive state to all
+             * other cells except cells within the same row or column.
+             *
+             * Possible options:
+             * - `'all'`- all cells (aka disabled)
+             * - `'column'` - all cells within the same row
+             * - `'row'` - all cells within the same column
+             * - `'column-row'` - all cells within the same row or column
+             * - `'self'` - only current cell
+             * Point padding for a single point.
+             *
+             * @sample         highcharts/series-heatmap/inactive-except/
+             *                 Compare different `except` options
+             *
+             * @validvalue     ["all", column", "row", "column-row", "self"]
+             * @since          7.1.4
+             * @product        highcharts highmaps
+             */
+            except: 'column-row'
         }
     }
 }, merge(colorSeriesMixin, {
@@ -264,9 +288,31 @@ seriesType('heatmap', 'scatter',
         this.valueMax = this.dataMax;
         // Get the extremes from the y data
         Series.prototype.getExtremes.call(this);
+    },
+    /**
+     * Mapping object between `inactive.except` and built-in methods
+     */
+    inactiveFilters: {
+        all: function () {
+            return true;
+        },
+        column: function (activePoint, otherPoint) {
+            return activePoint.x === otherPoint.x;
+        },
+        'column-row': function (activePoint, otherPoint) {
+            return activePoint.x === otherPoint.x ||
+                activePoint.y === otherPoint.y;
+        },
+        row: function (activePoint, otherPoint) {
+            return activePoint.y === otherPoint.y;
+        },
+        self: function (activePoint, otherPoint) {
+            return activePoint.x === otherPoint.x &&
+                activePoint.y === otherPoint.y;
+        }
     }
     /* eslint-enable valid-jsdoc */
-}), H.extend({
+}), merge({
     /**
      * Heatmap series only. Padding between the points in the heatmap.
      * @name Highcharts.Point#pointPadding
@@ -305,7 +351,27 @@ seriesType('heatmap', 'scatter',
         ];
     }
     /* eslint-enable valid-jsdoc */
-}, colorPointMixin));
+}, colorPointMixin, {
+    setState: function (state) {
+        var point = this, series = this.series, points = series.points, inactiveException = series.options.states
+            .inactive.except, filterFunction = series.inactiveFilters[inactiveException], i = 0;
+        if (filterFunction) {
+            series.options.inactiveOtherPoints = true;
+            for (; i < points.length; i++) {
+                // Check if points are not being destroyed
+                if (points[i] && points[i].series) {
+                    colorPointMixin.setState.call(points[i], state !== 'hover' ||
+                        (state === 'hover' &&
+                            filterFunction(point, points[i])) ?
+                        '' : 'inactive');
+                }
+            }
+            series.options.inactiveOtherPoints = true;
+        }
+        // Apply default state for current point:
+        colorPointMixin.setState.apply(this, arguments);
+    }
+}));
 /**
  * A `heatmap` series. If the [type](#series.heatmap.type) option is
  * not specified, it is inherited from [chart.type](#chart.type).
--- a/js/parts-map/HeatmapSeries.js
+++ b/js/parts-map/HeatmapSeries.js
@@ -307,8 +307,7 @@ seriesType('heatmap', 'scatter',
             return activePoint.y === otherPoint.y;
         },
         self: function (activePoint, otherPoint) {
-            return activePoint.x === otherPoint.x &&
-                activePoint.y === otherPoint.y;
+            return false;
         }
     }
     /* eslint-enable valid-jsdoc */
--- a/js/parts-map/HeatmapSeries.js
+++ b/js/parts-map/HeatmapSeries.js
@@ -136,6 +136,8 @@ seriesType('heatmap', 'scatter',
         padding: 0 // #3837
     },
     /** @ignore-option */
+    inactiveOtherPoints: true,
+    /** @ignore-option */
     marker: null,
     /** @ignore-option */
     pointRange: null,
@@ -311,7 +313,7 @@ seriesType('heatmap', 'scatter',
         }
     }
     /* eslint-enable valid-jsdoc */
-}), merge({
+}), H.extend({
     /**
      * Heatmap series only. Padding between the points in the heatmap.
      * @name Highcharts.Point#pointPadding
@@ -350,29 +352,7 @@ seriesType('heatmap', 'scatter',
         ];
     }
     /* eslint-enable valid-jsdoc */
-}, colorPointMixin, {
-    setState: function (state) {
-        var point = this, series = this.series, points = series.points, inactiveException = series.options.states &&
-            series.options.states.inactive &&
-            series.options.states.inactive.except, filterFunction, i = 0;
-        if (inactiveException) {
-            filterFunction = series.inactiveFilters[inactiveException];
-            series.options.inactiveOtherPoints = true;
-            for (; i < points.length; i++) {
-                // Check if points are not being destroyed
-                if (points[i] && points[i].series) {
-                    colorPointMixin.setState.call(points[i], state !== 'hover' ||
-                        (state === 'hover' &&
-                            filterFunction(point, points[i])) ?
-                        '' : 'inactive');
-                }
-            }
-            series.options.inactiveOtherPoints = true;
-        }
-        // Apply default state for current point:
-        colorPointMixin.setState.apply(this, arguments);
-    }
-}));
+}, colorPointMixin));
 /**
  * A `heatmap` series. If the [type](#series.heatmap.type) option is
  * not specified, it is inherited from [chart.type](#chart.type).
--- a/js/parts-map/HeatmapSeries.js
+++ b/js/parts-map/HeatmapSeries.js
@@ -174,7 +174,7 @@ seriesType('heatmap', 'scatter',
              * @sample         highcharts/series-heatmap/inactive-except/
              *                 Compare different `except` options
              *
-             * @validvalue     ["all", column", "row", "column-row", "self"]
+             * @validvalue     ["all", "column", "row", "column-row", "self"]
              * @since          7.1.4
              * @product        highcharts highmaps
              */
@@ -353,9 +353,11 @@ seriesType('heatmap', 'scatter',
     /* eslint-enable valid-jsdoc */
 }, colorPointMixin, {
     setState: function (state) {
-        var point = this, series = this.series, points = series.points, inactiveException = series.options.states
-            .inactive.except, filterFunction = series.inactiveFilters[inactiveException], i = 0;
-        if (filterFunction) {
+        var point = this, series = this.series, points = series.points, inactiveException = series.options.states &&
+            series.options.states.inactive &&
+            series.options.states.inactive.except, filterFunction, i = 0;
+        if (inactiveException) {
+            filterFunction = series.inactiveFilters[inactiveException];
             series.options.inactiveOtherPoints = true;
             for (; i < points.length; i++) {
                 // Check if points are not being destroyed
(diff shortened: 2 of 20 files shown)


ts/parts-map/HeatmapSeries.ts
```diff
@@ -514,7 +606,47 @@ seriesType<Highcharts.HeatmapSeriesOptions>(
...
+    colorPointMixin,
+    {
+        setState: function (
+            this: Highcharts.HeatmapPoint,
+            state?: string
+        ): void {
+            var point = this,
+                series = this.series,
+                points = series.points,
+                inactiveException = (series.options.states as any)
```
bre1470: This `any` is avoidable by `series.options.states || {}`.

ts/parts-map/HeatmapSeries.ts
```diff
@@ -514,7 +606,47 @@ seriesType<Highcharts.HeatmapSeriesOptions>(
...
+                    .inactive.except as Highcharts
```
bre1470: This cast should not be necessary in the same way. Also try to write types on one line and if it is not possible, do
```
something{:| as} (
    The.Extremly.Long.Type.Name
)
```

ts/parts-map/HeatmapSeries.ts
```diff
@@ -514,7 +606,47 @@ seriesType<Highcharts.HeatmapSeriesOptions>(
...
+                                '' : 'inactive'
+                        );
+                    }
+                }
+
+                series.options.inactiveOtherPoints = true;
+            }
+
+            // Apply default state for current point:
+            colorPointMixin.setState.apply(this, arguments as any);
```
bre1470: Can be avoided with `[].from(arguments)`

ts/parts-map/HeatmapSeries.ts
bre1470: Actually this will not work, you have to make a deep availability check for the option to make sure it is set.

js/parts-map/HeatmapSeries.js
```diff
@@ -155,6 +155,30 @@ seriesType('heatmap', 'scatter',
...
+             * - `'column'` - all cells within the same row
+             * - `'row'` - all cells within the same column
+             * - `'column-row'` - all cells within the same row or column
+             * - `'self'` - only current cell
+             * Point padding for a single point.
+             *
+             * @sample         highcharts/series-heatmap/inactive-except/
+             *                 Compare different `except` options
+             *
+             * @validvalue     ["all", column", "row", "column-row", "self"]
```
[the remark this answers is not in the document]
pawelfus: Yes, I saw that after PR'ing. Thanks!

ts/parts-map/HeatmapSeries.ts
pawelfus: Ok, thanks!

ts/parts-map/HeatmapSeries.ts
pawelfus: I had some problems here, could you clarify this a little to me?
`series.options.states.inactive` object **is always set**, because it's defined above:
https://github.com/highcharts/highcharts/pull/11548/files#diff-f8b433ee2021a9296168918bf2cf2288R298

In API it's optional, but in `series.options` it's always present. I think there's no need for checking like this:

```js
inactiveException = series.options.states &&
    series.options.states.inactive &&
    series.options.states.inactive.except;
```

ts/parts-map/HeatmapSeries.ts
pawelfus: I don't see anywhere this usage. Is it a planned change in future for all `arguments as any` -> `[].from(arguments)` (reminder: no IE support). 


Anyway, changing this throws errors in `npx gulp script`:
> ts/parts-map/HeatmapSeries.ts(647,53): error TS2339: Property 'from' does not exist on type 'never[]'.

bre1470: Well the `any` type are turning off all TypeScript checks and therefor have to be removed in the future. How we will do this is not clear yet.

ts/parts-map/HeatmapSeries.ts
bre1470: For the options we will have to find something that suits our and user needs. For now everything is optional, because every option can be set or not. As I understand you, a user is not able to make this option undefined. Is this correct?

pawelfus: No, no, I mean user can set options as undefined, but for the most (all?) options in Series class we have (in short):

```js
series.options = merge(
  defaultOptions.plotOptions.type,
  defaultOptions.plotOptions.series,
  userOptions.plotOptions.type,
  userOptions.series[index]
);
```

This way, once something is set in default options for series (`H.seriesType('type', default.plotOptions.type, ...)`) it is available later, for example, in the case above we might have:

```js
merge(
  { },
  { states: { inactive: { except: 'column-row' } } },
  undefined,
  { color: 'red' } // from:  "series: [{ color: 'red' }]"
);
```

And still, the resulting `series.options` will have `states.inactive.except` available. Of course if a user sets for example: `states: function () {}` it will throw errors but this is a misconfigured chart (and users with TS shouldn't be able to build project).

bre1470: TypeScript forces us to handle misconfigurations. This can already happen internally when making use of prototype applies with fake instances. Please do the tripple check.

samples/highcharts/series-heatmap/inactive-except/demo.js
```diff
@@ -0,0 +1,102 @@
...
+                            inactive: {
+                                except: this.value
+                            }
+                        }
+                    });
+                }
+
+                document
+                    .querySelectorAll('input[name="inactive-except"]')
+                    .forEach(function (radio) {
```
KacperMadej: IE doesn't support `forEach` on a `NodeList` element ([source](https://developer.mozilla.org/en-US/docs/Web/API/NodeList/forEach)). Previously it was resolved like in this code: https://github.com/highcharts/highcharts/blob/974b67d60923373f6b5d4abb565e62dc811e1111/ts/modules/oldie.src.ts#L1948

ts/parts-map/HeatmapSeries.ts
```diff
@@ -461,11 +508,45 @@ seriesType<Highcharts.HeatmapSeriesOptions>(
...
+                activePoint: Highcharts.HeatmapPoint,
+                otherPoint: Highcharts.HeatmapPoint
+            ): boolean {
+                return activePoint.y === otherPoint.y;
+            },
+            self: function (
+                activePoint: Highcharts.HeatmapPoint,
+                otherPoint: Highcharts.HeatmapPoint
+            ): boolean {
+                return false;
```
bre1470: In this case you could remove the function parameters as they do not serve any purpose. Or is this a temporary thing?

ts/parts-map/HeatmapSeries.ts
```diff
@@ -18,6 +18,15 @@ import H from '../parts/Globals.js';
  */
 declare global {
     namespace Highcharts {
+        type HeatmapSeriesStatesInactiveExceptValue = (
+            'row'|'column'|'column-row'|'self'|'all'
+        );
+
+        type HeatmapInactiveFilterFunction = (
```
bre1470: Please make function types as interfaces, because type declarations are just placeholder which results in messy tooltips when it comes to functions. In your case:
```
interface HeatmapInactiveFilterFunction {
    (activePoint: HeatmapPoint, otherPoint: HeatmapPoint): boolean;
}
```

ts/parts-map/HeatmapSeries.ts
pawelfus: Good point! I'm not convinced if we even need this filter, added because it seems to be quite short one. Will update args.

ts/parts-map/HeatmapSeries.ts
pawelfus: Ok, will update, thanks!

js/modules/boost/boost-overrides.js
```diff
@@ -331,6 +331,21 @@ wrap(Series.prototype, 'processData', function (proceed) {
...
 
+
+// In heatmap, inactive state tries to modify opacity for other points.
+// Disable this for canvas (where opacity is disabled) to improve performance.
+if (seriesTypes.heatmap) {
+    wrap(
+        seriesTypes.heatmap.prototype.pointClass.prototype,
+        'setState',
+        function (proceed) {
+            if (!this.series.isSeriesBoosting) {
```
jon-a-nygaard: This seems like it will disable all states when boosting, not just `inactive`. Is that on purpose?

bre1470:
Looks good to me!

ts/parts-map/HeatmapSeries.ts
```diff
@@ -514,7 +591,49 @@ seriesType<Highcharts.HeatmapSeriesOptions>(
...
+            var point = this,
+                series = this.series,
+                points = series.points,
+                inactiveException = series.options.states &&
+                        series.options.states.inactive &&
+                        series.options.states.inactive.except,
+                filterFunction,
+                i = 0;
+
+            if (inactiveException) {
```
jon-a-nygaard: Would it not be better to keep this logic in [`applyInactiveState`](https://github.com/highcharts/highcharts/blob/3afea52337b352c0f4f848312f3425c2105bc38e/ts/parts/Pointer.ts#L873), and keep `setState` as before?

To me it seems like `Point.setState` will now, maybe unexpectedly, alter the state of many points, instead of just the one `setState` was called on.

ts/parts-map/HeatmapSeries.ts
```diff
@@ -261,6 +283,30 @@ seriesType<Highcharts.HeatmapSeriesOptions>(
...
+                 * Point padding for a single point.
+                 *
+                 * @sample         highcharts/series-heatmap/inactive-except/
+                 *                 Compare different `except` options
+                 *
+                 * @validvalue     ["all", "column", "row", "column-row", "self"]
+                 * @since          7.1.4
+                 * @product        highcharts highmaps
+                 */
+                except: 'column-row'
```
jon-a-nygaard: > Why not simply filter?
> in tree-structure, we don't loop over all points

Not completely convinced about the naming. Would it be wrong to call it `filter` even though it is not looping all points?

Also some alternative synonyms to `except` are `omit`, `exclude`, `skip`. Not sure if they are any better though.

pawelfus: Personally, I like `exclude` name. I'm also worried that users may expect `filter` to be a function. Like `array.filter(fun)`.

ts/parts-map/HeatmapSeries.ts
pawelfus: Good point. Maybe instead of `pointer.applyInactiveState` the [`series.setAllPointsToState`](https://github.com/highcharts/highcharts/blob/3afea52337b352c0f4f848312f3425c2105bc38e/ts/parts/Interaction.ts#L1546-L1555) would we a better place.

Logic behind using `Point.setState` was inherited from networkgraph, where one point's state propagates to all connected nodes.

js/modules/boost/boost-overrides.js
pawelfus: No, not really, can be added a condition for `inactive` state. 

Just to be on the same page, I realized that boosted heatmap does not have any states: https://www.highcharts.com/demo/heatmap-canvas (and simplified for select: https://jsfiddle.net/BlackLabel/Le6otns3/) so it's just a shortcut.

jon-a-nygaard: > I realized that boosted heatmap does not have any states

If this can be confirmed by @TorsteinHonsi or another member that it is correct behavior, then I am fine with the current solution.

ts/parts-map/HeatmapSeries.ts
jon-a-nygaard: > the series.setAllPointsToState would we a better place

Sounds good to me.

> Logic behind using Point.setState was inherited from networkgraph, where one point's state propagates to all connected nodes.

I see, I have the same worry for this.
Though if it is fine to alter other points in `setState`, then maybe the `Series.setState` could be extended to call an `applyStateToOtherPoints` or something, if it exists on the prototype?
Then all series could just implement this, instead of re-implementing `setState`. 🤔

pawelfus: In a separate PR I will try to refactor `networkgraph` (and other node-based series) to use `series.setAllPointsToState` instead of `setState`. So the solution will be consistent across series.

I just realized one thing: `setState()` is part of our API. Maybe we should ask ourselves  a question: should `point.setState()`, called by user, alter other point's state? I think this might be considered as a bug, so I would rather make `setAllPointsToState` available in API (if someone request that). What do you think?

jon-a-nygaard: I agree to this.

> Maybe we should ask ourselves a question: should point.setState(), called by user, alter other points' state?

Networkgraph has been working around this issue by having a [third undocumented variable](https://github.com/highcharts/highcharts/blob/d18eaabe4749dff19afac6f66f7ccdd4c13d973d/ts/modules/networkgraph/networkgraph.src.ts#L985), but I believe what you describe is a cleaner solution.

pawelfus: Just to be on the same page:
The third argument, is actually the second argument and it's [documented](https://api.highcharts.com/class-reference/Highcharts.Series#setState) - note it's `series.setState()`, not `point.setState()`.

ts/parts-map/HeatmapSeries.ts
jon-a-nygaard: @bre1470 can this comment thread be closed as resolved?

ts/parts-map/HeatmapSeries.ts
jon-a-nygaard: Yes, you are right. The Networkgraph `Point.setState` is defined at [mixins/nodes.ts#L283-L309](https://github.com/highcharts/highcharts/blob/d18eaabe4749dff19afac6f66f7ccdd4c13d973d/ts/mixins/nodes.ts#L283-L309), and it seems that altering of other point states can not be toggled as it is now.

ts/parts-map/HeatmapSeries.ts
bre1470: Yes, it resolved.

ts/parts-map/HeatmapSeries.ts
pawelfus: It's refactored now, no more `setNodeState` method, everything is done using `inactiveFilters` in all node-based series.

bre1470:
It is a pity. With all the migration we now have huge merge conflicts. Unsure if the work should start from stretch. Maybe after tree shaking is completed?

pawelfus:
> It is a pity. With all the migration we now have huge merge conflicts. Unsure if the work should start from stretch. Maybe after tree shaking is completed?

Don't worry, that was just a draft (I think before GH drafts were introduced)

PR #11551, opened 2019-08-02
ts/colortype

bre1470:
Introduced ColorType placeholder.

### DIFF
--- a/ts/parts/Color.ts
+++ b/ts/parts/Color.ts
@@ -21,6 +21,7 @@ declare global {
         type ColorNone = [];
         type ColorRGBA = [number, number, number, number];
         type ColorString = string;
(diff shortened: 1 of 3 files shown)


KacperMadej:
@bre1470 

How about other usages of the full type? like: https://github.com/highcharts/highcharts/blob/dd07861701dc09c67082df2edc2cae5e002d5d0c/ts/parts/CandlestickSeries.ts#L22-L39

bre1470:
@KacperMadej In our internal namespace we can make use of it. I avoided to apply it for now, to reduce the risk of merge conflicts. It is not decided though, if this gets official used in API options and class reference - that is something @TorsteinHonsi has to decide.

PR #11559, opened 2019-08-05
Tools: Added linting to CI as per request from Sophie.

activist:
Sometimes when doing merges the local linting isn't always been rerun upon commits.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -230,7 +230,7 @@ workflows:
       - test_browsers:
           name: "test-Mac.Safari"
           requires:
-            - lint
+            - install_dependencies
           browsers: "Mac.Safari"
       - test_browsers:
           name: "test-Win.Chrome"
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -214,13 +221,16 @@ workflows:
...
           requires:
             - install_dependencies
+      - generate_ts_declarations:
+          requires:
+            - lint
       - test_browsers:
           name: "test-Mac.Safari"
           requires:
-            - install_dependencies
+            - lint
```
bre1470: We can leave it to `install_dependencies` here imho. It should be enough, if `generate_ts_declarations` gets linted.

activist: Ok, that means that generating the ts declarations a will run parallell to the `gulp lint` / continue to run even if the `gulp lint`  fails - or I can add the linting as the last step of install_dependencies. Is that ok?

bre1470: I though that we have:
```
checkout->dependencies->lint->lint-dts
                     \->Safari->Chrome->Firefox->Edge->IE
```

activist: Changed it now as discussed :)

bre1470: Thank you! :)
Looks good. 👍

PR #11619, opened 2019-08-06
docs/11555-axis-options

### LINKED ISSUE #11555: typescript: Axis class should have minRange field
#### Expected behaviour

`minRange` field should be present in typescript definition of `Axis` class

#### Actual behaviour

No `minRange` field in the typescript definition of Axis class
 
 

#### Product version
highcharts 7.1.2

bre1470:
Fixed #11555, Made `options` and `userOptions` properties in Axis class public.

@TorsteinHonsi Please confirm, that these can and should be public.


TorsteinHonsi:
> @TorsteinHonsi Please confirm, that these can and should be public.

Confirmed, thanks!

PR #11626, opened 2019-08-07
bugfix/11477-export-data-parallel-coords

### LINKED ISSUE #11477: Parallel coordinates - exporting CSV data
#### Expected behaviour

Don't parse whole data into a date format, just data attached to datetime axis.

#### Actual behaviour

Whole data is parsing into a date format, which for small or negative data could be unwanted. 

[Demo](https://jsfiddle.net/BlackLabel/akjdsxhy/) showing the problem.

#### Product version

pawelfus:
Fixed #11477, exporting parallel coordinates chart to CSV did not use correct formats for y-values.
___
I'm not sure about the name and placement for `getCategoryAndDatetimeMap`. It's just extracted part of code to be a function.


TorsteinHonsi:
I'm sorry for the delay @pawelfus, there are now conflicts because of the `export-data.src.js` file being ported to TS. Can you port your changes too?

pawelfus:
Done, let me ask @bre1470 for a review, I have some doubts about TS

ts/modules/export-data.src.ts
```diff
@@ -471,6 +476,42 @@ Highcharts.Chart.prototype.getDataRows = function (
...
+        getCategoryAndDatetimeMap = function (
+            series: Highcharts.Series,
+            pointArrayMap: Array<string>,
+            pIdx?: number
+        ): Highcharts.CategoryAndDatetimeMap {
+            var categoryMap: Highcharts.CategoryAndDatetimeMap['categoryMap'] =
+                {},
+                datetimeValueAxisMap:
+                Highcharts.CategoryAndDatetimeMap['datetimeValueAxisMap'] =
+                {};
```
pawelfus: @bre1470 - can I do something better here? It's uber long, doesn't look good, hard to read etc.

bre1470: You can create some `type Alias = ...` before line 27 and use them in line 28 and 29 and here in line 485 and 488.
And regarding the naming I suggest to prefix all with `Exporting` because otherwise you will have a lot of fun to find it later on. For example the interface `CategoryAndDatetimeMap` should be better named `ExportingCategoryAndDatetimeMap`.

bre1470: I also think that the `And` in the name is not necessary.

bre1470: Maybe I am to much influenced by C#, but I find it strange to write `Datetime` instead of `DateTime`.

pawelfus: Thanks!

pawelfus: Changed.

Issue #11294, opened 2019-07-03
Make SVG Pie Chart text selectable including legend items and data values


constantintaivan:
Hello,
#### Expected behaviour
SVG Pie Chart legend and data labels should be selectable to be able to copy/paste values.

#### Actual behaviour
Text is not selectable. I've tried styling but it does not work. Not sure what I'm missing or if this is not yet supported.
svg text {
    -moz-user-select: text;
    -khtml-user-select: text;
    -webkit-user-select: text;
    -ms-user-select: text;
    user-select: text;
    pointer-events: all;
    cursor: default;
}

#### Live demo with steps to reproduce

http://jsfiddle.net/8ndx0m2b/

#### Product version

 Highcharts JS v7.0.1 (2018-12-19)

#### Affected browser(s)
All

sebastianbochan:
Hi @constantintaivan,
Thank you for contacting us. At this moment the (selection and copy/paste) functionality is not supported yet. 

I advice you to vote for this idea in our uservoice here:
- https://highcharts.uservoice.com/forums/55896-highcharts-javascript-api/suggestions/12882225-title-and-chartnames-to-be-selectable-with-mouse

Im marking this suggestion as enhancement.

constantintaivan:
Thank you @sebastianbochan! I'm going to get more votes on this.

sebastianbochan:
Unstale

PR #11629, opened 2019-08-07
Added new option group, caption

TorsteinHonsi:
Added new top-level option, [caption](https://api.highcharts.com/highcharts/caption). The caption has similar structure to title and subtitle. It renders a chart description below the chart contents, supports reponsiveness and is part of exported charts.
___

* [jsFiddle demo](http://jsfiddle.net/highcharts/s5vy2k8n/)
* [UserVoice request](https://highcharts.uservoice.com/forums/55896-highcharts-javascript-api/suggestions/2948242-add-footnote-textbox-immediately-below-chart)

#### Note for core developers
The implementation is quite light weight, piggy-backing the options and layout functionality of title and subtitle, so it is not too much of a bloat.

We need to do a review on how this affects a11y. Are there any common areas with chart description options? Any potential for combining? @oysteinmoseng .

#### To do
 - [x] Fix Highstock crash with navigator
 - [ ] Not decided: Implement `maxHeight` with a default percentage setting. The problem is we don't currently have a mechanism to limit the height of a `text` or `label` element, so this one goes beyond the caption implementation. That said, a `height` or `maxHeight` feature would also benefit axis labels and data labels.
 - [x] Accessibility implementation.
 - [x] Fix issue with [bottom range selector](http://jsfiddle.net/BlackLabel/ax6j24ud/)
 - [x] Implement `caption.margin` option, separated from `title.margin`.

### DIFF
--- a/js/parts/Options.js
+++ b/js/parts/Options.js
@@ -27,7 +27,7 @@ import H from './Globals.js';
  * @return {void}
  */
 /**
- * Conaints common event information. Through the `options` property you can
+ * Contains common event information. Through the `options` property you can
  * access the series options that were passed to the `addSeries` method.
  *
  * @interface Highcharts.ChartAddSeriesEventObject
@@ -1690,8 +1690,8 @@ H.defaultOptions = {
          */
         /**
          * The vertical alignment of the title. Can be one of `"top"`,
-         * `"middle"` and `"bottom"`. When a value is given, the title behaves
-         * as floating.
+         * `"middle"` and `"bottom"`. When middle, the subtitle behaves as
+         * floating.
          *
          * @sample {highcharts} highcharts/subtitle/verticalalign/
          *         Footnote at the bottom right of plot area
@@ -1764,6 +1764,84 @@ H.defaultOptions = {
          * @since 4.2.5
          */
         widthAdjust: -44
+    },
+    /**
+     * The chart's caption, which will render below the chart and will be part
+     * of exported charts. The caption can be updated after chart initialization
+     * through the `Chart.update` or `Chart.caption.update` methods.
+     *
+     * @sample highcharts/caption/text/
+     *         A chart with a caption
+     * @since  next
+     */
+    caption: {
+        /**
+         * When the caption is floating, the plot area will not move to make
+         * space for it.
+         *
+         * @type      {boolean}
+         * @default   false
+         * @apioption caption.floating
+         */
+        /**
+         * CSS styles for the caption.
+         *
+         * In styled mode, the caption style is given in the
+         * `.highcharts-caption` class.
+         *
+         * @sample {highcharts} highcharts/css/titles/
+         *         Styled mode
+         *
+         * @type      {Highcharts.CSSObject}
+         * @default   {"color": "#666666"}
+         * @apioption caption.style
+         */
+        /**
+         * Whether to
+         * [use HTML](https://www.highcharts.com/docs/chart-concepts/labels-and-string-formatting#html)
+         * to render the text.
+         *
+         * @type      {boolean}
+         * @default   false
+         * @apioption caption.useHTML
+         */
+        /**
+         * The x position of the caption relative to the alignment within
+         * `chart.spacingLeft` and `chart.spacingRight`.
+         *
+         * @type      {number}
+         * @default   0
+         * @apioption caption.x
+         */
+        /**
+         * The y position of the caption relative to the alignment within
+         * `chart.spacingTop` and `chart.spacingBottom`.
+         *
+         * @type      {number}
+         * @apioption caption.y
+         */
+        /**
+         * The caption text of the chart.
+         *
+         * @sample {highcharts} highcharts/caption/text/
+         *         Custom caption
+         */
+        text: '',
+        /**
+         * The horizontal alignment of the caption. Can be one of "left",
+         *  "center" and "right".
+         *
+         * @type  {Highcharts.AlignValue}
+         */
+        align: 'left',
+        /**
+         * The vertical alignment of the caption. Can be one of `"top"`,
+         * `"middle"` and `"bottom"`. When middle, the caption behaves as
+         * floating.
+         *
+         * @type      {Highcharts.VerticalAlignValue}
+         */
+        verticalAlign: 'bottom'
     },
     /**
      * The plotOptions is a wrapper object for config objects for each series
--- a/js/parts/Options.js
+++ b/js/parts/Options.js
@@ -1784,6 +1784,10 @@ H.defaultOptions = {
          * @apioption caption.floating
          */
         /**
+         * The margin between the caption and the plot area.
+         */
+        margin: 15,
+        /**
          * CSS styles for the caption.
          *
          * In styled mode, the caption style is given in the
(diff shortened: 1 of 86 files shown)


ts/parts/Chart.ts
```diff
@@ -33,6 +33,9 @@ declare global {
             plotBox: BBoxObject;
             spacingBox: BBoxObject;
         }
+        interface CaptionObject extends SVGElement {
```
bre1470: Is this used anywhere, otherwise it should be removed?

oysteinmoseng:
Very cool! This definitely overlaps with a11y concerns on multiple levels, and also with Cloud.

My gut feeling says to display `chart.accessibility.description` in the caption by default - and the other way around, depending on which is defined, but we might want to think this through pretty carefully. I'll take a closer look tomorrow.

ts/parts/Chart.ts
```diff
@@ -1051,119 +1059,121 @@ extend(Chart.prototype, /** @lends Highcharts.Chart.prototype */ {
...
+     *        Either title, subtitle or caption
+     * @param {Highcharts.TitleOptions|Highcharts.SubtitleOptions|Highcharts.CaptionOptions|undefined} explicitOptions
+     *        The options to set, will be merged with default options.
+     *
+     * @return {void}
+     */
+    applyDescription: function (
+        this: Highcharts.Chart,
+        name: ('title'|'subtitle'|'caption'),
+        explicitOptions: (Highcharts.DescriptionOptionsType|undefined)
```
bre1470: This should be written as `explicitOptions?: Highcharts.DescriptionOptionsType`.

ts/parts/Chart.ts
```diff
@@ -119,6 +123,10 @@ declare global {
             public xAxis: Array<Axis>;
             public yAxis: Array<Axis>;
             public addCredits(credits?: CreditsOptions): void;
+            public applyDescription(
+                name: ('title'|'subtitle'|'caption'),
+                options: (DescriptionOptionsType|undefined)
```
bre1470: Here it has to be `explicitOptions?: DescriptionOptionsType`.

ts/parts/Chart.ts
```diff
@@ -1051,119 +1059,121 @@ extend(Chart.prototype, /** @lends Highcharts.Chart.prototype */ {
...
-            var name = arr[0] as string,
-                title = (chart as any)[name],
-                titleOptions = arr[1],
-                chartTitleOptions = arr[2] as (
-                    Highcharts.TitleOptions|Highcharts.SubtitleOptions
-                );
+            // Update methods, shortcut to Chart.setTitle, Chart.setSubtitle and
+            // Chart.setCaption
+            elem.update = function (
+                o: (Highcharts.DescriptionOptionsType)
```
bre1470: I suggest a more descriptive variable like (at least) `opt`. Compiler will reduce it anyway.

js/parts/Options.js
```diff
@@ -1765,6 +1765,84 @@ H.defaultOptions = {
...
+    /**
+     * The chart's caption, which will render below the chart and will be part
+     * of exported charts. The caption can be updated after chart initialization
+     * through the `Chart.update` or `Chart.caption.update` methods.
+     *
+     * @sample highcharts/caption/text/
+     *         A chart with a caption
+     * @since  next
+     */
+    caption: {
```
jon-a-nygaard: In my opinion `caption.enabled` would be useful to keep it consistent with other features.
I see that this is similar to `title`, but I believe to have the option `enabled` as in `credits` is a nicer pattern because you can just toggle it on an off.

ts/parts/Chart.ts
```diff
@@ -1051,119 +1059,121 @@ extend(Chart.prototype, /** @lends Highcharts.Chart.prototype */ {
...
-            }
-        });
-        chart.layOutTitles(redraw);
     },
 
     /**
-     * Internal function to lay out the chart titles and cache the full offset
-     * height for use in `getMargins`. The result is stored in
-     * `this.titleOffset`.
+     * Internal function to lay out the chart title, subtitle and caption and
```
bre1470: Commas missing: `Internal function to lay out the chart title, subtitle, and caption, and`

activist:
A really nice feature  :+1:  

Although you are probably aware there were two minor things that I noticed:
* The caption seems to be cut off on exported bitmaps / pdf.
* I have some challenges when trying to highlight/select and copy the caption, but that might be due to the nature of SVGs?

oysteinmoseng:
> * I have some challenges when trying to highlight/select and copy the caption, but that might be due to the nature of SVGs?

`<text>` elements in SVG can be selected (as evidenced by downloading the SVG and opening in a browser), but I think Highcharts might be messing with this by overriding mousedown events.. It is something we should ideally fix.

pawelfus:
Cool, simple feature yet quite common for Cloud!

Is it worth implementing Title/Description Class? With full lifecycle (`create`/`init`, `render`/`redraw`, `update`, `destroy`)? It's much easier to debug and maintain when we have these methods in one place (file) and chart is only maintaining the plotting area and delegates methods (like `update()`). Or that's the `so it is not too much of a bloat.` part in comment? :)

> <text> elements in SVG can be selected (as evidenced by downloading the SVG and opening in a browser), but I think Highcharts might be messing with this by overriding mousedown events.. It is something we should ideally fix.

Yes, we are messing with this. This is caused by all drag&drop (e.g. panning, zooming, selecting points stc) actions, where users don't want to select anything on the chart, but perform an action.

oysteinmoseng:
As to the panning/zoom etc features, could we only prevent the default browser handling in the cases where we are actually performing a custom action? This sounds like it should be easy to do.

KacperMadej:
1. Looks bad in Highstock: http://jsfiddle.net/BlackLabel/zx29waym/
2. No overflow protection: http://jsfiddle.net/BlackLabel/zx29waym/1/

oysteinmoseng:
Regarding the text selection issue we can follow that up here: https://github.com/highcharts/highcharts/issues/11294

bre1470:
I suggest to deprecate `Chart.setSubtitle` with the changes as it adds no functionality in addition to `Chart.setTitle`.

js/parts/Options.js
TorsteinHonsi: In that case we need to change title and subtitle too

TorsteinHonsi:
> The caption seems to be cut off on exported bitmaps / pdf.

Not for me... Probably that may happen if the fonts used on the client are not the same as the ones on the export server.

> Is it worth implementing Title/Description Class? With full lifecycle (create/init, render/redraw, update, destroy)?

Yes it could probably be worthwhile, instead of mocking lifecycle like we are currently doing. It can be done in a separate refactoring step though.

> Looks bad in Highstock

Ouch.

> No overflow protection

Subtitle and title also have no overflow protection, but of course it is more of a problem with the caption. I suggest a `maxHeight` option like that of the legend, which can be a percentage or pixel setting.

> I suggest to deprecate Chart.setSubtitle with the changes as it adds no functionality in addition to Chart.setTitle.

It (as well as `setCaption`) is used internally in `Chart.update` to map setters to the options.

ts/parts/Chart.ts
TorsteinHonsi: It is equivalent to `SubtitleObject`. I forgot to add doclets for `CaptionObject` and `CaptionObject#update`.

ts/parts/Chart.ts
TorsteinHonsi: Hmm. I don't think there should be a comma after "subtitle" (at least it's optional), but one after "caption" is probably needed.

sebastianbochan:
Looks bad with the bottom rangeSelector.

Demo:
- http://jsfiddle.net/BlackLabel/ax6j24ud/

PR #11631, opened 2019-08-07
ts/destroy-undefined

bre1470:
Changed some return types of destroy functions.


bre1470:
For internal arrays of elements we should consider a regular loop with a `Array.length = 0` afterwards.

PR #11637, opened 2019-08-08
Bugfix/11493 split tooltip 2

### LINKED ISSUE #7650: Hovered points outside plotting area

### LINKED ISSUE #7711: yAxis display out of pane after setExtremes

### LINKED ISSUE #11493: Y Axis scaled using set Extremes Is inappropriate

jon-a-nygaard:
Fixed #7650, hide hover points outside plot area. Fixed #11493, limit split tooltip to yAxis range.

---
# Description
Reverts #11531 due to unwanted side effects. Please see https://github.com/highcharts/highcharts/issues/11493#issuecomment-519074331 and https://github.com/highcharts/highcharts/issues/7711#issuecomment-518251273.
Applies a different solution to fix the related issues.

- Halo gets the same visibility as point.graphic or the stateMarkerGraphic
- stateMarkerGraphic is hidden if `point.isInside === false`
- Split tooltip target and anchorY is limited to (`yAxis.pos < target < yAxis.pos + yAxis.len`). Height of tooltip is also considered for `target`.

# Example(s)
- [Demo of issue #7650](https://jsfiddle.net/jon_a_nygaard/z0o94uwm/)
- [Demo of issue #7650 with fix applied](https://jsfiddle.net/jon_a_nygaard/9bq1jv4k/)
- [Demo of issue #11493](https://jsfiddle.net/jon_a_nygaard/nhudqsfa/)
- [Demo of issue #11493 with fix applied](https://jsfiddle.net/jon_a_nygaard/9weaxh2j/)

# Related issue(s)
Closes #7650
Closes #11493
Closes #7711


jon-a-nygaard:
IE8 is not happy with something related to visibility, I am currently looking into fixing this.

ts/parts/Interaction.ts
```diff
@@ -1264,6 +1259,9 @@ extend(Point.prototype, /** @lends Highcharts.Point.prototype */ {
                 'class': 'highcharts-halo highcharts-color-' +
                     pick(point.colorIndex, series.colorIndex) +
                     (point.className ? ' ' + point.className : ''),
+                'visibility': (
+                    (point.graphic || stateMarkerGraphic) as any
+                ).visibility,
```
pawelfus: Would it be better to move this into the condition above? So not rendering halo at all, when point/shared marker is not visible? By the condition above I mean:
https://github.com/highcharts/highcharts/blob/0bb12294b4f4f6c03b4b265488535e0e956b6039/ts/parts/Interaction.ts#L1246

jon-a-nygaard: Thanks, that is a good point. Will make the necessary change.

It will only be optimal when `(point.graphic || stateMarkerGraphic) as any).visibility` equals `hidden` though. Since `inherit` can be both `'visible'|'hidden'` depending on the parents, it will have to be rendered in this case.

jon-a-nygaard: @pawelfus Please see latest changes.

pawelfus: Looks good, thanks!

PR #11643, opened 2019-08-08
Export attr in parts/Utilities.js

jon-a-nygaard:
Title says it all.

Please take an extra look on the TypeScript function overloads to make sure these are good enough.

### DIFF
--- a/ts/parts/Utilities.ts
+++ b/ts/parts/Utilities.ts
@@ -157,9 +157,18 @@ declare global {
         function arrayMin(data: Array<any>): number;
         function attr(
             elem: (HTMLDOMElement|SVGDOMElement),
-            prop?: (string|HTMLAttributes|SVGAttributes),
-            value?: (number|string)
-        ): any;
+            prop: (HTMLAttributes|SVGAttributes)
+        ): undefined;
+        function attr(
+            elem: (HTMLDOMElement|SVGDOMElement),
+            prop: string,
+            value?: undefined
+        ): (string|null);
+        function attr(
+            elem: (HTMLDOMElement|SVGDOMElement),
+            prop: string,
+            value: (number|string)
+        ): undefined;
         function clearTimeout(id: number): void;
         function correctFloat(num: number, prec?: number): number;
         function createElement(
@@ -1427,6 +1436,20 @@ function defined<T>(obj: T): obj is NonNullable<T> {
     return typeof obj !== 'undefined' && obj !== null;
 }
 
+function attr(
+    elem: (Highcharts.HTMLDOMElement|Highcharts.SVGDOMElement),
+    prop: (Highcharts.HTMLAttributes|Highcharts.SVGAttributes)
+): undefined;
+function attr(
+    elem: (Highcharts.HTMLDOMElement|Highcharts.SVGDOMElement),
+    prop: string,
+    value?: undefined
+): (string|null);
+function attr(
+    elem: (Highcharts.HTMLDOMElement|Highcharts.SVGDOMElement),
+    prop: string,
+    value: (number|string)
+): undefined;
 /**
  * Set or get an attribute or an object of attributes. To use as a setter, pass
  * a key and a value, or let the second argument be a collection of keys and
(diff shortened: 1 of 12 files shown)


ts/parts/Utilities.ts
```diff
@@ -1427,6 +1444,23 @@ function defined<T>(obj: T): obj is NonNullable<T> {
     return typeof obj !== 'undefined' && obj !== null;
 }
 
+function attr(
+    elem: (Highcharts.HTMLDOMElement|Highcharts.SVGDOMElement)
+): undefined;
```
bre1470: I do not see any value in this definition. It is possible, true, but not useful.

ts/parts/Utilities.ts
```diff
@@ -1427,6 +1444,23 @@ function defined<T>(obj: T): obj is NonNullable<T> {
...
+function attr(
+    elem: (Highcharts.HTMLDOMElement|Highcharts.SVGDOMElement),
+    prop: string,
+    value?: undefined
+): (string|null);
+function attr(
+    elem: (Highcharts.HTMLDOMElement|Highcharts.SVGDOMElement),
+    prop: string,
+    value: (number|string)
+): undefined;
```
bre1470: You could combine it with the previous declaration to a conditional return type:
```ts
function attr<T extends (number|string|undefined)>(
     elem: (Highcharts.HTMLDOMElement|Highcharts.SVGDOMElement),
     prop: string,
     value?: T
 ): T extends undefined ? (number|string) : undefined;
```

bre1470: I realized that you can not overload this so good, so three declarations are necessary as it seams.

ts/parts/Utilities.ts
jon-a-nygaard: I agree that it is not useful, and probably a bug if used like this, but this is how it is currently implemented.

If we should remove this overload, then I would prefer to also correct the implementation.

bre1470: Still the compiler should not accept it.

bre1470: I actually do not see any problem with the implementation. You just have to remove the optional flag for the second parameter.

jon-a-nygaard: > remove the optional flag

Yes, that is a good suggestion, but then we can also optimize the following:
```javascript
} else if (defined(prop) && isObject(prop)) {
```
to:
```javascript
} else {
```
because `prop` is always defined, and if it is not a string then it has to be an object.

So I still believe an adjustment of the implementation is in order.

bre1470: Agreed. The adjustment will cover it.

jon-a-nygaard: Great, will make the changes ASAP

jon-a-nygaard: @bre1470  and @TorsteinHonsi please see latest changes.

ts/parts/Utilities.ts
```diff
@@ -157,9 +157,23 @@ declare global {
...
+            elem: (HTMLDOMElement|SVGDOMElement),
+            prop: string,
+            value: (number|string)
+        ): undefined;
+        function attr(
+            elem: (HTMLDOMElement|SVGDOMElement),
+            prop: (string|HTMLAttributes|SVGAttributes),
             value?: (number|string)
-        ): any;
+        ): (string|null|undefined);
```
bre1470: I am not sure, if this one is necessary. It looks like a merge of the three previous declarations.

ts/parts/Utilities.ts
bre1470: This looks good.

ts/parts/Utilities.ts
bre1470: I believe it is only for the actual implementation down in the code needed.

jon-a-nygaard: Yes, I believe you are right. This function type is not needed, since the more specific overloads will catch it first.
Will make another update.

jon-a-nygaard: @bre1470 Please see latest changes.

PR #11649, opened 2019-08-09
Fixed #11199, step miscalculated in axis when range is 0

### LINKED ISSUE #11199: grid line placement broken for single category y axis
#### Expected behaviour
There should be a grid line on the top of the axis and one on the bottom.

#### Actual behaviour
The top grid line has moved to the middle, going straight through the points.

broken in 7.0.2: https://jsfiddle.net/w0xfr2d3/
working in 7.0.1: https://jsfiddle.net/qu1pc5z2/

This only happens when there is only one category.

#### Product version
Highcharts 7.0.2 and above

#### Affected browser(s)
Firefox, chrome, Internet Explorer 11

robahs:
Fixed issue #11199


TorsteinHonsi:
Thanks for the fix, I'll merge it into a local branch and add a unit test.

Issue #11592, opened 2019-08-05
Migrate `modules/sankey` to TypeScript


bre1470:

bre1470:
- This needs code changes before it can migrated. Basically the point class is absolutely not compatible with the nodes mixin and therefore TypeScript will not accept the point type in the `Series.nodes` property.
- Additionally I wonder if we really need special node options in addition to the regular point options. This should in my understanding be the same.
- Another problem is the node formatters context which extends from point, but in the implementation expects instead `content.point` property for accessing the point.

PR #11662, opened 2019-08-12
docs/fix-sankey

bre1470:
Fixed sankey doclets.


bre1470:
Preparation for #11592

PR #11675, opened 2019-08-13
bugfix/11344-stock-icons-iconsURL

sebastianbochan:
Touch #11344, set common iconsURL for popup and stocktools.

### DIFF
--- a/js/annotations/navigationBindings.js
+++ b/js/annotations/navigationBindings.js
@@ -1365,10 +1365,9 @@ H.setOptions({
          * Path where Highcharts will look for icons. Change this to use
          * icons from a different server.
          *
-         * @since 7.1.3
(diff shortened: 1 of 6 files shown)


js/annotations/navigationBindings.js
```diff
@@ -1361,6 +1361,16 @@ H.setOptions({
          * @apioption navigation.events.deselectButton
          */
 
+        /**
+         * Path where Highcharts will look for icons. Change this to use
+         * icons from a different server.
+         *
+         * @since 7.1.3
+         * @apioption @navigation.iconsURL
```
bre1470: `@apioption` must be last tag in doclet and option path must start witch letter.

bre1470: @TorsteinHonsi will fix it in master

PR #11685, opened 2019-08-14
Added high contrast themes.

oysteinmoseng:
Added `high-contrast-light` and `high-contrast-dark` themes with 3:1 contrast against background.
___
Also added skipTest demos of both for use in docs. Themes have 10 series colors that all comply with the WCAG requirement of 3:1 contrast against background for graphical elements. Neighboring colors are tested for color blindness.

Tweaks appreciated as it is hard to find 10 colors that play well together within these constraints.

Looks like this:

Screenshot 2019-08-14 at 19 31 35
Screenshot 2019-08-14 at 19 31 44

### DIFF
--- a/js/masters/themes/high-contrast-dark.src.js
+++ b/js/masters/themes/high-contrast-dark.src.js
@@ -1,0 +1,11 @@
+/**
+ * @license Highcharts JS v@product.version@ (@product.date@)
+ * @module highcharts/themes/high-contrast-dark
+ * @requires highcharts
+ *
+ * (c) 2009-2019 Highsoft AS
+ *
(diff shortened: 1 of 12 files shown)


TorsteinHonsi:
@jon-a-nygaard Shouldn't the themes be defined as `.src.js` files?

oysteinmoseng:
Just the master files, if the existing themes is an indicator

oysteinmoseng:
Sophie says:
> I do not believe that there is a real conversion necessary until we figured out how to do ES6 module loading. In the end, they just set a bunch of options, so some might have some interface declarations, but that is all when it comes to TypeScript.

PR #11694, opened 2019-08-15
Export syncTimeout in parts/Utilities.ts

jon-a-nygaard:
Title says it all.

Please take an extra look at the function overloads to make sure these are good enough.

### DIFF
--- a/ts/parts/Utilities.ts
+++ b/ts/parts/Utilities.ts
@@ -313,9 +313,9 @@ declare global {
         function stop(el: SVGElement, prop?: string): void;
         function syncTimeout(
             fn: Function,
-            delay?: number,
-            context?: any
-        ): (number|undefined);
+            delay: number,
+            context?: unknown
+        ): number;
         function uniqueKey(): string;
         function wrap(
             obj: any,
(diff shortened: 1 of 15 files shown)


bre1470:
I agree. This can be reduced to a single function.

jon-a-nygaard:
@TorsteinHonsi and @bre1470 please see latest changes.

ts/parts/Utilities.ts
```diff
@@ -1525,26 +1525,27 @@ function splat(obj: any): Array<any> {
...
  */
-H.syncTimeout = function (
+function syncTimeout(
     fn: Function,
-    delay?: number,
-    context?: any
-): (number|undefined) {
-    if (delay) {
+    delay: number,
+    context?: unknown
```
bre1470: Type in namespace has to change as well.

jon-a-nygaard: Thanks, should be fixed in last update.

jon-a-nygaard:
I see there is now a conflict between master and this branch, I will solve it locally and force push an updated version of the branch.

PR #11703, opened 2019-08-16
Tools/visual test fixes

TorsteinHonsi:
- Fixed the override for `$.getJSON` to look for `Highcharts.getJSON` and files loaded from jsdelivr, will now catch those and insert content from local sample files.
- Added more Highcharts modules, allowing more samples to pass.
- Invalidated external `image` tags, allowing more SVGs to compare.
- Validated stock and maps demo. We can now run with `--tests highcharts/demo/*,maps/demo/*,stock/demo/*`.


activist:
Looks good, but tests are failing. 

As for Interaction and Tooltip I assume changes are good :+1:

PR #11704, opened 2019-08-16
 bugfix/11608-boost-points-objects-extremes

### LINKED ISSUE #11608: Tooltips not showing on scatter plot running with boost module
#### Expected behaviour
When hovering mouse to the plot point should show tooltips

#### Actual behaviour
Hovering mouse to data point didnt show anything

https://jsfiddle.net/p13mcLqv/1/

#### Product version
Highstock 7.1.2

#### Affected browser(s)
Chrome

pawelfus:
Fixed #11608, tooltip was not displayed in boosted mode when yAxis extremes were set and points were configured as objects (`{ x, y }`).

### DIFF
--- a/js/modules/boost/boost-init.js
+++ b/js/modules/boost/boost-init.js
@@ -193,8 +193,8 @@ function init() {
 
                 if (!chartDestroyed) {
                     if (useRaw) {
-                        x = d[0];
-                        y = d[1];
+                        x = H.pick(d[0], d.x);
+                        y = H.pick(d[1], d.y);
                     } else {
                         x = d;
                         y = yData[i];
--- a/samples/unit-tests/boost/tooltip/demo.js
+++ b/samples/unit-tests/boost/tooltip/demo.js
@@ -1,4 +1,4 @@
-QUnit.test('Tooltip on a boosted chart with categories', function (assert) {
+QUnit.test('Tooltip on a boosted chart', function (assert) {
     var chart = Highcharts.chart('container', {
             xAxis: {
                 categories: ['categoryName', 'B', 'C']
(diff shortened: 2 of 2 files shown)


cvasseng:
Agree with @TorsteinHonsi - at the very least it should be expanded to be e.g. `d.x || d[0]` to avoid the function call overhead. There may also be other point object logic that fails in weird places after opening this door as boost was written explicitly without support for them, so we need to decide to what degree we want to support it to avoid confusions/communicate clear expectations.

pawelfus:
Good point with `pick()` vs `typeof` check. Actually, `pick` is even worse, because `null` values won't be handled properly. 

I consider fixing this a good idea - it seems that only tooltip doesn't work, rendering is fine.

cvasseng:
...there's actually a further optimization that can be done too:

It's safe to assume that if a couple points have `.x|.y` set, the whole set does (or is expected to). So you could sample a couple of points before looping (for instance the first and last), check for the existance of `.x|.y`, and set a bool. Then just check the bool in the loop. That will add almost no overhead (doing a typecheck 2M times is not cheap). :grimacing:

pawelfus:
I think we do that for the first point to check `turboThreshold` - if the first point is an object config then error (12) is thrown.

TorsteinHonsi:
> There may also be other point object logic that fails in weird places after opening this door as boost was written explicitly without support for them

Yes. I don't think we should open this door. Boost should be optimized for speed only. For example, arearange series will fail if we apply axis extremes (like we recommend): https://jsfiddle.net/highcharts/ov3gwkf5/1/

pawelfus:
Right, good point, thanks! 

How about adding an error (or a warning) to our list? Alternatively improving description for https://www.highcharts.com/errors/12/ ?

TorsteinHonsi:
> Alternatively improving description for https://www.highcharts.com/errors/12/ ?

Yes, I think that's the way to go. So the idea here now will be that when boosting, turbo mode is forced.

PR #11710, opened 2019-08-19
bugfix/11290-solid-gauge-leaks

### LINKED ISSUE #11290: SolidGauge leaks labelCollectors
#### Expected behaviour
Updating options of a solidGauge should not degrade performance.

#### Actual behaviour
Every time a solidgauge is updated like this:
```
    chartSpeed.update({
      yAxis: {
        min: 0,
        max: 200,
        title: {
          text: 'Speed'
        }
      },

      series: [{
        data: [10],
      }]
```
a new function is added to Chart.labelCollectors. This happens in RadialAxis.js, line 537 (the init event). There is no corresponding cleanup event. 
If the update is done periodically, the application quickly becomes irresponsive. When profiling, you see that a lot of time is spent in hideOverlappingLabels.

http://jsfiddle.net/zmu9wfov/

#### Product version
Highcharts 7.1.2

#### Affected browser(s)
Tested on Chrome

pawelfus:
Fixed #11290, updating `yAxis` in solid gauge series caused memory leaks.
___
Please review the TS too. I'm not sure about `this.labelCollector as Highcharts.ChartLabelCollectorFunction` - I simply want `this.labelCollector` to be either function or not defined/false.

### DIFF
--- a/js/parts-more/RadialAxis.js
+++ b/js/parts-more/RadialAxis.js
@@ -379,13 +379,34 @@ radialAxisMixin = {
                     center[2]) +
                 (titleOptions.y || 0))
         };
+    },
+    /* *
+     * Attach and return collecting function for labels in radial axis for
+     * anti-collision.
+     */
+    createLabelCollector: function () {
+        var axis = this;
+        return function () {
+            if (axis.isRadial &&
+                axis.tickPositions &&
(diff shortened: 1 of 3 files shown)


TorsteinHonsi:
Thanks! Awaiting @bre1470 's approval before merge.

PR #11714, opened 2019-08-19
bugfix/11546-parentNodes-missing-after-update

### LINKED ISSUE #11546: PackedBubble: Missing parentNodes after chart.update
#### Expected behaviour
Chart rendered the same way as before `chart.update()` (no missing parentNodes)

#### Actual behaviour
parentNodes vanished

Demo: https://jsfiddle.net/BlackLabel/7ut0aLf8/

#### Product version
v7

FYI @Izothep

Izothep:
Fixed #11546, in packed bubble series parentNodes were missing after chart and series update.

### DIFF
--- a/ts/parts-more/PackedBubbleSeries.ts
+++ b/ts/parts-more/PackedBubbleSeries.ts
@@ -1113,14 +1113,13 @@ seriesType<Highcharts.PackedBubbleSeriesOptions>(
                 width: (series.parentNodeRadius as any) * 2,
                 height: (series.parentNodeRadius as any) * 2
             }, parentOptions);
-            if (!series.graph) {
+            if (!(series.parentNode as any).graphic) {
                 series.graph = (series.parentNode as any).graphic =
                     chart.renderer.symbol(parentOptions.symbol)
                         .attr(parentAttribs)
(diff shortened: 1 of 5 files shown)


ts/parts-more/PackedBubbleSeries.ts
```diff
@@ -1113,14 +1113,13 @@ seriesType<Highcharts.PackedBubbleSeriesOptions>(
                 width: (series.parentNodeRadius as any) * 2,
                 height: (series.parentNodeRadius as any) * 2
             }, parentOptions);
-            if (!series.graph) {
+            if (!(series.parentNode as any).graphic) {
                 series.graph = (series.parentNode as any).graphic =
                     chart.renderer.symbol(parentOptions.symbol)
                         .attr(parentAttribs)
```
pawelfus: Is that line necessary? It's done after if-else. Or some of `parentAttribs` are expected **before** adding to DOM?

Izothep: Thanks for the feedback! It was a relict after removing `if-else` from the code, I have already removed unnecessary `attr` in the newest commit.

PR #11733, opened 2019-08-20
Tools: Added CI testing of stock and maps samples

activist:
.. and fixed test verification of additional samples.

Extracted karma js template for the visual test samples into separate function. 

Now an assert for checking if the chart exists is always run for all visual tests. I'm a bit divided with regards to splitting it up or not. I.e. sample verification could instead be done only when explicitly asking for it through an argument
E.g `karma start test/karma-conf.js  --browsercount 1 --tests highcharts/*/* --verify-only`

Also it is not apparent that the above will not run unit tests, but rather wrap demo files as QUnit tests. Should we make that more explicit or is it ok as it is?

### DIFF
--- a/test/karma-conf.js
+++ b/test/karma-conf.js
@@ -511,7 +511,12 @@ module.exports = function (config) {
                     // Skipped from demo.details
                     if (handleDetails(path) === false) {
                         file.path = file.originalPath + '.preprocessed.js';
-                        done(`QUnit.skip('${path}');`);
+                        if (argv.visualcompare) {
+                            // QUnit will explode when all tests within a module are skipped. Omitting test instead.
+                            done(`console.log('Not adding test ${path} due to being skipped from demo.html');`);
+                        } else {
+                            done(`QUnit.skip('${path}');`);
+                        }
                         return;
                     }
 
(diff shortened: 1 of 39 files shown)


test/karma-conf.js
```diff
@@ -397,7 +397,23 @@ module.exports = function (config) {
             'samples/highcharts/demo/organization-chart/demo.js',
             'samples/highcharts/demo/pareto/demo.js',
             'samples/highcharts/demo/pyramid3d/demo.js',
-            'samples/highcharts/demo/synchronized-charts/demo.js'
+            'samples/highcharts/demo/synchronized-charts/demo.js',
```
TorsteinHonsi: All these successfully create SVGs on the master using `--reference`. For the renderer samples, the clue is that they don't create a `chart` instance, therefor we look for `document.getElementByTagName('svg')` instead. We used to check for `chart || document.getElementByTagName('svg').length`, but in this PR it has changed to `&&`, which was not the intention.

As for the lates 3 samples, the `/maps/` ones, they load data from Google Spreadsheet. This also works in the master, both for reference and for comparison.

test/karma-conf.js
```diff
@@ -740,3 +680,74 @@ module.exports = function (config) {
...
+            
+            ${scriptBody}
+
+            let attempts = 0;
+            function assertIfChartExists() {
+                var chart = Highcharts.charts[
+                    Highcharts.charts.length - 1
+                ];
+
+                if (chart && document.getElementsByTagName('svg').length) {
```
TorsteinHonsi: Should be `chart || document.getElementsByTagName('svg').length` in order to allow renderer-only samples.

test/karma-conf.js
activist: This was a messup on my side. This is now fixed.

As as sidenote all the visual tests will be retried until there's a chart or svg found. For most tests this is really not necessary and we should consider making a separate and more explicit test template. However, for some of the maps/ samples that first loads after the (geo)json data has been retrieved it is necessary to retry (or we need to change the demos accordingly so that they more easily can be incorporated in a test).

test/karma-conf.js
activist: Fixed

PR #11737, opened 2019-08-21
ts/modules-draggable-points

### LINKED ISSUE #11572: Migrate `modules/draggable-points` to TypeScript

bre1470:
Fixed #11572, Migrated draggable-points to TypeScript.

### DIFF
--- a/ts/modules/draggable-points.src.ts
+++ b/ts/modules/draggable-points.src.ts
@@ -9,36 +9,211 @@
  *  !!!!!!! SOURCE GETS TRANSPILED BY TYPESCRIPT. EDIT TS FILE ONLY. !!!!!!!
  *
  * */
+
 'use strict';
+
 import H from '../parts/Globals.js';
+
+/**
+ * Internal types
+ * @private
+ */
+declare global {
+    namespace Highcharts {
+        interface Chart {
+            /** @requires modules/draggable-points */
+            dragDropData?: DragDropDataObject;
+            /** @requires modules/draggable-points */
+            dragHandles?: DragHandlesObject;
+            /** @requires modules/draggable-points */
+            dragGuideBox?: SVGElement;
+            /** @requires modules/draggable-points */
+            hasAddedDragDropEvents?: boolean;
+            /** @requires modules/draggable-points */
+            isDragDropAnimating?: boolean;
+            /** @requires modules/draggable-points */
+            unbindDragDropMouseUp?: Function;
+            /** @requires modules/draggable-points */
+            setGuideBoxState(
+                state: string,
+                options?: Dictionary<DragDropGuideBoxOptionsObject>
+            ): SVGElement;
+            /** @requires modules/draggable-points */
+            hideDragHandles(): void;
+            /** @requires modules/draggable-points */
+            zoomOrPanKeyPressed(e: Event): boolean;
+        }
+        interface ChartOptions {
+            zoomKey?: string;
+        }
+        interface DragDropDataObject {
+            draggedPastSensitivity?: boolean;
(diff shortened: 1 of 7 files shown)


ts/modules/draggable-points.src.ts
```diff
@@ -0,0 +1,3045 @@
...
+        interface PointDragStartCallbackFunction {
+            (this: Point, event: PointDragStartEventObject): void;
+        }
+        interface PointDragStartEventObject extends MouseEvent {
+            updateProp?: string;
+        }
+        interface PointDropCallbackFunction {
+            (this: Point, event: PointDropEventObject): void;
+        }
+        interface PointDropEventObject {
```
oysteinmoseng: Both the PointDragEventObject and this PointDropEventObject should probably have an optional `newPoint` and `newPointId` prop, as these are sent in for convenience if there is only one point being updated.

ts/modules/draggable-points.src.ts
```diff
@@ -0,0 +1,3045 @@
...
+
+/**
+ * Style options for the guide box. The guide box has one state by default, the
+ * `default` state.
+ *
+ * @type         {Highcharts.Dictionary<Highcharts.PlotSeriesDragDropGuideBoxDefaultOptions>}
+ * @since        6.2.0
+ * @optionparent plotOptions.series.dragDrop.guideBox
+ *
+ * @private
```
oysteinmoseng: These options should go in the API - does the `@private` prevent this?

ts/modules/draggable-points.src.ts
```diff
@@ -0,0 +1,3045 @@
...
+};
+
+
+/**
+ * Options for the drag handles.
+ *
+ * @since        6.2.0
+ * @optionparent plotOptions.series.dragDrop.dragHandle
+ *
+ * @private
```
oysteinmoseng: Same as `defaultGuideBoxOptions` - these should go in the public API - does the `@private` prevent this?

bre1470: The `@private` prevents them to go into the global namespace of the class-reference. `@ignore-option` is the equivalent for the API option tree.

ts/modules/draggable-points.src.ts
bre1470: The `@private` prevents them to go into the global namespace of the class-reference. `@ignore-option` is the equivalent for the API option tree.

ts/modules/draggable-points.src.ts
```diff
@@ -0,0 +1,3045 @@
...
+        interface PointDragCallbackFunction {
+            (this: Point, event: PointDragEventObject): void;
+        }
+        interface PointDragDropObject {
+            newValues: Dictionary<number>;
+            point: Point;
+        }
+        interface PointDragEventObject {
+            newPoints: Dictionary<PointDragDropObject>;
+            origin: object;
```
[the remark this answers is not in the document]
bre1470: If you are sure about this, we need to change the doclets as well, because this is taken from the official class reference.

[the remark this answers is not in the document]
oysteinmoseng: Where is this mentioned in the doclets?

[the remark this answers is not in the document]
bre1470: It is not in the documentation since the gulp task split. Doclet sources are in this file behind the internal namespace.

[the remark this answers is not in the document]
oysteinmoseng: I see, yes, it should be changed, `object` is just being lazy. It is technically a `DragDropPositionObject`.

PR #11746, opened 2019-08-22
ts/modules-exporting

### LINKED ISSUE #11575: Migrate `modules/exporting` to TypeScript

bre1470:
- Fixed #11575, Migrated exporting module to TypeScript.
- Improved typing to avoid `arguments as any` and `any` after `isObject(o, true)`.

### DIFF
--- a/ts/parts/Time.ts
+++ b/ts/parts/Time.ts
@@ -516,7 +516,7 @@ Highcharts.Time.prototype = {
         var d, offset, newOffset;
 
         if (this.useUTC) {
-            d = this.Date.UTC.apply(0, arguments as any);
+            d = this.Date.UTC.apply(0, arguments);
             offset = this.getTimezoneOffset(d);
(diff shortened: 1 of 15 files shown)


ts/parts/Time.ts
```diff
@@ -744,7 +744,7 @@ Highcharts.Time.prototype = {
         this: Highcharts.Time,
         f: (string|Array<T>|Highcharts.Dictionary<T>)
     ): Highcharts.Dictionary<T> {
-        if (!isObject(f, true)) { // check for string or array
+        if (!isObject(f)) { // check for string or array
```
TorsteinHonsi: This now changed functionality from checking that f is not an object, to checking that f is neither an object nor an array.

bre1470: Otherwise the purpose of the function to output only dictionaries would not work.

PR #11748, opened 2019-08-22
ts/modules-funnel

### LINKED ISSUE #11577: Migrate `modules/funnel` to TypeScript

bre1470:
Fixed #11577, Migrated FunnelSeries and PyramidSeries.


bre1470:
It looks like pie and funnel are actually not supporting arrays in data labels options. If this is correct I have to change the PR.

Issue #11712, opened 2019-08-19
Dependency Wheel - this.linksTo is undefined after calling node.update in afterAnimate event


jmt406:
#### Expected behaviour
Related nodes get highlighted and the hover popup appears.  No JavaScript error.

We use the afterAnimate event to turn the labels outward like they were when we used D3 Chord diagram.

#### Actual behaviour
Highlighting related nodes and showing the hover popup are broken.  There is a JavaScript error in setNodeState.

Chrome/Edge:
"Cannot read property 'concat' of undefined"
sankey.src.js: 187

Firefox:
"this.linksTo is undefined"
sankey.src.js: 187

#### Live demo with steps to reproduce
http://jsfiddle.net/nsk2rzvf/

Attempt to hover over any part of the dependency wheel.  It does not highlight related nodes or show the hover popup as expected.  Check the browser console for JavaScript errors.

#### Product version
Highcharts 7.1.1 (still affects 7.1.3)

#### Affected browser(s)
Occurs in Windows 10 x64 with Chrome 76.0.3809.100 (x64), Firefox 68.0.2 (x64), and Edge 44.18362.1.0 (EdgeHTML 18.18362).

#### Workaround
On one node in the Dependency Wheel, this.linksTo, this.fromNode, and this.toNode are undefined.  If we update the source to accommodate those situations, the errors do not throw anymore, and everything appears to work as expected.  We aren't experts though.

The workaround requires modifying sankey.src.js (7.1.1):

```
setNodeState: function () {
    var args = arguments,
        others = (this.isNode && this.linksTo) ? this.linksTo.concat(this.linksFrom) :
            [this.fromNode, this.toNode];

    others.forEach(function (linkOrNode) {
        if (linkOrNode) {
            Point.prototype.setState.apply(linkOrNode, args);

            if (!linkOrNode.isNode) {
                if (linkOrNode.fromNode.graphic) {
                    Point.prototype.setState.apply(linkOrNode.fromNode, args);
                }
                if (linkOrNode.toNode.graphic) {
                    Point.prototype.setState.apply(linkOrNode.toNode, args);
                }
            }
        }
    });

    Point.prototype.setState.apply(this, args);
}
```

pawelfus:
Thank your for reporting the issue @jmt406 

First of all, thanks for sharing the workaround!

[...]

Issue #9679, opened 2018-12-17
Chart Sankey reports errors when there is too much data


cheeses:
When there is too much data, it can lead to the accumulation of graphics and error reporting. Is there a good solution? I want to make it scalable by scrolling the mouse, but I don't know how to do it. Looking forward to reply!
http://thyrsi.com/t6/635/1545028630x2728278865.png
http://thyrsi.com/t6/634/1545028310x2890171450.png

TorsteinHonsi:
Thanks for reporting!

Can you demonstrate this live? In my attempt it seems to work: https://jsfiddle.net/highcharts/a3rm16g8/3/

cheeses:
> Thanks for reporting!
> 
> Can you demonstrate this live? In my attempt it seems to work: https://jsfiddle.net/highcharts/a3rm16g8/3/

Hi, @TorsteinHonsi 
Thanks, our data structure is different: https://jsfiddle.net/cheeses/a3rm16g8/5/

KacperMadej:
Here's a smaller demo: https://jsfiddle.net/BlackLabel/zcbm2v83/

Issue #8218, opened 2018-04-20
chart sankey does not support a->b->a pattern


shshen:
#### Expected behaviour
sankey should be able to render a->b->a 

#### Actual behaviour
sankey render incorrectly to  b->a only.

#### Live demo with steps to reproduce
https://jsfiddle.net/zbnu6w36/

#### Product version
highcharts 6.1, but i think this does not work on all versions.

#### Affected browser(s)
all browser types.

pawelfus:
Hi @shshen 

Thank you for reporting. You are right, circular references are not supported right now.

TorsteinHonsi:
I did some work here, making the layout work in some cases. At least it is better now. But now we need to set chart margins to make room for the links, the next improvement would be to make the layout flow automatically.

* https://jsfiddle.net/highcharts/zbnu6w36/8/
* https://jsfiddle.net/highcharts/zbnu6w36/9/
* https://jsfiddle.net/highcharts/zbnu6w36/7/
* https://jsfiddle.net/highcharts/zbnu6w36/10/

hurcy:
@TorsteinHonsi Could you see this case? 
`a->b, b->c, c->b` doesn't work.
https://jsfiddle.net/m5yjrkv1/

TorsteinHonsi:
Fixed, thanks for reporting! https://jsfiddle.net/highcharts/jfeo8tdy/2/

pawelfus:
Hi @niran-manandhar 

Could you share an example? Note that new version (with fix) is not released, only available on our master branch: 

```
<script src="https://github.highcharts.com/highcharts.js"></script>
<script src="https://github.highcharts.com/modules/sankey.js"></script>
```

niran-manandhar:
**with released** 
https://jsfiddle.net/niran_manandhar/q9b05amd/13/ 
has the stack size error
**with master** 
https://jsfiddle.net/niran_manandhar/tuj874Lz/5/ 
doesnt have the stack size error but chart doesn't seem right

TorsteinHonsi:
You're right, there's still a geometry issue even though the stack size issue is resolved. As a workaround we can override the `translate` function, though there's still something wrong with the green link: https://jsfiddle.net/highcharts/jfeo8tdy/9/

niran-manandhar:
managed to do it with margins and clip:false
https://jsfiddle.net/niran_manandhar/vL913mkb/

sruchot:
Hello,
[...]

PR #11752, opened 2019-08-22
bugfix/sankey-multiple-a-points-b-and-b-points-a

cs3b:
Issue occur when i have multiple circular dependencies - this is based on the observation on 20 cases 4 that have larger dataset was not rendering at all I've received:

```
TypeError: fromNode is undefined sankey.src.js:865
    createNodeColumns sankey.src.js:865
    forEach self-hosted:266
    createNodeColumns sankey.src.js:846
    translate sankey.src.js:1102
   ...
```

Sample chart that did render after patch was applied:

case chart didnt render without patch

After applying those small changes, i haven't found any case when it didn't worked.

It might be related with #11712 and #9679 and #8218 - and this only assumption based on reading those tickets, haven't tested those cases. 

I do not now highcharts source code - this the first time i've look inside - still the loop looks broken from general perspective.

### DIFF
--- a/js/modules/sankey.src.js
+++ b/js/modules/sankey.src.js
@@ -464,7 +464,7 @@ seriesType('sankey', 'column',
                 }
                 else {
                     for (i = 0; i < node.linksTo.length; i++) {
-                        point = node.linksTo[0];
+                        point = node.linksTo[i];
                         if (point.fromNode.column > fromColumn) {
                             fromNode = point.fromNode;
                             fromColumn = fromNode.column;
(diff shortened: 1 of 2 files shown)


TorsteinHonsi:
Thanks for your PR! To get us started with the review, can you first provide a demo where the bug is demonstrated? Ideally based on jsFiddle.

cs3b:
I also found another interesting case, when there is no data in the first column (all the nodes goes from middle to right or from right to middle). I've also made patch for that (not included in this pull request).

As i'm expieriencing / fixing that issues on production dataset, it is some work that is needed to reproduce that in jsfiddle. Will do that next week.

TorsteinHonsi:
I'm closing this due to reproduction. Please open an issue if able to reproduce.

PR #11779, opened 2019-08-26
feature/11315-vertical-panning

sebastianbochan:
Added support for `y` and `xy` panning.

----

In case of delete feature/11315-vertical-panning branch please also remove enhancement/11315-vertical-panning branch.

Demo: http://jsfiddle.net/BlackLabel/nvyrpufq/


bre1470:
Not necessary to add me for code review, as I shall not point out code style. The rest is covered by the integrated tests.

ts/parts/Interaction.ts
```diff
@@ -753,12 +757,26 @@ extend(Chart.prototype, /** @lends Chart.prototype */ {
...
+        if (typeof panning === 'object') {
+            panningOptions = panning;
+        } else {
+            panningOptions = {
+                enabled: panning,
+                type: 'x'
+            };
+        }
+
+        (chart.options.chart as any).panning = panningOptions;
```
KacperMadej: I assume that casting to `any` is needed here because the property (second `chart` in `chart.options.chart`) is optional. 

Casting to `any` in new code is depreciated in favor of resolving the problem another way. In this case, you could add an `if` statement that will ensure the existance of `chart.options.chart`.

ts/parts/Interaction.ts
```diff
@@ -815,37 +844,49 @@ extend(Chart.prototype, /** @lends Chart.prototype */ {
...
-                }
-                spill = newMax - paddedMax;
-                if (spill > 0) {
-                    newMax = paddedMax;
-                    newMin -= spill;
-                }
+                // It is not necessary to calculate extremes on ordinal axis,
+                // because the are already calculated, so we don't want to
+                // override them.
+                if (!axis.options.ordinal) {
```
KacperMadej: `axisOpt` could be used since you have `axisOpt = axis.options,` in line No. 805

ts/parts/OrdinalAxis.ts
```diff
@@ -918,9 +918,15 @@ addEvent(Chart, 'pan', function (this: Highcharts.Chart, e: any): void {
         xAxis = chart.xAxis[0],
         overscroll = xAxis.options.overscroll,
         chartX = e.originalEvent.chartX,
+        panning = (chart.options.chart as any).panning,
```
KacperMadej: Same as before. It could be replaced by `chart.options.chart && chart.options.panning`.

ts/parts/OrdinalAxis.ts
```diff
@@ -1018,7 +1024,7 @@ addEvent(Chart, 'pan', function (this: Highcharts.Chart, e: any): void {
     }
 
     // revert to the linear chart.pan version
-    if (runBase) {
+    if (runBase || panning && (panning.type === 'y' || panning.type === 'xy')) {
```
KacperMadej: How about `/y/.test(panning.type)`?

ts/parts/StockChart.ts
```diff
@@ -201,18 +205,20 @@ H.StockChart = H.stockChart = function (
...
         // Always disable startOnTick:true on the main axis when the navigator
         // is enabled (#1090)
         navigatorEnabled = pick(
             options.navigator && options.navigator.enabled,
             (defaultOptions.navigator as any).enabled,
             true
         ),
-        disableStartOnTick = navigatorEnabled ? {
+        verticalPanningEnabled = panning &&
+                    (panning.type === 'y' || panning.type === 'xy'),
```
KacperMadej: Same here - how about `/y/.test(panning.type)`?

ts/parts/StockChart.ts
```diff
@@ -1092,3 +1102,37 @@ addEvent(Chart, 'update', function (
...
 });
+
+// Extend the Axis prototype to calculate start min/max values
+// (including min/maxPadding). This is related to using vertical panning
+// (#11315).
+addEvent(Axis, 'afterSetScale', function (
+    this: Highcharts.Axis
+): void {
+    var axis = this,
+        panning = (axis.chart.options.chart as any).panning,
```
KacperMadej: Another one for avoiding new `any`s.

ts/parts/StockChart.ts
```diff
@@ -1092,3 +1102,37 @@ addEvent(Chart, 'update', function (
...
+        !axis.isXAxis &&
+        !defined(options.startMin) &&
+        !defined(options.startMax)
+    ) {
+
+        var min = Number.MAX_VALUE,
+            max = Number.MIN_VALUE;
+
+        axis.series.forEach(function (series): void {
+            min = Math.min(Math.min.apply(null, (series.yData as any)), min) -
```
KacperMadej: Maybe it's not relevant, but we have `H.arrayMin` to be [safe](https://github.com/highcharts/highcharts/blob/3f9a9b8f4d52283f41dec3585d311acd02a0bd78/ts/parts/Utilities.ts#L2127-L2150).

samples/stock/chart/panning/demo.html
```diff
@@ -0,0 +1,5 @@
+<div id="container" style="height: 400px; min-width: 600px"></div>
+
```
TorsteinHonsi: Add a description to instruct the user to use the `Alt` key for panning. With the current demo, I didn't know that until I read the code.

ts/parts/Options.ts
```diff
@@ -1245,16 +1245,31 @@ H.defaultOptions = {
...
          * @default   {highcharts} false
          * @default   {highstock} true
-         * @since     4.0.3
-         * @product   highcharts highstock gantt
-         * @apioption chart.panning
+         * @apioption chart.panning.enabled
          */
 
+        /**
+         * Decides in what dimensions the user can pan the chart. Cane be
```
TorsteinHonsi: Typo: "Cane be..."

ts/parts/Pointer.ts
```diff
@@ -1112,6 +1112,9 @@ Highcharts.Pointer.prototype = {
             selectionMarker = this.selectionMarker,
             mouseDownX = this.mouseDownX,
             mouseDownY = this.mouseDownY,
+            panningEnabled = isObject(chartOptions.panning) ?
+                chartOptions.panning && chartOptions.panning.enabled :
```
TorsteinHonsi: Since we already check `isObject`, isn't it safe to simplify to this?
```js
            panningEnabled = isObject(chartOptions.panning) ?
                chartOptions.panning.enabled :
                chartOptions.panning,
```

ts/parts/StockChart.ts
```diff
@@ -1092,3 +1101,38 @@ addEvent(Chart, 'update', function (
...
+            max = Number.MIN_VALUE;
+
+        axis.series.forEach(function (series): void {
+            min = Math.min(H.arrayMin(series.yData as any), min) -
+                (axis.min && axis.dataMin ? axis.dataMin - axis.min : 0);
+            max = Math.max(H.arrayMax(series.yData as any), max) +
+                (axis.max && axis.dataMax ? axis.max - axis.dataMax : 0);
+        });
+
+        options.startMin = min;
```
TorsteinHonsi: We shouldn't write internal properties/state back to the `options`. Find another way place for it, either directly on the `Axis`, or, maybe better, create a preliminary `Axis.panning` or `Axis.panningState` object for it. Then the whole object can be destroyed when the panning finishes.

samples/stock/chart/panning/demo.html
Denyllon: Description added as chart's subtitle.

ts/parts/Pointer.ts
Denyllon: I agree. Will remove it.

Denyllon: While tried to simplify it, I've noticed that checking `chart.panning` is done because of TypeScript rules. This property can also be type of `undefined`, so after doing it as you suggested, TS thrown the error, because we're trying to refer something what can be undefined (by `chartOptions.panning.enabled`). In short, TS does not know what `isObject` is actually doing, and shows the error in this case. Maybe @bre1470 would have any idea how to achieve it in other way?

ts/parts/StockChart.ts
Denyllon: After moving those values to separate object located directly on `Axis`, works fine.
Regarding your last sentence, I wondering whether understood it correctly, because I think it would be better to calculate those `start` values once, and keep them on the `Axis.panningState` object, than calculating/removing them on every panning action. Could you try to explain it in more detail?

ts/parts/Pointer.ts
TorsteinHonsi: Okay, we can leave it as is for now.

ts/parts/StockChart.ts
TorsteinHonsi: Well.. I don't have the full details, so I just suggested an architecture. If there are implementation problems with it, then of course that is more important.

TorsteinHonsi:
Pending CI tests.

TorsteinHonsi:
@sebastianbochan In the main demo, it seems that horizontal panning is no longer working, can only pan in vertical direction? https://highcharts.github.io/highcharts-utils/samples/?1574850672884#gh/3bc87384e9/sample/stock/chart/panning

Denyllon:
It seems that was noticeable since v7.1.3, so it's not related with vertical panning feature apparently. I will try to bisect it and find out what causes the issue.

PR #11823, opened 2019-08-30
ts/readme

bre1470:
Added readme for internal TypeScript.

### DIFF
--- a/ts/README.md
+++ b/ts/README.md
@@ -9,10 +9,11 @@ out of the doclets to be in sync with the official API documentation.
 Main Rules
 ----------
 - You have to sync types in code, doclets, and internals yourself
+- Order everything in the internal namespace alphabetical
 - Make use of import and export where possible; consult `parts/Utilities.ts` for
   reference
+- Make use of parantheses around type lists. (`Array<(number|null|string)>`)
 - Do not use `any` type in new code as it is only used for the migration phase
-- Order everything in the internal namespace alphabetical
 - Add paragraphs around conditional types (`(A extends B ? C : D)`), type lists
   (`(A|B)`), and type unions (`(A&B)`).
 
@@ -80,9 +81,9 @@ function a(el?: (Element|undefined)): void {
 ```
 
 
-### Test agains "undefined" and "null"
+### Test against "undefined" and "null"
 
-To prevent yourself from pitfalls during runtime, you should check agains
+To prevent yourself from pitfalls during runtime, you should check against
 optional properties as soon as possible to rule them out in further code.
 
 *Incorrect:*
@@ -134,7 +135,7 @@ of them:
   `function (required: number, optional: string = ''): void;`
 
 * Destructing arrays:
-  `const sentence = ['Hello', 'world']; let [greeting, pronoun] = test;`
+  `const sentence = ['Hello', 'world']; let [greeting, pronoun] = sentence;`
 
 * Destructing objects:
   `const { chart, isDOMElement } = Highcharts; chart('container');`
--- a/ts/README.md
+++ b/ts/README.md
@@ -21,7 +21,7 @@ Main Rules
 Do's and Don'ts
 ---------------
 
-### Avoid `any` casting
+### Avoid "any" casting
 
 Even in strict mode some casts to `any` are allowed. In most situations you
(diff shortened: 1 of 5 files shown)


bre1470:
@KacperMadej Please review the readme and add additional rules that should be included.

KacperMadej:
Are there any newer JS features that could be used in TS code that will later be saved in the corresponding JS file in a safer (more backward compatible) way, e.g. due to targeted ES version? Like what happens when we use `let` and/or `const`? Is it possible and safe to use [default parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters), spread, rest, [destructuring assignments] for array and/or objects? Maybe other newer features too?

Other than this the file looks nice and clear.

bre1470:
Thanks for your input! If I remember correctl, @jon-a-nygaard was able to use default parameters, spread, etc without a problem. When it comes to let and const, we have to try, but it should be safe as well. The target is set to ES5, with only keeping import/export in ES6 syntax. So many thinks get correctly transpiled for older configurations. I will add this to the readme.

jon-a-nygaard:
> If I remember correctl, @jon-a-nygaard was able to use default parameters, spread, etc without a problem.

**Appears safe to use:**
- `const` and `let` is transpiled nicely to `var`
- Destructuring assignments are also transpiled nicely and are safe to use, except for in `import` and `export` statements which is not transpiled.
- Default parameters appear to be transpiled nicely as well.

**May be unsafe to use:**
- Spread and rest operator seems to be handled well, but the transpilation here is more complex so I would be careful of using this for now. At least validate the result before committing.

bre1470:
Thank you for testing the open points out! 👍

bre1470:
@KacperMadej @jon-a-nygaard Please review the additions.

ts/README.md
```diff
@@ -113,40 +113,43 @@ function a(data?: Array<(number|undefined)>, dOffset?: number): number {
...
+  `(a, b) => {console.log('yeah')};`
+
+* Constants and limited variables:
+  `const yes: boolean = true; let degree: number = 180;`
+
+* Default parameter values:
+  `function (required: number, optional: string = ''): void;`
+
+* Destructing arrays:
+  `const sentence = ['Hello', 'world']; let [greeting, pronoun] = test;`
```
jon-a-nygaard: Small typo with `let [greeting, pronoun] = test;`, should be `let [greeting, pronoun] = sentence;`

ts/README.md
```diff
@@ -113,40 +113,43 @@ function a(data?: Array<(number|undefined)>, dOffset?: number): number {
...
+like Internet Explorer 9+, Chrome 5+, Firefox 4+, Safari 5+. So you can make use
+of them:
 
-Good to Know
-------------
+* Arrow functions (without own `this` scope):
+  `(a, b) => {console.log('yeah')};`
+
+* Constants and limited variables:
+  `const yes: boolean = true; let degree: number = 180;`
```
jon-a-nygaard: Is not the type assignment in `const yes: boolean = true;` redundant?

Since the type `boolean` can be inferred from the assignment, and while it is a constant then we can be certain the type does not change as well.
I guess with `let degree: number = 180;` it makes more sense to make sure the type is not changed in an later reassignment of the variable.

bre1470: Constant assignments can be dynamic as well. So the type makes sure, that yes is always `boolean` if the `true` constant is replaced by a function return for example to determine the runtime environment which results in a type list.

jon-a-nygaard: Hmm I see 🤔 but if replaced by a function then the type would be inferred from the return value of the function. But I guess it has its value when you really want to enforce the type.

bre1470: An example: `const header = document.getElementById('header');` - the header const becomes type `Element|null`;

ts/README.md
```diff
@@ -80,7 +80,7 @@ function a(el?: (Element|undefined)): void {
 ```
 
 
-### Test agains `undefined` and `null`
+### Test agains "undefined" and "null"
```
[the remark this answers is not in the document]
bre1470: Thanks, fixed.

ts/README.md
bre1470: Thanks, fixed.

PR #11831, opened 2019-09-02
ts/indicators-indicators

### LINKED ISSUE #11817: Migrate `indicators/indicators` to TypeScript

bre1470:
Fixed #11817, Migrated SmaIndicator to TypeScript.

### DIFF
--- a/ts/indicators/indicators.src.ts
+++ b/ts/indicators/indicators.src.ts
@@ -16,6 +16,11 @@ import H from '../parts/Globals.js';
  */
 declare global {
     namespace Highcharts {
+        interface IndicatorValuesObject {
+            values: Array<Array<number>>;
+            xData: Array<number>;
+            yData: Array<number>;
+        }
         interface LineSeriesOptions {
             useOhlcData?: boolean;
         }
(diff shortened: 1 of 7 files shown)


ts/indicators/indicators.src.ts
```diff
@@ -42,8 +42,8 @@ declare global {
             public getName(): string;
             public getValues(
                 series: Series,
-                params: object
-            ): IndicatorValuesObject;
+                params: SmaIndicatorParamsOptions
+            ): (boolean|IndicatorValuesObject);
```
sebastianbochan: @bre1470 
What is the reason of setting here a boolean type? Because in all cases it returns array of points (empty array)

bre1470: Inherited indicators might return boolean and you can not change a return type between related classes.

sebastianbochan: I do not know any scenario when getValues can be boolean, could you elaborate how you are able reproduce that?

bre1470: Basically when there are less values available on the referenced series than the period param requires: https://github.com/highcharts/highcharts/pull/11835/files#diff-bb3b226e917ae389d52c66c3939658ddR186-R188

PR #11832, opened 2019-09-02
bugfix/11812-color-axis-chart-update

### LINKED ISSUE #11812: Chart.update doesn't update colorAxis if the initial config didn't contain it
#### Expected behaviour

```js
 var chart = Highcharts.chart('container', {
        chart: {
            type: 'heatmap',
        },
				title: {
        		text: 'Resize window to fail'
        },
        //colorAxis: { },
        series: [{
            data: [[0, 0, 10], [0, 1, 19], [0, 2, 8], [0, 3, 24], [0, 4, 67], [1, 0, 92], [1, 1, 58], [1, 2, 78], [1, 3, 117], [1, 4, 48]]
        }]
    });
    chart.update({ 
    	colorAxis: {
    		reversed: true
      }
    });
```

`Chart#update` should update `colorAxis` regardless of the initial config.

In general I would expect 
```
Highcharts.chart('container', config)
``` 
and 
```
let chart = Highcharts.chart('container', {})
chart.update(config, true, true)
``` 
to be identical.
#### Actual behaviour
`colorAxis` is not updated if the initial config doesn't contain `colorAxis`.

http://jsfiddle.net/7u9g6h30/1/
#### Product version

7.1.3

ppotaczek:
Fixed #11812, point colors were not changed if `colorAxis` was added in update.

### DIFF
--- a/ts/parts/Dynamics.ts
+++ b/ts/parts/Dynamics.ts
@@ -376,7 +376,9 @@ extend(Chart.prototype, /** @lends Highcharts.Chart.prototype */ {
         (chartOptions as any)[type].push(userOptions);
 
         if (isColorAxis) {
-            this.isDirtyLegend = true;
+            this.series.forEach(function (series: Highcharts.Series): void {
+                series.update({}, false);
+            });
         }
 
         if (pick(redraw, true)) {
@@ -565,8 +567,8 @@ extend(Chart.prototype, /** @lends Highcharts.Chart.prototype */ {
         'xAxis',
         'yAxis',
         'zAxis',
-        'series',
         'colorAxis',
+        'series',
         'pane'
     ],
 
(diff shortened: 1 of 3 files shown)


ts/parts/Dynamics.ts
```diff
@@ -376,7 +376,9 @@ extend(Chart.prototype, /** @lends Highcharts.Chart.prototype */ {
         (chartOptions as any)[type].push(userOptions);
 
         if (isColorAxis) {
-            this.isDirtyLegend = true;
+            this.series.forEach(function (series: Highcharts.Series): void {
+                series.update({}, false);
```
TorsteinHonsi: Is it necessary to call `series.update()` on every series in the chart in order to make it bind to the color axis? Can we do this in a cheaper way?

ppotaczek: Yes, it looks like that we can use only `bindAxes` method:

        if (isColorAxis) {
            this.isDirtyLegend = true;
            this.series.forEach(function (series: Highcharts.Series): void {
                series.bindAxes();
            });
        }

PR #11833, opened 2019-09-02
Bugfix/11690 undefined global

### LINKED ISSUE #11690: Regression: win is not defined in Globals.js
#### Expected behaviour
Should load Highcharts as an ESM module and create a chart without errors.

#### Actual behaviour
Highcharts errors due to trying to access `win`, while it should access `glob`.
Regression since https://github.com/highcharts/highcharts/pull/11214

https://jsfiddle.net/jon_a_nygaard/6emvkysu/

#### Product version
Highcharts v7.1.3

jon-a-nygaard:
Fixed #11690, removed use of undefined global variable that broke support for ES modules.

---
# Description
Replaced use of `win` with `glob` to solve undefined error, and updated relative dependency paths to fix ESM support.

Added a sample `samples/highcharts/esm/simple` to test that ESM support is not broken in future versions.

# Related issue(s)
- Closes #11690

### DIFF
--- a/js/modules/networkgraph/networkgraph.src.js
+++ b/js/modules/networkgraph/networkgraph.src.js
@@ -95,8 +95,8 @@ import U from '../../parts/Utilities.js';
 var defined = U.defined;
 import '../../parts/Options.js';
 import '../../mixins/nodes.js';
-import '/layouts.js';
-import '/draggable-nodes.js';
+import './layouts.js';
+import './draggable-nodes.js';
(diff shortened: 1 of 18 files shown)


jon-a-nygaard:
Yes, that will probably be a cleaner solution as we could simplify and remove `win` and `glob` and just fetch what we need directly from `window`.

But this will require a change in the `highcharts-assembler` before modifying the `highcharts` code.
Should I make this adaptation in this PR @TorsteinHonsi?

jon-a-nygaard:
Also worth noting that the PR https://github.com/highcharts/highcharts-utils/pull/25 is required to be able to test the `samples/highcharts/esm/simple`.

PR #11883, opened 2019-09-03
docs/docusaurus-setup

goransle:
Cleaned up documentation for use with the doc-builder.


goransle:
> I noticed the PNG's are stored in the folder next to md files - wouldn't it be better to set up a separate file structure for images?

It is probably down to preference, but I would imagine that when editing the documentation it would be easier to keep them together. It also works on Github without changing to relative urls.

TorsteinHonsi:
@cvasseng Opinions?

cvasseng:
@goransle does the image handler in Docusaurus support relative paths? If so, I guess there's no issue with having a separate folder with images. 

It might be easier to find things though if they're not in a separate image folder due to the file hierarchy limiting the number of them in any given folder, but we should probably formulate some semantics in terms of naming them regardless so they're easy to identify. 

I'm actually a bit torn, as neither keeping them in the same folder as the markdown nor having a root `images` folder seem ideal. The former adds a lot of noise when browsing the docs here, and the latter requires variable amounts of path backtracking to get to it from the md files (since the docs themselves are in a nested folder hierarchy).

There's another point too though that we should start considering once we have the base up and running, which is that we should replace chart images with embeds. Right now, some of the screens are from Highcharts 4.x, plus it's a bit silly to have rasterized versions instead of actually demoing the interactivity in the same go. So maybe the answer here is "depends on how many images we end up with if we remove all the chart images"?

TorsteinHonsi:
> that we should replace chart images with embeds

Definately. 

> I'm actually a bit torn, as neither keeping them in the same folder as the markdown nor having a root images folder seem ideal.

I would prefer keeping the images separate, but I'll merge this then we can consider it later if we don't like working with it.

PR #11884, opened 2019-09-03
bugfix/11405-plotlines-bottom-chart-edge

### LINKED ISSUE #11405: The plotline on y axis is not visible when value is given 0 to 0.99
#### Expected behaviour
The plot line on y - axis is rendered outside the view port when the series data is positive [0, 0.3, 0.5, 1, 2] and the plotline value is 0 to 0.99.

#### Actual behaviour
Please refer the negative data [0, -0.3, -0.5, -1, -2], the plot line is visible for value 0.

https://jsfiddle.net/tushar_chandra_030389/ov71mzrg/6/

#### Product version

#### Affected browser(s)
Chrome, rest as well.

image

ppotaczek:
Fixed #11405, plotLines were rendered outside the bottom and left chart edge if the value was the same as the axis extreme.

### DIFF
--- a/js/parts/Axis.js
+++ b/js/parts/Axis.js
@@ -3484,8 +3484,9 @@ H.extend(Axis.prototype, /** @lends Highcharts.Axis.prototype */ {
             }
             return x;
         };
-        // #11405
-        if (!this.horiz && value === this.min) {
+        // Move the line inside the chart (#11405)
+        if ((!this.reversed && value === this.min) ||
+            (this.reversed && value === this.max)) {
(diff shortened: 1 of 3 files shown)


TorsteinHonsi:
This change lead to a lot of diffs in http://utils.highcharts.local/samples/#nightly. The problem is that the bottom grid line, present in almost every chart, now is moved one  pixel up and placed next to the axis line. It should be covered by the axis line like before.

The OP demo had no axis line, so maybe we need to check for the axis line width in addition.

PR #11887, opened 2019-09-04
Tests: POC tests fail when run in random order.

activist:
Just a small proof-of-concept that we have tests that fails when being run in "random" order (other than default). 

This indicates that some tests are not resetting some global parameters or are leaking data into the global scope.


activist:
@TorsteinHonsi Not exactly high priority, but probably worth looking into this at some point.

PR #11888, opened 2019-09-04
Tools: Testing to increase Win.Chrome+ Mac.Firefox browsercount

activist:
Testing to split tests across two instances when running Win.Chrome and Mac.Firefox in order to migate the disconnect issue.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -338,6 +338,7 @@ workflows:
           browsers: "Mac.Firefox"
       - test_browsers:
           name: "test-Win.Chrome"
+          browsercount: 2
           requires:
             - "test-Mac.Firefox"
           browsers: "Win.Chrome"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -333,6 +333,7 @@ workflows:
           browsers: "Mac.Safari"
       - test_browsers:
           name: "test-Mac.Firefox"
+          browsercount: 2
           requires:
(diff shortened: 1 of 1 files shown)


activist:
So far the tests haven't failed. As for the implications on this we will just have to see how it goes. Given a test that is split across several(2) browser instances it works as follows:
1. First part is queued in Browserstack and waits until there is an availiable slot in BrowserStack.
2. When first part finishes the next part is queued in BrowserStack and have to wait for an available slot.

Basically this means that Win.Chrome and Mac.Firefox tests will have to wait for an available slot in Browserstack twice. So depending on the how many other builds are running in CI there might be a lot of waiting time, but still we would save a lot of time if the disconnects are now mitigated.

activist:
> I wonder if we should make this for all browsers so that in case of test errors it will stop sooner and free up the BS slots.
If I remember correctly the karma-sharding works so that both parts will always run even if there is an failure in one of them. I guess that is expected as we are asking it to run a full test suite. 
There is another option to let CircleCI split the tests into multiple containers, but in that case I also think per default that all tests will run. 

As a sidenote, if we are going to let the CI do the splitting we will also run into the same problems as described here: https://github.com/highcharts/highcharts/pull/11887

bre1470:
Okay, then maybe wait and see how helpful the current change will be. I thought, that the problem was covered by Torstein's PR https://github.com/highcharts/highcharts/pull/11889.

PR #11902, opened 2019-09-05
ts/internal-states

### LINKED ISSUE #11886: Simplify `series.states` options as a Dictionary<SeriesOptions> internally
Technically every point state can have the same set of point options, even though it is not mentioned in the API documentation (because only visual ones makes sense).

The types in doclets will be not changed for now.

bre1470:
Close #11886, Improved internal series factory and states.

Further on you can use `seriesType<Highcharts.MySeries>(…)` and it will automatically reflect the options, series class, and point class. This also means, that you do not have to define the `this` scope of the class functions anymore.

The reflection paths are `MySeries.options` and `MySeries.pointClass` (= `typeof MyPoint`).

Additionally internal `MySeriesOptions.state` has been simplified. Now the generic type `SeriesStatesOptionsObject<MySeries>` is available.


bre1470:
Unfortunately there is an exception to what I wrote. The `merge` function is masking the `this` scope. So in case of mixin-merges you still have to write the `this` scope in the class functions.

Issue #11329, opened 2019-07-05
Transform issue with tooltip.outside


kzoon:
The hover on line charts is off when you use css **transform** on a parent container.

On http://jsfiddle.net/kzoon/jswhn2v8/3/ hover activates data points to the left of the actual mouse pointer. The more to the right, the worse it gets.
The column series does NOT have this issue ?!?!

It looks like the fix of https://github.com/highcharts/highcharts/issues/10498 doesn't work in all cases.
In the meantime, my own workaround for tooltip positions (see https://github.com/highcharts/highcharts/issues/2405) doesn't help here anymore.

For me this issue is quite urgent, because the problem happens in a version already released to my customers.

pawelfus:
Hi @kzoon 

Let me answer your questions:
> The column series does NOT have this issue ?!?!

Columns show tooltip when hover happens directly over a rectangle. For other series types (like area, line, spline etc.) we use k-d tree for searching the closest point to the pointer. Values used for searching and rendering are the same, but `transform: scale()` changes how rendered chart looks.

> It looks like the fix of #10498 doesn't work in all cases.

Issue #10498 refers to the width of the chart, not related to the tooltip's position.



Is that issue any different from #2405? At this moment it seems to be a duplicate which I should close..

Workaround posted originally by @TorsteinHonsi seems to be fine (simplified a little): http://jsfiddle.net/BlackLabel/6jng50dy/

kzoon:
The workaround works in the fiddle I supplied. My problem now is that it doesn't in the real world. So, the fiddle was not representative of the real problem. I will try to make a fiddle that shows my problem.

kzoon:
Found it. You added scale: 0.7 to the chart JSON in order to make the workaround work. I can't do that. The transform happens in a different application layer from the one that adds the chart, so I can't just put the scale into the chart config.

So I'm looking for a workaround that doesn't require the chart to know the transform scale.

I had such a workaround based on getBoundingClientRect and offsetWidth of the chart container (see my code in https://github.com/highcharts/highcharts/issues/2405) but that one doesn't work anymore.

kzoon:
Found the real problem:
 ```
 tooltip: {
    outside: true
  },
```
makes the tooltip to go all over the place when using transform scale.
See   http://jsfiddle.net/kzoon/0arw8t7j/6/

pawelfus:
Thanks for the details! 

We have a few problems in the demo:
- different scales (for tooltip and chart)
- wrong position
[...]

PR #11927, opened 2019-09-10
Bugfix/9590 split tooltip outside

### LINKED ISSUE #9590: Highstock tooltip.outside=true fails for tooltip.split=true
#### Expected behaviour
Tooltip (g.highcharts-tooltip ) should be rendered outside of the svg.highcharts-root when tooltip.outside = true and tooltip.split = true.

#### Actual behaviour
Tooltip is rendered within svg.highcharts-root, resulting in a tooltip that is hidden behind the edge of the highcharts container.

http://jsfiddle.net/qa7hvb2o/

#### Product version
http://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/tooltip/outside

#### Affected browser(s)
Chrome
Safari

jon-a-nygaard:
Fixed #9590, added support for split tooltip with `tooltip.outside=true`

---
# Description
Add support for using split tooltip with outside container.

There is potential to simplify and use the `updateTooltipContainer` function on "non-split" tooltips, but I ran into issues with calculating label positions and width across different types.

**NB!:** The solution does not include support for the edge cases of transform and translate, similar to what is described in #11329.

# Example(s)
- [Demo of issue](http://jsfiddle.net/qa7hvb2o/) - hover one of the charts to see the tooltip is inside
- [Demo of issue with fix applied](http://jsfiddle.net/jon_a_nygaard/8w6km0Lu/) - hover one of the charts to see the tooltip is outside

# Related issue(s)
- Closes #9590

### DIFF
--- a/ts/parts/Tooltip.ts
+++ b/ts/parts/Tooltip.ts
@@ -257,6 +257,40 @@ var doc = H.doc,
     syncTimeout = H.syncTimeout,
     timeUnits = H.timeUnits;
 
+/**
+ * Used to update the tooltip container when split tooltip has outside=true.
+ *
+ * @private
+ * @todo Reuse functionality on other tooltip types. Issue with finding a common
+ * solution to calculate label position and size.
+ * @todo Export the method to make it available for unit testing
+ * @param {HTMLElement} chartContainer The chart container to align tooltip to.
+ * @param {HTMLElement} container The container for the outside tooltip.
+ * @param {Highcharts.SVGRenderer} renderer The renderer for the outside
+ * tooltip.
+ * @param {Highcharts.SVGElement} label The tooltip label
+ * @return {void}
+ */
+function updateTooltipContainer(
+    chartContainer: HTMLElement,
+    container: HTMLElement,
+    renderer: Highcharts.SVGRenderer,
+    label: Highcharts.SVGElement
+): void {
+    // Position the tooltip container to the chart container
+    const offset = H.offset(chartContainer);
+    container.style.left = offset.left + 'px';
+    container.style.top = offset.top + 'px';
+
+    // Set container size to fit the tooltip
+    const { width, height, x, y } = label.getBBox();
+    renderer.setSize(
+        width + x,
+        height + y,
+        false
+    );
+}
+
 /* eslint-disable no-invalid-this, valid-jsdoc */
 
 /**
@@ -377,12 +411,9 @@ H.Tooltip.prototype = {
          * Split tooltip does not support outside in the first iteration. Should
          * not be too complicated to implement.
          */
-        this.outside = (
-            pick(
-                options.outside,
-                Boolean(chart.scrollablePixelsX || chart.scrollablePixelsY)
-            ) &&
-            !this.split
+        this.outside = pick(
+            options.outside,
+            Boolean(chart.scrollablePixelsX || chart.scrollablePixelsY)
         );
     },
 
@@ -562,7 +593,9 @@ H.Tooltip.prototype = {
                 this.label.addClass('highcharts-tooltip-' + this.chart.index);
             }
 
-            if (this.outside) {
+            // Split tooltip use updateTooltipContainer to position the tooltip
+            // container.
+            if (tooltip.outside && !tooltip.split) {
                 set = {
                     x: this.label.xSetter as any,
                     y: this.label.ySetter as any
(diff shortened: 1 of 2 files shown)


ts/parts/Tooltip.ts
```diff
@@ -1169,7 +1192,13 @@ H.Tooltip.prototype = {
                 } as any);
             }
 
-            this.isHidden = false;
+            // show it
+            if (tooltip.isHidden && tooltip.label) {
```
TorsteinHonsi: What happens if `tooltip.label` is undefined? The idea with the old structure, calling `tooltip.getLabel()` initially, is that the label is generated the first time `refresh` is called.

ts/parts/Tooltip.ts
```diff
@@ -257,6 +257,40 @@ var doc = H.doc,
     syncTimeout = H.syncTimeout,
     timeUnits = H.timeUnits;
 
+/**
```
TorsteinHonsi: This file doesn't use private functions for anything else, I prefer keeping it consistent with the existing code, which means adding a class function. Then the arguments here are not needed, as they are part of the `Tooltip` context.

ts/parts/Tooltip.ts
```diff
@@ -1385,6 +1414,25 @@ H.Tooltip.prototype = {
...
+            renderer
+        } = tooltip;
+        if (outside && container && renderer) {
+            updateTooltipContainer(
+                chart.container,
+                container,
+                renderer,
+                tooltipLabel
+            );
+        }
```
TorsteinHonsi: Seems a little bloaty to pass all the arguments to a private function when we can just a) do it inline or b) use a class function.

ts/parts/Tooltip.ts
```diff
@@ -257,6 +257,40 @@ var doc = H.doc,
...
+    label: Highcharts.SVGElement
+): void {
+    // Position the tooltip container to the chart container
+    const offset = H.offset(chartContainer);
+    container.style.left = offset.left + 'px';
+    container.style.top = offset.top + 'px';
+
+    // Set container size to fit the tooltip
+    const { width, height, x, y } = label.getBBox();
+    renderer.setSize(
```
TorsteinHonsi: We already do `renderer.setSize` in the `updatePositions` function. Is it not possible to combine?

ts/parts/Tooltip.ts
jon-a-nygaard: `tooltip.label` should at this point always be defined, because `getLabel()` is called both in `renderSplit` and in the `else` section. The checking that it exists before use is only to satisfy TypeScript.
Previously it was called twice on split tooltip, which is the reason for this change.

ts/parts/Tooltip.ts
jon-a-nygaard: Preferably yes, but it is complicated as `updatePositions` is currently not used for split tooltip.

ts/parts/Tooltip.ts
jon-a-nygaard: I see your point here and will adapt this if you believe it is for the better, but in my opinion there is the following downsides with this:

a) do it inline `updateTooltipContainer(tooltip);`
- There is a loss in clarity when doing this. It is more difficult to know what is required for the function to execute when it is asking for a complete `Tooltip` object. Having the required variables as parameters ensures they are passed to the function.
- It is easier to unit-test (if possible) when the function is asking for "only" [`HTMLElement`, `HTMLElement`, `Highcharts.SVGRenderer`, `Highcharts.SVGElement`], which is although more parameters, it is less complicated structures then a complete `Highcharts.Tooltip` object.
- The usability for the function is less flexible as it now requires `Tooltip`. This may seem redundant now, but you never know in the future.

b) use a class function `tooltip.updateTooltipContainer();`
In addition to what I believe is downsides in a) this would also in my opinion add the following
- It bloats the Tooltip prototype as `updateTooltip` is only a utility function and not really meant for the public
- Using context adds state to the function, which again is more difficult to test.

Choosing between these two I will prefer a)

ts/parts/Tooltip.ts
jon-a-nygaard: I have made some relevant comments to this down below in one of your other comments.

ts/parts/Tooltip.ts
TorsteinHonsi: I hear you, but I think it's more important that we are consistent. Many functions on the Tooltip property are private by nature. The downside is that we're bloating the prototype, while the upside is that we can hack and fix the Tooltip from the outside.

By doing alternative a) I meant not creating a function at all. Since we're only doing the operation once, there's no real need for calling a function. What we will do in the future isn't relevant now. If we decide we want to reuse this block of code, we can always create a function later.

ts/parts/Tooltip.ts
TorsteinHonsi: Okay I didn't notice that.

jon-a-nygaard:
@TorsteinHonsi Please see latest changes

ts/parts/Tooltip.ts
```diff
@@ -1385,6 +1380,31 @@ H.Tooltip.prototype = {
...
+         * the operation.
+         */
+        const {
+            container,
+            outside,
+            renderer
+        } = tooltip;
+        if (outside && container && renderer) {
+            // Position the tooltip container to the chart container
+            const offset = H.offset(chart.container);
```
TorsteinHonsi: Change this to use the existing `chart.pointer.chartPosition`, or get it if undefined. Avoids traversion the DOM unnecessarily. Assigned to @TorsteinHonsi .

PR #11934, opened 2019-09-11
Tools: Reduced number of cpus by one

activist:
.. so that at least one physical core is available for other stuff. This prevents karma from disconnecting in some cases.


test/karma-conf.js
```diff
@@ -240,7 +240,7 @@ module.exports = function (config) {
         browsers = Object.keys(browserStackBrowsers);
     }
 
-    const browserCount = argv.browsercount || (os.cpus().length - 1);
+    const browserCount = argv.browsercount || (os.cpus().length - 2);
```
TorsteinHonsi: What if the os has only CPUs available? Shouldn't it be something like this?
`const browserCount = argv.browsercount || Math.max(1, os.cpus().length - 2);`

activist: You are right. I corrected it to the above.

PR #11951, opened 2019-09-12
docs/boost-array-data

### LINKED ISSUE #11608: Tooltips not showing on scatter plot running with boost module

pawelfus:
Improved docs and errors for boost and turboThreshold. Close #11608.

Follow up: #11704 

- [x] improve description for error 12
- [x] improve docs for turbo threshold
- [ ] improve description in [docs](https://www.highcharts.com/docs/advanced-chart-features/boost-module) - I think `caveats` is the right place. Could you update this @TorsteinHonsi ?

Implementation explanation:
My first attempt was to simply let Highcharts throw error by modifying `series.options.turboThreshold` temporary, like this:
```js
var threshold = series.options.turboThreshold;

if (isSeriesBoosted) {
  series.options.turboThreshold = 0.1;
  proceed ( ... );
}

series.options.turboThreshold = threshold;
```

But unfortunately whole logic for checking if `isSeriesBoosted` is based on initialized series, while error 12 is thrown during series initialization. This is not a problem to check if `series.boostThreshold` is exceeded, but to check if the chart should be in boosted mode with shared context for all series (e.g.`boost.allowForce`).


ts/parts/Series.ts
```diff
@@ -4177,6 +4179,32 @@ H.Series = H.seriesType<Highcharts.LineSeries>(
...
+         * Find and return the first non null point in the data
+         *
+         * @private
+         * @function Highcharts.Series.getFirstNotNullPoint
+         * @param {Array<Highcharts.PointOptionsType>} data
+         *        Array of options for points
+         *
+         * @return {Highcharts.PointOptionsType}
+         */
+        getFirstNotNullPoint: function (
```
TorsteinHonsi: Nice refactoring, but it would be easier on the brain if the function is called `getFirstValidPoint`.

TorsteinHonsi:
> improve description in docs - I think caveats is the right place. Could you updated this @TorsteinHonsi ?

Sure, but I'll wait until we have got the docs into markdown.

ts/parts/Series.ts
pawelfus: I agree, this was the first name I used :) I changed this to be more accurate - we have `point.isValid()` that really validates a point, here we just check if the point is null. If you think it's not a problem, I will change to `getFirstValidPoint`.

TorsteinHonsi: Yes, `isValid` and `isNull` are basicly antonyms.

TorsteinHonsi:
@pawelfus Will you also add the appropriate language to https://github.com/highcharts/highcharts/blob/master/docs%2Fadvanced-chart-features%2Fboost-module.md?

pawelfus:
Done. Refactored, added descriptions and additionally fixed minor typos in the `boost-module.md` file.

PR #11963, opened 2019-09-13
ts/indicators-atr

### LINKED ISSUE #11840: Migrate `indicators/atr` to TypeScript

activist:
Closes #11840 , Migrated AtrIndicator to TypeScript.

### DIFF
--- a/ts/indicators/atr.src.ts
+++ b/ts/indicators/atr.src.ts
@@ -1,0 +1,215 @@
+/* *
+ *
+ *  License: www.highcharts.com/license
+ *
+ *  !!!!!!! SOURCE GETS TRANSPILED BY TYPESCRIPT. EDIT TS FILE ONLY. !!!!!!!
+ *
+ * */
+
+'use strict';
+
+import H from '../parts/Globals.js';
+
+declare global {
+    namespace Highcharts {
+        class AtrIndicator extends SmaIndicator {
+            public data: Array<AtrIndicatorPoint>
+            public yData: Array<Array<number>>;
+            public getValues(
+                series: Series,
+                params: AtrInidicatorParamsOptions
+            ): (boolean|IndicatorValuesObject);
+        }
+
+        interface AtrIndicatorOptions extends SmaIndicatorOptions {
+            params?: AtrInidicatorParamsOptions;
+        }
+
+        interface AtrInidicatorParamsOptions {
+            period?: number;
+        }
+
+        class AtrIndicatorPoint extends SmaIndicatorPoint {
+            public series: AtrIndicator
+        }
+
+        interface SeriesTypesDictionary {
+            atr: typeof AtrIndicator;
+        }
+    }
+}
+
+
+import U from '../parts/Utilities.js';
+var isArray = U.isArray;
+
+var seriesType = H.seriesType,
+    UNDEFINED: undefined;
+
+/* eslint-disable valid-jsdoc */
+// Utils:
+/**
+ * @private
+ */
+function accumulateAverage(
+    points: Array<Array<(number|Array<number>)>>,
+    xVal: Array<number>,
+    yVal: Array<Array<number>>,
+    i: number
+): void {
+    var xValue = xVal[i],
+        yValue = yVal[i];
+
+    points.push([xValue, yValue]);
+}
+
+/**
+ * @private
+ */
+function getTR(currentPoint: Array<number>, prevPoint: Array<number>): number {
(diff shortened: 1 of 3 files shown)


ts/indicators/atr.src.ts
```diff
@@ -0,0 +1,215 @@
...
+ *
+ *  !!!!!!! SOURCE GETS TRANSPILED BY TYPESCRIPT. EDIT TS FILE ONLY. !!!!!!!
+ *
+ * */
+
+'use strict';
+
+import H from '../parts/Globals.js';
+
+declare global {
```
bre1470: Missing doclet
```ts
/**
 * Internal types
 * @private
 */
```

ts/indicators/atr.src.ts
```diff
@@ -0,0 +1,215 @@
...
+var seriesType = H.seriesType,
+    UNDEFINED: undefined;
+
+/* eslint-disable valid-jsdoc */
+// Utils:
+/**
+ * @private
+ */
+function accumulateAverage(
+    points: Array<Array<(number|Array<number>)>>,
```
bre1470: Maybe you can use the type `Array<[number|Array<number>]>` to narrow it down.

ts/indicators/atr.src.ts
```diff
@@ -0,0 +1,215 @@
...
+                xValue = (xVal as any)[0],
+                yValue = yVal[0],
+                range = 1,
+                prevATR = 0,
+                TR = 0,
+                ATR: Array<Array<number>> = [],
+                xData = [],
+                yData = [],
+                point,
+                i,
```
bre1470: Types missing.

ts/indicators/atr.src.ts
```diff
@@ -0,0 +1,215 @@
...
+ * is inherited from [chart.type](#chart.type).
+ *
+ * @extends   series,plotOptions.atr
+ * @since     6.0.0
+ * @product   highstock
+ * @excluding dataParser, dataURL
+ * @apioption series.atr
+ */
+
+''; // to include the above in jsdocs
```
bre1470: Should be something like `to include the above in the js output`

ts/indicators/atr.src.ts
```diff
@@ -0,0 +1,215 @@
...
+        TR = Math.max(HL, HCp, LCp);
+
+    return TR;
+}
+
+/**
+ * @private
+ */
+function populateAverage(
+    points: any,
```
bre1470: This should be `Array<[number|Array<number>]>` too.

ts/indicators/atr.src.ts
```diff
@@ -0,0 +1,215 @@
...
+                yValue = yVal[0],
+                range = 1,
+                prevATR = 0,
+                TR = 0,
+                ATR: Array<Array<number>> = [],
+                xData = [],
+                yData = [],
+                point,
+                i,
+                points: Array<Array<(number|Array<number>)>>;
```
bre1470: This could be `Array<[number|Array<number>]>` as well.

ts/indicators/atr.src.ts
activist: Having a hard time to get this to work
```
Argument of type '[number, number[]]' is not assignable to parameter of type '[number | number[]]'.
  Types of property 'length' are incompatible.
    Type '2' is not assignable to type '1'.ts(2345)
``
Any clues?

ts/indicators/atr.src.ts
activist: Yes, it isn't used in the function so I tried the simple fix :)

ts/indicators/atr.src.ts
bre1470: You have to use the tuple everywhere in the module. You are mixing `[number , Array<number>]` (length = 2) with `Array<(number|Array<number>)>` (length = 1).

activist: > Maybe you can use the type Array<[number|Array<number>]> to narrow it down.
Yes, I didn't get the tuple of number or Array<number> thing, but Array<[number,Array<number>]> makes sense :)

activist: > Maybe you can use the type Array<[number|Array<number>]> to narrow it down.

Yes, I didn't get the whole tuple of number or Array<number> thing, but `[number, Array<number>] makes sense :)

activist:
@bre1470, please review revised code with fixes. Slowly getting there.. :)

PR #11964, opened 2019-09-13
bugfix/11476-inverted-variwide-labels-error

### LINKED ISSUE #11476: Inverted Variwide with hidden X-axis labels - Cannot set property 'labelrank' of null
#### Expected behaviour

Inverted Variwide with hidden X-axis labels renders chart with out any error.

#### Actual behaviour
Inverted Variwide with hidden X-axis labels results in,
*Uncaught TypeError: Cannot set property 'labelrank' of null*

Issue URL: https://jsfiddle.net/rajeshdanabal/hm38gw1x/

Go to URL and in the console you can see the error.

#### Product version
Highcharts v7.1.2 (2019-06-03)

#### Affected browser(s)
Version 75.0.3770.142 (Official Build) (64-bit)

ppotaczek:
Fixed #11476, inverted variwide chart with disabled X axis labels threw an error.

### DIFF
--- a/js/modules/variwide.src.js
+++ b/js/modules/variwide.src.js
@@ -264,14 +264,18 @@ addEvent(H.Axis, 'afterDrawCrosshair', function (e) {
 addEvent(H.Axis, 'afterRender', function () {
     var axis = this;
 
-    if (!this.horiz && this.variwide && this.options.labels.enabled) {
+    if (!this.horiz && this.variwide) {
         this.chart.labelCollectors.push(function () {
-            return axis.tickPositions.map(function (pos, i) {
-                var label = axis.ticks[pos].label;
+            return axis.tickPositions
+                .filter(function (pos) {
+                    return axis.ticks[pos].label;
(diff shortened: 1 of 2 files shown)


js/modules/variwide.src.js
```diff
@@ -264,7 +264,7 @@ addEvent(H.Axis, 'afterDrawCrosshair', function (e) {
 addEvent(H.Axis, 'afterRender', function () {
     var axis = this;
 
-    if (!this.horiz && this.variwide) {
+    if (!this.horiz && this.variwide && this.options.labels.enabled) {
```
TorsteinHonsi: Instead of checking the `enabled` option, I think it would be safer to just filter the `tickPositions` by the ones that have a label, before we do the `map`. We do something similar in RadialAxis: https://github.com/highcharts/highcharts/blob/v7.2.0/ts%2Fparts-more%2FRadialAxis.ts/#L700-L704 .

PR #11965, opened 2019-09-13
bugfix/11100-3d-column-animation

### LINKED ISSUE #11100: 3D column is not animating on data update
#### Expected behaviour
Smooth animation on points' update.

#### Actual behaviour
No animation.

https://jsfiddle.net/BlackLabel/goyb9fpc/

Bisected to commit af24cb21a1 (in 3D, nodeName is `g`, while `shapeType` is cuboid).

#### Product version
v7.1.0+

pawelfus:
Fixed #11100, 3D columns were not animated on data update.
___
Two minor things to mention:

1.
There's one small problem with this solution: `hasNewShapeType` is added to `Point` class, not to `ColumnPoint`. This is caused by behaviour of `seriesType(..args, /* pointClass */)` - which does not use parent's series `pointClass`, but original `Point` class when adding some new props to new `pointClass`. For example (demo: http://jsfiddle.net/BlackLabel/tp82b4rh/)

```js
Highcharts.seriesType(
  'newColumnOK', 'column', {}, {},
  { something: true }
);

Highcharts.seriesType(
  'newColumnBAD', 'newColumnOK', {}, {},
  { }
);

console.log(
  typeof Highcharts
    .seriesTypes.newColumnOK.prototype
    .pointClass.prototype
    .something
); // boolean

console.log(
  typeof Highcharts
    .seriesTypes.newColumnBAD.prototype
    .pointClass.prototype
    .something
); // undefined
```

I'm not sure if this is expected? 

2. 
Is `...args` really fine in TS? Probably `for`-loop is faster than `Array.prototype.splice.call(arguments, 1)`.

### DIFF
--- a/js/modules/cylinder.src.js
+++ b/js/modules/cylinder.src.js
@@ -49,7 +49,11 @@ seriesType('cylinder', 'column',
 {}, {}, 
 /** @lends Highcharts.seriesTypes.cylinder#pointClass# */
 {
-    shapeType: 'cylinder'
+    shapeType: 'cylinder',
+    hasNewShapeType: H
+        .seriesTypes.column.prototype
+        .pointClass.prototype
+        .hasNewShapeType
 });
 /**
  * A `cylinder` series. If the [type](#series.cylinder.type) option is not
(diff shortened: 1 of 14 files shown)


TorsteinHonsi:
1. Yes this is expected, as Column doesn't (and doesn't need to) have its own Point class.
2. It looks fine to me. The generated legacy code may be a bit on the verbose side, but we should focus on keeping the modern code clean. Soon we will offer options to compile to ES6 and/or ES6 modules and then the rest operator will not be transpiled.

PR #11981, opened 2019-09-16
docs/11803-bar-datalabels

### LINKED ISSUE #11803: Option (series.)dataLabels.enabled not mentioned in documentation
On e.g. https://api.highcharts.com/highcharts/plotOptions.bar.dataLabels the "enabled" attribute is not listed as an possible option for dataLabels, it is only mentioned briefly as an example within the general description of dataLabels. I think it should be added.

bre1470:
Fixed #11803, bar dataLabels option.

Changes require reinstallation of `highchart-documentation-generators` package.

### DIFF
--- a/js/annotations/types/ElliottWave.js
+++ b/js/annotations/types/ElliottWave.js
@@ -52,7 +52,7 @@ H.extendAnnotation(ElliottWave, CrookedLine,
              */
 
             /**
-             * @ignore-options
+             * @internal
              */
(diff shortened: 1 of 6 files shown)


bre1470:
The issue is about the lack of documentation of (unset default) options for data labels. Except for the interface itself the defaults are lacking some options. So instead of having a tree giving a wrong impression of available options, it will become a JSON default and the details are in the common interface.

PR #11984, opened 2019-09-16
docs/typescript

bre1470:
Updated TypeScript beta documentation.


TorsteinHonsi:
There are a lot of changes here, what are the relevant ones? And was this forked out from master before the docusaurus branch was merged in? Maybe it would be easier with a new PR?

bre1470:
This has merged the docusaurus PR, so I wonder why there is a conflict at all. When can I change the documentation?

bre1470:
Is this supposed to be editable at all? I actually do not know where to make it happen, that the TypeScript documentation gets updated. Probably this here is wrong.

PR #12000, opened 2019-09-17
bugfix/11727-log-axis-precission

### LINKED ISSUE #4360: Column Chart - logarithmic scale - column value is integer but appears as long weird decimal
jsfiddle: https://jsfiddle.net/s5bya1uj/1/

The column value which is set to '30' appears as '30.000000000003'.

This happens with a chart of type column, yAxis set to logarithmic and the rest of the yvalues set to 0.

That also makes the chart appear unexpectedly messed up.

### LINKED ISSUE #11727: In logarithmic chart with only one category with value as 650, the yaxis label shows 650.0000000001 instead of 650
#### Expected behaviour
Yaxis label should have 650

#### Actual behaviour
Yaxis label shows 650.0000000001

https://cloud.highcharts.com/edit/243992

#### Product version
latest version

#### Affected browser(s)
All browsers i presume

pawelfus:
Fixed #11727, logarithmic yAxis with just one label sometimes was rendered with numerical errors.
___
Just a small doubt: I'm not sure if this is a way to go (what if there are other values that will require value higher than browsers' supported precision?), I was simply inspired by other commit that resolved #4360

### DIFF
--- a/js/parts/Axis.js
+++ b/js/parts/Axis.js
@@ -3948,8 +3948,8 @@ H.extend(Axis.prototype, /** @lends Highcharts.Axis.prototype */ {
             // The correctFloat cures #934, float errors on full tens. But it
             // was too aggressive for #4360 because of conversion back to lin,
             // therefore use precision 15.
-            axis.min = correctFloat(axis.log2lin(axis.min), 15);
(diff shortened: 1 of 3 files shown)


TorsteinHonsi:
> Just a small doubt

Let's keep it this way, it is likely a matter of tuning.

PR #12032, opened 2019-09-19
Feature/chart menu a11y rewrite

oysteinmoseng:
Improved screen reader compatibility for chart export menu.
___
This PR changes a few things logic wise for the export menu a11y handling. The main difference is the different ARIA attribs used, and the location in the DOM of the proxy containers. In addition, the relevant code has been cleaned up some.

### Changes to non-a11y code:

An exception was added to the unit test testing for unwanted attribs in styled mode, if the `highcharts-a11y-proxy-button` class is set. This because we don't want to confuse users by adding style for the visually hidden proxy buttons in our CSS. Previously these buttons were outside the container tested with the unit test, so this was not an issue.

A new event, `chart.exportMenuShown`, was added in exporting.ts - complimenting the existing `chart.exportMenuHidden`.

### DIFF
--- a/js/modules/accessibility/AccessibilityComponent.js
+++ b/js/modules/accessibility/AccessibilityComponent.js
@@ -10,11 +10,44 @@
 
 'use strict';
 import Highcharts from '../../parts/Globals.js';
-
-var win = Highcharts.win,
+import U from '../../parts/Utilities.js';
+var extend = U.extend,
+    win = Highcharts.win,
     doc = win.document,
     merge = Highcharts.merge,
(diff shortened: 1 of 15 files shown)


oysteinmoseng:
One comment: After the rewrite, a comma was left over at the end of a function call argument list. This was not caught by linting, but could it be? `var a = myfunc(a, b,);` IE chokes on this.

TorsteinHonsi:
> This was not caught by linting, but could it be?
Yes, if there's an ESLint rule for it, we should enable it

Issue #4035, opened 2015-03-25
HighCharts 4.1.4: off-by-one error in line hover width (JSFiddle)


mcayland:
Within our application we allow the user to specify the hover style (particularly line width) on a per-series basis. Since upgrading to HighCharts 4.1.x series, there appears to be an off-by-one error in the lineWidth calculation which gives a lineWidth change on hover even when series.lineWidth === series.states.hover.lineWidth.

To reproduce:

1) Visit the JSFiddle at http://jsfiddle.net/gfuc2j1o/2/

2) Hover the mouse over the line series

Note that while both the series lineWidth and the series hover lineWidth are both set to 3, there is still a visible change in lineWidth during the hover.

Some initial investigation indicates that the lineWidth is being altered in Series.setState():

<pre>
if (state) {
    lineWidth = (stateOptions[state].lineWidth || lineWidth) + (stateOptions[state].lineWidthPlus || 0);
}
</pre>

Since lineWidthPlus always appears to be present and set to 1, the hover lineWidth will always be different to the series lineWidth during the hover state, even when set to the same numeric value.

highslide-software:
#### Workaround
Set `lineWidthPlus` to 0. [View it on JSFiddle](http://jsfiddle.net/highcharts/gfuc2j1o/3/).

PR #12044, opened 2019-09-21
tests/moving-issue-tests-rests

sigrunnu:
Moved issue test #4035 to unit-tests


samples/unit-tests/chart/linewidth/demo.js
```diff
@@ -0,0 +1,93 @@
+QUnit.test('normal:lineWidth and hover:lineWidthPlus #4035', function (assert) {
```
TorsteinHonsi: Move to `unit-tests/series-line/linewidth` instead, as this is a property of line-based series.

PR #12070, opened 2019-09-26
ts/indicators-dpo

activist:
Closes ##11844, Migrated DPOIndicator to TypeScript.

### DIFF
--- a/ts/indicators/dpo.src.ts
+++ b/ts/indicators/dpo.src.ts
@@ -1,0 +1,197 @@
+/* *
+ *
+ *  License: www.highcharts.com/license
+ *
+ *  !!!!!!! SOURCE GETS TRANSPILED BY TYPESCRIPT. EDIT TS FILE ONLY. !!!!!!!
+ *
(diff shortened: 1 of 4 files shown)


ts/indicators/dpo.src.ts
```diff
@@ -0,0 +1,197 @@
...
+                params: DPOIndicatorParamsOptions
+            ): (boolean|IndicatorValuesObject);
+        }
+
+        interface DPOIndicatorOptions extends SMAIndicatorOptions {
+            params?: DPOIndicatorParamsOptions;
+        }
+
+        interface DPOIndicatorParamsOptions extends SMAIndicatorParamsOptions {
+            period: number;
```
bre1470: This should be optional.

activist: Ok, so then it should be cast as any in the top of the implementation? (period is optional in SMAIndicator)

activist: Fixed

PR #12086, opened 2019-09-27
ts/indicators-ema

### LINKED ISSUE #11845: Migrate `indicators/ema` to TypeScript

activist:
Closes #11845 , Migrated EMAIndicator to TypeScript.

### DIFF
--- a/ts/indicators/aroon.src.ts
+++ b/ts/indicators/aroon.src.ts
@@ -169,12 +169,12 @@ H.seriesType<Highcharts.AroonIndicator>(
         pointValKey: 'y',
         linesApiNames: ['aroonDown'],
         getValues: function (
-            series: Highcharts.AroonIndicator,
+            series: Highcharts.Series,
             params: Highcharts.AroonIndicatorParamsOptions
         ): Highcharts.IndicatorMultipleValuesObject {
             var period = (params.period as any),
                 xVal: Array<number> = (series.xData as any),
-                yVal: Array<Array<number>> = series.yData,
+                yVal: Array<Array<number>> = (series.yData as any),
                 yValLen = yVal ? yVal.length : 0,
                 // 0- date, 1- Aroon Up, 2- Aroon Down
                 AR: Array<Array<number>> = [],
--- a/ts/indicators/aroon.src.ts
+++ b/ts/indicators/aroon.src.ts
@@ -30,7 +30,6 @@ declare global {
             public pointClass: typeof AroonIndicatorPoint;
             public points: Array<AroonIndicatorPoint>;
             public pointValKey: MultipleLinesMixin['pointValKey'];
-            public yData: Array<Array<number>>;
             public drawGraph: MultipleLinesMixin['drawGraph'];
             public getTranslatedLinesNames: MultipleLinesMixin[
                 'getTranslatedLinesNames'
--- a/ts/indicators/atr.src.ts
+++ b/ts/indicators/atr.src.ts
@@ -138,15 +138,15 @@ seriesType<Highcharts.ATRIndicator>(
      */
     {
         getValues: function (
-            series: Highcharts.ATRIndicator,
+            series: Highcharts.Series,
             params: Highcharts.ATRIndicatorParamsOptions
         ): (boolean|Highcharts.IndicatorValuesObject) {
-            var period = (params.period as any),
-                xVal = (series.xData as any),
-                yVal = series.yData,
-                yValLen = yVal ? yVal.length : 0,
-                xValue = (xVal as any)[0],
-                yValue = yVal[0],
+            var period: number = (params.period as any),
+                xVal: Array<number> = (series.xData as any),
+                yVal: Array<Array<number>> = (series.yData as any),
+                yValLen: number = yVal ? yVal.length : 0,
+                xValue: number = (xVal as any)[0],
+                yValue: Array<number> = yVal[0],
                 range = 1,
                 prevATR = 0,
                 TR = 0,
@@ -169,7 +169,7 @@ seriesType<Highcharts.ATRIndicator>(
 
             for (i = 1; i <= yValLen; i++) {
 
-                accumulateAverage(points, (xVal as any), yVal, i);
+                accumulateAverage(points, xVal, yVal, i);
 
                 if (period < range) {
                     point = populateAverage(
--- a/ts/indicators/atr.src.ts
+++ b/ts/indicators/atr.src.ts
@@ -20,7 +20,6 @@ declare global {
             public data: Array<ATRIndicatorPoint>
             public pointClass: typeof ATRIndicatorPoint;
             public points: Array<ATRIndicatorPoint>;
-            public yData: Array<Array<number>>;
             public getValues(
                 series: Series,
                 params: ATRIndicatorParamsOptions
--- a/ts/indicators/bollinger-bands.src.ts
+++ b/ts/indicators/bollinger-bands.src.ts
@@ -196,13 +196,13 @@ H.seriesType<Highcharts.BBIndicator>(
         },
         getValues: function (
             this: Highcharts.BBIndicator,
-            series: Highcharts.BBIndicator,
+            series: Highcharts.Series,
             params: Highcharts.BBIndicatorParamsOptions
         ): (boolean|Highcharts.IndicatorMultipleValuesObject) {
             var period: number = (params.period as any),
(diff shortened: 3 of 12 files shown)


ts/indicators/ema.src.ts
```diff
@@ -0,0 +1,230 @@
...
+        calculateEma: function (
+            xVal: (Array<number>|undefined),
+            yVal: Array<Array<number>>,
+            i: number,
+            EMApercent: number,
+            calEMA: (number|undefined),
+            index: number,
+            SMA: number
+        ): [number, number] {
+            var x: number = (xVal as any)[i - 1],
```
activist: @bre1470  I'm a bi uncertain if this is correct as xVal could be undefined. But if not it could possibly mean yet another type of `IndicatorValueObject`. Unless there is another way around of course..

bre1470: I think it is actually not possible at the stage of `getValues`, that `xData` of the linkes series is undefined. So this is probably not correct.

ts/indicators/ema.src.ts
```diff
@@ -0,0 +1,230 @@
...
+
+            y = calEMA === undefined ?
+                SMA : correctFloat(((yValue as any) * EMApercent) +
+                (calEMA * (1 - EMApercent)));
+
+            return [x, y];
+        },
+        getValues: function (
+            this: Highcharts.EMAIndicator,
+            series: Highcharts.EMAIndicator,
```
bre1470: Here the `series` parameters is about the linked parent series. Therefore type `Highcharts.Series` is correct. At the same time you do not need `yData` on the indicator class, so it can be removed in the declaration.

ts/indicators/ema.src.ts
```diff
@@ -0,0 +1,230 @@
...
+            params: Highcharts.EMAIndicatorParamsOptions
+        ): (boolean|Highcharts.IndicatorValuesObject) {
+            var period: number = (params.period as any),
+                xVal: (Array<number>|undefined) = series.xData,
+                yVal: Array<Array<number>> = series.yData,
```
bre1470: You have to make use of `as any` on the right side for both value arrays. Additionally `xVal` should not be undefinable.

ts/indicators/ema.src.ts
activist: Ok, I have made some changes, but this still confuses me in the sense that we previously moved the `yData` in order to be able to define it in the indicator. So, that change was perhaps not necessary given that it is the linkedParent that is passed to getValues?

Also given the above, I think the same fix needs to be applied in other indicators as well.

bre1470: That is then probably incorrect.`yData` on the indicator is not in direct usage as far as I know.

The `series` parameter of `Indicator.getValues` is always about the linked parent series, not about the indicator.

We created in CMF an interface for OHLC-like series with a two-dimensional `yData` property to cover any compatible series type: https://github.com/highcharts/highcharts/blob/master/ts/indicators/cmf.src.ts#L62-L64

activist: I have now changed the Series type in the getValues(..) implementation of several indicators. Please review @bre1470 :)

ts/indicators/aroon.src.ts
```diff
@@ -169,12 +169,12 @@ H.seriesType<Highcharts.AroonIndicator>(
...
         linesApiNames: ['aroonDown'],
         getValues: function (
-            series: Highcharts.AroonIndicator,
+            series: Highcharts.Series,
             params: Highcharts.AroonIndicatorParamsOptions
         ): Highcharts.IndicatorMultipleValuesObject {
             var period = (params.period as any),
                 xVal: Array<number> = (series.xData as any),
-                yVal: Array<Array<number>> = series.yData,
+                yVal: Array<Array<number>> = (series.yData as any),
```
bre1470: Additionally the `AroonIndicator.yData` property should be removed in the class declaration.

ts/indicators/atr.src.ts
```diff
@@ -138,15 +138,15 @@ seriesType<Highcharts.ATRIndicator>(
...
         ): (boolean|Highcharts.IndicatorValuesObject) {
-            var period = (params.period as any),
-                xVal = (series.xData as any),
-                yVal = series.yData,
-                yValLen = yVal ? yVal.length : 0,
-                xValue = (xVal as any)[0],
-                yValue = yVal[0],
+            var period: number = (params.period as any),
+                xVal: Array<number> = (series.xData as any),
+                yVal: Array<Array<number>> = (series.yData as any),
```
bre1470: Additionally the `ATRIndicator.yData` property should be removed in the class declaration.

ts/indicators/bollinger-bands.src.ts
```diff
@@ -196,13 +196,13 @@ H.seriesType<Highcharts.BBIndicator>(
...
             this: Highcharts.BBIndicator,
-            series: Highcharts.BBIndicator,
+            series: Highcharts.Series,
             params: Highcharts.BBIndicatorParamsOptions
         ): (boolean|Highcharts.IndicatorMultipleValuesObject) {
             var period: number = (params.period as any),
                 standardDeviation: number = (params.standardDeviation as any),
                 xVal: Array<number> = (series.xData as any),
-                yVal: Array<Array<number>> = series.yData,
+                yVal: Array<Array<number>> = (series.yData as any),
```
bre1470: Additionally the `BBIndicator.yData` property should be removed in the class declaration.

ts/indicators/cci.src.ts
```diff
@@ -111,12 +111,12 @@ seriesType<Highcharts.CCIIndicator>(
...
     {
         getValues: function (
-            series: Highcharts.CCIIndicator,
+            series: Highcharts.Series,
             params: Highcharts.CCIIndicatorParamsOptions
         ): (boolean|Highcharts.IndicatorValuesObject) {
             var period: number = (params.period as any),
                 xVal: Array<number> = (series.xData as any),
-                yVal: Array<Array<number>> = series.yData,
+                yVal: Array<Array<number>> = (series.yData as any),
```
bre1470: Additionally the `CCIIndicator.yData` property should be removed in the class declaration.

ts/indicators/dpo.src.ts
```diff
@@ -123,15 +123,16 @@ H.seriesType<Highcharts.DPOIndicator>(
...
             params: Highcharts.DPOIndicatorParamsOptions
         ): (boolean|Highcharts.IndicatorValuesObject) {
             var period: number = (params.period as any),
                 index: number = (params.index as any),
                 offset: number = Math.floor(period / 2 + 1),
                 range: number = period + offset,
                 xVal: Array<number> = series.xData || [],
-                yVal: (Array<number>|Array<Array<number>>) = series.yData || [],
+                yVal: (Array<number>|Array<Array<number>>) =
+                    (series.yData as any) || [],
```
bre1470: Additionally the `DPOIndicator.yData` property should be removed in the class declaration.

ts/indicators/wma.src.ts
```diff
@@ -151,12 +151,12 @@ seriesType<Highcharts.WMAIndicator>(
...
     {
         getValues: function (
-            series: Highcharts.WMAIndicator,
+            series: Highcharts.Series,
             params: Highcharts.WMAIndicatorParamsOptions
         ): (boolean|Highcharts.IndicatorValuesObject) {
             var period: number = params.period as any,
                 xVal: Array<number> = (series.xData as any),
-                yVal: Array<Array<number>> = series.yData,
+                yVal: Array<Array<number>> = (series.yData as any),
```
bre1470: Additionally the `WMAIndicator.yData` property should be removed in the class declaration.

ts/indicators/ema.src.ts
bre1470: This one is solved. 👍

activist:
> The `yData` property in the indicator classes should be removed.

I assume this also goes for ABandsIndicator, ADIndicator and AOIndicator?

bre1470:
> > The `yData` property in the indicator classes should be removed.
> 
> I assume this also goes for ABandsIndicator, ADIndicator and AOIndicator?

Probably yes - but until now, they are not part of this PR. :)

activist:
> > > The `yData` property in the indicator classes should be removed.
> > 
> > 
> > I assume this also goes for ABandsIndicator, ADIndicator and AOIndicator?
> 
> Probably yes - but until now, they are not part of this PR. :)

Now they are :)

PR #12118, opened 2019-10-02
ts/indicators-keltner-channels

### LINKED ISSUE #11847: Migrate `indicators/keltner-channels` to TypeScript

activist:
Closes #11847, Tools: Migrated KeltnerChannelsIndicator to TS.

### DIFF
--- a/ts/indicators/keltner-channels.src.ts
+++ b/ts/indicators/keltner-channels.src.ts
@@ -23,13 +23,9 @@ declare global {
             extends SMAIndicator implements MultipleLinesIndicator {
             public data: Array<KeltnerChannelsIndicatorPoint>;
             public linesApiNames: MultipleLinesMixin['linesApiNames'];
-            public nameBase: string;
-            public nameComponents: Array<string>;
             public options: KeltnerChannelsIndicatorOptions;
-            public pointArrayMap: MultipleLinesMixin['pointArrayMap'];
             public pointClass: typeof KeltnerChannelsIndicatorPoint;
             public points: Array<KeltnerChannelsIndicatorPoint>;
-            public pointValKey: MultipleLinesMixin['pointValKey'];
             public getTranslatedLinesNames: MultipleLinesMixin[
                 'getTranslatedLinesNames'
             ];
@@ -204,7 +200,7 @@ H.seriesType<Highcharts.KeltnerChannelsIndicator>(
                 BL: number,
                 date: number,
                 seriesEMA: (boolean|Highcharts.IndicatorValuesObject) =
-                    (EMA as any).prototype.getValues(series,
+                    EMA.prototype.getValues(series,
                         {
                             period: period,
                             index: index
@@ -259,4 +255,4 @@ H.seriesType<Highcharts.KeltnerChannelsIndicator>(
  * @apioption    series.keltnerchannels
  */
 
-''; // to avoid removal of the above jsdoc
+''; // to include the above in the js output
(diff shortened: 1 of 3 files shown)


ts/indicators/keltner-channels.src.ts
```diff
@@ -0,0 +1,258 @@
...
+    /**
+     * @lends Highcharts.Series#
+     */
+    merge(multipleLinesMixin, {
+        pointArrayMap: ['top', 'middle', 'bottom'],
+        pointValKey: 'middle',
+        nameBase: 'Keltner Channels',
+        nameComponents: ['period', 'periodATR', 'multiplierATR'],
+        linesApiNames: ['topLine', 'bottomLine'],
+        requiredIndicators: ['ema', 'atr'],
```
bre1470: All these indicator properties are missing in the namespace declaration of the indicator. Additionally you should cast this second merge parameter to the indicator type.

ts/indicators/keltner-channels.src.ts
```diff
@@ -0,0 +1,258 @@
...
+                seriesATR: (boolean|Highcharts.IndicatorValuesObject) =
+                    ATR.prototype.getValues(series,
+                        {
+                            period: periodATR
+                        }),
+                pointEMA: Array<number>,
+                pointATR: Array<number>,
+                xData: Array<number> = [],
+                yData: Array<Array<number>> = [],
+                i;
```
bre1470: Type is missing.

ts/indicators/keltner-channels.src.ts
```diff
@@ -0,0 +1,258 @@
...
+                },
+                bottomLine: {
+                    styles: {
+                        lineColor: this.color
+                    }
+                }
+            }, this.options);
+        },
+        getValues: function (
+            series: Highcharts.KeltnerChannelsLinkedParentSeries,
```
bre1470: Again this is about the reference series (the linked parent).

ts/indicators/keltner-channels.src.ts
```diff
@@ -0,0 +1,258 @@
...
+                yData.push([TL, ML, BL]);
+            }
+
+            return {
+                values: KC,
+                xData: xData,
+                yData: yData
+            };
+        }
+    })
```
bre1470: Here you can cast it like `} as Partial<Highcharts.KeltnerChannelIndicator>)`

ts/indicators/keltner-channels.src.ts
bre1470: My fault - it is. :)

ts/indicators/keltner-channels.src.ts
activist: Thanks missed it. I was a bit too quick to submit this PR. As for the optional properties in the implemented interface I understand it is needed, but do we also explicitly add them again even when they are present in the parent class?

activist: Ok, we should explicitly add the properties even though they are declared in parent class? A good practice, although not strictly needed?

bre1470: Correct. The general problem is, that we are still in the transition phase and because of this, types might change or properties get optional. Conflicts can be easier detected, if every declaration contains the properties that are set during initial run.

ts/indicators/keltner-channels.src.ts
```diff
@@ -0,0 +1,263 @@
...
+            marker?: PointMarkerOptionsObject;
+            params?: KeltnerChannelsIndicatorParamsOptions;
+            tooltip?: TooltipOptions;
+            topLine?: Dictionary<CSSObject>;
+        }
+
+        interface KeltnerChannelsIndicatorParamsOptions
+            extends SMAIndicatorParamsOptions {
+            periodATR?: number;
+            multiplierATR: number;
```
bre1470: This one should be optional.

activist: Fixed

PR #12138, opened 2019-10-03
Fixed #12090, a11y module breaks legend navigation.

### LINKED ISSUE #12090: Accessibility module breaks legend's navigation
#### Expected behaviour
Legend navigation should work.

#### Actual behaviour
It doesn't because of accessibility module. Everything works well if the module is not imported.

https://jsfiddle.net/BlackLabel/1d5qarn9/ 
Click legend navigation arrow.

#### Product version
Highcharts 7.2.0

oysteinmoseng:
Fixed #12090, a11y module breaks legend navigation.
___
A11y legend proxy buttons were overlaying navigation arrows. These are now hidden unless the corresponding legend item is on the current page.

Added a `legend.afterScroll` event that fires after legend scroll animation is complete in order to update proxies.

No other changes to non-a11y files.

### DIFF
--- a/ts/parts/Legend.ts
+++ b/ts/parts/Legend.ts
@@ -1542,7 +1542,8 @@ Highcharts.Legend.prototype = {
         scrollBy: number,
         animation?: (boolean|Highcharts.AnimationOptionsObject)
     ): void {
-        var pages = this.pages,
+        var chart = this.chart,
+            pages = this.pages,
             pageCount = pages.length,
             currentPage = (this.currentPage as any) + scrollBy,
             clipHeight = this.clipHeight,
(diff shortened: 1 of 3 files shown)


ts/parts/Legend.ts
```diff
@@ -1622,8 +1623,15 @@ Highcharts.Legend.prototype = {
 
             this.currentPage = currentPage;
             this.positionCheckboxes();
-        }
 
+            // Fire event after scroll animation is complete
+            const animOptions = H.animObject(
+                pick(animation, chart.renderer.globalAnimation, true)
+            );
+            setTimeout((): void => {
```
TorsteinHonsi: Use `syncTimeout` so that it will run sync when duration is 0.

oysteinmoseng: Fixed in separate PR

PR #12141, opened 2019-10-03
Tests/moving last issue tests

sigrunnu:
Moved the last issue tests.
All committed with `--no-verify`


TorsteinHonsi:
@sigrunnu There is one test failing, it can be resolved by adding the containers programmatically like you did with the other similar cases.

samples/unit-tests/3d/reversedstacks/demo.js
```diff
@@ -0,0 +1,67 @@
...
+
+    // reversed stacks
+    for (var j = chart2.series.length - 1; j >= 1; j--) {
+        assert.strictEqual(
+            chart2.series[j].options.zIndex < chart2.series[j - 1].options.zIndex,
+            true,
+            'Series descendig order'
+        );
+    }
+});
```
TorsteinHonsi: Missing clean-up ofter the test finishes. The two generated charts should be destroyed, and the two new divs be removed from the DOM.

Issue #1171, opened 2012-08-13
Pie chart - stickytracking / tooltip


sebastianbochan:
If set stickyTracking is false on a pie chart, and move your mouse over a tooltip,it starts to oscillate. 

http://jsfiddle.net/kzoon/C7bq2

In case when /master/highcharts.src.js is used, then tooltip appears in the top left corner.

PR #12146, opened 2019-10-04
bugfix/1172-column-navigator

### LINKED ISSUE #1172: Column type navigator
Example: http://jsfiddle.net/KeGjx/4/

When using columns in navigator, they will be drawn outside navigator.  Should be cut  or takes less space, I think.

### LINKED ISSUE #3136: cropThreshold doesn't work
Example: http://jsfiddle.net/n6EhL/3/

Try to zoom-in by dragging one of navigator's handles. First column is missing even when we set `cropThreshold`.

Maybe be connected to #2437.

Workaround is to set `cropShoulder` directly in series prototype.

pawelfus:
Fixed #1172, handles positions and `xAxis` extremes were incorrect when using column type series in navigator.
___
Note: This does not fix #3136 - it would be nice to have an official API option to change `cropShoulder`. Otherwise most left/right columns are hidden even when xAxis labels (due to extra space created by columns) are rendered. 

Fixed demos from #1171:

- http://jsfiddle.net/BlackLabel/6rzqoasu/1/
- http://jsfiddle.net/BlackLabel/Lmhqngu5/
- http://jsfiddle.net/BlackLabel/yncevr5j/1/

### DIFF
--- a/ts/parts/Navigator.ts
+++ b/ts/parts/Navigator.ts
@@ -505,6 +505,15 @@ extend(defaultOptions, {
              * The type of the navigator series. Defaults to `areaspline` if
              * defined, otherwise `line`.
              *
+             * Heads up:
+             * When using `column` type navigator, don't forget to change
+             * [pointRange](#navigator.series.pointRange) to `null`.
+             * In column-type navigator, zooming is limited to at least one
+             * point with it's `pointRange`.
+             *
+             * @sample {highstock} stock/navigator/column/
+             *         Column type navigator
+             *
              * @type    {string}
              * @default areaspline
              */
--- a/ts/parts/Navigator.ts
+++ b/ts/parts/Navigator.ts
@@ -797,6 +797,7 @@ Axis.prototype.toFixedRange = function (
     fixedMax?: number
 ): Highcharts.RangeObject {
     var fixedRange = this.chart && this.chart.fixedRange,
+        halfPointRange = (this.pointRange || 0) / 2,
         newMin = pick<number|undefined, number>(
             fixedMin, this.translate(pxMin as any, true, !this.horiz) as any
         ),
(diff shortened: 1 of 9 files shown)


pawelfus:
That's interesting. This (new) demo should have small diff: http://utils.highcharts.local/samples/#test/stock/navigator/column

> Left version: 7.2.0; right version: 7.2.0-modified
The generated SVG is different, checking exported images...
The exported images are different (dissimilarity index: 0.02)

ts/parts/Navigator.ts
```diff
@@ -505,6 +505,15 @@ extend(defaultOptions, {
              * The type of the navigator series. Defaults to `areaspline` if
              * defined, otherwise `line`.
              *
+             * Heads up:
```
TorsteinHonsi: Could we handle this with smarter default options so that we don't need to do manual warnings?

ts/parts/Navigator.ts
```diff
@@ -505,6 +505,15 @@ extend(defaultOptions, {
...
+             * When using `column` type navigator, don't forget to change
+             * [pointRange](#navigator.series.pointRange) to `null`.
+             * In column-type navigator, zooming is limited to at least one
+             * point with it's `pointRange`.
```
TorsteinHonsi: Typo: ... with its `pointRange`.

ts/parts/Navigator.ts
pawelfus: Sure, not a problem

pawelfus:
Default `pointRange` was a bit more tricky than expected - `type` can be defined in multiple ways (default `defaultSeriesType`, `navigator.series.type` and `series.navigatorOptions.type`), so first type is resolved, then related `pointRange` is picked up.

ts/parts/Navigator.ts
```diff
@@ -2234,6 +2283,17 @@ Navigator.prototype = {
                     baseNavigatorOptions
                 );
 
+                // Once nav series type is resolved, pick correct pointRange
+                mergedNavSeriesOptions.pointRange = pick(
+                    // Stricte set pointRange in options
+                    userNavOptions.pointRange,
+                    baseNavigatorOptions.pointRange,
+                    // Fallback to default values, e.g. `null` for column
+                    (defaultOptions.plotOptions as any)[
```
pawelfus: @bre1470 - I know I introduce here new `as any`, but the same happens in other files where we need to reference `defaultOptions.plotOptions`. I assume this will be picked altogether with other changes?

bre1470: I believe this should not be necessary inside the pick function anymore. @jon-a-nygaard made it type safe.

bre1470: Oh I see now, that it is used with an indexer. `any` is okay in this case.

pawelfus: Ok, thanks! Sorry, should have highlight more lines to make it more readable..

PR #12148, opened 2019-10-04
ts/indicators-pivot-points

### LINKED ISSUE #11852: Migrate `indicators/pivot-points` to TypeScript

activist:
Closes #11852, Tools: Migrated PivotPointsIndicator to TS.

### DIFF
--- a/ts/indicators/pivot-points.src.ts
+++ b/ts/indicators/pivot-points.src.ts
@@ -52,6 +52,8 @@ declare global {
         }
 
         class PivotPointsIndicatorPoint extends SMAIndicatorPoint {
+            public destroy(): void;
+            public destroyElements(): void;
             public P: number;
             public pivotLine: string;
             public series: PivotPointsIndicator;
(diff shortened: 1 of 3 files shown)


ts/indicators/pivot-points.src.ts
```diff
@@ -0,0 +1,483 @@
...
+
+        interface PivotPointsIndicatorParamsOptions
+            extends SMAIndicatorParamsOptions {
+            algorithm?: string;
+        }
+
+        class PivotPointsIndicatorPoint extends SMAIndicatorPoint {
+            public P: number;
+            public pivotLine: string;
+            public series: PivotPointsIndicator;
```
bre1470: The `PivotPointsIndicatorPoint` class is lacking the `destroy` and `destroyElements` functions, that are defined below.

activist: But aren't these defined by Point and thus declared and available through inheritance?

bre1470: No, only the type is defined by inheritence, but the implementation is not. You overwrite the inherited functions. Therefor you have to define it again, so that inheritence conflicts in future changes show up on the right spot.

activist: Added missing function declarations to `PivotPointsIndicatorPoint`

PR #12149, opened 2019-10-04
bugfix/12073-annotations-draggable-points

### LINKED ISSUE #12073: Dragging annotation changing the data values of draggable points.
#### Expected behaviour
Changing annotation position should not affect on data values.

#### Actual behaviour
When a series have draggable points enabled, and annotation is added to the chart, then the data of specific point (above which the "annotation" is currently being dragged) is changing while moving the annotation's label.

Live example: https://jsfiddle.net/ag5jsf63/
1. Click on **Add Annotation** button
2. Try to move the added annotation, then observe what's happening with closest point values

#### Product version
From v7.1.3

pawelfus:
Fixed #12073, dragging an annotation used to change points' values too.
___

@oysteinmoseng - do you think it's okay to add this `if` in draggable-points module? Maybe you have a better idea? I would rather keep annotations logic in annotations plugin..

@bre1470 - annotations are not TS-ready, so I'm not sure if this is fine: `(chart as any).hasDraggedAnnotation`

### DIFF
--- a/ts/modules/draggable-points.src.ts
+++ b/ts/modules/draggable-points.src.ts
@@ -29,6 +29,8 @@ declare global {
             dragGuideBox?: SVGElement;
             /** @requires modules/draggable-points */
             hasAddedDragDropEvents?: boolean;
+            /** @requires modules/annotations */
+            hasDraggedAnnotation?: boolean;
             /** @requires modules/draggable-points */
             isDragDropAnimating?: boolean;
             /** @requires modules/draggable-points */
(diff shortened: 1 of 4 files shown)


ts/modules/draggable-points.src.ts
```diff
@@ -2962,6 +2962,11 @@ function mouseDown(
         return;
     }
 
+    // Ignore if dragging an annotation
+    if ((chart as any).hasDraggedAnnotation) {
```
bre1470: Please do not cast to `any`. Instead go in this file up to the namespace and add the property as an optional boolean to the chart interface (and add a mini doclet, that it requires the module).

pawelfus: Done, thanks! :)

PR #12150, opened 2019-10-04
bugfix/11650-highcharts-tooltip-header-class

### LINKED ISSUE #11650: 'highcharts-tooltip-header' class missing for the tooltip header
#### Expected behaviour
Expecting a default class 'highcharts-tooltip-header' for the tooltip

#### Actual behaviour
Not able to differentiate the Tooltip Header from the other series tooltips

1. Hit the url https://jsfiddle.net/289e40xy/ with highstock v6.2.0
2. Try Removing the callout rendered in svg using 

```
g.highcharts-tooltip-header>path{
    fill: transparent !important;
    stroke-width: 0px !important;
}
```
3. As useHTML set to true , add styles to the haeder

```
.highcharts-tooltip-header>span{
    border: 2px solid darkgrey;
    height: 25px;
    border-radius: 6px;
    background: blue;
    color:yellow !important;
}
```
4. Add the same set of classes in v7.1.2..It doesnt seem to work

_Observation_

In v7.1.2, https://jsfiddle.net/6mgyscnj/ -> Styles are not getting affected

style not affected

In v6.2.0, https://jsfiddle.net/289e40xy/1/ -> Styles are getting affected

styles getting affected

#### Product version

Tested out in v6.2.0 > class 'highcharts-tooltip-header' is present
Tested out in v7.1.2 > class 'highcharts-tooltip-header' is missing

#### Affected browser(s)
Charome v72

pawelfus:
Fixed #11650, added missing `highcharts-tooltip-header` classname to the tooltip's header.
___
Additionally we can:

- remove `highcharts-header` class: https://github.com/highcharts/highcharts/blob/5e5553521eb50408d1fb4674d71cc5328c85ae7a/ts/parts/Tooltip.ts#L1716-L1719
- and move CSS under `highcharts-tooltip-header`: https://github.com/highcharts/highcharts/blob/5e5553521eb50408d1fb4674d71cc5328c85ae7a/css/highcharts.scss#L205-L207

### DIFF
--- a/docs/chart-design-and-style/style-by-css.md
+++ b/docs/chart-design-and-style/style-by-css.md
@@ -443,6 +443,7 @@ Text styles for the title. Replaces [title.style](https://api.highcharts.com/hi
     .highcharts-tooltip  
     .highcharts-tooltip-box  
     .highcharts-tooltip text
+    .highcharts-tooltip-header
 
 Styles for the tooltip. The tooltip box is the shape or path where the background and border can be set. Text styles should be applied to the text element.
 
--- a/js/parts/Tooltip.js
+++ b/js/parts/Tooltip.js
@@ -855,7 +855,10 @@ H.Tooltip.prototype = {
                         .label(null, null, null, (point.isHeader ?
                         options.headerShape :
                         options.shape) || 'callout', null, null, options.useHTML)
-                        .addClass('highcharts-tooltip-box ' + colorClass)
+                        .addClass(point.isHeader ?
+                        'highcharts-tooltip-header ' : '' +
+                        'highcharts-tooltip-box ' +
+                        colorClass)
                         .attr(attribs)
                         .add(tooltipLabel);
                 }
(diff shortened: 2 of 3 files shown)


pawelfus:
The problem is with missing `hc-tooltip-header` class, not `hc-header` class, compare these two examples:

- v6.1: https://jsfiddle.net/BlackLabel/6utL0zq5/
Screen Shot 2019-10-15 at 15 24 19

- v7.2: https://jsfiddle.net/BlackLabel/dkh41rp7/
Screen Shot 2019-10-15 at 15 24 03

Without `hc-tooltip-header` class it's harder to style the label box (`g.highcharts-tooltip .highcharts-label:first-child>path` vs `.highcharts-tooltip-header path`).

Note that PR does not change how we style it, only adds the classname. I thought about changing the styling too (as in my comment above), but I wasn't sure if that won't break something (and as you pointed out, it would be inconsistent).

PR #12124, opened 2019-10-03
ts/indicators-ichimoku-kinko-hyo

activist:
Closes #11846, Tools: Migrated IKHIndicator to TypeScript.

PR #12153, opened 2019-10-07
ts/indicator-price-envelopes

### LINKED ISSUE #11854: Migrate `indicators/price-envelopes` to TypeScript

activist:
Closes #11854, Tools: Migrated PriceEnvelopesIndicator to TS.

### DIFF
--- a/ts/indicators/price-envelopes.src.ts
+++ b/ts/indicators/price-envelopes.src.ts
@@ -23,7 +23,7 @@ declare global {
             public init(): void;
             public getValues(
                 series: Series,
-                params: PriceEnvelopesIndicatorOptionsParams
+                params: PriceEnvelopesIndicatorParamsOptions
             ): (boolean|IndicatorMultipleValuesObject);
             public nameBase: string;
             public options: PriceEnvelopesIndicatorOptions;
@@ -38,7 +38,15 @@ declare global {
             public translate(): void;
         }
 
-        interface PriceEnvelopesIndicatorOptionsParams
+        interface PriceEnvelopesIndicatorGappedExtensionObject {
+            options?: PriceEnvelopesIndicatorGappedExtensionOptions;
+        }
+
+        interface PriceEnvelopesIndicatorGappedExtensionOptions {
+            gapSize?: number;
+        }
+
+        interface PriceEnvelopesIndicatorParamsOptions
             extends SMAIndicatorParamsOptions {
             topBand?: number;
             bottomBand?: number;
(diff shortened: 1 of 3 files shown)


activist:
Awaiting merge of #12124 for further work.

bre1470:
There is no change in `.eslintignore`.

activist:
> There is no change in `.eslintignore`.

Thanks, surprisingly the commit didn't produce any errors after being converted to ts. Any idea why? Too me it looks like the output js should fail on e.g. max line length.

bre1470:
I guess, somehow every case is matching the excluding pattern: `(eslint-disable-line|pointFormat|headerFormat| highcharts\/|\\]\\(#|@apioption|@borrows|@default|@extends|@name|@param|@property|@return|@sample|@see|@type|@typedef|@validvalue)` 🤷‍♀️

ts/indicators/price-envelopes.src.ts
```diff
@@ -0,0 +1,339 @@
...
+            public points: Array<PriceEnvelopesIndicatorPoint>;
+            public pointValKey: string;
+            public toYData(
+                point: PriceEnvelopesIndicatorPoint
+            ): [number, number, number];
+            public translate(): void;
+        }
+
+        interface PriceEnvelopesIndicatorGappedExtensionObject {
+            [index: string]: Dictionary<(number|undefined)>;
```
bre1470: If you go that far, you can remove the interface and use `Dictionary<Dictionary<(number|undefined)>>`.

ts/indicators/price-envelopes.src.ts
```diff
@@ -0,0 +1,339 @@
...
+        }
+
+        interface PriceEnvelopesIndicatorOptionsParams
```
bre1470: Has to be named `PriceEnvelopesIndicatorParamsOptions`.

ts/indicators/price-envelopes.src.ts
activist: This same structure seems to appear in a few different indicators. Could it be better to define this as separate interfaces in one of the the indicators and the reuse the same in the other indicators, or do you think that they should be declared a separate interfaces per indicator?

bre1470: That is hard to tell, but as there is no mixin in use, I would go with separate interfaces.

activist: Added separate interfaces. I hope the naming is ok `PriceEnvelopesIndicatorGappedExtensionObject` and `PriceEnvelopesIndicatorGappedExtensionOptions`

ts/indicators/price-envelopes.src.ts
activist: Fixed

PR #12155, opened 2019-10-07
ts/indicators-regressions

### LINKED ISSUE #11856: Migrate `indicators/regressions` to TypeScript

activist:
Closes #11856, Tools: Migrated LinearRegressionsIndicator to TS.

### DIFF
--- a/ts/indicators/regressions.src.ts
+++ b/ts/indicators/regressions.src.ts
@@ -1,0 +1,600 @@
+/**
+ *
+ *  (c) 2010-2019 Kamil Kulig
+ *
+ *  License: www.highcharts.com/license
+ *
+ *  !!!!!!! SOURCE GETS TRANSPILED BY TYPESCRIPT. EDIT TS FILE ONLY. !!!!!!!
+ *
+ * */
+
+'use strict';
+
+import H from '../parts/Globals.js';
+
+/**
+ * Internal types
+ * @private
+ */
+declare global {
+    namespace Highcharts {
+        class LinearRegressionIndicator extends SMAIndicator {
+            public data: Array<LinearRegressionIndicatorPoint>;
+            public findClosestDistance(
+                xData: Array<number>
+            ): (number|undefined);
+            public getEndPointY(
+                lineParameters: RegressionLineParametersObject,
+                endPointX: number
+            ): number;
+            public getRegressionLineParameters(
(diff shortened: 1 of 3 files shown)


ts/indicators/regressions.src.ts
```diff
@@ -0,0 +1,601 @@
...
+         *          of a straight line function
+         * @param {number} endPointX - x coordinate of the point
+         * @return {number} - y value of the point that lies on the line
+         */
+        getEndPointY: function (
+            lineParameters: Highcharts.RegressionLineParametersObject,
+            endPointX: number
+        ): number {
+            return (lineParameters.slope as any) * endPointX +
+                (lineParameters.intercept as any);
```
bre1470: The interface should have the properties non-optional, so that you can remove the `any` cast.

ts/indicators/regressions.src.ts
```diff
@@ -0,0 +1,601 @@
...
+    {},
+    /**
+     * @lends Highcharts.Series#
+     */
+    {
+        nameBase: 'Linear Regression Slope Indicator',
+        getEndPointY: function (
+            lineParameters: Highcharts.RegressionLineParametersObject
+        ): number {
+            return (lineParameters.slope as any);
```
bre1470: Same here.

ts/indicators/regressions.src.ts
```diff
@@ -0,0 +1,601 @@
...
+        */
+        slopeToAngle: function (slope: number): number {
+            return Math.atan(slope) * (180 / Math.PI); // rad to deg
+        },
+
+        getEndPointY: function (
+            this: Highcharts.LinearRegressionAngleIndicator,
+            lineParameters: Highcharts.RegressionLineParametersObject
+        ): number {
+            return this.slopeToAngle(lineParameters.slope as any);
```
bre1470: Same here.

ts/indicators/regressions.src.ts
activist: While I wholeheartedly agree it is a bit unclear to me as to when it is ok to make interfaces with non-optional properties in phase one of the migration. If it is used as a pure value object it is ok?

activist:
@bre1470  Added requested fixes

activist:
> Please resolve conflicts.

Fixed :)

PR #12166, opened 2019-10-08
ts/indicators-supertrend

### LINKED ISSUE #11860: Migrate `indicators/supertrend` to TypeScript

activist:
Closes #11860, Tools: Migrated SupertrendIndicator to TS.

### DIFF
--- a/ts/indicators/supertrend.src.ts
+++ b/ts/indicators/supertrend.src.ts
@@ -1,0 +1,651 @@
+/* *
+ *
+ *  License: www.highcharts.com/license
+ *
+ *  !!!!!!! SOURCE GETS TRANSPILED BY TYPESCRIPT. EDIT TS FILE ONLY. !!!!!!!
+ *
+ * */
+
+'use strict';
+
+import H from '../parts/Globals.js';
+
+/**
+ * Internal types
+ * @private
+ */
+declare global {
+    namespace Highcharts {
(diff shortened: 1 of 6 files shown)


ts/indicators/supertrend.src.ts
```diff
@@ -0,0 +1,651 @@
...
+
+        interface SupertrendGroupedPointsObject {
+            bottom: Array<Partial<Point>>;
+            intersect: Array<Partial<Point>>;
+            top: Array<Partial<Point>>;
+        }
+
+        interface SupertrendLineOptionsObject {
+            [index: string]: (Dictionary<CSSObject>|undefined);
+        }
```
bre1470: As far as I see, this interface is not part of the API options and therefore should be called something ending on `Object`, but not `OptionsObject` or `Options`. Additionally I am not sure, if this interface has any benefit, as it is only used once and can be replaced by `Dictionary<(Dictionary<CSSObject>|undefined)>`.

activist: Ok, will fix the naming :+1: 

Regarding the interface I'm too was unsure, but I felt that putting it in a interface is slightly clearer as (at least to me) `Dictionary<(Dictionary<CSSObject>|undefined)>` is somewhat harder to grasp.

activist:
> Please resolve conflicts.

Fixed

PR #12197, opened 2019-10-11
bugfix/12182-dtl-format-timezone-tooltip

### LINKED ISSUE #12182: Custom dateTimeLabelFormat only shows time when using custom timezone
#### Expected behaviour
Date formatter should use correct custom dateTimeFormat format when setting specific timezone or offset from UTC.

#### Actual behaviour
Date formatter only shows the "minute" formatting, even when showing historical data

https://jsfiddle.net/yrja26ez/

- Only showing HH:MM in tooltip
- Showing full date and time when removing timezone

#### Product version
Highstock 7.2.0

#### Affected browser(s)
Tested in chrome

pawelfus:
Fixed #12182, wrong Tooltip's `dateTimeLabelFormat` was picked when timezones were used.

### DIFF
--- a/js/parts/Tooltip.js
+++ b/js/parts/Tooltip.js
@@ -1028,7 +1028,7 @@ H.Tooltip.prototype = {
      *         The optimal date format for a point.
      */
     getDateFormat: function (range, date, startOfWeek, dateTimeLabelFormats) {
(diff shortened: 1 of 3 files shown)


TorsteinHonsi:
Closing, turns out this is not a bug.

PR #12196, opened 2019-10-11
ts/indicators-tema

activist:
Closes #11875, Tools: Migrated TEMAIndicator to TS.

PR #12198, opened 2019-10-11
ts/indicators-trix

### LINKED ISSUE #11876: Migrate `indicator/trix` to TypeScript

activist:
Closes #11876, Tools: Migrated TRIXIndicator to TypeScript.

### DIFF
--- a/ts/indicators/trix.src.ts
+++ b/ts/indicators/trix.src.ts
@@ -1,0 +1,142 @@
+/* *
+ *
+ *  License: www.highcharts.com/license
+ *
+ *  !!!!!!! SOURCE GETS TRANSPILED BY TYPESCRIPT. EDIT TS FILE ONLY. !!!!!!!
+ *
+ * */
+
+'use strict';
+
+import H from '../parts/Globals.js';
+
+/**
+ * Internal types
+ * @private
+ */
+declare global {
+    namespace Highcharts {
+        class TRIXIndicator extends TEMAIndicator {
+            public data: Array<TRIXIndicatorPoint>;
+            public init(): void;
+            public getTemaPoint(
+                xVal: Array<number>,
+                tripledPeriod: number,
+                EMAlevels: EMAIndicatorLevelsObject,
+                i: number
+            ): ([number, (number|null)]|undefined);
+            public options: TRIXIndicatorOptions;
+            public pointClass: typeof TRIXIndicatorPoint;
+            public points: Array<TRIXIndicatorPoint>;
+        }
+
+        interface TRIXIndicatorParamsOptions
+            extends TEMAIndicatorParamsOptions {
+            // for inheritance
+        }
+
+        class TRIXIndicatorPoint extends TEMAIndicatorPoint {
+            public series: TRIXIndicator;
+        }
+
+        interface TRIXIndicatorOptions extends TEMAIndicatorOptions {
+            params?: TRIXIndicatorParamsOptions;
+        }
+
+        interface SeriesTypesDictionary {
+            trix: typeof TRIXIndicator;
+        }
+    }
+}
+
+import '../parts/Utilities.js';
+import requiredIndicator from '../mixins/indicator-required.js';
+
+var correctFloat = H.correctFloat,
+    TEMA = H.seriesTypes.tema;
+
+/**
+ * The TRIX series type.
+ *
+ * @private
+ * @class
+ * @name Highcharts.seriesTypes.trix
+ *
+ * @augments Highcharts.Series
+ */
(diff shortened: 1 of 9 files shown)


activist:
Awaiting review and merge of #12196 before further work.

ts/indicators/trix.src.ts
```diff
@@ -0,0 +1,142 @@
...
+                var TRIXPoint: ([number, (number|null)]|undefined) = [
+                    xVal[i - 3],
+                    EMAlevels.prevLevel3 !== 0 ?
+                        correctFloat(EMAlevels.level3 - EMAlevels.prevLevel3) /
+                        EMAlevels.prevLevel3 * 100 : null
+                ];
+            }
+
+            return TRIXPoint;
+        }
```
activist: So, because this can return null there are changes also in several other indicators as the parent indicator also must be able to support this.

bre1470: These are clearly design issues of indicators. It is probably just working, because the functionality is so simple. Maybe we should consider an `interface` for all indicators and ignore the `SMAIndicator` inheritence in the code. The Interface can then become generic to support whatever values an indicator is returning.
```ts
interface Indicator<TValues extends IndicatorValuesType> {
    getValues(series: Series, options: IndicatorParamsOptions): TValues;
}
type IndicatorValuesType = …
```
... or we integrate this interface into `SMAIndicator` to continue the current inheritence:
```ts
interface SMAIndicator<TValues extends IndicatorValuesType> {
    getValues(series: Series, options: SMAIndicatorParamsOptions): TValues;
}
type IndicatorValuesType = …
```
What do you think?

pawelfus: It works because returned object by `getValues()` is the same format as `series.xData` and `series.yData`:
```js

            public xData?: Array<number>;
            public yData?: (
                Array<(number|null|undefined)>|
                Array<Array<(number|null|undefined)>>
            );
```
And `TRIXPoint` is just `[xData[i], yData[i]]` => `[number, (number|null)]` or `undefined`.

PS: `series.xData` can contain `null`'s too

activist: @bre1470  I agree that we need to refactor this, but I'm not sure it will be for the better if we are not allowed to change any implementation code. Maybe we can discuss it next week?

bre1470: @pawelfus - So the indicator's xData and yData are always the same as in the reference series?

@activist - Yes, we can do that.

pawelfus: I'm not sure what do you mean by reference series, but indicator is a series. So `xData` and `yData` in series and indicators should be the same type

TorsteinHonsi:
Awaiting @bre1470 's review.

bre1470:
We will discuss this today.

ts/indicators/trix.src.ts
bre1470: So the outcome of our discussion is as following:

- [ ] Introduce a generic values object:
```ts
        interface IndicatorValuesObject<TLinkedSeries extends LineSeries> {
            values: Array<[
                Extract<Required<TLinkedSeries['xData']>, '0'>,
                Extract<Required<TLinkedSeries['yData']>, '0'>
            ]>;
            xData: TLinkedSeries['xData'];
            yData: TLinkedSeries['yData'];
        }
```

- [ ] Make indicators generic:
```ts
class SMAIndicator<TLinkedSeries extends LineSeries> extends LineSeries {
    public data: Array<SMAIndicatorPoint<TLinkedSeries>>;
    public linkedParent: TLinkedSeries;
    public processData: TLinkedSeries['processData'];
    public points: Array<SMAIndicatorPoint<TLinkedSeries>>;
    public getValues(
        series: TLinkedSeries,
        params: SMAIndicatorParamsOptions
    ): (boolean|IndicatorValuesObject<TLinkedSeries>);
}
class SMAIndicatorPoint<TLinkedSeries extends LineSeries> extends LinePoint {
    public series: SMAIndicator<TLinkedSeries>;
}
```
```ts
class IKHIndicator<TLinkedSeries extends OHLCSeries> extends SMAIndicator<TLinkedSeries> {
    public getValues(
        series: TLinkedSeries,
        params: IKHIndicatorParamsOptions
    ): (boolean|IndicatorValuesObject<TLinkedSeries>);
}
```

- In an other pull request replace `boolean` returns of `getValues` with `undefined`.

activist:
@bre1470  Will you approve this? Then I will create two other PRs. One for making indicators generic and one for getting rid of the boolean return type from `getValues(..)`

Issue #11829, opened 2019-09-02
Treemap - update issue for point's with declared parent


Basts:
The update feature doesn't work for points with declared parent property. 

Last working version is 6. 

#### Live demo with steps to reproduce

[Demo](https://jsfiddle.net/BlackLabel/mvbwjroc/)

#### Product version

versions > 6

pawelfus:
Duplicate of #11738

PR #12199, opened 2019-10-11
bugfix/11738-treemap-update

### LINKED ISSUE #11738: Treemap - update issue when same amount of points
Update feature doesn't work in the case of updating the same amount of points as the number of points for initial series.

Last working version is 7.0.3 - [demo](https://jsfiddle.net/BlackLabel/pxdjk4yo/)

[Demo](https://jsfiddle.net/BlackLabel/8ux9j7d4/)

#### Product version
versions > 7.0.3

pawelfus:
Fixed #11738, updating treemap series with the same amount of points caused cells to disappear.

### DIFF
--- a/js/modules/treemap.src.js
+++ b/js/modules/treemap.src.js
@@ -1163,6 +1163,7 @@ seriesType('treemap', 'scatter'
                         zIndex: 1000 - levelDynamic
                     })
                         .add(series.group);
+                    series[groupKey].survive = true;
                 }
             }
             // Draw the point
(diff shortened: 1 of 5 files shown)


pawelfus:
Hmm.. strange, in the first visual test I see no difference (maybe because of the gif's quality), while the second one has different `Zoom` text rendering - not related to the commit

lezan:
I do not know if I am missing something, but I am trying to use master branch to fix the problem mentioned here #11829 but still It can not update correctly in my case.
[Here](https://www.highcharts.com/forum/viewtopic.php?f=9&t=42574&p=149765#p149765) there is conversation with me and Bastss.
You can find a jsfiddle here: https://jsfiddle.net/e8uda9gw/.

Thank you.

pawelfus:
Thanks! I will check this

pawelfus:
Created separate ticket for this issue with ID's: https://github.com/highcharts/highcharts/issues/12333

PR #12229, opened 2019-10-16
enhancement/exported numberFormat

marvin19:


pawelfus:
> A better approach would be a new callback option, global.numberFormatter. Then if that callback exists, the custom formatter will run. It will also be able to call the original Highcharts.numberFormatter.

This is good one. Global formatter is a good start, if needed we can always move it to local config (like we did with timezones).

jon-a-nygaard:
I agree with @pawelfus's comment 👍 

We can maybe also, instead of having to check if the `global.numberFormatter` exists, have it default to `Highcharts.numberFormat`, and use `global` everywhere `numberFormat` is called today?

TorsteinHonsi:
> We can maybe also, instead of having to check if the global.numberFormatter exists, have it default to Highcharts.numberFormat, and use global everywhere numberFormat is called today?

Yes that may be worth checking out. Then instead of `var numberFormatter = U.numberFormatter`, the modules will use `var numberFormat = Highcharts.defaultOptions.global.numberFormat`.

TorsteinHonsi:
Conclusion after discussion:
* We add a new option, `chart.numberFormatter` to allow setting individual number formats per chart.
* A `numberFormatter` property is set on the `Chart` instance (next to `this.time`), then this function is used in place of the old `H.numberFormat`.
* The `format` and the `formatSingle` functions are changed to take `chart` as the third argument instead of `time`, then `time` and `numberFormatter` are read from the `chart`.

marvin19:
samples/unit-tests/datalabels/members/ there is a bug here that @jon-a-nygaard is going to fix.

PR #12237, opened 2019-10-17
Bugfix/a11y minor fixes

oysteinmoseng:
Improved handling of null point accessibility.
___
- Always describe the series element if `exposeAsGroupOnly` is true.
- Describe null points to AT by creating invisible dummy elements.

Some changes to existing null-related unit tests to disable accessibility. Otherwise only a11y changes.


oysteinmoseng:
Forced to merge master in due to tooling issues, so apologize for the commit history mess.

PR #12241, opened 2019-10-17
docs/9892-module-requirements

bre1470:
- [x] Added bunch of requires tags in options.
- [x] Added more requires tags in options.


js/annotations/annotations.src.js
```diff
@@ -238,6 +239,8 @@ merge(
              * Options for annotation's labels. Each label inherits options
              * from the labelOptions object. An option from the labelOptions
              * can be overwritten by config for a specific label.
+             *
+             * @requires modules/annotations
```
TorsteinHonsi: If the `labelOptions` have a `@required` tag, its brother `shapeOptions` should also have it.

bre1470: Please review once more.

PR #12253, opened 2019-10-21
docs/error-17

### LINKED ISSUE #9892: Enhancement suggestion: improve error #17 (missing series)
#### Expected behaviour
Error with more details.

1 . What is the name of the missing series - in console

2 . What file should be loaded to avoid the error - on the error site, as 
**full table with all series and all related required files** visible to a user **or** error site with additional request about the missing series type, so 
**only the relevant info decided through the GET request** will be shown e.g. `https://www.highcharts.com/errors/17/?type=scatter` and an error about possible typo in case no match.

#### Actual behaviour
Error message not precise enough to help resolve the problem.

bre1470:
- Close #9892
- Added H.error hint in URL of error 17

### DIFF
--- a/js/parts/Chart.js
+++ b/js/parts/Chart.js
@@ -347,7 +347,7 @@ extend(Chart.prototype, /** @lends Highcharts.Chart.prototype */ {
             optionsChart.defaultSeriesType), series, Constr = seriesTypes[type];
         // No such series type
         if (!Constr) {
-            H.error(17, true, chart);
+            H.error(17, true, chart, { missingModuleFor: type });
         }
         series = new Constr();
         series.init(this, options);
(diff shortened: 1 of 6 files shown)


TorsteinHonsi:
I know this is still a WIP, but instead of having bulky code inside the error handler that deals with error-specific stuff, I think it is better that this information is passed over with the actual call. If we introduce a new parameter for extra information (`params` or the like), the error can be called with something like this:

```js
        // No such series type
        if (!Constr) {
            H.error(17, true, chart, { type });
        }
```

The params can be passed directly to the website via a querystring if we want that.

bre1470:
Okay, I will do that. I was not sure, if I should introduce a new parameter, when it is just in use for one error, but maybe there are more in the future.

TorsteinHonsi:
Yes, by allowing a object we can extend it with error-specific properties in the future. It may be that we should just pass `{ msg }` though, then we don't need any logic downstream.

PR #12254, opened 2019-10-21
bugfix/11471-chart-update-error

### LINKED ISSUE #11471: Chart.update fails when a new series with a previously unknown id is provided, and oneToOne is set
#### Expected behaviour
When I use Chart.update with oneToOne set to true, and I add a new series with a new ID, update should not raise an exception.

#### Actual behaviour
Highcharts raises:
It tries to find the series object by calling `Chart.get` on the id, and when a series with a corresponding id is not found, it tries to check each point in `series.points` for each `series` in `Chart.series`. However, at this point, all the points are actually null values, so you end up calling `null.id`, which raises an exception. 

This is the console error:
```
Uncaught TypeError: Cannot read property 'id' of null
    at itemById (highstock.src.js:24127)
    at Array.find (<anonymous>)
    at H.find (highstock.src.js:1929)
    at H.Chart.get (highstock.src.js:24136)
    at highstock.src.js:33374
    at highstock.src.js:33371
    at H.Chart.update (highstock.src.js:33353)
    at (index):50
```

https://jsfiddle.net/he7f43kt/3/
If you remove "id" of the "eurusd" series in the update call, it works.

#### Product version
Highstock 7.2.0

#### Affected browser(s)
Chrome

pawelfus:
Fixed #11471, updating chart with new series and `oneToOne` set to true used to throw errors when data was grouped.
___
Note it's a tricky solution, perhaps it's better to implement the workaround from the ticket instead. But from the perf perspective, it's better this way..


pawelfus:
No idea about pictorial demo, I guess it's a new one and my master branch wasn't up-to-date?

PR #12259, opened 2019-10-23
docs/12250-point-index

### LINKED ISSUE #12250: Typescript type for "Point" does not contain "index" property
Given a populated "TooltipFormatterContextObject" called "context", typescript complains when accessing `context.point.index` property.

To access the property while using typescript, you must cast `context.point` to `any`.

`const index = (context.point as any).index`

Can we get the type definitions updated please?

N/A

#### Product version
Highcharts 7.2.0

#### Affected browser(s)
Tested using Chrome

bre1470:
- Close #12250, Made `Point.index` public.
- Made `Series.index` public.


bre1470:
> * Both descriptions start with "Returns ...". This sounds like a function that returns a value, while this is rather a numeric property. Maybe we should skip this word?
> * Related to the above, the `index` properties are internal properties that cannot be set directly, only read. Should we add the `@readonly` tag?

In pure TypeScript we would implement this as a get function, that only returns the value. See https://www.typescriptlang.org/docs/handbook/classes.html#accessors
I will add the `@readonly` tag in the meanwhile.

> * Both properties are typed `number|undefined`. In what cases can they be undefined?

I believe that is a mistake. I probably misread the conditions.

PR #12280, opened 2019-10-28
Samples: Patched details files Gantt

goransle:
Patched `demo.details` files for Gantt demos and samples

### DIFF
--- a/samples/gantt/chart/scrollable-plotarea-vertical/demo.details
+++ b/samples/gantt/chart/scrollable-plotarea-vertical/demo.details
@@ -1,6 +1,8 @@
 ---
- name: Gantt Chart with Scrollable Plot Area
- js_wrap: b
- authors:
-   - Torstein Hønsi
+name: Gantt Chart with Scrollable Plot Area
+js_wrap: b
+authors:
+  - Torstein Hønsi
+tags: []
+categories: []
 ...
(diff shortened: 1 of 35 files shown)


oysteinmoseng:
Agree. It should currently be easy to parse this from the HTML source, unless I am missing something. I think this would be a better option than forcing duplication, since alternative 2 would require us to parse the HTML anyway.

goransle:
Should have been clearer in the original message. The new additions are all parsed via the demo-manager, so keeping the details file as it is should not be problem in that regard.

jon-a-nygaard:
I am wondering what is the purpose or benefit of having the name of these files in the demo.details? Is it to be able to filter on the modules used in the demo?

Another possible option could be to remove the scripts from `demo.html` and add them as `resources` to JSFiddle and Codepen, and have the demo manager add all resources as link/script tags in the output.

TorsteinHonsi:
> The new additions are all parsed via the demo-manager, so keeping the details file as it is should not be problem in that regard.

My concern is about keeping it consistent in the future. We constantly add and update samples.

> Another possible option could be to remove the scripts from demo.html and add them as resources to JSFiddle and Codepen, and have the demo manager add all resources as link/script tags in the output.
> downside is that the script tags are not visible in the html

Yes, we tried this in the beginning. The main downside is that is very cumbersome to add/modify/remove scripts when testing in jsFiddle.

jon-a-nygaard:
> Having it in demo.html is more visible and much more handy

I agree, so if it is not very beneficial that the modules are mentioned in the `demo.details`, then I believe alternative 1 is a good choice.

goransle:
Changed to only patch names, tags and categories. Currently names are set using the `demo/index.htm` file in all cases, but perhaps it should prefer the name in `demo.details` (if it is not 'Highcharts Demo' etc)?

PR #12285, opened 2019-10-28
Bugfix/regression windbarb datagrouping

TorsteinHonsi:
Data grouping failed if applied to the first wind barb series on the page.

Fail case: http://utils.highcharts.local/samples/#test/highcharts/plotoptions/windbarb-datagrouping


TorsteinHonsi:
Well, the comment explains it, it is actually a module load order problem. If the data grouping module is loaded before the windbarb, the approximation is registered immediately, otherwise it is registered on series init.

PR #12297, opened 2019-10-29
Feature/basic demo a11y

oysteinmoseng:
Added new option `accessibility.linkedDescription`.
___
Added a11y to some demos (gone through alphabetically from A-G). More to come, but want to merge current work before the grand Typescript migration.

Some a11y bugs have been fixed in the process, and a new option `accessibility.linkedDescription` has been added.

No non-a11y files modified except for demos.


js/modules/accessibility/options/options.js
```diff
@@ -215,6 +218,31 @@ var options = {
          * @apioption accessibility.description
          */
 
+        /**
+         * The ID of an HTML element describing the contents of the chart.
```
TorsteinHonsi: Using id's doesn't scale well. If we have 10 charts on the page, it would  be preferable to avoid using id's, but rather rely on the HTML structure. For example, default to the next sibling of the container that has a `.highcharts-caption` tag. Maybe we should consider that the default `linkedDescription` be a query selector, but in that case it needs to be relative to the `container`'s parent node, since we can't currently select the parent directly.

oysteinmoseng: The caption texts defined in JS are using the `caption` feature, while the ones in HTML are changed to use the `linkedDescription`. I am thinking for the rest of the demos we might want to only use the `linkedDescription` since we consider it better practice. Thoughts on that?

As for IDs, I am not sure if it is safe to make it relative to the parent node of the container. I can imagine users using a CMS where the description is somewhere else in the DOM. Or maybe we can make it relative to the parent node, and also accept an HTML element as parameter - allowing users to make their own fancy queries. E.g:

```javascript
accessibility: {
    linkedDescription: '.highcharts-caption',
    linkedDescription: document.getElementById('myIDForElementSomewhereFarAway')
}
```

Thoughts?

oysteinmoseng:
Again CI issues, so had to merge in master (conflicts, so no rebase). Apologies for review mess 😬

TorsteinHonsi:
See my [comment 5 days ago](https://github.com/highcharts/highcharts/pull/12297#pullrequestreview-309012562).

PR #12300, opened 2019-10-30
Added functionality for showing previous price in Stockchart tooltip

eldjuve:
I did something like this using wrap on one of my implementations, and really liked the effect.

Basically when showing compares on stockcharts I want to always show a valid tick for all series that has data/ticks before the hover point. I have customers that have been thinking it is a bug when the shared tooltip falls out mid chart when there are gaps in the data for a less volatile symbol, but this would make it very clear.

Not been able to compile and test(have to set env variable for python and stuff) but hopefully it compiles. Data manipulation in demo to show data with gaps.


eldjuve:
Demo of our current implementation:
chart

PR #12310, opened 2019-10-30
bugfix/12248-waterfall-overlap-stacking 

### LINKED ISSUE #12248: waterfall with overlap stacking misaligned after reflow
I'm trying to make a waterfall chart using `stacking: "overlap"` with multiple stacks. If the container has explicitly set width and height, the chart draws as expected, but as soon as anything causes a `reflow`, the first stack is misaligned. Further reflows compound the misalignment. A reflow can be triggered by changing the width or height of the container and then explicitly running `reflow()` on the chart, as in the example below, however, a reflow can/will be caused by other events as well, for instance...
- if you click on one of the series in the legend to hide it and then click again to reshow it
- if the container's height and/or width is not explicitly set, the browser causes and immediate reflow on load, so that the chart will never plot correctly
- etc.

#### Expected behaviour
after reflow, chart should look the same (relatively speaking, if the container size has changed)

#### Actual behaviour
after reflow, chart looks different

https://jsfiddle.net/Loqpta41/

#### Product version
Highcharts JS v7.2.0 (2019-09-03)
https://code.highcharts.com/highcharts.js
https://code.highcharts.com/highcharts-more.js

#### Affected browser(s)
on macOS 10.15
Google Chrome Version 77.0.3865.120 (Official Build) (64-bit)
Safari Version 13.0.2 (15608.2.30.1.1)
Firefox Quantum 69.0.3 (64-bit)

PaulDalek:
Fixed #12248, the Y axis' range was disturbed on stacked (overlapping) waterfall chart.

### DIFF
--- a/js/parts-more/WaterfallSeries.js
+++ b/js/parts-more/WaterfallSeries.js
@@ -33,6 +33,10 @@ addEvent(Axis, 'afterInit', function () {
         };
     }
 });
+addEvent(Axis, 'afterBuildStacks', function () {
+    this.waterfallStacks.changed = false;
+    delete this.waterfallStacks.alreadyChanged;
+});
 addEvent(Chart, 'beforeRedraw', function () {
     var axes = this.axes, series = this.series, i = series.length;
     while (i--) {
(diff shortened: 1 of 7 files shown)


PaulDalek:
Thanks!

PR #12311, opened 2019-10-30
bugfix/12267-getCrispPath-error

### LINKED ISSUE #12267: Waterfall chart error when invisble and first value below 0
#### Expected behaviour
Create a waterfall serie that is invisible by default.

#### Actual behaviour
Highcharts-more crashes in the method `getCrispPath`.

When specifying `visible: false` and a value below 0 in the series configuration it crashes: https://jsfiddle.net/2q3Loz5h/
Otherwise it renders correctly and can be set visible or invisible: https://jsfiddle.net/2q3Loz5h/1/

#### Product version
Last versions of Higcharts and Highcharts-more

#### Affected browser(s)
Reproduced on Chrome and Firefox

PaulDalek:
Fixed #12267, error in waterfall chart when the series was invisible and the first value was below 0

### DIFF
--- a/js/parts-more/WaterfallSeries.js
+++ b/js/parts-more/WaterfallSeries.js
@@ -455,7 +455,7 @@ seriesType('waterfall', 'column', {
                     yPos
                 ];
             }
-            if (!stacking &&
+            if (!stacking && d &&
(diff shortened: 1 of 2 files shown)


PaulDalek:
Thanks!

PR #12321, opened 2019-11-02
docs/data-labels-structure

### LINKED ISSUE #10401: Funnel datalabels left align
#### Expected behaviour
How do I keep all the data labels in funnel on left side ?

#### Actual behaviour
All data labels are coming on right side.  I kept datalabels.x and kept -450. The items are coming on left side but alignments are messed up. And the softconnector are on right side.

https://www.highcharts.com/demo/funnel

#### Product version
highcharts 6.0.4

#### Affected browser(s)
Chrome, Firefox

### LINKED ISSUE #11803: Option (series.)dataLabels.enabled not mentioned in documentation

### LINKED ISSUE #12185: connectorWidth / connectorColor are missing in docs
#### Expected behaviour
add options to datalabels like it was in the past.

#### Actual behaviour
The connectorWidth / connectorColor are missing in docs

#### Product version
7.2.0

@bre1470

### LINKED ISSUE #12194: Doc of sankey need update
#### Expected behaviour
https://github.com/highcharts/highcharts/blob/0017f6c643/js/modules/sankey.src.js#L76

#### Actual behaviour
The nodeFormat and nodeFormatter are not listed in the latest document.
image

bre1470:
- Close #10401, Added position option to dataLabels.
- Close #11803, Restored enabled option of dataLabels.
- Close #12185, Restored connectorColor and connectorWidth options of dataLabels.
- Close #12194, Made doclet types and structure consistent.


bre1470:
It is the only way to get dataLabels fail proof consistent with minimum overload.
- If we have default values as expandable leafs without documentation it looks like a bug (see issue reports).
- If we add documentation a second time beside the interface we end up with out of sync doclets like in timeline.
- If we remove the common dataLabels interface, the declarations will grow by about 20% as every series gets its own dataLabels interface.

bre1470:
> The inheritance structure of the `dataLabel` is probably more complex, but to me this feels like an inconsistency.

If you are worried about navigation inconsistency, I can migrate `dragDrop` and `marker` to interfaces as well. There is no need for 100+ different ones imho. They are just bloating the declarations.

bre1470:
Conclusions:
- This gets reverted to the original structure inside the default values (like in v6).
- A new tag is introduced for the declarations generator to mark shared interfaces.
- Optimization for other `OptionsObject` interfaces will be done after all source code is migrated to TS files.

PR #12324, opened 2019-11-04
Bugfix/11617 scrollable split tooltip

### LINKED ISSUE #11617: split tooltip and scrollbalePlotArea
#### Expected behaviour
Tooltip with full opacity

#### Actual behaviour
Tooltip hidden behind the scrollable mask

https://jsfiddle.net/BlackLabel/jbk9qpnx/

Scroll down to the xAxis and hover any point.

Screenshot:

Screen Shot 2019-08-06 at 14 56 35

#### Product version
v7.1.2

jon-a-nygaard:
Fixed #11617, support `tooltip.split` with `chart.scrollablePlotArea`.

---
# Description
Added support for split tooltip with scrollable plot area.

- Limit the tooltip to render within the plot area when `scrollablePlotArea` is enabled.
- Updated `getChartPosition` to use `scrollingContainer` when it exists, to position tooltip correctly when `container` is moved during scrolling. [See #diff-31a7e6aaf7bdf7daa73993cea91fd9efR375]( https://github.com/highcharts/highcharts/compare/bugfix/11617-scrollable-split-tooltip?expand=1#diff-31a7e6aaf7bdf7daa73993cea91fd9efR375)
- Added utility function `clamp`
- Refactored to `renderSplit` to gather logic in smaller sections, added scoped functions `getAnchor`, `defaultPositioner`, and `updateTooltip`. 
- Removed `any` casts in `renderSplit` where possible.

# Example(s)
- [Demo of issue](https://jsfiddle.net/BlackLabel/jbk9qpnx/)
- [Demo of issue with fix applied](https://jsfiddle.net/jon_a_nygaard/q5dhzv2y/)
- [Demo of #11493 with current changes](https://jsfiddle.net/jon_a_nygaard/w3cq1Lt5/) - Manually tested against regressions.
- [Demo of #9590 with current changes](http://jsfiddle.net/jon_a_nygaard/w4L7brqs/) - Manually tested against regressions.

# Related issue(s)
- Closes #11617

### DIFF
--- a/js/parts/Tooltip.js
+++ b/js/parts/Tooltip.js
@@ -846,8 +846,6 @@ H.Tooltip.prototype = {
         var clamp = function (value, min, max) {
             return value > min ? value < max ? value : max : min;
         };
-        var clampToHorizontalPlotArea = function (value) { return clamp(value, plotLeft, plotLeft + plotWidth - scrollablePixelsX); };
-        var clampToVerticalPlotArea = function (value) { return clamp(value, plotTop, plotTop + plotHeight - scrollablePixelsY); };
         var boundaries = {
             left: scrollablePixelsX ? plotLeft : 0,
             right: scrollablePixelsX ?
@@ -952,8 +950,8 @@ H.Tooltip.prototype = {
                         - scrollTop;
                 }
                 // Limit values to plot area
-                anchorX = clampToHorizontalPlotArea(anchorX);
-                anchorY = clampToVerticalPlotArea(anchorY);
+                anchorX = clamp(anchorX, boundaries.left, boundaries.right);
+                anchorY = clamp(anchorY, boundaries.top, boundaries.bottom);
                 if (isHeader) {
                     x = clamp(x, boundaries.left, boundaries.right - boxWidth);
                 }
--- a/js/parts/Tooltip.js
+++ b/js/parts/Tooltip.js
@@ -1044,13 +1044,13 @@ H.Tooltip.prototype = {
          */
         var container = tooltip.container, outside = tooltip.outside, renderer = tooltip.renderer;
         if (outside && container && renderer) {
+            // Set container size to fit the tooltip
+            var _h = tooltipLabel.getBBox(), width = _h.width, height = _h.height, x = _h.x, y = _h.y;
+            renderer.setSize(width + x, height + y, false);
             // Position the tooltip container to the chart container
             var chartPosition = pointer.getChartPosition();
             container.style.left = chartPosition.left + 'px';
             container.style.top = chartPosition.top + 'px';
-            // Set container size to fit the tooltip
-            var _h = tooltipLabel.getBBox(), width = _h.width, height = _h.height, x = _h.x, y = _h.y;
-            renderer.setSize(width + x, height + y, false);
         }
     },
     /**
--- a/js/parts/Tooltip.js
+++ b/js/parts/Tooltip.js
@@ -856,13 +856,8 @@ H.Tooltip.prototype = {
         };
         var tooltipLabel = tooltip.getLabel();
         var headerTop = Boolean(opposite);
-        var boxes = [];
         var distributionBoxTop = plotTop;
         var headerHeight = 0;
-        // Graceful degradation for legacy formatters
-        if (isString(labels)) {
-            labels = [false, labels];
-        }
         // Calculate the x and y position for the anchor
         function getAnchor(point) {
             var isHeader = point.isHeader, _a = point.plotX, plotX = _a === void 0 ? 0 : _a, _b = point.plotY, plotY = _b === void 0 ? 0 : _b, series = point.series;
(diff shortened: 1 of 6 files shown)


ts/parts/Utilities.ts
```diff
@@ -1314,6 +1314,19 @@ H.merge = function<T> (): T {
...
+/**
+ * Constrain a value to within a lower and upper threshold.
+ *
+ * @private
+ * @param {number} value The initial value
+ * @param {number} min The lower threshold
+ * @param {number} max The upper threshold
+ * @return {number} Returns a number value within min and max.
+ */
+function clamp(value: number, min: number, max: number): number {
```
TorsteinHonsi: This is a quite frequent operation in Highcharts. Now that we have a utility function for that, we should at the same time identify all other places where it is done, and implement the utility function. Search for places with nested `Math.min` and `Math.max` calls.

js/parts/Tooltip.js
```diff
@@ -8,9 +8,25 @@
  *
  * */
 'use strict';
+var __read = (this && this.__read) || function (o, n) {
```
TorsteinHonsi: See if we can avoid the code that makes TypeScript include this polyfill. Likely candidate is array destructuring.

ts/parts/Tooltip.ts
```diff
@@ -1188,202 +1189,281 @@ H.Tooltip.prototype = {
...
+                xAxis: [{ opposite }]
+            },
+            options,
+            options: {
+                distance = 16,
+                positioner
+            }
+        } = tooltip;
+
+        const boundaries = {
```
TorsteinHonsi: From reading this code it looks like in a scrolling chart the boundaries are the _plot area_, whereas it is the whole chart otherwise? Is that because of the masks? In that case, a comment here would help.

ts/parts/Tooltip.ts
```diff
@@ -1188,202 +1189,281 @@ H.Tooltip.prototype = {
...
+
+        const tooltipLabel = tooltip.getLabel();
+        const headerTop = Boolean(opposite);
+
+        let distributionBoxTop = plotTop;
+        let headerHeight = 0;
+        let rightAlign = false;
+
+        /**
+         * Calculates the anchor position for the tooltip
```
TorsteinHonsi: A bit confusing description - we are talking about a partial tooltip here, right? So the comment would be clearer if it read something like "Calculates the anchor position for the partial tooltip"

ts/parts/Tooltip.ts
```diff
@@ -1188,202 +1189,281 @@ H.Tooltip.prototype = {
...
+            }
+
+            // Limit values to plot area
+            anchorX = clamp(anchorX, boundaries.left, boundaries.right);
+            anchorY = clamp(anchorY, boundaries.top, boundaries.bottom);
+            return [anchorX, anchorY];
+        }
+
+        /**
+         * Calculates the position of the tooltip
```
TorsteinHonsi: As above, confusing whether this applies to the `Tooltip` instance (`this`) or a partial tooltip for the split tooltip.

ts/parts/Tooltip.ts
```diff
@@ -1188,202 +1189,281 @@ H.Tooltip.prototype = {
...
+                    x = boundaries.right - boxWidth;
+                }
+            }
+
+            // NOTE: y is relative to distributionBoxTop
+            return { x, y };
+        }
+
+        /**
+         * Updates the attributes and styling of the tooltip. Creates a new
```
TorsteinHonsi: Partial tooltip...

ts/parts/Tooltip.ts
```diff
@@ -1188,202 +1189,281 @@ H.Tooltip.prototype = {
...
+         * tooltip if it does not exists.
+         *
+         * @private
+         * @param {Highcharts.Tooltip|undefined} tooltip The tooltip to update
+         * @param {Highcharts.Point} point The point related to the tooltip
+         * @param {boolean|string} str The label for the tooltip
+         * @return {Highcharts.SVGElement} Returns the updated tooltip
+         */
+        function updateTooltip(
+            tooltip: (Highcharts.SVGElement|undefined),
```
TorsteinHonsi: Use `partialTooltip` to avoid confusion with the Tooltip instance.

ts/parts/Utilities.ts
TorsteinHonsi: Still an issue.

TorsteinHonsi:
To do: 
 - [ ] Use `tooltip.distance` instead of `tooltip.options.distance`.
 - [ ] Rename partial tooltips in code and comments.
 - [ ] Use `clamp` in other parts of the code where appropriate.

jon-a-nygaard:
@TorsteinHonsi Please see latest changes.

> Use clamp in other parts of the code where appropriate.

Yes, I agree, but if you do not mind, then I would like to apply this in a follow up PR.
This is to avoid potential visual diffs to blend in with the ones in this PR.

TorsteinHonsi:
> Yes, I agree, but if you do not mind, then I would like to apply this in a follow up PR

Ok, makes sense!

PR #12328, opened 2019-11-04
Tools: getValues() now returns undefined

activist:
..rather than boolean.

I have tried to keep the changes as light as possible for now.

### DIFF
--- a/ts/indicators/acceleration-bands.src.ts
+++ b/ts/indicators/acceleration-bands.src.ts
@@ -198,7 +198,7 @@ H.seriesType<Highcharts.ABandsIndicator>(
                 i: (number|undefined);
 
             if (yValLen < period) {
-                return undefined;
+                return;
             }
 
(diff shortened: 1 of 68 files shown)


ts/indicators/acceleration-bands.src.ts
```diff
@@ -198,7 +198,7 @@ H.seriesType<Highcharts.ABandsIndicator>(
                 i: (number|undefined);
 
             if (yValLen < period) {
-                return false;
+                return undefined;
```
bre1470: Do not explicitly return `undefined` - just return without anything. Actually a lint rule should catch this. This has to be changed in all indicators.

activist: Ok, I'll fix, but why is returning nothing preferred?

bre1470: In some situations you can redeclare `undefined` as a variable. So use it only as a type and not as a value.

PR #12338, opened 2019-11-06
ts/indicators-getvalues-generics

### LINKED ISSUE #12264: Migrate indicators to use generics
Should extend LineSeries and return appropriate generic structure from `getValues(..)` based on its parent LineSeries.

activist:
Initial work on making the indicators `getValues(..)` generic. This makes it possible to remove all the different IndicatorValues* interfaces.

In some cases there were used LinkedParent interfaces to narrow down the xData and yData types in the getValues interface. However in most cases this collided with the generic basetype definition and I could not find a way around it. 

Closes #12264


bre1470:
There are merge conflicts because of the activated `no-undefined` rule.

ts/indicators/indicators.src.ts
```diff
@@ -22,7 +22,7 @@ declare global {
             public data: Array<SMAIndicatorPoint>;
             public dataEventsToUnbind: Array<Function>;
             public hasDerivedData: boolean;
-            public linkedParent: Series;
+            public linkedParent: LineSeries;
```
pawelfus: I'm not sure about this one - does `LineSeries` cover data format from OHLCSeries? For example SMA indicator can work with all possible linked parents (Line, Range, OHLC/Candlestick).

activist: Ok, then it should probably still use `Series`. Then I assume we need to change the generic type to extend from `Series`and not `LineSeries`as well?

pawelfus: I think so. Any thoughts @bre1470 ?

bre1470: Practically both should be the same, but in type sense it might be better to rather use `Series` than `LineSeries`.

activist: :+1: Ok, I'll change it

activist: Changed to Series. FYI @pawelfus @bre1470

pawelfus: Thanks!

PR #12348, opened 2019-11-06
bugfix/11148-tracking-outside-plotarea-polar

### LINKED ISSUE #11148: Trackable point outside yAxis extremes (polar)
#### Expected behaviour
No points can be tracked if their value is higher than yAxis.max

#### Actual behaviour
Although the points' y is bigger than yAxis.max, they can be tracked as long as they are within the plot area.

polar

https://jsfiddle.net/BlackLabel/goL7b1r6

#### Product version
Highcharts 7.1.2

#### Affected browser(s)
All

raf18seb:
Fixed #11148, a point should not be trackable when is out of his pane. Also, isInsidePlot() works for panes now.


ts/parts/Chart.ts
```diff
@@ -697,13 +697,29 @@ extend(Chart.prototype, /** @lends Highcharts.Chart.prototype */ {
...
     ): boolean {
-        var x = inverted ? plotY : plotX,
-            y = inverted ? plotX : plotY;
+        var chart = this,
+            x = inverted ? plotY : plotX,
+            y = inverted ? plotX : plotY,
+            center,
+            isInsidePlot;
+
+        if (chart.polar) {
```
TorsteinHonsi: The core Highcharts functionality shouldn't deal with polar chart functionality. Instead, all this functionality should be gathered in the Polar.ts file. You can resolve that by implementing an event pattern like seen with the `getAxes` and `afterDrawChartBox` events in Polar.ts. In order to modify a return value, the return value must be part of event arguments, something like this:

```js
    isInsidePlot: function (
        this: Highcharts.Chart,
        plotX: number,
        plotY: number,
        inverted?: boolean
    ): boolean {
        var x = inverted ? plotY : plotX,
            y = inverted ? plotX : plotY;

        var e = {
            isInsidePlot: x >= 0 &&
                x <= this.plotWidth &&
                y >= 0 &&
                y <= this.plotHeight
        };
        // A hook for an event listener can mutate e.isInsidePlot
        fireEvent(this, 'isInsidePlot', e);

        return e.isInsidePlot;
    },
```

ts/parts/Pointer.ts
```diff
@@ -637,16 +637,53 @@ Highcharts.Pointer.prototype = {
...
         e?: Highcharts.PointerEventObject
     ): Highcharts.PointerHoverDataObject {
-        var hoverPoint: Highcharts.Point,
+        var chart = this.chart,
+            hoverPoint: Highcharts.Point,
             hoverPoints = [] as Array<Highcharts.Point>,
             hoverSeries = existingHoverSeries,
             useExisting = !!(isDirectTouch && existingHoverPoint),
             notSticky = hoverSeries && !hoverSeries.stickyTracking,
+            isInsidePane = (
```
TorsteinHonsi: As above, we shouldn't pollute core Pointer stuff with polar features. Instead, write an event hook and extend it from the Polar.ts file.

PR #12349, opened 2019-11-06
bugfix/12298-crosshair-after-update

### LINKED ISSUE #12298: Two labels from left and right Y axis are displayed
#### Expected behaviour
In a case when user clicks on chart to enter "edit" mode (see jsfiddle for implementation) if there is disabled and "shared: false" tooltip it will display TWO data labels for right series (red color on example).

#### Actual behaviour
Should display only ONE data label per crosshair on the left/right depending on the series Y axis.

https://jsfiddle.net/fameriks/2pxvgs3o/50/

Steps to reproduce:

1. open jsfiddle
2. move cursor to any point from red series
3. hit mouse click AND don't move cursor
4. observe two labels on left and right Y axes

Looks like this setting affects it (even when tooltip is disabled) https://api.highcharts.com/highstock/tooltip.shared - One label when shared is "true". Two labels when shared is "false"

#### Product version
Highstock 7.2.0

#### Affected browser(s)
All

pawelfus:
Fixed #12298, enabling crosshair label while a point is hovered rendered crosshair labels for all axes.

### DIFF
--- a/ts/parts/Pointer.ts
+++ b/ts/parts/Pointer.ts
@@ -1002,7 +1002,10 @@ Highcharts.Pointer.prototype = {
                 } else if (hoverPoint) { // #2500
                     hoverPoint.setState(hoverPoint.state, true);
                     chart.axes.forEach(function (axis: Highcharts.Axis): void {
-                        if (axis.crosshair) {
+                        if (
+                            axis.crosshair &&
+                            (hoverPoint as any).series[axis.coll] === axis
+                        ) {
                             axis.drawCrosshair(null as any, hoverPoint);
                         }
                     });
--- a/js/parts/Pointer.js
+++ b/js/parts/Pointer.js
@@ -646,7 +646,8 @@ Highcharts.Pointer.prototype = {
                 else if (hoverPoint) { // #2500
                     hoverPoint.setState(hoverPoint.state, true);
                     chart.axes.forEach(function (axis) {
-                        if (axis.crosshair) {
+                        if (axis.crosshair &&
+                            hoverPoint.series[axis.coll] === axis) {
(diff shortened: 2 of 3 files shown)


ts/parts/Pointer.ts
```diff
@@ -1002,7 +1002,10 @@ Highcharts.Pointer.prototype = {
                 } else if (hoverPoint) { // #2500
                     hoverPoint.setState(hoverPoint.state, true);
                     chart.axes.forEach(function (axis: Highcharts.Axis): void {
-                        if (axis.crosshair) {
+                        if (
+                            axis.crosshair &&
+                            (hoverPoint as any).series[axis.coll] === axis
```
pawelfus: Added casting to any similar to this one:
https://github.com/highcharts/highcharts/blob/0db656018c3a976d7e8ac38e9ff4d4778256603a/ts/parts/Pointer.ts#L868-L870

The difference here is `(hoverPoint as any).series` vs `(hoverPoint.series as any)`. For some reason TS said that `hoverPoint` can be undefined, which clearly is checked above. Did I miss something?

bre1470: That is typical for TypeScript, if you use an outer variable in an inner function. TypeScript is not able to determ, when the inner function gets called and the outer variable might change. Therefor it has to assume, that every of the possible variable types can be possible in the inner function. So you have to do an other validate check.

That said, in this case a cast to any is still necessary, because you use hoverPoint with an indexer, which the Point class does not have. So a inner validation does not help in type sense.

bre1470: In few words: The any cast is fine as it is for now.

pawelfus: Thanks for the explanation!

PR #12351, opened 2019-11-06
ts/modules-sonification-fixes

bre1470:
- Removed internal sonification interface.
- Optimized sonification types.


bre1470:
It looks like I have to revert the branch to the last visual safe commit and reapply all changes. Maybe in several commits so we can see, where it breaks.

PR #12359, opened 2019-11-07
Tools: Updated CI cache key for now to force

activist:
..update. Will remove npm caching later if I cant find a fix for it.

This will fix issues with CI not using the last TS declaration generator version.

FYI @bre1470

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -79,13 +79,13 @@ jobs:
      - run: node --version && npm --version
      - restore_cache:
          keys:
-           - v1-npm-deps-{{ .Branch }}-{{ checksum "package.json" }}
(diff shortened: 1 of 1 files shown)


bre1470:
Thank you! 👍  I am not sure, if it is necessary anymore. The latest workflow cycles are not failing anymore.

bre1470:
Maybe it is enough to create the PR to flush the cache, and the merge is not necessary.

activist:
> Maybe it is enough to create the PR to flush the cache, and the merge is not necessary.
Unfortunately, it would need to be updated on the each branch where it has a stale cache. For new PRs it wouldn't be stale so it is likely not an issue for those.

PR #12364, opened 2019-11-07
bugfix/12145-series-color-option

### LINKED ISSUE #12145: Wrong 'Tooltip.pointFormatter' type definition after parsing to d.ts
#### Expected behaviour
Context should indicate the Point object, and all series properties should exist on `this.series` within formatter function.

#### Actual behaviour
The `Tooltip.pointFormatter` definition is not correct after parsing to `highcharts.d.ts` file.
Before it looks like:
```ts
pointFormatter?: FormatterCallbackFunction<Point>;
```
when the `FormatterCallbackFunction` is:
```ts
interface FormatterCallbackFunction<T> {
    (this: T): string;
}
```
but after parse it looks like not complete or overwritten by something else:
```ts
pointFormatter?: () => void;
```

This cause errors within custom formatter function, a few 'does not exist' warnings, and wrong context
image

bre1470:
Fixed #12145, missing option `supertrend.color`.


ts/indicators/supertrend.src.ts
```diff
@@ -281,14 +281,20 @@ H.seriesType<Highcharts.SupertrendIndicator>(
                     top: {
                         styles: {
                             lineWidth: indicOptions.lineWidth,
-                            lineColor: indicOptions.fallingTrendColor,
+                            lineColor: (
+                                indicOptions.color ||
+                                indicOptions.fallingTrendColor
```
pawelfus: This is general request everywhere:

Please first use detailed option, then general: `indicOptions.fallingTrendColor || indicOptions.color`. In this PR, `color` option can be used instead of both options (`fallingTrendColor` and `risingTrendColor`) making `color` option universal, so should be picked last.

bre1470: - The color options gets the inherited description.
- The reason for the order of `series.color` is in line with the current behaviour for `point.color`, which overrules both other color options. @TorsteinHonsi What do we do?

TorsteinHonsi: I think `point.color` should override any series color. But it doesn't look like the `point.color` is separated from the `series.color` in the code?

bre1470: The order of the PR is: `point.color` > `series.color` > `series.xTrendColor`.
The requested change is: `point.color` > `series.xTrendColor` > `series.color`.
So I assume, you agree with Pawel?

pawelfus: The main reason for this change is to get compatible TS (this is the only series that doesn't support `series.color`, instead implements `series.risingTrendColor` and `fallingTrendColor`).

Other options (e.g. `negativeColor`, `upColor`) take precedence over `series.color`, so I think this should work the same.

Alternatively, instead of making `series.color` an alternative option for both: `risingTrendColor` and `fallingTrendColor`, my suggestion in #12145 was to make `series.color` an alternative to `risingTrendColor` **or** `fallingTrendColor` (`risingTrendColor` IMHO should be replaced with `color`). For example we don't have.. `positiveColor` in regards to `negativeColor`. But to keep backward compatibility, I would leave `risingTrendColor` anyway.

Regarding docs description: In the PR it's not explained that `color` takes precedence over  `risingTrendColor` and `fallingTrendColor`.

bre1470: I was just trying to be consistent with current behaviour and as point.color takes precedence, I thought that this is the case for `series.color` as well. But I will change it, so the docs can be left as they are and everything works the same as before and similar to other options.

TorsteinHonsi:
Everything okay now @pawelfus ?

TorsteinHonsi:
Pending CI tests, have merged in master to see if visual diff improves.

PR #12408, opened 2019-11-14
Docs: Note that zones takes precedence over negativeColor

### LINKED ISSUE #12396: Adding zones to a series stops negativeColor from applying for negative values
#### Expected behaviour
When I have a series with negative values and have defined a negative value colour I would expect negative values to be displayed in this color when there is a zone defined.

#### Actual behaviour
As soon as a zone is added the negative colour is removed and it just uses the series color.
My example has an empty zone entry, removing this brings the negative color back.

https://jsfiddle.net/1vh8ygwd/

#### Product version
Highcharts.js 7.2.1

#### Affected browser(s)
Chrome 78

TorsteinHonsi:
Closes #12396

### DIFF
--- a/js/parts/Series.js
+++ b/js/parts/Series.js
@@ -763,7 +763,9 @@ null,
      */
     /**
      * The color for the parts of the graph or points that are below the
-     * [threshold](#plotOptions.series.threshold).
+     * [threshold](#plotOptions.series.threshold). Note that `zones` takes
+     * precedence over the negative color. Using `negativeColor` is
+     * equivalent to applying a zone with value of 0.
      *
      * @see In styled mode, a negative color is applied by setting this option
      *      to `true` combined with the `.highcharts-negative` class name.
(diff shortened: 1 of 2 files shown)


pg1671:
Hi Torstein .. thanks for addressing this issue.

I was using negativeColor for red colors below 0 (and was setting a fillOpacity for styling) and I also wanted to use zones to style the series line in an (areaspline) graph to use dashes to indicate the current period rather than a solid line.

I have 2 problems:
1. How do I apply multiple styles with zones. Eg. If I use a zone to set red below zero as in [{ value: 0, color: 'red' }] then this overrides the fillOpacity that applied when I was using negativeColor and just uses a solid red colour for the fill.

2. How do I target multiple axis with zones.  I want to target the y axis for positive\negative colors and also the x axis to style the series line.  The zoneAxis setting sits outside the zones array so if there are multiple zones how do you target an axis?

thanks.

pawelfus:
Hi @pg1671 

Sorry for the late reply! To answer your questions:

1. Don't use named colors, but hex/rgb(a) formats instead: https://jsfiddle.net/BlackLabel/2jwb7kvm/3/ - this way opacity can be applied

2. Only one axis can apply zones. You can create request for this functionality on our [UserVoice](https://highcharts.uservoice.com/forums/55896-highcharts-javascript-api).

PR #12418, opened 2019-11-15
Tools: Force checkout branch for visual comparison

activist:
To make sure visual test comparison runs even if checking someone has forgot to check in something  (with reference to ts compiled js files)

@TorsteinHonsi  Merging this could potentially make the CI not uncover these non-commited files unless their inner workings has changed someway that it would be caught by a test instead. So if you don't want to merge this it is totally fine, but ideally we should write a different test/pre-commit check in a more sensible place that would uncover such errors. 

We could also implement a check on CI in the initial step that exits with an error code if there are found any uncommited js files after running `gulp scripts`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -182,8 +182,8 @@ jobs:
       - <<: *load_workspace
       - run: sudo apt-get install -y rsync
       - run: > # checkout master at diverge point of current branch and install
-          git checkout master && git fetch origin master && git reset --hard origin/master &&
-          git checkout $(git merge-base master ${CIRCLE_BRANCH}) && git log --oneline -5 &&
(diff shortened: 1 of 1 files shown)


TorsteinHonsi:
What do you think @bre1470 ? We discussed this earlier and concluded that it is a temporary situation and not worth the effort checking it, plus the CI failure will put pressure on us to fix the compiled JS files...

bre1470:
If the `js` folder would go directly into package distribution, it might be necessary to check it in CI. So I agree that this going to far for the situation we have. We can probably come up with some modification in the linting task to force sync  of `js` and `ts` during commits.

TorsteinHonsi:
Okay, thanks. I'm closing this then, but leaving the branch.

PR #12420, opened 2019-11-15
bugfix/12355-series-destroy-leak

### LINKED ISSUE #12355: Memory leak when calling series.update
#### Expected behaviour
Updating series should not result in a memory `leak`.

#### Actual behaviour
When calling `series.update()` new events are registered on the `series` and old ones are not removed.

Open one of the `codepens` provided below and see how `events` array grows.

Polar [CodePen](https://codepen.io/skrtheboss/pen/YzzLxPr)
Non Polar [CodePen](https://codepen.io/skrtheboss/pen/gOOzGPV)

#### Product version
Highcharts 7.2.1

#### Affected browser(s)
all

bre1470:
- Fixed #12355, multiple destroy events on series.

### DIFF
--- a/ts/indicators/indicators.src.ts
+++ b/ts/indicators/indicators.src.ts
@@ -33,6 +33,7 @@ declare global {
             public requiredIndicators: Array<string>;
             public useCommonDataGrouping: boolean;
             public init(chart: Chart, options: SMAIndicatorOptions): void;
+            public destroy(keepEvents?: boolean): void;
             public getName(): string;
             public getValues<TLinkedSeries extends Series>(
                 series: TLinkedSeries,
@@ -537,13 +538,16 @@ seriesType<Highcharts.SMAIndicator>(
                 yData: yData
             } as Highcharts.IndicatorValuesObject<TLinkedSeries>;
         },
-        destroy: function (this: Highcharts.SMAIndicator): void {
+        destroy: function (
+            this: Highcharts.SMAIndicator,
+            keepEvents?: boolean
+        ): void {
             this.dataEventsToUnbind.forEach(function (
                 unbinder: Function
             ): void {
                 unbinder();
             });
-            Series.prototype.destroy.call(this);
+            Series.prototype.destroy.call(this, keepEvents);
         }
     }
 );
--- a/ts/indicators/indicators.src.ts
+++ b/ts/indicators/indicators.src.ts
@@ -33,7 +33,6 @@ declare global {
             public requiredIndicators: Array<string>;
             public useCommonDataGrouping: boolean;
             public init(chart: Chart, options: SMAIndicatorOptions): void;
-            public destroy(keepEvents?: boolean): void;
             public getName(): string;
             public getValues<TLinkedSeries extends Series>(
                 series: TLinkedSeries,
@@ -538,16 +537,13 @@ seriesType<Highcharts.SMAIndicator>(
                 yData: yData
             } as Highcharts.IndicatorValuesObject<TLinkedSeries>;
         },
-        destroy: function (
-            this: Highcharts.SMAIndicator,
-            keepEvents?: boolean
-        ): void {
+        destroy: function (this: Highcharts.SMAIndicator): void {
             this.dataEventsToUnbind.forEach(function (
                 unbinder: Function
             ): void {
                 unbinder();
             });
-            Series.prototype.destroy.call(this, keepEvents);
+            Series.prototype.destroy.apply(this, arguments);
         }
     }
 );
--- a/ts/parts-more/Polar.ts
+++ b/ts/parts-more/Polar.ts
@@ -51,6 +51,7 @@ declare global {
             connectEnds?: boolean;
             data: Array<PolarPoint>;
             group: SVGElement;
+            hasClipCircleSetter?: boolean;
             kdByAngle?: boolean;
             points: Array<PolarPoint>;
             preventPostTranslate?: boolean;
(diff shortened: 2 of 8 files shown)


bre1470:
The issue is somehow fixed, but one unrelated `unit-tests/indicator-sma/recalculations` is failing in headless only for me.

@TorsteinHonsi Do you have an idea, why this is happening?

bre1470:
As far as I see, sma is not passing destroy arguments through to generic series. Is there a reason for this.

pawelfus:
I'm afraid this solution does not work when events are defined: https://jsfiddle.net/BlackLabel/yhbqxrg5/ - probably `keepEvents` should always be set to true on update, like it was. Right now `keepEvents` tells us when all events should be removed (to clean the memory completely eg on `series.remove()`) or shouldn't at all (e.g. on `series.update()`). What we need is kind of intermediate solution: keep some events and others should always be removed.

Maybe it's better to create `eventsToUnbind` array, where internal events are pushed (see Navigator.ts) and on destroy are always removed?

Alternative solution requires a bit of refactoring. Instead of binding/unbinding events, move logic to prototyped event:

```js
H.addEvent(
  H.Chart,
  'resize',
  () => {
    this.series.forEach(s => s.invertGroups())
  }
);
```

PS: I don't know why we even need that event. In theory, on `chart.redraw`/`chart.render` method is called anyway because of `series.invertGroups = setInvert;`, right?

> As far as I see, sma is not passing destroy arguments through to generic series. Is there a reason for this?

It's a bug, `apply(...arguments)` should be used instead of `call()`.

bre1470:
Then I am not sure, if this is possible to fix without breaking API. A year ago I already tried to fix the problem of leaking event callbacks, but then it also did not apeased every use case. The problem lies basically in anonymous callback functions, which are not comparable.

bre1470:
I will try the "events to unbind" approach from Navigator. At least it could help with `destroy` event. It would be good, if we could distinguish between internal and user event callbacks everywhere in some upper generic Highchart `Component` class. This one could also provide support for unique IDs/symbols, serialization, and reflection.

pawelfus:
> At least it could help with destroy event.

Why it won't help with `afterRender` event? `hasClipCircleSetter` and `clipCircle ` are probably cleared like all other props from series, so I think it might be a safe solution.

> It would be good, if we could distinguish between internal and user event callbacks everywhere in some upper generic Highcharts.Component class. This one could also provide support for unique IDs/symbols, serialization, and reflection.

Could you elaborate? I'm curious how do you see it. Right now we have kind of pub/sub solution, where events are stored in instances/prototypes (so we can remove them even when defined as anonymous function). Thanks!

bre1470:
> > At least it could help with destroy event.
> Why it won't help with `afterRender` event? `hasClipCircleSetter` and `clipCircle ` are probably cleared like all other props from series, so I think it might be a safe solution.

I did not found the reason. I hope that is now fixed with the last commit.

> > It would be good, if we could distinguish between internal and user event callbacks everywhere in some upper generic Highcharts.Component class. This one could also provide support for unique IDs/symbols, serialization, and reflection.
> 
> Could you elaborate? I'm curious how do you see it. Right now we have kind of pub/sub solution, where events are stored in instances/prototypes (so we can remove them even when defined as anonymous function). Thanks!

Well the idea behind is to have every instance to manage events itself instead of a utility function. For global events (on the prototype) we can create a static instance that manages this kind of events and gets triggered by the common instance implementation. That way it is easier to have additional custom modifications for event handling, though differentiate between internal and user events should be done already in the common instance implementation.

pawelfus:
Interesting, my demo doesn't work as polar anymore: https://jsfiddle.net/BlackLabel/yhbqxrg5/3/

bre1470:
> Interesting, my demo doesn't work as polar anymore: https://jsfiddle.net/BlackLabel/yhbqxrg5/3/

If I change the scripts sources to the local branch, there is no problem with the bugfix at all. Maybe it is some loading problem with the server?

bre1470:
@pawel Can you please try it locally and confirm? http://jsfiddle.net/cksaqu9r/

ts/parts-more/Polar.ts
```diff
@@ -352,74 +351,70 @@ if (seriesTypes.spline) {
...
+                    !series.yAxis.reversed &&
+                    (points[i].y as any) < (series.yAxis.min as any)
                 ) {
                     points[i].isNull = true;
                 }
             }
         }
 
         // Perform clip after render
-        if (!this.hasClipCircleSetter) {
```
pawelfus: I think this `if` should stay, otherwise we have leak here: http://jsfiddle.net/BlackLabel/em2hbrwn/ 

Click on legend multiple times and observe results in legend. The same happens on window's resize.

pawelfus:
> Can you please try it locally and confirm? http://jsfiddle.net/cksaqu9r/

Yes, locally works fine! Sorry for the false alarm :)

bre1470:
> > Can you please try it locally and confirm? http://jsfiddle.net/cksaqu9r/
> 
> Yes, locally works fine! Sorry for the false alarm :)

I am so thankful to this outcome. :D

ts/indicators/indicators.src.ts
```diff
@@ -537,13 +538,16 @@ seriesType<Highcharts.SMAIndicator>(
...
+            this: Highcharts.SMAIndicator,
+            keepEvents?: boolean
+        ): void {
             this.dataEventsToUnbind.forEach(function (
                 unbinder: Function
             ): void {
                 unbinder();
             });
-            Series.prototype.destroy.call(this);
+            Series.prototype.destroy.call(this, keepEvents);
```
pawelfus: Wouldn't it be better to use `Series.prototype.destroy.call(this, ...arguments);` or `Series.prototype.destroy.apply(this, arguments);`? `keepEvents` wasn't always there, so if another argument will be added we will need to fix this again..

bre1470:
I agree, but simply following `Navigator.removeEvents` for consistency. So there it should be renamed too. How about `Series.removeInternalEvents`?

pawelfus:
Perhaps we should simply move all events removing to this new function, so this:
```
            // remove all events
            if (!keepEvents) {
                removeEvent(series);
            } else {
                series.removeEvents();
            }
```
Will become this:
```
            series.removeEvents(keepEvents);
```

TorsteinHonsi:
Yes, I support that. And then `keepEvents` should be renamed to something more descriptive, like `isUpdating` or `keepEventsForUpdate`.

Issue #11305, opened 2019-07-04
New Series Type: Radial bar chart


Izothep:
Implementing Radial bar chart.

Internal: 
It should be possible to base on current polar chart logic, combined with methods used for creating gauge and solidgauge series.

Izothep:
Closed by https://github.com/highcharts/highcharts/pull/12426

Issue #565, opened 2011-11-14
[BUG] Wrong months labels and 31st October issue


floernoult:
I just tried your last release 2.1.9, and it appears that months labels are shifted by one. (October is displayed as November, November as December and so on....). You can see it on the website examples  (http://www.highcharts.com/demo/spline-irregular-time).

Moreover, I made a graph with one value per day during one month from October 13 to November 9, and the graph is displaying two data entries on November 1st, but October 31st doesn't exist. Consequently, from November 1st, the day label is wrong as I have to points of data on Monday November 1st (which actually is a tuesday).

Here my chart code if you want to test :

var myChart = new Highcharts.Chart({
                chart: {
                    renderTo: 'myChart',
                    defaultSeriesType: 'line',
                    height: 250,
                    marginBottom: 100,
                    width: chartWidth,
                    zoomType: 'x'
                },
                title: {
                    text: null
                },
                tooltip: {
                    crosshairs: true,
                    shared: true
                },
                xAxis: {
                    type: 'datetime',
                    maxZoom: 3 \* 24 \* 3600000,
                    labels: {
                        rotation: -45,
                        y: 25,
                        formatter: function() {
                            return Highcharts.dateFormat('%d. %b', this.value);
                        }
                    }
                },
                yAxis: [{
                    min: 0,
                    title: {
                        text: null
                    },
                    labels: {
                        style: {
                           color: '#558133'
                        }
                    }
                }, { 
                    min: 0,
                    title: {
                        text: null
                    },
                    opposite: true,
                    labels: {
                        style: {
                           color: '#3B5998'
                        }
                    }
                  }, { 
                    min: 0,
                    title: {
                        text: null
                    },
                    labels: {
                        style: {
                           color: '#AA4643'
                        }
                    },
                  }],
                series: [{
                    name: "Data 1",
                    yAxis: 1,
                    color: '#3B5998',
[...]

PR #12426, opened 2019-11-18
feature/11305-radial-bar-chart

PaulDalek:
Added inverted column series on polar (#11305).

Demos:
- http://jsfiddle.net/BlackLabel/ocL4eur8/
- http://jsfiddle.net/BlackLabel/dmerh72w/
- http://jsfiddle.net/BlackLabel/82qkoz60/
- http://jsfiddle.net/BlackLabel/5q4srbzt/
- http://jsfiddle.net/BlackLabel/t198cpm5/
- http://jsfiddle.net/BlackLabel/g7r3neyp/
- http://jsfiddle.net/BlackLabel/kmpcoxwe/
- http://jsfiddle.net/BlackLabel/anmLt18s/


js/parts-more/RadialAxis.js
```diff
@@ -436,15 +461,21 @@ addEvent(Axis, 'init', function (e) {
...
+        isCircular = !inverted && coll === 'xAxis' ||
+            inverted && coll === 'yAxis';
+        this.defaultRadialOptions = isCircular ?
+            this.defaultCircularRadialOptions :
+            merge(this.defaultYAxisOptions, this.defaultNonCircularRadialOptions);
+        // Prevents from rendering an additional undesirable grid line
+        // and label
+        if (!isCircular) {
+            this.defaultRadialOptions.minPadding = 0;
+        }
```
sebastianbochan: Why the IF is not in merge?

js/parts/PieSeries.js
```diff
@@ -597,15 +597,22 @@ seriesType('pie', 'line',
                     // start values
                     graphic.attr({
                         // animate from inner radius (#779)
-                        r: point.startR || (series.center[3] / 2),
+                        // If chart is inverted polar and series is column,
+                        // allow 0 as radius
+                        r: series.isRadialBar ?
+                            (args).start === (args).end ?
```
sebastianbochan: Do we need brackets in args?

js/parts/PieSeries.js
```diff
@@ -597,15 +597,22 @@ seriesType('pie', 'line',
...
                         end: startAngleRad
                     });
                     // animate
                     graphic.animate({
-                        r: args.r,
-                        start: args.start,
-                        end: args.end
+                        r: series.isRadialBar &&
+                            (args).start === (args).end ?
+                            0 : (args).r,
```
sebastianbochan: Do we need brackets?

js/parts/Series.js
```diff
@@ -4192,6 +4196,7 @@ null,
         if (series.visible) {
             series.drawPoints();
         }
+        fireEvent(series, 'afterDrawPoints');
         /* series.points.forEach(function (point) {
```
sebastianbochan: Do we need commented area?

js/parts/Series.js
```diff
@@ -4081,6 +4081,10 @@ null,
         if (!series.xAxis) {
             return;
         }
+        // Don't invert series on inverted polar
+        if (series.isRadialSeries) {
+            inverted = false;
```
sebastianbochan: Consider to use it as default option of serie `defaultCircularRadialOptions`

js/parts/Stacking.js
```diff
@@ -284,12 +284,20 @@ Chart.prototype.getStacks = function () {
...
 Axis.prototype.buildStacks = function () {
-    var axisSeries = this.series, reversedStacks = pick(this.options.reversedStacks, true), len = axisSeries.length, i;
+    var axisSeries = this.series, reversedStacks = pick(this.options.reversedStacks, true), len = axisSeries.length, actualSeries, options, i;
     if (!this.isXAxis) {
         this.usePercentage = false;
         i = len;
         while (i--) {
-            axisSeries[reversedStacks ? i : len - i - 1].setStackedPoints();
+            actualSeries = axisSeries[reversedStacks ? i : len - i - 1];
+            options = actualSeries.options;
```
sebastianbochan: I guess that `options` are series options. If yes, I think that `seriesOptions` will be more intuitive.

js/parts-more/Polar.js
```diff
@@ -397,77 +413,53 @@ if (seriesTypes.column) {
...
-                        yMin = yAxis.translate(yAxis.min);
-                        yMax = yAxis.translate(yAxis.max);
-                        if (!yAxis.reversed) {
-                            if (start < yMin) {
-                                start = end = 0;
-                            }
-                            else if (end < yMin) {
-                                end = yMin;
-                            }
+                        _a = __read([end, start], 2), start = _a[0], end = _a[1];
```
sebastianbochan: Please add comment what it is

js/parts-more/Polar.js
```diff
@@ -535,34 +532,75 @@ if (seriesTypes.column) {
...
+                    if (angle > 20 && angle < 160) {
+                        align = 'left'; // right hemisphere
+                    }
+                    else if (angle > 200 && angle < 340) {
+                        align = 'right'; // left hemisphere
+                    }
+                    else {
+                        align = 'center'; // top or bottom
+                    }
+                    options.align = align;
```
sebastianbochan: The wrap is really long, consider to extract logic of angles to separated function.

js/parts-more/RadialAxis.js
```diff
@@ -341,7 +341,39 @@ radialAxisMixin = {
...
+                    if (shapeArgs.start) {
+                        // Find a true value of the point based on the
+                        // angle
+                        value = axis.translate(options.point.rectPlotY, true);
+                    }
+                }
+                end = axis.getPosition(value);
+                x2 = end.x;
+                y2 = end.y;
+            }
```
sebastianbochan: The same situation like in wrap section above (extract function).

ts/parts-more/Polar.ts
```diff
@@ -654,6 +657,25 @@ if (seriesTypes.column) {
         if (xAxis.isRadial) {
             points = series.points;
             i = points.length;
+            yMin = yAxis.translate(yAxis.min as any);
+            yMax = yAxis.translate(yAxis.max as any);
+
+            // Finding a correct threshold
+            if (chart.inverted && H.isNumber(threshold)) {
+                (thresholdAngleRad as any) = yAxis.translate(threshold);
```
sebastianbochan: We should avoid to casting `as any`

ts/parts-more/Polar.ts
```diff
@@ -685,81 +707,52 @@ if (seriesTypes.column) {
...
-                        // inverted polar bars
-                        start = 0;
+                        // Initial start and end angles for radial bar
+                        start = thresholdAngleRad;
                         end = point.plotY;
                     }
 
-                    // Prevent from rendering point outside acceptable
-                    // circle range
                     if ((start as any) > (end as any)) {
```
sebastianbochan: We should avoid casting `as any`

ts/parts-more/Polar.ts
```diff
@@ -260,14 +261,14 @@ seriesProto.toXY = function (
...
-        plotY = point.plotY as any,
+        plotY = point.plotY,
         series = point.series,
         inverted = chart.inverted,
         clientX;
 
     // Corrected y position of inverted series other than column
     if (inverted && series && !series.isRadialBar) {
-        plotY = (point.plotY as any) = yAxis.translate(point.y as any);
+        plotY = point.plotY = yAxis.translate(point.y as any) || 0;
```
sebastianbochan: what is the reason of using `as any` ?

sebastianbochan:
In many places you use `as any`. Consider to replace it with correct types.

ts/parts/Axis.ts
```diff
@@ -2906,21 +2903,6 @@ extend(Axis.prototype, /** @lends Highcharts.Axis.prototype */{
...
-         * @sample {highcharts} highcharts/yaxis/angle/
-         *         Dual axis polar chart
-         *
-         * @type      {number}
-         * @default   0
-         * @since     4.2.7
-         * @product   highcharts
-         * @apioption yAxis.angle
-         */
-
```
pawelfus: Why this is removed?

ts/parts/Axis.ts
```diff
@@ -2946,22 +2928,6 @@ extend(Axis.prototype, /** @lends Highcharts.Axis.prototype */{
...
-         *
-         * @sample {highcharts} highcharts/demo/polar-spider/
-         *         Polygon grid lines
-         * @sample {highcharts} highcharts/yaxis/gridlineinterpolation/
-         *         Circle and polygon
-         *
-         * @type       {string}
-         * @product    highcharts
-         * @validvalue ["circle", "polygon"]
-         * @apioption  yAxis.gridLineInterpolation
```
pawelfus: Why this is removed?

js/parts-more/RadialAxis.js
PaulDalek: I can see now that this should apply to xAxis on inverted polar chart only to prevent from rendering undesirable grid line and label (-1 in case of default options). Made required changes.

js/parts/PieSeries.js
PaulDalek: It is already corrected in the newest version.

js/parts/PieSeries.js
PaulDalek: It is already corrected in the newest version too.

js/parts/Series.js
PaulDalek: This piece of code is already inside a comment block on the master script.

js/parts/Series.js
PaulDalek: I'm not sure if I understand this one but the isRadialSeries is a simple flag that is set on afterInit event to recognize series on inverted polar. It is used to prevent from inverting group of inverted polar series even though chart.inverted is set to true. I think it is unnecessary to place it in defaultCircularRadialOptions object, as it contains options for axis. What do you think @TorsteinHonsi ?

js/parts/Stacking.js
PaulDalek: I agree, corrected. Thanks!

js/parts-more/Polar.js
PaulDalek: I have used here destructuring arrays to swap variables. I have added a fitting comment.

js/parts-more/Polar.js
PaulDalek: I moved the logic that finds the right alignments to new function called findAlignments. It reduces the alignDataLabel wrap a little bit.

js/parts-more/RadialAxis.js
PaulDalek: Made a similar changes to ones mentioned comment above. I have created getCrosshairPosition function and place crosshair logic there.

ts/parts-more/Polar.ts
PaulDalek: Corrected by adding a simple check if thresholdAngleRad returns a truthy value.

ts/parts-more/Polar.ts
PaulDalek: Previously mentioned check on the thresholdAngleRad property and checking the start variable resolves this problem as well.

ts/parts-more/Polar.ts
PaulDalek: I have added an additional check to see if the point.y is a number.

ts/parts-more/Polar.ts
```diff
@@ -793,6 +1049,60 @@ H.addEvent(H.Chart, 'afterDrawChartBox', function (
...
+
+H.addEvent(H.Series, 'afterInit', function (
+    this: Highcharts.Series
+): void {
+    var chart = this.chart;
+
+    // Add flags that identifies radial inverted series
+    if (chart.inverted && chart.polar) {
+        this.isRadialSeries = true;
+        if (!(['column', 'bar'].indexOf(this.type) < 0)) {
```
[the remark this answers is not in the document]
PaulDalek: Corrected, thank you for the suggestion!

ts/parts-more/Polar.ts
```diff
@@ -607,38 +629,221 @@ if (seriesTypes.column) {
...
+                                start = yAxis.translate(stackValues[0]);
+                                end = yAxis.translate(stackValues[1]);
+
+                                // If starting point is beyond the
+                                // range, set it to 0
+                                if (start < 0) {
+                                    start = 0;
+                                } else if (start > visibleRange) {
+                                    start = visibleRange;
+                                }
```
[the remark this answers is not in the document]
PaulDalek: Corrected, thank you for the suggestion!

ts/parts-more/RadialAxis.ts
```diff
@@ -772,17 +882,31 @@ addEvent(Axis as any, 'init', function (
...
     } else if (polar) {
         extend(this, radialAxisMixin);
-        isCircular = isX;
-        this.defaultRadialOptions = isX ?
-            this.defaultRadialXOptions :
-            merge(this.defaultYAxisOptions, this.defaultRadialYOptions);
 
+        // Check which axis is circular
+        isCircular = !inverted && coll === 'xAxis' ||
+            inverted && coll === 'yAxis';
```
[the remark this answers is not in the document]
PaulDalek: I have replaced conditional with the this.horiz but I have left isCircular property because of its many occurrences later in the code.

ts/parts/Axis.ts
PaulDalek: Right now the angle property can be used both on, xAxis and yAxis, so I have moved it to the RadialAxis.ts part instead.

ts/parts/Axis.ts
PaulDalek: Same thing as above but with the gridlineinterpolation option.

PaulDalek:
@TorsteinHonsi When it comes to the visual differences in some non-inverted polar chart examples, it was a result of applying the `minPadding` (set to 0) not only to xAxis on inverted but also to yAxis on non-inverted chart. I made a correction to fix that.

ts/parts/Axis.ts
pawelfus: Ahh! Missed that, thanks :)

ts/parts/Axis.ts
pawelfus: Got it!

ts/parts-more/RadialAxis.ts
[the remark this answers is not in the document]
pawelfus: Thanks 👍

js/parts-more/RadialAxis.js
TorsteinHonsi: On a category X axis the `minPadding` should be 0, maybe the problem is somewhere else?

js/parts-more/Polar.js
TorsteinHonsi: Unfortunately this polyfill is inserted to support destructuring arrays. The problem with this is that so much junk code is inserted in order to simplify the simple task of swapping variables. Also, the current bundler isn't smart enough to add this to the upper scope, so it will potentially be added to every module. Long story short, we'll have to wait a while with array destructuring, and for the time being swap variables the old way :(

js/parts/Series.js
sebastianbochan: 1. We should not re-define function's parameters (like `inverted`)
2. I suggest to move this param to serie's init / default polar options.

js/parts-more/RadialAxis.js
PaulDalek: The additional tick (-1) was a result of merging the `defaultNonCircularRadialOptions` object with the `defaultYAxisOptions` where minPadding is set to 0.05 and `startOnTick` is set to true no matter if vertical axis was X or Y type. Now the `defaultNonCircularRadialOptions` is merged into a correct set of options (the `defaultOptions` in case of `xAxis` and the `defaultYAxisOptions` in case of `yAxis`).

js/parts/Series.js
PaulDalek: Ok, thanks for the further information. This options is already set on the event but it is `afterInit`. As for the modifying the `inverted` parameter, I've got rid of it, instead I added a simple check in `setInvert`.

js/parts-more/Polar.js
PaulDalek: Corrected, thanks!

ts/parts/Stacking.ts
```diff
@@ -535,13 +536,24 @@ Axis.prototype.buildStacks = function (this: Highcharts.Axis): void {
...
     if (!this.isXAxis) {
         this.usePercentage = false;
         i = len;
         while (i--) {
-            axisSeries[reversedStacks ? i : len - i - 1].setStackedPoints();
+            actualSeries = axisSeries[reversedStacks ? i : len - i - 1];
+            options = actualSeries.options;
+
+            // Threshold should always be set to 0 in case of percent
+            // stacking in inverted polar bars
```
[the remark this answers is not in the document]
PaulDalek: I think that setting `threshold` with some other value than 0 in case when `stacking` is set to `percent` doesn't make much sense in general. I got rid of this piece of code. Should we let the user set a custom `threshold` value and make it affect percent stacking (that's how it works right now on cartesian plots)? Chart will work but the data won't be correctly placed.

ts/parts/ColumnSeries.ts
```diff
@@ -1066,9 +1066,12 @@ seriesType<Highcharts.ColumnSeries>(
 
                     if (graphic) { // update
                         graphic[verb](
-                            merge(shapeArgs)
+                            merge(shapeArgs),
+                            null,
+                            function (): void {
+                                H.fireEvent(point, 'afterPointAnimate');
```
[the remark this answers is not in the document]
PaulDalek: Unfortunately, I cannot change the `shapeArgs` in the translate override for polar the same way as the graphic is updated on `afterPointAnimate` because then points would animate directly to the center of the pane (like in the following demo after hiding the Series 2: http://jsfiddle.net/BlackLabel/ycrb6qf5/). As for using one of the existing events, the closest one is the `afterRender` at the end of the series' render function but that would result in points simply disappear (the first three points: http://jsfiddle.net/BlackLabel/xv04jby1/). Could we maybe instead of firing the `afterPointAnimate` event for every column point, make that every radial bar point would have reference to a custom complete function and the call of the animate function in the `drawPoints` would look like the following: `graphic[verb](merge(shapeArgs), void 0, point.complete)`. This way the `drawPoints` function would run just as before for regular column points but in case of radial bar points the `complete` function would be added and resulting in a correct animation process.

ts/parts-more/Polar.ts
```diff
@@ -607,38 +629,221 @@ if (seriesTypes.column) {
...
+                    // A correct value for stacked or not fully visible
+                    // point
+                    point.plotY = (start <
+                        (series.translatedThreshold as any) ?
+                        start : end) - startAngleRad;
+
+                    if (!point.graphic) {
+                        // The graphic cannot be added to a group here
+                        // as the group doesn't exist yet, it is added
+                        // later on the afterDrawPoints event
```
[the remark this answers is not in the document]
PaulDalek: In a nutshell I was able to resolve this by using and changing the `point.shapeType` property.

ts/parts-more/Polar.ts
```diff
@@ -793,6 +1049,60 @@ H.addEvent(H.Chart, 'afterDrawChartBox', function (
...
+    // Add flags that identifies radial inverted series
+    if (chart.inverted && chart.polar) {
+        this.isRadialSeries = true;
+        if (!(['column', 'bar'].indexOf(this.type) < 0)) {
+            this.isRadialBar = true;
+        }
+    }
+});
+
+H.addEvent(H.Series, 'afterDrawPoints', function (
```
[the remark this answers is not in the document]
PaulDalek: Corrected, I got rid of the `afterDrawPoints` event entirely.

PaulDalek:
@TorsteinHonsi As for the issue with cutting the last category on `xAxis` near the max value, I have corrected that by modifying the `beforeSetTickPositions` function but I'm not sure if it is a correct place.

js/parts-more/Polar.js
```diff
@@ -391,25 +374,40 @@ if (seriesTypes.column) {
             if (threshold === null) {
                 threshold = 0;
             }
```
sebastianbochan: Instead of IF try use  `threshold = options.threshold || 0`

js/parts-more/Polar.js
```diff
@@ -529,8 +514,16 @@ if (seriesTypes.column) {
                             }
                         }
                     }
+                    else {
+                        point.shapeType = 'arc';
```
sebastianbochan: in IF and ELSE there is 'arc'.

Consider to set the `arc` by default and add 
```
if (point.graphic.element.nodeName === 'path') {
   point.shapeType = 'path';
}
```
only in IF

js/parts-more/RadialAxis.js
```diff
@@ -243,6 +245,11 @@ radialAxisMixin = {
             typeof pick(this.userMax, this.options.max) === 'undefined' &&
             correctFloat(this.endAngleRad - this.startAngleRad) ===
                 correctFloat(2 * Math.PI));
+        // This will lead to add an extra tick to xAxis in order to display a
+        // correct range on inverted polar
+        if (!this.isCircular && this.chart.inverted) {
```
sebastianbochan: Use the `inverted` variable which is also used in #565 line.

js/parts-more/Polar.js
PaulDalek: Corrected, thanks for the suggestion!

js/parts-more/Polar.js
PaulDalek: Corrected, thanks for the suggestion!

ts/parts/Stacking.ts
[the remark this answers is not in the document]
TorsteinHonsi: Yes I think we should go for consistence. So if this is a bug with polar, then it is also a bug with cartesian.

ts/parts-more/Polar.ts
```diff
@@ -511,46 +533,54 @@ var polarAnimate = function (
...
+            }
+        } else {
+            // Enable animation on polar charts only in SVG. In VML, the scaling
+            // is different, plus animation would be so slow it would't matter.
+            if (chart.renderer.isSVG) {
 
-        // Enable animation on polar charts only in SVG. In VML, the scaling
-        // is different, plus animation would be so slow it would't matter.
-        if (chart.renderer.isSVG) {
+                if (animation === true) {
```
TorsteinHonsi: Use `Highcharts.animObject` instead.

ts/parts-more/Polar.ts
```diff
@@ -607,82 +637,341 @@ if (seriesTypes.column) {
...
+                        end: end
+                    };
+
+                    // A correct value for stacked or not fully visible
+                    // point
+                    point.plotY = (defined(series.translatedThreshold) &&
+                        (start < series.translatedThreshold ? start : end)) -
+                            startAngleRad;
+
+                    if (point.graphic &&
```
TorsteinHonsi: This part should be refactored out of the `translate` function. In general, the translate function should only create the shapeArgs and do processing of positions. All actual rendering and creating/updating of `graphic` should happen in the `drawPoints` function. If necessary, you can create a `drawPoints` event that is fired at the top of the ColumnSeries `drawPoints` function.

js/parts-more/RadialAxis.js
```diff
@@ -8,13 +8,29 @@
  *
  * */
 'use strict';
+var __read = (this && this.__read) || function (o, n) {
```
TorsteinHonsi: This is a polyfill for array destructuring assignments. At the time being we need to avoid those descructuring assignments in order to prevent our modules to be bloated with this polyfill.

ts/parts-more/Polar.ts
PaulDalek: Corrected, thanks!

js/parts-more/RadialAxis.js
PaulDalek: Sorry, I previously missed that destructuring assignment. Corrected.

docs/chart-and-series-types/radial-bar-chart.md
```diff
@@ -0,0 +1,40 @@
+Radial Bar (Inverted Polar) Chart
```
TorsteinHonsi: Radial (or Circular) Bar Chart

docs/chart-and-series-types/radial-bar-chart.md
```diff
@@ -0,0 +1,40 @@
...
+A radial bar (or inverted polar bar) series is a form of visualization column data on polar coordinate system. Since the `xAxis` is vertical and `yAxis` is circular, as opposed to non-inverted variant, shape of points are circular.
+
+<iframe style="width: 100%; height: 600px; border: none;" src=https://www.highcharts.com/samples/embed/highcharts/demo/polar-radial-bar allow="fullscreen"></iframe>
+
+Click [here](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/demo/polar-radial-bar/) to check the code.
+
+Getting started
+---------------
+
+In order to create radial bar chart, the 'highcharts.js' and 'highcharts-more.js' scripts are required to be loaded, just like in case of non-inverted version of a polar chart.
```
TorsteinHonsi: Also show how we make a column into a radial/circular bar by setting polar and inverted to true.

ts/parts-more/Polar.ts
```diff
@@ -793,6 +1039,39 @@ H.addEvent(H.Chart, 'afterDrawChartBox', function (
...
+    // Add flags that identifies radial inverted series
+    if (chart.inverted && chart.polar) {
+        this.isRadialSeries = true;
+        if (this instanceof seriesTypes.column) {
+            this.isRadialBar = true;
+        }
+    }
+});
+
+H.addEvent(H.Series, 'afterRender', function (
```
TorsteinHonsi: We don't use `point.complete` anymore, we can get rid of this now, right? Also make sure it is removed from the TypeScript interface.

ts/parts-more/RadialAxis.ts
```diff
@@ -202,7 +221,7 @@ radialAxisMixin = {
     },
 
     // Circular axis around the perimeter of a polar chart
-    defaultRadialXOptions: {
+    defaultCircularRadialOptions: {
```
TorsteinHonsi: Should be `defaultCircularOptions`

ts/parts-more/RadialAxis.ts
```diff
@@ -220,8 +239,47 @@ radialAxisMixin = {
     },
 
     // Radial axis, like a spoke in a polar chart
-    defaultRadialYOptions: {
+    defaultNonCircularRadialOptions: {
```
TorsteinHonsi: Should be `defaultRadialOptions`

ts/parts-more/RadialAxis.ts
```diff
@@ -773,17 +927,27 @@ addEvent(Axis as any, 'init', function (
...
         extend(this, radialAxisMixin);
-        isCircular = isX;
-        this.defaultRadialOptions = isX ?
-            this.defaultRadialXOptions :
-            merge(this.defaultYAxisOptions, this.defaultRadialYOptions);
 
+        // Check which axis is circular
+        isCircular = this.horiz;
+
+        this.defaultRadialOptions = isCircular ?
```
TorsteinHonsi: This property should probably be called `this.defaultPolarOptions` since it can hold either radial or circular axis options.

ts/parts/ColumnSeries.ts
```diff
@@ -58,10 +58,15 @@ declare global {
             borderWidth?: number;
             color?: (ColorString|GradientColorObject|PatternObject);
         }
+        interface CompleteAnimateFunction {
```
TorsteinHonsi: No longer used, can be removed

ts/parts/ColumnSeries.ts
```diff
@@ -58,10 +58,15 @@ declare global {
...
+            (this: SVGElement): void;
+        }
         class ColumnPoint extends LinePoint {
             public allowShadow?: boolean;
             public barX: number;
+            public complete?: CompleteAnimateFunction;
```
TorsteinHonsi: No longer used, remove it

samples/unit-tests/polar/column-inverted/demo.js
```diff
@@ -0,0 +1,204 @@
+QUnit.skip('Positions of the points.', function (assert) {
```
TorsteinHonsi: Fix these tests for the refactored code (`startR` no longer exists on point).

docs/chart-and-series-types/radial-bar-chart.md
PaulDalek: Corrected, thanks!

docs/chart-and-series-types/radial-bar-chart.md
PaulDalek: Added the `Creating a radial bar chart` subsection in `Getting started` section and new `Data format` section.

ts/parts-more/Polar.ts
PaulDalek: Got rid of the event, thanks!

ts/parts-more/RadialAxis.ts
PaulDalek: Corrected, thanks!

ts/parts-more/RadialAxis.ts
PaulDalek: Corrected, thanks!

ts/parts-more/RadialAxis.ts
PaulDalek: Corrected, thanks!

ts/parts/ColumnSeries.ts
PaulDalek: Removed, thanks!

ts/parts/ColumnSeries.ts
PaulDalek: Removed, thanks!

samples/unit-tests/polar/column-inverted/demo.js
PaulDalek: Instead of using the `startR`, the `start` and `end` are compared in order to check if points are outside the visible range.

PR #12433, opened 2019-11-18
Update internationalization.md

mkrkushwah:


TorsteinHonsi:
Thanks for the PR, but there is some random change in the header, plus I don't see a relevant change in the reference to the Chinese fiddle?

KacperMadej:
Closing due to inactivity.

PR #12446, opened 2019-11-19
docs/12409-remove-html-tags

### LINKED ISSUE #12409: Docs: Missed closing html tags
#### Expected behaviour
Adding closing tags to the docs.

#### Actual behaviour
Warnings about requiring `end` tag in Xcode.

Please see https://github.com/highcharts/highcharts-ios/issues/255 for more info.
#### Product version

7.2.1

bre1470:
Close #12409, Removed HTML from doclets.

### DIFF
--- a/js/parts-gantt/Tree.js
+++ b/js/parts-gantt/Tree.js
@@ -23,7 +23,7 @@ var isFunction = function (x) {
  * @function Highcharts.Tree#getListOfParents
  *
  * @param {Array<*>} data
- *        List of points set in options. `Array<*>.parent`is parent id of point.
+ *        List of points set in options. `Array.parent` is parent id of point.
  *
  * @param {Array<string>} ids
  *        List of all point ids.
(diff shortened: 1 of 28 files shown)


js/parts-gantt/Tree.js
```diff
@@ -23,7 +23,7 @@ var isFunction = function (x) {
  * @function Highcharts.Tree#getListOfParents
  *
  * @param {Array<*>} data
- *        List of points set in options. `Array<*>.parent`is parent id of point.
```
ihnatmoisieiev: will this affect in `tree.json`?

ts/parts-gantt/Tree.ts
```diff
@@ -83,7 +83,7 @@ var isFunction = function (x: unknown): x is Function {
  * @function Highcharts.Tree#getListOfParents
  *
  * @param {Array<*>} data
- *        List of points set in options. `Array<*>.parent`is parent id of point.
```
ihnatmoisieiev: the same question as above :)

TorsteinHonsi: No it won't affect the structure of the tree. This is just the description.

PR #12448, opened 2019-11-19
feature/slow-stochastic-indicator

pawelfus:
Added Slow Stochastic indicator.

### DIFF
--- a/ts/indicators/slow-stochastic.src.ts
+++ b/ts/indicators/slow-stochastic.src.ts
@@ -1,0 +1,191 @@
+/* *
+ *
+ *  License: www.highcharts.com/license
+ *
+ *  !!!!!!! SOURCE GETS TRANSPILED BY TYPESCRIPT. EDIT TS FILE ONLY. !!!!!!!
+ *
+ * */
+
+'use strict';
+
+import H from '../parts/Globals.js';
+
+/**
+ * Internal types
+ * @private
+ */
+declare global {
+    namespace Highcharts {
+        class SlowStochasticIndicator extends StochasticIndicator {
+            public nameBase: string;
+            public init(): void;
+            public getValues<TLinkedSeries extends Series>(
+                series: TLinkedSeries,
+                params: SlowStochasticIndicatorParamsOptions
+            ): (IndicatorValuesObject<TLinkedSeries>|undefined);
+        }
+
+        interface SlowStochasticIndicatorParamsOptions
+            extends StochasticIndicatorParamsOptions {
+            // for inheritance
+        }
+
+        class SlowStochasticIndicatorPoint extends StochasticIndicatorPoint {
+            public series: SlowStochasticIndicator;
+        }
+
+        interface SlowStochasticIndicatorOptions
+            extends StochasticIndicatorOptions, MultipleLinesIndicatorOptions {
+            params?: SlowStochasticIndicatorParamsOptions;
+        }
+
+        interface SeriesTypesDictionary {
+            slowstochastic: typeof SlowStochasticIndicator;
+        }
+    }
+}
+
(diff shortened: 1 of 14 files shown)


ts/indicators/slow-stochastic.src.ts
```diff
@@ -0,0 +1,191 @@
...
+                    this,
+                    ({
+                        xData: slowValues.xData,
+                        yData: fastYData
+                    } as any),
+                    {
+                        index: 1,
+                        period: periods[2]
+                    }
+                );
```
pawelfus: Can I do something with this little monster? Especially lines 137-139.. @bre1470

ts/indicators/slow-stochastic.src.ts
```diff
@@ -0,0 +1,191 @@
...
+            this: Highcharts.SlowStochasticIndicator,
+            series: TLinkedSeries,
+            params: Highcharts.SlowStochasticIndicatorParamsOptions
+        ): (Highcharts.IndicatorValuesObject<TLinkedSeries>|undefined) {
+            const periods: Array<number> = (params.periods as any),
+                fastValues = seriesTypes.stochastic.prototype.getValues.call(
+                    this,
+                    series,
+                    params
+                ) as any,
```
pawelfus: Decided to cast to any, I couldn't resolve that issue. It seems that getting values from other indicators is a bit cumbersome. FYI @bre1470

ts/indicators/slow-stochastic.src.ts
```diff
@@ -0,0 +1,191 @@
...
+
+/**
+ * Internal types
+ * @private
+ */
+declare global {
+    namespace Highcharts {
+        class SlowStochasticIndicator extends StochasticIndicator {
+            public nameBase: string;
+            public init(): void;
```
pawelfus: I assume that I should define here all methods that are overwritten, right? @bre1470

bre1470: Yes, except if you do a super call with all arguments (`apply(this, arguments)`).

ts/indicators/slow-stochastic.src.ts
bre1470: Please see `ts/README.md` and make use of parentheses with type unions. We have to discuss the increase of line characters though.

ts/indicators/slow-stochastic.src.ts
bre1470: If I understand it correctly, stochastic is expecting a two-dimensional series (OHLC), but limiting the generic variable `TValueSeries` leads to inheritence conflict. So I agree with that.

ts/indicators/slow-stochastic.src.ts
pawelfus: Ahh, I see, will do, thanks!

ts/indicators/slow-stochastic.src.ts
pawelfus: Maybe somewhere `Highcharts.IndicatorValuesObject<TLinkedSeries>` makes this incompatible?

Just tried to compare with `ppo` indicator, where it's first cast as `IndicatorValuesObject`, then casted to `any`:
```js
            LPE = EMA.prototype.getValues.call(this, series, {
                index: index,
                period: periods[1]
            }) as Highcharts.IndicatorValuesObject<TLinkedSeries>;

            ...

            for (...) {
                PPO.push([(LPE as any).xData[i], oscillator]);
                xData.push((LPE as any).xData[i]);
                yData.push(...);
            }
```

Anyway, if this makes sense to you @bre1470 then I'm fine :)

ts/indicators/slow-stochastic.src.ts
pawelfus: Ok, thanks! 👍

PR #12460, opened 2019-11-20
enhancement/11308-center-column-nulls

raf18seb:
Added new property `series.ignoreNulls` which centers the columns within a group when others have a null value or missing points.

**Live demos**
column series: https://jsfiddle.net/BlackLabel/oeLvzgdw
columnrange series: https://jsfiddle.net/BlackLabel/yzj2pvht

Possible options:
- `false` (old behavior)
- `'normal'` (center columns and align them to the center of category),
- `'evenlySpaced'` (place columns evenly relative to the category width),
- `'fillSpace'` (columns fill the category width regardless of their quantity in the category).
Any naming or API structure suggestions are welcome.

To do:
- [x] tests
- [x] demos
- [x] stackLabels should be positioned correctly
- [x] `ignoreNulls: 'normal'` should work with multiple Y axes
- [x] `ignoreNulls` with `data` module

If `ignoreNulls: 'normal'` is gonna be the new default, I suggest adding this as an experimental feature with `ignoreNulls: false` in the next release and change the default in the future.

### DIFF
--- a/js/parts-more/ColumnRangeSeries.js
+++ b/js/parts-more/ColumnRangeSeries.js
@@ -126,6 +126,9 @@ seriesType('columnrange', 'arearange', merge(defaultPlotOptions.column, defaultP
     drawGraph: noop,
     getSymbol: noop,
     // Overrides from modules that may be loaded after this module
+    addPoint: function () {
+        return colProto.addPoint.apply(this, arguments);
+    },
     crispCol: function () {
         return colProto.crispCol.apply(this, arguments);
     },
@@ -135,14 +138,17 @@ seriesType('columnrange', 'arearange', merge(defaultPlotOptions.column, defaultP
     drawTracker: function () {
         return colProto.drawTracker.apply(this, arguments);
     },
-    hasPointInX: function (point, otherSeries) {
-        return colProto.hasPointInX.apply(this, arguments);
+    findMinColumnWidth: function () {
+        return colProto.findMinColumnWidth.apply(this, arguments);
     },
     getColumnCount: function (point) {
         return colProto.getColumnCount.apply(this, arguments);
     },
     getColumnMetrics: function () {
         return colProto.getColumnMetrics.apply(this, arguments);
+    },
+    hasPointInX: function (point, otherSeries) {
+        return colProto.hasPointInX.apply(this, arguments);
     },
     pointAttribs: function () {
         return colProto.pointAttribs.apply(this, arguments);
--- a/js/parts-more/ColumnRangeSeries.js
+++ b/js/parts-more/ColumnRangeSeries.js
@@ -135,6 +135,12 @@ seriesType('columnrange', 'arearange', merge(defaultPlotOptions.column, defaultP
     drawTracker: function () {
         return colProto.drawTracker.apply(this, arguments);
     },
+    hasPointInX: function (point, otherSeries) {
+        return colProto.hasPointInX.apply(this, arguments);
+    },
+    getColumnCount: function (point) {
+        return colProto.getColumnCount.apply(this, arguments);
+    },
     getColumnMetrics: function () {
         return colProto.getColumnMetrics.apply(this, arguments);
     },
@@ -154,7 +160,9 @@ seriesType('columnrange', 'arearange', merge(defaultPlotOptions.column, defaultP
         return colProto.translate3dShapes.apply(this, arguments);
     }
 }, {
-    setState: colProto.pointClass.prototype.setState
+    setState: colProto.pointClass.prototype.setState,
+    remove: colProto.pointClass.prototype.remove,
+    update: colProto.pointClass.prototype.update
 });
 /**
  * A `columnrange` series. If the [type](#series.columnrange.type)
--- a/js/parts-more/ColumnRangeSeries.js
+++ b/js/parts-more/ColumnRangeSeries.js
@@ -138,8 +138,11 @@ seriesType('columnrange', 'arearange', merge(defaultPlotOptions.column, defaultP
     drawTracker: function () {
         return colProto.drawTracker.apply(this, arguments);
     },
-    findMinColumnWidth: function () {
-        return colProto.findMinColumnWidth.apply(this, arguments);
+    getMaxColumnCount: function () {
+        return colProto.getMaxColumnCount.apply(this, arguments);
+    },
+    getMinColumnWidth: function () {
+        return colProto.getMinColumnWidth.apply(this, arguments);
     },
     getColumnCount: function (point) {
         return colProto.getColumnCount.apply(this, arguments);
@@ -147,8 +150,8 @@ seriesType('columnrange', 'arearange', merge(defaultPlotOptions.column, defaultP
     getColumnMetrics: function () {
         return colProto.getColumnMetrics.apply(this, arguments);
     },
-    hasPointInX: function (point, otherSeries) {
-        return colProto.hasPointInX.apply(this, arguments);
+    hasValueInX: function (point, otherSeries) {
+        return colProto.hasValueInX.apply(this, arguments);
     },
     pointAttribs: function () {
         return colProto.pointAttribs.apply(this, arguments);
--- a/js/parts-more/ColumnRangeSeries.js
+++ b/js/parts-more/ColumnRangeSeries.js
@@ -125,7 +125,9 @@ seriesType('columnrange', 'arearange', merge(defaultPlotOptions.column, defaultP
     trackerGroups: ['group', 'dataLabelsGroup'],
     drawGraph: noop,
     getSymbol: noop,
-    // Overrides from modules that may be loaded after this module
+    // Overrides from modules that may be loaded after this module.
+    // All below methods always fire the currect method from colProto
+    // even when we change/wrap them in colProto.
     addPoint: function () {
         return colProto.addPoint.apply(this, arguments);
     },
--- a/js/parts-more/ColumnRangeSeries.js
+++ b/js/parts-more/ColumnRangeSeries.js
@@ -131,6 +131,9 @@ seriesType('columnrange', 'arearange', merge(defaultPlotOptions.column, defaultP
     addPoint: function () {
         return colProto.addPoint.apply(this, arguments);
     },
+    correctStackLabels: function (point) {
+        return colProto.correctStackLabels.apply(this, arguments);
+    },
     crispCol: function () {
         return colProto.crispCol.apply(this, arguments);
     },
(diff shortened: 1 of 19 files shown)


js/parts-more/ColumnRangeSeries.js
```diff
@@ -138,17 +138,20 @@ seriesType('columnrange', 'arearange', merge(defaultPlotOptions.column, defaultP
...
         return colProto.drawTracker.apply(this, arguments);
     },
-    findMinColumnWidth: function () {
-        return colProto.findMinColumnWidth.apply(this, arguments);
+    getMaxColumnCount: function () {
+        return colProto.getMaxColumnCount.apply(this, arguments);
+    },
+    getMinColumnWidth: function () {
+        return colProto.getMinColumnWidth.apply(this, arguments);
     },
```
sebastianbochan: What is the reason of using function instead of `getMinColumnWidth: colProto.getMinColumnWidth` ?

js/parts/ColumnSeries.js
```diff
@@ -513,59 +547,81 @@ seriesType('column', 'line',
...
-                    });
-                }
-            });
-        }
+        chart.series.forEach(function (series) {
+            // Make sure that this series is the last
+            // of column/columnrange/bar.
+            if ((series.type === 'column' ||
+                series.type === 'columnrange' ||
+                series.type === 'bar') && series.points) {
```
sebastianbochan: Consider to use match with regexp like: 

```
series.type.match(/column|columnrange|bar/g)
```

js/parts/ColumnSeries.js
```diff
@@ -609,8 +665,11 @@ seriesType('column', 'line',
             (groupPadding +
                 colIndex * pointOffsetWidth -
                 (categoryWidth / 2)) * (reverseStacks ? -1 : 1);
-        if (chart.maxColumnCount &&
-            series.options.ignoreNulls === 'centered') {
+        // Handle pointXOffset when series.ignoreNulls: 'normal'
+        // so the points are aligned to the center of category.
+        if (chart.maxColumnCount && ignoreNulls &&
+            ignoreNulls !== 'evenlySpaced' &&
+            ignoreNulls !== 'fillSpace') {
```
sebastianbochan: Consider to use `ignoreNulls === 'center'` instead of two negations

js/parts/ColumnSeries.js
```diff
@@ -624,13 +683,42 @@ seriesType('column', 'line',
...
+     *
+     * @private
+     * @function Highcharts.seriesTypes.column#getMaxColumnCount
+     * @return {number}
+     */
+    getMaxColumnCount: function () {
+        var series = this;
+        var maxColumnCount = 0;
+        series.points.forEach(function (point) {
+            var columnCount = series.getColumnCount(point);
```
sebastianbochan: Missing verification if series is  "visible".

How about null points?

js/parts/ColumnSeries.js
pawelfus: Don't check `series.type` string, instead use `series instanceof seriesTypes.column` (which will be sooner or later replaced anyway). Other series types, like `errorbar` should share this functionality too.

raf18seb: @pawelfus the condition `series instanceof seriesTypes.column` is not true for `columnrange` series type which does not inherit from `column`.

js/parts/ColumnSeries.js
raf18seb: The purpose of this is to allow this implementation to work not only for `ignoreNulls: 'normal'`, but also for `ignoreNulls: true` and similar.
If you think it's no needed and it's not a good idea, then, please, let me know and I'll change it.

js/parts-more/ColumnRangeSeries.js
raf18seb: There is a difference between using `getMinColumnWidth: colProto.getMinColumnWidth` and function with `return colProto.getMinColumnWidth.apply(...)`.

- `getMinColumnWidth: colProto.getMinColumnWidth`
When we load core code and modules, then `columnrange.prototype.getMinColumnWidth` gets a value of `column.prototype.getMinColumnWidth`. Then, when we change (e.g. wrap) the `column.prototype.getMinColumnWidth`, the `columnrange`'s method is not changed.
- function with `return colProto.getMinColumnWidth.apply(...)`
Here, after we edited `getMinColumnWidth` on `column`'s prototype, the `columnrange` series will always fire the fresh, edited method.

js/parts/ColumnSeries.js
pawelfus: Aahh.. thanks for the explanation! FYI @TorsteinHonsi

js/parts/ColumnSeries.js
raf18seb: This verification is inside `getColumnCount` method.

TorsteinHonsi:
I would like to discuss the available options. I assume the existing functionality plus the `normal` setting will be most popular. Users have been requesting what we are now implementing as `normal`. I have checked three of our biggest competitors, and they all resolve null points in grouped columns the same way as we currently do.

The question is, do we need the `evenlySpaced` and `fillSpace` options? Are those frequently requested? If not, and if there are not common use cases for them, they would be a bloat, and it would be better not to add them. We can always add them later upon request.

js/parts-more/ColumnRangeSeries.js
sebastianbochan: Makes sense and I agree.

js/parts/ColumnSeries.js
sebastianbochan: Indeed, I missed that.

raf18seb:
> I would like to discuss the available options. I assume the existing functionality plus the `normal` setting will be most popular. Users have been requesting what we are now implementing as `normal`. I have checked three of our biggest competitors, and they all resolve null points in grouped columns the same way as we currently do.
> 
> The question is, do we need the `evenlySpaced` and `fillSpace` options? Are those frequently requested? If not, and if there are not common use cases for them, they would be a bloat, and it would be better not to add them. We can always add them later upon request.

For the last years, as a workaround, we were providing this code https://jsfiddle.net/BlackLabel/zvjw0h85 which behaves like `fillSpace` option and, from my experience, users were happy with that.

Furthermore, the `fillSpace` option was really easy to implement. The problem was with `normal` and `evenlySpaced` which required more new code. The `normal` option was the hardest to implement and is very similar to `evenlySpaced`. My point is that `normal` option requires logic that is already responsible for other implementations, so I don't see a reason why should we resign from them.

TorsteinHonsi:
Ok, thanks. Another thing, I notice there is some added code dealing with making other series dirty when one point is added/updated/removed. Basically, the other points at the same X position need to redraw in order to preserve the desired layout. However, this is not a new situation, we already have the same concept in stacks, where a change in one point affects the other series: http://jsfiddle.net/highcharts/on2qm6xz/. 

The question is, can we tap into the existing stacking logic instead of adding a new path og logic for making the other series dirty?

raf18seb:
You mean the logic from `H.Chart.redraw()`?

https://github.com/highcharts/highcharts/blob/fccbd474954486f6f4378e8379fb1cf543bfd7f3/ts/parts/Chart.ts#L854-L876

If yes, then IMHO my dirting logic has less performance impact, because dirties only necessary series, not all of them:
```
dirtyTheSeries: function (series, point): void {
            series.xAxis.series.forEach(function (
                otherSeries: Highcharts.Series
            ): void {
                if (
                    otherSeries.type === series.type &&
                    (otherSeries as any).hasValueInX(point, otherSeries)
                ) {
                    otherSeries.isDirty = true;
                }
            });
        }
```

We can consider merging both logics into one in the near future.
Or you still believe that the current logic from stacking is better?

TorsteinHonsi:
> If yes, then IMHO my dirting logic has less performance impact, because dirties only necessary series, not all of them

Yes, I think you are correct. The old logic goes dumbly over all series, regardless of series type and regardless of whether the series has actually changes. So a refactoring and alignment of functionality here is probably a good idea.

js/parts-more/ColumnRangeSeries.js
pawelfus: Could you add a comment with explanation there? Or is it somehow protected by a test, so no one will try to refactor this accidentally?

raf18seb: Writing a test for each method would be overkill. On the other hand, writing comments at each method would be overkill too. What is the best approach then?

pawelfus: I think one test to rule them all should be enough :) It would be nice to have all methods commented anyway

TorsteinHonsi:
> If yes, then IMHO my dirting logic has less performance impact, because dirties only necessary series, not all of them

> Yes, I think you are correct. The old logic goes dumbly over all series, regardless of series type and regardless of whether the series has actually changes. So a refactoring and alignment of functionality here is probably a good idea.

Did you fix this? I don't see any changes to the code block in `Chart.redraw()`.

raf18seb:
After further investigation, I realized that it will be more complicated to integrate `ignoreNulls` functionality with `stacking` - it is not only about dirting the required series. To merge both logics, I will have to go deeper into stacking, understand it and then refactor the code. It's gonna take a while and I will prioritize it if you want me to.

TorsteinHonsi:
> It's gonna take a while and I will prioritize it if you want me to.

Yes please. Stacking and grouping are related concepts. While the stack renders columns on top of each other for the same X value, grouping renders them side by side. It is important that we don't bloat our code base by implementing two so similar concepts independently.

raf18seb:
I apologize it took so much time. I thought it will be much more complex than it was. Is this what you meant (last commit)?

js/parts/Chart.js
```diff
@@ -433,6 +433,29 @@ extend(Chart.prototype, /** @lends Highcharts.Chart.prototype */ {
...
+            }
+        }
+        if (hasDirtyStacks) { // mark others as dirty
+            i = series.length;
+            while (i--) {
+                serie = series[i];
+                if (serie.options.stacking &&
+                    // These series have their own dirting logic
+                    !serie.type.match(/column|columnrange|bar/g)) {
+                    serie.isDirty = true;
```
sebastianbochan: Could we set the `serie.isDirty = true` flag in 444 line and avoid double while loop?

raf18seb: In short: this is the old logic for series other than column, columnrange, bar. We need to loop one more time separately to dirty all series, not only the ones found after first dirty.
@TorsteinHonsi will know more.

raf18seb:
@TorsteinHonsi FYI: it is ready to be reviewed.

TorsteinHonsi:
Continued in https://github.com/highcharts/highcharts/pull/12818

PR #12461, opened 2019-11-20
feature/11303-flame

### LINKED ISSUE #11303: Add Flame Chart in a demo/article section
Adding a Flame Chart in Highcharts - as a demo or a blog article.

Internal:
There is a possibility to use currently existing series, like a columnrange for achieving similar graph. There will also be a need for testing performance after creating the series.

KacperMadej:
Added flame series demo. Closes #11303.


samples/highcharts/demo/flame/demo.js
```diff
@@ -0,0 +1,1205 @@
...
+                },
+                enabled: true
+            }
+        }
+    }],
+    tooltip: {
+        headerFormat: "",
+        pointFormat: 'selfSize of <b>{point.name}</b> is <b>{point.value}</b>'
+    }
+}, function(chart) {
```
TorsteinHonsi: Add this functionality to the `load` event instead, so that we get a clean configuration of the chart instance. Then the event handler would apply to all charts in the page, but it should run a check before adding the buttons. For example check that the chart has a `flame` series, or that it has _only_ flame series. This event should be added within the same IIFE that adds the series type, so that it is an easily identifiable entity in the demo.

samples/highcharts/demo/flame/demo.js
```diff
@@ -0,0 +1,1205 @@
...
+        startOnTick: false,
+        endOnTick: false
+    }],
+    series: [{
+        type: 'flame',
+        data: data,
+        yAxis: 1,
+        xAxis: 1
+    }, {
+            visible: false,
```
TorsteinHonsi: Indenting is off several places in this file.

samples/highcharts/demo/flame/demo.details
```diff
@@ -0,0 +1,12 @@
...
+name: Flame chart
+authors:
+  - Kacper Madej
+js_wrap: b
+alt_text: >-
+  Highcharts flame chart JavaScript example graph shows
+  a chart rendering process as stacked tasks and subtasks.
+tags: []
+categories:
+  - More chart types
```
TorsteinHonsi: We need to add it to `index.htm` also, the new system is not in production yet.

samples/highcharts/demo/flame/demo.js
TorsteinHonsi: On second thought, I actually think this kind of toolbar should be implemented in HTML. The only reason for adding it in the chart SVG is to preserve it in the exported chart, but dynamic buttons don't really make sense in the exported chart, you can't click them anyway...

KacperMadej: Changed to HTML buttons format and style based on [this Gantt demo with buttons](https://www.highcharts.com/gantt/demo/interactive-gantt).

samples/highcharts/demo/flame/demo.js
KacperMadej: Fixed.

samples/highcharts/demo/flame/demo.details
KacperMadej: Added.

PR #12481, opened 2019-11-22
bugfix/12425-sunburst-drilldown-after-update

### LINKED ISSUE #12425: sunburst - series update breaks the chart
Doing an update on the series breaks chart functionality (drilldown stops work after it).

[Demo](http://jsfiddle.net/BlackLabel/nv38Lmx7/)
1. Update doesn't change colors,
2. After the update, drilldown stops working,
3. After the second click, drilldown works again,

#### Product version

Highcharts JS v7.2.1 (2019-10-31)

pawelfus:
Fixed #12425, updating sunburst series used to disable traversing.
___
It was general problem with events leaking. Multiple click events were added. Additionally removed leaking `setOptions` event.

### DIFF
--- a/js/modules/treemap.src.js
+++ b/js/modules/treemap.src.js
@@ -580,7 +580,7 @@ seriesType('treemap', 'scatter'
             this.colorAttribs = colorMapSeriesMixin.colorAttribs;
         }
         // Handle deprecated options.
-        addEvent(series, 'setOptions', function (event) {
+        series.eventsToUnbind.push(addEvent(series, 'setOptions', function (event) {
             var options = event.userOptions;
             if (defined(options.allowDrillToNode) &&
                 !defined(options.allowTraversingTree)) {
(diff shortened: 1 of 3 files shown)


TorsteinHonsi:
Pending @jon-a-nygaard 's approval

pawelfus:
Thanks @jon-a-nygaard!

> Update doesn't change colors

Could you share a demo? https://jsfiddle.net/BlackLabel/ojxrL7h4/ - color has changed to red.

> I also noticed that the data label animation is changed after update. It should fade in after updating positions.

I think it's unrelated to this PR, see v7.2.1 demo: https://jsfiddle.net/BlackLabel/ojxrL7h4/4/show  This is current animation:
drilldown-labels-animation-current

jon-a-nygaard:
> Could you share a demo?

https://jsfiddle.net/jon_a_nygaard/h058xjoz/1/

> I think it's unrelated to this PR, see v7.2.1

Yes, it is not caused by this PR, but I was thinking it may be related to this issue with sunburst and Series.update. Similar to the color issue reported in the reproduction steps in the issue.

pawelfus:
Thanks! 

It seems that changing from `contrast` color to any other color is broken in general: https://jsfiddle.net/BlackLabel/jd5L0s6z/

Changing from any color to any different color is fine: https://jsfiddle.net/BlackLabel/upabsm6w/

Regarding dataLabels animation, it seems some hacks are added there, probably another one needed? 

I wasn't aware of other issues, my PR resolves only memory leaks.

pawelfus:
Aah, I missed that! Created separate ticket: #12500 - thanks for pointing this out!

PR #12485, opened 2019-11-22
bugfix/12459-annotations-control-points

### LINKED ISSUE #12459: StockTools control points are not working
#### Expected behaviour
It should be possible to change the size of the added circle.

#### Actual behaviour
It is not possible to change the size of the added circle.

http://jsfiddle.net/BlackLabel/4m7pyx9f/

1. Add a circle
2. Try to change its size

#### Product version
Highstock

Caused probably by this commit: https://github.com/highcharts/highcharts/commit/d4101aadc

pawelfus:
Fixed #12459, annotation's control points were not draggable.

### DIFF
--- a/js/annotations/eventEmitterMixin.js
+++ b/js/annotations/eventEmitterMixin.js
@@ -23,6 +23,14 @@ var eventEmitterMixin = {
     addEvents: function () {
         var emitter = this;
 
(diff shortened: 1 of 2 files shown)


pawelfus:
Will check that FF test failure today.

PR #12503, opened 2019-11-26
docs/series-setstate

### LINKED ISSUE #12467: Highcharts.Series#setState: Please add 'normal', 'select' and 'inactive' to the official documentation
#### Expected behaviour
https://api.highcharts.com/class-reference/Highcharts.Series#setState mentions not only `'hover'` and `undefined`, but also `'normal'`, `'select'` and `'inactive'` 

Also, there is a mention to use this in combination with https://api.highcharts.com/class-reference/Highcharts.Series#update as mentioned here: https://github.com/highcharts/highcharts/issues/11750#issuecomment-524296009

#### Actual behaviour
https://api.highcharts.com/class-reference/Highcharts.Series#setState mentions only `'hover'` and `undefined` and there's no mention of using it in combination with https://api.highcharts.com/class-reference/Highcharts.Series#update as mentioned here: https://github.com/highcharts/highcharts/issues/11750#issuecomment-524296009

https://jsfiddle.net/afv6tmug/

#### Product version
Highcharts 7.2.1

bre1470:
- Close #12467, Mentioned all possible values for series state (and point).

### DIFF
--- a/js/parts/Interaction.js
+++ b/js/parts/Interaction.js
@@ -686,8 +686,9 @@ extend(Point.prototype, /** @lends Highcharts.Point.prototype */ {
      * @function Highcharts.Point#setState
      *
      * @param {Highcharts.PointStateValue|""} [state]
-     *        The new state, can be one of `''` (an empty string), `hover`,
-     *        `select` or `inactive`.
+     *        The new state, can be one of `'hover'`, `'select'`, `'inactive'`,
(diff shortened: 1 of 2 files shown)


swiss-chris:
Why is it the changes I asked for are not publicly visible ? 

https://api.highcharts.com/class-reference/Highcharts.Series#setState

bre1470:
The documentation is related to releases instead of the master branch. The change is expected to be visible in about two weeks.

PR #12508, opened 2019-11-27
enhancement/heatmap-custom-markers

Denyllon:
Adds ability to set marker’s symbol of heatmap points to a custom shapes defined within `SVGRenderer.prototype.symbols` object.

Custom shapes are set through `series.marker.symbol`, as it is in case of configuring other series which have markers.

To-Do:
- [x] Conversion to TypeScript
- [x] Add tests
- [x] Refactoring API docs

Demo: http://jsfiddle.net/BlackLabel/f58vwznd/


js/modules/tilemap.src.js
```diff
@@ -387,6 +387,19 @@ seriesType('tilemap', 'heatmap'
      */
     tileShape: 'hexagon'
 }, {
+    // Use drawPoints method from old heatmap implementation
+    // Consider standarizing heatmap and tilemap into more consistent form.
+    drawPoints: function () {
+        var _this = this;
+        // In styled mode, use CSS, otherwise the fill used in the style
+        // sheet will take precedence over the fill attribute.
+        var func = this.chart.styledMode ? 'css' : 'animate';
```
sebastianbochan: 1. Do we need to declare the `func` variable? In other parts of HC we use it inline.
2. Wy do you use `css` instead of `attr` method?

js/parts-map/HeatmapSeries.js
```diff
@@ -136,8 +136,13 @@ seriesType('heatmap', 'scatter',
         overflow: false,
         padding: 0 // #3837
     },
-    /** @ignore-option */
-    marker: null,
+    marker: {
+        symbol: 'rect',
+        /** @ignore-option */
+        radius: 0,
+        lineColor: UNDEFINED
```
sebastianbochan: use i.e `void 0` instead `UNDEFINED`

js/parts-map/HeatmapSeries.js
```diff
@@ -179,32 +184,122 @@ seriesType('heatmap', 'scatter',
...
+                .match(/url/), cellAttr = point.getCellAttributes(), shapeArgs = {
+                x: Math.min(cellAttr.x1, cellAttr.x2),
+                y: Math.min(cellAttr.y1, cellAttr.y2),
+                width: Math.max(Math.abs(cellAttr.x2 - cellAttr.x1), 0),
+                height: Math.max(Math.abs(cellAttr.y2 - cellAttr.y1), 0)
             };
+            // If marker shape is regular (symetric), find shorter
+            // cell's side.
+            if (hasRegularShape) {
+                dif = Math.abs(shapeArgs.width - shapeArgs.height);
```
sebastianbochan: `dif` is a little general, suggest `sizeDif`

js/parts-map/HeatmapSeries.js
```diff
@@ -179,32 +184,122 @@ seriesType('heatmap', 'scatter',
...
+        attr.stroke = (point.marker && point.marker.lineColor ||
+            seriesOptions.marker && seriesOptions.marker.lineColor ||
+            this.color);
+        if (state) {
+            stateOptions = merge(seriesOptions.states[state], (seriesOptions.marker &&
+                seriesOptions.marker.states)[state], point.options.states &&
+                point.options.states[state] || {});
+            brightness = stateOptions.brightness;
+            attr.fill =
+                stateOptions.color || (brightness !== UNDEFINED &&
```
sebastianbochan: Replace `UNDEFINED `with `void 0`

js/parts-map/HeatmapSeries.js
```diff
@@ -179,32 +184,122 @@ seriesType('heatmap', 'scatter',
...
+    pointAttribs: function (point, state) {
+        var attr = seriesTypes.scatter.prototype.pointAttribs
+            .call(this, point, state), seriesOptions = this.options || {}, stateOptions, brightness;
+        // Apply lineColor, or set it to default series color.
+        attr.stroke = (point.marker && point.marker.lineColor ||
+            seriesOptions.marker && seriesOptions.marker.lineColor ||
+            this.color);
+        if (state) {
+            stateOptions = merge(seriesOptions.states[state], (seriesOptions.marker &&
+                seriesOptions.marker.states)[state], point.options.states &&
```
sebastianbochan: What in case when `seriesOptions.marker.states` is not defined?

js/parts-map/HeatmapSeries.js
```diff
@@ -213,11 +308,16 @@ seriesType('heatmap', 'scatter',
     drawPoints: function () {
         // In styled mode, use CSS, otherwise the fill used in the style
         // sheet will take precedence over the fill attribute.
-        var func = this.chart.styledMode ? 'css' : 'animate';
-        seriesTypes.column.prototype.drawPoints.call(this);
-        this.points.forEach(function (point) {
-            point.graphic[func](this.colorAttribs(point));
-        }, this);
+        var func = this.chart.styledMode ? 'css' : 'animate', seriesMarkerOptions = this.options.marker || {};
```
sebastianbochan: Check the comment about `func` at the top.

js/parts-map/HeatmapSeries.js
```diff
@@ -263,9 +357,32 @@ seriesType('heatmap', 'scatter',
...
+     * @private
+     * @borrows Highcharts.seriesTypes.line.setOptions as Highcharts.seriesTypes.heatmap#setOptions
+     */
+    setOptions: function (itemOptions) {
+        var newOptions, propsMap = {
+            borderWidth: 'lineWidth',
+            borderColor: 'lineColor'
+        };
+        // Keep backward compatibility for borderWidth
+        for (var prop in propsMap) {
```
sebastianbochan: Use `forEach`

js/parts-map/HeatmapSeries.js
```diff
@@ -263,9 +357,32 @@ seriesType('heatmap', 'scatter',
...
+            if (H.defined(itemOptions[prop])) {
+                var markerOptions = {};
+                markerOptions[propsMap[prop]] =
+                    itemOptions[prop];
```
sebastianbochan: You can remove `markerOptions` because you use it only in merge.

js/parts-map/HeatmapSeries.js
```diff
@@ -263,9 +357,32 @@ seriesType('heatmap', 'scatter',
...
+                newOptions = merge(newOptions, itemOptions, {
+                    marker: merge(itemOptions.marker, markerOptions)
```
sebastianbochan: Replace: 
`marker: merge(itemOptions.marker, markerOptions)`
with
`marker: merge(itemOptions.marker, itemOptions[prop])`

ts/modules/tilemap.src.ts
```diff
@@ -790,7 +791,22 @@ seriesType<Highcharts.TilemapSeries>('tilemap', 'heatmap'
...
     }, { // Prototype functions
-
+        // Use drawPoints method from old heatmap implementation
+        // Consider standarizing heatmap and tilemap into more consistent form.
+        drawPoints: function (
+            this: Highcharts.TilemapSeries
+        ): void {
+            // In styled mode, use CSS, otherwise the fill used in the style
+            // sheet will take precedence over the fill attribute.
+            var func = this.chart.styledMode ? 'css' : 'animate';
```
sebastianbochan: Again about `func`

ts/parts-map/HeatmapSeries.ts
```diff
@@ -303,82 +352,143 @@ seriesType<Highcharts.HeatmapSeries>(
...
-                point.shapeArgs = {
-                    x: Math.min(x1, x2) + pointPadding,
-                    y: Math.min(y1, y2) + pointPadding,
-                    width: Math.max(Math.abs(x2 - x1) - pointPadding * 2, 0),
-                    height: Math.max(Math.abs(y2 - y1) - pointPadding * 2, 0)
+                var pointAttr,
+                    dif,
+                    hasImage =
+                        (point.marker && point.marker.symbol || shape || '')
+                            .match(/url/),
```
pawelfus: Why we have two different checks for `hasImage`? In `markerAttribs` we have: 
> point.hasImage = (symbol && symbol.indexOf('url') === 0) || false;

Can we perhaps use `point.hasImage` here? And unify logic in `markerAttribs`?

ts/parts-map/HeatmapSeries.ts
```diff
@@ -293,6 +320,28 @@ seriesType<Highcharts.HeatmapSeries>(
...
+        /**
+         * @private
+         * @function Highcharts.seriesTypes.heatmap#setClip
+         * @return {void}
+         */
+        setClip: function (
+            this: Highcharts.HeatmapSeries
+        ): void {
+            Series.prototype.setClip.apply(this, arguments);
+            if (this.options.clip !== false && this.markerGroup) {
```
pawelfus: Hm.. isn't `markerGroup` always defined once you made the transition to markers?

ts/parts-map/HeatmapSeries.ts
```diff
@@ -293,6 +320,28 @@ seriesType<Highcharts.HeatmapSeries>(
...
+                this.markerGroup
+                    .clip(this.chart.clipRect);
```
pawelfus: I think `this.sharedClipRect` should be used (if defined). Otherwise x/yAxis top/left/height/width can clip incorrectly?

ts/parts-map/HeatmapSeries.ts
```diff
@@ -293,6 +320,28 @@ seriesType<Highcharts.HeatmapSeries>(
...
             // general point range
             this.yAxis.axisPointRange = options.rowsize || 1;
+
+            // Bind new symbol names
+            extend(symbols, {
+                ellipse: symbols.circle,
+                rect: symbols.square
+            });
+        },
+        getSymbol: seriesTypes.scatter.prototype.getSymbol,
```
pawelfus: Why `seriesTypes.scatter`, not `Series` ?

ts/parts-map/HeatmapSeries.ts
```diff
@@ -237,8 +258,14 @@ seriesType<Highcharts.HeatmapSeries>(
             padding: 0 // #3837
         },
 
-        /** @ignore-option */
-        marker: null as any,
+        marker: {
+            symbol: 'rect',
+            /** @ignore-option */
+            radius: 0,
+            lineColor: UNDEFINED
```
pawelfus: Replace with `void 0`

js/parts-map/HeatmapSeries.js
pawelfus: Replace with `typeof`*

js/parts-map/HeatmapSeries.js
pawelfus: `objectEach`*

js/modules/tilemap.src.js
Denyllon: This method is only moved here the old heatmap implementation, in order to keep the way of how the tilemap has worked before, but agree. I removed that variable. Regarding `css` call, you can find the explanation in the comment above the code.

js/parts-map/HeatmapSeries.js
Denyllon: They're always inherited from parent series `defaultOptions`

ts/parts-map/HeatmapSeries.ts
```diff
@@ -303,82 +352,143 @@ seriesType<Highcharts.HeatmapSeries>(
...
+            state?: string
+        ): Highcharts.SVGAttributes {
+            var attr = seriesTypes.scatter.prototype.pointAttribs
+                    .call(this, point, state),
+                seriesOptions = this.options || {},
+                stateOptions,
+                brightness;
+
+            // Apply lineColor, or set it to default series color.
+            attr.stroke = (point && point.marker && point.marker.lineColor ||
```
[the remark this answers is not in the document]
Denyllon: Yes it is, but in other way than it was calculated before refactoring. So far, it was taking the color from `series.color` in the first place, or if it was not defined, then from `point.color` (column's prototype `pointAttribs` method). Currently used prototyped method checks above properties in the other way around, so it was required to keep it behave as before.

ts/parts-map/HeatmapSeries.ts
Denyllon: It was done because of TypeScript definitions. The `Series.markerGroup` is optional, so can be `undefined`. I assumed that would be better to check `this.markerGroup` exists instead of casting it `as any`.
If it is not correct, I can also override the `markerGroup` property within `HeatmapSeries` interface, and make it required. What do you think?

ts/parts-map/HeatmapSeries.ts
Denyllon: Agree. From now on, setting the `point.hasImage` just once within `series.translate()`, and only read that property in `markerAttribs()` function.

ts/parts-map/HeatmapSeries.ts
```diff
@@ -389,15 +499,20 @@ seriesType<Highcharts.HeatmapSeries>(
...
-            }, this);
+            var func = this.chart.styledMode ? 'css' : 'animate',
+                seriesMarkerOptions = this.options.marker || {};
+
+            if (seriesMarkerOptions.enabled || this._hasPointMarkers) {
+                seriesTypes.scatter.prototype.drawPoints.call(this);
+                this.points.forEach(function (
+                    point: Highcharts.HeatmapPoint
+                ): void {
+                    if (!point.isNull) {
```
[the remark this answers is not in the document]
Denyllon: Currently, null points are rendered as before.

Denyllon:
@activist Unfortunately visual tests are still not passing after merging master into this branch. The difference is noticeable in comparison between reference and candidate gif files, but I'm not able to reproduce them in JS fiddle. I've took the `highcharts/plotoptions/series-nullformat` unit-test, recreate it with reference to github's master and my branch, and both samples looks the same. Here are the examples:

Github master: http://jsfiddle.net/BlackLabel/3jwxnz42/
Last commit of this branch: http://jsfiddle.net/BlackLabel/gn6pc2eb/
 
While debugging I can see that all cells are having the same dimensions within `shapeArgs` object. However, on the code generated in visual tests they're differ with 1 pixel of width, and I'm not sure why that discrepancy occurs. Could you take a look on it, and try to tell me what can cause this type of problem and where to try to look for its source?

ts/parts-map/HeatmapSeries.ts
pawelfus: It's ok 👍

activist:
As for the visual tests (just to clarify) the intent is that is it should help identify changes and it is not something that **have** to pass. I.e it can be merged even though the visual review hasn't been done. 

But, I agree with you that there likely is something wrong here. For instance when looking at the (textual) diff there are differences like `aria-hidden="false"` in the reference, while in the candidate it is `aria-hidden="true"`, as well as `<rect x="385" y="86" width="31"height="29" fill="rgb(231,213,226)" ` vs `<pathfill="rgb(240,234,243)" d="M 543 29 L 575 29 575 57 543 57 Z"` which to me indicates that there are differences as to how the tests (reference creation vs comparison).

This is likely caused by some sort of pollution of the global scope or differences in how the test runs are set up. I will look into this soon. 

@TorsteinHonsi  do you have any clue as to what may cause such differences as listed above?

Denyllon:
> as well as <rect x="385" y="86" width="31"height="29" fill="rgb(231,213,226)" vs <pathfill="rgb(240,234,243)" d="M 543 29 L 575 29 575 57 543 57 Z" which to me indicates that there are differences as to how the tests (reference creation vs comparison).
Those changes are caused by the fact that heatmap have had a points generated as `<rect>`'s before refactoring. For now, all points are always `<path>`'s, regardless a shape which is set by user.

activist:
@Denyllon  After carefully looking into your JSFiddle samples I actually find that they do differ for that particular sample:

Open in two different tabs, and zoom to max (500% depending on how it looks on your screen):
http://jsfiddle.net/BlackLabel/3jwxnz42/show
http://jsfiddle.net/BlackLabel/gn6pc2eb/show

If you toggle back and forth between the tabs you will indeed see that the center vertical border between the "tiles" is slightly more to the left on the latter one.

Screenshots:
jsfiddle net_BlackLabel_gn6pc2eb_show
jsfiddle net_BlackLabel_3jwxnz42_show

TorsteinHonsi:
The `aria` stuff wouldn't result in a visual diff, so we can ignore that. It is probably because there is some a11y change in the master that has not been merged into this branch. 

The other changes are related to pixel-precision of shapes it seems. I would start by merging the master in. Maybe something is changed there. 

It is a general problem that the side-by-side comparison in utils does not work exactly like the CI. We should strive for consistency.

Denyllon:
In short: we decided to approve all visual tests differences, because they were not be reproducible (or noticeable) locally, even after zooming in both overlapping versions of charts.

samples/unit-tests/series-heatmap/marker/demo.js
```diff
@@ -0,0 +1,188 @@
...
+            lineColor: 'rgba(0, 255, 0, 1)'
+        }
+    });
+
+    var marker = heatmap.points[0].graphic;
+
+    assert.strictEqual(
+        marker.element.getAttribute('stroke'),
+        'rgba(0, 255, 0, 1)',
+        'Marker\'s line color is set.'
```
TorsteinHonsi: Use the word "should" in all descriptions. It makes it easy to understand what to _expect_ versus what actually happens. Applies to the other asserts as well.

samples/unit-tests/series-heatmap/marker/demo.js
```diff
@@ -0,0 +1,188 @@
...
+            height: 50
+        }
+    });
+
+    var bBox = heatmap.points[0].graphic.getBBox();
+
+    assert.strictEqual(
+        bBox.width - heatmap.options.marker.lineWidth === 50 &&
+        bBox.height - heatmap.options.marker.lineWidth === 50,
+        true,
```
TorsteinHonsi: Instead of a complex condition that strictly equals `true`, split it up in to `strictEqual` assertions that tests for the right number. For truthy/falsy conditions we have `assert.ok`, but in this case it's better to split it up.

samples/unit-tests/series-heatmap/marker/demo.js
```diff
@@ -0,0 +1,188 @@
...
+            lineColor: 'red',
+            lineWidth: 5
+        }
+    });
+
+    bBox = point.graphic.getBBox();
+
+    assert.strictEqual(
+        bBox.width - point.options.marker.lineWidth === 20 &&
+        bBox.height - point.options.marker.lineWidth === 20,
```
TorsteinHonsi: As above

ts/parts-map/ColorSeriesMixin.ts
```diff
@@ -106,7 +107,7 @@ H.colorSeriesMixin = {
 
             color = point.options.color ||
                 (
-                    point.isNull ?
+                    point.isNull || point.value === null ?
```
TorsteinHonsi: This addition shouldn't be necessary. When `point.value` is `null`, `point.isNull` should always be true. See if this can be handled in the `isValid` function for the given Point class.

ts/parts-map/HeatmapSeries.ts
```diff
@@ -456,6 +799,42 @@ seriesType<Highcharts.HeatmapSeries>(
...
+        applyOptions: function (
+            this: Highcharts.HeatmapPoint,
+            options: Highcharts.HeatmapPointOptions,
+            x?: number
+        ): Highcharts.HeatmapPoint {
+            var point = H.Point.prototype
+                .applyOptions.call(this, options, x) as any;
+
+            point.formatPrefix =
+                point.isNull || point.value === null ?
```
TorsteinHonsi: Shouldn't be necessary to check for `point.value === null`

ts/parts-map/HeatmapSeries.ts
```diff
@@ -456,6 +799,42 @@ seriesType<Highcharts.HeatmapSeries>(
...
+         * a null point
+         * @private
+         * @function Highcharts.HeatmapPoint.isValid
+         * @return {boolean}
+         */
+        isValid: function (this: Highcharts.HeatmapPoint): boolean {
+            // undefined is allowed
+            return (
+                this.value !== Infinity &&
+                this.value !== -Infinity
```
TorsteinHonsi: This function should also check for `this.value !== null`, then we can use `point.isNull` safely above.

ts/parts-map/HeatmapSeries.ts
```diff
@@ -435,11 +739,50 @@ seriesType<Highcharts.HeatmapSeries>(
 
             // Get the extremes from the y data
             Series.prototype.getExtremes.call(this);
+        },
+
+        /**
+         * @private
+         * @borrows Highcharts.seriesTypes.line.setOptions as Highcharts.seriesTypes.heatmap#setOptions
+         */
+        setOptions: function (
```
TorsteinHonsi: Mutating the options can be considered bad practice. Instead, this functionality should probably be moved to the `pointAttribs`, and the old options be applied directly to the attributes.

ts/parts-map/ColorSeriesMixin.ts
Denyllon: Moving above condition into `isValid` function does not give any effect, because if `point.value` will be null, then the point graphic (marker) will not be rendered at all, as it is in case of scatter series on which a heatmap is based on since refactoring.

Before implementing the ability to customise heatmap markers, this series used `drawPoints` method from column series, where the flag `point.isNull` is not checked during point rendering process, but only whether the `point.y` is not equal to `null`. That’s the reason of why heatmap points with` point.value = null` had graphics before, and most probably we should keep it to work in the same way in the new implementation.

We can either override `Series.drawPoints` function and move a whole logic into the new function, but with only one change in if statement (what does not look as a good way out), or keep the current implementation which does work correctly, but can be a bit difficult to understand at first glance.
We can also consider other solutions, so if you would have any ideas or suggestions of how to resolve it, will be great to discuss about it.

ts/parts-map/HeatmapSeries.ts
TorsteinHonsi: Still an issue as of March 13.

PR #12538, opened 2019-12-02
Bugfix/10263 heatmap pointplacement between

### LINKED ISSUE #10263: Heatmap uses wrong min/max values on datetime x-axis
#### Expected behaviour
Setting the min and max values on the x-axis should properly be applied to the extremes.

#### Actual behaviour
Setting min/max on heatmap seems to always add an additionial/hidden tick. I also tried both setting the min/max properties of the x-axis and using the setExtremes() method but neither seem to work

I am uncertain if the other options in my example contribute to the problem, but this is the configuration I need for my use case:
https://jsfiddle.net/lanimelrok/4vw2stjn/

#### Product version
Highstock 7.0.3

jon-a-nygaard:
Fixed #10263, axis had wrong min and max values when Heatmap series had only a single point.

---
# Description
When `xAxis` has only a single tick, the min and max values gets adjusted to display the point better.
Not sure if this is the best way to solve the issue, but I added a filter to avoid this adjustment if the axis has a heatmap series with `pointPlacement: 'between'`.

While working on this issue I also did some refactor of `HeatmapSeries.translate` and `Series.pointPlacementToXValue` to fix some type issues.

# Example(s)
- [Demo of issue](https://jsfiddle.net/jon_a_nygaard/az4032ye/)
- [Demo of issue with fix applied](https://jsfiddle.net/jon_a_nygaard/ax9n6fb8/)

# Related issue(s)
- Closes #10263


ts/parts/StockChart.ts
```diff
@@ -349,24 +349,17 @@ addEvent(Series, 'setOptions', function (
...
 
-        if (is('column') || is('columnrange')) {
+        if (this.is('column') || this.is('columnrange')) {
             overrides = {
                 borderWidth: 0,
                 shadow: false
             };
 
-        } else if (is('line') && !is('scatter') && !is('sma')) {
+        } else if (this.is('line') && !this.is('scatter') && !this.is('sma')) {
```
KacperMadej: What is the point of checking for `line` if all series types inherit from it?

TorsteinHonsi: Right

PR #12541, opened 2019-12-02
Added 1 as a default value to zIndex, in no data to display

### LINKED ISSUE #12343: Gauge chart and No data to display
#### Expected behaviour
When a gauge chart has no data to display, the defined noData message should be visible. 

#### Actual behaviour
The message is rendered but behind the main pane, so we can't see it.

https://jsfiddle.net/nq0hos8m/

#### Product version
Highcharts 6.2, but same results with Highcharts 7.2.1

#### Affected browser(s)
All

marvin19:
Fixed #12343, `noData` message was not visible with gauge series. Added 1 as default `zIndex` to render the message in front.

---
# Examples
- [Demo of issue](https://jsfiddle.net/nq0hos8m/)
- [Demo of issue with fix applied](https://jsfiddle.net/kxnuh3j0/)
# Related issues
- Closes #12343

### DIFF
--- a/samples/unit-tests/no-data/members/demo.js
+++ b/samples/unit-tests/no-data/members/demo.js
@@ -1,0 +1,15 @@
+QUnit.test("defaultOptions", assert => {
+    const {
+        noData: {
+            attr: {
+                zIndex
+            }
+        }
+    } = Highcharts.getOptions();
+
+    assert.equal(
+        zIndex,
+        1,
+        "Default z index should be 1 (#12343)"
+    );
+});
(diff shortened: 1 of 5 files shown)


samples/unit-tests/no-data/members/demo.js
```diff
@@ -0,0 +1,15 @@
...
+            attr: {
+                zIndex
+            }
+        }
+    } = Highcharts.getOptions();
+
+    assert.equal(
+        zIndex,
+        1,
+        "should have defaultOptions.noData.attr.zIndex to equal 1 (#12343)"
```
TorsteinHonsi: Better write as a natural sentence with upper case first: "Default z index should be 1 (#12343)"

PR #12552, opened 2019-12-03
Fixed #12462, replacing slash in filename

### LINKED ISSUE #12462: Filenames that contain "/" are missing all content before the "/" and the "/" is not replaced with a different, safe, character
#### Expected behaviour
When exporting a chart with a custom filename that contains a "/" character, the "/" is replaced with a safe character such as "-" or "_". When the chart is exported, the filename contains all of the characters, minus the "/".

Example, exporting a chart with a custom filename of "Medical/Dental Office" should export as "Medical-Dental Office" or "Medical_Dental Office". 

#### Actual behaviour
When you export a chart with a custom filename such as "Medical/Dental Office" the filename that is generated is "Dental Office" and you lose all the characters that are before the "/". 

#### Notes
We are dynamically generating filenames based on user submitted names, so we cannot guarantee that the user will not place a "/" in the name. 

I have noticed that other frameworks used to export charts handle replacing "/" characters automatically. Tested in html2pdf.js and Filesaver.js

Export the chart in this demo and compare the filename exported versus the filename set in the code. 
demo: https://jsfiddle.net/2wrxe7v3/1/

#### Product version
highcharts version 7.2.0

#### Affected browser(s)
Chrome (Mac and Windows)

goransle:
Fixed #12462, filename with forward slash (`/`) not being replaced with safe character when exported. Now replaced with `-`.

# Description
Replaces `/` in the filename set in `exportChart` and `getFilename` in the export module.

# Example(s)
* [Demo of issue](https://jsfiddle.net/2wrxe7v3/1/)
* [Demo of issue with fix applied](https://jsfiddle.net/goransle/4jos3bkc/7/)
## Related issue(s)
Closes #12462

### DIFF
--- a/js/modules/exporting.src.js
+++ b/js/modules/exporting.src.js
@@ -1153,7 +1153,7 @@ extend(Chart.prototype, /** @lends Highcharts.Chart.prototype */ {
     getFilename: function () {
         var s = this.userOptions.title && this.userOptions.title.text, filename = this.options.exporting.filename;
         if (filename) {
-            return filename;
+            return filename.replace(/\//g, '-');
         }
         if (typeof s === 'string') {
             filename = s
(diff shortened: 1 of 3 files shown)


TorsteinHonsi:
Awaiting CI run, otherwise approved 👍

PR #12563, opened 2019-12-04
Fixed missing drag markers when loading highcharts-3d module.

### LINKED ISSUE #12484: Draggable point + 3d = no drag markers for columns
#### Behaviour
If you will load `highcharts-3d` and `draggable-points` module then columns will not show the drag markers on hover making them impossible to drag.

This problem also affects Gantt charts if both "draggable" and "3d" files are loaded.

https://jsfiddle.net/BlackLabel/9zpL7tdy/

Hover over a column and notice the lack of the drag marker.
(It is sometimes possible to get the drag markers to show on columns after dragging the line in the demo.)

#### Product version
7.2.1

#### Affected browser(s)
Tested on Chrome v78 / Win10

marvin19:
Fixed #12484, drag markers for columns were not showing when `highcharts-3d` and `draggable-points` were loaded.

---
# Examples 
- [Demo of issue](https://jsfiddle.net/BlackLabel/9zpL7tdy/)
- [Demo of issue with fix applied](https://jsfiddle.net/k64y5ute/)
# Related issues
- Closes #12484

### DIFF
--- a/ts/modules/draggable-points.src.ts
+++ b/ts/modules/draggable-points.src.ts
@@ -2725,7 +2725,8 @@ function getFirstProp<T>(obj: Highcharts.Dictionary<T>): (T|undefined) {
 function mouseOver(point: Highcharts.Point): void {
     var series = point.series,
         chart = series && series.chart,
-        dragDropData = chart && chart.dragDropData;
+        dragDropData = chart && chart.dragDropData,
+        is3d = chart && chart.is3d && chart.is3d();
 
     if (
         chart &&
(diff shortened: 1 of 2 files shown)


ts/modules/draggable-points.src.ts
```diff
@@ -2726,6 +2726,8 @@ function mouseOver(point: Highcharts.Point): void {
     var series = point.series,
         chart = series && series.chart,
         dragDropData = chart && chart.dragDropData;
+    const is3DEnabled = chart && chart.options && chart.options.chart &&
```
TorsteinHonsi: Instead of checking the options structure, use the test designed for this: `is3d = chart && chart.is3d && chart.is3d()`.

TorsteinHonsi: And while we're at it, we should consolidate the code to either use only 'const' or only 'var' in the same function block, for the sake of consistency.

PR #12568, opened 2019-12-04
Fix zoom issues in treemap

### LINKED ISSUE #12488: Area selection error treemap
#### Expected behaviour
The area that is selected should increase

#### Actual behaviour
The area is exactly the opposite.

image-2019-10-04-17-37-49-565

#### Product version
Highcharts 7.2

#### Affected browser(s)
In all browsers

jon-a-nygaard:
Fixed #12488, selection zoomed to wrong area in the Treemap series.

---
# Description
The treemap series used `csvCoord=false` when calling `Axis.translate` for the y values, while `Axis.toValue` that is used to find new min and max for `Chart.zoom` assumes `csvCoord=true` for the y axis.
Solved by setting `csvCoord=true` when translating y values, and reversing the y-values before passing them into translate.

# Example(s)
- [Demo of issue](https://jsfiddle.net/jon_a_nygaard/s4xqouf2/)
- [Demo of issue with fix applied](https://jsfiddle.net/jon_a_nygaard/v529z0hs/)

# Related issue(s)
- Closes #12488

### DIFF
--- a/ts/modules/treemap.src.ts
+++ b/ts/modules/treemap.src.ts
@@ -1188,71 +1188,34 @@ seriesType<Highcharts.TreemapSeries>(
             });
         },
         setPointValues: function (this: Highcharts.TreemapSeries): void {
-            var series = this,
-                xAxis = series.xAxis,
-                yAxis = series.yAxis;
+            const series = this;
+            const { points, xAxis, yAxis } = series;
+            const styledMode = series.chart.styledMode;
 
-            series.points.forEach(function (
-                point: Highcharts.TreemapPoint
-            ): void {
-                var node = point.node,
-                    values = node.pointValues,
-                    x1,
-                    x2,
-                    y1,
-                    y2,
-                    crispCorr = 0;
+            // Get the crisp correction in classic mode. For this to work in
+            // styled mode, we would need to first add the shape (without x,
+            // y, width and height), then read the rendered stroke width
+            // using point.graphic.strokeWidth(), then modify and apply the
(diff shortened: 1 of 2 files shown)


ts/modules/treemap.src.ts
```diff
@@ -1188,71 +1188,40 @@ seriesType<Highcharts.TreemapSeries>(
...
-                    crispCorr = (
-                        (series.pointAttribs(point)['stroke-width'] || 0) % 2
-                    ) / 2;
-                }
+            const translateX = (value: number): number => Math.round(
+                xAxis.translate(value, false, false, false, true) as any
+            );
+            const translateY = (value: number): number => Math.round(
+                // Flip y-values so that 0 is top of the axis
+                yAxis.translate(100 - value, false, true, false, true) as any
```
pawelfus: Won't be better to use `xAxis.toPixels(value, true)` and `yAxis.toPixels(100 - value, true)` instead of defining another `translateX(value)` and `translateY(value)` method?

jon-a-nygaard: Thanks, good catch. `Axis.toPixels` is a better choice.
Was a bit unsure if I still should keep the local methods to avoid repeating rounding and the y-values logic, but I ended up removing the functions altogether.

jon-a-nygaard: Thanks, good catch. Axis.toPixels is a better choice.
Was a bit unsure if I still should keep the local methods to avoid repeating rounding and the y-values logic, but I ended up removing the functions altogether.

PR #12570, opened 2019-12-04
bugfix/7255-print-cancel-safari

### LINKED ISSUE #7255: Print is not working correctly on Highcharts.com demo page in Safari browser
#### Expected behaviour
Print functionality should work the same on all browsers.

#### Actual behaviour
In Safari browser printing chart is not working correctly and page refresh is needed.

https://www.highcharts.com/demo/line-basic

1. Click on print button
2. Click on cancel 

screen shot 2017-10-11 at 14 26 41

3. Click again on print button

The alert message is shown:
screen shot 2017-10-11 at 14 26 51

After clicking on 'print' chart is not displayed correctly:
screen shot 2017-10-11 at 14 26 57

#### Affected browser(s)
Safari (10.1.2)

pawelfus:
Fixed #7255, trying to printing chart again, after cancelling in Safari, elements other than chart were not hidden.
___
### Solution (Safari only)
In Safari (I assume `Highcharts.isSafari` is safe to use), instead of `setTimeout` for hiding and restoring DOM elements for printing, `window.matchMedia('print')` will be used. That required a bit refactoring in `Chart.print` method.

### Other browsers:
- Chrome - can use `matchMedia` too, can be replaced
- FireFox - doesn't work because of this bug: https://bugzilla.mozilla.org/show_bug.cgi?id=774398 
- IE 10-11/Edge - doesn't work, haven't found the reason for that
- IE 6-9 - don't support `matchMedia`

### Notes
- Solution pollutes `Highcharts` namespace with: `beforePrint`, `afterPrint`, `moveContainers` methods and `printingChart` property. I'm open for suggestions here.
- No test, I'm afraid karma won't support such test. Needs to be verified manually.

### Samples
- without fix: https://jsfiddle.net/BlackLabel/bj9hd6mg/
- with fix: https://jsfiddle.net/BlackLabel/kfpatbs5/

Steps (in Safari): Try to print the chart, but cancel it. Try to print again, so confirm box will show up. Click "print" in the box. in Safari "Test" elements will be visible in the print preview.

### DIFF
--- a/js/modules/exporting.src.js
+++ b/js/modules/exporting.src.js
@@ -933,116 +933,16 @@ H.post = function (url, data, formAttributes) {
     // clean up
     discardElement(form);
 };
-/**
-* Move the chart container(s) to another div.
-*
-* @function Highcharts#moveContainers
-*
-* @private
-*
-* @param {Highcharts.Chart} chart
-*        Move from chart
-* @param {Highcharts.HTMLDOMElement} moveTo
-*        Move target
-* @return {void}
-*/
-H.moveContainers = function (chart, moveTo) {
-    (chart.fixedDiv ? // When scrollablePlotArea is active (#9533)
-        [chart.fixedDiv, chart.scrollingContainer] :
-        [chart.container]).forEach(function (div) {
-        moveTo.appendChild(div);
-    });
-};
-/**
-* Prepare chart and document before printing a chart.
-*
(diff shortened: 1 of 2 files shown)


pawelfus:
Good point @TorsteinHonsi - refactored. No need for all methods to make them avail in the Highcharts namespace.

PR #12571, opened 2019-12-04
Tools: Add publishing to npm next tag in nightly build

activist:
This adds publishing of nightly builds to npm using the next tag on the highcharts package on npm (meaning not a separate highcharts-nightly package on npm).


activist:
Solved via https://github.com/highcharts/highcharts/pull/12587

PR #12574, opened 2019-12-05
ts/modules-annotations

bre1470:
- Migrating annotations to TypeScript.


ts/annotations/MockPoint.ts
```diff
@@ -0,0 +1,541 @@
...
+ */
+declare global {
+    namespace Highcharts {
+        interface AnnotationMockLabelOptionsObject {
+            x: (number|null);
+            y: (number|null);
+            point: AnnotationMockPoint;
+        }
+        class AnnotationMockPoint {
+            public static fromPoint(point: AnnotationPoint): AnnotationMockPoint;
```
pawelfus: `point: AnnotationPoint`: shouldn't be `point: Highcharts.Point` ? This static methods creates a new MockPoint from `series.points[index]` point.

ts/annotations/MockPoint.ts
```diff
@@ -0,0 +1,541 @@
...
+        }
+        interface AnnotationMockSeries {
+            chart: AnnotationChart;
+            getPlotBox: Series['getPlotBox'];
+            xAxis?: (Axis|null);
+            yAxis?: (Axis|null);
+            visible: boolean;
+        }
+        interface AnnotationPoint {
+            mock?: undefined;
```
pawelfus: Actually `AnnotationPoint` is just `Highcharts.Point`. The way to distinguish between `Point` and `MockPoint` is that `MockPoint.mock` is defined and set to true. Should we then replace `AnnotationPoint` with `Point`?

ts/annotations/annotations.src.ts
```diff
@@ -0,0 +1,1608 @@
...
+            symbol: string;
+            visible: boolean;
+            width: number;
+        }
+        interface AnnotationControlPointPositioner {
+            (target: AnnotationControllable): PositionObject;
+        }
+        type AnnotationDraggableValue = (''|'x'|'y'|'xy');
+        type AnnotationLabelType = AnnotationControllableLabel;
+        type AnnotationShapeType = AnnotationControllable;
```
pawelfus: I know it's not finished, just a heads-up: `AnnotationShapeType` can be probably one of many (`ControllableRect`, `ControllablePath`, `ControllableCircle` etc.)

ts/annotations/annotations.src.ts
```diff
@@ -0,0 +1,1608 @@
...
+        interface AnnotationMockPointOptionsObject {
+            x: number;
+            y: number;
+            xAxis?: (number|string|Axis|null);
+            yAxis?: (number|string|Axis|null);
+        }
+        interface AnnotationPoint extends Point {
+            series: AnnotationSeries;
```
pawelfus: Ah, here is the definition for `AnnotationPoint`. One of other comments (about `mock`) doesn't matter, sorry.

ts/annotations/annotations.src.ts
```diff
@@ -0,0 +1,1608 @@
...
+            text?: string;
+            useHTML: boolean;
+            verticalAlign: VerticalAlignValue;
+            x: number;
+            y: number;
+        }
+        interface AnnotationsLabelsOptions extends AnnotationsLabelOptions {
+            point?: (string|AnnotationMockPointOptionsObject);
+        }
+        interface AnnotationsOptions extends AnnotationControllableOptionsObject {
```
pawelfus: I'm not sure about this one. Are really `AnnotationsOptions` extended by `AnnotationControllableOptionsObject`? I trust you here, so could you double check this?

ts/annotations/annotations.src.ts
```diff
@@ -0,0 +1,1608 @@
...
+            controlPointOptions: AnnotationControlPointOptionsObject;
+            draggable: AnnotationDraggableValue;
+            events: AnnotationsEventsOptions;
+            id?: (number|string);
+            labelOptions: AnnotationsLabelOptions;
+            labels: Array<AnnotationsLabelsOptions>;
+            shapeOptions: AnnotationsShapeOptions;
+            shapes: Array<AnnotationsShapesOptions>;
+            typeOptions: AnnotationsTypeOptions;
```
pawelfus: I think these `typeOptions` should be defined in specific files (e.g. Fibonacci, ElliottWave) etc. Every type can provide different set of options.

ts/annotations/controllable/ControllableLabel.ts
```diff
@@ -0,0 +1,561 @@
...
+ */
+declare global
+{
+    namespace Highcharts
+    {
+        interface AnnotationAlignObject extends AlignObject {
+            height?: number;
+            width?: number;
+        }
+        class AnnotationControllableLabel implements AnnotationControllable {
```
pawelfus: Wouldn't be better to use `extends` ? There will be more classes that can use `AnnotationControllable` base and overwrite it's methods.

ts/annotations/eventEmitterMixin.ts
```diff
@@ -0,0 +1,349 @@
...
+ * */
+
+import H from '../parts/Globals.js';
+
+/**
+ * Internal types.
+ */
+declare global {
+    namespace Highcharts {
+        interface AnnotationEventEmitter extends AnnotationEventEmitterMixin {
```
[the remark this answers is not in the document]
bre1470: The mixin describes only the export type, while the emitter describes the instance for mixin functions. Not everything of the emitter is part of the mixin.

ts/annotations/MockPoint.ts
bre1470: `AnnotationPoint` is only an internal interface to describe the `Point` instance during annotation processing, so that `any` casts can be avoided. Same for `AnnotationChart` and `AnnotationSeries`.

ts/annotations/MockPoint.ts
bre1470: See previous comment.

ts/annotations/annotations.src.ts
bre1470: Thanks! 👍

ts/annotations/annotations.src.ts
bre1470: See my second comment.

ts/annotations/annotations.src.ts
bre1470: Well yesterday it was the conclusion, as all controllables take AnnotationsOptions, but I really do not know. It is just the typing that leads to this conclusion.

ts/annotations/controllable/ControllableLabel.ts
bre1470: AnnotationControllable is not a class - just a mixin object - you can not extend a class from an interface.

ts/annotations/eventEmitterMixin.ts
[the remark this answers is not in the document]
pawelfus: I see, thanks!

ts/annotations/annotations.src.ts
pawelfus: Ah, right, this `x`/`y` options that actually works. Thanks!

ts/annotations/controllable/ControllableLabel.ts
pawelfus: Ah, okay!

ts/annotations/MockPoint.ts
pawelfus: Yes, realized that after writing this comment, thanks!

ts/annotations/annotations.src.ts
bre1470: As you can merge interfaces but not property typing, I kept the generic type for inheritance purpose.

ts/annotations/navigationBindings.ts
```diff
@@ -0,0 +1,1626 @@
...
+            public initEvents(): void;
+            public initUpdate(): void;
+            public removeEvents(): void;
+            public update(options: NavigationOptions): void
+        }
+        interface NavigationBindingsButtonEventsObject {
+            button: HTMLDOMElement;
+            events: NavigationBindingsOptionsObject;
+        }
+        interface NavigationBindingsOptionsObject {
```
pawelfus: Not available in API: http://localhost:9005/class-reference/Highcharts.NavigationBindingsOptionsObject

Referenced for example from: http://localhost:9005/highstock/navigation.bindings.arrowInfinityLine

ts/modules/marker-clusters.src.ts
```diff
@@ -288,6 +288,8 @@ var Series = H.Series,
  * @product      highcharts highmaps
  * @since        next
  * @optionparent plotOptions.scatter.cluster
+ *
+ * @private
```
pawelfus: Is this related to Annotations?

ts/annotations/controllable/markerMixin.ts
```diff
@@ -0,0 +1,227 @@
...
+        interface AnnotationMarkerMixin {
+            markerEndSetter(this: SVGElement, value: string): void;
+            markerStartSetter(this: SVGElement, value: string): void;
+            setItemMarkers(this: AnnotationControllablePath, item: AnnotationControllablePath): void;
+        }
+        interface AnnotationChart {
+            afterGetContainer(): void;
+        }
+        interface Options {
+            defs?: Dictionary<SVGDefinitionObject>;
```
pawelfus: Aren't `defs` already defined? See: https://api.highcharts.com/highcharts/defs

ts/annotations/navigationBindings.ts
bre1470: I do not understand. Should this internal interface be public or should this not be created internally?

ts/modules/marker-clusters.src.ts
bre1470: This is related to the class references. The tag gets ignored by the API scripts and just makes sure, that JSDoc is not picking it up itself (to put it in the global namespace).

ts/annotations/controllable/markerMixin.ts
bre1470: Yes, but this is about the internal namespace, not about the API documentation.

ts/annotations/navigationBindings.ts
pawelfus: I think this should be public. Since here: http://localhost:9005/class-reference/Highcharts.NavigationBindingsOptionsObject - we can click reference to this interface:

Screen Shot 2020-01-06 at 11 09 58

I'm pretty sure we users don't want to see this after clicking:
Screen Shot 2020-01-06 at 11 11 57

bre1470: It should be fixed now.

Issue #12152, opened 2019-10-06
Column filled with pattern disappears after reflow


dhilgarth:
#### Expected behaviour
After reflowing the chart as a reaction to a resize of the container, all columns should remain visible

#### Actual behaviour
Columns with a pattern fill disappear

#### Live demo with steps to reproduce
https://stackblitz.com/edit/angular-highcharts-392f9c
* 4 columns are visible. Two red with pattern, two blue without pattern
* Resize the window
--> The two red columns disappeared. Hovering over the blue columns to show the tooltip and then moving the mouse to where the red columns should be will make them re-appear.

#### Product version
highcharts: 7.2.0
highcharts-angular: 2.4.0

#### Affected browser(s)
Tested with Chrome 77 and Firefox 69

sebastianbochan:
Hi @dhilgarth,
It looks like similar to the known bug https://github.com/highcharts/highcharts/issues/12053

@oysteinmoseng could you confirm?

dhilgarth:
I did however see this issue also with v7.0.6 of Highcharts.

oysteinmoseng:
No, this seems to be an unrelated issue. The pattern definition being referred to is deleted from the DOM on window resize, and reinserted on mouse hover. I think there might be some logic looking for unused patterns and removing them, so that could be at play here.

Are we able to reproduce this in a jsFiddle without Angular?

oysteinmoseng:
Turns out the pattern garbage collector looking for patterns in use is looking for SVG fill attributes like `fill="url(#...`. In your demo the fill attribute is `fill="url(https://angular-highcharts-392f9c.stackblitz.io/#highcharts-pattern-0-170de4bb61efed1d4)"`.

Absolute URLs should be okay in these, so we need to change the GC code to look for those patterns as well.

PR #12577, opened 2019-12-05
Bugfix/pattern fill fixes

### LINKED ISSUE #12053: When creating the chart for the second time, patterns don't appear (pattern-fill.js 7.1.3+)
When the chart is created for the first time, everything works fine.

#### Actual behaviour
Since 7.1.3, when we create the chart again, pattern fill is not applied.

https://jsfiddle.net/BlackLabel/mq4Ltkvd

#### Product version
Highcharts 7.1.3+
Works fine in 7.1.2

oysteinmoseng:
Fixed #12053 and #12152, pattern fill issues. Introduced `color.patternIndex` option.
___
Note:
~~Color options of the type `color: 'url(xyz)'` are now passed through `complexColor` as well, to allow for additional processing. We now automatically add the chart index both to the `url()` reference and the pattern `id` to avoid duplicate IDs between charts.~~

See [new comment](https://github.com/highcharts/highcharts/pull/12577#issuecomment-562606374).

### DIFF
--- a/js/modules/pattern-fill.src.js
+++ b/js/modules/pattern-fill.src.js
@@ -481,13 +481,13 @@ H.addEvent(H.Chart, 'redraw', function () {
     if (patterns.length) {
         // Look through the DOM for usage of the patterns. This can be points,
         // series, tooltips etc.
-        [].forEach.call(this.renderTo.querySelectorAll('[color^="url(#"], [fill^="url(#"], [stroke^="url(#"]'), function (node) {
+        [].forEach.call(this.renderTo.querySelectorAll('[color^="url("], [fill^="url("], [stroke^="url("]'), function (node) {
             var id = node.getAttribute('fill') ||
                 node.getAttribute('color') ||
                 node.getAttribute('stroke');
             if (id) {
                 usedIds.push(id
-                    .substring(id.indexOf('url(#') + 5)
+                    .substring(id.indexOf('url(') + 5)
                     .replace(')', ''));
             }
         });
--- a/js/modules/pattern-fill.src.js
+++ b/js/modules/pattern-fill.src.js
@@ -246,10 +246,11 @@ H.SVGRenderer.prototype.addPattern = function (options, animation) {
     }, attribs;
     if (!id) {
         this.idCounter = this.idCounter || 0;
-        id = 'highcharts-pattern-' + (this.chartIndex || 0) + '-' +
-            this.idCounter;
+        id = 'highcharts-pattern-' + this.idCounter;
         ++this.idCounter;
     }
+    // Add chart index
+    id += '-' + (this.chartIndex || 0);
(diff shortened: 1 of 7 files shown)


oysteinmoseng:
Visual difference is probably a bug, looking into it

TorsteinHonsi:
Okay, awaiting your comment. Otherwise approved.

oysteinmoseng:
Patterns are now back to function as they were prior to the changes in `v7.1.3`, meaning IDs are left alone (chart index is not added). If ID is not specified, a unique one is made.

Instead we change the way default patterns are handled. We introduce a global `Highcharts.patterns` array with some default patterns, allowing users to add their own patterns here as well by pushing to the array before creating the chart. To use patterns from this array, specify the `color.patternIndex` option, as an index into the array.

Example:

```js
    series: [{
        data: [{
            y: 1,
            color: {
                patternIndex: 0 // References the first default pattern
            }
        },
        // ...
```

This has a few implications: It breaks backwards compatibility since default patterns now get unique IDs and can't be referenced the old way (`url(#highcharts-default-pattern-0)`). On the bright side it should fix the issues with missing patterns/duplicate IDs across charts. It also means charts do not create pattern elements for patterns they do not use. The `patterns` array gives users an accessible global pool where pattern definitions can be reused - similar to the `colors` array. This could be useful for e.g. themes.

charlwillia6:
@oysteinmoseng - Do you have an example of how you would still use a custom pattern with this new patternIndex?

TorsteinHonsi:
@charlwillia6 Here's a live demo: https://highcharts.github.io/highcharts-utils/samples/#gh/8369a808f5/sample/highcharts/series/pattern-fill-pie

oysteinmoseng:
@charlwillia6 To briefly expand on the demo above, custom patterns can be used in the exact same way as before, e.g. like this:

```js
    series: [{
        data: [{
            y: 1,
            color: {
                pattern: {
                    // Your pattern here
                }
            }
        }
        // ...
```

New with `patternIndex` is that you can also push your custom patterns to the `Highcharts.patterns` global array and pick them up with `patternIndex`. Otherwise there should be no change to the behavior of custom patterns. Example:

```js
const numDefaultPatterns = Highcharts.patterns.length;
Highcharts.patterns.push({
    // Your custom pattern here
});

    // ...

    series: [{
        data: [{
            y: 1,
            color: {
                patternIndex: numDefaultPatterns // References the first pushed custom pattern
            }
        },
        // ...
```

rammohan201:
> ```js
> patternIndex: numDefaultPatterns
> ```

Highcharts.patterns.length; is complaining that patterns is undefined in Highcharts, I am using 8.0.0 version.
what is the version in which this Patterns is available on HighCharts?

pawelfus:
Have you imported `modules/pattern-fill` module?

PR #12588, opened 2019-12-06
Bugfix/6553-legend-symbol-usehtml

### LINKED ISSUE #6553: Legend symbol click is not hiding series when legend.useHTML is set to true
#### Expected behaviour
Clicking on legend symbol should work the same when legend useHTML is set and when legend is not using HTML.

#### Actual behaviour
Legend symbol click is not hiding series when legend.useHTML is set to true

http://jsfiddle.net/5n948mqe/

#### Affected browser(s)
All

Basts:
Fixed #6553, legend symbol functionality with the `useHTML` setting.

### DIFF
--- a/ts/parts/Interaction.ts
+++ b/ts/parts/Interaction.ts
@@ -30,7 +30,7 @@ declare global {
             setItemEvents(
                 item: (BubbleLegend|Point|Series),
                 legendItem: SVGElement,
-                useHTML?: boolean
+                itemToSet: SVGElement
             ): void;
         }
         interface Point {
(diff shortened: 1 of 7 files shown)


ts/parts/Interaction.ts
```diff
@@ -456,7 +456,7 @@ extend(Legend.prototype, {
 
         // Set the events on the item group, or in case of useHTML, the item
         // itself (#1249)
-        (useHTML ? legendItem : (item.legendGroup as any))
+        itemToSet
```
TorsteinHonsi: Instead of changing the function parameters and the way they are called, I think it would be cleaner just to change it here. Something like this:

```js
const elements = useHTML ? [legendItem, item.legendSymbol] : [item.legendGroup];
elements.forEach(element => {
    element.on('mouseover' ......
```

PR #12594, opened 2019-12-09
Added column-comparison demo. Closes #11312.

### LINKED ISSUE #11312: Add Election demo to a demo/article section
Add Election demo as an article and include it in official Highcharts demos. 

Internal: 
It is a good idea to combine new **Sorted Categories Labels Animation** functionality with this demo.

TorsteinHonsi:
Sorry for creating a new PR, I merged it too early. Had some comments.

### DIFF
--- a/samples/highcharts/demo/column-comparison/demo.css
+++ b/samples/highcharts/demo/column-comparison/demo.css
@@ -1,46 +1,46 @@
 #container {
-    min-width: 310px;
-    max-width: 800px;
-    height: 400px;
-    margin: 0 auto
-  }
-  
-  .buttons {
-    text-align: center;
-    margin-bottom: 1.5rem;
-    font-size: 0;
-  }
-  
-  .buttons button {
-    cursor: pointer;
-    border: 1px solid silver;
-    border-right-width: 0;
-    background-color: #f8f8f8;
-    font-size: 1rem;
-    padding: 0.5rem;
-    outline: none;
-    transition-duration: 0.3s;
-  }
-  
-  .buttons button:first-child {
-    border-top-left-radius: 0.3em;
-    border-bottom-left-radius: 0.3em;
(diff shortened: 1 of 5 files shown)


Basts:
@TorsteinHonsi 

Back to the data separation - something like here is that what you had in mind? Demo: http://jsfiddle.net/BlackLabel/g164txvn/

I have a question about setting the point color from the dictionary. Do you have a better idea than using a `series.colors` array with additional function? That's the only one which came to my mind... Demo: http://jsfiddle.net/BlackLabel/Laoz783g/

TorsteinHonsi:
> Back to the data separation - something like here is that what you had in mind?

Yes, exactly like that. It's a clean array of name/value pairs.

> Do you have a better idea than using a series.colors array with additional function? 

I think it's fine the way you do it. Another approach would be to `map` the `data` itself just before you apply it to the series, and set the `color` directly to the data point. But I don't see any real advantages with that.

Basts:
@TorsteinHonsi 

I just found that [this solution ](http://jsfiddle.net/BlackLabel/Laoz783g/) breaks the animation. Open the demo and click the 2004 button to see what I have in mind. 

I tried your other suggestion and here is what I figured out - [demo](http://jsfiddle.net/BlackLabel/mkjg5143/). Animation works fine. Do you approve this solution?

TorsteinHonsi:
Yes, except I think the `countries` should be an array of objects, so make it easier to read, and more configuration-like:

```js
var countries = [{
  name: 'South Korea',
  flag: 197582,
  color: 'rgb(201, 36, 39)'
}, {
  name: 'Japan',
  flag: 197604,
  color: 'rgb(201, 36, 39)'
}, ...
```

TorsteinHonsi:
Nice work, thanks!

PR #12611, opened 2019-12-11
Tools: Added gulp task for bumping version and cleaning up after release.

activist:
Can be tested using `gulp prep-release --nextversion x.y.z` and `gulp prep-release --cleanup (--commit)`

@TorsteinHonsi Behind the scenes this is using sed which can be slightly different on OSX. Please test the command and let me know if you experience any issues.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -73,11 +73,7 @@ nbproject/
 /samples/**/diff.gif
 /samples/**/reference.svg
 
-/tmp/report.html
-
-/tmp/missing.in.code.json
-
-/tmp/processed.api.json
+tmp/
 /package-lock.json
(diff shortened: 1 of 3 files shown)


activist:
@TorsteinHonsi  please test again now. I have made some changes that should sort out the differences on OSX.

TorsteinHonsi:
Yes, it works  fine now! But it produces three backup files, those should be git-ignored so we don't accidentally commit them, right? Or just write them to `/tmp/` or somewhere that we already ignore.

activist:
@TorsteinHonsi Hopefully it should work now by adding the files to tmp/ as suggested.

PR #12616, opened 2019-12-11
enhancement/Inverted-3d-improvements

### LINKED ISSUE #5508: Datalabels misplacing on 3D bar chart.
#### Expected behaviour

When dataLabels.enabled is set to true, the dataLabels should be displayed properly.
#### Actual behaviour

The dataLabels are displayed incorrectly.

https://jsfiddle.net/scottszb1987/n603jpt4/
#### Affected browser(s)

Chrome

Izothep:
Fixed #5508, data labels were wrongly positioned on inverted 3D charts.

___
Enhancement with multiple 3D column series improvements, such as

- [x] Creating additional argument in H.perspective method
- [x] Improved dataLabels position with 180 alpha rotation
- [x] Improved dataLabels position with inverted graphs
- [x] Improved stackLabels position
- [x] Improved Tooltip position with inverted graphs 
- [x] Added Normalisation for 3D chart angles calculation


ts/parts-3d/Column.ts
```diff
@@ -503,22 +494,41 @@ wrap(Series.prototype, 'alignDataLabel', function (
     proceed: Function
 ): void {
 
+    // In 3D we need to pass point.outsidePlot option to the justifyDataLabel
+    // method for disabling justifying dataLabels in columns outside plot
+    arguments[3].outside3dPlot = arguments[1].outside3dPlot;
```
KacperMadej: See the comment below.

ts/parts-3d/Column.ts
```diff
@@ -503,22 +494,41 @@ wrap(Series.prototype, 'alignDataLabel', function (
...
-            chart = series.chart;
-
         var args = arguments,
             alignTo = args[4],
-            point = args[1];
+            point = args[1],
+            dLOptions = args[3];
 
-        var pos = ({ x: alignTo.x, y: alignTo.y, z: series.z });
+        var series = this as Highcharts.ColumnSeries,
```
KacperMadej: `var`, `var` -> single `var` will work well here, or even better: use `const` and `let`.

ts/parts-3d/Column.ts
```diff
@@ -533,18 +543,42 @@ wrap(H.StackItem.prototype, 'getStackBox', function (
...
+            series: Highcharts.Series
+        ): void {
+            if (series instanceof seriesTypes.column && series.data.length) {
+                columnSeries = series;
+            }
         });
+        // If any series is a column series, use its barW, z and depth
+        // parameters for correct stackLabels position calculation
+        if (columnSeries) {
+            var height: number = arguments[6],
```
KacperMadej: `var` in here doesn't look good - consider using `let` / `const` or placing all `var` at the top of the scope

ts/parts-3d/Column.ts
```diff
@@ -503,22 +494,41 @@ wrap(Series.prototype, 'alignDataLabel', function (
...
-        var pos = ({ x: alignTo.x, y: alignTo.y, z: series.z });
+        var series = this as Highcharts.ColumnSeries,
+            seriesOptions: Highcharts.ColumnSeriesOptions = series.options,
+            chart = series.chart,
+            inside = pick(dLOptions.inside, !!series.options.stacking),
+            pos = ({
+                x: alignTo.x + point.pointWidth / 2,
+                y: alignTo.y,
+                z: series.z + (seriesOptions as any).depth / 2
+            });
```
KacperMadej: is the `()` a leftover from testing some TS castings? `pos = ({...})` should be the same as `pos = {...}`

ts/parts-3d/Column.ts
```diff
@@ -503,22 +494,41 @@ wrap(Series.prototype, 'alignDataLabel', function (
...
+
+        if (this.chart.inverted && inside) {
```
KacperMadej: `chart` is available, so no need for `this.chart`

ts/parts-3d/Column.ts
```diff
@@ -533,18 +543,42 @@ wrap(H.StackItem.prototype, 'getStackBox', function (
...
-            y: stackBox.y,
-            z: 0
+        var columnSeries: Highcharts.ColumnSeries | undefined;
+        // Check if any series iherits from column 3D
+        chart.series.forEach(function (
+            series: Highcharts.Series
+        ): void {
+            if (series instanceof seriesTypes.column && series.data.length) {
+                columnSeries = series;
+            }
```
KacperMadej: Why looping over all series if only one `columnSeries` is later used? This looks like a mistake, but if this is intended then consider `Array.some` instead of `Array.forEach`.

ts/parts-3d/Column.ts
```diff
@@ -533,18 +543,42 @@ wrap(H.StackItem.prototype, 'getStackBox', function (
...
-        pos = H.perspective([pos], chart, true)[0];
-        stackBox.x = pos.x;
-        stackBox.y = pos.y;
+            if (chart.inverted) {
+                // Do not use default offset calculation logic for 3D
+                // inverted stackLabels.
+                stackBox.width = 0;
+            }
+
+            pos = H.perspective([pos], chart, true, false)[0];
```
KacperMadej: How about renaming `pos` to something more descriptive? and adding comments when the value of the pos variable is changing significantly (from flat to 3d, from not adjusted to adjusted, etc.), or maybe even using more variables for better readability (and TS types might be easier to write).

Izothep:
All of the mentioned changes are done. There is still a need for preparing proper tests for 3D columns (like testing H.perspective) before the PR will be considered as final.

Izothep:
Visual comparison of the old and new version.

Old:
Screenshot 2019-12-20 at 15 54 52

New:
Screenshot 2019-12-20 at 15 54 37
Screenshot 2019-12-20 at 15 54 31

js/parts-3d/Column.js
```diff
@@ -265,6 +259,7 @@ addEvent(Series, 'afterInit', function () {
                 z = (stacks.totalStacks * 10) - z;
             }
         }
+        seriesOptions.depth = seriesOptions.depth || 25;
```
sebastianbochan: I suggest to set 25 as value in defaultOptions instead of hardcoding.

ts/parts-3d/Column.ts
```diff
@@ -140,10 +130,10 @@ wrap(H.Series.prototype, 'justifyDataLabel', function (
 seriesTypes.column.prototype.translate3dPoints = function (): void {};
 seriesTypes.column.prototype.translate3dShapes = function (): void {
 
-    var series = this as Highcharts.ColumnSeries,
+    var series: Highcharts.ColumnSeries = this,
         chart = series.chart,
         seriesOptions = series.options,
-        depth = seriesOptions.depth || 25,
+        depth = (seriesOptions as any).depth,
```
sebastianbochan: seriesOptions has own type (`SeriesOptions` or `ColumnSeriesOptions`), use it instead of `any`

ts/parts-3d/Column.ts
```diff
@@ -218,10 +218,11 @@ seriesTypes.column.prototype.translate3dShapes = function (): void {
                 [{
                     x: (tooltipPos as any)[0],
                     y: (tooltipPos as any)[1],
```
sebastianbochan: You can try to use `PositionObject` or `TooltipPositionerPointObject` instead of `any`

js/parts-3d/Column.js
```diff
@@ -339,16 +349,35 @@ wrap(Series.prototype, 'alignDataLabel', function (proceed) {
...
+            if (series instanceof seriesTypes.column && series.data.length) {
+                columnSeries = series;
+            }
         });
-        pos = H.perspective([pos], chart, true)[0];
-        stackBox.x = pos.x;
-        stackBox.y = pos.y;
+        // If any series is a column series, use its barW, z and depth
+        // parameters for correct stackLabels position calculation
+        if (columnSeries) {
```
sebastianbochan: Why this If is not in forEach from line 356?

ts/parts-3d/Column.ts
```diff
@@ -500,27 +513,56 @@ if (seriesTypes.columnrange) {
...
+            options3d = (chart.options.chart as any).options3d;
+
+        let dLPosition = {
+            x: alignTo.x + point.pointWidth / 2,
+            y: alignTo.y,
+            z: series.z + (seriesOptions as any).depth / 2
+        };
+
+        if (chart.inverted) {
+            // dLPosition.y += (point.shapeArgs as any).width / 2;
```
KacperMadej: Is this a comment or a commented code line?

ts/parts-3d/Column.ts
```diff
@@ -500,27 +513,56 @@ if (seriesTypes.columnrange) {
...
+            // logic and there is no need to position them using
+            // non-3D algorighm (that use alignTo.width)
+            if (inside) {
+                alignTo.width = 0;
+                dLPosition.x += (point.shapeArgs as any).height / 2;
+            }
+            // When chart is upside down
+            // (alpha angle between 180 and 360 degrees)
+            // it is needed to add column width to calculated value.
+            if (options3d.alpha >= 90 && options3d.alpha <= 270) {
```
KacperMadej: I am unable to find in code normalization to 0-360 degree range for alpha. Is rotating the chart for x + 360*n covered? or not really supported anywhere and this place shouldn't be different?

ts/parts-3d/Column.ts
```diff
@@ -530,21 +572,51 @@ wrap(Series.prototype, 'alignDataLabel', function (
...
+    stackItem: Highcharts.StackItem,
+    x: number,
+    y: number,
+    xWidth: number,
+    h: number,
+    axis: Highcharts.Axis
 ): void { // #3946
     var stackBox = proceed.apply(this, [].slice.call(arguments, 1));
+    // Only do this for 3D column and iherited series.
+    if (chart.is3d() && stackItem.base) {
```
KacperMadej: How is this check ensuring that we have a 3D column or it's "iherited" (btw: should be inherited) series?
The check for "columness" is a bit lower in the code as far as I can tell.

js/parts-3d/Column.js
Izothep: We had similar discussion with @pawelfus about the above code and unfortunately due to a fact that we are not creating new series type but just extend existing one, we should keep the old way of this declaration (this is the same code as before but just moved into another place). However, I am totally with you about this line of code, had my own doubts about it and I think in future it should be a separate series type (column-3d) where you can add this option. This will also solve typescript problems.

ts/parts-3d/Column.ts
Izothep: Unfortunately not possible for now due to a fact that for now depth might be undefined (it is added as an optional parameter in TS). I have also discussed this line with Pawel and we both agreed that before finding more global solution for similar use-cases we should not focus on that issue.

js/parts-3d/Column.js
Izothep: The reason was that we wanted to find any column series in the graph and then assign it to columnSeries variable. However, this part of code is already refactored so we can mark it as resolved.

ts/parts-3d/Column.ts
Izothep: Thanks! I will test it later today!

ts/parts-3d/Column.ts
Izothep: Thanks! I will add it in the next commit. To be honest, I found the same issue just an hour ago. We should cover this normalisation for every use case, not only for dataLabels.

ts/parts-3d/Column.ts
Izothep: let's call it a ghost from a past. I will remove it in the next commit.

ts/parts-3d/Column.ts
Izothep: this check for columnSeries is just 10 lines below.

Izothep: ... so I will move the comment to this place and correct a typo :)

ts/parts-3d/Column.ts
Izothep: We use `Highcharts.Point.tooltipPos?` Which is an optional parameter and that is the reason of the problem.

Izothep: I would leave it for now and focus on the problem in future as it is more global one (similar `as any` is used in `dumbbel`, `varivide`, `xrange`, `polar`, `waterfall` etc.)

Izothep:
All of the previously discussed issues are resolved now. Further enhancements are added. The Pull Request is ready to be considered as final in my opinion.

ts/parts-3d/Column.ts
```diff
@@ -223,15 +218,33 @@ seriesTypes.column.prototype.translate3dShapes = function (): void {
...
+
+            // Recalculate point positions for inverted graphs
+            if (chart.inverted) {
+                point2dPos.x = (shapeArgs as any).height;
+                point2dPos.y = point.clientX;
+            }
+
+            // Calculate and store point's position in 3D,
+            // using perspective method.
+            point.pos3d = perspective([point2dPos], chart, true, false)[0];
```
pawelfus: Just nitpicking: since `pos3d` is equivalent to plotX/Y then maybe `plot3d` name would be better?

samples/unit-tests/3d/column-data-labels-visibility/demo.js
```diff
@@ -28,9 +28,133 @@ QUnit.test('3D columns dataLabels initial visibility', function (assert) {
...
+            }
+        }
+    });
+
+    dataLabel = points[0].dataLabel;
+    assert.close(
+        dataLabel.x + dataLabel.padding / 2,
+        points[0].pos3d.x,
+        4,
+        'dataLabel is in proper x position in inverted graph'
```
pawelfus: Replace all `is` with `should be`.

samples/unit-tests/3d/column-data-labels-visibility/demo.js
```diff
@@ -28,9 +28,133 @@ QUnit.test('3D columns dataLabels initial visibility', function (assert) {
...
+    );
+    assert.close(
+        dataLabel.y + dataLabel.absoluteBox.height / 2,
+        points[1].series.xAxis.height - points[1].pos3d.y,
+        4,
+        'dataLabel is in proper y position in inverted graph'
+    );
+
+    chart.update({
```
pawelfus: Is this update necessary?

pawelfus:
I see a lot of diff changes. I assume now tooltip should be displayed in the geometric center - but not always it looks great. Perhaps `column3d` series should set `tooltip.distance` to `0`?

samples/unit-tests/3d/column-data-labels-visibility/demo.js
Izothep: Done with 65ed49e

samples/unit-tests/3d/column-data-labels-visibility/demo.js
Izothep: Removed in 65ed49e

ts/parts-3d/Column.ts
Izothep: I have changed this name in 65ed49e

Izothep:
@activist Could you please look at the issue with styled mode and Karma, mentioned by @pawelfus? It looks to be general issue, not related to this specific branch.

activist:
@Izothep @pawelfus Karma is currently not loading "external styles" when styled mode is uses. In that case it would have to fetch and inline the styles. These styles are used instead, which I assume explains the difference: https://github.com/highcharts/highcharts/blob/f2fb01cb33fd562098001078ea5d7c8cb0237a83/test/karma-setup.js#L430-L436

pawelfus:
Yes, that explains it. Thanks!

PR #12618, opened 2019-12-11
Tools: Added gulp test for docs not in sidebar

goransle:
Adds a pre-commit test that checks if there are any documents in `/docs/` that are not added to `/docs/sidebars.js` or the the new `doc-config.js`.

Added the current offenders as unlisted as to not interrupt any work in progress.


goransle:
> Thanks! While we're at it, we should probably also validate the other way around - check that any item listed in the `sidebar.js` actually exists in the file system. Unless that is already handled by the docs builder?

It currently errors out on the build when a file in the sidebar is not found, which should be picked up by CI. But it should be a quick addition to add it here also.

PR #12622, opened 2019-12-12
JS fiddle for demo issue

multiplechuks:
Added a working JSfiddle


TorsteinHonsi:
Thanks for the PR!

We won't link to fiddles that we don't control ourselves, but as far as I can see, the only change you did to the fiddle was to update the Highcharts path to use `https`. So I made that change to our own fiddle, and we don't need to change the docs.

I'm closing this PR, please comment if I overlooked something.

PR #12624, opened 2019-12-12
Tools: Fail CI build when workspace is dirty

activist:
Added missing transpiled js file changes.

If someone forgets to e.g commit a js file after editing a TS file the build will now fail early.


activist:
The build is still failing since the missing file is not present on master, but after this is merged the build will fail earlier in the process.

PR #12649, opened 2019-12-17
Tools: Report slow tests to create awareness. Browserstack timezone set to UTC

activist:
Just some small changes that may - or may not be merged depending if you think it is reasonable.

Tests slower than 3 seconds will be logged in the output in order to create some awareness around slow tests and to reduce amount of slow tests.

Set browserstack timezone to UTC in order to have a deterministic timezone to relate to. Not sure if this will fix/mask the issue on the nightly, but at least we now have a timezone to relate to that can be changed if we want to test other timezones.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -142,6 +142,18 @@ jobs:
             - slack/status:
                 fail_only: true
 
+  test_timezones:
+    <<: *defaults
+    description: Run tests with different browser configs at BrowserStack
+    parameters:
+      browsers:
+        description: "Which browser to test?"
+        type: string
+        default: "ChromeHeadless"
+    steps:
+      - <<: *load_workspace
+      - run: ./.circleci/scripts/test_timezones.sh -b <<parameters.browsers>>
+
   check_for_docs_changes_and_deploy:
     <<: *defaults
     description: Checks for changes in docs/ and triggers deploy via doc-builder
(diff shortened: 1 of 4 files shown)


TorsteinHonsi:
> Set browserstack timezone to UTC in order to have a deterministic timezone to relate to.

That is good for consistency, but on the other hand, we now risk missing issues with various time zones. Could it be an idea to run the time tests in other timezones? Especially Indian time is tricky because it is 30 minutes off. `gulp test --tests unit-tests/time/*` should be enough.

activist:
> That is good for consistency, but on the other hand, we now risk missing issues with various time zones. Could it be an idea to run the time tests in other timezones? Especially Indian time is tricky because it is 30 minutes off. `gulp test --tests unit-tests/time/*` should be enough.

I agree that we could risk missing out on those kind of issues. Since this looks like a browser dependent issue we would have to run it on Browserstack for a variety of browsers/platforms. It is a simple thing to add, but a concern is that it will add more pressure on our existing bottleneck called Browserstack. How about adding it for nightly for now?

activist:
Added test that runs on this specific branch for testing purposes and on the nightly build (testing ChromeHeadless, Safari, Firefox (Mac) and Chrome (Win)).

It tests the following timezones per run:
Australia/Melbourne
America/Los_Angeles
Asia/Kolkata
Atlantic/Reykjavik
Pacific/Fiji

Issue #12589, opened 2019-12-06
Bar series tooltip in wrong position if xAxis has top option set


KacperMadej:
#### Behaviour
Inverted column chart or bar series is having a tooltip displayed in the wrong position if the xAxis connected to the series has `top` option set.

#### Live demo with steps to reproduce
https://jsfiddle.net/BlackLabel/a8ksLnz4/

#### Product version
7.2.1

PR #12655, opened 2019-12-17
bugfix/12589-tooltip-wrong-position

wchmiel:
Fixed issue with tooltip wrong position on a bar chart when `xAxis` top or height set, see #12589.


TorsteinHonsi:
Thanks for the changes! I did a simplification for the fix, instead of reading and parsing the options over again, tried to read `Axis.top` directly, as this is already the result of handling the options. So let's see how the visual tests go.

PR #12659, opened 2019-12-18
Tools: Align comparison and storage of visual test samples

### LINKED ISSUE #12640: Tools: Visual test comparison can report diff even when there seemingly is none
The diff produced by karma seems to sometimes be a false positive. Take for instance https://vrevs.highsoft.com/pr/12508/review from #12508 where there are diffs reported. In the produced gif there is possible to see a small diff that seems like font rendering or delay in rendering between the samples. 

When downloading SVGs in question, converting them to PNG and comparing with pixelmatch no diff is found which seems correct to me when visually comparing the two SVG files.

E.g 
https://s3.eu-central-1.amazonaws.com/staging-vis-dev.highcharts.com/visualtests/diffs/pullrequests/12508/highcharts/demo/heatmap/reference.svg
and 
https://s3.eu-central-1.amazonaws.com/staging-vis-dev.highcharts.com/visualtests/diffs/pullrequests/12508/highcharts/demo/heatmap/candidate.svg
seems identical to me.

We should do some digging to see if we can find the culprit or perhaps use pixelmatch to do the diffing in both cases.

activist:
Made sure visual test samples are stored, loaded and compared in the same way.

Fixes #12640

### DIFF
--- a/test/karma-setup.js
+++ b/test/karma-setup.js
@@ -448,14 +448,19 @@ function getSVG(chart) {
  */
 function compare(data1, data2) { // eslint-disable-line no-unused-vars
     var i = data1.length,
-        diff = 0;
+        diff = 0,
+        pixels = [],
+        pixel;
 
     // loops over all reds, greens, blues and alphas
     while (i--) {
-        if (!data1[i] === data2[i]) {
+        pixel = Math.floor(i / 4);
+        if (Math.abs(data1[i] - data2[i]) !== 0 && !pixels[pixel]) {
+            pixels[pixel] = true;
             diff++;
         }
     }
+
     return diff;
 }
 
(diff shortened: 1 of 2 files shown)


test/karma-setup.js
```diff
@@ -448,29 +444,24 @@ function getSVG(chart) {
...
-        pixels = [],
-        pixel;
+        diff = 0;
 
     // loops over all reds, greens, blues and alphas
     while (i--) {
-        pixel = Math.floor(i / 4);
-        if (Math.abs(data1[i] - data2[i]) !== 0 && !pixels[pixel]) {
-            pixels[pixel] = true;
+        if (!data1[i] === data2[i]) {
```
TorsteinHonsi: By removing the check for `pixels[pixel]`, we are now no longer counting the amount of different pixels, but the amount of different color values. Each pixel has a red, green, blue and alpha value. So the resulting number will now be up to four times the actual amount of changed pixels. Is that intended?

activist: Sorry, that was a mistake on my end. Clearly it is RGBA values. I have reverted the change now.

PR #12686, opened 2019-12-27
bugfix/4369-wrong-zindex-3d-column

### LINKED ISSUE #4369: Incorrect column 3d when reversedStacks enabled
Example: http://jsfiddle.net/k0rfv8bw/6/

Without reversedStacks, chart is printed properly.

Izothep:
Fixed #4369, 3D columns had wrong zIndex with long charts.


Izothep:
The difference between old and new version.

Old:
Screenshot 2019-12-30 at 10 06 08


New:
Screenshot 2019-12-30 at 10 05 47

ts/parts-3d/SVGRenderer.ts
```diff
@@ -715,13 +742,15 @@ H.SVGRenderer.prototype.cuboidPath = function (
...
     ): Array<number|Array<number>> {
-        var ret = [[] as any, -1];
+        var ret = [[] as any, -1],
+            pathSide1: Array<number> = path1.map(mapSidePath) as any,
+            pathSide2: Array<number> = path2.map(mapSidePath) as any;
 
         path1 = path1.map(mapPath) as any;
         path2 = path2.map(mapPath) as any;
-        if (H.shapeArea(path1 as any) < 0) {
+        if (H.shapeArea(pathSide1 as any) < 0) {
```
pawelfus: Quite a lot these castings for `pathSideX` variables - above first are defined as  `Array<number>` then casted to `any` and now here used as `any` again?

ts/parts-3d/SVGRenderer.ts
```diff
@@ -700,6 +710,23 @@ H.SVGRenderer.prototype.cuboidPath = function (
      * helper method to decide which side is visible
      * @private
      */
+    function mapSidePath(i: number): Highcharts.Position3dObject {
+        // Added support for 0 value in columns, where height is 0
+        //  but the shape is rendered
+        if (h === 0 && i > 1 && i < 6) {
+            return {
+                x: pArr[i].x,
+                y: pArr[i].y + 10,
```
pawelfus: These hardcoded values: `6` and `10` - what those mean?

Izothep: I have commented the function in 78f4008

ts/parts-3d/SVGRenderer.ts
Izothep: It was needed to refactor and comment the whole method to make it more clear for everyone. I have additionally removed all castings in the method. Related commit: 78f4008

ts/parts-3d/Math.ts
```diff
@@ -169,6 +170,9 @@ H.perspective3D = function (
  * @param {boolean} [insidePlotArea]
  * Wether to verifiy the points are inside the plotArea
  *
+ * @param {boolean} [useInvertedPersp]
+ * Wether to use inverted perspective in calculations
```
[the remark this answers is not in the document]
Izothep: yes, its a copy-pasted error. I will correct all of the "wether" mistakes in 3D code.

ts/parts-3d/Math.ts
```diff
@@ -258,10 +267,11 @@ H.pointCameraDistance = function (
             z: pick(options3d.depth, 1) * pick(options3d.viewDistance, 0) +
                 options3d.depth
         },
+        // Added support for objects with plotX or x coordinates.
         distance = Math.sqrt(
-            Math.pow(cameraPosition.x - coordinates.plotX, 2) +
-            Math.pow(cameraPosition.y - coordinates.plotY, 2) +
-            Math.pow(cameraPosition.z - coordinates.plotZ, 2)
+            Math.pow(cameraPosition.x - (coordinates.plotX || coordinates.x), 2) +
```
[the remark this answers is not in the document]
Izothep: `pick` will be the best in this case, nice catch!

TorsteinHonsi:
Thanks @Izothep! Any comment on the visual diff? It seems the only difference is that 0 height columns now have a stronger edge/border. Is that intentional? It is no blocker, but we should be concious about why it changed.

Izothep:
> Any comment on the visual diff? It seems like all the 0 columns got a stronger edge/border. It may not be a bug, but we should be concious about why it happens.
> 
> Otherwise this is good to go, just need clarification on the visual diff.

Thank you for the feedback @TorsteinHonsi! The visual diff is expected:
- In the previous version of 3D column, if pickSide method was not able to choose between left and right (because of 0 height in column) we were not drawing path for any side.
- In the current version pickSide should be able to choose between left and right every time. This cause the situation where we draw a side for the face of cuboid with 0 height (so we simply draw a path).

This approach is needed for dealing with unreported bug with 3d Columns animation from 0 height. 

Master:
master

Animation after current changes:
candidate

PR #12694, opened 2019-12-30
PoC: module loader/synchronizer.

pawelfus:
A simple implementation for module loader/synchronizer. We have more and more modules that depend on each other. That means it's getting harder to manage them, especially for users: what's the order of loading modules and where it's documented?

PR shows an example how we can sync column-3d and highcharts-more.

In this solution, every module should fire `moduleLoader.loadedModule('module-A');`. Then all other modules can add wraps/events to be applied only when a specific module is loaded: 

```js
// module B
moduleLoader.onLoadModule('module-A').then(function () {
  addEvent(Something.That.ModuleA.Created, function () {});
  
  wrap(Something.That.ModuleA.Created.prototype, function () {});
});
```

If `ModuleA` was already loaded, callback is executed immediately. Otherwise it waits for the module to be loaded.

Example in demos:
- correct order of files: http://jsfiddle.net/BlackLabel/njL697m4/
- wrong order of files: http://jsfiddle.net/BlackLabel/7ojp5tnu/6/ (overlapping columns)
- wrong order of files, with PR'ed version: http://jsfiddle.net/BlackLabel/7ojp5tnu/9/

I think we can depromisify this loader, using `addEvent`/`fireEvent`. Maybe even we can place this automatically in distribution files when assembling those? Or perhaps there's much better solution available (something with register module)?

Why we can't use `addEvent`/`fireEvent` out of the box? Because if `fireEvent` would be called before `moduleB` registers it's callback then it never would be called.

### DIFF
--- a/ts/parts/Utilities.ts
+++ b/ts/parts/Utilities.ts
@@ -105,6 +105,24 @@ declare global {
         interface FormatterCallbackFunction<T> {
             (this: T): string;
         }
+        interface ModuleLoaderModuleObject {
+            resolved: boolean;
+            dependencies: Function[];
+        }
+        interface ModuleLoaderModulesLookupObject {
+            [key: string]: ModuleLoaderModuleObject;
+        }
+        interface LoadedModuleFunction {
+            (moduleName: string): void;
+        }
+        interface OnLoadModuleFunction {
+            (moduleName: string): Promise<void>;
+        }
+        interface ModuleLoaderObject {
+            modules: ModuleLoaderModulesLookupObject;
+            loadedModule: LoadedModuleFunction;
+            onLoadModule: OnLoadModuleFunction;
+        }
         interface ObjectEachCallbackFunction<T> {
             (this: T, value: any, key: string, obj: any): void;
         }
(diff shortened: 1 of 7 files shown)


sebastianbochan:
Tested with multiple modules (solid-gauge) and works correct! (One of tests in react native wrapper)

https://jsfiddle.net/BlackLabel/q67vk5ho/

ts/parts/Utilities.ts
```diff
@@ -105,6 +105,24 @@ declare global {
         interface FormatterCallbackFunction<T> {
             (this: T): string;
         }
+        interface ModuleLoaderModuleObject {
+            resolved: boolean;
+            dependencies: Function[];
+        }
+        interface ModuleLoaderModulesLookupObject {
+            [key: string]: ModuleLoaderModuleObject;
+        }
```
KacperMadej: This is the same as `Dictionary<ModuleLoaderModuleObject>` if you don't need this `ModuleLoaderModulesLookupObject` to be defined in here then you could use that dictionary when needed (e.g. in `interface ModuleLoaderObject` below).

TorsteinHonsi:
I think the goal should be that we can also use script tags to load modules independently. We have resolved similar things in the past by using `fireEvent` with an `order`. 

> Why we can't use addEvent/fireEvent out of the box? Because if fireEvent would be called before moduleB registers it's callback then it never would be called.

We should be able to resolve that by initializing modules from the `Chart.init` event or similar?

pawelfus:
> We should be able to resolve that by initializing modules from the Chart.init event or similar?

I think we can, it could be something like this:

- hc-3d:
```js
Chart.moduleInitializers.push(
  // Code required for correct 3D columnrange series:
  addEvent(Highcharts, 'loaded.highcharts-more', function () { ... });
);

addEvent(
  Chart,
  'init',
  function () {
    fireEvent('loaded.highcharts-3d');
  }
);
```
- hc-more:
```js
addEvent(
  Chart,
  'init',
  function () {
    fireEvent('loaded.highcharts-more');
  }
);
```
- hc-core:
```js
// Clean up after initialization
Chart.afterInit = function () {
  Chart.moduleInitializers.forEach(unbind => unbind());
  Chart.moduleInitializers.length = 0;
};
```

But somehow it feels wrong, to init modules when the first chart is initialized. I'm not sure if we won't break something.. +we fire incorrect events (`loaded.something`) on every chart init. What do you think?

bre1470:
It might be useful to have a special `load` event, that can be triggered just ones, and after that any attachment to it will be fired instantly with the saved callback context. An call order can not be supported in this case.

bre1470:
Sorry, duplicate suggestion.

PR #12699, opened 2019-12-31
bugfix/7204-ohlc-stem-extend

### LINKED ISSUE #7204: OHLC Ticks are Drawn Improperly When Line Width is > 1
#### Expected behaviour
The lineWidth should be taken into account when drawing the ticks. 

#### Actual behaviour
image

http://jsfiddle.net/7zadtvo1/

#### Affected browser(s)
Firefox, Chrome, etc.

pawelfus:
Fixed #7204, OHLC ticks were rendered incorrectly with higher `series.lineWidth` values.
___
Now stem is stretched to open/close values, to prevent case when it seems that open/close are below/above low/close values.

Before:
Screen Shot 2019-12-31 at 15 34 43
After:
Screen Shot 2019-12-31 at 15 34 36

I expect some visual changes.

### DIFF
--- a/js/parts/OHLCSeries.js
+++ b/js/parts/OHLCSeries.js
@@ -201,9 +201,18 @@ seriesType('ohlc', 'column'
      * @return {void}
      */
     drawPoints: function () {
-        var series = this, points = series.points, chart = series.chart;
+        var series = this, points = series.points, chart = series.chart, 
+        /**
+         * Extend vertical stem to open and close values.
+         */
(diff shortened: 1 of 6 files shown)


pawelfus:
@activist - could you take a look? Seems that some files weren't generated?

activist:
> @activist - could you take a look? Seems that some files weren't generated?

Seems like that did the trick. Dunno why it happened in the first place :)

Issue #12337, opened 2019-11-06
stackedLabel has incorrect position after resize chart container


balkarov:
Hello. 

**Case:**
Resize chart container and update chart size chart.reflow(); and return the size back.

![1](https://user-images.githubusercontent.com/6030335/68295413-6f961f00-00a3-11ea-90da-35ea5de3dd48.gif)

#### Expected behaviour
stackedLabel the same position.
![image](https://user-images.githubusercontent.com/6030335/68295478-905e7480-00a3-11ea-82ea-1282acef1914.png)

#### Actual behaviour
stackedLabel incorrect position.

![image](https://user-images.githubusercontent.com/6030335/68295330-35c51880-00a3-11ea-8ae1-95931645998e.png)

#### Live demo with steps to reproduce
https://jsfiddle.net/nhxy8a91/

#### Product version
Highcharts 7.2.1

#### Affected browser(s)
Google Chrome 78.0.3904.87

pawelfus:
Thanks for reporting this bug @balkarov 

@Basts - could you take a look? Looks like a regression since v7.1.3, v7.1.2 is fine: https://jsfiddle.net/BlackLabel/hp85q3km/

TorsteinHonsi:
Bisected to ed7ea403acc4299600cf8dc5a7d98d53b9ce78b6

Basts:
Closed by #12779

Issue #12308, opened 2019-10-30
For stacked labels changed poistion after 7.1.3


balkarov:
In version 7.1.2 -> 7.1.3 for stackedLabels changed position.
7.1.2
![image](https://user-images.githubusercontent.com/6030335/67851157-e028c300-fb1a-11e9-92b1-6ea0a99964b4.png)

7.1.3
![image](https://user-images.githubusercontent.com/6030335/67851202-f8004700-fb1a-11e9-8a1e-a000e4c9f627.png)

#### Expected behaviour
Did not change position

#### Actual behaviour
Changed position

#### Live demo with steps to reproduce
7.1.2
https://jsfiddle.net/ju1b86c4/
 
7.1.3
https://jsfiddle.net/txoLpnec/

#### Product version
Highcharts 7.1.2 - 7.2.0

#### Affected browser(s)
Chrome 77.0.3865.120

pawelfus:
Hi @balkarov 

Thanks for reporting. I'm not sure if this is an intended change (@TorsteinHonsi , @Basts, could you confirm?), but in v7.1.3 `stackLabels` were improved to support features like overflow/crop/padding/border/background etc.

Now you can set padding (which defaults to 0) for labels, to get the same output as in v7.1.2: https://jsfiddle.net/BlackLabel/dnvL2ebh/

balkarov:
Hello @pawelfus 

thank you for your answer.
Why when you added new properties did not leave this padding (6)? Why did 0?

Basts:
Hi @balkarov, @pawelfus,

It was an intended change. Before implementing overflow and crop features into stack labels, labels were rendered as a text (not a label) with fixed padding set to 6 without possibility to change it. Now it is more 'flexible' and the user has the possibility to set wanted padding. 

[Demo](https://jsfiddle.net/BlackLabel/pdwuahkm/) - v7.1.3 with padding set to 6
[Demo](https://jsfiddle.net/BlackLabel/81f3c6ub/) - previous version with padding set to 20 which doesn't change anything

balkarov:
@Basts @pawelfus 
For `column` chart you have space between column and stacklabels by default
https://jsfiddle.net/jrea47Lv/

![image](https://user-images.githubusercontent.com/6030335/67934065-24c66400-fbd8-11e9-9732-6d894782924c.png)


But for `bar` not
[...]

Issue #12294, opened 2019-10-29
StackLabels outside xAxis min & max range are displayed


aroue2:
Having a stacked column graph with two series of 10 columns each, but displaying only columns with xAxis >= 2 and xAxis <= 8

#### Expected behaviour
StackLabels of not showed columns are not showed

#### Actual behaviour
StackLabels of not showed columns are showed

#### Live demo with steps to reproduce
https://jsfiddle.net/so7jnzwg/ : A chart with stackLabels and dataLabels. We can see that dataLabels of not showed columns are not displayed while stackLabels are displayed.

#### Product version
Highcharts 7.2.0

#### Affected browser(s)
At least Firefox, and Google Chrome

Basts:
Closed by #12779

Issue #11982, opened 2019-09-16
Overlapping stackLabels


danazelniker:
#### Expected behaviour

labels shouldn't overlaps one another when there is not enough space

#### Actual behaviour

labels overlaps

#### Live demo with steps to reproduce
![insights bug](https://user-images.githubusercontent.com/16417615/64968804-6d85c000-d8ab-11e9-823a-7f5d67abf5ff.png)

#### Product version
 Highcharts: 6.1.1

#### Affected browser(s)
all

pawelfus:
Hi @danazelniker 

Thanks for reporting this issue.

I see two issues here:

- first arrow: 
  There's no logic in overlapping dataLabels and other columns, only overlapping between labels is supported. I suggest to create an idea on our [UserVoice](https://highcharts.uservoice.com/forums/55896-highcharts-javascript-api) - if the idea gets popular it will be implemented to the core (or as a plugin) :) 

- second arrow:
  That looks like a bug. Could you confirm that [`series.dataLabels.allowOverlap`](https://api.highcharts.com/class-reference/Highcharts.DataLabelsOptionsObject#allowOverlap) is set to false? Could you also check if current version (v7.2.0) has the same issue? If none of the above resolve the issue, please provide a live demo (e.g. jsFiddle). Thanks!

danazelniker:
Hi @pawelfus 

I'm not sure where I should add `series.dataLabels.allowOverlap: false`.
currently the `series` key is an array of my data.
you can see it here:
https://jsfiddle.net/vf2dw6xa/1/
Thanks!

pawelfus:
If you didn't set it, then by default it's set to false, no need to change anything. Just in case, by `series.dataLabels.allowOverlap` I mean on of:
- `{ plotOptions: { series: { dataLabels: { allowOverlap: false } } } }`
- `series: [{ dataLabels: { allowOverlap: false } }]`

Anyway, your demo (https://jsfiddle.net/vf2dw6xa/1/) does not show issue of overlapping two dataLabels.

danazelniker:
you can see it on the left side:
![image](https://user-images.githubusercontent.com/16417615/65154212-c2544280-da33-11e9-80a5-e27973dd3c17.png)

pawelfus:
Ah, you mean `stackLabel`! Thanks, you are right, it's broken, thanks for pointing this out! :) It's a regression - in v7.1 this used to work fine: https://jsfiddle.net/BlackLabel/q36safp9/


[...]

PR #12700, opened 2019-12-31
bugfix/12453-zeronode-sankey

### LINKED ISSUE #12453: Zero node are shown in dependency wheel
#### Expected behaviour
Nodes with only 0 weights are not shown

#### Actual behaviour
Nodes with only 0 weights are shown

http://jsfiddle.net/kzoon/9u5vfq7e/

Workaround: filter out the 0 values from your data

Basts:
Fixed #12453, node with 0 value was rendered on the Sankey and dependency wheel.

### DIFF
--- a/ts/modules/sankey.src.ts
+++ b/ts/modules/sankey.src.ts
@@ -648,8 +648,10 @@ seriesType<Highcharts.SankeySeries>(
                     totalNodeOffset;
 
                 for (var i = 0; i < column.length; i++) {
-                    if (column[i].getSum()) {
-                        totalNodeOffset = column[i].getSum() * factor + nodePadding;
+                    const sum = column[i].getSum();
+
+                    if (sum) {
+                        totalNodeOffset = sum * factor + nodePadding;
                     } else {
                         // If node sum equals 0 nodePadding is missed #12453
                         totalNodeOffset = 0;
--- a/ts/modules/sankey.src.ts
+++ b/ts/modules/sankey.src.ts
@@ -648,7 +648,6 @@ seriesType<Highcharts.SankeySeries>(
                     totalNodeOffset;
 
                 for (var i = 0; i < column.length; i++) {
-                    totalNodeOffset = column[i].getSum() * factor + nodePadding;
                     if (column[i].getSum()) {
                         totalNodeOffset = column[i].getSum() * factor + nodePadding;
                     } else {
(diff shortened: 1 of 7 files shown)


Basts:
@TorsteinHonsi 

This solution is the easiest way to not render the nodes for the point with set weight to 0 however, it changes the order of rendered points (what is logic because point and his node weren't rendered) - it is shown clearly in the failed visual tests. 

I just want to ask what do you think about this solution? 

Alternatively, what comes to my mind is to render this point, but with the width or height set to 0. Point also will be unvisible. But I think that it will be much more code and could be complicated and on the end, the output will be the same, but the order will be correct (which actually user is able to control while pasting data).

TorsteinHonsi:
Hi Basts, thank you for your PR!

This is a nice fix, but I think it is problematic that the order is modified. I can imagine implementers want to control the order by adding zero links. Can you explore the alternative option too? I assume we wouldn't actually _render_ the empty nodes, but they will exist in the model, and we need to make sure that the `nodePadding` is not added on each side of an empty node.

Basts:
> Hi Basts, thank you for your PR!
> 
> This is a nice fix, but I think it is problematic that the order is modified. I can imagine implementers want to control the order by adding zero links. Can you explore the alternative option too? I assume we wouldn't actually _render_ the empty nodes, but they will exist in the model, and we need to make sure that the `nodePadding` is not added on each side of an empty node.

Ok, so I will dig more into code to find another solution which will keep the order of rendered points. Thank you for the reply!

TorsteinHonsi:
Great. Actually I think the only thing you need to figure out is the `nodePadding`, then just avoid rendering 0 weighted links.

Basts:
> To do:
> 
> * Figure out `nodePadding` so that it is not added on both sides next to 0 nodes. Also test this for the first and last node, both on sankey and dependency wheel.
> * Don't render the graphics for links and nodes with 0 value.

@TorsteinHonsi 
Thank you for pointing out the guideline where to look for the solution. Currently, I am working on setOffset function in stackLabels (#12337, #12308, #12294, #11982). When I will finish it I will back to this issue.

Basts:
Visual tests shows that this solution breaks the organization chart. I will fix it.

Basts:
@TorsteinHonsi

I think that the PR is ready for the review. Before the fix the link from/to node with 0 weight was rendered, now it's not - that's the only difference which has left in the visual tests.

ts/modules/sankey.src.ts
```diff
@@ -648,7 +648,12 @@ seriesType<Highcharts.SankeySeries>(
                     totalNodeOffset;
 
                 for (var i = 0; i < column.length; i++) {
-                    totalNodeOffset = column[i].getSum() * factor + nodePadding;
+                    if (column[i].getSum()) {
+                        totalNodeOffset = column[i].getSum() * factor + nodePadding;
```
TorsteinHonsi: Put the results of `column[i].getSum()` into a variable (or rather a constant) so that we don't have to compute it twice.

Issue #12326, opened 2019-11-04
Column to treemap drilldown is not working properly


wchmiel:
#### Expected behaviour
After drillup the column series should be rendered properly.

#### Actual behaviour
After drillup the column series is broken.

#### Live demo with steps to reproduce
https://jsfiddle.net/BlackLabel/7f1tkva2/

1. Drill down to `treemap` series
2. Drill up to column series by clicking "Back to" button

#### Product version
Highcharts

TorsteinHonsi:
_Internal note_

I would focus the bug hunt around this part:

https://github.com/highcharts/highcharts/blob/cda57fe0b501a46406ab350f31ce3594ac5c100f/ts/modules/treemap.src.ts#L2176-L2195

In that code, the `max` and the `dataMax` is set to 100, which may explain at least parts of what we see in the demo. Seems like it is not unset after drilling up.

FYI @jon-a-nygaard (not assigning it to you, leaving it up for grabs, but any insight is valuable).

jon-a-nygaard:
> In that code, the max and the dataMax is set to 100, which may explain at least parts of what we see in the demo. Seems like it is not unset after drilling up.

Yes, I am quite sure this must be what is causing the issue.
I am not sure how to solve this, but it would be preferable in my opinion that the series is not modifying the `Axis.options`, but somehow is able to impact what options is used in `Axis.render`.

pawelfus:
It would be nice to have a common solution for options like that (not only on Axis instance, but also for Chart) - for example changing "column"->"bar" has the same limitation: https://jsfiddle.net/BlackLabel/phgc4oum/

Maybe improving `Chart.propFromSeries` method to work with drilldown module, then sharing it with Axis could work?

TorsteinHonsi:
I believe instead of letting the series hijack the whole axis, it should add an event listener to the axis, so that those properties are modified only if a treemap series is present. I am not sure at what stage it should go in and override the behaviour though, and it should probably not touch the options.

PR #12703, opened 2020-01-02
bugfix/12326-wrong-options-after-drilldown

wchmiel:
Fixed issue with wrong axis options after `drilldown` to treemap series, see #12326.
___

Also, fixed issue mentioned by @pawelfus with "column"->"bar" `drilldown`.

### DIFF
--- a/js/modules/drilldown.src.js
+++ b/js/modules/drilldown.src.js
@@ -131,7 +131,7 @@ import '../parts/Chart.js';
 import '../parts/Series.js';
 import '../parts/ColumnSeries.js';
 import '../parts/Tick.js';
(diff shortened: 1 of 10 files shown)


pawelfus:
Note: Once #12538 will be merged, we can replace `instanceof` with `series.is()`.

wchmiel:
So perhaps it will be a good idea to split those cases? One PR to solve the bug "column" -> "treemap" `drilldown`, resolving #12326 with `afterBindAxes` event. Another issue and PR to solve the problem with reflecting series types and options?

TorsteinHonsi:
Sorry for the delay! Yes a split is probably a good idea, then we can verify each of the cases before proceeding.

PR #12706, opened 2020-01-02
Tools: Quick test of using worker_threads compilation when compiling

activist:
~300% decrease in compilation time. Command `gulp scripts-compile`

Note that worker_threads were stable from node v12.11, so in case this change is merged later on we should change the "engines" version for node in package.json to v12.11 (unless there is something else in way). This is due to earlier node versions had to be run with `--experimental-worker` in order for this to work.


TorsteinHonsi:
Nice, thanks!

PR #12717, opened 2020-01-06
Fixed #12690, a11y module caused render issues for 3D funnel.

### LINKED ISSUE #12690: accessibility module causing render issues in funnel 3d chart
#### Expected behaviour

use the live demo below and remove accessibility.js

image

#### Actual behaviour

image

https://jsfiddle.net/r2skt4ju/

#### Product version
8.0.0

#### Affected browser(s)
reproduced in Chrome

oysteinmoseng:
Fixed #12690, a11y module caused render issues for 3D funnel.
___

A11y module reverses DOM order for series where it does not match order of points in points array. Changed to not do this for 3D charts.

### DIFF
--- a/ts/modules/accessibility/components/SeriesComponent/SeriesDescriber.ts
+++ b/ts/modules/accessibility/components/SeriesComponent/SeriesDescriber.ts
@@ -579,15 +579,17 @@ function describeSeriesElement(
  * @param {Highcharts.Series} series The series to add info on.
  */
 function describeSeries(series: Highcharts.AccessibilitySeries): void {
-    var chart = series.chart,
+    const chart = series.chart,
         firstPointEl = getSeriesFirstPointElement(series),
-        seriesEl = getSeriesA11yElement(series);
+        seriesEl = getSeriesA11yElement(series),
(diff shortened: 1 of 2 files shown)


ts/modules/accessibility/components/SeriesComponent/SeriesDescriber.ts
```diff
@@ -579,15 +579,17 @@ function describeSeriesElement(
  * @param {Highcharts.Series} series The series to add info on.
  */
 function describeSeries(series: Highcharts.AccessibilitySeries): void {
-    var chart = series.chart,
+    const chart = series.chart,
         firstPointEl = getSeriesFirstPointElement(series),
-        seriesEl = getSeriesA11yElement(series);
+        seriesEl = getSeriesA11yElement(series),
+        is3d = chart.options.chart?.options3d?.enabled;
```
pawelfus: Wouldn't be better to use `chart.is3d && chart.is3d()`?

oysteinmoseng: Oh cool, was not aware of this function. Will change

PR #12722, opened 2020-01-06
Fixed #12718, markers invisible on chart update with a11y module.

### LINKED ISSUE #12718: chart.update line markers not showing when accessibility module included
#### Expected behaviour
New line series should have their markers shown on updating chart

#### Actual behaviour
It doesnt.

https://jsfiddle.net/2745rxdt/1/

Push the "Click Me" button, new series markers are not visible. If you remove the accessibility module though, it seems to work fine.

#### Product version
Latest

@oysteinmoseng

oysteinmoseng:
Fixed #12718, markers invisible on chart update from null points with a11y module.
___

~Trying a simple fix first, if there are perceptible visual differences we'll have to get more clever.~

See comments.

### DIFF
--- a/ts/parts/Dynamics.ts
+++ b/ts/parts/Dynamics.ts
@@ -1040,10 +1040,15 @@ extend(Point.prototype, /** @lends Highcharts.Point.prototype */ {
 
             point.applyOptions(options);
 
-            // Update visuals
-            if (point.y === null && graphic) { // #4146
+            // Update visuals, #4146
+            // Handle dummy graphic elements for a11y, #12718
+            const hasDummyGraphic = graphic && point.hasDummyGraphic;
+            const shouldDestroyGraphic = point.y === null ? !hasDummyGraphic : hasDummyGraphic;
+            if (graphic && shouldDestroyGraphic) {
                 point.graphic = graphic.destroy();
+                delete point.hasDummyGraphic;
             }
+
             if (isObject(options, true)) {
                 // Destroy so we can get new elements
                 if (graphic && graphic.element) {
--- a/js/modules/accessibility/components/SeriesComponent/SeriesDescriber.js
+++ b/js/modules/accessibility/components/SeriesComponent/SeriesDescriber.js
@@ -52,7 +52,9 @@ function makeDummyElement(point, pos) {
     dummy.attr({
         'class': 'highcharts-a11y-dummy-point',
         fill: 'none',
-        opacity: 0
+        opacity: 0,
+        'fill-opacity': 0,
+        'stroke-opacity': 0
     });
     return dummy;
 }
@@ -73,6 +75,7 @@ function addDummyPointElement(point) {
     }, dummyElement = makeDummyElement(point, dummyPos);
     if (parentGroup && parentGroup.element) {
         point.graphic = dummyElement;
+        point.hasDummyGraphic = true;
         dummyElement.add(parentGroup);
         // Move to correct pos in DOM
         parentGroup.element.insertBefore(dummyElement.element, firstGraphic ? firstGraphic.element : null);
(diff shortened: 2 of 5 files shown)


oysteinmoseng:
Well then.

oysteinmoseng:
Different solution, setting a flag on point and handling in `point.update`.

Turns out this is hard to solve without recreating the graphic, as `point.update` tries to reuse the existing graphic regardless. This means that the marker will be wrong for e.g. circular markers, where the dummy `rect` element is reused, even if we can get the marker visibility correct.

Solved here by checking if the point has a dummy graphic and then destroying it in `point.update` when points are not null. Could alternatively be done in an event callback, at the performance cost of event dispatching for each point update. It's not much code, so I kept it in the core.

ts/parts/Dynamics.ts
```diff
@@ -1040,10 +1040,15 @@ extend(Point.prototype, /** @lends Highcharts.Point.prototype */ {
 
             point.applyOptions(options);
 
-            // Update visuals
-            if (point.y === null && graphic) { // #4146
+            // Update visuals, #4146
```
TorsteinHonsi: Can we avoid this juggling of conditions? Under what circumstances do we create a dummy graphic? Is it always added when the a11y module is loaded? In that case, maybe we can have a series property `series.useDummyGraphics`, and simplify the condition like so:

```js
if (point.y === null && graphic && !series.useDummyGraphics) {
    point.graphic = graphic.destroy();
}
```

oysteinmoseng: The dummy graphics are added for null points only - to ensure points always have a graphic if they should be announced to screen readers. So this condition needs to destroy graphics for:
- Points that are no longer null that have a dummy graphic
- Points that are null now and no longer should have a graphic, unless it has a dummy graphic

So it could be
``` javascript
// Option #1
if (
  graphic && (
    point.y !== null && point.hasDummyGraphic ||
    point.y === null && !point.hasDummyGraphic
  )
)
```
or slightly more explicit for future readers, the `shouldDestroyGraphic` definition could be replaced with
```javascript
// Option #2
const shouldDestroyGraphic = point.y !== null && hasDummyGraphic ||
    point.y === null && !hasDummyGraphic
```
which accomplishes the same as in the PR, but is a bit longer. Minification will reduce it down regardless, so it depends on what we find readable. I have no strong opinions either way, maybe leaning slightly towards Option 1 since it is pretty clear already that we are destroying the graphic.

pawelfus: Sorry to interrupt, but did you consider not using `point.graphic`? Something like `point.a11nGraphic` instead? Something you can freely add/remove in `describePointsInSeries`.

Then simple change change in `point.destroy` from 
```
            // Remove all events and elements
            if (point.graphic || point.dataLabel || point.dataLabels) {
                removeEvent(point);
                point.destroyElements();
            }
```
To:
```
            // Remove all events and elements
            if (point.shouldDestroyGraphic()) {
                removeEvent(point);
                point.destroyElements();
            }
```
Could help solving the issue? It's not the first time we wish to add a graphic-like element to a point which will be automatically removed (even when point is null) on point remove.

oysteinmoseng: Yes, it was considered, and might still be a viable option. However because a point should never have a dummy graphic and a real graphic at the same time, it will lead to more changes in the core - managing the lifecycle of the dummy graphic.

The dummy graphic will have to be managed in `point.update` as well as in `point.destroy`, and it will cause a domino effect in the a11y module having to look for `point.dummyGraphic` everywhere `point.graphic` does not exist. If `point.graphic` is overwritten at any point in the code (now or in the future), the `point.dummyGraphic` should also be removed, which could be a source of bugs.

TorsteinHonsi: Okay, I see it is not as simple as I thought.

PR #12747, opened 2020-01-09
tools/utilities-fx-class

bre1470:
Converted Fx to TypeScript class pattern.

### DIFF
--- a/ts/parts/Utilities.ts
+++ b/ts/parts/Utilities.ts
@@ -132,10 +132,11 @@ declare global {
                 prop: string
             );
             [key: string]: any;
-            public elem: (HTMLDOMElement|SVGElement);
+            public elem: (HTMLElement|SVGElement);
             public options: AnimationOptionsObject;
-            public paths: [SVGPathArray, SVGPathArray];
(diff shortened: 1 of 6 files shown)


ts/parts/Utilities.ts
```diff
@@ -3407,6 +3441,7 @@ if ((win as any).jQuery) {
 
 // TODO use named exports when supported.
 const utils = {
+    Fx,
```
KacperMadej: @bre1470 
I'm not sure if this is needed here since you have already set it up as `H.Fx = Fx` in 1270. I don't see any other option that is having both (it either exported via the `utils` or set up on `H`).

Any option you will choose, you might also want to check the `utils` reference in the `highcharts.src.js` master (https://github.com/highcharts/highcharts/blob/master/js/masters/highcharts.src.js)

I know it's already after the merge, but if any of this info will be of any help then great :D

bre1470: @KacperMadej 
Both is needed for migration, as one is for module export and the other for the global namespace.

Issue #12716, opened 2020-01-04
Moving the Navigator handles cause axis label to appear outside of chart


amerikan:
#### Expected behaviour

When resizing the Navigator handles, the chart labels should NOT be shown if data is not visible.

#### Actual behaviour

![Screen Shot 2020-01-03 at 8 59 04 PM](https://user-images.githubusercontent.com/179356/71760028-e6655380-2e6b-11ea-9622-18bab3e3d2e6.png)

#### Live demo with steps to reproduce
https://jsfiddle.net/v0k2f8o7/

Slowly drag the left handle to the left side in the navigator and pay attention to the chart labels.

You'll be able to see the behavior as shown below:

![Screen Shot 2020-01-03 at 9 00 37 PM](https://user-images.githubusercontent.com/179356/71760050-35ab8400-2e6c-11ea-8a57-5f89ab427d33.png)

#### Product version
In the jsfiddle Highstock JS v8.0.0 (2019-12-10) is used. This is true also for v6.2.0 which I am using.

#### Affected browser(s)
Tested on Firefox 71.0

pawelfus:
Thanks for reporting the issue @amerikan 

It's caused by ordinal xAxis. Did you consider disabling [`xAxis.ordinal`](https://api.highcharts.com/highstock/xAxis.ordinal) option? Take a look: https://jsfiddle.net/BlackLabel/trv8o7su/

Just a word of explanation: these labels shouldn't be visible because are pointing something outside the plotting area, not because data is not visible. There's no direct relation between xAxis label and dataset. If you need this, you should use [`xAxis.tickPositioner`](https://api.highcharts.com/highstock/xAxis.tickPositioner), for example like this: https://jsfiddle.net/BlackLabel/cu8fd20b/

PR #12761, opened 2020-01-10
bugfix/12716-axis-label-outside

wchmiel:
Fixed issue with axis labels outside plot range when ordinal enabled, see #12716.


ts/parts/Axis.ts
```diff
@@ -5496,6 +5498,20 @@ extend(Axis.prototype, /** @lends Highcharts.Axis.prototype */{
             }
             if (!tickPositionsOption && !tickPositioner) {
                 this.adjustTickAmount();
+
+                // Make sure ticks are within axis min and max range (#12716).
+                if (this.isOrdinal) {
```
pawelfus: Can we move this logic to `OrdinalAxis` file? E.g. add `afterAdjustTickAmount` event and modify config there.

ts/parts/Axis.ts
```diff
@@ -5496,6 +5498,20 @@ extend(Axis.prototype, /** @lends Highcharts.Axis.prototype */{
...
+                    min = this.min;
+                    max = this.max;
+
+                    tickPositions = tickPositions.filter(
+                        (pos: number): boolean =>
+                            defined(min) && pos >= min &&
+                            defined(max) && pos <= max
```
pawelfus: Instead of checking `defined(min)` in every iteration, can we don't call `filter` at all?

pawelfus:
General question: Why these ticks and labels are rendered in the first place? Instead of filtering out ticks, we shouldn't create them at all. Not that tick is not always present there:

ticks

Seems like some threshold is set there, once crossed label hides/shows up (half of the range, e.g. half day?).

PS: Since it's a Highstock issue, please add `Product: Highstock` label

wchmiel:
Additional tick is rendered because of this condition:
```js
while ((this.min as any) - minPointOffset > tickPositions[0]) {
    tickPositions.shift();
}
```

In some cases, this condition is false and tick is not removed.

TorsteinHonsi:
Pending @pawelfus 's review

ts/parts/OrdinalAxis.ts
```diff
@@ -1074,6 +1074,25 @@ addEvent(Axis, 'afterSetScale', function (this: Highcharts.Axis): void {
...
+
+// Make sure ticks are within axis min and max range (#12716).
+addEvent(Axis, 'afterTrimTicks', function (
+    this: Highcharts.Axis,
+    e: { tickPositions: Array<number> }
+): void {
+    var axis = this,
+        tickPositions = e.tickPositions;
+
+    if (axis.isOrdinal && axis.min && axis.max && tickPositions) {
```
pawelfus: Marginal case: this won't work with axis.min/max = 0

pawelfus:
Interesting. In theory `minPointOffset` incorporates ordinalCorrection. Data is irregular (for one month number of milliseconds varies) so it takes an effect. @TorsteinHonsi - how exactly `ordinalSlope` and `ordinalOffset` work? Is this kind of an approximation function (ax + b)?

TorsteinHonsi:
> Is this kind of an approximation function (ax + b)?

Yes, that's the idea. I don't have it all in my head, but I _think_ it applies to the extremes of the dataset.

PR #12764, opened 2020-01-10
Tools: Move highcharts-assembler code

activist:
Changed name of test main file from test.js to test-main.js, and added a few eslint ignore line rules, otherwise this is "as-is" in the highcharts-assembler repo.

Added mocha and chai as dependencies for now since that is what the highcharts-assembler uses for testing.


activist:
- [ ]  Move code
- [ ]  Move code

activist:
Going to move this into `tools/`folder after meeting we had today. Previous assumption was that since `exporting-server` had its own folder on the root the assembler would logically be on the same level.

activist:
Moved assembler code into `tools/`. Please review :)

samples/unit-tests/annotations/annotations-shapes/demo.details
```diff
@@ -1,6 +1,7 @@
 ---
  resources:
+   - https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.6/require.min.js
```
bre1470: Why is this needed?

package.json
```diff
@@ -27,7 +27,8 @@
   ],
   "scripts": {
     "test": "npx gulp test --force --speak",
-    "testall": "npx gulp test --browsers all --force --speak",
+    "testall": "npm run test:assembler && npx gulp test --browsers all --force --speak",
+    "test:assembler": "npx mocha tools/assembler/test/test-main.js",
```
bre1470: Why did you renamed `index.js` to `test-main.js`?

activist: Renamed back to `index.js`.

samples/unit-tests/annotations/annotations-shapes/demo.details
activist: Thanks for noticing. A sloppy eye on my end. I was playing with karma and forgot to undo this one.

activist:
> How will github.highcharts.com work when we request builds after this commit? Has that been tested? We should probably do the required modifications there and deploy at the same time as this one.

Yes, merging this can wait. But note that github.highcharts.com will still use the `highcharts-assembler` as-is. The main hurdle for github.highcharts.com will be when the `js/`-folder is removed from this repo as the application then will need to compile ts. I'm working on a solution for that now.

TorsteinHonsi:
Closing this, as we agreed to postpone refactoring the assembler.

PR #12774, opened 2020-01-13
bugfix/11148-polar-tracking-outside-plotarea

### LINKED ISSUE #11148: Trackable point outside yAxis extremes (polar)

raf18seb:
Fixed #11148, a point should not be trackable when is out of his pane. Also, isInsidePlot() works for panes now.

This PR is a continuation of this: https://github.com/highcharts/highcharts/pull/12348
I am not able to build a project and make changes on that branch anymore. The old PR can be closed.


ts/parts/Utilities.ts
```diff
@@ -3380,6 +3385,16 @@ H.isFunction = function (obj: unknown): obj is Function {
...
+H.isInsidePane = function (
+    x: number,
+    y: number,
+    center: Array<number>
+): boolean {
+    return Math.sqrt(
+        Math.pow(x - center[0], 2) + Math.pow(y - center[1], 2)
+    ) < center[2] / 2;
+};
+
```
pawelfus: Can we move this function to `Pane.ts` file?

ts/parts/Pointer.ts
```diff
@@ -834,7 +856,7 @@ Highcharts.Pointer.prototype = {
             chart.hoverPoint = hoverPoint;
 
             // Draw tooltip if necessary
-            if (tooltip) {
+            if (tooltip && (!chart.polar || hoverPoint.isInsidePane)) {
```
pawelfus: Instead of checking `hoverPoint.isInsidePane`, how about `fireEvent(this, 'afterGetHoverData', hoverPoint);` can remove hoverPoint? Something like this:

```js
let e = { hoverPoint };
fireEvent(this, 'afterGetHoverData', e);
hoverPoints = e.hoverPoint;
```

And `hoverPoint` will be undefined when outside the pane.

ts/parts/Utilities.ts
raf18seb: Yes. Should I leave this function hanging in the pane module's scope? Also, I would have to move 2 events from `Polar.ts` to `Pane.ts` that use this function. What do you think about moving all the new code (all 3 events) from `Polar.ts` to `Pane.ts`?

pawelfus: Do you mean `beforeGetHoverData` and `afterGetHoverData`? I'm fine with that.

TorsteinHonsi:
OK from me, awaiting @pawelfus 's approval 👍

raf18seb:
@pawelfus, changes made.

PR #12778, opened 2020-01-13
Tools: Wait for svg file writes to complete

activist:
There seems to be an issue with `karma-imagecapture-reporter` exiting before all file writes are complete. At least that is the theory, but it is hard to reproduce. Latest example is https://circleci.com/gh/highcharts/highcharts/25460 where there are no reported file write errors, which indicates that nothing has gone wrong (as it then should produce a log line with a short error message).

Thus I have added some code to make sure the files are written before karma is exiting, which is more or less same approach as [karma-junit-reporter uses](https://github.com/karma-runner/karma-junit-reporter/blob/2f43e5063e050b7853ebcc82c24ecc26310e0cff/index.js#L251-L257).

### DIFF
--- a/test/karma-imagecapture-reporter.js
+++ b/test/karma-imagecapture-reporter.js
@@ -15,6 +15,11 @@ function ImageCaptureReporter(baseReporterDecorator, config, logger, emitter) {
     baseReporterDecorator(this);
     const LOG = logger.create('reporter.imagecapture');
     const gitSha = getLatestCommitShaSync();
+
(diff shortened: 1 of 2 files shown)


activist:
Yet another example where the files were SVG seemingly doesn't exist even no write errors are reported: https://circleci.com/gh/highcharts/highcharts/25492?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link #12777

PR #12779, opened 2020-01-13
bugfix/11982-overlapping-stacklabels

Basts:
General stackLabels setOffset functionality refactoring. Touch #11982, #12308, #12337, #12294.


Basts:
@TorsteinHonsi 

There is a difference in the visual test because before the fix the default distance was a fixed value, now it is modulated by padding value.

Basts:
Most of the visual differences are related to the change the fixed distance value (which was set to 4) to the value defined by padding (which I set to the 5, as it is in the dataLabels). I can set it to 4, as the fixed value was, so the visual difference should be infinitesimal. But I found a bug in the last two visual test - label disappears when their y or x is set. I am going to fix it.

PR #12780, opened 2020-01-13
enhancement/highcharts-vertical-panning

### LINKED ISSUE #12671: Panning: maps panning type 'xy' not working
#### Behaviour
Looks like enabling vertical panning for Stock broke it for Maps.

@Denyllon 
https://github.com/highcharts/highcharts/commit/f5b56c90d4e254e616737c2b599a74c34f3b378a

The new state is added only for Stock. When debugging a Highmaps demo - the needed property is undefined.

By default panning of type 'xy' should be enabled for the `map` constructor in Highmaps - otherwise, the vertical panning feature will cause a regression for Highmaps.

BTW: `panningState` is added to colorAxis as well as to yAxis, so this could be fixed too.

I have checked if loading the Stock module will work as a workaround, but it doesn't help since panningState is being filled with `NaN`s - series.yData for map series is full of `undefined`.

fyi: @sebastianbochan 

https://jsfiddle.net/BlackLabel/q74vr8aj/

Zoom in and try panning vertically.
Only horizontal panning is visible.

#### Product version
8.0.0

Denyllon:
Added vertical panning to Highcharts. The `chart.panning.type` now accepts `y` directions.

~~Fixed #12671, added vertical panning to Highcharts.~~


ts/parts/Interaction.ts
```diff
@@ -899,6 +906,15 @@ extend(Chart.prototype, /** @lends Chart.prototype */ {
                             false,
                             { trigger: 'pan' }
                         );
+
+                        if (
+                            !chart.resetZoomButton &&
+                            type === 'xy' || type === 'y'
```
sebastianbochan: you can replace it with `match`

js/parts/Interaction.js
```diff
@@ -549,7 +549,7 @@ extend(Chart.prototype, /** @lends Chart.prototype */ {
                         axis.setExtremes(newMin, newMax, false, false, { trigger: 'pan' });
                         if (!chart.resetZoomButton &&
                             !hasMapNavigation &&
-                            type === 'xy' || type === 'y') {
+                            type.match('y') !== null) {
```
sebastianbochan: Not need to compare with null. its array.

ts/parts/Chart.ts
```diff
@@ -2888,3 +2888,45 @@ extend(Chart.prototype, /** @lends Highcharts.Chart.prototype */ {
...
+    ) {
+        let min: (number|undefined),
+            max: (number|undefined);
+
+        axis.series.forEach(function (series): void {
+            if (!(isNumber(min) && isNumber(max))) {
+                min = Number.MAX_VALUE;
+                max = Number.MIN_VALUE;
+            }
+            min = Math.min(H.arrayMin(series.yData as any), min) -
```
TorsteinHonsi: Get rid of the `any` casts. The whole block can be wrapped in a type check for `series.yData`

Denyllon:
@TorsteinHonsi @sebastianbochan Thank you for the review.
I also moved the logic of automatically disabling yAxis `startOnTick` and `endOnTick` from `stockChart` constructor into a Chart's `init` function, in order to keep consistency of API.

_NOTE: It's still not ready to merge, because some docs are missing._

ts/parts/Chart.ts
```diff
@@ -478,6 +478,24 @@ extend(Chart.prototype, /** @lends Highcharts.Chart.prototype */ {
...
+            ): Highcharts.YAxisOptions {
+                return merge(
+                    yAxisOptions,
+                    (verticalPanningEnabled ? {
+                        startOnTick: false,
+                        endOnTick: false
+                    } : null)
+                );
+            });
+
```
TorsteinHonsi: I think it is bad practice to modify the options set based on other options (although we do it in other places too). I think this code block should be rewritten to do the test in the places where `startOnTick` and `endOnTick` are actually used, in `Axis.ts`. You could create a private function that tests for `hasVerticalPanning`, to use in the conditions checking for `startOnTick` or `endOnTick`. Then refactor when you see more economic ways of doing it.

For example, in the top of `setTickPositions`, we could do:
```js
var .....,
    verticalPanning = hasVerticalPanning(chart),
    startOnTick = !verticalPanning && options.startOnTick,
    endOnTick = !verticalPanning && options.endOnTick
```

ts/parts/Axis.ts
```diff
@@ -5785,6 +5809,42 @@ extend(Axis.prototype, /** @lends Highcharts.Axis.prototype */{
...
 
+        // Extend the Axis to calculate start min/max values
+        // (including min/maxPadding). This is related to using vertical
+        // panning (#11315).
+        if (
+            panning &&
+            (
+                panning.type &&
+                panning.type.match('y')
+            ) &&
```
TorsteinHonsi: Use the new `hasVerticalPanning` test instead.

yarn.lock
```diff
@@ -0,0 +1,11289 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
TorsteinHonsi: We don't use yarn.lock in this repo, please delete it or .gitignore it.

ts/parts/Chart.ts
```diff
@@ -2885,6 +2886,20 @@ extend(Chart.prototype, /** @lends Highcharts.Chart.prototype */ {
...
+    /**
+    * Check whether the chart has vertical panning ('y' or 'xy' type).
+    *
+    * @private
+    * @function Highcharts.Chart#hasVerticalPanning
+    * @return {boolean}
+    *
+    */
+    hasVerticalPanning: function (
```
TorsteinHonsi: This function is only used in `Axis.ts`. To avoid polluting the Chart class, move it over to Axis.ts and make it a local function there.

Denyllon:
Things which are done:
- demo added to Highcharts API docs
- setting yAxis min/max values does not affect on vertical panning bounds
- `hasVerticalPanning` function moved to `Axis.ts`, and marked as private
- Setting yAxis `minPadding` and `maxPadding` works the same way like on x axis

ts/parts/Axis.ts
```diff
@@ -5785,6 +5812,41 @@ extend(Axis.prototype, /** @lends Highcharts.Axis.prototype */{
             axis.cleanStacks();
         }
 
+        // Extend the Axis to calculate start min/max values
```
TorsteinHonsi: What is the purpose of this block? Why not just use `axis.dataMin` and `axis.dataMax`? They are available at this point: https://jsfiddle.net/highcharts/65L208qv/.

ts/parts/Axis.ts
```diff
@@ -5364,8 +5382,14 @@ extend(Axis.prototype, /** @lends Highcharts.Axis.prototype */{
             tickPositionsOption = options.tickPositions,
             minorTickIntervalOption = this.getMinorTickInterval(),
             tickPositioner = options.tickPositioner,
-            startOnTick = options.startOnTick,
-            endOnTick = options.endOnTick;
+            hasVerticalPanning = this.hasVerticalPanning(),
+            isColorAxis = this.coll === 'colorAxis',
+            startOnTick = isColorAxis ?
```
TorsteinHonsi: More compact:
`startOnTick = !hasVerticalPanning && !isColorAxis && options.startOnTick,`

ts/parts/Axis.ts
Denyllon: The `panningState` object is required, because there is the difference of how **getExtremes()** function works on x and y axis, and we need to store initial data min and max values.

Let us take a situation, when a user zoomed the chart or set initial extremes on axis. In case of xAxis, `getExtremes` returns the `dataMin` and `dataMax` values, which are actual data extreme values of series related with this axis. However, after calling `getExtremes()` on yAxis, we're always getting values of currently visible range (excluding all data points located out of the current visible axis range).

https://jsfiddle.net/BlackLabel/ca0qku1x/

ts/parts/Axis.ts
Denyllon: Unfortunately, this does not work as expected, so I changed it to:
`startOnTick = (isColorAxis || !hasVerticalPanning) && options.startOnTick`

ts/parts/Axis.ts
TorsteinHonsi: Thanks for the explanation. I will try to refactor `Series.getExtremes` to make it return the data extremes, then we can call it here instead of recreating the series data loop. This is something we discussed earlier too, that the `getExtremes` function should return values rather than setting them on the instance.

TorsteinHonsi:
Pending myself to work on `Series.getExtremes`.

TorsteinHonsi:
### To do
 - [ ] Wait for https://github.com/highcharts/highcharts/pull/13162 to pass and be merged into the master.
 - [ ] Merge the master into the PR branch.
 - [ ] Refactor the calculation of axis extremes to use the new `Series.getExtremes`. Consider allowing a new parameter, `getExtremesFromAll`, to be passed into the function.

Denyllon:
After using the `Series.getExtremes` function, and adding the new parameter called `getExtremesFromAll` the result is the same - that function works on a processed data calculated before. After discussion we decided to split `Series.processData` function into a smaller parts responsible for getting and setting processed data, and then get it within vertical panning, but only when it's required.

- [ ] rebuild processData (split into smaller parts responsible for getting and setting processed data)
- [ ] move a whole logic of calculating panning state directly into a panning function (or also into separate functions)
- [ ] add flag `hasChanged` info panningState object, which will decide about when we should process data in order to get new panning state (e.g data updates)

TorsteinHonsi:
Pending tests after merge

Issue #12714, opened 2020-01-03
Chart.update on regular pie chart to 3d


kai2468:
#### Expected behaviour

Pie chart changes from a regular pie chart to a 3d one.
Looks to be a regression as it used to work in v7:
https://jsfiddle.net/BlackLabel/1036gu2f/

#### Actual behaviour
Pie does not change over to a 3d version

#### Live demo with steps to reproduce
https://jsfiddle.net/35yqdcga/1/
Push the "Click me" button to try changing the pie chart to a 3d pie. It doesnt work.

#### Product version
Latest.

pawelfus:
Bisected to commit af24cb21a1

PR #12787, opened 2020-01-14
bugfix/12714-pie-options3d-update

wchmiel:
Fixed issue with pie series update to 3d, see #12714.

### DIFF
--- a/samples/unit-tests/chart/options3d/demo.js
+++ b/samples/unit-tests/chart/options3d/demo.js
@@ -8,9 +8,9 @@ QUnit.test('options3d update', function (assert) {
             data: [1, 2, 3]
         }]
     });
-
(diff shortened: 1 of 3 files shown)


samples/unit-tests/chart/options3d/demo.js
```diff
@@ -1,5 +1,32 @@
...
+            options3d: {
+                enabled: true,
+                alpha: 45,
+                beta: 0
+            }
+        }
+    });
+
+    assert.ok(
+        !series.points[0].graphic.isFlagged,
```
pawelfus: Just a nitpicking: could you check instead if correct elements are rendered in DOM? 3D slice should have a few sides: `top`/`side1`/`side2`/`out` (with inner radius also `inn` side).

PR #12791, opened 2020-01-15
Tools: Added feature request template

cvasseng:
Converts ISSUE_TEMPLATE.md to the new issue template style, and adds a separate template for feature requests that automagically tags issues as such, and use a different template.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,22 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+#### Expected behaviour
+
+
+#### Actual behaviour
+
+
+#### Live demo with steps to reproduce
+<!-- template: https://jsfiddle.net/highcharts/LLExL/ -->
+
+#### Product version
+<!--- Highcharts, Highstock or Highmaps plus version number -->
+
+#### Affected browser(s)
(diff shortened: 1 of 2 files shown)


cvasseng:
We do have the "Type: Feature Request" label, but it's not used currently I guess.

My vote goes to using this one for the template, mainly so that we can easily sort out requests from a user (and then relabel when it's properly classified - judging from Uservoice there's going to be several that are actually bugs or something else), and because it's more apparent that it's a request and not a guaranteed thing. I feel the Enhancement label is a bit vague, primarily in the sense of "is this actually being worked on?". I do see a place for the Enhancement label though, but my gutfeel is that it's better suited as a tag assigned by us rather than by users. You guys are more in tune with how the labels flow though, so take this with a grain of salt. ;)

pawelfus:
Can we do this?
- "Type: Feature Request" for tagging github issues
- "Type: Enhancement" for tagging PR's (so change log generates automatically)

Won't that be too complicated?

activist:
I think @pawelfus suggestion sounds good.

TorsteinHonsi:
> "Type: Enhancement" for tagging PR's (so change log generates automatically)

Sorry for the belated answer, but the changelog only has two types of changes: Features and Bugfix. Currently features are those that are tagged changelog, but not bugfix. So I don't see a reason to complicate that.

But I'll merge the PR, I agree that Enhancement is a bit vague, that "Feature Request" is more descriptive. We can change the tagging accordingly for issues that we want to move over.

pawelfus:
> Currently features are those that are tagged changelog, but not bugfix

Hm.. it seems we have "Changelog: Feature" label 😅It seems that `"Type: Feature Request"` and `"Type: Enhancement" ` are just duplicates and I vote for b):
> b) rename our current label, or

Issue #12705, opened 2020-01-02
Chart.update() api does not align legend properly vertically, when font-size is changed


tejaspchordiya:
#### Expected behaviour
Chart.update() api should align legend properly vertically, when font-size is changed.

#### Actual behaviour
Chart.update() api when called with legend options with increased font-size, legend items gets vertically aligned to top. Ideally, it should vertically aligned in middle as the default case.

#### Live demo with steps to reproduce
[Demo](https://jsfiddle.net/5phbxv9s/2/)

pawelfus:
Thanks for reporting this issue @tejaspchordiya 

Comparison demo: https://jsfiddle.net/BlackLabel/n96c0bus/

sebastianbochan:
_Internal note_
Can be related with https://github.com/highcharts/highcharts/issues/12683

PR #12795, opened 2020-01-15
bugfix/12705-legend-font-size-update

kamilkulig:
Fixed issue with updating legend's font-size, see #12705.


samples/unit-tests/legend/legend-general/demo.js
```diff
@@ -118,4 +118,31 @@ QUnit.test('Legend redraws', function (assert) {
...
+        legend: {
+            itemStyle: {
+                fontSize: "24px"
+            }
+        }
+    });
+
+    assert.ok(
+        itemOriginalYPos < +chart.legend.allItems[0].legendItem.element.attributes.y.value,
+        'Item positioned correctly'
```
pawelfus: "Item **should** be positioned correctly"

TorsteinHonsi: Or more descriptive, "The item should be moved down" or something like that

PR #12796, opened 2020-01-16
bugfix/12736-mouse-over-sticky-tooltip

### LINKED ISSUE #12736: Mousing over tooltip should not dismiss it, move it, or change points.
#### Expected behaviour
Hover content, such as tooltips, are required by WCAG 2.1 to be persistent when hovered over. When hovering over the tooltip content, the tooltip should stay in place. This has several use cases, but particularly helps users of magnification tools.

#### Actual behaviour
Currently, if hovering over the tooltip causes the mouse pointer to be closer to a different point, the tooltip will move to that point instead.

#### Suggested improvement
Keep the tooltip fixed when mousing over it.

With this behavior, the tooltip may block access to points beneath it. We may want to allow the tooltip to be closed by clicking it, or add a close icon on tooltip hover.

https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/demo/line-basic/

Hover over the "Other" series, and try to hover over the tooltip. The tooltip will move to the above series.

bre1470:
Added new option, `tooltip.stickOnHover`, to avoid moving the tooltip when hovering on it and points were hiding below it.

Fixed #12736, Mouse over tooltip could change it.


bre1470:
There is no handling (manual close) for oversized tooltips integrated and also the tooltip text is still not selectable (which should be a separate option anyway).

ts/parts/Pointer.ts
```diff
@@ -1716,6 +1725,75 @@ Highcharts.Pointer.prototype = {
...
+    isStickyTooltip: function (this: Highcharts.Pointer, e: Highcharts.PointerEventObject): boolean {
+        const chart = this.chart;
+        const point = chart.hoverPoint;
+        const tooltip = chart.tooltip;
+        const eventPosition: Highcharts.PositionObject = {
+            x: e.chartX,
+            y: e.chartY
+        };
+
+        const absoluteBounding = function (element: Highcharts.SVGElement): Highcharts.BBoxObject {
```
TorsteinHonsi: Instead of the new `absoluteBounding` function , try using the existing `Highcharts.offset` function. It reports the position relative to the _page_, but then `chart.pointer.chartPosition` can be substracted.

ts/parts/Tooltip.ts
```diff
@@ -345,6 +346,13 @@ H.Tooltip.prototype = {
          */
         this.isHidden = true;
 
+        /**
+         * @readonly
+         * @name Highcharts.Tooltip#isSticky
+         * @type {boolean}
+         */
+        this.isSticky = (options.sticky || false);
```
TorsteinHonsi: We don't really need a property on the `Tooltip` since this is a simple reflection of the option. Instead, we can add the default value `false` directly to the tooltip default options, and read from `options.sticky` whenever needed. That is the common and simplest pattern as long as we're not dealing with options that depend on some condition.

ts/parts/Options.ts
```diff
@@ -3675,6 +3676,16 @@ H.defaultOptions = {
...
          */
 
+        /**
+         * Prevents the tooltip from switching or closing, when touched or
+         * pointed.
+         *
+         * @type      {boolean}
+         * @since     8.0.1
+         * @product   highcharts highstock
+         * @apioption tooltip.sticky
```
TorsteinHonsi: Is `sticky` a descriptive name? I think something like `stickOnHover` better describes the feature.

ts/parts/Pointer.ts
bre1470: This should be solved.

ts/parts/Tooltip.ts
bre1470: This should be solved.

ts/parts/Options.ts
bre1470: This should be solved.

ts/parts/Pointer.ts
```diff
@@ -1716,6 +1725,68 @@ Highcharts.Pointer.prototype = {
 
     },
 
+    /**
+     * @private
+     * @param {Highcharts.PointerEventObject} e
+     * @return {boolean}
+     */
+    isStickyTooltip: function (this: Highcharts.Pointer, e: Highcharts.PointerEventObject): boolean {
```
pawelfus: Could you add a description what method actually does? From `Pointer.isStickyTooltip(event)` method name I expect something quite simple (compare for example with `chart.is3d()` or `series.is('seriesType')`, but it checks also DOM and position (?).

ts/parts/Options.ts
```diff
@@ -3677,6 +3678,19 @@ H.defaultOptions = {
...
+        /**
+         * Prevents the tooltip from switching or closing, when touched or
+         * pointed.
+         *
+         * @sample highcharts/tooltip/stickonhover/
+         *         Tooltip sticks on hover event
+         *
+         * @type      {boolean}
+         * @since     8.0.1
+         * @product   highcharts highstock
```
pawelfus: No support for Gantt and Highmaps?

ts/parts/Pointer.ts
bre1470: This should be solved with the added description. Let me know if something is unclear.

ts/parts/Options.ts
bre1470: Good catch! I added support by removing the tag. :)

ts/parts/Pointer.ts
pawelfus: Aaah! Yes, that's clear. Thank you! :)

bre1470:
@TorsteinHonsi Please review again.

bre1470:
Or better later. I will have to solve the merge conflicts first.

PR #12797, opened 2020-01-16
Refactor SVG paths to array of typed arrays

TorsteinHonsi:
This change modifies all path definitions from a one-dimensional array to an array of strongly typed arrays designating segments.

Before:
`['M', 0, 0, 'L', 100, 100]`

After
`[
  ['M', 0, 0],
  ['L', 100, 100]
]`

The *benefits* are strong type checking on all segments, and easier computation on arrays, particularly during animation, where we can compute on each segment instead of slicing and splicing on a one-dimensional array. The *downsides* are compatibility, risk of regressions during refactoring, and that we need a backwards-compat function in the middle to avoid breaking custom code.

#### To do
 - [x] Make all unit tests pass
 - [x] Make all visual tests pass
 - [x] Add text description for error 32
 - [x] Change some (not all) of the demos/tests using the old syntax
 - [x] Work on a more graceful `pathToSegment` function.


samples/unit-tests/3d/column-animation/demo.js
```diff
@@ -1,6 +1,8 @@
 /* eslint func-style:0 */
 
-QUnit.test('Point animation', function (assert) {
+// Skipped, can't seem to find out why this fails after SVG path refactoring.
```
bre1470, line 3: Maybe it is a timing problem?

samples/unit-tests/svgrenderer/pathtosegments/demo.js
```diff
@@ -0,0 +1,120 @@
...
+    testUpperAndLower(
+        ['M', 0, 0, 'A', 1, 1, 1, 1, 1, 1, 1],
+        [['M', 0, 0], ['A', 1, 1, 1, 1, 1, 1, 1]],
+        'Arc'
+    );
+
+    // Fails
+    /*
+    assert.deepEqual(
+        pathToSegments(['M', 0, 0, 0, 'L', 1, 1]),
```
bre1470, line 113: Maybe it fails, because it is missing the inner arrays.

TorsteinHonsi: It fails since I changed `pathToSegments` function to be more sloppy, using two `any` casts 😬.

ts/parts/SplineSeries.ts
```diff
@@ -253,10 +253,10 @@ seriesType<Highcharts.SplineSeries>(
...
             ret = [
                 'C',
-                pick(lastPoint.rightContX, lastPoint.plotX),
-                pick(lastPoint.rightContY, lastPoint.plotY),
-                pick(leftContX, plotX),
-                pick(leftContY, plotY),
+                pick(lastPoint.rightContX, lastPoint.plotX) as any,
+                pick(lastPoint.rightContY, lastPoint.plotY) as any,
+                pick(leftContX, plotX) as any,
+                pick(leftContY, plotY) as any,
```
bre1470, line 259: This is just missing a final non-null/undefined value: `pick(…, …, 0)` to avoid the `any` cast.

ts/parts/SvgRenderer.ts
```diff
@@ -83,6 +104,9 @@ declare global {
             style?: (string|CSSObject);
             translateX?: number;
             translateY?: number;
+            // width?: number;
+            // x?: number;
+            // y?: number;
```
bre1470, line 109: Are these future property types, when we migrated the usage from `number/string` to `number` only?

ts/annotations/controllable/ControllablePath.ts
```diff
@@ -148,32 +148,31 @@ merge<Highcharts.AnnotationControllablePath, Partial<Highcharts.AnnotationContro
...
+
+                while (++pointIndex < len && showPath) {
+                    point = points[pointIndex];
+                    command = point.command || 'L';
+                    position = this.anchor(point).absolutePosition;
+
+                    if (command === 'M') {
+                        d.push([command, position.x, position.y]);
+                    } else if (command === 'L') {
+                        d.push([command, position.x, position.y]);
```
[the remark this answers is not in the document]
TorsteinHonsi: One should think so, but apparently TypeScript gets confused as to whether this is a `SVGPathLineTo` or `SVGPathMoveTo`. If you have a proposal to fix that, it would be great.
`Argument of type '["L" | "M", number, number]' is not assignable to parameter of type 'SVGPathSegment'`

ts/parts-3d/SVGRenderer.ts
```diff
@@ -221,7 +221,7 @@ SVGRenderer.prototype.toLineSegments = function (
         m = true;
 
     points.forEach(function (point: Highcharts.PositionObject): void {
-        result.push(m ? 'M' : 'L', point.x, point.y);
+        result.push(m ? ['M', point.x, point.y] : ['L', point.x, point.y]);
```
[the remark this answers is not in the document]
TorsteinHonsi: Same as above:

`Argument of type '["L" | "M", number, number]' is not assignable to parameter of type 'SVGPathSegment'`

ts/parts/SplineSeries.ts
TorsteinHonsi: Thanks, will do a bit of rewrite here.

ts/parts/SvgRenderer.ts
TorsteinHonsi: I don't remember the context, but I vaguely remember wanting these properties there, but something failed.

ts/annotations/controllable/ControllablePath.ts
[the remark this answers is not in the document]
bre1470: I see. I did just noticed, that it was (kind of) working here: https://github.com/highcharts/highcharts/blob/022878f65ff240f81a5e107818dfcb7b45ac9264/ts/parts-more/RadialAxis.ts#L766, but that is probably because of `path` beeing of type `Array<any>`.

samples/unit-tests/3d/column-animation/demo.js
TorsteinHonsi: Turns out it's a real problem with path animation. It jumps to the end path before it starts animating. Thanks for noticing, I will work on a fix.

TorsteinHonsi: Fixed

bre1470: That was fast! 😄

ts/annotations/controllable/ControllablePath.ts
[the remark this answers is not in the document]
TorsteinHonsi: It was `xy` that was `any`. Fixed that now.

PR #12821, opened 2020-01-21
enhancement/11296-toggle-fullscreen

raf18seb:
Fullscreen module added. Allows using _chart.toggleFullscreen()_ method.


raf18seb:
@bre1470 , could you take a look at (**chart as any**) in _full-screen.src.ts_ file and maybe propose a solution?

ts/modules/full-screen.src.ts
```diff
@@ -18,63 +15,169 @@ import H from '../parts/Globals.js';
  */
 declare global {
     namespace Highcharts {
+        interface Chart {
+            browserProps: {
+                fullscreenChange: string;
+                requestFullscreen: string;
```
bre1470: Here you should set requestFullscreen type to the literal string types, that are possible: `requestFullscreen: ('msRequestFullscreen'|'mozRequestFullscreen'|'requestFullscreen'|'webkitRequestFullscreen')`

ts/modules/full-screen.src.ts
```diff
@@ -18,63 +15,169 @@ import H from '../parts/Globals.js';
...
+            if (chart.isFullscreen) {
+                chart.isFullscreen = false;
+                chart.closeFullscreen();
+            } else {
+                chart.isFullscreen = true;
+            }
+        }
+    );
+
+    (chart as any).container.parentNode[chart.browserProps.requestFullscreen]();
```
bre1470: parentNode needs a check agains `instanceof Element`, so that you can use the full screen function type safe.
```
    if (chart.container.parentNode instanceof Element) {
        chart.container.parentNode[chart.browserProps.requestFullscreen]();
    }
```

raf18seb: Thanks, I made changes, could you take a look once again? Especially at Globals.ts and `instanceof Document`

ts/modules/full-screen.src.ts
```diff
@@ -18,63 +15,189 @@ import H from '../parts/Globals.js';
...
-    init: function (
-        this: Highcharts.FullScreen,
-        container: Highcharts.HTMLDOMElement
-    ): void {
-
-        let promise;
-
-        if (container.requestFullscreen) {
-            promise = container.requestFullscreen();
+H.Chart.prototype.toggleFullscreen = function (this: Highcharts.Chart): void {
```
bre1470: You can remove the `this` argument as `H.Chart.prototype` provides the scope automatically to the function. (And with that you can remove the off-flag for the linting rule.)

ts/modules/full-screen.src.ts
```diff
@@ -18,63 +15,189 @@ import H from '../parts/Globals.js';
...
+    var chart = this,
+        container = chart.container.parentNode as Highcharts.HTMLDOMElement;
```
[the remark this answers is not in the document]
raf18seb: Hmm, I wonder why linter didn't catch vars...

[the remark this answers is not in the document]
bre1470: It looks like it is turned off for the migration phase. Still it would be good practice to use `let` at least. `let` and `const` get both transpiled, but it helps with the typing checks.

ts/modules/full-screen.src.ts
```diff
@@ -18,63 +15,194 @@ import H from '../parts/Globals.js';
...
+        exportDivElements.length &&
+        exportingOptions?.menuItemDefinitions?.viewFullscreen &&
+        menuItems &&
+        chart.options.lang?.viewFullscreen
+    ) {
+        exportDivElements[menuItems.indexOf('viewFullscreen')].innerHTML = (
+            exportingOptions.menuItemDefinitions.viewFullscreen.text ||
+            chart.options.lang.viewFullscreen
+        );
+    }
```
TorsteinHonsi: This code block is very similar to the above - refactor it out to a function. It can be local to the module.

ts/modules/full-screen.src.ts
```diff
@@ -18,63 +15,194 @@ import H from '../parts/Globals.js';
...
         }
+    }
 
-        if (promise) {
-            promise['catch']((): void => { // eslint-disable-line dot-notation
-                alert('Full screen is not supported inside a frame'); // eslint-disable-line no-alert
-            });
-        }
+    if (!chart.isFullscreen) {
+        chart.openFullscreen();
```
TorsteinHonsi: The error message is now removed. Previously I would get an alert when trying to full-screen a chart from inside a frame, now it just silently fails but with a JS error in the console. This can be tested in the utils.

bre1470:
> We could do that either by keeping them private to the module or reintroducing a Fullscreen object to control it. Then it would be a one-to-one relation between `Chart` and `Fullscreen`. What is your opinion on this @bre1470 ?

I would find it helpful to have a new class like `FullscreenChartController` with these functions. It could be set as a backreference on the chart instance like `(chart as Partial<FullscreenChart>.fullscreen = new FullscreenChartController(chart);` during `Chart#event:load`. This also can become handy for modifications with accessibility.

If we go with private functions in the module scope instead, we have to fire a bunch of events with context object to allow modification: `Chart#event:fullscreenOpen`, `Chart#event:afterFullscreenOpen`, `Chart#event:fullscreenClose`, `Chart#event:afterFullscreenClose`, `Chart#event:fullscreenToggle`, `Chart#event:afterFullscreenToggle`

TorsteinHonsi:
Great, let's make it a controller object. Except the `FullscreenChartController` name is a bit on the verbose end compared to the current practice. We have a number of objects already where the class has the same name as the instance:

```js
chart.legend = new Legend;
chart.tooltip = new Tooltip;
chart.navigator = new Navigator;
chart.rangeSelector = new RangeSelector;
```

Continuing this convention, I propose
`chart.fullscreenController = new FullscreenController(chart)`

raf18seb:
All methods are stored on `chart.fullscreenController` now. Also, I added `catch-alert` when someone is trying to open fullscreen inside an iframe.

I have 2 `as any` like in the other classes (Tooltip, Navigator, etc.). Is it ok or we can avoid it somehow?

ts/modules/full-screen.src.ts
```diff
@@ -1,80 +1,237 @@
...
+     *
+     * @since       next
+     *
+     * @sample      highcharts/members/chart-togglefullscreen/
+     *              Toggle fullscreen mode from a HTML button
+     *
+     * @function    Highcharts.Chart#toggleFullscreen
+     * @return      {void}
+     * @requires    modules/exporting
+     * @requires    modules/fullscreen
```
bre1470: It is `modules/full-screen`

ts/modules/full-screen.src.ts
```diff
@@ -1,80 +1,237 @@
...
+ * @type {boolean|undefined}
+ * @since next
  */
-var FullScreen = H.FullScreen = function (
-    this: Highcharts.FullScreen,
-    container: Highcharts.HTMLDOMElement
+
+/* eslint-disable no-invalid-this, valid-jsdoc */
+
+H.FullscreenController = function (
```
bre1470: Please write this in ES6 class pattern with doclets. That way you also avoid the `any` casts:
```ts
/**
 * ... Class description ...
 * @class
 * @name Highcharts.FullscreenController
 * @hideconstructor
 * @requires modules/exporting
 * @requires modules/full-screen
 */
class FullscreenControll {
    /* *
     *
     *  Constructors
     *
     * */
    /** @private */
    public constructor(chart: Highcharts.Chart) {
        this.chart = chart;
        this.browserProps = { ... };
    }
    /* *
     *
     *  Properties
     *
     * */
    /**
     * Chart managed by the fullscreen controller.
     * @name Highcharts.FullscreenController#chart
     * @type {Highcharts.Chart}
     */
    public chart: Highcharts.Chart;
    /** @private */
    public browserProps: Highcharts.FullscreenControll['browserProps'];
    /* *
     *
     *  Functions
     *
     * */
    /**
     * ... Description ...
     * @function Highcharts.FullscreenController.toggleFullscreen
     */
    public toggleFullscreen(): void {
    }
}
addEvent(Chart, 'beforeRender', function (this: Highcharts.Chart): void {
    /**
     * ...Description...
     * @name Highcharts.Chart#fullscreenController
     * @type {Highcharts.FullscreenController}
     * @requires modules/exporting
     * @requires modules/full-screen
     */
     this.fullscreenController = new FullscreenController(this);
 });
H.FullscreenController = FullscreenController;
export default H.FullscreenController;
```

ts/parts/Globals.ts
```diff
@@ -87,15 +87,19 @@ declare global {
...
+        webkitExitFullscreen: Function;
         webkitHidden: boolean;
     }
     interface Element {
         mozRequestFullScreen: Function;
         msMatchesSelector: Element['matches'];
         msRequestFullscreen: Function;
         webkitMatchesSelector: Element['matches'];
-        webkitRequestFullscreen: Function;
+        webkitRequestFullScreen: Function;
```
bre1470: Please add the `/** @deprecated */` doclet in front of each vendor-specific implementation.

bre1470:
Please note, that we decided to have the FullscreenController public, so that the `Chart#fullscreenController#toggleFullscreen` function is available to users. The rest of the fullscreenController, except for a new `isFullscreen` property maybe, should be private.

TorsteinHonsi:
> @TorsteinHonsi What do you think?
I think that now that we don't have to worry about polluting the `chart`, it may also be handy to have the open and close methods exposed. I imagine that toggling is not always what implementers would want. Sometimes it may be more beneficial to do `if it is closed, openIt(), otherwise doNothing()`.

ts/modules/full-screen.src.ts
raf18seb: Actually, I think it is not needed at all because the `full-screen` module is a part of the `exporting` module - not a separate one.

bre1470: If you look into the masters, you will find a `full-screen` module, independent of the `exporting` module, as the `exporting` module is only required for the menu.

bre1470: @TorsteinHonsi `full-screen` is part of the `exporting` package, which looks to me like a bug, as there is nothing related between both except for the menu point. Can we remove it from there as a dependency?

ts/modules/full-screen.src.ts
```diff
@@ -227,11 +121,187 @@ H.FullscreenController.prototype = {
...
+     *
+     *  Properties
+     *
+     * */
+    /**
+     * Chart managed by the fullscreen controller.
+     * @name Highcharts.FullscreenController#chart
+     * @type {Highcharts.Chart}
+     */
+    public chart: Highcharts.Chart;
```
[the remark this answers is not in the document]
raf18seb: So it is not as is your previous comment? Should I move lines 130-135 into line 124?

ts/modules/full-screen.src.ts
```diff
@@ -227,11 +121,187 @@ H.FullscreenController.prototype = {
...
+     *
+     * @since       next
+     *
+     * @sample      highcharts/members/chart-togglefullscreen/
+     *              Toggle fullscreen mode from a HTML button
+     *
+     * @function Highcharts.FullscreenController#toggleFullscreen
+     * @return      {void}
+     * @requires    modules/exporting
+     * @requires    modules/full-screen
```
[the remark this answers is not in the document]
raf18seb: Ok, so the fact that `full-screen` module is a part of `exporting` is actually ok?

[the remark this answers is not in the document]
raf18seb: Nevermind, it's 'exporting', not 'full-screen'.

ts/modules/full-screen.src.ts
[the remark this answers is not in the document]
bre1470: Yes, my example code was wrong for this case. It is the line `this.chart = chart;` in the constructor code.

ts/modules/full-screen.src.ts
```diff
@@ -1,80 +1,309 @@
...
+     * */
+
     /**
-     * Init function
+     * Stops displaying the chart in fullscreen mode.
+     * Exporting module required.
+     *
+     * @since       next
+     *
+     * @function Highcharts.FullscreenController#toggleFullscreen
```
bre1470: The name has to end with `#closeFullscreen`.

ts/modules/full-screen.src.ts
bre1470: The conclusion is that we can not remove it from `exporting` package as it is now expected there.

raf18seb:
Thanks for all suggestions! I made all changes, it is ready for a final review.

ts/modules/full-screen.src.ts
```diff
@@ -71,127 +70,22 @@ declare global {
...
-     * @function    Highcharts.Chart#toggleFullscreen
-     * @return      {void}
-     * @requires    modules/exporting
-     * @requires    modules/fullscreen
+     * @param {Highcharts.Chart} chart
      */
-    toggleFullscreen: function (): void {
-        const fullscreenController = this,
-            chart = fullscreenController.chart;
+    public constructor(chart: Highcharts.Chart) {
```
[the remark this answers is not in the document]
bre1470: I am sorry, but somehow the `@private` tag is removing the class from the TypeScript declarations. So my previous request is wrong and the `@private` tag has to be removed. If you like, you can restore your constructor doclet from a previous commit. I will later investigate why `@hideconstructor` is not working (#12909).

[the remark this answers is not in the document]
raf18seb: Sure, I removed the `@private` tag. I don't know much about how the doclets work inside so, when you figure that out, could you commit this part or provide me the proper doclet?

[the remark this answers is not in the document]
bre1470: Yes, I can do that. I just hesitate to commit randomly in other peoples branches. :)

ts/modules/full-screen.src.ts
```diff
@@ -95,11 +94,7 @@ class FullscreenController {
          */
         this.chart = chart;
 
-        // IE8 compatibility
-        const HTMLelement = typeof HTMLElement === 'undefined' ?
-            Element : HTMLElement;
-
-        if (!(chart.container.parentNode instanceof HTMLelement)) {
+        if (!(chart.container.parentNode instanceof Element)) {
```
sebastianbochan: Check if `.is()` will work here.

bre1470: What do you mean? `isDOMElement`?

sebastianbochan: My mistake, ignore my idea.

PR #12822, opened 2020-01-21
bugfix/10162-text-not-visible-due-to-outline #10162

### LINKED ISSUE #10162: Labels do not appear in pie charts if text is in RTL language
#### Expected behaviour
The labels in Arabic and Hebrew should appear

#### Actual behaviour
The charts shows only text in English and Simplified Chinise
But Arabic and Hebrew text is not displayed at all
image

https://jsfiddle.net/yat9L67p/1/

#### Product version

Highcharts
#### Affected browser(s)
Google Chrome

mateuszkornecki:
Fixed #10162, RTL text was covered when using a text outline.


samples/unit-tests/datalabels/i18n/demo.js
```diff
@@ -0,0 +1,55 @@
+QUnit.test('RTL characters on data labels', function (assert) {
```
TorsteinHonsi: The text outline is general to all labels, not only data labels. So instead of creating a full chart instance in order to run a test on its labels, we should run the test on the `SVGRenderer.text` API directly. Rewrite the test and move it to `unit-tests/svgrenderer/text/demo.js`.

PR #12818, opened 2020-01-21
enhancement/11308-center-column-with-nulls

raf18seb:
Added new property `series.ignoreNulls` which centers the columns within a category when other series have a null value or missing points.

This PR is a continuation of https://github.com/highcharts/highcharts/pull/12460

PR #12846, opened 2020-01-24
Added new stacking option, category-center

### LINKED ISSUE #11308: New functionality: Centering columns inside the categories with nullPoints
When null points are present in the data (and should be ignored on graph), they should not occupy space in related category.

Current version:
Screenshot 2019-07-04 at 11 55 28

Current workaround:
http://jsfiddle.net/hgjvpqab/

Demo with a workaround:
Screenshot 2019-07-04 at 11 55 36

TorsteinHonsi:
Added new option,  `plotOptions.column.centerInCategory`. It centers columns within each category group, but avoiding gaps for `null` or missing points.

Fixes #11308.

This PR offers an alternative to #12818. By tapping into the existing stacking logic, we avoid having to create lots of logic that basically deals with the same problem - points on the same X value need to be aware of each other. By using the stacking logic we simply reduce the logic to compute the final position of the columns.

### To do
 - [x] Update sample and docs
 - [x] Implement in combination with simple and multiple stacks
 - [x] Check out multiple points at the same X value
 - [x] Move over and adapt relevant unit tests from #12818
 - [x] Fix two issues mentioned in [comment](https://github.com/highcharts/highcharts/pull/12846#issuecomment-579196086) (Torstein).


pawelfus:
Nice! 
But I have one problem with this approach: user can't use regular stacking with stacks **and** centered columns, for example this output is impossible (manually edited SVG):
- (`stack: 'A', stack: 'B'` + `stacking: 'normal'` + centered columns):
Screen Shot 2020-01-24 at 13 03 53

We can get only one of those:
- (`stack: 'A', stack: 'B'` + `stacking: 'normal'`):
Screen Shot 2020-01-24 at 13 05 53

- (`stack: 'A', stack: 'B'` + `stacking: 'category-center'`):
Screen Shot 2020-01-24 at 13 06 14

TorsteinHonsi:
> user can't use regular stacking with stacks and centered columns

Yes... the question is if that edge case is worth all the extra code. Alternatively, we should be able to combine the two concepts to avoid parallel code.

pawelfus:
I'm not sure if we can call it an edge case: users simply want columns centered over a category. Just a design change, position of a column.

Stacking in this case is just visual representation of a bar components. For me it's another feature that I would expect to be compatible with the solution above.

raf18seb:
Nice, I agree that the performance is much better and we have less new code. But, as Paweł said, we are missing stacking + centering and also we are missing more customization options like e.g. "fill available space" (columns have different width depending on a number of points in category).

Also, I got used to filling the whole available space by setting `series.pointPadding` and `series.groupPadding` to 0: http://jsfiddle.net/BlackLabel/y5za097e/
There's also a small problem with adding multiple same points.

TorsteinHonsi:
Okay, see the new commit. I now removed the centering from the `stacking` options, to avoid conflicts there. So now it should work in combination with `stacking` (except it looks like crap currently :).

The new option is `centerInCategory`. 

The change is that is still uses the stacking logic, but instead of being a separate stack option, it creates its own stack just for this purpose. This is not a new concept, Highcharts already supports positive and negative stacks for the same series/points, so I just added a `group` stack.

With this approach we get all the X-value mapping and dynamic stuff for free.

Now we need to figure out how _real_ stacks affects the positioning. 

> we are missing more customization options like e.g. "fill available space"

If we want those, it should be simple to implement it in the last scaling step in `ColumnSeries.translate`.

ts/parts/Stacking.ts
```diff
@@ -661,8 +670,14 @@ Axis.prototype.cleanStacks = function (this: Highcharts.Axis): void {
  * @function Highcharts.Series#setStackedPoints
  * @return {void}
  */
-Series.prototype.setStackedPoints = function (this: Highcharts.Series): void {
-    if (!this.options.stacking ||
+Series.prototype.setStackedPoints = function (
+    this: Highcharts.Series,
+    stackingParam: string
```
pawelfus: I think this param should be optional

ts/parts/Stacking.ts
```diff
@@ -652,7 +656,12 @@ Axis.prototype.cleanStacks = function (this: Highcharts.Axis): void {
 };
 
 
-// Stacking methods defnied for Series prototype
+// Stacking methods defined for Series prototype
+Series.prototype.setGroupedPoints = function (this: Highcharts.Series): void {
+    if (this.options.centerInCategory) {
+        Series.prototype.setStackedPoints.call(this, 'group');
```
pawelfus: Hmm.. why not simply `this.setStackedPoints('group')`?

TorsteinHonsi: Because many of the box-like series types override `setStackedPoints`. Haven't tested it though.

pawelfus: Ah, that's the case! Thanks 👍

TorsteinHonsi:
I think I've got it working with stacking now. Tested no stacking, single stack and multiple stacks. @raf18seb will you move over and adapt the relevant unit tests?

raf18seb:
Still isn't working with the different series type (see commented line series) and isn't bulletproof for data commented in line 19: http://jsfiddle.net/BlackLabel/9gk152zv/

sebastianbochan:
Im curious, what is the reason of removing / comment test? Is it temporary solution for Safari?

shashank1010:
Just curious to know by when would this be tagged in a release.

sebastianbochan:
The release should be available this week (if all test will pass).

PR #12847, opened 2020-01-24
bugfix/12653-custom-options

### LINKED ISSUE #12653: Map in Highcharts throw error when data object has "state" property
#### Expected behaviour
Highcharts shouldn't throw error when object has property such as "state"

#### Actual behaviour
Highcharts throw error when data object has "state" property

https://codepen.io/gopinathshiva/pen/BayQMKx

1. Click on any of the lat lon
2. Then click on any of the 2 lat lon points
3. Find error in console

#### Product version
8.0.0

#### Affected browser(s)
All

bre1470:
Introducing the new options `plotOptions.series.custom`, `series.custom`, and `series.data.custom` to store additional data for custom formatters and custom event callbacks.

* Fixed #12653, added support for sub option `custom`.


pawelfus:
I don't quite understand the need for this feature. In the test I can see nested properties used in sorting module. Does this PR add an option to use `point.myValue` instead of `point.custom.myValue`? Custom formats already work that we can use this: https://jsfiddle.net/BlackLabel/nyu2d07k/

Just to be on the same page, IMHO #12653 is not a bug - `point.state` is an option that should be an object, not a string. Similar error we can observe with other options, for example `zones`: https://jsfiddle.net/BlackLabel/ubrzLde0/

Regarding code: we already have nested-properties traversing, for example used in [`series.keys`](https://api.highcharts.com/highstock/series.line.keys) - can we reuse that logic?

bre1470:
It adds an single property for custom data, to prevent name conflicts between custom properties and official ones. Your example `point.myValue` will become invalid, when we are done with the TypeScript migration.

This is about issue #12847.

No, the logic in keys is a setter loop.

bre1470:
Thanks for your help! All issues should be solved as far as I understand the impact in the demos. Please re-review. :)

ts/parts/Point.ts
```diff
@@ -609,7 +612,9 @@ class Point {
         // For higher dimension series types. For instance, for ranges, point.y
         // is mapped to point.low.
         if (pointValKey) {
-            point.y = (point as any)[pointValKey];
+            point.y = pointValKey.indexOf('custom.') === 0 ?
+                getPropertyValue(pointValKey, point) :
+                (point as any)[pointValKey];
```
pawelfus: Can we move all repeated code into a separate method? Like this:
```js
getValueFor = function (point, key) {
  return key.indexOf('custom.') === 0 ? getPropertyValue(key, point) : (point as any)[key];
};
```

So the line 615 above would be simple `point.y = getValueFor(pointValKey, point);` Otherwise, if we ever will need to use that method again there's a high chance that `something.indexOf('custom.')` will be forgotten.

bre1470: The condition is more for performance reasons, to skip a function call. But there is another thing, why I make a difference. `getPropertyValue` should only work on `point.options` (so that is the mistake in the code snippet).

pawelfus: > The condition is more for performance reasons, to skip a function call

Yes, I understand that, but I still believe in this case it's better to make a function call instead of code repetition.

> `getPropertyValue` should only work on point.options (so that is the mistake in the code snippet)

The snippet was created from the lines above comment:

```js
            point.y = pointValKey.indexOf('custom.') === 0 ?
                getPropertyValue(pointValKey, point) :
                (point as any)[pointValKey];
```
I noticed that sometimes you use `point` otherwise `point.options`. I thought it's intentional (Highcharts.Point instance vs user provided point options).

bre1470:
I added tests and moved the inline conditions to the new `Point.getNestedProperty` function.

samples/unit-tests/series/datasorting/demo.js
```diff
@@ -95,3 +95,54 @@ QUnit.test('Data sorting ', function (assert) {
...
+                custom: {
+                    myValue: 'a'
+                },
+                y: 2
+            }],
+            dataSorting: {
+                enabled: true
+            }
+        }]
+    }, function (chart) {
```
pawelfus: Just a question: Any particular reason to do the test in callback?

bre1470: I do not think so. I guess, I used it just because of the similarity to the TestTemplate pattern.

Issue #12850, opened 2020-01-24
Enhancement: Inner size for pane


PaulDalek:
# Description of the feature
Implementing the possibility to set an inner size for pane.

PaulDalek:
Closed by #12852

PR #12852, opened 2020-01-24
enhancement/12850-pane-inner-size

PaulDalek:
Added inner size for pane (#12850).

Demos:
- https://jsfiddle.net/BlackLabel/jx6a0pry/ - polar bars
- https://jsfiddle.net/BlackLabel/cj4weny2/ - polar columns


ts/mixins/centered-series.ts
```diff
@@ -88,16 +88,34 @@ H.CenteredSeriesMixin = {
...
+            innerSize = parseFloat(innerSize);
+        }
+
+        positions = [
+            pick(centerOption[0] as any, '50%' as any),
+            pick(centerOption[1] as any, '50%' as any),
+            // Prevent from negative values
+            pick(size && size < 0 ? void 0 : options.size, '100%'),
+            pick(innerSize && innerSize < 0 ? void 0 : options.innerSize, '0%')
+        ];
```
pawelfus: Can we use `getCenter()` method from cetnered-series mixin?

ts/parts-more/Polar.ts
```diff
@@ -78,7 +82,7 @@ declare global {
                 high: number,
                 start: number,
                 end: number
-            ): SVGPathArray;
+            ): SVGPathArray|SVGAttributes;
```
pawelfus: I don't see related changes in code

ts/parts-more/Polar.ts
```diff
@@ -521,61 +540,93 @@ var polarAnimate = function (
...
-                        translateY: center[1] + plotTop,
-                        scaleX: 0.001, // #1499
-                        scaleY: 0.001
-                    };
-
-                    group.attr(attribs);
-                    if (markerGroup) {
-                        markerGroup.attr(attribs);
+                // A different animation needed for column like series
+                if (series instanceof seriesTypes.column) {
```
pawelfus: Use `series.is('column')` instead

pawelfus: And why in general we need to switch from animating scaling to animating radius?

ts/parts-more/Polar.ts
```diff
@@ -600,21 +651,48 @@ if (seriesTypes.column) {
...
-                start: start,
-                end: end,
-                innerR: len - pick(low, len)
+            len = this.yAxis.len,
+            paneInnerR = center[3] / 2,
+            r = len - high + paneInnerR,
+            innerR = len - pick(low, len) + paneInnerR;
+
+        // Render columnrange series the old way, enhance in the future
+        if (this instanceof seriesTypes.columnrange) {
```
pawelfus: Use `series.is('...')`

ts/parts-more/Polar.ts
```diff
@@ -792,23 +874,28 @@ if (seriesTypes.column) {
...
-                        )
-                    };
+
+                    // Changed the way polar columns are drawn in order to make
+                    // it more consistent with the drawing of inverted columns
+                    // (they are using the same function now). Also, it was
+                    // essential to make the animation work correctly (the
+                    // scaling of the group) is replaced by animating each
+                    // element separately.
+                    point.shapeArgs = series.polarArc(
```
pawelfus: `point.shapeArgs` should be an object with config for the shape, here it's SVGPathArray - why we need this inconsistency? When `point.shapeType = 'arc';` is set, then I expect `shapeArgs` be configured via `{ x, y, r, innerR, start, end }` options.

ts/parts-more/Polar.ts
```diff
@@ -792,23 +874,28 @@ if (seriesTypes.column) {
...
                 }
 
                 // Provided a correct coordinates for the tooltip
                 series.toXY(point);
 
                 if (chart.inverted) {
-                    tooltipPos = xAxis.postTranslate((point as any).rectPlotY,
-                        point.barX + point.pointWidth / 2);
+                    tooltipPos = yAxis.postTranslate((point as any).rectPlotY,
+                        barX + point.pointWidth / 2);
```
pawelfus: Is that the reason why tooltips is not rendered in demos?

ts/parts-more/RadialAxis.ts
```diff
@@ -746,17 +766,26 @@ radialAxisMixin = {
 
         // Spokes
         if (axis.isCircular) {
-            a = (typeof innerRadius === 'string') ?
-                relativeLength(innerRadius, 1) : (
-                    innerRadius /
-                    Math.sqrt(Math.pow(x2 - x1, 2) + Math.pow(y2 - y1, 2))
-                );
+            pyth =
```
pawelfus: Took me a while :D Wouldn't be `distance` more descriptive name?

ts/parts-more/RadialAxis.ts
```diff
@@ -776,16 +805,20 @@ radialAxisMixin = {
...
-            if (!options.isCrosshair &&
-                (value as any < 0 || value as any > axis.height) && inverted) {
-                value = 0;
+            if (value) {
+                if (value < 0) {
+                    value = 0;
+                } else if (value > height) {
+                    value = isCrosshair ? height :
+                        (axis.reversed ? height : 0);
+                }
```
pawelfus: Consider using `H.clamp(value, min, max)`

ts/parts-more/Pane.ts
```diff
@@ -313,6 +313,19 @@ extend(Pane.prototype, {
          */
         size: '85%',
 
+        /**
+         * The size of pane's inner area (area in the center that is not
+         * occupied), either as a number defining pixels, or a
+         * percentage defining a percentage of the plot area.
```
TorsteinHonsi: If I remember correctly, it is a percentage of the pane's _radius_, as defined in the `size` option. This should be communicated more precisely.

ts/parts-more/Polar.ts
```diff
@@ -1009,7 +1104,9 @@ H.SVGRenderer.prototype.clipCircle = function (
             id: id
         }).add(this.defs);
 
-    wrapper = this.circle(x, y, r).add(clipPath);
+    wrapper = innerR ?
+        this.arc(x, y, r, innerR, 0, 2 * Math.PI).add(clipPath) :
+        wrapper = this.circle(x, y, r).add(clipPath);
```
TorsteinHonsi: The `wrapper = ` part shouldn't be necessary in this ternary?

ts/parts/Axis.ts
```diff
@@ -7434,7 +7443,32 @@ extend(Axis.prototype, /** @lends Highcharts.Axis.prototype */{
...
+                    chart.renderer.clipCircle(
+                        clipOptions.x,
+                        clipOptions.y,
+                        clipOptions.r,
+                        clipOptions.innerR
+                    );
+
+                if (this.cross) {
+                    this.cross.clip(this.clipCrosshair);
+                }
```
TorsteinHonsi: This part should be refactored out so we don't pollute the basic `Axis` with polar-specific stuff. We could possibly do it in an event call, `afterDrawCrosshairs`, but I'm wondering if clipping is necessary at all. The crosshair path is calculated using `getPlotLinePath` further up, and I would expect this to be a correct path within the pane`s `size` and `innerSize` bounds.

ts/mixins/centered-series.ts
PaulDalek: This change is applied inside the `getCenter` function. I have added it in order to prevent from negative values for radiuses in general when generating positions by the `getCenter` function. Right now setting the `pane.size` as some negative results in error (https://jsfiddle.net/BlackLabel/fLxosc71/). What do you think about it? Should I change that to how it worked before?

ts/parts-more/Polar.ts
PaulDalek: This is added because the `polarArc` function can return not only path array but also attributes specific for arcs (`x`, `y`, `r`, `innerR`, `start`, `end`). The reason why I have changed it to work that way is explained in the comment below (question about the animation).

ts/parts-more/Polar.ts
PaulDalek: As for the first, thanks for the suggestion, I'll change it! As for the second question, the reason for this change is that the scaling is performed from the center no matter what the `innerSize` is set to. Before, when there was no `innerSize` option, it wasn't a problem but right now it just looks weird if there is some `innerSize` set for a pane and points are still animating from the center. So in order to make columns start animating from the places determined by the `innerSize` (`center[3]`) the `r` and `innerR` are attr'ed with pane's inner size and then animated to actual values. Is there a better way to do it?

ts/parts-more/Polar.ts
PaulDalek: Thanks! Will change that.

ts/parts-more/Polar.ts
PaulDalek: I have changed the `shapeType` (from the `path`) when it comes to columns (only) series (no matter if inverted or not). In the actual release (v8.0.0) the radial bar series elements have `shapeType` set to `arc` while non inverted polar columns' `shapeType` are set to `path`. Because of the scaling animation problem (explained above) I have changed the way how non-inverted columns are drawn to the same way as the radial bar series elements (both have the `shapeType` set to `arc` and this way the `polarArc` function must return the `shapeArgs` object in the following form -> `{x, y, r, innerR, start, end}` for non-inverted columns too) so the function is able to return objects with arguments for arcs. Right now the `polarArc` returns such object but it also can return the path array (for columnrange series for now but this also can be changed). That's why the `SVGPathArray|SVGAttributes` is set as a possible type of returned value from the `polarArc` (Pawel's second question above). What do you think about it? Should it be done some other way?

ts/parts-more/Polar.ts
PaulDalek: It is caused by this commit: https://github.com/highcharts/highcharts/commit/0b5d8744b7a781130afe3b521dc1da4fc8e946bf.

ts/parts-more/RadialAxis.ts
PaulDalek: Agree, the `pyth` is a little bit enigmatic :D.

ts/parts-more/RadialAxis.ts
PaulDalek: Thanks!

ts/parts-more/Pane.ts
PaulDalek: I thought that it is a similar way as `size`. The `size` descriptions says: _'The size of the pane, either as a number defining pixels, or a percentage defining a percentage of the plot are.'_ but I will change that. Thanks!

ts/parts-more/Polar.ts
PaulDalek: Yes, sorry missed that. Thank you!

ts/parts/Axis.ts
PaulDalek: Yes, this could probably be done inside the `getPlotLinePath` (in the section where `a` and `b` properties are calculated to find an adequate ratio). I'll check that and correct it. Thanks!

ts/parts-more/Pane.ts
TorsteinHonsi: Yes, actually the size is a percentage of the smallest of the plot height or plot width. So a `size` og 100% fills the smallest dimension of the plot area.

ts/parts-more/Polar.ts
pawelfus: Thanks 👍

ts/parts-more/Polar.ts
pawelfus: Ah, thanks! Created an example with comparison: https://jsfiddle.net/BlackLabel/hp1nL5d0/ (the first one does animate with extra offset, not available right now). But this won't work with wider arcs: https://jsfiddle.net/BlackLabel/Lcurjomh/

I don't know any simple solution here, how about you @TorsteinHonsi or @Izothep ?

ts/mixins/centered-series.ts
pawelfus: > This change is applied inside the getCenter function.

Oh right, my bad!

ts/parts-more/Polar.ts
pawelfus: Can `polarArc` simply return either `{x, y, r, innerR, start, end}` (arc) or `{ d }`(path) ? And can we define different methods in proto for column and overwrite for other series (apart from columnrange - do we have other series?).

TorsteinHonsi: I agree this dualism is unfortunate. I think the only sensible thing to do is that it returns `SVGAttributes` always. But we should seek consistency wherever possible. If we can use `shapeType = 'arc'` all over, it would probably be the right combination of consistency and flexibility.

ts/parts-more/Polar.ts
PaulDalek: Right now the `polarArc` returns only `SVGAttributes` object.

ts/parts-more/Polar.ts
PaulDalek: Added the use of the `is` function.

ts/parts-more/Polar.ts
PaulDalek: Corrected.

ts/parts-more/Polar.ts
PaulDalek: Changing the `shapeType` from `path` to `arc` for points of columnrange series allowed to get rid of the necessity of generating the path array (`SVGPathArray` type) with the `symbols.arc` (the old way, inside the `polarArc`) and instead, the `polarArc` function returns `SVGAttributes` object (only) for columnrange too (just as for column series).

ts/parts-more/RadialAxis.ts
PaulDalek: Changed to the `distance`.

ts/parts-more/RadialAxis.ts
PaulDalek: After some thinking I came to the conclusion that the `if else` statement here is not necessary (no matter if `value < 0` or `value > height`, it always must be 0 then).

ts/parts-more/Pane.ts
PaulDalek: I have corrected the description for both, `size` and `innerSize`. In case when the `innerSize` is percentage, it is percentage of the `size`.

ts/parts-more/Polar.ts
PaulDalek: Corrected.

ts/parts/Axis.ts
PaulDalek: It seems that I have already provided such logic for grid lines and somehow I missed that it also works for circular `crosshair` therefore clipping (which I have added before) was unnecessary.

ts/parts-more/RadialAxis.ts
```diff
@@ -340,18 +341,28 @@ radialAxisMixin = {
...
+        radius?: number,
+        innerRadius?: number
     ): Highcharts.RadialAxisPath {
         var center = this.center,
             end,
             chart = this.chart,
             r = pick(radius, center[2] / 2 - this.offset),
             path: Highcharts.RadialAxisPath;
 
+        if (innerRadius === void 0) {
```
pawelfus: Just nitpicking: everywhere else we check `typeof variable === 'undefined'`

ts/parts/Axis.ts
```diff
@@ -319,6 +322,7 @@ declare global {
             public height: number;
             public horiz?: boolean;
             public isBroken?: boolean;
+            public isCircular?: (undefined|boolean);
```
pawelfus: I think this should be defined in RadialAxis, not in Axis? The same for `clipCrosshair `. Perhaps I'm wrong, @bre1470 ?

ts/parts/Axis.ts
```diff
@@ -7401,7 +7406,7 @@ extend(Axis.prototype, /** @lends Highcharts.Axis.prototype */{
 
             // Draw the cross
             if (!graphic) {
-                this.cross = graphic = this.chart.renderer
+                this.cross = graphic = chart.renderer
```
pawelfus: Nitpicking again: since you change `this` -> `chart`, then `this.cross` should be changed too? ;)

ts/parts-more/Polar.ts
pawelfus: Shouldn't `SVGPathArray` be removed from the above then?

ts/parts-more/Polar.ts
```diff
@@ -601,21 +652,33 @@ if (seriesTypes.column) {
         high: number,
         start: number,
         end: number
-    ): Highcharts.SVGPathArray {
+    ): Highcharts.SVGPathArray|Highcharts.SVGAttributes {
```
pawelfus: `Highcharts.SVGPathArray` should be removed?

ts/parts-more/Polar.ts
PaulDalek: Ahh yes, I have changed the function itself but forgot about getting rid of the return type. Corrected, thanks!

ts/parts-more/RadialAxis.ts
PaulDalek: Corrected, thanks!

ts/parts/Axis.ts
PaulDalek: Got rid of the `clipCrosshair` as it isn't necessary any more (crosshair's length is calculated in the `getPlotLinePath` function) and as it seems the `isCircular`  it was already defined in the RadialAxis part.

ts/parts/Axis.ts
PaulDalek: Here `this` is referring to the Axis. Do you mean creating `axis` variable for `this`?

ts/parts-more/Polar.ts
PaulDalek: Removed, thanks!

ts/parts/Axis.ts
pawelfus: Ah, my bad, thanks!

PR #12855, opened 2020-01-27
bugfix/9682-scroll-page-while-touching-legend

### LINKED ISSUE #9682: Not possible to scroll site on mobile, when touching legend items.
#### Expected behaviour

Site should be able to scroll when touching legend item, and legend item should be disabled only after touch end (as it is in the case of handling standard mouse events).

#### Actual behaviour

When touching the legend on mobile devices, and want to scroll down/up a site, it's `legendItemClick` default event called, just on start touch.

[Official basic line demo](https://www.highcharts.com/demo/line-basic)
Try to swipe the screen on mobile, touching any legend item.

#### Affected browser(s)
Mobile Chrome v71.0.3578.98

kamilkulig:
Fixed #9682, page couldn't be scrolled while touching a legend item.

### DIFF
--- a/ts/parts/SvgRenderer.ts
+++ b/ts/parts/SvgRenderer.ts
@@ -1923,7 +1923,7 @@ extend((
         if (hasTouch && eventType === 'click') {
             element.ontouchstart = function (e: Event): void {
                 svgElement.touchEventFired = Date.now(); // #2269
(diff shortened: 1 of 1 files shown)


Izothep:
I believe we can simply close this PR as we have merged another solution from different PR.

PR #12860, opened 2020-01-28
series.flags.marker shouldn't be visible in the API

kamilkulig:
It is already disabled in series.flags.data


ts/parts/FlagsSeries.ts
```diff
@@ -143,7 +143,7 @@ seriesType<Highcharts.FlagsSeries>(
      * @extends      plotOptions.column
      * @excluding    animation, borderColor, borderRadius, borderWidth,
      *               colorByPoint, dataGrouping, pointPadding, pointWidth,
-     *               turboThreshold
+     *               turboThreshold, marker
```
KacperMadej: It's cleaner to keep the alphabetical order of the excluded options. @kamilkulig

PR #12862, opened 2020-01-28
Make tooltip.stickOnHover default with a11y module.

oysteinmoseng:
Make the new `tooltip.stickOnHover` option on by default with the a11y module loaded.

### DIFF
--- a/js/modules/accessibility/options/options.js
+++ b/js/modules/accessibility/options/options.js
@@ -734,6 +734,12 @@ var options = {
              */
             enabled: true
         }
+    },
+    /**
+     * @optionparent tooltip
+     */
+    tooltip: {
+        stickOnHover: true
     }
 };
 export default options;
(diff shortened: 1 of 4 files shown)


pawelfus:
Should we improve docs for `tooltip.stickOnHover`? So docs actually informs users that default value has changed.

PS: It seems that current value (`undefined`?) conflicts with type `boolean` in jsDoc :)

oysteinmoseng:
Yes, good catch! I will add a note in the doclet. According to Sophie, we don't need to mention `undefined` types in option doclets, as this applies to a lot of options that are undefined by default.

pawelfus:
> According to Sophie, we don't need to mention undefined types in option doclets, as this applies to a lot of options that are undefined by default.

Sorry for not being clear! I don't want to mention `undefined` in docs, but to set `@defaults false`. For example [`tooltip.useHTML`](https://api.highcharts.com/highstock/tooltip.useHTML).

But! I think it's unnecessary: in this particular case `undefined` is fine - actually `undefined` means to be `true` or `false` value, depending on other modules to be loaded (or not). For example [`tooltip.outside`](https://api.highcharts.com/highstock/tooltip.outside) or [`scrollbar.liveRedraw`](https://api.highcharts.com/highstock/scrollbar.liveRedraw).

Issue #12584, opened 2019-12-06
Annotations: Add defer options similar to dataLabels


pawelfus:
#### Expected behaviour
It should be possible for annotations to defer rendering until initial animation for series has finished.

#### Actual behaviour
Annotations show up right away.

Basts:
Closed as #13688

PR #12867, opened 2020-01-28
enhancement/12584-annotations-defer-options

Basts:
Enhancement #12584, defer options for annotations.


bre1470:
You have to introduce the `AnnotationDeferOptionsObject` twice - one time as the internal interface in TypeScript syntax, the second time in doclet syntax for the external TypeScript declarations.

ts/annotations/annotations.src.ts
```diff
@@ -1198,13 +1219,29 @@ merge(
...
+                // else inherits from animation set in plotOptions
+                this.options.defer = isObject(this.options.defer) ? this.options.defer : (
+                    plotOptions.series && defined(plotOptions.series.animation) ?
+                        plotOptions.series.animation :
+                        plotOptions.line.animation
+                );
+                if (this.chart.renderer.forExport) {
+                    this.options.defer = false;
+                }
+            }
```
pawelfus: Please, move this to `annotation.init` method. Then store this logic in `this.defer` property so we can easily access this later (e.g. for exporting) instead of modifying `this.options.defer`.

pawelfus: @bre1470 - can we define `this.defer` as `false|AnimationOptionsObject`? So later it can be simply written as `if (this.defer) { ..animate(defer) }` ? Any cons?

samples/highcharts/annotations/defer/demo.details
```diff
@@ -0,0 +1,6 @@
+---
+ name: Highcharts Anootations Demo
```
pawelfus: Typo "Anoo"

ts/annotations/annotations.src.ts
```diff
@@ -1608,6 +1645,27 @@ extend(chartProto, /** @lends Highcharts.Chart# */ {
...
+    },
+
+    delayRendering: function (this: Highcharts.AnnotationChart): void {
+        var chart = this;
+
+        chart.annotations.forEach(function (annotation): void {
+            var defer = annotation.options.defer,
+                seriesAnimDuration = animObject(defer).duration,
+                fadeInDuration = seriesAnimDuration ? Math.min(seriesAnimDuration, 200) : 0,
+                annotationsGroup = [annotation.shapesGroup, annotation.labelsGroup];
```
pawelfus: We should simply use `annotation.graphic` which is the container for  `shapesGroup` and `labelsGroup`

samples/highcharts/annotations/defer/demo.js
```diff
@@ -0,0 +1,41 @@
...
+        data: [{
+            y: 29.9,
+            id: 'min'
+        }, 71.5, 106.4, 129.2, 144.0, 176.0, 135.6, 148.5, {
+            y: 216.4,
+            id: 'max'
+        }, 194.1, 95.6, 54.4]
+    }],
+
+    annotations: [{
```
pawelfus: Could you split this annotation into three separate ones (label with max, label with min, just a shape) - and set different options (`defer=true`, `defer=false`, `defer={duration}`)?

Setting title for this demo would be nice too :)

ts/annotations/annotations.src.ts
```diff
@@ -497,6 +505,19 @@ merge(
              */
             visible: true,
 
+            /**
+             * Whether to defer displaying the annotations until the set
+             * duration time has finished. Setting to `'false'` disable
```
pawelfus: - Shouldn't be `'false'` -> `false` ? The same below with `true`.
- disable -> disables
- perhaps simply "Setting to `false` renders annotations immediately." ?

ts/annotations/annotations.src.ts
bre1470: Yes, that would improve the situation with the typing, but then it needs to be changed in the options as well (doclets & interfaces).

Basts: @bre1470 Should the mentioned option `this.defer` be introduce in doclets syntax too? It is an internal property which is defined by `options.defer` (which user is able to control) and other conditions.

bre1470: No, I do not think so. This was a comment about the type used in API options doclets, but as I see `true` is there in use to activate defaults. So nothing has to be changed. Sorry for the confusion.

ts/annotations/navigationBindings.ts
```diff
@@ -1623,6 +1623,19 @@ H.setOptions({
...
+             * Whether to defer displaying the annotations until the set
+             * duration time has finished. Setting to `false` renders
+             * annotation immediately. If set to `true` inherits the duration
+             * time set in [plotOptions.series.animation](#plotOptions.series.animation).
+             *
+             * @sample highcharts/annotations/defer
+             *         Set defer duration time
+             *
+             * @type {boolean|Highcharts.AnnotationDeferOptionsObject}
+             */
```
pawelfus: Do we need this definition again? `annotationsOptions` already extends `annotations`, so `defer`'s definition should show up anyway. Probably simple `defer: false` should be enough.

ts/annotations/annotations.src.ts
```diff
@@ -1023,12 +1055,28 @@ merge(
...
             this.addShapes();
             this.addLabels();
             this.addClipPaths();
             this.setLabelCollector();
+
+            if (defer) {
+                // If defer duration is set use set value,
+                // else inherits from animation set in plotOptions
+                this.defer = isObject(this.options.defer) ?
+                    this.options.defer :
```
pawelfus: Perhaps I miss something here but:
- if `this.options.defer` is object -> okay, use that options
- if `this.options.defer` is true -> okay, inherit defer/animation from plotOptions
- if `this.options.defer` is false -> we still inherit from plotOptions ?

pawelfus: Bonus point: `this.options.defer` is already defined above in line 1059.

ts/annotations/annotations.src.ts
```diff
@@ -497,6 +516,19 @@ merge(
...
+             * Whether to defer displaying the annotations until the set
+             * duration time has finished. Setting to `false` renders
+             * annotation immediately. If set to `true` inherits the duration
+             * time set in [plotOptions.series.animation](#plotOptions.series.animation).
+             *
+             * @sample highcharts/annotations/defer
+             *         Set defer duration time
+             *
+             * @type {boolean|Highcharts.AnnotationDeferOptionsObject}
+             */
```
pawelfus: Missing `@since 8.0.1` tag.

TorsteinHonsi:
In the current implementation, the annotations are deferred, but when the time is in, they jump suddenly on to the scene. Compare that to how data labels and series labels behave. They are deferred, but then fade in with a smooth animation: https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/demo/line-basic/.

Actually, since this is a repeating pattern, I would like to discuss a different approach. What if we make a `defer` option directly on the animation parameters and implement it as low down in the animation engine as we can?

This would allow us to remove the `defer` implementation from series labels and annotations and probably from data labels.

A defer-and-fade-in sequence would then look like this:
```js
element.attr({ opacity: 0 })
    .animate({
        defer: 1000,
        duration: 150,
        opacity: 1
    });
```
What do you think?

Basts:
@TorsteinHonsi 

It sounds like a good idea. I am going to open a new ticket with this enhancement and I will back to trying to implement it in nearly future when I finish my current tasks.

Should this PR be closed?

Basts:
New ticket: https://github.com/highcharts/highcharts/issues/12901

TorsteinHonsi:
Closing this in favor of #12901.

PR #12873, opened 2020-01-29
bugfix/12566-data-module-null-point

### LINKED ISSUE #12566: Data module parses absent points to undefined instead of null.
#### Expected behaviour
I expect to get `null` point when I provide data in a table without a corresponding `<td></td>` tag.

#### Actual behaviour
I get `undefined` point.

https://jsfiddle.net/BlackLabel/nbjurf5y

#### Product version
Highcharts 7.2.1

mateuszkornecki:
Fixed #12566, in the data module, empty points should be parsed to `null` instead of `undefined`.


samples/unit-tests/data/general/demo.js
```diff
@@ -193,4 +193,58 @@ QUnit.test('Data config on updates', function (assert) {
...
+            table: 'secondTable',
+            decimalPoint: ','
+        }
+    });
+
+    assert.strictEqual(
+        chart.data.columns[2][0],
+        null,
+        'Empty point should be parsed to null instead of undefined.'
+    );
```
TorsteinHonsi: After the assertion, clean up. Remove the table from the page.

PR #12878, opened 2020-01-31
Added patternTransform option for pattern fills. Closes #12763.

### LINKED ISSUE #12763: Scale not working on Patterns
Looking at the SVGAttributes I thought the properties scaleX/scaleY/rotation would work in a pattern:path since a path receives a SVGAttribute object. For some reason they are not working. Am I doing this correctly?
https://jsfiddle.net/panintelligence/ojcm5grd/4/

https://api.highcharts.com/class-reference/Highcharts.PatternOptionsObject#path
https://api.highcharts.com/class-reference/Highcharts.SVGAttributes

oysteinmoseng:
Added new option `pattern.patternTransform` that allows SVG `patternTransform` functions on a pattern.
___
Debatable if naming should be `pattern.transform` instead, but leaning towards `pattern.patternTransform` since it is a one-to-one with the SVG attribute.

### DIFF
--- a/js/modules/pattern-fill.src.js
+++ b/js/modules/pattern-fill.src.js
@@ -58,7 +58,7 @@ import H from '../parts/Globals.js';
 * @name Highcharts.PatternOptionsObject#path
 * @type {string|Highcharts.SVGAttributes}
 */ /**
-* SVG patternTransform to apply to the entire pattern.
+* SVG `patternTransform` to apply to the entire pattern.
 * @name Highcharts.PatternOptionsObject#patternTransform
 * @type {string}
 * @see [patternTransform demo](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/series/pattern-fill-transform)
--- a/js/modules/pattern-fill.src.js
+++ b/js/modules/pattern-fill.src.js
@@ -58,6 +58,11 @@ import H from '../parts/Globals.js';
 * @name Highcharts.PatternOptionsObject#path
 * @type {string|Highcharts.SVGAttributes}
 */ /**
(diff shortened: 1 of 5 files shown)


js/modules/pattern-fill.src.js
```diff
@@ -58,6 +58,11 @@ import H from '../parts/Globals.js';
 * @name Highcharts.PatternOptionsObject#path
 * @type {string|Highcharts.SVGAttributes}
 */ /**
+* SVG patternTransform to apply to the entire pattern.
+* @name Highcharts.PatternOptionsObject#patternTransform
+* @type {string}
+* @see [patternTransform demo](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/series/pattern-fill-transform)
```
pawelfus: I'm just curious: why `@see`, not `@sample`?

oysteinmoseng: `@sample` is a custom tag that unfortunately doesn't work for interface doclets..

pawelfus: Ah, thanks! 👍

ts/modules/pattern-fill.src.ts
```diff
@@ -111,6 +112,11 @@ declare global {
  * @name Highcharts.PatternOptionsObject#path
  * @type {string|Highcharts.SVGAttributes}
  *//**
+ * SVG patternTransform to apply to the entire pattern.
```
TorsteinHonsi: Code should be written in backticks: `patternTransform`.

TorsteinHonsi:
... otherwise approved :+1:

PR #12879, opened 2020-01-31
tools/utilities-error

marvin19:
Tools: Migrated Utilities.error to ES6 pattern.
This PR is based on https://github.com/highcharts/highcharts/pull/12877


bre1470:
Conflict-free :)

PR #12891, opened 2020-02-04
bugfix/11993-wordcloud-setdata

### LINKED ISSUE #11993: Incorrect wordcloud animation on unsorted data
This report is about wordclouds where the data is _not_ sorted by weight.

## Expected behaviour
1 Words move from the old to the new position
2 Correct color on words after animation

#### Actual behaviour
1 Words start their animation from an seemingly arbitrary position and then move to their new position
2 The words get new colors assigned. The color of the first word in the data no longer corresponds with the first color in the plotoptions colors.

https://jsfiddle.net/kzoon/tmowrf9j/
Click 'update chart'. Clicking 'update chart' repeatedly makes the issue even clearer, because the data is exactly the same...

https://jsfiddle.net/kzoon/akyn5u9w/, which uses sorted data, shows the correct behavior.

kamilkulig:
Fixed #11993, wordcloud looked incorrect after running `setData()`.

### DIFF
--- a/ts/modules/wordcloud.src.ts
+++ b/ts/modules/wordcloud.src.ts
@@ -980,13 +980,20 @@ var wordCloudSeries: Partial<Highcharts.WordcloudSeries> = {
                 return p.weight;
             }),
             maxWeight = Math.max.apply(null, weights),
-            data = series.points.sort(function (
+            // concat() prevents from sorting the original array.
+            data = series.points.concat().sort(function (
                 a: Highcharts.WordcloudPoint,
                 b: Highcharts.WordcloudPoint
             ): number {
                 return b.weight - a.weight; // Sort descending
             }),
             field: Highcharts.WordcloudFieldObject;
+
+        // Reset the scale before finding the dimensions (#11993)
+        series.group.attr({
+            scaleX: 1,
+            scaleY: 1
+        });
 
         // Get the dimensions for each word.
         // Used in calculating the playing field.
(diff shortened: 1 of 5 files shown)


ts/modules/wordcloud.src.ts
```diff
@@ -980,14 +980,21 @@ var wordCloudSeries: Partial<Highcharts.WordcloudSeries> = {
...
+            // concat() prevents from sorting the original array.
+            data = series.points.concat().sort(function (
                 a: Highcharts.WordcloudPoint,
                 b: Highcharts.WordcloudPoint
             ): number {
                 return b.weight - a.weight; // Sort descending
             }),
             field: Highcharts.WordcloudFieldObject;
 
+        // Reset the scale before finding the dimensions (#11993)
```
pawelfus: A comment why we need to reset the scale would be nice. It's quite unusual reason AFAIR ;)

kamilkulig:
The visual change is not deliberate. However, the reference image doesn't seem like the intended look in the first place. The position of tooltip isn't something that we can recreate in the demo just by moving the mouse:
 
image

Also, the words overlap in both images which leads me to suspicion that screenshot has been made during the initial animation of the series.

TorsteinHonsi:
> The visual change is not deliberate.

The test tool opens the first point's tooltip by `point.onMouseOver()`. Maybe the change in the sorting order explains why a different point is hovered now? Please check that, then come back with an advice on whether the visual change should be approved or not.

kamilkulig:
Now I see. I added `concat()` in my commit to preserve the original order in `series.points` array.  `series.setData()` doesn't work correctly without it.
After my change the tooltip displays for word `Lorem` - the first data point.  So, I corrected the demo and the visual change should be approved.

kamilkulig:
Now I see. I added `.concat()` in my commit to preserve the original order of `series.points` array. `.setData()` doesn't work correctly without it.
After my change, the tooltip displays for the word "Lorem" - the first point. So my commit corrects the demo and the visual change should be approved.

PR #12892, opened 2020-02-04
bugfix/12885-sticky-follow-pointer

### LINKED ISSUE #12885: Tooltip stickOnHover and followPointer
#### Expected behaviour
Expected the `followPointer` option to work even when `stickOnHover` is true.

#### Actual behaviour
When both options are set to true, neither of them work. 

https://jsfiddle.net/highcharts/eLag1cw9/3/

1. Drag the mouse over a slice and observe that the tooltip no longer follows the pointer.
2. Hover over "Internet Explorer" so that the tooltip partly covers IE, partly Firefox. Now move the mouse up over the tooltip into Firefox. The tooltip will now move to Firefox, thereby failing to comply with the `stickOnHover` option.

#### Product version
Pre-production, post v8.0.0.

bre1470:
Fixed #12885, no stickOnHover when followPointer.


bre1470:
Please review once more.

ts/parts/Tooltip.ts
```diff
@@ -1134,6 +1147,45 @@ class Tooltip {
...
+     * @private
+     * @function Highcharts.Tooltip#isStickOnContact
+     *
+     * @param {Highcharts.PointerEventObject} e
+     * Pointer event to test.
+     *
+     * @return {boolean}
+     * True, if pointer is in contact with sticky tooltip.
+     */
+    public isStickyOnContact(e: Highcharts.PointerEventObject): boolean {
```
TorsteinHonsi: Now that we have refactored the `stickOnContact` logic to work with real elements and mouseover/out events, do we need this computational hit testing at all? We have the following situations:

* Mouse from point to tooltip tracker => Tooltip sticks. Events detect it.
* Mouse from point to anything else => Tooltip hides. Events detect it.
* Mouse from tooltip tracker to own points and series (tooltip is aware of its point) => Tooltip sticks. Events detect it.
* Mouse from tooltip to anything else => Tooltip hides because related target is not its own point or series.

I can't see a situation where we need hit testing?

bre1470: I need this in sankey at the moment, but maybe there is an other way to solve the issue with the link hover handling.

bre1470: I partially fixed it, but it still jumps sometimes into the center link for unknown reason.

ts/parts/Tooltip.ts
```diff
@@ -1689,6 +1729,52 @@ class Tooltip {
         }
     }
 
+    /**
+     * If the `stickOnContact` option is active, this will add a tracker shape.
+     *
+     * @private
+     * @function Highcharts.Tooltip#stickOnContact
+     */
+    private stickOnContact(): void {
```
TorsteinHonsi: This function name isn't really descriptive anymore. I think a better name would be simply `drawTracker`, as we now have a tracker that is equivalent to the series tracker.

ts/parts/Tooltip.ts
```diff
@@ -1689,6 +1729,52 @@ class Tooltip {
...
+        const labelBBox = label.getBBox();
+
+        // Combine anchor and tooltip
+        const size: Highcharts.SizeObject = {
+            width: Math.max((anchorPos[0] + chart.plotLeft - label.translateX), labelBBox.width),
+            height: Math.max((anchorPos[1] + chart.plotTop - label.translateY), labelBBox.height)
+        };
+
+        if (!tooltip.stickOnContactTracker) {
+            tooltip.stickOnContactTracker = label.renderer
```
TorsteinHonsi: Also equivalent to series, I think we should simply call it `tooltip.tracker`.

ts/parts/Tooltip.ts
```diff
@@ -1689,6 +1729,52 @@ class Tooltip {
...
+                .rect(0, 0, size.height, size.width)
+                .attr({ fill: 'rgba(0,0,0,0)' })
+                .add(label);
+        } else {
+            tooltip.stickOnContactTracker.attr('width', size.width.toString());
```
TorsteinHonsi: Use an object to set multiple attributes. And since the `size` is already on the right format, this is enough:
`tooltip.stickOnContactTracker.attr(size);`

ts/parts/Tooltip.ts
```diff
@@ -1689,6 +1729,52 @@ class Tooltip {
...
+        // Combine anchor and tooltip
+        const size: Highcharts.SizeObject = {
+            width: Math.max((anchorPos[0] + chart.plotLeft - label.translateX), labelBBox.width),
+            height: Math.max((anchorPos[1] + chart.plotTop - label.translateY), labelBBox.height)
+        };
+
+        if (!tooltip.stickOnContactTracker) {
+            tooltip.stickOnContactTracker = label.renderer
+                .rect(0, 0, size.height, size.width)
+                .attr({ fill: 'rgba(0,0,0,0)' })
```
TorsteinHonsi: We need to consider styled mode:
1. Add a class name, presumably `.highcharts-tracker`, and add the fill to highcharts.scss.
2. Make the `.attr` conditional, checking if `chart.styledMode`.
See similar code in the Series `drawTracker` function.

bre1470:
1. Pie chart defaults working now accordingly: https://jsfiddle.net/cbk2usdp/
2. Tracker size is now correct calculated: https://jsfiddle.net/3ro84neh/

TorsteinHonsi:
Great work!
Awaiting tests, otherwise ready to merge 👍

bre1470:
Thank you! :)

PR #12893, opened 2020-02-04
docs/area-threshold-null

### LINKED ISSUE #12824: plotOptions.area.threshold Typescript type does not accept null
#### Expected behaviour

Typescript linter will allow `plotOptions.area.threshold` to be `null`.

#### Actual behaviour

The Typescript linter throws the error below when setting `plotOptions.area.threshold` to `null`. According to the documentation this should be allowed. 

https://api.highcharts.com/highcharts/plotOptions.area.threshold

image

image

_None_

#### Product version
8.0.0

#### Affected browser(s)
Issue appears in the VSCode IDE.

bre1470:
Fixed #12824, Added null type to series.area.threshold option.

### DIFF
--- a/js/parts/AreaSeries.js
+++ b/js/parts/AreaSeries.js
@@ -149,6 +149,7 @@ seriesType('area', 'line',
      * @sample {highcharts} highcharts/plotoptions/area-threshold-infinity/
      *         A threshold of Infinity
      *
+     * @type    {number|null}
      * @since   2.0
      * @product highcharts highstock
      */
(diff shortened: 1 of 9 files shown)


js/parts/AreaSeries.js
```diff
@@ -149,6 +149,7 @@ seriesType('area', 'line',
      * @sample {highcharts} highcharts/plotoptions/area-threshold-infinity/
      *         A threshold of Infinity
      *
+     * @type    {number|null}
```
TorsteinHonsi: It is not specific to area series and its children, it also applies to [column series](https://api.highcharts.com/highcharts/series.column.threshold).

In `line` series however, it doesn't seem to have an effect, because the Y axis doesn't react to any `threshold` setting. So for line series it seems that `number` is the only type that makes sense. Although for consistency we could allow `null` there too, it will affect the `negativeColor` the same way as setting it to 0.

Issue #10663, opened 2019-05-06
Improve a set visibility in venn during interactivity


jon-a-nygaard:
#### Expected behaviour
When hovering a set in a venn diagram, it should distinguish itself clearly from the other sets.

Some possible behaviours that may achieve this is:
- When hovering a shape, the shape gets the highest zIndex which pushes it on top of the other shapes. This will make it very clear which point/shape is hovered, and how large it is. Downside is that it will hide the intersection shapes underneath itself during hover. Similar functionality exists in treemap today.
- When hovering a shape then it would use hover effects on all the other shapes it includes. E.g. a chart with sets A and B, when hovering A it would apply effects to the set A and the intersection A&B. This should help distinguish how large set A is, while still keeping the intersections visible. Downside is that it may still be visually difficult to see what shape is actually hovered, and it may be a more complicated and performance heavy solution.
- When hovering a point, the other points get an inactive state which fades them out. This functionality already exists, and may be the more natural choice, but at the moment it is not tested how well it works with the venn series.

#### Actual behaviour
The point is hovered with effects and its tooltip is displayed, but if the point is overlapped by many other points it will still be difficult to view.

See the following example where it can be difficult to see the actual size of the sets A, B, C and D due to their large overlap https://codesandbox.io/s/74vny3ry7q

pawelfus:
ad 3) This is how Venn could work with inactive state + zIndexing: https://codesandbox.io/s/busy-davinci-cmnjn

Config required to achieve that: 
```js
          point: {
            events: {
              mouseOver: function() {
                this.graphic && this.graphic.toFront();
              },
              mouseOut: function() {
                this.graphic &&
                  this.graphic.attr({
                    zIndex: this.sets && this.sets.length ? this.sets.length : 0
                  });
              }
            }
          },
          inactiveOtherPoints: true,
```

Basts:
@TorsteinHonsi

Might this issue by closed by this merge: https://github.com/highcharts/highcharts/pull/12898?

pawelfus:
Yes, we can close it now, thanks for pointing this out!

PR #12898, opened 2020-02-05
enhancement/#10663-venn-inactive-visiblity

Basts:
Changed the default `inactive.opacity` for venn diagrams to ease perception, see #10663.

### DIFF
--- a/js/modules/venn.src.js
+++ b/js/modules/venn.src.js
@@ -726,6 +726,9 @@ var vennOptions = {
             color: '${palette.neutralColor20}',
             borderColor: '${palette.neutralColor100}',
             animation: false
+        },
+        inactive: {
+            opacity: 0.075
         }
     },
     tooltip: {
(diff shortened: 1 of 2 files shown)


Basts:
After digging in the #12898 and testing possibilities in the venn chart I decided that the easiest and most qualities solution is to change the default **inactive opacity** to a much lower value. Setting the opacity to 0.075 (which is a value which I suggest to use) improves much point visibility while the point is hovered with keeping their intersection shapes. 

Suggested opacity:
ezgif com-video-to-gif (3)

The previous one (pay attention to point while the **C** or **AB** labels are hovered, their visibility is weak): 
ezgif com-video-to-gif (4)

Demo to test: https://jsfiddle.net/BlackLabel/twc3gabv/

Any suggestion or opinion is welcome.

@TorsteinHonsi 
I have a question how to change the PR message to use it in the changelog if this PR will be fine?
What should be included in the message?

Basts:
@TorsteinHonsi 

I tried to dig in it, but I am not sure exactly what to look for. I compared the venn chart animation with [pie](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/demo/pie-semi-circle/), bar and some other series and animation work the same for both charts.

The only difference with I noticed is with the [heat map  ](https://www.highcharts.com/demo/heatmap) where animation works as you described - smoothly between states for points.

Do you suggest that the same behaviour should be implemented in the Venn chart and other series? Maybe a new ticket to it will be a good approach.

PR #12917, opened 2020-02-10
Feature/annotations a11y

oysteinmoseng:
Added accessibility functionality for annotations. Also added new option `accessibility.point.valueDescriptionFormat`.
___
Annotations are now accessible to screen readers and other assistive technology relying on the AOM.

A list of the chart's annotations is added to the hidden info region before the chart. In addition, points with annotations connected to them will include a description of these annotations in their aria-labels by default.

New i18n options have been added under `langOptions.accessibility.screenReaderSection.annotations`, as well as `langOptions.accessibility.series.pointAnnotationsDescription`.

Because of the way annotation descriptions are computed, we have refactored out the logic that deals with describing the axis values of the data points. This allowed for the addition of the `accessibility.point.valueDescriptionFormat` option, which makes it a lot simpler to define custom descriptions in a declarative way, rather than having to write a description formatter function. Demos using description formatter functions have been updated to use this new option where possible.

No changes have been made to the annotations module.


oysteinmoseng:
Some tooling issues, but will look at this tomorrow and see if we can merge in master without too much drama.

oysteinmoseng:
Gave up on tooling issues, created new branch from master and pulled this one in there. Use this instead: https://github.com/highcharts/highcharts/pull/12927

PR #12942, opened 2020-02-12
docs/11300-missing-quotes-api

### LINKED ISSUE #11300: Misleading description of gapUnit option
#### Expected behaviour
The words `relative` and `value` should be surrounded by quotes (to indicate that these are strings).

#### Actual behaviour
There're no quotes.

The problem was reported to our support team:
> the fact is that your documentation is written without quotes (when the gapUnit is value) and I understood it as (when the gapUnit is [numeric] value).
> Pls fix it in documentation, eg: when the gapUnit is "value".

https://api.highcharts.com/highstock/series.line.gapUnit

#### Product version
v7.1.2-modified

raf18seb:
Fixed #11300, properties string values in API should be inside string quotes.

### DIFF
--- a/js/modules/broken-axis.src.js
+++ b/js/modules/broken-axis.src.js
@@ -385,9 +385,9 @@ H.Series.prototype.gappedPath = function () {
      * that if the distance between two points is greater than five times
      * that of the two closest points, the graph will be broken.
      *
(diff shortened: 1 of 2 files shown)


pawelfus:
I think this shouldn't show up (docs-fix) in the changelog.

PR #12944, opened 2020-02-12
tools/utilities-seriesType

marvin19:
Tools: Migrated Utilities.seriesType to ES6 pattern

### DIFF
--- a/ts/modules/tilemap.src.ts
+++ b/ts/modules/tilemap.src.ts
@@ -101,8 +101,18 @@ const {
 
 import '../parts-map/HeatmapSeries.js';
 
-// Utility func to get padding definition from tile size division
-var tilePaddingFromTileSize = function (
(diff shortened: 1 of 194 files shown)


ts/modules/tilemap.src.ts
```diff
@@ -95,25 +95,25 @@ const {
...
-        yDiv: number
-    ): Highcharts.TilemapPaddingObject {
-        var options = series.options;
-
-        return {
-            xPad: (options.colsize || 1) / -xDiv,
-            yPad: (options.rowsize || 1) / -yDiv
-        };
+// Utility func to get padding definition from tile size division
+var tilePaddingFromTileSize = function (
```
bre1470: Maybe you can change that to the regular function syntax, as it does not need to be variable.

PR #12960, opened 2020-02-13
bugfix/12870-hover-point-boost-module.

### LINKED ISSUE #12870: Regression in 7.1.3: Tooltip hover markers are not rendered in boosted charts
#### Expected behaviour
Tooltip hover markers should be visible in boosted charts

image

#### Actual behaviour
Tooltip markers don't show up when a chart is in boost mode since `7.1.3` (so `7.2.0`, `7.2.1`, and `8.0.0` as well)

image

Example for 8.0.0: https://jsfiddle.net/m3hwok5d/ (not working)
Example for 7.1.2: https://jsfiddle.net/m3hwok5d/1/ (working)

#### Product version
Highcharts `7.1.3`, `7.2.0`, `7.2.1`, and `8.0.0`

#### Affected browser(s)
Chrome, Safari, Firefox

mateuszkornecki:
Fixed #12870, hover point is not displayed when boost module is enabled.

### DIFF
--- a/js/modules/boost/boost-overrides.js
+++ b/js/modules/boost/boost-overrides.js
@@ -94,6 +94,7 @@ Series.prototype.getPoint = function (boostPoint) {
         point.plotX = boostPoint.plotX;
         point.plotY = boostPoint.plotY;
         point.index = boostPoint.i;
+        point.isInside = this.isPointInSight(boostPoint);
     }
     return point;
 };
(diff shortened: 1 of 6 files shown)


mateuszkornecki:
The newly created method takes into account the axes on which the series is drawn instead of the entire plot area. Using the existing one may cause some bugs similar to #7650.

TorsteinHonsi:
Thanks!

PR #12964, opened 2020-02-14
tools/package-dependencies

bre1470:
Updated package dependencies.


bre1470:
Maybe that helps also with the visual tests.

PR #12965, opened 2020-02-14
Refactored Time.get and .set to class members

TorsteinHonsi:
So we don't have to typecheck whenever using them.

### DIFF
--- a/ts/parts/Time.ts
+++ b/ts/parts/Time.ts
@@ -22,7 +22,7 @@ declare global {
         Date: any;
     }
     namespace Highcharts {
-        type TimeUnit = (
+        type TimeUnitValue = (
             'Date'|
             'Day'|
             'FullYear'|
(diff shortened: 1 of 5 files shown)


ts/parts/Time.ts
```diff
@@ -395,16 +405,100 @@ class Time {
...
+    /**
+     * Get the value of a date object in given units, and subject to the Time
+     * object's current timezone settings. This function corresponds directly to
+     * JavaScripts `Date.getXXX / Date.getUTCXXX`, so instead of calling
+     * `date.getHours()` or `date.getUTCHours()` we will call
+     * `time.get('Hours')`.
+     *
+     * @function Highcharts.Time#get
+     *
+     * @param {Highcharts.TimeUnit} unit
```
bre1470: In public (API-docs-sense) members all types have to be available as a doclet. The name should end on `Value` as it contains string literals only. So in this case, the minimum is
```ts
/**
 * Time units
 * @typedef {"Date"|"Day"|"FullYear"|"Hours"|"Milliseconds"|"Minutes"|"Month"|"Seconds"} Highcharts.TimeUnitValue
 */
```

TorsteinHonsi: Done, thanks.

PR #12974, opened 2020-02-17
tools/utilities-refactor-formatSingle

marvin19:
Integrated `formatSingle` into `format`. Removed `datePropsToTimestamps`.

### DIFF
--- a/js/parts/Utilities.js
+++ b/js/parts/Utilities.js
@@ -1302,62 +1302,6 @@ var wrap = H.wrap = function wrap(obj, method, func) {
     };
 };
 /**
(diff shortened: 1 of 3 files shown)


TorsteinHonsi:
Approved, but there's a visual diff that seems to be wrong in the master, correct in the candidate. Before merging, we need to understand why it differs.

bre1470:
As far as I remember, the test was using the wrong variables (data options) for formatting.

bre1470:
Maybe I should clarify this further: The test has been fixed in this PR, so it is expected.

PR #12975, opened 2020-02-17
bugfix/category-center

raf18seb:
Fixed 2 bugs in https://github.com/highcharts/highcharts/pull/12846


ts/parts/Stacking.ts
```diff
@@ -671,6 +671,7 @@ Axis.prototype.cleanStacks = function (this: Highcharts.Axis): void {
 Series.prototype.setGroupedPoints = function (this: Highcharts.Series): void {
     if (
         this.options.centerInCategory &&
+        this.type.match(/column|columnrange|bar/g) &&
```
TorsteinHonsi: Use `this.is` instead of string type matching, otherwise we won't be able to catch inherited series. Since `bar` inherits `column`, it should be enough to do `(this.is('column') || this.is('columnrange'))`

PR #12983, opened 2020-02-19
tools/utilities-small-fixes

marvin19:
Tools: Added missing imports


ts/parts-map/ColorAxis.ts
```diff
@@ -142,6 +142,7 @@ const {
     addEvent,
     erase,
     extend,
+    Fx,
```
bre1470: I believe, this import lacks usage further down in the code (search for H.Fx).

bre1470:
Looks good to me. 👍

bre1470:
@marvin19 
There are now some conflict with the master, that have to be adressed.

PR #12987, opened 2020-02-19
bugfix/9682-scroll-page-mobile

### LINKED ISSUE #9682: Not possible to scroll site on mobile, when touching legend items.

Izothep:
Fixed #9682, touching legend prevented page scroll.

### DIFF
--- a/js/parts/SvgRenderer.js
+++ b/js/parts/SvgRenderer.js
@@ -1210,7 +1210,7 @@ extend(SVGElement.prototype, /** @lends Highcharts.SVGElement.prototype */ {
      *         The SVGElement for chaining.
      */
     on: function (eventType, handler) {
-        var svgElement = this, element = svgElement.element, touchEventTime = 0, touchStartPos;
+        var svgElement = this, element = svgElement.element, touchStartPos, touchEventFired;
         // touch
         if (hasTouch && eventType === 'click') {
             element.ontouchstart = function (e) {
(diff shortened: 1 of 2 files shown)


Izothep:
Thanks for the review! 

With current code `hasMoved` is checked for both Android and iOS devices as `touchEnd` event is fired for both Android and iOS. 

Additionally, I don't think we need to check `userAgent` anymore. I have tested the code without this line with multiple examples (also mentioned in first created issue) and all looks good on both Android and iOS.

Izothep:
@TorsteinHonsi I have refactored the code removing userAgent dependency (and additionally time dependency at it was created for the same issue) and instead I have created a touchEventFired check that will help with legendItemClick issue on old Android. 

I have tested it on Android 4.0.4 and 4.2.0 and everything worked as expected.

TorsteinHonsi:
Very nice, thanks!

PR #12996, opened 2020-02-20
Annotations a11y: Demo consistency across OS/browsers.

oysteinmoseng:
Seems to be variations in fonts, as height is not consistent across browsers/OS. Setting axis min avoids any issues with not having enough space for the first annotation to fall below the point.


TorsteinHonsi:
Approved, awaiting conflict resolution and CI tests.

PR #13000, opened 2020-02-20
new options to support northern and southern hemis

cgalvarino:
By default, the windbarbs module presents barbs that are Northern hemisphere friendly.  Our friends in the Southern hemisphere view things differently (the dashes and pennants are on the left of the stem).

I needed a few extra options for our setup here.  Perhaps others might benefit.

```
/**
 * Include directional arrow head.
 */
arrowHead: true,
/**
 * Wind barbs in the northen hemisphere have barbs and pennants drawn on
 * the right of the stem.  Southern on the left.
 */
hemisphere: 'northern',
/**
 * True if data needs to be converted to knots.
 */
convertToKnots: true
```

I freely admit that I have not tested this as a standalone module inside Highcharts.  I have tested it inside our application, though.  Thanks for such an awesome project!

Northern hemi
image

Southern hemi
image

### DIFF
--- a/js/modules/windbarb.src.js
+++ b/js/modules/windbarb.src.js
@@ -155,7 +155,20 @@ seriesType('windbarb', 'column'
      *
      * @since 6.1.0
      */
-    xOffset: 0
+    xOffset: 0,
+    /**
+     * Include directional arrow head.
+     */
+    arrowHead: true,
+    /**
+     * Wind barbs in the northen hemisphere have barbs and pennants drawn on
+     * the right of the stem.  Southern on the left.
+     */
+    hemisphere: 'northern',
+    /**
+     * True if data needs to be converted to knots.
+     */
+    convertToKnots: true
 }, {
     pointArrayMap: ['value', 'direction'],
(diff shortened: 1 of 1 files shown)


pawelfus:
Thanks for sharing your solution!

cgalvarino:
@TorsteinHonsi I agree that the `convertToKnots` is unattractive, but it appears that the original code assumes the data needs it?  https://github.com/highcharts/highcharts/blob/73bcd6a7987912772516f551b4f290712cbc99b8/ts/modules/windbarb.src.ts#L300

TorsteinHonsi:
> but it appears that the original code assumes the data needs it?

Yes, apparently the definition of the barbs is based on knots.

cgalvarino:
Our data happened to already be in knots, so when we passed them through the windbarb module, the barbs were incorrect.  That's why I added the flag to reduce the ambiguity.  I suppose an alternative would be to document that windbarbs expects the data to be in m/s?

TorsteinHonsi:
> I suppose an alternative would be to document that windbarbs expects the data to be in m/s?

Yes, we should do that

Issue #12985, opened 2020-02-19
Pie datalabel y position offset ignored on points


kzoon:
#### Expected behaviour
I can set the y position offset per point

#### Actual behaviour
The  y position offset per point is ignored

#### Live demo with steps to reproduce
See https://jsfiddle.net/kzoon/fvcs1h8x/
The data label y position offset of -30 specified for Firefox is ignored in favor of the -3 specified for the pie chart.

(I am trying to correct the position of data labels that have a different font size)

#### Product version
Latest

pawelfus:
Hi @kzoon 

Thanks for reporting the issue. `series.data[].dataLabels.y` (and `x`) were never implemented for a pie chart - it's a bug in docs. I don't see any reason for this inconsistency for a pie series - I think we should add it (so it will work just like in column: https://jsfiddle.net/BlackLabel/5t1zd8as/)

kamilkulig:
Thank you for reporting the problem @kzoon - it's a bug.
_Internal note:_
It seems that pie series has always ignored `data[i].dataLabels.y` properties.

PR #13004, opened 2020-02-21
enhancement/12985-pie-datalabel-y-position

wchmiel:
Added pie series `dataLabel` x and y position per point, see #12985.

### DIFF
--- a/js/parts/DataLabels.js
+++ b/js/parts/DataLabels.js
@@ -886,13 +886,19 @@ if (seriesTypes.pie) {
                 };
                 dataLabel._pos = {
                     x: (x +
-                        options.x +
+                        ( // (#12985)
+                        ((point.options || {}).dataLabels || {}).x ||
+                            options.x) +
                         ({
                             left: connectorPadding,
                             right: -connectorPadding
                         }[labelPosition.alignment] || 0)),
                     // 10 is for the baseline (label vs text)
-                    y: y + options.y - 10
+                    y: (y +
+                        ( // (#12985)
+                        ((point.options || {}).dataLabels || {}).y ||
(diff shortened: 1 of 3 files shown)


js/parts/DataLabels.js
```diff
@@ -1027,6 +1033,9 @@ if (seriesTypes.pie) {
      * @return {void}
      */
     seriesTypes.pie.prototype.placeDataLabels = function () {
+        // Zmienic pozycje labelek tutaj
+        // (jak to zrobic zeby na serii jak sie ustawi
+        // to nie bylo brane pod uwage tylko w punkcie?)
```
KacperMadej: ❤️

js/parts/DataLabels.js
```diff
@@ -886,13 +886,19 @@ if (seriesTypes.pie) {
                 };
                 dataLabel._pos = {
                     x: (x +
-                        options.x +
+                        ( // (#12985)
+                        ((point.options || {}).dataLabels || {}).x ||
```
KacperMadej: `(point.options || {}).dataLabels || {})` - is used twice in code, so it could be done once and stored as some variable

ts/parts/DataLabels.ts
```diff
@@ -1521,18 +1521,26 @@ if (seriesTypes.pie) {
                     align: (labelPosition as any).alignment
                 };
 
+                // (#12985)
+                pointDataLabelsOptions =
+                    merge(options, point.options.dataLabels);
```
pawelfus: Do we need to use `merge()` ? It's quite heavy.
In this case, simple `pick()` (because of numerical values for x/y) should be used.

Issue #12901, opened 2020-02-05
Add defer option as animation parameter


Basts:
@pawelfus @TorsteinHonsi 

As in the title, there was a proposal to add the defer option as the animation parameter in the animation engine.

Suggested defer-and-fade-in sequence: 

```
element.attr({ opacity: 0 })
    .animate({
        defer: 1000,
        duration: 150,
        opacity: 1
    });

```

When this feature will be ready to use current defer logic could be removed from series labels, data labels and annotations and replaced by the new one. Additionally, I think that it should be easy to implement it for stackLabels. 

I would like to discuss the idea that the **defer** could be also a number which value could be set as the defer time for particular label or annotation in their options if user would like to set it too (not only a value inherited from series).

pawelfus:
I suggest to implement this is steps:
- [ ] - implement `SVGElement.animate({ defer })` - can be PR-drafted, before proceeding further
- [ ] - implement `series.animation.defer`
- [ ] - replace `series.dataLabels.defer` with [to be decided] `series.dataLabels.animation.defer` 
- [ ] - replace logic in `series.label` and add `series.label.animation.defer` option (if needed)
- [ ] - add `annotation.animation.defer`, `annotation.shapes[].animation.defer` and `annotation.labels[].animation.defer`
- [ ] - consider implementing `yAxis.stackLabels.animation.defer`

Quite a lot of new options, perhaps some of them are not needed (e.g. `series.label.animation.defer`, or `annotation.shapes[].animation.defer`) and can be implemented later, on demand. Every suggested option above should inform in docs that `defer` is applied only on chart/series/annotation/etc. init.

PR #13005, opened 2020-02-21
enhancement/12901-defer-animation-parameter

Basts:
Enhancement #12901, defer option as an animation parameter.


Basts:
Demo to test current implementation: https://jsfiddle.net/BlackLabel/zn53secu/

Suggested sequence to use: 

`element.attr({ opacity: 0 })
    .animate({
        opacity: 1
    }, {
        defer: 1000,
        duration: 150,
    });
`

Where the first obj is a params attributes, second one is the animOptions to trigger. 

To do: 

- [ ]  - implement SVGElement.animate({ defer }) - can be PR-drafted, before proceeding further
 

- [ ] - implement series.animation.defer

 

- [ ] - replace series.dataLabels.defer with [to be decided] series.dataLabels.animation.defer

 

- [ ] - replace logic in series.label and add series.label.animation.defer option (if needed)

- [ ]  - add annotation.animation.defer, annotation.shapes[].animation.defer and annotation.labels[].animation.defer

- [ ]  - consider implementing yAxis.stackLabels.animation.defer

ts/parts/SvgRenderer.ts
```diff
@@ -976,7 +978,9 @@ extend((
             if (complete) {
                 animOptions.complete = complete;
             }
-            animate(this, params, animOptions);
+            H.syncTimeout((): void => {
```
KacperMadej: Load `syncTimeout` from `Utils` as shown in https://github.com/highcharts/highcharts/commit/4aa3cff7e9277f5648f209c8fd671e3a9de1b026

ts/parts/Utilities.ts
```diff
@@ -64,6 +64,7 @@ declare global {
             complete?: Function;
             curAnim?: Dictionary<boolean>;
             duration?: number;
+            defer?: number;
```
KacperMadej: Interface properties should be sorted alphabetically (move `defer` above `duration`)

Basts:
@KacperMadej Thx 👍

ts/parts/Series.ts
```diff
@@ -6707,6 +6707,11 @@ H.Series = seriesType<Highcharts.LineSeries>(
             // overwrite the animation.complete option which should be available
             // to the user).
             if (!hasRendered) {
+                // Additional time if defer is defined before afterAnimate
+                // will be triggered
+                if (animDuration && defined(animObject(options.animation).defer)) {
```
TorsteinHonsi: Just like `animObject` always returns an object with a duration, it should also always return a `defer` property (0 by default). That will save us from type checking in cases like this.

ts/parts/SvgRenderer.ts
```diff
@@ -961,8 +962,10 @@ extend((
         complete?: Function
     ): Highcharts.SVGElement {
         var animOptions = animObject(
-            pick(options, this.renderer.globalAnimation, true)
-        );
+                pick(options, this.renderer.globalAnimation, true)
+            ),
+            deferTime = animOptions.defer || 0;
```
TorsteinHonsi: As above

ts/parts/Series.ts
Basts: @TorsteinHonsi 
I have a problem with this change and I would like to discuss it.

```
const animObject = H.animObject = function animObject(
    animation?: (boolean|Highcharts.AnimationOptionsObject)
)Highcharts.AnimationObject {
    return isObject(animation) ?
        H.merge(
            { duration: 500, defer: 0 },
            animation as Highcharts.AnimationOptionsObject
        ) as any :
        { duration: animation as boolean ? 500 : 0, defer: 0 };
}; 
```

That will be the best implementation because defer will be always assigned however requires changes in the current **Highcharts.AnimationOptionsObject** definition and probably create the new definition, like **Highcharts.AnimationObject**, which will look like this: 

```
        interface AnimationOptionsObject {
            complete?: Function;
            curAnim?: Dictionary<boolean>;
            defer: number;
            duration: number;
            easing?: (string|Function);
            step?: AnimationStepCallbackFunction;
        }
```

This implementation will avoid getting the defer or duration undefined. 

What do you think? What is your opinion about it?

ts/annotations/annotations.src.ts
```diff
@@ -1047,12 +1050,38 @@ merge(
...
+                        (plotOptions.series && defined(plotOptions.series.animation) ?
+                            (
+                                (
+                                    plotOptions.series.animation.defer &&
+                                    plotOptions.series.animation.duration
+                                ) ?
+                                // If defer and duration are set the animation
+                                //  will be triggered after their sum value
+                                    plotOptions.series.animation.defer + plotOptions.series.animation.duration :
+                                    plotOptions.series.animation.defer ? plotOptions.series.animation.defer :
```
KacperMadej: If I am counting correctly this is the 5th level of the nested ternary operators. Readability is low. It would be nice to see this rewritten into something maintainable more easily.

js/parts/Series.js
```diff
@@ -4971,7 +4972,7 @@ null,
         if (!hasRendered) {
             // Additional time if defer is defined before afterAnimate
             // will be triggered
-            if (animDuration && defined(animObject(options.animation).defer)) {
+            if (animDuration && (animObject(options.animation).defer)) {
```
KacperMadej: `(animObject(options.animation).defer)` -> `animObject(options.animation).defer`

Basts: I will fix it, but it is not the final code, because how this logic will look like depends on the above conversation about implementing the defer inside the **animObject**.

ts/annotations/annotations.src.ts
Basts: @KacperMadej please check the current implementation and let me know what do you think about it.

KacperMadej: Thnak you :+1: - it's easier to follow the code now.

whispers: _(You could also try to eliminate the any casting and set the default value (from the last else) at the top (when the variable is created), but those are just minor details and really NOT worth taking care of right now if the code is not final.)_

ts/parts/Series.ts
Basts: This is the only thing which left to implement (I think). Any opinion is welcome.

TorsteinHonsi: Yes, defer and duration will now always be a number in the object that `animObject` returns. As opposed to the `AnimationOptionsObject` where those are optional.

One solution would be using two interfaces like you suggest.

The other solution I think would be that the _option_ is a partial of the object. @bre1470 do you think that solution would apply in this case?

bre1470: @TorsteinHonsi 
I fear that this would lead to complications later on, as both interfaces are not logically connected, just by the current implementation. I would suggest something shorter like
```ts
interface AnimationObject extends AnimationOptionsObject {
    defer: number;
    duration: number;
}
```
That connects output with input in the way, the implementation is working.

Regarding the implementation, the following line can be reduced by ` as boolean` as it provides no benefit:
```ts
{ duration: animation as boolean ? 500 : 0, defer: 0 };
```

ts/modules/series-label.src.ts
```diff
@@ -152,6 +153,21 @@ H.setOptions({
              */
             label: {
 
+                /**
+                 * Whether to defer displaying the series label until the set
+                 * time in milliseconds has finished. Setting to `false` renders
+                 * the series label immediately. If set to `true` inherits
+                 * the defer time set in [series.animation](#series.animation).
+                 *
+                 * @sample highcharts/plotoptions/defer
```
pawelfus: Wrong path

ts/annotations/annotations.src.ts
```diff
@@ -497,6 +501,21 @@ merge(
...
+             * time set in [plotOptions.series.animation](#plotOptions.series.animation).
+             *
+             * @sample highcharts/annotations/defer
+             *         Set defer time
+             *
+             * @since        8.0.3
+             *
+             * @type {boolean | number}
+             */
+            defer: true,
```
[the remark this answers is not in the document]
TorsteinHonsi: Yes, absolutetly! The `defer` option shouldn't be standalone now that it is so closely coupled to animation. 

As for `dataLabels.animation`, we still need to support the `dataLabels.defer` option, but it should relay directly to the animation object.

ts/modules/series-label.src.ts
```diff
@@ -1127,12 +1144,17 @@ function drawLabels(this: Highcharts.Chart, e: Event): void {
...
                     animObject(series.options.animation).duration as any
                 );
             }
-
+            // Add the label or the series defer time
+            if (options.defer || defer) {
+                isNumber(options.defer) ?
+                    (delay as any) += options.defer :
+                    delay += defer;
+            }
```
KacperMadej: It took me a while - something doesn't feel right here. The docs say: 
```
                                              ...Setting to `false` renders
                 * the series label immediately. If set to `true` inherits
                 * the defer time set in [series.animation](#series.animation).
```
If `options.defer` is `true` then `delay += defer` runs no matter what the value of `defer` is. I'm ok with that.
If `options.defer` is `false` then `delay += defer` runs if `defer` is not zero (as it has to be a number), so this makes me think that something is wrong here.
And if `options.defer` is zero then `delay += options.defer` runs (if `delay` is not zero) or nothing happens, so basically the outcome is the same - no change to `delay`. This 3rd scenario fits better the description of `renders the series label immediately`.

Basts:
@TorsteinHonsi @pawelfus 

I have an idea for a sequence to control the defer in dataLabels: 

```
dataLabels: {
   defer: true, // boolean value to support the old option
   animation: {
       defer: number // here, the user has a possibility to set a value for delaying the animation
   }
}
```

The dataLabels.defer set to the false makes that value set for animation.defer is not respected. What do you think?

pawelfus:
I would say `defer` and `animation` defer should work the same way:
```js
dataLabels: {
  defer: boolean | number,
  animation: {
    defer: boolean | number
  }
}
```
So we keep this consistent with other `series/annoatations.animation.defer` options.

When resolving `defer` for dataLabels, simply decide which has higher priority (can be `defer`). Like this:
```js
var defer = pick(dataLabels.defer, dataLabels.animation.defer, true);

if (defer) {
  ...
}
```

Basts:
Demo to test options: http://jsfiddle.net/BlackLabel/ebhrygm9/

Demo to test defer in annotations: https://jsfiddle.net/BlackLabel/amb9Loq7/

Demo to test defer implementation in SVGElement.animate: https://jsfiddle.net/BlackLabel/h0m6zrgs/

ts/annotations/annotations.src.ts
```diff
@@ -1640,7 +1670,14 @@ extend(chartProto, /** @lends Highcharts.Chart# */ {
         this.plotBoxClip.attr(this.plotBox);
 
         this.annotations.forEach(function (annotation): void {
+            const deferDiff = annotation.deferTime - annotation.durationTime;
             annotation.redraw();
+            annotation.graphic.animate({
+                opacity: 1
+            }, {
+                defer: deferDiff === 0 ? annotation.deferTime : deferDiff,
```
pawelfus: Are you sure this ternary op is needed? It's quite the difference between setting `defer=200` (and below 200) and `defer=201` (and above):
- defer = 200 -> after 200ms run 200ms animation
- defer = 201 -> after 1ms run 200ms animation

ts/parts/Axis.ts
```diff
@@ -3590,6 +3590,33 @@ class Axis {
          */
         stackLabels: {
 
+            /**
+             * The animation configuration object for the stackLabels. Please
```
pawelfus: stackLabels -> `stackLabels`

ts/parts/Stacking.ts
```diff
@@ -626,6 +633,17 @@ Axis.prototype.renderStackTotals = function (this: Highcharts.Axis): void {
...
+
+    deferTime = getDeferTime(chart, defer);
+    durationTime = Math.min(deferTime, 200);
+    deferDiff = deferTime - durationTime;
+
+    stackTotalGroup.animate({
+        opacity: 1
+    }, {
+        duration: durationTime,
+        defer: deferDiff === 0 ? deferTime : deferDiff
```
pawelfus: The same as above: defer=200 vs defer=201.

ts/parts/Series.ts
```diff
@@ -1075,6 +1077,13 @@ H.Series = seriesType<Highcharts.LineSeries>(
             duration: 1000
         },
 
+        /**
+         * @default   0
+         * @type      {number}
```
pawelfus: Type: number vs number|boolean

ts/parts/Axis.ts
```diff
@@ -3590,6 +3590,33 @@ class Axis {
...
+            animation: {
+                /** @internal */
+                defer: true
+            },
+
+            /**
+             * @default   true
+             * @type      {boolean|number}
+             * @since     next
+             * @apioption yAxis.stackLabels.animation.defer
```
pawelfus: Why sometimes `defer` is defined as `boolean|number` (annotation, dataLabels, stackLabels) and for series is just `number` ?

Basts: I think that setting `defer` as a boolean in series doesn't make sense, because setting it to true will set it as `series.animation.defer = 1`, meanwhile setting defer to `true` for others (annotation, dataLabels, stackLabels) makes that defer value is taken from this one set in series.

ts/annotations/annotations.src.ts
Basts: After internal discussion - we back to the previous implementation, so:

```js
    durationTime = Math.min(deferTime, 200);
    stackTotalGroup.animate({
        opacity: 1
    }, {
        duration: durationTime,
        defer: deferTime - durationTime
    });
```

which means that: 

```js
//duration is set to 200
if (defer < duration) {
defer = 0
duration = defer
}
```

pawelfus:
> Do we need public options for dataLabels.animation and stackLabels.animation at all?

Stack labels should provide an option to go back to v8.0.4 (in short: to disable deferral).
`dataLabels.animation` is not necessary, but we decided it would be inconsistent across all other options (`series.animation.defer`, `annotation.animation.defer`, `stackLabels.animation.defer`). We can use `series.defer`, `annotation.defer` and `stackLabels.defer` instead, but it feels wrong in terms of implementation (`item.animate(options, { defer, duration })`. I'm afraid if future requests will show up, we may end up with different API options; I would like avoid that.

Note that implementation of stackLabels is quite easy: `getDeferTime(chart, item.options.animation.defer);` and removing this is as simple as changing to `getDeferTime(chart, true);`.

Other suggestions (`defer: undefined` and TS`) - I agree with you 👍

TorsteinHonsi:
> Other suggestions (defer: undefined and TS`) - I agree with you 👍

Ok, let's do those changes then.

ts/parts/Utilities.ts
```diff
@@ -2641,6 +2654,54 @@ const getStyle = H.getStyle = function (
...
+ * @param {Highcharts.Chart} chart
+ *        The chart instance.
+ *
+ * @param {number | boolean} defer
+ *        Defer defined in the stackLabels/annotations options
+ *
+ * @return {number}
+ *         The numeric value.
+ */
+const getDeferTime = H.getDeferTime = function (
```
TorsteinHonsi: I think a bit of refactoring is needed here.
* Is this function supposed to mimick the current behaviour of data labels? In this case it should also be used by data labels.
* It currently only looks at `plotOptions`-level options, but what if the user changes the `defer` or `duration` settings of the series item? Probably a general Utilities.ts function is not the way to go, it should rather be a `Series` method. For stack labels it can simply do an array.reduce over the stack series and use the highest `deferTime` reported by the series.
* In the last condition it looks for `plotOptions.line.animation` which can't be right for other series types... But all this will hopefully sort out correctly if we move the log to the Series object and only look for `this.options.animation`.

Basts: @TorsteinHonsi

Your proposal to rewrite it as a `Series` method totally makes sense for the `stackLabels` and the `dataLabels` - I agree. But what about the `annotation`? If `annotation.animation.defer` will be enabled without any number value - where the defer value should come from? From the `plotOptions` as it is now?

TorsteinHonsi: I think in that case it should be reduced from `chart.series`, the same way as the defer from stack labels is reduced from `yAxis.series`.

Basts:
To sum up:

1. 
```js
stackLabels/annotation: {
   animation: {
       defer: 0 - show label immediately,
       defer: number - show after those number value
   }
   // animation object is not defined (default) - the defer time is getting by the getDeferTime 
}
```
2.
```js
dataLabels: {
   defer: boolean - works as it used to be,
   // animation object is not defined (default) - above options works,
   animation:  {
       defer: 0 - show label immediately,
       defer: number - show after those number value
   }
}
```
3. **getDeferTime** logic has been changed - function gets the the highest deferTime reported by the series.

Annotation demo: https://jsfiddle.net/BlackLabel/ztj9xkb1/

Options demo: http://jsfiddle.net/BlackLabel/gx1h03np/

samples/unit-tests/axis/stacklabels/demo.js
```diff
@@ -324,3 +324,79 @@ QUnit.test('StackLabels outside xAxis min & max range are displayed #12294', fun
...
+            }),
+            done = assert.async(),
+            slOpacity;
+
+        setTimeout(function () {
+            // animation started
+            slOpacity = chart.yAxis[0].stackTotalGroup.opacity;
+            assert.strictEqual(
+                slOpacity === 0,
+                true,
```
pawelfus, line 371: Instead of 
```
            assert.strictEqual(
                slOpacity === 0,
                true,
```

Use
```
            assert.strictEqual(
                slOpacity,
                0,
```

+ change `slOpacity` variable name to something more readable :)

samples/unit-tests/axis/stacklabels/demo.js
```diff
@@ -324,3 +324,79 @@ QUnit.test('StackLabels outside xAxis min & max range are displayed #12294', fun
...
+                    'Animation should be started but not finished'
+                );
+            }, 200);
+
+            setTimeout(function () {
+                slOpacity = chart.yAxis[0].stackTotalGroup.opacity;
+                // animation finished
+                assert.strictEqual(
+                    slOpacity === 1,
+                    true,
```
pawelfus, line 390: The same as above.

samples/unit-tests/axis/stacklabels/demo.js
```diff
@@ -324,3 +324,79 @@ QUnit.test('StackLabels outside xAxis min & max range are displayed #12294', fun
...
+                    'Animate should be finished - stackLabels should be visible'
+                );
+                // all tests are done
```
pawelfus, line 393: Nitpicking :) `all` -> `All`

ts/parts/SvgRenderer.ts
```diff
@@ -959,26 +960,31 @@ extend((
...
+
 
         // When the page is hidden save resources in the background by not
         // running animation at all (#9749).
         if (pick(doc.hidden, doc.msHidden, doc.webkitHidden, false)) {
             animOptions.duration = 0;
         }
 
-        if (animOptions.duration !== 0) {
+        if (animOptions.duration !== 0 && defined(deferTime)) {
```
pawelfus, line 978: Is that necessary? Doesn't now `animObject`always return `{ duration, defer }` ?

ts/parts/Utilities.ts
```diff
@@ -2641,6 +2650,32 @@ const getStyle = H.getStyle = function (
...
+ *        The chart instance.
+ *
+ * @return {number}
+ *        The numeric value.
+ */
+const getDeferTime = H.getDeferTime = function (chart): number {
+
+    const seriesDelay = chart.series.map(function (series): number {
+        var seriesAnim = series.options.animation as any,
+            seriesDuration = seriesAnim.duration;
```
pawelfus, line 2668: Should be `seriesAnim && seriesAnim.duration`. Otherwise can throw errors that you check below (`if (seriesAnim)`).

ts/parts/Utilities.ts
```diff
@@ -2641,6 +2650,32 @@ const getStyle = H.getStyle = function (
...
+
+        if (seriesAnim) {
+            return seriesAnim.defer ? seriesDuration + seriesAnim.defer : seriesDuration;
+        }
+        // If seriesAnim doesn't exist - return 0;
+        return 0;
+    });
+    return Math.max(...seriesDelay);
```
pawelfus, line 2676: Not a big deal, but we can replace this with `Math.max.apply(null, seriesDelay)` - avoiding `__spread` implementation in JS.

Perhaps even faster would be to use `forEach` + variable that is returned at the end:
```js
let maxDefer = 0;

series.forEach(s =>
  if (s.options.animation) {
      Math.max(maxDefer, (duration ? duration : 0) + (defer ? defer : 0));
  }
);

return maxDefer;
```

ts/annotations/annotations.src.ts
```diff
@@ -1057,12 +1088,17 @@ merge(
...
+            const chart = this.chart,
+                defer = animObject(this.options.animation).defer;
+
             this.linkPoints();
             this.addControlPoints();
             this.addShapes();
             this.addLabels();
             this.addClipPaths();
             this.setLabelCollector();
+            this.deferTime = defined(defer) ? defer : getDeferTime(chart);
```
pawelfus, line 1100: `defer` should now be always defined, right? Then:
- `defer = animObject(this.options.animation).defer;` -> `defer = 0`
- `this.deferTime = defined(defer) ? defer : getDeferTime(chart);` -> `this.deferTime = 0`

ts/parts/Utilities.ts
```diff
@@ -50,6 +50,7 @@ declare global {
         interface AnimationOptionsObject {
             complete?: Function;
             curAnim?: Dictionary<boolean>;
+            defer?: number;
             duration?: number;
```
TorsteinHonsi, line 54: The `duration` property should not be optional in this interface.

ts/parts/Utilities.ts
```diff
@@ -2641,6 +2650,32 @@ const getStyle = H.getStyle = function (
...
+ * @param {Highcharts.Chart} chart
+ *        The chart instance.
+ *
+ * @return {number}
+ *        The numeric value.
+ */
+const getDeferTime = H.getDeferTime = function (chart): number {
+
+    const seriesDelay = chart.series.map(function (series): number {
+        var seriesAnim = series.options.animation as any,
```
TorsteinHonsi, line 2667: Use `animObject` so you don't have to do type checking or any casting.

ts/parts/Utilities.ts
```diff
@@ -2362,11 +2368,14 @@ const setAnimation = H.setAnimation = function setAnimation(
  *         An object with at least a duration property.
  */
 const animObject = H.animObject = function animObject(
-    animation?: (boolean|Highcharts.AnimationOptionsObject)
-): Highcharts.AnimationOptionsObject {
+    animation?: (boolean|Partial<Highcharts.AnimationOptionsObject>)
+): Partial<Highcharts.AnimationOptionsObject> {
```
TorsteinHonsi, line 2372: This function should not return a Partial. By returning the full set, we know that `duration` is always a number.

ts/parts/Stacking.ts
```diff
@@ -603,10 +603,10 @@ Axis.prototype.renderStackTotals = function (this: Highcharts.Axis): void {
...
         renderer = chart.renderer,
         stacks = axis.stacks,
-        deferTime,
-        durationTime,
-        stackLabelsAnim = animObject(axis.options.stackLabels.animation),
-        defer = stackLabelsAnim.defer,
+        stackLabelsAnim = axis.options.stackLabels.animation,
+        deferTime = stackLabelsAnim && typeof stackLabelsAnim.defer === 'undefined' ?
+            getDeferTime(chart) : animObject(stackLabelsAnim).defer,
+        durationTime = Math.min(deferTime, 200),
```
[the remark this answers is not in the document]
Basts: @TorsteinHonsi 

Hm, I am not sure, I borrowed this 'logic' from `dataLabels`, where `fadeInDuration` was calculated in this same way. Do you think that fixed value, like 200, will be a better option?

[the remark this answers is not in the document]
pawelfus: Ah.. I see it now, thanks! 👍

ts/parts/Utilities.ts
```diff
@@ -50,8 +50,8 @@ declare global {
         interface AnimationOptionsObject {
```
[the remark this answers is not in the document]
Basts: @TorsteinHonsi 

Referring to the message about using zapier/slack message alert - could you take a look and share your opinion on it and the above requires?

Basts:
Closed as #13688

Issue #12840, opened 2020-01-23
Column (Bar) Filled with pattern disappears when we click on another chart in the same page. PATTERN FILL


rammohan201:
#### Expected behavior
I want each chart to behave on its own, Pattern Fill columns should not be dependent on any other chart, pattern filled column should display even after chart reflow or update or legend toggle of other column, 

#### Actual behavior
my application page has multiple charts of same type rendered in different accordions. It is an angular application, I have three bars in a chart, one of them is having pattern fill, I am facing issue with pattern filled bar. technically, chart components are independent, but some how, pattern fill is displaying in other charts only if first chart is opened. If I close the first chart by collapsing the accordion, pattern filled column is not displaying, I tried adding id to pattern, when I clicked on legend of another column to hide and show, then pattern filled column is not displaying again, legend symbol is not displaying. Please let me know what I am doing wrong?

I referred to below existing similar issues and updated highcharts version to 8.0.0 but even then issue is not resolved.

https://github.com/highcharts/highcharts/issues/12053
https://github.com/highcharts/pattern-fill/issues/32

#### Live demo with steps to reproduce
https://stackblitz.com/edit/angular-highcharts-uk7qzx

In this demo, I added two charts, Chart 1 is hidden by default, we can see 2nd Chart by default, and it is not showing pattern filled column.
Then click on Show/Hide Chart1 button, then we can see pattern filled column displaying in two charts. 
Then hide first chart by clicking on Show/Hide Chart1 button, then we can see pattern filled column not displaying.

why 2nd chart's pattern filled column is dependent on first one? 
why hiding 1st chart is impacting 2nd chart's pattern filled column?

#### Product version
8.0.0

#### Affected browser(s)
Chrome, IE 11, Mozilla

pawelfus:
Hi @rammohan201 

Thanks for reporting the issue. This is caused by using the same pattern for different charts. When creating multiple charts with the same patterns' ID's all points actually refer to the first patterns (because of the first matching ID). When you hide first chart, you hide also the pattern and it seems that hidden pattern can't be use on a visible element.

I'm not sure what we can do with this issue, any ideas @oysteinmoseng ? Built-in patterns are generated with unique ID per chart.

Workaround is to use separate ID's per charts if you want to use a patterns that is hidden. For example: https://stackblitz.com/edit/angular-highcharts-v86vbt?file=app/app.component.ts

rammohan201:
[...]

PR #13024, opened 2020-02-26
Bugfix: Fixed issue with pattern fill on redraw when renderer.url is set

goransle:
Fixes an issue with pattern fills when `renderer.url` is set. Part of #12840

**Demo without fix:** https://jsfiddle.net/goransle/qLv8e2rn/
**Demo with fix applied:** https://jsfiddle.net/goransle/9z56mpxo/

Note: this PR does not solve the original issue of #12840, which is a side effect of two charts sharing the same custom pattern IDs.


ts/modules/pattern-fill.src.ts
```diff
@@ -740,13 +740,11 @@ addEvent(H.Chart, 'redraw', function (): void {
...
                         node.getAttribute('stroke');
-
                 if (id) {
-                    usedIds.push(
-                        id
-                            .substring(id.indexOf('url(') + 5)
-                            .replace(')', '')
-                    );
+                    const sanitizedId = id.replace(renderer.url, '').replace('url(#', '').replace(')', '');
+                    if (usedIds.indexOf(sanitizedId) === -1) {
```
oysteinmoseng: I see that we are using an array to keep track of the used IDs. We should probably just use a hash-type lookup object for this, e.g. `usedIds[sanitizedId] = true` where `usedIds = {}`. Then below where we loop through we can just run `usedIds[patternId]` to test if it exists - this should be much faster than searching the array.

goransle: Like so?

oysteinmoseng: Yep, (assuming this works)!

PR #13029, opened 2020-02-27
bugfix/11397-word-cloud-focus-border

### LINKED ISSUE #11397: Word Cloud keyboard access broken - accessibility
Hi to all! 
Working with Microsoft and came across this error.
Any help will be greatly appreciated!
Thanks!

#### Expected behavior
When navigating in the word cloud with the keyboard and use arrows up and down to select and focus on words, the words are selected with an outline around them.

#### Actual behavior
There is a non-moving outline in the center instead of around each word.

https://jsfiddle.net/b3enhvw5/

Select the word cloud and try moving around word with the arrows.

#### Product version
    highcharts: 7.1.1

#### Affected browser(s)
All

mateuszkornecki:
Fixed #11397, incorrect position of the focus border on text elements.

### DIFF
--- a/ts/modules/accessibility/focusBorder.ts
+++ b/ts/modules/accessibility/focusBorder.ts
@@ -40,6 +40,10 @@ declare global {
     }
 }
 
+interface TextAnchorCorrectionObject {
+    x: number;
+    y: number;
+}
 
 /* eslint-disable no-invalid-this, valid-jsdoc */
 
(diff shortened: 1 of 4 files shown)


ts/modules/accessibility/focusBorder.ts
```diff
@@ -73,9 +73,27 @@ extend(H.SVGElement.prototype, {
...
+            const isFirefox = H.isFirefox;
+            borderPosX = this.attr('x') as number - (bb.width * 0.5) - pad +
+                // Correct the offset caused by the browser.
+                (isRotated ? bb.height * 0.068 : 0);
+            borderPosY = this.attr('y') as number - (bb.height * 0.5) - pad +
+                // Firefox needs different correction value.
+                (isRotated ? 0 : -bb.height * (isFirefox ? 0.25 : 0.068));
+        } else {
+            borderPosX = bb.x - pad;
+            borderPosY = bb.y - pad;
```
KacperMadej: If you will set those values as default ones in lines 76-77 then the type of the variables will be closer to the set definitions (`number`) - now they start as `undefined`.

ts/modules/accessibility/focusBorder.ts
```diff
@@ -73,9 +73,27 @@ extend(H.SVGElement.prototype, {
         bb.x += this.translateX ? this.translateX : 0;
         bb.y += this.translateY ? this.translateY : 0;
 
+        let borderPosX: number,
+            borderPosY: number;
+
+        // For text elements, apply x and y offset, #11397.
+        if (this.element.nodeName === 'text') {
+            const isRotated = Boolean(this.rotation);
+            const isFirefox = H.isFirefox;
```
KacperMadej: If `H.isFirefox` is used once then there is no point for setting a variable for it.

ts/modules/accessibility/focusBorder.ts
```diff
@@ -73,9 +73,27 @@ extend(H.SVGElement.prototype, {
         bb.x += this.translateX ? this.translateX : 0;
         bb.y += this.translateY ? this.translateY : 0;
 
+        let borderPosX: number,
+            borderPosY: number;
+
+        // For text elements, apply x and y offset, #11397.
+        if (this.element.nodeName === 'text') {
+            const isRotated = Boolean(this.rotation);
```
KacperMadej: `!!this.rotation` will give you the same output and is more common in the rest of the code.

ts/modules/accessibility/focusBorder.ts
```diff
@@ -73,9 +73,27 @@ extend(H.SVGElement.prototype, {
...
+            const isFirefox = H.isFirefox;
+            borderPosX = this.attr('x') as number - (bb.width * 0.5) - pad +
```
KacperMadej: `as number` casting seems to avoidable - either through a check for `undefined` or parsing into a number. This depends on the TS error you were getting.

TorsteinHonsi:
Pending @oysteinmoseng 's review

ts/modules/accessibility/focusBorder.ts
```diff
@@ -67,17 +71,70 @@ extend(H.SVGElement.prototype, {
...
+         * @param {Highcharts.SVGElement} text
+         *
+         * @return {TextAnchorCorrectionObject}
+         */
+        function getTextCorrection(text: Highcharts.SVGElement): TextAnchorCorrectionObject {
+            let posXCorrection = 0,
+                posYCorrection = 0;
+
+            if (text.attr('text-anchor') === 'middle') {
+                posXCorrection = H.isFirefox && text.rotation ? 0.25 : 0.5;
```
TorsteinHonsi, line 99: The special treatment for Firefox, does it have a connection to this? https://github.com/highcharts/highcharts/blob/v8.0.4/ts/parts/SvgRenderer.ts/#L3705-L3729

ts/modules/accessibility/focusBorder.ts
```diff
@@ -67,17 +71,70 @@ extend(H.SVGElement.prototype, {
...
+                    {
+                        x: isRotated ? 1 : 0,
+                        y: 0
+                    };
+
+            borderPosX = +this.attr('x') - (bb.width * correction.x) - pad;
+            borderPosY = +this.attr('y') - (bb.height * correction.y) - pad;
+
+            if (isLabel && isRotated) {
+                [borderWidth, borderHeight] = [borderHeight, borderWidth];
```
TorsteinHonsi, line 127: Note that array destructuring adds a big blob of code in the compiled JS, so it should be avoided.

ts/modules/accessibility/focusBorder.ts
mateuszkornecki: I don't think it is related to this case. In the RTL case, Firefox was changing the order of elements, and now it has a different baseline than other browsers.

ts/modules/accessibility/focusBorder.ts
```diff
@@ -67,31 +71,70 @@ extend(H.SVGElement.prototype, {
...
+                posYCorrection = 0;
+
+            if (text.attr('text-anchor') === 'middle') {
+                posXCorrection = H.isFirefox && text.rotation ? 0.25 : 0.5;
+                posYCorrection = H.isFirefox && !text.rotation ? 0.75 : 0.5;
+            } else if (!text.rotation) {
+                posYCorrection = 0.75;
+            } else {
+                posXCorrection = 0.25;
+                posYCorrection *= (H.isFirefox ? 2 : 0);
```
KacperMadej, line 105: Isn't `posYCorrection` equal to `0` on entering this code line, so multiplying it doesn't change anything?

mateuszkornecki: You are right! That's probably the fragment of the previous correction logic that I forgot to delete. I will check that and fix it!

ts/modules/accessibility/focusBorder.ts
TorsteinHonsi: Makes sense, I think we have the same situation in the `textOutline` implementation.

PR #13030, opened 2020-02-27
enhancement/pane-class-es6

marvin19:
Tools: Migrated Pane class to ES6


ts/parts-more/Pane.ts
```diff
@@ -615,3 +608,11 @@ addEvent(H.Pointer, 'afterGetHoverData', function (
...
     }
 });
+
+H.Pane = Pane as any;
+
+const paneModule = {
+    Pane
+};
+
+export default paneModule;
```
bre1470: This should be replaced with a simple `export default H.Pane;`.

PR #13032, opened 2020-02-27
bugfix/12994-tooltip-hidedelay

### LINKED ISSUE #12994: The label appears after using the show() method on visible series
#### Expected behaviour
Tooltip doesn't show up.

#### Actual behaviour
If `tooltip.hideDelay` property is set to 0 and `tooltip.split` is enabled, the tooltip will appear after using show() method on a visible series.

Steps to reproduce:
   1. Hover over the graph so the tooltip element is rendered.
   2. Click the "show" button which will make the tooltip appear unnecessarily.

https://jsfiddle.net/BlackLabel/yL8902zh/

#### Possilble workaround 
`tooltip.hideDelay` higher than `0`

#### Product version

#### Affected browser(s)
All

kamilkulig:
Fixed #12994, `tooltip.hideDelay` of 0 was causing unexpected tooltip behavior.


ts/parts/Tooltip.ts
```diff
@@ -1033,7 +1033,7 @@ class Tooltip {
         delay = pick(delay, this.options.hideDelay, 500);
         if (!this.isHidden) {
             this.hideTimer = syncTimeout(function (): void {
-                tooltip.getLabel()[delay ? 'fadeOut' : 'hide']();
+                tooltip.getLabel().fadeOut();
```
KacperMadej: So now fadeOut animation for about 150ms is unavoidable even when setting delay to 0?
Then the comment above this function should reflect that - `A value of 0 disables the fade out animation.` will no longer be true.

kamilkulig: Right, I didn't notice the comment - thanks for pointing that out. On the other hand: is there a difference in current implementation between `hideDelay: 0` & `hideDelay: 1`: https://jsfiddle.net/BlackLabel/xevg84r6/
I can't spot the difference. Or am I missing something?

TorsteinHonsi: The `fadeOut` function takes one argument, `duration`. So I think we should pass a duration of 0 if delay is 0:

`tooltip.getLabel().fadeOut(delay ? void 0 : 0);`

Then we need to fix the `fadeOut` function so it accepts 0, which is a bug in any case:

`duration: pick(duration, 150),`

PR #13034, opened 2020-02-27
bugfix/12982-failing-packedbubble-test

### LINKED ISSUE #12982: Tests: A test seems to be failing at random or is unstable
#### Behaviour
There is an unstable test or a sign of instability of the tests environment shown in this PR as the failed visual test result - https://github.com/highcharts/highcharts/pull/12978

fyi: @activist 

More details:
The test is `series-packedbubble/spiral`.
The difference is in the opacity of a single element. Looks like this could be related to some unfinished animation or, as @pawelfus suggested, related to `defer`.

Izothep:
Fixed #12982, packedBubble test was failing randomly.

___

Test was failing due to a small time difference of hiding overlapping labels. 

Long story short: The reason was that "dataLabels overlapping" check was fired every time new series was rendered. Even though this check is necessary when standard packed bubble series (with simulation) is used, for spiral-packing only one verification is needed.

### DIFF
--- a/js/parts-more/PackedBubbleSeries.js
+++ b/js/parts-more/PackedBubbleSeries.js
@@ -570,7 +570,12 @@ seriesType('packedbubble', 'bubble',
                     });
                 }
             });
-            series.chart.hideOverlappingLabels(dataLabels);
+            // Only hide overlapping dataLabels for layouts that
+            // use simulation. Spiral packedbubble don't need
+            // additional dataLabel hiding on every simulation step
+            if (series.options.useSimulation) {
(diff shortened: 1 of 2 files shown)


Izothep:
Thanks for the feedback! That is definitely true. 

I have considered adding `setAnimation` as well and discussed this approach with @pawelfus, but for now we thought that we will focus on the specific packedbubble issue more and add the `setAnimation` fix somewhere in future if needed. 

It is really great that we were able to fix both problems at once with one simple visual test difference.

Issue #12789, opened 2020-01-15
Annotations not showing up in data table


oysteinmoseng:
#### Expected behaviour
Annotations connected to a point should be displayable in data tables, perhaps with an option in the annotations (and the individual annotation) to turn the behavior on/off.

#### Actual behaviour
Annotations are ignored when building the data table.

#### Live demo with steps to reproduce
https://jsfiddle.net/4Lvc63e9/

KacperMadej:
Is there a detailed design for the feature?

Each annotation can have multiple labels. Each label has a position that is set by x and y values (scaled from set axes) or taken from a point (set by point id) in the chart. It is a similar problem to multiple scatter type series with multiple points with the same x values used - how should the data be sorted by default and what control should be allowed from the public API level?

Should an annotation column be added to the data table after all series? or after each series, if the point from the given series is being annotated?

If there is no detailed specification for this I would like to recommend for the initial feature implementation:
- annotations columns added after series columns,
- add more columns if there are more labels for the same x/category,
- use different annotation only to track order of columns correctly,
- include `includeInDataExport` on each annotation object level and on each label level.

A basic POC for the visual output could be seen under a chart in a demo here: https://jsfiddle.net/BlackLabel/xc1p96qd/

@oysteinmoseng

oysteinmoseng:
@KacperMadej No spec from our side, but I think your proposal is sensible behavior. Alternatively a single column for annotations per series could be used, and the annotations could be concatenated. This would be similar to what we do for screen reader accessibility.

Regardless it will definitely be important that this is behavior that can be turned off/on. I like the idea of supporting the `includeInDataExport` option as well, similar to what we have for series. Having it per label might be overkill, I don't think we support it per point today.

@TorsteinHonsi Thoughts?

TorsteinHonsi:
I don't know... As Kacper indicates, annotations are by nature unstructured. If they were structured, I think in most cases it would be better to use data labels. This also works well with data exporting: http://jsfiddle.net/highcharts/LLExL/14423/ .

oysteinmoseng:
Well, looking at our example demo for CSUN for example, doing something like this with data labels would be a lot more work for the chart creator: https://highcharts.github.io/highcharts-utils/samples/#gh/fa024da5f4/sample/highcharts/accessibility/accessible-annotations

In this case, all but one of the annotations are part of the structured data, and something that would make sense to export/import across systems.

I think we have a few obvious challenges here:
[...]

PR #13035, opened 2020-02-27
enhancement/12789-annotations-in-data-table

madepiet:
The first proof of concept of displaying annotations in the data table (#12789).

> todo list with checkboxes
- [x] columnHeaderFormatter support
- [x] Expansion of polyfill, support for .fill
- [x] New API options
- [x] Tests


ts/annotations/annotations.src.ts
```diff
@@ -1674,6 +1674,73 @@ chartProto.callbacks.push(function (
...
+
+                        newRow.fill('');
+                        newRow.push(annotationText);
+                        newRow.xValues = [];
+                        newRow.xValues[xAxisIndex] = annotationX;
+                        event.dataRows.push(newRow);
+                        wasAdded = true;
+                    }
+
+                    if (!wasAdded) {
```
KacperMadej: Above this line should be a comment about the case being processed/resolved, so we will know what happens in this if.

ts/annotations/annotations.src.ts
```diff
@@ -1674,6 +1674,73 @@ chartProto.callbacks.push(function (
...
+                                xAxisIndex !== void 0 &&
+                                annotationX === row.xValues[xAxisIndex]
+                            ) {
+                                row.push(annotationText);
+                                wasAdded = true;
+                            }
+                        });
+                    }
+
+                    if (!wasAdded) {
```
KacperMadej: Above this line should be a comment about the case being processed/resolved, so we will know what happens in this if.

ts/annotations/annotations.src.ts
```diff
@@ -1674,6 +1674,73 @@ chartProto.callbacks.push(function (
...
+                        const newRow: any = new Array(event.dataRows[0].length);
+
+                        newRow.fill('');
+                        newRow[0] = annotationX;
+                        newRow.push(annotationText);
+                        newRow.xValues = [];
+                        if (xAxisIndex !== void 0) {
```
KacperMadej: Please add a comment above the `if` to inform about the case that is processed/resolved below. When will be `xAxisIndex` undefined?

js/annotations/annotations.src.js
```diff
@@ -1093,6 +1093,59 @@ chartProto.callbacks.push(function (chart) {
...
+    addEvent(chart, 'exportData', function (event) {
+        var annotations = chart.annotations;
+        annotations.forEach(function (annotation) {
+            annotation.labels.forEach(function (label) {
+                var annotationText = label.options.text;
+                label.points.forEach(function (points) {
+                    var annotationX = points.x, xAxisIndex = points.series.xAxis ? points.series.xAxis.options.index : -1;
+                    var wasAdded = false;
+                    if (xAxisIndex === -1) {
+                        var newRow = new Array(event.dataRows[0].length);
```
sebastianbochan: Im curious of using array instead of an object here.

KacperMadej: It is an array to match the existing format coming from export-data module. `newRow` is later pushed into `dataRows` array that stores arrays - rows of table for data export.

sebastianbochan: Now makes sense, thanks!

madepiet:
> todo list with checkboxes
- [ ] columnHeaderFormatter support
- [ ] Expansion of polyfill, support for .fill
- [ ] New API options
- [ ] Tests

madepiet:
> The first proof of concept of displaying annotations in the data table (#12789).

> todo list with checkboxes
- [ ] columnHeaderFormatter support
- [ ] Expansion of polyfill, support for .fill
- [ ] New API options
- [ ] Tests

ts/modules/export-data.src.ts
```diff
@@ -250,6 +251,18 @@ Highcharts.setOptions({
...
+            * When several labels are assigned to a specific point,
+            * they will be displayed in one field in the table.
+            *
+            * @sample highcharts/export-data/concatenate-point-annotations/
+            *         Concatenate point annotations.
+            *
+            * @since   next
+            * @requires modules/annotations
+            */
+            concatenatePointAnnotations: false,
```
sebastianbochan, line 264: IMHO the parameters is too long. 

Maybe linkedPointAnnotations or linkPointAnnotations is more intuitive / user-friendly ?

madepiet: I created it based on this discussion: https://github.com/highcharts/highcharts/issues/12789#issuecomment-585633261
Of course, we can change it. @KacperMadej any thoughts?

pawelfus: How about `joinAnnotations: boolean|string` ? String can be the glue in the concatenated string (right now called `annotationSeparator = '; '`)

pawelfus:
It would be nice to hear some feedback from other devs, I think we have two possible solutions:
```js
csv: {
  joinAnnotations: boolean,
  annotationsSeparator: string
}
```
and:
```js
csv: {
  annotations: {
    join: boolean,
    separator: string
  }
}
```
Perhaps other alternatives?

What do you think? @sebastianbochan @TorsteinHonsi @oysteinmoseng @bre1470 @KacperMadej @Basts 

And what's the difference between `join=true` and `join=false` @madepiet ? Could you show simple comparison? Thanks!

madepiet:
For now, it is working like this:
http://jsfiddle.net/BlackLabel/L1w3qdv6/

```javascript
exporting: {
  csv: {
    joinAnnotations: false // default
    annotationSeparator: string
  }
}
```

Basts:
I am tilting more to the nested option in the `annotations` object because it will be easier to find all available options for exporting annotations than looking for them in the whole API list - they could be far away from each other because of their alphabetical context.

One thought: what do you think about changing the `join` option name to `concat`? For me the `join` option intuitively sounds like an option which decides if the annotation will be attached to the data-label at all.

ts/modules/export-data.src.ts
sebastianbochan: or even `includeAnnotations` or `addAnnotations`?

oysteinmoseng:
Great work so far. Agree that nesting it in an `annotations` object is clearer.

`join` could be more descriptive, it's not very clear what is being joined to what, but I am not sure what would be the best name. `annotations.joinMultiple` might be better? Thoughts @TorsteinHonsi ?

TorsteinHonsi:
I agree that the `exporting.csv.annotations` endpoint is better. I am trying to think of similar options for the `join` stuff, but I can't think of any. So I'm open for suggestions here. Except maybe even a `format` would be more flexible, then we could also add stuff like a bullet point in front of each item.

pawelfus:
> Except maybe even a format would be more flexible, then we could also add stuff like a bullet point in front of each item.

That's a good idea. Probably could be also implemented for other values in exporting (e.g. we have `csv.dateFormat` but we don't have `csv.numberFormat`). No requests for that, though.

> I am trying to think of similar options for the join stuff, but I can't think of any. So I'm open for suggestions here

Just realized we have, directly in CSV options: [`csv.itemDelimiter`](https://api.highcharts.com/highcharts/exporting.csv.itemDelimiter). So analogous `csv.annotations.itemDelimiter` should be the best name, I think.

pawelfus:
I'm curious where should we place logic for adding annotations in data export: annotations.js or export-data.js. I don't have strong opinion here, but export-data sounds like a better place?

madepiet:
When starting work on this, idea was to put the annotations export logic in this annotations.js module, so that it would be used if the user actually needs to use annotations and export these labels. But yes, I agree that here too the opinion of others would be needed.

oysteinmoseng:
I think the convention would dictate to keep the logic in `annotations.js`, but this might also be influenced by the upcoming dashboard product and reworked data layer. @TorsteinHonsi would know best

TorsteinHonsi:
Agree with @oysteinmoseng

ts/mixins/legend-symbol.ts
```diff
@@ -138,12 +138,8 @@ H.LegendSymbolMixin = {
...
         this.legendLine = renderer
             .path([
-                'M',
-                0,
-                verticalCenter,
-                'L',
-                symbolWidth,
-                verticalCenter
+                ['M', 0, verticalCenter],
+                ['L', symbolWidth, verticalCenter]
```
oysteinmoseng, line 142: Is this from a different PR?

ts/Extensions/Annotations/Annotations.ts
```diff
@@ -1732,6 +1745,129 @@ chartProto.callbacks.push(function (
...
+            columnHeaderFormatter = function (index: any): any {
+                let s;
+                if (csvColumnHeaderFormatter) {
+                    s = csvColumnHeaderFormatter(index);
+                    if (s !== false) {
+                        return s;
+                    }
+                }
+
+                s = 'Annotations ' + index;
```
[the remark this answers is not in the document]
oysteinmoseng: Also it seems this column header is not added to the exported CSV

ts/mixins/legend-symbol.ts
madepiet: Unrelated bug in code outside my branch (I added a fix here). The point is that some time ago it happened that paths have to be passed as segments and it is enough to fix the code for something like that otherwise it causes problems.

ts/Extensions/Annotations/Annotations.ts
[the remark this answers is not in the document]
madepiet: Thanks! I corrected it in a new commit.

sebastianbochan:
Looks good !

oysteinmoseng:
@madepiet Want to change the PR to point to the vacation branch? Otherwise Torstein can look at it when he gets back

madepiet:
@oysteinmoseng Done!

sebastianbochan:
Please solve the conflict.

PR #13051, opened 2020-03-02
bugfix/12091-shadow-update

### LINKED ISSUE #12091: series.update shadow does not change shadow without disabling it first
#### Expected behaviour
Shadow should change color with series.update({shadow: { color: 'blue}}).

#### Actual behaviour
series.update({shadow: false}) must be called first and then series.update({shadow: { color: 'blue}}).

https://jsfiddle.net/wnrq1tfy/1/

'update'  button should change color, but does not.
'update with disable' button disables shadow before updating color.

#### Product version
Highcharts 7.2.0

#### Affected browser(s)
Chrome 77

#### Original issue
https://github.com/whawker/react-jsx-highcharts/issues/232

wchmiel:
Fixed #12091, could not update series shadows without disabling them first.

~~Fixed #12091, the issue with update series shadows.~~


ts/parts/SvgRenderer.ts
```diff
@@ -2721,6 +2721,15 @@ extend((
             // compensate for inverted plot area
             transform;
 
+        // Update shadow when options change (#12091).
+        if (shadowOptions) {
+            if (JSON.stringify(this.oldShadowOptions) !== JSON.stringify(shadowOptions)) {
```
pawelfus: `JSON.stringify` is not a safe solution when comparing objects. See demo: https://jsfiddle.net/BlackLabel/3k2j8ht9/

ts/parts/SvgRenderer.ts
```diff
@@ -2717,10 +2717,28 @@ extend((
...
             shadowElementOpacity,
+            update = false,
 
             // compensate for inverted plot area
             transform;
 
+        // Update shadow when options change (#12091).
+        if (isObject(shadowOptions)) {
+            if (this.oldShadowOptions) {
+                Object.keys(this.oldShadowOptions || {}).forEach((key): void => {
```
pawelfus: We have `objectEach` method you can use here.

pawelfus: Extras:
What if `oldShadowOptions = { width: 1, opacity: 1 }` and `newShadowOptions = { offsetX: 10, offsetY: 10 }` ? From code (a few `pick()` calls) it seems that might be the case.

ts/parts/SvgRenderer.ts
```diff
@@ -2717,10 +2717,28 @@ extend((
...
+                        update = true;
+                    }
+                });
+            }
+
+            if (update) {
+                this.destroyShadows();
+            }
+
+            this.oldShadowOptions = shadowOptions;
```
pawelfus: Shouldn't we define `oldShadowOptions` in TS?

wchmiel:
I've improved the fix, please review again @TorsteinHonsi.

PR #13052, opened 2020-03-02
bugfix/13042-getTimezoneOffset-param-type

### LINKED ISSUE #13042: getTimeZoneOffset callback passes inconsistent value type
#### Expected behaviour
`time.getTimezoneOffset` should take a parameter `timestamp` that is milliseconds since January 1 1970 [according to docs](https://api.highcharts.com/highstock/time.getTimezoneOffset)

#### Actual behaviour
Effectively `timestamp` is sometimes the number of milliseconds, but also a `Date()` object. I think it depends on what kind of "thing" is being transformed (label, axis, value, ...).

https://jsfiddle.net/tw6gujfv/ (observe in console)

#### Product version

#### Affected browser(s)
all

bre1470:
Fixed #13042, `getTimeZoneOffset` callback passed inconsistent value type.

### DIFF
--- a/ts/parts/Time.ts
+++ b/ts/parts/Time.ts
@@ -152,7 +152,7 @@ declare global {
  * @callback Highcharts.TimezoneOffsetCallbackFunction
  *
  * @param {number} timestamp
- * Timestamp in terms of milliseconds since January 1 1970.
+ * Date instance or timestamp in terms of milliseconds since January 1 1970.
  *
  * @return {number}
  * Timezone offset in minutes.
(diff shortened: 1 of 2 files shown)


ts/parts/Time.ts
```diff
@@ -152,7 +152,7 @@ declare global {
  * @callback Highcharts.TimezoneOffsetCallbackFunction
  *
  * @param {number} timestamp
- * Timestamp in terms of milliseconds since January 1 1970.
+ * Date instance or timestamp in terms of milliseconds since January 1 1970.
```
mediafreakch: Since you changed the occurences in the code to return `.valueOf()`, is this always a timestamp now?

bre1470: Thanks, good catch! Yes, this doclet change is therefor invalid.

PR #13061, opened 2020-03-03
bugfix/11635-wheelscroll-click-tooltip

### LINKED ISSUE #11635: Scrolling doesn't trigger tooltip hide when cursor leaves the chart's container
#### Expected behaviour
1. consider there are 10 charts in one page from top to bottom.
2. Now scroll the page multiples time top to bottom using mouse wheel
3. All the tool-tips get clustered and do not disappear.

#### Actual behaviour
1. Tool tips should disappear and do not get clustered

#### Product version
    "highcharts": "7.1.2",

#### Affected browser(s)
Chrome

bre1470:
Fixed #11635, scrolling didn't trigger hiding of the tooltip when the cursor left the chart's container.

___

- Close #11635, Tooltip hides with wheelscroll click. https://jsfiddle.net/g643wuLj/
- Improved TestController.

### DIFF
--- a/ts/parts/Interaction.ts
+++ b/ts/parts/Interaction.ts
@@ -291,10 +291,20 @@ TrackerMixin = H.TrackerMixin = {
             pointer = chart.pointer,
             renderer = chart.renderer,
             snap = (chart.options.tooltip as any).snap,
+            tooltip = chart.tooltip,
             tracker = series.tracker,
-            i,
-            onMouseOver = function (): void {
-                if (chart.hoverSeries !== series) {
+            i: number,
+            onMouseOver = function (e: Highcharts.PointerEventObject): void {
+
+                pointer.normalize(e);
+
+                if (
+                    chart.hoverSeries !== series &&
+                    (
+                        !tooltip ||
+                        !tooltip.isStickyOnContact()
+                    )
+                ) {
                     series.onMouseOver();
                 }
             },
@@ -1386,7 +1396,10 @@ extend(Series.prototype, /** @lends Highcharts.Series.prototype */ {
     onMouseOver: function (this: Highcharts.Series): void {
         var series = this,
             chart = series.chart,
-            hoverSeries = chart.hoverSeries;
+            hoverSeries = chart.hoverSeries,
+            pointer = chart.pointer;
+
+        pointer.setHoverChartIndex();
 
         // set normal state to previous series
         if (hoverSeries && hoverSeries !== series) {
--- a/ts/parts/Interaction.ts
+++ b/ts/parts/Interaction.ts
@@ -291,20 +291,10 @@ TrackerMixin = H.TrackerMixin = {
             pointer = chart.pointer,
             renderer = chart.renderer,
             snap = (chart.options.tooltip as any).snap,
-            tooltip = chart.tooltip,
             tracker = series.tracker,
             i: number,
             onMouseOver = function (e: Highcharts.PointerEventObject): void {
-
-                pointer.normalize(e);
-
-                if (
-                    chart.hoverSeries !== series &&
-                    (
-                        !tooltip ||
-                        !tooltip.isStickyOnContact()
-                    )
-                ) {
+                if (chart.hoverSeries !== series) {
                     series.onMouseOver();
                 }
             },
--- a/ts/parts/Pointer.ts
+++ b/ts/parts/Pointer.ts
@@ -90,7 +90,7 @@ declare global {
             public destroy(): void;
             public drag(e: PointerEventObject): void;
             public dragStart(e: PointerEventObject): void;
-            public drop(e: PointerEventObject): void;
+            public drop(e: Event): void;
             public findNearestKDPoint(
                 series: Array<Series>,
                 shared: (boolean|undefined),
@@ -122,15 +122,15 @@ declare global {
                 e: (T|PointerEvent|TouchEvent),
                 chartPosition?: OffsetObject
             ): T;
-            public onContainerClick(e: PointerEventObject): void
-            public onContainerMouseDown(e: PointerEventObject): void;
-            public onContainerMouseLeave(e: PointerEventObject): void;
-            public onContainerMouseMove(e: PointerEventObject): void;
+            public onContainerClick(evt: MouseEvent): void
+            public onContainerMouseDown(evt: MouseEvent): void;
+            public onContainerMouseLeave(e: MouseEvent): void;
+            public onContainerMouseMove(e: MouseEvent): void;
             public onContainerTouchMove(e: PointerEventObject): void;
             public onContainerTouchStart(e: PointerEventObject): void;
             public onDocumentTouchEnd(e: PointerEventObject): void;
-            public onDocumentMouseMove(e: PointerEventObject): void;
-            public onDocumentMouseUp(e: PointerEventObject): void;
+            public onDocumentMouseMove(e: MouseEvent): void;
+            public onDocumentMouseUp(e: MouseEvent): void;
             public onTrackerMouseOut(e: PointerEventObject): void;
             public pinch(e: PointerEventObject): void;
             public pinchTranslate(
@@ -623,18 +623,16 @@ class Pointer {
      * @private
      * @function Highcharts.Pointer#drop
      *
-     * @param {global.Event} e
-     *
-     * @return {void}
+     * @param {global.Event} evt
      */
-    public drop(e: Event): void {
+    public drop(evt: Event): void {
         var pointer = this,
             chart = this.chart,
             hasPinched = this.hasPinched;
 
         if (this.selectionMarker) {
             var selectionData = {
-                    originalEvent: e, // #4890
+                    originalEvent: evt, // #4890
                     xAxis: [],
                     yAxis: []
                 },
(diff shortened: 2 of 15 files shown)


ts/parts/Interaction.ts
```diff
@@ -291,10 +291,20 @@ TrackerMixin = H.TrackerMixin = {
...
+            i: number,
+            onMouseOver = function (e: Highcharts.PointerEventObject): void {
+
+                pointer.normalize(e);
+
+                if (
+                    chart.hoverSeries !== series &&
+                    (
+                        !tooltip ||
+                        !tooltip.isStickyOnContact()
```
TorsteinHonsi: Is this related to #11635, or is it a random issue related to the `stickOnContact` feature? Ideally, this PR should be focused on only the changes required to fix #11635. We can leave it for now, I just want to understand the connection.

ts/parts/Pointer.ts
```diff
@@ -1065,15 +1064,20 @@ class Pointer {
...
         this.isDirectTouch = false;
 
         if (
             series &&
             relatedTarget &&
             !series.stickyTracking &&
+            (
+                !tooltip ||
+                !tooltip.isStickyOnContact()
+            ) &&
```
TorsteinHonsi: As above, are we mixing in some unrelated fix?

ts/parts/Pointer.ts
```diff
@@ -1283,15 +1287,21 @@ class Pointer {
         // Normalize before the 'if' for the legacy IE (#7850)
         e = this.normalize(e);
 
-        if ((e as any).button !== 2) {
+        this.onContainerMouseMove(e);
```
TorsteinHonsi: It seems a little strange to call `onContainerMouseMove()` from the `onContainerMouseDown` function... Why are we doing that? Do we know the side effects? Are there parts of the `onContainerMouseMove` function that we can call directly instead?

At the least, we should have a comment explaining why we do this strange call.

ts/parts/Pointer.ts
```diff
@@ -1283,15 +1287,21 @@ class Pointer {
...
 
-            this.zoomOption(e);
+        if (
+            typeof e.button === 'undefined' ||
+            e.button === 0 // #11635, limiting to primary button
```
TorsteinHonsi: This logic now changed from `e.button !== 2` to `e.button === 0`. If I remember correctly, the primary button in IE is or was 1, hence the !== 2 part. Has this been tested with IE? If so, we should have a comment.

ts/parts/Pointer.ts
```diff
@@ -1283,15 +1287,21 @@ class Pointer {
...
-            if (e.preventDefault) {
+            this.zoomOption(e);
+            /*
+            // #295, dragging not always working in Firefox
+            // #11635, turn off condition because of wheel click issue
+            if (H.isFirefox) {
                 e.preventDefault();
             }
-
+             */
```
TorsteinHonsi: Instead of commenting out obsolete code, we should remove it straight away. If a regression happens, we will find the cause using `git blame`.

ts/parts/Pointer.ts
```diff
@@ -1333,13 +1351,16 @@ class Pointer {
...
 
+        /*
         if (
             !defined(H.hoverChartIndex) ||
             !charts[H.hoverChartIndex as any] ||
             !(charts[H.hoverChartIndex as any] as any).mouseIsDown
         ) {
             H.hoverChartIndex = chart.index;
         }
+         */
```
TorsteinHonsi: As above, obsolete code can safely be deleted rather than commented.

ts/parts/Pointer.ts
```diff
@@ -2106,51 +2127,83 @@ class Pointer {
             ownerDoc = container.ownerDocument;
 
         container.onmousedown = function (e: MouseEvent): void {
-            pointer.onContainerMouseDown(e as any);
+            pointer.onContainerMouseDown(pointer.normalize(e));
```
TorsteinHonsi: `Pointer.onContainerMouseDown` already does `normalize` internally...

ts/parts/Pointer.ts
```diff
@@ -2106,51 +2127,83 @@ class Pointer {
...
         };
         container.onmousemove = function (e: MouseEvent): void {
-            pointer.onContainerMouseMove(e as any);
+            pointer.onContainerMouseMove(pointer.normalize(e));
```
TorsteinHonsi: `Pointer.onContainerMouseMove` already does `normalize` internally...

ts/parts/Pointer.ts
```diff
@@ -2106,51 +2127,83 @@ class Pointer {
...
         };
         container.onclick = function (e: MouseEvent): void {
-            pointer.onContainerClick(e as any);
+            pointer.onContainerClick(pointer.normalize(e));
```
TorsteinHonsi: `Pointer.onContainerClick` already does `normalize` internally...

ts/parts/Pointer.ts
```diff
@@ -2106,51 +2127,83 @@ class Pointer {
...
                 }
             );
             if (!H.unbindDocumentTouchEnd) {
                 H.unbindDocumentTouchEnd = addEvent(
                     ownerDoc,
                     'touchend',
-                    pointer.onDocumentTouchEnd as any
+                    function (e: TouchEvent): void {
+                        pointer.onDocumentTouchEnd(pointer.normalize(e));
+                    }
```
TorsteinHonsi: With all the `addEvent` and `on[XXX]` assignments above, would it be better to just use the `Function.bind` function to bind it to the pointer?

Also, as noted, many of the already normalize the event arguments, should we do that with the rest where it is needed, so we can have simple assignments of bound functions?

ts/parts/Pointer.ts
```diff
@@ -2106,51 +2127,83 @@ class Pointer {
...
+    /**
+     * Sets the index of the hovered chart and leaves the previous hovered
+     * chart, to reset states like tooltip.
+     *
+     * @private
+     * @function Highcharts.Pointer#setHoverChartIndex
+     */
+    public setHoverChartIndex(): void {
+        const chart = this.chart;
+        const hoverChart = H.charts[H.hoverChartIndex || -1];
```
TorsteinHonsi: Should be:
`const hoverChart = H.charts[pick(H.hoverChartIndex, -1)];`

... to prevent the 0 case failure.

ts/parts/Pointer.ts
```diff
@@ -2106,51 +2127,83 @@ class Pointer {
...
+
+        if (
+            hoverChart &&
+            hoverChart.index !== chart.index
```
TorsteinHonsi: I think we can safely do `if (hoverChart && hoverChart !== chart)` ?

ts/parts/Pointer.ts
```diff
@@ -2169,10 +2222,13 @@ class Pointer {
             pinchDown,
             isInside;
 
+        /*
         if (chart.index !== H.hoverChartIndex) {
             this.onContainerMouseLeave({ relatedTarget: true } as any);
         }
         H.hoverChartIndex = chart.index;
+         */
```
TorsteinHonsi: Remove the block instead of commenting.

ts/parts/Pointer.ts
bre1470: The event order between Firefox and Chrome after click scrolling is different. Firefox does not fire reliable a move event after releasing the click wheel. Chrome fires always down and click.

ts/parts/Pointer.ts
bre1470: Just to make sure, I understand the issue, the original condition was already flawed, correct? So we need additionally some IE specific code here?

ts/parts/Pointer.ts
bre1470: That is a type thing, because I liked to remove the `any` casts. Without normalize `onContainerMouseDown` has to accept regular `MouseEvent`s without normalizing. It is the same with the other pointer callbacks.

ts/parts/Interaction.ts
bre1470: This comes from an old merge during the first try of solving the issue.

ts/parts/Pointer.ts
bre1470: Same here.

bre1470:
Please review again.

ts/parts/Pointer.ts
```diff
@@ -623,18 +623,16 @@ class Pointer {
      * @private
      * @function Highcharts.Pointer#drop
      *
-     * @param {global.Event} e
-     *
-     * @return {void}
+     * @param {global.Event} evt
      */
-    public drop(e: Event): void {
+    public drop(evt: Event): void {
```
TorsteinHonsi: Why was this renamed to `evt`? In all other aspects, we use `e` for event arguments, so we should keep consistent whenever we can.

ts/parts/Pointer.ts
```diff
@@ -1312,18 +1306,19 @@ class Pointer {
...
      *
-     * @param {Highcharts.PointerEventObject} e
+     * @param {global.MouseEvent} evt
      *
      * @return {void}
      */
-    public onContainerMouseLeave(e: Highcharts.PointerEventObject): void {
+    public onContainerMouseLeave(evt: MouseEvent): void {
         const chart = charts[pick(H.hoverChartIndex, -1)];
+        const pEvt = this.normalize(evt);
```
TorsteinHonsi: Do we need to create a new variable for the normalized event arguments?

Issue #13097, opened 2020-03-06
Tools: prep-release fails


TorsteinHonsi:
#### Expected behaviour
`gulp prep-release` should prepare for release by updating the version in files.

#### Actual behaviour
Fails with JS error

#### Live demo with steps to reproduce
In the repo root, run `gulp prep-release --nextversion x.y.z`

Also test `npx gulp prep-release --cleanup --commit` 

#### Product version
v8.0.3

#### Affected browser(s)
None, this is a tooling issue. Fails on the Mac. Possibly works on Linux, where it was authored.

A suggested debugging path is to use GitHub blame to identify when the function was first introduced, then test if it ever worked on the Mac. If it did, we can bisect to find the regression.

goransle:
@TorsteinHonsi [this commit](https://github.com/highcharts/highcharts/commit/5e221800f5aed3185b3518aedb39bd0bc3db9bf3) seems to be the issue. OSX does not support the 'tmp/*.bak' syntax for the `-i` argument.

Simplest fix may be to use `sed -i'.bak'` and add `.bak` files to `.gitignore`.

It also seems that the `grep` script for setting next version in docs does not work not work on OSX.

PR #13102, opened 2020-03-06
Tools: Fix for #13097 prep-release failing in OSX

goransle:
* Fixes issues with the `prep-release` gulp task on OSX. #13097
* It now stores the backup files at the same location as the original (not in `/tmp`)
* Added `.bak` files to gitignore


goransle:
> The current versions declared in files package.json, bower.json and build.properties does not match!

This is probably because the version in `build.properties` has `-modified` appended after cleanup. Perhaps this should be ignored in the version comparison?

It does not fail to finish for me. I found some inconsistent use of quotation marks that may have caused issues (but it does not seem to be an issue on Catalina).

TorsteinHonsi:
> This is probably because the version in build.properties has -modified appended after cleanup. Perhaps this should be ignored in the version comparison?

Yes, I think `-modified` should be stripped out.

goransle:
Confirmed to be working in Ubuntu 18.04.

PR #13111, opened 2020-03-10
Fixed #13110, chart.sonify no longer working.

### LINKED ISSUE #13110: Sonification demos no longer working with v8.0.1
Because of the new sonification options, where sonification by default is `enabled: false`, none of the sonification demos using `chart.sonify` are working anymore.

Example: https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/demo/sonification/

We should probably change the behavior of `chart.sonify()` to sonify the charts regardless of whether or not the option is enabled on the chart - and not call this function internally if the option is `false`.

Workaround for now is either rolling back to v8.0.0 of the sonification module, or setting the `sonification.enabled` option to true for the chart.

oysteinmoseng:


oysteinmoseng:
Closing, try in new PR. Something gone wrong with untracked files.

PR #13119, opened 2020-03-11
enhancement/stacking-class-es6

marvin19:
Tools: Migrated Stacking class to ES6 pattern

### DIFF
--- a/ts/parts/Stacking.ts
+++ b/ts/parts/Stacking.ts
@@ -213,66 +213,80 @@ var Axis = H.Axis,
  * @param {number} x
  * @param {Highcharts.OptionsStackingValue} [stackOption]
  */
-H.StackItem = function (
-    this: Highcharts.StackItem,
-    axis: Highcharts.Axis,
-    options: Highcharts.YAxisStackLabelsOptions,
-    isNegative: boolean,
-    x: number,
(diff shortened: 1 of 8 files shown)


ts/parts/Stacking.ts
```diff
@@ -213,66 +213,80 @@ var Axis = H.Axis,
...
+    public constructor(
+        axis: Highcharts.Axis,
+        options: Highcharts.YAxisStackLabelsOptions,
+        isNegative: boolean,
+        x: number,
+        stackOption?: Highcharts.OptionsStackingValue
+    ) {
 
-    var inverted = axis.chart.inverted;
+        var inverted = axis.chart.inverted;
```
bre1470: This should become a constant.

ts/parts/Stacking.ts
```diff
@@ -947,3 +956,6 @@ Series.prototype.getStackIndicator = function (
 
     return stackIndicator as any;
```
bre1470: We should try to remove the `any` casts. Maybe the key string can be narrowed down to string literal types?

`stackIndicator` should be type safe, too.

bre1470: I am referring to all the code in the function. Somehow GitHub forgot the marking. 🤷‍♀️

PR #13127, opened 2020-03-12
Fixed #13001, scrollablePlotArea not inheriting chart style

### LINKED ISSUE #13001: Chart style not work in scrollablePlotArea
#### Expected behaviour
```
chart :  {
        className: 'relative-pos',
        style   : {
            fontFamily: 'Roboto',
        },
        scrollablePlotArea: {
            minWidth: 800
        },
},
```
Chart style will apply styles, class to all fixed and scrolling container (**highcharts-fixed and highcharts-scrolling**).

In my case, it's should apply **fontFamily: 'Roboto' to "highcharts-fixed"** too.

#### Actual behaviour
Chart style only applies styles to **highcharts-scrolling**. 

http://jsfiddle.net/sirhvd/7xdsg6ez/

#### Product version
Highcharts 8.0.0

#### Affected browser(s)
All

goransle:
Fixed #13001, `scrollablePlotArea` not inheriting chart style. Also applied chart styles to tooltip container when `tooltip.outside` is `true`.

~~Note that the style options does not apply to the tooltip unless `tooltip.outside` is set to `false` (defaults to `true` with scrollablePlotArea). If this is not considered an issue it should be noted in the API.~~

### DIFF
--- a/ts/parts/Tooltip.ts
+++ b/ts/parts/Tooltip.ts
@@ -728,7 +728,7 @@ class Tooltip {
                     container,
                     0,
                     0,
-                    {},
+                    { ...this.chart.options.chart?.style },
                     void 0,
                     void 0,
                     renderer.styledMode
--- a/ts/parts/Tooltip.ts
+++ b/ts/parts/Tooltip.ts
@@ -728,7 +728,7 @@ class Tooltip {
                     container,
                     0,
                     0,
-                    { ...this.chart.options.chart?.style },
+                    this.chart.options.chart?.style,
                     void 0,
                     void 0,
                     renderer.styledMode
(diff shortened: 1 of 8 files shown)


pawelfus:
Please check guidelines for PR and changelog: https://github.com/highcharts/highcharts/tree/master/changelog#writing-content-for-the-changelog (missing changelog label and first paragraph for changelog is missing).

ts/parts/Tooltip.ts
```diff
@@ -728,7 +728,7 @@ class Tooltip {
                     container,
                     0,
                     0,
-                    {},
+                    { ...this.chart.options.chart?.style },
```
pawelfus: I think we don't need this - simple `this.chart.options.chart?.style` should be enough? And we won't create extra polyfill.

goransle: You're right, fix incoming.

PR #13138, opened 2020-03-13
bugfix/12883-accessibility-pagination-not-working

### LINKED ISSUE #12883: Legend pagination arrow buttons sometimes don't work
#### Expected behaviour
Clicking on the pagination arrow buttons should move the the next/previous legend page

#### Actual behaviour
There are cases where we have a lot of series in the legend, and when clicking on the arrow down, instead of going to the next page, the chart acts as I clicked on one of the legend series so I can't navigate to the next page.

I contacted the support center and recieved the following response:
"It looks like the accessibility module is responsible for that behavior. When I inspected the area where clicking results in switching series' visibility, I found out that there is a hidden button that belongs to a div element with '**highcharts-a11y-proxy-container**' class which indicates the accessibility module. Without it, it seems everything works correctly. If you want, you can report it as an issue here: https://github.com/highcharts/highcharts/issues.
​
Example:
https://jsfiddle.net/BlackLabel/9x8jsuom/​

https://jsfiddle.net/stco3nLp/1/

This isn't always reproduced. In my nachine, it's reproduces all the time. I see in the legend 6 pages instead of 8. So my guess is that there is another hidden line in each legend page, and when clicking on the arrow, the click is caoght by the series button instead of the arrow.

#### Product version
7.0.0

#### Affected browser(s)
Chrome, Edge

Izothep:
Fixed #12883, legend pagination was not working correctly when the accessibility module was included.

### DIFF
--- a/js/modules/accessibility/components/LegendComponent.js
+++ b/js/modules/accessibility/components/LegendComponent.js
@@ -93,9 +93,9 @@ extend(LegendComponent.prototype, /** @lends Highcharts.LegendComponent */ {
      * @private
      */
     updateLegendItemProxyVisibility: function () {
-        var legend = this.chart.legend, items = legend.allItems || [], curPage = legend.currentPage || 1;
+        var legend = this.chart.legend, items = legend.allItems || [], curPage = legend.currentPage || 1, clipHeight = legend.clipHeight || 0;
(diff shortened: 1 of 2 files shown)


oysteinmoseng:
Looks good to me!

PR #13148, opened 2020-03-16
enhancement/data-class-es6

marvin19:
Tools: Migrated Data class to ES6 pattern

### DIFF
--- a/ts/modules/data.src.ts
+++ b/ts/modules/data.src.ts
@@ -258,6 +258,7 @@ const {
 
 import '../parts/Chart.js';
 import '../mixins/ajax.js';
+import H from '../parts/Globals.js';
 
 // Utilities
 var addEvent = Highcharts.addEvent,
@@ -265,8 +266,7 @@ var addEvent = Highcharts.addEvent,
     win = Highcharts.win,
     doc = win.document,
     merge = Highcharts.merge,
-    fireEvent = Highcharts.fireEvent,
-    SeriesBuilder: typeof Highcharts.SeriesBuilder;
+    fireEvent = Highcharts.fireEvent;
 
 /**
  * The Data module provides a simplified interface for adding data to
@@ -675,17 +675,26 @@ var addEvent = Highcharts.addEvent,
  *
  * @param {Highcharts.Chart} [chart]
  */
-var Data = function (
-    this: Highcharts.Data,
-    dataOptions: Highcharts.DataOptions,
-    chartOptions?: Highcharts.Options,
-    chart?: Highcharts.Chart
-): void {
-    this.init(dataOptions, chartOptions, chart);
-};
+class Data {
+    public constructor(
+        dataOptions: Highcharts.DataOptions,
+        chartOptions?: Highcharts.Options,
+        chart?: Highcharts.Chart
+    ) {
+        this.init(dataOptions, chartOptions, chart);
+    }
 
-// Set the prototype properties
-extend(Data.prototype, {
+    public alternativeFormat?: string;
+    public chart: Highcharts.Chart = void 0 as any;
+    public chartOptions: Highcharts.Options = void 0 as any;
+    public columns?: Array<Array<Highcharts.DataValueType>>;
+    public dateFormat?: string;
+    public decimalRegex?: RegExp;
+    public firstRowAsNames: boolean = void 0 as any;
+    public liveDataTimeout?: number;
+    public rawColumns: Array<Array<string>> = void 0 as any;
+    public options: Highcharts.DataOptions= void 0 as any;
+    public valueCount?: Highcharts.DataValueCountObject;
 
     /**
      * Initialize the Data object with the given options
@@ -696,8 +705,7 @@ extend(Data.prototype, {
      * @param {Highcharts.Options} [chartOptions]
      * @param {Highcharts.Chart} [chart]
      */
-    init: function (
-        this: Highcharts.Data,
+    public init(
         options: Highcharts.DataOptions,
         chartOptions?: Highcharts.Options,
         chart?: Highcharts.Chart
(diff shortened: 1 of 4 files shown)


ts/modules/data.src.ts
```diff
@@ -258,15 +258,15 @@ const {
...
 
 // Utilities
 var addEvent = Highcharts.addEvent,
     Chart = Highcharts.Chart,
     win = Highcharts.win,
     doc = win.document,
     merge = Highcharts.merge,
-    fireEvent = Highcharts.fireEvent,
-    SeriesBuilder: typeof Highcharts.SeriesBuilder;
+    fireEvent = Highcharts.fireEvent;
```
bre1470: Utilities function should be used via import. Also import statements should be sorted, so that named imports come first and anonymous come last.

ts/modules/data.src.ts
```diff
@@ -720,7 +728,7 @@ extend(Data.prototype, {
         this.options = options;
         this.columns = (
             options.columns ||
-            this.rowsToColumns(options.rows) ||
+            this.rowsToColumns((options.rows as any)) ||
```
bre1470: `any` should be worked out.

ts/modules/data.src.ts
```diff
@@ -742,7 +750,7 @@ extend(Data.prototype, {
         this.rawColumns = [];
 
         // No need to parse or interpret anything
-        if (this.columns.length) {
+        if ((this.columns as any).length) {
```
bre1470: `any` should be worked out.

ts/modules/data.src.ts
```diff
@@ -2641,195 +2648,190 @@ addEvent(
...
+            point =
+                pointIsArray ? [] as Array<T> : {} as Highcharts.Dictionary<T>,
+            columnIndexes;
+
+        // Loop each reader and ask it to read its value.
+        // Then, build an array or point based on the readers names.
+        builder.readers.forEach(function (
+            reader: Highcharts.SeriesBuilderReaderObject
+        ): void {
+            var value = columns[reader.columnIndex as any][rowIndex];
```
bre1470: `any` should be worked out.

ts/modules/data.src.ts
```diff
@@ -2641,195 +2648,190 @@ addEvent(
...
+            enoughColumns = true;
+
+        // Loop each reader and give it an index if its missing.
+        // The freeIndexes.shift() will return undefined if there
+        // are no more columns.
+        builder.readers.forEach(function (
+            reader: Highcharts.SeriesBuilderReaderObject
+        ): void {
+            if (typeof reader.columnIndex === 'undefined') {
+                reader.columnIndex = freeIndexes.shift() as any;
```
bre1470: `any` should be worked out.

ts/modules/data.src.ts
```diff
@@ -2641,195 +2648,190 @@ addEvent(
...
+
+                // Sort the remaining
+                columnIndexes.sort(function (a: number, b: number): number {
+                    return a - b;
+                });
 
-            // Now use the lowest index as name column
-            this.name = (columns[columnIndexes.shift() as any] as any).name;
+                // Now use the lowest index as name column
+                this.name = (columns[columnIndexes.shift() as any] as any).name;
```
bre1470: `any` should be worked out.

ts/modules/data.src.ts
bre1470: Not new. Ignoring.

ts/modules/data.src.ts
bre1470: Not new. Ignoring.

ts/modules/data.src.ts
bre1470: Not new. Ignoring.

ts/modules/data.src.ts
```diff
@@ -248,25 +248,25 @@ declare global {
...
-import '../parts/Chart.js';
+import H from '../parts/Globals.js';
 import '../mixins/ajax.js';
+import '../parts/Chart.js';
 
 // Utilities
-var addEvent = Highcharts.addEvent,
-    Chart = Highcharts.Chart,
+var Chart = Highcharts.Chart,
     win = Highcharts.win,
```
bre1470: These two have to use the imported `H.` instead of `Highcharts.`. Maybe look for similar cases in the file, where `Highcharts.` is not used as a type but as an object.

ts/modules/data.src.ts
```diff
@@ -2203,26 +2216,24 @@ extend(Data.prototype, {
...
     /**
      * Reorganize rows into columns.
      *
      * @function Highcharts.Data#rowsToColumns
      *
-     * @param {Array<Array<(number|string)>>} rows
+     * @param {Array<Array<Highcharts.DataValueType>>} rows
      *
-     * @return {Array<Array<(number|string)>>}
+     * @return {Array<Array<Highcharts.DataValueType>>}
```
bre1470: The return type is not in sync with the function header.

ts/modules/data.src.ts
```diff
@@ -2579,7 +2586,7 @@ addEvent(
              * @name Highcharts.Chart#data
              * @type {Highcharts.Data|undefined}
              */
-            chart.data = new (Data as any)(extend(userOptions.data, {
+            chart.data = new (H.Data as any)(extend(userOptions.data, {
```
bre1470: `new (H.Data as any)(…` -> `new H.Data(…`

ts/modules/data.src.ts
```diff
@@ -2641,195 +2648,190 @@ addEvent(
...
+        ): void {
+            var value = columns[reader.columnIndex as any][rowIndex];
+
+            if (pointIsArray) {
+                (point as any).push(value);
             } else {
-                (point as any)[reader.configName] = value;
+                if (reader.configName.indexOf('.') > 0) {
+                    // Handle nested property names
+                    Highcharts.Point.prototype.setNestedProperty(
```
bre1470: `Highcharts.Point` -> `Point` + `import Point from…` at the top.

PR #13151, opened 2020-03-16
bugfix/12027-packedbubble-print

### LINKED ISSUE #12027: Printing chart is all messed up for packed bubble chart
#### Expected behaviour
The chart is expected to be printed as in jpeg.
carbon-emissions-around (2)

#### Actual behaviour
Since **enableSimulation** is set **true** for UI, the chart gets printed while in motion.
image

I tried to set **enableSimulation** to false in beforeprint. But I could not disable the simulation when **Print chart** is clicked.

Izothep:
Fixed #12027, print was not working correctly in packed bubble charts.

___

Added update method for layouts in both packedbubble and networkgraph series

### DIFF
--- a/ts/modules/networkgraph/layouts.ts
+++ b/ts/modules/networkgraph/layouts.ts
@@ -120,7 +120,7 @@ declare global {
             public init(options: NetworkgraphLayoutAlgorithmOptions): void;
             public initPositions(): void;
             public isStable(): boolean;
-            public updateSimulation(force: boolean, enable?: boolean): void;
+            public updateSimulation(enable?: boolean): void;
             public removeElementFromCollection<T>(
                 element: T, collection: Array<T>
             ): void
@@ -211,22 +211,11 @@ extend(
 
             this.approximation = options.approximation;
         },
-        update: function (
-            this: Highcharts.NetworkgraphLayout,
-            options: Highcharts.NetworkgraphLayoutAlgorithmOptions,
-            redraw?: boolean
-        ): void {
-            this.options = merge(this.options, options);
-            if (pick(redraw, true) && this.chart) {
-                this.chart.redraw();
-            }
-        },
         updateSimulation: function (
             this: Highcharts.NetworkgraphLayout,
-            forced: boolean,
             enable?: boolean
         ): void {
-            this.enableSimulation = forced ? enable : this.options.enableSimulation;
+            this.enableSimulation = pick(enable, this.options.enableSimulation);
         },
         start: function (this: Highcharts.NetworkgraphLayout): void {
             var layout = this,
@@ -937,4 +926,25 @@ addEvent(Chart as any, 'render', function (
             });
         }
     }
+});
+
+// disable simulation before print if enabled
+addEvent(Chart as any, 'beforePrint', function (
+    this: Highcharts.PackedBubbleChart
+): void {
+    this.graphLayoutsLookup.forEach(function (layout): void {
+        layout.updateSimulation(false);
+    });
+    this.redraw();
+});
+
+// re-enable simulation after print
+addEvent(Chart as any, 'afterPrint', function (
+    this: Highcharts.PackedBubbleChart
+): void {
+    this.graphLayoutsLookup.forEach(function (layout): void {
+        // return to default simulation
+        layout.updateSimulation();
+    });
+    this.redraw();
 });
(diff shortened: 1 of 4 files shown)


ts/parts-more/PackedBubbleSeries.ts
```diff
@@ -1891,6 +1875,27 @@ addEvent(Chart as any, 'beforeRedraw', function (
...
+
+// disable simulation before print if enabled
+addEvent(Chart as any, 'beforePrint', function (
+    this: Highcharts.PackedBubbleChart
+): void {
+    this.graphLayoutsLookup.forEach(function (layout): void {
+        layout.updateSimulation(true, false);
+    });
+    this.redraw();
+});
```
pawelfus: Shouldn't this be defined in Layout? Does it fix networkgrap issue when packed bubble series is not included?

ts/modules/networkgraph/layouts.ts
```diff
@@ -206,6 +211,23 @@ extend(
...
         },
+        update: function (
+            this: Highcharts.NetworkgraphLayout,
+            options: Highcharts.NetworkgraphLayoutAlgorithmOptions,
+            redraw?: boolean
+        ): void {
+            this.options = merge(this.options, options);
+            if (pick(redraw, true) && this.chart) {
+                this.chart.redraw();
+            }
```
pawelfus: I don't see any use of this method - is method necessary?
Also, do we need `chart.redraw()` here? Layout up to now was independent from the chart instance.

ts/modules/networkgraph/layouts.ts
```diff
@@ -206,6 +211,23 @@ extend(
...
+        },
+        updateSimulation: function (
+            this: Highcharts.NetworkgraphLayout,
+            forced: boolean,
+            enable?: boolean
+        ): void {
```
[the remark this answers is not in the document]
Izothep: Currently, `updateSimulation(true)` is setting the simulation to false (`undefined` to be exact).

Yep, I had some doubts as well, currently it uses:
`this.enableSimulation = forced ? enable : this.options.enableSimulation;`

So what should happen in case when `updateSimulation(false, true)` is used? Maybe instead of `forced` it should be more like:
`this.enableSimulation = useDefaultVal ? this.options.enableSimulation : enable;`

ts/parts-more/PackedBubbleSeries.ts
Izothep: I will move it to Layout, then Networkgraph print will be fixed as well.

ts/modules/networkgraph/layouts.ts
Izothep: yeah, Right now the method is not used anywhere but in general it works so it is possible to use it. I think we can keep it as it is not too long method. `chart.redraw` might be called outside of the function so it is not needed to be in here.

ts/modules/networkgraph/layouts.ts
```diff
@@ -194,6 +197,8 @@ extend(
             this.integration =
                 H.networkgraphIntegrations[options.integration as any];
 
+            this.enableSimulation = options.enableSimulation;
```
pawelfus: Missing TS definition ;)

ts/modules/networkgraph/layouts.ts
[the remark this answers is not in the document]
pawelfus: Ah I see, so we can say it's resetting flag to the default value. Yes, it makes sense.

ts/modules/networkgraph/layouts.ts
pawelfus: I'm not sure if this generally works - for example updating max number of iterations, or force-functions. Simulation must be reset in order to apply all changes. It seems that only `options.enableSimulation` and `options.theta` options were directly used - so updating `this.options` affects them :)

ts/modules/networkgraph/layouts.ts
Izothep: TS definition was created before (not in this specific commit) as it was used in draggable-nodes

ts/modules/networkgraph/layouts.ts
Izothep: we can use `layout.init` for updating all the other layout parameters. However, after deeper investigation I believe we don't need this method for now. I will add it to the code when necessary.

Izothep:
I believe we can consider the PR as final.

PR #13155, opened 2020-03-17
enhancement/drag-panes-class-es6

marvin19:
Tools: Migrated AxisResizer class to ES6 pattern

### DIFF
--- a/ts/modules/drag-panes.src.ts
+++ b/ts/modules/drag-panes.src.ts
@@ -675,9 +675,6 @@ class AxisResizer {
     }
 }
 
-H.AxisResizer = AxisResizer as any;
-export default H.AxisResizer;
-
 // Keep resizer reference on axis update
 Axis.prototype.keepProps.push('resizer');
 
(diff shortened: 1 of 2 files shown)


ts/modules/drag-panes.src.ts
```diff
@@ -250,32 +264,26 @@ var hasTouch = H.hasTouch,
              */
             y: 0
         }
-    };
-
-merge(true, Axis.prototype.defaultYAxisOptions, resizerOptions);
```
bre1470: That is missing. You should add something like `merge(true, Axis.defaultYAxisOptions, AxisResizer.resizerOptions);` at the end of the file.

ts/modules/drag-panes.src.ts
```diff
@@ -675,13 +673,15 @@ H.AxisResizer.prototype = {
             (resizer as any)[key] = null;
         });
     }
-} as any;
+}
+
+H.AxisResizer = AxisResizer as any;
+export default H.AxisResizer;
```
bre1470: Exports should go to the file end.

PR #13162, opened 2020-03-18
Refactor Series.getExtremes

TorsteinHonsi:
Refactored `Series.getExtremes` so that it now only returns values instead of setting them on the series instance. It makes it easier to override, easier to use in different contexts, and more consistent with the `get` naming.

### DIFF
--- a/ts/modules/bullet.src.ts
+++ b/ts/modules/bullet.src.ts
@@ -317,13 +317,13 @@ seriesType<Highcharts.BulletSeries>('bullet', 'column'
                     this,
                     targetData
                 );
-                if (typeof targetExtremes.dataMin === 'number') {
+                if (isNumber(targetExtremes.dataMin)) {
                     dataExtremes.dataMin = Math.min(
                         pick(dataExtremes.dataMin, Infinity),
                         targetExtremes.dataMin
                     );
                 }
-                if (typeof targetExtremes.dataMax === 'number') {
+                if (isNumber(targetExtremes.dataMax)) {
                     dataExtremes.dataMax = Math.max(
                         pick(dataExtremes.dataMax, -Infinity),
                         targetExtremes.dataMax
--- a/ts/modules/bullet.src.ts
+++ b/ts/modules/bullet.src.ts
@@ -35,7 +35,7 @@ declare global {
             public points: Array<BulletPoint>;
             public targetData: Array<number>;
             public drawPoints(): void;
-            public getExtremes(yData?: Array<number>): void;
+            public getExtremes(yData?: Array<number>): DataExtremesObject;
         }
         interface BulletPointOptions extends ColumnPointOptions {
             borderColor?: ColorType;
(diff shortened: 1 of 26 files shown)


ts/modules/bullet.src.ts
```diff
@@ -304,21 +304,33 @@ seriesType<Highcharts.BulletSeries>('bullet', 'column'
...
-                yMax = series.dataMax;
-                yMin = series.dataMin;
-                columnProto.getExtremes.call(this, targetData);
-                series.dataMax = Math.max(series.dataMax, yMax);
-                series.dataMin = Math.min(series.dataMin, yMin);
+                const targetExtremes = columnProto.getExtremes.call(
+                    this,
+                    targetData
+                );
+                if (typeof targetExtremes.dataMin === 'number') {
```
pawelfus: Wouldn't be better to use `isNumber()`? We get for free checking for `NaN` and if value is finite.

bre1470: I agree, but I wonder why it is optional in the first place. Shouldn't DataExtremesObject always provide these numbers?

ts/parts/ColumnSeries.ts
```diff
@@ -836,7 +836,7 @@ seriesType<Highcharts.ColumnSeries>(
                     // in visible range (#7046)
                     if (
                         point.y === threshold &&
-                        series.dataMax <= (threshold as any) &&
+                        (series.dataMax as any) <= (threshold as any) &&
```
bre1470: I doubt that this new `any` cast is necessary. If it is optional, just put a `series.dataMax &&` before, but it is actually not optional.

ts/parts/Series.ts
```diff
@@ -218,6 +219,10 @@ declare global {
         interface Chart {
             runTrackerClick?: boolean;
         }
+        interface DataExtremesObject {
+            dataMin?: number;
+            dataMax?: number;
```
bre1470: Why are the properties optional? Also consider use of `RangeObject` interface with `min` / `max`.

ts/modules/sonification/utilities.ts
```diff
@@ -241,13 +241,13 @@ var utilities: Highcharts.SonificationUtilitiesObject = {
      */
     virtualAxisTranslate: function (
         value: number,
-        dataExtremes: Highcharts.RangeObject,
+        DataExtremesObject: Highcharts.RangeObject,
```
bre1470: This looks like a refactoring problem.

ts/parts/Series.ts
TorsteinHonsi: In some cases, like when boosting is active, or in general for the windbarb series, there are no data extremes. So I thought it was better to return an object with undefined keys than to return undefined.

> Also consider use of RangeObject interface with min / max.

Yes, we could use that. Then we would have to change the above to check for existence of the whole object. Any preferences @pawelfus , @bre1470 ?

ts/parts/ColumnSeries.ts
TorsteinHonsi: Lazyness I guess. But we need to typecheck, as it can be 0.

ts/modules/bullet.src.ts
TorsteinHonsi: They can be undefined, see comment below.

ts/parts/Series.ts
pawelfus: I'm fine with all solutions (return: undefined, object without keys or object with keys that values are undefined):

- `undefined` is the easiest to work with (instead of checking `isNumber(dataExtremes.dataMax)` check if `dataExtremes` are defined), but at the same time it's a little strange that `getExtremes()` returns nothing
- `{ dataMin: undefined|number, dataMax: undefined|number }` is kind of standarized output from this function I would expect, assuming that dataMin/Max can be undefined eg when data is empty

TorsteinHonsi: Yes I agree to the second point (or both actually). But the second point is the reason why I did it this way, given that `series.dataMin` can also be undefined.

PR #13164, opened 2020-03-18
bugfix/12979-annotations-update-errors

### LINKED ISSUE #12979: Annotation module: update causing errors in the browser console
#### Behaviour
If an axis related to an annotation is updated at the same time as the annotation then there is an error in browser console like:

```red
Error: <rect> attribute x: Expected length, "undefined".
30 Error: <rect> attribute x: Expected length, "undefinedNaN".
30 Error: <rect> attribute width: Expected length, "undefinedNaNNaN".
```

Details are rather complex. The annotation needs to have a label that will have a point that will have set the axis that is being updated. The update needs to be done in a single chart.update or could be done for the axis first as long as it will be done without a redraw.

Debugging leads to the `getClipBox` function of the Annotation in the annotation module. Relevant axis does not have top, left, width nor height properties set, so the error is later produced.
https://github.com/highcharts/highcharts/blob/5d7a807ace923cabcbe2f896b437d7a356e577d8/ts/annotations/annotations.src.ts#L1144-L1151

As a workaround, you could update the axis with a redraw before updating the annotation or you could use a simple fix as seen in the demo below:
https://jsfiddle.net/BlackLabel/c0jx2egt/12/

This is a demo with the problem visible (and workarounds commented out):
https://jsfiddle.net/BlackLabel/c0jx2egt/11/

#### Product version
8.0.0

#### Affected browser(s)
Tested only in Chrome

madepiet:
Fixed #12979, console errors when updating an axis at the same time as updating a related annotation.

~~Fixed #12979, console errors after the update.~~

### DIFF
--- a/ts/annotations/annotations.src.ts
+++ b/ts/annotations/annotations.src.ts
@@ -1143,10 +1143,10 @@ merge(
 
         getClipBox: function (this: Highcharts.Annotation): Highcharts.BBoxObject {
             return {
-                x: (this.clipXAxis as any).left,
-                y: (this.clipYAxis as any).top,
-                width: (this.clipXAxis as any).width,
-                height: (this.clipYAxis as any).height
+                x: (this.clipXAxis as any).left || 0,
+                y: (this.clipYAxis as any).top || 0,
+                width: (this.clipXAxis as any).width || 0,
+                height: (this.clipYAxis as any).height || 0
             };
         },
 
--- a/ts/annotations/annotations.src.ts
+++ b/ts/annotations/annotations.src.ts
@@ -95,7 +95,10 @@ declare global {
             public setLabelCollector(): void;
             public setOptions(userOptions: AnnotationsOptions): void;
             public setVisibility(visible?: boolean): void;
-            public update(userOptions: DeepPartial<AnnotationsOptions>): void;
+            public update(
+                userOptions: DeepPartial<AnnotationsOptions>,
+                redraw?: boolean
+            ): void;
         }
         interface AnnotationChart extends Chart {
             annotations: Array<Annotation>;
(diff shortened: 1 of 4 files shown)


ts/annotations/annotations.src.ts
```diff
@@ -1143,10 +1143,10 @@ merge(
...
         getClipBox: function (this: Highcharts.Annotation): Highcharts.BBoxObject {
             return {
-                x: (this.clipXAxis as any).left,
-                y: (this.clipYAxis as any).top,
-                width: (this.clipXAxis as any).width,
-                height: (this.clipYAxis as any).height
+                x: (this.clipXAxis as any).left || 0,
+                y: (this.clipYAxis as any).top || 0,
+                width: (this.clipXAxis as any).width || 0,
+                height: (this.clipYAxis as any).height || 0
```
TorsteinHonsi: This is not a mistake, but I think we should take the opportunity to fix the typing in this block.

1. First, according to TypeScript, `this.clipXAxis` can be undefined. So we should check for that, or alternatively find out why it is typed that way if it is in practice always defined.
2. Next, the `y`, `top`, `width` and `height` properties of the Axis type are `number`. But here we have a case where they are `undefined`. Why is that?

pawelfus:
Removed `Highstock` label - it's a bug in annotations core.

Denyllon:
IMO: This issue is a bit too complex, to resolve it in this way. The problem lies on the process of updating annotation, the sequence of functions call, and finally the way of calculating clip box in general.

To-do's:
 - [ ] rebuild the process of updating `Annotation`
 - [ ] change the way of calculating clip box

pawelfus:
Regarding this:
>  change the way of calculating clip box

This is work in progress, see #13168 PR. Clipping for the whole annotation will be moved to shapes' logic.

PR #13168, opened 2020-03-19
bugfix/12897-annotations-outside-plot-area

### LINKED ISSUE #12897: annotations outside plot area issue (overflow regression)
#### Expected behaviour

Overflow set to 'allow' should allow rendering annotation outside the plot area.
Demo (Highcharts v 6): https://jsfiddle.net/BlackLabel/1cr98bkp/

#### Actual behaviour

In the current version, annotations disappear when is render outside the plot area. 

Demo: https://jsfiddle.net/BlackLabel/p2xh5kba/
Demo with useHTML, where only the text is displayed: https://jsfiddle.net/BlackLabel/f2kgr5xe/

#### Product version

wchmiel:
Fixed #12897, a regression that caused annotations outside the plot area to be hidden even if the `overflow` setting was `allow`. 

~~Fixed #12897, the issue with annotations outside the plot area.~~

### DIFF
--- a/js/annotations/annotations.src.js
+++ b/js/annotations/annotations.src.js
@@ -10,7 +10,7 @@
 'use strict';
 import H from '../parts/Globals.js';
 import U from '../parts/Utilities.js';
-var addEvent = U.addEvent, defined = U.defined, destroyObjectProperties = U.destroyObjectProperties, erase = U.erase, extend = U.extend, find = U.find, fireEvent = U.fireEvent, merge = U.merge, pick = U.pick, splat = U.splat, wrap = U.wrap;
+var addEvent = U.addEvent, defined = U.defined, destroyObjectProperties = U.destroyObjectProperties, erase = U.erase, extend = U.extend, find = U.find, fireEvent = U.fireEvent, isString = U.isString, merge = U.merge, pick = U.pick, splat = U.splat, wrap = U.wrap;
 import '../parts/Chart.js';
 import controllableMixin from './controllable/controllableMixin.js';
 import ControllableRect from './controllable/ControllableRect.js';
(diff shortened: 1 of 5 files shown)


pawelfus:
> As far as I can see, the new `clip` option works in the same way as the crop option of data labels. If that is the case, `annotation.clip` should be renamed `crop` to avoid confusion.

This option is similar to `series.clip` - turns on/off the clipping on the annotation. `annotation.labels[].crop` is already available. Perhaps, instead of adding `annotation.clip/crop` we should add `annotation.shapes[].crop/clip` and refactor clipping in annotations:
- remove current clipping for the parent group (group that stores both, shapes and labels)
- allow `annotation.labels[].crop` do it's thing
- implement clipping for shapes
- add new option `annotation.shapes[].crop` (or clip) to control shapes' clipping. Note that `overflow` option doesn't make sense when shapes are bound to values/points on the chart.

Thoughts @wchmiel , @TorsteinHonsi ?

wchmiel:
I agree. Otherwise, to allow labels to render outside the plot area we will have to set `annotations.clip`, `annotation.labels[].crop` and `annotation.labels[].overflow`.

TorsteinHonsi:
Okay, I see. How about the samples? They look bad now that they are clipped to the plot area. Shouldn't the settings be updated in the samples?

pawelfus:
I think so. Actually, this is a bug, that annotations aren't clipped in these demos. In v8.0.4 all annotations should be clipped to it's pane (xAxis + yAxis).

What do you think about the proposal above @TorsteinHonsi ? Implement `annotations.shapes[].clip` option and let `annotations.labels[].crop` option work for labels.

Just tested this solution, it seems to be broken for labels that are below the plotting area: https://jsfiddle.net/BlackLabel/zpr3bxwt/ - works nice when drag to the left/right or up. Drag a point to the second pane - label (and line!) is still visible:
Screen Shot 2020-03-27 at 13 43 41


We can't allow that in Highstock, example: 
Screen Shot 2020-03-27 at 13 42 04

TorsteinHonsi:
> What do you think about the proposal above @TorsteinHonsi ? Implement annotations.shapes[].clip option and let annotations.labels[].crop option work for labels.

Okay, agreed 👍

pawelfus:
@wchmiel - please keep in mind related issue with clipping: #13164

TorsteinHonsi:
Please merge the master in to this PR, then it's ready to merge

wchmiel:
Actually, we decided in the comments that the solution to this issue should be completely different. Can we close this PR and open a new one when the new fix will be finished?

TorsteinHonsi:
All right, I lost track of the discussion. I'm closing.

Issue #13095, opened 2020-03-05
Tooltip shows outside of plot area for scrollablePlotArea


wdrobinson:
#### Expected behaviour
Tooltip should not display outside of plot when using scrollablePlotArea = true or there should be a way to conditionally remove it. There are cases where having a legend below the chart is necessary and setting opacity to 1 works except the tooltip still shows when you mouse over the bottom of the chart. 

#### Live demo with steps to reproduce
https://jsfiddle.net/Lxo728a5/

#### Product version
Highcharts latest

![image](https://user-images.githubusercontent.com/10948156/76004765-0818a080-5ed8-11ea-8e0e-0ef326a27203.png)

wdrobinson:
In addition to the bug of having the tooltip display where it shouldn't, if you scroll down halfway and move your mouse to the bottom of the chart below the legend so the tooltip causes the page to expand, the chart resizes which resets your scrollbar to the top.

sebastianbochan:
Hi @wdrobinson,
Thank you for reporting about the problem which is caused by `tooltip.outside` parameter. Setting that as a false solves the problem.

Demo:
- https://jsfiddle.net/BlackLabel/Lkxehv8m/

Docs:
- https://api.highcharts.com/highcharts/tooltip.outside

wdrobinson:
tooltip.outside is true by default for scrollable charts and in our case we need the outside tooltip for the chart with the issue so we need a fix or a better workaround for the bugs.

wdrobinson:
Also even with changing tooltip.outside to false the tooltip appears at times because the .highcharts-scrollable-mask doesn't extend the entire width of the chart:
![image](https://user-images.githubusercontent.com/10948156/76352635-c3c93e00-62e5-11ea-9eb1-1c51c1cd7235.png)

https://jsfiddle.net/5j9tb4or/

sebastianbochan:
You are right, now I see the problem. Thank you for the details!

pawelfus:
*Internal note:*
There are some constraints in Tooltip (for scrollablePlotArea) already implemented. Check there what's missing to prevent rendering this tooltip for points outside of the visible area. Note that default kd-tree won't resolve that, because points are rendered within the given extremes.

Izothep:
As a workaround, it should be possible to move function checking constraints outside of split tooltip, and then use it in in `Tooltip.refresh`:
http://jsfiddle.net/BlackLabel/n812y4mp/3/

___
Internal note: 

[...]

PR #13172, opened 2020-03-20
bugfix/12832-fullscreen-scrollableplotarea

### LINKED ISSUE #12832: Scroll doesn't work in the full screen mode
Scroll disappears in the full-screen mode.

[Vertical scroll demo](https://jsfiddle.net/BlackLabel/pg4Lx1yc/)
[Horizontal scroll demo](https://jsfiddle.net/BlackLabel/9r82vzoe/)

Choose the 'view in full screen' in the export menu to see the bug. 

#### Product version

### Internal note:
It works fine after changes in the code and back to previous state.
Should be clearly shown here: https://imgur.com/a/saeKSbL 

------ EDIT -------

scrollablePlotArea changes chart DOM and separate some parts to different divs like: 

```html
<div id="container">
   <div id="highcharts-....." class="highcharts-container">
   </div>
</div>
```

to: 

```html
<div id="container">
   <div class="highcharts-fixed">
   </div>
   <div class="highcharts-scrolling">
      <div class="highcharts-inner-container">
          <div id="highcharts-....." class="highcharts-container">
          </div>
      </div>
   </div>
</div>
```

Current `Fullscreen.prototype.open` implementation affects only on the `"highcharts-inner-container"`. There will be needed some additional condition to affect on other parents when scrollable is applied. 

Current workaround only for the vertical scroll (description inside): https://jsfiddle.net/BlackLabel/1pvLn49e/

Basts:
Fixed #12832, scrollable plot area didn't work in full screen mode.

~~Fixed #12832, fullscreen for scrollable-plot-area.~~

### DIFF
--- a/ts/modules/full-screen.src.ts
+++ b/ts/modules/full-screen.src.ts
@@ -99,10 +99,6 @@ class Fullscreen {
          */
         this.isOpen = false;
 
-        if (!(chart.container.parentNode instanceof Element)) {
-            return;
-        }
-
         const container = chart.renderTo;
 
         // Hold event and methods available only for a current browser.
--- a/ts/modules/full-screen.src.ts
+++ b/ts/modules/full-screen.src.ts
@@ -225,7 +225,7 @@ class Fullscreen {
             );
 
             if (chart.container.parentNode instanceof Element) {
-                const promise = chart.container.parentNode[
+                const promise = chart.renderTo[
                     fullscreen.browserProps.requestFullscreen
                 ]();
 
--- a/ts/modules/full-screen.src.ts
+++ b/ts/modules/full-screen.src.ts
@@ -103,7 +103,7 @@ class Fullscreen {
             return;
         }
 
-        const container = chart.container.parentNode;
+        const container = chart.renderTo;
 
         // Hold event and methods available only for a current browser.
         if (!this.browserProps) {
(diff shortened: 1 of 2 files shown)


Basts:
1. This is the easiest solution which I found, however, I would like to ask if there are any obstacles to use `chart.renderTo` rather than `chart.container.parentNode`? (That's why it is a draft) 

2. Tooltip doesn't work on the full-screen. I think that it is related to this bug #13095.

3. Demo to test: https://jsfiddle.net/BlackLabel/1bkaefpw/

ts/modules/full-screen.src.ts
```diff
@@ -225,7 +225,7 @@ class Fullscreen {
             );
 
             if (chart.container.parentNode instanceof Element) {
-                const promise = chart.container.parentNode[
+                const promise = chart.renderTo[
```
pawelfus: The above `if` does not make sense at all now :)

Basts: Yeah, you are right I will fix it, but first I want to know if there are any obstacles to use `renderTo` rather than `container.parentNode`.

raf18seb: I'm fine with it. You should also change the container on the constructor (line 106).

Basts:
@raf18seb 

Which one do you have in mind? This:

```js
if (chart.container.parentNode instanceof Element) {
...
}
```

I don't think so. Everything works fine without it - I did some tests manually.

raf18seb:
Sorry, I am learning GH ;)
I mean this one:
```
if (!(chart.container.parentNode instanceof Element)) {
    return;
}
```

Basts:
@raf18seb 

It seems that this if is unnecessary.

Changing to:

```js
if (!(chart.renderTo)) {
    return;
}
```
Doesn't make sense, because if the `chart.renderTo` will be unknown the error will have occurred before the Fullscreen module will be applied.

PR #13176, opened 2020-03-20
bugfix/13175-fullscreen-stockTools-bindings

raf18seb:
Fixed #13176, Stock Tools' fullscreen button did not work.


sebastianbochan:
In the past the fullscreen did not work in IFRAMEs by default (javascript limits). If it works, we should focus on the lines:

https://github.com/highcharts/highcharts/blob/1e6571d179ba4ab3a36c4b2afb8cc4721343f741/js/modules/stock-tools-gui.js#L815

Please check that.

raf18seb:
Seba is right, it works in iFrame now so I removed redundant code. I also turned `promise['catch']` into dot notation in fullscreen.ts, it's more readable now (probably TS has changed lately).

TorsteinHonsi:
Please resolve conflicts, otherwise perfect 👍

Issue #12738, opened 2020-01-08
Switch to named import/export and ES6 class pattern


bre1470:
Before we can compile the masters in TypeScript, we have to switch the sources to named import/export, so that a partial namespace is not a problem anymore.

- [x] Accessibility
- [x] Core/Axis/Axis
- [x] Core/Axis/Axis3D
- [x] Core/Axis/BrokenAxis
- [x] Core/Axis/ColorAxis
- [x] Core/Axis/DateTimeAxis
- [x] Core/Axis/GridAxis
- [x] Core/Axis/HiddenAxis
- [x] Core/Axis/LogarithmicAxis
- [x] Core/Axis/MapAxis
- [x] Core/Axis/NavigatorAxis
- [x] Core/Axis/OrdinalAxis
- [x] Core/Axis/PlotLineOrBand
- [x] Core/Axis/RadialAxis
- [x] Core/Axis/ScrollbarAxis
- [x] Core/Axis/StackingAxis
- [x] Core/Axis/Tick
- [x] Core/Axis/Tick3D
- [x] Core/Axis/TreeGridAxis
- [x] Core/Axis/TreeGridTick
- [x] Core/Axis/VMLAxis3D
- [x] Core/Axis/ZAxis
- [x] Core/Chart/Chart
- [x] Core/Chart/Chart3D
- [x] Core/Chart/GanttChart
- [x] Core/Chart/StockChart
- [x] Core/Color
- [x] Core/Globals
- [x] Core/Legend
- [x] Core/MSPointer
- [x] Core/Navigator
- [x] Core/Options (a bit tricky, similar to Utilities)
- [x] Core/Pointer
- [x] Core/Renderer/HTML
- [x] Core/Renderer/SVG/SVGElement
- [x] Core/Renderer/SVG/SVGLabel
- [x] Core/Renderer/SVG/SVGPath
- [x] Core/Renderer/SVG/SVGRenderer
- [x] Core/Renderer/SVG/SVGRenderer3D
- [x] ~Core/Responsive~
- [x] Core/Scrollbar
- [x] Core/Series/DataLabels
- [x] Core/Series/Point
- [x] Core/Series/Series3D
- [x] Core/Time
- [x] Core/Tooltip
- [x] Core/Utilities
- [x] Extensions/Annotations
- [x] Extensions/Annotations/ControlPoint
- [x] Extensions/Annotations/MockPoint
- [x] Extensions/Annotations/NavigationBindings
[...]

PR #13179, opened 2020-03-22
ts/logarithmicaxis

bre1470:
Migrated LogarithmicAxis to ES6 pattern.
Related to #12738.

### DIFF
--- a/js/parts/Axis.js
+++ b/js/parts/Axis.js
@@ -518,12 +518,6 @@ var Axis = /** @class */ (function () {
                 addEvent(axis, eventType, event);
             }
         });
-        // extend logarithmic axis
-        axis.lin2log = options.linearToLogConverter || axis.lin2log;
-        if (axis.isLog) {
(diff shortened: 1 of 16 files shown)


bre1470:
@TorsteinHonsi Both tests wrap (now) non-existence log/lin functions on the Axis prototype. How do we address that? Do we modify the demos to point to the composition property, or do we duplicate functions to be backward compatible with both demos?

pawelfus:
It's not only demos, users are actually using these solutions - we will break compatibility when changing methods.. Just saying, deprecation process could help here ;) Perhaps for now, modify demos to use new solution and keep backward compatibility?

bre1470:
@pawelfus The problem with the deprecate process works only for documented functions. That is the problem with all these half-official stuff floating around. ;)

bre1470:
That said, I will restore the compatibility. It will hinder the migration though.

PR #13190, opened 2020-03-24
enhancement/stock-tools-gui-es6

marvin19:
Tools: Migrated stock-tools-gui to ES6 pattern.


ts/modules/stock-tools-gui.ts
```diff
@@ -934,69 +936,59 @@ addEvent(H.Chart, 'redraw', function (): void {
...
+    public guiEnabled: (boolean|undefined) = void 0 as any;
+    public iconsURL: string = void 0 as any;
+    public lang: (Highcharts.Dictionary<string>|undefined) = void 0 as any;
+    public listWrapper: Highcharts.HTMLDOMElement = void 0 as any;
+    public options: Highcharts.StockToolsGuiOptions = void 0 as any;
+    public placed: boolean =void 0 as any;
+    public showhideBtn: Highcharts.HTMLDOMElement = void 0 as any;
+    public submenu: Highcharts.HTMLDOMElement = void 0 as any;
+    public toolbar: Highcharts.HTMLDOMElement= void 0 as any;
+    public visible: boolean = void 0 as any;
```
bre1470: Properties, that get right away a value in the constructor, should not be initialized (`void 0 as any`). As far as I see, this has to change for `chart`, `eventsToUnbind`, `guiEnabled`, `iconsURL`, `lang`, `options`, `placed`, `visible`.

PR #13191, opened 2020-03-24
ts/logarithmicaxis

bre1470:
- Migrated LogarithmicAxis to ES6 pattern. Related to #12738.
- Improved keepProps handling in Axis.

### DIFF
--- a/js/parts/Utilities.js
+++ b/js/parts/Utilities.js
@@ -1596,7 +1596,9 @@ var discardElement = H.discardElement = function discardElement(element) {
  *         The corrected float number.
  */
 var correctFloat = H.correctFloat = function correctFloat(num, prec) {
-    return parseFloat(num.toPrecision(prec || 14));
+    var v = parseFloat(num.toPrecision(prec || 14));
+    console.log(num, v);
+    return v;
 };
 /**
  * Set the global animation to either a given value, or fall back to the given
--- a/js/parts/Utilities.js
+++ b/js/parts/Utilities.js
@@ -1596,9 +1596,7 @@ var discardElement = H.discardElement = function discardElement(element) {
  *         The corrected float number.
  */
 var correctFloat = H.correctFloat = function correctFloat(num, prec) {
-    var v = parseFloat(num.toPrecision(prec || 14));
-    console.log(num, v);
-    return v;
+    return parseFloat(num.toPrecision(prec || 14));
 };
 /**
  * Set the global animation to either a given value, or fall back to the given
--- a/ts/parts/Axis.ts
+++ b/ts/parts/Axis.ts
@@ -320,6 +320,7 @@ declare global {
             public static defaultRightAxisOptions: AxisOptions;
             public static defaultTopAxisOptions: AxisOptions;
             public static defaultYAxisOptions: YAxisOptions;
+            public static keepProps: Array<string>;
             public constructor(chart: Chart, userOptions: AxisOptions);
             public _addedPlotLB?: boolean;
             public allowZoomOutside?: boolean;
@@ -360,7 +361,7 @@ declare global {
             public isRadial?: boolean;
             public isXAxis?: boolean;
             public isZAxis?: boolean;
-            public keepProps: Array<string>;
+            public keepProps?: Array<string>;
             public labelAlign?: AlignValue;
             public labelEdge: Array<null>;
             public labelFormatter: (
@@ -442,6 +443,7 @@ declare global {
             public generateTick(pos: number, i?: number): void;
             public getClosest(): number;
             public getExtremes(): ExtremesObject;
+            public getKeepProps(): Array<string>;
             public getLinePath(lineWidth: number): SVGPathArray;
             public getLinearTickPositions(
                 tickInterval: number,
@@ -3767,6 +3769,10 @@ class Axis implements AxisComposition {
         }
     };
 
+    // Properties to survive after destroy, needed for Axis.update (#4317,
+    // #5773, #5881).
+    public static keepProps = ['extKey', 'hcEvents', 'names', 'series', 'userMax', 'userMin'];
+
     /* *
      *
      *  Constructors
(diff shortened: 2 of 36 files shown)


ts/parts/Axis.ts
```diff
@@ -5603,8 +5596,8 @@ class Axis implements AxisComposition {
                     this.closestPointRange,
                     true
                 );
-            } else if (this.isLog) {
-                tickPositions = axis.getLogTickPositions(
+            } else if (this.isLog && axis.logarithmic) {
```
TorsteinHonsi: Can we remove `isLog` and just simply test for `axis.logarithmic`? That also means when updating an axis from `log` to `lin` we need to delete `axis.logarithmic`.

ts/parts/Axis.ts
```diff
@@ -7659,13 +7652,9 @@ interface Axis {
...
-extend(Axis.prototype, {
-
-    // Properties to survive after destroy, needed for Axis.update (#4317,
-    // #5773, #5881).
-    keepProps: ['extKey', 'hcEvents', 'names', 'series', 'userMax', 'userMin']
-
-});
+// Properties to survive after destroy, needed for Axis.update (#4317,
+// #5773, #5881).
+Axis.prototype.keepProps = ['extKey', 'hcEvents', 'names', 'series', 'userMax', 'userMin'];
```
TorsteinHonsi: Why is this set separately?

bre1470: This is a prototype property shared by all instances. Alternative would be to add it to the class as a static property.

ts/parts/Axis.ts
bre1470: No as isLog is different in the sense, that it indicates, if the functionality is active and not just loaded/available.

ts/parts/Axis.ts
TorsteinHonsi: Wouldn't that look cleaner?

ts/parts/Axis.ts
TorsteinHonsi: That's what I mean.... Should `axis.logarithmic` only be present when the functionality is active? Then we would need some kind of construct like

```js
if (type === 'logarithmic') {
    this.logarithmic = new LogarithmicAxisAdditions(this);
}
```

... and then this object would be deleted again if the axis is updated to non-logarithmic.

ts/parts/Axis.ts
bre1470: Yes, I agree. When I tried it, unfortunately tests were failing, and since the issue with the Color class, I do not like to make big restructures. I can give it an other try.

ts/parts/Axis.ts
bre1470: I see. With the last two tries, I did not thought of deleting it later on. We can try that tomorrow.

ts/parts/Axis.ts
bre1470: The difference is, that you can mask the prototype property by setting a instance property with the same name, but we could work around that with a getter function, if necessary.

js/parts/Utilities.js
```diff
@@ -1596,7 +1596,9 @@ var discardElement = H.discardElement = function discardElement(element) {
  *         The corrected float number.
  */
 var correctFloat = H.correctFloat = function correctFloat(num, prec) {
-    return parseFloat(num.toPrecision(prec || 14));
+    var v = parseFloat(num.toPrecision(prec || 14));
+    console.log(num, v);
```
pawelfus: :)

bre1470: Thanks, solved. ;)

ts/parts/LogarithmicAxis.ts
```diff
@@ -19,180 +26,267 @@ import H from './Globals.js';
...
-            _minorAutoInterval?: (null|number);
-            getLogTickPositions(
-                interval: number,
-                min: number,
-                max: number,
-                minor?: boolean
-            ): Array<number>;
+            /** @deprecated */
             lin2log(num: number): number;
+            /** @deprecated */
```
TorsteinHonsi: These are marked deprecated, but are used heavily internally?

ts/parts/LogarithmicAxis.ts
```diff
@@ -19,180 +26,267 @@ import H from './Globals.js';
...
         }
+        return positions;
     }
 
-    // Set the axis-level tickInterval variable
-    if (!minor) {
-        axis.tickInterval = interval;
+    public lin2log(num: number): number {
+        const axis = this.axis;
+        const lin2log = axis.options.linearToLogConverter;
```
TorsteinHonsi: Remove the `linearToLogConverter`. It is not officially supported, plus without a reverse function, `logToLinearConverter`, it is unusable.

ts/parts/LogarithmicAxis.ts
bre1470: Yes. It is necessary to support the wrapping in v8.

ts/parts/LogarithmicAxis.ts
```diff
@@ -19,180 +26,267 @@ import H from './Globals.js';
...
+        /**
+         * @deprecated
+         * @private
+         * @function Highcharts.Axis#lin2log
+         *
+         * @param {number} num
+         *
+         * @return {number}
+         */
+        axisProto.lin2log = function (this: Highcharts.Axis, num: number): number {
```
TorsteinHonsi: The to converter functions are duplicated, defined both on the LogarithmicAxisAdditions and on the Axis itself. We need to avoid that. Either one of them need to call the other (probably best if the `axis` calls `axis.logarithmic.lin2log`), or one of them need to be borrowed from the other. Like `axisProto.lin2log = LogarithmicAxisAdditions.prototype.lin2log`.

TorsteinHonsi: .. Or define local functions that are set on both classes.

bre1470: I do this because the `this` scope is different, so you would need wrapper functions anyway. You can not simply assign them.

Issue #13065, opened 2020-03-03
DashStyle for BoxPlot charts doesn't work


thejensi1:
Cannot use the dashStyle property for box plot charts. Only stemDashStyle works but this will just change to style of the stems instead the style for all lines/dashes.

#### Expected behaviour
Setting a dashStyle for a box plot chart should result in changing the style for all lines/dashes.

#### Actual behaviour
Doesn't work at all.

#### Live demo with steps to reproduce
https://jsfiddle.net/thejensi1/Lgms6fn9/3/

raf18seb:
Hi @thejensi1 ,

The API for _series.botplot_ is incorrect (should not be here): https://api.highcharts.com/highcharts/series.boxplot.dashStyle
For this moment, as you said above, setting dashStyle in boxplot works only for a stem.

PR #13193, opened 2020-03-24
enhancement/13065-boxplot-dashStyle

raf18seb:
Added individual dash style settings for various parts of box plots. See #13065. 

~~Enhancement #13065 , added more ways for styling dash styles in the boxplot series.~~

### DIFF
--- a/samples/unit-tests/series-boxplot/boxplot/demo.js
+++ b/samples/unit-tests/series-boxplot/boxplot/demo.js
@@ -191,3 +191,78 @@ QUnit.test(
         );
     }
 );
+
+QUnit.test(
+    'Individual dash styles for box, median, stem and whisker lines (#13065)',
+    function (assert) {
+        var chart = Highcharts.chart('container', {
+
+            chart: {
+                type: 'boxplot'
+            },
+
+            series: [{
+                dashStyle: 'DashDot',
+                medianDashStyle: 'ShortDash',
+                whiskerDashStyle: 'dot',
+                data: [
+                    {
+                        low: 760,
+                        q1: 801,
+                        median: 848,
+                        q3: 895,
+                        high: 965,
+                        boxDashStyle: 'dot',
+                        medianDashStyle: 'dot',
+                        stemDashStyle: 'dot',
+                        whiskerDashStyle: 'dot',
+                        whiskerWidth: 1,
+                        medianWidth: 1
+                    }, {
+                        low: 733,
+                        q1: 853,
+                        median: 939,
+                        q3: 980,
+                        high: 1080
+                    }
+                ]
+            }]
+        });
+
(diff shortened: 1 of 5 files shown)


samples/unit-tests/series-boxplot/boxplot/demo.js
```diff
@@ -190,4 +190,79 @@ QUnit.test(
...
+
+        var series = chart.series[0],
+            firstPoint = series.points[0],
+            secondPoint = series.points[1];
+
+        ['box', 'medianShape', 'stem', 'whiskers'].forEach(function (elem) {
+            assert.strictEqual(
+                firstPoint[elem].attr('stroke-dasharray'),
+                '1,3',
+                'Point\'s ' + elem + 'DashStyle'
```
pawelfus: Friendly reminder for this an other tests: make this a full sentence, e.g. `...dashStyle should be applied to a point`.

ts/parts-more/BoxPlotSeries.ts
```diff
@@ -464,8 +514,11 @@ seriesType<Highcharts.BoxPlotSeries>('boxplot', 'column', {
                         options.stemWidth,
                         options.lineWidth
                     );
-                    stemAttr.dashstyle =
-                        point.stemDashStyle || options.stemDashStyle;
+                    stemAttr.dashstyle = pick(
+                        point.stemDashStyle,
+                        options.stemDashStyle,
+                        options.dashStyle
+                    );
```
pawelfus: Not a problem, just a hint: in general `pick()` is useful when dealing with numerical values, where `a || b` can not be used (as `a=0` would pick `b` value).

ts/parts-more/BoxPlotSeries.ts
```diff
@@ -723,4 +788,64 @@ seriesType<Highcharts.BoxPlotSeries>('boxplot', 'column', {
...
+ * @sample {highcharts} highcharts/css/boxplot/
+ *         Box plot in styled mode
+ *
+ * @type      {Highcharts.DashStyleValue}
+ * @default   Solid
+ * @since     next
+ * @product   highcharts
+ * @apioption series.boxplot.data.boxDashStyle
+ */
+
```
pawelfus: Do need to define twice all options for docs?

raf18seb: I think yes is we want to show them in API not only on the series level but on the point level too: https://api.highcharts.com/highcharts/series.boxplot.data
What is the better practice? Should I remove them from `boxplot.data` level?

pawelfus: No, my bad, didn't realize it's `boxplot.data`, not `boxplot`. Thanks!

PR #13194, opened 2020-03-24
ts/datetimeaxis

bre1470:
Migrated DateTimeAxis to ES6 pattern.
Related to #12738.

### DIFF
--- a/ts/parts/Axis.ts
+++ b/ts/parts/Axis.ts
@@ -4791,12 +4791,13 @@ class Axis implements AxisComposition {
                 });
 
             } else if (
+                axis.dateTime &&
                 axis.isDatetimeAxis &&
                 this.getMinorTickInterval() === 'auto'
             ) { // #1314
                 minorTickPositions = minorTickPositions.concat(
                     axis.getTimeTicks(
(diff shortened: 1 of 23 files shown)


ts/parts/Axis.ts
```diff
@@ -4791,12 +4791,13 @@ class Axis implements AxisComposition {
                 });
 
             } else if (
+                axis.dateTime &&
                 axis.isDatetimeAxis &&
```
TorsteinHonsi: Instead of the `isDatetimeAxis` flag, can we simply test for `axis.dateTime`? That means we also have to delete `axis.dateTime` when updating.

bre1470: It is different in the sense that it indicates, if the axis type is set to `datetime`, not if the axis is capable of date/time support like the composition does.

bre1470: Data grouping uses the composition functions without checking for `isDateTimeAxis`. If that can be solved somehow, we can remove `isDateTimeAxis`.

bre1470: I believe, I have figured it out.

PR #13197, opened 2020-03-25
ts/brokenaxis

bre1470:
Migrated broken-axis to ES6 pattern.


ts/modules/broken-axis.src.ts
```diff
@@ -241,457 +158,626 @@ Axis.prototype.setBreaks = function (
...
+     *
+     * @return {void}
+     */
+    public setBreaks(
+        breaks?: Array<Highcharts.XAxisBreaksOptions>,
+        redraw?: boolean
+    ): void {
+        const brokenAxis = this;
+        const axis = brokenAxis.axis;
+        const isBroken = (isArray(breaks) && !!breaks.length);
```
TorsteinHonsi: As an equivalent to what we have now done with datetime axis and logarithmic axis, shouldn't `isBroken` be either removed so we solely look for `Axis.brokenAxis`, or alternatively be moved to `Axis.brokenAxis`?

ts/modules/broken-axis.src.ts
```diff
@@ -19,33 +30,7 @@ import H from '../parts/Globals.js';
...
-        interface AxisBreakBorderObject {
-            move: string;
-            size?: number;
-            value: number;
-        }
-        interface AxisBreakObject {
-            from: number;
-            len: number;
-            to: number;
+            breakArray: Array<AxisBreakObject>;
```
TorsteinHonsi: Shouldn't this be moved to the `brokenAxis` object to keep the `Axis` clean from _broken_ stuff?

bre1470: Even though I had changed the demos, we have to be backward compatible here.

ts/modules/broken-axis.src.ts
bre1470: I am not sure about this. This is part of the broken axis composition to actually check, what to do, if there are no breaks. I will rename it to `hasBreaks`.

ts/modules/broken-axis.src.ts
```diff
@@ -484,57 +485,48 @@ class BrokenAxis {
...
-                        }
+            if (
+                axis && // #5950
+                axis.brokenAxis &&
+                axis.brokenAxis.hasBreaks
+            ) {
+                const brokenAxis = axis.brokenAxis;
+
+                keys.forEach(function (key: string): void {
+                    breaks = brokenAxis && brokenAxis.breakArray || [];
```
bre1470: If we would test agains `brokenAxis` instead of `axis && axis.brokenAxis`, we do not have to test in the loop again.
```ts
const brokenAxis === (
    axis && // #5950
    axis.brokenAxis
); 
if (
    brokenAxis && 
    brokenAxis.hasBreaks
) {
    keys.forEach(function (key: string): void {
         breaks = brokenAxis.breakArray || [];
```

TorsteinHonsi: Yes I initially did that but then I needed an additional test for `axis` because that's used below. If I understand your comment correctly...

bre1470: In that case, maybe something like this:
```ts
if (!axis) { // #5950
    return;
}

const brokenAxis = axis.brokenAxis;

if (!brokenAxis || !brokenAxis.hasBreaks) {
    return;
}

keys.forEach(function (key: string): void {
         breaks = brokenAxis.breakArray || [];
```

TorsteinHonsi: IMHO the existing logic looks cleaner...

bre1470: Okay. I had just code optimization in mind.

PR #13203, opened 2020-03-25
Bugfix/13161-xrange-disable-stacking

### LINKED ISSUE #13161: xRange - point placement with stacking enabled
#### Expected behaviour
After enabling stacking the points should be stack on the same "row".

#### Actual behaviour
The first point is placed on the second row.

The green bar should be placed next to the red one, like yellow is placed next to blue.
https://jsfiddle.net/BlackLabel/aLty3nug/

#### Product version

#### Affected browser(s)
All

sebastianbochan:
Fixed #13161, stacking was applied on xrange.

### DIFF
--- a/ts/modules/xrange.src.ts
+++ b/ts/modules/xrange.src.ts
@@ -298,6 +298,9 @@ seriesType<Highcharts.XRangeSeries>('xrange', 'column'
 
             swapAxes();
 
+            // #13161 - overwrite unsupported stacking option
+            this.options = merge(this.options, { stacking: void 0 });
+
             metrics = columnType.prototype.getColumnMetrics.call(this);
 
             swapAxes();
--- a/ts/modules/xrange.src.ts
+++ b/ts/modules/xrange.src.ts
@@ -278,7 +278,7 @@ seriesType<Highcharts.XRangeSeries>('xrange', 'column'
         init: function (this: Highcharts.XRangeSeries): void {
             seriesTypes.column.prototype.init.apply(this, arguments as any);
 
-            this.options.stacking = false as any; // #13161
+            this.options.stacking = void 0; // #13161
         },
 
         /**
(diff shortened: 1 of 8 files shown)


ts/modules/xrange.src.ts
```diff
@@ -269,6 +270,17 @@ seriesType<Highcharts.XRangeSeries>('xrange', 'column'
...
 
+        /**
+         * @private
+         * @function Highcarts.seriesTypes.xrange#init
+         * @return {void}
+         */
+        init: function (this: Highcharts.XRangeSeries): void {
+            seriesTypes.column.prototype.init.apply(this, arguments as any);
+
+            this.options.stacking = false as any; // #13161
```
TorsteinHonsi: `false` is not a valid value for stacking. Why not set it to `void 0` so we can avoid the `any` cast?

sebastianbochan: I extracted this value from OHLC series. 

https://github.com/highcharts/highcharts/blob/b5ef0bcb2fcad2af185f8a5f2bd8450b76157f97/ts/parts/OHLCSeries.ts#L219

So I guess that (void 0) we should set this value in both series. correct?

TorsteinHonsi: Yes, correct. This is a survivor from pre-TypeScript code. When converting to TypeScript, we systematically cast all inconsistencies like this to `any` to avoid altering the functionality.  But as we spot obvious mistakes like this, we should seek to remove the `any` casts while working on related parts of the code.

The final step in the TypeScript conversion will be to go over all `any` casts and refactor them.

sebastianbochan: ok, so will commit changes in both files here.

sebastianbochan: @TorsteinHonsi fixed, also in navigator and tests.

PR #13206, opened 2020-03-26
bugfix/13095-scrollable-tooltip-show-outside

### LINKED ISSUE #13095: Tooltip shows outside of plot area for scrollablePlotArea

Izothep:
Fixed #13095, tooltip was showing outside `plotArea` for scrollable graphs.

### DIFF
--- a/js/modules/overlapping-datalabels.src.js
+++ b/js/modules/overlapping-datalabels.src.js
@@ -13,7 +13,7 @@
 'use strict';
 import H from '../parts/Globals.js';
 import U from '../parts/Utilities.js';
-var addEvent = U.addEvent, fireEvent = U.fireEvent, isArray = U.isArray, objectEach = U.objectEach, pick = U.pick;
+var addEvent = U.addEvent, fireEvent = U.fireEvent, isArray = U.isArray, objectEach = U.objectEach, pick = U.pick, isIntersectRect = U.isIntersectRect;
 import '../parts/Chart.js';
 var Chart = H.Chart;
 /* eslint-disable no-invalid-this */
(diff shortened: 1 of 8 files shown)


Izothep:
The current visual diffs are related to problem when points are not visible inside plotArea.  The issue is related to below karma setup:

```
/*
 * Display the tooltip so it gets part of the comparison
 */
Highcharts.prepareShot = function (chart) {
    if (
        chart &&
        chart.series &&
        chart.series[0]
    ) {
        var points = chart.series[0].nodes || // Network graphs, sankey etc
            chart.series[0].points;

        if (points) {
            for (var i = 0; i < points.length; i++) {
                if (
                    points[i] &&
                    !points[i].isNull &&
                    !( // Map point with no extent, like Aruba
                        points[i].shapeArgs &&
                        points[i].shapeArgs.d &&
                        points[i].shapeArgs.d.length === 0
                    ) &&
                    typeof points[i].onMouseOver === 'function'
                ) {
                    points[i].onMouseOver();
                    break;
                }
            }
        }
    }
};
```

In here we are taking first series point, even though it might be outside plotArea (we see this situation in diffs).

The solution is to simply edit `karma-setup`, adding check for points if they are inside plotArea. Good idea might be to use the same method as developed in this PR. However, it looks like a lot of refactoring (adding new method always contains several lines of code) and because of that I would like to know your opinion @TorsteinHonsi, @pawelfus. Do you think current solution is good enough? I had some doubts but it is hard for me to get with something else. I will send you some of my thoughts in separate comment to keep it "short".

Izothep:
The issue in here is related to tooltip visible for points outside `plotArea`. This is caused by no check for `point.inside` in tooltip.refresh (no matter split or not-split). 

At the beginning I thought that using split methods will be enough for creating proper functionality, however, in split we are only limiting tooltip to plotArea using `clamp`. The issue for split tooltip is visible as well, the only difference is that when hovering points outside of `plotArea`, tooltip is rendered inside plotArea (it should not be rendered at all)

In our case we should limit `runPointActions` (no action should be called when point is outside plotArea). 

At first I though that good would be to use already created `chart.isPointInside` or `chart.isInsidePlot` for making this check in runPointActions, but unfortunately these methods are used for rendering points as well and in case of Highstock or Scrollable area we need this points to be rendered - we just want to disable pointActions. That is main reason why I created separate method.

Now I started to think that we might need this values stored in our points - this might help us to make all of the overlapping calculations for inverted/non-inverted graphs etc. but I am not sure if we really need it for now.

That is in short my process of analysing this bug. It would be great if you could add some of your remarks on it.

pawelfus:
Regarding karma: I think adding `point.x >= xAxis.min` (not for pie charts etc.) is reasonable solution. Indeed, we render at least two extra points to keep e.g. line continuous at  the beginning and ending of the chart. Even more points is rendered with other configs (`getExtremesFromAll`).

Regarding solution: I'm okay with proposed solution. If you expect more situations where this might be needed, then I would store this value in point, instead od calculating this every time.

PR #13208, opened 2020-03-26
Series label format and formatter

TorsteinHonsi:
Added new series label options, `series.label.format` and `series.label.formatter`.

The PR also fixes some issues related to the series-label module combined with the rest of the unit tests.

### DIFF
--- a/ts/modules/series-label.src.ts
+++ b/ts/modules/series-label.src.ts
@@ -1073,7 +1073,8 @@ Chart.prototype.drawSeriesLabels = function (): void {
                     let animationOptions: Highcharts.AnimationOptionsObject|undefined;
                     if (isNew) {
                         animationOptions = animObject(series.options.animation);
-                        (animationOptions.duration as number) *= 0.2;
+                        // @todo: Safely remove any cast after merging #13005
+                        (animationOptions.duration as any) *= 0.2;
                     }
 
                     series.labelBySeries
--- a/ts/modules/series-label.src.ts
+++ b/ts/modules/series-label.src.ts
@@ -58,6 +58,8 @@ declare global {
             connectorAllowed?: boolean;
             connectorNeighbourDistance?: number;
             enabled?: boolean;
+            format?: string;
+            formatter?: FormatterCallbackFunction<Series>;
             maxFontSize?: (number|null);
             minFontSize?: (number|null);
             onArea?: (boolean|null);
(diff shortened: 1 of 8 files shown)


ts/modules/series-label.src.ts
```diff
@@ -1032,25 +1067,23 @@ Chart.prototype.drawSeriesLabels = function (): void {
...
                         attr = {};
                     }
-                    (series.labelBySeries as any)
+
+                    // Default initial animation to a fraction of the series
+                    // animation (#9396)
+                    let animationOptions: Highcharts.AnimationOptionsObject|undefined;
+                    if (isNew) {
+                        animationOptions = animObject(series.options.animation);
+                        (animationOptions.duration as number) *= 0.2;
```
bre1470: This kind of casting (`as number`) should be avoided, as it hides undefined/null problems.
- If duration is always set, declaration typing of the property should be changed instead.
- If optional, do `animationOptions.duration = (animationOptions.duration || 0) * 0.2` instead.
- If you can not deal with it now, cast it to `any` and we will pick it up later.

ts/modules/series-label.src.ts
```diff
@@ -174,6 +177,28 @@ H.setOptions({
...
+                 * A format string for the label, with support for a subset of
+                 * HTML. Variables are enclosed by curly brackets. Available
+                 * variables are `name`, `options.xxx`, `color` and other
+                 * members from the `series` object. Use this option also to set
+                 * a static text for the label.
+                 *
+                 * @type string
+                 * @since next
+                 */
+                format: void 0,
```
pawelfus: Should this be simply `format: {name}`?

pawelfus: Ah, I see, then `formatter` will never be called.

ts/modules/series-label.src.ts
pawelfus: If (or when) #13005 will be merged, then `animObject()` will always return `{ duration, defer }` and casting to `any` will be unnecessary. So I would go for now with the third option :)

bre1470: Fantastic! 👍

TorsteinHonsi: Great!

TorsteinHonsi:
> Where AnimationOptionsObject is used as an input, it should become Partial<AnimationOptionsObject>. What do you think?

Yes absolutely. This pattern is all over in the Highcharts source code. Would love to see it implemented.

PR #13213, opened 2020-03-26
bugfix/12834-hidden-series-wrong-extremes

### LINKED ISSUE #12834: Highstock: trying to move the scrollbar when series are hidden and xAxis min and max are set breaks the min and max
#### Expected behaviour

Nothing should happen when a user try to move the scrollbar and no series is visible.

#### Actual behaviour

If a min and a max are set for the x axis and all the series are not visible, trying to move the scrollbar breaks the min and max.

https://codepen.io/Korbraan/pen/jOEXoRW

Hide the series with the button and then try to move the scrollbar.

#### Product version
Highstock 6.0.3+

#### Affected browser(s)
Firefox, Opera, Chrome, ...

sebastianbochan:
Fixed #12834, hidden series applied wrong extremes on X axis when scrolling.

~~Fixed #12834, hidden series applied wrong extremes on xaxis when scroll.~~

### DIFF
--- a/ts/parts/ScrollbarAxis.ts
+++ b/ts/parts/ScrollbarAxis.ts
@@ -235,9 +235,7 @@ class ScrollbarAxis {
         addEvent(AxisClass, 'afterGetOffset', function (): void {
             var axis = this as ScrollbarAxis,
                 index = axis.horiz ? 2 : 1,
-                scrollbar = axis.scrollbar,
-                axisMin = pick(axis.options.min, axis.min as any),
-                axisMax = pick(axis.options.max, axis.max as any);
+                scrollbar = axis.scrollbar;
 
             if (scrollbar) {
                 axis.chart.scrollbarsOffsets = [0, 0]; // reset scrollbars offsets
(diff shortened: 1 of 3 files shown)


sebastianbochan:
Will debug it.

ts/parts/ScrollbarAxis.ts
```diff
@@ -224,7 +235,9 @@ class ScrollbarAxis {
         addEvent(AxisClass, 'afterGetOffset', function (): void {
             var axis = this as ScrollbarAxis,
                 index = axis.horiz ? 2 : 1,
-                scrollbar = axis.scrollbar;
+                scrollbar = axis.scrollbar,
+                axisMin = pick(axis.options.min, axis.min as any),
+                axisMax = pick(axis.options.max, axis.max as any);
```
pawelfus: I don't see these values used in this scope - is that necessary?

TorsteinHonsi:
Awaiting @pawelfus' approval

PR #13223, opened 2020-03-30
ts/parallelaxis

bre1470:
Migrated ParallelAxis to ES6 pattern.

### DIFF
--- a/ts/modules/parallel-coordinates.src.ts
+++ b/ts/modules/parallel-coordinates.src.ts
@@ -216,7 +216,7 @@ class ParallelAxisAdditions {
         const axis = parallel.axis;
         const chart = axis.chart;
 
-        var fraction = ((parallel.position || NaN) + 0.5) /
+        var fraction = (pick(parallel.position, NaN) + 0.5) /
             (chart.parallelInfo.counter + 1);
 
         if (chart.polar) {
--- a/ts/modules/parallel-coordinates.src.ts
+++ b/ts/modules/parallel-coordinates.src.ts
@@ -216,7 +216,7 @@ class ParallelAxisAdditions {
         const axis = parallel.axis;
         const chart = axis.chart;
 
-        var fraction = (pick(parallel.position, NaN) + 0.5) /
+        var fraction = ((parallel.position || 0) + 0.5) /
             (chart.parallelInfo.counter + 1);
 
         if (chart.polar) {
--- a/ts/modules/parallel-coordinates.src.ts
+++ b/ts/modules/parallel-coordinates.src.ts
@@ -251,6 +251,9 @@ class ParallelAxis {
 
         /* eslint-disable no-invalid-this */
 
+        // On update, keep parallel additions.
+        Axis.keepProps.push('parallel');
+
         // Add parallel addition
         addEvent(AxisClass, 'init', function (): void {
             const axis = this;
--- a/ts/modules/parallel-coordinates.src.ts
+++ b/ts/modules/parallel-coordinates.src.ts
@@ -13,6 +13,7 @@
 'use strict';
 
 import type RadialAxis from '../parts-more/RadialAxis';
+import Axis from '../parts/Axis.js';
 import H from '../parts/Globals.js';
 
 /**
@@ -21,14 +22,6 @@ import H from '../parts/Globals.js';
  */
 declare global {
     namespace Highcharts {
-        interface Axis {
-            parallelPosition?: ParallelAxis['parallelPosition'];
-            /** @requires modules/parallel-coordinates */
-            setParallelPosition(
-                axisPosition: Array<string>,
-                options: AxisOptions
-            ): void;
-        }
         interface Chart {
             hasParallelCoordinates?: ParallelChart['hasParallelCoordinates'];
             parallelInfo?: ParallelChart['parallelInfo'];
@@ -38,10 +31,6 @@ declare global {
         interface ChartOptions {
             parallelAxes?: XAxisOptions;
             parallelCoordinates?: boolean;
-        }
-        interface ParallelAxis extends Axis {
-            chart: ParallelChart;
-            parallelPosition: number;
         }
         interface ParallelChart extends Chart {
             hasParallelCoordinates?: boolean;
@@ -77,10 +66,8 @@ import '../parts/Chart.js';
 import '../parts/Series.js';
 
 // Extensions for parallel coordinates plot.
-var Axis = H.Axis,
-    Chart = H.Chart,
-    ChartProto = Chart.prototype,
-    AxisProto = H.Axis.prototype;
+var Chart = H.Chart,
+    ChartProto = Chart.prototype;
 
 var defaultXAxisOptions = {
     lineWidth: 0,
@@ -88,6 +75,8 @@ var defaultXAxisOptions = {
     opposite: true,
     type: 'category'
 };
+
+/* eslint-disable valid-jsdoc */
 
 /**
  * @optionparent chart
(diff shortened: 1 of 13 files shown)


ts/modules/parallel-coordinates.src.ts
```diff
@@ -315,114 +485,6 @@ extend(ChartProto, /** @lends Highcharts.Chart.prototype */ {
 });
 
 
-// On update, keep parallelPosition.
-Axis.keepProps.push('parallelPosition');
```
pawelfus: I think this is necessary?

bre1470: If that is the case, do I have to add the `parallel` property here?

bre1470: Please review.

pawelfus: In general, when Axis is updated, it's recreated from scratch. However, some properties are stored (from array `keepProps`) and assigned back.

I don't know if that is still necessary for `parallelPosition` property (or the new one `parallel.position`). Here is the case: https://jsfiddle.net/BlackLabel/c0m8L7d6/3/ - after 1s, updated second yAxis without issues. Here is current build: https://jsfiddle.net/BlackLabel/c0m8L7d6/2/ - btw, first axis is missing.

pawelfus: Note that other compositions may not need that property at all.

bre1470: Okay, I fixed it and add compositions to keepProps only, if it makes sense. You fiddle works with the last commit: https://jsfiddle.net/1zu08j4t/

ts/parts/DateTimeAxis.ts
```diff
@@ -211,6 +211,8 @@ class DateTimeAxis {
      */
     public static compose(AxisClass: typeof Axis): void {
 
+        AxisClass.keepProps.push('dateTime');
```
pawelfus: So `dateTime` does need to be in `keepProps`? Why?

bre1470: For performance reasons, as we skip creating a new instance during re-init of the axis.

ts/parts/NavigatorAxis.ts
```diff
@@ -155,11 +155,16 @@ class NavigatorAxis {
      */
     public static compose(AxisClass: typeof Axis): void {
 
+        AxisClass.keepProps.push('navigatorAxis');
```
pawelfus: I guess this one makes sense because of `Axis.navigatorAxis.previousZoom`?

ts/parts/OrdinalAxis.ts
```diff
@@ -355,11 +355,12 @@ class OrdinalAxis {
         SeriesClass: typeof Series
     ): void {
 
+        AxisClass.keepProps.push('ordinal');
```
pawelfus: I don't see a reason to keep `ordinal` prop.

ts/parts/LogarithmicAxis.ts
```diff
@@ -215,7 +215,7 @@ class LogarithmicAxis {
      */
     public static compose(AxisClass: typeof Axis): void {
 
-        /* eslint-disable no-invalid-this */
+        AxisClass.keepProps.push('logarithmic');
```
pawelfus: I think this one actually makes sense 🤔I see there's code to be executed when switching from `log` to `linear` type. Nice catch! 👍

ts/parts/OrdinalAxis.ts
bre1470: So we instantiate it only ones per axis instance and not with every axis init: https://github.com/highcharts/highcharts/blob/5133f3e50d90fac52261571ac43fd0c76a076de5/ts/parts/OrdinalAxis.ts#L939 Condition is missing for that though.

ts/parts/LogarithmicAxis.ts
bre1470: I disagree with you. Saving constructor calls is sensible, too.

ts/parts/DateTimeAxis.ts
pawelfus: Just to explain: It's similar to `logarithmic`, but the only action on `destroy` is to delete `axis.dateTime.axis`.

ts/parts/LogarithmicAxis.ts
pawelfus: But the code is clearing props only on `axis.linear` object. I'm not so convinced anymore.. :)

pawelfus: Above comment was added before your loaded for me, I'm commenting myself comment above. Saying that, I don't understand what do you mean.

ts/parts/DateTimeAxis.ts
pawelfus: > For performance reasons, as we skip creating a new instance during re-init of the axis.

Yes, that makes sense! 👍

bre1470:
I wonder, if we should introduce a better way than keepProps for complex objects. How about a `keepMe` property, that `Axis.destroy` checks on its properties? For example `Axis.ordinal.keepMe` would be true and so it would skip the keepProps check.

ts/parts/LogarithmicAxis.ts
bre1470: I removed the destroy over design. :)
I mean this part here: https://github.com/highcharts/highcharts/blob/5133f3e50d90fac52261571ac43fd0c76a076de5/ts/parts/LogarithmicAxis.ts#L242-L244

ts/parts/OrdinalAxis.ts
bre1470: I added the condition for ordinal composition.

pawelfus:
> For example `Axis.ordinal.keepMe` would be true and so it would skip the `keepProps` check.

FYI: We do that in Series: `Series.propertyName.survive` (for SVG elements), but it's not much different from `keepProps` check.

bre1470:
@pawelfus I addressed all you concerns (I hope). Please review ones more.

bre1470:
@TorsteinHonsi Should we introduce a sub-property `survive` for the `Axis.destroy` loop?

TorsteinHonsi:
> Should we introduce a sub-property survive for the Axis.destroy loop?

Can't do that, these properties may be any type. `extKey` for example is a string, `userMin` is `number|undefined`.

The `keepProps` concept is light-weight and has been working well (pragmatic in other words). I don't quite see a reason for changing it?

bre1470:
@TorsteinHonsi For me this is isolated: https://github.com/highcharts/highcharts/blob/d17c85e73866e46ac9cb99f9969eda811d23ea5f/ts/modules/parallel-coordinates.src.ts#L166-L237 Or did you meant something else?

TorsteinHonsi:
Yes, I meant these additional properties on the Axis and Chart:

```js
        interface Axis {
            parallelPosition?: ParallelAxis['parallelPosition'];
            /** @requires modules/parallel-coordinates */
            setParallelPosition(
                axisPosition: Array<string>,
                options: AxisOptions
            ): void;
        }
        interface Chart {
            hasParallelCoordinates?: ParallelChart['hasParallelCoordinates'];
            parallelInfo?: ParallelChart['parallelInfo'];
            /** @requires modules/parallel-coordinates */
            setParallelInfo(options: Options): void;
        }
```

bre1470:
We did not discussed the Chart migrasjon yet, so I leave it everywhere out.

bre1470:
I will create new PRs. This becomes confusing because of the keepProps side stuff.

PR #13224, opened 2020-03-30
deprecate

bre1470:
Added deprecation warnings in console.

### DIFF
--- a/ts/parts/Utilities.ts
+++ b/ts/parts/Utilities.ts
@@ -703,7 +703,7 @@ var charts = H.charts,
  *
  * @return {void}
  */
-const error = H.error = function (
+function error(
     code: (number|string),
     stop?: boolean,
     chart?: Highcharts.Chart,
@@ -718,7 +718,10 @@ const error = H.error = function (
                 throw new Error(message);
             }
             // else ...
-            if (win.console) {
+            if (
+                win.console &&
+                error.messages.indexOf(message) === -1 // prevent console flooting
+            ) {
                 console.log(message); // eslint-disable-line no-console
             }
         };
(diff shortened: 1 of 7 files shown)


ts/parts/Utilities.ts
```diff
@@ -2698,6 +2699,44 @@ const find = H.find = (Array.prototype as any).find ?
...
+            }
+        }
+    }
+    if (
+        fnName &&
+        deprecateFnNames.indexOf(fnName) === -1
+    ) {
+        deprecateFnNames.push(fnName);
+        // eslint-disable-next-line no-console
+        console.log(
```
pawelfus: Check if `win.console` is available. Unless we have some polyfill for that, but I guess we don't, since we check that in `H.error()` too. Why? Old IE didn't initiate `win.console` unless dev-tools were open and threw error..

bre1470:
Yes, I agree, but this PR is about the function. The PR did not deprecated functions, the doclets already state, that they are. The function just makes it visible. The warnings will have to be addressed in a separate PR though.

bre1470:
I see this not as a question of complexity but as of support. Customers have to be informed, when they are using deprecated functions.

bre1470:
That said, we could as an alternative use the `H.error` function instead and introduce an error 100 for deprecated stuff.

pawelfus:
> Do we need it?

I think we do. Let me copy&paste my thoughts from core_devs channel:

> How about we define a procedure for making options (and methods/classes) deprecated?
>
> We keep adding new options, sometimes those are just shortcuts for the same config (I know we have some equivalent options, e.g. `navigator.scrollbar `is just passed on to `navigator.xAxis[0].scrollbar`). How about we consider cleaning up the code base from time to time? Or setting a warning, that option will be deprecated at version/time? I’m not sure if we even can do that (it’s hard to predict a release date/version with irregular releases), but just wanted to share my thoughts.
>
> For example we are about to implement `dataLabels.animation` config for initial dataLabels animation, and keep `dataLabels.defer` for backward compatibility. This is fine, but I think in 2021 we could remove this option? Of course, we would keep this in API docs, like any other deprecated options.


> Is it worth the added complexity?

Could you elaborate?

PR #13232, opened 2020-03-31
bugfix/tilemap-incorrect-marker-attributes

Denyllon:
Tilemap produced some errors after heatmap refactoring (on `master`), when hovering points.
Some prototype functions and features from the old tilemap series had to be restored back, in order to keep it work as before.

### DIFF
--- a/samples/unit-tests/series-tilemap/coloraxis/demo.js
+++ b/samples/unit-tests/series-tilemap/coloraxis/demo.js
@@ -30,4 +30,23 @@ QUnit.test('Tilemap and ColorAxis', function (assert) {
         'ColorAxis.max should be the same as max value in points (#11644)'
     );
 
+    chart.series[0].update({
+        tileShape: 'circle'
+    });
+
+    var point = chart.series[0].points[1];
+    point.setState('hover');
+
+    assert.notEqual(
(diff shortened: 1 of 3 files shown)


sebastianbochan:
Looks good 👍

samples/unit-tests/series-tilemap/coloraxis/demo.js
```diff
@@ -30,4 +30,23 @@ QUnit.test('Tilemap and ColorAxis', function (assert) {
...
+    chart.series[0].update({
+        tileShape: 'circle'
+    });
+
+    var point = chart.series[0].points[1];
+    point.setState('hover');
+
+    assert.notEqual(
+        point.graphic.element.getAttribute('cx'),
+        "NaN",
```
sebastianbochan, line 42: Consider to use 

```
assert.ok(
        Highcharts.isNaN(point.graphic.element.getAttribute('cx')),
```

TorsteinHonsi: I prefer the original `assert.notEqual`. It is stricter...

sebastianbochan: Good to know, thanks!

PR #13235, opened 2020-03-31
ts/parallel-axis

bre1470:
Migrated ParallelAxis to ES6 pattern. Replaces partially PR #13223.

### DIFF
--- a/ts/modules/parallel-coordinates.src.ts
+++ b/ts/modules/parallel-coordinates.src.ts
@@ -13,6 +13,7 @@
 'use strict';
 
 import type RadialAxis from '../parts-more/RadialAxis';
+import Axis from '../parts/Axis.js';
 import H from '../parts/Globals.js';
 
 /**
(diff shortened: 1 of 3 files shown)


ts/modules/parallel-coordinates.src.ts
```diff
@@ -174,6 +163,190 @@ H.setOptions({
...
+        });
+
+        /* eslint-enable no-invalid-this */
+
+    }
+}
+
+interface ParallelAxis extends Axis {
+    chart: Highcharts.ParallelChart;
+    parallel: ParallelAxisAdditions;
```
TorsteinHonsi, line 343: I think this one should be called `parallelCoordinates` to avoid any future confusion.

bre1470:
This is not the right time to work on chart compositions. Chart class is not migrated and the assembler does not support several type exports in one file.

bre1470:
Or putted in other words, I will work on this after axis migration. :)

TorsteinHonsi:
> Or putted in other words, I will work on this after axis migration. :)

Okay!

@pawelfus , are you okay with the latest changes? Adding the parallel coordinates stuff into an `Axis.parallelCoordinates` object?

PR #13241, opened 2020-04-01
enhancement/packedbubble-smooth-animation

Izothep:
Packedbubble, improved bubbles animation.


Izothep:
http://jsfiddle.net/rfuah3dp/ - mock-up of solution in case of dataLabels animation

js/modules/networkgraph/node-animation.js
```diff
@@ -0,0 +1,150 @@
...
+                            label.addClass('highcharts-simulation-dataLabels-fadeout');
+                            if (!styledMode) { // <--- !styledMode :
+                                label.css({
+                                    opacity: 1,
+                                    transition: 'opacity 2000ms'
+                                });
+                            }
+                        }
+                        else {
+                            complete = function () {
```
sebastianbochan, line 71: These lines are similar to IF, except place it in the complete. I would consider to extract it as an internal function and call in IF or in Complete.

function should be also with param (opacity 0 or 1 and adapt classes).

js/modules/networkgraph/node-animation.js
```diff
@@ -0,0 +1,150 @@
...
+            var complete, newOpacity;
+            if (label) {
+                newOpacity = label.newOpacity;
+                if (label.oldOpacity !== newOpacity) {
+                    // Make sure the label is completely hidden
+                    // to avoid catching clicks (#4362)
+                    if (label.alignAttr && label.placed) { // data labels
+                        if (newOpacity) {
+                            label.removeClass('highcharts-simulation-dataLabels');
+                            label.addClass('highcharts-simulation-dataLabels-fadeout');
```
sebastianbochan, line 64: 1. Do we need two classes, instead od one?
2. Classes should not be applied only when styled mode?

js/modules/networkgraph/node-animation.js
Izothep: Thanks! I have refactored this part :)

ts/parts-more/PackedBubbleSeries.ts
```diff
@@ -1447,6 +1451,79 @@ seriesType<Highcharts.PackedBubbleSeries>(
...
+         *        The Point to inspect.
+         *
+         * @param {string} [state]
+         *        The state, can be either `hover`, `select` or undefined.
+         *
+         * @return {Highcharts.SVGAttributes}
+         *         A hash containing those attributes that are not settable from
+         *         CSS.
+         */
+        markerAttribs: function (
```
TorsteinHonsi, line 1474: Is all this just to fix the problem with `Math.round`? In that case we should just have a series property that dictates whether the position should be rounded...

css/highcharts.scss
```diff
@@ -823,3 +823,13 @@ text.highcharts-drilldown-data-label,
 .highcharts-grid-axis .highcharts-axis-line {
     stroke-width: 1px;
 }
+
+/* NetworkGraph and Packed Bubble */
+.highcharts-simulation-dataLabels-fadeout {
```
TorsteinHonsi, line 828: First, in CSS, don't use camel case. We already have classes like `.highcharts-data-labels`.
Second, I don't think the words _simulation_ and _fadeout_ ready belong in this class name. It should probably be more generic, like `.highcharts-packedbubble-series .highcharts-data-label` and `.highcharts-packedbubble-series .highcharts-data-label-hidden`. Because what we're really define in the CSS is _how_ we show/hide the labels.

ts/parts-more/PackedBubbleSeries.ts
Izothep: Definitely I am with you on this one! I was afraid to make this kind of adjustment in Highcharts core but I will do it in my next commit.

Izothep:
> Thanks for the PR!
> 
> Apart from the inline comments, I am wondering what the big rewrite/addition to `hideOverlappingLabels` is about? The fading in and out of labels already works almost perfectly out of the box, just by applying styles: https://jsfiddle.net/highcharts/9khmbsaL/ . What problem are we resolving? Maybe we don't need a heavy fix with lots of duplicated code?

In case of this refactoring, I have made it to not work/rewrite overlapping dataLabels for whole chart when the change is needed only for packed bubble and networkgraph. I have thought that good idea might be to divide big `hideOverlappingLabels` method into three smaller methods - which allows us to modify only small parts of this method if needed instead of copy-pasting whole function.

I will take some deeper look into eventual simpler solution on flickering dataLabels problem visible in your demo - maybe this refactoring won't be needed.

css/highcharts.scss
Izothep: Corrected, Thank you for catching this :)

ts/parts-more/PackedBubbleSeries.ts
Izothep: Edited - used crisp option as discussed

ts/modules/overlapping-datalabels.src.ts
```diff
@@ -243,22 +234,23 @@ Chart.prototype.hideOverlappingLabels = function (
...
+                        }
+                        label[newOpacity ? 'show' : 'hide'](true);
+                        label.placed = !!newOpacity;
+                    };
 
                     isLabelAffected = true;
 
                     // Animate or set the opacity
                     label.alignAttr.opacity = newOpacity;
+                    animOptions = label.options && label.options.animation || null;
```
TorsteinHonsi, line 250: I don't think using JavaScript animations here is a good idea. It is very slow as it competes with the simulation in resources. In fact, in the network-graph main demo it doesn't seem to work properly at all, all the labels are stuck semi-transparent.

Instead, I think we should disable JS animation as we currently do, and use CSS transitions. It runs much smoother. Network graph and packed bubble should have a _default_ `dataLabel.style.transition: 'opacity 2s'` or something similar.

ts/parts/Globals.ts
```diff
@@ -31,6 +31,8 @@ declare global {
             frameShapes?: any; // @todo highcharts 3d
             isBoosting?: any; // @todo boost module
             hideOverlappingLabels: Function; // @todo overlapping module
+            getLabelBoxes: Function; // @todo overlapping module
+            hideOrShowLabels: Function; // @todo overlapping module
```
TorsteinHonsi, line 35: These are redundant now, right?

ts/parts/DataLabels.ts
```diff
@@ -44,6 +44,7 @@ declare global {
         interface DataLabelsOptions {
             align?: (AlignValue|null);
             allowOverlap?: boolean;
+            animation?: boolean|AnimationOptionsObject;
```
TorsteinHonsi, line 47: Remove if we go for CSS animation

ts/parts/Series.ts
```diff
@@ -2788,7 +2807,16 @@ H.Series = seriesType<Highcharts.LineSeries>(
...
+            y: 0,
+
+            /**
+             * The animation for changing labels opacity.
+             *
+             * @since     next
+             * @default   false
+             * @type    {boolean|Highcharts.AnimationOptionsObject}
+             */
+            animation: false
```
TorsteinHonsi, line 2819: If we go for CSS animation of the opacity, this should be removed.

ts/parts/Utilities.ts
```diff
@@ -1382,6 +1382,26 @@ function clamp(value: number, min: number, max: number): number {
...
 
+/**
+ * Utility function checking if two rectangles are overlapping
+ *
+ * @private
+ * @param {Highcharts.BBoxObject} box1 BBox of the first rectangle
+ * @param {Highcharts.BBoxObject} box2 BBox of the second rectangle
+ * @return {number} Returns a boolean value if two rectangles intersect.
+*/
+function isIntersectRect(
```
TorsteinHonsi, line 1393: Since we're using this function only in `overlapping-datalabels`, we can move it back to that file. If, in the future, we want to use it in other places, we can move it to Utilities and expose it.

ts/parts/Globals.ts
Izothep: Yes, sorry, I forgot to remove them!

ts/modules/overlapping-datalabels.src.ts
Izothep: Thanks! Sure, it looks that all I have to do in here is to remove hide/show in complete method an return to 
```
label[label.isOld ? 'animate' : 'attr'](
        label.alignAttr,
        null as any,
        complete
);
```

It looks to be working correctly now but I will make this PR draft again as I need to test if everything works good with this small change.

samples/unit-tests/axis/stacklabels/demo.js
```diff
@@ -232,8 +232,9 @@ QUnit.test('Stack labels overlapping issue #11982', function (assert) {
 
 
     assert.strictEqual(
```
sebastianbochan, line 234: you can use `assert.ok`

samples/unit-tests/datalabels/overlapping/demo.js
```diff
@@ -128,7 +128,8 @@ QUnit.test(
         });
 
         assert.strictEqual(
```
sebastianbochan, line 130: you can use assert.ok

samples/unit-tests/series/datalabels/demo.js
```diff
@@ -272,7 +272,9 @@ QUnit.test(
 
         assert.strictEqual(
```
sebastianbochan, line 273: you can use assert.ok

ts/modules/overlapping-datalabels.src.ts
```diff
@@ -126,6 +125,17 @@ Chart.prototype.hideOverlappingLabels = function (
         box1,
         box2,
         isLabelAffected = false,
+        isIntersectRect = function (
+            box1: Highcharts.BBoxObject,
```
sebastianbochan: What is `box1` / `box2`, its a little unclear. 

Maybe we should use` box` and `nextBox` or `siblingBox` ?

ts/parts/Interaction.ts
```diff
@@ -1255,7 +1255,7 @@ extend(Point.prototype, /** @lends Highcharts.Point.prototype */ {
 
                 // Some inactive points (e.g. slices in pie) should apply
                 // oppacity also for it's labels
-                if (series.options.inactiveOtherPoints) {
+                if (series.options.inactiveOtherPoints && pointAttribs.opacity) {
```
sebastianbochan, line 1258: Is pointAttribs always defined? 

Opacity is not default like 0 or 1 ?

Izothep: not in case of pointAttribs in inactive state, when it is undefined the dataLabels opacity is `NaN`. Live example: https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/plotoptions/sankey-minlinkwidth

Hover over graph and then check dataLabels opacity

Izothep: but `pointAttribs` is always defined, yes

ts/modules/overlapping-datalabels.src.ts
Izothep: I would prefer to leave it for now as it is - it is an existing method, that I have just re-added. I have tried to move it to utils.js but for now it has no point so needed to move it back in here. 

When this method will go to Utils, the comments and naming may change.

samples/unit-tests/axis/stacklabels/demo.js
Izothep: done.

samples/unit-tests/datalabels/overlapping/demo.js
Izothep: done

samples/unit-tests/series/datalabels/demo.js
Izothep: done :)

TorsteinHonsi:
Looks good to me now. Is it ready to merge?

Izothep:
Yes, I believe it is ready to merge now.

Issue #11782, opened 2019-08-26
Selection is not provided in split packed bubble


Surekaraja:
#### Expected behaviour
Need to select each group in split packed bubble as like in pie chart

#### Actual behaviour
Selection can't be achieved as like in pie chart. Provide solution 

#### Live demo with steps to reproduce
template: https://jsfiddle.net/hvecayfx/1/ 

#### Product version
highcharts v7

Izothep:
Hi @Surekaraja!

Thank you for reporting your issue. 

Could you please send us more specific information how you would like to select split packed bubbles? 

For now it is already possible to select packed bubbles using allowPointSelect option.

Live example: https://jsfiddle.net/BlackLabel/175xhsek/

Surekaraja:
Hi

Here in your reply you have given selection only for individual points of particular split group. I need selection for that particular split group not for the points inside that particular split group. Kindly reply as early as possible.


Kind Regards,

Sureka R

BI Developer - Trainee

[cid:52521dd3-6bef-4982-9f80-0236d7c06395]

Look Forward. Think Ahead.

Main: +91-9710062203
Mobile: +91-8610845038

Innovation & Services Hub<http://visualbi.com/about-us/locations/>

________________________________

Izothep:
Thank you for providing us with further information about your idea. 

Unfortunately, for now it is not possible to use similar feature as it was not intended to give point events callback functionality to parentNodes. I think it might be good idea to add similar option in future Highcharts releases.

For now, as a workaround it might be possible to manually call parentNode.select:
https://jsfiddle.net/BlackLabel/t8vd61mx/

Surekaraja:
Hi
 You have given the solution for single series but I need solution for series array without doing iteration. If possible can you provide solution without that extra function. Do you have any other options for this kind of selection?

 https://jsfiddle.net/ekcqfg7z/1/


Kind Regards,

Sureka R

BI Developer - Trainee

[cid:985a5c6e-44ff-4e0c-819e-b5e9efe320d7]

Look Forward. Think Ahead.

Main: +91-9710062203
Mobile: +91-8610845038

Innovation & Services Hub<http://visualbi.com/about-us/locations/>

________________________________

sebastianbochan:
Hi @Surekaraja,
Thank you for the feedback. 

Updated demo:
- https://jsfiddle.net/BlackLabel/6e7k0t1y/

Please be aware of limits of the solution because at this moment its only workaround.

[...]

PR #13244, opened 2020-04-02
enhancement/11782-packedbubble-parent node selection

mateuszkornecki:
Enhancement #11782, added support for selecting and dragging parentNode in the packedbubble.


js/parts-more/PackedBubbleSeries.js
```diff
@@ -1123,14 +1172,30 @@ seriesType('packedbubble', 'bubble',
...
-    onMouseMove: dragNodesMixin.onMouseMove,
+    onMouseMove: function (point, event) {
+        var isParentNode = point.isParentNode, parentNodeLayout = this.parentNodeLayout;
+        if (isParentNode) {
+            parentNodeLayout.setMaxIterations();
+            parentNodeLayout.currentStep = 0;
+            parentNodeLayout.step();
+        }
+        dragNodesMixin.onMouseMove.apply(this, arguments);
+    },
```
Izothep, line 1198: As we use the same code in onMouseMove and onMouseDown, maybe we can move them to separate method?

mateuszkornecki: You are absolutely right, I will fix it.

ts/parts-more/PackedBubbleSeries.ts
```diff
@@ -1860,6 +1958,123 @@ seriesType<Highcharts.PackedBubbleSeries>(
...
+            eventArgs?: (Highcharts.Dictionary<any>|Event),
+            defaultFunction?: (
+                Highcharts.EventCallbackFunction<Highcharts.Point>|Function
+            )
+        ): void {
+            const point = this,
+                series = this.series,
+                seriesOptions = series.options;
+
+            // load event handlers on demand to save time on mouseover/out
```
pawelfus, line 1974: Please, start comments with capital letter :)

ts/parts-more/PackedBubbleSeries.ts
```diff
@@ -1860,6 +1958,123 @@ seriesType<Highcharts.PackedBubbleSeries>(
...
+                (
+                    point.options &&
+                    point.options.events &&
+                    (point.options.events as any)[eventType]
+                )
+                ) {
+                    point.importEvents();
+                }
+
+                // add default handler if in selection mode
```
pawelfus, line 1986: Capital

ts/parts-more/PackedBubbleSeries.ts
```diff
@@ -1860,6 +1958,123 @@ seriesType<Highcharts.PackedBubbleSeries>(
...
+                         * @name Highcharts.Point#selected
+                         * @type {boolean}
+                         */
+                        parentNode.selected = parentNode.options.selected = selected;
+                        (series.options.data as any)[series.data.indexOf(parentNode)] =
+                            parentNode.options;
+
+                        parentNode.setState((selected as any) && 'select');
+
+                        // unselect all other points unless Ctrl or Cmd + click
```
pawelfus, line 2044: Capital

ts/parts-more/PackedBubbleSeries.ts
```diff
@@ -1860,6 +1958,123 @@ seriesType<Highcharts.PackedBubbleSeries>(
...
+                        }
+                    };
+                }
+
+                fireEvent(point, eventType, eventArgs, defaultFunction);
+            } else {
+                return Point.prototype.firePointEvent.apply(this, arguments as any);
+            }
+        },
+        select: function (
```
pawelfus, line 2009: This method looks like a copy&paste of `point.select()`. How about instead doing something like this:

```
if (isParent) {
  this.series.getSelectedPoints = this.getSelectedParents;
  Point.proto.select(this, args);
  this.series.getSelectedPoints = H.Series.proto.getSelectedPoints;
} else {
  Point.proto.select(this, args);
}
```

pawelfus: The same applies to the method above `firePointEvent` - can we reuse method instead of copy&paste, and just modify options that are used inside the method?

ts/parts/Interaction.ts
```diff
@@ -52,7 +52,7 @@ declare global {
             importEvents(): void;
             onMouseOut(): void;
             onMouseOver(e?: PointerEventObject): void;
-            select(selected?: boolean, accumulate?: boolean): void;
+            select(selected?: boolean | null, accumulate?: boolean): void;
```
pawelfus, line 55: Why select can be `null`?

mateuszkornecki: According to the API:
 `When true, the point is selected. When false, the point is unselected. When null or undefined, the selection state is toggled.` 
https://api.highcharts.com/class-reference/Highcharts.Point#select

ts/parts/Chart.ts
```diff
@@ -144,6 +144,7 @@ declare global {
             public getContainer(): void;
             public getMargins(skipAxes?: boolean): void;
             public getSelectedPoints(): Array<Point>;
+            public getSelectedParentNodes(): Array<PackedBubblePoint>;
```
pawelfus, line 147: Don't set it in `Chart.ts`. Instead extend Chart in the file that creates method. For an example check `BubbleLegend.ts`.

ts/parts-more/PackedBubbleSeries.ts
```diff
@@ -1853,9 +1854,7 @@ seriesType<Highcharts.PackedBubbleSeries>(
             const isParentNode = point.isParentNode,
                 parentNodeLayout = this.parentNodeLayout;
             if (isParentNode) {
-                parentNodeLayout.setMaxIterations();
-                parentNodeLayout.currentStep = 0;
-                parentNodeLayout.step();
+                parentNodeLayout.restartSimulation();
```
pawelfus, line 1857: Btw, why do we need to restart the simulation? In theory this is the process should be already defined in `dragNodesMixin`:
- if simulations is running, call `layout.resetSimulation()` to extend the running simulation
- if simulation has ended, call `layout.start()` to start new simulation
- if simulation is disabled, call `chart.redraw()` to run all simulations in sync mode

The above `restartSimulation` seems to be unnecessary?

mateuszkornecki: We need to restart the simulation to fix the parent node's behavior on drag&drop (without that repulsing didn't work correctly when the simulation has ended). We could use the layout.start() as well but to achieve the goal we don't need the whole logic of the start() method, so I extracted just the necessary parts to the restartSimulation().

pawelfus: Ok, I see. Then perhaps we should do this:
- in drag-nodes mixin, instead of using `series.layout` we should use all layouts (check `chart.graphLayoutsStorage`) - look over all available layouts and reset them all
- test `layout.resetSimulation` with all cases (disabled simulation, running simulation and finished simulation)
- `resetSimulation()` should be able to replace current logic for that:
https://github.com/highcharts/highcharts/blob/d1be8b45541bc903e7c3c3cacd010956a3a5db01/ts/modules/networkgraph/draggable-nodes.ts#L138-L156

Izothep: I am totally for this solution. Great idea to put it all in dragNodesMixin.

Izothep:
Good job, thanks! The only concern I have is related to visual tests but they looks to be unrelated to the change in code.

sebastianbochan:
@Izothep its issue also in other PR. You can ignore it.

PR #13259, opened 2020-04-06
bugfix/13226-tooltip-prevents-interaction

### LINKED ISSUE #13226: Tooltip preventing interaction with chart components
#### Expected behaviour
Navigator should be draggable after interacting with tooltip even if tooltip overlaps the navigator

#### Actual behaviour
Top part of navigator is not draggable after interacting with tooltip around navigator

https://jsfiddle.net/Lvtrzna7/

1. Hover over the last point in the chart to display the tooltip
2. Try dragging the navigator by using the right handle. The top part of the navigator handle is not clickable.

TooltipNavigatorIssue

The navigator height in our app is set to be smaller than the default. This is how I found the issue.

Just an observation: I noticed "pointer-events: none" was being set on the group with the .highcharts-tooltip-header class before. If I set it manually in Chrome, this works as expected. 

#### Product version
Highstock v8.0.4 (8.0.3 does not have this behavior)

#### Affected browser(s)
Chrome

Izothep:
Fixed #13226, tooltip prevented interaction on other chart elements.

### DIFF
--- a/ts/parts/SvgRenderer.ts
+++ b/ts/parts/SvgRenderer.ts
@@ -2528,10 +2528,7 @@ extend((
             duration: pick(duration, 150),
             complete: function (): void {
                 // #3088, assuming we're only using this for tooltips
-                elemWrapper.attr({
-                    y: -9999,
-                    visibility: 'hidden' // #13226
-                });
+                elemWrapper.attr({ y: -9999 }).hide();
             }
         });
     },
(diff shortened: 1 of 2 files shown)


ts/parts/SvgRenderer.ts
```diff
@@ -2528,7 +2528,10 @@ extend((
             duration: pick(duration, 150),
             complete: function (): void {
                 // #3088, assuming we're only using this for tooltips
-                elemWrapper.attr({ y: -9999 });
+                elemWrapper.attr({
+                    y: -9999,
+                    visibility: 'hidden' // #13226
```
sebastianbochan, line 2533: I would consider to use `.hide()` instead of inline setting visibility.

Izothep: Definitely, thanks for suggestion!

PR #13263, opened 2020-04-06
 bugfix/13025-auto-margin

### LINKED ISSUE #13025: Issue with auto margin computation
#### Expected behaviour
stockChart to compute the margin in relation to other elements above the plot area

#### Actual behaviour
margin computed is quite large when both the legend and rangeSelector are placed above the plot area

https://jsfiddle.net/5v3L1jch/2/

Product version
Highstock - 8.0.0

Affected browser(s)
Chrome

Note: This can be worked around by setting a margin in userland but I find the auto computed margin to be extremely useful so I'd prefer if we don't have to set the margin manually.

kamilkulig:
Fixed #13025, legend was overlapping range selector when they were both vertically aligned to the top.

### DIFF
--- a/samples/unit-tests/rangeselector/legend-position/demo.js
+++ b/samples/unit-tests/rangeselector/legend-position/demo.js
@@ -1,5 +1,5 @@
 QUnit.test('Legend should be rendered below the range selector', function (assert) {
-    Highcharts.stockChart('container', {
+    var chart = Highcharts.stockChart('container', {
         rangeSelector: {
             enabled: true,
             inputEnabled: false,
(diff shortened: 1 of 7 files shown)


samples/unit-tests/rangeselector/legend-position/demo.js
```diff
@@ -0,0 +1,24 @@
...
+            enabled: true,
+            align: 'left',
+            verticalAlign: 'top'
+        },
+        series: [{
+            data: [101, 343, 11]
+        }]
+    });
+    assert.ok(
+        document.getElementsByClassName('highcharts-legend')[0].getBoundingClientRect().top >=
```
pawelfus, line 20: Don't use `document` - you may get elements from test-templates that are existing in the background. Use `chart.container` instead.

PR #13269, opened 2020-04-07
bugfix/13101-pie-negative-values

### LINKED ISSUE #13101: Cannot read property 'animate' of undefined
Hello.

Case:

1. Piechart with drilldown and with negative value.
2. Call `.select(false, false)` on negative value.

#### Expected behaviour
Unselect point

#### Actual behaviour
Error `Cannot read property 'animate' of undefined`
image

PS: on positive values works.

https://jsfiddle.net/39stjv0x/1/

#### Product version
Highcharts JS v8.0.3 (2020-03-05)

#### Affected browser(s)
Google Chrome 80.0.3987.132

karolkolodziej:
Fixed #13101, pie with negative values threw error on drilldown.

### DIFF
--- a/ts/parts/PieSeries.ts
+++ b/ts/parts/PieSeries.ts
@@ -1413,7 +1413,7 @@ seriesType<Highcharts.PieSeries>(
                 point.options;
 
             if (point.graphic) {
-                (point.graphic as any).animate(this.getTranslate());
+                point.graphic.animate(this.getTranslate());
             }
 
             if (point.shadowGroup) {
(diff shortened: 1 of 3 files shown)


ts/parts/PieSeries.ts
```diff
@@ -1412,7 +1412,9 @@ seriesType<Highcharts.PieSeries>(
             (series.options.data as any)[series.data.indexOf(point)] =
                 point.options;
 
-            (point.graphic as any).animate(this.getTranslate());
+            if (point.graphic) {
+                (point.graphic as any).animate(this.getTranslate());
```
pawelfus, line 1416: Casting `as any` should be unnecessary now

PR #13282, opened 2020-04-09
bugfix/8280-wrong-3d-columns-zIndex

### LINKED ISSUE #8280: 3D column wrong display order with large chart width
#### Expected behaviour
The display order should look like in the case of smaller chart widths.

screen shot 2018-05-04 at 12 03 40

#### Actual behaviour
In my case, if chart width is more than 1223 and the y value is 0, the display order is wrong.

screen shot 2018-05-04 at 12 03 19

https://jsfiddle.net/BlackLabel/8co8yjb9/

#### Product version
Highcharts, 6.1.0

#### Affected browser(s)
All

Izothep:
Fixed #8280, 3D column `zIndex` was wrongly calculated on wide screens.

### DIFF
--- a/js/parts-3d/SVGRenderer.js
+++ b/js/parts-3d/SVGRenderer.js
@@ -198,6 +198,8 @@ element3dMethods = {
         // store original destroy
         elem3d.originalDestroy = elem3d.destroy;
         elem3d.destroy = elem3d.destroyParts;
+        // Store information if any side of element was rendered by force.
+        elem3d.forcedSides = paths.forcedSides;
     },
     /**
      * Single property setter that applies options to each part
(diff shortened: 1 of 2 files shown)


js/parts-3d/SVGRenderer.js
```diff
@@ -274,26 +276,34 @@ cuboidMethods = merge(element3dMethods, {
...
     fillSetter: function (fill) {
-        this.singleSetterForParts('fill', null, {
+        var elem3d = this;
+        elem3d.forcedSides = elem3d.forcedSides || [];
+        elem3d.singleSetterForParts('fill', null, {
             front: fill,
-            top: color(fill).brighten(0.1).get(),
-            side: color(fill).brighten(-0.1).get()
+            // Do not change color if side was forced to render.
+            top: color(fill).brighten(elem3d.forcedSides.includes('top') ? 0 : 0.1).get(),
```
KacperMadej, line 301: @Izothep Sorry for writing after the merge but it came to me just now. Is the usage of `includes` safe? It's not IE friendly and as far as I have checked oldie-polyfill doesn't cover this.

Izothep: I may change it to classic indexOf

PR #13290, opened 2020-04-09
bugfix/11750-series-init-opacity

### LINKED ISSUE #11750: How To: Dynamically set opacity of line series
I am attempting to set the opacity of a line series dynamically - basically I want to "ghost" all of the series on a particular axis when the user clicks a button or ghost an individual series when the user clicks a button.

I have tried the following options
series.update({ opacity: 0.4 })
series.update({ opacity: 0.4 }, true)
series.options.opacity = 0.4 with chart.redraw();

Though I see the opacity value changing in the chart object, the chart does not actually change, as if something else is overriding the opacity when it renders.

What am i doing wrong?  Is it possible to do this dynamically?

#### Product version
Highcharts 7.1.3 with Stock module installed

pawelfus:
Fixed #11750, `series.opacity` option was not respected when creating or updating series.


ts/parts/Series.ts
```diff
@@ -76,6 +76,7 @@ declare global {
             public linkedSeries: Array<Series>;
             public markerGroup?: SVGElement;
             public name: string;
+            public opacity?: number;
```
TorsteinHonsi, line 79: Why do we need this series property? Can't we just read it directly from the options where we use it?

ts/parts/Series.ts
```diff
@@ -6565,15 +6567,25 @@ H.Series = seriesType<Highcharts.LineSeries>(
...
+                    zIndex: zIndex || 0.1 // IE8 and pointer logic use this
+                };
+
+                if (
+                    !this.chart.styledMode &&
+                    this.opacity &&
+                    this.opacity !== 1
+                ) {
+                    newAttrs.opacity = this.opacity;
+                }
```
TorsteinHonsi, line 6585: While reading directly from the options, this seems to do the job:
```js
if (!this.chart.styledMode) {
                    newAttrs.opacity = this.options.opacity;
                }
```

Also, the proposed logic doesn't allow users setting the opacity to 0.

TorsteinHonsi: That said, it seems that neither of the solutions work with `Series.update`?

pawelfus: `this.options.opacity;` was implemented in my first commit, then visual diffs show me that `series.opacity` option for venn and treemap is already used: https://api.highcharts.com/highcharts/series.treemap.opacity

So I decided to use `series.opacity` that can be removed in these series in `series.init`.

But since you are right, `series.update()` doesn't work, I will look for another solution (I guess plot groups are not removed on series.update).

ts/parts/Series.ts
pawelfus: See my explanation below.

pawelfus:
Fixed issue on-update. I decided to stick to `plotGroup()` method. I realized that `zIndex` should have similar problem and it was solved in `series.update()`. Instead of adding more code in `series.update()` I decided to refactor this a bit: `plotGroup()` on init only creates group, and other attributes (visibility, zIndex and opacity) are applied on every call.

It's really tempting to replace `visibility` and `zIndex` arguments with one `attributes={}` argument.

PR #13311, opened 2020-04-14
Fixed JSFiddle with trellis chart example

zhdanova:
I noticed that trellis example in JSFiddle is broken due to loading HighCharts over http instead of https. I forked existing example and updated URL.


TorsteinHonsi:
Thanks! I updated the fiddle, so we don't have to change the URL.

TorsteinHonsi:
.. Also removed the jQuery dependency.

PR #13325, opened 2020-04-16
bugfix/13136-tickAmount-small-chart-height

### LINKED ISSUE #13136: tickAmount don't works when chart size is small
#### Expected behaviour
tickAmount working.

#### Actual behaviour
tickAmount don't changes nothing.

https://jsfiddle.net/bsqn1v0u/1/

See how there's no 3 ticks, but only 2. But if you change the height to 400px you will get 3 ticks (or whatever amount you put, like 100 ticks, so is not a density problem).

#### Product version
Latest

#### Affected browser(s)
Chrome

karolkolodziej:
Fixed #13136, couldn't set `yAxis.tickAmount` when the chart size was small.

~~Fixed #13136, didn't able to set the tick amount when the chart size was small.~~

### DIFF
--- a/ts/parts/Axis.ts
+++ b/ts/parts/Axis.ts
@@ -5794,11 +5794,11 @@ class Axis implements AxisComposition {
         var axis: Highcharts.Axis = this as any,
             options = this.options,
             tickAmount = options.tickAmount,
-            tickPixelInterval = options.tickPixelInterval;
+            tickPixelInterval = options.tickPixelInterval as number;
 
         if (
             !defined(options.tickInterval) &&
-            this.len < (tickPixelInterval as any) &&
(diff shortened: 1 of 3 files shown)


ts/parts/Axis.ts
```diff
@@ -5794,11 +5794,11 @@ class Axis implements AxisComposition {
         var axis: Highcharts.Axis = this as any,
             options = this.options,
             tickAmount = options.tickAmount,
-            tickPixelInterval = options.tickPixelInterval;
+            tickPixelInterval = options.tickPixelInterval as number;
```
TorsteinHonsi, line 5797: We should leave this as `any` for now. Later on, we will fix typing for the options (using Partial and DeepPartial), so that this option is always a number.

PR #13335, opened 2020-04-17
bugfix/13202-lollipop-data-format

### LINKED ISSUE #13202: The lollipop series - data issue
The suggested data configurations options in the [documentation](https://api.highcharts.com/highcharts/series.lollipop.data) don't work in the lollipop series. The only one which works correctly is an array of y values:

`data: [0, 5, 3, 5]`

The array of objects or an array of arrays doesn't work. 

Demo: https://jsfiddle.net/BlackLabel/bgtLk79s/

#### Product version

raf18seb:
Fixed #13202 , lollipop series didn't work with all data formats.

@bre1470 could you suggest how to avoid `any` in this commit?
**Background:** lollipop series (one-dimension data format, ['y']) inherits from dumbbell series (two-dimension ['low', 'high']).

### DIFF
--- a/js/modules/lollipop.src.js
+++ b/js/modules/lollipop.src.js
@@ -56,6 +56,11 @@ seriesType('lollipop', 'dumbbell', {
         pointFormat: '<span style="color:{series.color}">●</span> {series.name}: <b>{point.low}</b><br/>'
     }
 }, {
+    pointArrayMap: ['y'],
+    pointValKey: 'y',
+    toYData: function (point) {
+        return point.y;
+    },
     translatePoint: areaProto.translate,
     drawPoint: areaProto.drawPoints,
     drawDataLabels: colProto.drawDataLabels,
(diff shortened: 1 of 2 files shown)


bre1470:
> @bre1470 could you suggest how to avoid `any` in this commit?
> **Background:** lollipop series (one-dimension data format, ['y']) inherits from dumbbell series (two-dimension ['low', 'high']).

It depends on how the types differ. You can always reduce the type union in overloads.

```ts
declare class Series {
  public toYData(): (number|string);
}
declare class MySeries extends Series {
  public toYData(): number;
}
```

In your case though the original return type is a tuple, which is a really narrowed array type. You could change `public toYData(point: AreaRangePoint): [number, number];` to `public toYData(point: AreaRangePoint): Array<number>;` to work around the limitation (actually you do not need to overload then anymore).

Otherwise TypeScript is correct in saying, that Lollipop series can never be a 100% AreaRange series. `any` is then the only options.

raf18seb:
Thanks @bre1470 ! I will look into that soon.

I would also like to discuss one thing - whether we should keep the old naming `data: [{low: 7}, {low: 10}]` or we should change it as I did above to `data: [{y: 7}, {y: 10}]` to keep it consistent with other one-dimension series.

@pawelfus suggested that it should be changed to `y` (demos must be changed too) and users who have already been using lollipop series will need to change the data config if they want to use the newest Highcharts version.
@sebastianbochan @TorsteinHonsi what is your opinion?

TorsteinHonsi:
> Should we leave low or change it to y?

I agree with both. It should be called y, but we can't just change it, breaking it for current users. We need to make a compatibility snippet. Probably an override to `Point.applyOptions` is the appropriate place, to catch it as early as possible. Or it may even be in `init`, because we only need to handle object options. Something like this. Make sure `low: null` is also handled, therefor using `low in options`:

```js
if (isObject(options) && low in options) {
    options.y = options.low;
    delete options.low;
}
```

PR #13350, opened 2020-04-21
bugfix/13287-wrong-vbp-position

### LINKED ISSUE #13287: Volume by price indicator isn't working
The VBP indicator is not working.

Open, below demo - VBP is not displayed. When moving navigator, it is displayed for a while, then throws many errors into the console.

https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/stock/indicators/volume-by-price

#### Product version
Highstock 8.0.4

FYI @PaulDalek

PaulDalek:
Fixed #13287, incorrect position of the `vpb` (volume by price) series after the initial animation.

### DIFF
--- a/ts/indicators/volume-by-price.src.ts
+++ b/ts/indicators/volume-by-price.src.ts
@@ -361,10 +361,16 @@ seriesType<Highcharts.VBPIndicator>(
             init: boolean
         ): void {
             var series = this,
-                attr: Highcharts.SVGAttributes = {};
+                inverted = series.chart.inverted,
+                attr: Highcharts.SVGAttributes = {},
+                translate,
+                position;
 
             if (!init) {
-                attr.translateX = series.yAxis.pos;
+                translate = inverted ? 'translateY' : 'translateX';
+                position = inverted ? series.yAxis.top : series.xAxis.left;
+                (series.group as any)['forceAnimate:' + translate] = true;
+                attr[translate] = position;
                 (series.group as any).animate(
                     attr,
                     extend(animObject(series.options.animation), {
--- a/ts/indicators/volume-by-price.src.ts
+++ b/ts/indicators/volume-by-price.src.ts
@@ -362,7 +362,7 @@ seriesType<Highcharts.VBPIndicator>(
         ): void {
             var series = this,
                 inverted = series.chart.inverted,
-                group = series.group as Highcharts.SVGElement,
+                group = series.group,
                 attr: Highcharts.SVGAttributes = {},
                 translate,
                 position;
--- a/ts/indicators/volume-by-price.src.ts
+++ b/ts/indicators/volume-by-price.src.ts
@@ -362,6 +362,7 @@ seriesType<Highcharts.VBPIndicator>(
         ): void {
             var series = this,
                 inverted = series.chart.inverted,
+                group = series.group as Highcharts.SVGElement,
                 attr: Highcharts.SVGAttributes = {},
                 translate,
                 position;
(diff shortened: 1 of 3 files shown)


ts/indicators/volume-by-price.src.ts
```diff
@@ -361,10 +361,16 @@ seriesType<Highcharts.VBPIndicator>(
...
+                inverted = series.chart.inverted,
+                attr: Highcharts.SVGAttributes = {},
+                translate,
+                position;
 
             if (!init) {
-                attr.translateX = series.yAxis.pos;
+                translate = inverted ? 'translateY' : 'translateX';
+                position = inverted ? series.yAxis.top : series.xAxis.left;
+                (series.group as any)['forceAnimate:' + translate] = true;
```
sebastianbochan, line 372: What is the reason of using `series.group as any` instead of `SVGElement` ?

PaulDalek: Corrected. Thanks for noticing!

ts/indicators/volume-by-price.src.ts
```diff
@@ -361,11 +361,18 @@ seriesType<Highcharts.VBPIndicator>(
             init: boolean
         ): void {
             var series = this,
-                attr: Highcharts.SVGAttributes = {};
+                inverted = series.chart.inverted,
+                group = series.group as Highcharts.SVGElement,
```
TorsteinHonsi, line 365: `Series.group` is typed as `(Highcharts.SVGElement|undefined)`. By casting this to `SVGElement`, we're ignoring the possibility that it can be undefined, which would cause failure below. The more robust approach would probably be that line 370 read `if (!init && group) {`, then everything is properly guarded.

The only "problem" is that we know that `group` is always defined at this point. It is only before the first render, and after destruction, it may be undefined. Any idea for how we should handle such cases @bre1470 ? It is a recurring pattern.

sebastianbochan: Its more complicated then I expected. 
Maybe we should consider different type or union?

bre1470: The official approach is to initialize everything during the construct cycle. With the refactoring we sometimes do an `void 0 as any` assignment, if the init -> render chain takes care of the property value:
```ts
class Series {
    public group: SVGElement = void 0 as any; // dirty!!!
    public render(): void {
        group = this.renderer.createElement('g');
    }
}
```

IMHO the best solution would be `Object.define` support in all our browsers, so we can make use of getter/setter syntax in classes, which would solve all the hassle by automatically creating an initial instance for a property if not set yet:
```ts
class Series {
    private _group?: SVGElement;
    public get group(): SVGElement {
        if (!this._group) {
           this._group = this.renderer.createElement('g');
        }
        return this._group;
    }
}
```

TorsteinHonsi: > which would solve all the hassle by automatically creating an initial instance for a property if not set yet

A bit like the `getLabel` function of the tooltip, that creates the label if it isn't already defined. But this would produce a ton of extra code if we were to implement it consistently. So I guess, at least in this case we just need to check for `group`. 

`if (!init && group) {`

There's no doubt it is also the safest, as there is, at least in theory, a possibility that this function can run before initialization or after descruction.

ts/indicators/volume-by-price.src.ts
```diff
@@ -361,11 +361,18 @@ seriesType<Highcharts.VBPIndicator>(
...
-            if (!init) {
-                attr.translateX = series.yAxis.pos;
-                (series.group as any).animate(
+                inverted = series.chart.inverted,
+                group = series.group as Highcharts.SVGElement,
+                attr: Highcharts.SVGAttributes = {},
+                translate,
+                position;
+
+            if (!init && group) {
```
TorsteinHonsi, line 370: Now that we check the group, we can also remove the cast:

`group = series.group as Highcharts.SVGElement,` 
=>
`group = series.group,`

PaulDalek: Ahh, yes. Corrected, thanks!

PR #13364, opened 2020-04-23
bugfix/13330-stacklabels-backgroundColor-issue

### LINKED ISSUE #13330: Background color styling options for stackLabels and dataLabels
#### Expected behaviour
stackLabels and dataLabels should share styling options and allow setting the background color from the API.

#### Actual behaviour
The documentation shows that we are able to change the background color and text color for dataLabels using options such as backgroundColor and color. However, when we want to use the dataLabels.style, we are deprived of the option of giving the background color:

https://jsfiddle.net/BlackLabel/rdzp7y5x/
https://jsfiddle.net/BlackLabel/Lhvud1pe/

When there is a need to stylize the background color for stackLabels, we are not able to do it in any of the available ways as for dataLabels:

https://jsfiddle.net/BlackLabel/8ugay6sn/

#### Workaround with formatter and useHTML:
https://jsfiddle.net/BlackLabel/ot6zqman/

#### Product version
v8.0.4

Basts:
Added box options for `yAxis.stackLabels`: `backgroundColor`, `borderWidth`, `borderRadius` and `borderColor`. See #13330.

~~Fixed #13330, adding the styling options, like: backgroundColor, borderWidth, borderRadius and borderColor for stack labels.~~

### DIFF
--- a/js/parts/Axis.js
+++ b/js/parts/Axis.js
@@ -5711,6 +5711,36 @@ var Axis = /** @class */ (function () {
              */
             allowOverlap: false,
             /**
+             * The background color or gradient for the stack label.
+             *
+             * @type      {Highcharts.ColorString|Highcharts.GradientColorObject|Highcharts.PatternObject}
+             * @since     next
+             * @apioption yAxis.stackLabels.backgroundColor
+             */
+            /**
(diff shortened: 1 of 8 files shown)


sebastianbochan:
Looks good!

PR #13367, opened 2020-04-24
bugfix/treegrid-module

bre1470:
Fixed dependency issue in TreeGrid module.


goransle:
Nodes are in the right place, but now they won't collapse: https://jsfiddle.net/goransle/kv8ux79y/

bre1470:
This is now fixed, too.

PR #13377, opened 2020-04-27
Fixed dead link to Download section.

Denyllon:


bre1470:
Maybe @goransle has some input regarding this.

goransle:
Without `/blog/` is preferred. (I think the plan is is to move these pages when the front page transitions to Wordpress as well).

Issue #13166, opened 2020-03-19
Unable to internationalize "Category", and "DateTime" words in export-data


gtanczyk:
#### Expected behaviour
I would like to be able to internationalize all contents of XLS export

#### Actual behaviour
It's not possible.

https://github.com/highcharts/highcharts/blob/6cae030f630f092d4de845dea4e2e054a32d100a/js/modules/export-data.src.js#L308

sebastianbochan:
HI @gtanczyk,
Thank you for reporting about the problem. I agree that we can extract it to the lang object.

raf18seb:
@sebastianbochan I think we should take it from xAxis title if defined instead of `lang`. See multiple x-axes: https://jsfiddle.net/BlackLabel/7r9js4fw

sebastianbochan:
I agree that we can extract all static labels from data module to the lang object. 
@oysteinmoseng what do you think?

oysteinmoseng:
Agree 100% that we should put all static text in `lang` - and also agree with @raf18seb that it could be useful to add a note in the API somewhere (or in the tutorial docs for the export data module) on how the table is built.

Looking at the current `lang` options, we might want to pull the export-related keys into a new `lang.exporting` option group, and add deprecation/compatibility fixes. Not sure how many new keys the export data module would add. @TorsteinHonsi might have further thoughts on that

PR #13383, opened 2020-04-27
enhancement/13166-export-data-lang

karolkolodziej:
Internationalized export-data table text, see #13166.

### DIFF
--- a/js/modules/export-data.src.js
+++ b/js/modules/export-data.src.js
@@ -237,18 +237,18 @@ Highcharts.setOptions({
         /**
          * The text for exported table.
          *
-         * @since    8.0.4
+         * @since    next
          * @requires modules/export-data
          */
         exportData: {
             /**
              * The category column title.
              */
-            categoryColumnTitle: 'Category',
+            categoryHeader: 'Category',
             /**
              * The category column title when axis type set to "datetime".
              */
-            categoryColumnTitleDT: 'DateTime'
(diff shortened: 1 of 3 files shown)


js/modules/export-data.src.js
```diff
@@ -234,6 +234,22 @@ Highcharts.setOptions({
          * @requires modules/export-data
          */
         downloadXLS: 'Download XLS',
+        /**
+         * The text for exported table.
+         *
+         * @since    8.0.4
```
sebastianbochan, line 240: use `@since next`

js/modules/export-data.src.js
```diff
@@ -234,6 +234,22 @@ Highcharts.setOptions({
...
+         * @requires modules/export-data
+         */
+        exportData: {
+            /**
+             * The category column title.
+             */
+            categoryColumnTitle: 'Category',
```
sebastianbochan, line 247: I would consider to shorter name of param ie: `categoryHeader`.

js/modules/export-data.src.js
```diff
@@ -234,6 +234,22 @@ Highcharts.setOptions({
...
+            /**
+             * The category column title when axis type set to "datetime".
+             */
+            categoryColumnTitleDT: 'DateTime'
```
sebastianbochan, line 251: DT is not intuitive. 

I would consider to shorter name of param ie: `categoryDatetimeHeader`.

PR #13388, opened 2020-04-28
bugfix/12804-back-button-sunburst

### LINKED ISSUE #12804: Sunburst incorrectly shows back button after drill up
#### Expected behaviour
After drilling down to a specific layer, the back button should only be displayed until the user clicks it the number of layers that they drilled down to

#### Actual behaviour
Back button always shows one layer more than it should. For example, after clicking to the second layer, the back button should only need to be clicked one time before it disappears. Instead, it needs to be clicked two times in order for it to disappear. 

https://www.highcharts.com/demo/sunburst

1. Click on Asia. The sunburst should drill down to the Asia section of the chart
2. Click the back button. The sunburst drills up to the full chart. At this time, since the entire chart is being displayed, the back button should not still be shown. Instead, it needs to be clicked one extra time in order to disappear. 

#### Product version
Highcharts 8.0.0

#### Affected browser(s)
Chrome

kamilkulig:
Fixed #12804, in sunburst charts, the drill-up button was visible after drilling up to the top level.

~~Fixed #12804: drillup button was visible after drilling up to parent.~~

### DIFF
--- a/ts/modules/sunburst.src.ts
+++ b/ts/modules/sunburst.src.ts
@@ -202,7 +202,13 @@ var CenteredSeriesMixin = H.CenteredSeriesMixin,
     rad2deg = 180 / Math.PI,
     seriesTypes = H.seriesTypes,
     setTreeValues = mixinTreeSeries.setTreeValues,
-    updateRootId = mixinTreeSeries.updateRootId;
+    updateRootId = function (series: any): string {
+        var rootId = mixinTreeSeries.updateRootId(series),
+            nodeMap = series.nodeMap;
+        // In sunburst the first level should be treated the same as
+        // the '' level (#12804).
+        return nodeMap && nodeMap[rootId].parent === '' ? '' : rootId;
+    };
 
 // TODO introduce step, which should default to 1.
 var range = function range(from: unknown, to: unknown): Array<number> {
--- a/ts/modules/sunburst.src.ts
+++ b/ts/modules/sunburst.src.ts
@@ -202,13 +202,7 @@ var CenteredSeriesMixin = H.CenteredSeriesMixin,
     rad2deg = 180 / Math.PI,
     seriesTypes = H.seriesTypes,
     setTreeValues = mixinTreeSeries.setTreeValues,
-    updateRootId = function (series: any): string {
-        var rootId = mixinTreeSeries.updateRootId(series),
-            nodeMap = series.nodeMap;
-        // In sunburst the first level should be treated the same as
-        // the '' level (#12804).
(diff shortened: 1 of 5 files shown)


ts/modules/sunburst.src.ts
```diff
@@ -202,7 +202,13 @@ var CenteredSeriesMixin = H.CenteredSeriesMixin,
     rad2deg = 180 / Math.PI,
     seriesTypes = H.seriesTypes,
     setTreeValues = mixinTreeSeries.setTreeValues,
-    updateRootId = mixinTreeSeries.updateRootId;
+    updateRootId = function (series: any): string {
```
TorsteinHonsi, line 205: How does this differ from the treemap series? In stead of an override, can we come up with a general solution in the mixin itself? I don't have the answer, just asking... Maybe the same issue also applies to treemap series, but we can't see it?

kamilkulig: I analyzed the problem again and it turned out that my previous solution is not right.
The problem here is that the sunburst chart looks exactly the same on the level 0 (represented in code as `“”`) and level 1 but only if there’s only one element on level 1.
If there are at least two children on level 1 everything works fine: http://jsfiddle.net/BlackLabel/8vt5rqy7/ 

So, I think that the fix here is to handle the scenario when there’s only 1 element on level 1. Here’s a demonstration version of that fix - hiding traverse up button: http://jsfiddle.net/BlackLabel/0dv35mfr/
```
    if (rootId === '' ||
      (series instanceof H.seriesTypes.sunburst &&
        rootId === series.tree.children[0].id && series.tree.children.length === 1)) {
```
Of, course the code should be placed in sunburst module (not treemap) so some refactoring of `renderTraverseUpButton `method is needed.

The blinking effect you mentioned is caused by the fact that the sunburst switches between the level 0 and 1 (and they both look the same).

TorsteinHonsi: Thanks! Is this ready to merge now? It seems that the PR was marked ready for review, but the mentioned changes are not committed. Did you forget to push?

kamilkulig: Not ready yet. Just wanted to make sure that my idea is good.  I'm going to commit probably today or tomorrow.

TorsteinHonsi: Yes I think the idea is good. But maybe, for consistency, we can have the same situation in tree maps? What if we have a tree map with 1 item on level 1? Maybe it is enough just to look for the number of items on level 1, regardless of the series type?

PR #13399, opened 2020-04-29
bugfix/13177-zones-update-color

### LINKED ISSUE #13177: Update of column series color does not work if zones are used
#### Expected behaviour
Calling series.update() with a different color than what the series was created with, will update the column colors and point marker colors appropriately.

#### Actual behaviour
Calling series.update() with a different color does not update the column color if zones are used.  It appears that the individual point colors and marker colors are not updated if zones are used, even for line charts.

https://jsfiddle.net/kevinfernandes/2j7xLnz9/

This is forked from one of your column chart demos.

Clicking the "Color" button on the bottom of the chart updates the legend with the new colors, and the line updates as well, but the columns and the point markers on the line do not.

If you comment out the zones on each series and run it again, the points and columns update as expected.

#### Product version
Found in version 7.2.1
Not sure what version the demo is using.

#### Affected browser(s)
Tested in Chrome v80

Izothep:
Fixed #13177, update was not changing points color when zones were applied.

### DIFF
--- a/js/parts/Point.js
+++ b/js/parts/Point.js
@@ -706,6 +706,8 @@ var Point = /** @class */ (function () {
      */
     Point.prototype.resolveColor = function () {
         var series = this.series, colors, optionsChart = series.chart.options.chart, colorCount = optionsChart.colorCount, styledMode = series.chart.styledMode, colorIndex;
+        // remove points nonZonedColor for later recalculation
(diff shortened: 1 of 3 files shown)


TorsteinHonsi:
Is it ready?

Izothep:
Yes, I believe it is. The fix itself was pretty simple.

PR #13413, opened 2020-05-01
TypeScript: Stricter typing for Axis.tickSize

TorsteinHonsi:
It is now a tuple|undefined, and properly guarded.


js/parts/Axis.js
```diff
@@ -2037,11 +2037,11 @@ var Axis = /** @class */ (function () {
      * @param {string} [prefix]
      * 'tick' or 'minorTick'
      *
-     * @return {Array<number>}
+     * @return {[number, number]|undefined}
```
bre1470, line 2040: JSDoc types have to be written without spaces.

PR #13429, opened 2020-05-04
bugfix/13074-uniqueSerial

### LINKED ISSUE #13074: change static method uniqueKey not working
before version 8.0.2
changing the method worked (in order to Overrides dynamic id generation for testing)
```
Highcharts.uniqueKey = (function() {
  var idCounter = 0;
  return function() {
    return 'highcharts-' + idCounter++;
  };
})();

HighchartsMore(Highcharts);
```
now it is not changing anything

bre1470:
Added `Highcharts.serialKeys` option, to generate `Highcharts.uniqueKeys` suitable for automatic tests.

Close #13074, Added H.serialKeys mode for H.uniqueKey.


bre1470:
`serial` means 1, 2, 3... etc., see https://en.wikipedia.org/wiki/Serial_number

`Highcharts.useUniqueIds` is imho confusing, because if you disable it, Highcharts will still use the `Highcharts.uniqueKey` function, just with a serial result. How about `useSerialKey`?

TorsteinHonsi:
Hmm, yes that makse sense. I think we need to take one step back and look at it with an implementer's eye. Highcharts currently applies unique id's to any element in any chart across a page, so that in a setup with multiple charts, no id will be the same. 

With the new setting, it will create the same serial ids for any two charts on the same page. So the id's will no longer be unique. And this all applies to the id's of elements, which is a more precise description than `key`. What about `useSerialIds`? Then elaborate a bit in the description that this only makes sense when comparing the generated SVG of two or more charts.

TorsteinHonsi:
Okay, sorry to nag, but ... why a setter function? Can't we just do it like in the previous commit?
`const hash = H.useSerialIds ? ... : ...`

bre1470:
ES6 imports are readonly. So TypeScript allows something like `H.charts.push(…)` but not `H.charts = …`.

bre1470:
I considered as an alternative to have the flag property on the uniqueKey function `H.uniqueKey.useSerialIds`, but that would be bad practice and hard to document properly. So I went with a simple set/get-like function.

PR #13444, opened 2020-05-06
bugfix/13375-plotbands-events-not-working

### LINKED ISSUE #13375: plotBand events not working if it was added with the api addPlotBand()
If the chart is initialized without a plotBand option within the xAxis option, no events are attached if it get's added at a later point.
If there is an existing plotBand option, the event's will be attached, the plotBand option must not include any plotBand, just the option needs to exist.

Problem is that `isNew` is false on second loop without an existing plotBand option.
I checked this with:
`console.log('create plot', isNew, options.id, path);`

before

```
// common for lines and bands
if ((isNew || !svgElem.d) && path && path.length) {
```

If the `plotBand` option **does not** exists, isNew is within the first call `true` and on second `false`
If the `plotBand` option **does** exists, on first and second isNew is `true`

#### Expected behaviour

Events should get attached correctly.

#### Actual behaviour

`if ((isNew || !svgElem.d) && path && path.length) {` is false and the events are not attached.

http://jsfiddle.net/d2k6r1bs/13/

Saved version has the issue, uncomment the plotBand option to get the expected behaviour

#### Product version

#### Affected browser(s)
All, (Verified in Firefox + Chrome in Win + Linux, IE 11 + Edge (not the chrome version))

Izothep:
Fixed #13375, events were not added to dynamically created plot bands and lines.

### DIFF
--- a/js/parts/PlotLineOrBand.js
+++ b/js/parts/PlotLineOrBand.js
@@ -126,8 +126,8 @@ var PlotLineOrBand = /** @class */ (function () {
             return;
         }
         // common for lines and bands
-        if ((isNew || !svgElem.d) && path && path.length) {
-            svgElem.attr({ d: path });
+        // Add events only if they were not added before.
+        if (!svgElem.eventsAdded) {
             // events
             if (events) {
                 objectEach(events, function (event, eventType) {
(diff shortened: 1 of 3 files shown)


Izothep:
Actually, I want to add one more fix in this PR, I am converting it to draft

js/parts/PlotLineOrBand.js
```diff
@@ -106,7 +106,7 @@ var PlotLineOrBand = /** @class */ (function () {
              * @name Highcharts.PlotLineOrBand#svgElement
              * @type {Highcharts.SVGElement}
              */
-            plotLine.svgElem = svgElem = renderer
+            plotLine.svgElem = svgElem = svgElem || renderer
```
Izothep, line 109: ok, its a relict of different solution, additional check might be removed.

PR #13453, opened 2020-05-06
enhancement/deprecate

bre1470:
Added deprecation error.

Follow up to PR https://github.com/highcharts/highcharts/pull/13224.


oysteinmoseng:
@bre1470 Since all this is doing is introducing an error code (unless I misunderstood), the only thing that would have to change is the `warn` function in `accessibility/options/deprecatedOptions.ts`, which already uses the `H.error` function.

TorsteinHonsi:
Please perform Øystein's suggestion and merge the master in here, then this PR is ready to merge.

PR #13459, opened 2020-05-07
bugfix/13180-axis-before-tick

### LINKED ISSUE #13180: Migrate Axis.beforeSetTickPositions to event system
This should be adjusted to make use of events.

bre1470:
Fixed #13180, moved beforeSetTickPositions to events.

### DIFF
--- a/js/parts/Axis.js
+++ b/js/parts/Axis.js
@@ -1431,9 +1431,7 @@ var Axis = /** @class */ (function () {
         // set the translation factor used in translate function
         axis.setAxisTranslation(true);
         // hook for ordinal axes and radial axes
-        if (axis.beforeSetTickPositions) {
-            axis.beforeSetTickPositions();
-        }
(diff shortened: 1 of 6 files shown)


js/parts/Axis.js
```diff
@@ -1431,9 +1431,7 @@ var Axis = /** @class */ (function () {
         // set the translation factor used in translate function
         axis.setAxisTranslation(true);
         // hook for ordinal axes and radial axes
-        if (axis.beforeSetTickPositions) {
-            axis.beforeSetTickPositions();
-        }
+        fireEvent(this, 'setTickInterval');
         // hook for extensions, used in Highstock ordinal axes
         if (axis.ordinal) {
```
TorsteinHonsi, line 1436: Since this block always runs right after `beforeSetTickPositions`, and the ordinal axis already has a hook for that, can we just move this into the same hook?

PR #13472, opened 2020-05-08
bugfix/13379-item-series-no-data 

### LINKED ISSUE #13379: Crash in item chart with no or empty data in horizontal/vertical layout
#### Expected behaviour
No crash when series object does not hold data or data is empty.

#### Actual behaviour
Crash when series object has no data property, or data is empty

https://jsfiddle.net/axnu0smv/

* Run -> crash
* Change chart type to `pie` -> works
* Change layout to circular by defining `startAngle` and `endAngle` -> works

#### Product version
8.0.4

#### Affected browser(s)
Tested in Chromium

karolkolodziej:
Fixed #13379, item series with no data threw an error.


ts/modules/item-series.src.ts
```diff
@@ -199,6 +199,9 @@ seriesType<Highcharts.ItemSeries>(
     {
         markerAttribs: void 0,
         translate: function (this: Highcharts.ItemSeries): void {
+            if (this.total === 0) {
```
TorsteinHonsi, line 202: Add a short comment on why we are doing this, and a reference to the issue number.

samples/unit-tests/series-item/dynamic/demo.js
```diff
@@ -39,3 +39,16 @@ QUnit.test('Item series dynamics', assert => {
         'Point length modified'
     );
 });
+
+QUnit.test('Errors in the console during initialization with no data should not be visible, #13379.', assert => {
```
TorsteinHonsi, line 43: This test is good, but I'm thinking we should take this opportunity to go further, and make a general unit test so that we don't do this error again.

We should have a generic test that checks that all series types behave nicely with 
- missing `series` option
- `series: []`
- missing `data` option
- `data: []`

See https://github.com/highcharts/highcharts/blob/v8.1.0/samples/unit-tests/styled-mode/series-types/demo.js#L45-L121 for inspiration. We can create this test in `unit-tests/series/seriestypes`

PR #13479, opened 2020-05-11
Refactor renderer.label into new class SVGLabel

TorsteinHonsi:
The new class inherits SVGElements instead of overriding methods per instance.


ts/parts/SVGLabel.ts
```diff
@@ -0,0 +1,573 @@
...
+            public text: SVGElement;
+            public textAlign: string;
+            public textStr?: string;
+            public width: number;
+            public widthSetting?: number;
+            public x?: number;
+            public xSetting: number;
+            public y?: number;
+            public ySetting: number;
+        }
```
bre1470, line 69: This should be replaced with `let SVGLabel: SVGLabel;` as it is only for backwards compatibility. Instead use `import type SVGLabel from './SVGLabel';` in other files to reference the type and save real imports.

ts/parts/SVGLabel.ts
```diff
@@ -0,0 +1,573 @@
...
+
+    public ySetter(value: number): void {
+        this.ySetting = this.y = Math.round(value);
+        this.attr('translateY', this.ySetting);
+    }
+}
+
+H.SVGLabel = SVGLabel;
+
+export default H.SVGLabel;
```
bre1470, line 573: Here export SVGLabel without `H.` as we do not need to be backward compatible.

ts/parts/SVGLabel.ts
```diff
@@ -0,0 +1,573 @@
...
+        }
+    }
+}
+
+/**
+ * @private
+ */
+interface SVGLabel extends Highcharts.SVGLabel {
+    // takes interfaces from internal namespace
+}
```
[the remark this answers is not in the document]
TorsteinHonsi: It doesn't compile if I remove it...

ts/parts/SVGLabel.ts
```diff
@@ -0,0 +1,573 @@
...
+        y?: number,
+        shape?: Highcharts.SymbolKeyValue | string,
+        anchorX?: number,
+        anchorY?: number,
+        useHTML?: boolean,
+        baseline?: boolean,
+        className?: string
+    ) {
+        super();
+        this.init(renderer, 'g');
```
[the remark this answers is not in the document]
TorsteinHonsi: I think we need to consider that later, as it requires a major change to the base SVGElement.

ts/parts/SVGElement.ts
```diff
@@ -416,8 +398,8 @@ class SVGElement {
         this.parentInverted = parent && (parent as any).inverted;
 
         // build formatted text
-        if (typeof this.textStr !== 'undefined') {
-            renderer.buildText(this as any);
+        if (typeof this.textStr !== 'undefined' && !(this instanceof H.SVGLabel)) {
```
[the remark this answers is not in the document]
TorsteinHonsi: This causes a runtime error if I change it. There will be a circular dependency if I import `SVGLabel` at the top of `SVGElement.ts`, as `SVGLabel` depends of `SVGElement`.

ts/parts/SVGLabel.ts
TorsteinHonsi: Could I ask you to have a look at this?

ts/parts/SVGLabel.ts
```diff
@@ -0,0 +1,548 @@
...
+        str: string,
+        x: number,
+        y?: number,
+        shape?: Highcharts.SymbolKeyValue | string,
+        anchorX?: number,
+        anchorY?: number,
+        useHTML?: boolean,
+        baseline?: boolean,
+        className?: string
+    ) {
```
pawelfus, line 58: Did you consider decreasing number of params? Either, `label(renderer, string, position, options)` or simple `label(renderer, options)` ?

Perhaps `anchorX/Y` could be removed from init (unless these options are crucial when initializing) and moved to `setAnchor()` method?

ts/parts/SVGLabel.ts
```diff
@@ -0,0 +1,548 @@
+/* *
+ *
+ *  (c) 2010-2020 Torstein Honsi
```
pawelfus, line 3: Not related to the PR itself, just a note: can we automate years? For example `annotations` have 2009-2017 (?) :) For example `@product.copyrightYears@`

ts/parts/SVGLabel.ts
```diff
@@ -0,0 +1,548 @@
...
+    public kind: string = 'label';
+
+    /**
+     * For labels, these CSS properties are applied to the `text` node directly.
+     *
+     * @private
+     * @name Highcharts.SVGLabel#textProps
+     * @type {Array<string>}
+     */
+    public textProps: Array<string> = [
```
pawelfus, line 119: How about making this one a public static array?

ts/parts/SVGLabel.ts
```diff
@@ -0,0 +1,548 @@
...
+        'textOutline', 'textOverflow', 'width'
+    ]
+
+    /* *
+     *
+     *  Functions
+     *
+     * */
+
+    public alignSetter(value: 'left' | 'center' | 'right'): void {
```
pawelfus, line 131: How about using [`H.AlignValue`](https://api.highcharts.com/class-reference/Highcharts.html#.AlignValue) instead of `string|string|string` ?

ts/parts/SVGLabel.ts
```diff
@@ -0,0 +1,548 @@
...
+
+        // Release local pointers (#1298)
+        /*
+        wrapper =
+        renderer =
+        text =
+        updateBoxSize =
+        updateTextPadding =
+        boxAttr = null as any;
+        */
```
pawelfus, line 239: Is this comment necessary?

ts/parts/SVGLabel.ts
TorsteinHonsi: Yes I considered that, but forgot. I think I'll do that. Makes it more future proof.

ts/parts/SVGLabel.ts
TorsteinHonsi: Yes we should. I don't think we need a variable for it though. We can have the build process look for actual touch dates in the files for example. I'm closing this comment and we can pick it up in a separate issue.

ts/parts/SVGLabel.ts
TorsteinHonsi: I think this will make sense:
`label(string, x, y, options)` 

Then it can be aligned directly with `text(string, x, y, options)`. 

Will create a separate PR for that, so we have better comparison.

PR #13498, opened 2020-05-12
bugfix/13342-tooltip-position-update-height

### LINKED ISSUE #13342: Tooltip does not to properly pop up when changing chart height dynamically since Chrome 81
### Expected behaviour

After dynamically increasing chart height (say, from 0.1 to 400), tooltip should still pop up not only when the mouser cursor is on the line but also when the cursor is on the empty space in a chart.

### Actual behaviour

**After upgrading from chrome 80 to chrome 81,**
Tooltip does not pop up as you move the mouse cursor on the empty space in the chart, but it does pop up if you move the cursor onto the line.

### Live demo with steps to reproduce
[Reproduction in CodeSandbox](https://codesandbox.io/s/highcharts-vue-demo-93b6i)

**Google Chrome 80.0.3987.149**
chrome80

**Google Chrome 81.0.4044.113**
chrome81

### Product version
* highcharts: 8.0.3
* highcharts-vue: 1.3.5

### Affected browser(s)
Chrome 81

sebastianbochan:
Fixed #13342, tooltip tracking was broken after height update.

### DIFF
--- a/ts/parts/Chart.ts
+++ b/ts/parts/Chart.ts
@@ -2652,8 +2652,12 @@ class Chart {
     public updateContainerScaling(): void {
         const container = this.container;
         if (
-            container.offsetWidth &&
-            container.offsetHeight &&
+            /**
+             * #13342 - tooltip was not visible in Chrome, when chart
+             * updates height.
+             */
+            container.offsetWidth > 2 &&
+            container.offsetHeight > 2 &&
             container.getBoundingClientRect
         ) {
             const bb = container.getBoundingClientRect(),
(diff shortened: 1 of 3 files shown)


ts/parts/Chart.ts
```diff
@@ -2652,8 +2652,12 @@ class Chart {
     public updateContainerScaling(): void {
         const container = this.container;
         if (
-            container.offsetWidth &&
-            container.offsetHeight &&
+            /**
```
TorsteinHonsi, line 2655: Use the `//` syntax for inline comments. `/**` signifies a doclet.

sebastianbochan:
I committed your suggestions. The PR is ready to merge.
@TorsteinHonsi

PR #13501, opened 2020-05-12
Crashes when shadowOptions is not an object

Llorx:


Llorx:
Nope. My fault. Setting `null` to disable it instead of `false`.

Issue #13293, opened 2020-04-09
Missing table data in export-data module for bubble charts


weltall:
#### Expected behaviour

Directly from the highchart samples https://www.highcharts.com/demo/bubble on the main site i'd expect an export (view data table, export csv and xlsx) should export all the data from the series. Including at least name, x, y, z

#### Actual behaviour
What happens is some of the data is not exported. Specifically the x column, at least, is skipped and we get only the name, y and z column as it can be seen here:
![image](https://user-images.githubusercontent.com/1127663/78909782-4cbdcb80-7a84-11ea-996a-331a94de2eed.png)

#### Live demo with steps to reproduce
 https://www.highcharts.com/demo/bubble try to use export-data.js functionalities from the menu.

#### Product version
highchart 8.0.4

#### Affected browser(s)
Seems to affect all (i tried firefox and chrome last version, ie 11, edge

pawelfus:
Hi @weltall 

Thank you for reporting this bug.

**Workaround:**
Set `series.keys = ['x', 'y', 'z']`
Demo: https://jsfiddle.net/BlackLabel/v12gdtb9/

Basts:
After digging into this issue below are proposals of solutions which I would discuss. The main problem is with this line https://github.com/highcharts/highcharts/blob/master/ts/modules/export-data.src.ts#L536 -> `pointArrayMap = keys || series.pointArrayMap || ['y']` and `series.pointArrayMap` which in this case doesn't include the 'x'.

1. Leave it as it - by default when point doesn't have the `name` property the 'x' value is rendered under the `category` header, but the user has control of what should be displayed in the table by defining the `keys`. Demo: http://jsfiddle.net/BlackLabel/oksp4L7t/

2. Create a function which will check if data has defined name in the series (this functionality could be change - it is just a proposal). It is easy to see what I have in mind in the demo (line 68-83 and 92-94). 

Demo: http://jsfiddle.net/BlackLabel/mog4d9wc/

@oysteinmoseng @TorsteinHonsi

sebastianbochan:
Im close to solution from point 1, which seems to be more optimal and does not require overwriting methods or create one one.

TorsteinHonsi:
I think the logical problem here is that we assume that `name` is a replacement for `x`, which is usually true in category axes. But in the OP example, the X axis is not a category axis, and the name and the x value are two different dimensions.

[...]

PR #13513, opened 2020-05-13
bugfix/13293-export-table-data

Basts:
Fixed, #13293, adding x value in export data-table if the point has defined name and xAxis is not a category type.


js/modules/export-data.src.js
```diff
@@ -354,12 +354,26 @@ Highcharts.Chart.prototype.getDataRows = function (multiLevelHeaders) {
             categoryMap: categoryMap,
             dateTimeValueAxisMap: dateTimeValueAxisMap
         };
+    }, 
+    // Create point array depends if xAxis is category
+    // or point.name is defined #13293
+    getPointArray = function (series, xAxis) {
+        if (series.data.filter(function (d) { return d.name; }).length &&
```
sebastianbochan, line 361: I don't think that series.data.filter should be as a part of condition. Its unreadable.

I suggest to extract the filter into extra variable i.e. `namedPoints` and then use "namedPoints.length"

js/modules/export-data.src.js
```diff
@@ -354,12 +354,26 @@ Highcharts.Chart.prototype.getDataRows = function (multiLevelHeaders) {
...
+            xAxis && !xAxis.categories &&
+            series.type !== 'gantt') {
+            if (series.pointArrayMap && series.pointArrayMap.filter(function (p) { return p === 'x'; }).length) {
```
sebastianbochan, line 364: The same situation like above (line 361)

ts/modules/export-data.src.ts
```diff
@@ -523,6 +523,25 @@ Highcharts.Chart.prototype.getDataRows = function (
...
+        // Create point array depends if xAxis is category
+        // or point.name is defined #13293
+        getPointArray = function (
+            series: Highcharts.Series,
+            xAxis: Highcharts.Axis
+        ): string[] {
+            if (
+                series.data.filter((d: Highcharts.Point): string | undefined => d.name).length &&
+                xAxis && !xAxis.categories &&
+                series.type !== 'gantt'
```
pawelfus, line 535: Avoid checking `series.type`. Instead use `series.is('series-type')`. 

But why, in the first place, we need to filter out `gantt` type? How about x-range?

Basts: `Gantt` chart has defined properly the `series.pointArrayMap` - (`start, end`) and doesn't require any additional changes like adding the 'x' to this array.

TorsteinHonsi:
Please fix the commented cases and merge the master into this PR to update it.

PR #13520, opened 2020-05-14
Fixed #13494, bubble animation failed when redrawing

### LINKED ISSUE #13494: bubble series not visible after second call to addSeries
#### Expected behaviour
`addSeries` is called twice and two series are visible.
image

#### Actual behaviour
Only the second series is visible
image

https://jsfiddle.net/alvarezg/3xdkucny/
(resizing viewport or toggling series visibility in legend seems to fix the issue but initial render is wrong)
#### Product version

Highcharts, Highstock 8.0.3 and newer (works with 8.0.2)

TorsteinHonsi:
Fixed #13494, a regression causing bubble animation to fail when redrawing, like when running `addSeries` in quick succession.

### DIFF
--- a/samples/unit-tests/series-bubble/marker/demo.js
+++ b/samples/unit-tests/series-bubble/marker/demo.js
@@ -109,3 +109,74 @@ QUnit.test('Bubble data points without z-param.(#8608)', function (assert) {
         'Has marker'
     );
 });
+
+QUnit.test('Bubble animation and async redraws (#13494)', assert => {
+    const clock = TestUtilities.lolexInstall();
+
+    try {
+
+        const chart = Highcharts.chart('container', {
+            chart: {
+                type: 'bubble'
+            },
+            plotOptions: {
+                series: {
+                    animation: {
+                        duration: 100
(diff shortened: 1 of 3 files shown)


samples/unit-tests/series-bubble/marker/demo.js
```diff
@@ -108,4 +108,75 @@ QUnit.test('Bubble data points without z-param.(#8608)', function (assert) {
...
+                'First series points should continue animating'
+            );
+            assert.strictEqual(
+                chart.series[1].points[0].graphic.attr('width'),
+                1,
+                'Second series points should be in animation start position'
+            );
+        }, 50);
+
+        setTimeout(() => {
```
pawelfus, line 167: Since it's async test, shouldn't we use `done = asert.async()`? See: https://api.qunitjs.com/assert/async

TorsteinHonsi: Actually that's not required when we use lolex, since `setTimeout()` is overridden and made sync.

PR #13523, opened 2020-05-14
Fixed typos that broke multiple links in docs for exporting

cascading-jox:
I noticed that several links were broken in the api documentation for exporting and it looks like there was a couple of small typos. I think I managed to fix them but I am not familiar with how the documentation system works. 

An example of a broken link can be found [here](https://api.highcharts.com/highmaps/exporting.buttons) at the navigation.buttonOptions link. The link is generated as https://api.highcharts.com//navigation.buttonOptions instead of https://api.highcharts.com/highmaps/navigation.buttonOptions and I am guessing because the... I realized when thinking about why that my commit will not actually fix that particular link but it should fix some other links. I can't find a simple fix for that but feel free to edit my pull request if there is a solution.

### DIFF
--- a/js/modules/exporting.src.js
+++ b/js/modules/exporting.src.js
@@ -494,7 +494,7 @@ defaultOptions.exporting = {
      *
      * @sample {highcharts} highcharts/exporting/chartoptions-data-labels/
      *         Added data labels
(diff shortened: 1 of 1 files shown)


cascading-jox:
I checked the links again and it seems to be fixed. I am guessing that #13495 fixed the issue.

PR #13527, opened 2020-05-15
Fixed #13526, gradient fills not working in edge

### LINKED ISSUE #13526: Gradient fill with CSS doesn't work in Edge
#### Expected behaviour
Column bars to be gradient filled

#### Actual behaviour
Column bars not shown

https://jsfiddle.net/kr0tx7a1/

#### Product version
v8.1.0

#### Affected browser(s)
Edge only (AFAIK)

I am pretty sure this is a bug in Edge. If you manually remove the `highcharts-color-0` class from the `highcharts-series` *<g>* tag, it works. If you remove the same css class from the *<rect>* tag it works too. But if you have the same class applied to both, nothing is shown! 

I will do a pull request fix for this - either the `highcharts-color-0` class can be removed from the `highcharts-point` *<rect>* or removed from the group. As far as I can tell one series cannot have multiple colors so it doesn't make sense to have the class applied at the point level - it might as well just be removed. I have tried this for my graphs and it's working ok. I assume you guys can do better testing than me :)

rocklan:
This fixes gradient fills not working in edge browser as logged in #13526. An alternate fix would be to not add the _highcharts-color-x_ class at the group level, which I think is done inside `Series.js` inside the *plotGroup* function.

### DIFF
--- a/js/parts/Point.js
+++ b/js/parts/Point.js
@@ -495,8 +495,6 @@ var Point = /** @class */ (function () {
             (point.selected ? ' highcharts-point-select' : '') +
             (point.negative ? ' highcharts-negative' : '') +
             (point.isNull ? ' highcharts-null-point' : '') +
-            (typeof point.colorIndex !== 'undefined' ?
-                ' highcharts-color-' + point.colorIndex : '') +
             (point.options.className ? ' ' + point.options.className : '') +
             (point.zone && point.zone.className ? ' ' +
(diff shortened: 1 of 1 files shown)


TorsteinHonsi:
Thanks! See my comment in https://github.com/highcharts/highcharts/issues/13526#issuecomment-629205395. I'm closing this PR now, but we'll bring the idea into the fix.

PR #13528, opened 2020-05-15
Enhancement/options-es6

marvin19:
Migrated Options-file to ES6 pattern.

However, there are still some refactoring issues that needs to be solved for this file to be completely migrated. DefaultOptions still needs H. in the Utilties.ts file and Time.ts file. This has to be addressed at a later stage.


bre1470:
@TorsteinHonsi I am unsure about the changes in the visual tests.

PR #13529, opened 2020-05-15
bugfix/13334-ordinal-axis-panning

### LINKED ISSUE #13334: Can't move/drag to another period
#### Expected behaviour
Can drag / move to another period with all data.

#### Actual behaviour
I Can't drag / move to another period with data **data: dataArrObj** or **data: dataArr** but I can drag / move to another perior with **data: data** or **data: conversionData**.

Can't drag/move
image

https://jsfiddle.net/3L64r1mv/2/

#### Product version

#### Affected browser(s)
Chrome, Firefox, Edge

```HTML
<script src="https://code.highcharts.com/stock/highstock.js"></script>
<script src="https://code.highcharts.com/stock/modules/data.js"></script>
<script src="https://code.highcharts.com/stock/modules/exporting.js"></script>

<div id="container" style="height: 400px; min-width: 310px"></div>
```
```JS
const start = 1584388800000; // Mon Mar 16 2020 21:00:00 GMT+0100 (hora estándar de Europa central)
const step = 24 * 60 * 60 * 1000; // 24 hours

const getRandomValue = (a, b) => {
  const min = Math.ceil(a);
  const max = Math.floor(b);

  const random = Math.random();

  if (random < 0.1) {
    return null;
  }

  return Math.floor(random * (max - min + 1)) + min;
};

const generateArr = (len, arrOfObj = true) => {
  const arr = [];

  for (let i = 0; i < len; i++) {
    const date = start + i * step;

    if (arrOfObj) {
      arr.push({
        x: new Date(date).getTime(),
        y: getRandomValue(250, 750)/1000,
      });
    } else {
      arr.push([
        new Date(date).getTime(),
        getRandomValue(250, 750)/1000,
      ]);
    }
  }
  return arr;
};


Highcharts.getJSON('https://cdn.jsdelivr.net/gh/highcharts/highcharts@v7.0.0/samples/data/usdeur.json', function(data) {

  var conversionData = data.map(function(d) {
    return {
      x: d[0],
      y: d[1]*1000,
    }
  });

  // Generate same number of items that data
  const dataArrObj = generateArr(data.length, true);
  const dataArr = generateArr(data.length, false);


  console.log(data)
  console.log(conversionData)
  console.log(dataArrObj);
  console.log(dataArr);


  Highcharts.stockChart('container', {

    rangeSelector: {
      selected: 2,
      enabled: true,
      allButtonsEnabled: true,
      buttons: [
        // zoom button list
        {
          type: 'day', // if number of items does not complete a day, the selected style is not inserted
          count: 1,
          text: 'Day',
        },
        {
          type: 'week', // if number of items does not complete a week, the selected style is not inserted
          count: 1,
          text: 'Week',
        },
        {
          type: 'month', // if number of items does not complete a month, the selected style is not inserted
          count: 1,
          text: 'Month',
        },
      ],
    },

    title: {
      text: 'USD to EUR exchange rate'
    },

    yAxis: {
      title: {
        text: 'Exchange rate'
      },
    },

    series: [{
      turboThreshold: 0,
      name: 'USD to EUR',
      /* data: data, */
      /* data: conversionData, */
      /* data: dataArrObj, */
      data: dataArr,
      connectNulls: true,
      tooltip: {
        valueDecimals: 0
      }
    }]
  });
});

```

Denyllon:
Fixed #13334, panning an ordinal axis with equal spaced data did not work correctly.

### DIFF
--- a/js/parts/Interaction.js
+++ b/js/parts/Interaction.js
@@ -489,6 +489,7 @@ extend(Chart.prototype, /** @lends Chart.prototype */ {
                 xy = [0];
             }
             xy.forEach(function (isX) {
+                var _a;
                 var axis = chart[isX ? 'xAxis' : 'yAxis'][0], axisOpt = axis.options, horiz = axis.horiz, mousePos = e[horiz ? 'chartX' : 'chartY'], mouseDown = horiz ? 'mouseDownX' : 'mouseDownY', startPos = chart[mouseDown], halfPointRange = (axis.pointRange || 0) / 2, pointRangeDirection = (axis.reversed && !chart.inverted) ||
                     (!axis.reversed && chart.inverted) ?
                     -1 :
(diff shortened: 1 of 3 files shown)


js/parts/Interaction.js
```diff
@@ -523,41 +524,40 @@ extend(Chart.prototype, /** @lends Chart.prototype */ {
                     axis.toValue(axis.toPixels(extremes.max) +
                         axis.minPixelPadding));
                 axis.panningState = panningState;
-                // It is not necessary to calculate extremes on ordinal axis,
-                // because the are already calculated, so we don't want to
-                // override them.
-                if (!axisOpt.ordinal) {
```
pawelfus, line 529: Instead of changing the logic, would be sufficient to check `axis.isOrdinal`? That flag determines if data on the chart requires ordinal positions. Could you check this?

ts/parts/Interaction.ts
```diff
@@ -884,58 +884,58 @@ extend(Chart.prototype, /** @lends Chart.prototype */ {
 
                 axis.panningState = panningState;
 
-                // It is not necessary to calculate extremes on ordinal axis,
-                // because the are already calculated, so we don't want to
-                // override them.
-                if (!axisOpt.ordinal) {
```
pawelfus, line 890: Instead of changing logic, would be it sufficient to change to `!axis.isOrdinal`? That flag determines if ordinal positions were calculated, or if data on the chart is equally spaced and `ordinal` option does nothing.

PR #13545, opened 2020-05-18
bugfix/utilities-objecteach

bre1470:
Fixed typing of objectEach.


ts/modules/boost/wgl-renderer.ts
```diff
@@ -1645,7 +1645,7 @@ function GLRenderer(
         shader.destroy();
         if (gl) {
 
-            objectEach(textureHandles, function (key: string): void {
+            objectEach(textureHandles, function (_val, key): void {
```
TorsteinHonsi, line 1648: Why the dangling underscore? We have been trying to avoid those, though we have some of them for historic reasons.

ts/modules/boost/wgl-renderer.ts
```diff
@@ -1645,7 +1645,7 @@ function GLRenderer(
...
                 if (textureHandles[key].handle) {
```
TorsteinHonsi, line 1649: Now that we have `val` passed in, we can probably use that directly. `if (val.handle) {` etc.

ts/modules/boost/wgl-renderer.ts
bre1470: It is used in ESLint and TypeScript to mark unused function parameters, so that no complaints appear.

PR #13546, opened 2020-05-18
bugfix/pointer-class

bre1470:
Fixed typing of Pointer class.


bre1470:
I can not spot any difference in the visual tests.

PR #13552, opened 2020-05-19
bugfix/13371-item-series-without-size

### LINKED ISSUE #13371: Crash in item chart when enabling data labels without setting size
#### Expected behaviour
The item chart should not crash and automatically scale to optimally fit chart and data labels (like the pie chart)

#### Actual behaviour
The item chart crashes with: `highcharts.src.js:36896 Uncaught RangeError: Maximum call stack size exceeded`

https://jsfiddle.net/2swtvh1a/

* Crashes when running
* changing the chart type to `pie` fixes the issue
* adding `size` fixes the issue

#### Product version
Highcharts 8.0.4

#### Affected browser(s)
Tested in Chromium and Firefox

karolkolodziej:
Fixed #13371, unable to display data labels in item series without setting series size.


ts/modules/item-series.src.ts
```diff
@@ -198,15 +198,17 @@ seriesType<Highcharts.ItemSeries>(
...
+        ): void {
             if (!this.slots) {
                 this.slots = [];
             }
             if (
                 isNumber(this.options.startAngle) &&
                 isNumber(this.options.endAngle)
             ) {
-                H.seriesTypes.pie.prototype.translate.call(this);
+                H.seriesTypes.pie.prototype.translate.call(this, positions);
```
pawelfus, line 211: Did you consider using `apply(this, arguments)` instead? With `.call()` we may have similar issues in future if we decide to add more arguments.

karolkolodziej: You're right! It seems reasonable.

PR #13555, opened 2020-05-19
Fixed #11972, responsive pane

### LINKED ISSUE #11972: Gauge Chart vs. Responsive Rules: center of gauge needle moving upon value update
I'm having this issue where the center of the needle of a gauge chart moves when updating the value dynamically.

I'm seeing this occur when there is also a `responsive` rule (which contains a left margin) is set.

http://jsfiddle.net/bramus/k40oztb6/

Note the use of the `responsive` block along with the `setInterval` which updates the value dynamically.

#### Expected behaviour

The base of the needle remains in the center of the chart

#### Actual behaviour

The center of the needle shifts upon update of the series' value

#### Product version

Using highcharts 7.2.0 now. The bug itself got introduced with version 7.1.3.

If you alter the demo to use version 1.7.2 then you'll see the whole gauge pane get shifted (due to the `responsive` rule) upon initializing. Starting with version 7.1.3 the gauge pane itself does not move upon load.

#### Affected browser(s)

Seeing this in latest Chrome, latest Firefox, latest Safari

#### Further findings

1. The tooltip is immediately positioned at the correct position.
2. When wrapping the `#container` into a `#wrapper`, and setting the width/height on the `#container` (see http://jsfiddle.net/bramus/b8v35L2s/), the rendering issue can be corrected by resizing the window : you'll see the gauge pane get moved around upon a window resize.
3. When immediately setting the left margin (instead of through a responsive rule) the chart renders correctly

Points 2 and 3 above leave me to think that the responsive rules aren't applied properly after load.

TorsteinHonsi:
Fixed #11972, not all chart elements responded to responsive margins

### DIFF
--- a/js/parts/Dynamics.js
+++ b/js/parts/Dynamics.js
@@ -461,9 +461,13 @@ extend(Chart.prototype, /** @lends Highcharts.Chart.prototype */ {
                     chart.isDirtyBox = true;
                 }
                 // Chart setSize
-                if (!isResponsiveOptions &&
-                    chart.propsRequireReflow.indexOf(key) !== -1) {
-                    runSetSize = true;
+                if (chart.propsRequireReflow.indexOf(key) !== -1) {
+                    if (isResponsiveOptions) {
+                        chart.isDirtyBox = true;
+                    }
+                    else {
+                        runSetSize = true;
+                    }
                 }
             });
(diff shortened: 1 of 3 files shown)


sebastianbochan:
Tested all my use cases and all of them are passed.

The bug was really tricky.

Thanks!

Issue #13482, opened 2020-05-11
Markers can't be disabled on styled mode when accessibility module is loaded


alvarezguille:
#### Expected behaviour
Initial setting or update of marker.enabled to false should hide markers.

#### Actual behaviour
Only marker from legend is disabled, markers in data series are still show.

#### Live demo with steps to reproduce
https://jsfiddle.net/alvarezg/42e35c86/

#### Product version
Highcharts and Highstock 7.1.2 or later

sebastianbochan:
Hi @alvarezguille,
Thank you for reporting. 

FYI @oysteinmoseng

TorsteinHonsi:
As @oysteinmoseng mentioned in the PR, I agree this fix is quick and effective, but perhaps not a permanent solution. The better approach would probably be to add a class name to the marker group (`.highcharts-markers-hidden` for example) and add that to the default CSS.

```
.highcharts-markers-hidden .highcharts-point:not(.highcharts-point-hover) {
    display: none;
}
```

### Workaround for v8.1.0
Add that CSS to the `.highcharts-markers` group:
https://jsfiddle.net/highcharts/f1jxt98v/

oysteinmoseng:
Note: For better a11y use `opacity: 0` instead of `display: none`.

PR #13563, opened 2020-05-20
Improved fix for #13482, forced markers in styled mode.

oysteinmoseng:
Improved fix for #13482, support for forced markers in styled mode.
___
As discussed in #13482, adds proper support for forced markers in styled mode. 3 new classes are added:
- `highcharts-a11y-markers-hidden` for indicating a series with invisible markers (`marker.enabled: false` on series)
- `highcharts-a11y-marker-hidden` indicates a point with an invisible marker (`marker.enabled: false` on point).
- `highcharts-a11y-marker-visible` indicates a forced visible marker (`marker.enabled: false` on series, but `enabled: true` on point).

The latter two are only used when a series has individual point marker settings. All of the above is only added in styled mode.

### DIFF
--- a/css/highcharts.scss
+++ b/css/highcharts.scss
@@ -282,6 +282,10 @@ g.highcharts-series,
     stroke-width: 1px;
     stroke: $background-color;
 }
+.highcharts-a11y-markers-hidden .highcharts-point:not(.highcharts-point-hover):not(.highcharts-a11y-marker-visible),
+.highcharts-a11y-marker-hidden {
+    opacity: 0;
+}
 .highcharts-point {
     stroke-width: 1px;
 }
--- a/js/modules/accessibility/components/SeriesComponent/forcedMarkers.js
+++ b/js/modules/accessibility/components/SeriesComponent/forcedMarkers.js
@@ -26,21 +26,10 @@ function isWithinDescriptionThreshold(series) {
 /**
  * @private
  */
-function isWithinNavigationThreshold(series) {
-    var navOptions = series.chart.options.accessibility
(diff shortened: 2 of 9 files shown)


oysteinmoseng:
Re: visual differences - is the tool not taking opacity into account? Is it just showing additional SVG elements? I can not visually see markers in these demos, even when adding the a11y module.

TorsteinHonsi:
> Re: visual differences - is the tool not taking opacity into account?

The tool applies dummy CSS to the generated SVG, so if the element is hidden using a CSS rule in `highcharts.css`, it won't take effect.

oysteinmoseng:
Indeed, which is why we have this option (defaults to 200) for screen readers: https://api.highcharts.com/highcharts/accessibility.series.pointDescriptionEnabledThreshold.

But we also draw the forced markers if we are below the threshold for keyboard navigation (https://api.highcharts.com/highcharts/accessibility.keyboardNavigation.seriesNavigation.pointNavigationEnabledThreshold), which is infinite by default. This is so that we can focus data points dynamically. I am wondering if we can get away with not doing that, and just show the tooltip as long as we are above the screen reader threshold. I'll look into the details. We also have to consider voice input support here.

Another possibility is to lower the keyboard nav threshold, but given the repeat speed of most keyboards we can still efficiently navigate pretty large charts.

oysteinmoseng:
I think we can get away with not drawing markers for kbd navigation and voice input. That means the default limit for forcing markers is now 200 (visible) points. If there is something I am not considering, we can always reevaluate if we get bug reports.

PR #13596, opened 2020-05-27
bugfix/13372-horizontal-item-series-drilldown

### LINKED ISSUE #13372: Crash in item chart when using drilldown in horizontal/vertical layout
#### Expected behaviour
Clicking a data point should update the displayed data to the specified drilldown data, exactly like in the circular layout.

#### Actual behaviour
When clicking a data point for drilldown the data disappears.

https://jsfiddle.net/6sbcwrjq/1/

* Click on any of the data points for drilldown -> crash
* Uncomment the *circular options*
* Try again -> works

#### Product version
Highcharts 8.0.4

#### Affected browser(s)
Chromium

karolkolodziej:
Fixed #13372, `drilldown` in horizontal item series was not allowed.

### DIFF
--- a/ts/modules/drilldown.src.ts
+++ b/ts/modules/drilldown.src.ts
@@ -12,8 +12,13 @@
 
 'use strict';
 
+import Chart from '../parts/Chart.js';
 import Color from '../parts/Color.js';
 import H from '../parts/Globals.js';
+import O from '../parts/Options.js';
+const {
+    defaultOptions
+} = O;
 import Point from '../parts/Point.js';
 import SVGRenderer from '../parts/SVGRenderer.js';
 import Tick from '../parts/Tick.js';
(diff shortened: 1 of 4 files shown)


ts/modules/drilldown.src.ts
```diff
@@ -1276,39 +1276,45 @@ if (PieSeries) {
...
-                    var animateTo = point.shapeArgs;
-
-                    if (!styledMode) {
-                        (animateFrom as any).fill = level.color;
-                        (animateTo as any).fill = point.color;
-                    }
+                    (this.chart.options.drilldown as any).animation;
+
+            // Unable to drill down in the horizontal item series #13372
+            if (this.type !== 'item' && this.center) {
```
TorsteinHonsi, line 1282: Don't do hard checks on `this.type`, as it won't work for future series types that build upon the item series. Instead, use `this.is('item')`, or preferably do a feature check where it actually goes wrong.

PR #13597, opened 2020-05-27
bugfix/13202-data-format-lollipop

### LINKED ISSUE #13202: The lollipop series - data issue

raf18seb:
Fixed #13202, the lollipop series didn't work with all data formats.

This is a continuation of https://github.com/highcharts/highcharts/pull/13335 - had some issues with TS while merging master so I decided to create a fresh PR. The old one can be removed.

### DIFF
--- a/js/modules/lollipop.src.js
+++ b/js/modules/lollipop.src.js
@@ -56,13 +56,25 @@ seriesType('lollipop', 'dumbbell', {
         pointFormat: '<span style="color:{series.color}">●</span> {series.name}: <b>{point.low}</b><br/>'
     }
 }, {
+    pointArrayMap: ['y'],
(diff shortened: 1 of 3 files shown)


raf18seb:
Not yet, there is a problem with a tooltip. I need to check all other possible issues

raf18seb:
Now it should be OK (only some visual differences, but not related to the changes I made).

PR #13601, opened 2020-05-28
bugfix/12012-gantt-collapse-chart-height

### LINKED ISSUE #12012: Collapsing/Expanding sometimes visually not happening when the Gantt has a fixed height
#### Expected behaviour
Collapsing and expanding should hide/show data.

#### Actual behaviour
When setting a fixed height, some combinations of collapse/expands don't visually update the Gantt.

https://jsfiddle.net/g6t8yh0r/1/
This demo is just the official subtask demo. I removed the title and added a fixed height of `300`.

To reproduce, in this example: 
- first, click on "Design" (works OK) 
- then, click on "Planning" (visually, nothing happens)
- finally, click on "Develop" (this will cause both this Tick and Planning to collapse)

And some similar issues happen when trying to expand.

I thought simply calling `chart.redraw()` or `chart.update({})` would fix it but it didn't.

#### Product version
Product: Highcharts Gantt v7.2.0

#### Affected browser(s)
Chrome, but I assume any other too.

goransle:
Fixed #12012, collapse not working when chart had set height

### DIFF
--- a/ts/parts-gantt/TreeGridAxis.ts
+++ b/ts/parts-gantt/TreeGridAxis.ts
@@ -643,9 +643,9 @@ namespace TreeGridAxis {
 
             // If staticScale is not defined on the yAxis
             // and chart height is set, set axis.isDirty
-            // to ensure collapsing works (#13601)
+            // to ensure collapsing works (#12012)
             addEvent(axis, 'afterBreaks', function (): void {
-                if (axis.coll === 'yAxis' && !axis.staticScale && axis.chart.userOptions.chart?.height) {
+                if (axis.coll === 'yAxis' && !axis.staticScale && axis.chart.options.chart?.height) {
                     axis.isDirty = true;
                 }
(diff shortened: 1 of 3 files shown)


ts/parts-gantt/TreeGridAxis.ts
```diff
@@ -641,6 +641,15 @@ namespace TreeGridAxis {
                 }
             });
 
+            // If staticScale is not defined on the yAxis
+            // and chart height is set, set axis.isDirty
+            // to ensure collapsing works (#13601)
+            addEvent(axis, 'afterBreaks', function (): void {
+                if (axis.coll === 'yAxis' && !axis.staticScale && axis.chart.userOptions.chart?.height) {
```
TorsteinHonsi, line 648: Use `axis.chart.options` instead of `userOptions`, to pick up settings from themes etc.

PR #13603, opened 2020-05-28
Fixed issue with oneToOne and ids

### LINKED ISSUE #13541: The 'oneToOne' parameter of 'Chart.update' does not work properly.
#### Expected behaviour
Updating the chart with `oneToOne` flag set equal to `true` should work as it is described in documentation.

#### Actual behaviour
Updating chart with the new series array, where all series contain `id` does not work as it is described in documentation.

Per the documentation:
> If the series have id's, the new series options will be matched by id, and the remaining ones removed.

However, the following example shows that it does not work in that way, because it should result in removing the first series, keeping the second without any changes, and add the third. The effect is keeping only the third one.

This issue could be related with #9606.

https://jsfiddle.net/BlackLabel/bt0acn2w/

TorsteinHonsi:
Fixed #13541, `chart.update` with `oneToOne` failed when removing and adding series, if all series had explicit id's.

### DIFF
--- a/js/parts/Dynamics.js
+++ b/js/parts/Dynamics.js
@@ -526,8 +526,21 @@ extend(Chart.prototype, /** @lends Highcharts.Chart.prototype */ {
                     });
                 }
                 splat(options[coll]).forEach(function (newOptions, i) {
-                    var item = (defined(newOptions.id) &&
-                        chart.get(newOptions.id)) || chart[coll][indexMap ? indexMap[i] : i];
+                    var hasId = defined(newOptions.id);
+                    var item;
+                    // Match by id
+                    if (hasId) {
+                        item = chart.get(newOptions.id);
+                    }
+                    // No match by id found, match by index instead
+                    if (!item) {
+                        item = chart[coll][indexMap ? indexMap[i] : i];
+                        // Check if we grabbed an item with an exising but
+                        // different id (#13541)
+                        if (item && hasId && defined(item.options.id)) {
+                            item = void 0;
+                        }
+                    }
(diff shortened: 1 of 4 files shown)


sebastianbochan:
In my tests all work properly.

@Denyllon could you take a look too?

Denyllon:
@sebastianbochan Thank you for summoning.
What about the situation, where we have defined two series with `id`s, and calling the `update()` with also two series, though the one without id? Shouldn't that works in such a way that the series with id will be updated, and the series without it might be ignored?
Currently when doing something like that, we can see that only the series without id is visible on the chart. Maybe it's just a wrong use case, but I was thinking that would be good to make you to know about it.

Example: https://jsfiddle.net/BlackLabel/bgu536sL/

sebastianbochan:
Mea culpa, I found the next edge case: https://jsfiddle.net/BlackLabel/9rd1sj70/ (Large + click update). Maybe Im wrong, but it looks like also inconsistant.

TorsteinHonsi:
@Denyllon 
What happens is that the second series has no id, so it updates by position. In position 1 there is a series with an id, (before the first series is removed), so it updates that. The question it comes down to, is what to do if the new config has an id and the old one does not. In the current state of the PR, the series will get updated in this case.

@sebastianbochan 
My head explodes when trying to understand your sample, I'm back on Monday, have a nice weekend!

sebastianbochan:
@TorsteinHonsi 
Sorry for breaking your Friday :D 
I simplified my demo and turned our that was my mistake, sorry again! We can ignore this use case.

PR #13604, opened 2020-05-28
 bugfix/13420-annotations-not-added-on-ios

### LINKED ISSUE #13420: Adding annotations on Highstock GUI doesn't work properly on iOS devices.
#### Expected behaviour
Adding annotations from GUI should work in the same way as on Android devices.

#### Actual behaviour
It's hard to add an annotation to the chart, using GUI implemented in the below demo.
An attend to add any annotation stops on initialization. For example, after choosing the **Eliot 5 Line**, it is possible to start drawing the first line only, but not to create its second anchor point.
ezgif com-optimize

Live example: https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/stock/demo/stock-tools-gui/

#### Product version
Highstock v8.0.4 

#### Affected browser(s)
OS: iOS 13.1.1 (iPad 7th Gen)
Chrome v80.0.3987.95
Safari v13

Izothep:
Fixed #13420, adding annotation was not working on iOS devices.

### DIFF
--- a/js/annotations/eventEmitterMixin.js
+++ b/js/annotations/eventEmitterMixin.js
@@ -24,16 +24,9 @@ var eventEmitterMixin = {
      */
     addEvents: function () {
         var emitter = this, addMouseDownEvent = function (element) {
-            if (Highcharts.isTouchDevice) {
-                addEvent(element, 'touchstart', function (e) {
-                    emitter.onMouseDown(e);
-                });
-            }
-            else {
-                addEvent(element, 'mousedown', function (e) {
-                    emitter.onMouseDown(e);
-                });
-            }
+            addEvent(element, Highcharts.isTouchDevice ? 'touchstart' : 'mousedown', function (e) {
+                emitter.onMouseDown(e);
+            });
         };
         addMouseDownEvent(this.graphic.element);
         (emitter.labels || []).forEach(function (label) {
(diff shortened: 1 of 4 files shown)


js/annotations/eventEmitterMixin.js
```diff
@@ -98,23 +110,44 @@ var eventEmitterMixin = {
...
-        emitter.removeMouseUp = addEvent(H.doc, 'mouseup', function (e) {
-            emitter.cancelClick = emitter.hasDragged;
-            emitter.hasDragged = false;
-            emitter.chart.hasDraggedAnnotation = false;
-            // ControlPoints vs Annotation:
-            fireEvent(pick(emitter.target, emitter), 'afterUpdate');
-            emitter.onMouseUp(e);
-        });
+        if (Highcharts.isTouchDevice) {
+            emitter.removeDrag = addEvent(H.doc, 'touchmove', function (e) {
```
sebastianbochan, line 114: Instead of duplicate the block of code, can we use: 

`const dragType = Highcharts.isTouchDevice ? 'touchmove' : 'mousemove'`

and then use it as a param od addEvent?

Izothep: All is corrected and working - thanks for this info!

PR #13607, opened 2020-05-29
Fixed #13544, errors with axis count and responsive options

### LINKED ISSUE #13544: Responsive chart error
#### Expected behaviour

I have two series and two yAxis respectivly. If chart width smaller than 500px, only one vertical axis remains, and both of series are connected to it. When I increase the display, the second yAxis appears, and second series should be connected to it. 

#### Actual behaviour

When I reduce the display width, all works properly, but when I increase it, I get an error `#18`:
It seems like series haven't been updated.

image

http://jsfiddle.net/rsaktaev/pywqu6ot/

Please try to reduce the display, then increase it. Am I doing something wrong?
#### Product version

Hightcharts 8.1

TorsteinHonsi:
Fixed #13544, JavaScript error when the axis count changed in responsive options

### DIFF
--- a/js/parts/Responsive.js
+++ b/js/parts/Responsive.js
@@ -208,7 +208,10 @@ Chart.prototype.matchResponsiveRule = function (rule, matches) {
 /**
  * Get the current values for a given set of options. Used before we update
  * the chart with a new responsiveness rule.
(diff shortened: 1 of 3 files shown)


sebastianbochan:
Looks good 👍

PR #13609, opened 2020-05-29
bugfix/13369-tick-amount-changes-scatter

### LINKED ISSUE #13369: Tick amount changes when changing series visibility- scatter
#### Expected behaviour
Tick amount should remain the same when you change series visibility.

#### Actual behaviour
When changing the series visibility, tick amount changes. 

https://jsfiddle.net/BlackLabel/6wbzfnsd/

#### Product version
Highcharts 8.0.4

#### Affected browser(s)
All

karolkolodziej:
Fixed #13369, ticks were different after changing series visibility.

### DIFF
--- a/js/parts/Axis.js
+++ b/js/parts/Axis.js
@@ -1442,7 +1442,7 @@ var Axis = /** @class */ (function () {
         // This applies only if tickInterval is not defined.
         minTickInterval = pick(options.minTickInterval, 
         // In datetime axes, don't go below the data interval, except when
-        // there are scatter-like series involved (#13669).
+        // there are scatter-like series involved (#13369).
         axis.dateTime &&
             !axis.series.some(function (s) { return s.noSharedTooltip; }) ?
             axis.closestPointRange : 0);
--- a/js/parts/Axis.js
+++ b/js/parts/Axis.js
@@ -1071,8 +1071,7 @@ var Axis = /** @class */ (function () {
             this.series.forEach(function (series) {
                 var seriesClosest = series.closestPointRange, visible = series.visible ||
                     !series.chart.options.chart.ignoreHiddenSeries;
-                if (!series.noSharedTooltip &&
-                    defined(seriesClosest) &&
+                if (defined(seriesClosest) &&
                     visible) {
                     ret = defined(ret) ?
                         Math.min(ret, seriesClosest) :
--- a/js/parts/Axis.js
+++ b/js/parts/Axis.js
@@ -1071,7 +1071,8 @@ var Axis = /** @class */ (function () {
             this.series.forEach(function (series) {
                 var seriesClosest = series.closestPointRange, visible = series.visible ||
                     !series.chart.options.chart.ignoreHiddenSeries;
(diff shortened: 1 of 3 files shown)


karolkolodziej:
Sure, I will examine it closer!

karolkolodziej:
> All code is there for a reason, so when fixing a bug by removing code we also need to go back (via blame) to try to understand what that condition does in the first place. In this case it is there to prevent the computed `pointRange` from becoming to small. The computed `pointRange` is used for columns and derived series.

You're right! I passed all test and I didn't think about this that way. Do you think I should create a simple workaround or some fancier, more complex one?

TorsteinHonsi:
I'm not sure... I would start by trying to understand why there is only one tick rendered for a line series in this case: https://jsfiddle.net/highcharts/tm6fdL8h/ . I am guessing it has something to do with not rendering more ticks than there are data points, but it seems a bit aggressive in this case. 

Though I'm not sure it is a bug. The idea is that a line series is sequential, so we don't want too many intermediate ticks between points. A scatter series is, well, scattered throughout the cartesian plane. 

What is probably a bug is that when we have a combination of these, the axis should use the smallest tick interval, so it wouldn't change when we toggle the line series.

karolkolodziej:
The direct reason why we get only one tick in your example is the function called `trimTicks` which trims ticks which are not visible. As you see in the demo when we comment parts responsible for this we can watch them. However, we go back we see that the `tickPositions` are calculated based on the `tickInterval`. 
Interval mostly depends on the function called `getClosest`. But because scatter don't have shared tooltip we receive `undefined` here and that is why we get different ticks on a series toggle. 
Uncomment the second wrap to see quick workaround: https://jsfiddle.net/BlackLabel/4ozh6nk2/
Please let me know what do you think about it.

TorsteinHonsi:
I think I found how it is connected. The problem is that `axis.closestPointRange` is used for two things (at least):

1. For determining the `pointRange` in column-like series, hence the regression that I pointed out above.
2. For providing a default `minTickInterval` for datetime axes.

So the problem with this issue here is that the default `minTickInterval` doesn't make sense with scattered data, only with ordered data. Here is a solution that I think works well. Instead of applying the `closestPointRange` always, we first check if there is a `noSharedTooltip` series in the chart. The `noSharedTooltip` property, though perhaps not well named, tells us that the data isn't sorted by X.

```js
        // Before normalizing the tick interval, handle minimum tick interval.
        // This applies only if tickInterval is not defined.
        minTickInterval = pick(
            options.minTickInterval,
            // In datetime axes, don't go below the data interval, except when
            // there are scatter-like series involved (#13669).
            axis.dateTime &&
            !axis.series.some((s): boolean|undefined => s.noSharedTooltip) ?
                axis.closestPointRange : 0
        );
```

If you agree with this solution, please apply it, then set up unit tests for this issue and also for the regression with column/scatter that I pointed out above.

karolkolodziej:
Yes, I agree! 
Your solution works better and looks clean too.
Thanks for helping me understand this complex topic.

PR #13617, opened 2020-06-01
bugfix/lollipop-point

bre1470:
Fixed Point reference in lollipop series.


bre1470:
Visuals fail because of compiling problem in master.

PR #13623, opened 2020-06-02
bugfix/packedbubble-chart

bre1470:
Fixed usage of Chart class in packed bubble series.

### DIFF
--- a/ts/parts-more/PackedBubbleSeries.ts
+++ b/ts/parts-more/PackedBubbleSeries.ts
@@ -10,7 +10,25 @@
 
 'use strict';
 
+import Chart from '../parts/Chart.js';
+import Color from '../parts/Color.js';
+const color = Color.parse;
 import H from '../parts/Globals.js';
+import Point from '../parts/Point.js';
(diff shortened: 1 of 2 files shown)


ts/parts-more/PackedBubbleSeries.ts
```diff
@@ -10,15 +10,33 @@
...
+} = U;
 
 /**
  * Internal types
  * @private
  */
 declare global {
     namespace Highcharts {
-        interface Chart {
+        interface ChartLike {
```
sebastianbochan, line 39: Is it a proper name? Sounds a little strange...

bre1470: This has to change that way, otherwise you will not be able to reference the ES6-imported Chart type in the internal namespace.

bre1470: Also: The internal namespace is a temporary one to help during the migration. You should not rely on it and always favor ES6 imports. A good example in how to do that is the accessibility code.

PR #13625, opened 2020-06-02
Fixed #13590, Improved removeClass whitespace logic

### LINKED ISSUE #13590: Pie chart: CSS classes not being added/removed correctly
#### Expected behaviour
When you clickOn/mouseOver a chart segment it should add a class to the svg path and when you  clickOff/mouseOut it should remove the class.

Similarly `point.update({ className: 'foo' })` should add `foo` class to the path and `point.update({ className: null })` should remove the `foo` class.

#### Actual behaviour
Classes get added without a space separating them and you see something like
`highcharts-pointhighcharts-color-0`.

Or you see something like `highcharts-point highcharts-color-0foo` and then `point.update({ className: null })` has no effect.

1. right click on the "Chrome" chart segment and "inspect"
2. notice the class attribute is OK, eg `highcharts-point highcharts-point-select highcharts-color-0`
3. left-click the "Chrome" segment
4. notice the class attribute has broken `highcharts-pointhighcharts-color-0`.

PS. same issue exists when I use `point.update({ className: 'foo' })` but I don't have a demo ready.

#### Product version
8.x

#### Affected browser(s)
All?

webdevan:
This should fix the bug identified in #13590 by ensuring that there is always whitespace when a class is removed.

It also fixes another potential bug where the `replace` function might replace a substring of a CSS class.


TorsteinHonsi:
Thanks @webdevan !

I have merged your PR into a new working branch so that we can run the CI.

sebastianbochan:
@TorsteinHonsi,
I run all our internal tests + test the related bug, all worked properly.

TorsteinHonsi:
Great, thanks! I had to remove the `.trim()` though, it will break old IE. But it's not critical.

sebastianbochan:
How about unit-test? Can we skip that or add?

TorsteinHonsi:
@sebastianbochan 
Yes we should definately have that. Should be added to the other PR.

PR #13628, opened 2020-06-03
Fixed #13590, CSS classes not being removed correctly

### LINKED ISSUE #13590: Pie chart: CSS classes not being added/removed correctly

TorsteinHonsi:
Fixed #13590, CSS classes not being removed correctly, causing unwanted behaviour in pie chart hovering and unhovering.

___

### To do
 - [x] Unit tests


sebastianbochan:
Looks like incorrect because of few other commits (different bugs).

TorsteinHonsi:
> Looks like incorrect because of few other commits (different bugs).

Okay, feel free to start over again with a fresh PR if needed. Don't know where that history came from.

sebastianbochan:
The problem was caused by this PR: https://github.com/highcharts/highcharts/pull/13625 where a few other commits are inside. As I see, all of them have been already merged so it should not be a problem during merge of this PR.

PR #13638, opened 2020-06-04
docs/point-id

### LINKED ISSUE #13614: Gantt - point.id  missing type + api
#### Expected behaviour
According to the API:

```
series.gantt.data.dependency

The ID of the point (task) that this point depends on in Gantt charts. Aliases connect. Can also be an object, specifying further connecting options between the points. Multiple connections can be specified by providing an array.

```
```
series.gantt.data.dependency.to

The ID of the point to connect to.
```

It should be possible to add the id property to the data point, as we can see in this official demo: https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/gantt/demo/project-management#save

#### Actual behaviour

Such type doesn't exist on the GanttPoint, moreover, it doesn't exist in the API too: https://api.highcharts.com/gantt/series.gantt.data

```javascript
// GanttPointOptionsObject pasted from highcharts.d.ts

export interface GanttPointOptionsObject {
    /**
     * (Gantt) Accessibility options for a data point.
     */
    accessibility?: PointAccessibilityOptionsObject;
    /**
     * (Gantt) Whether the grid node belonging to this point should start as
     * collapsed. Used in axes of type treegrid.
     */
    collapsed?: boolean;
    /**
     * (Gantt) Progress indicator, how much of the task completed. If it is a
     * number, the `fill` will be applied automatically.
     */
    completed?: (number|XrangePointPartialFillOptionsObject);
    /**
     * (Gantt) A reserved subspace to store options and values for customized
     * functionality. Here you can add additional data for your own event
     * callbacks and formatter callbacks.
     */
    custom?: Dictionary<any>;
    /**
     * (Gantt) The ID of the point (task) that this point depends on in Gantt
     * charts. Aliases connect. Can also be an object, specifying further
     * connecting options between the points. Multiple connections can be
     * specified by providing an array.
     */
    dependency?: (string|XrangePointConnectorsOptionsObject|Array<(string|XrangePointConnectorsOptionsObject)>);
    /**
     * (Gantt) A description of the point to add to the screen reader
     * information about the point.
     */
    description?: string;
    /**
     * (Gantt) Point specific options for the draggable-points module. Overrides
     * options on `series.dragDrop`.
     */
    dragDrop?: SeriesLineDataDragDropOptions;
    /**
     * (Highcharts) The `id` of a series in the drilldown.series array to use
     * for a drilldown for this point.
     */
    drilldown?: string;
    /**
     * (Gantt) The end time of a task.
     */
    end?: number;
    /**
     * (Gantt) The rank for this point's data label in case of collision. If two
     * data labels are about to overlap, only the one with the highest
     * `labelrank` will be drawn.
     */
    labelrank?: number;
    /**
     * (Highcharts, Highstock) Options for the point markers of line-like
     * series.
     */
    marker?: PointMarkerOptionsObject;
    /**
     * (Gantt) Whether this point is a milestone. If so, only the `start` option
     * is handled, while `end` is ignored.
     */
    milestone?: boolean;
    /**
     * (Gantt) The name of a task. If a `treegrid` y-axis is used (default in
     * Gantt charts), this will be picked up automatically, and used to
     * calculate the y-value.
     */
    name?: string;
    /**
     * (Gantt) The ID of the parent point (task) of this point in Gantt charts.
     */
    parent?: string;
    /**
     * (Gantt) The start time of a task.
     */
    start?: number;
    /**
     * (Gantt) The Y value of a task.
     */
    y?: number;
}
```

#### Product version
8.1.0

bre1470:
Close #13614, Added public doclet for Point.id property.

### DIFF
--- a/js/parts/Point.js
+++ b/js/parts/Point.js
@@ -192,6 +192,13 @@ var Point = /** @class */ (function () {
          */
         this.colorIndex = void 0;
         this.formatPrefix = 'point';
+        /**
+         * An ID for the point instance. This is automatically generated if the
+         * point option is unset.
+         *
+         * @name Highcharts.Point#id
+         * @type {string}
+         */
         this.id = void 0;
         this.isNull = false;
         /**
@@ -625,7 +632,6 @@ var Point = /** @class */ (function () {
     Point.prototype.init = function (series, options, x) {
         this.series = series;
         this.applyOptions(options, x);
-        // Add a unique ID to the point if none is assigned
         this.id = defined(this.id) ? this.id : uniqueKey();
         this.resolveColor();
         series.chart.pointCount++;
--- a/js/parts/Series.js
+++ b/js/parts/Series.js
@@ -5423,7 +5423,7 @@ null,
  * @apioption series.line.data.description
  */
 /**
- * An id for the point. This can be used after render time to get a
+ * An ID for the point. This can be used after render time to get a
  * pointer to the point object through `chart.get()`.
  *
  * @sample {highcharts} highcharts/point/id/
--- a/ts/parts/Point.ts
+++ b/ts/parts/Point.ts
@@ -367,6 +367,13 @@ class Point {
 
     public formatPrefix: string = 'point';
 
+    /**
+     * An ID for the point instance. This is automatically generated if the
+     * point option is unset.
+     *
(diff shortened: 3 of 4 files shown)


bre1470:
This PR is not necessary.

PR #13642, opened 2020-06-04
Fixed #13554, a11y not supporting proximate legend layout.

### LINKED ISSUE #13554: Accessibility Module and Legend Proximate
#### Expected behaviour
When using the legend proximate, when clicking on each item it should toggle the visibility.

#### Actual behaviour
When the accessibility module is loaded the legend items stop referencing the correct line so when an item moves to the bottom of the list and you then try and enable it another legend is affected.

https://jsfiddle.net/garybrain/m9e37zyp/10/

#### Product version
8.1.0

#### Affected browser(s)
All

oysteinmoseng:
Fixed #13554, a11y not supporting proximate legend layout.
___
Moderately pragmatic approach taken. We need to recalculate positions of the proxy overlays when legend item position changes, but this is not trivial because of animation. Decided to take the same approach as with legend scrolling and fire an event after the animation duration has passed. It seems overkill to recalculate the position for every step (like we do with focus border by overriding position setters in the renderer).

There is a repeating pattern of the a11y module having to create HTML overlays for SVG elements that are animated, so we might want to find an overall better solution to this when we have time, rather than firing events and always calculating bounding boxes.

### DIFF
--- a/ts/parts/Legend.ts
+++ b/ts/parts/Legend.ts
@@ -561,12 +561,21 @@ class Legend {
             legendGroup = item.legendGroup;
 
         if (legendGroup && legendGroup.element) {
-            legendGroup[defined(legendGroup.translateY) ? 'animate' : 'attr']({
+            const shouldAnimate = defined(legendGroup.translateY);
+
+            legendGroup[shouldAnimate ? 'animate' : 'attr']({
                 translateX: ltr ?
                     itemX :
                     legend.legendWidth - itemX - 2 * (symbolPadding as any) - 4,
                 translateY: itemY
             });
+
+            const animOptions = animObject(
+                pick(this.chart.renderer.globalAnimation, true)
+            );
+            syncTimeout((): void => {
+                fireEvent(this, 'afterPositionItem', { item });
+            }, shouldAnimate && animOptions.duration || 0);
         }
 
         if (checkbox) {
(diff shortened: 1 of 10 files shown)


ts/parts/Legend.ts
```diff
@@ -561,12 +561,21 @@ class Legend {
...
-            legendGroup[defined(legendGroup.translateY) ? 'animate' : 'attr']({
+            const shouldAnimate = defined(legendGroup.translateY);
+
+            legendGroup[shouldAnimate ? 'animate' : 'attr']({
                 translateX: ltr ?
                     itemX :
                     legend.legendWidth - itemX - 2 * (symbolPadding as any) - 4,
                 translateY: itemY
             });
+
```
TorsteinHonsi, line 572: I think (without testing it) that it may be both safer and cheaper to use the `complete` callback for animation. Can you check that? Something like this:

```js
            const complete = (): void => {
                fireEvent(this, 'afterPositionItem', { item });
            };
            const attribs = {
                translateX: ltr ?
                    itemX :
                    legend.legendWidth - itemX - 2 * (symbolPadding as any) - 4,
                translateY: itemY
            };

            if (defined(legendGroup.translateY)) {
                legendGroup.animate(attribs, { complete });
            } else {
                legendGroup.attr(attribs);
                complete();
            }
```
I haven't counted the bytes though :) But I assume under any circumstance it is better practice to use the `complete` callback than a timeout.

PR #13643, opened 2020-06-04
bugfix/13534-draggable-points-touch-events-error

### LINKED ISSUE #13534: Chart with draggable points throws errors on touch devices
#### Expected behaviour
Touch devices supported

#### Actual behaviour
Using charts with draggable points with touch devices make them unresponsive

This can be reproduced on any official jsFiddle from the documentation.
1) Go here: https://api.highcharts.com/highcharts/plotOptions.series.dragDrop
2) Select "[Draggable column and line series](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/dragdrop/resize-column)"
3) With Chrome Developer Tools enable "Device Toolbar" and select iPad as device.
4) Just click anywhere on the chart. It throws this:
````
highcharts.src.js:22772 Uncaught TypeError: Cannot read property '0' of undefined
    at f.normalize (highcharts.src.js:22772)
    at x (draggable-points.src.js:1322)
    at HTMLDocument.<anonymous> (draggable-points.src.js:2260)
````
I've been able to track this down to this file: https://github.com/highcharts/highcharts/blob/master/ts/parts/Pointer.ts

In the [line 1194](https://github.com/highcharts/highcharts/blob/39793279986ac351d42b66e7ae789579380f5234/ts/parts/Pointer.ts#L1194): 
````
// iOS (#2757)
const ePos = (
    touches ?
        touches.length ?
            touches.item(0) as Touch :
            touches.changedTouches[0] :
         e as unknown as PointerEvent
);
````

It looks like ``touches.changedTouches`` is ``undefined``, because ``e.touches.changedTouches`` is ``undefined``. It should be ``e.changedTouches`` instead. I tried locally and that simple change seems to fix the problem and make it work for both touch and non-touch devices.

Possible fix:
````
// iOS (#2757)
const ePos = (
    touches ?
        touches.length ?
            touches.item(0) as Touch :
            e.changedTouches[0] :
         e as unknown as PointerEvent
);
````
#### Product version
Highcharts: 8.1.0
Draggable Points: 8.1.0

#### Affected browser(s)
Safari on iPad Pro
Chrome with iPad mode enabled

sebastianbochan:
Fixed #13534, dragging point caused an error on iOS.

----
Being honest, I have no idea how to prepare an unit-test for iOS only.

Any thoughts?

### DIFF
--- a/js/parts/Pointer.js
+++ b/js/parts/Pointer.js
@@ -742,7 +742,8 @@ var Pointer = /** @class */ (function () {
         var ePos = (touches ?
             touches.length ?
                 touches.item(0) :
-                touches.changedTouches[0] :
+                (pick(// #13534
+                touches.changedTouches, e.changedTouches))[0] :
             e);
         // Get mouse position
         if (!chartPosition) {
--- a/ts/parts/Pointer.ts
+++ b/ts/parts/Pointer.ts
@@ -1094,7 +1094,7 @@ class Pointer {
         chartPosition?: Highcharts.OffsetObject
     ): T {
         const touches = (e as TouchEvent).touches;
-    
+
         // iOS (#2757)
         const ePos = (
             touches ?
@@ -1106,7 +1106,6 @@ class Pointer {
                     )[0] :
                 e as unknown as PointerEvent
         );
-
 
         // Get mouse position
         if (!chartPosition) {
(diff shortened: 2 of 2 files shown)


sebastianbochan:
Makes sense, thanks for explanation!

PR #13645, opened 2020-06-04
bugfix/13576-boxplot-series-draggable-values

### LINKED ISSUE #13576: Drag and drop in the boxplot series doesn't work.
#### Expected behaviour
The low, high, q1 and q3 values should be draggable.
As it is in the version 6.2.0: https://jsfiddle.net/BlackLabel/2go6sf43/ 

#### Actual behaviour
Only two of this values are draggable. 

https://jsfiddle.net/BlackLabel/e42uhg80/

#### Product version
Highcharts v8.1.0 

#### Affected browser(s)
All

karolkolodziej:
Fixed #13576, a regression causing drag and drop to fail for some values in the boxplot series.

### DIFF
--- a/js/modules/draggable-points.src.js
+++ b/js/modules/draggable-points.src.js
@@ -473,7 +473,7 @@ if (seriesTypes.boxplot) {
             axis: 'y',
             move: true,
             resize: true,
-            resizeSide: 'draggableQ1',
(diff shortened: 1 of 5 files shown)


TorsteinHonsi:
Pending @oysteinmoseng 's review

PR #13654, opened 2020-06-05
bugfix/13195-dataclasses-hidden-points

### LINKED ISSUE #13195: Tooltip is shown for hidden points when using colorAxis.dataClasses
#### Expected behaviour
Tooltip should not be shown for hidden points (the same behaviour if we just hide serie in legend).

#### Actual behaviour
After user hides points from chart with `colorAxis.dataClasses` used it is still possible to hover over hidden points to get tooltip.

https://jsfiddle.net/Lj6t4590
STR:
1. open demo from link
2. hide points by clicking ranges in the legend
3. move mouse pointer over places where points were present

#### Product version
 Highcharts JS v8.0.4 (2020-03-10) (latest CDN version)

#### Affected browser(s)
Google Chrome 80.0.3987.149 64-bit MacOS

ppotaczek:
Fixed #13195, tooltip was shown for points that were hidden by data classes.

### DIFF
--- a/js/parts-map/ColorSeriesMixin.js
+++ b/js/parts-map/ColorSeriesMixin.js
@@ -33,6 +33,7 @@ H.colorPointMixin = {
                 point[key][method]();
             }
         });
+        this.series.buildKDTree(); // rebuild kdtree #13195
     }
     /* eslint-enable valid-jsdoc */
 };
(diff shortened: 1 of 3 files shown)


sebastianbochan:
Looks good! 👍

PR #13679, opened 2020-06-09
bugfix/13608-parallel-axis-update

### LINKED ISSUE #13608: Updating chart.parallelCoordinates: true doesn't work since 8.1.0
#### Expected behaviour
Worked fine in 8.0.4. I expect the chart working when we update the chart with
```
chart: {
      parallelCoordinates: true
}
```

#### Actual behaviour
The chart disappears after update.

https://jsfiddle.net/BlackLabel/pubng3mz/

#### Product version
8.1.0 (worked fine in 8.0.4)

bre1470:
Fixed #13608, a regression causing parallel axes to lose their positions after update.


pawelfus:
I don't see any related changes? Or it's hidden somewhere between the file re-ogranization?

bre1470:
The logical change can be found in the `else` statement of the `afterSetOptions` listener.

PR #13702, opened 2020-06-12
enhancement/area-3d-series

Izothep:
Enhancement, added support for area series in 3d charts.


ts/parts/AreaSeries.ts
```diff
@@ -520,6 +530,13 @@ seriesType<Highcharts.AreaSeries>(
             }
 
             topPath = getGraphPath.call(this, graphPoints, true, true);
+            if (series.chart.is3d()) {
+                bottomPoints = H.perspective(
+                    bottomPoints as any, this.chart, true
+                ).map(function (point): Highcharts.AreaPoint {
+                    return { plotX: point.x, plotY: point.y, plotZ: point.z } as any;
+                }) as any;
```
sebastianbochan, line 538: Too many casts "as any". What is the reason of not creating an interface of them or use existed i.e Points?

Izothep: I am not sure if you are aware that it is still Draft :) The first PR is made to check if all visual test are passing with new series approach.

sebastianbochan: Mea culpa, just waiting for a final version :) Sorry!

Izothep: However, I think that in general problem of casting to any in Area series is something we should work on. I thought about creating simple interface of point positions as `{x?, y?, z?, plotX?, plotY?, plotZ?}` that will solve needs of both standard and 3d graph in this case.

ts/parts/AreaSeries.ts
```diff
@@ -531,11 +531,20 @@ seriesType<Highcharts.AreaSeries>(
 
             topPath = getGraphPath.call(this, graphPoints, true, true);
             if (series.chart.is3d()) {
+                var options3d = (series as any).chart.options.chart.options3d;
```
Izothep, line 534: Do you have an idea how to remove this `any` cast? We have a case where we are in 3d graph so `options3d` are defined, however, without this casting I would need to make unnecessary check in `if` statement

ts/parts/AreaSeries.ts
```diff
@@ -422,6 +423,7 @@ seriesType<Highcharts.AreaSeries>(
                     options.connectNulls,
                     stacking === 'percent'
                 ),
+                rawPoints = series.rawPoints,
```
Izothep, line 426: relict, I will remove it in the next commit

ts/parts/AreaSeries.ts
```diff
@@ -506,7 +508,7 @@ seriesType<Highcharts.AreaSeries>(
                     // true
                     if (!(isNull && !stacking && connectNulls)) {
                         graphPoints.push(points[i]);
-                        bottomPoints.push({ // @todo make real point object
+                        bottomPoints.push({
```
Izothep, line 511: Bring back the comment

ts/parts/AreaSeries.ts
Izothep: All is moved into new file and casting is significantly reduced.

ts/parts-3d/Chart3D.ts
```diff
@@ -954,6 +954,9 @@ namespace Chart3D {
             if (e.options.type === 'scatter') {
                 e.options.type = 'scatter3d';
             }
+            if (e.options.type === 'area') {
+                e.options.type = 'area3d';
+            }
```
pawelfus, line 959: This keeps increasing. Perhaps we should just switch every `xyz` series to `xyz3d` type (e.g. pie)? Do you see any downside, @TorsteinHonsi ?

ts/parts-3d/Series.ts
```diff
@@ -87,6 +111,7 @@ H.Series.prototype.translate3dPoints = function (): void {
             y: rawPoint.plotY as any,
             z: rawPoint.plotZ as any
         });
+        series.rawPoints = H.extend([], rawPoints);
```
pawelfus, line 114: Any reason to use extend/merge here? 3D is quite heavy for the CPU already and this will just lower the performance.

ts/parts/SVGElement.ts
```diff
@@ -240,7 +240,7 @@ declare global {
             ): void;
             public xGetter(key: string): (number|string|null);
             public yGetter(key: string): (number|string|null);
-            public zIndexSetter(value: number, key: string): boolean;
+            public zIndexSetter(value?: number, key?: string): boolean;
```
pawelfus, line 243: Is that change necessary? What's the point of calling `zIndexSetter` without arguments? Why 3D needs this?

ts/parts-3d/Area.ts
```diff
@@ -0,0 +1,362 @@
...
+     * @optionparent plotOptions.area3d
+     */
+    {
+        tooltip: {
+            pointFormat: 'x: <b>{point.x}</b><br/>y: <b>{point.y}</b><br/>z: <b>{point.z}</b><br/>'
+        }
+
+    // Series class
+    }, {
+        getGraphPath: function (
```
pawelfus, line 95: It seems to be a copy-paste from area with minor changes. Maybe it's time to split `getGraphPath` into minor methods?

samples/highcharts/3d/area/demo.js
```diff
@@ -0,0 +1,31 @@
...
+        }
+    },
+    plotOptions: {
+        series: {
+            depth: 100,
+            groupZPadding: 100
+        }
+    },
+    title: {
+        text: 'a 3D Area Chart'
```
pawelfus, line 24: Don't start with `a` :)

ts/parts-3d/Chart3D.ts
Izothep: I would be with @pawelfus on this one - e.g. now we are creating support for `polygon3d` that should also be added in here.

ts/parts-3d/Series.ts
Izothep: Yes, from line 123 we are changing value of `rawPoint` so I have created a copy to this change in series `rawPoints` array. However, for now I think that easier is to simply create new point object instead of changing `rawPoint` (so the naming will be correct).

ts/parts/SVGElement.ts
Izothep: In shortcut, you don't need `key` in many use-cases.

ts/parts-3d/Chart3D.ts
TorsteinHonsi: The difference between `scatter` and `scatter3d` is the data structure. Scatter takes `[x, y]`, while scatter3d takes `[x, y, z]`. But that is not the case with the area/area3d, is it? As far as I understand the data structure is `[x, y]`, but it is rendered in the 3d room, in which case we shouldn't change the `type`.

Izothep: I would agree in general that there are differences between `scatter3d` and `area3d` related to the data structure, however making separate series type for 3d looks to have more benefits:

- keeping core clean and small
- set some basic series options, like `directTouch`
- chance for further improvements, like adding z support in future and much more.

I have made simple test for adding z value to `area3d` and it works well in some cases without any significant work:
Screenshot 2020-06-19 at 15 35 44

Izothep: @TorsteinHonsi so would you prefer more the approach shown in the old commit? https://github.com/highcharts/highcharts/pull/13702/commits/547d96e230484d8c0125170ab7e28eec4f298aea

TorsteinHonsi: In my understanding, these are two different concepts. I don't think an area series with `[x, y, z]` _for each point_ makes much sense. It would be hard to read and understand an area series like that. Instead, I think what we really want is an area to be rendered on a plane within the 3d space. That means equivalent to the current behaviour of column series in 3D space: https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/3d/column-zpadding/ . Here, the `groupZPadding` is used to place the are in the Z dimension.

Alternatively, we should probably have a `z` value on the series level, so that it can be placed on the Z axis. I think the current `groupZPadding` option was implemented before we even had the Z axis implementation. Then this option should be available for line, column, 2d-scatter etc.

Izothep: Ok, Thanks for the comment! So I will reverse code to the previous approach with `groupZPadding` and without a need to create new series type. In the meantime, I may look at adding `z` option on the Series level but I would like to deal with it separately of this PR as this might cause some zIndexing calculation issues.

PR #13718, opened 2020-06-15
Fixed #13708, a regression breaking Highcharts.keys

### LINKED ISSUE #13708: H.keys doesn't work
#### Expected behaviour
`H.keys()` should work just as `Object.keys()`:

#### Actual behaviour
```
Object.keys({"foo":"bar"})
["foo"]

Highcharts.keys({"foo":"bar"})
Highcharts error #32: www.highcharts.com/errors/32/?Highcharts.keys=Object.keys
Highcharts.keys: Object.keys
highcharts.src.js:2026 Uncaught TypeError: Cannot convert undefined or null to object
    at Arguments.keys (<anonymous>)
    at Object.H.keys (highcharts.src.js:2026)
```

https://jsfiddle.net/j4aq97ds/

Just run, you'll the output from Object.keys() and the error from Highchart.keys()

#### Product version
Highcharts 8.1.1 

#### Affected browser(s)
Tested on Chrome, probably affects all browsers

TorsteinHonsi:
Fixed #13708, a regression breaking `Highcharts.keys`

### DIFF
--- a/js/parts/Utilities.js
+++ b/js/parts/Utilities.js
@@ -1914,9 +1914,9 @@ var find = H.find = Array.prototype.find ?
  * @return {Array<string>}
  *         An array of strings that represents all the properties.
  */
-H.keys = function () {
+H.keys = function (obj) {
     error(32, false, void 0, { 'Highcharts.keys': 'Object.keys' });
-    return Object.keys.apply(arguments);
+    return Object.keys(obj);
 };
 /**
  * Get the element's offset position, corrected for `overflow: auto`.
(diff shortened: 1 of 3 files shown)


sebastianbochan:
I found the same problem with `forEach` here https://github.com/blacklabel/custom_events/issues/143

TorsteinHonsi:
> Is it a global problem with polyfills?

No, it is just that we started raising warnings on using deprecated utility functions. Your plugins should stop using these functions. 

Issue #13708 was different, in that the utility function itself was wrongly implemented.

PR #13730, opened 2020-06-16
bugfix/13532-annotations-editable-issue

### LINKED ISSUE #13532: Label, Circle, Rectangle annotations are not editable in Highcharts 8.1.0
#### Expected behaviour
Annotations of type "label", "circle" and "rectangle" to be editable.

#### Actual behaviour
In version 8.1.0 the mentioned annotations can't be edited.

Basts:
Fixed #13532, a regression in 8.1.0 causing annotations of type "label", "circle" and "rectangle" not to be editable.

~~Fixed #13532, basic annotations editable issue.~~


ts/annotations/annotations.src.ts
```diff
@@ -169,9 +169,11 @@ declare global {
         interface AnnotationsTypeOptions {
             background?: AnnotationsShapeOptions;
             height?: number;
+            labelOptions?: Partial<AnnotationsShapeOptions>;
             line?: AnnotationsShapeOptions;
             point: AnnotationMockPointOptionsObject;
             points?: Array<AnnotationsTypePointsOptions>;
+            shapes?: AnnotationsShapeOptions;
```
pawelfus, line 176: Is this type used later to generate ts definitions for users? If yes, then we can not set it here. We don't have `shapes` and `labelOptions` options available for any annotation type.

ts/annotations/types/BasicAnnotation.ts
```diff
@@ -199,16 +199,29 @@ H.extendAnnotation(
...
+
+                if (options.shapes[0].type === 'circle') {
+                    annotationType = 'cricle';
+                } else if (options.shapes[0].type === 'path') {
+                    annotationType = 'rectangle';
+                }
+            } else {
+                options.typeOptions.labelOptions = options.labelOptions;
+                annotationType = 'label';
+            }
```
pawelfus, line 218: Define `annotationType` in init, and store it for example as `this.basicType`.

ts/annotations/types/BasicAnnotation.ts
```diff
@@ -210,26 +210,41 @@ class BasicAnnotation extends Annotation {
...
+    public init(): void {
+        const options = this.options;
+
+        if (options.shapes) {
+            delete options.labelOptions;
+            if (options.shapes[0].type === 'circle') {
+                this.basicType = 'circle';
+            } else if (options.shapes[0].type === 'path') {
+                this.basicType = 'rectangle';
+            }
```
pawelfus, line 232: For a sanity check, I would just use:
```js
            if (options.shapes[0].type === 'circle') {
                this.basicType = 'circle';
            } else {
                this.basicType = 'rectangle';
            }
```

To avoid situation when `basicType` will be undefined.

PR #13735, opened 2020-06-17
bugfix/13205-range-selector-input-date

### LINKED ISSUE #13205: highstock rangeselector behaviour
#### Expected behaviour
In highstock, In range selector, in mentioned example bellow, when I change 'from' datetime to '2019/12/23 00:00:00.000', the input will change to 7 hours past to '2019/12/22 17:00:00.000'.
Any datetime in 'from' input will have same behaviour. I believe this can be a bug. I could not find any explanation in documents or internet.

#### Actual behaviour
to show '2019/12/23 00:00:00.000' in the input and filter the data properly (which I believe it does filter properly)

https://jsfiddle.net/skd4t382/1/

#### Product version
8.0.4 also in 6.x

#### Affected browser(s)
Chrome Version 80.0.3987.149

karolkolodziej:
Fixed #13205, range selector showed wrong values ​​after update.

### DIFF
--- a/ts/parts/RangeSelector.ts
+++ b/ts/parts/RangeSelector.ts
@@ -1255,8 +1255,15 @@ class RangeSelector {
      */
     public defaultInputDateParser(inputDate: string, useUTC: boolean): number {
         var date = new Date();
+
         if (useUTC) {
+            if (H.isSafari) {
+                return Date.parse(inputDate.split(' ').join('T'));
+            }
             return Date.parse(inputDate + 'GMT');
+        }
+        if (H.isSafari) {
+            return Date.parse(inputDate.split(' ').join('T'));
         }
         return Date.parse(inputDate) - date.getTimezoneOffset() * 60 * 1000;
     }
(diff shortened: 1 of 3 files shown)


karolkolodziej:
I'm not sure how to test this. I will appreciate some advice.

ts/parts/RangeSelector.ts
```diff
@@ -1268,22 +1280,24 @@ class RangeSelector {
...
+            var inputValue = input.value as any,
+                value: (number|Highcharts.RangeSelectorParseCallbackFunction|undefined),
                 chartAxis = chart.xAxis[0],
                 dataAxis = chart.scroller && chart.scroller.xAxis ?
                     chart.scroller.xAxis :
                     chartAxis,
                 dataMin = dataAxis.dataMin,
                 dataMax = dataAxis.dataMax;
 
+            value = (options.inputDateParser || defaultInputDateParser)(inputValue, chart.time.useUTC);
```
pawelfus, line 1299: Good! Now, since we added new argument to this function call (`useUTC`), we also need to update documentation  for `options.inputDateParser` and it's TS definition.

PR #13770, opened 2020-06-25
bugfix/13710-pie-drawempty-regression

### LINKED ISSUE #13710: Error rendering empty pie charts after upgrading to 8.1.1
#### Expected behaviour
An empty pie chart renders without errors.

#### Actual behaviour
An error appears:
`ReferenceError: Highcharts is not defined`
https://github.com/highcharts/highcharts/blob/decfc3ab47ec88e28472f065fd2454b46c9f1fd1/js/parts/PieSeries.js#L789

https://codesandbox.io/s/highcharts-vue-demo-ye65e

#### Product version
8.1.1

#### Affected browser(s)
All

goransle:
Fixed #13710, a regression causing errors on empty pie series.

Seems this one snuck out from the fix in 8.12.

With fix: https://codesandbox.io/s/highcharts-vue-demo-b37nm?file=/src/main.js
Without fix: https://codesandbox.io/s/highcharts-vue-demo-ye65e

### DIFF
--- a/js/parts/PieSeries.js
+++ b/js/parts/PieSeries.js
@@ -9,6 +9,7 @@
  * */
 'use strict';
 import H from './Globals.js';
+import SVGRenderer from './SVGRenderer.js';
 import LegendSymbolMixin from '../mixins/legend-symbol.js';
 import Point from './Point.js';
 import U from './Utilities.js';
@@ -786,7 +787,7 @@ seriesType('pie', 'line',
                     .add(this.group);
             }
             this.graph.attr({
-                d: H.SVGRenderer.prototype.symbols.arc(centerX, centerY, this.center[2] / 2, 0, {
+                d: SVGRenderer.prototype.symbols.arc(centerX, centerY, this.center[2] / 2, 0, {
                     start: start,
                     end: end,
                     innerR: this.center[3] / 2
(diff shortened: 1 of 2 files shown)


pawelfus:
I might be wrong, but shouldn't we:
- import renderer: `import SVGRenderer from '../parts/SVGRenderer.js';`
- then use `SVGRenderer.prototype.symbols` ?

What do you think @bre1470 ?

felipemfp:
@TorsteinHonsi when this fix will be released? I'm asking because in our project we need the vulnerability fixed on latest, but Pie charts are broken without this PR, so it would be awesome if you release patch.

pawelfus:
Next release will be probably in August, date is not specified yet. It's a regression, so you can use either v8.1.0 or modify the source code (replace `Highcharts.SVGRenderer` with `H.SVGRenderer`).

felipemfp:
> Next release will be probably in August, date is not specified yet. It's a regression, so you can use either v8.1.0 or modify the source code (replace `Highcharts.SVGRenderer` with `H.SVGRenderer`).

`v8.1.0` has a [vulnerability issue](https://www.npmjs.com/advisories/1227), and I really didn't want to patch `highcharts`. We had four releases on March, 2 on June... why couldn't we get 1 in July?

bre1470:
Hi @felipemfp!
Consider using the current nightly for development until the next release: `npm install --save highcharts/highcharts-dist#nightly`. See https://github.com/highcharts/highcharts-dist#install-nightly-build
The nightly will stay the same for the next two weeks because of national vacations in Norway.

Issue #13690, opened 2020-06-10
`view data table` option be togglable


Ardeshir81:
# Description of the feature
In the exporting context menu, `view data table` functionality fits better if it is togglable.

# Library related to the feature
Might be Highcharts - Highcarts/exporting

# Proof of Concept/ Live example for the feature
https://codepen.io/Regie/full/OwBJjE
(use view data table option)

# Workaround ?
I used this: https://stackoverflow.com/a/55417530/2908277
as workaround, which is not very safe and depends on Highcharts internals

You can vote for the feature by adding thumbs-up reaction for this post.

pawelfus:
*Internal note:*
Steps:
- implement functionality to remove the data table
- change "View data table" text to e.g. "Hide data table" when table is rendered
- add lang option to change the "Hide data table" text
- in PR, don't forget to list API options that were added/changed

PR #13773, opened 2020-06-25
enhancement/13690-toggle-data-table

karolkolodziej:
Enhancement #13690, added the possibility to toggle the data table.

Added API Chart method called `hideData`.


ts/modules/export-data.src.ts
```diff
@@ -379,7 +380,14 @@ setOptions({
          * @since    6.0.0
          * @requires modules/export-data
          */
-        viewData: 'View data table'
+        viewData: 'View data table',
+        /**
+         * The text for the menu item.
+         *
+         * @since    6.0.0
```
pawelfus, line 387: I'm pretty sure it's not supported since v6 :) Should be `@since next` or `@since v8.1.3`

ts/modules/exporting.src.ts
```diff
@@ -1993,9 +1993,10 @@ extend(Chart.prototype, /** @lends Highcharts.Chart.prototype */ {
                             },
                             innerHTML: (
                                 (item as any).text ||
-                                (chart.options.lang as any)[
-                                    (item as any).textKey as any
-                                ]
+                                (chart.dataTableDiv && item.textKey === 'viewData') ?
```
pawelfus, line 1996: Instead of making decision here, can we change `item.textKey` when table is being generated/hidden? For example `viewData()` can change `textKey` to `viewData` and `hideData()` can change to `hideData`.

ts/modules/export-data.src.ts
```diff
@@ -1162,25 +1171,46 @@ Chart.prototype.downloadXLS = function (): void {
  * @fires Highcharts.Chart#event:afterViewData
  */
 Chart.prototype.viewData = function (): void {
```
pawelfus, line 1173: I don't think we should modify `viewData()`. This is our API method, which, well, should display the table. Now it's toggling the data, so how about this (pseudo code):

```js
Chart.toggleDataTable = function () {
  if (table) {
    this.hideData();
  } else {
    this.showData();
  }
}
```

Then, we link `toggleDataTable` to exporting button and we don't have to worry further. Plus we make equivalent API method for `showData` we already have, plus short method for toggling (can also be in API).

Could you also check why `event:afterViewData` was added? Perhaps we should already implement `afterHideData` event.

karolkolodziej: The `event:afterViewData` was added to focus on the data table when it appears.
But here I'm not sure how this should behave after hiding the data table.
@oysteinmoseng , might you give us an opinion about it?

oysteinmoseng: When we hide the table we would not want to change the focus. So as long as the event is fired after the table is shown we are good, I think

PR #13779, opened 2020-06-26
bugfix/13302-marker-clusters-custom-symbol

### LINKED ISSUE #13302: Clusters - Points with custom marker symbol are not displayed after zoom
#### Expected behaviour
Points should be visible without errors.

#### Actual behaviour
Points are not displayed.
Error in console: `Error: <image> attribute y: Expected length, "NaNNaNNaN".`

1. Open jsfiddle: https://jsfiddle.net/BlackLabel/w160t4eo/
2. Wait 2 seconds

#### Product version
Highcharts v8.0.4

#### Affected browser(s)
All

wchmiel:
Fixed #13302, marker cluster image symbol not displayed after the zoom.

### DIFF
--- a/samples/unit-tests/series/marker-clusters/demo.js
+++ b/samples/unit-tests/series/marker-clusters/demo.js
@@ -379,9 +379,17 @@ QUnit.test('General marker-clusters', function (assert) {
                     point = chart.series[0].points[0];
                     graphicBBox = point.graphic.getBBox();
 
-                    assert.deepEqual(
-                        [point.plotX.toFixed(2), point.plotY.toFixed(2)],
-                        [graphicBBox.x.toFixed(2), graphicBBox.y.toFixed(2)],
+                    result = true;
+
+                    if (
+                        Math.abs(point.plotX - graphicBBox.x) > 1 &&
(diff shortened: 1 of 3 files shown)


samples/unit-tests/series/marker-clusters/demo.js
```diff
@@ -359,6 +359,74 @@ QUnit.test('General marker-clusters', function (assert) {
         resultVisibility,
         'Parent point should be hidden before animation.'
     );
+
+    // (#13302)
+    var done = assert.async(),
+        url = location.host.substr(0, 12) === 'localhost:98' ?
```
pawelfus, line 365: I think it's a bad idea - for example I'm using only `utils.highcharts.com` so this won't work in browser.

Could you use for example `https://www.highcharts.com/samples/graphics/sun.png` ?

wchmiel: I've tried `https://www.highcharts.com/samples/graphics/sun.png` yet it doesn't work in tests. I found this approach in other tests, and for me, it works fine in `utils.highcharts.com`.

pawelfus: Okay then! 👍

PR #13797, opened 2020-06-30
bugfix/13533-negative-xValue-on-log-axis

### LINKED ISSUE #13553: Exception on logarithmic scatter chart with negative x-values
#### Expected behaviour
Points with values <0 are ignored om logarithmic scaled axis when the axis minimum is set above 0.

As stated in the documentation:

> As a consequence, Highcharts will remove 0 or negative points associated to the axis, and if you try to set the axis.min option to 0 or negative, it will fail with an error.

#### Actual behaviour
When the x value of a scatter point is <0 you get Highcharts exception 10.

http://jsfiddle.net/kzoon/cpey9qkf/
Notice that points with negative y-values _are_ ignored without the exception:
http://jsfiddle.net/kzoon/5yvkxqLo/

#### Product version
Highcharts latest, probably earlier as well

#### Affected browser(s)
All

mateuszkornecki:
Fixed #13553, negative xValues on logarithmic axis.


ts/parts/Series.ts
```diff
@@ -5189,11 +5189,13 @@ H.Series = seriesType<Highcharts.LineSeries>(
...
-                    yValue !== null &&
-                    (yValue as any) <= 0
-                ) {
+                // Discard disallowed x and y values for log axes #3434, #13533
+                const isPositiveAxes = yAxis.positiveValuesOnly ||
+                    xAxis.positiveValuesOnly,
+                    isValidValue = (yValue !== null && (yValue as any) <= 0) ||
+                    (xValue && xValue <= 0);
+
+                if (isPositiveAxes && isValidValue) {
```
pawelfus, line 5198: We need to make sure that only nulls/not positive values should set `isNull` flag. Correct my if I'm wrong but this case will be wrong:

- xAxis: log, yAxis: linear -> `isPositiveAxes = true`
- xValue: 1, yValue: 0 -> `isValidValue=true` 

Which means we will set `isNull` to true, but we shouldn't. It's a valid point.

`isValidValue` - that variable name is a bit misleading.

Perhaps it's a good idea, to create `Axis.proto.validateValue(value)` method and call it here and in above `getSeriesExtremes()` (instead of two filters: `isPositive` and isNumber`).

ts/parts/Axis.ts
```diff
@@ -4359,6 +4359,12 @@ class Axis implements AxisComposition, AxisLike {
                     if (axis.isXAxis) {
                         xData = series.xData as any;
                         if (xData.length) {
+                            const isPositive = (number: number): boolean => number > 0;
+
+                            xData = axis.logarithmic ?
+                                xData.filter(isNumber).filter(isPositive) :
```
pawelfus, line 4365: Try to avoid using `Array.filter().filter()` - it's not optimal

samples/unit-tests/axis/type-logarithmic/demo.js
```diff
@@ -264,4 +264,36 @@ QUnit.test('Y axis minimum got stuck (#3353)', function (assert) {
         postUpdatesTick,
         "Y minimum value should not be changed when updating yAxis type"
     );
+});
+
+QUnit.test('Negative values on log axes', function (assert) {
```
pawelfus, line 269: Can we test all possible cases? 

- xAxis: log, yAxis: log:
  - x: positive, y: positive,
  - x: positive, y: negative,
  - x: negative, y: positive,
  - x: negative, y: negative

- xAxis: log, yAxis: linear:
  - x: positive, y: positive,
  - x: positive, y: negative,
  - ...
- ...

Issue #13804, opened 2020-06-30
3d pie chart error when updating on inactive browser tab 


Serg-Maximchuk:
#### Expected behaviour
No errors expected

#### Actual behaviour
Error thrown
![errors](https://user-images.githubusercontent.com/13038578/86141999-5e4ec680-bafb-11ea-870f-d7ca64733615.png)

#### Live demo with steps to reproduce
Steps to reproduce:
* open [this example](https://codesandbox.io/s/highcharts-pie3d-example-6qc7g) **in new tab**
* go back to this tab
* wait 2-3 sec
* switch back to example
* now you should be able to see error

This happens when chart is being updated on inactive tab. Example is based on react wrapper, I don't know how to reproduce it without react, but I think there's nothing to do with react wrapper.

#### Code affected
[this function](https://github.com/highcharts/highcharts/blob/master/js/parts-3d/SVGRenderer.js#L661) is being called [here](https://github.com/highcharts/highcharts/blob/c7e1976467a3b0e1620d44a414af2470716732b9/js/parts/SVGElement.js#L492),  check second argument - that's why it fails [here](https://github.com/highcharts/highcharts/blob/master/js/parts-3d/SVGRenderer.js#L670), that object does not have any `elem` field. 
Related to [this bugfix](https://github.com/highcharts/highcharts/issues/9749)

#### Product version
Highcharts 8.1.2

#### Affected browser(s)
Google Chrome 83.0.4103.116

pawelfus:
Thank you for reporting the issue @Serg-Maximchuk 

**Simplified demo:**
https://jsfiddle.net/BlackLabel/zg8tmsak/

*Internal note:*
It's a bug since v7.0.2, v7.0.1 used to work fine.

Bisected to commit 25506e5375

kraussjo:
I have exactly the same problem with version v8.2.2 (2020-10-22)

I would be happy about a prompt fix.

PR #13805, opened 2020-06-30
#13804 safe function call

Serg-Maximchuk:
[Fix for this issue](https://github.com/highcharts/highcharts/issues/13804)


TorsteinHonsi:
Thanks for the PR! The fix needs to be applied to the `.ts` file and tests should be set up, but I'll leave it here so our devs can see how you resolve the issue.

PR #13818, opened 2020-07-02
bugfix/13785-map-data-exporting

### LINKED ISSUE #13785: Wordmap export based on x and y axis
#### Expected behaviour
Export data

#### Actual behaviour
Export x and y axis with incremental integer

http://jsfiddle.net/wbyn6cx3/

View data table:
Category	Value
x	y
Canada	0	
United States of America	1	
Japan	2	
Russia	3	

#### Product version
Highcharts 8.1.2 (tested on >= 8.1)

#### Affected browser(s)
Tested on Firefix

Basts:
Fixed #13785, wrong CSV output for data without defined `y` value.

### DIFF
--- a/ts/modules/export-data.src.ts
+++ b/ts/modules/export-data.src.ts
@@ -536,8 +536,8 @@ Chart.prototype.getDataRows = function (
             series: Highcharts.Series,
             xAxis: Highcharts.Axis
         ): string[] {
-            const namedPoints = series.data.filter((d): string | 0 | null | undefined =>
-                (d.y === 0 || d.y === null || d.y) && d.name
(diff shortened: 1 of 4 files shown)


ts/modules/export-data.src.ts
```diff
@@ -536,7 +536,8 @@ Chart.prototype.getDataRows = function (
             series: Highcharts.Series,
             xAxis: Highcharts.Axis
         ): string[] {
-            const namedPoints = series.data.filter((d): string | undefined => d.name);
+            const namedPoints = series.data.filter((d): string | 0 | null | undefined => d.y && d.name);
```
pawelfus, line 539: How about `d.y = 0` ? ;)

PR #13821, opened 2020-07-02
docs/generate-url-download

marvin19:
Added function that generates download link for the .zip files.


TorsteinHonsi:
Already covere in another PR

Issue #13798, opened 2020-06-30
Range selector button text color not applied for all buttons.


wchmiel:
#### Expected behaviour
`rangeSelector.buttonTheme.style.color` should be set for all buttons.

#### Actual behaviour
`rangeSelector.buttonTheme.style.color` is set only for the first button text.

#### Live demo with steps to reproduce
v8.1.1 (regression) -> https://jsfiddle.net/BlackLabel/y4zg85vt/
v8.1.0 -> https://jsfiddle.net/BlackLabel/o3es2d1y/

#### Product version
Highstock above v8.1.0

Bisected to commit https://github.com/highcharts/highcharts/commit/d8237eaa4bfa0492a4a3cc59783c10336f75fe0c

PR #13836, opened 2020-07-06
bugfix/13798-range-selector-buttons-text

wchmiel:
Fixed issue with range selector buttons text color, see #13798.


samples/unit-tests/rangeselector/buttons-range-state/demo.js
```diff
@@ -1,3 +1,83 @@
+QUnit.test('Range selector buttons general tests.', function (assert) {
```
TorsteinHonsi, line 1: Actually we don't need a unit test for the button row, because the visual test (stock/rangeselector/styling) picks it up. Instead, we could use a unit test for the actual problem, which lies in the button function mutating its options. So instead of loading a full stock chart, create a much faster and to-the-point test on the `SVGRenderer.button` itself: https://jsfiddle.net/highcharts/hnt7g9zL/

ts/parts/SVGRenderer.ts
```diff
@@ -1620,6 +1620,9 @@ class SVGRenderer {
             ),
             curState = 0,
             styledMode = this.styledMode,
+            // Make a copy of normalState (#13798)
+            // (reference to options.rangeSelector.buttonTheme)
+            normalState = normalState ? H.merge(normalState, {}) : normalState,
```
TorsteinHonsi, line 1625: The merge can be simplified to `merge(normalState)` I think. Also, `merge` is already available in the local scope, so we don't need to access it through `H`.

TorsteinHonsi:
All good, pending tests after merging the master 👍

PR #13852, opened 2020-07-08
bugfix/13816-scatter-update-line-width

### LINKED ISSUE #13816: Scatter plot doesn't redraw when setting lineWidth to 0
#### Expected behaviour
1. I have a scatter plot with one series, and that series has lineWidth=1
2. I use chart.update with redraw=true to change the series' lineWidth=0
3. The chart should redraw so that the line connecting the points of the series isn't visible.

#### Actual behaviour
Instead, after step 2, the line doesn't disappear from the chart until I move my cursor over the chart.

https://jsfiddle.net/wxo5n6hL/21/
1. Run the jsfiddle
2. Click the Connect button - a line should be drawn between the points
3. Click the Disconnect button - I expect the line to disappear, but it doesn't
4. Mouseover the chart - the line disappears

#### Product version
Noticed in v8.1.0. I'm upgrading from Highcharts v6.2.0 to v8.1.0, and did not see this issue before upgrading.

#### Affected browser(s)
I tried out the jsfiddle on Chrome, Firefox, and Edge and got the same results.

I think I've figured out all the other relevant (to me) changes between the two versions, so I'm hoping there's a workaround for this!

wchmiel:
Fixed #13816, scatter plot didn't redraw when updating `lineWidth` to 0.


ts/Core/Interaction.ts
```diff
@@ -1533,6 +1533,7 @@ extend(LineSeries.prototype, /** @lends Highcharts.Series.prototype */ {
                 (series.chart.options.chart as any).animation
             ),
             attribs,
+            lineWidthPlus,
```
TorsteinHonsi, line 1536: This variable is no longer in use, please remove it

ts/Series/ScatterSeries.ts
```diff
@@ -196,7 +196,14 @@ BaseSeries.seriesType<typeof Highcharts.ScatterSeries>(
          * @function Highcharts.seriesTypes.scatter#drawGraph
          */
         drawGraph: function (this: Highcharts.ScatterSeries): void {
-            if (this.options.lineWidth) {
+            if (
+                this.options.lineWidth ||
+                (
+                    this.options.lineWidth === 0 &&
```
TorsteinHonsi, line 202: Add a comment and a reference to the issue number so we know why we do this

TorsteinHonsi (changes requested):
Two minor things to improve, otherwise fine!

TorsteinHonsi (approved):
Thanks!

PR #13861, opened 2020-07-08
tools/folder-core-parts-3

bre1470:
Moved parts files into Core, Extensions and Series folders.


bre1470:
There are no `TS` variants of them, and the JS files are not in use. So these are leftovers of some file movement or consolidation.

@oysteinmoseng 
Do you have objections agains deleting the following `JS` files?
- `js/modules/accessibility/components/InfoRegionsComponent/AnnotationsA11y.js`
- `js/modules/accessibility/components/InfoRegionsComponent/InfoRegionsComponent.js`

oysteinmoseng:
All good to delete those js files. Agree that they are just leftovers from a different file structure.

PR #13878, opened 2020-07-10
Bugfix/sonification tweaks

oysteinmoseng:
Minor fixes, probably no need for changelog as the use cases came up after adding the new sonification options.


oysteinmoseng:
Fails because of issue with test in master. The issue should apparently go away after 6pm so we can try again then, or merge in fix from master when Sophie's PR has been added.

PR #13891, opened 2020-07-13
enhancement/mixins-legend-symbol-es6

marvin19:
Migrated legend-symbol to ES6 pattern.

### DIFF
--- a/ts/Core/Axis/ColorAxis.ts
+++ b/ts/Core/Axis/ColorAxis.ts
@@ -1579,7 +1579,7 @@ namespace ColorAxis {
         chart: Chart;
         name: string;
         options: object;
-        drawLegendSymbol: LegendSymbolMixin['drawRectangle'];
+        drawLegendSymbol: typeof LegendSymbolMixin['drawRectangle'];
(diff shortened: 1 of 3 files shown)


ts/Core/Axis/ColorAxis.ts
```diff
@@ -1579,7 +1579,7 @@ namespace ColorAxis {
         chart: Chart;
         name: string;
         options: object;
-        drawLegendSymbol: LegendSymbolMixin['drawRectangle'];
+        drawLegendSymbol: typeof LegendSymbolMixin['drawRectangle'];
```
bre1470, line 1582: Why is this reduced to a function type?

marvin19: It appeared as a problem in the editor. I got the message 'LegendSymbolMixin' refers to a value, but is being used as a type here. Did you mean 'typeof LegendSymbolMixin'?

bre1470: Okay, thanks.

PR #13896, opened 2020-07-13
bugfix/13846-dataLabels-hidden-pie

### LINKED ISSUE #13846: Certain data labels are not visible on pie chart's init.
#### Expected behaviour
All data labels should be visible from the beginning.

#### Actual behaviour
Certain data labels are not visible on pie chart's init, but they appearing after hovering any of slices. It looks like they're treated as if they would overlap in some way while the `alignTo: 'plotEdges'` is set.

Live example: https://jsfiddle.net/BlackLabel/zdtx6ajp/

Izothep:
Fixed #13846, data labels were initially hidden in some pie chart instances due to overlapping detection.


sebastianbochan:
The unit test is missing.

PR #13906, opened 2020-07-14
bugfix/13694/area-low-value-breaks

### LINKED ISSUE #13694: Area is broken by low value
#### Expected behaviour
Correct area shape.

#### Actual behaviour
The point in a form `{ y: 5, low: 3 }` breaks the area series, which should be draw according to series' threshold, not `point.low` value.

https://jsfiddle.net/BlackLabel/6wva4nt9/

#### Product version
8.1.1

#### Affected browser(s)
All

Basts:
Fixed #13694, threshold wasn't respected when low was defined.

### DIFF
--- a/ts/Series/AreaSeries.ts
+++ b/ts/Series/AreaSeries.ts
@@ -480,7 +480,7 @@ seriesType<Highcharts.AreaSeries>(
 
                 isNull = points[i].isNull;
                 plotX = pick(points[i].rectPlotX, points[i].plotX);
-                yBottom = pick(points[i].yBottom, translatedThreshold);
+                yBottom = translatedThreshold;
 
                 if (!isNull || connectNulls) {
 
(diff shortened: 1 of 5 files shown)


ts/Series/AreaSeries.ts
```diff
@@ -480,7 +480,7 @@ seriesType<Highcharts.AreaSeries>(
 
                 isNull = points[i].isNull;
                 plotX = pick(points[i].rectPlotX, points[i].plotX);
-                yBottom = pick(points[i].yBottom, translatedThreshold);
+                yBottom = stacking ? points[i].yBottom : translatedThreshold;
```
pawelfus, line 483: Just a question: is `yBottom` always defined when stacking is set? Or can be e.g. null? On the left we used `pick()`. Should we do the same here? `stacking ? pick(points[i].yBottom, translatedThreshold) : translatedThreshold;`

Basts: I checked the code and find this part: https://github.com/highcharts/highcharts/blob/master/ts/Core/Series/Series.ts#L5286. Correct me if I am wrong, but it seems that `yBottom` is always defined when stacking is set.

pawelfus: Looks good, thanks for checking! 👍

PR #13912, opened 2020-07-15
enhancement/data-layer

bre1470:
To-Do:
- [x] DataTable
- [x] DataTableRow
- [x] DataConverter converts between primitives, Date, and DataTable
- [x] DataJSON to support serialization of any instance
- [x] DataModifier
- [x] ChainModifier
- [ ] ColumnModifier
- [ ] FlattenModifier
- [x] GroupModifier
- [x] RangeModifier
- [ ] SeriesDataModifier
- [x] CSVStore
- [x] GoogleStore
- [x] HTMLTableStore
- [x] CSVParser
- [x] GoogleSheetsParser (DataJSON support is missing)
- [x] HTMLTableParser
- [x] unit-tests
- [x] demo samples
- [x] Data module makes use of data layer
- [x] SeriesDataConverter creates DataTable from data option and vice versa
- [ ] Removing `Data` from class/file names in inherited modifiers, parsers, and stores
- [x] Events have to support context identifier (Event.detail: Record<string, string>)
- [x] DataTable and DataTableRow have to take a context identifier or provide their own id as context to events
- [x] DataStore has to provide a context identifier (Event.detail)
- [x] DataStore needs static function and properties to register Store class and create instances
- [ ] Utilities.fireEvent support optionally zero-timeout calls
- [ ] Series makes use of data layer
- [ ] Series should understand and listen to DataTable
- [ ] Series.update should go to SeriesDataConverter
- [ ] SeriesDataConverter needs to fire events instead of serial returning results


ts/Data/DataStore.ts
```diff
@@ -0,0 +1,90 @@
...
+    ): void {
+
+    }
+
+}
+
+export interface MetaColumn {
+    name?: string;
+    metadata?: any;
+}
```
bre1470, line 88: This needs to be wrapped in a namespace named `DataStore`, so that it becomes accessible via the default export (see DataTable for an example). The default export is the only mechanism supported by the Highcharts assembler. TypeScript will automatically merge the namespace types with the previous class type of the same name.

sebastianbochan: I missed it by accident.

ts/Data/GoogleDataStore.ts
```diff
@@ -0,0 +1,227 @@
...
+                text: (string|Error)
+            ): void {
+                /* *
+                 * TODO:
+                 * catch error
+                 * ...
+                 *
+                 * */
+                // console.log(text);
+            }
```
bre1470, line 198: We need something here, to inform anyone when the async load is finished and the dataTable is available.

* One way would be a callback parameter taken by `load` and passed through to `fetchSheet`. The `callback` function should receive either a `DataTable` instance or an `Error` instance.

* Alternatively we can fire events `load`/`afterLoad`, `parse`/`afterParse` and `fail`, which would be similar to `DataTable` events and maybe more flexible. Any opinions?

ts/Data/GoogleDataStore.ts
```diff
@@ -0,0 +1,227 @@
...
+        });
+
+        return true;
+    }
+
+    public load(): boolean {
```
bre1470, line 204: The return type of `load` should be void, so that we later can replace it with a promise, when a refactoring is possible.

ts/Data/GoogleDataStore.ts
sebastianbochan: Maybe we should consider to use `await / async` ?

From two suggestions, I am closer to the solutuon based on events. It sounds better because of consistency (of both modules).

Any other comments here?

bre1470: We unfortunately can not, because of our browser target ES5. `await` / `async` and `Promise` are not part of it.

ts/Data/GoogleDataStore.ts
sebastianbochan: Done

ts/Data/DataStore.ts
sebastianbochan: Done

ts/Data/GoogleDataStore.ts
sebastianbochan: Ahh indeed, I always do not remember about that. 

So well, events sounds better than callbacks, but maybe lets wait for other comments?

wchmiel: In my opinion, events would be a better choice.

sebastianbochan:
@goransle,
Please lint the code ;)

ts/Data/GoogleDataStore.ts
wchmiel: So, we will try to implement events for now if no other opinions.

ts/Data/DataStore.ts
```diff
@@ -80,6 +85,21 @@ abstract class DataStore {
...
+            for (let i = 0; i < noOfRows; ++i) {
+                const row = new DataRow();
+                for (let j = 0; j < noOfCols; ++j) {
+                    row.insertColumn((headers.length ? headers[j] : uniqueKey()), cols[j][i]);
+                }
+                table.insertRow(row);
+            }
+        }
+        return table;
+    }
```
bre1470, line 102: In my opinion this should not be part of the DataStore class. It is not related to store functionality, but more to parsing. It should better be outsourced into a new `DataParser` module file similar to the DataConverter class.

Beside that the variable names should be clearer to avoid confusion. Instead of `noOfRows` use `rowsLength`. Use `columns` instead of `cols`. (See https://en.wikipedia.org/wiki/Col)

ts/Data/CSVDataStore.ts
```diff
@@ -0,0 +1,430 @@
...
+             */
+            function push(): void {
+                if (startColumn > actualColumn || actualColumn > endColumn) {
+                    // Skip this column, but increment the column count (#7272)
+                    ++actualColumn;
+                    token = '';
+                    return;
+                }
+
+                if (!isNaN(parseFloat(token)) && isFinite(token as any)) {
```
bre1470, line 161: No any casts like this, please.

ts/Data/GoogleDataStore.ts
```diff
@@ -172,17 +192,27 @@ class GoogleDataStore extends DataStore {
...
+                                function (): void {
+                                    store.fetchSheet();
+                                },
+                                dataRefreshRate * 1000
+                            );
+                        }
+
+                        fireEvent(store, 'afterLoad', { table });
+                    }
+                );
```
bre1470: The event firing looks good, but I see no `on` function to attach a listener. Additionally the typing for the callback functions and the event properties are missing. See DataTable and DataRow for examples.

ts/Data/DataParser.ts
```diff
@@ -0,0 +1,45 @@
...
+    uniqueKey
+} = U;
+
+class DataParser {
+
+    /**
+     * Functions
+     */
+
+    public columnArrayToDataTable(columns: Array<Array<Highcharts.DataValueType>>, headers: string[] = []): DataTable {
```
bre1470, line 28: Referencing the Highcharts namespace (which exists only for migration) makes it not a real ES6 module.

ts/Data/GoogleDataStore.ts
```diff
@@ -0,0 +1,268 @@
...
+    public columns: Array<Array<Highcharts.DataValueType>>;
+
+    private dataParser = new DataParser();
+
+    /* *
+    *
+    *  Functions
+    *
+    * */
+    private getSheetColumns(json: Highcharts.JSONType): Array<Array<Highcharts.DataValueType>> {
```
bre1470, line 74: We need to integrated the types somewhere, so that we do not use the global Highcharts namespace.

ts/Data/Stores/CSVDataStore.ts
```diff
@@ -0,0 +1,225 @@
...
+        let metadata;
+
+        for (let i = 0, iEnd = this.metadata.length; i < iEnd; i++) {
+            metadata = this.metadata[i];
+
+            json.metadata.push([
+                metadata.name,
+                metadata.metadata
+            ]);
+        }
```
bre1470, line 195: @wchmiel 
This code can be found in all store implementations and therefor should go into the `DataStore` class as a `protected getMetadataJSON(): DataStore.MetadataJSON` function.

sebastianbochan: We just discuss it and will refactor ;) 

Thanks!

ts/Data/Stores/DataStore.ts
```diff
@@ -49,6 +50,41 @@ class DataStore {
...
+            if (elem instanceof Array && typeof elem[0] === 'string') {
+                metadata.push({
+                    name: elem[0],
+                    metadata: elem[1]
+                });
+            }
+        }
+
+        return metadata;
+    }
```
bre1470, line 86: Thanks! This looks good. One thing: This should become static and moved to the top of the class, as it is no requiring an instance.

ts/Data/Stores/GoogleDataStore.ts
```diff
@@ -0,0 +1,331 @@
...
+        }
+
+        // Insert null for empty spreadsheet cells (#5298)
+        for (i = 0; i < colCount; i++) {
+            for (j = 0; j < cellCount; i++) {
+                if (typeof columns[i][j] === 'undefined') {
+                    columns[i][j] = null as any;
+                }
+            }
+        }
```
goransle, line 184: This loop seems to be causing some issues: 
image.

This works:
```js
	columns.forEach(function (
            column: Array<DataValueType>
        ): void {
            if(column.length){
                for (i = 0; i < column.length; i++) {
                    if (typeof column[i] === 'undefined') {
                        column[i] = null as any;
                    }
                }
            }
        });
```
Which makes it seem that we are adding additional columns somewhere.

bre1470: Maybe `colCount` is set to early?

sebastianbochan: Will check it

ts/Data/Stores/DataStore.ts
```diff
@@ -114,43 +185,87 @@ implements DataEventEmitter<TEventObject>, DataJSON.Class {
...
+        metadata?: Partial<MetadataType>;
+    }
+
+    /**
+     * A metadata object
+     */
+    export type MetadataType = {
+        title: string;
+        dataType: DataJSON.Types;
+        // validator: Function;
```
goransle, line 267: Maybe a topic for the next meeting: how should we handle this and other functions in options?

bre1470: Yes, I agree. We should discuss this and also start with some guidelines in the readme to solve these kind of problems.

ts/Data/Parsers/HTMLTableParser.ts
```diff
@@ -59,15 +76,30 @@ class HTMLTableParser extends DataParser<DataParser.EventObject> {
...
-        this.tableElement = tableElement;
+
+        if (tableElement) {
+            this.tableElement = tableElement;
+            this.tableElementID = tableElement.id;
+        } else if (options?.tableHTML) {
+            this.tableElement = options.tableHTML;
+            this.tableElementID = options?.tableHTML.id;
+        }
     }
```
goransle: Thoughts on the parser constructors? Should we try to stick with `constructor(element, options)` or just `constructor(options)`.

Currently the CSVParser constructor only deals with an options object. The HTMLTableParser does both the element to parse and options (with the element also being an option, which should probably change).

bre1470: It could be beneficial to share constructor arguments between parsers. The HTMLTableParser is probably the exception when it comes to the need of a DOM element. Maybe just switch the arguments order?

ts/Data/DataTable.ts
```diff
@@ -371,6 +384,70 @@ class DataTable implements DataEventEmitter<DataTable.EventObject>, DataJSON.Cla
...
+     * Retrieves the given columns, either by the canonical column ID,
+     * or by an alias
+     *
+     * @param {...string} columnIDOrAlias
+     * IDs or aliases for the columns to get, aliases taking precedence.
+     *
+     * @return {Array<Array<DataTableRow.CellType>>}
+     * A two-dimensional array of the specified columns
+     */
+    public getColumns(...columnIDOrAlias: Array<string>): Array<Array<DataTableRow.CellType>> {
```
bre1470, line 397: please use `columnNameOrAlias`

ts/Data/DataTable.ts
```diff
@@ -371,6 +384,70 @@ class DataTable implements DataEventEmitter<DataTable.EventObject>, DataJSON.Cla
...
+        const columns = this.toColumns(),
+            { aliasMap } = this;
+
+        const columnIDs = Object.keys(columns),
```
bre1470, line 401: These are column names, not IDs.

ts/Data/DataTable.ts
```diff
@@ -93,25 +93,30 @@ class DataTable implements DataEventEmitter<DataTable.EventObject>, DataJSON.Cla
...
 
                 if (!columnsObject[cellName]) {
                     columnsObject[cellName] = [];
                 }
-                let cellValue: DataJSON.Types;
+                let cellValue;
 
                 if (cell instanceof DataTable) {
-                    cellValue = cell.toJSON();
+                    cellValue = JSON.stringify(cell.toJSON());
```
bre1470, line 117: Why is the table becoming a string?

ts/Data/DataTable.ts
```diff
@@ -0,0 +1,902 @@
...
+    }
+
+    /**
+     * Converts the datatable instance to a record of columns
+     * @return {DataTable.ColumnCollection}
+     * A record of columns
+     */
+    public toColumns(): DataTable.ColumnCollection {
+        return DataTable.toColumns(this);
+    }
```
bre1470, line 397: I find this confusing. We should decide, if this is a static function or instance function. I favor instance like with toJSON and toString, so the implementation of `DataTable.toColumns` would move to this. What do you think?

goransle: Agree, I mostly made it static to match `fromColumns` but I don't see any use cases where it is necessary.

test/typescript-karma/Data-Modifiers/InvertModifier/tests.js
```diff
@@ -0,0 +1,113 @@
...
+        result2,
+        rowId,
+        columnName;
+
+    assert.ok(
+        invertedTable !== table,
+        'The inverted table should be a new table instance.'
+    );
+
+    assert.ok(
```
bre1470, line 50: It is better to use `assert.strictEqual` here.

ts/Data/DataTable.ts
```diff
@@ -0,0 +1,902 @@
...
+    public deleteRow(
+        rowId: string,
+        eventDetail?: DataEventEmitter.EventDetail
+    ): (DataTableRow|undefined) {
+        const row = this.rowsIdMap[rowId],
+            index = this.rows.indexOf(row);
+
+        this.emit({ type: 'deleteRow', detail: eventDetail, index, row });
+
+        this.rows[index] = row;
```
wchmiel, line 308: Why row is still a rows array element after delete? Is it correct?

bre1470: Good question. I guess it is incorrect.

bre1470: Fixed,

ts/Data/DataConverter.ts
```diff
@@ -197,6 +547,22 @@ namespace DataConverter {
...
         boolean|null|number|string|Date|DataTable|undefined
     );
+
+    /**
+     * The shared options for all DataParser instances
+     */
+    export interface Options extends DataJSON.JSONObject {
+        dateFormat?: string;
+        alternativeFormat?: string;
+    }
```
bre1470, line 557: If these refactored options are used only by parsers, it should be part of the base class there and not be part of the data core.

wchmiel: Options are used by DataConverter. The comment is wrong, should be `Internal options for DataConverter`. I will fix it.

bre1470:
Failing with `series.adjustForMissingColumns is not a function` locally and in CI, when running `npx cross-env NODE_OPTIONS=--max_old_space_size=3072 karma start test/karma-conf.js --single-run --splitbrowsers Win.Chrome --browsercount 2` locally. Data layer branch is possibly not working in Windows Chrome.

bre1470:
@goransle Last commit continues to fail in CI.

goransle:
> @goransle Last commit continues to fail in CI.

Does not seem related to the actual commit. Same error as your previous comment (but with safari). Did merging in the master fix that one?

bre1470:
@goransle Okay, we will see it with the next commit. 👀

bre1470:
@goransle It is working now in the new commit, so I guess it was some problem with the CI cache.

ts/Data/DataSeriesConverter.ts
```diff
@@ -0,0 +1,267 @@
...
+        return {
+            y: column
+        };
+    }
+
+    getRangePoint(column: number): PointOptions {
+        return {
+            y: column
+        };
+    }
```
bre1470, line 130: Why are these separate functions?

ts/Data/DataSeriesConverter.ts
```diff
@@ -0,0 +1,267 @@
...
+}
+/* *
+ *
+ *  Namespace
+ *
+ * */
+namespace DataSeriesConverter {
+    export interface Options {
+        type?: string;
+        columnMap?: Record<string, string>;
```
bre1470, line 259: One usage difficulty with `columnMap` is, that you can not map a column index. So you can not simply convert `Series.parallelArrays` to a map, where the column index is equal to the array index.

bre1470 (commented):
I struggle making use of the DataSeriesConverter. It needs doclets and probably improvements. Any input?

ts/Data/DataSeriesConverter.ts
sebastianbochan: The idea is: distinguish the point type. I.e getRangePoint will return `y` and `yLow` values.  

I agree that the `getPiePoint` is unnecessary.

PR #13916, opened 2020-07-16
enhancement/moving-and-renaming-mixins

marvin19:
Renamed the filenames in the mixin-folder to CamelCase and moved them according to new file structure.


bre1470:
Looks like some mixin references are not updated to capital case.

PR #13919, opened 2020-07-16
bugfix/13914-axis-negative-log

### LINKED ISSUE #13914: Logarithmic axis - negative values.
#### Expected behaviour

According to this [article](https://www.highcharts.com/blog/snippets/alternative-maths-plotting-negative-values-logarithmic-axis/), the negative values should be displayed (thanks to this custom code), however, something is broken after the 8.0.4 version. 

Last working version: http://jsfiddle.net/BlackLabel/bj3dwe4t/

#### Actual behaviour

The current version, the negative points are hidden: http://jsfiddle.net/BlackLabel/cdguh2zp/

#### Product version

Highcharts JS v8.1.2 (2020-06-16)

bre1470:
- Fixed #13914, negative logarithmic values not visible.
- Tools: Updated moved file references.


ts/Core/Axis/LogarithmicAxis.ts
```diff
@@ -282,3 +287,13 @@ interface LogarithmicAxis extends Axis {
 LogarithmicAxis.compose(Axis); // @todo move to factory functions
 
 export default LogarithmicAxis;
+
+/**
+ * Activates rendering of negative logarithmic values.
+ *
+ * @type      {boolean}
+ * @since     8.1.0
+ * @apioption xAxis.allowNegativeLog
```
Basts, line 296: Shouldn't this option be related only to the yAxis (not xAxis)?

ts/Core/Axis/LogarithmicAxis.ts
```diff
@@ -282,3 +287,13 @@ interface LogarithmicAxis extends Axis {
...
+ */
```
pawelfus, line 297: Can we add a demo here? It's not like setting `allowNegativeLog` works out of the box, but some extra code is needed, right?

bre1470: I would expect this to work out of the box. Maybe we should adjust the usage of the `lin2log` function instead than having an other hackish sample. The `lin2log` and `log2lin` function are supposed to move to the composition instance, so this will break with HC 9.

pawelfus: Previously it wasn't an API option, rather a plugin - so it was fine. With these changes perhaps we should probably provide full support then? Also, maybe it's a good idea to update related article with v9+ config?

ts/Core/Axis/LogarithmicAxis.ts
bre1470: Every axis could be logarithmic: http://jsfiddle.net/wgy5qnf9/
You just have to use point objects, because the array index is not sufficient.

Basts: Ahh, you are right, my mistake!

bre1470:
I do not see, how the visual test results are related to the changes of the log function. Somehow the axis scale changes a bit.

ts/Core/Axis/LogarithmicAxis.ts
```diff
@@ -204,11 +209,27 @@ class LogarithmicAxisAdditions {
...
-        return Math.pow(10, num);
+        const isNegative = num < 0 && this.axis.options.allowNegativeLog;
+
+        let result = Math.pow(10, Math.abs(num));
+
+        if (result < 10) {
+            result = (10 * (result - 1)) / (10 - 1);
+        }
+
+        return isNegative ? -result : result;
```
pawelfus, line 220: This is the reason why all demos has changed, compare values in console: 
- current: https://jsfiddle.net/BlackLabel/8yrn47mj/
- candidate: https://jsfiddle.net/BlackLabel/y56fsam1/

I don't know the reasoning behind these normalizations, but perhaps we should simply use `Math.pow(10, num);` when `allowNegativeLog` is not defined. I'm referring to both `lin2log` and `log2lin` functions here :)

bre1470:
Added sample. Please merge, @oysteinmoseng !

Issue #13359, opened 2020-04-22
Gantt chart with yAxis scrollbar and grids multiplies the scrollbar


JohanBrodinSinch:
#### Expected behaviour
Only one scrollbar is expected

#### Actual behaviour
Each grid is causing the chart to add an extra scrollbar. So four grid columns renders four scrollbars.

#### Live demo with steps to reproduce
https://jsfiddle.net/JohanBrodin/bwc183gp/4/

#### Product version
Latest. Did use one of your demos

#### Affected browser(s)
Tested on Chrome but assume all

<img width="830" alt="Screenshot 2020-04-22 at 18 18 30" src="https://user-images.githubusercontent.com/50442378/80007195-3a04b700-84c6-11ea-9061-e7256e5d4a49.png">

JohanBrodinSinch:
Did read below workaround suggestion. However where should you add the corresponding min and max? Independently if you put it under yAxis or in the first column the scrollbar is always 100% extended. So not a workaround if you cannot use the scrollbar.

https://github.com/highcharts/highcharts/issues/10641

sebastianbochan:
Hi @JohanBrodinSinch,
Thank you for reporting. It looks like a bug.

As a workaround you can disable scrollbars in each grid.

Demo:
- https://jsfiddle.net/BlackLabel/fb712kys/

FYI @oysteinmoseng

JohanBrodinSinch:
Thx for your quick replay! Yes it is working fine with your workaround. 

For me this also looks like a bug. You cannot do scrolling per grid so having to disable them individually is not an expected user behaviour.

sebastianbochan:
Hi @JohanBrodinSinch,
I found also a simpler solution: declare scrollbar in the grid, which should include it.

Demo:
- https://jsfiddle.net/BlackLabel/tsmhyLug/

As a result you do not need to disable in others grids.

I agree that is a bug.

PR #13920, opened 2020-07-16
bugfix/13072-gantt-ticks-while-scrolling

### LINKED ISSUE #13072: Gantt Chart Scrolling looks bad
#### Expected behaviour
Scrolling is smooth and text is between the expected lines.

#### Actual behaviour
It is behaving oddly. I think in general there’s a lag in response. Is there a way to speed it up? I have included three screenshots and a video of the problem on our real project.

Simplified problem below;

https://jsfiddle.net/BlackLabel/L78mkans/

#### Affected browser(s)
Chrome

Files describing problem more:

scroll-disappearingNameAndLine
scroll-overlapping-name
scrollOverlappingDatetime

karolkolodziej:
Fixed #13072, ticks during scrolling were missing.

Todo:
- [x] Fix missing ticks on the `yAxis`.
- [ ] Check/fix ticks on the `yAxis` with the `grid`.
- [x] Check/fix the flashing labels on the `xAxis`.
- [x] Extra borderline and missing one.
- [x] Misplaced vertical grids.
- [x] Extra tick label at the top


ts/Core/Axis/TreeGridAxis.ts
```diff
@@ -902,10 +903,12 @@ namespace TreeGridAxis {
 
             return Object.keys(axis.treeGrid.mapOfPosToGridNode || {}).reduce(
                 function (arr: Array<number>, key: string): Array<number> {
-                    var pos = +key;
+                    var pos = +key,
+                        roundedMin = Math.floor(axis.min / axis.tickInterval) * axis.tickInterval,
+                        roundedMax = Math.ceil(axis.max / axis.tickInterval) * axis.tickInterval;
```
pawelfus, line 908: Instead of defining in `reduce()`, could we define in in the higher scope? To calculate this just once.

karolkolodziej: That is a good point. I will change in the next commit.

ts/Core/Axis/Axis.ts
```diff
@@ -7250,10 +7250,13 @@ class Axis implements AxisComposition, AxisLike {
         const isLinked = axis.isLinked;
         const ticks = axis.ticks;
         const slideInTicks = axis.chart.hasRendered && isNumber(axis.oldMin);
+        const tickInterval = axis.tickInterval;
 
         // Linked axes need an extra check to find out if
         if (!isLinked ||
-            (pos >= (axis.min as any) && pos <= (axis.max as any))
+            (pos >= (axis.min as any) && pos <= (axis.max as any)) ||
+            axis.grid &&
```
pawelfus, line 7258: We should avoid polluting core with product-specific code

ts/Core/Axis/Axis.ts
```diff
@@ -7250,10 +7250,13 @@ class Axis implements AxisComposition, AxisLike {
...
+            (pos >= (axis.min as any) - 0.5 * tickInterval && pos <= (axis.max as any) + 0.5 * tickInterval)
```
pawelfus, line 7259: Hmm.. where this `0.5` comes from?

karolkolodziej: To avoid collisions with the column title text we have to hide the grid name exactly at the middle of the first line.

ts/Core/Axis/Axis.ts
karolkolodziej: You are right. I will try to find another solution!

karolkolodziej: I'm going to examine closely #13359, they might be connected.

pawelfus: Thanks!

ts/Core/Axis/Axis.ts
pawelfus: I'm sorry, I don't quite understand.. Anyway, should be moved out from here if possible :)

karolkolodziej:
I went once again through that issue and I'm afraid I can't find another solution.
The left-hand axis(column), is linked to the axis with index 0.
```
if (!isLinked || (pos >= axis.min && pos <= axis.max)) {
  ...
  ticks[pos].render(i);
}
```
Because of that and the conditions in the Axis.ts file in the 'renderTick' method, we are only allowed to render ticks on this axis which are divided by 1.
For example, the new min is 0.1, and the positions of ticks(pos) are 0 and 1. Only tick with the value of 1 is going to pass this condition.
This condition is correct for the ordinary charts because we can somehow change the tick interval/amount but in the Gantt during scroll, every tick maters.
I agree with @pawelfus, that the Axis.ts file is not a perfect place for these changes, but I'm afraid it's the only solution for this time.
@Basts maybe you have another perspective on this issue?
Working example: https://jsfiddle.net/BlackLabel/juk5zg6f/

pawelfus:
If you can't find any other solution, then perhaps it's a bug in core, not in Gantt. Just tested categories+scrollbar in Highstock: https://jsfiddle.net/BlackLabel/7uo3cr69/ - looks equally wrong. Similar in Highcharts: https://jsfiddle.net/BlackLabel/7uo3cr69/1/

Regarding this extra `0.5` - check `Axis.tickmarkOffset` and `Axis.minPointOffset`(?) properties. Using one of these options might be better than calculating `0.5 * tickInterval`. Probably `tickmarkOffset` is the one I'm thinking of: ticks are not translated if tickInterval is different than `1`. No translation = no issue, right? https://jsfiddle.net/BlackLabel/juk5zg6f/2/

PS: There's still some bugs:
- extra label: Screen Shot 2020-07-27 at 10 59 32

- extra border line and missing one: Screen Shot 2020-07-27 at 11 03 55

- misplaced vertical grids (might be unrelated): 
Screen Shot 2020-07-27 at 11 07 02

karolkolodziej:
You are right, I can change the `0.5` to `tickmarkOffset` it looks reasonable!

Maybe, I should create custom logic for the vertical axis to generate these 'ticks' in the grid instead of using actual ticks.
At least, I think I should create a logic to handle the upper and lower border, like in the example below.
There is still an issue with the extra-label and the misplaced vertical grids.

I'm still not quite sure how this should behave, check the screen below, and tell me which in your opinion looks the best.
1.
Screenshot 2020-07-27 at 15 21 50
2.
Screenshot 2020-07-27 at 15 26 17
3. 
Screenshot 2020-07-27 at 15 29 36


Live example: https://jsfiddle.net/BlackLabel/8f63pxzu/

pawelfus:
> Maybe, I should create custom logic for the vertical axis to generate these 'ticks' in the grid instead of using actual ticks.

I don't know.. We already have something similar for the first tick. Check `Axis.render()`. When we use existing ticks we get nice animation for free, right? For example when clicking on scrollbar's buttons. And we don't have to worry about options that are set (lineWidth, colors etc) for ticks or for a gridline.

> I'm still not quite sure how this should behave, check the screen below, and tell me which in your opinion looks the best.

I would say second one. First one has problem that tick is visible, but gridline is missing. The third one has missing blue point - we don't want to change that (`cropThreshold` should control this points visibility). I don't know if we extra line between partial category and yAxis column titles. @oysteinmoseng - what do you think?

oysteinmoseng:
Yes, agree with @pawelfus and prefer the second image for the behavior when scrolling. Keeping the extra lines is good for illustrating to the user that they are scrolling, so I like them. I think it would look weird if the row suddenly expanded to take up more space if you scroll the neighboring point out of view.

PR #13921, opened 2020-07-16
bugfix/13789-polar-connect-ends

### LINKED ISSUE #13789: polar - connectEnds issue
#### Expected behaviour

There shouldn't be any additional path between last and the first pair of points for the `arearange` series. 

#### Actual behaviour

When the `connectEnds` property is set to false, the chart is not rendering properly - some additional path is added between last and the first pair of points. 

Demo: https://jsfiddle.net/BlackLabel/5yz4gsjc/

#### Product version

Highcharts JS v8.1.2 (2020-06-16)

PaulDalek:
Fixed #13789, an incorrect arearange path on polar chart when the `connectEnds` option is set to false.

### DIFF
--- a/js/Series/AreaRangeSeries.js
+++ b/js/Series/AreaRangeSeries.js
@@ -179,24 +179,29 @@ seriesType('arearange', 'area', {
      * @private
      */
     getGraphPath: function (points) {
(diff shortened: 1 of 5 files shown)


oysteinmoseng:
Leaving the branch in case we want to amend for `chart.inverted` and reverse axes etc.

PaulDalek:
Yes, inverted is broken as the `arearange` itself needs to be fixed to work correctly when `chart.inverted` is set to true, so as for now it will work for non-inverted charts only.

PR #13922, opened 2020-07-16
bugfix/13762-polar-labels-use-html

### LINKED ISSUE #13762: Spiderweb/Windrose/Radar charts not displaying all x labels
#### Expected behaviour
Given: I have a spiderweb chart
When: I enable `useHTML` on the chart labels
Then: I should see all labels on the spiderweb chart

#### Actual behaviour
It is only drawing half of the labels.
This also happens in windrose and Polar (radar) chart

https://jsfiddle.net/panintelligence/kqt0r9w1/

#### Product version
We updated from 7.2.0 to 8.1.0 and the problem started to appear.
This is also a problem in the last version

PaulDalek:
Fixed #13762, missing `xAxis` labels on a polar chart when the `useHTML` option was set to true.

### DIFF
--- a/ts/Extensions/OverlappingDataLabels.ts
+++ b/ts/Extensions/OverlappingDataLabels.ts
@@ -190,7 +190,7 @@ Chart.prototype.hideOverlappingLabels = function (
                     left: '0',
                     center: '0.5',
                     right: '1'
-                }[label.alignValue as Highcharts.AlignValue];
+                }[label.alignValue as Highcharts.AlignValue || label.textAlign];
 
                 if (alignValue) {
                     xOffset = +alignValue * boxWidth;
@@ -199,7 +199,8 @@ Chart.prototype.hideOverlappingLabels = function (
                 }
 
                 return {
-                    x: pos.x + (parent.translateX || 0) + padding - xOffset,
+                    x: pos.x + (parent.translateX || 0) + padding -
+                        (xOffset || 0),
                     y: pos.y + (parent.translateY || 0) + padding -
                         lineHeightCorrection,
                     width: label.width - 2 * padding,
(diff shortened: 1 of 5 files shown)


ts/Extensions/OverlappingDataLabels.ts
```diff
@@ -199,7 +199,8 @@ Chart.prototype.hideOverlappingLabels = function (
                 }
 
                 return {
-                    x: pos.x + (parent.translateX || 0) + padding - xOffset,
+                    x: pos.x + (parent.translateX || 0) + padding -
+                        (xOffset || 0),
```
oysteinmoseng, line 203: Is this for `xOffset === NaN`? It is initialized to 0 above it seems

PaulDalek: Yes but there are some cases where `alignValue`, `textAlign` and `translateX` properties can be undefined therefore the expression in the `else if` section can result in `NaN`.

ts/Extensions/OverlappingDataLabels.ts
```diff
@@ -190,7 +190,7 @@ Chart.prototype.hideOverlappingLabels = function (
                     left: '0',
                     center: '0.5',
                     right: '1'
-                }[label.alignValue as Highcharts.AlignValue];
+                }[label.alignValue as Highcharts.AlignValue || label.textAlign];
```
pawelfus, line 193: I might be wrong here, but isn't `align` and `textAlign` different options? At least that's the case in StackLabels: https://api.highcharts.com/highcharts/yAxis.stackLabels.textAlign

PaulDalek: It seems that in case when the `useHTML` is set to true, the `textAlign` property is created in the `alignSetter` for HTML elements (https://github.com/highcharts/highcharts/blob/vacation2020/js/Core/Renderer/HTML/HTML.js#L289) and it receives the value of the `align` option. I suppose it works that way to ensure the right position of a text for elements (e.g. https://api.highcharts.com/highcharts/title) that don't provide `textAlign` as an option that can be set through the API.

PR #13933, opened 2020-07-20
bugfix/13353-passive-events

### LINKED ISSUE #11353: Multi-[Violation] Added non-passive event listener to a scroll-blocking 'touchstart' event ++
### Situation
**On MacOS/Chrome**
_CandleStick: When the StockTools are enabled_
Lot of noticeable non proper handled eventListners thrown in the console. mainly refers to touchstart event
`[Violation] Added non-passive event listener to a scroll-blocking 'touchstart' event. Consider marking event handler as 'passive' to make the page more responsive.`

### To reproduce
This can be easily noticeable in the console of any react-based demo of StockChart with enabled stock-toos module.

Here is a quick live demos:
https://codesandbox.io/s/10yv629397

Video:
https://i.imgur.com/Z80XISw.mp4

Screenshot 2019-07-08 at 16 38 38

**Tested on**
MacOS (Mojave) Chrome (Version 75.0.3770.100) and other Mac machines and Chrome versions.

raf18seb:
Fixed #11353, added support for passive touch events.

### DIFF
--- a/ts/Core/Globals.ts
+++ b/ts/Core/Globals.ts
@@ -72,6 +72,7 @@ declare global {
         const symbolSizes: Dictionary<SizeObject>;
         const win: GlobalWindow;
         const seriesTypes: SeriesTypesDictionary;
+        const supportPassiveEvents: boolean;
         const svg: boolean;
         const version: string;
         let theme: (Options|undefined);
@@ -166,7 +167,23 @@ var glob = ( // @todo UMD variable named `window`, and glob named `win`
     hasBidiBug = (
         isFirefox &&
         parseInt(userAgent.split('Firefox/')[1], 10) < 4 // issue #38
-    );
+    ),
+    noop = function (): void {},
+    // Checks whether the browser supports passive events, (#11353).
+    checkPassiveEvents = function (): boolean {
+        let supportsPassive = false;
+
+        const opts = Object.defineProperty({}, 'passive', {
+            get: function (): void {
+                supportsPassive = true;
+            }
+        });
+
+        glob.addEventListener('testPassive', noop, opts);
+        glob.removeEventListener('testPassive', noop, opts);
+
+        return supportsPassive;
+    };
 
 var H: typeof Highcharts = {
     product: 'Highcharts',
@@ -185,10 +202,11 @@ var H: typeof Highcharts = {
     chartCount: 0,
     seriesTypes: {} as Highcharts.SeriesTypesDictionary,
     symbolSizes: {},
+    supportPassiveEvents: checkPassiveEvents(),
     svg: svg,
     win: glob,
     marginNames: ['plotTop', 'marginRight', 'marginBottom', 'plotLeft'],
-    noop: function (): void {},
+    noop: noop,
 
     /**
      * Theme options that should get applied to the chart. In module mode it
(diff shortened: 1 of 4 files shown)


ts/Core/Globals.ts
```diff
@@ -185,10 +202,11 @@ var H: typeof Highcharts = {
     chartCount: 0,
     seriesTypes: {} as Highcharts.SeriesTypesDictionary,
     symbolSizes: {},
+    supportPassiveEvents: checkPassiveEvents(),
```
sebastianbochan, line 205: 1. `isPassiveEvent` would be shorter.

raf18seb:
Thank for your comments. @[**oysteinmoseng**](https://github.com/oysteinmoseng) I was afraid of it. We are not able to detect whether users run `preventDefault` or not. Do you guys have any more specific suggestions for the next steps?

pawelfus:
How about we add a `passive` option to `options` config in `addEvent(element, type, callback, options)`? By default, it will be passive for touch events, unless set `options.passive = false`? That should resolve issue with our build-in event listeners - we could control them as wish, reducing amount of warnings in console.

Regarding user-defined callbacks - I don't see a perfect solution. The problem we have: we provide `callback` option in API, but not `options`. We can enable this in API, so users need to decide whether event will be passive or not:

```js
chart: {
  events: {
    click: callback, // assuming for a click event, we attach touchstart+touchend too, like in `SVGElement.on()`
    clickOptions: {
      passive: boolean
    }
  }
}
```

PS: Checked how frameworks resolved that - angular, react and vue still haven't adapted to this.

oysteinmoseng:
I like supporting `passive` in the `addEvent` options. Not too sure about making it default for touch events, but Torstein might have input here.

As for setting it for the user, maybe either making it part of a returned object in the event handler or setting a property on the event argument object could be easier than adding new options in the API.

Example:
```javascript
events: {
  click: function (e) {
    e.iAmDefinitelyPassive = true;
    console.log('I am passive');      
  }
}
```

raf18seb:
@oysteinmoseng are you sure we can go this way? Maybe I am misunderstanding something.
We can decide whether an event is passive only while adding an event, not later when firing the event's handler function.
Going this way, we could only (I don't see any other way) parse function to string (inside `addEvent` method), try to find something like `iAmDefinitelyPassive` in string, and then add an event with a proper `passive` flag, but we definitely don't want to go this way.

oysteinmoseng:
@raf18seb You are 100% right, disregard that suggestion. My opinion then would be that we should make all user-added events non-passive by default. We can then introduce one or more options for marking relevant events passive if needed. Maybe it is enough to have a `passiveTouchEvents` boolean option that covers all touch events? @TorsteinHonsi probably has input on this too.

TorsteinHonsi:
The most flexible solution I can come up with is similar to @pawelfus', but instead of providing an `events.clickOptions` property next to `events.click`, I think it would look better to support an object config:

```js
events: {
   click: {
      listener: function () {},
      options: {
         passive: true
      }
   }
}
```

On the downside, we will introduce complexity to the typing as `events.click` would become `ChartClickCallbackFunction|ChartClickCallbackObject`.

If the current suggestion works for our internal cases, I'd say we start with that. But obviously we need to check if we break something, by identifying if there are places we are currently calling `e.preventDefault` on touch events.

On another note, do we really need to check if passive mode is supported by the browser? Do we break anything if we pass in `{ passive: true }` and the browser doesn't support it?

raf18seb:
Thanks, will do.

> On another note, do we really need to check if passive mode is supported by the browser? Do we break anything if we pass in { passive: true } and the browser doesn't support it?

Yes, for example IE doesn't accept the 4th argument as an object but only a `boolean` that indicates `useCapture` parameter.
If we don't check if passive mode is supported by the browser, then IE will treat an object argument as `true` and break most of the functionalities (`capture` should be `false`).
```
if ({}) {
    console.log('This statement is true and I am logged in the console')
}
```

Issue #13937, opened 2020-07-20
The 'eventsToUnbind' array refers the same array between different series and charts.


Denyllon:
#### Expected behaviour
The `eventsToUnbind` array should be generated per series instance.

#### Actual behaviour
When there are multiple charts on the page, and any of them is updated, then the events within `series.eventsToUnbind` of all those charts are unbind.

#### Live demo with steps to reproduce
https://jsfiddle.net/BlackLabel/1kp9hoqx/
https://jsfiddle.net/BlackLabel/619uokyL/1/ - drilling down sunburst's nodes calls the series update on the second chart, what causes unbinding the series `click` handler on the first chart, and drilldown stops working.

#### Internal note
FYI @bre1470 
_The problem is caused by the https://github.com/highcharts/highcharts/commit/495fee94bd2a007f48720e78bfdba19f75270c80 commit, where the `eventsToUnbind` property is initialized on `Series.prototype` instead of e.g `init()`_

PR #13945, opened 2020-07-22
bugfix/12959-13937-sunburst-treemap-drilldown

### LINKED ISSUE #12959: Sunburst doesn't drilldown after re-rendered (with multiple charts on the same page)
#### Expected behaviour
We should be able to keep drilling down in all charts after we have a re-render

#### Actual behaviour
The sunburst chart doesn't "drilldown" after it's re-rendered with more than one chart in the page.
If only one sunburst is in isolation it works even after re-render.
After a re-render, only the last sunburst one keeps "working".

https://codesandbox.io/s/silent-surf-yxi5i

#### Product version
8.0.0

#### Affected browser(s)
All

I also opened the same issue on highcharts-react (don't know where is the "best" place)
https://github.com/highcharts/highcharts-react/issues/200

Denyllon:
Fixed #12959, #13937, drilldown did not work on treemap and sunburst series after update.

### DIFF
--- a/js/Core/Series/Series.js
+++ b/js/Core/Series/Series.js
@@ -2666,7 +2666,6 @@ null,
     colorCounter: 0,
     cropShoulder: 1,
     directTouch: false,
-    eventsToUnbind: [],
     isCartesian: true,
     // each point's x and y values are stored in this.xData and this.yData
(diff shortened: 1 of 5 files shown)


oysteinmoseng:
Good stuff

PR #13949, opened 2020-07-23
bugfix/12275-pointInterval-pointStart-updates-through-plotOptions

### LINKED ISSUE #12275: X axis is not getting updated on changing pointStart
#### Expected behaviour
BarChart: The start value on X axis must updated when `plotOptions.series.pointStart` was updated.

#### Actual behaviour
When dynamically updating `plotOptions.series.pointStart` and updating the chart with new data the X axis is not getting updated, it stays with the initial value.

#### Product version
Highcharts 7.1.2

#### Example
https://stackblitz.com/edit/react-jrd9mx

My options are like this:
```javascript
const options = {
    plotOptions: {
        series: {
            pointStart: firstXAxisValue, // receives new value on each update
            stacking: 'normal',
            pointPadding: 0.1,
            groupPadding: 0,
            shadow: false
        }
    },
    tooltip: {
        formatter: createTooltipFormatter(), // custom formatter
        backgroundColor: 'rgba(255, 255, 255, 0.95)',
        borderColor: '#444444',
        borderRadius: 3,
        style: {
            padding: 0,
            fontFamily
        },
        shared: true
    },
    xAxis: {
        labels: {
            step: 1,
            formatter() {
                if (this.isFirst || this.isLast) return this.value;
                return null;
            }
        },
        tickInterval: 1,
        title: {
            text: 'Years'
        },
        minorTicks: false,
        crosshair: false
    }
}
```

mateuszkornecki:
Fixed #12275, `pointInterval` and `pointStart` through `plotOptions` didn't work.


ts/Core/Dynamics.ts
```diff
@@ -968,6 +970,35 @@ extend(Chart.prototype, /** @lends Highcharts.Chart.prototype */ {
...
+     *
+     * @param {string} property
+     *
+     * @return {boolean}
+     */
+    isInPlotOptions: function (this: Chart, property: string): boolean {
+        const chart = this,
+            plotOptions = chart.options.plotOptions;
+        if (plotOptions) {
+            for (const seriesType in plotOptions) {
```
pawelfus, line 989: We already have `objectEach()` in utils. Could you reuse it?

ts/Core/Dynamics.ts
```diff
@@ -1485,6 +1516,8 @@ extend(Series.prototype, /** @lends Series.prototype */ {
                 (newType && newType !== this.type) ||
                 // New options affecting how the data points are built
                 typeof options.pointStart !== 'undefined' ||
+                chart.isInPlotOptions('pointInterval') ||
+                chart.isInPlotOptions('pointStart') ||
```
pawelfus, line 1520: It's not enough to check if option exists in `plotOptions`. We should keep points when value did not change. 

Also missing `pointIntervalUnit` and `keys` options too.

ts/Core/Dynamics.ts
```diff
@@ -1667,6 +1669,26 @@ extend(Series.prototype, /** @lends Series.prototype */ {
...
+     *
+     * @return {boolean}
+     */
+    hasOptionChanged(this: Highcharts.Series, optionName: string): boolean {
+        const chart = this.chart,
+            plotOptions = chart.options.plotOptions;
+
+        return (this.options as any)[optionName] !== pick(
+            (plotOptions as any)[this.type][optionName],
+            (plotOptions as any).series[optionName],
```
pawelfus, line 1689: This won't work, see: https://jsfiddle.net/BlackLabel/397zvwm5/2/

ts/Core/Dynamics.ts
```diff
@@ -1667,6 +1669,26 @@ extend(Series.prototype, /** @lends Series.prototype */ {
...
+     *
+     * @param {string} option
+     *
+     * @return {boolean}
+     */
+    hasOptionChanged(this: Highcharts.Series, optionName: string): boolean {
+        const chart = this.chart,
+            plotOptions = chart.options.plotOptions;
+
+        return (this.options as any)[optionName] !== pick(
```
pawelfus, line 1687: Simple `!==` works only for primitives. For objects (e.g. `dataGrouping`) won't work: https://jsfiddle.net/BlackLabel/397zvwm5/4/

pawelfus: Use case: https://jsfiddle.net/BlackLabel/397zvwm5/6/ - should it return false?

ts/Core/Dynamics.ts
```diff
@@ -1667,6 +1671,33 @@ extend(Series.prototype, /** @lends Series.prototype */ {
...
+     * Check if the option has changed.
+     *
+     * @private
+     * @function Highcharts.Series#hasOptionChanged
+     *
+     * @param {string} option
+     *
+     * @return {boolean}
+     */
+    hasOptionChanged(this: Highcharts.Series, optionName: string): boolean {
```
pawelfus, line 1685: This method works nicely for primitive properties (string, number etc), but won't work with arrays/objects. Comparing `{ x: 5 }` with `{ x: 5 }` won't work (`!==` compares references), unless we implement deep recursive comparison. That seems to be an overhead for `dataGrouping`/`keys` option only. Regarding `keys` see my disclaimer below, but `dataGrouping` is related to Stock only, not Highcharts. Perhaps Stock should overwrite this methods (or use events)?

Anyway, I see a few possible solutions here:
- ignore that updating `plotOptions.series.dataGrouping` doesn't work -> as already in current release, no changes then
- merge this branch, but this affects performance -> `{} !== {}` so points are recreated always when `dataGrouping` is set like this `series: [{ dataGrouping: { ... } }]`. Sole purpose of `dataGrouping` is to increase the performance so this solution should be avoided
- implement deep copy (in Highcharts or in Stock) -> resolves the issues, but produces extra code. And we end up with another recursive function to maintain (`merge()`, `cleanRecursive()`, `extend()` and something in tooltip and in networkgraph and who knows where else :))
- implement simple comparison of objects/arrays - limited to the first level only (instead of recursive solution) -> could be lightweight solution, with just a few lines of code

Disclaimer: I said "it won't work with arrays" but it's not true. It works, because our `merge()` does copy arrays (and nodes afaik) directly, not item by item (performance reasons) :)

What do you think @oysteinmoseng @sebastianbochan @bre1470 ? Any other ideas? Which one do you prefer?

pawelfus: Ping @oysteinmoseng @sebastianbochan @bre1470 and now also @TorsteinHonsi :)

oysteinmoseng: I would suggest writing some specific performance tests so we're not shooting in the dark here. It can be hard to predict what will be performance bottlenecks. If I understand the issue correctly the ideal solution would be that `hasOptionChanged` does a deep comparison between objects (and arrays?). This might not necessarily be a noticeable performance hit at all. We don't see deep nested objects and arrays passed to `series.update` normally.

Not a big fan of only solving this in Stock since we are probably eventually going to add things to plotOptions in Highcharts that requires deep comparisons too (if we don't have some already?).

Leaving it as is (maybe documenting it) is the pragmatic approach of course, until someone complains.

sebastianbochan: I would say that `implement simple comparison of objects/arrays` sounds like a good plan and an usefull tool also for other issues. 

Additionally, it can be useful in a bug like: exporting (compare userOptions and options for an axis).

mateuszkornecki: @TorsteinHonsi Do you have any thoughts, about choosing the right solution?

TorsteinHonsi: Actually, after trying different approaches, I like this one best. 

> this affects performance -> {} !== {} so points are recreated always when dataGrouping is set like this series: [{ dataGrouping: { ... } }]

I think it is okay that points are always recreate in this scenario. When `dataGrouping` is set in the updated options, we must assume it is because it should be changed.

TorsteinHonsi (approved):

TorsteinHonsi:
Is it ready to merge? It is still in Draft mode... @mateuszkornecki

mateuszkornecki:
Yes, you are right the PR is ready. I totally forgot that it is still a draft, sorry!

PR #13970, opened 2020-07-29
bugfix/10779-yAxis-max-issue

### LINKED ISSUE #10779: Gantt - yAxis max value
Gantt chart with yAxis.max value higher than point numbers is rendering wrong. 

[Workaround](https://jsfiddle.net/BlackLabel/0fuxb8e2/)

[Demo](https://jsfiddle.net/BlackLabel/jmcvx7ga/)

#### Product version

Highcharts JS v7.1.1 (2019-04-09)

#### Affected browser(s)
All

EDIT:
[The solution suggested by a user](http://jsfiddle.net/BlackLabel/q9ejdh4b/)

Basts:
Fixed #10779, yAxis max value occurred the wrong axis rendering.

### DIFF
--- a/ts/Core/Axis/Tick.ts
+++ b/ts/Core/Axis/Tick.ts
@@ -818,7 +818,6 @@ class Tick {
             axis = tick.axis,
             labelPos,
             reversed = axis.reversed,
-            isVertical = !axis.horiz,
             xPos,
             yPos;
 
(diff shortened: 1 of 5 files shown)


pawelfus:
Any problems with this solution and `addPoint`/`setData`/`update` ?

PS: Missing labels ;)

Basts:
@pawelfus 

Now - not! 😄  Demo for tests: http://jsfiddle.net/BlackLabel/1pfanvj7/

ts/Core/Axis/Tick.ts
```diff
@@ -848,10 +848,19 @@ class Tick {
...
         if (!moved && (tick.labelPos || label)) {
             labelPos = tick.labelPos || (label as any).xy;
+            // Use the inverted logic for the yAxis
+            if (!axis.isXAxis) {
+                if (inverted) {
+                    inverted = false;
+                } else {
+                    inverted = true;
+                }
+            }
```
pawelfus, line 858: I don't think it's a good idea. Perhaps instead of `chart.inverted`, we should use `axis.horiz`?

Basts:
@oysteinmoseng 

I have an idea for the simplest solution ever in this case - set name as `i + ' '`(space in the empty string) rather than `i + ''`. What do you think?

oysteinmoseng:
@Basts Haha, well why not. I say we go for it, at least until it turns out to be an issue for someone. I am sure it is an edge case anyway.

Basts:
@oysteinmoseng I think that it has an impact for the string alignment because in the Gantt labels start to align from the left side of the label text - demo: https://jsfiddle.net/BlackLabel/gwk7zbqo/

oysteinmoseng:
@Basts Try inserting something like this instead of a real space: http://www.fileformat.info/info/unicode/char/200B/index.htm

oysteinmoseng:
Awaiting review from @pawelfus as well.

PR #13974, opened 2020-07-30
bugfix/10555-scrollablePlotArea-absolute

### LINKED ISSUE #10555: scrollablePlotArea does not work properly on Firefox
#### Expected behaviour
scrollablePlotArea should work correctly on all browsers.

#### Actual behaviour
scrollablePlotArea doesn't work on Firefox.

view: https://jsfiddle.net/BlackLabel/ko92esqx/show
editor: https://jsfiddle.net/BlackLabel/ko92esqx/

#### Product version
Highcharts

#### Affected browser(s)
Firefox 64.0 (macOS Mojave 10.14.3)

goransle:
Fixed #10555, scrollablePlotArea not rendering correctly on FireFox.

No fix applied: https://jsfiddle.net/BlackLabel/ko92esqx/show
Fix applied: https://jsfiddle.net/L5zbtc9y/show

### DIFF
--- a/ts/Extensions/ScrollablePlotArea.ts
+++ b/ts/Extensions/ScrollablePlotArea.ts
@@ -42,6 +42,7 @@ declare global {
             fixedRenderer?: Renderer;
             innerContainer?: HTMLDOMElement;
             scrollingContainer?: HTMLDOMElement;
+            scrollingParent?: HTMLDOMElement;
             scrollableMask?: SVGElement;
             scrollablePixelsX?: number;
             scrollablePixelsY?: number;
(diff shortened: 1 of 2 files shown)


ts/Extensions/ScrollablePlotArea.ts
```diff
@@ -254,10 +254,22 @@ Chart.prototype.setUpScrolling = function (): void {
...
+    // Insert a container with position relative
+    // that scrolling and fixed container renders to (#10555)
+    const relContainer = createElement(
+        'div',
+        void 0,
+        {
+            position: 'relative'
+        },
+        this.renderTo,
+        true
```
oysteinmoseng, line 266: We probably don't need the `nopad` argument to be `true` for a div?

ts/Extensions/ScrollablePlotArea.ts
```diff
@@ -254,10 +254,22 @@ Chart.prototype.setUpScrolling = function (): void {
         attribs.overflowY = 'auto';
     }
 
+    // Insert a container with position relative
+    // that scrolling and fixed container renders to (#10555)
+    const relContainer = createElement(
+        'div',
+        void 0,
```
oysteinmoseng, line 261: Would consider adding a class name `highcharts-scrolling-parent` or similar just in case someone needs access to this container. Maybe also store the element on the chart object

ts/Extensions/ScrollablePlotArea.ts
```diff
@@ -355,14 +366,14 @@ Chart.prototype.applyFixed = function (): void {
...
+                top: 0
             },
             null as any,
             true
         );
-        this.renderTo.insertBefore(
-            this.fixedDiv,
-            this.renderTo.firstChild
+        this.scrollingContainer?.parentNode.appendChild(
+            this.fixedDiv
```
oysteinmoseng, line 376: This puts the fixed container after the scrolling container in the DOM, while it used to be before. Assuming this is not an issue?

goransle: Does not seem to be an issue, but I'll restore the order.

PR #13975, opened 2020-07-30
tools/update

bre1470:
Updated dependencies.


bre1470:
The file size comparison is huge. I wonder if this is the result of a change in the ZIP dependency.

PR #13976, opened 2020-07-30
bugfix/13971-color-axis-import

### LINKED ISSUE #13971: Download builder generates invalid bundles
A number of elements from `H`/`parts/Globals.js` are reported as `undefined` when loading files generated by the [Download Builder][1].

#### Steps to reproduce
1. Using the [Download Builder][1], create a bundle containing all available modules (with the expectation that this combination cannot result in missing dependencies).
2. Save that file to a directory and create a simple html file that loads it: `<html><head><script src="./highcharts-custom.js"></script></head></html>`
3. Open said HTML file in the browser of your choice and look at the JS console.

#### Expected behaviour
The file should load correctly and no errors should be logged.

#### Actual behaviour
The following error is reported while loading `ColorAxis.js`:

```
Uncaught TypeError: can't access property "prototype", Series is undefined
```

Looking at the site where the exception was thrown, this is because `H.Series` appears to be undefined while the module is loading.

#### Product version
* Highcharts 8.1.2
* Download builder from highcharts.com
* Possibly [the section of the doc on custom builds](https://www.highcharts.com/docs/getting-started/how-to-create-custom-highcharts-files) (untested)

#### Affected browser(s)
* Firefox 79
* Chrome 84

[1]:https://www.highcharts.com/blog/download/#downloadbuilder

bre1470:
Fixed #13971, missing Series import in ColorAxis.

### DIFF
--- a/js/Accessibility/Components/SeriesComponent/SeriesKeyboardNavigation.js
+++ b/js/Accessibility/Components/SeriesComponent/SeriesKeyboardNavigation.js
@@ -19,6 +19,8 @@ import KeyboardNavigationHandler from '../../KeyboardNavigationHandler.js';
 import EventProvider from '../../Utils/EventProvider.js';
 import ChartUtilities from '../../Utils/ChartUtilities.js';
 var getPointFromXY = ChartUtilities.getPointFromXY, getSeriesFromName = ChartUtilities.getSeriesFromName, scrollToPoint = ChartUtilities.scrollToPoint;
+import '../../../Series/ColumnSeries.js';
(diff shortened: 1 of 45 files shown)


sm-Fifteen:
Like I said in the bug report, ColorAxis is probably not the only place where this problem occurs, it just happens to be the first one in loading order. Removing Heatmap and all the Highmaps components causes similar errors to occur at different points in the loading procces.

bre1470:
@sm-Fifteen Thanks for pointing it out. I fixed all additional series dependencies.

Issue #13583, opened 2020-05-25
Fullscreen on iOS


hillerrv:
#### Expected behaviour
Full screen view opens.

#### Actual behaviour
Full screen view does not open.

#### Live demo with steps to reproduce
- Open https://www.highcharts.com/demo/line-basic in Safari on iOS 13
- Click burger menu -> 'view in full screen'

#### Product version
Highcharts v 8.1.0

#### Affected browser(s)
Safari and Chrome on iOS

sebastianbochan:
Hi @hillerrv,
Thank you for reporting. As I see in the scope of functionality, the Fullscreen API is not supported in iOS Safari and has partial support in Safari: [more on this](https://caniuse.com/#feat=fullscreen).

As I tested the fullscreen works on iPad, but not on iPhone.

hillerrv:
Hi @sebastianbochan 
I do see the problem that the official Fullscreen API is not supported on iPhone. However, using a decent library I would not expect the need to implement the workaround myself.

raf18seb:
Hi @hillerrv

As @sebastianbochan said, the scope of our fullscreen module functionality is limited to the browser [Full Screen API](https://caniuse.com/#feat=fullscreen) which is limited in some browsers.

We could do some research and extend it in the future, but it is not on our high priority features list right now. I agree with you that, for now, we could create our official support matrix and add it into our docs, but, in the end, it would be very similar to the one shown in the link above.

I am sorry for the inconvenience, but there is not much we can do with it at this moment.

urvesh86:
Hi @sebastianbochan & @raf18seb, 

we are facing the same issue in IOS device too. I went through whole thread. It is really helpful if you fix it asap.

Thank you!!

hillerrv:
Hi @raf18seb 
I agree that your documentation would be very similar to the link of caniuse, but as a developer I do not know that you implemented the highcharts fullscreen mode with the Full Screen API. Thus, it would be very useful if you would add a note in the documentation for other developers.

raf18seb:
@hillerrv good point, thanks for sharing your opinion.

urvesh86:
Hi @sebastianbochan & @sebastianbochan,

I have upgraded latest version of highchart (8.1.2), still facing the same issue. If you have any plan to solve this issue in near future, please share tentative date of when this feature will be available. It would helpful to us in planning.

sebastianbochan:
HI @urvesh86,
[...]

PR #14003, opened 2020-08-05
docs/13583-fullscreen

sebastianbochan:
#13583, docs: fullscreen, added information about limits.


ts/Stock/StockToolsBindings.ts
```diff
@@ -1994,6 +1994,8 @@ var stockToolsBindings: Highcharts.Dictionary<Highcharts.NavigationBindingsOptio
     /**
      * Displays chart in fullscreen.
      *
+     * Fullscreen is not supported on iPhone due to iOS limits.
```
pawelfus, line 1997: Could you use bold `**Note**:` at the beginning? To highlight this :)

sebastianbochan: Fixed

sebastianbochan:
Im not sure why test not passed, especially that is only docs edit. It looks like an issue in the branch.

sebastianbochan:
> Shouldn't it be "iOS limitations" rather than "iOS limits"? 🤔

Indeed, will commit.

sebastianbochan:
Done

Issue #13768, opened 2020-06-25
Gantt- unable to create the chart using the keys


karolkolodziej:
#### Expected behaviour
Create the chart by setting the `keys` and putting data into an array.

#### Actual behaviour
The chart is not creating.

#### Live demo with steps to reproduce
https://jsfiddle.net/BlackLabel/uz6Lf74y/1/

#### Product version
Highcharts Gantt JS v8.1.2 (2020-06-16)

#### Affected browser(s)
All

PR #14016, opened 2020-08-07
bugfix/13838-exporting-collapsed-points

### LINKED ISSUE #13838: Gantt - exporting collapsed points
#### Expected behaviour

Toggling the node by the user should change the `point.collapsed` property which will make visible nodes after for the exporting.  

#### Actual behaviour

Toggling the nodes doesn't have an impact on the exported chart (the chart is rendering for the export with the initial options). 

Demo: https://jsfiddle.net/BlackLabel/5ygt6pqv/

**Workaround:**
https://jsfiddle.net/BlackLabel/cy27xwz5/

#### Product version

 Highcharts Gantt JS v8.1.2 (2020-06-16)

Basts:
Fixed #13838, toggled nodes in the tree grid were not reflected correctly in the exported chart.

### DIFF
--- a/samples/unit-tests/gantt/treegrid/demo.js
+++ b/samples/unit-tests/gantt/treegrid/demo.js
@@ -439,4 +439,10 @@ QUnit.test('series.data[].collapsed', assert => {
         false,
         'This point should be expanded #13838'
     );
+
+    assert.strictEqual(
+        axis.series[0].options.data[0].collapsed,
+        false,
+        'This point should be expanded #13838'
+    );
 });
--- a/samples/unit-tests/gantt/treegrid/demo.js
+++ b/samples/unit-tests/gantt/treegrid/demo.js
@@ -433,4 +433,10 @@ QUnit.test('series.data[].collapsed', assert => {
         2,
         'should have axis.max equal 2 when "Node 1" is expanded.'
     );
+
+    assert.strictEqual(
+        axis.series[0].points[0].collapsed,
+        false,
+        'This point should be expanded #13838'
+    );
 });
--- a/ts/Core/Axis/TreeGridAxis.ts
+++ b/ts/Core/Axis/TreeGridAxis.ts
@@ -99,6 +99,7 @@ namespace TreeGridAxis {
         collapseStart?: number;
         depth: number;
         descendants?: number;
+        id?: number | string | null;
         height?: number;
         name: string;
         nodes: [TreeGridNode];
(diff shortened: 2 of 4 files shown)


ts/Core/Axis/TreeGridAxis.ts
```diff
@@ -851,6 +853,19 @@ namespace TreeGridAxis {
...
             breaks.push(obj);
+            // Change the collapsed flag #13838
+            node.collapsed = true;
+            axis.series.forEach(function (series): void {
+                const data = series.options.data;
+
+                if (data && data.length) {
+                    data.forEach(function (data): void {
+                        if ((data as PointOptionsObject).id === node.id) {
+                            (data as PointOptionsObject).collapsed = true;
```
pawelfus, line 864: Can Gantt a point be defined as an array instead of an object? If yes, then this won't work with arrays. For the reference, you can check how `PieSeries.slice()` is implemented (where `point.sliced` option is updated).

ts/Core/Axis/TreeGridAxis.ts
```diff
@@ -851,6 +853,19 @@ namespace TreeGridAxis {
                 obj = getBreakFromNode(node, axis.max);
 
             breaks.push(obj);
+            // Change the collapsed flag #13838
+            node.collapsed = true;
+            axis.series.forEach(function (series): void {
+                const data = series.options.data;
+
+                if (data && data.length) {
+                    data.forEach(function (data): void {
```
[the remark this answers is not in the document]
pawelfus: Perhaps `var point = chart.get('id')` will be enough?

samples/unit-tests/gantt/treegrid/demo.js
```diff
@@ -433,4 +433,10 @@ QUnit.test('series.data[].collapsed', assert => {
         2,
         'should have axis.max equal 2 when "Node 1" is expanded.'
     );
+
+    assert.strictEqual(
+        axis.series[0].points[0].collapsed,
```
pawelfus, line 438: Please check also `series.options.data[0].collapsed`

ts/Core/Axis/TreeGridAxis.ts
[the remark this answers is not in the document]
oysteinmoseng: Even better

ts/Core/Axis/TreeGridAxis.ts
oysteinmoseng: [I don't think it works right now](https://github.com/highcharts/highcharts/issues/13768), but it should.

pawelfus (approved):
I see nothing wrong :)

TorsteinHonsi (approved):

PR #14031, opened 2020-08-12
bugfix/14025-ie11-support-annotations-in-exportdata

### LINKED ISSUE #14025: Pre-production: Annotation data export fail in IE11
Consider this config:
```js

    series: [{
        type: 'pie',
        data: [2, {
            y: 4,
            id: 'pointI'
        }, 3, {
            y: 4,
            id: 'pointII'
        }, 6]
    }],

    annotations: [{
        labelOptions: {
            includeInDataExport: true //default
        },
        labels: [{
            point: 'pointI',
            text: 'This is my annotation I'
        }]
    }, {
        labelOptions: {
            includeInDataExport: false
        },
        labels: [{
            point: 'pointII',
            text: 'This is my annotation II'
        }]
    }],

    exporting: {
        showTable: true
    }
});
```

With the current master, this will fail in IE11 because `Array.prototype.fill` is not supported. The polyfill in `oldie-polyfills.js` is valid, but that file is only intended for IE6-8.

Probably the `.fill('')` call should just be refactored to a simple loop filling the empty slots.

madepiet:
Fixed #14025, changing .fill and annotation logic.

### DIFF
--- a/ts/Extensions/Annotations/Annotations.ts
+++ b/ts/Extensions/Annotations/Annotations.ts
@@ -1796,10 +1796,20 @@ chartProto.callbacks.push(function (
 
                             // Annotation not connected to any xAxis -
                             // add new row.
-                            if (xAxisIndex === -1) {
-                                const newRow: any = new Array(event.dataRows[0].length);
+                            let isPie, i;
+                            const series = chart.series;
 
-                                newRow.fill('');
+                            for (i = 0; i < series.length; i++) {
+                                isPie = series[i].is('pie') === false;
+                            }
+
+                            if (isPie && xAxisIndex === -1) {
+                                const n = event.dataRows[0].length,
+                                    newRow: any = new Array(n);
+
+                                for (let i = 0; i < n; ++i) {
+                                    newRow[i] = '';
+                                }
                                 newRow.push(annotationText);
                                 newRow.xValues = [];
                                 newRow.xValues[xAxisIndex] = annotationX;
@@ -1834,9 +1844,12 @@ chartProto.callbacks.push(function (
                             // Annotation not placed on any exported data point,
                             // but connected to the xAxis - add new row
                             if (!wasAdded) {
-                                const newRow: any = new Array(event.dataRows[0].length);
+                                const n = event.dataRows[0].length,
+                                    newRow: any = new Array(n);
 
-                                newRow.fill('');
+                                for (let i = 0; i < n; ++i) {
+                                    newRow[i] = '';
+                                }
                                 newRow[0] = annotationX;
                                 newRow.push(annotationText);
                                 newRow.xValues = [];
--- a/ts/Extensions/Annotations/Annotations.ts
+++ b/ts/Extensions/Annotations/Annotations.ts
@@ -1796,14 +1796,7 @@ chartProto.callbacks.push(function (
 
                             // Annotation not connected to any xAxis -
                             // add new row.
-                            let isNotPie, i;
-                            const series = chart.series;
-
-                            for (i = 0; i < series.length; i++) {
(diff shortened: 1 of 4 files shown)


ts/Extensions/Annotations/Annotations.ts
```diff
@@ -1796,10 +1796,20 @@ chartProto.callbacks.push(function (
...
                             // Annotation not connected to any xAxis -
                             // add new row.
-                            if (xAxisIndex === -1) {
-                                const newRow: any = new Array(event.dataRows[0].length);
+                            let isPie, i;
+                            const series = chart.series;
 
-                                newRow.fill('');
+                            for (i = 0; i < series.length; i++) {
+                                isPie = series[i].is('pie') === false;
```
pawelfus, line 1803: What if we have mixed types? For example:
```js
series: [{ type: 'column' }, { type: 'pie' }, { type: 'column' }]
```

And actually it's not `isPie` but probably `isNotPie`? :)

ts/Extensions/Annotations/Annotations.ts
```diff
@@ -1796,10 +1796,20 @@ chartProto.callbacks.push(function (
...
+                            }
+
+                            if (isPie && xAxisIndex === -1) {
+                                const n = event.dataRows[0].length,
+                                    newRow: any = new Array(n);
```
pawelfus, line 1808: Can type be `string[]` instead of any?

ts/Extensions/Annotations/Annotations.ts
madepiet: This is a very good question that I also came across while creating this table and has not been discussed before. I think a little redesign of this table would be necessary here, as it may turn out to be less readable (this is just my opinion). It would be great to hear the opinions of others.

isNotPie, agree 👍

ts/Extensions/Annotations/Annotations.ts
madepiet: It cannot be used there because not all slots of the created table will be just a type string.

ts/Extensions/Annotations/Annotations.ts
pawelfus: What I mean is your `for () { ... }` is always setting `isPie` flag only for the last series :) It's no different from `const isPie = series[series.length - 1].is('pie') === false;`. The last item in `for()` has the highest priority.

Anyway, can we focus in this PR in supporting IE11? If there's something else we should discuss or rethink, then could you create separate issue? Thanks! :)

ts/Extensions/Annotations/Annotations.ts
pawelfus: Right, just checked: `getDataRows()` returns `Array<Array<(number|string)>>`. Still, it's not any :)

ts/Extensions/Annotations/Annotations.ts
madepiet: Right, I'll create a new ticket for this and improve this logic there.

madepiet:
@TorsteinHonsi yes, I think it can be added at the moment.  
I will fix the logic for the export annotation for the pie series with the new ticket. There we will also need to determine what it should look like because the table itself without annotations for mixed series types and pie looks like this: https://jsfiddle.net/BlackLabel/6omwnhfd/

I wonder what it should look like with the annotations added 🤔
@pawelfus what do you think?

ts/Extensions/Annotations/Annotations.ts
madepiet: Here it will be necessary to rebuild this logic, I will improve it in the new ticket.

pawelfus: Okay, thanks!

PR #14032, opened 2020-08-12
data-layer/events

bre1470:
Improved data events (making it type safe).


bre1470:
I hope so, at least that was the idea. So type sense works similar to the DOM events.

PR #14045, opened 2020-08-14
Refactor: TextBuilder

TorsteinHonsi:
A designated text builder that makes it easier to filter incoming HTML-like options. It breaks down the incoming HTML to an AST, filters it by tag and attribute whitelists, and inserts node by node into the DOM.


TorsteinHonsi:
I accidentally merged this into the master, which was not the intension yet. Have rolled back master and this branch. When ready, merge the branches directly without the PR.

KacperMadej:
@jasonclg 
Not through the options as this was never officially a part of API to support HTML - only text is supported.
In other words, secure support for `useHTML` for this element is not implemented.
Please open a new issue (a [feature request](https://github.com/highcharts/highcharts/issues/new?assignees=&labels=Type%3A+Feature+Request&template=feature_request.md&title=)) for this problem if you would like for this option to be fully supported.

However, you could use CSS a prepend or append content - https://jsfiddle.net/BlackLabel/jpc4gard/2/
(The above demo is base on an old forum post: https://www.highcharts.com/forum/viewtopic.php?t=17735)

jasonclg:
Thanks for the explanation!
Just reported this as a bug https://github.com/highcharts/highcharts/issues/15449
Looks like I need to open a new feature request?

KacperMadej:
@jasonclg The new issue is fine - thank you 👍

PR #14049, opened 2020-08-14
bugfix/13950-stocktools-dg-approximation

### LINKED ISSUE #13950: SMA volume in stock tools is not visible in high price ranges
#### Expected behaviour
The stock tools demo allows adding the SMA indicator to AAPL line chart. Alternative to price can volume be referenced. The volume thought is much lower than the price and in the default price view not visible.

#### Actual behaviour
SMA volume should get its own (hidden axis), so that the volume is visible in every view.

#### Product version
Highstock 8.1.2

#### Affected browser(s)
all

pawelfus:
Fixed #13950, in StockTools, adding technical indicators to column-type series did not use the same `dataGrouping.approximation`.
___
Alternative approach is to add this logic in Indicators core (instead of StockTools, like in this PR), where indicator will decide its approximation by overwriting `getDGApproximation()`. But that would be inconsistent (DG approx would depend on parent series type) and might be misleading for developers ("I'm adding SMA to A-type and B-type series, why lines are different?!"). Right now it's up to the developer to decide DG approx when using indicator. We need this change in StockTools because end-user has no way to choose an approximation, so we must set approx to the best one we can guess.

### DIFF
--- a/ts/Stock/StockToolsBindings.ts
+++ b/ts/Stock/StockToolsBindings.ts
@@ -295,7 +295,7 @@ bindingsUtils.manageIndicators = function (
         // Make sure that indicator uses the SUM approx if SUM approx is used
         // by parent series (#13950).
         if (
-            parentSeries &&
+            typeof parentSeries !== 'undefined' &&
             parentSeries instanceof Highcharts.Series &&
             parentSeries.getDGApproximation() === 'sum' &&
             // If indicator has defined approx type, use it (e.g. "ranges")
--- a/ts/Stock/StockToolsBindings.ts
+++ b/ts/Stock/StockToolsBindings.ts
@@ -19,11 +19,13 @@ import type Point from '../Core/Series/Point';
 import H from '../Core/Globals.js';
 import NavigationBindings from '../Extensions/Annotations/NavigationBindings.js';
 import U from '../Core/Utilities.js';
+import dataGrouping from '../Extensions/DataGrouping';
 const {
     correctFloat,
     defined,
     extend,
     fireEvent,
+    getOptions,
     isNumber,
     merge,
     pick,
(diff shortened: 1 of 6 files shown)


ts/Stock/StockToolsBindings.ts
```diff
@@ -19,11 +19,13 @@ import type Point from '../Core/Series/Point';
 import H from '../Core/Globals.js';
 import NavigationBindings from '../Extensions/Annotations/NavigationBindings.js';
 import U from '../Core/Utilities.js';
+import dataGrouping from '../Extensions/DataGrouping';
```
bre1470, line 22: This needs to have the file extension, or otherwise it will fail as an ES6 module.

ts/Stock/StockToolsBindings.ts
```diff
@@ -286,6 +290,25 @@ bindingsUtils.manageIndicators = function (
...
         seriesConfig.id = uniqueKey();
         navigation.fieldsToOptions(data.fields, seriesConfig);
+        parentSeries = chart.get(seriesConfig.linkedTo);
+        defaultOptions = getOptions().plotOptions as any;
+
+        // Make sure that indicator uses the SUM approx if SUM approx is used
+        // by parent series (#13950).
+        if (
+            parentSeries &&
+            parentSeries instanceof Highcharts.Series &&
```
bre1470, line 300: For optimization consider testing with `typeof parentSeries !== 'undefined'`, as it does not go up the `super` constructor.

pawelfus: Okay!

ts/Stock/StockToolsBindings.ts
pawelfus: I'm surprised it's even here. Already made a commit that removes that line, probably IDE auto-imported this during development.. thanks!

pawelfus: Done 👍

ts/Stock/StockToolsBindings.ts
pawelfus: Done 👍

PR #14072, opened 2020-08-18
Move Sonification to Extensions directory

oysteinmoseng:
Internally reorganized source file structure and naming.

#### Upgrade note
The file structure of ES modules has changed. This has no consequences when using master files as advised (like [http://code.highcharts.com/es-modules/masters/highcharts.src.js](http://code.highcharts.com/es-modules/masters/highcharts.src.js)), but may break when loading module files directly.


TorsteinHonsi (approved):
Thanks!

Issue #3965, opened 2015-03-10
tickAmount - serie is not centered


sebastianbochan:
Let's compare:
- http://jsfiddle.net/rb9p6b6t/1/
- http://jsfiddle.net/rb9p6b6t/2/

When we set tickAmount I think that serie should be in the middle, like in the first example.

jflambert:
This issue is affecting me. As a workaround I wrote code that detects min==max values and forces min-- and max++ for the xAxis ticks.

pawelfus:
Another workaround is to set `minRange` option instead, for example: http://jsfiddle.net/rb9p6b6t/3/

PR #14077, opened 2020-08-18
bugfix/13749-negative-ticks-on-positive-axes

### LINKED ISSUE #13749: Issue with negative ticks caused by using tickAmount and min together
#### Expected behaviour
When no negative values exist in data set, the expectation is that the x and y axes must start at 0 (min: 0 is given) and must plot only the positive ticks (tickAmount: 5 is given). Such as shown below using Highcharts version 5:
Correct-version5

#### Actual behaviour
Highcharts upgrade from version 5 to version 7 is causing negative ticks on x and y axes even when no negative data exists. Various chart plots are wrongly displaying negative axes. This worked perfectly with version 5 but is behaving weird with version 7 as shown below:
Issue

Look at the starting tick, -2.5, whereas there are no negative values, so it must start at 0. Also observe the last tick is not appropriate width as the other tick intervals and ends abruptly in between.

Here is the link for demonstration of the issue: https://jsfiddle.net/3awvyn0b/3/, see config for x-axis which is not being reflected in the plot:
xAxis: {
        tickAmount: 5,
        min: 0
    },
If I remove tickAmount for the x-axis, the negative ticks are resolved but then the ticks are not correctly marked. See here:
Capture1

#### Product version
Highcharts 7

#### Affected browser(s)
Chrome

mateuszkornecki:
Fixed #13749, negative ticks on positive axes with the `xAxis.min` = 0.

### DIFF
--- a/ts/Core/Axis/Axis.ts
+++ b/ts/Core/Axis/Axis.ts
@@ -5890,6 +5890,18 @@ class Axis implements AxisComposition, AxisLike {
 
         this.tickAmount = (tickAmount as any);
     }
+    /**
+     * Check if the series assigned to the axis have equal values
+     *
+     * @private
+     * @function Highcharts.Axis#haveSeriesEqualValues
+     * @return {boolean}
+     */
+    public haveSeriesEqualValues(): boolean {
+        const series = this.series,
+            seriesWithEqualValues = series.filter((series): boolean => series.dataMin === series.dataMax);
+        return seriesWithEqualValues.length === series.length;
+    }
 
     /**
      * When using multiple axes, adjust the number of ticks to match the highest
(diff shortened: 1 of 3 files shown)


ts/Core/Axis/Axis.ts
```diff
@@ -5899,27 +5911,29 @@ class Axis implements AxisComposition, AxisLike {
...
+            threshold = pick(axis.threshold, axis.softThreshold ? 0 : null);
+
+        let min, len, i;
 
         if (axis.hasData()) {
             if (currentTickAmount < tickAmount) {
                 min = axis.min;
 
                 while (tickPositions.length < tickAmount) {
+                    const isMinHigherThanTickInterval = (min || min === 0) && (min < tickInterval);
```
pawelfus, line 5931: Probably should be `Lower`, not `Higher`

mateuszkornecki: Indeed setting the `x Axis.softThreshold` will resolve this issue, unfortunately, it is causing more than 100 visual changes, and breaking a few tests.

mateuszkornecki:
> Just an idea: It works fine for yAxis, does not work for xAxis. It seems that yAxis gets its threshold (or softThreshold) from series. xAxis by default has both props disabled (or nullified), but if we set them, everything seems to be fine: https://jsfiddle.net/BlackLabel/3pu6axed/2/
> 
> Can we explore this solution? What do you think @KacperMadej @mateuszkornecki ?

mateuszkornecki:
> Just an idea: It works fine for yAxis, does not work for xAxis. It seems that yAxis gets its threshold (or softThreshold) from series. xAxis by default has both props disabled (or nullified), but if we set them, everything seems to be fine: https://jsfiddle.net/BlackLabel/3pu6axed/2/
> 
> Can we explore this solution? What do you think @KacperMadej @mateuszkornecki ?

Indeed enabling the `softThreshold` for `xAxis` will resolve this issue, unfortunately, it is causing more than 100 visual changes, and breaking a few tests.

mateuszkornecki:
One visual test is currently failing. I discussed that edge-case with @pawelfus and it seems that both results are correct. I can prevent changing axes with parallel coordinates by adding a simple condition, but I am not sure if it's really necessary.

TorsteinHonsi:
I think we need to reconsider how to fix this - the original fix for #3965 also seems to work only by accident. If the data is all negative, the `softThreshold` is not respected. Here's what should happen:
* If `min` is set, we should not prepend ticks below `min`. Likewise for `max`.
* If `softThreshold`, and the data is all-positive, we should not prepend ticks below `min`. Equivalent for all-negative data.

I'm creating a new PR with this approach.

PR #14126, opened 2020-08-25
ts/series-base

bre1470:
- Introduced base series class for series migration.
- Introduced line series for series migration.
- Moved old series class to catesian series.


bre1470:
At the moment CatesianSeries is identical with LineSeries. In the next migration step I will break this up, so that a real Series => CatesianSeries => LineSeries class inheritance gets established.

TorsteinHonsi (changes requested):
All good except
- CatesianSeries should be CartesianSeries
- Data labels should extend CartesianSeries
- Resolve conflicts

TorsteinHonsi (approved):

PR #14131, opened 2020-08-26
enhancement/offlineexporting-es6

marvin19:
Migrated OfflineExporting to ES6 pattern.


TorsteinHonsi (approved):
Looks good, pending @bre1470 's review.

PR #14141, opened 2020-08-28
bugfix/10365-gantt-tooltip-wrong-timezone

### LINKED ISSUE #10365: Timezone is not applied in Gantt's tooltip
#### Expected behaviour
The tooltip should display:
```
Start: 01:00
End: 02:00
```

#### Actual behaviour
The timezone is correctly applied to the point's graphic but it doesn't work in the toolitp.
It shows:
```
Start: 00:00
End: 01:00
```

#### Live demo
http://jsfiddle.net/BlackLabel/myr0Lxke/

#### Product version
7.0.3

karolkolodziej:
Fixed #10365, tooltip in Gantt did not respect time zones.


TorsteinHonsi (approved):
Thanks!

PR #14155, opened 2020-09-01
bugfix/13591-grid-axis-label-format

### LINKED ISSUE #13591: Timezone Problems in Gantt
Timezone "Europe/Berlin" is set in Gantt chart.

#### Expected behaviour

1. Weekday indication shall be correct in all zoom levels
2. Conversion of time shall be correct

#### Actual behaviour

1. The weekday indication depends on the zoom level of the Gantt. In the reproduction scenario, look at week 23. If zoomed-out, week 23 has the weekdays SMTWTFS (incorrect for timezone Europe/Berlin), if zoomed-in, the week has the weekdays "Mon, 1. Jun", "Tue, 2. Jun", ..., "Sun, 7. Jun" (correct for timezone Europe/Berlin).

2. Consider the task with the id "SU-INTE-Input" in the scenario, which has the start-timestamp 1591048800000. The Gantt converts this to "Monday, 1 Jun", which is wrong for timezone Europe/Berlin. Verify here - https://www.andre-jochim.de/unix_timestamp_umrechnen.htm - that the timestamp 1591048800 maps to 2020-06-02 00:00:00 for timezone Europe/Berlin.

http://jsfiddle.net/yf1Letjn/1/

#### Product version

Highcharts JS v8.1.0 (2020-05-05)
Highcharts Gantt JS v8.1.0 (2020-05-05)

//! moment.js
//! version : 2.26.0

//! moment-timezone.js
//! version : 0.5.31

#### Affected browser(s)

Reproduced on Firefox

karolkolodziej:
Fixed #13591, time zone declared per chart was not respected for calculating axis labels.

### DIFF
--- a/ts/Core/Axis/GridAxis.ts
+++ b/ts/Core/Axis/GridAxis.ts
@@ -273,7 +273,8 @@ H.dateFormats.W = function (this: Highcharts.Time, timestamp: number): string {
 
 // First letter of the day of the week, e.g. 'M' for 'Monday'.
 H.dateFormats.E = function (timestamp: number): string {
-    return dateFormat('%a', timestamp, true).charAt(0);
+    const time = new Highcharts.Time(this.options as any);
+    return time.dateFormat('%a', timestamp, true).charAt(0);
 };
 
 /* eslint-disable no-invalid-this */
--- a/ts/Core/Axis/GridAxis.ts
+++ b/ts/Core/Axis/GridAxis.ts
@@ -272,9 +272,8 @@ H.dateFormats.W = function (this: Highcharts.Time, timestamp: number): string {
 };
 
 // First letter of the day of the week, e.g. 'M' for 'Monday'.
-H.dateFormats.E = function (timestamp: number): string {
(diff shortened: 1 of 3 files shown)


ts/Core/Axis/GridAxis.ts
```diff
@@ -273,7 +273,8 @@ H.dateFormats.W = function (this: Highcharts.Time, timestamp: number): string {
 
 // First letter of the day of the week, e.g. 'M' for 'Monday'.
 H.dateFormats.E = function (timestamp: number): string {
-    return dateFormat('%a', timestamp, true).charAt(0);
+    const time = new Highcharts.Time(this.options as any);
+    return time.dateFormat('%a', timestamp, true).charAt(0);
```
pawelfus, line 277: Why can't we use `this.dateFormat()` but we need to create another `Time` instance? Is that a problem with TS or something else? :)

TorsteinHonsi (changes requested):
Pending change to `return this.dateFormat`

TorsteinHonsi (approved):
Thanks!

PR #14159, opened 2020-09-02
docs/14095-area-color-options

### LINKED ISSUE #14095: v:8.1.0:area color  wrong when set color with transparency
#### Expected behaviour

color can set myself transparency

#### Actual behaviour
when i set color:'rgba(237,235,255,0.3)'  but it's render color rgba(127,79,255,0.75)

Transparency fixed write dead

**Transparency is 0.3**:
image
**Transparency is 0.75**:
image

live demo:https://codesandbox.io/s/boring-moon-3wlig?file=/src/components/HelloWorld.vue

bre1470:
Fixed #14095, Added links between relevant color options in area series.

### DIFF
--- a/js/Series/AreaRangeSeries.js
+++ b/js/Series/AreaRangeSeries.js
@@ -54,6 +54,8 @@ BaseSeries.seriesType('arearange', 'area', {
      * @see [color](#plotOptions.arearange.color)
      * @see [fillColor](#plotOptions.arearange.fillColor)
      *
+     * @default   {highcharts} 0.75
+     * @default   {highstock} 0.75
      * @apioption plotOptions.arearange.fillOpacity
      */
     /**
@@ -643,6 +645,8 @@ BaseSeries.seriesType('arearange', 'area', {
  * @see [color](#series.arearange.color)
  * @see [fillColor](#series.arearange.fillColor)
  *
+ * @default   {highcharts} 0.75
+ * @default   {highstock} 0.75
  * @apioption series.arearange.fillOpacity
  */
 /**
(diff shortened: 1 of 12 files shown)


pawelfus (commented):
Just a comment, because I'm not sure how it's related to #14095 😅 

To be on the same page:
- `color` changes all: markers, line and fill, without `fillOpacity`
- `fillColor` changes only fill and fillOpacity (if provided as `rgba()`)
- `fillOpacity` changes only, well, fill's opacity :) 

Demo: https://jsfiddle.net/BlackLabel/k83x75rj/

And regarding #14095, it seems I wasn't clear enough, sorry. [API](https://api.highcharts.com/highcharts/series.area.fillOpacity) says it's by default set to `undefined`, but actually it's `0.75` (and doclet seems to be fine?)

bre1470:
Thanks for clarification. I will check the issue with the default value.

Just to make sure with the rest. I see the following: color changes markers color and opacity, line color and opacity, and fill color without opacity. Is that correct?

pawelfus:
> Thanks for clarification. I will check the issue with the default value.

Thanks!

> Just to make sure with the rest. I see the following: color changes markers color and opacity, line color and opacity, and fill color without opacity. Is that correct?

Exactly. When you put this in that way it may look wrong, but this happens only because `fillOpacity` has default value (`0.75`). There are no counterparts to `fillOpacity` for marker and line (like `series.marker.opacity` or `series.lineOpacity`).

pawelfus (commented):
Looks good thanks!

Just a side question: I see you defined this for `area` and some others (`arearange`, `areaspline` etc) separately. Does it mean we should also define this for `streamgraph` and `bellcurve`?

bre1470:
Default values get not inherited, so your mentioned series types need similar doclets.

TorsteinHonsi (approved):
Thanks!

PR #14162, opened 2020-09-02
Extracted AST functionality to new class.

oysteinmoseng:
@TorsteinHonsi See if you agree with this setup, or if we want to make some changes - definitely many ways to structure this. Will add to a11y once it's merged.

The PR adds a new `AST` class. The class just has a constructor and an `addToDOM` function, and a public `nodes` member that holds a reference to an array of nodes that make up the tree (renamed `ASTObject` to `ASTNode` to avoid confusion between these objects and the new class). To modify the tree before adding to the DOM, we would in other words just access the public `nodes` property.

It could be cleaner at times to have `ASTNode/ASTObject` keep all of this functionality itself, e.g. `ASTNode.addToDOM()` and `new ASTNode(markup)`, as well as `astNode.attributes[key] = val`. There seems to be some code where root properties on elements are copied over to the nodes for backwards compatibility, though, so I figured it was best avoided.

### DIFF
--- a/js/Core/Renderer/HTML/AST.js
+++ b/js/Core/Renderer/HTML/AST.js
@@ -1,0 +1,214 @@
+/* *
+ *
+ *  (c) 2010-2020 Torstein Honsi
+ *
+ *  License: www.highcharts.com/license
+ *
+ *  !!!!!!! SOURCE GETS TRANSPILED BY TYPESCRIPT. EDIT TS FILE ONLY. !!!!!!!
+ *
+ * */
+'use strict';
+import H from '../../Globals.js';
+import U from '../../Utilities.js';
+var attr = U.attr, objectEach = U.objectEach, splat = U.splat;
+/**
+ * Represents an AST
+ * @private
+ * @class
+ * @name Highcharts.AST
+ */
+var AST = /** @class */ (function () {
+    // Construct an AST from HTML markup, or wrap an array of existing AST nodes
+    function AST(source) {
+        this.nodes = typeof source === 'string' ?
+            this.parseMarkup(source) : source;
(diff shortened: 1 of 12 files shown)


TorsteinHonsi:
Thanks @oysteinmoseng !

> It could be cleaner at times to have ASTNode/ASTObject keep all of this functionality itself, e.g. ASTNode.addToDOM() and new ASTNode(markup), as well as astNode.attributes[key] = val. 

Yes, it could be cleaner at times. Much like native `DOMElement`s are nodes and trees at the same time, and they don't have wrapper objects.

> There seems to be some code where root properties on elements are copied over to the nodes for backwards compatibility, though, so I figured it was best avoided.

Can you show me the examples of that?

oysteinmoseng:
@TorsteinHonsi Here - unless I misunderstood: https://github.com/highcharts/highcharts/pull/14162/files#diff-ba6ff8eee12713d480647179346bb689R164-R175

TorsteinHonsi:
@oysteinmoseng I don't think that's a problem. What it does is to copy properties from a node to its attributes. It's for preserving legacy from the current `H.definition` function.

```js
node.someAttributeName => node.attributes.someAttributeName;
```

oysteinmoseng:
@TorsteinHonsi Maybe I am misunderstanding this, but doesn't that mean that the code would copy over any and all internal properties that we put on the `ASTNode` object? And add them as attributes in the DOM using `attr`? Seems potentially messy - but maybe we could filter against allowed attributes or something? Just thinking about possible future bugs if we add `this.referenceToX = x`, or if code (ours or clients) using `ASTNode` does something similar.

PR #14184, opened 2020-09-07
docs/14094-axis-oposite-gantt

### LINKED ISSUE #14094: Gantt/Docs- xAxis.opposite docs are incorrect
#### Expected behaviour
Change the default property of the `xAxis.opposite` in Gantt documentation.

#### Actual behaviour
The xAxis.opposite property is by default set in Gantt as `true`.
It is not consistent with the documentation.

API: https://api.highcharts.com/gantt/xAxis.opposite
https://jsfiddle.net/BlackLabel/m8a1dyx0/

#### Product version
Highcharts Gantt JS v8.1.2 (2020-06-16)

#### Affected browser(s)
All

karolkolodziej:
Fixed #14094, changed docs for the `opposite` property on `xAxis` in Gantt.

### DIFF
--- a/js/Core/Axis/Axis.js
+++ b/js/Core/Axis/Axis.js
@@ -4310,7 +4310,10 @@ var Axis = /** @class */ (function () {
          *         Y axis on left side
          *
          * @type      {boolean}
-         * @default   false
+         * @default   {highcharts} false
+         * @default   {highstock} false
+         * @default   {highmaps} false
+         * @default   {gantt} true
          * @apioption xAxis.opposite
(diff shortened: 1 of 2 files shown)


KacperMadej (commented):
@karolkolodziej 
Defaults could be grouped like in here:
https://github.com/highcharts/highcharts/blob/521442070932641983efe1998c40eea216ddda61/js/Core/Axis/Axis.js#L4085

TorsteinHonsi (approved):
Thanks!

PR #14185, opened 2020-09-07
Fixed #13749, ticks exceeded axis extremes when tickAmount was set

### LINKED ISSUE #13749: Issue with negative ticks caused by using tickAmount and min together

TorsteinHonsi:
Fixed #13749, when `tickAmount` was set, the tick positions would sometimes violate settings of `min`, `max` or `softThreshold`.

### To do
 - [ ] Find out what's up with the visual difference in the solid gauge demo

Replaces PR #14077


ts/Core/Axis/Axis.ts
```diff
@@ -5891,6 +5891,13 @@ class Axis implements AxisComposition, AxisLike {
         this.tickAmount = (tickAmount as any);
     }
 
+    /**
+     * When using multiple axes, adjust the number of ticks to match the highest
+     * number of ticks in that group.
+     *
+     * @private
+     * @function Highcharts.Axis#adjustTickAmount
+     */
```
pawelfus, line 5900: Unnecessary (duplicated) doclet

pawelfus (changes requested):
Just a minor thing (doclet).

Solution is not perfect: `min`+`max`+`tickAmount` still fails for xAxis: https://jsfiddle.net/BlackLabel/3pu6axed/5/ but I think it's good for now - the config in my demo is rather uncommon.

ts/Core/Axis/Axis.ts
TorsteinHonsi: Thanks!

PR #14221, opened 2020-09-14
bugfix/13884-scrollablePlotArea-collaps-nodes

### LINKED ISSUE #13884: Gantt : Collapsable nodes + navigator +  chart.scrollablePlotArea
#### Expected behaviour
I want to fix header position (with chart.scrollablePlotArea), 
using navigator option and collapsable nodes. 

#### Actual behaviour
When i add scrollableplotarea to my charts, i have several bugs : 
 - navigator is mostly not displayed, and i have a white bloc in the page bottom
 - when i close nodes (all nodes for example), all the chart disappear 
 - opactiy of the header is OK on the first display, then when we close nodes, opacity becomes to 0

template: https://jsfiddle.net/yfojxqb4/4/ 

#### Product version
Highcharts.ganttChart 8.1.2

#### Affected browser(s)
I am using Chrome latest version.

karolkolodziej:
Fixed #13884, collapsing node with `scrollablePlotArea` crashed the chart.

### DIFF
--- a/ts/Extensions/StaticScale.ts
+++ b/ts/Extensions/StaticScale.ts
@@ -98,12 +98,11 @@ Chart.prototype.adjustHeight = function (): void {
 
                 diff = height - chart.plotHeight;
 
-                if (!chart.scrollablePixelsY) {
-                    if (Math.abs(diff) >= 1) {
-                        chart.plotHeight = height;
-                        chart.redrawTrigger = 'adjustHeight';
-                        chart.setSize(void 0, chart.chartHeight + diff, animate);
-                    }
+                // Adjust the size, but not if scrollable plot area (#13884)
+                if (!chart.scrollablePixelsY && Math.abs(diff) >= 1) {
+                    chart.plotHeight = height;
+                    chart.redrawTrigger = 'adjustHeight';
+                    chart.setSize(void 0, chart.chartHeight + diff, animate);
                 }
 
                 // Make sure clip rects have the right height before initial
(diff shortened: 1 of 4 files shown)


ts/Extensions/StaticScale.ts
```diff
@@ -98,10 +98,12 @@ Chart.prototype.adjustHeight = function (): void {
 
                 diff = height - chart.plotHeight;
 
-                if (Math.abs(diff) >= 1) {
-                    chart.plotHeight = height;
-                    chart.redrawTrigger = 'adjustHeight';
-                    chart.setSize(void 0, chart.chartHeight + diff, animate);
+                if (!chart.scrollablePixelsX && !chart.scrollablePixelsY) {
```
TorsteinHonsi, line 101: Do we need to check both the X and the Y direction? Isn't it enough to block static scale only in the vertical direction?

TorsteinHonsi (changes requested):
Thanks! In addition to the inline comment, do we have a good explanation for the difference in the visual test?

ts/Extensions/StaticScale.ts
karolkolodziej: You are right! There is no need to check both.

karolkolodziej:
I went through the demo from visual differences, and it's hard to find the origin of that difference.
When you review the file difference, you can see that the width of the chart was changed.
Inside this PR, I didn't play with the chart width, so I guess that these changes are irrelevant. Correct me please if I'm wrong.

TorsteinHonsi (approved):
Thanks! Approved, awaiting CI tests after the last commit.

TorsteinHonsi:
The actual fix in the PR (checking `scrollablePixelsY`) is already applied in the master. So if the original issue still persists, it must be caused by something else.

PR #14229, opened 2020-09-15
ts/shadow-options-object-type

marvin19:
Put ShadowOptionsObject in its own type definition file.


bre1470 (changes requested):
So far it looks good. `SVGRenderer.ts` also defines `Highcharts.ShadowOptionsObject`, so it should be removed there, too.
https://github.com/highcharts/highcharts/blob/ts/shadow-options-object-type/ts/Core/Renderer/SVG/SVGRenderer.ts#L73-L79

bre1470 (approved):
Looks good! 👍

TorsteinHonsi (approved):

PR #14245, opened 2020-09-18
ts/animations

marvin19:
Created animation folder and extracted animation functions and Fx class into their own files. Added type definition files.


TorsteinHonsi (approved):
Thanks!

TorsteinHonsi:
Pending @bre1470 's approval

ts/Core/Animation/FxLike.d.ts
```diff
@@ -0,0 +1,26 @@
...
+ * */
+
+/* *
+ *
+ *  Declarations
+ *
+ * */
+export interface FxLike {
+
+}
```
bre1470, line 18:
```ts
/**
 * Optional properties for the Fx class. Should be defined outside of this file by `declare module` pattern.
 */
export interface FxLike {
    // nothing here yet
}
```

bre1470 (changes requested):
Looks good. 👍

Though `FxLike` needs some explanation for others to understand the purpose.

bre1470 (approved):

PR #14249, opened 2020-09-21
docs/13838-fix-tooltip-positioner

bre1470:
- Fixed arguments typing of TooltipPositionerCallbackFunction.


bre1470:
@TorsteinHonsi There are three visual tests failing, but I do believe it is unrelated and just caused by a resize of the container.

TorsteinHonsi (approved):
Thanks! Yes, the visual tests can't possibly derive from this PR

PR #14252, opened 2020-09-22
docs/fix-point-options

bre1470:
Fixed options types of data point.


TorsteinHonsi (changes requested):
Do we want to expose `undefined` as a valid point value and thereby clutter up the API? It does not work in plain number config: https://jsfiddle.net/highcharts/06euoc5n/. It probably works just by accident in array configs.

bre1470:
Probably not, because it is not part of the JSON specifications. Will fix.

TorsteinHonsi (approved):

PR #14258, opened 2020-09-23
bugfix/13751-series-clippath

### LINKED ISSUE #13751: With stock module, the righthand side of the gragh becomes missing when the the number of digits on Y-axis is dynamically decreased.
#### Expected behaviour
The points, lines, and areas of the graph should be drawn without loss even if the value of the Y axis changes dynamically.

#### Actual behaviour
**With stock module,**
the righthand side of the gragh becomes missing when the the number of digits on Y-axis is dynamically decreased.

1. press Add new series 
2. press remove series
 => the righthand side of the gragh becomes missing

##### With stock module
[Reproduction in CodeSandbox](https://codesandbox.io/s/with-stock-module-7w392)
With stock module

##### Without stock module
[Reproduction in CodeSandbox](https://codesandbox.io/s/without-stock-module-p5yfz)
Without stock module

#### Product version
highcharts: 8.1.2
highcharts-vue: 1.3.5

#### Affected browser(s)
Chrome 81

pawelfus:
Fixed #13751, wrong clip path after adding/removing series.

### DIFF
--- a/js/Core/Chart/StockChart.js
+++ b/js/Core/Chart/StockChart.js
@@ -724,7 +724,7 @@ addEvent(LineSeries, 'render', function () {
             }
         }
         // First render, initial clip box
-        if (!this.clipBox && this.animate) {
+        if (!this.clipBox && !chart.hasRendered) {
             this.clipBox = merge(chart.clipBox);
             this.clipBox.width = this.xAxis.len;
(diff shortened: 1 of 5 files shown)


TorsteinHonsi (approved):
Thanks!

PR #14263, opened 2020-09-24
Fixed #14230, time options not always updating

### LINKED ISSUE #14230: time.useUTC doesn't update the second time
#### Expected behaviour
useUTC should get updated every time.

#### Actual behaviour
useUTC is updated only the first time.

https://jsfiddle.net/aunvoq4f/2/

#### Product version
Highcharts 8.0.4

#### Affected browser(s)
All

TorsteinHonsi:
Fixed #14230, `chart.update` failed to update `time` options in some cases.


pawelfus (approved):
Thanks! 👍

PR #14271, opened 2020-09-25
bugfix/13956-annotations-different-yAxis-size

### LINKED ISSUE #13956: Incorrect annotation labels position for multiple annotations in multiple axis chart
#### Expected behaviour
Annotation should indicate the point as it is set to the point.
Visualization (setting y pixel position - it doesn't work well when moving):
https://jsfiddle.net/BlackLabel/3n5tgeL7/

#### Actual behaviour
Annotation for duplicate axes does not align correctly with the set point:
https://jsfiddle.net/BlackLabel/rxoczyva/

#### Product version
v8.1.2

Basts:
Fixed #13956, annotations had the wrong position for resized Y axis.

### DIFF
--- a/js/Extensions/Annotations/Mixins/ControllableMixin.js
+++ b/js/Extensions/Annotations/Mixins/ControllableMixin.js
@@ -126,7 +126,7 @@ var controllableMixin = {
         return {
             relativePosition: anchor,
             absolutePosition: merge(anchor, {
-                x: anchor.x + plotBox.translateX,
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (changes requested):
Thanks!

While we're at it, we should do the same fix on the X axis: https://jsfiddle.net/highcharts/02xad3mp/

TorsteinHonsi (approved):

PR #14284, opened 2020-09-30
Fixed #14277, update scrollablePlotArea fixed elements in every applyFixed

### LINKED ISSUE #14277: Updating scrollablePlotArea.minHeight breaks yAxis and bar category labels
#### Expected behaviour

* Updating `scrollablePlotArea.minHeight` redraws the chart so that the y-axis remains at the proper location.
* Updating `scrollablePlotArea.minHeight` redraws the chart so that the bar categories labels are redrawn if there is now space for them/no longer space for them.

#### Actual behaviour

1. If `yAxis` is present, changing `minHeight` doesn't update `yAxis` position properly (it gets "stuck" in the middle of the chart, if the chart has beccome bigger).
2. If `series`/`categories` are increased and *then* `minHeight` is increased, canvas size increased but bars aren't redrawn, so it just creates extra blank space.
3. If `series`/`categories` and `minHeight` are updated simultaneously, the category labels aren't updated (some will be missing when increasing the chart size). Only if `series`/`categories` is set *after* `minHeight` does it update the labels.

https://codesandbox.io/s/beautiful-hellman-7tnvp

highcharts-minheight-bugs

Note: reload the page in between reproduction attempts to reset the chart state.

**Bug #​1 reproduction steps:**

1. Click `Toggle y-axis definition`.
2. Click `Update entire options object`  (y-axis label should update to `My y-axis`).
3. Click `Toggle size`.
4. Click `Update entire options object` (increases the bars/minHeight).
5. Click `Toggle size`.
6. Click `Update entire options object` (decreases the bars/minHeight).
7. Click `Toggle size`.
8. Click `Update entire options object` (increases the bars/minHeight again).
9. Observe that the y-axis is in the wrong location.
10. Reload and do the same with the yAxis definition turned *off*, and the yAxis will behave as expected.

**Bug #​2 reproduction steps:**

1. Click `Toggle size`.
2. Click `Update series/categories`.
3. Click `Update minHeight`.
4. Observe that the canvas size has increased, but the bars haven't been redrawn to fill the space.

**Bug #​3 reproduction steps:**

1. Click `Toggle size`.
2. Click `Update entire options object` or `Update minHeight and series/categories simultaneously`.
3. Observe that some bar categories don't have labels even though there is now space for them.
4. Click `Update series/categories`.
5. Observe that the bars categories are now redrawn so that all have labels.

#### Product version

Highcharts v8.2.0

#### Affected browser(s)

Chrome and Firefox, haven't tried any others.

mayhewluke:
This fixed the first issue mentioned in #14277. The issue seems to occur because `moveFixedElements` is only called on the _initial_ call to `applyFixed`. If the `minHeight` is then changed after that initial call, subsequent `applyFixed` invocations do not run `moveFixedElements`. This changes `applyFixed` to _always_ run `moveFixedElements`, which ensures that the fixed elements are always placed appropriately on every render.

I don't know if running it every time (rather than on some specific event) is overkill or might introduce performance or other concerns, so if there's a better place for `moveFixedElements` to be run I'm happy to update the PR accordingly.

### DIFF
--- a/ts/Extensions/ScrollablePlotArea.ts
+++ b/ts/Extensions/ScrollablePlotArea.ts
@@ -400,8 +400,6 @@ Chart.prototype.applyFixed = function (): void {
             .addClass('highcharts-scrollable-mask')
             .add();
 
-        this.moveFixedElements();
-
         addEvent(this, 'afterShowResetZoom', this.moveFixedElements);
         addEvent(this, 'afterLayOutTitles', this.moveFixedElements);
 
@@ -413,6 +411,8 @@ Chart.prototype.applyFixed = function (): void {
             this.chartHeight
         );
     }
+
+    this.moveFixedElements();
 
     // Increase the size of the scrollable renderer and background
     scrollableWidth = this.chartWidth + (this.scrollablePixelsX || 0);


TorsteinHonsi:
Thanks!

These lines indicates in what circumstances we should do `moveFixedElements`:

```js
        addEvent(this, 'afterShowResetZoom', this.moveFixedElements);
        addEvent(this, 'afterLayOutTitles', this.moveFixedElements);
```

As you say, the reason for not applying it all the time would be performance, therefore we do it the first time and subsequently on certain events. Probably it should be done after `chart.update` too, but I don't have time to test this now.

mayhewluke:
Thanks for the feedback @TorsteinHonsi! I get that those are the circumstances that `moveFixedElements` is _currently_ being done under, but based on the bug report in #14277 it appears that that's not enough.

I'd be happy to revise the PR to use the correct event instead, but I can't seem to find any documentation on these events - the two you mention above aren't in the docs anywhere. Would you be able to point me to where I can find a list of events like these, or let me know which is the right event to use?

khlieng:
This can probably be closed because of #15013.

PR #14286, opened 2020-10-01
docs/fix-typos

khlieng:
This fixes some typos in the docs, the series.name newline change prevents a space being inserted after the dot in the generated HTML.


ts/Core/Options.ts
```diff
@@ -4050,8 +4050,8 @@ H.defaultOptions = {
 
         /**
          * The HTML of the point's line in the tooltip. Variables are enclosed
-         * by curly brackets. Available variables are point.x, point.y, series.
-         * name and series.color and other properties on the same form.
+         * by curly brackets. Available variables are point.x, point.y,
```
TorsteinHonsi, line 4053: While we're at it, we should add backticks to code snippets too: 

> Available variables are \`point.x\`, \`point.y\`, \`series.name\` and \`series.color\` and other properties on the same form.

TorsteinHonsi (changes requested):
Thanks! Good observations!

TorsteinHonsi (approved):
Thanks!

PR #14309, opened 2020-10-07
bugfix/14214-negative-value

### LINKED ISSUE #14214: Parliament Chart throws error for negative values
Expected behaviour
        If any one of the inputs contains negative value, parliament chart can simply ignore that negative value and plot rest of the values.

Actual behaviour
        Parliament chart throws error if the input contains negative values.

Demo
       https://jsfiddle.net/s3dgua1t/

Product version
      Highcharts v8.2.0

marvin19:
Fixed #14214, parliament chart (item series type) threw error on negative values. Negative values should be ignored like in the pie series.

~~Fixed #14214, replacing negative values with null in pie-based series.~~


ts/Series/PieSeries.ts
```diff
@@ -864,6 +862,16 @@ BaseSeries.seriesType<typeof Highcharts.PieSeries>(
          */
         generatePoints: function (this: Highcharts.PieSeries): void {
             LineSeries.prototype.generatePoints.call(this);
+            const data = this.data;
```
TorsteinHonsi, line 865: Why do we need this new block? The original fiddle seems to work without it...

TorsteinHonsi (changes requested):
Thanks! In addition to the inline comment, please add a unit test.

ts/Series/PieSeries.ts
marvin19: The fix replaces negative values with null on the processed data. Data labels require null values on point instance to align correctly to their related element.

This is the incorrect rendering without the block:
highcharts-item-chart

This is the correct rendering with the block:
highcharts-item-chart (1)

TorsteinHonsi: I see! I moved this check to the translate function where the angles are calculated, I think that's sufficient. Commit coming up.

TorsteinHonsi (changes requested):
Pending updated tests that actually check if the graphic is visible or not.

TorsteinHonsi (approved):
Thanks!

PR #14311, opened 2020-10-07
ExportData Data-layer integration

goransle:
ToDo:
- [x] Visual tests
- [x] Await #14045 and integrate AST where applicable


ts/Core/Renderer/HTML/AST.ts
```diff
@@ -243,6 +243,33 @@ class AST {
         }
     }
 
+
+    public static serialize(node: Highcharts.ASTNode): string {
```
goransle, line 247: @TorsteinHonsi Any objections to moving this `serialize` functionality to the AST class?

PR #14318, opened 2020-10-08
bugfix/14298-cancelable-touch-events

### LINKED ISSUE #14298: Check cancelable on touch events before calling preventDefault
I enabled reporting api (`Report-To`) on my website, and get an intervention report from Chrome browsers with the following message:

```json
{
  "columnNumber": 169,
  "id": "IgnoredEventCancel",
  "lineNumber": 63,
  "message": "Ignored attempt to cancel a touchend event with cancelable=false, for example because scrolling is in progress and cannot be interrupted.",
  "sourceFile": "highstock.js?v=32550416a08fb3d3"
}
```

Looking at the minified file, I believe it is being caused by this:
https://github.com/highcharts/highcharts/blob/aaa230381d4423928080c00b7187dd78d01b459c/ts/Core/Renderer/SVG/SVGElement.ts#L2123

I think the fix is to check if `e.cancelable` is true before calling it.

#### Product version

Highstock v8.2.0

#### Affected browser(s)
Chrome 85 on Android (or touch devices)

khlieng:
Fixed #14298, Chrome complained about touch events not being `cancelable` when calling `preventDefault`.

### DIFF
--- a/ts/Accessibility/AccessibilityComponent.ts
+++ b/ts/Accessibility/AccessibilityComponent.ts
@@ -469,7 +469,9 @@ AccessibilityComponent.prototype = {
                 }
 
                 e.stopPropagation();
-                e.preventDefault();
+                if (e.cancelable) {
+                    e.preventDefault();
+                }
             });
         });
     },
--- a/ts/Accessibility/AccessibilityComponent.ts
+++ b/ts/Accessibility/AccessibilityComponent.ts
@@ -469,9 +469,7 @@ AccessibilityComponent.prototype = {
                 }
 
                 e.stopPropagation();
-                if (e.cancelable) {
-                    e.preventDefault();
-                }
+                e.preventDefault();
             });
         });
     },
--- a/ts/Core/Pointer.ts
+++ b/ts/Core/Pointer.ts
@@ -1436,7 +1436,8 @@ class Pointer {
 
         // On touch devices, only proceed to trigger click if a handler is
         // defined
-        if (hasZoom && self.initiated && !fireClickEvent && e.cancelable) {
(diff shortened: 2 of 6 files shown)


ts/Accessibility/AccessibilityComponent.ts
```diff
@@ -469,7 +469,9 @@ AccessibilityComponent.prototype = {
                 }
 
                 e.stopPropagation();
-                e.preventDefault();
+                if (e.cancelable) {
+                    e.preventDefault();
```
oysteinmoseng, line 473: This should always run `preventDefault` - regardless of `cancelable`.

oysteinmoseng (changes requested):
Just two concerns:
1. What happens in browsers that do not support `event.cancelable`?
2. In AccessibilityComponent we should always run `preventDefault` as the event is moved to a different element. See inline comment.

ts/Core/Pointer.ts
```diff
@@ -1436,7 +1436,8 @@ class Pointer {
 
         // On touch devices, only proceed to trigger click if a handler is
         // defined
-        if (hasZoom && self.initiated && !fireClickEvent && e.cancelable) {
+        if (hasZoom && self.initiated && !fireClickEvent &&
+            (typeof e.cancelable !== 'boolean' || e.cancelable)) {
```
oysteinmoseng, line 1440: Just nitpicking, but you might be able to get away with `e.cancelable !== false` and save a few bytes of filesize.

oysteinmoseng (changes requested):
Looks good, just a minor comment

oysteinmoseng (approved):
Looks good to me, thanks!

TorsteinHonsi (approved):

PR #14322, opened 2020-10-08
bugfix/14310-plotband-axis-visibility

### LINKED ISSUE #14310: plotBand added with addPlotBand on not visible xAxis doesn't respond to chart zooming
#### Expected behaviour

1. Prevent adding plotBand by using `addPlotBand` on `xAxis` with `visible: false`. If you add plotBand directly in `xAxis` section with `visible: false` then plotBand won't be shown.
2. (preferable) Respond to zoom correctly

#### Actual behaviour

`plotBand` added with `addPlotBand` on not visible `xAxis` doesn't respond to chart zooming:

Oct-07-2020 15-28-13

When `xAxis` is visible:

Oct-07-2020 15-29-12

https://jsfiddle.net/smokir/qx0tnf87/

#### Product version

Highcharts

#### Affected browser(s)

Chrome 85.0.4183.1697

khlieng:
Fixed #14310, `addPlotBandOrLine` didn't check if the axis was visible before rendering.

### DIFF
--- a/js/Core/Axis/PlotLineOrBand.js
+++ b/js/Core/Axis/PlotLineOrBand.js
@@ -947,7 +947,10 @@ extend(Axis.prototype, /** @lends Highcharts.Axis.prototype */ {
      * @return {Highcharts.PlotLineOrBand|undefined}
      */
     addPlotBandOrLine: function (options, coll) {
-        var obj = new PlotLineOrBand(this, options).render(), userOptions = this.userOptions;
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (changes requested):
Thanks! We also need a test to verify that a) the band is not rendered in a hidden axis, and b) when the same axis' visibility is changed to true (after `update()`), the band is rendered too.

TorsteinHonsi (approved):
Looking good 👍

PR #14339, opened 2020-10-13
bugfix/14254-missing-options-arg

### LINKED ISSUE #14254: Highstock plotBands acrossPanes flag has no effect
#### Expected behaviour
Setting `acrossPanes` to false for plotBands in Highstock should only render plot band in single pane.

#### Actual behaviour
Setting `acrossPanes` to false for plotBands in Highstock has no effect. (Possibly related to fix  for #6644 )

https://jsfiddle.net/xkuc5dvt/ (Notice plot band spans all panes)

#### Product version
Highstock 8.2.0

#### Affected browser(s)
Chrome

khlieng:
Fixed #14254, the `plotBands.acrossPanes` option had no effect.

~Fixed #14254, `getPlotBandPath` had no options argument, but its interface did and other callers were using it.~

### DIFF
--- a/ts/Core/Axis/PlotLineOrBand.ts
+++ b/ts/Core/Axis/PlotLineOrBand.ts
@@ -1151,17 +1151,17 @@ extend(Axis.prototype, /** @lends Highcharts.Axis.prototype */ {
         this: Highcharts.Axis,
         from: number,
         to: number,
-        options?: (Highcharts.AxisPlotBandsOptions|Highcharts.AxisPlotLinesOptions)
+        options: (Highcharts.AxisPlotBandsOptions|Highcharts.AxisPlotLinesOptions) = this.options
     ): SVGPath {
         var toPath = this.getPlotLinePath({
                 value: to,
                 force: true,
-                acrossPanes: options?.acrossPanes ?? (this.options as any).acrossPanes
+                acrossPanes: options.acrossPanes
             } as Highcharts.AxisPlotLinePathOptionsObject),
             path = this.getPlotLinePath({
                 value: from,
                 force: true,
-                acrossPanes: options?.acrossPanes ?? (this.options as any).acrossPanes
+                acrossPanes: options.acrossPanes
             } as Highcharts.AxisPlotLinePathOptionsObject),
             result = [] as SVGPath,
             i,
--- a/ts/Core/Axis/PlotLineOrBand.ts
+++ b/ts/Core/Axis/PlotLineOrBand.ts
@@ -1141,23 +1141,27 @@ extend(Axis.prototype, /** @lends Highcharts.Axis.prototype */ {
      * @param {number} to
      *        The axis value to end on.
      *
+     * @param {Highcharts.AxisPlotBandsOptions|Highcharts.AxisPlotLinesOptions} options
+     *        The plotBand or plotLine configuration object.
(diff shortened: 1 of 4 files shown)


ts/Core/Axis/PlotLineOrBand.ts
```diff
@@ -1141,23 +1141,27 @@ extend(Axis.prototype, /** @lends Highcharts.Axis.prototype */ {
...
         from: number,
-        to: number
+        to: number,
+        options?: (Highcharts.AxisPlotBandsOptions|Highcharts.AxisPlotLinesOptions)
     ): SVGPath {
         var toPath = this.getPlotLinePath({
                 value: to,
                 force: true,
-                acrossPanes: (this.options as any).acrossPanes
+                acrossPanes: options?.acrossPanes ?? (this.options as any).acrossPanes
```
bre1470, line 1159: Avoid optional markers as they bloat the code. Instead do shorter `&&` `||` checks and remove the `any` cast at the same time.

ts/Core/Axis/PlotLineOrBand.ts
```diff
@@ -1141,23 +1141,27 @@ extend(Axis.prototype, /** @lends Highcharts.Axis.prototype */ {
...
             } as Highcharts.AxisPlotLinePathOptionsObject),
             path = this.getPlotLinePath({
                 value: from,
                 force: true,
-                acrossPanes: (this.options as any).acrossPanes
+                acrossPanes: options?.acrossPanes ?? (this.options as any).acrossPanes
```
bre1470, line 1164: Avoid optional markers as they bloat the code. Instead do shorter `&&` `||` checks and remove the `any` cast at the same time.

ts/Core/Axis/PlotLineOrBand.ts
```diff
@@ -1141,23 +1141,27 @@ extend(Axis.prototype, /** @lends Highcharts.Axis.prototype */ {
...
+     *
      * @return {Highcharts.SVGPathArray}
      *         The SVG path definition in array form.
      */
     getPlotBandPath: function (
         this: Highcharts.Axis,
         from: number,
-        to: number
+        to: number,
+        options?: (Highcharts.AxisPlotBandsOptions|Highcharts.AxisPlotLinesOptions)
```
bre1470, line 1154: Make use of default values:
```ts
getPlotBandPath: function (
    this: Highcharts.Axis,
    from: number,
    to: number,
    options: (Highcharts.AxisPlotBandsOptions|Highcharts.AxisPlotLinesOptions) = this.options
```

bre1470 (changes requested):
Some optimization is needed.

ts/Core/Axis/PlotLineOrBand.ts
khlieng: How do I get rid of the `any` cast when `AxisOptions` doesn't have `acrossPanes`?

bre1470 (approved):
Thanks! Looks good. 👍

TorsteinHonsi (changes requested):
Thanks!

Please add a unit test to. We can for example create one plot band which is across panes and one which is not, and verify that the height of the bands is different.

TorsteinHonsi (approved):

Issue #12528, opened 2019-11-28
New Functionality: Drilldown go-to-level + breadcrumbs


Izothep:
Adding go-to-level functionality for drilldown to specific level. Additionally, create breadcrumbs in drilldown module.

bbrudnoy:
Presumably this will introduce breadcrumbs in treemap charts. Certainly thumbs up. This a desired feature in my company.

karolkolodziej:
The breadcrumbs functionality introduced here: https://github.com/highcharts/highcharts/pull/15767

PR #14341, opened 2020-10-13
enhancement/12528-breadcrumbs-functionality

karolkolodziej:
Feature #12528, initial commit.

To do: 
- [x] POC
- [x] Support for treemap
- [ ] a11y
- [x] `styledMode`
- [x] Tests
- [ ] Demos
- [ ] TS
- [ ] Styling


ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,658 @@
...
+     * @requires  modules/drilldown
+     *
+     * @function Highcharts.Chart#createList
+     * @param {Highcharts.Breadcrumbs} this
+     *        Breadcrumbs class.
+     */
+    public createList(this: Breadcrumbs): void {
+        const chart = this.chart,
+            drilldownLevels = chart.drilldownLevels,
+            initialSeriesLevel = -1;
```
Izothep, line 334: It would be good to add comment why it is set to -1

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,658 @@
...
+        let breadcrumbsList: Array<any> = this.breadcrumbsList || [],
+            // If the list doesn't exist treat the initial series
+            // as the current level- first iteration.
+            currentLevelNumber: number = breadcrumbsList.length ?
+                breadcrumbsList[breadcrumbsList.length - 1][0] : initialSeriesLevel;
+
+        if (drilldownLevels && drilldownLevels.length) {
+            // Add the initial series as the first element.
+            breadcrumbsList = !breadcrumbsList.length ?
+                [[initialSeriesLevel, drilldownLevels[0].seriesOptions]] : breadcrumbsList;
```
Izothep, line 345: It looks that the `currentLevelNumber` from `339` is never used as here we use `initialSeriesLevel` and in `351` we are editing `currentLevelNumber`.

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,658 @@
...
+     *
+     * @requires  modules/drilldown
+     *
+     * @function Highcharts.Chart#destroy
+     * @param {Highcharts.Breadcrumbs} this
+     *        Breadcrumbs class.
+     * @param {Highcharts.Breadcrumbs} lastVisibleLevel
+     *        Number of levels to destoy.
+     */
+    public destroy(this: Breadcrumbs, lastVisibleLevel: number): void {
```
Izothep, line 370: I would also add the possibility when lastVisibleLevel is not set - then it would destroy all the breadcrumbs.

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,658 @@
...
+     * @param {Highcharts.Breadcrumbs} this
+     *        Breadcrumbs class.
+     */
+    public draw(this: Breadcrumbs): void {
+        const breadcrumbs = this,
+            chart = breadcrumbs.chart,
+            breadcrumbsList = breadcrumbs.breadcrumbsList,
+            breadcrumbsOptions = chart.options.drilldown?.breadcrumbs,
+            lastBreadcrumbs = breadcrumbsList[breadcrumbsList.length - 2][2],
+            buttonPadding = 5;
```
Izothep, line 429: should it not be configurable?

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,658 @@
...
+
+        // A main group for the breadcrumbs.
+        if (!this.breadcrumbsGroup) {
+            this.breadcrumbsGroup = chart.renderer
+                .g('breadcrumbs-group')
+                .attr({
+                    zIndex: 7
```
Izothep, line 436: I think that zIndex should be configurable.

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,658 @@
...
+     * @function Highcharts.Chart#draw
+     * @param {Highcharts.Breadcrumbs} this
+     *        Breadcrumbs class.
+     */
+    public draw(this: Breadcrumbs): void {
+        const breadcrumbs = this,
+            chart = breadcrumbs.chart,
+            breadcrumbsList = breadcrumbs.breadcrumbsList,
+            breadcrumbsOptions = chart.options.drilldown?.breadcrumbs,
+            lastBreadcrumbs = breadcrumbsList[breadcrumbsList.length - 2][2],
```
Izothep, line 428: `lastBreadcrumb`?

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,658 @@
...
+     *
+     * @requires  modules/drilldown
+     *
+     * @function Highcharts.Chart#multipleDrillUp
+     * @param {Highcharts.Breadcrumbs} this
+     *        Breadcrumbs class.
+     * @param {Highcharts.Breadcrumbs} drillAmount
+     *        A number of drillups that needs to be performed.
+     */
+    public multipleDrillUp(this: Breadcrumbs, drillAmount: number): void {
```
Izothep, line 483: If this method should stay, it should be moved outside of Breadcrumbs class

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,658 @@
...
+    });
+
+    addEvent(Chart, 'redraw', function (e: Chart): void {
+        this.drilldown?.breadcrumbs?.redraw();
+    });
+
+    addEvent(Chart, 'afterDrilldown', function (e: Chart): void {
+        const chart = this;
+
+        if (!chart.drilldown?.breadcrumbs && chart.drilldown) {
```
Izothep, line 647: I would add check if breadcrumbs are enabled.

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,658 @@
...
+            chart.drilldown.breadcrumbs = new Breadcrumbs(chart as Chart, chart.options as Highcharts.Options);
+        }
+
+        if (chart.drilldown && chart.options.drilldown?.breadcrumbs?.enabled && chart.drilldown?.breadcrumbs) {
```
Izothep, line 651: Actually, we can combine these two if's into one

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,658 @@
...
+                function (): void {
+                    if (breadcrumbsOptions.events && breadcrumbsOptions.events.click) {
+                        breadcrumbsOptions.events.click(button as any, breadcrumbs as any);
+                        // Prevent from click on the current level
+                    } else if (breadcrumbsList[breadcrumbsList.length - 1][1].name !== button.textStr) {
+                        breadcrumbs.multipleDrillUp(breadcrumb[0]);
+                    }
+                })
+                .attr({
+                    padding: 3
```
Izothep, line 552: It should be configurable

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,658 @@
...
+    * @requires  modules/drilldown
+    *
+    * @function Highcharts.Chart#update
+    * @param {Highcharts.Breadcrumbs} this
+    *        Breadcrumbs class.
+    * @param {Highcharts.Options} userOptions
+    *        Breadcrumbs class.
+    */
+    public update(this: Breadcrumbs, userOptions: Highcharts.Options): void {
+        const chartOptions = H.merge(userOptions, this.options);
```
Izothep, line 616: We are updating the options but not the buttons themselves. I think we should draw buttons again in here (or in `redraw` method). If in `redraw`, we should add `redraw` flag as `update` method's argument.

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,658 @@
...
+
+/* eslint-disable no-invalid-this */
+
+if (!H.Breadcrumbs) {
+    H.Breadcrumbs = Breadcrumbs as any;
+
+    addEvent(Chart, 'getMargins', function (e: Chart): void {
+        const chart = this,
+            breadcrumbsOptions = chart.options.drilldown?.breadcrumbs,
+            extraMargin = chart.rangeSelector ? 40 : 30;
```
Izothep, line 628: why the value is fixed to 30/40?

Izothep (changes requested):

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,658 @@
...
+    public breadcrumbsList: Array<Array<any>> = [];
+    public options: Highcharts.BreadcrumbsOptions = void 0 as any;
+
+    public constructor(chart: Chart, userOptions: Highcharts.Options) {
+        this.init(chart, userOptions);
+    }
+
+    public init(chart: Chart, userOptions: Highcharts.Options): void {
+        const chartOptions = H.merge(userOptions, this.options),
+            breadcrumbsOptions = chartOptions.drilldown?.breadcrumbs as Highcharts.BreadcrumbsOptions;
```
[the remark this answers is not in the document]
karolkolodziej: Removed optional chaining from file

ts/Extensions/Breadcrumbs.ts
karolkolodziej: It's the last rendered button, used for calculating the position of the next one. Any clue for a better name?

ts/Extensions/Breadcrumbs.ts
karolkolodziej: It was not in the docs, but I don't see any reason why it shouldn't be configurable.

ts/Extensions/Breadcrumbs.ts
karolkolodziej: Same as padding

ts/Extensions/Breadcrumbs.ts
karolkolodziej: I hardcoded this value, to make it visually acceptable for the PoC. I would like to hear the @TorsteinHonsi opinion about the overall design before I implement this.

ts/Extensions/Breadcrumbs.ts
karolkolodziej: Do you mean into `Chart`?

ts/Extensions/Breadcrumbs.ts
karolkolodziej: In `350` we are using `currentLevelNumber` to compare levels, but in this case (going only one level down) I agree that the `351` line is not necessary.

ts/Extensions/Breadcrumbs.ts
karolkolodziej: I changed  `40` into the range selector height.
Second value `30` is a bit more tricky because at that point the breadcrumbs don't exist and why I'm not able to get their height that is why I assumed some default height. I will be able to calculate that height when we have the final design.

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,898 @@
...
+        }
+        interface BreadcrumbsClickCallbackFunction {
+            (e: Event, breadcrumbs: Breadcrumbs): (boolean|undefined);
+        }
+        interface BreadcrumbsButtonsEventsOptions {
+            click?: BreadcrumbsClickCallbackFunction;
+        }
+        interface BreadcrumbsButtonsFormatter {
+            (
+                breadcrumb: Array<any>,
```
pawelfus, line 73: Could we define what's passed on? To avoid using `any`

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,898 @@
...
+            ): (string);
+        }
+        interface Options {
+            breadcrumbs?: BreadcrumbsOptions;
+        }
+        interface ChartLike {
+            breadcrumbs?: Breadcrumbs;
+        }
+        interface LangOptions {
+            mainBreadCrumb?: string;
```
pawelfus, line 84: Maybe `mainBreadcrumb`? (without capital "C")

ts/Extensions/Breadcrumbs.ts
sebastianbochan: The Rangeselector is dynamic (responsive design) so you cannot hardcode the height.

ts/Extensions/Breadcrumbs.ts
sebastianbochan: We should not use casting to `any`.

You can create your own type and use it here.

ts/Extensions/Breadcrumbs.ts
sebastianbochan: I would say that main sounds strange. Maybe we should consider to use "basicBreadcrump"  / "parentBreadcrump" /  "rootBreadcrump"

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,898 @@
...
+            click?: BreadcrumbsClickCallbackFunction;
+        }
+        interface BreadcrumbsButtonsFormatter {
+            (
+                breadcrumb: Array<any>,
+                breadcrumbs: Breadcrumbs
+            ): (string);
+        }
+        interface Options {
+            breadcrumbs?: BreadcrumbsOptions;
```
pawelfus, line 78: Breadcrumbs is a navigation aid feature. Perhaps we should place options in `navigation.breadcrumbs`?

ts/Extensions/Breadcrumbs.ts
pawelfus: I don't see this "Main" rendered anywhere. Did I miss something?

samples/unit-tests/breadcrubs/buttons/demo.js
```diff
@@ -0,0 +1,261 @@
...
+            name: "Supply",
+            data: [{
+                name: "Fruits",
+                y: 5,
+                drilldown: "Fruits"
+            },
+            {
+                name: "Vegetables",
+                y: 6,
+                drilldown: "Vegetables"
```
pawelfus, line 22: In this demo, go back to the top level and click on "Vegetables to drill down. Nothing happens even when drilldown is defined below. 

Demo: http://utils.highcharts.local/samples/#test/unit-tests/breadcrubs/buttons

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,898 @@
...
+            /**
+            * The default padding for each button and separator.
+            *
+            * @type      {boolean}
+            * @since     next
+            */
+            buttonPadding: 5,
+
+            /**
+            * Show only last button instead of whole drillUp tree.
```
pawelfus, line 144: Just an idea:
Instead of `showOnlyLast`, let's call it `showFullPath` or `showAboslutePath`. The option would be self-describing.

That reminds me, how about long paths? For example this: http://jsfiddle.net/BlackLabel/5hctznvw/2/

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,898 @@
...
+            * for the buttons.
+            * This property won't work when positioned in the middle.
+            *
+            * @type      {boolean}
+            * @since     next
+            */
+            floating: true,
+
+            /**
+            * A format string for the breadcrumbs button.
```
pawelfus, line 189: Could you improve a little this description? For example, what is the context for this format? Is that a `point.value`?

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,898 @@
...
+                /**
+                 * The Y offset of the breadcrumbs button group.
+                 */
+                y: 0
+            },
+
+            /**
+            * A CSS styles for all breadcrumbs.
+            *
+            * @type      {string}
```
pawelfus, line 242: I have a few questions here :)
- Breadcrumbs' buttons or texts? Does it apply to separators too? Can I set for example `max-width` for the button, and text will get ellipsis?
- `@type` seems wrong ;) 
- add `@defaults`, maybe I'm wrong, but thatthis is necessary for [`CSSObject`](https://api.highcharts.com/class-reference/Highcharts.CSSObject) interface

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,898 @@
...
+            */
+            useHTML: false,
+
+            /**
+            * The zIndex of the group.
+            *
+            * @type      {boolean}
+            * @since     next
+            */
+            zIndex: 7
```
pawelfus, line 281: Perhaps `8` would be better (to be higher than the selection marker in zoom). And we might as well increase tooltips label from 8 to 9 (so tooltip will be above the breadcrumbs).

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,898 @@
...
+    public level: number = -1;
+    public options: Highcharts.BreadcrumbsOptions = void 0 as any;
+
+    public constructor(chart: Chart, userOptions: Highcharts.Options) {
+        this.init(chart, userOptions);
+    }
+
+    public init(chart: Chart, userOptions: Highcharts.Options): void {
+        const chartOptions = merge(userOptions, this.options),
+            breadcrumbsOptions = chart.options.breadcrumbs;
```
pawelfus, line 312: `chartOptions` seems to not used here. Can be removed?

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,898 @@
...
+     * @param {Highcharts.Breadcrumbs} this
+     *        Breadcrumbs class.
+     * @param {Highcharts.Breadcrumbs} lastVisibleLevel
+     *        Number of levels to destoy.
+     */
+    public destroy(this: Breadcrumbs, lastVisibleLevel: number|undefined): void {
+        const breadcrumbs = this,
+            chart = breadcrumbs.chart,
+            breadcrumbsList = breadcrumbs.breadcrumbsList,
+            breadcrumbsOptions = chart.options.breadcrumbs;
```
pawelfus, line 478: General issue (see also other methods):
Instead of referring to `chart.options.breadcrumbs` use `breadcrumbs.options`

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,898 @@
...
+        // A main group for the breadcrumbs.
+        if (!this.breadcrumbsGroup && breadcrumbsOptions) {
+            this.breadcrumbsGroup = chart.renderer
+                .g('breadcrumbs-group')
+                .attr({
+                    zIndex: breadcrumbsOptions.zIndex
+                })
+                .addClass('highcharts-breadcrumbs')
+                .add();
+            chart.redraw();
```
pawelfus, line 565: Why do we need to redraw the chart here? In most scenarios, we use `chart.redraw()` at the end (if needed).

ts/Extensions/Breadcrumbs.ts
sebastianbochan: We use navigation also in stockTools, I'm afraid that it can be tricky here.

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,898 @@
...
+     * @function Highcharts.Breadcrumbs#multipleDrillUp
+     * @param {Highcharts.Breadcrumbs} this
+     *        Breadcrumbs class.
+     * @param {Highcharts.Breadcrumbs} drillAmount
+     *        A number of drillups that needs to be performed.
+     */
+    public multipleDrillUp(this: Breadcrumbs, drillAmount: number): void {
+        const chart = this.chart,
+            breadcrumbsList = this.breadcrumbsList,
+            drillNumb = defined(drillAmount) ?
```
pawelfus, line 646: 🎶   I've become so numb... 🎶

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,898 @@
...
+            }
+
+            const button = chart.renderer.button(
+                breadcrumbsOptions.formatter ? breadcrumbsOptions.formatter(breadcrumb, breadcrumbs) : void 0 ||
+                    format(arrow + breadcrumbsOptions.format as string,
+                        { value: breadcrumb[1].name || (lang && lang.mainBreadCrumb) }, chart),
+                posX,
+                posY,
+                function (e: (Event|Highcharts.Dictionary<any>)): void {
+                    // extract events from button object and call
```
pawelfus, line 727: In general, when we write a full-line comment, we start it with a capital letter, for example:

```js
// My comment
doSomething();
```
But this should not use capital letter:
```js
doSomething(); // my comment
```

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,898 @@
...
+    * @param {Highcharts.Breadcrumbs} posY
+    *        Initial vertical position
+    * @return {SVGElement|void}
+    *        Returns the SVG button
+    */
+    public renderSeparator(this: Breadcrumbs, posX: number, posY: number): SVGElement|undefined {
+        const breadcrumbs = this,
+            chart = this.chart,
+            breadcrumbsOptions = chart.options.breadcrumbs,
+            size = 10;
```
pawelfus, line 775: Should we somehow connect size and font-size? Or make this configurable, so the change could be like this:

```js
separator: 'triangle'
```
->
```js
separator: {
  symbol: 'triangle',
  size: 10
  ... // place for more options in the future, maybe styles/attributes?
}
```

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,898 @@
...
+    *        Initial horizontal position
+    * @param {Highcharts.Breadcrumbs} posY
+    *        Initial vertical position
+    * @return {SVGElement|void}
+    *        Returns the SVG button
+    */
+    public renderSeparator(this: Breadcrumbs, posX: number, posY: number): SVGElement|undefined {
+        const breadcrumbs = this,
+            chart = this.chart,
+            breadcrumbsOptions = chart.options.breadcrumbs,
```
pawelfus, line 774: Another example when referring to `this.options` would be better: `renderSeparator()` then could always return SVGElement, right?

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,898 @@
...
+
+        if (breadcrumbsOptions && breadcrumb && breadcrumb[1]) {
+            if (breadcrumbsOptions && breadcrumbsOptions.showOnlyLast) {
+                arrow = '◁ ';
+            }
+
+            const button = chart.renderer.button(
+                breadcrumbsOptions.formatter ? breadcrumbsOptions.formatter(breadcrumb, breadcrumbs) : void 0 ||
+                    format(arrow + breadcrumbsOptions.format as string,
+                        { value: breadcrumb[1].name || (lang && lang.mainBreadCrumb) }, chart),
```
pawelfus, line 723: Something is wrong here (I guess only), see: http://jsfiddle.net/BlackLabel/5hctznvw/3/ - click on Chrome and check the button

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,898 @@
...
+        const chart = this,
+            breadcrumbs = chart.breadcrumbs as Highcharts.Breadcrumbs,
+            breadcrumbsOptions = chart.options.breadcrumbs;
+
+        if (breadcrumbsOptions && breadcrumbsOptions.enabled) {
+
+            if (!breadcrumbs) {
+                chart.breadcrumbs = new Breadcrumbs(chart as Chart, chart.options as Highcharts.Options);
+            }
+            chart.redraw();
```
pawelfus, line 871: This is an interesting case:
- drilldown call `chart.redraw()`, then fires `afterDrilldown` event
- breadcrumbs add `afterDrilldown` and fires `chart.redraw()`

I think we can somehow improve this, in future similar cases can happen too. I see two possible solutions:

- create `afterDrilldownBeforeRedraw` event and place it before `chart.redraw()`
- move `fireEvent(this, 'afterDrilldown');` before `chart.redraw()`

The first one pollutes events scope. The second one might have unexpected results and bugs. The current solution (multiple redraws) is from the performance perspective is a bad idea.

Any other ideas how to solve it? What do you think @TorsteinHonsi ?

samples/unit-tests/breadcrubs/breadcrumbs-in-treemap/demo.html
```diff
@@ -0,0 +1,10 @@
+<script src="https://code.highcharts.com/highcharts.js"></script>
+<script src="https://code.highcharts.com/modules/data.js"></script>
+<script src="https://code.highcharts.com/modules/heatmap.js"></script>
+<script src="https://code.highcharts.com/modules/treemap.js"></script>
+<script src="https://code.highcharts.com/modules/breadcrumbs.js"></script>
```
TorsteinHonsi, line 5: Breadcrumbs should not be a separate module in the distribution. It should be a separate _ES module_, but imported into the drilldown and treemap distribution modules/JS files.

ts/Extensions/Breadcrumbs.ts
TorsteinHonsi: Yes, multiple redraws should be avoided. Other than that I don't have full overview over this, but I prefer alternative two instead of adding specialized event calls.

ts/Extensions/Breadcrumbs.ts
karolkolodziej: Yes, the capital `C` isn't necessary in this case, `mainBreadcrumb` looks much better.
It's only used when the series name is not declared. I will prepare the appropriate tests and demos.

ts/Extensions/Breadcrumbs.ts
karolkolodziej: `drilldown.breadcrumbs` is unfortunate because in `treemap` we don't use that and creating that whole structure might be confusing for the user. 
`navigation.breadcrumbs` sound reasonable and maybe more natural in this case where breadcrumbs are used to navigate through the chart.
What do you think @TorsteinHonsi ?

samples/unit-tests/breadcrubs/buttons/demo.js
karolkolodziej: In this case, the `drilldown` property is declared for a series but after that, this property isn't mentioned in the `drilldown` object. 
It might be misleading, I will change that.
```
series: [{
            name: "Supply",
            data: [{
                name: "Fruits",
                y: 5,
                drilldown: "Fruits"
            }, {
                name: "Vegetables",
                y: 6,
                drilldown: "Vegetables"
            }, {
                name: "Meat",
                y: 3
            }]
        }],
        drilldown: {
            animation: false,
            series: [{
                name: "Fruits",
                id: "Fruits",
                data: [{
                    name: "Citrus",
                    y: 2,
                    drilldown: "Citrus"
                }, {
                    name: "Tropical",
                    y: 5,
                    drilldown: "Tropical"
                }, ['Other', 1]
                ]
            }]
        }
```

ts/Extensions/Breadcrumbs.ts
karolkolodziej: The code for creating a button and changing the text was invoked simultaneously.
I added a separate method to handle the text change for a singular button with `format/formatter` support.

ts/Extensions/Breadcrumbs.ts
TorsteinHonsi: I think we should have `drilldown.breadcrumbs: BreadCrumbsOptions` and `plotOptions.treemap.breadcrumbs: BreadCrumbsOptions`. Sunburst will inherit from treemap. 

The idea is that this places the drill-up button for drilldown and the traverse-up button for treemap.

ts/Extensions/Breadcrumbs.ts
pawelfus: This is a more general issue, the second solution does not resolve this issue. The same problem we have with RangeSelector, Navigator, etc components that need to resize the plotting area. Some components (title/subtitle and legend) already do this in `chart.redraw()`. We will try to do this:
- create method in `chart` that will internally call `chart.layoutTitles()` and `legend.render()`
- method will fire event, allowing other chart components to reserve some space in the chart
- RangeSelector may now remove workaround, implement new event solution and fix a related bug - that will additionally improve performance (will be separate PR)
- Navigator may now easily fix related bugs (where Navigator width does not match the plotting area) -  separate bugfix/PR
- StockTools - no one reported a bug, but I will check this

ts/Extensions/Breadcrumbs.ts
karolkolodziej: In the `update` method we are merging the `options` with the `userOptions` so it isn't necessary to use them here.

ts/Extensions/Breadcrumbs.ts
karolkolodziej: @TorsteinHonsi We have changed that- the breadcrumbs options now are available inside `drilldown.breadcrumbs` and `treemap.breadcrumbs`.
What I would like to ask is where the breadcrumbs class should be?  Now you might reach it in the `chart`. Should we move it to `drilldown` and `treemap`, or can we leave that as it is now on the `chart`?

ts/Extensions/Breadcrumbs.ts
TorsteinHonsi: I agree. I prefer `showFullPath`.

ts/Extensions/Breadcrumbs.ts
TorsteinHonsi: Not sure if I understand. Do we need to register the class anywhere at all? Can't we just import the module into the drilldown and the treemap modules?

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,1035 @@
...
+    }
+}
+
+/* eslint-disable no-invalid-this */
+
+if (!H.Breadcrumbs) {
+    H.Breadcrumbs = Breadcrumbs as any;
+
+    addEvent(Chart, 'setOffsets', function (): void {
+        this.breadcrumbs && this.breadcrumbs.setMargins();
```
karolkolodziej, line 947: When `floating: false` space is not reserved after the first click. Fix and add a test.

PR #14353, opened 2020-10-14
bugfix/14338-global-exportingOptions-data-table

### LINKED ISSUE #14338: "View data table" export option always toggles the 11th button in the list, regardless of actual position
#### Expected behaviour

When using the export-data.js "View data table" option, that option's button text should toggle between "View data table" and "Hide data table".

#### Actual behaviour

The 11th item in the list is the button text that changes, regardless of the actual position of "View data table".

The code to determine `menuItems` in `Chart.prototype.toggleDataTable` does not look at the user-defined `menuItems` in `this.options`. It always gets the default `menuItems` setting, where `viewData` is at index position 10. If you have fewer than 11 options, it throws an error.

http://jsfiddle.net/79jdfvtb/

#### Product version

#### Affected browser(s)
All

karolkolodziej:
Fixed #14338, wrong button was updated after showing the data table.

### DIFF
--- a/ts/Extensions/ExportData.ts
+++ b/ts/Extensions/ExportData.ts
@@ -1259,7 +1259,9 @@ Chart.prototype.hideData = function (): void {
 
 Chart.prototype.toggleDataTable = function (): void {
     var exportDivElements = this.exportDivElements,
-        menuItems = this.options.exporting?.buttons?.contextButton.menuItems,
+        menuItems = this.options.exporting &&
+            this.options.exporting.buttons &&
+            this.options.exporting.buttons.contextButton.menuItems,
         lang = this.options.lang;
 
     if (this.isDataTableVisible) {
--- a/ts/Extensions/ExportData.ts
+++ b/ts/Extensions/ExportData.ts
@@ -1259,9 +1259,8 @@ Chart.prototype.hideData = function (): void {
 
 Chart.prototype.toggleDataTable = function (): void {
     var exportDivElements = this.exportDivElements,
-        menuItems = this.options.exporting &&
-            this.options.exporting.buttons &&
-            this.options.exporting.buttons.contextButton.menuItems,
+        options = this.options.exporting,
+        menuItems = options && options.buttons && options.buttons.contextButton.menuItems,
         lang = this.options.lang;
 
     if (this.isDataTableVisible) {
--- a/ts/Extensions/ExportData.ts
+++ b/ts/Extensions/ExportData.ts
@@ -1259,7 +1259,7 @@ Chart.prototype.hideData = function (): void {
 
 Chart.prototype.toggleDataTable = function (): void {
     var exportDivElements = this.exportDivElements,
-        menuItems = exportingOptions?.buttons?.contextButton.menuItems,
+        menuItems = this.options.exporting?.buttons?.contextButton.menuItems,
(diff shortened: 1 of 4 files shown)


ts/Extensions/ExportData.ts
```diff
@@ -1259,7 +1259,7 @@ Chart.prototype.hideData = function (): void {
 
 Chart.prototype.toggleDataTable = function (): void {
     var exportDivElements = this.exportDivElements,
-        menuItems = exportingOptions?.buttons?.contextButton.menuItems,
+        menuItems = this.options.exporting?.buttons?.contextButton.menuItems,
```
pawelfus, line 1262: Should we cast `this.options` to `any` @bre1470 instaed of using `.?` ?

karolkolodziej: Or I can cast it like this, what do you think?
`menuItems = (this.options.exporting as Highcharts.ExportingOptions).buttons?.contextButton.menuItems,`

bre1470: There should be no forced casts without conditional tests. Also new code should not introduce `any` casts.

IMHO this code should be tested with `&&` instead of optional accessors. Patterns like `something?.else` are bloating the JS output because of `undefined` and `null` tests (to avoid filtering zero and empty string).

ts/Extensions/ExportData.ts
```diff
@@ -1259,7 +1259,9 @@ Chart.prototype.hideData = function (): void {
 
 Chart.prototype.toggleDataTable = function (): void {
     var exportDivElements = this.exportDivElements,
-        menuItems = exportingOptions?.buttons?.contextButton.menuItems,
+        menuItems = this.options.exporting &&
+            this.options.exporting.buttons &&
+            this.options.exporting.buttons.contextButton.menuItems,
```
pawelfus, line 1264: I'm just curious, why `exporting.buttons` must be checked, but `exporting.buttons.contextButton` is always available? Any ideas @karolkolodziej and @bre1470 ?

bre1470: Because the exporting module might not be loaded.

bre1470: Though you are right. It is strange that contextButton is not tested. User could replace buttons with their own options and it would break.

pawelfus: Hm.. I might be still missing something, sorry :) 
- `this.options.exporting` -> can be undefined, when exporting module not loaded
- `this.options.exporting.buttons` -> can be undefined because..?
- `this.options.exporting.buttons.contextButton` -> is always defined because exporting module was loaded

karolkolodziej: I guess it can be undefined because it's an option?
```
interface ExportingOptions {
            allowHTML?: boolean;
            buttons?: ExportingButtonsOptions;
           ...
}
```

bre1470: I would say so too. You could set `buttons` to `undefined`, but the same is true for `contextButton`. So I guess that `contextButton` is an internal type bug.

BTW Maybe consider something like `const options = this.options.exporting` to make the `&&` condition a bit shorter.

TorsteinHonsi (approved):
Thanks!

PR #14354, opened 2020-10-14
bugfix/14166-currentDateIndicator-time

### LINKED ISSUE #14166: currentDateIndicator doesn't show the correct date when time declared per chart
#### Expected behaviour
Show correctly calculated time.

#### Actual behaviour
The `currentDateIndicator` doesn't respect `time` declared in the chart and shows UTC time regardless of options declared.

https://jsfiddle.net/BlackLabel/Lned3pzy/

#### Product version
v8.2.0

#### Affected browser(s)
All

khlieng:
Fixed #14166, `currentDateIndicator` ignored per-chart time options, default label formatter used global `dateFormat`.

### DIFF
--- a/js/Extensions/CurrentDateIndication.js
+++ b/js/Extensions/CurrentDateIndication.js
@@ -10,8 +10,6 @@
  *
  * */
 import Axis from '../Core/Axis/Axis.js';
-import O from '../Core/Options.js';
-var dateFormat = O.dateFormat;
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (changes requested):
Looking good! Please add a test too.

TorsteinHonsi (approved):

PR #14357, opened 2020-10-14
bugfix/13353-handling-passive-events

### LINKED ISSUE #11353: Multi-[Violation] Added non-passive event listener to a scroll-blocking 'touchstart' event ++

raf18seb:
Fixed #11353, added support for passive events.

This PR is a continuation of https://github.com/highcharts/highcharts/pull/13933 (can be closed). I couldn't resolve some issues again so had to install a fresh repo.

### DIFF
--- a/ts/Core/Globals.ts
+++ b/ts/Core/Globals.ts
@@ -66,6 +66,7 @@ declare global {
         const marginNames: Array<string>;
         const noop: () => void;
         const product: string;
+        const isPassiveEvent: boolean;
         const symbolSizes: Dictionary<SizeObject>;
         const win: GlobalWindow;
         const svg: boolean;
@@ -164,7 +165,27 @@ var glob = ( // @todo UMD variable named `window`, and glob named `win`
     hasBidiBug = (
         isFirefox &&
         parseInt(userAgent.split('Firefox/')[1], 10) < 4 // issue #38
-    );
+    ),
+    noop = function (): void {},
+    // Checks whether the browser supports passive events, (#11353).
+    checkPassiveEvents = function (): boolean {
+        let supportsPassive = false;
+
+        // Object.defineProperty doesn't work on IE as well as passive events -
+        // instead of using polyfill, we can exclude IE totally.
+        if (!isMS) {
+            const opts = Object.defineProperty({}, 'passive', {
+                get: function (): void {
+                    supportsPassive = true;
+                }
+            });
+
+            glob.addEventListener('testPassive', noop, opts);
+            glob.removeEventListener('testPassive', noop, opts);
+        }
+
+        return supportsPassive;
+    };
 
 var H: typeof Highcharts = {
     product: 'Highcharts',
@@ -182,11 +203,12 @@ var H: typeof Highcharts = {
     SVG_NS: SVG_NS,
     chartCount: 0,
     seriesTypes: {} as Highcharts.SeriesTypesDictionary,
+    isPassiveEvent: checkPassiveEvents(),
     symbolSizes: {},
     svg: svg,
     win: glob,
     marginNames: ['plotTop', 'marginRight', 'marginBottom', 'plotLeft'],
-    noop: function (): void {},
+    noop: noop,
 
     /**
      * Theme options that should get applied to the chart. In module mode it
--- a/ts/Core/Globals.ts
+++ b/ts/Core/Globals.ts
@@ -66,7 +66,7 @@ declare global {
         const marginNames: Array<string>;
         const noop: () => void;
         const product: string;
-        const isPassiveEvent: boolean;
+        const supportsPassiveEvents: boolean;
         const symbolSizes: Dictionary<SizeObject>;
         const win: GlobalWindow;
         const svg: boolean;
@@ -203,7 +203,7 @@ var H: typeof Highcharts = {
     SVG_NS: SVG_NS,
     chartCount: 0,
     seriesTypes: {} as Highcharts.SeriesTypesDictionary,
-    isPassiveEvent: checkPassiveEvents(),
+    supportsPassiveEvents: checkPassiveEvents(),
     symbolSizes: {},
     svg: svg,
     win: glob,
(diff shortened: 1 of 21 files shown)


TorsteinHonsi:
Thanks!

As [I noted in the original PR](https://github.com/highcharts/highcharts/pull/13933#issuecomment-690227183), I was concerned about the added complexity when allowing an event handler to be a config object instead of a function:

> If the current suggestion works for our internal cases, I'd say we start with that. But obviously we need to check if we break something, by identifying if there are places we are currently calling `e.preventDefault` on touch events.

Did we check this out? Have we concluded that we actually need the added complexity?

raf18seb:
I tested all the cases that I could think of where we are calling `e.preventDefault` and made those events to be `passive: false`.

> Did we check this out? Have we concluded that we actually need the added complexity?
We don't need it internally. Internally, we can decide whether an event should be passive or not before adding an event.
The main reason for adding that complexity is to let the users decide about their events before they're added.

We can get rid of it and wait for users to request, however, it's already implemented - it's your call whether you want to keep it or add it later/never.

TorsteinHonsi:
In my experience it's better to be minimalistic and tight on the features than to add features and complexity that is rarely used. @pawelfus , @oysteinmoseng what do you think?

pawelfus:
Given the popularity of the issue, I would say fix it all

oysteinmoseng:
How much overhead is it to support the event config objects? If it is a lot, I suppose the alternative is to make user-provided touch events passive by default. That should cover most cases without having to add complexity.

TorsteinHonsi:
> How much overhead is it to support the event config objects?

@oysteinmoseng You can read that from the code in this PR, as it is implemented here.

TorsteinHonsi:
_If_ we decide to go all in, we should refactor out the part where we `objectEach` over event options and add either a function event or an event object.

It is currently implemented in [Chart.ts](https://github.com/highcharts/highcharts/pull/14357/files#diff-eec054049d33d5ccd9ad7485a735dfeb5bda93fba3e975bdd69cdff552d400fbR546) and [Interaction.ts](https://github.com/highcharts/highcharts/pull/14357/files#diff-936632e577cd99839dbf2228c424f197aa2d9e1e9ff0531db5fa5f2c3e2fe8a0R1125). 

But should also be applied in [Axis.ts](https://github.com/highcharts/highcharts/blob/d35fb6c/ts/Core/Axis/Axis.ts#L4241-L4245) and possibly [PlotLineOrBand.ts](https://github.com/highcharts/highcharts/blob/d35fb6c/ts/Core/Axis/PlotLineOrBand.ts/#L330-L332)

oysteinmoseng:
If I am reading it correctly that's just 8 or so lines of extra code so far? That being said, there is the refactor mentioned by Torstein, and it looks like it is not added to the API documentation yet, and it is extra complexity to explain to users.

Unless I am missing something, the only value we are adding here is that users can add touch events that call `preventDefault` from the config options. Since we have the pattern of returning `false` to prevent the default Highcharts action, I am not sure how much value there is in the extra `preventDefault`. Except to actually prevent scrolling, I suppose. Users could also perhaps use `addEvent` directly in many cases if they want to control the passive option.

I am leaning towards not implementing this for now.

TorsteinHonsi:
> I am leaning towards not implementing this for now.

Me too... So far, "it may come in handy sometime in the future" has not been a valid reason to add something to Highcharts. 

> Given the popularity of the issue, I would say fix it all

@pawelfus Don't you think that's mainly because of the error message in the console, that will be fixed under any circumstance?

pawelfus:
> Don't you think that's mainly because of the error message in the console, that will be fixed under any circumstance?

Hmm.. good point. I just realized, after Oystein's comment, that only touch events need this change (and a few of `click` events too, because those are translated to touch events sometimes). And those are not that popular afaik. I haven't seen anyone reporting this issue for anything except our internal events.

> If I am reading it correctly that's just 8 or so lines of extra code so far? That being said, there is the refactor mentioned by Torstein, and it looks like it is not added to the API documentation yet, and it is extra complexity to explain to users.

Actually, instead of refactoring, we could change `addEvent(element, eventName, callback, options)` to allow callback or config: `addEvent(element, eventName, { callback, options } | callback, options)`. That would simplify code in the core too. But readability.. Just thinking out loud, nevermind me ;)

TorsteinHonsi (changes requested):
Okay, we need a decision...

I hereby decide that we implement passive events for our internal events, but not for custom events added through chart config. That way we solve the initial reported issue, and we don't overcomplicate our API options structure.

In the future, if frequently requested, we can consider support for passive events through the API options structure.

ts/Core/Globals.ts
```diff
@@ -182,11 +203,12 @@ var H: typeof Highcharts = {
     SVG_NS: SVG_NS,
     chartCount: 0,
     seriesTypes: {} as Highcharts.SeriesTypesDictionary,
+    isPassiveEvent: checkPassiveEvents(),
```
TorsteinHonsi, line 206: Wouldn't a better name here be `supportsPassiveEvents`? I feel that _is_ is not the right word here, because it's not like Highcharts _is_ a passive event.

TorsteinHonsi (changes requested):

ts/Core/Globals.ts
raf18seb: Absolutely right 👍

TorsteinHonsi (approved):
Thanks!

PR #14360, opened 2020-10-15
bugfix/14050-empty-dataLabels

### LINKED ISSUE #14050: Empty `dataLabels` array breaks `errorbar` series
#### Expected behaviour

`errorbar`-type series should be resilient to the `dataLabels` config option being an empty array, like other series types.

#### Actual behaviour

Setting an `errorbar` series' `dataLabels` to `[]` throws an error:

```
TypeError: Cannot read property 'enabled' of undefined
    at e.drawDataLabels (highcharts-more.src.js:1727)
    at e.seriesType.drawDataLabels (highcharts-more.src.js:3985)
    at e.render (highcharts.js:409)
    at highcharts.js:333
    at B.renderSeries (highcharts.js:332)
    at B.render (highcharts.js:335)
    at B.firstRender (highcharts.js:338)
    at B.<anonymous> (highcharts.js:309)
    at g.fireEvent (highcharts.js:33)
```

This error does not occur for other series types like `bar`, `column`, `scatter`, `line` or `area`.

https://jsfiddle.net/5xjezhn6/

#### Product version
Highcharts 8.1.2

#### Affected browser(s)
All

khlieng:
Fixed #14050, empty `dataLabels` array passed to `columnrange` and related series threw error.

### DIFF
--- a/js/Series/AreaRangeSeries.js
+++ b/js/Series/AreaRangeSeries.js
@@ -298,7 +298,7 @@ BaseSeries.seriesType('arearange', 'area', {
                 lowerDataLabelOptions = dataLabelOptions[1];
             }
             else {
-                upperDataLabelOptions = dataLabelOptions[0];
+                upperDataLabelOptions = dataLabelOptions[0] || { enabled: false };
                 lowerDataLabelOptions = { enabled: false };
             }
         }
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (changes requested):
Thanks!

In addition to the fix, we need a unit test to make sure we don't reintroduce the same error in future refactoring (which is a common cause for bugs). In the unit test, we should simplify the demo as much as possible, [for example like this](https://jsfiddle.net/highcharts/c7m0q2zk/).

Here's another case that still fails: https://jsfiddle.net/highcharts/c7m0q2zk/1/

TorsteinHonsi (commented):
Perfect!

TorsteinHonsi (approved):

PR #14366, opened 2020-10-16
bugfix/boost-dx9-shaders

cvasseng:
Fixed #14366, error in boost module on older Windows installations.

Boost: Fixed D9 issue with texture access in conditionals.

This issue manifests itself on older Windows installations where DirectX >9 support is non-existant. The issue is browser independent. Specifically, it fixes this error:
```
[highcharts boost] shader error - C:\fakepath(40,12): error X6077: texld/texldb/texldp/dsx/dsy instructions with r# as source cannot be used inside dynamic conditional 'if' blocks, dynamic conditional subroutine calls, or loop/rep with break*. 
```


TorsteinHonsi (approved):
Thanks!

TorsteinHonsi:
Pending checks

PR #14369, opened 2020-10-16
bugfix/14352-data-table-button-text

### LINKED ISSUE #14352: Improper button text when the data table is initially shown
#### Expected behaviour
Adjust button text to the table visibility no matter how it was enabled.

#### Actual behaviour
When the data table is enabled by the `showTable` property, `toggleDataTable()` or `viewData()` method the button has a wrong text.

Check the button text in the export menu
https://jsfiddle.net/BlackLabel/ef96348y/
https://jsfiddle.net/BlackLabel/0vznf1hp/

#### Product version

#### Affected browser(s)
All

karolkolodziej:
Fixed #14352, wrong button text displayed when chart initialized with the table.


TorsteinHonsi (approved):
Thanks!

PR #14370, opened 2020-10-16
Fixed #14356 `dataURLtoBlob` failed

### LINKED ISSUE #14356: H.dataURLtoBlob - Regex does not accommodate filename=generated.pdf;
#### Expected behaviour

When calling 'getSVGForLocalExport' on a chart, 'dataURLtoBlob' is eventually calls and falls over, when the passed in 'dataURL' argument includes the following string: 'filename=generated.pdf;'

I found another user who was having a similar issue, and was glad to find a workaround in the form of a plugin proposed:

https://github.com/highcharts/highcharts/issues/10243#issuecomment-493015314

Unfortunately this workaround is no longer applicable, and @pawelfus asked me to create this bug.

#### Actual behaviour

dataURLtoBlob fails to convert the dataURL to a blob.

goransle:
Fixed #14356, offline PDF export failed for larger datasets.

This was due to `dataURLtoBlob` failing when the `dataURL` contained the filename.

[With fix applied](http://jsfiddle.net/goransle/0h7o631m/)
[Without](http://jsfiddle.net/goransle/d7v95ahn/)

### DIFF
--- a/js/Extensions/DownloadURL.js
+++ b/js/Extensions/DownloadURL.js
@@ -22,7 +22,9 @@ var win = Highcharts.win, nav = win.navigator, doc = win.document, domurl = win.
  *         Blob
  */
 var dataURLtoBlob = Highcharts.dataURLtoBlob = function (dataURL) {
-    var parts = dataURL.match(/data:([^;]*)(;base64)?,([0-9A-Za-z+/]+)/);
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #14377, opened 2020-10-19
ts/line-column-series

bre1470:
Migrated LineSeries and ColumnSeries to class pattern.

This migration differs in some important aspects of the previous approaches.
1. There is no class split of the LineSeries, so for now it is still the base class.
2. The series classes have only the default constructor without arguments.
3. Non-optional properties are part of the prototype interface, not of the class pattern.

The following technique helps with the series migration. You can move every series function one by one and test in between, if something breaks:
```ts
class SomeSeries extends ParentSeries {
  public static defaultOptions = merge(ParentSeries.defaultOptions, {
    // some options
  });
  // here are the functions after movement
  // you can use super instead of seriesTypes.parent like
  public init(): void {
    super.init.apply(this, arguments as any);
  }
}
interface SomeSeries extends ParentSeries {
  // here are the class definitions taken from Highcharts namespace
}
extend(
  SomeSeries.prototype,
  {
    // here are the functions before movement (like in the seriesType factory)
    // properties from the factory will stay here and not be moved,
    // because they need to be on the prototype (not instance)
  }
);
Series.addSeries('some', SomeSeries);
```


TorsteinHonsi (approved):
Thanks, looks good!

Logically it looks a bit quirky that the LineSeries now is the base series (for example in Legend, where a legend item can either be a `Point|LineSeries`... in reality it is a point or any type of series), but I assume this is temporary.

bre1470:
> Thanks, looks good!
> 
> Logically it looks a bit quirky that the LineSeries now is the base series (for example in Legend, where a legend item can either be a `Point|LineSeries`... in reality it is a point or any type of series), but I assume this is temporary.

Yes, it is temporary and just represents the situation on runtime right now.

TorsteinHonsi (changes requested):
Pending conflict resolution

TorsteinHonsi (approved):

PR #14378, opened 2020-10-20
bugfix/14289-infinite-loop-histogram

### LINKED ISSUE #14289: Infinite loop in histogram caused by series values
When using highchart to render histogram with certain values it ends up in an infinite loop in histogram-bellcurve.js inside the function derived data on this loop 

```
// If binWidth is 0 then max and min are equaled,
                // increment the x with some positive value to quit the loop
                for (x = min; 
                // This condition is needed because of the margin of error while
                // operating on decimal numbers. Without that, additional bin
                // was sometimes noticeable on the graph, because of too small
                // precision of float correction.
                x < max &&
                    (series.userOptions.binWidth ||
                        correctFloat(max - x) >= binWidth ||
...
```

The cause seems to be the value of x never varies with certain values (almost surely because of precision issues)

#### Expected behaviour
That the chart would display.

#### Actual behaviour
Highchart ends in an infinite loop till the browser asks to stop the script or to wait more, as it's an infinite loop the only way out is stopping the script

Just run this
https://jsfiddle.net/ytLhqa8j/1/

By removing a last digit in the value the loop is not happening anymore: https://jsfiddle.net/ytLhqa8j/

```
        "data": [{
            "y": 0.508304338907063,
        }
```

to 
```

        "data": [{
            "y": 0.50830433890706,
        }
```

#### Product version

#### Affected browser(s)
Firefox Chrome last versions

khlieng:
Fixed #14289, histogram series entered an infinite loop when passed certain data values.


Denyllon (approved):
Looks correct to me. Thank you!

TorsteinHonsi (changes requested):
Thanks! Please add a test too.

khlieng:
Would the tests hanging forever be an acceptable way of failing?

Denyllon (approved):

TorsteinHonsi (approved):

PR #14384, opened 2020-10-20
Bugfix/14359 column range series should not stack

DJTechnoo:
Prevent stacking side-effect in columnRange

### DIFF
--- a/ts/Series/ColumnRangeSeries.ts
+++ b/ts/Series/ColumnRangeSeries.ts
@@ -133,25 +133,32 @@ BaseSeries.seriesType<typeof Highcharts.ColumnRangeSeries>('columnrange', 'arear
 ), {
     setOptions: function (this: Highcharts.ColumnRangeSeries): Highcharts.ColumnRangeSeriesOptions {
         var options = arearangeProto.setOptions.apply(this, arguments),
-            dataLabels = options.dataLabels || [];
+            dataLabels = options.dataLabels || [],
+            isInverted = this.chart.inverted;
 
         if (options.stacking) {
             if (!(dataLabels instanceof Array)) {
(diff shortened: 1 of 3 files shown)


ts/Series/ColumnRangeSeries.ts
```diff
@@ -130,6 +134,13 @@ BaseSeries.seriesType<typeof Highcharts.ColumnRangeSeries>('columnrange', 'arear
     (defaultOptions.plotOptions as any).arearange,
     columnRangeOptions
 ), {
+    setOptions: function (this: Highcharts.ColumnRangeSeries): Highcharts.ColumnRangeSeriesOptions {
+        var options = arearangeProto.setOptions.apply(this, arguments);
+        if (options.stacking) {
```
TorsteinHonsi, line 139: Add a little comment with a reference to the issue number, so that future us can understand why we're doing this.

ts/Series/ColumnRangeSeries.ts
```diff
@@ -130,6 +134,13 @@ BaseSeries.seriesType<typeof Highcharts.ColumnRangeSeries>('columnrange', 'arear
...
+            options.stacking = columnRangeOptions.stacking;
```
TorsteinHonsi, line 140: Might as well just delete it, since `columnRangeOptions.stacking` should always be unefined.

TorsteinHonsi (changes requested):

ts/Series/ColumnRangeSeries.ts
DJTechnoo: The `columnRangeOptions.stacking` was still able to be overwritten.  For now this was the workaround, but I will have a look at this to see if there is a better way of doing this.

TorsteinHonsi (approved):
Thanks!

PR #14392, opened 2020-10-21
14389-stockTools-options-type

### LINKED ISSUE #14389: Documentation and Typescript type of stockTools.gui.buttons should be Array<string>
The documentation and typescript types are wrong and specify this property as a string, this was changed between 7.2.1 and 8.2.0

### Wrong
https://api.highcharts.com/highstock/stockTools.gui.buttons
https://unpkg.com/highcharts@8.2.0/highcharts.d.ts

### Correct
https://github.com/highcharts/highcharts/blob/master/ts/Stock/StockToolsGui.ts#L85

bre1470:
Fixed #14389, type of stockTools.gui.buttons option.

### DIFF
--- a/js/Stock/StockToolsGui.js
+++ b/js/Stock/StockToolsGui.js
@@ -184,9 +184,10 @@ setOptions({
             toolbarClassName: 'stocktools-toolbar',
             /**
              * A collection of strings pointing to config options for the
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #14393, opened 2020-10-21
bugfix/timeline-duplicate-events

### LINKED ISSUE #13277: Possible `addEvent` usage with memory leaks
Potential events that are not removed on instance delete/update. List:

- http://utils.highcharts.local/samples/#test/unit-tests/series-timeline/datetime-axis - `Highcharts.charts[1].hcEvents.afterHideOverlappingLabel` - 2x from timeline
- http://utils.highcharts.local/samples/#test/unit-tests/axis/extremes -> Test: `#5493, #5823 - Extremes for xAxis with hidden series and dataGrouping` -> `Highcharts.charts[0].hcEvents.resize` - 6x `setInvert` (!)
- http://utils.highcharts.local/samples/#test/unit-tests/rangeselector/buttons-datagrouping -> `Highcharts.charts[0].hcEvents.destroy` - 9x `destroyExport` (!)
- http://utils.highcharts.local/samples/#test/unit-tests/3d/column-animation -> `Highcharts.charts[0].hcEvents.destroy` - 3x `destroyExport` + 1x other

Needs further debugging to check if those events are expected, or really are memory leaks.

Tested with: (simple example) https://jsfiddle.net/BlackLabel/v4bfpq6L/ - and implemented for all tests

#### Product version
v8.0.4-master

khlieng:
Fixed #13277, event listener leakage in several series and export module.

The same internal listeners were being re-added on every `update()`/`redraw()` and never removed.

### DIFF
--- a/samples/unit-tests/series/seriestypes/demo.js
+++ b/samples/unit-tests/series/seriestypes/demo.js
@@ -73,8 +73,8 @@ QUnit.test('#13277: Event listener memory leak', assert => {
                 data: []
             });
 
-            assert.ok(before === eventCount(chart.series[0]), `No leakage into series.hcEvents for ${type} series`);
-            assert.ok(beforeChart === eventCount(chart), `No leakage into chart.hcEvents for ${type} series`);
+            assert.strictEqual(eventCount(chart.series[0]), before, `${type} update() should not leak into series.hcEvents`);
+            assert.strictEqual(eventCount(chart), beforeChart, `${type} update() should not leak into chart.hcEvents`);
         }
     });
 });
--- a/samples/unit-tests/series/seriestypes/demo.js
+++ b/samples/unit-tests/series/seriestypes/demo.js
@@ -40,3 +40,41 @@ Object.keys(Highcharts.seriesTypes).forEach(function (type) {
         });
     }
 });
+
+QUnit.test('#13277: Event listener memory leak', assert => {
+    Object.keys(Highcharts.seriesTypes).forEach(type => {
+        if (
+            !('linkedTo' in Highcharts.defaultOptions.plotOptions[type]) &&
+            type !== 'scatter3d' &&
+            type !== 'map' && type !== 'mapline' // Transform error on redraw
+        ) {
+            const chart = Highcharts.chart('container', {
+                chart: {
+                    type: type
+                },
+                series: [{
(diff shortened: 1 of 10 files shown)


TorsteinHonsi (changes requested):
Thanks!

- Do we have an issue number to refer to in the PR description?
- We should also have an accompanying unit test (perhaps also think bigger and create a generic test for all `seriesTypes`)

khlieng:
There is no issue for this one.

pawelfus:
This issue: #13277 mentions that bug

khlieng:
Thanks, the `setInvert` case looks to be working as intended, theres one per series + navigator. The `destroyExport` one made me think that it might be a good idea to prevent adding the same function reference twice to the same event in `addEvent`?

pawelfus:
Well, I reported this in April, so probably someone has already fixed some of them :)

> The `destroyExport` one made me think that it might be a good idea to prevent adding the same function reference twice to the same event in `addEvent`?

We can, cool idea. I don't see any real use-case when the same function should be added twice? Of course, except the case when event was previously removed.

On the other hand, we should not call `addEvent()` multiple times with the same object and function. If we do that, then probably we are making a mistake, and the protection in `addEvent()` might hide the real bug.

samples/unit-tests/series/seriestypes/demo.js
```diff
@@ -39,4 +39,42 @@ Object.keys(Highcharts.seriesTypes).forEach(function (type) {
...
+            };
+
+            const before = eventCount(chart.series[0]);
+            const beforeChart = eventCount(chart);
+
+            chart.series[0].update({
+                data: []
+            });
+
+            assert.ok(before === eventCount(chart.series[0]), `No leakage into series.hcEvents for ${type} series`);
```
TorsteinHonsi, line 76: Best practices:
* Use `assert.strictEqual` instead of `assert.ok` and `===`.
* Include the word "should" in the description. This way it is clear what we _expect_ to see, rather than what we actually see, which is wrong when the test fails.

TorsteinHonsi (changes requested):
Nice work 👍  

Just a couple of best-practice adjustments for the assertions, then it's good to merge.

TorsteinHonsi (approved):

PR #14394, opened 2020-10-21
samples/delete-jquery

khlieng:


samples/maps/members/axis-update/demo.js
```diff
@@ -34,23 +34,31 @@ Highcharts.getJSON('https://cdn.jsdelivr.net/gh/highcharts/highcharts@v7.0.0/sam
     });
 });
 
-var blackAndWhite = true,
+let blackAndWhite = true,
     log = true;
 
-$('#update-color').click(function () {
-    var colorAxis = chart.colorAxis[0];
+const updateColor = () => {
```
TorsteinHonsi, line 40: Why a named function instead of an anonymous function like before?

DJTechnoo: At the time it made sense, but I see now that it might look cleaner with an anonymous function instead.

samples/highcharts/chart/renderto-jquery/demo.js
```diff
@@ -1,15 +0,0 @@
-var $container = $('<div>').appendTo(document.body);
```
TorsteinHonsi, line 1: This sample is referred to from the API docs. If we remove the sample, we must remove the reference too: https://github.com/highcharts/highcharts/blob/samples/delete-jquery/ts/Core/Options.ts/#L1754-L1755

TorsteinHonsi (changes requested):
Excellent job! In addition to the removed sample issue, please also check what happened to the diffing visual sample.

TorsteinHonsi (approved):
Perfect, thanks!

PR #14396, opened 2020-10-21
bugfix/13934-xAxis-range-stock-crash

### LINKED ISSUE #13934: Page crashes with memory limit when xAxis intervals are out of range
#### Expected behaviour
Chart should be displayed without any points. 

#### Actual behaviour
When `plotOptions.series.dataGrouping` is enabled, the page crashes. When disabled the chart is rendered.

http://jsfiddle.net/brkxgwq8/

#### Product version
Highstock JS v8.1.2 (2020-06-16)

#### Affected browser(s)
Chrome, Firefox

karolkolodziej:
Fixed #13934, chart crashed when `dataGrouping` was disabled and axis extremes were set outside the data range.

### DIFF
--- a/ts/Core/Axis/Axis.ts
+++ b/ts/Core/Axis/Axis.ts
@@ -6452,7 +6452,7 @@ class Axis implements AxisComposition, AxisLike {
             step,
             bestScore = Number.MAX_VALUE,
             autoRotation: any,
-            range = (this.max as any) - (this.min as any),
+            range = Math.abs((this.max as any) - (this.min as any)),
             // Return the multiple of tickInterval that is needed to avoid
             // collision
             getStep = function (spaceNeeded: number): number {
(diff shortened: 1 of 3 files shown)


ts/Core/Axis/Axis.ts
```diff
@@ -6452,7 +6452,7 @@ class Axis implements AxisComposition, AxisLike {
             step,
             bestScore = Number.MAX_VALUE,
             autoRotation: any,
-            range = (this.max as any) - (this.min as any),
+            range = Math.abs((this.max as any) - (this.min as any)),
```
TorsteinHonsi, line 6455: I'm not sure I understand why a negative value should be turned positive. Isn't the point to keep it above 0 so that we don't enter an eternal loop? In that case, this would be more logically correct (and easier to understand for future us):

`range = Math.max((this.max as any) - (this.min as any), 0),`

TorsteinHonsi (changes requested):

ts/Core/Axis/Axis.ts
karolkolodziej: Yep, you're right, that makes much more sense!

TorsteinHonsi (approved):
Thanks!

PR #14405, opened 2020-10-22
bugfix/14023-numberFormat-returns-wrong-value

### LINKED ISSUE #14023: Value in tooltip not formatted properly
#### Expected behaviour
tooltipOptions.valueDecimals set to 0 should work regardless on number size. It worked in 5.0.12, in 7.0.1 it does not work anymore.

#### Actual behaviour
The tooltip values are not rounded if the number is too small ( 2e-7) and valueDecimals is set to 0. If valueDecimals is set to 1 or 2, the tooltip works properly.

Correct behavior: http://jsfiddle.net/hLprzecq/  
Correct

Incorrect behavior: http://jsfiddle.net/bn7tj1fw/ 
Incorrect

I have created a ticket 7844 for this issue, but instead of fixing the bug I have received a hint that I should create a bug report over here. This is ineffective, the ticket 7844 should be automatically converted to a bug and I should not be forced to retype the whole thing again here in GitHub.

#### Product version
Highcharts 7.0.1 (the bug was introduced in 6.*,I think)

karolkolodziej:
Fixed #14023, for small numbers and decimals argument set to zero, the `numberFormater` returned wrong values.

### DIFF
--- a/js/Core/Utilities.js
+++ b/js/Core/Utilities.js
@@ -1281,7 +1281,7 @@ var timeUnits = H.timeUnits = {
 var numberFormat = H.numberFormat = function numberFormat(number, decimals, decimalPoint, thousandsSep) {
     number = +number || 0;
     decimals = +decimals;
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (approved):
Thanks!

PR #14406, opened 2020-10-22
samples/stock-tools-gui-css

goransle:
Slight modification to css of the `stock-tools-gui` demo, that will mainly result in a nicer thumbnail on the website. Also removed a style that was noted to be outdated.


TorsteinHonsi (approved):
👍

Issue #14130, opened 2020-08-26
Column stacking is broken after chart update with specific options. 


wchmiel:
#### Expected behavior
After chart update stacking should be applied correctly.

#### Actual behaviour
After chart update stacking is broken.

#### Live demo with steps to reproduce
https://jsfiddle.net/BlackLabel/w17ctxbd/

1. Just run the demo
2. Both charts should look the same, as in v8.1.0: https://jsfiddle.net/BlackLabel/e3o4fay9/

#### Product version
Highcharts v8.1.1

Bisected to commit: https://github.com/highcharts/highcharts/commit/57ce06ee444abb7b9c5e962c68dc6f5f831b64ae

robguinness:
I confirm a similar issue when using v8.2.0. The stacked chart is broken on first load (some data missing), and when reloaded, it displays all data correctly. In v8.1.0, everything works fine.

pawelfus:
FYI @TorsteinHonsi 

*Internal note:*
Removed wrong label

Issue #13995, opened 2020-08-04
setSize on dynamically built chart prevents render of second series


anajavi:
#### Expected behaviour
When adding series dynamically all series should show up after setSize is called.

#### Actual behaviour
When setSize is called with:
```js
// any of these will fail
chart.setSize(undefined, 200);
chart.setSize(400, 200);
chart.setSize(400);
chart.setSize(undefined, undefined);
```

Second series on second axis does not render.

#### Live demo with steps to reproduce
https://jsfiddle.net/2mb6vyrn/1/

Second series does not render unless call to setSize is removed.

#### Product version
Highcharts 8.1.1 and 8.1.2
In 8.1.0 this works.

#### Affected browser(s)
Chrome 84

#### Original issue
https://github.com/whawker/react-jsx-highcharts/issues/299#issuecomment-668433074

anajavi:
This is most likely related to #13680, as that fixed setSize for single series.

pawelfus:
Thanks for reporting this issue @anajavi - you are right, it's a bug since v8.1.1

FYI @TorsteinHonsi

PR #14414, opened 2020-10-23
Fixed regressions related to chart.update and axis redraw

TorsteinHonsi:
Fixed regressions with `chart.update` and `chart.setSize`, causing side effects like stacks not updating and series not
showing after adding dynamically. See #13995 and #14130.


pawelfus (approved):
Nice! Thanks :)

PR #14415, opened 2020-10-23
bugfix/14379-scrollablePlotArea-radial-axes

### LINKED ISSUE #14379: scrollablePlotArea- yAxis is not scrolling with the gauge
#### Expected behaviour
Move the gauge yAxis while scrolling.
Workaround: https://jsfiddle.net/BlackLabel/jovu6k21/

#### Actual behaviour
When the scrollablePlotArea is enabled, the gauge yAxis is not scrolling. 

https://jsfiddle.net/BlackLabel/wp3ud9k8/

#### Product version
Highcharts JS v8.2.0

#### Affected browser(s)
All

karolkolodziej:
Fixed #14379, when `scrollablePlotArea` was enabled, gauge axis was not scrolling along.


ts/Extensions/ScrollablePlotArea.ts
```diff
@@ -327,7 +332,23 @@ Chart.prototype.moveFixedElements = function (): void {
...
+ * user scrolls the chart
+ * @private
+ */
+Chart.prototype.moveFixedElements = function (): void {
+    var container = this.container,
+        fixedRenderer = this.fixedRenderer;
+    this.fixedSelectors = this.getFixedElements();
+
+    // Add the possibility to edit the list.
+    fireEvent(this, 'afterGetFixedElements');
```
pawelfus, line 349: 1. `fireEvent(this, 'afterGetFixedElements');` should be called in the method, so if we ever call this method, anywhere again, event will be fired too

2. Instead of defining another object/array in Chart class, could you try this? See:
```js
fireEvent(this, 'afterGetFixedElements',  { fixedSelectors });
return fixedSelectors;
```

Then in Radial axis you could do this:
```js
addEvent(Highcharts.Chart, 'afterGetFixedElements', function (event): void { 
  if (event.fixedSelectors) { /* modify event.fixedSelectors */ } 
});
```

pawelfus (changes requested):
Minor change 👍

karolkolodziej:
I've checked the visual differences and I don't know where they coming from.
In the demo below, one of them- works as expected.
Demo: https://jsfiddle.net/BlackLabel/mrb6tpvq/
Might it be caused by the way the test is performed?

pawelfus:
You didn't load the `highcharts-more` module. Take a look: https://jsfiddle.net/BlackLabel/mrb6tpvq/1/

pawelfus (approved):
Thanks!

TorsteinHonsi (changes requested):
Thanks for the PR!

Actually I will propose another solution - instead of hooking in the radial axis module to postprocess the fixed elements of the scrollable plot area module, I think we should stick with the current simple concept of identifying elements by CSS selectors. 

Currently there is no CSS class revealing that we are dealing with a radial axis, but that is something we should have under any circumstance. This seems to do the trick (based on the current master):

```diff
diff --git a/ts/Core/Axis/Axis.ts b/ts/Core/Axis/Axis.ts
index 8d3c1538d..9ba0c3282 100644
--- a/ts/Core/Axis/Axis.ts
+++ b/ts/Core/Axis/Axis.ts
@@ -6916,10 +6916,12 @@ class Axis implements AxisComposition, AxisLike {
 
         // Create the axisGroup and gridGroup elements on first iteration
         if (!axis.axisGroup) {
+            const radialClassName = this.isRadial ? 'highcharts-radial-axis ' : '';
             axis.gridGroup = renderer.g('grid')
                 .attr({ zIndex: options.gridZIndex || 1 })
                 .addClass(
                     'highcharts-' + this.coll.toLowerCase() + '-grid ' +
+                    radialClassName +
                     (className || '')
                 )
                 .add(axisParent);
@@ -6927,6 +6929,7 @@ class Axis implements AxisComposition, AxisLike {
                 .attr({ zIndex: options.zIndex || 2 })
                 .addClass(
                     'highcharts-' + this.coll.toLowerCase() + ' ' +
+                    radialClassName +
                     (className || '')
                 )
                 .add(axisParent);
@@ -6934,6 +6937,7 @@ class Axis implements AxisComposition, AxisLike {
                 .attr({ zIndex: (labelOptions as any).zIndex || 7 })
                 .addClass(
                     'highcharts-' + axis.coll.toLowerCase() + '-labels ' +
+                    radialClassName +
                     (className || '')
                 )
                 .add(axisParent);
diff --git a/ts/Extensions/ScrollablePlotArea.ts b/ts/Extensions/ScrollablePlotArea.ts
index 75f901f6a..8bfb57286 100644
--- a/ts/Extensions/ScrollablePlotArea.ts
+++ b/ts/Extensions/ScrollablePlotArea.ts
@@ -325,7 +325,12 @@ Chart.prototype.moveFixedElements = function (): void {
         axisClass = '.highcharts-yaxis';
     }
 
-    fixedSelectors.push(axisClass as any, axisClass + '-labels');
+    if (axisClass) {
+        fixedSelectors.push(
+            `${axisClass}:not(.highcharts-radial-axis)`,
+            `${axisClass}-labels:not(.highcharts-radial-axis)`
+        );
+    }
 
     fixedSelectors.forEach(function (className: string): void {
         [].forEach.call(
```

TorsteinHonsi (approved):

PR #14450, opened 2020-10-28
bugfix/14138-misplaced-tooltip-column

### LINKED ISSUE #14138: Misplaced tooltip when yAxis max value set (column chart)
I configured a column chart so that the yAxis has a max value and allowed to draw a tooltip outside of the chart as well. Columns that are cut off because of the max value seem to effect the tooltip position with their original size. 
When displaying negative values in combination with a min value set the same problem occurs.

image

I would expect the tooltip to be positioned on top of the visible column.

Please see https://jsfiddle.net/rarakiroa/zbm92j7k/4/ to reproduce the issue. Hover any of the points exceeding the value of 100.

Izothep:
Fixed #14138, tooltip was displayed outside plotArea for columns when they were exceeding yAxis extremes.

### DIFF
--- a/samples/unit-tests/series-column/tooltip/demo.js
+++ b/samples/unit-tests/series-column/tooltip/demo.js
@@ -53,3 +53,47 @@ QUnit.test("Shared tooltip should compare point.distX, not point.dist to find ab
         "Proper hovered point."
     );
 });
+
+
+QUnit.test("Tooltip should stay inside plotArea even when columns are outside axis extremes.", function (assert) {
+    var chart = new Highcharts.chart('container', {
+        chart: {
+            type: 'column',
+            marginTop: 150
+        },
+        plotOptions: {
+            series: {
+                stacking: true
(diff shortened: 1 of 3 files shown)


samples/unit-tests/series-column/tooltip/demo.js
```diff
@@ -52,4 +52,48 @@ QUnit.test("Shared tooltip should compare point.distX, not point.dist to find ab
...
+
+    var point = chart.series[0].data[0],
+        controller = new TestController(chart);
+
+    controller.mouseOver(
+        chart.plotLeft + point.plotX + 4,
+        chart.plotTop
+    );
+
+    assert.strictEqual(
```
pawelfus, line 93: Could be just `ok()` instead of `strictEqual()`

pawelfus (approved):
Thanks!

TorsteinHonsi (approved):

Issue #14342, opened 2020-10-13
Highmaps: Using symbol as an image lagging while zooming in / out 


YashDhandha:
#### Expected Behavior
While using symbol as an image, symbol of cluster / marker shouldn't be lagging same behind while, number on cluster stays in its place and the same should be with image 

#### Actual Behavior
While we are using symbol as an image, image is lagging behind for a second this is only occurring only when we have used the 
image URL in symbol

#### Live demo with steps to reproduce
https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/maps/marker-clusters/basic 

In above example if we add symbol with image URL, behavior mentioned above will be replicated

#### Product version
Highmaps Version: 8.2.0

PaulDalek:
Hi @YashDhandha

Thank you for reporting this issue. Could you provide an example where this lagging issue occurs? I tried to reproduce it on my side by using an image as a symbol but it works fine.

**Demo**
https://jsfiddle.net/BlackLabel/usfm8oqc/

YashDhandha:
Hi Pawel,

Thanks for reverting.

PFA - video where I have recorded the same jsfiddle you sent.

While zooming in / out, you will be able to see that number of that cluster stays at its place but image lags behind for like a second that you can see in the attached video.

Also, if we don’t use the image, this is not replicating and working as expected.

Hope we can have some workaround on this

Regards,
Yash Dhandha

pawelfus:
**Simplified demo:**
https://jsfiddle.net/BlackLabel/usfm8oqc/1/
Steps:

Click on `+` then on `-` on both charts. Image animation starts from wrong position. It's fine for an SVG marker:

![image](https://user-images.githubusercontent.com/1453926/95969465-0e2c1380-0e0f-11eb-9550-69a9d9fb8e9e.gif)
![marker](https://user-images.githubusercontent.com/1453926/95969475-11bf9a80-0e0f-11eb-9cea-3cdd5f122183.gif)

FYI @wchmiel

YashDhandha:
Hi @wchmiel / @pawelfus do we have any update here?

pawelfus:
Hi @YashDhandha 
Issue was classified as a bug.

PR #14452, opened 2020-10-28
bugfix/14342-cluster-image-wrong-position

### LINKED ISSUE #14343: Highmaps: No states(hover / select) is available to change the symbol image URL
#### Expected Behavior
While using symbol as an image, select / hover states should change symbol image URL when its applied to that specific state using the property of Highmaps

#### Actual Behavior
While using symbol as an image, should be able to change the image URL on hover / select state using symbol property

https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/maps/marker-clusters/basic

Using above URL, if we place the symbol with image URL, while placing states as hover / select on marker and changing the URL in that states, there is no change for symbol and doesn't get updated

#### Product version
Highmaps Version: 8.2.0

wchmiel:
Fixed #14343, cluster marker image animation had wrong position

~~Fixed issue with cluster marker image animation wrong position, see #14342.~~


TorsteinHonsi (approved):
Thanks!

PR #14456, opened 2020-10-29
bugfix/14372-patternfill-opacity 

### LINKED ISSUE #14372: Inactive state not applied to color patterns
#### Expected behaviour
Inactive state properly set on elements that use pattern fill.

#### Actual behaviour
Opacity is not set.

https://jsfiddle.net/BlackLabel/wm2trzsf/5/

Hover over "A" circle, "A&B" circle should be dimmed out, just like with a regular color.

karolkolodziej:
Fixed #14372, unable to set opacity in the pattern.

Works with the `Venn` series, however, the animation is missing... is it acceptable or should I find another solution?
Demo: https://jsfiddle.net/BlackLabel/y6ne1a8x/

### DIFF
--- a/js/Series/VennSeries.js
+++ b/js/Series/VennSeries.js
@@ -868,9 +868,10 @@ var vennSeries = {
         // Return resulting values for the attributes.
         return {
             'fill': color(options.color)
-                .setOpacity(options.opacity)
                 .brighten(options.brightness)
                 .get(),
+            // Set opacity directly to the SVG element, not to pattern #14372.
+            opacity: options.opacity,
             'stroke': options.borderColor,
             'stroke-width': options.borderWidth,
             'dashstyle': options.borderDashStyle
--- a/samples/unit-tests/series-venn/integration/demo.js
+++ b/samples/unit-tests/series-venn/integration/demo.js
@@ -98,3 +98,53 @@ QUnit.module('Options', () => {
         });
     });
 });
+
+QUnit.test('The inactive state should be set to the patterns the same as for colors, #14372.', function (assert) {
+    const chart =  Highcharts.chart('container', {
+        series: [{
+            type: 'venn',
+            data: [{
+                sets: ['A'],
+                value: 2
+            }, {
+                sets: ['B'],
+                value: 2
+            }, {
+                sets: ['A', 'B'],
+                value: 1,
+                name: 'A&B',
+                color: {
+                    pattern: {
+                        path: {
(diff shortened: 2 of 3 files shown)


ts/Extensions/PatternFill.ts
```diff
@@ -658,6 +658,10 @@ addEvent(SVGRenderer, 'complexColor', function (
     // Set the fill/stroke prop on the element
     element.setAttribute(prop, value);
 
+    addEvent(Highcharts.Color, 'setPatternOpacity', function (e): void {
```
khlieng, line 661: This is going to leak event handlers.

TorsteinHonsi (changes requested):
I'm wondering if we're looking at this the wrong way. Instead of adding a hack to allow opacity in patterns, let's look at the pie series, where opacity works:
https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/series/pattern-fill-pie/

This is because the pie series sets the `opacity` attribute directly on the point graphic, which works like so: https://jsfiddle.net/Lb9qoehc/1/

In the pie series' `pointAttribs` function (borrowed from , the `opacity`) attribute is returned. But in the venn series, the `opacity` is applied to the `fill` color (using `setOpacity`) before returned, which is why it fails with patterns.

So I think we should start by looking at why the Venn series does it this way and not like other series.

Edit: Here's the proposed fix applied to the OP: https://jsfiddle.net/highcharts/zuf0sncq/1/

Next, _if we find out Venn has a good reason for doing it differently_, we should find a way to integrate opacity in pattern fills, preferably without tailored event hooks in the Color.ts function.

pawelfus:
> So I think we should start by looking at why the Venn series does it this way and not like other series.

Just to be on the same page: setting `opacity` attribute sets opacity also to border, while setting opacity in `fill` sets opacity only in `fill`. Apart from that, I don't see any other good reason.

TorsteinHonsi (changes requested):
I like the cleanness of 47fd255 better - apply a global opacity regardless of pattern. Three visual samples were diffing because the other series' data labels were no semi-transparent, but that is in accordance with [how other series work](https://jsfiddle.net/highcharts/tef0k4s9/), and should be accepted.

karolkolodziej:
I went back to the previous commit and accepted visual changes with a comment about that change.

TorsteinHonsi (approved):
Thanks!

PR #14461, opened 2020-10-30
bugfix/14444-clip-path-boost-inverted

### LINKED ISSUE #14444: Chart goes outside plot on zoom with boost
#### Expected behaviour
Chart should be redrawn within the plot area

#### Actual behaviour

image

As you can see, on zooming in the bars go out of the plot area. In many cases, they go a lot out distorting other content in the page.

Minimal reproducer: https://jsfiddle.net/v7t9kefn/2/

#### Product version
Highcharts JS v8.2.2 (2020-10-22)

#### Affected browser(s)
Chrome

karolkolodziej:
Fixed #14444, boosted inverted chart had wrong clip-path.

Last changes are a compromise solution because they fixed the issue with inverted charts however the issue with multiple panes is still open and the chart so far it looks the same as before this fix.

### DIFF
--- a/js/Extensions/Boost/BoostOverrides.js
+++ b/js/Extensions/Boost/BoostOverrides.js
@@ -61,10 +61,13 @@ Chart.prototype.getBoostClipRect = function (target) {
     };
     if (target === this) {
         var verticalAxes = this.inverted ? this.xAxis : this.yAxis; // #14444
-        verticalAxes.forEach(function (axis) {
-            clipBox.y = Math.min(axis.pos, clipBox.y);
-            clipBox.height = Math.min(axis.pos - this.plotTop + axis.len, clipBox.height);
-        }, this);
+        if (verticalAxes.length <= 1) {
+            clipBox.y = Math.min(verticalAxes[0].pos, clipBox.y);
+            clipBox.height = verticalAxes[0].pos - this.plotTop + verticalAxes[0].len;
+        }
+        else {
+            clipBox.height = this.plotHeight;
+        }
     }
     return clipBox;
 };
--- a/js/Extensions/Boost/BoostOverrides.js
+++ b/js/Extensions/Boost/BoostOverrides.js
@@ -60,16 +60,11 @@ Chart.prototype.getBoostClipRect = function (target) {
         height: this.plotHeight
     };
     if (target === this) {
-        if (!this.inverted) {
-            this.yAxis.forEach(function (yAxis) {
-                clipBox.y = Math.min(yAxis.pos, clipBox.y);
-                clipBox.height = Math.max(yAxis.pos - this.plotTop + yAxis.len, clipBox.height);
-            }, this);
-        }
-        else { // #14444
-            clipBox.height = this.plotHeight;
-            clipBox.y = this.plotTop;
-        }
+        var verticalAxes = this.inverted ? this.xAxis : this.yAxis; // #14444
+        verticalAxes.forEach(function (axis) {
+            clipBox.y = Math.min(axis.pos, clipBox.y);
+            clipBox.height = Math.min(axis.pos - this.plotTop + axis.len, clipBox.height);
+        }, this);
     }
     return clipBox;
 };
--- a/js/Extensions/Boost/BoostOverrides.js
+++ b/js/Extensions/Boost/BoostOverrides.js
@@ -60,10 +60,16 @@ Chart.prototype.getBoostClipRect = function (target) {
         height: this.plotHeight
     };
     if (target === this) {
-        this.yAxis.forEach(function (yAxis) {
-            clipBox.y = Math.min(yAxis.pos, clipBox.y);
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (changes requested):
It seems to me like the existing code takes into account that there may be multiple axes with different positions, but that the new code ignores this. See https://jsfiddle.net/m4hzrkgu/.

I assume something like this would work, but haven't tested it:
```ts

    if (target === this) {
        const verticalAxes = this.inverted ? this.xAxis : this.yAxis; // #14444
        verticalAxes.forEach(function (axis): void {
            clipBox.y = Math.min(axis.pos, clipBox.y);
            clipBox.height = Math.max(
                axis.pos - this.plotTop + axis.len,
                clipBox.height
            );
        }, this);
    }
```

karolkolodziej:
After a slight modification of your code (changed `Math.max` to `Math.min` to properly calculate the size of the clip-path and takes into account the possibility of manually positioning the axis).
However, it looks like the tooltip is not working correctly and (which might be connected) we still can zoom below the plot area.
**Demo:** http://jsfiddle.net/BlackLabel/18jLcy02/

Should I create a separate ticket for that issue?

TorsteinHonsi (changes requested):
> After a slight modification of your code (changed Math.max to Math.min to properly calculate the size of the clip-path and takes into account the possibility of manually positioning the axis).

I think that change effectively stops us from using multiple panes: https://jsfiddle.net/highcharts/49cahqom/ . Compare that to the current official release.

> Should I create a separate ticket for that issue?

Yes, I think it's best with a separate ticket if the issue is not related.

karolkolodziej:
After further investigation, I see that the multiple panes don't work well with the current release. 
Demo: http://jsfiddle.net/BlackLabel/d0ayob6n/

Current code in boost module (creating one clip-path for the whole chart) isn't particularly well optimalized. The `getBoostClipRect` method is somehow invoked for each series and always returns the same clip. Changing that might save some resources for the charts reach with series.

However multiple panes work properly in the Stock (without boost) chart where the clip-path is calculated for each series individually, not for the whole chart. 
Demo: http://jsfiddle.net/BlackLabel/uwy4n06t/
https://github.com/highcharts/highcharts/blob/72babf379e387c189b3af4b8b471c166c81c5250/ts/Core/Chart/StockChart.ts#L1026

@TorsteinHonsi, @pawelfus   I would appreciate it if you could throw some light on it and point in which direction the code should go.

TorsteinHonsi:
Yes, boost generalizes the clip rectangle for the whole set of series. I am not sure why it is done that way - I imagine it would be just as simple to apply individual clip rectangles for each series, based on their own axes' positions.

It may have to do with [boost.seriesThreshold](https://api.highcharts.com/highcharts/boost.seriesThreshold) though. Maybe we need to one combined clip area when that threshold is passed.

Feel free to investigate this.

PS: Here is the demo with graphs bleeding out of their panes: http://jsfiddle.net/highcharts/yqjbe0L7/

pawelfus:
> It may have to do with boost.seriesThreshold though. Maybe we need to one combined clip area when that threshold is passed.

I would say the same. Boost can either create one canvas per series, or use just one, shared across all series. This is related to boosting dozens of series and limited number of webgl contexts (16). It's strange, but it seems we always use one canvas, no matter the reason (boosting because of number of series or number of points): http://jsfiddle.net/BlackLabel/p3xy9mLh/ - but it's unrelated to the issue.

In theory we can go down this route: create one canvas per pane, but practically 16+ panes is not that surprising (e.g. real-time data dashboard with scrollable plot area). However:

> I would appreciate it if you could throw some light on it and point in which direction the code should go.

I would focus on the inverted-clip bug here, not on related multiple-panes issue. Could you show a demo where solution suggested by Torstein doesn't work? Thanks! :)

karolkolodziej:
> Could you show a demo where solution suggested by Torstein doesn't work?
>
This solution fails while the axis is positioned manually because of the code below.
`clipBox.height = Math.max(yAxis.pos - this.plotTop + yAxis.len, clipBox.height);`
In this case `clipBox.height  = chart.plotHeight` is always bigger than any axis and automatically the clip box is going to be the size of the plot area.
But in case of this issue (standard axis options only inverted chart), it works like a charm.
Demo: https://jsfiddle.net/BlackLabel/ux5r6tdj/

Switching `max` to `min` fixes that, however this time it fails with multiple panes (last commit). 
`Chart.prototype.getBoostClipRect` method sets one clipPath to the whole chart not for each series/pane.
Demo: https://jsfiddle.net/BlackLabel/o9Lkfhnc/

I guess we can leave the Torstein solution for this bug and create a separate issue for the other connected elements and solve them together after some discussion which way to choose.

pawelfus:
Okay, in inverted chart, we should use `plotWidth` instead of `plotHeight`: https://jsfiddle.net/BlackLabel/ux5r6tdj/1/

TorsteinHonsi (approved):
Looks good to me, thanks!

PR #14469, opened 2020-11-02
docs/14196-fix-sunburst-level

### LINKED ISSUE #14196: Typescript declarations: Sunburst chart series levels
#### Expected behaviour
The type _PlotSunburstLevelsOptions_ is missing the _'level?: number'_ property in highcharts 8. I assumed that this property was not necessary anymore. However, the _'levelSize'_ property among others are ignored without _'level'_ defined. Either add _'level?: number'_ back to _PlotSunburstLevelsOptions_ or make it work without it (treat the first _PlotSunburstLevelsOptions_ in the _'levels'_ array as level 1 and so on...)

#### Actual behaviour
The _PlotSunburstLevelsOptions[]_ typed levels defined in the sunburst options series don't work without a defined _'level'_ property

Sunburst levelSize example:
https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/plotoptions/sunburst-levelsize/

The example doesn't work anymore if you remove the _'level'_ properties from the _PlotSunburstLevelsOptions_ object defined in _'levels'_:
https://jsfiddle.net/xtfbzn4o/

#### Product version
highcharts 8.2.0

#### Affected browser(s)
Chrome

bre1470:
Fixed #14196, added level doclet to sunburst options.

### DIFF
--- a/js/Series/SunburstSeries.js
+++ b/js/Series/SunburstSeries.js
@@ -447,6 +447,16 @@ var sunburstOptions = {
      * @apioption plotOptions.sunburst.levels.dataLabels
      */
     /**
+     * Decides which level takes effect from the options set in the levels
+     * object.
+     *
+     * @sample highcharts/demo/sunburst
+     *         Sunburst chart
+     *
+     * @type      {number}
+     * @apioption plotOptions.sunburst.levels.level
+     */
+    /**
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #14471, opened 2020-11-02
bugfix/13981-marker-cluster-empty-data

### LINKED ISSUE #13981: Marker Cluster module with xAxis min/max undefined and data set to empty array - chart does not display
#### Expected behaviour
The chart will show with no data until the series data is updated.

#### Actual behaviour
Nothing renders. Seems like the module gets stuck in an infinite loop.

https://codepen.io/codywsears/pen/jOWgBJy

#### Product version
Highcharts v8.1.2

#### Affected browser(s)
Found on chrome

wchmiel:
Fixed #13981, no map displayed when using marker clusters and empty data array.

~~Fixed issue with marker clusters empty data array, see #13981.~~

### DIFF
--- a/js/Extensions/MarkerClusters.js
+++ b/js/Extensions/MarkerClusters.js
@@ -1291,7 +1291,8 @@ Scatter.prototype.generatePoints = function () {
         clusterOptions.enabled &&
         series.xData &&
         series.yData &&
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (approved):
Thanks!

PR #14472, opened 2020-11-02
bugfix/14417-incorrect-min-max-behaviour

### LINKED ISSUE #14417: yAxis min property and startOnTick/endOnTick not working properly
#### Expected behaviour

When I use startOnTick and endOnTick as false, and min property in yAxis, if that min value is greater than all the values of the points then chart should be empty (it shouldn't draw any point)

#### Actual behaviour

When I use startOnTick and endOnTick as false and a min property in yAxis, if the min value is greater than the values of the points, it actually draws some points (that are lower than the min value). Also, it seems that the chart is inverted

https://jsfiddle.net/ladet/0k45zcdb/2/

#### Product version
Highcharts 8.1.2 with Highcharts-react

**Note**: I need both startOnTick and endOnTick as false, and it shouldn't be drawing any points when the min is greater than the points. If I set startOnTick and endOnTick as true, this doesn't happen. If I set a max property it works correctly, but since both max and min are set dynamically by the user, it should be possible to only select a minimum limit and it should be working properly whether the values are lower than said minimum or higher.

image

This also happens with columns charts but is worse since it draws a full column when it shouldn't draw any column
image

DJTechnoo:
Fixed #14417, charts displayed parts of the data when min or max exceeded extreme values.

~~Fixed #14417 charts displaying parts when min or max exceed extreme values.~~

### DIFF
--- a/ts/Core/Axis/Axis.ts
+++ b/ts/Core/Axis/Axis.ts
@@ -5485,6 +5485,18 @@ class Axis implements AxisComposition, AxisLike {
             }
         }
 
+        // If min is bigger than highest,
+        // or if max less than lowest value,
+        // the chart should not render points. (#14417)
+        if (axis.min !== null && axis.max !== null) {
+            if (axis.min > axis.max) {
+                if (defined(axis.options.min)) {
+                    axis.max = axis.min;
+                } else if (defined(axis.options.max)) {
+                    axis.min = axis.max;
+                }
+            }
+        }
 
         // get tickInterval
         if (
@@ -5791,19 +5803,6 @@ class Axis implements AxisComposition, AxisLike {
                 while ((this.max as any) + minPointOffset <
                         tickPositions[tickPositions.length - 1]) {
                     tickPositions.pop();
-                }
-            }
-
-            // If min is bigger than highest,
-            // or if max less than lowest value,
-            // the chart should not render points. (#14417)
-            if (this.min && this.max) {
-                if (this.min > this.max) {
-                    if (typeof this.options.min !== 'undefined' && this.options.min !== null) {
-                        this.max = this.min;
-                    } else if (typeof this.options.max !== 'undefined' && this.options.max !== null) {
-                        this.min = this.max;
-                    }
                 }
             }
 
--- a/ts/Core/Axis/Axis.ts
+++ b/ts/Core/Axis/Axis.ts
@@ -4743,7 +4743,7 @@ class Axis implements AxisComposition, AxisLike {
     }
 
     /**
-     * Internal function to et the tick positions of a linear axis to round
+     * Internal function to get the tick positions of a linear axis to round
      * values like whole tens or every five.
      *
      * @function Highcharts.Axis#getLinearTickPositions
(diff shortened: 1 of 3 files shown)


ts/Core/Axis/Axis.ts
```diff
@@ -5794,6 +5794,19 @@ class Axis implements AxisComposition, AxisLike {
                 }
             }
 
+            // If min is bigger than highest,
+            // or if max less than lowest value,
+            // the chart should not render points. (#14417)
+            if (this.min && this.max) {
+                if (this.min > this.max) {
+                    if (typeof this.options.min !== 'undefined' && this.options.min !== null) {
```
pawelfus, line 5802: This looks now exactly as `Highcharts.defined()` implementation

ts/Core/Axis/Axis.ts
```diff
@@ -5794,6 +5794,19 @@ class Axis implements AxisComposition, AxisLike {
                 }
             }
 
+            // If min is bigger than highest,
+            // or if max less than lowest value,
+            // the chart should not render points. (#14417)
+            if (this.min && this.max) {
```
pawelfus, line 5800: This probably will fail when `this.min = 0` and `this.max = -1`?

pawelfus (commented):
I know the solution here is based on my workaround, but I used `trimTicks` in the workaround because that was the easiest place to inject some code. Instead it might be better to fix issue where it happens (where `axis.min` and `axis.max` are being defined) - check `setTickInterval` method.

DJTechnoo:
Thank you for the review Pawel! Yes you are right, I shall have one more go at it, thank you for advice and feedback!

ts/Core/Axis/Axis.ts
```diff
@@ -5485,6 +5485,18 @@ class Axis implements AxisComposition, AxisLike {
             }
         }
 
+        // If min is bigger than highest,
+        // or if max less than lowest value,
+        // the chart should not render points. (#14417)
```
TorsteinHonsi, line 5490: Let the comment break at line-width 80, so it looks less random.

ts/Core/Axis/Axis.ts
```diff
@@ -5485,6 +5485,18 @@ class Axis implements AxisComposition, AxisLike {
...
+        if (axis.min !== null && axis.max !== null) {
+            if (axis.min > axis.max) {
```
TorsteinHonsi, line 5492: These two conditions can be combined into one.

TorsteinHonsi (changes requested):
Thanks! In addition to the two inline comments, please also [make the PR ready for the changelog](https://github.com/highcharts/highcharts/tree/master/changelog#writing-content-for-the-changelog).

ts/Core/Axis/Axis.ts
DJTechnoo: Thank you Torstein, will do the changes and prepare the PR after resolving the failing tests that are occuring.

TorsteinHonsi (approved):
Thanks!

PR #14473, opened 2020-11-02
bugfix/14434-stickOnContact-outside

### LINKED ISSUE #14434: Tooltip `stickOnContact` has confilct with `outside`
#### Expected behaviour

Tooltip should not change while mouse pointer is on it.

#### Actual behaviour

Tooltip is closed

https://jsfiddle.net/fv5qkr4c/

#### Product version
v8.2.2

#### Affected browser(s)
Tested on Chrome 86.0.4240.111

khlieng:
Fixed #14434, `tooltip.outside` broke `tooltip.stickOnContact`.

### DIFF
--- a/ts/Core/Pointer.ts
+++ b/ts/Core/Pointer.ts
@@ -1233,6 +1233,14 @@ class Pointer {
         const chart = charts[pick(H.hoverChartIndex, -1)];
         const tooltip = this.chart.tooltip;
 
+        if (tooltip && tooltip.container) { // #14434
+            const rect = tooltip.container.getBoundingClientRect();
+            if (e.clientX >= rect.left && e.clientX <= rect.right &&
+                e.clientY >= rect.top && e.clientY <= rect.bottom) {
+                return;
+            }
+        }
+
         e = this.normalize(e);
 
         // #4886, MS Touch end fires mouseleave but with no related target
--- a/ts/Core/Tooltip.ts
+++ b/ts/Core/Tooltip.ts
@@ -503,6 +503,7 @@ class Tooltip {
         if (this.renderer) {
             this.renderer = this.renderer.destroy() as any;
             discardElement(this.container as any);
+            this.container = void 0 as any;
         }
         U.clearTimeout(this.hideTimer as any);
         U.clearTimeout(this.tooltipTimeout as any);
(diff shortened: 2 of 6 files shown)


ts/Core/Pointer.ts
```diff
@@ -1233,6 +1233,14 @@ class Pointer {
...
         const tooltip = this.chart.tooltip;
 
+        if (tooltip && tooltip.container) { // #14434
+            const rect = tooltip.container.getBoundingClientRect();
+            if (e.clientX >= rect.left && e.clientX <= rect.right &&
+                e.clientY >= rect.top && e.clientY <= rect.bottom) {
+                return;
+            }
+        }
+
```
TorsteinHonsi, line 1243: I was hoping for a simpler check, something like this:
```js
if (tooltip && tooltip.isStickyOnContact()) {
    return;
}
```

But it doesn't work. The internal `tooltip.inContact` flag is false, that should work here. Can you have a look at why that doesn't work?

ts/Core/Tooltip.ts
```diff
@@ -503,6 +503,7 @@ class Tooltip {
         if (this.renderer) {
             this.renderer = this.renderer.destroy() as any;
             discardElement(this.container as any);
+            this.container = void 0 as any;
```
TorsteinHonsi, line 506: Looks like `container` is wrongly typed when we have to cast `void 0` to any. The type should be `HTMLElement|undefined`. But we may not need this change if we get the above to work.

TorsteinHonsi (changes requested):

ts/Core/Pointer.ts
khlieng: I think its because the tooltip label is smaller than the container, and the handler that sets inContact is attached to the label.

khlieng: Or its not really about size since the charts `onContainerMouseLeave` is going to fire first anyway.

TorsteinHonsi: Hmm. In that case we can use something like `if (this.inClass(relatedTarget, 'highcharts-tooltip-container'))` => return.

@bre1470 Any suggestions on how we can avoid having double logic to detect whether to stick on contact?

khlieng: Using `setTimeout(,0)` in `onContainerMouseLeave` could work.

khlieng: That sets `pointerEvents` to `auto` when `stickOnContact` is `true`, otherwise the handlers that set `inContact` wouldnt fire at all. And its probably `none` when `stickOnContact` is `false` so that you can click through the tooltip.

PR #14475, opened 2020-11-02
bugfix/14433-dependency-wheel-datalabels

### LINKED ISSUE #14433: Dependency wheel: Datalabels are not place properly if there are only 2 nodes.
#### Expected behaviour
Datalabels should place as per 'Brazil' in given example

#### Actual behaviour
Datalabel 'Portugal' placed inside the circle

https://jsfiddle.net/sgzdt758/

khlieng:
Fixed #14433, dependency wheel data labels had wrong position for items covering more than 180 degrees of the wheel.


TorsteinHonsi (approved):
Thanks!

PR #14476, opened 2020-11-02
enhancement/3d-area-series

Izothep:
Enhancement, added support for area series in 3d charts.


ts/Core/Renderer/SVG/SVGElement.ts
```diff
@@ -223,7 +223,7 @@ declare global {
             ): void;
             public xGetter(key: string): (number|string|null);
             public yGetter(key: string): (number|string|null);
-            public zIndexSetter(value: number, key: string): boolean;
+            public zIndexSetter(value: number, key?: string): boolean;
```
Izothep, line 226: It looks that changing key to optional value is not needed with current solution.

ts/Core/Series/Series3D.ts
```diff
@@ -51,14 +53,22 @@ addEvent(LineSeries, 'afterTranslate', function (): void {
...
         projectedPoint: Position3DObject,
         zValue: (number|null|undefined),
-        i: number;
+        i: number,
+        stack = seriesOptions.stacking ?
+            (seriesOptions.stack || 0) :
+            series.index, // #4743
+        rawPointsX = [] as Array<number>;
+
+    series.zPadding = (stack as any) *
```
Izothep, line 70: Instead, we can use:
```
stack = seriesOptions.stacking ?
            (H.isNumber(seriesOptions.stack) ? seriesOptions.stack : 0) :
            series.index || 0
```

So the casting won't be needed.

ts/Core/Series/Series3D.ts
```diff
@@ -73,18 +83,23 @@ LineSeries.prototype.translate3dPoints = function (): void {
...
-            rawPoint.plotZ = 0;
+            // add value of zPadding to final z position of calculated point.
+            rawPoint.plotZ = series.zPadding;
         }
 
         rawPoint.axisXpos = rawPoint.plotX;
         rawPoint.axisYpos = rawPoint.plotY;
         rawPoint.axisZpos = rawPoint.plotZ;
 
+        rawPointsX.push(rawPoint.plotX as any);
```
Izothep, line 94: `rawPointsX.push(rawPoint.plotX || 0);` will solve the casting

PR #14480, opened 2020-11-03
bugfix/14428-update-point-drilldown

### LINKED ISSUE #14428: Enabling drilldown on existing series does not initially work
#### Expected behaviour
Drilldown columns are clickable from the start

#### Actual behaviour
Drilldown columns are clickable only after going one level deeper and then coming back to start.
As you see in my example, clicking on chart columns does not work.
Initially you are only able to click on xAxis ticks (years), then you click "Back" button and only then you can click on the columns.

https://codepen.io/SoWhat/pen/RwRoMQq

#### Product version
All versions up to 8.0.0. When testing with newer versions, my chart did not work at all.

#### Affected browser(s)
All

I think this is a bug, because the point.update({ drilldown: true }, false); function only changes cursor to pointer, but does not actually enable drilldown initially.

khlieng:
Fixed #14428, enabling `drilldown` through `Point.update()` did not work.


TorsteinHonsi (approved):
Nice, thanks!

PR #14482, opened 2020-11-03
bugfix/14426-vertical-pan-zoom

### LINKED ISSUE #14426: Vertical panning after zooming
#### Expected behaviour
Pan on the vertical axis after zoom to `0`.

#### Actual behaviour
After zooming, new min equal the minimal data is set on the y-axis.
Unable to pan on the vertical axis below the `1.2`.

Zoom in, then try to pan below `1.2` on the y-axis.
https://jsfiddle.net/BlackLabel/v27u1tks/

#### Product version
Highcharts JS v8.2.2 (Vertical panning available from 8.1.0)

#### Affected browser(s)
All

khlieng:
Fixed #14426, it was not possible to pan vertically to 0 after zooming.

### DIFF
--- a/samples/unit-tests/interaction/panning/demo.js
+++ b/samples/unit-tests/interaction/panning/demo.js
@@ -1,5 +1,5 @@
 QUnit.test('#14426: Vertical panning after zooming', assert => {
-    const chart = Highcharts.chart('container', {
+    let chart = Highcharts.chart('container', {
         chart: {
             type: 'column',
             zoomType: 'xy',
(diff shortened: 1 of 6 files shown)


samples/unit-tests/interaction/panning/demo.js
```diff
@@ -0,0 +1,64 @@
...
+
+    const offset = Highcharts.offset(chart.container);
+
+    // Zoom
+    chart.pointer.onContainerMouseDown({
+        type: 'mousedown',
+        pageX: 250 + offset.left,
+        pageY: 50 + offset.top,
+        target: chart.container
+    });
```
pawelfus, line 37: FYI: Instead of playing with `offset` you can use `TestController` :)

khlieng: I got "inspired" by the other panning tests :)

ts/Core/Interaction.ts
```diff
@@ -853,7 +853,9 @@ extend(Chart.prototype, /** @lends Chart.prototype */ {
                             isNumber(dataExtremes.dataMax)
                         ) {
                             panningState.startMin = Math.min(
-                                dataExtremes.dataMin, panningState.startMin
+                                pick(series.options.threshold, Infinity),
```
TorsteinHonsi, line 856: The handling of `startMin` and `startMax` was previously equivalent, but now `startMin` also takes the `threshold` into account. Gut feeling says they should still be equivalent, and the fail case is when all values are negative, so that the `threshold` is the max value. Try to zoom in, then we should expect to be able to pan up to 0: https://jsfiddle.net/highcharts/zupgLn59/

TorsteinHonsi (changes requested):

ts/Core/Interaction.ts
```diff
@@ -853,10 +853,14 @@ extend(Chart.prototype, /** @lends Chart.prototype */ {
...
+                                dataExtremes.dataMin,
+                                panningState.startMin
                             );
                             panningState.startMax = Math.max(
-                                dataExtremes.dataMax, panningState.startMax
+                                pick(series.options.threshold, Infinity),
```
TorsteinHonsi, line 861: This means that when `series.options.threshold` is `null` or undefined, `startMax` becomes `Infinity`. And the consequence can be see here: https://jsfiddle.net/highcharts/zupgLn59/1/

I think it should be `pick(series.options.threshold, -Infinity)`

TorsteinHonsi (changes requested):

TorsteinHonsi (approved):

PR #14483, opened 2020-11-03
ts/pie-series

bre1470:


ts/Series/Pie/PiePoint.ts
```diff
@@ -0,0 +1,430 @@
...
+ * */
+
+interface PiePoint {
+    connectorShapes: Record<string, PiePoint.ConnectorShapeFunction>;
+    labelDistance: number;
+    name: string;
+    options: PiePointOptions;
+    series: PieSeries;
+}
+extend(PiePoint.prototype, {
```
TorsteinHonsi, line 301: Why do we do extend in this strange manner instead of just putting it in the class?

TorsteinHonsi (changes requested):

ts/Series/Pie/PiePoint.ts
bre1470: In the class they need a default value in the constructor. A later `init` call by chart is not sufficient.

bre1470: Also to keep compatibility, the properties in the `extend` call continue to be set on the prototype. In a class they would be set for each instance separately, which can break code, fiddles, tests, and wraps.

bre1470: I will improve the situation with `any` casts. It has to be refactored anyway later on.

TorsteinHonsi (approved):

PR #14484, opened 2020-11-04
Fixed #14260, series color affected by marker.fillColor in Boost

### LINKED ISSUE #14260: Line color same as marker fill with large datasets
#### Expected behaviour
Line color should be default and only marker fill should change. 

#### Actual behaviour
Line color is same as marker color due to which line disappears when `marker.fillColor: '#ffffff'`.

https://codepen.io/govinda-totla/pen/JjXxbPa

Changing the number of points to 1000 fixes the issue.

#### Product version

#### Affected browser(s)
Chrome

TorsteinHonsi:
Fixed #14260, series graph color was affected by `marker.fillColor` in Boost

### DIFF
--- a/js/Extensions/Boost/WGLRenderer.js
+++ b/js/Extensions/Boost/WGLRenderer.js
@@ -831,7 +831,9 @@ function GLRenderer(postRenderCallback) {
             }
             else {
                 fillColor =
-                    (s.series.pointAttribs && s.series.pointAttribs().fill) ||
+                    (s.drawMode === 'points' && // #14260
+                        s.series.pointAttribs &&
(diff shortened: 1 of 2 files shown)


KacperMadej:
Hi @viralBot 

The best option for older code version is to upgrade to the up to date one.
The second best we could do, for an older version, is a workaround - it is possible to have 2 series to control series plot and markers separately, like in this live example: https://jsfiddle.net/BlackLabel/c7oawd9s/

In general, for technical support questions please contact through Technical Support channels relevant to your questions and license-wise availability - https://www.highcharts.com/support/

PR #14486, opened 2020-11-04
bugfix/14078-boost-preallocated

### LINKED ISSUE #14078: Boosted chart with `usePreallocated: true` causes all series points to appear at the same location
#### Expected behaviour
Each point should appear in its provided coordinates
image

#### Actual behaviour
With `boost.usePreallocated: true`, all series' points overlap
image

https://codepen.io/burtyish/pen/jOqrKax?editors=1011

#### Product version
8.1

#### Affected browser(s)
Chrome

khlieng:
Fixed #14078, all points rendered in all series with `boost.usePreallocated` on.


TorsteinHonsi (approved):
Thanks!

PR #14494, opened 2020-11-05
bugfix/14397-networkgraph-update

### LINKED ISSUE #14397: Updating networkgraph series causes simulation to start from the beginning
#### Expected behaviour
`series.update({ dataLabels })` should keep existing positions instead of starting the simulation from the scratch.

#### Actual behaviour
`series.update({ dataLabels })` starts simulation like when creating a chart.

https://jsfiddle.net/highcharts/18vrf9su/2/

*Internal note:*
Might be connected to the way Series is updated: by removing old one and initialising new one. Perhaps `points` (links) are preserved, but `nodes` are not.

khlieng:
Fixed #14397, `networkgraph` simulation restarted on `series.update()`.

### DIFF
--- a/js/Core/Dynamics.js
+++ b/js/Core/Dynamics.js
@@ -1084,6 +1084,8 @@ extend(LineSeries.prototype, /** @lends Series.prototype */ {
         animation = series.finishedAnimating && { animation: false }, kinds = {};
         if (keepPoints) {
             preserve.push('data', 'isDirtyData', 'points', 'processedXData', 'processedYData', 'xIncrement', 'cropped', '_hasPointMarkers', '_hasPointLabels', 
+            // Networkgraph (#14397)
+            'nodes', 'layout', 
             // Map specific, consider moving it to series-specific preserve-
             // properties (#10617)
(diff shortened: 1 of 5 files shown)


pawelfus (approved):
Looks good to me, thanks!
Changes in `Dynamic.ts` up to @TorsteinHonsi - probably it could be a nice idea to let series specific type provide props that should be stored. No strong opinion here, though.

pawelfus:
Just realized - karma + chrome headless do not have `requestAnimationFrame`, so is this test actually testing anything?

khlieng:
The test is just testing if nodes and layout are preserved on update, so that should be fine?

pawelfus:
Good point 👍  Then `enableSimulation` can be disabled (default in karma tests).

TorsteinHonsi (approved):
Looks good!

TorsteinHonsi:
> probably it could be a nice idea to let series, of a specific type, provide props that should be stored. No strong opinion here, though.

Yes, absolutely. But we can come back to that if we find the list hard to maintain.

PR #14496, opened 2020-11-05
bugfix/14458-3d-drilldown-same-name

### LINKED ISSUE #14458: Drilldown to point of same name as parent column on 3d column chart causes error
#### Expected behaviour
Drilldown when series has 1 point with same name as parent point should work.

#### Actual behaviour
Drilldown when series has 1 point with same name as parent point produces javascript error.
Axis3D.wrapGetSlotWidth - nextTick.label is undefined

http://jsfiddle.net/marklofdahl/re1fjh6y/
Click on "Museums" to drilldown - causes javascript error

#### Product version
Highcharts 8.2

#### Affected browser(s)
all

If the point name is not the same as the parent point, then it does not error.

I worked around it by setting nextTick.label...

    /* wrap Axis.getSlotWidth for 3D, to fix missing label bug */
    Highcharts.wrap(Highcharts.Axis.prototype, 'getSlotWidth', function (proceed, tick) {
        var axis = this;
        var chart = axis.chart;
        var ticks = axis.ticks;
        var gridGroup = axis.gridGroup;
        var width;

        if (axis.categories &&
            chart.frameShapes &&
            chart.is3d() &&
            gridGroup &&
            tick &&
            tick.label) {

            var tickId = tick.pos,
                prevTick = ticks[tickId - 1],
                nextTick = ticks[tickId + 1],
                prevLabel,
                nextLabel;

            if (tickId !== 0 && prevTick && !prevTick.label) {
                prevTick.label = {};
                prevLabel = true;
            }
            if (nextTick && !nextTick.label) {
                nextTick.label = {};
                nextLabel = true;
            }
            width = proceed.apply(axis, [].slice.call(arguments, 1));
            if (prevLabel) {
                delete prevTick.label;
            }
            if (nextLabel) {
                delete nextTick.label;
            }
        } else {
            width = proceed.apply(axis, [].slice.call(arguments, 1));
        }
        return width;
    });
})(Highcharts);

khlieng:
Fixed #14458, drilling down 3d chart points with the same name threw.

### DIFF
--- a/ts/Core/Axis/Axis3D.ts
+++ b/ts/Core/Axis/Axis3D.ts
@@ -766,19 +766,19 @@ class Axis3D {
 
             // Check whether the tick is not the first one and previous tick
             // exists, then calculate position of previous label.
-            if (tickId !== 0 && prevTick && prevTick.label && (prevTick.label as any).xy) {
+            if (tickId !== 0 && prevTick && prevTick.label && prevTick.label.xy) {
                 prevLabelPos = perspective3D({ // #8621
-                    x: (prevTick.label as any).xy.x,
-                    y: (prevTick.label as any).xy.y,
+                    x: prevTick.label.xy.x,
+                    y: prevTick.label.xy.y,
                     z: null as any
                 }, origin, origin.vd);
             }
             // If next label position is defined, then recalculate its position
             // basing on the perspective.
-            if (nextTick && nextTick.label && (nextTick.label as any).xy) {
+            if (nextTick && nextTick.label && nextTick.label.xy) {
                 nextLabelPos = perspective3D({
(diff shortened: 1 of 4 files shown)


ts/Core/Axis/Axis3D.ts
```diff
@@ -775,7 +775,7 @@ class Axis3D {
             }
             // If next label position is defined, then recalculate its position
             // basing on the perspective.
-            if (nextTick && (nextTick.label as any).xy) {
+            if (nextTick && nextTick.label && (nextTick.label as any).xy) {
```
pawelfus, line 778: Does it mean we can get rid of `as any` castings above and below? :) Could you try it?

TorsteinHonsi (approved):
Thanks!

Issue #433, opened 2011-09-08
rangeSelector doesn't respect UTC setting


four43:
The rangeSelector seems to select the range - local time offset. I'm using UTC: true and offsetting my data server-side before sending my data to the chart. For example, if I set the date to today. It will set it to 5AM this morning. The expected result is 12AM today, since the data is already localized.

kakoni:
Getting into this old issue;
```
if (!defaultOptions.global.useUTC) {
  value = value + new Date().getTimezoneOffset() * 60 * 1000;
}
```
Doesnt take into account that TimezoneOffset differs during the year(daylight savings...)
Example (using moment)
```
moment.tz.zone('Europe/Helsinki').offset(1403465838805)
-180
moment.tz.zone('Europe/Helsinki').offset(1388563200000)
-120
```

So you would need something like
```
value = value + moment.tz.zone(SUPPLIED).offset(value) * 60 * 1000;
```

adamwehmann:
@TorsteinHonsi... 

Following up @kakoni's comment, when:

```
global: {
  useUTC: false
}
```

then
```
rangeSelector: {
  inputDateFormat: '%Y-%m-%d',
  inputEditDateFormat: '%Y-%m-%d',
  inputDateParser: function (value) {
    var temp_date = new Date(value);
    var temp_utc = temp_date.valueOf();
    temp_utc = temp_utc + (temp_date.getTimezoneOffset() - new Date().getTimezoneOffset()) * 60 * 1000;
    return temp_utc;
  }
}
```

seems to be a currently necessary workaround for DST.

PR #14503, opened 2020-11-06
bugfix/14416-rangeselector-timezone

### LINKED ISSUE #14416: Range input changes date when losing focus (blur)
#### Expected behaviour
1. Click inside the "To" input (focus)
2. Click outside the "To" input (blur)
3. Nothing should happen because nothing was modified

#### Actual behaviour
1. Click inside the "To" input (focus)
2. Click outside the "To" input (blur)
3. The "To" date changes from `Jan 4 1970` to `Jan 3 1970`
hc-range-input-bug

https://jsfiddle.net/nunoarruda/ys9qzubn/22/

This issue seems to happen only with the following set of options:
```
time: {
    timezoneOffset: 420
},
xAxis: {
    minRange: 3600 * 1000 // one hour
},
series: [
    {
      pointInterval: 24 * 3600 * 1000,
      data: [1, 2, 3, 4, 5]
    }
]
```

#### Product version
Highstock JS v8.2.2 (2020-10-22)

#### Affected browser(s)
All of them

khlieng:
Fixed #14416, range selector ignored `chart.time.timezoneOffset`.

### DIFF
--- a/ts/Extensions/RangeSelector.ts
+++ b/ts/Extensions/RangeSelector.ts
@@ -1318,6 +1318,7 @@ class RangeSelector {
                 dataMax = dataAxis.dataMax;
 
             value = (options.inputDateParser || defaultInputDateParser)(inputValue, chart.time.useUTC);
+            value += (chart.time.timezoneOffset || 0) * 60 * 1000; // #14416
 
             if (value !== input.previousValue) {
                 input.previousValue = value;
--- a/ts/Extensions/RangeSelector.ts
+++ b/ts/Extensions/RangeSelector.ts
@@ -1270,15 +1270,14 @@ class RangeSelector {
      * @function Highcharts.RangeSelector#defaultInputDateParser
      */
     public defaultInputDateParser(inputDate: string, useUTC: boolean, time?: Highcharts.Time): number {
-        let date;
-
-        if (H.isSafari) {
-            date = Date.parse(inputDate.split(' ').join('T'));
-        } else if (useUTC) {
-            date = Date.parse(inputDate + 'Z');
-        } else {
-            date = Date.parse(inputDate);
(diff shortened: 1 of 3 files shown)


ts/Extensions/RangeSelector.ts
```diff
@@ -1318,6 +1318,7 @@ class RangeSelector {
                 dataMax = dataAxis.dataMax;
 
             value = (options.inputDateParser || defaultInputDateParser)(inputValue, chart.time.useUTC);
+            value += (chart.time.timezoneOffset || 0) * 60 * 1000; // #14416
```
TorsteinHonsi, line 1321: In the [defaultInputDateParser](https://github.com/highcharts/highcharts/blob/v8.2.2/ts/Extensions/RangeSelector.ts/#L1282), we are already correction the values for `date.getTimezoneOffset`. I am wondering if we can improve that correction to use the current state of `chart.time`, by using `time.getTimezoneOffset` instead. We may have to pass in the `time` as an argument though, in case `defaultInputDateParser` is not bound to the `RangeSelector` scope.

TorsteinHonsi (changes requested):

ts/Extensions/RangeSelector.ts
khlieng: The `date.getTimezoneOffset` correction in `defaultInputDateParser` seems to just be correcting for the #433 fix that happened before `defaultInputDateParser` existed :)

TorsteinHonsi: Okay, but can we combine them? Using `time.getTimezoneOffset` should in theory apply to the old issue as well.

TorsteinHonsi (approved):
Nice work, thanks!

PR #14510, opened 2020-11-09
there should be space between two words Update chart-types.md

farhanbadr09:


TorsteinHonsi:
Sorry, "columnrange" is the correct spelling for the `chart.type` setting. See https://api.highcharts.com/highcharts/plotOptions.columnrange.

PR #14517, opened 2020-11-09
Fixed #14436, range selector styling with kbd navigation.

### LINKED ISSUE #14436: Range selector button styling with keyboard navigation
#### Expected behaviour
Navigating through the range selector using TAB and arrow keys should mirror hovering and selecting buttons using the mouse.

#### Actual behaviour
When navigating through the buttons, the styling makes it look like the buttons are being selected. Selecting the buttons resets the styling to neutral, so that it is unclear which button is selected.

#### Product version
Highstock v8.2.2

oysteinmoseng:
Fixed #14436, range selector styling with keyboard navigation.
___
Styling of range selector buttons as the user navigated them with the keyboard was suboptimal, due to incorrect state handling. Now it mirrors the experience with hover/click using a mouse, with the additional focus border.

### DIFF
--- a/js/Accessibility/Components/RangeSelectorComponent.js
+++ b/js/Accessibility/Components/RangeSelectorComponent.js
@@ -44,17 +44,24 @@ function shouldRunInputNavigation(chart) {
  * @return {boolean}
  */
 H.Chart.prototype.highlightRangeSelectorButton = function (ix) {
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #14519, opened 2020-11-09
bugfix/14362-add-series-no-animation

### LINKED ISSUE #14362: AddSeries has no animation
I'm using addSeries to add a curve, why is there no animation, it makes no difference if the third parameter is true or false, I want to animate the curve, just like the animation of the first render.How do I set this

wchmiel:
Fixed #14362, no animation when a series was added dynamically.

~~Fixed issue with animation when series is added dynamically, see #14362.~~


TorsteinHonsi (approved):
Thanks! Fixing an issue by removing code is risky business, as the check for `chart.hasRendered` obviously is there for a reason. But I am not able to find out what breaks if we remove it. It may be that the check on [this line](https://github.com/highcharts/highcharts/commit/5b06bed1a73f146b604a5a5fae2782989c04ed88#diff-231838343b2e90b134f6a741cf16e449772eb59b4105940628c33717c2ebf8faR6618) is already enough, so the one in the `animate` function is redundant. So I'll approve this and hope it doesn't come back to bite us.

PR #14521, opened 2020-11-09
Added rangeSelector.buttons.description option for AT. Closes #14437.

### LINKED ISSUE #14437: Custom aria labels for range selector buttons
Looking at [this demo](http://jsfiddle.net/gh/get/jquery/1.7.2/highcharts/highcharts/tree/master/samples/stock/accessibility/accessible-stock/), the range selector buttons are accessible to screen reader users, but the labels mirror the button text.

It should be possible to configure a custom label text, so that the time abbreviations can be explained.

oysteinmoseng:
Added `rangeSelector.buttons.title` option for assistive technology and general accessability, see #14437.
___
It is now possible to define custom accessible labels for each individual range selector button. This is done through the optional `rangeSelector.buttons.description` option. The default buttons have these descriptions added out of the box. Where a description is not supplied, the button text is used, as before.

It is recommended to add the descriptions since many screen readers will announce confusing information when presented with these abbreviations. `1m` can for example be announced as `One minute`.

We might want to consider if these buttons should have a tooltip so that non-AT users can benefit from the abbreviation explanations on hover.

```js
buttons: [{
  type: 'month',
  count: 1,
  text: '1m',
  description: '1 month'
}, {
  // ...
}, {
  type: 'ytd',
  text: 'YTD',
  description: 'Year to Date'
}, {
  type: 'all',
  text: 'All'
 }]
```

### DIFF
--- a/js/Accessibility/Components/RangeSelectorComponent.js
+++ b/js/Accessibility/Components/RangeSelectorComponent.js
@@ -73,14 +73,16 @@ extend(RangeSelectorComponent.prototype, /** @lends Highcharts.RangeSelectorComp
      * Called on first render/updates to the chart, including options changes.
      */
     onChartUpdate: function () {
+        var _a;
         var chart = this.chart, component = this, rangeSelector = chart.rangeSelector;
         if (!rangeSelector) {
             return;
         }
-        if (rangeSelector.buttons && rangeSelector.buttons.length) {
-            rangeSelector.buttons.forEach(function (button) {
+        if ((_a = rangeSelector.buttons) === null || _a === void 0 ? void 0 : _a.length) {
+            rangeSelector.buttons.forEach(function (button, ix) {
+                var btnOptions = rangeSelector.buttonOptions[ix];
                 unhideChartElementFromAT(chart, button.element);
-                component.setRangeButtonAttrs(button);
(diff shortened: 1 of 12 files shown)


TorsteinHonsi (changes requested):
Thanks!

Now that we put this in the API, perhaps we should also supply it as a `title` to the buttons? It would be a simple addition, and also make the buttons more accessible regardless of screen readers.

And if so, we should probably call it `title` in accordance with for example `contextButton.titleKey`.

PS: How about internationalization?

oysteinmoseng:
Yes, even better if we can add this as a `<title>` element (in which case we might also be able to ditch the `aria-label`).

As for internationalization, I assume that would be the same as for the `1m`, `6m`, `YTD`, `All` strings etc. I am not sure it is necessary to add them to the `lang` options when they can be configured elsewhere?

TorsteinHonsi (changes requested):
> Added now in a11y module, could consider adding the title element in the core instead.

Yes, I think we should do it in the core. And we don't need a new function to set the title, we have it built in to the `SVGElement`: https://jsfiddle.net/highcharts/zecLkhqn/

TorsteinHonsi (approved):
Looking great! Pending CI tests.

Issue #14382, opened 2020-10-20
Additional options for Xaxis and Yaxis are not applied in "getSVG" method


MajidSh854:
#### Expected behaviour
See red in the "getSVG" output instead of green

#### Actual behaviour
See green in the "getSVG" output

#### Live demo with steps to reproduce
https://jsfiddle.net/MajidSh/qxza57pL/1/

#### Product version
highcharts 7.1.2

MajidSh854:
Thanks @PaulDalek

mulanxia:
Hmmm it looks like after https://github.com/highcharts/highcharts/pull/15250/ was merged, this workaround no longer works? Or am I missing something?

KacperMadej:
The problem is no longer visible making the workaround look like not working.
Closing as resolved.

(If the problem is still visible please add a live demo and the issue could be reopened.)

PR #14523, opened 2020-11-09
bugfix/14495-cross-hair-bg-colour-overwritten

### LINKED ISSUE #14495: The lastVisiblePrice label in price-indicator module overwrites the backgroudColor prop of crosshair label
When the LastVisiblePrice label is enabled, the crosshair label backgroudColor gets the value of the LastVisiblePlace label and ignores its own value. 
If LastVisiblePrice is not defined, then the crosshair label has correct color. 

Live demo: https://jsfiddle.net/BlackLabel/13qmwfkr/
Steps to reproduce:  
Hover the chart, and notice, that the crosshair label color is same, as the last defined LastVisiblePrice label color. 

Workaround: 
https://jsfiddle.net/BlackLabel/0qcwprjn/

#### Product version
 Highstock JS v8.2.2
#### Affected browser(s)
All

DJTechnoo:
Fixed #14495, the `lastVisiblePrice` label in price-indicator module overwrote the `backgroudColor` property of the crosshair label

~~Fixed background color not overwritten by lastVisiblePrice label background color.~~

### DIFF
--- a/js/Extensions/PriceIndication.js
+++ b/js/Extensions/PriceIndication.js
@@ -73,6 +73,7 @@ var addEvent = U.addEvent, isArray = U.isArray, merge = U.merge;
  */
 /* eslint-disable no-invalid-this */
 addEvent(LineSeries, 'afterRender', function () {
+    var _a, _b;
     var serie = this, seriesOptions = serie.options, pointRange = seriesOptions.pointRange, lastVisiblePrice = seriesOptions.lastVisiblePrice, lastPrice = seriesOptions.lastPrice;
     if ((lastVisiblePrice || lastPrice) &&
         seriesOptions.id !== 'highcharts-navigator-series') {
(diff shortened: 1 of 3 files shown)


pawelfus (changes requested):
Can we fix all other options too? For example padding: https://jsfiddle.net/BlackLabel/13qmwfkr/2/

Just a guess, but it seems `yAxis.options.crosshair` is overwritten, but never restored, maybe that's it? I see below comment: `// Restore crosshair:` that some props are restored.

pawelfus (approved):
Thanks!
I thought we use single quotes, even in tests, not double quotes, but it seems eslint is fine with that, so am I! 👍

DJTechnoo:
> I thought we use single quotes, even in tests

Whoops, wasn't aware of that! I'll keep it in mind for next time, thanks Pawel 👍

TorsteinHonsi:
> I thought we use single quotes, even in tests, not double quotes, but it seems eslint is fine with that, so am I!

@pawelfus We prefer single quotes, but we have some tests with large chunks of copy-pasted JSON. Maybe we should add a warning rather than an error.

TorsteinHonsi (approved):
Thanks!

TorsteinHonsi:
PS @DJTechnoo See https://github.com/highcharts/highcharts/tree/master/changelog#writing-content-for-the-changelog
* Start with "Fixed #14382..."
* Wrap code in backticks: `lastVisiblePrice`

PR #14526, opened 2020-11-09
ts/funnel-pyramid-series

bre1470:
- Migrated funnel series to class pattern.
- Migrated pyramid series to class pattern.


TorsteinHonsi (approved):
Thanks!

PR #14531, opened 2020-11-10
ts/bullet-series

marvin19:
Migration of bullet series.


ts/Series/Bullet/BulletSeries.ts
```diff
@@ -0,0 +1,511 @@
...
+        if (this.targetGraphic) {
+            this.targetGraphic = this.targetGraphic.destroy() as any;
+        }
+        super.destroy.apply(this, arguments);
+        return;
+    }
+
+    /* eslint-enable valid-jsdoc */
+
+}
```
bre1470, line 392: This should go into a `BulletPoint.ts` (instead of the `BulletPoint.d.ts`).

bre1470 (changes requested):
Thank you. One thing with the point class can be improved.

bre1470 (approved):
Thank you! Looks great. 👍

bre1470:
There is a visual test differing.

TorsteinHonsi (changes requested):
Pending a fix for the visual diff. Apparently the `target` value is no longer considered.

bre1470:
We unable to reproduce it locally. Additionally the graph looks completely different in the browser.

TorsteinHonsi (approved):

PR #14534, opened 2020-11-10
bugfix/14505-axis-minRange-NaN

### LINKED ISSUE #14505: Can drag navigator's drag-bar beyond the minimum x-axis range of stockchart
#### Expected behaviour

Navigator's handle stops when the resized range goes smaller than the minimum range.

#### Actual behaviour

Navigator's handle does not stop, and the chart zooms in beyond the limit.

In the jsfiddle below, grab the one of the navigator dragbars and shrink the range.

https://jsfiddle.net/ktakahashimtb/qf5bumtg/

Normally, navigator won't allow the zoomed range to go beyond the minimum limit, but in this case it allows it to go beyond the limit.

This appears to happen only when the first series contains only a single data point.

Inside the highstock code, this looks to be due to `axis.minRange` being set to `NaN` in `Axis.prototype.adjustForMinRange` function.  The code figuring out `closestDataRange` should probably skip the series with a single data point.  I think this will be very clear once stepping though the code under this scenario.

image

I found that I can get around this problem by explicitly setting xAxis.minRange.

#### Product version
Highstock v.8.2.2

#### Affected browser(s)

Chrome, Firefox, Safari, Edge

khlieng:
Fixed #14505, when one of the chart series had a single point, the navigator handles could be dragged beyond the default `xAxis.minRange`. The computed `minRange` was `NaN`.

~~Fixed #14505, `axis.minRange` was `NaN` when the first series had only a single point.~~

### DIFF
--- a/js/Core/Axis/Axis.js
+++ b/js/Core/Axis/Axis.js
@@ -1004,7 +1004,7 @@ var Axis = /** @class */ (function () {
      * @function Highcharts.Axis#adjustForMinRange
      */
     Axis.prototype.adjustForMinRange = function () {
-        var axis = this, options = axis.options, min = axis.min, max = axis.max, log = axis.logarithmic, zoomOffset, spaceAvailable, closestDataRange, i, distance, xData, loopLength, minArgs, maxArgs, minRange;
+        var axis = this, options = axis.options, min = axis.min, max = axis.max, log = axis.logarithmic, zoomOffset, spaceAvailable, closestDataRange = 0, i, distance, xData, loopLength, minArgs, maxArgs, minRange;
         // Set the automatic minimum range based on the closest point distance
         if (axis.isXAxis &&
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (changes requested):
There's still a case when `axis.minRange` becomes `NaN`: https://jsfiddle.net/highcharts/tb6re42o/. The any cast on line 4989 is suspicious in this regard. When `closestDataRange` is undefined, `axis.minRange` becomes NaN. So we should type check this and remove the any casts. Probably a `minRange` of 0 is acceptable in this case, as `dataMin` and `dataMax` are equal.

PS: when writing for the changelog, make sure to describe the symptoms, not only the underlying cause. End users often won't recognize the internal cause when browsing the changelog to see if their issues is fixed.

TorsteinHonsi (approved):
Looks good!

PR #14539, opened 2020-11-11
bugfix/14262-tickinterval-broken-by-label-rotation

### LINKED ISSUE #14262: xAxis.labels.rotation causing missing lables
#### Expected behaviour
After setting `xAxis.lables.rotation` to 180 the number of labels should be the same as when rotation is not set.

#### Actual behaviour
With the `xAxis.labels.rotation` set to 180 the xAxis will contain just the first and the last label.

https://jsfiddle.net/BlackLabel/j0sw62uo/

#### Product version
8.2.1

#### Affected browser(s)
All

DJTechnoo:
Fixed #14262 180 degrees rotation of labels without skipping any.

### DIFF
--- a/js/Core/Axis/Axis.js
+++ b/js/Core/Axis/Axis.js
@@ -2118,6 +2118,8 @@ var Axis = /** @class */ (function () {
         };
         if (horiz) {
             autoRotation = !labelOptions.staggerLines &&
+                // #14262
+                !labelOptions.rotation &&
                 !labelOptions.step &&
                 ( // #3971
                 defined(rotationOption) ?
--- a/samples/unit-tests/axis/ticks/demo.js
+++ b/samples/unit-tests/axis/ticks/demo.js
@@ -44,7 +44,8 @@ QUnit.test('Ticks were drawn in break(#4485)', function (assert) {
 
 });
 
-QUnit.test('alignTicks should consider only axes with series.(#4442)', function (assert) {
+QUnit.test('alignTicks should consider only ' +
+    'axes with series.(#4442)', function (assert) {
     var chart = $('#container').highcharts({
         yAxis: [{
             endOnTick: false,
@@ -66,7 +67,8 @@ QUnit.test('alignTicks should consider only axes with series.(#4442)', function 
     );
 });
 
-QUnit.test("tickInterval option should take precedence over data range(#4184)", function (assert) {
+QUnit.test('tickInterval option should take ' +
+    'precedence over data range(#4184)', function (assert) {
     var chart = $("#container").highcharts({
         xAxis: {
             min: 0,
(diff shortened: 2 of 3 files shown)


TorsteinHonsi (changes requested):
The proposed solution creates a regression in cases where the labels would overlap:

https://jsfiddle.net/highcharts/kxtdjfhs/

The current algorithm in this cases computes potential bounding boxes and overlaps, and skips every _n_'th label. With the new code, labels will overlap.

I think instead we need to decide what to do with between 90 and 270. Rendering the text upside down is obviously not a recommended setting. @mateuszkornecki do you remember what the use case was here?

And if the use case is worth fixing, maybe we just need to fix the bug in the logic that determines what labels to skip.

mateuszkornecki:
> The proposed solution creates a regression in cases where the labels would overlap:
> 
> https://jsfiddle.net/highcharts/kxtdjfhs/
> 
> The current algorithm in this cases computes potential bounding boxes and overlaps, and skips every _n_'th label. With the new code, labels will overlap.
> 
> I think instead we need to decide what to do with between 90 and 270. Rendering the text upside down is obviously not a recommended setting. @mateuszkornecki do you remember what the use case was here?
> 
> And if the use case is worth fixing, maybe we just need to fix the bug in the logic that determines what labels to skip.
@TorsteinHonsi 
Sorry, but as far as I remember I've never modified anything in this algorithm. Are you sure that you wanted to mention me?

TorsteinHonsi:
@mateuszkornecki You reported the original bug #14262, so I was just wondering if you remember the original use case. Personally I can't imagine when we would want 180 degrees rotated text 🙃

mateuszkornecki:
> @mateuszkornecki You reported the original bug #14262, so I was just wondering if you remember the original use case. Personally I can't imagine when we would want 180 degrees rotated text 🙃

Oh sorry, I totally forgot about it! I don't think that any user wanted to plot labels that are upside-down, I think that someone had a problem with the labels, and I while debugging I noticed that they are disappearing when the rotation is set to 180. Anyway, the same thing is happening with the rotation = 1 https://jsfiddle.net/BlackLabel/pkot59nw/

DJTechnoo:
> > @mateuszkornecki You reported the original bug #14262, so I was just wondering if you remember the original use case. Personally I can't imagine when we would want 180 degrees rotated text 🙃
> 
> I noticed that they are disappearing when the rotation is set to 180

I did some experimenting and found out the number to be around 156.18115

DJTechnoo:
> The proposed solution creates a regression in cases where the labels would overlap:
To me it seems like the `Highcharts.getOptions().lang.months` is causing the overlap here. 
Everything works fine if the months are manually put in, if I am not mistaken?

TorsteinHonsi:
> Anyway, the same thing is happening with the rotation = 1

What happens is that labels are hidden when the _projection_ of each label's bounding box overlaps the next. So in case of rotation = 1, the algorithm sees that _if the text were longer_, it would overlap. And this happens before we know each bounding box. If we want to make it smarter, maybe we can see what is the max length of any label.

> To me it seems like the Highcharts.getOptions().lang.months is causing the overlap here. Everything works fine if the months are manually put in, if I am not mistaken?

Yes you are 😬  . It is [the length of the strings](https://jsfiddle.net/highcharts/kxtdjfhs/1/) that matters, not where we pull them from.

DJTechnoo:
> Yes you are 😬 . It is [the length of the strings](https://jsfiddle.net/highcharts/kxtdjfhs/1/) that matters, not where we pull them from.

Oops, yes I just noticed the month abbreviations 😅

PR #14551, opened 2020-11-12
bugfix/variableTimezone

khlieng:
Fixed the initialization of `Time.variableTimezone`, it caused `Time` to sometimes return the wrong time when `useUTC` was set to `false`.

This should also fix the randomly failing range selector test on CI.


TorsteinHonsi:
Regarding the failure in Safari, be aware that the CI tests run on servers in randomly different time zones. Which is both a curse and a blessing.

khlieng:
Yes, but I'm trying to compare local time to local time, not sure whats happening.

TorsteinHonsi:
Yes that would be the easiest way out, but it would be good if we actually understand why the test fails, so we don't miss anything. You should be able to debug it by setting the time on your machine then run the test in Safari.

TorsteinHonsi (approved):
Looks good!

PR #14564, opened 2020-11-13
tools/test-cli-enhancements

goransle:
Added two additional arguments to the `test` gulp task:
* `--dots`: Switches the reporter to the less verbose "dots".
* `--timeout <milliseconds, default is 2000>`: sets the `browserDisconnectTimeout` option


TorsteinHonsi (approved):
Thanks!

Issue #6405, opened 2017-02-23
After adding "pointIntervalUnit" to column chart, highcharts crashes tab with out of memory exception.


AntanasBarauskas:
#### Expected behaviour

After adding pointIntervalUnit property to plotOptions.series, browser tab results in out of memory exception. This behavior occurs in Chrome/FF/IE. Added jsFiddle example with crashing code.
 
#### Actual behaviour

Crashes with out of memory exception. 
NOTE: Changing "type":"datetime" results in freezing tab instead of out of memory exception.

#### Live demo with steps to reproduce
http://jsfiddle.net/720aboww/

#### Affected browser(s)

Chrome/FF/IE

pawelfus:
Hi @AntanasBarauskas 

Thank you for this report. Could you post also a demo which doesn't crash the browser? It's a bit hard to debug.. :)

Thanks!

AntanasBarauskas:
Hello @pawelfus 

Sure: https://jsfiddle.net/06b4kvge/ :)

pawelfus:
Thanks!

The problem is setting `xAxis.categories`. That means Highcharts will try to render every category (`tickInterval = 1unit`, for a `datetime` axis unit is 1ms) for a given range. So, if you set `pointerInveralUnit` to one year, it means ~365*24*360*1000*number_of_points categories on xAxis.. :) 

After removing `xAxis.categories`, it works fine: https://jsfiddle.net/06b4kvge/2/

AntanasBarauskas:
You're right, removed xAxis.categories and works like a charm!

Thanks.

inoyakaigor:
@pawelfus Hi!
Yesterday I've got an out of memory error. I had wrong config (with xAxis.categories) and in this line browser throw exception.
https://github.com/highcharts/highcharts/blob/b9c683576110b58cedd46e09668e987053e74829/js/parts/Time.js#L745

It was happen because, I believe,  in `tickPositions` array had been containing ≈9 500 000 000 time numbers. Of course it is my mistake but if Highcharts notified me about potentially memory leak I'd save a lot of time.

Therefore I suggest to add warning in this line about big length of tickPositions array. As sample if length more than 10 000 000

pawelfus:
Hi @inoyakaigor 

[...]

PR #14565, opened 2020-11-13
bugfix/14555-big-max

### LINKED ISSUE #14555: Huge values issue
#### Expected behaviour
Can plot chart for series [0, 1.7e+308] with correct scaling and labels on yAxis.

#### Actual behaviour
Empty plot and no yAxis.

https://jsfiddle.net/rsb9Lwym/
Discussion on your forum https://www.highcharts.com/forum/viewtopic.php?f=9&t=45318.

#### Product version
Highcharts JS v8.2.2

#### Affected browser(s)
Google Chrome 86.0.4240.198
Firefox 82.0.2 (x64)

khlieng:
Fixed #14555, big data values rendered empty chart.

### DIFF
--- a/ts/Core/Axis/Axis.ts
+++ b/ts/Core/Axis/Axis.ts
@@ -4809,10 +4809,10 @@ class Axis {
             tickPositions.push(pos);
 
             // Always add the raw tickInterval, not the corrected one.
-            pos = correctFloat(
+            pos = Math.min(correctFloat(
                 pos + tickInterval,
                 precision
-            );
+            ), Number.MAX_VALUE);
 
             // If the interval is not big enough in the current min - max range
             // to actually increase the loop variable, we need to break out to
@@ -5542,9 +5542,10 @@ class Axis {
                 categories ?
                     1 :
                     // don't let it be more than the data range
-                    ((axis.max as any) - (axis.min as any)) *
-                    (tickPixelIntervalOption as any) /
-                    Math.max(axis.len, tickPixelIntervalOption as any)
+                    Math.min(
+                        ((axis.max as any) - (axis.min as any)) * (tickPixelIntervalOption as any),
+                        Number.MAX_VALUE
+                    ) / Math.max(axis.len, tickPixelIntervalOption as any)
             );
         }
 
--- a/js/Core/Axis/Axis.js
+++ b/js/Core/Axis/Axis.js
@@ -918,7 +918,7 @@ var Axis = /** @class */ (function () {
             // Place the tick on the rounded value
             tickPositions.push(pos);
             // Always add the raw tickInterval, not the corrected one.
-            pos = correctFloat(pos + tickInterval, precision);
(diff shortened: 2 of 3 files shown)


TorsteinHonsi (changes requested):
Works well, but something fails with a negative big number: https://jsfiddle.net/highcharts/muzk2yd8/. In the official release it renders a blank chart like the positive big number. With the new PR it renders a line, but the Y axis scale is failing.

ts/Core/Axis/Axis.ts
```diff
@@ -4783,7 +4783,7 @@ class Axis {
         var pos,
             lastPos,
             roundedMin =
-                correctFloat(Math.floor(min / tickInterval) * tickInterval),
+                Math.max(correctFloat(Math.floor(min / tickInterval) * tickInterval), -Number.MAX_VALUE),
```
TorsteinHonsi, line 4786: This `Math.max` doesn't seem to make a difference, as far as I can see this is handled in the new `clamp`?

ts/Core/Axis/Axis.ts
```diff
@@ -4824,6 +4824,11 @@ class Axis {
             // Record the last value
             lastPos = pos;
         }
+
+        if (pos > roundedMax && tickPositions[tickPositions.length - 1] < roundedMax) {
+            tickPositions.push(roundedMax);
```
TorsteinHonsi, line 4829: Is this block necessary? What is the fail case?

TorsteinHonsi (changes requested):

ts/Core/Axis/Axis.ts
khlieng: There would be no 0 tick in the big negative case so the top gets cut off, since the last `pos + tickInterval` is greater than `roundedMax` which is 0.

ts/Core/Axis/Axis.ts
khlieng: The first tick position would be -Infinity without it.

TorsteinHonsi (commented):
Working with this high numbers should be considered an edge case, so we should be careful how much we change core functionality around it. The proposed solution also has its flaws, like the ticks becoming too dense with respect to the `tickPixelInterval` option.

Another possibility would be to not go into the `getLinearTickPositions` function, but rather skip it, like we do in case #6405.

This modification solves the reported case, and puts a min and max tick on the axis:
```js
            if (this.tickInterval > Number.MAX_VALUE) {
                tickPositions = [
                    Math.max(this.min as any, -Number.MAX_VALUE),
                    Math.min(this.max as any, Number.MAX_VALUE)
                ];

            // Too many ticks (#6405). Create a friendly warning and provide two
            // ticks so at least we can show the data series.
            } else if (
```

It however fails with `data: [-1.7e+308, 1.7e+308]`, probably because the combined range then becomes too large, and translation fails. But so does the proposed PR. Arguably, the above poses a cleaner fix. I may also be more compact to combine it with the condition below it, where error 19 is thrown.

PR #14567, opened 2020-11-13
ts/spline-series

bre1470:
Migrated spline series to class pattern.


bre1470:
Visual test difference is probably unrelated. The position of title and tspan in data labels shifted - just drilldown sunburst.

TorsteinHonsi (approved):
Thanks!

PR #14569, opened 2020-11-13
ts/stock-chart

bre1470:
Improved ES module pattern of stock chart.


bre1470:
Change in suburst drilldown SVG is unrelated.

TorsteinHonsi (approved):
Thanks!

PR #14570, opened 2020-11-13
Added AT announcement on range selector btn click. Closes #14438.

### LINKED ISSUE #14438: Selecting range selector button should inform screen reader user of change.
#### Expected behaviour
When selecting a range selector button, screen reader users should be informed that something has changed.

#### Actual behaviour
There is no notification to screen reader users after pressing the buttons.

#### Product version
Highstock v8.2.2

oysteinmoseng:
Added assistive technology announcement on range selector button click. Closes #14438.
___
Assistive technology can now announce when a range selector button is clicked.

A new option `lang.accessibility.rangeSelector.clickButtonAnnouncement` has been introduced for control of the announcement contents. Set to empty to disable the announcement.

By default the announcement contains the new range of the X-axis shown in the chart. The range description format follows the existing [`lang.accessibility.axis`](https://api.highcharts.com/highcharts/lang.accessibility.axis) range description formats, and will adapt to the axis type. The date/time formats follow the existing [`accessibility.screenReaderSection.axisRangeDateFormat`](https://api.highcharts.com/highcharts/accessibility.screenReaderSection.axisRangeDateFormat).

A new internal event `chart.afterRangeSelectorBtnClick` was added.


oysteinmoseng:
Visual differences seem to be unrelated and only affecting IDs in code?

bre1470:
In local tests I saw that the SVG order of title and tspan shifted for datalabels - but only in sunburst drilldowns. Is this related to some A11Y fix?

oysteinmoseng:
@bre1470 Yes, I assume this PR: https://github.com/highcharts/highcharts/pull/14521. `<title>` should now always be added as the first child element, as per the spec.

ts/Extensions/RangeSelector.ts
```diff
@@ -951,6 +951,8 @@ class RangeSelector {
                 }
             );
         }
+
+        fireEvent(chart, 'afterRangeSelectorBtnClick');
```
TorsteinHonsi, line 955: Is there a good reason for firing the event on the chart instead of the `RangeSelector`? Seems like a bit of a pollution of the `Chart` with something that it shouldn't be concerned with...

TorsteinHonsi (changes requested):

ts/Extensions/RangeSelector.ts
oysteinmoseng: I initially had it on RangeSelector, but that class is not set up to deal with events when `destroy`ing and such, so it would be a more elaborate refactoring. Should we change it?

TorsteinHonsi: Yes I think so... I think it would be cleaner

TorsteinHonsi: PS: If you add it to the RangeSelector class rather than the instance, you should not have to worry about destroying.

```js
addEvent(RangeSelector, 'afterBtnClick', e => {});
```

TorsteinHonsi (approved):
Thanks!

PR #14572, opened 2020-11-14
ts/sankey-series

bre1470:
Added initial class pattern to sankey series.


bre1470:
Change in suburst drilldown SVG is unrelated.

Issue #14090, opened 2020-08-19
temporaryDisplay messes up parent DOM


naksenzh:
Regression: #7508

https://github.com/highcharts/highcharts/issues/7508

Expected behavior
Highstock doesn't affect DOM of its container's parent elements

Actual behavior
If the chart is created (or redrawn) when it's hidden, Highstock will call temporaryDisplay, which affects the parent elements' DOM. In particular, I've been having problems where it changes the "important" attribute of "display:none" on a parent element.

Would it be possible to add a setting that tells Highstock never to do this? Maybe in that case it could just delay drawing until it's made visible.

Live demo with steps to reproduce
Affected browser(s)
all

naksenzh:
@oysteinmoseng FYI we still have the same issue and it introduced regression when the team upgrades to the latest highcharts and latest accessibility module

oysteinmoseng:
It looks like there was a workaround in the old issue, but that the issue was never solved in the core. @TorsteinHonsi can probably confirm. The workaround in the old issue seems to still work with the latest version for me. The issue also seems unrelated to the accessibility module - at least in my tests.

@naksenzh Could you try applying the workaround from the old issue and report back? If you are still experiencing issues, would you be able to create a jsFiddle demo for debugging?

oysteinmoseng:
Old issue workaround (run before creating chart):

```javascript
(function (H) {
	var doc = document,
    	each = H.each;
    /** 
	 * If the renderTo element has no offsetWidth, most likely one or more of
	 * its parents are hidden. Loop up the DOM tree to temporarily display the
	 * parents, then save the original display properties, and when the true
	 * size is retrieved, reset them. Used on first render and on redraws.
	 *
	 * @private
	 * 
	 * @param  {Boolean} revert
	 *         Revert to the saved original styles.
	 */
	H.Chart.prototype.temporaryDisplay = function (revert) {
		var node = this.renderTo,
			tempStyle;
		if (!revert) {
			while (node && node.style) {

				// When rendering to a detached node, it needs to be temporarily
				// attached in order to read styling and bounding boxes (#5783,
				// #7024).
				if (!doc.body.contains(node) && !node.parentNode) {
					node.hcOrigDetached = true;
					doc.body.appendChild(node);
				}
				if (
					H.getStyle(node, 'display', false) === 'none' ||
					node.hcOricDetached
				) {
					node.hcOrigStyle = [
						['display', node.style.getPropertyValue('display'),
[...]

PR #14575, opened 2020-11-16
Fixed a11y issue with absolute positioning, touches #14090.

oysteinmoseng:
Fixed a11y issue with absolute positioning, touches #14090.
___
Incorporated workaround from issue into core - no longer position hidden elements absolutely without a relative container.


TorsteinHonsi (approved):
Thanks!

PR #14578, opened 2020-11-16
Fixed #14460, kbd click not firing series.click event.

### LINKED ISSUE #14460: series.events.click is not fired via keyboard enter / return / space
#### Expected behaviour
If I add a click event under `series.events.click`:

```js
series: [{
        events: {
            click: () => alert("Clicked!")
        }
    }]
```

That event should be fired when I click on the series with the mouse, **or** when the series has the keyboard focus and I press enter / return / space.

#### Actual behaviour
The event is only fired when I click on the series. The event does not fire when I press enter / return / space.

If I add a click event under `series.point.events.click` that event *does* fire when I navigate to a point in the chart and press enter / return / space.

Event does not fire on keyboard interaction with `series.events.click`:
http://jsfiddle.net/njb15qc2/

Event fires with mouse and keyboard interaction with `series.point.events.click`: 
http://jsfiddle.net/njb15qc2/1/

#### Product version
`highcharts.js` 8.2.0
`modules/accessibility.js` 8.2.0

#### Affected browser(s)
Google Chrome - Version 86.0.4240.75 (Official Build) (64-bit)

oysteinmoseng:
Fixed #14460, keyboard enter/space press not firing `series.click` event.
____
`series.point.click` works as before, but now `series.click` is also fired - similarly to the behavior with mouse/touch.

### DIFF
--- a/js/Accessibility/Components/SeriesComponent/SeriesKeyboardNavigation.js
+++ b/js/Accessibility/Components/SeriesComponent/SeriesKeyboardNavigation.js
@@ -16,7 +16,7 @@ import Chart from '../../../Core/Chart/Chart.js';
 import LineSeries from '../../../Series/Line/LineSeries.js';
 import Point from '../../../Core/Series/Point.js';
 import U from '../../../Core/Utilities.js';
-var defined = U.defined, extend = U.extend;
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #14582, opened 2020-11-17
ts/dumbbell-series

goransle:
Migrated `dumbbell` series to class pattern


ts/Series/Dumbbell/DumbbellPoint.ts
```diff
@@ -0,0 +1,146 @@
...
+    isValid: typeof AreaRangePoint.prototype.isValid;
+}
+
+extend(DumbbellPoint.prototype, {
+    // seriesTypes doesn't inherit from arearange point proto so put below
+    // methods rigidly.
+    destroyElements: AreaRangePoint.prototype.destroyElements,
+    isValid: AreaRangePoint.prototype.isValid
+});
+
```
bre1470, line 139: I believe we do not need to do this, if we have `class DumbbellPoint extends AreaRangePoint`.

bre1470 (approved):
Thank you! Looks good. 👍

PR #14586, opened 2020-11-17
enhancement/rangeselector-date-input

khlieng:
Added native browser date pickers to the range selector. Supported browsers are Chrome, Firefox, Edge and Safari for mobile. Other browsers fall back to text inputs.

### To do

 - [x] When picking a date with the mouse, it would be nice if the date input were blurred (so the SVG box appears). What happens now is that it stays visible and focused after I have picked a date.
 - [x] Now that we use `inputDateFormat` to do a smart guess on what type of date picker to render, we need visual samples to show users how it works. Add samples to https://api.highcharts.com/highstock/rangeSelector.inputDateFormat.
 - [x] Unit tests
 - [x] Update descriptions in API (`inputDateFormat`?) and [docs](https://www.highcharts.com/docs/stock/range-selector) on how this works.
 - [x] We need a team decision on how to deal with the greater space needed for the date inputs. Preliminary solution is to center it above the date box and add a shadow.
 - [x] On iPhone, after picking dates, there are small remnants of down arrows left on screen. Also, it seems that `min` and `max` is doesn't work.
 - [x] Assert accessibility. Are there any showstoppers?


TorsteinHonsi:
Thanks so far! Some comments:

* When picking a date with the mouse, it would be nice if the date input were blurred (so the SVG box appears). What happens now is that it stays visible and focused after I have picked a date.
* Now that we use `inputDateFormat` to do a smart guess on what type of date picker to render, we need visual samples to show users how it works. Add samples to https://api.highcharts.com/highstock/rangeSelector.inputDateFormat.
* Unit tests
* We need a team decision on how to deal with the greater space needed for the date inputs.

TorsteinHonsi:
Thanks so far! Some comments:

* When picking a date with the mouse, it would be nice if the date input were blurred (so the SVG box appears). What happens now is that it stays visible and focused after I have picked a date.
* Now that we use `inputDateFormat` to do a smart guess on what type of date picker to render, we need visual samples to show users how it works. Add samples to https://api.highcharts.com/highstock/rangeSelector.inputDateFormat.
* Unit tests
* We need a team decision on how to deal with the greater space needed for the date inputs.
* Currently the date input activates on the first click, but it takes a second click to open the date picker widget. Should the calendar open on first click? Or perhaps we should reveal the date input on hover to reduce the amount of clicks?
* On iPhone, after picking dates, there are small remnants of down arrows left on screen. Also, it seems that `min` and `max` is doesn't work.

js/Extensions/RangeSelector.js
```diff
@@ -1056,16 +1110,29 @@ var RangeSelector = /** @class */ (function () {
...
+        var keyDown = false;
         // handle changes in the input boxes
-        input.onchange = updateExtremes;
+        input.onchange = function () {
+            updateExtremes();
+            // Blur input when clicking date input calendar
+            if (!keyDown) {
+                rangeSelector.hideInput(name);
+                input.blur();
+            }
```
TorsteinHonsi, line 1124: Unfortunately, this also hides the input when we're arrowing through dates. To reproduce, click the date box so the input appears. Now using the left and right arrows works. Up and down arrows correctly sets the date for the chart, but it hides the input which it shouldn't do. Ideally we should be able to arrow up and down through dates, and the chart should change and input stay visible.

TorsteinHonsi (changes requested):

js/Extensions/RangeSelector.js
khlieng: Ah yes, need the keydown event then probably.

oysteinmoseng:
I've done some a11y testing, and there are definitely some issues. I'll take a look and see how much we can do and how much is inherent to the native datepicker. There seems to be some conflicts in terms of keyboard navigation functionality, though, and that we should be able to fix.

ts/Extensions/RangeSelector.ts
```diff
@@ -527,6 +535,8 @@ extend(defaultOptions, {
          * The first argument passed is a value to parse,
          * second is a boolean indicating use of the UTC time.
          *
+         * This will only get called for inputs of type `text`.
```
pawelfus, line 538: Should add info since which version it does happen?

Please, don't forget to add this info in changelog about breaking changes 👍

ts/Extensions/RangeSelector.ts
```diff
@@ -1294,7 +1383,7 @@ class RangeSelector {
         }
 
         if (time && useUTC) {
-            date += time.getTimezoneOffset(date) * 60 * 1000;
+            date += time.getTimezoneOffset(date);
```
pawelfus, line 1386: Is this change expected?

pawelfus (approved):
Looks great, thank you!

Two minor comments, not showstoppers.

ts/Extensions/RangeSelector.ts
khlieng: Yes I wrote the original code recently, it was broken, `getTimezoneOffset` does the multiplication itself.

pawelfus: Okay, thanks! 👍

TorsteinHonsi: @khlieng

ts/Extensions/RangeSelector.ts
TorsteinHonsi: Done

TorsteinHonsi (approved):

PR #14591, opened 2020-11-17
docs/14583-fix-series-threshold-type

### LINKED ISSUE #14583: plotOptions.series.threshold Typescript type does not accept null
#### Expected behaviour
plotOptions.series.threshold can be set to null in order to automatically scale the y-axis according to the data shown, not e.g.  have as lower limit default 0 but a value close to the minimum value of the data.

#### Actual behaviour
plotOptions.series.threshold cannot be set to null, since typescript definition only allows Number and undefined.

I copied this example from https://github.com/highcharts/highcharts/issues/12824, where the problem seems to be solved at least for some cases. "threshold: null" cannot be used when using typescript.

    xAxis: {
    	categories: ['Apples', 'Pears', 'Bananas', 'Oranges']
    },
    series: [{
        data: [101, 104, 103, 105],
        type: 'area',
        threshold: null
    }]
});

#### Product version
8.2.2

#### Affected browser(s)
Has nothing do with a browser.

bre1470:
Fixed #14583 the types of series threshold options.


ts/Core/Series/Series3D.ts
```diff
@@ -67,7 +67,7 @@ class Series3D extends LineSeries {
      *
      * */
 
-    public static defaultOptions = merge(LineSeries.defaultOptions);
+    public static defaultOptions = merge(LineSeries.defaultOptions, {});
```
TorsteinHonsi, line 70: Why the change? The `merge` function creates a deep copy when fed a single argument...

ts/Series/Spline/SplineSeries.ts
```diff
@@ -62,7 +62,7 @@ class SplineSeries extends LineSeries {
      * @product      highcharts highstock
      * @optionparent plotOptions.spline
      */
-    public static defaultOptions: SplineSeriesOptions = merge(LineSeries.defaultOptions);
+    public static defaultOptions: SplineSeriesOptions = merge(LineSeries.defaultOptions, {});
```
TorsteinHonsi, line 65: As above

TorsteinHonsi (changes requested):

ts/Core/Series/Series3D.ts
bre1470: This is to be consistent with the other change.

ts/Series/Spline/SplineSeries.ts
bre1470: Here it is because of the JSDoc parser, as it needs curly brackets.

TorsteinHonsi: What's stopping future us from removing the empty object because we know it doesn't make a difference logically, then breaking JSDoc? Can we fix JSDoc instead? Are there more cases like this?

bre1470: That is the only one, I am aware of. We might get JSDoc fixed instead. I will give it a try tomorrow.

TorsteinHonsi (approved):
Thanks!

PR #14592, opened 2020-11-17
Refactor palette into an ES module

TorsteinHonsi:
## Current situation as of v8.2.2
Palette colors in the source code are denoted as `${palette.xxx}`. This is replaced by the build script in highcharts-assembler. The highcharts-assembler’s `getPalette` function parses `highcharts.scss` and extracts the palette colors. The results are returned as an object and are used to replace the strings in the built source code.

## Downsides of the current situation
The palette is part of a secondary build step though it doesn’t have to be. As a result, for example the local transpiled ES modules have `${palette.xxx}` pseudo variables and can’t be loaded directly. If not the `highcharts.css` is loaded, all colors appear black. This can be seen in http://jsfiddle.net/highcharts/jhem90dL/ if we remove the CSS import statement.

Additionally the code would appear cleaner if we didn’t have these post-build replacements.

### To do
 - [x] Make the palette into an ES module. Use a modification of the highcharts-assembler’s `getPalette` function to build the ES module. It’s probably best to move this function over to the main repo and create a gulp task for it. The palette very rarely changes, so we don’t have to automate the process. Just make a task for building the palette module, then we can run this manually the next time it changes.
 - [x] Now import the palette module into every place it is used within the Highcharts source code, and replace all the current pseudo-variables.
 - [x] Fix the API. Currently the API is ignorant of the palette stuff, because the highcharts-assembler runs first and replaces the string. So all the API sees is the actual color values, and it uses those for default values. Both inferred defaults and `@default` tags. As part of the refactoring process, we need to import the palette ES module into the gulp jsdoc task, and use its values to populate the defaults.
 - [x] Clean up the highcharts-assembler. Hopefully we can remove all the palette related stuff now.


ts/Core/Palette.ts
```diff
@@ -0,0 +1,88 @@
...
+    /**
+     * Positive indicator color
+     */
+    indicatorPositiveLine: '#06b535' as ColorString,
+    /**
+     * Negative indicator color
+     */
+    indicatorNegativeLine: '#f21313' as ColorString,
+
+};
```
bre1470, line 87: If you write this as a namespace with a bunch of constants, you avoid forced casting and the JS outcome would be the same, plus you get a uppercase namespace object.
```ts
namespace Palette {
    …
    /**
     * Negative indicator color
     */
    export const indicatorNegativeLine: ColorString = '#f21313';
}
export default Palette;
```

bre1470: Sorry, I forgot that we need a JSON-like structure.

bre1470 (approved):
Looks good. 👍 

Though I suggest to have `Palette.ts` (and .js) in the `Core/Color` folder as it is highly color related.

TorsteinHonsi:
> Though I suggest to have Palette.ts (and .js) in the Core/Color folder as it is highly color related.

Good point... But then `highcharts-assembler` and `highcharts-documentation-generators` would have to be updated too, and versioned, package.json be updated etc :(

bre1470:
Better do it now than later, I would say. Otherwise we adjust to new exceptions in our tooling and it becomes harder to do the change in the long run. Also changing it twice is still better than the one time where I had four releases of the documentation-generators in one hour.

PR #14593, opened 2020-11-17
ts/packedbubble-series

bre1470:
Migrated packedbubble series to class pattern.


bre1470:
@Izothep @sebastianbochan Can you confirm, that the diffs of the bubble series (packed bubble halo) are wrong? For me the candidate after migration looks more correct than the reference.

bre1470:
Just to clarify: The reason that it differs is that now the packedbubble point inherits from the bubble point. With the old factory function it would inherit from the linepoint instead. Nevertheless I believe the new outcome in the visuals is correct.

pawelfus:
I agree, the candidate's halo looks much better 👍

Izothep:
I agree, the candidate is much better!

TorsteinHonsi (approved):
Thanks!

PR #14596, opened 2020-11-18
Tools: moved XML prettifying on reference SVGs

goransle:
@TorsteinHonsi Issue with visual tests seems to be that we inserted a newline inside `title` elements, which is interpreted as a whitespace on render.

It is probably best to run the test on the raw SVG and move the pretty printing to the visual test application (which also needs to be updated as the file diff is broken without newlines).


TorsteinHonsi (approved):
Thanks!

PR #14599, opened 2020-11-18
ts/atr-indicator

sebastianbochan:
Migrated atr indicator to class pattern.

### DIFF
--- a/ts/Stock/Indicators/ATR/ATRIndicator.ts
+++ b/ts/Stock/Indicators/ATR/ATRIndicator.ts
@@ -128,6 +128,7 @@ class ATRIndicator extends SMAIndicator {
 
     public data: Array<ATRPoint> = void 0 as any;
     public points: Array<ATRPoint> = void 0 as any;
+    public options: ATROptions = void 0 as any;
 
     /* *
      *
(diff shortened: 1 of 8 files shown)


sebastianbochan:
@TorsteinHonsi @bre1470 
The 3 charts are completely the same, so I we can approve them.

TorsteinHonsi (commented):
> The 3 charts are completely the same, so we can approve them.

Yes, this is a problem with the SVG comparison script, we have now fixed it for new forks from the master.

Pending @bre1470 's approval.

ts/Stock/Indicators/ATR/ATRIndicator.ts
```diff
@@ -0,0 +1,250 @@
...
+    } as ATROptions);
+
+    /* *
+     *
+     *  Properties
+     *
+     * */
+
+    public data: Array<ATRPoint> = void 0 as any;
+    public points: Array<ATRPoint> = void 0 as any;
```
bre1470, line 130: `options` is missing here.

bre1470 (changes requested):
Thank you! One typed property is missing in the class pattern. See comment.

ts/Stock/Indicators/ATR/ATRIndicator.ts
sebastianbochan: In the interface, I was missed.

Fixed.

bre1470 (approved):
Thank you! Looks good. 👍

PR #14601, opened 2020-11-18
ts/aroon-indicator

wchmiel:
Migrated aroon indicator to the class pattern.


bre1470 (approved):
Thank you! Looks good. 👍

PR #14604, opened 2020-11-19
ts/arearange-series

goransle:
Migrated `arearange` series to class pattern.

### DIFF
--- a/ts/Series/AreaRange/AreaRangeDataLabelsOptionsObject.d.ts
+++ b/ts/Series/AreaRange/AreaRangeDataLabelsOptionsObject.d.ts
@@ -1,0 +1,37 @@
+/* *
+ *
+ *  (c) 2010-2020 Torstein Honsi
+ *
+ *  License: www.highcharts.com/license
+ *
+ *  !!!!!!! SOURCE GETS TRANSPILED BY TYPESCRIPT. EDIT TS FILE ONLY. !!!!!!!
+ *
+ * */
+
+/* *
+ *
+ *  Imports
+ *
+ * */
+
+import type DataLabelOptions from '../../Core/Series/DataLabelOptions';
+
+/* *
+ *
+ *  Declarations
+ *
+ * */
+export interface AreaRangeDataLabelsOptionsObject extends DataLabelOptions {
+    xHigh?: number;
+    xLow?: number;
+    yHigh?: number;
+    yLow?: number;
+}
+
+/* *
+ *
+ *  Default export
+ *
+ * */
+
+export default AreaRangeDataLabelsOptionsObject;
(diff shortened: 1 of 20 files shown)


goransle:
@bre1470 do you see why the AreaRangeSeries ends up in the `highcharts-more.d.ts` file like this?

```js
/**
 * Adds the module to the imported Highcharts namespace.
 *
 * @param highcharts
 *        The imported Highcharts namespace to extend.
 */
export function factory(highcharts: typeof Highcharts): void;
declare module "./highcharts.src" {

	....

    /**
     * Class
     */
    let AreaRangeSeries: any;
}
....
```

ts/Series/AreaRange/AreaRangeDataLabelsOptionsObject.d.ts
```diff
@@ -0,0 +1,37 @@
...
+ * */
+
+import type DataLabelOptions from '../../Core/Series/DataLabelOptions';
+
+/* *
+ *
+ *  Declarations
+ *
+ * */
+export interface AreaRangeDataLabelsOptionsObject extends DataLabelOptions {
```
bre1470, line 24: This all (interface, file name, import, export) should become `AreaRangeDataLabelOptions`.

ts/Series/AreaRange/AreaRangePoint.ts
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    public isValid(): boolean {
+        return isNumber(this.low) && isNumber(this.high);
+    }
+}
+
+extend(AreaRangePoint.prototype, {
+
+});
```
bre1470, line 177: You can delete that one. :)

ts/Series/AreaSplineRangeSeries.ts
```diff
@@ -10,8 +10,12 @@
 
 'use strict';
 
+import type AreaRangePoint from './AreaRange/AreaRangePoint';
+import type AreaRangePointOptions from './AreaRange/AreaRangePointOptions';
+import type AreaRangeSeriesOptions from './AreaRange/AreaRangeSeriesOptions';
 import type { SeriesStatesOptions } from '../Core/Series/SeriesOptions';
 import type SplineSeries from './Spline/SplineSeries';
+import AreaRangeSeries from './AreaRange/AreaRangeSeries.js';
```
bre1470, line 18: `import type` should be enough.

ts/Series/DumbbellSeries.ts
```diff
@@ -38,11 +42,10 @@ const {
...
 } = U;
 
-import './AreaRangeSeries.js';
 import './ColumnRangeSeries.js';
 import '../Core/Interaction.js';
 
-var areaRangeProto = seriesTypes.arearange.prototype,
+var areaRangeProto = AreaRangeSeries.prototype,
     columnRangeProto = seriesTypes.columnrange.prototype,
     areaRangePointProto = areaRangeProto.pointClass.prototype,
```
bre1470, line 50: Maybe it is better to do a deconstruct on BaseSeries here:
```
import BaseSeries from '../Core/Series/Series.js';
const {
    seriesTypes: {
        arearange: {
            prototype: areaRangeProto
        },
        columnrange: {
             prototype: columnRangeProto
        }
    }
} = BaseSeries;
```

bre1470 (changes requested):
Thank you! See comments above and below.

bre1470 (approved):
Thank you! Looks good. 👍

PR #14605, opened 2020-11-19
ts/dpo-indicator

madepiet:
Added initial class pattern to dpo indicator.


ts/Stock/Indicators/DPO/DPOIndicator.ts
```diff
@@ -0,0 +1,232 @@
...
+
+/* *
+*
+*   Prototype Properties
+*
+* */
+
+interface DPOIndicator {
+    nameBase: string;
+    options: DPOOptions;
```
bre1470, line 192: The options property should go into the class pattern.

bre1470 (changes requested):
Thank you! One small change needed, otherwise it is good.

ts/Stock/Indicators/DPOIndicator.ts
```diff
@@ -1,217 +0,0 @@
...
-             * Period for Detrended Price Oscillator
-             */
-            period: 21
-        }
-    },
-    /**
-     * @lends Highcharts.Series#
-     */
-    {
-        nameBase: 'DPO',
```
bre1470, line 136: I believe that the value of nameBase somehow got removed. This part needs to be restored with the `extend` call right under the interface of the prototype properties.
```ts
extend(DPOIndicator.prototype, {
    nameBase: 'DPO'
});
```

bre1470 (changes requested):
Thank you so far. I found an other issue. Please see the comment.

Note that you also have to fix the auto merge conflict caused by changes of other indicators.

bre1470 (approved):
Thank you! Looks good. 👍

sebastianbochan (approved):

PR #14606, opened 2020-11-19
ts/cci-indicator

sebastianbochan:
Added initial class pattern to CCI indicator.


ts/Stock/Indicators/CCI/CCIIndicator.ts
```diff
@@ -0,0 +1,218 @@
...
+     * */
+
+    public data: Array<CCIPoint> = void 0 as any;
+    public points: Array<CCIPoint> = void 0 as any;
+    public options: CCIOptions = void 0 as any;
+
+    public getValues<TLinkedSeries extends LineSeries>(
+        series: TLinkedSeries,
+        params: CCIParamsOptions
+    ): (IndicatorValuesObject<TLinkedSeries>|undefined) {
```
bre1470, line 113: `getValues` is a function. You have two options: Add a generic "Functions" comment or remove the "Properties" one. This is important to distinguish, because property assignments in class pattern go on the instance, functions on the prototype.

bre1470 (changes requested):
Thank you! See comments.

sebastianbochan:
@bre1470 all issues have been already fixed.

bre1470 (approved):
Thank you! Looks good. 👍

PR #14608, opened 2020-11-19
enhancement/series-area-3d

Izothep:
Added support for area series in 3D charts.


ts/Series/Area3DSeries.ts
```diff
@@ -0,0 +1,107 @@
...
+        for (var i = 0; i < series.points.length; i++) {
+            bottomPoints.push({
+                x: series.rawPointsX[i],
+                y: options.stacking ? series.points[i].yBottom : translatedThreshold,
+                z: series.zPadding
+            } as any);
+        }
+    }
+
+    if (series.chart.is3d() && series.chart.options && series.chart.options.chart) {
```
pawelfus, line 74: You checked `is3d()` already above

pawelfus (changes requested):
Apart from that one minor thing, amazing work!

ts/Series/Area3DSeries.ts
Izothep: I have removed this check. Thanks!

TorsteinHonsi (changes requested):
Thanks, it looks great!

It seems the Z index is not 100% though: https://jsfiddle.net/highcharts/dk486t5m/

In addition I think we need some more visual samples both for user education and for regression testing:
* Multiple series non-stacked
* Multiple series with stacking
* Perhaps a combination of stacking and not

Izothep:
All corrected, thanks @TorsteinHonsi! I will add the demo to the related docs article in a separate PR.

TorsteinHonsi (changes requested):
Thanks!

There is still a Z index issue with the markers: https://jsfiddle.net/highcharts/dkqe7m5y/. Note how _Series 2_'s markers are hidden behind _Series 1_'s fill area.

> I will add the demo to the related docs article in a separate PR.

Can we have the docs and the nice demo in the same PR please? So that we don't lose track of this new feature.

Izothep:
Thanks for the review. `zIndex` problem is solved in the newest version of the code. I have added a demo and prepared docs explanation (I am not sure if they are written correctly though).

TorsteinHonsi (approved):
Perfect, thanks!

PR #14609, opened 2020-11-19
Increased timeouts in local tests.

pawelfus:
Empirically, the longest tests took around 15s (the slowest tests are related to panning, we can improve that too - probably `mousemove`  event is fired continuously).

Sometimes 20s could fail, so 30s sounds safe.

I don't see any noticeable time increase in tests (for me, still tests need less than 5mins). When running tests and observing CPU monitor, CPU usage jumped to 250%+ on one core (?!) and timeout error showed up in the console.


TorsteinHonsi (approved):
Thanks!

PR #14610, opened 2020-11-19
ts/momentum-indicator 

karolkolodziej:


js/masters/indicators/momentum.src.js
```diff
@@ -11,4 +11,4 @@
  * License: www.highcharts.com/license
  */
 'use strict';
-import '../../Stock/Indicators/MomentumIndicator.js';
+import '../../Stock/Indicators//Momentum/MomentumIndicator.js';
```
bre1470, line 14: Not sure if it is a problem on any platform, but to make sure the double slash should be fixed.

bre1470 (changes requested):
Thank you! One small thing otherwise good.

js/masters/indicators/momentum.src.js
karolkolodziej: True! I missed that. Thanks for spotting that

bre1470 (approved):
Thank you! Looks good. 👍

TorsteinHonsi (changes requested):
What's up with the visual diff? It seems like the series name is suddenly all upper case, but I can't see any change in the code related to it...

ts/Stock/Indicators/Momentum/MomentumIndicator.ts
```diff
@@ -0,0 +1,187 @@
...
+     * @optionparent plotOptions.momentum
+     */
+    public static defaultOptions: MomentumOptions = merge(SMAIndicator.defaultOptions, {
+        params: {
+            period: 14
+        }
+    } as MomentumOptions);
+
+    public data: Array<MomentumPoint> = void 0 as any;
+    public nameBase: string = void 0 as any;
```
bre1470, line 79: `nameBase` musty be set like in the original file and goes on the prototype.

ts/Stock/Indicators/Momentum/MomentumIndicator.ts
```diff
@@ -0,0 +1,187 @@
...
+/* *
+ *
+ *  Prototype Properties
+ *
+ * */
+
+interface MomentumIndicator {
+    pointClass: typeof MomentumPoint;
+}
+
```
bre1470, line 151: Here the extend of prototype with properties has to be added:
```ts
extend(MomentumIndicator.prototype, {
    nameBase: 'Momentum'
});
```

bre1470 (changes requested):
Spotted by @TorsteinHonsi , the baseName value is missing in the (missing) `extend` call.

ts/Stock/Indicators/Momentum/MomentumIndicator.ts
bre1470: To clarify it: `nameBase` has to be removed from the class pattern.

TorsteinHonsi:
Pending @bre1470 's approval

bre1470 (approved):
Thank you! Looks good. 👍

TorsteinHonsi (approved):

PR #14613, opened 2020-11-19
bugfix/13473-bar-shrinking-when-scrolling

### LINKED ISSUE #13473: disappearing column/bar when scrolling beyond point value
#### Expected behaviour
All points should be visible after scrolling.

#### Actual behaviour
When you scroll out the chart so the first point will be outside of the scroll area it will disappear after scrolling back.

https://jsfiddle.net/BlackLabel/x40da2gq/
Steps: 
1/ Scroll fully to the right, 
2/ Return to the starting position
3/ The first point won't be visible anymore

#### Workaround
After setting the min value everything is working correctly:
https://jsfiddle.net/BlackLabel/tz69qwa1/

#### Product version
8.1.0

#### Affected browser(s)
All

DJTechnoo:
Fixed #13473 bar shrinking when scrolling.

### DIFF
--- a/ts/Core/Axis/ScrollbarAxis.ts
+++ b/ts/Core/Axis/ScrollbarAxis.ts
@@ -75,6 +75,12 @@ class ScrollbarAxis {
                     this: Highcharts.Scrollbar,
                     e: Highcharts.ScrollbarChangedEventObject
                 ): void {
+                    if ((this.chart.options.chart?.type === 'column' ||
+                        this.chart.options.chart?.type === 'bar') &&
+                        !axis.options.min
+                    ) {
+                        axis.options.min = 0;
+                    }
                     var axisMin = pick(
                             axis.options && axis.options.min,
                             axis.min as any
--- a/ts/Core/Axis/ScrollbarAxis.ts
+++ b/ts/Core/Axis/ScrollbarAxis.ts
@@ -75,6 +75,7 @@ class ScrollbarAxis {
                     this: Highcharts.Scrollbar,
                     e: Highcharts.ScrollbarChangedEventObject
                 ): void {
+                    // #13473 Bars shrink if min is not set while scrolling
                     if ((this.chart.options.chart?.type === 'column' ||
                         this.chart.options.chart?.type === 'bar') &&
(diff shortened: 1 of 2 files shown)


ts/Core/Axis/ScrollbarAxis.ts
```diff
@@ -75,6 +75,13 @@ class ScrollbarAxis {
                     this: Highcharts.Scrollbar,
                     e: Highcharts.ScrollbarChangedEventObject
                 ): void {
+                    // #13473 Bars shrink if min is not set while scrolling
+                    if ((this.chart.options.chart?.type === 'column' ||
+                        this.chart.options.chart?.type === 'bar') &&
+                        !defined(axis.options.min)
```
pawelfus, line 81: Ouch, I'm afraid this is a bit more tricky 😅 
Checking for `chart.type` is not enough. This option can be set to `line` but user may add `column` type series: `series: [{ type: 'column }]`. 

Checking for every series is rather cumbersome. I think (and might be wrong!) this is related to `series.threshold` and `Axis.threshold`.

Just a side note, perhaps you already know that, but you can use `series.is('column')` to check if a series is type of `column` or based on `column` (like `bar`, `columnrange` itp).

ts/Core/Axis/ScrollbarAxis.ts
```diff
@@ -75,6 +75,13 @@ class ScrollbarAxis {
...
+                    ) {
+                        axis.options.min = 0;
```
pawelfus, line 83: Setting `options.something` is not a good idea, from my experience creates more bugs later. Since this callback sets extremes on Axis, perhaps extremes are wrong? As I mentioned above, I don't see `Axis.threshold` used anywhere here, so maybe `axisMin` should consider that value too?

DJTechnoo: Yes I was worried about that as well, I've been trying to come up with a better solution. Thanks, I'll look into `Axis.threshold` and see what happens 😅

pawelfus:
Closing since #13473 is is closed too

PR #14618, opened 2020-11-20
ts/networkgraph-series

bre1470:
Migrated networkgraph series to class pattern.

### DIFF
--- a/js/Series/Networkgraph/Networkgraph.js
+++ b/js/Series/Networkgraph/Networkgraph.js
@@ -10,6 +10,19 @@
  *
  * */
 'use strict';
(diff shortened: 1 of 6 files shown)


bre1470:
I will not go further. Types, implementations and inheritence do not fit.

bre1470:
Actually the layout concept makes no sense to me, because there is only one implementation everywhere and the draggable mixin is only used for that layout. So there are many unnecessary splits and naming of basically the same thing over and over again.

TorsteinHonsi (approved):
Thanks!

TorsteinHonsi:
@pawelfus @Izothep any comments about @bre1470 's layout refactor suggestion?

PR #14621, opened 2020-11-20
ts/chaikin-indicator

wchmiel:
Migrated chaikin indicator to the class pattern.


bre1470 (approved):
Thank you! Looks good. 👍

Issue #14584, opened 2020-11-17
Sankey chart, overlapping data labels link doesn't work


hihajdus:
#### Expected behavior
When overlapping is set to false data labels links should not be overlapped.
This feature works correctly in the 8.0.0 version.
Demo: [https://jsfiddle.net/BlackLabel/e0kxn56c/](https://jsfiddle.net/BlackLabel/e0kxn56c/)

#### Actual behaviour
Data labels links are overlapped when overlapping is set to false.

#### Live demo with steps to reproduce
https://jsfiddle.net/BlackLabel/8gx1yhtu/

#### Product version
Highcharts JS v8.2.2 (2020-10-22)

#### Affected browser(s)
All

pawelfus:
Bisected to commit 6dc7be2b8d

TorsteinHonsi:
Please see https://github.com/highcharts/highcharts/pull/14622 for a suggested fix.

PR #14622, opened 2020-11-20
bugfix: Sankey chart, overlapping data labels link doesn't work

murthyvitara:
Bug Fix: #14584 

bug: when series.nodes are present, series.points are ignored

fix: series.nodes and series.points are merged whenever available


TorsteinHonsi:
Thanks for the PR! However we need the fix to be applied to the TypeScript files, and preferrable tests to be added. I've linked to this PR in the issue, so we can pick it up and save time debugging/fixing it.

PR #14628, opened 2020-11-23
ts/natr-indicator

sebastianbochan:
Migrated natr indicator to the class pattern.


ts/Stock/Indicators/NATRIndicator.ts
```diff
@@ -1,143 +0,0 @@
...
-    {
-        tooltip: {
-            valueSuffix: '%'
-        }
-    },
-    /**
-     * @lends Highcharts.Series#
-     */
-    {
-        requiredIndicators: ['atr'],
```
bre1470, line 97: `requiredIndicator` is here set in the series factory, which indicates that as a prototype property it can not be part of the class pattern (because they become instance properties there). So you need to set the value in a following `extend` call.
```
extend(NATRIndicator.prototype, {
    requiredIndicators: ['atr']
});
```

ts/Stock/Indicators/NATR/NATRIndicator.ts
```diff
@@ -0,0 +1,156 @@
...
+
+/* *
+ *
+ *  Prototype Properties
+ *
+ * */
+interface NATRIndicator {
+    pointClass: typeof NATRPoint;
+}
+
```
bre1470, line 120: Here the extend call for the `requiredIndicators` values goes right after the interface.

bre1470 (changes requested):

ts/Stock/Indicators/NATRIndicator.ts
sebastianbochan: I was not aware of it, will change it!

Thanks!

sebastianbochan:
@bre1470 I hope that it is correct, now ;)

bre1470 (approved):
Thank you! Looks good. 👍

bre1470:
Please merge latest master to solve conflicts.

PR #14629, opened 2020-11-23
bugfix/14623-colorIndex-update

### LINKED ISSUE #14623: Marker colors do not update when series.colorIndex is changed via Chart.update or Series.update
#### Expected behaviour
Programmatically changing the `colorIndex` of a series would change the color of any markers as well as the path, line, etc, of the series.

#### Actual behaviour
The marker colors do not change and retain the className `highcharts-color-${x}` from before the update.

https://jsfiddle.net/josterman/g2tkwx9n/

Wait for the two timeouts to show the behavior.
#### Product version

Highcharts v8
#### Affected browser(s)
Observed on Chrome and Firefox, latest stable.

khlieng:
Fixed #14623, `Point.colorIndex` did not pick up `Series.colorIndex` updates.

### DIFF
--- a/js/Core/Series/Point.js
+++ b/js/Core/Series/Point.js
@@ -731,7 +731,7 @@ var Point = /** @class */ (function () {
         else {
             colorIndex = series.colorIndex;
         }
-        this.colorIndex = pick(this.colorIndex, colorIndex);
+        this.colorIndex = pick(colorIndex, this.colorIndex);
     };
     /**
      * Set a value in an object, on the property defined by key. The key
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (changes requested):
With this change, individual point `colorIndex` no longer works: https://jsfiddle.net/highcharts/5udkf3e1/.

I think what we need to do is to read the point _options_ instead of the property that has already been set:

```js
this.colorIndex = pick(this.options.colorIndex, colorIndex);
```

In addition, please add this case (point individual `colorIndex`) to the unit test, as it apparently isn't covered yet.

TorsteinHonsi (approved):

PR #14630, opened 2020-11-23
ts/aroonoscillator-indicator

madepiet:
Migrated aroonoscillator indicator to the class pattern.


bre1470 (approved):
Thank you! Looks good. 👍

PR #14635, opened 2020-11-23
ts/TEMA-indicator

karolkolodziej:


karolkolodziej:
I guess that this part also should be removed from the global namespace. @bre1470 where should I put that?

```
/**
 * Internal types
 * @private
 */
declare global {
    namespace Highcharts {
        interface EMAIndicatorLevelsObject {
            level1: number;
            level2: number;
            level3: number;
            prevLevel3: number;
        }
    }
}
```

bre1470:
> I guess that this part also should be removed from the global namespace. @bre1470 where should I put that?
> 
> ```
> /**
>  * Internal types
>  * @private
>  */
> declare global {
>     namespace Highcharts {
>         interface EMAIndicatorLevelsObject {
>             level1: number;
>             level2: number;
>             level3: number;
>             prevLevel3: number;
>         }
>     }
> }
> ```

Yes, that is correct. As the type is used in the indicator functions, it should go into a class namespace:
```ts
/* *
 *
 *  Class Namespace
 *
 * */
namespace EMAIndicator {
    export interface LevelsObject {
        level1: number;
        level2: number;
        level3: number;
        prevLevel3: number;
    }
}
```
The type references become `EMAIndicator.LevelsObject`.

bre1470 (approved):
Thank you! Looks good. 👍

PR #14647, opened 2020-11-24
bugfix/14624-vertical-panning-not-disablelable

### LINKED ISSUE #14624: panning.type = 'y' or 'xy' sometimes messes up aligned y-axes
#### Expected behaviour
Enabling or disabling y-axis panning should not impact y-axis tick positioning of aligned axes when the chart has not been panned and is fully zoomed out. Data in a series mapped to an automatically-sized axis should not be out of view.

#### Actual behaviour

If the panning type is "xy" or "y" and there are two or more y-axes, with automatic bounds but set to startOnTick and endOnTick, the extrema may not be calculated correctly, and data will be drawn off-screen. Whether or not they will be drawn correctly depends on the data and on the vertical space allocated to the chart.

Demo: https://jsfiddle.net/vjub85az/9/

Observe that the data from series y goes off the bottom of the chart; it should not.

Modify the values set in the first two lines to test different panning configurations. The data problem exists with panningType = "xy" or "y" but not "x". panningEnabled = true or false has no impact.

#### Product version
Highcharts JS v8.2.2

#### Affected browser(s)
Chrome, Firefox, and Safari; others not tested

DJTechnoo:
Fixed #14624, `panning.type` enabled panning while `panning.enabled = false`.

### DIFF
--- a/js/Core/Axis/Axis.js
+++ b/js/Core/Axis/Axis.js
@@ -3054,12 +3054,11 @@ var Axis = /** @class */ (function () {
     *
     */
     Axis.prototype.hasVerticalPanning = function () {
-        var _a, _b, _c, _d;
-        if (!((_b = (_a = this.chart.options.chart) === null || _a === void 0 ? void 0 : _a.panning) === null || _b === void 0 ? void 0 : _b.enabled)) {
-            // #14624
-            return false;
-        }
-        return /y/.test(((_d = (_c = this.chart.options.chart) === null || _c === void 0 ? void 0 : _c.panning) === null || _d === void 0 ? void 0 : _d.type) || '');
+        var _a;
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (changes requested):
Thanks!

As you can see from the transpiled `Axis.js`, and the file size report, the repeated optional chaining produces a lot of junk code in JS. I'd suggest somthing more optimized for bytesize, for example:

```js
    public hasVerticalPanning(): boolean {
        const panningOptions = this.chart.options.chart?.panning;
        return Boolean(
            panningOptions &&
            panningOptions.enabled &&
            /y/.test(panningOptions.type)
        );
    }
```

TorsteinHonsi (approved):
Thanks, looks good!

PR #14648, opened 2020-11-24
ts/pc-indicator

madepiet:
Migration PCIndicator to the class pattern.

### DIFF
--- a/ts/Stock/Indicators/PC/PCIndicator.ts
+++ b/ts/Stock/Indicators/PC/PCIndicator.ts
@@ -8,51 +8,27 @@
 
 'use strict';
 
-import type CSSObject from '../../Core/Renderer/CSSObject';
-import type IndicatorValuesObject from './IndicatorValuesObject';
-import type LineSeries from '../../Series/Line/LineSeries';
+import type IndicatorValuesObject from '../IndicatorValuesObject';
+import type LineSeries from '../../../Series/Line/LineSeries';
 const {
     seriesTypes: {
         sma: SMAIndicator
     }
 } = BaseSeries;
 import type {
-    SMAOptions,
-    SMAParamsOptions
-} from './SMA/SMAOptions';
-import type SMAPoint from './SMA/SMAPoint';
-import BaseSeries from '../../Core/Series/Series.js';
-import palette from '../../Core/Color/Palette.js';
-import MultipleLinesMixin from '../../Mixins/MultipleLines.js';
-import ReduceArrayMixin from '../../Mixins/ReduceArray.js';
-import U from '../../Core/Utilities.js';
+    PCOptions,
+    PCParamsOptions
+} from '../PC/PCOptions';
+import type PCPoint from './PCPoint';
+import BaseSeries from '../../../Core/Series/Series.js';
+import palette from '../../../Core/Color/Palette.js';
+import MultipleLinesMixin from '../../../Mixins/MultipleLines.js';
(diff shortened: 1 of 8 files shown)


ts/Stock/Indicators/PC/PCIndicator.ts
```diff
@@ -0,0 +1,239 @@
...
+    seriesTypes: {
+        sma: SMAIndicator
+    }
+} = BaseSeries;
+import type {
+    PCOptions,
+    PCParamsOptions
+} from '../PC/PCOptions';
+import type PCPoint from './PCPoint';
+import BaseSeries from '../../../Core/Series/Series.js';
```
bre1470, line 23: The deconstruction of seriesTypes should be near the BaseSeries import.

ts/Stock/Indicators/PC/PCIndicator.ts
```diff
@@ -0,0 +1,239 @@
...
+    pointValKey: string;
+}
+
+extend(PCIndicator.prototype, merge(MultipleLinesMixin, {
+    pointArrayMap: ['top', 'middle', 'bottom'],
+    pointValKey: 'middle',
+    nameBase: 'Price Channel',
+    nameComponents: ['period'],
+    linesApiNames: ['topLine', 'bottomLine']
+}));
```
bre1470, line 199: The mixin has to be manually assigned, to make it clear for future maintenance, what actually gets replaced in the class pattern. So it should become:
```ts
extend(PCIndicator.prototype, {
    getTranslatedLinesNames: MultipleLinesMixin.getTranslatedLinesNames,
    …
    translate: MultipleLinesMixin.translate
});
```
As you will notice, some values/functions like `linesApiNames` are actually not taken from the mixin like the interface suggests, but are redefined in the class pattern or the extend call.

ts/Stock/Indicators/PC/PCOptions.d.ts
```diff
@@ -0,0 +1,40 @@
...
+    SMAParamsOptions
+} from '../SMA/SMAOptions';
+
+/* *
+*
+*  Declarations
+*
+* */
+export interface PCOptions extends SMAOptions,
+    Highcharts.MultipleLinesIndicatorOptions {
```
bre1470, line 25: This can and should be on one line.

bre1470 (changes requested):

ts/Stock/Indicators/PC/PCIndicator.ts
madepiet: So, could you please suggest how to deal with these linesApiNames? 
Since leaving such a configuration causes errors:
```ts
extend(PCIndicator.prototype, {
    getTranslatedLinesNames: MultipleLinesMixin.getTranslatedLinesNames,
    pointArrayMap: ['top', 'middle', 'bottom'],
    pointValKey: 'middle',
    nameBase: 'Price Channel',
    nameComponents: ['period'],
    linesApiNames: ['topLine', 'bottomLine'],
    translate: MultipleLinesMixin.translate
});
```

bre1470: You have to assign all the parts of the mixin, that are not defined in the PCIndicator class pattern or extend call. As far as I can see, it is missing: `drawGraph`, `toYData`.

bre1470 (approved):
Thank you! Looks good. 👍

PR #14651, opened 2020-11-24
bugfix/14649-flag-update-visuals

### LINKED ISSUE #14649: Update flag fillColor
#### Expected behaviour
In a flag array, I change the portion of the same flag array that includes the fill color method as well. When I change the portion of the array everything should be updated including the fill color.

#### Actual behaviour
In a flag array, I change the portion of the same flag array that includes the fill color method as well. When I change the portion of the array everything updated except the fill color. The fill color change only when I hover the mouse over each and every changed flag. Do you have a method to change the fill color without hovering each flag when I change the portion of the flag array?

Here the JSFiddle regarding my concern. Here initially all flags are in green color, then when clicks on the button "Change Flag" the first two buttons change their title from A to B but the fillColor does not change. But in the array, I change the fillColor to blue. BTW when hovering the flags the color change to blue.

https://jsfiddle.net/munasinghecm/32qLdhr7/36/

How can I change the fillColor when the button click event happens without hovering over flags?
#### Product version

highcharts-angular "2.8.2"
highcharts: "8.2.2"

khlieng:
Fixed #14649, dynamically updating visual attributes on points in flag series did not work.

### DIFF
--- a/js/Series/Flags/FlagsSeries.js
+++ b/js/Series/Flags/FlagsSeries.js
@@ -110,28 +110,26 @@ var FlagsSeries = /** @class */ (function (_super) {
                 !outsideRight) {
                 // Create the flag
                 if (!graphic) {
-                    graphic = point.graphic = renderer.label('', null, null, shape, null, null, options.useHTML);
-                    if (!chart.styledMode) {
-                        graphic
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (approved):
Thanks!

PR #14655, opened 2020-11-24
ts/psar-indicator

sebastianbochan:
Migrated psar indicator to the class pattern.


bre1470 (approved):
Thank you! Looks good. 👍

PR #14656, opened 2020-11-24
ts/zigzag-indicator

sebastianbochan:
Migrated zigzag indicator to the class pattern.


sebastianbochan:
Correct PR: https://github.com/highcharts/highcharts/pull/14657

PR #14657, opened 2020-11-24
ts/zigzag-indicator-fixed

sebastianbochan:
Migrated zigzag indicator to the class pattern.


bre1470 (approved):
Thank you! Looks good. 👍

PR #14659, opened 2020-11-25
bugfix/14637-line-pointPlacement-between

### LINKED ISSUE #14637: Regression issue with series.pointPlacement since v8.0
#### Expected behaviour
when pointPlacement is set to "between" for line series, point marker should be placed between the ticks on the category axis.

#### Actual behaviour
when pointPlacement is set to "between" for line series, point markers are placed on the tick.

https://jsfiddle.net/praveenx9/p2L597tz/48/

#### Product version
Highcharts 8.0

#### Affected browser(s)
Chrome

khlieng:
Fixed #14637, setting `pointPlacement` to `between` placed line series points on ticks instead of between them.


TorsteinHonsi (approved):
Thanks!

PR #14664, opened 2020-11-25
ts/lollipop-series 🍭

goransle:
Migrated `lollipop` series to class pattern

### DIFF
--- a/ts/Series/Dumbbell/DumbbellPoint.ts
+++ b/ts/Series/Dumbbell/DumbbellPoint.ts
@@ -44,7 +44,6 @@ class DumbbellPoint extends AreaRangePoint {
     public options: DumbbellPointOptions = void 0 as any;
     public connector: Highcharts.SVGElement = void 0 as any;
     public pointWidth: number = void 0 as any;
-    public pointSetState: typeof AreaRangePoint.prototype.setState = super.setState;
 
     /* *
      *
@@ -115,6 +114,19 @@ class DumbbellPoint extends AreaRangePoint {
         point.connector[verb](series.getConnectorAttribs(point));
     }
 }
+
+/* *
+ *
+ *  Prototype properties
+ *
+ * */
+interface DumbbellPoint{
+    pointSetState: typeof AreaRangePoint.prototype.setState;
+}
+
+extend(DumbbellPoint.prototype, {
+    pointSetState: AreaRangePoint.prototype.setState
+});
 
 /* *
  *
--- a/ts/Series/Lollipop/LollipopPoint.ts
+++ b/ts/Series/Lollipop/LollipopPoint.ts
@@ -23,11 +23,19 @@ const {
     seriesTypes: {
         area: {
             prototype: areaProto
+        },
+        dumbbell: {
+            prototype: {
+                pointClass: DumbbellPoint
+            }
+        },
+        line: {
+            prototype: {
+                pointClass: Point
+            }
         }
     }
 } = BaseSeries;
-import DumbbellPoint from '../Dumbbell/DumbbellPoint.js';
-import Point from '../../Core/Series/Point.js';
 import U from '../../Core/Utilities.js';
 const {
     isObject,
@@ -63,7 +71,7 @@ extend(LollipopPoint.prototype, {
         series: LollipopSeries,
         options: LollipopPointOptions,
         x?: number
-    ): Point {
+    ): typeof Point.prototype {
         if (isObject(options) && 'low' in options) {
             options.y = options.low;
             delete options.low;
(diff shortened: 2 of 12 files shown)


ts/Series/Dumbbell/DumbbellPoint.ts
```diff
@@ -116,6 +115,19 @@ class DumbbellPoint extends AreaRangePoint {
...
+ *
+ * */
+interface DumbbellPoint{
+    pointSetState: typeof AreaRangePoint.prototype.setState;
+}
+
+extend(DumbbellPoint.prototype, {
+    pointSetState: AreaRangePoint.prototype.setState
+});
+
```
bre1470, line 130: DumbbellPoint is extending now AreaRangePoint, so you can actually simply delete it. It was in the factory, because there all point prototypes inherited from the base Point class.

ts/Series/Lollipop/LollipopSeries.ts
```diff
@@ -0,0 +1,253 @@
...
+/* *
+ *
+ *  Imports
+ *
+ * */
+
+import type LollipopSeriesOptions from './LollipopSeriesOptions';
+import LollipopPoint from './LollipopPoint.js';
+import AreaSeries from '../Area/AreaSeries.js';
+const { prototype: areaProto } = AreaSeries;
```
bre1470, line 22: Please use the deconstruction on BaseSeries for all types: Area, Dumbbell and Column.

ts/Series/Lollipop/LollipopSeries.ts
```diff
@@ -0,0 +1,253 @@
...
+ * */
+
+interface LollipopSeries extends DumbbellSeries {
+    pointClass: typeof LollipopPoint;
+    pointArrayMap: Array<string>;
+    pointValKey: string;
+    translatePoint: AreaSeries['translate'];
+    drawPoint: AreaSeries['drawPoints'];
+    drawDataLabels: ColumnSeries['drawDataLabels'];
+    setShapeArgs: ColumnSeries['translate'];
```
bre1470, line 130: This will become `typeof Series.prototype.prop`.

bre1470 (changes requested):

ts/Series/Dumbbell/DumbbellPoint.ts
goransle: AreaRangePoint does not have a `pointSetState` as far as I can see? But we could just call super.setState in the dumbbellPoint setstate, which would be cleaner.

goransle: (which I now remember becomes problematic, as LollipopPoint overwrites pointSetState again)

bre1470: Oh, my mistake. It is a different name. I like your idea, but that is kind of refactoring outside of the task. So just ignore my request - it is nonsense. 😅

bre1470 (changes requested):
Sorry, more things.

ts/Series/Lollipop/LollipopSeries.ts
```diff
@@ -0,0 +1,259 @@
...
+    drawDataLabels: typeof colProto['drawDataLabels'];
+    setShapeArgs: typeof colProto['translate'];
+}
+
+extend(LollipopSeries.prototype, {
+    pointArrayMap: ['y'],
+    pointValKey: 'y',
+    toYData: function (point: LollipopPoint): Array<number> {
+        return [pick(point.y, point.low)];
+    },
```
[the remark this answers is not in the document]
goransle: What should I do about the return type? The inherited type is a `[number, number]` tuple. Should I just `any`-cast this or change the type in AreaRangeSeries?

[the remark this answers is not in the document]
bre1470: It is better for inheritance to avoid tuples, otherwise you have to specify all the different tuple sizes down the road, even so in its core it is just an array.

ts/Series/Lollipop/LollipopPoint.ts
```diff
@@ -63,7 +71,7 @@ extend(LollipopPoint.prototype, {
         series: LollipopSeries,
         options: LollipopPointOptions,
         x?: number
-    ): Point {
+    ): typeof Point.prototype {
```
bre1470, line 74: In case this is the only usage of the Point class in the file, it is better to not use deconstruction, but an `import type`. Sorry for the confusion.

bre1470 (changes requested):
Thank you! One more thing...

ts/Series/Lollipop/LollipopPoint.ts
goransle: This function also calls `Point.prototype.init` :)

bre1470: Yes, okay. Well then everything is settled. 😅

bre1470 (approved):
Thank you! Looks good. 👍

PR #14665, opened 2020-11-25
bugfix/14315-column-empty-extremes

### LINKED ISSUE #14315: Setting extremes on the histogram chart with no data to display generates an error in the console
#### Expected behaviour
Being able to set the extremes without error.

#### Actual behaviour
When the extremes for the histogram axis such that there is no data there to display, errors in console occurs.
```
"highcharts.src.js:3636 Error: <rect> attribute y: Expected length, "NaN"."
"highcharts.src.js:3636 Error: <rect> attribute height: Expected length, "NaN"."
```

Click the button below the chart and check the console.
https://jsfiddle.net/BlackLabel/9yv5j6bq/

#### Product version
Highcharts JS v8.2.0

#### Affected browser(s)
All

khlieng:
Fixed #14315, setting X axis extremes outside the chart series' data range threw.

This regression happened because `clamp(undefined, -1e5, 1e5)` returns `-1e5` while `Math.min(Math.max(-1e5, undefined), 1e5)` returns `NaN` and `ColumnSeries.drawPoints` uses `isNumber(plotY)` to determine if it should draw.

### DIFF
--- a/ts/Series/Line/LineSeries.ts
+++ b/ts/Series/Line/LineSeries.ts
@@ -4665,7 +4665,7 @@ class LineSeries {
             // Set the the plotY value, reset it for redraws
             // #3201
             point.plotY = void 0;
-            if (typeof yValue === 'number' && yValue !== Infinity) {
+            if (isNumber(yValue)) {
                 const translated = yAxis.translate(
                     yValue, false, true, false, true
                 );
--- a/ts/Series/Line/LineSeries.ts
+++ b/ts/Series/Line/LineSeries.ts
@@ -4664,13 +4664,15 @@ class LineSeries {
 
             // Set the the plotY value, reset it for redraws
             // #3201
-            point.plotY = (
-                (typeof yValue === 'number' && yValue !== Infinity) ?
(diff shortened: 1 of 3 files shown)


ts/Series/Line/LineSeries.ts
```diff
@@ -4664,13 +4664,15 @@ class LineSeries {
...
             // #3201
-            point.plotY = (
-                (typeof yValue === 'number' && yValue !== Infinity) ?
-                    limitedRange(yAxis.translate(
-                        yValue, 0 as any, 1 as any, 0 as any, 1 as any
-                    ) as any) :
-                    void 0
-            );
+            point.plotY = void 0;
+            if (typeof yValue === 'number' && yValue !== Infinity) {
```
TorsteinHonsi, line 4668: Use the `isNumber` utility function that checks that the type is number, that it is not Infinity or -Infinity, and not NaN.

TorsteinHonsi (changes requested):

TorsteinHonsi (approved):
Thanks!

PR #14670, opened 2020-11-25
bugfix/14246-pie-legend-click

### LINKED ISSUE #14246: Pie, ignoreHiddenPoint set to false - point doesn't show up when clicking legendItem
In pie, when `series.ignoreHiddenPoint` is set to `false`, clicking twice on the same legendItem doesn't show the point again.

https://jsfiddle.net/BlackLabel/f1zjL86t

#### Product version
8.2.0
Works fine in 7.1.2, but it's broken since 7.1.3

khlieng:
Fixed #14246, pie point stayed hidden after first legend item click when `ignoreHiddenPoint` was set to `false`.


TorsteinHonsi (approved):
Thanks!

PR #14672, opened 2020-11-26
bugfix/14147-xRange-center-toolTipPosition

### LINKED ISSUE #14147: xrange chart with more than one series tooltip position
#### Expected behaviour
tooltip should appear above the xrange point in the middle

#### Actual behaviour
the tooltip does not appear in the center

https://jsfiddle.net/L5o21pyt/2/
image

#### Product version
Highcharts

#### Affected browser(s)
all

K009:
Fixed #14147, in xrange/gantt series, the tooltip wasn't centered above the point if there were more than one series in the chart.

~~Fixed #14147 tooltip wasn't centered.~~

### DIFF
--- a/js/Series/XRangeSeries.js
+++ b/js/Series/XRangeSeries.js
@@ -251,7 +251,8 @@ BaseSeries.seriesType('xrange', 'column'
      * @param {Highcharts.Point} point
      */
     translatePoint: function (point) {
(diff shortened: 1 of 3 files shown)


K009:
In visual tests, the position of the tooltip is moved by probably one pixel, so I would say that it's acceptable. Also, the tooltip is displayed one time from the left and the other time from the top because of the blue point that is out of our visible area. In my mind, it looks good, but I'm open to any suggestions.

TorsteinHonsi (approved):
Thanks, looks good!

I'm doing some changes to the description and tagging, for future PRs make sure to comply to the [guidelines for the changelog](https://github.com/highcharts/highcharts/tree/master/changelog#writing-content-for-the-changelog).

PR #14673, opened 2020-11-26
ts/vwap-indicator

sebastianbochan:
Migrated vwap indicator to the class pattern.


bre1470 (approved):
Thank you! Looks good. 👍

PR #14674, opened 2020-11-26
Refactored events to handle class inheritance

### LINKED ISSUE #14661: Events don't work with extended classes
#### Expected behaviour
Events should work when added anywhere in the inheritance chain including on the instance

#### Actual behaviour
Events added to children behave flimsy. As a consequence, for example an event added to `ScatterSeries` would also fire on the `LineSeries`, which is its parent. 

1. When adding an event on an inherited class, firing the event on the parent also causes the child event to run: http://jsfiddle.net/highcharts/h1k6prwa/
2. When adding an event to first the inherited class then the parent class, only the inherited class' event handler runs: http://jsfiddle.net/highcharts/h1k6prwa/1/

#### Product version
v8.2.2

#### Affected browser(s)
All

TorsteinHonsi:
Fixed #14661, the event system didn't properly work with class inheritance. For example an event set on the inherited `ColumnSeries` class, would also fire on the parent `LineSeries`.

### To do
 - [x] Implement events in the inheritance chain
 - [x] Add unit tests for the new functionality
 - [x] Make use of the new functionality in cases where we currently set an event on the `LineSeries` then do an `instanceof` check within the handler. That should not be needed anymore.
 - [x] IE8 polyfill of `getPrototypeOf`

### DIFF
--- a/js/Core/Dynamics.js
+++ b/js/Core/Dynamics.js
@@ -1136,6 +1136,8 @@ extend(LineSeries.prototype, /** @lends Series.prototype */ {
         }
         if (seriesTypes[newType || initialType]) {
             extend(series, seriesTypes[newType || initialType].prototype);
+            // The events are tied to the prototype chain, don't copy
+            delete series.hcEvents;
         }
         else {
             error(17, true, chart, { missingModuleFor: (newType || initialType) });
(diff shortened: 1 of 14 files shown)


pawelfus (approved):
Nothing to add, nice! 👍

PR #14676, opened 2020-11-26
ts/ppo-indicator

madepiet:
Migration ppo indicator to the class pattern.


ts/Stock/Indicators/PPO/PPOIndicator.ts
```diff
@@ -0,0 +1,243 @@
...
+}
+
+/* *
+*
+*   Prototype Properties
+*
+* */
+
+interface PPOIndicator {
+    options: PPOOptions;
```
bre1470, line 194: `options` should got into the class pattern.

bre1470 (changes requested):
Thank you! One requested change is needed.

bre1470 (approved):
Thank you! Looks good. 👍

PR #14680, opened 2020-11-26
ts/trendline-indicator

madepiet:
Migration trendline indicator to the class pattern.


bre1470 (approved):
Thank you! Looks good. 👍

PR #14681, opened 2020-11-26
Auto compute heading level for a11y description.

oysteinmoseng:
Auto compute heading level for the pre-chart accessibility description.
___
The a11y description inserted before the chart has by default used an `h5` heading. This often comes up in audits, as chart authors tend not to configure it differently regardless of the page structure.

The a11y module now looks for previous headings in siblings/parents/parent siblings, and sets the level accordingly (`h2` to `h6`). A `p` tag is used if no headings are found. In some cases, such as where the headings are nested under `div`s for layout purposes, headings will not be detected.

It is still possible to manually set the heading level as before in [`accessibility.screenReaderSection.beforeChartFormat`](https://api.highcharts.com/highcharts/accessibility.screenReaderSection.beforeChartFormat). Existing charts where this has already been configured will not change behavior.


TorsteinHonsi (changes requested):
Thanks, nice feature! Please resolve the conflict.

TorsteinHonsi (approved):

PR #14686, opened 2020-11-27
ts/columnrange-series

goransle:
Migrated columnrange series to class pattern

### DIFF
--- a/ts/Series/ColumnRange/ColumnRangePoint.ts
+++ b/ts/Series/ColumnRange/ColumnRangePoint.ts
@@ -1,0 +1,61 @@
+/* *
+ *
+ *  (c) 2010-2020 Torstein Honsi
+ *
+ *  License: www.highcharts.com/license
+ *
+ *  !!!!!!! SOURCE GETS TRANSPILED BY TYPESCRIPT. EDIT TS FILE ONLY. !!!!!!!
+ *
+ * */
+
+'use strict';
+
+/* *
+ *
+ *  Imports
+ *
+ * */
+
+import type ColumnRangePointOptions from './ColumnRangePointOptions.js';
+import type ColumnRangeSeries from './ColumnRangeSeries.js';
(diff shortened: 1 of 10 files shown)


ts/Series/ColumnRange/ColumnRangePoint.ts
```diff
@@ -0,0 +1,61 @@
...
+    barX: typeof ColumnPoint.prototype['barX'];
+    pointWidth: typeof ColumnPoint.prototype['pointWidth'];
+    shapeArgs: SVGAttributes;
+    shapeType: typeof ColumnPoint.prototype['shapeType'];
+}
+class ColumnRangePoint extends AreaRangePoint {
+    public series: ColumnRangeSeries = void 0 as any;
+    public options: ColumnRangePointOptions = void 0 as any;
+    public setState = ColumnPoint.prototype.setState;
+}
```
bre1470, line 54: That here is misleading, because this time it is a real point class. So it should be just the class pattern instead of declaration with an interface, I believe.

ts/Series/ColumnRange/ColumnRangePoint.ts
```diff
@@ -0,0 +1,61 @@
...
+interface ColumnRangePoint {
+    barX: typeof ColumnPoint.prototype['barX'];
+    pointWidth: typeof ColumnPoint.prototype['pointWidth'];
+    shapeArgs: SVGAttributes;
+    shapeType: typeof ColumnPoint.prototype['shapeType'];
+}
+class ColumnRangePoint extends AreaRangePoint {
+    public series: ColumnRangeSeries = void 0 as any;
+    public options: ColumnRangePointOptions = void 0 as any;
+    public setState = ColumnPoint.prototype.setState;
```
bre1470, line 53: `setState` is a function assignment and should go into an extend call (prototype property).

bre1470 (changes requested):

bre1470 (approved):
Thank you! Looks good. 👍

PR #14687, opened 2020-11-27
max-len-tests-fix

K009:
Splitted too long lines into the shorter ones.


TorsteinHonsi (approved):

TorsteinHonsi:
If a PR like this is blocking us from doing another (important) fix, it is better to do the whitespace change along with the other PR.

PR #14693, opened 2020-11-27
ts/tilemap-series

bre1470:
Migrated tilemap series to class pattern.


bre1470:
I can not see a difference in the last unsolved visual diff. @TorsteinHonsi How to proceed?

TorsteinHonsi:
@bre1470 The difference is in the overlap between the hovered tile and the one above it, as indicated by the red pixels in the "Computed difference" graphic. If you toggle "Show file diff", you'll see the change in the SVG. Apparently the candidate has a `data-z-index` attribute that the reference doesn't have. That also explains why the `<path>` element is in a different place in the DOM, and why there is a visual difference in the overlap. The same difference can also be observed in the utils.

TorsteinHonsi (changes requested):
@bre1470 The difference is in the overlap between the hovered tile and the one above it, as indicated by the red pixels in the "Computed difference" graphic. If you toggle "Show file diff", you'll see the change in the SVG. Apparently the candidate has a data-z-index attribute that the reference doesn't have. That also explains why the <path> element is in a different place in the DOM, and why there is a visual difference in the overlap. The same difference can also be observed in the utils.

TorsteinHonsi (approved):
Thanks!

PR #14694, opened 2020-11-28
Fix getTimeTicks: ensure min value of 1 for setUTCDate

robbertbrak:
Same as https://github.com/highcharts/highcharts/pull/9344, but for a different branch of the `getTimeTicks` function.

See http://jsfiddle.net/xjrop93q/ for a reproduction scenario. The chart width must be less than about 1200px for the error behavior to occur:
image


TorsteinHonsi:
Thanks! We'll need to apply this to the TypeScript source files, verify the idea and set up regression tests. I'm adding it to the backlog.

PR #14698, opened 2020-11-28
fix the linkedTo series is not 1st series issue

cksiow:
fix the linkedTo series is not 1st series issue


TorsteinHonsi:
Thanks @cksiow 

@pawelfus can you or your team have a look at this? And if it is valid, prepare a PR according to guidelines?

pawelfus:
Yes, it's a bug @karolkolodziej will take this from this point

PR #14707, opened 2020-11-30
ts/pareto-series

goransle:
Migrated `pareto` series to class pattern

### DIFF
--- a/ts/Series/ParetoSeries/ParetoSeries.ts
+++ b/ts/Series/ParetoSeries/ParetoSeries.ts
@@ -16,10 +16,8 @@
  *
  * */
 
-import type LinePoint from '../Line/LinePoint';
-import type LinePointOptions from '../Line/LinePointOptions';
-import type LineSeriesOptions from '../Line/LineSeriesOptions';
-import type { SeriesStatesOptions } from '../../Core/Series/SeriesOptions';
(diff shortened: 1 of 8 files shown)


ts/Series/ParetoSeries/ParetoSeries.ts
```diff
@@ -0,0 +1,278 @@
...
+    merge(DerivedSeriesMixin, {
+        /* eslint-disable no-invalid-this, valid-jsdoc */
+        /**
+         * Calculate sum and return percent points.
+         *
+         * @private
+         * @function Highcharts.Series#setDerivedData
+         * @requires modules/pareto
+         */
+        setDerivedData: (function (this: ParetoSeries): void {
```
bre1470, line 197: To avoid confusion during future refactoring, convert it to manual mixin apply. See https://github.com/highcharts/highcharts/pull/14719 for procedure.

bre1470 (changes requested):
Thank you! Please see comment for requested change.

ts/Series/ParetoSeries/ParetoSeries.ts
bre1470: After that you can move the custom implementation setDerivedData to the class pattern.

bre1470 (approved):
Thank you! Looks good. 👍

PR #14708, opened 2020-11-30
ts/sunburst-series

bre1470:
Migrated sunburst series to class pattern.


bre1470:
Since SunburstPoint is now inheriting from TreemapPoint (and ScatterPoint/LinePoint/Point), things work a bit different.
But I say that the differences have become so marginal, that I would call the job done. What do you think, @TorsteinHonsi ?

TorsteinHonsi:
Yes, for the visual side of it, it's fine!

TorsteinHonsi (approved):
Thanks!

PR #14711, opened 2020-11-30
ts/gauge-series

goransle:
Migrated `gauge` series to class pattern

### DIFF
--- a/ts/Series/Gauge/GaugeSeries.ts
+++ b/ts/Series/Gauge/GaugeSeries.ts
@@ -16,30 +16,30 @@
  *
  * */
 
-import type ColorType from '../Core/Color/ColorType';
-import type LinePoint from './Line/LinePoint';
-import type LinePointOptions from './Line/LinePointOptions';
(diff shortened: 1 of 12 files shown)


ts/Series/Gauge/GaugeSeries.ts
```diff
@@ -0,0 +1,735 @@
...
+     * */
+
+    public data: Array<GaugePoint> = void 0 as any;
+    public points: Array<GaugePoint> = void 0 as any;
+    public options: GaugeSeriesOptions = void 0 as any;
+
+    public yAxis: RadialAxis = void 0 as any;
+    public pivot?: SVGElement;
+    public fixedBox: boolean = true;
+    public forceDL: boolean = true;
```
bre1470, line 388: `fixedBox` and `forceDL` are prototype properties (set in extend call) and need to be moved to the interface.

bre1470 (changes requested):
Thank you! Please see comment for requested change.

bre1470 (approved):
Thank you! Looks good. 👍

PR #14713, opened 2020-11-30
ts/gantt-series

bre1470:
Migrated gantt series to class pattern.


bre1470:
merge conflict makes no sense - is in sync

TorsteinHonsi (approved):
Thanks!

PR #14714, opened 2020-11-30
ts/waterfall-series

marvin19:
Migration of WaterfallSeries.


ts/Series/Waterfall/WaterfallAxis.ts
```diff
@@ -0,0 +1,246 @@
...
+ * @private
+ */
+interface WaterfallAxis extends Axis {
+    waterfall: WaterfallAxis.Composition;
+}
+
+/**
+ * @private
+ */
+namespace WaterfallAxis {
```
bre1470, line 49: This should be moved to Core/Axis.

bre1470 (changes requested):
Thank you! I forgot one thing...

bre1470 (approved):
Thank you! Looks good. 👍

PR #14717, opened 2020-11-30
enhancement/rangeselector-dropdown

khlieng:
- [x] Button collapsing
- [x] Select
- [x] Make button animation work again
- [x] Fix padding visual diffs
- [x] Unit tests
- [x] Docs
- [x] Visual sample
- [x] Find out whats up with Firefox on CI
- [x] Accessibility


ts/Extensions/RangeSelector.ts
```diff
@@ -2164,18 +2289,174 @@ class RangeSelector {
...
+        this.positionButtons();
+        this.showDropdown();
+    }
+
+    /**
+     * @private
+     * @function Highcharts.RangeSelector#showButtons
+     * @return {void}
+     */
+    public showButtons(): void {
```
TorsteinHonsi, line 2373: Is this the counterpart to `collapseButtons`? If so, maybe a better suited name is `expandButtons`?

TorsteinHonsi (changes requested):

* Consider http://utils.highcharts.local/samples/#view/stock/demo/area. The 3M button is preselected. Now if we grab the left handle and drag left, the dropdown shows 1M in an unselected state. Until the visible range fits the range of the 6M button, then it falls back to 1M etc. This is somewhat counterintuitive. I suggest we create a disabled placeholder option first, using `lang.zoomText` as the text content. See https://jsfiddle.net/highcharts/ze03cLjd/
* Click events on the buttons are not triggered by the dropdown: https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/stock/rangeselector/button-click/ . Currently, in the `renderButtons` function, the event handler of the buttons checks for click events before running `clickButton`. We need either to extract this function, or move the whole event handling to the `clickButton` function.
* The dropdown does not appear when inputs are enabled so that the button row exceeds the chart width: https://jsfiddle.net/highcharts/wtuybdeg/ . Perhaps not a very important use case, but on the other hand it may be simple to fix by taking the max available width into account when determining the responsive option.

ts/Extensions/RangeSelector.ts
```diff
@@ -1807,16 +1833,58 @@ class RangeSelector {
...
+            options
+        } = this;
         const lang = defaultOptions.lang;
-        const renderer = this.chart.renderer;
-        const options = this.options;
-        const buttons = this.buttons;
-        const buttonTheme = options.buttonTheme;
+        const renderer = chart.renderer;
+
+        const buttonTheme = merge(options.buttonTheme);
```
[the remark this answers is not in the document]
khlieng: Are you sure? According to my editor its able to infer the type.

ts/Core/Renderer/SVG/SVGLabel.ts
```diff
@@ -434,7 +429,7 @@ class SVGLabel extends SVGElement {
             bBox.height || Infinity
         );
 
-        if (this.needsBox) {
+        if (this.needsBox && (!this.renderer.styledMode || this.text.added)) {
```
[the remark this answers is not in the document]
khlieng: Maybe I should just get rid of it and move the crisping outside the `!this.box` block instead so that it handles `stroke-width` changes as well?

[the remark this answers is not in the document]
TorsteinHonsi: Yes I still don't understand the case, but it seems wrong to add more to the condition when `this.needsBox` is itself a condition.

ts/Extensions/RangeSelector.ts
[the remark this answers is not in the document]
TorsteinHonsi: Yes, for me to. I could have sworn it was `any` yesterday... 🤔

ts/Core/Renderer/SVG/SVGLabel.ts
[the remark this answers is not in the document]
khlieng: The case is that `stroke-width` needs to be known for crisping to work, and in styled mode you dont know until the text is added to the DOM, so I added that condition since the crisping was set only when creating the box. And this problem was discovered when I made the button width actually update when setting padding.

samples/stock/rangeselector/dropdown/demo.js
```diff
@@ -0,0 +1,41 @@
...
+                dropdown
+            }
+        });
+    }
+});
+
+document.getElementById('width').addEventListener('click', e => {
+    const width = e.target.dataset.width;
+    if (width) {
+        document.getElementById('container').style.width = width;
```
pawelfus, line 38: I'm just curious here: why not using `chart.setSize()`?

ts/Core/Renderer/SVG/SVGLabel.ts
```diff
@@ -243,10 +250,11 @@ class SVGLabel extends SVGElement {
     public getBBox(): BBoxObject {
         const bBox = this.bBox;
         const padding = this.padding;
+        const paddingLeft = this.paddingLeft || padding;
```
pawelfus, line 253: Use `pick()` instead of `||`. Otherwise, when `this.paddingLeft = 0` we will user `this.padding`

ts/Core/Renderer/SVG/SVGLabel.ts
```diff
@@ -463,20 +458,19 @@ class SVGLabel extends SVGElement {
...
      */
-    private updateTextPadding(): void {
+    public updateTextPadding(): void {
         const text = this.text;
 
         // Determine y based on the baseline
         const textY = this.baseline ? 0 : this.baselineOffset;
 
-        let textX = this.paddingLeft + this.padding;
+        let textX = this.paddingLeft || this.padding;
```
pawelfus, line 473: Use `pick()`

ts/Core/Renderer/SVG/SVGLabel.ts
```diff
@@ -512,11 +507,17 @@ class SVGLabel extends SVGElement {
         this.widthSetting = isNumber(value) ? value : void 0;
     }
 
+    public getPaddedWidth(): number {
+        const padding = this.padding;
+        const paddingLeft = this.paddingLeft || padding;
+        const paddingRight = this.paddingRight || padding;
```
pawelfus, line 513: Use `pick()`

ts/Extensions/RangeSelector.ts
```diff
@@ -106,6 +105,7 @@ declare global {
             buttons?: Array<RangeSelectorButtonsOptions>;
             buttonSpacing: number;
             buttonTheme: SVGAttributes;
+            dropdown: 'always'|'never'|'responsive';
```
pawelfus, line 108: Could we also define a new type? For example `RangeSelectorDropdownValue`

ts/Extensions/RangeSelector.ts
pawelfus: A little nitpicking here, but.. `showButtons` and `collapseButtons` are actually not only hiding/showing buttons, but also hide logic to show/hide the dropdown. Could we untangle this? What do you think?

Just in case, I assume `zoomText` as part of buttons and should be kept here

PS: Function description is missing in the doclet.

ts/Extensions/RangeSelector.ts
```diff
@@ -2164,18 +2289,174 @@ class RangeSelector {
...
+    }
+
+    /**
+     * Position the select element on top of the button.
+     *
+     * @private
+     * @function Highcharts.RangeSelector#showDropdown
+     * @return {void}
+     */
+    public showDropdown(): void {
```
pawelfus, line 2426: How about `positionDropdown` instead of `showDropdown`?

pawelfus: Nevermind, I see now how it's hidden :)

pawelfus (changes requested):
Just a few minor things. Overall, great work!

samples/stock/rangeselector/dropdown/demo.js
khlieng: It was just the first thing that came into my mind.

ts/Extensions/RangeSelector.ts
khlieng: I think they should stay tangled, they dont make much sense on their own and then you would have 2 function calls everywhere.

ts/Extensions/RangeSelector.ts
khlieng: I think they should stay tangled, they dont make much sense on their own and then you would have 2 function calls everywhere.

ts/Core/Renderer/SVG/SVGLabel.ts
khlieng: What would you pass to `attr()` to clear it in this case, some string like `auto` and then check for that in the setter?

TorsteinHonsi (approved):
I'm happy with the result now, tested on mobile and various browsers. I'm marking it approved by me. 

@oysteinmoseng  anything to add regarding a11y before merging in?

oysteinmoseng:
No, looks good! Native HTML elements are fairly a11y-safe. Looks like we will have to do some work in the a11y module to adapt, but we can merge this in and I can work from master as long as we don't do a release yet.

PR #14720, opened 2020-12-01
ts/solidgauge-series

goransle:
Migrated `solidgauge` series to class pattern


bre1470 (approved):
Thank you! Looks good. 👍

PR #14724, opened 2020-12-01
ts/PivotPoints-indicator 

K009:


ts/Stock/Indicators/PivotPoints/PivotPointsIndicator.ts
```diff
@@ -0,0 +1,447 @@
...
+    }
+}
+
+interface PivotPointsIndicator{
+    endPoint: number;
+    nameBase: string;
+    plotEndPoint: number;
+    pointArrayMap: Array<string>;
+    pointClass: typeof PivotPointsPoint;
+    pointValKey: string;
```
bre1470, line 410: `endPoint` and `plotEndPoint` are not set in extend and should therefor be moved to the properties in the class pattern.

ts/Stock/Indicators/PivotPoints/PivotPointsPoint.ts
```diff
@@ -0,0 +1,81 @@
...
+        this: PivotPointsPoint
+    ): void {
+        destroyExtraLabels(this, 'destroyElements');
+    }
+    // This method is called when removing points, e.g. series.update()
+    destroy(
+        this: PivotPointsPoint
+    ): void {
+        destroyExtraLabels(this, 'destroyElements');
+    }
```
bre1470, line 67: Functions in class pattern should explicitly set public.

ts/Stock/Indicators/PivotPoints/PivotPointsPoint.ts
```diff
@@ -0,0 +1,81 @@
...
+}
+interface PivotPointsPoint {
+    P: number;
+    pivotLine: string;
+    series: PivotPointsIndicator;
+}
```
bre1470, line 73: All properties can be moved to the class pattern as they are not set in an extend call.

bre1470 (changes requested):

bre1470 (approved):
Thank you! Looks good. 👍

PR #14726, opened 2020-12-01
bugfix/tema-dema-indicator-linkedTo

karolkolodziej:
Fixed #14698, issue with how the `EMAPercent` was calculated in the DEMA and TEMA indicator.


TorsteinHonsi (approved):
Thanks!

PR #14734, opened 2020-12-02
bugfix/13804-3d-pie-inactive-animation

### LINKED ISSUE #13804: 3d pie chart error when updating on inactive browser tab

khlieng:
Fixed #13804, 3d pie animation in inactive tab threw.

### DIFF
--- a/js/Core/Renderer/SVG/SVGElement.js
+++ b/js/Core/Renderer/SVG/SVGElement.js
@@ -497,7 +497,7 @@ var SVGElement = /** @class */ (function () {
             // Call the end step synchronously
             objectEach(params, function (val, prop) {
                 if (animOptions.step) {
-                    animOptions.step.call(this, val, { prop: prop, pos: 1 });
+                    animOptions.step.call(this, val, { prop: prop, pos: 1, elem: this });
                 }
             }, this);
         }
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (approved):
Thanks!

PR #14737, opened 2020-12-02
bugfix/13490-polar-update-hover

### LINKED ISSUE #13490: Updating the chart to be non-polar breaks point hover
**Expected behavior:**
Point hover effect (halo, tooltip) should continue working after updating a chart to be non-polar.

**Actual behaviour:**
Hovering over datapoints while polar _and then_ subsequently updating the chart to be non-polar makes the points on the sides (but not the ones in the middle) lose their hover effects, the halo and the tooltip.

**Fiddle:**
https://jsfiddle.net/L444Z/pw9uemLh/21/

Steps:
1. Hover over the chart points to see their tooltips (don't skip this!)
2. Click the button to make the chart non-polar
3. Hover over a point in the beginning or in the end of the series

**Version:** Highcharts 8.1.0

I'd also be happy with a workaround!

khlieng:
Fixed #13490, point hovering effects such as `tooltip` stopped working after disabling `Chart.polar` with `update`.

### DIFF
--- a/js/Extensions/Pane.js
+++ b/js/Extensions/Pane.js
@@ -388,6 +388,9 @@ addEvent(Pointer, 'beforeGetHoverData', function (eventArgs) {
                 (!chart.hoverPane || s.xAxis.pane === chart.hoverPane));
         };
     }
+    else {
+        chart.hoverPane = void 0;
+    }
 });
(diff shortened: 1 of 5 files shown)


TorsteinHonsi (approved):
Thanks!

PR #14739, opened 2020-12-02
bugfix/13948-offline-pdf-lineargradient

### LINKED ISSUE #13948: Offline exporting color axis to pdf with 'highlighted' value results in a file reader can't open
#### Expected behavior
Exporting a chart (with the below described properties) offline, to pdf should result in a well-behaved pdf file, containing the chart roughly as it appears in the browser.

Scenario:
Have a chart with a color axis, with custom stops defined. Using the stops, 'highlight' a special value with a different color that is not part of any gradient (see config in fiddle)

#### Actual behavior
Exporting the above described chart offline to pdf results in a file Adobe reader can't open properly (a warning is shown, dismissing it will show the chart with many parts missing, typically all the axes). Browsers do open it, but the color axis (apart from the ticks and labels) appears transparent (tried with Chrome and Edge).

https://jsfiddle.net/L444Z/fyctm67q/14/

Steps:
1. Click the hamburger and then click download pdf document menu item
2. Open the result file with either Acrobat reader or your browser (they produce different behavior as described above)

#### Product version
Highcharts 8.1.2

goransle:
Fixed #13948, exported PDFs containing linear gradients with multiple 0 offset stops had errors.

### DIFF
--- a/js/Extensions/OfflineExporting.js
+++ b/js/Extensions/OfflineExporting.js
@@ -210,6 +210,20 @@ function downloadSVGLocal(svg, options, failCallback, successCallback) {
         [].forEach.call(svgElement.querySelectorAll('*[visibility="hidden"]'), function (node) {
             node.parentNode.removeChild(node);
         });
+        // Workaround for #13948, multiple stops in linear gradient set to 0
+        // causing error in Acrobat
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #14741, opened 2020-12-02
ts/ikh-indicator

pawellysy:


js/Stock/Indicators/IKHIndicator.js
```diff
@@ -6,12 +6,26 @@
  *
```
bre1470, line 6: This whole file needs to be deleted and with that the referencing masters files needs to be updated.

ts/Stock/Indicators/IKH/IKHIndicator.ts
```diff
@@ -0,0 +1,960 @@
...
+const color = Color.parse;
+import H from '../../../Core/Globals.js';
+import U from '../../../Core/Utilities.js';
+import {
+    IKHDrawSenkouSpanObject,
+    IKHGapExtensionObject,
+    IKHOptions,
+    IKHParamsOptions,
+    IKHSenkouSpanOptions
+} from './IKHOptions';
```
bre1470, line 38: The import should make it clear that it is about types. Therefor explicitly do `import type`.

ts/Stock/Indicators/IKH/IKHIndicator.ts
```diff
@@ -0,0 +1,960 @@
...
+    *  Properties
+    *
+    * */
+
+    public data: Array<IKHPoint> = void 0 as any;
+    public options: IKHOptions = void 0 as any;
+    public points: Array<IKHPoint> = void 0 as any;
+    public pointValKey = 'tenkanSen';
+    public nameComponents = ['periodSenkouSpanB', 'period', 'periodTenkan'];
+
```
bre1470, line 374: The last two properties should be kept in the extend call for the class prototype.

ts/Stock/Indicators/IKH/IKHIndicator.ts
```diff
@@ -0,0 +1,960 @@
...
+    ): IndicatorValuesObject<TLinkedSeries> | undefined;
+    graphCollection: Array<string>;
+    graphsenkouSpan: SVGElement | undefined;
+    ikhMap: Highcharts.Dictionary<Array<IKHPoint>>;
+    nextPoints?: Array<IKHPoint>;
+    options: IKHOptions;
+    pointValKey: string;
+    pointClass: typeof IKHPoint;
+    points: Array<IKHPoint>;
+    pointArrayMap: Array<keyof IKHPoint>;
```
bre1470, line 926: Every property here, that will be not part of the extend call (has a default value) needs to be moved to class pattern and in some cases initialized with `void 0 as any`. Exception to this rule is the pointClass property, which stays in the interface.

bre1470 (changes requested):

js/Stock/Indicators/IKHIndicator.js
pawellysy: Deleted file, and updated masters files

ts/Stock/Indicators/IKH/IKHIndicator.ts
pawellysy: Added 'type' import

ts/Stock/Indicators/IKH/IKHIndicator.ts
pawellysy: Moved those two properties to extend call

ts/Stock/Indicators/IKH/IKHIndicator.ts
pawellysy: Moved given properties to class pattern

bre1470 (approved):
Thank you! Looks good. 👍

sebastianbochan:
I had to re-run tests (disconnect issue) and now, the indicator is ready to merge.

PR #14744, opened 2020-12-03
Refactor series.update, use setPrototypeOf to cast series type

TorsteinHonsi:
The `Object.setPrototypeOf` function is supported by IE11 and all modern browsers. This ensures that when running `Series.update({ type })`, the series is a true instance of the new type, instead of an imitation with copied-over members like currently.


bre1470 (approved):
Neat. 👍

pawelfus (approved):
Nothing to add, nice!

PR #14748, opened 2020-12-03
Fix creating tickPositions with [ NaN, NaN ] and therefore calling axisFormatter

sag-tobias-frey:
Consider the following scenario: There is a series with x / y values; however all y values are "null" because there is no data yet. There are two cases now:

1) this.len >= tickPixelInterval => tickAmount stays at undefined =>tickPositions is an empty array => axisFormatter is not called
2) this.len < tickPixelInterval => tickAmount is set to 2 => tickAmount is set to 5 at then end => tickPositions is [NaN, NaN] in the end => axisFormatter is called with garbage

This change fixes this issue by only setting tickAmount to 2 if tickAmount was only 0 not undefined.

### DIFF
--- a/js/Core/Axis/Axis.js
+++ b/js/Core/Axis/Axis.js
@@ -1696,7 +1696,7 @@ var Axis = /** @class */ (function () {
     Axis.prototype.getTickAmount = function () {
         var axis = this, options = this.options, tickAmount = options.tickAmount, tickPixelInterval = options.tickPixelInterval;
         if (!defined(options.tickInterval) &&
-            !tickAmount && this.len < tickPixelInterval &&
+            tickAmount == 0 && this.len < tickPixelInterval &&
             !this.isRadial &&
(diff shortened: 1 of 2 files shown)


TorsteinHonsi:
Thanks for writing!

Can you provide a demo for the fail case? Then we can create a fix with a regression test and verify that your suggested fix is valid.

sag-tobias-frey:
@TorsteinHonsi I have created a jsfiddle showing the error ( https://jsfiddle.net/hLo7avxr/ ). With height = 150 an alert is triggered; this does not happen with height = 250. Do you need something else?

pawelfus:
Thank you for the demo @sag-tobias-frey (and for the PR!)

Created a bug report based on your demo and PR: #14769

TorsteinHonsi:
I'm closing this PR now, the issue will take over. Thanks again for reporting!

Issue #13868, opened 2020-07-09
Pie series in Highstock - Error when hovering 


ppotaczek:
#### Expected behaviour
Hovering over the pie series without errors.

#### Actual behaviour
Error when hovering over the pie series: `Uncaught TypeError: Cannot read property 'lineWidth' of undefined`

#### Live demo with steps to reproduce
http://jsfiddle.net/BlackLabel/fLxbv0pk/

#### Product version
Highstock v8.1.2

#### Affected browser(s)
All

ppotaczek:
**Workaround:** disable splitted tooltip and set at least an empty configuration object for marker inactive state:

	tooltip: {
	    split: false
	},
	series: [{
	    ...,
	    marker: {
	        states: {
	            inactive: {}
	        }
	    }
	}]

**Live example:** http://jsfiddle.net/BlackLabel/09z8tcnq/

PR #14749, opened 2020-12-03
bugfix/13868-error-when-hovering-on-pie

DJTechnoo:
Fixed: #13868 error when hovering on pie.


pawelfus:
Fixed in https://github.com/highcharts/highcharts/pull/15872

PR #14750, opened 2020-12-03
Demo/xmas card 2020

nasin:
files for the Christmas card


TorsteinHonsi:
Changes to the ESLint settings at the root will affect the whole project, including the Highcharts source code, and must be well tested and based on a team discussion.

TorsteinHonsi:
PS: We deliberately allowed long strings in samples in our current config:

https://github.com/highcharts/highcharts/blob/6809d394295bc107ddbff9a8a336bf680fe0c9fb/samples/.eslintrc/#L32

TorsteinHonsi:
Closing this, it's too risky to mess with the global ESLint settings...

PR #14752, opened 2020-12-04
bugfix/14746-datetime-null-local

### LINKED ISSUE #14746: Null columns & yAxis datatime & disabled useUTC bug
#### Expected behaviour

No console error and the empty chart should be rendered - the above config should work as it used to be in the  <v8.0.0 

#### Actual behaviour

The error occurs in the console and the empty chart doesn't show up. 

Demo: https://jsfiddle.net/BlackLabel/mf01x3wz/

#### Product version

     > 8.0.1

khlieng:
Fixed #14746, column chart with null data, datetime axis and `useUTC` set to `false` threw.


TorsteinHonsi (approved):
Thanks, looks good!

PR #14753, opened 2020-12-04
bugfix/13716-initialize-responsive-rules

### LINKED ISSUE #13716: Initializing annotations in responsive rules does not work
#### Expected behaviour
Annotation added to the chart when responsive rule is applied on chart init.

#### Actual behaviour
Annotation is not added and error is being thrown.

https://jsfiddle.net/BlackLabel/vhyu1sjd/2/

#### Product version
n/a

#### Affected browser(s)
Chrome Version 83.0.4103.97, probably others too

#### Workaround:
Define an empty array in Chart.prototype and in chart options:
```js
Highcharts.Chart.prototype.annotations = []; // Workaround, part I

var chart = Highcharts.chart('container', {


    annotations: [], // Workaround, part II

    responsive: {
      rules: [{
        condition: {
          maxWidth: 500
        },
        chartOptions: {
          annotations: [{
            shapes: [{
              point: '3',
              type: 'circle',
              r: 10
            }]
          }]
        }
      }]
    }
  })
```

Demo: https://jsfiddle.net/BlackLabel/vhyu1sjd/1/

goransle:
Fixed #13716, annotations added in responsive rules did not work.

This was due to `setResponsive` being called before callbacks were added.

### DIFF
--- a/js/Core/Chart/Chart.js
+++ b/js/Core/Chart/Chart.js
@@ -1668,10 +1668,6 @@ var Chart = /** @class */ (function () {
         chart.renderLabels();
         // Credits
         chart.addCredits();
-        // Handle responsiveness
-        if (chart.setResponsive) {
-            chart.setResponsive();
-        }
         // Handle scaling
         chart.updateContainerScaling();
         // Set flag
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (approved):
Thanks!

PR #14757, opened 2020-12-04
ts/ajax-namespace

bre1470:
Removed ajax types from internal namespace.


bre1470:
I do not get the Circle CI diffs fixed. It makes no sense.

PR #14761, opened 2020-12-04
Prevents adding conflicting classNames to SVGElement

ostermanj:
`SVGElement.prototype.addClass` will currently add a class as long the current className string does not include it exactly. If for example you programmatically change point colors  you will get a className string that includes conflicting classNames. This is true for any "serialized" class ending in hyphen-number.

for example `class="highcharts-point highcharts-color-0 highcharts-color-1"`

if you programmatically try to change that point back to `highcharts-color-0` the `highcharts-color-1` will remain and override it.

this fix checks for the presence of "serialized" classes and will remove any that would be overridden by the new one.

there may be a better way to do it -- this relies on some RegEx sniffing. I left out the typescript typing.

### DIFF
--- a/ts/Core/Renderer/SVG/SVGElement.ts
+++ b/ts/Core/Renderer/SVG/SVGElement.ts
@@ -633,6 +633,11 @@ class SVGElement {
                 name: string
             ): Array<string> {
                 if ((currentClassName as any).indexOf(name) === -1) {
+                    let split = name.split(/-\d+$/); // see if the `name` is a serialized class that ends in hyphen-number, i.e., `highcharts-color-1`
+                    if (split.length > 1) { // if so the split array will have 2 entries
+                        let regex = new RegExp(split[0] + '-\\d+$'); // remove the corresponding existing className from what will be returned (i.e., remove `highcharts-color-0`)
+                        newClassName[0] = newClassName[0].replace(regex, '');
+                    }
                     newClassName.push(name);
                 }
                 return newClassName;


pawelfus:
Hi @ostermanj 

Thank you for the PR! Could you attach a demo where it's broken? We will check it

ostermanj:
Thanks, @pawelfus . Here's  fiddle: https://jsfiddle.net/6weLda1z/

It's a pie chart in styledMode. The colorIndexes of the points are changed after 2 seconds by the `Chart.update` method and then changed back after 2 more second. The first color change works but the second one does not because the `highcharts-color-X` class introduced by the first color change is not removed.

Perhaps using `Series.update` would be better?

TorsteinHonsi:
@pawelfus I think the `addClass` and related functions are up for a refactoring.
* Use `classList` in stead of setting `className`. It is supported by IE > 9. Provide polyfills for old browsers.
* Support passing an array of strings to `addClass`. If a string is passed (for legacy), we can just split it and pass it as arguments to `classList.add`.

I think those changes also should fix this issue.

pawelfus:
Thank you for the demo @ostermanj!

FYI: In general, I don't recommend changing `Chart.userOptions`. Using just `chart.update( newOptions )` should be enough. We constantly improve dynamic methods and in the future, when we won't see a change in comparison to `userOptions`, we may omit such options.

**Simplified demo:**
https://jsfiddle.net/BlackLabel/pt3eock9/3/

Created a separate bug report: https://github.com/highcharts/highcharts/issues/14773

ostermanj:
Thanks, all. @pawelfus : point well taken — I was just grabbing userOptions to get a reference to the config object. Thanks for all your work.

TorsteinHonsi:
Thanks again for posting, I'm closing this as it will be taken care of in #14773

PR #14762, opened 2020-12-04
bugfix/getTimeTicks-negative-values

karolkolodziej:
Fixed, getTimeTicks returned negative values.
Replacement for the #14694 PR.

### DIFF
--- a/js/Core/Time.js
+++ b/js/Core/Time.js
@@ -566,11 +566,11 @@ var Time = /** @class */ (function () {
             if (interval === timeUnits.week) {
                 // get start of current week, independent of count
                 minDay = time.get('Day', minDate);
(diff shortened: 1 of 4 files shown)


karolkolodziej:
After some test and approaches, it looks like this fix won't fit. 
We need that negative ticks in some cases to calculate when the new week begins. 
We can't force to set that date to 1 or forbid the negative ticks.

Here we are calculating that exact date in order to get a correct start of the week. Setting it here to `1` looks like ignoring that piece of code.
https://github.com/highcharts/highcharts/blob/c836b306b70aeef585e1462850d749179a603176/ts/Core/Time.ts#L836-L842
Or am I missing something in this mechanism?

Onixmiha:
bugfix/getTimeTicks-negative-values

PR #14763, opened 2020-12-04
bugfix/13283-hexcolor-treated-as-comment)

### LINKED ISSUE #13283: Hex color format is ignored in data from CSV
#### Expected behaviour
Hex color should work like other formats: http://jsfiddle.net/BlackLabel/6m4e8x0y/4946/

#### Actual behaviour
Hex color is interpreted as `null` value.

http://jsfiddle.net/BlackLabel/0gjbn4m6/1/

#### Product version
Highcharts v8.0.4

#### Affected browser(s)
All
___
_Interna note:_
Probably caused by:
https://github.com/highcharts/highcharts/blob/2aef02113d9a157bb621200cb795f3b58a8a1a04/ts/modules/data.src.ts#L1449-L1457

DJTechnoo:
Fixed #13283, an issue with the data module's CSV parser. Hex colors (and any other strings following a hash) were treated as a comment and ignored.

~~Fixed #13283, Hex color ignored as a comment.~~

### DIFF
--- a/samples/unit-tests/data/csv/demo.js
+++ b/samples/unit-tests/data/csv/demo.js
@@ -643,34 +643,6 @@ QUnit.test('startRow, endRow, startColumn, endColumn', function (assert) {
     });
 });
 
-QUnit.test('Comments in CSV', function (assert) {
-    var data = [
-        '# -------',
-        '# Comment',
-        '# ----',
-        'Apples,Pears',
-        '1,2# Inline comment',
-        '3,4',
-        '5,6'
-    ].join('\n');
-
-    Highcharts.data({
-        csv: data,
-        parsed: function () {
-            assert.strictEqual(
-                this.columns[0].join(','),
-                'Apples,1,3,5',
-                'First column ok'
-            );
-            assert.strictEqual(
-                this.columns[1].join(','),
-                'Pears,2,4,6',
-                'Second column ok'
-            );
-        }
-    });
-});
-
 // Highcharts 4.0.4, Issue #3437
 // Data module fails with numeric data in first column
 QUnit.test('Data module numeric x (#3437)', function (assert) {
--- a/samples/unit-tests/data/csv/demo.js
+++ b/samples/unit-tests/data/csv/demo.js
@@ -713,3 +713,27 @@ QUnit.test('Dot date format', function (assert) {
         }
     });
 });
+
+QUnit.test(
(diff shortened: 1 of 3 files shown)


TorsteinHonsi:
@cvasseng Arguably we should not support hash-style comment at all. Apparently the CSV spec doesn't mention it. When putting the issue report's CSV in a .csv file and opening it in Numbers, it parses the comments as regular strings. Likewise with https://csvjson.com/. So an alternative solution here would be to remove this altogether:

```js
                // Quoted string
                if (c === '#') {
                    // The rest of the row is a comment
                    push();
                    return;
                }
```

What do you think?

DJTechnoo:
> So an alternative solution here would be to remove this altogether:

I would agree to remove it. Shall I proceed to do so?

cvasseng:
This was initially done more out of convention, there's some (obscure) software that exports CSV with column formatting in comments, and it was low-effort to add, so I just did it. 

I don't see any reason not to remove it, especially if it's causing issues elsewhere.

samples/unit-tests/data/csv/demo.js
```diff
@@ -713,3 +685,27 @@ QUnit.test('Dot date format', function (assert) {
         }
     });
 });
+
+QUnit.test(
+    'Hex values are not treated as comments (#13283).',
+    function (assert) {
+        var chart = Highcharts.chart('container', {
```
TorsteinHonsi, line 692: Instead of creating a full `Chart` instance, we can test this using the Data module only. See some of the other tests in the same file for how we call `Highcharts.data` and do the assertion in the `parsed` callback.

samples/unit-tests/data/csv/demo.js
```diff
@@ -713,3 +685,27 @@ QUnit.test('Dot date format', function (assert) {
...
+
+                itemDelimiter: ';',
+                seriesMapping: [{
+                    color: 2
+                }]
+            }
+        });
+
+        var actualColor = chart.series[0].data[0].options.color;
+        assert.strictEqual(actualColor, '#F0F', 'Color must be a hex value');
```
TorsteinHonsi, line 709: Not a big deal, but our general rule of thumb is that assertion descriptions should contain the word "should". We may add a pre-commit rule for it in the future.

TorsteinHonsi (changes requested):

TorsteinHonsi (approved):

PR #14767, opened 2020-12-07
bugfix/14765-global-patterns

### LINKED ISSUE #14765: Highcharts.patterns is not defined
#### Expected behaviour

[In v8.2.0 and prior](https://jsfiddle.net/goransle/Lf1acxg5/), the pattern-fill module registered a global array `Highcharts.patterns` containing available patterns. As this was not mentioned in the [changelog](https://www.highcharts.com/blog/changelog/#highcharts-v8.2.2) I assume this was not intended.

#### Actual behaviour

[In v8.2.2 and later, this array is `undefined`](https://jsfiddle.net/goransle/vq03fuz7/)

#### Product version
Highcharts 8.2.2 + current master

#### Affected browser(s)
Tested on latest Chrome, Edge, Firefox

khlieng:
Fixed #14765, `Highcharts.patterns` global was missing.

### DIFF
--- a/js/Extensions/PatternFill.js
+++ b/js/Extensions/PatternFill.js
@@ -14,6 +14,7 @@
 import A from '../Core/Animation/AnimationUtilities.js';
 var animObject = A.animObject;
 import Chart from '../Core/Chart/Chart.js';
+import H from '../Core/Globals.js';
 import LineSeries from '../Series/Line/LineSeries.js';
 import Point from '../Core/Series/Point.js';
 import SVGRenderer from '../Core/Renderer/SVG/SVGRenderer.js';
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (commented):
@bre1470 Is this way of registering members on the namespace okay? Or should we do it in the master file?

bre1470:
There exists no API documentation, that `Highcharts.patterns` is available. So there is no guidance what it correct.

In general I like the idea with the master files. That way we can easily remove all the global namespace usage from `es-module/masters` in a future postprocessing and at the same time be backwards compatible in the regular masters. Maybe that should be one of the next migration jobs: Moving all global assignments to the masters.

TorsteinHonsi (approved):
Thanks. I'll merge this now as we need to deal with internal namespace later anyway.

Issue #13488, opened 2020-05-11
Chart inside a scrollable div: persistent crosshair offset when scrolling the div


olejorgenb:
#### Live demo with steps to reproduce

https://jsfiddle.net/95crzfu7/2/

Srcoll the inner container (overflow: auto) while the pointer is inside the chart, observe that the crosshair get an vertical offset. Only way to reset the offset is to move the cursor out and into the chart again.

Probably related: #12799 (I haven't tried yet, but I suspect the linked SO answer would work: https://stackoverflow.com/a/59769430/1517969)

#### Product version

Highcharts JS v8.1.0 (2020-05-05)

#### Affected browser(s)

Tested on recent versions of Firefox and Chrome

pawelfus:
Thanks for reporting this bug @olejorgenb 

**Additional workaround:**
Demo: https://jsfiddle.net/BlackLabel/djh84vqo/
Plugin: 
```js
(function(H) {
  H.wrap(
    H.Pointer.prototype,
    'getChartPosition',
    function(p) {
      delete this.chartPosition;

      return p.apply(this, Array.prototype.slice.call(arguments));
    }
  )
})(Highcharts);
```

*Internal note:*
In theory, we do clear this cache on `this.scrollingContainer`'s `scroll` event. We need to check why it does not solve the issue.

olejorgenb:
Yes, I ended up clearing all the `Highchart.charts[*].pointer.chartPosition` on the `scroll` event :) (assuming there's some performance hit in recalculation the position on every mouse move)

khlieng:
Fixed in #14768.

PR #14768, opened 2020-12-07
bugfix/13996-pointer-position-scroll

### LINKED ISSUE #13996: Zoom rectangle appears at a wrong place after scrolling inside containing element
#### Expected behaviour
The left-top of the zoom rectangle appears where the mouse is clicked, and the right-bottom at the current cursor position.

#### Actual behaviour
The zoom rectangle appears at a wrong spot (looks to be off by the scrolled distance)

https://jsfiddle.net/ktakahashimtb/eno1pvz8

Move the cursor on top of the chart and scroll down inside the container element using mouse wheel, trackpad or arrow keys.
Now, try zooming in the chart.  The zoom rectangle appears at a wrong place, thus zooming in at the unintended plot area.

This behavior can cause **unexpected zoom-in** to happen when scrolling horizontally inside the chart.

https://jsfiddle.net/ktakahashimtb/5wndme3r

In the above fiddle, try:
1. zoom in
2. scroll down (in the container element) with mouse wheel or trackpad
3. scroll horizontally in the graph
4. **Unexpected zoom** happens because the mouse-drag action **on the horizontal scrollbar** is misinterpreted to have happened in the plot area, thus zoom-in is now activated.

#### Diagnosis ####
The problem appears to be happening inside the `Pointer.prototype.normalize` function, where `chartPosition` was not updated properly after an scroll event, thus computed `chartY` and `chartX` values are not correct.

https://github.com/highcharts/highcharts/blob/e06fb000fa9fc2ad816a9c5ec4f29e311d90bf9b/ts/Core/Pointer.ts#L1118

It appears to me that, instead of computing `chartY` and `chartX` based on the _current viewport offset_ of an mouse event (`e.PageX`, `e.PageX`) and the chart position (`chartPoition.Left`, `chartPoition.Top`),  we can just copy values from `e.offsetX` and `e.offsetX` to obtain the mouse position relative to the chart.

https://jsfiddle.net/ktakahashimtb/w7ucx40j/

In the fiddle above, I'm overwriting `e.PageX`, and `e.PageY` values with `e.offsetX` and `e.offsetX` using [wrap](https://www.highcharts.com/docs/extending-highcharts/extending-highcharts#wrapping-up-a-plugin).  Now, the zoom rectangle appears at the correct location.  I'm not sure if there are reasons for not setting `e.PageX` and `e.PageY` this way, but so far this workaround has fixed the issue with no side effect.  That said, I'd like to have this issue fixed in the Highcharts code.

#### Product version
Highstock v8.1.2 (Highstock v7.2.1 shows the same issue)

#### Affected browser(s)
Chrome, Firefox, Edge,

goransle:
Fixed #13996, #13488, pointer position was wrong after chart element was scrolled inside another element.

[JSfiddle without fix applied](https://jsfiddle.net/ktakahashimtb/eno1pvz8)
[With fix applied](https://jsfiddle.net/goransle/8koyr305/)

### DIFF
--- a/js/Core/Pointer.js
+++ b/js/Core/Pointer.js
@@ -1381,6 +1381,7 @@ var Pointer = /** @class */ (function () {
      * @function Highcharts.Pointer#setDOMEvents
      */
     Pointer.prototype.setDOMEvents = function () {
+        var _this = this;
         var container = this.chart.container, ownerDoc = container.ownerDocument;
         container.onmousedown = this.onContainerMouseDown.bind(this);
         container.onmousemove = this.onContainerMouseMove.bind(this);
@@ -1390,6 +1391,11 @@ var Pointer = /** @class */ (function () {
         if (!H.unbindDocumentMouseUp) {
             H.unbindDocumentMouseUp = addEvent(ownerDoc, 'mouseup', this.onDocumentMouseUp.bind(this));
         }
+        // In case we are dealing with overflow, reset the chart position when
+        // scrolling
+        addEvent(this.chart.renderTo.parentElement, 'scroll', function () {
+            delete _this.chartPosition;
+        });
         if (H.hasTouch) {
             addEvent(container, 'touchstart', this.onContainerTouchStart.bind(this), { passive: false });
             addEvent(container, 'touchmove', this.onContainerTouchMove.bind(this), { passive: false });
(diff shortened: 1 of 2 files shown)


khlieng:
Probably also fixes #13488.

goransle:
> Probably also fixes #13488.

Seems so: https://jsfiddle.net/goransle/qphb4j0v/

TorsteinHonsi (changes requested):
Thanks!

The fix only looks at the parent element. When adding scrolling further up the DOM tree, it is not handled: https://jsfiddle.net/highcharts/vwouhftd/

TorsteinHonsi (approved):
Thanks!

PR #14776, opened 2020-12-08
bugfix/14549-clip-path-added-axis

### LINKED ISSUE #14549: Multiple panes don't create correct clip-path for dynamically added axis
#### Expected behaviour
Create clip-path for dynamically added axis the same as for static.
This feature works correctly in the 8.0.2 version.
Demo: https://jsfiddle.net/BlackLabel/pqco92kv/

#### Actual behaviour
Series clip-path exceeds the axis when it's added dynamically.

https://jsfiddle.net/BlackLabel/n91a3rmw/

#### Product version
Highstock JS v8.2.2 (2020-10-22)

#### Affected browser(s)
All

karolkolodziej:
Fixed #14549, incorrect clip-path for added axis.

~~This change improves #13751. It applies the max available clip-path width to make sure that none of the series points is going to be cut.  An extra part of the series clip-path is then trimmed by the `plotArea`.~~

The previous approach procured to many visual differences. Latest commit is about to refix the issue introduced in #13751.

### DIFF
--- a/js/Core/Chart/StockChart.js
+++ b/js/Core/Chart/StockChart.js
@@ -685,9 +685,9 @@ addEvent(LineSeries, 'render', function () {
             }
         }
         // First render, initial clip box
-        if (!this.clipBox && !chart.hasRendered) {
-            this.clipBox = merge(chart.clipBox);
-            this.clipBox.width = this.xAxis.len;
+        if (!this.clipBox) {
+            this.clipBox = merge({}, chart.clipBox);
+            this.clipBox.width = chart.chartWidth;
             this.clipBox.height = clipHeight;
             // On redrawing, resizing etc, update the clip rectangle
         }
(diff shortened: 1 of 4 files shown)


pawelfus (changes requested):
Visual diff in Highcharts Stock: ✅   - previously line was overflowing in this demo, now it's clipped. Only this one demo shows the diff, because probably only this one adds series in `chart.load` event.
Visual diffs in Highcharts Maps: 🔴   - this 1px looks strange, could you please check it?

karolkolodziej:
As it comes to the visual differences in maps, it is strange indeed.
When you compare (see the difference in northern Greenland):
- v8.2.2 https://jsfiddle.net/BlackLabel/9407pvyn/
- master http://jsfiddle.net/BlackLabel/ucvmwexp/

You can observe that the differences were introduced to the master in some earlier commit.
Why this change is first caught in this PR?

TorsteinHonsi (approved):
Looks good, thanks!

PR #14781, opened 2020-12-09
docs/accessibility-module

marvin19:
Added more more content and headings to the docs.


docs/accessibility/accessibility-module.md
```diff
@@ -1,10 +1,14 @@
 Accessibility module
 ===
 
-The accessibility of Highcharts is critical to us. Every Highcharts license includes our Accessibility module, which helps ensure your charts are as accessible as possible. We use the WCAG 2.1 standard as our guideline for this, as well as involving users with disabilities in our testing and feature development.
+The accessibility of Highcharts is critical to us. Accessibility should be fundamental to any software project, not an additional add on. Every Highcharts license includes our Accessibility module, which helps ensure your charts are as accessible as possible.  We use the WCAG 2.1 standard as our guideline for this, as well as involving users with disabilities in our testing and feature development.
```
oysteinmoseng, line 4: Maybe replace "not an additional add on" with e.g. "not an afterthought". Also looks like an extra space before "We use the WCAG 2.1 standard...".

docs/accessibility/accessibility-module.md
```diff
@@ -1,10 +1,14 @@
...
 
 We recommend to always include the accessibility module, unless there is an explicit reason not to do so. Including this module will make your charts compatible with assistive technologies used by people with disabilities, and also improve the usability of your charts, helping you reach a broader audience. Including the Accessibility module will also help with SEO by making your charts more visible to web crawlers.
 
+Getting started with accessibility
+----------------------------------
+You can read more about [installation of Highcharts](https://www.highcharts.com/docs/getting-started/installation) and [setting up your first chart](https://www.highcharts.com/docs/getting-started/your-first-chart).
```
oysteinmoseng, line 10: Maybe instead of "You can read more about" just do "Read more about", and put it in _italics_?

docs/accessibility/accessibility-module.md
```diff
@@ -19,29 +23,74 @@ We also recommend including the exporting and export-data modules. This lets use
...
 </figure>
 ```
+<iframe style="width: 100%; height: 470px; border: none;" src=https://www.highcharts.com/samples/embed/highcharts/accessibility/accessible-avg-temp allow="fullscreen"></iframe>
 
-This behavior can be configured with the [`accessibility.linkedDescription`](https://api.highcharts.com/highcharts/accessibility.linkedDescription) option. It is possible to link the chart to any element on the page using this option. The use of the `<figure>` element in the example above is optional, but can help convey to screen reader users that they are navigating a diagram. Note that use of the `<figcaption>` element should be avoided in this context, as it will interfere with the accessibility features of Highcharts.
+**Caption:**
+
+It is possible to add the text description directly on the chart SVG as well, using the [`caption`](https://api.highcharts.com/highcharts/caption) option. If this option is used, the caption will automatically be made available to screen reader users. The caption renders in the bottom, and is a part of the exported chart.
```
oysteinmoseng, line 50: We might want to replace "The caption renders in the bottom" with "The caption renders at the bottom of the chart, and is included if the chart is exported" or something similar

docs/accessibility/accessibility-module.md
```diff
@@ -19,29 +23,74 @@ We also recommend including the exporting and export-data modules. This lets use
...
 
-It is possible to add the text description directly on the chart SVG as well, using the [`caption`](https://api.highcharts.com/highcharts/caption) option. If this option is used, the caption will automatically be made available to screen reader users.
+```js
+Highcharts.chart('container', {
+    caption: {
+        text: '<b>The caption renders in the bottom, and is part of the exported chart.</b>'
```
oysteinmoseng, line 55: Same comment here as above

oysteinmoseng (changes requested):
Great work! Only minor comments.

oysteinmoseng (approved):
Nice!

TorsteinHonsi (approved):

PR #14782, opened 2020-12-09
bugfix/14769-nan-tickpositions

### LINKED ISSUE #14769: `tickPositions` are NaN's when chart is narrow
#### Expected behaviour
No `Axis.labels.formatter` calls with `NaN`'s as values

#### Actual behaviour
`formatter` called with `NaN` values

https://jsfiddle.net/BlackLabel/nsh7pqu9/

#### Product version
v8.2.2

Related PR: https://github.com/highcharts/highcharts/pull/14748

To do:
- [ ] run tests to check if the given solution does not break something
- [ ] create a regression test

K009:
Fixed #14769, axis label formatter was called with NaN values when the axis was short and series consisted of only null values.

~~Fixed #14769, formatter called with Nan values.~~


TorsteinHonsi:
As the visual diff shows, this also changes all cases where the `tickAmount` option is undefined (which is default) and the total axis length is less than the given `tickPixelInterval`, which is default 72. I'm posting another proposal for a fix, where we prevent `adjustTickAmount` from extending the `tickPositions` when there is no axis min and max.

TorsteinHonsi (approved):

Issue #14775, opened 2020-12-08
Add zoom feature for single touch displays ( highcharts.js )


djzurg:
I am intrested in boosted line chart.
It would be great if you will add zoom feature for single touch displays.
If you will add this option then I will buy the library.

# Description of the feature
Add zoomSingleTouch propery to make possible touch-and-drag zoom on single touch monitors.

# Library related to the feature
highcharts.js

# Proof of Concept/Live example for the feature
![image](https://user-images.githubusercontent.com/75671326/101475584-a51fd280-395d-11eb-82d2-94eb8400942a.png)

----

**You can vote for this feature by adding a thumbs-up reaction to this post.**

TorsteinHonsi:
Thanks for the request @djzurg !

I have been experimenting with this now, and see that it is quite easy to implement. Will post a PR later today implementing the feature.

TorsteinHonsi:
PR created. See a working demo at https://jsfiddle.net/highcharts/10cbpru3/show. Open in Chrome, enable Developer Tools and emulate a mobile device.

djzurg:
Thank you!
I've just requested the price of the library.

PR #14783, opened 2020-12-09
Added feature chart.zoomBySingleTouch

TorsteinHonsi:
Added new feature [chart.zoomBySingleTouch](https://api.highcharts.com/highcharts/chart.zoomBySingleTouch) to provide optional mouse-like selection zooming for touch devices. See #14775.

### DIFF
--- a/ts/Core/Pointer.ts
+++ b/ts/Core/Pointer.ts
@@ -1272,7 +1272,7 @@ class Pointer {
             pEvt.returnValue = false;
         }
 
-        if (chart.mouseIsDown === 'mousedown') {
+        if (chart.mouseIsDown === 'mousedown' || this.touchSelect(pEvt)) {
             this.drag(pEvt);
         }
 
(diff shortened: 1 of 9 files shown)


ts/Core/Pointer.ts
```diff
@@ -1272,7 +1272,7 @@ class Pointer {
             pEvt.returnValue = false;
         }
 
-        if (chart.mouseIsDown === 'mousedown') {
+        if (chart.mouseIsDown === 'mousedown' || this.touchSelect(pEvt)) {
             this.drag(pEvt);
         }
```
bre1470, line 1277: Can we use this as an opportunity to do `e = this.normalize(e)` instead of creating pEvt?

bre1470 (approved):
Looks good. 👍

ts/Core/Pointer.ts
bre1470: Okay, I see now that it is in type sense not working.

pawelfus (approved):
Nice!

oysteinmoseng (approved):
Cool!

PR #14786, opened 2020-12-10
ts/removed-dictionary

marvin19:
Removed Dictionary from the internal namespace.


bre1470 (approved):
Thank you! Looks good. 👍

PR #14787, opened 2020-12-10
bugfix/14522-clip-path-multiple-panes

### LINKED ISSUE #14522: Incorrect clip path with multiple panes
#### Expected behaviour
Create identical clip-paths in Stock and in Chart.

#### Actual behaviour
In the Chart, the clip-path is generated globally for the whole chart.

Zoom in to see the difference.
Chart: http://jsfiddle.net/BlackLabel/27n8mt9u/
Stock: http://jsfiddle.net/BlackLabel/ecp1qgzv/

#### Product version
Highcharts / Highstock JS v8.2.2 (2020-10-22)

#### Affected browser(s)
All

karolkolodziej:
Fixed #14522, incorrect series clip-path for the chart with multiple panes.

Use existing code from the Stock and reuse it globally.


TorsteinHonsi (changes requested):
I think the time is ripe no for a thorough refactor of the clipping logic, to make it more simpler, more typesafe and easier to read. The basic idea of the shared clip keys is the initial animation. The logic would be simpler if we just added one clip path for each series, and animated them one by one. But in a complex chart this would lead to simultaneously animating multiple clippings, which could lead to performance problems. Therefore we combine the identical paths for multiple series, and animate just one clip path. With that in mind, I think we should do some refactoring:

* The `getClipBox` function should return a box (`{ x, y, width, height })` for the series, and take the series' own xAxis and yAxis lengths into account instead of the chart plot height and chart plot width. It looks like the contents of `getSeriesClipPath` should be taken into account in the `getClipBox` function rather than having a separate function for the pane stuff.
* The `sharedClipKey` needs to also take clipBox x, width and height into account, and then probably doesn't need to take `option.xAxis` and `options.yAxis` into account.
* Avoid storing the dynamic keys directly on the chart like `(chart as any)[sharedClipKey]`. Instead, create a dictionary on the chart so we can access it in a type safe way.
* Set up unit tests. Inspiration: http://jsfiddle.net/highcharts/sL6e2y79/ . For multiple panes the clip rects should not be identical.

PR #14790, opened 2020-12-10
bugfix/14788-input-type-ie

### LINKED ISSUE #14788: IE11 chokes on the new RangeSelector inputs
#### Expected behaviour
Graceful fallback to text inputs

#### Actual behaviour
Fatal crash. IE77 doesn't like that we set the input type to `date`. A simple solution would be to check for `isMS` or `isIE` in the `preferredInputType` function and return `text` for IE. But be careful that we don't disable date inputs for versions of Edge that support it.

https://jsfiddle.net/highcharts/Lwakgxj7/

#### Product version
Pre release

#### Affected browser(s)
IE11 (probably all IE). Edge not tested.

khlieng:
Fixed #14788, range selector date input threw in IE instead of falling back to text.

### DIFF
--- a/js/Extensions/RangeSelector.js
+++ b/js/Extensions/RangeSelector.js
@@ -1082,9 +1082,21 @@ var RangeSelector = /** @class */ (function () {
         // to the right size when focused.
         var input = createElement('input', {
             name: name,
-            className: 'highcharts-range-selector',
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #14792, opened 2020-12-11
bugfix/8742-stackLabels-dataLabels

### LINKED ISSUE #8742: Some of stackLabels are not displayed when dataLabels are
#### Expected behaviour
**All** stackLabels should be displayed independently from dataLabels.

#### Actual behaviour
Some stackLabels are not displayed when dataLabels are displayed.

#### Live demo
https://jsfiddle.net/BlackLabel/rht1j40w/

#### Product version
6.1.1

#### Workaround 
Set yAxis.stackLabels.allowOverlap to true.

Looks like regression after commit ae932768305de4824ae0d1584f96d3343fb9bc4c

khlieng:
Fixed #8742, some `stackLabels` did not render with `dataLabels` enabled.

### DIFF
--- a/ts/Extensions/Stacking.ts
+++ b/ts/Extensions/Stacking.ts
@@ -373,7 +373,7 @@ class StackItem {
         }
 
         // Rank it higher than data labels (#8742)
-        this.label.labelrank = chart.plotHeight;
+        this.label.labelrank = chart.inverted ? chart.plotWidth : chart.plotHeight;
     }
 
     /**
--- a/ts/Extensions/Stacking.ts
+++ b/ts/Extensions/Stacking.ts
@@ -373,7 +373,7 @@ class StackItem {
         }
 
         // Rank it higher than data labels (#8742)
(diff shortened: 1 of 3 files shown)


ts/Extensions/Stacking.ts
```diff
@@ -373,7 +373,7 @@ class StackItem {
         }
 
         // Rank it higher than data labels (#8742)
-        this.label.labelrank = chart.plotHeight;
+        this.label.labelrank = chart.inverted ? chart.plotWidth : chart.plotHeight;
```
TorsteinHonsi, line 376: We have `chart.plotSizeY`, it would save us a few bytes

TorsteinHonsi (changes requested):
Thanks! Only a minor optimization tip, otherwise fine.

ts/Extensions/Stacking.ts
TorsteinHonsi: In addition, please add this issue to the comment above:

TorsteinHonsi (approved):

PR #14799, opened 2020-12-14
ts/indicators-interfaces

sebastianbochan:
Fixed missing properties in interfaces.


ts/Stock/Indicators/PC/PCIndicator.ts
```diff
@@ -179,10 +179,13 @@ class PCIndicator extends SMAIndicator implements Highcharts.MultipleLinesIndica
...
 interface PCIndicator {
-    getTranslatedLinesNames: Highcharts.MultipleLinesMixin[
+    getTranslatedLinesNames: typeof MultipleLinesMixin[
         'getTranslatedLinesNames'
     ];
-    linesApiNames: Highcharts.MultipleLinesMixin['linesApiNames'];
+    drawGraph: typeof MultipleLinesMixin.drawGraph;
+    toYData: typeof MultipleLinesMixin.toYData;
+    translate: typeof MultipleLinesMixin.translate;
+    linesApiNames: typeof MultipleLinesMixin['linesApiNames'];
```
bre1470, line 188: Only use brackets when not using typeof.

bre1470 (changes requested):
Nice! One thing...

ts/Stock/Indicators/PC/PCIndicator.ts
sebastianbochan: Fixed

bre1470 (approved):
Thank you! Looks good. 👍

PR #14801, opened 2020-12-14
bugfix/14283-colorAxis-update-error

### LINKED ISSUE #14283: Update ColorAxis visible property issues
#### Actual behaviour

Issue 1:
If we start with visible color axis on init, if we update it to hide, ticks and labels hide but not the color bar.

Issue 2:
If we start with not visible color axis on init, if we update it to show, it doesn't

#### Expected behaviour

Color axis should update correctly

http://jsfiddle.net/t3mugL0x/3/

Issue 1 :  Visible on init > Update visible
Issue 2:  Not visible on init > Update visible

#### Product version

Highcharts 8.2.0
https://www.npmjs.com/package/highcharts/v/8.2.0

#### Affected browser(s)

All

K009:
Fixed #14283, problems with dynamically updating the color axis `visible` property

### DIFF
--- a/ts/Core/Axis/ColorAxis.ts
+++ b/ts/Core/Axis/ColorAxis.ts
@@ -40,7 +40,6 @@ import LineSeries from '../../Series/Line/LineSeries.js';
 import palette from '../../Core/Color/Palette.js';
 import Point from '../Series/Point.js';
 import U from '../Utilities.js';
-import { param } from 'jquery';
 const {
     addEvent,
     erase,
--- a/ts/Core/Axis/ColorAxis.ts
+++ b/ts/Core/Axis/ColorAxis.ts
@@ -40,6 +40,7 @@ import LineSeries from '../../Series/Line/LineSeries.js';
 import palette from '../../Core/Color/Palette.js';
 import Point from '../Series/Point.js';
 import U from '../Utilities.js';
+import { param } from 'jquery';
 const {
     addEvent,
     erase,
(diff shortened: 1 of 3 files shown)


ts/Core/Axis/ColorAxis.ts
```diff
@@ -1292,6 +1292,14 @@ class ColorAxis extends Axis implements AxisLike {
         chart.isDirtyLegend = true;
     }
 
+    // Removing the whole axis (#14283)
+    public destroy(): void {
+        this.chart.isDirtyLegend = true;
+
+        this.destroyItems();
+        H.Axis.prototype.destroy.call(this);
```
TorsteinHonsi, line 1300: Use `super.destroy();`. In addition, we probably need to pass over the arguments (`keepEvents`, or more future proof, all arguments) to the super method.

TorsteinHonsi (changes requested):

ts/Core/Axis/ColorAxis.ts
```diff
@@ -40,6 +40,7 @@ import LineSeries from '../../Series/Line/LineSeries.js';
 import palette from '../../Core/Color/Palette.js';
 import Point from '../Series/Point.js';
 import U from '../Utilities.js';
+import { param } from 'jquery';
```
TorsteinHonsi, line 43: This line should be removed, right?

TorsteinHonsi (changes requested):

TorsteinHonsi (approved):

PR #14811, opened 2020-12-15
ts/utilites-namespace-functions

marvin19:
Removed functions from internal namespace in Utilities.


bre1470 (approved):
Thank you! Looks good. 👍

ts/Core/Animation/Fx.ts
```diff
@@ -533,6 +541,19 @@ interface Fx extends FxLike {
...
 }
 
+/* *
+ *
+ *  Compatibility
+ *
+ * */
+
 H.Fx = Fx;
+(H as any).timers = Fx.timers;
```
TorsteinHonsi, line 551: I don't think we need to keep this on the namespace. If the tests pass without it I'd say we remove it.

TorsteinHonsi (changes requested):

ts/Core/Animation/Fx.ts
bre1470: test-utilities requires an update then. (highcharts and highcharts-utils repositories)

TorsteinHonsi: Okay. We'll leave it for now then.

TorsteinHonsi (approved):

PR #14816, opened 2020-12-16
bugfix/14209-Histogram-update-with-empty-array

### LINKED ISSUE #14209: Histogram does not update when empty array passed as new data
#### Expected behaviour
When a histogram is updated from having data to having none, it should update to match what appears when the histogram has no data initially.

#### Actual behaviour
The chart does not change when updated with an empty array.

https://jsfiddle.net/uabwevzL/

#### Product version
Highcharts 8.1.2

#### Affected browser(s)
Chrome (probably all)

pawellysy:
Fixed #14209, histogram after `setData` with empty array wasn't updated.

### DIFF
--- a/js/Series/Histogram/HistogramSeries.js
+++ b/js/Series/Histogram/HistogramSeries.js
@@ -160,6 +160,7 @@ var HistogramSeries = /** @class */ (function (_super) {
     HistogramSeries.prototype.setDerivedData = function () {
         var yData = this.baseSeries.yData;
         if (!yData.length) {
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (changes requested):
Thanks!

Before merging, please add a [description to go into the changelog](https://github.com/highcharts/highcharts/tree/master/changelog#writing-content-for-the-changelog), and tag the PR accordingly.

TorsteinHonsi (approved):

PR #14817, opened 2020-12-16
Fixed #14691, FireFox popup reversed.

### LINKED ISSUE #14691: First of the Bolded subtitles in the annotation edit popup is moved to the last place
When clicking the edit button on annotation, the first bolded subtitle is displayed in the wrong place. Should be on top of the list of properties, and is on the bottom. 

- Add the rectangle shape annotation to the chart
- Click the edit button
- Notice, that the subtitle (in this case 'shape options') is in the wrong position
Firefox ---->
image

Chrome ----> 
image

For different shape it looks like the first title, that is displayed in the options is moved to the last place: 
Firefox --->
image
Chrome --->
image

#### Product version
Highstock v8.2.2

#### Affected browser(s)
Firefox

DJTechnoo:
Fixed #14691, incorrect order of annotation subtitles in pop-ups in Firefox.


ts/Extensions/Annotations/Popup.ts
```diff
@@ -686,10 +686,14 @@ H.Popup.prototype = {
             });
 
             if (isRoot) {
-                storage = storage.sort(function (a): number {
-                    return (a as any)[1].match(/format/g) ? -1 : 1;
+                H.stableSort(storage, function (a: any): number {
```
TorsteinHonsi, line 689: Instead of referencing `H.stableSort`, get it from the Utilities directly. See for example Legend.ts:
```js
import U from './Utilities.js';
const {
    [...],
    stableSort,
    [...]
} = U;
```

ts/Extensions/Annotations/Popup.ts
```diff
@@ -686,10 +686,14 @@ H.Popup.prototype = {
...
+                    return a[1].match(/format/g) ? -1 : 1;
                 });
 
+                if (H.isFirefox) {
```
TorsteinHonsi, line 693: Same here, we're transitioning away from using the namespace, so access `isFirefox` from Utils instead.

TorsteinHonsi (changes requested):

ts/Extensions/Annotations/Popup.ts
DJTechnoo: It appears to me that `isFirefox` does not come from Utils yet.🤔

TorsteinHonsi (approved):
Looking good, just waiting for the tests to pass then it can be merged 👍

PR #14825, opened 2020-12-17
XMas card responsive minimum adaptions

TorsteinHonsi:
Just scaling down the container and making sure other elements don't overflow


nasin:
On very small screens, the audio controls aren't wrapping.

In the CSS, if you add flex-wrap:wrap to the "#controls" style, that will fix it. Otherwise, it looks good.

PR #14836, opened 2020-12-22
bugfix/8635-variwide-zoom

### LINKED ISSUE #8635: Variwide chart - zoom not working properly.
#### Expected behaviour
After zoom a variwide chart points should be rendered properly (next to each other).

#### Actual behaviour
After zoom points are overlapping and they are often on wrong positions.

https://jsfiddle.net/wchmiel/fwL7t3hn/

#### Product version
Highcharts

#### Affected browser(s)
all

khlieng:
Fixed #8635, zooming variwide chart did not work.

Zooming inverted variwide charts still doesnt work.


TorsteinHonsi (approved):
Thanks!

PR #14839, opened 2020-12-22
bugfix/14833-column-axis-breaks

### LINKED ISSUE #14833: Column series display incorrect values with axis breaks
#### Expected behaviour
A column series on a chart with an x-axis break should behave the same as a line series: columns between `from` and `to` should be removed.

#### Actual behaviour
The columns between `from` and `to` are shifted to the `from` x position. The last column in the break ends up on top, resulting in an incorrect y value for that position.

Adapted from the [docs example](https://api.highcharts.com/highcharts/xAxis.breaks):
https://jsfiddle.net/ug7xhofL/

#### Product version
Highcharts 8.2.2

#### Affected browser(s)
Tested in Chrome 87

khlieng:
Fixed #14833, column series axis breaks did not render correctly.

### DIFF
--- a/ts/Series/Column/ColumnSeries.ts
+++ b/ts/Series/Column/ColumnSeries.ts
@@ -1115,7 +1115,7 @@ class ColumnSeries extends LineSeries {
                 verb = graphic && chart.pointCount < animationLimit ?
                     'animate' : 'attr';
 
-            if (isNumber(plotY) && point.y !== null) {
+            if (isNumber(plotY) && point.y !== null && point.visible !== false) {
                 shapeArgs = point.shapeArgs;
 
                 // When updating a series between 2d and 3d or cartesian and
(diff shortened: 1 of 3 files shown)


ts/Series/Column/ColumnSeries.ts
```diff
@@ -1115,7 +1115,7 @@ class ColumnSeries extends LineSeries {
                 verb = graphic && chart.pointCount < animationLimit ?
                     'animate' : 'attr';
 
-            if (isNumber(plotY) && point.y !== null) {
+            if (isNumber(plotY) && point.y !== null && point.visible !== false) {
```
TorsteinHonsi, line 1118: For the sake of consistency, I think we should do this the same way as we do with pie series. We may want to refactor this repeated pattern in the future.

https://github.com/highcharts/highcharts/blob/35163c02456af62b4f26a93a068510844ae7a25d/ts/Series/Pie/PieSeries.ts#L951-L953

TorsteinHonsi (changes requested):

TorsteinHonsi (approved):

PR #14844, opened 2020-12-23
bugfix/13943-split-tooltip-alignment

### LINKED ISSUE #13943: Tooltip gets rendered on the wrong side
If you have a small width for the chart and set tooltip/outside to true and put the chart on the right side of the screen
the tooltip gets rendered on the right, therefore you don't see it. it should be rendered on the left side

#### Expected behaviour
With a small chart and tooltip outside true, when there is no space on the right, the tooltip should be rendered on the left side

#### Actual behaviour
the tooltip gets rendered on the right side and exceeds the screen, although there is space on the left

http://jsfiddle.net/nickwinger/qtg4a8rL/1/

#### Product version

latest

goransle:
Fixed #13943, split tooltip was rendered to wrong side on smaller charts.

## Tooltip.outside: false
No fix: http://jsfiddle.net/goransle/h0c1d5xw/
With fix applied: http://jsfiddle.net/goransle/fqmo69kw/

## Tooltip.outside: true
`splitRenderer` now checks if the tooltip is inside the viewport when rendering an outside tooltip. The `.highcharts-tooltip-container` now has it's width set to the width of the root element instead of the width of the tooltip.
 
No fix: http://jsfiddle.net/goransle/jx2p3vzn/
Fix: http://jsfiddle.net/goransle/Lt4y7mdu/


TorsteinHonsi (changes requested):
In the bug report, the chart is aligned to the right, so the tooltips overflowed when rendered to the right of the point.

However, when the chart is aligned left, with the fix applied, the tooltips are clipped to the left: http://jsfiddle.net/goransle/fqmo69kw/.

So we're basically just pushing the same problem over to the left side...

Ideally, when `toooltip.outside`, the `bounds` should be the whole page.

TorsteinHonsi (changes requested):
In the beginning of the tooltip placement logic, we set up the `bounds` that the tooltip is allowed to render within:

https://github.com/highcharts/highcharts/blob/6a6ed6d5844f48e8e6f2afb8c1e1ac8d928619b6/ts/Core/Tooltip.ts#L1456-L1463

In this PR, an exception is made for outside tooltip, where the tooltip placement is post-computed inline to align inside the viewport instead of the bounds. It would be cleaner if the bounds in this case _were_ the viewport, then hopefully the placement logic itself wouldn't have to consider special cases.

Can you have a try at that?

ts/Core/Tooltip.ts
```diff
@@ -1723,8 +1734,18 @@ class Tooltip {
             return boxes;
         }, []);
 
+        const isInsideRightBounds = (box: Record<string, any>): boolean => (
+            (tooltip.outside ?
+                box.anchorX + distance + box.boxWidth :
+                box.point.plotX + box.boxWidth) < bounds.right
```
TorsteinHonsi, line 1740: The left-overflow test simply tests for `box.x < scrollLeft`. Why is the right-overflow test more complicated? I would expect `box.x + box.boxWidth` to reveal the right edge of the tooltip regardless of `tooltip.outside`, since `box.x` is used regardless of `tooltip.outside` in the left-overflow check. If so, it would be easier to read and maintain this code if the left and right checks looked the same.

ts/Core/Tooltip.ts
```diff
@@ -1773,18 +1795,17 @@ class Tooltip {
...
         if (outside && container && renderer) {
-            // Set container size to fit the tooltip
-            const { width, height, x, y } = tooltipLabel.getBBox();
+            // Set container size to fit the bounds
+            const { width, x } = tooltipLabel.getBBox();
             renderer.setSize(
-                width + x,
-                height + y,
+                scrollablePixelsX ? width + x : bounds.right - bounds.left,
+                bounds.bottom - bounds.top,
```
TorsteinHonsi, line 1802: Does the renderer have to cover the whole bounds? Which are effectively the whole viewport when `tooltip.outside` and the whole chart when not? The initial idea here was that the renderer doesn't have to be larger than the tooltip itself.

TorsteinHonsi (changes requested):

goransle:
@TorsteinHonsi @Izothep This one fell by the wayside for a bit. Any thoughts on the latest attempt? Also, I noticed that the stockChart constructor seems to mutate the options object on the current master (issue #15445).

TorsteinHonsi:
Thanks! The code looks good.

Though in the [tooltip outside:true demo](http://jsfiddle.net/goransle/Lt4y7mdu/) it seems like the bounds are not working consistently.
* When hovering the last points of the left chart, the tooltips are to the left of the points. Why does that happen when the bounds are the whole document?
* When hovering the first points or the right chart, the _header tooltip_ is moved to the right. Also, why does that happen when the bounds are the whole document?

In the [tooltip outside:false demo](http://jsfiddle.net/goransle/fqmo69kw/) I can't see a difference at all...

Also, we need tests to prevent regressions.

goransle:
> * When hovering the last points of the left chart, the tooltips are to the left of the points. Why does that happen when the bounds are the whole document?

This is actually the current behaviour of the split tooltip, left-aligned until there is not enough space. See http://jsfiddle.net/goransle/hovtf0kd/

The header tooltip may still be clamped to the chart position, will investigate.

TorsteinHonsi:
> This is actually the current behaviour of the split tooltip, left-aligned until there is not enough space

Oh, stupid me, I made that decision myself 😬

TorsteinHonsi (approved):
Perfect, thanks!

PR #14846, opened 2020-12-23
bugfix/14834-colorAxis-update-reversed

### LINKED ISSUE #14834: Heatmap's coloraxis legend flips the order with every update
#### Expected behaviour
Heatmap renders coloraxis in the order specified with [colorAxis.reversed](https://api.highcharts.com/highcharts/colorAxis.reversed) option.

#### Actual behaviour
Heatmap alternates the color-axis order every time update happens.

Here is a demo of how we ran into this issue when Heatmap is created through Highcharts React wrapper.
(please press the update button several times to see the issue)

https://stackblitz.com/edit/highcharts-heatmap-coloraxis-flipping?file=src/App.js

This is a more stripped down version, calling [Chart.update](https://api.highcharts.com/class-reference/Highcharts.Chart#update) manually, simulating Hightcharts React wrapper updating the chart.

https://jsfiddle.net/ktakahashimtb/c5osgm0v/

It appears that, in order to cause this issue, the colorAxis.reversed option needs to be set to a non-default value, and also [colorAxis.tickPositions](https://api.highcharts.com/highcharts/colorAxis.tickPositions) need to be set.   There may be other properties besides tickPositions, which can also trigger this issue.

This happens with both 'vertical' or 'horizontal' [layout](https://api.highcharts.com/highcharts/colorAxis.layout)

Also, the issue seems to start happening only from the second update.  It does not happens on the first update.

#### Product version
Highcharts & Highstock, with currently released version:  8.2.2

#### Affected browser(s)
Chrome, Firefox, Edge, Safari

khlieng:
Fixed #14834, `colorAxis` ignored `reversed` option set in chart config on `update`.


TorsteinHonsi (changes requested):
I think we need to handle this in a more general way. We shouldn't have to handle `reversed` specifically on line 1250. In fact, the same problem also applies to other options, at least `side`: https://jsfiddle.net/highcharts/ac13gdfL/

The problem lies in the way default options are merged with user options, so we should go back to the offending commit and see how this can be done safely in a generic way.

TorsteinHonsi (approved):
Thanks! Approved, just waiting for CI checks.

PR #14851, opened 2020-12-28
bugfix/14793-zAxis-update

### LINKED ISSUE #14793: Scatter3D chart - issue after changing title of Z axis
#### Expected behaviour
Label on Z axis changes

#### Actual behaviour
Label disappears, axis gets distorted

https://jsfiddle.net/stonaudr/ozr28pjc/

#### Product version
Highcharts 8.2.2

#### Affected browser(s)
All

khlieng:
Fixed #14793, Z axis broke on `update`.

### DIFF
--- a/js/Core/Axis/ZAxis.js
+++ b/js/Core/Axis/ZAxis.js
@@ -141,6 +141,8 @@ var ZAxis = /** @class */ (function (_super) {
             offset: 0,
             lineWidth: 0
         }, userOptions);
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (changes requested):
Thanks!

The fix looks fine, but for test performance reasons, please add the assertion to one of the existing tests for `zAxis` update instead of creating a new test with a new chart.

TorsteinHonsi (approved):

PR #14856, opened 2020-12-29
bugfix/rangeselector-event-leak

khlieng:
Fixed memory leak in range selector, chart `load` event listeners leaked on `update`.


TorsteinHonsi (changes requested):
Thanks, the fix is good.

In light of today's meeting and the concern of adding tests for each issue, I think this is a test that can be combined with an existing test. We already have tests for `rangeselector/vertical-update` and `rangeselector/enabled-update`. Probably all these three can be squashed into `rangeselector/update` to avoid unnecessarily creating the expensive stock charts.

TorsteinHonsi (approved):

PR #14867, opened 2021-01-04
bugfix/14858-callout-anchor-line

### LINKED ISSUE #14858: Annotation with default shape 'callout' and align: 'right' don't display the line to point
When an annotation is displayed with the shape: 'callout' and align: 'right', the line, that should indicate the original anchor place is not displayed. 

#### DEMO
https://jsfiddle.net/BlackLabel/abyfm931/

#### Product version
HC v8.2.2

#### Affected browser(s)
All

khlieng:
Fixed #14858, left and right-aligned callout labels with the anchor above or below sometimes missed anchor lines when there was no room for a chevron.


TorsteinHonsi (approved):
Thanks!

PR #14870, opened 2021-01-04
bugfix/14826-candlestick-tooltip-wrong-color

### LINKED ISSUE #14826: Tooltip pointFormatter not using the current candlestick color
Hi,

Here is the link to the Highchart forum. It contains the problem that i found for the upColor candlestick for the tooltip.

https://www.highcharts.com/forum/viewtopic.php?p=162437#p162437

wchmiel:
Fixed #14826, wrong tooltip border color in candlestick series.


TorsteinHonsi (changes requested):
Thanks for the PR!

We should try to resolve this without bringing in the OHLC/Candlestick specific `upColor` property in the Tooltip code.

In fact, I think that `point.color` should reflect the series-level `upColor` whenever the point is up. We can do that by overriding the `resolveColor` function on the `OHLCPoint`:

```ts
    public resolveColor(): void {
        super.resolveColor();
        if (this.open < this.close && !this.options.color) {
            this.color = this.series.options.upColor;
        }
    }
```

Since candlestick inherits from OHLC, this will apply to both instead of only the candlestick. 

Now after this change, you can go over OHLC and candlestick's `pointAttrib` functions. Since `point.color` now reflects the series' `upColor`, those can probably be simplified.

As for the unit test, we discussed in last meeting an effort to reduce the amount of generated charts. So in this case, I suggest that the test can be combined with the existing `unit-tests/series-candlestick/linecolor-up` and renamed to something general like `color`.

TorsteinHonsi:
Thanks! I'm unsure about the changed dot color for OHLC in the visual diffs. They now get a darker color. Is that correct?

wchmiel:
No, dots darker color is not correct. I'll fix it in a minute, thanks!

TorsteinHonsi:
Thanks! Please resolve conflicts, then it's ready to merge.

TorsteinHonsi (approved):

PR #14871, opened 2021-01-04
bugfix/10322-data-parse-dates

### LINKED ISSUE #10322: Data module parses all dates as local
#### Expected behaviour
To detect whether they are local or UTC. 

#### Actual behaviour
Always local.

* Minimal demo: http://jsfiddle.net/highcharts/2scbLq5e/

#### Product version
7.0.3

#### Discussion
Responsible code:
https://github.com/highcharts/highcharts/blob/f0d795ebc1308b228898404f943e1dee0645865b/js/modules/data.src.js#L1867-L1880

In order to fix this, we must do it conditionally so that we still parse date strings without timezone information as local dates. We must add a check for whether timezone information exists in the string, for example `2018-03-13T17:00:00+00:00`, `2018-03-13T17:00:00Z`, `2018-03-13T17:00:00 CET` or `2018-03-13T17:00:00 GMT-0400`. It is complex and not standardized.

bre1470:
Fixed #10322, limit local timezone to timestamps without timezone.

### DIFF
--- a/ts/Extensions/Data.ts
+++ b/ts/Extensions/Data.ts
@@ -685,14 +685,6 @@ class Data {
 
     /* *
      *
-     *  Static Properties
-     *
-     * */
-
-    private static timezoneOffset = (new Date()).getTimezoneOffset();
-
-    /* *
-     *
      *  Constructors
      *
      * */
@@ -2183,8 +2175,7 @@ class Data {
             key,
             format,
             dateFormat = this.options.dateFormat || this.dateFormat,
-            match,
-            timezoneOffset = Data.timezoneOffset;
+            match;
 
         if (parseDate) {
             ret = parseDate(val);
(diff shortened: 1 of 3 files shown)


ts/Extensions/Data.ts
```diff
@@ -2201,13 +2237,13 @@ class Data {
...
                         (match as any).getTime() -
-                        (match as any).getTimezoneOffset() *
+                        timezoneOffset *
                         60000
                     );
 
                 // Timestamp
                 } else if (isNumber(match)) {
-                    ret = match - (new Date(match)).getTimezoneOffset() * 60000;
+                    ret = match - timezoneOffset * 60000;
```
TorsteinHonsi, line 2246: We previously substracted the timezone of the actual date of the match. In the new PR, the `timezoneOffset` is statically set on line 692, which means that it will vary depending on DST or not. As a consequence, check out how a non-UTC summer-time date is parsed (it is winter time as of writing this). Compare this to the results with the PR code http://jsfiddle.net/72kon4ch/

TorsteinHonsi (changes requested):
In addition to the inline comment, we should have tests, especially now that we are likely to move this functionality to the data layer. Current date parsing tests are in `unit-tests/data/csv/demo.js`, though for just testing date parsing (without CSV) we could have a separate sample.

TorsteinHonsi (approved):

PR #14874, opened 2021-01-05
bugfix/5244-clippath-pushstate-update

### LINKED ISSUE #5244: The series path is out of clippath when use history.push()
#### Actual behaviour

The line serie is out of clippath.
#### Live demo of the issue

http://jsfiddle.net/2Ls0Lmdf/
#### Steps to reproduce the behaviour
1. Click the switch button
2. Move handle 
3. Line serie is out of clippath.
   zrzut ekranu 2016-04-25 o 15 03 38
#### Affected browser(s)

All browsers

goransle:
Fixed #5244, renderer url was wrong after location was changed using history.pushstate.

## Reported cases with fix implemented
* Vanilla, clip-path http://jsfiddle.net/goransle/9t1u5woc/
* Angular, clip-path https://stackblitz.com/edit/angular-p1dow7?file=src/app/chart.component.ts
* React, xMin + xMax after update state + resize https://stackblitz.com/edit/react-qt2zd9?file=index.js
* Vue, gradient. [Before](https://codepen.io/gewisser/pen/NWNgWwb) | [After](https://codepen.io/goransle/pen/jOMebJN)
* Angular, fillcolor. [Before](https://stackblitz.com/edit/angular-ivy-mbykwn?file=src/app/chart/chart.component.ts) | [After]( https://stackblitz.com/edit/angular-ivy-mgshrh?file=src/app/chart/chart.component.ts)
* Patterns [before](http://jsfiddle.net/goransle/s9L5epuf/) | [After](http://jsfiddle.net/goransle/j8ehb2rv/)


TorsteinHonsi:
Probably this fix belongs entirely in the renderer, as the issue also applies without a Chart instance: http://jsfiddle.net/highcharts/ehLc653k/ . Though it may be hard to find a performant way to to check when we need to update the URL. 

Additionally, remember to fix also the cases mentioned by Pawel: (gradient, clip-paths, filters, patterns).

goransle:
Seems we have a related regression on master with pattern fills. 

Current master with a base tag and url changed via state: http://jsfiddle.net/goransle/1rsw0ckh/
Latest release: http://jsfiddle.net/goransle/kdpswnb6/

Bisect led me to [this commit](https://github.com/highcharts/highcharts/commit/f639b8c2a27473efe82b747df570cdd1eb4b55b6), but I can't quite see how those changes would affect the patterns being drawn. Any ideas @TorsteinHonsi @oysteinmoseng ?

TorsteinHonsi:
I confirm the bisect, and agree it doesn't make sense. But there are more quirks. 

Even with the latest release, the pattern fills disappear if you resize the chart by dragging the separator between the jsFiddle frames.

Another quirk is that in both cases (master and latest release), we have two identical pattern elements with the same id. That should not happen.

goransle:
Created a separate issue for the pattern-fill situation: #14973.

goransle:
Took a stab at a fix that will work for a standalone SVGRenderer as well. Here's an example of that: https://jsfiddle.net/goransle/7tka489b/. Here it is with a chart with a larger dataset with multiple gradients/patterns: https://jsfiddle.net/goransle/c2y71t8j/

TorsteinHonsi (changes requested):
@goransle With #14976 now merged into the master, can you check if this PR is still relevant?

goransle:
@TorsteinHonsi It still applies as this will fix problems with history-based routing in Angular, Vue, and the likes. 

See these examples in Angular :
* Master https://stackblitz.com/edit/angular-ivy-hqkszk?file=src/app/chart/chart.component.ts 
* With fix: https://stackblitz.com/edit/angular-ivy-mgshrh?file=src/app/chart/chart.component.ts

goransle:
Will be fixed by #15068, closing this.

PR #14881, opened 2021-01-06
bugfix/8095-reset-zoom-drilldown

### LINKED ISSUE #8095: Reset zoom button does not show up when performing zooming after drilldown
#### Expected behaviour
Zooming on any drilldown level should result in displaying reset zoom button.
#### Actual behaviour
The reset zoom button does not display when performing zooming on any drilldown level.

http://jsfiddle.net/502k9msp/

1. Perform a drilldown.
2. Zoom anywhere.

karolkolodziej:
Fixed #8095, reset zoom button was not visible after drilldown.

### DIFF
--- a/samples/unit-tests/drilldown/resetzoom/demo.js
+++ b/samples/unit-tests/drilldown/resetzoom/demo.js
@@ -155,6 +155,11 @@ QUnit.test('Drilldown and reset zoom', function (assert) {
         'object',
         'We have a drillUp button'
     );
+    assert.strictEqual(
+        typeof chart.resetZoomButton,
+        'object',
+        'Reset zoom button should also be visible.'
+    );
 
     // Drill up
     chart.drillUp();
--- a/samples/unit-tests/drilldown/resetzoom/demo.js
+++ b/samples/unit-tests/drilldown/resetzoom/demo.js
@@ -160,6 +160,11 @@ QUnit.test('Drilldown and reset zoom', function (assert) {
         'object',
         'Reset zoom button should also be visible.'
     );
+    assert.ok(
+        chart.resetZoomButton.xSetting >
+            chart.drillUpButton.xSetting + chart.drillUpButton.width,
+        'Buttons should not overlap.'
+    );
 
     // Drill up
     chart.drillUp();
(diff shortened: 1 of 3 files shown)


ts/Extensions/Drilldown.ts
```diff
@@ -1633,3 +1634,59 @@ addEvent(Point, 'afterSetState', function (): void {
...
+            this.drillUpButton.align({
+                x: buttonOptions.position.x,
+                y: buttonOptions.position.y,
+                align: 'right'
+            },
+            false,
+            buttonOptions.relativeTo || 'plotBox'
+            );
+        }
+    }
```
sebastianbochan, line 1682: Lines 1668-1681 are exactly the same as 1649-1661. I suggest to extract it and avoid code rendudant.

sebastianbochan (changes requested):

sebastianbochan (approved):

TorsteinHonsi (changes requested):
We already have an event handler for `beforeShowResetZoom` on line 1057. Instead of creating a complicated workaround for our own code, we should modify the existing code to do what we want. The current event handler disables the `resetZoomButton` if the `drillUpButton` is present. That's obviously not what we want with this fix, we want to align them side by side instead.

ts/Extensions/Drilldown.ts
```diff
@@ -68,6 +68,7 @@ declare module '../Core/Chart/ChartLike' {
         drilldown?: Highcharts.ChartDrilldownObject;
         drilldownLevels?: Array<Highcharts.DrilldownLevelObject>;
         drillUpButton?: SVGElement;
+        temporaryDrillUpButton?: SVGElement;
```
TorsteinHonsi, line 71: This line is obsolete

TorsteinHonsi (changes requested):
In addition to the inline comments, there is a fail case. With the [OP demo](http://jsfiddle.net/502k9msp/):

1. Perform drilldown by clicking first column
2. Zoom in
3. Click "Back to"
4. Click first column => Crash

samples/unit-tests/drilldown/resetzoom/demo.js
```diff
@@ -155,6 +155,11 @@ QUnit.test('Drilldown and reset zoom', function (assert) {
         'object',
         'We have a drillUp button'
     );
+    assert.strictEqual(
+        typeof chart.resetZoomButton,
+        'object',
+        'Reset zoom button should also be visible.'
+    );
```
TorsteinHonsi, line 163: In addition to checking for the button's existence, it would probably be a good idea to check that it does not overlap with the drillUp button.

TorsteinHonsi (changes requested):

karolkolodziej:
Thanks for pointing that case! I added a piece of code (together with a proper test) covering that example.

TorsteinHonsi (approved):

PR #14885, opened 2021-01-06
A11y adjustments for new range selector features.

oysteinmoseng:
A11y adjustments for new range selector features.
___
Follow up to #14586 and #14717, ensuring screen reader and keyboard navigation is on track with changes. Native datepickers are still surprisingly sub-optimal with screen readers, but not much we can do except wait for browsers and screen reader vendors to fix.

Also changes the dropdown options to `button.title` content if available, as it is easier to understand than the abbreviations.


TorsteinHonsi (approved):
Thanks!

PR #14887, opened 2021-01-07
bugfix/14876-minPointLength-negative-zero

### LINKED ISSUE #14876: minPointLength not showing bar in case of zero value for scenario where one series has all zero values and other has all negative values
#### Expected behaviour
minPointLength should show small bar in case of zero values

#### Actual behaviour
minPointLength not showing small bar in case of zero values

https://jsfiddle.net/4z5s80ra/

`xAxis: {
        categories: ['c1', 'c2', 'c3']
    },
    series: [
    	{
      	name: 's1',
        data : [0, 0, 0]
      },
      {
      	name: 's2',
        data: [-12, -13.4, -5]
      }
    ],
    plotOptions: {
    	series: {
      	minPointLength: 2
      },
      column: {
      	dataLabels: {
        	enabled: true
        }
      }
    }`

When all values of one series has zero values and other series has only negative values then var is not appearing.
If one of the value is non zero then it works as expected

FYI
When all values of one series has zero values and other series has only negative values then var is not appearing.
image

If one of the value is non zero then it works as expected
image

#### Product version

highcharts: 8.2.2 and 7.2.0

#### Affected browser(s)
chrome

khlieng:
Fixed #14876, `minPointLength` columns did not show for series with only zero values when combined with series with only negative values.

### DIFF
--- a/js/Series/Column/ColumnSeries.js
+++ b/js/Series/Column/ColumnSeries.js
@@ -346,8 +346,9 @@ var ColumnSeries = /** @class */ (function (_super) {
                     // and if there's room for it (#7311)
                     (yAxis.min || 0) < threshold &&
                     // if all points are the same value (i.e zero) not draw
-                    // as negative points (#10646)
-                    dataMin !== dataMax) {
+                    // as negative points (#10646), but only if there's room
+                    // for it (#14876)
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (approved):
Thanks!

PR #14889, opened 2021-01-07
bugfix/14868-gantt-drilldown-labels-postion

### LINKED ISSUE #14868: Incorrect labels position in drilldown when Gantt module is activated
#### Expected behaviour
image

#### Actual behaviour
image

Data labels are initially rendered correctly. **After entering a deeper level of dropdown, data labels are in the incorrect position.** The effect stays even after returning back to the root level.

https://codesandbox.io/s/highcharts-react-demo-forked-tcltb?file=/demo.jsx

1. Uncomment line no. `15` from `demo.jsx` to turn on the Gantt module.
2. Save & refresh the demo project
3. Click on the label of given month to go deeper into the drilldown.
4. **Check how new labels are positioned. (x axis)**
5. Click on the button to go back to root level
6. **Check how the position of labels changed. (x axis)**

#### Product version
Highcharts v8.2.2

#### Affected browser(s)
- Chrome v89.0
- Safari v14.0
- Firefox v84.0

karolkolodziej:
Fixed #14868, axis labels were positioned incorrectly in non-Gantt charts when the Gantt package was imported.

### DIFF
--- a/ts/Core/Axis/GridAxis.ts
+++ b/ts/Core/Axis/GridAxis.ts
@@ -911,7 +911,7 @@ class GridAxis {
         const userLabels = axis.userOptions.labels || {};
 
         // Fire this only for the Gantt type chart, #14868.
-        if (axis.userOptions.grid && axis.userOptions.grid.enabled) {
+        if (axis.options.grid && axis.options.grid.enabled) {
             if (axis.horiz) {
                 if (gridOptions.enabled === true) {
                     axis.series.forEach(function (series): void {
--- a/ts/Core/Axis/GridAxis.ts
+++ b/ts/Core/Axis/GridAxis.ts
@@ -911,13 +911,11 @@ class GridAxis {
         const userLabels = axis.userOptions.labels || {};
 
         // Fire this only for the Gantt type chart, #14868.
-        if (axis.options.grid && axis.options.grid.enabled) {
+        if (gridOptions.enabled) {
             if (axis.horiz) {
-                if (gridOptions.enabled === true) {
-                    axis.series.forEach(function (series): void {
-                        series.options.pointRange = 0;
-                    });
-                }
+                axis.series.forEach(function (series): void {
+                    series.options.pointRange = 0;
+                });
 
                 // Lower level time ticks, like hours or minutes, represent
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (changes requested):
Nice, thanks! I've been wondering why the labels were sometimes slightly off center.

Change requests:
* Is there a specific reason for checking `userOptions.grid.enabled` rather than the usual `options`? Unless looking for things that are set explicitly, we should use the processed `options`.
* Can you also go over the 192 differing visual samples and verify them? Most or all seem to correctly adjust the axis label back to center. But we must also make sure there is not a real regression hiding in between the valid diffs.

karolkolodziej:
You are right, there is no need to use the `userOptions` the `options` works fine in this case. 
I went through all visual differences and each one looked as correct adjustment of label positioning.

ts/Core/Axis/GridAxis.ts
```diff
@@ -910,38 +910,41 @@ class GridAxis {
...
         const userLabels = axis.userOptions.labels || {};
 
-        if (axis.horiz) {
-            if (gridOptions.enabled === true) {
-                axis.series.forEach(function (series): void {
-                    series.options.pointRange = 0;
-                });
-            }
+        // Fire this only for the Gantt type chart, #14868.
+        if (axis.options.grid && axis.options.grid.enabled) {
```
TorsteinHonsi, line 914: On line 910, we already have `gridOptions = options.grid || {}`. That means line 914 can probably be optimized to `if (gridOptions.enabled) {`

ts/Core/Axis/GridAxis.ts
```diff
@@ -910,38 +910,41 @@ class GridAxis {
...
+            if (axis.horiz) {
+                if (gridOptions.enabled === true) {
```
TorsteinHonsi, line 916: Now since we're already checking for `gridOptions.enabled` on line 914, we should be able to safely remove the check on line 916.

TorsteinHonsi (changes requested):
Thanks! There seems to be room for some more micro-optimization.

TorsteinHonsi (approved):

Issue #1151, opened 2012-08-06
Tooltip position with shared: false


pawelfus:
Example: http://jsfiddle.net/27Tau/4/

Tooltip is positioned as for first series when shared is set to false, whenever I hover first os second series. Expected behaviour: the same as shared: true.

PR #14891, opened 2021-01-07
bugfix/14771-inverted-area-tooltip-position

### LINKED ISSUE #14771: Incorrect tooltip position on inverted area chart with adjusted width
#### Expected behaviour
For an inverted area chart with adjusted width, the tooltips should be positioned near the data points, as for a bar chart:
correct_bar_chart

#### Actual behaviour
The tooltips are misplaced:
incorrect_line_chart

http://jsfiddle.net/8qdkvocz/
http://jsfiddle.net/8m1qdbjo/3/

#### Product version
Highcharts 8.2.2

#### Affected browser(s)
Firefox 83.0, Microsoft Edge 87.0.664.55 and others

wchmiel:
Fixed #14771, incorrect tooltip position on inverted chart with adjusted axis width.

### DIFF
--- a/js/Core/Tooltip.js
+++ b/js/Core/Tooltip.js
@@ -289,7 +289,7 @@ var Tooltip = /** @class */ (function () {
      * @return {Array<number>}
      */
     Tooltip.prototype.getAnchor = function (points, mouseEvent) {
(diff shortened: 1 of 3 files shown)


TorsteinHonsi:
Thanks!

I took your idea and refactored the `getAnchor` function a bit. Issue #14771 is the same as #1151, only for inverted charts. So I moved the calculations to the same place and also made sure that plotX was adjusted the same way. Then made sure shared tooltip are placed the same way whether inverted or not.

Now pending CI checks.

PR #14895, opened 2021-01-08
bugfix/14351-solidgauge-pane-flicker

### LINKED ISSUE #14351: Solidgauge pane track flickers on update
It seems that in the new version (8.1.0) the solid gauge started to flick after the update, it has an impact on the React wrapper working - reported [here](https://github.com/highcharts/highcharts-react/issues/260). Not sure if it is a regression or intended behavior.

It could be disabled when the `chart.animation` is set to false: https://jsfiddle.net/BlackLabel/sjmxdzb9/

Demo: https://jsfiddle.net/BlackLabel/dusLaz3h/

khlieng:
Fixed #14351, `solidgauge` pane flickered on `update`.

### DIFF
--- a/js/Core/Animation/Fx.js
+++ b/js/Core/Animation/Fx.js
@@ -79,8 +79,8 @@ var Fx = /** @class */ (function () {
                     var startItem = startSeg[j];
                     var endItem = endSeg[j];
                     // Tween numbers
-                    if (typeof startItem === 'number' &&
-                        typeof endItem === 'number' &&
+                    if (isNumber(startItem) &&
+                        isNumber(endItem) &&
                         // Arc boolean flags
                         !(endSeg[0] === 'A' && (j === 4 || j === 5))) {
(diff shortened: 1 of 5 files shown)


TorsteinHonsi (changes requested):
Thanks!

In addition to a proper test for `NaN`, we should also ask ourselves why a NaN value was passed. I have traced it to the following lines, where `this.left` and `this.top` are `undefined`:

https://github.com/highcharts/highcharts/blob/46c31dc6fb5da66aff8e1c5477fa5dc780d74e36/ts/Core/Axis/RadialAxis.ts#L277-L278

It is probably because we are in the middle of an `axis.update`. We can simply fix this by doing `(this.left || 0)`, though it is problematic that TypeScript has `this.left` as a `number` rather than `number|undefined`.

When fixing this, it allows smooth animation in situations like this: https://jsfiddle.net/gx0th2k1/

TorsteinHonsi (approved):

PR #14899, opened 2021-01-09
ts/new-line-series

bre1470:
Created a new line series and replaced references of core series (previous line series).


Onixmiha (approved):

Onixmiha (approved):

TorsteinHonsi:
> Created a new empty line series with the temporary key `_line`.

How temporary is that? Looks like a hack that doesn't belong in the master, but if it is just a required quick step on the way to the permanent solution then we can merge it.

bre1470:
When all line-specific code is moved to the new `LineSeries`, we can replace the `line` registry and remove `_line` again. Though I have no idea how much work it is. In many places like axis, label, legend, pointer, stock are references, that only directly work with `Series` class and rely on the inheritance of the old `line` series by all other series (so in future the reduced core series). All these code lines have make use of `SeriesRegistry.series` instead of `SeriesRegistry.seriesTypes.line` before we can even think about moving line-specific code to the new `LineSeries` and removing `_line`.

bre1470:
@TorsteinHonsi Please review ones more.

TorsteinHonsi (approved):
Thanks, looks good!

Now there are other functions that can probably be moved over too, like `getSymbol`, `getGraphPath`, 'getZonesGraphs`, though it is not very important.

TorsteinHonsi (approved):

PR #14900, opened 2021-01-09
bugfix/13572-incorrect-path-rendering-for-area

### LINKED ISSUE #13572: Incorrect path rendering for area
#### Expected behaviour
The ability to correctly set the color of the line surrounding the area using the CSS options.

#### Actual behaviour
When we want to set the line color surrounding an area using the CSS:
`.highcharts-area { 
	stroke-width: 3;
	stroke: red;
}
`
There is no border on one edge because the path is not closed:
https://jsfiddle.net/BlackLabel/7d0fq2ah/

Possible workaround - improved core function responsible for rendering this path (adding 'Z'):
https://jsfiddle.net/BlackLabel/wnypd0gs/

#### Product version
Highcharts v8.1.0

madepiet:
Fixed #13572, area series fill was not closed, making it hard to add a stroke for the area itself.

### DIFF
--- a/js/Core/Animation/Fx.js
+++ b/js/Core/Animation/Fx.js
@@ -295,7 +295,7 @@ var Fx = /** @class */ (function () {
                 // causing the middle two points to be sliced out, since an area
                 // path starts at left, follows the upper path then turns and
                 // follows the bottom back.
-                var segmentToAdd = arr[arr.length / positionFactor - 1].slice();
+                var segmentToAdd = arr[Math.floor(arr.length / positionFactor) - 1].slice();
                 // Disable the first control point of curve segments
                 if (segmentToAdd[0] === 'C') {
                     segmentToAdd[1] = segmentToAdd[5];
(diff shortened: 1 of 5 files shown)


sebastianbochan:
Looks good 👍

sebastianbochan (approved):

TorsteinHonsi (changes requested):
Thanks!

Sideways animation is broken (when adding point with `shift`): https://jsfiddle.net/y6Lz03pa/

sebastianbochan:
It looks like a problem with double `areaPath.push(['Z']);` 

Suggested fix:
https://jsfiddle.net/BlackLabel/e2p36k9u/

madepiet:
The problem seems to be more complex as it works fine with this code, but you can see the jump on the left side. I wonder how this is supposed to work - smoothly moving from the left side?

In the code, I only push 'Z' once, but it also requires a change in the animation function. Because right now we move the whole hardcoded path which is not correct.

TorsteinHonsi:
@sebastianbochan 

> Is it the desired effect?

No, as you can see with your fiddle, the series doesn't move at all. It is supposed to slide sideways to make room for the new point.

> In the code, I only push 'Z' once, but it also requires a change in the animation function. Because right now we move the whole hardcoded path which is not correct.

Off the top of my head, I think what happens in the animation is that it splits the area path in two and injects the new point in the middle. So you need to make sure the added length of the path segments (the `Z`) is accounted for.

TorsteinHonsi (approved):

PR #14901, opened 2021-01-11
bugfix/14053-plotband-add-update-redraw

### LINKED ISSUE #14053: Plot Bands do not show occasionally on v8.1.1 and up
Hello! I've recently tried updated Highcharts from version **8.1.0** to **8.1.1** and I have come across a bug where my plot bands do not always show after a full redraw of my chart.

I have a Refresh button for the graph that removes all series, plot bands, and plot lines, then re-adds them all and calls `chart.redraw()` once they are all added. I have noticed that the plot bands are sometimes not redrawn.

I add all of my plot bands via the [addPlotBand](https://api.highcharts.com/class-reference/Highcharts.Axis#addPlotBand) function. All of the plot band objects I add are found in the `axis.userOptions.plotBands` and `axis.options.plotBands` arrays, but only the **last item** in these arrays is drawn and found in my `axis.plotLinesAndBands` array.

I dug through the changelog between these versions and I only saw [this issue](https://github.com/highcharts/highcharts/pull/13444/commits) that touched on these items. Maybe they are related? @Izothep any ideas?

#### Expected behaviour

Adding multiple plot bands should render all of the plot bands

#### Actual behaviour

Adding multiple plot bands sometimes only renders the last plot band added.

This is hard to reproduce reliably, it does not always happen. However I do believe it happens more often when removing and re-adding plot bands back to back with the same `id`.

#### Product version
Highcharts/Highstock v8.1.1 and up

#### Affected browser(s)

Tested on Chrome 84.0.4147.125

khlieng:
Fixed #14053, plot bands added before `update` with `redraw` set to `false` were not added when dynamically adding plot bands between the `update` and a `redraw`.

### DIFF
--- a/js/Core/Axis/Axis.js
+++ b/js/Core/Axis/Axis.js
@@ -2763,12 +2763,12 @@ var Axis = /** @class */ (function () {
             }
             // custom plot lines and bands
             if (!axis._addedPlotLB) { // only first time
+                axis._addedPlotLB = true;
                 (options.plotLines || [])
                     .concat(options.plotBands || [])
                     .forEach(function (plotLineOptions) {
                     axis.addPlotBandOrLine(plotLineOptions);
                 });
-                axis._addedPlotLB = true;
             }
         } // end if hasData
         // Remove inactive ticks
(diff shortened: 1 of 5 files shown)


TorsteinHonsi (approved):
Thanks!

Issue #14872, opened 2021-01-05
Legend title not announced by screen readers


oysteinmoseng:
#### Expected behaviour
When adding a title to the chart legend, this should be read by screen readers.

#### Actual behaviour
Legend titles are not exposed to assistive tech.

#### Live demo with steps to reproduce
https://jsfiddle.net/e32sfayo/

oysteinmoseng:
Note: Should also add basic support for color axis announcements - e.g. axis range.

PR #14905, opened 2021-01-11
Added a11y support for legend title. Touches #14872.

oysteinmoseng:
Added a11y support for legend title. Touches #14872.
___
Legend title is now exposed as expected to assistive technology if defined.


TorsteinHonsi (approved):
Thanks!

TorsteinHonsi (approved):

PR #14907, opened 2021-01-12
Docs: Clarified confusing default value #14893.

### LINKED ISSUE #14893: Specifying axis label format breaks thousand separator
#### Expected behaviour
yAxis.labels.format is specified using `{value}`. The values on the labels should adhere to the normal thousands separator rules.

#### Actual behaviour

No thousands separator is used, even though the format option contains nothing more than the default setting `{value}`. Numeric symbols are also not in use, although that part might be by design.

https://jsfiddle.net/zbw23s7h/1/

If the line `format: '{value}'` is removed, the thousands separator works normally.

#### Product version
Highcharts ~8.2.0

#### Affected browser(s)

Tested in Chrome, Firefox, Edge.

DJTechnoo:
Fixed #14893, removed misleading default value in the `labels.format` docs.

### DIFF
--- a/js/Core/Axis/Axis.js
+++ b/js/Core/Axis/Axis.js
@@ -3929,14 +3929,17 @@ var Axis = /** @class */ (function () {
              */
             enabled: true,
             /**
-             * A [format string](https://www.highcharts.com/docs/chart-concepts/labels-and-string-formatting)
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #14908, opened 2021-01-12
bugfix/11494-outside-tooltip-zindex

### LINKED ISSUE #11494: Tooltip z-index problem with outside:true
#### Expected behaviour
When the following parameters are met the tooltip should be completely opaque:
1. A <section> around the chart (#outer in the jsfiddle) has a z-index greater than 3
2. The container div for the chart has position: relative (.container in the jsfiddle)
3. The tooltip has outside: true
4. The tooltip span has a fully opaque color (i.e. rgba(100,100,100,1)

#### Actual behaviour
Tooltip is not opaque. But if you change any one of the items above (other than 4) then it will be opaque.

https://jsfiddle.net/te1qc2wn/

The issue seems to be that when Outside is true, the z-index of the tooltip is set to 3 and there is no way to change it using css. In previous versions the z-index was not set to 3 and it worked fine.

#### Product version
Is not a problem in Highcharts version 7.1.1, but is a problem in more recent versions.

#### Affected browser(s)
Only tested in Chrome

khlieng:
Fixed #11494, setting `z-index` on tooltip with `outside` set to `true` through chart options was not possible.

### DIFF
--- a/js/Core/Tooltip.js
+++ b/js/Core/Tooltip.js
@@ -414,11 +414,11 @@ var Tooltip = /** @class */ (function () {
      * @return {Highcharts.SVGElement}
      */
     Tooltip.prototype.getLabel = function () {
-        var _a, _b;
+        var _a, _b, _c;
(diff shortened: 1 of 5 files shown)


TorsteinHonsi (approved):
Thanks!

PR #14912, opened 2021-01-12
bugfix/14910-centerInCategory-double-zoom

### LINKED ISSUE #14910: The centerInCategory property fails on zoom
#### Actual behaviour
The columns overlap each other after zooming twice.

https://jsfiddle.net/BlackLabel/art21mog/

#### Product version
8.2.2

#### Affected browser(s)
All

khlieng:
Fixed #14910, `centerInCategory` broke after zooming twice.

### DIFF
--- a/js/Core/Axis/Axis.js
+++ b/js/Core/Axis/Axis.js
@@ -1820,10 +1820,7 @@ var Axis = /** @class */ (function () {
             }
         }
         else if (axis.stacking) {
-            var keys = Object.keys(axis.stacking.stacks);
-            if (!(keys.length === 1 && keys[0].slice(-5) === 'group')) {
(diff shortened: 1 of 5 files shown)


TorsteinHonsi (commented):
I'm not sure why this group stack should have special treatment versus a normal stack. I assume a normal stack is rebuilt before rendering. Can you do some digging and see why doesn't happen with the group stack?

khlieng:
@TorsteinHonsi so it turns out `hasStackedSeries` in `chart.redraw` is false because `series.stacking` is not set, so I should probably just include `centerInCategory` in that check as well instead?

TorsteinHonsi:
Yes, that sounds like the best solution!

TorsteinHonsi (approved):

PR #14917, opened 2021-01-13
docs/14710-column-pointwidth

### LINKED ISSUE #14710: pointWidth typings mismatch
#### Expected behaviour
In my code I do:
image

I consider `null` to be expected value for `pointWidth` because the doc says so: https://api.highcharts.com/highcharts/series.bar.pointWidth.

#### Actual behaviour
I get type error:
image

I guess, that
`pointWidth?: number;`
should be something like
`pointWidth?: number | null;`

n/a

#### Product version
Highcharts 8.2.2

#### Affected browser(s)
n/a

bre1470:
Fixed #14710, added `undefined` as the valid type/value for default behavior  of `column.pointWidth` and inherited series options.

### DIFF
--- a/js/Accessibility/Components/RangeSelectorComponent.js
+++ b/js/Accessibility/Components/RangeSelectorComponent.js
@@ -121,8 +121,8 @@ extend(RangeSelectorComponent.prototype, /** @lends Highcharts.RangeSelectorComp
         }
     },
     /**
-     * @private
      * Hide buttons from AT when showing dropdown, and vice versa.
+     * @private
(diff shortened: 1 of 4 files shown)


TorsteinHonsi (changes requested):
The code doesn't distinguish between `null` and `undefined`. The option is handled in a `pick`. So instead of further bloating the type I think it is better that we stick with `number|undefined` and change the description text instead.

TorsteinHonsi (approved):

PR #14926, opened 2021-01-13
Fixed missing import after merge.

oysteinmoseng:
Fixed missing import after merge.
___
The `stripHTMLTagsFromString` call is needed here - not for security, but to avoid html code being part of the label in case of HTML titles.


bre1470 (approved):
Looks good! 👍

PR #14943, opened 2021-01-15
bugfix/14932-update-compare-plotOptions

### LINKED ISSUE #13251: Switching from non-compared to compared does not work
#### Expected behaviour
When the property `plotOptions.series.compare='compare'` is set, the chart should go into compare mode

#### Actual behaviour
This only works if the compare is set on first load, doesn't work if you dynamically update the chart with 'compare'

https://jsfiddle.net/npxyvg6t/3/

#### Product version

#### Affected browser(s)
Tested in Chrome and Firefox

### LINKED ISSUE #14932: Chart plotOption value "compare" does not update dynamically
**Expected behaviour**
I expect the chart to reload itself when chart.update() is called with different compare options.

**Actual behaviour**
When I change the compare option and I update the chart nothing happens.
Live demo with steps to reproduce

https://codepen.io/justinphung/pen/JjRejJG

Click on button 'Toggle Compare'
Notice that the subtitle changes but the x- and y-axis does not.
When changing the compare value in the code for initializing the chart the x- and y-axis is different after codepen reloads.

**Product version**
Highstock v8.2.2

**Affected browser(s)**
Chrome v87

khlieng:
Fixed #14932, #13251, updating `compare` through `plotOptions` did not work.

Closes #13251.


TorsteinHonsi (approved):
Thanks!

Issue #1338, opened 2012-10-30
labels.useHTML + labels.step renders labels in top left corner slightly visible


dwick:
here is an example: http://jsfiddle.net/nXu2T/

PR #14945, opened 2021-01-18
bugfix/14904-double-click-throws-icon-out-of-chart

### LINKED ISSUE #14904: Gantt: animation a triangle symbol in a drop down table on yAxis, after double click.
#### Expected behaviour
The triangle animation should point to the position of the vertex to the right if the drop down menu is closed.
When the menu is opened, the vertex of the triangle should point downwards.

Screen Shot 2021-01-11 at 4 02 15 PM

#### Actual behaviour
When you drop down the menu at the table on yAxis, collapse it and double click it, when it has not collapsed completely, the triangle symbol changes animation.

template: https://jsfiddle.net/BlackLabel/w8zrj0kn/

Screen Shot 2021-01-11 at 4 08 54 PM

Screen Shot 2021-01-11 at 4 03 09 PM

#### Product version
Highcharts Gantt JS v8.2.2 (2020-10-22)

#### Affected browser(s)
All

DJTechnoo:
Fixed #14904, drop-down icon on tree grid child disappearing after two clicks on parent.


ts/Core/Axis/TreeGridTick.ts
```diff
@@ -179,6 +179,8 @@ namespace TreeGridTick {
         // Set the new position, and show or hide
         if (!shouldRender) {
             icon.attr({ y: -9999 }); // #1338
+        } else {
+            icon.attr({ y: 0 }); // #14904
```
TorsteinHonsi, line 183: More compact:

`icon.attr({ y: shouldRender ? 0 : -9999 }); // #1338, #14904`

TorsteinHonsi (changes requested):
Thanks! Just a minor filesize optimization.

ts/Core/Axis/TreeGridTick.ts
DJTechnoo: Thank you,! I've made the change

TorsteinHonsi (approved):

TorsteinHonsi:
All good, just waiting for CI to run.

PR #14946, opened 2021-01-18
docs/14401-timeline-legendType

### LINKED ISSUE #14401: Options other than "point" for legendType?
https://github.com/highcharts/highcharts/blob/5ee7fe79e88ebe870b6a7874c9ccb775da3d2b3d/js/Series/TimelineSeries.js#L87-#L87

What are the other options other than "point"? I'm trying to have my timeline chart display unique values for each of type of event not every one in the legend.

khlieng:
Fixed #14401, removed `legendType` from `timeline` series API docs.


TorsteinHonsi (approved):
Thanks!

PR #14953, opened 2021-01-19
bugfix/14110-html-label-hover-ie

### LINKED ISSUE #14110: label.element.contains gives an error in SVGLabel on Internet Explorer11
#### label.element.contains or span.element.contains throws an unhandled exception
on highcharts for ie11. It happens on hover

#### Product version
<!--- Highcharts, :8.0.0>

khlieng:
Fixed #14110, hovering labels with `useHTML` set to `true` threw in IE.


TorsteinHonsi (approved):
Thanks!

PR #14957, opened 2021-01-20
bugfix/14954-export-styledMode-ie

### LINKED ISSUE #14954: Client Side Exporting is raising an error in IE11
#### Expected behaviour

The Client Side Exporting of PNG/JPG should work in IE11 when including the exporting and offline-exporting modules.

#### Actual behaviour

When going to download, get the error message: "Object doesn't support property or method 'remove'"

Unfortunately it is an IE11 issue and jsFiddle doesn't support. This is the [feature example](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/exporting/offline-download-demo).

I think the issue is related to this [change](https://github.com/highcharts/highcharts/pull/14097/files#diff-bcca82be8ec262727e0d70b38b4c0dfcd827c11bd3f355aa5f5e90161ff2e87dR1875). IE11 doesn't support the method remove, so changing the code from removeChild to remove is causing the issue.

I confirmed removing the `remove()` method does allow the export to process properly.

#### Product version
v8.2.2

#### Affected browser(s)
IE11

khlieng:
Fixed #14954, offline export with `styledMode` enabled sometimes failed in IE.


TorsteinHonsi (approved):
Thanks!

PR #14961, opened 2021-01-20
enhancement/data-layer

bre1470:
Restoration of valid branch for https://github.com/highcharts/highcharts/pull/13912
To-Do:
- [x] DataTable
- [x] DataTableRow
- [x] DataConverter converts between primitives, Date, and DataTable
- [x] DataJSON to support serialization of any instance
- [x] DataModifier
- [x] ChainModifier
- [x] GroupModifier
- [x] RangeModifier
- [x] SortModifier
- [x] CSVStore
- [x] GoogleStore
- [x] HTMLTableStore
- [x] CSVParser
- [x] GoogleSheetsParser (DataJSON support is missing)
- [x] HTMLTableParser
- [x] unit-tests
- [x] demo samples
- [x] Data module makes use of data layer
- [x] SeriesDataConverter creates DataTable from data option and vice versa
- [x] Removing `Data` from class/file names in inherited modifiers, parsers, and stores
- [x] Events have to support context identifier (Event.detail: Record<string, string>)
- [x] DataTable and DataTableRow have to take a context identifier or provide their own id as context to events
- [x] DataStore has to provide a context identifier (Event.detail)
- [x] DataStore needs static function and properties to register Store class and create instances
- [x] Series makes use of data layer
- [x] Series should understand and listen to DataTable
- ~Series.update should go to SeriesDataConverter~ Not possible, as point data has to be kept as a reference.
- Unclear task: SeriesDataConverter needs to fire events instead of serial returning results — Why?
- Unclear task: Utilities.fireEvent support optionally zero-timeout calls — Why?


sebastianbochan:
@bre1470
Only what comes to my mind is verifying our files (from the previous branch) with the new branch. Does it make sense?

bre1470:
@sebastianbochan I do not think, that it is necessary. I did not needed to cherry pick. All the old commits (= history) are the same as the hash values. I just jumped back before my last six commits, where somehow 20 unrelated files were changed and the merge conflicts with master could not be solved in any way.

ts/Data/Parsers/DataParser.ts
```diff
@@ -149,18 +149,19 @@ implements DataEventEmitter<TEventObject>, DataJSON.Class {
...
      */
     public static getTableFromColumns(
         columns: Array<Array<DataTableRow.CellType>> = [],
-        headers: Array<string> = []
+        headers: Array<string> = [],
+        dataTable?: DataTable
     ): DataTable {
         const columnsLength = columns.length,
-            table = new DataTable();
+            table = dataTable?.clone() || new DataTable();
```
bre1470, line 164: Avoid bloating code by using optional selectors. Use inline condition.

ts/Data/Parsers/CSVParser.ts
```diff
@@ -504,11 +504,14 @@ class CSVParser extends DataParser<DataParser.EventObject> {
...
+     * Optional DataTable to use as base for the parsed dataTable
+     *
      * @returns {DataTable}
      * A DataTable from the parsed CSV
      */
-    public getTable(): DataTable {
-        return DataParser.getTableFromColumns(this.columns, this.headers);
+    public getTable(dataTable?: DataTable): DataTable {
+        return DataParser.getTableFromColumns(this.columns, this.headers, dataTable);
     }
```
bre1470: Modifying should not be part of get-functions.

ts/Data/Parsers/GoogleSheetsParser.ts
```diff
@@ -273,11 +273,14 @@ class GoogleSheetsParser extends DataParser<DataParser.EventObject> {
...
+     * Optional DataTable to use as base for the parsed dataTable
+     *
      * @return {DataTable}
      * A DataTable from the parsed Google Sheet
      */
-    public getTable(): DataTable {
-        return DataParser.getTableFromColumns(this.columns, this.headers);
+    public getTable(dataTable?: DataTable): DataTable {
+        return DataParser.getTableFromColumns(this.columns, this.headers, dataTable);
     }
```
bre1470: Modifying should not be part of get-functions.

ts/Data/Parsers/HTMLTableParser.ts
```diff
@@ -273,11 +273,14 @@ class HTMLTableParser extends DataParser<DataParser.EventObject> {
...
+     * Optional DataTable to use as base for the parsed dataTable
+     *
      * @return {DataTable}
      * A DataTable from the parsed HTML table
      */
-    public getTable(): DataTable {
-        return DataParser.getTableFromColumns(this.columns, this.headers);
+    public getTable(dataTable?: DataTable): DataTable {
+        return DataParser.getTableFromColumns(this.columns, this.headers, dataTable);
     }
```
bre1470: Modifying should not be part of get-functions.

ts/Data/Stores/CSVStore.ts
```diff
@@ -239,7 +239,7 @@ class CSVStore extends DataStore<CSVStore.EventObjects> implements DataJSON.Clas
                 table: store.table
             });
             store.parser.parse({ csv });
-            store.table = store.parser.getTable();
+            store.table = store.parser.getTable(store.table);
```
bre1470, line 242: Modifying should not be part of get-functions.

ts/Data/Stores/GoogleSheetsStore.ts
```diff
@@ -142,7 +142,7 @@ class GoogleSheetsStore extends DataStore<GoogleSheetsStore.EventObject> impleme
             dataType: 'json',
             success: function (json: Highcharts.JSONType): void {
                 store.parser.parse(json);
-                store.table = store.parser.getTable();
+                store.table = store.parser.getTable(store.table);
```
bre1470, line 145: Modifying should not be part of get-functions.

ts/Data/Stores/HTMLTableStore.ts
```diff
@@ -190,7 +190,7 @@ class HTMLTableStore extends DataStore<HTMLTableStore.EventObjects> implements D
             eventDetail
         );
 
-        store.table = store.parser.getTable();
+        store.table = store.parser.getTable(store.table);
```
bre1470, line 193: Modifying should not be part of get-functions.

bre1470 (commented):
I believe it is clearer if these changes go into a `DataTable.insertRows` function uses by stores, than mixing all the getter-functions of parsers with setter code.

oysteinmoseng (approved):
Milestone! No comments from me

ts/Extensions/Data.ts
```diff
@@ -1429,95 +1453,34 @@ class Data {
...
+                    csv: csv,
+                    startRow: options.startRow,
+                    endRow: options.endRow,
+                    startColumn: options.startColumn,
+                    endColumn: options.endColumn,
+                    firstRowAsNames: options.firstRowAsNames,
+                    switchRowsAndColumns: options.switchRowsAndColumns,
+                    decimalPoint: options.decimalPoint,
+                    itemDelimiter: options.itemDelimiter,
+                    lineDelimiter: options.lineDelimiter
```
TorsteinHonsi, line 1477: Why the repetition? Can we just pass on either `options` or `merge(options)`?

ts/Extensions/Data.ts
```diff
@@ -1533,68 +1496,43 @@ class Data {
...
+            htmlStore = this.dataStore = new HTMLTableStore(
+                new DataTable(),
+                {
+                    table: typeof table === 'string' ? table : table.id || ''
+                },
+                new HTMLTableParser({
+                    startRow: startRow,
+                    endRow: endRow,
+                    startColumn: startColumn,
+                    endColumn: endColumn,
```
TorsteinHonsi, line 1519: Use object literal shorthand syntax

ts/masters/highcharts.src.ts
```diff
@@ -29,6 +29,7 @@ import '../Core/Legend.js';
 import Chart from '../Core/Chart/Chart.js';
 import '../Extensions/ScrollablePlotArea.js';
 import '../Extensions/Stacking.js';
+import DataTable from '../Data/DataTable.js';
```
TorsteinHonsi, line 32: Do we need this in the core at this point? This import is probably responsible for the bundle size going up by 3 kB (gzipped).

At this point we are not using the data layer in the core. The end goal is to replace the series' internal data handling with the data layer, so when we have accomplished that, the bundle size should be balanced. But we should not load redundant code into a bundle that will reach production with the next release.

samples/highcharts/demo/series-table/demo.details
```diff
@@ -0,0 +1,10 @@
+---
+name: Series table
+authors:
+  - Sophie Bremer
+js_wrap: b
+tags:
+  - Highcharts demo
+categories:
+  - Dynamic charts
```
TorsteinHonsi, line 9: Should this sample be in www.highcharts.com/demo at this point? Also, it's not working... JS error on adding row.

samples/highcharts/studies/data-layer/demo.mjs
```diff
@@ -0,0 +1,191 @@
+import '../../../../code/es-modules/Core/Renderer/SVG/SVGRenderer.js';
```
TorsteinHonsi, line 1: This demo fails for me with JS errors...

ts/Data/DataConverter.ts
```diff
@@ -0,0 +1,646 @@
...
+            floatVal = parseFloat(trimInsideVal);
+
+        let result: ('string' | 'Date' | 'number') = 'string',
+            dateVal;
+
+        // is numeric
+        if (+trimInsideVal === floatVal) {
+
+            // If the number is greater than milliseconds in a year, assume
+            // datetime.
```
TorsteinHonsi, line 340: There's some regression, it fails to guess Date here: https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/data/delimiters/

ts/Core/Utilities.ts
```diff
@@ -461,6 +461,41 @@ namespace error {
     export const messages: Array<string> = [];
 }
 
+// eslint-disable-next-line valid-jsdoc
+/**
+ * Reduces tree-like objects to a simple object with keys in dot syntax.
+ * @private
+ */
+function flat(
```
TorsteinHonsi, line 469: Apparently the `flat` and `unflat` functions are not used at all, so they should be removed. If we need them down the line, only add them to the core if they are actually used by the core.

ts/Data/AJAX/Authenticators/AuthenticatorType.d.ts
```diff
@@ -0,0 +1,31 @@
+/* *
+ *
+ *  Data Layer
+ *
+ *  (c) 2012-2020 Torstein Honsi
```
TorsteinHonsi, line 5: I didn't write this... And not in 2012 or 2020.

If determining the real author for the files is not straight forward, we should write `(c) 2020-2021 Highsoft`

TorsteinHonsi (changes requested):
Thanks for all the great work!

samples/highcharts/studies/data-layer/demo.mjs
goransle: This demo is a bit outdated in a few ways, we have changed how the modifiers work, the DataSeriesConverter and the DataTable. So it might be simplest to retire this iteration and create a new demo for the series-dataTable integration when that is ready.

ts/Extensions/Data.ts
bre1470: As the options are a combination for table and parser constructor, this assignment is nearly as good as it can get. Your suggestions would flood each constructor with options for the other one. Or is this what your meant?

bre1470: It is not possible because of overlapping options. :(

ts/Data/DataConverter.ts
bre1470: Is it fixed now? When I do a local test on the branch, it looks the same with your fiddle. I need more information about the failure to continue.

bre1470: Skjermbilde 2021-05-31 kl  12 41 45

goransle: Seems to still be broken for me: https://jsfiddle.net/goransle/9437yp8g/ (also note that the CSV in the sample is slightly wrong on the first line 😬 )

bre1470: Now I see. Chrome breaks, Firefox and Safari not.

bre1470: Please confirm, that it works in Chrome now.

bre1470:
The last genuine commit introduces a new modifier function, that is async and is used by the table for bulk processing during attachment of a modifier.
*async:*
- Modifier.modify(table) = Promise(modifyTable(table))
*sync:*
- Modifier.modifyCell(table) 
- Modifier.modifyColumns(table)
- Modifier.modifyRows(table)
- Modifier.modifyTable(table)

With that design, only one setter in DataTable needs to be async.
*async:*
- Table.setModifier(modifier)
*sync:*
- Table.setCell()
- Table.setColumnAlias()
- Table.setColumns()
- Table.setRows()
- Table.setRowObjects()

ts/Data/DataConverter.ts
goransle: It works! 🙌

TorsteinHonsi (commented):
Thanks! Looks good now, it's reassuring that the core files are not modified (yet).

The changes are now concentrated in the Data Module, which has seen a significant increase in file size:
|                                | master | enhancement/data-layer |
|--------------------------------|--------|------------------------|
| data.js (compiled)             | 16 kB  | 50 kB                  |
| data.js (compiled and gzipped) | 6 kB   | 15 kB                  |

An increase in file size is expected, given the added structure and improvements in architecture, but tripling the size is quite much.

Are we sure we are not loading anything into this module that is not necessary at this point?

bre1470:
I will check ones more. I recall dead code some months ago.

bre1470:
Improved the code, but still no significant change in size. 🤔

goransle:
Mind that the `data.js (compiled)` would be the compiled `data.src.js` which includes jsdoc and comments. There also seems to be some code in `Data.ts` that could be removed if we find a way to handle the `columns` and `rows` options. (Might be possible with a pure DataStore?)

TorsteinHonsi:
> Mind that the data.js (compiled) would be the compiled data.src.js which includes jsdoc and comments

No, it shouldn't be. I compared `data.js` after running `gulp compile` which would run Closure Compiler, which strips all jsdocs and comments. Then I compared the gzipped result of the same. Comparing `data.src.js` is not relevant.

goransle:
> > Mind that the data.js (compiled) would be the compiled data.src.js which includes jsdoc and comments
> 
> No, it's not including jsdoc and comments. I compared `data.js` after running `gulp compile`. Then I compared the gzipped result of the same. Comparing `data.src.js` is not relevant.

Right, I misread my own code in the `write-file-sizes` gulp task 😅

TorsteinHonsi (approved):

TorsteinHonsi:
Approved 🎉 

Pending approvals by @goransle, @sebastianbochan and @wchmiel .

goransle (approved):
Only thing that came to mind is that the Google sheet implementation is outdated, but as it is not implemented in the data-module in this PR it should not block a merge.

sebastianbochan:
Looks amazing! 

ps. How about the Googlesheet like @goransle said.

sebastianbochan (approved):

wchmiel (approved):
Looks good, great job!

bre1470:
Yes, good job everyone! 🎉 
Ready for merge, @TorsteinHonsi .

bre1470:
Yes, good job everyone! 🎉 

> Only thing that came to mind is that the Google sheet implementation is outdated, but as it is not implemented in the data-module in this PR it should not block a merge.

> ps. How about the Googlesheet like @goransle said.

I suggest, that we address this in the layout branches, were we need it for our demo.

PR #14965, opened 2021-01-21
bugfix/14439-not-stacking-on-first-render

### LINKED ISSUE #11439: Highcharts Gantt: addPoint called once, while using parent (uniqueNames), creates 2 items
#### Expected behaviour
Calling `addPoint({<options>})` should add one point to the chart, under the proper `parent`, when using `yAxis.uniqueNames`

#### Actual behaviour
Works fine on the first call, but any following one appends 2 Points instead (first, a clone of the previously added one, and then the new one).

Actually, it gets weirder:
- On 1st call, the Point is added correctly. From then on, any addition of a new Point also appends a duplication of itself (or the one before)
- Calling `update({})` twice after that seems to auto correct some of the cases (the duplicates are removed etc...)
- Adding a Point as child of a Task (parent) that is followed by other Tasks (parent) "shifts" the TreeGrid's style and makes its (the added Point's) label look like one of a Task (parent), meaning `bold` and `drop down icon`.

**_I haven't tried updating the whole series (and including the new data) like I usually do, I was just fiddling with the other Isssue's demo and noticed this behaviour._**

My demo is based on #11350 's. There, the issue was adding an Item (Series, not a Point) when the last current item of the graph was collapsed created problems making drawing the new Item (Series, in his case) weirdly and preventing the user from expanding the other Point. A "fix" was found but:
> weirdly enough, the bug still occur if (and only if) a second item is added while the other one is collapsed. But if you expand, add the second, then collapse again, the following item additions don't cause the visual bug

_Originally posted by @FilipeCosta06 in https://github.com/highcharts/highcharts/issues/11350#issuecomment-512147775_

In my demo, I have buttons to both add Points with and without the mention issue's fix.

### Use Case
Just to clarify: I wish to add Points to a Gantt chart. These new Points may or may not be children (sub tasks) of others and said parent may or may not be collapsed and may or may not be the last item.

http://jsfiddle.net/v2uqn3cz/4/
(_optional_) 1- Open the console, I print the Series' data's userOptions there
(_optional_) 2- Expand the "Develop" Task (parent)
3- Click one of the "Add to develop" button
4- Repeat to start seeing some weird stuff
4- Click the "update({})" twice button and see how it, sometimes, fix the graphic by removing the duplicated data

#### Product version
Highcharts Gantt v7.1.2

#### Affected browser(s)
Chrome, Version 74.0.3729.169 (Official Build) (64-bit)
Any ?

### LINKED ISSUE #14439: Circular stacking diagram ball problem
#### 
The small balls do not stack and are not outside the big ones

#### 
The spheres are stacked on top of each other, outside of the larger spheres

https://jshare.com.cn/public/E0GmQv
微信截图_20201027215808

DJTechnoo:
Fixed #14439, simulation of `packedBubble` stopping too early.


TorsteinHonsi (approved):
Thanks!

PR #14967, opened 2021-01-21
Suggestion to fix #14966

### LINKED ISSUE #14966: Accessibility module included & point y value null leads to errors in browser console
#### Expected behaviour
No errors in console. 

#### Actual behaviour
In case the accessibility module is included, there is a error (or multiple errors) in the console when selecting a point whose y value is null programmatically:
Error: <rect> attribute y: Expected length, "NaN".

In addition, the point that is added in the DOM has a y value of NaN.

image

https://jsfiddle.net/L8kaxdpw/20/

Rerpo steps:

1. Open attached jsfiddle
2. Open browser dev tools
3. Click on the button and check the console afterwards (not the console in jsfiddle but the actual browser console)

#### Product version
I was able to reproduce it in Highcharts 8.1.2 and 8.2.2

kemusag:
This would fix the browser console outputs described in #14966

### DIFF
--- a/js/Core/Series/Point.js
+++ b/js/Core/Series/Point.js
@@ -1136,7 +1136,7 @@ var Point = /** @class */ (function () {
             markerAttribs = series.markerAttribs(point, state);
         }
         // Apply hover styles to the existing point
-        if (point.graphic) {
+        if (point.graphic && !point.hasDummyGraphic) {
             if (previousState) {
                 point.graphic.removeClass('highcharts-point-' + previousState);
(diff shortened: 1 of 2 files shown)


PaulDalek:
Hi @kemusag

Thank you for proposing this fix and creating a PR.

TorsteinHonsi:
Thanks! I've tagged the issue as Backlog, and this PR is linked. We will create a new PR with a test and that runs through the CI.

PR #14975, opened 2021-01-22
bugfix/14879-currentPriceIndicator-label-disappearing

### LINKED ISSUE #14879: Highstocks: Current Price Indicator Label disappearing sometimes
#### Expected behaviour
When I enable the current price indicator, I expect to see the label showing the current price at all times.

#### Actual behaviour
The label showing the current price disappears and reappears when I change the view in the navigator. It also disappears and reappears when changing the view using the range selector.

http://jsfiddle.net/rogerwa/5bvpgdfr/9/
1. Enable current price indicators, notice that the current price indicator label works fine.
2. Select 3m in the range selector. Notice that the current price indicator label disappears, depending on the window size.
3. Move the time window using the navigator below, and notice that the current price indicator label disappears and reappears.

#### Product version
Highstock v8.2.2

#### Affected browser(s)
Chrome v87
Firefox v84

karolkolodziej:
Fixed #14879, `currentPriceIndicator` label was not always visible when navigating the data set.

~Fixed #14879, currentPriceIndicator label was not visible.~


ts/Core/Chart/StockChart.ts
```diff
@@ -725,17 +725,20 @@ addEvent(Axis, 'afterDrawCrosshair', function (
         point[this.isXAxis ? 'x' : 'y'] :
         this.toValue(horiz ? e.chartX : e.chartY);
 
+    // Crosshair should be rendered within Axis range (#7219)
+    // Also, the point of currentPriceIndicator
+    // should be inside the plot area, #14879.
+    const isInside = point ? point.series.isPointInside(point) :
+        (value as any) > (min as any) || (value as any) < (max as any);
```
raf18seb, line 732: You changed `<` and `>` operators to be opposite - be careful with that :)

raf18seb (changes requested):

ts/Core/Chart/StockChart.ts
karolkolodziej: Indeed, I changed that to make it more clear and readable. 
I also changed the visibility values in the ternary operator in order to preserve the logic. 
I checked these changes once again and everything works as expected.

raf18seb: Sorry for not explaining enough.
The old condition `!(A+B)` is not the same as `!A+!B`. I should be `!A*!B`.

Your condition always returns `true`. Let's take `value = 15`, `min = 0` and `max = 10`
`15 > 1 || 15 < 10` <==> `true || false` <==> `true` but, obviously, 15 is not between 0 and 10.

Also, when it comes to negation, remember that `!<` is `>=`, not `>`

karolkolodziej: You are right! 
Thanks for pointing that out! ;)

raf18seb: No problem.
Have you checked whether it should be `>` or `>=`?

karolkolodziej: I didn't find a single test case in which `>` will fail.

raf18seb (approved):
Thanks!

pawelfus (approved):
Looks good, thanks!

pawellysy (approved):
Looks Good!

TorsteinHonsi:
All good, pending CI

TorsteinHonsi (approved):
Thanks! I added some typesafing. Awaiting CI before merge.

PR #14976, opened 2021-01-22
bugfix/14973-patternfill-base-pushState

### LINKED ISSUE #14973: Pattern fill fails to render when <base> tag is present and window.location is manipulated via history.pushState
#### Expected behaviour
Should render as it does in the latest release: http://jsfiddle.net/goransle/kdpswnb6/

#### Actual behaviour
Pattern fills fails to render.

http://jsfiddle.net/goransle/mr4jqa3e/

#### Product version
Current master

#### Affected browser(s)
Latest Chrome + Firefox

khlieng:
Fixed #14973, pattern fill failed to render when `<base>` tag was present and `window.location` was manipulated via `history.pushState`.

The pattern disappearing on `redraw` gets fixed by #14874.

Threw in some unrelated optimizations in `getChartPosition`.

### DIFF
--- a/js/Core/Chart/Chart.js
+++ b/js/Core/Chart/Chart.js
@@ -1847,6 +1847,7 @@ var Chart = /** @class */ (function () {
             }
         });
         chart.render();
+        chart.pointer.getChartPosition(); // #14973
         // Fire the load event if there are no external images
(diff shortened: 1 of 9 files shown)


TorsteinHonsi (approved):
Thanks!

PR #14981, opened 2021-01-23
tools/increase-comparison-time

goransle:
Fixed `visual_comparison_with_master` failing on the last step. It seems our visual tests are finishing faster, which caused the script that waited for the background job to time out. Removed the waiting step by combining the two gulp tasks.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -244,7 +244,7 @@ jobs:
       - run:
           name: Wait for file size comparison and comment on PR
           command: while [ ! -f ./tmp/filesizes/comparison.md ]; do sleep 1; done && npx gulp pr-comment-sizes --pr ${CIRCLE_PULL_REQUEST##*/} --user circleci-mu
-          no_output_timeout: 2m #wait for max 2 minutes
+          no_output_timeout: 5m #wait for max 2 minutes
 
   nightly_visual_diff:
     <<: *defaults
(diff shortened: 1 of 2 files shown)


bre1470:
Wouldn't it be better to produce some output than increasing the timeout?

goransle:
I don't think we would want the step to wait forever, in case the file it's waiting for is not made. But, we should probably look into doing this in a separate step by storing artifacts.

bre1470:
I do not understand, why this would lead to a forever wait.
At the moment the issue is only, that CI has to assume after 2 minutes without console output, that the task is probably dead.
So instead of waiting even 5 minutes for dead tasks, you could make a verbose task output that indicates vitality.
With that you can even reduce the timeout to one minute.

goransle:
### File size comparison
| | master | candidate | difference |
|-------------|-------------:|-------------:|-------------:|
| highcharts | 281.9 kB | 276.1 kB | -5872 B |
| highcharts, gzipped | 95.2 kB | 94.1 kB | -1158 B|
| highstock | 372.1 kB | 359.3 kB | -13136 B |
| highstock, gzipped | 124.6 kB | 121.6 kB | -3045 B|
| highmaps | 339.4 kB | 331.6 kB | -7998 B |
| highmaps, gzipped | 112.0 kB | 110.6 kB | -1462 B|
| highcharts-gantt | 387.5 kB | 372.7 kB | -15143 B |
| highcharts-gantt, gzipped | 129.5 kB | 125.7 kB | -3942 B|

TorsteinHonsi (approved):
Thanks!

PR #14986, opened 2021-01-25
bugfix/14980-toggle-stacking-centerInCategory

### LINKED ISSUE #14980: Column chart bug: switching to / from stacked with centerInCategory: true
#### Expected behaviour
Toggling stacked option on series should switch series between two states

#### Actual behaviour
Toggling stacked option the second time makes stacks overlap

https://jsfiddle.net/nulle/x7pLeatr/
1. Loaded state is a stacked chart with **two stacks** (two series in each stack), and **centerInCategory: true** (let's call this state1).
2. Click "Toggle Stacked" button once, and stacked option is removed, chart now has 4 columns in each category (state2). 
3. Click "Toggle Stacked" button once more. Expected: state1 (same as on initial load). Actual result: stacked bars are not next to eachother, but overlapping (clearly visible when tooltipping).

Note that with either "centerInCategory" turned off, or just one stack, everything works as expected.
Possibly connected to https://github.com/highcharts/highcharts/issues/14921 ? Though, just one y axis here.

#### Product version
Highcharts 8.2.2

#### Affected browser(s)
Probably all (using Chrome 87.0.4280.141 on MacOS)

khlieng:
Fixed #14980, toggling stacking with `centerInCategory` enabled did not work.


TorsteinHonsi (changes requested):
There's still a bug if we start out with `isStacked = false` : https://jsfiddle.net/highcharts/8cvk9q4x/

Everything is good until we click "Toggle" the second time, then the columns fly off to the left only to return on the third click.

Also, instead of doing the `skip` inside that loop, I think we should dig deeper. In the [OP demo](https://jsfiddle.net/nulle/x7pLeatr/), why is it that there is no `column,group` stack initially, but it is there when we're updating? It seems like there is something that should be destroyed when we toggle stacking.

khlieng:
Its not there initially because it doesnt get created if real stacking is enabled, but all the stacks should probably be destroyed somehow when `stacking` or `centerInCategory` changes.

TorsteinHonsi:
> all the stacks should probably be destroyed somehow when stacking or centerInCategory changes.

Yes, can you look into that? So that when we reach `adjustForMissingColumns` the state is the same regardless of what it was before the update.

TorsteinHonsi (approved):

Issue #14010, opened 2020-08-06
New demo: Stock chart with all indicators


pawelfus:
Current demos: https://www.highcharts.com/demo/stock

We have three demos of indicators: 
- StockTools (general demo)
- MACD + Pivot points
- SMA + Volume by Price

Idea is to add a new demo (or replace one of the existing) where users can get overview of all indicators, not only 4 out of 40.

Additionally, it would be nice to improve demos titles
- `MACD pivot points` -> `MACD and Pivot points`
- `SMA volume by price` -> `SMA and Volume by price`

And change title of the paragraph:
- `Flags and Indicators` -> `Flags and Technical indicators`

PR #14987, opened 2021-01-25
enhancement/14010-new-demo-technical-indicators

karolkolodziej:
Feature #14010, added the demo with all technical indicators.

Edited demo: http://jsfiddle.net/BlackLabel/czpj5h74/

### DIFF
--- a/samples/stock/demo/all-indicators/demo.css
+++ b/samples/stock/demo/all-indicators/demo.css
@@ -1,17 +1,14 @@
-.main {
-	position: absolute;
-	min-width: 800px;
-}
-
 .selectors {
 	height: 20px;
 	font-size: 20px;
-	display: flex;
-}
-
+  }
+  
 .left {
-	flex: 1;
+	position: absolute;
+	left: 0%;
+	top: 1%;
 	margin-left: 20px;
+	width: 50%;
 }
 
 .left__labels {
@@ -23,8 +20,11 @@
 }
 
 .right {
-	flex: 1;
-	margin-left:20px;
+	position: absolute;
+	left: 50%;
+	top: 1%;
+	margin-left: 20px;
+	width: 50%;
 }
 
 .right__labels {
--- a/samples/stock/demo/all-indicators/demo.css
+++ b/samples/stock/demo/all-indicators/demo.css
@@ -1,0 +1,42 @@
+.main {
+	position: absolute;
+	min-width: 800px;
+}
+
+.selectors {
+	height: 20px;
+	font-size: 20px;
+	display: flex;
+}
+
+.left {
+	flex: 1;
+	margin-left: 20px;
+}
+
+.left__labels {
+	display: block;
+}
+
+.left__select {
+	width: 90%;
+}
+
+.right {
+	flex: 1;
+	margin-left:20px;
+}
+
+.right__labels {
+	display: block;
+}
+
+.right__select {
+	width: 90%;
+}
+
+#container {
+	position: relative;
+	top: 30px;
+}
+  
--- a/samples/stock/demo/all-indicators/demo.css
+++ b/samples/stock/demo/all-indicators/demo.css
@@ -1,35 +1,24 @@
 .selectors-container {
-	font-size: 1.2em;
+	background: #f2f2f2;
+	margin-bottom: 1rem;
+	font-size: 0;
 }
 
-.lhs-col {
-	width: 48%;
+.selectors-container .col {
+	font-size: 1.2rem;
+	width: calc(50% - 1em);
+	padding: 0.5em;
 	display: inline-block;
 }
 
-.left-select {
+.selectors-container select {
 	width: 100%;
-}
-
-.rhs-col {
-	width: 48%;
-	margin-left: 2%;
-	display: inline-block;
-}
-
-.right-select {
-	width: 100%;
+	font-size: 16px; /* prevent page zoom in iOS */
 }
 
 @media (max-width: 768px) {
-	.lhs-col,
-	.rhs-col {
+	.selectors-container .col {
 		display: block;
-		width: 100%;
-	}
-
-	.rhs-col {
-		margin-left: 0px;
-		margin-top: 15px;
+		width: calc(100% - 1em);
 	}
 }
--- a/samples/stock/demo/all-indicators/demo.css
+++ b/samples/stock/demo/all-indicators/demo.css
@@ -4,39 +4,27 @@
   }
   
 .left {
-	position: absolute;
-	left: 0%;
-	top: 1%;
-	margin-left: 20px;
+	position: relative;
+	left: 2%;
 	width: 50%;
 }
-
-.left__labels {
-	display: block;
-}
-
+  
 .left__select {
 	width: 90%;
 }
-
+  
 .right {
-	position: absolute;
-	left: 50%;
-	top: 1%;
-	margin-left: 20px;
+	position: relative;
 	width: 50%;
+	left: 52%;
+	transform: translate(0, -100%);
 }
-
-.right__labels {
-	display: block;
-}
-
+  
 .right__select {
 	width: 90%;
 }
-
+  
 #container {
 	position: relative;
-	top: 30px;
+	top: 35px;
 }
-  
--- a/samples/stock/demo/all-indicators/demo.css
+++ b/samples/stock/demo/all-indicators/demo.css
@@ -1,30 +1,35 @@
-.selectors {
-	height: 20px;
-	font-size: 20px;
-  }
-  
-.left {
-	position: relative;
-	left: 2%;
-	width: 50%;
+.selectors-container {
+	font-size: 1.2em;
 }
-  
-.left__select {
-	width: 90%;
+
+.lhs-col {
+	width: 48%;
+	display: inline-block;
 }
-  
-.right {
-	position: relative;
-	width: 50%;
-	left: 52%;
-	transform: translate(0, -100%);
+
+.left-select {
+	width: 100%;
 }
-  
-.right__select {
-	width: 90%;
+
+.rhs-col {
+	width: 48%;
+	margin-left: 2%;
+	display: inline-block;
 }
-  
-#container {
-	position: relative;
-	top: 35px;
+
+.right-select {
+	width: 100%;
+}
+
+@media (max-width: 768px) {
+	.lhs-col,
+	.rhs-col {
+		display: block;
+		width: 100%;
+	}
+
+	.rhs-col {
+		margin-left: 0px;
+		margin-top: 15px;
+	}
 }
--- a/samples/stock/demo/all-indicators/demo.details
+++ b/samples/stock/demo/all-indicators/demo.details
@@ -1,7 +1,7 @@
 ---
 name: All technical indicators
 authors:
-  - Karol Kołodziej
+  - Karol Kolodziej
 js_wrap: b
 tags:
   - Highcharts Stock demo
--- a/samples/stock/demo/all-indicators/demo.details
+++ b/samples/stock/demo/all-indicators/demo.details
@@ -1,10 +1,10 @@
 ---
-name: MACD pivot points
+name: All technical indicators
 authors:
   - Karol Kołodziej
 js_wrap: b
 tags:
   - Highcharts Stock demo
 categories:
-  - Flags and Indicators
+  - Flags and Technical indicators
 ...
--- a/samples/stock/demo/all-indicators/demo.details
+++ b/samples/stock/demo/all-indicators/demo.details
@@ -1,0 +1,10 @@
+---
+name: MACD pivot points
+authors:
+  - Karol Kołodziej
+js_wrap: b
+tags:
+  - Highcharts Stock demo
+categories:
+  - Flags and Indicators
+...
--- a/samples/stock/demo/all-indicators/demo.html
+++ b/samples/stock/demo/all-indicators/demo.html
@@ -3,11 +3,11 @@
 <script src="https://code.highcharts.com/stock/modules/exporting.js"></script>
 <script src="https://code.highcharts.com/stock/indicators/indicators-all.js"></script>
 
-<div class='main'>
-  <div class='selectors'>
-    <div class="left">
+<div class="main-wrapper">
+  <div class="selectors-container">
+    <div class="lhs-col">
       <label for="overlays">Overlays:</label>
-      <select class='left__select' id="overlays">
+      <select class='left-select' id="overlays">
         <option value="abands">Acceleration Bands</option>
         <option value="bb">Bollinger Bands</option>
         <option value="dema">DEMA (Double Exponential Moving Average)</option>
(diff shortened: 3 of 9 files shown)


samples/stock/demo/all-indicators/demo.html
```diff
@@ -0,0 +1,65 @@
...
+<script src="https://code.highcharts.com/stock/modules/data.js"></script>
+<script src="https://code.highcharts.com/stock/modules/exporting.js"></script>
+<script src="https://code.highcharts.com/stock/indicators/indicators-all.js"></script>
+
+<div class='main'>
+  <div class='selectors'>
+    <div class="left">
+      <label class='left__labels' for="overlays">Overlays:</label>
+      <select class='left__select' id="overlays">
+        <option value="none">None</option>
```
raf18seb, line 11: I haven't seen many HTML tags in our official demos. We could consider using Highcharts tools e.g. `H.createElement` and then we could position 'selects' using Highcharts coordinates (without CSS). @pawelfus what do you think?

samples/stock/demo/all-indicators/demo.js
```diff
@@ -0,0 +1,193 @@
...
+        }, {
+            type: 'column',
+            id: 'volume',
+            name: 'Volume',
+            data: volume,
+            yAxis: 1
+        }]
+    });
+
+    document.getElementById("overlays").addEventListener("change", function (e) {
```
raf18seb, line 64: I would put this code either into the chart's callback or load event.

samples/stock/demo/all-indicators/demo.js
```diff
@@ -0,0 +1,193 @@
...
+        const seriesToAdd = e.target.value;
+
+        // Destroy previously existing series.
+        if (chart.series[2] && chart.series[2].yAxis.options.index === 0) {
+            chart.series[2].destroy(false);
+        }
+        if (chart.series[3] && chart.series[3].yAxis.options.index === 0) {
+            chart.series[3].destroy(false);
+        }
```
raf18seb, line 73: What is the purpose of this if statement?

samples/stock/demo/all-indicators/demo.js
```diff
@@ -0,0 +1,193 @@
...
+            }
+
+            if (seriesToAdd !== 'aroon') {
+                chart.addSeries({
+                    type: seriesToAdd,
+                    linkedTo: 'aapl',
+                    yAxis: 3,
+                    showInLegend: true
+                }, false);
+            }
```
raf18seb, line 189: We could put it into one `if-else` statement.

raf18seb (changes requested):

samples/stock/demo/all-indicators/demo.js
```diff
@@ -0,0 +1,193 @@
+Highcharts.getJSON('https://demo-live-data.highcharts.com/aapl-ohlcv.json', function (data) {
+    // split the data set into ohlc and volume
+    var ohlc = [],
+        volume = [],
+        dataLength = data.length,
+
+        i = 0;
+
+    for (i; i < dataLength; i += 1) {
```
raf18seb, line 9: We can shorten this to one line.

raf18seb (changes requested):

samples/stock/demo/all-indicators/demo.html
pawelfus: I think simple HTML in the demo is fine. For example Periodic table: http://utils.highcharts.local/samples/#view/highcharts/series-heatmap/periodic-table

samples/stock/demo/all-indicators/demo.js
```diff
@@ -0,0 +1,193 @@
...
+
+    // create the chart
+    const chart = Highcharts.stockChart('container', {
+        chart: {
+            height: 600
+        },
+        title: {
+            text: 'AAPL Historical'
+        },
+        subTitle: {
```
pawelfus, line 32: `subtitle`

samples/stock/demo/all-indicators/demo.js
```diff
@@ -0,0 +1,193 @@
...
+    });
+
+    document.getElementById("oscilators").addEventListener("change", function (e) {
+        const seriesToAdd = e.target.value;
+
+        // Add third axis if needed.
+        if (chart.yAxis.length === 3) {
+            chart.yAxis[0].update({
+                height: '55%'
+            });
```
pawelfus, line 131: When using multiple updates, set `redraw=false` (the same applies to the `update()` and `addAxis()` below)

samples/stock/demo/all-indicators/demo.js
```diff
@@ -0,0 +1,193 @@
...
+                        styles: {
+                            lineColor: 'red'
+                        }
+                    },
+                    senkouSpan: {
+                        color: 'rgba(0, 255, 0, 0.3)',
+                        styles: {
+                            fill: 'rgba(0, 0, 255, 0.1)'
+                        }
+                    }
```
pawelfus, line 111: Do we need to define all styles for lines in IKH?

samples/stock/demo/all-indicators/demo.js
```diff
@@ -0,0 +1,193 @@
...
+            text: 'All indicators'
+        },
+        legend: {
+            enabled: true
+        },
+        rangeSelector: {
+            selected: 2
+        },
+        tooltip: {
+            split: true
```
pawelfus, line 42: This is the default, isn't it?

samples/stock/demo/all-indicators/demo.css
```diff
@@ -0,0 +1,42 @@
+.main {
+	position: absolute;
+	min-width: 800px;
+}
+
+.selectors {
+	height: 20px;
+	font-size: 20px;
+	display: flex;
```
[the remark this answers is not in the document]
pawelfus: + `flex` is tricky when comes to auto-resizing the chart

samples/stock/demo/all-indicators/demo.js
karolkolodziej: No, we don't have to, actually most of them (especially the fill) are set by default.

samples/stock/demo/all-indicators/demo.js
karolkolodziej: User can add first an oscillator (which will be series with the index 2), then can add an overlay (which will be with index 3). The only thing (as far as I can see) which you can use to distinguish them is the yAxis index.
The main purpose for that statements is to detect which series is assigned to which axis and destroy the correct one.
Is there any room for improvement?

samples/stock/demo/all-indicators/demo.css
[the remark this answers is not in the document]
karolkolodziej: I removed the `flex` property and I used some different style attributes.

raf18seb (approved):
Looks good, thanks!

samples/stock/demo/all-indicators/demo.js
```diff
@@ -0,0 +1,147 @@
...
+            } else {
+                if (seriesToAdd === 'aroon') {
+                    chart.addSeries({
+                        yAxis: 3,
+                        type: 'aroon',
+                        linkedTo: 'aapl',
+                        color: 'green',
+                        lineWidth: 1,
+                        showInLegend: true,
+                        aroonDown: {
```
pawelfus, line 129: Can we simplify it? I mean, let's treat `aroon` as any other tech. indicator, without defining styles (just use defaults)

samples/stock/demo/all-indicators/demo.js
```diff
@@ -0,0 +1,147 @@
...
+                chart.addSeries({
+                    type: seriesToAdd,
+                    linkedTo: 'aapl',
+                    showInLegend: true
+                }, false);
+            }
+            chart.redraw();
+        });
+
+        document.getElementById("oscilators").addEventListener("change", function (e) {
```
pawelfus, line 79: Sorry for missing this earlier, just a typo: "oscillators"

pawelfus (commented):
Can we change the demo, I mean the way it works? For example:

1. Let's init chart and select by default one overlay and one oscillator (with its own yAxis)
2. Remove `None` option from dropdowns. That simplifies a lot demo (managing yAxis heights, indexes, etc)
3. Add `id` to overlay and oscillator series. So when changing indicator, we simply use `let series = chart.get('series-id'); if (series) { series.remove(false) }` then add new series

karolkolodziej:
I've changed that logic and simplified the demo as you mentioned.

The demo looked strange in utils because of the CSS, I also fixed that.
Updated demo: http://jsfiddle.net/BlackLabel/0qnjfxtv/

sebastianbochan (changes requested):
- change class names
- replace px with em (font-size)
- remove relative / top and use margins / display-inline etc.

samples/stock/demo/all-indicators/demo.js
```diff
@@ -0,0 +1,107 @@
...
+        ]);
+
+        volume.push([
+            data[i][0], // the date
+            data[i][5] // the volume
+        ]);
+    }
+
+    // create the chart
+    const chart = Highcharts.stockChart('container', {
```
raf18seb, line 23: `const` will not work in the old browsers. We can remove it (see the comment below).

samples/stock/demo/all-indicators/demo.js
```diff
@@ -0,0 +1,107 @@
...
+            showInLegend: true,
+            yAxis: 0
+        }, {
+            type: 'macd',
+            id: 'oscillator',
+            linkedTo: 'aapl',
+            showInLegend: true,
+            yAxis: 2
+        }]
+    }, function () {
```
raf18seb, line 72: Instead of declaring a `const chart` variable, we can use `chart` as function's argument like
```js
}, function (chart) {
```

samples/stock/demo/all-indicators/demo.js
```diff
@@ -0,0 +1,107 @@
...
+        document.getElementById("overlays").addEventListener("change", function (e) {
+            const series = chart.get('overlay');
```
raf18seb, line 74: Another `const`. Please, find all of them. The same with double quotes.

raf18seb, line 74: Mixing single and double quotes. Prefer single `"` -> `'`

samples/stock/demo/all-indicators/demo.js
```diff
@@ -0,0 +1,107 @@
...
+
+            if (series) {
+                series.remove(false);
+                chart.addSeries({
+                    type: e.target.value,
+                    linkedTo: 'aapl',
+                    id: 'overlay',
+                    yAxis: 0,
```
raf18seb, line 82: Is it necessary? `yAxis: 0` is by default. Or maybe this indicator doesn't stick to the first axis by default?

samples/stock/demo/all-indicators/demo.js
```diff
@@ -0,0 +1,107 @@
...
+                    showInLegend: true
```
raf18seb, line 83: There are many `showInLegend: true`. If all series should be shown in the legend, you could define it in `plotOptions.series`.

samples/stock/demo/all-indicators/demo.details
```diff
@@ -0,0 +1,10 @@
+---
+name: All technical indicators
+authors:
+  - Karol Kołodziej
```
raf18seb, line 4: By the occasion - don't you want to replace `ł` with `l` as we discussed?

samples/stock/demo/all-indicators/demo.html
```diff
@@ -0,0 +1,64 @@
+<script src="https://code.highcharts.com/stock/highstock.js"></script>
+<script src="https://code.highcharts.com/stock/modules/data.js"></script>
+<script src="https://code.highcharts.com/stock/modules/exporting.js"></script>
+<script src="https://code.highcharts.com/stock/indicators/indicators-all.js"></script>
+
+<div class="main-wrapper">
+  <div class="selectors-container">
```
raf18seb, line 7: You can leave it as it is, but do we need this `<div class="selectors-container">`? I don't see any good reasons (maybe there is one) why this
```html
<div class="a">
    <div class="b">
        <p></p>
        <p></p>
        <p></p>
    </div>
</div>
```
cannot be replaced with (with corrected CSS styles and classes maybe)
```html
<div class="a b">
    <p></p>
    <p></p>
    <p></p>
</div>
```

Just a suggestion for the future.

samples/stock/demo/all-indicators/demo.html
```diff
@@ -0,0 +1,64 @@
...
+    <div class="lhs-col">
+      <label for="overlays">Overlays:</label>
+      <select class='left-select' id="overlays">
```
raf18seb, line 10: In this case, a double quote is preferable `"`.

raf18seb (changes requested):

samples/stock/demo/all-indicators/demo.html
karolkolodziej: `main-wrapper` class is to... well.. wrap the whole selectors and the chart together as it might be a part of some bigger design/site

raf18seb (approved):
Thanks!

pawelfus (approved):
Thanks!

samples/stock/demo/all-indicators/demo.html
```diff
@@ -0,0 +1,64 @@
...
+        <option value="vwap">WMA (Weighted Moving Average)</option>
+        <option value="wma">VWAP (Volume Weighted Average Price)</option>
+        <option value="zigzag">Zig Zag</option>
+      </select>
+    </div>
+
+    <div class="rhs-col">
+      <label for="oscillators">Oscillators:</label>
+      <select class="right-select" id="oscillators">
+        <option value="none">None</option>
```
pawelfus, line 35: Accepted, and then realized: There is no `none` technical indicator :)

samples/stock/demo/all-indicators/demo.html
```diff
@@ -0,0 +1,64 @@
...
+        <option value="apo">Absolute price indicator</option>
+        <option value="ad">A/D (Accumulation/Distribution)</option>
+        <option value="aroon">Aroon</option>
+        <option value="aroonoscillator">Aroon oscilator</option>
```
pawelfus, line 39: Oscillator :)

samples/stock/demo/all-indicators/demo.html
```diff
@@ -0,0 +1,64 @@
...
+        <option value="atr">ATR (Average True Range)</option>
+        <option value="ao">Awesome oscilator</option>
```
pawelfus, line 41: Oscillator

pawelfus (changes requested):
minor polishing

pawelfus (approved):
Thanks!

pawellysy (approved):
Looks good!

TorsteinHonsi (approved):
Thanks, nice demo! I did some final CSS polishing, now merging.

PR #14988, opened 2021-01-25
bugfix/13473-scrollbar-threshold

### LINKED ISSUE #13473: disappearing column/bar when scrolling beyond point value

khlieng:
Fixed #13473, it was not possible to scroll back down to the `threshold` after scrolling up with `yAxis.scrollbar` enabled and no explicit `yAxis.min` set.

### DIFF
--- a/ts/Core/Axis/ScrollbarAxis.ts
+++ b/ts/Core/Axis/ScrollbarAxis.ts
@@ -67,13 +67,15 @@ class ScrollbarAxis {
                     Math.min(
                         axisMin,
                         axis.min as any,
-                        Math.min(pick(axis.threshold, Infinity), axis.dataMin)
+                        axis.dataMin,
+                        pick(axis.threshold, Infinity)
                     ) : axisMin,
                 scrollMax: defined(axis.dataMax) ?
                     Math.max(
                         axisMax,
                         axis.max as any,
-                        Math.max(pick(axis.threshold, -Infinity), axis.dataMax)
+                        axis.dataMax,
+                        pick(axis.threshold, -Infinity)
                     ) : axisMax
             };
         };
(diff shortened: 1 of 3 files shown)


ts/Core/Axis/ScrollbarAxis.ts
```diff
@@ -87,13 +87,15 @@ class ScrollbarAxis {
                             Math.min(
                                 axisMin,
                                 axis.min as any,
-                                axis.dataMin as any
+                                (axis.dataMin as any) > (axis.threshold as any) ?
+                                    axis.threshold : (axis.dataMin as any)
```
TorsteinHonsi, line 91: Wouldn't it be more compact and readable to write `Math.min(axis.threshold, axis.dataMin)` in this expression and the three others?

TorsteinHonsi (changes requested):

ts/Core/Axis/ScrollbarAxis.ts
khlieng: `Math.min` return `NaN` if any of the values are undefined.

khlieng: Maybe something like `Math.min(pick(axis.threshold, Number.MAX_VALUE), axis.dataMin)` could work though.

khlieng: Or not, tests are failing because `Math.min` turns `null` into `0` :/

TorsteinHonsi (approved):
Thanks!

PR #14993, opened 2021-01-26
bugfix/13170-boost-scatter-cull-y

### LINKED ISSUE #13170: Markers outside of zoomed area are displayed at the bottom of the visible plot area
#### Expected behaviour
These markers should not be displayed.

#### Actual behaviour
Using Highstock for scrolling both x, y directions and Boost module to plot a large number of data points, these markers, which are below the current zoomed area, are displayed at the bottom of the visible plot area.  This is more noticeable when scrolling upward because markers look 'stuck' at the bottom instead of being scrolled out.

https://jsfiddle.net/ktakahashimtb/5ca6qg8w/

#### Product version
Highstock v8.0.4 (2020-03-10)  

#### Affected browser(s)
Chrome, Edge, Firefox, Safari

khlieng:
Fixed #13170, some of the points below a zoomed-in boosted scatter chart were not culled properly and displayed along the bottom of the plot area.

### DIFF
--- a/ts/Extensions/Boost/WGLRenderer.ts
+++ b/ts/Extensions/Boost/WGLRenderer.ts
@@ -748,11 +748,11 @@ function GLRenderer(
                 continue;
             }
 
-            if (nx && nx >= xMin && nx <= xMax && series.type !== 'scatter') {
+            if (nx && nx >= xMin && nx <= xMax && !series.is('scatter')) {
                 nextInside = true;
             }
 
-            if (px && px >= xMin && px <= xMax && series.type !== 'scatter') {
+            if (px && px >= xMin && px <= xMax && !series.is('scatter')) {
                 prevInside = true;
             }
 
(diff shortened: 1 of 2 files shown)


ts/Extensions/Boost/WGLRenderer.ts
```diff
@@ -748,11 +748,11 @@ function GLRenderer(
                 continue;
             }
 
-            if (nx && nx >= xMin && nx <= xMax) {
+            if (nx && nx >= xMin && nx <= xMax && series.type !== 'scatter') {
```
TorsteinHonsi, line 751: Hard checking for `series.type` should not be used because it will fail for inherited children of scatter. Instead, use `!series.is('scatter')`, which will also be true for inherited children. Alternatively, in this case we may actually be looking for `&& series.is('line')`

TorsteinHonsi (changes requested):

ts/Extensions/Boost/WGLRenderer.ts
khlieng: But scatter inherits from line right? I'll do `!series.is('scatter')`

TorsteinHonsi (approved):

PR #14999, opened 2021-01-27
bugfix/12100-mapNavigation-indicators-menu-scrollable

### LINKED ISSUE #12100: StockTools indicators menu is not scrollable with mapNavigation.enabled: true
#### Expected behaviour

#### Actual behaviour

karolkolodziej:

To do:
- [x] test


samples/stock/stocktools/scrollable-popup/demo.html
```diff
@@ -0,0 +1,14 @@
...
+<script src="https://code.highcharts.com/stock/highstock.js"></script>
+<script src="https://code.highcharts.com/stock/modules/data.js"></script>
+<script src="https://code.highcharts.com/maps/modules/map.js"></script>
+
+<script src="https://code.highcharts.com/stock/indicators/indicators-all.js"></script>
+
+<script src="https://code.highcharts.com/modules/annotations-advanced.js"></script>
+
+<script src="https://code.highcharts.com/modules/stock-tools.js"></script>
+
```
raf18seb, line 13: Do we need these spaces between `<script>` tags?

samples/stock/stocktools/scrollable-popup/demo.js
```diff
@@ -0,0 +1,31 @@
...
+        i = 0;
+    for (i; i < dataLength; i += 1) {
+        ohlc.push([
+            data[i][0], // the date
+            data[i][1], // open
+            data[i][2], // high
+            data[i][3], // low
+            data[i][4] // close
+        ]);
+    }
```
raf18seb, line 13: I think we could use simple line series with few points - no need to load `data.js` module, fetch data from server and iterate through the whole array. We could also remove 3 lines in the series' config (type, id, name).

raf18seb (changes requested):

raf18seb (approved):
Thanks!

samples/stock/stocktools/scrollable-popup/demo.js
```diff
@@ -0,0 +1,15 @@
...
+    },
+    stockTools: {
+        gui: {
+            enabled: true,
+            buttons: ['indicators']
+        }
+    },
+    series: [{
+        type: 'line',
+        data: Array.from(Array(50)).map(() => Math.random() * 10)
```
pawellysy, line 13: isn't it better to generate the random array this way? 
`Array.from({length: 50}, () => Math.random() * 10)`

pawellysy (changes requested):

js/Extensions/Annotations/Popup.js
```diff
@@ -726,6 +737,7 @@ addEvent(NavigationBindings, 'showPopup', function (config) {
             (this.chart.options.stockTools &&
                 this.chart.options.stockTools.gui.iconsURL) ||
             'https://code.highcharts.com/@product.version@/gfx/stock-icons/'));
+        this.popup.chart = this.chart;
```
pawelfus, line 740: Instead of assigning it here, should we wait for #15015 to be merged? `popup.chart` is already implemented there. 

PS: It seems #15015 is missing `popup.chart` definition in TS Popup class.

karolkolodziej: I will wait till the merge and then I'm going to change that :)

samples/stock/stocktools/scrollable-popup/demo.js
karolkolodziej: it does look slightly better thanks!

pawellysy (approved):
Looks good!

TorsteinHonsi:
Sorry for the mess, I approved and merged this, then regretted and reverted it. See my [comment in the issue](https://github.com/highcharts/highcharts/issues/12100#issuecomment-779811670).

PR #15000, opened 2021-01-27
bugfix/12548-scatter3d-null-z

### LINKED ISSUE #12548: 3dscatter points are not displayed when the first one has z=null
Setup: 3dscatter chart, with ``        data: [
			[50,50, null],
            [15, 35, 50],
            [10, 20, 30],
        ]``
#### Expected behaviour
All three points are displayed (null is counted as 0)
OR
The first point is not displayed, but other points are
#### Actual behaviour
No points are displayed

https://jsfiddle.net/BlackLabel/kgv6rd1a/

The problem occurs only in a situation when the first point of data have z = null. Surprisingly if that point is not the first point in data, all three points are displayed correctly. In any other situation, the point, where one of the coordinates(x or y) is null, the point is not displayed but it doesn't affect the behavior of other points.

#### Related forum post
https://www.highcharts.com/forum/viewtopic.php?f=9&t=42737&start=15

#### Product version
Highcharts JS v7.2.1 (2019-10-31)

#### Affected browser(s)
All

khlieng:
Fixed #12548, 3d scatter chart rendered no points when the first point had `z` set to `null`.

### DIFF
--- a/js/Series/Scatter3D/Scatter3DPoint.js
+++ b/js/Series/Scatter3D/Scatter3DPoint.js
@@ -24,6 +24,8 @@ var __extends = (this && this.__extends) || (function () {
     };
 })();
 import ScatterSeries from '../Scatter/ScatterSeries.js';
+import U from '../../Core/Utilities.js';
+var defined = U.defined;
 /* *
  *
  *  Class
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (approved):
Thanks!

PR #15003, opened 2021-01-27
bugfix/12517-scrollablePlotArea-resetZoomButton

### LINKED ISSUE #12517: Reset button not showing when using scrollablePlotArea
#### Expected behaviour
The Reset Zoom should show on the top right of the chart after a zoom.

#### Actual behaviour
Regardless of whether I use HighCharts or GanttCharts, if the `scrollablePlotArea.minWidth`value is big enough to cause the right end of the chart to appear off-screen, the Reset Zoom button won't show.

According to [this issue](https://github.com/highcharts/highcharts/issues/10466) (#10466), it was normal for the Reset Zoom Button to stick to the right end of the chart when zooming, even if that end is off-screen. Furthermore, this behaviour seemed to have been fixed https://github.com/highcharts/highcharts/commit/8e46848c4068ca5cb18073802c5da9cd64691244. 

Therefore, I suspect it to be a regression, but maybe I missed something.

From my observations, the `transform=translate(x,y)` seems to be computed based on the `minWidth` value and, as such, the absolute positioning isn't reachable since it isn't affected by the plot area being scrolled.

Gantt demo
https://jsfiddle.net/1sto4fqg/1/

(make sure to edit the `minWidth` value if you happen to have a screen bigger than 3000 pixels, the value I set in this demo.

1. Zoom (xAxis) anywhere by clicking and dragging
2. Scroll horizontally (`Shift + scroll`, on Windows) to reach the right end of the chart
3. The Reset Zoom button isn't there 
(you can see it exists in the DOM tree by searching for the `highcharts-reset-zoom` class. The translate value for the button seems to `transform="translate(2896,112)"`).

Same thing happens in the first issue I mentioned (#10466)

#### Product version
HighCharts and Gantt v7.2.1 

#### Affected browser(s)
Chrome
Any ?

khlieng:
Fixed #12517, `resetZoomButton` did not show when using `scrollablePlotArea`.

### DIFF
--- a/js/Core/Chart/Chart.js
+++ b/js/Core/Chart/Chart.js
@@ -2449,12 +2449,7 @@ var Chart = /** @class */ (function () {
         var chart = this, lang = defaultOptions.lang, btnOptions = chart.options.chart.resetZoomButton, theme = btnOptions.theme, states = theme.states, alignTo = (btnOptions.relativeTo === 'chart' ||
             btnOptions.relativeTo === 'spaceBox' ?
             null :
-            {
-                x: chart.plotLeft,
-                y: chart.plotTop,
-                width: chart.chartWidth - chart.plotLeft - (chart.marginRight || 0),
-                height: chart.chartHeight - chart.plotTop - (chart.marginBottom || 0)
-            });
+            this.scrollablePlotBox || 'plotBox');
         /**
          * @private
          */
--- a/js/Core/Chart/Chart.js
+++ b/js/Core/Chart/Chart.js
@@ -2452,8 +2452,8 @@ var Chart = /** @class */ (function () {
             {
                 x: chart.plotLeft,
                 y: chart.plotTop,
-                width: chart.spacingBox.x + chart.spacingBox.width - chart.plotLeft,
-                height: chart.spacingBox.y + chart.spacingBox.height - chart.plotTop
+                width: chart.chartWidth - chart.plotLeft - (chart.marginRight || 0),
+                height: chart.chartHeight - chart.plotTop - (chart.marginBottom || 0)
             });
         /**
          * @private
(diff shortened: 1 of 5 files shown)


js/Core/Chart/Chart.js
```diff
@@ -2449,7 +2449,12 @@ var Chart = /** @class */ (function () {
         var chart = this, lang = defaultOptions.lang, btnOptions = chart.options.chart.resetZoomButton, theme = btnOptions.theme, states = theme.states, alignTo = (btnOptions.relativeTo === 'chart' ||
             btnOptions.relativeTo === 'spaceBox' ?
             null :
-            'plotBox');
+            {
+                x: chart.plotLeft,
+                y: chart.plotTop,
+                width: chart.chartWidth - chart.plotLeft - (chart.marginRight || 0),
+                height: chart.chartHeight - chart.plotTop - (chart.marginBottom || 0)
+            });
```
TorsteinHonsi, line 2457: So with the `scrollablePlotArea` module we first mutate `chart.plotBox`...

https://github.com/highcharts/highcharts/blob/5553fad1efb18cb65c279c0d7517bd3ac8633db8/ts/Extensions/ScrollablePlotArea.ts#L157-L158

... then we need to unmutate it for this fix. I feel like we need to find a smarter way. Maybe we should store a copy of the original `plotBox` as `chart.scrollablePlotBox`? Or maybe more intiitively, the other way around, avoid mutating the `plotBox` and create a new `scrollablePlotBox`. So wherever we need to access the full width, access `(chart.scrollablePlotBox || chart.plotBox).width`. But first see in what cases the extended width is used.

TorsteinHonsi (changes requested):

TorsteinHonsi (approved):
Thanks!

PR #15006, opened 2021-01-28
bugfix/13222-fullscreen-exit-size

### LINKED ISSUE #13222: Pie Chart - The chart does not display the original size after exiting full screen mode
#### Actual behaviour
The initial demo version of the pie chart is displayed as with the given sizes.

After running "View full screen" / "Exit full screen" note that the size of the chart changes (it has been reduced or increased).

#### The problem occurs in the official Highcharts demo:

#### As well as in the editor with a simplified code:
https://jsfiddle.net/BlackLabel/weo19uvy/

#### Product version
Highcharts v8.0.4

#### Affected browser(s)
Chrome 80.0.3987.149

khlieng:
Fixed #13222, chart size was wrong after exiting fullscreen.

This also fixes fullscreening of fixed size charts (`options.chart.width`/`height` set).

### DIFF
--- a/js/Extensions/FullScreen.js
+++ b/js/Extensions/FullScreen.js
@@ -100,7 +100,7 @@ var Fullscreen = /** @class */ (function () {
      * @requires    modules/full-screen
      */
     Fullscreen.prototype.close = function () {
-        var fullscreen = this, chart = fullscreen.chart;
+        var fullscreen = this, chart = fullscreen.chart, optionsChart = chart.options.chart;
         // Don't fire exitFullscreen() when user exited using 'Escape' button.
         if (fullscreen.isOpen &&
             fullscreen.browserProps &&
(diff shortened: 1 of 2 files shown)


Waqar06:
When this bug is expected to be fixed? How can I access its solution. I don't see fullscreen.js file in my libraray

KacperMadej:
@Waqar06 
To test a fix before it gets merged into the master branch you could use the commit hash like in this demo:
https://jsfiddle.net/BlackLabel/6vpgLje5/1/

There is no ETA for the fix, but the PR passed all tests and needs only a review and acceptance to be merged into the master branch, so this should be resolved soon - in the upcoming (in the next few days) or the next (in the next few weeks) release.

Waqar06:
@KacperMadej  I have tested it thoroughly, It works fine for normal zoom. The only case I see this failed is when someone zooms his/her browser window then the screen.height returns the wrong value. thus full-screen view will not display the full view of the chart instead it will show a small view. 

For example, if my screen resolution is 1920 x 1080 pixels but I have a zoom browser to 150 % then the screen.height returns 864 height.

khlieng:
@Waqar06 there is no screen.height in this fix, you can just ignore the commented out part in the demo linked above.

TorsteinHonsi (approved):
Thanks!

PR #15007, opened 2021-01-28
Range selector inputs design upgrade

TorsteinHonsi:
New graphic design for the range selector date inputs. Simplified visual appearance and flexible width based on the rendered content.

#### Upgrade note
The range selector date input fields have a new visual appearance. See [this fiddle](https://jsfiddle.net/highcharts/vto4af69/) on how to revert to the old design.

### DIFF
--- a/ts/Extensions/RangeSelector.ts
+++ b/ts/Extensions/RangeSelector.ts
@@ -543,7 +543,8 @@ extend(defaultOptions, {
          * @sample {highstock} stock/rangeselector/styling/
          *         Styling the buttons and inputs
          *
-         * @since     1.3.7
+         * @type   {number|undefined}
+         * @since  1.3.7
          */
         inputBoxWidth: void 0,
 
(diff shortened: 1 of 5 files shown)


ts/Extensions/RangeSelector.ts
```diff
@@ -537,14 +537,16 @@ extend(defaultOptions, {
         inputBoxHeight: 17,
 
         /**
-         * The pixel width of the date input boxes.
+         * The pixel width of the date input boxes. When `undefined`, the width
+         * is fitted to the renderred content.
```
pawelfus, line 541: Typo: `renderred` -> `rendered`

pawelfus (approved):
Looks good, thanks!

Please, don't forget about "Upgrade notes" for the changelog :)

pawelfus:
Just realized that not all visual changes are expected, for example: https://jsfiddle.net/BlackLabel/1bjvftkp

Screenshot 2021-01-28 at 11 46 54

Previous boxes were aligned to the right margin, now it seems inputs are a little misplaced. In the visual tests, the difference is even bigger.

pawelfus (changes requested):
Alignment needs a fix

pawelfus (approved):
Everything looks good now, thanks!

One visual test shows something we can improve in the future: https://jsfiddle.net/BlackLabel/xat1rvop/1/. Right now the arrow is a little bit confusing on a chart without data (like "click on exporting button!")

PR #15011, opened 2021-01-29
bugfix/15009-venn-point-className

### LINKED ISSUE #15009: series.venn.data.className not adding to circle's class attribute since version 8.2
#### Expected behaviour
Previously, specifying the series.venn.data.className property value set the class of the circle element within the Venn SVG. (See JSFiddle of previous behaviour in 8.1: https://jsfiddle.net/079dbo4m/16/)

#### Actual behaviour
Now, the series.venn.data.className property is ignored, and no class attribute is added to the circle elements, meaning styling is not added. (JSFiddle of current behaviour in 8.2: https://jsfiddle.net/cfr0hed4/4/)

Previous: https://jsfiddle.net/079dbo4m/16/
Current: https://jsfiddle.net/cfr0hed4/4/

Steps to reproduce:
1. Create a Venn diagram using highcharts
2. Add series.venn.data with sets and values properties set to render the chart
3. Add the series.venn.data.className property and add a corresponding class to the .css
4. Notice that the class is not pulled through on the element

#### Product version
Highcharts 8.2+

#### Affected browser(s)
This was reproduced on Chrome

khlieng:
Fixed #15009, setting `className` on points in venn series did not work.

### DIFF
--- a/js/Series/Venn/VennPoint.js
+++ b/js/Series/Venn/VennPoint.js
@@ -57,7 +57,7 @@ var VennPoint = /** @class */ (function (_super) {
     return VennPoint;
 }(ScatterSeries.prototype.pointClass));
 extend(VennPoint.prototype, {
-    draw: DrawPointMixin.draw
+    draw: DrawPointMixin.drawPoint
 });
 /* *
  *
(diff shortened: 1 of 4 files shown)


TorsteinHonsi (approved):
Thanks!

PR #15021, opened 2021-02-01
bugfix/12422-boost-zones-zoom

### LINKED ISSUE #9124: Zoom in/out when using zones not working with boost

### LINKED ISSUE #12422: Boosted charts have unboosted data remaining on the chart after zoom and reset zoom
#### Expected behaviour
Start with a boosted chart that uses zones.  Zoom in, chart is no longer boosted.  Reset zoom, chart becomes boosted again.

#### Actual behaviour
Start with a boosted chart that uses zones.  Zoom in, chart is no longer boosted.  Reset zoom, chart becomes boosted again.  However, the unboosted traces remain on the graph.  **Seems to only happen if zones are being used.**

[https://jsfiddle.net/5xq29pwn/](https://jsfiddle.net/5xq29pwn/)

Upon load you'll see:
image
Zoom way in like this:
image
Now reset Zoom:
image

#### Product version
Highcharts & Boost Module, current versions (7.1.1+)

#### Affected browser(s)
Chome, no other browsers tested.

khlieng:
Fixed #12422, #9124, un-boosted graphics remained visible after resetting zoom on boosted charts when using zones.

Closes #9124.

### DIFF
--- a/ts/Extensions/Boost/BoostInit.ts
+++ b/ts/Extensions/Boost/BoostInit.ts
@@ -159,6 +159,15 @@ function init(): void {
             if (this.points || this.graph) {
                 this.destroyGraphics();
             }
+            if ((this as any).getZonesGraphs) {
+                const props: string[][] = (this as any).getZonesGraphs([['graph', 'highcharts-graph']]);
+                props.forEach((prop): void => {
+                    const zoneGraph: Highcharts.SVGElement = (this as any)[prop[0]];
+                    if (zoneGraph) {
+                        (this as any)[prop[0]] = zoneGraph.destroy();
+                    }
+                });
+            }
 
             // If we're rendering per. series we should create the marker groups
             // as usual.
--- a/ts/Extensions/Boost/BoostInit.ts
+++ b/ts/Extensions/Boost/BoostInit.ts
@@ -159,15 +159,6 @@ function init(): void {
             if (this.points || this.graph) {
                 this.destroyGraphics();
             }
-            if ((this as any).getZonesGraphs) {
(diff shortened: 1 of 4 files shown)


ts/Extensions/Boost/BoostInit.ts
```diff
@@ -159,6 +159,15 @@ function init(): void {
...
             }
+            if ((this as any).getZonesGraphs) {
+                const props: string[][] = (this as any).getZonesGraphs([['graph', 'highcharts-graph']]);
+                props.forEach((prop): void => {
+                    const zoneGraph: Highcharts.SVGElement = (this as any)[prop[0]];
+                    if (zoneGraph) {
+                        (this as any)[prop[0]] = zoneGraph.destroy();
+                    }
+                });
+            }
```
TorsteinHonsi, line 170: I think this block should be in the `destroyGraphics` function itself along with the destruction of the graphs and the points.

TorsteinHonsi (changes requested):

TorsteinHonsi (approved):

PR #15024, opened 2021-02-01
bugfix/13875-lastVisiblePrice-and-crosshairs-options

### LINKED ISSUE #13875: lastVisiblePrice label is overriding the yAxis crosshairs label settings.
#### Expected behaviour
Individual settings for the `yAxis.crosshair.label` should be respected while using `lastVisiblePrice`

#### Actual behaviour
The `yAxis.crosshair.label` background color is overridden.

1/ Without `lastVisiblePrice.label` enabled the `yAxis.crosshair.label` background color is red.
2/ After enabling `lastVisiblePrice`, the background it will no longer be red.

https://jsfiddle.net/BlackLabel/0gzjyqrn/

#### Product version
8.1.2

#### Affected browser(s)
All

karolkolodziej:
Fixed #13875, crosshair labels options were inherited from lastVisiblePrice.

The issue does not occur on the master branch, I only added a test to prevent from happening again.

### DIFF
--- a/samples/unit-tests/indicator-price-indicator/recalculations/demo.js
+++ b/samples/unit-tests/indicator-price-indicator/recalculations/demo.js
@@ -147,3 +147,54 @@ QUnit.test(
         );
     }
 );
+
+QUnit.test('CurrentPriceIndicator && yAxis crosshair label options, #13875.', function (assert) {
+    const chart = Highcharts.stockChart('container', {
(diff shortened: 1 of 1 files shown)


karolkolodziej:
Looks like this issue was raised in #14495 and solved via #14523 PR. 
So additional test seems unnecessary.

PR #15027, opened 2021-02-02
Tests/a11y portal demo tweaks

oysteinmoseng:
Tweaks to demos for use in feature overview on a11y portal.


TorsteinHonsi (approved):
Thanks!

PR #15032, opened 2021-02-02
bugfix/12031-tooltip-distance-scale

### LINKED ISSUE #12031: Y pointer hover position after transform scale on line charts (after render)
#### Expected behaviour

When a DOM ancestor of a chart div has a transform scale value > 1, the tooltip should correctly track the X,Y position of the rendered points on hover.

NOTE: This issue is similar to #11975.  The proposed solution, i.e., to re-render the chart after scaling the container, corrects the issue for the X-coordinate but not the Y-coordinate.

#### Actual behaviour
If the container scale is greater than 1, the Y-coordinate of the tooltip is incorrect.  It appears above the point by an offset that appears to be proportional to the scale factor.

https://jsfiddle.net/tkmiller3/5jgk6c98/

Use the +/- buttons to scale the ancestor.  Note the Y position of the tooltip when hovering over the points.

#### Product version
Highcharts v7.2.0

khlieng:
Fixed #12031, tooltip `distance` did not scale correctly with `outside` set to `false`.

### DIFF
--- a/js/Core/Tooltip.js
+++ b/js/Core/Tooltip.js
@@ -584,8 +584,9 @@ var Tooltip = /** @class */ (function () {
          */
         firstDimension = function (dim, outerSize, innerSize, scaledInnerSize, // #11329
         point, min, max) {
-            var scaledDist = dim === 'y' ?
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (approved):
Thanks!

PR #15035, opened 2021-02-02
docs/15034-rangeselector-enabled

### LINKED ISSUE #15034: Wrong typing for `RangeSelectorOptions.enabled` in 9.0.0
#### Expected behaviour

The type of `RangeSelectorOptions.enabled` should be `boolean` (as it was before version 9), or it should be documented.

#### Actual behaviour

Since V9, the type is `string`, and the documentation doesn't say what the accepted values are.
The example about this option (https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/stock/rangeselector/enabled/) still shows `false`, a boolean, not a string as value.

#### Product version

9.0.0

#### Affected browser(s)
NA

khlieng:
Fixed #15034, `rangeSelector.enabled` missed doclet type.

### DIFF
--- a/ts/Extensions/RangeSelector.ts
+++ b/ts/Extensions/RangeSelector.ts
@@ -428,7 +428,7 @@ extend(defaultOptions, {
          * @sample {highstock} stock/rangeselector/enabled/
          *         Disable the range selector
          *
-         * @type {boolean}
+         * @type {boolean|undefined}
          * @default {highstock} true
          */
         enabled: void 0,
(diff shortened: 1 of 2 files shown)


ts/Extensions/RangeSelector.ts
```diff
@@ -428,6 +428,7 @@ extend(defaultOptions, {
          * @sample {highstock} stock/rangeselector/enabled/
          *         Disable the range selector
          *
+         * @type {boolean}
```
TorsteinHonsi, line 431: There's now a mismatch between the doclet `@type` and the actual TypeScript type. I think it should be `{boolean|undefined}`. 

https://github.com/highcharts/highcharts/blob/ea08706707c0532d71e0398b93ec9bc331c8a28e/ts/Extensions/RangeSelector.ts#L116

TorsteinHonsi (changes requested):

TorsteinHonsi (approved):

PR #15039, opened 2021-02-03
Fix HTML parsing in Phantom

### LINKED ISSUE #15037: Version 9.0.0 fails with Batik
#### Expected behaviour
No errors

#### Actual behaviour
Fails with DOMParser because it can't parse HTML

All charts

#### Product version
v9.0.0

#### Affected browser(s)
Batik

TorsteinHonsi:
Fixed #15037, a regression causing failure with text parsing in PhantomJS.

___

The feature check was too narrow, used browser sniffing instead of feature sniffing.

@cvasseng Can you verify that this works in PhantomJS without the `DOMParser` polyfill?


cvasseng:
Confirmed working on `node-export-server` with no polyfill. :)

cvasseng (approved):

PR #15043, opened 2021-02-03
bugfix/11930-yAxis-remove-options

### LINKED ISSUE #11930: Highstock - Dynamically added and removed yAxis is unnecessarily placed in exported chart
#### Expected behaviour

chart (78)

#### Actual behaviour

chart (77)

1. Open jsfiddle: http://jsfiddle.net/BlackLabel/zuqs0bmk/
2. Export the chart

#### Product version
Highstock v7.2.0

#### Affected browser(s)
All

**Original use case:** https://github.com/highcharts/highcharts-react/issues/161

khlieng:
Fixed #11930, removing dynamically added y-axes did not work correctly with `navigator` enabled, causing the removed axis to show in exported charts.

### DIFF
--- a/ts/Core/Axis/Axis.ts
+++ b/ts/Core/Axis/Axis.ts
@@ -7927,8 +7927,20 @@ class Axis {
         erase(chart.axes, this);
         erase((chart as any)[key], this);
 
-        if (isArray((chart.options as any)[key])) {
-            (chart.options as any)[key].splice(this.options.index, 1);
+        const options: Highcharts.AxisOptions[] = (chart.options as any)[key];
+
+        if (isArray(options) && defined(this.options.index)) {
+            const option = options[this.options.index];
+            if (option && option.index === this.options.index) {
+                options.splice(this.options.index, 1);
+            } else {
+                for (let i = 0; i < options.length; i++) {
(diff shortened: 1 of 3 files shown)


ts/Core/Axis/Axis.ts
```diff
@@ -7927,8 +7927,20 @@ class Axis {
         erase(chart.axes, this);
         erase((chart as any)[key], this);
 
-        if (isArray((chart.options as any)[key])) {
-            (chart.options as any)[key].splice(this.options.index, 1);
+        const options: Highcharts.AxisOptions[] = (chart.options as any)[key];
+
+        if (isArray(options) && defined(this.options.index)) {
```
TorsteinHonsi, line 7932: Please explain in a comment what is going on here, and add a reference to the issue. Both for the sake of the PR, and for the sake of ourselves finding this piece of code in the future. It might be that we should fix this issue in other parts of the code, but under any circumstance this is so hard to grasp that we need comments to explain it to us. As for alternative fixes, check out what we're doing with the `isInternal` option in other parts of the code base.

TorsteinHonsi (changes requested):

TorsteinHonsi (approved):

PR #15044, opened 2021-02-03
bugfix/11909-boost-update

### LINKED ISSUE #11909: Boost options are not updated when using chart.update
#### Expected behaviour
Boost options should be updated when using chart.update method.

#### Actual behaviour
Boost is not updated, even thought other chart options are updated properly

https://jsfiddle.net/BlackLabel/mwdpo8Lt/4/

#### Product version
Highcharts 7.2.0

khlieng:
Fixed #11909, updating boost options did not work.


TorsteinHonsi (approved):
Thanks!

PR #15047, opened 2021-02-03
bugfix/14197-MACD-changing-shortPeriod

### LINKED ISSUE #14197: MACD indicator dosn't work properly after changing shortPeriod
#### Expected behaviour
Calculate this indicator for the whole available data.

#### Actual behaviour
After changing `shortPeriod` to a smaller number, few last points on the MACD indicator are missing.

https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/stock/indicators/macd/
Change `shortPeriod` to 11.

#### Product version
Highstock JS v8.2.0

#### Affected browser(s)
ALL

karolkolodziej:
Fixed #14197, after changing `shortPeriod` in MACD, points were missing.

First point of MACD should start on the 'longPeriod' and the indicator should end the same as the main series.
A spreadsheet where I calculate the MACD for different params: https://docs.google.com/spreadsheets/d/1tnM4NX5lsKE4mpk2Qm8cyNzbtoRwGIAHnsVv3sypmQk/edit?usp=sharing


samples/unit-tests/indicator-macd/recalculations/demo.js
```diff
@@ -488,3 +488,78 @@ QUnit.test('Test algorithm on data updates.', function (assert) {
...
+                    ]
+                }, {
+                    yAxis: 1,
+                    type: 'macd',
+                    linkedTo: 'main',
+                    params: {
+                        shortPeriod: 10,
+                        longPeriod: 26,
+                        signalPeriod: 9,
+                        period: 26
```
pawelfus, line 553: Can we define only `shortPeriod`?

samples/unit-tests/indicator-macd/recalculations/demo.js
```diff
@@ -488,3 +488,78 @@ QUnit.test('Test algorithm on data updates.', function (assert) {
...
+                    }
+                }
+            ]
+        }),
+        points = chart.series[1].points;
+
+    assert.strictEqual(
+        points[points.length - 1].category,
+        36,
```
pawelfus, line 562: It would be better to check if last point's `x` matches main series last point, for example: `series[0].points[points.length - 1].category`.

Instead of `point.category` you can use `point.x`

ts/Stock/Indicators/MACD/MACDIndicator.ts
```diff
@@ -418,7 +418,8 @@ class MACDIndicator extends SMAIndicator {
                 defined(shortEMA[i + (params.shortPeriod as any) + 1][0])
             ) {
                 MACD.push([
-                    shortEMA[i + (params.shortPeriod as any) + 1][0],
+                    (params.shortPeriod as number) >= 12 ?
```
pawelfus, line 421: Does not work with `shortPeriod` bigger than 12: http://jsfiddle.net/BlackLabel/30t4p9me/6/ - first points are missing in the bottom pane

pawelfus (changes requested):

pawelfus (approved):
Nice catch!

raf18seb (approved):

ts/Stock/Indicators/MACD/MACDIndicator.ts
```diff
@@ -374,7 +374,8 @@ class MACDIndicator extends SMAIndicator {
         series: TLinkedSeries,
         params: MACDParamsOptions
     ): (IndicatorValuesObject<TLinkedSeries>|undefined) {
-        var j = 0,
+        var indexToShift: number = (params.longPeriod as any) - (params.shortPeriod as any), // #14197
```
pawellysy, line 377: isn't it better to have `as number` ?

pawellysy (approved):
One minor change, apart from that looks good!

ts/Stock/Indicators/MACD/MACDIndicator.ts
karolkolodziej: `params.longPeriod` type everywhere is declared as `any`
I think the convention is to leave types of previously declared variable that during the TS refactoring they all might be changed together to better types.

TorsteinHonsi (changes requested):
Thanks! Please resolve the conflicts, then after CI it's ready to merge.

TorsteinHonsi (approved):

PR #15051, opened 2021-02-04
bugfix/15045-centerInCategory-columnrange

### LINKED ISSUE #15045: 'adjustForMissingColumns' doesn't work for columnrange series type
#### Expected behaviour
The visible columns should ignore the missing ones and should be centered in their group when some columns are missing. 

#### Actual behaviour
The columns are positioned in a way, to leave the space for missing columns: 

https://jsfiddle.net/BlackLabel/j6rhLex9/

#### Product version
v9.0.0

#### Affected browser(s)
all

khlieng:
Fixed #15045, `centerInCategory` did not work with `columnrange` series.


TorsteinHonsi (approved):
Thanks!

PR #15052, opened 2021-02-04
bugfix/15042-ikh-line-color-switcharoo

### LINKED ISSUE #15042: IKH indicator has wrong color after interacting with chart
#### Expected behaviour
Lines should be the same color as initially (color taken from `chart.colors` array)

#### Actual behaviour
Any interaction (causing chart to be redrawn) renders IKH lines as red.

https://jsfiddle.net/znw4khrj/
Steps:
- click on any `rangeSelector` button, drag&drop or use navigator

ikh

khlieng:
Fixed #15042, IKH indicator lines changed color on the first redraw.


TorsteinHonsi (approved):
Thanks!

PR #15058, opened 2021-02-05
bugfix/11493-split-tooltip-3

### LINKED ISSUE #11493: Y Axis scaled using set Extremes Is inappropriate

khlieng:
Fixed #11493, split tooltip showed outside y-axis for points outside y-axis extremes.

Commented out some assertions until the behaviour is settled.

Demo: https://jsfiddle.net/akvmp0qy/1/

Throwing in some `if (yAxis.toPixels(yAxis.dataMin) > scrollTop)` and vice versa might be nice, to avoid showing a clamped tooltip if theres just a completely empty part of a yAxis within the visible part of the plot?

### DIFF
--- a/js/Core/Tooltip.js
+++ b/js/Core/Tooltip.js
@@ -1006,12 +1006,13 @@ var Tooltip = /** @class */ (function () {
             }
             else {
                 var xAxis = series.xAxis, yAxis = series.yAxis;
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (approved):
Thanks!

Issue #15062, opened 2021-02-06
Offline export to PNG/PDF not rendering properly with rangeSelector (BTC, ETH Unicode issue fixed!).


roychowdhuryrohit-dev:
**Offline export to PNG/PDF not rendering properly with rangeSelector.**

Browser: Safari, Chrome

Font: Google's Noto Sans

<img width="1070" alt="Screenshot 2021-02-06 at 19 42 28" src="https://user-images.githubusercontent.com/24897721/107120449-be88c080-68b3-11eb-8d98-86eb83fe0051.png">

UPDATE: Mentioning a separate font for exporting doesn't seem to solve the issue.

```
exporting: {
    chartOptions: {
        chart: {
            style: { 
                fontFamily: 'Times New Roman', 
            }
        }
    }
},
```

UPDATE: Unicode characters like Bitcoin(₿), Ethereum(Ξ), Euro(€) are not rendering properly in offline export (only with _Download PDF document_).

<img width="1379" alt="Screenshot 2021-02-06 at 21 18 23" src="https://user-images.githubusercontent.com/24897721/107123383-41654780-68c3-11eb-9bbb-d5bf78b491f2.png">

UPDATE: Also facing an issue in Chrome where it's impossible to type the correct dates unlike in Safari where I'm able to type the whole date just like before. Suppose when I want to set 'November' as month, I cannot type two 1s to represent number 11. Although it is possible to set 'November' month by clicking on calendar datepicker as a workaround.

KacperMadej:
Hi @roychowdhuryrohit-dev 

Thank you for reporting the problem.

Please add a live demo for the issue.
Please let us know the browsers' versions and the OS with the version that you are using to recreate the problem.

If you have multiple problems then please open a new ticket for each different problem.

roychowdhuryrohit-dev:
![demo](https://user-images.githubusercontent.com/24897721/107153565-34fcef80-6994-11eb-8509-7004a0005774.gif)

OS: macOS 10.15.7
Browser: Safari 14.0.2/Chrome 88.0.4324.96
Font: Default font
Theme: Same problem with default theme.

roychowdhuryrohit-dev:
Finally got it working.
[...]

Issue #15060, opened 2021-02-05
Stock rangeSelector UI issue in Internet Explorer 11 IE11 in version 9


raf18seb:
#### Expected behaviour
![image](https://user-images.githubusercontent.com/23310742/107038372-b3069e00-67bc-11eb-963f-d26df9873824.png)

#### Actual behaviour
The new UI (v9) of rangeSelector doesn't fully work in IE11.
![image](https://user-images.githubusercontent.com/23310742/107038415-c580d780-67bc-11eb-80a8-555fc36a1d6f.png)

#### Live demo with steps to reproduce
Open in IE11
https://www.highcharts.com/demo/stock/basic-line

#### Product version
Highcharts Stock 9

#### Affected browser(s)
Internet Explorer 11

khlieng:
This is probably because of the shorter width, Safari also has some problems but IE is trying to cram a clearing cross in there aswell.

roychowdhuryrohit-dev:
Also facing an issue in Chrome where it's impossible to type the correct dates unlike in Safari where I'm able to type the whole date just like before. Suppose to set 'November' month, I cannot type two 1s to represent number 11.

KacperMadej:
@roychowdhuryrohit-dev 
Please provide a live demo link and precise steps required to recreate the problem.

I have tried with [the basic demo](https://www.highcharts.com/demo/stock/basic-line). The "From" input works fine and "To" intpu (2nd, the one on the right side) has no data from that month in that year, so doesn't work either by input or datepicker - this is expected and correct. Tested on Chrome / Windows 10.

khlieng:
@KacperMadej set the from input to 2019 first, then try to enter 1 in as the month. This is related to the data starting after January.

roychowdhuryrohit-dev:
@KacperMadej The issue persists in your basic demo as well. 
First click “All” button -> Click on the month section of “From” -> Try typing 11 (6/11/2019).

PR #15066, opened 2021-02-08
bugfix/rangeselector-november

khlieng:
Fixed issue mentioned in #15060 and #15062, it was not possible to enter months starting with 1 into the range selector date input under certain conditions. When typing the extremes now only update on blur and when pressing enter, which is the same behaviour the text inputs have.


TorsteinHonsi (approved):
Thanks!

PR #15068, opened 2021-02-08
Fixed #5244, issue with internal references

### LINKED ISSUE #5244: The series path is out of clippath when use history.push()

TorsteinHonsi:
Fixed #5244, internal SVG references were wrong after location was changed using `history.pushState`. Removed workaround for old browser bug.

### DIFF
--- a/js/Core/Options.js
+++ b/js/Core/Options.js
@@ -3350,7 +3350,7 @@ H.defaultOptions = {
          */
         borderRadius: 3,
         /**
-         * For series on a datetime axes, the date format in the tooltip's
+         * For series on datetime axes, the date format in the tooltip's
          * header will by default be guessed based on the closest data points.
          * This member gives the default string representations used for
          * each unit. For an overview of the replacement codes, see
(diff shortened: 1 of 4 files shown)


goransle (approved):
Looks good 👍

pawelfus (approved):
Looks good!
At first, I thought `innerHML = ...` may cause some security issues, but `uniqueKey` can't be overridden, so it's okay 👍

TorsteinHonsi:
> Looks good!
> At first, I thought `innerHML = ...` may cause some security issues, but `uniqueKey` can't be overridden, so it's okay 👍

Thanks! I'll do a try with an AST too, see how it turns out.

PR #15070, opened 2021-02-08
bugfix/15069-panning-default-type

### LINKED ISSUE #15069: Panning default type is not set.
#### Expected behaviour
Chart options merged with defaults (with values as [API reference describes](https://api.highcharts.com/highcharts/chart.panning.type)).

#### Actual behaviour
The type is not set causing an error in some cases - see below.

https://jsfiddle.net/BlackLabel/u6dw9smv/

1. Shift + drag&drop to pan to the right side.
2. Error in the console `Cannot read property 'match' of undefined` as the `type` option is not defined.

#### Product version
v9.0.0

khlieng:
Fixed #15069, `panning` default `type` was missing.

This also fixes missing panning in maps API docs.


js/Core/Options.js
```diff
@@ -993,29 +993,34 @@ H.defaultOptions = {
          * @sample  {highcharts} highcharts/chart/pankey/ Zooming and panning
          * @sample  {highstock} stock/chart/panning/ Zooming and xy panning
          *
-         * @product highcharts highstock gantt
+         * @product highcharts highstock highmaps gantt
```
KacperMadej, line 996: Listing all 4 products should be the same as not setting `@product` at all.
Example `chart.styledMode`:
 https://github.com/highcharts/highcharts/blob/a2d73c62eaef8f597f57e190d2e6801a2351eef4/js/Core/Options.js#L1041-L1051

TorsteinHonsi (approved):
Thanks!

Issue #14930, opened 2021-01-14
RSI indicator on the line type chart.


miloszwojcik:
#### Expected behavior
Add an RSI Indicator line to the chart

#### Actual behaviour
Add a blank series.

#### Live demo with steps to reproduce
https://jsfiddle.net/12hnszkg/9/

Replicate.

Add an RSI indicator from the indicators list. Note nothing happens. You should get a new series on the chart. This happens only on the line type chart. On ohlc it's working fine. 

The issue might also exist in other indicators like: PSAR, NATAR, KETLNER CHANNEL, VWAP, VOLUME BY PRICE, MONEY FLOW INDEX, ACCUMULATION/DISTRIBUTION

KacperMadej:
Hi @miloszwojcik 

Thank you for reporting the problem.

Some indicators require OHLC data format because some parts of the data are used for the indicator calculations. **RSI requires close value.**

It is possible to create custom technical indicators as explained in the docs here: https://www.highcharts.com/docs/stock/custom-technical-indicators

**Workaround**: a demo with modified RSI to work with line series data - https://jsfiddle.net/BlackLabel/rps7f9to/

sebastianbochan:
Hi @miloszwojcik,
You should declare a [useOhlcData](https://api.highcharts.com/highstock/series.line.useOhlcData) option.

Demo:
- https://jsfiddle.net/BlackLabel/h24nb6Lu/

miloszwojcik:
I think this would be much more efficient if we could operate just on the line data [open date, value] and use the RSI. 
RSI just needs `[open date, close value]` so I'm confused why ohlc is required in this case. Does RSI indicator in high charts necessarily need some other values than  `[open date, close value]`?

pawelfus:
Hi @miloszwojcik 

RSI does not need more information, only the close values. We can assume that the only value provided in a line series is what a user wants to use for calculations. Nice catch, thanks!

You mentioned other technical indicators:
- PSAR requires high and low values
- NATR is based on ATR, which needs high and low values
- Keltner Channel is based on ATR, also requires high and low
- VWAP does not require OHLC data, should work with line and OHLC series
- Volume by Price, requires only close values, could be improved too
- in MFI we use `(H+L+C)/3` to calculate the typical price
- Accumulation/Distribution requires low, high, and close values

[...]

PR #15071, opened 2021-02-08
Enhancement/14930-rsi-indicator-on-line-series

pawellysy:
Added line series as available type for the RSI Indicator. See #14930.

~Feature #14930, Added line series as available type for RSI Indicator.~


samples/unit-tests/indicator-rsi/recalculations/demo.js
```diff
@@ -47,44 +47,41 @@ QUnit.test('Test RSI calculations on data updates.', function (assert) {
...
+        [430000, 470000, 420000, 457835],
+        [430000, 470000, 420000, 453548],
+        [430000, 470000, 420000, 440288],
+        [430000, 470000, 420000, 441783],
+        [430000, 470000, 420000, 442181],
+        [430000, 470000, 420000, 445672],
+        [430000, 470000, 420000, 434205],
+        [430000, 470000, 420000, 426628],
+        [430000, 470000, 420000, 431314]], false);
+
```
raf18seb, line 84: The previous code format was more natural, I would stick to that one.

samples/unit-tests/indicator-rsi/recalculations/demo.js
```diff
@@ -150,4 +147,81 @@ QUnit.test('Test RSI calculations on data updates.', function (assert) {
...
+            440288,
+            441783,
+            442181,
+            445672,
+            434205,
+            426628,
+            431314
+        ],
+        id: 'line-series',
+        yAxis: 0
```
raf18seb, line 188: Isn't `yAxis: 0` by default?

samples/unit-tests/indicator-rsi/recalculations/demo.js
```diff
@@ -150,4 +147,81 @@ QUnit.test('Test RSI calculations on data updates.', function (assert) {
...
+    });
```
raf18seb, line 189: You could set the `redraw` flag here to `false` to improve the performance (the chart will be redrawn within the next `chart.addSeries()` method call.

pawelfus, line 189: With multiple dynamic methods, set `redraw=false`

pawelfus, line 189: With multiple dynamic methods, set `redraw=false` :)

ts/Stock/Indicators/RSI/RSIIndicator.ts
```diff
@@ -88,9 +88,10 @@ class RSIIndicator extends SMAIndicator {
         series: TLinkedSeries,
         params: RSIParamsOptions
     ): (IndicatorValuesObject<TLinkedSeries>|undefined) {
+
```
raf18seb, line 91: Empty unnecessary line

ts/Stock/Indicators/RSI/RSIIndicator.ts
```diff
@@ -106,20 +107,28 @@ class RSIIndicator extends SMAIndicator {
             change: number,
             avgGain: number,
             avgLoss: number,
-            i: number;
+            i: number,
+            values: Array<number>;
 
         // RSI requires close value
```
pawelfus, line 113: I think it does not anymore ;)

ts/Stock/Indicators/RSI/RSIIndicator.ts
```diff
@@ -88,9 +88,10 @@ class RSIIndicator extends SMAIndicator {
         series: TLinkedSeries,
         params: RSIParamsOptions
     ): (IndicatorValuesObject<TLinkedSeries>|undefined) {
+
         var period = (params.period as any),
             xVal: Array<number> = (series.xData as any),
-            yVal: Array<Array<number>> = (series.yData as any),
+            yVal: Array<Array<number>> | Array<number> = (series.yData as any),
```
raf18seb, line 94: Hmm I know it is a small detail, but it would look more natural if you defined that TS definition in the opposite order
`Array<number> | Array<Array<number>>`.

ts/Stock/Indicators/RSI/RSIIndicator.ts
```diff
@@ -106,20 +107,28 @@ class RSIIndicator extends SMAIndicator {
...
+        if ((xVal.length < period)) {
             return;
         }
+        if (!isArray(yVal[0])) {
+            values = yVal as Array<number>;
+        } else if (yVal[0].length < 4) {
+            values = (yVal as Array<Array<number>>).map(
+                (value: Array<number>): number =>
+                    value[(params.index as number) || (yVal as Array<Array<number>>)[0].length - 1]
+            );
```
pawelfus, line 123: I'm not sure what is the purpose of this block. Is this support for series types that have less than four values in data (e.g. arearange)? Could we instead check this way: 

```js
if (!isArray()) {
 ...
} else {
  let index = Math.min(params.index, yVal[0].length - 1);
  values = (yVal as any).map((value: Array<number>): number => value[index]);
}
```

pawelfus (changes requested):
Additionally, make sure API is up to date:
https://api.highcharts.com/highstock/series.rsi.params.index - default value in docs says it's 0, while it's 3 (close)

ts/Stock/Indicators/RSI/RSIIndicator.ts
```diff
@@ -106,20 +107,28 @@ class RSIIndicator extends SMAIndicator {
...
 
         // RSI requires close value
-        if (
-            (xVal.length < period) || !isArray(yVal[0]) ||
-            yVal[0].length !== 4
-        ) {
+        if ((xVal.length < period)) {
             return;
         }
+        if (!isArray(yVal[0])) {
```
raf18seb, line 117: I think the `isNumber(yVal[0])` check would be simpler.

ts/Stock/Indicators/RSI/RSIIndicator.ts
```diff
@@ -106,20 +107,28 @@ class RSIIndicator extends SMAIndicator {
...
+            values = yVal as Array<number>;
+        } else if (yVal[0].length < 4) {
```
raf18seb, line 119: Why `yVal[0].length < 4`? If this is a preparation for cases like `HLC` series, could you add a comment explanation?

ts/Stock/Indicators/RSI/RSIIndicator.ts
```diff
@@ -106,20 +107,28 @@ class RSIIndicator extends SMAIndicator {
...
+            values = (yVal as Array<Array<number>>).map(
+                (value: Array<number>): number =>
+                    value[(params.index as number) || (yVal as Array<Array<number>>)[0].length - 1]
```
raf18seb, line 122: Can you assign `(yVal as Array<Array<number>>)[0].length - 1` to a variable outside of `Array.map`? It's just an index (number). With dozens thousands of points, we could have one simple variable of `number` type instead of calculating `(yVal[0].length - 1` each iteration many times.

raf18seb (changes requested):

raf18seb:
I have one general question - do we have to add two separate `line` and `rsi` series? Can't we just update
```js
chart.series[0].update({
    type: 'line',
    data: [...]
});
```
?

ts/Stock/Indicators/RSI/RSIIndicator.ts
karolkolodziej, line 119: Why `4`? What does it represent?

karolkolodziej (changes requested):

ts/Stock/Indicators/RSI/RSIIndicator.ts
pawellysy: well, in fact as you suggested earlier if I can calculate the index in a way: `Math.min(params.index, yVal[0].length - 1)` it 
can ensure, that we don't reach outside of the bounds of the array, so this `else if` is redundant.

ts/Stock/Indicators/RSI/RSIIndicator.ts
pawellysy: well, in fact as you suggested earlier if I can calculate the index in a way: Math.min(params.index, yVal[0].length - 1) it
can ensure, that we don't reach outside of the bounds of the array, so this else if is redundant.

ts/Stock/Indicators/RSI/RSIIndicator.ts
pawellysy: As @raf18seb suggested earlier, it can prepare us to the cases, when you can define the RSI indicator for some other types of series like HLC, range etc.

ts/Stock/Indicators/RSI/RSIIndicator.ts
pawellysy: It already was checked this way, and In my opinion, if you check if it is an array it is more readable.

raf18seb: It was checked this way because, previously, it could be only an array. It made sense then, but now you changed it. Now, it can be either number or array so, if you are checking whether something is a number, then, in my humble opinion, a simple check `isNumber()` is just perfect for this. It's your call

raf18seb: Ok, I reconsidered, in this case both are ok 👍

samples/unit-tests/indicator-rsi/recalculations/demo.js
pawellysy: I switched from creating new series to updating the previous series.

ts/Stock/Indicators/RSI/RSIIndicator.ts
raf18seb: Both are correct and I shouldn't request changes in this case. However, I disagree that it's more readable this way, because you don't indicate what type is in the first check (we only know that it's not an array).
The most readable and safe check would be:
```js
if (isNumber()) {
...
} else if (isArray() && array.length < 4) {
...
}
```

pawellysy: currently this block looks like this:  
```
 if (!isArray(yVal[0])) {
            values = yVal as Array<number>;
        } else {
            // in case of the situation, where the series type has data length shorter
            // then 4 (HLC, range), this ensures that we are not trying to
            // reach the index out of bounds
            index = Math.min(params.index as number, yVal[0].length - 1);
            values = (yVal as Array<Array<number>>).map((value: Array<number>): number => value[index]);
        }

```

pawellysy: so in your example it would be ```js
if (isNumber()) {
...
} else if (isArray() && array.length < 4) {
...
} else{ //??????
}
```

File: ts/Stock/Indicators/RSI/RSIIndicator.ts [hunk 11]
```diff
@@ -25,6 +25,8 @@ const {
 import U from '../../../Core/Utilities.js';
 const {
     isArray,
+    isNumber,
+    error,
```
pawelfus, line 29: I don't see any usage of this, is it necessary? :)

ts/Stock/Indicators/RSI/RSIIndicator.ts
```diff
@@ -99,27 +102,34 @@ class RSIIndicator extends SMAIndicator {
...
             return;
         }
 
+        if (isNumber(yVal[0])) {
+            values = yVal as Array<number>;
+        } else {
+            // in case of the situation, where the series type has data length
+            // longer then 4 (HLC, range), this ensures that we are not trying
+            // to reach the index out of bounds
+            index = Math.min(params.index as number, yVal[0].length - 1);
```
pawelfus, line 125: You can reuse `index` variable from the definitions, instead of `params.index`

pawelfus (changes requested):
Minor changes, overall good!

ts/Stock/Indicators/RSI/RSIIndicator.ts
```diff
@@ -64,7 +64,8 @@ class RSIIndicator extends SMAIndicator {
     public static defaultOptions: RSIOptions = merge(SMAIndicator.defaultOptions, {
         params: {
             period: 14,
-            decimals: 4
+            decimals: 4,
+            index: 3
```
karolkolodziej, line 68: An index declared here is not consistent with the [API](https://api.highcharts.com/highstock/series.rsi.params), which one is correct?

karolkolodziej (changes requested):

ts/Stock/Indicators/RSI/RSIIndicator.ts
pawellysy: Previously, the default index value was hard coded in the getValues function, and wasn't taking into consideration the value passed through series configuration. Because the index value was not defined here, the api specification was inherited from SMA indicator, where default index value was 3. Adding the default index value here not only allows us to change the configuration of the RSI indicator, which allows to calculate the indicator over other values, but also changes the API to the correct value.

pawelfus (approved):
All good, thanks!

karolkolodziej (approved):

raf18seb (approved):
Thanks!

PR #15073, opened 2021-02-08
bugfix/11293-boost-responsive-duplicate

### LINKED ISSUE #11293: Boost and responsive rules makes chart is drawing twice
As in the title: Boost module and responsive rules to hide chart title and subtitle make chart is drawing twice.

[Demo](https://jsfiddle.net/BlackLabel/9moru07g/). After window resizes, the chart is rendering properly. 

#### Product version

#### Affected browser(s)
All

khlieng:
Fixed #11293, boosted chart rendered duplicate series when combined with responsive options.


TorsteinHonsi (approved):
Thanks!

PR #15078, opened 2021-02-09
samples/organization

goransle:
Updated the `organization-chart` demo to more closely resemble the current organization. Also updated the photos to those on the "About us" page.


TorsteinHonsi (approved):
Thanks!

Issue #15076, opened 2021-02-08
SVG filtered out of tooltip HTML


harrismd:
#### Expected behaviour
SVG graphics render in tooltips when useHTML is true

#### Actual behaviour
SVG graphics are filtered out of the HTML and are not displayed in the tooltip.

#### Live demo with steps to reproduce
https://jsfiddle.net/mharrisqb/9wrbn7fz/
Simply hover over the chart and notice there is no blue circle to the left of the text "Series 1" and no black diamond to the left of the text "Series 2".

#### Product version
Highcharts 9.0.0

#### Affected browser(s)
All

KacperMadej:
Hi @harrismd 

Thank you for reporting the problem.

In the browser console you should be able to find a warning - `(highcharts.src.js:518) Highcharts warning: Invalid tagName 'svg' in config`. The behavior is by design:

> Improved security by filtering all HTML configuration strings and abstract syntax trees through allow lists before inserting into the DOM.

(v9.0.0 - first point - https://www.highcharts.com/blog/changelog/)

See the related issue for more info: https://github.com/highcharts/highcharts/issues/15031

harrismd:
Shall we change this to a feature request then? Highcharts used to support <SVG> graphics in HTML tooltips. It was removed in version 9. We'd like it back.

harrismd:
I would argue that supporting `<svg>` in HTML is just as important as supporting `<img>`.

Yash2737:
Issue for me, cannot render svg html directly

PR #15080, opened 2021-02-09
Slight refactoring of SVG clip hit test

TorsteinHonsi:
Possible addition to #15068


KacperMadej:
This leads to an unpleasant side-effect of allowing `svg` tag, but adding it without the NS correct for an SVG.
(The unpleasant part is that the element is there and there is no warning in the browser console while it doesn't work.)
Demo (based on the issue https://github.com/highcharts/highcharts/issues/15076): https://jsfiddle.net/BlackLabel/15purg9v/

KacperMadej (approved):

Issue #15026, opened 2021-02-02
Simple approximation demo


pawelfus:
We need to create a simple `dataGrouping.approximation` demo, and present it here:

- https://api.highcharts.com/highstock/series.line.dataGrouping.approximation

We already have an example with grouping points and using custom data. In addition, we can also provide a very simple demo that returns simply the first point from the group (instead of average value).

PR #15081, opened 2021-02-09
enhancement/15026-new-approximation-demo

karolkolodziej:
Feature #15026, added simple approximation demo to the documentation.


samples/stock/plotoptions/series-datagrouping-simple-approximation/demo.js
```diff
@@ -0,0 +1,43 @@
...
+        data: ADBE
+    }, {
+        name: 'ADBE start',
+        data: ADBE,
+        dataGrouping: {
+            approximation: function () {
+                var groupStart = this.dataGroupInfo.start;
+
+                return this.options.data[groupStart][1];
+            }
```
pawelfus, line 29: Can be simpler :) `approximation` takes an argument (or many if data is e.g. OHLC), which is an array of values:
```js
            approximation: function (data) {
                return data[0];
            },
```

pawelfus (changes requested):
See comment

samples/stock/plotoptions/series-datagrouping-simple-approximation/demo.js
```diff
@@ -0,0 +1,41 @@
...
+                return groupData[0];
+            }
+        }
+    }, {
+        name: 'ADBE end',
+        data: ADBE,
+        dataGrouping: {
+            approximation: function (groupData) {
+                var groupEnd = this.dataGroupInfo.start +
+                     this.dataGroupInfo.length - 1;
```
pawelfus, line 35: This doesn't work. `groupData` contains a list of values that should be used to calculate new, grouped value. For example: `[1, 2, 3]`, which means, DataGrouping wants to group these three points into one. Average would be 2, start=1, end=3.

`this.dataGroupInfo` is used when referring original dataset, not `groupData`.

Simple `return groupData[groupData.length - 1];` should be enough :)

pawelfus (changes requested):

samples/stock/plotoptions/series-datagrouping-simple-approximation/demo.html
```diff
@@ -0,0 +1,5 @@
+<div id="container" style="height: 400px; min-width: 600px"></div>
+
+<script src="https://code.highcharts.com/stock/highstock.js"></script>
+<script src="https://code.highcharts.com/stock/modules/exporting.js"></script>
+<script type="text/javascript" src="https://www.highcharts.com/samples/data/three-series-1000-points.js"></script>
```
raf18seb, line 5: Is it necessary? `type="text/javascript"`

raf18seb (commented):
I have only one small question, see the comment above.

samples/stock/plotoptions/series-datagrouping-simple-approximation/demo.html
karolkolodziej: I think it's not necessary, I copied that from [approximation demo](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/stock/plotoptions/series-datagrouping-approximation)

TorsteinHonsi: Not necessary.

TorsteinHonsi (changes requested):
Please fix the comment and merge in the master, then after CI, it should be ready to merge.

pawelfus (approved):
Nice! :)

pawellysy (approved):
Looks good!

raf18seb (approved):
All good, thanks!

TorsteinHonsi (approved):

PR #15091, opened 2021-02-09
docs/dependency wheel API url fix 

katherine-tully:
Dependency wheel API url provided in docs is invalid and results in a 404.


TorsteinHonsi (approved):
Thanks!

PR #15093, opened 2021-02-10
bugfix/10296-navigator-event-leak

### LINKED ISSUE #10296: Calling Series.update results in additional show and hide challbacks execution
#### Expected behaviour

Series.update should not affect how many times show and hide callbacks are executed.

#### Actual behaviour

Each Series.update call results in one additional show and hide callback call (not during update, but for example on legend item click). Both functions are set via plotOptions.

https://jsfiddle.net/futj0a83/

Open developer console, click on button and then toggle menu item.

#### Product version
Highstock JS v7.0.3

#### Affected browser(s)
Tested on Chrome 72.0.3626.121 (Official) (64-bit)

khlieng:
Fixed #10296, navigator leaked event handlers into series and x-axis on `Series.update`.


TorsteinHonsi (approved):
Thanks!

PR #15098, opened 2021-02-10
Added additional lang context for legend item a11y.

oysteinmoseng:
Added additional `item` object context for [lang.accessibility.legend.legendItem](https://api.highcharts.com/highcharts/lang.accessibility.legend.legendItem).
___
Minimal change, just added extra context after user mentioned a use case where it would be useful to have other info than `itemName` in the a11y description of legend items.


TorsteinHonsi (changes requested):
Please merge in the master to make the tests pass

TorsteinHonsi (approved):

PR #15100, opened 2021-02-10
Attempt to fix errors in CI in accessible-venn demo.

oysteinmoseng:
See title. Something fails, could be that fontawesome is not loading in CI, and is causing elements to be undefined? Tried a bandaid fix, crossing fingers.


khlieng:
I think the root problem here is the `setTimeout`, the DOM is probably gone by the time it runs. And this PR (and my exclude PR :() is failing because its trying to run visual tests from master first.

TorsteinHonsi:
This is not needed now (fingers crossed), I made a change in karma so it always runs lolex on visual samples.

TorsteinHonsi:
Hopefully redundant after #15102 .

PR #15106, opened 2021-02-10
homepage animation 2

nasin:
Tryin this again on a new branch.


TorsteinHonsi (changes requested):
Thanks for the PR! I believe the `apple.svg` file is missing now.

TorsteinHonsi (approved):
Thanks!

PR #15111, opened 2021-02-11
bugfix/9998-exporting-menu-clipped

### LINKED ISSUE #9998: Export menu gets clipped if chart size is too small
#### Expected behaviour
There should be an `outside` property for the export menu just like the recently added `outside` property for tooltips so that the export menu does not get clipped with small charts.

#### Actual behaviour
Export menu is clipped.

Open export menus in demo: https://jsfiddle.net/lanimelrok/o2hbp46r/

#### Product version
Highcharts/Highstock 6.2.0

khlieng:
Fixed #9998, exporting menu was clipped below the chart.


TorsteinHonsi (approved):
Thanks!

PR #15112, opened 2021-02-11
Fixed #2284, print media query

### LINKED ISSUE #2284: Printing chart with @media queries
Example: http://jsfiddle.net/j3xvg/4/show/

Simple try to print whole page (don't use exporting module). Bottom div is scaled correctly, chart is cut off.

TorsteinHonsi:
Fixed #2284, charts were clipped when printing the whole page with media query rules.

### DIFF
--- a/js/Extensions/Exporting.js
+++ b/js/Extensions/Exporting.js
@@ -948,18 +948,29 @@ H.post = function (url, data, formAttributes) {
     // clean up
     discardElement(form);
 };
-if (H.isSafari) {
-    H.win.matchMedia('print').addListener(function (mqlEvent) {
-        if (!H.printingChart) {
-            return void 0;
-        }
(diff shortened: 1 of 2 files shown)


TorsteinHonsi:
Going back to draft. The problem with this solution is that it requires the export module in order to print a page correctly. So looking in the `setReflow` function, I found this:

https://github.com/highcharts/highcharts/blob/440b8e004487e808253a48bf8a1a243c1cba885d/ts/Core/Chart/Chart.ts#L1693-L1702

I think we can make that work with Firefox by adding a print media query dynamically.

RemyMachado:
Any update on this fix?
I'm struggling to get a responsive graph on print view on my React/Nextjs app do you have any workaround by the time it gets released?

PR #15114, opened 2021-02-11
bugfix/9738-groupAll-ohlc

### LINKED ISSUE #9738: ohlc / candlestick groupAll + zoom doesnt calculate correctly
#### Expected behaviour
When zoomed and moving left/right chart should include all points in grouping and data shouldnt change once grouped. When moving left / right points shouldnt change at all.

#### Actual behaviour
When moving every point is changing its values. Also xAxis date and actual points are not aligned once zoomed in and moving to far right. 

groupAll: true (all points in zoomed area are changing)
http://jsfiddle.net/79jdzhtw/

groupAll: false (only first and last points in zoomed area are changing)
http://jsfiddle.net/79jdzhtw/4/

Also i tested with both ohlc and candlestick types and same problem. With other types it works as expected.

#### Product version
Highstock 7.0.1 (>=6.1.0)

#### Affected browser(s)
Chrome 71.0.3578.98 (Official Build) (64-bit)

khlieng:
Fixed #9738, `groupAll` did not work correctly with OHLC-based series.

### DIFF
--- a/js/Extensions/DataGrouping.js
+++ b/js/Extensions/DataGrouping.js
@@ -136,6 +136,7 @@ var approximations = H.approximations = {
     }
 };
 var groupData = function (xData, yData, groupPositions, approximation) {
+    var _a;
     var series = this, data = series.data, dataOptions = series.options && series.options.data, groupedXData = [], groupedYData = [], groupMap = [], dataLength = xData.length, pointX, pointY, groupedY, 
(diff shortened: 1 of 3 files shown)


pawelfus (approved):
Looks good, thanks! :)

TorsteinHonsi (approved):
Thanks!

PR #15115, opened 2021-02-11
bugfix/ast-allow-description-list-tags

ryanbonial:
Fixed the removal of description list ( `<dl>`, `<dt>`, and `<dd>`) tags by adding them as allowed tags in the AST.


KacperMadej:
Hi @ryanbonial 

Thank you for submitting the PR.

KacperMadej:
@ryanbonial 
Could you create an issue for the problem that this PR solves? With a live demo and maybe a longer description so we could add a test for the issue and reference the fix later in the changelog., if possible, please.

TorsteinHonsi:
Thanks again! It's okay without an issue, we will add the list tags to the allow list. I'll be adding them in a separate PR so this one will be closed. Leaving it open until then. No further action required.

ryanbonial:
Thank you!

PR #15122, opened 2021-02-12
Bugfix/6332-split-tooltip-on-flags

### LINKED ISSUE #6332: A split tooltip does not work correctly with flags
#### Actual behaviour
When the flag does not cover the point (x value), a tooltip for the flags does not show up - instead the tooltip for the next point is shown (with empty text).

#### Expected behaviour
I think a split tooltip should work in a similar way as a shared tooltip for flags
http://jsfiddle.net/qdg47eu0

http://jsfiddle.net/qdg47eu0/1/
1. Move mouse over the flag

pawellysy:
Fixed #6332, split tooltip doesn't work correctly with flag series type with empty text property.


samples/unit-tests/series-flags/tooltip/demo.js
```diff
@@ -85,5 +85,13 @@ QUnit.test(
...
-    }
-);
+        // Empty shared tooltip visible, when text is not defined in flag point, #6332
+        chart.tooltip.update({ split: true });
+
+        chart.tooltip.refresh(chart.series[2].points[0]);
+        assert.strictEqual(
+            chart.container.getElementsByClassName('highcharts-label-box').length,
+            6,
+            'The tooltip box next to the flag with empty title should be ivisible (#6332).'
```
karolkolodziej, line 95: This test is not checking visibility only length of the label box. 
Isn't there a better way to test that?

karolkolodziej (changes requested):
Fix looks ok, improve the test.

samples/unit-tests/series-flags/tooltip/demo.js
pawellysy: Well, in fact, the description of the test isn't accurate, because the tooltip is not only invisible, it is not created, so I have nothing to reference to. @pawelfus  is there a better way?

raf18seb: You could check whether `charts.tooltip.label.element.childNodes.length === 1` (only tt tooltip is rendered).

Also, there is a typo `ivisible`.

pawelfus: It's checking the number of elements with `label-box` class in the whole chart. We can improve it to check only the number of elements in `.highcharts-tooltip` parent. After the fix, there should be exactly one element (`headerFormat`, the one displayed on xAxis). That would be a nice improvement in the test, because we don't have to worry if more elements on the chart will receive the `label-box` class (e.g. extra buttons from modules etc.).

I don't see any better way to test it: we need to make sure that the box was not created (so it's not empty).

I agree, the test description could be improved a little :)

pawelfus (approved):
Nothing to add, thanks! :)

karolkolodziej (approved):

raf18seb (approved):
All good, thanks!

TorsteinHonsi (approved):
Thanks!

TorsteinHonsi:
Awaiting CI, then ready to merge.

PR #15125, opened 2021-02-12
bugfix/15124-dropdown-no-buttons

### LINKED ISSUE #15124: Highstocks: Uncaught TypeError: Cannot read property 'getBBox' of undefined
The issue is related with range selector I think. So, when the width of chart is too small (smaller than the expected width to fit the range selector inputs), then we have the exception and chart is crashing.

```
highstock.src.js:55594 Uncaught TypeError: Cannot read property 'getBBox' of undefined
    at c.showDropdown (highstock.src.js:55594)
    at c.collapseButtons (highstock.src.js:55530)
    at c.handleCollision (highstock.src.js:55459)
    at c.alignElements (highstock.src.js:55243)
    at c.render (highstock.src.js:55056)
    at a.<anonymous> (highstock.src.js:55935)
    at highstock.src.js:2114
    at q (highstock.src.js:2111)
    at a.firstRender (highstock.src.js:30760)
```

#### Expected behaviour

I would expect that inputs just dive over the border (at least that shouldn't crash the chart)

#### Actual behaviour

When the width of screen is less than sum of width + margins of the range selector inputs, then the chart is crashing with exception.

https://jsfiddle.net/this_self/j3cdarz6/18/

Please set the next width:

```js
{
  ...,
  chart: {
    width: 300,
  }
}

```

#### Product version
Highstock 9.0.0

#### Affected browser(s)
All (but for firefox we have another exception):
```
Uncaught TypeError: c[this.currentButtonIndex()] is undefined
    showDropdown highstock.src.js:55594
    collapseButtons highstock.src.js:55530
    handleCollision highstock.src.js:55459
    alignElements highstock.src.js:55243
    render highstock.src.js:55056
    <anonymous> highstock.src.js:55935
    q highstock.src.js:2114
    q highstock.src.js:2111
    firstRender highstock.src.js:30760
    init highstock.src.js:29085
    q highstock.src.js:2121
    init highstock.src.js:28950
    getArgs highstock.src.js:28919
    a highstock.src.js:28895
    y highstock.src.js:56204
    <anonymous> (index):147

```

khlieng:
Fixed #15124, range selector with inputs wider than the plot and no buttons threw.

### DIFF
--- a/js/Extensions/RangeSelector.js
+++ b/js/Extensions/RangeSelector.js
@@ -1256,7 +1256,9 @@ var RangeSelector = /** @class */ (function () {
                 height: 0,
                 zIndex: inputsZIndex
             });
-            this.renderButtons();
+            if (this.buttonOptions.length) {
+                this.renderButtons();
+            }
             // First create a wrapper outside the container in order to make
             // the inputs work and make export correct
             if (container.parentNode) {
@@ -1723,7 +1725,7 @@ var RangeSelector = /** @class */ (function () {
                 hasActiveButton = true;
             }
         });
-        if (!hasActiveButton && buttons.length > 0) {
+        if (!hasActiveButton) {
             if (dropdown) {
                 dropdown.selectedIndex = 0;
             }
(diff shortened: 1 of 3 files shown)


KacperMadej:
Is there a reason to run `renderButtons` if there are no buttons to render? Maybe the check could be less nested? Like in here: https://github.com/highcharts/highcharts/blob/270b4c4a73110711b657b51e5f6c50f346883866/ts/Extensions/RangeSelector.ts#L1774 [Demo](https://jsfiddle.net/BlackLabel/e6aczjy8/) 

fyi @pawelfus

TorsteinHonsi (approved):
Thanks!

PR #15144, opened 2021-02-15
bugfix/15128-series-clip-false-stock

### LINKED ISSUE #15128: Line drawn incorrectly with series clipping disabled
#### Expected behaviour
On resizing the output pane, the line should be drawn correctly.

#### Actual behaviour
Line goes missing on resizing the output pane. Markers are drawn but line is not. 

[https://jsfiddle.net/vmfop43L/](https://jsfiddle.net/vmfop43L/)

Resize the output pane to reproduce the issue. Note that when `series.clip` is enabled, it works fine.

#### Product version
Highstock v9.0.0

#### Affected browser(s)
Chrome

khlieng:
Fixed #15128, parts of the series disappeared when redrawing a regular chart with `series.clip` set to `false` and stock loaded.

### DIFF
--- a/js/Core/Chart/StockChart.js
+++ b/js/Core/Chart/StockChart.js
@@ -682,7 +682,8 @@ addEvent(Series, 'render', function () {
     if (!(chart.is3d && chart.is3d()) &&
         !chart.polar &&
         this.xAxis &&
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (approved):
Thanks!

PR #15150, opened 2021-02-16
bugfix/15064-incorrect-default-multiples

### LINKED ISSUE #15064: X-axis custom intervals for charts xrange
hi~
Thank you for opening my issue.

It is expected to be able to customize the x-axis interval of the chart. At present, the labels of the x-axis of the graph generated by the default method will display incompletely and fold lines. When calculating according to the "day" method, the interval is always two days. It is expected that the setting of the interval can be customized here.

For example, as shown in the demo in the link, it is expected that the x-axis labels will not be folded, and all the last labels will be displayed. At present, the display form of "week" has not taken effect. How to achieve the above effect~

demo:https://jshare.com.cn/public/C0Gn1m

DJTechnoo:
Fixed #15064, corrected default multiples in the docs.


TorsteinHonsi (approved):
Thanks!

Issue #8603, opened 2018-07-10
Highstock: Dynamically added axes do not have same defaults


whawker:
#### Expected behaviour
Default axis options are applied to dynamically added axes (via `chart.addAxis()`)

#### Actual behaviour
Default options not applied, meaning default options like `ordinal` are not used

#### Live demo with steps to reproduce
https://jsfiddle.net/brojuf06/1/

In the demo above note how in the first chart, the xAxis is `ordinal: true` and has no gap. However in the second chart the default options are *not* applied, so a gap in the data is present.

Seems to be related to [this section](https://github.com/highcharts/highcharts/blob/master/js/parts/StockChart.js#L160-L214) of the code, where default are applied for `Highcharts.stockChart()` but no defaults are applied to `chart.addAxis`

#### Product version
Highstock 6.1.1

#### Affected browser(s)
All

whawker:
Judging by the inactivity, I guess this is a won't fix?

pawelfus:
Hi @whawker 

This ticket wasn't prioritized. At this moment please provide all of the options to `addAxis` you need to use. My apologies for the inconvenience!

PR #15164, opened 2021-02-17
bugfix/stock-default-axis-options

khlieng:
Fixed dynamically added axis defaults in stock, axes dynamically added to stock charts got different defaults than the ones added through the chart constructor.


TorsteinHonsi (changes requested):
Thanks for the PR!

We need some tests too. Does the `StockChart.createAxis` function actually run at all? As far as I can see, the `StockChart` class isn't used. The `stockChart` factory returns an instance of `Chart`. And `Highcharts.StockChart` is a reference to the `stockChart` factory.

@bre1470 FYI

khlieng:
So I guess most of the factory function should be moved into an init function now that theres a real class, and then change `Highcharts.StockChart` to the class?

ts/Core/Chart/StockChart.ts
```diff
@@ -1158,8 +1169,8 @@ function getDefaultAxisOptions(type: string, options: Highcharts.AxisOptions): H
  *
  * */
 
-(H as any).stockChart = stockChart;
-(H as any).StockChart = StockChart;
+H.stockChart = stockChart;
+H.StockChart = StockChart;
```
bre1470, line 1173: This should be moved to the master file for stock (modules/stock). All files except masters have to work without Highcharts namespace and automatic modifications to become compatible with ES6 tree shaking (async). And it is okay to have any casts in the master file. No need to define the namespace there.

ts/masters/modules/stock.src.ts
```diff
@@ -10,6 +10,7 @@
  * License: www.highcharts.com/license
  */
 'use strict';
+import Highcharts from '../highcharts.src.js';
```
bre1470, line 13: That needs to be taken from `'../../Core/Globals.js'` otherwise it bloats the product size.

TorsteinHonsi (approved):

pawelfus:
`xAxis.type` still fails: https://jsfiddle.net/BlackLabel/g7cfja6q/
Related ticket: https://github.com/highcharts/highcharts/issues/8603 

@khlieng

PR #15165, opened 2021-02-17
bugfix/15134-networkgraph-update-color

### LINKED ISSUE #15134: Networkgraph - series update
#### Expected behaviour

The networkgraph series should be changed after triggering the update event.

Workaround: https://jsfiddle.net/BlackLabel/8w6xgjkp/

#### Actual behaviour

The networkgraph series doesn't update. 

Demo: https://jsfiddle.net/BlackLabel/g2L7ateu/

It seems that the color update used to work in the previous version: https://jsfiddle.net/BlackLabel/szx8Lhmt/, however, the update of the `layoutAlgorithm` property didn't. 

#### Product version

Highcharts JS v9.0.0 (2021-02-02)

khlieng:
Fixed #15134, updating networkgraph series color did not work.

### DIFF
--- a/js/Core/Series/Series.js
+++ b/js/Core/Series/Series.js
@@ -2795,6 +2795,7 @@ var Series = /** @class */ (function () {
      * @fires Highcharts.Series#event:afterUpdate
      */
     Series.prototype.update = function (options, redraw) {
+        var _a;
         options = cleanRecursively(options, this.userOptions);
         fireEvent(this, 'update', { options: options });
         var series = this, chart = series.chart, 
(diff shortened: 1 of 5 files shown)


TorsteinHonsi (changes requested):
Thanks!

It would be better if we were able to avoid polluting the base series with functionality targeted at a relatively marginal series type like the network graph.

Presumably we can do this in a similar way to this code:

https://github.com/highcharts/highcharts/blob/94566e9dd40cf50d923bbe105895e230c385f5d4/ts/Series/Networkgraph/Networkgraph.ts#L749-L753

The `afterUpdate` event should work, but it may be that the `updatedData` event is also suited.

TorsteinHonsi (approved):

PR #15181, opened 2021-02-19
bugfix/14966-accessibility-null-points

### LINKED ISSUE #14966: Accessibility module included & point y value null leads to errors in browser console

K009:
Fixed #14966, selected null points generated: "Error: attribute y: Expected length, "NaN"".

### DIFF
--- a/js/Core/Series/Point.js
+++ b/js/Core/Series/Point.js
@@ -1136,7 +1136,8 @@ var Point = /** @class */ (function () {
             markerAttribs = series.markerAttribs(point, state);
         }
         // Apply hover styles to the existing point
-        if (point.graphic) {
+        // Prevent from dummy null points (#14966)
+        if (point.graphic && !point.hasDummyGraphic) {
             if (previousState) {
                 point.graphic.removeClass('highcharts-point-' + previousState);
             }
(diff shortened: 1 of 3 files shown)


TorsteinHonsi:
Thanks! To avoid creating unnecessary new charts in the test suite, can you see if any of the other tests in `unit-tests/accessibility/null-points/demo.js` can do this check? There are other charts with null points, you maybe you can just add a `.select()` and an assertion on one of them?

oysteinmoseng (approved):
Makes sense!

TorsteinHonsi (approved):

PR #15182, opened 2021-02-19
 bugfix/15178-theme-console-warning

### LINKED ISSUE #15178: AST trows a warning for a button with theme with states
#### Behaviour
If states are used in buttonTheme I get a warning in the console:
`Highcharts warning: Invalid attribute 'states' in config `

https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/stock/rangeselector/styling/

or 

https://api.highcharts.com/highcharts/chart.resetZoomButton.theme
Steps: zoom to show the button -> warning

#### Product version
v9.0.0+

#### Workaround
Ignore it - everything works fine anyway.

#### Internal notes
`states` end up in `normalState` (from the theme), but could be safely deleted before running `AST.filterUserAttributes(normalState)` because `normalState` is already a copy and `states` are provided to `SVGRenderer.prototype.button` as arguments and later used from them

DJTechnoo:
Fixed #15178, console warning when using `states` in `rangeSelector.buttonTheme` options.


TorsteinHonsi (changes requested):
Thanks for the PR! 

This PR got me started on a [refactoring of SVGAttributes](https://github.com/highcharts/highcharts/pull/15197) because `states` is not a valid part of SVGAttributes, and should be picked up by TS. 

So now, after you have merged in master, TS should complain that `states` does not exist on type SVGAttributes.

The fix is to delete the `states` before we create the button, like we do in Exporting:

https://github.com/highcharts/highcharts/blob/23414aed782fc2a7ecf5ce632673572a298298cd/ts/Extensions/Exporting.ts#L2142

Also, please fix the same problem for map navigation buttons: https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/maps/mapnavigation/button-theme/

ts/Core/Renderer/SVG/SVGRenderer.ts
```diff
@@ -37,6 +37,7 @@ import SVGLabel from './SVGLabel.js';
 import AST from '../HTML/AST.js';
 import TextBuilder from './TextBuilder.js';
 import U from '../../Utilities.js';
+import options from '../../../Accessibility/Options/Options';
```
TorsteinHonsi, line 40: This is not intended, right?

TorsteinHonsi (changes requested):

ts/Core/Renderer/SVG/SVGRenderer.ts
DJTechnoo: Woops, must've snuck in there from an auto import

TorsteinHonsi (approved):

PR #15183, opened 2021-02-19
ts/remove-js-folder 🎂

bre1470:
Removed versioning of JS folder.


goransle (approved):
Looks good 🍰 . I was curious if we had any gulp tasks that would have issues, but they seem to work as they should.

bre1470:
Indeed, the lint tasks had issues, when we migrated the product masters to TypeScript, but that is already part of the master branch.

PR #15184, opened 2021-02-19
docs/accessibility-docs

marvin19:
Added content to a11y documentation.


docs/accessibility/patterns-and-contrast.md
```diff
@@ -4,8 +4,33 @@ Patterns and contrast
 The default palette of Highcharts is designed with accessibility in mind, so that any two neighbor colors are tested for different types of color blindness. In addition to that, there are a few ways to increase contrast, both for the visually impaired or for grayscale prints, but also for the charts to be more readable in general.
```
oysteinmoseng, line 5: Suggested additional paragraph here:

Keep in mind that you should not rely on color alone to convey information in the chart if possible. Consider using data labels for points to show the values with numbers, and make sure readers of the chart can understand the data without relying on color.

docs/accessibility/tables.md
```diff
@@ -9,6 +9,23 @@ Including the [export-data](https://code.highcharts.com/modules/export-data.js)
 <script src="https://code.highcharts.com/modules/accessibility.js"></script>
 ```
 
+There is two different ways to view the data table. The first option is by clicking the menu and then "Show data table".
```
oysteinmoseng, line 12: "There are two"

oysteinmoseng (changes requested):
Awesome work, this will be a serious improvement! Just minor comments as per usual.

PR #15185, opened 2021-02-19
bugfix/13247-stock-empty-series-NaN

### LINKED ISSUE #13247: time.getTimezoneOffset called with NaN when using stockChart
#### Expected behaviour
The chart renders without any errors.

#### Actual behaviour
Rendering the chart throws an error because it got `timestamp` that is `NaN`.

https://codesandbox.io/s/happy-night-hoqp2

#### Product version
8.0.4

#### Affected browser(s)
Chrome

#### Comments

I am using `time.getTimezoneOffset` to avoid Moment.js. Instead, I am using `date-fns-tz`. But the conversion of the timestamp to the desired timezone fails, because `getTimezoneOffset` is called with `NaN` for an unknown reason.

- I assume it should be possible to display 2 different series (one type line, the other type column) on the same axis.
- The xAxis min, max, minRange settings seem to affect this bug as well (try to comment them, while type="column" is commented)
- I included the unminified highcharts source, it's easy to debug the internals like this. Make sure though you open the preview in a separate tab

khlieng:
Fixed #13247, `getTimezoneOffset` got called with `NaN` timestamp when rendering stock chart with at least one empty series.


ts/Extensions/RangeSelector.ts
```diff
@@ -1459,7 +1459,7 @@ class RangeSelector {
             );
         }
 
-        if (time && useUTC) {
+        if (time && useUTC && isNumber(date)) {
```
pawelfus, line 1462: Could you tell us (or leave a comment?) why `date` can be NaN?

pawelfus (commented):
Nice catch! In other words, when other series forced dataGrouping, a series without data caused this. One question though, see above

ts/Extensions/RangeSelector.ts
khlieng: Date.parse can return NaN

pawelfus: Ok, then what does this change do? If `date` can be NaN, then no matter what we do here, we will return `NaN` anyway. Maybe I'm missing something?

khlieng: It prevents calling `getTimezoneOffset` with a NaN timestamp, which may be a user-provided callback, which is what the issue was all about. We dont really need the offset if we have no date to apply it to anyway. Returning NaN is fine.

pawelfus: Thanks for the explanation! :) 

I will leave the decision to Torstein, but I think this might be unnecessary and may cause some issues in the future. If any similar issue (like the one above, with empty data series and dataGrouping) shows up, in the bug report we will have the title: "`getTimezoneOffset` not called" instead of "`getTimezoneOffset` called with NaN's" - not a bug difference. This only hides the real culprit, looks like a workaround 🤔

TorsteinHonsi (approved):

TorsteinHonsi:
All good, awaiting CI

PR #15186, opened 2021-02-19
bugfix/11687-scrollablePlotArea-drillUpButton-fixed

### LINKED ISSUE #11687: scrollablePlotArea and drilldown button position
When I click on the bar of the chart, back button to return to the previous level appears in the upper left corner, which if there are many elements disappears from the view when moving the 
horizontal scroll, I need that the button can be fixed in the same corner of the chart widget.

image

Please check my demo with the same bug when you click a bar in the chart = > 
 https://jsfiddle.net/x7w09u6j/1/

I'm working with the following versions of HC => "highcharts": "^7.0.3" &
"highcharts-angular": "^2.4.0" (angular 7 +.net core development)

This bug happens in Chrome, Safari in my case.

khlieng:
Fixed #11687, `drillUpButton` did not have a fixed position when using `scrollablePlotArea`.


TorsteinHonsi (approved):
Thanks!

Issue #14888, opened 2021-01-07
Highcharts: Current Price Indicator for multiple series


rogerwalt:
#### Expected behaviour
I expect the current price indicator to indicate the current price for every series.

#### Actual behaviour
When using the current price indicator from the stock-tools bar, the current price indicator is only displayed for one series.

#### Live demo with steps to reproduce
http://jsfiddle.net/rogerwa/5bvpgdfr/9/

1. Enable the current price indicator from the stock toolbar.
2. Notice that only one label and one red line is displayed instead of 3 each.

#### Product version
Highstock v8.2.2

#### Affected browser(s)
Chrome v87
Firefox v84

KacperMadej:
Hi @rogerwalt 

Thank you for reporting the problem.

The feature is not fully/clearly described in our documentation, so I see how there might be different expectations for this feature. IMHO the tool works as designed and only documentation should be improved.

Another option would be to either change how this stock tool works (to meet your requirements), add some option to have this work both ways, or add a new tool like in the workaround below.

@rogerwalt Please let me know how would like to see this issue resolved.

Stock tools are limited here by the price indicator which is limited by axis because it allows only a single crosshair label.

**Workaround**: define custom stock tool (docs: https://www.highcharts.com/docs/stock/stock-tools) that will work as needed. Demo: http://jsfiddle.net/BlackLabel/Ld6cb4uk/

I had to disable `compare: 'percent'` due to another issue #14879
I had to use multiple axes due to the axis crosshair label limit.

JustinPhung:
Hi @KacperMadej,

i am working together with @rogerwalt.
We could implement the current Price Indicator as suggest in the workaround.

That's why in my opinion this issue is resolved.

Thank you!

pawelfus:
We will improve logic for showing/hiding price indicators:

- button will toggle price indicators for all series (price series, volume series, and technical indicators)
- to solve a potential issue with multiple overlapping price indicators, we will resolve it by assigning the same color to the price indicator as the main series has (e.g. if the series has an orange color, the price indicator's line and the label will have an orange stroke and fill)

Issue #15029, opened 2021-02-02
StockTools: LastPrice button functionality improvements


pawelfus:
#### Expected behaviour
Use cases:
- initializing chart with `lastVisiblePrice.enabled` and `lastPrice.enabled` **disabled** (default): button should show "show" icon. The first click on the button should show labels and plot lines on price series (tech. indicators: to be decided, navigator series: don't). Next clicks toggle the price indicators
- initializing chart where at least one series has `lastVisiblePrice.enabled` and `lastPrice.enabled` enabled: the same behaviour as above
- initializing chart where all price series (excluding navigator series and tech. indicators) have both options enabled: button should show "hide" icon. The first click on the button should hide all last price indicators. Then every subsequent click toggles the price indicators

#### Actual behaviour
The button toggles the visibility of `lastPrice` and `lastVisiblePrice` independently. Initializing series with `lastPrice.enabled = false` and `lastVisiblePrice.enabled = true` means first click on button with switch `lastPrice.enabled` to true and `lastVisiblePrice.enabled` to false. This is strange behaviour, the end-user can not get rid of horizontal lines not matter what, and icons on the button are misleading (show/hide icon is based on `lastPrice.enabled` value).

#### Live demo with steps to reproduce
https://jsfiddle.net/BlackLabel/42jxyf6g/3/

#### Product version
v8+

PR #15195, opened 2021-02-22
bugfix/14888-currentPriceIndicator-all-series

karolkolodziej:
Added `currentPriceIndicator` and `lastPrice` for multiple series (also to indicators and volume), see #14888.

To do:
- [x] Update docs


ts/Stock/StockToolsBindings.ts
```diff
@@ -2047,8 +2047,8 @@ var stockToolsBindings: Record<string, Highcharts.NavigationBindingsOptionsObjec
             button: HTMLDOMElement
         ): void {
             var chart = this.chart,
-                series = chart.series[0],
-                options = series.options,
+                series = chart.series,
+                options = series[0].options,
```
karolkolodziej, line 2051: This is related to #15029 and will be fixed then.

samples/unit-tests/indicator-price-indicator/recalculations/demo.js
```diff
@@ -318,4 +318,84 @@ QUnit.test('The lastPrice color, #15074.', function (assert) {
...
+    assert.strictEqual(
+        chart.series[1].lastPrice.attr('stroke'),
+        chart.series[1].color,
+        'Volume lastPrice line should have color as series.'
+    );
+    assert.strictEqual(
+        chart.series[1].crossLabel.attr('fill'),
+        chart.series[1].color,
+        'Volume lastVisiblePrice label should have color as series.'
+    );
```
pawelfus, line 389: You test twice the same series, probably should be `series[2]`.

How about placing these tests in `forEach()`?

ts/Stock/StockToolsBindings.ts
```diff
@@ -2069,18 +2069,22 @@ var stockToolsBindings: Record<string, Highcharts.NavigationBindingsOptionsObjec
...
+                    // label
+                    lastVisiblePrice: {
+                        enabled: !lastVisiblePrice,
+                        label: {
+                            enabled: true,
+                            color: serie.options.lastPrice && serie.options.lastPrice.color || serie.color // #14888
+                        }
                     }
-                }
+                });
```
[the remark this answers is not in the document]
pawelfus: I think we should follow the change with docs (`By default, the line has the same color as the series.`), so the change should be done in PriceIndicator.ts`, not in `StockTools.ts`. This button should only toggle visibility, not perform any action related to the color.

samples/unit-tests/indicator-price-indicator/recalculations/demo.js
```diff
@@ -318,4 +318,84 @@ QUnit.test('The lastPrice color, #15074.', function (assert) {
...
+                height: '60%'
+            }, {
+                top: '65%',
+                height: '35%'
+            }],
+            stockTools: {
+                gui: {
+                    enabled: true,
+                    buttons: [
+                        "currentPriceIndicator"
```
pawelfus, line 335: Let's go with single quotes :)

pawelfus (changes requested):
See comments

raf18seb (commented):
Shouldn't the line follow the series' color? 

**edit:** it seems the problem occurs only with loaded `highcharts.css` styles. 
image

ts/Extensions/PriceIndication.ts
```diff
@@ -126,9 +127,7 @@ addEvent(Series, 'afterRender', function (): void {
         lastVisiblePrice = seriesOptions.lastVisiblePrice,
         lastPrice = seriesOptions.lastPrice;
 
-    if ((lastVisiblePrice || lastPrice) &&
-            seriesOptions.id !== 'highcharts-navigator-series') {
-
+    if ((lastVisiblePrice || lastPrice) && seriesOptions.id !== 'highcharts-navigator-series') {
```
raf18seb, line 130: Wouldn't it be more readable when written in multiple lines?
```js
if (
    (lastVisiblePrice || lastPrice) &&
    seriesOptions.id !== 'highcharts-navigator-series'
) {
```

samples/unit-tests/indicator-price-indicator/recalculations/demo.js
```diff
@@ -318,4 +318,68 @@ QUnit.test('The lastPrice color, #15074.', function (assert) {
...
+            }, {
+                top: '65%',
+                height: '35%'
+            }],
+            stockTools: {
+                gui: {
+                    enabled: true,
+                    buttons: [
+                        'currentPriceIndicator'
+                    ]
```
raf18seb, line 336: It can be written in only 1 line
```js
buttons: ['currentPriceIndicator']
```

samples/unit-tests/indicator-price-indicator/recalculations/demo.js
```diff
@@ -318,4 +318,68 @@ QUnit.test('The lastPrice color, #15074.', function (assert) {
...
+    chart.series.forEach(function (serie) {
+        if (serie.options.id !== 'highcharts-navigator-series') {
+            assert.strictEqual(
+                serie.lastPrice.attr('stroke'),
+                serie.color,
+                'Each series lastPrice line should have color as series.'
+            );
+            assert.strictEqual(
+                serie.crossLabel.attr('fill'),
+                serie.color,
```
raf18seb, line 369: The single form of "series" is "series. I've got my own crusade to correct it in the code wherever I can.

I am aware there was a discussion about it a while ago, but I still think that we should be consistent with a correct form and 95%+ of "series" cases, not "serie".

samples/unit-tests/indicator-price-indicator/recalculations/demo.js
```diff
@@ -318,4 +318,68 @@ QUnit.test('The lastPrice color, #15074.', function (assert) {
...
+                'Each series lastVisiblePrice label should have color as series.'
```
raf18seb, line 370: series'

ts/Stock/StockToolsBindings.ts
```diff
@@ -2069,19 +2069,22 @@ var stockToolsBindings: Record<string, Highcharts.NavigationBindingsOptionsObjec
...
-                lastPrice: {
-                    enabled: !lastPrice
-                },
-                // label
-                lastVisiblePrice: {
-                    enabled: !lastVisiblePrice,
-                    label: {
-                        enabled: true
+            series.forEach(function (serie): void {
+                serie.update({
```
raf18seb, line 2073: "serie" again

raf18seb (commented):
The code itself is perfect. I have only few polishing suggestions (not changes request, just suggestions). Sorry for being picky

pawellysy (approved):
Looks good, thanks

samples/unit-tests/indicator-price-indicator/recalculations/demo.js
karolkolodziej: You are right, we should be consistent with the names, so I will join your crusade and I will try to change it everywhere.

raf18seb (approved):
Perfect, thanks ;3

karolkolodziej:
The question raised by Rafał looks like a separate issue that involves CSS stylesheets.
Reported here: #15222

ts/Extensions/PriceIndication.ts
```diff
@@ -144,9 +146,12 @@ addEvent(Series, 'afterRender', function (): void {
             crop;
 
         if (lastPrice && lastPrice.enabled) {
-
             yAxis.crosshair = yAxis.options.crosshair = seriesOptions.lastPrice;
```
pawelfus, line 149: Is this line still necessary? Below, we do the merge, so maybe we can remove it? :)

ts/Core/Chart/StockChart.ts
```diff
@@ -692,6 +692,7 @@ addEvent(Axis, 'afterDrawCrosshair', function (
             crossLabel
                 .attr({
                     fill: options.backgroundColor ||
+                        point && point.color || // #14888
```
pawelfus, line 695: Q: Is this expected that only `crossLabel` has the same color as the last point? How about the line?

Check out:
- `series.colorUp`: https://jsfiddle.net/BlackLabel/ba1m7edk/3/
- `point.color`: https://jsfiddle.net/BlackLabel/ba1m7edk/2/ (less popular case, than the top one)

pawelfus (changes requested):

ts/Core/Chart/StockChart.ts
karolkolodziej: Let's stick for now with the code which set color for the line and label from the series.
Changed in the latest commit.

ts/Extensions/PriceIndication.ts
```diff
@@ -144,9 +146,14 @@ addEvent(Series, 'afterRender', function (): void {
...
 
+            if (!series.chart.styledMode) {
+                yAxis.crosshair = yAxis.options.crosshair = merge({
+                    color: series.color // default color from the series #14888
+                }, seriesOptions.lastPrice);
```
pawelfus, line 154: Can we simplify this a little? In general `merge`s are time-consuming, so perhaps this could be easier and faster? Something like this:
```js
yAxis.crosshair.color = yAxis.options.crosshair.color = seriesOptions.lastPrice.color || series.color
```

pawelfus (changes requested):

pawelfus (approved):
Thanks!

PR #15197, opened 2021-02-22
TS: Strict SVGAttributes

TorsteinHonsi:
No more rogue keys.


ts/Accessibility/AccessibilityComponent.ts
```diff
@@ -342,8 +342,8 @@ AccessibilityComponent.prototype = {
             }, attributes);
 
         Object.keys(attrs).forEach(function (prop: string): void {
-            if (attrs[prop] !== null) {
-                proxy.setAttribute(prop, attrs[prop]);
+            if ((attrs as any)[prop] !== null) {
+                proxy.setAttribute(prop, (attrs as any)[prop]);
```
bre1470, line 346: If you set the `attrs` variable to the type `(SVGAttributes&Record<string, any>)`, you do not have to make `any` casts. On the other hand I like to see the Record<string, any> type from all the place removed as well. So maybe we should introduce in the global namespace of `Core/Globals` a type alias `type AnyRecord = Record<string, any>`. So `attrs` should then get the type `(SVGAttributes&AnyRecord)`. What do you think about that idea? Would this work for all the other new `any` casts that pop up?

bre1470 (changes requested):
If a `any` cast is the solution, we need to do something about it.
- In general we can not make SVGAttributes a strict sealed interface, in case you try to accomplish this. I can always come along with a `ExtendedSVGAttributes` interface.
- As an alternativ to removing the generic key indexer, we can add all valid and invalid properties (split it, to make it easier to maintain). Example:
```ts
// valid attributes
interface SVGAttributes extends Record<string, any> {
    'stroke-width': string;
    tabindex: number;
}
// invalid attributes
interface SVGAttributes {
    borderRadius: never;
    strokeWidth: never;
}
```

ts/Accessibility/AccessibilityComponent.ts
TorsteinHonsi: It looks like in this case the problem was actually that `attrs` can be undefined.

Apart from that, I think most of those new `any`s have popped up in situations where we are looping objects. It looks like it can be resolved by using `keyof`. Should we go with that solution? Then we preserve the strict typing.
```ts
        Object.keys(attrs || {}).forEach(function (prop: keyof HTMLAttributes): void {
            if (attrs && attrs[prop] !== null) {
                groupDiv.setAttribute(prop, String(attrs[prop]));
            }
        });
```

bre1470 (approved):
We will leave it as it for now and fix `any` later.

PR #15200, opened 2021-02-23
bugfix/13858-stock-responsive-clip

### LINKED ISSUE #13858: Stock Line Chart broken with responsive options
#### Expected behaviour
When using the responsive feature the line chart should not be broken

#### Actual behaviour
if the container of the highstock-line-chart has a fixed height and you use the responsive feature
(to disable rangeSelector and legend for example) the Line in the Chart is broken.

http://jsfiddle.net/nickwinger/pk5rdq6g/5/

#### Product version

Highstock

#### Affected browser(s)
Chrome, Safari, latest

khlieng:
Fixed #13858, clipping was wrong after plot height had been changed by a responsive rule firing during chart load with stock loaded.


js/Core/Chart/StockChart.js
```diff
@@ -699,9 +699,11 @@ addEvent(Series, 'render', function () {
...
             }
         }
-        // First render, initial clip box
-        if (!this.clipBox && this.isDirty && !this.isDirtyData) {
-            this.clipBox = merge(chart.clipBox);
+        // First render, initial clip box. clipBox also needs to be updated if
+        // the series is rendered again before starting animating, in
+        // compliance with a responsive rule (#13858).
+        if (!chart.hasLoaded || (!this.clipBox && this.isDirty && !this.isDirtyData)) {
+            this.clipBox = this.clipBox || merge(chart.clipBox);
```
pawelfus, line 706: Is there any reason to use the original object, then the copy (if not defined)?

pawelfus (commented):
Again, a short question, sorry 😅

js/Core/Chart/StockChart.js
khlieng: Just to avoid the copy. The demos with visual differences dont even load the code that got changed by this PR, so not sure whats up there.

pawelfus: Just realized it's `this.clipBox` vs `chart.clipBox`, sorry!

Regarding the visual diff, perhaps we load all modules in CircleCi visual tests? Just like in Karma. @goransle might have more info :)

goransle: Yes, in fact the visual tests are a subset of the karma tests 🤓

khlieng: So I was kind of able to reproduce the staticscale one by loading stock, but its reversed from what the visual diff says, broken on master and fixed in this PR, strange.

Master: https://jsfiddle.net/qp18r2xb/
PR: https://jsfiddle.net/wzfvnp1t/12/

goransle: @khlieng you might want to run karma in debug mode to see what is happening there. 
From the highcharts root: `npx karma start ./test/karma-conf.js --debug --browsers Chrome --no-single-run --tests 'highcharts/studies/axis-staticscale'`

khlieng: Now the staticscale diff is working as intended, but im not able to reproduce the other one.

That demo also seems to be broken, the scrollable part doesnt work, and try this:
1. Open https://www.highcharts.com/samples/gantt/chart/scrollable-plotarea-vertical
2. Click "New product launch once"
3. Click "New offices" 3 times

TorsteinHonsi (changes requested):
Thanks! Visual tests need to be worked on.

js/Core/Chart/StockChart.js
pawelfus: Okaay, some magic happens here. If this is not already reported could you report it? (Gantt+scrollable+tree grid axis)

khlieng: #13884 looks similar, and it has an open and approved PR 🤔

khlieng: I added the fix from that PR here just to test and now the clipping in the diff is fine.

TorsteinHonsi: There are still two diffing samples? https://vrevs.highsoft.com/pr/15200/review

khlieng: The first one is working as intended, it now has the same clipping as a regular bar chart and the same clipping as that sample without stock loaded: https://www.highcharts.com/samples/highcharts/studies/axis-staticscale.

And the other one is the same 15px diff as in #13884, not sure why that happens, but the 600px width it has now kind of sounds more correct.

TorsteinHonsi (approved):

PR #15207, opened 2021-02-23
bugfix/13646-empty-pie-color

### LINKED ISSUE #13646: series.pie.color is not changing color
#### Expected behaviour
According to [docs](https://api.highcharts.com/highcharts/series.pie.color), the color of the border in an empty pie chart should be customizable. If I provide a valid hex code, the color of this border should change.

#### Actual behaviour
The border for an empty chart remains `#cccccc` regardless of what I provide this option.

This is the example included in docs: https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/plotoptions/pie-emptyseries/

The color `#c0ffee` should be a mint-like color, but you can see the border is not respecting the provided hex code.

#### Product version
Highcharts latest version (8.1.0)

#### Affected browser(s)
Reproduced on Mac: Chrome, FF, Safari

madepiet:
Fixed #13646, the `color` option did not take effect in empty pie series.

~~Fixed #13646 Color property in the empty pie.~~


TorsteinHonsi:
Thanks for the PR! 

In general, we should avoid `userOptions`, because ideally the class in question should have to distinguish between what options are default and which are user options. So I did a rewrite to try to avoid the case where we set `options.color = null` in the first place. Let's see how the tests go.

TorsteinHonsi (approved):

PR #15208, opened 2021-02-23
bugfix/15029-LastPrice-stockTools-functionality

### LINKED ISSUE #15029: StockTools: LastPrice button functionality improvements

karolkolodziej:
Fixed #15029, improved `currentPriceIndicator` button functionality.

Todo: 
- [x] Change button icon
- [x] Add test for button change
- [x] Merge master after merging  #15195  and #15226  it into master


raf18seb (changes requested):

ts/Stock/StockToolsGui.ts
```diff
@@ -1223,6 +1227,34 @@ class Toolbar {
...
+        series.forEach(function (serie): void {
+            const lastVisiblePrice = serie.lastVisiblePrice && serie.lastVisiblePrice.enabled,
+                lastPrice = serie.lastPrice && serie.lastPrice.enabled;
+
+            if (lastVisiblePrice || lastPrice) {
+                priceIndicatorEnabled = 'atLeastOneEnabled';
+            }
+
+            if (!lastVisiblePrice && !lastPrice) {
+                priceIndicatorEnabled = 'bothDisabled';
```
pawellysy, line 1253: Having the boolean variable stored as a string seems a bit strange for me. A better way would be creating some kind of enum, where all options are available, which would prevent us from making some typos. If there are only 2 possibilities, the obvious choice seems `boolean`.

pawellysy (changes requested):
Overall good, see comments :)

pawellysy (approved):
Looks Good!

ts/Stock/StockToolsGui.ts
```diff
@@ -1223,6 +1227,34 @@ class Toolbar {
...
+                lastPrice = serie.lastPrice && serie.lastPrice.enabled;
+
+            if (lastVisiblePrice || lastPrice) {
+                priceIndicatorEnabled = true;
+            }
+
+            if (!lastVisiblePrice && !lastPrice) {
+                priceIndicatorEnabled = false;
+            }
+        });
```
pawelfus, line 1255: This `forEach` in fact is testing only the last series. No matter what we decide for the first series, the last one can define `priceIndicatorEnabled` according to its own internal options.

For example, the first series has `lastPrice` enabled and it's setting `priceIndicatorEnabled` to true. Then the last series has both options disabled, resulting in `priceIndicatorEnabled = false`. If we reverse the series order, then we will always receive `priceIndicatorEnabled=true`, no matter what was set in the first series.

ts/Stock/StockToolsBindings.ts
```diff
@@ -2047,41 +2047,50 @@ var stockToolsBindings: Record<string, Highcharts.NavigationBindingsOptionsObjec
...
             var chart = this.chart,
-                series = chart.series[0],
-                options = series.options,
-                lastVisiblePrice = (
-                    options.lastVisiblePrice &&
-                    options.lastVisiblePrice.enabled
-                ),
-                lastPrice = options.lastPrice && options.lastPrice.enabled,
+                series = chart.series,
+                priceIndicatorEnabled = chart.stockTools && chart.stockTools.priceIndicatorEnabled,
```
pawelfus, line 2051: Question: Is it safe to store this as a property? Can something change the button state, so we should check it on demand (`isPriceIndicatorEnabled(..)`)? For example on series add/remove?

pawelfus (changes requested):
See comments below

ts/Stock/StockToolsBindings.ts
pawelfus: I still have a small problem with storing `priceIndicatorEnabled` variable in `stockTools`. It is a bit bloating the properties, that are related to the tools in general (HTML references, options, chart references etc.). A property specific to just one button.

Can we go for now with the function call in both places? Upon `chart.render` event and here, when clicking on a button. So below would be changed from:

```js
if (priceIndicatorEnabled) {
  ...
}
```
to:
```js
if (chart.stockTools.isPriceIndicatorEnabled(series)) {
 ...
}
```

Or even better, store `isPriceIndicatorEnabled` in `bindingsUtils`, and use it:
```js
if (bindingsUtils.isPriceIndicatorEnabled(series)) {
 ...
}
```

What do you think?

pawelfus (changes requested):

ts/Stock/StockToolsBindings.ts
```diff
@@ -2047,41 +2047,40 @@ var stockToolsBindings: Record<string, Highcharts.NavigationBindingsOptionsObjec
...
+                            lastPrice: {
+                                enabled: true
+                            },
+                            lastVisiblePrice: {
+                                enabled: true,
+                                label: {
+                                    enabled: true
+                                }
+                            }
+                        }, false);
```
pawelfus, line 2078: Q: Would that make sense to store `bindingsUtils.isPriceIndicatorEnabled(chart.series)` in a local variable? Then we could simplify this long if-else to:

```js
const priceIndicatorEnabled = bindingsUtils.isPriceIndicatorEnabled(chart.series);

if (gui && gui.guiEnabled && chart.stockTools) {
  series.forEach(function (series): void {
    series.update({
      lastPrice: { enabled: !priceIndicatorEnabled },
      lastVisiblePrice: { enabled: !priceIndicatorEnabled, label: { enabled: true } }
    }, false);
  });
}
```

pawelfus (changes requested):
Just a minor clean up :)

ts/Extensions/Annotations/NavigationBindings.ts
```diff
@@ -1486,6 +1504,24 @@ setOptions({
         }
     }
 });
+addEvent(H.Chart, 'render', function (): void {
+    const chart = this,
+        stockTools = chart.stockTools,
+        button = chart.stockTools &&
+            chart.stockTools.toolbar &&
+            chart.stockTools.toolbar.querySelector('.highcharts-current-price-indicator');
```
raf18seb, line 1512: You can use only `stockTools` variable instead of `chart.stockTools`

ts/Extensions/Annotations/NavigationBindings.ts
```diff
@@ -1486,6 +1504,24 @@ setOptions({
         }
     }
 });
+addEvent(H.Chart, 'render', function (): void {
```
raf18seb, line 1507: It would be nice to comment why this event is added or at least add a comment with a related ticket number only. Might be useful in the future

ts/Extensions/Annotations/NavigationBindings.ts
```diff
@@ -1486,6 +1504,24 @@ setOptions({
...
+            chart.stockTools.toolbar &&
+            chart.stockTools.toolbar.querySelector('.highcharts-current-price-indicator');
+
+    // Change the initial button background.
+    if (stockTools && chart.options.series && button) {
+        if (bindingsUtils.isPriceIndicatorEnabled(chart.series)) {
+            (button as any).firstChild.style['background-image'] =
+            'url("' + stockTools.getIconsURL() + 'current-price-hide.svg")';
+        } else {
+            (button as any).firstChild.style['background-image'] =
```
raf18seb, line 1520: Nothing big, but you could declare the variable like this
```js
button = chart.stockTools &&
            chart.stockTools.toolbar &&
            (chart.stockTools.toolbar.querySelector('.highcharts-current-price-indicator') as any).firstChild;
```
and use a shorter version
```js
button.style['background-image'] = ...
```

ts/Stock/StockToolsBindings.ts
```diff
@@ -2047,41 +2047,40 @@ var stockToolsBindings: Record<string, Highcharts.NavigationBindingsOptionsObjec
...
-                gui: Highcharts.Toolbar = chart.stockTools as any,
-                iconsURL = gui.getIconsURL();
-
-            if (gui && gui.guiEnabled) {
-                if (lastPrice) {
-                    (button.firstChild as any).style['background-image'] =
-                        'url("' + iconsURL +
-                        'current-price-show.svg")';
+                series = chart.series,
+                gui: Highcharts.Toolbar = chart.stockTools as any;
```
raf18seb, line 2051: Hmm, I'm not a TypeScript ninja, but is everything ok with this line?

ts/Stock/StockToolsBindings.ts
```diff
@@ -2047,41 +2047,40 @@ var stockToolsBindings: Record<string, Highcharts.NavigationBindingsOptionsObjec
...
+
+            if (gui && gui.guiEnabled && chart.stockTools) {
+                // Reset all and disable both indicators
+                if (bindingsUtils.isPriceIndicatorEnabled(chart.series)) {
```
raf18seb, line 2055: You have already declared `var series = chart.series` so use a variable instead

raf18seb (changes requested):

raf18seb (approved):
Thanks 👍

pawelfus (approved):
Approved! 👍

TorsteinHonsi (approved):
Thanks!

Issue #12268, opened 2019-10-24
Stock tools GUI doesn't work on the second axis


Basts:
#### Expected behaviour

Stock tools GUI should work for both axes. 

#### Actual behaviour

It is impossible to add any stock tools on the second axis. 

#### Live demo with steps to reproduce

[Official GUI demo](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/stock/demo/stock-tools-gui/)

#### Product version
Highcharts JS v7.2.0 (2019-09-03)

raf18seb:
Hi, thanks for reporting.

There is an **addClipPaths()** method that adds clipPath on annotations that are outside of axes.
You can simply disable this functionality:
```Highcharts.Annotation.prototype.addClipPaths = function() {};```

Demo: https://jsfiddle.net/BlackLabel/hydLu05f

mubeenellahi1:
@raf18seb Thank you for your suggestion it was super helpful.

But this partially solves the issue. What I am able to do is:
Create a flag annotation on the first axis and them move it around to all the axis.
The functionality required is that I want to create separate annotations for separate y axis. For now clicking on the second y axis does not create an annotation.

raf18seb:
Hi @mubeenellahi1 

Unfortunately, the yAxis is hardcoded and, for this moment, adding annotations to other axes is impossible.
As @Basts reported above, it's an enhancement and maybe will be added to Highcharts in the future.
I strongly advise reporting this on our UserVoice channel here: https://highcharts.uservoice.com - it might speed up the development process if other users want this feature too.

For now, you can add annotations only to the first axis and, if you want to move them over the second axis, disable clipPaths as suggested above.

raghavakrishnab:
Hi 
i m  getting error  **addClipPaths doesn't exist on type 'Annotation'**  when i tried to disable clippaths as per your suggestion.
Highcharts.Annotation.prototype.addClipPaths = function() {};

Please assist.

raf18seb:
I am not seeing any error in the above [demo](https://jsfiddle.net/BlackLabel/hydLu05f) (latest Chrome, Windows 10). Can you provide more info/steps so I can reproduce your issue?

raghavakrishnab:
Hi,
In the same [demo](https://jsfiddle.net/BlackLabel/hydLu05f) (latest Chrome, Windows 10).

[...]

PR #15209, opened 2021-02-23
enhancement/12268-stock-tools-multiple-axes

raf18seb:
Improved #12268, added possibility to draw Stock Tools on multiple axes.

~~Initially, this is a draft PR. The commit has been made with `--no-verify` flag because of some TS errors - it is being investigated now.~~

~~Potential code review ideas will be discussed later.~~

~~FYI @goransle and @bre1470 - this is the branch we were discussing my TS issue on.~~


bre1470:
The test are passing for me on this branch. Maybe ts-node behaves different on a Windows environment. 🤔
Does `npx gulp test` passes for you?

raf18seb:
> Does npx gulp test passes for you?

No, it throwns another error:
```
TypeError: test\karma-conf.js: Emit skipped
    at getOutput (C:\Users\Rafal\Desktop\repos\highcharts\node_modules\ts-node\src\index.ts:755:17)
    at Object.compile (C:\Users\Rafal\Desktop\repos\highcharts\node_modules\ts-node\src\index.ts:968:32)
```

Goran is going to take a look at it on Windows.

raf18seb:
The `getAssignedAxis` method checks whether (x, y) coordinates are within the axes `min` and `max` extremes. I am opened to discuss other approaches how we can valid coordinates.

ts/Extensions/Annotations/NavigationBindings.ts
```diff
@@ -1214,22 +1261,31 @@ setOptions({
...
                     this: NavigationBindings,
                     e: PointerEvent
-                ): Annotation {
+                ): Annotation|void {
                     var coords = this.chart.pointer.getCoordinates(e),
+                        coordsX = this.getAssignedAxis(coords, 'x'),
+                        coordsY = this.getAssignedAxis(coords, 'y'),
                         navigation = this.chart.options.navigation;
 
+                    // Exit if clicked out of axes area
```
karolkolodziej, line 1270: I would add the same comment everywhere you have that `if`. Not everyone is looking through the whole file, very often we concentrate on one method and I think we should have similar comments.

karolkolodziej (changes requested):

ts/Extensions/Annotations/Annotations.ts
```diff
@@ -179,6 +179,8 @@ declare global {
             markerStart?: string;
             point?: (string|AnnotationMockPointOptionsObject);
             points?: Array<(string|AnnotationMockPointOptionsObject)>;
+            xAxis: number;
```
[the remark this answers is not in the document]
raf18seb: It's index. I didn't change the name because it was named like this previously. We can consider changing it, but there are so many places when an axis' index is called this way that I think it should be a part of potential Stock Tools refactoring in the future.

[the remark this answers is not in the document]
pawelfus: It's the same as with option `series.xAxis` or `series.yAxis`. It's an index, even when the name does not suggest it.

However, do we have `shapeOptions.xAxis` or `shapes[index].xAxis`? Only `shapes[index].point` or `points` have this option, according to the API: https://api.highcharts.com/highstock/annotations.shapes.point.xAxis (commented with `(1)` the same places)

And the format is `number | string | null`, could you double-check this?

ts/Extensions/Annotations/NavigationBindings.ts
```diff
@@ -1214,22 +1261,31 @@ setOptions({
...
+
                     return this.chart.addAnnotation(
                         merge(
                             {
                                 langKey: 'circle',
                                 type: 'basicAnnotation',
                                 shapes: [{
                                     type: 'circle',
+                                    xAxis: coordsX.axis.options.index,
+                                    yAxis: coordsY.axis.options.index,
```
pawelfus, line 1283: See my comment above, could you double-check this? I don't recall this option

ts/Extensions/Annotations/NavigationBindings.ts
```diff
@@ -195,17 +195,24 @@ const bindingsUtils = {
...
-            width = coords.xAxis[0].value - options.point.x,
-            height = options.point.y - coords.yAxis[0].value;
-
-        annotation.update({
-            typeOptions: {
-                background: {
-                    width: chart.inverted ? height : width,
-                    height: chart.inverted ? width : height
+            coordsX = NavigationBindings.prototype.getAssignedAxis(coords, 'x'),
+            coordsY = NavigationBindings.prototype.getAssignedAxis(coords, 'y'),
```
pawelfus, line 199: Any particular reason to use `NavigationBindings.prototype.getAssignedAxis` over `chart.navigationBindings.getAssignedAxis` in this one annotation?

I'm fine with static methods, but then assign it to `NavigationBindings` instead of `NavigationBindings.prototype`.

ts/Stock/StockToolsBindings.ts
```diff
@@ -385,27 +395,44 @@ bindingsUtils.updateHeight = function (
...
 bindingsUtils.attractToPoint = function (
     e: PointerEvent,
     chart: Chart
-): Highcharts.NavigationBindingsAttractionObject {
+): Highcharts.NavigationBindingsAttractionObject | void {
     var coords = chart.pointer.getCoordinates(e),
-        x = coords.xAxis[0].value,
-        y = coords.yAxis[0].value,
+        coordsX = NavigationBindings.prototype.getAssignedAxis(coords, 'x'),
+        coordsY = NavigationBindings.prototype.getAssignedAxis(coords, 'y'),
```
pawelfus, line 420: `NavigationBindings.prototype.getAssignedAxis` vs `chart.navigationBindings.getAssignedAxis`

pawelfus (commented):
Amazing! 🎉 
More questions than changes, see below

ts/Stock/StockToolsBindings.ts
```diff
@@ -750,26 +778,39 @@ var stockToolsBindings: Record<string, Highcharts.NavigationBindingsOptionsObjec
...
-                            }, {
-                                x: coords.xAxis[0].value,
-                                y: coords.yAxis[0].value
-                            }]
-                        }
-                    },
-                    navigation.annotationsOptions,
-                    (navigation.bindings as any).segment.annotationsOptions
-                );
+                options;
```
pawellysy, line 789: is it necessary to declare the options here, instead of after the if statement, where the options var is assigned?
Additionally, why not `const`?

pawellysy (approved):
Overall good, one question regarding the declaration of the variable.

ts/Extensions/Annotations/Annotations.ts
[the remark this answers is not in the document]
TorsteinHonsi: > And the format is number | string | null, could you double-check this?

When `series.xAxis` or `series.yAxis` is a string, it refers to the `id` of the axis. For consistency, I think we should recreate that feature for the Annotations too. 

Also, I think we can simplify a bit and disallow `null`. I think 0 has the same effect, right?

TorsteinHonsi (changes requested):
Please resolve comments and merge in master

ts/Extensions/Annotations/Annotations.ts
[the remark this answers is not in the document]
raf18seb: Nice catch, @pawelfus 
I corrected TS definitions and now I'm taking axis index from shapes[index] point. Also, the API description should be corrected or even removed (the user doesn't need to define https://api.highcharts.com/highstock/annotations.shapes.point.xAxis because the click finds a proper axis automatically.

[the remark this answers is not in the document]
raf18seb: Nice catch, @pawelfus
I corrected TS definitions and now I'm taking axis index from shapes[index] point. Also, the API description should be corrected or even removed (the user doesn't need to define https://api.highcharts.com/highstock/annotations.shapes.point.xAxis because the click finds a proper axis automatically.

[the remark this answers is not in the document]
raf18seb: I think this option can be removed from API anyway because now the user doesn't need to define on which axis he is adding an annotation - new functionality finds the proper axis automatically.

When it comes to TS definitions, I think the Annotation's MockPoint's `xAxis` or `yAxis` props are only `number | AxisType` type now. I removed `string` and `null` from TS definitions.

Are you OK with these changes? https://github.com/highcharts/highcharts/pull/15209/commits/fd88d5ef63d43b515446d12f904475c0faed7b23

ts/Stock/StockToolsBindings.ts
raf18seb: It's a habit related to `var` being hoisted. I didn't change `var` to `let` or `const` because this ticket is not about refactoring. If I wrote it from scratch, I would definitively write it your way.

raf18seb:
Do you know where those CI errors come from? Something wrong with merge?

bre1470:
The annotation type tests break:
```
test/typescript/modules/annotations.ts(37,22): error TS2769: No overload matches this call.
  Overload 1 of 2, '(options: Options, callback?: ChartCallbackFunction | undefined): Chart', gave the following error.
    Type '"container"' has no properties in common with type 'Options'.
```
Run `npx gulp lint-dts` to test local.

Also visual tests fail:
```
HeadlessChrome 89.0.4389 (Linux 0.0.0) Highcharts **********/annotations-advanced/measure FAILED
	Died on test #1     at samples/**********/annotations-advanced/measure/demo.js.preprocessed.js:2:15: Cannot read property 'toPixels' of undefined
	TypeError: Cannot read property 'toPixels' of undefined
	    at Measure.addCrosshairs (code/modules/annotations-advanced.src.js:5347:38)
	    at Measure.addShapes (code/modules/annotations-advanced.src.js:5319:22)
	    at Measure.Annotation.init (code/modules/annotations-advanced.src.js:2593:22)
	    at Measure.init (code/modules/annotations-advanced.src.js:5193:43)
	    at Measure.Annotation (code/modules/annotations-advanced.src.js:2582:22)
	    at new Measure (code/modules/annotations-advanced.src.js:5181:31)
	    at Chart.initAnnotation (code/modules/annotations-advanced.src.js:3550:34)
	    at code/modules/annotations-advanced.src.js:3630:44
	    at Chart.<anonymous> (code/modules/annotations-advanced.src.js:3624:39)
	HeadlessChrome 89.0.4389 (Linux 0.0.0): Executed 57 of 828 (1 FAILED) (skipped 8) (0 secs / -836235 mins 52.162 secs)
HeadlessChrome 89.0.4389 (Linux 0.0.0) Highcharts **********/annotations-advanced/pitchfork FAILED
	Died on test #1     at samples/**********/annotations-advanced/pitchfork/demo.js.preprocessed.js:2:15: Cannot read property 'left' of null
	TypeError: Cannot read property 'left' of null
	    at Function.InfinityLine.findEdgePoint (code/modules/annotations-advanced.src.js:4501:38)
	    at Pitchfork.middleLineEdgePoint (code/modules/annotations-advanced.src.js:4850:37)
	    at ControllablePath.point (code/modules/annotations-advanced.src.js:1040:56)
	    at ControllablePath.linkPoints (code/modules/annotations-advanced.src.js:1061:34)
	    at ControllablePath.init (code/modules/annotations-advanced.src.js:940:22)
	    at new ControllablePath (code/modules/annotations-advanced.src.js:1519:22)
	    at Pitchfork.Annotation.initShape (code/modules/annotations-advanced.src.js:2846:29)
	    at Pitchfork.addLines (code/modules/annotations-advanced.src.js:4871:22)
	    at Pitchfork.addShapes (code/modules/annotations-advanced.src.js:4867:22)
	    at Pitchfork.Annotation.init (code/modules/annotations-advanced.src.js:2593:22)
```

ts/Extensions/Annotations/Annotations.ts
```diff
@@ -1342,17 +1342,6 @@ merge<Annotation>(
...
 
-                /**
-                 * This number defines which xAxis the point is connected to.
-                 * It refers to either the axis id or the index of the axis in
-                 * the xAxis array. If the option is not configured or the axis
-                 * is not found the point's x coordinate refers to the chart
-                 * pixels.
-                 *
-                 * @type      {number|string|null}
-                 * @apioption annotations.labels.point.xAxis
```
pawelfus, line 1353: `labels[index].point.xAxis` is supported: https://api.highcharts.com/highstock/annotations.labels.point.xAxis

ts/Extensions/Annotations/Types/Measure.ts
```diff
@@ -822,18 +822,6 @@ Measure.prototype.defaultOptions = merge(
...
-             * It refers to either the axis id or the index of the axis
-             * in the xAxis array.
-             */
-            xAxis: 0,
-            /**
-             * This number defines which yAxis the point is connected to.
-             * It refers to either the axis id or the index of the axis
-             * in the yAxis array.
-             */
-            yAxis: 0,
```
pawelfus, line 836: `measure.typeOptions.xAxis` is supported: https://api.highcharts.com/highstock/annotations.measure.typeOptions.xAxis

ts/Extensions/Annotations/Types/CrookedLine.ts
```diff
@@ -156,19 +156,6 @@ CrookedLine.prototype.defaultOptions = merge(
...
-             * in the xAxis array.
-             */
-            xAxis: 0,
-            /**
-             * This number defines which yAxis the point is connected to.
-             * It refers to either the axis id or the index of the axis
-             * in the xAxis array.
-             */
-            yAxis: 0,
-
```
pawelfus, line 171: I think `typeOptions` does support `xAxis` and `yAxis`: https://api.highcharts.com/highstock/annotations.crookedLine.typeOptions.xAxis

ts/Extensions/Annotations/NavigationBindings.ts
```diff
@@ -969,6 +976,46 @@ class NavigationBindings {
...
+     *        All the chart's axes with a current pointer's axis value.
+     *
+     * @param {string} xOrY
+     *        'x' or 'y' indicating axis.
+     *
+     * @return {Highcharts.PointerAxisCoordinateObject}
+     *         Object with a first found axis and its value that pointer
+     *         is currently pointing.
+     */
+    public getAssignedAxis(
```
pawelfus, line 995: Can we move this function to `bindingUtils`?

pawelfus (changes requested):
Just a minor polishing

ts/Extensions/Annotations/Types/Measure.ts
raf18seb: Right, totally forgot that annotations still can be drawn in the chart's config...

samples/unit-tests/stock-tools/bindings/demo.js
```diff
@@ -398,3 +398,142 @@ QUnit.test('Bindings general tests', function (assert) {
...
+            coordsY;
+
+        // The yAxes overlap - y coord on both of them
+        coords = getCoordinates({
+            chartX: chart.yAxis[1].left + offset,
+            chartY: chart.yAxis[1].top + offset
+        });
+
+        coordsX = getAssignedAxis(coords, 'x');
+        coordsY = getAssignedAxis(coords, 'y');
```
pawellysy, line 438: Why you split declaration and assigning of the values to those variables? Can't you do assign them right away?

ts/Extensions/Annotations/NavigationBindings.ts
```diff
@@ -228,6 +235,46 @@ const bindingsUtils = {
...
+     * @param {string} xOrY
+     *        'x' or 'y' indicating axis.
+     *
+     * @return {Highcharts.PointerAxisCoordinateObject}
+     *         Object with a first found axis and its value that pointer
+     *         is currently pointing.
+     */
+    getAssignedAxis(
+        coords: Highcharts.PointerAxisCoordinatesObject,
+        xOrY: ('x'|'y')
```
pawellysy, line 257: I find the `xOrY` parameter name a bit odd. Couldn't it be something like `isXAxis: boolean` like in Axis class?

pawellysy (changes requested):
see comments, apart from that general question: Why are you splitting declaration and assignment of the variables? in some examples it just makes additional variable, which in case of early return statement call is redundant.

ts/Extensions/Annotations/NavigationBindings.ts
raf18seb: `axis.isXAxis` would not be useful in this case, because I am not operating on a specific axis, but on an object like this:
```js
coords = {
    xAxis: [{axis: axis, value: number}, ...],
    yAxis: [{axis: axis, value: number}, ...]
}
```

However, when I'm looking at this code after few days, I see that I don't need to loop through both properties `xAxis` and `yAxis` in `objectEach` there, because I am returning only one of them and I know which one from the beginning. I'll take a closer look at it tomorrow and probably improve.

raf18seb:
@TorsteinHonsi please hold merging, I'll improve something here.

pawelfus (approved):
Approved!

raf18seb:
@TorsteinHonsi ready to merge if no other changes requests.

karolkolodziej (approved):

pawellysy (approved):
Looks Good! Nothing to add :)

ts/Extensions/Annotations/MockPoint.ts
```diff
@@ -224,11 +225,12 @@ class MockPoint {
...
         point: Highcharts.AnnotationPointType
     ): Highcharts.AnnotationMockPointOptionsObject {
+
         return {
             x: point.x as any,
             y: point.y as any,
-            xAxis: point.series.xAxis,
-            yAxis: point.series.yAxis
+            xAxis: point.series.xAxis as AxisType,
+            yAxis: point.series.yAxis as AxisType
```
TorsteinHonsi, line 233: I don't think we should do the cast here... We don't _know_ that the series has an axis. What happens if the series is a pie series? Instead of casting, we should have a type check or allow the `AnnottationMockPointOptionsObject.xAxis` and `yAxis` to be undefined.

ts/Extensions/Annotations/NavigationBindings.ts
```diff
@@ -228,6 +235,35 @@ const bindingsUtils = {
...
+     * @return {Highcharts.PointerAxisCoordinateObject}
+     *         Object with a first found axis and its value that pointer
+     *         is currently pointing.
+     */
+    getAssignedAxis(
+        coords: Array<Highcharts.PointerAxisCoordinateObject>
+    ): Highcharts.PointerAxisCoordinateObject {
+        return coords.filter(function (coord): boolean {
+            const axisMin = coord.axis.min as number,
+                axisMax = coord.axis.max as number,
```
TorsteinHonsi, line 257: In this case, as the previous comment, we cannot cast `axis.min` and `axis.max` to `number`, because they aren't necessarily numbers. Instead we need to code for the case that they are not numbers.

In practice, we can test this by setting `data: []` in the given unit test.

ts/Extensions/Annotations/NavigationBindings.ts
```diff
@@ -1250,9 +1293,12 @@ setOptions({
...
                         annotation: Annotation
                     ): void {
-                        var point = annotation.options.shapes[0].point,
-                            x = this.chart.xAxis[0].toPixels((point as any).x),
-                            y = this.chart.yAxis[0].toPixels((point as any).y),
+                        var mockPointOpts = annotation.options.shapes[0]
+                                .point as Highcharts.AnnotationMockPointOptionsObject,
+                            x = this.chart.xAxis[mockPointOpts.xAxis as number]
+                                .toPixels(mockPointOpts.x),
+                            y = this.chart.yAxis[mockPointOpts.yAxis as number]
```
TorsteinHonsi, line 1300: As above, avoid casting and instead do type checking.

ts/Extensions/Annotations/NavigationBindings.ts
```diff
@@ -1304,23 +1360,23 @@ setOptions({
...
-                                        yAxis: 0,
+                                        xAxis: coordsX.axis.options.index,
+                                        yAxis: coordsY.axis.options.index,
                                         x: x,
                                         y: y
                                     }, {
-                                        xAxis: 0,
-                                        yAxis: 0,
+                                        xAxis: coordsX.axis.options.index,
+                                        yAxis: coordsY.axis.options.index,
```
TorsteinHonsi, line 1379: This block can be shortened considerably.

```ts
const xAxis = coordsX.axis.options.index;
const yAxis = coordsY.axis.options.index;
...
points: [
    { xAxis, yAxis, x, y },
    { xAxis, yAxis, x, y },
...
```

ts/Extensions/Annotations/Types/Pitchfork.ts
```diff
@@ -124,8 +125,8 @@ class Pitchfork extends InfinityLine {
         return {
             x: ((points[1].x as any) + (points[2].x as any)) / 2,
             y: ((points[1].y as any) + (points[2].y as any)) / 2,
-            xAxis: points[0].series.xAxis,
-            yAxis: points[0].series.yAxis
+            xAxis: points[0].series.xAxis as AxisType,
+            yAxis: points[0].series.yAxis as AxisType
```
TorsteinHonsi, line 129: As above, avoid casting and instead do type checking.

ts/Extensions/Annotations/Types/VerticalLine.ts
```diff
@@ -69,7 +70,7 @@ class VerticalLine extends Annotation {
 
         return {
             x: point.x as any,
-            xAxis: point.series.xAxis,
+            xAxis: point.series.xAxis as AxisType,
```
TorsteinHonsi, line 73: As above, avoid casting and instead do type checking.

ts/Extensions/Annotations/Types/VerticalLine.ts
```diff
@@ -94,7 +95,7 @@ class VerticalLine extends Annotation {
...
```
TorsteinHonsi, line 98: As above, avoid casting and instead do type checking.

ts/Stock/StockToolsBindings.ts
```diff
@@ -385,27 +395,43 @@ bindingsUtils.updateHeight = function (
...
 bindingsUtils.attractToPoint = function (
     e: PointerEvent,
     chart: Chart
-): Highcharts.NavigationBindingsAttractionObject {
+): Highcharts.NavigationBindingsAttractionObject | void {
     var coords = chart.pointer.getCoordinates(e),
-        x = coords.xAxis[0].value,
-        y = coords.yAxis[0].value,
+        coordsX = chart.navigationBindings?.utils.getAssignedAxis(coords.xAxis),
+        coordsY = chart.navigationBindings?.utils.getAssignedAxis(coords.yAxis),
```
TorsteinHonsi, line 419: Better do one check for `chart.navigationBindings` then assign coordsX and coordsY. Optional chaining is quite bulky when transpiled to ES5, and here it's twice over.

ts/Stock/StockToolsBindings.ts
```diff
@@ -419,8 +445,8 @@ bindingsUtils.attractToPoint = function (
         y: (closestPoint as any).y,
         below: y < (closestPoint as any).y,
         series: (closestPoint as any).series,
-        xAxis: (closestPoint as any).series.xAxis.index || 0,
-        yAxis: (closestPoint as any).series.yAxis.index || 0
+        xAxis: (closestPoint as any).series.xAxis.options.index || 0,
+        yAxis: (closestPoint as any).series.yAxis.options.index || 0
```
TorsteinHonsi, line 449: I know this is not new to this PR, but what if `closestPoint` is undefined? That case should probably be handled. Imagine that the series has `data: []`.

TorsteinHonsi (changes requested):
Thanks, great work! I added a bunch of comments, but they are all basically the same request: To avoid treating types of `(SomeType|undefined)` like `SomeType`, but also code for the case that it can be undefined.

TorsteinHonsi (changes requested):
Thanks for the updates! Please resolve conflicts before we merge.

TorsteinHonsi (approved):

PR #15215, opened 2021-02-24
bugfix/15075-stock-tools-yaxis-top

### LINKED ISSUE #15075: Some Stock Tools annotations (e.g. infinity line or pitchfork) aren't working with yAxis.top
The Pitchfork tool, Infinity Line, Vertical Line and similar aren't working properly when `yAxis.top` is defined.

https://jsfiddle.net/BlackLabel/4toy0fa8/

#### Product version
Highcharts Stock 9.0.0

#### Affected browser(s)
All

raf18seb:
Fixed #15075, some Stock Tools annotations were being drawn incorrectly when `yAxis.top` was set.

To do:
- [x] tests


ts/Extensions/Annotations/Types/VerticalLine.ts
```diff
@@ -58,31 +59,34 @@ class VerticalLine extends Annotation {
...
+            inverted = chart.inverted,
             point = annotation.points[0],
-            xy = MockPoint.pointToPixels(point, true),
-            y = xy.y,
-            offset = annotation.options.typeOptions.label.offset;
-
-        if (annotation.chart.inverted) {
-            y = xy.x;
-        }
+            left = pick(point.series.yAxis?.left, 0),
```
karolkolodziej, line 65: I know that optional chaining is convenient, however, it generates a lot of nasty variables in JS.
Could you change that to `left = pick(point.series.yAxis && point.series.yAxis.left, 0),`

ts/Extensions/Annotations/Types/VerticalLine.ts
```diff
@@ -58,31 +59,34 @@ class VerticalLine extends Annotation {
...
+            top = pick(point.series.yAxis?.top, 0),
+            offset = annotation.options.typeOptions.label.offset,
+            y = MockPoint.pointToPixels(point, true)[inverted ? 'x' : 'y'];
 
         return {
             x: point.x as any,
             xAxis: point.series.xAxis,
-            y: y + offset
+            y: y + offset +
+                (inverted ? (left - chart.plotLeft) : (top - chart.plotTop))
```
karolkolodziej, line 74: Personally, I wouldn't split that into two lines, but it's up to You ;)

ts/Extensions/Annotations/Types/VerticalLine.ts
```diff
@@ -91,7 +95,8 @@ class VerticalLine extends Annotation {
         return {
             x: point.x as any,
             xAxis: point.series.xAxis,
-            y: y + yOffset
+            y: y + yOffset +
+                (inverted ? (left - chart.plotLeft) : (top - chart.plotTop))
```
karolkolodziej, line 99: Same here ;)

samples/unit-tests/stock-tools/bindings/demo.js
```diff
@@ -398,3 +398,65 @@ QUnit.test('Bindings general tests', function (assert) {
...
+        {
+            chartX: point.plotX,
+            chartY: point.plotY + yAxis.top
+        }
+    );
+
+    let annotationBBox = chart.annotations[0].graphic.getBBox();
+
+    assert.strictEqual(
+        annotationBBox.y + annotationBBox.height + 10,
```
karolkolodziej, line 432: Could you please explain from where this `10` came?
Is that some kind of predefined offset?

karolkolodziej (changes requested):

ts/Extensions/Annotations/Types/VerticalLine.ts
raf18seb: We've already had a conversation about it a year ago and agreed that one optional chaining check is not that bad. The problem is getting bigger with more checks than one. However, I agree, we can get rid of them totally.

ts/Extensions/Annotations/Types/VerticalLine.ts
raf18seb: I like to keep 80 characters line length :) in this case it's readable enough so I would leave it as it is.

samples/unit-tests/stock-tools/bindings/demo.js
raf18seb: I thought that the assert description ```'Annotation\'s element should be placed 10px from the second point.'``` will be enough comment. Some annotations are placed 10px from a point. I haven't searched the code where exactly those 10 pixels are set, but it's a standard position. I can add an additional comment.

karolkolodziej: The test description is pretty self-explanatory and checks that condition perfectly.
I was curious if there is an option to replace `10` with something like (it's just a guess) `annotation.offset`

raf18seb: I'll check

pawelfus (approved):
Looks good, thanks!

pawellysy (approved):
approved!

TorsteinHonsi (changes requested):
Thanks! Please fix the merge conflict, apart from that it looks good.

samples/unit-tests/stock-tools/bindings/demo.js
raf18seb: You were close, thanks :) `typeOptions.yOffset`
https://github.com/highcharts/highcharts/blob/695395b8bd374a70e14a7863cc3dd34b8a0e1acb/ts/Extensions/Annotations/Types/VerticalLine.ts#L189

karolkolodziej (approved):

TorsteinHonsi (approved):

PR #15223, opened 2021-02-25
bugfix/15217-centerInCategory-bar-tooltip

### LINKED ISSUE #15217: Bar chart with centerInCategory enabled - tooltips display in the "non centered" position
#### Expected behaviour
after setting series.bar.centerInCategory to true, the tooltip will display right next to the bar.

#### Actual behaviour
the tooltip displays where it would if the bars weren't centered.

[https://jsfiddle.net/seaberry/5d2trz90/13/](https://jsfiddle.net/seaberry/5d2trz90/13/)

This is the demo from the series.bar.centerInCategory section of the docs. The only difference is that the chart type is "bar" instead of "column".

You'll also notice that the data appears to be in reverse order, but setting xAxis.reverse to false fixes this. 

#### Product version
Highcharts 8.2.2

#### Affected browser(s)
I tried this on Chrome 88 and Firefox 85

khlieng:
Fixed #15217, bar chart tooltip position was wrong with `centerInCategory` enabled.


TorsteinHonsi (approved):
Thanks!

PR #15226, opened 2021-02-25
bugfix/14192-disable-stock-tools-gui-buttons

### LINKED ISSUE #14192: forEach console error
Hello,

I'm having some issues with the console error. 
I recreated the issue in this js fiddle https://jsfiddle.net/40khrmuj/1/

You can replicate the issue by doing these steps: 
1. Unselect items from the highcharts-legend-box
2. Try to add Vertical labels or Flags to the chart space

When I'm trying to add the label I'm getting the error. Is there any way to disable the specific GUI tab if there is no series shown? 

Thanks!

pawellysy:
Fixed #14192, Disabled stock tools GUI buttons, if all series are invisible.

- [x] TODO: Add tests

Implemented automated disabling of the stock tools buttons if the plotting area is empty (when none series is visible). See the #14192. 

Added new option to some stock tools bindings, to determine, if the button should be auto-disabled. e.g.
```js
stockToolsBindings: {
    fullScreen: {
         alwaysEnabled: true/false/undefined (undefined = auto)
    }
}
```

Demo: https://jsfiddle.net/BlackLabel/zetv2gor/4/

Steps:
- hide both series
- drawing tools' buttons are disabled, others didn't change (for example FullScreen)


ts/Extensions/Annotations/NavigationBindings.ts
```diff
@@ -1486,6 +1487,51 @@ setOptions({
         }
     }
 });
+addEvent(H.Chart, 'render', function (): void {
+    var chart = this,
+        navigationBindings = chart.navigationBindings,
+        disabledClassName = 'highcharts-disabled-btn';
+
+    if (chart && navigationBindings) {
+        // check if the buttons should be enabled/disabled based on
```
karolkolodziej, line 1496: When the comment occupies the whole line please treat it as a sentence- start with a capital letter, ends with a dot.
`// My comment sentence.`

Comments at the end of the code could be treated as words.
` if (chart.stockTools && chart.stockTools.toolbar) { // some comment`

ts/Stock/StockToolsBindings.ts
```diff
@@ -2185,11 +2186,12 @@ var stockToolsBindings: Record<string, Highcharts.NavigationBindingsOptionsObjec
      *
      * @type    {Highcharts.NavigationBindingsOptionsObject}
      * @product highstock
-     * @default {"className": "highcharts-save-chart", "init": function() {}}
+     * @default {"className": "highcharts-save-chart", alwaysEnabled: true, "init": function() {}}
```
karolkolodziej, line 2189: The `alwaysEnabled` default property is missing in  `fullScreen`.

karolkolodziej (changes requested):
It looks like the fix doesn't fully cover that issue. Altho disabling the GUI works well.
Using the demo from the issue and that PR, I was able to get the same error in a simple case.
1. Disable one series.
2. Try to add a flag -> error in console.

Additionally, please see the comments in the code.

raf18seb (changes requested):

ts/Stock/StockToolsGui.ts
```diff
@@ -1489,6 +1489,10 @@ class Toolbar {
             // main button in first level og GUI
             mainNavButton = (buttonWrapper as any).parentNode.parentNode;
 
+        // if the button is disabled, don't do anything
+        if (buttonWrapperClass.indexOf('highcharts-disabled-btn') > -1) {
```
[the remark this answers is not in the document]
pawellysy: Less code, and seems more accurate in my opinion. I can change if you like.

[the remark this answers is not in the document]
KacperMadej: psst... [here](https://dev.to/werninator/a-neat-little-trick-with-javascripts-indexof-4dj5)

ts/Extensions/Annotations/NavigationBindings.ts
```diff
@@ -1486,6 +1487,52 @@ setOptions({
...
+            navigationBindings.boundClassNames,
+            function (value: Highcharts.NavigationBindingsOptionsObject, key: string): void {
+                // Get the HTML element coresponding to the
+                // className taken from StockToolsBindings.
+
+                if (chart.stockTools && chart.stockTools.toolbar) {
+                    const buttonNode = chart.stockTools.toolbar.querySelectorAll('.' + key);
+
+                    if (buttonNode) {
+                        if (!buttonsEnabled && !value.alwaysEnabled) {
```
pawelfus, line 1516: Can we move `value.alwaysEnabled` to the line 1512? Whole checking DOM, and classnames operations are unnecessary when buttons should be always enabled.

pawelfus (changes requested):
Looks good! Minor changes.

Question: How hard it will be to disable the hover effect when buttons are disabled? To match RangeSelector's buttons design:

hover

pawelfus:
@TorsteinHonsi 
We are open to the API option name changes. See API description & demo in the OP comment.

Should we also show this in the changelog as a new feature or a bugfix? It was reported & classified as a bug, but fixing a bug is a side effect of the general improvements in StockTools (this PR and #15209 PR => both together fix #14192)

raf18seb (approved):
Nothing more to add from my side, thanks.

samples/unit-tests/stock-tools/gui/demo.js
```diff
@@ -61,3 +61,115 @@ QUnit.test('Stocktools GUI', function (assert) {
     );
     */
 });
+
+QUnit.test('Disabling and Enabling stock Tools buttons, when series are invisible, #14192', function (assert) {
```
raf18seb, line 65: Mixing lowercase and uppercase letters

samples/unit-tests/stock-tools/gui/demo.js
```diff
@@ -61,3 +61,115 @@ QUnit.test('Stocktools GUI', function (assert) {
...
+    toolsContainer.className += 'tools-container';
+    var button = document.createElement('button');
+    button.className += 'dummy-button';
+    button.innerHTML = 'dummy button';
+
+    toolsContainer.appendChild(button);
+
+    document.getElementById('container').parentNode.insertBefore(toolsContainer, document.getElementById('container'));
+
+    var chart = Highcharts.stockChart('container', {
```
raf18seb, line 79: Not a mistake, but it's a better practice to keep the `var` declarations on top of the function scope. You could also use `const` and `let`.

samples/unit-tests/stock-tools/gui/demo.js
```diff
@@ -61,3 +61,115 @@ QUnit.test('Stocktools GUI', function (assert) {
...
+            bindingsClassName: 'tools-container'
+        }
+    });
+
+    var controller = new TestController(chart);
+    toolsContainer.style.position = 'absolute';
+    toolsContainer.style['z-index'] = 99999;
+
+    controller.click(10, 10);
+    assert.equal(wasInitCalled, false, 'Init function should not be executed, when there is no series.');
```
raf18seb, line 103: Any reason why this assert is in one line? It would be more readable divided into multiple lines (like in the other tests).

samples/unit-tests/stock-tools/gui/demo.js
```diff
@@ -61,3 +61,115 @@ QUnit.test('Stocktools GUI', function (assert) {
...
+        data: [
+            [0, 1, 2, 3, 4],
+            [1, 1, 2, 3, 4],
+            [2, 1, 2, 3, 4],
+            [3, 1, 2, 3, 4],
+            [4, 1, 2, 3, 4],
+            [5, 1, 2, 3, 4],
+            [5, 1, 2, 3, 4]
+        ]
+    });
```
raf18seb, line 148: You are adding a default `line` series with `ohlc` data. I would stick with `line` series and simplify data.

samples/unit-tests/stock-tools/gui/demo.js
```diff
@@ -61,3 +61,115 @@ QUnit.test('Stocktools GUI', function (assert) {
...
+    controller.click(10, 10);
+    assert.equal(wasInitCalled, true, 'Init function should not be called, when series are visible.');
+
+    chart.series[0].remove();
+    wasInitCalled = false;
+    controller.click(10, 10);
+    assert.equal(wasInitCalled, false, 'Init function should not be called, after deleting the series.');
+
+    chart.addSeries({
+        id: 'aapl',
```
raf18seb, line 138: Is this `id` necessary?

samples/unit-tests/stock-tools/gui/demo.js
```diff
@@ -61,3 +61,115 @@ QUnit.test('Stocktools GUI', function (assert) {
...
+                dummyButton: {
+                    className: 'dummy-button',
+                    alwaysEnabled: true,
+                    init: function () {
+                        wasInitCalled = true;
+                    }
+                }
+            }
+        }
+    });
```
raf18seb, line 162: When using multiple updates (`chart.addSeries`, `chart.update` and `series.setVisible`), set `redraw` flag to `false`.

ts/Extensions/Annotations/NavigationBindings.ts
```diff
@@ -1486,6 +1487,60 @@ setOptions({
...
+            if (!series.options.isInternal && series.visible) {
+                buttonsEnabled = true;
+            }
+        });
+
+        objectEach(
+            navigationBindings.boundClassNames,
+            function (value: Highcharts.NavigationBindingsOptionsObject, key: string): void {
+                // Get the HTML element coresponding to the
+                // className taken from StockToolsBindings.
```
raf18seb, line 1510: Is this a proper place for this comment?

raf18seb (changes requested):

pawelfus (approved):
Nothing to add 👍

samples/unit-tests/stock-tools/gui/demo.js
TorsteinHonsi: FYI, we have `max-len` at 80 in our ESLint settings, but with an exception for strings. That's why ESLint didn't complain about this. But best practice is still 80.

TorsteinHonsi:
> We are open to the API option name changes.

@pawelfus So the new option is `alwaysEnabled`? I would say that the word _always_ may be redundant, because `enabled: true` means always. Then `enabled: false` means never, and `enabled: undefined` would mean auto. This is in accordance with how other `enabled` options work.

Is there a difference between enabling the button and showing it in the first place? What option do we have for removing the button completely?

pawelfus:
> So the new option is `alwaysEnabled`? I would say that the word always may be redundant, because enabled: true means always. Then `enabled: false` means never, and `enabled: undefined` would mean auto. This is in accordance with how other `enabled` options work.
> 
> Is there a difference between enabling the button and showing it in the first place? What option do we have for removing the button completely?

@TorsteinHonsi 
Enabling/disabling buttons completely has the same API as exporting menu: [`stockTools.gui.buttons`](https://api.highcharts.com/highstock/stockTools.gui.buttons) vs [`exporting.buttons.contextButton.menuItems`](https://api.highcharts.com/highstock/exporting.buttons.contextButton.menuItems). Here, we want to implement auto-disable feature for a button. Something similar to the disabled state in RangeSelector buttons. Some buttons should be automatically disabled (e.g. adding annotations when all series are hidden) while others should always be enabled (e.g. Fullscreen).

The best option would be `disabled: true/false/undefined` but everywhere we use `enabled` so I believe it's a bad idea to introduce this name for an option.

TorsteinHonsi (changes requested):
For the records, we decided the new option should be called

`noDataState: 'normal'|'disabled'|undefined`

Then we stay consistent with our current button state names (normal, hover, select, disabled). We also have the three outcomes of the original suggestion (always, never, auto).

pawellysy:
@TorsteinHonsi just to make sure, that I understand correctly, 'normal' state makes the buttons stay active when no series is visible (like current alwaysEnabled: true), 'disabled' -> disables the buttons when there is no data, and undefined behaves like disabled?

samples/unit-tests/stock-tools/gui/demo.js
raf18seb: @pawellysy when dividing into multiple lines, it's better to do it this way:
```js
assert.ok(
    true,
    'some text'
);
```
instead of this way:
```js
assert.ok(
    true,
    'some text');
```

TorsteinHonsi:
@pawellysy 
* `noDataState: 'normal'` should replace `alwaysEnabled: true`
* `noDataState: 'disabled'` should replace `alwaysEnabled: false`
* `noDataState: undefined` should replace `alwaysEnabled: undefined`

TorsteinHonsi (approved):

css/stocktools/gui.scss
```diff
@@ -262,4 +262,8 @@ $button-hover-color: #e6ebf5;
   width: 50%;
   height: 20px;
   float: left;
-}
\ No newline at end of file
+}
+
+li.highcharts-disabled-btn:hover, .highcharts-disabled-btn .highcharts-submenu-item-arrow:hover {
+  background-color: #f7f7f7;
```
TorsteinHonsi, line 268: Use the `$button-background-color` variable instead. Makes theming easier more predictable.

TorsteinHonsi (changes requested):
In addition to the inline comment, please resolve the merge conflict. Other than that, everything ready for merge. Nice work!

TorsteinHonsi (approved):

TorsteinHonsi:
Thanks! Pending approvals from @raf18seb and @karolkolodziej .

karolkolodziej (approved):

raf18seb (approved):
👍

PR #15233, opened 2021-02-26
bugfix/15218-liveRedraw-default

### LINKED ISSUE #15218: performance is different regarding how scrollbar is defined
#### Expected behaviour
Performance should be the same whether in which place scrollbar is defined.

#### Actual behaviour
Performance is different depending on how scrollbar is defined. When it's defined on xAxis it works like _liveRedraw_ property was set to true, and when it's defined directly on the chart it seems that _liveRedraw_ property is set to false. According to documentation in both ways, the property should be defined the same and work the same.

https://jsfiddle.net/BlackLabel/doLyrhe8/

#### Product version
Highcharts JS v9.0.1 (2021-02-25)

khlieng:
Fixed #15218, `liveRedraw` defaulting to true in modern browsers did not work when scrollbar was set in chart options instead of axis options.


pawelfus (approved):
Thanks!

One question though, does it make sense to refactor this, and instead of repeating the same code over and over again, just put this in one function?

TorsteinHonsi (approved):
Thanks, looks good!

PR #15236, opened 2021-03-01
bugfix/15222-css-overwriting-crosshair-color

### LINKED ISSUE #15222: CSS properties overwrite the lastPrice color
#### Expected behaviour
When `styledMode: false`, the CSS properties should not overwrite the `lastPrice` element attributes.

#### Actual behaviour
When you import the stylesheet from `https://code.highcharts.com/css/highcharts.css`, CSS class `.highcharts-crosshair-thin` overwrite the `lastPrice` line color.

http://jsfiddle.net/BlackLabel/sukvqo08/
Comment/remove the stylesheet to see the difference.

#### Product version

#### Affected browser(s)
All

DJTechnoo:
Fixed #15222, CSS color class not added to `lastPrice` cross-hair.


ts/Core/Axis/Axis.ts
```diff
@@ -7780,6 +7775,12 @@ class Axis {
                             dashstyle: (options as any).dashStyle
                         });
                     }
+                } else {
+                    graphic.addClass( // #15222
+                        'highcharts-crosshair highcharts-crosshair-' +
+                        (categorized ? 'category ' : 'thin ') +
+                        (options as any).className
```
karolkolodziej, line 7782: When `className` does not exist, the class name looks strange: `class="highcharts-crosshair highcharts-crosshair-thin undefined"`.
I would add some condition here like `options.className ? options.className : ''` to prevent that.

pawelfus (changes requested):
I'm not sure if this is a bug (#15222). No matter if styled mode or not, classnames are available anyway. Check simple demo with CSS for the background: https://jsfiddle.net/BlackLabel/me10k4wu/

To make this work the same way in styled and non-styled modes (so color does match the series color), I would consider adding related color-class (`.highcharts-color-{n}`), to match the series color-class. But I believe this should be part of #15195 🤔

pawelfus:
Agreed on: Let's try `.highcharts-color-{n}` solution, after #15195 merge :)

pawelfus (changes requested):
Looks good, thank you! 
Requesting a test :)

pawelfus (approved):
Thanks! 👍

TorsteinHonsi (approved):
Thanks!

PR #15248, opened 2021-03-03
bugfix/button-align-update-size

khlieng:
Fixed reset zoom and drillup button alignment after plot size changes, button alignment did not update when the plot size changed due to hiding or showing elements such as legend or title.

### DIFF
--- a/ts/Core/Chart/Chart.ts
+++ b/ts/Core/Chart/Chart.ts
@@ -3725,8 +3725,8 @@ class Chart {
 }
 
 addEvent(Chart, 'afterSetChartSize', function (e: { skipAxes: boolean }): void {
-    if (!e.skipAxes && this.resetZoomButton) {
-        this.resetZoomButton.align();
+    if (!e.skipAxes) {
+        this.renderer.align();
     }
 });
 
--- a/ts/Core/Chart/Chart.ts
+++ b/ts/Core/Chart/Chart.ts
@@ -1922,7 +1922,7 @@ class Chart {
                 axis.setAxisSize();
                 axis.setAxisTranslation();
             });
-            this.renderer.alignElements();
+            renderer.alignElements();
         }
 
         fireEvent(chart, 'afterSetChartSize', { skipAxes: skipAxes });
--- a/ts/Core/Chart/Chart.ts
+++ b/ts/Core/Chart/Chart.ts
@@ -1922,6 +1922,7 @@ class Chart {
                 axis.setAxisSize();
                 axis.setAxisTranslation();
             });
+            this.renderer.alignElements();
         }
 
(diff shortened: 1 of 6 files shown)


TorsteinHonsi (changes requested):
We have an array, `renderer.alignedObject`, that should take care of re-aligning all items after the renderer size changes. See https://jsfiddle.net/highcharts/j3tmvbLd/.

I would expect that to work on these two buttons too. Can you investigate that? Maybe we should run alignment not only on renderer size change but also on plot size change?

khlieng:
Yes, it does work on these buttons, but the renderer doesnt change size in this case. Ill look into running it on plot size changes also.

ts/Core/Chart/Chart.ts
```diff
@@ -3724,6 +3724,12 @@ class Chart {
 
 }
 
+addEvent(Chart, 'afterSetChartSize', function (e: { skipAxes: boolean }): void {
+    if (!e.skipAxes) {
+        this.renderer.align();
```
TorsteinHonsi, line 3729: Shouldn't we just move this call inside the `setChartSize` function instead?

https://github.com/highcharts/highcharts/blob/4905235f0d4a6ffd93668478b48434fb182621b0/ts/Core/Chart/Chart.ts#L1921-L1924

```ts
        if (!skipAxes) {
            chart.axes.forEach(function (axis: Highcharts.Axis): void {
                axis.setAxisSize();
                axis.setAxisTranslation();
            });            
            renderer.align();
        }
```

ts/Core/Renderer/SVG/SVGRenderer.ts
```diff
@@ -2474,6 +2471,20 @@ class SVGRenderer {
...
     }
+
+    /**
+     * Re-align all aligned elements.
+     *
+     * @private
+     * @function Highcharts.SVGRenderer#align
+     * @return {void}
+     */
+    public align(): void {
```
TorsteinHonsi, line 2482: I think we should call this function `alignElements` to avoid (cognitive) confusion with `SVGElement.align` and the `align` property of SVGAttributes.

ts/Core/Renderer/SVG/SVGRenderer.ts
```diff
@@ -2474,6 +2471,20 @@ class SVGRenderer {
...
+        let i = this.alignedObjects.length;
+        while (i--) {
```
TorsteinHonsi, line 2484: Since speed is not critical here (we're only doing a handful of elements), maybe we should go with the more compact `forEach` loop.

TorsteinHonsi (changes requested):

TorsteinHonsi (approved):

PR #15256, opened 2021-03-03
enhancement/15252-map-gantt-class

### LINKED ISSUE #15252: Refactor: Proper classes for mapChart and ganttChart
Since https://github.com/highcharts/highcharts/pull/15164, `StockChart` is a proper class inheriting from `Chart`, instead of just a factory. To avoid inconsistency we need to do the same change with map and gantt charts.

khlieng:
Added map and gantt classes, closes #15252.


ts/Core/Chart/GanttChart.ts
```diff
@@ -43,8 +37,129 @@ const {
...
+     * @function Highcharts.GanttChart#init
+     *
+     * @param {Highcharts.Options} userOptions
+     *        Custom options.
+     *
+     * @param {Function} [callback]
+     *        Function to run when the chart has loaded and and all external
+     *        images are loaded.
+     *
+     * @return {void}
```
bre1470, line 54: If these should be public classes and functions, we need on top of the class pattern some doclets similar to
```ts
/**
 * Gantt-optimized chart. Use {@link Highcharts.Chart|Chart} for common charts.
 *
 * @requires modules/gantt
 *
 * @class
 * @name Highcharts.GanttChart
 * @extends Highcharts.Chart
 */
```

Otherwise the class doclets lack a `@private` tag one line above `@function`.

ts/masters/modules/map.src.ts
```diff
@@ -21,4 +22,8 @@ import '../../Series/MapPoint/MapPointSeries.js';
 import '../../Series/MapBubble/MapBubbleSeries.js';
 import '../../Series/Heatmap/HeatmapSeries.js';
 import '../../Extensions/GeoJSON.js';
-import '../../Maps/Map.js';
+import M from '../../Core/Chart/MapChart.js';
```
bre1470, line 25: I would prefer if the imports in masters are more descriptive, because later on all imports are named, and a single letter will not work then anymore. I suggest in this case here `MapChart`.

ts/Core/Chart/MapChart.ts
```diff
@@ -0,0 +1,218 @@
...
+
+    return SVGRenderer.prototype.pathToSegments(arr);
+}
+
+const mapModule = {
+    maps,
+    mapChart,
+    MapChart,
+    splitPath
+};
```
bre1470, line 210: I have an idea, that is similar to static functions/props. You could create a namespace with the name `MapChart` and put all the functions/props into this namespace as exports. Then TypeScript merges it with the class and you just need to export MapChart class. Functions/props can be accessed like `MapChart.maps` etc.

bre1470 (changes requested):
In general we should be careful in making doclets public. Also we should consider creating the helper functions/props into a separate Utilities file. I remember that I wrote something else yesterday, but the implementations here make me realize, that it is better if all class modules export the class as default for future compatibility and consistency. See suggestion for namespace workaround, if (Gantt|Map|Stock)Utilities files are not an option.

samples/unit-tests/highcharts/constructors/demo.js
```diff
@@ -190,7 +190,7 @@ QUnit.test('Highmaps', function (assert) {
         }
     });
 
-    chart = new Highcharts.Map({
+    chart = new Highcharts.MapChart({
```
TorsteinHonsi, line 193: I think we need an alias for this constructor. `new Highcharts.Map` was the documented way of constucting until v5 or so, to we risk pulling the carpet under some old implementations with this change. It should be quite simple to set it in the master file, `Highcharts.Map = MapChart`.

TorsteinHonsi (changes requested):

samples/unit-tests/highcharts/constructors/demo.js
khlieng: Its already there.

TorsteinHonsi: Okay, but then we need to keep the old constructor in some of the tests so that we have coverage.

samples/maps/series/latlon-transform/demo.js
```diff
@@ -3,7 +3,7 @@ var chart,
     position;
 
 // Initiate the chart
-chart = Highcharts.Map({
+chart = new Highcharts.Map({
```
bre1470, line 6: I believe that this requires, that `Highcharts.Map = Highcharts.MapChart.mapChart`.

ts/masters/modules/map.src.ts
```diff
@@ -21,4 +22,8 @@ import '../../Series/MapPoint/MapPointSeries.js';
 import '../../Series/MapBubble/MapBubbleSeries.js';
 import '../../Series/Heatmap/HeatmapSeries.js';
 import '../../Extensions/GeoJSON.js';
-import '../../Maps/Map.js';
+import MapChart from '../../Core/Chart/MapChart.js';
+
+(Highcharts as any).MapChart = (Highcharts as any).Map = MapChart;
+(Highcharts as any).mapChart = MapChart.mapChart;
```
bre1470, line 28: Here the `Highcharts.Map` assignment has to change, I believe. Like `Highcharts.Map` the factory function `Highcharts.mapChart` does not care, if it is called with the `new` keyword or not.

bre1470 (commented):
Looks good overall! Though I believe it is not 100% backwards compatible.

samples/maps/series/latlon-transform/demo.js
bre1470: Sorry, to be more specific: IMHO we should not change this test.

bre1470 (approved):

samples/maps/series/latlon-transform/demo.js
khlieng: Since this is example code intended for people to look at, maybe it should just use `mapChart`?

bre1470: Agreed. Maybe you can find some other test for `Highcharts.Map` without instantiation.

khlieng: Yeah I added one to the constructor unit test.

TorsteinHonsi:
> Though I believe it is not 100% backwards compatible.

@bre1470 Now you're scaring me. I see that we can't call `Highcharts.Map` without the `new` keyword, but I think that's fine. Do you know other caveats?

bre1470:
> > Though I believe it is not 100% backwards compatible.
> 
> @bre1470 Now you're scaring me. I see that we can't call `Highcharts.Map` without the `new` keyword, but I think that's fine. Do you know other caveats?

I just recall that I messed with one of the constructors some time ago by removing the ability to use it without `new` keyword, and we had failing customer solutions. That was when we had for some time one bugfix release per week. 

So as we had this in a one of our Map demos, we should support it. And it is easy to do, because the return value of a function is always used as the instance when the function has been called with the `new` keyword.

khlieng:
> > Though I believe it is not 100% backwards compatible.
> 
> @bre1470 Now you're scaring me. I see that we can't call `Highcharts.Map` without the `new` keyword, but I think that's fine. Do you know other caveats?

I'll just mention that we can call `Map` without `new`, check the last commit.

TorsteinHonsi (changes requested):
These are the constructors that we currently support. And since we already broke it once, we need to make sure we continue to support them as a minimum:

Current stable: https://jsfiddle.net/highcharts/Ltwc286d/
Current PR fails: https://jsfiddle.net/highcharts/Ltwc286d/1/

There are a couple of inconsistencies
* `Chart` cannot be called without `new`, but `StockChart` can.
* `MapChart` doesn't exist, instead `Map` can be called with or without `new`.
* `GanttChart` doesn't exist, and doesn't have to. Gantt was added after we recommended using the lowercase first factories.

samples/unit-tests/highcharts/constructors/demo.js
```diff
@@ -139,7 +139,7 @@ QUnit.test('Highstock', function (assert) {
     assert.strictEqual(typeof chart.rangeSelector, 'object', 'Is stock chart');
 
     calledBack = false;
-    chart = Highcharts.StockChart(
+    chart = new Highcharts.StockChart(
```
khlieng: Shouldnt we have some coverage of this and Map without `new`?

TorsteinHonsi: Yes... You fix?

TorsteinHonsi (approved):

PR #15257, opened 2021-03-03
Demo/nw landing page

nasin:
Hi, these are demos we want to incorporate on a landing page for the Newswatch campaign.


samples/highcharts/blog/nw-landing-page/demo.js
```diff
@@ -0,0 +1,1185 @@
...
+    title: {
+        text: ''
+    },
+    tooltip: {
+        enabled: false,
+        formatter: function () {
+            var info = '';
+            switch (this.color) {
+            case dirDist50:
+                info = 'is an aiport <b>more than 50</b> direct distinations';
```
TorsteinHonsi, line 252: Typo: "aiport"

samples/highcharts/blog/nw-landing-page/demo.js
```diff
@@ -0,0 +1,1185 @@
...
+        enabled: false
+    },
+    credits: {
+        enabled: false
+    },
+    title: {
+        text: ''
+    },
+    tooltip: {
+        enabled: false,
```
TorsteinHonsi, line 247: Why do we disable the tooltip (but still have advanced formatters for it)? If we can, I think we should keep the tooltip as it is a great interactive and assistive feature. The demos are nicely set up with real world examples like planets or airports, all of this is lost if we don't have the tooltip.

samples/highcharts/blog/nw-landing-page/demo.js
```diff
@@ -0,0 +1,1185 @@
...
+            [-1, 1],
+            [0, 0],
+            [1, 1],
+            [2, 2],
+            [3, 3]
+        ]
+    }]
+});
+
+/*NETWOR GRAPH*/
```
TorsteinHonsi, line 208: Typo: "NETWOR"

samples/highcharts/blog/nw-landing-page/demo.js
```diff
@@ -0,0 +1,1185 @@
...
+        innerSize: '70%',
+        data: planetAtmosphere
+    }]
+},
+function () {
+    const chart = this,
+        series = chart.series[0];
+    let newData = [];
+    let count = 0;
+    setInterval(function () {
```
TorsteinHonsi, line 119: All the animations are quite processor heavy, so if an end user leaves this page open, the CPU is working hard. I propose one of two solutions:
a. Start the animations initially, and stop them after a fixed time when we assume the end user has scrolled down. Optionally we can start them again in a new sequence if the mouse/pointer interacts with the container div (just restart it with a mouseover event).
b. Use the [jQuery appear plugin](https://github.com/morr/jquery.appear), or something equivalent non-jQuery, to run the animations only when inside the viewport. But here too they should stop after a while in case the page is left open.

samples/highcharts/blog/nw-landing-page/demo.html
```diff
@@ -0,0 +1,44 @@
...
+<script src="https://cdn.jsdelivr.net/npm/bootstrap@5.0.0-beta2/dist/js/bootstrap.bundle.min.js"></script>
+<link href="https://cdn.jsdelivr.net/npm/bootstrap@5.0.0-beta2/dist/css/bootstrap.min.css" rel="stylesheet">
+
+</head>
+<body>
+
+ <div id="content" class="site-content">
+	 
+		<div class="row" ><!-- style="background-color: #47475c" -->
+	 		<div class="col-lg-12 d-lg-flex justify-content-center bkgWhite" >
```
TorsteinHonsi, line 16: It seems like not all sizes work perfectly - on a wide screen I see 6 charts side by side. When reducing the width of my window gradually, the sixth and the fifth chart are partially clipped before they disappear. Then when below 768px, only four charts are visible, but I assume that is intentional, to avoid taking up too much vertical space.

TorsteinHonsi (changes requested):
Thanks! In addition to the inline comments, I think the demo should be called something else than `nw-landing-page`. In two years it is no longer new, but hopefully still in production. Maybe `blog-landing-page` or something is more descriptive.

samples/highcharts/blog/nw-landing-page/demo.html
nasin: Yes, I intentionally clipped the container on small screens to limit it to 4 graphs.

TorsteinHonsi: Okay, so then we have a problem between 768 and the full container width.

nasin: I fixed the CSS so all charts display regardless of the screen size.

samples/highcharts/blog/nw-landing-page/demo.js
nasin: I create a function that runs one time on load.

samples/highcharts/blog/nw-landing-page/demo.js
nasin: I put the tooltips back and fixed the first graph so it actually means something.

samples/highcharts/blog/nw-landing-page/demo.js
nasin: fixed the typo

TorsteinHonsi (changes requested):
Thanks!

Now we can set [tooltip.outside](https://api.highcharts.com/highcharts/tooltip.outside) to prevent the tooltips from overlapping the content inside the small chart boxes.

TorsteinHonsi (approved):

PR #15263, opened 2021-03-04
Removed unnecessary formatters from art-grants demo.

oysteinmoseng:
Removed some old and unnecessary formatters from art-grants demo as discussed.


TorsteinHonsi (approved):
Thanks!

PR #15267, opened 2021-03-04
bugfix/15188-debugger-no-chart

### LINKED ISSUE #15188: Debugger module doesn't register AST's warning
#### Behaviour
`chart` is not passed into the warning call, so the debugger module is not showing info for it.

https://jsfiddle.net/BlackLabel/Ltokcnmz/
The warning is in the console but is not picked up by the module.

#### Product version
Since 9.0.0

KacperMadej:
Fixed #15188, added error display fallback to the most recently added chart for debugger module.


ts/Core/Utilities.ts
```diff
@@ -515,6 +516,8 @@ function error(
         message += additionalMessages;
     }
 
+    // Display error on the chart causing the error or the last created chart.
+    chart = chart || find(charts.slice().reverse(), (c?: Chart): boolean => !!c);
```
TorsteinHonsi, line 520: Can we do this logic in the debugger module instead? Because when the module is not loaded we show the error in the console, and we don't need the chart context.

TorsteinHonsi (changes requested):

ts/Core/Utilities.ts
KacperMadej: @TorsteinHonsi 
`chart` is needed in this function context only to fire the event. To keep the debugger module running on an event we would have to fire the event on something else and pass `chart` as an additional argument - so it will still be in this context. The "something else" to fire the event on could be the `Highcharts` instance - this makes some sense because if there is no chart then we loop the `charts` array, but I haven't seen any event fired like this. An alternative event source that I can think of is `window` or `document`.

TorsteinHonsi: Yes, I think events that are not related to a chart should be called on one of those objects. `Highcharts`, `window` or `document`. We probably cannot use the `Chart` class either because we can have a pure renderer setup without a `Chart`.

@bre1470 Which object do you think makes more sense? When we get rid of the namespace, we still have Highcharts as the return for Globals.ts, right? And `Highcharts.win` should also always be given.

bre1470: Yes, we can use the Globals module for that purpose, but we could also consider a chart registry similar to one for series.

TorsteinHonsi: We have `Highcharts.charts`. So if we pass `Highcharts` (=Globals)  to the debugger module, it will use Highcharts.charts to find the last one.

bre1470: Sorry, I confused instance and factory/constructor.

KacperMadej: @bre1470 I changed the context of the event to `Highcharts` to provide more info (in the commit below). If `Highcharts` is ok, then all should be done.

TorsteinHonsi (approved):
Great, thanks!

TorsteinHonsi (changes requested):
Oops, please fix the linting.

TorsteinHonsi:
What about the new visual diffs? It seems it now reports errors that were previously lost. Is it because we didn't have the `chart` scope previously?

KacperMadej:
@TorsteinHonsi Yes, those errors were there but not displayed. Most (if not all) could be resolved by fixing the source of the warnings.
Maybe I should set up a new issue, use the visual diffs from here to fix those problems, and after that will be resolved we'll get back here?

TorsteinHonsi:
Yes, that's a good idea! So we put this one on hold for now then.

TorsteinHonsi (approved):
Thanks! Changes approved, now awaiting CI. The two remaining visual diffs should be fixed with the last master merge.

PR #15273, opened 2021-03-05
bugfix/15260-networkgraph-className

### LINKED ISSUE #15260: Unable to add class to network graph data
#### Expected behaviour

Following [the relevant documentation](https://api.highcharts.com/highcharts/series.networkgraph.data.className), I expect a class name to be added to the lines in the network graph. I'm using the following data (more in the jsfiddle below):

```
data: [{
  from: "A",
  to: "B",
  className: "foo",
  color: "green",
}, {
  ...
}]
```

#### Actual behaviour

No class is added at all. Adding [`color`](https://api.highcharts.com/highcharts/series.networkgraph.data.color) _does_ work. As do `to` and `from`.

https://jsfiddle.net/ordyx9bu/

#### Product version

#### Affected browser(s)
Firefox, Chromium

khlieng:
Fixed #15260, networkgraph point `className` option did not work.

### DIFF
--- a/samples/unit-tests/series-networkgraph/networkgraph/demo.js
+++ b/samples/unit-tests/series-networkgraph/networkgraph/demo.js
@@ -29,7 +29,11 @@ QUnit.test('Network Graph', function (assert) {
         },
         keys: ['from', 'to'],
         data: [
-            ['A', 'B'],
+            {
+                from: 'A',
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #15274, opened 2021-03-05
bugfix/15179-crisp-subpixel-radius

### LINKED ISSUE #15179: Point marker and grid line don't always line up (crisping)
#### Expected behaviour

In order to center point markers on 1px wide grid lines, I've set the marker radius to 2.5. I expect every marker to perfectly line up with the grid lines, crosshairs etc.

#### Actual behaviour

Some of the markers are not horizontally centered on the grid line.

Demo: https://jsfiddle.net/up2ch65b/

On my screen markers at 0 and 2 are both offset 1 pixel to the right. It might be different for you depending on your screen resolution and scaling factor, so here's a screenshot of the issue as well: 
image

#### Analysis

I've identified the cause for this behavior in the following piece of code:

https://github.com/highcharts/highcharts/blob/8de4d3091cd6f59c230e3dcfd682d0c49b71fdfd/ts/Core/Series/Series.ts#L5221-L5223

The value of `point.plotX` is rounded _before_ subtracting the radius. In order to account for floating point radii, the rounding should be done _after_ subtracting the radius.

Here is the same demo but with my proposed fix: https://jsfiddle.net/aw2exbpr/

khlieng:
Fixed #15179, crisping did not work correctly for subpixel radius markers.

### DIFF
--- a/samples/unit-tests/series/marker/demo.js
+++ b/samples/unit-tests/series/marker/demo.js
@@ -72,6 +72,7 @@ QUnit.test('Marker size and position', function (assert) {
                 data: [1, 2, 3],
                 animation: false,
                 marker: {
+                    radius: 2.5,
                     animation: false,
                     states: {
                         hover: {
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #15281, opened 2021-03-08
TS: no nulls for Axis.min and Axis.max

TorsteinHonsi:
No need for `null`, better use `undefined`.


TorsteinHonsi:
Approved, but decided to wait until next major release and do a systematic assessment of every place we use `null` where it should be `undefined`.

TorsteinHonsi:
Closing this for now

PR #15284, opened 2021-03-09
Tools: added demos/samples deploy to CI setup

goransle:
Added triggers for building samples and demos via the highcharts-demo-manger CI setup.

Samples and demos should now be updated nightly, if there are changes in the relevant folders. It should also update demo thumbnails if any new demos are added.

I also added a full refresh on release, but this may be redundant (except perhaps in the case of thumbnails being affected by changes in HC).


TorsteinHonsi (approved):
Thanks!

PR #15285, opened 2021-03-09
bugfix/13449-update-dataLabels-allowOverlap

### LINKED ISSUE #13449: Update dataLabels allowOverlap property via plotOptions
#### Expected behaviour

Updating `dataLabels.allowOverlap` doesn't work when this option has been updated via using `chart.update` and trying to change plotOptions.

#### Actual behaviour

Should work as it used to be in the 6.2.0 version: https://jsfiddle.net/BlackLabel/9gy7s64L/

As a workaround, it is possible to enabling allowOverlap by updating `series.dataLabels` options, demo:  https://jsfiddle.net/BlackLabel/vshj95to/

Demo: https://jsfiddle.net/BlackLabel/g9z831jm/

#### Product version

Highcharts JS v8.1.0 (2020-05-05)

#### Affected browser(s)
All

khlieng:
Fixed #13449, updating `dataLabels.allowOverlap` did not work.

### DIFF
--- a/samples/unit-tests/datalabels/overlapping/demo.js
+++ b/samples/unit-tests/datalabels/overlapping/demo.js
@@ -42,6 +42,23 @@ QUnit.test('Overlapping dataLabels should be hidden', function (assert) {
         true,
         'Rotated dataLabel hidden (#7362).'
     );
+
+    chart.update({
+        plotOptions: {
+            series: {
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #15286, opened 2021-03-09
DeepPartial for Axis options

TorsteinHonsi:
DeepPartial for Axis options, prevent having to cast `options.label.someThing` to any. Following up this, other options in `AxisOptions` should also be make non-optional.


bre1470 (approved):
Awesome! 👍

Issue #15140, opened 2021-02-15
Directional Movement Index (DMI) technical indicator


pawelfus:
Implement DMI technical indicator:

- [x] - algorithm (ATR should not be included, loaded manually)
- [x] - algorithm tests - (@karolkolodziej)
- [x] - create a demo (e.g. https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/stock/indicators/apo/)
- [x] - describe `params` in API (e.g. https://api.highcharts.com/highstock/series.ad.params)
- [x] - add it to list of indicators (https://www.highcharts.com/docs/stock/technical-indicator-series) - if necessary, add to the bottom list too: https://www.highcharts.com/docs/stock/technical-indicator-series#multiple-series-bindings
- [x] - add it to indicators-all.js (http://code.highcharts.com/stock/indicators/indicators-all.src.js)
- [x] - test it in StockTools (e.g. separate yAxis for Oscillators)
- [ ] - test it in Popup (e.g. if params are not missing)
- [x] - if necessary, add `lang` in API (https://api.highcharts.com/highstock/lang.navigation.popup)
- [x] - add the indicator to the demo with all indicators (http://utils.highcharts.local/samples/#view/stock/demo/all-indicators)

imdanshraaj:
how can i remove dx line in highcharts? i just want to show plus di and minus di line.

karolkolodziej:
@imdanshraaj The easiest would be simply hiding that line by setting the line width to zero. 
In this case, you also have to change the tooltip - [demo](https://jsfiddle.net/BlackLabel/8Lcrx4hz/).

If that is not what you have been looking for you can always create your custom indicator - [guide](https://www.highcharts.com/docs/stock/custom-technical-indicators).

For future reference, with questions like this, please contact us via [available support channels.](https://www.highcharts.com/blog/support/)

PR #15287, opened 2021-03-09
enhancement/15140-DMI-indicator

karolkolodziej:
Enhancement #15140 
Initial commit only with basic algorithm test.


ts/Stock/Indicators/DMI/DMIIndicator.ts
```diff
@@ -0,0 +1,398 @@
...
+    ): number {
+        const currentHigh = yVal[i][1],
+            currentLow = yVal[i][2],
+            previousHigh = yVal[i - 1][1],
+            previousLow = yVal[i - 1][2];
+
+        // If any value is null, return 0.
+        if (!currentHigh || !currentLow || !previousHigh || !previousLow) {
+            return 0;
+        }
```
raf18seb, line 156: I have a question here. I wanted to handle `null` values for `open` or `low`, but it clearly changes the further indicator values. Any suggestions?

ts/Stock/Indicators/DMI/DMIIndicator.ts
```diff
@@ -0,0 +1,398 @@
...
+    ): number {
+        return smoothedDM / (tr || 1) * 100;
+    }
+
+    public calculateDX(
+        plusDI: number,
+        minusDI: number
+    ): number {
+        return correctFloat(
+            Math.abs(plusDI - minusDI) / (Math.abs(plusDI + minusDI) || 1) * 100
```
raf18seb, line 183: Dividing - I want to make it safe from 0, but it might change the indicators values. Well, rather change values than got an error. Is it OK? If not, any ideas?

ts/Stock/Indicators/DMI/DMIIndicator.ts
```diff
@@ -0,0 +1,398 @@
...
+     * @requires     stock/indicators/indicators
+     * @requires     stock/indicators/dmi
+     * @optionparent plotOptions.dmi
+     */
+    public static defaultOptions: DMIOptions = merge(SMAIndicator.defaultOptions, {
+        /**
+         * @excluding index
+         */
+        params: {
+            index: void 0, // DMI has unchangeable index
```
raf18seb, line 78: The algorithm is strictly based on specific `high`, `low` and `close` values. I excluded it from API, but the `index` field is still displayed in a popup in Stock Tools. Should we correct it in Stock Tools (bug/feature)? Or can it be somehow handled here?

raf18seb (commented):
Except a few questions in the comments: @pawelfus me and @karolkolodziej had a lot of corrections at the final DMI values. I am pretty sure that the values are OK now (based on a spreadsheet found somewhere on the web) but I think that the final values should be double-checked one more time.

ts/Stock/Indicators/DMI/DMIIndicator.ts
```diff
@@ -0,0 +1,398 @@
...
+        i: number,
+        isPositiveDM?: boolean
+    ): number {
+        const currentHigh = yVal[i][1],
+            currentLow = yVal[i][2],
+            previousHigh = yVal[i - 1][1],
+            previousLow = yVal[i - 1][2];
+
+        // If any value is null, return 0.
+        if (!currentHigh || !currentLow || !previousHigh || !previousLow) {
```
raf18seb, line 154: TODO: use `H.defined` (valid 0 values)

ts/Stock/Indicators/DMI/DMIIndicator.ts
```diff
@@ -0,0 +1,398 @@
...
+            accumulatedValues - accumulatedValues / period + currentValue
+        );
+    }
+
+    public getTR(
+        currentPoint: Array<number>,
+        prevPoint?: Array<number>
+    ): number {
+        // If any value is null, return 0.
+        if (!currentPoint[1] || !currentPoint[2] || !currentPoint[3]) {
```
raf18seb, line 202: TODO: use `H.defined` (valid 0 values)

raf18seb (changes requested):

ts/Stock/Indicators/DMI/DMIIndicator.ts
karolkolodziej: Mentioned issue is reported here: #15362

ts/Stock/Indicators/DMI/DMIIndicator.ts
```diff
@@ -0,0 +1,398 @@
...
+        let prevSmoothedPlusDM: number = 0,
+            prevSmoothedMinusDM: number = 0,
+            prevSmoothedTR: number = 0,
+            i: number;
+
+        for (i = 1; i < yValLen; i++) {
+            let smoothedPlusDM: number,
+                smoothedMinusDM: number,
+                smoothedTR: number,
+                currentPlusDM: number, // +DM
```
karolkolodziej, line 245: You don't have `prevPlusDM`, I would consider removing `current`

ts/Stock/Indicators/DMI/DMIIndicator.ts
```diff
@@ -0,0 +1,398 @@
...
+    public getTR(
+        currentPoint: Array<number>,
+        prevPoint?: Array<number>
+    ): number {
+        // If any value is null, return 0.
+        if (!currentPoint[1] || !currentPoint[2] || !currentPoint[3]) {
+            return 0;
+        }
+
+        const HL = currentPoint[1] - currentPoint[2],
```
karolkolodziej, line 206: You are not using these variables again, I would make it in one line with some additional comments what is what. 
Like:
```
return correctFloat(
   Math.max(
       currentPoint[1] - currentPoint[2], // high - low
       ....
   );

)

File: ts/Stock/Indicators/DMI/DMIIndicator.ts (hunk 2 above)
Comment (karolkolodziej) [reply ts/Stock/Indicators/DMI/DMIIndicator.ts:183]: But do we have to really be concerned about that? 
I don't see a chance that the sum of two positive values give 0. 
Both of them must be 0 which is impossible in this case... 
Correct me if I'm wrong?

File: samples/unit-tests/indicator-dmi/recalculations/demo.js [hunk 8]
```diff
@@ -0,0 +1,266 @@
...
+            testNumberOfPoints = textDescription => {
+                assert.strictEqual(
+                    chart.series[0].points.length,
+                    DMIIndicator.points.length +
+                        DMIIndicator.options.params.period,
+                    textDescription
+                );
+            };
+
+        testNumberOfPoints(
```
karolkolodziej, line 57: An interesting idea for writing the tests! 
I've never seen this before ;)
However, I would choose one way and remain constant through the file.

samples/unit-tests/indicator-dmi/recalculations/demo.js
```diff
@@ -0,0 +1,266 @@
...
+                [59.525469402892, 37.607393331916384, 9.541683831350067],
+                [44.821122205165, 34.70030407492836, 13.221302313080818]
+            ],
+            'The DMI (DX, +DI, -DI) values should be correct.'
+        );
+
+        chart.series[0].addPoint([1555335000000, 49.65, 49.96, 49.5, 49.81]);
+
+        testNumberOfPoints(
+            'After addPoint: the number of DMI points should be equal to the' +
```
karolkolodziej, line 83: Use backticks instead of adding two strings.

samples/unit-tests/indicator-dmi/recalculations/demo.js
```diff
@@ -0,0 +1,266 @@
...
+            [15, 5, 26.9441, 26.1332, 26.2221],
+            [16, 5, 26.5189, 25.4307, 26.0144],
+            [17, 5, 26.5189, 25.3518, 25.4605],
+            [18, 5, 27.0927, 25.8760, 27.0333],
+            [19, 5, 27.6860, 26.9640, 27.4487]
+        ]);
+
+        testNumberOfPoints(
+            'After setData: the number of DMI points should be equal to the ' +
+            'number of the main series\' points plus period.'
```
karolkolodziej, line 131: Backticks will help you avoid this part `main series\' points`

ts/Stock/Indicators/DMI/DMIIndicator.ts
```diff
@@ -0,0 +1,398 @@
...
+                /**
+                 * Pixel width of the line.
+                 */
+                lineWidth: 1,
+                /**
+                 * Color of the line.
+                 *
+                 * @type {Highcharts.ColorString}
+                 */
+                lineColor: '#00ff00'
```
karolkolodziej, line 109: Maybe we might use some predefined colors? Like it is done in PC indicator
`lineColor: palette.colors[2],` which is `#90ed7d` and at least for me this kind of green is more pleasant to the eye than `#00ff00`.
Just an idea to discuss

karolkolodziej (commented):
The algorithm looks correct! Great job!
Some minor things to improve/discus.

pawelfus (approved):
I have nothing to add, nice! 👍

ts/Stock/Indicators/DMI/DMIIndicator.ts
pawelfus: I'm afraid none of the indicators is now resolving `null` values in any way. If we decide to do that, then it will be probably resolved in the indicators core, not per indicator (by filtering out null/invalid points/values).

ts/Stock/Indicators/DMI/DMIIndicator.ts
raf18seb: Probably you're right, I was thinking the same. Yea, I'll remove it and, if any errors occur in the future, we can easily fix it.

ts/Stock/Indicators/DMI/DMIIndicator.ts
raf18seb: Hmm, I'm gonna defend this. Yes, initially I had it written the way you are proposing, but, in this case (4 variables each used twice) it is much more readable with properly named and defined variables.

However, if others share your opinion, I will change it.
@pawellysy @pawelfus what do you think?

samples/unit-tests/indicator-dmi/recalculations/demo.js
raf18seb: I'm gonna defend this. We have plenty of tests where, for example, asserts are inside `forEach` instead of repeating them one by one. I know this case is slightly different, but it's not wrong - I test the same assert 6 times so using a function to make it a little bit shorter seems fine to me.

However, if others share your opinion, I will change it.
@pawellysy @pawelfus what do you think?

ts/Stock/Indicators/DMI/DMIIndicator.ts
raf18seb: Sure, I agree.

ts/Stock/Indicators/DMI/DMIIndicator.ts
raf18seb: Without assigning `undefined` here, it would inherit the value from SMA and that value would be displayed in Stock Tools popup. I decided to assign `undefined` to clearly let everyone know that the index is no needed. Furthermore, we will fix it in Stock Tools popup, probably just by filtering out `undefined` values or similarly.

raf18seb: I would leave it as it is and change/improve in the future if needed.

raf18seb: I would leave it as it is and change/improve in the future if needed (unless you have any objections).

raf18seb:
@bre1470 just to make sure - would you check TS definitions, please? Especially in `DMIOptions.d.ts` - are all definitions there needed?

raf18seb (approved):

ts/Stock/Indicators/DMI/DMIOptions.d.ts
```diff
@@ -0,0 +1,46 @@
...
+}
+
+export interface DMILineOptions {
+    styles?: CSSObject;
+}
+
+export interface DMIZonesOptions {
+    startIndex?: number;
+}
+
```
bre1470, line 41: As far as I can see, this interface is not in use and can be removed.

bre1470 (changes requested):
Except one interface it looks good. 👍

karolkolodziej:
Since I opened this PR, I'm not able to approve it, but for me now it looks correct. 
Great work!

ts/Stock/Indicators/DMI/DMIOptions.d.ts
raf18seb: Thanks!

pawellysy (approved):
Looks good!

bre1470 (approved):

TorsteinHonsi (approved):
Nice work, thanks!

samples/unit-tests/indicator-dmi/recalculations/demo.js
bre1470: The two concerns I have about it are:
- Arrow functions are not supported by Internet Explorer. So it will throw a syntax error.
- Inconsistent syntax could make it harder to refactor all tests to a different setup in the future.

raf18seb: These are only tests, not official demos - more and more tests are written in ES6, so the IE should not be a concern here.
When it comes to future refactoring - good point, didn't know about refactoring plans. If there are any, you could make a short note and let us know how you would like us to write tests and what should we pay attention to.

PR #15288, opened 2021-03-09
bugfix/12100-mapNavigation-indicators-menu-scrollable

### LINKED ISSUE #12100: StockTools indicators menu is not scrollable with mapNavigation.enabled: true

karolkolodziej:


ts/Extensions/Annotations/Popup.ts
```diff
@@ -1229,7 +1217,7 @@ H.Popup.prototype = {
             return createElement(
                 DIV,
                 {
-                    className: PREFIX + 'tab-item-content'
+                    className: PREFIX + 'tab-item-content highcharts-no-mousewheel'// #12100
```
raf18seb, line 1220: The `PREFIX` variable is `highcharts-`. I think it would be better if you added those classes this way:
```js
className: PREFIX + 'tab-item-content' + ' ' + PREFIX + 'no-mousewheel'// #12100

```

raf18seb (changes requested):
A few questions:
- there are few strange files like e.g. `js/Extensions/Debugger/Debugger.js`. I don't think they have been added intentionally.
- what about tests? I assume they already exist, am I right?

Except that, I have only one comment question about `PREFIX`

PR #15289, opened 2021-03-09
TS: Improved ES6 pattern and types of Globals

bre1470:
- Added internal custom H.Obj.keys type reference, returning `keyof object` for usage on objects without indexer.
- Migrated shared properties to ES6 namespace Globals (= window.Highcharts in masters).
- Clarified todo's of UMD workaround.
- Simplified assignment of SeriesRegistry.
- Deprecated custom browser properties, as TypeScript tends to remove them gradually from the internal type definitions.
- Removed outdated global types

### DIFF
--- a/ts/Core/Globals.ts
+++ b/ts/Core/Globals.ts
@@ -10,9 +10,47 @@
 
 'use strict';
 
+/* *
+ *
+ *  Imports
+ *
+ * */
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (approved):
Thanks!

PR #15293, opened 2021-03-10
bugfix/12100-indicators-menu-scrollable-refectoring

### LINKED ISSUE #12100: StockTools indicators menu is not scrollable with mapNavigation.enabled: true

karolkolodziej:


pawelfus (approved):
Thanks! 👍

TorsteinHonsi (approved):
Perfect, thanks!

PR #15296, opened 2021-03-10
Bugfix/15291-inconsistent-addition-of-symbols

### LINKED ISSUE #15291: Inconsistent addition of symbols.rect in modules
#### Expected behaviour
A predictable implementation of the symbol

#### Actual behaviour
* `symbols.rect` is not defined in the core
* The dotplot module adds it as a call to `callout`.
* The heatmap module adds it (upon heatmap series initialization)  as a call to `square`.

* Dotplot: https://jsfiddle.net/highcharts/sb1r04gz/
* Heatmap, before series creation: https://jsfiddle.net/highcharts/sb1r04gz/1/
* Heatmap, after series creation: https://jsfiddle.net/highcharts/sb1r04gz/2/

Note that the rectangles in point 1 and 3 are not identical. Crisping is applied to the heatmap/square implementation but not to the dotplot/callout implementation.

#### Product version
v9.0.1

#### Affected browser(s)
All

___

Instead of letting the modules load this feature in an inconsistent way, we should probable just apply `rect` as an alias of `square` in the core, and remove the module additions. It will be a lightweight addition to the core.

DJTechnoo:
Fixed #15291, inconsistent addition of symbols.


TorsteinHonsi (changes requested):
Awesome!

Please provide a description for the changelog.

Other than that, and if the tests pass, it's ready for merge.

TorsteinHonsi (approved):

PR #15299, opened 2021-03-10
bugfix/8603-stock-forced-axis-options

### LINKED ISSUE #8603: Highstock: Dynamically added axes do not have same defaults

khlieng:
Fixed #8603, dynamically added stock axes missed forced options, addition to #15164.

### DIFF
--- a/samples/unit-tests/stockchart/general/demo.js
+++ b/samples/unit-tests/stockchart/general/demo.js
@@ -167,6 +167,14 @@ QUnit.test('Default options for stock chart', function (assert) {
     assert.strictEqual(
         chart.yAxis[1].options.title.text,
         null,
-        'Axis should have stock defaults applied'
+        '#8603: Axis should have stock defaults applied'
+    );
+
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #15302, opened 2021-03-11
bugfix/15170-stocktools-popup-lang

### LINKED ISSUE #15170: StockTools: Missing lang options for `<indicator>.params`
#### Expected behaviour
Available options to change default `params` names, like `period` or `index`.

#### Actual behaviour
We display the name of the option:
Screenshot 2021-02-18 at 12 11 51

List of options:
- `index`
- `period`
- `standardDeviation`
- `periodTenkan`
- `periodSenkouSpanB`
- `periodATR`
- `multiplierATR`
- `shortPeriod`
- `longPeriod`
- `signalPeriod`
- `decimals`
- `algorithm`
- `topBand`
- `bottomBand`
- `initialAccelerationFactor`
- `maxAccelerationFactor`
- `increment`
- `multiplier`
- `ranges`
- `highIndex`
- `lowIndex`
- `deviation`
- `xAxisUnit`
- `factor`

Note: Options work when defined in `lang`: https://jsfiddle.net/BlackLabel/jpyeu2mh/ (choose IKH indicator) - but default options are not defined (thus not listed in API: https://api.highcharts.com/highstock/lang.navigation.popup )

Steps:
- click on the Indicators button
- check some of the indicators (MACD, ZigZag etc)

raf18seb:
Fixed #15170, added `lang` options for indicators parameters in Stock Tools popup.

### DIFF
--- a/ts/Extensions/Annotations/Popup.ts
+++ b/ts/Extensions/Annotations/Popup.ts
@@ -35,6 +35,7 @@ const {
     isString,
     objectEach,
     pick,
+    setOptions,
     stableSort,
     wrap
 } = U;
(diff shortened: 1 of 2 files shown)


pawelfus (approved):
Thanks 👍

pawelfus (changes requested):
Sorry, prematurely accepted. Could move lang definitions from `Popup` to `StockToolsGUI`, where the rest of lang options are set? See: https://github.com/highcharts/highcharts/blob/master/ts/Stock/StockToolsGui.ts#L249

ts/Extensions/Annotations/Popup.ts
```diff
@@ -165,6 +166,41 @@ var indexFilter = /\d/g,
     LI = 'li',
     H3 = 'h3';
 
+// Lang definitions for indicators parameters in Stock Tools popup, #15170
+setOptions({
```
karolkolodziej, line 170: Rather question than request.
Is there any difference in performance or anything else between `setOptions` and `merge` like below?
```
merge(defaultOptions, {
    lang: {
       ...
    }
})
```

karolkolodziej (approved):

ts/Extensions/Annotations/Popup.ts
raf18seb: Yes. `setOptions` changes (extends) the first argument when merge only returns the whole new object (without assigning or changing the first one). So you would have to either do something like this:
```js
defaultOptions = merge(defaultOptions, {
    lang: {
       ...
    }
})
```
or
```js
merge(true, defaultOptions, {
    lang: {
       ...
    }
})
```

https://api.highcharts.com/class-reference/Highcharts#.merge%3CT%3E

karolkolodziej: Ok, `setOptions` looks correctly in this case! Thanks ;)

TorsteinHonsi (approved):
Thanks! Pending @pawelfus' final approval

pawelfus (approved):
Looks great, thanks!

PR #15304, opened 2021-03-11
Typesafe AxisOptions

TorsteinHonsi:
Make options that are not undefined in real life, not undefined in TypeScript. This is a follow-up from using `DeepMerge` in axis options.

Go over each option and figure out which ones may actually be `undefined`
 - [x] `Axis.options` root 
 - [x] `Axis.options.labels`
 - [x] `Axis.options.title`
 - [x] `Axis.crosshair`
 - [x] `Axis.crosshair.label`
 - [x] `Axis.breaks`
 - [x] `YAxisOptions` (minColor and maxColor probably belong on `ColorAxisOptions` instead?


ts/Series/Funnel3D/Funnel3DSeries.ts
```diff
@@ -257,16 +256,16 @@ class Funnel3DSeries extends ColumnSeries {
...
+        } as DeepPartial<Highcharts.AxisOptions>);
+        merge(true, this.yAxis.options, {
             gridLineWidth: 0,
-            title: null,
+            title: void 0,
             labels: {
                 enabled: false
             }
-        });
+        } as DeepPartial<Highcharts.AxisOptions>);
```
bre1470, line 268: The `something as somethingElse` is forced typing and reduces the type check to the primitive scope, in the sense that return type and forced type share the same primitive type (object).

ts/Core/Axis/Tick.ts
```diff
@@ -432,7 +432,7 @@ class Tick {
             // (#8809)
             if (
                 label.textWidth &&
-                !(labelOptions.style && labelOptions.style.width) &&
+                !labelOptions.style.width &&
                 !(label.styles as any).width
```
bre1470, line 436: All the `style` types are the one thing I wonder, if we should go with the built-in variant `CSSStyleDeclaration` of `lib.dom.d.ts`.

ts/Core/Axis/TreeGridAxis.ts
```diff
@@ -106,7 +106,7 @@ namespace TreeGridAxis {
      *
      * */
 
-    export interface AxisBreakObject extends AxisTypes.AxisBreakObject {
+    export interface AxisBreakObject extends Highcharts.XAxisBreaksOptions {
```
bre1470, line 109: We have to get away from referencing Highcharts namespace. In general it would make sense, to move all namespace types into DTS files, so that they become extendable like in a namespace.

ts/Core/Axis/Axis3D.ts
```diff
@@ -368,7 +368,7 @@ class Axis3D {
     /**
      * @optionparent xAxis
      */
-    public static defaultOptions: Highcharts.AxisOptions = {
+    public static defaultOptions: DeepPartial<Highcharts.AxisOptions> = {
```
bre1470, line 371: Why is this? Did you tried `Highcharts.XAxisOptions` without deep partial?

On a side note: Maybe we should build Highcharts.AxisOptions as a deep partial intersect type instead of an union. Defaults should not use it though, instead their specific axis interface.

bre1470 (commented):
The removed `any` types look great! 👍 

Usage of `DeepPartial` makes a flawed impression. The idea was to have something like
```ts
class {
  static default: DefaultType = merge({…});
  options: DefaultType;
  userOptions: DeepPartial<DefaultType>;
  init(options: DeepPartial<DefaultType>);
  update(options: DeepPartial<DefaultType>);
}
```

bre1470:
In general it should be possible to do `const merged: DefaultType = merge(DefaultType, DeepPartial<DefaultType>);` without any special `any` or `as` casting. If that is not working, the typing of the merge function might need improvements.

ts/Core/Axis/Tick.ts
TorsteinHonsi: Yes maybe, but I think it's a bit out of scope for this PR

ts/Core/Axis/Axis3D.ts
TorsteinHonsi: It's the `defaultOptions` of the Axis3D component, so it shouldn't hold the full `AxisOptions`

ts/Series/Funnel3D/Funnel3DSeries.ts
TorsteinHonsi: Yes, we should have the merge return the same type as its first argument, (or second argument if the first is bool).

TorsteinHonsi: Alright, this _does_ work. The `as` was probably just needed for an intermediate step.

TorsteinHonsi:
> Usage of DeepPartial makes a flawed impression. The idea was to have something like

Agreed. But I think we need to start with the lower (leafier) levels then work our way up, right? Then when we see the full picture we can probably just set DeepPartial on the top level and for every update function, userOptions property etc.

ts/Core/Axis/Axis3D.ts
bre1470: Wasn't the idea of the full XAxisOptions to only have the defaults non-optional? I still do not understand why defaults need to be partial. They should by definition contain all non-optionals, therefor default values and have to work without DeepPartial. Something is not working, if this is necessary.

TorsteinHonsi: The `Axis3D.defaultOptions` are a partial of the full `AxisClass.defaultOptions` because they are merged in here:

https://github.com/highcharts/highcharts/blob/fa028f8596197eb0809b6524123317e9c8bb58b1/ts/Core/Axis/Axis3D.ts#L490

If not partial, we would need a separate type for the 3d axis options overrides.

bre1470 (approved):
Okay, let's merge it. 👍 After discussion we agreed to merge this so far, even though it is not type safe.

To-do's:
- We have to fix `Highcharts.AxisOptions` for `Axis.options`, as `XAxisOptions|YAxisOptions` is incorrect and type is not helpful `XAxisOptions|DeepPartial<YAxisOptions>`, unless...
- we also change `YAxisOptions`, because it is not extending `XAxisOptions`, but just defining additions of defaults for `XAxisOptions`, similar to options of compositions.
- In the end, we also need to find a better way to integrate the composition option types.

PR #15308, opened 2021-03-12
bugfix/event-leaks

khlieng:
Fixed event handler leaks in fullscreen module, flags series and `scrollablePlotArea`.


TorsteinHonsi (approved):
Thanks!

PR #15314, opened 2021-03-12
bugfix/errors-clean-up

KacperMadej:
Fixed errors in demos.
Prepared for https://github.com/highcharts/highcharts/pull/15267

<s>There is still a problem with advanced-annotations markers. It could be resolved by adding `display` to the `allowedAttributes` array in the AST part. I tried changing the code to use `visibility` but it is also not allowed.

@TorsteinHonsi Could we add `display` and `visibility` to `allowedAttributes`? or should this be fixed in a different way?</s>

EDIT: After a closer look, the elements causing problems are parts of `defs` in `SVG`, so `display: "none"` isn't needed and related functionality currently doesn't work anyway - reported here: https://github.com/highcharts/highcharts/issues/15315.


KacperMadej:
Commenting options merge was not intended - revert for the merge line will be added shortly. This should resolve the bad visual changes.

TorsteinHonsi (approved):
Thanks!

PR #15317, opened 2021-03-14
Prevent DOM access from format

TorsteinHonsi:
Improved parsing of deep paths in format strings to prevent reading DOM properties.


cvasseng (approved):
Looks pretty solid - I at least have not been able to find a way to refer to global variables so far.

PR #15318, opened 2021-03-15
bugfix/15160-touch-inactive-state

### LINKED ISSUE #15160: In iPad, Legend disable and enable of series1, disables series2 and does not restore even on click outside of the chart container.
#### Expected behaviour
Moving the cursor away from the legend area and out of chart container should show all the series in their respective solid colors with original opacity. (Same as in the browser mode)

#### Actual behaviour
When a series in a chart is disabled and enabled from the legend all the other series plots are dimmed. It does not show up in solid color even after clicking outside the chart container.

Problem Simulation: 
1.	Open fiddle https://jsfiddle.net/qszxe6ar/4/ in ipad mode in chrome browser
2.	Disable Series1 and enable it back.
3.	Series 2 is in disabled state. Even if I click outside the chart container, Series 2 is still in disabled state. (Refer screenshot below)

This happens in iPad devices, can also be seen in Chrome simulator mode of iPad.

#### Product version
iPad Legend disable and enable
Highcharts 8.1.2

#### Affected browser(s)
iPad Safari and other touch devices.

khlieng:
Fixed #15160, inactive state was not always cleared on touch devices when clicking outside the chart.

### DIFF
--- a/ts/Core/Pointer.ts
+++ b/ts/Core/Pointer.ts
@@ -1268,11 +1268,6 @@ class Pointer {
         ) {
             this.reset();
         }
-
-        // #15160
-        if (H.hasTouch) {
-            this.chart.series.forEach((series: Series): void => series.setState());
-        }
     }
 
     /**
@@ -1838,6 +1833,11 @@ class Pointer {
             });
 
             pointer.hoverX = chart.hoverPoints = chart.hoverPoint = null as any;
+        }
+
+        // #15160
+        if (H.hasTouch) {
+            this.chart.series.forEach((series: Series): void => series.setState());
         }
     }
 
--- a/ts/Core/Pointer.ts
+++ b/ts/Core/Pointer.ts
@@ -253,8 +253,6 @@ class Pointer {
         this.chart = chart;
         this.hasDragged = false;
         this.options = options;
-        this.unbindContainerMouseLeave = function (): void {};
-        this.unbindContainerMouseEnter = function (): void {};
         this.init(chart, options);
     }
 
@@ -300,9 +298,7 @@ class Pointer {
 
     public tooltipTimeout?: number;
 
-    public unbindContainerMouseLeave: Function;
-
-    public unbindContainerMouseEnter: Function;
+    public eventsToUnbind: Array<Function> = [];
 
     public unDocMouseMove?: Function;
 
@@ -378,11 +374,8 @@ class Pointer {
     public destroy(): void {
         var pointer = this;
 
-        if (typeof pointer.unDocMouseMove !== 'undefined') {
-            pointer.unDocMouseMove();
-        }
-
-        this.unbindContainerMouseLeave();
+        this.eventsToUnbind.forEach((unbind): void => unbind());
+        this.eventsToUnbind = [];
 
         if (!H.chartCount) {
             if (H.unbindDocumentMouseUp) {
@@ -1275,6 +1268,11 @@ class Pointer {
         ) {
             this.reset();
         }
+
+        // #15160
+        if (H.hasTouch) {
+            this.chart.series.forEach((series: Series): void => series.setState());
+        }
     }
 
     /**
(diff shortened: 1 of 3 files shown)


ts/Core/Pointer.ts
```diff
@@ -1275,6 +1268,11 @@ class Pointer {
         ) {
             this.reset();
         }
+
+        // #15160
+        if (H.hasTouch) {
+            this.chart.series.forEach((series: Series): void => series.setState());
```
TorsteinHonsi, line 1274: Did you consider doing this in the `Pointer.reset` function? In general, the job of that function is to "Reset the tracking by hiding the tooltip, the hover series state and the hover point". It is responsible of resetting the state of the `hoverSeries`:

https://github.com/highcharts/highcharts/blob/41e481abd12743502dcc1600ed70d0814e0ace5b/ts/Core/Pointer.ts#L1823-L1825

So maybe that's where we should add the code to reset inactive states?

... And why does this apply only to touch devices?

TorsteinHonsi (changes requested):

ts/Core/Pointer.ts
khlieng: Because the `mouseout` handler on the legend item resets the state for non-touch devices. Doing it in `reset` seems to work a bit better since it gets called when clicking outside the plot also.

khlieng: `mouseout` normally fires when clicking outside elements after clicking them on touch devices, not sure why it doesnt happen here. The container `mouseleave` also doesnt get fired if you only click a legend item and then outside the chart, which means `reset` doesnt get called in that case.

TorsteinHonsi: Legend `mouseout` currently sets inactive items state to '', and also sets item state to ''. It may be that it would be smarter to just call `pointer.reset` from legend `mouseout`. I think that's essentially what we are doing anyway.

TorsteinHonsi: .. and then we wouldn't have to double-check for 'H.hasTouch'.

khlieng: But the problem is the `mouseout` doesnt fire on touch devices because of this: https://github.com/highcharts/highcharts/blob/a55cae7ddaa73936a32fe430b83eff1e5dc4341e/ts/Core/Renderer/SVG/SVGElement.ts#L2128, #9682.

khlieng: Could we perhaps get rid of the 300ms click delay fix since I believe this is no longer an issue in modern browsers?

TorsteinHonsi: > But the problem is the mouseout doesnt fire on touch devices

Won't touching anywhere outside the plot area trigger `Pointer.reset` on touch devices? So if we centralize this inactive `setState` to `Pointer.reset`, then it will be handled there regardless of legend `mouseout`?

> Could we perhaps get rid of the 300ms click delay fix

Maybe... Where is that?

khlieng: No, touching a legend item and then outside the chart container wont fire the container `mouseleave` so `reset` wont be called. This is because of the `preventDefault` I linked above, its part of the click delay fix.

TorsteinHonsi: Okay... Then I agree that's where we should look for a solution. If we can avoid the `preventDefault` and still avoid the linked issue, it's the most consistent fix I think.

TorsteinHonsi (changes requested):
I'm afraid the latest change brought back #9682. Can be seen with Chrome on recent Androids using BrowserStack. Tested with the `area-basic` sample in Utils. Click the Mobile button for standalone view.

khlieng:
> I'm afraid the latest change brought back #9682. Can be seen with Chrome on recent Androids using BrowserStack. Tested with the `area-basic` sample in Utils. Click the Mobile button for standalone view.

It happens even if I add the `preventDefault` back in, if I remove the accessibility module everything is fine again :/

TorsteinHonsi (approved):

PR #15319, opened 2021-03-15
Added to context of axis label format/formatter

TorsteinHonsi:
Added `text` to axis label context, making it possible to insert the default formatting inside [labels.format](https://api.highcharts.com/highcharts/xAxis.labels.format).


bre1470 (approved):
Looks good! 👍 
Though some samples are failing, but maybe it comes from master.

PR #15329, opened 2021-03-16
bugfix/15005-dataGroup-point-start

### LINKED ISSUE #15005: Highstock chart point get wrong dataGroup start index when dataGrouping's options groupAll is true
#### expected behavior
get correct point.dataGroup.start index no matter zoom range is used or not

#### Actual behaviour
get wrong point.dataGroup.start when zoom range is selected. 

this problem only happens when dataGrouping setting groupAll: true, if groupAll:false, both scenario get corrected index.

https://jsfiddle.net/98t1v4cp/3/

#### Product version
lastest

#### Affected browser(s)
bug

karolkolodziej:
Fixed #15005, point `dataGroup` start was not calculated correctly.


ts/Core/Series/Series.ts
```diff
@@ -4265,7 +4267,16 @@ class Series {
...
-                point.dataGroup = (series.groupMap as any)[i];
+                if (series && series.groupMap) {
+                    for (j; j < series.groupMap.length; j++) {
+                        if ((series.groupMap[j].start as number) + (series.groupMap[j].length as number) > cropStart) {
+                            groupCropStartIndex = j;
+                            j = series.groupMap?.length;
+                        }
+                    }
+                }
+
```
pawelfus, line 4278: The whole idea of calculating `groupCropStartIndex` was to get the correct `cropStart` (`cropStart` for grouped points, not for the original dataset).

Now, we get `series.cropStart` correct for the grouped points too, so can we reuse it? Something like this:
```js
let groupCropStartIndex: number = options.dataGrouping?.groupAll ? cropStart : 0;
```

pawelfus: Or, completely different solution (not touching Highcharts core, only Stock):

Instead of trying to find correct indexes, crop `groupMap` too?  so the only change would be ( in `DataGrouping.ts`):

```js
series.cropStart = croppedData.start; // #15005
```
=>
```js
series.groupMap = series.groupMap.slice(croppedData.start, croppedData.end + 1); // #15005
```

samples/unit-tests/series/datagrouping/demo.js
```diff
@@ -827,6 +827,97 @@ QUnit.test('When groupAll: true, group point should have the same start regardle
     );
 });
 
+QUnit.test('When groupAll: true, group point should have the same start regardless of axis extremes (case with cropStart), #15005.', function (assert) {
```
pawelfus, line 830: Can we reuse the previous demo for this extra use-case? The demo is defined a few lines above new one

pawelfus (changes requested):

ts/Core/Series/Series.ts
karolkolodziej: I think the `series.groupMap` should remain untouched and inside all groups should exist (even the ones which are currently invisible).
That is why I applied the first suggestion.

samples/unit-tests/series/datagrouping/demo.js
```diff
@@ -825,6 +825,76 @@ QUnit.test('When groupAll: true, group point should have the same start regardle
         chart.series[1].points[1].dataGroup.start,
         'When the groupAll: false, and new extremes don\t influence the group, the start should not be changed.'
     );
+
+    // Change the data set to check the group start,
+    // when the extremes overlap with the point x.
```
raf18seb, line 830: It might be useful to add an issue ticket's number here in comment to indicate what issue you are testing.

samples/unit-tests/series/datagrouping/demo.js
```diff
@@ -825,6 +825,76 @@ QUnit.test('When groupAll: true, group point should have the same start regardle
...
+            [1611138383000, 0.835],
+            [1611159135000, 0.77],
+            [1611162097000, 0.7825],
+            [1611162097000, 0.7825]
+        ]
+    }, false);
+    chart.series[1].remove();
+
+    const point = chart.series[0].points[10],
+        pointX = point.x;
```
raf18seb, line 870: You defined 2 variables that you are using only once, and below, you are using `chart.series[0].points[9]` few times. You could either use variables more efficiently, or (which I prefer in this case) you could not use variables at all.

raf18seb (commented):
Just minor suggestions about the code cleanliness. I'm not familiar with dataGrouping that much.

raf18seb (approved):
Good job 👍

TorsteinHonsi (approved):
Thanks! Approved, awaiting @pawelfus and @pawellysy 's approvals.

pawelfus (approved):
Approved! 👍

PR #15330, opened 2021-03-16
TS typesafing - partial root option and required options.chart

TorsteinHonsi:
Getting my feet wet on making the top options partial. Beginning by making `Options.chart` a required property.


bre1470 (approved):
Thank you! Looks very good. 👍

PR #15338, opened 2021-03-17
Tools: fixed `unauthorized` error from demo CI job

goransle:
Fixed the CI job trying to use a context the nightly bot does not have access to. Also resolved a few problems with the script itself, and added a way to pass arguments to the deployer (`-dryrun` mostly).


TorsteinHonsi (approved):
Thanks!

PR #15340, opened 2021-03-17
bugfix/14906-split-tooltip-followPointer

### LINKED ISSUE #14906: Split outside tooltip w/ followPointer creates flashing/hiding tooltip
I've searched issues and can't find anything that mentions this. Might be semi related to [#9590](https://github.com/highcharts/highcharts/issues/9590)
#### Expected behaviour
Tooltip is rendered normally no matter the `followPointer` or `followTouchMove` settings

#### Actual behaviour
Tooltip flickers on move but disappears entirely while just hovering

https://jsfiddle.net/wy19gumq/71/

#### Product version

v8.2.2

#### Affected browser(s)
Chrome (latest)
* Disappears entirely in Chrome 
Safari (latest)
* Stays on screen but gets cut off in Safari
Potentially more

khlieng:
Fixed #14906, split tooltip flickered with `followPointer` enabled.

### DIFF
--- a/samples/unit-tests/tooltip/split/demo.js
+++ b/samples/unit-tests/tooltip/split/demo.js
@@ -184,7 +184,8 @@ QUnit.test('Split tooltip with useHTML and outside', function (assert) {
         tooltip: {
             split: true,
             useHTML: true,
-            outside: true
+            outside: true,
+            followPointer: true
         }
     });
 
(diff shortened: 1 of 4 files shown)


TorsteinHonsi (changes requested):
Thanks! We also need to mention [in the API](https://api.highcharts.com/highcharts/tooltip.followPointer) that `followPointer` doesn't apply when `split` is true.

TorsteinHonsi (approved):

PR #15341, opened 2021-03-17
Typesafe Utilities.extend

TorsteinHonsi:
As a general rule, the second argument of `extend` should be a subset/partial of the first. Exceptions are when two objects are merged to create a third type, in which case we can use an explicit type.

With this PR some new `any` casts are added. Previously these were inferred any's. It is better with explicit any's that we can systematically refactor out, than implicit any's that allowed us to extend typed objects with anything.


bre1470 (approved):
Looks good! 👍

PR #15351, opened 2021-03-18
Typesafe merge

TorsteinHonsi:
Typesafe `merge` when the first parameter is `true`.

### DIFF
--- a/ts/Core/Axis/GridAxis.ts
+++ b/ts/Core/Axis/GridAxis.ts
@@ -1034,7 +1034,7 @@ class GridAxis {
                     'year',
                     null
                 ]]
-            }, userOptions);
+            } as Highcharts.AxisOptions, userOptions);
 
             // X-axis specific options
             if (this.coll === 'xAxis') {
--- a/ts/Core/Axis/GridAxis.ts
+++ b/ts/Core/Axis/GridAxis.ts
@@ -953,11 +953,11 @@ class GridAxis {
      */
     public static onAfterSetOptions(
         this: Axis,
-        e: { userOptions: Highcharts.AxisOptions }
+        e: { userOptions: DeepPartial<Highcharts.AxisOptions> }
     ): void {
         var options = this.options,
             userOptions = e.userOptions,
-            gridAxisOptions: Highcharts.AxisOptions,
+            gridAxisOptions: DeepPartial<Highcharts.AxisOptions>,
             gridOptions: GridAxis.Options = (
                 (options && isObject(options.grid)) ? (options.grid as any) : {}
             );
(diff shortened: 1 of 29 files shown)


ts/Core/Axis/GridAxis.ts
```diff
@@ -1034,7 +1034,7 @@ class GridAxis {
                     'year',
                     null
                 ]]
-            }, userOptions);
+            } as Highcharts.AxisOptions, userOptions);
```
bre1470, line 1038: I played a bit with the generic typing to make this working with `merge<Highcharts.AxisOptions>(true,…` and failed. My conclusion is here that TypeScript is right about the typing. With the forced type the problem only gets ignored, that you do in-place merge on a new created object which does not contain all defaults.
- Therefor the correct solution would be to simply do a regular merge without in-place flag here.
- Also `userOptions` and `gridAxisOptions` should be deep partial.
```ts
        e: { userOptions: DeepPartial<Highcharts.AxisOptions> }
    ): void {
        var options = this.options,
            userOptions = e.userOptions,
            gridAxisOptions: DeepPartial<Highcharts.AxisOptions>,
            gridOptions: GridAxis.Options = (
                (options && isObject(options.grid)) ? (options.grid as any) : {}
            );

        if (gridOptions.enabled === true) {

            // Merge the user options into default grid axis options so
            // that when a user option is set, it takes presedence.
            gridAxisOptions = merge<DeepPartial<Highcharts.AxisOptions>>({

                className: (
```

ts/Extensions/Annotations/Annotations.ts
```diff
@@ -1631,8 +1631,8 @@ H.extendAnnotation = function <T extends typeof Annotation> (
     merge(
         true,
         Constructor.prototype,
-        BaseConstructor.prototype,
-        prototype
+        BaseConstructor.prototype as any,
+        prototype as any
     );
```
bre1470, line 1636: This should use `extend` not `merge` imho, because merge works recursive.
```ts
    extend(
        Constructor.prototype,
        merge(
            BaseConstructor.prototype,
            prototype
        )
    );
```

bre1470 (changes requested):
The type change of `merge` looks good! 👍 

The usage needs some attention...

ts/Core/Axis/GridAxis.ts
```diff
@@ -966,7 +966,7 @@ class GridAxis {
 
             // Merge the user options into default grid axis options so
             // that when a user option is set, it takes presedence.
-            gridAxisOptions = merge(true, {
+            gridAxisOptions = merge<DeepPartial<Highcharts.AxisOptions>>(true, {
```
bre1470, line 969: This should be no in-place merge. -> no `true`

bre1470 (changes requested):

ts/Core/Axis/GridAxis.ts
TorsteinHonsi: Why not? We are creating a new object, so instead of creating another object we might just as well merge the `userOptions` into our local defaults?

bre1470: You are right.

bre1470 (approved):

PR #15352, opened 2021-03-18
bugfix/12116-boost-stacked-rangeselector

### LINKED ISSUE #12116: Boost + rangeSelector with selected range + stacking = breaks the chart
#### Behaviour
If rangeSelector is active and selected range (for init) is not full and stacking is enabled then an error crashes the boosted chart.

https://jsfiddle.net/BlackLabel/4x6bwq5v/

1. Open and run the demo --> get an error in the browser console

#### Product version
7.2.0

#### Workaround(s)
Setting `getExtremesFromAll` to `true` for the series in the chart resolves the problem.

khlieng:
Fixed #12116, boosted stacked chart with range selector enabled and a partial range selected threw.

### DIFF
--- a/ts/Extensions/Boost/BoostInit.ts
+++ b/ts/Extensions/Boost/BoostInit.ts
@@ -225,6 +225,10 @@ function init(): void {
                     chartDestroyed = typeof chart.index === 'undefined',
                     isYInside = true;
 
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #15364, opened 2021-03-19
bugfix/11493-13095-tooltip-bounds

### LINKED ISSUE #11493: Y Axis scaled using set Extremes Is inappropriate

### LINKED ISSUE #13095: Tooltip shows outside of plot area for scrollablePlotArea

khlieng:
Fixed #11493, #13095, tooltip showed outside axis and plot bounds.

Closes #13095.


ts/Core/Pointer.ts
```diff
@@ -1965,9 +1965,14 @@ class Pointer {
         // Update positions (regardless of kdpoint or hoverPoint)
         } else if (followPointer && tooltip && !tooltip.isHidden) {
             anchor = tooltip.getAnchor([{} as any], e);
-            tooltip.updatePosition(
-                { plotX: anchor[0], plotY: anchor[1] } as any
-            );
+            if (
```
Izothep, line 1968: It looks that we are not using the `anchor` variable in any other place than the one inside of visible if block so we should be able to simply move the second if inside the first one:

```
followPointer &&
tooltip && !tooltip.isHidden &&
tooltip.isInsideX(anchor[0]) &&
tooltip.isInsideY(anchor[1])
```

ts/Core/Tooltip.ts
```diff
@@ -1370,8 +1376,14 @@ class Tooltip {
         } else {
             // update text
             if (tooltip.split) {
-                this.renderSplit(text as any, splat(pointOrPoints));
-            } else {
+                this.renderSplit(text as any, points);
+            } else if (
+                chart.polar ||
+                (
```
Izothep, line 1382: Shouldn't we check, if point's series has axes? This might help with current problems of tooltip not displaying for some series not using axes

ts/Core/Tooltip.ts
```diff
@@ -1788,6 +1800,92 @@ class Tooltip {
...
+    /**
+     * Checks if a point is within vertical visible plot and axis bounds.
+     *
+     * @private
+     * @param {number} plotY
+     * @param {Highcharts.Series} series
+     * @param {boolean} paneCoordinates
+     * @return {boolean}
+     */
+    public isInsideY(plotY: number, series?: Series, paneCoordinates?: boolean): boolean {
```
Izothep, line 1855: It looks that `isInsideY` and `isInsideX` are almost the same methods so maybe we can create one common method for both X and Y?

Izothep (changes requested):

ts/Core/Tooltip.ts
khlieng: Its just going to check if its within the plot if there are no axes, but which current problems?

ts/Core/Pointer.ts
khlieng: But then you need to move the `getAnchor` outside without really gaining a whole lot.

Izothep: I am not sure if we need to move it outside - I cannot find any other place where it is used

ts/Core/Tooltip.ts
Izothep: It looks that the tooltip is not displaying for Windbarb. Also, it is not displaying for Sankey in some specific cases (but maybe it was planned :) )

Izothep: It looks that the tooltip is not displaying for Windbarb. Also, it is not displaying for Sankey in some specific cases (but maybe it was planned :) )

ts/Core/Pointer.ts
Izothep: Yes, that is right. So we can leave it as it is for now :)

ts/Core/Tooltip.ts
khlieng: I think the visual diffs are mostly fine now, they are either related to points that are outside or due to the inverted and node fixes in `getChartCoordinatesFromPoint`.

Izothep: Ok, that's good. I would only check the windbarb graph - just in case (points in windbarb are rendered outside of axis extremes if I remember correctly). Apart of that all is good :)

ts/Core/Tooltip.ts
```diff
@@ -1788,6 +1813,79 @@ class Tooltip {
...
+     * Checks if a point is within visible plot and axis bounds.
+     *
+     * @private
+     * @param {number|boolean} plotX
+     * @param {number} plotY
+     * @param {Highcharts.Series} series
+     * @param {boolean} paneCoordinates
+     * @return {boolean}
+     */
+    public isInside(plotX: (number|boolean), plotY: number, series?: Series, paneCoordinates?: boolean): boolean {
```
TorsteinHonsi, line 1826: Better stay within `max-len` 80 when we can.

TorsteinHonsi (changes requested):
Nice work! See the inline comments.

ts/Core/Tooltip.ts
```diff
@@ -1370,51 +1380,69 @@ class Tooltip {
...
+                let checkY = y;
 
-                // Prevent the tooltip from flowing over the chart box (#6659)
-                if (!(options.style as any).width || styledMode) {
-                    label.css({
-                        width: this.chart.spacingBox.width + 'px'
-                    });
+                if (mouseEvent && currentSeries.is('xrange')) {
+                    checkX = mouseEvent.chartX - chart.plotLeft;
+                    checkY = mouseEvent.chartY - chart.plotTop;
```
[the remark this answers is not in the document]
khlieng: The anchor (`tooltipPos`) can be outside the plot while parts of the point is inside, so I check if the pointer is inside instead, maybe `directTouch` handles this too?

ts/Core/Tooltip.ts
```diff
@@ -1370,51 +1380,69 @@ class Tooltip {
...
-                        'highcharts-color-' +
-                        pick(
-                            (point as any).colorIndex,
-                            currentSeries.colorIndex
-                        )
-                    );
+                if (
+                    chart.polar ||
+                    this.isInside(checkX, checkY, currentSeries, !point.tooltipPos) ||
+                    currentSeries.is('windbarb') // Windbarb is outside plot
```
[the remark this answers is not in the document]
khlieng: So `directTouch` just seems to be a flag set once on the series prototype and never changed, there is also `Pointer.isDirectTouch` but thats going to break the scrollablePlotArea part of this.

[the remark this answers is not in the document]
TorsteinHonsi: Okay, here's a case that is not a windbarb: https://jsfiddle.net/highcharts/8q2vyzfk/ . Hover the top column. The thing is that this should have a tooltip because it is directly touched. In v9.0.1 it works.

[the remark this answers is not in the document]
TorsteinHonsi: Still there's the `currentSeries.is('windbarb')` check...

TorsteinHonsi:
Now pending:
- My two inline comments 
- The sankey and org chart positioning of visual diffs seems legit, though I am not sure what causes it. Please approve if valid.
- The diff in _stock/tooltip/positioner_ is dangerous. Since this is an approach that we have recommended in one of our demos, it will probably cause tooltips to disappear in many client's production environments. Why is it hidden? And maybe we should simply not evaluate `isInside` when `tooltip.options.positioner` is set, similar to what we do when `series.options.clip` is false?

khlieng:
The sankey and org positioning is caused by using `shapeArgs` instead of `plotX/Y` for nodes in `getChartCoordinatesFromPoint`, since nodes have no proper `plotX/Y`.

The stock/tooltip/positioner one is because the point karma calls `onMouseOver` on (Sep 28) is outside, the first visible point is Sep 29. So you would not be able to make this tooltip show through normal interaction with the chart.

ts/Core/Pointer.ts
```diff
@@ -763,18 +763,30 @@ class Pointer {
...
+
+            if (
+                (point as Highcharts.NodesPoint).isNode &&
+                shapeArgs &&
+                isNumber(shapeArgs.x) &&
+                isNumber(shapeArgs.y)
+            ) {
+                x = shapeArgs.x;
+                y = shapeArgs.y;
+            }
```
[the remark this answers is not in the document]
khlieng: I dont see any good way to combine them.

TorsteinHonsi:
Thanks! Now only the hard-coded windbarb check remaining.

TorsteinHonsi:
Thanks again for the changes. I'm being super critical now, but the tooltip/pointer stuff is a part of the code that has undergone several refactorings over the years, and we need to be extremely cautious with. 

Instead of adding another series property, `hasTooltipBounds`, I am wondering if we can use the `Series.isPointInside` function. In that case the windbarb series can have its own implementation of that function, checking only the X dimension. Then, going further with that thought:
* Are there other ways the new Tooltip logic can piggy-back the `Series.isPointInside` function or the `Point.isInside` property?
* And is the `Series.isPointInside` function safe to use with scrollable plot area?

I know I'm increasing the scope of this PR now, but we need to view this in a wholistic way.

khlieng:
That would require replacing the `Series.isPointInside` implementation with the new one from `Toolltip.isInside`, but that would change the meaning `Point.isInside` and break things. How about pulling `isInside` out of Tooltip, putting it in Chart (its needed in places that have no series), adding a new function (`shouldShowTooltip`?) to Series that calls it, and then Windbarb can override it and skip a dimension.

TorsteinHonsi:
Yes, we could do that. Though `Chart` already has a `isInsidePlot` function that is frequently used. But again, that function may also be flawed by `scrollablePlotArea`. So all in all, I think all of this must be revisited. Scrollable plot area is a relatively new feature. We must make sure these `isInside` functions are consistent, that we don't repeat ourselves, and that scrollable plot area is properly handled throughout.

TorsteinHonsi (changes requested):
Okay, we now have `Chart.isInsidePlot` and `Chart.isInside` that live side by side. That's not intuitive, as they partly overlap. I think the final step now is to merge the new `isInside` function into an enhanced `isInsidePlot` that takes scrollable plot area and axis length into account. 

My proposed path forward:
1. Create an override for `isInsidePlot` that takes an _options object_ as the third parameter. The existing third parameter is `inverted`, which actually shouldn't be necessary because it is already available as `this.inverted`. We should do a check of the use cases to see if we need to keep the boolean for of the parameter.
2. Now see what parameters we need to pass for the options. Initial thoughts:
```ts
{
   series // because the new `isInside` logic uses it
   ignoreX // instead of having plotX as (number|true)
   ignoreY // instead of having plotY as (number|true)
}
```
3. We can probably use the existance of `options.series` to decide whether we should use the axis postions or fall back to `plotWidth` and `plotHeight`.
4. Last, go over all existing uses of `isInsidePlot` to see if we should make use of any new functionality. For example, in [this fiddle](https://jsfiddle.net/highcharts/shcg4rd8/), the data label for an off-axis point is rendered. In this case, the data label logic should use the enhanced `isInsidePlot` function with the `series` option. The tooltip though is already fixed with this PR.
5. Also check that the implementations of `afterIsInsidePlot` events make sense with the new enhanced function.

khlieng:
So the main use cases of `inverted` is polar always passing `false` and data labels passing `isCartesian && chart.inverted`.

Theres also some differences between how `inverted` works in `isInside` and `isInsidePlot`. `isInsidePlot` needs to swap `plotX` and `plotY` because they come from points, while `isInside` only needs to swap axes since it gets passed tooltip anchors or mouse coordinates. Should there be an `isPoint` option too? 🤔

khlieng:
I'm not able to reproduce the new visual diff locally, even when loading the same scripts and using the same default options as karma it still works fine.

ts/Core/Series/Series.ts
```diff
@@ -7269,6 +7269,30 @@ class Series {
...
 
+    /**
+     * Checks if a tooltip should be shown for a given point.
+     *
+     * @private
+     * @param {number|true} plotX
+     * @param {number|true} plotY
+     * @return {boolean}
+     */
+    public shouldShowTooltip(
```
TorsteinHonsi, line 7280: Instead of `{number|true}`, maybe it would be cleaner here too to follow the pattern from `Chart.isInsidePlot` and have pure numeric `plotX` and `plotY`, then `options` as the third argument? Then the third argument can just be extended with `series` and possibly `visiblePlotOnly`, and everything passed on to `Chart.isInsidePlot`.

ts/Core/Chart/Chart.ts
```diff
@@ -622,28 +622,87 @@ class Chart {
...
+                scrollLeft,
+                scrollTop
+            } = {
+                scrollLeft: 0,
+                scrollTop: 0
+            }
+        } = this;
+
+        const series = options.series;
+        const box = (options.visiblePlotOnly && scrollablePlotBox) || plotBox;
```
TorsteinHonsi, line 652: It seems that the new `visiblePlotOnly` property is passed frequently. Is it more often true or more often false? If it is more often `true`, maybe we can make `true` the default? It would save bytes and be more backwards compatible?

TorsteinHonsi:
> I'm not able to reproduce the new visual diff locally

Me neither. It is probably a pixel perfection thing, let's not spend more time on it.

ts/Core/Chart/Chart.ts
khlieng: About 12 out of 34 calls pass it.

khlieng:
https://github.com/highcharts/highcharts/blob/ce7bc5ffdbd79047b599e10338413d96741a8e9b/ts/Extensions/Pane.ts#L558-L571 is the only usage of `afterIsInsidePlot`, should I just replace it with a wrap? Should perform better.

TorsteinHonsi:
> should I just get rid of the event and replace it with a wrap

In general we're trying to get rid of wraps and use events instead, so let's keep it. Events are quite performant.

TorsteinHonsi (approved):
Thanks for your patience and good work!

PR #15365, opened 2021-03-19
Demo/blog update options

nasin:
Hi, this is the first demo that will load on the homepage after the animation. The goal of the demo is to show how easy it is to manipulate options, and it's also meant to be fun.


TorsteinHonsi (changes requested):
Thanks, good idea!

Some minor issues
- For the pie chart, the _slice_ is a boolean, and no matter what I do with the slider, the slices remain sliced. It's a bit confusing.
- For the column chart, the "border" option confused me, until I realized it's actually the border _radius_ that is being changed. The columns don't have a visible border. Maybe it's better to call it "rounding" or something.
- With the line chart, the slider's initial position doesn't reflect the initial line width or marker size, so when sliding up, the line width or marker size actually increases.
- Also, in general, instead of updating the chart with animation _after_ the slider is changed, consider updating it during dragging, but without animation. It feels more responsive. See samples like css[/em](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/css/em/), [3d column interactive](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/demo/3d-column-interactive/) or [labels-autorotation-default](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/xaxis/labels-autorotation-default/).

TorsteinHonsi (changes requested):
It seems that #15365 and #15402 both contain (more or less) the same changes for the same files. Please make sure that the PRs don't interfere with each other. New PRs should be forked from the master branch, not a PR branch.

Also, please note my comments above ☝️

nasin:
Sorry for the confusion. I'm still working on your first set of comments (and learning git.)

TorsteinHonsi:
Thanks! You now did the changes locally, but did you push to GitHub? I can't see any changes in the online repo.

nasin:
Sorry about that. The correct files should be there now.

samples/highcharts/blog/blog-update-options/demo.css
```diff
@@ -0,0 +1,127 @@
+.container {
```
TorsteinHonsi, line 1: I think all CSS rules should be namespaced. I see in the utils that the styles of the demo affects the rest of the page, and we risk the same on the web page:

`.highcharts-demo h5 ...`
`.highcharts-demo .container.controls ...`

samples/highcharts/blog/blog-update-options/demo.css
```diff
@@ -0,0 +1,127 @@
...
+  transition:all 100ms;
+}
+h5{
+  font-size: 16px;
+  font-weight: bold;
+  display: block;
+  color:#333;
+  margin-bottom: 5px;
+}
+.cContainer{
```
TorsteinHonsi, line 14: Use kebab case like we usually do.

samples/highcharts/blog/blog-update-options/demo.js
```diff
@@ -0,0 +1,787 @@
...
+    borderRadius: 5,
+    borderWidth: 0,
+    allowOverlap: true,
+    dragDrop: {
+        draggableY: false,
+        draggableX: false,
+        dragHandle: {
+            lineWidth: 0
+        }
+    }
```
TorsteinHonsi, line 84: _Probably_ not needed as we are not loading the `drag-drop` module on this page?

samples/highcharts/blog/blog-update-options/demo.js
```diff
@@ -0,0 +1,787 @@
...
+let seriesType = 'item';
+
+const controls = [
+    ['startAngle', 'endAngle', 'innerSize', 'size'],
+    ['innerSize', 'size', 'slices', 'animation'],
+    ['xAxis', 'yAxis', 'width', 'radial'],
+    ['xAxis', 'yAxis', 'lineWidth', 'marker']
+];
+const controlLabels = [
+    ['start angle', 'end angle', 'inner aize', 'size'],
```
TorsteinHonsi, line 117: Typo: "inner aize"

samples/highcharts/blog/blog-update-options/demo.js
```diff
@@ -0,0 +1,787 @@
...
+                                startAngle: physics.angle
+                            }
+                        }
+                    }
+                );
+
+            }
+            if (physics.prevAngle <= physics.angle) {
+                button.disabled = false;
+            }
```
TorsteinHonsi, line 710: Instead of building the physics from scratch, we should use (and demonstrate how to use) our built-in animation for this. Plus it is much shorter. The approach is to define an easing function on the Math object, and call that from the animation options: https://jsfiddle.net/highcharts/cyLub6nm/

TorsteinHonsi (changes requested):
Thanks! In addition to the inline comments, I am also wondering if animation of the xAxis max and yAxis max for line/column charts is useful. Can we find something more useful to demonstrate?

One suggestion is to make the X axis a `datetime` axis (which according to some sources 75% of all cartesian charts have), then animate the `tickPixelInterval`. That would demonstrate one of Highcharts' core capabilites.

Are there other settings we can change? Linear/logarithmic axis perhaps? But that is an on/off setting. What about making the line actually a area and animating the fill opacity of the area fill?

nasin:
Hi, I changed some of the demos. I added a 3d area chart as the last series type. Then for the column chart, I added zones and plot band demos. I made the pie a variable pie.

samples/highcharts/blog/blog-update-options/demo.js
TorsteinHonsi: Did you decide to remove the spinning pie altogether?

TorsteinHonsi:
Thanks! I noticed to errors now:
* The `+` and `-` buttons stopped working. Only the slider itself works now.
* It seems like the `min` and `max` of the input elements don't have an effect. The 3d area has overlapping labels when setting the alpha to max, and I tried to change it in the HTML, but nothing changed. Actually the HTML has max="30", while the rendered slider goes to 100.

nasin:
Hi, I believe I fixed the issues with the "update options" demo. In this PR, I also added 3 simple homepage demos for demonstrating CSS styling, themes and I took a codepen map demo from the live blog and made it into a sample. I also fixed some issues with the timing of the homepage animation.

TorsteinHonsi (approved):
Great work! Approved, awaiting CI.

PR #15375, opened 2021-03-22
Updated series.line.marker.radius in Highcharts

KrishnaKumarHariprasannan:
The docs had the default value for stock charts but not for highcharts - this commit fixed this discrepancy.


TorsteinHonsi (approved):
Thanks! For some reason it seems the docs builder is not able to pick up the actual value.

PR #15376, opened 2021-03-23
bugfix/10464-boost-halo-axis-pos

### LINKED ISSUE #10464: Highcharts boost module renders markers on wrong Y-Axis on Zoom
#### Expected behaviour
The markers should be rendered on the same Y-Axis as the the point.

#### Actual behaviour
Markers are rendered on a different Y-Axis.

https://jsfiddle.net/o1pzc5jg/

Zoom anywhere on the chart such that a couple of hours of data is shown.  Observe the markers for the line chart drawn on the Y-Axis of the scatter chart.

#### Product version
Highchart 6.1.4, Highchart 7.1.0

#### Affected browser(s)
Chrome, Safari, Firefox
Fiddle Screenshot

khlieng:
Fixed #10464, halo position was wrong when using boost with multiple axes.

### DIFF
--- a/ts/Extensions/Boost/BoostInit.ts
+++ b/ts/Extensions/Boost/BoostInit.ts
@@ -476,6 +476,9 @@ function init(): void {
                     (xAxis && xAxis.pos !== prevX) ||
                     (yAxis && yAxis.pos !== prevY)
                 ) {
+                    // #10464: Keep the marker group position in sync with the
+                    // position of the hovered series axes since there is only
+                    // one shared marker group when boosting.
                     chart.markerGroup.translate(xAxis.pos, yAxis.pos);
 
                     prevX = xAxis.pos;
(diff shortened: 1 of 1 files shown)


TorsteinHonsi:
What is the root cause of the problem? Is it that the marker group wasn't translated properly in the first place? Or is it that we have only one chart-level marker group versus the normal series-level marker group?

Can we fix the root cause so we don't have to do the translation twice?

If not, please add a verbose comment to explain the situation.

TorsteinHonsi (changes requested):
What is the root cause of the problem? Is it that the marker group wasn't translated properly in the first place? Or is it that we have only one chart-level marker group versus the normal series-level marker group?

Can we fix the root cause so we don't have to do the translation twice?

If not, please add a verbose comment to explain the situation.

khlieng:
Its because of only one chart-level group.

TorsteinHonsi (changes requested):
> If not, please add a verbose comment to explain the situation.

... and add a reference to the issue number, so future us can understand what this peculiar code does.

TorsteinHonsi (approved):

PR #15379, opened 2021-03-23
bugfix/15377-inactive-state-zero-opacity

### LINKED ISSUE #15377: Network graph: dataLabel inactive opacity doesn't allow 0
#### Expected behaviour
Zero accepted as **opacity for dataLabels** in inactive state the same way as markers.

#### Actual behaviour
0 works like 1 for dataLabels. Markers are working fine.

Close to zero value works on hover:
https://jsfiddle.net/BlackLabel/ckovs632/5/

Zero doesn't work:
https://jsfiddle.net/BlackLabel/ckovs632/6/

#### Product version
9.0.1

#### Affected browser(s)
all

khlieng:
Fixed #15377, inactive state with 0 opacity applied due to `inactiveOtherSeries` being `true` did not work for data labels and connectors.

### DIFF
--- a/samples/unit-tests/series-pie/datalabel/demo.js
+++ b/samples/unit-tests/series-pie/datalabel/demo.js
@@ -7,6 +7,11 @@ QUnit.test('Pie data labels general tests', function (assert) {
                 {
                     animation: false,
                     type: 'pie',
+                    states: {
+                        inactive: {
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #15385, opened 2021-03-23
bugfix/15382-polar-zero-column

### LINKED ISSUE #15382: Polar plot showing columns, when there is no data present
#### Expected behavior
Wind Rose Polar plot should not show visual columns, when there is no data for this "wind-sector".
Columns not correctly plotted for interval 20-29 m/s looks very similar in structure to data used from Data Interval 0-9 m/s.

#### Actual behavior
When choosing Data interval 2 (in this case 20-29 m/s), the first 5 columns shows a green visualization, even though it has no data. Both the JSON object and when hovering the actual column "0" data is reported (eg: the column "N" reports 0%).

https://jsfiddle.net/ham28y1p/2/

#### Product version
 Version? I use the CDN, so the latest version?
WindRose2

#### Affected browser(s)
Only tested with chrome.
WindRose

khlieng:
Fixed #15382, updating polar column to zero value did not work.

### DIFF
--- a/samples/unit-tests/svgrenderer/symbol/demo.js
+++ b/samples/unit-tests/svgrenderer/symbol/demo.js
@@ -155,13 +155,7 @@ QUnit.test('Symbol tests', function (assert) {
 });
 
 QUnit.test('Arc', assert => {
-    const ren = new Highcharts.Renderer(
-        document.getElementById('container'),
-        400,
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Nice, thanks!

PR #15388, opened 2021-03-24
bugfix/15384-flag-update-shape

### LINKED ISSUE #15384: flag.shape can't be dynamically update()'
#### Expected behaviour
flag.shape to change on update() 

#### Actual behaviour
flag.shape is unchanged

https://jsfiddle.net/BlackLabel/cm6zh231/
https://jsfiddle.net/doxtry1w/
https://www.highcharts.com/forum/viewtopic.php?f=12&t=46158&sid=dfb89f23e05530ff9f9ae9a2e5c5e6f3

#### Product version

#### Affected browser(s)
Chrome / all

khlieng:
Fixed #15384, updating flag shape did not work.

### DIFF
--- a/ts/Series/Flags/FlagsPoint.ts
+++ b/ts/Series/Flags/FlagsPoint.ts
@@ -80,6 +80,14 @@ class FlagsPoint extends ColumnSeries.prototype.pointClass {
         return isNumber(this.y) || typeof this.y === 'undefined';
     }
 
+    /**
+     * @private
+     */
+    public hasNewShapeType(): (boolean|undefined) {
+        const shape = this.options.shape || this.series.options.shape;
+
+        return this.graphic && shape && shape !== this.graphic.symbolKey;
+    }
 }
 
 /* *
--- a/ts/Series/Flags/FlagsSeries.ts
+++ b/ts/Series/Flags/FlagsSeries.ts
@@ -420,7 +420,7 @@ class FlagsSeries extends ColumnSeries {
                 !outsideRight
             ) {
                 // #15384
-                if (graphic && shape && shape !== graphic.symbolKey) {
+                if (graphic && point.hasNewShapeType()) {
                     graphic = graphic.destroy();
                 }
 
--- a/ts/Series/Flags/FlagsSeries.ts
+++ b/ts/Series/Flags/FlagsSeries.ts
@@ -421,7 +421,7 @@ class FlagsSeries extends ColumnSeries {
             ) {
                 // #15384
                 if (graphic && shape && shape !== graphic.symbolKey) {
-                    graphic = point.graphic = graphic.destroy();
+                    graphic = graphic.destroy();
                 }
 
                 // Create the flag
--- a/ts/Series/Flags/FlagsSeries.ts
+++ b/ts/Series/Flags/FlagsSeries.ts
@@ -419,6 +419,10 @@ class FlagsSeries extends ColumnSeries {
                 (plotX as any) >= 0 &&
                 !outsideRight
             ) {
+                // #15384
+                if (graphic && shape && shape !== graphic.symbolKey) {
+                    graphic = point.graphic = graphic.destroy();
+                }
 
                 // Create the flag
                 if (!graphic) {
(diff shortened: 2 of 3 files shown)


ts/Series/Flags/FlagsSeries.ts
```diff
@@ -419,6 +419,10 @@ class FlagsSeries extends ColumnSeries {
                 (plotX as any) >= 0 &&
                 !outsideRight
             ) {
+                // #15384
+                if (graphic && shape && shape !== graphic.symbolKey) {
+                    graphic = point.graphic = graphic.destroy();
+                }
```
TorsteinHonsi, line 425: We have a point method for this, called `hasNewShapeType`. Plus, we don't have to unset `point.graphic` because it is reassigned directly after. So this check can be shortened to:

```ts
                // #15384
                if (graphic && point.hasNewShapeType()) {
                    graphic = graphic.destroy();
                }
```

TorsteinHonsi (changes requested):

ts/Series/Flags/FlagsSeries.ts
khlieng: I dont think it can be used for this, not in its current state at least, in it `oldShapeType` is `g`, `this.shapeType` is `rect`, so it always returns true and neither of them have anything to do with `series.options.shape`/`point.options.shape` in this case.

pawelfus: We had a similar issue with 3D, where point's graphic is a group: `<g>`, but shapeType is `rect`. 3D overwrites `hasNewShapeType` and adds its own check. If we want to follow the convention, then `point.hasNewShapeType()` should be defined for Flag series too

ts/Series/Flags/FlagsPoint.ts
```diff
@@ -80,6 +80,14 @@ class FlagsPoint extends ColumnSeries.prototype.pointClass {
...
     }
 
+    /**
+     * @private
+     */
+    public hasNewShapeType(): (boolean|undefined) {
+        const shape = this.options.shape || this.series.options.shape;
+
+        return this.graphic && shape && shape !== this.graphic.symbolKey;
+    }
```
TorsteinHonsi, line 90: This is not necessary, we [already have this on Point class](https://github.com/highcharts/highcharts/blob/8a4bd3653771624aefb1b4cf9f2d9e51bab7dac2/ts/Core/Series/Point.ts#L882-L893).

TorsteinHonsi (changes requested):

ts/Series/Flags/FlagsPoint.ts
khlieng: https://github.com/highcharts/highcharts/pull/15388#discussion_r600432760

ts/Series/Flags/FlagsSeries.ts
TorsteinHonsi: Actually I think we should come up with a general check that applies in all cases. The idea of the `hasNewShapeType` function is to determine when we need to generate a new graphic in cases like casting series type or switching between 2D and 3D. The 3D override seems to be too categoric since it always returns false when the graphic is grouped. Basically the idea is to return false only when the existing type can _not_ be animated into the new type. What do you think @pawelfus ?

pawelfus: > Basically the idea is to return false only when the existing type can not be animated into the new type.

We don't have morphism implemented (only [v7 demo](https://jsfiddle.net/BlackLabel/dhrxcs1m/8/)), so at this moment none of the shapes can be animated into the new one, but correct me if I'm wrong. We use `hasNewShapeType` method to check if the new shape was set and if yes, then destroy the old graphic. All the other visual aspects can be animated (color, radius, width etc.).

Regarding the solution, maybe instead of checking the node's name, we can check `point.shapeType`?  In 3d we have `arc3d` and `cuboid`. The flags series inherits `shapeType = rect` from column series. While every flag is a label, we can still use `shapeType`. This should be smart enough and safe too. The only problem I see, is probably with `shapeType` being changed before we draw a shape (`hasNewShapeType()` called too late?).

TorsteinHonsi (approved):

ts/Series/Flags/FlagsSeries.ts
TorsteinHonsi: But we still have the flags, where I don't think the shape is reflected in `shapeType`. I'll merge this at the current state, then we can revive the problem when we get to morphing.

PR #15390, opened 2021-03-24
Samples: Faster topojson demo

TorsteinHonsi:
The old demo was slow because
- We loaded a 2.3 MB topojson map, which is overkill for web.
- For each of the ~30k points, we ran `proj4` with the projection string, so the string had to be parsed every time. With the change, we run projection of each point on the pre-compiled projection. Projection time of the original 2.3 MB map went down from ~450ms to ~30ms.


oysteinmoseng (approved):
Fantastic!

PR #15393, opened 2021-03-24
Updated custom-component a11y demo.

oysteinmoseng:
Demo was not updated to work with latest a11y module changes, discovered after a customer had a related question.


TorsteinHonsi (approved):
Thanks!

PR #15394, opened 2021-03-24
feature/15139-disparity-index-indicator

raf18seb:
Added new technical indicator, Disparity Index (DI). See #15139.


raf18seb:
Merge with master is on the way (after code review)

samples/stock/indicators/disparity-index/demo.js
```diff
@@ -0,0 +1,44 @@
...
+            data: data
+        }, {
+            type: 'disparityindex',
+            linkedTo: 'aapl',
+            yAxis: 1,
+            params: {
+                period: 9
+            }
+        }]
+
```
karolkolodziej, line 42: Unnecessary space

ts/Stock/Indicators/DisparityIndex/DisparityIndexOptions.d.ts
```diff
@@ -0,0 +1,34 @@
...
+ *  Declarations
+ *
+ * */
+
+export interface DisparityIndexOptions extends SMAOptions {
+    params?: DisparityIndexParamsOptions;
+}
+
+export interface DisparityIndexParamsOptions extends SMAParamsOptions {
+    average: 'sma'|'ema'|'dema'|'tema'|'wma';
```
karolkolodziej, line 31: Shouldn't we place there only the type of the `average: string`?

ts/Stock/Indicators/DisparityIndex/DisparityIndexIndicator.ts
```diff
@@ -0,0 +1,282 @@
...
+
+    /* *
+     *
+     *  Functions
+     *
+     * */
+
+    public init(this: DisparityIndexIndicator): void {
+        const args = arguments,
+            ctx = this, // Disparity Index indicator
```
karolkolodziej, line 119: Why `ctx`?

ts/Stock/Indicators/DisparityIndex/DisparityIndexIndicator.ts
```diff
@@ -0,0 +1,282 @@
...
+    /* *
+     *
+     *  Properties
+     *
+     * */
+
+    public data: Array<DisparityIndexPoint> = void 0 as any;
+    public options: DisparityIndexOptions = void 0 as any;
+    public points: Array<DisparityIndexPoint> = void 0 as any;
+
```
karolkolodziej, line 110: Shouldn't the `averageIndicator` also be declared here?

ts/Stock/Indicators/DisparityIndex/DisparityIndexIndicator.ts
```diff
@@ -0,0 +1,282 @@
...
+        const averageType = params.average,
+            index = params.index,
+            period = params.period as any,
+            xVal: Array<number> = (series.xData as any),
+            yVal: Array<number>|Array<Array<number>> = (series.yData as any),
+            yValLen: number = yVal ? yVal.length : 0,
+            disparityIndexPoint: Array<Array<number>> = [],
+            xData: Array<number> = [],
+            yData: Array<number> = [],
+            averageIndicator: any = this.averageIndicator,
```
karolkolodziej, line 178: I think you might improve the types here.

karolkolodziej: As I think about it more thoroughly it might not be that easy. In the beginning, I thought that it's only the string, but I think we should accept `any` here.

ts/Stock/Indicators/DisparityIndex/DisparityIndexIndicator.ts
```diff
@@ -0,0 +1,282 @@
...
+            values: disparityIndexPoint,
+            xData: xData,
+            yData: yData
+        } as IndicatorValuesObject<TLinkedSeries>;
+    }
+
+}
+
+interface DisparityIndexIndicator {
+    averageIndicator?:
```
karolkolodziej, line 231: I'm not sure where it should be declared, maybe @bre1470 might help?

ts/Stock/Indicators/DisparityIndex/DisparityIndexIndicator.ts
```diff
@@ -0,0 +1,282 @@
...
+        typeof EMAIndicator|
+        typeof DEMAIndicator|
+        typeof SMAIndicator|
+        typeof TEMAIndicator|
+        typeof WMAIndicator;
+    pointClass: typeof DisparityIndexPoint;
+}
+
+extend(DisparityIndexIndicator.prototype, {
+    nameBase: 'Disparity Index'
```
karolkolodziej, line 241: _Discusion:_
We should also show in the tooltip and legend the type of `average` we are using

ts/Stock/Indicators/DisparityIndex/DisparityIndexIndicator.ts
```diff
@@ -0,0 +1,282 @@
...
+                function (indicator: Highcharts.Indicator): undefined {
+                    indicator.prototype.init.apply(ctx, args);
+                    return;
+                }
+            );
+        }
+    }
+
+    public calculateDisparityIndex(
+        close: number,
```
karolkolodziej, line 159: If this indicator allows changing the `index` of the price for a certain point in time, maybe use some more universal name like `currentPrice` for example?

karolkolodziej (changes requested):
Great work!

docs/stock/technical-indicator-series.md
```diff
@@ -75,7 +75,7 @@ Click [here](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree
 yAxis bindings
 -------------
 
-Some technical indicators (Bollinger Bands, EMA, Ichimoku Kinko Hyo, Pivot Points, Price Envelopes, PSAR, SMA, VbP, VWAP, WMA, Zig Zag) can be placed on the same yAxis as the main series. However, other indicators (A/D, ATR, CCI, CMF, DMI, MACD, MFI, Momentum, RoC, RSI, Stochastic) should use a separate yAxis. This is caused by values calculated by algorithms: yAxis extremes for the main series can be <250, 255> but for the Stochastic technical indicator, values are within <0, 100> extremes. A technical indicator can be placed on a separate yAxis as any other series:
+Some technical indicators (Bollinger Bands, EMA, Ichimoku Kinko Hyo, Pivot Points, Price Envelopes, PSAR, SMA, VbP, VWAP, WMA, Zig Zag) can be placed on the same yAxis as the main series. However, other indicators (A/D, ATR, CCI, CMF, Disparity Index, DMI, MACD, MFI, Momentum, RoC, RSI, Stochastic) should use a separate yAxis. This is caused by values calculated by algorithms: yAxis extremes for the main series can be <250, 255> but for the Stochastic technical indicator, values are within <0, 100> extremes. A technical indicator can be placed on a separate yAxis as any other series:
```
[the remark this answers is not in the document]
raf18seb: What do you mean by saying "the other part"?

samples/stock/indicators/disparity-index/demo.js
raf18seb: Do you mean "enter"? Line break? Is there a 2-lines break or a 1-line break? If 1-line, then it is intended.

ts/Stock/Indicators/DisparityIndex/DisparityIndexOptions.d.ts
raf18seb: I think that both are fine, however, `'sma'|'ema'|'dema'|'tema'|'wma'` type is more accurate.

docs/stock/technical-indicator-series.md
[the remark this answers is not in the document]
karolkolodziej: The table/list of indicators.

ts/Stock/Indicators/DisparityIndex/DisparityIndexOptions.d.ts
pawelfus: I believe `someRandomString` is an invalid type of `average` 🤔

samples/stock/indicators/disparity-index/demo.js
karolkolodziej: yep- 1-line break- I accept that it might be intentional ;)

ts/Stock/Indicators/DisparityIndex/DisparityIndexIndicator.ts
pawelfus: Good point!

ts/Stock/Indicators/DisparityIndex/DisparityIndexIndicator.ts
raf18seb: `ctx` means "context". In this case, the main context is the Disparity Index indicator. The variable is not called `indicator` because this name is used later in
```js
function (indicator: Highcharts.Indicator): undefined {
    indicator.prototype.init.apply(ctx, args);
    return;
}
```

It's not my idea/pattern, but it's named same way in every indicator that uses `IndicatorRequired.js`. If I'm not wrong, you did exactly the same in your indicator :) 

image

But I also added a comment that indicates the context `// Disparity Index indicator`

karolkolodziej: Yes, I did it exactly the same, but as you can see I didn't think about that as you!
Thanks for explaining!

ts/Stock/Indicators/DisparityIndex/DisparityIndexIndicator.ts
raf18seb: Yes, I thought about it. The problem I encountered is that I would have to define
```js
extend(DisparityIndexIndicator.prototype, {
    nameBase: some_name
});
```
in 2 different places. I send a commit with a code showing how would it look like now.

ts/Stock/Indicators/DisparityIndex/DisparityIndexIndicator.ts
```diff
@@ -0,0 +1,282 @@
...
+            break;
+        case 'tema':
+            averageIndicator = TEMAIndicator;
+            break;
+        case 'wma':
+            averageIndicator = WMAIndicator;
+            break;
+        default: // use sma if any of the above strings do not match
+            averageIndicator = SMAIndicator;
+        }
```
pawelfus, line 140: Could we simplify this to: `averageIndicator = seriesTypes[averageType]`? Or `averageIndicator = seriesTypes[averageType] || SMAIndicator` ?

docs/stock/technical-indicator-series.md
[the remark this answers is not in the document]
raf18seb: Yes, I didn't change it. I left it for #15456.
Should I leave the above too?

I thought the problem was only with editing the table. In the above case, editing is very easy.

ts/Stock/Indicators/DisparityIndex/DisparityIndexIndicator.ts
```diff
@@ -0,0 +1,282 @@
...
+            range = (3 * period - 2) - 1;
+            break;
+        case 'ema':
+        case 'wma':
+        default: // use sma if any of the above strings do not match
+            range = period - 1;
+        }
+
+        // Check period, if bigger than points length, skip
+        if (!index || yVal.length <= range) {
```
pawelfus, line 202: Is `index = 0` incorrect?

ts/Stock/Indicators/DisparityIndex/DisparityIndexIndicator.ts
pawelfus: Could you elaborate? `averageIndicator` can be either undefined or one of the predefined other Indicator types, right?

ts/Stock/Indicators/DisparityIndex/DisparityIndexIndicator.ts
```diff
@@ -0,0 +1,282 @@
...
+            range = (2 * period - 1) - 1;
+            break;
+        case 'tema':
+            range = (3 * period - 2) - 1;
+            break;
+        case 'ema':
+        case 'wma':
+        default: // use sma if any of the above strings do not match
+            range = period - 1;
+        }
```
pawelfus, line 199: Do we need this checking? Instead of manually checking the period, can we base on what's returned from `getValues()`? If `getValues()` returns `undefined`, then we don't have enough points => return undefined;

We can get the first point from: 

```js
const start = xVal.indexOf(getValues().xData[0]);
for (let i = start; i < yValLen ; i++) { ... }
```

ts/Stock/Indicators/DisparityIndex/DisparityIndexIndicator.ts
raf18seb: Here is a commit: https://github.com/highcharts/highcharts/pull/15394/commits/24d8d6cf44cc13f341a6e12ebca9139261710ec0
As you can see, `nameBase` needs to be defined in 2 places:
- before indicator init, because it needs to be displayed in popup (otherwise it will be shown as `DISPARITYINDEX` instead of `Disparity Index`
image
https://github.com/highcharts/highcharts/blob/22fd2b1e8b995f5ce7cfad4ee96c5db4d3898dd6/ts/Extensions/Annotations/Popup.ts#L896-L905
- and while initing an indicator, when we know which average has been chosen.

Is it OK now?

pawelfus (changes requested):
- Doesn't work in StockTools GUI: https://jsfiddle.net/BlackLabel/2z30acdw/ (EMA works fine)
- My comments regarding `switches` are ideas for simplifying future maintenance. For example, adding a new average type indicator (e.g. Welles Winder or Hull), all we need to change are docs and TS. No need to dive into the indicator's algorithm. The same when someone decides to create a custom algorithm and wants to use it in Disparity Index

ts/Stock/Indicators/DisparityIndex/DisparityIndexIndicator.ts
```diff
@@ -0,0 +1,286 @@
...
+            averageIndicator = WMAIndicator;
+            break;
+        default: // use sma if any of the above strings do not match
+            averageIndicator = SMAIndicator;
+        }
+
+        ctx.averageIndicator = averageIndicator;
+
+        extend(DisparityIndexIndicator.prototype, {
+            nameBase: 'Disparity Index (' + averageType + ')'
```
pawelfus, line 145: Use `nameComponents`, so the resulting format will match other technical indicators:
` Name (paramA, paramB, ... , paramC)` => `Disparity Index (14, EMA)`

ts/Stock/Indicators/DisparityIndex/DisparityIndexIndicator.ts
pawelfus: The average type should not be defined in the list in the popup. It's a parameter, so only tooltip/legend should include this info

ts/Stock/Indicators/DisparityIndex/DisparityIndexIndicator.ts
raf18seb: So obvious...

ts/Stock/Indicators/DisparityIndex/DisparityIndexIndicator.ts
raf18seb: Nice idea. I'll check it

pawellysy (approved):
Nothing to add! 👍

ts/Stock/Indicators/DisparityIndex/DisparityIndexIndicator.ts
```diff
@@ -0,0 +1,246 @@
+/* *
+ *  (c) 2010-2021 Rafal Sebestjanski
+ *
+ *  Disparity Index indicator for highstock
```
pawelfus, line 4: - technical indicator*
- Highcharts Stock*

ts/Stock/Indicators/DisparityIndex/DisparityIndexIndicator.ts
```diff
@@ -0,0 +1,246 @@
...
+     * @requires     stock/indicators/disparity-index
+     * @optionparent plotOptions.disparityindex
+     */
+    public static defaultOptions: DisparityIndexOptions =
+    merge(SMAIndicator.defaultOptions, {
+        params: {
+            /**
+             * The average used to calculate the Disparity Index indicator.
+             * By default it uses SMA. To use other averages, e.g. EMA,
+             * the stock/indicators/ema.js file needs to be loaded.
```
pawelfus, line 84: stock/indicators/ema.js -> `stock/indicators/ema.js`

pawelfus (changes requested):
Overall great!
Just a minor polishing, sorry I missed that in the previous review..

ts/Stock/Indicators/DisparityIndex/DisparityIndexIndicator.ts
```diff
@@ -0,0 +1,246 @@
...
+interface DisparityIndexIndicator {
+    nameBase: string;
+    nameComponents: Array<string>;
+
+    pointClass: typeof DisparityIndexPoint;
+}
+
+extend(DisparityIndexIndicator.prototype, {
+    nameBase: 'Disparity Index',
+    nameComponents: ['period', 'average']
```
karolkolodziej, line 205: The `average` property is not displayed in the tooltip/legend

karolkolodziej (changes requested):
Two things to change: 
- `average` in tooltip
- I still can't add the `disparityIndex` through the StoockTools when the `average: 'sma'`

pawelfus (approved):
All good, thanks! 
PS: I can't recreate issues reported by Karol (tested here: https://jsfiddle.net/BlackLabel/2z30acdw/1/ )

raf18seb:
I've just tested, I can't recreate either.

karolkolodziej (approved):
Sorry for the confusion, my local branch was behind the main one and I didn't check that properly. Of course, everything works fine, sorry once again. ;)
The only thing left if the conflict now ;)

TorsteinHonsi (approved):
Thanks, nice work!

PR #15395, opened 2021-03-24
bugfix/15383-indicator-recalc

### LINKED ISSUE #15383: Indicators recalculate its values wrongly, when you change the type of the main series
#### Expected behaviour
When updating main series, the indicator should recalculate correctly, when you change the type.

#### Actual behaviour
The indicator doesn't take into consideration the new type of the main series. It recalculates assuming that the main series is still the pevious type.

https://jsfiddle.net/BlackLabel/m1sf506h/

#### Product version
v9.0.1

#### Affected browser(s)
all

khlieng:
Fixed #15383, indicators did not recalculate when updating series type.

### DIFF
--- a/ts/Core/Series/Series.ts
+++ b/ts/Core/Series/Series.ts
@@ -6814,6 +6814,8 @@ class Series {
         series.initialType = initialType;
         chart.linkSeries(); // Links are lost in series.remove (#3028)
 
+        // #15383: Fire updatedData if the type has changed to keep linked
+        // series such as indicators updated
         if (casting && series.linkedSeries.length) {
             series.isDirtyData = true;
         }
--- a/ts/Core/Series/Series.ts
+++ b/ts/Core/Series/Series.ts
@@ -6634,6 +6634,8 @@ class Series {
             animation = series.finishedAnimating && { animation: false },
             kinds = {} as Record<string, number>;
 
+        newType = newType || initialType;
+
         if (keepPoints) {
             preserve.push(
                 'data',
@@ -6704,9 +6706,11 @@ class Series {
             delete (series as any)[prop];
         });
 
-        if (seriesTypes[newType || initialType]) {
+        let casting = false;
 
-            const casting = newType !== series.type;
+        if (seriesTypes[newType]) {
+
+            casting = newType !== series.type;
 
             // Destroy the series and delete all properties, it will be
             // reinserted within the `init` call below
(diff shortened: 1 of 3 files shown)


pawelfus (approved):
Looks good, thanks! 👍

ts/Core/Series/Series.ts
```diff
@@ -6763,6 +6767,9 @@ class Series {
 
         series.init(chart, options);
 
+        if (casting) {
+            series.isDirtyData = true;
+        }
```
TorsteinHonsi, line 6772: Since this bug fix only applies to indicators that need to recalculate the data, I'm concerned that `isDirtyData` is not necessary for other series. I think this is more precise:
```ts
if (casting && series.hasDerivedData) {
    series.isDirtyData = true;
}
```

TorsteinHonsi (changes requested):

ts/Core/Series/Series.ts
khlieng: This wont work since its the indicator that has `hasDerivedData` set, but its not the indicator that changes type and the indicator relies on the series its bound to firing `updatedData`.

khlieng: Maybe checking `linkedSeries` could work?

pawelfus: Hm.. there's also antoher solution. Instead of marking series after an update with `isDirtyData` flag, technical indicators could listen to `series.update()` events and check if `type` and `data` changed in any way.

But still, it is strange that `isDirtyData` flag was not set when updating the series with `{ type, data }` object. Most of the technical indicators recalculate only on `updatedData` event (Volume By Price is recalculated on `afterSetExtremes`) .  Perhaps the best solution is to check: `casting && dataChange`?

khlieng: `isDirtyData` was only set when the series already had points previously and casting clears the points. Checking `casting && dataChange` sounds like it would break the `setData + update({ type })` case.

TorsteinHonsi (approved):

ts/Core/Series/Series.ts
```diff
@@ -6811,6 +6814,10 @@ class Series {
         series.initialType = initialType;
         chart.linkSeries(); // Links are lost in series.remove (#3028)
 
+        if (casting && series.linkedSeries.length) {
```
TorsteinHonsi, line 6817: Nice! Now we just need a small comment and a reference to the issue.

TorsteinHonsi (changes requested):

TorsteinHonsi (approved):

PR #15396, opened 2021-03-24
Tools: corrected samples CI script

goransle:
The CI job was incorrectly set to run the `check_and_trigger_docs...` script.


TorsteinHonsi (approved):
Thanks!

PR #15397, opened 2021-03-24
bugfix/15392-boost-empty-data-turbo

### LINKED ISSUE #15392: Error 12 on empty array when boosted
#### Expected behaviour
No error - empty array shouldn't be a problem.

#### Actual behaviour
When boost is enabled through `seriesThreshold` a series with an empty `data` array is flagged through error 12.

https://jsfiddle.net/BlackLabel/f79j21nd/4/
No steps required - the error shows on init.

#### Product version
v9.0.1

khlieng:
Fixed #15392, boosted chart with empty data showed turbo warning.


TorsteinHonsi (approved):
Thanks!

PR #15402, opened 2021-03-25
Demo/blog landing page

nasin:
Hi, Marketing requested that I remove the grayscale filter from the graphs.


TorsteinHonsi (changes requested):
It seems that #15365 and #15402 both contain (more or less) the same changes for the same files. Please make sure that the PRs don't interfere with each other. New PRs should be forked from the master branch, not a PR branch.

For this PR, I think only the removal of the grayscale filter really belongs here. Since this is a minor change, the fastest solution may be just to delete this PR and continue with #15365 that already contains the removal of the grayscale.

nasin:
Ok, thanks. I'll do that.

nasin:
Closing this because changes are already included in an existing PR.

PR #15403, opened 2021-03-25
Tools: Updated packages and activated linting of no-optional-chain.

bre1470:


TorsteinHonsi (approved):
Great! That paid off in the ES5 bundle sizes.

PR #15404, opened 2021-03-25
 Fixed scope and typing of Utilities.getStyle function.

bre1470:
Basically the function becomes compatible to ES import.

### DIFF
--- a/ts/Core/Utilities.ts
+++ b/ts/Core/Utilities.ts
@@ -1834,6 +1834,16 @@ function getNestedProperty(path: string, parent: unknown): unknown {
     return parent;
 }
 
+function getStyle(
+    el: HTMLDOMElement,
+    prop: string,
+    toInt: true
+): number;
+function getStyle(
+    el: HTMLDOMElement,
+    prop: string,
+    toInt?: false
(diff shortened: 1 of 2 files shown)


ts/Core/Utilities.ts
```diff
@@ -1850,16 +1860,17 @@ function getNestedProperty(path: string, parent: unknown): unknown {
...
 function getStyle(
     el: HTMLDOMElement,
     prop: string,
     toInt?: boolean
-): (number|string) {
+): (number|string|undefined) {
+    const customGetStyle: typeof getStyle = ((H as any).getStyle || getStyle);
 
-    var style;
+    let style: (number|string|undefined) = (toInt ? NaN : void 0);
```
TorsteinHonsi, line 1873: I can't imagine a case where we actually want to use `NaN`. Would it be a lot of overhead if we return `number|undefined` and check validity in the implementations?

bre1470: `parseInt` is also returning NaN, so I see it as the natural value for failed integer parsing.

TorsteinHonsi (approved):
Great, thanks!

PR #15408, opened 2021-03-26
bugfix/15406-responsive-mapnav

### LINKED ISSUE #15406: mapNavigation breaks with responsive rule
#### Expected behaviour
I want to hide `mapNavigation` buttons on mobile devices which seems to be pretty straightforward:
```
responsive: {
    rules: [{
        condition: { maxWidth: 576 },
        chartOptions: {
            mapNavigation: {
                enabled: false,
            },
        }
    }]
},
```

#### Actual behaviour
- If initial width is under the responsive breakpoint - the map stops responding to resize events and the following error is printed in the console:
```
TypeError: Cannot read property 'alignedObjects' of undefined
    at v.H.align (highstock.js:52)
    at a.<anonymous> (map.js:38)
    at highstock.js:19
    at q (highstock.js:19)
    at a.onload (highstock.js:380)
    at a.firstRender (highstock.js:380)
    at a.<anonymous> (highstock.js:350)
    at q (highstock.js:19)
    at a.init (highstock.js:349)
```
- If initial width is over the responsive breakpoint - when the map gets resized down to the responsive breakpoint the `mapNavigation` buttons are hidden (which is correct) but when the map's width is back over the breakpoint they are in the wrong place

https://jsfiddle.net/veLmp4oy/1/

#### Product version
Highmaps 9.0.1

#### Affected browser(s)
All

khlieng:
Fixed #15406, responsive rule broke map navigation.

### DIFF
--- a/samples/unit-tests/maps/map-navigation/demo.js
+++ b/samples/unit-tests/maps/map-navigation/demo.js
@@ -104,7 +104,7 @@ QUnit.test('Map navigation button alignment', assert => {
     const chart = Highcharts.mapChart('container', {
         chart: {
             plotBorderWidth: 1,
-            width: 600
+            width: 400
         },
 
         mapNavigation: {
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #15417, opened 2021-03-28
Enhancement/a11y population dashboard

oysteinmoseng:
Added an accessible population dashboard demo on request for an a11y showcase on short notice. **NB: Please do not merge & delete before Easter, as the branch-specific URL is used for a demo.**

Not sure if this belongs in `/demo`, since it is a bit elaborate, but could be good to showcase somewhere perhaps - maybe with some more polishing and some input from Nancy on the design? We don't really have any elaborate a11y demos currently, especially that deal with changing data and dashboard-style interactions.

Also fixed some mostly minor issues and made some enhancements that were discovered in the process.

One issue was a regression in master that had caused the keyboard focus indicator to break (https://github.com/highcharts/highcharts/commit/e49d4830f98fcb3d2ec847f91bc83dbbccb630bc), so that should probably be included before the next release, even if this PR does not get ready by then.

Screen Shot 2021-03-28 at 00 44 36

### DIFF
--- a/samples/highcharts/accessibility/population-dashboard/demo.js
+++ b/samples/highcharts/accessibility/population-dashboard/demo.js
@@ -1,0 +1,550 @@
+var populationHistory = {
+    // 1984, 1989, 1994, 1999, 2004, 2009, 2014, 2019
+    se: [
+        8336605, 8492964, 8780745, 8857874, 8993531, 9298515, 9696110, 10285453
+    ],
+    dk: [
+        5111619, 5131594, 5206180, 5321799, 5404523, 5523095, 5643475, 5818553
+    ],
+    fi: [
+        4881803, 4964371, 5088333, 5165474, 5228172, 5338871, 5461512, 5520314
+    ],
+    no: [
+        4140099, 4226901, 4336613, 4461913, 4591910, 4828726, 5137232, 5347896
+    ],
+    is: [
+        239511, 252852, 266021, 277381, 292074, 318499, 327386, 361313
+    ],
+    gl: [
+        52700, 55300, 55500, 56100, 56911, 56323, 56295, 56225
+    ],
+    fo: [
+        44891, 47166, 45856, 46436, 47600, 47808, 47960, 48678
+    ]
+};
+var areaData = {
+    se: 450295,
+    no: 385207,
+    dk: 42933,
+    fi: 338440,
+    is: 103000,
+    fo: 1399,
+    gl: 2166000
+};
+var topCities = {
+    se: [
+        ['Göteborg', 599011],
+        ['Stockholm', 1515017],
+        ['Malmö', 316588]
+    ],
+    no: [
+        ['Bergen', 420000],
+        ['Oslo', 1000467],
+        ['Stavanger', 222697]
+    ],
+    fi: [
+        ['Tampere', 341696],
+        ['Helsinki', 1305893],
+        ['Turku', 277677]
+    ],
+    dk: [
+        ['Aarhus', 280534],
+        ['Copenhagen', 1330993],
+        ['Odense', 180302]
+    ],
+    is: [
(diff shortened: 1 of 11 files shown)


samples/highcharts/accessibility/population-dashboard/demo.js
```diff
@@ -0,0 +1,550 @@
...
+        width: 150,
+        type: 'solidgauge'
+    },
+    pane: {
+        startAngle: 0,
+        endAngle: 360,
+        background: [{
+            outerRadius: '115%',
+            innerRadius: '95%',
+            backgroundColor: Highcharts.color('#FDD089')
```
TorsteinHonsi, line 172: Use `rgba` instead

samples/highcharts/accessibility/population-dashboard/demo.js
```diff
@@ -0,0 +1,550 @@
...
+        color: {
+            pattern: {
+                path: 'M 0 1.5 L 5 1.5 M 0 4 L 5 4',
+                backgroundColor: '#FDD089',
+                color: '#FF8F99',
+                width: 5,
+                height: 5
+            }
+        }
+    });
```
TorsteinHonsi, line 397: All this should be available natively using `point.select()`. The pattern would then be set in the chart configuration, as the `select` state. So unless there is a bug somewhere, this code is overengineered?

TorsteinHonsi (changes requested):
Thanks for the PR! 

Yes it would be nice as a general dashboard demo. We don't have a good demo for tying together chart components like this.

Some comments:
* The export menu only exports the geo map, which is a bit counter intuitive, I was expecting the full dashboard. Maybe better to disable exporting?
* Map zoom - is it necessary, or just redundant noise? It also has a cost in terms of blocking mousewheel/touch scroll.
* The map and details work both ways, so I can select an item in one component and the others get updated. Not so with the table. I would expect the map and details to update when I click a row in the table.
* Along the same lines, I would expect the table row to be highlighted when a country is selected in the map or details?

samples/highcharts/accessibility/population-dashboard/demo.js
TorsteinHonsi: It works, programmatically and through UI: https://jsfiddle.net/highcharts/79gydo31/

oysteinmoseng: Agreed - The only reason for not using `point.select` was the lack of support for it in the a11y module vs `point.update`, but this is something we should fix now that we don't have a deadline. Also we need to block selecting multiple points by holding down Shift - should we have an API option for this?

oysteinmoseng:
Thanks! Comments to the general comments:

- Agreed on the expectations of the exporting. Being able to export the map as SVG is very useful for a11y, though (tactile graphics), so maybe I should look into either exporting the whole dashboard (even as separate files if needed), or changing the wording in the menu.
- Map zoom: Agree, can probably just be removed, unless we want to highlight the feature.
- Table: Much agreed.

samples/highcharts/accessibility/population-dashboard/demo.js
TorsteinHonsi: Yes, we have the `select` event with `e.accumulate`: https://api.highcharts.com/class-reference/Highcharts.PointInteractionEventObject.

We can cancel all select event where the UI passes `e.accumulate = true`: https://jsfiddle.net/highcharts/79gydo31/1/

TorsteinHonsi:
> maybe I should look into either exporting the whole dashboard (even as separate files if needed), or changing the wording in the menu.

Yes, either of those. If you want to have a go at full export, there's a boilerplate at https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/exporting/multiple-charts/

Issue #15199, opened 2021-02-23
lastVisiblePrice label should be customizable - docs


karolkolodziej:
#### Expected behaviour
Test that compatibility and add documentation under `lastVisiblePrice.label`.

#### Actual behaviour
It looks like [options](https://api.highcharts.com/highstock/yAxis.crosshair.label) for the `crosshair` label can be also applied to the `lastVisiblePrice` indicator label.

#### Live demo with steps to reproduce
https://jsfiddle.net/BlackLabel/xLzsaqkd/

#### Product version
Highstock JS v9.0.1 (2021-02-15)

PR #15418, opened 2021-03-29
docs/15199-lastVisiblePrice-label-docs

pawellysy:
Done #15199, Updated lastVisiblePrice label docs.

### DIFF
--- a/ts/Extensions/PriceIndication.ts
+++ b/ts/Extensions/PriceIndication.ts
@@ -93,7 +93,7 @@ declare global {
  * horizontal axes.
  *
  * @type      {Highcharts.AlignValue}
- * @since     7.0
+ * @since     7.0.0
  * @product   highstock
  * @apioption plotOptions.series.lastVisiblePrice.label.align
  *
@@ -104,7 +104,7 @@ declare global {
  * color, or `#666666` if that is not available.
  *
  * @type      {Highcharts.ColorType}
- * @since     7.0
+ * @since     7.0.0
  * @product   highstock
  * @apioption plotOptions.series.lastVisiblePrice.label.backgroundColor
  */
@@ -113,7 +113,7 @@ declare global {
  * The border color for the crosshair label
  *
  * @type      {Highcharts.ColorType}
- * @since     7.0
+ * @since     7.0.0
  * @product   highstock
  * @apioption plotOptions.series.lastVisiblePrice.label.borderColor
  */
@@ -123,7 +123,7 @@ declare global {
  *
  * @type      {number}
  * @default   3
- * @since     7.0
+ * @since     7.0.0
  * @product   highstock
  * @apioption plotOptions.series.lastVisiblePrice.label.borderRadius
  */
@@ -133,7 +133,7 @@ declare global {
  *
  * @type      {number}
  * @default   0
- * @since     7.0
+ * @since     7.0.0
  * @product   highstock
  * @apioption plotOptions.series.lastVisiblePrice.label.borderWidth
  */
@@ -152,7 +152,7 @@ declare global {
  * numeric axes and `{value:%b %d, %Y}` for datetime axes.
  *
  * @type      {string}
- * @since     7.0
+ * @since     7.0.0
  * @product   highstock
  * @apioption plotOptions.series.lastVisiblePrice.label.format
  */
@@ -161,7 +161,7 @@ declare global {
  * Formatter function for the label text.
  *
  * @type      {Highcharts.XAxisCrosshairLabelFormatterCallbackFunction}
- * @since     7.0
+ * @since     7.0.0
  * @product   highstock
  * @apioption plotOptions.series.lastVisiblePrice.label.formatter
  */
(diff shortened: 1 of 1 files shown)


ts/Extensions/PriceIndication.ts
```diff
@@ -108,6 +147,67 @@ declare global {
  * @apioption plotOptions.series.lastPrice.enabled
  */
 
+/**
+ * A format string for the crosshair label. Defaults to `{value}` for
+ * numeric axes and `{value:%b %d, %Y}` for datetime axes.
+ *
+ * @type      {string}
+ * @since     2.1
```
pawelfus, line 155: Wrong version

ts/Extensions/PriceIndication.ts
```diff
@@ -108,6 +147,67 @@ declare global {
...
+ * @product   highstock
+ * @apioption plotOptions.series.lastVisiblePrice.label.format
+ */
+
+/**
+ * Formatter function for the label text.
+ *
+ * @type      {Highcharts.XAxisCrosshairLabelFormatterCallbackFunction}
+ * @since     2.1
```
pawelfus, line 164: Wrong version

ts/Extensions/PriceIndication.ts
```diff
@@ -108,6 +147,67 @@ declare global {
...
+ * @product   highstock
+ * @apioption plotOptions.series.lastVisiblePrice.label.shape
+ */
+
+/**
+ * Text styles for the crosshair label.
+ *
+ * @type      {Highcharts.CSSObject}
+ * @default   {"color": "white", "fontWeight": "normal", "fontSize": "11px", "textAlign": "center"}
+ * @since     2.1
```
pawelfus, line 194: Wrong version

ts/Extensions/PriceIndication.ts
```diff
@@ -108,6 +147,67 @@ declare global {
...
+ */
+
+/**
+ * The line marks the last price from all points.
+ *
+ * @sample {highstock} stock/indicators/last-price
+ *         Last price
+ *
+ * @declare   Highcharts.SeriesLastPriceOptionsObject
+ * @product   highstock
```
pawelfus, line 206: Missing version

pawelfus (changes requested):
General question: Instead of defining options one by one, can we somehow reuse already-defined doclets from Crosshair? E.g. using `@extends`, like we do for series

Then we would need to define only those options, that were changed in the `last(Visible)Price` (`color`, `classname`, `snap`, `enabled`) config?

ts/Extensions/PriceIndication.ts
```diff
@@ -108,6 +147,68 @@ declare global {
  * @apioption plotOptions.series.lastPrice.enabled
```
karolkolodziej, line 147: If all options for the lastPrice are available from the v7.0.0, we might add that information to the `enabled` option.

TorsteinHonsi:
Pending @pawelfus ' approval. Specifically, was this discussed? 
 > General question: Instead of defining options one by one, can we somehow reuse already-defined doclets from Crosshair? E.g. using @extends, like we do for series

pawellysy:
> Pending @pawelfus ' approval. Specifically, was this discussed?
> 
> > General question: Instead of defining options one by one, can we somehow reuse already-defined doclets from Crosshair? E.g. using `@extends`, like we do for series

@pawelfus @torsteinHonsi I added the `@extends yAxis.crosshair` option to the docs but some of the API options description needed to be updated, hence some of them are defined one more time.

pawelfus (changes requested):
Something is still wrong with versions:
Zrzut ekranu 2021-04-7 o 17 30 45

raf18seb (approved):
Nothing to add.

ts/Extensions/PriceIndication.ts
```diff
@@ -74,17 +74,104 @@ declare global {
...
+ *
+ * @type      {number}
+ * @default   0
+ * @since     7.0
+ * @product   highstock
+ * @apioption plotOptions.series.lastVisiblePrice.label.borderWidth
+*/
+
+/**
+ * Padding inside the lastVisiblePrice label.
```
pawelfus, line 168: Apply option-name code formatting. Change:
> Padding inside the lastVisiblePrice label.

To:
> Padding inside the `lastVisiblePrice` label.

(applies to every option)

ts/Extensions/PriceIndication.ts
```diff
@@ -74,17 +74,104 @@ declare global {
...
+ *
+ * @type      {number}
+ * @default   8
+ * @since     7.0
+ * @product   highstock
+ * @apioption plotOptions.series.lastVisiblePrice.label.padding.
```
pawelfus, line 174: Change:
> `plotOptions.series.lastVisiblePrice.label.padding.`

To:
> `plotOptions.series.lastVisiblePrice.label.padding`

Otherwise we get:

Zrzut ekranu 2021-04-8 o 18 02 35

pawelfus (changes requested):
Just a minor polishing! :)

ts/Extensions/PriceIndication.ts
raf18seb: Clue: it is possible to generate local API using `npx gulp jsdoc-watch`. I know it takes time to generate, but it is worth checking at least once before committing.

pawelfus (approved):
Looks great, thank you! :)

TorsteinHonsi (approved):

PR #15419, opened 2021-03-29
bugfix/11336-tooltip-small-chart

### LINKED ISSUE #11336: Pie chart tooltip not displayed when hovering series
#### Expected behaviour

Tooltip should be displayed on all series like in version `7.0.3`.

#### Actual behaviour

Tooltip of first series is displayed once, then tooltip is never shown again for any series after hovering others series.

Not working in version `7.1.2`: https://jsfiddle.net/37v4ehbk/

Working in version `7.0.3`: https://jsfiddle.net/g9utn8py/

#### Product version

Highcharts version 7.1.2

#### Affected browser(s)

Tested in Firefox 67.0.4 and Chromium 75.0.3770.90

khlieng:
Fixed #11336, no tooltip showed when hovering very small chart.

### DIFF
--- a/samples/unit-tests/legend/legend-height/demo.js
+++ b/samples/unit-tests/legend/legend-height/demo.js
@@ -68,6 +68,12 @@
                 chart.legend.legendHeight < 150,
                 'Legend is less than full height'
             );
+
+            chart.setSize(20, 20);
+            assert.notOk(
+                chart.legend.nav,
+                '#11336: There should be no navigation if there is no space'
+            );
         }
     );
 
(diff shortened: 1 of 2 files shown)


TorsteinHonsi:
Thanks! Do you know what the visual diff is? It looks like the top of some text content. Could it be parts of the legend that should be clipped away? (And why does legend clipping affect the tooltip 🤔 ?)

khlieng:
Its the top of the legend, it was pushed further down by the nav buttons before. The tooltip didnt work because the nav button tracker covered the whole chart.

TorsteinHonsi (approved):

PR #15423, opened 2021-03-30
bugfix/15420-zones-clip-leak

### LINKED ISSUE #15420: Memory leak on chart.update when negativeColor is true
#### Expected behaviour

Calling `chart.update()` should not significantly (or at all) change the DOM count when updating series with identical lengths of data items.

#### Actual behaviour

When `negativeColor` is enabled on a series, the `chart.update()` call is causing new DOM elements to be created every time, and they are never cleaned up. For long-running applications this can eventually lead to an out of memory exception.

Upon closer inspection of the DOM tree, it appears that new clipPath elements are created on each call to `chart.update()`. The old clipPaths are never removed from the DOM tree. This only occurs when `negativeColor` is enabled; if it is not, the clipPath defs are cleaned up appropriately.

**Note**: It is possible this clipPath leak is happening with other options as well. The `negativeColor` option was just one that I observed.

https://jsfiddle.net/qj725cs3/

#### Product version

Highcharts 9.0.1

#### Affected browser(s)

Chrome 89.0.4389.90 (64-bit)

khlieng:
Fixed #15420, series with `zones` or `negativeColor` enabled leaked memory on `update`.

### DIFF
--- a/samples/unit-tests/series/zones/demo.js
+++ b/samples/unit-tests/series/zones/demo.js
@@ -284,9 +284,17 @@ QUnit.test('Adding and removing zones', function (assert) {
         ]
     });
 
+    const clips = chart.series[0].clips;
+
     chart.series[0].update({
         zones: []
     });
+
+    assert.strictEqual(
+        chart.series[0].clips,
+        clips,
+        '#15420: Clips array should have been preserved'
+    );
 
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #15427, opened 2021-03-31
bugfix/15412-passive-leftovers

### LINKED ISSUE #15412: SVGElement on Event Handlers - [Violation] Added non-passive event listener to a scroll-blocking event
#### Expected behaviour
Inspector does not complain about non-passive event listeners

#### Actual behaviour
Each rendered chart creates a Violation

#### Product version
HighCharts v9.0.1

#### Affected browser(s)
Chrome 89.0.4389.90

For some background, I have a dashboard that contains 6 gauges. Whenever the page loads the console shows a Violation for each of them, for several events.

I have followed the other related issues to this problem that were addressed with some recent changes, however it appears that the new mechanism for adding Events with the passive option is not applied by default to ColumnSeries, SVGElements?

I am using the es6 modules and was able to make a local edit to the `SVGElement.prototype.on` method in `Core/Renderer/SVG/SVGElement.js`.  If the device is not a touch screen with an event of type 'click', then the current default does the following: (~line 1480) 
`element['on' + eventType] = handler;`
This method is being called twice  from the **drawTacker method** in `Series/Column/ColumnSeries.js` (lines 574, 575)

If I assign U.addEvent to addEvent at the top of the file, and then use `addEvent(element, eventType, handler, { passive: false});` then I no longer see the error.  Just to note - I haven't tested to see if the events are still working properly, but this change does remove the Violations.

Even so, it appears that this case has not been handled properly.

Edit: As an observation, if the intent of this code is to respond to a user touching or mousing in/out the tracker in the dial/gauge, and showing the data for the tracker, then the events are in fact working, because I am seeing the tooltip appear/disappear when I mouse over and then away.

khlieng:
Fixed #15412, some charts showed non-passive scroll-blocking warnings.

### DIFF
--- a/samples/unit-tests/datalabels/overlapping/demo.js
+++ b/samples/unit-tests/datalabels/overlapping/demo.js
@@ -100,7 +100,7 @@ QUnit.test(
             ]
         });
 
-        chart.legend.allItems[1].legendGroup.element.onclick();
+        Highcharts.fireEvent(chart.legend.allItems[1].legendGroup.element, 'click');
         chart.xAxis[0].setExtremes(0.5);
 
         assert.strictEqual(
--- a/samples/unit-tests/drilldown/resetzoom/demo.js
+++ b/samples/unit-tests/drilldown/resetzoom/demo.js
@@ -180,7 +180,7 @@ QUnit.test('Drilldown and reset zoom', function (assert) {
     );
 
     // Click reset zoom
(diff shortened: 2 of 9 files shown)


raf18seb (approved):
Looks fine, thanks 👍

TorsteinHonsi (approved):
Thanks!

PR #15431, opened 2021-03-31
docs/15190-changelog-for-14883

KacperMadej:


TorsteinHonsi (changes requested):
Thanks!

The `changelog.html` file is generated, so the the change needs to be applied upstream, at [9.0.0.md](https://github.com/highcharts/highcharts/blob/master/changelog/highcharts/9.0.0.md). Otherwise the change will be nuked on next update.

TorsteinHonsi (approved):

PR #15444, opened 2021-04-06
bugfix/15438-polar-NaN

### LINKED ISSUE #15438: Radar chart with accessibility plugin generates "Expected length, NaN" error on null value.
#### Expected behaviour
* No console errors are generated when rendering a Radar chart with null values in the series with the accessibility module included.

#### Actual behaviour
* When creating a radar chart with the latest Highchart version, if the accessibility plugin is used and the series has null values, it will generate the following console error:

```
highcharts.js:50 Error: <rect> attribute x: Expected length, "NaN".
highcharts.js:50 Error: <rect> attribute y: Expected length, "NaN".
```

The original issue shown in this example:
https://jsfiddle.net/obcs95d3/

If you change the null value to a regular value the error is gone:
https://jsfiddle.net/obcs95d3/1/

If you remove the accessibility module but still have null value, the error is also gone:
https://jsfiddle.net/obcs95d3/2/

#### Product version

"highcharts": "9.0.1"

#### Affected browser(s)
* Chrome: 89.0.4389.90
* Safari:  14.0.3 (16610.4.3.1.7)

khlieng:
Fixed #15438, polar chart with null data and accessibility module loaded threw.

### DIFF
--- a/samples/unit-tests/chart/polar/demo.js
+++ b/samples/unit-tests/chart/polar/demo.js
@@ -1,6 +1,5 @@
-QUnit.test('Polar chart with no data (#5226)', function (assert) {
-    assert.expect(0);
-    Highcharts.chart('container', {
+QUnit.test('Polar chart data', function (assert) {
+    const chart = Highcharts.chart('container', {
         chart: {
             polar: true
         },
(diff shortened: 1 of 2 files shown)


TorsteinHonsi:
Thanks! I did a minor optimization to also avoid the unnecessary `postTranslate` call when we already know `plotY` is `undefined`. Now awaiting CI checks.

oysteinmoseng (approved):
No comment from me - except it seems the null dummy point handling in a11y module does not work properly with polar. We can treat that as a separate issue I think.

PR #15446, opened 2021-04-06
bugfix/15434-plotband-hidden-series

### LINKED ISSUE #15434: When series are hidden, the plotBand is still visible as a thin line
#### Expected behaviour
When you hide all of the series, the plotbands should be invisible.

#### Actual behaviour
The plotbands remain visible as a thin line and translate to the beginning of the axis.

Click on the legend to hide the series, and notice, that the plotBand remain visible with the width of 0.5 px at the beginning of the axis. 
https://jsfiddle.net/BlackLabel/Lqxf9e17/

#### Possible Workaround
https://jsfiddle.net/BlackLabel/e4tkdug2/

#### Product version
v9.0.1

#### Affected browser(s)
all

khlieng:
Fixed #15434, plot band was not completely hidden when hiding series.


TorsteinHonsi (approved):
Thanks!

PR #15451, opened 2021-04-06
bugfix/15445-stock-options-mutate

### LINKED ISSUE #15445: Stock constructor mutates options object
#### Expected behaviour
Constructor should not mutate the options.

http://jsfiddle.net/goransle/c7ep8nka/

Here, the second chart should be identical to the first.

#### Product version
Current master, works as expected on [latest release](http://jsfiddle.net/goransle/jcpg4oq2/)

khlieng:
Fixed #15445, options got mutated when creating stock chart.

I also got rid of the series hoop jumping in `Chart` since `merge` is just going to do `copy.series = original.series`, which is the same thing, so I dont see where the performance gain was supposed to come from.

### DIFF
--- a/samples/unit-tests/axis/overscroll/demo.js
+++ b/samples/unit-tests/axis/overscroll/demo.js
@@ -84,7 +84,7 @@ Highcharts.each([true, false], function (ordinal) {
                 xAxis.max - xAxis.min,
                 xAxis.series[0].options.data.length -
                     1 +
-                    options.xAxis[0].overscroll,
+                    xAxis.options.overscroll,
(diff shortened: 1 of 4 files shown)


pawelfus (approved):
Looks good! 👍 

> I also got rid of the series hoop jumping in Chart since merge is just going to do copy.series = original.series, which is the same thing, so I dont see where the performance gain was supposed to come from.

It's probably with us since day 1, perhaps `merge` used to work differently?

TorsteinHonsi (approved):
Looks good, thanks!

Issue #15370, opened 2021-03-22
Missing demos in /stock/demo pages 


pawelfus:
https://www.highcharts.com/demo/stock

At the bottom of the page, the "Flags and Technical indicators" category is missing (with all demos and thumbnails)

PR #15457, opened 2021-04-07
Added demo-config to samples folder

goransle:
Moved this config file from the highcharts-demo-manager repo. Related to #15370.

### DIFF
--- a/samples/demo-config.js
+++ b/samples/demo-config.js
@@ -1,0 +1,29 @@
+/**
+ * Config for the demo-site builder. Each key represents a page, i.e. 'Higcharts demos'.
+ * Select which categories or tags to include with `tags` or `categories`.
+ * `Filter` will further limit the selection to the specified tags (or categories)
+ * `Path` specifies the relative path the page. In other words, what comes after `highcharts.com/demo`.
+ * The key is also used for the page title
+ */
+module.exports = {
+    Highcharts: {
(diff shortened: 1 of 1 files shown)


karolkolodziej:
In the future, we might want to change the `name` in the demo details from `Highstock` to `Highcharts Stock` like [here](https://github.com/highcharts/highcharts/blob/12540607569b99467d91c447e23eb1d3b869e16d/samples/stock/indicators/ema/demo.details).
Will it make any difference for the demo builder?
FYI: @pawelfus

goransle:
@karolkolodziej  `name` is used mainly for the page title, so that would not break anything. The demos which appear on [https://highcharts.com/demo](highcharts.com/demo), are selected by `tags` ([example .details file](https://github.com/highcharts/highcharts/blob/497b60df028d6a123d753bcb0140b3a1b28a517b/samples/stock/demo/arearange/demo.details)). I already changed these to `Highcharts Stock demo`, so they should be OK :)

TorsteinHonsi (approved):
Thanks!

TorsteinHonsi:
@goransle Are the [samples/{product}/demo/index.htm](https://github.com/highcharts/highcharts/blob/feature/mapgeometry/samples/highcharts/demo/index.htm) files now redundant? In that case, they should be deleted.

goransle:
@TorsteinHonsi  Yes, they should be redundant now. I will check and see.

PR #15458, opened 2021-04-07
bugfix/15436-dynamic-colorAxis

### LINKED ISSUE #15436: Color axis not updating as expected
#### Expected behaviour
https://jsfiddle.net/9sbq2wLj/ - On directly loading chart

#### Actual behaviour
https://jsfiddle.net/b6tvk842/1/ - On updating the config

Added above

#### Product version
Highcharts 9

#### Affected browser(s)
Chrome

khlieng:
Fixed #15436, previous legend items remained visible after dynamically adding color axis.

I believe the case of not being able to add a color axis through `update` with `oneToOne=false` is working as intended.


TorsteinHonsi (approved):
Thanks!

PR #15466, opened 2021-04-08
Fixed #15447, wordcloud kbd nav issues.

### LINKED ISSUE #15447: Inconsistent behaviour for keyboard focus on wordcloud
#### Expected behaviour
Keyboard navigation works consistently for wordcloud similar to that of charts. That is, if container is set to receive focus first, it receives focus on first 'Tab' key press. Focus shifts to data point only upon next 'Tab' key press.

#### Actual behaviour
Upon navigating to wordcloud using 'Tab' key, focus is grabbed by data point even if container is set to receive focus first. This is seen in v8.2.0 but seems to be working fine at least partially in the latest version (note that data points never receive focus).

**Broken Version**
8 2 0-Broken-Version

**Partially Working Version**
9 0 1-Partially-Working-Version

Broken: https://jsfiddle.net/franklinefrancis/b7420k6w/
Partially Working: https://jsfiddle.net/franklinefrancis/nkdoq048/

#### Product version
8.2.0

#### Affected browser(s)
All browsers

oysteinmoseng:
Fixed #15447, wordcloud keyboard navigation, and added support for making the chart container a separate tab-stop.
___
Added support for using the `container` component in keyboard navigation, to make the container itself a tab-stop. This is done by adding `container` in the [`keyboardNavigation.order`](https://api.highcharts.com/highcharts/accessibility.keyboardNavigation.order) list.

### DIFF
--- a/ts/Accessibility/Components/ContainerComponent.ts
+++ b/ts/Accessibility/Components/ContainerComponent.ts
@@ -14,6 +14,7 @@ import type {
     SVGDOMElement
 } from '../../Core/Renderer/DOMElementType';
 import AccessibilityComponent from '../AccessibilityComponent.js';
+import KeyboardNavigationHandler from '../KeyboardNavigationHandler.js';
 import ChartUtilities from '../Utils/ChartUtilities.js';
 const {
     unhideChartElementFromAT,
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (approved):
Thanks!

PR #15469, opened 2021-04-08
Tools: Switched to native TS watch in scripts-watch task.

bre1470:


TorsteinHonsi (changes requested):
Wow, nice, that's way faster to work with!

How about also making this the default `gulp` task? As far as I understand, the default `gulp` is broken since it depends on the .js files being compiled beforehand. Before we started migration, `gulp` would start a watch on the source files.

TorsteinHonsi (approved):

PR #15478, opened 2021-04-09
samples/remove-demo-index

goransle:
Removed the `index.htm` pages from the `samples/*/demo/` folders. Updated some demo attributes which fell back to the `index.htm` if they were undefined. Also updated the precommit test, which now checks that the attributes in `demo.detail` are set to match `demo-config.js`.

Set priority for some demos that should be at the top of their categories. If we want to keep the order as-is, we could either save all the current priorities, or perhaps add an ordered list of demo names to `demo-config.js`


TorsteinHonsi:
My only concern with this change is that team members create new samples by copying one of the listed demos, and they accidentally end up in the demos on the website. And the other way around, they create new demo content and forget to tag them properly.

We could overcome that by:
a) Only show samples from `{product}/demo` in the website demos. (If that's the case already, ignore this :)
b) Throw a pre-commit error when a sample from `{product}/demo` is not tagged correctly.

goransle:
@TorsteinHonsi a) is the case at the moment. b) should be covered by the updates [here](https://github.com/highcharts/highcharts/blob/8a5f10052d6ee20e19ae345fecf9f1aeec45380c/tools/gulptasks/test.js#L77).

TorsteinHonsi (approved):
Ah, nice!

PR #15479, opened 2021-04-09
bugfix/15362-cleaning-indicators-params-docs

### LINKED ISSUE #15362: Stocktools indicator popup - unnecessary params for some indicators
#### Expected behaviour
Remove the `index` and `period` parameter from Stocktools (possibly also from documentation)
for indicators that don't use them.

#### Actual behaviour
For some indicators, there is an option inside the Stocktools indicator popup to add the `index` and `period` parameter.
Where some indicators like for example VBP doesn't need those.

https://api.highcharts.com/highstock/series.ad.params

#### Product version

#### Affected browser(s)
All

_Inernal note:_
Go through each indicator file, check listed `params` and compare them with Stocktools and documentation (add `excluding`).

raf18seb:
Fixed #15362, corrected API and Stock Tools popup list of indicators' params.

#### Upgrade note
The default value of [linearregression.params.xAxisUnit](https://api.highcharts.com/highstock/plotOptions.linearregression.params.xAxisUnit) has been changed from `undefined` to `null`.

### DIFF
--- a/ts/Stock/Indicators/Chaikin/ChaikinIndicator.ts
+++ b/ts/Stock/Indicators/Chaikin/ChaikinIndicator.ts
@@ -84,6 +84,12 @@ class ChaikinIndicator extends EMAIndicator {
              */
             volumeSeriesID: 'volume',
             /**
+             * Parameter used indirectly for calculating the `AD` indicator.
+             * Decides about the number of data points that are taken
+             * into account for the indicator calculations.
+             */
+            period: 9,
+            /**
              * Periods for Chaikin Oscillator calculations.
              *
              * @type    {Array<number>}
--- a/ts/Stock/Indicators/Chaikin/ChaikinIndicator.ts
+++ b/ts/Stock/Indicators/Chaikin/ChaikinIndicator.ts
@@ -73,9 +73,10 @@ class ChaikinIndicator extends EMAIndicator {
          * Paramters used in calculation of Chaikin Oscillator
          * series points.
          *
-         * @excluding index, period
+         * @excluding index
          */
         params: {
+            index: void 0, // unused index, do not inherit (#15362)
             /**
              * The id of volume series which is mandatory.
              * For example using OHLC data, volumeSeriesID='volume' means
--- a/ts/Stock/Indicators/LinearRegression/LinearRegression.ts
+++ b/ts/Stock/Indicators/LinearRegression/LinearRegression.ts
@@ -82,6 +82,9 @@ class LinearRegressionIndicator extends SMAIndicator {
                  * computing the slope and the intercept. This may enchance the
                  * legiblitity of the indicator's values.
                  *
+                 * Default value is the closest distance between two data
+                 * points.
+                 *
                  * In `v9.0.2`, the default value has been changed
                  * from `undefined` to `null`.
                  *
--- a/ts/Stock/Indicators/LinearRegression/LinearRegression.ts
+++ b/ts/Stock/Indicators/LinearRegression/LinearRegression.ts
@@ -82,8 +82,8 @@ class LinearRegressionIndicator extends SMAIndicator {
                  * computing the slope and the intercept. This may enchance the
                  * legiblitity of the indicator's values.
                  *
-                 * Default value is the closest distance between two data
-                 * points.
+                 * In `v9.0.2`, the default value has been changed
+                 * from `undefined` to `null`.
                  *
                  * @sample {highstock} stock/plotoptions/linear-regression-xaxisunit
                  *         xAxisUnit set to 1 minute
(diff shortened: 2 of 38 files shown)


ts/Stock/Indicators/LinearRegression/LinearRegression.ts
```diff
@@ -108,7 +108,7 @@ class LinearRegressionIndicator extends SMAIndicator {
                  * @type    {number}
                  * @product highstock
                  */
-                xAxisUnit: void 0
+                xAxisUnit: null
```
pawelfus, line 111: Please change the doclet to inform users about `null` value above:

```
                 * Default value is the closest distance between two data
                 * points.
```
And change type in doclet: `@type    {number}`


Additionally, add an upgrade note (see 4th point: https://github.com/highcharts/highcharts/tree/master/changelog#writing-content-for-the-changelog )

pawelfus (changes requested):
Looks great! Just one minor thing:

ts/Stock/Indicators/Chaikin/ChaikinIndicator.ts
```diff
@@ -73,9 +73,10 @@ class ChaikinIndicator extends EMAIndicator {
          * Paramters used in calculation of Chaikin Oscillator
          * series points.
          *
-         * @excluding index, period
+         * @excluding index
```
karolkolodziej, line 76: If we're giving users the ability to change the `period`, we should inform them where this parameter is used.
Based on the `ChainkinIndicator.ts` this value is only used for `AD` calculation.
``` TS
        ADL = AD.prototype.getValues.call(this, series, {
            volumeSeriesID: params.volumeSeriesID,
            period: period
        }) as IndicatorValuesObject<TLinkedSeries>;

```
This information should be included in the API. My suggestion: 
```
Parameter used indirectly for calculating the accumulation/distribution indicator.  
Decides about the number of data points that are taken into account for the indicator calculations.
```

karolkolodziej (changes requested):
Apart from one minor thing, it looks good!

ts/Stock/Indicators/Momentum/MomentumIndicator.ts
```diff
@@ -67,11 +67,8 @@ class MomentumIndicator extends SMAIndicator {
      * @requires     stock/indicators/momentum
      * @optionparent plotOptions.momentum
      */
-    public static defaultOptions: MomentumOptions = merge(SMAIndicator.defaultOptions, {
-        params: {
-            period: 14
-        }
-    } as MomentumOptions);
+    public static defaultOptions: MomentumOptions =
+    merge(SMAIndicator.defaultOptions, {} as MomentumOptions);
```
pawellysy, line 71: Is th `merge` function needed here? 
couldn't it be just 
`public static defaultOptions = SMAIndicator.defaultOptions as MomentumOptions`

pawellysy (changes requested):
See comment

ts/Stock/Indicators/Momentum/MomentumIndicator.ts
raf18seb: Unfortunately, it generates an error ```code tagged with @optionparent must be an object``` when generating API. I'm leaving it as it is.

pawelfus: FYI @bre1470

bre1470: This is correct, it creates a clone and also follows the pattern for JSDoc.

raf18seb: Anyway, it's resolved now, because I'm not merging an empty object anymore.

ts/Stock/Indicators/LinearRegression/LinearRegression.ts
```diff
@@ -82,8 +82,8 @@ class LinearRegressionIndicator extends SMAIndicator {
                  * computing the slope and the intercept. This may enchance the
                  * legiblitity of the indicator's values.
                  *
-                 * Default value is the closest distance between two data
-                 * points.
+                 * In `v9.0.2`, the default value has been changed
+                 * from `undefined` to `null`.
```
pawelfus, line 86: Please, keep the explanation how default value (null) works :)

pawelfus (changes requested):
Nice! 
Just a minor thing

karolkolodziej (approved):

pawelfus (approved):
Nothing to add, thanks! :)

pawellysy (approved):
Nothing to add 👍

TorsteinHonsi (approved):

PR #15481, opened 2021-04-09
Tools: Activated new warnings for Highcharts migration.

bre1470:
Warnings for:
- no-highcharts-namespace
- no-mixin
- no-wrap


TorsteinHonsi (approved):
Thanks!

Issue #15413, opened 2021-03-27
Keyboard navigation freezes after clicking a value in Heatmap


franklinefrancis:
#### Expected behaviour
Keyboard navigation across data points of Heatmap works without any glitches

#### Actual behaviour
After using keyboard to navigate a few data points of Heatmap and then clicking on one of the data points causes keyboard navigation not to work anymore.

![Heatmap Keyboard Navigation](https://user-images.githubusercontent.com/1227574/112712344-e7f9bc00-8ef4-11eb-9460-aebceeee2ccf.png)

#### Live demo with steps to reproduce
https://jsfiddle.net/franklinefrancis/upd0bv3w/

#### Product version
8.2.0

#### Affected browser(s)
Chrome, Firefox

KacperMadej:
Thank you for reporting the bug.

a11y - fyi: @oysteinmoseng  

_Internal note:_
This is also visible when a point is focused via a11y module selection and a mouse hover doesn't hide the selection.

PR #15482, opened 2021-04-09
Fixed #15416, #15413, heatmap kbd nav issues.

### LINKED ISSUE #15416: Incorrect behaviour of backward keyboard navigation in Treemap and Heatmap
#### Expected behaviour
Backward navigation using keyboard in Treemap and Heatmap should work similar to other charts

#### Actual behaviour
Backward navigation using Shift+Tab keys within Treemap and Heatmap needs to be done twice to move focus out of the chart.

1. Run the JSFiddle for Treemap or Heatmap
2. Start focus with the first text field labelled 'First focus'
3. Press Tab key to move focus to chart (works as expected)
4. Press Tab key again to move focus to text field labelled 'Last focus' (works as expected)
5. Now, press Shift+Tab from 'Last focus' text field. Focus moves back to chart data point as expected.
6. Press Shift+Tab to move focus back to 'First focus' text field. But focus moves to another data point within chart. Only upon pressing Shift+Tab again, the focus moves back to 'First focus' text field. Thus backward keyboard navigation order is different from forward navigation.

Heatmap Backward Keyboard Navigation

Treemap Backward Keyboard Navigation

Treemap - https://jsfiddle.net/franklinefrancis/yaj849uo/
Heatmap - https://jsfiddle.net/franklinefrancis/4wj9efdg/

#### Product version
8.2.0

#### Affected browser(s)
Chrome, Firefox

oysteinmoseng:
Fixed #15416, #15413, heatmap keyboard navigation not working after mouse interaction, and backwards navigation not working as expected.
___
Two issues at play here, one being that `setState` in colorMap-series alters the z-index of the point, causing elements to lose focus. This is solved now by adding an event listener in the a11y module that reapplies focus after `setState` when relevant. This required some refactoring of how `setState` is handled in colorMap-series.

The other issue was related to the first one, but dealt with the auto-init of the first keyboard nav module when the container receives focus. This should not happen when exiting the chart, but is only an issue in combination with the z-index problem, since the focus is lost when the module inits.

### DIFF
--- a/samples/unit-tests/series-heatmap/members/demo.js
+++ b/samples/unit-tests/series-heatmap/members/demo.js
@@ -191,51 +191,19 @@ QUnit.test('Point range after setData(#3758)', function (assert) {
 });
 
 QUnit.test('seriesTypes.heatmap.pointClass.setState', function (assert) {
-    var series = Highcharts.seriesTypes.heatmap,
-        setState = series.prototype.pointClass.prototype.setState,
-        pointAttribs = series.prototype.pointAttribs,
-        noop = Highcharts.noop,
-        point = {
-            graphic: {
-                attr: function (obj) {
-                    var graphic = this,
-                        keys = Object.keys(obj);
-                    keys.forEach(function (key) {
-                        var value = obj[key];
-                        graphic[key] = value;
-                    });
-                },
-                animate: noop,
-                addClass: noop,
-                removeClass: noop
+    const chart = new Highcharts.Chart('container', {
+            chart: {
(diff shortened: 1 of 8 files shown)


TorsteinHonsi (changes requested):
Thanks! We just need a little more verbose description for the changelog, otherwise approved.

oysteinmoseng:
Edited - should ideally be two entries in the changelog perhaps?

TorsteinHonsi (approved):
Nah, that's fine!

PR #15484, opened 2021-04-09
Fixed #15343, treemap and sunburst kbd nav broken.

### LINKED ISSUE #15343: Accessibility focus freezes when navigating through treemap points
Like in the title.

https://jsfiddle.net/BlackLabel/mxzo5rcn/

#### Product version
Highcharts 9.0.1

FYI: @oysteinmoseng

oysteinmoseng:
Fixed #15343, regression with treemap and sunburst keyboard navigation.
___
After earlier fix in a11y module that excluded `isInside === false` points from navigation.


TorsteinHonsi (approved):
Thanks!

PR #15485, opened 2021-04-09
Migration of animation core

bre1470:
- Moved global assignments and types to masters.
- Replaced var with const and let
- Removed warnings from task output (kept in editor)


TorsteinHonsi (approved):
Thanks!

PR #15490, opened 2021-04-12
tools/lock-file

khlieng:
This is a fresh lockfile created just now on the latest commit.


TorsteinHonsi (changes requested):
It blocks CI... Any insight @bre1470 ?

khlieng:
I think it might be a cache issue.

bre1470:
Yes, that might be the case. Let's see...

khlieng:
Hmm, when I ran `npm install` on this branch now npm added a few empty `requires` objects to the lockfile, so thats probably what happens on CI too then. I'll try to commit it.

bre1470:
No, it is not the cache. The error says "Untracked files found. Did you forget to commit any transpiled files? Failing build now as this likely will trigger errors later in the build.", so it might be a false positive. We should consider a forced merge.

bre1470:
I get locally the error: This version of npm is compatible with lockfileVersion@1, but package-lock.json was generated for lockfileVersion@2. I'll try to do my best with it!

bre1470:
This will likely always change a bit and produce an error. I suggest to remove the file from CI monitoring.

From the docs:
> Note that the file format changed significantly in npm v7 to track information that would have otherwise required looking in node_modules or the npm registry. Lockfiles generated by npm v7 will contain lockfileVersion: 2.

No version provided: an "ancient" shrinkwrap file from a version of npm prior to npm v5.
1: The lockfile version used by npm v5 and v6.
2: The lockfile version used by npm v7, which is backwards compatible to v1 lockfiles.
3: The lockfile version used by npm v7, without backwards compatibility affordances. This is used for the hidden lockfile at node_modules/.package-lock.json, and will likely be used in a future version of npm, once support for npm v6 is no longer relevant.

TorsteinHonsi:
Pending @bre1470 's approval

bre1470 (approved):

TorsteinHonsi (approved):

PR #15494, opened 2021-04-13
enhancement/15488-rect

### LINKED ISSUE #15488: Tooltip borderRadius only works for callout, not square (I think workaround available)
#### Expected behaviour

Setting border radius for a tooltip with shape square should have rounded corners. In the jsfiddle it doesn't.

So I poked around in the source code and just using `rect` seems to work. it's not in the typescript definitions or official docs, though https://api.highcharts.com/highcharts/tooltip.shape - would it be reasonable to add rect as an official option and make `square` legacy?

#### Actual behaviour

no rounded corners

https://jsfiddle.net/zvwoskfb/

#### Product version
9.0.1

#### Affected browser(s)

All

khlieng:
Fixed #15488, square shape did not support rounded corners.

### DIFF
--- a/ts/Core/Options.ts
+++ b/ts/Core/Options.ts
@@ -3747,7 +3747,7 @@ H.defaultOptions = {
 
         /**
          * The name of a symbol to use for the border around the tooltip. Can
-         * be one of: `"callout"`, `"circle"`, `"square"` or `"rect"`. When
+         * be one of: `"callout"`, `"circle"` or `"rect"`. When
          * [tooltip.split](#tooltip.split)
          * option is enabled, shape is applied to all boxes except header, which
          * is controlled by
--- a/ts/Core/Options.ts
+++ b/ts/Core/Options.ts
@@ -3747,7 +3747,7 @@ H.defaultOptions = {
 
         /**
          * The name of a symbol to use for the border around the tooltip. Can
-         * be one of: `"callout"`, `"circle"`, or `"square"`. When
+         * be one of: `"callout"`, `"circle"`, `"square"` or `"rect"`. When
          * [tooltip.split](#tooltip.split)
          * option is enabled, shape is applied to all boxes except header, which
          * is controlled by
--- a/ts/Core/Tooltip.ts
+++ b/ts/Core/Tooltip.ts
@@ -177,7 +177,7 @@ declare global {
             plotX: number;
             plotY: number;
         }
-        type TooltipShapeValue = ('callout'|'circle'|'square');
+        type TooltipShapeValue = ('callout'|'circle'|'square'|'rect');
     }
 }
 
--- a/samples/unit-tests/svgrenderer/symbol/demo.js
+++ b/samples/unit-tests/svgrenderer/symbol/demo.js
@@ -170,3 +170,29 @@ QUnit.test('Arc', assert => {
         '#15382: Y radius should be 0'
     );
 });
+
+QUnit.test('Square/rect', assert => {
+    ['square', 'rect'].forEach(shape => {
+        const fn = Highcharts.SVGRenderer.prototype.symbols[shape];
+
+        let path = fn(0, 0, 10, 10);
+        assert.strictEqual(
+            path.length,
+            5,
+            `${shape}, no options: Path should have no curves`
+        );
+
+        path = fn(0, 0, 10, 10, { r: 0 });
+        assert.strictEqual(
+            path.length,
+            5,
+            `${shape}, r=0: Path should have no curves`
+        );
+
+        path = fn(0, 0, 10, 10, { r: 5 });
+        assert.ok(
+            path.length > 5,
+            `${shape}, r=5: Path should have curves`
+        );
+    });
+});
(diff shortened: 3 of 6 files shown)


TorsteinHonsi (changes requested):
There's no logical reason why `rect` should allow rounded borders while `square` doesn't. In fact, if they were different, it should be that square should have equal width and height. But we don't need that, because the use case (point markers) pass equal arguments.

So instead of relaying the symbol setting of the heatmap I will propose that we implement rounded corners in the shared symbol function. Then the tooltip border radius would be changed by the `borderRadius` option like expected. In this proposal (based on master) I have refactored out and shared the rounded corner setup from `callback`:

```diff
diff --git a/ts/Core/Renderer/SVG/SVGRenderer.ts b/ts/Core/Renderer/SVG/SVGRenderer.ts
index db2528ffa9..a1bdd59330 100644
--- a/ts/Core/Renderer/SVG/SVGRenderer.ts
+++ b/ts/Core/Renderer/SVG/SVGRenderer.ts
@@ -2540,18 +2540,40 @@ SVGRenderer.prototype.escapes = {
     '"': '&quot;'
 };
 
+const roundedRect: Highcharts.SymbolFunction = (
+    x,
+    y,
+    w,
+    h,
+    options
+): SVGPath => {
+    const r = (options && options.r) || 0;
+    return [
+        ['M', x + r, y],
+        ['L', x + w - r, y], // top side
+        ['C', x + w, y, x + w, y, x + w, y + r], // top-right corner
+        ['L', x + w, y + h - r], // right side
+        ['C', x + w, y + h, x + w, y + h, x + w - r, y + h], // bottom-rgt
+        ['L', x + r, y + h], // bottom side
+        ['C', x, y + h, x, y + h, x, y + h - r], // bottom-left corner
+        ['L', x, y + r], // left side
+        ['C', x, y, x, y, x + r, y] // top-left corner
+    ];
+};
+
 // #15291
-const rect = (
-    x: number,
-    y: number,
-    w: number,
-    h: number): SVGPath => [
-    ['M', x, y],
-    ['L', x + w, y],
-    ['L', x + w, y + h],
-    ['L', x, y + h],
-    ['Z']
-];
+const rect: Highcharts.SymbolFunction = function (x, y, w, h, options): SVGPath {
+    if (options && options.r) {
+        return roundedRect.apply(0, arguments);
+    }
+    return [
+        ['M', x, y],
+        ['L', x + w, y],
+        ['L', x + w, y + h],
+        ['L', x, y + h],
+        ['Z']
+    ];
+};
 
 /**
  * An extendable collection of functions for defining symbol paths.
@@ -2722,17 +2744,7 @@ SVGRenderer.prototype.symbols = {
             anchorY = options && options.anchorY || 0,
             path: SVGPath;
 
-        path = [
-            ['M', x + r, y],
-            ['L', x + w - r, y], // top side
-            ['C', x + w, y, x + w, y, x + w, y + r], // top-right corner
-            ['L', x + w, y + h - r], // right side
-            ['C', x + w, y + h, x + w, y + h, x + w - r, y + h], // bottom-rgt
-            ['L', x + r, y + h], // bottom side
-            ['C', x, y + h, x, y + h, x, y + h - r], // bottom-left corner
-            ['L', x, y + r], // left side
-            ['C', x, y, x, y, x + r, y] // top-left corner
-        ];
+        path = roundedRect(x, y, w, h, { r });
 
         if (!isNumber(anchorX)) {
             return path;
```

ts/Core/Options.ts
```diff
@@ -3747,7 +3747,7 @@ H.defaultOptions = {
 
         /**
          * The name of a symbol to use for the border around the tooltip. Can
-         * be one of: `"callout"`, `"circle"`, or `"square"`. When
+         * be one of: `"callout"`, `"circle"`, `"square"` or `"rect"`. When
```
TorsteinHonsi, line 3750: Actually `square` doesn't make sense for the tooltip border anymore. It is a rectangle, not a square. Should we just remove this from the description? It will still work, but we don't have to encourage it by listing it here.

ts/Core/Tooltip.ts
```diff
@@ -177,7 +177,7 @@ declare global {
             plotX: number;
             plotY: number;
         }
-        type TooltipShapeValue = ('callout'|'circle'|'square');
+        type TooltipShapeValue = ('callout'|'circle'|'square'|'rect');
```
TorsteinHonsi, line 180: As above - `tooltip.shape = 'square'` will actually render a rectangle, so I think it is better to remove `square`

TorsteinHonsi (changes requested):

ts/Core/Options.ts
khlieng: But wont typescript users with `square` tooltips get complaints from the compiler then? Do we expect them to just replace it with `rect`?

TorsteinHonsi: Yes good point. I suggest we remove it from the description, but leave it in the type.

TorsteinHonsi (approved):

PR #15504, opened 2021-04-14
bugfix/15502-verticalCounter-annotation

### LINKED ISSUE #15502: Highstock - Issue found with the vertical counter in the Stock Tool
Hi there,

The counter of the vertical counter label seems to don't work. I tested with the latest (v.9.0.1) and is not fixed.

When you select a vertical counter and implement it in the chart, the variable _this.verticalCounter_ seems to don't exists and is never caught, then the label always applies the text "0" by default.

image

It can be reproduced in:

am I not implementing well or missing a parameter?

Thanks in advance!

khlieng:
Fixed #15502, vertical counter annotation did not work.


pawelfus (approved):
Thanks!

TorsteinHonsi (approved):

PR #15511, opened 2021-04-15
bugfix/empty-area-svg-error

khlieng:
Fixed a regression, area series with empty data caused SVG error.

### DIFF
--- a/samples/unit-tests/series-area/update/demo.js
+++ b/samples/unit-tests/series-area/update/demo.js
@@ -70,4 +70,14 @@ QUnit.test('Updating series stacked property', assert => {
         chart.series[0].areaPath.slice(-1)[0].includes("Z"),
         'The last index of the path array contains the closure'
     );
+
+    chart.series[0].update({
+        data: []
+    });
+
+    assert.strictEqual(
+        chart.series[0].areaPath.length,
+        0,
+        'Path should be empty when there is no data'
+    );
 });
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (commented):
How did this issue manifest? In this case it looks okay: https://jsfiddle.net/highcharts/qhact8ou/

Also, we have a generic test for all series types and no data/empty data. Maybe we should adapt that to catch this case? 

https://github.com/highcharts/highcharts/blob/908d8c4f7a25fe34f203a1bff2584fb2a7aad34a/samples/unit-tests/series/seriestypes/demo.js#L1-L43

khlieng:
Its a unreleased regression caused by #14900, since its not a regular thrown error that can be caught that test didnt catch it.

TorsteinHonsi:
How does it look then? What happens?

khlieng:
Looks fine, just a console error: https://jsfiddle.net/grvfqnce/

TorsteinHonsi:
Do you know where the regression comes from? Have you bisected? What changed? Why is this an issue now when it wasn't earlier?

Trying to understand why we need to add a check...

khlieng:
The issue comes from the PR I linked above, which closes the path to make strokes look nicer, this PR prevents pushing the `Z` if the path is empty since setting `d` to just `Z` logs an error.

TorsteinHonsi (approved):

PR #15516, opened 2021-04-15
bugfix/15435-stock-clip-update

### LINKED ISSUE #15435: The phenomenon that the series is not drawn or is cut off when using highstock
#### Actual behaviour
Series and y-axis are dynamically added to express various metrics.
스크린샷 2021-04-05 오전 10 57 41
스크린샷 2021-04-05 오전 11 33 59
The end of the series is not drawn like the attached image

I need help. Help

#### Product version
v9.0.1

#### Affected browser(s)
ALL

khlieng:
Fixed #15435, clipping was sometimes wrong after `update` with stock loaded.

### DIFF
--- a/samples/unit-tests/series/clip/demo.js
+++ b/samples/unit-tests/series/clip/demo.js
@@ -5,12 +5,17 @@ QUnit.test('General series clip tests', assert => {
         const chart = Highcharts.chart('container', {
                 series: [{
                     data: [10, 20]
+                }, {
+                    data: [1, 2, 3],
+                    yAxis: 1
                 }],
-                yAxis: {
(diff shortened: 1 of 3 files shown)


pawelfus (approved):
Thanks, looks good!

Just a question: Would it make sense to get rid of `animation` argument in `series.setClip()`? It's taken from `this.options` anyway

khlieng:
No, because there are calls to `setClip` without `animation`, passing in `animation` means it should prepare to animate. So we would still need a boolean at least.

TorsteinHonsi (approved):
Excellent, thanks!

PR #15517, opened 2021-04-15
docs/class-reference-links

codyatwork:
Fixed some links in docs to class reference pages and made slight grammatical tweaks to contributing guidelines.

(This may seem so minor as to not even be worth it, but I was clicking links to see the supported format keys in `dateFormat()` and thought that information was removed because after looking in the sidebar I ended up on `dateFormats`.)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -43,7 +43,7 @@ Your issue should:
 
 ## Suggesting Features and Enhancements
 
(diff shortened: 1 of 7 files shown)


bre1470 (approved):
Thank you! Looks good. 👍

TorsteinHonsi (approved):

PR #15530, opened 2021-04-19
bugfix/15509/allowedtags-missing-u-tag

### LINKED ISSUE #15509: <u> getting stripped for useHTML: true
#### desc
i assume it's getting stripped because the `<u>` tag is deprecated? 
but it's back and not deprecated as per https://developer.mozilla.org/en-US/docs/Web/HTML/Element/u

#### Expected behaviour
`<u>` tag should render the `<u>` tag

#### Actual behaviour
`<u>` tag getting stripped

https://jsfiddle.net/yhtz0ma4/

1. set title useHTML: true
2. set text to '<u>some text</u>'

#### Product version
9.0.0

#### Affected browser(s)
chrome

DJTechnoo:
Fixed #15509, `<u>` tag not working with `useHTML`.


TorsteinHonsi (approved):
Thanks! Approved, ready to merge as soon as the tests pass

khlieng:
Merging master now should fix CI (hopefully 🙏) (#15533).

TorsteinHonsi (approved):

PR #15538, opened 2021-04-20
bugfix/15534-area-null-empty-update

### LINKED ISSUE #15534: Area chart crashes when transitioning from data to 0 data
The expectation would be for area series to adapt to changes in data, similarly to other series types (e.g. a series of type line doesn't crash: https://codesandbox.io/s/highcharts-react-simple-chart-forked-qslj1?file=/demo.jsx ) and an empty series is presented, which is the expected behaviour.

#### Expected behaviour
1. Press the Update button
2. An empty area chart appears

#### Actual behaviour
1. Press the Update button
2. The chart crashes with an error

https://codesandbox.io/s/highcharts-react-simple-chart-forked-q0kvm?file=/demo.jsx

#### Product version
highcharts: 9.0.1
highcharts-react-official: 3.0.0

#### Affected browser(s)
All.

khlieng:
Fixed #15534, updating area series containing null data to no data threw.

### DIFF
--- a/samples/unit-tests/series-area/update/demo.js
+++ b/samples/unit-tests/series-area/update/demo.js
@@ -3,7 +3,8 @@ QUnit.test('Updating series stacked property', assert => {
         chart: {
             type: 'area',
             width: 600,
-            height: 350
+            height: 350,
+            animation: true
         },
         xAxis: {
             categories: [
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #15539, opened 2021-04-20
Replace Hebraic with Hebrew

pyrocks:
The language is called Hebrew, not Hebraic. No other code changes.


TorsteinHonsi (approved):
Oh, thanks! That's embarrassing. Norwenglish there 🤦

PR #15545, opened 2021-04-20
ts/migration-utilities

marvin19:
Migration: Refactoring of Utilities.

### DIFF
--- a/ts/Core/Options.ts
+++ b/ts/Core/Options.ts
@@ -40,13 +40,30 @@ import palette from './Color/Palette.js';
 import Time from './Time.js';
 import U from './Utilities.js';
 const {
-    merge
+    getNestedProperty,
+    isNumber,
+    merge,
+    pick,
+    pInt
 } = U;
+
+/* *
+ *
+ *  Declarations
+ *
+ * */
 
 declare module './Chart/ChartLike'{
     interface ChartLike {
         marginRight: Highcharts.ChartOptions['marginRight'];
         polar: Highcharts.ChartOptions['polar'];
+    }
+}
+
+declare module './GlobalsLike' {
+    interface GlobalsLike {
+        defaultOptions: Highcharts.Options;
+        time: Time;
     }
 }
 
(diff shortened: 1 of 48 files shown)


ts/Core/Options.ts
```diff
@@ -4370,18 +4385,289 @@ H.time = new Time(
...
+             * @name Highcharts.time
+             * @type {Highcharts.Time}
+             */
+            H.time = defaultTime;
+        }
+    }
+
+    return defaultOptions;
+}
+H.setOptions = setOptions; // @todo move to master
```
bre1470, line 4661: `@todo` is left to do, to move the H modification.

ts/Core/Options.ts
```diff
@@ -4370,18 +4385,289 @@ H.time = new Time(
...
+ * object.
+ *
+ * @function Highcharts.getOptions
+ *
+ * @return {Highcharts.Options}
+ */
+function getOptions(): Highcharts.Options {
+    return defaultOptions;
+}
+H.getOptions = getOptions; // @todo move to master
```
bre1470, line 4494: `@todo` to do.

bre1470 (changes requested):
Some `@todo`s to move to master.

bre1470 (approved):
Thank you! Looks good. 👍

TorsteinHonsi (commented):
Thanks!

Why move `format` and `numberFormat` from Utilities.ts to Options.ts? Functionality wise, they are not tied to the options. They are general utility functions that can be applied in callbacks, event handlers, extensions and other situations not related to options.

bre1470:
After the globals modification has been removed in utilities, a dependency cycle occurred between options and utilities. This is caused by both functions depending on options defaults.
The logical solution was to move the functions to the options. That way the rest of the utilities functions do not depend on options defaults. So for example you undergo the need of the unrelated Highcharts options for the data layer, when it makes use of the event and object functions.

bre1470 (changes requested):
Internal discussion resolved that we should create a new `Core/FormatUtilities.ts` containing the formatting functions.

bre1470 (approved):

TorsteinHonsi (approved):
Nice, thanks!

PR #15554, opened 2021-04-21
bugfix/15512-rangeSelector-dataGrouping-dynamic-series

### LINKED ISSUE #15512: dataGrouping doesn't apply to indicator added via stockTools
#### Expected behaviour
New Indicator has the same dataGrouping enabled as the main series.

#### Actual behaviour
The dataGrouping, that is defined in rangeSelector buttons isn't applied to the new indicator added via stockTools.

https://jsfiddle.net/BlackLabel/5u2hty3L/

Add ema indicator via stockTools.

#### Product version
v9.0.1

#### Affected browser(s)
all

khlieng:
Fixed #15512, dynamically added series did not use range selector data grouping.

### DIFF
--- a/samples/unit-tests/series/datagrouping/demo.js
+++ b/samples/unit-tests/series/datagrouping/demo.js
@@ -435,9 +435,10 @@ QUnit.test('Switch from grouped to non-grouped', function (assert) {
         'Monthly columns'
     );
 
-    const dataGrouping = chart.addSeries({}).options.dataGrouping;
-    assert.ok(
-        dataGrouping.forced && dataGrouping.units,
+    const series = chart.addSeries({ data: chart.series[0].options.data });
+    assert.deepEqual(
+        series.currentDataGrouping,
+        chart.series[0].currentDataGrouping,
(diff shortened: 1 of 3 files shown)


samples/unit-tests/series/datagrouping/demo.js
```diff
@@ -435,6 +435,12 @@ QUnit.test('Switch from grouped to non-grouped', function (assert) {
         'Monthly columns'
     );
 
+    const dataGrouping = chart.addSeries({}).options.dataGrouping;
+    assert.ok(
+        dataGrouping.forced && dataGrouping.units,
+        '#15512: Datagrouping from range selector should be used'
```
pawelfus, line 441: Should we rather check if options are exactly the same? Something like `assert.deepEqual(dataGrouping, rangeSelector.buttonOptions[rangeSelector.selected].dataGrouping)`?

pawelfus (approved):
Looks good, just a minor question

TorsteinHonsi (approved):
Thanks!

PR #15556, opened 2021-04-21
bugfix/15555-dependencywheel-animation-update

### LINKED ISSUE #15555: Dependency wheel - error after update while animation running
#### Behaviour
A destroyed element tries to animate causing an error.

https://jsfiddle.net/BlackLabel/y01gcmbq/

#### Product version
9.0.1

khlieng:
Fixed #15555, removing dependency wheel points during animation threw.


TorsteinHonsi (approved):
Thanks!

PR #15562, opened 2021-04-22
bugfix/4406-addSeries-animation-clip

### LINKED ISSUE #4406: No animation when adding/removing series in highstock
Hello

I try to dynamically add/remove series in a highstock chart. I have tried various options, but I can't get the added series to animate. 

The only animation showing are:
- the first serie plotting during chart creation
- the zoom/scale animation when adding data

Here is what it looks like:
http://jsfiddle.net/mtasxa0o/

i even tried to force animation when adding a serie `mchart.addSeries({},true,true);` but with no luck.

**This only applies to highstock.** If I call the same code with highchart it works:
http://jsfiddle.net/rjj4hL59/

Am I missing something?

raf18seb:
Fixed #4406, series animation didn't work when adding a new series using the `chart.addSeries()` method.

### DIFF
--- a/ts/Core/Series/Series.ts
+++ b/ts/Core/Series/Series.ts
@@ -2735,6 +2735,8 @@ class Series {
 
     public symbol?: string;
 
+    public useTemporaryClip?: boolean;
+
     public tooltipOptions: Highcharts.TooltipOptions = void 0 as any;
 
     public touched?: boolean;
@@ -4864,6 +4866,7 @@ class Series {
         }
 
         const sharedClipKey = [
+            this.useTemporaryClip ? '_temporaryClip' : '', // #4406
             animation && animation.duration,
             animation && animation.easing,
             animation && animation.defer,
@@ -4894,6 +4897,10 @@ class Series {
             inverted = chart.inverted,
             seriesClipBox = this.clipBox,
             clipBox = this.getClipBox(animation),
+            tempClipKey = ( // #4406
+                this.sharedClipKey &&
+                this.sharedClipKey.split(',')[0] === '_temporaryClip'
+            ) ? this.sharedClipKey : void 0,
             sharedClipKey = this.getSharedClipKey(animation), // #4526
             clipRect = chart.sharedClips[sharedClipKey],
             markerClipRect = chart.sharedClips[sharedClipKey + 'm'];
@@ -4923,7 +4930,8 @@ class Series {
 
 
             }
-            chart.sharedClips[sharedClipKey] = clipRect = renderer.clipRect(clipBox);
+            chart.sharedClips[sharedClipKey] = clipRect =
+                renderer.clipRect(clipBox);
 
             // Create hashmap for series indexes
             clipRect.count = { length: 0 };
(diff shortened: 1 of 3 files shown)


ts/Core/Series/Series.ts
```diff
@@ -4864,6 +4866,7 @@ class Series {
         }
 
         const sharedClipKey = [
+            this.useTemporaryClip ? '_temporaryClip' : '', // #4406
```
khlieng, line 4869: You could probably get away with only adding a temp prefix if a shared clip with that key already exists.

ts/Core/Series/Series.ts
```diff
@@ -4894,6 +4897,10 @@ class Series {
             inverted = chart.inverted,
             seriesClipBox = this.clipBox,
             clipBox = this.getClipBox(animation),
+            tempClipKey = ( // #4406
+                this.sharedClipKey &&
+                this.sharedClipKey.split(',')[0] === '_temporaryClip'
+            ) ? this.sharedClipKey : void 0,
```
khlieng, line 4903: I think this could be moved below the `getSharedClipKey` call and simplified to something like `isTemp = sharedClipKey.slice(0, 5) === '_temp'`.

ts/Core/Series/Series.ts
```diff
@@ -4961,7 +4969,15 @@ class Series {
...
-                    chart.sharedClips[sharedClipKey + 'm'] = markerClipRect.destroy();
+                    chart.sharedClips[sharedClipKey + 'm'] =
+                        markerClipRect.destroy();
+                }
+
+                // Remove the temporary clipping rectangle when the series
+                // animation has finished after chart.addSeries(), (#4406)
+                const tempClip = tempClipKey && chart.sharedClips[tempClipKey];
+                if (tempClipKey && tempClip) {
+                    chart.sharedClips[tempClipKey] = tempClip.destroy();
```
khlieng, line 4980: I think this should be removed since the existing `clipRect.destroy` is going to deal with it for the non-stock case (no `clipBox`) and then add a `else if (isTemp)` to it that chops off the temp prefix and (since shared clips are not destroyed in stock):

- If a shared clip with that key already exists, changes the group clip to that and destroys the temp one.
- If not, keeps the temp one and moves it to the non-temp key and updates `this.sharedClipKey`.

This should fix the visual diff.

khlieng, line 4980: I think this should be removed since the existing `clipRect.destroy` is going to deal with it for the non-stock case (no `clipBox`) and then add a `else if (isTemp)` block to it that chops off the temp prefix and (since shared clips are not destroyed in stock):

- If a shared clip with that key already exists, changes the group clip to that and destroys the temp one.
- If not, keeps the temp one and moves it to the non-temp key and updates `this.sharedClipKey`.

This should fix the visual diff.

pawellysy (approved):
Looks good! 👍

raf18seb:
Solved by https://github.com/highcharts/highcharts/pull/16341

PR #15566, opened 2021-04-23
bugfix/15525-line-dataLabel-drilldown

### LINKED ISSUE #15525: Clicking Data Labels doesn't trigger drilldown for Area and Line charts
#### Expected behaviour
In column charts with drilldown, when we click on the data labels it triggers drilldown, but the same doesn't happen to "area"  and "line" charts. For area it triggers the drilldown only if we click on the data point itself or in the axis label, but not in the data labels.

#### Actual behaviour
Clicking on data labels should trigger the drilldown. In the jsfiddle below, you can try changing the chart type to "column" and you'll see it works fine.

https://jsfiddle.net/jgtkcrvy/

#### Product version
Highcharts JS v9.0.1

#### Affected browser(s)
Tested in Chrome Version 89

khlieng:
Fixed #15525, clicking data label did not drill down line and area series.

### DIFF
--- a/samples/unit-tests/datalabels/events/demo.details
+++ b/samples/unit-tests/datalabels/events/demo.details
@@ -1,0 +1,5 @@
+---
+ resources:
+   - https://code.jquery.com/qunit/qunit-2.0.1.js
+   - https://code.jquery.com/qunit/qunit-2.0.1.css
+...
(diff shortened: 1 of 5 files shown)


TorsteinHonsi (approved):
Thanks!

PR #15570, opened 2021-04-23
Migration: SVGElement

bre1470:
- Removed SVGElement type from global namespace.
- Moved global additions from SVGElement to Globals.
- Replaced `var` in SVGElement with `const`/`let`.


TorsteinHonsi (approved):
Thanks!

PR #15571, opened 2021-04-23
bugfix/#13895-boost-zoneAxis

### LINKED ISSUE #13895: Zones for boosted charts do not work for an "x" zoneAxis
#### Expected behaviour
When configuring a **boosted** series, setting up the zones and setting the zoneAxis to **"x"** should display zones based on x-axis.

#### Actual behaviour
The current behavior is it displays zones based on the y-axis regardless if you set the zoneAxis to "x". Meaning that it will always use the plot's y-values to zone ignoring the zoneAxis.

There was this fix some time ago: #10439
Looking at it briefly I see that it checks for y values only and doesn't seem to take the zoneAxis into consideration?
https://github.com/highcharts/highcharts/blob/master/js/Extensions/Boost/WGLRenderer.js#L524

https://jsfiddle.net/34bLw0m9/

#### Product version
Highcharts v8.1.2

#### Affected browser(s)
All

tzaitoun:
Fixed #13895, boost will use zoneAxis to determine zone point color.

### DIFF
--- a/ts/Extensions/Boost/WGLRenderer.ts
+++ b/ts/Extensions/Boost/WGLRenderer.ts
@@ -397,6 +397,7 @@ function GLRenderer(
             isXInside = false,
             isYInside = true,
             firstPoint = true,
+            zoneAxis = options.zoneAxis || 'y',
             zones = options.zones || false,
             zoneDefColor: (Color|undefined) = false as any,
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (approved):
Thanks, works well!

PR #15582, opened 2021-04-27
bugfix/15523-lollipop-negativeColor

### LINKED ISSUE #15523: Negative color is used even for positive values in Lollipop
#### Expected behaviour
When defining a color and a negativeColor parameters, we should get positive values colored with the "color" parameter and negative values colored with the "negativeColor" parameter

#### Actual behaviour
Both positive and negative values are colored with the "negativeColor" parameter

https://jsfiddle.net/vegaelce/ztkdm21v/

#### Product version
Highcharts 9.1

#### Affected browser(s)
All

khlieng:
Fixed #15523, `negativeColor` did not work for lollipop series.

### DIFF
--- a/samples/unit-tests/series-lollipop/lollipop/demo.js
+++ b/samples/unit-tests/series-lollipop/lollipop/demo.js
@@ -5,10 +5,16 @@ QUnit.test('Lollipop offset affection.', function (assert) {
         },
         series: [
             {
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (approved):
Awesome, thanks for cleaning up typing too!

PR #15584, opened 2021-04-27
Migration: Time

marvin19:
Migration: Removed internal namespace from time.ts


ts/Core/Axis/DateTimeAxis.ts
```diff
@@ -26,14 +28,19 @@ declare module '../Series/SeriesOptions' {
     }
 }
 
+declare module './TickPositionsArray' {
+    interface TickPositionsArray extends Array<number> {
+    }
+}
```
bre1470, line 34: This should be part of the TickPositionsArray file, because it is the essential definition of the type.

ts/Core/Globals.ts
```diff
@@ -75,6 +75,11 @@ declare global {
      * @todo: Rename UMD argument `win` to `window`
      */
     const win: Window|undefined;
+
+    interface Window {
+        moment: any;
+        Date: any;
+    }
```
bre1470, line 82: No more `any` types in Globals (only `AnyRecord`). Delete this interface definition and cast the `win`/`window` object to `AnyRecord`, if the code access the `moment` or `Date` properties in an unsafe way.

bre1470 (changes requested):
Looks good! Some smaller things need to be addressed.

bre1470 (approved):
Thank you! Looks good. 👍

TorsteinHonsi (approved):

PR #15586, opened 2021-04-27
Docs: Security page

TorsteinHonsi:
A page explaining XSS filtering including options for pre v9 users.


cvasseng (approved):
Looks real good!

PR #15591, opened 2021-04-27
bugfix/#15432-mutated-zone-default-color

### LINKED ISSUE #15432: Removing zone from boosted graph turns curve black
#### Expected behaviour
Calling
```
chart.update({
  plotOptions: {
    series: {
      zones: []
    }
  }
})
```
should restore boosted graph to original color.

#### Actual behaviour
The curve becomes black when calling the above method on chart.

https://jsfiddle.net/c3pzuef5/

In this demo, click the remove zone button so that it calls the chart.update mentioned previously and the curve will turn black.

#### Product version
v9.0.1

#### Affected browser(s)
All

tzaitoun:
Fixed #15432, boost zone default color is cloned instead of mutated.

The **zoneDefColor** rgba array is mutated for each point in the series:
```
(pcolor as any)[0] /= 255.0;
(pcolor as any)[1] /= 255.0;
(pcolor as any)[2] /= 255.0;
```
So the first point that uses **zoneDefColor** will have the right color and everything after will have rgb = [0, 0, 0] (black).
https://jsfiddle.net/5tcpuedx/

### DIFF
--- a/ts/Extensions/Boost/WGLRenderer.ts
+++ b/ts/Extensions/Boost/WGLRenderer.ts
@@ -816,7 +816,7 @@ function GLRenderer(
 
             // Note: Boost requires that zones are sorted!
             if (zones) {
-                pcolor = (zoneDefColor as any).rgba;
+                pcolor = (zoneDefColor as any).rgba.slice();
                 zones.some(function ( // eslint-disable-line no-loop-func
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (approved):
Awesome, thanks!

PR #15592, opened 2021-04-28
bugfix/15491-rangeselector-vertical-center

### LINKED ISSUE #15491: Range selector button adaptive width
#### Expected behaviour
Range selector buttons adjust their size depending on the content.

#### Actual behaviour
When the text of the button is long enough it just overflows beyond buttons boundaries instead of expanding the container.

https://jsfiddle.net/j8xv6u2d/2/ (see 'Seit Vergleich' button)

#### Product version
Highstock 9.0.1

#### Affected browser(s)
All

khlieng:
Fixed #15491, range selector button text vertical centering did not work well and `paddingLeft` and `paddingRight` in `buttonTheme` were not supported.

### DIFF
--- a/ts/Core/Renderer/HTML/AST.ts
+++ b/ts/Core/Renderer/HTML/AST.ts
@@ -203,6 +203,7 @@ class AST {
         'orient',
         'padding',
         'paddingLeft',
+        'paddingRight',
         'patternUnits',
         'r',
         'refX',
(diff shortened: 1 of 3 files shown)


khlieng:
So the visual diffs are working as intended, the text is centered now. The reason there are 100 more diffs on the last commit is that more of the stock visual tests actually ran (but still not all of them, first commit: `TOTAL: 131 FAILED, 12 SUCCESS`, second commit: `TOTAL: 230 FAILED, 18 SUCCESS` and theres 296 of them 🤔).

TorsteinHonsi:
Ok, as troublesome as it is for the review process, I agree that this is correct. Will you go over all the samples in the review tool, copy/paste a descriptive text and Approve them?

TorsteinHonsi (approved):

PR #15595, opened 2021-04-28
Migration: var

bre1470:
Changed `var` to `let`. `const` rule is now a warning until the const migration step is completed. This will be separate PRs, because most `const` warnings need manual review. BTW warnings er hidden in the linting task recently, so they are only in editors visible (where they can be filtered).


TorsteinHonsi (approved):
Nice, thanks!

PR #15597, opened 2021-04-28
Migration: Utilities

marvin19:
Removed Highcharts namespace from Utilities.


ts/Core/FormatUtilities.ts
```diff
@@ -302,10 +302,16 @@ function numberFormat(
...
     dateFormat,
     format,
     numberFormat
 };
 
-export default exports;
+namespace FormatUtilities {
+    export interface FormatterCallbackFunction<T> {
+        (this: T): string;
+    }
```
bre1470, line 314: This name can be simplified to `FormatCallback`.

ts/Core/Utilities.ts
```diff
@@ -2457,4 +2386,42 @@ const utilitiesModule = {
...
+        order: number;
+    }
+    export interface ObjectEachCallbackFunction<TObject, TContext> {
+        (
+            this: TContext,
+            value: TObject[keyof TObject],
+            key: keyof TObject,
+            obj: TObject
+        ): void;
+    }
```
bre1470, line 2415: This name can be simplified to `ObjectEachCallback`.

bre1470 (changes requested):

bre1470 (approved):
Thank you! Looks good. 👍

TorsteinHonsi (approved):

PR #15598, opened 2021-04-28
bugfix/15593-stocktools-overflow

### LINKED ISSUE #15593: StockTools GUI doesn't render correctly in a wrapper with multiple charts
#### Expected behaviour
Inside wrappers add StockTools GUI the same as in the JS (same height as the chart with additional pagination).

#### Actual behaviour
Inside wrappers, StockTools GUI pagination is missing and it occupies too much space, which results in overlaping.

Correct JS: https://jsfiddle.net/BlackLabel/g40s5vzo/
Angular: https://stackblitz.com/edit/highcharts-angular-stock-gui-5byr8x
Vue: https://codesandbox.io/s/highcharts-stockgui-tools-see-whats-happen-with-large-toolbar-20pbm
React: https://codesandbox.io/s/stocktools-gui-hrmbi

#### Product version

#### Affected browser(s)
All

khlieng:
Fixed #15593, Stock Tools overflowed the chart container when the class `chart` was not set on the container.

### DIFF
--- a/samples/unit-tests/indicator-price-indicator/recalculations/demo.js
+++ b/samples/unit-tests/indicator-price-indicator/recalculations/demo.js
@@ -185,6 +185,10 @@ function (assert) {
         controller = new TestController(chart),
         button = chart.stockTools.listWrapper.childNodes[0].childNodes[0];
 
+    document.getElementsByClassName(
+        'highcharts-stocktools-wrapper'
+    )[0].style.display = 'none';
+
     // Show croshair with the label.
     controller.moveTo(200, 200);
     assert.strictEqual(
--- a/samples/unit-tests/indicator-price-indicator/recalculations/demo.js
+++ b/samples/unit-tests/indicator-price-indicator/recalculations/demo.js
@@ -185,9 +185,7 @@ function (assert) {
         controller = new TestController(chart),
         button = chart.stockTools.listWrapper.childNodes[0].childNodes[0];
 
-    document.getElementsByClassName(
-        'highcharts-stocktools-wrapper'
-    )[0].style.display = 'none';
+    chart.stockTools.wrapper.style.display = 'none';
 
     // Show croshair with the label.
     controller.moveTo(200, 200);
--- a/ts/Stock/StockToolsGui.ts
+++ b/ts/Stock/StockToolsGui.ts
@@ -1415,7 +1415,7 @@ class Toolbar {
             className: PREFIX + 'stocktools-wrapper ' +
                 guiOptions.className + ' ' + bindingsClassName
         });
-        container.insertBefore(wrapper, container.childNodes[0]);
+        container.appendChild(wrapper);
 
         // Mimic event behaviour of being outside chart.container
         [
--- a/ts/Stock/StockToolsGui.ts
+++ b/ts/Stock/StockToolsGui.ts
@@ -1408,18 +1408,28 @@ class Toolbar {
             navigation = chart.options.navigation,
             bindingsClassName = navigation && navigation.bindingsClassName,
             listWrapper,
-            toolbar,
-            wrapper;
-
-        // #15593
-        css(container.parentNode, { position: 'relative' });
+            toolbar;
 
         // create main container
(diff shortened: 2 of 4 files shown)


ts/Stock/StockToolsGui.ts
```diff
@@ -1411,12 +1411,15 @@ class Toolbar {
             toolbar,
             wrapper;
 
+        // #15593
+        css(container.parentNode, { position: 'relative' });
```
pawelfus, line 1415: Usually, when we set hardcoded styles, there is a case where users come back and say they need to change it. I think it would make sense to add `stockTools.gui.style` option with this predefined  `{ position }` config

pawelfus (changes requested):
Added Sebastian to confirm I'm not missing anything.

I really appreciate the changes in TS 👍

ts/Stock/StockToolsGui.ts
khlieng: Calling it just `style` makes it sound like it gets applied to the stocktools and not `renderTo`, should it perhaps be called something like `renderToStyle`?

pawelfus: Oh, I thought you added this style to the toolbar container, not the chart's container. I'm not sure if can we do this in [`chart.style`](https://api.highcharts.com/highcharts/chart.style)? So the module would just set this config. I don't know how would we inform users that the Stock Tools module adds this style. Perhaps @TorsteinHonsi has some ideas here

TorsteinHonsi: It's dangerous to fiddle too much with the `renderTo`, as it's the implementer's element. Is there a reason why we are not using the `container` itself? It already has `position: relative`.

ts/Stock/StockToolsGui.ts
```diff
@@ -1473,9 +1476,9 @@ class Toolbar {
         // Show hide toolbar
         this.showhideBtn = showhideBtn = createElement(DIV, {
             className: PREFIX + 'toggle-toolbar ' + PREFIX + 'arrow-left'
-        }, null as any, wrapper);
+        }, void 0, wrapper);
```
sebastianbochan, line 1479: Im pretty curios why we use `void 0` instead of an empty object `{}` ?

samples/unit-tests/indicator-price-indicator/recalculations/demo.js
```diff
@@ -185,6 +185,10 @@ function (assert) {
         controller = new TestController(chart),
         button = chart.stockTools.listWrapper.childNodes[0].childNodes[0];
 
+    document.getElementsByClassName(
```
pawelfus, line 188: In tests, it's better to use `chart.container.getElementsByClassName()` instead of `document.getElementsByClassName()`, because we have templates available in karma (which are separate charts). Alternatively, you can use `chart.stockTools.wrapper` which should be a reference directly to this DOM element.

pawelfus (changes requested):
It seems this PR breaks exporting: https://jsfiddle.net/BlackLabel/g40s5vzo/2/ - try to download the first and the second chart. The second one works fine. I'm sorry, I don't know why 😓  Because extra HTML is available in the internal container?

pawelfus (approved):
Looks good, thanks! 👍 
I have one question though:
> // Mimic event behaviour of being outside chart.container

What kind of behaviour is being mimicked? Is it somehow related to the tooltip? Perhaps we can reuse `no-tooltip` class instead?

khlieng:
Yes, but its also about preventing clicks on a submenu from hitting the chart, to not start drawing an annotation when you click the button. We could probably replace some of it with `no-tooltip` when that gets merged.

ts/Stock/StockToolsGui.ts
TorsteinHonsi: When we don't have any styles to add, it is more effective to add undefined than to create an empty object, which internally in `createElement` will be truthy and passed over to `css`, where again it will be truthy and inspected for keys etc.

TorsteinHonsi (approved):
Looks good!

PR #15599, opened 2021-04-28
Fixed #15521, inactive state remained after kbd nav.

### LINKED ISSUE #15521: Chart remains to be greyed out even after losing focus
#### Expected behaviour
Chart is not greyed out after losing focus

#### Actual behaviour
Data points of Donut, pie and bubble charts remain to be greyed out after losing focus

https://jsfiddle.net/franklinefrancis/1a3xwLh9/

1. Focus on any data point
2. Use left/right arrow keys to navigate - inactive data points are greyed out
3. Press Tab or Shift+Tab - all data points are greyed out unless mouse is moved

#### Product version
8.2.0

#### Affected browser(s)
All browsers

oysteinmoseng:
Fixed #15521, inactive state remained on points after exiting chart with keyboard navigation.


TorsteinHonsi (approved):
Thanks!

PR #15602, opened 2021-04-28
Migration: SVGRenderer

bre1470:
- Added Core/Renderer/RendererRegistry to handle Highcharts.Renderer
- Added Core/Renderer/SVG/SymbolRegistry to handle Core/Renderer/SVG/Symbols
- Removed SVGRenderer from internal namespace.
- Removed FontMetrics from internal namespace.
- Removed SVGDefinitionObject.
- Removed SymbolFunction and SymbolOptions from internal namespace.


bre1470:
@TorsteinHonsi Overhead before symbol registry was 562 bytes, this increased to 1026 bytes (zipped 89 bytes to 249 bytes). I wonder, if this overhead is worth it. Should I build everything back into `SVGRenderer.prototype`?

TorsteinHonsi:
> I wonder, if this overhead is worth it.
It depends, I am wondering if we need the `SymbolRegistry.register` function instead of just a simple `Symbols` object with `Record<string, SymbolFunction>` where we can add the functions directly? What are the benefits of using an abstraction? Probably it causes at least a part of the extra bytesize.

TorsteinHonsi:
Besides, we risk breaking stuff for those who added custom symbols to `Highcharts.SVGRenderer.prototype.symbols`.

bre1470:
> > I wonder, if this overhead is worth it.
> It depends, I am wondering if we need the `SymbolRegistry.register` function instead of just a simple `Symbols` object with `Record<string, SymbolFunction>` where we can add the functions directly? What are the benefits of using an abstraction? Probably it causes at least a part of the extra bytesize.
It is necessary to comply with ESM specs as Symbols themself are readonly, unless we keep the extra namespace containing just for the symbols constant.
> Besides, we risk breaking stuff for those who added custom symbols to Highcharts.SVGRenderer.prototype.symbols.
That is safe, as Highcharts.SVGRenderer.prototype.symbols is the reference to the registry.

I will cut the registry but keep the extra Symbols and type files to see, if the overhead is gone.

TorsteinHonsi (approved):

PR #15615, opened 2021-04-29
bugfix/14742-scrolling-navigator-range

### LINKED ISSUE #14742: Chart-Scrolling not working when xAxis-Range is equal to range of initial data
#### Expected behaviour
When adding new data with a given range, the chart should scroll and only show the range

#### Actual behaviour
When you set the range exactly like the initial data the scrolling does not work

if you look here, the scrolling works because the range is smaller than the initial data:
(initial data is one minute, range is only 40 seconds):
https://stackblitz.com/edit/ionic-v4-angular-tabs-kcdmqy?file=src/app/tab1/tab1.page.ts

however if i set the range to one minute (which is desired), the chart gets squeezed when updating and does not scroll anymore:
https://stackblitz.com/edit/ionic-v4-angular-tabs-phr1tk?file=src/app/tab1/tab1.page.ts

#### Product version

latest

#### Affected browser(s)
safari/chrome macos

karolkolodziej:
Fixed #14742, the axis `range` property was ignored when there were fewer data points.

### DIFF
--- a/ts/Core/Navigator.ts
+++ b/ts/Core/Navigator.ts
@@ -218,6 +218,10 @@ declare global {
                 addEvent: boolean,
                 redraw?: boolean
             ): void;
+            public shouldStickToMin(
+                baseSeries: Series,
+                navigator: Navigator
+            ): boolean|undefined;
         }
     }
 }
@@ -2527,8 +2531,7 @@ class Navigator {
     public updatedDataHandler(this: Series): void {
         const navigator = this.chart.navigator as Highcharts.Navigator,
             baseSeries = this,
-            navigatorSeries = this.navigatorSeries,
-            xDataMin = navigator.getBaseSeriesMin((baseSeries.xData as any)[0]);
+            navigatorSeries = this.navigatorSeries;
 
         // If the scrollbar is scrolled all the way to the right, keep right as
         // new data  comes in.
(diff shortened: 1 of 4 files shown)


ts/Core/Navigator.ts
```diff
@@ -2555,6 +2553,36 @@ class Navigator {
         }
     }
 
+    /**
+     * Detect if the zoomed area should stick to the minimum or maximum, #14742.
```
pawellysy, line 2557: Is this function also checking, if the zoomed area should stick to maximum? If yes, the function name should be renamed e.g. to `shouldStickToExtreme`. If not, then change the comment.

pawellysy (approved):
See comment

pawellysy (changes requested):
see comment

ts/Extensions/RangeSelector.ts
```diff
@@ -950,6 +950,12 @@ class RangeSelector {
                 return;
             }
         } else if (type === 'all' && baseAxis) {
+            // If the navigator exist and the axis range is declared reset that
+            // range and from now on only use the range set by a user, #14742.
+            if (chart.navigator) {
+                chart.navigator.baseSeries[0].xAxis.options.range = void 0 as any;
```
pawelfus, line 956: Please, check if `baseSeries[]` is defined. I know, it might not be a real use-case, but it's better to not introduce this bug: https://jsfiddle.net/BlackLabel/4nq0ca2b/ (click on "All" button and observe console)

pawelfus (changes requested):

TorsteinHonsi (changes requested):
Thanks! Here's a blocking regression:
1. Open https://jsfiddle.net/highcharts/gbhx09jq/
2. Click the left navigator arrow once so that the navigator range is disconnected from the data max.
3. Click the "Add points" button five times or more. The navigator range is now squeezed. It should maintain the same range and stick to the data min.

raf18seb (approved):
Looks good 👍

TorsteinHonsi:
One last thing... Can you also add the case reported in my previous comments to the unit tests? It slipped through the cracks the first time, so we must make sure we cover that case in the future.

pawellysy (approved):
Looks good!

TorsteinHonsi (approved):

PR #15616, opened 2021-04-29
bugfix/15607-random-zero-stackLabels

### LINKED ISSUE #15607: Random 0(Zeroes) are displaying when zooming in the series
#### Expected behaviour
we dont want to display these random 0's

#### Actual behaviour
we dont have 0's in series, but when zooming into series we get random o's  using the stacklabels, its confusing our users if something is broken.

https://jsfiddle.net/6r0udyev

#### Product version

#### Affected browser(s)
Chrome
image

khlieng:
Fixed #15607, random stack labels with the value 0 showed when zooming in.

### DIFF
--- a/ts/Extensions/OverlappingDataLabels.ts
+++ b/ts/Extensions/OverlappingDataLabels.ts
@@ -69,7 +69,12 @@ addEvent(Chart, 'render', function collectAndHide(): void {
                 objectEach(stack, function (
                     stackItem: Highcharts.StackItem
                 ): void {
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (approved):
Thanks!

AJTSN:
i still see same issue with my sample,what's the fix?

khlieng:
The fix is not released yet, heres what it looks like on master: https://jsfiddle.net/qnys6mop/1/

PR #15625, opened 2021-04-30
bugfix/14403-tooltip-covers-elements

### LINKED ISSUE #14403: Tooltip covers clickable elements
#### Expected behaviour
The tooltip should be hidden when hovering clickable element like buttons.

#### Actual behaviour
The buttons are still clickable through the tooltip, but it is unexpected.

- https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/demo/treemap-large-dataset/
  - Click South-East Asia
  - Hover the Back button. The Bangladesh-tooltip now covers the button
- https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/demo/line-time-series/
  - Zoom in on the right part of the chart (with the high values)
  - Hover the Zoom out button. If there are values close to the button, the tooltip will cover the button.
- https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/annotations/gui-buttons/
  - Click the TXT button to add a text label
  - Click a point. A label is added, but the tooltip immediately covers it.

#### Discussion
These cases represent different elements where the tooltip should yield. The first two are buttons, so we could have a general rule that all buttons should make the tooltip hide. The last case is a bit different though - as long as we're allowed to edit the label we need to be able to click it, but when just viewing a chart, we may argue that the tooltip should show for the point connected to the label.  

Under any circumstance, perhaps adding a general CSS class, like `highcharts-hide-tooltip`, would be a solution. Then this class name would be applied to all buttons, and optionally to other elements, for example text labels. We have functionality in the `Pointer.onDocumentMouseMove` to hide the tooltip when outside the plot area, so we can probably easily extend this to check if the mouse is over a `highcharts-hide-tooltip` element (using `this.inClass`).

khlieng:
Fixed #14403, tooltip covered interactive elements like the legend, buttons or annotations. Added a CSS class, `highcharts-no-tooltip` that can be applied to elements to prevent the data tooltip covering them.

@pawelfus where is this popup tooltip hiding mechanism? The tooltip showed behind the popup and on top of it if `outside` was `true`.


pawelfus:
> where is this popup tooltip hiding mechanism? The tooltip showed behind the popup and on top of it if outside was true.

Just checked, it seems there is nothing we can replace - as you said tooltip was displayed under the popup.

ts/Core/Legend.ts
```diff
@@ -1237,6 +1237,7 @@ class Legend {
              * @type {Highcharts.SVGElement}
              */
             legend.group = legendGroup = renderer.g('legend')
+                .addClass('highcharts-no-tooltip')
```
pawelfus, line 1240: Q: Should add an option `legend.className`, similar to annotation's label? So users can switch back to the old behaviour?

pawelfus (changes requested):
Looks good, thanks! Just tests are missing

ts/Core/Legend.ts
TorsteinHonsi: Agreed. It could also be convenient for styling purposes.

ts/Core/Renderer/SVG/SVGLabel.ts
```diff
@@ -98,6 +98,8 @@ class SVGLabel extends SVGElement {
 
         if (className !== 'button') {
             this.addClass('highcharts-label');
+        } else {
+            this.addClass('highcharts-no-tooltip');
```
TorsteinHonsi, line 102: Can be micro optimized by putting a ternary inside the `addClass` call
```ts
this.addClass(
            className === 'button' ?
                'highcharts-no-tooltip' :
                'highcharts-label'
        );
```

TorsteinHonsi (changes requested):
Looks great! In addition to my inline comment:

* I agree with Pawel's suggestion on adding a `legend.className` option.
* In the [time-series](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/demo/line-time-series/) demo, the new functionality doesn't work because the accessibility module is adding a proxy button. Apparently the class name(s) are not copied over. Maybe we need to copy only the `highcharts-no-tooltip` class name over to avoid unforeseen styling issues. @oysteinmoseng FYI.

khlieng:
> Please check the failing test in Firefox, otherwise perfect 👍

Any idea whats wrong with it? 🤔

TorsteinHonsi:
> Any idea whats wrong with it?

@khlieng Apparently the green label with "custom format" is not visible at all in Firefox. It has opacity=0 and a CSS class indicating it is hidden, though the `visibility` attribute says `visible`. It may be a quirk with the annotations.

samples/unit-tests/annotations/annotations-dynamic/demo.js
```diff
@@ -204,6 +204,22 @@ QUnit.test("Annotation's dynamic methods", function (assert) {
         'blue',
         '#15424: Fourth line should be blue (lineColor)'
     );
+
+    // TODO: Fails in firefox on CI
```
TorsteinHonsi, line 208: We need to resolve this one before we merge

TorsteinHonsi (changes requested):

samples/unit-tests/annotations/annotations-dynamic/demo.js
khlieng: I replaced it with just checking the class, we already have some `SVGLabel` hovering coverage in the button test.

TorsteinHonsi (approved):

PR #15626, opened 2021-05-02
Fixed #14320, the table was not re-rendered after an update.

### LINKED ISSUE #14320: Export data table update issue
#### Expected behaviour

Last properly working version is 8.1.2: https://jsfiddle.net/BlackLabel/gf8bum23/

#### Actual behaviour

In the actual version, the data table doesn't update after the series updating.

Demo of the current version: https://jsfiddle.net/BlackLabel/7vx1uykj/

#### Product version

yipei1028:
Credit to @karolkolodziej 💯

### DIFF
--- a/samples/unit-tests/export-data/basics/demo.js
+++ b/samples/unit-tests/export-data/basics/demo.js
@@ -1168,11 +1168,9 @@ QUnit.test('Point name with category (#13293)', function (assert) {
 
 QUnit.test('Toggle data table (#13690)', function (assert) {
     var chart = Highcharts.chart('container', {
-        series: [
-            {
-                data: [2, 5, 1, 6, 7, 8, 5]
-            }
-        ]
+        series: [{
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (changes requested):
Thanks!
1. I think it is ineffective to re-render the table every time the chart renders. The chart internally has mechanisms to avoid re-rendering parts that have not changed, but with this change the table would have no such mechanism. I propose adding event listeners to `updatedData` on all the affected series of the chart. These event listeners would set a flag, `isDirtyData`, on the chart level. Then only when this flag is true, the table should be rendered again.
2. Instead of the insecure `innerHTML`, use the AST like so:

https://github.com/highcharts/highcharts/blob/a8a9529a92e6ef3bebb796c4acaf4941f8ce850f/ts/Extensions/ExportData.ts#L1380-L1382

PR #15637, opened 2021-05-04
Samples: tweaked end date for stock/lazy-loading demo

goransle:
Changed the final date by 1h 59min to make it look better if the viewer were to select "1h" on the rangeselector without navigating. [Fiddle](https://jsfiddle.net/goransle/tj749wns/)


TorsteinHonsi (approved):
Thanks!

PR #15638, opened 2021-05-04
Migration: BulletSeries

marvin19:
Renamed file, moved import and variable declarations.


bre1470 (commented):
The interface in the d.ts file should be renamed, too.

bre1470 (approved):
Thank you! Looks good. 👍

Issue #14280, opened 2020-09-29
Navigation.events.closePopup does not fire on "X" click


pawelfus:
#### Expected behaviour
Event should be fired when clicking on "X" button in popup.

#### Actual behaviour
Event is not fired, callback not executed.

#### Live demo with steps to reproduce
https://jsfiddle.net/BlackLabel/vqo6Lcy4/

1. Add some basic annotation, e.g. "Txt"
2. Click on annotation
3. Click on "X" button to close popup - event is not fired (and annotation does not get deselected)

#### Product version
N/A

#### Affected browser(s)
N/A

PR #15641, opened 2021-05-04
enhancement/popup-refactor

sebastianbochan:
Refactored popup, added support for inheritance.

### DIFF
--- a/ts/Extensions/Annotations/Popup.ts
+++ b/ts/Extensions/Annotations/Popup.ts
@@ -47,7 +47,7 @@ const {
 declare global {
     namespace Highcharts {
         class Popup {
-            public constructor(parentDiv: HTMLDOMElement, iconsURL: string, chart: Chart|undefined);
+            public constructor(parentDiv: HTMLDOMElement, iconsURL: string, chart?: Chart);
             public annotations: PopupAnnotationsObject;
             public container: HTMLDOMElement;
             public iconsURL: string;
@@ -69,7 +69,7 @@ declare global {
             public deselectAll(): void;
             public getFields(parentDiv: HTMLDOMElement, type: string): PopupFieldsObject;
             public getLangpack(): Record<string, string>;
-            public init(parentDiv: HTMLDOMElement, iconsURL: string, chart: Chart|undefined): void;
+            public init(parentDiv: HTMLDOMElement, iconsURL: string, chart?: Chart): void;
             public showForm(
                 type: string,
                 chart: AnnotationChart,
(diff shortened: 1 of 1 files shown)


ts/Extensions/Annotations/Popup.ts
```diff
@@ -218,18 +218,22 @@ H.Popup.prototype = {
...
+                (
+                    iconsURL.match(/png|svg|jpeg|jpg|gif/ig) ?
+                        iconsURL : iconsURL + 'close.svg'
+                ) + ')';
 
         ['click', 'touchstart'].forEach(function (eventName: string): void {
             addEvent(closeBtn, eventName, function (): void {
-
-                fireEvent(_self.chart.navigationBindings, 'closePopup');
+                _self.closePopup();
```
pawelfus, line 236: We need a `fireEvent` here. @pawellysy could you remind us what was the reason?

ts/Extensions/Annotations/Popup.ts
```diff
@@ -183,7 +183,7 @@ wrap(Pointer.prototype, 'onContainerMouseDown', function (this: Pointer, proceed
     }
 });
 
-H.Popup = function (this: Highcharts.Popup, parentDiv: HTMLDOMElement, iconsURL: string, chart: Chart): void {
+H.Popup = function (this: Highcharts.Popup, parentDiv: HTMLDOMElement, iconsURL: string, chart: Chart|undefined): void {
```
pawelfus, line 186: There are so many `chart` references in the code, are you sure it's safe? For example `chart.plotTop`, `chart.series` etc. 

It seems `this.chart` is not defined in Class, so defining this as `chart?: Chart` should be a good idea to review this all

pawelfus (changes requested):

ts/Extensions/Annotations/Popup.ts
sebastianbochan: Can we leave both calls?

ts/Extensions/Annotations/Popup.ts
sebastianbochan: Makes sense, we can replace the `chart` with the optional `chart`

ts/Extensions/Annotations/Popup.ts
pawellysy: This line was added because clicking on the 'x' button only hid the popup, and didn't call the event. 
here is the issue: https://github.com/highcharts/highcharts/issues/14280

sebastianbochan: So can we leave both? ClosePopup and fireEvent? It will be a little bit more flexible for other popups.

pawelfus: I'm not sure, let's try it

pawellysy: I think that it can be both 👍

ts/Extensions/Annotations/Popup.ts
sebastianbochan: Done

ts/Extensions/Annotations/Popup.ts
sebastianbochan: Done

pawellysy (approved):
Looks good! 👍

pawelfus (approved):
Looks good, thanks!

pawellysy (approved):
The popup works correctly, looks good! 👍

PR #15645, opened 2021-05-04
bugfix/15400-stock-clip-load-setSize

### LINKED ISSUE #15400: Bug with events.load setSize()
#### Expected behaviour
Show me all bars

#### Actual behaviour
When I calculate the height of the chart dynamically, it shows me only the last bars in the bar chart. The others come sometime later or when I change the height of the page so it has to recalculate the chart.
It works if I take the highchart.js (https://code.highcharts.com/highcharts.js) instead of highstocks.js (https://code.highcharts.com/stock/highstock.js)

http://jsfiddle.net/wmys9xkc/

#### Product version
Latest stable Highstock

#### Affected browser(s)
Google Chrome, Firefox in MacOS

khlieng:
Fixed #15400, clipping was wrong after calling `setSize` in `load` event with stock loaded.

### DIFF
--- a/samples/unit-tests/series/clip/demo.js
+++ b/samples/unit-tests/series/clip/demo.js
@@ -128,8 +128,26 @@ QUnit.test('General series clip tests', assert => {
 
 QUnit.test('Each series should have their own clip-path, (#14549).', assert => {
     const chart = Highcharts.stockChart('container', {
-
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #15651, opened 2021-05-05
bugfix/15636-draggable-guide-size

### LINKED ISSUE #15636: draggable-point guide box has wrong dimentions when not whole point is visible on the plotArea
#### Expected behaviour
When dragging the point, whose part isn't visible, the guide box should have dimensions of the actual size of the point.

#### Actual behaviour
When dragging the point, the guide box size is the size of the visible part of the point, which is not correct.

https://jsfiddle.net/BlackLabel/dmywo7jL/

move the bar with name 'Example bar chart' and notice, that the guide box, that is rendered, when you are dragging the point, that should indicate the place, that the moved guide box will be is not in correct size. 

#### Product version
v9.0.1

#### Affected browser(s)
all

khlieng:
Fixed #15636, draggable point guide box was not big enough when the point was not fully inside.

### DIFF
--- a/ts/Extensions/DraggablePoints.ts
+++ b/ts/Extensions/DraggablePoints.ts
@@ -2284,9 +2284,9 @@ Series.prototype.getGuideBox = function (
 
         if (bBox && (bBox.width || bBox.height || bBox.x || bBox.y)) {
             changed = true;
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (approved):
Thanks!

PR #15658, opened 2021-05-05
Fixed #6943, could not update axis events

### LINKED ISSUE #6943: Can't update axis events?
#### Expected behavior

Calling xAxis[0].update({ events: { ... } }); should update the event handlers such as afterSetExtremes.

#### Actual behavior

Event handlers don't change. In the demo, the original afterSetExtremes persists after we set `events` via `Axis.update`.

http://jsfiddle.net/by3hryj1/

TorsteinHonsi:
Fixed #6943, could not update axis events that were set by options.

### DIFF
--- a/ts/Core/Foundation.ts
+++ b/ts/Core/Foundation.ts
@@ -1,0 +1,87 @@
+/* *
+ *
+ *  (c) 2010-2021 Torstein Honsi
+ *
+ *  License: www.highcharts.com/license
+ *
+ *  !!!!!!! SOURCE GETS TRANSPILED BY TYPESCRIPT. EDIT TS FILE ONLY. !!!!!!!
+ *
+ * */
+
+'use strict';
+
+/* *
+ *
+ *  Imports
+ *
+ * */
+import Axis from 'Axis/Axis.js';
+import Series from 'Series/Series.js';
+
+import U from './Utilities.js';
+const {
+    addEvent,
+    isFunction,
+    objectEach,
+    removeEvent
+} = U;
+
+
+/* *
+ *
+ *  Functions
+ *
+ * */
+
+/*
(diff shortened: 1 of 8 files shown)


ts/Core/Foundation.ts
```diff
@@ -0,0 +1,87 @@
...
+ * instance. #6943, #10861.
+ */
+const registerEventOptions = (
+    component: Axis|Series
+): void => {
+
+    // A lookup over those events that are added by _options_ (not
+    // programmatically). These are updated through .update()
+    component.eventOptions = component.eventOptions || {};
+
```
bre1470, line 49: I was expecting a class structure.
```ts
class Foundation {
  eventOptions: Record<string, Function>;
  protected registerEventOptions(eventOptions: Record<string, Function>): void {…};
}
export default Foundation;
```
```ts
class Axis extends Foundation {
…
    this.registerEventOptions(this.options.events);
…
}
```
```ts
class Series extends Foundation {
…
    this.registerEventOptions(this.options.events);
…
}
```

TorsteinHonsi: Yes, eventually, when we see more patterns... Isn't this okay for a start, since we only have one function to share between the two classes?

bre1470: It is for me. I just fear, that someone else will come along and add to the temporary pattern, and eventually it will become difficult to refactor it as a class.

bre1470: I suggest to wrap it in a static class function.
```ts
class Foundation {
  public static registerEventOptions = (
    component: Axis|Series
  ): void => {…}
}
export default Foundation;
```

bre1470 (approved):

PR #15659, opened 2021-05-05
bugfix/15640-legend-state-keyboard-nav

### LINKED ISSUE #15640: Disabled legend items are losing disabled style with keyboard nav.
See https://jsfiddle.net/eh2uLw5p/

1. Navigate to legend with keyboard.
2. Disable an item with enter/space, and move to the next item.
3. Navigate back over the item again with arrow keys.
4. Navigate to the next item again.

The text color of the disabled item has changed back to normal, rather than remaining disabled. 

Also:

1. Navigate to legend with keyboard.
2. Disable an item with enter/space.

After animation, the disabled item will receive the normal state again, and the text style will no longer be disabled. The normal state does not go away by navigating to a different legend item.

khlieng:
Fixed #15640, legend items lost disabled state on keyboard navigation.

### DIFF
--- a/ts/Accessibility/Components/LegendComponent.ts
+++ b/ts/Accessibility/Components/LegendComponent.ts
@@ -25,7 +25,8 @@ const {
     addEvent,
     extend,
     find,
-    fireEvent
+    fireEvent,
+    isNumber
 } = U;
 
 import AccessibilityComponent from '../AccessibilityComponent.js';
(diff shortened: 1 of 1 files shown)


oysteinmoseng (approved):
Nice!

TorsteinHonsi (approved):
Thanks!

PR #15660, opened 2021-05-05
bugfix/15646-boost-zone-invisible

### LINKED ISSUE #15646: Boosted zone color is "invisible" when color is undefined
#### Expected behaviour
The expected behavior should be if a zone object doesn't define a color, it should fallback to the default color. Which is how the non-boosted zones seem to work.

https://github.com/highcharts/highcharts/blob/ce7bc5ffdbd79047b599e10338413d96741a8e9b/ts/Core/Series/Point.ts#L876-L880

#### Actual behaviour
In a boosted graph, if a zone object doesn't have a color, the graph within the zone is invisible. 

https://jsfiddle.net/m846aLu7/

The first graph is the non-boosted graph and the second one is the boosted graph, it works as expected in the non-boosted graph.
#### Product version

Latest
#### Affected browser(s)
All

khlieng:
Fixed #15646, zones with no color set were invisible in boosted chart.

### DIFF
--- a/ts/Extensions/Boost/WGLRenderer.ts
+++ b/ts/Extensions/Boost/WGLRenderer.ts
@@ -417,29 +417,29 @@ function GLRenderer(
             zoneColors = [];
 
             zones.forEach(function (zone, i): void {
-                if (!zoneDefColor && typeof zone.value === 'undefined') {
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (approved):
Thanks!

PR #15662, opened 2021-05-06
bugfix/15661-map-gantt-options-mutate

### LINKED ISSUE #15661: [Regression][9.1.0][Highmaps] Highmaps deletes userOptions.series
#### Expected behaviour
Highmaps should not modify `userOptions` object in any way.

#### Actual behaviour
Highmaps deletes `userOptions.series` after the initialization.
The issue is not present in v9.0.1.

#### Product version
9.1.0

khlieng:
Fixed #15661, options got mutated when creating map or gantt chart.


TorsteinHonsi (approved):
Thanks!

PR #15678, opened 2021-05-07
bugfix/scrollablePlotArea-tooltip-getValidPoints

khlieng:
Fixed a `scrollablePlotArea` regression, tooltip did not show sometimes.

Repro:
1. Open https://www.highcharts.com/samples/highcharts/chart/scrollable-plotarea
2. Scroll left


TorsteinHonsi (approved):
Thanks!

PR #15680, opened 2021-05-07
bugfix/15676-stock-scrollablePlotArea-clip

### LINKED ISSUE #15676: chart options scrollablePlotArea not working when Chart is created using highstock.js
#### Expected behaviour
According to the documentation _"Highcharts is already included in Highcharts Stock, so it not necessary to load both."_
So creating a standard highchart using Highcharts Stock should work.

#### Actual behaviour
Using the **Highcharts**-package to creating a simple chart with `scrollablePlotArea`-options works fine: https://jsfiddle.net/NielsHolt/tygn239u/2/

When creating the exact same chart using the **Highcharts Stock**-package: https://jsfiddle.net/NielsHolt/v7bgzsyo/3/ 
The scrolling effect do _not_ work correct: The first part of the chart is not drawn

 

Created with Highcharts: https://jsfiddle.net/NielsHolt/tygn239u/2/
Created with Highcharts Stock: https://jsfiddle.net/NielsHolt/v7bgzsyo/3/

#### Product version
9.1.0

#### Affected browser(s)
Tested in Chrome on Windows

khlieng:
Fixed #15676, clipping was wrong with `scrollablePlotArea` enabled and stock loaded.

### DIFF
--- a/samples/unit-tests/scrollable-plotarea/options/demo.html
+++ b/samples/unit-tests/scrollable-plotarea/options/demo.html
@@ -1,4 +1,4 @@
-<script src="https://code.highcharts.com/highcharts.js"></script>
+<script src="https://code.highcharts.com/stock/highstock.js"></script>
 
 
 <div id="qunit"></div>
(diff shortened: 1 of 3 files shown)


pawelfus (changes requested):
I'm afraid this is much more complicated than just removing clipping for the Stock chart. Take a look: https://jsfiddle.net/BlackLabel/vpuqz4wc/1/

We are in a process of refactoring clips in Stock, related to #4406 - probably after this, we would simply check if the scrollable area is active, then increase the width (if I understand the issue correctly)

PR #15682, opened 2021-05-07
Bugfix/15508-stock-tools-measure-x-annotation

### LINKED ISSUE #15508: Annotation with measure type "x" does not work as expected
Hi there!
When I'll try to implement the annotation measure tools I'd realize that when I don't want to use the range selector config, the measure tool for "x" doesn't show the labels at all:
image

But if I put the rangeSelector enabled the behavior works as expected
image

https://jsfiddle.net/fpmu4n2o/83/

#### Affected browser(s)
Google Chrome

am I not implementing it correctly?
Thanks in advance and good works anyway!

pawellysy:
Fixed #15508, the measure annotation's label wasn't correctly rendered inside the annotation.

### DIFF
--- a/samples/unit-tests/annotations/annotations-measure/demo.js
+++ b/samples/unit-tests/annotations/annotations-measure/demo.js
@@ -52,7 +52,30 @@ QUnit.test('#13664 - annotation measure on yAxis', function (assert) {
     assert.close(
         chart.annotations[0].shapesGroup.getBBox().x,
         chart.annotations[0].labels[0].graphic.anchorX,
-        5,
+        0.5,
         "Annotation's label's X position should be close to the X position of the annotation."
+    );
+
+    chart.update({
+        chart: {
+            spacingTop: 100
+        },
+        title: {
+            text: ''
+        }
+    });
+
+    assert.close(
+        chart.annotations[0].shapesGroup.getBBox().y,
+        chart.annotations[0].labels[0].graphic.anchorY,
+        0.5,
+        "Annotation's label's Y position should be close to the Y position of the annotation after updates."
+    );
+
+    assert.close(
+        chart.annotations[0].shapesGroup.getBBox().x,
+        chart.annotations[0].labels[0].graphic.anchorX,
+        0.5,
+        "Annotation's label's X position should be close to the X position of the annotation after updates."
     );
 });
--- a/samples/unit-tests/annotations/annotations-measure/demo.js
+++ b/samples/unit-tests/annotations/annotations-measure/demo.js
@@ -47,14 +47,16 @@ QUnit.test('#13664 - annotation measure on yAxis', function (assert) {
         bbox.y,
         chart.annotations[0].labels[0].graphic.anchorY,
         0.5,
-        "Annotation's label's Y position should be close to the Y position of the annotation."
+        `Annotation's label's Y position should be
+        close to the Y position of the annotation.`
     );
 
     assert.close(
         bbox.x,
         chart.annotations[0].labels[0].graphic.anchorX,
         0.5,
-        "Annotation's label's X position should be close to the X position of the annotation."
+        `Annotation's label's X position should
+        be close to the X position of the annotation.`
     );
 
     chart.update({
@@ -72,13 +74,15 @@ QUnit.test('#13664 - annotation measure on yAxis', function (assert) {
         bbox.y,
         chart.annotations[0].labels[0].graphic.anchorY,
         0.5,
-        "Annotation's label's Y position should be close to the Y position of the annotation after updates."
+        `Annotation's label's Y position should be close
+        to the Y position of the annotation after updates.`
     );
 
     assert.close(
         bbox.x,
         chart.annotations[0].labels[0].graphic.anchorX,
         0.5,
-        "Annotation's label's X position should be close to the X position of the annotation after updates."
+        `Annotation's label's X position should be close
+        to the X position of the annotation after updates.`
     );
 });
--- a/samples/unit-tests/annotations/annotations-measure/demo.js
+++ b/samples/unit-tests/annotations/annotations-measure/demo.js
@@ -29,37 +29,30 @@ QUnit.test('#13664 - annotation measure on yAxis', function (assert) {
             }
         ],
 
-        series: [
-            {
-                data: [1, 2, 3, 2, 3, 4, 5, 6, 7, 8, 3, 2, 4, 4, 4, 4, 3]
-            },
-            {
-                yAxis: 1,
-                data: [
-                    1,
-                    2,
-                    3,
-                    2,
-                    3,
-                    4,
-                    5,
-                    6,
-                    7,
-                    8,
-                    3,
-                    2,
-                    4,
-                    4,
-                    4,
-                    4,
-                    3
-                ].reverse()
-            }
-        ]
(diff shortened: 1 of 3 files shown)


samples/unit-tests/annotations/annotations-measure/demo.js
```diff
@@ -29,37 +29,53 @@ QUnit.test('#13664 - annotation measure on yAxis', function (assert) {
...
+        }]
     });
 
     assert.ok(
         chart.annotations[0].shapesGroup.getBBox().y === chart.yAxis[1].top,
         'Annotation measure should be visible on vary yaxis (#13664).'
     );
+
+    assert.close(
+        chart.annotations[0].shapesGroup.getBBox().y,
```
karolkolodziej, line 46: I would assign this to a variable in order to reduce the number of times when this method is fired.

ts/Extensions/Annotations/Types/Measure.ts
```diff
@@ -19,10 +19,15 @@ import U from '../../../Core/Utilities.js';
 const {
     extend,
     isNumber,
-    merge
+    merge,
+    pick
 } = U;
 
 /* eslint-disable no-invalid-this, valid-jsdoc */
+interface ExtendedPositionObject extends PositionObject{
```
karolkolodziej, line 27: @bre1470 could you please confirm if that is the optimal way of extending that interface?

ts/Extensions/Annotations/Types/Measure.ts
```diff
@@ -498,22 +503,20 @@ class Measure extends Annotation {
...
+                point: function (target: any): ExtendedPositionObject {
+                    const annotation: Measure = target.annotation;
+                    const options = target.options;
 
                     return {
-                        x: (inverted ? top : 10) +
-                            xAxis.toPixels(annotation.xAxisMin, !inverted),
-                        y: (inverted ? -left + 10 : top) +
-                            yAxis.toPixels(annotation.yAxisMin)
+                        x: annotation.xAxisMin,
```
karolkolodziej, line 516: IMHO the design without that extra 10px padding looks cleaner. 
If everyone agrees with that, please accept and add some comment on the visual difference.

ts/Extensions/Annotations/Annotations.ts
```diff
@@ -165,6 +165,8 @@ declare global {
             point?: (string|MockPointOptions);
             itemType?: string;
             vertical?: VerticalAlignValue;
+            xAxis?: number;
```
karolkolodziej, line 168: A few lines above- do we need this `extends` with an almost identical name?
When I merge that, everything works as expected, but maybe there is another reason for that?
``` TS
 interface AnnotationsLabelsOptions extends AnnotationsLabelOptions
```

karolkolodziej (changes requested):

ts/Extensions/Annotations/Types/Measure.ts
```diff
@@ -498,22 +503,20 @@ class Measure extends Annotation {
...
-                point: function (target: any): PositionObject {
-                    const annotation: Measure = target.annotation,
-                        chart = annotation.chart,
-                        inverted = chart.inverted,
-                        xAxis = chart.xAxis[typeOptions.xAxis],
-                        yAxis = chart.yAxis[typeOptions.yAxis],
-                        top = chart.plotTop,
-                        left = chart.plotLeft;
+                xAxis: 0,
+                yAxis: 0,
```
raf18seb, line 510: Could you double-check if it's working on multiple panes?

ts/Extensions/Annotations/Types/Measure.ts
```diff
@@ -498,22 +503,20 @@ class Measure extends Annotation {
...
+                point: function (target: any): ExtendedPositionObject {
+                    const annotation: Measure = target.annotation;
+                    const options = target.options;
```
raf18seb, line 513: Would be nice to stick with our pattern of declaring multiple variables with only one `const` key word:
```js
const annotation: Measure = target.annotation,
    options = target.options;
```

raf18seb (changes requested):
The label is correct when you create annotation by dragging the mouse from left to right, but it's not working when dragging from right to left. Could you check it?

ts/Extensions/Annotations/Types/Measure.ts
pawellysy: It does work in multiple panes 👍

ts/Extensions/Annotations/Annotations.ts
pawellysy: @bre1470 could you take a look?

samples/unit-tests/annotations/annotations-measure/demo.js
```diff
@@ -29,37 +29,53 @@ QUnit.test('#13664 - annotation measure on yAxis', function (assert) {
...
+        chart.annotations[0].labels[0].graphic.anchorY,
+        0.5,
+        "Annotation's label's Y position should be close to the Y position of the annotation."
```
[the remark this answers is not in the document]
pawellysy: I used double quotes, to use single quote as a punctuation mark in the name of the test. Isn't it desired solution?
@pawelfus

ts/Extensions/Annotations/Annotations.ts
bre1470: https://api.highcharts.com/highcharts/annotations.labelOptions is not identical to https://api.highcharts.com/highcharts/annotations.labels options.

ts/Extensions/Annotations/Types/Measure.ts
pawellysy: isn't it exactly this?
```js
{
x: number;
y: number;
xAxis?:number;
yAxis?:number;
}
```

TorsteinHonsi:
@pawellysy , regarding the PR description, see [writing content for the changelog](https://github.com/highcharts/highcharts/blob/master/changelog/readme.md#writing-content-for-the-changelog), points 5 and 6 .

ts/Extensions/Annotations/Types/Measure.ts
pawelfus: `x/yAxis?: number|string;` should be correct too. In theory `xAxis: chart.xAxis[index]` is also supported, but I have never tested it

I'm not sure if `ExtendedPositionObject` name is correct here, sorry for the confusion. Just checked, we already have `MockPointOptions` interface defined, let's reuse it

pawelfus (changes requested):

karolkolodziej (approved):
Please fix the TS and accept visual differences.

raf18seb (approved):
Thanks!

ts/Extensions/Annotations/Annotations.ts
```diff
@@ -165,6 +165,8 @@ declare global {
...
+            yAxis?: number;
```
pawelfus, line 169: Could you change it to: `yAxis?: number|string`? The same for `xAxis`

pawelfus (changes requested):
Sorry, I missed this in the previous review, could you fix this minor issue? See below
Thanks!

pawelfus (approved):
Thanks! :)

TorsteinHonsi (approved):

PR #15686, opened 2021-05-10
bugfix/15685-boost-scatter-ghost

### LINKED ISSUE #15685: Boosted Scatterplot shows Ghost points when first series data array is empty
#### Expected behaviour
Only points in series with data are rendered.

#### Actual behaviour
"Ghost" points are rendered each corresponding to the real data.

https://jsfiddle.net/zenzizen/7w9rcsay/2/

#### Product version
Highcharts 8.1.2

#### Affected browser(s)
Chrome and Firefox (at least)

#### Discussion
This only happens when the first data series array is empty.  Note that this seems to only affect graphs whose *first* series is the empty one; swapping the first and second series makes those ghost points disappear.  My workaround is to forcibly insert an empty series before it and hide it in the chart and legend - uncomment the first empty series in the series data array.
These ghost points are also "tied" in some way to the real points; removing one removes a specific ghost point and moving that real point will shift the corresponding ghost point.

khlieng:
Fixed #15685, boosted scatter chart with empty first series rendered ghost points.

### DIFF
--- a/ts/Extensions/Boost/WGLRenderer.ts
+++ b/ts/Extensions/Boost/WGLRenderer.ts
@@ -1255,10 +1255,7 @@ function GLRenderer(
 
             if (
                 s.segments.length === 0 ||
-                (
-                    (s as any).segmentslength &&
-                    s.segments[0].from === s.segments[0].to
-                )
+                s.segments[0].from === s.segments[0].to
             ) {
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (approved):
Thanks!

PR #15688, opened 2021-05-10
bugfix/15325-multi-axis-panning

### LINKED ISSUE #15325: Glitchy xy panning behavior with secondary y axis
#### Expected behaviour
When panning is enabled and multiple y axes are present, my expectation is that when panning, all series on all axes would be panned together, maintaining their positional relationship and scale relative to one another. Effectively, when zooming in and then panning, I would expect everything to remain static as you pan around the zoomed-in chart.

#### Actual behaviour
It appears as though panning is only being applied to one axis. I took a look at the panning implementation and it appears as though it assumes that only a single y axis is present ([see line 3570 in Chart.ts](https://github.com/highcharts/highcharts/blob/ecd5a15df7805601cf7585eedf8b6944e79614c5/ts/Core/Chart/Chart.ts#L3570)). Additionally, it appears that enabling panning dynamically results in a change in the autoscale of one or both axes, which is undesired and unexpected.

I was able to simply extend the JSFiddle example linked to from the [multiple axes page](https://www.highcharts.com/demo/combo-multi-axes) to enable panning in `xy` mode to demonstrate the issue. See this JSFiddle: https://jsfiddle.net/fh927wmy/ 

To reproduce both issues I've mentioned:
- Zoom in to an area and attempt to pan; observe that only one series is moved relative to the others, when all should move together
- Note in this JSFiddle that the scales have changed between this demo and the demo on the multiple axes page, due to the fact that panning is enabled (and no other changes were made).

#### Product version
HighCharts 9+

#### Affected browser(s)
Developing in Chrome.

khlieng:
Fixed #15325, panning multiple axes did not work.

### DIFF
--- a/ts/Core/Axis/Axis.ts
+++ b/ts/Core/Axis/Axis.ts
@@ -294,6 +294,7 @@ declare global {
             ordinal?: boolean;
             overscroll?: number;
             pane?: number;
+            panningEnabled: boolean;
             range?: number;
             reversed?: boolean;
             reversedStacks: boolean;
@@ -432,6 +433,7 @@ declare global {
             public options: AxisOptions;
             public overlap: boolean;
             public paddedTicks: Array<number>;
+            public panningEnabled: boolean;
             public panningState?: AxisPanningState;
             public plotLinesAndBands: Array<PlotLineOrBand>;
             public plotLinesAndBandsGroups: Record<string, SVGElement>;
@@ -1259,6 +1261,12 @@ class Axis {
          * @since     4.1
          * @apioption xAxis.crosshair.zIndex
          */
+
+        /**
+         * Whether to pan axis. If `chart.panning` is enabled, the option
+         * allows to disable panning on an individual axis.
+         */
+        panningEnabled: true,
 
         /**
          * The Z index for the axis group.
@@ -4171,6 +4179,7 @@ class Axis {
          */
         axis.reversed = pick(options.reversed, axis.reversed);
         axis.visible = options.visible;
+        axis.panningEnabled = options.panningEnabled;
         axis.zoomEnabled = options.zoomEnabled;
 
         // Initial categories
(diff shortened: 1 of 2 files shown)


ts/Core/Axis/Axis.ts
```diff
@@ -4171,6 +4179,7 @@ class Axis {
          */
         axis.reversed = pick(options.reversed, axis.reversed);
         axis.visible = options.visible;
+        axis.panningEnabled = options.panningEnabled;
```
TorsteinHonsi, line 4182: As far as I can see, we don't need to make this an instance property, because it is always the same as the option. So we can just refer to the option directly instead.

ts/Core/Chart/Chart.ts
```diff
@@ -3762,10 +3764,14 @@ class Chart {
...
 
                     // set new reference for next run:
-                    (chart as any)[mouseDown] = mousePos;
+                    nextMousePos[mouseDown] = mousePos;
                 }
             });
 
+            objectEach(nextMousePos, (pos, down): void => {
+                (chart as any)[down] = pos;
+            });
```
TorsteinHonsi, line 3773: It would be nice if we take this opportunity to refactor this. We already have `Pointer.mouseDownX` and `Pointer.mouseDownY`. Can you see if we can work on them instead of having additional and similar props on the `Chart`?

TorsteinHonsi (changes requested):
Thanks, nice work! Please see inline comments.

ts/Core/Chart/Chart.ts
khlieng: The `Chart` props are updated on every `pan` call since `pan` needs to know how far the cursor has moved since the previous call, while the `Pointer` props is the actual mousedown position. So it seems we would need some sort of state in addition to the `Pointer` props anyway.

TorsteinHonsi: Yes... Actually the `Chart` props have a misleading name. It is not the _mouse down X_, it is actually the X of the last pan event. Okay, let's just leave it for now since this PR is about something else.

TorsteinHonsi (approved):

PR #15689, opened 2021-05-10
docs/15495-padding-ordinal 

### LINKED ISSUE #15495: Docs: min/maxPadding does not explain relation with xAxis.ordinal option
#### Expected behaviour
API should explain that `Axis.ordinal` option takes precendence over`Axis.minPadding` and `Axis.maxPadding` options and ignores them. For `Axis.ordinal` users add a note about `Axis.overscroll` option. Add the explanation under `min/maxPadding` options (Stock only)

#### Actual behaviour
Missing explanation

#### Product version
Highcharts Stock

karolkolodziej:
Fixed #15495, added explanation about paddings with an ordinal axis.

### DIFF
--- a/ts/Core/Axis/Axis.ts
+++ b/ts/Core/Axis/Axis.ts
@@ -1939,7 +1939,8 @@ class Axis {
          * is set using `axis.setExtremes()`, the maxPadding will be ignored.
          *
          * @productdesc {highstock}
-         * When the axis `ordinal: true`, the `maxPadding` is ignored.
+         * For the ordinal axis, the `minPadding` and `maxPadding` are ignored
+         * use the `overscroll` instead.
          *
          * @sample {highcharts} highcharts/yaxis/maxpadding/
          *         Max padding of 0.25 on y axis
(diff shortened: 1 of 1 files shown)


ts/Core/Axis/Axis.ts
```diff
@@ -1938,6 +1938,9 @@ class Axis {
          * of the plot area. When the axis' `max` option is set or a max extreme
          * is set using `axis.setExtremes()`, the maxPadding will be ignored.
          *
+         * @productdesc {highstock}
+         * When the axis `ordinal: true`, the `maxPadding` is ignored.
```
raf18seb, line 1942: I think we could improve this sentence. What about the same what you used in `xAxis.ordinal`?
For example:
```
* For the ordinal axis, the `minPadding` and `maxPadding` are ignored -
* use the `overscroll` instead.
```

pawellysy (approved):
See @raf18seb  comment. Apart from that, looks good!

raf18seb (approved):
Thanks!

pawelfus (approved):
👍

PR #15693, opened 2021-05-10
tools/cypress-setup

goransle:
Initial Cypress setup. 

**To run locally**:
Requires highcharts-utils to be running. Run with `npx cypress open` (for the GUI) and `npx cypress run` (for the CLI).

**CI Setup**
Currently runs sequentially in Chrome and Firefox. We can also add Edge (and most Chromium-based browsers that has Linux versions), if that is necessary.


test/cypress/integration/example_spec.js
```diff
@@ -0,0 +1,36 @@
+describe('Name of the group', () => {
```
TorsteinHonsi, line 1: We usually don't do underscore file names in Highcharts...

goransle: Ah yes. Do we have a preference towards naming conventions? I see in the Cypress docs they sometimes use `example.spec.js`, but any `.js` file works so we could also omit the `spec` part.

TorsteinHonsi: In `samples/unit-tests` we just use dashes. In `test/ts-node-unit-tests/test` I see that we are using dashes _and_ dots. So I guess either works.

TorsteinHonsi (approved):
Great! Pending @pawelfus ' approval.

test/cypress/integration/example.spec.js
```diff
@@ -0,0 +1,36 @@
+describe('Name of the group', () => {
+    beforeEach(() => {
+        // reset before tests
+        cy.viewport(1000, 500);
+    });
+
+    afterEach(()=>{
+        //
+    });
```
pawelfus, line 9: I would comment this out with a note to avoid this hook

test/cypress/integration/example.spec.js
```diff
@@ -0,0 +1,36 @@
...
+
+    it('There should be a chart', () => {
+        cy.visit('/highcharts/3d/area');
```
pawelfus, line 12: Question: Should we use the `before` hook for `visit()`? Or couple it under one `it()`? I'm referring this ["best pracite"](https://docs.cypress.io/guides/references/best-practices#Having-tests-rely-on-the-state-of-previous-tests)

pawelfus (changes requested):
I think we are missing docs update (where the test files should be created, the structure of directories, etc.) Probably we can follow the current structure? Something like `cypress/integration/{pruductName}/{moduleName}/test.name.js`?

Where `{product}` can be Dashboard/Stock etc.?
Is the `{moduleName}` necessary? Sounds artificial, perhaps `{productName}/moduleName.js` would be enough?

test/cypress/integration/example.spec.js
goransle: I somehow completely missed the `before` hook. It's probably cleaner to use those, and should make it easier to move tests around 👍 .

.gitignore
```diff
@@ -69,4 +69,4 @@ tmp/
 /temp.*
 /env.file
 
-*.bak
```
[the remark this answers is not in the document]
goransle: No, it is not 😬

pawelfus (approved):
Thank you!

PR #15702, opened 2021-05-11
bugfix/15698-bad-instance-access

### LINKED ISSUE #15698: Error happened under development: Error "Highcharts is not defined" thrown when invalid HTML attributes are set under tooltip:pointFormatter
#### Expected behaviour
There should only be warning or error in console under development instead of throwing error when invalid HTML attributes are set under tooltip:pointFormatter

#### Actual behaviour
The application is broken under development when invalid HTML attributes are set under tooltip:pointFormatter

#### Screenshot
image

Below is related code in my project (never mind on the code styling though, it's a demo project....):
image

#### Product version

Highcharts 9.1.0

#### Affected browser(s)
Chrome

By the way, it seems like we introduced related changes in this PR:  #15675, @KacperMadej could you help when there's chance? Thanks a lot! :)

KacperMadej:
Fixed #15698, Highcharts instance was accessed incorrectly on error.

### DIFF
--- a/ts/Core/Utilities.ts
+++ b/ts/Core/Utilities.ts
@@ -449,7 +449,7 @@ function error(
     }
 
     fireEvent(
-        Highcharts,
+        H,
         'displayError',
         { chart, code, message, params },
         defaultHandler


TorsteinHonsi (approved):
Thanks!

PR #15704, opened 2021-05-11
bugfix/13238-panning-ordinal-lin2val-refactoring 

### LINKED ISSUE #13238: Chart zoom changes while panning on mobile devices
#### Actual behaviour
Scrollbar resets the zoom on mobile devices, I can observe the problem only if the scroll is exactly in this position:
mobile

When you open this demo on your phone, by swiping to the right from the full screen, then from the right edge to the left, the graph will constantly increase to 31 Mar, then when you try to swipe to the left from the position presented in the picture the zoom out will fire off.

Live demo:
https://jsfiddle.net/BlackLabel/me82Lcrz/show

#### Product version
Highcharts v8.0.4

#### Affected browser(s)
Chrome on Android | 80.0.3987.149

karolkolodziej:
Fixed #13238, panning on mobile devices with `ordinal` axis didn't work correctly.

Refactored `lin2val`:
- changed what value is passed to that method in case of `ordinal` axis (`Axis.ts`)
-  used `extendOrdinalPositions`  instead of `ordinalPositions` to calculate values outside of the current extremes
-  isolated `indexToVal` method - for panning on PC
- split into smaller methods


samples/unit-tests/axis/ordinal/demo.js
```diff
@@ -147,3 +147,114 @@ QUnit.test('Ordinal axis and lazy loading', function (assert) {
...
+            xAxis: {
+                min: 1585595880000
+            },
+            series: [{
+                data: data
+            }]
+        }),
+        axis = chart.xAxis[0];
+
+    function emulatePannign() {
```
pawellysy, line 246: typo

samples/unit-tests/axis/ordinal/demo.js
```diff
@@ -147,3 +147,114 @@ QUnit.test('Ordinal axis and lazy loading', function (assert) {
...
+        const newMin = axis.toValue(-20),
+            newMax = axis.toValue(400);
+
+        chart.xAxis[0].setExtremes(newMin, newMax);
```
pawellysy, line 250: is there no better way of emulating panning?

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -860,76 +952,37 @@ namespace OrdinalAxis {
...
+            if (axis.ordinal && !extendedOrdinalPositions) {
+                extendedOrdinalPositions = axis.ordinal.getExtendedPositions();
+                this.extendedOrdinalPositions = extendedOrdinalPositions;
+            }
 
-                // Loop down along the ordinal positions. When the linear
-                // equivalent of i matches an ordinal position, interpolate
-                // between the left and right values.
-                } else {
-                    while (i--) {
```
pawellysy, line 903: Now we are not handling the case if you want to get the pixel of a value, that is not included in extendedOrdinalPosition.

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -845,6 +852,91 @@ namespace OrdinalAxis {
...
+                // Distance in pixels between two points
+                // on the ordinal axis in the current zoom.
+                ordinalPointPixelInterval = axis.len / ordinalPositions.length - 1,
+                shiftIndex = val / ordinalPointPixelInterval;
+
+            // Make sure that the returned index does not exceed array
+            // boundaries. If so, return the first or the last index of array.
+            return Math.min(
+                Math.max(
+                    extendedOrdinalPositions.indexOf(axis.min as any) + shiftIndex,
```
pawellysy, line 888: I think, that `extendedOrdinalPositions.indexOf(axis.min as any)` can return -1. E.g. if you set the extremes using navigator

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -845,6 +852,91 @@ namespace OrdinalAxis {
...
+         */
+        axisProto.getIndexOfPoint = function (
+            val: number,
+            ordinalPositions: Array<number>,
+            extendedOrdinalPositions: Array<number>
+        ): number {
+            const axis = this,
+                // Distance in pixels between two points
+                // on the ordinal axis in the current zoom.
+                ordinalPointPixelInterval = axis.len / ordinalPositions.length - 1,
```
pawellysy, line 881: if axis.min and axis.max don't overlap values from extendedOrdinalPositions, this is inaccurate. Apart from the error below. I think that this whole thing should be calculated relative to some point from ordinalPositions.

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -845,6 +852,91 @@ namespace OrdinalAxis {
...
+            }
+
+            if (typeof distance !== 'undefined' &&
+                typeof ordinalPositions[i] !== 'undefined') {
+                return ordinalPositions[i] + (distance ?
+                    distance *
+                        (ordinalPositions[i + 1] - ordinalPositions[i]) :
+                    0);
+            }
+            return index;
```
pawellysy, line 937: is it even possible to get this return value?

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -845,6 +852,91 @@ namespace OrdinalAxis {
...
+         */
+        axisProto.index2val = function (index: number): number {
+            let axis = this,
+                ordinal = axis.ordinal,
+                // Context could be changed to extendedOrdinalPositions.
+                ordinalPositions = ordinal.positions;
+
+            // The visible range contains only equally spaced values.
+            if (!ordinalPositions) {
+                return index; // TODO Find better return value if ordinal doesn't exist.
```
pawellysy, line 915: is it even possible that ordinalPositions is undefined?

pawellysy (changes requested):
See comments

ts/Core/Axis/OrdinalAxis.ts
karolkolodziej: Indeed, the current implementation only works for points that are included in the `extendedOrdinalPositions` array. 
If not, we get the first point from that array.
But I'm not sure if the fix should also resolve that? 
What is the potential use case for that? When we might need that?

ts/Core/Axis/OrdinalAxis.ts
karolkolodziej: This method is not public, so the risk that the `ordinalPositions` doesn't exist are small. 
However for now this condition might be pointless because I've changed what is passed to that method. 
And it won't prevent from crashing.
I think we might remove that additional if statement? 
@pawelfus could you please advise?

Part changed in the `Axis.ts`:
``` TS
returnValue = axis.lin2val(axis.logarithmic ? returnValue : val);
```

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -845,6 +852,104 @@ namespace OrdinalAxis {
...
+        ): number {
+            const axis = this;
+            let firstPointVal,
+                firstPointX,
+                secondPointX;
+
+            if (axis.series[0].points) {
+                firstPointVal = axis.series[0].points[0].x;
+                firstPointX = axis.series[0].points[0].plotX;
+                secondPointX = axis.series[0].points[1].plotX;
```
pawellysy, line 885: what if there is no series or there is less then 2 points in this series? Is return 0 correct way of handling this?

pawellysy (approved):
See comments

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -860,76 +965,37 @@ namespace OrdinalAxis {
...
-                        val = ordinalPositions[i];
-                    } else { // split it up
-                        i = Math.floor(val);
-                        distance = val - i; // the decimal
-                    }
+            // When iterating for the first time,
+            // get the extended ordinal positional and assign them.
+            if (axis.ordinal && !extendedOrdinalPositions) {
+                extendedOrdinalPositions = axis.ordinal.getExtendedPositions();
+                this.extendedOrdinalPositions = extendedOrdinalPositions;
```
pawelfus, line 988: Could we store them in `axis.ordinal` instead?

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -860,76 +965,37 @@ namespace OrdinalAxis {
...
-                    } else if (val > i) { // out of range, panning to the right
-                        val = ordinalPositions[i];
-                    } else { // split it up
-                        i = Math.floor(val);
-                        distance = val - i; // the decimal
-                    }
+            // When iterating for the first time,
+            // get the extended ordinal positional and assign them.
+            if (axis.ordinal && !extendedOrdinalPositions) {
+                extendedOrdinalPositions = axis.ordinal.getExtendedPositions();
```
pawelfus, line 987: Could we calculate all ordinal positions, only if it's not possible to translate `val` using `ordinalPositions`?

ts/Core/Axis/OrdinalAxis.ts
pawelfus: We might not need that internally, but we will break `xAxis.toValue()`, which is widely used by developers. Compare results in  these demos:
- https://jsfiddle.net/BlackLabel/rh51sbw8/ (current)
- https://jsfiddle.net/BlackLabel/rh51sbw8/1/ (candidate)

pawelfus (changes requested):

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -845,6 +852,104 @@ namespace OrdinalAxis {
...
+         * @param {Array<number>} [ordinalPositions]
+         *        An array of points visible in the current plot area.
+         *
+         * @param {Array<number>} [extendedOrdinalPositions]
+         *        An array of all points available on the axis
+         *        for the given data set.
+         *
+         * @return {number}
+         */
+        axisProto.getIndexOfPoint = function (
```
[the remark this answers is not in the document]
karolkolodziej: That sounds like a good idea!

ts/Core/Axis/OrdinalAxis.ts
karolkolodziej: From the TS point of view, it is, so I think it should stay here.

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -530,6 +533,53 @@ namespace OrdinalAxis {
             return groupIntervalFactor;
         }
 
+        /**
+         * Get index of point inside the extended ordinal positions array.
```
pawellysy, line 537: This comment is a bit inaccurate. Should be: 
inside the ~~extended~~ ordinal positions array.

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -530,6 +533,53 @@ namespace OrdinalAxis {
...
+         *
+         * @return {number}
+         */
+        public getIndexOfPoint(
+            val: number,
+            ordinallArray: Array<number>
+        ): number {
+            const ordinal = this,
+                axis = ordinal.axis;
+            let firstPointVal = (axis.series[0].points && axis.series[0].points[0].x) || 0,
```
pawellysy, line 557: If you are later looping through all of the series (including the first one) You don't need to assign this value to `(axis.series[0].points && axis.series[0].points[0].x)`. You can assign it to 0 right away. This value will later get overwritten.

By the way, can't we just use `axis.dataMin`?
@pawelfus do you have any better idea, how to get those values?

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -854,84 +949,73 @@ namespace OrdinalAxis {
          * @param {number} val
          *        The linear abstracted value.
          *
-         * @param {boolean} [fromIndex]
+         * @param {boolean} [isInside]
          *        Translate from an index in the ordinal positions rather than a
```
pawellysy, line 953: Update comment

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -854,84 +949,73 @@ namespace OrdinalAxis {
...
-                                (ordinalOffset as any);
-                            // something between 0 and 1
-                            distance = (val - linearEquivalentLeft) /
-                                (linearEquivalentRight - linearEquivalentLeft);
-                            break;
-                        }
+                    const EOPlength = extendedOrdinalPositions.length,
+                        leftVisiblePoint = extendedOrdinalPositions[0],
+                        rightVisiblePoint = extendedOrdinalPositions[EOPlength - 1],
+                        slope = (rightVisiblePoint - leftVisiblePoint) / (EOPlength - 1);
```
pawellysy, line 997: can `EOPLength` be equal to 1?

pawellysy (changes requested):
see comments

samples/unit-tests/axis/ordinal/demo.js
```diff
@@ -147,3 +147,200 @@ QUnit.test('Ordinal axis and lazy loading', function (assert) {
...
+        lin2val(420),
+        7.4,
+        `For the pixel value higher than the first visible point, the function
+        should calculate value between that point and next using EOP array.`
+    );
+    assert.strictEqual(
+        lin2val(1000),
+        12.2,
+        `For the pixel value higher than any point in extendedOrdinalPositions,
+        array, the function should return the last element of that array.`
```
pawelfus, line 344: The test description is misleading

ts/Core/Axis/OrdinalAxis.ts
pawelfus: Or remove extra argument, as mentioned above :)

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -854,84 +949,73 @@ namespace OrdinalAxis {
...
+            // (array contains also points that are currently out of range).
+            if (!isInside) {
+                // When iterating for the first time,
+                // get the extended ordinal positional and assign them.
+                if (axis.ordinal && !extendedOrdinalPositions) {
+                    extendedOrdinalPositions = axis.ordinal.getExtendedPositions();
+                    this.ordinal.extendedOrdinalPositions = extendedOrdinalPositions;
+                }
+
+                if (ordinalPositions && extendedOrdinalPositions && extendedOrdinalPositions.length) {
```
pawelfus, line 980: Do we need a check for `ordinalPositions`? In this case, we use only `extendedOrdinalPositions`

ts/Core/Axis/OrdinalAxis.ts
pawelfus: I'm not sure, good question!
That would mean we have exactly one tick on the chart. And we somehow zoomed in. But with just one tick ordinal logic should not be used

pawelfus (changes requested):

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -65,7 +66,7 @@ declare module './Types' {
             findHigherRanks?: boolean
         ): TickPositionsArray;
         /** @deprecated */
-        lin2val(val: number, fromIndex?: boolean): number;
+        lin2val(val: number, isInside?: boolean|undefined): number;
```
raf18seb, line 69: Is `|undefined` necessary here? The `?` should be enough

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -97,7 +97,9 @@ interface OrdinalAxis extends Axis {
         closestDistance?: number,
         findHigherRanks?: boolean
     ): TickPositionsArray;
-    lin2val(val: number, fromIndex?: boolean): number;
+    index2val(val: number): number;
+    lin2val(val: number, isInside?: boolean|undefined): number;
```
raf18seb, line 101: Same as above

raf18seb (changes requested):
Excellent work! Just a small note from me :)

ts/Core/Axis/OrdinalAxis.ts
karolkolodziej: I've tried different scenarios and I must say that getting EOP length as zero isn't possible.

TorsteinHonsi (changes requested):
Thanks! In addition to the inline comment, please revert the whitespace changes in Axis.ts so we reduce the diff.

I will continue the review after the behaviour of `lin2val` is fixed.

ts/Core/Utilities.ts
```diff
@@ -1450,6 +1450,48 @@ function discardElement(element?: HTMLDOMElement): void {
...
+            // Continue searching to the right.
+            start = middle + 1;
+        } else {
+            // Continue searching to the left.
+            end = middle - 1;
+        }
+    }
+    // Key could not be found.
+    return !indirectSearch ? -1 : (middle as number);
+}
```
pawelfus, line 1493: Since only the OrdinalAxis module requires this utility method, can we move the definition to the OrdinalAxis file?

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -854,84 +949,73 @@ namespace OrdinalAxis {
...
-                            distance ?
-                                distance *
-                                (ordinalPositions[i + 1] - ordinalPositions[i]) :
-                                0
-                        ) :
-                        val
-                );
             }
-            return ret;
+            return val;
```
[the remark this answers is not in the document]
pawelfus: @karolkolodziej

pawelfus (changes requested):

ts/Core/Axis/OrdinalAxis.ts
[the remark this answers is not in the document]
karolkolodziej: I've tried to apply that simple mechanism, however, it seems like it's not that straightforward. 
It will require some additional properties, cases or if conditions that will complicate the code.  
I think that the iterative approach might be more readable in a complex calculation like this.

samples/unit-tests/axis/breaks/demo.js
```diff
@@ -821,3 +821,34 @@ QUnit.test('connectNulls and stacking', assert => {
...
+                [6, 2],
+                [7, 2],
+                [8, 2]
+            ]
+        }]
+    });
+
+    assert.close(
+        chart.xAxis[0].toValue(100),
+        0.26227848101265816,
```
pawellysy, line 850: This value seems a bit too accurate if you are using the accuracy of 0.05. Either round the tested value, or increase the accuracy

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -534,6 +578,40 @@ namespace OrdinalAxis {
...
+         */
+        public getIndexOfPoint(
+            val: number,
+            ordinalArray: Array<number>
+        ): number {
+            const ordinal = this,
+                axis = ordinal.axis,
+                firstPointVal = ordinal.positions ? ordinal.positions[0] : 0,
+                // toValue for the first point.
+                firstPointX = ordinal.slope ? ordinal.slope * axis.transA : 0;
```
pawellysy, line 604: can `axis.transA` also be undefined?

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -858,84 +981,76 @@ namespace OrdinalAxis {
...
+                return val;
+            }
 
-            } else {
+            // Convert back from modivied value to pixels.
+            val = (val - (localMin as any)) * localA;
+
+            // If the value is not inside the plot area,
+            // use the extended positions.
+            // (array contains also points that are currently out of range).
```
pawellysy, line 1006: not a clear comment. Maybe "...that are currently outside of the plotArea"?

pawellysy (changes requested):
Minor polishing

TorsteinHonsi (approved):
Looks good, thanks! Pending approval from the other reviewers.

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -858,84 +981,76 @@ namespace OrdinalAxis {
...
-                                ((ordinalSlope as any) *
-                                (i + 1)) +
-                                (ordinalOffset as any);
-                            // something between 0 and 1
-                            distance = (val - linearEquivalentLeft) /
-                                (linearEquivalentRight - linearEquivalentLeft);
-                            break;
-                        }
+                    const EOPlength = extendedOrdinalPositions.length,
+                        leftVisiblePoint = extendedOrdinalPositions[0],
```
pawellysy, line 1030: since You are taking the first and last index of extendedOrdinalPositions, the variable name is a bit misleading, because this point doesn't have to be visible (mostlikely isn't)

pawellysy (changes requested):

ts/Core/Axis/OrdinalAxis.ts
karolkolodziej: TS said that it should exist and I believe him ;)

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -858,84 +981,66 @@ namespace OrdinalAxis {
...
-                    typeof ordinalPositions[i] !== 'undefined' ?
-                        ordinalPositions[i] + (
-                            distance ?
-                                distance *
-                                (ordinalPositions[i + 1] - ordinalPositions[i]) :
-                                0
-                        ) :
-                        val
-                );
+                // For cases when the index is not in the EOP array,
```
pawelfus, line 1030: A little nitpicking here, sorry!

What is `EOP` array? It would be nice in general to avoid abbreviations. Does it make sense to name variables simply `index` and `length`? 

Similar question to `firstEOPPoint`. In `positions` we have x-values, right? Not points (as Point instance classes)

pawelfus (changes requested):

pawellysy (approved):
Nothing to add 👍

pawelfus (approved):
Thanks!

TorsteinHonsi:
* Pending @raf18seb 's review
* Before the final merge, can you have another look at the performance? A [rough benchmark](https://jsfiddle.net/highcharts/zswej8u1/) shows that on my system, this chart typically rendered in 350 ms on the last release, but with this PR it goes up to about 365 on average. Not a big deal, but we should have a quick look in case there is something going on that could be more effective. It should be noted that the same chart _without_ ordinal axis renders in ~200ms, so ordinal logic is quite expensive to begin with.

karolkolodziej:
These additional `20ms` are taken to calculate the `extendedOrdinalPositions`.
In this case, this method (`lin2val`) is fired twice to calculate the positions of the navigator which tries to render far away from the chart. 
Maybe there might be room for a small improvement however, this fix and the one from @pawellysy are part of the bigger refactoring which significantly improves the performance of the ordinal axis.

Just to show a scale of the issue:
- `lin2val` method is fired twice during the initialization
- `val2lin` is fired over 300 times

Using the binary search algorithm instead of the ordinary loop allows us to dramatically improve the performance despite the fact that we have to calculate the `extendedOrdinalPositions`.

TorsteinHonsi:
Ok, thanks! Now only pending approval from @raf18seb

TorsteinHonsi:
PS: Benchmarked the other PR (with binary search) and the results are awesome! Down from ~346 to ~257 ms on the fiddle above 🎉

raf18seb (commented):
Approved, perfect job, thanks! (after merging, but still wanted to say GJ!)

PR #15705, opened 2021-05-12
bugfix/15639-plotlinesAndBands-console-error

### LINKED ISSUE #15639: Uncaught TypeError: Cannot read property 'length' of undefined when plotband is removed
#### Expected behavior
Error is not thrown

#### Actual behaviour
When xAxis is already destroyed calling [removePlotBand](https://api.highcharts.com/class-reference/Highcharts.Axis#removePlotBand) throws an error

1. Click at "Remove plot band" button
image
2. See error in browser console
image

https://jsfiddle.net/vq5amytu/1/

#### Product version
Highcharts v9.1.0 (2021-05-03)

#### Affected browser(s)
Checked in Chrome Version 90.0.4430.93

DJTechnoo:
Fixed #15639, error when calling `removePlotBand` after destroying `Axis`.


TorsteinHonsi (changes requested):
Nice, thanks!

Can you also add a unit test? Should be sufficient to add `axis.destroy` then `removePlotBand` at the end of an existing test.

Also, while working on this part of the code, it's a nice opportunity to resolve `const` vs `let` inside the `removePlotBandOrLine` function.

DJTechnoo:
Unit test is still pending

TorsteinHonsi (approved):

PR #15707, opened 2021-05-12
bugfix/15701-tooltip-fontSize

### LINKED ISSUE #15701: Incorrect text position in tooltip
This is reproduced when use the fontFamilty in tooltip.

#### Expected behaviour
image

#### Actual behaviour
image

https://jsfiddle.net/h4rqobu2/

#### Product version
Highcharts 9.1

#### Affected browser(s)
Google Chrome 89.0.4389.128

khlieng:
Fixed #15701, tooltip text alignment broke when setting font size.

Since this improves vertical centering there will be visual diffs.

### DIFF
--- a/ts/Core/Renderer/SVG/SVGElement.ts
+++ b/ts/Core/Renderer/SVG/SVGElement.ts
@@ -23,6 +23,7 @@ import type {
     HTMLDOMElement,
     SVGDOMElement
 } from '../DOMElementType';
+import type FontMetricsObject from '../FontMetricsObject';
 import type GradientColor from '../../Color/GradientColor';
 import type RectangleObject from '../RectangleObject';
 import type ShadowOptionsObject from '../ShadowOptionsObject';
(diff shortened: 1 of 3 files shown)


TorsteinHonsi:
Demo broken down to the Renderer: https://jsfiddle.net/highcharts/ugvm5w8f/

@khlieng In the TextBuilder we already analyze the line height in order to get the SVG text layout right, so I have created a modified fix that caches the first line's height in that loop, and applies it in the `updateBoxSize` function. Should be more effective both in terms of performance and bytesize. Plus we can avoid visual diffs on the default tooltips, though there will be some legitimate diffs on non-default tooltips. Let's see how the visual tests go.

Demo with the fix applied: https://jsfiddle.net/highcharts/ugvm5w8f/1/

khlieng:
So, looking at new diffs, should we just center everything and approve it instead? Either by somehow caching `fontSize` of the first line in TextBuilder or doing a reverse fontMetrics on `firstLineHeight`.

TorsteinHonsi:
Yes, I'll do some experimenting.

TorsteinHonsi:
In my latest change I store the font metrics so we can read the baseline directly, instead of reverse-engineering it. But now we're probably back to the initial case, that all the default tooltips will change, and we get 600 diffs. I'm inclined to resign to that fact though. It is the correct outcome, although it produces a lot of diff that can hide other regressions. 

I suggest we create a preliminary workaround then do the final switch as a last-minute change before the next release.

TorsteinHonsi (approved):

PR #15709, opened 2021-05-12
bugfix/15708-heatmap-null-point-class

### LINKED ISSUE #15708: highcharts-null-point doesn't apply to the heatmap
#### Expected behaviour
Change the null point colour through the CSS properties via `highcharts-null-point` class.

#### Actual behaviour
Can't change the null point colour in styled mode.
It works in maps, in heatmap it fails.

Maps: https://jsfiddle.net/BlackLabel/r2u4gpj5/
Heatmap: https://jsfiddle.net/BlackLabel/v4z39mqg/

#### Product version
 Highcharts JS v9.1.0 (2021-05-03)

#### Affected browser(s)
All

khlieng:
Fixed #15708, heatmap null points missed `highcharts-null-point` class.

### DIFF
--- a/samples/unit-tests/series-heatmap/marker/demo.js
+++ b/samples/unit-tests/series-heatmap/marker/demo.js
@@ -20,7 +20,7 @@ QUnit.test('General marker tests', function (assert) {
         series: [
             {
                 data: [
-                    [0, 0, 1],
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #15711, opened 2021-05-12
Migration: Options

marvin19:
Moved interfaces to type definition files.


ts/Core/Axis/TreeGridAxis.ts
```diff
@@ -22,13 +21,13 @@ import type {
...
 } from '../Series/PointOptions';
 import type SizeObject from '../Renderer/SizeObject';
-import type SVGAttributes from '../Renderer/SVG/SVGAttributes';
 import Axis from './Axis.js';
 import Tick from './Tick.js';
 import Tree from '../../Gantt/Tree.js';
 import TreeGridTick from './TreeGridTick.js';
 import mixinTreeSeries from '../../Mixins/TreeSeries.js';
 const { getLevelOptions } = mixinTreeSeries;
+import type { ChartAddSeriesEventObject } from '../Chart/ChartOptions';
```
bre1470, line 30: Type imports should be sorted separately from regular imports.

ts/Core/Chart/ChartDefaults.ts
```diff
@@ -0,0 +1,6 @@
+import type ChartOptions from './ChartOptions';
+
+const chartDefaults: ChartOptions = {
+};
+
+export default chartDefaults;
```
bre1470, line 6: If you do not move the `chart: { … }` default from `Options.ts` into this file, you can delete `ChartDefaults.ts`.

ts/Core/Chart/ChartOptions.d.ts
```diff
@@ -0,0 +1,165 @@
...
+ * */
+
+declare module './Chart/ChartLike'{
+    interface ChartLike {
+        marginRight: ChartOptions['marginRight'];
+        polar: ChartOptions['polar'];
+    }
+}
+
+export interface ChartOptions {
```
bre1470, line 43: I would have expect something like `declare module '../OptionsLike' {` for `chart: ChartOptions`.

ts/Core/Globals.ts
```diff
@@ -17,6 +17,7 @@
  * */
 
 import type GlobalsLike from './GlobalsLike';
+import O from './Options.js';
```
bre1470, line 20: This should be a type import.

ts/Core/GlobalsLike.d.ts
```diff
@@ -18,6 +18,7 @@ import type Chart from './Chart/Chart';
 import type { SeriesTypeRegistry } from './Series/SeriesType';
 import type SizeObject from './Renderer/SizeObject';
 import type Time from './Time';
+import O from './Options.js';
```
bre1470, line 21: Here a type import as well.

ts/Core/MSPointer.ts
```diff
@@ -19,6 +19,7 @@ const {
     noop,
     win
 } = H;
+import O from './Options.js';
```
bre1470, line 22: This should be a type import as well. I will stop pointing these out. Please check all files, if only the type is used. It will help you to move determ the correct file after OptionsLike becomes Options (and Options becomes DefaultOptions).

bre1470 (changes requested):
The import statements need some attentions, but I suggest that you do the cleanup and file naming first, so that differences between type and namespace become clearer than in the first commit.

bre1470 (commented):
Looks good. 👍 Now the renaming is left...

bre1470 (approved):
Thank you! Nicely done. 👍

TorsteinHonsi (approved):
👍

PR #15718, opened 2021-05-14
bugfix/4406-series-clip-refactoring

### LINKED ISSUE #4406: No animation when adding/removing series in highstock

raf18seb:
Fixed #4406, the newly added series wasn't animating.

Also, the `Series.prototype.setClip` method has been refactored and divided into 2 separate methods.

TO DO:
- [ ] make sure that the changed test is correct
- [ ] add test for this issue

### DIFF
--- a/samples/unit-tests/series/clip/demo.js
+++ b/samples/unit-tests/series/clip/demo.js
@@ -103,18 +103,16 @@ QUnit.test('General series clip tests', assert => {
 
         setTimeout(() => {
             assert.strictEqual(
-                parseFloat(
-                    chart.container
-                        .querySelectorAll(
-                            chart.series[0].group
-                                .attr('clip-path')
-                                .slice(4)
-                                .slice(0, -1) + ' rect'
-                        )[0]
-                        .getAttribute('width')
-                ),
-                chart.clipBox.width,
-                "Correct clippath's width after updating chart (#13751)."
+                chart.container
+                    .querySelectorAll(
+                        chart.series[0].group
+                            .attr('clip-path')
+                            .slice(4)
+                            .slice(0, -1) + ' rect'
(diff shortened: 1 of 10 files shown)


pawelfus (approved):
Great, thanks!

karolkolodziej (approved):

pawellysy (approved):

TorsteinHonsi:
Thanks! I'm putting this on hold until after vacation, I would like to revisit it then. I think we need to take a step back and see if we can make the footprint of the clipping logic smaller. Plus the stock-specific stuff should be implemented in the core. The clip rectangle update on `Chart.render` is very similar to what we're doing in `Series.animate`.

I will try to refactor this with a smaller footprint than currently, and without Stock chart hacks. Treating stock charts the same as non-stock also means that we can ignore the code specific for non-stock charts. Perhaps we can remove `chart.clipRect` altogether and do this only on the series level.

pawelfus:
> Plus the stock-specific stuff should be implemented in the core.

That would simplify logic a lot. We just hit the wall in #16009 PR (pinch + multiple panes) - where we were able to implement most of the fix, but clipping remains broken (because of using `chart.clipRect` for all series, while one clip per pane is necessary)

PR #15724, opened 2021-05-17
Updated product names in license.txt

KacperMadej:


TorsteinHonsi (changes requested):
I think we shouldn't call them separate libraries anymore. Proposed text:
"This is the source code repository for Highsoft's JavaScript charting library Highcharts, including its Stock, Maps and Gantt packages."

TorsteinHonsi (approved):

PR #15738, opened 2021-05-18
bugfix/15730-popup-toolbar-hide-annotation

### LINKED ISSUE #15730: Stock Tools: Annotation's mini popup is not hidden after hiding all annotations
#### Expected behaviour
Annotation's mini popup should be hidden after hiding all annotations

#### Actual behaviour
Annotation's mini popup is not hidden after hiding all annotations

Steps:
- add an annotation
- selected new annotation, mini popup should show up
- click on "Hide all annotations" button

hide

#### Product 
Highcharts Stock

khlieng:
Fixed #15730, toolbar popup did not get hidden when hiding the selected annotation.


pawelfus (changes requested):
The solution looks good, but we need a test for it. We have too many fixes in Stock Tools GUI and Popup, that are untested, and manually testing this is problematic and time-consuming.

I have discussed this problem (missing tests) with Torstein, and we agreed on trying to create an isolated environment for Stock Tools, where we will be able to load CSS styles etc., that won't affect existing tests. Of course, tests will be part of CircleCI

For now, I suggest keeping this open until we solve the tests' issue, then create a test for it. Most likely you would be able to create a test for it which we could move later to the other directory, but let's not add extra work (we already have some tests that should be moved)

khlieng:
@TorsteinHonsi

TorsteinHonsi:
Pending approval by @pawelfus

pawelfus (approved):
Thanks!

TorsteinHonsi (approved):
Please resolve conflict, otherwise approved 👍

PR #15748, opened 2021-05-19
bugfix/15729-annotation-drag-selected

### LINKED ISSUE #15729: Stock Tools: After an annotation's drag&drop, annotation should remain selected
#### Expected behaviour
After drag&drop action, the annotation should remain selected (all control points should be visible, the popup should be visible)

#### Actual behaviour
Annotation is deselected. Control points and popup are hidden

Steps:
- draw an annotation
- select the annotation (popup shows up)
- drag&drop annotation
- popup and control points are missing

edit

#### Product
Highcharts Stock

khlieng:
Fixed #15729, annotation did not stay selected after dragging.


pawelfus (changes requested):
See: https://github.com/highcharts/highcharts/pull/15738#pullrequestreview-662251718

khlieng:
@pawelfus should it also stay selected after dragging a control point?

pawelfus:
> @pawelfus should it also stay selected after dragging a control point?

Yes, it should, nice catch!

ts/Extensions/Annotations/NavigationBindings.ts
```diff
@@ -571,19 +571,27 @@ class NavigationBindings {
...
+                // Element could be removed in the child action, e.g. button
+                (clickEvent.target as any).parentNode &&
+                // TO DO: Polyfill for IE11?
+                !closestPolyfill(clickEvent.target as any, '.' + PREFIX + 'popup')
+            ) {
+                fireEvent(navigation, 'closePopup');
+            } else if (activeAnnotation.cancelClick) {
+                setTimeout((): void => {
+                    activeAnnotation.cancelClick = false;
+                }, 0);
```
pawelfus, line 593: Could you add an explanation (a comment) why we need to use `setTimeout()` here? The idea of this block was to simply close popup and deselect the annotation. Then, if any button is selected, start drawing a new annotation.

pawelfus (changes requested):

pawelfus (approved):

TorsteinHonsi (approved):

PR #15749, opened 2021-05-19
Tools: Updated BS versions for karma tests.

bre1470:
The outdated Safari version of macOS Big Sur got deactivated. Took the chance to update all browser versions.


bre1470:
It is not deactivated, but simply replaced with 14.1 instances. So BS deployment is the (temporary) bottleneck.

PR #15751, opened 2021-05-19
bugfix/15746-heatmap-select-state-color

### LINKED ISSUE #15746: Heatmap: setting select: true doesn't select the point
#### Expected behaviour
When defining the data, if you specify the `select: true` property for the point in the data array, the point should be selected initially. 

#### Actual behaviour
The given point isn't selected

https://jsfiddle.net/BlackLabel/mk5x4wL1/

#### Workaround
select the point in load event via

```js
events: {
	load: function () {
		this.series[0].points[index].select();
	}
}
```
https://jsfiddle.net/BlackLabel/8dgu7vhz/

#### Product version
v9.1.0

#### Affected browser(s)
all

khlieng:
Fixed #15746, heatmap select state color did not work correctly.

### DIFF
--- a/samples/unit-tests/series-heatmap/marker/demo.js
+++ b/samples/unit-tests/series-heatmap/marker/demo.js
@@ -21,7 +21,12 @@ QUnit.test('General marker tests', function (assert) {
             {
                 data: [
                     [0, 0, null],
-                    [0, 1, 1.1],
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #15753, opened 2021-05-19
Migration: Chart

bre1470:
- Moved chart defaults into separate file.
- Moved chart factory to master.
- Moved title options to chart namespace.
- Moved AxisOptions interface into separate file.
- Optimized constant variables in chart.
- Fixed jsdoc task.


TorsteinHonsi (approved):
Thanks!

PR #15758, opened 2021-05-20
Tools: changed upload path for errors

goransle:
Changed the path errors were uploaded to to match the links provided in the console. (These were served via joomla until the new hompage was launched).


TorsteinHonsi (approved):
Thanks!

PR #15761, opened 2021-05-20
Samples/demo priorities

goransle:
Restored the order of stock and gantt demos from the deleted `index.htm` files.


TorsteinHonsi (approved):
Thanks!

PR #15765, opened 2021-05-21
bugfix/15725-annotation-steps-axis

### LINKED ISSUE #15725: Stock Tools: Adding multi-step annotation uses wrong yAxis coordinates
#### Expected behaviour
When adding multi-step annotation, and moving the mouse out of the current pane, another pane coordinates are used. The first step should decide yAxis/pane, and all other steps should reuse that

#### Actual behaviour
drag steps annotations

The same happens, when adding an annotation to the first pane, and moving the mouse to the bottom pane.

#### Product version
Highcharts Stock

khlieng:
Fixed #15725, adding annotations with multiple steps used the wrong axis coordinates when moving the cursor to a different pane.


ts/Stock/StockToolsBindings.ts
```diff
@@ -533,18 +533,17 @@ bindingsUtils.updateNthPoint = function (
         annotation: Annotation
     ): void {
         const options = annotation.options.typeOptions,
-            coords = this.chart.pointer.getCoordinates(e),
-            coordsX = this.utils.getAssignedAxis(coords.xAxis),
-            coordsY = this.utils.getAssignedAxis(coords.yAxis);
+            xAxis = isNumber(options.xAxis) && this.chart.xAxis[options.xAxis],
+            yAxis = isNumber(options.yAxis) && this.chart.yAxis[options.yAxis];
```
TorsteinHonsi, line 537: Are these options always `number|undefined` @pawelfus ? In other, related options, we can also apply the id of the associated axis as a string.

TorsteinHonsi:
Pending approval by @pawelfus and @raf18seb

ts/Stock/StockToolsBindings.ts
pawelfus: `updateNthPoint()` is an internal method and what's defined in `options.xAxis` is also set by us. And this is always a `number` (index). In fact, we don't need `isNumber(options.xAxis)` check, but probably TS requires this

pawelfus (changes requested):
Looks good, thanks!
However, the test is missing ;)

TorsteinHonsi:
Pending test as requested

raf18seb (approved):
All good (except the mentioned tests) 👍

TorsteinHonsi (approved):

PR #14920, opened 2021-01-13
bugfix/9744-legend-title-stocktools-overlap

khlieng:
Fixed #9744, stock toolbar overlapped left-aligned title and legend.

PR #15767, opened 2021-05-21
enhancement/12528-breadcrumbs-v2

karolkolodziej:
Added [breadcrumbs](https://api.highcharts.com/highcharts/drilldown.breadcrumbs) functionality for drilldown charts as well as hierarchical series like treemaps and sunburst. This replaces the drill-up and traverse-up buttons.

#### Upgrade note
The `drilldown.drillUpButton` options set is deprecated and replaced by `drilldown.breadcrumbs`. Some but not all settings are backwards compatible. See our [demo of mimicking the legacy button style](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/breadcrumbs/single-button/) with breadcrumbs. The same is true for `treemap.traverseUpButton`.


ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,1103 @@
...
+            });
+            treemapEventsAdded = true;
+        }
+    });
+
+    // Remove resize/afterSetExtremes at chart destroy
+    addEvent(Chart, 'destroy', function destroyEvents(): void {
+        if (this.breadcrumbs) {
+            this.breadcrumbs.destroyGroup();
+            this.breadcrumbs = void 0 as any;
```
TorsteinHonsi, line 1066: Why `as any`? Shouldn't Chart.breadcrumbs be `Breadcrumbs|undefined`?

ts/Core/Chart/Chart.ts
```diff
@@ -825,66 +819,10 @@ class Chart {
...
-        });
-
-        // handle added or removed series
-        if (redrawLegend && legend && legend.options.enabled) {
-            // draw legend graphics
-            legend.render();
-
-            chart.isDirtyLegend = false;
-        }
+        chart.setOffsets();
```
TorsteinHonsi, line 822: I think all the changes to Chart.ts are redundant. Unless I am missing something, the whole point of this refactoring is to add the `setOffsets` event. But the handling itself is flawed, it is not enough to set `yAxis.options.offsets`: https://jsfiddle.net/highcharts/LLExL/14497/ .

Instead, I think we must do like the range selector did, to use the `Chart` `getMargins` event and add to `this.plotTop`.

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,1103 @@
...
+    public createList(this: Breadcrumbs, e?: any): void {
+        const breadcrumbsList: Array<Array<number|null|Point|SeriesType|SVGAElement>> = this.breadcrumbsList || [],
+            chart = this.chart,
+            drilldownLevels = chart.drilldownLevels;
+            // If the list doesn't exist treat the initial series
+            // as the current level- first iteration.
+        let currentLevelNumber: number|null = breadcrumbsList.length ?
+            (breadcrumbsList[breadcrumbsList.length - 1][0] as number) : null;
+
+        if (this.isTreemap && e) {
```
TorsteinHonsi, line 481: In principle, I am not sure the Breadcrumbs class should care whether it is implemented in a Drilldown or a Treemap. The contents of `createList` and `calculateLevel` I believe belong in the Drilldown and Treemap implementations. Probably those classes should have some function that prepares the list and sends it over to the breadcrumbs, either as options, or as events if it needs to be kept updated. 

I think the litmus test for a good Breadcrumbs API would be if we can add it to a Chart and attach it to some random data structure. Can we set it up to highlight elements of a nested HTML list for example? Then we have a good modular design. If we manage that, we can drop the `isTreemap` property.

Ideally we should make it able to run without the Chart too, only depending on the Renderer, but that is probably not worthwhile.

What do you think?

TorsteinHonsi (commented):
Thanks! Now the boring last 10% of work starts :)

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,1103 @@
...
+    public defaultFormatter(this: Breadcrumbs, breadcrumb: Array<number|null|Point|SeriesType|SVGAElement>): string {
+        const breadcrumbs = this,
+            chart = breadcrumbs.chart,
+            breadcrumbsOptions = breadcrumbs.options,
+            arrow = breadcrumbsOptions.showFullPath ? '' : '◁ ',
+            lang = chart.options.lang;
+
+        return breadcrumbsOptions.formatter && breadcrumbsOptions.formatter(breadcrumb, breadcrumbs) ||
+            arrow + (
+                format(breadcrumbsOptions.format as string, { point: (breadcrumb[1] as Point) }, chart) ||
```
TorsteinHonsi, line 554: If not sure `point` is the right name, as this may also be a series in drilldown or a level in treemap. Would `level` be a better name?

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,1103 @@
...
+    /**
+     * Align the breadcrumbs group depending on the alignment.
+     *
+     * @requires  modules/breadcrumbs
+     *
+     * @function Highcharts.Breadcrumbs#alingGroup
+     * @param {Highcharts.Breadcrumbs} this
+     *        Breadcrumbs class.
+     */
+    public alignGroup(this: Breadcrumbs): void {
```
TorsteinHonsi, line 384: Most of this function is probably not needed. See how the current Drillown module uses the `relativeTo` option, which is passed directly on to `renderer.align`.

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,1103 @@
...
+     * @product highcharts
+     */
+    public static defaultBreadcrumbsOptions = {
+        /**
+         * The default padding for each button and separator in each direction.
+         *
+         * @type      {number}
+         * @since     next
+         */
+        buttonPadding: 5,
```
TorsteinHonsi, line 188: As far as I can see from the code, this is actually the space _between_ the buttons/separators. The padding of the buttons is hardcoded to 3 (see separate comment). Instead, this option should be called `buttonSpacing`, which would be consistent with the range selector option.

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,1103 @@
...
+                    ) {
+                        breadcrumbs.multipleDrillUp(breadcrumb[0] as number);
+                    }
+
+                    if (callDefaultEvent !== false && !breadcrumbsOptions.showFullPath) {
+                        breadcrumbs.multipleDrillUp(null as any);
+                    }
+                })
+                .attr({
+                    padding: 3
```
TorsteinHonsi, line 840: Avoid hard coding. In accordance with the range selector and other modules, these attributes should be exposed as `options.buttonTheme`.

TorsteinHonsi (changes requested):
When the inline comments have been discussed and resolved, we must have a look at the graphic design. I propose two different design modes:
* When `showFullPath` is false, the default design matches how the current `drillUpButton` looks. A button look, right aligned inside `plotBox`, with a bigger padding.
* When `showFullPath` is true, go for a more lightweight text look. No buttons, just text links with a simple arrow between them, like the current range selector inputs.
* This implies that the `options.buttonTheme` (see separate comment) be `undefined` by default, and changes based ont the `showFullPath` option. The `align` would also be `undefined` and depend on that option.

ts/Core/Renderer/SVG/Symbols.ts
```diff
@@ -333,6 +333,20 @@ function triangleDown(
     ];
 }
 
+function triangleRight(
```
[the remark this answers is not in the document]
TorsteinHonsi: Oh, this is _between_ the buttons. But I still think we should use a simple character.

ts/Extensions/Breadcrumbs.ts
karolkolodziej: I would like to set `this.breadcrumbs = undefined`, is there any other way to do this without using `void 0 as any`?

samples/unit-tests/drilldown/resetzoom/demo.js
```diff
@@ -178,10 +190,11 @@ QUnit.test('Drilldown and reset zoom', function (assert) {
...
     );
-    assert.ok(
-        chart.tooltip.isHidden,
-        '#14403: Tooltip should be hidden when hovering drillUpButton'
-    );
+    // @todo
+    // assert.ok(
+    //     chart.tooltip.isHidden,
+    //     '#14403: Tooltip should be hidden when hovering drillUpButton'
+    // );
```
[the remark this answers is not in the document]
pawelfus: With the current master, I think this should work out of the box (all buttons receive `highcharts-no-tooltip` class)

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,1103 @@
...
+        }
+        interface TreemapSeriesOptions {
+            breadcrumbs?: BreadcrumbsOptions;
+        }
+        interface ChartLike {
+            breadcrumbs?: Breadcrumbs;
+        }
+        interface LangOptions {
+            mainBreadcrumb?: string;
+        }
```
bre1470, line 118: The interfaces should be part of `declare module` patterns, or in case of breadcrumb-specific go into the class `namespace Breadcrumbs` right before the default export.

In general nothing new should be added to the internal namespace. It also has a deprecated tag for some time.

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,1103 @@
...
+            public init(chart: Chart, userOptions: Highcharts.BreadcrumbsOptions, isTreemap: boolean): void;
+            public redraw(): void;
+            public resetList(this: Breadcrumbs): void
+            public renderButton(breadcrumb: Array<number|Point|SVGAElement>, posX: number, posY: number):
+            SVGElement|void;
+            public renderSeparator(posX: number, posY: number): SVGElement|undefined;
+            public update(userOptions: Highcharts.BreadcrumbsOptions, redrawFlag: boolean): void;
+            public updateSingleButton(this: Breadcrumbs): void;
+            public setMargins(this: Breadcrumbs): void;
+        }
```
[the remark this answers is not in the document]
bre1470: We should not need this.

ts/Extensions/Breadcrumbs.ts
bre1470: Chart.breadcrumbs is optional in ChartLike, but does not allow `undefined` as a possible type. Extend the type there like Torstein wrote, or use here `delete chart.breadcrumbs` instead of a `void 0` assignment.

samples/unit-tests/drilldown/resetzoom/demo.js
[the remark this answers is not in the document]
TorsteinHonsi: Ok, so the `@todo` then would be to merge in the master, then uncomment this block.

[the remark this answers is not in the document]
karolkolodziej: It looks that the new functionality is working for the buttons when they are inside of the `plotArea`. In the other case, they are working similar to how our button works. Additionally, when we will move our button inside the `plotArea`, it will hide the tooltip just as the `reset zoom button` does.

ts/Core/Chart/Chart.ts
pawelfus: I remember discussing this issue with Karol and Grzesiek, they had similar problems as Sebastian had with StockTools and RangeSelector. `getMargins()` is not sufficient in some cases, and we end up with multiple events added to fix minor issues.

- In StockTools we have callbacks dedicated to: `getMargins`, `beforeRender` and `beforeRedraw` (last two added this year: https://github.com/highcharts/highcharts/pull/14920) events
- In RangeSelector, we have a small monster, called `alignElements()`, where we check other components (Captions, Legend, Exporting button), but I guess this might inevitable (collision detection). And of course, we have events added to `beforeRender` and `getMargins` (I'm curious if adding `beforeRedraw` might resolve some issues we are not aware of and/or let us slim down the RangeSelector module)
- In Navigator, we have `getMargins` and `afterSetChartSize`

Instead, I suggested refactoring that part of the `chart.redraw()` and let all external components (Legend, Title/Subtitle, RangeSelector, Navigator, StockTools, Breadcrumbs) register the space they need. The order of the events would define the order of components that reserve space. The Legend, Title/Subtitle are Core components, so they are simply added in the correct place. I might be wrong, but to me, it seems other components just find workarounds on how to solve it. Perhaps, there's a better way to solve it? 🤔

pawelfus (changes requested):
I didn't finish the review, it depends on what will be decided with `getMargins()`. If we decide to use `getMargins()` and possibly other events, then I expect some changes then

ts/Extensions/Breadcrumbs.ts
```diff
@@ -375,15 +307,14 @@ class Breadcrumbs {
...
     public breadcrumbsGroup: SVGElement = void 0 as any;
     public breadcrumbsList: Array<Array<any>> = [];
     public isDirty: boolean = true;
     public isTreemap?: boolean = void 0 as any;
     public level: number = -1;
-    public options: Highcharts.BreadcrumbsOptions = void 0 as any;
+    public options: BreadcrumbsOptions = void 0 as any;
 
-    public constructor(chart: Chart, userOptions?: Highcharts.BreadcrumbsOptions, isTreemap?: boolean) {
+    public constructor(chart: Chart, userOptions?: BreadcrumbsOptions, isTreemap?: boolean) {
```
bre1470, line 317: BreadcrumbsOptions have defaults, not used by users. Therefor user options should be of type `DeepPartial<BreadcrumbsOptions>`.

ts/Extensions/Breadcrumbs.ts
```diff
@@ -940,7 +871,7 @@ class Breadcrumbs {
     * @param {boolean} redraw
     *        Redraw flag
     */
-    public update(this: Breadcrumbs, options: Highcharts.BreadcrumbsOptions, redraw: boolean = true): void {
+    public update(this: Breadcrumbs, options: BreadcrumbsOptions, redraw: boolean = true): void {
```
bre1470, line 874: These options should be `DeepPartial` as well.

ts/Extensions/Breadcrumbs.ts
```diff
@@ -1015,7 +946,7 @@ if (!H.Breadcrumbs) {
 
         if (breadcrumbs && e.options.drilldown && e.options.drilldown.breadcrumbs) {
             breadcrumbs.isDirty = true;
-            breadcrumbs.update(e.options.drilldown.breadcrumbs as Highcharts.BreadcrumbsOptions, false);
+            breadcrumbs.update(e.options.drilldown.breadcrumbs as BreadcrumbsOptions, false);
```
bre1470, line 949: No forced cast in arguments. If you have to do this, typing is not matching.

ts/Extensions/Breadcrumbs.ts
```diff
@@ -1033,11 +964,11 @@ if (!H.Breadcrumbs) {
 
     addEvent(Chart, 'applyDrilldown', function (): void {
         const chart = this,
-            breadcrumbs = chart.breadcrumbs as Highcharts.Breadcrumbs,
+            breadcrumbs = chart.breadcrumbs,
             breadcrumbsOptions = chart.options.drilldown && chart.options.drilldown.breadcrumbs;
 
         if (!breadcrumbs) {
-            chart.breadcrumbs = new Breadcrumbs(chart as Chart, breadcrumbsOptions, false);
+            chart.breadcrumbs = new Breadcrumbs(chart, breadcrumbsOptions as any, false);
```
bre1470, line 971: No `any` cast. If you have to do this, typing is not matching.

ts/Extensions/Breadcrumbs.ts
```diff
@@ -1080,7 +1011,7 @@ if (!H.Breadcrumbs) {
 
                 if (breadcrumbs && e.options.breadcrumbs) {
                     breadcrumbs.isDirty = true;
-                    breadcrumbs.update(e.options.breadcrumbs as Highcharts.BreadcrumbsOptions, false);
+                    breadcrumbs.update(e.options.breadcrumbs as BreadcrumbsOptions, false);
```
bre1470, line 1014: No forced cast in arguments. If you have to do this, typing is not matching.

ts/Extensions/Breadcrumbs.ts
```diff
@@ -1097,6 +1028,65 @@ if (!H.Breadcrumbs) {
...
+
+interface Breadcrumbs {
+    breadcrumbsGroup: SVGElement;
+    breadcrumbsList: Array<Array<any>>;
+    chart: Chart;
+    isDirty: boolean;
+    isTreemap?: boolean;
+    level: number;
+    options: BreadcrumbsOptions;
+}
```
bre1470, line 1045: The properties of interface `Breadcrumbs` should move to the class and the interface itself deleted.

ts/Extensions/Breadcrumbs.ts
```diff
@@ -1097,6 +1028,65 @@ if (!H.Breadcrumbs) {
...
+interface BreadcrumbsButtonsFormatter {
+    (
+        breadcrumb: Array<number|null|Point|SeriesType|SVGAElement>,
+        breadcrumbs: Breadcrumbs
+    ): (string);
+}
+interface SeparatorOptions {
+    symbol?: string;
+    size?: number;
+}
```
bre1470, line 1082: Except for Breadcrumbs interface the rest should go into a `namespace Breadcrumbs`, to be accessible with the default export.

bre1470 (dismissed):

ts/Core/Chart/Chart.ts
bre1470: It would be great if we find a common API for that. A thing from my side, just lookin into the responsive module: It might be helpful if every component also takes optional a maximum to be available for them. This would allow us to cover narrowed layout situations and responsiveness.

TorsteinHonsi: I have also been thinking about this. A possible solution is to introduce a Regions concept. The regions may be called something like north, northeast, east, southeast... (you get the picture :). Every element would get assigned to a region. So the default title, and the subtitle, would be region "north". The range selector would be the same. Legend `align: right` would internally mean region east. Inside each region, the elements would have an order. This would allow us to do standardized layout calculation of all elements in the various regions first (at the point in time when we currently run `Chart.getMargins`), then find out how much is left for the plot area, which is the center region.

TorsteinHonsi: @pawelfus I see you are suggesting something further down here, but I can't comment there...

pawelfus: GitHub is duplicating my comment, I don't know why... It's the same as the one above. Regions idea looks good, do you have some API in your mind?

samples/unit-tests/drilldown/drillupbutton/demo.js
```diff
@@ -144,7 +25,7 @@ QUnit.test('Drill up button with responsive rules', function (assert) {
                     },
                     chartOptions: {
                         drilldown: {
-                            drillUpButton: {
+                            breadcrumbs: {
```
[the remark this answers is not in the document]
Izothep: Should work well now

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,1103 @@
...
+            events?: BreadcrumbsButtonsEventsOptions;
+            floating?: boolean;
+            format?: string;
+            formatter?: BreadcrumbsButtonsFormatter;
+            position?: BreadcrumbsAlignOptions;
+            separator?: SeparatorOptions;
+            showFullPath?: boolean;
+            style?: CSSObject;
+            useHTML?: boolean;
+            zIndex?: number;
```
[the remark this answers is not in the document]
Izothep: Options are updated now.

Izothep:
> When the inline comments have been discussed and resolved, we must have a look at the graphic design. I propose two different design modes:
> 
> * When `showFullPath` is false, the default design matches how the current `drillUpButton` looks. A button look, right aligned inside `plotBox`, with a bigger padding.
> * When `showFullPath` is true, go for a more lightweight text look. No buttons, just text links with a simple arrow between them, like the current range selector inputs.
> * This implies that the `options.buttonTheme` (see separate comment) be `undefined` by default, and changes based ont the `showFullPath` option. The `align` would also be `undefined` and depend on that option.
> 
> Edit: On second thought, maybe we should just go for `showFullPath` by default, and keep a sleek and minimal design. Then we can just provide a demo of how to set it up like the old button, with `showFullPath: false`, right-side alignment and a button-like design. It would be more predictable, and show more information for deep drilling/traversing.
> 
> What do you think?

Hm, I would rather go with the second idea - however in this case proper demo "how to go back to the previous version" will be needed.

ts/Core/Chart/Chart.ts
Izothep: I agree with that point. I will prepare it similar to the rangeSelector for now, until we will decide to implement `regions` approach.

Izothep:
I have made significant changes in Breadcrumbs code, it is still not a final version, however it is much closer to becoming one. @pawelfus could you please take a look at the code for eventual improvements?

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,1180 @@
...
+     * @requires  modules/breadcrumbs
+     *
+     * @function Highcharts.Breadcrumbs#createList
+     * @param {Highcharts.Breadcrumbs} this
+     *        Breadcrumbs class.
+     * @return {any}
+     *        Formatted text.
+     */
+    public createList(this: Breadcrumbs): any {
+        const breadcrumbsList: Array<Array<number|null|Point|SeriesType|SVGAElement>> = this.breadcrumbsList || [],
```
pawelfus, line 382: Q: Why we need so many different types in this array of arrays? Point/Series probably comes from Treemap vs regular drilldown, but how about the other types?

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,1180 @@
...
+                breadcrumbsList.push([0, (drilldownLevels[0].seriesOptions as Point)]);
+            }
+
+            const lastBreadcrumb = breadcrumbsList[breadcrumbsList.length - 1];
+
+            drilldownLevels.forEach(function (level): void {
+                // If level is already added to breadcrumbs list,
+                // don't add it again- drilling categories
+                // + 1 because of the wrong levels numeration
+                // in drilldownLevels array.
```
pawelfus, line 399: Q: Why it's wrong?

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,1180 @@
...
+                const button = el[2],
+                    separator = el[3];
+
+                // Remove SVG elements fromt the DOM.
+                button && button.destroy();
+                separator && separator.destroy();
+            });
+            // Clear the breadcrums list array.
+            breadcrumbsList.length = 0;
+            this.breadcrumbsGroup.destroy();
```
pawelfus, line 435: Q: Do we need to destroy each button first if we destroy the group later?

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,1180 @@
...
+     * @param {Highcharts.Breadcrumbs} this
+     *        Breadcrumbs class.
+     * @param {Highcharts.Breadcrumbs} jumpsAmount
+     *        A number of drillups that needs to be performed.
+     */
+    public jumpBy(this: Breadcrumbs, jumpsAmount: number): void {
+        const breadcrumbsList = this.breadcrumbsList;
+
+        // if defined, new level is a current level - jumps amount
+        // if not defined, new level is usually 0 - the first level in list.
```
pawelfus, line 513: Capital letters

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,1180 @@
...
+        if (breadcrumbsList.length) {
+            const jumpsNumber = defined(jumpsAmount) ?
+                breadcrumbsList[breadcrumbsList.length - 1][0] - jumpsAmount :
+                breadcrumbsList[breadcrumbsList.length - 1][0];
```
pawelfus, line 517: Instead of checking `breadcrumbsList[index][0]`, could we simply use `this.level`? The same in all other places.

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,1180 @@
...
+    /**
+     * Performs multiple drillups based on the provided number.
+     *
+     * @requires  modules/breadcrumbs
+     *
+     * @function Highcharts.Breadcrumbs#jumpBy
+     * @param {Highcharts.Breadcrumbs} this
+     *        Breadcrumbs class.
+     * @param {Highcharts.Breadcrumbs} jumpsAmount
+     *        A number of drillups that needs to be performed.
```
pawelfus, line 507: I think I'm missing something here. `jumpsAmount` suppose to be the number of drillups, yet later in the function `const jumpsNumber` is calculated to get the number of drillups.

pawelfus: Perhaps the method name should be `jumpTo(level)`?

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,1180 @@
...
+                breadcrumbsList[breadcrumbsList.length - 1][0];
+
+            if (this.options.showFullPath) {
+                for (let i = 0; i < jumpsNumber; i++) {
+                    this.jumpUp();
+                }
+            } else {
+                this.jumpUp();
+            }
+            this.chart.getMargins();
```
pawelfus, line 526: Is this `getMargins()` necessary?

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,1180 @@
...
+
+            if (!this.level) {
+                this.destroyGroup();
+                this.chart.redraw();
```
pawelfus, line 530: Will this `jumpBy` method be part of the API? I'm a bit afraid of hiding `chart.redraw()` here

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,1180 @@
...
+            this.breadcrumbsGroup.destroy();
+            this.breadcrumbsGroup = void 0 as any;
+        } else if (
+            chart.drillUpButton
+        ) {
+            breadcrumbsList.length = 0;
+            chart.drillUpButton.destroy();
+            chart.drillUpButton = void 0;
+        }
+        chart.getMargins();
```
pawelfus, line 444: Do we need `getMargins()` here?

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,1180 @@
...
+                    prevConector = breadcrumbsList[breadcrumbsList.length - 2] &&
+                        breadcrumbsList[breadcrumbsList.length - 2][3];
+
+                // Remove connector from the previous button.
+                prevConector && prevConector.destroy();
+                breadcrumbsList[breadcrumbsList.length - 2].length = 3;
+
+                // Remove SVG elements fromt the DOM.
+                button && button.destroy();
+                conector && conector.destroy();
```
pawelfus, line 585: Duplicated code from `destroyGroup`.

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,1180 @@
...
+     * @param {Highcharts.Breadcrumbs} this
+     *        Breadcrumbs class.
+     */
+    public renderSingleButton(this: Breadcrumbs): void {
+        const breadcrumbs = this,
+            chart = breadcrumbs.chart,
+            breadcrumbsList = breadcrumbs.breadcrumbsList,
+            breadcrumbsOptions = breadcrumbs.options,
+            lastBreadcrumbs = breadcrumbsList[breadcrumbsList.length - 2] &&
+                breadcrumbsList[breadcrumbsList.length - 2][2],
```
pawelfus, line 706: Q: Why sometimes it refers to the last breadcrumb as `length - 2, and sometimes `length - 1`?

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,1180 @@
...
+
+                    breadcrumb.push(separator);
+                    posX += separator ? separator.getBBox().width + buttonSpacing : 0;
+                }
+            });
+        }
+    }
+
+    /**
+     * Create a group, then draw breadcrumbs together with the separators.
```
pawelfus, line 692: Unrelated description

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,1180 @@
...
+            }
+        } else if (positionOptions.verticalAlign === 'bottom') {
+            translationY -= height + bBox.y;
+        }
+
+        if (positionOptions.align === 'right') {
+            translationX = -width;
+        } else if (positionOptions.align === 'center') {
+            translationX = -width / 2;
+        }
```
pawelfus, line 776: Q: Do we need all this logic related to `translationX` and `translationY`? Other aligned groups (Legend, ResetZoomButton, etc.) seems to be fine without extra logic, just using `align()`.

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,1180 @@
...
+        let breadcrumbs = this.breadcrumbs,
+            breadcrumbsHeight;
+        if (
+            breadcrumbs &&
+            !breadcrumbs.options.floating &&
+            breadcrumbs.breadcrumbsList.length
+        ) {
+            breadcrumbsHeight = breadcrumbs.breadcrumbsGroup ?
+                breadcrumbs.breadcrumbsGroup.getBBox().height :
+                30;
```
pawelfus, line 990: Use `buttoneTheme.height` instead (probably with padding/spacing)

pawelfus (changes requested):

ts/Maps/MapNavigationOptions.d.ts
```diff
@@ -32,7 +32,7 @@ declare module '../Core/Options' {
     }
 }
 
-export type ButtonRelativeToValue = ('plotBox'|'spacingBox');
+export type ButtonRelativeToValue = ('chart'|'plotBox'|'spacingBox');
```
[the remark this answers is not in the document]
Izothep: It is an option from the rangeSelector, it was implemented before in rangeSelector, it is just the TS correction

ts/Extensions/Breadcrumbs.ts
Izothep: The initial main level is not added, every next is added with increased index, starting from 0. That means that the first drilldown level is actually with index 0.

ts/Extensions/Breadcrumbs.ts
Izothep: I don't think we need, I have removed it from destroyGroup method.

ts/Extensions/Breadcrumbs.ts
Izothep: It is slightly different, as it removes only on button/connector. I have also removed `button.destroy` and `connector.destroy` from `destroyGroup` so it is not duplicate anymore

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,1180 @@
...
+    };
+
+    /* *
+     *
+     * Properties
+     *
+     * */
+
+    public breadcrumbsGroup: SVGElement = void 0 as any;
+    public breadcrumbsList: Array<Array<any>> = [];
```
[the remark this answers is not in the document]
Izothep: Good point, let me correct that!

ts/Extensions/Breadcrumbs.ts
Izothep: Done

ts/Extensions/Breadcrumbs.ts
Izothep: yes, it should be `jumpTo`, going to the level related to clicked breadcrumb

ts/Extensions/Breadcrumbs.ts
Izothep: nope, not necessary

ts/Extensions/Breadcrumbs.ts
Izothep: no :)

ts/Extensions/Breadcrumbs.ts
Izothep: I will move it inside the `destroyGroup` method if that's better.

ts/Extensions/Breadcrumbs.ts
Izothep: Sometimes we want to refer to the last drawn breadcrumb (before drawing the new one) or we want to get previous breadcrumbs connector - then we use -2. 

It would be much more clear when switching naming to this.level - then it will be easier to understand from what levels we are taking parameters.

ts/Extensions/Breadcrumbs.ts
Izothep: Improved with new code version

ts/Extensions/Breadcrumbs.ts
Izothep: Legend is checking doing something similar:
  `alignTo.y += rangeSelector.getHeight();`

The other difference is that legend is relative to the chart and breadcrumbs are not - because of that maybe the functionality is slightly different. We may investigate it further though.

ts/Extensions/Breadcrumbs.ts
Izothep: hm, on the other hand, we may need it, lack of it cause an error in one test

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,1202 @@
...
+            chart = breadcrumbs.chart,
+            list = breadcrumbs.list,
+            breadcrumbsOptions = breadcrumbs.options;
+
+        // A main group for the breadcrumbs.
+        if (!breadcrumbs.group && breadcrumbsOptions) {
+            breadcrumbs.group = chart.renderer
+                .g('breadcrumbs-group')
+                .addClass('highcharts-no-tooltip')
+                .addClass('highcharts-breadcrumbs')
```
pawelfus, line 624: Can be replaced with just one `addClass()` call

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,1202 @@
...
+            list.forEach(function (breadcrumb, index): void {
+                const breadcrumbButton = breadcrumb.button,
+                    separatorElement = breadcrumb.separator;
+
+                // Render a button.
+                if (!breadcrumbButton) {
+                    const button = breadcrumbs.renderButton(breadcrumb, posX, posY);
+                    breadcrumb.button = button;
+                    posX += button ? button.getBBox().width + buttonSpacing : 0;
+                    breadcrumbs.group.lastXPos = posX;
```
pawelfus, line 685: Unused property

ts/Extensions/Breadcrumbs.ts
pawelfus: Hint: Use `width` and `height`

pawelfus (changes requested):

ts/Extensions/Breadcrumbs.ts
TorsteinHonsi: Status November 16th: Still many references to treemap and drilldown inside the Breadcrumbs, which I think should be refactored.

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,1187 @@
...
+         * Variables are enclosed by curly brackets.
+         * Available values are passed in the declared point options.
+         *
+         * @type      {string|undefined}
+         * @since     next
+         * @default   undefined
+         * @sample {highcharts} highcharts/breadcrumbs/format
+         *          Display custom values in breadcrumb button.
+         */
+        format: void 0 as any,
```
TorsteinHonsi, line 198: Unnecessary any casts. Undefined is a valid type.

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,1187 @@
...
+
+            /**
+             * The X offset of the breadcrumbs button group.
+             */
+            x: 0,
+
+            /**
+             * The Y offset of the breadcrumbs button group.
+             */
+            y: 0
```
TorsteinHonsi, line 251: This doesn't match with the fact that we are reserving space for the button above the plot area. Probably the default `y` should be negative. https://jsfiddle.net/highcharts/fyqL74nj/

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,1187 @@
...
+        },
+
+        /**
+         * Options object for Breadcrumbs separator.
+         *
+         * @since     next
+         */
+        separator: {
```
TorsteinHonsi, line 259: Is this used at all now? In the demos, there is now an arrow between the items.

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,1187 @@
...
+         * CSS styles for all breadcrumbs.
+         *
+         * In styled mode, the breadcrumbs buttons are styled by the
+         * `.highcharts-range-selector-buttons .highcharts-button` rule with its
+         * different states.
+         *  @type {Highcharts.SVGAttributes}
+         *  @since     next
+         */
+        style: {
+            'fill': '#ffffff',
```
TorsteinHonsi, line 305: Use pallette.

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,1187 @@
...
+            'position': 'absolute'
```
TorsteinHonsi, line 306: Should be unnecessary. By default the button is SVG, otherwise the SVGRenderer should position it automatically.

TorsteinHonsi (changes requested):

ts/Extensions/Breadcrumbs.ts
Izothep: That's true, with the current version symbol was replaced by the string type arrows. Would you like to have the possibility to choose a symbol as a separator?

ts/Extensions/Breadcrumbs.ts
Izothep: Yes, actually, I had a question in this matter as well. 

Do we want to reserve space at all? I thought that when a group is relative to `plotBox` and similar, then we should not reserve space - buttons are inside plotArea so no extra margin is needed. 

On the other hand, I would like to reserve this space when group is relative to a `chart`. But then we need to move the button a little bit lower, so the default 0 position won't really be at the top of a chart anymore.

ts/Extensions/Breadcrumbs.ts
Izothep: Will do, thanks!

ts/Extensions/Breadcrumbs.ts
Izothep: It looks to be unnecessary, I will remove it in the next commit if all tests pass.

ts/Extensions/Breadcrumbs.ts
Izothep: corrected

ts/Extensions/Breadcrumbs.ts
Izothep: Actually, I have found your answer from some time ago:

> Oh, this is between the buttons. But I still think we should use a simple character.

So I will just go with removing symbol options and keeping full text support for separators.

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,1187 @@
...
+    defaultOptions.lang,
+    /**
+     * @optionparent lang
+     */
+    {
+        /**
+         * @since    next
+         * @product  highcharts
+         */
+        breadcrumbsToLabel: '&#8592',
```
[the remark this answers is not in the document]
Izothep: Changed accordingly

samples/unit-tests/breadcrumbs/breadcrumbs-in-treemap/demo.js
```diff
@@ -0,0 +1,115 @@
...
+    );
+    Highcharts.fireEvent(chart.breadcrumbs, 'up', { newLevel: 1 });
+
+    assert.strictEqual(
+        buttons[buttons.length - 1].textContent,
+        'Fruits',
+        'The last button should have text Fruits.'
+    );
+
+    // Highcharts.fireEvent(chart.breadcrumbs, 'up', { newLevel: 0 });
```
pawelfus, line 108: Q: Why is it commented out?

samples/unit-tests/breadcrumbs/buttons/demo.js
```diff
@@ -0,0 +1,427 @@
...
+        }
+    });
+
+    chart.series[0].points[0].doDrilldown();
+    assert.ok(
+        chart.drillUpButton.element,
+        'Initially, the breadcrumbs should be disabled and the drillUp button should exist.'
+    );
+    // @todo
+    //chart.drillUp();
```
pawelfus, line 91: Unncessary comments

samples/unit-tests/breadcrumbs/reserve-space/demo.js
```diff
@@ -0,0 +1,62 @@
...
+        'Space should be reserved after the drilldown.'
+    );
+    // @todo
+    //chart.drillUp();
+
+    // assert.strictEqual(
+    //     chart.yAxis[0].top,
+    //     10,
+    //     'Spacing should reset after drillUp.'
+    // );
```
pawelfus, line 61: Commented out test

ts/Extensions/Drilldown.ts
```diff
@@ -1089,22 +1140,33 @@ addEvent(Chart, 'afterShowResetZoom', function (): void {
...
+        const groupBox = this.breadcrumbs.group.getBBox();
+        this.breadcrumbs.group.align(
+            {
+                x: buttonOptions.position.x - bbox.width - 10,
+                y: buttonOptions.position.y,
+                width: groupBox.width,
+                align: buttonOptions.position.align
+            },
+            true,
+            breadcrumbsOptions.relativeTo
```
pawelfus, line 1169: Q: Shall we move the whole `"afterShowResetZoom"` logic to breadcrumbs? I don't like fact that Drilldown uses `Breadcrumbs.group.align()`. The same regarding the `"selection"` event below. Then perhaps we could refactor the `if()` clause above to an internal function (or a simple flag, that is set on init/update)?

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,1166 @@
...
+    public elementList: { [x: string]: Breadcrumbs.BreadcrumbElement } = {};
+    public chart: Chart;
+    public isDirty: boolean = true;
+    public level: number = 0;
+    public options: Breadcrumbs.BreadcrumbsOptions = void 0 as any;
+    public series?: TreemapSeries = void 0 as any;
+    public plotX: number = 0;
+    public plotY: number = 0;
+
+    // TO DO move merge with drillupbutton upper in chain
```
pawelfus, line 364: TO DO?

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,1166 @@
...
+
+        this.destroySingleButton();
+
+        // Destroy elements one by one. It's necessary beacause
+        // g().destroy() does not remove added HTML
+        this.destroyListElements(true);
+
+        // Clear the breadcrums list array.
+        if (this.group) {
+            this.group.destroy();
```
pawelfus, line 807: To comment above does not seem to be related to the actual code

ts/Extensions/Breadcrumbs.ts
```diff
@@ -0,0 +1,1166 @@
...
+        objectEach(
+            this.elementList,
+            (element): void => {
+                element.updated = false;
+            }
+        );
+    }
+
+    /**
+     * Reset state for all buttons in elementList.
```
pawelfus, line 882: Incorrect description. Seems to be description from `resetElementListState` method

pawelfus (changes requested):
See my comments/questions inline

ts/Extensions/Breadcrumbs.ts
Izothep: oh.

ts/Extensions/Drilldown.ts
Izothep: The code is moved. I have refactored the `if` but I have not moved it to a separate method for now.

ts/Extensions/Breadcrumbs.ts
Izothep: I believe it should be ready now.

pawelfus (approved):
Looks good, thanks!

One thing still concerns me: how the integration Treemap/Drilldown <-> Breadcrumbs should be implemented. Some of the event handlers in Treemap (`setRootNode`, `destroy`, `updated`) and in Drilldown (`afterDrilldown`, `afterDrillUp`, `update`) could be removed, and code placed directly inside those methods. The current solution seems to be clear, but I'm a bit biased..

ts/Extensions/Breadcrumbs.ts
TorsteinHonsi: I made a modification so that `y` is now automatically set based of the `floating` option.

Izothep: Ok, sounds great :)

TorsteinHonsi:
I did some modifications now, related to styling, samples, options etc. Not so much the business logic.

One remaining question - is the [Chart.prototype.showDrillUpButton](https://github.com/highcharts/highcharts/blob/71f67268a2e0a60de1d8938c384c274bda07668d/ts/Extensions/Drilldown.ts#L900-L942) used anywhere now?

Izothep:
> I did some modifications now, related to styling, samples, options etc. Not so much the business logic.
Great, thank you for that! 

> One remaining question - is the Chart.prototype.showDrillUpButton used anywhere now?

With the Breadcrumbs in place, the `Chart.prototype.showDrillUpButton` is not used and may be removed from the code. I was not sure if I should remove it though, as I think many customers used it as a custom feature, out of the standard Highcharts API. @pawelfus do you have some thoughts on this matter?

pawelfus:
> I was not sure if I should remove it though, as I think many customers used it as a custom feature, out of the standard Highcharts API. @pawelfus do you have some thoughts on this matter?
> 
> EDIT: On the other hand, we may remove it and just send info to the support team on how to create the same functionality with Breadcrumbs.

I think, if it's not part of the official API, then it can be safely removed. If there's a client who uses `showDrillUpButton()` method, then it's most likely a workaround. We can either provide a new workaround or fix the related issue.

TorsteinHonsi (changes requested):
Some more things to be done:
* In the API, only some of the `drilldown.breadcrumbs` child options are listed. Under `plotOptions.treemap` and sunburst, no breadcrumbs object is listed at all.
* The `lang.drillUpText` option is redundant and should probably be removed.
* Is the `lang.mainBreadcrumb` text used at all?
* These functions are no longer in use: `Chart.prototype.showDrillUpButton`, `Chart.prototype.getDrilldownBackText`, `TreemapSeries.renderTraverseUpButton`.
* The drilldown module adds a Chart method, `createList`. That is unfortunate because we want to avoid bloating the Chart object with module stuff, and also the name would be too general. Ideally we should have a `drilldown` composition for stuff like this, but for now it is probably better to just make `createList` a local function for the module.

Izothep:
> Some more things to be done:
> 
> * In the API, only some of the `drilldown.breadcrumbs` child options are listed. Under `plotOptions.treemap` and sunburst, no breadcrumbs object is listed at all.

Corrected

> * The `lang.drillUpText` option is redundant and should probably be removed.

We have left it as `deprecated` for now, if customers would like to still use it.

> * Is the `lang.mainBreadcrumb` text used at all?

Yes, inside the `getButtonText` method.

> * These functions are no longer in use: `Chart.prototype.showDrillUpButton`, `Chart.prototype.getDrilldownBackText`, `TreemapSeries.renderTraverseUpButton`.

Removed.

> * The drilldown module adds a Chart method, `createList`. That is unfortunate because we want to avoid bloating the Chart object with module stuff, and also the name would be too general. Ideally we should have a `drilldown` composition for stuff like this, but for now it is probably better to just make `createList` a local function for the module.

Local function is used for the now.

TorsteinHonsi (approved):
Okay, let's do this 🤞

TorsteinHonsi:
Pending a sample for standalone breadcrumbs

PR #15768, opened 2021-05-21
Migration: Annotations

marvin19:
Migration: Removed Highcharts namespace from Annotations

### DIFF
--- a/ts/Accessibility/Components/AnnotationsA11y.ts
+++ b/ts/Accessibility/Components/AnnotationsA11y.ts
@@ -13,7 +13,8 @@
 'use strict';
 
 import type Annotation from '../../Extensions/Annotations/Annotations';
-import type Point from '../../Core/Series/Point';
+import type AnnotationChart from '../../Extensions/Annotations/AnnotationChart';
+import type PointClass from '../../Core/Series/Point';
 
 import HTMLUtilities from '../Utils/HTMLUtilities.js';
 const {
@@ -30,14 +31,14 @@ const {
  * @return {Array<object>} The labels, or empty array if none.
  */
 function getChartAnnotationLabels(
-    chart: Highcharts.AnnotationChart
-): Array<Highcharts.AnnotationLabelType> {
+    chart: AnnotationChart
+): Array<Annotation.LabelType> {
     const annotations = chart.annotations || [];
 
     return annotations.reduce((
-        acc: Array<Highcharts.AnnotationLabelType>,
+        acc: Array<Annotation.LabelType>,
         cur: Annotation
-    ): Array<Highcharts.AnnotationLabelType> => {
+    ): Array<Annotation.LabelType> => {
         if (
             cur.options &&
             cur.options.visible !== false
@@ -56,7 +57,7 @@ function getChartAnnotationLabels(
  * @param {object} label The annotation label object
  * @return {string} The text in the label.
  */
-function getLabelText(label: Highcharts.AnnotationLabelType): string {
+function getLabelText(label: Annotation.LabelType): string {
     return (
         (
             label.options &&
@@ -80,7 +81,7 @@ function getLabelText(label: Highcharts.AnnotationLabelType): string {
  * @param {object} label The annotation label object to describe
  * @return {string} The description for the label.
  */
-function getAnnotationLabelDescription(label: Highcharts.AnnotationLabelType): string {
+function getAnnotationLabelDescription(label: Annotation.LabelType): string {
     const a11yDesc = (
         label.options &&
         label.options.accessibility &&
@@ -94,7 +95,7 @@ function getAnnotationLabelDescription(label: Highcharts.AnnotationLabelType): s
     const chart = label.chart;
     const labelText = getLabelText(label);
     const points = label.points as Array<Highcharts.AccessibilityPoint>;
-    const getAriaLabel = (point: Point): string => (
+    const getAriaLabel = (point: PointClass): string => (
         point.graphic &&
         point.graphic.element &&
         point.graphic.element.getAttribute('aria-label') ||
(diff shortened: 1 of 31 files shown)


ts/Extensions/Annotations/Annotations.ts
```diff
@@ -77,6 +75,40 @@ const {
...
+        annotations: Array<Annotation>;
+        controlPointsGroup: SVGElement;
+        options: Annotation.ChartOptions;
+        plotBoxClip: SVGElement;
+        addAnnotation(userOptions: AnnotationsOptions, redraw?: boolean): Annotation;
+        drawAnnotations(): void;
+        initAnnotation(userOptions: AnnotationsOptions): Annotation;
+        removeAnnotation(idOrAnnotation: (number|string|Annotation)): void;
+    }
+}
```
bre1470, line 90: `AnnotationChart` is only declared in this file, so this should be moved to `Annotation.Chart` namespace and the `AnnotationChart.d.ts` should be deleted to simplify things.

bre1470 (changes requested):

ts/Accessibility/Components/AnnotationsA11y.ts
```diff
@@ -13,7 +13,8 @@
 'use strict';
 
 import type Annotation from '../../Extensions/Annotations/Annotations';
-import type Point from '../../Core/Series/Point';
+import type AnnotationChart from '../../Extensions/Annotations/AnnotationChart';
+import type PointClass from '../../Core/Series/Point';
```
bre1470, line 17: As there is no name conflict, `import type Point` should be the favored pattern.

ts/Accessibility/Options/Options.ts
```diff
@@ -26,6 +26,7 @@ import type Series from '../../Core/Series/Series';
 
 import palette from '../../Core/Color/Palette.js';
 import ColorType from '../../Core/Color/ColorType';
+import Annotation from '../../Extensions/Annotations/Annotations';
```
bre1470, line 29: Not needed here.

ts/Extensions/Annotations/Annotations.ts
```diff
@@ -77,6 +76,26 @@ const {
  * Declarations
  *
  * */
+declare module './AnnotationsOptions'{
+    // @todo AnnotationOptions.d.ts
```
bre1470, line 80: `@todo` to do and delete declare module here.

ts/Extensions/Annotations/Annotations.ts
```diff
@@ -29,22 +31,19 @@ import type ColorString from '../../Core/Color/ColorString';
...
+import type SeriesClass from '../../Core/Series/Series';
 import type SVGElement from '../../Core/Renderer/SVG/SVGElement';
 import type SVGPath from '../../Core/Renderer/SVG/SVGPath';
 import type { SymbolKey } from '../../Core/Renderer/SVG/SymbolType';
 
 import A from '../../Core/Animation/AnimationUtilities.js';
 const { getDeferredAnimation } = A;
 import Chart from '../../Core/Chart/Chart.js';
-const chartProto: Highcharts.AnnotationChart = Chart.prototype as any;
+const chartProto: AnnotationChart = Chart.prototype as any;
```
bre1470, line 46: Use deconstruct and no `any`:
```ts
const { prototype: chartProto } = Chart;
```

ts/Extensions/Annotations/MockPoint.ts
```diff
@@ -6,12 +6,14 @@
...
 
+import type Annotation from '../Annotations/Annotations';
+import type AnnotationChart from '../Annotations/AnnotationChart';
 import type MockPointOptions from './MockPointOptions';
 import type PositionObject from '../../Core/Renderer/PositionObject';
 import Series from '../../Core/Series/Series.js';
 
 declare module './MockPointOptions' {
-    interface MockPointOptions {
+    interface MockPointOptions { // @todo -> MockPointOptions
```
bre1470, line 16: Not part of Annotation class, I know, but could you move it in the d.ts file, please? We already mock with MockPoint, so time is right for this.

ts/masters/modules/annotations-advanced.src.ts
```diff
@@ -22,3 +23,6 @@ import '../../Extensions/Annotations/Types/VerticalLine.js';
 import '../../Extensions/Annotations/Types/Measure.js';
 import '../../Extensions/Annotations/NavigationBindings.js';
 import '../../Extensions/Annotations/Popup.js';
+const G: AnyRecord = Highcharts;
+// Annotations
+G.extendAnnotation = Annotation.extendAnnotation;
```
bre1470, line 28: Define G.Annotation as well.
```ts
G.Annotation = Annotation;
```

ts/masters/modules/annotations.src.ts
```diff
@@ -10,6 +10,10 @@
...
  */
 'use strict';
-import '../../Extensions/Annotations/Annotations.js';
+import Highcharts from '../../Core/Globals.js';
+import Annotation from '../../Extensions/Annotations/Annotations.js';
 import '../../Extensions/Annotations/NavigationBindings.js';
 import '../../Extensions/Annotations/Popup.js';
+const G: AnyRecord = Highcharts;
+// Annotations
+G.extendAnnotation = Annotation.extendAnnotation;
```
bre1470, line 19: Define G.Annotation, too.

bre1470 (changes requested):
Good work! Some more fine tuning needed in the comments.

In addition, the AnnotationsOptions (and file) should be renamed to AnnotationOptions to stop the plural confusion. I you find other interfaces with `Annotations` plural prefix, please rename them without the `s` as well.

ts/masters/highcharts.src.ts
```diff
@@ -12,6 +12,7 @@ import Utilities from '../Core/Utilities.js';
 import DefaultOptions from '../Core/DefaultOptions.js';
 import Fx from '../Core/Animation/Fx.js';
 import Animation from '../Core/Animation/AnimationUtilities.js';
+import Annotation from '../Extensions/Annotations/Annotation.js';
```
bre1470, line 15: Annotation should not be part of the regular highcharts package.

bre1470 (changes requested):
One thing...

ts/Accessibility/Components/AnnotationsA11y.ts
bre1470: Oi, and this needs to be done.

PR #15770, opened 2021-05-21
bugfix/15727-annotation-popup-input-selection

### LINKED ISSUE #15727: Stock Tools: Selecting input in popup and firing event mouseup outside the popup hides it
#### Expected behaviour
Text in the input should be highlighted and editable. The popup should not hide

#### Actual behaviour
Popup disappears

drag popup

#### Product
Highcharts Stock

khlieng:
Fixed #15727, annotation popup was hidden when dragging from input to outside popup.

This PR needs #15748.


khlieng:
But we still want to be able to close the popup by clicking outside it right? I think we could just add my event handler to the popup itself instead.

pawelfus:
> But we still want to be able to close the popup by clicking outside it right?

Of course, we still want this

pawelfus:
It doesn't seem to be fixed: https://jsfiddle.net/BlackLabel/30Lpdz2r/

pawelfus:
Ah, right, my bad, sorry!

khlieng:
@TorsteinHonsi this one should be ready now.

TorsteinHonsi (approved):
Thanks! Pending final approval from @pawelfus

pawelfus (changes requested):
Thanks, looks good! Just missing a Cypress test

TorsteinHonsi:
@khlieng Pending Pawel's request

pawelfus (approved):
Thanks!

TorsteinHonsi (approved):

PR #15778, opened 2021-05-25
bugfix/15771-drilldown-crop

### LINKED ISSUE #15771: highstock drill down doesn't work on axis label for boundary conditions when chart is scrolled using navigator
We have a stock chart with a navigator. When navigator is scrolled towards right edge of x-axis such that the
data points towards left edge of x-axis are being hidden. After that if we do a drill down
on axis label which is rendered at right edge of x-axis then drill down should have worked and drill up button should have appeared.

However drill down is not working and drill up button appears without any text or value.
navigatorDrilldownIssue_LI

To reproduce the issue load any of the below demo. Drag the navigator's left handle towards right edge of the x-axis. Then try to perform drilldown by clicking the axis label that appears at the right edge of x-axis (the last one)
https://jsfiddle.net/vikas29/jewv89L3/4/
https://jsfiddle.net/BlackLabel/kjLod890/
We have even posted the issue to highcharts forums: 
https://www.highcharts.com/forum/viewtopic.php?f=12&t=46524&p=167118#p167118

we are using  Highstock JS v9.0.1

have checked on chrome

khlieng:
Fixed #15771, drilldown failed when points were cropped from the start of the series.

### DIFF
--- a/samples/unit-tests/drilldown/drilldown-general/demo.js
+++ b/samples/unit-tests/drilldown/drilldown-general/demo.js
@@ -201,6 +201,21 @@ QUnit.test('Drilldown methods', function (assert) {
         ['Cats2', 'Dogs2', 'Cows2', 'Sheep2', 'Pigs2'],
         'First drilldown after update (#7600)'
     );
+
+    chart.drillUp();
+    chart.update({
+        series: [{
+            cropThreshold: 1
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #15785, opened 2021-05-25
bugfix/15740-stock-tools-error-after-destroy

### LINKED ISSUE #15740: `chart.xAxis[0].destroy()` followed by `chart.destroy()` causes error
#### Expected behaviour
After `chart.xAxis[0].destroy()` followed by `chart.destroy()` there should be no error.

#### Actual behaviour
After `chart.xAxis[0].destroy()` followed by `chart.destroy()` the console gives the following error:
`Uncaught TypeError: Cannot read property 'offsets'`.

https://jsfiddle.net/L4ntg8c3/1/

#### Product version
Highcharts 9.1.0

#### Affected browser(s)
All

DJTechnoo:
Fixed #15740, error in some cases when destroying elements of the chart and Stock Tools were present.


TorsteinHonsi (approved):
Thanks, looks good to me! Test coverage will follow in another PR. Requestion review from @pawelfus .

TorsteinHonsi:
I'm a little bit concerned about unknown side effects of moving around the `chart.redraw` call and the event that triggers the Stock Tools. For a more atomic fix, we could simply add a `redraw` parameter to the `destroy` function, let it default to true and pass it as false when called from the chart destroy event.

```ts
addEvent(Chart, 'destroy', function (): void {
    if (this.stockTools) {
        this.stockTools.destroy(false);
    }
});
```

```ts
        // redraw
       if (redraw) {
            this.chart.isDirtyBox = true;
            this.chart.redraw();
       }
```
What do you think @pawelfus ?

pawelfus:
Hm.. is this bug some isolated part of the scenario when it breaks? `Axis.destroy()` is not part of the official API, it's an internal method. Replace with `Axis.remove()` and will work fine: https://jsfiddle.net/BlackLabel/wukem45j/

> Thanks, looks good to me! Test coverage will follow in another PR. Requestion review from @pawelfus .

Hmm.. why a separate PR for the test?

> I'm a little bit concerned about unknown side effects of moving around the chart.redraw call and the event that triggers the Stock Tools. For a more atomic fix, we could simply add a redraw parameter to the destroy function, let it default to true and pass it as false when called from the chart destroy event.

Me too. Adding `chart.redraw()` there will call `chart.update()` twice resulting in `chart.redraw` event called twice, demo: https://jsfiddle.net/BlackLabel/r1thwoxv/) - but I'm not sure if there are any side effects apart from performance. In the first place, I was surprised `destroy()` doesn't take `redraw` argument like any other func.. oh wait, Toolbar doesn't have `remove([redraw])`. In general, we have this architecture (simplified):
```ts
ClassName.remove = (redraw?) => {
  this.destroy(); 
  if (pick(redraw, true))
    this.chart.redraw();
}
```

```ts
ClassName.update = (options, redraw?) => {
  this.destroy();
  this.init();
 
  if (pick(redraw, true))
    this.chart.redraw();
} 
```

And `destroy()` just removes elements from DOM (and events, if necessary), not touching `chart.redraw()` at all. Perhaps we should do this little refactoring, which would resolve the issue?

TorsteinHonsi:
> resulting in chart.redraw event called twice, demo
Oh, thanks! I suspected it would call redraw twice, but couldn't demonstrate it. 

> Perhaps we should do this little refactoring, which would resolve the issue?
Yes I agree, it would make things more consistent. @DJTechnoo .

pawelfus (approved):
Nice!

ts/Stock/StockToolsGui.ts
```diff
@@ -1620,6 +1620,11 @@ class Toolbar {
         if (this.chart.navigationBindings) {
             this.chart.navigationBindings.update();
         }
+
+        if (pick(redraw, true)) {
+            this.chart.isDirtyBox = true;
```
TorsteinHonsi, line 1625: The idea of the dirty/redraw design is that we mark items dirty when updated, but the redraw itself is updated. So that we can optimize chart updates by doing only one chart redraw. Consequently, `isDirtyBox` must be set regardless of `redraw` (move it outside the condition).

TorsteinHonsi (changes requested):

TorsteinHonsi (approved):
Thanks! Now pending CI tests.

PR #15786, opened 2021-05-25
Migration: Tick

bre1470:
- Removed Tick types from internal namespace.
- Improved const/let pattern in Tick class.


TorsteinHonsi (approved):
Thanks!

PR #15787, opened 2021-05-25
feature/3424-zoom-on-polar

PaulDalek:
Implementing the zoom functionality on polar series.


TorsteinHonsi:
Thanks, it looks good to me so far, except all the polar functionality must me refactored out of the core files to keep the weight down. It can probably be hooked in using events.

PR #15788, opened 2021-05-26
bugfix/14780-offline-export-fallback

### LINKED ISSUE #14780: Offline export to PNG/JPEG with HTML content without fallback to export server
#### Expected behaviour
Offline export to PNG/JPEG with HTML content without fallback to export server should work as it is supported by all modern browsers

#### Actual behaviour
Offline export to PNG with HTML content without export server fails

[https://jsfiddle.net/7qdxp526/2/](https://jsfiddle.net/7qdxp526/2/)

Fiddle with default PNG Export failing and simple workaround showing that export is possible and shouldn't fail in default export

#### Product version

All

khlieng:
Fixed #14780, exporting charts with HTML content to PNG or JPEG fell back to the export server even in browsers that supported it.


TorsteinHonsi (changes requested):
Thanks! With this fresh in mind, could you also do a critical review of the [Browser Support Table](https://github.com/highcharts/highcharts/blob/bugfix/13238-panning-ordinal-lin2val-refactoring/docs/export-module/client-side-export.md/#browser-support-table)?

khlieng:
I think the only thing to change there might be that `Safari (latest)` should probably be just `OK` since Safari has supported the `download` attribute since 2017, iOS has also supported it since 2019.

TorsteinHonsi:
@khlieng What about Edge? I assume it is similar to Chrome now? But it should be tested in case or browser/feature sniffing plays tricks with it.

khlieng:
Yes probably, but the support table matches what we currently do in any version of Edge at least.

TorsteinHonsi:
Edge is an evergreen, so the old (pre-Webkit) versions are not so relevant anymore. To avoid confusion, we can write Edge >= 79.

TorsteinHonsi:
I've fixed an iOS regression (Safari v14)  and updated the support table.

But there's now a regression in Safari when exporting SVG with [embedded images from a CORS enabled server](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/exporting/offline-download-embedded/).

This apparently trickles down to Safari (iOS) and Chrome (iOS) which also runs Safari internally. 

I a sense it is out of scope for this PR, but on the other hand if the fix is simple we might as well do it here so that the whole support table is cleaned up. When this is fixed, offline-export supports all modern browsers, except perhaps external images for PDF export. Though that may be fixable too. Will continue looking into this tomorrow.

khlieng:
Dont we need to add an exception for Edge to make PDF and PNG/JPEG w/ images actually work?:

https://github.com/highcharts/highcharts/blob/cc36c66d8429de847513dfa8af3122c500a48201/ts/Extensions/OfflineExporting.ts#L852

TorsteinHonsi:
I'll test that. Anyway, I'm merging this now, and start working on a new PR with the Safari problem.

TorsteinHonsi (approved):

PR #15791, opened 2021-05-26
bugfix/15790-axis-offset-type

### LINKED ISSUE #15790: Typing regression: YAxisOptions.offset is typed as string instead of number
#### Expected behaviour
`YAxisOptions.offset` should be typed (and documented) as number, as it was before. 

#### Actual behaviour
It's now typed and documented as string, altough the text of the documentation and the examples show that it should be a number

#### Product version

9.1.0

khlieng:
Fixed #15790, axis `offset` doclet type was wrong.


TorsteinHonsi (approved):
Thanks!

PR #15794, opened 2021-05-26
bugfix/9612-zoom-pan-pointPlacement

### LINKED ISSUE #9612: Panning with chart.plotOptions causes zooming issue.
#### Expected behaviour
Configure highcharts to zoom and pan:
`chart: {
  zoomtype: 'x',
  panning: true,
  panKey: 'ctrl'
},
plotOptions: {
  series: {
     pointPlacement: 'on'
   }
}`
When holding the ctrl button chart should pan smoothly and remain at the same zoom level.

#### Actual behaviour
When holding ctrl and dragging the mouse, the chart pans but also zooms, slowly if zoom is low, but gets exponentially more pronounced as the zoom increases.

The plotOptions property seems to be the culprit upon first inspection. Removing the option produces expected behaviour.

Configure highcharts with the above options.
I'm unsure if any other options affect the issue.

Here is partial a reproduction of my chart.
http://jsfiddle.net/wmqfryh8/3/

#### Product version

6.2.0
#### Affected browser(s)
Latest Chrome and Firefox, maybe others I didn't test.

khlieng:
Fixed #9612, panning after zooming sometimes broke with `pointPlacement` set to `on` or `between`.

### DIFF
--- a/samples/unit-tests/axis/pointplacement/demo.js
+++ b/samples/unit-tests/axis/pointplacement/demo.js
@@ -1,26 +1,54 @@
 QUnit.test('Axis pointPlacement', assert => {
     var chart = Highcharts.chart('container', {
         chart: {
-            width: 600
+            width: 600,
+            zoomType: 'x',
+            panning: true,
+            panKey: 'shift'
         },
         xAxis: {
             categories: ['Apples', 'Pears', 'Bananas', 'Oranges']
         },
         series: [
             {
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks! Can you add a unit test as well?

TorsteinHonsi (changes requested):
Request changes

TorsteinHonsi (approved):

PR #15801, opened 2021-05-27
Chart update and events

### LINKED ISSUE #6538: Chart update didn't update chart.events functions
#### Expected behaviour
Using chart.update() method, I think that chart events methods should be updated.

#### Actual behaviour
WIth chart.update() chart events aren't updated

http://jsfiddle.net/ahv1sLxu/1/

To reproduce click on the button and make drilldown. Drilldown event function is not working.

#### Affected browser(s)
All

TorsteinHonsi:
Fixed #6538, `chart.update` didn't update events given in `options.chart.events`


pawelfus (approved):
Looks good 👍

PR #15802, opened 2021-05-27
Bugfix: Safari 14.1 time calculations and formatting

bre1470:
Fixed handling of local time with Safari 14.1.

### DIFF
--- a/ts/Core/Time.ts
+++ b/ts/Core/Time.ts
@@ -52,72 +52,22 @@ declare module './Axis/TickPositionsArray'{
     }
 }
 
-/**
- * Normalized interval.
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (commented):
Those two changes, are they both related to the same bug in Safari < 14.1? If so, can we somehow isolate the feature sniffing so that we can easily identify and remove it some time in the future when we don't have to support pre-14.1 Safari? In pseudo code:

`const hasOldSafariBug = H.isSafari && !Intl.DateTimeFormat.prototype.formatRange`

TorsteinHonsi (approved):

PR #15806, opened 2021-05-28
Migration: Responsive

bre1470:
- Improved MSPointer class.
- Migrated responsive to a composition.
- Removed responsive interfaces from internal namespace.
- Improved const/let pattern in responsive composition.


TorsteinHonsi (approved):
Thanks!

PR #15808, opened 2021-05-28
bugfix/15803-navigator-left

### LINKED ISSUE #15803: Setting navigator.xAxis.left causes navigator's selected area to be incorrectly offset
#### Expected behaviour
The navigator should work properly while honoring the defined offset in `navigator.xAxis.left`

#### Actual behaviour
The entire navigator does offset the desired amount, but the navigator's selected area does not honor the defined offset after being dragged.

nav

https://jsfiddle.net/8dmjv6fp/

#### Product version
Highstock latest (v9.1.0)

#### Affected browser(s)
All

khlieng:
Fixed #15803, navigator positioned using `navigator.xAxis.left` flickered when dragging.


TorsteinHonsi (approved):
Thanks!

PR #15809, opened 2021-05-28
Demo/website/homepage

nasin:
Hi, I modified the homepage animation to work with reduced motion preferences for accessibility (you have to turn the setting on to see it, obviously.) Not sure if I reduced the motion enough.


TorsteinHonsi (approved):
Thanks!

PR #15810, opened 2021-05-30
Removed duplicate entry in docs sidebar

goransle:
`stock/custom-technical-indicators` appears twice in the the stock category.


TorsteinHonsi (approved):
Thanks!

PR #15812, opened 2021-05-31
Studies: remove php files in studies folder

goransle:
These files are now served via S3, so a) PHP does not work, and b) there may be trouble with Cloudflare if we leave them accessible.


TorsteinHonsi (changes requested):
As far as I can see there are 5 .php files in the `/studies` folder. All of these come from early studies for features that are now official. So I think rather than filter them from the upload, we should delete them, and the studies where they are used, from the repo. We also need to check (search) whether these studies are referred from the docs.

TorsteinHonsi (approved):

PR #15814, opened 2021-05-31
Migration: BrokenAxis

bre1470:
- Improved BrokenAxis composition.
- Removed internal namespace reference in BrokenAxis.
- Improved const/let pattern.


TorsteinHonsi (approved):
Thanks!

Issue #15734, opened 2021-05-17
Stock Tools: Slow Stochastic is missing "period" fields in the popup


pawelfus:
#### Expected behaviour
"Periods" label should be displayed above the period inputs

#### Actual behaviour
"Periods" label  is missing:

![Zrzut ekranu 2021-05-17 o 17 55 16](https://user-images.githubusercontent.com/1453926/118519066-0fbd1180-b739-11eb-9d8f-c6c2dd71589c.png)

#### Live demo with steps to reproduce
[Demo](https://www.highcharts.com/demo/stock/stock-tools-gui)

#### Product
Highcharts Stock

Issue #15732, opened 2021-05-17
Stock Tools: Chaikin oscillator has unclear "period" fields


pawelfus:
#### Expected behaviour
Periods in Chaikin oscillator should distinguish between `period` for A/D and `periods` for EMA's.

#### Actual behaviour
All fields are labeled as "Period".

#### Live demo with steps to reproduce
[Demo](https://www.highcharts.com/demo/stock/stock-tools-gui)

![Zrzut ekranu 2021-05-17 o 17 39 06](https://user-images.githubusercontent.com/1453926/118516658-d08dc100-b736-11eb-9ba3-9f05fa66bc7f.png)

#### Product
Highcharts Stock

PR #15815, opened 2021-05-31
bugfix/15733-missing-label-stock-tools

### LINKED ISSUE #15733: Stock Tools: APO is missing "period" fields in the popup
#### Expected behaviour
"Periods" label should be displayed above the period inputs

#### Actual behaviour
Only "Index" label is displayed, "Periods" is missing:

Zrzut ekranu 2021-05-17 o 17 41 57

[Demo](https://www.highcharts.com/demo/stock/stock-tools-gui)

#### Product
Highcharts Stock

DJTechnoo:
Fixed #15733, missing 'Periods' label in Stock Tools pop-up for some indicators.


ts/Extensions/Annotations/Popup.ts
```diff
@@ -295,6 +295,23 @@ H.Popup.prototype = {
             lang = this.lang,
             inputName = PREFIX + type + '-' + optionName;
 
+        if (optionParamList.length === 3 && optionName === '0') { // If a label refers to multiple fields (15733)
```
sebastianbochan, line 298: Why we use `optionParamList.length === 3` ?

DJTechnoo: Because in the case of 'Periods' (plural), which represents an array of numbers for the input fields, it appends an index. So the `optionsParamList` will look something like `['params', 'periods', '0']`. 
'Periods' seems to be the only case like this.

sebastianbochan (approved):

sebastianbochan:
@pawellysy any feedback here?

khlieng:
This also fixes #15734 and possibly #15732 right?

DJTechnoo:
> This also fixes #15734 and possibly #15732 right?

Yes, indeed it does fix both 😄

TorsteinHonsi (approved):
Thanks!

PR #15816, opened 2021-05-31
bugfix/15811-title-verticalAlign-middle

### LINKED ISSUE #15811: Highcharts | Title is aligned above instead of the actual center in donut (pie) chart
#### Expected behaviour

`title.verticalAlign: 'middle'` should vertically align the title to the middle of the chart

#### Actual behaviour

image
The title is positioned a few pixels above the actual middle of the chart.

This seems to be less visible when [the chart's size is larger](https://jsfiddle.net/AmitBecken/wjs0onuv/2/), but in my case (a smaller chart, only 80X80 pixels) it's easily noticeable.

You can also see this in [the following Stack Overflow answer](https://stackoverflow.com/a/15964921/2928233), its [attached example image](https://i.stack.imgur.com/5WQhS.png) and its [attached example JSFiddle](http://jsfiddle.net/pDdcd/) which no longer matches the result in the example image.

https://jsfiddle.net/AmitBecken/264apbL7/3/

#### Product version
Highcharts 9.1.0

#### Affected browser(s)
Chrome, Firefox

khlieng:
Fixed #15811, setting `title.verticalAlign` to `middle` did not vertically center the title completely.

### DIFF
--- a/ts/Core/Chart/Chart.ts
+++ b/ts/Core/Chart/Chart.ts
@@ -1266,7 +1266,8 @@ class Chart {
             const title = (this as any)[key],
                 titleOptions: Chart.DescriptionOptionsType = (this as any).options[key],
                 verticalAlign = titleOptions.verticalAlign || 'top',
-                offset = key === 'title' ? -3 :
+                offset = key === 'title' ?
+                    verticalAlign === 'top' ? -3 : 0 :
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (approved):
Thanks!

PR #15817, opened 2021-05-31
bugfix/6765-dataGrouping-order 

### LINKED ISSUE #6765: Highstock : stacked area chart does not render on regions of intersection among series
* Chart type - area
* Stacking - normal
* Highstock version used - 5.0.11

https://jsfiddle.net/Lx3jum2n/2/

In the fiddle above, the portion of the chart in which all 3 series intersect is not rendered

(Series 'Three' starts at the left edge of the un-rendered portion; series 'One' and 'Two' end at the right edge of the un-rendered portion)

When connectNulls is true, the region of intersection shows up distorted.

https://jsfiddle.net/6bk6pemq/1/

Similar issues open - https://github.com/highcharts/highcharts/issues/1991, https://github.com/highcharts/highcharts/issues/3327

karolkolodziej:
Fixed #6765, changed how and when the `dataGrouping` should be fired for multiple series when one of them doesn't require grouping.

Todo
- [x] Tests

### DIFF
--- a/ts/Core/Axis/Axis.ts
+++ b/ts/Core/Axis/Axis.ts
@@ -1836,7 +1836,7 @@ class Axis {
             // First process all series assigned to that axis.
             axis.series.forEach(function (series): void {
                 // Allows filtering out points outside the plot area.
-                series.forceCrop = series.forceCropping();
+                series.forceCrop = series.forceCropping && series.forceCropping();
 
                 series.processData(
                     axis.min !== (axis.old && axis.old.min) ||
--- a/ts/Core/Axis/Axis.ts
+++ b/ts/Core/Axis/Axis.ts
@@ -1833,12 +1833,24 @@ class Axis {
         // This is in turn needed in order to find tick positions in ordinal
         // axes.
         if (isXAxis && !secondPass) {
+            // First process all series assigned to that axis.
             axis.series.forEach(function (series): void {
+                const chart = series.chart,
+                    options = series.options,
+                    dataGroupingOptions = options.dataGrouping,
+                    groupingEnabled = series.allowDG !== false && dataGroupingOptions &&
+                        pick(dataGroupingOptions.enabled, chart.options.isStock);
+
+                // Allows filtering out points outside the plot area.
+                series.forceCrop = groupingEnabled; // #334
+
                 series.processData(
                     axis.min !== (axis.old && axis.old.min) ||
                     axis.max !== (axis.old && axis.old.max)
                 );
             });
+            // Then apply grouping if needed.
+            fireEvent(this, 'afterProcessData');
         }
 
         // set the translation factor used in translate function
--- a/ts/Core/Axis/Axis.ts
+++ b/ts/Core/Axis/Axis.ts
@@ -1844,7 +1844,7 @@ class Axis {
                 );
             });
             // Then apply grouping if needed.
-            fireEvent(this, 'afterProcessData');
+            fireEvent(this, 'postProcessData');
         }
 
         // set the translation factor used in translate function
--- a/ts/Core/Axis/Axis.ts
+++ b/ts/Core/Axis/Axis.ts
@@ -1835,14 +1835,8 @@ class Axis {
         if (isXAxis && !secondPass) {
             // First process all series assigned to that axis.
             axis.series.forEach(function (series): void {
-                const chart = series.chart,
-                    options = series.options,
-                    dataGroupingOptions = options.dataGrouping,
-                    groupingEnabled = series.allowDG !== false && dataGroupingOptions &&
-                        pick(dataGroupingOptions.enabled, chart.options.isStock);
-
                 // Allows filtering out points outside the plot area.
-                series.forceCrop = groupingEnabled; // #334
+                series.forceCrop = series.forceCropping();
 
                 series.processData(
                     axis.min !== (axis.old && axis.old.min) ||
--- a/ts/Extensions/DataGrouping.ts
+++ b/ts/Extensions/DataGrouping.ts
@@ -1024,13 +1024,13 @@ seriesProto.generatePoints = function (): void {
 
 // When all series are processed, calculate the group pixel width and then
 // if this value is different than zero apply groupings.
-addEvent(Axis, 'afterProcessData', function (): void {
+addEvent(Axis, 'postProcessData', function (): void {
     const axis = this,
         series = axis.series;
 
     series.forEach(function (series): void {
         // Reset the groupPixelWidth, then calculate if needed.
-        series.groupPixelWidth = null as any; // #2110
+        series.groupPixelWidth = void 0; // #2110
 
         series.groupPixelWidth = axis.getGroupPixelWidth && axis.getGroupPixelWidth();
 
--- a/ts/Extensions/DataGrouping.ts
+++ b/ts/Extensions/DataGrouping.ts
@@ -72,6 +72,7 @@ declare module '../Core/Series/SeriesLike' {
         hasGroupedData?: boolean;
         hasProcessed?: boolean;
         preventGraphAnimation?: boolean;
+        applyGrouping(): void;
         destroyGroupedData(): void;
         generatePoints(): void;
         getDGApproximation(): string;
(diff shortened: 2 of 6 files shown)


ts/Core/Axis/Axis.ts
```diff
@@ -1833,12 +1833,24 @@ class Axis {
...
+            // First process all series assigned to that axis.
             axis.series.forEach(function (series): void {
+                const chart = series.chart,
+                    options = series.options,
+                    dataGroupingOptions = options.dataGrouping,
+                    groupingEnabled = series.allowDG !== false && dataGroupingOptions &&
+                        pick(dataGroupingOptions.enabled, chart.options.isStock);
+
+                // Allows filtering out points outside the plot area.
+                series.forceCrop = groupingEnabled; // #334
```
pawelfus, line 1845: Do we need this in Axis core? Stock related code should be in the Stock modules

pawelfus (changes requested):

ts/Core/Axis/Axis.ts
karolkolodziej: Inside that loop the `series.processData()` is triggered. Then this method fires the `getProcessedData()`.
If I don't add this line (`series.forceCrop = groupingEnabled`) the method (`getProcessedData`) returns uncropped data that causes some further issues.
At that point, I should know if the data should be cropped or not.

I could an event like for example `beforeProcessData` because I guess we shouldn't play with `cropThreshold` here?

pawelfus: Perhaps we should replace the flag `forceCrop` with a method that returns this info? Something like this:

```js
forceCropping () {
  return this.forceCrop || (
    this.allowDG !== false &&
    dataGroupingOptions &&
    pick(dataGroupingOptions.enabled, this.chart.options.isStock);
  );
}
``` 

And use it in `getProcessedData()`? What do you think @TorsteinHonsi ?

Just to explain the solution we are trying to achieve here:
Currently we process and group data in this way (pseudo-code):
```js
axis.series.forEach(series => {
  series.processData();
  if (groupingNeeded) { series.groupData(); }
});
```

This way first series might not be grouped but the second one will be. That causes problems with stacking (different x-values for stacks).

Instead, we decided to solve it this way:
```js
axis.series.forEach(series => {
  series.processData();
});
// All series processed, we know groupPixelWidth for the most dense series
axis.series.forEach(series => {
  if (groupingNeeded) { series.groupData(); }
});
```

TorsteinHonsi: Maybe I am missing something here, but we already have logic that is supposed to go over all the series and identify a common `groupPixelWidth`:

https://github.com/highcharts/highcharts/blob/627e742c23721b32e2c744cee92a5a04659e41b5/ts/Extensions/DataGrouping.ts#L1163-L1165

karolkolodziej: We do have the `getGroupPixelWidth` method, but in order to get the `groupPixelWidth` the data needs to be processed. And in this case, we have to flag that **before** processing.

``` TS
// If one of the series needs grouping, apply it to all (#1634)
    i = len;
    while (i--) {
        dgOptions = series[i].options.dataGrouping;

        if (dgOptions) { // #2692

            dataLength = (series[i].processedXData || series[i].data).length;

            // Execute grouping if the amount of points is greater than the
            // limit defined in groupPixelWidth
            if (
                series[i].groupPixelWidth ||
                dataLength >
                ((this.chart.plotSizeX as any) / groupPixelWidth) ||
                (dataLength && dgOptions.forced)
            ) {
                doGrouping = true;
            }
        }
    }
```

The `forceCrop` flag rather checks if the grouping might be needed in the future. 

If you take a look at the `getProcessedData` the part that is failing to crop is the one below. I don't think that we can add any other property to make it pass.

``` TS
        // optionally filter out points outside the plot area
        if (isCartesian &&
            series.sorted &&
            !getExtremesFromAll &&
            (
                !cropThreshold ||
                dataLength > cropThreshold ||
                series.forceCrop
            )
        ) {
```

I will suggest adding a function like Pawel suggested (`forceCropping`)

TorsteinHonsi:
Pending new `forceCropping` function

ts/Core/Axis/Axis.ts
```diff
@@ -1833,12 +1833,18 @@ class Axis {
...
+                // Allows filtering out points outside the plot area.
+                series.forceCrop = series.forceCropping();
+
                 series.processData(
                     axis.min !== (axis.old && axis.old.min) ||
                     axis.max !== (axis.old && axis.old.max)
                 );
             });
+            // Then apply grouping if needed.
+            fireEvent(this, 'afterProcessData');
```
TorsteinHonsi, line 1847: The convention in Highcharts is that events with the name `after...` are fired at the end of the function they're named after. So in this case, it either has to have a different name, or the call has to be moved to the end of the `processData` function. 

I would prefer the latter. If the current event handler should only be called in special cases,  we can pass over event arguments to help decide. A potential problem is that StockChart overrides `Series.prototype.processData` so that there is code executing after the parent function. If that is a problem, we can move the whole StockChart processData logic, combined with the new logic, inside an `afterProcessData` event handler.

ts/Extensions/DataGrouping.ts
```diff
@@ -1012,6 +1022,26 @@ seriesProto.generatePoints = function (): void {
...
+// if this value is different than zero apply groupings.
+addEvent(Axis, 'afterProcessData', function (): void {
+    const axis = this,
+        series = axis.series;
+
+    series.forEach(function (series): void {
+        // Reset the groupPixelWidth, then calculate if needed.
+        series.groupPixelWidth = null as any; // #2110
+
+        series.groupPixelWidth = axis.getGroupPixelWidth && axis.getGroupPixelWidth();
```
TorsteinHonsi, line 1035: Simplify the assignment and get rid of the any cast (if we're not using `null` as a special meaning, in which case we need to add typing for it).

`series.groupPixelWidth = (axis.getGroupPixelWidth && axis.getGroupPixelWidth()) || void 0;`

TorsteinHonsi (changes requested):
Thanks! In addition to the inline comments:
* Please make sure the PR description is understandable in the context of the changelog. Right now it doesn't mention data grouping, only grouping which is ambiguous. Should also probably better explain when the problem arose.
* Please check out the visual diffs. Why is data grouping not applied to the scrolbar demo?

ts/Extensions/DataGrouping.ts
karolkolodziej: I agree that the `null as any` doesn't look nice.. 
I will change it to `void 0` however I can't simplify that assignment because before using the `getGroupPixelWidth` method the `groupPixelWidth` should be reset.

TorsteinHonsi (approved):

TorsteinHonsi:
Pending approval by @pawelfus and @pawellysy

ts/Core/Axis/Axis.ts
```diff
@@ -1833,12 +1833,18 @@ class Axis {
         // This is in turn needed in order to find tick positions in ordinal
         // axes.
         if (isXAxis && !secondPass) {
+            // First process all series assigned to that axis.
             axis.series.forEach(function (series): void {
+                // Allows filtering out points outside the plot area.
+                series.forceCrop = series.forceCropping();
```
pawelfus, line 1839: `forceCropping` is defined only in Highstock, so this will break all Highcharts charts: https://jsfiddle.net/BlackLabel/3p7ftcxh/

Replace with simple `series.forceCrop = series.forceCropping && series.forceCropping();`

PS: Why didn't TS warn about the issue?

ts/Extensions/DataGrouping.ts
```diff
@@ -1012,6 +1022,26 @@ seriesProto.generatePoints = function (): void {
...
 
+// When all series are processed, calculate the group pixel width and then
+// if this value is different than zero apply groupings.
+addEvent(Axis, 'postProcessData', function (): void {
+    const axis = this,
+        series = axis.series;
+
+    series.forEach(function (series): void {
+        // Reset the groupPixelWidth, then calculate if needed.
+        series.groupPixelWidth = void 0; // #2110
```
pawelfus, line 1033: Q: I know it's not directly from this PR, but I'm curious if this is enough? Clearing only **current** series `groupPixelWidth`. That means all others still keep old `groupPixelWidth`, isn't it?

pawelfus (changes requested):
See comments below:

ts/Core/Axis/Axis.ts
TorsteinHonsi: 😮  This is the danger of always loading all the modules in the karma tests...

ts/Extensions/DataGrouping.ts
karolkolodziej: Hmm... that is an interesting question. 
Theoretically, it seems as that might not be enough however I haven't found any failing scenario for that...

After some thinking, I assume that it should be ok because in the next step the `getGroupPixelWidth` method already knows about all series and if needed this whole loop should set the same `groupPixelWidth` for each series. In this case, it doesn't matter that some series might have old `groupPixelWidth`.

pawelfus (approved):
Great, thanks!

pawellysy (approved):
Nothing to Add! 👍

PR #15818, opened 2021-05-31
Migration: Chart3D

bre1470:
- Moved compositions of Chart3D into master.
- Removed Chart3D types from internal namespace.
- Replaced SVGAttributesExtendeded with SVGAttributes3D.


TorsteinHonsi (approved):
Thanks!

PR #15829, opened 2021-06-02
Enhancement/patterns and demo

oysteinmoseng:
Made default pattern-fills more subtle and usable. Changed `accessible-pie` demo for better pattern handling.


TorsteinHonsi (changes requested):
Thanks! As for CSS style guide, we are (almost) consistent throughout our code base at using  kabab-case class names and id's. So I think it should be `#patterns-enabled`. And also `#peLbl` is hardly human readable, I think it looks friendlier to spell out `#patterns-enabled-label`.

TorsteinHonsi (approved):

PR #15833, opened 2021-06-02
bugfix/2406-val2lin-refactor

### LINKED ISSUE #2406: Line/point calculations in Highstock
Example: http://jsfiddle.net/nywsa/8/

Zoom in on the chart. For example to see only last week. Lines are crossing in the middle, while shouldn't, values should go to:
- black: 190-206, while it's a full line (34-206)
- blue: 150-160, while it's 64-160

Related topic: http://forum.highcharts.com/highstock-usage/line-series-on-candlestick-chart-render-incorrectly-on-zoom-t27890/

This issue happens only when `ordinal=true` is set

pawellysy:
Fixed #2406, fixed `val2lin` method to correctly calculate positions outside the visible `plotArea` in ordinal axis.

### DIFF
--- a/ts/Core/Axis/OrdinalAxis.ts
+++ b/ts/Core/Axis/OrdinalAxis.ts
@@ -458,6 +458,7 @@ namespace OrdinalAxis {
                         getGroupIntervalFactor: this.getGroupIntervalFactor
                     },
                     ordinal2lin: axisProto.ordinal2lin, // #6276
+                    getIndexOfPoint: axisProto.getIndexOfPoint,
                     val2lin: axisProto.val2lin // #2590
                 } as any;
                 fakeAxis.ordinal.axis = fakeAxis;
@@ -1049,7 +1050,7 @@ namespace OrdinalAxis {
          * Translate from a linear axis value to the corresponding ordinal axis
          * position. If there are no gaps in the ordinal axis this will be the
          * same. The translated value is the value that the point would have if
-         * the axis were linear, using the same min and max.
+         * the axis was linear, using the same min and max.
          *
          * @private
          * @function Highcharts.Axis#val2lin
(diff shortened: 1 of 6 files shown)


ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -1146,17 +1327,22 @@ namespace OrdinalAxis {
                     // range, else it happens on the current x axis which is
                     // smaller and faster.
                     chart.fixedRange = max - min;
+
+                    // @todo debug session, to change later
```
TorsteinHonsi, line 1331: > to change later

When? We should remove debug stuff before merging it into the master

TorsteinHonsi (changes requested):
Thanks! In addition to the inline comment, please adapt the PR description according to [writing content for the changelog](https://github.com/highcharts/highcharts/tree/master/changelog#writing-content-for-the-changelog). Points 5 and 8.

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -955,53 +1063,126 @@ namespace OrdinalAxis {
...
+
+                if (!slope) {
+                    slope =
+                        (extendedOrdinalPositions[EOPlength - 1] -
+                            extendedOrdinalPositions[0]) /
+                        EOPlength;
+                }
+                // this value represents the value, where the first index of
+                // ordinalPositions lays in extendedOrdinalPositions.
+                // in other words, how much the OP is moved inside the EOP
```
karolkolodziej, line 1125: I would unify the names here (`EOP`, `ExtendedOrdinalPositions`, `extendedOrdinalPositions`)
IMO the `rxtendedOrdinalPositions` is the best.

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -955,53 +1063,126 @@ namespace OrdinalAxis {
...
-                // two nearest values
-                i = ordinalLength - 1;
-                while (i--) {
-                    if (val > ordinalPositions[i] || i === 0) { // interpolate
-                        // something between 0 and 1
-                        distance = (val - ordinalPositions[i]) /
-                            (ordinalPositions[i + 1] - ordinalPositions[i]);
-                        ordinalIndex = i + distance;
-                        break;
+                let EOPlength = extendedOrdinalPositions.length,
```
karolkolodziej, line 1114: Could it be `const`?

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -955,53 +1063,126 @@ namespace OrdinalAxis {
...
+                // If the searched value is outside the visiblePlotArea,
+                // check if it is inside ExtendedOrdinalPositions.
+                if (
+                    val >= extendedOrdinalPositions[0] &&
+                    val <=
+                        extendedOrdinalPositions[
+                            extendedOrdinalPositions.length - 1
+                        ]
+                ) {
+                    let index = Composition.findIndexOf(
```
karolkolodziej, line 1140: Could it be `const`?

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -955,53 +1063,126 @@ namespace OrdinalAxis {
...
+                        extendedOrdinalPositions,
+                        val,
+                        true
+                    );
+
+                    if (extendedOrdinalPositions[index] === val) {
```
karolkolodziej, line 1146: As it was in the `bugfix/13238` branch, I think here we could apply the same.
Replace the `extendedOrdinalPositions` and `ordinalPositions` with a clever switch because the algorithm seems to be the same in both cases.  The array is the only thing that changes.

ts/Core/Axis/OrdinalAxis.ts
karolkolodziej, line 1331: This one should go back to the previous state.

karolkolodziej (changes requested):
Great work!
Few changes in the code and some visual differences needs to be checked.

ts/Core/Axis/OrdinalAxis.ts
pawellysy: My bad, forgot to delete it before pushing.

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -1062,53 +1063,110 @@ namespace OrdinalAxis {
...
-                // two nearest values
-                i = ordinalLength - 1;
-                while (i--) {
-                    if (val > ordinalPositions[i] || i === 0) { // interpolate
-                        // something between 0 and 1
-                        distance = (val - ordinalPositions[i]) /
-                            (ordinalPositions[i + 1] - ordinalPositions[i]);
-                        ordinalIndex = i + distance;
-                        break;
+                const EOPlength = extendedOrdinalPositions.length;
```
TorsteinHonsi, line 1116: Please use some other name to adhere to the style of using lower case first on JS const and let's. Suggest `positionsLength` or just `length`.

TorsteinHonsi (changes requested):

TorsteinHonsi (approved):
All good from my end 👍

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -1062,53 +1082,95 @@ namespace OrdinalAxis {
...
+                    ordinalPositions[0]
+                );
+
+                // If the searched value is outside the visiblePlotArea,
+                // check if it is inside extendedOrdinalPositions.
+                if (
+                    val >= extendedOrdinalPositions[0] &&
+                    val <=
+                        extendedOrdinalPositions[
+                            extendedOrdinalPositions.length - 1
```
pawelfus, line 1141: You already have `length` variable

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -1062,53 +1082,95 @@ namespace OrdinalAxis {
...
+                        let diff = extendedOrdinalPositions[0] - val,
+                            approximateIndexOffset = diff / slope;
+                        ordinalIndex =
+                            -originalPositionsReference -
+                            approximateIndexOffset;
+                    } else {
+                        let diff =
+                                val -
+                                extendedOrdinalPositions[
+                                    extendedOrdinalPositions.length - 1
```
pawelfus, line 1161: `length` variable is already defined

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -1062,53 +1082,95 @@ namespace OrdinalAxis {
...
                 }
-                ret = toIndex ?
-                    ordinalIndex :
-                    (ordinal.slope as any) *
-                    (ordinalIndex || 0) +
-                    (ordinal.offset as any);
             }
-            return ret;
+
+            return toIndex ? ordinalIndex : (slope as any) * (ordinalIndex || 0) +
```
pawelfus, line 1172: Is this `slope as any` necessary? I think you already check if `slope` is not defined

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -1062,53 +1082,95 @@ namespace OrdinalAxis {
...
+                    val >= extendedOrdinalPositions[0] &&
+                    val <=
+                        extendedOrdinalPositions[
+                            extendedOrdinalPositions.length - 1
+                        ]
+                ) {
+                    // Return Value
+                    ordinalIndex = getIndexInArray(extendedOrdinalPositions, val) - originalPositionsReference;
+                } else {
+                    // since ordinal.slope is the average distance between 2
```
pawelfus, line 1147: In general, I would skip this, but since there are some other changes, could you start comment from the capital letter? ;)

pawelfus (changes requested):
Some nitpicking, overall great!

ts/Core/Axis/OrdinalAxis.ts
pawellysy: typescript still says, it can be undefined ¯\_(ツ)_/¯

pawelfus (approved):
Great, thanks!

karolkolodziej (approved):

raf18seb (approved):

PR #15834, opened 2021-06-02
Better legend a11y by placing elements in a list.

oysteinmoseng:
Improved legend accessibility by placing legend buttons in a list.


TorsteinHonsi (approved):
Thanks!

PR #15836, opened 2021-06-03
New feature: relativeXValue

### LINKED ISSUE #1602: Data Compression Support - Base Date to Shorten Timestamps
Highcharts can "compress" very well the data by using "regular intervals" for time series (since only the start timestamp and the interval is required).

The problem is with irregular intervals :(.

Please add support for "Base Date to Shorten Timestamps".
See an example how AnyChart is doing it:
http://www.anychart.com/products/stock/docs/users-guide/index.html?data-set-input-data-settings.html#timestamp-units-cuttting

It's very efficient, and it cuts for many of our charts the transferred data to ~70%.

Thank you very much.

TorsteinHonsi:
Added new feature, `series.relativeXValue`, allowing point X values to be relative to the `series.pointStart` and subject to `series.pointInterval` and `series.pointIntervalUnit`.

Closes #1602

## To do
 - [x] Consider the naming. The name `pointStartAsBase` is nice as it refers directly to `pointStart`, but it fails to communicate that `pointInterval` and `pointIntervalUnit` also affect the X value. Maybe a better name would be simply `relativeX` or `relativeXValues`. Decision: Rename to `relativeXValue` (singular).
 - [x] API docs. Point to the two existing samples (Highcharts and Stock)
 - [x] Unit tests. Check crossing of DST specifically. Check dynamically updating data and settings. Issue with `setData` [not recognizing modified X values](https://jsfiddle.net/highcharts/pa7zxy9h/1/).
 - [x] General docs.

## Demo
https://jsfiddle.net/highcharts/pa7zxy9h/


bre1470 (approved):
Nice! 👍

PR #15837, opened 2021-06-03
bugfix/15777-plotLine-label-clip

### LINKED ISSUE #15777: Plot line label extends past the plot area into chart area
#### Expected behaviour
Since the plot line is cut off at the plot area I would expect that the label text should also get cut off in the plot area

#### Actual behaviour
image

[https://jsfiddle.net/3uy672o4/](https://jsfiddle.net/3uy672o4/)

#### Product version
Highcharts JS v9.1.0

#### Affected browser(s)
Tested Firefox/Chrome. Probably all browsers.

khlieng:
Fixed #15777, plot line labels were not clipped.


TorsteinHonsi (changes requested):
I think in this case it would be better to go for text overflow handling instead of clipping. It has some advantages:
* We get optionally either an ellipsis or line-wrap instead of a clean cut.
* If ellipsis, we get a `title` built-in tooltip to read the whole text
* It is more compact

See https://jsfiddle.net/highcharts/vqdhyc8w/. Instead of the fixed width I am using in the demo, we would internally insert the length of the plot line as default `style.width`. Maybe we should also change the default `textOverflow`. I am not sure what is best, line-wrap or ellipsis.

ts/Core/Axis/PlotLineOrBand.ts
```diff
@@ -452,35 +440,61 @@ class PlotLineOrBand {
...
+            width = axis.height - (label.alignAttr.y - axis.top);
+        } else if (rotation === 180) {
+            width = label.alignAttr.x - axis.left;
+        } else if (rotation === 270) {
+            width = label.alignAttr.y - axis.top;
+        }
+
+        if (width) {
+            label.css({ width: width + 'px' });
+        }
```
TorsteinHonsi, line 496: Since we're already computing the box height and width further up, I think this can be simplified. The following works for me, but it doesn't take into account the rotation of 180 and 270 - but I am not able to make those work in the first place...

```js
        const x = arrayMin(xBounds);
        const y = arrayMin(yBounds);
        const width = arrayMax(xBounds) - x;
        const height = arrayMax(yBounds) - y;

        label.align(optionsLabel, false, {
            x: x,
            y: y,
            width,
            height
        });

        label.css({
            width: (
                axis.horiz ?
                    height - (optionsLabel.y || 0) :
                    width - (optionsLabel.x || 0)
            ) + 'px'
        });
        ```

khlieng: Hmm, heres 180: https://jsfiddle.net/epLx81bq/2/

We should possibly handle right-aligned labels as well?

TorsteinHonsi: > Hmm, heres 180

I'm not sure it's useful though. It is also possible to set any rotation: https://jsfiddle.net/highcharts/krogm2u4/ . To keep it simple, maybe we should just stick with rotations of 0 and 90?

> We should possibly handle right-aligned labels as well?

Yes, possibly. But I'm afraid of bloat. Just wondering if there is any synergy with other similar concepts here. Chart title and subtitle perhaps.

TorsteinHonsi (approved):

PR #15850, opened 2021-06-07
bugfix/15845-selection-marker-export

### LINKED ISSUE #15845: Remains zoom area after using download with exporting.
#### Expected behaviour
The zoom area should disappear after downloading the chart image by exporting.

#### Actual behaviour
The zoom area remains after download the chart image.

https://jsfiddle.net/BlackLabel/rn35bud0/

#### Product version
Highcharts JS v9.1.1 (2021-06-03)

#### Affected browser(s)
All

khlieng:
Fixed #15845, selection marker showed without mousedown after export.

### DIFF
--- a/ts/Core/Globals.ts
+++ b/ts/Core/Globals.ts
@@ -223,7 +223,7 @@ namespace Globals {
      *
      * */
 
-    export let chartCount: 0;
+    export let chartCount = 0;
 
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (approved):
Thanks!

PR #15856, opened 2021-06-08
Expose Point.doDrilldown

TorsteinHonsi:
Exposed `Point.doDrilldown` and `Axis.drilldownCategory` as official API methods.

### DIFF
--- a/samples/highcharts/drilldown/programmatic/demo.css
+++ b/samples/highcharts/drilldown/programmatic/demo.css
@@ -1,0 +1,11 @@
+#container {
+    min-width: 310px;
+    max-width: 800px;
+    height: 400px;
+    margin: 0 auto;
+}
+#button-row {
+    min-width: 310px;
+    max-width: 800px;
+    margin: 0 auto;
+}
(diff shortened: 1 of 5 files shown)


pawelfus (changes requested):
It would be nice to have a demo like other API functions have (`sth.update()`, `Axis.addPlotLine()` etc)

Extra question:
I don't see support for multi-drilldown, for example drilling down two points at the same time (like drilldown to a category). I would expect something like this:
```js
series_1.points[0].doDrilldown(false);
series_2.points[0].doDrilldown(true); // redraw=true ?
```

TorsteinHonsi:
Thanks for the review! Will add a sample. 

> I don't see support for multi-drilldown, for example drilling down two points at the same time (like drilldown to a category)

The `redraw` parameter of `doDrilldown` doesn't quite work that way. Instead, there's a separate `Axis.prototype.drilldownCategory` function that does what you say. I will expose that too, and tie them together with `@see` tags.

pawelfus (approved):
Thanks!

PR #15857, opened 2021-06-08
Fixed #6173, CSP errors on tooltip drop shadow in styled mode

### LINKED ISSUE #6173: HC5 - Inline styles
#### Expected behaviour
No inline styles

#### Actual behaviour
Inline styles are still applied.

We still have inline styles: 
- http://jsfiddle.net/yLu0zreg/ (check a title <text>)

#### Affected browser(s)
All browsers

zrzut ekranu 2016-12-23 o 11 02 08

EDIT:
- errors when content-security-policy header is applied

TorsteinHonsi:
Fixed #6173, CSP errors on tooltip drop shadow, even in styled mode.

Set tooltip shadow filter as attribute instead of inline style.

### DIFF
--- a/ts/Core/Renderer/SVG/SVGAttributes.d.ts
+++ b/ts/Core/Renderer/SVG/SVGAttributes.d.ts
@@ -57,6 +57,7 @@ export interface SVGAttributes {
     end?: number;
     fill?: ColorType;
     'fill-opacity'?: number;
+    filter?: string;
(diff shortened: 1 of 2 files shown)


TorsteinHonsi:
Pending @KacperMadej 's approval

KacperMadej (approved):
Looks good. Unable to find inline styles 👍

PR #15858, opened 2021-06-08
bugfix/15843-tooltip-shared-stickOnContact

### LINKED ISSUE #15843: Shared tooltip with stickOnContact doesn't always update
#### Expected behaviour
In displaying a shared tooltip with stickOnContact, specifically for a stacked column chart, the tooltip should display and update appropriately as the cursor hovers over different columns in the chart.

#### Actual behaviour
The tooltip displays but sometimes does not update when you move the cursor to other columns. If you move the cursor beyond the approximate width of the tooltip, it finally triggers an update.

[jsFiddle](https://jsfiddle.net/7f1scvpm/1/)

To recreate: mouseover Aug Series 1 around where Series 1 and 2 meet. Tooltip correctly appears for Aug. Now move the cursor horizontally left through July and June-- the tooltip doesn't update. Keep going to May (staying on Series 1) and it updates. It also updates if the cursor touches Series 2 in any column. 

#### Product version
Highcharts 9.1.1

#### Affected browser(s)
Chrome for sure. Haven't tested in others

**A few notes and theories:**

- I can only get it to happen with combo of stickOnContact, and shared tooltip. (Haven't tried on other chart types besides column)
- It seems like the area for stickOnContact is somehow being miscalculated. If I move my mouse horizontally left/right or downwards beyond the approximate height and width of the tooltip, it updates
- Moving the cursor upwards towards the tooltip keeps it sticky, even when I'm not hovered over the tooltip itself
- It only seems to be a problem when the tooltip was triggered with a mouseover towards the bottom of the chart, not when it's higher up on the chart
- Doing the same actions above but going horizontally right doesn't recreate the problem. Probably due to miscalculation of the bounds of the tooltip and the fact that it's displayed to the left of the hovered column

khlieng:
Fixed #15843, shared tooltip with `stickOnContact` enabled did not always update position.

### DIFF
--- a/samples/unit-tests/tooltip/shared/demo.js
+++ b/samples/unit-tests/tooltip/shared/demo.js
@@ -212,14 +212,15 @@ QUnit.test(
     }
 );
 
-QUnit.test('Shared tooltip with pointPlacement (#5832)', function (assert) {
+QUnit.test('Shared tooltip with pointPlacement and stickOnContact', function (assert) {
     var chart = Highcharts.chart('container', {
         chart: {
             type: 'column'
         },
 
         tooltip: {
-            shared: true
+            shared: true,
+            stickOnContact: true
         },
         plotOptions: {
             column: {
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

Issue #15827, opened 2021-06-01
Stock Annotations: Freeform drawing


pawelfus:
# Description of the feature
Create an annotation type of a freeform drawing. PoC needed before implementing this. Probably there will be no control points, only drag&drop to move the annotation around.

# Library related to the feature
Highcharts Stock

# Proof of Concept/Live example for the feature
PoC needed

----

**You can vote for this feature by adding a thumbs-up reaction to this post.**

pawelfus:
Unstale

Issue #15826, opened 2021-06-01
Stock Annotations: Time cycles


pawelfus:
# Description of the feature
Add a Time cycles annotation. Indicates regular intervals on the timeline.

- Two control points:
  - one to control x position
  - second one to control the diameter of the first cycle
- Add this annotation type to Stock Tools
- Separate demo
- API options similar to Circle/Ellipse (fill, stroke, stroke-width)

# Library related to the feature
Highcharts Stock

# Proof of Concept/Live example for the feature
![Zrzut ekranu 2021-06-1 o 15 46 16](https://user-images.githubusercontent.com/1453926/120333987-86890b80-c2f0-11eb-9b95-bf6d137e4a1f.png)

----

**You can vote for this feature by adding a thumbs-up reaction to this post.**

Issue #15825, opened 2021-06-01
Stock Annotations: Fibonacci Timezones


pawelfus:
# Description of the feature
Add a Fibonacci timezones annotation. Indicates Fibonacci intervals on the timeline.

- Each vertical line should have a label with a value
- Two control points:
  - one to control where the first line is rendered
  - second one to control the distance between the 1st and the 2nd (4th?) vertical line
- API options similar to Fibonacci annotations (probably implementation can be based on that Class too)

# Library related to the feature
Highcharts Stock

# Proof of Concept/Live example for the feature
![Zrzut ekranu 2021-06-1 o 15 34 03](https://user-images.githubusercontent.com/1453926/120332782-5c831980-c2ef-11eb-9bfa-8a1fee1fd9a1.png)

# Todo's:
- [x] - create new annotation type (see [other types](https://github.com/highcharts/highcharts/tree/master/ts/Extensions/Annotations/Types))
- [x] - add to `annotations-advanced` module
- [x] - add to [API](https://api.highcharts.com/highstock/annotations)
- [x] - create a demo (similar to [this one](https://jsfiddle.net/BlackLabel/fxp3L96s/))
- [x] - add to Stock Tools ([icon](https://github.com/highcharts/highcharts/blob/master/gfx/stock-icons/fibonacci-timezone.svg))
----

**You can vote for this feature by adding a thumbs-up reaction to this post.**

Issue #15348, opened 2021-03-17
Heikin Ashi


pawelfus:
# Description of the feature
In regular candlestick series, the body of the candle is rendered as OPEN+CLOSE values and wicks/shadows as HIGH+LOW values.

In Heikin Ashi, values are transformed according to the formula:
```
NEW_CLOSE = (OPEN+HIGH+LOW+CLOSE) / 4
NEW_OPEN = (PREVIOUS_OPEN + PREVIOUS_CLOSE) / 2 
NEW_HIGH = Math.max(HIGH, NEW_CLOSE, NEW_OPEN)
NEW_LOW = Math.min(LOW, NEW_CLOSE, NEW_OPEN)
```
The first candlestick is calculated as:
```
NEW_CLOSE = (OPEN+HIGH+LOW+CLOSE) / 4
NEW_OPEN = (OPEN + CLOSE) / 2 
NEW_HIGH = HIGH
NEW_LOW = LOW
```
Then rendered as regular candlestick shape, using new values

# Library related to the feature
Highcharts Stock

# Proof of Concept/Live example for the feature
N/A

# To do's:
- [x] - new series/options implementation
- [x] - update [SCSS](https://github.com/highcharts/highcharts/blob/master/css/highcharts.scss) (in case of new styled-mode options) and the [list](https://github.com/highcharts/highcharts/blob/master/docs/chart-design-and-style/style-by-css.md) (if necessary)
- [x] - add to [Stock Tools'](https://www.highcharts.com/demo/stock/stock-tools-gui) button list of series types
- [x] - official [demo](https://www.highcharts.com/demo/stock) (this will cover visual test too)
- [x] - official [general documentation](https://www.highcharts.com/docs/stock/candlestick-chart)
- [x] - update [API](https://api.highcharts.com/highstock/) (series/options description)

----

**You can vote for this feature by adding a thumbs-up reaction to this post.**

PR #15868, opened 2021-06-08
Feature/stock annotation icons

nasin:
These are icons for the following new stock annotation/series:

- Heikin Ashi, touch #15348
- Hollow candlestick, touch #15347
- HLC, touch #15346
- Time cycles, touch #15826
- Fibonacci Timezones touch #15825
- Freeform drawing, touch #15827


TorsteinHonsi (changes requested):
Thanks! I added a change to make the existing icons render regardless of sizing.

In addition, it seems that there are some  sample file changes in this PR that are out of scope. They hasn't got do do with the annotation icons.

TorsteinHonsi (commented):
Regarding the icons themselves, it seems like at least some of the new ones have a thinner line style than the existing ones. Is that a problem?

Skjermbilde 2021-06-09 kl  12 38 28

pawelfus:
> Thanks! I added a change to make the existing icons render regardless of sizing.

Icons are a little too big right now 😅  (and separators too small) Compare:
- [current](https://jsfiddle.net/BlackLabel/sgbukwmj/4/)
- [candidate](https://jsfiddle.net/BlackLabel/sgbukwmj/5/) (you can observe new icons "in action")

|Zrzut ekranu 2021-06-9 o 14 27 23| Zrzut ekranu 2021-06-9 o 14 27 32 |

pawelfus (changes requested):
Could you add `series-` prefix to all series-related icons, so we match the current format? For example `series-hlc` instead of `hlc-series`

Thanks!

TorsteinHonsi:
> Icons are a little too big right now

Oh. Since I removed the width and height attributes, the background is now stretched to the size of the button. But I'm inclined to think that this is better behaviour (more responsive). Is there a way to make them render like they did, but using CSS? 

In addition, the new icons in this PR do not have width and height attributes, so in order to make it consistent, we need to either add attributes to those or remove them from the old ones.

pawelfus:
> Oh. Since I removed the width and height attributes, the background is now stretched to the size of the button. But I'm inclined to think that this is better behaviour (more responsive). Is there a way to make them render like they did, but using CSS?

@sebastianbochan - I need your assist, what do you think? Can we get the old behaviour in CSS?

nasin:
I created new icons for the follow: 

- Ellipse, touch #15008 
- Disable Zoom (panning.svg), no ticket yet, 
- linear.svg and logarithmic.svg for axes switching, #15695 
- Clear chart -- I suggest using the trash can icon (destroy.svg), no ticket yet. 

I also renamed series icons. And I updated the width of the lines on some of the icons (drawing, time-cycles, fibonacci timezones.)

TorsteinHonsi:
> @sebastianbochan - I need your assist, what do you think? Can we get the old behaviour in CSS?

@pawelfus, @sebastianbochan We can use [background-size](https://developer.mozilla.org/en-US/docs/Web/CSS/background-size). I added a commit with it. In my opinion it is better to do this in CSS than in the SVG file itself, as it then possible for designers to override it (for example in responsive designs) without modifying the icons. Feel free to revert if you disagree.

pawelfus:
Added all icons to the demo: https://jsfiddle.net/BlackLabel/sgbukwmj/7/ (the first group) - looks great! 

> In my opinion it is better to do this in CSS than in the SVG file itself, as it allows designers to override it (for example in responsive designs) without modifying the icons

Convinced 👍 


@nasin - could we decrease the number of vertical lines in Fibonacci timezones? Or decrease the line-width of the first one (I guess this is the one Torstein was referring to in one of his previous comments):
Zrzut ekranu 2021-06-10 o 11 54 54

nasin:
Hi, I made the lines on the fibonacci icon all the same width.

TorsteinHonsi (changes requested):
@nasin Can you remove the `samples/highcharts/website/*` stuff from this PR? They are out of scope for this PR, I assume they belong in another branch?

TorsteinHonsi (approved):

TorsteinHonsi:
Pending approval by @pawelfus

TorsteinHonsi:
Merging, as Pawel's change request seems to be fixed.

PR #15870, opened 2021-06-09
Migration: Legend

marvin19:
Removed Highcharts namespace from Legend.


ts/Core/Legend.ts
```diff
@@ -1857,6 +1757,30 @@ class Legend {
...
+
+/* *
+ *
+ *  Class Namespace
+ *
+ * */
+
+namespace Legend {
+    export interface LegendCheckBoxElement extends HTMLDOMElement {
+        checked?: boolean;
```
bre1470, line 1778: The interface name should be reduced to `CheckBoxElement` as it is in the `Legend` namespace.

bre1470 (approved):
Thank you! Looks good. 👍

TorsteinHonsi (approved):

PR #15872, opened 2021-06-09
bugfix/13868-hovering-pie-in-stock 

### LINKED ISSUE #13868: Pie series in Highstock - Error when hovering

### LINKED ISSUE #16085: Pie in stock with accessibility- error in the console caused by navigator
#### Expected behaviour
No errors in the console.

#### Actual behaviour
When the accessibility module is loaded, hovering over the pie series with `stockChart` constructor generates an error in the console.
Without a navigator works fine.

http://jsfiddle.net/BlackLabel/yx54bkq9/

Without navigator: http://jsfiddle.net/BlackLabel/f1zm87hc/

#### Product version
Highstock JS v9.1.2 (2021-06-16) on branch `bugfix/13868-hovering-pie-in-stock` which fixes some issues 

#### Affected browser(s)
All

karolkolodziej:
Fixed #13868, #16085, hovering over a `pie` series in a stock chart generated errors in the console.

Closes #16085.


TorsteinHonsi (changes requested):
Thanks!

We need a more general solution, as this bug is not specific to stock charts: https://jsfiddle.net/highcharts/pbj5ruqx/

The pie series class (along with some other classes) has a property called `noSharedTooltip` to prevent similar cases like this for shared tooltip. Split tooltips are similar to shared in the sense that they are grouped by the X axis value. So I would look for a solution around the `noSharedTooltip` property.

ts/Core/Tooltip.ts
```diff
@@ -1379,6 +1380,12 @@ class Tooltip {
 
         // single point tooltip
         } else {
+            // Hovering over series that tooltip shouldn't be split, #13868.
+            if (!isArray(pointOrPoints) && pointOrPoints.series.noSplitTooltip) {
+                tooltip.options.split = false;
+                tooltip.split = false;
+            }
```
pawelfus, line 1387: I'm afraid this might not work, because there's no reset to `split=true`. I wanted to try this demo: https://jsfiddle.net/BlackLabel/5wdz4a8y/1/ - but it fails

pawelfus (changes requested):

ts/Core/Tooltip.ts
```diff
@@ -69,6 +69,7 @@ declare module './Series/PointLike' {
 declare module './Series/SeriesLike' {
     interface SeriesLike {
         noSharedTooltip?: boolean;
+        noSplitTooltip?: boolean;
```
TorsteinHonsi, line 72: Probably better to check `noSharedTooltip` for split tooltip too. A split tooltip is also a form of shared tooltip, in the sense that multiple series are shown together. Also, `noSharedTooltip` is not limited to pie series.

pawellysy (changes requested):
Awaiting working version

ts/Core/Tooltip.ts
karolkolodziej: I decided to use `noSplitTooltip` because series like for example `scatter` might have splitted tooltip, when shared doesn't work for them.

ts/Series/PackedBubble/PackedBubbleSeries.ts
```diff
@@ -1473,7 +1473,6 @@ interface PackedBubbleSeries {
     forces: Array<string>;
     hasDraggableNodes: boolean;
     isCartesian: boolean;
-    noSharedTooltip: boolean;
```
pawelfus, line 1476: Q: Why removing props from the interface? By default `noSharedTooltip` is optional, but in this particular series it's always defined (and set to `true`). Probably for Booleans, it's not a big deal, but with other types, it might change (and require) more code.

karolkolodziej: I removed `noSharedTooltip` from the `PackedBubbleSeries` interface because this property has been already declared in the `Tooltip.ts` file as a `SeriesLike` interface.

Should it (`noSharedTooltip` property) be declared in the interface for each series individually?

@bre1470 could you please help us decide which pattern is better?

bre1470: If it is possible, that a series with shared tooltips inherits from a series without shared tooltips, it is better to define the property more generic. In that case `SeriesLike` in `Tooltip.ts` should be enough.

pawelfus: > Should it (noSharedTooltip property) be declared in the interface for each series individually?

Sorry, let me rephrase. Not for each series, but just the ones that overwrite the property. In other (coding) words:

```ts
interface SeriesLike {
  noSharedTooltip?: boolean
} 

class ColumnSeries extends SeriesLike {
}

class PieSeries extends ColumnSeries {
  noSharedTooltip: boolean // or even "true"
}

PieSeries.prototype.noSharedTooltip = true;

// Later in code:
// if (this.noSharedTooltip) { ... } <-- fine! 
```

As I said above, it's not a big deal with booleans, I just wanted to discuss this :) With e.g. objects it's a bit more troublesome:

```ts
interface customObjectInterface {
  x: number,
  y: number
}

interface SeriesLike {
  customObject?: customObjectInterface
} 

class ColumnSeries extends SeriesLike {
}

class PieSeries extends ColumnSeries {
  customObject: customObjectInterface
}

PieSeries.prototype.customObject = { x: 0, y: 0 };

// Later in code:
// `this` in an instance of PieSeries
if (this.customObject) {
   doSomething(
     this.customObject.x,
     this.customObject.y
   )
}
// Or
   doSomething(
     this.customObject?.x,
     this.customObject?.y
   )

// vs

doSomething(
  this.customObject.x,
  this.customObject.y
) // <-- `customObject` is defined
```

pawellysy (approved):

ts/Core/Series/Point.ts
```diff
@@ -1670,7 +1670,7 @@ class Point {
                     }
                 }
 
-                if (!chart.styledMode && stateMarkerGraphic) {
+                if (!chart.styledMode && stateMarkerGraphic && point.state !== 'inactive') {
```
raf18seb, line 1673: Could you shorten this to keep the 80 characters limit?

ts/Core/Tooltip.ts
```diff
@@ -1392,8 +1396,25 @@ class Tooltip {
         if (text === false) {
             this.hide();
         } else {
+            // When previously split tooltip was rendered and now if
+            // hovering over another series where split tooltip should not
+            // be used, destroy the label in order to create ordinary one.
+            // #13868
+            if (tooltip.split && point.series.noSplitTooltip && tooltip.tt && tooltip.label) {
```
raf18seb, line 1403: Could you shorten this to keep the 80 characters limit?

ts/Core/Tooltip.ts
```diff
@@ -1392,8 +1396,25 @@ class Tooltip {
...
+                tooltip.label = tooltip.label.destroy();
+                tooltip.tt = tooltip.tt.destroy();
+                this.cleanSplit();
+            }
+            // When previously tooltip was generated for noSplitTooltip series
+            // and now split one should be rendered, destroy the label.
+            if (tooltip.split && !point.series.noSplitTooltip &&
+                !(tooltip.tt || !tooltip.options.headerFormat) && tooltip.label
+            ) {
```
raf18seb, line 1412: More readable
```js
if (
    tooltip.split &&
    !point.series.noSplitTooltip &&
    !(tooltip.tt || !tooltip.options.headerFormat) &&
    tooltip.label
) {
...
```

samples/unit-tests/series-pie/pie/demo.js
```diff
@@ -320,3 +320,41 @@ QUnit.test(
...
+    chart.tooltip.refresh(chart.series[0].points[0]);
+    chart.tooltip.refresh(chart.series[1].points[0]);
+    chart.tooltip.refresh(chart.series[0].points[0]);
+    assert.ok(
+        true,
+        `After hovering over two series where one of them should not have
+        a splitter tooltip, while tooltip split explicitly set to true,
+        there should be no errors in the console.`
+    );
+});
```
raf18seb, line 360: Could we make it one QUnit test instead of 2 separate? Using `chart.update()` for example. It could improve the performance of the tests.

samples/unit-tests/series-pie/pie/demo.js
```diff
@@ -320,3 +320,41 @@ QUnit.test(
         );
     }
 );
+
+QUnit.test('Pie chart initialized through the stockChart constructor, (#14773).', function (assert) {
```
raf18seb, line 324: Using arrow function would shorten this line a little

raf18seb (changes requested):

samples/unit-tests/series-pie/pie/demo.js
karolkolodziej: Made it but found one issue while testing in karma #16085

raf18seb (approved):
Thanks!

pawelfus (approved):

khlieng:
@oysteinmoseng do internal series need forced markers?

oysteinmoseng:
@khlieng If you are referring to the navigator, then yes

ts/Core/Tooltip.ts
```diff
@@ -1392,8 +1396,31 @@ class Tooltip {
...
+            // When previously tooltip was generated for noSplitTooltip series
+            // and now split one should be rendered, destroy the label.
+            if (tooltip.split &&
+                !point.series.noSplitTooltip &&
+                !(tooltip.tt || !tooltip.options.headerFormat) &&
+                tooltip.label
+            ) {
+                tooltip.label = tooltip.label.destroy();
+            }
+
```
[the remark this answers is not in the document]
karolkolodziej: I've managed to move it to `getLabel` and slightly simplify it.
Checking for that class covers some edge cases here and there which I didn't consider so thanks for that hint ;)

ts/Core/Tooltip.ts
karolkolodziej: Unfortunately, it is the case for the flag series.
I'm afraid we have to have separate properties for that.

TorsteinHonsi: As I recall it was never the intention that flags should have a split tooltip. I think it just happened and has been patched after. I'm pushing a commit with some simplifications.

TorsteinHonsi:
@karolkolodziej I committed som simplifications, specifically removed `noSplitTooltip` because of the above discussion. I don't think it was intentional that flag series should have a split tooltip. So now we're going back to the single tooltip as is was before `split` was introduced.

Also, I simplified the destruction bit to simply run `this.destroy()`. Can you see if I am missing something there? 

https://github.com/highcharts/highcharts/blob/de1a1a699c1687da99097602b0d02091c4e559ed/ts/Core/Tooltip.ts#L493-L498

karolkolodziej:
It's fine for me if you are ok with the flag having a different tooltip ;)

The only thing I would change is the comment above that if statement describing why we are destroying it.
```
When previously split tooltip was rendered and now 
if hovering over another series where split tooltip should not be used
(or the other way around), destroy it to create a proper one
for the currently hovering series, #13868.
```

TorsteinHonsi (approved):

PR #15873, opened 2021-06-09
bugfix/15862-pane-remove-oneToOne

### LINKED ISSUE #15862: When using 9.x and Highcharts More, responsive pane size causes errors to throw on resize
#### Expected behaviour
When a chart has a `responsive` rule that affects `pane.size`, resizing the viewport in a way that causes the chart width to go past the `maxWidth` of that rule should lead to the pane size changing.

#### Actual behaviour
On some resizes, the chart content instead stays in its pre-resize position, and the following error appears in the console:

```
Uncaught TypeError: a.remove is not a function
    at eval (VM410 highcharts.js:397)
    at a.update (VM410 highcharts.js:397)
    at a.setResponsive (VM410 highcharts.js:582)
    at a.redraw (VM410 highcharts.js:363)
    at a.setSize (VM410 highcharts.js:378)
    at eval (VM410 highcharts.js:376)
```

Interestingly, it appears this only occurs if Highcharts More has been activated in the environment.

Due to the nature of the issue, I've produced two demos of the problem:

https://codesandbox.io/s/highcharts-responsive-error-demo-filgn?file=/index.html
https://jsfiddle.net/MichaelTDoyle93/Ldec6ht8/

The CodeSandbox issue more closely resembles the original setup (i.e. using the npm package rather than the CDN scripts), but the error message covers the viewport which obfuscates one of the symptoms; the jsfiddle example also illustrates that the cause isn't something specific to the package form of the library.

In both of these demos, he most consistent way I could find to reproduce the issue was to resize the preview pane down to as narrow as it could go, and then resize it out to as far as it could go.

#### Product version
HIghcharts (and Highcharts More) >= 9.0.0

#### Affected browser(s)
Chrome, Edge, Safari and Firefox (and potentially others)

In terms of a workaround, because the issue doesn't appear on version 8.2.2 a revert would likely be the best option, but obviously since 9.0.0 provided major security updates it would be less than ideal. Unfortunately, the addition of Highcharts More is also essential in our use case, as we use the polar area chart type. Thanks for your help!

khlieng:
Fixed #15862, `Chart.update` with `oneToOne` attempting to remove pane threw.

### DIFF
--- a/samples/unit-tests/pane/pane-update/demo.js
+++ b/samples/unit-tests/pane/pane-update/demo.js
@@ -73,12 +73,6 @@ QUnit.test('Pane update, single', function (assert) {
         2,
         'New border width'
     );
-
-    assert.strictEqual(
-        chart.pane[0].options.endAngle,
-        chart.options.pane.endAngle,
-        'Pane options in chart options are updated (#9917)'
-    );
 });
 
 QUnit.test('Pane update through chart.update', function (assert) {
--- a/samples/unit-tests/pane/pane-update/demo.js
+++ b/samples/unit-tests/pane/pane-update/demo.js
@@ -198,6 +198,15 @@ QUnit.test('Pane update through chart.update', function (assert) {
         'silver',
         'Pane updated by id'
     );
+
+    chart.update({
+        pane: []
+    }, true, true);
+
+    assert.ok(
+        true,
+        '#15862: oneToOne update attempting to remove pane should not throw'
+    );
 });
 
 QUnit.test('Pane update, backgrounds', function (assert) {
--- a/ts/Core/Chart/Chart.ts
+++ b/ts/Core/Chart/Chart.ts
@@ -3385,7 +3385,7 @@ class Chart {
         });
 
         itemsForRemoval.forEach(function (item: any): void {
-            if (item.chart) { // #9097, avoid removing twice
+            if (item.chart && item.remove) { // #9097, avoid removing twice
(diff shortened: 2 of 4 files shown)


TorsteinHonsi:
Thanks for the PR!

I'm trying to understand what is going on here...

The issue is marked a regression. What commit is it bisected to?

And the fact that the item is attempted removed twice may indicate that the preceding logic is flawed. Do you know why this happens?

khlieng:
So whats going on is that `Pane` has no `remove` function, so it throws. Its a regression since 9.0.0 because in 8.2.2 there was a `if (item.remove)` since the `remove` function used to get deleted from series after removing.

TorsteinHonsi:
> the size should just be updated right?

Yes, but I'm not sure whether the Pane has an `update` function at all...

khlieng:
> > the size should just be updated right?
> 
> Yes, but I'm not sure whether the Pane has an `update` function at all...

How about this one?:

https://github.com/highcharts/highcharts/blob/master/ts/Extensions/Pane.ts#L503-L519

TorsteinHonsi:
So it has one. That means we should have another look at why it is removed in the responsive setup.

khlieng:
@TorsteinHonsi the problem was that the pane options needed to be an array, because of `currentOptions()` and `collectionsWithUpdate` etc.

TorsteinHonsi (approved):

PR #15878, opened 2021-06-10
bugfix/15849-candlestick-update-color

### LINKED ISSUE #15849: Dynamic candle update does not update color
#### Expected behaviour
A red candle (open >= close) should be filled with red color.

#### Actual behaviour
 I have a problem with mass updating of chart. Sometimes some candles are filled with the wrong color (green instead of red).

I was able to reproduce this here: http://jsfiddle.net/otsybulsky/6L8t7g1y/ Just press the button "Swap". 

#### Product version
Highcharts 9.1.0

#### Affected browser(s)
All

khlieng:
Fixed #15849, candlestick fill was wrong when updating point data from rising to falling.


pawelfus (approved):
Not much to review, looks good! ;)

TorsteinHonsi (approved):
Thanks!

PR #15887, opened 2021-06-11
Migration: Tooltip

marvin19:
Removed Highcharts namespace from Tooltip.


ts/Core/Tooltip.ts
```diff
@@ -85,122 +85,6 @@ declare module '../Core/TooltipOptions'{
...
-            plotX: number;
-            plotY: number;
-        }
-        type TooltipShapeValue = ('callout'|'circle'|'square'|'rect');
-    }
-}
-
 /**
  * Callback function to format the text of the tooltip from scratch.
  *
```
bre1470, line 90: The API Declarations should be moved below the default export, followed by an `('');`, to reduce distraction.

bre1470 (approved):
Thank you! Looks good. 👍

TorsteinHonsi (changes requested):
Thank you! Please resolve conflict, otherwise approved.

TorsteinHonsi (approved):

PR #15888, opened 2021-06-11
Migration: Point

marvin19:
Removed Highcharts namespaces from Point.


bre1470 (approved):
Thanks! Looks good. 👍

bre1470 (changes requested):
API declarations (doclets) should be moved under the default export.

bre1470 (approved):

TorsteinHonsi (approved):

PR #15889, opened 2021-06-11
enhancement/12521-arc-diagram

madepiet:
Arc Diagram proof of concept: http://jsfiddle.net/BlackLabel/sycuqj2k/


TorsteinHonsi:
Judging from the description, this is not ready for (final) review, right? There is also obvious stuff missing, like samples and docs. So I think it should be converted to a draft. You can still discuss the concept with team members.

Izothep:
> Judging from the description, this is not ready for (final) review, right? There is also obvious stuff missing, like samples and docs. So I think it should be converted to a draft. You can still discuss the concept with team members.

Yes, it is still a work in progress. @madepiet could you please convert the PR to draft?

madepiet:
Sorry for my mistake, I was sure it was marked earlier

samples/highcharts/series-arcdiagram/inverted/demo.css
```diff
@@ -0,0 +1,5 @@
+#container {
+    height: 400px;
+    max-width: 800px;
```
TorsteinHonsi, line 3: The inverted arc diagram demo doesn't look good on desktop because it is 800px wide, so most of the plot area is white unused space. I suggest we simply make the `max-width` smaller.

Another room for improvement in this demo is the data labels. Currently they cover the nodes, even though there is plenty of space. So we should align them to the left of the nodes, and also make sure the overlapping prevention algorithm is not too aggressive (try setting a smaller `padding` on the data labels).

ts/Series/ArcDiagram/ArcDiagramPoint.ts
```diff
@@ -0,0 +1,154 @@
...
+    /* *
+     *
+     *  Functions
+     *
+     * */
+
+    /* eslint-disable valid-jsdoc */
+
+    /**
+     * Return a text path that the data label uses.
```
TorsteinHonsi, line 85: This is for links, right? Shouldn't we have a visual demo for this?

ts/Series/ArcDiagram/ArcDiagramSeries.ts
```diff
@@ -0,0 +1,603 @@
...
+     * @exclude      dataSorting
+     * @since        next
+     * @product      highcharts
+     * @requires     modules/arc-diagram
+     * @optionparent plotOptions.arcdiagram
+     */
+    public static defaultOptions: ArcDiagramSeriesOptions = merge(SankeySeries.defaultOptions, {
+        nodeShape: 'circle',
+        /**
+         * Wheter the nodes with different values should have the same size.
```
TorsteinHonsi, line 74: Whether nodes with...

ts/Series/ArcDiagram/ArcDiagramSeriesOptions.d.ts
```diff
@@ -0,0 +1,43 @@
...
+
+/* *
+ *
+ *  Declarations
+ *
+ * */
+
+export interface ArcDiagramSeriesOptions extends SankeySeriesOptions {
+    center?: Array<(number|string|null)>;
+    startAngle?: number;
```
TorsteinHonsi, line 31: Do we need `startAngle`? Or is it just a remnant from the dependency wheel series?

ts/Series/ArcDiagram/ArcDiagramSeries.ts
```diff
@@ -0,0 +1,603 @@
...
+     *         Arc Diagram
+     *
+     * @extends      plotOptions.sankey
+     * @exclude      dataSorting
+     * @since        next
+     * @product      highcharts
+     * @requires     modules/arc-diagram
+     * @optionparent plotOptions.arcdiagram
+     */
+    public static defaultOptions: ArcDiagramSeriesOptions = merge(SankeySeries.defaultOptions, {
```
TorsteinHonsi, line 71: Missing docs for some of the options, like `majorRadius` and `linkWeight`.

TorsteinHonsi (changes requested):
Great work! Some change requests.

samples/highcharts/series-arcdiagram/link-weight/demo.js
```diff
@@ -0,0 +1,64 @@
+Highcharts.chart('container', {
+
+    title: {
+        text: 'Highcharts Arc Diagram'
+    },
+    subtitle: {
+        text: 'Arc Diagram with fixed link weight'
+    },
+    series: [{
+        linkWeight: 1,
```
pawelfus, line 10: I would add `centeredLinks: true,` to this demo:
Zrzut ekranu 2022-01-12 o 11 00 11

ts/Series/ArcDiagram/ArcDiagramSeries.ts
```diff
@@ -0,0 +1,603 @@
...
+        nodeShape: 'circle',
```
pawelfus, line 72: Missing doclet.

Would it make sense to reuse `marker` API for nodes? Then use the marker's logic which is already implemented (color, stroke etc.)

ts/Series/ArcDiagram/ArcDiagramSeries.ts
```diff
@@ -0,0 +1,603 @@
...
+ */
+
+/**
+ * An array of data points for the series. For the `arcdiagram` series
+ * type, points can be given in the following way:
+ *
+ * An array of objects with named values. The following snippet shows only a
+ * few settings, see the complete options set below. If the total number of data
+ * points exceeds the series' [turboThreshold](#series.area.turboThreshold),
+ * this option is not available.
```
pawelfus, line 575: Since it's the only option to configure the data, then shouldn't be `turboThreshold` disabled? 🤔 

It seems it's not the only series with this bug. other types: dependency wheel, flags, networkgraph, organization, sankey. Perhaps default `keys` should be set to `['from', 'to', 'weight']`? This is the way `MapSeries` and `Wordcloud` solve this issue (provide two default ways to configure the data).

ts/Series/ArcDiagram/ArcDiagramSeries.ts
```diff
@@ -0,0 +1,603 @@
...
+ * @product   highcharts
+ * @excluding outgoing, dataLabels
+ * @apioption series.arcdiagram.data
+ */
+
+/**
+ * Individual data label for each node. The options are the same as
+ * the ones for [series.arcdiagram.dataLabels](#series.arcdiagram.dataLabels).
+ *
+ * @apioption series.arcdiagram.nodes.dataLabels
```
pawelfus, line 600: Is this the only option in `nodes` array? I think something is missing here: http://localhost:9005/highcharts/series.arcdiagram.nodes.dataLabels (e.g. `id`)
Zrzut ekranu 2022-01-12 o 11 22 28

ts/Series/ArcDiagram/ArcDiagramSeriesOptions.d.ts
```diff
@@ -0,0 +1,43 @@
...
+    states?: SeriesStatesOptions<ArcDiagramSeries>;
+    equalNodes?: boolean;
+    centeredLinks?: boolean;
+    majorRadius?: number;
+    reversed: boolean;
+    scale?: number;
+    nodeShape?: string; // experimental
+    centerPos?: string;
+    linkWeight?: number;
```
pawelfus, line 40: It seems all of the options listed above are available, but undocumented

ts/Series/DependencyWheel/DependencyWheelPointOptions.d.ts
```diff
@@ -26,6 +26,7 @@ import type SankeyPointOptions from '../Sankey/SankeyPointOptions';
 
 export interface DependencyWheelPointOptions extends SankeyPointOptions {
     // nothing here yet
```
pawelfus, line 28: misleading comment

ts/Series/NodesComposition.ts
```diff
@@ -239,7 +239,6 @@ namespace NodesComposition {
                 coll: string
             ): (number|undefined) {
                 let offset = 0;
-
```
pawelfus, line 242: Unnecessary change ;)

ts/Series/Organization/OrganizationSeries.ts
```diff
@@ -656,11 +629,11 @@ extend(OrganizationSeries.prototype, {
...
-namespace OrganizationSeries {
-    export interface ColumnArray<T = OrganizationPoint> extends SankeySeriesType.ColumnArray<T> {
-        offset(node: T, factor: number): (Record<string, number>|undefined);
-    }
-}
+// namespace OrganizationSeries {
+//     export interface ColumnArray<T = OrganizationPoint> extends SankeySeriesType.ColumnArray<T> {
+//         offset(node: T, factor: number): (Record<string, number>|undefined);
+//     }
+// }
```
pawelfus, line 636: If unnecessary, please remove

ts/Series/Organization/OrganizationSeries.ts
```diff
@@ -31,14 +31,14 @@ import type {
...
 import type { SankeyDataLabelFormatterContext } from '../Sankey/SankeyDataLabelOptions';
-import type SankeySeriesType from '../Sankey/SankeySeries';
 import type { StatesOptionsKey } from '../../Core/Series/StatesOptions';
 import type SVGAttributes from '../../Core/Renderer/SVG/SVGAttributes';
 import type SVGLabel from '../../Core/Renderer/SVG/SVGLabel';
 import type SVGPath from '../../Core/Renderer/SVG/SVGPath';
 import OrganizationPoint from './OrganizationPoint.js';
 import { Palette } from '../../Core/Color/Palettes.js';
 import SeriesRegistry from '../../Core/Series/SeriesRegistry.js';
+import SankeyColumnComposition from '../Sankey/SankeyColumnComposition.js';
```
pawelfus, line 41: It seems importing just type should be enough here

ts/Series/ArcDiagram/ArcDiagramSeries.ts
```diff
@@ -0,0 +1,603 @@
...
+            chart = series.chart,
+            // column needs casting, to much methods required at the same time
+            column = SankeyColumnComposition.compose([] as Array<ArcDiagramPoint>, series);
+
+        column.sankeyColumn.maxLength = chart.inverted ?
+            chart.plotHeight : chart.plotWidth;
+
+        // Get the translation factor needed for each column to fill up the
+        // plot height
+        column.sankeyColumn.getTranslationFactor = (
```
pawelfus, line 122: Q: It seems right now, Sankey adds `sankeyColumn` to the point via composition. Then ArcDiagram overwrites some of the methods, right? Would it make sense to use a generic NodeColumn composition, and both, Sankey and ArcDiagram implement separate compositions, both conforming to the NodeColumn one?

pawelfus (changes requested):
Nice work!

A few minor things in addition to inline comments:
- Missing `readme.md` for `samples/highcharts/demo/arc-diagram`
- API is not excluding options from Sankey/Dependency wheel/Line (for example: http://localhost:9005/highcharts/series.arcdiagram.curveFactor, dragDrop, getExtremesFromAll, nodePadding)
- Description refers to `Sankey`: http://localhost:9005/highcharts/series.arcdiagram.dataLabels

ts/Series/ArcDiagram/ArcDiagramSeries.ts
Izothep: `SankeyColumn` already takes methods from both `Arc Diagram` and `Sankey`, however, if the same method is used in both series types differently, it takes it from the `Sankey` by default. It might be possible to simply move `SankeyColumn` to the other, more generic place but the code will not change much.

ts/Series/ArcDiagram/ArcDiagramSeries.ts
Izothep: Partially done the migration to the `marker` (`symbol` might be used now) but the rest would require a change of the concept of how nodes are drawn - currently, it is an extension of the `sankey` which is an extension of the `column` series.

Izothep:
@TorsteinHonsi @pawelfus Thank you for the comments. I have made the requested changes now. There are still two issues that should be discussed though: 
1. Renaming and moving `SankeyColumn` as a general composition.
2. Changing how points are drawn so it will use `markerAttribs`.

ts/Series/ArcDiagram/ArcDiagramSeries.ts
pawelfus: Aaah, right, thanks 👍

ts/Series/ArcDiagram/ArcDiagramSeries.ts
pawelfus: Okay, thanks 👍

Izothep:
After a discussion with @pawelfus, I believe that we may skip point 1. from the above message for the time being. 
I have worked on changing how the `nodes` are drawn so now they are using `markerAttribs`. 

I think that this PR is ready for the second review.

ts/Series/ArcDiagram/ArcDiagramSeries.ts
```diff
@@ -0,0 +1,782 @@
...
+         * @default false
+         * @product highcharts
+         */
+        centeredLinks: false,
+
+        /**
+         * The radius of the link arc. If not set,
+         * it is calculated automatically.
+         *
+         * @type    {boolean}
```
TorsteinHonsi, line 96: Wrong type

ts/Series/ArcDiagram/ArcDiagramSeries.ts
```diff
@@ -0,0 +1,782 @@
...
+         * @since next
+         * @default false
+         * @product highcharts
+         */
+        centeredLinks: false,
+
+        /**
+         * The radius of the link arc. If not set,
+         * it is calculated automatically.
```
TorsteinHonsi, line 94: Give a more elaborate explanation of how it works. By experimenting I found that:
- If not set, it renders a semi-circle between the nodes, except when overflowing the edge of the plot area, in which case an arc touching the edge is rendered.
- If set, an arc extending to the given `radius` is always rendered.

ts/Series/ArcDiagram/ArcDiagramSeries.ts
```diff
@@ -0,0 +1,782 @@
...
+
+        /**
+         * The offset of an arc diagram nodes column in relation
+         * to the plotArea. The offset equal to 50% places nodes
+         * in the center of a chart.
+         *
+         * @type    {string}
+         * @since next
+         * @default undefined
+         * @product highcharts
```
TorsteinHonsi, line 111: It looks to me that the default behaviour is like setting this to `100%`? We can spell out that in the options (in JavaScript), then we don't need to handle `undefined`, and it makes the typing simpler.

ts/Series/ArcDiagram/ArcDiagramSeries.ts
```diff
@@ -0,0 +1,782 @@
...
+         * @apioption series.arcdiagram.offset
+         */
+
+        /**
+         * The global link weight. If not set, width will is calculated
```
TorsteinHonsi, line 116: Typo: "width will is"

ts/Series/ArcDiagram/ArcDiagramSeries.ts
```diff
@@ -0,0 +1,782 @@
...
+        /**
+         * Whether nodes with different values should have the same size.
+         *
+         * @type    {boolean}
+         * @since next
+         * @default false
+         * @product highcharts
+         */
+        equalNodes: false,
+        keys: ['from', 'to', 'weight'],
```
TorsteinHonsi, line 135: The `keys` should not be part of the default options. Use `pointArrayMap` instead. In this case I think it should be inherited.

ts/Series/ArcDiagram/ArcDiagramSeries.ts
```diff
@@ -0,0 +1,782 @@
...
+        /**
+         * Whether series should be placed on the other side of plotArea.
```
TorsteinHonsi, line 137: Whether _the_ series. And write either plot area in plain language or `plotArea` in backticks.

ts/Series/ArcDiagram/ArcDiagramSeries.ts
```diff
@@ -0,0 +1,782 @@
...
+         * @excluding enabled, enabledThreshold, height, radius, width
+         */
+        marker: {
+            symbol: 'circle',
+            /**
+             * In arc-diagram charts, the radius is overridden and determined
+             * based on the point's data value.
+             *
+             * @ignore-option
+             */
```
TorsteinHonsi, line 185: Missing docs on how to control the size of the marker when `equalNodes` is true. It looks - from the code - to be possible to control it using `nodePadding`. But maybe we also should allow setting `marker.radius` so that it is possible to set it to a fixed value? For example if I want to make it the same height as my data labels.

pawelfus (changes requested):
Looks great, just one thing:
- open any demo, hover any link -> tooltip is not displayed. It works in other node-based series (Sankey, Networkgraph, Dependency wheel, etc.)

TorsteinHonsi (changes requested):
Nice! The main demo is beautiful. Further comments inline.

Izothep:
Thanks again for the review! I believe that I have changed all the mentioned parts of the code now. Tooltip problem is solved, also one `dataLabels` issue is fixed with the newest code.

pawelfus (approved):
Thanks! 👍

TorsteinHonsi (approved):
Awesome 🎉

PR #15893, opened 2021-06-11
bugfix/5430-state-marker-zone-color

### LINKED ISSUE #5430: Zone color on hover incorrect when markers are hidden
#### Expected behaviour

When markers are hidden because of small distances between points (large points, zoomed-out axis), markers still show on hover. Hover colors should match what they would be if those markers were not hidden.
#### Actual behaviour

Marker colors don't match the correct zone, though line colors do

http://jsfiddle.net/ncchm8v1/
#### Affected browser(s)

Happens in both Firefox and Chrome in Linux, Chrome in Windows

Note, I mentioned this in issue a comment on #4670 but thought it might be better to open a new issue rather than reopen that one, as I'm not sure if they're related

khlieng:
Fixed #5430, marker shown on hover with disabled markers missed CSS class.

### DIFF
--- a/samples/unit-tests/series/marker/demo.js
+++ b/samples/unit-tests/series/marker/demo.js
@@ -176,6 +176,11 @@ QUnit.test('Marker size and position', function (assert) {
         Math.floor(series.stateMarkerGraphic.attr('y')),
         'Correct image y-position (#7273)'
     );
+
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #15898, opened 2021-06-14
bugfix/15896-gantt-scrollbar

### LINKED ISSUE #15896: Gantt chart scroll bar issue on upgrade to 9.1.1
#### Expected behaviour
Gantt Chart should show Y axis scroll bar with below settings

		yAxis:{
					max:2,
            scrollbar: {
                enabled: true,
            },
			},

This works fine with the version 8.0.0, when upgraded to 9.1.1, scroll bar disappeared with the same config

In the provided jsfiddle, which is pointing to 8.0.0, you will see the scrollbar, but if you change it to point to 9.1.1, the scrollbar will disappear. Kindly guide us what needs to be done for scroll bar to appear in version 9.1.1

https://jsfiddle.net/sreejithga/ph703yz8/6/

#### Product version
9.1.1

#### Affected browser(s)
Chrome

khlieng:
Fixed #15896, scrollbar did not work with gantt.

### DIFF
--- a/ts/masters/modules/stock.src.ts
+++ b/ts/masters/modules/stock.src.ts
@@ -22,10 +22,8 @@ import '../../Core/Navigator.js';
 import '../../Extensions/RangeSelector.js';
 import StockChart from '../../Core/Chart/StockChart.js';
 const G: AnyRecord = Highcharts;
-// Compositions
-if (!G.Scrollbar) {
-    Scrollbar.compose(G.Axis);
-}
 // Classes
 G.Scrollbar = Scrollbar;
 G.StockChart = G.stockChart = StockChart.stockChart;
+// Compositions
+Scrollbar.compose(G.Axis);
--- a/ts/masters/modules/stock.src.ts
+++ b/ts/masters/modules/stock.src.ts
@@ -22,8 +22,10 @@ import '../../Core/Navigator.js';
 import '../../Extensions/RangeSelector.js';
 import StockChart from '../../Core/Chart/StockChart.js';
 const G: AnyRecord = Highcharts;
+// Compositions
+if (!G.Scrollbar) {
+    Scrollbar.compose(G.Axis);
+}
 // Classes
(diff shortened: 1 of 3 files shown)


ts/masters/modules/stock.src.ts
```diff
@@ -22,8 +22,10 @@ import '../../Core/Navigator.js';
...
 const G: AnyRecord = Highcharts;
+// Compositions
+if (!G.Scrollbar) {
+    Scrollbar.compose(G.Axis);
+}
 // Classes
 G.Scrollbar = Scrollbar;
 G.StockChart = G.stockChart = StockChart.stockChart;
-// Compositions
-Scrollbar.compose(G.Axis);
```
bre1470, line 29: The order should not be important for the composition here. (= compositions usually come last, after all namespace assignments) Was something failing?

bre1470 (approved):

bre1470 (changes requested):
Wait a second. The change means, that composition has no detection agains running twice. This should be fixed with a private array in the compose function, taking all modified axis class and then next time check, if the array already contains the axis class to modify.

khlieng:
> Wait a second. The change means, that the composition has no detection agains running twice. This should be fixed with a private array in the compose function, taking all modified axis class and then next time check, if the array already contains the axis class to modify.

Doesnt the `if (!G.Scrollbar)` handle that?

TorsteinHonsi (approved):
Thanks!

TorsteinHonsi (changes requested):
Pending bre1420's comment

bre1470:
Without fixing the composition only the master is compatible to tree shaking. If a user calls the composition in a ES6 module it has not protection agains double modification. (which triggers the bug)

bre1470 (approved):
Thank you! Looks good. 👍

TorsteinHonsi (approved):

PR #15899, opened 2021-06-14
enhancement/15348-heikin-ashi

karolkolodziej:
Enhancement #15348, added Heikin Ashi series.

To do's:
- [x] - new series/options implementation
 - [x]  - ~~update [SCSS](https://github.com/highcharts/highcharts/blob/master/css/highcharts.scss) (in case of new styled-mode options) and the [list](https://github.com/highcharts/highcharts/blob/master/docs/chart-design-and-style/style-by-css.md) (if necessary)~~
- [x] - add to [Stock Tools'](https://www.highcharts.com/demo/stock/stock-tools-gui) button list of series types
- [x] - official [demo](https://www.highcharts.com/demo/stock) (this will cover visual test too)
- [x] - official [general documentation](https://www.highcharts.com/docs/stock/candlestick-chart)
- [x] - update [API](https://api.highcharts.com/highstock/) (series/options description)
- [x] - fix `heikinashi` with `colorAxis`
- [x] - TS
- [x] - add `dataGrouping` option
- [x] - position demo


docs/stock/heikinashi.md
```diff
@@ -0,0 +1,10 @@
+Candlestick chart
```
TorsteinHonsi, line 1: Wrong header?

docs/stock/heikinashi.md
```diff
@@ -0,0 +1,10 @@
...
+================
+
+A Heiken Ashi series is typically used to present the modified open, high, low and close price over a period of time. 
+A Heiken Ashi chart is composed of a body and an upper and a lower wick similarly to the candlestick. The only difference is that the open, high, low and close values had been modified by the formula.
```
TorsteinHonsi, line 5: > A Heiken Ashi chart

Should we use a different word than "chart"? This is an item or a point, right?

.. _have_ been modified by the formula.

samples/stock/demo/heikinashi/demo.details
```diff
@@ -0,0 +1,11 @@
+---
+name: Candlestick
```
TorsteinHonsi, line 2: Wrong name?

samples/stock/demo/heikinashi/demo.js
```diff
@@ -0,0 +1,44 @@
...
+            height: '50%'
+        }, {
+            title: {
+                text: 'Heikin Ashi'
+            },
+            top: '50%',
+            height: '50%',
+            offset: 0
+        }],
+        series: [{
```
TorsteinHonsi, line 33: It is not easy to see the difference between the candlestick and the Heikin Ashi without zooming in. I propose setting `rangeSelector.selected` to 1 in the config.

samples/stock/demo/heikinashi/demo.js
```diff
@@ -0,0 +1,44 @@
...
+
+    for (i; i < dataLength; i += 1) {
+        ohlc.push([
+            data[i][0], // the date
+            data[i][1], // open
+            data[i][2], // high
+            data[i][3], // low
+            data[i][4] // close
+        ]);
+    }
```
TorsteinHonsi, line 14: As far as I can see, this is not necessary. It just copies the data. We can just use `data` directly on both series.

TorsteinHonsi (changes requested):

TorsteinHonsi:
Thanks for the PR! Please see my inline comments.

samples/stock/demo/heikinashi/demo.js
```diff
@@ -0,0 +1,33 @@
...
+    Highcharts.stockChart('container', {
+        title: {
+            text: 'Candlestick and Heiken Ashi series comparison.'
+        },
+        rangeSelector: {
+            selected: 1
+        },
+        yAxis: [{
+            title: {
+                text: 'OHLC'
```
pawellysy, line 11: `candlestick`?

pawellysy (changes requested):
minor changes

ts/Series/HeikinAshi/HeikinAshiPoint.ts
```diff
@@ -0,0 +1,148 @@
...
+        const point = this,
+            series = point.series,
+            yAxis = series.yAxis,
+            translated = [
+                'plotOpen',
+                'plotHigh',
+                'plotLow',
+                'plotClose',
+                'yBottom'
+            ]; // translate OHLC for
```
raf18seb, line 70: Is this comment finished? Or is it indicating the below `forEach` loop?

ts/Series/HeikinAshi/HeikinAshiPoint.ts
```diff
@@ -0,0 +1,148 @@
...
+
+        [point.open, point.high, point.low, point.close, point.low]
+            .forEach(
+                function (value, i): void {
+                    if (value !== null) {
+                        (point as any)[translated[i]] =
```
raf18seb, line 76: Does it have to be `as any`? Can't we add a TS definition for `point.translated`?

ts/Series/HeikinAshi/HeikinAshiPoint.ts
```diff
@@ -0,0 +1,148 @@
...
+     *
+     */
+    public modifyFirstPointValue(): void {
+        const point = this,
+            newClose = (point.open + point.high + point.low + point.close) / 4,
+            newOpen = (point.open + point.close) / 2;
+
+        point.open = newOpen;
+        point.close = newClose;
+    }
```
raf18seb, line 104: Just a suggestion, this could be shorter
```js
public modifyFirstPointValue(): void {
    this.open = (this.open + this.high + this.low + this.close) / 4,
    this.close = (this.open + this.close) / 2;
}
```

ts/Series/HeikinAshi/HeikinAshiPoint.ts
```diff
@@ -0,0 +1,148 @@
...
+
+    /**
+     * Calculate and modify the point value.
```
raf18seb, line 107: ... point's values.

ts/Series/HeikinAshi/HeikinAshiPoint.ts
```diff
@@ -0,0 +1,148 @@
...
+            newClose = (point.open + point.high + point.low + point.close) / 4,
+            newHigh = Math.max(point.high, newClose, newOpen),
+            newLow = Math.min(point.low, newClose, newOpen);
+
+        point.open = newOpen;
+        point.close = newClose;
+        point.high = newHigh;
+        point.low = newLow;
+    }
+}
```
raf18seb, line 130: Same as above - could be shorter

raf18seb (changes requested):

ts/Series/HeikinAshi/HeikinAshiPoint.ts
karolkolodziej: I'm using `newOpen` and `newClose` later.
I am afraid I won't be able to make it shorter.

ts/Series/HeikinAshi/HeikinAshiPoint.ts
karolkolodziej: It's not that easy to get rid of that `any`. 
Additional after some research, I see that we use that approach in other similar cases like this one so I guess for now it has to stay like this.

pawellysy (approved):
nothing to add! 👍

raf18seb (approved):
GJ, thanks!

pawelfus (changes requested):
Looks great! Just minor polishing.

TODO: Decide if should be part of the Stock module, or a separate one

karolkolodziej:
> TODO: Decide if should be part of the Stock module, or a separate one

I guess this is up to @TorsteinHonsi 
Could you please help us decide?

pawelfus (changes requested):
Could you test this?
Try to use `series.modifyValue(point)`. I know you don't have access to the previous point with the current implementation, but adding an extra argument (`series.modifyValue(point, index)`) would let you access the previous point. Then you could remove the extra loop from translate()

pawelfus:
> > TODO: Decide if should be part of the Stock module, or a separate one
> 
> I guess this is up to @TorsteinHonsi
> Could you please help us decide?

The longer I think about this, the more I'm leaning towards a separate module. Heikin Ashi and Hollow Candlestick series are just the two first series out of a few new ones I want to be implemented (Kagi, Renko, and Point&Figure). If every series will add 1kB+ to the package (I expect even more from the next three types), then it's the right moment to start releasing the Stock series as modules. I'm not sure how does this align with tree-shaking, though.

TorsteinHonsi:
> it's the right moment to start releasing the Stock series as modules. I'm not sure how does this align with tree-shaking, though.

Agreed. To my knowledge, tree shaking shouldn't be affected by how the master files are set up.

> The only problem I see (with modularized series) is dataGrouping, which doesn't have a mechanism to easily extend it.

FYI @bre1470 . This should be taken into account when refactoring/migrating the data grouping.

TorsteinHonsi (approved):
Looks good to me, pending final approval by @pawelfus

ts/Extensions/DataGrouping.ts
```diff
@@ -923,6 +923,9 @@ const baseProcessData = seriesProto.processData,
         },
         ohlc: {
             groupPixelWidth: 5
+        },
+        heikinashi: {
```
pawelfus, line 927: Please, add a comment to move this to `HeikinAshiSeries.ts` when dataGrouping will support any form of extending with new types

pawelfus (changes requested):

ts/Series/HeikinAshi/HeikinAshiSeries.ts
```diff
@@ -0,0 +1,364 @@
...
+     * @return {void}
+     *
+     */
+    public getHeikinashiData(): void {
+        const series = this,
+            processedYData = series.allGroupedData || series.yData, // procesed and grouped data
+            heikiashiData = series.heikiashiData;
+
+        if (!heikiashiData.length && processedYData && processedYData.length) {
+            const firstPoint: any = processedYData[0];
```
pawelfus, line 157: Why casting to `any`? The same below:
```
                const dataPoint: any = processedYData[i],
                    previousDataPoint: any = heikiashiData[i - 1];
```

ts/Series/HeikinAshi/HeikinAshiSeries.ts
```diff
@@ -0,0 +1,364 @@
...
+}
+
+// Modify points after base translation.
+addEvent(HeikinAshiSeries, 'afterTranslate', function (): void {
+    this.assignDataPoints();
+});
+
+// Force to recalculate the heikinashi data set after updating data.
+addEvent(HeikinAshiSeries, 'updatedData', function (): void {
+    if (!this.heikiashiData) {
```
pawelfus, line 235: This check seems to be unnecessary (`this.heikinashiData` is set on `init`)

pawelfus (changes requested):
Looks great, just some minor polishing, thanks!

ts/Series/HeikinAshi/HeikinAshiSeries.ts
karolkolodziej: The `firstPoint` is an element from the `processedYData` so it might have a type like: `number|null|undefined`.
However, after removing the `undefined` from all definitions everything works as expected. 
(I guess this type might be omitted but I didn't want to make a mess in that PR.)

So my only concern was the `number|null` types. 
In calculation, adding `null` to something doesn't change the result, so I cast it to any to avoid checks or conversion from `null` to `0`.
Is there some easy hack to avoid that casting?

karolkolodziej: Comment about that is added below because previously I cast while invoking the method.

`this.modifyDataPoint(dataPoint, previousDataPoint);`

pawelfus: Okay, thanks for the explanation!

pawelfus (approved):
Great work! Thank you :)

pawellysy (approved):

raf18seb (approved):
GJ, thanks!

Issue #6218, opened 2017-01-09
Can't update tooltip's valuePrefix and valueSuffix 


zhy1stgg:
I want to update tooltip's valuePrefix or valueSuffix  by chart.update, but it no effect.

demo:  https://jsfiddle.net/m9uzuht9/

sebastianbochan:
_Workaround_
Update each series:
- https://jsfiddle.net/of56mrsr/

zhy1stgg:
Hi @sebastianbochan 

I just want to make sure it's a bug while chart.update can update toolip.share and other options but can't update valuePrefix or valueSuffix.

thanks

oysteinmoseng:
Hi @zhy1stgg, thanks for reporting, it's indeed a bug. Sebastian's posted fiddle is just a workaround.

PR #15904, opened 2021-06-15
Enhancement: Chart.getOptions function

TorsteinHonsi:
Added new function, [Chart.getOptions()](https://api.highcharts.com/class-reference/Highcharts.Chart#getTable), to get the current active configuration options for the chart. 

#### Upgrade notes
For Highcharts Stock, the ordering of series and axes in the `chart.series`, `chart.xAxis`, `chart.yAxis` and `chart.axes` arrays were changed. Previously the navigator series and axes were last of the initial items, and when adding items they were appended. From now on, the internal navigator series and axes are always last, so the index of items now corresponds to that of the current options.

Some refactoring was made
- `chart` user options are now kept in sync with series and axis user options, so that `chart.xAxis[0].userOptions === chart.userOptions.xAxis[0]`. If that is not the case, for example it is broken after an update, it should be considered a bug.
- The navigator y axis and x axis are now pushed to the end of `chart.xAxis` and `chart.yAxis` so that the indexing of the axes corresponds to that of the options.

# To do (critical, better to land fast)
 - [x] Make sure collections like series and axes are reflected correctly in `chart.options`.
 - [x] Internal refactoring, combine `cleanRecursive` with the Responsive module's `getCurrent` function.
 - [x] Create the new `getOptions` function (POC in [studies](http://utils.highcharts.local/samples/view?path=highcharts/studies/chart-getoptions))
    - [x]  Return `diffObjects` on `chart.options` (or `chart.userOptions`?) compared to `defaultOptions`.
 - [x] Internal axes (navigator) are now part of `chart.options.yAxis` and `xAxis`. Reconsider that. Tests need to be changed, but are there other consequences of removing them?
 - [x] Sample and more elaborate doclet on `getOptions`

# To do (non critical, can wait)
 - [x]  Unit tests and cleaning of results. Results of `getOptions` should `deepEqual` the initial options, possibly except some internal, unharmful modifications to the options. And except items that can be objects or arrays, like `series`, `xAxis` and `yAxis`. These should be in array form in `getOptions`.
 - [ ] Figure out how to deal with `colorAxis` on top level. Currently setting `colorAxis: {}` enables the color axis. Test how themes with `colorAxis` defined affects charts. It may be that this needs to be changed altogether, for example by requiring `colorAxis: { enabled: true }`.
 - [ ] Figure out how to deal with `Highcharts.setOptions` currently modifying `defaultOptions`, making it unusable for identifying the actual user options. We need a new property storing the default default options (`factoryDefaults` ?).
 - [ ] Use the new `getOptions` in the exporting module? Low gain, high risk. Probably now worth it.
 - [ ] Go over the usage of `userOptions`. The original intention of this property was to distinguish between actual user options and default options. But there has been an inflation in its usage. In most cases it could simply be replaced by `options`, that should always be up to date. In other cases, we can use `cleanRecursive` to find out whether an option is set by the user or not. For complex objects, like axes and series, the `options` are built in multi-level merges. In those cases, we probably need to split the merge so that it is possible to built a composite default options object to be able to diff out the user options.
    - [x] Check specifically the #6218 fix, that sets `userOptions` inside the `tooltip.options`. Should be refactored.


pawelfus:
> Make sure collections like series and axes are reflected correctly in `chart.options`.

Do you expect any issues with `navigator` axes?

ts/Core/Chart/Chart.ts
```diff
@@ -664,29 +641,112 @@ class Chart {
...
+     *
+     * @private
+     * @function Highcharts.Chart#insertItem
+     * @param  {Highcharts.Series|Highcharts.Axis} item
+     *         The item to insert
+     * @param  {Array<Highcharts.Series>|Array<Highcharts.Axis>} collection
+     *         A collection of items, like `chart.series` or `xAxis.series`.
+     * @return {number} The index of the series in the collection.
+     */
+    public insertItem(
```
pawelfus, line 657: Q: Would it make sense to move this function to Utilities? Since there's no need for the context I'm not sure if we need to keep it in Chart class.

Perhaps, with some minor refactoring in force-directed-graph I could use it in the future too (instead of `addElementsToCollection` and `removeElementFromCollection`)

pawelfus (approved):
Amazing work! Just a two minor questions.

PS: Since there's a change in indexing in Stock (navigator is the last one), do we need an "Upgrade notes" part?

ts/Core/Chart/Chart.ts
TorsteinHonsi: Sure, will do 👍

TorsteinHonsi:
Ready for merge, pending CI

PR #15905, opened 2021-06-15
Removed nightly trigger for demo deploy from ci config

goransle:
Removed the `check_for_samples_changes_and_deploy` job from the nightly workflow, as it seems the scheduled workflows are a bit more limited when it comes to CI credentials. This functionality is being moved to [highcharts/highcharts-demo-manager](/highcharts/highcharts-demo-manager).


TorsteinHonsi (approved):
Thanks!

PR #15906, opened 2021-06-15
bugfix/15902-legend-proxy-series-update

### LINKED ISSUE #15902: A11Y Proxy Container doesn't update when chart data updates
#### Expected behaviour

When the series of a chart changes, the hidden A11Y buttons/elements should update as well.

#### Actual behaviour
The A11Y buttons still displays the original series.

http://jsfiddle.net/BlackLabel/kz5yneb6/

Open the console, click the button and notice that the original series is still listed

#### Product version
Highcharts 9.1.1 with Accessibility Module

#### Affected browser(s)
Chrome

Note: I originally reported this in [highcharts-react](https://github.com/highcharts/highcharts-react/issues/303) and Paweł sent me here.

khlieng:
Fixed #15902, legend proxy buttons did not update when updating series.

- [x] Fix `addSeries` too?

### DIFF
--- a/samples/unit-tests/accessibility/series-showhide/demo.js
+++ b/samples/unit-tests/accessibility/series-showhide/demo.js
@@ -42,5 +42,15 @@ QUnit.test(
             'true',
             'Series should be hidden from AT'
         );
+
+        chart.series[0].update({
+            name: 'Bean sprouts'
+        });
+
+        assert.notStrictEqual(
+            chart.series[0].a11yProxyElement.ariaLabel.indexOf('Bean'),
+            -1,
+            '#15902: Proxy button aria-label should have been updated'
+        );
     }
 );
--- a/samples/unit-tests/accessibility/series-showhide/demo.js
+++ b/samples/unit-tests/accessibility/series-showhide/demo.js
@@ -48,7 +48,7 @@ QUnit.test(
         });
 
         assert.notStrictEqual(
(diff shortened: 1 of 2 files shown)


TorsteinHonsi:
> Fix addSeries too?

It should reflect all CRUD operations, right @oysteinmoseng ?

oysteinmoseng:
Yes, this is just an HTML overlay for the legend, so needs to update whenever the legend moves somewhere (`chart.update`) or has changed content. Also note `series.addPoint` if we have points in the legend (e.g. in pies).

Without testing, it looks like the current solution would be good if we also recreate the proxies if the list of proxies doesn't match the series/points.

Honestly we could probably get away with always just recreating them on every render as well, since there aren't usually that many legend items..

khlieng:
Should probably just try to find some way to check if the legend has been dirty.

oysteinmoseng:
Yes, perhaps instead of doing this on `chart.render` it would be better to attach an event (use the EventProvider) to whenever the legend has re-rendered

TorsteinHonsi:
If it is hooked up to `Legend.renderItem` or `Legend.layoutItem` it should automagically handle whatever is thrown at it. Or use the Legend `afterRender` event, which can access `Legend.allItems`.

TorsteinHonsi (approved):
Looks good to me 👍 . Pending approval by @oysteinmoseng .

oysteinmoseng (approved):
Nice work!

PR #15912, opened 2021-06-16
bugfix/15909-pie-dataLabel-useHTML-visible

### LINKED ISSUE #15909: 3d Pie chart Error when using HTML dataLabels and data with 'visible: false' property
#### Expected behaviour
Display 3d Pie chart, and data where "visible: false" will not be shown

#### Actual behaviour
Chart does not display and there is a Uncaught TypeError: g is undefined error

This seems to only be a problem when using useHTML: true in the plotOptions.pie.dataLabels property

https://jsfiddle.net/acwgz0hm/1/

#### Product version

Highcharts 9.1.1

#### Affected browser(s)
Firefox 89.0

khlieng:
Fixed #15909, pie point with `visible` set to `false` and `useHTML` data labels threw.

### DIFF
--- a/samples/unit-tests/series-pie/datalabel/demo.js
+++ b/samples/unit-tests/series-pie/datalabel/demo.js
@@ -464,7 +464,7 @@ QUnit.test(
                         data: [
                             ['Firefox', 44.2],
                             ['IE7', 26.6],
-                            ['IE6', 20],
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #15918, opened 2021-06-16
bugfix/15864-navigator-rangeSelector-regresion 

### LINKED ISSUE #15864: Navigator and rangeSelector in 9.1.1 different from 9.1.0
#### Expected behaviour
When selected one  of the ranges I expect the range in the navigator to be selected on the right

#### Actual behaviour
It is selected on the left so the first part of the navigator is selected iso the last part.

Explanation and jsFiddle can be seen here: https://www.highcharts.com/forum/viewtopic.php?f=12&t=46704#p167638

#### Product version
9.1.1

#### Affected browser(s)
All

karolkolodziej:
Fixed #15864, a regression in the navigator. When `rangeSelector.selected` was set, the navigator had the wrong position after adding a series to an empty chart.

### DIFF
--- a/samples/unit-tests/navigator/navigator/demo.js
+++ b/samples/unit-tests/navigator/navigator/demo.js
@@ -1300,8 +1300,9 @@ QUnit.test('Initiation chart without data but with set range, #15864.', function
             pointInterval: 36e7
         }]
     });
-    assert.ok(
-        chart.xAxis[0].max !== 0,
+    assert.notStrictEqual(
+        chart.xAxis[0].max,
+        0,
         `After adding series to the chart that has set the range,
         the navigator shouldn't stick to min.`
     );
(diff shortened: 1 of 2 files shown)


samples/unit-tests/navigator/navigator/demo.js
```diff
@@ -1289,3 +1289,20 @@ QUnit.test('Scrolling when the range is set, #14742.', function (assert) {
...
+QUnit.test('Initiation chart without data but with set range, #15864.', function (assert) {
+    const chart = Highcharts.stockChart('container', {
+        rangeSelector: {
+            selected: 1
+        },
+        series: [{
+            pointInterval: 36e7
+        }]
+    });
+    assert.ok(
```
TorsteinHonsi, line 1303: Use `assert.notStrictEqual` instead, it prints more information on failure.

TorsteinHonsi (changes requested):

TorsteinHonsi (approved):

PR #15923, opened 2021-06-17
bugfix/9047-boost-primcount-zero

### LINKED ISSUE #9047: Chrome RENDER WARNING: Render count or primcount is 0.
#### Expected behaviour
Using zoom or pan does not create render warnings when boost is enabled.

#### Actual behaviour
Zoom or pan does create render warnings in Chrome when boost is enabled.
screen shot 2018-09-24 at 11 05 50

https://jsfiddle.net/BartBM/k7d8s2mr/

#### Product version
Latest Highcharts, boost and maps version, currently 6.1.3

#### Affected browser(s)
Chome 67, 68 & 69

khlieng:
Fixed #9047, boosted series with null points showed warnings in some browsers.


TorsteinHonsi (approved):
Thanks!

PR #15934, opened 2021-06-18
bugfix/15922-heatmap-border-dataLabels-useHTML

### LINKED ISSUE #15922: Heatmap issue - 8.0.0 versus 9.1.1
#### Expected behaviour
When we were using version 8.0.0, the Risk heat map was coming alright.
but when upgraded to 9.1.1, the cell borders are lost also the numbers (value) which was being shown inside each cell is also lost..

Here is the JSFiddle created with 8.0.0 version

https://jsfiddle.net/sreejithga/rwbd1u8g/1/

which is as per our expectation.

Now if you change the url to point to 9.1.1 (replace 8.0.0 by 9.1.1), we get a map with no borders and no cell values, which is not correct.

Screenshot from 9.1.1 version which is the issue with no borders and and no cell values
-------------------------------------------------------------------------------------------------
image

Kindly help us fix this issue in 9.1.1, so it works as it was working in 8.0.0

https://jsfiddle.net/sreejithga/rwbd1u8g/1/

#### Product version
9.1.1

#### Affected browser(s)
Chrome

khlieng:
Fixed #15922, heatmap data labels with `useHTML` enabled and default formatter did not work, per-point `borderColor` and `borderWidth` did not work.

### DIFF
--- a/samples/unit-tests/series-heatmap/datalabel/demo.js
+++ b/samples/unit-tests/series-heatmap/datalabel/demo.js
@@ -11,7 +11,13 @@ QUnit.test('Check last point visible (#5254)', function (assert) {
         series: [
             {
                 data: [
-                    [0, 0, 10],
+                    {
+                        x: 0,
+                        y: 0,
+                        value: 10,
+                        borderColor: 'red',
+                        borderWidth: 1
+                    },
                     [0, 1, 19],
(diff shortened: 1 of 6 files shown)


TorsteinHonsi (approved):
Thanks!

PR #15942, opened 2021-06-21
Migration: RadialAxis (and HiddenAxis)

bre1470:
- Moved HiddenAxis composition into RadialAxis composition.
- Moved automatic modification to master.
- Improved HiddenAxis composition and compose.
- Improved `const`/`let` in HiddenAxis.
- Removed internal namespace.
- Removed `this` aliases from RadialAxis.


TorsteinHonsi (approved):
Thanks!

PR #15943, opened 2021-06-21
bugfix/15449-exporting-menu-html

### LINKED ISSUE #15449: Cannot add image or icon in menu item since Highcharts 9.0.0
#### Expected behaviour
Image/icon was allowed to be added in the menu item in Highcharts 8.2.0.

#### Actual behaviour
It's no longer supported. HTML tags will be converted to text.

https://jsfiddle.net/5sra4ufe
In HTML, comment line 1&2 and uncomment line 3&4 to compare the behavior in Highcharts 8.2.0.

#### Product version
Highcharts [9.0.0, 9.0.1]

khlieng:
Fixed #15449, HTML in exporting menu item text did not work.


TorsteinHonsi (approved):
Thanks!

PR #15947, opened 2021-06-22
bugfix/15946-marker-symbol-update-animate

### LINKED ISSUE #15946: Specified marker symbol not being animated
#### Expected behaviour
When I specify a marker symbol, e.g. 'circle' or 'triangle', for a given series and the values are updated, the symbol movement should be animated like the rest of the chart. If I do not specify a symbol but let Highcharts choose for me (I am guessing based on series index), the symbols are animated correctly.

#### Actual behaviour
The symbols jump to their new positions without being animated

https://jsfiddle.net/Ly0e2c56/

Use the button to update the data, the markers for the 'custom' series will jump, the markers for 'default' will not.

#### Product version
The version published when using the https://code.highcharts.com/highcharts.js

#### Affected browser(s)
Only tested in Chrome

khlieng:
Fixed #15946, unchanged markers set in options did not animate when updating data through `update`.

### DIFF
--- a/samples/unit-tests/series/update/demo.js
+++ b/samples/unit-tests/series/update/demo.js
@@ -126,7 +126,19 @@ QUnit.test('Series.update', function (assert) {
     assert.strictEqual(
         chart.series[0].points[0].graphic.symbolName,
         'square',
-        'The symbol name should update for all markers (#10870'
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #15948, opened 2021-06-22
Enhancement: Templating in string formats

### LINKED ISSUE #15933: Logic in format strings
# Description of the feature
Support some simple logic in [format strings](https://www.highcharts.com/docs/chart-concepts/labels-and-string-formatting#format-strings). 

Format strings are preferred over formatter callbacks in situations where 
* The configuration is XSS critical because it comes from an unfiltered, untrusted source (like is the case on our own export server), or
* The configuration is relayed from another programming language so function callbacks can not easily be serialized.
* The configuration must be serialized to JSON for storage etc.

For these reasons, it is sometimes required that the format strings support some simple logic. We should model it on well-known and well-tested template languages like Handlebars. Primarily, these features should be considered:
* Conditions. Model it on Handlebars `#if`/`else` syntax.
* Loops. Handlebars have `#each` loops. These can take blocks of template code, unlike our internal Accessibility module's `#each` function. Consider either providing a compliance rewrite or using a different name.
* Simple math. An analysis of our own demos reveals that we often use multiplication and division in our formatters because string formats don't support those operations. For example by converting a fraction to a percentage, or a value to an `{n}k` numeric symbol.
* Array indexing? The accessibility module has this.
* Nested functions? Handlebars has this.

This should be implemented as a pluggable system, allowing contributors to add their own formatting functions. If the core implementation becomes too bulky, consider not adding it to core `highcharts.js`, but as a module script instead.

### Use cases
Conditional tooltip format
```js
pointFormat: '<span class="{#if point.negative}negative{else}positive{/if}">{y:.0f}</span>'
```
Math
```js
labels: {
  format: '{value * 100}%' // Alternatively function called multiply
}
```

Highcharts

----

TorsteinHonsi:
Adding support for [templating](https://www.highcharts.com/docs/chart-concepts/templating), using simple logic in format strings.

### Todo
 - [x] Rename FormatUtilities to Templating
 - [x] Error handling (missing helpers, missing config etc. Silent errors?)
 - [x] `log` function. Show in the docs how to include one-liner `log` or `debugger` functions.
 - [x] Consider `unless` function
 - [x] Conflicts with the a11y module. How to handle `#each`? Alternatives `#for` or `#foreach`.
 - [x] Add `accessibility.point.descriptionFormat` option and use in samples
 - [x] Sub-expressions? A demo could be conversion between Celsius and Fahrenheit, that involves addition and multiplication.
 - [x] Tests
 - [x] Should a11y `#each(` and `#plural(` be deprecated and replaced in language strings?
 - [x] Docs
   - [x] `tooltip.format`
   - [x] A11y point `descriptionFormat`. Two locations in options tree.
   - [x] General templating docs
 - [x] Samples
   - [x] `tooltip.format`
   - [x] Subexpressions: data labels with Celsius/Fahrenheit
   - [x] General
   - [x] Go over the rest of the samples, demos in particular. Some use equality checks or number comparison, consider what to do. Adding native support for [relational helpers](https://gist.github.com/tracend/7522125) would have minimal impact on bundle size.
#### Issues
   - [x] Empty if block fails and passes not else. Space works. Use `format: '{#if point.isNull}{else}{point.name}{/if}` in the `all-areas-as-null` demo. Also consider `#unless` helper if this is a common pattern in maps. Or is it covered by `nullFormat`?
   - [x] Nested if/else fails

Closes #15933 .


docs/chart-concepts/templating.md
```diff
@@ -0,0 +1,116 @@
+Templating
+==========
+
+Highcharts supports templating in format strings. Since v11.1 (2023) the templates support logic, and are generally [recommended](https://www.highcharts.com/docs/chart-concepts/labels-and-string-formatting#format-strings) over [formatter callbacks](https://www.highcharts.com/docs/chart-concepts/labels-and-string-formatting#formatter-callbacks) when the configuration needs to be secure and JSON compatible. The Highcharts templating style is inspired by well-proven languages like Handlebars and Mustache, but is more focused on numeric operations since charting is all about numeric data.
+
+## Expressions
+Expressions in format strings are enclosed by `{single brackets}`. They can be simple variables or constants, or conditional blocks or functions called _helpers_.
+
+**Variables** are inserted directly inside the bracket, for example `"The point value at {point.x} is {point.y}"`.
```
oysteinmoseng, line 9: Should we mention nested property access here?

ts/Extensions/SeriesLabel/SeriesLabel.ts
```diff
@@ -48,7 +48,7 @@ import type SymbolOptions from '../../Core/Renderer/SVG/SymbolOptions';
 import A from '../../Core/Animation/AnimationUtilities.js';
 const { animObject } = A;
 import Chart from '../../Core/Chart/Chart.js';
-import FU from '../../Core/FormatUtilities.js';
+import FU from '../../Core/Templating.js';
```
pawelfus, line 51: Not a blocker: I think it would be better to either use `F` or `Templating` everywhere. FU is just.. misleading ;)

docs/chart-concepts/templating.md
```diff
@@ -0,0 +1,116 @@
...
+* **divide**. Divide the first number with the second. For example `{divide 10 2}` prints 5. Division by zero returns an empty string. [Demo](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/plotoptions/series-datalabels-format-subexpression).
+* **eq**. Returns `true` for loose equality (JavaScript `==`) between the first and second argument. Can be used either as a block helper, `{#eq index 0}First item{/eq}`, or in a subexpression `{#if (eq index 0)}First item{/if}`.
+* **#each**. Iterate over an array of items. The context of each child is given as `{this}` in the block body. Additional variables in the block body are `@index`, `@first` and `@last`. Example `{#each points}{@index}) {name}, {#if @last}and {/if}`. [Demo](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/tooltip/format-shared).
+* **ge**. Greater than or equal, JavaScript `>=`. Doubles as block helper and subexpression.
+* **gt**. Greater than, JavaScript `>`. Doubles as block helper and subexpression.
+* **#if**. Conditional block helper. `{#if point.isNull}The point is null{else}The value is {point.y}{/if}`.
```
pawelfus, line 99: Are multiple conditions supported? Eg `{#if point.isNull || (point.y eq 0)}Zero{else}{point.y}{/if}`

pawelfus (changes requested):
Amazing work! Just a minor nitpicking

Extra question:
Shall we replace **all possible** demos with templates? Like these:
- [compare-value](http://utils.highcharts.local/samples/#view/stock/plotoptions/series-compare-value)
- [compare-percent](http://utils.highcharts.local/samples/#view/stock/plotoptions/series-compare-percent)
- [navigator/inverted](http://utils.highcharts.local/samples/#view/stock/navigator/inverted)
- [axis-setcompare](http://utils.highcharts.local/samples/#view/stock/members/axis-setcompare)

(checked only Stock-related)

bre1470 (approved):
Works great. 👍

TorsteinHonsi:
> Shall we replace all possible demos with templates? Like these:

@pawelfus I replaced the ones in the `demos`, but didn't go over all in samples. But that is something we can do and probably should do at a later stage.

docs/chart-concepts/templating.md
TorsteinHonsi: Currently not. For that we would require custom helpers for `and` and `or`. Your example would be:

```js
Highcharts.Templating.helpers.and = (a, b) => a && b;
```
`{#if (or point.isNull (eq point.y 0))}Zero{else}{point.y}{/if}`

I didn't include that in the built-in helpers because to my surprise, `#if` was hardly needed in the demos. I think it is better to add them later if requested.

pawelfus: Okay, thanks!

Just the background: the idea came from `axis-setcompare`. I don't see any other (stock) demo needing this.

PR #15949, opened 2021-06-22
Stricter typing of options.lang

TorsteinHonsi:
So we can skip any-casting whenever accessing language strings that we know are defined.

### DIFF
--- a/ts/Accessibility/Options/LangOptions.ts
+++ b/ts/Accessibility/Options/LangOptions.ts
@@ -192,7 +192,7 @@ declare global {
     }
 }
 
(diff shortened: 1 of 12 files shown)


ts/Accessibility/Options/LangOptions.ts
```diff
@@ -192,7 +192,7 @@ declare global {
     }
 }
 
-const langOptions: LangOptions = {
+const langOptions: Partial<LangOptions> = {
```
bre1470, line 195: Shouldn't this be deep partial to cover lang sets from extensions?

bre1470 (approved):

PR #15951, opened 2021-06-22
Migration: PiePoint

bre1470:
- Optimized PiePoint.
- Removed this aliases from PiePoint.


ts/Series/Pie/PiePoint.ts
```diff
@@ -123,7 +124,8 @@ class PiePoint extends Point {
      * @private
      */
     public getTranslate(): PiePoint.TranslationAttributes {
-        return this.sliced ? (this.slicedTranslation as any) : {
+        const point = this;
+        return point.sliced ? (point.slicedTranslation as any) : {
```
TorsteinHonsi, line 128: Correct me if I'm wrong, but this type of optimization doesn't have an effect on the file size. Closure compiler changes it back to using `this` directly, and it does so because the gzip algorithm zips the repeating string `this.`, causing the zipped output to be smaller than it would be if we used a local variable alias.

Tested with [Closure Compiler](https://closure-compiler.appspot.com/):
```js
// ==ClosureCompiler==
// @compilation_level SIMPLE_OPTIMIZATIONS
// @output_file_name default.js
// ==/ClosureCompiler==

// ADD YOUR CODE HERE
function hello(name) {
  const w = this;
  console.log(w.innerWidth, w.outerWidth);
}
hello('New user');
```
⬇️ 
```function hello(a){console.log(this.innerWidth,this.outerWidth)}hello("New user");```

File: ts/Series/Pie/PiePoint.ts [hunk 2]
```diff
@@ -181,7 +182,8 @@ class PiePoint extends Point {
      * @private
      */
     public isValid(): boolean {
-        return isNumber(this.y) && this.y >= 0;
+        const point = this;
+        return isNumber(point.y) && point.y >= 0;
```
TorsteinHonsi, line 186: As above

ts/Series/Pie/PiePoint.ts
```diff
@@ -352,18 +354,18 @@ extend(PiePoint.prototype, {
...
             this: PiePoint,
-            labelPosition: PiePositionObject,
+            labelPosition: PiePoint.PositionObject,
             connectorPosition: PiePoint.LabelConnectorPositionObject,
             options: PieDataLabelOptions
         ): SVGPath {
-
-            let touchingSliceAt = connectorPosition.touchingSliceAt,
-                series = this.series,
+            const point = this,
```
TorsteinHonsi, line 361: As above

TorsteinHonsi (changes requested):
Thanks! See the comments for a discussion that applies to nearly all of our code.

ts/Series/Pie/PiePoint.ts
bre1470: I see, I did not know. Back in my days™ it helped also with minification. 😅 

Besides I still do not like to use `this`, because it is not descriptive about its nature and is inconsistent with inline functions. 😬

BTW as soon as it is used in inline functions, the variable stays:
```js
// ==ClosureCompiler==
// @compilation_level SIMPLE_OPTIMIZATIONS
// @languageIn: ECMASCRIPT5_STRICT
// @languageOut: ECMASCRIPT5_STRICT
// @output_file_name default.js
// ==/ClosureCompiler==

// ADD YOUR CODE HERE
function CLSS() {
}
CLSS.prototype.hello = function(name) {
  const clss = this;
  console.log(clss.innerWidth, clss.outerWidth);
  [].forEach(() => { clss.hello('rofl'); });
}

(new CLSS()).hello('New user');
```
⬇️
```js
function CLSS(){}CLSS.prototype.hello=function(b){var a=this;console.log(a.innerWidth,a.outerWidth);[].forEach(function(){a.hello("lol")})}(new CLSS).hello("New user");
```

TorsteinHonsi: Yes, it is not usable with arrow functions, but TypeScript knows that. I think it is cleaner to use `this` directly, usually it is clear what class we are in. And the local variable aliases are not also well named either, for example when we use `wrapper` for an instance of `SVGElement`.

TorsteinHonsi (approved):
Thanks!

PR #15954, opened 2021-06-22
bugfix/15952-annotation-control-point-afterUpdate

### LINKED ISSUE #15952: AfterUpdate callback doesn't fire when editing shape annotation
#### Expected behaviour
The API reference states that the afterUpdate function is an 'Event callback when annotation is updated (e.g. drag and dropped **or resized by control points**).'. So when resizing an **existing** rectangle annotation, the afterUpdate callback should be called when the user releases their mouse after they have finished resizing it with the control point. 

Link to API reference: https://api.highcharts.com/highstock/navigation.annotationsOptions.events.afterUpdate

#### Actual behaviour
After resizing the afterUpdate callback is not called. AfterUpdate is only called when initially creating a rectangle annotation on the screen, but not when resizing an existing annotation using control points.

This fiddle prints "Updated Annotation" when the afterUpdate callback is called.

1. Click the rectangle annotation from the stocktools bar on the left and draw a rectangle anywhere on the chart.
2. Clear the console on the bottom right (should be full of entries from when afterUpdate was called during the creation of the initial annotation)
3. Resize the annotation using control point on the bottom right of the rectangle. 

You'll see that when releasing the rectangle after the resize, "Updated Annotation" is not in the console because afterUpdate is never called. (There might be an entry in the console caused by clicking on the annotation, but not by moving the control point to resize it).

#### Product version
highcharts 9.0.1

khlieng:
Fixed #15952, annotation `afterUpdate` event did not fire when using control point.

### DIFF
--- a/samples/unit-tests/annotations/annotations-events/demo.html
+++ b/samples/unit-tests/annotations/annotations-events/demo.html
@@ -1,5 +1,5 @@
 <script src="https://code.highcharts.com/highcharts.js"></script>
-<script src="https://code.highcharts.com/modules/annotations.js"></script>
+<script src="https://code.highcharts.com/modules/annotations-advanced.js"></script>
 
 <div id="qunit"></div>
 <div id="qunit-fixture"></div>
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (approved):
Looks good to me, pending approval by @pawelfus 👍

pawelfus (approved):
Looks good, thanks!

PR #15959, opened 2021-06-23
Migration: Themes

bre1470:
- Made themes compatible to ES6 tree shaking.
- Moved H.theme assignment to masters files.

### DIFF
--- a/ts/Core/DefaultOptions.ts
+++ b/ts/Core/DefaultOptions.ts
@@ -2931,7 +2931,7 @@ function getOptions(): Options {
  *         Updated options.
  */
 function setOptions(
-    options: Partial<Options>
+    options: DeepPartial<Options>
 ): Options {
 
     // Copy in the default options
--- a/ts/Core/Globals.ts
+++ b/ts/Core/Globals.ts
@@ -225,16 +225,6 @@ namespace Globals {
 
     export let chartCount = 0;
 
-    /**
-     * Theme options that should get applied to the chart. In module mode it
-     * might not be possible to change this property because of read-only
-     * restrictions, instead use {@link Highcharts.setOptions}.
-     *
-     * @name Highcharts.theme
-     * @type {Highcharts.Options}
-     */
-    export let theme: (DeepPartial<Options>|undefined);
-
 }
 
 /* *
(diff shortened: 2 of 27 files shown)


bre1470:
Is this good enough, or do we plan something more advanced like reset for themes, multiple themes, etc?

TorsteinHonsi:
> Is this good enough, or do we plan something more advanced like reset for themes, multiple themes, etc?

@bre1470 It's more than good enough - by design, there is strictly no such feature as "themes" for Highcharts, it is just `setOptions`. Years ago at one point I needed to store the "theme" somewhere (I think it had something to do with the samples), and I put it on `Highcharts.theme`. It is okay to leave it there as long as we don't bloat it.

TorsteinHonsi:
Do we need the extra abstraction to `apply` the theme? Ideally, the theme files should be just `setOptions` with a `DeepPartial<options>`. Also, do we need the master file/source file concept here?

I'm just thinking that something that was meant to be very simple, is turning more complicated. The themes are not really part of Highcharts, they should be considered examples of how to implement a design across multiple charts.

bre1470:
If we do not wrap it in a function call, therefor do instead auto modifications, it will not become ESM compatible. If you load two theme files, you will loose control. We should then clearly communicate, that these are not supposed to be used in production and are just demo code. Best would be to move them to samples and maybe even take the theme property out of the class reference.

oysteinmoseng:
Just a comment that some of the themes are meant to be used in production, e.g. the high contrast themes. We also have plans for making more themes further down the line.

TorsteinHonsi (approved):
Yes I guess all of them are meant to be used in production, or as inspiration for stuff being used in production. 

> If you load two theme files, you will loose control.

In such case, I would expect file A to merge-overwrite the default options, file B to overwrite the product of default options + file A etc. In other words, `setOptions` runs a standard merge operation.

> If we do not wrap it in a function call, therefor do instead auto modifications, it will not become ESM compatible

Yes, good point. 

> Best would be to move them to samples and maybe even take the theme property out of the class reference.

But we would still need to preserve this concept that the theme can be applied to all other samples. 

All in all, it's probably best to continue with this PR. I'm giving it an approval, if @oysteinmoseng has no other comments it's ready to merge.

oysteinmoseng (approved):

PR #15961, opened 2021-06-24
Migration: OrdinalAxis

marvin19:


ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -48,7 +48,7 @@ import '../Navigator.js';
 
 declare module './AxisComposition' {
     interface AxisComposition {
-        ordinal?: OrdinalAxis['ordinal'];
+        ordinal?: OrdinalAxis.Composition['ordinal'];
```
bre1470, line 51: This type should be replaced with OrdinalAxis.Additions.

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -137,10 +116,29 @@ namespace OrdinalAxis {
      *
      * */
 
+    export declare class Composition extends Axis {
+        forceOrdinal?: boolean;
+        isInternal?: boolean;
+        ordinal: OrdinalAxis.Additions;
```
bre1470, line 122: Inside the namespace it is not necessary to use reference the own namespace. So OrdinalAxis.Additions becomes Additions, and OrdinalAxis.Composition becomes Composition.

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -720,15 +927,15 @@ namespace OrdinalAxis {
...
-        ChartClass: typeof Chart,
-        SeriesClass: typeof Series
-    ): void {
+    export function compose<T extends typeof Axis>(
+        AxisClass: T,
+        SeriesClass: typeof Series,
+        ChartClass: typeof Chart
+    ): (typeof Composition&T) {
 
         AxisClass.keepProps.push('ordinal');
```
bre1470, line 936: The compose function needs double-compose protection. Create an non-exportet constant modifiedClasses, taking all class types in an array. Then do conditions around each of the three different class modifications to check whether the provided class argument is already in the array, otherwise modify and put the class in the array.

bre1470 (changes requested):

bre1470 (approved):
Thank you. Looks good! 👍

Issue #15966, opened 2021-06-25
centerInCategory column overlapping when mixing with other series types


Kanatorabo:
#### Expected behaviour
The placement of columns (X-Axis) in charts should not be affected by adding a line series

#### Actual behaviour
Depending on the position in the series Array the placement of columns (X-Axis) is affected by a line series.

#### Live demo with steps to reproduce
Line Series on the end of the array (works fine):
https://jsfiddle.net/Kanatorabo/mLz7wd5s/17/

Line Series in the middle of the array (broken): 
https://jsfiddle.net/Kanatorabo/mLz7wd5s/13/

Line Series on the beginning of the array (broken)
https://jsfiddle.net/Kanatorabo/mLz7wd5s/15/

#### Product version
Highcharts v9.1.2

#### Affected browser(s)
Chrome

Kanatorabo:
pull request https://github.com/highcharts/highcharts/pull/15967

PR #15967, opened 2021-06-25
fixes centerInCategory column overlapping when mixing with other series types #15966

Kanatorabo:


hubertkozik:
Closing due to inactivity.

PR #15980, opened 2021-06-30
Migration: Exporting

marvin19:


samples/highcharts/exporting/multiple-charts/demo.js
```diff
@@ -45,7 +45,7 @@ Highcharts.exportCharts = function (charts, options) {
     options = Highcharts.merge(Highcharts.getOptions().exporting, options);
 
     // Post to export server
-    Highcharts.post(options.url, {
+    Highcharts.Chart.prototype.post(options.url, {
```
bre1470, line 48: We have to continue with exposing `post` function on the `Highcharts` namespace. This goes for all modified samples. So the `post` function of the `Exporting` module has to be assigned in its masters file to `G.post`.

samples/unit-tests/exporting/backgroundcolor/demo.js
```diff
@@ -53,12 +53,12 @@ QUnit.test('Capture POST', function (assert) {
         }
     });
 
-    var originalPost = Highcharts.post;
+    var originalPost = Highcharts.Chart.prototype.post;
```
bre1470, line 56: File should be reverted.

samples/unit-tests/exporting/buttons/demo.js
```diff
@@ -122,21 +122,22 @@ QUnit.test('Export buttons', function (assert) {
 
     assert.strictEqual(button.nodeName, 'g', 'Button is there');
 
-    var originalPost = Highcharts.post;
+    var originalPost = Highcharts.Chart.prototype.post;
```
bre1470, line 125: File should be reverted.

samples/unit-tests/exporting/exportchart/demo.js
```diff
@@ -42,12 +42,12 @@ QUnit.test('Testing exportChart', function (assert) {
         }
     });
 
-    var originalPost = Highcharts.post;
+    var originalPost = Highcharts.Chart.prototype.post;
```
bre1470, line 45: File should be reverted.

samples/unit-tests/exporting/filename/demo.js
```diff
@@ -92,12 +92,12 @@ QUnit.test('POST filename', function (assert) {
         }
     });
 
-    var originalPost = Highcharts.post;
+    var originalPost = Highcharts.Chart.prototype.post;
```
bre1470, line 95: File should be reverted.

samples/unit-tests/exporting/scale/demo.js
```diff
@@ -50,9 +50,9 @@ QUnit.skip('Exported chart scale', function (assert) {
         count = 0;
 
     function testScale(scale) {
-        var originalPost = Highcharts.post;
+        var originalPost = Highcharts.Chart.prototype.post;
```
bre1470, line 53: File should be reverted.

samples/unit-tests/exporting/sourcewidth/demo.js
```diff
@@ -56,9 +56,9 @@ QUnit.skip('Exported chart sourceWidth and sourceHeight', function (assert) {
         }),
         done = assert.async();
 
-    var originalPost = Highcharts.post;
+    var originalPost = Highcharts.Chart.prototype.post;
```
bre1470, line 59: File should be reverted.

samples/unit-tests/exporting/type/demo.js
```diff
@@ -48,17 +48,17 @@ QUnit.test('Test type option with exportChart', function (assert) {
         }),
         postData;
 
-    var originalPost = Highcharts.post;
+    var originalPost = Highcharts.Chart.prototype.post;
```
bre1470, line 51: File should be reverted.

samples/unit-tests/exporting/width/demo.js
```diff
@@ -46,9 +46,9 @@ QUnit.skip('Exported chart width', function (assert) {
         }),
         done = assert.async();
 
-    var originalPost = Highcharts.post;
+    var originalPost = Highcharts.Chart.prototype.post;
```
bre1470, line 49: File should be reverted.

ts/masters/modules/exporting.src.ts
```diff
@@ -10,5 +10,8 @@
  * License: www.highcharts.com/license
  */
 'use strict';
+import Highcharts from '../../Core/Globals.js';
+const G: AnyRecord = Highcharts;
 import '../../Extensions/FullScreen.js';
-import '../../Extensions/Exporting.js';
+import Exporting from '../../Extensions/Exporting/Exporting.js';
+Exporting.compose(G.Chart);
```
bre1470, line 17: Here the `G.post` has to be defined.

bre1470 (changes requested):

samples/unit-tests/exporting/backgroundcolor/demo.js
bre1470: Sorry, I was wrong here.

bre1470 (approved):
Thank you! Looks good. 👍

ts/Core/Renderer/HTML/AST.ts
```diff
@@ -288,17 +288,21 @@ class AST {
...
+     * @param {string} html
+     * Markup string
      */
-    public static setElementHTML(el: Element, html: string): void {
-        el.innerHTML = ''; // Clear previous
-        if (html) {
-            const ast = new AST(html);
-            ast.addToDOM(el);
+    public static setElementHTML(el: (Element|null), html: string): void {
+        if (el) {
```
TorsteinHonsi, line 300: Why make `el` optional? If there is only one case where the element may be undefined/null, we should probably guard it in the call instead. Especially since Exporting.ts is outside the core.

ts/Extensions/Exporting/Exporting.ts
```diff
@@ -0,0 +1,1963 @@
...
+     * Post URL
+     *
+     * @param {object} data
+     * Post data
+     *
+     * @param {Highcharts.Dictionary<string>} [formAttributes]
+     * Additional attributes for the post request
+     */
+    export function post(
+        this: Chart,
```
TorsteinHonsi, line 1629: I don't think the `post` function should be attached to the (already overcrowded) `Chart.prototype`. It is a generic utility function that doesn't have anything to do with a chart, which is also why it previously was attached to `Highcharts`.

Do we have an alternative solution to this? Putting it in Utilities would be the most logic solution, but then it would bloat the core.

TorsteinHonsi (changes requested):

ts/Extensions/Exporting/Exporting.ts
bre1470: Would it be okay to move it to something similar to FormatUtilities, like HttpUtilities?

TorsteinHonsi: Yes I suppose so. I guess the challenge is to make it work with those tests that override Highcharts.post. If we can rethink how we do those, the `post` function could also be local to the Exporting module.

bre1470: The idea of a post event on the chart came to our mind, before we decided with the solution in question.

TorsteinHonsi: The features that are most related to `post` is the `ajax` and `getJSON` functions, both found in `ts/Extensions/Ajax.ts`. The cleanest solution, thematically, may be to add the `post` function there and rename the file to `HttpUtilities`.

TorsteinHonsi (approved):
Looks like we finally found a nice home for `post` :)

PR #15981, opened 2021-06-30
bugfix/1441-dateTime-tickPositions-format

### LINKED ISSUE #1441: tickpositions - datetime xAxis
In case when xAxis has datetime type and we set tickPositions as timestamps, then labels are not dates, but numbers.

http://jsfiddle.net/topatidle/AWeNf/1/ HC
http://jsfiddle.net/AWeNf/3/ HS

khlieng:
Fixed #1441, `datetime` axis ticks did not get date/time formatting applied when `tickPositions` were set.


TorsteinHonsi (changes requested):
Thanks for the PR!

This solution is good, but the result depends on `this.tickInterval` that depends on some other options, plus I think we can do it cheaper.

We already have similar logic for determining the date format for the tooltip header. I think we should try out a bit of refactoring that will allow us to use the same approach in Axis. It is at least worth a shot:
- Move the `getXDateFormat` function from Tooltip to the DateTimeAxis composition. Point.x and `dateTimeLabelFormats` can be passed directly, as we want Axis to use `Axis.options.dateTimeLabelFormats`.
- Move `Tooltip.getDateFormat` to `Time`. 
- In `Tick.addLabel`, if no `tickPositionInfo` is set, run `getXDateFormat`.

With this approach, the basis of the axis label formatting will be the `Axis.closestPointRange`. It would be more precise to compute the closest _tick positions_, but I think this will work for the majority of cases.

TorsteinHonsi:
Nice refactoring! It also makes sense for the other use cases to pull it out of `Tooltip`

TorsteinHonsi:
If the last changes pass I'll try to implement `DeepPartial` on Tooltip so that we can get rid of `dateTimeLabelFormats || {}` and any casting. Then modify this PR then merge it.

TorsteinHonsi:
To do
 - [ ] Merge #15998 into master
 - [ ] Merge master in to this PR
 - [ ] Fix redundant `tooltip.options.dateTimeLabelFormats || {}` pattern

TorsteinHonsi (approved):

PR #15985, opened 2021-07-01
bugfix/15983-update-remove-event

### LINKED ISSUE #15983: Axis events not updated on export
This seems to be related to #6943 that has been fixed recently*. However the same issue described there still occurs when calling `chart.exportChartLocal` programatically. I wonder if maybe this part still uses a faulty implementation.

\* The fix didn't appear in the changelog but I assume it was rolled out, as I saw the corresponding code in the latest npm package

#### Expected behaviour

Existing callbacks on the chart instance should not be invoked when overwritten in the options object passed to `chart.exportChartLocal` 

#### Actual behaviour

Callbacks are invoked

Demo: https://jsfiddle.net/5dwkj2q6/

Steps to reproduce:

1. Open your devtools console
2. Zoom into any area on the chart
3. Notice the console message
4. Click "Export local" button

Expected: Should not log the console message  
Actual: Logs the console message

#### Product version
9.1.2

#### Affected browser(s)
All

khlieng:
Fixed #15983, event did not get removed when updating it to `undefined`.

### DIFF
--- a/samples/unit-tests/axis/events/demo.js
+++ b/samples/unit-tests/axis/events/demo.js
@@ -94,4 +94,16 @@ QUnit.test('Axis events', function (assert) {
     );
 
     unbindClass();
+
+    chart.xAxis[0].update({
+        events: {
+            afterSetExtremes: void 0
+        }
+    });
+    chart.xAxis[0].setExtremes(3, 7);
+    assert.deepEqual(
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #15987, opened 2021-07-01
bugfix/15970-ordinal-annotation 

### LINKED ISSUE #15970: Annotation line with ordinal axis
#### Expected behaviour
Edit the line freely without changing its length.

#### Actual behaviour
Editing the line annotation while using the `ordinal` axis result in shortening the line.

Video: https://www.youtube.com/watch?v=S4dfMMO8c2Y

#### Product version
Highstock JS v9.1.2 (2021-06-16)

#### Affected browser(s)
All

#### Workaround
- Use the 9.1.1 version: http://jsfiddle.net/BlackLabel/p47ocy68/
- Set `xAxis` `ordinal` to false 

_Internal note:_
_Cause of the regression: https://github.com/highcharts/highcharts/pull/15704_

karolkolodziej:
Fixed #15970, annotation positions were not accurate with an ordinal axis.


pawelfus (approved):

raf18seb (approved):

pawellysy (approved):
Good job! 👍

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -622,14 +622,16 @@ namespace OrdinalAxis {
         ): number {
             const ordinal = this,
                 axis = ordinal.axis,
+                min = axis.min as any,
```
TorsteinHonsi, line 625: Avoid `any` casts in new code. `axis.min` is undefined when there are not data points in series related to the axis. What happens with this logic in such case?

TorsteinHonsi (changes requested):

ts/Core/Axis/OrdinalAxis.ts
pawelfus: > What happens with this logic in such case?

Probably the same as we do with non-ordinal axes: NaN's ([demo](https://jsfiddle.net/BlackLabel/xmbc5p38/)) 🤔

ts/Extensions/DataGrouping.ts
```diff
@@ -1036,11 +1036,11 @@ seriesProto.generatePoints = function (): void {
 
 // When all series are processed, calculate the group pixel width and then
 // if this value is different than zero apply groupings.
-addEvent(Axis, 'postProcessData', function (): void {
+addEvent(Axis, 'postProcessData', function (e: any): void {
```
raf18seb, line 1039: Any chance to get rid of this `any`?

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -622,17 +627,30 @@ namespace OrdinalAxis {
...
-                min = axis.min || 0, // #15987
-                firstPointVal = ordinal.positions ? ordinal.positions[0] : 0,
-                secondPointVal = ordinal.positions ? ordinal.positions[1] : 0;
+                firstPointVal = ordinal.positions ? ordinal.positions[0] : 0;
+            let firstPointX = axis.series[0].points &&
+                axis.series[0].points[0] &&
+                axis.series[0].points[0].plotX ||
+                axis.minPixelPadding; // #15987
+
+            // When more series asign to axis, find the smalles one, #15987.
```
raf18seb, line 636: "asign" -> "assigned"
"smalles" -> "smallest"

raf18seb (changes requested):

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -518,6 +520,9 @@ namespace OrdinalAxis {
 
                     series.processData.apply(fakeSeries);
 
+                    // Apply grouping if needed.
+                    fireEvent(axis, 'postProcessData', { fakeAxis: fakeAxis });
```
pawelfus, line 524: - Shouldn't `postProcessData` be fired **after** all series have processed the data?
- I think we shouldn't fire the event on the main axis (I know there's no other callback to `postProcessData` yet, but in the future, it may cause problems). I was hoping we could do something like this: `fireEvent(fakeAxis, 'postProcessData')`, but we can't. How about defining what's in the `postProcessData` callback (line 1039+) as a function, then call the method directly with a different context?

pawelfus (changes requested):

raf18seb (approved):
Thanks!

pawelfus (approved):
Thanks!

TorsteinHonsi (approved):

PR #15988, opened 2021-07-01
bugfix/15986-exit-anchor-chart-update

### LINKED ISSUE #15986: Regarding High charts highcharts-exit-anchor issue.
Hi,

I am shivaraju from OpenText. We have an issue while calling below code.
var clientChart = actuate.report.Chart[ "ClientChart" + chartElement.id ];
chart.update( {
  chart: {
    width: null,
    height: null
  }
} );

after calling this method with width and height as null, chartElement **highcharts-exit-anchor** div has been removed. because of this my custom tab order flow has getting problem. May I know why it is getting deleted after calling that. This is the effective div after calling the method.

 **<div class="highcharts-exit-anchor" tabindex="0" aria-hidden="false"></div>**

Please let me know is there any other way we can get that div back again after calling this.(We are considering based on exit anchor div we are moving focus to next other charts).

khlieng:
Fixed #15986, exit anchor disappeared on `Chart.update`.

### DIFF
--- a/samples/unit-tests/accessibility/accessibility-options/demo.js
+++ b/samples/unit-tests/accessibility/accessibility-options/demo.js
@@ -249,6 +249,12 @@ QUnit.test('Chart description', function (assert) {
             -1,
         'Chart description included in screen reader region'
     );
+
+    chart.update({});
+    assert.ok(
+        document.querySelector('.highcharts-exit-anchor'),
+        '#15986: There should still be an exit anchor after updating'
+    );
 });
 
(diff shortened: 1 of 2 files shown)


oysteinmoseng (approved):
Nice, good investigative work! The marker is (potentially) recreated on render, so `keyboardNavigation.update` won't catch it.

The whole end-of-chart-marker logic is a bit messy and probably needs refactoring, but I have it on my list to change the backwards tab navigation logic anyway, so I can clean it up more then.

TorsteinHonsi (approved):

PR #15990, opened 2021-07-01
bugfix/15235-ast-whitespace

### LINKED ISSUE #15235: Treatment of html in version 9
Hit `Run` repeatedly with the following jsfiddle:

http://jsfiddle.net/ohmkqzub/

This is meant to simulate that the title is made up of individual components, each of which is subject to different formatting based on user settings, and the word separator is also a user option... and a valid and common choice is just a space character (` `).

Now try it with highcharts version 9:

http://jsfiddle.net/tna93r6x/

Why does the space character get wiped out?  Anything to do with the new [filtering feature](https://www.highcharts.com/docs/chart-concepts/labels-and-string-formatting#filtering)?

khlieng:
Fixed #15235, whitespace between html elements got removed.

### DIFF
--- a/samples/unit-tests/svgrenderer/label/demo.js
+++ b/samples/unit-tests/svgrenderer/label/demo.js
@@ -92,6 +92,20 @@ QUnit.test('Left trim (#5261)', function (assert) {
         null,
         'Ending break should have no dy'
     );
+
+    label = ren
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #15998, opened 2021-07-06
TS: DeepPartial tooltip options

TorsteinHonsi:
Explicit tooltip options to avoid picks and any-casts.

### DIFF
--- a/ts/Core/DefaultOptions.ts
+++ b/ts/Core/DefaultOptions.ts
@@ -2700,7 +2700,8 @@ const defaultOptions: Options = {
          * @type {Highcharts.ColorString|Highcharts.GradientColorObject|Highcharts.PatternObject}
          */
         backgroundColor: color(palette.neutralColor3)
-            .setOpacity(0.85).get() as ColorType,
+            // @todo: Disallow undefined as input for colors
+            .setOpacity(0.85).get() as any,
 
         /**
          * The pixel width of the tooltip border.
--- a/ts/Core/DefaultOptions.ts
+++ b/ts/Core/DefaultOptions.ts
@@ -10,6 +10,7 @@
 
 'use strict';
 
+import type ColorType from './Color/ColorType';
 import type { Options } from './Options';
 import type Point from './Series/Point';
 import type Series from './Series/Series';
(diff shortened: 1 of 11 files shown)


ts/Core/DefaultOptions.ts
```diff
@@ -2704,7 +2700,7 @@ const defaultOptions: Options = {
          * @type {Highcharts.ColorString|Highcharts.GradientColorObject|Highcharts.PatternObject}
          */
         backgroundColor: color(palette.neutralColor3)
-            .setOpacity(0.85).get(),
+            .setOpacity(0.85).get() as ColorType,
```
bre1470, line 2703: We avoid forced types, because it disables some type checking. Why is this necessary, if `Color.get` returns `Color` (part of ColorType)?

ts/Core/Tooltip.ts
```diff
@@ -145,7 +145,7 @@ class Tooltip {
 
     public now: Record<string, number> = {};
 
-    public options: TooltipOptions = {};
+    public options: TooltipOptions = {} as any;
```
bre1470, line 148: Can we set it here with the defaults? Should the defaults be moved to the class maybe?

bre1470 (commented):
Some comments. Otherwise it looks good. 👍

ts/Core/DefaultOptions.ts
TorsteinHonsi: `.get()` apparently returns `ColorType|undefined`...

ts/Core/Tooltip.ts
TorsteinHonsi: Yes probably, but I didn't want to make too many structural changes in this PR.

ts/Core/DefaultOptions.ts
bre1470: Maybe we should stop with `undefined` as an input. It does not sound right to me. An alternative for special cases could be something like a `Color.Null` or `Color.NaC` constant instance of `Color`.

bre1470 (approved):

ts/Core/DefaultOptions.ts
TorsteinHonsi: I think disallowing `undefined` as a color input makes most sense. I'll make a `@todo` comment about it, so we don't double the scope of this PR.

PR #16002, opened 2021-07-06
bugfix/12416-scrollablePlotArea-tooltip-update

### LINKED ISSUE #12416: Split tooltip in scrollablePlotArea is not updated on scrolling
#### Expected behaviour
The tooltip should be refreshed when the chart is scrolled.

See https://github.com/highcharts/highcharts/issues/11617 for more information

#### Actual behaviour
The positioning of the tooltip stays according to what it should be before the scrolling, and is not updated during the scroll.

https://jsfiddle.net/Loepjy94/1/
1. Hover one of the points to see the split tooltip labels
2. Scroll down
3. The tooltip positions are not updated, and is wrongly displayed after the scrolling

#### Product version
Next version after v7.2.1

khlieng:
Fixed #12416, tooltip did not update when scrolling scrollable plot area.


LearnGrowAndShare:
Is there a plan on release for this pr?

TorsteinHonsi (approved):

PR #16003, opened 2021-07-06
Migration: ColorAxis

bre1470:
- Migrated ColorAxis to a composition.
- Removed ColorAxis code from Chart class.
- Preferred const over let.
- Fixed small bug in gulp tasks.


TorsteinHonsi (approved):
Thanks!

PR #16005, opened 2021-07-07
bugfix/16004-shared-tooltip-multiple-axes

### LINKED ISSUE #16004: Shared tooltip with multiple yAxis and "top" for positioning of the first yAxis won't work for all markers
#### Expected behaviour
Highcharts should shows the shared tooltip on every marker (if not disabled or configured otherwise).

#### Actual behaviour
In this case, the tooltip for all markers of "BD" and "AD" aren't shown.
I think this is because the first yAxis has `"top":"65%",`.
It works as expected for "AU" and "BU".

https://jsfiddle.net/6nke2uLo/

#### Product version
newest

#### Affected browser(s)
all

khlieng:
Fixed #16004, a regression causing shared tooltip not to show for some axes with multiple differently positioned axes.


raf18seb:
@TorsteinHonsi the users are reporting this issue on different support channels, so it would be great to push it if possible

ts/Core/Tooltip.ts
```diff
@@ -1059,7 +1059,13 @@ class Tooltip {
                 if (
                     chart.polar ||
                     currentSeries.options.clip === false ||
-                    currentSeries.shouldShowTooltip(checkX, checkY)
+                    (
+                        points.length === 1 ?
+                            currentSeries.shouldShowTooltip(checkX, checkY) :
```
TorsteinHonsi, line 1064: Can we optimize this by always going the `points.some` route? In other words, are there any cases where `points.length === 1` and `currentSeries` is not the same as `points[0].series`?

TorsteinHonsi (changes requested):
In addition to the inline comment, please set up a test.

TorsteinHonsi (approved):

PR #16008, opened 2021-07-07
Migration: PlotLineOrBand

bre1470:
- Moved PlotLineOrBand into a subfolder.
- Added composition for PlotLineOrBandAxis.
- Removed PlotLineOrBandAxis composition from AxisLike.
- Optimized const/let.


TorsteinHonsi (approved):
Thanks!

PR #16009, opened 2021-07-07
Bugfix/13217-mobile-pinch-zoom-issue

### LINKED ISSUE #13217: Pinch zoom issue - multiple pane (mobile devices)
#### Expected behaviour

The vertical pinch-zoom should work same as a desktop vertical zoom for multiple-pane: https://jsfiddle.net/BlackLabel/ac9oqpyd/

In the single-pane vertical pinch-zoom works fine on the mobile devices:
https://jsfiddle.net/BlackLabel/qc2pfgv0/4/

#### Actual behaviour

Pinch zoom does not work in the chart with multiple-pane.

Demo: https://jsfiddle.net/BlackLabel/ac9oqpyd/

#### Product version

pawellysy:
Fixed #13217, pinch zoom works correctly on multiple panes on mobile devices.

- [ ] test?

### DIFF
--- a/ts/Core/Pointer.ts
+++ b/ts/Core/Pointer.ts
@@ -70,6 +70,12 @@ declare module './Chart/ChartLike'{
     }
 }
 
+type Scale = {
+    translateX: number;
+    translateY: number;
+    scaleX: number;
+    scaleY: number;
+}
 /* *
  *
  *  Class
@@ -1292,7 +1298,7 @@ class Pointer {
             touchesLength = touches.length,
             lastValidTouch = self.lastValidTouch as any,
             hasZoom = self.hasZoom,
-            transform: Series.PlotBoxObject = {} as any,
+            transform: Partial<Series.PlotBoxObject> = {},
             fireClickEvent = touchesLength === 1 && (
                 (
                     self.inClass(e.target as any, 'highcharts-tracker') &&
@@ -1421,7 +1427,7 @@ class Pointer {
 
             // Scale and translate the groups to provide visual feedback during
             // pinching
-            self.scaleGroups(transform, clip as any, self.pinchedAxes);
+            self.scaleGroups(transform as any, clip as any, self.pinchedAxes);
 
             if (self.res) {
                 self.res = false;
@@ -1864,19 +1870,23 @@ class Pointer {
             axis.series.forEach((series: Series): void => {
                 const seriesAttribs = attribs || series.getPlotBox(), // #1701
                     isX = axis.isXAxis,
-                    attr: any = {};
-                let scale = `scale${isX ? 'X' : 'Y'}`,
-                    translate = `translate${isX ? 'X' : 'Y'}`;
-                attr[scale] = (seriesAttribs as any)[scale];
-                const transformScale = chart.inverted ? 1 / attr[scale] : attr[scale];
-                attr[translate] = (seriesAttribs as any)[translate] + (attribs ? (transformScale * axis.pos) : 0);
+                    attr: Partial<Scale> = {};
+                if (isX) {
+                    attr.scaleX = seriesAttribs.scaleX;
+                    const transformScale = chart.inverted ? 1 / attr.scaleX : attr.scaleX;
+                    attr.translateX = seriesAttribs.translateX + (attribs ? (transformScale * axis.pos) : 0);
+                } else {
+                    attr.scaleY = seriesAttribs.scaleY;
+                    const transformScale = chart.inverted ? 1 / attr.scaleY : attr.scaleY;
+                    attr.translateY = seriesAttribs.translateY + (attribs ? (transformScale * axis.pos) : 0);
+                }
 
                 if (series.xAxis && series.xAxis.zoomEnabled && series.group) {
                     series.group.attr(attr);
                     if (series.markerGroup) {
                         series.markerGroup.attr(attr);
                         series.markerGroup.clip(
-                            clip ? (chart.clipRect as any) : (null as any)
+                            clip ? chart.clipRect : clip
                         );
                     }
                     if (series.dataLabelsGroup) {
@@ -1887,7 +1897,9 @@ class Pointer {
         });
 
         // Clip
-        (chart.clipRect as any).attr(clip || chart.clipBox);
+        if (chart.clipRect) {
+            chart.clipRect.attr(clip || chart.clipBox);
+        }
     }
 
     /**
--- a/ts/Core/Pointer.ts
+++ b/ts/Core/Pointer.ts
@@ -1293,7 +1293,6 @@ class Pointer {
             lastValidTouch = self.lastValidTouch as any,
             hasZoom = self.hasZoom,
             transform: Series.PlotBoxObject = {} as any,
-
             fireClickEvent = touchesLength === 1 && (
                 (
                     self.inClass(e.target as any, 'highcharts-tracker') &&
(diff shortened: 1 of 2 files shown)


pawellysy:
@TorsteinHonsi there seems to be an issue with creating test to this example. In current test suite, when you check the `unit-tests/samples/touch` test `TouchPointer events` this test doesn't pass in utils, but it does in karma (probably because karma doesn't support karma events). Do you have any view on how to solve this problem?

ts/Core/Pointer.ts
```diff
@@ -1830,25 +1844,36 @@ class Pointer {
...
+                attr[scale] = (seriesAttribs as any)[scale];
+                const transformScale = chart.inverted ? 1 / attr[scale] : attr[scale];
+                attr[translate] = (seriesAttribs as any)[translate] + (attribs ? (transformScale * axis.pos) : 0);
+
+                if (series.xAxis && series.xAxis.zoomEnabled && series.group) {
+                    series.group.attr(attr);
+                    if (series.markerGroup) {
+                        series.markerGroup.attr(attr);
+                        series.markerGroup.clip(
+                            clip ? (chart.clipRect as any) : (null as any)
```
karolkolodziej, line 1869: What if the `chart.clipRect` isn't defined? I guess that is the reason for casting to `any`.

karolkolodziej (changes requested):
Nice catch with that!

Only a few TS things require some attention.
Additionally, I would add more comments that it's more clear in the future.
There is also a question about the tests...

ts/Core/Pointer.ts
```diff
@@ -1288,6 +1293,7 @@ class Pointer {
             lastValidTouch = self.lastValidTouch as any,
             hasZoom = self.hasZoom,
             transform: Series.PlotBoxObject = {} as any,
+
```
raf18seb, line 1296: Is this new line necessary?

ts/Core/Pointer.ts
```diff
@@ -1317,6 +1323,15 @@ class Pointer {
             return self.normalize(e);
         });
 
+        self.pinchedAxes = self.pinchedAxes || chart.axes.filter((axis: Axis): boolean => {
+            if ((!self.zoomX && axis.isXAxis) || (!self.zoomY && !axis.isXAxis) || !axis.zoomEnabled) {
```
raf18seb, line 1327: Few lines go out of the 80 characters limit

ts/Core/Pointer.ts
```diff
@@ -1540,8 +1555,7 @@ class Pointer {
         selectionMarker[wh] = selectionWH;
         selectionMarker[xy] = selectionXY;
         transform[scaleKey] = scale;
-        transform['translate' + XY] = (transformScale * plotLeftTop) +
-            (touch0Now - (transformScale * touch0Start));
+        transform['translate' + XY] = (touch0Now - (transformScale * touch0Start));
```
raf18seb, line 1558: Why removed `(transformScale * plotLeftTop) +`? I always ask when some part of code is removed, because there probably is a reason why that code was added in the first time, right?

ts/Core/Pointer.ts
```diff
@@ -1830,25 +1844,36 @@ class Pointer {
...
-                    series.markerGroup.clip(
-                        clip ? (chart.clipRect as any) : (null as any)
-                    );
-                }
-                if (series.dataLabelsGroup) {
-                    series.dataLabelsGroup.attr(seriesAttribs);
+        axes.forEach((axis: Axis): void => {
+            axis.series.forEach((series: Series): void => {
+                const seriesAttribs = attribs || series.getPlotBox(); // #1701
+                const isX = axis.isXAxis;
```
raf18seb, line 1855: In the Highcharts project, we use "1 key word declaration pattern". Use one `const` instead of multiple. The same with `let` below

raf18seb (changes requested):
Good job! Just small code polishing

ts/Core/Pointer.ts
```diff
@@ -1317,6 +1323,15 @@ class Pointer {
             return self.normalize(e);
         });
 
+        self.pinchedAxes = self.pinchedAxes || chart.axes.filter((axis: Axis): boolean => {
+            if ((!self.zoomX && axis.isXAxis) || (!self.zoomY && !axis.isXAxis) || !axis.zoomEnabled) {
+                return false;
+            }
+            if (axis.horiz) {
+                return (touches[0] as any).chartX > axis.left && (touches[0] as any).chartX < axis.left + axis.len;
```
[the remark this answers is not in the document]
pawellysy: The `touches` variable is casted to any in multiple places, and It is hard for me to pinpoint the exact solution to ts issues. The object, that is passed to the function is then mutated by `normalize` function, which adds additional properties. I don't really know, how to fix, so i kept the previous pattern with `any` casting of this var. Do you have some easy fix of this issue? @pawelfus @raf18seb

ts/Core/Pointer.ts
```diff
@@ -1830,25 +1844,36 @@ class Pointer {
...
+        axes.forEach((axis: Axis): void => {
+            axis.series.forEach((series: Series): void => {
+                const seriesAttribs = attribs || series.getPlotBox(); // #1701
+                const isX = axis.isXAxis;
+
+                const attr: any = {};
+                let scale = `scale${isX ? 'X' : 'Y'}`;
+                let translate = `translate${isX ? 'X' : 'Y'}`;
+                attr[scale] = (seriesAttribs as any)[scale];
+                const transformScale = chart.inverted ? 1 / attr[scale] : attr[scale];
```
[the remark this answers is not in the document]
pawellysy: I copied it from the place, where it was initially calculated. I don't know exact formula. It was originally calculated here: https://github.com/highcharts/highcharts/blob/master/ts/Core/Pointer.ts#L1539

ts/Core/Pointer.ts
pawellysy: This was the previous solution. I only edited the moment, when it is invoked. This part of code needs heavy refactoring, and clipping will be handled soon, so I'd suggest to leave it as it is.

ts/Core/Pointer.ts
pawellysy: we want to seperatly calculate this values for each series, so this part was moved to `scaleGroups` function. plotLeft/plotTop is different for each series, depending on the axis, it is added to, so it should be calculated separately.

ts/Core/Pointer.ts
```diff
@@ -1830,29 +1855,46 @@ class Pointer {
...
+                    attr: Partial<Series.PlotBoxObject> = {};
+
+                if (isX) {
+                    attr.scaleX = seriesAttribs.scaleX;
+                    const transformScale = chart.inverted ? 1 / attr.scaleX : attr.scaleX;
+                    attr.translateX = seriesAttribs.translateX + (attribs ? (transformScale * axis.pos) : 0);
+                } else {
+                    attr.scaleY = seriesAttribs.scaleY;
+                    const transformScale = chart.inverted ? 1 / attr.scaleY : attr.scaleY;
+                    attr.translateY = seriesAttribs.translateY + (attribs ? (transformScale * axis.pos) : 0);
```
pawelfus, line 1876: Generating `attr` object could be moved to `axes.forEach()` loop, instead of calculating this in `series.forEach()` loop.

pawelfus (changes requested):
Looks good, just one minor thing (see below)

ts/Core/Pointer.ts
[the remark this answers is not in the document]
karolkolodziej: I haven't thought about the bigger picture here.
I think we can leave that as it is now.

ts/Core/Pointer.ts
karolkolodziej: Yep! Fine ;)

karolkolodziej (approved):

ts/Core/Pointer.ts
```diff
@@ -1830,29 +1855,46 @@ class Pointer {
...
-                    );
+        axes.forEach((axis: Axis): void => {
+            axis.series.forEach((series: Series): void => {
+                const seriesAttribs = attribs || series.getPlotBox(), // #1701
+                    isX = axis.isXAxis,
+                    attr: Partial<Series.PlotBoxObject> = {};
+
+                if (isX) {
+                    attr.scaleX = seriesAttribs.scaleX;
+                    const transformScale = chart.inverted ? 1 / attr.scaleX : attr.scaleX;
```
raf18seb, line 1871: Are you 100% sure that `attr.scaleX !== 0` here?

ts/Core/Pointer.ts
```diff
@@ -1317,6 +1322,27 @@ class Pointer {
...
+                }
+                if (axis.horiz) {
+                    return (
+                        (touches[0] as any).chartX > axis.left &&
+                        (touches[0] as any).chartX < axis.left + axis.len
+                    );
+                }
+                return (
+                    (touches[0] as any).chartY > axis.top &&
+                    (touches[0] as any).chartY < axis.top + axis.len
```
raf18seb, line 1343: Is there a chance to get rid of these `as any`?

raf18seb (changes requested):

ts/Core/Pointer.ts
pawellysy: Yes

ts/Core/Pointer.ts
```diff
@@ -1395,7 +1421,7 @@ class Pointer {
 
             // Scale and translate the groups to provide visual feedback during
             // pinching
-            self.scaleGroups(transform, clip as any);
+            self.scaleGroups(transform as any, clip as any, self.pinchedAxes);
```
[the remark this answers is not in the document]
pawellysy: as I said above, there are a lot of issues with ts, where variables are mutated, and it is hard to pinpoint the exact type, that the variable has.

raf18seb (approved):
Thanks!

pawelfus (changes requested):
On hold

Awaiting refactoring related to https://github.com/highcharts/highcharts/pull/15718#issuecomment-876992475

TorsteinHonsi:
> Awaiting refactoring related to #15718 (comment)

Now that #16341 is merged, we can look into this again @pawellysy

pawelfus:
FYI @pawellysy - added to the next sprint's backlog

pawellysy:
Closed due to the refactoring

PR #16012, opened 2021-07-08
bugfix/16011-annotation-labels-shapes-update

### LINKED ISSUE #16011: responsive not working over multiple size changes
#### Expected behavior
Responsive rules will apply whenever condition is met

#### Actual behavior
Rules are applied the first time, but not subsequent times after the view has reverted back

https://jsfiddle.net/wevu95g2/1/

When graph first appears, it is above responsive threshold, and left axis (which is actually an x axis) and annotations appear as expected.
1. click 'Small' and left axis disappears as specified in responsive rule; the left annotation does not disappear but the right one does (or maybe it is just off-screen)
2. click 'Large' and left axis reappears as expected; annotation does not change
3. click 'Small' and left axis does not disappear
4. further clicking of 'Large', then 'Small' does not show any responsive effect.

#### Product version
master (9.1)

#### Affected browser(s)
Edge, Chrome, Firefox

#### Further notes:
I discovered this because I am trying to get responsive to work in my application, which is in Angular 9 and uses "highcharts-angular": "2.10.0" to integrate highcharts as an Angular component. In the Angular code, the chart HTML is:

```
  <highcharts-chart
      [Highcharts]="Highcharts"
      [options]="chartOptions"
      [oneToOne]="true"
      (chartInstance)="storeChartId($event)"
      style="display: block;"
    >
  </highcharts-chart>
```

I copied chartOptions from the debugger and pasted them into the jsfiddle. That is when I discovered this bug.

But my real problem is that responsive is not working at all in my app--the axis does not disappear no matter how narrow the container is.

/label ~Bug

khlieng:
Fixed #16011, some labels/shapes disappeared when updating annotation with multiple labels/shapes and not passing in any labels/shapes.

### DIFF
--- a/samples/unit-tests/annotations/annotations-dynamic/demo.js
+++ b/samples/unit-tests/annotations/annotations-dynamic/demo.js
@@ -105,6 +105,13 @@ QUnit.test("Annotation's dynamic methods", function (assert) {
                     xAxis: 0,
                     yAxis: 0
                 }
+            }, {
+                point: {
+                    x: 3,
+                    y: 125000,
+                    xAxis: 0,
+                    yAxis: 0
+                }
             }
         ]
     };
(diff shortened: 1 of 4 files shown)


pawelfus (approved):
Thanks!

TorsteinHonsi (approved):

PR #16018, opened 2021-07-09
docs/fix-product-api-links

goransle:
Replaced some links using the `highstock#rangeSelector` format. Also found one broken link to a demo while testing.


TorsteinHonsi (approved):
Thanks!

Issue #15361, opened 2021-03-19
Cumulative sum


pawelfus:
# Description of the feature
Similar to `compare` option. The cumulative sum should render a sum according to the first point within the visible range:

```
SUM[0] = P[0]
SUM[i] = SUM[i - 1] + P[i]
```

# Library related to the feature
Highcharts Stock

# Proof of Concept/Live example for the feature
n/a

# To do's:
- [x] - new series/options implementation
- [x] - update [SCSS](https://github.com/highcharts/highcharts/blob/master/css/highcharts.scss) (in case of new styled-mode options) and the [list](https://github.com/highcharts/highcharts/blob/master/docs/chart-design-and-style/style-by-css.md) (if necessary)
~- [x] - add to [Stock Tools'](https://www.highcharts.com/demo/stock/stock-tools-gui) button list of series types~
- [x] - official [demo](https://www.highcharts.com/demo/stock) (this will cover visual test too)
~- [ ] - official [general documentation](https://www.highcharts.com/docs/stock/candlestick-chart)~ - separte ticket: #16060
- [x] - update [API](https://api.highcharts.com/highstock/) (series/options description)

----

**You can vote for this feature by adding a thumbs-up reaction to this post.**

laventnc:
I posted the original feature recommendation, very happy to see it's taking off. Wanted to share a bit of info I've gathered through the different iterations, [here's a link to my current version](https://jsfiddle.net/nlaventis/jwea6sd4/2/).

- I ended up needing to store the original series returns in order to efficiently respond to date-range changes
- This creates a lot of issues in terms of updating the series' actual data
  - The relationship between series.update and series.setData caused issues as the difference between data updating when the date range changes & when the actual series' data is updated
  - _I believe the best solution would be to treat it like an indicator series, where you define the hidden series as raw return data and link a cumSum type series to it_
 - Adding series to a chart became much more complex as you need to calc the cumSum over the active date range
 - Capturing xAxis.setExtremes to trigger a re-calc of the cumSum is satisfactory; however, I've noticed that this event fires more often than expected (this is what I'm currently using).
 - The relationship with the navigator has been a particular issue... You either disable navigator.adaptToUpdatedData or you end up with a changing, often misleading, series visible
 - Series with different sets of dates (x values) forced me to apply the following logic in `CumSum.setExtremes( evt )`:
   1. find the index of each series' dates that is closest to the event's min/max property
[...]

PR #16020, opened 2021-07-09
feature/15361-stock-cumulative-sum

raf18seb:
Added new feature [Cumulative Sum](https://www.highcharts.com/docs/stock/cumulative-sum). See #15361


ts/Core/Chart/StockChart.ts
```diff
@@ -941,11 +1026,18 @@ Series.prototype.init = function (): void {
...
+    compare?: string,
+    redraw?: boolean
+): void {
     this.initCompare(compare);
 
     // Survive to export, #5485
-    this.userOptions.compare = compare;
+    this.options.compare = this.userOptions.compare = compare;
+
+    if (pick(redraw, true)) {
```
karolkolodziej, line 1038: Isn't it always `true`?

ts/Core/Chart/StockChart.ts
```diff
@@ -1120,35 +1202,141 @@ Axis.prototype.setCompare = function (
...
+Series.prototype.setCumulative = function (
+    cumulative?: boolean,
+    redraw?: boolean
+): void {
+    this.initCumulative(cumulative);
+
+    // Survive to export, #5485
+    this.options.cumulative = this.userOptions.cumulative = cumulative;
+
+    if (pick(redraw, true)) {
```
karolkolodziej, line 1234: Isn't it always `true`?

ts/Core/Chart/StockChart.ts
```diff
@@ -1120,35 +1202,141 @@ Axis.prototype.setCompare = function (
...
+
+        return sum;
+    });
+
+    return [cumulativeDataMin, cumulativeDataMax];
+};
+
+/**
+ * Highcharts Stock only. Set the cumulative mode on all series
+ * belonging to an Y axis after render time.
```
karolkolodziej, line 1309: What do you mean by `after render time`?

karolkolodziej (changes requested):

ts/Core/Chart/StockChart.ts
raf18seb: `H.pick(false, true)` returns `false`

ts/Core/Chart/StockChart.ts
raf18seb: `H.pick(false, true)` returns `false`

pawellysy (approved):
Good work!

ts/Core/Chart/StockChart.ts
```diff
@@ -910,13 +919,13 @@ addEvent(Axis, 'afterDrawCrosshair', function (
...
+ *  Start shared compare and cumulative logic                                 *
  * ************************************************************************** */
 
 /**
- * Extend series.init by adding a method to modify the y value used for plotting
- * on the y axis. This method is called both from the axis when finding dataMin
- * and dataMax, and from the series.translate method.
+ * Extend series.init by adding a methods to modify the y values used
+ * for plotting on the y axis. This method is called both from the axis
+ * when finding dataMin and dataMax, and from the series.translate method.
```
pawelfus, line 928: > This method is called both from the axis
> * when finding dataMin and dataMax
 
 This is a little bit off. For compare it's correct, but not for `cumulativeSum`

pawelfus (changes requested):

samples/stock/members/axis-setcumulative/demo.js
```diff
@@ -0,0 +1,32 @@
+Highcharts.stockChart('container', {
+
+    rangeSelector: {
+        selected: 4
+    },
+
+    tooltip: {
+        pointFormat: '<span style="color:{series.color}">{series.name}</span>: <b>{point.y}</b> ({point.cumulativeSum})<br/>',
```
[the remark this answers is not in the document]
raf18seb: Q1 answer: it has never worked with compare too. I should fix it for both cumulative and compare then

Q2: sure 👍

ts/Core/Chart/StockChart.ts
raf18seb: > What do you mean by after render time?
We should ask the author of the compare :) I think he meant that we don't have to define compare on the chart's init in the chart's options, but we can turn it on later - "after the chart has been initially rendered".

ts/Core/Chart/StockChart.ts
```diff
@@ -1120,35 +1202,141 @@ Axis.prototype.setCompare = function (
...
+ * @sample stock/members/axis-setcumulative/
+ *         Set compoare
+ *
+ * @function Highcharts.Axis#setCumulative
+ *
+ * @param {boolean} [redraw=true]
+ *        Whether to redraw the chart or to wait for a later call to
+ *        {@link Chart#redraw}.
+ */
+Axis.prototype.setCumulative = function (
```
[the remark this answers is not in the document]
raf18seb: Yes we can. I see 2 disadvantages of it:
- a little bit more difficult API for users that are using only one of compare/cumulative
- no backward compatibility for compare

What's your final decision? Leave it as it is or optimize the code?

pawelfus (approved):
Approved, but I'm thinking about refactoring the whole "compare"/"cumulative sum" modes. This extension is still overwriting some of the prototype methods (`Series.init`, `Series.processData` etc) and adding extra ones (`Series.initCompare`, `Series.setCompare` etc, however `Axis.setCompare()` should remain, as this is part of the official API). 

Instead, we could:
- use events instead of overwriting
- move to a separate file, make this an extension, e.g. `Series.dataModify = { init, set, others... }`

So we could use it later in tree shaking (extra ~400lines, not every Stock user needs these modes).

Refactoring of course would be in a separate PR/branch

What do you think @raf18seb @TorsteinHonsi @bre1470

bre1470:
I like the second idea, a composition, it has the least impact on the axis and series itself. The official functions could be a link to the composition instance. The naming though needs improvement, `dataModify` is to unspecific. As this is a regular composition, it should go directly under the `ts/Series` folder.

ts/Core/Chart/StockChart.ts
```diff
@@ -1192,14 +1245,7 @@ Axis.prototype.setCompare = function (
...
 ): void {
-    if (!this.isXAxis) {
-        this.series.forEach(function (series): void {
-            series.setCompare(compare);
-        });
-        if (pick(redraw, true)) {
-            this.chart.redraw();
-        }
-    }
+    this.setModifier('compare', compare, redraw);
```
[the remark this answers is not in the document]
bre1470: It depends. You should consider a regular object with `as const` cast.
- Example: https://www.typescriptlang.org/play?ts=3.9.7#code/KYOwrgtgBAKgygUXNA3gKCpqID2AaDLCAQwE8AjYArKU4AZzQF800BjHEegFygClEyKAF4o6GrgBcUAAzUiZStICM8zHXrSATMyjF6UDl24BuNEA
- Docs: https://www.typescriptlang.org/docs/handbook/enums.html

ts/Core/Chart/StockChart.ts
```diff
@@ -926,10 +941,124 @@ Series.prototype.init = function (): void {
...
+    }
+};
+
+/**
+ * Shared code for the axis.setCompare() and the axis.setCumulative() methods.
+ * @private
+ */
+Axis.prototype.setModifier = function (
+    mode: 'compare'|'cumulative',
+    modeState?: boolean|string,
```
karolkolodziej, line 959: I think there is no handler for the case when `modeState: null` 

`typeof modeState === 'object'`

ts/Core/Chart/StockChart.ts
```diff
@@ -926,10 +941,124 @@ Series.prototype.init = function (): void {
...
+ *
+ * @param {string} pointFormat
+ */
+Point.prototype.tooltipFormatter = function (pointFormat: string): string {
+    const point: any = this,
+        { numberFormatter } = point.series.chart,
+        replace = function (value: string): void {
+            pointFormat = pointFormat.replace(
+                '{point.' + value + '}',
+                ((point[value] as any) > 0 && value === 'change' ? '+' : '') +
```
karolkolodziej, line 1035: From what I see, casting to `any` isn't necessary.

ts/Core/Chart/StockChart.ts
```diff
@@ -1110,45 +1245,157 @@ Axis.prototype.setCompare = function (
...
+        this.modifyValue = function (
+            this: Series,
+            value?: number|null,
+            index?: number
+        ): (number|undefined) {
+            if (value === null) {
+                value = 0;
+            }
+
+            if (typeof value !== 'undefined' && typeof index !== 'undefined') {
```
karolkolodziej, line 1308: `defined(value) && defined(index)`

ts/Core/Chart/StockChart.ts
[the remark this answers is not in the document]
karolkolodziej: Both `enum` and casting `as const` sound better than having this as a hardcoded string.
But casting `as const`  looks better after compiling that is true

karolkolodziej (changes requested):

TorsteinHonsi (changes requested):
Supporting a refactoring as suggested by Pawel. In addition:
- Add docs
- My name is in the author list in the samples. Author should be removed or updated to the correct name.

pawelfus:
> * Add docs

Related PR: https://github.com/highcharts/highcharts/pull/16187

TorsteinHonsi (approved):
Thanks! Pending approval by @karolkolodziej

raf18seb:
@TorsteinHonsi sorry for the number of merges, I had some troubles with a project. This PR still needs some polishing.

TorsteinHonsi (changes requested):
Pending more polishing as stated above.

pawelfus (approved):
Thanks! 🎉

samples/stock/demo/cumulative-sum/demo.details
```diff
@@ -0,0 +1,12 @@
+---
+ name: Highcharts Stock Cumulative Sum Demo
+ authors:
+   - Rafal Sebestjanski
+ js_wrap: b
+ new_until: 2021-12-31
```
karolkolodziej, line 6: Is this a proper date? @pawelfus

karolkolodziej (approved):

samples/stock/demo/cumulative-sum/demo.details
pawelfus: Yes, it is 👍

TorsteinHonsi (approved):

PR #16028, opened 2021-07-12
Fixed #15229, avoid duplicate a11y region labels.

### LINKED ISSUE #15229: [a11y] Duplicate region labels with multiple charts
The landmark labels (such as "Chart screen reader information") are the same for all charts, meaning that with multiple charts on the page, there will be duplicate labels.

The simple solution is to add the chart title to these labels by default. Example: https://jsfiddle.net/wktusdrL/

oysteinmoseng:
Fixed #15229, avoid duplicate a11y region labels.
___
We now add the chart title to these by default.


TorsteinHonsi (approved):
Thanks!

PR #16029, opened 2021-07-12
Migration: Chart renderer

bre1470:
SVGRenderer will be the default fallback and imported mandatory together with Chart.

~Extended `chart.renderer` option with the class constructor type of `SVGRenderer`.~

~Without something like this, `Chart` class will never become full ESM compatible (async) as it expects registered renderers to be available during construction.~

~An alternative could be a fourth argument in the constructor taking the renderer class.~


bre1470:
In the sync meeting today, we decided to keep the options untouched. Instead we go with a general fallback to SVGRenderer directly in Chart, so no additional import or the Renderer class is needed.

TorsteinHonsi (approved):
Thanks!

Issue #8667, opened 2018-07-24
Modulize Highchart? Tree shaking capability.


envil:
I'm using Highcharts and Highstocks with Angular in my application. The functionality is good, however it's impossible to optimize the bundle size of Highcharts:
Take look at this screenshot of my application bundle information:
<img width="961" alt="screen shot 2018-07-24 at 12 07 05 pm" src="https://user-images.githubusercontent.com/775364/43116583-12e92d9a-8f3b-11e8-8dc4-99d943a39a4c.png">
You can see that Highcharts ate up to nearly 500kb (dark blue outline), even bigger than a chunky framework like Angular (red outline). Even if one uses only one chart in Highcharts, they have to carry on all of the Highcharts code, which is unacceptably unoptimized.

The solution I know here is to modulize Highcharts so that one can only import what they need. Also with new Webpack Tree Shaking, the bundle size will be reduced vastly.
I know this cannot be fixed soon but are you guys thinking about it?

sebastianbochan:
Hi @envil,
At the begining I advise you to load only `highstock.js` which included all Highcharts' features. You can also build your custom file by our tool here:
-  https://www.highcharts.com/download

envil:
Thank you for your responses.
I'm looking at the custom build feature. It gives me a `highchart-custom.src.js`which is not convenient to use at all since it's not a node_modules, so I cannot import it to my Angular application in a regular way. And all the type definitions for the typescript are gone.

envil:
Also what I mean is ES module, completely independent and can be imported to other ES modules. See how `lodash` and `date-fns` have defeated `underscore` and `momentjs`, one of the most important factor is the modularization capability like this.

TorsteinHonsi:
@jon-a-nygaard FYI

GeorgeKnap:
I am running into this as well.
I tried to go around with  `import * as Highstock from 'highcharts/highstock';` but it has no effect whatsoever.

jon-a-nygaard:
Hi, a while back we started publishing our parts files as ES modules which means you can do `import Highcharts from 'highcharts/es-modules/masters/highcharts.src.js'`. This may help some on the issue.

We are also currently working on making these parts files more standalone, which should improve tree shaking capabilities. It is a bit difficult at the moment to give a good ETA for when this will be published, but I will keep you updated as soon as I know some more.

ymanvieu:
hi,

[...]

PR #16030, opened 2021-07-12
Migration: Phase 1 clean-up

bre1470:
Improved ES module architecture with support for tree shaking, see #8667.


TorsteinHonsi (approved):
Thanks!

PR #16039, opened 2021-07-14
bugfix/16036-vbp-indicator-volumeDivision-graphic

### LINKED ISSUE #16036: Indicator VbP ("Volume by Price") rescale and positioning not working if volumeDivision enabled
#### Expected behaviour
Add an OHLC series, a volume series and a "vbp" indicator with parameter volumeDivision: { enabled: true }.
Then scale the chart, e.g. by using xAxis.setExtremes().
Expect both (the positive and negative colored) "vbp" bars to be correctly scaled and positioned into their respective zones.

#### Actual behaviour
The right part, negative/red "vbp" bars is updated correctly by VBPIndicator.prototype.translate.
But the left part, positive/green "vbp" bars are not translated and stay in their initial scale and position.

With volumeDivision disabled all works as expected.

The issue seems to lay in how VBPIndicator.prototype.drawPoints adds the two required "points".
Looks like the first added "point" (the positive/green) is overwritten by the second "point" (negative/red).
VBPIndicator.prototype.translate only finds the last added "points" (the negative/red ones) and is not able to translate the first "points" (positive/green ones).

Issue can be reproduces in the sample linked for the indicator:
https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/stock/indicators/volume-by-price

Simply change the period several times and you can see that the green bar points are not correctly resized and positioned any more.

#### Product version
Highstock 9.1.0 and 9.1.2

#### Affected browser(s)
Should not depend on browser, but tested in Chrome 91.0.4472.106 and Firefox Developer 86.0b9

khlieng:
Fixed #16036, VBP indicator graphic did not update correctly with `volumeDivision` enabled.

### DIFF
--- a/samples/unit-tests/indicator-volume-by-price/recalculations/demo.js
+++ b/samples/unit-tests/indicator-volume-by-price/recalculations/demo.js
@@ -195,6 +195,13 @@ QUnit.test('Test algorithm on data updates.', function (assert) {
         'volumeDataArray is correct after point remove on the base and the volume series.'
     );
 
+    const negativeGraphic = indicator.points[0].negativeGraphic;
+    indicator.points[0].destroy();
+    assert.notOk(
+        negativeGraphic.element,
+        '#16036: negativeGraphic should have been destroyed'
+    );
+
     indicator.remove();
     assert.ok(
         chart.series.indexOf(indicator) === -1,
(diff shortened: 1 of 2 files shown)


karolkolodziej (approved):
Thanks!

pawelfus (approved):
Thanks!

TorsteinHonsi (approved):

PR #16046, opened 2021-07-14
bugfix/16042-xrange-gaps

### LINKED ISSUE #16042: Random white gaps in X-Range chart between bars
#### Expected behaviour

There should be no gaps between bars when there are no gaps in the series data.

#### Actual behaviour

Random white gaps between bars are rendered. When zooming in the gaps appear at different places.

JSFiddle: https://jsfiddle.net/6gay3mcs/2/

image

After zooming:
image

#### Product version

Highcharts JS v9.1.2

#### Affected browser(s)

The screenshots above were taken in Chrome (Version 91.0.4472.114).
When using Firefox (89.0.2) the JSFiddle example initially does not show gaps but they still appear after zooming.

khlieng:
Fixed #16042, 1px gaps showed between some xrange points when there was no gaps in the data.

### DIFF
--- a/ts/Series/XRange/XRangeSeries.ts
+++ b/ts/Series/XRange/XRangeSeries.ts
@@ -391,10 +391,13 @@ class XRangeSeries extends ColumnSeries {
             );
         }
 
+        const x = Math.floor(Math.min(plotX, plotX2)) + crisper;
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (approved):
Thanks!

PR #16049, opened 2021-07-14
Migration: Color

bre1470:
- Improved const/let.
- Replaced `undefined` color with Color.None.
- Removed any casts in Color class.


bre1470:
Visual colors of two demos have been fixed as a side effect.

TorsteinHonsi (approved):
Thanks!

PR #16051, opened 2021-07-14
bugfix/16044-draggable-guide-size

### LINKED ISSUE #16044: draggable-point guide box has wrong dimentions when not whole point is visible on the plotArea
#### Expected behaviour
Point guide box have actual dimensions while dragging "example chart bar" from right to left

#### Actual behaviour
Point guide box have only visible and not actual dimensions   while dragging "example chart bar" from right to left

#### Product version
9.1.2

#### Affected browser(s)
all browsers

khlieng:
Fixed #16044, draggable point guide box was not big enough when the point was partially outside the end of the axis.

### DIFF
--- a/ts/Extensions/DraggablePoints.ts
+++ b/ts/Extensions/DraggablePoints.ts
@@ -50,6 +50,7 @@ import U from '../Core/Utilities.js';
 const {
     addEvent,
     clamp,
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (approved):
Thanks!

PR #16054, opened 2021-07-16
bugfix/16045-tooltip-hide-opacity-hover-button

### LINKED ISSUE #16045: Tooltip data not visible if we add opacity to tooltip  tyle object and hover over zoom out button
If we add opacity to tooltip, zoom it in and position of the chart bar is near the reset zoom button

https://i.ibb.co/ysvJ52t/example.jpg

hen hover over chart bar so the tooltip is being shown and then again hover over reset zoom button text from tooltip will disappear. If we remove the opacity from tooltip style object this will not happen.

#### Expected behaviour

Tooltip data is visible

#### Actual behaviour
Tooltip data is not visible

#### Product version
9.1.2

### LINKED ISSUE #16140: Tooltips has visibility: visible even though it is hidden
#### Expected behaviour
Tooltips that are not visible should have `visibility: hidden` as a style.

#### Actual behaviour
When using `useHTML: true` in conjunction with `hideDelay: 0` tooltips will keep their visibility set to `visibility: visible` even after vanishing via `opacity: 0`.
In our project this lead to an invisible tooltip covering part of our chart, which - as a result - did not receive ClickEvents anymore. A workaround for us was setting `hideDelay: 1` (any value above 0 works), but this should probably be fixed and in both cases `visibility` should be set to `hidden`.

https://jsfiddle.net/bg8x3tfs/6/

#### Product version
9.1.2

#### Affected browser(s)
All of them would be my guess.

#### Note
I tried finding the corresponding code in the highcharts project, but could not determine where the visibility is set. If somebody points me in the right direction I will open a PR.

khlieng:
Fixed #16045, #16140, tooltip with `style.opacity` set did not hide properly when hovering button.

- [ ] Find out what to do about the failing tests

Closes #16140.


TorsteinHonsi:
Closing as requested in https://github.com/highcharts/highcharts/pull/16891

PR #16057, opened 2021-07-16
bugfix/16053-colorAxis-show-hide-legend

### LINKED ISSUE #16053: Heatmap Enable Legend Not Working Dynamically
#### Expected behaviour
Dynamically show and hide the legend by toggling between `enabled: true` and `enabled: false`

#### Actual behaviour

Dynamically showing/hiding the legend by toggling between `enabled: true` and `enabled: false` works 
great on all other Highcharts charts. However, on a heatmap, even if the value for the legend is set to `enabled: true` the chart 
will not _dynamically_ show the legend. It will not expand and show the legend like all other charts.

To reproduce on a heat map: 

* Toggle the `Update Legend` button: https://jsfiddle.net/yesg2xc1/
* Notice it will dynamically hide the legend, but it won't dynamically show it again. You can check the console to verify that the enabled option is being toggled.
* The expected behavior with the same implementation on a bar chart:  https://jsfiddle.net/x8fkdoyj/1/

khlieng:
Fixed #16053, color axis disappeared after toggling legend.

### DIFF
--- a/samples/unit-tests/coloraxis/coloraxis-update/demo.js
+++ b/samples/unit-tests/coloraxis/coloraxis-update/demo.js
@@ -57,6 +57,26 @@ QUnit.test('Color axis updates', function (assert) {
         undefined,
         'No extra undefined properties after update'
     );
+
+    chart.update({
+        legend: {
+            enabled: false
+        }
+    });
+    assert.notOk(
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #16073, opened 2021-07-19
bugfix/15334-waterfall-negative-dataLabels

### LINKED ISSUE #15334: Waterfall dataLabel not always properly aligned when inside: false and negative values
#### Expected behaviour
Align data properly for `inside: false` and negative values.

#### Actual behaviour
When in the waterfall series `dataLabel.inside: false` and the sum has negative values, data labels are rendered incorrectly.
Adding `max` to `yAxis` makes some of the labels (sum) disapear. 

https://jsfiddle.net/BlackLabel/zarg8cfh/

#### Product version

#### Affected browser(s)
All

khlieng:
Fixed #15334, some waterfall data labels did not show for points below `threshold`.

### DIFF
--- a/samples/unit-tests/series-waterfall/waterfall/demo.js
+++ b/samples/unit-tests/series-waterfall/waterfall/demo.js
@@ -74,7 +74,10 @@ QUnit.test('General waterfall tests', function (assert) {
                     },
                     data: [-5.8, -5.3, -4, -2.7, -6.9]
                 }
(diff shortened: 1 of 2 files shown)


karolkolodziej (approved):
Thanks!

KacperMadej (approved):

TorsteinHonsi (approved):

PR #16088, opened 2021-07-21
bugfix/16087-stocktools-ie

### LINKED ISSUE #16087: Stock-tools error in IE11
#### Behaviour
Error in:
```
// if the button is disabled, don't do anything
                if (buttonWrapperClass.indexOf('highcharts-disabled-btn') > -1) {
```
in stock-tools.js
The error message is about a fail with `indexOf` on undefined.

#### Product version
9.1.0+

khlieng:
Fixed #16087, stock tools threw in IE.

### DIFF
--- a/ts/Extensions/Annotations/NavigationBindings.ts
+++ b/ts/Extensions/Annotations/NavigationBindings.ts
@@ -1602,27 +1602,30 @@ addEvent(Chart, 'render', function (): void {
 
                     if (buttonNode) {
                         if (value.noDataState === 'normal') {
-                            buttonNode.forEach(function (button): void {
+                            for (let i = 0; i < buttonNode.length; i++) {
+                                const button = buttonNode[i];
                                 // If button has noDataState: 'normal',
                                 // and has disabledClassName,
                                 // remove this className.
                                 if (button.className.indexOf(disabledClassName) !== -1) {
                                     button.classList.remove(disabledClassName);
                                 }
-                            });
+                            }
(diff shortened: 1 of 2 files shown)


pawelfus (approved):
Thanks!

ts/Extensions/Annotations/NavigationBindings.ts
```diff
@@ -1602,27 +1602,30 @@ addEvent(Chart, 'render', function (): void {
...
+                                const button = buttonNode[i];
                                 if (button.className.indexOf(disabledClassName) === -1) {
                                     button.className += ' ' + disabledClassName;
                                 }
-                            });
+                            }
                         } else {
-                            buttonNode.forEach(function (button): void {
+                            for (let i = 0; i < buttonNode.length; i++) {
+                                const button = buttonNode[i];
```
raf18seb, line 1623: What about simply convert a nodeList to an array like this?
```js
const buttonNode = Array.prototype.slice.call(
    chart.navigationBindings.container[0].querySelectorAll('.' + key)
);
```

raf18seb (commented):
Just a small suggestion

ts/Extensions/Annotations/NavigationBindings.ts
khlieng: I moved the for loop outside the ifs instead.

TorsteinHonsi (approved):
Thanks!

Issue #16019, opened 2021-07-09
Stock Tools: Create filter/searchbox for technical indicators popup


pawelfus:
# Description of the feature
The number of technical indicators keeps growing and it's getting harder and harder to find the one we want to add to the chart. Implement a simple filter box that will filter out indicators. 

# Library related to the feature
Highcharts stock

# Proof of Concept/Live example for the feature
<img src="https://user-images.githubusercontent.com/1453926/125079321-c13a4c80-e0c3-11eb-8546-12a0a3dc9f59.png" width="250">

----

**You can vote for this feature by adding a thumbs-up reaction to this post.**

PR #16089, opened 2021-07-21
enhancement/16019-indicators-searchbox 

karolkolodziej:
Enhancement #16019, added an indicator search box into the indicator popup window.

To do:
- [ ] `handleInputChange`
- [ ] `handleResetInput`
- [ ] tests
- [ ] CSS
- [ ] list of indicators in order

### DIFF
--- a/ts/Extensions/Annotations/Popup.ts
+++ b/ts/Extensions/Annotations/Popup.ts
@@ -66,7 +66,12 @@ declare global {
             ): HTMLDOMElement;
             public addCloseBtn(): void;
             public addColsContainer(container: HTMLDOMElement): Record<string, HTMLDOMElement>;
-            public addInput(option: string, type: string, parentDiv: HTMLDOMElement, value: string): void;
+            public addInput(
+                option: string,
+                type: string,
+                parentDiv: HTMLDOMElement,
+                inputAttributes: InputAttributes
+            ): void;
             public closePopup(): void;
             public deselectAll(): void;
             public getFields(parentDiv: HTMLDOMElement, type: string): PopupFieldsObject;
@@ -125,6 +130,7 @@ declare global {
                 rhsColWrapper: HTMLDOMElement
             ): void;
             addIndicatorList(this: Popup, chart: AnnotationChart, parentDiv: HTMLDOMElement, listType: string): void;
+            addSearchBox(this: Popup, chart: AnnotationChart, parentDiv: HTMLDOMElement): void;
             addParamInputs(
                 this: Popup,
                 chart: AnnotationChart,
@@ -151,6 +157,10 @@ declare global {
             init(this: Popup, chart: AnnotationChart): void;
             selectTab(this: Popup, tab: Element, index: number): void;
             switchTabs(this: Popup, disableTab: number): void;
+        }
+        interface InputAttributes {
+            value: string|number;
+            type: string;
         }
     }
 }
@@ -290,7 +300,11 @@ H.Popup.prototype = {
      * Default value of input i.e period value is 14, extracted from
      * defaultOptions (ADD mode) or series options (EDIT mode)
      */
-    addInput: function (option: string, type: string, parentDiv: HTMLDOMElement, value: string): void {
+    addInput: function (
+        option: string, type: string,
+        parentDiv: HTMLDOMElement,
+        inputAttributes: Highcharts.InputAttributes
+    ): void {
         const optionParamList = option.split('.'),
             optionName = optionParamList[optionParamList.length - 1],
             lang = this.lang,
@@ -310,14 +324,13 @@ H.Popup.prototype = {
         }
 
         // add input
-        if (value !== '') {
-
+        if (inputAttributes) {
             createElement(
                 INPUT,
                 {
                     name: inputName,
-                    value: value[0],
-                    type: value[1],
+                    value: inputAttributes.value as any,
+                    type: inputAttributes.type,
                     className: PREFIX + 'popup-field'
                 },
                 void 0,
@@ -796,13 +809,17 @@ H.Popup.prototype = {
 
             // ADD tab
             this.addColsContainer(tabsContainers[0] as any);
+            indicators.addSearchBox.call(
+                this,
+                chart,
+                tabsContainers[0] as any
+            );
             indicators.addIndicatorList.call(
                 this,
                 chart,
                 tabsContainers[0] as any,
                 'add'
             );
-
             buttonParentDiv = tabsContainers[0]
                 .querySelectorAll('.' + PREFIX + 'popup-rhs-col')[0];
 
@@ -931,6 +948,33 @@ H.Popup.prototype = {
             if (indicatorList.childNodes.length > 0) {
                 (indicatorList.childNodes[0] as any).click();
             }
+        },
+        /**
+         * Add searchbox.
+         * @private
+         */
+        addSearchBox: function (
+            this: Highcharts.Popup,
+            chart: Highcharts.AnnotationChart,
+            parentDiv: HTMLDOMElement
+        ): void {
+            // Main parrent.
+            const lhsCol = parentDiv.querySelectorAll('.' + PREFIX + 'popup-lhs-col')[0],
+                options = 'type.searchIndicators',
+                inputAttributes = {
+                    value: '',
+                    type: 'text'
+                },
+                clearFilterText = this.lang.clearFilter;
+
+            const handleResetInput = function (): void {
+            };
+
+            // Add input field with the label.
+            this.addInput(options, INPUT, lhsCol as any, inputAttributes);
+
+            // Add button to reset the input.
+            const button = this.addButton(lhsCol as any, clearFilterText, 'button', handleResetInput, lhsCol as any);
         },
         /**
          * Extract full name and type of requested indicator.
@@ -1173,7 +1217,7 @@ H.Popup.prototype = {
                             parentFullName,
                             type,
                             parentDiv,
-                            ''
+                            '' as any
                         );
                         addParamInputs.call(
                             _self,
@@ -1192,7 +1236,10 @@ H.Popup.prototype = {
                             parentFullName,
                             type,
                             parentDiv,
-                            [value, 'text'] as any // all inputs are text type
+                            {
+                                value: value,
+                                type: 'text'
+                            } // all inputs are text type
                         );
                     }
                 }
--- a/ts/Stock/StockToolsGui.ts
+++ b/ts/Stock/StockToolsGui.ts
@@ -311,6 +311,8 @@ setOptions({
                 background: 'Background',
 
                 // Indicators' params (#15170):
+                searchIndicators: 'Search Indicators',
+                clearFilter: 'clear filter',
                 index: 'Index',
                 period: 'Period',
                 periods: 'Periods',
--- a/ts/Core/Renderer/HTML/HTMLAttributes.d.ts
+++ b/ts/Core/Renderer/HTML/HTMLAttributes.d.ts
@@ -31,6 +31,7 @@ export interface HTMLAttributes {
     name?: string;
     onclick?: any;
     onload?: any;
+    placeholder?: string;
     scope?: string;
     textContent?: string;
     title?: string;


ts/Stock/StockToolsGui.ts
```diff
@@ -311,6 +311,8 @@ setOptions({
                 background: 'Background',
 
                 // Indicators' params (#15170):
+                'search-indicators': 'Search Indicators',
```
pawelfus, line 314: `searchIndicators`*

ts/Extensions/Annotations/Popup.ts
```diff
@@ -830,17 +861,62 @@ H.Popup.prototype = {
...
+                sorted[array[key]] = series[array[key]];
+            }
+            return sorted;
+        },
+        /**
+         * Create an array of filtered and sorted series based on the filter.
+         * @private
+         */
+        filterSeries: function (series: Highcharts.FilteredSeries, filter?: string): Highcharts.FilteredSeries {
+            let filteredSeries = {} as Highcharts.FilteredSeries;
```
pawelfus, line 902: Just a discussion point. It seems that to create a list of indicators we do this:
```
A) Get the list for ADD/EDIT:
  a.1) If it's EDIT -> use array from `chart.series`
  a.2) If it's ADD -> use the object `plotOptions`
    a.2.1) Filter out series which are not indicators
    a.2.2) Filter out series which don't match the name from input 
    a.2.3) Sort series:
      a.2.2.1) Map object to array
      a.2.2.1) Sort array
      a.2.2.1) Map array to object
```
    
 How about:
 - we merge a.2.1) and a.2.2)?
 - we do the mapping from object to array in the first filter, and later use only an array of series (objects)? That would simplify TS (no Highcharts.FitleredSeries interface) and reduce the number of unnecessary loops/iterations over an object.

pawelfus (changes requested):

ts/Stock/StockToolsGui.ts
karolkolodziej: That is the tricky part because for now the `search-indicators` is used also as a class for that input.
But I guess that I might implement a separate mechanism to add the class name.

pawelfus: Or, change the name. For example `indicatorsSearchBox`, or something similar 🤔

karolkolodziej:
While playing with it I encountered some undesirable behaviour.
Take for example this scenario:
1. Type `a` into the input- you should have 7 elements
2. When you try to access the `Acceleration Bands` by typing the `ac` it doesn't work...
In order to make it possible you have to write `ab` which might be confusing for someone.

Also shouldn't we match the series with its lang interpretation? What if you translate that into other language or give them other names. Searching won't be possible. @pawelfus

pawelfus:
Nice catch! We need to implement aliases for the searching box. By default, we can pull values from lang

ts/Extensions/Annotations/Popup.ts
```diff
@@ -830,127 +876,321 @@ H.Popup.prototype = {
...
+            this: Highcharts.Popup,
+            series: Array<Series>
+        ): [[string, string, Series]] {
+            let filteredSeries: [[string, string, Series]] = [] as any;
+
+            // Allow only indicators.
+            series.forEach(function (series): void {
+                const seriesOptions = series.options;
+
+                if ((series as any).params || seriesOptions && (seriesOptions as any).params) {
```
raf18seb, line 964: `series.is('sma')`

ts/Extensions/Annotations/Popup.ts
```diff
@@ -830,127 +876,321 @@ H.Popup.prototype = {
...
+         * Filter an array of series and map its names and types.
+         *
+         * @private
+         *
+         * @param {Highcharts.FilteredSeries} series
+         *        All series are available in the plotOptions.
+         *
+         * @return {void}
+         */
+        filterSeriesArray: function (
```
raf18seb, line 954: Please, do not exceed 80 in this method

ts/Extensions/Annotations/Popup.ts
```diff
@@ -830,127 +876,321 @@ H.Popup.prototype = {
...
+         *        Type of list depending on the selected bookmark.
+         *        Might be 'add' or 'edit'.
+         *
+         * @param {string|undefined} filter
+         *        Applied filter string from the input.
+         *        For the first iteration, it's an empty string.
+         *
+         * @return {void}
          */
         addIndicatorList: function (
```
raf18seb, line 1001: Please, do not exceed 80 in this method

ts/Extensions/Annotations/Popup.ts
```diff
@@ -830,127 +876,321 @@ H.Popup.prototype = {
...
+                item = createElement(
+                    LI,
+                    {
                         className: PREFIX + 'indicator-list'
-                    }, void 0, indicatorList);
-                    item.appendChild(doc.createTextNode(
-                        indicatorNameType.name
-                    ));
+                    },
+                    void 0,
```
raf18seb, line 1066: Sometimes you use `null as any`, sometimes `void 0` for styles

sebastianbochan (changes requested):

ts/Extensions/Annotations/Popup.ts
```diff
@@ -830,127 +876,321 @@ H.Popup.prototype = {
...
+         *
+         * @param {Highcharts.AnnotationChart} chart
+         *        The chart object.
+         *
+         * @param {HTMLDOMElement} parentDiv
+         *        HTML parent element.
+         *
+         * @return {HTMLSelectElement}
+         */
+        addSearchBox: function (
```
raf18seb, line 1123: Please, do not exceed 80 in this method

ts/Extensions/Annotations/Popup.ts
```diff
@@ -830,127 +876,321 @@ H.Popup.prototype = {
...
+            };
+
+            // Add input field with the label and button.
+            const input = this.addInput(options, INPUT, lhsCol, inputAttributes) as HTMLInputElement,
+                button = this.addButton(lhsCol, clearFilterText, 'button', lhsCol);
+
+            button.classList.add('clear-filter-button');
+
+            // Add input change events.
+            ['input'].forEach(function (eventName: string): void {
```
raf18seb, line 1150: It `['input'].length === 1`, do we have to use `forEach` here?

ts/Extensions/Annotations/Popup.ts
```diff
@@ -133,8 +152,11 @@ declare global {
                 type: string,
                 parentDiv: HTMLDOMElement
             ): void;
+            filterSeries(series: SeriesTypePlotOptions, filter?: string): [[string, string, Series]];
+            filterSeriesArray(series: Array<Series>): [[string, string, Series]];
+            sortSeries(series: any): any;
```
karolkolodziej, line 157: unnececary?

ts/Extensions/Annotations/Popup.ts
```diff
@@ -1173,7 +1413,7 @@ H.Popup.prototype = {
                             parentFullName,
                             type,
                             parentDiv,
-                            ''
+                            '' as any
```
raf18seb, line 1416: Can we get rid of this `as any`?

ts/Extensions/Annotations/Popup.ts
```diff
@@ -872,55 +877,116 @@ H.Popup.prototype = {
...
-            for (key in series) {
-                if (Object.prototype.hasOwnProperty.call(series, key)) {
-                    array.push(key);
-                }
-            }
+        filterSeries: function (
+            this: Highcharts.Popup,
+            series: Highcharts.FilteredSeries,
+            filter?: string
+        ): [[string, string, Series]] {
```
sebastianbochan, line 910: Would an object be better than the array in the array?

raf18seb (changes requested):

ts/Extensions/Annotations/Popup.ts
```diff
@@ -310,14 +324,13 @@ H.Popup.prototype = {
...
-        if (value !== '') {
-
+        if (inputAttributes) {
             createElement(
                 INPUT,
                 {
                     name: inputName,
-                    value: value[0],
-                    type: value[1],
+                    value: inputAttributes.value as any,
```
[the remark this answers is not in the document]
karolkolodziej: In line 177 I've declared an unfortunate interface `InputAttributes`.
Changing it I can avoid casting to `any`.

css/annotations/popup.scss
```diff
@@ -238,4 +240,23 @@ $button-hover-color: #e6ebf5;
...
+}
+
+label[for="highcharts-input-search-indicators"] {
+  text-align: center;
+  font-weight: bold;
+  color: black;
+  margin-top: 0.5rem;
+}
+
+input[name="highcharts-input-search-indicators"] {
```
sebastianbochan, line 252: I would replace it with the class, context selectors are not supported in the older IE.

css/annotations/popup.scss
```diff
@@ -238,4 +240,23 @@ $button-hover-color: #e6ebf5;
 .highcharts-indicator-title {
   float: left;
   padding-bottom: 15px;
+}
+
+label[for="highcharts-input-search-indicators"] {
```
sebastianbochan, line 245: I would replace it with the class, context selectors are not supported in the older IE.

sebastianbochan (changes requested):

ts/Extensions/Annotations/Popup.ts
karolkolodziej: We've discussed that with @pawelfus and there are two main reasons why I chose array:
1. Avoiding looping through an object multiple times.
2. In the case of the `EDIT` bookmark, series are placed in the array. With array here, we can use one logic to handle both cases.

Later in the code, I've tried to give some descriptive names for each element in the array.
`[indicatorFullName, indicatorType, series]`

css/annotations/popup.scss
```diff
@@ -238,4 +240,23 @@ $button-hover-color: #e6ebf5;
...
+
+input[name="highcharts-input-search-indicators"] {
+   width: 80%;
+   margin-left: 10%;
+}
+
+.highcharts-popup button.clear-filter-button {
+  margin: 0 auto;
+  display: block;
+  float: unset;
```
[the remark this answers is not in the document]
karolkolodziej: Without that, button is floating to the right because of float declared in`.highcharts-popup button {`

ts/Extensions/Annotations/Popup.ts
```diff
@@ -830,127 +876,321 @@ H.Popup.prototype = {
...
+                indicators = popup.indicators;
+            let filteredSeries: [[string, string, Series]] = [] as any;
+
+            objectEach(series, function (
+                series: Series,
+                value: string
+            ): void {
+                const seriesOptions = series.options;
+                // Allow only indicators.
+                if ((series as any).params || seriesOptions && (seriesOptions as any).params) {
```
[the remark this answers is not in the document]
karolkolodziej: The `series` here is taken from the `chart.options.plotOptions` and they are not initialized/rendereed.
Thus I can't use `if (series.is('sma'))` and I have to use some specific properties available only for the indicators.

ts/Extensions/Annotations/Popup.ts
karolkolodziej: Here as the series is rendered I can check its' type.

ts/Extensions/Annotations/Popup.ts
karolkolodziej: However, an array of objects looks a bit better.

ts/Extensions/Annotations/Popup.ts
karolkolodziej: It's not my part of the code.
I've only specified the type which the `addInput` method should take but for some reason in some cases, it's required to be an empty string that is why I casted it to `any`.

pawelfus: Perhaps it should be an empty object? 🤔 
It seems `''` is leveraging undefined references later: `emptyString.value`/`htmlFor`/`labelClassName` etc resolve to `undefined`

ts/Extensions/Annotations/Popup.ts
```diff
@@ -795,162 +875,381 @@ H.Popup.prototype = {
...
-                if (
-                    (serie as any).params ||
-                    seriesOptions && (seriesOptions as any).params
-                ) {
+            rhsColWrapper = rhsCol
+                .querySelectorAll('.' + PREFIX + 'popup-rhs-col-wrapper')[0] as HTMLElement;
 
-                    const indicatorNameType = _self.indicators.getNameType(serie as any, value),
-                        indicatorType = indicatorNameType.type;
+            filteredSeriesArray.forEach(function (seriesSet: Highcharts.FilteredSeries): void {
```
pawelfus, line 1119: Rafał missed: 80+ ;)

ts/Extensions/Annotations/Popup.ts
```diff
@@ -795,162 +875,381 @@ H.Popup.prototype = {
...
+                const { indicatorFullName, indicatorType, series } = seriesSet;
```
pawelfus, line 1120: Are we allowed to destructure function arguments? 

```ts
filteredSeriesArray.forEach(function (seriesSet: Highcharts.FilteredSeries): void {
  const { indicatorFullName, indicatorType, series } = seriesSet;
```
->
```ts
filteredSeriesArray.forEach(function ({ indicatorFullName, indicatorType, series }): void {
```

@bre1470

ts/Extensions/Annotations/Popup.ts
```diff
@@ -795,162 +875,381 @@ H.Popup.prototype = {
...
-            indicatorList = createElement(UL, {
-                className: PREFIX + 'indicator-list'
-            }, null as any, lhsCol as any);
+            // Filter and sort the series.
+            if (!isEdit && series && !isArray(series)) {
+                // Apply filters only for the 'add' indicator list.
+                filteredSeriesArray = indicators.filterSeries.call(this, series, filter);
+            } else if (series && isArray(series)) {
+                filteredSeriesArray = indicators.filterSeriesArray.call(this, series);
+            }
```
pawelfus, line 1098: Maybe sort series here? Instead of duplicating sorting code in both `filterSeries` and `filterSeriesArray` functions.

ts/Extensions/Annotations/Popup.ts
```diff
@@ -795,162 +875,381 @@ H.Popup.prototype = {
...
+                if (
+                    (series as any).params || seriesOptions &&
+                    (seriesOptions as any).params
+                ) {
+                    const { indicatorFullName, indicatorType } =
+                        indicators.getNameType(series, value);
+
+                    if (filter) {
+                        const slicedValue = indicatorFullName.slice(0, filter.length);
+                        if (slicedValue.toLocaleLowerCase() === filter.toLocaleLowerCase()) {
```
pawelfus, line 970: In other words, typing `stochastic` will find only Stochastic technical indicator, and Slow Stochastic will be missing? Use safe regexp instead

pawelfus (changes requested):
PR is breaking all annotation inputs: https://jsfiddle.net/BlackLabel/xjdwemzu/1/

ts/Extensions/Annotations/Popup.ts
```diff
@@ -795,162 +875,381 @@ H.Popup.prototype = {
...
+                            series
+                        };
+
+                        filteredSeriesArray.push(filteredSeries);
+                    }
+                }
+            });
+
+            // Sort indicators alphabeticaly.
+            return filteredSeriesArray.sort(function (a, b): number {
```
pawelfus, line 992: Use `stableSort`

ts/Extensions/Annotations/Popup.ts
```diff
@@ -795,162 +875,381 @@ H.Popup.prototype = {
...
+                const seriesAName = a.indicatorFullName.toLowerCase(),
+                    seriesBName = b.indicatorFullName.toLowerCase();
+                return (seriesAName < seriesBName) ? -1 : (seriesAName > seriesBName) ? 1 : 0;
```
pawelfus, line 995: Do we need all of this? Simple `(a, b) => a.indicatorFullName > b.indicatorFullName` won't be enough? We are comparing strings here, right?

pawelfus (changes requested):
Missed two comments, sorry!

ts/Extensions/Annotations/Popup.ts
bre1470: Unlikely, unless Babel is doing something about it. TypeScript will not convert it, because we have a ES6 target for Babel.

karolkolodziej: With destructurization as Pawel suggested it works as before, however it compiles it a bit odd.
```JS
filteredSeriesArray.forEach(function (_a) {
   var indicatorFullName = _a.indicatorFullName, indicatorType = _a.indicatorType, series = _a.series;
  ...
```

I think the first approach is more readable and results in clearer code after compiling.

ts/Extensions/Annotations/Popup.ts
karolkolodziej: I'm afraid we do need this because the sorting function should return the number.
Having capital and lower case letters also makes sorting inaccurate.

karolkolodziej:
The current implementation (comparing the search phrase with the full name) might be limited in some cases.
Let's see that in the example.

We have the `Acceleration Bands` and `Bollinger Bands` indicators. However, the Bollinger Bands is named `BB` by default.
So if the user types 'band' there is only one result.
Screenshot 2021-07-30 at 11 51 24


What if we add in the `StockToolsGui.ts`(since that will be only used in the popup) to lang an array for each series with its' aliases and assigned them to each series. These aliases have to be visible in the `plotOptions` because based on that the list is created.
``` JS
// StockToolsGui
lang.stockTools.indicatorAliases:{
  sma: ['simple moving average']
  ...
}

// SMA INDICATOR
public static defaultOptions: SMAOptions = merge(LineSeries.defaultOptions, {
     ...
     aliases: langOptions.stockTools.indicatorAliases[indicatorType],
     ...
})
```

Then while looping through the `plotOptions` we would extract those aliases and filter through them :) 
@pawelfus @TorsteinHonsi  what do you think?

pawelfus:
Do we need to modify indicators? I think we don't need this information anywhere except for the Stock Tools, so I would keep the code there.

karolkolodziej:
That is even better yes!
```TS
// StockToolsGui.ts
lang.stockTools.indicatorAliases:{
  sma: 'simple moving average'
  ...
}
```
Then in the `filterSeries` we might get the aliases from the `getNameType` method.
``` TS
// Popup.ts
 const { indicatorFullName, indicatorType, indicatorAliases } =
                        indicators.getNameType(series, value);

 if (filter) {
       const regex = new RegExp(filter, 'i');
       if (indicatorFullName.match(regex) || indicatorAliases.match(regex)) {
              ...
       }
  }

```

pawellysy (approved):
Looks good!

pawelfus:
I'm wondering if aliases should be arrays. Do you think any of the indicators might need more than just one alias? Usually, we have abbreviations (SMA) that need an alias (Simple Moving Average). Perhaps when translating aliases? So that we could keep English name and e.g. Polish one.

karolkolodziej:
In the second comment, I added them as strings.

``` TS
// StockToolsGui.ts
lang.stockTools.indicatorAliases:{
  sma: 'simple moving average prosta średnia krocząca'
  ...
}
```
I agree it's better to have it as an array. We might have even multiple allies as a single string, not an array.

ts/Extensions/Annotations/Popup.ts
```diff
@@ -795,162 +880,383 @@ H.Popup.prototype = {
...
-                    (seriesTypes[type].prototype as any).nameBase || type.toUpperCase(),
-                seriesType = type;
+            series: Series,
+            indicatorType: string
+        ): Highcharts.IndicatorNameCouple {
+            const options = series.options,
+                seriesTypes = H.seriesTypes;
+            // add mode
+            let seriesName = seriesTypes[indicatorType] &&
+                    (seriesTypes[indicatorType].prototype as SMAIndicator).nameBase || indicatorType.toUpperCase(),
```
raf18seb, line 1258: 100

raf18seb (changes requested):

pawelfus (changes requested):
- There's again something wrong multi periods fields. It shows an extra input, for example, APO has two periods, but three inputs are displayed ([demo](http://jsfiddle.net/BlackLabel/v98j5nes/1/)): 
Zrzut ekranu 2021-08-3 o 17 30 15

- "Clear filters" button could use "pointer" cursor too :) And other styles:
  - normal state: text color and border: `#424BCD`, background: white
  - on hover: text color: white, background: `#424BCD`

karolkolodziej:
Duplicated inputs or inputs with undefined are fixed and tested in the #16097 PR.
I think there is no need to copy/paste that code ;)

raf18seb (changes requested):
See my previous "request changes" comments.

raf18seb (approved):
Thanks!

pawelfus (changes requested):
Missing border in "search indicators" ;)

pawelfus:
@TorsteinHonsi - could you check our discussion regarding aliases? See https://github.com/highcharts/highcharts/pull/16089#issuecomment-889828229 and next ~4 posts

TorsteinHonsi:
> could you check our discussion regarding aliases

I'm good with either solution you find best. But arrays are more flexible, like you suggest, translators can add english _and_ translated strings.

TorsteinHonsi:
Pending
* Conflicts to be resolved again
* Approvals by @sebastianbochan and @pawelfus

ts/Stock/StockToolsGui.ts
```diff
@@ -341,7 +343,399 @@ setOptions({
...
+                     */
+                    dmi: ['Directional Movement Index'],
+
+                    /**
+                     *  Detrended price alias.
+                     *
+                     * @default ['Detrended price’]
+                     * @type    {Array<string>}
+                     */
+                    dpo: ['Detrended price'],
```
pawelfus, line 607: `Detrended price oscillator`*

pawelfus (changes requested):
Looks great, just a minor polishing ([demo](http://jsfiddle.net/BlackLabel/81nyxogm/2/show)):
- missing input escaping: type `*`, `\` etc. in the input -> errors in the console
- could you add some spacing between input and the "clear filter" button (~10px?)?

Aliases:
- it seems you listed all, except for `Detrended price`
-

ts/Extensions/Annotations/Popup.ts
```diff
@@ -172,6 +195,12 @@ declare global {
             selectTab(this: Popup, tab: Element, index: number): void;
             switchTabs(this: Popup, disableTab: number): void;
         }
+        interface InputAttributes {
+            value?: string|undefined;
```
raf18seb, line 199: Is `undefined` required here?

ts/Extensions/Annotations/Popup.ts
```diff
@@ -816,20 +890,25 @@ H.Popup.prototype = {
                     storage.reverse(); // (#14691)
                 }
 
-                storage.forEach(function (genInput): void {
+                storage.forEach(function (genInput: any): void {
                     if ((genInput as any)[0] === true) {
```
raf18seb, line 894: You can remove `as any` here

ts/Extensions/Annotations/Popup.ts
```diff
@@ -816,20 +890,25 @@ H.Popup.prototype = {
...
-                            (genInput as any)[2]
+                            genInput[2]
                         ).appendChild(doc.createTextNode(
-                            (genInput as any)[1]
+                            genInput[1]
                         ));
 
                     } else {
-                        addInput.apply((genInput as any)[0], (genInput as any).splice(1));
+                        const inputAttributes = {
```
raf18seb, line 906: Do we need to declare a new variable here?

ts/Extensions/Annotations/Popup.ts
```diff
@@ -865,143 +944,393 @@ H.Popup.prototype = {
...
+            let rhsColWrapper: HTMLElement,
                 indicatorList: HTMLDOMElement,
-                item: HTMLDOMElement;
+                item: HTMLDOMElement,
+                filteredSeriesArray: [Highcharts.FilteredSeries] = [] as any;
 
-            if (!chart) {
+            if (!chart && series) {
                 return;
             }
```
raf18seb, line 1161: What about moving this `if` statement a few lines above the `let` declaration?

ts/Extensions/Annotations/Popup.ts
```diff
@@ -865,143 +944,393 @@ H.Popup.prototype = {
...
 
-            // create wrapper for list
-            indicatorList = createElement(UL, {
-                className: PREFIX + 'indicator-list'
-            }, null as any, lhsCol as any);
+            // Filter and sort the series.
+            if (!isEdit && series && !isArray(series)) {
+                // Apply filters only for the 'add' indicator list.
+                filteredSeriesArray = indicators.filterSeries.call(this, series, filter);
+            } else if (series && isArray(series)) {
```
raf18seb, line 1167: Do we need to check `series &&` here?

raf18seb (changes requested):
Nice! Only small polishing suggestions

pawellysy (approved):
Looks good

ts/Extensions/Annotations/NavigationBindings.ts
```diff
@@ -1184,14 +1191,14 @@ setOptions({
...
          *
-         * @since   7.0.0
          * @product highcharts highstock
          */
         navigation: {
             /**
              * Translations for all field names used in popup.
              *
              * @product highcharts highstock
+             * @since   7.0.0
```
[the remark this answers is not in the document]
karolkolodziej: I was testing one thing and I forgot to change that. Thanks ;)

ts/Extensions/Annotations/Popup.ts
```diff
@@ -865,143 +943,397 @@ H.Popup.prototype = {
...
+                // Allow only indicators.
+                if (
+                    (series as any).params || seriesOptions &&
+                    (seriesOptions as any).params
+                ) {
+                    const { indicatorFullName, indicatorType } =
+                        indicators.getNameType(series, value);
+
+                    if (filter) {
+                        try {
```
pawelfus, line 1041: Could you replace `try/catch` with `filter.replace(keys, flags)`? Thanks!

pawelfus (changes requested):
Just one, tiny thing. Thanks!

raf18seb (approved):
Good job, thanks!

ts/Extensions/Annotations/Popup.ts
```diff
@@ -307,55 +346,79 @@ H.Popup.prototype = {
...
+     * @param {string} indicatorType
+     *        Type of the indicator i.e. sma, ema...
+     *
+     * @param {HTMLDOMElement} parentDiv
+     *        HTML parent element.
+     *
+     * @param {Highcharts.InputAttributes} inputAttributes
+     *        Attributes of the input.
+     *
+     * @return {HTMLSelectElement}
```
bre1470, line 401: Here is a type mismatch and a description missing. Without one does not now, which element to expect (parent, label or input).

ts/Extensions/Annotations/Popup.ts
```diff
@@ -865,143 +943,397 @@ H.Popup.prototype = {
...
+            indicators.addSearchBox.call(
+                this,
+                chart,
+                tabsContainers[0] as HTMLDOMElement
+            );
             indicators.addIndicatorList.call(
                 this,
                 chart,
-                tabsContainers[0] as any,
+                tabsContainers[0] as HTMLDOMElement,
```
bre1470, line 955: Why do you force cast. Does the tabsContainers do not contain HTMLElements?

ts/Extensions/Annotations/Popup.ts
```diff
@@ -865,143 +943,397 @@ H.Popup.prototype = {
...
+         * @private
+         *
+         * @param {Highcharts.FilteredSeries} series
+         *        All series are available in the plotOptions.
+         *
+         * @param {string|undefined} filter
+         *        Applied filter string from the input.
+         *        For the first iteration, it's an empty string.
+         *
+         * @return {void}
```
bre1470, line 1010: `@return {void}` can be removed.

ts/Extensions/Annotations/Popup.ts
```diff
@@ -865,143 +943,397 @@ H.Popup.prototype = {
...
+         *
+         * @private
+         *
+         * @param {Highcharts.AnnotationChart} chart
+         *        The chart object.
+         *
+         * @param {HTMLDOMElement} parentDiv
+         *        HTML parent element.
+         *
+         * @return {HTMLSelectElement}
```
bre1470, line 1273: Return type mismatch, probably it can be removed.

ts/Extensions/Annotations/Popup.ts
```diff
@@ -1148,23 +1482,30 @@ H.Popup.prototype = {
...
-         * @param {string} - indicator type like: sma, ema, etc.
-         * @return {Object} - series name and type like: sma, ema, etc.
+         *        Series which name is needed(EDITmode - defaultOptions.series,
+         *        ADDmode - indicator series).
+         *
+         * @param {string} [IndicatorType]
+         *        Type of the indicator i.e. sma, ema...
+         *
+         * @return {Highcharts.IndicatorNameCouple}
+         *        Full name and series type.
```
bre1470, line 1496: Type `Highcharts.Dictionary<string>` is enough, no need to introduce an extra type for generics.

bre1470 (changes requested):

ts/Extensions/Annotations/Popup.ts
karolkolodziej: I get `Argument of type 'Element' is not assignable to parameter of type 'HTMLElement'.` and I thought that casting to `HTMLDOMElement` would be better than to `any`

bre1470: For sure it is, but it also disables most TypeScript validation (not as much as `any` though). I was wondering, if we could have `tabsContainers` limit itself to `HTMLDOMElement`. Or maybe `addIndicatorList` would be happy with `Element` type instead of `HTMLDOMElement` as well.

ts/Extensions/Annotations/Popup.ts
```diff
@@ -865,143 +944,388 @@ H.Popup.prototype = {
...
+                        indicatorFullName: series.name,
+                        indicatorType: series.type,
+                        series
+                    };
+
+                    filteredSeriesArray.push(filteredSeries);
+                }
+            });
+
+            return filteredSeriesArray;
```
bre1470, line 1102: The function doclet has no return information. The return type is a singlet with FilteredSeries. Though the loop can push more than one series. If it is really just one, then just return the first series directly out of the loop (classic for).

ts/Extensions/Annotations/Popup.ts
```diff
@@ -865,143 +944,388 @@ H.Popup.prototype = {
...
+                            indicatorType,
+                            series
+                        };
+
+                        filteredSeriesArray.push(filteredSeries);
+                    }
+                }
+            });
+
+            return filteredSeriesArray;
```
bre1470, line 1069: The same question here. Singlet or array?

bre1470 (changes requested):

ts/Extensions/Annotations/Popup.ts
bre1470: And otherwise the return type should be a regular array (`Array<Highcharts.FilteredSeries>`).

pawelfus (approved):
Nothing to add, thanks!

ts/Extensions/Annotations/Popup.ts
```diff
@@ -826,20 +826,25 @@ H.Popup.prototype = {
                     storage.reverse(); // (#14691)
                 }
 
-                storage.forEach(function (genInput): void {
+                storage.forEach(function (genInput: any): void {
```
sebastianbochan, line 894: Do we need this `any?

I noticed that the `getInput` is an array of objects with `type` and `value`.

Ref: https://github.com/highcharts/highcharts/commit/cc40c239b3f06baae0ed1167c808e178b239be5f#diff-276aaccade8e98f0f52f9dae7d9e60c247861424ae8a2cbbb84d8ceb1724393eR907

sebastianbochan (changes requested):

sebastianbochan:
One thing more: https://github.com/highcharts/highcharts/commit/cc40c239b3f06baae0ed1167c808e178b239be5f#diff-276aaccade8e98f0f52f9dae7d9e60c247861424ae8a2cbbb84d8ceb1724393eR907

I do know why I cannot request this change.

sebastianbochan:
One thing more: https://github.com/highcharts/highcharts/commit/cc40c239b3f06baae0ed1167c808e178b239be5f#diff-276aaccade8e98f0f52f9dae7d9e60c247861424ae8a2cbbb84d8ceb1724393eR894

I do know why I cannot request this change.

ts/Extensions/Annotations/Popup.ts
karolkolodziej: This casting to any was written whenever the `genInput` was used that is why I decided to simplify it a bit and I moved that `as any` here.
I guess there might be a way to get rid of that nasty casting however this isn't the purpose of that PR.

karolkolodziej: When it comes to `if (genInput[0] === true) {` it has to be like that.. 
Without that the popup in Annotation is failing

sebastianbochan (approved):

ts/Extensions/Annotations/Popup.ts
```diff
@@ -865,143 +944,394 @@ H.Popup.prototype = {
...
+         *
+         * @return {Array<Highcharts.FilteredSeries>} filteredSeriesArray
+         *         Returns array of filtered series based on filter string.
+         */
+        filterSeries: function (
+            this: Highcharts.Popup,
+            series: Series,
+            filter?: string
+        ): [Highcharts.FilteredSeries] {
+            const popup = this,
```
bre1470, line 1019: Return type should be `Array<Highcharts.FilteredSeries>`.

ts/Extensions/Annotations/Popup.ts
```diff
@@ -865,143 +944,394 @@ H.Popup.prototype = {
...
+                indicators = popup.indicators,
+                lang = popup.chart && popup.chart.options.lang,
+                indicatorAliases = lang &&
+                    lang.navigation &&
+                    lang.navigation.popup &&
+                    lang.navigation.popup.indicatorAliases;
+            let filteredSeriesArray: [Highcharts.FilteredSeries] = [] as any,
```
bre1470, line 1026: Array type should be Array<Highcharts.FilteredSeries>, then you can remove `any`.

ts/Extensions/Annotations/Popup.ts
```diff
@@ -865,143 +944,394 @@ H.Popup.prototype = {
...
+         * @param {Highcharts.FilteredSeries} series
+         *        All series that are available in the plotOptions.
+         *
+         * @return {Array<Highcharts.FilteredSeries>} filteredSeriesArray
+         *         Returns array of filtered series based on filter string.
+         */
+        filterSeriesArray: function (
+            this: Highcharts.Popup,
+            series: Array<Series>
+        ): [Highcharts.FilteredSeries] {
```
bre1470, line 1089: Return type should be `Array<Highcharts.FilteredSeries>`.

ts/Extensions/Annotations/Popup.ts
```diff
@@ -865,143 +944,394 @@ H.Popup.prototype = {
...
+            let filteredSeriesArray: [Highcharts.FilteredSeries] = [] as any,
```
bre1470, line 1090: Array type should be `Array<Highcharts.FilteredSeries>` and `any` be removed.

bre1470 (changes requested):
A little more typing, otherwise good.

bre1470:
Thank you! Looks good. 👍

bre1470 (approved):

css/annotations/popup.scss
```diff
@@ -1,7 +1,8 @@
 // Colors for buttons.
 $button-background-color: #f7f7f7;
 $button-hover-color: #e6ebf5;
-
+$blue: #424bcd;
+$white: #fff;
```
TorsteinHonsi, line 5: Use semantic variable names for theming (like the existing `$button-background-color` or `$button-hover-color`).  Imagine for example a designer styling the Stock Tools with a dark theme with orange highlight colors, then color variable names like $blue and $white are misleading.

TorsteinHonsi (changes requested):
Just a small comment on SASS variable names, otherwise ready to approve.

css/annotations/popup.scss
pawelfus: I was curious about the good practices for naming SCSS. Shall we add this to.. erm some guideline's readme? Not sure where could it fit nice

TorsteinHonsi: Not sure either... But if you look at the naming in highcharts.scss it is generally semantic. The idea would be that it would make sense regardless of the actual color used.

TorsteinHonsi: Descriptions like "light-color" and "dark-color" is also not good. In a typical dark mode design, the background would be dark and the typography light, so this would be counterintuitive. It is desireble to reduce the color palette in Highcharts, so I suggest we replace these color names _and_ values with the existing colors in `highcharts.scss`. We can for example use `$highlight-color-80` for the blue and `$background-color` for the white.

TorsteinHonsi (changes requested):
Sorry for coming in at this late stage, but I think the UX needs some polishing.
* The "Clear filter" button takes too much attention. It should either be hidden or disabled until a filter is actually active. Maybe also drop the border?
* There are double scrollbars.

The screenshot is taken from http://utils.highcharts.local/samples/#view/stock/demo/stock-tools-gui
Skjermbilde 2021-09-21 kl  11 09 16

pawelfus:
> The "Clear filter" button takes too much attention.

It is so big in the utils, check this demo http://jsfiddle.net/BlackLabel/s7nv2f46/

Zrzut ekranu 2021-09-21 o 11 35 56

>  It should either be hidden or disabled until a filter is actually active. Maybe also drop the border?

Any thoughts @oysteinmoseng @nasin ?

> There are double scrollbars.

+1

karolkolodziej:
1. I guess the `Clear filter` button could be shown on the `input` focus event.
But after that, should it always be visible or there are some cases when it should be hidden?

2. Double scrollbar could be fixed in two ways:
a) Input travels along with the lint (when we are at the bottom of the initial list, we don't see the input)
b) Input is positioned undependable at the top of the column (only the list is scrollable)

oysteinmoseng:
Initial UX/a11y thoughts:
- The "clear filter" button can definitely be hidden (`display: none`) until there are active filters.
- It should probably be labelled "Clear filters" if there can be multiple filters.
- You could remove the button border, but then I would underline it to indicate that it can be interacted with. Nancy might have more thoughts on this.
- I would expect the search input field to always be visible, and the list to scroll independently.
- The border around form input fields (including search field) has to be at least 3:1 contrast with background, not sure if that is the case with that light grey?
- Should the default font be a sans-serif, just to look more modern?

pawelfus:
> I guess the Clear filter button could be shown on the input focus event.
> After that, should it always be visible or there are some cases when it should be hidden?

Perhaps when input is empty -> hidden "clear filter" button. When not empty, show it. There's probably "onkeypressed" event that can hide/show input and button-click can hide it too 🤔 

> Input is positioned undependable at the top of the column (only the list is scrollable)

IMHO just the list enough

TorsteinHonsi:
I did some changes to the CSS. Re-used color variable names from `highcharts.scss` and also re-used the same font-family setting as suggested. Also updated the styling in utils so it doesn't intervene.

> Perhaps when input is empty -> hidden "clear filter" button. When not empty, show it.

That's what I was thinking. Unless it makes the content jump up and down when reflowing, which may be bad.

nasin:
> Initial UX/a11y thoughts:
> 
>     * The "clear filter" button can definitely be hidden (`display: none`) until there are active filters.
> 
>     * It should probably be labelled "Clear filters" if there can be multiple filters.
> 
>     * You could remove the button border, but then I would underline it to indicate that it can be interacted with. Nancy might have more thoughts on this.
> 
>     * I would expect the search input field to always be visible, and the list to scroll independently.
> 
>     * The border around form input fields (including search field) has to be at least 3:1 contrast with background, not sure if that is the case with that light grey?
> 
>     * Should the default font be a sans-serif, just to look more modern?

I think it's ok to hide the "clear filters" button until there are filters to clear, and then to display it as a link rather than a button so it takes up less space. The border on the inputs is too light. hex color 949494 will work.

nasin:
>     1. I guess the `Clear filter` button could be shown on the `input` focus event.
>        After that, should it always be visible or there are some cases when it should be hidden?
>        But to be honest I don't know if it's that necessary. As Paweł mentioned in the demo it looks much smaller than in utils.
> 
>     2. Double scrollbar could be fixed in two ways:
>        a) Input travels along with the lint (when we are at the bottom of the initial list, we don't see the input)
>        b) Input is positioned undependable at the top of the column (only the list is scrollable)

My intention was for "b" to happen...only the list scrolls.

pawelfus:
Thanks! Okay, let's try the suggestions @karolkolodziej, so we can test the UX and make the decision

karolkolodziej:
What has changed:

- [ ] only one scrollbar
- [ ] hide scrollbar when is not needed
- [ ] input border
- [ ] anchor instead of the button
- [ ] input always visible in a fixed position at the top

@pawelfus @TorsteinHonsi @nasin let me know how this looks like 😄

pawelfus (commented):
Just realized one thing: when filtering out indicators, the first one from the list is displayed in the right column. This is a nice feature, but when there's no match, shouldn't the right column be empty? Perhaps not empty, rather contain some info, like "No match" etc.?

Alternatively, don't update the right column, when filtering out indicators. For example, Acceleration Bands will be always displayed unless a user chooses a different indicator. 

What do you think? @TorsteinHonsi @karolkolodziej @nasin @oysteinmoseng 

PS: @karolkolodziej - missing "pointer" on the cursor when hovering the "clear" button

karolkolodziej:
> Just realized one thing: when filtering out indicators, the first one from the list is displayed in the right column. This is a nice feature, but when there's no match, shouldn't the right column be empty? Perhaps not empty, rather contain some info, like "No match" etc.?

I've added a simple POC of how it might work.
Let me know what you think about it ;)
If the feedback is positive, I will add it to `lang` and create some tests.

TorsteinHonsi:
> Let me know what you think about it ;)

Yup, I think that's a good approach

PR #16091, opened 2021-07-21
bugfix/16084-timeline-dataLabel-update-enabled

### LINKED ISSUE #13671: Inverting timeline results in incorrect label positioning
#### Expected behaviour
When chart.inverted changed, labels are in proper order/position.

#### Actual behaviour
Labels are out of order.

https://codepen.io/epfromer/pen/abdvEyY/ 
- Click on invert button to toggle

#### Product version
8.1.0 

#### Affected browser(s)
All

### LINKED ISSUE #16084: Dynamic update of datalabel on point click is not working for highcharts timeline graph
#### Expected behaviour

I am trying to show dataLabel of a point on click which is easy in normal line charts, but seems like it is no more case with highcharts timeline chart. Is there a workaround for this?

#### Actual behaviour
updating dataLabel on a series point click is not working for timeline charts.

Here is the fiddle for normal line chart which works for enabling dataLabel on click , https://jsfiddle.net/4zdLr70f/

Here is the fiddle for timeline chart which is not working , https://jsfiddle.net/zxpa8ned/

#### Product version
latest version

#### Affected browser(s)
Generic

khlieng:
Fixed #16084, #13671, updating timeline chart did not work properly in some cases.

Closes #13671.

### DIFF
--- a/samples/unit-tests/series-timeline/datalabels/demo.js
+++ b/samples/unit-tests/series-timeline/datalabels/demo.js
@@ -26,7 +26,10 @@ QUnit.test('Timeline: General tests.', function (assert) {
                     },
                     {
                         name: 'Date 3',
-                        label: 'Some label'
+                        label: 'Some label',
+                        dataLabels: {
+                            enabled: false
+                        }
                     }
                 ]
             }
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (approved):
Thanks!

PR #16093, opened 2021-07-22
bugfix/11712-dependencywheel-update-node

### LINKED ISSUE #11712: Dependency Wheel - this.linksTo is undefined after calling node.update in afterAnimate event

khlieng:
Fixed #11712, updating dependency wheel nodes did not work properly.

### DIFF
--- a/ts/Core/Series/Point.ts
+++ b/ts/Core/Series/Point.ts
@@ -978,7 +978,9 @@ class Point {
         function update(): void {
 
             point.applyOptions(options);
-            point.resolveColor();
+            if (!series.options.colorByPoint) {
+                point.resolveColor();
+            }
 
             // Update visuals, #4146
             // Handle dummy graphic elements for a11y, #12718
@@ -1009,19 +1011,21 @@ class Point {
                 }
             }
 
-            // record changes in the parallel arrays
-            i = point.index as any;
-            series.updateParallelArrays(point, i);
+            if (!(point as any).isNode) { // #11712
+                // record changes in the parallel arrays
+                i = point.index as any;
+                series.updateParallelArrays(point, i);
 
-            // Record the options to options.data. If the old or the new config
-            // is an object, use point options, otherwise use raw options
-            // (#4701, #4916).
-            (seriesOptions.data as any)[i] = (
-                isObject((seriesOptions.data as any)[i], true) ||
-                    isObject(options, true)
-            ) ?
-                point.options :
-                pick(options, (seriesOptions.data as any)[i]);
+                // Record the options to options.data. If the old or the new
+                // config is an object, use point options, otherwise use raw
+                // options (#4701, #4916).
+                (seriesOptions.data as any)[i] = (
+                    isObject((seriesOptions.data as any)[i], true) ||
+                        isObject(options, true)
+                ) ?
+                    point.options :
+                    pick(options, (seriesOptions.data as any)[i]);
+            }
 
             // redraw
             series.isDirty = series.isDirtyData = true;
--- a/samples/unit-tests/styled-mode/series-states/demo.js
+++ b/samples/unit-tests/styled-mode/series-states/demo.js
@@ -47,4 +47,18 @@ QUnit.test('Inactive state and styledMode', function (assert) {
         'Series should not have inactive state class, ' +
             'when points within inherit state'
     );
+
+    chart.series[0].nodes[0].update({});
+    assert.strictEqual(
+        chart.series[0].nodes[0].index,
+        0,
+        '#11712: Node index should be correct after update'
+    );
+
+    chart.series[0].nodes[2].update({});
+    assert.strictEqual(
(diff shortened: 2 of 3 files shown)


ts/Core/Series/Point.ts
```diff
@@ -978,7 +978,9 @@ class Point {
         function update(): void {
 
             point.applyOptions(options);
-            point.resolveColor();
+            if (!series.options.colorByPoint) {
```
pawelfus, line 981: It seems, checking the `series.options.colorByPoint` option is already done in `resolveColor()`. Perhaps we should fix something there?

ts/Core/Series/Point.ts
```diff
@@ -1009,19 +1011,21 @@ class Point {
...
+                // Record the options to options.data. If the old or the new
+                // config is an object, use point options, otherwise use raw
+                // options (#4701, #4916).
+                (seriesOptions.data as any)[i] = (
+                    isObject((seriesOptions.data as any)[i], true) ||
+                        isObject(options, true)
+                ) ?
+                    point.options :
+                    pick(options, (seriesOptions.data as any)[i]);
+            }
```
pawelfus, line 1028: There's no need for `else`? Like, which arrays (perhaps `options.nodes`?) should be updated when updating a node? Is there a chance to move this "else" logic to NodesMixin? 🤔 It seems to be tricky...

ts/Core/Series/Point.ts
khlieng: But I kind of need to skip it only on `update`, to fix a node update regression introduced by #15878. There might be some better way to deal with `colorCounter` though.

Also noticing `nodeLookup` is stored but doesnt actually get used the next time `generatePoints` runs, so all the nodes get re-created every time, not sure if intentional 🤔.

ts/Core/Series/Point.ts
khlieng: Yes its probably needed for exporting, maybe it will be dealt with by #15904?

ts/Core/Series/Point.ts
pawelfus: > But I kind of need to skip it only on `update`, to fix a node update regression introduced by #15878. There might be some better way to deal with `colorCounter` though.

Ah, I see. Is the dependency wheel the only series with this regression? Perhaps #15878 should add `resolveColor()` in candlestick's `point.update`?

> Also noticing nodeLookup is stored but doesnt actually get used the next time generatePoints runs, so all the nodes get re-created every time, not sure if intentional 🤔.

No idea 😅 But `networkgraph` series uses this lookup object to add/remove nodes, so it's expected that `nodesLuookup` is up to date. Of course, instead of recreating the whole structure (and nodes), updating would be faster

khlieng: Anything `colorByPoint`: https://jsfiddle.net/u1mc8tdh/2/, `resolveColor` on `update` could probably still be useful in other cases.

pawelfus:
Closing as the main ticket is closed.

Issue #15497, opened 2021-04-13
Stock Tools gui - pivot points: change algorithm selection from text field to dropdown


pawellysy:
# Description of the feature

When creating pivot points indicator through the stock tools gui, you have the option to select, what type of algorithm you want the indicator to use. The available options are `'standard'`, `'fibonacci'` and `'camarilla'`, but in the current version, you have to type the name of the option. The better solution to the selection of the algorithm would be a dropdown, with all options listed.

# Library related to the feature

Highcharts Stock

# Proof of Concept/Live example for the feature

https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/stock/demo/stock-tools-gui
In the official demo, select indicator -> pivot Points and try to change algorithm type.

----

**You can vote for this feature by adding a thumbs-up reaction to this post.**

PR #16097, opened 2021-07-23
enhancement/15497-gui-pivot-points-dropdown 

karolkolodziej:
Added a selection box for algorithm parameter inside the Stock Tools popup window.  See #15497.

- [x] test


test/cypress/support/index.js
```diff
@@ -31,3 +31,7 @@ Cypress.Commands.add('chart', () =>
         }
     }))
 );
+
+Cypress.Commands.add('openIndicators', () =>
+    cy.get('.highcharts-menu-item-btn').first().click()
```
pawelfus, line 36: Instead of the first button, would it be better to get `.highcharts-indicators .highcharts-menu-item-btn`?

ts/Extensions/Annotations/Popup.ts
```diff
@@ -137,11 +151,11 @@ declare global {
             getNameType(series: SMAIndicator, type: string): Record<string, string>;
             listAllSeries(
                 this: Popup,
-                type: string,
+                indicatorTypepor: string,
```
pawelfus, line 154: typepor?

pawelfus (changes requested):

ts/Extensions/Annotations/Popup.ts
```diff
@@ -932,6 +955,132 @@ H.Popup.prototype = {
...
+         * @param {string} [optionName]
+         *        Name of the option into which selection is being added.
+         *
+         * @param {HTMLSelectElement} [selectBox]
+         *        HTML select box element to which the options are being added.
+         *
+         * @param {string|undefined} indicatorType
+         *        Type of the indicator i.e. sma, ema...
+         *
+         * @param {string|undefined} indicatorType
```
karolkolodziej, line 1029: Wrong param

ts/Stock/Indicators/PivotPoints/PivotPointsOptions.d.ts
```diff
@@ -24,6 +24,9 @@ import type {
 
 export interface PivotPointsOptions extends SMAOptions {
     params?: PivotPointsParamsOptions;
+    options: {
+        algorithm: Array<string>;
+    };
```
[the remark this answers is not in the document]
karolkolodziej: Right now they are declared per series. However, I will move them into the StockTools file.

ts/Extensions/Annotations/Popup.ts
```diff
@@ -410,11 +444,22 @@ H.Popup.prototype = {
             }
         });
 
+        selectList.forEach(function (select: HTMLSelectElement): void {
+            const id = select.id;
+
+            // Get inputs only for the parameters, not for series and volume.
+            if (id !== PREFIX + 'select-series' && id !== PREFIX + 'select-volume') {
+                const parameter = select.id.split('highcharts-select-')[1];
```
raf18seb, line 452: You can use the `id.split('...` (on variable) here instead of `select.id.split('...`

ts/Extensions/Annotations/Popup.ts
```diff
@@ -932,6 +977,140 @@ H.Popup.prototype = {
...
+                SELECT,
+                {
+                    name: selectName,
+                    className: PREFIX + 'popup-field'
+                },
+                null as any,
+                parentDiv
+            ) as HTMLSelectElement;
+
+            selectBox.setAttribute('id', PREFIX + 'select-' + optionName);
```
raf18seb, line 1030: Can we add the `id` inside `createElement` instead of calling `setAttribute` again?

ts/Extensions/Annotations/Popup.ts
```diff
@@ -932,6 +977,140 @@ H.Popup.prototype = {
...
+            indicatorType?: string,
+            parameterName?: string,
+            selectedOption?: string
+        ): void {
+            const popup = this;
+
+            // Get and apply selection options for the possible series.
+            if (optionName === 'series') {
+                // List all series which have id - mandatory for indicator.
+                chart.series.forEach(function (serie): void {
```
raf18seb, line 1073: `serie` :D

ts/Extensions/Annotations/Popup.ts
```diff
@@ -932,6 +977,140 @@ H.Popup.prototype = {
...
+                                value: seriesOptions.id
+                            },
+                            void 0,
+                            selectBox
+                        ).appendChild(doc.createTextNode(
+                            seriesOptions.name || seriesOptions.id
+                        ));
+                    }
+                });
+            } else if (chart.options.plotOptions && indicatorType && parameterName) {
```
raf18seb, line 1093: Why `chart.options.plotOptions` is required here?

ts/Extensions/Annotations/Popup.ts
```diff
@@ -932,6 +977,140 @@ H.Popup.prototype = {
...
+                // Get and apply options for the possible parameters.
+                const parameterOption = pivotPointsAlgorithm;
+
+                parameterOption && parameterOption.forEach(function (element): void {
```
raf18seb, line 1097: Is `parameterOption &&` required here? We know that this is a nonempty array.

Even if we didn't know whether this is an array, the condition should be different. I think we can just remove `parameterOption &&`.

raf18seb (changes requested):

ts/Extensions/Annotations/Popup.ts
```diff
@@ -932,6 +977,140 @@ H.Popup.prototype = {
...
+                            void 0,
+                            selectBox
+                        ).appendChild(doc.createTextNode(
+                            seriesOptions.name || seriesOptions.id
+                        ));
+                    }
+                });
+            } else if (chart.options.plotOptions && indicatorType && parameterName) {
+                // Get and apply options for the possible parameters.
+                const parameterOption = pivotPointsAlgorithm;
```
pawelfus, line 1095: Q: It seems we should use lookup object here, like:

```js
const parameterOption = dropdownParameters[parameterName];
```

Where `dropdownParameters` is defined as:
```js
const dropdownParameters = {
    'params.algorithm': ['standard', 'fibonacci', 'camarilla']
}
```

PS: Using `indicatorType` in the key might be necessary, in the case of two indicators using the same option name, but with a different list of values.

pawelfus (commented):
Just one discussion point from me

ts/Extensions/Annotations/Popup.ts
```diff
@@ -369,35 +404,39 @@ H.Popup.prototype = {
...
-            linkedTo: linkedTo && linkedTo.getAttribute('value'),
+            linkedTo: linkedTo && linkedTo.getAttribute('value') || '',
             fields: { }
-        } as any;
+        };
 
-        [].forEach.call(inputList, function (input: HTMLInputElement): void {
-            param = input.getAttribute(PREFIX + 'data-name');
-            seriesId = input.getAttribute(PREFIX + 'data-series-id');
+        inputList.forEach(function (input: HTMLInputElement): void {
```
pawelfus, line 437: This won't work in IE

ts/Extensions/Annotations/Popup.ts
```diff
@@ -410,11 +449,22 @@ H.Popup.prototype = {
             }
         });
 
+        selectList.forEach(function (select: HTMLSelectElement): void {
```
pawelfus, line 452: This won't work in IE

pawelfus (changes requested):
Just realized IE support will be broken with these two changes

ts/Extensions/Annotations/Popup.ts
raf18seb: IMHO I still prefer calling `forEach` directly on a converted array. To convert nodelist to array in old IE you can use `slice` like this:
```js
const inputList = Array.prototype.slice.call(parentDiv.querySelectorAll(INPUT));

inputList.forEach(function (...
...
```

pawelfus: Is there any difference between `[].forEach.call(inputList, ...)` and `Array.prototype.slice.call(parentDiv.querySelectorAll(INPUT)).forEach(...)`?

karolkolodziej: Apart from that it's look clearer (at least for me) it's the same.

pawellysy (changes requested):
I've noticed, that in indicators, which require volume series, it is impossible to select the volume series from the dropdown. 
image

raf18seb (approved):
Thanks!

pawelfus:
> I've noticed, that in indicators, which require volume series, it is impossible to select the volume series from the dropdown.

@karolkolodziej - could you also add a test for this case, please?

pawelfus (changes requested):
These changes also duplicate "periods" params for some of the technical indicators: http://jsfiddle.net/BlackLabel/d8y5hzmw/5/

Zrzut ekranu 2021-08-3 o 15 22 54

pawellysy (approved):

pawelfus (approved):
Thanks!

TorsteinHonsi (changes requested):
Thanks! Please resolve conflicts, then it's all good.

TorsteinHonsi (approved):

PR #16099, opened 2021-07-23
Tools: scripts-ts v2

bre1470:
- Made default export compatible to UMD.
- Removed `compare-api` in favor of `test-tree` task.
- Optimized `scripts-code` task.


TorsteinHonsi (approved):
Thanks!

Issue #16037, opened 2021-07-13
Docs: Annotations.shapes/labels.points are missing `function() {}` API


pawelfus:
#### Actual behaviour
[`annotations.shapes.points`](https://api.highcharts.com/highcharts/annotations.shapes.points)(and `labels`) don't mention that `points` can be an array of functions which should return `{x, y, xAxis, yAxis}` object.

#### API example
A basic example that could be added to the API: https://jsfiddle.net/BlackLabel/pau75hcg/ (calculates and shows local min/max/avg values, zoom-in to change the labels)

#### Product version
Highcharts

PR #16108, opened 2021-07-27
enhancement/16037-annotation-shapes-point-docs

karolkolodziej:
Enhancement #16037, improved annotation shape point documentation.

### DIFF
--- a/ts/Extensions/Annotations/Annotations.ts
+++ b/ts/Extensions/Annotations/Annotations.ts
@@ -1985,7 +1985,7 @@ export default Annotation;
 /**
  * Callback function that returns the annotation shape point.
  *
- * @callback Highcharts.AnnotationShapePointFunction
+ * @callback Highcharts.AnnotationMockPointFunction
  *
  * @param  {Highcharts.Annotation} annotation
  *         An annotation instance.
@@ -1997,8 +1997,10 @@ export default Annotation;
 /**
  * Shape point as object or function.
  *
- * @type {Highcharts.AnnotationMockPointOptionsObject|Highcharts.AnnotationShapePointFunction}
- * @name Highcharts.AnnotationShapePointOptions
+ * @typedef {
+ *          Highcharts.AnnotationMockPointOptionsObject|
+ *          Highcharts.AnnotationMockPointFunction
+ *     }Highcharts.AnnotationShapePointOptions
  */
 
 ''; // required by JSDoc parsing
--- a/ts/Extensions/Annotations/Annotations.ts
+++ b/ts/Extensions/Annotations/Annotations.ts
@@ -1365,45 +1365,6 @@ merge<Annotation>(
                  */
 
                 /**
-                 * The x position of the point. Units can be either in axis
-                 * or chart pixel coordinates.
-                 *
-                 * @type      {number}
-                 * @apioption annotations.labels.point.x
-                 */
-
-                /**
-                 * The y position of the point. Units can be either in axis
-                 * or chart pixel coordinates.
-                 *
-                 * @type      {number}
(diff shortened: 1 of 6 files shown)


raf18seb (approved):

pawelfus (changes requested):
Two things:
- see circleCI errors
- it sets correct types for `Shape`, but not for [`Label`](http://localhost:9005/highcharts/annotations.labels.point). Then perhaps the name should be more universal (`AnnotationShapePointFunction` -> `AnnotationMockPointFunction`) like it already has for object (`AnnotationMockPointObject`).

pawellysy (approved):

pawelfus (changes requested):
> - it sets correct types for Shape, but not for Label. Then perhaps the name should be more universal (AnnotationShapePointFunction -> AnnotationMockPointFunction) like it already has for object (AnnotationMockPointObject).

ts/Extensions/Annotations/Annotations.ts
```diff
@@ -1359,50 +1359,11 @@ merge<Annotation>(
                  *         Attach annotation to a mock point
                  *
                  * @declare   Highcharts.AnnotationMockPointOptionsObject
-                 * @type      {string|*}
+                 * @type      {string|Highcharts.AnnotationMockPointOptionsObject}
```
pawelfus, line 1362: This can also be `AnnotationMockPointFunction` (like in the new demo)

ts/Extensions/Annotations/Annotations.ts
```diff
@@ -1359,50 +1359,11 @@ merge<Annotation>(
                  *         Attach annotation to a mock point
```
pawelfus, line 1359: Add a new demo here too (it has an example for `labels.point` with functions)

ts/Extensions/Annotations/Annotations.ts
```diff
@@ -1419,21 +1380,27 @@ merge<Annotation>(
...
+                 *
+                 * This option is available
                  * for shapes which can use multiple points such as path. A
                  * point can be either a point object or a point's id.
                  *
                  * @see [annotations.shapes.point](annotations.shapes.point.html)
                  *
-                 * @declare   Highcharts.AnnotationMockPointOptionsObject
-                 * @type      {Array<string|*>}
+                 * @type      {Array<Highcharts.AnnotationShapePointOptions>}
```
pawelfus, line 1403: Missing `string` type (it has in the description above: `A point can be either a point object or a point's id.`

pawelfus (changes requested):

ts/Extensions/Annotations/Annotations.ts
bre1470: Then it can not declare Highcharts.AnnotationMockPointOptionsObject automatically. You have to define that interface yourself.

bre1470:
Honestly you should consider one callback function producing all points dynamically as an extra option like `pointGenerator` than trying to have an array mixing objects and functions. It just becomes messy right now.

pawelfus:
> Honestly you should consider one callback function producing all points dynamically as an extra option like `pointsGenerator` than trying to have an array mixing objects and functions. It just becomes messy right now.

Yes, that's a nice idea. Though this ticket and PR is only about documenting what we already have, so it's more about cleaning the mess :) Usually, you either have `[ { }, { }, ... ]` or `[ fun(), fun(), ... ]`, but if a user needs a mix of any combination (e.g. `[ {}, fun(), string, ... ]`) then it will work

pawelfus (approved):
Thank you! :)

TorsteinHonsi (approved):
Thanks!

PR #16113, opened 2021-07-28
bugfix/16112-polar-axis-array-setOptions

### LINKED ISSUE #16112: Polar axis ignores custom default options
#### Expected behaviour
Polar axis should respect options set with 'Highcharts.setOptions' in case when 'yAxis' definition is array.

#### Actual behaviour
Default options for yAxis are ignored.

https://jsfiddle.net/nc9ruy3x/

#### Product version
Latest

#### Affected browser(s)
All

khlieng:
Fixed #16112, axis options set by `setOptions` were not picked up by polar chart when chart axis options were set as an array.


TorsteinHonsi (approved):
Thanks!

PR #16117, opened 2021-07-29
docs/15596-sunburst-rotation-api-path

### LINKED ISSUE #15596: Sunburst rotation and rotationMode incorrectly documented in API
#### Actual behaviour
The API path is incorrect.
`sunburst.levels.rotation` suggests that the points in the sunburst chart can be rotated per level.

#### Expected behaviour
It is not about points, but **dataLabels**. The correct path should be `sunburst.levels.dataLabels.rotation`
The same for `rotationMode`.

https://jsfiddle.net/BlackLabel/6dhjmrne/

#### Product version
Highcharts 9.0.1

bm64:
Fixed #15596, removed unnecessary `levels.rotation` and `levels.rotationMode` API properties for sunburst.

### DIFF
--- a/ts/Series/Sunburst/SunburstSeries.ts
+++ b/ts/Series/Sunburst/SunburstSeries.ts
@@ -524,20 +524,6 @@ class SunburstSeries extends TreemapSeries {
          */
 
         /**
-         * Can set a `rotation` on all points which lies on the same level.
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (approved):
Thanks!

Issue #15876, opened 2021-06-10
DOM order should follow keyboardNavigation.order


oysteinmoseng:
#### Expected behaviour
Specific use case: If the legend is placed above the chart, setting [keyboardNavigation.order](https://api.highcharts.com/highcharts/accessibility.keyboardNavigation.order) will allow tab navigation order to change, but the DOM order should change as well. This impacts screen reader users navigating by virtual cursor.

#### Actual behaviour
The DOM order of proxy elements is fixed.

#### Live demo with steps to reproduce
https://jsfiddle.net/oysteinmoseng/nyv6as9f/
Navigating with virtual cursor, the order is data points first, then legend, then menu, while the Tab order is opposite.

oysteinmoseng:
This will be closed once PR #16124 is merged

PR #16124, opened 2021-08-01
A11y proxy refactoring

oysteinmoseng:
Fixed issue #15876, DOM order not following keyboardNavigation.order.
___
Pretty comprehensive rewrite of the proxy element logic. Now the a11y module has one central ProxyProvider that keeps track of the various groups, and can reorder them in the DOM properly (ref #15876). Previously the logic was disjointed, as each component kept track of its own group.

Fixed several edge case bugs/glitches in the process, and the new solution should be more robust in general, and easier to debug since critical logic is gathered in one place.


TorsteinHonsi (approved):
Thanks!

TorsteinHonsi (changes requested):
Please resolve conflicts.

PR #16128, opened 2021-08-02
bugfix/16126-a11y-full-screen-print-error

### LINKED ISSUE #16126: Print chart in the full screen mode - console error
The error occurs in the console after triggering the print event while the full-screen mode is applied. As a consequence full-screen mode is forced to be disabled but in the export menu the option to close still exists.

`accessibility.src.js:431 Uncaught TypeError: a.hasAttribute is not a function`

Demo: https://www.highcharts.com/demo/line-basic

Steps to reproduce:
1. Open export menu,
2. Click `View in full screen`,
3. Open export menu again,
4. Click `Print chart`,
5. Check the console and export menu.

khlieng:
Fixed #16126, printing chart in full screen with accessibility loaded threw.


TorsteinHonsi (approved):
Thanks!

PR #16145, opened 2021-08-04
bugfix/16143-hidden-boost-local-export

### LINKED ISSUE #16143: OOB exception when exporting SVG
#### Expected behaviour
SVG exporting works without generating an error.

#### Actual behaviour
Highcharts generates error 28 which is caused by an out of bounds exception in getSVGForLocalExport (offline-exporting.ts)

Cause: The for loop in getSVGForLocalExport doesn't correctly handle node removal. No adjustments is being made to i nor l to correctly account for the deletion.  

https://jsfiddle.net/72q3y68x/

- Load example and select download SVG option from export menu.
- Open developer console to see generated error. 

#### Product version

Highcharts

#### Affected browser(s)
Chrome

khlieng:
Fixed #16143, offline exporting hidden boosted series threw error 28.


TorsteinHonsi (approved):
Thanks!

TorsteinHonsi (changes requested):
Please resolve conflicts.

TorsteinHonsi (approved):

PR #16151, opened 2021-08-05
bugfix/14055-boost-with-odrinal 

### LINKED ISSUE #14055: setting ordinal = true is not working when I use boost.js in Highchart
ordinal property should remove the no data points in the series and working fine normally.

But if we include boost.js to improve the performance, the above said behavior is not working and displaying all the dates between the start and end date

Please find the below demo page in jsFiddle. Please remove/add boost.js to reproduce the issue. 

http://jsfiddle.net/rameshkan/7f4crsvL/5/

karolkolodziej:
Fixed #14055 ordinal axis and data grouping didn't work with the boost module.

Todo: 

- [x] POC
- [x] fix DG with boost
- [x] test
- [x] performance test
- [x] ~describe in API what has precedence over what~


ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -251,15 +251,15 @@ namespace OrdinalAxis {
                         }
                     }
 
-                    if (series.isSeriesBoosting) {
-                        hasBoostedSeries = true;
-                    }
+                    // if (series.isSeriesBoosting) {
```
karolkolodziej, line 254: Removing that enables the ordinal axis to be boosted properly however it has some negative impact on the performance. 
This could be checked in [the demo](http://jsfiddle.net/BlackLabel/ncv7gbh0/) (roughly 200ms increase in render time).
At that point, I was wondering If there should be some sort of mechanism which should decide whether to use grouping or boosting at that point? Or should it be included somewhere else?
I would be grateful for your opinions.

_DG is still to be fixed_

pawelfus: > Removing that enables the ordinal axis to be boosted properly however it has some negative impact on the performance.

Could you do some more tests?
- Is the ~200ms a constant increase, or it depends on the number of points? Could you compare 100k, 500k, and 2M?
- Does it increase the time when rendering the chart with `ordinal=false`?
- Does it increase the time when a dataset is regular (but with `ordinal=true`)?

> At that point, I was wondering If there should be some sort of mechanism which should decide whether to use grouping or boosting at that point? Or should it be included somewhere else?

In the current version, DG has a higher priority (it means when DG and Boost are enabled, DG will take precedence, so the developer must disable DG to enable Boost mode). I think we should keep this logic (there's no need to introduce this breaking change, right?)

karolkolodziej: > In the current version, DG has a higher priority (it means when DG and Boost are enabled, DG will take precedence, so the developer must disable DG to enable Boost mode). I think we should keep this logic (there's no need to introduce this breaking change, right?)

That makes sense! Having order what should be fired first would be enough.
It only needs a proper description in the API.

karolkolodziej: 1. **100k points**
- v9.1.2 - average rendering time: 220ms
- PR changes - average rendering time: 340ms

2. **500k points**
- v9.1.2 - average rendering time: 380ms
- PR changes - average rendering time: 600ms

3. **2000k points**
- v9.1.2 - average rendering time: 1600ms
- PR changes - average rendering time: 2400ms

<br>

> Does it increase the time when rendering the chart with ordinal=false?

No, it renders the same as the current version
<br>

> Does it increase the time when a dataset is regular (but with ordinal=true)?

Unfortunately yes, when `ordinal: true` and regular data set, the chart render takes more time (the same as for non-regular data).

pawelfus: > Unfortunately yes, when ordinal: true and regular data set, the chart render takes more time (the same as for non-regular data).

That's strange. In theory, a regular dataset should not trigger any of the ordinal methods, right? Could you check this?

karolkolodziej: I've tested it more thoroughly and my previous comment was a bit of an overstatement. 
Rendering regular data with `ordinal: true`, doesn't take as much time as non-regular data.

However, setting `ordinal: false` for ordinal data has a non-negligible impact on rendering time because of the few loops that are happening in the `beforeSetTickPositions` method in order to test if the data should be evenly distributed- check for `useOrdinal` property.

**To sum up, this PR doesn't affect the performance of the boosted stock chart with the regular data used.**

TorsteinHonsi: Does that mean that the processing is now slower, but only for ordinal axis? In that case it's fine, as boost with ordinal axis is a new feature.

TorsteinHonsi: No need for that I think, as we're actually adding support for boosted ordinal axis

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -984,15 +983,17 @@ namespace OrdinalAxis {
                         }
                     }
 
-                    if (series.isSeriesBoosting) {
-                        hasBoostedSeries = true;
-                    }
+                    // Enable boost for the ordinal axis, #14055.
+                    // if (series.isSeriesBoosting) {
+                    //     hasBoostedSeries = true;
+                    // }
```
pawelfus, line 989: I think there's no need to keep commented out code. If this will cause any kind of regression in the future, we will simply bisect. The same applies to the code below

pawelfus (changes requested):

PR #16154, opened 2021-08-05
bugfix/16153-chart-update-style-fontFamily-reset

### LINKED ISSUE #16153: chart.update could lead to fontFamily or fontSize reset to defaults
#### Expected behaviour

After calling chart.update correct values should be applied.

#### Actual behaviour

Currently if user calls chart.update with styles which is equal to current cleanRecursively call removes these values in chart.update and we lost these values, as a consequence fontFamily or fontSize could reset to defaults because they are hardcoded in getStyle method:

fontFamily: '"Lucida Grande", "Lucida Sans Unicode", Arial, Helvetica, sans-serif'
fontSize: '12px'

ts/Core/Chart/Chart.ts - update()
ts/Core/Renderer/SVG/SVGRenderer.ts - getStyle()

https://jsfiddle.net/g3p4ve6b/1/

#### Product version
Highcharts JS v9.1.2

#### Affected browser(s)
All

khlieng:
Fixed #16153, `fontFamily` reset when updating `chart.style`.

### DIFF
--- a/samples/unit-tests/chart/chart-update/demo.js
+++ b/samples/unit-tests/chart/chart-update/demo.js
@@ -146,6 +146,11 @@
         var chart = Highcharts.chart(
             $('<div>').appendTo('#container')[0],
             Highcharts.merge(getConfig(), {
+                chart: {
+                    style: {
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

Issue #15696, opened 2021-05-10
Stock Tools: Indicator on indicator


pawelfus:
# Description of the feature
Implement an option to choose an indicator as a base series (under the hood: `linkedTo` option). All new added technical indicators should get an ID, so we can use it later and display it in the dropdown:

<img width="597" alt="Zrzut ekranu 2021-05-10 o 17 36 15" src="https://user-images.githubusercontent.com/1453926/117685471-408bcc80-b1b6-11eb-9eec-a4161e93a14b.png">

Issue: How simple data technical indicators should behave when H/L/C values are required. 

# Library related to the feature
Highcharts Stock

# Proof of Concept/Live example for the feature
N/A

----

**You can vote for this feature by adding a thumbs-up reaction to this post.**

PR #16156, opened 2021-08-06
enhancement/15696-indicator-on-indicator

karolkolodziej:
Added support for calculating indicators on another indicator. See #15696.


pawelfus (changes requested):
Looks good, just one thing: http://jsfiddle.net/BlackLabel/q5hreoLj/1/ - add SMA, then edit SMA -> now you can choose SMA as a base for itself

ts/Extensions/Annotations/Popup.ts
```diff
@@ -1019,25 +1019,22 @@ H.Popup.prototype = {
             selectBox.setAttribute('id', PREFIX + 'select-' + optionName);
 
             // list all series which have id - mandatory for creating indicator
-            chart.series.forEach(function (serie): void {
+            chart.series.forEach(function (series): void {
+                seriesOptions = series.options as any;
```
raf18seb, line 1023: You are casting the whole `seriesOptions` variable to `any`, but only `seriesOptions.params` is throwing an error (other `seriesOptions.name` or `seriesOptions.id` are fine.

I would think about how we can get rid of `as any` totally or at least cast to `any` only `options.params` , not the whole variable (like previously, before your changes).

ts/Extensions/Annotations/Popup.ts
```diff
@@ -1019,25 +1019,22 @@ H.Popup.prototype = {
...
 
-                seriesOptions = serie.options;
+                if (seriesOptions.id !== PREFIX + 'navigator-series') {
+                    const seriesName = seriesOptions.params ?
+                        seriesOptions.name || series.name :
+                        seriesOptions.name || seriesOptions.id || '';
```
raf18seb, line 1028: Duplicated `seriesOptions.name`.

Isn't it better?:
```js
const seriesName = seriesOptions.name ||
    (seriesOptions as any).params ? series.name : seriesOptions.id || '';
```

raf18seb (changes requested):

pawellysy (approved):

test/cypress/integration/Highcharts/stock-tools.spec.js
```diff
@@ -34,3 +34,65 @@ describe('Stock Tools', () => {
...
+
+        cy.addIndicator(); // Add SMA indicator with period 20.
+
+        cy.window().then((win) => {
+            const H = win.Highcharts,
+                chart = H.charts[0];
+
+            // Select the first 3m period.
+            chart.xAxis[0].setExtremes(1565098200000, 1565098200000 + 36e5 *24 *90);
+        });
```
pawelfus, line 73: Can we use `cy.chart().then(chart => chart.xAxis[0].setExtremes(...)`?

pawelfus (approved):
Looks good, one question though ;)

pawelfus:
Edit: http://jsfiddle.net/BlackLabel/q5hreoLj/1/ - is is something with my build or is it still broken @karolkolodziej

karolkolodziej:
I guess it might be something on your end...
Add:
Screenshot 2021-08-10 at 11 59 13
Edit: 
Screenshot 2021-08-10 at 11 59 22

raf18seb (approved):
Thanks!

TorsteinHonsi (changes requested):
All good, except please resolve the conflicts.

TorsteinHonsi (approved):

PR #16157, opened 2021-08-06
bugfix/16146-sunburst-negative

### LINKED ISSUE #16146: Sunburst chart - Issue observed when drill down to first level.
Expected behavior - Chart should display properly

Actual behavior - Chart is broken after drill down

Live demo with steps to reproduce
https://jsfiddle.net/p0vgzr41/

Product version
9.1.2

Chart before drill down-
image

Chart after drill down-
image

khlieng:
Fixed #16146, negative values broke sunburst chart.


TorsteinHonsi (approved):
Thanks!

PR #16163, opened 2021-08-06
bugfix/16158-indicator-popup-default-series

### LINKED ISSUE #16158: Indicators Popup: Wrong defaults in series-dropdowns
#### Expected behaviour
1. When adding a technical indicator with multiple series links (e.g. ohlc + volume), by default, the first ("Series") dropdown should choose the main series, then the second one should select the "Volume" series (if exists):
Zrzut ekranu 2021-08-6 o 11 01 47

2. When editing a technical indicator with multiple series links, linked series should be selected by default:
Zrzut ekranu 2021-08-6 o 11 06 13

#### Actual behaviour
1. Always the first (main) series is selected.
2. The first dropdown selects always the "Volume" series (??), the second one selects the main series (??)

#### Product version
Highcharts Stock - Stock Tools module, v9.1.2

khlieng:
Fixed #16158, indicator popup had wrong series selected by default in some cases.

### DIFF
--- a/ts/Extensions/Annotations/Popup.ts
+++ b/ts/Extensions/Annotations/Popup.ts
@@ -141,7 +141,7 @@ declare global {
                 optionName: string,
                 chart: AnnotationChart,
                 parentDiv: HTMLDOMElement,
-                selectedOption: string
+                selectedOption?: string
             ): void;
         }
         interface PopupTabsObject {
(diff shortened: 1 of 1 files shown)


pawelfus (changes requested):
Looks good, thanks!
Could you also create related Cypress tests?

raf18seb (commented):
What if the main series and the volume series are both `column` series? For example OBV which doesn't require OHLC data

TorsteinHonsi (approved):

TorsteinHonsi (changes requested):
Pending approval by @pawelfus

pawelfus (approved):
Thanks!

PR #16178, opened 2021-08-11
Migration: Announcer

marvin19:
Removed global namespace from Announcer.

### DIFF
--- a/ts/Accessibility/Utils/Announcer.ts
+++ b/ts/Accessibility/Utils/Announcer.ts
@@ -12,10 +12,17 @@
 
 'use strict';
 
+/* *
+ *
+ *  Imports
+ *
+ * */
+
 import type Chart from '../../Core/Chart/Chart';
 import type {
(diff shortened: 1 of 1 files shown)


ts/Accessibility/Utils/Announcer.ts
```diff
@@ -115,6 +137,20 @@ class Announcer {
     }
 }
 
-H.Announcer = Announcer;
+/* *
```
bre1470, line 140: It got removed, but not set in masters.

bre1470 (changes requested):

ts/Accessibility/Utils/Announcer.ts
bre1470: Sorry, it is okay. Not a public class.

bre1470 (approved):

TorsteinHonsi (approved):

PR #16181, opened 2021-08-11
feature/15825-new-annotation-fibonacci-timezones

raf18seb:
Added new annotation Fibonacci Time Zones. See #15825.


ts/Extensions/Annotations/Types/FibonacciTimezones.ts
```diff
@@ -0,0 +1,292 @@
...
+    ): Function {
+        return function (target: any): PositionObject {
+            let annotation = target.annotation,
+                points = annotation.points,
+                // Offset between the first and the second line
+                deltaX = points.length > 1 ? points[1].x - points[0].x : 0,
+                // firstLine.x + fibb * offset
+                x = points[0].x + fibonacciIndex * deltaX;
+
+            // Horizontal and vertical lines have only one point,
```
raf18seb, line 68: Improve the comment

raf18seb (commented):
To to

karolkolodziej (changes requested):
Few smaller questions/improvements.

- [x] Additionally, I spotted an issue while the chart is being inverted, the control point isn't placed correctly. 
I was wondering if this should be even a concern for us?

samples/highcharts/annotations-advanced/fibonacci-timezones/demo.js
```diff
@@ -0,0 +1,60 @@
...
+
+    // Create the chart
+    Highcharts.stockChart('container', {
+
+        chart: {
+            events: {
+                load: function () {
+                    this.annotations.forEach(function (annotation) {
+                        annotation.setControlPointsVisibility(true);
+                        annotation.cpVisibility = true;
```
pawelfus, line 12: Do we need this `cpVisibility` property? Perhaps it would be better to make this a variable, so developers won't get the wrong idea that `cpVisibility` can change anything

pawelfus (changes requested):
Additional notes:
- [ ]  as discussed, please change from `timezones` to `time-zones`/`timeZones` 
- [ ] [demo](http://jsfiddle.net/BlackLabel/ouz035ts/1/), steps: click on the annotation, there's `undefined` in the mini-popup. Then click `edit` icon (pencil), we can set `fill` color (which is not possible), but can't change line's color (which should be possible)
- [ ] please merge with master, so we can test docs

After changes will continue review :)

ts/Extensions/Annotations/Types/Tunnel.ts
```diff
@@ -181,8 +181,6 @@ Tunnel.prototype.defaultOptions = merge(
      */
     {
         typeOptions: {
-            xAxis: 0,
-            yAxis: 0,
             /**
```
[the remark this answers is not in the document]
pawelfus: I had the same question, and probably the answer is: it's inherited from the CrookedLine, but let's wait for Rafał's answer :)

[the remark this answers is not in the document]
raf18seb: Yep, exactly :)

ts/Extensions/Annotations/Types/FibonacciTimezones.ts
```diff
@@ -0,0 +1,292 @@
...
+     *  Functions
+     *
+     * */
+
+    public addShapes(): void {
+        let fibb = 1,
+            nextFibb = 1,
+            temp: number;
+
+        for (let i = 0; i < 11; i++) {
```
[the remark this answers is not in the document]
raf18seb: I had the same concern. I would leave it hardcoded for now and, when users want to change the number of lines (more than 11), then make it configurable.
What do you think?

I will add a comment for sure describing what `11` means.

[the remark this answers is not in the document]
karolkolodziej: For me, it could be hardcoded the only thing I would change is a description of that property in case if in the future we might change it into something more configurable. ;)

ts/Extensions/Annotations/Types/FibonacciTimezones.ts
```diff
@@ -0,0 +1,292 @@
...
+        };
+    }
+
+    /* *
+     *
+     *  Constructors
+     *
+     * */
+
+    public constructor(chart: Highcharts.AnnotationChart, options: FibonacciTimezones.Options) {
```
pawellysy, line 106: useless constructor. It does the same, as the superior class

ts/Extensions/Annotations/Types/FibonacciTimezones.ts
```diff
@@ -0,0 +1,292 @@
...
+            const points = [
+                // Correct the first line position (we cannot correct the fibb
+                // variable because it's needed for further iterations)
+                FibonacciTimezones.edgePoint(1, 0, !i ? 0 : fibb),
+                FibonacciTimezones.edgePoint(0, 1, !i ? 0 : fibb)
+            ];
+
+            temp = nextFibb;
+            nextFibb = fibb + nextFibb;
+            fibb = temp;
```
pawellysy, line 131: you can replace it with:
```
nextFibb = fibb + nextFibb
fibb = nextFibb - fibb;
```

pawellysy (changes requested):
Please add cypress tests

ts/Extensions/Annotations/Types/FibonacciTimezones.ts
```diff
@@ -0,0 +1,292 @@
...
+
+            this.initShape(
+                merge(
+                    this.options.typeOptions.line,
+                    {
+                        type: 'path',
+                        points: points
+                    }
+                ),
+                false as any
```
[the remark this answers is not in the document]
raf18seb: I corrected it in all annotations types, it was a general problem. I got rid of `as any`'s and linted those parts of code.
FYI @pawelfus

samples/highcharts/annotations-advanced/fibonacci-timezones/demo.js
raf18seb: Sorry, could you elaborate or explain in other words?

ts/Extensions/Annotations/Types/FibonacciTimezones.ts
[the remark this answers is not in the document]
karolkolodziej: Could you please, for clarity, describe the `index` parameter in the `Annotation.ts` file where the `public initShape` method is declared?

ts/Stock/StockToolsGui.ts
```diff
@@ -284,6 +285,7 @@ setOptions({
                 verticalLabel: 'Vertical label',
                 verticalArrow: 'Vertical arrow',
                 fibonacci: 'Fibonacci',
+                fibonacciTimeZones: 'Fibonacci Time Zones',
```
karolkolodziej, line 288: Is it `Time Zone` or `Time Zones` (plural)? I can see different variants.

@pawelfus  Could you please help with the decision?
IMO it should be `Fibonacci Time Zones`

After the decision, please unify it everywhere.

ts/Extensions/Annotations/Types/FibonacciTimeZones.ts
```diff
@@ -0,0 +1,306 @@
...
+     *  Functions
+     *
+     * */
+
+    public addShapes(): void {
+        const numberOfLines = 11;
+        let fibb = 1,
+            nextFibb = 1;
+
+        for (let i = 0; i < numberOfLines; i++) {
```
karolkolodziej, line 134: No hard opinion here just a proposal to consider ;)

What if you start that loop from `1` and handle the first point outside that loop that you could avoid the `correctedFibb`

karolkolodziej (changes requested):

ts/Stock/StockToolsGui.ts
raf18seb: Plural, I'll look for them and correct, thanks

ts/Extensions/Annotations/Types/FibonacciTimeZones.ts
raf18seb: I considered few approaches here and I think this one is short and easy. Maybe there is another one, but as long as it's not simpler/shorter, I leave it as it is now.

ts/Stock/StockToolsGui.ts
karolkolodziej: Remember about changing the name of the `fibonacci-timezone.svg` file ;)

raf18seb: Seems like only the `svg` icon is not in the plural form

samples/highcharts/annotations-advanced/fibonacci-timezones/demo.js
pawelfus: Sure! Is `Annotation.cpVisibility` necessary? Or can it be a simple variable: `let cpVisibility`?

pawellysy (approved):
Looks good

karolkolodziej (changes requested):
So far so good.
But still, a few things missing:
- [x]  Cypress
- [x] . Edit popup (`fill` doesn't make sense and the `line color` is missing)
- [ ]  Should it be possible to use this annotation with the inverted chart or we skip this part right now?

pawelfus (changes requested):
Looks good thanks!
Could you also please add the `@since next` to the doclets?

pawellysy (approved):
Looks good! 👍

ts/Extensions/Annotations/Types/FibonacciTimezones.ts
```diff
@@ -0,0 +1,292 @@
...
+                FibonacciTimezones.edgePoint(1, 0, !i ? 0 : fibb),
+                FibonacciTimezones.edgePoint(0, 1, !i ? 0 : fibb)
+            ];
+
+            temp = nextFibb;
+            nextFibb = fibb + nextFibb;
+            fibb = temp;
+
+            // Save the second line for the control point
+            if (i === 1) {
```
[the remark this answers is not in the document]
pawelfus: By definition, Fibonacci time zones define the starting point and the range. Then all corresponding zones are calculated as Fibonacci sequence. That makes sense to render the control point at the second vertical line. Of course, if clients in the future would ask for a different solution, we may change it

pawelfus (changes requested):
> Edit popup (fill doesn't make sense and the line color is missing)

This is still broken

- [x] [demo](http://utils.highcharts.local/samples/#view/stock/demo/stock-tools-gui) - add timezones (at the beginning of the dataset), zoom in on the right side (quite a lot). Try drag&drop on the line. Annotation disappears. Most probably related to the cropped data and trying to calculate the zones. Perhaps only the first and the second lines should be draggable?


- [ ] It seems the distance between lines is calculated in xAxis values. Is this configurable? Current/default behaviour seems a bit strange, [demo](http://utils.highcharts.local/samples/#view/stock/demo/stock-tools-gui): 
- click `1m` button, add time zones. First point: 3rd of Aug, second point: 5th of Aug. Distance between 3rd and 4th lines is shorter than the 2nd and the 3rd one.
- the same steps as above, then drag&drop the annotation. Lines jump around

Using a non-ordinal axis probably won't cause this effect

test/cypress/integration/Highcharts/annotations/fib-time-zones/timezones.spec.js
```diff
@@ -0,0 +1,110 @@
+describe('Stock Tools Fibonacci Tome Zones, #15825', () => {
+    beforeEach(() => {
+        cy.viewport(1000, 500);
```
karolkolodziej, line 3: After fixing the issue with flicking annotation on the ordinal axis please add a test.

karolkolodziej (changes requested):

TorsteinHonsi:
Pending resolution to Pawel and Karol's comments.

raf18seb:
> Additional notes:
> 
> * [ ]  [demo](http://jsfiddle.net/BlackLabel/ouz035ts/1/), steps: click on the annotation, there's `undefined` in the mini-popup. Then click `edit` icon (pencil), we can set `fill` color (which is not possible), but can't change line's color (which should be possible)

It's a general issue. See an official `pitchfork` [demo](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/annotations-advanced/pitchfork/) (add these popup modules):
```html
<link rel="stylesheet" type="text/css" href="https://code.highcharts.com/css/stocktools/gui.css">
<link rel="stylesheet" type="text/css" href="https://code.highcharts.com/css/annotations/popup.css">
```

ts/Stock/StockToolsBindings.ts
```diff
@@ -2079,6 +2079,79 @@ const stockToolsBindings: Record<string, Highcharts.NavigationBindingsOptionsObj
...
+                    typeOptions: {
+                        xAxis: coordsX.axis.options.index,
+                        yAxis: coordsY.axis.options.index,
+                        points: [{
+                            x: x
+                        }, {
+                            x: coordsX.value
+                        }]
+                    }
+                });
```
pawellysy, line 2151: From what I see, you are only translating point. Don't you wanna use the `annotation.translatePoint()` method?

pawellysy (changes requested):

ts/Extensions/Annotations/Types/FibonacciTimeZones.ts
```diff
@@ -83,11 +86,19 @@ class FibonacciTimeZones extends CrookedLine {
     ): Function {
         return function (target: any): PositionObject {
             const annotation = target.annotation;
+
             let points = annotation.points;
             // Offset between the first and the second line
-            const deltaX = points.length > 1 ? points[1].x - points[0].x : 0,
+
+            const xAxis = points[0].series.xAxis,
```
karolkolodziej, line 93: Because of that nasty casting `target: any` we are losing all TS here.
Are you sure that `points` always exist?

ts/Extensions/Annotations/Types/FibonacciTimeZones.ts
```diff
@@ -83,11 +86,19 @@ class FibonacciTimeZones extends CrookedLine {
...
+                // Distance between the two first lines in pixels
+                deltaX = points.length > 1 ?
+                    xAxis.toPixels(points[1].x) - xAxis.toPixels(points[0].x) :
+                    0,
                 // firstLine.x + fibb * offset
-                x = points[0].x + fibonacciIndex * deltaX;
+                x = xAxis.toValue(
+                    xAxis.toPixels(points[0].x) + fibonacciIndex * deltaX
```
karolkolodziej, line 100: Can't you use `points[0].plotX` instead of `toPixels` calculation?

karolkolodziej (changes requested):
To do:
- [ ] check if improving TS is possible
- [ ] simplify the `x` calculation

ts/Stock/StockToolsBindings.ts
raf18seb: No, sorry. I'm not translating the point, I'm updating the distance between 2 lines and then translating all the lines here https://github.com/highcharts/highcharts/blob/25c57a09fe2b0474ef6dc0c69a60bbb364fe991a/ts/Extensions/Annotations/Types/FibonacciTimeZones.ts#L269

raf18seb: If I used `translatePoint()` here, I'd probably move (translate) only the first line.

pawelfus (approved):
Thank you!

ts/Extensions/Annotations/Types/FibonacciTimeZones.ts
raf18seb: This is a huge TS refactoring. This should be done not only for the FibonacciTimeZones annotation, but for all others too. In my opinion, TS refactor here should be in a separate ticket. What do you think?

karolkolodziej: I'm ok with that ;) 
It seems like all other annotations have the same TS issue and they are working fine. So let's leave it for now.

karolkolodziej (approved):

pawellysy (approved):
Nothing to add!

TorsteinHonsi (approved):
Thanks!

PR #16182, opened 2021-08-11
Migration: ChartUtilities

marvin19:
Removed global namespace from ChartUtilties.


ts/Accessibility/Accessibility.ts
```diff
@@ -86,7 +87,7 @@ declare global {
             options: Required<SeriesOptions>;
             points: Array<AccessibilityPoint>;
         }
-        let A11yChartUtilities: A11yChartUtilities;
+        let A11yChartUtilities: ChartUtilties.A11yChartUtilities;
```
bre1470, line 90: Here fix the import name and use `typeof ChartUtilities`.

ts/Accessibility/Utils/ChartUtilities.ts
```diff
@@ -411,8 +391,33 @@ function scrollToPoint(point: Point): void {
...
+            chart: Chart,
+            element: DOMElementType
+        ): void;
+        hideSeriesFromAT(series: Series): void;
+        scrollToPoint(point: Point): void;
+    }
+}
 
-const ChartUtilities: Highcharts.A11yChartUtilities = {
+const ChartUtilities: ChartUtilities.A11yChartUtilities = {
```
bre1470, line 420: Namespace with `A11yChartUtilities` can be removed. Leave the type of the `ChartUtilities` constant blank.

bre1470 (changes requested):

bre1470 (approved):
Thank you! Looks good. 👍

PR #16199, opened 2021-08-16
bugfix/missing-series-in-demo

karolkolodziej:
Added missing modules for the new series into the stock tools GUI demo.

After adding the new series `heikinashi` and `hollowcandlestick` they are missing in the stock tools GUI demo.


pawelfus (approved):

TorsteinHonsi (approved):
Thanks!

PR #16202, opened 2021-08-16
Docs: Added article about ES6 core modules and tree shaking.

bre1470:


TorsteinHonsi (changes requested):

TorsteinHonsi (changes requested):
Better, but compiled+gzipped size is what matters for production. You can retrieve gzipped size by `gzip -k -9 code/highcharts.js`

TorsteinHonsi (approved):

PR #16205, opened 2021-08-16
bugfix/candlestick-point-up-color 

karolkolodziej:
Fixed the CSS fill property of bullish candlestick and the hollow candles.


TorsteinHonsi (approved):
Thanks!

PR #16206, opened 2021-08-16
bugfix/ordinal-grouping-visual-differences 

karolkolodziej:
Fixed how the ordinal axis is being calculated after grouping.

In cases when ordinal enabled and data grouped with evenly spaced points (days/weeks), the ordinal calculation should be forced.


TorsteinHonsi (approved):
Thanks!

PR #16211, opened 2021-08-17
docs/blog-article-calendar-demo

raf18seb:
Added demo to the blog article about using Stock with a calendar.


TorsteinHonsi (approved):
Thanks!

Issue #16017, opened 2021-07-08
Ability to change the orientation(aka direction) of scrollbar to left side in case of RTL lagnuagues.


sisaki-n:
# Description of the feature
When we use Bar, Heatmap charts and loading in RTL languages, the scrollbar is always shown on the right side of the chart instead of left handside. which is not in sync with regular scroll bar of html controls like div.

# Library related to the feature
 Highcharts / Highcharts Stock 

# Proof of Concept/Live example for the feature

In this chart, even if you set the x axis as opposite to true and y-axis reversed to true. required RTL changes, scroll bar is always shown on left side and there is no provision to change the direction or orientation of the scroll bar.
https://jsfiddle.net/gh/get/library/pur ... scrollbar/

POC: By updating the value of the Transform property of g element inside the xaxis scrollbar as below in both load and reflow events in chart options. it worked. 

```js
load: function () {
    if (isRTL) {
        this.xAxis[0].scrollbar.group.element.attributes["transform"].value = "translate(0,10)";
        this.xAxis[0].scrollbar.group.element.attributes["transform"].nodeValue = "translate(0,10)";
        this.xAxis[0].scrollbar.group.element.attributes["transform"].textContent = "translate(0,10)";
    }
},
```
by making xvalue of translate argument as 0 , scrollbar position got updated from right to left.

but it should have been provided in some API property which enables the developers to get reliable solution.

**You can vote for this feature by adding a thumbs-up reaction to this post.**

raf18seb:
Hi @sisaki-n, thank you for creating the ticket. Your jsFiddle URL doesn't seem to work. I created a simple demo: https://jsfiddle.net/BlackLabel/obfz8mdt/

xAxis scrollbar is available only in Highcharts Stock.

PR #16214, opened 2021-08-17
enhancement/16017-scrollbar-position 

karolkolodziej:
Enhancement #16017, added the possibility to position the `scrollbar` on the left.


ts/Core/Axis/ScrollbarAxis.ts
```diff
@@ -21,6 +21,7 @@ import type Scrollbar from '../Scrollbar';
 import type ScrollbarOptions from '../ScrollbarOptions';
 
 import U from '../Utilities.js';
+import Chart from '../Chart/Chart.js';
```
karolkolodziej, line 24: Unused- remove

pawellysy (approved):
Looks good!

pawelfus (changes requested):
- how about calling the option `opposite` (like e have for the Axis)? By default `opposite = true` (current behaviour), and setting `opposite = false` would translate the scrollbar: to the left (vertical Axis) position, and to the top (horizontal Axis, in the future if requested)
- xAxis labels in the inverted chart overlap the scrollbar: http://jsfiddle.net/BlackLabel/a76oxdj9/2/
- please, create a demo

karolkolodziej:
1. I agree, `opposite` sounds better and it's more intuitive. Additionally, it might be easier to scale it in the future.
2. When the `marginLeft` is set on the chart, it has a higher priority than the `scrollbar` margin calculation and I think it should stay like this.
3. Done :)

raf18seb (approved):

pawelfus:
> When the marginLeft is set on the chart, it has a higher priority than the scrollbar margin calculation and I think it should stay like this.

Nice catch! I missed the `merginLeft` option. Agree 👍

ts/Core/ScrollbarDefaults.ts
```diff
@@ -118,6 +118,21 @@ const ScrollbarDefaults: ScrollbarOptions = {
...
+     * which positions the scrollbar on the left.
+     *
+     * @sample stock/yaxis/scrollbar-opposite/
+     *         A scrollbar not on the opposite side
+     *
+     * @type      {boolean}
+     * @default   true
+     * @apioption scrollbar.opposite
+     */
+    opposite: true,
```
pawelfus, line 134: That's a little tricky. Now all docs, [`xAxis`](http://localhost:9005/highstock/xAxis.opposite), [`yAxis`](http://localhost:9005/highstock/yAxis.opposite), and [`scrollbar`](http://localhost:9005/highstock/scrollbar.opposite) share this option, while only vertical Axis (yAxis in non-inverted and xAxis in the inverted chart) supports. I think this option should be removed from `scrollbar`, while `xAxis` and `yAxis` should mention when it works.

pawelfus (changes requested):
Just a minor thing, related to dosc. Sorry I missed that in the previous review. Apart from that - looks great!

pawelfus:
Extra note: Missing `@since next` in the doclets

ts/Core/ScrollbarDefaults.ts
karolkolodziej: Could you please elaborate because I don't fully get it?

> Now all docs, xAxis, yAxis,

The linked docs lead to `xAxis.opposit` which is a completely different parameter.

> while xAxis and yAxis should mention when it works.

Where this should be mentioned?

pawelfus: > The linked docs lead to xAxis.opposit which is a completely different parameter.

My apologies! Should be `xAxis.scrollbar.opposite`, `xAxis.scrollbar.opposite` and `scrollbar.opposite`.

> Where this should be mentioned?

In the doclet, I think

TorsteinHonsi:
Pending solution to Pawel's comments and fixing the Firefox test (use `assert.close` if pixel precision is the problem).

samples/unit-tests/scrollbar/positions/demo.js
```diff
@@ -1,296 +1,325 @@
...
-                },
-                series: [
-                    {
-                        data: [4, 20, 100, 5, 2, 33, 12, 23]
-                    }
-                ]
-            },
-            chart = Highcharts.chart('container', options),
-            position = chart.xAxis[0].scrollbar.group.translateY;
+// QUnit.test(
```
pawelfus, line 1: I guess this test shouldn't be commented out? ;)

ts/Core/ScrollbarDefaults.ts
pawelfus: > I think this option should be removed from `scrollbar`, while `xAxis` and `yAxis` should mention when it works.

Now we have this option defined in `scrollbar` only :)

pawelfus (changes requested):

pawelfus (approved):
Thanks!

TorsteinHonsi (approved):

PR #16215, opened 2021-08-17
Update of Google Sheets API to v4 (temporary fix)

apumapho:
This is an incomplete and rushed patch to update Google Sheets API to v4.  v3 was deprecated as of August 2, 2021 and users are beginning to receive failure/404 notices when trying to render highcharts that rely on legacy Google Sheets API data.

The v4 API requires an API key to be created at [API Services / Credentials](https://console.cloud.google.com/apis/credentials) and then added in your chart config.  Note also that `googleSpreadsheetWorksheet` has changed from a number to the name of the sheet.

```
data: {
            googleApiKey: 'YOUR_apiKey',
            googleSpreadsheetKey: 'YOUR_spreadsheetKey',
            googleSpreadsheetWorksheet: 'YOUR_sheetName'
}
```

TODO:
- there was code in the v3 version that limited to the data to start/endRow and start/endColumn which, if needed by users, should be reconsidered to instead use the [RANGE api now available](https://developers.google.com/sheets/api/samples/reading)
- there was code in the v3 version that tried to cast numbers as such which may be needed for some users depending on how they configured their charts; we could add some checks to see if a number is truly a number, or could [expand how the data is requested to format numbers/dates](https://developers.google.com/sheets/api/reference/rest/v4/spreadsheets.values/get)

Hope this helps people whose charts are broken with a quick fix!

### DIFF
--- a/ts/Extensions/Data.ts
+++ b/ts/Extensions/Data.ts
@@ -103,6 +103,7 @@ declare global {
             endRow?: number;
             error?: Function;
             firstRowAsNames?: boolean;
+            googleApiKey?: string;
             googleSpreadsheetKey?: string;
             googleSpreadsheetWorksheet?: string;
             itemDelimiter?: string;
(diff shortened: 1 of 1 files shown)


TorsteinHonsi:
Thanks for the PR! We will look into it.

TorsteinHonsi:
Thanks again for the inspiration! I've created an inhouse PR that allows us to run CI tests and stuff, see https://github.com/highcharts/highcharts/pull/16232 . Closing this.

PR #16221, opened 2021-08-18
bugfix/hollowcandlestick-line-color

karolkolodziej:
Fixed how the color is applied for bullish hollow candlestick line.


pawelfus (approved):
Thanks!

raf18seb (changes requested):
Seems ok, thanks!
Shouldn't we add some tests for the point's colors? Or the visuals are enough?

karolkolodziej:
At the beginning of creating the new series, we agreed that the visual test is enough and I think we shouldn't add more tests.

TorsteinHonsi:
I agree that the visuals are enough for the general color handling. Merging now. Thanks for the fix!

PR #16240, opened 2021-08-23
hotfix/docs-config-compare-cumulative

raf18seb:
Fixed docs error on master (caused by the compare and cumulative articles).

### DIFF
--- a/docs/doc-config.js
+++ b/docs/doc-config.js
@@ -6,7 +6,9 @@ module.exports = {
      * Refer to doc by relative path without extension, i.e.: 'maps/drilldown'
      */
     unlisted: [
-        'export-module/legacy-export-servers'
+        'export-module/legacy-export-servers',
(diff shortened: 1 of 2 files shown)


docs/doc-config.js
```diff
@@ -6,7 +6,9 @@ module.exports = {
      * Refer to doc by relative path without extension, i.e.: 'maps/drilldown'
      */
     unlisted: [
-        'export-module/legacy-export-servers'
+        'export-module/legacy-export-servers',
+        'stock/compare',
+        'stock/cumulative-sum'
```
TorsteinHonsi, line 11: Are you sure you want them as unlisted? It means they are not reachable from the navigation menu...

raf18seb: Of course...  corrected

TorsteinHonsi (approved):

PR #16241, opened 2021-08-23
bugfix/16238-reset-datagrouping

### LINKED ISSUE #16238: Regression: Data grouped tooltip not reset
#### Expected behaviour
The tooltip should reflect the current data grouping

#### Actual behaviour
When zooming to an un-grouped level, the tooltip suggests still grouped.

https://jsfiddle.net/highcharts/5o9bqer8/2/
Click "1M" to ungroup the data. Observe that the tooltip still says we're looking at a week. Click the point. The point still holds `dataGroup` information. 

The same demo in v9.1: https://jsfiddle.net/highcharts/5o9bqer8/3/ . No group tooltip or `Point.dataGroup`.

#### Product version
v9.1.0

#### Affected browser(s)
All

raf18seb:
Fixed #16238, a regression causing data grouping information not to reset when zooming in, both in the tooltip and in internal properties.

### DIFF
--- a/docs/doc-config.js
+++ b/docs/doc-config.js
@@ -6,7 +6,9 @@ module.exports = {
      * Refer to doc by relative path without extension, i.e.: 'maps/drilldown'
      */
     unlisted: [
-        'export-module/legacy-export-servers'
+        'export-module/legacy-export-servers',
+        'stock/compare',
+        'stock/cumulative-sum'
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (approved):
Thanks! Approved, just waiting for CI tests to finalize.

PR #16242, opened 2021-08-23
bugfix/16231-export-styledMode

### LINKED ISSUE #16231: export & styled mode
#### Expected behaviour

The last properly working version is 9.1.2: https://jsfiddle.net/BlackLabel/6qhgfLk4/

#### Actual behaviour

It seems that the styled config is not applied in the exported chart.

Demo: https://www.highcharts.com/demo/styled-mode-column

khlieng:
Fixed #16231, a regresssion in v9.2 causing exporting not to work with `styledMode` enabled.


TorsteinHonsi (approved):
Thanks!

PR #16226, opened 2021-08-19
Migration: Earcon

marvin19:
Migrated Earcon to ES6 pattern. This needs [this PR](https://github.com/highcharts/highcharts/pull/16222) to be merged first.

PR #16243, opened 2021-08-23
Migration: Instrument

marvin19:
Migrated Instrument to ES6 pattern. This needs [ChartSonify](https://github.com/highcharts/highcharts/pull/16222) and [Earcon](https://github.com/highcharts/highcharts/pull/16226) to be merged first.

### DIFF
--- a/ts/Extensions/Sonification/Instrument.ts
+++ b/ts/Extensions/Sonification/Instrument.ts
@@ -28,78 +28,654 @@ const {
     pick,
     uniqueKey
 } = U;
+/* eslint-disable no-invalid-this, valid-jsdoc */
+
+/* *
+ *
+ *  Class
+ *
+ * */
 
 /**
- * Internal types.
- * @private
+ * The Instrument class. Instrument objects represent an instrument capable of
+ * playing a certain pitch for a specified duration.
+ *
+ * @sample highcharts/sonification/instrument/
+ *         Using Instruments directly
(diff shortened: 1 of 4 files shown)


ts/Extensions/Sonification/Instrument.ts
```diff
@@ -28,79 +28,655 @@ const {
...
-            public setPan(panValue: number): void;
-            public stop(
-                immediately: boolean,
-                onStopped?: Function,
-                callbackData?: unknown
-            ): void;
+class Instrument {
+
+    // Default options for Instrument constructor
+    public static defaultOptions: Instrument.Options = {
```
bre1470, line 59: Section comment missing.

ts/Extensions/Sonification/Instrument.ts
```diff
@@ -28,79 +28,655 @@ const {
...
 
+/* *
+ *
+ *  Class Namespace
+ *
+ * */
+
+namespace Instrument {
+
+    export let audioContext: AudioContext;
```
bre1470, line 640: This is not type safe. Change to optional static class property.

bre1470 (changes requested):

ts/Extensions/Sonification/Instrument.ts
bre1470: Sorry, my mistake. The init function exits with error, if the context is missing. So this is fine.

bre1470 (approved):

PR #16252, opened 2021-08-24
bugfix/16176-funnel-dataLabel-rotation

### LINKED ISSUE #16176: Data labels rotation throws console error when value is other than 0
Hi Team,

Whenever we are giving data label's rotation with any value(other than 0) throws errors in the console and doesn't behave as expected. And when data label is long and if rotation value is provided, then it doesn't show any data labels.

JSFiddle example:-
[https://jsfiddle.net/KrishnaPriyaE/h06L3vzg/2/](https://jsfiddle.net/KrishnaPriyaE/h06L3vzg/2/)

We have used highcharts V8.x in an angular project.

All browser have the same issue.

khlieng:
Fixed #16176, funnel data labels with `rotation` set showed console error.


TorsteinHonsi (approved):
Thanks!

PR #16257, opened 2021-08-25
bugfix/16150-columnpyramid-inverted-yAxis-width

### LINKED ISSUE #16150: The columnpyramid series isn't working with yAxis width and inverted
Hi, 

I posted this issue in Highcharts Forum. So  Higcharts Developer @K009   suggested that I post it here to have a closer look into it. 

We tried  to get a vertical version of this:
https://jsfiddle.net/Theresa_Joy/4x6rfom9/16/

We tried to get it done, and this is our result.
https://jsfiddle.net/Theresa_Joy/j270L9y5/36/

There seems to be no script error, but we only see kind of bars with a slight angle instead of pyramids.
We checked this in Chrome, Mozilla and Edge.

Reply of @K009 
It happens because the axis is smaller (eg. 25% instead of 100%) and the points don't have enough place to be displayed, which results in no angled corners. Simplified demo with this issue: https://jsfiddle.net/BlackLabel/o390gwje/

I think that it might be a bug. Could you report it on our GitHub so we could take a closer look at it?
Here's a link: https://github.com/highcharts/highchart ... new/choose

**Can you pls. check it out at the earliest and let us know.** This is one very important feature we want to have, for our graphs. Since Highcharts has already an inbuilt Column Pyramid Chart, it was a very good candidate for our choice.

bm64:
Fixed #16150, pyramid shape was incorrect for inverted charts when `yAxis.width` set.

### DIFF
--- a/samples/unit-tests/series-columnpyramid/basic/demo.js
+++ b/samples/unit-tests/series-columnpyramid/basic/demo.js
@@ -54,3 +54,51 @@ QUnit.test('Column pyramid series - 0 dataLabel #12514', function (assert) {
         'Label should be translated and had different position translate attributes than starting values #12514'
     );
 });
+
+QUnit.test(
+    "Column pyramid series inverted - yAxis width #16150",
+    function (assert) {
+        const chart = Highcharts.chart("container", {
+            chart: {
+                type: "columnpyramid",
+                inverted: true,
+                width: 600,
+                animation: false,
+                height: 600
+            },
+            yAxis: {
+                width: "40%"
+            },
+
+            series: [
+                {
+                    data: [5, 4, 2]
+                }
+            ]
+        });
+
+        var roundPath = function (dArray) {
+            return dArray.map(item =>
+                Highcharts.map(item, function (value) {
+                    var number = Math.round(value);
+                    return Highcharts.isNumber(number) ? number : value;
+                })
+            );
+        };
+
+        const validPath = [
+            ["M", 370, 37],
+            ["L", 370, 37],
+            ["L", 335, 223],
+            ["L", 406, 223],
+            ["Z"]
+        ];
+
+        const testedPath = roundPath(chart.series[0].points[0].shapeArgs.d);
+        assert.deepEqual(
+            testedPath,
+            validPath,
+            "testedPath should be equal to validPath, #16150"
+        );
+    }
+);
--- a/samples/unit-tests/series-columnpyramid/basic/demo.js
+++ b/samples/unit-tests/series-columnpyramid/basic/demo.js
@@ -56,49 +56,37 @@ QUnit.test('Column pyramid series - 0 dataLabel #12514', function (assert) {
 });
 
 QUnit.test(
-    "Column pyramid series inverted - yAxis width #16150",
-    function (assert) {
-        const chart = Highcharts.chart("container", {
+    'The columnpyramid series inverted with yAxis.width, #16150',
+    assert => {
+        const chart = Highcharts.chart('container', {
             chart: {
-                type: "columnpyramid",
-                inverted: true,
-                width: 600,
-                animation: false,
-                height: 600
+                type: 'columnpyramid',
+                inverted: true
             },
             yAxis: {
-                width: "40%"
+                width: '40%'
             },
-
-            series: [
-                {
-                    data: [5, 4, 2]
-                }
-            ]
+            series: [{
+                data: [5, 4, 2]
+            }]
         });
 
-        var roundPath = function (dArray) {
-            return dArray.map(item =>
-                Highcharts.map(item, function (value) {
-                    var number = Math.round(value);
-                    return Highcharts.isNumber(number) ? number : value;
-                })
-            );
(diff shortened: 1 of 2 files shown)


ts/Series/ColumnPyramid/ColumnPyramidSeries.ts
```diff
@@ -228,9 +228,9 @@ class ColumnPyramidSeries extends ColumnSeries {
 
             // inverted chart
             if (chart.inverted) {
-                invBarPos = chart.plotWidth - barY;
+                invBarPos = yAxis.width - barY;
                 stackHeight = (topPointY -
-                (chart.plotWidth - (translatedThreshold as any)));
+                (yAxis.width - (translatedThreshold as any)));
```
raf18seb, line 233: @TorsteinHonsi shouldn't the linter catch the lack of indentation here?

It would be more readable written this way:
```js
stackHeight =
    (topPointY - (yAxis.width - (translatedThreshold as any)));
```
or even this way:
```js
stackHeight = (
    topPointY - (yAxis.width - (translatedThreshold as any))
);
```

samples/unit-tests/series-columnpyramid/basic/demo.js
```diff
@@ -54,3 +54,51 @@ QUnit.test('Column pyramid series - 0 dataLabel #12514', function (assert) {
...
+
+QUnit.test(
+    "Column pyramid series inverted - yAxis width #16150",
+    function (assert) {
+        const chart = Highcharts.chart("container", {
+            chart: {
+                type: "columnpyramid",
+                inverted: true,
+                width: 600,
+                animation: false,
```
raf18seb, line 66: Is it necessary? (`animation: false`)

samples/unit-tests/series-columnpyramid/basic/demo.js
```diff
@@ -54,3 +54,51 @@ QUnit.test('Column pyramid series - 0 dataLabel #12514', function (assert) {
...
+                height: 600
```
raf18seb, line 67: We usually set the chart's size in css.

samples/unit-tests/series-columnpyramid/basic/demo.js
```diff
@@ -54,3 +54,51 @@ QUnit.test('Column pyramid series - 0 dataLabel #12514', function (assert) {
...
+            },
+            yAxis: {
+                width: "40%"
+            },
+
+            series: [
+                {
+                    data: [5, 4, 2]
+                }
+            ]
```
raf18seb, line 77: I like to use this pattern:
```js
series: [{
    data: [5, 4, 2]
}]
```

samples/unit-tests/series-columnpyramid/basic/demo.js
```diff
@@ -54,3 +54,51 @@ QUnit.test('Column pyramid series - 0 dataLabel #12514', function (assert) {
...
+        });
+
+        var roundPath = function (dArray) {
+            return dArray.map(item =>
+                Highcharts.map(item, function (value) {
```
raf18seb, line 82: It's deprecated. Please, use the `Array.prototype.map` instead.

samples/unit-tests/series-columnpyramid/basic/demo.js
```diff
@@ -54,3 +54,51 @@ QUnit.test('Column pyramid series - 0 dataLabel #12514', function (assert) {
...
+            ["L", 335, 223],
+            ["L", 406, 223],
+            ["Z"]
+        ];
+
+        const testedPath = roundPath(chart.series[0].points[0].shapeArgs.d);
+        assert.deepEqual(
+            testedPath,
+            validPath,
+            "testedPath should be equal to validPath, #16150"
```
raf18seb, line 101: Could you elaborate on this sentence? Could be:
```
The pyramid shape should be correct, #16150."
```

raf18seb (changes requested):
Thanks!

Few clues:
- you're mixing ES5 with ES6 - please use one of them (ES6 preferable)
- please use the single quotes `'` in JavaScript
- test is failing in Safari

raf18seb (approved):
Good job, thanks!

TorsteinHonsi (approved):

samples/unit-tests/series-columnpyramid/basic/demo.js
```diff
@@ -54,3 +54,39 @@ QUnit.test('Column pyramid series - 0 dataLabel #12514', function (assert) {
         'Label should be translated and had different position translate attributes than starting values #12514'
     );
 });
+
+QUnit.test(
```
TorsteinHonsi, line 58: This test is good, but in order to keep the processing time down in our test suite, it can be combined with an existing test. The first test in this file ('Column pyramid series') can be extended by setting an `yAxis.width`.

TorsteinHonsi (changes requested):

TorsteinHonsi (approved):
Thanks! Pending approval by @sebastianbochan

PR #16268, opened 2021-08-30
Fixed CSP and style props issues with a11y module in styled mode.

oysteinmoseng:


TorsteinHonsi (changes requested):
Thanks! In the [css/a11y](http://utils.highcharts.local/samples/#view/highcharts/css/a11y) demo, there's still one `<li style="visibility: visible;">...</li>` that probably could be set in CSS.

oysteinmoseng:
`Utils.attr` in this branch or master does not support removing attributes by setting them to `null` or `undefined`. Was this added in a different PR that is under review @TorsteinHonsi? Or does it have to be added?

TorsteinHonsi (approved):

PR #16269, opened 2021-08-30
Improvements to series keyboard navigation.

oysteinmoseng:
Improvements to series keyboard navigation, now supporting `End`/`Home`/`PgUp`/`PgDown` keys.
___
We no longer go to last point when tabbing backwards into the chart, as this was shown in user testing to be confusing for some users.

Instead we now support these additional keyboard shortcuts:
- `Home`: Go to first point in chart
- `End`: Go to last point in chart
- `PageUp`: Go to next series
- `PageDown`: Go to previous series


TorsteinHonsi (approved):
Thanks!

PR #16281, opened 2021-09-01
Update render-charts-serverside.md

goransle:
Replaced a broken link on https://www.highcharts.com/docs/export-module/render-charts-serverside

@TorsteinHonsi ~~This article is pretty much an expanded version of https://www.highcharts.com/docs/export-module/setting-up-the-server. Perhaps we should retire one of them to avoid outdated information? I would suggest to keep the shorter one at /docs and have the export-server documentation on the GH repo.~~

Edit: the second article is actually mostly about the legacy export server, so perhaps we should just remove that from the sidebar?


TorsteinHonsi:
@goransle It's time to bury all the stuff dealing with the old export servers. Having an article on command-line rendering is still useful, so maybe we should copy the legacy part over to a separate page which is only accessible from a subtle link at the end of https://www.highcharts.com/docs/export-module/render-charts-serverside?

goransle:
We had a hidden doc already at [docs/export-module/legacy-export-servers](https://www.highcharts.com/docs/export-module/legacy-export-servers), so moved the instructions on the pure PhantomJS implementation there.

TorsteinHonsi (approved):
Thanks!

PR #16285, opened 2021-09-02
feature/15225-offset-direction

### LINKED ISSUE #15225: Organization chart: add new offsetDirection property to allow how offset property gets applied
# Description of the feature

Create a new organization chart property named "offsetDirection" that takes values of "vertical" or "horizontal".  This would be used in conjunction with the existing "offset" property as current offset is only applied in fixed direction depending if chart is inverted or not.   This would allow organization nodes to be shifted up or down, or left or right.

https://api.highcharts.com/highcharts/plotOptions.organization

https://jsfiddle.net/ttbsoftware/a5h1vwdp/
----
image

### LINKED ISSUE #15255: Polygon chart update with wrong data with button click in highcharts react
Goal : To change the polygon series used in the chart with a button. For example, I want to display a yellow square initially, then with a button, display a different red square. 

#### Expected behaviour

Yellow square is displayed correctly. When button is clicked, red square is displayed correctly 

#### Actual behaviour

Both yellow and red squares are not displayed correctly. 

https://stackblitz.com/edit/react-a4whqv?file=index.js

#### Product version
    "highcharts": "8.2.2",
    "highcharts-react-official": "3.0.0",

#### Affected browser(s)

Chrome

KacperMadej:
Added new offset options for nodes in Sankey and Organization chart, [`nodes.offsetHorizontal`](https://api.highcharts.com/highcharts/series.sankey.nodes.offsetHorizontal) and [`nodes.offsetVertical`](https://api.highcharts.com/highcharts/series.sankey.nodes.offsetVertical). See [#15225](https://github.com/highcharts/highcharts/issues/15225).

Closes #15225.

### DIFF
--- a/samples/highcharts/plotoptions/sankey-node-column/demo.js
+++ b/samples/highcharts/plotoptions/sankey-node-column/demo.js
@@ -18,7 +18,9 @@ Highcharts.chart('container', {
         ],
         nodes: [{
             id: 'US',
-            offset: 130
+            //offset: 130 // deprecated
+            offsetHorizontal: -50,
+            offsetVertical: 130
         }, {
             id: 'EU',
             column: 2
--- a/ts/Series/Sankey/SankeySeries.ts
+++ b/ts/Series/Sankey/SankeySeries.ts
@@ -468,11 +468,16 @@ class SankeySeries extends ColumnSeries {
                 nodePadding = series.nodePadding;
 
             for (let i = 0; i < column.length; i++) {
-                const sum = column[i].getSum();
-                const height = Math.max(
-                    sum * factor,
-                    series.options.minLinkWidth as any
-                );
+                const sum = column[i].getSum(),
+                    height = Math.max(
+                        sum * factor,
+                        series.options.minLinkWidth as any
+                    ),
+                    directionOffset = node.options[chart.inverted ?
+                        'offsetHorizontal' :
+                        'offsetVertical'
+                    ],
+                    optionOffset = node.options.offset || 0;
 
                 if (sum) {
                     totalNodeOffset = height + nodePadding;
@@ -482,9 +487,15 @@ class SankeySeries extends ColumnSeries {
                 }
                 if (column[i] === node) {
                     return {
-                        relativeTop: offset + relativeLength(
-                            node.options.offset || 0,
-                            totalNodeOffset
+                        relativeTop: offset + (defined(directionOffset) ?
+                            // directionOffset is a percent of the node height
+                            relativeLength(
+                                directionOffset,
+                                height
+                            ) : relativeLength(
+                                optionOffset,
+                                totalNodeOffset
+                            )
                         )
                     };
                 }
@@ -914,14 +925,13 @@ class SankeySeries extends ColumnSeries {
             toY = getY(toNode, 'linksTo'),
             nodeLeft = fromNode.nodeX,
             nodeW = this.nodeWidth,
-            right = (toNode.column as any) * this.colDistance,
+            right = toNode.nodeX,
             outgoing = point.outgoing,
             straight = right > nodeLeft + nodeW;
 
         if (chart.inverted) {
             fromY = (chart.plotSizeY as any) - fromY;
             toY = (chart.plotSizeY || 0) - toY;
-            right = (chart.plotSizeX as any) - right;
             nodeW = -nodeW;
             linkHeight = -linkHeight;
             straight = nodeLeft > right;
@@ -1060,6 +1070,7 @@ class SankeySeries extends ColumnSeries {
                 Math.round(sum * translationFactor),
                 this.options.minLinkWidth as any
             ),
+            nodeWidth = Math.round(this.nodeWidth),
             crisp = Math.round(options.borderWidth as any) % 2 / 2,
             nodeOffset = column.offset(node, translationFactor),
             fromNodeTop = Math.floor(pick(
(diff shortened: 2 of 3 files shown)


ts/Series/Sankey/SankeySeries.ts
```diff
@@ -1258,16 +1269,50 @@ export default SankeySeries;
  * @type {string}
  * @product highcharts
  *//**
+ * This option is deprecated, use
+ * {@link Highcharts.SankeyNodeObject#offsetHorizontal} and
+ * {@link Highcharts.SankeyNodeObject#offsetVertical}  instead.
```
raf18seb, line 1274: Double space before "instead"

ts/Series/Sankey/SankeySeries.ts
```diff
@@ -1258,16 +1269,50 @@ export default SankeySeries;
...
  *
+ * @deprecated
  * @name Highcharts.SankeyNodeObject#offset
- * @type {number}
+ * @type {number|string}
  * @default 0
  * @since 6.0.5
+ *//**
+ * The horizontal offset of a node. Positive values shift the node right,
+ * negative shift it left.
```
raf18seb, line 1292: "Positive velues shift" or "Positive value shifts"?

ts/Series/Sankey/SankeySeries.ts
```diff
@@ -1258,16 +1269,50 @@ export default SankeySeries;
...
+ * The vertical offset of a node. Positive values shift the node down,
+ * negative shift it up.
+ *
+ * If a percantage string is given, the node is offset by the percentage of the
+ * node size.
+ *
+ * @see {@link https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/plotoptions/sankey-node-column/|Highcharts-Demo:}
+ *         Specified node offset
+ *
+ * @name Highcharts.SankeyNodeObject#offsetVertical
```
raf18seb, line 1313: What about naming it `horizontalOffset` and `verticalOffset`?

Idea: instead of deprecating `offset`, can we use the old `offset` as `horizontalOffset`, and add only `verticalOffset` as a new option?

ts/Series/Sankey/SankeySeries.ts
```diff
@@ -1258,16 +1269,50 @@ export default SankeySeries;
...
+ * @type {number|string}
+ * @since next
+ *//**
+ * The vertical offset of a node. Positive values shift the node down,
+ * negative shift it up.
+ *
+ * If a percantage string is given, the node is offset by the percentage of the
+ * node size.
+ *
+ * @see {@link https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/plotoptions/sankey-node-column/|Highcharts-Demo:}
```
raf18seb, line 1310: Shouldn't we update the demo?

ts/Series/Sankey/SankeySeries.ts
KacperMadej: A copy-paste based on the previous `nodes.offset` description. There is no `Positive value shifts` that I can find in the file (and it probably should be `A positive value shifts`). However, I'm for the currently used wording. @TorsteinHonsi ?

raf18seb (commented):
Few small questions/suggestions

ts/Series/Sankey/SankeySeries.ts
KacperMadej: It is updated with the new options being used. I'll tag you there.

samples/highcharts/plotoptions/sankey-node-column/demo.js
```diff
@@ -18,7 +18,9 @@ Highcharts.chart('container', {
         ],
         nodes: [{
             id: 'US',
-            offset: 130
+            //offset: 130 // deprecated
+            offsetHorizontal: -50,
+            offsetVertical: 130
```
KacperMadej, line 23: @raf18seb The demo update.

ts/Series/Sankey/SankeySeries.ts
KacperMadej: The problem with keeping the old one as one of the new ones is if that the old `offset` is either horizontal or vertical depending on the chart being inverted or not inverted, so it won't be very intuitive IMO.

I like the current names as they get to be placed near the `offset` when sorted alphabetically 😄 
Why different names?

ts/Series/Sankey/SankeySeries.ts
TorsteinHonsi: I think it's ok

TorsteinHonsi (commented):
Looks good, thanks! 

Last minute question - have we considered making these options behave the same regardless of `chart.inverted`? If we wanted that, we could call them `offsetLateral` and  `offsetLongitudinal` and remove the logic that deals with inversion.

KacperMadej:
@TorsteinHonsi 
The description of the submitted feature request was about vertical and horizontal offset, so this started the variable names and the logic matches the names. I see no problem with changing the names, logic, and docs if this will work more consistently with the other chart options. However, we already have [`xOffset`](https://api.highcharts.com/highcharts/series.windbarb.xOffset) and [`yOffset`](https://api.highcharts.com/highcharts/series.windbarb.yOffset) that doesn't work the same regardless of `chart.inverted` - https://jsfiddle.net/BlackLabel/gfdjzr28/. This also provides another pair of possible names for the options - if it will be decided to keep the current logic.

IMO: let's keep the current logic and change names to `xOffset` and `yOffset`.

TorsteinHonsi:
@KacperMadej I don't think `xOffset` and `yOffset` makes so much sense in a sankey chart, because we don't have X and Y axes like we do in the wind barb. Let's do a quick discussion on Slack.

TorsteinHonsi (approved):

Issue #15409, opened 2021-03-26
IndicatorRequired mixin isn't working in v9


raf18seb:
#### Expected behaviour
When running an indicator that requires another indicator module to be loaded (for example the dema indicator except `dema.js` also requires `ema.js` to be loaded) without the required module, the `console.log()` with an error description should be thrown into the console.

#### Actual behaviour
No description in the console. Instead, other errors are displayed.

#### Live demo with steps to reproduce
https://jsfiddle.net/BlackLabel/9msbrceq/

#### Product version
Since v9.0.0

pawelfus:
It's not a bug/regression in the `indicatorRequired` mixin. It works the same way it used to, but it's a problem related to tree shaking and modularizing series dependencies (under the hood, every technical indicator is just a series). Once similar issues will be resolved (e.g. sankey and dependency wheel series) we will be able to pick this up.

Status: On hold

bre1470:
It is impossible to combine this with tree shaking. I tried several approaches, but it depends on the deprecated seriesType factory when it comes to options and constructor.

If we move the logic out of init right before the class, we could at least trigger the error message right before it fails. Opinions?

bre1470:
@raf18seb @pawelfus Could you take a look?

pawelfus:
> If we move the logic out of init right before the class, we could at least trigger the error message right before it fails. Opinions?

The idea to implement this in `init` was to throw an error and prevent running all the calculations (which could throw a lot of other errors, which would be really messy and easy to miss the most important error). I think we can safely move this out of the `init` function, as you suggest 👍 If we could also somehow prevent defining the indicator class, that would be great. So in the end, if we would end up with only two errors (missing dependency and error-17 (missing series)).

Perhaps we should resign from `indicatorRequired` mixin? Instead, let's create a mechanism (or refactor the existing one) that does the check for any module? Demo for the Sankey and Dependency-wheel: https://jsfiddle.net/BlackLabel/fmdLj4cs/1/ (missing Sankey dependency)

raf18seb:
There's still `Uncaught TypeError: Cannot read property 'prototype' of undefined` error before error-17. If this is not a concern for us, maybe we could reuse the logic from error-17?

bre1470:
The prototype error comes from the transpiled code, that TypeScript produces for the class inheritance.

bre1470:
[...]

PR #16286, opened 2021-09-02
Migration: Indicator requirements

bre1470:
Removed `requiredIndicators` info and added EMA to the default `indicators` master. See #15409


bre1470:
Might address https://github.com/highcharts/highcharts/issues/15409

changelog/highcharts/9.0.0.md
```diff
@@ -7,6 +7,7 @@
 
 ## Upgrade notes
 - Due to improved security, some tags or attributes may be removed from HTML configuration options like labels or tooltip formats and formatters. Read more [in the docs](https://www.highcharts.com/docs/chart-concepts/labels-and-string-formatting#filtering) on how to allow them.
+- IndicatorRequired mixin no longer works, because the class architecture (related to tree-shaking) has changed.
```
pawelfus, line 10: 1. I'm afraid this might not be really clear to users. Can we simply inform that error of missing a technical indicator dependency requirements will no longer be there? 

2. Wrong file. Not Highcharts changelog, but Stock should be updated: https://github.com/highcharts/highcharts/blob/master/changelog/highcharts-stock/9.0.0.md

ts/masters/indicators/indicators.src.ts
```diff
@@ -11,4 +11,5 @@
  * License: www.highcharts.com/license
  */
 'use strict';
+import '../../Stock/Indicators/EMA/EMAIndicator.js';
```
pawelfus, line 14: Q: Does the order not matter here?

ts/Stock/Indicators/LinearRegressionAngle/LinearRegressionAngle.ts
```diff
@@ -154,6 +154,7 @@ export default LinearRegressionAngleIndicator;
  * @product   highstock
  * @excluding dataParser,dataURL
  * @requires  stock/indicators/indicators
+ * @requires  stock/indicators/linearregression
```
pawelfus, line 157: Just checked the dist folder: there is no `linearregression`, `linearregressionangle` etc modules. Only one, that combines them all: https://code.highcharts.com/stock/indicators/regressions.src.js

pawelfus (changes requested):
A lot of work, amazing! Thank you!

There are still a few indicators that unnecessary define `requiredindicators`:
- KeltnerChannel
- MACD
- NATR
- Supertrend
- SMA (I know it's empty, but without the mixin, it's a redundant property on the class. I don't know why it wasn't defined in the mixin, it shouldn't be there I guess)

Docs:
- APO, Chaikin, DEMA, MACD, PPO, TEMA, and TRIC still mention `ema` is required. Some of them still have `@requires stock/indicators/ema`, some others mention this in the doclet too
- Disparity Index uses EMA as an example of loading different average algorithms, could you replace it with DEMA instead?:

             * By default it uses SMA. To use other averages, e.g. EMA,
             * the `stock/indicators/ema.js` file needs to be loaded.
->

             * By default it uses SMA. To use other averages, e.g. TEMA,
             * the `stock/indicators/tema.js` file needs to be loaded.

ts/masters/indicators/indicators.src.ts
bre1470: Outside of karma it probably does. Fixed.

pawelfus (changes requested):
Nice catch with demos, thanks!

This still hasn't changed:
> There are still a few indicators that unnecessary define `requiredindicators`:
> 
> - KeltnerChannel
> - MACD
> - NATR
> - Supertrend
> - SMA (I know it's empty, but without the mixin, it's a redundant property on the class. I don't know why it wasn't defined in the mixin, it should be there I guess)

bre1470:
Then I probably do not understand, what you mean with that. I was searching for it and nothing pops up.

bre1470:
Got it now! It was a matter of character case. `requiredIndicators`

pawelfus:
> Got it now! It was a matter of character case. `requiredIndicators`

Aaah, sorry!

pawelfus (approved):
Great job, thank you!

bre1470:
Thanks and thank you for your help! :)

TorsteinHonsi (approved):

PR #16294, opened 2021-09-03
bugfix/15656-wheel

### LINKED ISSUE #15656: Highcharts highmap mouse wheel zoom doesn't work on windows machine
#### Expected behaviour
In Edge or Chrome mouse wheel should trigger zoom.
On MacOs, user has to hold shift with mouse wheel to zoom the map.

#### Actual behaviour
Mouse wheel doesn't zoom in map.

https://jsfiddle.net/wL6j9ten/

#### Product version
 Highmaps JS v9.1.0

#### Affected browser(s)
Windows:
- Edge
- Chrome

KacperMadej:
Fixed #15656, resolved broken mouse scroll on some browsers.

--------
Looks like `e.deltaY` is problematic as it is not in the same units - in different browser+OS combinations are giving drastically different values. If `e.wheelDelta` will be prioritized then other possible options (`e.deltaY` and `e.detail`) will be used mostly be some old or probably some future browsers (`e.wheelDelta` is [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/WheelEvent)).

Tested manually on Win and Mac in Chrome, Edge, Firefox and Safari (Mac). The broken browsers are now fixed and some other browsers are slightly changed (smaller jumps on scroll and IMHO better zoom, so an overall improvement).

### DIFF
--- a/ts/Maps/MapPointer.ts
+++ b/ts/Maps/MapPointer.ts
@@ -14,6 +14,7 @@ import type PointerEvent from '../Core/PointerEvent';
 import Pointer from '../Core/Pointer.js';
 import U from '../Core/Utilities.js';
 const {
+    defined,
     extend,
     pick,
     wrap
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (approved):
Thanks for the fix!

PR #16304, opened 2021-09-08
Migration: PointSonify

marvin19:
Added composition and ES6 pattern to PointSonify.


bre1470 (approved):
Thank you! Looks good. 👍

PR #16306, opened 2021-09-08
Tools: Migrated Palette to const enum

bre1470:
Tools: Migrated Palette to const enum for in-place injection to reduce file size.


TorsteinHonsi (approved):
Thanks!

PR #16311, opened 2021-09-09
Migration: compare/cumulative composition

raf18seb:
Stock: refactoring compare and cumulative into a composition.


ts/Series/CompareCumulativeComposition.ts
```diff
@@ -0,0 +1,741 @@
...
+    interface PointLike {
+        change?: number;
+        cumulativeSum?: number;
+    }
+}
+
+declare module '../Core/Series/SeriesLike' {
+    interface SeriesLike {
+        compareValue?: number;
+        getCumulativeExtremes(activeYData: Array<number>): [number, number];
```
pawelfus, line 68: Instead of extending `SeriesLike` with every method/property, perhaps it's better to store it all under one property?  What do you think @bre1470 ?

Like this:
```
interface ModifiersAdditions {
  compareValue?: number;
  getCumulativeExtremes(activeYData: Array<number>): [number, number];
  ...
}

declare module '../Core/Series/SeriesLike' {
    interface SeriesLike {
        modifiers: ModifiersAdditions
    }
}
```

raf18seb:
This PR is a merge to https://github.com/highcharts/highcharts/pull/16020
Please hold with a code review, I'll add `Series.dataModify = {}` class to keep all the methods and props on it.

ts/Series/CompareCumulativeComposition.ts
pawelfus: Just realized the [comment above](https://github.com/highcharts/highcharts/pull/16311#issuecomment-915977970) - okay!

ts/Series/CompareCumulativeComposition.ts
```diff
@@ -0,0 +1,741 @@
...
+        }
+
+        if (composedClasses.indexOf(AxisClass) === -1) {
+            composedClasses.push(AxisClass);
+
+            const axisProto = AxisClass.prototype as AxisComposition;
+
+            axisProto.setCompare = axisSetCompare;
+            axisProto.setModifier = setModifier;
+            axisProto.setCumulative = AxisSetCumulative;
```
karolkolodziej, line 174: Why does this one start with a capital letter?

karolkolodziej (changes requested):

ts/Series/CompareCumulativeComposition.ts
bre1470: I agree.

raf18seb:
Reach for review. I'm open to any naming suggestions, especially for `dataModify`, `CompareCumulativeComposition` and class `Additions`

ts/Series/CompareCumulativeComposition.ts
```diff
@@ -0,0 +1,789 @@
...
+    }
+
+    /**
+     * Extend series.processData by finding the first y value in the plot area,
+     * used for comparing the following values
+     *
+     * @ignore
+     * @function Highcharts.Series#processData
+     */
+    function afterProcessData(this: Series): (boolean|undefined) {
```
karolkolodziej, line 369: I'm not sure if this function returns any of that.

karolkolodziej (changes requested):
Please check how the documentation is generated because there is no information about the   `setCompare` method in API, while it is being used in the demo and it works fine.

ts/Series/CompareCumulativeComposition.ts
```diff
@@ -0,0 +1,789 @@
...
+        dataModify?: CompareCumulativeComposition.Additions;
+        setCompare(compare?: string|null, redraw?: boolean): void;
+        setCumulative(cumulative?: boolean|null, redraw?: boolean): void;
+    }
+}
+
+declare module '../Core/Series/SeriesOptions' {
+    interface SeriesOptions {
+        compare?: string|null;
+        compareBase?: (0|100);
```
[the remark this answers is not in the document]
bre1470: I disagree. It makes it clear that this is a union type. All union and intersect types should have parentheses.

pawelfus (approved):
Thanks!
I guess `Additions` may stay (many other compositions use it, without any conflicts).
For `dataModify` and `CompareCumulativeComposition` I don't have any great alternatives.. perhaps `dataModify` -> `dataMode`/`dataRescalers`/`dataAnalyze` or `performanceMode` (`compare="percent"` is often called a `performance chart`, but it's not the case for `cumulative sum`)? And `CompareCumulativeComposition` -> `DataModeComposition`? Performance is a bit misleading, though.

ts/Series/CompareCumulativeComposition.ts
```diff
@@ -0,0 +1,789 @@
...
+        change?: number;
+        cumulativeSum?: number;
+    }
+}
+
+declare module '../Core/Series/SeriesLike' {
+    interface SeriesLike {
+        dataModify?: CompareCumulativeComposition.Additions;
+        setCompare(compare?: string|null, redraw?: boolean): void;
+        setCumulative(cumulative?: boolean|null, redraw?: boolean): void;
```
bre1470, line 69: Place all composition states and functions inside the `Additions` class, as this already receives the series instance via constructor. It avoids name conflicts with other compositions and future changes. Also the property for the composition additions should reflect the name of the composition: `Series.compareCumulative` or `DataModifyComposition.Additions` - you choice. The same with axis and point.

```ts
interface SeriesLike {
  compareCumulative?: CompareCumulativeComposition.SeriesAdditions
}
interface AxisLike {
  compareCumulative?: CompareCumulativeComposition.AxisAdditions;
}
interface Point {
  compareCumulative?: CompareCumulativeComposition.PointAdditions; // probably an interface for the two values
}

File: ts/Series/CompareCumulativeComposition.ts [hunk 6]
```diff
@@ -0,0 +1,789 @@
+/* *
+ *
+ *  !!!!!!! SOURCE GETS TRANSPILED BY TYPESCRIPT. EDIT TS FILE ONLY. !!!!!!!
+ *
+ * */
```
bre1470, line 5: What and how should the copyright and license information be placed her, @TorsteinHonsi?

bre1470 (changes requested):

ts/Series/CompareCumulativeComposition.ts
raf18seb: Copied from the old `compare`. Corrected

samples/unit-tests/series/cumulative/demo.js
```diff
@@ -16,7 +16,8 @@ QUnit.test('Stock: general tests for the Cumulative Sum', function (assert) {
         dataArr2 = [4, 5, 1, -3];
 
     assert.deepEqual(
-        Highcharts.Series.prototype.getCumulativeExtremes([1, 2, 3, 4, 5, 0]),
+        chart.series[0].dataModify.constructor
```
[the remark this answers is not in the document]
raf18seb: Thanks ;) without `.prototype` to be more precise

ts/Series/CompareCumulativeComposition.ts
raf18seb: All the methods outside the `Additions` class are API methods and should be available for users methinks.

Changed to `DataModifyComposition.Additions` :)

bre1470 (approved):
Looks good. 👍

karolkolodziej (approved):

pawellysy (approved):
Looks good!

TorsteinHonsi (approved):

ts/Series/DataModifyComposition.ts
```diff
@@ -0,0 +1,788 @@
...
+        this.dataModify = new Additions(this as SeriesComposition);
+
+        if (this.options.compare) {
+            // Set comparison mode
+            this.dataModify.initCompare(this.options.compare);
+        } else if (this.options.cumulative) {
+            // Set Cumulative Sum mode
+            this.dataModify.initCumulative(this.options.cumulative);
+        }
+    }
```
TorsteinHonsi, line 288: I am wondering if we should set `this.dataModify` only when we have an active modifier, and otherwise delete it. Then in Series.ts we can test for `series.dataModify`, which would always have a `modifyValue`. Instead of a deep test for `series.dataModify && series.dataModify.modifyValue` which we are currently doing repeatedly. By implication, we would not set `modifyValue` to null, but instead do something like this:

```ts

        let dataModify: Additions|undefined;
        if (this.options.compare || this.options.cumulative) {
            dataModify = new Additions(this as SeriesComposition);
            if (this.options.compare) {
                // Set comparison mode
                dataModify.initCompare(this.options.compare);
            } else {
                // Set Cumulative Sum mode
                dataModify.initCumulative(this.options.cumulative);
            }
        }
        this.dataModify = dataModify;
```
This pattern is more in line with other compositions we have. Like for example `Axis.logarithmic`- when `logarithmic` is set, we can safely use `logarithmic.getLogTickPositions`.

TorsteinHonsi (changes requested):

ts/Series/DataModifyComposition.ts
```diff
@@ -0,0 +1,805 @@
...
+    ): void {
+        if (!this.isXAxis) {
+            this.series.forEach(function (series): void {
+                if (
+                    mode === 'compare' &&
+                    (
+                        modeState === 'percent' ||
+                        modeState === 'value' ||
+                        typeof modeState === 'undefined' ||
+                        modeState === null
```
pawelfus, line 229: I might be missing something, but do we need the second part of this `if`, which checks all possible types of `modeState`?  The same below, where we would check types for `cumulative` mdoe.

pawelfus (commented):
See my comments below

karolkolodziej (approved):

TorsteinHonsi (approved):

PR #16313, opened 2021-09-09
Enhancement/more series a11y options

oysteinmoseng:
Support `series.accessibility.point` options matching the top level `accessibility.point` options.
___
This makes it possible to define things like `accessibility.point.dateFormat` and `accessibility.point.valueDescriptionFormat` on a per-series or per-series-type basis.

Note: This deprecates `series.accessibility.pointDescriptionFormatter` in favor of `series.accessibility.point.descriptionFormatter` for consistency. The deprecated option is automatically moved over, and an error 32 warning is given, like the other deprecated options.

Existing demos using `pointDescriptionFormatter` have been changed where possible to use the new options.


TorsteinHonsi (approved):
Thanks!

PR #16321, opened 2021-09-10
bugfix/13683-legend-nav-pages

### LINKED ISSUE #13683: Legend is missing the last page
#### Expected behaviour
Legend has the 7th page containing series 97 to 100. (in the attached JSFiddle case) 

#### Actual behaviour
Legend does not have the 7th page.

https://jsfiddle.net/ktakahashimtb/shLpjcqb/

** Might need to open it on Windows to see the issue **

This issue seems to happen only when a legend has N x N arrangement of the items. 
In this example, the container element is sized so that the legend has 4 x 4 items on a page.

In addition, setting maxHeight seems to be a necessary ingredient.

#### Product version
Highcharts v8.1.1 (to at least v 7.2.0)

#### Affected browser(s)
Chrome, Edge, Firefox (on Windows)
(could not reproduce this on MacBook - Chrome and Safari)

KacperMadej:
Fixed #13683, the last row of items in a paginated legend was not fully visible in edge cases.

### DIFF
--- a/samples/unit-tests/legend/navigation/demo.details
+++ b/samples/unit-tests/legend/navigation/demo.details
@@ -1,0 +1,6 @@
+---
+ resources:
+   - https://code.jquery.com/qunit/qunit-2.0.1.js
+   - https://code.jquery.com/qunit/qunit-2.0.1.css
+ js_wrap: b
+... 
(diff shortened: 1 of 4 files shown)


TorsteinHonsi (approved):
Thanks!

PR #16326, opened 2021-09-12
Stock Tools: corrected "langKey" for crooked5 annotation

### LINKED ISSUE #16318: Stock-Tools: Annotation "crooked5" has generic langKey "crookedLine"
#### Expected behaviour
I'd expect stock-tools.js to set up the crooked5 Annotation with `langKey: "crooked5"`, like e.g. crooked3 does.
This way it would be more specific for translation of the Annotation names.

Not sure if it is a bug (like copy&paste issue) or intended as is. If intended, please feel free to decline my request. It is for sure no big deal to use it as is.

#### Actual behaviour
stock-tools.js defines the generic `langKey: "crookedLine"` for "crooked5"

n/a

#### Product version
Highstocks 9.2.2

#### Affected browser(s)
n/a

raf18seb:
Fixed #16318, corrected the `langKey` property in Stock Tools gui for the Crooked Line 5 annotation.

### DIFF
--- a/ts/Stock/StockToolsBindings.ts
+++ b/ts/Stock/StockToolsBindings.ts
@@ -1312,7 +1312,7 @@ const stockToolsBindings: Record<string, Highcharts.NavigationBindingsOptionsObj
      *
      * @type    {Highcharts.NavigationBindingsOptionsObject}
      * @product highstock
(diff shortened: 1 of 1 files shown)


pawelfus (approved):
Thanks!

Issue #16052, opened 2021-07-15
Added series points have markers unexpectedly appear


gwatts-klipfolio:
#### Expected behaviour

- If the marker.enabled is set to false on data points within a new series the markers should not appear. 

#### Actual behaviour

- When a new series is added followed by a Chart.redraw specifically disabled markers appear.
- Note, this also occurs when clicking on legend items to have them disappear then re-appear (This behaviour can be seen in the jsfiddle provided)

#### Live demo with steps to reproduce
https://jsfiddle.net/e96xy0z1/

- Click on a legend item to have an area disappear
- Click on the hidden legend item to have it reappear
- Notice the marker in the middle of the area, also notice in the spec that the marker is specifically disabled.
- Occurs with area and line charts

#### Product version
Highcharts: 9.0.1

khlieng:
Related to #15480 @oysteinmoseng

gwatts-klipfolio:
The workaround described in #15480 (Making the markers the same size as the line) works to hide the markers, however, we also need to disable the hover state (tooltip) for the markers in question. These will still appear.

PR #16337, opened 2021-09-14
Fixed #15480 and #16052, point markers appear on redraw.

### LINKED ISSUE #15480: a11y module shows disabled point markers
#### Behaviour
accessibility module is showing disabled markers after chart resize or update.
Without the module markers are working correctly - still hidden after resize or update if were disabled at the start.

https://jsfiddle.net/BlackLabel/msdr1ohx/3/

#### Product version
9.0.1

oysteinmoseng:
Fixed #15480 and #16052, disabled point markers sometimes appeared on redraw with accessibility module loaded.


TorsteinHonsi (approved):
Thanks!

Issue #4550, opened 2015-09-09
Offline exporting: Firefox is unable to create gradient color for PNG


KacperMadej:
Using Firefox (40.0.3 - latest) and offline-exporting module for exporting chart in offline mode to PNG file gradient color is not created properly - is not displaying.

Example: http://jsfiddle.net/7ponnv7r/2/
Steps: 
1. Export to PNG via chart's context menu.
2. Open downloaded PNG file of the chart and see that linear nor radial gradients are not visible.

highslide-software:
@oysteinmoseng It may be related to [this code block](https://github.com/highslide-software/highcharts.com/blob/v4.1.8/js/parts/SvgRenderer.js#L1217-L1222). Firefox has turned out to be particularly sensitive to how internal URLs are used in SVG.

oysteinmoseng:
This seems to be an issue with Firefox, see minimal example: http://jsfiddle.net/wa1xmvt9

Canvas.drawImage does not appear to work with internal references.

oysteinmoseng:
This is apparently a known (and old) [issue](https://bugzilla.mozilla.org/show_bug.cgi?id=841920) with Firefox and Blobs.

PR #16338, opened 2021-09-14
Fixed #12169, allow Blob offline export in Firefox.

### LINKED ISSUE #12169: Offline exporting: workaround for previous browser bugs no longer needed
`Highcharts.svgToDataUrl (modules/offline-exporting.src.ts)`

#### Expected behaviour
A CSP blocking "data:" scheme for img-src would still allow image export via "blob:"

#### Actual behaviour
Export fails in Firefox and (presumably) Safari, due to a workaround put in place for previous browser bugs.

```
// Safari requires data URI since it doesn't allow navigation to blob
// URLs. Firefox has an issue with Blobs and internal references,
// leading to gradients not working using Blobs (#4550)
```

Previous issue #4550 highlighted an issue with Firefox not handling internal references correctly, workaround was to fallback to "data:".

This has subsequently been fixed in Firefox. Related bugs:
https://bugzilla.mozilla.org/show_bug.cgi?id=841920
https://bugzilla.mozilla.org/show_bug.cgi?id=1280584
https://bugzilla.mozilla.org/show_bug.cgi?id=1447262

Also, there's a workaround in there for Safari with comment saying Safari does not allow navigation to "blob:", which I believe is also fixed (based on discussion in this issue, but I don't have a mac to check): https://bugs.webkit.org/show_bug.cgi?id=102914 

#### Affected browser(s)
Firefox & Safari (?)

oysteinmoseng:
Fixed #12169, allow Blob offline export in Firefox.
___
Removes old workaround for #4550. Tested and verified that it is no longer needed. The Safari workaround appears to still be necessary.

### DIFF
--- a/ts/Extensions/OfflineExporting/OfflineExporting.ts
+++ b/ts/Extensions/OfflineExporting/OfflineExporting.ts
@@ -833,10 +833,9 @@ namespace OfflineExporting {
 
         try {
             // Safari requires data URI since it doesn't allow navigation to
-            // blob URLs. Firefox has an issue with Blobs and internal
-            // references, leading to gradients not working using Blobs (#4550).
-            // foreignObjects also dont work well in Blobs in Chrome (#14780).
-            if (!webKit && !H.isFirefox && svg.indexOf('<foreignObject') === -1) {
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (approved):
Thanks!

PR #16341, opened 2021-09-14
Refactor series clipping

TorsteinHonsi:
* Merge Stock specific functionality into Highcharts (panes and axis line avoidance).
* Simpler logic


ts/Core/Series/Series.ts
```diff
@@ -2572,30 +2476,69 @@ class Series {
...
-                }, animation);
-            }
+        // Run the animation
+        } else if (
+            animationClipRect &&
+            // Only first series in this pane
+            !animationClipRect.hasClass('highcharts-animating')
+        ) {
+            const finalBox = series.getClipBox();
+            animation.step = (val, fx): void => {
```
pawelfus, line 2529: Q: Is it necessary? In theory, `step` is a valid option in the animation object.

pawelfus (changes requested):
Looks great!

There are still a few commented-out lines - do we need them?

I don't know the reason, but in Stock, every second series-adding creates 4x clip, instead of 2x: http://jsfiddle.net/BlackLabel/3ujox9ye/ (which animates the navigator series). In fact, with disabled markers, only 1x clip would be enough (`markerAnimationClipRect` is created always, not only when markers exist)

Tested also multiple panes (horizontal and vertical) in Stock - no issues.

This PR also fixes #4406

ts/Core/Series/Series.ts
TorsteinHonsi: Keeping the marker group update in the step of the general group animation will ensure they are in sync, and will probably also be faster because we don't set up extra timers for the marker group. But I'll make sure the custom `animation.step` option is not overwritten.

pawelfus (approved):
Nice, thanks! 🎉

Issue #14928, opened 2021-01-13
Highcharts Zooming Enhancement


markzolotoy:
I am showing a short video for new Zoom modes enhancement, specifically for Horizontal  only and Vertical only modes.
Please let me know if a similar feature has been requested and a time frame consideration.

Thanks

https://user-images.githubusercontent.com/7363944/104508958-a1135000-55ae-11eb-9c48-803181e283db.mp4

PaulDalek:
Hi @markzolotoy

Thank you for contacting us and suggesting this enhancement.

markzolotoy:
Any timeframe for the feature to be implemented?

sebastianbochan:
Hi @markhobson,
At this moment we do not have any ETA for the ticket.

KacperMadej:
Here is a PoC demo: https://jsfiddle.net/BlackLabel/aetbf9vz/

@markhobson Please check the PoC demo - is this matching the feature request? or something should be improved/changed? 

As for possible options, using one could be enough:
- size limit (hard-coded to 10 pixels in the PoC demo) for the 1-dimensional zoom
By default, it could be set to 0, so the feature will be disabled - providing required backwards compatibility.

If there are any options that will be useful to be added to the API please leave a comment about that.

markhobson:
@sebastianbochan @KacperMadej I think you've got the wrong Mark :)

KacperMadej:
Ugh. Right. Very sorry about this - thank you for letting us know 😅 

@markzolotoy

markzolotoy:
@KacperMadej. That looks very good. The only thing is that it's not very intuitive where to position the mouse to initiate one-dimensional zoom.

KacperMadej:
Right, I think we could add this too.

A partial border (frame) that will match the size limit in length on corners (2D zoom) or sides (1D zoom).

![frame](https://user-images.githubusercontent.com/8960408/104941657-a4924700-59b3-11eb-8fa5-1741196619c6.jpg)


For the API options: currently, the only styling is done through `chart.selectionMarkerFill`, but it might be worth considering some options structure like:
```
chart: {
  selectionMarker: {
    fill: color,
    ...new options to set color, width and/or style of the partial border...
```

Potential options:
- frame stroke width,
- frame color (stroke),
- frame dash style,
- frame padding (for placement inside or outside of the selection marker, but still with the proper paths to correctly indicate required mouse position to toggle the number of zoom dimensions dynamically)

[...]

PR #16342, opened 2021-09-14
Feature/14928-zoom-options-refactor

hihajdus:
Refactored zoom options, introduce the new object structure. Related to #14928.

### DIFF
--- a/ts/Core/Chart/ChartDefaults.ts
+++ b/ts/Core/Chart/ChartDefaults.ts
@@ -966,6 +966,55 @@ const ChartDefaults: ChartOptions = {
     zoomBySingleTouch: false,
 
     /**
+     * Chart zooming options.
+     */
+    zooming: {
+
+        /**
+         * Enables zooming by a single touch, in combination with
+         * [chart.zoomType](#chart.zoomType). When enabled, two-finger pinch
+         * will still work as set up by [chart.pinchType](#chart.pinchType).
+         * However, `zoomBySingleTouch` will interfere with touch-dragging the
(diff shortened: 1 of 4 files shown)


ts/Core/Chart/ChartOptions.d.ts
```diff
@@ -170,6 +171,10 @@ export interface ChartSelectionContextObject {
     yAxis: Array<ChartSelectionAxisContextObject>;
 }
 
+export interface ChartZoomingOptions {
```
KacperMadej, line 174: and pinchType?

KacperMadej (changes requested):
This PR looks more like a draft for now. It would be helpful to add in the top comment a checkbox list with planned features for the refactor.

IMHO: If this PR should keep the changelog label then the opening line should inform about the added feature and refactor notes should go into upgrade notes. Just adding a note to keep an eye on it, but this will be probably resolved later by @TorsteinHonsi

ts/Core/Chart/ChartDefaults.ts
```diff
@@ -965,6 +965,55 @@ const ChartDefaults: ChartOptions = {
      */
     zoomBySingleTouch: false,
 
+    /**
+     * Chart zooming options.
```
pawelfus, line 969: This needs `@since` and all options below should not define `@since`. It's not possible to declare:
```js
zooming: {
  singleTouch: true
}
```
In version 9.0.0, right? :)

pawelfus (changes requested):

PR #16346, opened 2021-09-15
bugfix/11756-svgrenderer-img-parameters

### LINKED ISSUE #11756: SVG renderer - width/height image parameters issue
In the [documentation](https://api.highcharts.com/class-reference/Highcharts.SVGRenderer#image), we can find information saying that:
> The image width. If omitted, it defaults to the image file width.
> The image height. If omitted it defaults to the image file height.

Meanwhile omitting these parameters is impossible. It is a documentation mistake or this functionality fails? 

[Demo](http://jsfiddle.net/BlackLabel/w7h9t52k/) where I tried to set on of those parameters to: undefined, '' - empty string, or even null.  

#### Product version

Highcharts JS v7.1.3 (2019-08-14)

#### Affected browser(s)

All

mfilipiec:
Fixed #11756, adding image without optional parameters threw an error.

### DIFF
--- a/samples/unit-tests/svgrenderer/symbol/demo.js
+++ b/samples/unit-tests/svgrenderer/symbol/demo.js
@@ -196,3 +196,23 @@ QUnit.test('Square/rect', assert => {
         );
     });
 });
+
+QUnit.test('Image', assert => {
+    const renderer = new Highcharts.Renderer(
+        document.getElementById('container'),
+        400,
+        300
+    );
+
+    renderer.image(
+        'https://www.highcharts.com/samples/graphics/sun.png',
+        100,
+        null,
+        ''
+    ).add();
+
+    assert.ok(
+        true,
+        'No errors after adding an image without optional parameters, #11756.'
+    );
+});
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (changes requested):
Thanks for the PR!

Although this fix is valid, we always try to keep the file size down, and since we fix many issues every day, every byte matters. 

This bug is not a critical one, but it is inconvenient that an error pops up when doing what we describe in the docs. However, a full typechecking for numbers I think is going too far for an edge case like this. Plus it is quite bulky (why not set `attribs.x`, `attribs.y` etc directly instead of going via `param`?) 

I suggest looping over the `arguments`. I expect it to be smaller, but only testing will show:

```ts
        const args = arguments

        ...

        // Optional properties
        (['x', 'y', 'width', 'height'] as (keyof SVGAttributes)[])
            .forEach((prop, i): void => {
                if (args[i + 1]) { // if we want to be stricter, we can do isNumber(args[i + 1])
                    attribs[prop] = args[i + 1];
                }
            });
```

TorsteinHonsi:
Thanks! Local tests shows that we could chip off some (compiled) bytes by testing the arguments one by one, so lets see how the last change goes.

TorsteinHonsi (approved):

Issue #16159, opened 2021-08-06
Stock Tools gui: disparity index indicator add dropdown selection for Average type


pawellysy:
# Description of the feature

When creating disparity index indicator through the stock tools gui, you have the option to select, what type of average algorythm you want the indicator to use. The available options are indicators like 'sma', 'ema' etc. but in the current version they should be defined as string. Better solution would be to create dropdown selection with dropdown, with all options written down.

# Library related to the feature
Highcharts Stock

# Proof of Concept/Live example for the feature
Similar to https://github.com/highcharts/highcharts/issues/15497

----

**You can vote for this feature by adding a thumbs-up reaction to this post.**

pawelfus:
*Internal note:*
To be picked up after merging #16097

PR #16347, opened 2021-09-15
enhancement/16159-disparity-index-dropdown 

karolkolodziej:
Added Disparity Index `average` parameter as a dropdown in the Stock Tools popup. See #16159.

### DIFF
--- a/ts/Extensions/Annotations/Popup.ts
+++ b/ts/Extensions/Annotations/Popup.ts
@@ -204,7 +204,7 @@ enum DropdownProperties {
  */
 const dropdownParameters: Highcharts.DropdownParameters = {
     'algorithm-pivotpoints': ['standard', 'fibonacci', 'camarilla'],
-    'average-disparityindex': ['sma', 'ema']
+    'average-disparityindex': ['sma', 'ema', 'dema', 'tema', 'wma']
 };
 
 /* eslint-disable no-invalid-this, valid-jsdoc */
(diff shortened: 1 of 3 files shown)


pawelfus (approved):
Thanks!

ts/Extensions/Annotations/Popup.ts
```diff
@@ -194,15 +194,17 @@ const indexFilter = /\d/g,
...
 }
 
 /**
  * List of available algorithms for the specific indicator.
  * @private
  */
 const dropdownParameters: Highcharts.DropdownParameters = {
-    'algorithm-pivotpoints': ['standard', 'fibonacci', 'camarilla']
+    'algorithm-pivotpoints': ['standard', 'fibonacci', 'camarilla'],
+    'average-disparityindex': ['sma', 'ema']
```
raf18seb, line 207: The available base indicators are:
- sma
- ema
- dema
- tema
- wma

Shouldn't we add all of them?

raf18seb (changes requested):

ts/Extensions/Annotations/Popup.ts
karolkolodziej: I will also change `|` into `,` in documentation because it misled me.
Screenshot 2021-09-16 at 09 54 08

raf18seb: What about this?
If a value is different than `'ema'`, `'dema'`, `'tema'` or `'wma'`, then `'sma'` is used.

raf18seb (approved):
Thanks!

pawellysy (approved):
Looks good!

TorsteinHonsi (approved):

PR #16352, opened 2021-09-16
Allow removing a11y info sections by setting format to null.

oysteinmoseng:
Allow removing accessibility information sections by setting section formats to `null`.
___
This allows for a cleaner DOM in cases where one or more of these sections are not needed.

See [accessibility.screenReaderSection.beforeChartFormat](https://api.highcharts.com/highcharts/accessibility.screenReaderSection.beforeChartFormat).


TorsteinHonsi (changes requested):
Instead of `null`, can we make an empty string remove the info sections? The typing would be simpler as it would always be a string rather than `string|null`.

oysteinmoseng:
Nice, changed. Also found a semi-related unreleased bug, so fixed that too

TorsteinHonsi (approved):

PR #16359, opened 2021-09-17
bugfix/16041-flag-distribute-maxDistance 

### LINKED ISSUE #16041: Flag series is hidden when chart width is fixed when text is long
#### Expected behaviour
Flag series is hidden when chart width is fixed when text is long.

#### Actual behaviour
Flag series should show up, may be with some ellipsis text overflow.

https://jsfiddle.net/07zok3t8/

On commenting out line no 3 in JS code, we can see flag series showing up.

#### Product version
Highstock JS v9.1.2 

#### Affected browser(s)
Chrome

karolkolodziej:
Fixed #16041, long flag labels were not distributed properly which resulted in removing them.


samples/unit-tests/series-flags/general/demo.js
```diff
@@ -201,4 +201,36 @@ QUnit.test('Scrolling inverted chart with a flag series.', function (assert) {
...
+    const chart = Highcharts.chart('container', {
+        chart: {
+            width: 800
+        },
+        series: [{
+            data: [
+                [1, 10],
+                [6, 60]
+            ],
+            type: 'line'
```
raf18seb, line 216: This is the default, not needed

samples/unit-tests/series-flags/general/demo.js
```diff
@@ -201,4 +201,36 @@ QUnit.test('Scrolling inverted chart with a flag series.', function (assert) {
...
+            }],
+            type: "flags"
+        }]
+    });
+
+    assert.ok(
+        chart.series[1].points[0].graphic.x > 0,
+        'The flag graphic should have a positive x position.'
+    );
+    chart.setSize(400, undefined);
```
raf18seb, line 230: The second argument `undefined` is the default one, not needed here, but it can stay of course.

samples/unit-tests/series-flags/general/demo.js
```diff
@@ -201,4 +201,36 @@ QUnit.test('Scrolling inverted chart with a flag series.', function (assert) {
...
+                [1, 10],
+                [6, 60]
+            ],
+            type: 'line'
+        }, {
+            data: [{
+                x: 5,
+                title: "Very long long text very long long text"
+            }],
+            type: "flags"
```
raf18seb, line 222: Double quote

samples/unit-tests/series-flags/general/demo.js
```diff
@@ -201,4 +201,36 @@ QUnit.test('Scrolling inverted chart with a flag series.', function (assert) {
...
+        series: [{
+            data: [
+                [1, 10],
+                [6, 60]
+            ],
+            type: 'line'
+        }, {
+            data: [{
+                x: 5,
+                title: "Very long long text very long long text"
```
raf18seb, line 220: Double quote

raf18seb (changes requested):

pawelfus (approved):
Please resolve Rafal's comments, the rest looks good, thanks!

raf18seb (approved):
Thanks!

pawellysy (approved):

TorsteinHonsi (changes requested):
I think 300px is too distant in most situations. See https://jsfiddle.net/highcharts/xLyzbt16/1/ and test it with the PR.

What if we analyze the boxes and look for the widest box? Then we could set the `maxDistance` to something like `Math.min(greatestBoxWidth, 100)`?

karolkolodziej:
Setting `Math.min(greatestBoxWidth, 100)` won't work in this case. 
For smaller flags, numbers are so small that they disappear.

I decided to use `Math.max..` to preserve the previous behaviour and for the longer labels distribute them properly 😄

TorsteinHonsi:
> won't work in this case

Yes of course, `Math.max` was what I meant 🙃

PR #16360, opened 2021-09-17
Migration: Global win to window, restricted strict mode

bre1470:
Migrated global win to window. This will mask the global window object in every Highcharts bundle, so that always `Highcharts.win === window`.

To do:
- [x] Release assembler v1.4.0 with changed UMD variable
- [x] Switch dependency in this branch to v1.4.0 (package.json & package-lock.json)


bre1470:
[ ] Release assembler v1.4.0 with changed UMD variable
[ ] Switch dependency in this branch to v1.4.0 (package.json & package-lock.json)

pawelfus (approved):
It's ok, I guess?

TorsteinHonsi (changes requested):
1.5 kB increase in compiled+gzipped bundle size, just over this small refactoring? We need to find out where that comes from.

TorsteinHonsi (approved):

PR #16361, opened 2021-09-17
enhancement/control-point-cursor 

karolkolodziej:
Enhancement changed the cursor for the annotation control point.


ts/Extensions/Annotations/Types/Measure.ts
```diff
@@ -448,9 +448,16 @@ class Measure extends Annotation {
...
     public addControlPoints(): void {
+        const options = this.options.controlPointOptions;
         let selectType = this.options.typeOptions.selectType,
             controlPoint;
 
+        if (selectType === 'x') {
+            options.style.cursor = 'ew-resize';
+        } else if (selectType === 'y') {
+            options.style.cursor = 'ns-resize';
+        }
```
pawelfus, line 459: I think this should be changed only when a developer has not defined `style.cursor`

pawelfus (changes requested):
Just a minor thing, otherwise looks good 👍

ts/Extensions/Annotations/Types/Measure.ts
```diff
@@ -452,10 +453,15 @@ class Measure extends Annotation {
         let selectType = this.options.typeOptions.selectType,
             controlPoint;
 
-        if (selectType === 'x') {
-            options.style.cursor = 'ew-resize';
-        } else if (selectType === 'y') {
-            options.style.cursor = 'ns-resize';
+        if (
+            !defined(this.userOptions.controlPointOptions &&
+                this.userOptions.controlPointOptions.style.cursor)
```
pawelfus, line 458: The value in `options.style.cursor` can be set in `controlPointOptions` (general, for all control points), or in `contrPoints` array. 

Perhaps it will be better (and easier) to check if the option is already set in the object? Like this:
```js
if (!defined(options.style.cursor)) {
   ...
}
```

pawelfus (changes requested):

pawellysy (approved):
Looks good!

ts/Extensions/Annotations/Types/Measure.ts
karolkolodziej: Since the `cursor` is already defined in the `controlPointOptions`, this condition always fails.

ts/Extensions/Annotations/Types/Measure.ts
```diff
@@ -448,9 +449,21 @@ class Measure extends Annotation {
...
 
     public addControlPoints(): void {
+        const options = this.options.controlPointOptions;
         let selectType = this.options.typeOptions.selectType,
             controlPoint;
 
+        if (
+            !defined(this.userOptions.controlPointOptions &&
+                this.userOptions.controlPointOptions.style.cursor)
+        ) {
```
raf18seb, line 459: I was wondering if we could improve the bracket position here to make it a little bit more readable, maybe like this?:
```js
if (!defined(
    this.userOptions.controlPointOptions &&
    this.userOptions.controlPointOptions.style.cursor
)) {
```

raf18seb (changes requested):
There should be a cursor swap for a measure in the **inverted** chart

ts/Extensions/Annotations/Types/Measure.ts
pawelfus: Aaah, right, fine then! 👍

pawelfus (approved):

raf18seb (approved):
Thanks!

TorsteinHonsi (approved):

PR #16365, opened 2021-09-20
Migration: OnSeries

bre1470:
- Moved OnSeries mixin as OnSeriesComposition.
- Improved typing of stableSort.


TorsteinHonsi (approved):
Thanks!

Issue #16358, opened 2021-09-17
Tools: gulp filesize eats last stash when no edits


TorsteinHonsi:
#### Expected behaviour
When there are no changes compared to HEAD, the script should stop running with a friendly message.

#### Actual behaviour
When there are no changes, a new stash is not created. However, the script continues to the point where it applies _and deletes_ the last stash. 

#### Live demo with steps to reproduce
1. Create a bogus stash, or at least ake sure your last stash is not important.
2. Make sure the working tree is clean.
3. Run `gulp filesize`

goransle:
Should have been closed when https://github.com/highcharts/highcharts/pull/16374 was merged

PR #16374, opened 2021-09-20
Tools: fixed filesize script eating stash and added TS compilation 

### LINKED ISSUE #16357: Tools: gulp filesize not working
#### Expected behaviour
When there are local changes, `gulp filesize` should compare against HEAD.

#### Actual behaviour
It always results in 0B changed.

Load a clean master. Remove an expendable code block, like this one (which is soon to be modified, but still expendable:

https://github.com/highcharts/highcharts/blob/8065af38c4530c7f0643f3a5d855fbd6a4a94766/ts/Core/Renderer/SVG/SVGRenderer.ts#L1270-L1278

Now run `gulp filesize`.

goransle:
Fixed `gulp filesize` always popping the latest stash by using a named stash. Also runs `scripts-ts` if `scripts-watch` is not running.  Fixes #16357 and #16358.

### DIFF
--- a/tools/gulptasks/unsorted/filesize.js
+++ b/tools/gulptasks/unsorted/filesize.js
@@ -1,4 +1,8 @@
 /* eslint func-style: 0, no-console: 0 */
+const { randomUUID } = require('crypto');
+const processLib = require('../lib/process');
+const logLib = require('../lib/log');
+
 const gulp = require('gulp');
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (approved):
Thanks! The failing Cypress test is fixed in master.

Issue #13150, opened 2020-03-16
Improve navigation.bindings docs


pawelfus:
1. Main description is suggesting only exporting menu: https://api.highcharts.com/highcharts/navigation

> navigation
>
> A collection of options for buttons and menus appearing in the exporting module.

2. Only demo in Highstock is available, add others:
- https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/stock/stocktools/basic-gui
- https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/stock/stocktools/custom-popup

3. Split document into two separate docs (Stock Tools and Annotations GUI / bindings):
- https://www.highcharts.com/docs/stock/stock-tools - note that using bindings we can add annotations, as well as modify chart (e.g. change series type, enable/disable zooming etc).

4. Link the article above in API.

raf18seb:
We could also add more simple demos e.g.:
- custom navigation buttons in Highcharts
- demo for _navigation.bindings_ - _init/steps/end_

karolkolodziej:
Documentation will be splited under a separate task: #16392

shayded-exe:
Why is this feature not mentioned anywhere in the annotations documentation page?

It's hard to discover that this feature even exists at all.

pawelfus:
Hello @shayded-exe 

Good point @shayded-exe! At this moment, `navigation.bindings` is described only in [Stock Tools docs](https://www.highcharts.com/docs/stock/stock-tools), but should also get its own documentation page, mentioned in the Annotations page. Thank you for pointing this out! Created a separate ticket for that: https://github.com/highcharts/highcharts/issues/21227

PR #16378, opened 2021-09-21
docs/13150-navigation-bindings-docs

karolkolodziej:
Enhancement #13150, improved `navigation.bindings` documentation.

To do:

- [x] Main description is suggesting only exporting menu
- [x] Only demo in Highstock is available, add others

### DIFF
--- a/samples/highcharts/annotations/bindings-custom-annotation/demo.html
+++ b/samples/highcharts/annotations/bindings-custom-annotation/demo.html
@@ -1,0 +1,6 @@
+<script src="https://code.highcharts.com/highcharts.js"></script>
+<script src="https://code.highcharts.com/stock/modules/annotations-advanced.js"></script>
+
+
+<div id="container"></div>
+<button id="add-annotation">Add annotation</button>
--- a/samples/highcharts/annotations/bindings-custom-annotation/demo.html
+++ b/samples/highcharts/annotations/bindings-custom-annotation/demo.html
@@ -1,6 +1,5 @@
 <script src="https://code.highcharts.com/highcharts.js"></script>
 <script src="https://code.highcharts.com/stock/modules/annotations-advanced.js"></script>
 
-
 <div id="container"></div>
 <button id="add-annotation">Add annotation</button>
--- a/samples/highcharts/annotations/bindings-custom-annotation/demo.html
+++ b/samples/highcharts/annotations/bindings-custom-annotation/demo.html
@@ -1,5 +1,7 @@
 <script src="https://code.highcharts.com/highcharts.js"></script>
-<script src="https://code.highcharts.com/stock/modules/annotations-advanced.js"></script>
+<script src="https://code.highcharts.com/modules/annotations-advanced.js"></script>
 
 <div id="container"></div>
-<button id="add-annotation">Add annotation</button>
+<div class="tools-container">
+  <button class="highcharts-threshold-annotation">Add annotation</button>
+</div>
--- a/samples/highcharts/annotations/bindings-custom-annotation/demo.js
+++ b/samples/highcharts/annotations/bindings-custom-annotation/demo.js
@@ -1,0 +1,78 @@
+Highcharts.chart('container', {
+    navigation: {
+        bindings: {
+            thresholds: {
+                className: 'highcharts-threshold-annotation',
+                start: function (event) {
+                    var chart = this.chart,
+                        x = chart.xAxis[0].toValue(event.chartX),
+                        y = chart.yAxis[0].toValue(event.chartY),
+                        colors = chart.options.colors,
+                        series = chart.series[0],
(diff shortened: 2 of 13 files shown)


ts/Extensions/Annotations/NavigationBindings.ts
```diff
@@ -1257,8 +1257,14 @@ setOptions({
          * - `end`: last event to be called after last step event
          *
          * @type         {Highcharts.Dictionary<Highcharts.NavigationBindingsOptionsObject>|*}
+         *
          * @sample       stock/stocktools/stocktools-thresholds
          *               Custom bindings in Highcharts Stock
+         * @sample       highcharts/annotations/bindings/
+         *               Simple bingind without Stock Tools
+         * @sample       highcharts/annotations/bindings-custom-annotation/
+         *               Custom annotation binding without Stock Tools
```
pawelfus, line 1266: 1. I think there's no need to mention it it without Stock Tools. Simply add `@sample {highcharts} highcharts/...` so it will be available only under Highcharts API 
2. Could you also add the product to the first demo? So the Stock demo will be available only in Highcharts Stock API. Like this:
```
         * @sample {highstock} stock/stocktools/stocktools-thresholds
         *               Custom bindings in Highcharts Stock
```

ts/Extensions/Annotations/NavigationBindings.ts
```diff
@@ -1257,8 +1257,14 @@ setOptions({
          * - `end`: last event to be called after last step event
          *
          * @type         {Highcharts.Dictionary<Highcharts.NavigationBindingsOptionsObject>|*}
+         *
          * @sample       stock/stocktools/stocktools-thresholds
          *               Custom bindings in Highcharts Stock
+         * @sample       highcharts/annotations/bindings/
+         *               Simple bingind without Stock Tools
```
pawelfus, line 1264: ing*

pawelfus (changes requested):

pawellysy (approved):

samples/highcharts/annotations/bindings-custom-annotation/demo.html
```diff
@@ -0,0 +1,6 @@
+<script src="https://code.highcharts.com/highcharts.js"></script>
+<script src="https://code.highcharts.com/stock/modules/annotations-advanced.js"></script>
+
+
```
raf18seb, line 4: unnecessary line break

samples/highcharts/annotations/bindings-custom-annotation/demo.js
```diff
@@ -0,0 +1,78 @@
...
+                        series = chart.series[0],
+                        zones = series.userOptions.zones || [];
+
+                    chart.customColorIndex = chart.customColorIndex || 1;
+
+                    chart.customColorIndex++;
+
+                    if (
+                        chart.customColorIndex === colors.length
+                    ) {
```
raf18seb, line 20: Can be shorted to 1 line

raf18seb (changes requested):
What about migrating to ES6 wherever we can in those demos?

Also, some time ago me and Paweł created a support demo: https://jsfiddle.net/BlackLabel/3cqjn7bz/
Maybe you'll make use of it.

Additionally, see the comments below

karolkolodziej:
I've added your requested changes.

When it comes to the demo, it looks cool however the idea behind that issue was to show that annotations could be created without stock and stock tools. But maybe in the future, we might be able to reuse it ;)

samples/highcharts/annotations/bindings-custom-annotation/demo.js
```diff
@@ -0,0 +1,76 @@
...
+        data: [2, 5, 1, 6, 7, 8, 5]
+    }]
+}, function () {
+    const chart = this;
+    document.getElementById('add-annotation').addEventListener('click', function (e) {
+        const button = this,
+            annotationOptions =
+                chart.navigationBindings.options.bindings.thresholds;
+
+        chart.navigationBindings.bindingsButtonClick(
```
pawelfus, line 70: Hmm.. I'm not sure if we want to recommend `bindingsButtonClick` - it's not part of the API. Wouldn't it be better to do the same as in the `highcharts/annotations/bindings` demo? So simply setting up a container for tools, then adding a button with proper classname?

pawelfus (commented):
See my comment below

samples/highcharts/annotations/bindings-custom-annotation/demo.js
karolkolodziej: Yep! It seems easier and more reasonable! ;)

raf18seb (approved):
Thanks!

pawelfus (approved):
Thanks!

TorsteinHonsi (approved):
Thanks!

PR #16381, opened 2021-09-21
Website/icons

nasin:
Hi, these are the animated "icons to charts" for Highcharts, Stock, Maps and Gantt.

In the highcharts utils sample viewer, the map won't load when the demo is in the iframe. Same for the external buttons in the Gantt chart. I don't know what I'm doing wrong, but both of them work in the mobile view (outside of the iframe.)

### DIFF
--- a/samples/highcharts/website/icon-map/demo.js
+++ b/samples/highcharts/website/icon-map/demo.js
@@ -20,7 +20,7 @@ Math.easeOutBounce = pos => {
 
 const big = window.matchMedia("(min-width: 500px)").matches;
 
-
+const imgPath = 'https://cdn.jsdelivr.net/gh/highcharts/highcharts@feb8baf043cffb5e141ab065f95b8ca397569297/samples/graphics/homepage/';
 const maps = {
     chart: {
         animation: {
@@ -183,7 +183,7 @@ const maps = {
         enabled: false
     },
     tooltip: {
-        enabled: true
+        enabled: false
     },
     plotOptions: {
         series: {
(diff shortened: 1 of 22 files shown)


samples/highcharts/website/icon-map/demo.js
```diff
@@ -0,0 +1,685 @@
...
+                            }
+                        }
+                    },
+                    {
+                        condition: {
+                            minWidth: 499
+                        },
+                        chartOptions: {
+                            subtitle: {
+                                text: 'Click a country to zoom to it. Use buttons below map for selected tests.'
```
TorsteinHonsi, line 666: Which buttons? The demo doesn't have buttons below the map - maybe the website does?

TorsteinHonsi (changes requested):
Awesome!

In addition to the inline question, the stock and maps demos throw an error 15, that the data should be sorted. Non-sorted data may result in problems with the tooltip.

samples/highcharts/website/icon-map/demo.js
nasin: The buttons below the interactive gantt demo on the website (this is the same demo, just styled.) 

I think what is happening is there's a 'styledMode' check script that runs in the utils frame. I'm loading one chart in the utils frame, then wiping it out and loading another chart, and this causes a JS error, which then causes the buttons not to appear.

In the "mobile" view of the chart in utils, everything works fine, because that styledMode checker script doesn't seem to run.
[
image
](url)

nasin:
Hi, I fixed error 15 on the charts, plus I created reduced-motion versions.

TorsteinHonsi (approved):

PR #16382, opened 2021-09-21
Update charts' names to match up with the categories and goals

mekhatria:
Update the charts' names, for example, from just bar to categorical-composition-bar.


TorsteinHonsi (approved):
Thanks!

PR #16386, opened 2021-09-22
Migration: A11y utils

bre1470:
- Migrated DOMElementProvider to class pattern.
- Migrated EventProvider to class pattern.
- Updated package-lock.


TorsteinHonsi (approved):
Thanks!

Issue #15757, opened 2021-05-20
Heatmap borderRadius


karolkolodziej:
#### Expected behaviour
Apply `borderRadius` to the `heatmap` series and update the API.

#### Actual behaviour
In [heatmap API](https://api.highcharts.com/highcharts/series.heatmap) the `borderRadius` property isn't listed.
However, in version 8.2.0 it worked properly. 

Workaround: https://jsfiddle.net/BlackLabel/tphydk24/

#### Live demo with steps to reproduce
8.2.0: https://jsfiddle.net/u2qn3e87/
8.2.2 and higher : https://jsfiddle.net/u3kfmsLb/

#### Product version
 Highcharts JS v9.1.0 (2021-05-03)

#### Affected browser(s)
All

khlieng:
Caused by #12508 in 8.1.0.

PR #16388, opened 2021-09-22
bugfix/16165-borderRadius-disappear-hover

### LINKED ISSUE #16165: "borderRadius" disappears after hovering the pointer over plot in heatmap
Hello! 
Faced such a problem:

#### Expected behaviour
"borderRadius" doesn't disappear 

#### Actual behaviour
If you set the "borderRadius" of the plots, then after hovering over them with the pointer, the rounding disappears 
(https://api.highcharts.com/highcharts/series.heatmap.borderRadius)

Live: https://jsfiddle.net/Lpsb2c1v/6/
1. Move the pointer over some plot
2. Move the pointer away

#### Product version
Version: 9.1.2

#### Affected browser(s)
OSX
Firefox: 90.0.2
Chrome: 89.0.4389.114

K009:
Fixed #16165, border radius disappeared after hovering a point in heatmap.

### DIFF
--- a/samples/unit-tests/series-heatmap/members/demo.js
+++ b/samples/unit-tests/series-heatmap/members/demo.js
@@ -211,9 +211,19 @@ QUnit.test('seriesTypes.heatmap.pointClass.setState', function (assert) {
         0,
         'When state:normal zIndex is 0'
     );
+    assert.strictEqual(
+        point.graphic.d.split(' ')[1] - point.graphic.getBBox().x,
+        point.series.options.borderRadius,
+        'Border radius should have the value set in options on state:normal'
+    );
 
     setState.call(point, 'hover');
     assert.strictEqual(point.graphic.zIndex, 1, 'When state:hover zIndex is 1');
+    assert.strictEqual(
+        point.graphic.d.split(' ')[1] - point.graphic.getBBox().x,
+        point.series.options.borderRadius,
+        'Border radius should have the value set in options on state:hover'
+    );
 
     setState.call(point, 'select');
     assert.strictEqual(
@@ -223,19 +233,6 @@ QUnit.test('seriesTypes.heatmap.pointClass.setState', function (assert) {
     );
 
     setState.call(point, '');
-    assert.strictEqual(
-        point.graphic.d.split(' ')[1] - point.graphic.getBBox().x,
-        point.series.options.borderRadius,
-        'Border radius should have the value set in options on state:normal'
-    );
-
-    setState.call(point, 'hover');
-    assert.strictEqual(
-        point.graphic.d.split(' ')[1] - point.graphic.getBBox().x,
-        point.series.options.borderRadius,
-        'Border radius should have the value set in options on state:hover'
-    );
-
     assert.strictEqual(
         point.graphic.d.split(' ')[1] - point.graphic.getBBox().x,
         point.series.options.borderRadius,
--- a/samples/unit-tests/series-heatmap/members/demo.js
+++ b/samples/unit-tests/series-heatmap/members/demo.js
@@ -219,3 +219,46 @@ QUnit.test('seriesTypes.heatmap.pointClass.setState', function (assert) {
         'When state:select zIndex is 0'
     );
 });
+
+QUnit.test('Check if borderRadius is correct according to user settings', function (assert) {
+    const chart = new Highcharts.Chart('container', {
+            chart: {
+                type: 'heatmap'
+            },
+
+            series: [{
+                borderWidth: 1,
+                data: [
+                    [0, 0, 10],
+                    [0, 1, 19],
+                    [0, 2, 8],
+                    [0, 3, 24]
+                ],
+                borderRadius: 39
+            }]
+        }),
+        point = chart.series[0].points[0],
+        setState = Highcharts.seriesTypes.heatmap.prototype
+            .pointClass.prototype.setState;
+
+    setState.call(point, '');
+    assert.strictEqual(
+        point.graphic.r,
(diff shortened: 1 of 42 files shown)


samples/unit-tests/series-heatmap/members/demo.js
```diff
@@ -219,3 +219,46 @@ QUnit.test('seriesTypes.heatmap.pointClass.setState', function (assert) {
         'When state:select zIndex is 0'
     );
 });
+
+QUnit.test('Check if borderRadius is correct according to user settings', function (assert) {
```
raf18seb, line 223: Usually, we don't want to create a separate test with a new chart's instance, because creating thousands of new charts during the tests increases theirs time. Can we reuse the above test that tests something very similar?

raf18seb (changes requested):
I'm thinking whether we could add the possibility for setting the `borderWidth` and `borderRadius` per point here. It's a very popular users request in heatmap.

Something like
```js
if (
    this.options.borderRadius ||
    point.options.borderRadius
) {
    point.graphic.attr({
        r: point.options.borderRadius || this.options.borderRadius
    });
}
if (point.shapeArgs) {
    point.shapeArgs.r = point.options.borderRadius || this.options.borderRadius;
}
```

Also, please see comments below.

ts/Series/Heatmap/HeatmapSeries.ts
```diff
@@ -440,6 +440,10 @@ class HeatmapSeries extends ScatterSeries {
                         });
                     }
 
+                    // Applying borderRadius from options (#16165)
+                    (point.shapeArgs || {}).r = this.options.borderRadius;
+                    (point.shapeArgs || {}).d = point.graphic.pathArray;
```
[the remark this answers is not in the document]
K009: Yes, without that the `d` value doesn't match the actual shape of path.

samples/unit-tests/series-heatmap/members/demo.js
K009: Sure 👍

samples/unit-tests/series-heatmap/members/demo.js
```diff
@@ -210,12 +211,34 @@ QUnit.test('seriesTypes.heatmap.pointClass.setState', function (assert) {
...
+        'Border radius should have the value set in options on state:normal'
+    );
+
+    setState.call(point, 'hover');
+    assert.strictEqual(
+        point.graphic.d.split(' ')[1] - point.graphic.getBBox().x,
+        point.series.options.borderRadius,
+        'Border radius should have the value set in options on state:hover'
+    );
+
```
raf18seb, line 238: I think `setState.call(point, '');` is missing here

samples/unit-tests/series-heatmap/members/demo.js
```diff
@@ -210,12 +211,34 @@ QUnit.test('seriesTypes.heatmap.pointClass.setState', function (assert) {
...
     );
+
+    setState.call(point, '');
+    assert.strictEqual(
+        point.graphic.d.split(' ')[1] - point.graphic.getBBox().x,
+        point.series.options.borderRadius,
+        'Border radius should have the value set in options on state:normal'
+    );
+
+    setState.call(point, 'hover');
```
raf18seb, line 232: Instead of setting these states again, can we reuse the previouse `setState.call(point, '');` and `setState.call(point, 'hover');`?

So, instead of doing `setState.call(...` 6 times, we could do it 4 times.

samples/unit-tests/series-heatmap/members/demo.js
```diff
@@ -210,12 +211,34 @@ QUnit.test('seriesTypes.heatmap.pointClass.setState', function (assert) {
...
+    assert.strictEqual(
+        point.graphic.d.split(' ')[1] - point.graphic.getBBox().x,
+        point.series.options.borderRadius,
+        'Border radius should have the value set in options on state:hover'
+    );
+
+    assert.strictEqual(
+        point.graphic.d.split(' ')[1] - point.graphic.getBBox().x,
```
raf18seb, line 240: Why `point.graphic.d.split(' ')[1] - point.graphic.getBBox().x` instead of `point.graphic.r`?

raf18seb (changes requested):
Few questions, see below comments

samples/unit-tests/series-heatmap/members/demo.js
K009: that's true, thanks!

samples/unit-tests/series-heatmap/members/demo.js
K009: Yes, it should work. Thanks 👍

samples/unit-tests/series-heatmap/members/demo.js
K009: Unfortunately, `point.graphic.r` had the wrong (old) value and couldn't be used for the test. That's why I calculated it in a different way.

raf18seb (approved):
Good for me, thanks!

TorsteinHonsi:
I think we need to dig deeper. The fix for #15757 was already a patch for something that should be handled in `translate` or `pointAttribs`/`markerAttribs`, and now we're creating a patch for the patch. I've create a commit where it is added to `shapeArgs` directly, but had to modify `Series.drawPoints`. Let's see how the visual tests go.

TorsteinHonsi:
The visual tests failed. I found myself digging deeper and deeper until the depth of the `SVGRenderer.symbol` function. So resolving that would be time consuming and dangerous. I added a comment so that we can consider refactoring this at a later stage.

PR #16402, opened 2021-09-24
bugfix/15543-seriesthreshold-default-value

### LINKED ISSUE #15543: Docs: seriesThreshold wrong default value
#### Expected behaviour
In the API it is said that _boost.seriesThreshold_ value is set to null by default.
https://api.highcharts.com/highcharts/boost.seriesThreshold

#### Actual behaviour
This property is set to 50 by default and the description in API may confuse users.

mfilipiec:
Fixed #15543, setting a proper default value for seriesThreshold in docs

### DIFF
--- a/ts/Extensions/Boost/BoostOptions.ts
+++ b/ts/Extensions/Boost/BoostOptions.ts
@@ -113,7 +113,7 @@ declare global {
  * amount of series.
  *
  * @type      {number|null}
- * @default   null
+ * @default   50
  * @apioption boost.seriesThreshold
  */
 


KacperMadej (commented):
Is `null` ever a valid value for `seriesThreshold`?
`pick` in the code below will use `50` in case `null` is set:
https://github.com/highcharts/highcharts/blob/5ffa5328634c09f3d4978a82730e7091a14379d9/ts/Extensions/Boost/BoostOverrides.ts#L118-L129

TorsteinHonsi (changes requested):
Kacper is right, `null` is never used. So it should be removed also from the `@type` in the doclet, and from the TypeScript declaration.

PR #16403, opened 2021-09-24
Website/new homepage

nasin:
Hi,

Here's the new homepage animation. I know the first bubble chart is all squished on mobile, but it's looks really cool. I can certainly fix it if you want.

The "stop animation" link will be in the homepage HTML.

### DIFF
--- a/samples/highcharts/website/new-homepage/demo.css
+++ b/samples/highcharts/website/new-homepage/demo.css
@@ -1,0 +1,365 @@
+@import 'https://code.highcharts.com/css/highcharts.css';
+
+.hero-container {
+	background: rgb(79,70,90);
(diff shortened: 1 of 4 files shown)


TorsteinHonsi (changes requested):
Beautiful, great work! Only a few, non-critical remarks:

- There's a moment before and after the sankey animation where the lines are bulging inwards - is that intentional?
Skjermbilde 2021-09-30 kl  09 54 06

- Maybe we should pick another name and `new-homepage`. In some years, if we start working on another redesign, this will not be the new one... I suggest `homepage-2021`.

- The indenting in `demo.css` is a bit messy.

TorsteinHonsi (approved):
Looks great!

PR #16410, opened 2021-09-27
bugfix/16316-annotation-measue-with-ordinal 

### LINKED ISSUE #16316: Annotation with measure type "y" does not work when the data has weekends
#### Expected behaviour
I'm expected the below behaviour as shown in the documentation
https://jsfiddle.net/Lindus93/jxm17dnz/

#### Actual behaviour
in this configuration i only had one week and one monday for the next week and it not works but if i remove the last field the yAxis measure works.

https://jsfiddle.net/e257nqzx/

https://www.highcharts.com/demo/stock/stock-tools-custom-gui

PD: i try to put the xAxis breaks for this weekend and the line for the measure works but the measure data is not showing

https://www.highcharts.com/demo/stock/stock-tools-custom-gui

#### Product version
Highcharts/ Highstock 9.1.2

#### Affected browser(s)
All

karolkolodziej:
Fixed #16316, the position of measure annotation was not calculated properly on the ordinal axis.

### DIFF
--- a/test/cypress/integration/Highcharts/stock-tools.spec.js
+++ b/test/cypress/integration/Highcharts/stock-tools.spec.js
@@ -181,3 +181,36 @@ describe('An indicator on indicator, #15696.', () => {
             .should('not.contain', 'SMA (14)')
     });
 });
+
+describe('Measure annotation, #15696.', () => {
+    beforeEach(() => {
+        cy.viewport(1000, 800);
+    });
+
+    before(() => {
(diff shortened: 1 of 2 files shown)


pawelfus (approved):

TorsteinHonsi (approved):
Thanks! Pending approval by @pawellysy .

pawellysy (approved):

raf18seb (approved):

PR #16413, opened 2021-09-28
Migration: AccessibilityComponent

bre1470:
Migrated AccessibilityComponent to class pattern.


bre1470:
@oysteinmoseng I wonder in this case, if the dummy functions should actually become abstract class members with the class itself also abstract.

oysteinmoseng:
@bre1470 Good thinking, but they are not quite dummy functions, but more like default functions that can be overridden. So better keep it as is I think.

TorsteinHonsi (changes requested):
Thanks! Please resolve conflicts.

TorsteinHonsi (approved):
Thanks! Pending approval by @oysteinmoseng

oysteinmoseng (approved):
Looks great!

PR #16419, opened 2021-09-29
Tools: removed demos deploy from nightly

goransle:
This removes the demo deploy from the nightly. Will replace the trigger with a gulp task after tweaking some things on the demo-manager end.


TorsteinHonsi (approved):
Thanks!

PR #16432, opened 2021-10-01
Tools: Cleaned up import paths for webpack.

bre1470:


TorsteinHonsi (approved):
Thanks! And enter the Foundation 👍

PR #16437, opened 2021-10-04
bugfix/16275-number-precision-float

### LINKED ISSUE #16275: Line chart y-axis range is not computed correctly if the data point values are large numbers with small differences
#### Expected behavior
Data points [238863224762451, 238863224762452, 238863224762453, 238863224762453] can be shown correctly.

#### Actual behaviour
Only the first data point is shown.
image

http://jsfiddle.net/6dam1fjr/

#### Product version
Highcharts JS v9.2.2 (2021-08-24)

#### Affected browser(s)
Can be repro on Chrome/Edge, other browsers should be affected too.

K009:
Fixed #16275, ticks on the axis weren't displayed when their value was too high.

### DIFF
--- a/ts/Core/Axis/Axis.ts
+++ b/ts/Core/Axis/Axis.ts
@@ -1053,18 +1053,21 @@ class Axis {
         min: number,
         max: number
     ): Array<number> {
-        const roundedMin = correctFloat(Math.floor(min / tickInterval) * tickInterval),
-            roundedMax = correctFloat(Math.ceil(max / tickInterval) * tickInterval),
-            tickPositions = [];
+        const tickPositions = [];
 
         let pos,
             lastPos,
-            precision;
+            precision,
+            roundedMin = correctFloat(Math.floor(min / tickInterval) * tickInterval),
+            roundedMax = correctFloat(Math.ceil(max / tickInterval) * tickInterval);
 
         // When the precision is higher than what we filter out in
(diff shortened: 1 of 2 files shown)


ts/Core/Axis/Axis.ts
```diff
@@ -1053,18 +1053,21 @@ class Axis {
...
         let pos,
             lastPos,
-            precision;
+            precision,
+            roundedMin = correctFloat(Math.floor(min / tickInterval) * tickInterval),
+            roundedMax = correctFloat(Math.ceil(max / tickInterval) * tickInterval);
 
         // When the precision is higher than what we filter out in
         // correctFloat, skip it (#6183).
+        // Comment TODO (#16275)
```
KacperMadej, line 1066: TODO

TorsteinHonsi (changes requested):
This fix is based on my drop-in workaround, not on my [suggested fix](https://github.com/highcharts/highcharts/issues/16275#issuecomment-911339066), which is smaller, but not available as a drop-in workaround. 

Is there a specific reason for that, or is it just a misunderstanding?

KacperMadej:
My bad. The latter from me and I have suggested this to Dominik. My initial thought was that we want to keep some kind of precision to avoid floating point rounding errors even for bigger numbers.

KacperMadej (approved):

TorsteinHonsi (approved):

PR #16462, opened 2021-10-06
bugfix/16415-sma-index-value 

### LINKED ISSUE #16415: SMA's value is wrong, it should use index 3, which is 'close', instead of default 0, which is 'open'
#### Expected behaviour
Correct sma value, the same as python's and the security software's.

#### Actual behaviour
```
In [1]: import json 

In [2]: import requests

In [3]: r = requests.get('https://demo-live-data.highcharts.com/aapl-ohlcv.json')

In [4]: r.json()
In [5]: d = r.json()
In [7]: sum = 0

In [8]: for i in d[-14:]:
   ...:     sum += i[4]
   ...: 

In [10]: sum / 14
Out[10]: 146.98857142857142

The sma(14) on the graph is: 147.8929

In [17]: sum = 0

In [18]: for i in d[-15:]:
    ...:     sum += i[4]
    ...:

In [19]: sum / 15
Out[19]: 147.53

sma(15) on the graph is: 148.4987.
```
Later I wrote a small function to see which data was hs's ma using, and found it used the 'open' value:
```
In [21]: def ma(cnt, idx):
    ...:     sum = 0
    ...:     for i in d[-cnt:]:
    ...:         sum += i[idx]
    ...:     avg = sum / cnt
    ...:     print(avg)
    ...:     return avg
    ...:

In [22]: ma(15, 4)
147.53
Out[22]: 147.53

In [23]: ma(15, 3)
146.40666666666667
Out[23]: 146.40666666666667

In [24]: ma(15, 2)
149.29866666666663
Out[24]: 149.29866666666663

In [25]: ma(15, 1)
148.49866666666668
Out[25]: 148.49866666666668

In [26]: ma(14, 1)                                                                                                                                          
147.89285714285714
Out[26]: 147.89285714285714
```

https://www.highcharts.com/demo/stock/sma-volume-by-price/dark-unica
--> "Edit in codepen"
Modify the "week" in the datagrouping to "day", easy to check it.

#### Product version
highstock 7.1.2 and 9.2.2 both have this problem.
I checked the SMA indicator in the repository, and found it used the 'open' as the default, but we should use the 'close' value as default.

```
https://github.com/highcharts/highcharts/blob/master/ts/Stock/Indicators/SMA/SMAIndicator.ts
        params: {
            /**
             * The point index which indicator calculations will base. For
             * example using OHLC data, index=2 means the indicator will be
             * calculated using Low values.
             */
            index: 0,
            /**
             * The base period for indicator calculations. This is the number of
             * data points which are taken into account for the indicator
             * calculations.
             */
            period: 14
        }
    } as SMAOptions);
```
#### Affected browser(s)
Google chrome 93.

karolkolodziej:
Fixed #16415, changed default SMA `index` parameter, added an `index` to Keltner channels indicator.

 #### Upgrade note
Changed the default SMA `index` parameter from `0` (open) to `3`(close). This also has an impact on the other indicators where SMA is being used, namely  MACD, Price Envelopes and the Linear regression family.

### DIFF
--- a/ts/Core/Axis/OrdinalAxis.ts
+++ b/ts/Core/Axis/OrdinalAxis.ts
@@ -490,7 +490,7 @@ namespace OrdinalAxis {
 
             // Check if the index is inside position array.
             // If true, read/approximate value for that exact index.
-            if (index >= 0 && index < positions.length) {
+            if (index >= 0 && index < positions.length - 1) {
                 const leftNeighbour = positions[Math.floor(index)],
                     rightNeighbour = positions[Math.ceil(index)],
                     distance = rightNeighbour - leftNeighbour;
--- a/ts/Stock/Indicators/KeltnerChannels/KeltnerChannelsIndicator.ts
+++ b/ts/Stock/Indicators/KeltnerChannels/KeltnerChannelsIndicator.ts
@@ -61,6 +61,12 @@ class KeltnerChannelsIndicator extends SMAIndicator {
      */
     public static defaultOptions: KeltnerChannelsOptions = merge(SMAIndicator.defaultOptions, {
         params: {
+            /**
+             * The point index which indicator calculations will base. For
+             * example using OHLC data, index=2 means the indicator will be
+             * calculated using Low values.
+             */
+            index: 0,
             period: 20,
             /**
              * The ATR period.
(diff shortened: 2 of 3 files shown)


pawelfus (commented):
I see a few visual diffs, all related to MACD, Linear regression indicators family, and Price Envelopes.
It's probably the desired change for all of them, but could you please confirm? If we are going to change this, then it should be described in the breaking change too

pawelfus (approved):

pawellysy (approved):

PR #16471, opened 2021-10-07
bugfix/16449-rectangle-missing-side

### LINKED ISSUE #16449: Rectangle drawn with stock tools has only 3 sides, shape is not closed.
#### Expected behaviour
Rectangle drawn with stock tools should have 4 sides.

#### Actual behaviour
Rectangle has only 3 drawn sides, shape is not closed.

https://jsfiddle.net/BlackLabel/3y8zbofk/

#### Product version
9.2.2

#### Affected browser(s)
N/A

pawellysy:
Fixed #16449, the rectangle annotation added via Stock Tools was rendering without one side.


test/cypress/integration/Highcharts/annotations/rectangle/rectange.spec.js
```diff
@@ -0,0 +1,24 @@
...
+        cy.visit('/stock/demo/stock-tools-gui');
+    });
+    it('Rectangle annotation should have closed shape, 16449', () => {
+        cy.get('.highcharts-label-annotation').children().eq(1).click();
+        cy.get('.highcharts-rectangle-annotation').click();
+        cy.get('.highcharts-container')
+            .click(250, 200, { force: true })
+            .click(350, 100, { force: true });
+        cy.chart().then(chart => {
+            const path = chart.annotations[0].shapes[0].graphic.d;
```
pawelfus, line 16: Instead of checking `graphic.d` (which is our property), could you check the actual path in DOM?

PR #16480, opened 2021-10-11
Website/highcharts timeline

nasin:
Hi,

I re-created the 10-year milestone chart in the new style for use on the About Us page.

I'm not sure what to do with it at very small widths.


TorsteinHonsi (changes requested):
Thanks!

> I'm not sure what to do with it at very small widths.

As far as I can see, the only problem is that the flags overlap. You can try experimenting with `allowOverlapX` to make some of them move. 

Another possiblility is to set a `scrollablePlotArea`.

PR #16489, opened 2021-10-13
Enhancement/highcharts themes

nasin:
Hi,

Ferdinand asked me to put together "Highcharts" chart themes based on the new design system for use in the theme drop down on the demo pages.

I took a stab as making the TS files. (I modified existing ones.)


goransle:
I took the liberty to add the master files for these themes, which we need to build the `.js` files that end up at https://code.highcharts.com/themes/dark-unica.js

TorsteinHonsi (changes requested):
Thanks! Looks good. There are some issues after going over the samples in the utils (pull the last version of the utils and choose Theme in the select box).

Now after fixing these issues, I think we should swap the existing DarkUnica theme on the website with this new brand theme, which looks better. But that's a task for the web team.

### BrandDark theme
* http://utils.highcharts.local/samples/#view/highcharts/demo/annotations => the credits text color is too dark on the dark background.
* http://utils.highcharts.local/samples/#view/highcharts/demo/dumbbell => The lower bullets are dark, they should be light on the dark background
* http://utils.highcharts.local/samples/#view/highcharts/demo/timeline => The labels are not themed
* http://utils.highcharts.local/samples/#view/highcharts/demo/treemap-large-dataset => The color axis labels are not visible on the dark background. This can also be observed in some of the map demos.
* http://utils.highcharts.local/samples/#view/maps/demo/all-areas-as-null => The theme should probably set a better `plotOptions.map.nullColor`. It may even be that we should change the product default `nullColor`. The flight-routes demo has `nullColor: 'rgba(200, 200, 200, 0.3)',` which works well for both light and dark themes. Except we need to use a pallette color if we do this.
* All (or most) maps demos: The `mapNavigation` buttons are not dark. I would expect the same style as `rangeSelector.buttonTheme`.
* http://utils.highcharts.local/samples/#view/stock/demo/area =>  The white stroke on the  hover and select range selector buttons makes the text blurry. I don't think it works as intended. This is how to set a white stroke to the button without blurring the text:
```js
select: {
                        fill: '#000003',
                        style: {
                            color: '#fff'
                        },
                        'stroke-width': 1,
                        stroke: 'white'
                    }
```

### BrandLight theme
In general, it looks like this theme has accidentally inherited some of the dark material. Instead, it would be better to start from an existing light theme, which less overrides for labels etc.
* http://utils.highcharts.local/samples/#view/stock/demo/area => Fix scrollbar, range selector etc.
* http://utils.highcharts.local/samples/#view/highcharts/demo/accessible-pie => As above. Data labels are light, they should be dark like in the default theme.

nasin:
Hi,

I made the changes in both theme files. If I missed anything, let me know. Thanks!

nasin:
One other thing...I noticed on the existing dark themes, the map navigation buttons are light as well. I tried to change the buttons to dark gray, but could not figure out how to do it.

TorsteinHonsi (changes requested):
Thanks!
* Here's the docs for map navigation styling: https://api.highcharts.com/highmaps/mapNavigation.buttonOptions.theme
* I also noted that the [export button theme](https://github.com/highcharts/highcharts/pull/16489/commits/27810b057e08cbdcc5c735c5c93a51eab54a3d74#diff-b16c47535aeb0ed6e293a69abb2a99cd201d04e93b20b2fbbbd4954352f87513L90-L105) got removed in the latest commit. Shouldn't it be dark?

nasin:
I updated the map and exporting buttons for the dark theme.

TorsteinHonsi (approved):

PR #16498, opened 2021-10-15
bugfix/16398-incorrect-legend-item-spacing

### LINKED ISSUE #16398: legend item spacing incorrect for empty-name
#### Expected behaviour
legend rows should be spaced the same regardless of name

#### Actual behaviour
providing a blank name for data leads to the legend row icon being "compressed" against the next row.

https://jsfiddle.net/zenzizen/maj9hosg/7/ pie chart
https://jsfiddle.net/zenzizen/maj9hosg/14/ scatterplot

#### Product version
Highcharts & Highstock latest

#### Affected browser(s)
All presumably; checked Firefox, Brave and Chrome

bm64:
Fixed #16398, spacing was incorrect for legend items without name.


TorsteinHonsi:
_Internal note (to self)_
Debugging output shows that the 500+ visual diffs comes from the `bBox.height` being 15 while `fontMetricsH` being 15 on the test machine (Chrome on Linux). On desktop they are both 15, so it may be okay to just ignore this test output. Before that we can do a run where a `bBox.height` of 14 is hard-replaced to 15, to be able to catch side effects.

TorsteinHonsi (approved):
_Conclusion_: In the major browsers on Mac, all visual tests pass, except those with non-default fonts or styled mode. Therefore the 500+ failing tests on CI will not be that big a problem, as we do the final regression test locally anyway. And I agree with Kacper's advice that the new behaviour has the advantage that it is more predictable across systems.

Awaiting CI tests to pass, except visuals.

PR #16499, opened 2021-10-15
Docs/maps mapseries

madepiet:
Adding documentation for map-series in the maps folder:
https://www.highcharts.com/docs/maps/getting-started

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,23055 @@
 {
   "name": "highcharts",
-  "version": "9.2.2",
-  "lockfileVersion": 1,
+  "version": "9.3.0",
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "highcharts",
+      "version": "9.3.0",
+      "license": "SEE LICENSE IN <license.txt>",
+      "dependencies": {
+        "jspdf-yworks": "^2.0.1",
+        "svg2pdf.js": "^1.3.3"
+      },
+      "devDependencies": {
+        "@babel/core": "^7.15.0",
+        "@babel/preset-env": "^7.15.0",
+        "@octokit/rest": "^18.10.0",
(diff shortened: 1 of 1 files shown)


KacperMadej (changes requested):
Looks good. Small adjustments to apply:
- [x] add a simple demo as well (maybe [this one](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/maps/demo/category-map/), or maybe one from [mapData](https://code.highcharts.com/mapdata/), or even another as long as it is a fairly simple one)
- [x] add short info about using `mapData` with links to the complete info in https://www.highcharts.com/docs/maps/getting-started (namely "Load the map" and "Add and join data")
- [x] format options as inline `code` - `mapChart`, `map` (in "default series type is map.") and other is will be added
- [ ] remove package-lock.json from the PR  (if I understand correctly this:  https://highsoft.slack.com/archives/C2SD4K2MN/p1631102679038400 @TorsteinHonsi ?)

madepiet:
> As far as I can see, this new article is just a collection of URLs pointing to API and other articles' paragraphs and demos. I'm not sure, but I think the whole point of writing this article is to collect all the necessary information about the `map` series in one simple article so the users can easily find everything here instead of jumping URLs.
> 
> Is it OK if we leave it as it is (maybe just for now) or should we think about more general refactoring?
> 
> Also, please see my comments below.

I agree with @raf18seb , it would be necessary to find out more about how we should write our documentation. Oftentimes, we are not able to duplicate the same information in many places, so references are a good solution (I mean, for example, initializing the map constructor or loading the map itself). But I wonder if we should approach the creation of specific series in sufficient detail at the level of these documents or leave it, for example, for articles/tutorials - on the principles of how to do / how to create? This arrangement can be helpful and a good guide for creating our next documentation files.

PR #16409, opened 2021-09-27
enhancement/15579-custom-indicator-on-event 

karolkolodziej:
Enhancement #15579, added documentation with an example on how to calculate custom indicators on events.

PR #16507, opened 2021-10-18
bugfix/16397-indicator-recalculation 

### LINKED ISSUE #16397: VBP indicator recalculates only for the first main series update
#### Expected behaviour
I expect the VBP to always follow the main series, no matter how many data updates are.

#### Actual behaviour
VBP recalculates only for the first update. The next main series data updates don't change the VBP indicator.

Click the first button - VBP is recalculating.
Click the second button - VBP is **not** recalculating (but should).
Click the third button - VBP is **not** recalculating (but should).

https://jsfiddle.net/BlackLabel/t1hp89ud/

#### Product version
Highcharts Stock 9.2.2

karolkolodziej:
Fixed #16397, the VBP indicator was not recalculated after adding the point.

- [x] tests
- [x] ~braking change message (after reviewers approve the PR)~ no need for that, we modified internal properties
- [ ] adjust documentation (after reviewers approve the PR) #16409

### DIFF
--- a/ts/Stock/Indicators/VBP/VBPIndicator.ts
+++ b/ts/Stock/Indicators/VBP/VBPIndicator.ts
@@ -828,7 +828,10 @@ interface VBPIndicator {
 extend(VBPIndicator.prototype, {
     nameBase: 'Volume by Price',
     nameComponents: ['ranges'],
-    useAdditionalEvents: true,
+    calculateOn: {
+        chart: 'render',
+        xAxis: 'afterSetExtremes'
+    },
     pointClass: VBPPoint,
     markerAttribs: noop as any,
     drawGraph: noop,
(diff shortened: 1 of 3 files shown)


ts/Stock/Indicators/VBP/VBPIndicator.ts
```diff
@@ -829,10 +829,7 @@ interface VBPIndicator {
 extend(VBPIndicator.prototype, {
     nameBase: 'Volume by Price',
     nameComponents: ['ranges'],
-    bindTo: {
-        series: false,
-        eventName: 'afterSetExtremes'
-    },
+    calculateOnExtremesChagne: true,
```
pawelfus, line 832: Change*

Would it make sense to merge somehow both properties: `calculateOn` and `calculateOnExtremesChange`?

karolkolodziej: I combined them but I'm not sure if this is a correct approach.
Each one is responsible for something else and the main issue that I have with it is finding a proper name for this combined property...

bre1470 (approved):
Looks good! 👍

TorsteinHonsi (approved):

PR #16546, opened 2021-10-26
bugfix/16532-ordinal-regression

### LINKED ISSUE #16532: Updating from 8.2.2 to 9.3.0 causes 10x+ page load times
#### Expected behaviour
When updating to a new version, I'd expect performance to stay the same or be better.

#### Actual behaviour
After updating from Highcharts 8.2.2 to 9.3.0, I noticed our project's pages taking about 10x the time to load than what they did before the update. Each chart on the page went from loading in about 2-5 seconds or less to loading in about 20-45 seconds. In total, the page load time went from about 15 seconds to an unusual 2 minutes. One user reported page load times in excess of 10 minutes.

Both JSFiddles contain the same data from a production environment. Each data series is a two-dimensional array of a timestamp & value, each containing about 60,000 pairs of data. There are 6 series in total, all having the same timestamp values, so a total of about 360,000 data pairs. 8.2.2 loads the chart in about one second, where 9.3.0 takes almost 15 seconds.

I'm aware that the boost.js module exists, but the performance drop is large enough for me to consider this a bug.

8.2.2: https://jsfiddle.net/Kopo942/1bkpvgd0/7/
9.3.0: https://jsfiddle.net/Kopo942/218hg0u9/13/

Run the fiddle manually to get the best results, automatic runs produced odd numbers. Load times are printed in the console.

#### Product version
9.3.0

#### Affected browser(s)
Tested on Windows 10 with latest Chrome (94.0.xx).

karolkolodziej:
Fixed #16532, performance regression in ordinal axis.

### DIFF
--- a/ts/Core/Axis/OrdinalAxis.ts
+++ b/ts/Core/Axis/OrdinalAxis.ts
@@ -1226,14 +1226,6 @@ namespace OrdinalAxis {
                     series.processData.apply(fakeSeries);
                 });
 
-                // Apply grouping if needed.
-                // axis.applyGrouping.call(
-                //     fakeAxis,
-                //     {
-                //         hasExtemesChanged: false
-                //     }
-                // );
-
                 // Force to use the ordinal when points are evenly spaced
                 // (e.g. weeks), #3825.
                 if (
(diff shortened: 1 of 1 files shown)


pawellysy (approved):
image

TorsteinHonsi:
Pending approval by @pawelfus

pawelfus (approved):
Thanks!

PR #16578, opened 2021-11-02
Fixed #15551, colorAxis outside screen.

### LINKED ISSUE #15551: Wrong automatic width for vertical map legend/color axis
#### Expected behaviour
Legend/color axis get correct automatic width 

#### Actual behaviour
Vertical Legend/color axis is too narrow when using longer labels

See http://jsfiddle.net/kzoon/9jqwg3mL/

When aligned right: Legend will only fit when you fiddle around with x
When aligned left: Legend will only fit when you fiddle around with margin

#### Product version
Highmaps any version

DJTechnoo:
Fixed #15551, improved legend width calculation for `colorAxis` with long labels.

~~Fixed #15551, improved colorAxis calculation when using long labels.~~


ts/Core/Axis/Axis.ts
```diff
@@ -3403,6 +3403,11 @@ class Axis {
                 Math.max((clipOffset as any)[invertedSide], clip);
         }
 
+        if (coll === 'colorAxis') {
+            this.chart.legend.render();
+            chart.getMargins(true);
+        }
+
```
TorsteinHonsi, line 3410: In order to avoid polluting the core with code dealing with ColorAxis, check if this code could be in an event handler for `afterGetOffset`. Then that event handler would be added from the ColorAxis module.

ts/Core/Legend/Legend.ts
```diff
@@ -789,15 +789,21 @@ class Legend {
...
-            options.itemWidth ||
+            (options.itemWidth ||
             item.legendItemWidth ||
-            bBox.width + itemExtraWidth;
+            bBox.width + itemExtraWidth) + (
+                (options.layout === 'vertical') &&
+                (item as ColorAxis).maxLabelLength &&
+                ((item as ColorAxis).maxLabelLength > itemExtraWidth) &&
+                ((item as ColorAxis).maxLabelLength - itemExtraWidth) ||
+                0
```
TorsteinHonsi, line 805: Also to avoid polluting the core with code dealing with ColorAxis, check if this code could be moved to the ColorAxis. The legend is already checking for `item.legendItemWidth`. So it should be possible to make a `legendItemWidth` property on the ColorAxis, and set it there in order for the legend to pick it up.

KacperMadej:
TODO: add a test

ts/Core/Axis/Color/ColorAxis.ts
```diff
@@ -572,13 +573,29 @@ class ColorAxis extends Axis implements AxisLike {
...
+        const isSeries = !(item as any).series;
+        const series = !isSeries && (item as any).series.drawLegendSymbol ?
+                (item as any).series :
+                item,
+            seriesOptions = series.options,
+            showCheckbox = legend.createCheckboxForItem &&
+                seriesOptions &&
+                seriesOptions.showCheckbox;
+        const itemExtraWidth = legend.symbolWidth + symbolPadding +
+                        itemDistance + (showCheckbox ? 20 : 0);
```
KacperMadej, line 587: A consistent format would look better - not a mix of `const` in each line and a comma-separated list. Pick one and fight [here](https://highsoft.slack.com/archives/C2SD4K2MN/p1631862207052200).

ts/Core/Axis/Color/ColorAxis.ts
```diff
@@ -572,13 +573,29 @@ class ColorAxis extends Axis implements AxisLike {
         ).attr({
             zIndex: 1
         }).add(item.legendGroup);
-
+        const options = legend.options;
+        const symbolPadding = options.symbolPadding || 0;
+        const isSeries = !(item as any).series;
```
KacperMadej, line 578: The `item` is never a series. `series` property on `item` is an array of series because the `item` is an axis - colorAxis to be precise.

KacperMadej (commented):
The PR doesn't look ready for review, but I am requested for review, so here are some notes.
`¯\_(ツ)_/¯`

ts/Core/Axis/Color/ColorAxis.ts
```diff
@@ -564,21 +565,38 @@ class ColorAxis extends Axis implements AxisLike {
...
-            width,
-            height
-        ).attr({
-            zIndex: 1
-        }).add(item.legendGroup);
-
+        if (!item.legendSymbol) {
+            item.legendSymbol = this.chart.renderer.rect(
+                0,
+                (legend.baseline as any) - 11,
```
KacperMadej, line 571: We could (try to) reduce `as any` in this file while the code is modified.
Here `(legend.baseline || 0)` could be used.

ts/Core/Axis/Color/ColorAxis.ts
```diff
@@ -564,21 +565,38 @@ class ColorAxis extends Axis implements AxisLike {
...
         axis.legendItemWidth = (
             width +
             padding +
-            (horiz ? itemDistance : labelPadding)
-        );
+            (horiz ? itemDistance : labelPadding) + (
+                (options.layout === 'vertical') &&
+                (item as ColorAxis).maxLabelLength &&
+                ((item as ColorAxis).maxLabelLength > itemExtraWidth) &&
+                ((item as ColorAxis).maxLabelLength - itemExtraWidth) ||
```
KacperMadej, line 597: Those castings don't look like needed - `item` has its type defined as the function argument.

KacperMadej (changes requested):
Looks like padding should be doubled for the legend border to not touch the color axis labels - http://jsfiddle.net/BlackLabel/3gwyejxs/1/

raf18seb (approved):
Nothing from me, good job!

TorsteinHonsi (changes requested):
I think we can simplify this a bit, as we don't need to handle `showCheckbox`. I think the following will suffice. It will make the legend smaller in the cases where the text labels are shorter, but that is to be expected.

```ts

        // Set how much space this legend item takes up
        axis.legendItemWidth = (
            width +
            padding +
            (
                horiz ?
                    itemDistance :
                    (legendOptions.symbolPadding || 0) +
                    this.maxLabelLength +
                    padding
            )
        );
```

ts/Core/Axis/Color/ColorAxis.ts
```diff
@@ -1001,6 +1010,21 @@ namespace ColorAxis {
 // Properties to preserve after destroy, for Axis.update (#5881, #6025).
 Array.prototype.push.apply(Axis.keepProps, ColorAxis.keepProps);
 
+addEvent(ColorAxis, 'afterGetOffset', function (): void { // (#15551)
```
TorsteinHonsi, line 1013: It should not be necessary with an event handler here, as we already have an override for the `getOffset` function on the ColorAxis object. It should work the same if we put this logic in the function override right after `super.getOffset()`

https://github.com/highcharts/highcharts/blob/e2238b6e0432bbbd26ca0de91de1ecd3ebedde04/ts/Core/Axis/Color/ColorAxis.ts#L500-L501

TorsteinHonsi (changes requested):

TorsteinHonsi (approved):

madepiet (approved):

KacperMadej (approved):

PR #16595, opened 2021-11-04
Created an icon for the editor page

nasin:
Hi,

I needed to create an icon/chart animation for the Editor page. Here it is.

All the JS is vanilla.


TorsteinHonsi (approved):
Thanks!

PR #16597, opened 2021-11-05
Bugfix #16549: Keyboard navigation in heatmap 

### LINKED ISSUE #16549: [Accessibility] Keyboard navigation does not work correctly on the Heatmap in the 9.3 version.
#### Expected behaviour
Tab key should navigate between the chart elements. Arrow keys should navigate between the points.

#### Actual behaviour
Tab key behaviour is correct, but the arrow keys do not navigate between points on the Heatmap.
The latest working version is 9.2.2.

Use the tab key to choose a chart to navigate and try to navigate between the points using arrow keys.

Demo:
https://jsfiddle.net/BlackLabel/kh50Lzc2/

#### Product version
Highcharts 9.3.0

#### Affected browser(s)
N/A

bre1470:
Fixed #16549, a regression causing arrow navigation in map-like series types not to work.

~~Fixed #16549; Arrow navigation in map-like series types.~~


oysteinmoseng (approved):
Nice, no comments from me!

TorsteinHonsi (approved):

PR #16602, opened 2021-11-05
Changed initial overlay to match new design

nasin:
Hi, 

The overlay we were using to prevent people from mousing over the diagram was out of style with the new design. I changed it. (Dimming instead of an overlay.)


TorsteinHonsi (approved):
Thanks!

PR #16606, opened 2021-11-07
Removed title and subtitle

nasin:
Hi, Marketing requested that I remove the title and subtitle to make the chart less clutered.


TorsteinHonsi (approved):
Thanks!

PR #16607, opened 2021-11-07
Shortened copy to prevent overflow

nasin:
Hi,

I shortened the copy for the data labels and titles so that it's less crowded on smaller screens.


TorsteinHonsi (approved):
Approved, pending final CI run

PR #16613, opened 2021-11-08
bugfix/11636-update-series-mapData

### LINKED ISSUE #11636: Change mapData and data dinamically
#### Expected behaviour
Change `mapData` and `data` with the `update` method of the series.

#### Actual behaviour
The `mapData` doesn't change unless `hasDerivedData` is set to `true`.

[Demo](https://jsfiddle.net/DiDream/64vdk30p/9)

#### Product version
Highmaps JS v7.1.2 (2019-06-03)

#### Affected browser(s)
All

raf18seb:
Fixed #11636, updating [series.mapData](https://api.highcharts.com/highmaps/series.map.mapData) didn't work correctly.


KacperMadej (changes requested):
In short: the test is not working as expected.

In addition to the comments in code and as a reference (and info for @DJTechnoo and @madepiet ): in samples/data/map/bn-all.js we have `Highcharts.maps["countries/bn/bn-all"] = ...` so that registers the map for the karma tests. The "redirect" to the map-collection (top 2) or old maps copy (in samples dir) is defined in [karma-files.json](https://github.com/highcharts/highcharts/blob/a04cfc3dd0e08e7cd7c0c8cb4e4b99b4953e1aff/test/karma-files.json#L72-L79). This means that those files are loaded/executed for the whole karma environment that is shared for all tests running in headless on pre-commit as well as in CI.

ts/Core/Series/Series.ts
```diff
@@ -4219,7 +4221,19 @@ class Series {
...
+            // to be set before series.setData(), #11636
+            if (mapData) {
+                hasMapDataChanged = series.options.mapData !== mapData;
+
+                if (hasMapDataChanged) {
+                    series.options.mapData = mapData;
+                }
+            }
+
+            if (options.data || hasMapDataChanged) {
```
TorsteinHonsi, line 4236: We want to keep non-generic stuff out of the core Series.ts file. However there's a mechanism where we do a test for whether options have changed in order to determine `keepPoints`. I tried setting `keepPoints` to false, and that works for the use case. However I was expecting that `series.hasOptionChanged('mapData')` would be true, but it's not. So I would start with finding out why.

An alternative possible fix is the set [hasDerivedData to true](https://jsfiddle.net/highcharts/7raL89ep/) on the MapSeries prototype, but I am not sure of the consequences. In a sense it is semantically true because the data is composite and derived from `mapData` and `data`, but then again this flag is used with series where the data derived from another series.

TorsteinHonsi (changes requested):

ts/Core/Series/Series.ts
raf18seb: At the beginning I thought about the `series.hasOptionChanged` method too, but I don't think it's doing what we think. It's not comparing old option to an updated (new) option (during series.update()), but it's checking whether the option has been changed directly on the series level.

For example:
```js
series.update({ name: 'Test' });
// inside the series.update method
series.hasOptionChanged('name'); // -> false
```
but
```js
series.name = 'Test";
series.hasOptionChanged('name'); // -> true
```

The unit tests suggest the same. So I'm leaving this method as it is.

----------------------------------------------------------------------------
I'm also trying to fix it without setting `hasDerivedData: true` on the MapSeries prototype. Commit will be pushed soon

TorsteinHonsi: Okay, so I misunderstood what it did.

TorsteinHonsi:
There's also a case when `joinBy` is null, and we're just setting numeric data. In that case, the data will be purged with the new approach: https://jsfiddle.net/highcharts/thc1yqdb/1/ . 

I am not sure how realistic it is, but it is at least a bug. It not be worthwhile fixing it here.

Otherwise the PR is fine (except resolution to the comments above).

raf18seb:
All comments resolved, new tests added - ready for a final review.

> There's also a case when joinBy is null, and we're just setting numeric data. In that case, the data will be purged with the new approach: https://jsfiddle.net/highcharts/thc1yqdb/1/ .
I am not sure how realistic it is, but it is at least a bug. It not be worthwhile fixing it here.

I'm working on another ticket with updating `joinBy` - I'll investigate and maybe fix it there.

samples/unit-tests/maps/globalmap/demo.js
```diff
@@ -13,29 +14,40 @@ QUnit.test('Set basemap on chart object', function (assert) {
...
+
+    chart.series[0].update({
+        mapData: Highcharts.maps['countries/bn-all.js']
     });
+
+    assert.strictEqual(
+        chart.series[0].mapData.length,
+        51,
+        'Map on the series object should overrule chart-wide setting.'
+    );
```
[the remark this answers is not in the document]
raf18seb: `'invalid-map'` wasn't added by me. Probably it was there to notify us that the `chart.map` should no longer being taken into consideration after setting `series.mapData`. I removed it 🤷

ts/Series/Map/MapSeries.ts
```diff
@@ -1006,6 +1007,8 @@ class MapSeries extends ScatterSeries {
             if (mapData.type === 'FeatureCollection') {
                 this.mapTitle = mapData.title;
                 mapData = H.geojson(mapData, this.type, this);
+            } else {
+                mapData = isArray(mapData) ? mapData : mapData.features;
```
TorsteinHonsi, line 1011: What is the use case here? As I understand it, we either pass a full GeoJSON structure, in case the type must be FeatureCollection, or we pass the results of Highcharts.geojson, in chich case the result is an array. I'm concerned we are introducing more sloppy data structures, which may backfire. 

Also note we are about to [throw TopoJON into the mix](https://github.com/highcharts/highcharts/pull/16533/files#diff-928c6c929e43a646c2dbb842adc8ae6c2012597d40928d41e74f6b085948e2d0R1007-R1010).

TorsteinHonsi (changes requested):

ts/Series/Map/MapSeries.ts
raf18seb: When updating `series.update({ mapdata: ... })`, the `mapData.type` is being cleaned https://github.com/highcharts/highcharts/blob/d324ceb1af2d64caecd1a6094b04c27c199e0b47/ts/Core/Series/Series.ts#L4133
so the `mapData.type === 'FeatureCollection'` condition if `false`

Not sure how to solve it without interference to the Highcharts core

TorsteinHonsi: Okay... I need to think about that. Without the type it is not valid GeoJSON, so we need to find some other way.

KacperMadej: A merge of the current options (in userOptions) and the new options in [here](https://github.com/highcharts/highcharts/pull/16613/files#diff-928c6c929e43a646c2dbb842adc8ae6c2012597d40928d41e74f6b085948e2d0R1291) instead of just setting the new options might resolve this issue.

ts/Series/Map/MapSeries.ts
```diff
@@ -1288,7 +1286,8 @@ addEvent(MapSeries, 'update', function (
 
     if (options.mapData) {
         // Save the current (new) option that will be used in series.setData
-        series.options.mapData = options.mapData;
+        series.options.mapData =
+            merge(series.userOptions.mapData, options.mapData);
```
KacperMadej, line 1290: In the case ([demo](http://jsfiddle.net/BlackLabel/q201cfgy/)) of `mapData` being updated to a new array the merge I have suggested fails because when the old option is `undefined` (because `map` is set on chart level) - this forces an object out of the new array and fails in `forEach`.

How about `merge(true, series.options, { mapData: options.mapData });` instead?

TorsteinHonsi (changes requested):
I think we're about to overcomplicate things now... I added this line in `Series.update`, then all tests including the OP case works. Then we can remove all the stuff in MapSeries.ts. Yes, it's a modification of the core, but it is minimal.

In addition, the `keepPoints` flag is meant to be true when we simply want to update existing points because they can easily transition from one state to another. But that is not the case when we are replacing the whole map. But the number of points and what they represent is likely to change.

```ts
        const keepPoints = !(
            // Indicators, histograms etc recalculate the data. It should be
            // possible to omit this.
            this.hasDerivedData ||
            // New type requires new point classes
            (newType && newType !== this.type) ||
            // New options affecting how the data points are built
            options.mapData || // <======= Added this
            // Changes to data grouping requires new points in new group
            series.hasOptionChanged('dataGrouping') ||
            series.hasOptionChanged('pointStart') ||
            series.hasOptionChanged('pointInterval') ||
            series.hasOptionChanged('pointIntervalUnit') ||
            series.hasOptionChanged('keys')
        );
```

raf18seb:
> I think we're about to overcomplicate things now... I added this line in `Series.update`, then all tests including the OP case works. Then we can remove all the stuff in MapSeries.ts. Yes, it's a modification of the core, but it is minimal.
> 
> In addition, the `keepPoints` flag is meant to be true when we simply want to update existing points because they can easily transition from one state to another. But that is not the case when we are replacing the whole map. But the number of points and what they represent is likely to change.
> 
> ```ts
>         const keepPoints = !(
>             // Indicators, histograms etc recalculate the data. It should be
>             // possible to omit this.
>             this.hasDerivedData ||
>             // New type requires new point classes
>             (newType && newType !== this.type) ||
>             // New options affecting how the data points are built
>             typeof options.pointStart !== 'undefined' ||
>             typeof options.pointInterval !== 'undefined' ||
>             typeof options.relativeXValue !== 'undefined' ||
>             options.mapData || // <======= Added this
>             // Changes to data grouping requires new points in new group
>             series.hasOptionChanged('dataGrouping') ||
>             series.hasOptionChanged('pointStart') ||
>             series.hasOptionChanged('pointInterval') ||
>             series.hasOptionChanged('pointIntervalUnit') ||
>             series.hasOptionChanged('keys')
>         );
> ```

If the `data` is empty, the old map is not being destroyed: http://jsfiddle.net/BlackLabel/5t6dkzfs/
I should have added one more test testing that.

Also, the animation is gone.

TorsteinHonsi:
> If the data is empty, the old map is not being destroyed: http://jsfiddle.net/BlackLabel/5t6dkzfs/

The problem is that when using `allAreas: true`, we are internally generating a `data` array matching the given `mapData`. Then when we update `mapData`, the old `data` is still there, with geometries and all. I am not sure how to deal with that. As you see, if you first load the map with Andorra data, then load new mapData, the problem doesn't occur. I think in order to resolve this we must create a cleaner separation between `Series.data` and `Series.mapData`. But maybe it goes too far for this PR.

> Also, the animation is gone.

That's to be expected. Animation is for cases where the points are transitioning into another position, color, layout etc. But in this case we are replacing the regions with new regions of another country, so an animated transition in most cases doesn't make sense.

DJTechnoo (approved):

madepiet (approved):

TorsteinHonsi (approved):

PR #16618, opened 2021-11-09
Fixed #16173, word wrap failed inside anchors

### LINKED ISSUE #16173: Overlapping of long labels with anchor tags in it and extra marginBottom issue
#### ISSUE
In version 6.2.0 :- When we have long labels having an anchor tag as we set useHTML as true, they are overlapping and they are not being wrapped. Also because of the autoRotationLimit ( that is 60 ), when they are being rotated, the text is clipped.
For the normal text, we set useHTML as false. So when we have a longer text, it is wrapped but we get extra margin bottom between the chart and the legend-item.

In the latest version that is 9.2.0, behaviour is very weird when we set useHTML as true. Please refer the fiddle links for more clarity on this.

#### Expectation
We would like to have a consistent behaviour in which we don't have to worry about the overlapping or extra margin bottom that is being talked about.

#### Live demo with steps to reproduce for version Highcharts-6.2.0
 https://jsfiddle.net/ujjwal03/2jkdxL45/27/

#### Live demo with steps to reproduce for version Highcharts-9.1.2
 https://jsfiddle.net/ujjwal03/2jkdxL45/29/

TorsteinHonsi:
Fixed #16173, a regression since v9.0.0 causing word wrap and ellipsis to fail with text directly inside anchors.

### DIFF
--- a/samples/unit-tests/svgrenderer/text/demo.js
+++ b/samples/unit-tests/svgrenderer/text/demo.js
@@ -141,6 +141,15 @@ QUnit.test('Text word wrap with markup', function (assert) {
             'The text node width should be less than 100'
         );
     }
+
+    text.attr({
+        text: '<a href="https://www.highcharts.com">The quick brown fox jumps over the lazy dog</a>'
+    });
+
+    assert.ok(
+        text.getBBox().width <= 100,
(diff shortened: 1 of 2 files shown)


KacperMadej (approved):
Works with the demo reported in the issue 👍

PR #16623, opened 2021-11-09
Created static chart versions

nasin:
Hi, 

The web team asked me to create stand-alone static versions of the final charts for the product pages. I put them in their own directories.

I also fixed a bunch of accessibility issues and linted the CSS.


samples/highcharts/website/static-chart/demo.details
```diff
@@ -0,0 +1,7 @@
+---
+ name: Highcharts Static Chart for Highcharts Product Page
+ authors:
+   - Nancy Dillon
+ js_wrap: b
+ skipTest: true
```
TorsteinHonsi, line 6: Unless there's a good reason, remove `skipTest: true` so that the static charts become part of the visual test routine.

TorsteinHonsi (changes requested):

nasin:
I updated the static versions demo.details files and removed the skipTest

oysteinmoseng:
A11y review ongoing

- [ ] Highcharts demo
- [ ] Stock demo
- [ ] Map demo
- [ ] Gantt demo

oysteinmoseng (changes requested):
Some additional points that I did not fix @nasin :

- Icon-chart does not seem to be animating?
- I assume with reduced motion we would only show the static charts, so probably don't need much logic in the demos for this?
- Would recommend all the animating charts (icon-...) to have `enableMouseTracking: false` or `pointer-events: none` or similar, and remove the a11y module, then we'll give a description in the page where it is used.

**static-chart:**
- Is there any way to make the tooltip not move when hovered over? Thinking of low vision magnifier users.
- The green tabular point and brown wedge point both have too low contrast against the background. Could we add a dark border, or would this mess with the design too much?

**static-stock:**
- Tempted to add sonification to this, but perhaps wait until after the upcoming sonification revamp.

**static-map:**
- The darker points, such as the British isles, have low contrast against the background. Could we fix this somehow, either by changing colors or by adding a thin lighter border?
- Should the map title be "Population density" and not "Zoom to point"? Does it need a visible title at all?
- The subtitle text is invisible, but also not sure if it makes sense?
- Hovering over the color axis makes the text disappear
- Somehow it seems when zooming all the way out, you are zoomed back in automatically

**static-gantt:**
- The selected state for points needs to be obvious, since the user can remove the selected point. Currently you can only tell points are selected if you added them yourself.
- Could we reduce the contrast (!) on the vertical grid lines a bit? Or add a background color to the data labels for readability? Or both for that matter..
- Recommend not showing the "Remove selected" button unless something is selected.
- If possible, could we make the tooltip stick when hovered over? `stickOnContact: true` for some reason made the background color change for me..

nasin:
Hi,

I fixed all of the issues except the iceberg chart tooltip. I don't see how it's moving when I mouse over it.

oysteinmoseng:
Sorry, I meant change, not move. When moving the mouse over the text it changes to different points.

nasin:
Hi, I think I resolved the issue with the tooltip on the icon-chart chart (the icebergs.)

oysteinmoseng:
Hm, I think the tooltip might have been better before the fix actually.. It now moves whenever the mouse is moving, which makes it even harder. `stickOnContact: true` makes it tricky to use as well, so perhaps just revert to how it was for now.

One more question: Is there a static version of the home page stock chart somewhere? And is the stock chart after the animation supposed to be different from the one that happens if you click "Stop Animation"?

TorsteinHonsi (changes requested):
The gantt chart tooltip now looks a bit bad. There are borders on top and left but not right and bottom.

nasin:
I fixed the tooltip issues.

The static homepage chart is called static-candlestick.

oysteinmoseng:
I will make some a11y tweaks to the candlestick

oysteinmoseng:
Did my changes to static-map somehow get removed @nasin?

nasin:
I guess it's possible when I fixed the zoom issue on the map @oysteinmoseng

oysteinmoseng (approved):
All good from my side now I think!

TorsteinHonsi (approved):

PR #16632, opened 2021-11-10
Fixed accessibility, created static version

nasin:
Hi,

I created a static version of the editor page chart for accessibility purposes, and fixed the accessibility on both the animated and static charts.


samples/highcharts/website/static-editor/demo.details
```diff
@@ -0,0 +1,7 @@
+---
+ name: Highcharts Icon for Highcharts Editor
+ authors:
+   - Nancy Dillon
+ js_wrap: b
+ skipTest: true
```
TorsteinHonsi, line 6: When `skiptTest: true`, these charts are not considered when doing visual regression tests. But if there's not a specific reason for leaving them out (like custom animation, timeouts etc.), all the static charts should be part of the regression tests.

TorsteinHonsi (changes requested):

nasin:
I removed the skipTest from demo.details

TorsteinHonsi (approved):

PR #16636, opened 2021-11-11
Fixed: #15782, export hamburger icon hiding behind mapNavButtons.

### LINKED ISSUE #15782: Hanburger exporting menu not visible.
Hello

Issuing HighCharts Map, if exporting menu is left align with map navigation enabled. the exporting button is not visible if there is no title.

A workaround is to move the navigation menu.

"mapNavigation": {
    "enabled": true,
    "buttons": {
      "zoomIn": {
        "y": 33
      },
      "zoomOut": {
        "y": 60
      }
    }

#### Product version
Highmaps 9.0.1

#### Affected browser(s)
Chrome at least.

DJTechnoo:
Fixed #15782, export hamburger icon hiding behind `mapNavButtons`.


ts/Maps/MapNavigation.ts
```diff
@@ -240,6 +241,25 @@ MapNavigation.prototype.update = function (
                 height: 2 * button.height
             });
 
+            if (button.element) {
+                const expOptions = chart.options.exporting,
+                    expBtn = (expOptions && expOptions.buttons && expOptions.buttons.contextButton);
+
+                // If the nav buttons are on the same side
+                if (expBtn && defined(buttonOptions.y) && chart.exportingGroup) {
+                    if (chart.options && chart.title &&
```
raf18seb, line 250: These 2 if statements can be merged into 1

ts/Maps/MapNavigation.ts
```diff
@@ -240,6 +241,25 @@ MapNavigation.prototype.update = function (
...
+                        chart.title.textStr === '' &&
```
raf18seb, line 251: What about one title or subtitle is enabled?

raf18seb (changes requested):
In this demo http://jsfiddle.net/BlackLabel/rLcyx5jw/ the buttons still overlap. When debugging, the `chart.exportingGroup` is `undefined` and the added code is not fired.

Also, see the comments below

ts/Maps/MapNavigation.ts
DJTechnoo: If either has text, the bug is gone, so we check if both are empty

KacperMadej: An edge case when this is not true - http://jsfiddle.net/BlackLabel/rLcyx5jw/1/

ts/Maps/MapNavigation.ts
```diff
@@ -246,18 +247,45 @@ MapNavigation.prototype.update = function (
...
+                            chart.title && chart.title.textStr === '' &&
+                            chart.subtitle && chart.subtitle.textStr === '' &&
+                            (
+                                (
+                                    buttonOptions.align === 'right' &&
+                                    !expBtn.align
+                                ) ||
+                                (expBtn.align === buttonOptions.align)
+                            )
+                        ) {
```
raf18seb, line 271: I think this is a place where we could think about simplifying this `if` statement.
What about putting some of these conditions to a variable and using only a variable (or a few variables) in the `if` statement?

raf18seb (changes requested):
The logic seems to be fine for me, however, the `if` statement is very long and the readability here (code cleanliness) could be improved, see comment below.

samples/unit-tests/maps/setdata/demo.js
```diff
@@ -867,4 +870,13 @@ QUnit.test('Map set data with updated data (#3894)', function (assert) {
...
     );
+
+    const mapNavY = chart.mapNavButtons[0].alignAttr.y,
+        exportIconY = chart.exportSVGElements[0].alignAttr.translateY;
+
+    assert.notEqual(
+        mapNavY,
+        exportIconY,
+        '#15782, mapNav should not overlap with export icon.'
+    );
```
raf18seb, line 881: Also, I'm wondering about how safe this test is. What about the buttons overlap, but there's only 1px difference in y? This test would not catch this.
What about testing something like this?
```
assert.ok(
    mapNavY > exportIconY + exportIconY's height,
    '#15782, mapNav should not overlap with export icon.'
);
```

ts/Maps/MapNavigation.ts
```diff
@@ -246,18 +247,45 @@ MapNavigation.prototype.update = function (
...
+                            // (#15782) Make way for hamburger icon.
+                            buttonOptions.y +=
+                                chart.exportingGroup.getBBox().height * 1.2;
```
raf18seb, line 274: Suggestion: I'm wondering whether `* 1.2` is a good approach here - it might lead to unexpected offsets depending on the button's height. What about simply adding a constant distance, e.g. 10? What do you (and others) think about it?

raf18seb (changes requested):
Few small comments

ts/Maps/MapNavigation.ts
KacperMadej: @DJTechnoo ☝️

ts/Maps/MapNavigation.ts
```diff
@@ -241,23 +242,55 @@ MapNavigation.prototype.update = function (
...
+                if ( // Has expBtn and empty title/subtitle
+                    expBtn && defined(buttonOptions.y) &&
+                    chart.exportingGroup && chart.options &&
+                    chart.title && chart.title.textStr === '' &&
+                    chart.subtitle && chart.subtitle.textStr === '' &&
+                    (
+                        ( // mapNav and expBtn on right side
+                            buttonOptions.align === 'right' &&
+                            !expBtn.align
+                        ) || // or left side
```
KacperMadej, line 263: A comment like "or the same align option" would be closer to the truth.

ts/Maps/MapNavigation.ts
```diff
@@ -241,23 +242,55 @@ MapNavigation.prototype.update = function (
...
             });
 
+            // (#15782) Make way for hamburger icon.
+            function adjustMapNavBtn(): void {
+                const expOptions = chart.options.exporting,
+                    expBtn = (
+                        expOptions &&
+                        expOptions.buttons &&
+                        expOptions.buttons.contextButton
+                    );
```
KacperMadej, line 252: Alternative shorter code version:
`const expBtn = ((chart.options.exporting || {}).buttons || {}).contextButton;`

KacperMadej (approved):
There are some edge cases but none of them seems like a huge problem - this is good enough IMO. Optional comments in code.

madepiet (approved):

raf18seb (approved):
Thanks!

TorsteinHonsi (changes requested):

raf18seb:
We have a new solution with checking BBoxes, but we still can improve checking the `y` collision and fix some issues with updating.

TorsteinHonsi:
It's still in draft... Should it be Ready for review?

DJTechnoo:
oops, yes it is ready now

ts/Maps/MapNavigation.ts
```diff
@@ -63,6 +64,7 @@ declare global {
 
         interface MapNavigationChart extends Chart {
             mapNavButtons: Array<SVGElement>;
+            navButtonsGroup: SVGElement;
```
TorsteinHonsi, line 67: Instead of adding a property to the already crowded `Chart` class, add this to the `MapNavigation` class. 

We should do that with `mapNavButtons` too, but that may be out of scope for this PR.

ts/Maps/MapNavigation.ts
```diff
@@ -252,13 +256,65 @@ MapNavigation.prototype.update = function (
...
+                    });
+                }
+            }
+        };
+
+        if (!chart.hasLoaded) {
+            // Align it after the plotBox is known (#12776)
+            // and after the hamburger button's position is known
+            // so they don't overlap (#15782)
+            addEvent(chart, 'render', (): void => {
```
TorsteinHonsi, line 314: The callback can be passed directly. `addEvent(chart, 'render', adjustMapNavBtn);`

TorsteinHonsi (changes requested):
Looking good! I made a small refactoring in order to reuse a copy the `isIntersectRect` function from other parts of the code base. That makes it clearer for us if we want to share that from a module later on.

raf18seb (approved):

madepiet (approved):

TorsteinHonsi:
All good now! But I couldn't help my self using this PR to refactor the `mapNavButtons` away from Chart.

Pending CI run before final approval.

TorsteinHonsi (approved):

PR #16639, opened 2021-11-11
Fixed CSS error causing color issues

nasin:
Hi, there was a CSS error causing the second chart to appear yellow.


nasin:
I fixed some timing issues and CSS.

TorsteinHonsi (approved):
Thanks!

PR #16642, opened 2021-11-12
Fixed #16562, Windows high contrast mode keyboard nav issues.

### LINKED ISSUE #16562: Toggling a line from the legend through the keyboard in high contrast mode does not work
#### Expected behaviour
In High contrast mode: It should be possible to un-select a line in a line chart by pressing space bar when the name is selected in the legend next to the chart.

#### Actual behaviour
The line is not hidden. Clicking on the label works.

Looking at the code: it seems the keydown event is register twice and leads to the line being hidden and immediately being shown again.

This happens in `KeyboardNavigation.prototype.init = function (chart, components)`  (accessibility.src.js).
Wher in `Accessibility.prototype.init = function (chart) {`
1) `this.keyboardNavigation = new KeyboardNavigation(chart, this.components);` registers a key down event **(number 1)**
2) the next line is `this.update();` 
3) which calls `setHighContrastTheme` 
4) which triggers an update of the series 
5) which updates of the chart and `updateA11yEnabled`  
7) which leads to the creation of a `new Accessibility(` 
8) which leads to the creation of a new `KeyboardNavigation` 
9) which registers a key down event **(number 2)**
10) it will not register another one, as step 3) `setHighContrastTheme` is not called again, as `highContrastModeActive` is true at this point.

https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/demo/line-basic
The issue is visible when I turn on the Windows high contrast mode and open the above jsfiddle in the browser
Steps:
- in high contrast mode
- navigate to "installation" in the legend on the right side of the graph
- press "space"
- "installation" line stays visible / is not deactivated
- click on the same label
- "installation" line disappears / is deactivated

#### Product version
v9.3.0 (2021-10-21)

#### Affected browser(s)
Edge Version 95.0.1020.30 (Official build) (64-bit)

oysteinmoseng:
Fixed #16562, Windows high contrast mode keyboard navigation issues.

### DIFF
--- a/ts/Accessibility/Accessibility.ts
+++ b/ts/Accessibility/Accessibility.ts
@@ -160,7 +160,6 @@ class Accessibility {
         this.keyboardNavigation = new (KeyboardNavigation as any)(
             chart, this.components
         );
-        this.update();
     }
 
 
@@ -454,6 +453,9 @@ namespace Accessibility {
                 a11y.update();
             } else {
                 this.accessibility = a11y = new (Accessibility as any)(this);
+                if (a11y) {
+                    a11y.update();
+                }
             }
         } else if (a11y) {
             // Destroy if after update we have a11y and it is disabled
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (approved):
Thanks!

PR #16646, opened 2021-11-15
bugfix/14267-tilemap-square-error

### LINKED ISSUE #14267: Tilemap - tileshape `square` doesn't work
#### Expected behaviour

It's a regression. The last properly working version is an 8.0.4, so the chart should work as it used to be. 
Demo: https://jsfiddle.net/BlackLabel/t9hnvu7g/

#### Actual behaviour

The points don't draw and the error occurs in the console.

Demo: https://jsfiddle.net/BlackLabel/f7dx9ysa/

#### Product version

 Highmaps JS v8.2.0 (2020-08-20)

raf18seb:
Fixed #14267, tilemap [tileShape](https://api.highcharts.com/highmaps/series.tilemap.tileShape) `square` was throwing an error into the console.

### DIFF
--- a/samples/unit-tests/series-tilemap/coloraxis/demo.js
+++ b/samples/unit-tests/series-tilemap/coloraxis/demo.js
@@ -35,6 +35,11 @@ QUnit.test('Tilemap and ColorAxis', function (assert) {
         'ColorAxis.max should be the same as max value in points (#11644)'
     );
 
+    // Square shape: no errors should be visible in the console, #14267
+    chart.series[0].update({
+        tileShape: 'square'
+    });
+
     chart.series[0].update({
         tileShape: 'circle'
     });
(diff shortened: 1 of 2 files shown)


DJTechnoo (commented):
Looks good to me!

madepiet (approved):

KacperMadej (approved):

TorsteinHonsi (changes requested):
Thanks! I think now that we have guarded `markerOptions` we should also take the opportunity and get rid of the `any` cast. When typing the function input, we can resolve the rest of the anys:

```ts
        [['width', 'x'], ['height', 'y']].forEach(function (
            dimension: ['width'|'height', 'x'|'y']
        ): void {
```

TorsteinHonsi (approved):
Thanks!

PR #16654, opened 2021-11-15
docs/maps-series-link

madepiet:
Fixed link demo.

### DIFF
--- a/docs/maps/map-series.md
+++ b/docs/maps/map-series.md
@@ -17,7 +17,7 @@ For an overview of the `map` series options see the [API reference](https://api
        ...
     });
 
-<iframe style="width: 100%; height: 520px; border: none;" src="https://www.highcharts.com/samples/embed/maps/demo/category-map" allow="fullscreen"></iframe>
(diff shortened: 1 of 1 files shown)


docs/maps/map-series.md
```diff
@@ -15,7 +17,7 @@ For an overview of the `map` series options see the [API reference](https://api
        ...
     });
 
-<iframe style="width: 100%; height: 520px; border: none;" src="https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/maps/demo/category-map/" allow="fullscreen"></iframe>
+<iframe style="width: 100%; height: 520px; border: none;" src="https://www.highcharts.com/samples/embed/maps/demo/category-map" allow="fullscreen"></iframe>
```
raf18seb, line 20: Why `www.` is added here but not in the demo above? Not sure which one is correct (probably both), but we can choose one and be consistent from now

raf18seb (changes requested):
Just a small question

KacperMadej (approved):
I agree with Rafał. Other than that looks good 👍

TorsteinHonsi (approved):
Thanks!

PR #16666, opened 2021-11-17
Website/homepage 2021

nasin:
Fixed a JS error.

### DIFF
--- a/samples/highcharts/website/homepage-2021/demo.js
+++ b/samples/highcharts/website/homepage-2021/demo.js
@@ -895,7 +895,8 @@ const sankey = {
                             from: sankeyData[ii][0],
                             to: sankeyData[ii][1],
                             weight: sankeyData[ii][2]
-                        });
+                        }, false);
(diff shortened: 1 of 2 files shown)


samples/highcharts/website/homepage-2021/demo.js
```diff
@@ -888,7 +888,15 @@ const sankey = {
                         } else {
                             clearInterval(sankey);
                         }
-                    }, 10);
+                    }, 10);*/
+
+                    for (let ii = 0; ii < sankeyData.length; ++ii) {
+                        chart.series[0].addPoint({
```
TorsteinHonsi, line 894: When in a loop, `addPoint` should be called with the `redraw` parameter set to false, and `chart.redraw()` be called after. Otherwise, the chart will waste browser resources on redrawing for every iteration.

TorsteinHonsi (changes requested):

TorsteinHonsi (approved):

PR #16668, opened 2021-11-17
Website/highcharts timeline

nasin:
Hi, I updated the revenue through 2020 (based on a spreadsheet from Anne Jorunn) and removed the employee names for GDPR reasons.


TorsteinHonsi (approved):
Thanks! I'm approving the chart because we need to go live, but I noticed a couple of things that ideally should be fixed:
- Collision of flags Award/First book.
- Legend crashing into chart in mobile view
- Turnover flatlining after 2019
- A strange issue when I mouse over and mouse out, the turnover series redraws

TorsteinHonsi:
PS: @nasin If and when you have fixed some or all of the comments, please create a new PR.

PR #16683, opened 2021-11-19
Update demo.css

nasin:
Removed white background


TorsteinHonsi:
What is the use case? Is this chart embedded in an iframe in our website?

The consequence of adding `body` rules to the demo is that in case it is embedded directly into the page, it will take over the styling of the whole page. Is it the intention that the demo should affect more than its own area? Otherwise the styles should be restricted to `.hero-container` or `.chart-container`.

TorsteinHonsi (approved):
Never mind the comment, I see that we're loading the `nonav` page directly inside our hero section. We'll deal with the body styling _if_ it becomes a problem in the future.

Issue #16667, opened 2021-11-17
Support Trusted Types


ecktom:
#### Expected behaviour
Enabling [Trusted Types](https://web.dev/trusted-types/) should work without issues

#### Actual behaviour
When [Trusted Types](https://web.dev/trusted-types/) is enabled on a website, following lines causes Trusted Types violation due to string assignment to a dangerous sink (`DOMParser().parseFromString`).

`This document requires ‘TrustedHTML’ assignment` error is thrown due to https://github.com/highcharts/highcharts/blob/master/ts/Core/Renderer/HTML/AST.ts#L44

This could be prevented by changing those lines to something like (untested):
```
if (window.trustedTypes) {
  let staticPolicy = window.trustedTypes.createPolicy("highcharts", { createHTML: (input) => input }, true);
  return Boolean(new DOMParser().parseFromString(staticPolicy.createHTML(''), 'text/html'));
} else {
  return Boolean(new DOMParser().parseFromString('', 'text/html'));
}
```

Note:
Any input other than `''` should be sanitized like `createHTML: (input) => DOMPurify.sanitize(input)`

TorsteinHonsi:
Thanks for reporting! I'll look into this tomorrow.

PR #16694, opened 2021-11-23
Added support for Trusted Types, #16667

TorsteinHonsi:
Added support for Trusted Types, a browser standard for preventing cross-site scripting vulnerabilities.

See #16667.

### DIFF
--- a/docs/chart-concepts/security.md
+++ b/docs/chart-concepts/security.md
@@ -13,6 +13,8 @@ Highcharts.AST.allowedAttributes.push('data-value');
 Highcharts.AST.allowedReferences.push('tel:');
 ```
 
+Since v9.4 Highcharts also supports [Trusted Types](https://web.dev/trusted-types/), and our Cypress tests are set up with Trusted Types enabled which enforces an added layer of security.
+
 ### Scope
 Chart _configuration options_ that are valid JSON are filtered for known XSS vectors. JSON support types of object, strings, numbers, arrays, boolean or null. This implies that function callbacks and event handlers in the chart options structure is not covered by our scope. Functions are inherently unsafe because they by nature allow scripting towards the web page.
 
--- a/docs/chart-concepts/security.md
+++ b/docs/chart-concepts/security.md
@@ -13,7 +13,7 @@ Highcharts.AST.allowedAttributes.push('data-value');
 Highcharts.AST.allowedReferences.push('tel:');
 ```
 
-Since v9.4 Highcharts also supports [Trusted Types](https://web.dev/trusted-types/), and our Cypress tests are set up with Trusted Types enabled which enforces an added layer of security.
+Since v9.4 Highcharts also supports [Trusted Types](https://web.dev/trusted-types/) under the built-in policy name `highcharts`. Our Cypress tests and developer tools are set up with Trusted Types enabled, which enforces an added layer of security and greatly reduces the risk for one of our developers to accidentally introduce a DOM XSS vulnerability.
 
 ### Scope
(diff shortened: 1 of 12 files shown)


TorsteinHonsi:
In its current state this prevents errors in the console. But it doesn't really add any value. Highcharts prevents XSS by using DOMParser to parse incoming strings, then operate on the results. In its current state, we are just adding a dummy function before DOMParser, but ideally the DOMParser should be part of the policy.

TorsteinHonsi:
Debugging
* Pull latest `highcharts-utils`
* Start it with `sudo node server --trustedTypes`

Discussion points:
* Do we need an option to enable this? [DOMPurify disables](https://github1s.com/cure53/DOMPurify/blob/main/src/purify.js#L248-L251) this feature by default, with an option to turn it on. Maybe we should do the same, as it is still not widely adopted by browsers. It could be a flag on the AST like `allowTags` etc.
* Do we need an optional policy name in case of multiple scripts loaded. [DOMPurify has one](https://github1s.com/cure53/DOMPurify/blob/main/src/purify.js#L39-L40). But Highcharts throws error 16 in the first place, so it should be redundant.

TorsteinHonsi:
@pawelfus Note for debugging: In the latest utils, I had to comment out the meta tag because Cypress broke with all the other branches. So in order to test it now, you need to uncomment this tag:

https://github.com/highcharts/highcharts-utils/blob/d50d62343fa01a1d6f2226945782352ea9525b0e/views/layout.hbs#L17-L20

docs/chart-concepts/security.md
```diff
@@ -13,6 +13,8 @@ Highcharts.AST.allowedAttributes.push('data-value');
 Highcharts.AST.allowedReferences.push('tel:');
 ```
 
+Since v9.4 Highcharts also supports [Trusted Types](https://web.dev/trusted-types/), and our Cypress tests are set up with Trusted Types enabled which enforces an added layer of security.
+
```
pawelfus, line 17: Q: Should we inform users about the policy name?

ts/Core/Renderer/HTML/AST.ts
```diff
@@ -20,28 +20,59 @@ import type HTMLAttributes from '../HTML/HTMLAttributes';
...
+        'highcharts', {
+            createHTML: (s: string): string => s
+        }
+    )
+);
+
+// Create a TrustedHTML object from an untrusted string. This should also not be
+// exposed, and only used in our controlled cases (empty string and prior to
+// running the DOMparser).
+const trustedHTML = function (html: string): string {
```
pawelfus, line 61: The only problem I have here it's about the `trustedHTML` name. It might return an unsafe string. I guess that's the moment where DOMParser comes in handy?

pawelfus (approved):
Looks good 👍 

Q: Should we also include info in TT's wiki? This one: https://github.com/w3c/webappsec-trusted-types/wiki/Integrations

ts/Core/Renderer/HTML/AST.ts
TorsteinHonsi: > we would need to check in other places

True, but only in two places - when creating the empty HTML, and before running `parseFromString`. I refactored this, got rid of the `trustedHTML` function.

pawelfus: Nice, thanks!

Techn1x:
I'd really like to see this disabled, or configurable, until this issue is fixed
(on `window.print()` in chrome, if the page contains a chart it irrecoverably breaks the page, due to trusted types)
https://github.com/highcharts/highcharts/issues/16931 (contains a repro jsfiddle at the end)
https://github.com/highcharts/highcharts/issues/17533
https://github.com/cure53/DOMPurify/issues/529

PR #16696, opened 2021-11-23
Tools: Doclet readme

bre1470:


karolkolodziej (changes requested):
Maybe it is worth mentioning how the `@productdesc` works and when it should be used?

karolkolodziej (approved):

raf18seb (approved):
Thanks! ;)

TorsteinHonsi (approved):
Thanks!

PR #16702, opened 2021-11-25
Fixed issue with data module updates with firstRowAsNames enabled.

oysteinmoseng:
Fixed issue with data module updates with `firstRowAsNames` enabled.
___
Pragmatic fix for this issue, which happens both with using `data.columns` and any other options that parse something and then set `data.columns`, such as the Google Sheet link.

The `data.complete()` function modifies the `data.columns` prop when removing the first row as names. This means that when the chart data options are updated (`data.update` runs), the `data.columns` prop already has the first row removed. Since `firstRowAsNames` is still true, the first row is removed again, and data rows are lost.

Pragmatic fix: Work on a copy of the columns array in `data.complete()` when using firstRowAsNames, and avoid modifying `data.columns`.

Not a huge fan of this fix for two reasons:
1. It slightly breaks backwards compatibility, because `data.columns` will now also include the column headers, and will no longer have the `.name` prop. We can of course re-add the `.name` prop purely for compatibility, but it will still include the column headers so chances are a rewrite will be needed by the users that modify this array in some cases.
2. We are using custom props on the column arrays that have to be copied over, and if we add more of these and forget to do this we will introduce bugs (`column.isNumeric`, `column.unsorted`, etc). It seems overkill to write logic to copy over all non-array elements, but that is a solution of course.

Suggestions for a better fix appreciated, but it seems to work at least.

### DIFF
--- a/samples/unit-tests/data/general/demo.js
+++ b/samples/unit-tests/data/general/demo.js
@@ -296,7 +296,7 @@ QUnit.test(
         });
 
         assert.strictEqual(
-            chart.data.columns[2][1],
+            chart.data.columns[2][0],
             null,
             'Empty point should be parsed to null instead of undefined.'
         );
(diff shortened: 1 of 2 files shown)


TorsteinHonsi:
What if we don't create copies, but instead makes sure the `name` is only shifted once? This passes all the tests:

```ts
            // Get the names and shift the top row
            if (this.firstRowAsNames) {
                for (i = 0; i < (columns as any).length; i++) {
                    if (!defined((columns as any)[i].name)) {
                        (columns as any)[i].name = pick(
                            (columns as any)[i].shift(),
                            ''
                        ).toString();
                    }
                }
            }
```

oysteinmoseng:
Yes much nicer. I thought this would run into issues with the user not being able to update the first row names, but in practice it works fine since in that case they are passing in new arrays without the `.name` prop anyway. I'll update!

TorsteinHonsi (approved):

Issue #5050, opened 2016-02-16
series.setData causes SVG error in highmaps


ranair:
If map is created and data added immediately using setData, It causes bunch of errors in console log.
Sample error:

 *Error: Invalid value for <path> attribute d="M 13.083870967741936 270.0354838709677 L 14.090322580645163 270.17926267281103 14.30599078341014 270.5387096774193 13.299539170506913 270.39493087557605 12.508755760368665 269.89170506912444 12.07741935483871 269.9635944700461 11.502304147465438 269.38847926267283 12.364976958525347 269.2447004608295 13.083870967741936 270.0354838709677 Z 13.083870967741936 270.0354838709677 Z 13.083870967741936 270.0354838709677 Z"*

This seem to happen in chorme, not in firefox.
Here is the code snippet.
Steps to reproduce:
Go to world with Palestine highres: map collection form Highmaps doc.
http://jsfiddle.net/gh/get/jquery/1.11.0/highslide-software/highcharts.com/tree/master/samples/mapdata/custom/world-palestine-highres

Replace the java script the one below.
Run and look at the JS console in chrome:
```js
$(function () {
  // Initiate the chart
    $('#container').highcharts('Map', {

        title : {
            text : 'Highmaps basic demo'
        },

        mapNavigation: {
            enabled: true,
            buttonOptions: {
                verticalAlign: 'bottom'
            }
        },

        colorAxis: {
            min: 0
        },

        series : [{
            data : null,
            mapData: Highcharts.maps['custom/world-palestine-highres'],
            joinBy: 'hc-key',
            name: 'Random data',
            states: {
                hover: {
                    color: '#BADA55'
                }
            },
            dataLabels: {
                enabled: true,
                format: '{point.name}'
            }
        }]
    });
});
$(function () {
    var data = [
        {
            "hc-key": "us",
            "value": 1000
        },
        {
            "hc-key": "in",
            "value": 10000
        },
    ];

    // Initiate the chart
    $('#container').highcharts().series[0].setData(data);
});
```

jon-a-nygaard:
Thanks for reporting. We will have a look at this as soon as possible.

pawelfus:
Looks like connected to #3440

jon-a-nygaard:
@TorsteinHonsi and @oysteinmoseng remember to check if the issue that Pawel referenced can be closed by this fix :wink:

oysteinmoseng:
@jon-a-nygaard: Unfortunately it seems to be unrelated

PR #16710, opened 2021-11-29
bugfix/15374-map-joinby-update

### LINKED ISSUE #15374: mapChart doesn't update colored areas after changing joinBy
#### Expected behaviour
I have a default joinBy parameter for map instantiation. For example: `joinBy: ['iso-a3', 'code']`. After retrieving new data I'm checking for actual country iso standard and update chart accordingly.
`chart.series[0].update({
  data: data,
  joinBy: ['iso-a2', 'code']
});`
I expect chart to bind country codes to a map correctly

#### Actual behaviour
Chart doesn't respect `joinBy` param in `update` method.

Actually I achieved updating of map by applying joinBy explicilty before calling for update method. Like this:
chart.series[0].joinBy = ['iso-a2', 'code'];
chart.series[0].update({data: data});

But this workaround doesn't work on jsfiddle

http://jsfiddle.net/Lu4ybx2p/
Press "update series"

#### Product version
I'm using highcharts 8.2.2. (Angular)
JSFiddle has version 9.0.1

#### Affected browser(s)
Google chrome Version 89.0.4389.82 (Official Build) (64-bit)

raf18seb:
Fixed #15374, maps `series.joinBy` update didn't work.

- [x] I have one `as any` in code. Any suggestion on how to fix it is more than welcome.


DJTechnoo (approved):

ts/Series/Map/MapSeries.ts
```diff
@@ -1274,6 +1275,46 @@ SeriesRegistry.registerSeriesType('map', MapSeries);
...
+                cutFrom = i;
+                return true;
+            } else {
+                return false;
+            }
+        });
+
+	    if (cutFrom) {
+            series.userOptions.data.length = cutFrom;
+            (eventOptions.options as any).data = series.userOptions.data;
```
KacperMadej, line 1304: This makes TS happy:
```ts
addEvent(MapSeries, 'update', function (
    eventOptions: { options: DeepPartial<MapSeriesOptions> }
): void {
    const series = this,
        data = eventOptions.options.data;
    let joinBy = eventOptions.options.joinBy;

    if (typeof data === 'undefined' && joinBy && series.userOptions.data) {
        // Cut the userOptions.data to get original data so the mapData is
        // re-calculated properly when updating joinBy, #15374
        let cutFrom;
        series.userOptions.data.some((point: any, i): boolean => {
            if (point._i !== void 0) {
                cutFrom = i;
                return true;
            } else {
                return false;
            }
        });

        if (cutFrom) {
            series.userOptions.data.length = cutFrom;
            (eventOptions.options as any).data = series.userOptions.data;
        }

        // Handle joinBy that might be a string
        const seriesJoinBy = splat(joinBy);
        if (!seriesJoinBy[1]) {
            seriesJoinBy[1] = seriesJoinBy[0];
        }

        // Save the current (new) option that will be used in series.setData
        series.joinBy = seriesJoinBy;
    }
});
```
Also, in your code, you have trailing spaces (1305) and a tab instead of spaces (1302) - fixed in the code above.

samples/unit-tests/maps/series-map/demo.js
```diff
@@ -8,24 +8,61 @@ QUnit.test('Mapping of joinBy with custom sub options', function (assert) {
...
+
+        series.update({ // #15374
+            joinBy: joinBy
+        });
+
+        assert.strictEqual(
+            series.points[i].color,
+            Highcharts.seriesTypes.map.defaultOptions.nullColor,
+            'The joinBy changed - previous territory should not be colored.'
+        );
```
KacperMadej, line 59: Add something to the asserts messages based on the `forEach` loop so in case of a failure it will be clear what exactly failed. Even using `i` should be enough.

KacperMadej (commented):
Looks good 👍 
Two suggestions are below.

ts/Series/Map/MapSeries.ts
raf18seb: Thanks!!!!

TorsteinHonsi:
Again I feel that we're creating a patch for something that should work out of the box. We should be able to add `options.joinBy` to the test for `keepPoints` (like we did with `mapData`), then all points should be recreated and everything should be sorted out by itself.

The root problem this time is that we're mutating `series.options.data`. Instead, we should leave that untouched and add the processed points to `series.data` or `series.points`, which would then be wiped out after `update`. That means we would need to do the `allAreas` block (and the culprit, [data.push](https://github.com/highcharts/highcharts/blob/2b7707c3b6cd7c875db22e5677ffd33de6353627/ts/Series/Map/MapSeries.ts#L1085)) later, perhaps in `generatePoints`. I'll dig more into it tomorrow.

TorsteinHonsi (changes requested):
I now refactored the `setData` and `processData` functions so that `data` and `mapData` are no longer mutated. Instead, `processData` produces an array, `processedData`, that is being picked up in `Series.generatePoints` instead of reading directly from `options.data`.

### To do
 - [x] Fix `maps/members/series-addpoint`. It fails now because `addPoint` usually splices added points into `options.data`, in the same position as in `Series.data`. But in `processedData`, the position is different, so the order of points versus data options is not correct in [generatePoints](https://github.com/highcharts/highcharts/blob/8422657f2d78523543b5644dd3020fcfaff4ea0c/ts/Core/Series/Series.ts#L1820-L1825).
 - [x] The fix for #5050 had to be removed, check if it is still relevant.

TorsteinHonsi (approved):
Sorry I took over the PR for a general refactoring task, but it was easier as the required tests were already set up. I guess this requires another review from @KacperMadej, @DJTechnoo as well as @raf18seb .

raf18seb:
I can't approve my own PR, but it makes sense and looks OK now 👍

DJTechnoo (approved):

madepiet (approved):

PR #16713, opened 2021-11-29
Blog Logarithmic and linear graph

mekhatria:
Three new demos for the upcoming blog: **When should you use logarithmic or linear scales in graphs?**


samples/highcharts/blog/simple-Logarithmic-graph/demo.js
```diff
@@ -0,0 +1,56 @@
+Highcharts.chart("container", {
```
TorsteinHonsi, line 1: This file doesn't lint because it has indenting of 2 spaces. It should be blocked for committing. Did you run `npm i` before commiting?

samples/highcharts/blog/simple-Logarithmic-graph/demo.css
```diff
@@ -0,0 +1,5 @@
+#container {
```
TorsteinHonsi, line 1: Only lower case letters should be used in samples folder names. In order to fix it you probably need to change the name to something temporary, then commit, then change it back to the final form and commit again.

samples/highcharts/blog/simple-Logarithmic-graph/demo.js
```diff
@@ -0,0 +1,56 @@
...
+    ]
+  },
+
+  yAxis: {
+    title: {
+      text: "Price"
+    },
+    type: "logarithmic",
+    labels: {
+      format: "{text}$"
```
TorsteinHonsi, line 24: I _think_ the [dollar sign should be before the number](https://english.stackexchange.com/questions/11326/what-is-the-difference-between-20-and-20) 🤔  . If so, that should be fixed throughout all these samples.

TorsteinHonsi (changes requested):
Thanks for the PR!

TorsteinHonsi (approved):
Thanks!

PR #16714, opened 2021-11-29
Fixed #16697, keyboard nav issue with legend styling.

### LINKED ISSUE #16697: Legend font color changes when using the keyboard to focus it
#### Expected behaviour
The legend font should not change color, OR at the very least the color should revert back to its original color when focus is lost .

#### Actual behaviour
Legend color changes when focused and does not revert back unless a mouse out event gets fired on the label.

https://jsfiddle.net/smayx41h/8/

#### Product version
9.3.1

#### Affected browser(s)
Google Chrome, Edge

oysteinmoseng:
Fixed #16697, legend item style changed when using the keyboard to focus it

### DIFF
--- a/ts/Accessibility/Components/LegendComponent.ts
+++ b/ts/Accessibility/Components/LegendComponent.ts
@@ -120,6 +120,24 @@ function shouldDoLegendA11y(chart: Chart): boolean {
 }
 
 
+/**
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (approved):
Thanks!

PR #16717, opened 2021-11-30
Fixed #16307, export menu button missing tooltip with a11y module.

### LINKED ISSUE #16307: Export menu button does not have a tooltip when accessibility module is attached
#### Expected behaviour
After hover on the export menu button should show a native tooltip with text: `Chart context menu` even when the accessibility module was attached.

#### Actual behaviour
When a module with accessibility is attached, the button has no native tooltip after hover on it. 

https://jsfiddle.net/BlackLabel/9px0gsan/

Workaround: https://jsfiddle.net/BlackLabel/orwcmxpu/

#### Product version
v9.2.2

oysteinmoseng:


TorsteinHonsi (approved):
Thanks!

TorsteinHonsi:
Pending CI

PR #16719, opened 2021-11-30
bugfix/16596-drag-column-category

### LINKED ISSUE #16596: Draggable variwide chart - First element of chart is not draggable on Y axis
#### Expected behaviour
Every datapoint of the variwide chart should be draggable on the y axis when the module is enabled.

#### Actual behaviour
The first datapoint is not draggable at any circumstances. Even if there is only one element in the chart, it is not draggable as it is the first datapoint.

https://jsfiddle.net/Gistiv/k2xvhyo0/96/

#### Product version
Found on 8.2.2 and is still present in 9.3.0 as you can see in the fiddle

#### Affected browser(s)
Reproduced with chrome, edge and firefox

magdalena-gut:
Fixed #16596, the first column point of categorized x-axis wasn't draggable.


raf18seb (approved):
Good job, thanks!

TorsteinHonsi (approved):
Thanks!

PR #16721, opened 2021-11-30
bugfix/10525-axis-cleared-undefined

### LINKED ISSUE #10525: Axis options can not be cleared with `undefined`
> If I set tickPositions to null, it magically works, but TS doesn't like it, null is not allowed.
> Setting single tickPositions to `undefined` is working with Highcharts 9.3.1.

See: https://github.com/highcharts/highcharts-react/issues/110

Demo: https://jsfiddle.net/43w5ch2L/

Workaround: https://jsfiddle.net/bre1470/tr6shfy4/
Setting tickPositions options to `[]` is working with rendering of Highcharts 9.3.1, though the previous options are still accessible via `Axis.options`.

mfilipiec:
Fixed #10525, setting [tickPositions](https://api.highcharts.com/highcharts/xAxis.tickPosition) to `undefined` didn't clear options.

### DIFF
--- a/samples/unit-tests/axis/ticks/demo.js
+++ b/samples/unit-tests/axis/ticks/demo.js
@@ -507,6 +507,17 @@ QUnit.test('The tickPositions option', function (assert) {
         'No extra ticks should be inserted when zooming between explicit ' +
             'ticks (#7463)'
     );
+
+    chart.xAxis[0].update({
+        tickPositions: undefined
+    });
+    
+    assert.deepEqual(
+        chart.xAxis[0].tickPositions,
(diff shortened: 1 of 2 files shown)


raf18seb (changes requested):
Thanks for the PR!
Few improvement suggestions below:

- [ ] can you reuse an existing test and update the chart instead of creating a new one from scratch? With thousands of tests, it decreases the time needed to all tests pass
- [x] you can use ES6 when writing tests

raf18seb (approved):
Good job, thanks!

TorsteinHonsi (approved):

PR #16723, opened 2021-11-30
bugfix/16670-indicator-series-update

### LINKED ISSUE #16670: Stock chart with GUI (indicators) cannot compatible with rangeSelector datagrouping
I am newbie in using HighStocks and Javascript. I am not sure whether it is actually a bug. What I actually do is copy the example of "Stock chart with GUI" and add a rangeSelector datagrouping function for grouping daily data into weekly and monthly data.

#### Expected behaviour

The indicators will be recalculated if the data type is changed. For example, if the data is on daily basis, the indicator should use the daily data for calculation. After the data is changed into monthly basis, the indicator should be recalculated based on monthly data.

#### Actual behaviour
Some indicators, such as ema, can function as expected.
But some indicators, such as sma, bb, cannot function well. For example, SMA on daily data is used. If I change to weekly data by clicking "week" button, all curves are missing. Only the sma is shown if I click the "week" button again. The original price data is missing.

https://jsfiddle.net/xmtjqLvw/

#### Product version

Highcharts Stock v9.3.1 (should be)

#### Affected browser(s)
Chrome 94

pawelfus:
Fixed #16670, Range Selector's buttons with `dataGrouping` setting caused error after adding an indicator using the Stock Tools GUI.

### DIFF
--- a/ts/Stock/Indicators/SMA/SMAIndicator.ts
+++ b/ts/Stock/Indicators/SMA/SMAIndicator.ts
@@ -379,15 +379,12 @@ class SMAIndicator extends LineSeries {
         let indicator = this,
             oldData = indicator.points || [],
             oldDataLength = (indicator.xData || []).length,
-            processedData: IndicatorValuesObject<typeof LineSeries.prototype> = (
-                indicator.getValues(
-                    indicator.linkedParent,
-                    indicator.options.params as any
-                ) || {
-                    values: [],
-                    xData: [],
-                    yData: []
-                }),
+            emptySet: IndicatorValuesObject<typeof LineSeries.prototype> = {
+                values: [],
+                xData: [],
+                yData: []
+            },
+            processedData: IndicatorValuesObject<typeof LineSeries.prototype>,
             croppedDataValues = [],
             overwriteData = true,
             oldFirstPointIndex,
@@ -396,6 +393,19 @@ class SMAIndicator extends LineSeries {
             min,
             max,
             i;
+
+        // Updating an indicator with redraw=false may destroy data.
+        // If there will be a following update for the parent series,
+        // we will try to access Series object without any properties
+        // (except for prototyped ones). This is what happens
+        // for example when using Axis.setDataGrouping(). See #16670
+        processedData = indicator.linkedParent.options ?
+            (
+                indicator.getValues(
+                    indicator.linkedParent,
+                    indicator.options.params as any
+                ) || emptySet
+            ) : emptySet;
 
         // We need to update points to reflect changes in all,
         // x and y's, values. However, do it only for non-grouped
--- a/samples/unit-tests/indicator-sma/recalculations/demo.js
+++ b/samples/unit-tests/indicator-sma/recalculations/demo.js
@@ -175,7 +175,10 @@ QUnit.test('Test algorithm on data updates.', function (assert) {
         false
     );
 
-    assert.ok('#16670: Update without redraw should not throw errors.');
+    assert.ok(
+        true,
+        '#16670: Update without redraw should not throw errors.'
+    );
 
     chart.redraw();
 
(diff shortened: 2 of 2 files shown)


ts/Stock/Indicators/SMA/SMAIndicator.ts
```diff
@@ -397,6 +394,19 @@ class SMAIndicator extends LineSeries {
             max,
             i;
 
+        // Updating an indicator with redraw=false may destroy data.
+        // If there will be a following update for the parent series,
+        // we will try to access Series object without any properties
+        // (except for prototyped ones). This is what happens
+        // for example when using Axis.setDataGrouping(). See #16670
+        processedData = indicator.linkedParent.options ?
```
pawelfus, line 402: I'm curious if there's a better way to check whether the series is being removed/updated. I decided to use `options`, because when there are `options` there must be a series too.

TorsteinHonsi: > I'm curious if there's a better way to check whether the series is being removed/updated

Usually we use `series.isDirtyData` for that purpose. Did you try that? Or maybe it doesn't cover this situation?

pawelfus: Thanks! Let me check..
Checked `series.isDirtyData`: it seems it can be undefined (and be valid), so we can accidentally skip processing the data, not good. Double-checked alternatives (according to TS and `series.init`): `chart`, `userOptions`, `data`, `eventsToUnbind`, `xAxis`, `yAxis`, `index`, `linkedSeries`, `tooltipOptions`, `points` (this one is definitively incorrect, DG can nullify this property), `zones` and some others.

I think, we used to have a property like `isRemoving`, but I don't see it anymore.

TorsteinHonsi: Have you considered the possibility that we may be forgetting to set `isDirtyData` in places where it should be set? The purpose of this flag is to set it whenever we are doing create, update or delete operations on the series data. So if we have a scenario where we modify the data and `isDirtyData` is not `true`, it should be considered a bug.

pawelfus: I have not. Good point, let me check that :)

pawelfus: Okay, here is the process:
1. Click the RangeSelector's button which has `dataGrouping` set
2. Delegate to `Axis.setDataGrouping`, which updates each series with new `dataGrouping` config:
    - update indicator series with new DG
        - indicator is removed, the connection with `linkedParent` is broken (which is correct, `update()` might change the parent etc.)
        - `afterLinkSeries` event is fired, but indicator was not initialized, so far so good
        - indicator is added back with new options, awaits `afterLinkSeries` event, to connect to the parent
    - update Navigator with new DG
        - `series.update()` calls `series.remove()` which calls `chart.linkSeries()`
        - indicator's parent series is still available, so linking succeeds
    - update the main series with new DG
        - `series.update()` calls `series.remove()` which calls `chart.linkSeries()`
        - because `linkSeries()` is inside `series.remove()`, the series is not completely removed. It's just an object with methods and properties from Prototype. Indicator still holds the reference to this series. Tries to access `series.something[0]` and the errors occures.

I think, there's no spot where `isDirtyData` should be set.

Extra:
In fact, I just realized that `linkSeries` clears `series.linkedSeries` array, but does not clear `series.linkedParent`. Just tried that solution [demo](https://jsfiddle.net/BlackLabel/pdyx0935/) - now it throws our error, that parent series was not found. This is correct because the parent series was removed.

TorsteinHonsi: Ok, thanks for the explanation!

TorsteinHonsi (approved):

PR #16724, opened 2021-11-30
bugfix/16492-getbaseseriesmin-empty-xdata

### LINKED ISSUE #16492: Highstock: series.addPoint() bug when there is no data in a second series
#### Behavior

If in the second series is no data and property `showInNavigator` is setting to `true` the animation on series.addPoint() doesn't work.

https://jsfiddle.net/BlackLabel/u4cz1woa/

#### Product version
9.0.0+

pawelfus:
Fixed #16492, Navigator did not stick to the max when adding a point and an additional, empty series was existing on the chart.


TorsteinHonsi (approved):
Thanks!

Issue #16728, opened 2021-12-01
Correct a usa map demo, remove unnecessary dataLabels placement code


raf18seb:
Continuation of https://github.com/highcharts/map-collection-dist/issues/89, related PR https://github.com/highcharts/map-collection/pull/15

Remove the unnecessary code responsible for custom dataLabels placement to fit them into a legend.

PR #16730, opened 2021-12-01
enhancement/16728-usa-map-datalabels-legend

raf18seb:
Done https://github.com/highcharts/highcharts/issues/16728, removed unnecessary code that fits the dataLabels to the legend in a USA map.
The dataLabels positions have been corrected directly in the source map in this PR https://github.com/highcharts/map-collection/pull/15.

**Merge only when a new map collection version is released.** FYI @TorsteinHonsi (keeping as a draft now)

### DIFF
--- a/samples/maps/demo/us-data-labels/demo.js
+++ b/samples/maps/demo/us-data-labels/demo.js
@@ -1,5 +1,5 @@
 // Load the data from the HTML table and tag it with an upper case name used for joining
-var data = [],
+const data = [],
     // Get the map data
     mapData = Highcharts.geojson(Highcharts.maps['countries/us/custom/us-small']);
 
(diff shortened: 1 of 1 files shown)


KacperMadej (commented):
It would be nice to take the opportunity and update the data in the demo to something more up-to-date than 2017 statistics.

TorsteinHonsi:
Thanks! For the records, stuff like this may not survive into the near future, because we are moving away from the serverside projected GeoJSONs. In this particular case for example, it may be that we should create a small client-side module that pulls out small areas and render data labels as squares on the side.

DJTechnoo (approved):

KacperMadej:
The expected result was achieved in another commit/PR and this PR is not needed - closing.

Issue #16634, opened 2021-11-11
Invisible series is still focusable


pvsleeper:
#### Expected behaviour
When a series is not visible, it should not be focusable, otherwise it gives the user the experience that focus is lost.

#### Actual behaviour
When a11y is enabled and the you click on the legend to remove the series, the removed series is still focusable.

#### Live demo with steps to reproduce
https://jsfiddle.net/wxch4rbj/4/

#### Product version
9.3.1

#### Affected browser(s)
Google Chrome, Edge

pvsleeper:
Hmmm... it seems like this may have been done on purpose. Guessing this is so that the screen reader has something to focus on to read out the chart title etc, even if there is no data.

KacperMadej:
Hi @pvsleeper 

Thank you for reporting the issue.

oysteinmoseng:
Can confirm that this is a bug and not intended behavior, we only want interactive elements to receive kbd focus.

oysteinmoseng:
Fixed in pending PR #16732. The fix should be part of the next official release.

Closing this issue manually as our automation won't catch it when multiple issues are fixed in one PR.

Issue #16633, opened 2021-11-11
Legend focus indicator remains after losing focus


pvsleeper:
#### Expected behaviour
Focus indicator should disappear when focus is lost from legend when using the mouse and a11y is enabled.

#### Actual behaviour
The issue is slightly intermitted, but happens 9/10 times. The blue focus indicator border does not get removed from the legend when using the mouse to set focus on another element in the DOM. Using the keyboard however appears to be working fine.  

#### Live demo with steps to reproduce
https://jsfiddle.net/Ld09htsz/13/

#### Product version
9.3.1

#### Affected browser(s)
Google Chrome, Edge

KacperMadej:
hi @pvsleeper 

Thank you for reporting the problem.

Not sure if the behavior is by design like that.

More specific a11y comments might follow soon.

pvsleeper:
Hi @KacperMadej - thanks for looking into it. I feel pretty sure that the behavior is incorrect as it is a false focus that is being displayed which goes against the a11y rules

oysteinmoseng:
Fixed in pending PR #16732. The fix should be part of the next official release.

Closing this issue manually as our automation won't catch it when multiple issues are fixed in one PR.

Issue #16625, opened 2021-11-10
Accessibility keyboard navigation not allowing wrap around from first point to last point


XiLiuRoy:
#### Expected behaviour
When setting KeyboardNavigation->wrapAround to be true, and focus is on first data point, hit "left arrow" should move focus to the last data point of current series as per previous version, e.g. 9.1.2

#### Actual behaviour
When setting KeyboardNavigation->wrapAround to be true, and focus is on first data point, hit "left arrow" and focus stays on the first data point.

#### Live demo with steps to reproduce
https://jsfiddle.net/eo2vk4u5/7/ -> version 9.3.1

1. Hit "Tab" to focus on Pie chart
2. Hit "Left Arrow"
3. Focus is not moved

and comparing to previous version 9.1.2 https://jsfiddle.net/eo2vk4u5/4/
1. Hit "Tab" to focus on Pie chart
2. Hit "Left Arrow"
3. Focus moves to last data point of current series

#### Product version
 version 9.3.1

#### Affected browser(s)
Chrome, Firefox, IE

KacperMadej:
Hi @XiLiuRoy ,

Thank you for reporting the problem.

oysteinmoseng:
Fixed in pending PR #16732. The fix should be part of the next official release.

Closing this issue manually as our automation won't catch it when multiple issues are fixed in one PR.

PR #16732, opened 2021-12-01
Fixed kbd nav issues #16633, #16634, #16625.

oysteinmoseng:
Fixed keyboard nav issues #16633, #16634, #16625.


TorsteinHonsi (approved):
Thanks!

PR #16733, opened 2021-12-01
Added new feature, alignThresholds

### LINKED ISSUE #16708: Aligning zero of multiple y-axis
# Description of the feature
It would be great if we had a feature to align 0 of multiple y-axis. I looked in the documentation but couldn't find anything.

I found an old post with custom code for it.
https://highcharts.uservoice.com/forums/55896-general/suggestions/2554384-multiple-axis-alignment-control

And also a project on github
https://github.com/chriskmnds/highcharts-zero-align-y-axes

Could you please consider making it an official feature. It is a very useful feature to have.

Highcharts

Here is a fiddle where you can see the 0 of the two axes are not aligned.
https://jsfiddle.net/fs6z4rLd/

Here is a fiddle where I got them to align my manually setting min on one of the axes.
https://jsfiddle.net/fs6z4rLd/1/

TorsteinHonsi:
Added new feature `chart.alignThresholds` for aligning the thresholds in charts with multiple axes.

Closes #16708.


ts/Core/Axis/Axis.ts
```diff
@@ -59,6 +59,7 @@ const { deg2rad } = H;
 import { Palette } from '../Color/Palettes.js';
 import Tick from './Tick.js';
 import U from '../Utilities.js';
+import PriceEnvelopesIndicator from '../../Stock/Indicators/PriceEnvelopes/PriceEnvelopesIndicator';
```
raf18seb, line 62: Unnecessary import

raf18seb (changes requested):
Poetry.
Just one small unnecessary import and one question/suggestion below

ts/Core/Axis/Axis.ts
```diff
@@ -2168,26 +2182,84 @@ class Axis {
...
+            const thresholdAlignment = thresholdAlignments.length > 1 ?
+                thresholdAlignments.reduce(
+                    (sum: number, n): number => (sum += n),
+                    0
+                ) / thresholdAlignments.length :
+                void 0;
+
+            alignedAxes.forEach((axis): number|undefined => (
+                axis.thresholdAlignment = thresholdAlignment
+            ));
```
raf18seb, line 2260: I was a little bit confused about this arrow function's return. It's working as expected, but semantically forEach shouldn't return a number type.

What about a standard (less confusing) loop?

alignedAxes.forEach((axis): void => {
    axis.thresholdAlignment = thresholdAlignment;
});

The same with `append` and `prepend` functions below.

pawelfus: Hah, just yesterday had the same problem. Arrow functions by definition return something, so TS requires us to define what should be returned. I have decided to use `function () {}` instead 🤷

raf18seb: It could still be an arrow function, but with curly brackets (without an immediate return in the first line). Then it just returns `undefined` (like in my proposal)

pawelfus (changes requested):
Great work!

Two minor issues:
- [demo](https://jsfiddle.net/BlackLabel/w70524kL/3/) - I believe only yAxes on the same pane (height+top) should be aligned
- [demo](https://jsfiddle.net/BlackLabel/w70524kL/4/) - I'm not sure if we can do something about it, but zooming in is limited, and the threshold is always displayed

ts/Core/Axis/Axis.ts
pawelfus: Oh, right! Missed that part, sorry

raf18seb: Yep, code like this might be confusing :D

TorsteinHonsi:
### To do
 - [ ] Deal with axes of different panes
 - [ ] Other review comments
 - [ ] Consider that individual axes can opt out of `alignThresholds`, equivalent to `Axis.alignTicks`
 - [ ] Axes without a threshold (`axis.threshold` is not a number), it should not be subject to `alignThreshold`. Line series must set `threshold: 0`. Implement and explain in docs.

ts/Core/Axis/Axis.ts
```diff
@@ -2168,26 +2182,84 @@ class Axis {
...
                     } else {
-                        (others as any)[key] = 1;
+                        others[key] = 1;
+                    }
+
+                    // Handle alignThresholds. The `thresholdAlignments` array
+                    // keeps records of where each axis in the group wants its
+                    // threshold, from 0 which is on `axis.min`, to 1 which is
+                    // on `axis.max`.
+                    if (alignThresholds) {
```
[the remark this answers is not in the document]
TorsteinHonsi: Okay, agreed

ts/Core/Axis/Axis.ts
```diff
@@ -2243,55 +2315,121 @@ class Axis {
...
+
+            // When the axis is subject to the alignThresholds option
+            if (isNumber(thresholdTickIndex) && isNumber(threshold)) {
+
+                // Throw away the previously computed tickPositions and start
+                // from scratch with only the threshold itself, then add ticks
+                // below the threshold first, then fill up above the threshold.
+                // If we are not able to fill up to axis.max, double the
+                // tickInterval and run again.
+                while (
```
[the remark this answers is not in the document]
TorsteinHonsi: Yes, a bit long function. But this block is so similar to the following that I think we shouldn't refactor it out

raf18seb (approved):
Just as a note to my comment: https://github.com/highcharts/highcharts/pull/16733#discussion_r760528650

The `append` and `prepend` functions use a similar syntax as well and can be replaced with this:
```ts
append = (): void => {
    tickPositions.push(correctFloat(
        tickPositions[tickPositions.length - 1] + tickInterval
    ))
},
```

Both syntaxes are fine, so approval from me.

pawelfus (approved):
Thanks!

PR #16734, opened 2021-12-01
api/15594-mappoint-missing-stickytracking

### LINKED ISSUE #15594: Maps: stickyTracking for mappoint is not documented in API
The API property `mappoint.stickyTracking` is missing in the API for Maps
https://api.highcharts.com/highmaps/plotOptions.mappoint.stickyTracking

#### Product version
Highcharts Maps 9.0.1

raf18seb:
Fixed #15594, the `mappoint.stickyTracking` property was missing in the API.

### DIFF
--- a/ts/Series/MapPoint/MapPointSeries.ts
+++ b/ts/Series/MapPoint/MapPointSeries.ts
@@ -86,7 +86,29 @@ class MapPointSeries extends ScatterSeries {
                 /** @internal */
                 color: Palette.neutralColor100
             }
-        }
+        },
+
+        /**
+         * Sticky tracking of mouse events. When true, the `mouseOut` event on a
+         * series isn't triggered until the mouse moves over another series, or
+         * out of the plot area. When false, the `mouseOut` event on a series is
+         * triggered when the mouse leaves the area around the series' graph or
+         * markers. This also implies the tooltip when not shared. When
+         * `stickyTracking` is false and `tooltip.shared` is false, the tooltip
+         * will be hidden when moving the mouse between series. Defaults to true
+         * for line and area type series, but to false for columns, pies etc.
+         *
+         * @sample {highcharts} highcharts/plotoptions/series-stickytracking-true/
+         *         True by default for line series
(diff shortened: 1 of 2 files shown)


madepiet (approved):

DJTechnoo (approved):

TorsteinHonsi (changes requested):
Thanks!

The API is set up so that objects extends each other. The `mappoint.plotOptions` extend the `scatter.plotOptions` here: 

https://github.com/highcharts/highcharts/blob/e7983c78b905f52cc6d400b23ea3eb780e723767/ts/Series/MapPoint/MapPointSeries.ts#L227

So we should expect that the `stickyTracking` option was also inherited, without us having to redefine it. The reason why it doesn't do so, I think is this line, where `highmaps` is missing:

https://github.com/highcharts/highcharts/blob/e7983c78b905f52cc6d400b23ea3eb780e723767/ts/Series/Scatter/ScatterSeries.ts#L126

raf18seb:
I saw this, but if we go this way, all maps series will be extended (including e.g. `map` series). That's why I decided to extend only the `mappoint` series.
I think `stickyTracking` shouldn't be listed in https://api.highcharts.com/highmaps/series.map

Knowing that, do you still want me to change it to use inheritance, or leave redefined `stickyTracking` only on the `mappoint` series as it is now in this PR?

TorsteinHonsi:
Okay, I see. In that case I think it would be better to exclude this option from the other map series types:

https://github.com/highcharts/highcharts/blob/f0b3c8e8951bf04f29ad78ff4ad246fee2267777/ts/Series/Map/MapSeries.ts#L160

It is better to exclude than to maintain two copies of the identical description. But before excluding you should check it agains the other map series. It may be that the feature actually works there too.

KacperMadej (approved):
No additional comments.

raf18seb:
@TorsteinHonsi, sorry, I was wrong in the first place. `stickyTracking` works for all current Maps series so it should be inherited.

TorsteinHonsi (approved):

PR #16737, opened 2021-12-01
bugfix/15133-drilldown-labels-fadein

### LINKED ISSUE #15133: Regression in map drilldown - data labels should fade in
#### Expected behaviour
On drill down, data labels should fade in slowly after the chart has zoomed

#### Actual behaviour
Data labels appear abruptly before the chart has zoom-animated

https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/maps/demo/map-drilldown/

#### Product version
A regression in v7.0.0. Worked as intended in v6.2.0.

pawelfus:
Fixed #15133, Data Labels were missing the fade-in effect upon drilling down.
___
tl;dr:
\*whew\* quite a tricky issue 😅 
Click on a point to drill down. Observe data labels:
| Current | PR |
|---|---|
| [Pie drilldown](https://jsfiddle.net/BlackLabel/8ecyqda9/1/) | [Pie drilldown](https://jsfiddle.net/BlackLabel/8ecyqda9/2/) |
| [Column drilldown](https://jsfiddle.net/BlackLabel/Ldwc2z4q/1/) | [Column drilldown](https://jsfiddle.net/BlackLabel/Ldwc2z4q/2/) |
| [Maps drilldown v6](https://jsfiddle.net/BlackLabel/fzx1w26t/1/) and [Maps drilldown v9](https://jsfiddle.net/BlackLabel/fzx1w26t/2/) | [Maps drilldown](https://jsfiddle.net/BlackLabel/fzx1w26t/4/) |

A bit longer description:
I have tested three possible solutions:
- set the opacity for each label separately (this is the current implementation in Column drilldown) then fade in - as can be observed above, it's not working. I did not debug this further, because I decided it's not optimal anyway. Better hide the group:
- set the opacity for the group, then fade in. This is broken as [Ken Havard described](https://github.com/highcharts/highcharts/issues/15133#issuecomment-779257583): due to hovering a shape, during the animation, the `setState()` is called. `setState()` operates on the group's opacity. We could add events, refactor a little `setState()` etc. to get the desired result, but that may pollute the code and I really wanted to understand why `drillUp` works fine. And that leads to the third solution:
- hide the group (using `visibility` attr), then fade in using `setTimeout`. Personally, I don't like `setTimeout`s, so it can be further refactored (for example use `complete` callback after drilling up/down animation. Not a big deal, but the current solution starts fading in ~50ms before the animation ends). While mentioning refactoring:

A few ideas on how we could improve Drilldown's codebase:
- make Drilldown a class then use it later as a composition. We already have an object: `chart.drilldown` that holds the `update()` method. Now I have added an extra one (+`chart` property)
- replace `syncTimeout` with a sync code. I think `animation.defer` could be a way to help syncing shapes animation and data labels animation. I'm not sure how tricky it can be..

___
TESTS NOTE:
Perhaps I'm using `lolex` incorrectly, but when I tried to add this test to existing ones it broke other tests in the same `demo.js` - like enabling animation for all of them. That's the reason for creating a separate test for animations.

### DIFF
--- a/ts/Extensions/Drilldown.ts
+++ b/ts/Extensions/Drilldown.ts
@@ -162,7 +162,9 @@ declare module '../Core/Series/SeriesOptions' {
 declare global {
     namespace Highcharts {
         interface ChartDrilldownObject {
+            chart: Chart,
             update(options: DrilldownOptions, redraw?: boolean): void;
+            fadeInGroup(group?: SVGElement): void;
         }
         interface ChartEventsOptions {
             drilldown?: DrilldownCallbackFunction;
@@ -1062,6 +1064,40 @@ Chart.prototype.drillUp = function (): void {
     fireEvent(chart, 'drillupall');
 };
 
+/**
+ * A function to fade in a group. First, the element is being hidden,
+ * then, using `opactiy`, is faded in. Used for example by `dataLabelsGroup`
+ * where simple SVGElement.fadeIn() is not enough, because of other features
+ * (e.g. InactiveState) using `opacity` to fadeIn/fadeOut.
+ *
+ * @requires module:modules/drilldown
+ *
+ * @param {undefined|SVGElement} [group]
+ * The SVG element to be faded in.
+ */
+function fadeInGroup(
+    this: Highcharts.ChartDrilldownObject,
+    group?: SVGElement
+): void {
+    const animationOptions = animObject(
+        (this.chart.options.drilldown as any).animation
+    );
+
+    if (group) {
+        group.hide();
+
+        syncTimeout(
+            function (): void {
+                // Make sure neither the group, or the chart, were destroyed
+                if (group && group.added) {
+                    group.fadeIn();
+                }
+            },
+            Math.max(animationOptions.duration - 50, 0)
+        );
+    }
+}
+
 /* eslint-disable no-invalid-this */
 
 // Add update function to be called internally from Chart.update
@@ -1070,6 +1106,8 @@ addEvent(Chart, 'afterInit', function (): void {
     const chart = this;
 
     chart.drilldown = {
+        chart,
+        fadeInGroup,
         update: function (
             options: Highcharts.DrilldownOptions,
             redraw?: boolean
@@ -1281,10 +1319,11 @@ ColumnSeries.prototype.animateDrilldown = function (init?: boolean): void {
                         animationOptions
                     );
             }
-            if (point.dataLabel) {
-                point.dataLabel.fadeIn(animationOptions);
-            }
         });
+
+        if (chart.drilldown) {
+            chart.drilldown.fadeInGroup(this.dataLabelsGroup);
+        }
 
         // Reset to prototype
         delete this.animate;
@@ -1368,6 +1407,7 @@ if (PieSeries) {
                 (this.chart.drilldownLevels as any)[
                     (this.chart.drilldownLevels as any).length - 1
                 ],
+                dataLabelsGroup = this.dataLabelsGroup,
                 animationOptions =
                     (this.chart.options.drilldown as any).animation;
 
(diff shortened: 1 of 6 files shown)


pawelfus:
Extra note: Drilling up in Maps is missing the animation of the shapes. Broken since v9.3.0, v9.2 works fine: https://jsfiddle.net/BlackLabel/fzx1w26t/5/

ts/Extensions/Drilldown.ts
```diff
@@ -1062,6 +1064,40 @@ Chart.prototype.drillUp = function (): void {
...
+ *
+ * @param {undefined|SVGElement} [group]
+ * The SVG element to be faded in.
+ */
+function fadeInGroup(
+    this: Highcharts.ChartDrilldownObject,
+    group?: SVGElement
+): void {
+    const animationOptions = animObject(
+        (this.chart.options.drilldown as any).animation
```
Izothep, line 1083: Do we need `as any` here? Is it possible for `this.chart.options.drilldown` to not exist in this place? If yes then maybe let's create an if statement above?

pawelfus: Nice catch! 
It would be another nice-to-improve place. If we could have a Drilldown class, with access to `this.options`, that are always defined in JS (and TS), then we could give up a lot of `as any` castings in the Drilldown module. Right now, this `as any` is one out of seven other cases. 

> Is it possible for this.chart.options.drilldown to not exist in this place?

It's not possible.

ts/Extensions/Drilldown.ts
```diff
@@ -1368,6 +1407,7 @@ if (PieSeries) {
                 (this.chart.drilldownLevels as any)[
                     (this.chart.drilldownLevels as any).length - 1
                 ],
+                dataLabelsGroup = this.dataLabelsGroup,
```
pawelfus, line 1410: Unused variable.

Nothing did complain about the existence of this line. Why? 🤔

TorsteinHonsi (commented):
Thanks, looks like a solid piece of work!

Regarding `syncTimeout`, I agree it is okay, though perhaps it would be better to hook up to either the complete callback or an drilldown-after-animation event. Not a big deal.

But there's a question while were dealing with drilldown and inactive opacity. If you load http://localhost:3030/samples/#view/highcharts/demo/pie-drilldown on either a real mobile or emultated in Chrome Dev Tools, you'll notice that the slices stay inactive after drilling down. I would consider that a bug. Is it related to this issue, or should we create a separate one for it?

ts/Extensions/Drilldown.ts
pawelfus: Oh, okay. Thanks for the info!

pawelfus:
> But there's a question while were dealing with drilldown and inactive opacity. If you load http://localhost:3030/samples/#view/highcharts/demo/pie-drilldown on either a real mobile or emultated in Chrome Dev Tools, you'll notice that the slices stay inactive after drilling down. I would consider that a bug. Is it related to this issue, or should we create a separate one for it?

Hmm.. seems to be fine. Could you check any other slice than Chrome? For example, IE looks great (the first slice. Meanwhile, Chrome highlights the first slice, which is really small. Current solution works the same way: https://jsfiddle.net/BlackLabel/8dtz32ne/show Just in case, highlighting the first slice is done by accessibility module

TorsteinHonsi:
It's not a change in the PR, so maybe we should just keep it out of this discussion. But the thing is that when I touch any item, it drills down, then the first item is visible but all other are inactive. I have to touch outside the plot area in order to make them "state normal". I don't think that's the behaviour we want. In column charts we don't have the problem with inactive state, but it still shows the tooltip for the first item, so maybe this was intended behaviour at some point.

pawelfus:
That's correct. This is the default behaviour when you load the a11y module. Check this demo: https://jsfiddle.net/BlackLabel/8dtz32ne/1/show (removed the a11y module). The inactive state in Pie is applied per slice, while in the column series, it is applied per series. If you try to drill down with multiple series, then the result will be similar (the first series will have a hover state, the second one will be inactive): https://jsfiddle.net/BlackLabel/3gnx1q5h/show (of course, it won't happen if you remove a11y)

TorsteinHonsi (approved):

PR #16739, opened 2021-12-02
Fixed #16731, correct TS type for PlotAreasplineOptions.compare.

### LINKED ISSUE #16731: PlotAreasplineOptions.compare mismatch in TypeScript
#### Expected behaviour
`compare` should be restricted only to `'percent' | 'value' | null | undefined`, similar to here: https://github.com/highcharts/highcharts/blob/6a600e6224309869539c7a6679cdfa40f7671171/ts/Series/DataModifyComposition.ts#L79

#### Actual behaviour
From: https://raw.githubusercontent.com/highcharts/highcharts-dist/v9.3.2/highcharts.d.ts

It's typed as optional `string`. Also affects other types: `PlotAreasplinerangeOptions`, `PlotAtrOptions` but I assume it is declared in only one place.

```
export interface PlotAreasplineOptions {
    /**
     * (Highstock) Compare the values of the series against the first non-null,
     * non- zero value in the visible range. The y axis will show percentage or
     * absolute change depending on whether `compare` is set to `"percent"` or
     * `"value"`. When this is applied to multiple series, it allows comparing
     * the development of the series against each other. Adds a `change` field
     * to every point object.
     */
    compare?: string;
```

#### Product version
v.9.3.2

mcenkar:
Fixed #16731, this changes TS declaration from:

```ts
export interface PlotAreasplineOptions {
    /**
     * (Highstock) Compare the values of the series against the first non-null,
     * non- zero value in the visible range. The y axis will show percentage or
     * absolute change depending on whether `compare` is set to `"percent"` or
     * `"value"`. When this is applied to multiple series, it allows comparing
     * the development of the series against each other. Adds a `change` field
     * to every point object.
     */
    compare?: string;
```

to:

```ts
export type OptionsCompareValue = ("percent"|"value");

export interface PlotAreasplineOptions {

    /**
     * (Highstock) Compare the values of the series against the first non-null,
     * non- zero value in the visible range. The y axis will show percentage or
     * absolute change depending on whether `compare` is set to `"percent"` or
     * `"value"`. When this is applied to multiple series, it allows comparing
     * the development of the series against each other. Adds a `change` field
     * to every point object.
     */
    compare?: OptionsCompareValue;
```

### DIFF
--- a/ts/Series/DataModifyComposition.ts
+++ b/ts/Series/DataModifyComposition.ts
@@ -738,6 +738,7 @@ export default DataModifyComposition;
  * @type      {string}
  * @since     1.0.1
  * @product   highstock
+ * @validvalue ["percent", "value"]
  * @apioption plotOptions.series.compare
  */
 


TorsteinHonsi (approved):
Thanks!

PR #16741, opened 2021-12-02
Fixed #16624, markers showing after hover after disabling a11y.

### LINKED ISSUE #16624: Turning off accessibility leaves markers on charts even with markers enabled false
#### Expected behaviour
When markers are not enabled for a chart, turning on or off Accessibility flag should not leave any markers on the chart after hovering.

#### Actual behaviour
A point marker left on the chart when hover over different point or move mouse away from widget.

https://jsfiddle.net/qsno4kxu/2/ -> version 9.3.1
https://jsfiddle.net/24mx7h6w/14/ -> version 9.1.2

1. Open the above jsfiddle example
2. Marker is disabled in series config
2. Click "Disable A11Y" button
3. Use mouse to hover through the chart
4. Markers are left on the chart after hovering

#### Product version
Highcharts 9.3.1
Highcharts 9.1.2

#### Affected browser(s)
Chrome, Firefox, IE

oysteinmoseng:


TorsteinHonsi (approved):
Thanks!

PR #16743, opened 2021-12-02
Fixed #16497, legend kbd nav cause page scroll with container scaling.

### LINKED ISSUE #16497: Legend focus causes unexpected scrolling when container is scaled and accessibility is enabled
#### Expected behaviour
When the legend receives focus, the chart should not scroll out of view.

#### Actual behaviour
When focusing on a legend and a scaling transform is applied, the chart scrolls half way out of view.

This is probably a strange use case, but we do use a scaling transform in our product which unfortunately results in this behaviour. From what I can tell, the root cause is because the "proxy" buttons that gets created is usually located exactly on top of the legend item. When a scaling transform is applied, it seems to be calculating its position incorrectly. This button is the actual element that receives focus, and the browser thinks the thing its focusing is outside of the viewport and tries to scroll it into view for you, causing this issue.

https://jsfiddle.net/bfuLdo5n/15/

#### Product version
9.2.2

#### Affected browser(s)
Google Chrome, Edge

oysteinmoseng:
Fixed #16497, legend keyboard navigation caused page scroll with container scaling.

### DIFF
--- a/ts/Accessibility/ProxyElement.ts
+++ b/ts/Accessibility/ProxyElement.ts
@@ -312,13 +312,15 @@ class ProxyElement {
 
         if (chartDiv && posElement && posElement.getBoundingClientRect) {
             const rectEl = posElement.getBoundingClientRect(),
-                rectDiv = chartDiv.getBoundingClientRect();
+                chartPos = this.chart.pointer.getChartPosition();
 
             return {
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (approved):
Thanks!

Issue #16720, opened 2021-11-30
highcharts.d.ts out of sync?


FrankTaubert:
With the update to 9.3.2 my IDE startet complaining about the signature of  Highcharts.Tooltip#defaultFormatter

#### Expected behaviour
IDE not complaining :stuck_out_tongue:

#### Actual behaviour

looking at  https://github.com/highcharts/highcharts-dist/blob/master/highcharts.d.ts line 140216 it notes the signature as 
`defaultFormatter(): void;`

however, https://github.com/highcharts/highcharts/blob/master/ts/Core/Tooltip.ts line 298  has 
 the signature as:https://github.com/highcharts/highcharts/blob/af8977bafd322c170f435c3872a179de9105694c/ts/Core/Tooltip.ts#L298-L301)

These don't seem to match up?

#### Product version
Highcharts 9.3.2

#### Affected browser(s)
VS Code

KacperMadej:
Hi @FrankTaubert 

Thank you for reporting the problem.

TS definitions in code are used internally for the whole code. Public API, including .d.ts in releases (like highcharts-dist repo, NPM package, etc.), is based on comments - like the one seen above the linked code. That comment lacks info about arguments and return type.

PetarNikValchev:
Hi @KacperMadej,

Can you tell me when is this fix expected to be released?

raf18seb:
Hi @PetarNikValchev,

We are planning to release the next Highcharts version soon. We don't know the exact date yet, but it should be released within a few next weeks at the most.

PetarNikValchev:
Hi @raf18seb,

I can see there was a new release but the fix for this bug was not part of the release. Can you please give us an ETA for this because we are already missing our deadlines because of it?

raf18seb:
Hi @PetarNikValchev,

We released v9.3.3 with one hotfix only and, as you said, unfortunately, the fix for this bug here is not included in v9.3.3.
We are preparing to release v10.0 and we are getting closer, but we still don't have the exact ETA. In my opinion, it would be approximately 2 weeks.
I am sorry for the inconvenience. I'll let you know if I have any specific info on this matter.

PR #16746, opened 2021-12-02
fix missing documentation for ToolTip.defaultFormater

FrankTaubert:
Fixed https://github.com/highcharts/highcharts/issues/16720
A previous change removed parts of the documentation for Highcharts.Tooltip#defaultFormatter leading to an incorrect generation of `.d.ts` files with the 9.3.2 release.

This PR adds that change back in as well as adding a plain `string` as a possible return value.

#### Expected behaviour
signature requires a tooltip parameter and returns a string or string array

#### Actual behaviour
generated signature is: `defaultFormatter(): void;`


TorsteinHonsi (approved):
Thanks, that looks correct.

PR #16756, opened 2021-12-03
feature/8288-boost-areaspline

pawelfus:
Added `areaspline` series to the supported series in the Boost module, see #8288.


samples/unit-tests/axis/ordinal/demo.js
```diff
@@ -564,7 +568,10 @@ QUnit.test('Ordianl axis, data grouping and boost module, #14055.', assert => {
     });
     assert.ok(
         chart.xAxis[0].ordinal.positions.length &&
-            chart.series[0].boostClipRect,
+        (
+            chart.series[0].boostClipRect ||
+            chart.boostClipRect
+        ),
```
pawelfus, line 574: These changes (and the ones above) are related to the fact that now this chart has two supported series types, so a common `chart.boostClipRect` was created instead of a single `series[x].boostClipRect`.

TorsteinHonsi (approved):
Thanks!

PR #16761, opened 2021-12-06
Fixed issue with a11y module and drilldown on mobile.

oysteinmoseng:
Fixed issue on mobile where a point would be highlighted visually after drilldown with a11y module loaded.


TorsteinHonsi (approved):
Thanks!

PR #16765, opened 2021-12-07
Samples: Flags on column comparison

TorsteinHonsi:
Now loading flags from another source, also used elsewhere in our demos. Also more ES6 syntax.


raf18seb (approved):
All looks good, I only added a little bit more ES6 syntax

Issue #15159, opened 2021-02-17
Luminosity ratio for the toggle buttons in legends becomes less than the required ratio on turning them off.


anubhutisaini12:
#### Expected behaviour
On turning the "Tokyo" OFF (any legend off), Luminosity of the color of button should be equal to or more than 3:1.

#### Actual behaviour
On turning the "Tokyo" OFF (any legend off), Its color change and luminosity of the color becomes 1.6:1 which is less than 3:1.

#### Live demo with steps to reproduce

https://www.highcharts.com/demo/column-basic
![image](https://user-images.githubusercontent.com/76141946/108160779-bc263200-710f-11eb-93a6-2ec2e02aec1e.png)

MAS Reference:
MAS 1.4.11 - Non-text Contrast

Browser: Edge chromium Version 87.0.634.0 (Official build) dev (64-bit),
Tool: Color contrast analyzer

KacperMadej:
Hi @anubhutisaini12 

Thank you for reporting the problem.

As a workaround, the color could be adjusted in the chart configuration: https://api.highcharts.com/highcharts/legend.itemHiddenStyle

The lightest grey on white seems to be #949494 (ration >3.03) if we go by the non-text contrast - [demo](https://jsfiddle.net/BlackLabel/wqyngrh0/).
However, shouldn't this go by the text contrast of at least 4.5 (MAS 1.4.3)? Then the lightest grey is #767676 (>4.54) - [demo](https://jsfiddle.net/BlackLabel/wqyngrh0/1/).

fyi (a11y) @oysteinmoseng

oysteinmoseng:
Yes, understanding the text would be considered essential to the operation of this control so this should be evaluated against text contrast - 4.5:1 in this case.

The symbol could still have a lighter color, as long as the text is intelligible. Example: https://jsfiddle.net/b65oenta/. I think this is a better solution - I am not a fan of having the difference between enabled/disabled be too small, that just makes it hard to determine which series are disabled.

We could consider adding an option like `legend.itemHiddenStyle.symbolColor` and then change the default values accordingly. What do you think @TorsteinHonsi?

TorsteinHonsi:
Not sure... 

> I am not a fan of having the difference between enabled/disabled be too small, that just makes it hard to determine which series are disabled.

Indeed. And from an accessibility point of view, that is also unfortunate.

[...]

PR #16768, opened 2021-12-08
bugfix/2667-area-series-legend-symbol

### LINKED ISSUE #2667: Area legend should display area transparency
The Area legend displays the series color and transparency. But the user will recognize the area by the area color/transparency, not by the line color. We even use area charts where the series only differ in transparency,where the current legend is of no use at all.

See http://jsfiddle.net/kNLL4/2/

Possible overlap with issue 1982

pawellysy:
Fixed #2667, legend item of area series didn't apply the opacity property.

### DIFF
--- a/samples/unit-tests/legend/legend-symbol-opacity/demo.css
+++ b/samples/unit-tests/legend/legend-symbol-opacity/demo.css
@@ -1,0 +1,10 @@
+@import "https://code.highcharts.com/css/highcharts.css";
+
+#container {
+    max-width: 800px;
+    margin: 1em auto;
+}
+
+g.highcharts-series {
+    transition: opacity 0ms;
+}
--- a/samples/unit-tests/legend/legend-symbol-opacity/demo.details
+++ b/samples/unit-tests/legend/legend-symbol-opacity/demo.details
@@ -1,0 +1,6 @@
+---
+ resources:
+   - https://code.jquery.com/qunit/qunit-2.0.1.js
+   - https://code.jquery.com/qunit/qunit-2.0.1.css
+ js_wrap: b
+...
--- a/samples/unit-tests/legend/legend-symbol-opacity/demo.html
+++ b/samples/unit-tests/legend/legend-symbol-opacity/demo.html
@@ -1,0 +1,7 @@
+<script src="https://code.highcharts.com/highcharts.js"></script>
+
+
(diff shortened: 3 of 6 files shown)


KacperMadej (approved):
👍

TorsteinHonsi:
I'm not 100% sure this is a good idea. Area charts are sometimes used with a very transparent fill. Like [in this case](https://jsfiddle.net/highcharts/udwxp1ya/), where it would better to use the color of the graphs and markers, not the fill.

It could be that we should make the legend symbol an option. Currently we have two methods in LegendSymbol; `drawRectangle` (which is by default a circle because of the `legend.symbolRadius` setting). What we could do is to refactor this so that we have a series-level `drawLegendSymbol` function that calls either of these functions based on a `series.legendSymbol` option. If so, we need to rename the functions to something more generic, like `lineMarker` and `simple`. 

```ts
public drawLegendSymbol(): void {
    LegendSymbol[this.options.legendSymbol].call(this);
}
```

@KacperMadej , @pawellysy , @pawelfus what do you think?

pawelfus:
> what do you think?

+1 
It will not only keep the old charts without changes but also give better flexibility to the designers. Unless we want this for all series (e.g. column or pie?), I would start with these types:
- area
- areaspline
- arearange
- areasplinerange
- streamgraph
- polygon

KacperMadej:
+1

and https://github.com/highcharts/highcharts/issues/15159 was planned as a separate issue but in case of considering refactor it should be included in the discussion IMHO

TorsteinHonsi:
Ok @pawellysy then we pivot this PR into a small refactor, introducing the new `legendSymbol` option like described above. 

> and #15159 was planned as a separate issue but in case of considering refactor it should be included in the discussion IMHO

I think we should leave that out of this PR. The suggested refactoring in this case is more like moving code around, so it would be easier to keep track of changes if we keep the scope down.

TorsteinHonsi (changes requested):
Ok @pawellysy then we pivot this PR into a small refactor, introducing the new `legendSymbol` option like described above. 

> and #15159 was planned as a separate issue but in case of considering refactor it should be included in the discussion IMHO

I think we should leave that out of this PR. The suggested refactoring in this case is more like moving code around, so it would be easier to keep track of changes if we keep the scope down.

PR #16772, opened 2021-12-08
Website/xmas 2021

nasin:
Hi,

Here's the design for the 2021 Christmas card.


samples/highcharts/website/xmas-2021/demo.js
```diff
@@ -0,0 +1,3937 @@
...
+                    chart.series[0].update({
+                        color: '#E05283'
+                    });
+
+                    chart.series[3].data.forEach(function (point) {
+                        point.update({
+                            low: -30,
+                            high: 60
+                        });
+                    });
```
TorsteinHonsi, line 655: As far as I can see, `series[3]` has multiple data points, so in this case we should call `point.update` with the `redraw` parameter set to false, then call `chart.redraw()` after the loop.

samples/highcharts/website/xmas-2021/demo.js
```diff
@@ -0,0 +1,3937 @@
...
+
+                    setInterval(function () {
+                        topTree.style.transition = 'all 500ms';
+                        if (count % 2 === 0) {
+                            chart.series[3].data.forEach(function (point) {
+                                point.update({
+                                    low: treeData[point.index][0],
+                                    high: treeData[point.index][1]
+                                });
+                            });
```
TorsteinHonsi, line 666: As above, avoid internal redraws in a loop

samples/highcharts/website/xmas-2021/demo.js
```diff
@@ -0,0 +1,3937 @@
...
+                            topTree.style.opacity = 0;
+                        } else {
+                            chart.series[3].data.forEach(function (point) {
+                                point.update({
+                                    low: -30,
+                                    high: 60
+                                });
+                            });
```
TorsteinHonsi, line 674: As above, avoid internal redraws in a loop

samples/highcharts/website/xmas-2021/demo.js
```diff
@@ -0,0 +1,3937 @@
...
+                    coordinates
+                },
+                lineWidth: y === 1 ? 0 : undefined
+            });
+        }
+
+        return data;
+    };
+
+    // Add flight route after initial animation
```
TorsteinHonsi, line 3458: Is this flight route used? I can't see it in the globe view. I should probably be removed (including its call).

samples/highcharts/website/xmas-2021/demo.js
```diff
@@ -0,0 +1,3937 @@
...
+                                setInterval(function () {
+                                    if (count < 360) {
+                                        chart.update({
+                                            mapView: {
+                                                projection: {
+                                                    name: 'Orthographic',
+                                                    rotation: [count, -30]
+                                                }
+                                            }
+                                        });
```
TorsteinHonsi, line 3546: It loooks like this `chart.update` is called with animation. Fourth parameter, true by default unless `chart.animation` is set in the chart options. When updating the chart in a fast loop like this, turning off animation is important so that the browser doesn't have to compute redundant frames between the updates.

samples/highcharts/website/xmas-2021/demo.js
```diff
@@ -0,0 +1,3937 @@
...
+                                        count = count + 1;
+                                    } else {
+                                        chart.update({
+                                            mapView: {
+                                                projection: {
+                                                    name: 'Orthographic',
+                                                    rotation: [count, -30]
+                                                }
+                                            }
+                                        });
```
TorsteinHonsi, line 3556: Not a big deal, but this block can be combined with the previous by just using the modulo operator for count: `rotation: [count % 360, -30]`. Which means that the count will continue to increase past 360, but the rotation will always be between 0 and 360.

TorsteinHonsi (changes requested):
Awesome ⛄  . Some review comments mostly related to performance.

nasin:
Hi, I think I fixed everything.

TorsteinHonsi (approved):

PR #16773, opened 2021-12-09
bugfix/16673-treeGrid-destroy-icon

### LINKED ISSUE #16673: TreeGrid axis does not remove label icons on data update
#### Expected behaviour
After updating data, unnecessary label icons are left.

#### Actual behaviour
The unnecessary label icons should be removed.

Wait 2 secs to observe the issue
https://jsfiddle.net/BlackLabel/3Ljoa0he/3/

#### Product version
Highcharts Gantt v8.1.0+

#### Workaround
Snippet:
```js
Highcharts['_modules']['Core/Axis/TreeGridTick.js'].Additions.prototype.destroy = function() {
  if (this.labelIcon) {
    this.labelIcon.destroy();
  }
}
```

Demo: https://jsfiddle.net/BlackLabel/3Ljoa0he/2/

*Internal note:*
It seems to be a regression after moving TreeGrid into a composition. Previously, all icons used to be removed automatically in `Tick.destroy()`, which simply goes over all properties in a Tick instance, and if `destroy()` is available - calls it. Now, with composition, `Tick.treeGridTick` does not have `destroy()`, so the property is simply deleted, leaving the icon in DOM. Workaround (and the solution) is to add `destroy()` to the composition (like above)

karolkolodziej:
Fixed #16673, label icons in TreeGrid were not destroyed after an update.

### DIFF
--- a/samples/unit-tests/gantt/treegrid/demo.js
+++ b/samples/unit-tests/gantt/treegrid/demo.js
@@ -516,4 +516,11 @@ QUnit.test('series.data[].collapsed', assert => {
         false,
         'This point should be expanded #13838'
     );
+
+    chart.series[0].setData([], true, true, false);
+    assert.notOk(
+        document.querySelector('.highcharts-label-icon'),
+        `After updating data,
+        the unnecessary label icons should be removed, 16673.`
+    );
 });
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #16777, opened 2021-12-10
bugfix/16431-series-trimmed

### LINKED ISSUE #16431: Series of scatter plot don't show under certain circumstances.
#### Expected behaviour
Series can be shown.

#### Actual behaviour
Series don't show up.
If I change the height of chart or max of yAxis, the series can be shown.

https://jsfiddle.net/q7vxj860/6/

#### Product version
Highcharts JS v9.2.2

#### Affected browser(s)
Chrome (didn't test on other browsers)

karolkolodziej:
Fixed #16431, the series was trimmed due to axis floating-point rounding.


TorsteinHonsi:
Seems like it has some side effect. Most of the visual diffs look valid, like the columns now being packed tightly without empty pixels between them. But there seems to be some unwanted visual diffs with polar charts. Can you look into those?

karolkolodziej:
After some debugging, it looks like the issue with the polar chart was connected with how it calculates the radial position of the tick and its label.

TorsteinHonsi (approved):
Thanks, looks good!

PR #16781, opened 2021-12-13
Fixed html error for xmas card 

nasin:
I fixed an error in the HTML that was causing the "nonav" option not to work.


TorsteinHonsi (approved):
Approved, just awaiting the CI run.

Issue #12526, opened 2019-11-28
New Functionality: Chart per point


Izothep:
The main idea of this functionality is to give more support for adding statistical series inside the graph, just like current mappie series works with map areas.
![Screenshot 2019-11-28 at 10 59 58](https://user-images.githubusercontent.com/15050812/69796403-482c0100-11ce-11ea-9f8c-ccf7048ba396.png)

KacperMadej:
> a further activity occurs <

KacperMadej:
a further activity occurs

PR #16783, opened 2021-12-14
Feature: series on point composition

raf18seb:
Added new feature, [series.onPoint](https://api.highcharts.com/highcharts/series.line.onPoint), allowing to display some series types (e.g. pie) over a point with a given id. See [#12526](https://github.com/highcharts/highcharts/issues/12526).


samples/highcharts/series-on-point/series-on-point/demo.js
```diff
@@ -0,0 +1,62 @@
...
+    series: [{
+        type: 'column',
+        data: [
+            { y: 3, id: 'firstPoint' },
+            { y: 5, id: 'secondPoint' },
+            { y: 5, id: 'thirdPoint' },
+            3,
+            2,
+            3
+        ]
```
DJTechnoo, line 30: When toggling the series with the legend, the pies remain while the columns disappear. In the map-pie demo I see that the pies do disappear. Is this intentional, or should it disappear along with the columns?

raf18seb: It should ;) I'll fix it, thanks

TorsteinHonsi (commented):
Looks awesome!

The only comment I can come up with is that it looks pretty tightly coupled to pie and sunburst. But the series-on-point concept should work for any series type that don't involve axes, like funnel, treemap, map and probably others.

KacperMadej (commented):
In map-pies demo: `mapChart` constructor is used and `pie` type series are added. `pie` type series are not a part of the Maps API.

Other than that we talked about smaller things.

ts/Series/SeriesOnPointComposition.ts
```diff
@@ -0,0 +1,517 @@
...
+ *
+ *  Default Export
+ *
+ * */
+
+export default SeriesOnPointComposition;
+
+/* *
+ *
+ *  API Options
```
KacperMadej, line 435: All API options should be available only for the selected series that supports this feature (so pie and sunburst for now). Right now the options are visible for every series type like area or any indicator in Stock.

ts/Series/SeriesOnPointComposition.ts
```diff
@@ -0,0 +1,517 @@
...
+/**
+ * Y position of the series center. By default, the series is displayed on a
+ * point that it's connected to.
+ *
+ * @requires   modules/series-on-point
+ * @since      next
+ * @type       {number}
+ * @apioption  plotOptions.series.onPoint.position.y
+ */
+
```
KacperMadej, line 516: It would be nice to have some API docs for onPoint - the main object. See marker on any series type for comparison. Same for `onPoint.position` and `onPoint.connectorOptions`.

ts/Series/SeriesOnPointComposition.ts
```diff
@@ -0,0 +1,517 @@
...
+ * @default     1
+ * @requires    modules/series-on-point
+ * @type        {number}
+ * @since       next
+ * @product     highcharts highstock
+ * @apioption   plotOptions.series.onPoint.connectorOptions.width
+*/
+
+/**
+ * An id of the point that we connect series to.
```
KacperMadej, line 469: Should the series type limitation be mentioned here? There are some series on which points we cannot (hopefully: yet) add series, like pie, 3d?, indicators?, etc.

KacperMadej (changes requested):
Additional API docs comments below.

raf18seb:
> The only comment I can come up with is that it looks pretty tightly coupled to pie and sunburst. But the series-on-point concept should work for any series type that don't involve axes, like funnel, treemap, map and probably others.

@TorsteinHonsi 
We decided to start with a basic series that work almost out of the box with bubbles' size `z` logic. For other series, an additional logic for size and center calculations must be added.
We can keep working on it now or release what we already have and continue work in the near future - let us know what you decide

KacperMadej (approved):
👍

DJTechnoo (approved):
Wonderful!

madepiet (approved):

TorsteinHonsi:
> We decided to start with a basic series that work almost out of the box with bubbles' size z logic

Okay, but what I'm thinking is that we currently pollute the PieSeries and SunburstSeries with extra functions like `getRadii`, `getRadius`, `getPxExtremes` etc. How would it work if those methods existed on the composition itself? So that the only additional property on the _series_ would be `onPoint`. As far as I can see, that would not be too hard, because the composition events can call those function using the series as scope?

raf18seb:
@TorsteinHonsi 

Hmm, it should be possible, but some changes in (at least) `BubbleSeries.ts` are required.

Let's consider `getRadii` method below. We would have to add a few events and `if` statements (`if (typeof series.getPxExtremes === 'function')...)` to get `getPxExtremes`, `getZExtremes` and `getRadius` values.
https://github.com/highcharts/highcharts/blob/d4e029a10956d675266c36db9f8a0d9af6e99771/ts/Series/Bubble/BubbleSeries.ts#L442-L497


Should I proceed?

TorsteinHonsi:
@raf18seb 

> Hmm, it should be possible, but some changes in (at least) BubbleSeries.ts are required.

I don't think so. The BubbleSeries should keep its `getRadii` and stuff. But it should not be added to `SeriesLike`. Instead, it should be available in the composition.

raf18seb:
Sorry for the late reply!

This is the best I've got (see a commit above) - some changes were still required in `BubbleSeries.ts` because the contexts were wrong (I didn't explain it well enough in my previous comment).
@TorsteinHonsi

ts/Series/SeriesOnPointComposition.ts
```diff
@@ -0,0 +1,576 @@
...
+        ChartClass: typeof Chart
+    ): (typeof SeriesComposition&T) {
+        if (composedClasses.indexOf(SeriesClass) === -1) {
+            composedClasses.push(SeriesClass);
+
+            const pieProto = pie.prototype;
+
+            pieProto.getCenter = seriesGetCenter;
+            pieProto.drawPoints = seriesDrawPoints;
+            pieProto.translate = seriesTranslate;
```
TorsteinHonsi, line 141: Historically we have done a lot of overwriting prototype functions, but I am wondering if we can avoid this. For `getCenter` and `translate` we can probably do it in a better way by using events. Try looking for existing event handlers that can get the job done in time, otherwise we could also add events for those functions. For the `drawPoints` override we can probably use the `afterRender` event hooks instead.

ts/Series/SeriesOnPointComposition.ts
```diff
@@ -0,0 +1,576 @@
...
+            pieProto.bubblePadding = true;
+            pieProto.useMapGeometry = true;
```
TorsteinHonsi, line 143: These are dangerous modifications because they change the prototype for all series even though we don't use `onPoint` functionality. For example, if you drop in the `series-on-point.js` file into the [basic bubble](http://utils.highcharts.local/samples/#view/highcharts/demo/bubble) demo, all bubbles are gone. I suspect that may be because `useMapGeometry` is set to true. Also, the `pie-basic` demo dies if we add the series-on-point file.

TorsteinHonsi: A better solution would probably be to set `this.bubblePadding` and `this.useMapGeometry` conditionally on the series instance inside an event handler like `afterInit`.

ts/Series/SeriesOnPointComposition.ts
```diff
@@ -0,0 +1,576 @@
...
+
+        public connector?: SVGElement;
+
+        public options?: OnPoint;
+
+        /**
+         * @ignore
+         */
+        public getRadii(): void {
+            bubble.prototype.getRadii.call(this.series);
```
TorsteinHonsi, line 305: If we copied over `zData`, `yData` and other properties required to calculate the radii to the composition instance - cloud we then avoid calling this with the main series scope? So we could simply do this:

`public getRadii = bubble.prototype.getRadii;`

And the same for `getRadius`, `getPxExtremes` and `getZExtremes`.

TorsteinHonsi (changes requested):
Thanks, it's getting cleaner! Maybe there is still some room for moving stuff over to the composition. And there's some issues with polluting the prototypes.

ts/Series/SeriesOnPointComposition.ts
raf18seb: I did my best, but I couldn't avoid the context swap in `BubbleSeries.ts` in some cases.

Also, I still needed to save `radii` on the `series` (`series.onPoint.radii` isn't enough)

ts/Series/SeriesOnPointComposition.ts
```diff
@@ -0,0 +1,563 @@
...
+                    position.x :
+                    connectedPoint.plotX,
+                yFrom = defined(position.y) ?
+                    position.y :
+                    connectedPoint.plotY,
+                xTo = xFrom + (position.offsetX || 0),
+                yTo = yFrom + (position.offsetY || 0),
+                width = connectorOpts.width || 1,
+                color = (connectorOpts as any).color || this.series.color,
+                dashStyle = (connectorOpts as any).dashStyle;
```
TorsteinHonsi: The any casts here are because `connectorOptions` is set as SVGAttributes type, but `color` and `dashStyle` don't exist on that type. Either we need a separate interface like `OnPointConnectorOptions`, or we need to use `stroke` and `dashstyle` instead of `color` and `dashStyle`. Either is fine for me.

TorsteinHonsi (changes requested):
Thanks, I think the separation from series is really good now.

I did a final thing to uncouple it from pie and sunburst.
* Add the events to all `Series`
* Check support in the series init event
* Add a global property `onPointSupported` (yes, I know I said we should only have the `onPoint` property, but this is read prior to that :/). With this property, we can easily drop in support for future series types or series types that we didn't think of, like so `SomeSeriesType.prototype.onPointSupported = true`.
@raf18seb Can you check that this works as expected? All tests and demos seems to work.

KacperMadej (changes requested):
- [ ] How is loading sunburst after series-on-point handled? (In case a sunburst series is placed on a point of course)
- [ ] Sunburst's initial animation is funky - it slides into the onPoint-based position from the initial center.
- [x] Connector is not animated on zoom - it just jumps into the final position instead of the expected smooth animation.
- <s>Orthographic projection seems to position a series on a point sometimes wrongly. [Demo](http://jsfiddle.net/BlackLabel/65e4c01g/1/).</s> nvm - it's a dataLabels placement [bug](https://github.com/highcharts/highcharts/issues/17261)

raf18seb:
@KacperMadej 
> Sunburst's initial animation is funky - it slides into the onPoint-based position from the initial center.

I agree, but this is how it works in sunburst by default, see pie vs sunburst comparison: https://jsfiddle.net/BlackLabel/gnydvo3e/

In my opinion, the series-on-point feature shouldn't change the default series behaviors, especially when support for more series types will be added in the future (each series should keep its unique behaviors with series-on-point).

If this should work differently, then it should be fixed on a Sunburst level in general, but not here based only on the series-on-point feature (maybe a separate ticket should be created?)

ts/Series/SeriesOnPointComposition.ts
```diff
@@ -0,0 +1,554 @@
...
+            chartGetZData,
+            seriesAfterInit,
+            seriesAfterRender,
+            seriesGetCenter,
+            seriesShowOrHide,
+            seriesTranslate
+        } = Additions.prototype;
+
+        pie.prototype.onPointSupported = true;
+        sunburst.prototype.onPointSupported = true;
```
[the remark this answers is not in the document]
TorsteinHonsi: I support setting the property in the sunburst module.

madepiet (approved):

KacperMadej (approved):

TorsteinHonsi:
All good regarding review comments, now pending CI test run.

TorsteinHonsi (approved):

PR #16786, opened 2021-12-14
bugfix/16759-rangeSelector-overwrite-dataGrouping

### LINKED ISSUE #16759: rangeSelector buttons overwrite series dataGrouping options when selected by default
#### Expected behaviour
When rangeSelector buttons do not specify an approximation method ([`rangeSelector.buttons.dataGrouping.approximation`](https://api.highcharts.com/highstock/rangeSelector.buttons.dataGrouping.approximation)) and one button is selected by default ([`rangeSelector.selected`](https://api.highcharts.com/highstock/rangeSelector.selected)) the approximation method defined in each Series should be preserved ([`series.line.dataGrouping.approximation`](https://api.highcharts.com/highstock/series.line.dataGrouping.approximation)).

#### Actual behaviour
The expected behaviour is already present when `rangeSelector.selected` is `undefined`: the button `dataGrouping` options are used to patch the series, so any missing property in the button is preserved in the series' `dataGrouping` options.

However, when a button is selected initially through `rangeSelector.selected`, the button's entire `dataGrouping` object *replaces* the series' instead of patching it. The demo below illustrates this inconsistency.

My use case is to dispaly multiple series with different data grouping options in the same chart. To this end, I need to specify those options in the series' `dataGrouping` options, but this can not work if the range selector buttons override them.

https://jsfiddle.net/tx1582qp/1/

#### Product version
Highstock 9.3.2 (latest)

#### Affected browser(s)
n/a

mfilipiec:
Fixed #16759, range selector buttons overwrote series data grouping options when selected by default.

### DIFF
--- a/samples/unit-tests/series/datagrouping/demo.js
+++ b/samples/unit-tests/series/datagrouping/demo.js
@@ -1018,24 +1018,23 @@ QUnit.test('The dataGrouping enabling/disabling.', function (assert) {
     });
 });
 
-QUnit.test('The dataGrouping with selected range.', function (assert) {
+QUnit.test('The dataGrouping with selected range.', (assert) => {
     const chart = Highcharts.stockChart('container', {
         rangeSelector: {
             selected: 0,
             buttons: [{
-                type: "all",
-                text: "All",
+                type: 'all',
+                text: 'All',
                 dataGrouping: {
                     forced: true,
                     units: [
-                        ["millisecond", [3]]
+                        ['millisecond', [3]]
                     ]
                 }
             }]
         },
-        
         series: [{
-            data: [1,2,3,4,5,6,7,8,9],
+            data: [1, 2, 3, 4, 5, 6, 7, 8, 9],
             dataGrouping: {
                 approximation: () => 3
             }
@@ -1045,6 +1044,7 @@ QUnit.test('The dataGrouping with selected range.', function (assert) {
     assert.strictEqual(
         chart.series[0].points[0].y,
         3,
-        `When range is selected dataGrouping should work properly (#16759)`
+        `When the scope is set in the dataGrouping options
+         it should workas well as without it (#16759)`
     );
 });
--- a/samples/unit-tests/series/datagrouping/demo.js
+++ b/samples/unit-tests/series/datagrouping/demo.js
@@ -734,29 +734,50 @@ QUnit.test('Data grouping, custom name in tooltip', function (assert) {
     );
 });
 
-QUnit.test('DataGrouping and update', function (assert) {
+QUnit.test('DataGrouping with selected range and update', function (assert) {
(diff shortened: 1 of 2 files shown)


samples/unit-tests/series/datagrouping/demo.js
```diff
@@ -1016,4 +1016,35 @@ QUnit.test('The dataGrouping enabling/disabling.', function (assert) {
             the series.${prop} property should be deleted.`
         );
     });
+});
+
+QUnit.test('The dataGrouping with selected range.', function (assert) {
```
raf18seb, line 1021: Not really a big deal, but you can use ES6 arrow function here

raf18seb (changes requested):
Thanks! Please see comments below.

- [ ] Also, is it necessary to create the whole QUnit.test? Maybe we could reuse an existing test instead.

samples/unit-tests/series/datagrouping/demo.js
raf18seb: Not really a big deal again, but if there's only one argument, you don't need parentheses. But you can leave it as it is not :)

mfilipiec:
> * [x]  Also, is it necessary to create the whole QUnit.test (and create the whole chart)? Maybe we could reuse an existing test instead.

@raf18seb I was trying to use one of the existing tests but neither of them matched in this case.

raf18seb (changes requested):
(edit)
I see, it needs to be checked on first chart load (when xAxis is not created yet).

I found that you can reuse this test:
```js
QUnit.test('DataGrouping and update', function (assert) {
    const chart = Highcharts.stockChart('container', {
        rangeSelector: {
            selected: 0,
            buttons: [{
                type: 'all',
                text: 'All',
                dataGrouping: {
                    forced: true,
                    units: [
                        ['millisecond', [3]]
                    ]
                }
            }]
        },
        series: [
            {
                id: 'usdeur',
                dataGrouping: {
                    forced: true,
                    approximation: () => 3
                },
                data: [1, 2, 3]
            }
        ]
    });

    assert.strictEqual(
        chart.series[0].points[0].y,
        3,
        `When the scope is set in the dataGrouping options
        it should workas well as without it (#16759)`
    );

    chart.update(
        {
            series: [
                {
                    id: 'usdeur',
                    data: [3, 2, 1]
                },
                {
                    id: 'eurusd',
                    data: [1, 2, 3]
                }
            ]
        },
        true,
        true
    );

    assert.ok(
        true,
        `Should be no errors when:
            - updating one-to-one with IDs
            - DG enabled
        (#11471)`
    );
});
```

- [ ] Please, change the test name accordingly.

TorsteinHonsi (approved):
Thanks!

raf18seb (approved):
Good job, thanks!

PR #16787, opened 2021-12-14
Org chart: Improve mixed layouts, touch #13915.

### LINKED ISSUE #13915: Organization Chart, wrong position for children with normal layout
#### Expected behaviour
Nodes are placed below their parent node if possible.
image

#### Actual behaviour
Position of node's children seems off. 
Both when having mixed layouts in a level or when having all normal layout in a level.
Nodes are shifted to the right.
image

All normal layout: https://jsfiddle.net/alvarezg/rb3ftuLy/
Mixed normal and hanging: https://jsfiddle.net/alvarezg/g7aepx49/

#### Product version

Highcharts, Highstock 8.1.2

pawelfus:
Added new options: `hangingIndentTranslation` and `minNodeLength` for Organization chart, closes #13915.

#### Upgrade note
Children of a parent with `hanging` layout will inherit this layout. To prevent this, set `layout = "normal"` for these nodes.
___
PR based on this comment: https://github.com/highcharts/highcharts/issues/13915#issuecomment-726098027

Current demo: https://jsfiddle.net/BlackLabel/1fh0d4pt/8/

Implemented:
- forced `hanging` layout for children of a parent with `hanging` layout. Downside: chart now requires more space than mixed layouts in `hanging` layout (e.g. [this screenshot](https://github.com/highcharts/highcharts/issues/13915#issuecomment-682197835) - will render blue nodes in a hanging layout, which will require higher container)
- improved finding column (it's an inverted chart, so visually those are rows) - prevents overlapping in hanging layout 
- implemented three possible solutions, how to handle unlimited children in one column (visual column)
  - current logic, [demo](https://jsfiddle.net/BlackLabel/1fh0d4pt/4/) - it's still hard to tell which child is linked to which parent
  - translating children to the right, [demo](https://jsfiddle.net/BlackLabel/1fh0d4pt/5/) - links now look much better, however, there's a risk of overlapping nodes. Simply moving the 'PMO' node next to the 'CORPORATE' node can cause this, [demo](https://jsfiddle.net/BlackLabel/1fh0d4pt/6/). As long as the developer can control the order of the nodes, that won't be a problem.
  - shrinking children, [demo](https://jsfiddle.net/BlackLabel/1fh0d4pt/7/) - links look good, no risk of overlapping, however, very long trees can produce very narrow nodes, making labels barely readable. Added a minimum length of a node, to prevent negative widths

Note:
PR does not fix the OP's issue: https://github.com/highcharts/highcharts/issues/13915#issue-657945624 (placing nodes directly under parents). It seems the discussion drifted away from the OP's report.

Note 2:
Just tested forcing the `hanging` layout. For certain datasets, current solution looks much better, compare these charts: [current](https://jsfiddle.net/BlackLabel/joa4wx7r/) vs [new](https://jsfiddle.net/BlackLabel/joa4wx7r/2/). Perhaps forcing an inherited layout is not the best idea. At least an option to turn this off would be nice


pawelfus:
FYI @TorsteinHonsi @KacperMadej

TorsteinHonsi:
Thanks! As for the different layout scenarios you describe, I think we need to make it all optional. What works best in one case is not necessarily best in other.

KacperMadej:
- Adding an option to translate to the left (with changed connectors placement - on the right side for the left translated nodes) could provide more flexibility.

- Adding an option to prevent translation (centered) might resolve the OP problem as the nodes will be placed like in the current hanging layout with connectors centered. However, this would work nicely only for a single child data set. For >1 children maybe they could be shrunk like in here:
image

pawelfus:
A short summary:

- Implemented new options:
  - `hangingIndentTranslation: 'inherit'|'cumulative'|'shrink'` (`inherit` = default, current). I'm open to suggestions for a better option name and values
  - `minNodeLength: number` - for `shrink` type, to prevent nodes become barely visible. I did not use `minNodeWidth` (like we have `nodeWidth`), because of the confusion in inverted charts. I can change this if necessary.
  - now, by default `layout.hanging` option will be inherited from the parent. First I forced this option, but then realized it may break charts I didn't think of.

Results (added as samples to `hangingIndentTranslation`):
- [current](https://jsfiddle.net/BlackLabel/1fh0d4pt/12/) vs [`inherit`](https://jsfiddle.net/BlackLabel/1fh0d4pt/15/) vs [`shrink`](https://jsfiddle.net/BlackLabel/1fh0d4pt/13/) (sample in API)
- [current](https://jsfiddle.net/BlackLabel/1fh0d4pt/11/) vs [`inherit`](https://jsfiddle.net/BlackLabel/1fh0d4pt/14/) vs [`cumulative`](https://jsfiddle.net/BlackLabel/1fh0d4pt/11/) (sample in API)

ts/Series/Sankey/SankeySeries.ts
```diff
@@ -537,6 +537,24 @@ class SankeySeries extends ColumnSeries {
     public createNodeColumns(): Array<SankeySeries.ColumnArray> {
         const columns: Array<SankeySeries.ColumnArray> = [];
 
+        // Get columns offset including all sibiling and cousins etc.
+        function getOffset(node: SankeyPoint): number {
```
pawelfus, line 541: I wish I could move this code from SankeySeries to OrganizationSeries. Correct me if I'm wrong, but the whole `layout === "hanging"` part below belongs to OrgSeries, right?

TorsteinHonsi: Yes, I can't see a scenario for hanging layout in Sankey.

pawelfus: Thanks! Will do it

TorsteinHonsi (approved):
Awesome, thanks!

KacperMadej (approved):

PR #16796, opened 2021-12-16
bugfix/16135-drillup-category-axis

### LINKED ISSUE #16135: Original series gets messed-up upon drill-up from drill-series back to original-series
#### Expected behaviour
The original series should remain the same before drilling down to drill-series and after drilling up back to original series.

#### Actual behaviour
Upon drill-up from drill down series back to original series, two problems occur:
1. There appear white spaces in original series.
2. The last xAxis label is faulty.

Reproducer of above bug - See [fiddle](https://jsfiddle.net/2dn37Lra/).

#### Product version
Highcharts version `9.1.2`

#### Affected browser(s)
Chrome

karolkolodziej:
Fixed #16135, after drill-up, the series did not go back to its original state.

### DIFF
--- a/ts/Extensions/Drilldown.ts
+++ b/ts/Extensions/Drilldown.ts
@@ -976,11 +976,6 @@ Chart.prototype.drillUp = function (): void {
                 }
             });
 
-            // Reset the names to start new series from the beginning, #16135.
-            if (oldSeries.xAxis && oldSeries.xAxis.names) {
-                oldSeries.xAxis.names.length = 0;
-            }
-
             addedSeries = addedSeries || chart.addSeries(seriesOptions, false);
             if (
                 addedSeries.type === oldSeries.type &&
@@ -993,6 +988,7 @@ Chart.prototype.drillUp = function (): void {
                 newSeries = addedSeries;
             }
         };
+    const drilldownLevelsNumber = (chart.drilldownLevels as any).length;
 
     while (i--) {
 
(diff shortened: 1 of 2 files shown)


ts/Extensions/Drilldown.ts
```diff
@@ -976,6 +976,14 @@ Chart.prototype.drillUp = function (): void {
...
             });
 
+            // For category, axis reset the names that the new series
+            // starts from the beginning, #16135.
+            if (oldSeries.xAxis &&
+                oldSeries.xAxis.options.type === 'category'
+            ) {
+                oldSeries.xAxis.names = [];
+            }
+
```
TorsteinHonsi, line 986: I am a bit concerned that hard-checking the `type` may be too strict. And also that we don't need to reset the names, we could simply empty the existing array. We do a similar thing here:

https://github.com/highcharts/highcharts/blob/7de35360524956689d564ea6b2179ae81a12084f/ts/Extensions/Drilldown.ts#L840-L843

Can you check if the we can use the same, simpler approach in the new case?

TorsteinHonsi (changes requested):

TorsteinHonsi:
There's also a regression with this one: http://utils.highcharts.local/samples/#view/highcharts/drilldown/multi-series
Click Republican, then back. Categories of the parent axis are lost.

TorsteinHonsi (approved):
Thanks!

PR #16801, opened 2021-12-17
Fixed #16722, container size preventing map rotation.

### LINKED ISSUE #16722: Projection Orthographic does not rotate map, when maps container is larger.
#### Expected behaviour
Projection Orthographic should rotate world map, no matter what container height is.

#### Actual behaviour
Projection Orthographic does not rotate world map when height of container is bigger than ~680px;

template: https://jsfiddle.net/RobertRochon/97y0eb2L/1/

When we have the demo `Built in projection` we had #container height set to 500px.
When height is modified to more than ~680px, rotation does not work and standard map navigation is working.

#### Product version

Highcharts Maps 9.3.2

#### Affected browser(s)
Tested on:
Google Chrome - 95.0.4638.54
Firefox - 94.0

DJTechnoo:
Fixed #16722, container size preventing map rotation.


KacperMadej:
There is no linear relation between `minZoom` and `minPlotSize` - tested [here](https://jsfiddle.net/BlackLabel/pje8v9t0/).
There is a linear relation between `minPlotSize` and `getScale` result - see [demo](https://jsfiddle.net/BlackLabel/pje8v9t0/1/).
For changing chart size the factor is constant - [demo](https://jsfiddle.net/BlackLabel/pje8v9t0/2/).
After zooming in the scale goes up, so it would be possible to pick a threshold for a different panning mode (move or rotate).

ts/Maps/MapView.ts
```diff
@@ -164,21 +164,24 @@ class MapView {
                 }
 
                 // Panning rotates the globe
```
TorsteinHonsi, line 166: This comment should be further down, before line 172. Otherwise the context is lost.

KacperMadej (approved):
A conflict to resolve before the merge. Other than that, looks good 👍

madepiet (approved):

samples/unit-tests/maps/map-navigation/demo.js
```diff
@@ -190,3 +190,150 @@ QUnit.test('Map navigation button alignment', assert => {
...
+                geometry: {
+                    type: 'LineString',
+                    coordinates
+                },
+                lineWidth: y === 0 ? 2 : undefined
+            });
+        }
+
+        return data;
+    };
```
raf18seb, line 231: Do we need a graticule in a unit test? It's additional tens of thousands of operations that maybe can be skipped?

raf18seb (changes requested):

samples/unit-tests/maps/map-navigation/demo.js
TorsteinHonsi: You're right. We need something that makes the initial zoom show the full world, but it doesn't need to be this complex. Either a few lines (equator + a handful of meridians), or some wisely placed map points (poles + some around the equator?) should be enough to make the ortho projection zoom out to the full globe.

TorsteinHonsi (approved):
Thanks! Nice test with the partial graticule.

Issue #12522, opened 2019-11-28
New Series Type: Treegraph


Izothep:
Implementing Treegraph series.

Internal note: Treegraph is similar to Organization chart so we should be able to base on its implementation for creating this series.

Izothep:
Internal Note: 
We need to choose which series we should base on.

PR #16805, opened 2021-12-20
Feature/12522-treegraph

pawellysy:
Added new series type: Treegraph. See #12522.
To be finished here: #17600


pawellysy:
#### TODO

- [ ] `alignDataLabel` in inverted chart
- [ ] new API option for Organization series - link, and adding depricated tag for proper options
- [ ] ToggleCollapse option for Treegraph (and possibly Sankey/Organization)
- [ ] VisibleChildren (?) 

#### API DRAFT: 
https://docs.google.com/document/d/1gfl3Be4qr2ONtMmj1mmbF6YpzMJ6BdQ-e4_GaaISxuY/edit#heading=h.h1mhydapqh1r

ts/Series/Treegraph/TreegraphLayout.ts
```diff
@@ -0,0 +1,367 @@
...
+                        child = createDummyNode(parent, child);
+                        gapSize--;
+                    }
+                }
+            });
+        });
+    },
+    // Init the values of the node. probably redundant, since this can be
+    // performed in the init of the chart, but might be useful for repositioning
+    // of the nodes on the update.
```
Izothep, line 161: Have we tested if moving it to the chart init will cause problems after the update? Could it be the part of the nodes constructor?

ts/Series/Treegraph/TreegraphLayout.ts
```diff
@@ -0,0 +1,367 @@
...
+            node.preX = 0;
+        });
+    },
+    // Adds the value to each node, which indicates, what is his sibling number.
+    calculateInitialX: function (
+        this: Highcharts.TreegraphLayout,
+        node: TreegraphNode,
+        index: number
+    ): void {
+        const treeLayout = this as Highcharts.TreegraphLayout;
```
Izothep, line 178: Is the casting needed?

ts/Series/Treegraph/TreegraphLayout.ts
```diff
@@ -0,0 +1,367 @@
...
+        node.linksFrom.forEach((link, index): void => {
+            treeLayout.calculateInitialX(link.toNode, index);
+        });
+        node.relativeXPosition = index;
+    },
+    // recursive post order traversal of the tree, where the initial position of
+    // the node is calculated.
```
Izothep, line 185: Do we want the layout algorithm to be following our API docs structure? I am wondering if that should be visible to our customers. What do you think?

ts/Series/Treegraph/TreegraphLayout.ts
```diff
@@ -0,0 +1,367 @@
...
+        if (node.isLeaf()) {
+            leftSibling = node.getLeftSibling();
+            if (leftSibling) {
+                node.preX = leftSibling.preX + siblingDistance;
+                node.mod = node.preX;
+            } else {
+                node.preX = 0;
+            }
+        } else {
+            // if the node has children perform the recursive first walk for its
```
Izothep, line 203: `children, perform`

ts/Series/Treegraph/TreegraphLayout.ts
```diff
@@ -0,0 +1,367 @@
...
+            // children, and then calculate its shift in the apportion function
+            // (most crucial part part ofthe algorythm)
```
Izothep, line 205: lack of space - `of the`

ts/Series/Treegraph/TreegraphLayout.ts
```diff
@@ -0,0 +1,367 @@
...
+            let defaultAncestor = node.getLeftMostChild() as TreegraphNode;
```
Izothep, line 206: Why do we need this cast?

ts/Series/Treegraph/TreegraphLayout.ts
```diff
@@ -0,0 +1,367 @@
...
+
+            node.linksFrom.forEach(function (link): void {
+                treeLayout.firstWalk(link.toNode as any);
```
Izothep, line 209: `as any`

ts/Series/Treegraph/TreegraphLayout.ts
```diff
@@ -0,0 +1,367 @@
...
+                defaultAncestor = treeLayout.apportion(
+                    link.toNode,
+                    defaultAncestor
+                );
+            });
+            treeLayout.executeShifts(node);
+
+            let leftChild = node.getLeftMostChild() as TreegraphNode;
+            let rightChild = node.getRightMostChild() as TreegraphNode;
```
Izothep, line 218: similar to above, do we need this cast?

ts/Series/Treegraph/TreegraphLayout.ts
```diff
@@ -0,0 +1,367 @@
...
+            // set the position of the parent as a middle point of its children
+            // and move it by the value of the leftSibling (if it exists)
+            let midPoint = (leftChild.preX + rightChild.preX) / 2;
```
Izothep, line 221: I am thinking about moving all the variables to the top part of this method

ts/Series/Treegraph/TreegraphLayout.ts
```diff
@@ -0,0 +1,367 @@
...
+    // shift all children of the current node from right to left.
+    executeShifts: function (
+        this: Highcharts.TreegraphLayout,
+        node: TreegraphNode
+    ): void {
+        let shift = 0,
+            change = 0;
+        for (let i = node.linksFrom.length - 1; i >= 0; i--) {
+            let link = node.linksFrom[i];
+            let childNode = link.toNode as TreegraphNode;
```
Izothep, line 258: do we need the cast?

ts/Series/Treegraph/TreegraphLayout.ts
```diff
@@ -0,0 +1,367 @@
...
+        leftNode: TreegraphNode,
+        rightNode: TreegraphNode,
+        shift: number
+    ): void {
+        let subtrees = rightNode.relativeXPosition - leftNode.relativeXPosition;
+        rightNode.change -= shift / subtrees;
+        rightNode.shift += shift;
+        leftNode.change += shift / subtrees;
+        rightNode.preX += shift;
+        rightNode.mod += shift;
```
Izothep, line 351: It looks that we may separate this block by the `rightNode` and the `leftNode`

ts/Series/Treegraph/TreegraphLayout.ts
```diff
@@ -0,0 +1,367 @@
...
+        if (leftSibling) {
+            let rightIntNode = node;
+            let rightOutNode = node;
+            let leftIntNode = leftSibling;
+            let leftOutNode =
+                rightIntNode.getLeftMostSibling() as TreegraphNode;
+            let rightIntMod = rightIntNode.mod as number;
+            let rightOutMod = rightOutNode.mod as number;
+            let leftIntMod = leftIntNode.mod as number;
+            let leftOutMod = leftOutNode.mod as number;
```
Izothep, line 281: All the above casts - do we need them?

ts/Series/Treegraph/TreegraphLayout.ts
```diff
@@ -0,0 +1,367 @@
...
+            while (
+                leftIntNode &&
+                leftIntNode.nextRight() &&
+                rightIntNode &&
+                rightIntNode.nextLeft()
+            ) {
+                leftIntNode = leftIntNode.nextRight() as any;
+                leftOutNode = leftOutNode.nextRight() as any;
+                rightIntNode = rightIntNode.nextLeft() as any;
+                rightOutNode = rightOutNode.nextLeft() as any;
```
Izothep, line 292: all the `as any` :)

ts/Series/Treegraph/TreegraphLayout.ts
```diff
@@ -0,0 +1,367 @@
...
+                rightOutMod += rightOutNode.mod;
+            }
+            if (
+                leftIntNode &&
+                leftIntNode.nextRight() &&
+                !rightOutNode.nextRight()
+            ) {
+                rightOutNode.thread = (
+                    leftIntNode as TreegraphNode
+                ).nextRight();
```
Izothep, line 324: `as TreegraphNode`

ts/Series/Treegraph/TreegraphLayout.ts
```diff
@@ -0,0 +1,367 @@
...
+                rightOutNode.mod += leftIntMod - rightOutMod;
+            }
+            if (
+                rightIntNode &&
+                rightIntNode.nextLeft() &&
+                !leftOutNode.nextLeft()
+            ) {
+                leftOutNode.thread = (rightIntNode as TreegraphNode).nextLeft();
```
Izothep, line 332: `as TreegraphNode`

ts/Series/Treegraph/TreegraphLayout.ts
```diff
@@ -0,0 +1,367 @@
...
+        for (let i = node.linksFrom.length - 1; i >= 0; i--) {
+            let link = node.linksFrom[i];
+            let childNode = link.toNode as TreegraphNode;
+            childNode.preX += shift;
+            childNode.mod += shift;
+            change += childNode.change;
+            shift += childNode.shift + change;
+        }
+    },
+    apportion: function (
```
Izothep, line 265: Can we do something to make this method a little bit smaller? It is pretty long now.

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -0,0 +1,659 @@
...
+        sankey: { prototype: sankeyProto }
+    }
+} = SeriesRegistry;
+
+import H from '../../Core/Globals.js';
+import U from '../../Core/Utilities.js';
+const { merge, pick, addEvent, stableSort } = U;
+
+import './TreegraphLayout.js';
+import SankeySeries from '../Sankey/SankeySeries.js';
```
Izothep, line 45: Do we need to import this with the 

```
const {
    seriesTypes: {
        sankey: { prototype: sankeyProto }
    }
} = SeriesRegistry;
```
Already imported?

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -0,0 +1,659 @@
...
+    seriesTypes: {
+        sankey: { prototype: sankeyProto }
+    }
+} = SeriesRegistry;
+
+import H from '../../Core/Globals.js';
+import U from '../../Core/Utilities.js';
+const { merge, pick, addEvent, stableSort } = U;
+
+import './TreegraphLayout.js';
```
Izothep, line 44: Shouldn't we instead import the exported part from this file (and create this exported part if does not exist)?

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -0,0 +1,659 @@
...
+             *
+             * @type {'Walker'}
+             * @since next
+             * @default 'Walker'
+             * @product highcharts
+             */
+            layout: 'Walker',
+            /**
+             * Whether the first node should be placed on the opposite side of
+             * the plotArea. In default, the oldest child is positioned on the
```
Izothep, line 110: `By default`

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -0,0 +1,659 @@
...
+             *         Treegraph series with reversed nodes.
+             */
+            reversed: false,
+            /**
+             * @extends   plotOptions.series.marker
+             * @excluding enabled, enabledThreshold
+             */
+            marker: {
+                style: {
+                    cursor: 'pointer'
```
Izothep, line 128: And what if the node is not collapsible? Do we want to have `cursor: 'pointer'` for such nodes as well?

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -0,0 +1,659 @@
...
+        NodesComposition.generatePoints.apply(this, arguments as any);
+
+        /**
+         * Order the nodes, starting with the root node(s). (#9818)
+         * @private
+         */
+        function order(node: TreegraphNode, level: number): void {
+            // Prevents circular dependencies in the data:
+            if (node.wasVisited) {
+                throw Error('circular dependency deteceted');
```
Izothep, line 271: shouldn't we add some links to the further description of an error?

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -0,0 +1,659 @@
...
+                states: options.states
+            }
+        });
+
+        this.layoutAlgorythm.calculatePositions(series);
+        series.layoutModifier = this.getLayoutModifiers();
+
+        // First translate all nodes so we can use them when drawing links
+        this.nodes.forEach((node): void => {
+            // To Change - wasVisited set to false to enable next generatePoints
```
Izothep, line 410: do we need this comment? Maybe we may set the `wasVisited `in point constructor and reset it in the update?

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -0,0 +1,659 @@
...
+            super.drawDataLabels.apply(this, arguments);
+
+            // Render link labels:
+            this.points = this.data;
+            this.options.dataLabels.textPath =
+                this.options.dataLabels.linkTextPath;
+            super.drawDataLabels.apply(this, arguments);
+
+            // Restore nodes
+            this.points = this.points.concat(this.nodes || []);
```
Izothep, line 447: Is there a case when `this.nodes` is not an array?

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -0,0 +1,659 @@
...
+        });
+        // Pass test in drawPoints
+        node.plotY = 1;
+        // Set the anchor position for tooltips
+        node.tooltipPos = chart.inverted ?
+            [plotSizeY - y, plotSizeX - x] :
+            [x, y];
+    }
+
+    public alignDataLabel(): void {
```
Izothep, line 510: Do we need this method?

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -0,0 +1,659 @@
...
+        sankeyProto.alignDataLabel.apply(this, arguments);
+    }
+
+    public createNodeColumns(): Array<
+    SankeyColumnComposition.ArrayComposition<TreegraphNode>
+    > {
+        const originalNodes = this.nodes;
```
Izothep, line 517: I would add some comment on why we are doing this part of code

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -0,0 +1,659 @@
...
+    collapseTreeFromPoint(point, point.collapsed);
+    this.redraw();
+});
+
+function collapseTreeFromPoint(
+    point: TreegraphNode,
+    collapsed: boolean
+): void {
+    point.linksFrom.forEach((link: TreegraphPoint): void => {
+        link.toNode.hidden = collapsed;
```
Izothep, line 540: do we need this option with the `toNode.update` in place?

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -0,0 +1,659 @@
...
+            minXSize = 0,
+            maxYSize = 0,
+            minYSize = 0;
+        this.nodes.forEach((node: TreegraphNode): void => {
+            const markerOptions = merge(
+                this.options.marker,
+                node.options.marker
+            );
+            const symbol = markerOptions.symbol;
+            node.symbol = symbol;
```
Izothep, line 326: Why are we adding it to this place? Looks to not be related to the method

Izothep (changes requested):
Good Job, Thanks! 

There are several changes and questions that might be needed though

ts/Series/Treegraph/TreegraphSeries.ts
pawellysy: I guess it is not necessary, it can be read straight from the options when it is necessary.

ts/Series/Treegraph/TreegraphSeries.ts
pawellysy: This prevents the nodes, that already were collapsed to stay collapsed.

ts/Series/Organization/OrganizationSeriesOptions.d.ts
```diff
@@ -34,11 +34,23 @@ import type { SeriesStatesOptions } from '../../Core/Series/SeriesOptions';
...
+
 export type OrganizationHangingIndentTranslationValue = (
     'inherit'|'cumulative'|'shrink'
 );
 
+
 export interface OrganizationSeriesLevelOptions extends SankeySeriesLevelOptions {
+    borderRadius?: number;
+    linkColor?: ColorString;
+    linkLineWidth?: number;
```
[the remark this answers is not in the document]
pawellysy: the linkLineWidth was an old option, which is now deprecated. The link.lineWidth is a new option. Both of them could be defined in options of individual points, which was possible earlier, but wasn't defined in TS.

ts/Series/Treegraph/TreegraphSeries.ts
pawellysy: same as in https://github.com/highcharts/highcharts/blob/master/ts/Series/Networkgraph/Networkgraph.ts#L58

I wanted to be consistent. What do you think @pawelfus?

ts/Series/Organization/OrganizationSeries.ts
```diff
@@ -509,34 +570,47 @@ class OrganizationSeries extends SankeySeries {
...
+                    (levelOptions.states as any)[state as any]) ||
+                {},
+            borderRadius = pick(
+                stateOptions.borderRadius,
+                options.borderRadius,
+                levelOptions.borderRadius,
+                series.options.borderRadius
+            ),
+
+            linkColor = pick(
```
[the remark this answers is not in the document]
pawellysy: ¯\_(ツ)_/¯. @pawelfus  what do you think?

ts/Series/Treegraph/TreegraphSeries.ts
pawellysy: Moved this property to translateNode.

ts/Series/Treegraph/TreegraphLayout.ts
```diff
@@ -0,0 +1,367 @@
...
+    // modifiers is performed.
+    // link to the paper: http://dirk.jivas.de/papers/buchheim02improving.pdf
+    calculatePositions: function (
+        this: Highcharts.TreegraphLayout,
+        series: TreegraphSeries
+    ): void {
+        const treeLayout = this;
+        const nodes = series.nodes as TreegraphNode[];
+        this.initValues(nodes);
+        this.siblingDistance = 50;
```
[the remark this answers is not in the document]
pawellysy: These are arbitrary values, can be also changed to 1. They only matter to calculate relative position, which can later be changed to match the current plotArea.

ts/Series/Treegraph/TreegraphLayout.ts
pawellysy: it could, but none of these methods would make sense on their own. I guess, that all methods should be made private apart from `calculatePositions`

ts/Series/Treegraph/TreegraphLayout.ts
pawellysy: same as above, yes :)

ts/Series/Treegraph/TreegraphLayout.ts
pawellysy: I would leave it as it is so that it is consistent with the paper, we based on.

ts/Series/Organization/OrganizationSeries.ts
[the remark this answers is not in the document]
pawellysy: I don't know if it really is the best way. @pawelfus  what is your take on this?

ts/Series/Treegraph/TreegraphLayout.ts
```diff
@@ -0,0 +1,367 @@
+import TreegraphSeries from './TreegraphSeries';
+import H from '../../Core/Globals.js';
+import U from '../../Core/Utilities.js';
+const { extend, merge } = U;
+import TreegraphNode from './TreegraphNode.js';
+declare global {
+    namespace Highcharts {
+        class TreegraphLayout {
+            public constructor();
```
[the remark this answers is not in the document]
pawellysy: I did the same think as it is in networkgraph, and I wanted to be consistent. It can be changed if you want.

samples/highcharts/series-treegraph/link-text-path/demo.html
```diff
@@ -0,0 +1,11 @@
+<script src="https://code.highcharts.com/highcharts.js"></script>
+<script src="https://code.highcharts.com/modules/sankey.js"></script>
```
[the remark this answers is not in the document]
pawellysy: yes

ts/Series/Organization/OrganizationSeries.ts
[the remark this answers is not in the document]
pawelfus: No strong opinion here yet, with `reduce` we have castings to `any`. Without, we have more code. I'm leaning towards new solution, because `borderRadius` has fewer options, so `reduce` would end up with just two options.

ts/Series/Treegraph/TreegraphLayout.ts
pawellysy: nvm, you can't do it, if you are extending the algorithm the way we are doing it.

ts/Series/Treegraph/TreegraphSeries.ts
pawelfus: In Networkgraph/PackedBubble, we need some higher-level API, to support multiple series subscribing (adding its nodes and links) to one running layout algorithm. Can we create a chart with multiple tree series (roots)? If yes, then are these rendered separately (each runs its own positioning algorithm, one-by-one) or use a shared process?

test/ts-node-unit-tests/test-utils.ts
```diff
@@ -27,8 +27,8 @@ export function reportError(error: (AssertionError & Error)): void {
     failure(`${code} ${errorMessage}
         ${stack?.split('\n')[1]}
 
-Got: ${printArrayOrString(actual)}
+Got: ${printArrayOrString(actual as string | [])}
```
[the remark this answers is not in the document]
pawelfus: It seems to be a problem with a fresh install. I have the same problem, it seems we have "versionless" `@types/node` dependency. `AssertionErrror` used to be `any` type, now it's `unknown` type. I would probably cast it to `any` as it used to be 🤔

ts/Series/Treegraph/TreegraphLayout.ts
```diff
@@ -0,0 +1,419 @@
...
+     */
+    secondWalk: function (
+        this: Highcharts.TreegraphLayout,
+        node: TreegraphNode,
+        modSum: number
+    ): void {
+        const treeLayout = this;
+        // When the chart is not inverted we want the tree to be positioned from
+        // left to right with root node close to the chart border, this is why
+        // x and y positions are switched.
```
pawelfus, line 243: Q: Shouldn't we somehow check if the chart is inverted then?

ts/Series/Treegraph/TreegraphLayout.ts
```diff
@@ -0,0 +1,419 @@
...
+     */
+    executeShifts: function (
+        this: Highcharts.TreegraphLayout,
+        node: TreegraphNode
+    ): void {
+        let shift = 0,
+            change = 0;
+        for (let i = node.linksFrom.length - 1; i >= 0; i--) {
+            let link = node.linksFrom[i];
+            let childNode = link.toNode;
```
pawelfus, line 263: FYI: Just `let childNode = node.linksFrom[i].toNode` would be enough

pawelfus (changes requested):
Amazing!

A few missing things, though:
- [docs page](https://www.highcharts.com/docs/chart-and-series-types/chart-types)

ts/Series/Treegraph/TreegraphLayout.ts
pawellysy: By naming the variables (e.g. leftIntNode, preX, getLeftChild, etc) the algorithm assumes, that the layout will be positioned from top to bottom, and the default (non-iverted) layout of the nodes in Sankey and organization are from left to right. 
I decided to continue the trend, and the easiest way of switching it was on the last step so that the algorithm assigns values that are consistent with our orientation.

pawelfus: Ah, okay, thanks for the explanation 👍

docs/chart-and-series-types/treegraph-chart.md
```diff
@@ -0,0 +1,32 @@
...
+
+Levels
+------
+
+In an org chart typically, we may want to define specific styling for all nodes on a specific level, for example having all C-level positions in the same color. For this, we can use the [[levels option](https://api.highcharts.com/highcharts/plotOptions.organization.levels) and set the properties there. An example can be seen in the main demo above.
+
+API options
+-----------
+
+For the full set of available options, [see the API](https://api.highcharts.com/highcharts/series.organization).
```
TorsteinHonsi, line 32: `series.organization`?

samples/highcharts/demo/treegraph-chart/demo.js
```diff
@@ -0,0 +1,259 @@
...
+const markerColor = '#0C4160';
+const kingBorderColor = '#cf9700';
+
+Highcharts.chart('container', {
+    chart: {
+        inverted: true,
+        plotBackgroundColor: plotBackgroundColor,
+        borderWidth: 1,
+        width: 800,
+        height: 1200
```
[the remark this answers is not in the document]
pawelfus: No access :(

samples/highcharts/series-treegraph/link-types/demo.js
```diff
@@ -0,0 +1,26 @@
+Highcharts.chart('container', {
+    title: {
+        text: 'Treegraph: Different type of links'
+    },
+    series: [{
+        marker: {
+            radius: 30
```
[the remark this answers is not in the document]
pawellysy: should I add it to the demo or as a default option for dataLabels in Treegraph?

ts/Series/Treegraph/TreegraphSeries.ts
pawellysy: Now, each series creates its own `layoutAlgorithm` instance of the class, and each series is rendered separately. This also means, that the series will overlap, if there is more than one `treegraph` series.

ts/Series/Treegraph/TreegraphNode.ts
```diff
@@ -1,121 +1,195 @@
...
-    public hidden = false;
-    public linksFrom: Array<TreegraphPoint> = void 0 as any;
-    public linksTo: Array<TreegraphPoint> = void 0 as any;
-    public nodeSizeX: number = void 0 as any;
-    public nodeSizeY: number = void 0 as any;
-    public nodeHeight?: number;
-    public series: TreegraphSeries = void 0 as any;
-    wasVisited: any;
+import type TreegraphSeries from './TreegraphSeries.js';
+import type TreemapNodeNS from '../Treemap/TreemapNode.js';
```
Izothep, line 12: should it be `TreemapNodeNS`?

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -124,44 +136,88 @@ class TreegraphSeries extends OrganizationSeries {
...
                  * The HTML of the point's line in the tooltip. Variables are
-                 * enclosed by curly brackets. Available variables are point.x,
-                 * point.y, series.name and series.color and other properties on
-                 * the same form. Furthermore, point.y can be extended by the
-                 * tooltip.valuePrefix and tooltip.valueSuffix variables. This
-                 * can also be overridden for each series, which makes it a good
-                 * hook for displaying units.In styled mode, the dot is
-                 * colored by a class name rather than the point color.
+                 * enclosed by curly brackets. Available variables are
+                 * `point.x`,  `point.y`, `series.name` and `series.color` and
```
Izothep, line 199: Do we use `point.y` in the tooltip at all? I am asking because in another place about the data structure it is mentioned that we don't require any values. Maybe we may create this info more about our specific series?

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -198,21 +253,18 @@ class TreegraphSeries extends OrganizationSeries {
...
-                useHTML: false
-            },
-            link: {
-                /**
-                 * @default 'curved'
-                 *
-                 */
-                type: 'curved'
+                enabled: true,
+                linkFormatter: (): string => '',
```
Izothep, line 257: Why do we need this function? Is it not interfering with the `linkFormat`?

Izothep (changes requested):
Good Job! Small changes requested.

test/ts-node-unit-tests/test-utils.ts
```diff
@@ -27,8 +27,8 @@ export function reportError(error: (AssertionError & Error)): void {
     failure(`${code} ${errorMessage}
         ${stack?.split('\n')[1]}
 
-Got: ${printArrayOrString(actual as string | [])}
+Got: ${printArrayOrString(actual as any)}
```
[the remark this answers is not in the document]
pawellysy: it is the same as in master. I changed it back and forth again, so please make sure you are looking at the diff with master, not since you checked last time :)

ts/Series/Treegraph/TreegraphSeries.ts
pawellysy: It has a lesser priority than `linkFormat`. I guess, that the best is to use the function with the lowest priority so that the users could overwrite it with `pointFormat` or `linkFormat` etc.

ts/Series/Treemap/TreemapSeries.ts
```diff
@@ -1139,7 +1100,7 @@ class TreemapSeries extends ScatterSeries {
                 renderer,
                 shadow,
                 shapeArgs,
-                shapeType: 'rect'
+                shapeType: point.shapeType as any || 'rect'
```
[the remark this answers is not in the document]
pawellysy: There is a collision of types. the point.draw accepts type defined as `DrawPointComposition.DrawParams.shapeType` whereas `point.shapeType` is `string`.

Izothep (approved):
Thanks!

ts/Series/Treegraph/TreegraphLayout.ts
```diff
@@ -0,0 +1,430 @@
...
+                index: number
+            ): TreegraphNode.Node;
+        }
+        let treeLayouts: Record<string, typeof TreegraphLayout>;
+    }
+}
+
+H.treeLayouts = {
+    Walker: function (): void {}
+} as any;
```
pawelfus, line 56: I'm curious if we need to pollute the Highcharts namespace. I know it's done that way in Networkgraph, but perhaps it's unnecessary there too? I'm not saying you should change the Networgraph series too

pawelfus: Could we simply export the layout?

ts/Series/Treegraph/TreegraphLayout.ts
```diff
@@ -0,0 +1,430 @@
...
+            public beforeLayout(nodes: TreegraphNode.Node[]): void;
+            public afterLayout(nodes: TreegraphNode.Node[]): void;
+            public firstWalk(node: TreegraphNode.Node): void;
+            public secondWalk(node: TreegraphNode.Node, modSum: number): void;
+            public executeShifts(node: TreegraphNode.Node): void;
+            public apportion(
+                node: TreegraphNode.Node,
+                defaultAncestor: TreegraphNode.Node
+            ): TreegraphNode.Node;
+            public setArea(x: number, y: number, w: number, h: number): void;
```
pawelfus, line 36: Unnecessary method

ts/Series/Treegraph/TreegraphLayout.ts
```diff
@@ -0,0 +1,430 @@
...
+    },
+    /**
+     * Set the area, in which the nodes should be positioned.
+     *
+     * @param x initial x position
+     * @param y initial y position
+     * @param w width of the area
+     * @param h height of the area
+     */
+    setArea: function (
```
pawelfus, line 403: Unnecessary method

ts/Series/Treegraph/TreegraphPointOptions.d.ts
```diff
@@ -0,0 +1,31 @@
...
+import type { LinkOptions } from '../Organization/OrganizationSeriesOptions';
+
+/* *
+ *
+ *  Declarations
+ *
+ * */
+
+export interface TreegraphPointOptions extends TreemapPointOptions {
+    borderRadius: number;
```
pawelfus, line 27: Q: Shouldn't be optional?

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -0,0 +1,831 @@
...
+             *
+             * @sample highcharts/demo/treegraph-chart
+             *         Treegraph chart
+             *
+             * @type {'Walker'}
+             * @since next
+             * @default 'Walker'
+             * @product highcharts
+             */
+            layout: 'Walker',
```
pawelfus, line 119: For now, I would ignore this option since there's no other possible value. We can add this option later, when more algorithms will be added

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -0,0 +1,831 @@
...
+     * render dataLabels for both sets.
+     */
+    public drawDataLabels(): void {
+        if (this.options.dataLabels) {
+
+            if (!isArray(this.options.dataLabels)) {
+                this.options.dataLabels = [this.options.dataLabels];
+            }
+
+            // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
```
pawelfus, line 566: It seems removing this eslint-disable doesn't cause any issues

pawelfus (changes requested):
Docs:
Are `allowDrillToNode`, `findNearestPointBy `, `layout`, `traverseUpButton`, `levels.layout*`, `zones` and `zoneAxis` options available for Treegraph series too?

pawelfus (approved):
Tremendous work!
Thank you!

TorsteinHonsi:
Thanks, getting close now! A couple of discussion points @pawelfus @pawellysy 

1. The treegraph's cousins, treemap and sunburst, have a [levels option](https://api.highcharts.com/highcharts/series.sunburst.levels). This [doesn't work in the treegraph](https://jsfiddle.net/highcharts/8ph5d1Le/2/). Is it low-hanging fruit to import this feature into the treegraph, or should we wait for the public to request it?

2. About the collapse feature. Currently the nodes are clickable so they can be collapsed. I feel we are missing the last pieces to realize the full potential of this logic:
- `collapse` and `expand` functions on the node/point classes, so that implementers can call this programmatically. Equivalent to the TreeGridAxis collaps and expand functions. Essentially this just means a refactoring to move this logic out of the click event and `collapseTreeFromPoint` function and into the `TreegraphNode`.
- A `collapsed` option on the nodes, equivalent to that of the [gantt series](https://api.highcharts.com/gantt/series.gantt.data.collapsed).
- Last missing piece, and I am not sure how to deal with this. With a collapsed option, we would need some way to convey to the user that they can click the node to expand it. Maybe a small plus sign like like in this [google image search](https://www.google.com/search?q=treegraph+collapsed&source=lnms&tbm=isch&sa=X&ved=2ahUKEwjeqc7pqaL3AhX1R_EDHRB5AecQ_AUoAXoECAEQAw&biw=2560&bih=1496&dpr=1#imgrc=I6tYqHQeGKEWvM)? At the very least we need a CSS class so implementers can style it.

pawellysy:
@TorsteinHonsi I will investigate how much effort it takes to implement both enhancements.

ts/Series/Treegraph/TreegraphPoint.ts
```diff
@@ -0,0 +1,86 @@
...
+
+class TreegraphPoint extends TreemapPoint {
+    public options: TreegraphPointOptions = void 0 as any;
+    public isLink = false;
+    public series: TreegraphSeries = void 0 as any;
+    public collapsed: boolean = false;
+    public node: TreegraphNode.Node = void 0 as any;
+    public level?: number;
+    public linkToParent?: TreegraphLink;
+    hidden: boolean = false;
```
TorsteinHonsi, line 48: We already have the [Point.visible](https://github.com/highcharts/highcharts/blob/338528d0b241d201eb7597cb20244fababe10ea2/ts/Core/Series/Point.ts#L237) property. Do we need to add another property that seems to do exactly the same, but with a negated sign?

pawelfus:
> Last missing piece, and I am not sure how to deal with this. With a collapsed option, we would need some way to convey to the user that they can click the node to expand it. Maybe a small plus sign like like in this [google image search](https://www.google.com/search?q=treegraph+collapsed&source=lnms&tbm=isch&sa=X&ved=2ahUKEwjeqc7pqaL3AhX1R_EDHRB5AecQ_AUoAXoECAEQAw&biw=2560&bih=1496&dpr=1#imgrc=I6tYqHQeGKEWvM)? At the very least we need a CSS class so implementers can style it.

Alternatively, we can add a `collapsed` state for markers. With a proper legend, with data-classes, for example:
- red marker - root (optional)
- blue marker - expanded node
- white/transparent with blue border - collapsed node
- green marker - leaves

ts/Series/Treegraph/TreegraphPoint.ts
pawellysy: You're right. Deleted and replaced :)

TorsteinHonsi:
> Alternatively, we can add a collapsed state for markers

I think that sounds a bit complicated - I mean the end user would have to look at the legend, compare colors etc. And the implementer must set up the legend. The small plus icons are more intuitive in my opinion, and require less configuration, as they can be there by default...

Skjermbilde 2022-04-21 kl  11 38 18

TorsteinHonsi:
> It's just the amount of possible configuration options that make me dizzy

I know... Let's try to keep it simple. So we only need a `distance` from the node, because the position is always pointing towards the children. But we can have `x` and `y` options for offset. Should we use a button and model the options after the context menu button, which has a `symbol`? Or would it be enough to use a `text` option for `+` and `-` like we do with [mapNavigation.buttons](https://api.highcharts.com/highmaps/mapNavigation.buttons)? Under any circumstance we need to think synergy with existing concepts, both in the API and in its implementation.

TorsteinHonsi:
Thanks, it's a good start! I'll come back with more detailed thoughts after releasing v10.0.1. For now I am thinking that we need some options to control whether these buttons are shown. Take for example the `highcharts/demo/treegraph-chart` demo, we don't need all the top nodes to be collapsible. It is also a bit noisy with all the buttons, so we could for example have a setting to only show them on hovering the chart. Plus we need to work on the layout itself.

pawellysy:
@TorsteinHonsi The button can be disabled with `series.collapseButton.enabled = false` as well as the level and point options. 

I think that showing the button on the hover is a good idea though. I will check how it works.

TorsteinHonsi:
Some more things to work on:

1. Since the treegraph now shares the data structure with treemap and sunburst, and each point has one or zero links attached to it, there should be no need for the node/link separation. Yet when I inspect `series.points`, the first half of them are link points.
2. `TreegraphPoint.collapse` and `TreegraphPoint.expand` functions are not implemented.
3.  The `collapsed` option is not working in the `levels` options. This would be a nice feature for example for responsive views where we can reveal more subnodes as the size of the chart gets greater.
4. For the layout of the `collapseButton`, it needs some more discussion. Right now it hides when trying to hover it. Maybe it should partly overlap the shape itself so that it behaves more like an indicator than a button? Also, when a node is collapsed, the button should probably be visible by default. And the centering of the `-` and `+` is not perfect, causing it to look a bit clunky.

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -0,0 +1,935 @@
...
+                 * @default 0.5
+                 * @type {number}
+                 * @since next
+                 * @product highcharts
+                 * @apioption series.treegraph.link.offset
+                 *
+                 */
+            },
+            /**
+             * Options applied to the collapse button.
```
TorsteinHonsi, line 178: What is it? Explain more. Add a live demo showing the options.

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -0,0 +1,935 @@
...
+             */
+            collapseButton: {
+                onlyOnHover: true,
```
TorsteinHonsi, line 181: Missing description

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -0,0 +1,935 @@
...
+                /**
+                 * Offset of the button in the x direction.
+                 * @default 0
+                 */
+                x: 0,
+                /**
+                 * Padding inside the button.
+                 * @default 2
+                 */
+                padding: 2,
```
TorsteinHonsi, line 196: This doesn't work... I tried to set more padding, but it looks like only width and height works.

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -0,0 +1,935 @@
...
+                /**
+                 * Height of the button.
+                 * @default 3
+                 */
+                height: 10,
+                /**
+                 * Width of the button.
+                 * @default 3
+                 */
+                width: 10,
```
TorsteinHonsi, line 211: The `@default` tags don't match the value. In fact, I think the `@default` can be skipped, as the docs generator can read it from the code (verify with `npm run jsdoc`).

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -0,0 +1,935 @@
...
+                 * values between 0 and 1, where 0 is a straight line, and 1 is
+                 * a shape close to the default one.
+                 *
+                 * @default 0.5
+                 * @type {number}
+                 * @since next
+                 * @product highcharts
+                 * @apioption series.treegraph.link.offset
+                 *
+                 */
```
TorsteinHonsi, line 175: Why is this option called `offset`? Judging from the description, it sounds similar to the Sankey [curveFactor](https://api.highcharts.com/highcharts/series.sankey.curveFactor). If it is indeed the same, it should have the same name.

TorsteinHonsi (changes requested):
In addition to the inline comments, the ability to set a level-specific `collapse` is not implemented: https://jsfiddle.net/highcharts/Lrqwkuhy/2/ . It would come in handy in a responsive setup where we could cut off leaf nodes to save space.

pawellysy:
@TorsteinHonsi I have one question about the logic of setting the level-specific collapse option. Take a look at this demo: https://jsfiddle.net/highcharts/Lrqwkuhy/2/

When setting the `collapsed: true` for level 5, some of the leaves, which have their level manually set, don't have the parent point on that level, and are not hidden. Is this solution acceptable?

TorsteinHonsi:
Interesting case... Yes I think it is acceptable at least for now. If it becomes a problem down the road we can discuss it then, perhaps introduce options for how to deal with it.

ts/Series/Treegraph/TreegraphPoint.ts
```diff
@@ -0,0 +1,205 @@
...
+                    align: 'center',
+                    zIndex: 1
+                })
+                .addClass('highcharts-tracker')
+                .addClass('highcharts-collapse-button')
+                .removeClass('highcharts-no-tooltip')
+                .css(style || {})
+                .add(parentGroup);
+
+            (point.collapseButton.element as any).point = point;
```
TorsteinHonsi, line 118: It's somewhat bad practice to put this reference to a DOM element. Can we refer it somewhere else?

TorsteinHonsi (changes requested):

TorsteinHonsi:
@bre1470 Can you review this PR with your TypeScript glasses on?

Some of it is refactoring, moving existing pieces of code around. The treegraph part is new. I have removed some any casts and suspicious typing. Most of what is left now I think is more general type problems like class members (initialized to `undefined` because we use the `init` instead of constructor, and any casts on options because we haven't done the overall change to non-optional series options.

ts/Series/Treegraph/TreegraphLayout.ts
```diff
@@ -0,0 +1,347 @@
...
+ *
+ *  (c) 2010-2022 Pawel Lysy Grzegorz Blachlinski
+ *
+ *  License: www.highcharts.com/license
+ *
+ *  !!!!!!! SOURCE GETS TRANSPILED BY TYPESCRIPT. EDIT TS FILE ONLY. !!!!!!!
+ *
+ * */
+
+import type TreegraphSeries from './TreegraphSeries';
```
bre1470, line 11: - use strict missing
- Imports section comment missing

ts/Series/Treegraph/TreegraphLayout.ts
```diff
@@ -0,0 +1,347 @@
...
+import TreegraphNode from './TreegraphNode.js';
+
+class TreegraphLayout {
```
bre1470, line 14: - Class section comment missing
- add doclet for the class that it is private

ts/Series/Treegraph/TreegraphLayout.ts
```diff
@@ -0,0 +1,347 @@
...
+    /**
```
bre1470, line 15: Functions section comment missing

ts/Series/Treegraph/TreegraphLink.ts
```diff
@@ -0,0 +1,68 @@
+import type PointOptions from '../../Core/Series/PointOptions.js';
```
bre1470, line 1: - File header missing
- use strict missing
- Imports section missing etc.

ts/Series/Treegraph/TreegraphLink.ts
```diff
@@ -0,0 +1,68 @@
...
+import U from '../../Core/Utilities.js';
+const { pick, extend } = U;
+
+export interface LinkPointOptions extends TreegraphPointOptions {
+    link?: LinkOptions
+}
+class LinkPoint extends Point {
+    options: LinkPointOptions = void 0 as any;
+    fromNode: TreegraphPoint = void 0 as any;
+    toNode: TreegraphPoint = void 0 as any;
```
bre1470, line 16: You can move these properties to class prototype interface to avoid the any assignment.

ts/Series/Treegraph/TreegraphNode.ts
```diff
@@ -0,0 +1,195 @@
...
+ *
+ *  (c) 2010-2022 Pawel Lysy Grzegorz Blachlinski
+ *
+ *  License: www.highcharts.com/license
+ *
+ *  !!!!!!! SOURCE GETS TRANSPILED BY TYPESCRIPT. EDIT TS FILE ONLY. !!!!!!!
+ *
+ * */
+
+import type TreegraphSeries from './TreegraphSeries.js';
```
bre1470, line 11: - use strict missing
- Imports section comment missing etc.

ts/Series/Treegraph/TreegraphNode.ts
```diff
@@ -0,0 +1,195 @@
...
+            prototype: {
+                NodeClass: TreemapNode
+            }
+        }
+    }
+} = SeriesRegistry;
+
+namespace TreegraphNode {
+    export interface NodeValuesObject extends TreemapNodeNS.NodeValuesObject {}
+    export class Node extends TreemapNode {
```
bre1470, line 28: Restructure to class namespace
```
class TreegraphNode
namespace TreegraphNode
    export NodeValuesObject
```

ts/Series/Treegraph/TreegraphNode.ts
```diff
@@ -0,0 +1,195 @@
...
+        public mod: number = 0;
+        public thread?: Node;
+        public ancestor: Node = void 0 as any;
```
bre1470, line 31: Move to class interface to avoid any cast.

ts/Series/Treegraph/TreegraphNode.ts
```diff
@@ -0,0 +1,195 @@
...
+        public shift: number = 0;
+        public change: number = 0;
+        public children: Array<Node> = [];
+        public x?: number;
+        public y?: number;
+        public preX: number = 0;
+        public point: TreegraphPoint = void 0 as any;
```
bre1470, line 38: Move to class interface to avoid any cast.

ts/Series/Treegraph/TreegraphNode.ts
```diff
@@ -0,0 +1,195 @@
...
+        public column: number = void 0 as any;
```
bre1470, line 39: Number properties should get a default value instead of `void 0 as any`. Use `NaN` if it should fail by default.

ts/Series/Treegraph/TreegraphPoint.ts
```diff
@@ -0,0 +1,205 @@
...
+ *
+ *  (c) 2010-2022 Pawel Lysy Grzegorz Blachlinski
+ *
+ *  License: www.highcharts.com/license
+ *
+ *  !!!!!!! SOURCE GETS TRANSPILED BY TYPESCRIPT. EDIT TS FILE ONLY. !!!!!!!
+ *
+ * */
+
+/* *
```
bre1470, line 11: use strict missing

ts/Series/Treegraph/TreegraphPoint.ts
```diff
@@ -0,0 +1,205 @@
...
+                pointClass: TreemapPoint
+            }
+        }
+    }
+} = SeriesRegistry;
+
+
+/* *
+ *
+ *  Class
```
bre1470, line 41: Should come before class, here it is Declarations section comment.

ts/Series/Treegraph/TreegraphPoint.ts
```diff
@@ -0,0 +1,205 @@
...
+ * */
+
+interface ShapeArgs {
+    x: number;
+    y: number;
+    height: number;
+    width: number;
+}
+
+class TreegraphPoint extends TreemapPoint {
```
bre1470, line 52: Class doclet to mark class private missing

ts/Series/Treegraph/TreegraphPoint.ts
```diff
@@ -0,0 +1,205 @@
...
+    public options: TreegraphPointOptions = void 0 as any;
+    public isLink = false;
+    public collapseButton?: SVGElement;
+    public series: TreegraphSeries = void 0 as any;
+    public collapsed?: boolean;
+    public node: TreegraphNode.Node = void 0 as any;
+    public level?: number;
+    public linkToParent?: TreegraphLink;
+
+    draw(): void {
```
bre1470, line 62: - Functions section comment missing
- public keyword for class functions are missing

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -0,0 +1,948 @@
...
+import U from '../../Core/Utilities.js';
+const { merge, pick, relativeLength, isArray } = U;
+
+import TreegraphLink from './TreegraphLink.js';
+import { Palette } from '../../Core/Color/Palettes.js';
+import { DataLabelTextPathOptions } from '../../Core/Series/DataLabelOptions.js';
+import TreegraphLayout from './TreegraphLayout.js';
+import { TreegraphSeriesLevelOptions } from './TreegraphSeriesOptions.js';
+
+interface LayoutModifiers {
```
bre1470, line 50: Declarations section comment missing

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -0,0 +1,948 @@
...
+     *               lineWidth, negativeColor, nodes, sortIndex, zoneAxis,
+     *               zones
+     *
+     * @since        next
+     * @product      highcharts
+     * @requires     modules/treemap.js
+     * @requires     modules/treegraph.js
+     * @optionparent plotOptions.treegraph
+     */
+    public static defaultOptions: TreegraphSeriesOptions = merge(
```
bre1470, line 105: Because of the size of the default options this should go into a separate file named `TreegraphSeriesDefaults.ts`.

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -0,0 +1,948 @@
...
+    );
+
+    /* *
+     *
+     *  Properties
+     *
+     * */
+
+    /* eslint-disable valid-jsdoc */
+    public data: Array<TreegraphPoint> = void 0 as any;
```
bre1470, line 305: Consider empty arrays.

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -0,0 +1,948 @@
...
+
+    public options: TreegraphSeriesOptions = void 0 as any;
```
bre1470, line 307: Consider empty objects.

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -0,0 +1,948 @@
...
+            by = maxY === minY ? plotSizeY / 2 : -ay * minY + minYSize / 2,
+            ax = maxX === minX ?
+                1 :
+                (plotSizeX - (maxXSize + maxXSize) / 2) / (maxX - minX),
+            bx = maxX === minX ? plotSizeX / 2 : -ax * minX + minXSize / 2;
+
+        return { ax, bx, ay, by };
+    }
+
+    getLinks(): TreegraphLink[] {
```
bre1470, line 413: No implicit function scope, add public, protected or private.

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -0,0 +1,948 @@
...
+            const dataLabelOptions = this.options.dataLabels.map(
+                (option): DataLabelTextPathOptions | undefined =>
+                    option.textPath
+            );
+            // Render node labels.
+            this.drawNodeLabels.apply(this, arguments);
+
+            // Render link labels.
+            const points = this.points;
+            this.points = this.links as any;
```
bre1470, line 675: If this needs `any`, this calls for a fix. Links and points should have a type relation.

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -0,0 +1,948 @@
...
+}
+/* *
+ *
+ *  Registry
+ *
+ * */
+
+TreegraphSeries.prototype.pointClass = TreegraphPoint;
+TreegraphSeries.prototype.NodeClass = TreegraphNode.Node;
+TreegraphSeries.prototype.LinkClass = TreegraphLink;
```
bre1470, line 829: - empty namespace can be removed
- property assignment should come before Registry section
- do we need the other point classes really on the prototype? The pointClass is for sake of compatibility.

ts/Series/Treemap/TreemapNode.ts
```diff
@@ -0,0 +1,55 @@
+import type TreemapSeries from './TreemapSeries.js';
```
bre1470, line 1: - File header missing
- use strict missing
- Imports section comment missing

ts/Series/Treemap/TreemapNode.ts
```diff
@@ -0,0 +1,55 @@
...
+import type BBoxObject from '../../Core/Renderer/BBoxObject.js';
+import type TreemapPoint from './TreemapPoint.js';
+
+namespace TreemapNode {
+    export interface NodeValuesObject extends BBoxObject {
+        direction: number;
+        val: number;
+    }
+
+    export class Node {
```
bre1470, line 11:
```
class TreemapNode
namespace TreemapNode
     interface NodeValuesObject
```

ts/Series/Organization/OrganizationSeriesOptions.d.ts
```diff
@@ -31,14 +31,30 @@ import type { SeriesStatesOptions } from '../../Core/Series/SeriesOptions';
  *  Declarations
  *
  * */
+export type OrganizationChartLinkTypes = 'curved' | 'straight' | 'default';
+
 
 export type OrganizationNodesLayoutValue = ('normal'|'hanging');
 
+export interface LinkOptions {
```
bre1470, line 39: Please name it `OrganizationLinkOptions`.

bre1470 (changes requested):
God job! 

I have requested mostly structural changes. After that you should check the any casts in Treegraph. Some are not related to options and can be fixed with propper typing. If you have questions or need help, let me know.

ts/Core/Series/DataLabelOptions.d.ts
```diff
@@ -33,6 +33,12 @@ import type SVGAttributes from '../Renderer/SVG/SVGAttributes';
 
 export type DataLabelFilterOperatorValue = ('>'|'<'|'>='|'<='|'=='|'===');
 
+export interface TextPathAttributes extends SVGAttributes {
+    startOffset?: string;
+    textAnchor?: 'start'|'middle'|'end';
+    dy?: number;
```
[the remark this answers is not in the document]
TorsteinHonsi: It is a special type of SVGAttribute, so I think we can keep it separate for now.

ts/Series/Treegraph/TreegraphPoint.ts
pawellysy: Isn't it the go-to way of making the custom renderer object have the same tooltip as a point?

ts/Core/Renderer/BBoxObject.d.ts
```diff
@@ -24,6 +24,7 @@ import type SizeObject from './SizeObject';
  * */
 
 export interface BBoxObject extends PositionObject, SizeObject {
+    centerX?: number;
```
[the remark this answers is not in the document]
pawellysy: I wanted to add this property but backed out from it, kept this property anyway, because I've seen, that it is used here: https://github.com/highcharts/highcharts/blob/34d9e7d7d07186a9da34b18b7c358000472f2501/ts/Core/Series/DataLabel.ts#L198

I can delete it, if you think so.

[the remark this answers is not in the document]
bre1470: Oh, I see. Please delete it and add a `// @todo` in DataLabel to create a custom dlBox type there.

ts/Series/Treegraph/TreegraphSeries.ts
pawellysy: I added an extra reference to `NodeClass` in a similar fashion to `pointClass` because I figured, that it would be a way of creating the object using the correct class in Treemap/Sunburst/Treegraph series. When overwriting this property, the method `buildTree` creates correct nodes: https://github.com/highcharts/highcharts/blob/c994a3a6d2500f0473b1aa6c96bc892c8a6a05d1/ts/Series/Treemap/TreemapSeries.ts#L1294-L1302

Is there a better way?

bre1470: It is possible to move it to the constructor, but I am not happy with the required type hack. So I have no objection anymore. Just for curiosity the necessary changes would look as follows:

```diff
diff --git a/ts/Series/Sunburst/SunburstSeries.ts b/ts/Series/Sunburst/SunburstSeries.ts
index 469330732f..28130f42f3 100644
--- a/ts/Series/Sunburst/SunburstSeries.ts
+++ b/ts/Series/Sunburst/SunburstSeries.ts
@@ -418,6 +418,8 @@ class SunburstSeries extends TreemapSeries {
      *
      * */
 
+    public static readonly NodeClass = SunburstNode;
+
     /**
      * A Sunburst displays hierarchical data, where a level in the hierarchy is
      * represented by a circle. The center represents the root node of the tree.
@@ -1153,10 +1155,10 @@ class SunburstSeries extends TreemapSeries {
  * */
 
 interface SunburstSeries {
+    constructor: typeof SunburstSeries;
     getCenter: typeof CU['getCenter'];
     pointClass: typeof SunburstPoint;
     utils: typeof SunburstUtilities;
-    NodeClass: typeof SunburstNode;
 }
 extend(SunburstSeries.prototype, {
     drawDataLabels: noop, // drawDataLabels is called in drawPoints
@@ -1165,9 +1167,8 @@ extend(SunburstSeries.prototype, {
     onPointSupported: true,
     pointAttribs: ColumnSeries.prototype.pointAttribs as any,
     pointClass: SunburstPoint,
-    NodeClass: SunburstNode,
     utils: SunburstUtilities
-});
+} as SunburstSeries);
 
 /* *
  *
diff --git a/ts/Series/Treegraph/TreegraphNode.ts b/ts/Series/Treegraph/TreegraphNode.ts
index 65f733fb0b..c734dc8aac 100644
--- a/ts/Series/Treegraph/TreegraphNode.ts
+++ b/ts/Series/Treegraph/TreegraphNode.ts
@@ -25,9 +25,7 @@ import SeriesRegistry from '../../Core/Series/SeriesRegistry.js';
 const {
     seriesTypes: {
         treemap: {
-            prototype: {
-                NodeClass: TreemapNode
-            }
+            NodeClass: TreemapNode
         }
     }
 } = SeriesRegistry;
diff --git a/ts/Series/Treegraph/TreegraphSeries.ts b/ts/Series/Treegraph/TreegraphSeries.ts
index 87066113be..4976afbd4e 100644
--- a/ts/Series/Treegraph/TreegraphSeries.ts
+++ b/ts/Series/Treegraph/TreegraphSeries.ts
@@ -69,12 +69,17 @@ interface LayoutModifiers {
  * @augments Highcharts.Series
  */
 class TreegraphSeries extends TreemapSeries {
+
     /* *
      *
      *  Static Properties
      *
      * */
 
+    public static readonly LinkClass = TreegraphLink;
+
+    public static readonly NodeClass = TreegraphNode;
+
     /**
      * A treegraph series is a diagram, which shows a relation between ancestors
      * and descendants with a clear parent - child relation.
@@ -419,7 +424,7 @@ class TreegraphSeries extends TreemapSeries {
             if (point.node.parent) {
                 const pointOptions = merge(levelOptions, point.options);
                 if (!point.linkToParent || !point.linkToParent.update) {
-                    const link = new series.LinkClass().init(
+                    const link = new series.constructor.LinkClass().init(
                         series,
                         pointOptions,
                         void 0,
@@ -808,16 +813,11 @@ class TreegraphSeries extends TreemapSeries {
  * */
 
 interface TreegraphSeries {
+    constructor: typeof TreegraphSeries;
     inverted?: boolean;
     pointClass: typeof TreegraphPoint;
-    NodeClass: typeof TreegraphNode;
-    LinkClass: typeof TreegraphLink;
-
 }
 
-namespace TreegraphSeries {
-
-}
 /* *
  *
  *  Registry
@@ -825,8 +825,6 @@ namespace TreegraphSeries {
  * */
 
 TreegraphSeries.prototype.pointClass = TreegraphPoint;
-TreegraphSeries.prototype.NodeClass = TreegraphNode;
-TreegraphSeries.prototype.LinkClass = TreegraphLink;
 declare module '../../Core/Series/SeriesType' {
     interface SeriesTypeRegistry {
         treegraph: typeof TreegraphSeries;
diff --git a/ts/Series/Treemap/TreemapSeries.ts b/ts/Series/Treemap/TreemapSeries.ts
index 85c6080a05..29003bd506 100644
--- a/ts/Series/Treemap/TreemapSeries.ts
+++ b/ts/Series/Treemap/TreemapSeries.ts
@@ -100,6 +100,8 @@ class TreemapSeries extends ScatterSeries {
      *
      * */
 
+    public static readonly NodeClass = TreemapNode;
+
     /**
      * A treemap displays hierarchical data using nested rectangles. The data
      * can be laid out in varying ways depending on options.
@@ -1291,7 +1293,7 @@ class TreemapSeries extends ScatterSeries {
             height = Math.max(child.height + 1, height);
             children.push(child);
         });
-        node = new series.NodeClass().init(
+        node = new series.constructor.NodeClass().init(
             id,
             index,
             children,
@@ -1922,6 +1924,7 @@ class TreemapSeries extends ScatterSeries {
 interface TreemapSeries extends ColorMapComposition.SeriesComposition, TU.Series {
     colorAttribs: ColorMapComposition.SeriesComposition['colorAttribs'];
     colorKey: string;
+    constructor: typeof TreemapSeries;
     directTouch: boolean;
     drawLegendSymbol: typeof LegendSymbol.drawRectangle;
     getExtremesFromAll: boolean;
@@ -1929,7 +1932,6 @@ interface TreemapSeries extends ColorMapComposition.SeriesComposition, TU.Series
     parallelArrays: Array<string>;
     pointArrayMap: Array<string>;
     pointClass: typeof TreemapPoint;
-    NodeClass: typeof TreemapNode;
     trackerGroups: Array<string>;
     utils: {
         recursive: typeof TreemapUtilities.recursive;
@@ -1947,12 +1949,11 @@ extend(TreemapSeries.prototype, {
     parallelArrays: ['x', 'y', 'value', 'colorValue'],
     pointArrayMap: ['value'],
     pointClass: TreemapPoint,
-    NodeClass: TreemapNode,
     trackerGroups: ['group', 'dataLabelsGroup'],
     utils: {
         recursive: TreemapUtilities.recursive
     }
-});
+} as TreemapSeries);
 ColorMapComposition.compose(TreemapSeries);
 
 /* *
```

ts/Series/Treegraph/TreegraphLink.ts
```diff
@@ -0,0 +1,89 @@
...
+export interface TreegraphLinkOptions extends OrganizationLinkOptions {}
+export interface LinkPointOptions extends TreegraphPointOptions {
+    link?: TreegraphLinkOptions
+}
+interface LinkPoint {
+    options: LinkPointOptions;
+    fromNode: TreegraphPoint;
+    toNode: TreegraphPoint;
+}
+class LinkPoint extends Point {
```
bre1470, line 37: - Class interface should follow the class itself
- Add section comments

ts/Series/Treegraph/TreegraphNode.ts
```diff
@@ -0,0 +1,206 @@
...
+    seriesTypes: {
+        treemap: {
+            prototype: {
+                NodeClass: TreemapNode
+            }
+        }
+    }
+} = SeriesRegistry;
+
+interface TreegraphNode {
```
bre1470, line 35: - Interface for class prototype should come after class itself
- Section comments

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -0,0 +1,736 @@
...
+        // Set the anchor position for tooltip.
+        point.tooltipPos = chart.inverted ?
+            [plotSizeY - nodeY - height / 2, plotSizeX - nodeX - width / 2] :
+            [nodeX + width / 2, nodeY];
+    }
+}
+
+/* *
+ *
+ *  Prototype Properties
```
bre1470, line 594: = `Class Prototype`

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -0,0 +1,736 @@
...
+}
+/* *
+ *
+ *  Registry
+ *
+ * */
+
+TreegraphSeries.prototype.pointClass = TreegraphPoint;
+TreegraphSeries.prototype.NodeClass = TreegraphNode;
+TreegraphSeries.prototype.LinkClass = TreegraphLink;
```
bre1470, line 617: This should come directly after the `interface TreegraphSeries` and you should consider use of `Utilities.extend` to save some code.

ts/Series/Treemap/TreemapNode.ts
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * Imports
+ *
+ * */
+
+import type TreemapSeries from './TreemapSeries.js';
+import type BBoxObject from '../../Core/Renderer/BBoxObject.js';
+import type TreemapPoint from './TreemapPoint.js';
+
+interface TreemapNode {
```
bre1470, line 23: - Interface for class properties should follow the class
- Section comments missing

bre1470 (changes requested):

ts/Series/DrawPointUtilities.ts
```diff
@@ -128,6 +128,8 @@ function draw(
  */
 function shouldDraw(point: Point): boolean {
     switch (point.series && point.series.type) {
+        case 'treegraph':
+            return isNumber(point.plotY) && point.y !== null && point.visible;
```
pawellysy, line 132: @bre1470 Can't this function be migrated as a SeriesPoint method?

bre1470 (approved):
Thank you. Looks good! 👍

PR #16812, opened 2021-12-21
website/font fix for xmas card

nasin:
Reduced the size of the font to make sure it does not get cut off.


TorsteinHonsi (approved):
Thanks!

PR #16736, opened 2021-12-01
Fixed legend kbd issues #16520, #16545, #16467.

oysteinmoseng:
Fixed legend keyboard navigation issues #16520, #16545, #16467.

PR #16825, opened 2021-12-23
Fixed #16818, issue with kbd nav and legend.labelFormatter.

### LINKED ISSUE #16818: Defining a legend.labelFormatter function breaks legend keyboard navigation
#### Expected behaviour
I'd like to be able to utilize `legend.labelFormatter` to customize the format of the legend labels, and maintain keyboard functionality in the legend—arrow keys to navigate between series and spacebar/return/enter to toggle each series.

#### Actual behaviour
Upon defining a formatter function—it doesn't matter how complex the function is—expected keyboard functionality in the legend is broken.
- Define a formatter function
- Use the keyboard to navigate to the legend
- Use the arrow keys to navigate between series
- Use the spacebar to toggle visibility of any series

...I'm no longer able to use the arrow keys to navigate between series or the spacebar to toggle visibilty.

https://jsfiddle.net/longmatthewh/qby4k80g/16/

#### Product version
Highcharts JS v9.3.2 w/ the Accessibility module

#### Affected browser(s)
At least Chrome

oysteinmoseng:
Fixed #16818, issue with keyboard navigation and `legend.labelFormatter`.
___
Attempted fixed in #16736, but clearly needed a more specific and robust fix for this one.

### DIFF
--- a/ts/Accessibility/Components/LegendComponent.ts
+++ b/ts/Accessibility/Components/LegendComponent.ts
@@ -338,7 +338,7 @@ class LegendComponent extends AccessibilityComponent {
             this.updateLegendTitle();
 
             if (shouldRestoreFocus) {
-                (focusedElement as HTMLElement).focus();
+                this.chart.highlightLegendItem(this.highlightedLegendItemIx);
             }
             return true;
         }


TorsteinHonsi (approved):
Thanks!

longmatthewh:
@TorsteinHonsi ⬆️

TorsteinHonsi:
@longmatthewh We're targeting a release within 2-3 weeks.

@oysteinmoseng Do we have a drop-in fix they can use until then?

oysteinmoseng:
This one is tricky to drop-in fix, unfortunately, so if a fix is required before the release I have no better suggestions than editing the `accessibility.src.js` file directly or getting the release candidate from http://github.highcharts.com/modules/accessibility.js.

Note that getting it from github.highcharts.com will include other changes as well, and could cause compatibility issues with other files in some cases, so you might want to grab the other files from github.highcharts.com as well in that case.

PR #16834, opened 2021-12-28
bugfix/16169-centerInCategory-reverses-stack 

### LINKED ISSUE #16169: centerInCategory setting reverses stack groups order
#### Expected behaviour
`centerInCategory: true` in combination with `reversedStacks: false` shouldn't affect order of different stacks.

#### Actual behaviour
`centerInCategory: true` affects order of different stacks (also labels are misplaced).

https://jsfiddle.net/gv4skudb/
(_series 1_ and _series 2_ stack group must be first)
#### Product version
8.1.1 and later

#### Affected browser(s)
All

karolkolodziej:
Fixed #16169, stack order was affected when `centerInCategory` enabled and `reversedStacks` set to false.

### DIFF
--- a/samples/unit-tests/series/stacking/demo.js
+++ b/samples/unit-tests/series/stacking/demo.js
@@ -699,4 +699,46 @@ QUnit.test('Date objects as X values, column', function (assert) {
             'Series 3 - Point should start from value=1 (#4024)'
         );
     });
+
+    QUnit.test(
+        'The centerInCategory, reversedStacks order of stacks, #16169.',
+        function (assert) {
+            const chart = Highcharts.chart('container', {
+                    chart: {
+                        type: "column"
+                    },
+                    plotOptions: {
+                        column: {
+                            stacking: "normal",
+                            centerInCategory: true
+                        }
+                    },
+                    yAxis: {
+                        reversedStacks: false,
+                        stackLabels: {
+                            enabled: true
+                        }
+                    },
+                    series: [{
+                        data: [5],
+                        id: 0
+                    }, {
+                        data: [2],
+                        id: 1
+                    }, {
+                        data: [4],
+                        stack: "1"
+                    }, {
+                        data: [1],
+                        stack: "1"
+                    }]
+                }),
+                series = chart.series;
+
+            assert.ok(
+                series[0].points[0].barX < series[2].points[0].barX,
+                `Enabling centerInCategory and setting reversedStacks to false
+                should not affect the stack order.`
+            );
+        });
 }());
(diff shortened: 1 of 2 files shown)


samples/unit-tests/series/stacking/demo.js
```diff
@@ -699,4 +699,46 @@ QUnit.test('Date objects as X values, column', function (assert) {
             'Series 3 - Point should start from value=1 (#4024)'
         );
     });
+
+    QUnit.test(
```
raf18seb, line 703: No chance to reuse the existing test?

samples/unit-tests/series/stacking/demo.js
```diff
@@ -699,4 +699,46 @@ QUnit.test('Date objects as X values, column', function (assert) {
...
+                        id: 0
+                    }, {
+                        data: [2],
+                        id: 1
+                    }, {
+                        data: [4],
+                        stack: "1"
+                    }, {
+                        data: [1],
+                        stack: "1"
```
raf18seb, line 733: Please use single quote when writing JavaScript code

samples/unit-tests/series/stacking/demo.js
```diff
@@ -699,4 +699,46 @@ QUnit.test('Date objects as X values, column', function (assert) {
...
+                        stackLabels: {
+                            enabled: true
+                        }
+                    },
+                    series: [{
+                        data: [5],
+                        id: 0
+                    }, {
+                        data: [2],
+                        id: 1
```
raf18seb, line 727: Do we need `id` here?

ts/Series/Column/ColumnSeries.ts
```diff
@@ -822,7 +823,8 @@ class ColumnSeries extends Series {
                                     indexInCategory = totalInCategory;
                                 }
                                 if (stackItem.hasValidPoints) {
-                                    totalInCategory++;
+                                    reversedStacks ?
```
raf18seb, line 826: What do you think about adding a comment with a ticket number here so the ticket can be easier to find in case of any problems in the future?

raf18seb (changes requested):
Good job! Just a few small suggestions, see the comments below

samples/unit-tests/series/stacking/demo.js
karolkolodziej: The config to recreate the issue is quite specific and I was not able to reuse any of the existing tests.

raf18seb (approved):
Thanks!

TorsteinHonsi (approved):
Thanks!

PR #16843, opened 2021-12-30
bugfix/16820-drilldown-tooltip-mouseover 

### LINKED ISSUE #16820: Tooltip error when adding drilldown series on point mouseover
Hover on point - drilldown series is being added and then an error is thrown into the console.

https://jsfiddle.net/BlackLabel/e5kpnro4/

#### Product version
Highcharts 9.3.2

#### Workaround
https://jsfiddle.net/BlackLabel/5kxucwpe/

`Internal note:` tagging as `Good for beginners` because the first solution (workaround above) is very simple to implement. However it might be worth investigating this issue deeper.

karolkolodziej:
Fixed #16820, adding drilldown series on point mouseover threw an error.

### DIFF
--- a/samples/unit-tests/drilldown/drilldown-general/demo.js
+++ b/samples/unit-tests/drilldown/drilldown-general/demo.js
@@ -216,6 +216,25 @@ QUnit.test('Drilldown methods', function (assert) {
         2,
         '#15771: Point should be at correct index in ddPoints'
     );
+
+    chart.series[0].update({
+        point: {
+            events: {
+                mouseOver() {
+                    this.series.chart.addSeriesAsDrilldown(this, {
+                        data: [1, 2, 3, 4]
+                    });
+                }
+            }
+        },
+        data: [2, 5]
+    });
+    Highcharts.fireEvent(chart.series[0].points[0].graphic.element, 'mouseover');
+    assert.ok(
+        true,
+        `Adding drill down series through mouseover should not
+        generate errors in console, #16820.`
+    );
 });
 
 QUnit.test('Chart type update after drilldown', function (assert) {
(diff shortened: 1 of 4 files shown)


samples/unit-tests/drilldown/drilldown-general/demo.js
```diff
@@ -216,6 +216,25 @@ QUnit.test('Drilldown methods', function (assert) {
...
+                mouseOver() {
+                    this.series.chart.addSeriesAsDrilldown(this, {
+                        data: [1, 2, 3, 4]
+                    });
+                }
+            }
+        },
+        data: [2, 5]
+    });
+    Highcharts.fireEvent(chart.series[0].points[0].graphic.element, 'mouseover');
```
raf18seb, line 232: Just a note: you can fire `onMouseOver` instead
```js
chart.series[0].points[0].onMouseOver();
```

raf18seb (commented):
@TorsteinHonsi this solution is OK, but I'm wondering whether the `tooltip.refresh()` (or even `point.onMouseOver()`) method should be fired in the first place. In this case, we hover over the point to add series as drilldown, but not because we want to display the tooltip. Maybe we could stop mouseOver event from firing further actions by adding something similar to `e.preventDefault` or `return false` or something else in the `point.events.mouseOver` callback function?

Also, small info in the comment below.

ts/Core/Tooltip.ts
```diff
@@ -1042,7 +1042,7 @@ class Tooltip {
             styledMode = chart.styledMode;
         let textConfig = {} as Tooltip.FormatterContextObject;
 
-        if (!options.enabled) {
+        if (!options.enabled || !splat(pointOrPoints)[0].series) { // #16820
```
TorsteinHonsi, line 1045: We already do `splat(pointOrPoints)` on line 1037, and refer to item 0 on line 1038, so we can simplify this line to `if (!options.enabled || !point.series) { // #16820`

TorsteinHonsi (changes requested):
Looks good to me except a small optimization.

raf18seb (approved):
Perfect 👍

TorsteinHonsi (approved):

PR #16847, opened 2022-01-03
New option: AST.bypassHTMLFiltering

### LINKED ISSUE #15345: Add option to disable v9 HTML filtering
# Description of the feature

Hi, please could you add an option that allows the developer to force-disable the strict HTML filtering that is applied in v9? The current behaviour is quite disruptive to my use case.

I have written a wrapper library around Highcharts that exposes its functionality in declarative Angular syntax. For example, to display a line chart with a tooltip the user can write something like

```
<chart zoomType="x">
    <chart-title>
        <h3 class="subheading-2">
            Critical response to <em class="em">Star Wars</em> movies over time
        </h3>
    </chart-title>

    <chart-x-axis type="datetime">
        <chart-axis-date-labels rotation="-45" format="D MMM 'YY"></chart-axis-date-labels>
        <chart-axis-crosshair></chart-axis-crosshair>
    </chart-x-axis>

    <chart-y-axis>
        <chart-axis-number-labels format="0%"></chart-axis-number-labels>
        <chart-axis-plot-line
            value="0"
            className="plotline-y-0"
        ></chart-axis-plot-line>
    </chart-y-axis>

    <chart-line-series
        *ngIf="hasDataLoaded"
        name="MetaCritic score"
        color="purple"
        [data]="metacriticData"
    >
        <chart-series-number-data-labels format="0%"></chart-series-number-data-labels>
        <chart-series-point-markers symbol="square" radius="3"></chart-series-point-markers>
    </chart-line-series>

    <chart-tooltip>
        <div *chartTooltipContent="let context" class="_tooltip">
            <h3 class="subheading-2">{{ context.points[0].point.name }}</h3>
            <h4 class="caption"><mat-icon>date</mat-icon> {{ context.x | moment: 'DD MMMM YYYY' }}</h4>
            <some-custom-angular-component></some-custom-angular-component>
            <ul>
                <li *ngFor="let point of context.points">
                    <strong class="strong">{{ point.series.name }}</strong
                    >:
                    <span class="code">{{ point.y | numeral: '0%' }}</span>
                </li>
            </ul>
        </div>
    </chart-tooltip>

    <chart-legend></chart-legend>
</chart>
```

The tooltip's content is rendered within Angular in an off-screen outlet and then the `innerHTML` of the outlet tag is passed into Highcharts' tooltip options. This hides any Highcharts-specific plumbing for the user. Chart/series titles and annotations are implemented in a similar way.

Under v9, this no longer works because any Angular components that the user includes in their dynamic content (such as `some-custom-angular-component` and `mat-icon` above) will be filtered out.

The recommended approach is to add these tag names to Highcharts' AST whitelist up-front, but this is really undesirable as a developer experience for consumers of my library (plus it is a Highcharts implementation detail that the library would be leaking).

Additionally, Angular applies unique host-scoping tags to every component context - these look like `_ngcontent-whp-c321` or similar - and there is no way to whitelist these or disable Highcharts' warnings for them.

All I am requesting is that an option be added to revert to the previous (<=v8) HTML behaviour. The option can have a very specific name (like Angular's [DomSanitizer#bypassSecurityTrustHtml](https://angular.io/api/platform-browser/DomSanitizer#bypassSecurityTrustHtml)) and be false by default. 

Highcharts

N/A - see Highcharts 8.x

----

raf18seb:
Added new option `Highcharts.AST.bypassHTMLFiltering` for disabling AST filtering.

Closes #15345.

TODO:
- [x] I'll simplify the test and add another attribute test. Converted to draft till then

### DIFF
--- a/samples/unit-tests/svgrenderer/xss/demo.js
+++ b/samples/unit-tests/svgrenderer/xss/demo.js
@@ -52,7 +52,7 @@ QUnit.module('XSS', function () {
             // JavaScript directive
             text = ren
                 .text(
-                    'This is a link to <a href="javascript:alert(\'XSS\')">an simple JS directive</a>, <br>' +
+                    'This is a link to <a href="javascript:alert(\'XSS\')">a simple JS directive</a>, <br>' +
                     'an image <IMG SRC="javascript:alert(\'XSS\');">, <br>' +
                     'an unquoted image <IMG SRC=javascript:alert(\'XSS\')>, <br>' +
                     'a case insensitive attack vector <IMG SRC=JaVaScRiPt:alert(\'XSS\')>, <br>' +
@@ -207,24 +207,41 @@ QUnit.module('XSS', function () {
                 'Tag tricks, scripts should be stripped out from DOM'
             );
 
-            // Disable AST filtering to allow black-listed tags
-            Highcharts.AST.bypassHTMLFiltering = true;
+            if (useHTML) {
+                // Disable AST filtering to allow black-listed tags
+                Highcharts.AST.bypassHTMLFiltering = true;
 
-            text = ren.text(
-                '<customTag></customTag>',
-                x,
-                y += lineHeight,
-                useHTML
-            )
-            .add();
+                text = ren.text(
+                    [
+                        '<customTag></customTag>',
+                        '<span id="elemWithCustomAttribute" customAttribute="testValue"></span>',
+                        '<a href="javascript:alert(\'XSS\')">a simple JS directive</a>',
+                    ].join(',<br>'),
+                    x,
+                    y += lineHeight,
+                    useHTML
+                )
+                .add();
 
-            assert[useHTML ? 'ok' : 'notOk'](
-                document.querySelector('customTag'),
-                'AST disabled: black-listed tag should ' +
-                    (useHTML ? '' : 'not') + ' be added, #15345'
-            );
+                assert.ok(
+                    document.querySelector('customTag'),
+                    'AST disabled: custom tag should be added, #15345'
+                );
 
-            Highcharts.AST.bypassHTMLFiltering = false;
+                assert.strictEqual(
+                    document.getElementById('elemWithCustomAttribute')
+                        .getAttribute('customAttribute'),
+                    'testValue',
+                    'AST disabled: custom attribute should be added, #15345'
+                );
+
+                assert.ok(
+                    /javascript/i.test(text.element.outerHTML),
+                    'AST disabled: custom JavaScript directive should be allowed, #15345'
+                );
+
+                Highcharts.AST.bypassHTMLFiltering = false;
+            }
 
             // Reset
             window.alert = realAlert;
--- a/samples/unit-tests/svgrenderer/xss/demo.js
+++ b/samples/unit-tests/svgrenderer/xss/demo.js
@@ -207,6 +207,25 @@ QUnit.module('XSS', function () {
                 'Tag tricks, scripts should be stripped out from DOM'
             );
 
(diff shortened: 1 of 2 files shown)


raf18seb:
I'm testing here tag only, not sure how can I test the attribute.
Also, is a test for `useHTML: false` necessary here?

TorsteinHonsi:
> I'm testing here tag only, not sure how can I test the attribute.

I think you can just add a custom attribute to the pseudo-HTML. Also test references. By default, evil references like `file://` are not allowed. So you can test something like `href="foo://bar"` and verify that it passes through.

> Also, is a test for useHTML: false necessary here?

I don't think so. It will pass through the same filtering if `useHTML` is false, but it will be modified later on, so testing would be harder.

raf18seb:
@TorsteinHonsi ready for review

TorsteinHonsi:
@raf18seb @pawelfus - two discussion points:

* We're consistently using the term allow-listing for what we're doing in the AST. And the lists are called `allowedTags` etc. Maybe a better name for the new setting is `bypassAllowLists`? And strictly speaking, `bypassHTMLFiltering` is wrong, because we also apply apply this on SVG structures...
* What about the `filterUserAttributes` function? We're calling this internally for things like button attributes that come directly from the chart config. The way things are set up now, filtering is applied to those even when `bypassHTMLFiltering` is true. Which in a way makes sense, since it is not HTML per se. Nor SVG, just attributes 🤔 . And now I am contradicting my first point. If we leave the functionality like it is now, maybe `bypassMarkupFiltering` would be the appropriate name, since it's bypassing filtering of HTML and SVG markup, but not attributes coming from config. 

Thoughts?

raf18seb:
`bypassMarkupFiltering` is closest to the truth (true meaning), however, this might not be as easy to understand as `bypassHTMLFiltering`.

In other words: when I read `bypassHTMLFiltering`, I know it's about filtering HTML even it's not semantically true in 100% of cases.
But when I was reading `bypassMarkupFiltering` for the first time, I needed a while to think what the `Markup` actually is :D

I have a third proposal: `bypassHTMLMarkupFiltering` - but it's still not related to SVG structures.

As I don't see a perfect naming, I'm good with all of the above.

Maybe @pawelfus ?

pawelfus:
How about this?

Since we have `allowedTags`, `allowedAttributes`, and `allowerReferences`, would it make sense to set three different options? For example `bypassFilteringTags`, `bypassFilteringAttributes` and `bypassFilteringReferences`. The [feature request](https://github.com/highcharts/highcharts/issues/15345) mentions only tags and attributes, nothing about refs.

Alternatively, we can add support for `*` to each of the arrays. Adding `*` (`Highcharts.AST.allowedTags.push('*')`) will simply allow all possible tags/attributes/references to pass on. Internally, we can check for `indexOf('*')` and store this in a variable, e.g. `bypassTags` etc. to speed up process.

Personally, I prefer `bypassFiltering*`, as the `*` can be easily missed when debugging or checking for a security issues.

TorsteinHonsi:
But I sense a bit of a scope creep then.. The initial idea was just to bypass the filtering of incoming HTML from the options. I suggest we stick with that for now, then we can increase the scope later if needed.

TorsteinHonsi (approved):

joeveiga:
Hi there 👋. Is this option supposed to be working with custom `tooltip.formatter` function? I just upgraded from version 8 to 10 and my custom tool tips stopped rendering. Tried setting this flag but it didn't work for me.

raf18seb:
Answered [here](https://github.com/highcharts/highcharts/issues/17089)

PR #16852, opened 2022-01-05
Fixed doclet

pawelfus:
Just a typo in doclets. The current one creates in `tree.json`: `extends: series,plotOptions.lollipop,`, which splitting by comma will cause searching for `""` (an empty string) as a parent.


TorsteinHonsi (approved):
Thanks!

PR #16853, opened 2022-01-05
bugfix/11115-dependency-wheel-labels 

### LINKED ISSUE #11115: Dependency wheel node labels not fully visible
#### Expected behaviour
Node labels on the Dependency Wheel are readable.

#### Actual behaviour
It's difficult (impossible?) to use labels that are long or belong to sectors that are very narrow.

When using textPath, long labels are cut off:
https://jsfiddle.net/7jtfacyk/

Screen Shot 2019-06-20 at 2 28 41 PM

Without textPath, full labels are displayed, but can overlap if the sectors are small:
https://jsfiddle.net/7bmnd1fg/

Screen Shot 2019-06-20 at 2 28 12 PM

Since this plot type is so new, there is only a single trivial example out there. I can't find any way to customize the label positions like I have for other plot types.

#### Product version
Highcharts 7.1.2

karolkolodziej:
Fixed #11115, `textPath` for dependency wheel.

- [x] Create a separate module `CircularDataLabels` for series like `sunburst` and `dependencywheel`
- [x] Add support for `rotationMode` in `dependencywheel`
- [x] tests
- [x] API
- [x] demos 
- [x]  logic that automatically aligns the labels
- [ ] Check CI
- [ ] Resolve conflicts
- [ ] In [datalabels-rotationmode-auto](https://highcharts.github.io/highcharts-utils/samples/#gh/b4d352c4b5/sample/highcharts/series-dependencywheel/datalabels-rotationmode-auto), the Mexico label is hidden until hovered. (The South Africa label has an issue with text shadow, I'll deal with that).
- [ ] Improve `rotationMode: 'auto'`
- [ ] Improve `rotationMode: 'circular'` 
- [ ] Improve `rotationMode: 'perperndicular'`


raf18seb (approved):
👍

ts/Series/DependencyWheel/DependencyWheelSeries.ts
```diff
@@ -285,8 +295,9 @@ class DependencyWheelSeries extends SankeySeries {
                     r: r,
                     innerR: innerR,
                     start: start,
-                    end: end
-                };
+                    end: end,
+                    radius: r
```
raf18seb, line 299: `r` is already there

ts/Series/DependencyWheel/DependencyWheelSeries.ts
```diff
@@ -295,8 +306,25 @@ class DependencyWheelSeries extends SankeySeries {
...
+                if (
+                    node.shapeArgs &&
+                    node.shapeArgs.innerR &&
+                    node.shapeArgs.r
+                ) {
+                    node.innerArcLength = radians * node.shapeArgs.innerR;
+                    node.outerArcLength = radians * node.shapeArgs.r;
+                }
+
+                node.dlOptions = (dependecyWheel).getDlOptions({
```
raf18seb, line 319: Unnecessary bracket

ts/Series/DependencyWheel/DependencyWheelSeries.ts
```diff
@@ -295,8 +306,25 @@ class DependencyWheelSeries extends SankeySeries {
...
+                    point: node,
+                    level: void 0,
```
raf18seb, line 321: Is it necessary to pass a property with `undefined` value?

ts/Series/CircularDataLabels.ts
```diff
@@ -0,0 +1,252 @@
...
+        (circularSeries as any).prototype.getDlOptions = getDlOptions;
+    }
+
+    function getDlOptions(
+        params: CicrularDlOptionsParams
+    ): any {
+        // Set options to new object to avoid problems with scope
+        let point = params.point,
+            shape: Partial<SunburstSeries.NodeValuesObject> =
+                isObject(params.shapeArgs) ? params.shapeArgs : {},
```
raf18seb, line 84: Is it possible that `shapeArgs` is not an object? There's probably something wrong with TS

raf18seb (changes requested):

ts/Series/DependencyWheel/DependencyWheelSeries.ts
```diff
@@ -295,8 +304,24 @@ class DependencyWheelSeries extends SankeySeries {
...
+
+                node.dlOptions = dependecyWheel.getDlOptions({
+                    point: node,
+                    optionsPoint: node.options,
+                    shapeArgs: node.shapeArgs
+                });
+
                 // Draw the links from this node
-                node.linksFrom.forEach(function (point): void {
+                node.linksFrom.forEach(function (point: any): void {
```
TorsteinHonsi, line 324: Why a new `any`? Perhaps some types need updating?

ts/Series/DependencyWheel/DependencyWheelSeries.ts
```diff
@@ -286,7 +295,7 @@ class DependencyWheelSeries extends SankeySeries {
                     innerR: innerR,
                     start: start,
                     end: end
-                };
+                } as any;
```
TorsteinHonsi, line 298: Why a new `any`?

ts/Series/DependencyWheel/DependencyWheelSeries.ts
```diff
@@ -254,8 +262,8 @@ class DependencyWheelSeries extends SankeySeries {
      * functions instead of the whole translate function.
      */
     public translate(): void {
-
-        const options = this.options,
+        const dependecyWheel = this,
```
TorsteinHonsi, line 265: Why do we need a local reference? And if we really need one, the convention is to use `series` in this case.

TorsteinHonsi (changes requested):
Thanks, nice work on the composition. This really makes sense the way it is implemented, and will make it easier to adapt to PieSeries in the future.

One thing we may be missing now is the kind of logic that automatically aligns the labels. Currently this is implemented in pie series, so labels on the left side are right aligned and vice versa. X axis labels in a polar chart also has this feature. In the visual demo with rotation mode perpendicular, the labels are centered aligned. This may be out of scope though, but I do feel it is the last missing piece.

I also took the liberty to move the sample, we've been cramming too much stuff into the `plotoptions` folder in the past.

TorsteinHonsi:
@raf18seb Do you want to have a go with auto-alignment? I imagine that we set the default `align` to `undefined`, which results in the left-side labels being right-aligned and the right-side labels being left-aligned.

For inspiration, similar logic can be found for polar charts. Though that logic actually writes the alignment back to `options.align`, which I think is unfortunate. The options should not be mutated.

https://github.com/highcharts/highcharts/blob/5f27de015659dc2f5ac32911e1e1b05d2962906c/ts/Extensions/Polar.ts#L966-L975

karolkolodziej:
@TorsteinHonsi could you please take a look?☝️

TorsteinHonsi:
@karolkolodziej We need more samples and docs. For example, the latest commit apparently should apply automatic label alignment when `rotationMode` and `align` are falsy, but there's no demo and I can't get it to work.

Also, the `rotationMode: 'auto'` setting in sunburst applies the best fit. It prefers a parallel layout, but falls back to perpendicular if there is no space. This doesn't seem to work with the dependency wheel?

karolkolodziej:
@TorsteinHonsi 
I fixed the auto alignment issue, but I'm not sure if I should add a separate demo for each alignment option?

ts/Series/DependencyWheel/DependencyWheelSeries.ts
```diff
@@ -469,6 +519,24 @@ export default DependencyWheelSeries;
...
+ * The `perpendicular` option works similar to `auto` and position labels
+ * perpendicularly to the circumference. The `parallel` option fits the labels
+ * inside of the node.
+ *
+ * The `rotation` and `textPath` option takes precedence over `rotationMode`.
+ *
+ * @type       {string}
+ * @sample {highcharts} highcharts/plotoptions/dependencywheel-datalabels-rotationmode-perpendicular/
+ *         Perpendicular rotation mode
+ * @validvalue ["auto", "perpendicular", "parallel"]
```
TorsteinHonsi, line 534: Shouldn't we support 'circular' like the sunburst?

ts/Series/DependencyWheel/DependencyWheelSeries.ts
```diff
@@ -469,6 +519,24 @@ export default DependencyWheelSeries;
...
+ * Decides how the data label will be rotated relative to the perimeter of the
+ * dependencywheel. Valid values are `auto`, `parallel` and `perpendicular`.
+ * The `perpendicular` option works similar to `auto` and position labels
+ * perpendicularly to the circumference. The `parallel` option fits the labels
+ * inside of the node.
+ *
+ * The `rotation` and `textPath` option takes precedence over `rotationMode`.
+ *
+ * @type       {string}
+ * @sample {highcharts} highcharts/plotoptions/dependencywheel-datalabels-rotationmode-perpendicular/
```
TorsteinHonsi, line 532: This sample reference is wrong, the sample has moved.

TorsteinHonsi (changes requested):
@karolkolodziej We need at least enough visual samples that we can demonstrate how this works. Right now there's only a sample for `rotationMode: 'perpendicular' `, but:
* The labels are centered, not auto-aligned. I would expect, at least if I set `align: 'auto'`, that the labels be aligned like in a pie chart. And then we need a sample for that. 
* When I change the `rotationMode` to `circular`, it still renders perpendicular.
* When I change the `rotationMode` to `auto`, all labels are still perpendicular. I would expect a mix between parallel and perpendicular labels.

TorsteinHonsi:
Thanks! Remaining issues:
* In [datalabels-rotationmode-auto](https://highcharts.github.io/highcharts-utils/samples/#gh/b4d352c4b5/sample/highcharts/series-dependencywheel/datalabels-rotationmode-auto), the Mexico label is hidden until hovered. (The South Africa label has an issue with text shadow, I'll deal with that).
* Re-iteration of the first point above: in the [perpendicular demo](https://highcharts.github.io/highcharts-utils/samples/#gh/b4d352c4b5/sample/highcharts/series-dependencywheel/datalabels-rotationmode-perpendicular), we need to show how to make the labels radiate out from the nodes. I expect `align: 'auto'` to do that. Instead, all labels are left-aligned: https://jsfiddle.net/highcharts/spcqvua5/. I would expect that labels on the left side of the wheel are right-aligned. We also need to implement [distance](https://api.highcharts.com/highcharts/series.sunburst.dataLabels.distance) for these labels. Distance currently works when `rotationMode` is circular, but not for perpendicular.

TorsteinHonsi:
@karolkolodziej Pending continued work. Maybe we should put this back in draft and add the remaining tasks as check boxes?

PR #16854, opened 2022-01-05
Fixed round errors on windows.

pawelfus:
As in the title. The difference for windows-based machines is 1px


raf18seb (approved):
All tests pass on Windows.
1.1 rounding is perfect (1.0 is too small, it needs to be higher than 1).

TorsteinHonsi (approved):
Thanks!

PR #16856, opened 2022-01-05
Docs: Fixed types of PointLabelObject and TooltipFormatterContextObject.

### LINKED ISSUE #16565: Bug (Typescript): TooltipFormatterContextObject.x is declared as number but is assigned a string
#### Expected behaviour
TooltipFormatterContextObject.x should be declared string because that is the type assigned to it.
I would be happy to contribute a fix that improves type safety.

#### Actual behaviour
TooltipFormatterContextObject.x is declared as number in the Typescript declarations and in the documentation.

TooltipFormatterContextObject.x is assigned Point.PointLabelObject.x which is assigned Point.category* which is declared as string.

(* If a Point has no category then it is assigned the number Point.x, but that should be converted toString to be consistent with the type declaration of Point.category)

Tooltip.ts:
```
            textConfig = {
                x: point.category, // string!
                y: point.y
            } as any;
            textConfig.points = pointConfig as any;

        // single point tooltip
        } else {
            textConfig = point.getLabelConfig() as any;
        }
        this.len = pointConfig.length; // #6128
        const text: (boolean|string) = (formatter as any).call(textConfig, tooltip);
```
```
namespace Tooltip {
    export interface FormatterCallbackFunction {
        (
            this: FormatterContextObject,
            tooltip: Tooltip
        ): (false|string|Array<string>);
    }
    export interface FormatterContextObject {
        color: ColorType;
        colorIndex?: number;
        key: number;
        percentage?: number;
        point: Point;
        points?: Array<FormatterContextObject>;
        series: Series;
        total?: number;
        x: number;
        y: number;
    }
```
Point.ts:
```
    public getLabelConfig(): Point.PointLabelObject {
        return {
            x: this.category, // string!
            y: this.y,
```

In the following fiddle, most tooltips are broken because this.x does not have a toFixed method, except for Feb because its category is undefined.
https://jsfiddle.net/96mLwf82/1

#### Product version
Highcharts 9.3

#### Affected browser(s)
n/a

bre1470:
- Fixed #16565, Synchronized typing of `TooltipFormatterContextObject`with `PointLabelObject`.
- Fixed typing of `PointLabelObject`.
- Fixed parameter type of `Highcharts.clearTimeout`.

### DIFF
--- a/ts/Core/Series/Series.ts
+++ b/ts/Core/Series/Series.ts
@@ -2302,9 +2302,10 @@ class Series {
             // some API data
             point.category = (
                 categories &&
-                typeof (categories as any)[point.x as any] !== 'undefined' ?
-                    (categories as any)[point.x as any] :
-                    point.x
+                point.x !== null &&
+                typeof categories[point.x] !== 'undefined' ?
+                    categories[point.x] :
+                    point.x as any
             );
 
             // Determine auto enabling of markers (#3635, #5099)
--- a/ts/Core/Series/Series.ts
+++ b/ts/Core/Series/Series.ts
@@ -1600,7 +1600,7 @@ class Series {
             xExtremes = xAxis.getExtremes();
             min = xExtremes.min;
             max = xExtremes.max;
-            updatingNames = xAxis.categories && !xAxis.names.length;
+            updatingNames = !!(xAxis.categories && !xAxis.names.length);
         }
 
         // optionally filter out points outside the plot area
@@ -2300,11 +2300,9 @@ class Series {
             );
 
             // some API data
-            point.category = (
-                categories &&
-                typeof categories[point.x] !== 'undefined' ?
-                    categories[point.x] :
-                    point.x as any
+            point.category = pick(
+                categories && categories[point.x],
+                point.x as any
             );
 
             // Determine auto enabling of markers (#3635, #5099)
--- a/ts/Core/Series/Series.ts
+++ b/ts/Core/Series/Series.ts
@@ -2302,7 +2302,6 @@ class Series {
             // some API data
             point.category = (
                 categories &&
-                point.x !== null &&
                 typeof categories[point.x] !== 'undefined' ?
                     categories[point.x] :
                     point.x as any
--- a/ts/Core/Axis/Axis.ts
+++ b/ts/Core/Axis/Axis.ts
@@ -187,7 +187,7 @@ class Axis {
     public axisTitle?: SVGElement;
     public axisTitleMargin?: number;
     public bottom: number = void 0 as any;
-    public categories: Array<string> = void 0 as any;
+    public categories?: Array<string>;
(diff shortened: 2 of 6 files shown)


ts/Core/Series/Series.ts
```diff
@@ -2302,9 +2302,10 @@ class Series {
             // some API data
             point.category = (
                 categories &&
-                typeof (categories as any)[point.x as any] !== 'undefined' ?
-                    (categories as any)[point.x as any] :
-                    point.x
+                point.x !== null &&
```
TorsteinHonsi, line 2305: Are there any actual cases where `point.x` is `null`, or does this test come from the typing? I suspect it mey be a relic of the past, and that we instead should do `isNumber(point.x)` to rule out `undefined`.

TorsteinHonsi (changes requested):

ts/Core/Series/Series.ts
bre1470: This type is coming from MockPoint: https://github.com/highcharts/highcharts/blob/master/ts/Extensions/Annotations/MockPoint.ts#L494-L508

ts/Core/Series/Series.ts
```diff
@@ -2302,9 +2302,9 @@ class Series {
             // some API data
             point.category = (
                 categories &&
-                typeof (categories as any)[point.x as any] !== 'undefined' ?
-                    (categories as any)[point.x as any] :
-                    point.x
+                typeof categories[point.x] !== 'undefined' ?
+                    categories[point.x] :
+                    point.x as any
```
TorsteinHonsi, line 2307: Hmm... I think now it can be further simplified by 'pick': 
```ts
            point.category = (
                categories &&
                pick(categories[point.x], point.x)
            );
```

TorsteinHonsi (changes requested):

ts/Core/Series/Series.ts
bre1470: Neither works:
```ts
point.category = categories && pick(categories[point.x], point.x);
point.category = pick(categories && categories[point.x], point.x);
```
> 3 FAILED, 1910 SUCCESS

TorsteinHonsi: Ok... Then I suspect something must be wrong with the typing. I'll check.

TorsteinHonsi: Confirmed that the typing of `axis.categories` is wrong. It can be a boolean, so in the failing cases it was `false`. Pushing a commit now.

ts/Core/Series/Series.ts
BistroStu: I think the following is what was intended:

```ts 
point.category = pick(categories && categories[point.x as number], point.x);
```
But to make category true to its type (string) you would need:
```ts 
point.category = pick(categories && categories[point.x as number], String(point.x));
```

PR #16859, opened 2022-01-05
Website/accessible venn icons

nasin:
Hi, I updated the icons and some of the colors to better match with the new design.


TorsteinHonsi (approved):
Thanks!

PR #16860, opened 2022-01-06
Docs: Added Point.plotX/Y to API docs.

### LINKED ISSUE #16822: TooltipFormatterContextObject has wrong type for "point"
#### Expected behaviour
The TooltipFormatterContextObject should type the point property as TooltipPositionerPointObject not Point. The plotY and plotX properties are available on TooltipPositionerPointObject but not on Point. plotY and plotX are available when you log this property.

#### Actual behaviour
TooltipFormatterContextObject.point says there's no plotY/X but there is. 

#### Product version

Highcharts 9.3.2
#### Affected browser(s)
N/A

bre1470:
Fixed #16822, Added doclets for `Point.plotX` and `Point.plotY` to make both properties visible in API documentation and TypeScript definitions.

### DIFF
--- a/ts/Core/Series/Series.ts
+++ b/ts/Core/Series/Series.ts
@@ -2188,8 +2188,10 @@ class Series {
             }
 
             /**
-             * This is the translated X value for the point. Depending
-             * on the series type this value might not be defined.
+             * The translated X value for the point in terms of pixels. Relative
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #16869, opened 2022-01-11
Fixed issue with data module updates and dataGrouping options.

oysteinmoseng:
Fixed issue with data module updates when `dataGrouping` options were present.
___
The issue was that when updating data options for a chart with Google Spreadsheet link or `data.columns`, Highcharts error 14 was thrown in some cases. This happened because the first data row was corrupted and changed to string if `firstRowAsNames` was enabled, even if the first row had previously been removed and `column.name` set.

Presumably the only reason `dataGrouping` is involved here is the logic in `series.update` that looks for `hasOptionChanged('dataGrouping')` - I can't see any other reason it should matter. It does not matter which dataGrouping options are added, even a chart like the below fails:

```js
    const chart = Highcharts.chart('container', {
        plotOptions: {
            series: {
                dataGrouping: {
                    whatever: 'dummyoption'
                }
            }
        },
        data: {
            // ...
        }
    });
```

Fixed by not changing the first data row to string if `column.name` is already set.

### DIFF
--- a/samples/unit-tests/data/general/demo.js
+++ b/samples/unit-tests/data/general/demo.js
@@ -254,7 +254,7 @@ QUnit.test('Data config on updates and setOptions', function (assert) {
         2,
         'Global data options should be merged with the chart options (#16568).'
     );
-    
+
     // Clear the default options for other tests
     delete Highcharts.defaultOptions.data;
 });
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #16871, opened 2022-01-11
Fixing CSP warnings

### LINKED ISSUE #6884: Errors because of Content Security Policy
#### Expected behaviour
no csp violation

#### Actual behaviour
When I set csp rule to `style-src 'self'` i receive a bunch of errors in console

The issue is  inline styling in `buildText` and `css` functions. 

jsfiddle has no CPS. Demo has no sense.

#### Affected browser(s)
I think it is all modern browsers. Checked in Chrome and Safari.

Is there a possibility to avoid a problem on my side?

TorsteinHonsi:
Fixed #6884, console errors related to style attributes because of Content Security Policy.

### To do
 - [x] Go over new code and look for optimization
 - [x] Check Firefox `scrollablePlotArea` case again
 - [x] Check legacy browsers manually

### DIFF
--- a/ts/Core/Renderer/SVG/SVGElement.ts
+++ b/ts/Core/Renderer/SVG/SVGElement.ts
@@ -1076,9 +1076,9 @@ class SVGElement implements SVGElementLike {
 
         // Filter out existing styles to increase performance (#2640)
         if (oldStyles) {
-            objectEach(styles, function (style, n): void {
-                if (oldStyles && oldStyles[n] !== style) {
-                    (newStyles as any)[n] = style;
+            objectEach(styles, function (value, n: keyof CSSObject): void {
+                if (oldStyles && oldStyles[n] !== value) {
+                    newStyles[n] = value;
                     hasNew = true;
                 }
             });
--- a/ts/Core/Renderer/SVG/SVGElement.ts
+++ b/ts/Core/Renderer/SVG/SVGElement.ts
@@ -1061,19 +1061,12 @@ class SVGElement implements SVGElementLike {
      * @return {Highcharts.SVGElement}
      *         Return the SVG element for chaining.
      */
-    public css(styles: CSSObject): this {
+    public css(styles?: CSSObject): this {
         const oldStyles = this.styles,
             newStyles: CSSObject = {},
-            elem = this.element,
-            // These CSS properties are interpreted internally by the SVG
-            // renderer, but are not supported by SVG and should not be added to
-            // the DOM. In styled mode, no CSS should find its way to the DOM
-            // whatsoever (#6173, #6474).
-            svgPseudoProps = ['textOutline', 'textOverflow', 'width'];
+            elem = this.element;
 
         let textWidth,
-            serializedCss = '',
-            hyphenate: Function,
             hasNew = !oldStyles;
 
         // convert legacy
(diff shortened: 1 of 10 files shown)


karolkolodziej (approved):

ts/Core/Renderer/SVG/SVGElement.ts
```diff
@@ -1061,19 +1061,12 @@ class SVGElement implements SVGElementLike {
      * @return {Highcharts.SVGElement}
      *         Return the SVG element for chaining.
      */
-    public css(styles: CSSObject): this {
+    public css(styles?: CSSObject): this {
```
pawelfus, line 1064: Q: I can't find the reason why `styles` can be undefined

pawelfus (approved):
Approved, thanks! Just one question (see inline)

ts/Core/Renderer/SVG/SVGElement.ts
TorsteinHonsi: I changed it because the implementation does `if (styles...` several places, so I assumed the typing was wrong. I'll check tomorrow if we actually have any use case where it is `undefined`. Running the test suite should reveal it if you didn't already try that...

aashayjoshi:
We were able to replace the top level highcharts.js file in our project and verify that the issue of the console error as a result of CSP were fixed.

- No more more console error
- No issue with the tooltip or chart render

That said, since we bundle highcharts library for our project we could not test other modules such as solid-gauge and bullet etc. by simply using the file loaded from github.highcharts.com but we do require the fix for all the various modules used in highcharts.

TorsteinHonsi:
Thanks @aashayjoshi !

You should be able to load all modules from github.highcharts.com, but it appears that something fails currently. Anyway, the fix should apply to all parts of Highcharts. I have verified solid gauge and bullet graph specifically.

ts/Core/Renderer/SVG/SVGElement.ts
TorsteinHonsi: It does appear to be used with `undefined`, although this probably should be refactored:

https://github.com/highcharts/highcharts/blob/2105d2b1f28241c03253205ea404ca4852e6bf04/ts/Core/Navigator.ts#L1145

That expression actually passes `false|CSSObject`.

There are also at least two other places (Legend navigation and drilldown module) where non-object arguments are passed, but I think we shouldn't address that in this PR.

pawelfus: Thanks for the explanation! I agree, not part of the this PR, however worth to refactor in the future

MI14251:
> We were able to replace the top level highcharts.js file in our project and verify that the issue of the console error as a result of CSP were fixed.
> 
> * No more more console error
> * No issue with the tooltip or chart render
> 
> That said, since we bundle highcharts library for our project we could not test other modules such as solid-gauge and bullet etc. by simply using the file loaded from github.highcharts.com but we do require the fix for all the various modules used in highcharts.

hi @aashayjoshi ,

i have angular project, can you guide how to update highcharts.js or other .js files to get the CSP issues resolved..

Thanks

raf18seb:
Hi @MI14251!

There are a lot of changes in this PR so making a plugin fully fixing a problem will be time-consuming. It will be a part of a new Highcharts release which will be released soon (in a few weeks at the soonest).

Here you have an example of a custom plugin/wrap in angular: https://stackblitz.com/edit/highcharts-angular-custom-plugin?file=src%2Fapp%2FcustomPlugin.ts

In case of any further questions, please contact our support engineers on one of our dedicated support channels. More info: https://www.highcharts.com/blog/support/

PR #16875, opened 2022-01-12
Fixed issues with enablePolling and Google Sheets.

oysteinmoseng:
Fixed issues with `data.enablePolling` and Google Sheets.
___
**NOTE: #16869 should be merged first.**

Fixes two issues with live polling and Google Sheets.
1. Starting with `enablePolling: false` and updating to `true` was not possible.
2. If updating the chart to use e.g. CSV data instead, the polling did not immediately stop, therefore overwriting the CSV.


TorsteinHonsi (approved):
Thanks!

PR #16876, opened 2022-01-12
demo/tooltip-extra-information

karolkolodziej:
Added API demo showing how to display extra information in the tooltip.


samples/highcharts/tooltip/formatter-extra-information/demo.js
```diff
@@ -0,0 +1,42 @@
...
+    xAxis: {
+        categories: ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec']
+    },
+
+    series: [{
+        data: [
+            29.9,
+            {
+                y: 71.5,
+                extraInfrmation: 'First month when XYZ introduced.'
```
raf18seb, line 22: I think we should push the users to add additional info in a `data.custom` that was added specifically for cases like this
https://api.highcharts.com/highcharts/series.line.data.custom

samples/highcharts/tooltip/formatter-extra-information/demo.details
```diff
@@ -0,0 +1,7 @@
+---
+ name: Highcharts Demo
+ authors:
+   - Torstein Hønsi
```
raf18seb, line 4: Don't you want to change the author?

raf18seb (changes requested):
Please see comments below

samples/highcharts/tooltip/formatter-extra-information/demo.html
```diff
@@ -0,0 +1,3 @@
+<script src="https://code.highcharts.com/highcharts.js"></script>
+
+<div id="container" style="height: 400px"></div>
```
TorsteinHonsi, line 3: Avoid style attributes, use a `demo.css` file instead.

TorsteinHonsi (changes requested):
Thanks for this useful demo!

I think we should encourage our users to use `pointFormat` instead of `formatter`. That's because it is JSON-compatible and thus more secure, avoiding callbacks in the options structure, and more compliant to wrappers and other languages. 

Arguably, formatters are more flexible, but in this case the implementer has control over `extraInformation`, so any logic can be added where this property is generated.

This is the equivalent setup with `pointFormat` instead of `formatter`: https://jsfiddle.net/highcharts/68pmut32/. As a consequence, the name of the sample must change too.

TorsteinHonsi (approved):

PR #16882, opened 2022-01-13
TS: stricter typing in legend and SVGElement.css

TorsteinHonsi:
- TS: Made legend options DeepPartial
- Stricter typing of CSSObjects
- Require styles argument for SVGRenderer.css


pawelfus (approved):
Thanks!

PR #16887, opened 2022-01-17
Fixed #16884, smarter defaults for a11y point descriptions.

### LINKED ISSUE #16884: [A11y] Line chart with a single data series does not announce series name, series contains undefined aria-label attribute
#### Expected behaviour
When there are multiple line series in a line chart, each line announces itself via an aria-label when the user is traversing through that line. We expect the behavior to be the same when there is only a single line series in the line chart as well.

#### Actual behaviour
When there is only a single line series, the 'name' for that line series is shown in the tooltip, but it is never announced through the screen reader because the tracker for the series contains an undefined aria-label, whereas this aria-label is defined when there are multiple line series.

Below is the jsfiddle where I have removed every line but 1 from the example provided in the 'Accessible line chart' in HighCharts documentation.

Please observe that the screen readers do not give the user any clue about the name of the data series when this is the case.
https://jsfiddle.net/h8c3Lkwy/

This is the HTML element in question, which displays an undefined aria-label
image

And yet the same aria-label is defined when I add another data series to the same chart:
image

#### Product version
v9.3.2

#### Affected browser(s)
Tested on Chromium Edge Version 97.0.1072.55

Since issue is with an empty aria-label, it can be assumed that all browsers who operate within aria-label defined accessible UIs are affected as well.

We would like to announce the name of the data series to our users, even if it is a single series (this is crucial in cases where the data point conveys additional information other than the chart title). Please let us know if this is the intended behavior and/or advise if there's a way to replicate our desired behavior through some chart config changes.

oysteinmoseng:
Fixed #16884, smarter defaults for accessible point descriptions.

### DIFF
--- a/ts/Accessibility/Components/SeriesComponent/SeriesDescriber.ts
+++ b/ts/Accessibility/Components/SeriesComponent/SeriesDescriber.ts
@@ -497,13 +497,13 @@ function defaultPointDescriptionFormatter(
     point: Accessibility.PointComposition
 ): string {
     const series = point.series,
-        chart = series.chart,
+        shouldExposeSeriesName = series.chart.series.length > 1 ||
+            series.options.name,
         valText = getPointValueDescription(point),
         description = point.options && point.options.accessibility &&
             point.options.accessibility.description,
         userDescText = description ? ' ' + description : '',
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (approved):
Thanks!

PR #16891, opened 2022-01-17
bugfix/16045-resetzoom-tooltip-opacity

### LINKED ISSUE #16045: Tooltip data not visible if we add opacity to tooltip  tyle object and hover over zoom out button

raf18seb:
Fixed #16045, tooltip with opacity was not fully hidden when hovering over a reset zoom button.

This is a continuation of https://github.com/highcharts/highcharts/pull/16054 (old PR can be closed)

### DIFF
--- a/samples/unit-tests/pointer/touch/demo.js
+++ b/samples/unit-tests/pointer/touch/demo.js
@@ -166,6 +166,8 @@ QUnit.test('followPointer and followTouchMove', function (assert) {
             target: points[0].graphic.element
         });
 
+        chart.pointer.res = false;
+
         chart.pointer.onContainerTouchMove({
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (approved):
Nice, thanks!

PR #16895, opened 2022-01-17
bugfix/11712-update-node

### LINKED ISSUE #11712: Dependency Wheel - this.linksTo is undefined after calling node.update in afterAnimate event

pawelfus:
Fixed #11712, in node-based series types like dependency wheel and network graph, calling `node.update()` in the `afterAnimate` event would throw errors.
___
Prioritized because TreeGraph series also needs this `node.update()` to be working (e.g. to hide/show children nodes).

Two issues under the hood:
- wrong `node.index` for all Node-based series (except the DependencyWheel)
- updating a node used to update `series.options.data` instead of `series.options.nodes`. Solved this via 'wrapping' `Point.update()` in `Node.update()` composition: let `point.update()` do their job, then replace configs with correct ones

Perhaps there's a different approach to solve this? I'm open to suggestions!

### DIFF
--- a/samples/unit-tests/series-sankey/sankey/demo.js
+++ b/samples/unit-tests/series-sankey/sankey/demo.js
@@ -219,6 +219,23 @@ QUnit.test('Sankey nodeFormat, nodeFormatter', function (assert) {
         -1,
         'Tooltip ok'
     );
+
+    series.nodes[0].update({
+        color: 'red'
+    });
+
+    // After update, nodes should still use nodeFormat etc
+    assert.strictEqual(
+        series.nodes[0].dataLabel.text.textStr,
+        'Nodez',
+        'Explicit nodeFormat'
+    );
+    series.nodes[0].onMouseOver();
+    assert.notEqual(
+        chart.tooltip.label.text.textStr.indexOf('Nodez'),
+        -1,
+        'Tooltip ok'
+    );
 });
 
 QUnit.test('Sankey column option', function (assert) {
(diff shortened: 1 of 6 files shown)


TorsteinHonsi (approved):
Thanks, looks good to me!

Issue #6647, opened 2017-04-26
Pie: Negative innerSize value for a fixed donut thickness


moniuch:
#### Expected behaviour
I would like to be able to deliver pie.innerSize as a constant thickness of the ring (donut), regardless of the pie current size. I would like to supply a negative value which would then be calculated by the chart during render: `innerSize: -8`  // would keep the ring 8 pixels thick.

I also think, with negative value supplied, the animation should not animate the innerSize, but the ring should draw clock or counterclockwise.

TorsteinHonsi:
Keep in mind that a percentage value for `innerSize` will also give a predictable width, but _relative to the outer size_. 

To avoid confusion, a fixed width would probably be best implemented as a new option, `pie.thickness` or similar.

moniuch:
Oh well, instead of just mentioning the need, I started to suggest the solution. Not a good practice ;-)

TorsteinHonsi:
We're always open for suggestions!

moniuch:
Issue closed due to inactivity. An example of openness. Pun intended.

billsaysthis:
Is there any ETA/release plan for this? We really need it.

raf18seb:
@billsaysthis, no ETA at this moment, but we will discuss this ticket later this week.

KacperMadej:
**PoC**: https://jsfiddle.net/BlackLabel/41tjs5ua/
**Description**: Added `thickness`: overrides `innerSize`; accepts positive numbers.
**Notes**: Not properly tested. Changed/added code marked with comments about CUSTOM. Callback with a red marker for thickness check.

Please give this a try and in case of any issues please let us know.
(FYI @billsaysthis )

PR #16899, opened 2022-01-18
Enhancement #6647, added thickness options to override innerSize.

hihajdus:
Added new option [series.pie.thickness](https://api.highcharts.com/highcharts/series.pie.thickness) describing thickness ring of the chart type donut, override [series.pie.innerSize](https://api.highcharts.com/highcharts/series.pie.innerSize). See #6647.


ts/Series/Pie/PieDataLabel.ts
```diff
@@ -656,6 +656,9 @@ namespace ColumnDataLabel {
                     relativeLength(options.innerSize || 0, newSize),
                     newSize
                 );
+                options.thickness ? // #6647
+                newSize - options.thickness * 2
+                :
```
raf18seb, line 661: The use of a ternary operator here is wrong. This line `newSize - options.thickness * 2` isn't doing anything.

ts/Series/CenteredUtilities.ts
```diff
@@ -78,7 +79,8 @@ namespace CenteredUtilities {
             plotWidth = chart.plotWidth - 2 * slicingRoom,
             plotHeight = chart.plotHeight - 2 * slicingRoom,
             centerOption: Array<(number|string|null)> = options.center as any,
-            smallestSize = Math.min(plotWidth, plotHeight);
+            smallestSize = Math.min(plotWidth, plotHeight),
+            thickness = options.thickness as number * 2 || {};
```
raf18seb, line 83: This not necessarily is a `number` because `thickness` can be `undefined` in options. You should rather check whether it is a number and then fire the code you added.

ts/Series/CenteredUtilities.ts
```diff
@@ -48,6 +48,7 @@ namespace CenteredUtilities {
         innerSize?: (number|string);
         size?: (number|string);
         slicedOffset?: number;
+        thickness?: (number|0);
```
raf18seb, line 51: Type `number|0` is actually a `number` type, because 0 is also a number.

ts/Series/CenteredUtilities.ts
```diff
@@ -129,6 +131,12 @@ namespace CenteredUtilities {
         if (positions[3] > positions[2]) {
             positions[3] = positions[2];
         }
+        // thickness overrides innerSize, less than size (#6647)  
+        if (thickness < positions[2]) {
+            if (isNumber(thickness) && thickness > 0) {
```
raf18seb, line 136: You can combine these two "if statements" into one

samples/unit-tests/series-pie/innersize/demo.js
```diff
@@ -86,3 +86,38 @@ QUnit.test(
         );
     }
 );
+
+QUnit.test('Thickness cannot be greater than size.(#6647)', function (assert) {
+    var chart = Highcharts.chart('container', {
```
raf18seb, line 91: You can (and should) use ES6 in unit tests (arrow function, const etc.)

raf18seb (changes requested):
I added few comments, please check below. We'll take care of unit tests later.

Also, is this animation correct?
pie-thickness
?

ts/Series/Pie/PieDataLabel.ts
hihajdus: Yes, I missed a line `relativeLength(options.innerSize || 0, newSize), newSize);`

ts/Series/Pie/PieSeries.ts
```diff
@@ -576,6 +576,26 @@ class PieSeries extends Series {
...
+         *
+         *
+         *
+         * `thickness` 
+         *
+         *
+         *
+         * 
+         *
+         *
```
KacperMadej, line 590: .
.
.
.
.
why?
.
.
.
.
.

but seriously, this should look more like `startAngle` (one option above).

ts/Series/Pie/PieDataLabel.ts
KacperMadej: - [ ] Is the comment here needed or useful? To get the related issue for a given like we could use git blame.

- [ ] The indentation is off and the below looks better (just the format, not the content):
```
                options.thickness ? // #6647
                    newSize - options.thickness * 2 :
                    this.translate(center);
```

ts/Series/CenteredUtilities.ts
```diff
@@ -129,6 +131,12 @@ namespace CenteredUtilities {
         if (positions[3] > positions[2]) {
             positions[3] = positions[2];
         }
+        // thickness overrides innerSize, less than size (#6647)
```
KacperMadej, line 134: `thickness` in `ts/Series/CenteredUtilities.ts` is not handled very well. Why force the cast to a number? why x2 at the start? why it falls back to an empty object? why check if it is a number after the first comparison? The comment here says "less than size" but what does this mean?

KacperMadej (changes requested):
If something (e.g. tests, API docs, etc.) is clearly missing then it should be a draft PR.

ts/Series/CenteredUtilities.ts
hihajdus: Thanks for the notes 👍
I corrected the type compatibility and improved the comment.

PR #16902, opened 2022-01-19
bugfix/16791-safari-exporting

### LINKED ISSUE #16791: Chart image export returns a completely black image when using styled mode on iOS 15.1 and macOS 12
#### Expected behaviour
When exporting a chart to an image I expect an image version of the chart to be returned

#### Actual behaviour
The call to the default highcharts export server is made, but a completely black image is returned

The bug occurs in some of these scenario's I've tested:
- iOS 15.1: both Safari and Chrome fail
- macOS 12.0.1: Safari fails, but Chrome works

But is not present here:
- ipadOS 14.7: Chrome works as expected

This can be verified on one of the samples provided on the highcharts.com website

1. Use an apple device with either Safari or Chrome with one of the versions specified above
2. Visit https://www.highcharts.com/samples/highcharts/css/exporting
3. Click the hamburger menu and export to any of the image formats
4. Verify that the image download is a black version of the chart.

My assumption is that there is a problem using / processing the CSS for the chart, since an SVG version of the chart will show a black rectangle if the CSS is deleted (either on disk or through the use of developer tools)

#### Product version
Highcharts 9.3.2

#### Affected browser(s)
iOS 15.1: safari, chrome
macOS: 12.0.1 Safari

I don't have access to more devices to test this issue on, so the issue might be more widespread than this.

pawelfus:
Fixed #16791, exporting a chart in styled mode on Safari v15 produced a black image.
___
Test case: https://jsfiddle.net/BlackLabel/fr2ajxt8/4/ (compare results in console in Safari v15+ and Chrome v97+)

### DIFF
--- a/ts/Extensions/Exporting/Exporting.ts
+++ b/ts/Extensions/Exporting/Exporting.ts
@@ -1472,7 +1472,7 @@ namespace Exporting {
                 }
 
                 // Loop through all styles and add them inline if they are ok
-                if (G.isFirefox || G.isMS) {
+                if (G.isFirefox || G.isMS || G.isSafari) {
                     // Some browsers put lots of styles on the prototype
                     for (const p in styles) { // eslint-disable-line guard-for-in
                         filterStyles((styles as any)[p], p);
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (approved):
Thanks! It took me a while to understand what was going on in that code block, but I concluded that the only difference is that those listed browsers skip the `hasOwnProperty` guard inside `objectEach`. So I rewrote the code a bit to become clearer for the next guy. Now pending CI.

PR #16904, opened 2022-01-20
bugfix/16896-hovering-map

### LINKED ISSUE #16896: Highmaps: Script error is seen on hover of map series when we have map and mapbubble series if shared tooltip is ON
#### Expected behaviour
Error should not be seen when user tries to view tooltip on highmap.

#### Actual behaviour
Plot highmap using map and mapbubble series with tooltip enabled. On hover of the series, script error is seen on console.

Uncaught TypeError: Cannot read properties of undefined (reading 'pos')
    at h.b.searchPoint (highmaps.src.js:38031)
    at highmaps.src.js:26988
    at b.findNearestKDPoint (highmaps.src.js:26984)
    at b.getHoverData (highmaps.src.js:27154)
    at b.runPointActions (highmaps.src.js:27829)
    at b.onContainerMouseMove (highmaps.src.js:27458)

https://jsfiddle.net/amrutaJgtp/4q7ysh8k/1

#### Product version
Latest version v9.3.2

#### Affected browser(s)
All

karolkolodziej:
Fixed #16896, hovering over `mapbubble` with shared tooltip caused an error.

### DIFF
--- a/samples/unit-tests/maps/mapbubble-linewidth/demo.js
+++ b/samples/unit-tests/maps/mapbubble-linewidth/demo.js
@@ -1,38 +1,37 @@
 QUnit.test('MapBubble', function (assert) {
     const chart = Highcharts.mapChart('container', {
-            series: [
-                {
-                    mapData: Highcharts.maps['countries/gb/gb-all']
-                },
-                {
-                    type: 'mapbubble',
-                    lineWidth: 2,
-                    data: [
-                        {
-                            lat: 51.507222,
-                            lon: -0.1275,
-                            z: 3
-                        },
-                        {
-                            lat: 52.483056,
-                            lon: -1.893611,
-                            z: 4
-                        },
-                        {
-                            x: 1600,
-                            y: -3500,
-                            z: 3
-                        },
-                        {
-                            x: 2800,
-                            y: -3800,
-                            z: 1
-                        }
-                    ]
-                }
-            ]
-        }),
-        controller = new TestController(chart);
+        series: [
+            {
+                mapData: Highcharts.maps['countries/gb/gb-all']
+            },
+            {
+                type: 'mapbubble',
+                lineWidth: 2,
+                data: [
+                    {
+                        lat: 51.507222,
+                        lon: -0.1275,
+                        z: 3
+                    },
+                    {
+                        lat: 52.483056,
+                        lon: -1.893611,
+                        z: 4
+                    },
+                    {
+                        x: 1600,
+                        y: -3500,
+                        z: 3
+                    },
+                    {
+                        x: 2800,
+                        y: -3800,
+                        z: 1
+                    }
+                ]
+            }
+        ]
+    });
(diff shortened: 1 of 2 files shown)


ts/Series/MapBubble/MapBubbleSeries.ts
```diff
@@ -248,6 +250,8 @@ extend(MapBubbleSeries.prototype, {
 
     processData: MapSeries.prototype.processData,
 
+    searchPoint: noop as any, // #16896
```
raf18seb, line 253: What about other series types (not only a `mapbubble`)? As far as I know, `tooltip.shared` doesn't work for all the Highcharts Maps series, so maybe we can remove this function from the whole Maps product (`map` series, the rest inherits from `map`)?

samples/unit-tests/maps/mapbubble-linewidth/demo.js
```diff
@@ -1,41 +1,55 @@
...
+        'MapBubble with linewidth- points should have stroke width.'
+    );
+
+    chart.update({
+        tooltip: {
+            shared: true
+        }
+    });
+    // Hover over the chart.
+    controller.mouseMove(200, 200);
```
raf18seb, line 49: Instead of using TestController, you can simulate hovering with
```js
chart.pointer.runPointActions({ chartX: 400, chartY: 300 });
```

TestController is good, it's your call what you choose. IMHO my way is a little bit easier and faster

raf18seb (changes requested):

ts/Series/MapBubble/MapBubbleSeries.ts
TorsteinHonsi: It looks like at least the `map` series is covered (it has `searchPoint: noop`.

However, it may be that we actually want to use k-d-search with the MapBubble. So I propose not setting it to noop but rather implement it:

```ts
    public searchPoint(e: PointerEvent, compareX?: boolean): Point | undefined {
        return this.searchKDTree({
            clientX: e.chartX - this.chart.plotLeft,
            plotY: e.chartY - this.chart.plotTop
        }, compareX, e);
    }
```

ts/Series/MapBubble/MapBubbleSeries.ts
```diff
@@ -248,6 +262,8 @@ extend(MapBubbleSeries.prototype, {
...
```
TorsteinHonsi, line 265: Forgot to remove this line? 🤔

TorsteinHonsi (changes requested):

raf18seb (approved):
Thanks!

ts/Series/MapBubble/MapBubbleSeries.ts
karolkolodziej: indeed! Thanks!

TorsteinHonsi (approved):

PR #16916, opened 2022-01-25
bugfix/16474-pie-3d-click

### LINKED ISSUE #16474: Clicking on edge of 3D Pie Chart has Inconsistent Behavior
#### Expected behaviour
Clicking on 3D Pie Chart Edge should act like a click on the surface itself.

#### Actual behaviour
Inconsistent - seems to depend on where the cursor came from.
Could do one of:
- Select Correct Segment
- Nothing
- Select Wrong Segment

https://www.highcharts.com/docs/chart-concepts/3d-charts#3d-pie
http://jsfiddle.net/ued470wz/
Click on the edge...
edge-click-0
edge-click-1
edge-click-2

#### Product version
Highcharts 9.1.2

#### Affected browser(s)
Google Chrome, Edge, Others?

pawelfus:
Fixed #16474, added missing click event callback on sides of the 3D pie slices.

### DIFF
--- a/samples/unit-tests/3d/series-pie/demo.js
+++ b/samples/unit-tests/3d/series-pie/demo.js
@@ -483,3 +483,47 @@ QUnit.test('#13804: Inactive tab animation threw', assert => {
 
     Highcharts.SVGElement.prototype.animate = animate;
 });
+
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #16917, opened 2022-01-25
bugfix/4259-setData-mutation

### LINKED ISSUE #4259: setData mutating source array
Seeing a strange behavior against Highcharts 4.1.5. When providing an Array of numbers (i.e, [1,2,3,4]), after some threshold the source array seems to be mutated into the following form:

[{y = 1}, {y=2}, {y=3}, {y=4}]

This seems to happen on larger sets of data (1300 - 1400 points + ), but we can't seem to get turboThreshold or cropThreshold to make a difference. Is there a setting we're missing, or is this a known bug?

Normally not an issue, but we're visualizing running time-series data windows and as a result are hanging on to the arrays we're calling setData() with. Once they are mutated, they are effectively "broke"...upstream sorting/binning code doesn't expect the inner objects, and stops working.

Thanks for your help!

karolkolodziej:
Fixed #4259, added `allowMutatingData` chart property along with the `cloningMethod` to prevent data mutation while updating series or point.

- [x] Check different data formats. - I decided to use `JSON.parse(JSON.stringify(...))` to create a deep copy (`Array.slice()` wasn't enough for nested object.) This method of copying isn't ideal, especially if among the data there is some callback function declared. That is why I also added the `cloningMethod` field where users might declare their own recursive method of copying suited for their data format.
- [x] Check if exporting to image works. - There is some issue with that but it isn't connected directly to that PR.
       Export the chart to the png/jpg and observe the divergence between charts + error 15.
       - PR demo: https://jsfiddle.net/BlackLabel/fme789ak/
       - v9.3.2: https://jsfiddle.net/BlackLabel/8wt5sn9a/
- [x] Deep copy of an object with functions. - As described in the first point.
- [x] Check other possible scenarios where data might be mutated. - Covered in tests.
   - [x] `Point.remove()`
   - [x] `Point.update()`
   - [x] `Series.addPoint()`
   - [x] `Series.remove()`
   - [x] `Series.update()`
   - [x] `Chart.update()`
- [x] Fix TS
- [x] Implement `try/catch`. - Handle errors caused by `JSON.stringify` limitations like for example `BigInts`, circular references etc.
- [x] Handle change of `allowMutatingData` property. - Because of memory performance reasons, the update only works in one way (from default one to the solution prohibiting mutation). In order to implement it both ways we would need to create additional references but I don't see any real-life use case for that.


ts/Core/Series/Point.ts
```diff
@@ -1021,6 +1021,17 @@ class Point {
...
 
+            // If this is the first time we're writing back to options,
+            // and createDataCoppy enabled create
+            // a copy so that we don't mutate the source array. (#4259)
+            if (seriesOptions.data &&
+                !seriesOptions.isCopy &&
+                !!chart.options.chart.createDataCoppy
+            ) {
+                seriesOptions.data = seriesOptions.data.slice();
+                seriesOptions.isCopy = true;
```
raf18seb, line 1032: `isCopy` property name is not enough specific IMHO. It's on series so it suggests that the series is a copy of something, not data.

ts/Core/Series/Point.ts
```diff
@@ -1021,6 +1021,17 @@ class Point {
             i = point.index as any;
             series.updateParallelArrays(point, i);
 
+            // If this is the first time we're writing back to options,
+            // and createDataCoppy enabled create
+            // a copy so that we don't mutate the source array. (#4259)
+            if (seriesOptions.data &&
+                !seriesOptions.isCopy &&
+                !!chart.options.chart.createDataCoppy
```
raf18seb, line 1029: Is `!!` operator necessary here?

ts/Core/Chart/ChartOptions.d.ts
```diff
@@ -95,6 +95,7 @@ export interface ChartOptions {
     borderWidth?: number;
     className?: string;
     colorCount?: number;
+    createDataCoppy?: boolean;
```
raf18seb, line 98: Typo `coppy`

samples/unit-tests/series/setdata/demo.js
```diff
@@ -647,4 +647,55 @@ QUnit.test('#8795: Hovering after zooming in and using setData with redraw set t
     controller.moveTo(150, 150);
 
     assert.ok(true, 'It should not throw');
-});
\ No newline at end of file
+});
+
+QUnit.test(
```
raf18seb, line 652: Can we add at least one test with
```js
chart.update({
    chart: {
        createDataCoppy: false
    }
});
```


?

samples/unit-tests/series/setdata/demo.js
```diff
@@ -647,4 +647,55 @@ QUnit.test('#8795: Hovering after zooming in and using setData with redraw set t
...
+                [4, 4]
+            ],
+            newData = [0, 2, 4, 6, 8],
+            chart = Highcharts.chart('container', {
+                chart: {
+                    createDataCoppy: true
+                },
+                series: [{}]
+            }, function () {
+                this.series[0].setData(oriData);
```
raf18seb, line 670: You can replace this with:
```js
series: [{
    data: oriData
}]
```

(1 redraw fewer)

raf18seb (changes requested):

ts/Core/Series/Point.ts
TorsteinHonsi: Agreed

ts/Core/Chart/ChartOptions.d.ts
TorsteinHonsi: I'm not sure if it's a good name. It says something about how we deal with the mutation problem internally, but is not immediately recognizable to the implementer. The original issue report is called "setData mutating source array". I think a more descriptive property name would be something along the lines of `chart.allowMutatingData`. Which would in turn mean the polarity would be reversed and `true` be default.

TorsteinHonsi (changes requested):
Thanks! In addition to the inline problems, there are more cases where the data is mutated. I found `Point.remove`, but all the CRUD operations on series and point should be examined. https://jsfiddle.net/highcharts/n6jL9pt3/

ts/Core/Series/Series.ts
```diff
@@ -474,6 +474,19 @@ class Series {
...
+        // and allowMutatingData enabled create
+        // a copy so that we don't mutate the source array. (#4259)
+        if (options.data &&
+            !options.copyOfOriginalData &&
+            !chart.options.chart.allowMutatingData
+        ) {
+            options.data = chart.options.chart.cloningMethod ?
+                chart.options.chart.cloningMethod(options.data) :
+                JSON.parse(JSON.stringify(options.data));
+            options.copyOfOriginalData = true;
```
raf18seb, line 487: Why do we need this flag?
If we do need this, then I think it shouldn't be stored on `options`. Maybe on a `series`? But if it's on a series, then again - name `series.copyOfOriginalData` is not clear for me.


What about `series.isDataMutated` or `series.isDataHardCoppied?


And, this goes only one way - we can make a hard copy and set `series.isDataMutated = true`, but we don't set it back to `false`.
Use-case: if I do
```js
chart.update({
    chart: {
        allowMutatingData: true
    }
});
```
and then I do `series.setData(data);`, then the object reference will exist, but `series.isDataMutated` will be still `true`, when it should be `false`.

raf18seb (changes requested):
Looks better! But I still have one naming question, see the comment below.

- [x] Also, I'd love to see a unit test with `chart.update()` ;)

raf18seb (approved):
Thanks! I have nothing more to say, except the case you'll soon describe below

karolkolodziej:
Because of memory performance reasons, the update only works in one way (from default one to the solution prohibiting mutation). 
In order to implement it both ways, we would need to create additional references.
I don't see any real-life use case for that but maybe you have a different angle on that? @TorsteinHonsi

TorsteinHonsi:
> I don't see any real-life use case for that but maybe you have a different angle on that? @TorsteinHonsi

I don't, I agree that this is not useful. In fact, I am not sure whether `allowMutatingData` should be a chart option at all, or a _mode_, a setting on for example the `Chart` class (like we do for example with `AST.allowedTags`). We can discuss that.

Another comment on the current state of the PR - it may be that we can circumvent the `JSON.parse`/`cloneMethod` problem by allowing our `merge` function to deep merge arrays. Which I think could come in handy in other situations too.  This is the code in question:

https://github.com/highcharts/highcharts/blob/a96b10039bb0239cf8a3b41d3d8fabf86a17fcd0/ts/Core/Utilities.ts#L223-L233

What happens is that `isObject(value, true)`, returns false on arrays. Resulting in primitives and arrays being copied over directly. Or rather, primitives being copied, arrays only referenced. What we would need to change is that arrays be treated like objects, that their members are also deep-copied over. But we need to do that optionally to avoid breaking existing functionality. Currently, there is one option for `merge`, if the first argument of `merge` is `true`, the following members are copied into the first object instead of creating a new clone. So how should we provide another option?

* One proposal would be to allow a number for the first argument. If we convert it to a binary, we can use the positions of the binary for separate options. Like '01' = copy = false, mergeArrays = true'. That is compact, but hardly human readable, and only allows boolean for future settings.
* Another proposal is to allow a settings object in the first argument. But since it is an object, we would need some way to discern it from the actual objects to merge. We may do that with a `type`:
```js
merge({ type: 'MergeOptions', copyIntoFirst: false, deepArrays: true });
```

I think the second structure is more future proof.

karolkolodziej:
From what I can see current merge method seems to work fine. 
I tested it in multiple ways and I am not able to break it. Am I missing something?

TorsteinHonsi:
The `merge(true, data)` call will just return a reference to the original data, so any future changes to the copy will also mutate the data. https://jsfiddle.net/highcharts/8tsqg0Le/

karolkolodziej:
I understand your point here but if you apply my recent PR to the chart with a proper flag, it works well.  
Also, all tests are passing in with this approach.

Demo: https://jsfiddle.net/BlackLabel/5va3nyk0/

ts/Core/Chart/ChartDefaults.ts
```diff
@@ -527,6 +527,35 @@ const ChartDefaults: ChartOptions = {
...
 
+    /**
+     * By default, (because of memory and performance reasons) the chart
+     * does not copy the data but keeps it as a reference.
+     * In some cases, this might result in mutating the original data source.
+     * In order to prevent that set that property to false.
+     * Please note that changing that might decrease performance,
+     * especially with bigger sets of data.
+     * If your data contains any callback function please use some alternative
+     * method of copying.
```
TorsteinHonsi, line 538: Avoid line breaks without paragraphs. Either break at 80 or do a full paragraph.
https://github.com/highcharts/highcharts/blob/master/repo-guidelines.md#1-avoid-line-breaks-without-paragraphs

ts/Core/Chart/ChartDefaults.ts
```diff
@@ -527,6 +527,35 @@ const ChartDefaults: ChartOptions = {
...
+     *
+     * @type       {boolean}
+     * @see        [chart.cloningMethod](#chart.cloningMethod)
```
TorsteinHonsi, line 541: `cloningMethod` removed?

ts/Core/Chart/ChartDefaults.ts
```diff
@@ -527,6 +527,35 @@ const ChartDefaults: ChartOptions = {
...
+     * Helpful especially if the declared data contains some functions
+     * which the default method of copying `JSON.parse(JSON.stringify(...))`
+     * won't handle. In a case like this, some recursive method of copying
+     * or another external library to handle this should be used.
+     *
+     * @type       {Highcharts.ChartCloningnCallbackFunction}
+     * @since      next
+     * @return     {Array<(PointOptions|PointShortOptions)>}
+     * @apioption  chart.cloningMethod
+     */
```
TorsteinHonsi, line 557: The `cloningMethod` is no longer in use, right?

ts/Core/Chart/ChartOptions.d.ts
```diff
@@ -94,7 +102,9 @@ export interface ChartOptions {
     borderRadius?: number;
     borderWidth?: number;
     className?: string;
+    cloningMethod?: ChartCloningnCallbackFunction
```
TorsteinHonsi, line 105: Remove?

ts/Core/DefaultOptions.ts
```diff
@@ -2885,6 +2885,22 @@ export default DefaultOptions;
...
+ * @param {Highcharts.Chart} this
+ *        The chart on which the cloning occurs.
+ *
+ * @param {Array<*>} data
+ *        The data array.
+ *
+ * @return {Array<*>}
+ *         Return the copied data array.
+ */
+
```
TorsteinHonsi, line 2903: Remove?

ts/Core/Series/Series.ts
```diff
@@ -1327,9 +1327,26 @@ class Series {
...
-            firstPoint = null;
+            firstPoint = null,
+            copiedData;
+
+        // If this is the first time we're writing back to options,
+        // and allowMutatingData enabled create
+        // a copy so that we don't mutate the source array. (#4259)
+        if (!chart.options.chart.allowMutatingData) {
+            // Remove old reference
+            if (series.options.data) {
```
TorsteinHonsi, line 1338: We already have a local reference, `options`.

ts/Core/Series/Series.ts
```diff
@@ -1327,9 +1327,26 @@ class Series {
             updatedData,
             indexOfX = 0,
             indexOfY = 1,
-            firstPoint = null;
+            firstPoint = null,
+            copiedData;
+
+        // If this is the first time we're writing back to options,
+        // and allowMutatingData enabled create
+        // a copy so that we don't mutate the source array. (#4259)
```
TorsteinHonsi, line 1335: Let comments extend to 80 characters.

TorsteinHonsi (changes requested):
Thanks, it looks quite precise and compact now. Just needs some cleaning up, especially around `cloningMethod` which I believe is obsolete?

ts/Core/Series/Series.ts
```diff
@@ -1327,9 +1327,26 @@ class Series {
             updatedData,
             indexOfX = 0,
             indexOfY = 1,
-            firstPoint = null;
+            firstPoint = null,
+            copiedData;
+
+        // If this is the first time we're writing back to options, and
+        // allowMutatingData enabled createa copy so that we don't mutate the
+        // source array. (#4259)
```
raf18seb, line 1335: Typo "createa"

Sorry but I don't understand some sentences here, e.g.
```we're writing back to options```

or
```and allowMutatingData enabled createa (should it be "created" or "creates" or "creates a"?) copy so that we don't mutate the source array```

raf18seb (changes requested):
One question

TorsteinHonsi (approved):
All good from my end, but @raf18seb 's change request should be dealt with before we do the merge

raf18seb (approved):
Thanks!

Issue #3632, opened 2014-12-03
3d pie - slice rendered incorrectly when innerSize is larger than the size of a pie


KacperMadej:
When we are setting a height for 3D doughnut less than 250 it will not render properly.
Example: http://jsfiddle.net/01jrtLqm/

KacperMadej:
Shouldn't be innerSize disallowed to be larger than the size? This could solve the issue.

Any way, the problem is not present after this commit https://github.com/highslide-software/highcharts.com/commit/650eee6ca9f3c1c4ae1b345997041840eb164de5
Example: http://jsfiddle.net/01jrtLqm/1/
Related to #3323 

Better demo for this issue: http://jsfiddle.net/4axtrtv2/
Still solved by mentioned commit: http://jsfiddle.net/4axtrtv2/1/

highslide-software:
> Shouldn't be innerSize disallowed to be larger than the size? This could solve the issue.

Yes, that makes sense.

highslide-software:
@KacperMadej Did you make any code changes to this issue?

KacperMadej:
I did not make any code changes to this issue. After the commit https://github.com/highslide-software/highcharts.com/commit/650eee6ca9f3c1c4ae1b345997041840eb164de5 I am no longer able to recreate the issue. In a chart with correct innerSize (not larger than a size of a pie) issue is not visible. Should I add a limitation in code to prevent innerSize from being set as a value larger than a pie size? If yes, should be a innerSize decreased to the same (or other?) value as a size? or should those values be swapped (innerSize <-> size)?

highslide-software:
I am not sure. I guess you can take a look at the problem, and if it doesn't add too much code, add a limitation in the code. I guess decreasing the innerSize to the same value as the size would make more sense.

KacperMadej:
Doesn't look like it adds too much code - limitation added. Commit https://github.com/highslide-software/highcharts.com/commit/de685deeb20b6d590471dd2fa55ae57cb01e553f

PR #16930, opened 2022-01-27
Added thickness options.

hihajdus:
Added new option [pie.thickness](https://api.highcharts.com/highcharts/series.pie.thickness) describing thickness ring of pie. This option takes precedence over [pie.innerSize](https://api.highcharts.com/highcharts/series.pie.innerSize). See #6647.


raf18seb:
Is this ready for review? (still draft)

hihajdus:
> Is this ready for review? (still draft)

Ready to review 😊

ts/Series/Pie/PieDataLabel.ts
```diff
@@ -653,9 +653,11 @@ namespace ColumnDataLabel {
             if (newSize < center[2]) {
                 center[2] = newSize;
                 center[3] = Math.min( // #3632
-                    relativeLength(options.innerSize || 0, newSize),
+                    options.thickness ?
+                        Math.max(0, newSize - options.thickness * 2) :
```
raf18seb, line 657: Do we need this `Math.max` here? It doesn't allow `thickness` to be higher than half `size`. But is there a reason why it shouldn't be higher than half `size`? It's working fine with the same use-case of a negative `innerSize` here (no `Math.max` here):
```js
relativeLength(options.innerSize || 0, newSize),
```

Or maybe we do want it to limit both `thickness` **and** `innerSize` based on this [ticket](https://github.com/highcharts/highcharts/issues/3632)?
It seems like the above bug hasn't been fixed for a use case when the dataLabels are enabled.
Demo: http://jsfiddle.net/BlackLabel/1ys0bkz5/

ts/Series/CenteredUtilities.ts
```diff
@@ -78,7 +79,8 @@ namespace CenteredUtilities {
             plotWidth = chart.plotWidth - 2 * slicingRoom,
             plotHeight = chart.plotHeight - 2 * slicingRoom,
             centerOption: Array<(number|string|null)> = options.center as any,
-            smallestSize = Math.min(plotWidth, plotHeight);
+            smallestSize = Math.min(plotWidth, plotHeight),
+            thickness = options.thickness || 0;
```
raf18seb, line 83: You don't need `|| 0` here because you are checking `isNumber(thickness)` below. If TypeScript will mess with you about it, send him to me and we'll explain this and that...! (if you know what I mean)

...
(simply check `isNumber(thickness)` **first** in your `if` statement)

ts/Series/CenteredUtilities.ts
```diff
@@ -129,6 +131,13 @@ namespace CenteredUtilities {
         if (positions[3] > positions[2]) {
             positions[3] = positions[2];
         }
+        // thickness overrides innerSize, need to be less than pie size (#6647)
+        if (
+            thickness < positions[2] &&
```
raf18seb, line 136: If you decide that a negative `innerSize` shouldn't be allowed, then it should be
```js
thickness * 2 < positions[2] &&
```

raf18seb (changes requested):
Looking better! I have a few last questions before we jump to unit tests

ts/Series/Pie/PieDataLabel.ts
hihajdus: `thickness` should be limited so that `innerSize` is not set to more than `size` and is not less than 0.

raf18seb (changes requested):
Thanks! We can start working on tests now, I'll add some suggestions soon.

TO DO:

- [x] unit tests

raf18seb (changes requested):
Some test ideas:
- [ ] set `thickness` without any other options, test if `thickness + series[0].points[0].shapeArgs.innerR === size / 2`
- [ ] with the same options, fire mouse hover event somewhere in the middle of a pie to see if no point is hovered
- [ ] then do `series.update({ thickness: ..., size: ... })` to a different values and test again
- [ ] then disable `thickness` and check the above again
- [ ] you can also try similar asserts with very huge thickness and negative thickness (another updates)
- [ ] you can also add `series.innerSize` and see if it doesn't change anything (`thickness` should override `innerSize`

Some about 5-6 asserts should be anought

samples/unit-tests/series-pie/innersize/demo.js
```diff
@@ -86,3 +86,76 @@ QUnit.test(
...
+        chart.plotTop + chart.series[0].center[1] + 5
+    );
+    assert.strictEqual(
+        clicks, 1, 'Clicking on a side of a pie slice should fire click event.'
+    );
+
+    series.update({
+        thickness: undefined
+    });
+    assert.ok(true, `After disable thickness, shouldn't be displayed.`);
```
raf18seb, line 153: `assert.ok(true);`  isn't testing anything specific (it fails only when errors occur). When updating **any options**, the actual expected result should be tested (you can reuse the above asserts).

samples/unit-tests/series-pie/innersize/demo.js
```diff
@@ -86,3 +86,76 @@ QUnit.test(
...
+
+    series.update({
+        thickness: 20,
+        size: '50%',
+        innerSize: '150px'
+    });
+    assert.ok(true, 'Should working after update and override innerSize.');
```
raf18seb, line 160: Same as above

raf18seb (changes requested):
Some tests should be still improved. See the comments below

TorsteinHonsi (changes requested):
Thanks, it's looking good. My only change request is that a note is added to the API for `innerSize`, stating that this setting is overridden by `thickness`.

raf18seb (approved):
Good job, thanks!!
I'll only clean up and merge some unit tests and it's ready for a final review by @TorsteinHonsi

TorsteinHonsi (approved):

PR #16935, opened 2022-01-28
bugfix/13033-toggle-datasorting

### LINKED ISSUE #13033: Toggling the dataSorting feature does not work.
#### Expected behaviour
Enabling, and disabling the data sorting e.g through `Series.update()` should work correctly, like  updating any other series parameter.

#### Actual behaviour
An attempt to toggle data sorting feature through `update()` does not work at all.

https://jsfiddle.net/BlackLabel/nys1kc5p/

#### Workaround
The one way that I found to make it work somehow, is by adding **x** values into data, calling `Series.update()` with the new value of `dataSorting.enabled`, and after that doing a `Series.setData()` with copied data in order to force re-render a series.
Live example: https://jsfiddle.net/BlackLabel/d8n71jzw/

pawelfus:
Fixed #13033, toggling `series.dataSorting` option didn't work.
___
I don't like it's happening in `series.update()` but it seems the right place. Unless I'm missing something. `dataSorting` already modifies `x` values, so now it's additionqlly removing these values. That means if a dataset like `[{ x: 2, y }, { x: 1, y}]` will be part of the config, then x's will be removed, leaving: `[ { y }, { y } ]` in `series.options.data`.


ts/Core/Series/Series.ts
```diff
@@ -4225,8 +4226,11 @@ class Series {
             typeof options.pointStart !== 'undefined' ||
             typeof options.pointInterval !== 'undefined' ||
             typeof options.relativeXValue !== 'undefined' ||
+            options.dataSorting &&
```
TorsteinHonsi, line 4229: Can we make it simpler by just testing `typeof options.dataSorting`?

ts/Core/Series/Series.ts
```diff
@@ -4225,8 +4226,11 @@ class Series {
...
+                typeof options.dataSorting.enabled !== 'undefined' ||
             options.joinBy ||
             options.mapData || // #11636
+            series.hasOptionChanged('enabled', 'dataSorting') ||
```
TorsteinHonsi, line 4233: Do we need this?

ts/Core/Series/Series.ts
```diff
@@ -4374,6 +4378,21 @@ class Series {
...
+        if (
+            options.data &&
+            options.dataSorting &&
+            options.dataSorting.enabled === false &&
+            oldOptions.dataSorting &&
+            oldOptions.dataSorting.enabled === true
+        ) {
+            options.data.forEach((point): void => {
+                if (isObject(point, true)) {
+                    point.x = point.index = void 0;
```
TorsteinHonsi, line 4391: It is kind of bad practice that the options are mutated so the we need to reset them here. Is there another way we could do this in the first place, for example storing a lookup table for data sorting?

ts/Core/Series/Series.ts
```diff
@@ -4374,6 +4378,21 @@ class Series {
             (series as any)[prop] = (preserve as any)[prop];
         });
 
+        // When switching back from sorted data to unsorted, reset indexes:
+        if (
+            options.data &&
+            options.dataSorting &&
+            options.dataSorting.enabled === false &&
+            oldOptions.dataSorting &&
+            oldOptions.dataSorting.enabled === true
```
TorsteinHonsi, line 4387: Do we need all the conditions? Maybe a less-than-perfect test like `if (options.data and 'dataSorting' in options)` would be good enough?

ts/Core/Series/Series.ts
```diff
@@ -4430,6 +4449,18 @@ class Series {
...
         }
 
+        // If dataSorting was turned on, set the indexes
+        if (
+            options.dataSorting &&
+            options.dataSorting.enabled === true &&
+            (
+                !oldOptions.dataSorting ||
+                oldOptions.dataSorting.enabled === false
+            )
```
TorsteinHonsi, line 4459: Again, do we need all the conditions? Or would it be good enough to test for `'dataSorting' in options`? Sure, it would kick in when changing anything else in `dataSorting`, but does it matter?

ts/Core/Series/Series.ts
```diff
@@ -4450,24 +4481,36 @@ class Series {
      * Check if the option has changed.
      * @private
      */
-    public hasOptionChanged(optionName: string): boolean {
+    public hasOptionChanged(optionName: string, parentName?: string): boolean {
```
TorsteinHonsi, line 4484: Do we really need to do a deep test for `enabled`?

TorsteinHonsi (changes requested):
Thanks for the PR! And sorry for the late reply. I spent the last month thinking about whether we should add this much code for a relatively low priority issue. 

So my comments also revolve around the same question - can we do this shorter and cheaper if we accept it is not perfect?

ts/Core/Series/Series.ts
pawelfus: Changed to `typeof options.dataSorting !== 'undefined'` 👍

ts/Core/Series/Series.ts
pawelfus: It will be cost us some performance but will work, changed 👍

ts/Core/Series/Series.ts
TorsteinHonsi: Thanks, I'll wait for that

PR #16936, opened 2022-01-28
tests/remove-highcharts-each-map-etc

pawelfus:
Just a short clean-up in tests (unit-tests and studies dirs). I did not change official demos or samples.


TorsteinHonsi (approved):
Awesome, thanks!

PR #16939, opened 2022-01-31
Website/icon editor

nasin:
Hi,

Ferdinand asked me to make a more appropriate icon for the Editor product page. I created a new animation for the icon.

For the static chart at the end of the animation, I thought it would be cool to show the actual editor in action. I used the code from the standalone editor and made a mini editor.

I disabled a lot of stuff in eslint that I probably shouldn't have. 

Let me know what you think and how I should proceed. If this is too complicated, I can make a regular chart.


TorsteinHonsi:
Looks good to me!

Architecture wise, for future projects I think it would be better to keep complex HTML structures like the dropdown box outside the chart instead of in the chart title configuration, instead of re-purposing the chart title to do something that is not actually part of the chart. It could be positioned on top of the chart container by CSS. But that's just thoughts for future projects, I don't think we should spend time=money changing that for this case.

One small detail that would be nice would be that the dropdown showed the current month. You can do `new Date().getMonth()` and set the selected item.

TorsteinHonsi (approved):
Looks good to me! I think it is better to ignore ESlint errors instead of start fixing what belongs in a different project.

PR #16949, opened 2022-02-02
bugfix/14175-datalabels-rotation-update

### LINKED ISSUE #14175: Misplaced dataLabel after rotation update.
#### Expected behaviour
Data labels shouldn't be misplaced - like in the v7.0.3

#### Actual behaviour
Since v7.1.0 data labels are misplaced after performing the update of it's rotation

Steps: 
- click on the "vertical" button, then on the "horizontal" one. 
- observe that the data labels position is not correct
- change highcharts version to the 7.0.3
- reply first step (now the data labels are working correctly)
https://jsfiddle.net/BlackLabel/zx64ftbq/

#### Product version
8.2.0

#### Affected browser(s)
All

pawelfus:
Fixed #14175, enabling and disabling `dataLabels.rotation` misplaced data label's position.
___
This is a regression since `soft series update` was implemented. Previously, we used to destroy labels and recreate them. This PR restores that logic (enabling/disabling rotation now destroys labels). Note: data label is created using `text()` (w/o rotation) or `label()` (w/ rotation), so it seems a reasonable solution.

I have tried to fix this on the alignment level too. It seems rotated labels set `alignFactor = 0.5` while unrotated don't (`alignFactor = 0`). That means returning back to unrotated ones, we render labels with the wrong factor. Tried to fix that, but then we have another and another and another.. issue. Unless we decide to implement animation for rotation change, I would not touch this.

### DIFF
--- a/samples/unit-tests/datalabels/update/demo.details
+++ b/samples/unit-tests/datalabels/update/demo.details
@@ -1,0 +1,6 @@
+---
+ resources:
+   - https://code.jquery.com/qunit/qunit-2.0.1.js
+   - https://code.jquery.com/qunit/qunit-2.0.1.css
+ js_wrap: b
+...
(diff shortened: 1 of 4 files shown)


TorsteinHonsi (approved):
Looks good!

PR #16952, opened 2022-02-03
docs/breadcrumbs 

karolkolodziej:
Added general documentation for breadcrumbs.


TorsteinHonsi (changes requested):
Thanks!

The article confuses the Breadcrumbs as a part of the Drilldown module that is used in treemap and sunburst. That is not strictly correct - the Breadcrumbs is a standalone feature that is used by both the Drilldown module and the treemap and sunburst series. 

The treemap and sunburst series have nothing to do with drilldown, it is just a related concept.

In addition to these two use cases, I think we should mention how the Breadcrumbs module can be used standalone. We have a sample for that already.

docs/chart-and-series-types/treemap.md
```diff
@@ -149,6 +149,11 @@ levelIsConstant is an option used together with the levels and allowDrillToNode
 
 An [example](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/demo/treemap-large-dataset) where levelIsConstant is set to false.
 
+##### Navigating through levels
+For multiple levels consider using breadcrumb from drilldown module to help navigate through the chart.
```
TorsteinHonsi, line 153: It's nothing to consider - the breadcrumbs are part of the treemap/sunburst and will appear when the end user traverses. Actually the process of traversing the levels is not well covered in the docs currently, so it should be mentioned too. Name this section "Traversing through levels" so that we stick to one terminology.

TorsteinHonsi (changes requested):

TorsteinHonsi (approved):
It's good now, thanks!

PR #16954, opened 2022-02-03
bugfix/14167-setVisible-docs

### LINKED ISSUE #14167: Point.setVisible() missing in API docs
Link: https://api.highcharts.com/class-reference/Highcharts.Point

Note `Point.setVisible()` is available only for Pie series. Most of the others don't have this method implemented.

karolkolodziej:
Fixed #14167, added missing documentation for point setVisible method.

### DIFF
--- a/ts/Series/Pie/PiePoint.ts
+++ b/ts/Series/Pie/PiePoint.ts
@@ -180,8 +180,8 @@ class PiePoint extends Point {
     }
 
     /**
-     * Toggle the visibility of the pie slice.
-     * Note that this method is available only for `pie` series.
+     * Toggle the visibility of the pie slice. Note that this method is
+     * available only for some series, like pie, treemap and sunburst.
      *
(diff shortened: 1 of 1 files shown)


ts/Series/Pie/PiePoint.ts
```diff
@@ -181,11 +181,19 @@ class PiePoint extends Point {
 
     /**
      * Toggle the visibility of the pie slice.
-     * @private
+     * Note that this method is available only for `pie` series.
```
TorsteinHonsi, line 184: Not only pies. Other series like funnel, treemap, timeline etc. also implement it. It is probably safest to say "available only for some series, like pie, treemap and sunburst". Then don't exclude anything.

TorsteinHonsi (changes requested):

TorsteinHonsi (changes requested):

TorsteinHonsi (approved):

PR #16956, opened 2022-02-03
bugfix/16944-tooltip-outside-pos

### LINKED ISSUE #16944: When using tooltip.outside, tooltip gets misaligned
#### Expected behavior
Tooltip should not get misaligned when using `outside: true` property

#### Actual behavior
Tooltip is not aligned properly when using the outside prop

Check the demo example https://jsfiddle.net/rizzotto/urhkswtj/ and set the property `outside` to `true`

tooltip

hihajdus:
Fixed #16944, wrong tooltip position when using the `positioner` callback and `outside: true`.


raf18seb:
Please remember to check the tooltip's position without `tooltip.positioner`

hihajdus:
TO DO:
- [ ]  unit tests

TorsteinHonsi (changes requested):
I don't think it's connected to the `distance`. In the OP demo, if you set the `distance` option to something non-default, like 0 or 50, you'll see that the position is different when toggling `outside`.

Instead, it's about the chart's position in the page. The positioner is not corrected for the chart position.

* When the body has a 0 margin, [the position is correct](https://jsfiddle.net/highcharts/13jxnq29/1/).
* When the body has another margin, [the position is off by that margin](https://jsfiddle.net/highcharts/13jxnq29/2/)
* In a [real-life web page, it is off](https://jsfiddle.net/highcharts/13jxnq29/3/).

### Workaround
```js
        positioner(labelWidth, labelHeight, point) {
            const chartPosition = this.chart.pointer.getChartPosition();
            return {
                x: chartPosition.left + point.plotX + this.chart.plotLeft - labelWidth / 2,
                y: chartPosition.top + this.chart.plotTop - 20,
            }
        },
```
[View it live on jsFiddle](https://jsfiddle.net/highcharts/13jxnq29/4/)

### Suggested fix
Adjust for chart position internally after calling `positioner`, but only if `outside`

hihajdus:
Thanks for the detailed explanation of how work positioning with `distance` and `outside`. 

With the `positioner` it was automatically set `this.distance = 16` and this caused an extra offset.

TorsteinHonsi:
Thanks! But I can't see that it is corrected for the chart position. Like I explained above, I don't think it is connected to the `distance`, but instead the chart position in the page. https://jsfiddle.net/highcharts/13jxnq29/5/

samples/unit-tests/tooltip/outside/demo.js
```diff
@@ -82,6 +82,53 @@ QUnit.test('Outside tooltip styling', function (assert) {
...
+QUnit.test('Tooltip with default positioner and set outside true on correct position (#16944)', assert => {
+    const chart = Highcharts.chart('container', {
+            chart: {
+                width: 400,
+                height: 400
+            },
+            tooltip: {
+                formatter() {
+                    return this.y + 'AAAAAAAA';
+                },
```
pawellysy, line 94: maybe use tooltip.format, or if not necessary delete this formatter function?

pawellysy (approved):
Looks good! Just one minor comment to the test.

samples/unit-tests/tooltip/outside/demo.js
```diff
@@ -82,6 +82,53 @@ QUnit.test('Outside tooltip styling', function (assert) {
     );
 });
 
+QUnit.test('Tooltip with default positioner and set outside true on correct position (#16944)', assert => {
```
raf18seb, line 85: I can see this bug is very much based on the margin set. What about adding one more assert with css margin or `chart.margin` being set?

Also, is it possible to reuse an already existing chart instead of creating a new `QUnit.test` with a new chart? We already have thousands of tests and if we will create a separate chart for every bug, we will soon end up with very low performance (every tests run will be very long).

raf18seb (changes requested):
Looks good now! I have 2 suggestions below

samples/unit-tests/tooltip/outside/demo.js
hihajdus: Thanks for the suggestions, I will keep in mind about test and performance.

In this case eventually, we can put in `samples/unit-tests/tooltip/position` but when I added asserts the code looked unreadable, I insert in a separate test.

samples/unit-tests/tooltip/outside/demo.js
```diff
@@ -82,6 +82,68 @@ QUnit.test('Outside tooltip styling', function (assert) {
...
+        chart: {
+            margin: 50
+        }
+    });
+
+    tooltip.refresh(point);
+
+    chart.container.style.margin = '200px';
+
+    tooltip.refresh(point);
```
pawellysy, line 138: Do you need to refresh the tooltip twice?

pawellysy (changes requested):

samples/unit-tests/tooltip/outside/demo.js
hihajdus: Thanks for a check! I don't need a second refresh.

TorsteinHonsi (changes requested):
Some remaining glitches:

With this test case: https://jsfiddle.net/highcharts/13jxnq29/7/ .  Change `distance` to 50. When `outside`, the tooltip misses the target, unlike when not outside. Then the tooltip is on target.

With this test case: https://jsfiddle.net/highcharts/13jxnq29/8/ . The `positioner` option is not in use. `outside` is false and the tooltip is aligned to the blue line. Change `outside` to true, and it is misaligned. The PR [changes position also when positioner is not in use](https://github.com/highcharts/highcharts/pull/16956/files#diff-b74ef0962c761aa1326fc9274578c213e2ce811d2d5ce3820bde2a64a6f772e3R1827-R1830), which I believe was not an issue?

hihajdus:
Yes, it was just a matter of positioning the tooltip with `postioner` and `outside` true.
Thanks for testing, I'll fix that.

TorsteinHonsi (approved):

TorsteinHonsi:
Pending approval by @raf18seb @karolkolodziej and @pawellysy

samples/unit-tests/tooltip/outside/demo.js
raf18seb: What about adding your updates and asserts to the test above?
You can rename the title from `"Outside tooltip styling"` to `"Outside tooltip styling and position"`. The chart's config is almost the same so it should be easy to combine these 2 tests into one.

pawellysy (approved):
Looks good 👍

package-lock.json
```diff
@@ -27,7 +27,7 @@
         "browserify": "^17.0.0",
         "cloudflare": "^2.9.1",
         "colors": "^1.4.0",
-        "cypress": "7.7.0",
+        "cypress": "^12.0.2",
```
karolkolodziej, line 30: `package-lock.json` should not be committed

karolkolodziej (approved):

raf18seb (approved):
Thanks!

PR #16961, opened 2022-02-04
docs/16947-linearGradient-gradientUnits 

### LINKED ISSUE #16947: Legend color doesn't show up for a line graph under certain circumstances
#### Expected behaviour

Color marker for legend (symbolColor) should show up under all circumstances, even when using a linear color gradient for a line with the marker disabled.

#### Actual behaviour

The symbolColor doesn't show up with this format:
 *   linearGradient: { x1: 0, y1: 0, x2: 1, y2: 0 }
While it works fine with this format:
 *   linearGradient: [ 0, 0, 0, 400 ]
I'd rather use the object format.

Demo: https://jsfiddle.net/jwinkle/pc09uy4m/19/

The fiddle graph has two green-to-white lines and while the symbolColor using the array format looks fine, you'll notice that the symbolColor using the object format doesn't show up.

I learned of workarounds (see post below), but it seems like I shouldn't need a workaround.

If it's helpful, my original post is here: https://stackoverflow.com/questions/70932592/how-to-make-legend-color-show-up-in-highcharts-for-a-line-graph-under-certain-co/70933811

#### Product version

Highcharts 9.3.2

#### Affected browser(s)

All tested: Safari, Chrome, Firefox

Thanks!

karolkolodziej:
Added explanation of `gradientUnits` in `linearGradient` documentation, resolves #16947.

### DIFF
--- a/docs/chart-design-and-style/colors.md
+++ b/docs/chart-design-and-style/colors.md
@@ -35,6 +35,12 @@ The color is given as an object literal containing two properties:
 
 In modern SVG enabled browsers the linear gradients can be applied to both fills (backgrounds) and strokes (lines). Internet Explorer 8 and less doesn't support gradients on lines, so in these browsers only the first stop color is used.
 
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (approved):
Thanks!

PR #16962, opened 2022-02-04
Fixed missing images and mobile issues

nasin:
Hi, I fixed a couple of issues with missing images and mobile styles for both the animation and the static editor.


TorsteinHonsi (approved):
Thanks!

PR #16963, opened 2022-02-04
bugfix/16921-heatmap-marker-hover-zoom

### LINKED ISSUE #16921: Mouse over of zoomed heat map regions resizes regions
#### Expected behaviour
When a heat map is zoomed in (with zoomType:['xy']), mousing over the heat map panels should produce a visual focus on the panel and leave the region in its current, zoomed-in size and the panels in their current locations.

#### Actual behaviour
Mousing over a heat map panel resizes the panel back to its original size and places them back where they were in 100% zoom.  Resetting zoom and mousing over un-zoomed panels resizes and moves them as if they were still zoomed.  Tooltips are misaligned.

https://jsfiddle.net/L1skxfv8/
Drag to select any area of the heatmap to zoom in
After zoom, mouse over the heat map panels.
> Panels are resized back to 100% zoom size and placed back in their original locations

Reset zoom
Mouse over any panels
> Panels are resized to the size they should have been in the zoomed-in view.

#### Product version

Highcharts 9.3.2

#### Affected browser(s)
All

pawelfus:
Fixed #16921, regression causing the wrong size of a hovered heatmap cell after zooming in.
___
I have moved the comment to the place. Note that `option.r` is supported (maybe it wasn't at a time when the comment was added) in `drawPoints()`. Probably if we move `borderRadius` to `marker.radius` or `marker.r` (in API), then it will work without the workaround. Casted to `any` because `r` is not a valid option, however, supported.


TorsteinHonsi (approved):
Thanks!

PR #16975, opened 2022-02-08
bugfix/16788-waterfall-focusable-points

### LINKED ISSUE #16788: Not all bars in waterfall chart is focusable
#### Expected behaviour
All the bars in a waterfall chart should be focusable when a11y is enabled. 

#### Actual behaviour
Certain bars in a waterfall chart are not focusable. This seems to happen when the bars have negative values but the axis doesn't have negative values.

https://jsfiddle.net/oehu0tx6/24/

#### Product version
9.3.2

#### Affected browser(s)
Chrome, Egde

mfilipiec:
Fixed #16788, accessibility issue with some columns not being focusable because of wrong `point.isInside` flag.

### DIFF
--- a/samples/unit-tests/series-waterfall/waterfall-array-points/demo.js
+++ b/samples/unit-tests/series-waterfall/waterfall-array-points/demo.js
@@ -20,21 +20,21 @@ QUnit.test('Waterfall point definitions(#4094)', function (assert) {
         ]
     });
 
-    assert.equal(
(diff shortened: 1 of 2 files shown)


raf18seb (approved):

oysteinmoseng (approved):
Works great now!

TorsteinHonsi (approved):
Thanks!

PR #16977, opened 2022-02-08
Feature/node-series-refactor

pawellysy:
Example refactor of the series which have nodes.


pawellysy:
In most of the series, which have nodes, there is the same `pointClass` for both links and nodes. These classes have properties and methods for both nodes and for links. In this PR there is a way of separating the classes responsible for links and nodes so that the objects only have the properties that make sense for them. What do you think about this solution @TorsteinHonsi?

TorsteinHonsi:
@pawellysy I'm good with that

TorsteinHonsi (approved):

PR #16980, opened 2022-02-09
Themes/high contrast themes

nasin:
Hi, here are the revised high contrast themes. I tested the color combinations for colorblindness, and all the colors are 4.5:1.


oysteinmoseng:
Looks great, and a clear improvement on the existing themes I think! The 4.5:1 is a bonus, only 3:1 is necessary for non-text here, but since it looks good as is I would keep it.

Did we also want to add some themes with fewer colors, in order to make them look better, or do you not think that is necessary?

TorsteinHonsi (approved):
Thanks!

PR #16992, opened 2022-02-14
Feature/zoom in polar

bm64:
To do: 
- [x] - Implement zoom x type
  - [x]  tests 
  - [x] - ~http://jsfiddle.net/BlackLabel/3nbyL76q/ - zooming above dataMax doesn't work correctly~ deleted, because issue was most likely caused by minRange 0
  - [x] - extract start/end angle normalization logic http://jsfiddle.net/BlackLabel/k37pwd5s/
  - [x] - correct selection for negative angles http://jsfiddle.net/BlackLabel/jp2egm84/
  - [x] - correct swap angle http://jsfiddle.net/BlackLabel/4da16efz/
- [x] - Zoom type xy tests
- [x] - **Check zooming behaviour on slice (e.g. start - end Angle 210 - 360)**
  - [x] - correct selection shape, both for x and y zooming
  - [x]  - test prevent selection outside the slice (check if selection exists)
- [x] - **Check all series with zoom in polar**
  - [x] - [Column](http://jsfiddle.net/BlackLabel/gaLhxwcz/) - cropping doesn't work, check _isValid_
    - [x] - correct animation
  - [x]  - [Line](http://jsfiddle.net/BlackLabel/yjmrostp/) - doesn't work correctly, out of range points disappears after reset zoom
  - [x] - [Scatter](http://jsfiddle.net/BlackLabel/aewzv546/) - removes out of range points after reset zoom, decide how to close shape
  - [x] - [Area](http://jsfiddle.net/BlackLabel/aewzv546/) - removes out of range points after reset zoom, decide how to close shape
  - [x] - [Areaspline](http://jsfiddle.net/BlackLabel/ybcjmn54/) - removes out of range points after reset zoom, decide how to close shape
  - [x] - [Bar](http://jsfiddle.net/BlackLabel/5eatcdkg/) - zooming below 1 doesn't for polar and cartesian (https://jsfiddle.net/BlackLabel/1wj4rxfs/)
  - [x] - [Arearange](http://jsfiddle.net/BlackLabel/L8aekp3g/) - removes out of range points after zoom reset
- [x] - **Check behaviour in inverted polar**
  - [x] - Add tests for zoom in inverted polar (https://jsfiddle.net/BlackLabel/0dxkou5t/)
- [x] - **Touch device zoom**
  - [x] - Not working: https://jsfiddle.net/BlackLabel/k37pwd5s/4/show (pinch is not implemented for polar coordinates)
- [x] - **Test zooming for two polars in single chart**  http://jsfiddle.net/BlackLabel/86pr4tc5/
- [x] - stacking https://jsfiddle.net/BlackLabel/zcwom9b1/ - zoom extremes are correct, series are incorrect
- [x] - check disappearing labels after reset zoom (https://jsfiddle.net/BlackLabel/4Lax5ruq/) - zoom 4 - max - existing bug, not related to PR
  - [x] create a ticket: https://github.com/highcharts/highcharts/issues/17187
- [x] - Spiderweb - correct selection shape
- [x] - Add demo with zooming - start with scatter
- [x] - make sure zoom is disabled for mobile devices


ts/Core/Chart/Chart.ts
```diff
@@ -3521,6 +3521,52 @@ class Chart {
...
     }
 
+    /**
+     * Checks if mouse pointer is within pane.
+     *
+     * @private
+     * @function Highcharts.Chart#isWithinPane
+     * @emits Highcharts.Chart#event:afterIsWithinPane
+     */
+    public isWithinPane(
```
TorsteinHonsi, line 3531: We already have `chart.isInsidePlot`. See if we can use that directly instead of adding a function that is quite similar.

ts/Core/Series/Point.ts
```diff
@@ -441,6 +438,19 @@ class Point {
         chart.pointCount--;
     }
 
+    /**
+     * Determine if point is null.
+     *
+     * @private
+     * @function Highcharts.Point#determineIsNull
+     */
+    public determineIsNull(): boolean {
```
TorsteinHonsi, line 447: Do we need a new function? Usually in extensions we just override the `Point.isValid` function.

ts/Core/Chart/Chart.ts
bm64: Perhaps the method should be called `isInsideAxis`? `isInsidePlot` checks if `x`+`y` position is within the plotting area. Meanwhile `isInsideAxis` checks if `x` or `y` is within `xAxis` or `yAxis`. This is used for example in multiple panes, like Stock, and zooming on yAxis (when selecting area within one pane, then only that one yAxis/pane should zoom in). Demo: https://www.highcharts.com/demo/stock/stock-tools-gui (select yAxis zoom from StockTools and zoom in on both panes)

ts/Core/Series/Point.ts
bm64: [This logic](https://github.com/highcharts/highcharts/pull/16992/files/e27540ab39c2341724e2e59dfe3c8eeae2a58dfd#diff-185b42d88f432e641196b68d7e7f2da4ac7c3fd3d87ccffa9da96eea6a458941L342-L345) was extracted into determineIsNull function in order to restore points’ graphics after zoom reset. Before the change it would be called only initially resulting in missing points: https://jsfiddle.net/BlackLabel/7nx50sm8/ (click apply zoom then reset zoom)

EWChina999:
Please kindly  add a feature -  we can set  Clockwise or Counterclockwise in polar charts . In semiconductor field , we need to check symmetrical damage and zoom the defect area for details. See the below example , during spinning, a physical stuff will cause symmetrical damage on both side , same location.
image

ts/Core/Chart/Chart.ts
TorsteinHonsi: Can you check that again? `isInsidePlot` does lok at `series.xAxis` and `series.yAxis`, and uses axis.pos and axis.len to determine wheter the point is within. That is the same as looking at axis.left, top, width and height. I suspect that `isInsidePlot` can be used either directly or with small modifications.

TorsteinHonsi: The `isInsidePlot` function also has `ignoreX` and `ignoreY` options so you can use it to inspect one axis at the time.

ts/Core/Pointer.ts
```diff
@@ -404,6 +460,33 @@ class Pointer {
...
+                marker.width,
+            height = marker.attr ?
+                marker.attr('height') :
+                marker.height;
+
+        let event = { x, y, width, height, marker };
+
+        fireEvent(
+            this,
+            'afterGetSelectionBox',
```
TorsteinHonsi, line 483: As above, unless we're extending on the default functionality, don't use the `after*` pattern.

TorsteinHonsi (changes requested):
In addition to the inline comments, I am not sure we need two separate methods (and events) for `createSelectionMarker` and `getSelectionMarkerAttrs`. 

The thing is that the selection marker can either a `rect`, `path` or `arc`. And the attributes are always `SVGAttributes`. All of these objects can be created without initial attributes, we we can probably just create one singe extendable function that returns the element type and the attributes. Much like `point.shapeType` and `point.shapeArgs`. 

```ts
            // make a selection
            const selectionMarkerShape = this.getSelectionMarker(chartX, chartY);
            if (
                (chart.hasCartesianSeries || chart.mapView) &&
                (this.zoomX || this.zoomY) &&
                clickedInside &&
                !panKey
            ) {
                if (!selectionMarker) {
                    this.selectionMarker = selectionMarker = chart.renderer[selectionMarkerShape.type]()
                        .attr({
                            'class': 'highcharts-selection-marker',
                            zIndex: 7
                        })
                        .add();

                    if (!chart.styledMode) {
                        selectionMarker.attr({
                            fill:
                                chartOptions.selectionMarkerFill ||
                                color(Palette.highlightColor80)
                                    .setOpacity(0.25).get()
                        });
                    }
                }
            }

            if (selectionMarker) {
                selectionMarker.attr(
                    selectionMarkerShape.attribs
                );
            }
```

bm64:
Hello @EWChina999! Sorry for longer response time and thanks for sharing the idea. This ticket is strictly related to polar zoom, so if you don’t mind, please create a separated feature request so we can carry the discussion there: https://github.com/highcharts/highcharts/issues/new/choose

ts/Core/Series/Point.ts
TorsteinHonsi: That logic is there because the `isValid` function is not part of the basic Point object, just some extended objects. So instead of wrapping it, I think we should just implement the function on the basic Point.

```ts
public isValid(): boolean {
   return this.x !== null && isNumber(this.y);
}
```

ts/Core/Chart/Chart.ts
```diff
@@ -742,11 +742,14 @@ class Chart {
...
                 x,
                 y,
-                isInsidePlot: true
+                isInsidePlot: true,
+                options
             };
 
         if (!options.ignoreX) {
             const xAxis = (
+                options.axis && options.axis.isXAxis && options.axis
```
TorsteinHonsi, line 751: Is this line necessary? Or maybe we can adapt the next line and make an exception for polar charts?

`series && (inverted && !chart.polar ? series.yAxis: series.xAxis)`.

ts/Core/Pointer.ts
```diff
@@ -262,6 +264,68 @@ class Pointer {
...
 
+    /**
+     * Calculate attrs for selection marker.
+     * @private
+     * @function Highcharts.Pointer#getSelectionMarkerAttrs
+     * @emits afterGetSelectionMarkerAttrs
+     */
+    public getSelectionMarkerAttrs(chartX: number, chartY: number): {
+        attrs: SVGAttributes
+        type: 'rect' | 'arc' | 'path'
```
TorsteinHonsi, line 275: In accordance with existing naming, this should be called `shapeType`

ts/Core/Pointer.ts
```diff
@@ -262,6 +264,68 @@ class Pointer {
...
+            e,
+            (e: any): void => {
+                const zoomHor = this.zoomHor,
+                    zoomVert = this.zoomVert,
+                    mouseDownX = (this.mouseDownX || 0),
+                    mouseDownY = (this.mouseDownY || 0),
+                    chart = this.chart;
+
+                let attrs: SVGAttributes = {},
+                    svgType: 'rect' | 'arc' | 'path' = 'rect',
```
TorsteinHonsi, line 295: In accordance with existing naming, this should be called `shapeType`

TorsteinHonsi (changes requested):
Thanks! Some more comments. Trying to reduce the filesize impact on `highcharts.js`.

pawelfus (approved):
Thank you!

pawelfus:
@bre1470 could you review too? I see a few `as any` castings, but I don't have any great ideas on how to solve them

bre1470:
Yes, I will check. In the meantime please merge the latest master into this branch, so that CI jobs get updated.

bre1470 (changes requested):
Good job! 👍  See the comments for `any` replacements.

ts/Series/PolarComposition.ts
```diff
@@ -486,6 +765,54 @@ function searchPointByAngle(
     });
 }
 
+/**
+ * Trim polygonal path
+ * @private
+ */
+function trimPath(
+    path: any,
```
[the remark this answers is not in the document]
bre1470: No, wait. That is not correct. It just needs to be of type `SVGPath`. No renaming needed.

bre1470 (approved):
Thank you. Looks good! 👍

TorsteinHonsi (approved):

PR #17003, opened 2022-02-16
Fixed #16974, esc keypress kbd focus issue.

### LINKED ISSUE #16974: Focus is not getting back on the Hamburger icon after hitting escape
Expected Behaviour:
After clicking enter on Hamburger icon of the piecharts menu and then hitting on escape button, the focus should stay on the hamburger icon of the pie charts menu.

Actual Behaviour
After hitting on escape button, the focus is being highlighted on the to Pie chart and the Bar diagram

Live demo with steps to reproduce:
https://jsfiddle.net/Marshen/u39ey50L/54/

Product version
HighCharts version : v9.3.3.

Affected browser(s)
Chrome, Edge

oysteinmoseng:
Fixed #16974, esc keypress keyboard focus issue with multiple charts on the same page.
___

### DIFF
--- a/ts/Accessibility/Components/MenuComponent.ts
+++ b/ts/Accessibility/Components/MenuComponent.ts
@@ -105,8 +105,6 @@ class MenuComponent extends AccessibilityComponent {
 
     public exportButtonProxy?: ProxyElement;
 
-    public isExportMenuShown?: boolean;
-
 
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (approved):
Thanks!

PR #17004, opened 2022-02-16
Bugfix/16979 percent stacking key

### LINKED ISSUE #16979: Percent stacking: Multiple points with the same x-value
#### Expected behaviour
Correct stacking, similar to normal stacking

#### Actual behaviour
Only one point is rendered

https://jsfiddle.net/BlackLabel/umn9c0kh/

Subtask of #12529

magdalena-gut:
Fixed #16979, only one point per x value was shown when `stacking` set to `percent`.


TorsteinHonsi (approved):
Thanks, looks good to me! Pending approval by @pawelfus

pawelfus (approved):
Thanks!

PR #17016, opened 2022-02-18
feature/breadcrumbs-navigation

Izothep:
Added breadcrumbs as a part of the navigation object.


TorsteinHonsi (approved):
Thanks! I did some simplifications, I hope we preserved all cases. Regarding the handling of `navigator.breadcrumbs` options, I think that belongs in the Breadcrumbs module itself, not in Drilldown/Treemap/Sunburst.

PR #17017, opened 2022-02-18
Enhancement/12522 treegraph inherits treemap

pawellysy:


pawellysy:
@TorsteinHonsi, @pawelfus, @Izothep  
This is how much it took to convert the `treegraph` series to inherit from the `treemap` series. 

Check it out, and let me know if there is something that should be changed, or we roll back the changes (close this pr and go back to the treegraph extending organization).

ts/Series/Sunburst/SunburstSeries.ts
```diff
@@ -55,6 +53,7 @@ const {
     updateRootId
 } = TU;
 import U from '../../Core/Utilities.js';
+import SunburstNode from './SunburstNode.js';
```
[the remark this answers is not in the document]
Izothep: I think we may need, you may ignore my above comment (but you may also double-check it :))

ts/Series/Treemap/TreemapNode.ts
```diff
@@ -0,0 +1,55 @@
+import type TreemapSeries from './TreemapSeries.js';
+import type BBoxObject from '../../Core/Renderer/BBoxObject.js';
+import type TreemapPoint from './TreemapPoint.js';
+
+namespace TreemapNode {
```
Izothep, line 5: I am now wondering - may we borrow something from the NodesComposition?

ts/Series/Treemap/TreemapAlgorithmGroup.ts
```diff
@@ -19,7 +19,7 @@
  * */
 
 import type PositionObject from '../../Core/Renderer/PositionObject';
-import type TreemapSeries from './TreemapSeries';
+import TreemapNode from './TreemapNode.js';
```
Izothep, line 22: `type`?

ts/Series/Treegraph/TreegraphPointOptions.d.ts
```diff
@@ -14,15 +14,17 @@
...
 
 /* *
  *
  *  Declarations
  *
  * */
 
-export interface TreegraphPointOptions extends OrganizationPointOptions {
+export interface TreegraphPointOptions extends TreemapPointOptions {
+    borderRadius: any;
```
Izothep, line 26: Why `any`?

ts/Series/Treegraph/TreegraphPointOptions.d.ts
```diff
@@ -14,15 +14,17 @@
...
+    link: any;
```
Izothep, line 27: like above

ts/Series/Treegraph/TreegraphSeriesOptions.d.ts
```diff
@@ -25,7 +25,11 @@ import type { PointMarkerOptions } from '../../Core/Series/PointOptions';
 
 
 export type TreegraphLayoutTypes = 'Walker';
-export interface TreegraphSeriesOptions extends OrganizationSeriesOptions {
+export type linkOptions = {
```
Izothep: We are exporting empty object, is it planned?

ts/Series/Treegraph/TreegraphSeriesOptions.d.ts
```diff
@@ -25,7 +25,11 @@ import type { PointMarkerOptions } from '../../Core/Series/PointOptions';
...
+
+};
+export interface TreegraphSeriesOptions extends TreemapSeriesOptions {
+    link: any;
```
Izothep: instead of `any`, we may create `linkOptionsObject` that will contain all the options for the link. This may also solve the empty object in the `linkOptions` above

ts/Series/Treegraph/TreegraphLayout.ts
```diff
@@ -222,19 +216,19 @@ extend(H.treeLayouts.Walker.prototype, {
...
-                    link.toNode,
+                    child,
                     defaultAncestor
                 );
             });
             treeLayout.executeShifts(node);
 
-            let leftChild = node.getLeftMostChild() as TreegraphNode;
-            let rightChild = node.getRightMostChild() as TreegraphNode;
+            let leftChild = node.getLeftMostChild() as TreegraphNode.Node;
```
Izothep, line 230: Do we still need this cast as `TreegraphNode.Node`?

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -254,30 +306,13 @@ class TreegraphSeries extends OrganizationSeries {
...
-    }
-
     /**
      * Calculate `a` and `b` parameters of linear transformation, where
      * `finalPosition = a * calculatedPosition + b`.
      *
      * @return {LayoutModifiers} `a` and `b` parameter for x and y direction.
      */
-    private getLayoutModifiers(): LayoutModifiers {
+    private getLayoutModifier(): LayoutModifiers {
```
Izothep, line 307: Is this change of name for some purpose? If yes, change the documentation in comment above :)

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -342,86 +378,120 @@ class TreegraphSeries extends OrganizationSeries {
         return { ax, bx, ay, by };
     }
 
+    getLinks(): TreegraphLink[] {
+        const series = this;
+        const links = [] as TreegraphLink[];
+        this.data.forEach((point): void => {
+            if (point.node.parent) {
+                const link = new series.LinkClass().init(
```
Izothep, line 378: Aren't we able to call `linkClass init` with all the below declarations? like `link.toNode`, `link.formatPrefix` etc.

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -342,86 +378,120 @@ class TreegraphSeries extends OrganizationSeries {
...
-                levelIsConstant: true,
-                linkColor: options.linkColor, // organization series
-                linkLineWidth: options.linkLineWidth, // organization series
-                linkOpacity: options.linkOpacity,
-                states: options.states
+                levelIsConstant: series.options.levelIsConstant,
+                colorByPoint: options.colorByPoint
             }
-        });
+        }) as any;
```
Izothep, line 442: `as any` required?

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -342,86 +378,120 @@ class TreegraphSeries extends OrganizationSeries {
...
+    }
+
     /**
      * Treegraph has two separate collecions of nodes and lines,
      * render dataLabels for both sets.
      */
     public drawDataLabels(): void {
         if (this.options.dataLabels) {
-            const textPath = this.options.dataLabels.textPath;
+            const textPath = (this as any).options.dataLabels.textPath;
```
Izothep: Before, `(this as any)` was not necessary, maybe we can get rid of it now?

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -342,86 +378,120 @@ class TreegraphSeries extends OrganizationSeries {
...
             // Render node labels.
             super.drawDataLabels.apply(this, arguments);
 
             // Render link labels.
-            this.points = this.data;
-            this.options.dataLabels.textPath =
-                this.options.dataLabels.linkTextPath;
-            super.drawDataLabels.apply(this, arguments);
+            const points = this.points;
+            this.points = this.links as any;
```
Izothep, line 479: I can imagine why we need `as any` in here but maybe lets at least cast to proper points type?

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -342,86 +378,120 @@ class TreegraphSeries extends OrganizationSeries {
...
+            (this as any).options.dataLabels.textPath = (
+                this as any
```
Izothep, line 481: `as any`

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -342,86 +378,120 @@ class TreegraphSeries extends OrganizationSeries {
...
+            ).options.dataLabels.linkTextPath;
+            Series.prototype.drawDataLabels.apply(this, arguments);
 
             // Restore nodes.
-            this.points = this.points.concat(this.nodes);
-            this.options.dataLabels.textPath = textPath;
+            this.points = this.points.concat(points);
+            (this as any).options.dataLabels.textPath = textPath;
```
Izothep, line 487: again, `as any`

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -430,84 +500,106 @@ class TreegraphSeries extends OrganizationSeries {
...
+                levelOptions.borderRadius,
+                series.options.borderRadius
+            ),
+            linkColor = pick(
+                stateOptions.link && stateOptions.link.color,
+                options.link && options.link.color,
+                levelOptions.link && levelOptions.link.color,
+                series.options.link && series.options.link.color
+            ),
+            linkLineWidth = pick(
```
Izothep, line 520: I am sure we may make this code smaller, e.g. using updated strings array and forEach.

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -430,84 +500,106 @@ class TreegraphSeries extends OrganizationSeries {
...
+                stateOptions.link && stateOptions.link.lineWidth,
+                options.link && options.link.lineWidth,
+                levelOptions.link && levelOptions.link.lineWidth,
+                series.options.link && series.options.link.lineWidth
             );
-            return attrs;
+
+        if ((point as any).isLink) {
```
Izothep, line 527: `as any`

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -430,84 +500,106 @@ class TreegraphSeries extends OrganizationSeries {
...
+            }
         }
-        return super.pointAttribs.apply(this, arguments);
+        return attribs;
     }
 
+    public drawPoints(): void {
+        super.drawPoints.apply(this, arguments);
+        const points = this.points;
+        this.points = this.links as any;
```
Izothep, line 542: I am not a fan of this casting. I was even wondering if we should not create a separate method for drawing links if they have different type and functionality, instead of doing this tricky solution. @pawelfus

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -430,84 +500,106 @@ class TreegraphSeries extends OrganizationSeries {
...
-        node.dlOptions = TreegraphSeries.getDLOptions({
-            level: (this.mapOptionsToLevel as any)[node.level],
-            optionsPoint: node.options
-        });
         // Set the anchor position for tooltip.
-        node.tooltipPos = chart.inverted ?
-            [plotSizeY - y, plotSizeX - x] :
-            [x, y];
+        point.tooltipPos = chart.inverted ?
+            [plotSizeY - y, x] :
```
Izothep, line 583: It looks that something is not working correctly in here, maybe we should use `nodeX` and `nodeY`?

ts/Series/Treemap/TreemapSeries.ts
```diff
@@ -1138,7 +1099,7 @@ class TreemapSeries extends ScatterSeries {
                 renderer,
                 shadow,
                 shapeArgs,
-                shapeType: 'rect'
+                shapeType: (point.shapeType as any) || 'rect'
```
Izothep, line 1102: do we need `as any`?

Izothep (changes requested):
Good job!

Additionally to mentioned changes, it would be great if we could test:
1. tooltip functionality
2. rendering (borderWidth is being cropped at the moment)
3. collapse - links are hidden but not nodes

samples/highcharts/series-treegraph/reversed-nodes/demo.js
```diff
@@ -2,18 +2,21 @@ Highcharts.chart('container', {
...
+    series: [
+        {
+            reversed: true,
+            marker: {
+                radius: 30
+            },
+            type: 'treegraph',
+            keys: ['id', 'parent'],
+            data: [
+                ['A'],
```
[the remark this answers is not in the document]
pawellysy: Not reversed: 
image
Reversed: 
image

samples/highcharts/demo/treegraph-chart/demo.js
```diff
@@ -205,14 +215,17 @@ Highcharts.chart('container', {
...
                 {
                     id: 'Dwalin',
+                    parent: 'Fundin',
                     name: 'Dwalin',
                     description: '2772 - 3112',
                     marker: {
                         fillColor: expeditionColor
                     }
                 },
+                { id: 'Groin', parent: 'Farin' },
```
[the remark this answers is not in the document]
pawellysy: I wanted to write them in order, they are also represented on the tree. It can be changed tho. up to you

ts/Series/Sunburst/SunburstPoint.ts
```diff
@@ -63,13 +64,13 @@ class SunburstPoint extends TreemapPoint {
 
     public outerArcLength?: number;
 
-    public node: SunburstSeries.NodeObject = void 0 as any;
+    public node: SunburstNode.Node = void 0 as any;
```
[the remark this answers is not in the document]
pawellysy: whereas in the Sunburst/treemap there are not so many options, in the treegraph there is a bunch of additional functionalities, so I would try to separate it somehow. Maybe not the namespace, but a different class/ file for sure. 
up to you :)

ts/Series/Treemap/TreemapNode.ts
pawellysy: The node structure is different. Could you pinpoint some common functionalities, that could be easily transferred?

ts/Series/Treegraph/TreegraphSeries.ts
pawelfus: Cool idea, however, I'm not sure if this should be part of this PR

ts/Series/Treegraph/TreegraphSeries.ts
pawelfus: + hardcoded props can be defined in constructor:
```
                link.formatPrefix = 'link';
                link.dataLabelOnNull = true;
```

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -128,6 +129,60 @@ class TreegraphSeries extends OrganizationSeries {
...
+                 *
+                 * @sample   highcharts/series-organization/different-link-types
+                 *           Different link types
+                 *
+                 * @type {'default' | 'curved' | 'straight'}
+                 * @default 'default'
+                 * @product highcharts
+                 *
+                 */
+                type: 'curved' as const
```
pawelfus, line 172: Q: as const?

ts/Series/Treegraph/TreegraphLayout.ts
pawellysy: yes

pawelfus (changes requested):
The distribution file imports:
- OrganizationPoint
- OrganizationSeries
- TreemapPoint
- TreemapSeries

I hope I have found all places where importing could be changed to simple type importing. In case I have missed anything, please double check https://code.highcharts.local/modules/treegraph.js

ts/Series/Treegraph/TreegraphSeries.ts
pawellysy: I had some troubles with defining the types as string literals, and I found something like this:  typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions

I thought it is useful. Is there some other way?

ts/Series/Treegraph/TreegraphPoint.ts
```diff
@@ -16,10 +16,10 @@
 
 import type TreegraphPointOptions from './TreegraphPointOptions';
 import type TreegraphSeries from './TreegraphSeries';
-import OrganizationPoint from '../Organization/OrganizationPoint.js';
 import type TreegraphNode from './TreegraphNode.js';
-import NodesComposition from '../NodesComposition.js';
+import type TreegraphLink from './TreegraphLink.js';
 import U from '../../Core/Utilities.js';
+import TreemapPoint from '../Treemap/TreemapPoint.js';
```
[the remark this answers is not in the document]
pawellysy: yes

ts/Series/Treegraph/TreegraphSeries.ts
pawellysy: copied from `TreemapSeries.ts` so it might need to be resolved higher in the inheritance chain, and I didn't want to resolve all of the issues of the `treemap` at once ;) it can be changed tho if you want.

pawellysy:
https://github.com/highcharts/highcharts/pull/17017#discussion_r816933735 Why not?

ts/Core/Renderer/BBoxObject.d.ts
```diff
@@ -24,6 +24,7 @@ import type SizeObject from './SizeObject';
  * */
 
 export interface BBoxObject extends PositionObject, SizeObject {
+    centerX?: number;
```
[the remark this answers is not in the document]
pawellysy: The positions of the `dataLabels` are ok without setting the `centerX`, the issue was when checking if they are overlapping. What should be used then? `point.plotX`?

Apart from that, this property does exist but is casted with 'as any' whenever it is used. Shouldn't it be fixed as well?

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -34,10 +31,15 @@ const { getLevelOptions } = TU;
 
 import H from '../../Core/Globals.js';
 import U from '../../Core/Utilities.js';
-const { merge, pick, addEvent, stableSort, error } = U;
+const { merge, pick, addEvent } = U;
 
 import './TreegraphLayout.js';
 import Point from '../../Core/Series/Point';
+import TreemapSeries from '../Treemap/TreemapSeries.js';
```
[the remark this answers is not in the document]
pawellysy: This is the series, that I am extending. Why it is unnecessary?

ts/Series/Treemap/TreemapAlgorithmGroup.ts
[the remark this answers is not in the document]
pawellysy: sometimes we need to create a new node, so we need fill import

samples/highcharts/demo/treegraph-chart/demo.js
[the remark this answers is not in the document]
Izothep: Oh, I meant the structure:
```
{
    id: 'Groin',
    parent: 'Farin'
},
```

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -34,10 +31,16 @@ const { getLevelOptions } = TU;
...
 
 import './TreegraphLayout.js';
-import Point from '../../Core/Series/Point';
+import type Point from '../../Core/Series/Point';
+import TreemapSeries from '../Treemap/TreemapSeries.js';
+import OrganizationSeries from '../Organization/OrganizationSeries.js';
+import TreegraphLink from './TreegraphLink.js';
+import { Palette } from '../../Core/Color/Palettes.js';
+import ColumnSeries from '../Column/ColumnSeries.js';
+import { support } from 'jquery';
```
TorsteinHonsi, line 43: Accidental import

TorsteinHonsi (changes requested):

ts/Series/Treegraph/TreegraphSeries.ts
pawelfus: Because this is not the only series with this issue. If we consider all series, we may come up with a more universal solution. I simply don't want this PR to grow indefinitely, it is already hard to review all of the changes

ts/Core/Renderer/BBoxObject.d.ts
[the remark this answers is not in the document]
pawelfus: > Apart from that, this property does exist but is casted with 'as any' whenever it is used. Shouldn't it be fixed as well?

That's what I started with. The only place I see this is in XRangeSeries, which seems to me to be a workaround. Correct me if I'm wrong 😅  I see many references to `centerX`, but not in relation to dataLabels.

> The positions of the dataLabels are ok without setting the centerX, the issue was when checking if they are overlapping. What should be used then? point.plotX?

Could you point me to the place where `centerX` is used in overlapping dataLabels? I think we have `isIntersectRect` method responsible for this, which checks rectangles (x+w+width+height), not centerX

ts/Series/Treemap/TreemapAlgorithmGroup.ts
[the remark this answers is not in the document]
pawelfus: It seems to be solved (just type imported)?

ts/Series/Treegraph/TreegraphSeries.ts
pawelfus: (agreed solution: use const assertion to full object, not for every property inside)

ts/Core/Renderer/BBoxObject.d.ts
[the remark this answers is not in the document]
pawellysy: I guess, that my problem was that the links didn't have `plotX` and `plotY` properties set correctly. 
The part, where `centerX` was taken into consideration was here: https://github.com/highcharts/highcharts/blob/master/ts/Core/Series/DataLabel.ts#L198

ts/Series/Treegraph/TreegraphSeries.ts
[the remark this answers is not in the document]
pawellysy: IMO it is necessary. How else will I extend this class?

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -34,10 +31,15 @@ const { getLevelOptions } = TU;
 
 import H from '../../Core/Globals.js';
 import U from '../../Core/Utilities.js';
-const { merge, pick, addEvent, stableSort, error } = U;
+const { merge, pick, addEvent } = U;
 
 import './TreegraphLayout.js';
 import Point from '../../Core/Series/Point';
+import TreemapSeries from '../Treemap/TreemapSeries.js';
+import OrganizationSeries from '../Organization/OrganizationSeries.js';
```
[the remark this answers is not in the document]
pawellysy: IMO it is necessary. The only other solution that comes to my mind would be to move the rendering of the links to other Utils, like `LinksComposition`.
Is this what should I do?

ts/Series/Treegraph/TreegraphNode.ts
```diff
@@ -8,171 +8,171 @@
  *
  * */
 
-import TreegraphPoint from './TreegraphPoint.js';
 import type TreegraphSeries from './TreegraphSeries.js';
+import TreemapNode from '../Treemap/TreemapNode.js';
```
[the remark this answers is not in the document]
pawellysy: IMO it is necessary. How else will I extend this class?

ts/Series/Treemap/TreemapAlgorithmGroup.ts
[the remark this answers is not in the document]
pawellysy: `TreegraphNode` needs to be imported. Sometimes it is necessary to create a dummy node using the constructor of this Class. Is there another way?

ts/Series/Treegraph/TreegraphSeries.ts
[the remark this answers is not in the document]
pawelfus: Try something like this:
```ts
const {
    seriesTypes: {
        treemap: TreemapSeries,
        column: ColumnSeries
    }
} = SeriesRegistry;
````

File: ts/Series/Treegraph/TreegraphSeries.ts (hunk 29 above)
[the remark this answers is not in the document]
Comment (pawelfus) [reply ts/Series/Treegraph/TreegraphSeries.ts:39]: Personally, I think it's better than importing the whole Organization series into TreeGraph. What do you think @TorsteinHonsi ? Only one method (`translateLink`) is used in the Treegraph series here.

File: ts/Series/Treegraph/TreegraphNode.ts (hunk 30 above)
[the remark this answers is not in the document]
Comment (pawelfus) [reply ts/Series/Treegraph/TreegraphNode.ts:12]: Try this way:
```ts
import type TreegraphSeries from './TreegraphSeries.js';
import type TreemapNodeNS from '../Treemap/TreemapNode.js';
import TreegraphPoint from './TreegraphPoint.js';

const {
    seriesTypes: {
        treemap: {
            prototype: {
                NodeClass: TreemapNode
            }
        }
    }
} = SeriesRegistry;

namespace TreegraphNode {
    export interface NodeValuesObject extends TreemapNodeNS.NodeValuesObject {}
    export class Node extends TreemapNode { ... }
}
```

ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -342,86 +375,140 @@ class TreegraphSeries extends OrganizationSeries {
...
+                (this.chart.plotSizeX as any) - link.dlBox.x
+            ] : [
+                link.dlBox.x,
+                link.dlBox.y
+            ];
+            if (inverted) {
+                link.dlBox.y = 0;
+                link.dlBox.x = 0;
+                // link.dlBox.y = (this.chart.plotSizeY as any) - link.dlBox.y,
+                // link.dlBox.x =  (this.chart.plotSizeX as any) - link.dlBox.x
```
pawelfus, line 476: To do?

ts/Series/Treegraph/TreegraphPoint.ts
```diff
@@ -16,10 +16,10 @@
...
 const { extend } = U;
```
pawelfus, line 23: Unused `extend`. That means `import U ...` is redundant too.

ts/Series/Treegraph/TreegraphPoint.ts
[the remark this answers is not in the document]
pawelfus: Try something like this instead:
```ts

const {
    seriesTypes: {
        treemap: {
            prototype: {
                pointClass: TreemapPoint
            }
        }
    }
} = SeriesRegistry;
```

pawelfus (changes requested):

ts/Series/Treegraph/TreegraphSeries.ts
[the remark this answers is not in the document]
pawellysy: I rewrote this method, and put it directly in Treegraph 👍

ts/Core/Renderer/SVG/SVGRenderer.ts
```diff
@@ -869,6 +869,79 @@ class SVGRenderer implements SVGRendererLike {
...
+
+    /**
+     * General function to apply corner radius to a path
+     * @private
+     */
+
+    /* eslint-enable valid-jsdoc */
+
+
+    public curvedPath(path: SVGPath, r?: number): SVGPath {
```
pawelfus, line 882: I'm not sure if this should be part of the core. @TorsteinHonsi

pawelfus (approved):
Approved! Great work!

I'm not sure about the `curvedPath` method in Core (extra 0.2kB), but I'm leaving the decision up to Torstein

TorsteinHonsi (changes requested):
Thanks! As Pawel commented, we want to keep `curvedPath` out of the core unless it is actually used in the core.

TorsteinHonsi (approved):

pawellysy:
@bre1470 This branch was not merged to master, but to some other branch, so I guess, that it should not be added to given milestone, am I right? (This might be a bug in GH projects 😓 )

PR #17021, opened 2022-02-21
Fixed #16598: AddPoint not adding point.

### LINKED ISSUE #16598: Maps: mapbubble without lat long not working on addPoint
#### Expected behaviour
The point is added correctly as on init.

#### Actual behaviour
The point is not rendered.

https://jsfiddle.net/BlackLabel/jy0kwgeL/4/

addPoint runs in 1234ms.
The existing point changes radius, so the new point is processed to some degree but not rendered in the chart.

#### Product version + internal notes
Current (v9.3) but looks like this never worked (tested v8 and v7), so not something done through v9.3.0 map geometry update.

DJTechnoo:
Fixed #16598: `addPoint()` not adding point to map.


madepiet (approved):

ts/Series/MapPoint/MapPointSeries.ts
```diff
@@ -159,6 +161,11 @@ class MapPointSeries extends ScatterSeries {
         }
         this.generatePoints();
 
+        if (this.bounds && this.getProjectedBounds) {
```
TorsteinHonsi, line 164: Just a minor optimization - in the current state, the bounds will always be re-calculated when translate runs again. For example in this case, it is not necessary: https://jsfiddle.net/highcharts/Lw6vxy8j/

I think what we want to do is to run this only when the data has changed, so we can simply add a check for `isDirtyData`:

`if (this.bounds && this.getProjectedBounds && this.isDirtyData) {`

TorsteinHonsi (changes requested):
Thanks! See the inline comment.

raf18seb (changes requested):
1. Unfortunately, when doing a code rev, I found 2 bugs.

The first one (separate bug) is already reported: https://github.com/highcharts/highcharts/issues/17057

I was wondering about the second one because it might be related. Could you check it? Is it worth fixing it here or should we create a separate bug ticket and handle it later?
(adding point is still broken when series.data is initially empty)
http://jsfiddle.net/BlackLabel/j61z9gLe/

2. Also, the code can be shortened by removing unnecessary lines (see the comments below).

3. I was also wondering about merging those 2 tests in the `mapbubble-linewidth/demo.js` file into one test, but it's difficult because of other bugs so maybe let's keep it as it is for now.

4. Another thing: should the folder name be changed now? "mapbubble-linewidth"

ts/Series/MapPoint/MapPointSeries.ts
KacperMadej: [2nd bug from Rafał](http://jsfiddle.net/BlackLabel/j61z9gLe/) (start with empty data) suggests that checking `this.bounds` is not needed - when debugging, I changed the undefined to true to enter the if and the chart works fine. In short, if data is dirty and getProjectedBounds is available then it should run after clearing bounds (clearing undefined on an object is not raising any errors).
`if (this.getProjectedBounds && this.isDirtyData) {`
Tested locally and works fine. Doesn't resolve the problem with [processedData](https://github.com/highcharts/highcharts/issues/17057).

samples/unit-tests/maps/mapbubble-linewidth/demo.js
```diff
@@ -56,7 +56,52 @@ QUnit.test('MapBubble', function (assert) {
...
+    const chart = Highcharts.mapChart('container', {
+        chart: {
+            map: 'countries/gb/gb-all'
+        },
+        series: [{
+            type: 'map',
+            data: data,
+            dataLabels: {
+                enabled: false
+            }
```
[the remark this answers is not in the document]
DJTechnoo: Thanks I'll remove unnecessary options and look at the bugs

ts/Series/MapPoint/MapPointSeries.ts
DJTechnoo: Thank you Torstein and Kacper for the suggestions!

samples/unit-tests/maps/mapbubble-linewidth/demo.js
```diff
@@ -56,7 +56,41 @@ QUnit.test('MapBubble', function (assert) {
     assert.strictEqual(
         chart.series[1].joinBy[0],
         'hc-key',
-        `The value of joinBy in the mapbubble series 
+        `The value of joinBy in the mapbubble series
         should be 'hc-key' by default.`
     );
 });
+
+QUnit.test('AddPoint', function (assert) {
```
TorsteinHonsi, line 64: To avoid having to create another chart instance, can we instead do the `addPoint` test on the other chart in the file? In order to keep `gulp test` time down we generally try to reuse instances whenever it makes sense.

ts/Series/MapPoint/MapPointSeries.ts
```diff
@@ -211,10 +218,12 @@ class MapPointSeries extends ScatterSeries {
  * */
 
 interface MapPointSeries {
+    bounds: MapBounds | undefined;
     pointClass: typeof MapPointPoint;
 }
 extend(MapPointSeries.prototype, {
     type: 'mappoint',
+    bounds: MapSeries.prototype.bounds,
```
TorsteinHonsi, line 226: Why? `MapSeries.prototype.bounds` is `undefined` at this stage. Since `bounds` is of type `MapBounds | undefined`, I think we can just delete this line.

TorsteinHonsi (changes requested):
Thanks, I found a couple of other minor issues.

TorsteinHonsi (approved):

KacperMadej (approved):

samples/unit-tests/maps/mapbubble-linewidth/demo.js
```diff
@@ -1,44 +1,68 @@
...
+
     assert.strictEqual(
         chart.series[1].graph['stroke-width'],
         2,
         'MapBubble with linewidth- points should have stroke width.'
     );
 
+    chart.series[1].update({
+        joinBy: ['iso_3166_2', 'name']
+    });
```
raf18seb, line 52: You can set `chart.redraw` flag to false to improve performance here

samples/unit-tests/maps/mapbubble-linewidth/demo.js
```diff
@@ -1,44 +1,68 @@
...
+
+    chart.series[1].addPoint({
+        name: 'GB-AGY',
+        z: 1000,
+        color: 'red'
+    });
+
+    const addedPoint = chart.series[1].data[4];
+    assert.ok(
+        (addedPoint.plotX !== undefined && addedPoint.plotY !== undefined),
```
raf18seb, line 62: And by the occasion, last small thing: these brackets here are unnecessary

raf18seb (changes requested):
Looks good! Last small suggestions, see comments below.

Also, should we change the folder's name from "maps/mapbubble-linewidth" to e.g. "maps/mapbubble"?

PR #17022, opened 2022-02-21
bugfix/17002-boost-dataGrouping

### LINKED ISSUE #17002: Boost module not working when using StockChart
#### Expected behaviour
Using the boost module in combination with a Highcharts StockChart increases the performance by drawing (sections) using WebGL, displaying (sections) using an HTML image tag instead of an svg.

#### Actual behaviour
The boost module doesn't kick in, and the entire chart is drawn using svg as if the boost module was never enabled resulting in a unusably slow chart.
When switching the type to the default "chart" instead of "stockChart", the performance is immediately increased and the chart is drawn as expected.

Two of the official Highcharts boost Samples from the Highcharts boost module docs:
1: https://www.highcharts.com/samples/highcharts/boost/line-series-heavy-stock
2: https://www.highcharts.com/samples/highcharts/boost/line-series-heavy-dynamic

Link to the docs: https://www.highcharts.com/docs/advanced-chart-features/boost-module#samples

#### Product version
NPM package Highcharts - Version: 9.3.3
Affected chart type: Highstock / stockChart

#### Affected browser(s)
All

karolkolodziej:
Fixed #17002, changed boost documentation and adjusted stock demos.

### DIFF
--- a/docs/advanced-chart-features/boost-module.md
+++ b/docs/advanced-chart-features/boost-module.md
@@ -44,7 +44,7 @@ Configuration Data Options
 
 In boost mode, [turbo mode](https://api.highcharts.com/highcharts/plotOptions.series.turboThreshold) is always turned on. That means all data points should be configured as an array of numbers (e.g. `[1, 2, 3]`) or a two dimensional array of numbers (e.g. `[ [1, 2], [2, 3], [3, 4] ]`).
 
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (approved):
Thanks!

PR #17032, opened 2022-02-24
Created icon/chart for new advantage page

nasin:
Hi, this is the static chart/ animated icon for the new Advantage page.

### DIFF
--- a/samples/highcharts/website/icon-support/demo.css
+++ b/samples/highcharts/website/icon-support/demo.css
@@ -1,0 +1,257 @@
+@import "https://code.highcharts.com/css/highcharts.css";
+
+.chart-container .thick {
+    stroke-width: 1;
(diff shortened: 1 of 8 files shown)


oysteinmoseng:
Hi @nasin, took a quick a11y look - will these animate in any way, or could they even just be static SVG files? Wondering if we want to enable tooltip at all for these.

nasin:
Hi,

For the animated life preserver icon, I removed the tooltips and put logic in the animation to handle reduced motion prefered.

For the static word cloud, I made the tooltips more meaningful instead of showing occurrences, which were mostly 1s. I added the a11y module to chart as well.

I'm not sure what you mean about using static SVG files.

TorsteinHonsi (approved):
Thanks!

PR #17050, opened 2022-03-03
bugfix/17042-panning-ordinal

### LINKED ISSUE #17042: OrdinalAxis: panning to the left is broken in full view
#### Expected behaviour
When the xAxis is spanned to full range and you try to pan the chart to the left, nothing should happen

#### Actual behaviour
The extremes change to the smallest range near max values.

hold the mouse button down and move the mouse to the right, like you would like to see smaller values of the xAxis.

#### Product version
v9.3.3.
#### Affected browser(s)
all

karolkolodziej:
Fixed #17042, panning on ordinal axis when whole data visible caused an unnecessary change of extremes.


TorsteinHonsi:
Thanks! Pending review by @pawelfus

pawelfus (approved):
Thanks! 👍

PR #17085, opened 2022-03-14
bugfix/16782-setdata-unsorted-joinby

### LINKED ISSUE #16782: Highmaps - `joinBy` is not respected in series update
#### Expected behaviour
Workaround with an empty point on init: https://jsfiddle.net/BlackLabel/j3vb21qm/

#### Actual behaviour
Points are not matched by `hc-key`.

1. Open jsfiddle: https://jsfiddle.net/BlackLabel/j2qrk1pL/
2. Wait one second
3. Check the most right-bottom point - it is Hessen and should be Bayern

#### Product version
Highmaps 9.3

#### Affected browser(s)
All

_Internal note:_
Regression bug since Highmaps 9.3

raf18seb:
Fixed #16782, point's `joinBy` was wrong when updating data with an unsorted order.

### DIFF
--- a/ts/Core/Chart/Chart.ts
+++ b/ts/Core/Chart/Chart.ts
@@ -3382,7 +3382,9 @@ class Chart {
         if (updateAllSeries) {
             chart.getSeriesOrderByLinks().forEach(function (series): void {
                 // Avoid removed navigator series
-                series.update({}, false);
+                if (series.chart) {
+                    series.update({}, false);
+                }
             }, this);
         }
 
--- a/ts/Core/Chart/Chart.ts
+++ b/ts/Core/Chart/Chart.ts
@@ -102,7 +102,6 @@ const {
     fireEvent,
     getStyle,
     isArray,
-    isFunction,
     isNumber,
     isObject,
     isString,
@@ -3383,9 +3382,7 @@ class Chart {
         if (updateAllSeries) {
             chart.getSeriesOrderByLinks().forEach(function (series): void {
                 // Avoid removed navigator series
-                if (series.chart) {
-                    series.update({}, false);
-                }
+                series.update({}, false);
             }, this);
         }
 
--- a/samples/unit-tests/maps/globalmap/demo.js
+++ b/samples/unit-tests/maps/globalmap/demo.js
@@ -87,6 +87,27 @@ QUnit.test('Set basemap on chart object', function (assert) {
         a null color, #11636.`
     );
 
+    assert.strictEqual(
+        chart.series[0].points[0].options.name,
+        'Temburong',
+        `The first point from a map source should be bn-te, its name should
+        be taken from a map source.`
+    );
+
+    chart.series[0].update({
+        dataLabels: {
+            enabled: true,
+            format: '{point.hc-key}'
+        }
+    });
+
+    assert.strictEqual(
+        chart.series[0].points[0].dataLabel.textStr,
+        'bn-te',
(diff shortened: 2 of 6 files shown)


DJTechnoo (approved):

TorsteinHonsi:
Thanks! Can you explain why sorting matters? I would expect the `joinBy` logic to make sure each new `data` option to be linked to existing points (from `mapData`) regardless of its position in the array?

raf18seb:
Sure!

If `series.data` is an empty array initially, then map points are created from the map source in a given order.
Then, when setting a real data with a different order than source map data but **with the same data.length**, it reaches this part of code: https://github.com/highcharts/highcharts/blob/eb5a3f607eafd088fd033cee4af2a6b8b48adf0e/ts/Core/Series/Series.ts#L1216-L1223

Here the order in `oldData` and `data` is different and points are mixed up.

The first solution that came to my mind was to set `updatePoints` flag to `false` in `setData(data, ...)`, but it doesn't solve a case with `chart.update({ series: [{ data: newData }] });`

Then I was thinking about adding new condition here to this `if` statement: https://github.com/highcharts/highcharts/blob/eb5a3f607eafd088fd033cee4af2a6b8b48adf0e/ts/Core/Series/Series.ts#L1342-L1351
but `dataLength !== oldDataLength` doesn't make sense, and it touches Highcharts code not related to Maps.

So I decided to simply sort the data by `joinBy` manually.

I'm open to other solution if there is a better one

TorsteinHonsi:
Thanks for the explanation! The order or keys I think is a fragile metric - as you can see it falls apart if we add one mismatch: https://jsfiddle.net/highcharts/qLsb9n5t/

I'd rather go with your second proposal, to extend the condition for when to run `this.updateData`. If we add `!series.processedData` it works. As you say it touches Highcharts code not related to Maps. And as you know I am probably the cheapest person on the planet when it comes to bloating the core, but this is only a few bytes. We do the same for `!series.hasGroupedData` and `!series.isSeriesBoosting`. 

In fact we could probably avoid touching the core by adding something like this to the MapSeries (and map bubble?):

```ts
public updateData(data) {
    if (this.processedData) {
         return false;
    }
    super();
}
```

raf18seb:
Nice, thanks!

It's working fine, but it revealed another problem with setData for mapbubble with `updatePoints: false`.
Reported as a separate bug here: https://github.com/highcharts/highcharts/issues/17112

KacperMadej (approved):

ts/Core/Chart/Chart.ts
```diff
@@ -3383,9 +3382,7 @@ class Chart {
         if (updateAllSeries) {
             chart.getSeriesOrderByLinks().forEach(function (series): void {
                 // Avoid removed navigator series
-                if (series.chart) {
-                    series.update({}, false);
-                }
+                series.update({}, false);
```
TorsteinHonsi, line 3385: Why was the check for `series.chart` removed? What consequences may it have? The comment (which is now dangling) describes a removed navigator series.

TorsteinHonsi (commented):
Check out the removed guard.

ts/Core/Chart/Chart.ts
raf18seb: Sorry, I rushed with it. I saw an unusual `if` check, I thought it was unnecessary and removed it. Don't know the consequences :(
Already restored.

TorsteinHonsi (approved):

PR #17087, opened 2022-03-14
Fixed #9687, vertical data label suppressed for last column

### LINKED ISSUE #9687: Vertical datalabel suppressed for last column
#### Expected behaviour
All categories in a column chart should have data labels when there is enough space to render them.

#### Actual behaviour
The last data label in a column chart gets suppressed when data labels are rotated. This happens when you limit the width of the chart. The new chart width leaves enough space to render the data label.

 http://jsfiddle.net/kzoon/2bd8vomc/1/

#### Product version
Highcharts v6, v7

#### Affected browser(s)
All

TorsteinHonsi:
Fixed #9687, vertical data label suppressed when close to the right edge of the plot area


raf18seb:
Works fine for `rotation: 270`, but still fails for other values, see http://jsfiddle.net/BlackLabel/myjrLvg6/

TorsteinHonsi:
Ready for new review @raf18seb

raf18seb (approved):
Looks good, especially visual test! And wow to the `getBBox` changes.

TorsteinHonsi:
Thanks! It still needs some work with dynamic content though: https://jsfiddle.net/highcharts/4kb7Loew/

The problem is in these lines - to get the actual position of the box when redrawing:

https://github.com/highcharts/highcharts/blob/2d4472a21657381834b57707fce11be157db898c/ts/Core/Series/DataLabel.ts#L333-L335

PR #17095, opened 2022-03-16
bugfix/16901-navigator-rangeSelector-visibility 

### LINKED ISSUE #16901: Ancestor div with a hidden visibility not hiding navigator and range selector
#### Expected behaviour
Ancestor div with a hidden visibility should hide navigator and range selector. This is useful when using Highcharts inside a tab which hides a tab content with visibility.

#### Actual behaviour
Ancestor div with a hidden visibility not hiding navigator and range selector

http://jsfiddle.net/p6ugybj8/

#### Product version
Highstock JS v9.3.2

#### Affected browser(s)
Chrome

karolkolodziej:
Fixed #16901, navigator and range selector visibility was not inherited.


TorsteinHonsi (approved):
Thanks!

PR #17102, opened 2022-03-17
Columns of column chart float above x-axis when using datetime y-axis

### LINKED ISSUE #17084: Columns of column chart float above x-axis when using datetime y-axis
#### Expected behaviour

In a column chart, it is expected that the bottom of each column starts on the x-axis.

#### Actual behaviour

In some cases, when using a datetime y-axis, the columns of a column chart float 20 or 30 pixels above the x-axis.

See the following JSFiddle: https://jsfiddle.net/robbertbrak/xnd6mwhq/1/

#### Product version

Highcharts v10.0.0, but this bug is also present in most older versions of Highcharts.

#### Affected browser(s)

Not browser dependent. It can be reproduced in Firefox, Chrome, Safari.

#### Notes

This is related to https://github.com/highcharts/highcharts/pull/9344.

The issue can be fixed by added `Math.max(1, ...)` around the `time.get('Date', ...)` expression here: https://github.com/highcharts/highcharts/blob/master/ts/Core/Time.ts#L745.

robbertbrak:
This fixes #17084.

### DIFF
--- a/ts/Core/Time.ts
+++ b/ts/Core/Time.ts
@@ -741,12 +741,14 @@ class Time {
                 time.set(
                     'Date',
                     minDate,
-                    (
-                        time.get('Date', minDate) -
-                        minDay + startOfWeek +
-                        // We don't want to skip days that are before
-                        // startOfWeek (#7051)
-                        (minDay < startOfWeek ? -7 : 0)
+                    Math.max(1,
+                        (
+                            time.get('Date', minDate) -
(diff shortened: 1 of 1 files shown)


TorsteinHonsi:
Thanks for the PR!

We need to merge it into an inhouse branch so that the CI works. We also need to add a regression test. Will you have a look at that @raf18seb?

PR #17107, opened 2022-03-18
bugfix/15735-remove-indicator

### LINKED ISSUE #15735: Stock Tools: Adding and removing panes misplaces yAxis with `top` predefined
#### Expected behaviour
Correctly positioned first/main yAxis

#### Actual behaviour
yAxis is translated to the top (negative `yAxis.top` value)

[Demo](https://jsfiddle.net/BlackLabel/kam2zwbg/1/)

- initial yAxis has `yAxis.top` defined.
- add 2 indicators with separate yAxes (for example 2x ATR)
- remove the first of them -> yAxis height is calculated incorrectly:

Zrzut ekranu 2021-05-17 o 18 00 00

#### Product version
Highcharts Stock

pawellysy:
Fixed #15735, axes were wrongly calculated when removing an indicator via Stock Tools GUI.


ts/Stock/StockToolsBindings.ts
```diff
@@ -348,9 +354,12 @@ bindingsUtils.manageIndicators = function (
...
 
             if (indicatorsWithAxes.indexOf(series.type) >= 0) {
-                const removedYAxisHeight = yAxis.options.height;
+                const removedYAxisProps = {
+                    height: yAxis.options.height,
+                    top: yAxis.options.top
+                } as Highcharts.AxisPositionsObject;
                 yAxis.remove(false);
-                navigation.resizeYAxes(removedYAxisHeight as string);
+                navigation.resizeYAxes(removedYAxisProps);
```
karolkolodziej, line 362: There is something wrong with `AxisPositionsObject` interface declaration if you still need to cast it here like this.

karolkolodziej (changes requested):
- [x] One TS issue to fix

ts/Stock/StockToolsBindings.ts
```diff
@@ -348,9 +354,12 @@ bindingsUtils.manageIndicators = function (
             series.remove(false);
 
             if (indicatorsWithAxes.indexOf(series.type) >= 0) {
-                const removedYAxisHeight = yAxis.options.height;
+                const removedYAxisProps = {
+                    height: yAxis.options.height,
+                    top: yAxis.options.top
+                } as Highcharts.AxisPositionsObject;
```
karolkolodziej, line 360: There is something wrong with `AxisPositionsObject` interface declaration if you still need to cast it here like this.

karolkolodziej (changes requested):

ts/Stock/StockToolsBindings.ts
karolkolodziej: Additionally, I think there is no need for adding the `Object` word into that interface's name.

pawellysy: As you can see below, the `removedYAxisHeight` was previously casted as well (to string).  Since we can only remove the axis we previously created by GUI, these options are certainly strings, so I decided to cast them here, since this is not an issue.

raf18seb (approved):
Nice!

karolkolodziej (approved):

TorsteinHonsi:
Pending approval by @pawelfus

pawelfus (changes requested):
Just some minor changes in doclets, the rest looks great, thanks!

samples/highcharts/cypress/custom-indicator-stock-tools-gui/demo.details
```diff
@@ -0,0 +1,10 @@
+---
+name: Demo which is a base for cypress tests
+skipTest: true
+js_wrap: b
+tags:
+  - Highcharts Stock demo
+categories:
+  - General:
+      priority: 11
```
pawelfus, line 9: No need for tags and categories in the test

pawelfus (changes requested):

pawelfus (approved):
Thanks!

PR #17126, opened 2022-03-24
bugfix/15863-remove-hide-by-translation

### LINKED ISSUE #15863: convert SVG in PowerPoint to shapes for Polar Chart is very large and off the chart
We are implementing the new highcharts and something we use is to convert SVG in PowerPoint to shapes.
This works for most Highchart graphs but we have an issue with the polar chart. 

https://www.highcharts.com/demo/polar
Export to SVG and paste in PowerPoint and convert to Shapes but then it looks not correct and one shape is too big.

#### Product version
Highcharts 9.1.1

pawellysy:
Part of #15863, the scrollbar is no longer initially rendered with `y: -9999`.


ts/Core/Scrollbar.ts
```diff
@@ -596,6 +596,10 @@ class Scrollbar {
         let xOffset = height,
             yOffset = 0;
 
+        // Make the scrollbar visible when it is repositioned, #15763.
+        scroller.group.attr({
+            visibility: 'visible'
```
TorsteinHonsi, line 601: Use the shorter `.show()` function.

ts/Core/Scrollbar.ts
```diff
@@ -661,8 +665,8 @@ class Scrollbar {
             size = scroller.size,
             styledMode = scroller.chart.styledMode,
             group = renderer.g('scrollbar').attr({
-                zIndex: options.zIndex,
-                translateY: -99999
+                visibility: 'hidden',
```
TorsteinHonsi, line 668: In this case too it will probably be better to use the shorthand `.hide()` function. It is chainable.

TorsteinHonsi (changes requested):
Thanks! Remarks regarding the `show` and `hide` shorthand functions.

pawelfus (approved):
Thanks!

ts/Core/Scrollbar.ts
```diff
@@ -660,10 +662,12 @@ class Scrollbar {
...
-            group = renderer.g('scrollbar').attr({
-                zIndex: options.zIndex,
-                translateY: -99999
-            }).add();
+            group = renderer.g('scrollbar')
+                .attr({
+                    zIndex: options.zIndex
+                })
+                .add()
+                .hide(); // initially hide the scrollbar #15863
```
TorsteinHonsi, line 670: If possible, I think it is better to hide it before add. That would prevent rendering and a potential flash of visible content.

TorsteinHonsi (changes requested):

TorsteinHonsi (approved):

PR #17129, opened 2022-03-25
bugfix/fix-cypress-test

pawellysy:
Fixed the cypress tests, which are based on the current data, changing the extremes and having an impact on the actual tests.


karolkolodziej (approved):

TorsteinHonsi (changes requested):
Thanks! 
* I think we should move the tests to `/samples/highcharts/cypress/**/demo.js`, to that we can see them in the utils.
* In demo.details, remove `requiresManualTesting`, because we now have `skipTest: true`
* In demo.details, can we remove font-awesome? We probably don't need to load this external resource for cypress to run?
* In stock-tools-gui, we can probably simplify the data set a lot, because we're not running tests against the data, right? So three or four data points would be enough, just to make the extremes right?

TorsteinHonsi (approved):

PR #17130, opened 2022-03-25
Provide additional info to tooltip positioner

dmitrygorbenko:
It would be very nice to provide the point itself to the tooltip positioner because in that way customers may provide extra information within series data array elements.
In my case we desparately need this ability, because there is no other way to get a link between positioner and the point.
You had this in previous releases (at least on 6.3.1 which we are currently using), and no it's gone.

### DIFF
--- a/ts/Core/Tooltip.ts
+++ b/ts/Core/Tooltip.ts
@@ -1160,7 +1160,8 @@ class Tooltip {
                         plotY: y,
                         negative: point.negative,
                         ttBelow: point.ttBelow,
-                        h: anchor[2] || 0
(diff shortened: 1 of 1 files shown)


TorsteinHonsi:
Thanks for the PR!

However, I think this is redundant because we have that information in `this.chart.hoverPoint`, and for shared tooltips in `this.chart.hoverPoints`. See https://jsfiddle.net/highcharts/mq1n5c9d/. 

We should add that information to the docs.

dmitrygorbenko:
hm.. . thank you !
And sorry for taking your time by opening MR.

PR #17131, opened 2022-03-28
bugfix/17105-mapline-linewidth-hover-state

### LINKED ISSUE #17105: lineWidth hover state for the mapLine series
# Description of the feature
Adding the interact capability/setting the `lineWidth` for the mapLine series from the `states.hover`.
https://api.highcharts.com/highmaps/series.mapline.states.hover

Highcharts Map

# Live example for the feature
https://jsfiddle.net/BlackLabel/Lo6cbks1/

----

madepiet:
Fixed #17105, issues with mapline `lineWidth` and map `borderWidth` options and state settings.

~~Fixed: #17105, mapline linewidth hover state.~~


ts/Series/MapLine/MapLineSeries.ts
```diff
@@ -105,12 +106,24 @@ class MapLineSeries extends MapSeries {
...
             point,
             state
         );
 
+        if (state === 'hover') {
+            if (point.series.options.states &&
+                point.series.options.states.hover) {
+                let lineWidth = point.series.options.states.hover.lineWidth;
+
+                if (mapView && lineWidth) {
```
KacperMadej, line 121: `lineWidth` should work with`0`. I see no practical usage for this, but in general, the value of `0` should be allowed.

samples/unit-tests/maps/mapline-linewidth/demo.js
```diff
@@ -24,3 +24,120 @@ QUnit.test(
         );
     }
 );
+
+(async () => {
+    const mapData = await fetch(
+        'https://code.highcharts.com/mapdata/countries/gb/gb-all.topo.json'
+    ).then(response => response.json());
```
pawellysy, line 31: If you are using `async await` isn't it better to write like below?
```js
const response = await fetch(...),
     mapData = await response.json();
```

DJTechnoo (approved):

TorsteinHonsi (changes requested):
Thanks for the PR!

If we go a bit deeper than the reported issue, we see that the same problem applies to `map.borderWidth` as with `mapline.lineWidth`: https://jsfiddle.net/highcharts/16cuxe40/ . 

Another thing with the suggested solution is that it is hardwired to look for `hover`, so it won't support other states (namely `select`) like we would expect.

Both these things suggest that we must look for a solution one level up in the inheritence chain, since mapline inherits the map series. The mapping of `borderWidth` and `lineWidth` to `stroke-width` is defined by the `pointAttrToOptions` property. And for generic handling of state options we can look to `ColumnSeries.pointAttribs`. All in all, I found that this is a working version of `MapSeries.pointAttribs`:

```ts
    public pointAttribs(
        point: MapPoint,
        state?: StatesOptionsKey
    ): SVGAttributes {
        const { mapView, styledMode } = point.series.chart;
        const attr = styledMode ?
            this.colorAttribs(point) :
            ColumnSeries.prototype.pointAttribs.call(
                this, point as any, state
            );
        const strokeWidthOptionName = (
            this.pointAttrToOptions &&
            (this.pointAttrToOptions as any)['stroke-width']
        ) || 'borderWidth';

        // Individual stroke width
        let pointStrokeWidth = (point.options as any)[
            strokeWidthOptionName
        ];

        // Handle state specific border or line width
        if (state) {
            const stateOptions = merge(
                (this.options as any).states[state],
                point.options.states &&
                (point.options.states as any)[state] ||
                {}
            );
            pointStrokeWidth = stateOptions[strokeWidthOptionName];
        }

        if (pointStrokeWidth && mapView) {
            pointStrokeWidth /= mapView.getScale();
        }

        // In order for dash style to avoid being scaled, set the transformed
        // stroke width on the item
        if (
            attr.dashstyle &&
            mapView &&
            (this.options as any)[strokeWidthOptionName]
        ) {
            pointStrokeWidth = (this.options as any)[strokeWidthOptionName] /
                mapView.getScale();
        }

        attr['stroke-width'] = pick(
            pointStrokeWidth,
            // By default set the stroke-width on the group element and let all
            // point graphics inherit. That way we don't have to iterate over
            // all points to update the stroke-width on zooming.
            'inherit'
        );

        return attr;
    }
```

ts/Core/Animation/Fx.ts
```diff
@@ -521,6 +521,19 @@ class Fx {
...
+        this.elem.attr(
+            'stroke-width',
+            (
+                typeof this.start !== typeof this.end ?
+                    this.end : this.now
+            ) as any,
+            void 0,
+            true
+        );
+    }
```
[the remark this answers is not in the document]
TorsteinHonsi: PS: Animation fails because it tries to animate from `inherit` to a number, or back. So we need to deal with this upstream, probably in the MapSeries/MaplineSeries. I haven't dug further in, but one thought would be to go into `setState` and change the starting/ending values from `inherit` to a real number. Or maybe overwrite `point.graphic.animate` itself so that it replaces `stroke-width: inherit` with `lineWidth`/`scale`.

TorsteinHonsi (changes requested):
So far so good, now the [animation](https://github.com/highcharts/highcharts/pull/17131#discussion_r854080551) must be dealt with.

ts/Core/Animation/Fx.ts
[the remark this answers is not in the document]
KacperMadej: @TorsteinHonsi  Overwriting `animate` on point's graphic could resolve `"inherit"` as the end state of the animation easily because this will be one of the arguments. For `"inherit"` as the start, `stroke-widthGetter` could be overwritten because it would be used in https://github.com/highcharts/highcharts/blob/d5fa45269c7735a9430a26ee6b8255755c603418/ts/Core/Animation/AnimationUtilities.ts#L211 - unless the usage of the AnimationUtilities's `animate` should be replaced inside of point.graphic.animate by a new function?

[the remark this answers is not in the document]
TorsteinHonsi: @KacperMadej For `inherit` as the start, it should be enough to set `graphic['stroke-width'` statically, because `.attr()` will attempt to read it directly. [Working POC here](https://jsfiddle.net/highcharts/16cuxe40/1/).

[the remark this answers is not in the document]
KacperMadej: Should the static change to the numeric value of the 'inherit' be done only when strokeWidth is animating to keep the 'inherit' value after animation (so on zoom the strokeWidth will be updated correctly from the group's strokeWidth property) ? [Demo](https://jsfiddle.net/BlackLabel/he5am6kr/).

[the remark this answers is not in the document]
TorsteinHonsi: @KacperMadej Absolutetly. It seems my POC is buggy in this regard.

ts/Series/Map/MapSeries.ts
```diff
@@ -682,6 +682,49 @@ class MapSeries extends ScatterSeries {
...
+                        animate = point.graphic.animate;
+
+                    graphic.animate = function (params,
+                        options, complete): SVGElement {
+
+                        let switchBack = false;
+                        // When strokeWidth is animating
+                        if (params['stroke-width']) {
+                            const inheritedStrokeWidth = (
+                                (point.series.options.borderWidth || 0) /
```
TorsteinHonsi, line 697: By hard coding this to `borderWidth`, we are missing the `lineWidth` option of the mapline (so it falls back to 0). Use the `pointAttrToOptions` approach that we use in `pointAttribs`.

And about that - since we're now doing that pattern three times in MapSeries, consider refactoring it out to a function. It can be local to the module if it works for both series types.

ts/Series/Map/MapSeries.ts
```diff
@@ -682,6 +682,49 @@ class MapSeries extends ScatterSeries {
 
             // Add class names
             this.points.forEach((point): void => {
+                if (point.graphic) {
```
TorsteinHonsi, line 685: We already had an `if (point.graphic) {` below, so we should just combine them. And use the local `graphic` variable below too.

TorsteinHonsi (changes requested):
Thanks, getting close now!

samples/unit-tests/maps/mapline-linewidth/demo.js
```diff
@@ -24,3 +24,45 @@ QUnit.test(
         );
     }
 );
+
+QUnit.test(
```
TorsteinHonsi, line 28: Instead of adding a new test, please combine this one with the previous one (setting `lineWidth` to 30). And I think instead of loading the US map for the other one, we should just add that simple assertion into this new test, which is slimmer.

samples/unit-tests/maps/mapline-linewidth/demo.js
```diff
@@ -24,3 +24,45 @@ QUnit.test(
...
+            },
+            data: [{
+                id: 'one - two',
+                path: [["M", 5, 1], ["Q", 2.9268292682926833, 7.8048780487804885, 1, 15]]
+            }]
+        }, true, false);
+
+        chart.series[1].points[0].setState('hover');
+
+        assert.strictEqual(
```
TorsteinHonsi, line 63: We need to test that the setting is actually applied to the SVG as well, not just the options structure. In the [previous test](https://github.com/highcharts/highcharts/blob/251c6e4eb428ff6a317acdf1dcec7944f79224af/samples/unit-tests/maps/mapline-linewidth/demo.js#L17-L24) you can see how to do that.

TorsteinHonsi (changes requested):
In addition to the inline comments, I realized that we also fixed hover state for the vanilla `map` series, so we need to add a test for that too. Use an existing unit test, then add a hover `borderWidth` and assert it the same way we did here.

TorsteinHonsi (approved):
Thanks 🎉

PR #17134, opened 2022-03-28
Tools: used absolute paths for cypress articfacts

goransle:
Fixes Chrome Cypress artifacts not being saved.

There is still an issue with the Firefox videos failing to process: 
image


TorsteinHonsi (approved):
Thanks!

PR #17135, opened 2022-03-28
bugfix/15863-axis-title-translation

pawellysy:
Part of #15863, Removed Axis title translation to `y: -9999`.

### DIFF
--- a/samples/unit-tests/axis/title/demo.js
+++ b/samples/unit-tests/axis/title/demo.js
@@ -389,6 +389,7 @@ QUnit.test('Axis title multiline', function (assert) {
         'The title should be on a single line'
     );
     assert.ok(chart.plotWidth > crammedPlotWidth, 'Plot width increased');
+
 });
 
 // Highcharts 4.0.1, Issue #3027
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (commented):
Thanks! 

Removing code is risky... Are we sure we know why it was there and that its function is preserved? It was [introduced here](https://github.com/highcharts/highcharts/commit/9d1a98f56494f981420d4e4d126fbf33fc90b2b1).

pawellysy:
@TorsteinHonsi thanks for the bisect, I didn't think about it. I was hoping that visual diff would show some issues. I will investigate the problem in 3d charts

TorsteinHonsi:
> I was hoping that visual diff would show some issues

The tests are a safety net, but they can't be fully trusted for all possibilities. When dealing with some code, especially removing code, we also need to try to understand what the code does and what may possibly break.

pawelfus (approved):
Thanks!

pawellysy:
@TorsteinHonsi I restored the code I deleted and replaced the translation with `SVGElement.hide()`. However, I didn't manage to get the method `getTitlePosition` to return `NaN` in any configuration (for 3D charts also) apart from the faulty config e.g. `axis.left: 'some string'`. Could you please take a look, and if some condition comes to your mind, let me know. Thanks!

TorsteinHonsi:
Okay, then I suppose we can take the change on removing it, and if it pops up again, we'll know what to do and create a regression test for it? We don't want to keep dead code either.

TorsteinHonsi (approved):

PR #14344, opened 2020-10-13
bugfix/14320-export-data-table-update

karolkolodziej:
Fixed #14320, the table was not re-rendered after a data update.

PR #17137, opened 2022-03-29
bugfix/14320-export-data-table 

### LINKED ISSUE #14320: Export data table update issue

karolkolodziej:
Fixed #14320, the table was not re-rendered after a data update.

Replaces https://github.com/highcharts/highcharts/pull/14344

### DIFF
--- a/ts/Extensions/ExportData.ts
+++ b/ts/Extensions/ExportData.ts
@@ -462,13 +462,9 @@ addEvent(Chart, 'render', function (): void {
         this.options &&
         this.options.exporting &&
         this.options.exporting.showTable &&
-        !this.options.chart.forExport &&
-        !this.dataTableDiv
+        !this.options.chart.forExport
     ) {
         this.viewData();
-    } else if (this.dataTableDiv) {
-        // Data table was not re-rendered after a data update, #14320.
-        AST.setElementHTML(this.dataTableDiv, this.getTable());
     }
 });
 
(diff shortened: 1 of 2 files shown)


ts/Extensions/ExportData.ts
```diff
@@ -466,6 +466,9 @@ addEvent(Chart, 'render', function (): void {
         !this.dataTableDiv
     ) {
         this.viewData();
+    } else if (this.dataTableDiv) {
+        // Data table was not re-rendered after a data update, #14320.
+        AST.setElementHTML(this.dataTableDiv, this.getTable());
```
TorsteinHonsi, line 471: This seems a bit out of place, especially since we're otherwise doing the `setElementHTML` in the `toggleData` function. I found that the table successfully updates if I just remove the `!this.dataTableDiv` part of the condition:

```ts
addEvent(Chart, 'render', function (): void {
    if (
        this.options &&
        this.options.exporting &&
        this.options.exporting.showTable &&
        !this.options.chart.forExport
    ) {
        this.viewData();
    }
});
```

TorsteinHonsi (changes requested):
Thanks, see the inline comment.

TorsteinHonsi (approved):

Issue #17091, opened 2022-03-15
TextPath DataLabel stays visible for invisible points


pawellysy:
#### Expected behaviour
When point is not visible, the `dataLabel` with `textPath`  enabled should not be visible.

#### Actual behaviour
The `dataLabel` stays visible.

#### Live demo with steps to reproduce
The dataLabel with text `Brazil -> Portugal` stays visible in a place, where the Link should be
https://jsfiddle.net/BlackLabel/2a10brgL/2/

#### Product version
v10.0.0

#### Affected browser(s)
all

### Internal Note:

The `dataLabels` are hidden by translation here, which doesn't make sense for textPath
  https://github.com/highcharts/highcharts/blob/master/ts/Core/Series/DataLabel.ts#L387

PR #17143, opened 2022-03-31
bugfix/15863-dataLabel-hide-by-translation

pawellysy:
Part of #15863, DataLabels are no longer hidden by translation. 
Closes https://github.com/highcharts/highcharts/issues/17091, translate to -9999 didn't affect the text path data labels.


ts/Extensions/OverlappingDataLabels.ts
```diff
@@ -268,7 +268,10 @@ Chart.prototype.hideOverlappingLabels = function (
                 box2 &&
                 label1 !== label2 && // #6465, polar chart with connectEnds
                 label1.newOpacity !== 0 &&
-                label2.newOpacity !== 0
+                label2.newOpacity !== 0 &&
+                // #15863 dataLabels are no longer hidden by translation
+                label1.attr('visibility') !== 'hidden' &&
+                label2.attr('visibility') !== 'hidden'
```
pawelfus, line 274: With given O(N2) detection complexity, I would calculate and store `visibility` of the label before. Otherwise, on every iteration, we reach DOM twice (`attr('optionName')` getter does it).

Perhaps we can filter out unnecessary labels, or don't collect them at all? 
(2mins later): Just checked other labelCollectors, I think it's better to filter out them in this method, for example before sorting?

ts/Extensions/OverlappingDataLabels.ts
```diff
@@ -109,9 +109,13 @@ addEvent(Chart, 'render', function collectAndHide(): void {
                                 point.shapeArgs && point.shapeArgs.height
                             ); // #4118
 
-                            if (!options.allowOverlap) {
+                            if (
+                                !options.allowOverlap &&
+                                label.attr('visibility') !== 'hidden'
```
pawelfus, line 114: If you decide to not push labels in the collector, then you need to do this for all labels (e.g. StackLabels, Radial, Annotations collectors). Additionally, someone must remember this in the future, to filter out hidden labels. I'm not sure what's the best. Filtering hidden labels in the `OvelappingLabels` module or preventing collecting hidden labels in the Collectors.

I will leave the decition up to @TorsteinHonsi

pawelfus (approved):
Approved, just leaving one question mark up to Torstein

ts/Extensions/OverlappingDataLabels.ts
TorsteinHonsi: PS: The stack labels already check visibility inside the collector. Which means if we do the check in the comparison itself, we can remove this:

https://github.com/highcharts/highcharts/blob/06dc2c70f9b04b32dd22857e6eabd2df0645f15b/ts/Extensions/OverlappingDataLabels.ts#L74

TorsteinHonsi (approved):

PR #17144, opened 2022-03-31
bugfix/17057-map-addpoint-redraw

### LINKED ISSUE #17057: Map addPoint isn't working after series.update()
#### Expected behaviour
A red point should be added

#### Actual behaviour
A red point is not added because there `series.processedData` is `undefined` and this condition is not `true`: https://github.com/highcharts/highcharts/blob/496367a50fe0675397bde348fd2a477f32e704f6/ts/Core/Series/Series.ts#L3980-L3985

https://jsfiddle.net/BlackLabel/t8dczbLx/

#### Product version
Highcharts Maps v9.3.3 on master (before v10)

Regression after https://github.com/highcharts/highcharts/commit/91f9740ba172a3e48caac5963b1f7b6a66a645df
FYI @TorsteinHonsi

raf18seb:
Fixed #17057, `addPoint()` wasn't working correctly after `series.update({}, false);`.

### DIFF
--- a/samples/unit-tests/maps/setdata/demo.js
+++ b/samples/unit-tests/maps/setdata/demo.js
@@ -856,7 +856,9 @@ QUnit.test('Map set data with updated data (#3894)', function (assert) {
         mapView.zoom
     );
 
-    chart.series[0].setData(data);
+    const series = chart.series[0];
+
+    series.setData(data);
 
     const after = Object.assign(
         {},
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

madepiet (approved):

KacperMadej (approved):

DJTechnoo (approved):

PR #17151, opened 2022-04-04
bugfix/17139-proj4-chart

### LINKED ISSUE #17139: Highcharts 10 proj4 chart option no longer works
#### Expected behaviour
Highcharts v10 works with proj4 passed as a chart option. This was previously working in v9. The [changelog](https://www.highcharts.com/changelog/#highcharts-v10.0.0) does not mention anything about this option no longer being supported.

#### Actual behaviour
Hovering over the map causes a `r.proj4 is not a function` error.

In this example, the chart has proj4 passed in as options but when you hover over the map, you can reproduce the above error.
https://codesandbox.io/s/highcharts-react-simple-chart-forked-9k4hcq?file=/demo.jsx

- If you comment in the that sets `window.proj4`, hovering over the chart works as expected
- If you downgrade to highcharts v9.3.3, the chart hover works as expected without `window.proj4` set

#### Product version

Highcharts 10
React Highcharts 3.1

#### Affected browser(s)
All

SpoorthySrinivas:
Fixed #17139, `chart.proj4` option didn't work.

### DIFF
--- a/ts/Extensions/GeoJSON.ts
+++ b/ts/Extensions/GeoJSON.ts
@@ -384,7 +384,7 @@ Chart.prototype.transformToLatLon = function (
         sinAngle = transform.sinAngle ||
             (transform.rotation && Math.sin(transform.rotation)),
         // Note: Inverted sinAngle to reverse rotation direction
-        projected = win.proj4(transform.crs, 'WGS84', transform.rotation ? {
+        projected = proj4(transform.crs, 'WGS84', transform.rotation ? {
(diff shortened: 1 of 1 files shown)


KacperMadej (approved):

raf18seb:
Hi @SpoorthySrinivas, thanks for the PR!

I need to merge it into our inhouse branch so that the CI runs. I'll also add a regression test.

Closing the PR here and continuing here: https://github.com/highcharts/highcharts/pull/17153

PR #17153, opened 2022-04-04
bugfix/17139-chart-proj4

raf18seb:
Fixed https://github.com/highcharts/highcharts/issues/17139, `chart.proj4` option didn't work.

### DIFF
--- a/samples/unit-tests/maps/geojson/demo.js
+++ b/samples/unit-tests/maps/geojson/demo.js
@@ -1,43 +1,54 @@
 QUnit.test('GeoJSON and map transforms', assert => {
 
-    // #17139
-    const proj4 = window.proj4;
-    window.proj4 = undefined;
-
     const chart = Highcharts.mapChart('container', {
         chart: {
-            map: 'custom/world',
-            proj4: proj4 // #17139
+            map: 'custom/world'
         },
         series: [{
         }]
     });
 
+    const proj4 = window.proj4;
 
     [
-        { lon: 0, lat: 0 },
-        { lon: 0, lat: 60 },
-        { lon: 90, lat: 0 },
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (changes requested):
Thanks for the PR! 

For legacy, proj4-based projection we already have tests in `/unit-tests/maps/geojson`. So it would be better if we adapt that to this case. It runs a roundtrip projection of five points, so I think the only adaptation we need to do is to run it with `proj4` as a chart setting in addition to the global property.

madepiet (approved):

samples/unit-tests/maps/geojson/demo.js
```diff
@@ -1,8 +1,13 @@
 QUnit.test('GeoJSON and map transforms', assert => {
 
+    // #17139
+    const proj4 = window.proj4;
+    window.proj4 = undefined;
```
TorsteinHonsi, line 5: So now we have a test for the `chart.proj4` option, but we no longer have a test for `window.proj4`. We need to refactor this so both cases are covered. Move the logic to a function, then run the same set of tests. Can probably run `chart.update` in between to change the setting.

TorsteinHonsi (changes requested):
We need to test both ways of defining 'proj4'.

samples/unit-tests/maps/geojson/demo.js
raf18seb: Sure! In progress

DJTechnoo (approved):

madepiet (approved):

KacperMadej (approved):

TorsteinHonsi (approved):

Issue #16972, opened 2022-02-08
Exporting data: sortable table


KacperMadej:
# Description of the feature
Add option to sort by option and dynamically via GUI (click on a column header).

# Library related to the feature
Highcharts (export-data module)

# Proof of Concept/Live example for the feature
https://codepen.io/gvaartjes/pen/XomNea?editors=0010

----

**You can vote for this feature by adding a thumbs-up reaction to this post.**

PR #17154, opened 2022-04-04
enhancement/16972-sortable-table 

karolkolodziej:
Enhancement #16972, added sortable table.

- [x] Tests
- [x] Add classes 

Updated demo to play with: https://jsfiddle.net/BlackLabel/bdo7jx3p/


ts/Extensions/ExportData.ts
```diff
@@ -469,6 +471,71 @@ addEvent(Chart, 'render', function (): void {
     }
 });
 
+addEvent(Chart, 'afterViewData', function (): void {
+    const chart = this,
+        dataTableDiv = chart.dataTableDiv,
+        row = document.getElementsByClassName('highcharts-table-sort-row')[0],
```
TorsteinHonsi, line 477: Do we need the `highcharts-table-sort-row` classname? The top row should be selectable as `thead tr`, while the other rows are not inside a `thead`.

ts/Extensions/ExportData.ts
```diff
@@ -469,6 +471,71 @@ addEvent(Chart, 'render', function (): void {
...
+        getCellValue = function (
```
TorsteinHonsi, line 478: Can use arrow function. That also applies to most or all of the functions below.

ts/Extensions/ExportData.ts
```diff
@@ -469,6 +471,71 @@ addEvent(Chart, 'render', function (): void {
...
+            th.addEventListener('click', function (): void {
+                const thNodeList = dataTableDiv.querySelectorAll(
+                        'tr:not(.highcharts-table-sort-row)'
+                    ),
+                    thParentNodeList = th.parentNode.children,
+                    thArrayList = [] as Array<HTMLElement>,
+                    thParentArray = [] as Array<HTMLElement>;
+
+
+                [].forEach.call(
```
TorsteinHonsi, line 512: Are you just copying into an array? Then you can use destructuring:

`thArrayList = [...thNodeList]`

ts/Extensions/ExportData.ts
```diff
@@ -1147,6 +1214,9 @@ Chart.prototype.getTableAST = function (
                 }
 
                 theadChildren.push({
+                    attributes: {
+                        'class': 'highcharts-table-sort-row'
+                    },
```
TorsteinHonsi, line 1219: Probably redundant, see above

ts/Extensions/ExportData.ts
```diff
@@ -51,6 +52,7 @@ const {
 
 declare module '../Core/Chart/ChartLike'{
     interface ChartLike {
+        ascendingOrderInTable?: boolean
```
TorsteinHonsi, line 55: It would be nice to avoid more decoration on the Chart, but okay if we have no better way to put it. In the future we probably need to make a composition for all exporting and export-data stuff.

TorsteinHonsi (changes requested):
Thanks, works great!

In addition to the inline comments, I am thinking we should have class names to make it possible to mark the states of the sorting:
* A class name indicating if a column is active for ascending sorting
* ... and for descending sorting
* And perhaps one to indicate if sorting is available at all? Or it is always available now?

ts/Extensions/ExportData.ts
karolkolodziej: I'm trying to convert `NodeListOf<Element>` into a proper array that I can letter use the Array `sort` method.

I can destruct it as you suggested but it won't be possible without some strange casting but I'm ok with that.
This approach is much more readable than this odd `[].forEach` that I used.

karolkolodziej:
@TorsteinHonsi 
From now, each column can be sorted so I presume there is no need for any additional class name here.

PR #17155, opened 2022-04-04
demo/chart-inside-tooltip

madepiet:
Added demo files.

### DIFF
--- a/samples/highcharts/blog/chart-inside-tooltip/demo.html
+++ b/samples/highcharts/blog/chart-inside-tooltip/demo.html
@@ -1,0 +1,3 @@
+<script src="https://code.highcharts.com/stock/highstock.js"></script>
+
+<div id="container"></div>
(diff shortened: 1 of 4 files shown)


samples/highcharts/blog/chart-inside-tooltip/demo.html
```diff
@@ -0,0 +1,3 @@
+<script src="https://code.highcharts.com/stock/highstock.js"></script>
```
TorsteinHonsi, line 1: Loading `highstock.js` is overkill, right?

samples/highcharts/blog/chart-inside-tooltip/demo.js
```diff
@@ -0,0 +1,51 @@
...
+    Highcharts.chart('hc-tooltip', {
+        chart: {
+            type: 'pie'
+        },
+        title: {
+            text: null
+        },
+        subtitle: {
+            text: 'Chart inside tooltip'
+        },
```
TorsteinHonsi, line 11: Why a null title and an active subtitle? If it is for styling reasons, it would be more semantically correct to set a smaller font size for the title and then skip the subtitle.

TorsteinHonsi (changes requested):
Thanks!

TorsteinHonsi (approved):

PR #17156, opened 2022-04-05
Fixed #17146, HTML and inline style parsing issue

### LINKED ISSUE #17146: Invalid sanitizing of secure urls in HTML formatters in v10.x
#### Expected behaviour

What I'm trying to achieve (and what works in v9.x) is to display SVG icons as labels on the yAxis. We load our icons from an external CDN (URL looks like this: https://static-assets.dev.xxxx.com/icons/module-xyz.svg), and set this SVG as a mask for a span element, by using `mask: url()`.

#### Actual behaviour

This works fine in v9x, but breaks in v10. I assume that this has to do with the introduction of the new security feature in v10 (allowBypassFiltering), because the URL does get sanitized from 'https://static-assets.dev.xxxx.com/icons/module-xyz.svg' to just 'https'. 

What I've tried so far is to try bypass filtering, adding all kinds of allowedReferences to the AST and trying bunch of other stuff, but it all comes down to the fact that this just doesn't work in v10.

https://jsfiddle.net/dmj9syar/

As you can see there, the Icon does not get loaded and the URL does get replaced.

TorsteinHonsi:
Fixed #17146, a regression in v10 with parsing of HTML with inline style rules containing a colon

### DIFF
--- a/samples/unit-tests/ast/ast/demo.details
+++ b/samples/unit-tests/ast/ast/demo.details
@@ -1,0 +1,5 @@
+---
+ resources:
+   - https://code.jquery.com/qunit/qunit-2.0.1.js
+   - https://code.jquery.com/qunit/qunit-2.0.1.css
+...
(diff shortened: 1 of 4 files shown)


raf18seb (approved):
Looks good 👍

Issue #16810, opened 2021-12-21
The pattern fills inside each map element is upside down.


karolkolodziej:
#### Expected behaviour
Show the pattern fills correctly as it used to be previously.

#### Actual behaviour
The pattern fills inside each map element is upside down.

#### Live demo with steps to reproduce
v9.3.2: https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/maps/demo/pattern-fill-map
v9.2.2(last working version): https://jsfiddle.net/BlackLabel/6Lvd5r8n/

#### Product version
Issue introduced in v9.3.0

#### Affected browser(s)
All

PR #17158, opened 2022-04-05
Fixed: #16810, inverted images.

DJTechnoo:
Fixed: #16810, inverted image pattern on maps.


TorsteinHonsi:
@DJTechnoo It looks like in the current state of the PR, the scaling is done on the pattern in general. But we already have a hook, `Point.calculatePatternDimensions`, for adapting the pattern to individual points. Maybe we should adapt that to not only work with width and height, but also the scale transform?

At least in theory, that would be more correct because it won't flip the pattern unless it is _actually_ applied to a flipped point. With the current approach it flips the pattern regardless of where it is used.

DJTechnoo:
Thank you, I'll dive into it right away!

samples/unit-tests/accessibility/accessible-highmaps/demo.js
```diff
@@ -1,13 +1,16 @@
 QUnit.test('Basic map', function (assert) {
-    var chart = Highcharts.mapChart('container', {
+    const polishFlagURL =
```
raf18seb, line 2: The variable name without indicating a value would be better, e.g.
```js
const url = ...
```

samples/unit-tests/accessibility/accessible-highmaps/demo.js
```diff
@@ -1,13 +1,16 @@
...
+        'https://upload.wikimedia.org/wikipedia/en/1/12/Flag_of_Poland.svg';
+    const chart = Highcharts.mapChart('container', {
             chart: {
                 map: 'custom/europe'
             },
             series: [
                 {
+                    keys: ['hc-key', 'color.pattern.image'],
```
raf18seb, line 10: You have 2 keys defined, but your data point has 3 values: `['pl', polishFlagURL, 2],`

ts/Extensions/PatternFill.ts
```diff
@@ -223,6 +224,15 @@ Point.prototype.calculatePatternDimensions = function (
         if (!bBox.width || !bBox.height) {
             pattern._width = 'defer';
             pattern._height = 'defer';
+
+            // Check if map image pattern is flipped (#16810)
```
raf18seb, line 228: This comment here is repeated in 2 different places but both parts of code are different so I think one of the comments should be changed or should be more descriptive

ts/Extensions/PatternFill.ts
```diff
@@ -341,10 +351,20 @@ SVGRenderer.prototype.addPattern = function (
...
+        if (options.patternTransform) {
+            options.patternTransform += ' scale(1, -1)';
+        }
+    }
+
     if (options.patternTransform) {
         attrs.patternTransform = options.patternTransform;
     }
 
+
```
raf18seb, line 367: Unnecessary line break methinks🤔

raf18seb (changes requested):
Just a few small things, see the comments below

KacperMadej (approved):
What Rafał wrote. Other than that, looks good to me 👍

samples/unit-tests/accessibility/accessible-highmaps/demo.js
```diff
@@ -23,6 +26,13 @@ QUnit.test('Basic map', function (assert) {
         description.indexOf('Europe') > 0,
         'There is a screen reader region, and it contains "Europe"'
     );
+
+    assert.strictEqual(
+        chart.series[0].transformGroups[0].scaleY < 0,
+        // eslint-disable-next-line no-underscore-dangle
+        chart.series[0].points[1].options.color.pattern._inverted,
```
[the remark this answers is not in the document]
DJTechnoo: After trying this, ESLint wanted me to use the dotted notation again  ಥ_ಥ

raf18seb (approved):
Thanks!

TorsteinHonsi (approved):
Awesome, thanks!

TorsteinHonsi:
Now pending CI run after merging master in.

PR #17164, opened 2022-04-06
bugfix/17141-dataGrouping-multiple-series

### LINKED ISSUE #17141: "hasGroupedData" with multiple series does not return the right status
#### Expected behaviour
Returning "true" or "undefined" when calling "hasGroupedData" on a series, depending if the data is currently beeing grouped.

#### Actual behaviour
When two or more series are displayed, it always returns "true" once it returned "true" at least once.  
When there is only one series inside the graph it works as expected - returns "true" when the data is beeing grouped and "undefined" when zoomed in far enough where grouping is not applied. Also returns back to "true" when you reset the zoom and back to "undefined" when zoomed in again.

[Stackoverflow](https://stackoverflow.com/questions/71687888/highcharts-get-if-data-is-currently-beeing-grouped-with-multiple-series)
[https://jsfiddle.net/Itsearnest/ho0a4zw9/](https://jsfiddle.net/Itsearnest/ho0a4zw9/)

#### Product version
Highstock v10.0.0 

#### Affected browser(s)
Tested on Firefox, Chrome and Edge

karolkolodziej:
Fixed #17141, data grouping was not applied properly in charts with multiple series.

### DIFF
--- a/samples/unit-tests/series/datagrouping/demo.js
+++ b/samples/unit-tests/series/datagrouping/demo.js
@@ -1038,3 +1038,28 @@ QUnit.test('The dataGrouping enabling/disabling.', function (assert) {
         );
     });
 });
+
+QUnit.test('Data grouping multiple series on zoom, #17141.', function (assert) {
+    const chart = Highcharts.stockChart('container', {
+        chart: {
+            width: 500
+        },
+        series: [{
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #17165, opened 2022-04-06
bugfix/15863-axis-tick-hide-by-translation

pawellysy:
Part of #15863, Deleted hiding by translation of the ticks.

### DIFF
--- a/samples/unit-tests/axis/labels/demo.js
+++ b/samples/unit-tests/axis/labels/demo.js
@@ -219,7 +219,7 @@ QUnit.test(
         });
 
         assert.ok(
-            chart.yAxis[0].ticks[50].label.attr('y') > 0,
+            chart.yAxis[0].ticks[50].label.attr('visibility') !== 'hidden',
             '50 label is placed'
         );
 
@@ -230,7 +230,7 @@ QUnit.test(
 
         setTimeout(function () {
             assert.ok(
-                chart.yAxis[0].ticks[50].label.attr('y') < 0,
+                chart.yAxis[0].ticks[50].label.attr('visibility') === 'hidden',
                 '50 label is hidden'
             );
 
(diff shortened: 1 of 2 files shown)


pawelfus (approved):
Thanks!

TorsteinHonsi (changes requested):
Thanks!

Running `show()` without parameters sets the `visibility` to `visible`. The consequence of that is that the element will be visible even when the containing element is hidden, which is usually not what implementers want: https://jsfiddle.net/j72b510w/

The quick fix is to run `show(true)` to set the `visibility` to `inherit` instead.

But I am starting to think this is an antipattern. We have had numerous issues with this before, because it is counter intuitive to our own developers. Besides - in what scenario does an implementer want a specific SVG element to be visible when the container (or a parent group) itself is hidden? 

My suggestion is that we do a change now, and change the default of the `inherit` parameter to `true`:

https://github.com/highcharts/highcharts/blob/612c2ce41eed450e286b7fe69aedd488c68c947b/ts/Core/Renderer/SVG/SVGElement.ts#L2245-L2249

Step two will be to search up all cases of `.show()` throughout the code base, and either remove the `true` parameter, or add a `false` if we really want to force a visible element through a hidden parent (and again, in what scenario would that be?). @pawelfus what do you think?

PS: We can close this PR with the quick fix, then do a separate refactor PR.

pawelfus:
> @pawelfus what do you think?

I agree. It's so easy to accidentally use `show()` even when `show(true)` should be used.

TorsteinHonsi:
Excellent, let's do it. @pawellysy will you handle this as part of the -9999 family of fixes?

pawellysy:
@TorsteinHonsi sure!

TorsteinHonsi (approved):

Issue #16968, opened 2022-02-07
Fullscreen: events


KacperMadej:
# Description of the feature
Add API events for fullscreen open and close to e.g. allow chart update, prevent the original open or close be returning false, etc.

Since print-related events are in `chart.events` then this would be a good place to add the fullscreen related ones.

# Library related to the feature
Highcharts (exporting module)

# Proof of Concept/Live example for the feature
https://jsfiddle.net/BlackLabel/p8yasnf3/

----

**You can vote for this feature by adding a thumbs-up reaction to this post.**

PR #17175, opened 2022-04-11
feature/16968-fullscren-events

mfilipiec:
Added new fullscreen events. Now you can use [afterFullscreenClose](https://api.highcharts.com/highcharts/chart.events.afterFullscreenClose), [beforeFullscreenClose](https://api.highcharts.com/highcharts/chart.events.beforeFullscreenClose), [afterFullscreenOpen](https://api.highcharts.com/highcharts/chart.events.afterFullscreenOpen) and [beforeFullscreenOpen](https://api.highcharts.com/highcharts/chart.events.beforeFullscreenOpen) callback functions to modify chart options during toggle fullscreen mode. See #16968.


mfilipiec:
I have not prepared unit tests for this feature because it is not possible to recreate the opening of the fullscreen in these conditions.

And I have two questions, or rather advice:
- [ ] Are the names of these callback functions appropriate? 
- [ ] This arrangement in the API is ok?

image

samples/highcharts/chart/events-fullscreenopen-fullscreenclose/demo.details
```diff
@@ -0,0 +1,6 @@
+---
+ name: Highcharts Demo
+ authors:
+   - Michał Filipiec
```
raf18seb, line 4: I suggest not to use polish letters (they don't display correctly in some cases)

samples/highcharts/chart/events-fullscreenopen-fullscreenclose/demo.js
```diff
@@ -0,0 +1,43 @@
...
+                            fontSize: '18px'
+                        }
+                    }
+                });
+            }
+        }
+    },
+
+    title: {
+        text: 'Fullscreen'
```
raf18seb, line 32: Maybe `text: 'Fullscreen events'`?

samples/highcharts/chart/events-fullscreenopen-fullscreenclose/demo.js
```diff
@@ -0,0 +1,43 @@
...
+    },
+
+    subtitle: {
+        text: 'Click the context menu and choose "Fullscreen".<br> The title size will be larger'
```
raf18seb, line 36: The button's text is "View in full screen", not "Fullscreen".

Also, there's unnecessary space after `<br>`.

Also, a dot is missing at the end of the sentence

ts/Extensions/FullScreen.ts
```diff
@@ -253,12 +254,16 @@ class Fullscreen {
...
+                        fireEvent(chart, 'beforeFullscreenClose');
                         fullscreen.isOpen = false;
                         fullscreen.close();
+                        fireEvent(chart, 'afterFullscreenClose');
                     } else {
+                        fireEvent(chart, 'beforeFullscreenOpen');
                         chart.setSize(null, null, false);
                         fullscreen.isOpen = true;
                         fullscreen.setButtonText();
+                        fireEvent(chart, 'afterFullscreenOpen');
```
raf18seb, line 266: Close events are not working when pressing the exporting menu button (they work only on `escape` button).

I suggest firing events at the beginning and end of the `open` and `close` methods.

raf18seb (changes requested):
Thanks!

I have a few comments, see below.

ps. The events names are fine imho

samples/highcharts/chart/events-fullscreenopen-fullscreenclose/demo.css
```diff
@@ -0,0 +1,3 @@
+#container {
```
raf18seb, line 1: I think you should use shorter folder names. What about
```samples/highcharts/chart/fullscreen-events/...```
?

Also, do we have a folder for fullscreen demos? (does it have to be in the `/chart/` folder?)

File: ts/Extensions/FullScreen.ts [hunk 6]
```diff
@@ -368,3 +377,118 @@ addEvent(Chart, 'beforeRender', function (): void {
...
+ *
+ *  API Options
+ *
+ * */
+
+/**
+ * Fires after fullscreen close
+ *
+ * @sample highcharts/chart/events-fullscreenopen-fullscreenclose/
+ *         Fullscreen title size change
```
raf18seb, line 446: Could the description be more descriptive? See e.g. API of `chart.events.afterPrint` for a reference

raf18seb (changes requested):

samples/highcharts/chart/events-fullscreenopen-fullscreenclose/demo.css
raf18seb: Ok

ts/Extensions/FullScreen.ts
raf18seb: @mfilipiec sorry, maybe I missclicked or maybe GitHub is showing it incorrectly.

I meant that the `Fires after fullscreen close` sentence could be more descriptive, not the sample's name.

raf18seb:
Please see my comment above.

Also, please add appropriate PR labels and edit the PR description (see similar descriptions in changelog) ("feature" -> "see", and dot at the end of the sentence).

TorsteinHonsi (changes requested):
Thanks! 

* Regarding the naming of the events, the convention in Highcharts is to name the `before` part the same as (or similar to) the function name, then add an `after...` event if we need it. So in this case it would be `fullscreenOpen`/`afterFullscreenOpen` and `fullscreenClose`/`afterFullscreenClose`.
* Another thing we commonly do, is that the `before` event holds the main function body as its default event handler. This pattern allows implementers to `return false` in order to cancel the default behaviour. I don't know if it makes sense in this case, but it is something we can consider.
* The last thing we need to consider is whether we now need the _after_ events at all? In what scenario would we need to run logic after open or close as opposed to before?

All these things can be seen in the [Series.bindAxes](https://github.com/highcharts/highcharts/blob/2ecc2c0f909197aa4fcb59fe04cf9da27ccd9f9d/ts/Core/Series/Series.ts#L558-L627) function.

mfilipiec:
When doing this feature, I was based on the printing events as Kacper suggested in the ticket, so I split it into `before` and `after` callback functions. It seems to me that it is quite useful. For example, it is better to set the chart options just before open fullscreen and run some animation after it. 

If it is unnecessary I can remove `after` callback functions, but then also include the approach from the second point?

As for the first point, I will include it in the changes.

TorsteinHonsi:
> I was based on the printing events

I see. But the printing logic actually has functions called `beforePrint` and `afterPrint`, so the event names reflect the function names. 

I think we should skip the `after*` events for now, we can always add them later if we find it is needed.

TorsteinHonsi (changes requested):
Thanks! Was it intentional to put the events at the end of the respective functions? In that case they need the `after*` prefix. The convention is:

* Events in the beginning of a function is named the same as the function, and are cancellable by `return false` or `e.preventDefault()`. 
* Events at the end of a function are named  `after*`.

Also, there is an issue with the TS declarations that need to be adressed.

TorsteinHonsi (approved):

PR #17178, opened 2022-04-12
bugfix/16550-week-dataTime-format 

### LINKED ISSUE #16550: Duplicated week numbers in the x axis
Hi,

#### Expected behaviour
Unique week numbers in the x axis 

#### Actual behaviour
Duplicated week numbers in the x axis 

**Live Demo**
https://jsfiddle.net/fkdacgs8/7/
image

#### Product version
Highcharts Gantt latest 

#### Affected browser(s)
All

karolkolodziej:
Fixed #16550, the week `dateFormat` was not calculated correctly when `useUTC: false`.

### DIFF
--- a/samples/unit-tests/gantt/grid-axis/demo.js
+++ b/samples/unit-tests/gantt/grid-axis/demo.js
@@ -2167,3 +2167,32 @@ QUnit.test('slotWidth', assert => {
         '#15742: Rightmost tick slotWidth should be much smaller than the other ticks'
     );
 });
+
+QUnit.test('Week dataTime format for useUTC: false, #16550.', assert => {
+    const chart = Highcharts.ganttChart('container', {
+        chart: {
+            width: 800
+        },
+        xAxis: [{
+            tickInterval: 7 * 24 * 3600 * 1000
+        }],
+        time: {
+            useUTC: false
+        },
+        series: [{
+            data: [{
+                start: new Date("2015-04-12T01:00:00").getTime(),
+                end: new Date("2015-04-22T03:15:00").getTime()
+            }]
+        }]
+    });
+    const axis = chart.xAxis[0],
+        ticks = axis.ticks,
+        tickPositions = axis.tickPositions;
+
+    assert.notEqual(
+        ticks[tickPositions[0]].label.textStr,
+        ticks[tickPositions[1]].label.textStr,
+        'The first two axis labels should not be the same.'
+    );
+});
--- a/ts/Core/Axis/GridAxis.ts
+++ b/ts/Core/Axis/GridAxis.ts
@@ -1444,7 +1444,14 @@ dateFormats.E = function (this: Time, timestamp: number): string {
 
 // Adds week date format
 dateFormats.W = function (this: Time, timestamp: number): string {
-    const d = new this.Date(timestamp);
+    const time = this,
+        d = new this.Date(timestamp),
+        unitsToOmmit = (['Hours', 'Milliseconds', 'Minutes', 'Seconds'] as Array<Time.TimeUnitValue>);
+
+    unitsToOmmit.forEach(function (format): void { // #16550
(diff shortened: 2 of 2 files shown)


samples/unit-tests/gantt/grid-axis/demo.js
```diff
@@ -2167,3 +2167,32 @@ QUnit.test('slotWidth', assert => {
         '#15742: Rightmost tick slotWidth should be much smaller than the other ticks'
     );
 });
+
+QUnit.test('Week dataTime format for useUTC: false, #16550.', assert => {
```
TorsteinHonsi, line 2171: Since this is a bug in `dateFormats`, it is probably not necessary to build a full fledged chart to test it. See how other [dateFormat tests are built](https://github.com/highcharts/highcharts/blob/master/samples/unit-tests/time/dateformat/demo.js), and add you case there instead. 

In addition, remember that when setting `useUTC` to false, the dates will be local. This may cause different results depending on in which timezone the CI is running our tests, and would typically be a difficult thing to debug. In the other tests we deal with this by either setting the time in a different way, or by making sure the test only runs in the CET timezone (which is convenient for Norway/Poland devs).

https://github.com/highcharts/highcharts/blob/4132352299f951424d424cc2221ef093501969a0/samples/unit-tests/time/dateformat/demo.js#L39

TorsteinHonsi (changes requested):
Thanks! The fix is fine, but the test can be optimized a bit.

samples/unit-tests/gantt/grid-axis/demo.js
TorsteinHonsi: PS: For a simplified, simplified demo, I think it boils down to something like this: https://jsfiddle.net/highcharts/df4x5308/

samples/unit-tests/time/dateformat/demo.js
```diff
@@ -146,3 +146,16 @@ QUnit.test(
         );
     }
 );
+
+QUnit[TestUtilities.isCET ? 'test' : 'skip'](
+    'Week dataTime in local timezone, #16550.',
```
TorsteinHonsi, line 151: Typo: Should be dateTime

samples/unit-tests/time/dateformat/demo.js
```diff
@@ -146,3 +146,16 @@ QUnit.test(
...
+    function (assert) {
+        const t = new Highcharts.Time({ useUTC: false });
+
+        assert.strictEqual(
+            t.dateFormat('%W', new Date("2015-04-12T01:00:00").getTime()),
+            '14',
+            'Week dataFormat for local timezone should return 14.'
```
TorsteinHonsi, line 158: Typo: should be dateFormat

ts/Core/Axis/GridAxis.ts
```diff
@@ -1444,7 +1444,14 @@ dateFormats.E = function (this: Time, timestamp: number): string {
 
 // Adds week date format
 dateFormats.W = function (this: Time, timestamp: number): string {
-    const d = new this.Date(timestamp);
+    const time = this,
+        d = new this.Date(timestamp),
+        unitsToOmmit = (['Hours', 'Milliseconds', 'Minutes', 'Seconds'] as Array<Time.TimeUnitValue>);
```
TorsteinHonsi, line 1449: Typo: should be "omit" with one m.

TorsteinHonsi (changes requested):
Some typos

TorsteinHonsi:
Did you forget to push the fix for the typos?

TorsteinHonsi (changes requested):
Pending typos

karolkolodziej:
Yep, I forget to push the changes. I apologise for that

TorsteinHonsi (approved):

PR #17179, opened 2022-04-12
Fixed #17048, added option to remember kbd focus for points.

### LINKED ISSUE #17048: Incorrect focus order in bar charts keyboard navigation
#### Expected behaviour
If we have the following actions:
1. Navigate with the keyboard to the series in a bar chart (with `tab` key).
2. Change the focus of the bars in the series (with `arrow` keys) to any bar that is not the first one.
3. Navigate out of the series element to the next element (`tab` key) or to the previous element (`shift + tab`).
4. Navigate back to the series element (`tab` or `shift + tab` depending on if you navigated to the previous or to the next element).

I expect that when I navigate back to the series element, the focus will fall on the last bar that it was on before I navigated out of the series element. Or at least there should be an option to allow such behavior. 

#### Actual behaviour
When I navigate back to the series element, the focus always falls on the first bar instead of on the last bar that it was on before I navigated out of the series element.

https://jsfiddle.net/mxw4ukez/1
Follow the steps I listed in **Expected behaviour**.

#### Product version
v9.3.3

#### Affected browser(s)
I tested on the following browser versions:
- Chrome Version 98.0.4758.102 (Official Build) (64-bit)
- Edge Version 98.0.1108.43 (Official Build) (64-bit)

oysteinmoseng:
Added option to remember keyboard focus for points: `accessibility.keyboardNavigation.seriesNavigation.rememberPointFocus`.
___
Closes #17048.

### DIFF
--- a/ts/Accessibility/Components/SeriesComponent/SeriesKeyboardNavigation.ts
+++ b/ts/Accessibility/Components/SeriesComponent/SeriesKeyboardNavigation.ts
@@ -444,11 +444,8 @@ class SeriesKeyboardNavigation {
                     }]
             ],
 
-            init: function (
-                this: KeyboardNavigationHandler
-            ): number {
-                highlightFirstValidPointInChart(chart);
-                return this.response.success;
+            init: function (this: KeyboardNavigationHandler): number {
+                return keyboardNavigation.onHandlerInit(this);
             },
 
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #17183, opened 2022-04-14
enhancement/replace-visible-inherit

pawellysy:
Changed the default value of [SVGElement.show](https://api.highcharts.com/class-reference/Highcharts.SVGElement#show)'s `inherit` argument to `true`.

~~Set default argument for SVGElement.show function to true.~~


pawelfus (approved):
Thanks!

TorsteinHonsi (approved):
Thanks, looking good! 

So I observe that we now have 0 cases of `.show(false)`. But it is probably a good idea to keep the argument anyway, and it doesn't cost much in terms of bytes.

PR #17195, opened 2022-04-19
bugfix/17186-csv-empty-cells

### LINKED ISSUE #17186: chart.getCSV() doesn't add empty cells at the end of the row
`chart.getCSV()`  doesn't add empty cells at the end of the row. Although according to the CSV specification `"Within the header and each record, there may be one or more fields, separated by commas.  Each line should contain the same number of fields throughout the file."` Taken from [rfc4180](https://datatracker.ietf.org/doc/html/rfc4180#section-2). So there should be the same number of delimiters in every row. 

Most CSV parsers do still open the CSV like this. But this thing causes some trouble for the simple solutions, e.g. when you do not want to explicitly ask the user about the delimiter used, but to detect it automatically. If you cannot expect that a CSV has the same number of columns within every row, this adds more errors.

Imagine we have 2 series of points:

    A = [ [1, 0.1], [2, 0.2], [3,0.3] ]
    B = [ [2, 0.5], [4, 1.0] ]  

#### Expected behavior (expected CSV output)
- all rows have equal number of cells  
    
      x,A,B
      1,0.1,
      2,0.2,0.5
      3,0.3,
      4,,1.0

#### Actual behavior

- rows with `x=1` and `x=3` have 2 cells, other rows have 3 cells  

      x,A,B
      1,0.1
      2,0.2,0.5
      3,0.3
      4,,1.0

- [https://jsfiddle.net/odjute63/1/](https://jsfiddle.net/odjute63/1/)

#### Product version
Highcharts 10.0.0 (latest)

#### Affected browser(s)
- all, but it is not a browser-dependent issue

mik-the-deutsch-dev:
Fixed #17186. Added appending delimiters to the row to markup empty cells when needed. 

The first row returned from `this.getDataRows()` is used to determine how many columns should the CSV have (Please, see [ExportData.ts :894](https://github.com/highcharts/highcharts/compare/master...mik-the-deutsch-dev:master#diff-cef6e25630962677be4a94d5ecb5a257df6355c123f2306b4389afd50c8c0456R894)). Is this a correct and reliable solution?

### DIFF
--- a/samples/unit-tests/export-data/basics/demo.js
+++ b/samples/unit-tests/export-data/basics/demo.js
@@ -379,9 +379,9 @@ QUnit.test('Pie chart, multiple', function (assert) {
 
     var csv = [
         '"Category","Categories","Subcategories"',
-        '"Animals",2',
+        '"Animals",2,',
         '"Cats",,1',
-        '"Plants",2',
+        '"Plants",2,',
         '"Dogs",,1',
         '"Potatoes",,1',
         '"Trees",,1'
@@ -1103,8 +1103,8 @@ QUnit.test('Point name (#13293)', function (assert) {
         }),
         csv =
             '"Category","Series 1 (x)","Series 1 (y)","Series 2"\n' +
-            '"Point2",1,9\n' +
-            '"Point1",2,6\n' +
+            '"Point2",1,9,\n' +
+            '"Point1",2,6,\n' +
             '20,,,9\n' +
             '30,,,6';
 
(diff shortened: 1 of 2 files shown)


raf18seb:
Hi @mik-the-deutsch-dev, sorry for the late reply and thanks for the PR!

I created a separate PR from our in-house branch to run CircleCI. Let's continue here: https://github.com/highcharts/highcharts/pull/17223

PR #17198, opened 2022-04-19
bugfix/15863-stacking-labels-translation

pawellysy:
Replaced hiding by translation of the stacking labels with `SVGElement.hide()`. Touches #15863


TorsteinHonsi (approved):
Thanks! The comment `// Move label away... ` is no longer precise, but I have dug through the blame history and as far as I see everything is working like it should.

PR #17199, opened 2022-04-20
Enhancement/a11y tweaks

oysteinmoseng:
Improved default behavior of accessibility module with axis descriptions as well as landmark roles.


TorsteinHonsi (approved):
Thanks!

PR #17206, opened 2022-04-21
Themes/high contrast

nasin:
Improved the theme colors for high contrast light and high contrast dark to show more hue variance.

~I updated the theme colors for high contrast light and high contrast dark to show more hue variance.~

### DIFF
--- a/ts/Extensions/Themes/HighContrastDark.ts
+++ b/ts/Extensions/Themes/HighContrastDark.ts
@@ -54,7 +54,9 @@ namespace HighContrastDarkTheme {
             '#6BA48F',
             '#EA3535',
             '#8D96B7',
-            '#ECCA15'
+            '#ECCA15',
+            '#20AA09',
+            '#E0C3E4'
         ],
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (changes requested):
Thanks! 

Please fix the lint error (it should be caught on pre-commit, have you done `npm install` correctly?

As for the colors themselves, it works much better now. More variance. Though I think it would be a good idea to keep the same total amount of colors as we do with the default palette. When we picked the current default colors, the idea was to maximize the variance between the first few colors at the expense of the ones at the end of the array. The assumption is that we are more likely to see a line or column chart with two series than three series, which is more likely than four series etcetera. So the first few colors are prioritized in terms of contrast. Following this, the last two colors could be for example variation in brightness of any of the preceding colors.

nasin:
Hi, I updated the colors to 10 and fixed the lint issue.

TorsteinHonsi (approved):

PR #17208, opened 2022-04-22
Added new option, accessibility.series.descriptionFormat.

oysteinmoseng:
Added new option, `accessibility.series.descriptionFormat`.
___
Avoids some unnecessary callback functions in the configurations.

### DIFF
--- a/ts/Accessibility/Components/SeriesComponent/SeriesDescriber.ts
+++ b/ts/Accessibility/Components/SeriesComponent/SeriesDescriber.ts
@@ -597,14 +597,17 @@ function defaultSeriesDescriptionFormatter(
         ): (boolean|Axis) {
             return chart[coll] && chart[coll].length > 1 && series[coll];
         },
+        index = series.index + 1,
+        numSeries = chart.series && chart.series.length,
+        numPoints = series.points && series.points.length,
         xAxisInfo = getSeriesAxisDescriptionText(series, 'xAxis'),
         yAxisInfo = getSeriesAxisDescriptionText(series, 'yAxis'),
         summaryContext = {
             name: series.name || '',
-            ix: (series.index as any) + 1,
-            numSeries: chart.series && chart.series.length,
-            numPoints: series.points && series.points.length,
-            series: series
+            ix: index,
+            numSeries,
+            numPoints,
+            series
         },
         combinationSuffix = chartTypes.length > 1 ? 'Combination' : '',
         summary = chart.langFormat(
@@ -613,13 +616,24 @@ function defaultSeriesDescriptionFormatter(
         ) || chart.langFormat(
             'accessibility.series.summary.default' + combinationSuffix,
             summaryContext
-        );
+        ),
+        axisDescription = (
+            shouldDescribeAxis('yAxis') ? ' ' + yAxisInfo + '.' : ''
+        ) + (
+            shouldDescribeAxis('xAxis') ? ' ' + xAxisInfo + '.' : ''
+        ),
+        formatStr = chart.options.accessibility.series.descriptionFormat || '';
 
-    return summary + (description ? ' ' + description : '') + (
-        shouldDescribeAxis('yAxis') ? ' ' + yAxisInfo : ''
-    ) + (
-        shouldDescribeAxis('xAxis') ? ' ' + xAxisInfo : ''
-    );
+    return format(formatStr, {
+        seriesDescription: summary,
+        authorDescription: (description ? ' ' + description : ''),
+        axisDescription,
+        series,
+        numSeries,
+        numPoints,
+        chart,
+        index
+    }, void 0);
 }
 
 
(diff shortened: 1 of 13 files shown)


ts/Accessibility/Options/Options.ts
```diff
@@ -404,13 +405,39 @@ const Options: DeepPartial<OptionsType> = {
...
+             * Format to use for describing the data series group to assistive
+             * technology - including screen readers. The series context is
+             * available as `{series}`.
+             *
+             * {seriesDescription} refers to the automatic description of the
+             * series type and number of points added by Highcharts by default.
+             * {authorDescription} refers to the description added in
+             * [series.description](#plotOptions.series.description) if one is
+             * present. {axisDescription} refers to the description added if
+             * the chart has multiple X or Y axes.
```
TorsteinHonsi, line 426: Put code in backticks

ts/Accessibility/Components/SeriesComponent/SeriesDescriber.ts
```diff
@@ -597,14 +597,17 @@ function defaultSeriesDescriptionFormatter(
...
+        numPoints = series.points && series.points.length,
         xAxisInfo = getSeriesAxisDescriptionText(series, 'xAxis'),
         yAxisInfo = getSeriesAxisDescriptionText(series, 'yAxis'),
         summaryContext = {
             name: series.name || '',
-            ix: (series.index as any) + 1,
-            numSeries: chart.series && chart.series.length,
-            numPoints: series.points && series.points.length,
-            series: series
+            ix: index,
```
TorsteinHonsi, line 607: In this `summaryContext` this property is called `ix`, while in the `descriptionFormat` context below the same property is called `index`. And in both cases it is `series.index + 1` which may itself be confusing. What about using [number](https://english.stackexchange.com/questions/366834/what-is-a-word-that-means-one-based-index) for the one-based index?

ts/Accessibility/Components/SeriesComponent/SeriesDescriber.ts
```diff
@@ -613,13 +616,24 @@ function defaultSeriesDescriptionFormatter(
...
+
+    return format(formatStr, {
+        seriesDescription: summary,
+        authorDescription: (description ? ' ' + description : ''),
+        axisDescription,
+        series,
+        numSeries,
+        numPoints,
+        chart,
+        index
```
TorsteinHonsi, line 635: Just thinking - do we need all of these shorthands? Or could for instance `{numSeries}` as easily be written `{chart.series.length}` if we add `chart` to the context and show this in docs and demos`. The demos previously, when using formatters, used more of those paths that correspond to the class reference.

TorsteinHonsi (changes requested):
Thanks! A couple of discussion points.

ts/Accessibility/Components/SeriesComponent/SeriesDescriber.ts
oysteinmoseng: Makes sense, will change!

ts/Accessibility/Components/SeriesComponent/SeriesDescriber.ts
oysteinmoseng: I see what you mean, although since number of series and points are very common to add to these format strings I figured it would help the users if we gave them shorthands. Happy to remove them if you think it's too bloaty though?

TorsteinHonsi: Yes I am leaning towards removing them and adding `chart` instead. Because `{chart.series.length}` is not more complicated to use than `{numSeries}`, provided that they are documented the same way. But `chart.series.length` additionally reflects the actual structure in Highcharts.

oysteinmoseng: Hm, looking at this further our fancy formatting functions like `#plural()` won't currently work with property access. I can try to add it, but could get a bit convoluted.

Context: `line: '{name}, line {seriesNumber} of {chart.series.length} with {series.points.length} data {#plural(series.points.length, points, point)}.'`

oysteinmoseng: Well, we already have `getNestedProperty` separated out in utils, so hopefully not too convoluted then - I'll give it a shot

TorsteinHonsi:
@oysteinmoseng Okay - I wasn't aware of that, I thought this was a new addition. We definately don't want to break anything existing. And the context should always be the same between a `formatter` and the accompanying `format`. Is that honored for the new `descriptionFormat`?

oysteinmoseng:
Sounds good! I would say the context is similar enough, since the `descriptionFormatter` receives the series, and from that can get the chart and everything else it needs?

TorsteinHonsi (approved):

PR #17214, opened 2022-04-26
Added option accessibility.point.describeNull.

oysteinmoseng:
Added option `accessibility.point.describeNull`.


TorsteinHonsi (approved):
Thanks!

PR #17215, opened 2022-04-26
Feature/14928-zoom-options-refactoring 

hihajdus:
Refactored zoom options, introduce the new object structure. Related to https://github.com/highcharts/highcharts/issues/14928.


ts/Core/Chart/ChartDefaults.ts
```diff
@@ -992,13 +994,61 @@ const ChartDefaults: ChartOptions = {
      * However, `zoomBySingleTouch` will interfere with touch-dragging the
      * chart to read the tooltip. And especially when vertical zooming is
      * enabled, it will make it hard to scroll vertically on the page.
-     * @since 9.0.0
+     * @since next
```
raf18seb, line 997: Should we change the version here? It was added in 9.0.0, and now it's deprecated- so the version when it was added should stay the same

ts/Core/Chart/ChartDefaults.ts
```diff
@@ -992,13 +994,61 @@ const ChartDefaults: ChartOptions = {
...
+        /**
+         * Equivalent to [type](#chart.zooming.type), but for multitouch
+         * gestures only. By default, the `pinchType` is the same as the
+         * `type` setting. However, pinching can be enabled separately in
+         * some cases, for example in stock charts where a mouse drag pans the
+         * chart, while pinching is enabled. When [tooltip.followTouchMove](
+         * #tooltip.followTouchMove) is true, pinchType only applies to
+         * two-finger touches.
+         *
+         * @type       {string}
```
raf18seb, line 1019: You should add version info to the new options instead of changing old ;)

ts/Core/Chart/ChartDefaults.ts
```diff
@@ -992,13 +994,61 @@ const ChartDefaults: ChartOptions = {
...
+         * @sample     highcharts/chart/zoombysingletouch
+         *             Zoom by single touch enabled, with buttons to toggle
+         * @product    highcharts highstock gantt
+         */
+        singleTouch: false,
+
+        /**
+         * Decides in what dimensions the user can zoom by dragging the mouse.
+         * Can be one of `x`, `y` or `xy`.
+         *
```
raf18seb, line 1043: Same here, missing `@since`

ts/Core/Chart/Chart.ts
```diff
@@ -477,6 +477,21 @@ class Chart {
...
+                optionsChart.zoomBySingleTouch = pick(
+                    optionsChart.zooming.singleTouch,
+                    optionsChart.zoomBySingleTouch
+                );
+                optionsChart.pinchType = pick(
+                    optionsChart.zooming.pinchType,
+                    optionsChart.pinchType
+                );
+            }
+
```
TorsteinHonsi, line 494: See if the filesize can be optimized by storing `optionsChart.zooming` in a local variable `zooming`. Another possible optimization to check would be `extend(optionsChart, { zoomType: pick(...`

ts/Core/Chart/ChartDefaults.ts
```diff
@@ -992,13 +994,65 @@ const ChartDefaults: ChartOptions = {
...
      * @product    highcharts highstock gantt
+     * @deprecated
      */
     zoomBySingleTouch: false,
 
+    /**
+     * Chart zooming options.
+     * @since       next
+     */
+    zooming: {
```
TorsteinHonsi, line 1009: In addition to these three zoom properties, we also have `zoomKey` which I believe belongs here.

TorsteinHonsi (changes requested):
Thanks!

raf18seb (approved):
Nothing else from me, thanks!

ts/Core/Chart/ChartDefaults.ts
```diff
@@ -992,13 +994,65 @@ const ChartDefaults: ChartOptions = {
...
+         * Decides in what dimensions the user can zoom by dragging the mouse.
+         * Can be one of `x`, `y` or `xy`.
+         *
+         * @type       {string}
+         * @product    highcharts highstock gantt
+         * @validvalue ["x", "y", "xy"]
+         * @since      next
+         * @apioption  chart.zooming.type
+         */
+        type: 'x'
```
KacperMadej, line 1053: Is this new default value correct? [The current default is undefined](https://api.highcharts.com/highcharts/chart.zoomKey).

hihajdus: Mistake, type should default to undefined.

ts/Core/Chart/Chart.ts
hihajdus: Thanks for the suggestions, adding a local variable reduced the file weight, extend increased.

ts/Core/Chart/ChartDefaults.ts
hihajdus: Added zoomKey, what do you think about adding `chart.resetZoomButton` to zooming options, in the future?

KacperMadej (approved):

ts/Core/Chart/Chart.ts
```diff
@@ -477,6 +477,27 @@ class Chart {
...
+                optionsChart.zoomBySingleTouch = pick(
+                    zooming.singleTouch,
+                    optionsChart.zoomBySingleTouch
+                );
+                optionsChart.zoomType = pick(
+                    zooming.type,
+                    optionsChart.zoomType
+                );
+                optionsChart.zoomKey = pick(
+                    zooming.zoomKey,
```
TorsteinHonsi, line 496: `zooming.zoomKey` is a bit repetetive isn't it? I think it should be `zooming.key`.

ts/Core/Chart/Chart.ts
```diff
@@ -477,6 +477,27 @@ class Chart {
             this.callback = callback;
             this.isResizing = 0;
 
+            const zooming = optionsChart.zooming;
+
+            if (zooming) {
+                optionsChart.pinchType = pick(
+                    zooming.pinchType,
+                    optionsChart.pinchType
+                );
```
TorsteinHonsi, line 486: Now that we have marked the old options deprecated, I think it would be better if we turn this expression upside down: 
```ts
                zooming.pinchType = pick(
                    zooming.pinchType,
                    optionsChart.pinchType
                );
```
Then we would use the `zooming` option in the implementations, so that the actual implementation is not referring to the deprecated names. Unless that results in something significantly more complex.

Subsequently, we should probably also make the `zooming` property mandatory on the `ChartOptions` interface.

`zooming: Partial<ChartZoomingOptions>;`

ts/Core/Chart/ChartDefaults.ts
TorsteinHonsi: Yes, I think `resetZoomButton` also belongs here. Then as `zooming.resetButton`.

hihajdus:
To do:

- [ ] Check sample demos

ts/Core/Chart/Chart.ts
```diff
@@ -477,6 +477,31 @@ class Chart {
             this.callback = callback;
             this.isResizing = 0;
 
+            const zooming = optionsChart.zooming = optionsChart.zooming || {};
+
+            if (zooming) {
```
TorsteinHonsi, line 482: Unnecessary condition? As far as I can see, `zooming` should now always be truthy.

TorsteinHonsi (changes requested):

ts/Core/Chart/Chart.ts
hihajdus: True, this condition it's no longer needed.

TorsteinHonsi (approved):
Thanks, looking good now!

PR #17216, opened 2022-04-26
Fixed missing map title a11y description.

oysteinmoseng:
Fixed missing accessible description of map titles with TopoJSON maps.


TorsteinHonsi (approved):
Thanks!

PR #17218, opened 2022-04-26
Enhancement/a11y warning

oysteinmoseng:
Added warning by default when the `accessibility.js` module is not loaded.
___
Also added the module to all remaining demos, with basic configuration.

The warning can be overridden either through options or through overriding the `Chart.warnIfA11yModuleNotLoaded` function.


oysteinmoseng:
@TorsteinHonsi Just occurred to me that we do have `H.error` that could be used with the `stop=false` parameter and a direct string input. Do you think there is a point in adding that overhead for consistency, or should keep as is with a simple `console.warn`?

TorsteinHonsi (changes requested):
Thanks! As you mentioned, I agree we should use the Highcharts `error` feature instead of `console.warn`, in order to get the warning into our usual error flow (with its hooks for error handling).

TorsteinHonsi (approved):
Thanks!

PR #17223, opened 2022-04-27
bugfix/17186-csv-empty-cells

### LINKED ISSUE #17186: chart.getCSV() doesn't add empty cells at the end of the row

raf18seb:
Fixed #17186, CSV rows were missing delimiters when data was missing at the end of a row.

This is a continuation of: https://github.com/highcharts/highcharts/pull/17195

I checked the above solution and I think it's the best when it comes to performance.
If we'd like to add this logic to the `getDataRows` method, we would have to loop rows one more time. In this case, we could add this code here: https://github.com/highcharts/highcharts/blob/96fb71da571456bd45724df75767e92bfdb014e3/ts/Extensions/ExportData.ts#L793

What do you think @TorsteinHonsi? Should it be even fixed in the first place? As a workaround, the `null` points could be added to data. But, the CSV should have the same number of cells in all rows by definition.

Demo: http://jsfiddle.net/BlackLabel/pxLgb2d6/

### DIFF
--- a/samples/unit-tests/export-data/basics/demo.js
+++ b/samples/unit-tests/export-data/basics/demo.js
@@ -379,9 +379,9 @@ QUnit.test('Pie chart, multiple', function (assert) {
 
     var csv = [
         '"Category","Categories","Subcategories"',
-        '"Animals",2',
+        '"Animals",2,',
         '"Cats",,1',
-        '"Plants",2',
+        '"Plants",2,',
         '"Dogs",,1',
         '"Potatoes",,1',
         '"Trees",,1'
@@ -1103,8 +1103,8 @@ QUnit.test('Point name (#13293)', function (assert) {
         }),
         csv =
             '"Category","Series 1 (x)","Series 1 (y)","Series 2"\n' +
-            '"Point2",1,9\n' +
-            '"Point1",2,6\n' +
+            '"Point2",1,9,\n' +
+            '"Point1",2,6,\n' +
             '20,,,9\n' +
             '30,,,6';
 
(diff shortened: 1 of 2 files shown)


TorsteinHonsi:
@raf18seb 

I think this fix makes sense. And it seems like HTML table is already handling this correctly: https://jsfiddle.net/highcharts/yg9em5dv/1/ .

TorsteinHonsi:
@raf18seb In the latest commit I tried just setting the length of the row, which seems to have the same effect as pushing undefined. It fixes the problem with the missing delimiters. Please have a look if you can see any problems with that.

raf18seb:
Looks good! I only improved a comment because it didn't fit after a variable was removed.

raf18seb:
@TorsteinHonsi

TorsteinHonsi (approved):

Issue #17229, opened 2022-04-28
Wrong API for borderColor and borderWidth in mapPoint


madepiet:
#### Expected behaviour
It should work like this:
```
 marker: {
    lineColor: 'pink'
    lineWidth: 5
}
```
Demo: https://jsfiddle.net/BlackLabel/pf65Lotk/

#### Actual behaviour
borderColor and borderWidth are buggy inherited from type: 'map' and shouldn't be in the API:
https://api.highcharts.com/highmaps/plotOptions.mappoint.borderColor
https://api.highcharts.com/highmaps/plotOptions.mappoint.borderWidth

#### Live demo with steps to reproduce
https://jsfiddle.net/BlackLabel/79hf4eya/

#### Product version
v10.0.0

PR #17230, opened 2022-04-28
docs/17229-borderColor-borderWidth-in-mapPoint

madepiet:
Fixed: #17229, API for borderColor and borderWidth in mapPoint


TorsteinHonsi (approved):
Thanks!

PR #17231, opened 2022-04-28
Docs: Made Point.shapeArgs property public.

### LINKED ISSUE #16824: Typescript: HighCharts.Point doesn't expose shapeArgs
#### Expected behaviour
When using Typescript, I should be able to access shapeArgs on a HighCharts.Point object, as follows:
```
const foo = (pointLabel:Highcharts.PointLabelObject): void => {
     const shape = pointLabel.point.shapeArgs;
     //do something cool with shape
}
```

#### Actual behaviour
In order to access shapeArgs on a HighCharts.Point object, I have to cast it as any:
```
const foo = (pointLabel:Highcharts.PointLabelObject): void => {
     const shape = (pointLabel.point as any).shapeArgs;
     //do something cool with shape
}
```

https://jsfiddle.net/longmatthewh/pxstdvz1/7/

#### Product version
Highcharts JS v9.3.2

#### Affected browser(s)
N/A

bre1470:
Made Point.shapeArgs property public. Closed #16824

### DIFF
--- a/ts/Core/Series/Point.ts
+++ b/ts/Core/Series/Point.ts
@@ -210,8 +210,13 @@ class Point {
     public series: Series = void 0 as any;
 
     /**
-     * The attributes to control the rendered SVG shape of the point.
+     * The attributes for the rendered SVG shape of the point.
      *
+     * Changes should only be made during translate events, including
+     * `afterTranslate`, otherwise the rendered SVG shape might not reflect the
+     * latest changes.
+     *
+     * @readonly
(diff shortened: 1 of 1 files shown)


raf18seb (approved):

bre1470:
I struggle to create a working demo for shapeArgs changes. `series.events.afterTranslate` is too early/or gets overwritten. I suggest we remove any info about changing it.

pawelfus (approved):

TorsteinHonsi (approved):
Thanks!

PR #17248, opened 2022-05-05
Fixed #17236, a11y timeout caused test issues.

### LINKED ISSUE #17236: accessibility warning in setTimeout breaks tests
Version 10.1.0 introduced an error message in case the accessibility module is not loaded. See the change here: https://github.com/highcharts/highcharts-dist/blob/master/es-modules/Core/Chart/Chart.js#L1953-L1966

Unfortunately, it does that in a setTimeout, and there is no way to disable this setTimeout, even by disabling accessibility or adding the accessibility module, because the checks are done in the setTimeout callback itself. So, this setTimeout is always executed, no matter what.
It might not seem like a big deal, but we have tests (in an Angular application), using a "fakeAsync" zone where tests fail shen a  timer is still in the queue. And this timeout does that: it introduces a new, unexpected timer, in the queue. 

#### Product version
Highcharts 10.1.0

#### Affected browser(s)
All

oysteinmoseng:
Fixed #17236, timeout for accessibility module warning caused test issues.
___
The `setTimeout` should be unnecessary, technically, so suggesting its removal since it caused issues for users.

Also added some minimal code to make the chart behave as an image with alt text = chart title if the module is not loaded, as opposed to the gibberish one would otherwise get with screen readers et al.

### DIFF
--- a/samples/unit-tests/accessibility/accessibility-options/demo.js
+++ b/samples/unit-tests/accessibility/accessibility-options/demo.js
@@ -26,6 +26,9 @@ QUnit.test('Accessibility disabled', function (assert) {
             accessibility: {
                 enabled: false
             },
+            title: {
+                text: 'No a11y'
+            },
             series: [
                 {
                     data: [1, 2, 3, 4, 5, 6]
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (approved):
Thanks!

PR #17251, opened 2022-05-05
Tools: Upgraded karma

bre1470:
- Karma upgraded
- Added karma downgrade for CI workflow of IE8
- Added note that Google Closure Compiler should never get upgraded because of Java-dependency in later versions
- Added fixed timezone for CI workflow of Cypress


.circleci/config.yml
```diff
@@ -152,6 +152,36 @@ jobs:
...
+          name: Downgrade karma
+          command: |
+            npm i --no-save karma@4.3.0
+      - run:
+          command: "npx cross-env NODE_OPTIONS=--max_old_space_size=3072 karma start test/karma-conf.js --single-run --splitbrowsers << parameters.browsers >> --oldie --browsercount 1"
+          no_output_timeout: 20m
+      - run:
+          name: Upgrade karma
+          command: |
+            npm i karma
```
bre1470, line 178: Unsure if this is necessary.

TorsteinHonsi (approved):
Thanks!

.circleci/config.yml
bre1470: It definitely does not hurt, but could be a step we could skip. 🤷‍♀️

PR #17259, opened 2022-05-06
Tools: Upgraded to TypeScript 4

bre1470:
- Upgraded to TypeScript 4.6
- Improved code according to TypeScript 4 validations
- Fixed CircleCI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -358,7 +358,7 @@ jobs:
           command: npx gulp write-file-sizes --filename master.json
       - generate_references_command:
           browsercount: 2
-      - run: git checkout ${CIRCLE_BRANCH}  && git log --oneline -5 && npm ci --quiet --cache .npm --prefer-offline && npx gulp scripts
+      - run: git checkout ${CIRCLE_BRANCH}  && git log --oneline -5 && npm ci --quiet --cache .npm --prefer-offline && npx gulp scripts --force
       - run:
           name: File size comparison
           command: |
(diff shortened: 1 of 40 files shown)


TorsteinHonsi:
Preliminary note: Checking where the extra 2k bytes come from...

bre1470:
They come from transpiling all `"" + ""` to `"".concat("")`.

The reason according to the Internet is that `object + object` can trigger `.toValue().toString()` if it exists, while concat will always trigger `.toString()`. This is done to be compliant with the template literal \`\` which will always trigger `.toString()`.

The transpilation seems to happen only:
a. When you do not use template literal
b. When you transpile down to ES3 and ES5

Some Reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#browser_compatibility

I wonder if you could do `["", ""].join("")` as a workaround. I will check that.

bre1470:
So we have some options:
- For now I propose to stick to TypeScript 4.3 and maybe have ESLint warnings agains template literals
- For the future we should migrate template literals to `"" + ""` or `["", ""].join("")` pattern.

bre1470:
In addition the internal `__extends` function has expanded. In our case I think it is safe to put something in the code processor.
Screenshot 2022-05-07 at 07 08 35

bre1470:
The file size comparison does not work for tools changes. I do not find the issue, but the logical conclusion is that it is not using the gulp tasks to create the files.

Real world comparison:
| | master | candidate | difference |
|-------------|-------------:|-------------:|-------------:|
| highcharts.src.js | 2,146,469 B | 2,146,631 B | 162 B |

bre1470:
Finally the file size table shows real world numbers. Please take another look, @TorsteinHonsi.

TorsteinHonsi (approved):
Awesome, thanks!

PR #17268, opened 2022-05-09
bugfix/15863-fadeOut-hide-by-translation

pawellysy:
Part of #15863, removed `SVGElement.fadeOut()` hiding by translation.


TorsteinHonsi (approved):
Thanks!

PR #17271, opened 2022-05-09
bugfix/15863-TreeGridTick-hide-by-translation

pawellysy:
Part of #15863, Removed hide by translation from `TreeGridTick`.


TorsteinHonsi (approved):
Thanks!

PR #17272, opened 2022-05-09
New blog demos and data

mekhatria:
Feel free to check these new demos for the upcoming blogs


samples/highcharts/blog/cognitive-impairment/demo.html
```diff
@@ -0,0 +1,6 @@
+<script src="https://code.highcharts.com/highcharts.js"></script>
+<script src="https://code.highcharts.com/modules/exporting.js"></script>
+<script src="https://code.highcharts.com/modules/accessibility.js"></script>
+<script src="https://code.highcharts.com/modules/series-label.js"></script>
+<script src="https://code.highcharts.com/themes/high-contrast-light.js"></script>
+ <div id="container"></div>
```
raf18seb, line 6: Unnecessary space before tag

samples/highcharts/blog/cognitive-impairment-no-labels/demo.js
```diff
@@ -0,0 +1,161 @@
...
+    series: [
+        {
+            name: "Food",
+            data: [
+                { x: 828000, y: 3.19 },
+                { x: 1008000, y: 38.09 },
+                { x: 1440000, y: 69.63 },
+                { x: 3456000, y: 73.93 },
+                { x: 13680000, y: 55.13 },
+                { x: 16956000, y: 40.64 },
```
raf18seb, line 69: You can use arrays here (they're faster and the code is cleaner)
```js
data: [
    [828000, 3.19],
    [..., ...],
   etc.
]
```

samples/highcharts/blog/read-csv-data-using-beforeparse/demo.js
```diff
@@ -0,0 +1,42 @@
+Highcharts.setOptions({
+    colors: ["#058DC7", "#c7058d"]
+});
```
raf18seb, line 3: You can move the `colors` array directly into the chart's options (you don't need to use Highcharts.setOptions())

samples/highcharts/blog/cognitive-impairment-no-labels/demo.js
```diff
@@ -0,0 +1,161 @@
+function padTo2Digits(num) {
+    return num.toString().padStart(2, "0");
```
raf18seb, line 2: We prefer to use single quotes when writing JavaScript code. In some demos you use double quotes, in others you use single - it's good to be consistence. Not a big issue tho

raf18seb (changes requested):
Nice, I have only a few small comments, see them below

samples/highcharts/blog/cognitive-impairment-no-labels/demo.js
```diff
@@ -0,0 +1,161 @@
+function padTo2Digits(num) {
+    return num.toString().padStart(2, '0');
+}
+
+function convertMsToHM(milliseconds) {
```
TorsteinHonsi, line 5: Add a comment explaining why we need this function, so that implementers are not left with the false impression that they need to do all this just to print time correctly in Highcharts. In this case the X axis is a special case of time axis with a running number of hours.

samples/highcharts/blog/cognitive-impairment/demo.js
```diff
@@ -0,0 +1,161 @@
+function padTo2Digits(num) {
+    return num.toString().padStart(2, '0');
+}
+
+function convertMsToHM(milliseconds) {
```
TorsteinHonsi, line 5: As above

TorsteinHonsi (changes requested):
Thanks, just add some comments for the time formatting code then it's good to go.

samples/highcharts/blog/cognitive-impairment-no-labels/demo.js
mekhatria: Good point

samples/highcharts/blog/cognitive-impairment/demo.js
mekhatria: Done

TorsteinHonsi (approved):

PR #17277, opened 2022-05-10
bugfix/17205-clusters-maps

### LINKED ISSUE #17205: Marker clusters, map zoom: error when all points in cluster have the same coordinates
#### Expected behaviour
No errors in the console. Either zooming to a precalculated `maxZoom` value, or not zooming at all.

#### Actual behaviour
Error in the console that also hangs a browser
The error occurs because points have the same coordinates and the library doesn't know how to calculate the `zoom`.

```
Error: <g> attribute transform: Expected number, "…late(0,0) scale(NaN NaN)".
```

Click on the marker cluster
https://jsfiddle.net/BlackLabel/r1whj48k/

#### Product version
Highcharts Maps 10.0.0

The error occurs because the `getScale` method returns invalid value in the `getSVGTransform` method (`mapView.zoom` is `Infinity`).

Workaround: define `mapView.maxZoom`: https://jsfiddle.net/BlackLabel/Lf4cxg8k/

KacperMadej:
Fixed #17205, the marker cluster of coinciding points was causing an infinite loop on click.


samples/unit-tests/series/marker-clusters/demo.js
```diff
@@ -824,3 +824,60 @@ QUnit.test('OptimizedKmeans algorithm tests.', function (assert) {
         'optimizedKmeans should use kmeans again when extremes range is greater than on chart init.'
     );
 });
+
+QUnit.test('Maps tests.', function (assert) {
```
TorsteinHonsi, line 828: Can we combine these assertions with an existing map chart to avoid increasing the processing time of the test suite?

And if not, can we load a lighter map, without proj4 and with simpler and fewer shapes?

TorsteinHonsi (changes requested):
Some test performance notes.

DJTechnoo (approved):
This looks good, nothing else to add 😄

TorsteinHonsi (approved):

PR #17287, opened 2022-05-12
Tools: Fixed conflict between CJS and ESM.

bre1470:
Solved compatibility issues occurring with the distribution workflow.


TorsteinHonsi (approved):
Thanks!

PR #17289, opened 2022-05-12
bugfix/17116-stack-labels-negative

### LINKED ISSUE #17116: Stack labels on graphs - when legend is on the right & there are negative values, stack labels include the legend width when trying to align themselves
#### Expected behaviour
I should see my labels aligned at the end of each bar

#### Actual behaviour
When the legend is on the right and there are values that are negative, the stack labels seem to include the legend width when trying to align themselves, which is causing so much space on the right of the smaller bars and the labels don't show for the largest one.

http://jsfiddle.net/guwefdac/2/

#### Product version
Highcharts v8.2 & v10

#### Affected browser(s)
All

mfilipiec:
Fixed #17116, wrong stack labels calculations when `inverted` is set to `true`

### DIFF
--- a/samples/unit-tests/series-column/stacking/demo.js
+++ b/samples/unit-tests/series-column/stacking/demo.js
@@ -244,3 +244,35 @@ QUnit.test('Null threshold (#7420)', function (assert) {
         'The points should have an extent'
     );
 });
+
+QUnit.test('Negative stack labels in the inverted chart should be properly calculated. (#17116)', function (assert) {
+    const chart = Highcharts.chart('container', {
+        chart: {
+            inverted: true,
+            spacingRight: 200
+        },
+        yAxis: {
+            min: -500,
+            stackLabels: {
+                enabled: true
+            }
+        },
+        series: [{
+            type: 'column',
+            stacking: 'normal',
+            data: [-50]
+        }]
+    });
+
+    const point = chart.series[0].points[0],
+        pointPos = chart.plotWidth + chart.plotLeft - point.plotY,
+        label = chart.yAxis[0].stacking.stacks['-column,,,'][0].label,
+        labelPos = chart.plotLeft + label.x,
+        labelWidth = label.width;
+        
+    assert.strictEqual(
+        pointPos - (labelPos + labelWidth) < 1,
+        true,
+        'The stack label should be in close range'
+    );
+});
--- a/ts/Extensions/Stacking.ts
+++ b/ts/Extensions/Stacking.ts
@@ -480,6 +480,12 @@ class StackItem {
                     stackBox
                 );
             }
+
+            // Check if chart is inverted with negative values #17116
+            if (chart.inverted) {
+                label.alignAttr.x += chart.spacing[1] * (isNegative ? 1 : 0);
+            }
+
             label.attr({
                 x: label.alignAttr.x,
                 y: label.alignAttr.y


TorsteinHonsi:
Thanks! I see we already have some code apparently dealing with related problems here: 

https://github.com/highcharts/highcharts/blob/8b624efdc43b9ef0d1324fd276ec6fc87b176bb9/ts/Extensions/Stacking.ts#L432-L444

Furthermore, it seems to be quite easy to break it, for example with a reversed Y axis: https://jsfiddle.net/highcharts/9jf356op/ 

I think instead of fixing just this case we should do a wider investigation of label placement with all combinations of `chart.inverted`, `xAxis.reversed`, `yAxis.reversed` and the `textAlign` property. Probably there are patterns that we have not considered. For example, the position of a label would be the same for a negative stack on a reversed axis, as for a positive stack on a non-reversed axis.

TorsteinHonsi (changes requested):
Thanks! I see we already have some code apparently dealing with related problems here: 

https://github.com/highcharts/highcharts/blob/8b624efdc43b9ef0d1324fd276ec6fc87b176bb9/ts/Extensions/Stacking.ts#L432-L444

Furthermore, it seems to be quite easy to break it, for example with a reversed Y axis: https://jsfiddle.net/highcharts/9jf356op/ 

I think instead of fixing just this case we should do a wider investigation of label placement with all combinations of `chart.inverted`, `xAxis.reversed`, `yAxis.reversed` and the `textAlign` property. Probably there are patterns that we have not considered. For example, the position of a label would be the same for a negative stack on a reversed axis, as for a positive stack on a non-reversed axis.

raf18seb (changes requested):
Agreed with Torstein, it could be fixed in a better place.

Additionally, is it possible to use an already existing test instead of creating a new one? We should think about a performance because we already have thousands of tests.

TorsteinHonsi:
Thanks so far @karolkolodziej . Regarding the visual diffs, as far as I can see they are all caused by text shifting 1px in one direction. I wouldn't consider that a problem if it is a consequence of an improved placement logic.

karolkolodziej:
Indeed, those changes are caused by the modified placement logic.
But at that point, I'm wondering which scenario this code should cover?

From what I can see there are two scenarios to discuss: 
1. How the labels should be positioned when `chart.inverted` and `axis.reversed`- [demo](https://jsfiddle.net/BlackLabel/4yjg31ar/)
Should it be inside or outside of the bar?
2. Should we also check a similar case but with `!isNegative`- [demo](https://jsfiddle.net/BlackLabel/ba56vgzn/) 
Right now the label is missing but I'm not sure if this output does make sense?

TorsteinHonsi:
@karolkolodziej Sorry for the late reply!

1. It should be outside. In terms of positioning, it is the exact equivalent of https://jsfiddle.net/highcharts/4m8caj92/
2. There is room for the label so I think it should be rendered. That also applies to the negated version of the same setup: https://jsfiddle.net/highcharts/b4egoy5k/

TorsteinHonsi (changes requested):
I set up a fiddle to try to get my brain around the different dimensions: https://jsfiddle.net/highcharts/9jf356op/

This is what it currently renders:
Skjermbilde 2022-08-18 kl  10 38 16

Some placement issues:
* In Chart 1, the stack labels are correctly positioned for positive stacks. For negative stacks, the -6 label is not moved inside. This is a regression since the current production release.
* The same applies to Chart 2.
* In Chart 3, the stack labels for positive stacks should be to the left of the bars. The -6 label is missing here too, should be moved inside the box.
* In Chart 4, all stack labels should be at the other end of the stack. 

In general, a positive stack when `yAxis.reversed: false` has the same layout as a negative stack when `yAxis.reversed: true`. That means we can do some general sign flipping like
```
let side = reversed ? -1 : 1;
if (isNegative) {
    side *= -1;
}
```

TorsteinHonsi:
Closing this, will be refactored.

PR #17291, opened 2022-05-13
bugfix/17253-maps-docs-brokenjsfiddle

### LINKED ISSUE #17253: /docs/maps/create-custom-maps - jsfiddle broken example
#### Expected behaviour
https://github.com/highcharts/highcharts/blob/master/docs/maps/create-custom-maps.md
at step 15 an example jsfiddle is presented: [jsfiddle.net/highcharts/TUy7x/1/](https://jsfiddle.net/highcharts/TUy7x/1/) running the fiddle would render the map

#### Actual behaviour
error: "<a class='gotoLine' href='#56:21'>56:21</a> Uncaught TypeError: $(...).highcharts is not a function"

#### Affected browser(s)
Chrome Version 100.0.4896.127 (Official Build) (64-bit)

mfilipiec:
Fixed #17253, broken demos in map docs

### DIFF
--- a/docs/maps/create-custom-maps.md
+++ b/docs/maps/create-custom-maps.md
@@ -47,7 +47,7 @@ Highcharts Maps can be used for more than geographic data, and more than the fea
 
 ![map-from-svg-10.png](map-from-svg-10.png)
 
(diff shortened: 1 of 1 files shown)


raf18seb (changes requested):
I don't think we should add any official demos from the Black Label account.
Either @TorsteinHonsi could fork/update the demos, or maybe we should add them to the samples (better solution for a longer period of time)

TorsteinHonsi:
I have updated the demos. Generally we prefer to host the demos in the `/samples` of the highcharts repo, but this one is so slim that I just updated it.

PR #17293, opened 2022-05-13
Linted/Migrated AreaSeries

bre1470:


ts/Series/Area/AreaSeries.ts
```diff
@@ -27,11 +27,7 @@ import Color from '../../Core/Color/Color.js';
 const { parse: color } = Color;
 import LegendSymbol from '../../Core/Legend/LegendSymbol.js';
 import SeriesRegistry from '../../Core/Series/SeriesRegistry.js';
-const {
-    seriesTypes: {
-        line: LineSeries
-    }
-} = SeriesRegistry;
+const { seriesTypes: { line: LineSeries } } = SeriesRegistry;
```
TorsteinHonsi, line 30: What is the rationale behind this change? This is a pattern we have never used before. In fact, we have consistently broken up deep objects into several lines... Should deconstructors be handled differently from object literals?

bre1470: Looking at this, it shouldn't be a deconstruct at all.

bre1470: Sorry, the change I had in mind is not possible at the moment. It has to wait until after assembling changes.

TorsteinHonsi (approved):
Thanks!

PR #17306, opened 2022-05-18
Fixed #17013, map view fit to bounds after setData

### LINKED ISSUE #17013: Highmaps - Updating the series in allAreas: false results in the wrong area being rendered
#### Expected behaviour
Setting a new series for HighCharts should be reflected correctly in the browser.

#### Actual behaviour
The dataLabels are updated, but the actual shape of the area remains the same.
Note: this works, if the count in the series changes (hence, in the linked CodeSandbox, switching to/from "USA + Mexico" works, but any switch that doesn't involve that, will fail)

https://codesandbox.io/s/hc-map-switching-countries-16lqxl?file=/src/Demo.jsx

#### Product version
* highcharts@9.3.2 & highcharts-react-official@3.1.0 (for example)
Note: this used to work in highcharts@9.0.1

#### Affected browser(s)
All

TorsteinHonsi:
Fixed #17013, map view did not fit correctly to bounds after running `Series.setData`.

### DIFF
--- a/samples/unit-tests/maps/setdata/demo.js
+++ b/samples/unit-tests/maps/setdata/demo.js
@@ -1013,4 +1013,20 @@ QUnit.test('Map set data with updated data (#3894)', function (assert) {
         '#15782, mapNav should not overlap with ' +
             'export icon (Bottom right side).'
     );
+
+    // Verify that bounds adapt when setting data (#17013)
+    const worldScale = chart.series[0].transformGroups[0].scaleX;
+    chart.series[0].update({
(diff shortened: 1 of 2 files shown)


raf18seb (approved):
Looks good 👍

TorsteinHonsi:
> Actually, consider this solution https://github.com/highcharts/highcharts/pull/17307

@raf18seb Thanks, it looks connected, but regardless I think the changes in this PR make sense, since redraw should be the last operation of the `setData` function.

PR #17307, opened 2022-05-18
bugfix/17012-map-series-bounds

### LINKED ISSUE #17012: Toggling allAreas on a map doesn't update the map as expected
#### Expected behaviour
Setting `allAreas` through the react wrapper should reflect the correct behaviour.
true: render all available areas.
false: render just the areas where there's data associated.
Forcing re-renders are needed to correctly see the state, but that leads to other problems (the whole map breaking - [ref](https://github.com/highcharts/highcharts/issues/17011))

#### Actual behaviour
Toggling `allAreas` has no effect, and subsequent toggles have the exact opposite effect.
Forcibly rerendering (using the `key` ) works, but breaks the map if it happens during an animation.

https://codesandbox.io/s/hc-map-all-charts-not-updating-iqk00s?file=/src/App.js

#### Product version
* highcharts@9.3.3 & highcharts-react-official@3.1.0
* highcharts@9.3.2 & highcharts-react-official@3.0.0
* probably more

#### Affected browser(s)
All

raf18seb:
Fixed #17012, toggling `allAreas` did not correctly update the bounds of the map.

### DIFF
--- a/samples/unit-tests/maps/setdata/demo.js
+++ b/samples/unit-tests/maps/setdata/demo.js
@@ -805,6 +805,10 @@ QUnit.test('Map set data with updated data (#3894)', function (assert) {
 
     // Initialize the chart
     const chart = Highcharts.mapChart('container', {
+        accessibility: {
+            enabled: false
+        },
+
         title: {
             text: ''
         },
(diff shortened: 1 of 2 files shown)


TorsteinHonsi:
Thanks!

We do not always want `fitToBounds`. When the implementer has set `zoom` and `center`, those should be respected. It works correctly in the current official release: https://jsfiddle.net/highcharts/aw6p5fzs/.

Maybe we should guard it like here:

https://github.com/highcharts/highcharts/blob/eeb61ae1555e5a5eed2638d21b6808b44ee97285/ts/Maps/MapView.ts#L1064-L1066

_But_... Architecture wise, I don't think it's the Series' job to tell the map view to fit the bounds. For example, multiple series would lead to multiple calls to `MapView.fitToBounds`. Instead, we should probably add a hook, maybe to the `chart.redraw` event, to run `fitToBounds` after all the series are translated.

raf18seb:
Right...

I looked into it and it turned out we can't fire `fitToBounds` in `chart.redraw` event because it would require another redraw and got looped. Even if we hack the loop, there would be 2 redraws at least. Any clues?

TorsteinHonsi:
@raf18seb After some digging, I can't see a good way either. The fitting needs to be done after `translate` but before `drawPoints`, but those two functions are called in quick succession from `Series.redraw`. So let's keep the proposed solution, but wrap it in a check for custom `center` and `zoom` options. We might have to check on `userOptions` because I think the options are mutated in the MapView's constructor.

TorsteinHonsi:
Pending reviews by @KacperMadej and @madepiet

KacperMadej (approved):

PR #17316, opened 2022-05-23
Cleaned up some typing in GridAxis

TorsteinHonsi:
Missing time types and any casts

Edit: PR mutated into a general time unit typing cleanup.

### DIFF
--- a/ts/Core/Time.ts
+++ b/ts/Core/Time.ts
@@ -888,7 +888,7 @@ class Time {
         range: number,
         timestamp: number,
         startOfWeek: number,
-        dateTimeLabelFormats: Record<Time.TimeUnit, Time.DateTimeLabelFormatOption>
+        dateTimeLabelFormats: Time.DateTimeLabelFormatsOption
     ): string|undefined {
         const dateStr = this.dateFormat('%m-%d %H:%M:%S.%L', timestamp),
             blank = '01-01 00:00:00.000',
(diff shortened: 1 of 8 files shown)


ts/Core/Time.ts
```diff
@@ -898,49 +898,52 @@ class Time {
...
 
         let format: string|undefined,
-            n,
-            lastN = 'millisecond'; // for sub-millisecond data, #4223
+            n: Time.TimeUnit|undefined,
+            // for sub-millisecond data, #4223
+            lastN: Time.TimeUnit = 'millisecond';
 
         for (n in timeUnits) { // eslint-disable-line guard-for-in
+            if (n) {
```
bre1470, line 909: In my opinion the test is not necessary when it comes to keys.

ts/Core/Time.ts
```diff
@@ -954,6 +957,21 @@ class Time {
...
+        to?: string;
+    }
+    export type DateTimeLabelFormatOption = (
+        string|
+        Array<string>|
+        Time.DateTimeLabelFormatObject
+    );
+    export type DateTimeLabelFormatsOption = (
+        Record<TimeUnit, DateTimeLabelFormatOption>
+    );
```
bre1470, line 974: The naming could be simplified as the Time class has a DTL function, so DTLFormatObject etc.

TorsteinHonsi: I don't know... The name of the internal `resolveDTLFormat` function is an attempt to save filesize, but for the typing I think we should spell it out.

bre1470 (approved):
Looks good! 👍

PR #17319, opened 2022-05-24
bugfix/12206-top-margin

### LINKED ISSUE #12206: Top margin incorrect
Take the following example:

https://jsfiddle.net/8wvkstxh/1/

The chart options include a `y` adjustment to the top `xAxis` labels, because, without it, the labels are too far from the axis itself (for my liking):

https://jsfiddle.net/8wvkstxh/2/

But with the `y` adjustment included (first example above), there is now a lot of dead space above the top x axis labels (I'm trying to create a chart layout which minimises dead margin space and maximises chart area, so I'm sensitive to this).

The top margin is seemingly "correct" for the default y position (without the y adjustment -- second example above), in that there is only a small gap between the labels and the edge.

But with the adjusted y position there is too much margin space.

It seems to me that the problem is that in order to set the y position to my liking (to move the labels **down** from their default position), I need a **negative** value for y.  

But a negative y value actually should normally be pushing the labels **up**... so it seems that highcharts is thinking that more margin space will be required, and adds margin space to compensate for this, even though the labels are actually **lower** than without the negative adjustment (so that the **opposite** should have happened... the top margin should have been reduced, not increased).

Compare this with setting y explicitly to `0`:

https://jsfiddle.net/sw36dqtp/

Therein lies the problem: setting `y: 0` gives a very different result to not setting `y` at all.  I need to set `y` to get the y position right, so I need to state a value, and I'm moving it **up** from `y: 0` so I need to specify a negative value, which highcharts is apparently compensating for by adding more margin space which isn't needed.  This leaves a lot of bloat in the top margin.

pawelfus:
Fixed #12206, `xAxis.labels.y` did not adapt to multiline labels.

### DIFF
--- a/ts/Core/Axis/Tick.ts
+++ b/ts/Core/Axis/Tick.ts
@@ -535,19 +535,23 @@ class Tick {
             ),
             pos = {} as PositionObject;
 
-        let yOffset = labelOptions.y,
+        let yOffset: number,
             line: number;
 
-        if (!defined(yOffset)) {
-            if (axis.side === 0) {
-                yOffset = label.rotation ? -8 : -label.getBBox().height;
-            } else if (axis.side === 2) {
-                yOffset = rotCorr.y + 8;
-            } else {
-                // #3140, #3140
-                yOffset = Math.cos((label.rotation as any) * deg2rad) *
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (approved):
Thanks!

PR #17327, opened 2022-05-25
bugfix/17322-flags-reversed-axis-tooltip 

### LINKED ISSUE #17322: Tooltip drawn on top of flag when Y axis is reversed
#### Expected behaviour
Tooltips should not be drawn on top of the flag they point at (similar issue to #13780).

(Ideally for my use case, the tooltip should be drawn above the flag regardless of the Y axis being reversed.)

#### Actual behaviour
Instead of being drawn right above the top of the flag (as they are when the Y axis is not reversed), the tooltip is drawn right *under* the top of the flag, resulting in an overlap:
Screenshot 2022-05-25 at 10 56 05

https://jsfiddle.net/Lx27wg35/

#### Product version
Highstock v10.1.0

#### Affected browser(s)
Firefox, Chrome

karolkolodziej:
Fixed #17322, when axis was reversed, the tooltip overlapped flags.

### DIFF
--- a/ts/Series/Flags/FlagsPoint.ts
+++ b/ts/Series/Flags/FlagsPoint.ts
@@ -27,10 +27,7 @@ const {
     }
 } = SeriesRegistry;
 import U from '../../Core/Utilities.js';
-const {
-    isNumber,
-    extend
-} = U;
+const { isNumber } = U;
 
 /* *
  *
@@ -64,6 +61,8 @@ class FlagsPoint extends ColumnSeries.prototype.pointClass {
 
     public style?: CSSObject;
 
+    public ttBelow?: boolean = false;
+
     public unbindMouseOver?: Function;
 
     /* *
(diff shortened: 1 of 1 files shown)


ts/Series/Flags/FlagsPoint.ts
```diff
@@ -90,6 +93,16 @@ class FlagsPoint extends ColumnSeries.prototype.pointClass {
...
 
+/* *
+ *
+ *  Class Prototype
+ *
+ * */
+
+extend(FlagsPoint.prototype, {
+    ttBelow: false
+});
```
TorsteinHonsi, line 104: Instead of extending the prototype, we could just add this to the class definition, around line 64:

`public ttBelow = false;`

TorsteinHonsi (changes requested):
Thanks! A nice and simple fix, but I don't think we need to use `extend`

TorsteinHonsi (approved):
Thanks!

PR #17330, opened 2022-05-26
bugfix/17281-bubble-marker-symbol 

### LINKED ISSUE #17281: Bubble chart image symbols not scaled when point z value is below minimum
#### Expected behaviour
When a point in a bubble series is out of the zMin/zMax range for the series and uses an image symbol (`symbol: "url(...)"`), it should be skipped or displayed at a small size.
Screenshot 2022-05-11 at 09 29 04

#### Actual behaviour
The image is instead rendered at the wrong size and in the wrong place.
Screenshot 2022-05-11 at 09 29 12

This error also appears in the console:
> Unexpected value NaN parsing y attribute.

Combined with #17280, points in bubble charts can easily fall out of range and trigger this bug if one of the series has a minimum z value of 0.

https://jsfiddle.net/6v2a1ow4/

#### Product version
Highcharts v10.1.0

#### Affected browser(s)
Firefox, Chrome

karolkolodziej:
Fixed #17281, an error in console when bubble with custom marker was out of `zThreshold`.

### DIFF
--- a/samples/unit-tests/series-bubble/marker/demo.js
+++ b/samples/unit-tests/series-bubble/marker/demo.js
@@ -174,3 +174,26 @@ QUnit.test('Bubble animation and async redraws (#13494)', assert => {
         TestUtilities.lolexUninstall(clock);
     }
 });
+
+QUnit.test('Bubble with custom symbol markers, #17281.', function (assert) {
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #17332, opened 2022-05-26
bugfix/17331-type-type

### LINKED ISSUE #17331: TS: definition for OptionsTypeValue is wrong
#### Behaviour
`OptionsTypeValue` is a merge of `chart.panning.type` and `series.map.data.geometry.type`. Those should not be of the same type. The property name collision (and following unwanted merge) is logged upon running `npx gulp jsdoc-dts`.

Run `npx gulp jsdoc-dts` in the repo.
`OptionsTypeValue` (could be checked [here](https://raw.githubusercontent.com/highcharts/highcharts-dist/05dcbf8e2dbc924c37eb6e8c719fed3e9f953bbd/highcharts.src.d.ts)) shows as:
`export type OptionsTypeValue = ("x"|"xy"|"y"|"LineString"|"MultiLineString"|"MultiPolygon"|"Polygon");`

#### Product version
Highcharts 10.1.0 (probably since v9.3.0)

KacperMadej:
Fixed #17331, resolved TS type collision.

### DIFF
--- a/ts/Series/Map/MapSeries.ts
+++ b/ts/Series/Map/MapSeries.ts
@@ -1543,13 +1543,15 @@ export default MapSeries;
  */
 
 /**
+ * @typedef {"LineString"|"Polygon"|"MultiLineString"|"MultiPolygon"} Highcharts.SeriesMapDataGeometryTypeValue
+ */
+/**
  * The geometry type. Can be one of `LineString`, `Polygon`, `MultiLineString`
  * or `MultiPolygon`.
  *
- * @type      {string}
+ * @type      {Highcharts.SeriesMapDataGeometryTypeValue}
  * @since 9.3.0
  * @product   highmaps
- * @validvalue ["LineString", "Polygon", "MultiLineString", "MultiPolygon"]
  * @apioption series.map.data.geometry.type
  */
 


KacperMadej:
Not sure if this should go into the changelog - I am not adding the label that would make it do so.

KacperMadej:
I have tried changing the added definition location based on other similar definitions but I am still getting this error message. I do not see how the error is related to the introduced change. Any ideas @bre1470 ?

KacperMadej:
If https://github.com/highcharts/highcharts/pull/17354 works this PR can be closed. However, it would be nice to know why it failed.

bre1470:
@KacperMadej `@typedef` is a tag for class reference and `@apioption` is a tag for API reference. Not every file is processed for class reference, so this file might be ignored by the gulp task. I commented in the other PR about the solution to solve conflicts with aliases in a simple way. In general your approach is correct, but this can be automated with a custom tag.

raf18seb:
@TorsteinHonsi Kacper is on vacation for 3 weeks until 11 of July. Can this PR wait that long?

TorsteinHonsi:
@raf18seb Sure, no problem!

TorsteinHonsi:
Pending reply from @KacperMadej

KacperMadej:
Let's continue with the other PR.

PR #17337, opened 2022-05-30
Tools: Updated API docs tool for type reflection fix.

bre1470:
Covers internal issues about lacking type for `chart.alignThresholds` option.


TorsteinHonsi (approved):
Thanks!

Issue #17340, opened 2022-05-30
treemap breadcrumb rtl supprot


kingofday:
it would be great to have support for rtl in treemap breadcrumb after drilldown

karolkolodziej:
Thank you for pointing that out! 
That is a really good idea.

PR #17343, opened 2022-05-31
feature/17340-breadcrumbs-rtl 

karolkolodziej:
Added RTL support for breadcrumbs, see #17340.

### DIFF
--- a/ts/Extensions/Breadcrumbs.ts
+++ b/ts/Extensions/Breadcrumbs.ts
@@ -271,8 +271,10 @@ class Breadcrumbs {
          *
          * @type       {boolean}
          * @since      next
+         * @sample     {highcharts} highcharts/breadcrumbs/rtl
+         *             Breadcrumbs in RTL
          */
-        rtl: void 0,
+        rtl: false,
 
         /**
          * Positioning for the button row. The breadcrumbs buttons will be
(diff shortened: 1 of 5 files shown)


ts/Extensions/Breadcrumbs.ts
```diff
@@ -1103,6 +1135,7 @@ namespace Breadcrumbs {
         format?: string;
         formatter?: BreadcrumbsButtonsFormatter;
         relativeTo?: ButtonRelativeToValue;
+        rtl?: boolean;
```
TorsteinHonsi, line 1138: I think it would be better to make it mandatory and `false` by default. Since only `true` and `false` make any sense.

ts/Extensions/Breadcrumbs.ts
```diff
@@ -265,6 +265,15 @@ class Breadcrumbs {
          */
         relativeTo: 'plotBox',
 
+        /**
+         * Whether to reverse the order of buttons. This is common in Arabic
+         * and Hebrew.
+         *
+         * @type       {boolean}
```
TorsteinHonsi, line 272: A visual sample would also be nice. It would double as a test and a demo. Create a button with `class="autocompare"` to perform the drilldown. Search for "autocompare" in the samples for inspiration.

TorsteinHonsi (changes requested):
Thanks!

TorsteinHonsi (approved):

PR #17354, opened 2022-06-03
bugfix/17331-panning-type-type

### LINKED ISSUE #17331: TS: definition for OptionsTypeValue is wrong

KacperMadej:
Fixed #17331, resolved TS type collision.

---
The collision is caused by the usage of properties with the same key/name. This fix is from the `chart.panning.type` point of view (instead of the attempt in #17332 where I tried with `series.map.data.geometry.type`).


bre1470 (changes requested):
This should be solvable with a `@declare` tag, but does not work:
```
         * @declare    Highcharts.ChartPanningTypeValue
         * @type       {string}
         * @validvalue ["x", "y", "xy"]
         * @default    {highcharts|highstock} x
         * @default    {highmaps} xy
```
I will fix the generator and give you a note.

bre1470:
You can install `github:highcharts/highcharts-declarations-generator#v1.1.22` now and combine `@declare` with `@validvalue` to avoid conflicts.

ts/Core/Chart/Chart.ts
```diff
@@ -4169,4 +4169,10 @@ export default Chart;
  * @type {boolean|undefined}
  */
 
+/**
+ * The dimensions the user can pan the chart.
+ *
+ * @typedef {"x"|"y"|"xy"} Highcharts.ChartPanningTypeValue
+ */
+
```
bre1470, line 4177: You can reduced this in the other doclet to `@declare Highcharts.OptionsChartPanningTypeValue` and keep there @validvalue and @type.

bre1470 (changes requested):

TorsteinHonsi:
@KacperMadej

bre1470 (approved):
Thank you! Looks good. 👍

TorsteinHonsi (approved):

PR #17360, opened 2022-06-06
bugfix/11816-gauge-path

### LINKED ISSUE #11816: Documentation on gauge dial `path`
#### Expected behaviour
There should be documentation about the `path` property in the `dial` property of `gauge` series.
I have seen in the code that the property can be used, and it can be used. But it is unclear to me how to actually use it.
It seems to be hard to make responsive paths like the default one with just the `path` property.
As you can see in the live demo below (if the chart gets bigger or smaller, the custom dial stays the same size, as there are no variables to incorporate into the path)

#### Actual behaviour
- no documentation
- no example usage

https://jsfiddle.net/gj3z2ucr/

#### Product version
latest

#### Affected browser(s)
every

karolkolodziej:
Fixed #11816, in gauge series, added dial `path` into API documentation.


ts/Series/Gauge/GaugeSeries.ts
```diff
@@ -161,118 +161,139 @@ class GaugeSeries extends Series {
...
+                 * @sample {highcharts} highcharts/plotoptions/gauge-dial/
+                 *         Dial options demonstrated
+                 *
+                 * @type      {Highcharts.ColorString|Highcharts.GradientColorObject|Highcharts.PatternObject}
+                 * @default   #cccccc
+                 * @since     2.3.0
+                 * @product   highcharts
+                 * @apioption plotOptions.gauge.dial.borderColor
+                 */
+                borderColor: '#cccccc',
```
TorsteinHonsi, line 223: Use `palette` color like before

samples/highcharts/plotoptions/gauge-path/demo.js
```diff
@@ -0,0 +1,33 @@
...
+        dial: {
+            path: [
+                "M", 0, -10,
+                "l", 100, 0,
+                "l", 0, -10,
+                "l", 20, 20,
+                "l", -20, 20,
+                "l", 0, -10,
+                "l", -100, 0
+            ]
```
TorsteinHonsi, line 27: The one-dimensional array form of paths is deprecated. Use the two-dimensional form.

Also, we prefer single quotes.

TorsteinHonsi (changes requested):
Thanks for the PR!

Moving the default options values from inline picks and `||` operations to the actual options object, is a desired change.

But since we now have a default value for for example `dial.backgroundColor`, we don't have to do fall back to a hard coded value at all:

https://github.com/highcharts/highcharts/blob/8b93aedfa8a8f7063620fb2bbb2e5c9f0ecae49d/ts/Series/Gauge/GaugeSeries.ts#L529-L530

This applies to all those options that have a default value set.

Furthermore, all properties except `path` can be made non-optional in the `GaugeSeriesDialOptions` interface.

TorsteinHonsi (approved):
Awesome, thanks!

PR #17371, opened 2022-06-09
bugfix-17352-tooltip-scrollablePlotArea 

### LINKED ISSUE #17352: Tooltip not working correctly after window resize and scrollablePlotArea must be enabled
#### Expected behaviour
Tooltip works fine. Look like this.
image

#### Actual behaviour
Tooltip missing.
image

[](https://jsfiddle.net/sLhw8n3p/)

Repo steps:
1 click button "small";
2 open console to force redrew;
3 minimize the console panel
4 click button “large”;
5 open console to force redrew;
6 minimize the console panel
7 you can see part of the tooltip disappeared.  
(Seems only V-Scrollbar could cause this issue)

This demo is similar but not exactly same as our case. In our case, after resize the window, some of the tooltip disappear totally. And if 
 you set a positioner callback function, some tooltip will trigger that function while others don`t. The "top" property of the missing tooltip will be "-9999px".

#### Product version

v10.1.0

#### Affected browser(s)
I just tested it on chrome.

PS: I am new to JSFiddle, Just copy the code here in case the demo not working.
HTML:
`<meta name="viewport" content="width=device-width, initial-scale=1" />
<script src="https://code.jquery.com/jquery-3.6.0.min.js"></script>
<script src="https://code.highcharts.com/highcharts.js"></script>

<div id="container"></div>
<button onclick="onLarge()">
Large
</button>
<button onclick="onSmall()">
Small
</button>
`

JS:
`function onLarge() {
$('#container').css("height","400px")
}
function onSmall() {
$('#container').css("height","190px")
}`

chart: {
        type: 'spline',
        scrollablePlotArea: {
            minHeight: 200,
            scrollPositionX: 1
        }
    },

I Just change a little bit of the official demo. 
https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/chart/scrollable-plotarea

Thank you for your time!

karolkolodziej:
Fixed #17352, the tooltip was not visible after updating scrollablePlotArea.

### DIFF
--- a/samples/unit-tests/tooltip/outside/demo.js
+++ b/samples/unit-tests/tooltip/outside/demo.js
@@ -81,3 +81,33 @@ QUnit.test('Outside tooltip styling', function (assert) {
         '#11494: Setting tooltip.style.zIndex should also work'
     );
 });
+
+QUnit.test('Tooltip visibility after the scrollablePlotArea change, #17352.',
+    assert => {
+        const chart = Highcharts.chart('container', {
+            chart: {
+                type: 'spline',
+                scrollablePlotArea: {
+                    minHeight: 200
+                }
+            },
+            series: [{
+                data: [0, 1, 2, 3, 4]
+            }]
+        });
+
+        document.getElementById('container').style.height = "190px";
+        chart.reflow();
+
+        document.getElementById('container').style.height = "400px";
+        chart.reflow();
+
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #17373, opened 2022-06-09
bugfix/17247-drilldown-zoomout 

### LINKED ISSUE #17247: Drilldown chart throws an error and does not return to the original series when drilling up when zooming in before drilling down
#### Expected behaviour

When drilling up the chart should return to the original series and not throw an error.

#### Actual behaviour

When drilling up the chart does not return to the original series and throws the following error.

```
Uncaught TypeError: Cannot convert undefined or null to object
    at hasOwnProperty (<anonymous>)
    at x (highcharts.src.js:1545:17)
    at k.destroy (highcharts.src.js:11364:17)
    at a.<anonymous> (drilldown.src.js:2237:40)
    at highcharts.src.js:1654:25
    at z (highcharts.src.js:1651:17)
    at a.g.drillUp (drilldown.src.js:1681:21)
    at a.<anonymous> (drilldown.src.js:1792:17)
    at highcharts.src.js:1654:25
```

To reproduce from the demo below

1. Open https://jsfiddle.net/iamrewt/18q9xamf/42/ in Chrome or Firefox
2. Open developer tools
3. Zoom in to the chart
4. Click any point
5. Click "TOP SERIES NAME" in the upper left corner to drill up
6. The chart will cease to return to the original series and throw the error shown above

#### Product version

Highcharts 10, Highstock 10

#### Affected browser(s)

Chrome, Firefox. Not tested in Edge.

#### Other notes

This appears to work in Highcharts 8 and is preventing us from migrating to Highcharts 10. Previously, we were blocked on this upgrade due to #16479 .

karolkolodziej:
Fixed #17247, errors while drilling up zoomed chart.

@oysteinmoseng could you please take a look- there is some issue with the accessibility module
Demo: https://jsfiddle.net/BlackLabel/h3kg1w0f/

### DIFF
--- a/samples/unit-tests/drilldown/resetzoom/demo.html
+++ b/samples/unit-tests/drilldown/resetzoom/demo.html
@@ -1,5 +1,6 @@
 <script src="https://code.highcharts.com/highcharts.js"></script>
 <script src="https://code.highcharts.com/modules/drilldown.js"></script>
+<script src="https://code.highcharts.com/modules/accessibility.js"></script>
 
 
 <div id="qunit"></div>
--- a/samples/unit-tests/drilldown/resetzoom/demo.js
+++ b/samples/unit-tests/drilldown/resetzoom/demo.js
@@ -457,3 +457,54 @@ QUnit.test('Drilldown and reset zoom should not crash the chart, #8095.', functi
         'Reset zoom button should be visible again.'
     );
 });
+
+
+QUnit.test(
+    `Reset zoom button with dynamically added multiple drilldowns, #17247.`,
+    function (assert) {
+        const chart = Highcharts.chart('container', {
+            chart: {
+                zoomType: 'x',
+                events: {
+                    drilldown: function (e) {
+                        if (!e.seriesOptions) {
+                            const chart = this;
+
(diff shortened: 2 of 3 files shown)


oysteinmoseng:
@karolkolodziej What is the a11y module issue? No errors are thrown for me testing in Chrome

karolkolodziej:
@oysteinmoseng I apologize, apparently, I put the wrong commit hash. 
In [this demo](https://jsfiddle.net/BlackLabel/hb3tnzks/), it should be visible.

To recreate that issue please:
1. Zoom in
2. Click on the first point to dirlldown
3. Drill up
4. Voilà 

CI won't catch it- I only get that error in karma or with th accessibility module in the demo.

TorsteinHonsi:
Safari in CI has an error in the accessibility module...

karolkolodziej:
@oysteinmoseng ☝️ could you please take a look?

oysteinmoseng:
@karolkolodziej It happens because `chart.resetZoomButton` is destroyed, but the `chart.resetZoomButton` property is still defined on the chart, so it points to a dead object. I could work around it in the a11y module, but it's probably not supposed to be that way...?

PR #17380, opened 2022-06-10
Migration: AreaRangeSeries

bre1470:
- Optimized ES6 imports of AreaRangePoint.
- Optimized ES6 pattern of AreaRangeSeries.
- Adds `high`, `low` properties to `Highcharts.Point` class reference.


TorsteinHonsi (approved):
Thanks!

PR #17389, opened 2022-06-14
bugfix/17279-mappoint-fill-color-nan

### LINKED ISSUE #17279: Maps: mappoint's fill color is black when its value is NaN
#### Expected behaviour
Highcharts 9.2.2 used the minimum color axis value for invalid values

#### Actual behaviour
Highcharts 10+ generates NaN for color axis values when data is invalid

https://jsfiddle.net/Lhdes4gp/1/

#### Product version
HighMaps

#### Affected browser(s)
Any

hubertkozik:
Fixed #17279, mappoint fill color didn't work with `NaN` values.


raf18seb:
@TorsteinHonsi this fix breaks 2 tests, but I think this behavior is more consistent with standard Highcharts now.
Let's discuss these demos:
Highcharts: https://jsfiddle.net/BlackLabel/vsoh9e51/
Highcharts Maps: https://jsfiddle.net/BlackLabel/prnve3th/

For both `null` and `NaN` values, the US and PL colors should be the same (`nullColor`, not black).

What's your opinion on this?

TorsteinHonsi:
@raf18seb  I agree with your conclusions, that `NaN` should be treated as `null`.

hubertkozik:
@TorsteinHonsi there is one (actually 6 tests, but similar) test which found differences in the demo. The demo gets data from this [Google Spreadsheet](https://docs.google.com/spreadsheets/u/0/d/1gXzu9TYT3UvDMcoxj_kS7PUXMmC1MNVSfewccOs2dkA/pub?hl=en_GB&hl=en_GB&hl=en_GB&output=html) where value for South Sudan is an empty cell, which then is parsed to `NaN` value. This bugfix relied on catching NaN's values and stopping coloring them, so in the mentioned demo it is done. I assume that changing the cell value for South Sudan in Google Spreadsheet to value 15, which is the population density of this country in 2010, would be the simplest solution.

TorsteinHonsi (approved):
Yes, absolutely! I agree it all works correctly now. I have updated the spreadsheet.

PR #17402, opened 2022-06-16
enhancement/17270-flowmap-series

DJTechnoo:


docs/maps/flowmap-series.md
```diff
@@ -0,0 +1,62 @@
+Flow map
+===============
+
+A flow map is a series type that allows to display route paths (e.g. flights or ships routes) or flows on a map.
+
+<iframe style="width: 100%; height: 520px; border: none;" src="https://highcharts.com/samples/embed/maps/demo/flowmap" allow="fullscreen"></iframe>
+
+Setting the Flow Map Series
+----------------------------------
+In the above example, our map data is based on regions, so we need to load and add a base map firs:
```
kamil-musialowski, line 10: Typo, `first:`

samples/maps/demo/flowmap-around-the-world/demo.html
```diff
@@ -0,0 +1,7 @@
+<script src="https://code.highcharts.com/maps/highmaps.js"></script>
+<script src="https://code.highcharts.com/maps/modules/exporting.js"></script>
```
kamil-musialowski, line 2: Just a suggestion, maybe we should import _exporting modules_ as the last ones here? Might be a tiny bit better in terms of clarity.

samples/maps/demo/flowmap-around-the-world/demo.html
```diff
@@ -0,0 +1,7 @@
...
+<script src="https://code.highcharts.com/maps/modules/offline-exporting.js"></script>
+<script src="https://code.highcharts.com/maps/modules/sankey.js"></script>
+<script src="https://code.highcharts.com/maps/modules/flowmap.js"></script>
+
```
kamil-musialowski, line 6: Also, what about the accessibility module? Do we use it here, or it's not necessary?

samples/maps/demo/flowmap-ship-route/demo.html
```diff
@@ -0,0 +1,8 @@
+<script src="https://code.highcharts.com/maps/highmaps.js"></script>
+<script src="https://code.highcharts.com/maps/modules/exporting.js"></script>
```
kamil-musialowski, line 2: Same as the previous one, it doesn't affect the code at all - but it might improve the readability, when we "sort" the modules, as maps related, exporting related, accessibility.

But I'd like to hear others opinion on this one.

samples/maps/demo/flowmap/demo.js
```diff
@@ -0,0 +1,111 @@
...
+        chart: {
+            map: topology
+        },
+
+        title: {
+            text: 'Highmaps basic lat/lon demo'
+        },
+
+        subtitle: {
+            text: 'Demo of Highcharts map.'
```
kamil-musialowski, line 16: I think we don't usually put dot at the end of the sentence in _title/subtitle_

ts/Series/FlowMap/FlowMapSeries.ts
```diff
@@ -0,0 +1,834 @@
...
+ *
+ *  Declarations
+ *
+ * */
+
+declare module '../../Core/Series/SeriesOptions' {
+    interface SeriesStateHoverOptions {
+
+    }
+}
```
kamil-musialowski, line 59: Just a question! Should this `interface` be empty? If you could explain what is going on here, that would be great. Thanks!

ts/Series/FlowMap/FlowMapSeries.ts
```diff
@@ -0,0 +1,834 @@
...
+    /* *
+     *
+     *  Static properties
+     *
+     * */
+
+    /**
+     * A flowmap series is a series laid out on top of a map series allowing to
+     * display route paths (e.g. flights or ships routes) or flows on a map. It
+     * is drawing a link between two mappoints.
```
kamil-musialowski, line 81: Maybe a better description would be _"It creates a link between two mappoints."_ ?

ts/Series/FlowMap/FlowMapSeries.ts
```diff
@@ -0,0 +1,834 @@
...
+            /**
+             * The HTML of the tooltip header line. Variables are enclosed by
+             * curly brackets. Available variables are point.key, series.name,
+             * series.color and other members from the point and series
+             * objects. The point.key variable contains the category name, x
+             * value or datetime string depending on the type of axis. For
+             * datetime axes, the point.key date format can be set using
+             * tooltip.xDateFormat.
+             *
+             * TODO: check why it's not generated in API
```
kamil-musialowski, line 168: Is this one resolved or should we investigate on it?

ts/Series/FlowMap/FlowMapSeries.ts
```diff
@@ -0,0 +1,834 @@
...
+             */
+            headerFormat: '<span style="font-size: 10px">{series.name}</span><br/>',
+            pointFormat: '{point.options.from} \u2192 {point.options.to}: <b>{point.weight}</b><br/>'
+        },
+
+        /**
+         * The weight of the link determines how thick it will be compared to
+         * other weights.
```
kamil-musialowski, line 176: Maybe a better description would be to say:

The weight of the link determines how thick it will be compared to other links weight (or thickness?).

ts/Series/FlowMap/FlowMapSeries.ts
```diff
@@ -0,0 +1,834 @@
...
+            width = this.getLength(
+                rCorner[0] - lCorner[0],
+                rCorner[1] - lCorner[1]) * percentage / 100;
+        }
+
+        let path: SVGPath = [];
+
+        // For arrow head calculation
+        if (type === 'arrow') {
+
```
kamil-musialowski, line 236: By looking at the other `if` statements in this file, I think this line break is unnecessary here.

ts/Series/FlowMap/FlowMapSeries.ts
```diff
@@ -0,0 +1,834 @@
...
+                [
+                    'Q',
+                    arcPointX + wX,
+                    arcPointY + wY,
+                    fromX + fromXToArc,
+                    fromY + fromYToArc
+                ],
+                ['Z']
+            ]
+        };
```
kamil-musialowski, line 601: Can we use formatting like this: `],[` instead of the current one? We could save a bit space, and the code might get more readable.

ts/Series/FlowMap/FlowMapSeries.ts
```diff
@@ -0,0 +1,834 @@
...
+ * type, points can be given in the following ways:
+ *
+ * 1.  An array of arrays with 6 values. In this case, the values correspond
+ *     to `from, to, weight`. Example:
+ *     ```js
+ *     data: [
+ *         ["Point 1", "Point 2", 4],
+ *     ]
+ *     ```
+ *
```
kamil-musialowski, line 686: We talk about 6 values in the array, but only three are given. That might be a bit confusing for the users.

ts/Series/FlowMap/FlowMapSeries.ts
```diff
@@ -0,0 +1,834 @@
...
+ * @sample {highmaps} maps/demo/flowmap-ship-route/
+ *         Example ship route
+ *
+ * @type      {number}
+ * @apioption series.flowmap.data.curveFactor
+ */
+
+/**
+ * The fill color of an individual link
```
kamil-musialowski, line 722: In general, there is some inconsistency when it comes to code comments, and some jsdocs too.
I didn't want to mark every one, but mainly it's about some missing dots at the end of sentences.

kamil-musialowski (changes requested):
Overall, fantastic job on this one! I only made some comments about styling, spelling and code formatting, so nothing major.

Good job!

ts/Series/FlowMap/FlowMapSeries.ts
```diff
@@ -0,0 +1,834 @@
...
+    /**
+     * Get point attributes.
+     * @private
+     */
+    public pointAttribs(
+        point: FlowMapPoint,
+        state: StatesOptionsKey
+    ): SVGAttributes {
+        const attrs =
+                MapSeries.prototype.pointAttribs.call(this, point, state);
```
hubertkozik, line 348: I think there should be one tabulator less.

hubertkozik (changes requested):
Great job! Actually, Kamil has found probably all the small mistakes :)

ts/Series/FlowMap/FlowMapSeries.ts
```diff
@@ -0,0 +1,834 @@
...
+ * */
+
+import type FlowMapSeriesOptions from './FlowMapSeriesOptions';
+import type { MarkerEndOptions } from './FlowMapPointOptions';
+import type SVGPath from '../../Core/Renderer/SVG/SVGPath';
+import FlowMapPoint from './FlowMapPoint.js';
+import MapSeries from '../Map/MapSeries.js';
+import SeriesRegistry from '../../Core/Series/SeriesRegistry.js';
+
+const {
```
magdalena-gut, line 25: There is a new way of series import, that we should probably follow, e.g.: https://github.com/highcharts/highcharts/pull/17618/files#diff-dfdcaa3c1114b26145bb94ad79d0d629864c59984464a77fe50398140163097c

ts/Series/FlowMap/FlowMapPoint.ts
```diff
@@ -0,0 +1,118 @@
...
+ * */
+
+import type FlowMapPointOptions from './FlowMapPointOptions';
+import type FlowMapSeries from './FlowMapSeries';
+import type Point from '../../Core/Series/Point';
+
+import ColorMapComposition from '../ColorMapComposition.js';
+import { PointShortOptions } from '../../Core/Series/PointOptions';
+import SeriesRegistry from '../../Core/Series/SeriesRegistry.js';
+const {
```
magdalena-gut, line 26: The way of Import (same as in the case of FlowMapSeries.ts)

magdalena-gut (commented):
PR description should be updated for the future changelog

magdalena-gut (changes requested):
Nice work! Just a couple of comments above

ts/Series/FlowMap/FlowMapSeries.ts
DJTechnoo: I agree with your suggestion, but personallly I think it's more readable for me. `SankeySeries` also has the same formatting. Maybe we should decide and have a lint rule about this?

kamil-musialowski: That's a good idea, I think we should have a general rule for that (and for objects, if we already don't have one?).

DJTechnoo: I've been doing some serious and deep thinking about this. In order to resolve this, I think the `], [` should be honored

samples/maps/demo/flowmap/demo.js
```diff
@@ -0,0 +1,143 @@
...
+            data: [{
+                id: "Oslo",
+                lat: 60.1975501,
+                lon: 11.1004152
+            }, {
+                id: 'Warszawa',
+                lat: 52.169192,
+                lon: 20.973514
+            }, {
+
```
madepiet, line 46: Blank space needs to be removed.

ts/Series/FlowMap/FlowMapSeries.ts
```diff
@@ -0,0 +1,829 @@
...
+
+        // Weight vector calculation for the middle of the curve.
+        let [wX, wY] = FlowMapSeries.normalize(dX, dY);
+
+        // The `fineTune` prevents an obvious mismatch along the curve.
+        const fineTune = 1 + Math.sqrt(curveFactor * curveFactor) * 0.25;
+        wX *= scaledWeight * fineTune;
+        wY *= scaledWeight * fineTune;
+
+        // Ccalculate the arc strength.
```
madepiet, line 529: Typo Cc

ts/Series/FlowMap/FlowMapSeries.ts
```diff
@@ -0,0 +1,829 @@
...
+            );
+
+        tmp = fromXToArc;
+        fromXToArc = fromYToArc;
+        fromYToArc = -tmp;
+
+        fromXToArc *= scaledWeight;
+        fromYToArc *= scaledWeight;
+
+
```
madepiet, line 547: Blank space needs to be removed.

ts/Series/FlowMap/FlowMapSeriesOptions.d.ts
```diff
@@ -0,0 +1,40 @@
...
+ *
+ *  !!!!!!! SOURCE GETS TRANSPILED BY TYPESCRIPT. EDIT TS FILE ONLY. !!!!!!!
+ *
+ * */
+
+import MapLineSeriesOptions from '../MapLine/MapLineSeriesOptions';
+import type { MarkerEndOptions } from './FlowMapPointOptions';
+
+/* *
+ *
```
madepiet, line 15: This comment should be moved above imports.

ts/Series/FlowMap/FlowMapPointOptions.d.ts
```diff
@@ -0,0 +1,47 @@
+/* *
+ *
+ *  FlowMap
```
madepiet, line 3: Is it needed here?

ts/Series/FlowMap/FlowMapSeries.ts
```diff
@@ -0,0 +1,829 @@
...
+             * @type  {number|string}
+             */
+            height: '40%',
+            /**
+             * Width of a flowmap point's symbol.
+             *
+             * @since next
+             * @type  {number|string}
+             */
+            width: '40%'
```
KacperMadej, line 138: Info about the percentage base (percent of what?) would be nice.

ts/Series/FlowMap/FlowMapSeries.ts
```diff
@@ -0,0 +1,829 @@
...
+     *
+     * */
+
+    /**
+     * A flowmap series is a series laid out on top of a map series allowing to
+     * display route paths (e.g. flights or ships routes) or flows on a map. It
+     * is drawing a link between two mappoints.
+     *
+     * @extends      plotOptions.mapline
+     * @excluding    dataLabels
```
KacperMadej, line 84: Excluding list should include all the options that are excluded at the series level for consistency.

ts/Series/FlowMap/FlowMapSeries.ts
```diff
@@ -0,0 +1,829 @@
...
+ *             markerEnd: {
+ *                 enabled: true,
+ *                 height: 15,
+ *                 width: 8
+ *             }
+ *         }]
+ *     ```
+ *
+ * @type      {Array<number|null|*>}
+ * @excluding labelrank, outgoing
```
KacperMadej, line 700: Why `outgoing` is excluded? This looks like not needed and not yet implemented. Sankey leftover?

ts/Series/FlowMap/FlowMapSeries.ts
```diff
@@ -0,0 +1,829 @@
...
+    public static getLength(x: number, y: number): number {
+        return Math.sqrt(x * x + y * y);
+    }
+
+    /**
+     * Return a normalized vector.
+     * @private
+     */
+    public static normalize(x: number, y: number): Array<number> {
+        const length = FlowMapSeries.getLength(x, y);
```
KacperMadej, line 204: Why not the simpler `this.getLength`? Doesn't look like it needs to be called on the class.

KacperMadej (changes requested):
- [x] 1. The flowmap-around-the-world demo is not used anywhere and <s>looks bad</s> highlights limitations that should be addressed in future development. It might be better to move it to `highcharts/studies/` directory for now.
- [x] 2. If this series type is ready then shouldn't it be possible to rewrite [flight routes](https://www.highcharts.com/demo/maps/flight-routes) demo to use `flowmap`?
- [x] 3. `series.flowmap.data` is missing API docs for `from` and `to`.
- [x] 4. `flowmap.affectsMapView` is documented (shows in local API docs) but not working - exclude or fix.

ts/Series/FlowMap/FlowMapSeries.ts
madepiet: @magdalena-gut after checking it, it turned out that the current way of importing it is correct. The presented method is rather a workaround in cases where it is not correctly interpreted by TypeScript and when, for example, it starts complaining about the type scope when generating native Declarations from our sources. In my opinion, the current way is more readable for this example

DJTechnoo:
> * [ ]  2. If this series type is ready then shouldn't it be possible to rewrite [flight routes](https://www.highcharts.com/demo/maps/flight-routes) demo to use `flowmap`?

Do we want to replace it? That demo shows what MapLine can do, so should we instead make a FlowMap version?

ts/Series/FlowMap/FlowMapPoint.ts
```diff
@@ -0,0 +1,118 @@
...
+        const point = super.applyOptions.call(this, options, x) as any;
+
+        // Flowmap point doesn't need `value`, but if it's not there, the point
+        // is treated as a null point. So we just set `value: 0`.
+        if (point.value === null) {
+            point.value = 0;
+        }
+
+        return point;
+    }
```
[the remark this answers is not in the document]
hubertkozik: `applyOptions` can be removed, because this logic was written before changing the `isValid` function, which now is taking care of null points in case if weight is set to `null`. I have added test for it also.

ts/Series/FlowMap/FlowMapPoint.ts
```diff
@@ -0,0 +1,118 @@
...
+}
+
+/* *
+ *
+ *  Class Prototype
+ *
+ * */
+
+interface FlowMapPoint {
+    isValid: ColorMapComposition.PointComposition['isValid'];
```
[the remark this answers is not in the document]
hubertkozik: This is true because it has the same types. Furthermore, Flowmap is inheriting from MapLinePoint -> MapPoint, which is also taking `isValid` type from `ColorMapComposition.PointComposition`, so I think it is the correct way to do this. ColorAxis is not working with Flowmap, due to `fillColor` etc. properties, which are not taken into account by colorAxis. I think we should leave it as it is now, is it ok for you?

ts/Series/FlowMap/FlowMapSeries.ts
hubertkozik: I think `labelrank` also is not needed to exclude. When removed nothing has changed in API.

madepiet (approved):

magdalena-gut (approved):

samples/maps/demo/flowmap-flight-routes/demo.js
```diff
@@ -0,0 +1,197 @@
...
+        subtitle: {
+            text: 'Demo of Highcharts map with flowmap'
+        },
+
+        mapNavigation: {
+            enabled: true
+        },
+
+        tooltip: {
+            formatter: function () {
```
TorsteinHonsi, line 21: Use `headerFormat` and `pointFormat` instead. Make one for each series if checking for `this.point.lat` is a problem.

samples/maps/demo/flowmap-flight-routes/demo.js
```diff
@@ -0,0 +1,197 @@
...
+                    align: 'left',
+                    x: 5,
+                    verticalAlign: 'middle'
+                }
+            }]
+        }, {
+            type: 'flowmap',
+            linkedTo: 'cities',
+            minWeight: 8,
+            maxWeight: 12,
```
TorsteinHonsi, line 97: That's quite heavy compared to the usual flight routes map found on the Internet™️. The usual is 1px or 2px lines. Besides, we're missing a demo of how to display flows as simple lines - the three proposed demos all have fat arrows with `growTowards`.

samples/maps/demo/flowmap-flight-routes/demo.js
```diff
@@ -0,0 +1,197 @@
...
+                curveFactor: 0.2
+            }, {
+                id: 'Lerwick - Liverpool',
+                from: 'Lerwick',
+                to: 'Liverpool',
+                weight: 0.1,
+                curveFactor: -0.3
+            }]
+        }, {
+            type: 'flowmap',
```
TorsteinHonsi, line 136: Why two series? I assume it's because of the coloring. Maybe give them a name, `From Lerwick` and `From London` to give it an explanation?

samples/maps/demo/flowmap-ship-route/demo.js
```diff
@@ -0,0 +1,158 @@
...
+            between cities`
+        },
+
+        mapNavigation: {
+            enabled: true
+        },
+
+        mapView: {
+            center: [10, 20],
+            zoom: 2.2
```
TorsteinHonsi, line 26: On mobile the flows are cropped. We really should have `mapView.bounds` option...

samples/maps/demo/flowmap-ship-route/demo.js
```diff
@@ -0,0 +1,158 @@
...
+        'https://code.highcharts.com/mapdata/custom/world.topo.json'
+    ).then(response => response.json());
+
+    Highcharts.mapChart('container', {
+        chart: {
+            map: topology
+        },
+
+        title: {
+            text: 'Example route of the ship from USA to China'
```
TorsteinHonsi, line 12: "of _a_ ship"

samples/maps/demo/flowmap-ship-route/demo.details
```diff
@@ -0,0 +1,10 @@
+---
+name: Example route of the ship from USA to China
```
TorsteinHonsi, line 2: "... of _a_ ship... "

samples/maps/demo/flowmap-ship-route/demo.js
```diff
@@ -0,0 +1,158 @@
...
+                if (this.series.name === 'Ship Routes') {
+                    return tooltip.defaultFormatter.call(this, tooltip);
+                }
+                return '<b>' + this.point.id + '</b>' + (
+                    this.point.lat ?
+                        '<br>Lat: ' + this.point.lat +
+                        ' Lon: ' + this.point.lon :
+                        ''
+                );
+            }
```
TorsteinHonsi, line 40: Instead of a formatter, use `tooltip.headerFormat` and `tooltip.pointFormat` and set specific `tooltip` options either on `plotOptions` or on the series items.

samples/maps/demo/flowmap-ship-route/demo.js
```diff
@@ -0,0 +1,158 @@
...
+        chart: {
+            map: topology
+        },
+
+        title: {
+            text: 'Example route of the ship from USA to China'
+        },
+
+        subtitle: {
+            text: `The route through the Suez Canal with a number of travel days
```
TorsteinHonsi, line 16: "with _the_ number"

samples/maps/demo/flowmap/demo.js
```diff
@@ -0,0 +1,142 @@
...
+        'https://code.highcharts.com/mapdata/custom/europe.topo.json'
+    ).then(response => response.json());
+
+    Highcharts.mapChart('container', {
+        chart: {
+            map: topology
+        },
+
+        title: {
+            text: 'Highmaps basic lat/lon demo'
```
TorsteinHonsi, line 12: Wrong title

samples/maps/demo/flowmap/demo.js
```diff
@@ -0,0 +1,142 @@
...
+        },
+
+        subtitle: {
+            text: 'Demo of Highcharts map'
```
TorsteinHonsi, line 16: Wrong subtitle?

samples/maps/demo/flowmap/demo.js
```diff
@@ -0,0 +1,142 @@
...
+        },
+
+        mapNavigation: {
+            enabled: true
+        },
+
+        series: [{
+            name: 'Basemap',
```
TorsteinHonsi, line 24: This shows in the legend, but doesn't make sense. Probably better to just set `legend.enabled` to false.

samples/maps/demo/flowmap/demo.js
```diff
@@ -0,0 +1,142 @@
...
+            }
+        }, {
+            type: 'mappoint',
+            id: 'europe',
+            name: 'Cities',
+            dataLabels: {
+                format: '{point.id}'
+            },
+            data: [{
+                id: "Oslo",
```
TorsteinHonsi, line 38: Some id's have double quotes, some have single. The style guide prefers single.

TorsteinHonsi (changes requested):
Thanks for all the great work! In addition to the inline comments, here are some discussion points:

* **More samples**. As noted inline, we must have visual samples for all or most of the options. But the main demos also are quite similar visually (weighted lines, grow towards), so it would be great if we can come up with some more variation. Go to Google Images for inspiration.
* **Initial animation**. It would be nice with some juice. One idea is to animate the `scale` of each link individually, where the origin of scaling is the `from` point.
* **Curves**. Also consider a default `curveFactor`. Doesn't have to be perfect, but most likely it looks better than straight lines.

samples/maps/demo/flowmap-ship-route/demo.js
TorsteinHonsi: Implemented here: https://github.com/highcharts/highcharts/pull/18123

ts/Series/FlowMap/FlowMapSeries.ts
```diff
@@ -0,0 +1,868 @@
...
+            point.oldFromPoint = fromPoint;
+            point.oldToPoint = toPoint;
+
+            // Calculate point shape
+            point.shapeType = 'path';
+            point.shapeArgs = this.getPointShapeArgs(point);
+
+            // plotX and plotY need to be defined for dataLabels. (#15863)
+            point.y = point.plotY = 1;
+            point.x = point.plotX = 1;
```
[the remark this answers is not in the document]
hubertkozik: I have found, that it is a leftover from Sankey's inheritance (an idea from the beginning of the series implementation). `plotX` and `plotY` is needed to show point, so I will change it to `plotX`/`plotY` of `from` position.

TorsteinHonsi (changes requested):
Thanks for the changes! I think the default `curveFactor` looks great.

Regarding the demo, they must be better organized. The contents of `samples/maps/demo` will display among the website demos, but we don't want that for demos that only show a small detail. So you should pick one or two main samples to put in the demos, the other ones should be in `samples/maps/series-flowmap`.

KacperMadej (approved):

hubertkozik (approved):

kamil-musialowski (approved):

TorsteinHonsi (changes requested):
For the sample paths on the form `samples/highcharts/series-flowmap/flowmap-*`, the word flowmap is now repeated. It doesn't have to be. For reference, see how we do it for existing series types. For example, `highcharts/series-flowmap/curve-factor` is good, without repetition.

samples/maps/series-flowmap/ship-route/demo.details
```diff
@@ -0,0 +1,10 @@
+---
+name: Example route of a ship from USA to China
+author:
+  - Hubert Kozik
+js_wrap: b
+tags:
+  - Highcharts Maps demo
+categories:
+  - General
```
TorsteinHonsi, line 9: Remove this

ts/Series/FlowMap/FlowMapPoint.ts
```diff
@@ -0,0 +1,96 @@
...
+     *  Functions
+     *
+     * */
+
+    /**
+     * @private
+     */
+    public isValid(): boolean {
+        return isString(this.options.from) &&
+            isString(this.options.to) &&
```
TorsteinHonsi, line 74: Does the `isString` test cover the case when `from` and `to` are objects or arrays of longitude and latitude?

TorsteinHonsi (changes requested):
Nice work folks, we're steadily on track for approval here. Some more nitpicking for you to chew on 👇

ts/Series/FlowMap/FlowMapSeries.ts
```diff
@@ -0,0 +1,1019 @@
...
+        /**
+         * The opacity of all the links.
+         *
+         * @apioption plotOptions.flowmap.opacity
+         */
+
+        /**
+         * The weight for all links with unspecified weights.
+         * The weight of a link determines its thickness compared to
+         * other links.
```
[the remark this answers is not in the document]
DJTechnoo: Currently, everything is forced to be scaled by default.
Instead of adding a new option,  we could use `weight`, remove the default scaling (remove the default values for `minWidth` and `maxWidth`). If a check is added, the scaling will only be activated when they are actually set. In theory it should accomplish the above

TorsteinHonsi (changes requested):
It seems like we don't have an option to set the `lineWidth` for the flow series? The `map` and `mapline` series handle this by setting it on the series-group level, with possible overrides per point. I assume that functionality can just be inherited by the flow map.

TorsteinHonsi (changes requested):
Hei, while working on a demo I just noticed I think the `weight` is wrong. [In this jsFiddle](https://jsfiddle.net/highcharts/x9yr245m/), the max width is set to 15, but the widest arrow seems to be more like 30 pixels. Maybe we forgot to divide by 2 somewhere and added the width on both sides of the arrow?

KacperMadej:
Adding Øystein review request for an a11y check.

TorsteinHonsi (approved):
Approved 🎉  Now pending CI and @oysteinmoseng 's a11y review.

oysteinmoseng (changes requested):
Great work, and mostly works as expected out of the box! Some smaller fixes should be made, mostly demo config, but also a few core changes. Just let me know if you need input or pair programming on anything!

**Code changes/a11y module changes:**
- X should not be part of default x description for flow points. See variable [`showXDescription`](https://github.com/highcharts/highcharts/blob/master/ts/Accessibility/Components/SeriesComponent/SeriesDescriber.ts#L482) in SeriesDescriber.
- In many cases, the map point series get a default description like "London, undefined. Cities." It would be good to see if we can avoid adding the "undefined" by default. Not sure how this should be fixed, but start looking from [here](https://github.com/highcharts/highcharts/blob/master/ts/Accessibility/Components/SeriesComponent/SeriesDescriber.ts#L489).


**Unclear:**
- In from-to-lon-lat demo the one flow has the label "x, 2, from: [object Object], to: [object Object], weight: undefined". The "[object Object]" should be figured out, is this something we should fix in the a11y module, or is it a special case that we should just configure in the demo?


**Demo changes:**
- In the flowmap-migration demo it looks like most of the migration-flows have undefined weight? See in tooltip. Also affects a11y, since the weight is reported as undefined.
- Point.weight is also undefined in flowmap-world-cup, and flight-routes. This does affect a11y. If it should be undefined, you can configure valueDescriptionFormat (see next point).
- valueDescriptionFormat should be configured for some of these series in various demos, follow the "Is my chart accessible" Slab article, and either inspect elements or test with a screen reader. Also for flight-routes demo there is no need for the series name after the data point descriptions, can be configured as well. Do this after fixing core changes, since that will impact the default descriptions.
- seriesDescription would be good to configure as well in some demos, since there are some hidden series. Follow the same Slab article.
- Not so much an a11y thing, but the lat/lon precision is pretty wild in some cases! Does impact screen reader experience & cognitive a11y. May get away with a few decimal places less, to make it easier to read?

docs/maps/flowmap-series.md
```diff
@@ -0,0 +1,83 @@
+Flow map
+===============
+
+A flow map is a series type that allows to display route paths (e.g. flights or ships routes) or flows (e.g. import/export between regions) on a map.
```
DJTechnoo, line 4: Note: Maybe change "flights" and "ships" -> "flight and ship routes"

oysteinmoseng (approved):
Great!

oysteinmoseng:
Looks like we need to fix conflicts before merge, however @DJTechnoo

PR #17404, opened 2022-06-17
Demos/demos data update

fanuntez:
Updated demo data.


samples/highcharts/demo/pie-monochrome/demo.js
```diff
@@ -21,7 +22,7 @@ Highcharts.chart('container', {
         type: 'pie'
     },
     title: {
-        text: 'Browser market shares at a specific website, 2014'
+        text: 'Browser market shares in february, 2022'
```
TorsteinHonsi, line 25: Upper case first, February

TorsteinHonsi (changes requested):
Thanks, looks good!

* In the column-drilldown sample, the data was previously sorted with the most popular browser first. I think this should be done with the new data set too.
* Also, for all the pie series, I think they used to be sorted with the most popular first.

samples/highcharts/demo/column-stacked/demo.js
```diff
@@ -27,9 +31,9 @@ Highcharts.chart('container', {
     },
     legend: {
         align: 'right',
-        x: -30,
+        x: -400,
```
TorsteinHonsi, line 34: When the chart is small (like in visual test), the legend comes off the chart area. Better align left and with a smaller x offset.

samples/highcharts/demo/spline-irregular-time/demo.js
```diff
@@ -42,123 +44,281 @@ Highcharts.chart('container', {
...
-            [Date.UTC(1971, 4,  6), 0.96],
-            [Date.UTC(1971, 4,  9), 0.87],
-            [Date.UTC(1971, 4, 12), 0.88],
-            [Date.UTC(1971, 4, 15), 0.79],
-            [Date.UTC(1971, 4, 18), 0.54],
-            [Date.UTC(1971, 4, 21), 0.34],
-            [Date.UTC(1971, 4, 25), 0]
-        ]
-    }],
+    series: [
```
TorsteinHonsi, line 47: This demo previously had 3 series, now 4 is proposed. But I think it looks a bit cluttered with four series - better to remove one of them?

samples/highcharts/demo/area-missing/demo.js
```diff
@@ -1,32 +1,33 @@
...
     chart: {
         type: 'area'
     },
     title: {
-        text: 'Fruit consumption *'
+        text: 'Born persons, by boys\' name'
     },
     subtitle: {
-        text: '* Jane\'s banana consumption is unknown',
+        text: '* The name Yasin was not given this year.',
```
TorsteinHonsi, line 10: This explanation indicates there were 0 instances of Yasin this year. But there is a fundamental difference between 0 and null, where 0 is a real value that would be plotted in the area chart, and `null` is missing data. Better change the text to `* Missing data for Yasin this year`

samples/highcharts/demo/areaspline/demo.js
```diff
@@ -1,45 +1,64 @@
...
-        y: 100,
+        x: 120,
+        y: 70,
         floating: true,
         borderWidth: 1,
         backgroundColor:
             Highcharts.defaultOptions.legend.backgroundColor || '#FFFFFF'
     },
     xAxis: {
         categories: [
```
TorsteinHonsi, line 25: The number of categories makes the X axis cluttered, plus the nature of this axis is numerical/temporal. I would use a numerical axis in this case, and instead let the tooltip say something like `Hunting season starting autumn 2020` to indicate that it spans over two calendar years?

samples/highcharts/demo/3d-column-stacking-grouping/demo.js
```diff
@@ -11,11 +12,11 @@ Highcharts.chart('container', {
     },
 
     title: {
-        text: 'Total fruit consumption, grouped by gender'
+        text: ' Electricity production in countries, grouped by continent'
     },
 
     xAxis: {
-        categories: ['Apples', 'Oranges', 'Pears', 'Grapes', 'Bananas'],
+        categories: ['2016', '2017', '2018', '2019', '2020'],
```
TorsteinHonsi, line 19: Better use a linear axis here. Remove the categories and instead add something like this:

```js
plotOptions: {
    series: {
        pointStart: 2016
    }
}
```

samples/highcharts/demo/advanced-accessible/demo.js
```diff
@@ -77,100 +79,59 @@ var chart = Highcharts.chart('container', {
...
-
     responsive: {
         rules: [{
             condition: {
                 maxWidth: 550
             },
             chartOptions: {
                 xAxis: {
-                    categories: ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec']
+                    categories: ['Finland', 'Iceland', 'Norway', 'Sweden', 'Japan', 'Russia', 'France', 'UK', 'Canada']
```
TorsteinHonsi, line 110: The whole `responsive` section is redundant, as it does not change the categories. The point of the original responsive setting was to use shorter names for the months. In case of the countries, you can also replace them by two- or three-letter ISO codes.

samples/highcharts/demo/area-inverted/demo.js
```diff
@@ -25,33 +26,28 @@ Highcharts.chart('container', {
...
-        categories: [
-            'Monday',
-            'Tuesday',
-            'Wednesday',
-            'Thursday',
-            'Friday',
-            'Saturday',
-            'Sunday'
-        ]
+        categories: ['2014', '2015', '2016', '2017', '2018', '2019', '2020', '2021']
```
TorsteinHonsi, line 29: Do not use categories for numerical/linear data

samples/highcharts/demo/area-inverted/demo.js
```diff
@@ -25,33 +26,28 @@ Highcharts.chart('container', {
...
     },
     yAxis: {
-        title: {
-            text: 'Number of units'
+        labels: {
+            formatter(axis) {
+                return axis.value / 100 + ' B';
```
TorsteinHonsi, line 34: Better give the data itself in Billons so we don't need the formatter (we prefer `format`). Also, the tooltip should state the units of the number.

samples/highcharts/demo/area-missing/demo.js
```diff
@@ -1,32 +1,33 @@
...
+        x: 150,
+        y: 60,
         floating: true,
         borderWidth: 1,
         backgroundColor:
             Highcharts.defaultOptions.legend.backgroundColor || '#FFFFFF'
     },
     xAxis: {
-        categories: ['Apples', 'Pears', 'Oranges', 'Bananas', 'Grapes', 'Plums', 'Strawberries', 'Raspberries']
+        categories: ['2014', '2015', '2016', '2017', '2018', '2019', '2020', '2021']
```
TorsteinHonsi, line 26: Don't use categories for a numeric axis.

TorsteinHonsi (changes requested):
Awesome work on replacing the fruit consumption data with more meaningful data!

Please see my inline comments.

Going forward, please only work review feedback in this PR, so that the reviewer can keep track of what has been changed since the last review. Once the review process has started, it is better to open up a new, parallel PR to work on other demos.

samples/highcharts/demo/advanced-accessible/demo.js
```diff
@@ -77,100 +79,59 @@ var chart = Highcharts.chart('container', {
...
     },
-
     plotOptions: {
         series: {
             stickyTracking: true,
             whiskerWidth: 5
         }
     },
-
     series: [{
```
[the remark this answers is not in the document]
oysteinmoseng: It's not overly critical that it has multiple series, but it does need to have `null` points. It is supposed to be a more complex demo, so it would be good if it had at least 2 series.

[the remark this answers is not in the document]
fanuntez: have changed back to the old data for the time being. will look for more relevant data.

TorsteinHonsi (approved):
All good, thanks! I've done some final minor adjustments, then we'll let CI run and I'll approve all the visuals, then merge.

PR #17408, opened 2022-06-20
Improved AreaSplineSeries, AreaSplineRangeSeries, SplineSeries

bre1470:


ts/Series/AreaSpline/AreaSplineSeries.ts
```diff
@@ -143,12 +107,56 @@ SeriesRegistry.registerSeriesType('areaspline', AreaSplineSeries);
...
+ *
+ * @apioption plotOptions.areaspline.fillColor
+ */
+
+/**
+ * @see [color](#plotOptions.areaspline.color)
+ * @see [fillColor](#plotOptions.areaspline.fillColor)
+ *
+ * @default   {highcharts} 0.75
+ * @default   {highstock} 0.75
```
TorsteinHonsi, line 156: Does this have to be repeated? I thought `@default 0.75` would do?

bre1470: Good catch! It was all over the place.

TorsteinHonsi (approved):

PR #17409, opened 2022-06-20
Improved structure of ScatterSeries.

bre1470:


bre1470:
Unsure why compression increases when sources decrease with less imports.

TorsteinHonsi (approved):
Thanks!

PR #17418, opened 2022-06-21
Fixed #17182, Boost and hiding series in FF

### LINKED ISSUE #17182: Hide or show series using boost mode with firefox doesn't work properly
#### Expected behaviour

Hide and show series when using boost module with firefox

#### Actual behaviour

Series are not correctly hidden or shown on firefox, works on Chrome 

I found something strange in the html there is many `<image class="highcharts-boost-canvas">` in the `<g class="highcharts-series-group">`, images with empty href corresponding to the hidden series are still displayed.

I found that refreshing these images by hidding and showing them (with jquery by example) fix the problem.

https://jsfiddle.net/q3us4m0w/

#### Product version
Highcharts + Boost module v10.0.0 

#### Affected browser(s)

Firefox

TorsteinHonsi:
Fixed #17182, hiding and showing boosted series in Firefox didn't work correctly.

### DIFF
--- a/samples/unit-tests/boost/enter-exit/demo.js
+++ b/samples/unit-tests/boost/enter-exit/demo.js
@@ -1,7 +1,11 @@
 (function () {
     function assertNonBoosted(assert, s) {
         assert.ok(
-            !s.renderTarget || s.renderTarget.attr('href') === '',
+            !s.renderTarget || (
+                s.renderTarget.attr('href') ===
+                // Blank pixel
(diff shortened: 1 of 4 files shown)


raf18seb (commented):
Looks good, it's working fine now.

Only one question - I'm not familiar with boost so I'm assuming that empty pixel is good enough in `boostClear`. But what about other places where we set `href` attribute? For example here: https://github.com/highcharts/highcharts/blob/d5137ac2988803ae63116e0f00038959984adba1/ts/Extensions/BoostCanvas.ts#L288

raf18seb (approved):
👍

PR #17421, opened 2022-06-21
Migration of polar module

bre1470:
- Removed internal namespace.
- Migrated let to const.
- Migrated to simplified composition pattern.


bre1470:
Visual test `unit-tests/series-areasplinerange/polar` fails.

TorsteinHonsi (approved):
Thanks, nice job!

PR #17428, opened 2022-06-22
bugfix/17405-zoom-ordinal 

### LINKED ISSUE #17405: Zoom breaks with datagrouping disabled and boost in some data sets
#### Expected behaviour
Zoom should work fine

#### Actual behaviour
I have observed a few different behaviours:
- Works fine for some data sets
- For me, it just hangs the page in both chrome and firefox
- For my colleague, it works in chrome but adds some spurious line corrupting the chart in firefox

https://jsfiddle.net/4z5caq9p/ - This uses a specific data set that is generated via the following python code:

```python
import numpy as np
x0 = np.arange(1650500000, 1650590000, 1)
y0 = np.sin(x0/1000/2/np.pi)
x1 = np.arange(1650500000, 1650590000, 10)
y1 = np.sin(x1/1000/2/np.pi + np.pi/8)
```

#### Product version
Highstock 10.1.0

#### Affected browser(s)
Chrome, Firefox

karolkolodziej:
Fixed #17405, zooming on boosted charts froze under certain conditions.

Word of comment about that issue: 

- Before each series/axis render we calculate if the ordinal logic is needed. We do that by checking the distance between points of the combined series.
- This is a very data-specific issue. Each of these two series doesn't need ordinal logic because each point is spaced equally. However, one series has much more points than the other.
- After zooming we might end up with two series where the first and/or last position might not be exactly equally spaced which will force the chart to use ordinal while in reality, it doesn't need that.
This behavior caused an infinite loop.

### DIFF
--- a/ts/Core/Axis/OrdinalAxis.ts
+++ b/ts/Core/Axis/OrdinalAxis.ts
@@ -947,7 +947,8 @@ namespace OrdinalAxis {
                 ordinalPositions = [] as Array<number>,
                 overscrollPointsRange = Number.MAX_VALUE,
                 useOrdinal = false,
-                adjustOrdinalExtremesPoints = false;
+                adjustOrdinalExtremesPoints = false,
+                isBoosted = false;
 
             // Apply the ordinal logic
             if (isOrdinal || hasBreaks) { // #4167 YAxis is never ordinal ?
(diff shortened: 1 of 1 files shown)


karolkolodziej:
[Demo with recent changes](https://jsfiddle.net/BlackLabel/Lne2g38v/)

TorsteinHonsi (approved):
Thanks!

PR #17429, opened 2022-06-22
bugfix/17305-tunnel-height-localstorage

### LINKED ISSUE #17305: Stock Tools - issue with saving height of parallel channel to localStorage
#### Expected behaviour

The height of the parallel channel should be overwritten in localStorage if it is changed.

#### Actual behaviour

The height of the parallel channel is saved in localStorage only if the parallel channel is created from scratch. If the user modifies the height and saves it again, the new height is not saved.
However, updating and saving other control points (e.g. position) works fine.

parallelChanelBug

https://jsfiddle.net/BlackLabel/8zgva10f/

1. Add a parallel channel to the chart.
2. Save the chart in local storage using the stock button.
3. Refresh the page to get a blank chart and click the button "Restore annotation".
4. Select the parallel channel and change its height using the control button (middle one).
5. Save the chart again.
6. Refresh the page and restore again - the height of the parallel channel is not being saved, when other positions are saved correctly.

#### Product version
Highcharts Stock v10.1.0

hubertkozik:
Fixed #17305, the height of the Tunnel annotation wasn't overwritten in `localStorage`.

### DIFF
--- a/ts/Extensions/Annotations/Types/Tunnel.ts
+++ b/ts/Extensions/Annotations/Types/Tunnel.ts
@@ -171,6 +171,7 @@ class Tunnel extends CrookedLine {
 
         this.options.typeOptions.height = (this.points[3].y as any) -
             (this.points[0].y as any);
+        this.userOptions.typeOptions.height = this.options.typeOptions.height;
     }
 
 }


TorsteinHonsi (approved):
Thanks!

PR #17432, opened 2022-06-23
Migration of data module

bre1470:
- Removed internal namespace.
- Converted let to const.
- Moved class-independent functions.
- Made data module ESM compatible.


TorsteinHonsi (approved):
Thanks!

PR #17433, opened 2022-06-23
Fixed typing of Axis.translate.

bre1470:
Fixing `undefined` return in `Axis.translate` no code lines like to deal with.


bre1470:
As I stated there is not a single line dealing with `undefined` to determine if the axis is empty. For future cases we should implement some `isEmpty` if really needed. And `0` means no translation, so I would argue, that is the sensible result for the way our code is using translate. Otherwise we will bloat the code with `|| 0` for something that is not in use.

TorsteinHonsi (approved):

PR #17436, opened 2022-06-24
bugfix/17167-legend-title-no-data

### LINKED ISSUE #17167: Legend title is visible when there is no data in chart
#### Expected behaviour
Legend title should not be show when chart has no data
Behavior should be similar to Legend items/ series names.
#### Actual behaviour
Legend title is shown even when no data is present in chart

[http://jsfiddle.net/L7cnkg98/](https://urldefense.com/v3/__http:/jsfiddle.net/L7cnkg98/__;!!Obbck6kTJA!OT680xgTU3IRhrL-UAlI0SNq2mc1VHurzfGS9RIciAx0mbOLIzj_p_EVt44HJ1px$)
#### Product version

Latest version
#### Affected browser(s)
chrome,FF, edge

hubertkozik:
Fixed #17167, legend title didn't hide when chart data was empty.

### DIFF
--- a/samples/unit-tests/legend/legend-general/demo.js
+++ b/samples/unit-tests/legend/legend-general/demo.js
@@ -141,8 +141,11 @@ QUnit.test('Legend redraws', function (assert) {
         }
     });
 
-    assert.strictEqual(chart.legend.group.attr('visibility'), 'hidden',
-        'Legend title is not visible (#17167)');
+    assert.strictEqual(
+        chart.legend.group.attr('visibility'),
+        'hidden',
+        'Legend title should not be visible (#17167)'
+    );
 });
 
 QUnit.test(
--- a/samples/unit-tests/legend/legend-general/demo.js
+++ b/samples/unit-tests/legend/legend-general/demo.js
@@ -130,6 +130,18 @@ QUnit.test('Legend redraws', function (assert) {
     chart.series[0].hide(true);
 
     assert.strictEqual(visible, false, 'Legend item text has changed (#2165)');
+
+    chart.series[0].remove();
+
+    chart.update({
+        legend: {
+            title: {
+                text: 'Legend'
+            }
+        }
+    });
+
+    assert.strictEqual(chart.legend.group.visibility === 'hidden', true, 'Legend title is not visible (#17167)');
 });
 
 QUnit.test(
--- a/samples/unit-tests/legend/legend-general/demo.js
+++ b/samples/unit-tests/legend/legend-general/demo.js
@@ -131,7 +131,7 @@ QUnit.test('Legend redraws', function (assert) {
 
     assert.strictEqual(visible, false, 'Legend item text has changed (#2165)');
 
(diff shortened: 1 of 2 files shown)


samples/unit-tests/legend/legend-general/demo.js
```diff
@@ -130,6 +130,18 @@ QUnit.test('Legend redraws', function (assert) {
     chart.series[0].hide(true);
 
     assert.strictEqual(visible, false, 'Legend item text has changed (#2165)');
+
+    chart.series[0].remove();
+
+    chart.update({
```
raf18seb, line 136: You redraw the chart twice here. Instead, you can fire `series.remove()` method with a `redraw` flag set to `false` to update only options, and the chart will be redrawn only once during the `chart.update()` method.
https://api.highcharts.com/class-reference/Highcharts.Series#remove (see a `redraw` argument)

samples/unit-tests/legend/legend-general/demo.js
```diff
@@ -130,6 +130,18 @@ QUnit.test('Legend redraws', function (assert) {
...
+        legend: {
+            title: {
+                text: 'Legend'
+            }
+        }
+    });
+
+    assert.strictEqual(chart.legend.group.visibility === 'hidden', true, 'Legend title is not visible (#17167)');
```
raf18seb, line 144: Please do not exceed 80 characters line length.

raf18seb, line 144: You have 2 checks here: `visibility === 'hidden'` and `true === true` (`strictEqual`);
Try to find a way how to get rid of `===` to keep one check only (`strictEqual`).

raf18seb, line 144: Instead of pulling the property value from an object, it's safer to use the `attr` method:
`chart.legend.group.attr('visibility')`

raf18seb (changes requested):
Thanks! I have a few comments, please check below

samples/unit-tests/legend/legend-general/demo.js
```diff
@@ -130,6 +130,19 @@ QUnit.test('Legend redraws', function (assert) {
...
+    chart.update({
+        legend: {
+            title: {
+                text: 'Legend'
+            }
+        }
+    });
+
+    assert.strictEqual(chart.legend.group.attr('visibility'), 'hidden',
+        'Legend title is not visible (#17167)');
```
raf18seb, line 145: Instead of this:
```js
someFunction(first, second,
    third, fourth);
```

It's better to read and looks much better this way:
```js
someFunction(
    first,
    second,
    third,
    fourth
);
```

(see similar assertions above)

raf18seb (changes requested):
Looks better! Another 2 things and we're all set (see comments below)

raf18seb (approved):
Thanks!!

TorsteinHonsi (approved):
Thanks!

PR #17441, opened 2022-06-27
Fixed #17440, text path was lost on label text update

### LINKED ISSUE #17440: Text path setting doesn't stick after updating text
#### Expected behaviour
The text path to be preserved

#### Actual behaviour
Text path is lost and the text is in the position it would be without the path

https://jsfiddle.net/highcharts/sxme14fq/1/

#### Product version
10.1.0

TorsteinHonsi:
Fixed #17440, text path was lost on label text update. Refactored text paths, so they now support `textOutline` too.


pawelfus (approved):
Looks good, thanks!

TorsteinHonsi:
@pawelfus This PR mutated into a whole refactoring of text paths (now working on the AST instead of the DOM), so please review again.

ts/Series/Sunburst/SunburstPoint.ts
```diff
@@ -108,27 +108,38 @@ class SunburstPoint extends TreemapPoint {
             moreThanHalf = true;
         }
 
-        if (this.dataLabelPath) {
-            this.dataLabelPath = this.dataLabelPath.destroy();
+        // First time
+        if (!this.dataLabelPath) {
+            // Destroy the path with the label
+            wrap(label, 'destroy', (proceed): undefined => {
```
pawelfus, line 114: Instead of defining this wrap on every series that supports `textPath` through a custom shape, could we optimize it? It's easy to forget to add this wrap when adding the textPath's support to a new series.

For example, all paths added to `defs` could be by default removed in `label.destroy()`, and all others could be kept (e.g. when the shape is just a point's marker, like in the Networkgraph series).

pawelfus (commented):
Just one comment to discuss, all other things look good.

Tested also performance (using [this demo](http://utils.highcharts.local/samples/#view/highcharts/plotoptions/series-datalabels-rotation-crop-visual)). I didn't notice any performance drop because of the new event in `modifyTree`

ts/Series/Sunburst/SunburstPoint.ts
TorsteinHonsi: Yes I was in doubt here. The previous behaviour of adding the text path to the label itself in order to be destroyed with it, was a great idea. But doesn't work with the new model. Your proposal is good. Or maybe even safer - use a class name to identify those paths that are tied to a label one-to-one? Or a property in the `SVGElement.textPath` interface. It could simply be `textPath.unique = true`, which would instruct the destructor to flush it down with the label?

PR #17448, opened 2022-06-28
featrue/17285-nodeOpacity 

### LINKED ISSUE #17285: series.sankey.states.inactive.nodeOpacity
# Description of the feature
We can set opacity for the links between nodes in the sankey diagram in inactive mode by using series.sankey.states.inactive.linkOpacity, however, we can not set  opacity for the nodes  in the sankey diagram in inactive mode.
So we need series.sankey.states.inactive.nodeOpacity similarly.

Highcharts Sankey
https://api.highcharts.com/highcharts/series.sankey.states.inactive.linkOpacity

image
https://echarts.apache.org/examples/zh/editor.html?c=sankey-energy

----

karolkolodziej:
Fixed #17285, made the `opacity` property work for different states in the `sankey` series.

### DIFF
--- a/ts/Series/Sankey/SankeySeries.ts
+++ b/ts/Series/Sankey/SankeySeries.ts
@@ -268,7 +268,7 @@ class SankeySeries extends ColumnSeries {
          *
          * @private
          */
-        nodeOpacity: 1,
+        opacity: 1,
 
         /**
          * The minimal width for a line of a sankey. By default,
@@ -319,7 +319,7 @@ class SankeySeries extends ColumnSeries {
                 /**
                  * Opacity for the nodes in the sankey diagram in hover mode.
                  */
-                nodeOpacity: 1
+                opacity: 1
             },
             /**
              * The opposite state of a hover for a single point node/link.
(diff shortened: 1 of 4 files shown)


ts/Series/Sankey/SankeySeries.ts
```diff
@@ -322,10 +334,8 @@ class SankeySeries extends ColumnSeries {
                 linkOpacity: 0.1,
 
                 /**
-                 * Opacity of inactive markers.
-                 *
-                 * @type      {number}
-                 * @apioption plotOptions.series.states.inactive.opacity
+                 * Opacity for the links between nodes in the sankey diagram in
```
pawelfus, line 337: for the nodes*

pawelfus (changes requested):
Two minor things:
- a tiny change in the docs 
- `dataLabels` should inherit the opacity too

I'm not sure about the organization and the dependency wheel series types, up to @TorsteinHonsi

karolkolodziej:
@raf18seb visual comparison should be sufficient ;)

pawelfus (approved):
> Are the visual changes acceptable or should I adjust the default settings?

I'm not sure. I think the Organization chart should either set the inactive state to links too or remain unchanged. 

The rest looks good 👍

TorsteinHonsi (changes requested):
> I'm not sure about the organization and the dependency wheel series types, up to @TorsteinHonsi

I think those are good once we fix the problem with the opacity being applied twice.

TorsteinHonsi (approved):
Thanks!

PR #17452, opened 2022-06-29
Improved ZoomComponent.

bre1470:
- Removed AxisComposition of ZoomComponent
- Simplified forEach calls


TorsteinHonsi (approved):
Thanks!

PR #17454, opened 2022-06-30
Improved GridAxis composition.

bre1470:
- Reduced # of code lines


TorsteinHonsi (approved):
Thanks!

PR #17457, opened 2022-07-01
Migration of Annotation's Controllable class

bre1470:
- Splitted `ControllableMixin` into `Controllable` class, `ControllableLike`, `ControllableOptions`, `ControllableType` registry.


TorsteinHonsi (approved):
Thanks!

PR #17461, opened 2022-07-04
bugfix/17438-point-update-categories 

### LINKED ISSUE #17438: chart-update-moving-bar-incorrectly
#### Expected behaviour

using chart.update() with similar, but not Exactly the same, values,  the changed value (bar) is updated in it's correct place.

More detail and a workaround at https://stackoverflow.com/questions/72736226/chart-update-moving-bar-incorrectly, but the jsfiddle demonstrates the issue.

#### Actual behaviour

Value (bar) is re-ordered and layered on top of the first identical value in the series.

https://jsfiddle.net/wxpe6v2r/28/

#### Product version
9.3.3, but reproduced in 10.0

#### Affected browser(s)
Chrome

karolkolodziej:
Fixed #17438, updating points with the category axis gave the wrong x position.

### DIFF
--- a/samples/unit-tests/axis/category/demo.js
+++ b/samples/unit-tests/axis/category/demo.js
@@ -654,6 +654,27 @@ QUnit.test(
             'Zero,One,,Three',
             'Preserved names'
         );
+
+        chart.series[0].remove(false);
+        chart.xAxis[0].update({
+            type: 'category',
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks! Been trying to break this (experimenting with `uniqueNames`) but to no avail :)

PR #17466, opened 2022-07-05
Migration of ExportData.

bre1470:
- Made ExportData ESM-compatible
- Kept chart composition directly on Chart for backward-compatibility


TorsteinHonsi (approved):
Thanks!

PR #17469, opened 2022-07-06
Improved boost composition

bre1470:
- Reduced boost pollution on chart.

### DIFF
--- a/ts/Extensions/Boost/BoostSeries.ts
+++ b/ts/Extensions/Boost/BoostSeries.ts
@@ -100,6 +100,8 @@ interface BoostAlteredObject {
 interface BoostPointMockup {
     x: (false|number);
     clientX: number;
+    dist?: number;
+    distX?: number;
     plotX: number;
     plotY: number;
     i: number;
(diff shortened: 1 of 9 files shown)


ts/Extensions/Boost/BoostSeries.ts
```diff
@@ -604,180 +654,171 @@ function eachAsync(
...
+ * Performs the actual render if the renderer is
+ * attached to the series.
  * @private
- * @function Highcharts.Series#exitBoost
  */
-function seriesExitBoost(
-    this: BoostSeriesComposition
+function renderIfNotSeriesBoosting(
+    renderer: WGLRenderer,
+    series: Series,
```
TorsteinHonsi, line 808: Apparently, the `renderIfNotSeriesBoosting` and `getPoint` function needs the `series` to be passed as an argument, so they are not in the series scope. Would it make more sense to attach these functions to the `chart.boost` object instead? It seems a bit strange when `series.boost.getPoint()` is not bound to the series. Unless it is a matter of organising - that it makes sense to put `getPoint` in the BoostSeries object because it basically deals with series stuff?

bre1470: I put `getPoint` to `Series.boost` to undergo the need, that `Core/Pointer` has to directly import it from `BoostSeries`, which bloats the core. In theory both do not need to sit on any boost. `Chart.boost` could be an alternative. 👍

TorsteinHonsi: Okay, I leave it to you to see where you think it belongs.

bre1470: With `BoostChartAdditions` statements get longer, but I have found a way to make it simple to use.

TorsteinHonsi (approved):
Looks good!

PR #17470, opened 2022-07-06
Fixed #17400, announcer container not destroyed.

### LINKED ISSUE #17400: Accessibility. div.highcharts-announcer-container is not destroyed after Chart.destroy() method
When accessibility module is included. It generated bunch of `<div>`s like

```html
<div id="highcharts-screen-reader-region-before-1">...</div>
<div id="highcharts-screen-reader-region-before-2">...</div>
<div class="highcharts-announcer-container">...</div>
--- here is main chart code ---
<div id="highcharts-screen-reader-region-after-1">...</div>
etc...
```

After call of `Chart.destroy()` all `<div>`s are removed, but `.highcharts-announcer-container` still exists.
```html
<div class="highcharts-announcer-container">...</div>
```

oysteinmoseng:


TorsteinHonsi (approved):
Thanks!

PR #17471, opened 2022-07-06
Fixed #17401, console error with numeric point id.

### LINKED ISSUE #17401: Accessibility. point.id.indexOf is not a function
#### Expected behaviour
In accessibility module's medthod `getPointXDescription()` there is a chunk
```js
point.id.indexOf('highcharts-') < 0
```
but point.id might be a number, Number class doesn't have method `indexOf` This causes the error `TypeError: point.id.indexOf is not a function`

#### Actual behaviour
```js
String(point.id).indexOf('highcharts-') < 0
```
#### Product version
Highcharts, 10.1.0

#### Affected browser(s)
All

oysteinmoseng:
Fixed #17401, console error from accessibility module when using numeric point id.


dknight:
Thanks for the fix. Yes, the problem that type might be sloppy.

PR #17476, opened 2022-07-07
bugfix/17359-polar-stickyTracking 

### LINKED ISSUE #17359: Sticky tracking doesn't work for charts with polar=true config
#### Expected behaviour
Sticky tracking works exactly as expected when polar = false, but not when polar = true

#### Actual behaviour
Sticky tracking should work with chart having config polar = true

Reference: https://www.highcharts.com/forum/viewtopic.php?t=45216

 https://jsfiddle.net/patonomatic/a5bmw0se/6 (polar) vs https://jsfiddle.net/patonomatic/a5bmw0se/9 (non-polar)

#### Product version
Highcharts - 9.3.3

karolkolodziej:
Fixed #17359, `stickyTracking` was not respected correctly for polar charts.

### DIFF
--- a/ts/Core/Pointer.ts
+++ b/ts/Core/Pointer.ts
@@ -812,10 +812,8 @@ class Pointer {
         if (hoverPoint) {
             // When tooltip is shared, it displays more than one point
             if (shared && !hoverSeries.noSharedTooltip) {
-                searchSeries = series.filter(function (s): boolean {
-                    return eventArgs.filter ?
-                        eventArgs.filter(s) : filter(s) && !s.noSharedTooltip;
-                });
+                searchSeries = series.filter((s): boolean => s.stickyTracking &&
+                    (eventArgs.filter || filter)(s));
 
                 // Get all points with the same x value as the hoverPoint
                 searchSeries.forEach(function (s): any {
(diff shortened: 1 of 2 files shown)


ts/Core/Pointer.ts
```diff
@@ -795,8 +795,9 @@ class Pointer {
             [hoverSeries as any] :
             // Filter what series to look in.
             series.filter(function (s): boolean {
-                return eventArgs.filter ? eventArgs.filter(s) : filter(s) &&
-                    s.stickyTracking;
+                return eventArgs.filter ?
+                    eventArgs.filter(s) && s.stickyTracking :
+                    filter(s) && s.stickyTracking;
```
TorsteinHonsi, line 800: Looks like we can trim this expression a bit?
```ts
            series.filter((s): boolean => s.stickyTracking &&
                (eventArgs.filter || filter)(s));
```

TorsteinHonsi (changes requested):
Thanks! I think we can combine the condition a bit to become denser.

ts/Core/Pointer.ts
TorsteinHonsi: I think the requested change was applied to the wrong filter function... I requested a change to the filter function on line 798, but it was applied to the one on line 815, probably altering the functionality.

TorsteinHonsi (changes requested):

TorsteinHonsi (approved):
Thanks!

PR #17485, opened 2022-07-08
bugfix/17472-searchPointByAngle-polar

### LINKED ISSUE #17472: Spiderweb Shared Tooltip not rendering
#### Expected behaviour
Shared Tooltip should render multiple datapoints for the respective variable.

#### Actual behaviour
Tooltip does not render at all and the following error is thrown in the console.

highcharts_spiderweb

Shared tool tip is no longer working on the demo provided via the Highcharts website. Just hover over a datapoint and you will see the tool tip does not render.
https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/demo/polar-spider

#### Product version

Latest version from https://code.highcharts.com

#### Affected browser(s)
Chrome, Firefox, IE.

karolkolodziej:
Fixed #17472, shared tooltip did not work in polar charts.

- [x] test

### DIFF
--- a/ts/Series/PolarComposition.ts
+++ b/ts/Series/PolarComposition.ts
@@ -146,7 +146,7 @@ export declare class PolarSeriesComposition extends Series {
     animate(init?: boolean): void;
     searchPoint: (
         PolarSeriesComposition['kdByAngle'] extends true ?
-            'searchPointByAngle' :
+            typeof searchPointByAngle :
             Series['searchPoint']
     );
     xAxis: RadialAxis.AxisComposition;
--- a/ts/Series/PolarComposition.ts
+++ b/ts/Series/PolarComposition.ts
@@ -146,7 +146,7 @@ export declare class PolarSeriesComposition extends Series {
     animate(init?: boolean): void;
     searchPoint: (
         PolarSeriesComposition['kdByAngle'] extends true ?
-            PolarAdditions['searchPointByAngle'] :
+            'searchPointByAngle' :
             Series['searchPoint']
     );
     xAxis: RadialAxis.AxisComposition;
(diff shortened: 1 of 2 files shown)


ts/Series/PolarComposition.ts
```diff
@@ -146,7 +146,7 @@ export declare class PolarSeriesComposition extends Series {
     animate(init?: boolean): void;
     searchPoint: (
         PolarSeriesComposition['kdByAngle'] extends true ?
-            PolarAdditions['searchPointByAngle'] :
+            'searchPointByAngle' :
```
bre1470, line 149: `'searchPointByAngle'` is now a string literal type. Try `typeof searchPointByAngle`

bre1470 (changes requested):
One thing about typing needs be fixed.

Otherwise it looks good. The only drawback might be that is not accessible anymore, but it does not look like it is used outside of polar anyway.

bre1470 (approved):
Thanks! Looks good. 👍

TorsteinHonsi (approved):

GabrieleT0:
it seems that the problem is not solved, in this example it is still present: https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/demo/polar-spider

bre1470:
Thank you for the feedback. The changes are not yet released, see https://jsfiddle.net/bre1470/dk6ypz3w/ for the working example based on the current GitHub master.

GabrieleT0:
Thank you so much!

GabrieleT0:
There is a date for the release? I'm doing my bechelor's thesis and I'm using Hightchart, it's very important for me. Thank you.

bre1470:
We expect the next release to be ready this month.

PR #17490, opened 2022-07-11
bugfix/17484-organization-link-opacity 

### LINKED ISSUE #17484: Org chart link opacity is not applied
This issue is more relevant since node opacity is changing on hover starting from https://github.com/highcharts/highcharts/pull/17448 . I didn't have time to report it before the release, but I think org charts should behave like sankey and dependency wheel.

#### Expected behaviour
In sankey and dependency wheel series, the link opacity is reduced for remote links when hovering a node. This doesn't happen with the org chart.

#### Actual behaviour
The link opacity does not change on hover

https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/blog/data-type/
* Hover the top node (Data Type)
* Expected the links between the second and third level to dim.

#### Product version
10.2.0

karolkolodziej:
Fixed #17484, changed the link opacity on hover in organization series.

### DIFF
--- a/ts/Series/Organization/OrganizationSeries.ts
+++ b/ts/Series/Organization/OrganizationSeries.ts
@@ -516,7 +516,7 @@ class OrganizationSeries extends SankeySeries {
             values: (
                 OrganizationPointOptions &
                 OrganizationSeriesOptions
-            ) = ['borderRadius', 'linkColor', 'linkLineWidth']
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (approved):
Thanks! Yes, all these visual changes look correct.

PR #17491, opened 2022-07-11
bugfix/17481-zones-scrollablePlotArea 

### LINKED ISSUE #17481: Invisible issue for line after chart width with zones.
That problem occurs on charts with long data that have [`zones`](https://api.highcharts.com/highstock/plotOptions.series.zones).
I searched for existing issues, but couldn't find any similar cases. 😢

I also tested the [`cropThreshold`](https://api.highcharts.com/highcharts/series.line.cropThreshold) attribute, but there was no change.

<br />

#### Actual behaviour

**[With `zones` - JsFiddle](https://jsfiddle.net/qpnb25fj/4/)**

image

**[Without `zones` - JsFiddle](https://jsfiddle.net/qpnb25fj/5/)**

image

<br />

#### Expected behaviour

The line `<path>` must also be connected to the point that exists after scrollWidth. 

image

<br />

#### Product version
`Highcharts 10.1.0`

<br />

#### Affected browser(s)
`Chrome 103.0.5060`

karolkolodziej:
Fixed #17481, zones clip was incorrectly calculated and the line was partially missing when `scrollablePlotArea` was enabled.

### DIFF
--- a/samples/unit-tests/scrollable-plotarea/dynamics/demo.js
+++ b/samples/unit-tests/scrollable-plotarea/dynamics/demo.js
@@ -41,6 +41,25 @@ QUnit.test('Test dynamic behaviour of Scrollable PlotArea', function (assert) {
         chart.fixedRenderer.box.contains(chart.xAxis[0].axisGroup.element),
         'X-axis should be outside the scrollable plot area (#8862)'
     );
+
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #17492, opened 2022-07-11
Migration of NavigationBindings

bre1470:
Based & pending on PR: https://github.com/highcharts/highcharts/pull/17487
- Added composition
- Splitted options and helper functions (touchs StockTools)
- Optimized typing


TorsteinHonsi (approved):
Thanks!

PR #17502, opened 2022-07-13
bugfix/17486-bubbleZExtremes-remove 

### LINKED ISSUE #17486: chart.bubbleZExtremes not cleared on series.remove()
#### Expected behaviour
Added series to look the same as on init.

#### Actual behaviour
The series with 3 points looks OK on init. Then I'm adding a fourth point and the Z extremes are changed.
Then I'm removing the series and I'm adding the initial series with original 3 points - the bubbles are different sizes because `chart.bubbleZExtremes` wasn't cleared when the series with 4 points was removed.

https://jsfiddle.net/BlackLabel/8chdkqjw/

#### Product version
10.2.0

Workaround: clear `chart.bubbleZExtremes` before adding a new series (it will be recalculated): https://jsfiddle.net/BlackLabel/37qr9bau/

karolkolodziej:
Fixed #17486, `bubbleZExtremes` was not removed when the bubble series was removed.

### DIFF
--- a/ts/Series/Bubble/BubbleSeries.ts
+++ b/ts/Series/Bubble/BubbleSeries.ts
@@ -719,7 +719,7 @@ addEvent(BubbleSeries, 'updatedData', (e): void => {
     delete e.target.chart.bubbleZExtremes;
 });
 
-// After removing series, delete the chart-level Z extremes cache.
+// After removing series, delete the chart-level Z extremes cache, #17502.
 addEvent(BubbleSeries, 'remove', (e): void => {
     delete e.target.chart.bubbleZExtremes;
 });
--- a/ts/Series/Bubble/BubbleSeries.ts
+++ b/ts/Series/Bubble/BubbleSeries.ts
@@ -719,6 +719,11 @@ addEvent(BubbleSeries, 'updatedData', (e): void => {
     delete e.target.chart.bubbleZExtremes;
 });
 
+// After removing series, delete the chart-level Z extremes cache.
+addEvent(BubbleSeries, 'remove', (e): void => {
+    delete e.target.chart.bubbleZExtremes;
+});
+
 /* *
  *
  *  Axis ?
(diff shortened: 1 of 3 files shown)


ts/Series/Bubble/BubbleSeries.ts
```diff
@@ -719,6 +719,11 @@ addEvent(BubbleSeries, 'updatedData', (e): void => {
     delete e.target.chart.bubbleZExtremes;
 });
 
+// After removing series, delete the chart-level Z extremes cache.
+addEvent(BubbleSeries, 'remove', (e): void => {
+    delete e.target.chart.bubbleZExtremes;
```
raf18seb, line 724: Do we have to use `e`? Isn't `this` (series) enough?

ts/Series/Bubble/BubbleSeries.ts
```diff
@@ -719,6 +719,11 @@ addEvent(BubbleSeries, 'updatedData', (e): void => {
     delete e.target.chart.bubbleZExtremes;
 });
 
+// After removing series, delete the chart-level Z extremes cache.
```
raf18seb, line 722: Would be nice to add a ticket number for a reference

raf18seb (changes requested):
Are you planning to add functional tests? It's an ideal use case for them.
Except that, see the comments below

ts/Series/Bubble/BubbleSeries.ts
raf18seb: Maybe a simple `if` check? Probably `e` doesn't have enough TS definition and is defined `as any`. In this case, it will be safer to add `if` checks and actually use better series TS

raf18seb: Ok, I see it now. We are inside the arrow function so `this` context is not a `series`. Ok, let's keep it as it is, otherwise we would have to change it to a normal function.

PR #17505, opened 2022-07-14
Migration of Annotations' types

bre1470:
Based & pending on PR https://github.com/highcharts/highcharts/pull/17501

- Improved class structures
- Improved typing
- Removed redundant constructors

PR #17507, opened 2022-07-14
Migration of StockTools

bre1470:
Based & pending on PR https://github.com/highcharts/highcharts/pull/17505

- Added StockTools compositions
- Splitted classes / helpers
- Improved typing
- Removed internal namespace


bre1470:
Review of Stock Tools only: https://github.com/highcharts/highcharts/compare/ts/annotations-types...ts/stocktools

sebastianbochan:
I have only the problem with TS
- a lot of castings to "any"
- `StockToolsGuiDefinitionsButtonOptions` and `StockToolsGuiDefinitionsButtonsOptions` are really similar in names (so its confusing a bit). I'm also thinking about rearranging types and using something like `Array<StockToolsGuiDefinitionsButtonOptions>` or considering a generic type. However its only a kind of a thing for consideration.

bre1470:
Thanks for the review! I agree with your comment points. StockTools needs more type improvements.

Focus of this PR though is getting it ESM-compatible. Maybe you can take over after merge.

sebastianbochan (approved):

TorsteinHonsi (approved):
Thanks!

PR #17534, opened 2022-07-22
feature/14928-zoom-markings

hihajdus:
Added new option [zooming.markings](https://api.highcharts.com/highcharts/chart.zooming.markings), markers allowing to select the entire zoom area (horizontal or vertical). See #14928.


hihajdus:
To do:

- [ ] unit tests
- [ ] demo API

TorsteinHonsi:
Thanks! 

I'm wondering - do we actually need the visual markers? We already have the selection marker, and I can't see that it really adds value to highlight the corners. I see @KacperMadej 's comment [here](https://github.com/highcharts/highcharts/issues/14928#issuecomment-762358027), but I am not convinced. Any insight?

hihajdus:
Thanks @TorsteinHonsi for the note.
Additional graphical options were as a feature request. 
We can leave the basic styles as non-configurable and keep markings.enabled and markings.length in the API. The markings functionality seems to make this more intuitive with the visual showing of corners.

TorsteinHonsi:
As far as I can see, this PR actually adds two concepts:
* When zooming xy, the end user can optionally switch betseen xy, x and y.
* Markers appear around the selection area.

But in this PR those concepts are entangled, so that `zooming.markings` turns on both. But I don't understand what the term `marking` has to do with the dynamic switching of zoom mode, except that the markers help visualize what happens.

Next, this is what Kacper wrote in the isse:
> A partial border (frame) that will match the size limit in length on corners (2D zoom) or sides (1D zoom).

I think that the last part of the sentence is forgotten in this PR. The markings are always at the corners, thereby failing to be a visual help for switching between 2D and 1D zoom.

@KacperMadej  any thoughts?

KacperMadej:
1. Zooming could have a new type - `'dynamic'` or `'auto'` to would allow for dynamic change of the zoom type while making the selection in the chart.

A new option like `mode: 'dynamic' | 'fixed'` - but `'dynamic'` would override `type` so doesn't look that good IMO.

This would allow for:
- control over the zooming behavior - through `type:'dynamic'|'x'|'y'|'xy'`,
- `markings.enabled` would control visual aid that would be optional - not required for the behavior,
- `markings.length` affects the behavior so it might be better to change the option name and move it to `zooming` (maybe `threshold` ?) - that new option could be also used for any other zoom type instead of the currently hardcoded `10px`.

https://github.com/highcharts/highcharts/blob/cd9d2532a666ca028cd1ed815d58142d0fd6c73f/ts/Core/Pointer.ts#L316-L322


2. Markings on the sides should indicate where to stay to keep the current zoom type. For example, if you make the selection so the zoom type changes into `'y'` then the markings should be horizontal lines showing the area where, while dragging, you will stay in that zoom type. This feature is not implemented yet, and this is the idea that was mentioned by "A partial border (frame) that will match the size limit in [...] sides (1D zoom)".
image

hihajdus:
Thanks, @Kacper for explaintation,
Now I understand how the dynamic zoom functionality should be built.
As for the name of the new type, `auto` suits me best, and I have other name suggestion `d-z`.

For now, I will look into markings on the sides and see what adding a new type to the zooming option entails.

hihajdus:
Hi @TorsteinHonsi, I have questions for this dynamic zoom feature:
1. When adding a new type to the `zooming.type`, we have to refer to many places in the code. Would adding the dynamic zooming behavior as new `zooming.dynamic` option will be a better solution?

2. At this point I have removed most of the marking options, maybe we will wait for user feedback if there will be more interest in using dynamic zooming mode to add them in the future?

3. The length of the zooming selection marks remains, I called it threshold, but this one seems unintuitive.

TorsteinHonsi:
> Would adding the dynamic zooming behavior as new `zooming.dynamic` option will be a better solution?

That's the way it is currently implemented in the PR, right? But it doesn't work now, so it is hard for me to tell. The sample called `zooming-markings` currently uses the `markings` API that we abandoned. Could you update this or create a new demo for dynamic zooming?

> At this point I have removed most of the marking options, maybe we will wait for user feedback if there will be more interest in using dynamic zooming mode to add them in the future?

Yes, good idea. If we can keep a low footprint for this, it's a good thing. Maybe later we can get away with creating a demo/plugin/snippet on how to add markings.

> The length of the zooming selection marks remains, I called it threshold, but this one seems unintuitive.

Again, hard to say because I can't get the demo to work. Threshold is probably a good name, but maybe extended by something like `dragThreshold`. Because it means the minimum number of pixels the end user must drag before it counts as a selection.

TorsteinHonsi:
I'm putting this PR on hold for now. It doesn't seem to be a very frequently requested feature, and I am concerned about bloating the core.

pawellysy:
Close as stale

PR #17543, opened 2022-07-28
feature/17413-datalabel-contrast-color

### LINKED ISSUE #17413: dataLabel color should be able to apply contrast color when it's in the bar chart
# Description of the feature

According to the documentation for `series.bar.dataLabels.color` : https://api.highcharts.com/highcharts/series.bar.dataLabels
>The text color for the data labels. Defaults to undefined. For certain series types, like column or map, the data labels can be drawn inside the points. In this case the data label will be drawn with maximum contrast by default. Additionally, it will be given a text-outline style with the opposite color, to further increase the contrast. This can be overridden by setting the text-outline style to none in the dataLabels.style option.

It seems that the contrast color for data label will only exist in `type: column` 
https://www.highcharts.com/demo/column-comparison 

image

But I think it's also very useful in ordinary bar chart where data labels sometime gets pushed into the bar:
https://codesandbox.io/s/elated-jones-i5y7qo?file=/src/App.js:774-784

截圖 2022-06-21 14 07 04

Is it possible to apply this auto contrast color for dataLabel in bar chart? Or are there any workarounds that you'd suggest me to try it out? 

Highcharts

https://codesandbox.io/s/elated-jones-i5y7qo?file=/src/App.js:774-784

----

hubertkozik:
Fixed #17413, data labels contrast color omitted the chart background.

### DIFF
--- a/samples/unit-tests/datalabels/contrast/demo.js
+++ b/samples/unit-tests/datalabels/contrast/demo.js
@@ -1,7 +1,7 @@
 QUnit.test(
     "#6487: Column's data label with contrast after justification.",
     function (assert) {
-        var chart = Highcharts.chart('container', {
+        const chart = Highcharts.chart('container', {
                 chart: {
                     width: 600,
                     height: 400,
@@ -66,11 +66,78 @@ QUnit.test(
             `After updating from contrast color,
             label should have new color (#12500)`
         );
+
+        chart.series[1].remove(false);
+        chart.update({
+            chart: {
+                type: 'bar',
+                backgroundColor: '#000000'
+            },
+            series: [{
+                data: [100],
+                dataLabels: {
+                    enabled: true,
+                    format: 'Some long long long long long long long long text'
+                }
+            }]
+        }, false);
+        chart.yAxis[0].setExtremes(null, 1000, false, false);
+        chart.redraw();
+
+        const points = chart.series[0].points;
+
+        assert.strictEqual(
+            Highcharts.color(
+                points[0].dataLabel.element.childNodes[0].style.color
+            ).get(),
+            Highcharts.color(
+                'rgb(255,255,255)'
+            ).get(),
+            `If background of chart is dark dataLabels outside the bar chart
+            should get the contrast (white) color (#17413).`
+        );
+
+        chart.update({
+            chart: {
+                plotBackgroundColor: '#ffffff'
+            }
+        });
+
+        assert.strictEqual(
+            Highcharts.color(
+                points[0].dataLabel.element.childNodes[0].style.color
+            ).get(),
+            Highcharts.color(
+                'rgb(0,0,0)'
+            ).get(),
+            `If background of chart is dark, but plot background color is light
+            dataLabels outside the bar chart should get the contrast (black) 
+            color (#17413).`
+        );
+
+        chart.update({
+            chart: {
+                backgroundColor: '#ffffff',
+                plotBackgroundColor: '#000000'
+            }
+        });
+
+        assert.strictEqual(
+            Highcharts.color(
+                points[0].dataLabel.element.childNodes[0].style.color
+            ).get(),
+            Highcharts.color(
+                'rgb(255,255,255)'
+            ).get(),
+            `If background of chart is light, but plot background color is dark
+            dataLabels outside the bar chart should get the contrast (black)
+            color (#17413).`
+        );
     }
 );
 
 QUnit.test('Pie dataLabels and contrast', function (assert) {
-    var chart = Highcharts.chart('container', {
+    const chart = Highcharts.chart('container', {
             plotOptions: {
                 series: {
                     dataLabels: {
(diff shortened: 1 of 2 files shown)


ts/Core/Series/DataLabel.ts
```diff
@@ -608,14 +608,36 @@ namespace DataLabel {
                                 point.contrastColor = renderer.getContrast(
                                     (point.color || series.color) as any
                                 );
+
+                                const background = chart.chartBackground,
+                                    plotBackground = chart.plotBackground;
+
+                                let elseColor = String(Palette.neutralColor100);
```
raf18seb, line 615: I think using the JS `String()` function is not necessary here because we're dealing with a TS error.

Better (but not good) solution would be to cast it ti String like this:
```ts
let elseColor = Palette.neutralColor100 as String;
```

The best solution would be to tell TypeScript that this is actually a `string` type variable since the very beginning of its declaration - TS definitions are fine now and we don't use any JS code.

samples/unit-tests/datalabels/contrast/demo.js
```diff
@@ -98,3 +98,68 @@ QUnit.test('Pie dataLabels and contrast', function (assert) {
         'DataLabels outside the pie chart should not get contrast color (#11140).'
     );
 });
+
+QUnit.test('Bar dataLabels and contrast', function (assert) {
+    const chart = Highcharts.chart('container', {
+            chart: {
+                type: "bar",
```
raf18seb, line 105: Please use single quotes when writing JavaScript code in Highcharts

samples/unit-tests/datalabels/contrast/demo.js
```diff
@@ -98,3 +98,68 @@ QUnit.test('Pie dataLabels and contrast', function (assert) {
...
+                backgroundColor: '#000000'
+            },
+            series: [{
+                data: [100, 1000],
+                dataLabels: {
+                    enabled: true,
+                    format: "Some long long long long long long long long text"
```
raf18seb, line 112: Double quotes

raf18seb (changes requested):
Nice!
Except the comments below I have one question - can we reuse one of the existing tests/charts? Creating more and more charts makes the test last longer and longer. Creating new charts (`Highcharts.chart`) for each test drastically affects performance when dealing with thousands of tests.

ts/Core/Series/DataLabel.ts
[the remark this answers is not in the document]
hubertkozik: I didn't want to use the word 'contrast' due to the same property name in point (in a few lines higher) but after consideration, it is also a good name.

hubertkozik:
I have moved the logic of the test to the existing chart instance (higher) with some necessary updates.

samples/unit-tests/datalabels/contrast/demo.js
```diff
@@ -66,11 +66,78 @@ QUnit.test(
...
+            series: [{
+                data: [100],
+                dataLabels: {
+                    enabled: true,
+                    format: 'Some long long long long long long long long text'
+                }
+            }]
+        }, false);
+        chart.yAxis[0].setExtremes(null, 1000, false, false);
+        chart.redraw();
```
raf18seb, line 85: Few notes here:
- you don't need to fire `chart.redraw();` - you can set the `redraw` flag to `true` like this: `chart.yAxis[0].setExtremes(null, 1000, true, false);`
- the `redraw` flag is `true` by default. Do you need to set `animation` flag to `false`? It could be simple `chart.yAxis[0].setExtremes(null, 1000);`
- do we need to remove the second series? I know, it looks better, but it's additional unnecessary code and with 2 series the test should work exactly the same. I'm good both (with `series.remove()` or without), it's your call to decide

raf18seb (changes requested):
One more thing, see a comment below

samples/unit-tests/datalabels/contrast/demo.js
raf18seb: And sorry, I should have added 3 separate comments...

hubertkozik: I would like to remove that 2nd series, to keep the cohesion of the test. Additional line with `chart.redraw();` deleted :)

raf18seb (approved):
Thanks!

ts/Core/Series/DataLabel.ts
```diff
@@ -608,14 +608,37 @@ namespace DataLabel {
...
+                                    );
+                                } else if (
+                                    background &&
+                                    background.attr('fill') !== 'none'
+                                ) {
+                                    contrastColor = renderer.getContrast(
+                                        background.attr('fill') as any
+                                    );
+                                }
+
```
TorsteinHonsi, line 633: I think it would be better to read the colors directly from the options: 

```ts
                                const contrastColor = renderer.getContrast(pick(
                                    chart.options.chart.plotBackgroundColor,
                                    chart.options.chart.backgroundColor,
                                    Palette.neutralColor10
                                ));
```

I get some type issues with this, but those are the same as you currently have with `.attr('fill')`. The problem is that the colors can be other types than strings. Maybe we should just type-check the options because gradients are hard to handle anyway.

Also, the chart and plot background is the same for all points, so we should move this calculation out of the `forEach` loop.

TorsteinHonsi (changes requested):
Thanks! Some room for optimizing code-size and performance here I think.

TorsteinHonsi (approved):

PR #17550, opened 2022-08-01
Fixed AST policy conflict with multiple ESM programs in Chrome.

bre1470:


bre1470:
It is impossible to solve this in Chrome-based browsers without exposing the policy.

PR #17553, opened 2022-08-02
Fixed boost refresh after empty series array.

bre1470:
- Fixed boost refresh when no series is present.
- Fixed a double wrap of BarSeries.translate inherited from ColumnSeries.


TorsteinHonsi (approved):
Thanks!

PR #17558, opened 2022-08-03
Fixed #17547, no forced boost with zoom.

### LINKED ISSUE #17547: Unable to exit boost when series contains null points.
#### Expected behaviour
When there are fewer points than `boostThreshold` chart should exit the boost module.

#### Actual behaviour
When one series contains null points, and after zooming and having fewer visible points than the `boostThreshold` the chart is still boosted.
When the null series is not visible on the chart the issue is not visible.

https://jsfiddle.net/BlackLabel/no0txe2c/

#### Product version

#### Affected browser(s)
All

bre1470:
Fixed #17547; When the `chart.zoomType` option is set, boosted rendering will not be forced as long as threshold is not reached.

### DIFF
--- a/ts/Extensions/Boost/Boost.ts
+++ b/ts/Extensions/Boost/Boost.ts
@@ -366,7 +366,8 @@ export default Boost;
 
 /**
  * If set to true, the whole chart will be boosted if one of the series
- * crosses its threshold, and all the series can be boosted.
+ * crosses its threshold, and all the series can be boosted. If the
+ * `chart.zoomType` option is set, it will default to `false`.
  *
  * @type      {boolean}
  * @default   true
--- a/ts/Extensions/Boost/Boost.ts
+++ b/ts/Extensions/Boost/Boost.ts
@@ -183,17 +183,12 @@ export default Boost;
  */
 
 /**
- * Set the series threshold for when the boost should kick in globally.
+ * The chart will be boosted, if one of the series crosses its threshold and all
+ * the series in the chart can be boosted.
  *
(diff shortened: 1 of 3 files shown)


ts/Extensions/Boost/Boost.ts
```diff
@@ -366,7 +366,8 @@ export default Boost;
 
 /**
  * If set to true, the whole chart will be boosted if one of the series
- * crosses its threshold, and all the series can be boosted.
+ * crosses its threshold, and all the series can be boosted. If the
+ * `chart.zoomType` option is set, it will default to `false`.
```
TorsteinHonsi, line 370: The doclet says `@default true`. If the description is right, this means the default should be `undefined`, which means an intelligent guess. The start of the description, "If set to true..." is also confusing, as it indicates that the default is not `true`.

TorsteinHonsi (changes requested):
In addition to the inline changes, see [Writing content for the changelog](https://github.com/highcharts/highcharts/tree/master/changelog#writing-content-for-the-changelog). Specifically, points 5, 6 and 7.

ts/Extensions/Boost/Boost.ts
bre1470: As the whole detection changed, I would argue that the default value is now still legit.

bre1470:
Fixing the root issue (boost checks too early) would come a rewrite near. So I instead changed multiple conditions so that is more often checking extremes to detect range/zoom change.

TorsteinHonsi (approved):

PR #17573, opened 2022-08-05
bugfix/17569-linlogswitch-demo-bug

### LINKED ISSUE #17569: custom stock tools button not deselecting after click
### Expected behavior
clicking the custom stock tool button should act like a toggle ( just like show annotation button)

### Actual behavior
clicking the custom stock tool button will activate log scale, then will require 2 clicks to activate linear scale. Will then require 2 clicks to activate log scale.

### Live demo with steps to reproduce

[jsFiddle demo](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/blog/lin-log-switch)

The 'Dynamically switch between linear and log axis scale using Stock Tools GUI' [demo ](https://www.highcharts.com/blog/tutorials/dynamically-switch-between-linear-and-log-axis-scale-using-stock-tools-gui/) requires two clicks in order to perform scale switch. I believe this is because the button isn't being deselected till the next click event. 

#### Product version
HighStock 10.2

### Affected browser(s)
I've only testing on Google Chrome

pawellysy:
Fixed #17569, custom button had to be clicked twice on official demo.

### DIFF
--- a/samples/highcharts/blog/lin-log-switch/demo.js
+++ b/samples/highcharts/blog/lin-log-switch/demo.js
@@ -67,7 +67,11 @@ Highcharts.getJSON('https://demo-live-data.highcharts.com/aapl-ohlcv.json', func
                         button.firstChild.style['background-image'] = `url(${iconURL})`;
 
                         // Deselect button after click.
-                        button.classList.remove('highcharts-active');
+                        Highcharts.fireEvent(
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (approved):
Thanks!

PR #17588, opened 2022-08-09
Fixed #17551; added Series.types and Series.registerType.

### LINKED ISSUE #17551: Type definitions for `PieSeries` not public
#### Expected behaviour
To be able to see and use these, a common use case is to call a method from a predefined series; in the docs there are many examples with this.

#### Actual behaviour
`seriesTypes`, `PieSeries` are not public

Am I missing something? It would be great to have these. 

Thanks for the info in advance!
----

bre1470:
Fixed #17551; Added `Highcharts.Series.types` and `Series.registerType(type, class)` for custom series.

### DIFF
--- a/test/typescript-dts/classes/Series.ts
+++ b/test/typescript-dts/classes/Series.ts
@@ -1,6 +1,7 @@
 import * as Highcharts from 'highcharts';
 
 test_Series();
+test_Series_registerType();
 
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (approved):
Thanks 🎉

PR #17590, opened 2022-08-10
bugfix/17574-incorrect-marker-demo

### LINKED ISSUE #17574: Incorrect demo for disabled markers in API-doc
#### Expected behaviour
[API-doc for markers](https://api.highcharts.com/highcharts/plotOptions.series.marker) includes two demo-links for the enabled-option:
[Disabled markers](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/plotoptions/series-marker-enabled/)
[Disabled in normal state but enabled on hover](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/plotoptions/series-marker-enabled-false/)
The disabled marker demo should have markers disabled at all times, also on hover.

#### Actual behaviour
The demos are identical. The disabled markers demo has markers enabled on hover.

Demo that works (markers disabled at all times): [https://jsfiddle.net/6509ucya/](https://jsfiddle.net/6509ucya/)

#### Product version
Highcharts v10.2.0

#### Affected browser(s)
All

kamil-musialowski:
Fixed #17574, markers in demo were still enabled.

### DIFF
--- a/samples/highcharts/plotoptions/series-marker-enabled/demo.js
+++ b/samples/highcharts/plotoptions/series-marker-enabled/demo.js
@@ -1,19 +1,19 @@
-Highcharts.chart("container", {
+Highcharts.chart('container', {
     xAxis: {
         categories: [
-            "Jan", "Feb", "Mar", "Apr", "May", "Jun",
-            "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"
+            'Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep',
+            'Oct', 'Nov', 'Dec'
         ]
     },
 
     plotOptions: {
         series: {
             marker: {
-                enabled: false
-            },
-            states: {
-                hover: {
-                    enabled: false
+                enabled: false,
+                states: {
+                    hover: {
+                        enabled: false
+                    }
                 }
             }
         }
@@ -21,8 +21,8 @@ Highcharts.chart("container", {
 
     series: [{
         data: [
-            29.9, 71.5, 106.4, 129.2, 144.0, 176.0,
-            135.6, 148.5, 216.4, 194.1, 95.6, 54.4
+            29.9, 71.5, 106.4, 129.2, 144.0, 176.0, 135.6, 148.5, 216.4, 194.1,
+            95.6, 54.4
         ]
     }]
 });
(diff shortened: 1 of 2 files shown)


samples/highcharts/plotoptions/series-marker-enabled/demo.js
```diff
@@ -1,17 +1,28 @@
...
+                }
             }
         }
     },
 
     series: [{
-        data: [29.9, 71.5, 106.4, 129.2, 144.0, 176.0, 135.6, 148.5, 216.4, 194.1, 95.6, 54.4]
+        data: [
+            29.9, 71.5, 106.4, 129.2, 144.0, 176.0,
+            135.6, 148.5, 216.4, 194.1, 95.6, 54.4
```
raf18seb, line 25: I'm not sure about this line break in the middle of array. It looks nice for our brains, but it's not a common rule.
Personally, as it's only a jsFiddle demo, I'd keep it in one line. Breaking into 12+2 lines doesn't make sense imho.

@TorsteinHonsi could you decide so we have one rule and stick to it in the future?

raf18seb (changes requested):

samples/highcharts/plotoptions/series-marker-enabled/demo.js
TorsteinHonsi: @raf18seb  I prefer breaking up the array at 80 characters because the alternatives are worse. There are basically three options.

* No break. This may cause horizontal scrolling in jsFiddle, code editors and in code previews.
```js
// bad
data: [29.9, 71.5, 106.4, 129.2, 144.0, 176.0, 135.6, 148.5, 216.4, 194.1, 95.6, 54.4]
```

* Break for all items. This will require too much vertical space, at least for single values
```js
// bad
data: [
  29.9,
  71.5,
  106.4,
  129.2,
  144.0,
  176.0,
  135.6,
  148.5,
  216.4,
  194.1,
  95.6,
  54.4
]
```

* Compromise, break at 80
```js
// good
data: [
   29.9, 71.5, 106.4, 129.2, 144.0, 176.0, 135.6, 148.5, 216.4,
   194.1, 95.6, 54.4
]
```

TorsteinHonsi (approved):

PR #17592, opened 2022-08-11
Migration of NetworkgraphSeries

bre1470:
- Removed internal namespace from Networkgraph.


ts/Series/Networkgraph/NetworkgraphSeriesOptions.d.ts
```diff
@@ -0,0 +1,94 @@
...
+}
+
+export interface NetworkgraphSeriesOptions
+    extends SeriesOptions, NodesComposition.SeriesCompositionOptions {
+
+    dataLabels?: NetworkgraphDataLabelsOptionsObject;
+    draggable?: boolean;
+    inactiveOtherPoints?: boolean;
+    layoutAlgorithm?: ReingoldFruchtermanLayout.Options;
+    link?: SVGAttributes;
```
TorsteinHonsi, line 83: I suspect this type is wrong because the `link` option contains `color` and `width`. If this is passed directly to an SVG element, the color will (most likely) not be applied. Most likely the color is read somewhere and a `stroke` is applied instead. If this is true, `color` can be removed from `SVGAttributes.d.ts` too.

TorsteinHonsi (changes requested):

TorsteinHonsi (approved):

PR #17596, opened 2022-08-12
Docs: Fixed API options and declarations.

### LINKED ISSUE #14029: SVGAttributes are missing dx and dy options
Docs: https://api.highcharts.com/class-reference/Highcharts.SVGAttributes

In `textPath` we also have `dx` and `dy` options too: https://api.highcharts.com/highcharts/series.line.dataLabels.textPath

Used for example in our official demo: https://www.highcharts.com/demo/dependency-wheel

bre1470:
- Fixed #14029, missing SVG attributes in options
- Fixed missing type information for options in DependencyWheel demo
- Fixed issues with duplicated type declarations

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,6 +11,7 @@
       "devDependencies": {
         "@babel/core": "^7.18.10",
         "@babel/preset-env": "^7.18.10",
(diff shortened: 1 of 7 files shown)


TorsteinHonsi (approved):
Thanks!

PR #17599, opened 2022-08-15
bugfix/15752-parallel-coordinates-arearange

### LINKED ISSUE #15752: The arearange series isn't working in parallel coordinates
#### Expected behaviour
Let's consider this arearange data:
```js
data: [
      [23, 25],
      [-1, -0.5],
      [1.5, 4],
      [1.5, 5],
      [4, 5]
]
```
The arearange series should be displayed like this:

image

#### Actual behaviour
With the above data, the arearange series is displayed incorrectly:
image

To display it correctly, we need to correct the data - calculate the `high` value based on the first yAxis scale:
```js
data: [
      [23, 25],
      [-1, 23.5],
      [1.5, 25],
      [1.5, 26],
      [4, 24]
]
```

Real data, not working: https://jsfiddle.net/BlackLabel/qobuj3tp/
Corrected data, but obviously it shouldn't work like that: https://jsfiddle.net/BlackLabel/kmgh8fad/

#### Product version
Highcharts 9.1.0

K009:
Fixed #15752, arearange didn’t work in parallel coordinates.

### DIFF
--- a/samples/unit-tests/chart/parallel-coordinates-updates/demo.js
+++ b/samples/unit-tests/chart/parallel-coordinates-updates/demo.js
@@ -144,11 +144,9 @@ QUnit.test('Update parallel coordinates plot', function (assert) {
         3,
         'After chart update parallelInfo.counter has correct value (#10081).'
     );
-});
 
-// Calculate yAxis extremes based on the range series points as well (#15752)
-QUnit.test('Take range series points into account while calculating yAxis extremes (#15752)', function (assert) {
-    const chart = Highcharts.chart('container', {
+    // Calculate yAxis extremes based on the range series points (#15752)
+    chart.update({
         chart: {
             parallelCoordinates: true
         },
@@ -167,49 +165,46 @@ QUnit.test('Take range series points into account while calculating yAxis extrem
     const firstAxisMin = chart.yAxis[0].dataMin,
         firstAxisMax = chart.yAxis[0].dataMax,
         thirdAxisMin = chart.yAxis[2].dataMin,
-        thirdAxisMax = chart.yAxis[2].dataMax;
+        thirdAxisMax = chart.yAxis[2].dataMax,
+        series = chart.series[0];
 
-    assert.deepEqual(23, firstAxisMin, '1st yAxis min should be 23.');
-    assert.deepEqual(25, firstAxisMax, '1st yAxis max should be 25.');
+    assert.strictEqual(
+        series.points[0].low,
+        firstAxisMin,
+        '1st yAxis min should be 23.'
+    );
+    assert.strictEqual(
+        series.points[0].high,
+        firstAxisMax,
+        '1st yAxis max should be 25.'
+    );
 
-    assert.deepEqual(1.5, thirdAxisMin, '3rd yAxis min should be 1.5.');
-    assert.deepEqual(4, thirdAxisMax, '3rd yAxis max should be 4.');
-});
+    assert.strictEqual(
+        series.points[2].low,
+        thirdAxisMin,
+        '3rd yAxis min should be 1.5.'
+    );
+    assert.strictEqual(
+        series.points[2].high,
+        thirdAxisMax,
+        '3rd yAxis max should be 4.'
+    );
 
-// Calculate plotHigh value based on each yAxis scale (#15752)
-QUnit.test('Calculate plotHigh value based on each yAxis scale (#15752)', function (assert) {
-    const chart = Highcharts.chart('container', {
-        chart: {
-            parallelCoordinates: true
-        },
-        series: [{
-            type: 'arearange',
-            data: [
-                [23, 25],
-                [-1, -0.5],
-                [1.5, 4],
-                [1.5, 5],
-                [4, 5]
-            ]
-        }]
-    });
-
-    const series = chart.series[0];
+    // Calculate plotHigh value based on each yAxis scale (#15752)
     series.points.forEach(function (
         point,
         i
     ) {
-        assert.deepEqual(
+        assert.strictEqual(
             chart.yAxis[i].translate(
-                series.dataModify ?
-                    series.dataModify.modifyValue(point.high) : point.high,
+                point.high,
                 0,
                 1,
                 0,
                 1
             ),
             point.plotHigh,
-            "PlotHigh value should be calculated based on the point's yAxis."
+            `PlotHigh value should be calculated based on the point's yAxis.`
         );
     });
 });
--- a/samples/unit-tests/chart/parallel-coordinates-updates/demo.js
+++ b/samples/unit-tests/chart/parallel-coordinates-updates/demo.js
@@ -147,11 +147,13 @@ QUnit.test('Update parallel coordinates plot', function (assert) {
 
     // Calculate yAxis extremes based on the range series points (#15752)
     chart.series[0].setData([]);
-    chart.series[1].setData([]);
+    chart.series[1].remove();
 
     chart.update({
+        chart: {
+            type: 'arearange'
+        },
         series: [{
-            type: 'arearange',
             data: [
                 [23, 25],
                 [-1, -0.5],
@@ -200,26 +202,7 @@ QUnit.test('Update parallel coordinates plot', function (assert) {
     });
 
     // Multiple series case (#15752)
-    chart.update({
-        series: [{
-            type: 'arearange',
-            data: [
-                [-20, -15],
-                [-10, -5],
-                [-10, -5]
-            ]
-        }, {
-            type: 'arearange',
-            data: [
-                [0, 1],
-                [0, 5],
-                [0, 8]
-            ]
-        }]
-    });
-
     chart.addSeries({
-        type: 'arearange',
         data: [
             [10, 15],
             [10, 15],
@@ -228,8 +211,7 @@ QUnit.test('Update parallel coordinates plot', function (assert) {
     });
 
     const firstSeries = chart.series[0],
-        secondSeries = chart.series[1],
-        thirdSeries = chart.series[2];
+        secondSeries = chart.series[1];
 
     firstAxisMin = chart.yAxis[0].dataMin;
     firstAxisMax = chart.yAxis[0].dataMax;
@@ -237,23 +219,23 @@ QUnit.test('Update parallel coordinates plot', function (assert) {
     thirdAxisMax = chart.yAxis[2].dataMax;
 
     assert.strictEqual(
-        firstSeries.points[0].low,
+        secondSeries.points[0].low,
         firstAxisMin,
         `The first yAxis' min should be equal to the first point's low value.`
     );
     assert.strictEqual(
-        thirdSeries.points[0].high,
+        firstSeries.points[0].high,
         firstAxisMax,
         `The first yAxis' max should be equal to the first point's high value.`
     );
 
     assert.strictEqual(
-        firstSeries.points[2].low,
+        secondSeries.points[2].low,
         thirdAxisMin,
         `The third yAxis' min should be equal to the third point's low value.`
     );
     assert.strictEqual(
-        secondSeries.points[2].high,
+        firstSeries.points[2].high,
         thirdAxisMax,
         `The third yAxis' max should be equal to the third point's high value.`
     );
--- a/samples/unit-tests/chart/parallel-coordinates-updates/demo.js
+++ b/samples/unit-tests/chart/parallel-coordinates-updates/demo.js
@@ -162,11 +162,11 @@ QUnit.test('Update parallel coordinates plot', function (assert) {
         }]
     });
 
-    const firstAxisMin = chart.yAxis[0].dataMin,
+    const points = chart.series[0].points;
+    let firstAxisMin = chart.yAxis[0].dataMin,
         firstAxisMax = chart.yAxis[0].dataMax,
         thirdAxisMin = chart.yAxis[2].dataMin,
-        thirdAxisMax = chart.yAxis[2].dataMax,
-        points = chart.series[0].points;
+        thirdAxisMax = chart.yAxis[2].dataMax;
 
     assert.strictEqual(
         points[0].low,
@@ -198,4 +198,63 @@ QUnit.test('Update parallel coordinates plot', function (assert) {
             `PlotHigh value should be calculated based on the point's yAxis.`
         );
     });
+
+    // Multiple series case (#15752)
+    chart.update({
+        series: [{
+            type: 'arearange',
+            data: [
+                [-20, -15],
+                [-10, -5],
+                [-10, -5]
+            ]
+        }, {
+            type: 'arearange',
+            data: [
+                [0, 1],
+                [0, 5],
+                [0, 8]
+            ]
+        }]
+    });
+
+    chart.addSeries({
+        type: 'arearange',
+        data: [
+            [10, 15],
+            [10, 15],
+            [-3, -1]
+        ]
+    });
+
+    const firstSeries = chart.series[0],
+        secondSeries = chart.series[1],
+        thirdSeries = chart.series[2];
+
+    firstAxisMin = chart.yAxis[0].dataMin;
+    firstAxisMax = chart.yAxis[0].dataMax;
+    thirdAxisMin = chart.yAxis[2].dataMin;
+    thirdAxisMax = chart.yAxis[2].dataMax;
+
+    assert.strictEqual(
+        firstSeries.points[0].low,
+        firstAxisMin,
+        `The first yAxis' min should be equal to the first point's low value.`
+    );
+    assert.strictEqual(
+        thirdSeries.points[0].high,
+        firstAxisMax,
+        `The first yAxis' max should be equal to the first point's high value.`
+    );
+
+    assert.strictEqual(
+        firstSeries.points[2].low,
+        thirdAxisMin,
+        `The third yAxis' min should be equal to the third point's low value.`
+    );
+    assert.strictEqual(
+        secondSeries.points[2].high,
+        thirdAxisMax,
+        `The third yAxis' max should be equal to the third point's high value.`
+    );
 });
--- a/samples/unit-tests/chart/parallel-coordinates-updates/demo.js
+++ b/samples/unit-tests/chart/parallel-coordinates-updates/demo.js
@@ -146,6 +146,9 @@ QUnit.test('Update parallel coordinates plot', function (assert) {
     );
 
     // Calculate yAxis extremes based on the range series points (#15752)
+    chart.series[0].setData([]);
+    chart.series[1].setData([]);
+
     chart.update({
         series: [{
             type: 'arearange',
--- a/samples/unit-tests/chart/parallel-coordinates-updates/demo.js
+++ b/samples/unit-tests/chart/parallel-coordinates-updates/demo.js
@@ -145,3 +145,71 @@ QUnit.test('Update parallel coordinates plot', function (assert) {
         'After chart update parallelInfo.counter has correct value (#10081).'
     );
 });
+
+// Calculate yAxis extremes based on the range series points as well (#15752)
+QUnit.test('Take range series points into account while calculating yAxis extremes (#15752)', function (assert) {
+    const chart = Highcharts.chart('container', {
+        chart: {
+            parallelCoordinates: true
+        },
+        series: [{
+            type: 'arearange',
+            data: [
+                [23, 25],
+                [-1, -0.5],
+                [1.5, 4],
+                [1.5, 5],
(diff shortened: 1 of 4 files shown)


ts/Series/AreaRange/AreaRangeSeries.ts
```diff
@@ -284,14 +286,28 @@ class AreaRangeSeries extends AreaSeries {
...
+                } else {
+                    point.plotHigh = yAxis.translate(
+                        series.dataModify ?
+                            series.dataModify.modifyValue(high) : high,
+                        0 as any,
+                        1 as any,
+                        0 as any,
+                        1 as any
+                    );
+                }
```
raf18seb, line 309: Repeated code. The only difference here is `yAxis` used, so I think you could simplify it to a simple variable:
```js
const yAxis = isParallelCor ? chart.yAxis[i] : series.yAxis;
```

ts/Series/AreaRange/AreaRangeSeries.ts
```diff
@@ -268,13 +268,15 @@ class AreaRangeSeries extends AreaSeries {
      */
     public translate(): void {
         const series = this,
-            yAxis = series.yAxis;
+            yAxis = series.yAxis,
+            chart = series.chart;
```
raf18seb, line 272: You're using `chart` variable in 2 places only.

If you'd like to keep your variable, then it's a good practice to use the smallest possible scope (block scope instead of function scope). (the same with `yAxis` variable - scope could be reduced)
If you don't have many characters in a line (less than 80), then you don't need to declare a `chart` variable at all.

Your call which one you decide :)

samples/unit-tests/chart/parallel-coordinates-updates/demo.js
```diff
@@ -145,3 +145,71 @@ QUnit.test('Update parallel coordinates plot', function (assert) {
...
+    const firstAxisMin = chart.yAxis[0].dataMin,
+        firstAxisMax = chart.yAxis[0].dataMax,
+        thirdAxisMin = chart.yAxis[2].dataMin,
+        thirdAxisMax = chart.yAxis[2].dataMax;
+
+    assert.deepEqual(23, firstAxisMin, '1st yAxis min should be 23.');
+    assert.deepEqual(25, firstAxisMax, '1st yAxis max should be 25.');
+
+    assert.deepEqual(1.5, thirdAxisMin, '3rd yAxis min should be 1.5.');
+    assert.deepEqual(4, thirdAxisMax, '3rd yAxis max should be 4.');
```
raf18seb, line 176: When comparing numbers, use `assert.strictEqual`.

Also, it's better to use universal values like:
```js
assert.strictEqual(
    series.points[0].low, // is it low? I don't remember but you'll know
    firstAxisMin,
    ....
);
```

samples/unit-tests/chart/parallel-coordinates-updates/demo.js
```diff
@@ -145,3 +145,71 @@ QUnit.test('Update parallel coordinates plot', function (assert) {
         'After chart update parallelInfo.counter has correct value (#10081).'
     );
 });
+
+// Calculate yAxis extremes based on the range series points as well (#15752)
+QUnit.test('Take range series points into account while calculating yAxis extremes (#15752)', function (assert) {
```
raf18seb, line 150: Please do not exceed 80 characters length

samples/unit-tests/chart/parallel-coordinates-updates/demo.js
```diff
@@ -145,3 +145,71 @@ QUnit.test('Update parallel coordinates plot', function (assert) {
...
+            chart.yAxis[i].translate(
+                series.dataModify ?
+                    series.dataModify.modifyValue(point.high) : point.high,
+                0,
+                1,
+                0,
+                1
+            ),
+            point.plotHigh,
+            "PlotHigh value should be calculated based on the point's yAxis."
```
raf18seb, line 212: Double quote

samples/unit-tests/chart/parallel-coordinates-updates/demo.js
```diff
@@ -145,3 +145,71 @@ QUnit.test('Update parallel coordinates plot', function (assert) {
...
+
+    const series = chart.series[0];
+    series.points.forEach(function (
+        point,
+        i
+    ) {
+        assert.deepEqual(
+            chart.yAxis[i].translate(
+                series.dataModify ?
+                    series.dataModify.modifyValue(point.high) : point.high,
```
raf18seb, line 205: When writing tests, you should simplify the code as much as you can.

In core, we don't know whether a user is using `dataModify` or not. But here, in test, we know that we're not using it, so you can get rid of it

samples/unit-tests/chart/parallel-coordinates-updates/demo.js
```diff
@@ -145,3 +145,71 @@ QUnit.test('Update parallel coordinates plot', function (assert) {
...
+            ]
+        }]
+    });
+
+    const series = chart.series[0];
+    series.points.forEach(function (
+        point,
+        i
+    ) {
+        assert.deepEqual(
```
raf18seb, line 202: `deepEqual` -> `strictEqual`

ts/Extensions/ParallelCoordinates.ts
```diff
@@ -691,8 +690,33 @@ namespace ParallelAxis {
...
             });
 
-            axis.dataMin = arrayMin(currentPoints);
-            axis.dataMax = arrayMax(currentPoints);
+            // Take into account range series points as well (#15752)
+            let min: number = NaN,
+                max: number = NaN;
+
+            currentPoints.forEach(function (point): void {
+                if (Array.isArray(point)) {
```
raf18seb, line 698: In Highcharts, we're using `Highcharts.isArray` (from Utilities) instead of `Array.isArray`

ts/Extensions/ParallelCoordinates.ts
```diff
@@ -691,8 +690,33 @@ namespace ParallelAxis {
...
+                    max = arrayMax(point);
+                    min = arrayMin(point);
+                } else {
+                    return;
+                }
+            });
+
+            if (
+                !Array.isArray(arrayMin(currentPoints)) ||
+                !Array.isArray(arrayMax(currentPoints))
```
raf18seb, line 708: Same as above (Array.isArray)

ts/Extensions/ParallelCoordinates.ts
```diff
@@ -691,8 +690,33 @@ namespace ParallelAxis {
...
+            let min: number = NaN,
+                max: number = NaN;
+
+            currentPoints.forEach(function (point): void {
+                if (Array.isArray(point)) {
+                    max = arrayMax(point);
+                    min = arrayMin(point);
+                } else {
+                    return;
+                }
```
raf18seb, line 703: Is this empty return necessary here? `forEach` returns `undefined` by definition. Is TS rude here again?

samples/unit-tests/chart/parallel-coordinates-updates/demo.js
```diff
@@ -145,3 +145,71 @@ QUnit.test('Update parallel coordinates plot', function (assert) {
...
+
+    assert.deepEqual(23, firstAxisMin, '1st yAxis min should be 23.');
+    assert.deepEqual(25, firstAxisMax, '1st yAxis max should be 25.');
+
+    assert.deepEqual(1.5, thirdAxisMin, '3rd yAxis min should be 1.5.');
+    assert.deepEqual(4, thirdAxisMax, '3rd yAxis max should be 4.');
+});
+
+// Calculate plotHigh value based on each yAxis scale (#15752)
+QUnit.test('Calculate plotHigh value based on each yAxis scale (#15752)', function (assert) {
```
raf18seb, line 180: 80

raf18seb (changes requested):
Thanks for the PR!
Except the comments below, I have one important note: when writing unit tests, please try to reuse existing tests (and use updates if necessary) instead of creating new test modules and new charts, especially if they have the same config.
It's very important for performance - we already have thousands of tests and it takes a lot of time at every commit to fire all of them.

- [ ] reuse existing tests instead of creating new ones

ts/Extensions/ParallelCoordinates.ts
K009: Not necessary 👍

samples/unit-tests/chart/parallel-coordinates-updates/demo.js
```diff
@@ -167,49 +165,46 @@ QUnit.test('Take range series points into account while calculating yAxis extrem
     const firstAxisMin = chart.yAxis[0].dataMin,
         firstAxisMax = chart.yAxis[0].dataMax,
         thirdAxisMin = chart.yAxis[2].dataMin,
-        thirdAxisMax = chart.yAxis[2].dataMax;
+        thirdAxisMax = chart.yAxis[2].dataMax,
+        series = chart.series[0];
```
raf18seb, line 169: Wouldn't it be better to declare it like this?
```js
const points = chart.series[0].points;
```

samples/unit-tests/chart/parallel-coordinates-updates/demo.js
```diff
@@ -167,49 +165,46 @@ QUnit.test('Take range series points into account while calculating yAxis extrem
...
-    assert.deepEqual(23, firstAxisMin, '1st yAxis min should be 23.');
-    assert.deepEqual(25, firstAxisMax, '1st yAxis max should be 25.');
-
-    assert.deepEqual(1.5, thirdAxisMin, '3rd yAxis min should be 1.5.');
-    assert.deepEqual(4, thirdAxisMax, '3rd yAxis max should be 4.');
-});
+    assert.strictEqual(
+        series.points[0].low,
+        firstAxisMin,
+        '1st yAxis min should be 23.'
```
raf18seb, line 174: Sorry, I forgot to point that you should also change `23` here too )(and below in other assertions). Something like this:
```The first yAxis' min should be equal to the first point's low value.```

File: samples/unit-tests/chart/parallel-coordinates-updates/demo.js [hunk 14]
```diff
@@ -167,49 +165,46 @@ QUnit.test('Take range series points into account while calculating yAxis extrem
...
+        assert.strictEqual(
             chart.yAxis[i].translate(
-                series.dataModify ?
-                    series.dataModify.modifyValue(point.high) : point.high,
+                point.high,
                 0,
                 1,
                 0,
                 1
             ),
```
raf18seb, line 205: I think it's short enough now to have it in one line

raf18seb (changes requested):
Few more comments, see below

samples/unit-tests/chart/parallel-coordinates-updates/demo.js
```diff
@@ -144,4 +144,67 @@ QUnit.test('Update parallel coordinates plot', function (assert) {
         3,
         'After chart update parallelInfo.counter has correct value (#10081).'
     );
+
+    // Calculate yAxis extremes based on the range series points (#15752)
+    chart.update({
+        chart: {
+            parallelCoordinates: true
+        },
```
raf18seb, line 152: Is it necessary? I think it's already in the config above

samples/unit-tests/chart/parallel-coordinates-updates/demo.js
```diff
@@ -144,4 +144,67 @@ QUnit.test('Update parallel coordinates plot', function (assert) {
...
+        series.points[2].high,
+        thirdAxisMax,
+        '3rd yAxis max should be 4.'
+    );
+
+    // Calculate plotHigh value based on each yAxis scale (#15752)
+    series.points.forEach(function (
+        point,
+        i
+    ) {
```
raf18seb, line 197: We can shorten these 4 lines into 1 line.

ts/Series/AreaRange/AreaRangeSeries.ts
```diff
@@ -267,23 +267,28 @@ class AreaRangeSeries extends AreaSeries {
...
         ): void {
-
             const high = point.high,
                 plotY = point.plotY;
 
             if (point.isNull) {
                 point.plotY = null as any;
             } else {
+                const yAxis = series.chart.hasParallelCoordinates ?
+                    series.chart.yAxis[i] :
```
raf18seb, line 286: Just a note: it's not a mistake and you can leave it as it is, but you could get rid of `i` parameter and use `point.index` (or something similar) instead ;)

raf18seb (changes requested):

ts/Series/AreaRange/AreaRangeSeries.ts
K009: Nice one, thanks! 👍

samples/unit-tests/chart/parallel-coordinates-updates/demo.js
K009: I hope that line that's length is equal to exactly 80 won't be a problem. 😄

raf18seb: No, 80 is good ;)

ts/Extensions/ParallelCoordinates.ts
```diff
@@ -691,8 +691,31 @@ namespace ParallelAxis {
...
+            currentPoints.forEach(function (point): void {
+                if (isArray(point)) {
+                    max = arrayMax(point);
+                    min = arrayMin(point);
+                }
+            });
+
+            if (
+                !isArray(arrayMin(currentPoints)) ||
+                !isArray(arrayMax(currentPoints))
```
raf18seb, line 707: `arrayMin` finds a minimum value in **array of values**. Here you're trying to find a minimum value of **array of arrays**

ts/Extensions/ParallelCoordinates.ts
```diff
@@ -691,8 +691,31 @@ namespace ParallelAxis {
...
-            axis.dataMin = arrayMin(currentPoints);
-            axis.dataMax = arrayMax(currentPoints);
+            // Take into account range series points as well (#15752)
+            let min: number = NaN,
+                max: number = NaN;
+
+            currentPoints.forEach(function (point): void {
+                if (isArray(point)) {
+                    max = arrayMax(point);
+                    min = arrayMin(point);
```
raf18seb, line 701: I guess you're here trying to find extremes of yAxis. What if you have more series? Only the last series' points will be really taken into consideration

raf18seb (changes requested):
I think an initial bug is still not fixed. What if we have multiple series? http://jsfiddle.net/BlackLabel/x0zqL6hd/ (uncomment `yAxis` extremes config to see how the graph should look like)

My below comments should help you find the cause.

ts/Extensions/ParallelCoordinates.ts
K009: You've got a point!

ts/Extensions/ParallelCoordinates.ts
K009: Exactly, to check if any of the points is not an array (to take into account also points from series like line, column, etc.)

raf18seb: But why using `arrayMin` for that? Can't you just do `currentPoints[0]`?

I mean, comparing arrays using `arrayMin` is semantically wrong. If you want to distinguish if the point is a value (line, column etc.) or array (range series), then you should do it differently in my opinion

raf18seb: I took a closer look and mixed up non-range and range series: http://jsfiddle.net/BlackLabel/kLhc5j0z/

We can't use `currentPoints[0]`, it was just a blind shot, forget about it.
But I still think that we shouldn't use `arrayMin` in this specific use case. I see you're still working on it so please let me know when it's ready to review ;)

K009: Ok, you're right. This solution to the problem with mixed non-range and range series is not the best. In fact, I've got another solution.

ts/Extensions/ParallelCoordinates.ts
```diff
@@ -678,19 +678,24 @@ namespace ParallelAxis {
...
 
             axis.series.forEach(function (series): void {
-                if (
-                    series.visible &&
-                    defined((series.yData as any)[index as any])
-                ) {
-                    // We need to use push() beacause of null points
-                    currentPoints.push((series.yData as any)[index as any]);
+                if (series.yData && series.visible) {
+                    const y = (series.yData)[index as any];
```
raf18seb, line 686: I believe this `as any` can be removed here by checking if `index` is a number type.

raf18seb (changes requested):
Now it looks simple and nice ;)

Small TS comment below.
Could you add a test covering multiple series?

ts/Extensions/ParallelCoordinates.ts
```diff
@@ -678,19 +678,28 @@ namespace ParallelAxis {
...
-                currentPoints: Array<Point> = [];
+            const index = parallelCoordinates.position;
+            let currentPoints: Array<number|null> = [];
 
             axis.series.forEach(function (series): void {
                 if (
+                    series.yData &&
                     series.visible &&
-                    defined((series.yData as any)[index as any])
+                    typeof index === 'number'
```
raf18seb, line 688: Please use our Utils (`Highcharts.isNumber()`)

samples/unit-tests/chart/parallel-coordinates-updates/demo.js
```diff
@@ -144,4 +144,117 @@ QUnit.test('Update parallel coordinates plot', function (assert) {
...
+    });
+
+    chart.addSeries({
+        type: 'arearange',
+        data: [
+            [10, 15],
+            [10, 15],
+            [-3, -1]
+        ]
+    });
```
raf18seb, line 228: Some few inconsistencies:

1. You're using different data for single series tests, and new data for multiple series. Could you use one dataset? (the second preferable)
2. You're doing `chart.update()` and `chart.addSeries()` that redraw chart twice - a good pattern is to set a `redraw` flag to `false` for the first update, and redraw chart only once (better performance).
3. Why you're adding the second series using `update()`, but the third series using `addSeries`?
4. You have `type: 'arearange'` lines duplicated multiple times

Perhaps using only 2 series instead of 3 would be better (only one `addSeries` would be required). And you can set `chart: { type: 'arearange' }`.

raf18seb (changes requested):
Still some polishing required, but you're close to the finish line

samples/unit-tests/chart/parallel-coordinates-updates/demo.js
K009: All right, thanks for the feedback 👍

ts/Extensions/ParallelCoordinates.ts
```diff
@@ -678,19 +679,28 @@ namespace ParallelAxis {
...
             axis.series.forEach(function (series): void {
                 if (
+                    series.yData &&
                     series.visible &&
-                    defined((series.yData as any)[index as any])
+                    isNumber(index)
                 ) {
-                    // We need to use push() beacause of null points
-                    currentPoints.push((series.yData as any)[index as any]);
+                    const y = (series.yData)[index];
```
TorsteinHonsi, line 691: Funny parentheses...

ts/Extensions/ParallelCoordinates.ts
```diff
@@ -678,19 +679,28 @@ namespace ParallelAxis {
...
+
+                    // Take into account range series points as well (#15752)
+                    if (isArray(y)) {
+                        currentPoints = currentPoints.concat(y);
```
TorsteinHonsi, line 695: According to the internet, push is way faster than concat, so I think we should refactor this. We can even combine it into one line, using the `splat` utility function:

´currentPoints.push.apply(splat(y))`

ts/Extensions/ParallelCoordinates.ts
```diff
@@ -678,19 +679,28 @@ namespace ParallelAxis {
...
+                    } else {
+                        currentPoints.push(y);
+                    }
                 }
             });
 
+            currentPoints = currentPoints.filter((y): boolean => defined(y));
```
TorsteinHonsi, line 702: Can probably be simplified to 

`currentPoints = currentPoints.filter(defined);`

Or, even safer and probably as efficient:

`currentPoints = currentPoints.filter(isNumber);`

TorsteinHonsi (changes requested):

ts/Extensions/ParallelCoordinates.ts
K009: Sure, thanks for all the feedback.

raf18seb (approved):
Good job, thanks!!

TorsteinHonsi (approved):

PR #17600, opened 2022-08-15
Treegraph series type completion

### LINKED ISSUE #12522: New Series Type: Treegraph

pawellysy:
Added new series type, treegraph. [Read more in the docs](https://www.highcharts.com/docs/chart-and-series-types/treegraph-chart).

See #12522.

### To do:
 - [x] Animation of opacity as the nodes expand and contract.
 - [x] Correct animation when [expanding a node that started out as hidden](http://localhost:3030/samples/#view/highcharts/series-treegraph/collapsed-node)
 - [x] Design polishing (Torstein)
 - [x] Demos
   - [x] Simpler and mobile-compliant inverted demo
   - [x] Dot and label design
   - [x] Better design on the phylogenetic tree demo
   - [x] Try [annotated phylogenetic tree](https://www.biology-pages.info/P/Primates.html). This will verify that we can create trees without graphics, only placing text on the lines.
   - [x] Test toggling between treegraph, treemap and sunburst which have the same data structure and should be interchangable. Set up demo if successful (Torstein)
 - [x] Data labels should fade in and out when collapsing/expanding.
 - [x] Data labels should defer and fade in initial load (like regular Series).
 - [x] Refactor the way data labels are hidden. Remove from core as much as possible (Torstein).
 - [x] Verify docs (Torstein). 
 - [x] Color control by level [like sunburst](http://localhost:3030/samples/#view/highcharts/demo/sunburst) . The `colorByPoint` and `colorVariation` options should apply.

Closes #12522.
Continuance of PR #16805, Improved design of the treegraph series.


TorsteinHonsi (commented):
This line is suspicious:

https://github.com/highcharts/highcharts/pull/17600/files#diff-823a66b043bc26ddbba73db7273550c959258897fe881ad1cba977cf0492bcc1R103

Because `point.series` is identical to `series`, so the first and the third line of the merge is the same object... Did you mean to set `point.options.collapseButton`?

pawellysy:
Yes, this is what I ment. Nice catch, thanks!

TorsteinHonsi:
To do:
 - [ ] Animation of opacity as the nodes expand and contract.
 - [ ] Correct animation when expanding a node that started out as hidden
 - [ ] Design polishing and more demos (Torstein)

TorsteinHonsi (approved):

PR #17601, opened 2022-08-15
Migration of Debugger

bre1470:
- Migrated debugger to ESM-compatible composition.
- Converted let to const.
- Converted forEach to for.


TorsteinHonsi (approved):
Thanks!

PR #17603, opened 2022-08-15
Migration of XRangeSeries

bre1470:
- Made xrange composition ESM-compatible.
- Converted let to const, forEach to for.
- Fixed typing.


TorsteinHonsi (approved):
Thanks!

PR #17605, opened 2022-08-16
Docs: Improved typing of colors option.

### LINKED ISSUE #14156: Gradient colors not working in plotOptions for specific charts types ex. bar
Highcharts TypeScript Declarations doens't allow to set `GradientColorObject[]` on colors instead we can only set `ColorString[]` (https://api.highcharts.com/highcharts/colors). It mentions that it's possible to also set default colors on series or series.type. So I investigated `plotOptions.[type].colors`. Here it's possible to set `GradientColorObject[]` (https://api.highcharts.com/highcharts/plotOptions.bar.colors).

#### Expected behaviour
If I set `GradientColorObject[]` on `plotOptions.[type].colors` I would expect that the gradient colors would be implemented on the different series. Ex. plotOptions.bar.colors

#### Actual behaviour
No gradient colors are visible on the chart. Instead we need to set `colorByPoint` to true in order to see gradient colors. But we don't want to see a different gradient color for each point. Instead this should be possible for each serie. I expect there are more chart types where this doesn't work. If we take a pie chart as example it seems to work (https://jsfiddle.net/pmkL2nho/).

If we don't look at the TypeScript Declarations and allow to set `GradientColorObject[]` on `colors` we are able to see the gradient colors on different series.  (https://jsfiddle.net/veLny8oz/1/)

https://jsfiddle.net/veLny8oz/

#### Product version
Highcharts v8.0.0

#### Affected browser(s)
/

bre1470:
- Fixed #14156, `GradientColorObject` was an invalid type in `Options.colors`.

### DIFF
--- a/ts/Core/DefaultOptions.ts
+++ b/ts/Core/DefaultOptions.ts
@@ -116,7 +116,7 @@ const defaultOptions: Options = {
      * @sample {highcharts} highcharts/chart/colors/
      *         Assign a global color theme
      *
-     * @type    {Array<Highcharts.ColorString>}
+     * @type    {Array<(Highcharts.ColorString|Highcharts.GradientColorObject|Highcharts.PatternObject)>}
      * @default ["#7cb5ec", "#434348", "#90ed7d", "#f7a35c", "#8085e9",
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (approved):
Thanks!

PR #17606, opened 2022-08-16
Migration of Annotations and Stock Tools

bre1470:
- Migrated to class pattern
- Migrared to ESM-compatible compositions
- Splitted options and defaults
- Fixed typing
- Removed internal namespace
- Optimized code


TorsteinHonsi (approved):
Thanks! 💪

PR #17607, opened 2022-08-16
Docs: Added Tooltip.getPosition to public API.

### LINKED ISSUE #14848: Allow call to defaultPositioner from TooltipPositionerCallbackFunction
# Description of the feature
It would be helpful to call the function `defaultPositioner` from a custom positioner function (type `TooltipPositionerCallbackFunction`), e.g. when I like to fix the tooltip position only on one axis. Unfortunately, this is not really possible. `defaultPositioner` is not part of the public interface of `Tooltip` and the required arguments are different than those of `TooltiPositionerCallbackFunction`.

Highcharts

bre1470:
Closed #14848, missing type-safe access to the default tooltip positioner in a `TooltiPositionerCallbackFunction `.

### DIFF
--- a/ts/Core/Tooltip.ts
+++ b/ts/Core/Tooltip.ts
@@ -615,11 +615,22 @@ class Tooltip {
     }
 
     /**
-     * Place the tooltip in a chart without spilling over
-     * and not covering the point it self.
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (approved):
Thanks!

PR #17608, opened 2022-08-16
Docs: Added short type to sankey.data option.

### LINKED ISSUE #14849: getting error while implementing dependency wheel  "Type '[string, string, number]' has no properties in common with type 'SeriesSankeyPointOptionsObject'.ts(2559)"
Type '[string, string, number]' has no properties in common with type 'SeriesSankeyPointOptionsObject'.ts(2559)

image

image

bre1470:
Fixed #14849, possible shorted type is missing for Sankey's data option.


TorsteinHonsi (approved):
Thanks!

PR #17618, opened 2022-08-18
Fixed indicator types import/export for DTS emitter.

bre1470:


TorsteinHonsi (approved):
Looks good to me! @pawelfus can you have a look too?

ts/Stock/Indicators/VBP/VBPIndicator.ts
```diff
@@ -34,11 +41,8 @@ const {
     noop
 } = H;
 import SeriesRegistry from '../../../Core/Series/SeriesRegistry.js';
-const {
-    seriesTypes: {
-        sma: SMAIndicator
-    }
-} = SeriesRegistry;
+const SMAIndicator: typeof SMAIndicatorType = SeriesRegistry.seriesTypes.sma;
+const columnPrototype = SeriesRegistry.seriesTypes.column.prototype;
```
pawelfus, line 45: Shouldn't be:
```
import type ColumnSeriesType from '../../../Series/Column/ColumnSeries';
....
const columnProto: ColumnSeriesType = SeriesRegistry.seriesTypes.column.prototype;
```
Just like in other indicators (AOIndicator, MACDIndicator). 

Or maybe the other way around. `const columnPrototype = SeriesRegistry.seriesTypes.column.prototype;` is perfectly fine, and we can simply replace `const SMAIndicator: typeof SMAIndicatorType = SeriesRegistry.seriesTypes.sma;` with `const SMAIndicator = SeriesRegistry.seriesTypes.sma;`? Tested on RSIIndicator, seems to be fine

pawelfus (commented):
Looks good! Just one question before merge (see below)

TorsteinHonsi (approved):
Thanks!

ts/Stock/Indicators/VBP/VBPIndicator.ts
bre1470: Right, I was thinking about that too and then decided to be rather consistent than mix deconstruction and regular assignment. In the produced codes deconstructions become regular assignments, so we do not produce overhead. Also some changes are temporary until we can produce native declarations and then have the assembler replaced to use regular named imports everywhere and remove all the import const and namespaces.

pawelfus: Okay, thanks!

PR #17619, opened 2022-08-18
demo/series-linecap

kamil-musialowski:
API: Added demo for linecap.

https://api.highcharts.com/highcharts/plotOptions.series.linecap


ts/Core/Series/SeriesDefaults.ts
```diff
@@ -451,6 +451,9 @@ const seriesDefaults: SeriesOptions = {
     /**
      * The line cap used for line ends and line joins on the graph.
      *
+     * @sample {highcharts|highstock} highcharts/series/linecap/
```
raf18seb, line 454: Is it necessary? Could you generate local docs and see if it's working the same way without it?
```{highcharts|highstock}```

File: samples/highcharts/series/linecap/demo.js [hunk 2]
```diff
@@ -0,0 +1,25 @@
...
+
+    series: [{
+        data: [2, 5, 2, 3, 6],
+        linecap: 'round' // default
+    }, {
+        data: [3, 6, 3, 4, 7],
+        linecap: 'square'
+    }]
+
+});
```
raf18seb, line 25: I would personally add this demo to the `highcharts\samples\highcharts\plotoptions` folder

raf18seb (changes requested):
Thanks for a clear demo!
One question and small comment, please see below

samples/highcharts/series/linecap/demo.js
TorsteinHonsi: The `plotoptions` folder is pretty cramped... So I would prefer adding a new folder for line series on the pattern `series-{{seriestype}}`:

`/samples/highcharts/series-line/linecap`

ts/Core/Series/SeriesDefaults.ts
kamil-musialowski: You're right! It's not necessary, everything is working fine without `{highcharts|highstock}`.

raf18seb (approved):
Thanks!

PR #17620, opened 2022-08-18
Fixed #16958, differences in mapbubble API docs.

### LINKED ISSUE #16958: Problem with the mapbubble series API documentation.
#### Expected behaviour
Proper operation of the API in accordance with the implementation of TS.

#### Actual behaviour
There are significant differences in the API documentation compared to what the TS files contain:

Take a look at the animationLimit property, which is set to 500 in the code and 250 in the API:
https://api.highcharts.com/highmaps/plotOptions.mapbubble.animationLimit

Same with joinBy - it doesn't appear for mapbubble in code, but appears as joinBy by default in API: 'hc-key' ( #16750 ). tooltip.pointFormat is set but does not appear in the API.

#### Product version
v9.3.3

hubertkozik:
---
If some option is not visible in API it is a nice way to "enable" it by setting some parameters, like `@default` for `animationLimit` in this PR.

### DIFF
--- a/ts/Series/MapBubble/MapBubbleSeries.ts
+++ b/ts/Series/MapBubble/MapBubbleSeries.ts
@@ -242,6 +242,9 @@ class MapBubbleSeries extends BubbleSeries {
          * @apioption plotOptions.mapbubble.zThreshold
          */
 
+        /**
+         * @default 500
+         */
         animationLimit: 500,
         joinBy: 'hc-key',
         tooltip: {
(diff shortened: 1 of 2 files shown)


raf18seb (approved):
Thanks!

madepiet (approved):

DJTechnoo (approved):

KacperMadej (approved):

raf18seb:
Is [this](https://github.com/highcharts/highcharts/pull/17629) something that could also resolve the issue without this PR?

PR #17625, opened 2022-08-19
Brushed up speedometer design

TorsteinHonsi:
Simplified speedometer design


nasin (approved):
The changes look good! I like the simplicity of the new design and the focus on the plotbands.

PR #17626, opened 2022-08-19
Fixed series types import/export for DTS emitter.

bre1470:


bre1470:
TypeScript 4.8.2 can handle deconstructed types.

PR #17629, opened 2022-08-22
Docs: Fixed conflict of default values.

bre1470:
Basically in option nodes local generic default will win now over an inherited product default.


TorsteinHonsi (approved):
Thanks!

PR #17636, opened 2022-08-23
bugfix/14103-lollipop-marker-color

### LINKED ISSUE #14103: Marker Color is not applied for Lollipop Chart when we have data color inside the series
#### Expected behaviour
Marker Color is not applied for Lollipop Chart when we have data color inside the series, but in area chart it is workning correctly

#### Actual behaviour
Marker Color is not applied for Lollipop Chart when we have data color inside the series, but in area chart it is workning correctly

https://jsfiddle.net/dL53g7fp/1/

#### Product version

Highcharts
#### Affected browser(s)
ALL

hubertkozik:
Fixed #14103, marker fill color was not applied for the Lollipop series.


hubertkozik:
Will be continued in PR #17905 as an inheritance refactor.

PR #17640, opened 2022-08-24
Tools: Updated generators to fix defaults reflection.

### LINKED ISSUE #15575: Improper typing of key splitSeries in PlotPackedbubbleOptions
### Describe the bug
Improper typing of `PlotPackedbubbleOptions` where the type of key `splitSeries` is  `string` rather than `boolean`

### Expected behavior
The type of `splitSeries` should be boolean.

### Demo
https://stackblitz.com/edit/highcharts-angular-packedbubble
(solution submitted by @karolkolodziej on StackOverflow as a solution for missing types)

#### Steps required to recreate the problem in the demo:
1. Remove the modified interface used as a workaround to produce error.
`Type 'true' is not assignable to type 'string'.`

### Setup used
 - NodeJS version 14.15.4
 - Angular version 11.2.6
 - TypeScript version 4.1.5

bre1470:
Fixed #15575, wrong type for `plotOptions.packedbubble.layoutAlgorithm.splitSeries` (and 10 other options of multilple series types)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,8 +11,8 @@
       "devDependencies": {
         "@babel/core": "^7.18.10",
         "@babel/preset-env": "^7.18.10",
-        "@highcharts/highcharts-declarations-generator": "github:highcharts/highcharts-declarations-generator#v1.1.23",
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #17644, opened 2022-08-25
Docs: Added resetSelection flag to pointer event.

### LINKED ISSUE #16076: `resetSelection` does not exist on `ChartSelectionContextObject` Typescript type
#### Expected behaviour

`resetSelection` is a prop on `ChartSelectionContextObject` type

#### Actual behaviour

I was following this example for determining if the chart is currently zoomed in: https://stackoverflow.com/questions/66536206/styling-highcharts-reset-zoom-button-with-material-ui-button

But with typescript, I have to extend the `ChartSelectionContextObject` type on the `selection` event to add support for the property.

```
const onChartSelection: ChartSelectionCallbackFunction = (
    e: ChartSelectionContextObject,
  ) => {
    if ((e as any).resetSelection) {
      setIsZoomed(false);
    } else {
      setIsZoomed(true);
    }
    return undefined;
  };

const options: Options = {
    chart: {
      zoomType: 'xy',
      panning: {
        enabled: true,
        type: 'xy',
      },
      panKey: 'shift',
      events: {
        selection: onChartSelection,
      },
    },
  };
 ```

#### Product version
9.1.2

bre1470:
- Fixed #16076, resetSelection does not exist on selection event

### DIFF
--- a/ts/Core/Chart/Chart.ts
+++ b/ts/Core/Chart/Chart.ts
@@ -3590,7 +3590,7 @@ class Chart {
             hasZoomed;
 
         // If zoom is called with no arguments, reset the axes
-        if (!event || (event as any).resetSelection) {
+        if (!event || event.resetSelection) {
             chart.axes.forEach(function (axis): void {
                 hasZoomed = (axis.zoom as any)();
             });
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (approved):
Thanks!

PR #17649, opened 2022-08-25
Refactor, make point.graphic an array of elements

TorsteinHonsi:
Multiple graphics per point is a common pattern among series types. Now we can have a uniform way of dealing with operations like destroying and updating/animation them instead of bloating the core with series-specific stuff like `upperGraphic`. 

We can (and should?) still use the `point.graphic` property for the main graphic. In cases where the z-index of multiple graphics is the same (like item series), `point.graphic` can be a group element that holds the multiple `graphics`. In other cases (like `temperatureColors`), the graphics are intermixed and a common group cannot be used.

### DIFF
--- a/samples/unit-tests/series-arearange/markers/demo.js
+++ b/samples/unit-tests/series-arearange/markers/demo.js
@@ -37,11 +37,11 @@ QUnit.test('Markers for arearange.', function (assert) {
 
     chart.series[0].points.forEach(point => {
         assert.ok(
-            point.lowerGraphic !== undefined,
+            point.graphics[0] !== undefined,
             'Bottom marker for point: x=' + point.x + ' exists.'
         );
         assert.ok(
-            point.upperGraphic !== undefined,
+            point.graphics[1] !== undefined,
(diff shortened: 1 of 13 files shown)


raf18seb (approved):
The code related to this PR is fine, thanks for that! Approved

However, I have 2 general questions about a code not related to this PR and I think this is a good moment to ask about it.

1. All automatic tests pass. But when I open [this one](http://utils.highcharts.local/samples/#test/unit-tests/series-item/negative) manually, it fails for a negative value because `point.graphic` is missing (it worked exactly the same before this PR)
2. In the `item` series, do we need `isActive` flag and [related code](https://github.com/highcharts/highcharts/blob/29b6d46bf3ebb07d3ec6fd59589b7e52b2104bf7/ts/Series/Item/ItemSeries.ts#L338-L354)? When I removed that code, all tests passed. I also checked `point.remove()` and `point.setVisible()` and everything works fine - am I missing something or this code could just be removed?

raf18seb:
@TorsteinHonsi could you merge it to master so I can merge it to another branch? Thanks in advance

PR #17651, opened 2022-08-25
Added required XRangeSeries.compose in gantt bundle.

bre1470:
Fixed visual glitch in `gantt/xrange-series/demo`.


TorsteinHonsi (approved):
Thanks!

PR #17657, opened 2022-08-26
Popup buttons upper case first

TorsteinHonsi:
The buttons should be capitalized as is common practice in GUIs


pawelfus (changes requested):
`ellipse.spec.js` test is failing
Otherwise good!

PR #17659, opened 2022-08-26
Docs: Restored missing options of NavigationButtonOptions.

### LINKED ISSUE #17138: NavigationButtonOptions no longer has verticalAlign property
#### Expected behaviour
Highchart Option's `navigation.buttonOptions` **type** should have `verticalAlign property`

#### Actual behaviour
Highchart Option's `navigation.buttonOptions` **type** no longer has `verticalAlign` property. However, this is still a valid property as demonstrated by this [fiddle](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/navigation/buttonoptions-verticalalign/). This property got removed from the typings in v9.3.0 - it was still working in v9.2.2.

The following jsfiddle still works the move the navigation menu to the bottom yet the typescript typings don't support it
https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/navigation/buttonoptions-verticalalign/

#### Product version

Highcharts v10

#### Affected browser(s)
N/A

bre1470:
Fixed #17138, NavigationButtonOptions no longer has verticalAlign property

### DIFF
--- a/ts/Extensions/Exporting/Exporting.ts
+++ b/ts/Extensions/Exporting/Exporting.ts
@@ -38,7 +38,7 @@ import AST from '../../Core/Renderer/HTML/AST.js';
 import Chart from '../../Core/Chart/Chart.js';
 import ChartNavigationComposition from '../../Core/Chart/ChartNavigationComposition.js';
 import D from '../../Core/DefaultOptions.js';
-const { defaultOptions } = D;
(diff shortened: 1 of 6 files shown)


TorsteinHonsi (approved):
Thanks!

PR #17661, opened 2022-08-26
Docs: Fixed parameter of print callbacks in exporting module.

### LINKED ISSUE #17245: Wrong definition of ExportingBeforePrintCallbackFunction
According to API reference [here](https://api.highcharts.com/class-reference/Highcharts#.ExportingBeforePrintCallbackFunction) and typescript definition, beforePrint event gets two arguments, `chart` and `event`.
When running, I get one argument with `preventDefault`, `target` (the chart) and `type`.

https://jsfiddle.net/kqsx8rbh/
Add a `console.log(arguments)` inside beforePrint

#### Product version
10.1.0

bre1470:
Fixed #17245; parameters mismatch for ExportingBeforePrintCallbackFunction.

### DIFF
--- a/ts/Extensions/Exporting/Exporting.ts
+++ b/ts/Extensions/Exporting/Exporting.ts
@@ -1814,7 +1814,7 @@ export default Exporting;
  *
  * @callback Highcharts.ExportingAfterPrintCallbackFunction
  *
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (approved):
Thanks!

PR #17673, opened 2022-08-30
blog/demo-synchronize-multiple-axes 

karolkolodziej:
Added demo for the blog post `How to synchronize the axes on multiple charts?`.


samples/highcharts/blog/synchronize-axes-multiple-charts/demo.html
```diff
@@ -0,0 +1,56 @@
...
+
+<div class="chart" id="container1"></div>
+<div class="flex-container">
+  <div class="chart" id="container2"></div>
+  <div class="chart" id="container3"></div>
+</div>
+<div class="chart" id="container4"></div>
+
+<!-- Highcharts plugin that makes all chart's x and y axes to follow the main chart's axes. -->
+<script>
```
TorsteinHonsi, line 14: Why is this script in the demo.html instead of demo.js? If it is to make a clear separation between the plugin and the setup code I think we should do it with comments instead, like we currently do with other similar cases.

TorsteinHonsi (changes requested):

TorsteinHonsi:
Awaiting internal Slack discussion on the content of the demo.

PR #17675, opened 2022-08-30
Optimized Axis3D compositions.

bre1470:
- Removed static classes
- Simplified typing
- Splitted default options


TorsteinHonsi (approved):
Thanks!

PR #17676, opened 2022-08-30
Optimized NavigatorAxis composition.

bre1470:
- Simplified typing


bre1470:
I do not see any possibility to reduce the size increase. It looks like namespaces in modules can be compressed better than constant objects.

TorsteinHonsi (approved):
Thanks!

PR #17680, opened 2022-08-31
Better default value for legend.valueDecimals

TorsteinHonsi:
Better default value for `legend.valueDecimals` for color axis with data classes. Now it preserves the decimals of the `from` and `to` settings by default.


bre1470 (approved):
Looks good! 👍

Issue #10334, opened 2019-03-18
Dependency Wheel - Datalabels are not aligned properly


jeffersonswartz:
#### Expected behaviour
The Data labels in the Dependency wheel should be placed on the specific arc.

#### Actual behaviour
 The Data labels are aligned on the left corner

#### Live demo with steps to reproduce

https://jsfiddle.net/tj9c625s/

#### Product version

Highcharts - 7.0.3

#### Affected browser(s)
Windows - Chrome Version 72.0.3626.121 (Official Build) (64-bit)

TorsteinHonsi:
Note that this is pre-release code, dependency wheels will be available with v7.1.

In the current state of the master, link labels are not rendered: https://jsfiddle.net/highcharts/0nqx7rvk/3/

jeffersonswartz:
@TorsteinHonsi I have tested with the latest release 7.1.0 and the labels are not aligned properly.

https://jsfiddle.net/g9d31ovn/

pawelfus:
Thanks for the heads-up @jeffersonswartz !

*Internal note:*
Tried to bisect, but it's a bit tricky:
- last working commit: f6cc2cc79
- after merge organization chart branch, regressions prevent us from bisecting
- after this commit: 980738e5a - we get working dependency wheel, but with this bug

jeffersonswartz:
@pawelfus Is there any workaround to align the labels to the exact location

pawelfus:
The easiest solution is to disable link labels: https://jsfiddle.net/BlackLabel/wtpn6z0r/

Snippet:
```js
    dataLabels: {
      enabled: true,
      format: '',
      nodeFormat: '{point.name}'
    }
```

ArulValan:
Can we use a custom datalabel formatter function to get data from the series and modify the original label to display like {LinkName + LinkValue} instead of {LinkLabel}?

pawelfus:
Hi @ArulValan - yes, just as you said, you can use `dataLabels.formatter` to modify what is rendered as a link label: https://jsfiddle.net/BlackLabel/s57281wc/ (line ~65). If you have more questions, please contact our support: https://www.highcharts.com/blog/support/

raf18seb:
@TorsteinHonsi any update on this? Should we treat it as an [API bug](https://api.highcharts.com/highcharts/series.dependencywheel.dataLabels.format) or feature request (or even bug)?

PR #17681, opened 2022-08-31
Docs: Clarified data label formatters in DependencyWheel.

bre1470:
PR is related to https://github.com/highcharts/highcharts/issues/10334


TorsteinHonsi (approved):
Thanks!

PR #17682, opened 2022-08-31
Fixed #17610, centerInCategory and duplicated x's.

### LINKED ISSUE #17610: centerInCategory leaves enough space for all of the x values, but does not take into consideration that the columns with the same x value are stacked
#### Prerequisites
Column chart with `grouping: true` and `centerInCategory: true` and series containing multiple points for same x value. Columns for repeated x values are displayed overlaying each other.

#### Expected behaviour
Overlaying columns displayed with the same 'normal' distance to neighbouring columns as between any other regular columns.

#### Actual behaviour
Instead overlaying columns are rendered with empty space, enough for all repeating x values be displayed separately, without taking into account that the columns with the same x value are stacked.

https://jsfiddle.net/Cortez/h85dgapz/

#### Product version
Highcharts 10.2.0

#### Affected browser(s)
Google Chrome 104.0.5112.79

#### Additional
Related forum discussion:
https://www.highcharts.com/forum/viewtopic.php?t=49260

pawelfus:
Fixed #17610, `centerInCategory` did not consider columns with the same x-value when centering them.


TorsteinHonsi (approved):
Awesome, thanks!

TorsteinHonsi:
@pawelfus Please resolve conflict

pawelfus:
Done @TorsteinHonsi

PR #17683, opened 2022-08-31
Improve UX on data classes legend item

TorsteinHonsi:
Improved the UX on legend items when using a color axis with data classes. Hovering now renders other map areas as inactive, and clicking the legend item renders them as `null` rather than completely hiding them.

Demo with the changes applied: https://jsfiddle.net/7o3qd4av/

### DIFF
--- a/ts/Core/Axis/Color/ColorAxis.ts
+++ b/ts/Core/Axis/Color/ColorAxis.ts
@@ -897,17 +897,13 @@ class ColorAxis extends Axis implements AxisLike {
         const valueDecimals = (legendOptions as any).valueDecimals;
         const valueSuffix = (legendOptions as any).valueSuffix || '';
 
-        const getPointsInDataClass = (i: number): Point[] => {
-            const points: Point[] = [];
-            axis.series.forEach((series): void => {
-                series.points.forEach((point): void => {
-                    if (point.dataClass === i) {
-                        points.push(point);
-                    }
-                });
-            });
-            return points;
-        };
+        const getPointsInDataClass = (i: number): Point[] =>
+            axis.series.reduce((points, s): Point[] => {
+                points.push(...s.points.filter((point): boolean =>
+                    point.dataClass === i
+                ));
+                return points;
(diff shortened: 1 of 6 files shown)


ts/Core/Axis/Color/ColorAxis.ts
```diff
@@ -897,6 +897,14 @@ class ColorAxis extends Axis implements AxisLike {
         const valueDecimals = (legendOptions as any).valueDecimals;
         const valueSuffix = (legendOptions as any).valueSuffix || '';
 
+        const getPointsInDataClass = (i: number): Point[] =>
```
bre1470, line 900: Use `Array<Point>` so that it is consistent and easy to see whether something is an array or a tuple.

ts/Core/Axis/Color/ColorAxis.ts
```diff
@@ -925,28 +933,32 @@ class ColorAxis extends Axis implements AxisLike {
...
                         visible: true,
-                        setState: noop,
                         isDataClass: true,
+
+                        // Override setState to set either normal or inactive
+                        // state to all points in this data class
+                        setState: (state?: (StatesOptionsKey|'')): void => {
+                            getPointsInDataClass(i).forEach((point): void =>
+                                point.setState(state)
+                            );
```
bre1470, line 950: `for of`?

ts/Core/Axis/Color/ColorAxis.ts
```diff
@@ -988,9 +1000,9 @@ namespace ColorAxis {
         options: object;
         drawLegendSymbol: typeof LegendSymbol['drawRectangle'];
         visible: boolean;
-        setState: Function;
+        setState: Point['setState'];
         isDataClass: true;
-        setVisible: () => void;
+        setVisible: Function;
```
bre1470, line 1005: You can write `setVisible()` if you like to define it as an object-related function.

bre1470 (approved):

PR #17687, opened 2022-09-01
Tools: Removed Firefox from CI to improve stability.

bre1470:
We should have some alternative test for Firefox. Especially Annotations / Stock Tools seems to have some issues with Firefox.


TorsteinHonsi (approved):
Thanks! It's not super critical since we run the visual samples manually in Firefox before each release. But we could consider testing a subset like we do with IE8.

PR #17689, opened 2022-09-02
Migration of RangeSelector

bre1470:
- Made RangeSelector ESM-compatible.
- Added RangeSelectorComposition.
- Splitted `RangeSelectorDefaults` and `RangeSelectorOptions`.
- Removed internal namespace from RangeSelector.
- Converted let to const.


TorsteinHonsi (approved):
Thanks!

PR #17690, opened 2022-09-02
Fixed #17678, error on updating axis with extremes and no data

### LINKED ISSUE #17678: Chart redraw exception with invisible series due to skipped extremes calculation
#### Expected behaviour
When (re)drawing a chart with an invisible series, it should not throw an error.

#### Actual behaviour
If you have a Series which is invisible, without a max or min set, HighCharts will throw an error on redraw:
`Uncaught TypeError: can't access property "length", tickPositions is undefined`.

The error happens in this line:
https://github.com/highcharts/highcharts/blob/06e510915541b2215f65be7535076a8599448107/ts/Core/Axis/Axis.ts#L2057

This happens because for hidden series, the extremes are never calculated, which over a bunch of corners leads to above failure.

I could fix this issue in my deployment by just always making the following condition true:
https://github.com/highcharts/highcharts/blob/06e510915541b2215f65be7535076a8599448107/ts/Core/Axis/Axis.ts#L705

https://jsfiddle.net/BtbN/1bm86c4g/12/

Initially, this will throw the mentioned error.
Commenting in the `ignoreHiddenSeries: false` option makes it works fine, as it makes above condition pass, and calculate the extremes, which ultimately leads to tickPositions not being undefined.

#### Product version
Latest HighCharts at time of writing: 10.2.1

#### Affected browser(s)
All

TorsteinHonsi:
Fixed #17678, error on updating axis with extremes and no data
___
Spiced up with some general typesafing of `tickPositions` and `axis.min`/`axis.max`.

### DIFF
--- a/samples/unit-tests/chart/chart-addaxis/demo.html
+++ b/samples/unit-tests/chart/chart-addaxis/demo.html
@@ -1,0 +1,10 @@
+<script src="https://code.highcharts.com/stock/highstock.js"></script>
+<script src="https://code.highcharts.com/stock/highcharts-3d.js"></script>
+<script src="https://code.highcharts.com/stock/highcharts-more.js"></script>
+<script src="https://code.highcharts.com/stock/modules/exporting.js"></script>
+
+
+<div id="qunit"></div>
+<div id="qunit-fixture"></div>
+
+<div id="container"></div>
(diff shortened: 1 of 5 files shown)


samples/unit-tests/chart/chart-addaxis/demo.html
```diff
@@ -0,0 +1,10 @@
+<script src="https://code.highcharts.com/stock/highstock.js"></script>
+<script src="https://code.highcharts.com/stock/highcharts-3d.js"></script>
```
karolkolodziej, line 2: There is no need for `3d` and `highcharts-more` module

karolkolodziej (approved):
One minor comment, besides that it looks good. 
I was not able to break it.

PR #17691, opened 2022-09-04
docs/17351-animation-limit

### LINKED ISSUE #17351: animationLimit not working on initial animation of chart
#### Expected behaviour

If I set an _animationLimit_ to some value, all animations will not be shown if the number of points is greater than this value.

#### Actual behaviour

If I set an _animationLimit_ to some value, then the initial animation is shown, but if I update the series, then the animation is not shown. API says: "there is a limit that shuts down initial animation by default when the total number of points in the chart is too high. " which is not correct with actual behaviour. @TorsteinHonsi is that an expected behaviour and I am wrong?

https://jsfiddle.net/BlackLabel/deoy2a3f/

1) First the chart is loaded with animation.
2) After two seconds there is an update and the animation is not shown.

#### Product version
Highcharts v10.1.0

#### Affected browser(s)
All

hubertkozik:
Fixed #17351, changed animationLimit description and added a simple demo.

### DIFF
--- a/samples/highcharts/plotoptions/series-animationlimit/demo.css
+++ b/samples/highcharts/plotoptions/series-animationlimit/demo.css
@@ -1,0 +1,3 @@
+#container {
+    height: 400px;
+}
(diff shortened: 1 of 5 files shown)


TorsteinHonsi (approved):
Thanks!

PR #17704, opened 2022-09-07
bugfix/17639-exporting-duplicated-points

### LINKED ISSUE #17639: Exporting functionality removing duplicate datapoints in CSV/XLS/View Data Table
Problem: Hi Team, In my case I am visualizing some forecasts in the form of hourly chart - 24 data points for a day per hour, the chart works fine for normal scenario.

But when it is a daylight saving day e.g 31st Oct 2021 we receive 25 data points, 2 duplicate data points for 2 am. The visualization works fine on the chart but when I click on Download CSV/XLS/ Data Table the duplicate value is removed which we would like to have.

Let me know if you need more information.

hubertkozik:
Fixed #17639, exporting feature was removing duplicated points.

### DIFF
--- a/samples/unit-tests/export-data/basics/demo.js
+++ b/samples/unit-tests/export-data/basics/demo.js
@@ -6,36 +6,16 @@ QUnit.test('Categorized', function (assert) {
 
         xAxis: {
             categories: [
-                'Jan',
-                'Feb',
-                'Mar',
-                'Apr',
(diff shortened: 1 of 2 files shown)


pawellysy (approved):
Looks good!

TorsteinHonsi (approved):
Thanks!

PR #17707, opened 2022-09-07
Migration of DataGrouping

bre1470:
- Made DataGrouping ESM-compatible
- Added ApproximationRegistry.
- Added DataGroupingAxisComposition.
- Added DataGroupingSeriesComposition.
- Added Tooltip composition for DataGrouping.
- Splitted options.
- Converted let to const.


TorsteinHonsi:
Thanks!

Now that we're migrating the data grouping concept, can you also have a look at how it can be used on the client side? See https://www.highcharts.com/docs/working-with-data/server-side-data-grouping and https://github.com/highcharts/highcharts/blob/67b64d7e0fe7332e89f53ebb49a73321d10c0e77/test/node-datagrouping.js, which actually fails in the master.

Consider if this can be fixed easily, or alternatively if it is better to wait for the new Data Layer to handle server-side data grouping properly.

bre1470:
The `groupData` example is not working in master branch and relies on a series instance, so it is not a stand-alone function anymore. The server side example needs an update with a mockup series or we need to split the `groupData` implementation.

bre1470:
@TorsteinHonsi Please review solution with changed example.

TorsteinHonsi (approved):
Thanks! The binding to the Series is unfortunate, but since we're refactoring this with the new data layer, I don't think we should put in more work here.

PR #17709, opened 2022-09-08
Optimization of TreegraphSeries

bre1470:
- Replaced let and forEach.
- Fixed doclets.

### DIFF
--- a/ts/Series/Treegraph/TreegraphSeriesDefaults.ts
+++ b/ts/Series/Treegraph/TreegraphSeriesDefaults.ts
@@ -33,7 +33,8 @@ import { Palette } from '../../Core/Color/Palettes';
  * The best examples of the dataStructures, which best reflect this chart
  * are e.g. genealogy tree or directory structure.
  *
- * @sample highchart/series-treegraph/treegraph-chart
+ * TODO change back the demo path
+ * @sample highcharts/demo/treegraph-chart
  *         Treegraph Chart
  *
  * @extends      plotOptions.treemap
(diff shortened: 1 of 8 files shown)


ts/Series/Treegraph/TreegraphSeriesDefaults.ts
```diff
@@ -35,7 +35,7 @@ import { Palette } from '../../Core/Color/Palettes';
  *
  * TODO change back the demo path
  * @sample highcharts/demo/treegraph-chart
- *          Treegraph Chart
+ *         Treegraph Chart
```
bre1470, line 38: @pawellysy There is a todo about the missing sample. I do not see how this needs to be changed back. Can you take a look, please?

pawellysy: We agreed with @TorsteinHonsi, that Treegraph won't be released until we change the design, so I changed all demos, to not be referenced in the official docs. More info here: #17600

ts/Series/Treegraph/TreegraphSeriesDefaults.ts
```diff
@@ -22,20 +22,19 @@ import type TreegraphSeriesOptions from './TreegraphSeriesOptions';
...
 
 /**
  * A treegraph series is a diagram, which shows a relation between ancestors
  * and descendants with a clear parent - child relation.
  * The best examples of the dataStructures, which best reflect this chart
  * are e.g. genealogy tree or directory structure.
  *
- * TODO change back the demo path
- * @sample highcharts/demo/treegraph-chart
- *          Treegraph Chart
```
pawellysy, line 38: @bre1470  the idea was to change the sample's path back when PR #17600 gets merged. (apart from the fact, that there is a typo `highchart`, was this intended?)

bre1470: I think it is best if you fix the issue with treegraph samples. The warnings about invalid `@sample` will show when `npm run jsdoc` on master.

TorsteinHonsi (approved):
Thanks!

PR #17714, opened 2022-09-08
Docs: Fixed radius type of solid gauge options.

### LINKED ISSUE #11659: solid-gauge point's width cannot be set as pixels
#### Expected behaviour
In docs, it is written that we can set the size of below properties in **pixels**
- yAxis.tickLength
- pane.size
- pane.background.innerRadius
- pane.background.outerRadius
- series.radius
- series.innerRadius

#### Actual behaviour
Below properties read **percentage size only**
- series.radius
- series.innerRadius

gauge

https://jsfiddle.net/BlackLabel/rzn2s0kb

#### Product version
7.1.2

bre1470:
- Fixed #11659, `series.radius` docs mentions pixel but only percentages are implemented.
- Also moved API options to new `SolidGaugeSeriesDefaults.ts` file.


ts/Series/SolidGauge/SolidGaugeSeriesDefaults.ts
```diff
@@ -0,0 +1,224 @@
...
+ *
+ * @extends      plotOptions.gauge
+ * @excluding    dial, pivot, wrap
+ * @product      highcharts
+ * @requires     modules/solid-gauge
+ * @optionparent plotOptions.solidgauge
+ */
+const SolidGaugeSeriesDefaults: SolidGaugeSeriesOptions = {
+    /**
+     * The inner radius for points in a solid gauge. Can be given as a
```
raf18seb, line 44: What about adding "only" word?
```Can be given only as a percentage string like `"100%"`.```

(the same in other 3 places - 4 in total)

File: ts/Series/SolidGauge/SolidGaugeSeriesDefaults.ts [hunk 2]
```diff
@@ -0,0 +1,224 @@
...
+     * percentage string like `"100%"`.
+     *
+     * @sample {highcharts} highcharts/plotoptions/solidgauge-radius/
+     *         Individual radius and innerRadius
+     *
+     * @type      {string}
+     * @default   60
```
raf18seb, line 51: Shouldn't it be `"60%"` instead of `60`?

ts/Series/SolidGauge/SolidGaugeSeriesDefaults.ts
```diff
@@ -0,0 +1,224 @@
...
+
+    /**
+     * The outer radius for points in a solid gauge. Can be given as a
+     * percentage string like `"100%"`.
+     *
+     * @sample {highcharts} highcharts/plotoptions/solidgauge-radius/
+     *         Individual radius and innerRadius
+     *
+     * @type      {string}
+     * @default   100
```
raf18seb, line 92: Same as above: `100` -> `"100%"`

raf18seb (changes requested):
Thanks, just a few questions, see comments below

ts/Series/SolidGauge/SolidGaugeSeriesDefaults.ts
bre1470: Good question. I will check, maybe it supports indeed numbers, but only as percentage.

TorsteinHonsi: It does actually support numbers as percentages. Which I think is a flaw, because elsewhere in similar Highcharts settings, numbers are interpreted as pixels, and percentages are strings. Not sure to do about this. Maybe we should just remove number from the type definition and description text and leave the functionality as is?

bre1470: I agree. Should I change the defaults to strings?

TorsteinHonsi: Yes I think we should

TorsteinHonsi (approved):

TorsteinHonsi:
Pending review by @raf18seb

TorsteinHonsi (approved):
Pending review by @raf18seb

raf18seb (approved):
Thanks!

PR #17715, opened 2022-09-08
Docs: Made lineWidth option of timeline series public.

### LINKED ISSUE #13994: plotOptions.timeline.lineWidth missing from API docs
[The API docs](https://api.highcharts.com/highcharts/plotOptions.timeline) make no mention of the `lineWidth` property, which can be used to adjust the width of the line of each series on a timeline. This setting is really essential when working with multiple series on a timeline, since otherwise the default lines with the various colors overlap lines and markers and look pretty poor.

I wonder if there are any other settings from `plotOptions.line` that are relevant to `plotOptions.timeline` but aren't spelled out in the docs?

bre1470:
Fixed #13994, `lineWidth` missing in timeline options.


TorsteinHonsi (approved):
Thanks!

PR #17716, opened 2022-09-08
Docs: Excluded linkedTo from pie options.

### LINKED ISSUE #14161: Property linkedTo not working for series in pie Chart
#### Expected behaviour
In the documentation it is mentioned that when two series are linked, only the first one appears in the legend. 
https://api.highcharts.com/highcharts/series.pie.linkedTo

#### Actual behaviour
Both series are visible in the legend

https://jsfiddle.net/mnq5xzuw/1/

#### Product version
Highcharts v8.0.0

#### Affected browser(s)
/

bre1470:
Resolved #14161, linkedTo option not working in pie series.


TorsteinHonsi (approved):
Thanks!

PR #17722, opened 2022-09-09
bugfix/17288-shadow-group-not-preserved

### LINKED ISSUE #17288: Piechart series shadow is not applied on update
#### Expected behaviour

When using `chart.update` to update series, new shadow should be applied to them from the options argument (if present).

#### Actual behaviour

Shadow is not applied.

https://codepen.io/ElStaino/pen/NWyRjvm?editors=1011
Click on 'Update Chart' button.

#### Product version
Highcharts JS v10.1.0

#### Affected browser(s)
Google Chrome Version 100.0.4896.88 (Official Build) (x86_64)

hubertkozik:
Fixed #17288, pie chart series were not applied on update

~~Fixed #17288, shadowGroup was not applied after the series update.~~


TorsteinHonsi (approved):
Thanks!

PR #17728, opened 2022-09-14
Fixed #14226, auto rotation not working with step.

### LINKED ISSUE #14226: Tick label auto rotation doesn't work when step set to 1
#### Expected behaviour
Rotate the labels when the `step` declared.

#### Actual behaviour
When the label `step` it set to 1 the auto-rotation doesn't work. 

https://jsfiddle.net/BlackLabel/2kduLxr1/

#### Product version
Highcharts JS v8.2.0

#### Affected browser(s)
All

bre1470:
Fixed #14226, auto rotation of tick labels didn't work when step set to 1.


TorsteinHonsi:
Now the `autoRotation` works, but the `step` option is ignored: https://jsfiddle.net/highcharts/6u9rksod/2/

Will try to commit a fix for that.

TorsteinHonsi (changes requested):
If it passes tests now, I think all is good.

Some notes for the description, see [writing content for the changelog](https://github.com/highcharts/highcharts/tree/master/changelog#writing-content-for-the-changelog). Specifically points 1 (tagging), 5 (lower case after comma) and 6 (past tense).

       .

bre1470:
With the last changes only 1632 of the 1941 unit tests are running through. It somehow hangs but does not detect it as a failure. 🤔

bre1470:
Auto rotation always favors the least amount of ticks.
- Change autoRotation to not search for the fewest horizontal tick labels when steps are set?
- Change demo to turn off autoRotation?
- Accept visual change with fewer ticks?

TorsteinHonsi:
@bre1470 Are you referring to the visual regression in `highcharts/xaxis/labels-step`? I think it's clearly a bug, as there is actually a hard `step` setting to 2 but we get 4...

ts/Core/Axis/Axis.ts
```diff
@@ -2976,35 +2976,35 @@ class Axis {
...
+                // number of steps and a rotation closest to horizontal.
+                for (const rot of autoRotation) {
                     if (
                         rot === rotationOption ||
                         (rot && rot >= -90 && rot <= 90)
                     ) { // #3891
 
-                        step = getStep(
+                        rotStep = step || getStep(
                             Math.abs(labelMetrics.h / Math.sin(deg2rad * rot))
```
bre1470, line 2993: I think the main issue is that we use line height to measure the size of horizontal ticks. We should have some line width measurement in place or otherwise it becomes always an unreliable guessing game.

TorsteinHonsi: I think we do... We don't start rotating in the first place unless the labels are actually overlapping.

pawelfus (commented):
In general, it looks good, except for the failing test - adding any comment to get a notification when it's fixed

karolkolodziej (approved):

TorsteinHonsi (approved):

PR #17729, opened 2022-09-14
bugfix/17177-followtouchmove-update

### LINKED ISSUE #17177: Updating `tooltip.followTouchMove` does not work
#### Expected behaviour
Updating `followTouchMove` should change the bahaviour of tooltip.

#### Actual behaviour
Nothing happens

[Demo 1](https://jsfiddle.net/BlackLabel/vbgmzy3c/2/show): Updating `followTouchMove = true` => `false`:  (tooltip is displayed all the time)
[Demo 2](https://jsfiddle.net/BlackLabel/vbgmzy3c/show): Updating `followTouchMove = false` => `true`:  (tooltip is not displayed)

#### Product version
N/A

#### Affected browser(s)
N/A

#### Workaround
JS/TS: 
```ts
Highcharts.addEvent(Highcharts.Chart, 'afterUpdate', function() {
  if (this.pointer) {
    this.pointer.followTouchMove = Highcharts.pick(this.options.tooltip.followTouchMove, true);
  }
});
```
iOS/Android wrappers: https://jsfiddle.net/BlackLabel/qe8ymhc2/18/show (click on a point to change the behaviour)

kamil-musialowski:
Fixed #17177, updating `tooltip.followTouchMove` didn't work.

Demo for testing: https://jsfiddle.net/BlackLabel/c3qkuzvp/
API: https://api.highcharts.com/highcharts/tooltip.followTouchMove


TorsteinHonsi (changes requested):
Thanks for the fix!

Moving the `pick` statement out to where `followTouchMove` is used is a good idea, since it was previously only set once. 

However, we now have the same code appearing three times in the codebase, and twice in the same function:
`pick((tooltip && tooltip.options.followTouchMove), true)`

To reduce the code size, can you try one of these?
1. Restore `this.followTouchMove`, but move the assignment to `Pointer.onContainerTouchStart`. Thereby it should always be up-to-date when needed.
2. In the `pinch` function, create a local variable `const followTouchMove = touchesLength === 1 && pick(....)`.

Then you can test the impact against master using `gulp filesize`. The first approach has the advantage of only doing the calculation once, and the advantage of the second is that we avoid setting a global property.

kamil-musialowski:
Thank you for the review, it's been really helpful in understanding different approaches while working with our Core.

> 1. Restore this.followTouchMove, but move the assignment to Pointer.onContainerTouchStart. Thereby it should always be up-to-date when needed.
> 2. In the pinch function, create a local variable const followTouchMove = touchesLength === 1 && pick(....).

I have tested both of the proposed solutions, and it turns out that the second approach is a bit better (305.28kB on no.1 vs 305.22kB on no.2), so that's the one commited to this PR.

TorsteinHonsi (approved):

PR #17736, opened 2022-09-19
Support HTML in export labels

herndlm:
This allows to set `useHTML` in export label options and passes it through to the renderer in order to, well surprise, render HTML :)
Refs: https://www.highcharts.com/forum/viewtopic.php?t=34732, http://jsfiddle.net/zfngxoow

I'm a first-time contributor, although I read the guidelines, I'm not sure if I'm doing everything the right way. E.g. should some tests be added somewhere? I'm also unsure about the parameter values I had to explicitly set, like `'rect` because `undefined` was not allowed for those optional params. Are API docs (e.g. https://api.highcharts.com/highcharts/exporting.buttons.contextButton) generated I assume?

### DIFF
--- a/ts/Extensions/Exporting/Exporting.ts
+++ b/ts/Extensions/Exporting/Exporting.ts
@@ -416,7 +416,12 @@ namespace Exporting {
                 0,
                 0,
                 callback as any,
-                attr
+                attr,
+                {},
+                {},
+                {},
+                'rect',
+                btnOptions.useHTML
             )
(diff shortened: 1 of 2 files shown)


TorsteinHonsi:
Thanks! Of course the export buttons should support HTML:) I took your changes over to a local branch so we can run CI, then added docs and sample: https://github.com/highcharts/highcharts/pull/17740.

herndlm:
Oh great, the!

PR #17740, opened 2022-09-19
Enhancement: exporting button useHTML

TorsteinHonsi:
Added `exporting.buttonOptions.useHTML` to allow HTML with icons etc. in exporting buttons.

Original external PR: https://github.com/highcharts/highcharts/pull/17736

Preview: https://highcharts.github.io/highcharts-utils/samples/#gh/9dce2aae16/sample/highcharts/exporting/buttons-text-usehtml

# To do
 - [x] Handle mouseenter and mouseleave on the HTML overlay
 - [x] Make transition work in the demo
 - [x] Figure out why the [standard button doesn't work](https://jsfiddle.net/highcharts/mpa29qcr/) with `useHTML`.

### DIFF
--- a/samples/highcharts/exporting/buttons-text-usehtml/demo.css
+++ b/samples/highcharts/exporting/buttons-text-usehtml/demo.css
@@ -1,0 +1,5 @@
+#container {
+    height: 400px;
+    margin: 0 auto;
+    max-width: 800px;
+}
--- a/samples/highcharts/exporting/buttons-text-usehtml/demo.css
+++ b/samples/highcharts/exporting/buttons-text-usehtml/demo.css
@@ -3,3 +3,7 @@
     margin: 0 auto;
     max-width: 800px;
 }
+
+.highcharts-button span {
+    transition: color 250ms;
+}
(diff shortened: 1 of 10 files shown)


herndlm:
Hey, regarding
> Figure out why the [standard button doesn't work](https://jsfiddle.net/highcharts/mpa29qcr/) with useHTML
if you mean the weird hover/click behaviour - this seems to be fixable with e.g.
```css
div.highcharts-button span {
    pointer-events: none;
}
```
See also http://jsfiddle.net/zfngxoow/6/, in my PR I linked an old state that was missing that CSS

TorsteinHonsi:
> this seems to be fixable with e.g.

Yes it does, but that has some side effects in cases where we actually want to be able to interact with the span. Before I accept that solution, I'll see if it can be fixed at a deeper level.

bre1470 (changes requested):
- The code side looks good, but something is off with the implementation.
- In the preview the symbol highlights disappear as soon as I am over the symbol. The click only works on the highlight (in the padding space).
- There is also a padding issue with the right side of the highlight.
- In highcharts-utils I do not see any symbols in the preview sample. There is no network failure so I guess the CSS lacks something.

bre1470 (approved):
Tested with the wrong commit. Works flawless. 👍

PR #17766, opened 2022-09-21
enhancement/13394-temperaturemap-series-main-demo

hubertkozik:
Added the main demo for `blendColors` property in Highcharts Maps.


DJTechnoo (approved):

samples/maps/demo/blend-colors/demo.js
```diff
@@ -4,7 +4,11 @@
         'https://code.highcharts.com/mapdata/custom/world.topo.json'
     ).then(response => response.json());
 
-    Highcharts.getJSON('https://cdn.jsdelivr.net/gh/highcharts/highcharts@v7.0.0/samples/data/world-population.json', function (data) {
+    Highcharts.getJSON('https://cdn.jsdelivr.net/gh/highcharts/highcharts@b0488c7087/samples/data/busiest-europe-airports-2021.json', data => {
```
raf18seb, line 7: Will the final URL look like this?

samples/data/busiest-europe-airports-2021.json
```diff
@@ -0,0 +1,1202 @@
...
+    "type": "large_airport",
+    "name": "Berlin Brandenburg Airport",
+    "total_passengers": 9947006,
+    "lat": 52.362247,
+    "lon": 13.500672,
+    "country_name": "Germany",
+    "iso_country": "DE",
+    "region_name": "Brandenburg",
+    "iso_region": "DE-BR",
+    "municipality": "Berlin"
```
raf18seb, line 12: `name`, `total_passengers`, `lat` and `lon` properties are necessary.
What about the rest? Will we need them in the future?

samples/maps/demo/blend-colors/demo.js
```diff
@@ -4,7 +4,11 @@
         'https://code.highcharts.com/mapdata/custom/world.topo.json'
```
raf18seb, line 4: I'm wondering - what would it look like with a Europe map instead of a world map with zoom?

samples/maps/demo/blend-colors/demo.js
```diff
@@ -35,9 +47,17 @@
...
+                zoom: 3.35,
+                center: [15, 50.5]
+            },
+
             plotOptions: {
                 mapbubble: {
-                    minSize: 20,
+                    minSize: 30,
                     maxSize: 200,
                     opacity: 0.3,
```
raf18seb, line 62: I personally like more vivid colors. What about setting opacity to 0.5?

It's interesting how differently we can show the same data by manipulating bubbles sizes 🤔 
FYI @KacperMadej @madepiet @DJTechnoo @TorsteinHonsi  if you have any insights about the design

Or maybe we could add an update button to let users see how the same data can be presented differently depending on different bubbles sizes?

raf18seb (commented):
More a discussion than a changes request. See comments below

samples/maps/demo/blend-colors/demo.js
hubertkozik: I did it for the first time in my life, so I am also not sure if that is correct, I was based on: https://github.com/highcharts/highcharts/tree/master/samples/data#sample-data-folde

samples/data/busiest-europe-airports-2021.json
hubertkozik: I left this data with extra properties because maybe in the future someone could use it.

samples/maps/demo/blend-colors/demo.js
hubertkozik: It would look like this: https://jsfiddle.net/BlackLabel/jaugx91p/

In the data, there are also values for Porto Santo Airport and some airports in Russia, which are not included in `europe.topo.json` map.

samples/maps/demo/blend-colors/demo.js
KacperMadej: Here are some options for the demo that could be discussed - https://jsfiddle.net/BlackLabel/jaugx91p/2/ - changes are in code and we can discuss them on tomorrow's daily.

madepiet: @KacperMadej demo looks great to me!

samples/maps/demo/blend-colors/demo.js
TorsteinHonsi: Agreed, the world map works well for this data set

samples/maps/demo/blend-colors/demo.js
TorsteinHonsi: I agree with the vivid colors, 0.4 or 0.5 look a bit fresher.

As for the design in general, I will tinker a bit with that before the final publish. In particular, I think we should drop the spacing/margin, so that the map extends to the edge. Also the title and subtitle are a bit long, which becomes more apparent when emulating mobile phone.

samples/data/busiest-europe-airports-2021.json
raf18seb: Personally, I would get rid of it. Usually we don't add unnecessary information that is not used in the demo itself

raf18seb (approved):
I don't have any other comments, demo looks very good to me ;) let's leave the final touches for @TorsteinHonsi

samples/maps/demo/blend-colors/demo.js
TorsteinHonsi: It is okay, we need to do it that way for new demo content. Later, when the next release is tagged, we can (optionally) change it.

madepiet (approved):

PR #17770, opened 2022-09-22
Migration to LegendItem.legendItem

### LINKED ISSUE #16846: HighCharts.Series doesn't expose color, legendItem, legendSymbol
#### Expected behaviour
When using Typescript, I should be able to access color, legendItem, legendSymbol on a HighCharts.Series object, as follows:
```
const foo = (series:Highcharts.Series): void => {
     const legendItemElement:Element = series.legendItem.element;
     const legendSymbolElement:Element = series.legendSymbol.element;
     const seriesColor:string = series.color;

     //do something cool with color, legendItem, legendSymbol
}
```

#### Actual behaviour
In order to access color, legendItem, legendSymbol on a HighCharts.Series object, I have to cast it as any:
```
const foo = (series:Highcharts.Series): void => {
     const legendItemElement:Element = (series as any).legendItem.element;
     const legendSymbolElement:Element = (series as any).legendSymbol.element;
     const seriesColor:string = (series as any).color;

     //do something cool with color, legendItem, legendSymbol
}
```

I think color is a bit different, where if I cast the Series object as a specific type of PlotColumnOptions object I can get access to it, but maybe there's a better way. I don't know of any alternative, though, to getting access to legendSymbol or legendItem.

https://jsfiddle.net/longmatthewh/jfpdso2t/63/

#### Product version
Highcharts JS v9.3.2

#### Affected browser(s)
N/A

bre1470:
- Moved legendItem/legendLine/legendSymbol to legendData.item/.line/.symbol.
- Added doclets for `Highcharts.LegendItemObject`, `Highcharts.Series#color`, and `Highcharts.Series#legendItem`.
- Added type test for `Highcharts.Series#legendItem`.
- Fixed #16846


TorsteinHonsi:
It's quite a big overhead. I didn't fine read the changes, but do you have any immediate thoughts about this?
Skjermbilde 2022-09-29 kl  13 27 42

bre1470:
I optimized access, but there is not much more left. The 300 kB likely come from the added path element `.legendData.` which sums up and can not be minified. I use local variables where useful, but handling happens in many different places.
57 times x 11 bytes = 627 bytes - declarations.

ts/Core/Legend/Legend.ts
```diff
@@ -482,16 +488,22 @@ class Legend {
...
 
         // destroy SVG elements
-        ['legendItem', 'legendLine', 'legendSymbol', 'legendGroup'].forEach(
+        ['legendLine'].forEach(
             function (key: string): void {
                 if ((item as any)[key]) {
                     (item as any)[key] = (item as any)[key].destroy();
                 }
             }
         );
```
TorsteinHonsi, line 501: Is this used? I thought `item.legendLine` as moved to `legendData.line`? And if so, we should add `line` to the array at line 502.

ts/Core/Legend/Legend.ts
```diff
@@ -506,23 +518,26 @@ class Legend {
...
 
         // Destroy items
-        this.getAllItems().forEach(function (item): void {
-            ['legendItem', 'legendGroup'].forEach(destroyItems, item);
-        });
+        for (const item of this.getAllItems()) {
+            legendData = item.legendData || {};
+            destroyThat(legendData, 'group' as const);
+            destroyThat(legendData, 'item' as const);
+        }
```
TorsteinHonsi, line 537: We can probably just call `legend.destroyItem` for each item.

TorsteinHonsi: ... Then the `destroyThat` function can be inlined again.

TorsteinHonsi (changes requested):
In addition to the inline comment, I think we should look into naming. 

I think `legendItem` is a more appropriate name than `legendData`. The only downside is a possible naming collision with older HC version, where `legendItem` means the text SVGLabel itself. But implementations refering to `legendItem` will fail anyway, so I don't see that as much of a problem. 

Then if we call it `legendItem` I think what we currently call `legendData.item` should be called `legendItem.text` (or label?), which is more descriptive.

TorsteinHonsi:
To do: 

> I think `legendItem` is a more appropriate name than `legendData`.

TorsteinHonsi (approved):

PR #17772, opened 2022-09-23
Tools: Reduced dependencies

bre1470:
- Removed cross-env, fs-extra, rimraf, pixelmatch, puppeteer, tsutils.
- Replaced mkdirp with fs.mkdir.
- Fixed sass.


TorsteinHonsi (changes requested):
Thanks! We're using stylelint though. It is the CSS equivalent of ESLint.

https://github.com/highcharts/highcharts/blob/9cc8e25da44d59a3cdb2017eb611cea5e4fbaeb9/package.json#L113-L122

bre1470:
stylelint = +75 node_modules 😞

TorsteinHonsi (approved):

PR #17777, opened 2022-09-26
bugfix/17664-ichimoku-cloud-position

### LINKED ISSUE #17664: ichimoku cloud
#### Expected behaviour
Cloud to match other providers

#### Actual behaviour
 Cloud do not match

#### Product version
Highstock JS v10.2.0 (2022-07-05)

 All technical indicators for Highcharts Stock

#### 
ick
ick2

kamil-musialowski:
Fixed #17664, Ichimoku cloud was incorrectly positioned.

Following @pawellysy suggestions, I have changed the position of `senkou span A` and `senkou span B`. Moreover, the position of `chikou span` was also incorrect by one day, so it's fixed now.

All of the above was tested on our AAPL data, and compared with the tradingview AAPL data from NASDAQ.
DEMO for testing: https://jsfiddle.net/BlackLabel/o1qv4m2n/

### DIFF
--- a/ts/Stock/Indicators/IKH/IKHIndicator.ts
+++ b/ts/Stock/Indicators/IKH/IKHIndicator.ts
@@ -894,27 +894,37 @@ class IKHIndicator extends SMAIndicator {
                 IKH[i] = [];
             }
 
-            if (typeof IKH[i + period] === 'undefined') {
-                IKH[i + period] = [];
+            if (typeof IKH[i + period - 1] === 'undefined') {
+                IKH[i + period - 1] = [];
             }
 
-            IKH[i + period][0] = TS;
-            IKH[i + period][1] = KS;
-            IKH[i + period][2] = void 0;
+            IKH[i + period - 1][0] = TS;
+            IKH[i + period - 1][1] = KS;
+            IKH[i + period - 1][2] = void 0;
 
-            IKH[i][2] = CS;
+            if (typeof IKH[i + 1] === 'undefined') {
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Looks good to me! Requesting review from @pawelfus too.

ts/Stock/Indicators/IKH/IKHIndicator.ts
```diff
@@ -894,27 +894,37 @@ class IKHIndicator extends SMAIndicator {
...
 
             if (i <= period) {
-                IKH[i + period][3] = void 0;
-                IKH[i + period][4] = void 0;
+                IKH[i + period - 1][3] = void 0;
+                IKH[i + period - 1][4] = void 0;
+            }
+
+            if (typeof IKH[i + 2 * period - 2] === 'undefined') {
+                IKH[i + 2 * period - 2] = [];
```
pawelfus, line 917: Do we need the `if` condition below? Instead of checking two indexes (`IKH[i + 2 * period]` and `IKH[i + 2 * period - 2]`), check just the new index (with `-2`, since it's the one being used below).

pawelfus (changes requested):

kamil-musialowski:
@TorsteinHonsi After consulting the case with @pawelfus, we have decided to improve the tests, which allowed us to remove the unnecessary `if` statement.

pawelfus (approved):
Thanks!

TorsteinHonsi (approved):

PR #17778, opened 2022-09-26
Added newline, changed to markdown list

goransle:
Docusaurus was not happy with HTML syntax starting without newline. Also converted to Markdown for consistency


TorsteinHonsi (approved):
Thanks!

PR #17781, opened 2022-09-26
Fixed #17685, tooltip stuck with useHTML and stickOnContact

### LINKED ISSUE #17685: When useHTML and stickOnContact set, tooltip freezes sometimes.
#### Expected behaviour
Smooth tooltip refreshing.

#### Actual behaviour
When `useHTML` and `stickOnContact` set to true, tooltip freezes sometimes while hovering over chart.

http://jsfiddle.net/BlackLabel/txavf7ec/

#### Product version

#### Affected browser(s)
All

TorsteinHonsi:
Fixed #17685, the tooltip sometimes got stuck when combining the `useHTML` and `stickOnContact` options.
___

@bre1470 What do you make of this refactoring?

It turns the implementation upside down, so instead of listening for `mouseenter` and `mouseleave` setting the `tooltip.inContact` property accordingly, it listens for `mousemove` on the main container and does nothing if the target element is inside a tooltip (SVG or HTML alike). Which should effectively prevent issues caused by `mouseleave` not triggering correctly.

The inspiration is an early refactoring of mouseenter/mouseleave on the chart container itself, where I found these events to be unstable and to fail for example when the window itself lost focus. Instead I set a global mousemove event on the whole page. So this PR's idea is kind of a smaller-scale approach of the same problem. 

Do you see any obvious caveats?

### DIFF
--- a/ts/Core/Pointer.ts
+++ b/ts/Core/Pointer.ts
@@ -1182,8 +1182,7 @@ class Pointer {
             // `useHTML` options.
             !(
                 tooltip &&
-                tooltip.shouldStickOnContact() &&
-                this.inClass(pEvt.target as any, 'highcharts-tooltip')
+                tooltip.shouldStickOnContact(pEvt)
             )
         ) {
             if (
@@ -1241,6 +1240,7 @@ class Pointer {
      */
     public onDocumentMouseMove(e: MouseEvent): void {
         const chart = this.chart;
+        const tooltip = chart.tooltip;
         const chartPosition = this.chartPosition;
         const pEvt = this.normalize(e, chartPosition);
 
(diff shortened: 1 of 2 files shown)


bre1470:
Thank you. This might be a better approach. The new implementation is though not stable with edge cases when the pointer is inside the tooltip over an HTML element but outside of the series group. If I recall correctly I had a priority issue with pointer tracking between chart vs tooltip.
https://jsfiddle.net/bre1470/enzosyx7/10/

Looking at the commits for `Pointer.ts` v9.2.0: https://github.com/highcharts/highcharts/commits/v9.2.0/ts/Core/Pointer.ts
vs. commits for `Pointer.ts` 9.1.2: https://github.com/highcharts/highcharts/commits/v9.1.2/ts/Core/Pointer.ts
it is possibly one of the last two changes of v9.2.0 that breaks it.

bre1470 (approved):
Thank you for the fix! Works flawless now. 🎉

PR #17797, opened 2022-09-29
Revised homepage animation timing

nasin:
Hi, I replaced the (long) homepage animation with single candlestick chart that animates on load.


TorsteinHonsi (approved):
Thanks!

Issue #17113, opened 2022-03-22
Venn Diagram Legend


cmoihub:
# Description of the feature
As seen in this [fiddle](https://jsfiddle.net/BlackLabel/ua6vc7n4/) the highcharts legend works per series rather than on a set by set basis 
This is because the venn library requires the series to be an array of sets

# Library related to the feature
highcharts/modules/venn

# Proof of Concept/Live example for the feature
Here's a custom Legend Implementation
![image](https://user-images.githubusercontent.com/10642539/159528167-f82a6b6b-c3c7-470b-9d59-96101bd6a627.png)

pawellysy:
Hello, @cmoihub thanks for creating the feature request!

This is already possible: https://jsfiddle.net/BlackLabel/ua6vc7n4/3/
There are some small adjustments, which need to be added: 

- Add `drawLegendSymbol` method to VennSeries prototype.
- Add default option `legendType: 'point'`
- Extend the legend text to pick between point name and point id

PR #17799, opened 2022-09-29
Made  venn diagrams legend go by points name instead of series name

MarkusBarstad:
Made the legend-items of venn diagrams list the names of points instead of the names of series, see #17113 

Questions:
Eventhough my code, to my eyes at least, does what the work-around in the fiddle does, I am a bit unsure about the last point in @pawellysy list of requested changes: "Extend the legend text to pick between point name and point id"

- How do I, from/for a series set chart-options like a default `legend.labelFormatter` callback, which I assume is what is required to do here?
- And if so, is there an example of this or something similar I should use as reference?


TorsteinHonsi (changes requested):
Thanks! Would be nice to have a visual sample of this as well, both to showcase the feature, and to guard against future regressions. You can refer to the sample in the [top node of the venn series itself](https://github.com/highcharts/highcharts/blob/9cc8e25da44d59a3cdb2017eb611cea5e4fbaeb9/ts/Series/Venn/VennSeries.ts#L131-L134).

MarkusBarstad:
Hey Torstein! I have made a demo, but I am unsure about a couple of things to be mindful of with this PR:

- Its folder-name `venn-point-legend`
- Its description-text in [demo.html](https://github.com/highcharts/highcharts/pull/17799/commits/41a16fcc801cd06d820e5466acc65c0aac67a183#diff-b06f5ae5744ec1ceb2e96eb4a1c803bf0dcedaf51f164226177ee51f2217acee)
- Its details (which are straight copied from the original venn-diagram demo because I do not know who counts as a demo author) at [demo.details](https://github.com/highcharts/highcharts/pull/17799/commits/41a16fcc801cd06d820e5466acc65c0aac67a183#diff-95460244809906b77d1a7ad0f5d225914704f88caa99a71d22cc0623f5d9f1b9)

samples/highcharts/demo/venn-point-legend/demo.details
```diff
@@ -0,0 +1,10 @@
+---
+name: Venn diagram
+authors:
+  - Jon Arild Nygård
+js_wrap: b
+tags:
+  - Highcharts demo
+categories:
+  - More chart types
```
TorsteinHonsi, line 9: Tags and categories should be removed, this applies only when part of the demos in the website.

samples/highcharts/demo/venn-point-legend/demo.html
```diff
@@ -0,0 +1,12 @@
...
+<script src="https://code.highcharts.com/modules/venn.js"></script>
+<script src="https://code.highcharts.com/modules/exporting.js"></script>
+<script src="https://code.highcharts.com/modules/accessibility.js"></script>
+
+<figure class="highcharts-figure">
+    <div id="container"></div>
+    <p class="highcharts-description">
+        A venn-diagram with a legend listing the different points of its series.
+    </p>
+</figure>
```
TorsteinHonsi, line 11: Remove figure and description, keep only container

samples/highcharts/demo/venn-point-legend/demo.js
```diff
@@ -0,0 +1,53 @@
...
+            descriptionFormatter: function (point) {
+                var intersection = point.sets.join(', '),
+                    name = point.name,
+                    ix = point.index + 1,
+                    val = point.value;
+                return ix + '. Intersection: ' + intersection + '. ' +
+                    (point.sets.length > 1 ? name + '. ' : '') + 'Value ' + val + '.';
+            }
+        }
+    },
```
TorsteinHonsi, line 13: Not needed for a simple option sample

samples/highcharts/demo/venn-point-legend/demo.js
```diff
@@ -0,0 +1,53 @@
...
+    series: [{
+        showInLegend: true,
+        legendType: 'point',
```
TorsteinHonsi, line 16: Since this is the default `legendType` for venn (as of this PR), it should not be part of the config. It is also not documented in the API since it has the `@ignore-option` tag.

samples/highcharts/demo/venn-point-legend/demo.js
```diff
@@ -0,0 +1,53 @@
...
+            value: 1,
+            name: 'Not the best quality'
+        }, {
+            sets: ['Fast', 'Cheap', 'Good'],
+            value: 1,
+            name: 'They\'re dreaming'
+        }]
+    }],
+
+    title: {
```
TorsteinHonsi, line 50: For a brief explanation of the sample, add a `subtitle` saying "A venn diagram with legend enabled"

ts/Series/Venn/VennSeries.ts
```diff
@@ -132,6 +133,8 @@ class VennSeries extends ScatterSeries {
      *         Venn diagram
      * @sample {highcharts} highcharts/demo/euler-diagram/
      *         Euler diagram
+     * @sample {highcharts} highcharts/demo/venn-legend-point/
+     *         Venn diagram with a legend listing points
```
TorsteinHonsi, line 137: Since points is the only option for the legend, it suffices to say "Venn diagram with a legend"

TorsteinHonsi (changes requested):
Thanks!

> Its folder-name `venn-point-legend`

Actually we shouldn't put this under `highcharts/demo`, because that's where the contents of the demo section on the website lives. This is a detailed sample, that does not show on the website, only in the API. So an appropriate path would be `highcharts/series-venn/point-legend`. 

> Its description-text in `demo.html`

We don't need that in a simple sample. Just remove it, and keep only the container. You'll see plenty of examples for this in other demos.

> Its details

You can keep the author and add yourself in a comma-separated list.

MarkusBarstad:
Thank you for your feedback, I will implement your changes today!

TorsteinHonsi (approved):

PR #17807, opened 2022-09-30
Docs: Splitted and limited errorbar to Highcharts product.

bre1470:
- Splitted errorbar options
- Removed incomplete Highcharts Stock options from errorbar.


TorsteinHonsi (approved):
Thanks!

PR #17818, opened 2022-10-03
Fixed #17578, showTable issues with keyboard nav.

### LINKED ISSUE #17578: Keybord navigation with scroll
#### Actual behaviour
- After selecting an item in the legend the keyboard navigation (tab/shift tab) breaks and scrolls to the top of the page.
- Selecting an item in the legend re-enables the data table after hiding.

https://jsfiddle.net/FFSteve/k9z8q7go/18/

#### Product version

#### Affected browser(s)
All

oysteinmoseng:


ts/Extensions/ExportData/ExportData.ts
```diff
@@ -1016,14 +1017,22 @@ function chartToggleDataTable(
...
 
         // Generate the data table
         if (show) {
             this.dataTableDiv.innerHTML = AST.emptyHTML;
             const ast = new AST([this.getTableAST()]);
             ast.addToDOM(this.dataTableDiv);
+
+            if (createContainer || oldDisplay !== style.display) {
+                // Was hidden, is shown
+                fireEvent(this, 'afterDataTableShown', this.dataTableDiv);
```
TorsteinHonsi, line 1033: Instead of adding another event that triggers immediately before the `afterViewData` event, maybe we can just extend the functionality of the original one? If we pass a parameter like `e.wasHidden` or similar, we can just check for that in the a11y handler.

TorsteinHonsi (changes requested):
Thanks!

ts/Extensions/ExportData/ExportData.ts
oysteinmoseng: That's way smarter!

PR #17819, opened 2022-10-03
Docs: Removed unsupported plotOptions.flags.cropThreshold.

### LINKED ISSUE #17176: Wrong amount of points for series.type = 'flags' when series.cropThreshold is set to 0
#### Expected behaviour
The number of points for the series.type = 'flags' changes based on the chart's current extremes (viewable) range.

#### Actual behaviour
No change in the points array size for series.type = 'flags' when series.cropThreshold is set to 0.

This causes another issue while attempting to draw an annotation _(eg. rectangleAnnotation)_ when the chart can be panned and has a flag outside the current extremes.
Because the X-axis coordinates returned by this function will be outside of the chart and Highcharts will end the drawing:
this.chart.pointer.getCoordinates(e)

[https://jsfiddle.net/jackytanjq/8w61dyxr/1/](https://jsfiddle.net/jackytanjq/8w61dyxr/1/)
Steps: 
(i) Zoom into 1 of the flags _(Eg. From X-axis's value of 8 to 12 to view flag 2)_
(ii) Click on the button 'Log series.points.length' to see how many points are recognized by the chart.
(iii) Click on the button 'Set cropThreshold to 0'.
(iv) Click on the button 'Log series.points.length' to see how many points are recognized by the chart after the change.

#### Product version

Highcharts Stock.

#### Affected browser(s)
All

bre1470:
Fixed #17176 by removing `plotOptions.flags.cropThreshold`.

### DIFF
--- a/ts/Series/Flags/FlagsSeriesDefaults.ts
+++ b/ts/Series/Flags/FlagsSeriesDefaults.ts
@@ -38,8 +38,8 @@ import { Palette } from '../../Core/Color/Palettes.js';
  *
  * @extends      plotOptions.column
  * @excluding    animation, borderColor, borderRadius, borderWidth,
- *               colorByPoint, dataGrouping, pointPadding, pointWidth,
- *               turboThreshold
+ *               colorByPoint, cropThreshold, dataGrouping, pointPadding,
+ *               pointWidth, turboThreshold
  * @product      highstock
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (approved):
Thanks!

PR #17825, opened 2022-10-05
bugfix/17814-boost-ordinal-multiple-series

### LINKED ISSUE #17814: First and last columns not visible in boost mode with certain datasets
#### Expected behaviour
All columns, that are visible in non-boosted mode, are visible in boost mode as well, no matter the data in dataset.

#### Actual behaviour
First and last columns are not visible in boosted mode with some specific datasets.

https://jsfiddle.net/Cortez/empkuc1r/

#### Product version
Highstock v10.2.1

#### Affected browser(s)
Google Chrome v106.0.5249.61

karolkolodziej:
Fixed #17814, the first point was not correctly removed for series with only one point when boost was enabled with ordinal.

### DIFF
--- a/ts/Core/Axis/OrdinalAxis.ts
+++ b/ts/Core/Axis/OrdinalAxis.ts
@@ -960,7 +960,8 @@ namespace OrdinalAxis {
                     // between points is identical throughout all series.
                     if (
                         i > 0 &&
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (approved):
Thanks!

PR #17826, opened 2022-10-05
bugfix/17315-custom-marker-symbol

### LINKED ISSUE #17315: Custom marker symbol bug
#### Expected behaviour

If I hide and unhide a series using a legend item, the custom marker symbol will be shown correctly.

#### Actual behaviour

If I hide and unhide a series using a legend item custom marker symbol is stretched.

[](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/series-item/symbols)
1. Hide a 'male' series using a legend item.
2. Unhide a 'male' series using a legend item - the custom symbol is stretched.

#### Product version
Highcharts v10.1.0

#### Affected browser(s)
All

hubertkozik:
Fixed #17315, item series custom marker was stretched after redrawing.

### DIFF
--- a/samples/unit-tests/svgrenderer/symbol/demo.js
+++ b/samples/unit-tests/svgrenderer/symbol/demo.js
@@ -199,10 +199,11 @@ QUnit.test('Square/rect', assert => {
 
 QUnit.test('Image', assert => {
     const renderer = new Highcharts.Renderer(
-        document.getElementById('container'),
-        400,
-        300
-    );
+            document.getElementById('container'),
+            400,
+            300
+        ),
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (changes requested):
I think we need to go deeper. The problem doesn't lie in the item series, but in the renderer itself. Looks like there's a conflict with how it is using the cached image size in `symbolSizes` when rendering the same symbol with a different size.

See [this minimal example](https://jsfiddle.net/yj0mh8xr/1/), where the symbols are expected to be identical.

TorsteinHonsi:
My proposed solution is to let each of the `widthSetter` and `heightSetter` update _both_ width and height. 

```ts
/**
             * Width and height setters that take both the image's physical size
             * and the label size into consideration, and translates the image
             * to center within the label.
             */
            ['width', 'height'].forEach(function (key: string): void {
                img[key + 'Setter'] = function (value: any, key: string): void {
                    this[key] = value;

                    const {
                        element,
                        width,
                        height,
                        imgwidth,
                        imgheight
                    } = this;

                    let imgSize = this['img' + key];

                    if (defined(imgSize)) {
                        let scale = 1;

                        // Scale and center the image within its container.
                        // The name `backgroundSize` is taken from the CSS spec,
                        // but the value `within` is made up. Other possible
                        // values in the spec, `cover` and `contain`, can be
                        // implemented if needed.
                        if (
                            options &&
                            options.backgroundSize === 'within' &&
                            width &&
                            height
                        ) {
                            scale = Math.min(
                                width / imgwidth,
                                height / imgheight
                            );

                            imgSize = Math.round(imgSize * scale);

                            attr(element, {
                                width: Math.round(imgwidth * scale),
                                height: Math.round(imgheight * scale)
                            });

                        } else if (element) {
                            element.setAttribute(key, imgSize);
                        }

                        if (!this.alignByTranslate) {
                            this.translate(
                                ((width || 0) - (imgSize * scale)) / 2,
                                ((height || 0) - (imgSize * scale)) / 2
                            );
                        }
                    }
                };
            });
```

TorsteinHonsi (changes requested):

TorsteinHonsi (approved):

PR #17828, opened 2022-10-05
Tools: Added internal source declarations to es-modules.

bre1470:
Internal declarations were incomplete.


TorsteinHonsi (approved):
Thanks!

PR #17834, opened 2022-10-06
bugfix/17734-pdf-offline-exporting

### LINKED ISSUE #17734: Offline exporting to PDF fails with stroke-width="inherit"
#### Behaviour
Due to `stroke-width="inherit"` used in `Highcharts.seriesTypes.map.prototype.pointAttribs`
https://github.com/highcharts/highcharts/blob/08169996c37bfb1fd3884dfd99880cc52068a492/ts/Series/Map/MapSeries.ts#L979-L982

offline export fails with an error.

This value doesn't seem valid - no SVG spec support it or I am unable to find one.
Without this, the map seems to work (stroke will not be set if there is nothing in stroke-width so this needed to be hacked in the below demo): https://jsfiddle.net/BlackLabel/jfru7w6s/1/

https://jsfiddle.net/BlackLabel/jfru7w6s/2/

Export to PDF

#### Product version
Highmaps 10.2.1

hubertkozik:
Fixed #17734, offline exporting to PDF failed for maps.

Actually, we don't have tests for offline exporting, that's why I didn't create any new ones for that bug. My question is, should we have tests for offline exporting or it does not have sense?


TorsteinHonsi:
> My question is, should we have tests for offline exporting or it does not have sense?

It does make sense, and we should absolutely have it. The closest thing we've got are manual tests at http://utils.highcharts.local/samples/#view/highcharts/exporting/offline-download and similar (the ones marked with `requiresManualTesting` in demo.details). But if you can think of a way to automate it it would be a great improvement. The problem is we would need to run some kind of evaluation on binary content like PNGs and PDFs.

ts/Series/Map/MapSeries.ts
```diff
@@ -60,12 +60,14 @@ const {
 } = SeriesRegistry;
 import SVGRenderer from '../../Core/Renderer/SVG/SVGRenderer.js';
 import U from '../../Core/Utilities.js';
+import { isNumeric } from 'jquery';
```
TorsteinHonsi, line 63: 👀

ts/Series/Map/MapSeries.ts
```diff
@@ -974,13 +976,11 @@ class MapSeries extends ScatterSeries {
...
-            // By default set the stroke-width on the group element and let all
-            // point graphics inherit. That way we don't have to iterate over
-            // all points to update the stroke-width on zooming.
-            'inherit'
-        );
+        if (defined(pointStrokeWidth)) {
+            (attr as any)['stroke-width'] = pointStrokeWidth;
+        } else {
+            delete (attr as any)['stroke-width'];
+        }
```
TorsteinHonsi, line 983: The `any` cast is no longer needed, because `pointStrokeWidth` is a number.

ts/Series/Map/MapSeries.ts
hubertkozik: Unnoticed miss click import 😅

hubertkozik:
For now, we are leaving this ticket for a while, waiting for @TorsteinHonsi to make some checks and decision about `stroke/stroke-width `workaround.

TorsteinHonsi (approved):

PR #17838, opened 2022-10-07
Fixed #17764, centerInCategory didn't work with multiple y axes

### LINKED ISSUE #17764: With multiple axis and 'centerInCategory: true', columns are rendered one over another instead of being aligned side by side, without any way to control this
#### Expected behaviour
Columns nicely aligned side by side no matter if they belong to different yAxis (or a special api setting to control this behavior)

#### Actual behaviour
Columns are rendered one over another if they belong to different yAxis (without any way to control this)

https://jsfiddle.net/Cortez/m83gbhta/

#### Product version
Highcharts v10.2.1

#### Affected browser(s)
Google Chrome 105.0.5195.125

TorsteinHonsi:
Fixed #17764, `centerInCategory` didn't work with multiple y axes

### DIFF
--- a/ts/Core/Axis/Stacking/StackingAxis.ts
+++ b/ts/Core/Axis/Stacking/StackingAxis.ts
@@ -350,11 +350,12 @@ function seriesSetStackedPoints(
     stackingParam?: string
 ): void {
 
-    const stacking = stackingParam || this.options.stacking;
+    const chart = this.chart,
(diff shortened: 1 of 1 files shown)


TorsteinHonsi:
Fix applied to the OP demo: https://jsfiddle.net/highcharts/90hrap8x/

One think still failing is when there are null points in the second series (June).

PR #17845, opened 2022-10-11
Improved demo for stackDistance

kamil-musialowski:
Original demo didn't show the `stackDistance` property behaviour, since the flags were not on the same X value.

**API:** https://api.highcharts.com/highstock/series.flags.stackDistance

**Original demo:** https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/stock/plotoptions/flags-stackdistance/

**Improved demo:** https://jsfiddle.net/BlackLabel/vur2n6gd/


TorsteinHonsi (approved):
Thanks!

PR #17846, opened 2022-10-12
Fixed the demo for crosshair label.

kamil-musialowski:
Original demo didn't have the `yAxis crosshair label` enabled. 

Since the same demo is also used for showing the crosshair functionality, I decided to change the title instead of creating a new one (since the differences between label enabled/disabled are minor).

**Old:** https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/stock/xaxis/crosshairs-xy/
**New:** https://jsfiddle.net/BlackLabel/rj83w0h5/

**API:** https://api.highcharts.com/highstock/yAxis.crosshair
https://api.highcharts.com/highstock/yAxis.crosshair.label.enabled

### DIFF
--- a/ts/Core/Axis/AxisDefaults.ts
+++ b/ts/Core/Axis/AxisDefaults.ts
@@ -232,7 +232,7 @@ namespace AxisDefaults {
          * @sample {highcharts} highcharts/xaxis/crosshair-both/
          *         Crosshair on both axes
          * @sample {highstock} stock/xaxis/crosshairs-xy/
-         *         Crosshair on both axes
+         *         Crosshair on both axes, with yAxis label
          * @sample {highmaps} highcharts/xaxis/crosshair-both/
          *         Crosshair on both axes
          *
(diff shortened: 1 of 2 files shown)


ts/Core/Axis/AxisDefaults.ts
```diff
@@ -232,7 +232,7 @@ namespace AxisDefaults {
          * @sample {highcharts} highcharts/xaxis/crosshair-both/
          *         Crosshair on both axes
          * @sample {highstock} stock/xaxis/crosshairs-xy/
-         *         Crosshair on both axes
+         *         Crosshair on both axes, with yAxis label
```
TorsteinHonsi, line 235: Code must be either be written in backticks, `yAxis` or in plain English: y axis. I think plain English works best in this case.

TorsteinHonsi (changes requested):
Thanks! A minor proofreading note.

ts/Core/Axis/AxisDefaults.ts
kamil-musialowski: Corrected, thank you!

TorsteinHonsi (approved):

PR #15585, opened 2021-04-27
bugfix/15520-firefox-space

khlieng:
Fixed #15520, space key fired legend item click twice in Firefox.

PR #17848, opened 2022-10-13
Fixed toggling legend with spacebar in firefox

### LINKED ISSUE #17839: Firefox - Keyboard toggling of legend works only with Return, not with space bar
#### Expected behaviour
When legend entries are selected with keyboard navigation (using the a11y module) you should be able to toggle the series on and off using space bar or the return key.

#### Actual behaviour
When using Firefox browser only the return key works, the space bar does nothing. The space bar does work on the drop down menu in the following demo. 

Can be reproduced on the basic template.
https://jsfiddle.net/highcharts/LLExL/

#### Product version
Highcharts 10.2.1

#### Affected browser(s)
Firefox 106.0b8

MarkusBarstad:
Fixed #17839, spacebar would not toggle items in the legend.

Bug at https://bugzilla.mozilla.org/show_bug.cgi?id=1552419, which previously caused preventDefault to not work when pressing spacebar, seems fixed, on my end, in Firefox version 105.0.2

Hence firefox-specific code, which was added in [#15585](https://github.com/highcharts/highcharts/pull/15585) has been removed

### DIFF
--- a/ts/Accessibility/Components/LegendComponent.ts
+++ b/ts/Accessibility/Components/LegendComponent.ts
@@ -482,12 +482,8 @@ class LegendComponent extends AccessibilityComponent {
                 [
                     [keys.enter, keys.space],
                     function (
-                        this: KeyboardNavigationHandler,
-                        keyCode: number
+                        this: KeyboardNavigationHandler
                     ): number {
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (approved):
Thanks!

PR #17858, opened 2022-10-18
Fixed gantts connection lines not clearing/redrawing properly

### LINKED ISSUE #17833: Gantt chart: Connecting lines will not be fully deleted/redrawn on Drag'nDrop
#### Expected behaviour
All connecting lines will be redrawn after drag n' drop

#### Actual behaviour
If more than one connecting lines are defined, only the last defined line will be redrawn/removed on d&d. All other lines are still visible.
grafik

https://jsfiddle.net/dmt3fypL/9/

#### Product version
Highcharts Gantt JS v10.2.1 (2022-08-29)

#### Affected browser(s)
FF 102.3.0esr (64-Bit)
Chrome 102.0.5200.0
Edge 106.0.1370.34

MarkusBarstad:
Fixed #17833, all connectors weren't redrawn/deleted when updating them

The reason for this issue is the code in `Pathfinder.ts`, from line [744 to 770](https://github.com/highcharts/highcharts/blob/master/ts/Gantt/Pathfinder.ts#L744), during the update-function where it stores what connections are still present, and destroys old ones.

Specifically this conditional at line [752](https://github.com/highcharts/highcharts/blob/master/ts/Gantt/Pathfinder.ts#L752) is the culprit:
```
if (
    oldConnections[j].fromPoint ===
        pathfinder.connections[k].fromPoint &&
    oldConnections[j].toPoint ===
        pathfinder.connections[k].toPoint
)
```

It makes no distinction between lines to and from the same datapoint since they by necessity have the same `toPoint` and `fromPoint` values. This causes the code to store more than one reference to the same connection, and not storing others, on `redraw`.

Question:
- Should I implement a unique id for each connection, maybe in `Connection.ts` we can then use to distinguish between connecting lines? 
- Another option is to also check for the types of the connections, which will make the [demo](https://jsfiddle.net/dmt3fypL/9/) in the issue function, but will cause the same issue if you have multiple connections with same type, source, and destination.

TODO:
- [x] Create a fix for connecting lines which are not being cleared/redrawn properly

### DIFF
--- a/ts/Gantt/Connection.ts
+++ b/ts/Gantt/Connection.ts
@@ -844,7 +844,7 @@ class Connection {
             path: SVGPath,
             options = merge(
                 chart.options.connectors, series.options.connectors,
-                fromPoint.options.connect, connection.options
+                fromPoint.options.connectors, connection.options
             ),
             attribs: SVGAttributes = {};
 
--- a/ts/Gantt/Connection.ts
+++ b/ts/Gantt/Connection.ts
@@ -844,7 +844,7 @@ class Connection {
             path: SVGPath,
             options = merge(
                 chart.options.connectors, series.options.connectors,
-                fromPoint.options.connectors, connection.options
+                fromPoint.options.connect, connection.options
             ),
             attribs: SVGAttributes = {};
 
--- a/ts/Gantt/Pathfinder.ts
+++ b/ts/Gantt/Pathfinder.ts
@@ -748,21 +748,25 @@ class Pathfinder {
             ++j
         ) {
             found = false;
+            const oldCon = oldConnections[j];
+
             for (k = 0; k < lenNew; ++k) {
+                const newCon = pathfinder.connections[k];
+
                 if (
-                    oldConnections[j].fromPoint ===
-                        pathfinder.connections[k].fromPoint &&
-                    oldConnections[j].toPoint ===
-                        pathfinder.connections[k].toPoint
+                    typeof oldCon.options !== 'undefined' &&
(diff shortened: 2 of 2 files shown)


oysteinmoseng:
The Pathfinder module was never made with the intention of having multiple connections for the same points, so this could break in several places if we are unlucky...

Adding a UID is probably tricky here, since what we are doing is looking at the chart options and trying to match those against the graphics that have been drawn. We also need to be able to update a connection by changing its options, which means what we are really trying to do is look at the options and see if the resulting connection from those options has already been drawn.

I don't see a way to accomplish that other than to compare the options property by property (or store a hash to compare or similar). That will be overkill for 99.8% of cases, so I suggest just comparing on `type` in addition to the existing checks.

Disclaimer: I have not touched this code in years and may be totally wrong 😬

oysteinmoseng:
Nice! You could probably just do `oldCon.options && newCon.options && oldCon.options.type === newCon.options.type` in this case.

oysteinmoseng:
Actually scratch that - the connections are also the same if the types are both undefined... So maybe `(oldCon.options && oldCon.options.type) === (newCon.options && newCon.options.type)`. Alternatively `oldCon.options?.type === newCon.options?.type` would be neat, but I think that still transpiles pretty verbosely and is to be avoided.

MarkusBarstad:
This functions on my end now!

TorsteinHonsi:
Looks good to me! Asking a review from @oysteinmoseng too, to be on the safe side.

oysteinmoseng (approved):
Nice!

PR #17862, opened 2022-10-19
bugfix/17856-heatmap-default-linecolor

### LINKED ISSUE #17856: Heatmap cell borders disappear on hover (inconsistent with documentation)
#### Expected behaviour

When a heatmap chart's [`series.heatmap.marker.states.hover.lineColor`](https://api.highcharts.com/highcharts/series.heatmap.marker.states.hover.lineColor) is left unset/unspecified, the hover state of a given cell should have the same line (border) color as the normal/default state.

I assume this behavior is expected based on the documentation:

* For [`series.heatmap.marker.states.hover.lineColor`](https://api.highcharts.com/highcharts/series.heatmap.marker.states.hover.lineColor): "When `undefined`, the series' or point's lineColor for normal state is used."
* For [`series.heatmap.marker.lineColor`](https://api.highcharts.com/highcharts/series.heatmap.marker.lineColor): "When `undefined`, the series' or point's color is used."

#### Actual behaviour

When `lineColor` for the heatmap chart's hover state is left unspecified, the hovered cells on a heatmap lose their border altogether.

https://jsfiddle.net/jm7huncp/

* Note that, by default, hovering over the cells causes that cell's border to disappear
* A workaround is to add `lineColor` explicitly to the hover state (uncomment that line in the code)
* Back before Highcharts 8.1, when using `borderColor`, the color inheritance across states worked in a manner consistent with the documentation

#### Product version

Highcharts v10.2.1 (current as of writing)

#### Affected browser(s)

* Chrome (at least: this is the only browser I tested so far)

neillrobson:
Fixed #17856.

Previously, when the `lineColor` for a given heatmap marker state was left undefined, the stroke got reset entirely rather than retaining the default value. For example, if `lineColor` was not specified in the heatmap's hover state (when other fields for that state were defined), then the cell borders would disappear upon hovering on a given data point.

This patch adds a null check on the state option object's `lineColor` field, and only assigns it to the point's stroke attribute if it is defined. Consequently, when undefined, the stroke retains its default (normal state) value, in accordance with the documented behavior.

### DIFF
--- a/ts/Series/Heatmap/HeatmapSeries.ts
+++ b/ts/Series/Heatmap/HeatmapSeries.ts
@@ -634,7 +634,9 @@ class HeatmapSeries extends ScatterSeries {
                 stateOptions.color ||
                 Color.parse(attr.fill).brighten(brightness || 0).get();
 
-            attr.stroke = stateOptions.lineColor;
+            if (stateOptions.lineColor) {
+                attr.stroke = stateOptions.lineColor;
+            }
         }
 
         return attr;


neillrobson:
@raf18seb Is there anything I can do to help move this PR along? I have more time than usual this week to help out, so just let me know if I can do anything.

raf18seb:
Hi @neillrobson, thanks for the PR! Unfortunately, we need to create a new PR with your solution from our in-house branch to run all our procedures (CircleCI). I'm sorry it cannot be done from your account.
We'll create a separate Pull Request with it and some functional tests tomorrow morning.

pawellysy:
Continue a discussion here: #17896

PR #17865, opened 2022-10-20
Fix styledMode series className applied to tooltip

### LINKED ISSUE #17815: Styled mode line chart series className not applied to shared tooltip
#### Expected behaviour
Given I have a line chart, in styled mode, with a shared tooltip
When I apply a className to my series
Then I expect that className to be applied to the tooltip element for that series
Screenshot 2022-10-03 at 12 25 51

#### Actual behaviour
The className specified is not added to the values for the series in the tooltip

https://codepen.io/fhoeben/pen/WNJyeNQ

Screenshot 2022-10-03 at 12 20 34

The line colours have been configured, these are applied to the labels but not to the tooltip. The tooltip elements do not have the `weight` or `bmi` class specified to their series
Screenshot 2022-10-03 at 12 23 17

#### Product version

#### Affected browser(s)
all

MarkusBarstad:
Fixed #17815 , the tooltip didn't receive colors from `series.options.className` in `styledMode`

Not sure if this is too messy, I tried a lot of things with restructuring the tooltips formatters but It was easier and felt 'less invasive' to operate on the point's `tooltipFormatter`-function. 

It is not entirely done yet, I could avoid using a loop at [line 1141](https://github.com/highcharts/highcharts/pull/17865/commits/6b987a8a81f93553a6e06e890076fc5c416e7847#diff-b74ef0962c761aa1326fc9274578c213e2ce811d2d5ce3820bde2a64a6f772e3R1141) and use a regex instead, as is done elsewhere in `Tooltip.ts`, but I felt like I could use some pointers on how we ought structure the fix.

This fix works for both shared and non-shared tooltips if the user:

- Supplies a custom CSS-class to a series using the `series.options.className`-property, 
- Supplies a complimentary custom CSS-class to the point as per our specification i.e. `.myClass .highcharts-point {}`

TODO

- [ ] Build regex at [line 1141](https://github.com/highcharts/highcharts/pull/17865/commits/6b987a8a81f93553a6e06e890076fc5c416e7847#diff-b74ef0962c761aa1326fc9274578c213e2ce811d2d5ce3820bde2a64a6f772e3R1141)


TorsteinHonsi:
Before going too far on this, it appears to me that we can just add the `className` inside the default `pointFormat` for styled mode? It works in the OP demo. Or am I missing something?

In this line:
https://github.com/highcharts/highcharts/blob/fd06cd8c3430b29843a7dace5698fc7776a32eab/ts/Core/Tooltip.ts#L1692

Add the class name:
`                'class="highcharts-color-{$1.colorIndex} {series.options.className}"'`

MarkusBarstad:
> Add the class name: ` 'class="highcharts-color-{$1.colorIndex} {series.options.className}"'`

Then if the user wishes to have one color for the series, and another one for the point, the tooltips point not receive the users custom point color, like for example the legend does.

This is more apparent if you use a different custom color for points and series, like if you replace the CSS in the OP demo with this:

```
@import "https://code.highcharts.com/css/highcharts.css";

.weight {
  stroke: lime;
  fill: lime;
}

.weight .highcharts-point {
  fill: blueviolet;
}

.bmi {
  stroke: red;
  fill: red;
}

.bmi .highcharts-point {
  fill: cyan;
}
```

TorsteinHonsi:
What about this?

`class="highcharts-color-{$1.colorIndex} {series.options.className} {point.options.className}"`

The last class name will take precedence, in this case the point class name (if set) trumps the series class name.

TorsteinHonsi:
Working demo of the last comment: https://jsfiddle.net/highcharts/Lyoxajs5/

MarkusBarstad:
Hmm, for some reason it does not work to do `'class="highcharts-color-{$1.colorIndex} {series.options.className} {point.options.className}"'` in `styledModeFormat`.

After the `bodyFormatter` has formatted a line, I end up with an element with a whitespace where we want `point.options.className` to be like so:
`<span class="highcharts-color-1 bmi ">●</span> BMI: <b>35</b><br/>`

Trying to explore the code around `pointFormat` in the points tooltipFormatter-function since it is functioning well in the demo you posted.

TorsteinHonsi:
It worked for me... Did you remember setting a `className` option for the point itself?

```ts
        data: [
            ['Apples', 29.9, false],
            ['Pears', 71.5, false],
            {
            
            	name: 'Oranges',
                y: 106.4,
                className: 'orange'
            },
            ['Plums', 129.2, false],
            ['Bananas', 144.0, false],
            ['Peaches', 176.0, false],
            ['Prunes', 135.6, true, true],
            ['Avocados', 148.5, false]
        ],
```

MarkusBarstad:
You are entirely correct, sorry about that! I had some wrong assumptions about `point.options.className` but I get it now. Cleaning my code and comitting now.

TorsteinHonsi:
@MarkusBarstad Excellent! 

We should have a sample too. For inspiration, see https://github.com/highcharts/highcharts/pull/17893. I suggest putting _point_ and _series_ `className` in the same sample. It must also demonstrate this concept that the point class name overrides the more general series class name.

MarkusBarstad:
> @MarkusBarstad Excellent!
> 
> We should have a sample too. For inspiration, see #17893. I suggest putting _point_ and _series_ `className` in the same sample. It must also demonstrate this concept that the point class name overrides the more general series class name.

True, I will read your commits on #17893 and get on it!

MarkusBarstad:
The demo functions on my end and I think it illustrates what we want:

- Tooltips borders can be colored by `series.options.className`
- Tooltips points can be colored by `point.options.className`

I have used both the [column demo](https://jsfiddle.net/highcharts/Lyoxajs5/) you posted for structuring the CSS, and the [pull](https://github.com/highcharts/highcharts/pull/17893) you referred to for structuring the demo in general.

I have put the demo under the `css/`-folder in our `samples/`-dir, and I named the folder `tooltip-point-and-border-color`, but I am not quite sure about this name.

ts/Core/Tooltip.ts
```diff
@@ -1134,6 +1133,20 @@ class Tooltip {
...
+                            function (s): void {
+                                label.removeClass(
+                                    s.options.className as string
+                                );
+                            }
+                        );
+
+                        label.addClass(
+                            currentSeries.options.className as string
+                        );
```
TorsteinHonsi, line 1149: This code block can be removed now, right?

samples/highcharts/css/tooltip-point-and-border-color/demo.js
```diff
@@ -0,0 +1,37 @@
...
+            0]
+    }, {
+        type: 'line',
+        name: 'Oranges',
+        className: 'oranges',
+        data: [
+            1,
+            3,
+            {
+                className: 'bloodOrange',
```
TorsteinHonsi, line 31: Use kebab-case for CSS classes and ids.

TorsteinHonsi (changes requested):
* I think a more appropriate name for the sample would be `highcharts/css/point-series-classname` because what we're demonstrating is how to use the class names, not only for the tooltip.
* Also, add references to this from the API for series and point `className` options. Add `@sample` tags to the doclets.

ts/Core/Tooltip.ts
MarkusBarstad: Unfortunately if we remove it we will loose the `className`-colored border of the tooltip. I have assumed that tooltips border should be colored by series, regardless of custom point color, however I am not sure if this is appropriate.

samples/highcharts/css/tooltip-point-and-border-color/demo.js
MarkusBarstad: Fixed!

ts/Core/Tooltip.ts
TorsteinHonsi: Okay, let's simplify this. We can create a function `Tooltip.prototype.getClassName` that always delivers the full, appropriate class name for the label, then we don't need to deal with adding, removing, regexing etc. at all. See [Point.prototype.getClassName](https://github.com/highcharts/highcharts/blob/3776e7a402618302a6ac5cc534edfa69cde9096c/ts/Core/Series/Point.ts#L562-L582) for inspiration. Remember to also use it in the Tooltip `getLabel` function (where point and series are undefined I guess).

TorsteinHonsi: PS: actually it may be that we don't need to set a class name at all in `getLabel`, because it always passes `refresh` before it is actually displayed? You can check that out.

MarkusBarstad: I will!

There is some code in the `getLabel`-function which takes any classes supplied in `tooltip.options.className` (can be seen at [line 446](https://github.com/highcharts/highcharts/blob/3776e7a402618302a6ac5cc534edfa69cde9096c/ts/Core/Tooltip.ts#L446)) and includes them in the creation of the label, last arg in the assignment at [line 522](https://github.com/highcharts/highcharts/blob/3776e7a402618302a6ac5cc534edfa69cde9096c/ts/Core/Tooltip.ts#L522)

Should we move that code to the new `getClassName`-function?

TorsteinHonsi: Yes, the new `getClassName` should add the tooltip class name, the series class name and the point class name in succession. That should work 🤞

ts/Core/Tooltip.ts
```diff
@@ -1116,13 +1146,9 @@ class Tooltip {
                     });
 
                     // Set the stroke color of the box to reflect the point
-                    label.removeClass(/highcharts-color-[\d]+/g)
+                    label.removeClass(/.*/g)
```
TorsteinHonsi, line 1149: `removeClass` is not necessary. The [addClass](https://api.highcharts.com/class-reference/Highcharts.SVGElement#addClass) method has a parameter, `replace` that flushes previous classes.

TorsteinHonsi (changes requested):

TorsteinHonsi (changes requested):
Thanks! Now we need some unit tests. Uset the `TestController` to show the tooltip, then test that the className options are correctly applied in vanilla, shared and split tooltips.

MarkusBarstad:
A few notes on the qUnit tests:

- I attempted to use `TestTemplate` and `chart.container`, and I made it work locally in utils, but `npx gulp test` and the pre-commit scripts fails the same tests (the chart also doesn't look right in the sample-viewer when using templates, lacking series and colors etc.)

- Should we also test for the classes assigned to the tooltips points? (perhaps only for the non-split ones?)

- We could also use `tooltip.refresh` on some point, then check `chart.tooltip.label.element.className` instead of using `TestController`

MarkusBarstad:
A few notes on the qUnit tests:

- I attempted to use `TestTemplate` and `chart.container`, and I made it work locally in utils, but `npx gulp test` and the pre-commit scripts fails the same tests (the chart also doesn't look right in the sample-viewer when using templates, lacking series and colors etc.)

- Should we also test for the classes assigned to the tooltips points? (perhaps only for the non-split ones?)

- We could also use `tooltip.refresh` on some point, then check `chart.tooltip.label.element.className` instead of using `TestController`

TorsteinHonsi:
Looks like the tests pass now 🎉 

The last remaining change (I think) is to optimize for the test suite performance. Every time we create a new chart it costs processing time. In this case we can generate the chart once, then do  `chart.update` or `chart.tooltip.update` between testing the different types of tooltips.

MarkusBarstad:
Alright! 🎉  I will get on it!

TorsteinHonsi (approved):

PR #17875, opened 2022-10-21
Samples: Sync axes and tooltips

TorsteinHonsi:
A continuation of the topic of the blog post [How to synchronize the axes of multiple charts](https://www.highcharts.com/blog/tutorials/how-to-synchronize-the-axes-on-multiple-charts/).

The following adjustments are made:
* No master chart. Zooming on either chart is reflected on all others. Initial extremes are set as the union of all charts' extremes instead of the master.
* One common Reset Zoom button.
* Tooltips are synchronized.
* The `tickPositions` are not hard set, instead allowed to be calculated based on the chart size and the synchronized min and max.

[See the preview here](https://highcharts.github.io/highcharts-utils/samples/#gh/1084683bb0/sample/highcharts/blog/synchronize-axes-multiple-charts).

### DIFF
--- a/samples/highcharts/blog/synchronize-axes-multiple-charts/demo.css
+++ b/samples/highcharts/blog/synchronize-axes-multiple-charts/demo.css
@@ -3,6 +3,20 @@
     flex: 1;
 }
 
+#reset-zoom {
+    float: right;
+    padding: 1em;
+    visibility: hidden;
+}
+
+#reset-zoom.visible {
+    visibility: visible;
+}
+
+.dashboard {
+    clear: both;
+}
+
 .flex-container {
     display: flex;
     justify-content: center;
--- a/samples/highcharts/blog/synchronize-axes-multiple-charts/demo.details
+++ b/samples/highcharts/blog/synchronize-axes-multiple-charts/demo.details
@@ -1,6 +1,6 @@
 ---
- name: Synchronize the axes on multiple charts
+ name: Synchronize the axes and tooltips on multiple charts
  authors:
-   - Karol Kolodziej
+   - Karol Kolodziej, Torstein Hønsi
  js_wrap: b
 ...
--- a/samples/highcharts/blog/synchronize-axes-multiple-charts/demo.html
+++ b/samples/highcharts/blog/synchronize-axes-multiple-charts/demo.html
@@ -3,9 +3,13 @@
 <script src="https://code.highcharts.com/modules/exporting.js"></script>
 <script src="https://code.highcharts.com/modules/accessibility.js"></script>
 
-<div class="chart" id="container1"></div>
-<div class="flex-container">
-  <div class="chart" id="container2"></div>
-  <div class="chart" id="container3"></div>
+<button id="reset-zoom">Reset zoom</button>
+
+<div class="dashboard">
+  <div class="chart" id="container1"></div>
+  <div class="flex-container">
+    <div class="chart" id="container2"></div>
+    <div class="chart" id="container3"></div>
+  </div>
+  <div class="chart" id="container4"></div>
 </div>
-<div class="chart" id="container4"></div>
--- a/samples/highcharts/blog/synchronize-axes-multiple-charts/demo.js
+++ b/samples/highcharts/blog/synchronize-axes-multiple-charts/demo.js
@@ -8,48 +8,91 @@ const global = [73, 73, 74, 75, 76, 78, 79, 80, 82, 83, 83, 84, 84, 84, 85, 85,
     seAsia = [64, 65, 65, 66, 66, 71, 72, 73, 75, 78, 80, 82, 83, 85, 87, 88,
         87, 90, 91, 90, 85, 82];
 
-// Highcharts plugin that makes all chart's x and y axes to follow the main
-// chart's axes.
+// Highcharts plugin that synchronizes axes and tooltips between charts
 (() => {
-    function syncAxis(e) {
-        Highcharts.charts.forEach(chart => {
-            const mainChart = Highcharts.charts[0];
 
-            if (chart !== mainChart) {
-                if (chart[e.target.coll][0].setExtremes) {
-                    chart[e.target.coll][0].update({
-                        min: e.min,
-                        max: e.max,
-                        tickPositions: this.tickPositions
-                    });
-                }
-            }
-        });
-    }
-
-    Highcharts.addEvent(Highcharts.Axis, 'afterSetExtremes', e => {
-        if (e.target.chart.options.chart.otherChartsFollowAxes) {
-            syncAxis.call(e.target, e);
-        }
-    });
-
+    // On loading the last chart, find the union extremes of all charts and
+    // update each
     Highcharts.addEvent(Highcharts.Chart, 'load', () => {
-        const mainChart = Highcharts.charts[0];
 
         if (
-            mainChart.options.chart.otherChartsFollowAxes &&
-        Highcharts.lastChartRender
+            // When the last chart is loaded
+            Highcharts.charts.length ===
+            document.querySelectorAll('.chart').length
         ) {
-            mainChart.xAxis[0].setExtremes(
-                mainChart.xAxis[0].min,
-                mainChart.xAxis[0].max
-            );
-            mainChart.yAxis[0].setExtremes(
-                mainChart.yAxis[0].min,
-                mainChart.yAxis[0].max
-            );
+            const extremes = Highcharts.charts.reduce((extremes, chart) => {
+                extremes.xMin = Math.min(extremes.xMin, chart.xAxis[0].dataMin);
+                extremes.xMax = Math.max(extremes.xMax, chart.xAxis[0].dataMax);
+                extremes.yMin = Math.min(extremes.yMin, chart.yAxis[0].dataMin);
+                extremes.yMax = Math.max(extremes.yMax, chart.yAxis[0].dataMax);
+
+                return extremes;
+
+            }, {
+                xMin: Infinity,
+                xMax: -Infinity,
+                yMin: Infinity,
+                yMax: -Infinity
+            });
+
+            Highcharts.charts.forEach(chart => {
+                chart.xAxis[0].setExtremes(extremes.xMin, extremes.xMax);
+                chart.yAxis[0].setExtremes(extremes.yMin, extremes.yMax);
+            });
         }
     });
+
+    // When the extremes of either axis is are changed, cascade to the
+    // corresponding axis of the other charts.
+    Highcharts.addEvent(Highcharts.Axis, 'afterSetExtremes', e => {
+        const axis = e.target;
+        if (e.trigger === 'zoom') {
+            Highcharts.charts.forEach(chart => {
+                if (chart !== axis.chart) {
+                    if (chart[e.target.coll][0].setExtremes) {
+                        chart[e.target.coll][0].setExtremes(e.min, e.max);
+                    }
+                }
+            });
+        }
+    });
+
+    // Use a common reset-zoom button
+    Highcharts.addEvent(Highcharts.Chart, 'beforeShowResetZoom', () => {
+        const btn = document.getElementById('reset-zoom');
+        btn.classList.add('visible');
+        if (!btn.classList.contains('initialized')) {
+            btn.addEventListener('click', () => {
+                Highcharts.charts.forEach(chart => chart.zoomOut());
+                btn.classList.remove('visible');
+            });
+        }
+        btn.classList.add('initialized');
+
+        return false;
+    });
+
+
+    // Use a common tooltip
+    let triggerPoint;
+    Highcharts.addEvent(Highcharts.Tooltip, 'refresh', e => {
+        if (!triggerPoint) {
+            const triggerChart = e.target.chart;
+            triggerPoint = triggerChart.hoverPoint;
+            Highcharts.charts
+                .filter(chart => chart !== triggerChart)
+                .forEach(chart => {
+                    const matchingPoint = chart.series[0].points.find(p =>
+                        p.x === triggerPoint.x
+                    );
+                    if (matchingPoint) {
+                        matchingPoint.onMouseOver();
+                    }
+                });
+            triggerPoint = undefined;
+        }
+    });
+
 })();
 
 Highcharts.setOptions({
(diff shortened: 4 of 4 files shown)


rdp1414:
Hello,

Thanks for addressing the changes. In our use-case, all the synced charts have equal plot-area size - See [fiddle](https://jsfiddle.net/gr6u1ojk/). I have 3 queries on which please give your thoughts:
1. When the data ranges are little apart, say - 1st chart range is `[0, 100]`, 2nd chart range is `[10, 110]`, ..., then upon zoom and reset-zoom, charts don't go back to their state when charts were initially rendered - See [fiddle](https://jsfiddle.net/40eyc7am/).
2. When the data plotted in charts are quite apart, say - 1st chart range is `[0, 100]`, 2nd chart range is `[50, 150]`, ..., then even in their initial render, 1st chart's yAxis is not in sync with other chart's yAxis - See [fiddle](https://jsfiddle.net/cx3vqyad/1/). 
> Initial extremes are set as the union of all charts' extremes instead of the master.

Till how far the data ranges in charts can be?

3. For the charts laid out in a row - See [fiddle](https://jsfiddle.net/jrdv76s2/2/). We want to share a common yAxis: Labels and title of only the left-most chart of a row are present and not others- See [fiddle](https://jsfiddle.net/Lgysbfc1/1/). But as you can see that there are 2 issues here. Can you please suggest something here?
    1. The yAxis labels disabled for 2nd and 3rd chart are causing wide gaps between.
    2. The yAxis title of 1st (leftmost) chart is cut.

TorsteinHonsi:
1. Fixed, see https://github.com/highcharts/highcharts/pull/17875/commits/d99429c7756bce3ab52ed408671cf366b0d7eeb4 . With this change, we synchronize the axes not only on initial render, but also after zooming out.

2. When setting min and max, the exact layout of ticks depends on what is rendered from before. To make it more predictable, try setting a `tickAmount` setting to all Y axes - see [jsfiddle](https://github.com/highcharts/highcharts/pull/17875/commits/d99429c7756bce3ab52ed408671cf366b0d7eeb4)

3. It is one setting that causes both these effects: 
    `marginLeft: 50, // Keep all charts left aligned`
This setting is there to assure vertical alignment. But in this case you don’t want vertical alignment, so you can either remove it from the global options, or set it to `undefined` in the individual chart options. See [fiddle](https://jsfiddle.net/highcharts/hkn3dcm6/).

4. Currently this event is not in the options structure. However, you can add it to chart instances like this:
```js
const chart = Highcharts.chart(…);
Highcharts.addEvent(chart.tooltip, ‘refresh’, myHandler);
```
Optionally, you can set a custom option to identify the charts that are part of the tooltip sync, and check that in the common handler.

5. What happens is that the y axes are kept synchronized, which is how we set up this demo. I agree it would make sense, when `zoomType === 'x'`, that only the x axes were synchronized. Then all charts would zoom to the same x extremes, but the y axes would respond to the individual charts' data extremes. This would also apply to initial values. If this is what you'd expect, then I can modify the demo to assure greater granularity. Perhaps we should even introduce a custom option to assign a sync-group to every axis. That way we could group charts with different levels of synchronization. What do you think about that?

rdp1414:
Hello,

Thanks for the first 4 points - Many queries are solved.

> If this is what you'd expect, then I can modify the demo to assure greater granularity. 

Yes, that would be desirable. If granularity is not achieved i.e. yAxis doesn't modify in x-axis range `[0, 10]` (see [fiddle](https://jsfiddle.net/v2u7rz4q/)) as it does in normal chart (see [fiddle](https://jsfiddle.net/v2u7rz4q/1/)) then the intent of user to "zoom" gets defeated, I think.

> Perhaps we should even introduce a custom option to assign a sync-group to every axis. That way we could group charts with different levels of synchronization. What do you think about that?

If I understand correctly; please correct me if not: After initial synchronization, zoom, reset-zoom, etc (all which is done in the demo now), invoke a function which does - 
1. Loop through each axis (over multiple charts) which are shared and get the max series value plotted on that axis in the current view.
2. Calculate the maximum among those maximums (say `maxDataMax`).
3. Similarly for minimums (say `minDataMin`).
4. Then apply `setExtremes(minDataMin, maxDataMax)` to any one of the axis because the already axes synchronization logic in place will sync it to other connected axes as well. Here, we should be careful about unwanted feedbacks, etc.

If you can share a demo which achieves granularity, that'll be helpful.

TorsteinHonsi:
Hi @rdp1414 , sorry for the late reply!

In the latest commit, I implemented a `custom.syncGroup` option. I think it is a promising approach. Please see the [live demo on jsFiddle](https://jsfiddle.net/highcharts/103tuavs/) and see what you think. Instructions in the fiddle.

nishikantparmariam:
Hi @TorsteinHonsi, I see there are couple of problems with the solution here, can you please help?

See reproducer - **https://jsfiddle.net/kbzarjg5/3/**. Expectations - 

- Y-axes of both charts are synced, in all types of zooms x, y or xy. 
- X-axes are not synced.
- Initial y range of both charts or after clicking reset zoom button is union (i.e. min is min of the two and max is max of the two)
- Consider only zooming into first chart should affect the second. Vice-versa is not required for simplicity of demo.

### Problem 1 - Y-axis scale is not changing upon x zoom

> There is one problem: If you zoom into any chart in the x-range [0, 10], correspondingly y-range doesn't change which happens in a normal chart

I reproduced it separately - **https://jsfiddle.net/rybzga3e/**, it seems to happens if we had called `setExtremes()` on y-axis (remove last line and things work), is there a way to resolve this? We need an initial `setExtremes` call just after all charts have been loaded to set y-axis to union of all y-axis ranges.

y-axis-is-not-zoomed

### Problem 2 - Y-axis sync does not work properly

Reproducer - **https://jsfiddle.net/kbzarjg5/2/** (Just changed zooming type to `y`)

y-axis-sync-is-wrong

I tried to set this range in other chart using `chart.yAxis[0].update()` so that y-range of both charts look exact but it does not work.

TorsteinHonsi:
@nishikantparmariam 

### Problem 1
I think resetting the `userMin` and `userMax` properties works: https://jsfiddle.net/highcharts/2yvmq4g3/

### Problem 2
The issue seems to be that `setExtremes` adapts to the `startOnTick`, `endOnTick`, `minPadding` and `maxPadding` settings, so that the ticks don't come out identical when the axes are of different pixel lengths. One way to overcome that is to set the `tickAmount` to the same value: https://jsfiddle.net/highcharts/z3fm0qr8/ . Possibly we could automate this too.

nishikantparmariam:
@TorsteinHonsi 

I tried this but when both x and y are shared their union extremes aren't applied e.g. in this example only y-axis is set to union and not y-axis even though both should be set to union - https://jsfiddle.net/Lt9a5zf2/4/

This works if I remove the userMin and userMax lines from both. Can you please check this?

nishikantparmariam:
Ok, I found out that setting userMin and userMax for both axes after setExtremes call works but individually right after setExtremes does not work - https://jsfiddle.net/7oz9b5jd/ (does not work) and https://jsfiddle.net/7oz9b5jd/1/ (works)

Is this correct?

nishikantparmariam:
@TorsteinHonsi I am facing multiple problems here and have tried many approaches, can you please help?

### Syncing the ticks

> One way to overcome that is to set the tickAmount to the same value: https://jsfiddle.net/highcharts/z3fm0qr8/ . Possibly we could automate this too.

While this works but even upon zooming the tickAmount is always the same if it has been set once which is not what users may want. Because normal zooming in can change tickAmount. 

So, I tried to sync tickPositions instead but ran into more problems - 


- Attempt 1 - If I set tickPositions w.r.t last chart things don't work - https://jsfiddle.net/n7cwzgkm/4/, ideally I expected that after setExtremes call I can take tickPositions from any arbitrary chart? Can you please comment on ideal thing here.

- Attempt 2 - Say I set w.r.t first chart then it worked but the `setExtremes` of other charts does not work when first chart is zoomed (https://jsfiddle.net/n7cwzgkm/2/, line 100).

- Attempt 3 - I changed setExtremes to instead use axis.update but ran into problem of Attempt 1 where tick positions are not proper upon zooming into last chart and then resetting zoom - https://jsfiddle.net/n7cwzgkm/5/. See below GIF - 

wrong-ticks


In all these attempts there are some problems.

----------

### Zooming into area having no data points

Another problem we're facing is that when a chart has been set to union of extremes of all other charts, then when zooming on area where it has no data, it zooms into the place where it has data points, it can be possible that other charts could have data in that area, how do we implement to zoom exactly where selection happened?

**https://jsfiddle.net/hac7tugx/**

Notice that I zoomed from -14 to -12 but final zoomed in chart's x-values are very different that will sync other chart's also to these x-values and not -14 to -12 - 

zooming-outside-data

I tried overriding default zooming behavior of chart using chart.events.selection - https://jsfiddle.net/4vr1mso5/ but again ran into Problem 1 from https://github.com/highcharts/highcharts/pull/17875#issuecomment-2161015654 where zooming into x does not change y-axis scale.

hubertkozik:
Hi @nishikantparmariam!

### Syncing the ticks
As Torstein suggested, we can dynamically calculate the `tickAmount` option and update the chart accordingly. Thanks to that, zooming can still change the tickAmount properly. What do you think about this solution?

**Live demo:** https://jsfiddle.net/BlackLabel/bk8xtuo1/

### Zooming into area having no data points
You are facing this issue because you've used the `setExtremes` method and that's why during the zooming chart "doesn't know" about the actual min and max values of the axis. I suggest using the `axis.update` method instead of `setExtremes` and everything seems to work correctly.

**Live demo:** https://jsfiddle.net/BlackLabel/0vy6os2q/

nishikantparmariam:
Hi @hubertkozik, thanks for the reply!

### Syncing the ticks
> zooming can still change the tickAmount properly

While we can calculate the initial tickAmount, it seems that zooming is always respecting the same tickAmount.

### Zooming into area having no data points
> I suggest using the `axis.update` method instead of `setExtremes` and everything seems to work correctly.

While this works, I again run into Problem 1 from https://github.com/highcharts/highcharts/pull/17875#issuecomment-2175629344 where y-axis scale is not changing upon x zooming, reproducer - https://jsfiddle.net/26tpvzfh/

hubertkozik:
@nishikantparmariam

### Syncing the ticks
Yes, `tickAmount` was always calculated only on chart load. If you need, you can calculate it also when the extremes of an axis is changed: https://jsfiddle.net/BlackLabel/kt9fc1sL/

### Zooming into area having no data points
In that case, you can simply reset the previously set min and max values on selection event: https://jsfiddle.net/BlackLabel/ze8sybfk/

nishikantparmariam:
Thanks @hubertkozik, that helped. I am running into few more issues - 

## Zooming into area having no data points

> In that case, you can simply reset the previously set min and max values on selection event:

Reproducer - **_https://jsfiddle.net/yb3fuwg6/_**

Implemented this but it seems like this reset can also cause other charts to run sync due to `afterSetExtremes` of this chart. So, should I prevent other charts from running the effect of this call or not, I tried both cases and both of them are buggy and have expected behavior in some cases - 

### Preventing other charts from syncing (lock in `chart.events.selection`)

In below GIF, you can notice that -
 
- Zooming into first chart changes y range from 0 to 20k, but second/third chart have no effect of it due to lock. Check console only x is synced to other two charts -
  image

- Subsequent zooming into second chart changes its y range and x range and that is synced perfectly to first and third chart as expected.
 
locking

### Allowing other charts to sync  (no lock in `chart.events.selection`)

(Commented line 185 and 194 in above reproducer)

- Zooming into first chart changes its y range and x range and both are synced as expected to second and third chart - 
  
  image

- Subsequent zooming to second chart never changes its y range itself similar to problem 1 from https://github.com/highcharts/highcharts/pull/17875#issuecomment-2161015654

unlocking


----------


## Infinite Zooming

Seems like due to min/max/tickAmount being set, upon zooming there is infinite zooming happening, notice that x-values go up to 8 decimals, that does not happen usually. Will it happen in datetime as well? What will happen for categorical axis, is there a way to prevent this in generalized way?

infinite-zooming

I found out [minRange](https://api.highcharts.com/highcharts/xAxis.minRange), should I set it manually based on its definition?

image


---------

# Expectations

In general we're looking for a solution like this - 

- X-axis and/or y-axis of all charts should show same ticks if they are synced.

- Say zoom happens (`x` / `y` / `xy` type) on a chart, there are two cases - 

   - Normal zoom - For this chart, zoom should happen as it would happen normally.
   - Where there exists no data in this chart in zoomed area (case like [this](https://github.com/highcharts/highcharts/pull/17875#:~:text=are%20some%20problems.-,Zooming%20into%20area%20having%20no%20data%20points,-Another%20problem%20we%27re)) -
        - If zoom type is `x` ( https://jsfiddle.net/2h7ud4jz/ ) - After zooming, the y range can be same as full min and max like before zooming, so ticks are shown atleast, instead of not showing any ticks like this that will cause problems in syncing other charts that share a y-axis with this chart -  
    
        zooming-into-no-data

      


        - If zoom type is `y` - Similar to `x` zoom
        - If zoom type is `xy` (https://jsfiddle.net/2h7ud4jz/1/) - I had to remove reseting min/max in `chart.events.selection` to make it work - 

           
        zooming-into-no-data-xy

   


  In all cases, we want that if x-axis and/or y-axis of a chart is synced to other charts, their extremes and ticks are aligned perfectly to this chart irrespective of those charts having data in that region. 

## Optimization Question

It seems that sometimes syncing both x and then say y in `afterSetExtremes` could be tricky and expensive also. Is there another way to run `the syncing to other charts at once after both axis` have finished changing extremes due to zoom instead of changing it individually in `afterSetExtremes`?

This way it would be faster as well, because we can pass `redraw=False` in `update` and `setExtremes` methods for both the axes and a final `chart.redraw()` would do the thing?

nishikantparmariam:
## Expectations

Consider a case where x & y-axis both are shared and zooming type is `x`.

  - On first render, x and y-axis of each chart should be set to union of all charts and ticks should align properly i.e. no. of ticks, their positions and ticks themselves should be identical across charts.

  - Say zooming in x happened on a chart (this chart may / may not have data in that zoomed region) so -
  
     - For x-axis - We zoom into this chart and all other charts sync it, at this point y-range of individual charts could be different depending upon data in them for this x-region.
     - For y-axis - We take union of y-range from all charts for that x-region and then set y-axis of all charts to y-min and y-max found out. Exactly the point 5 from https://github.com/highcharts/highcharts/pull/17875#issuecomment-1291834738.

I'll discuss internally also upon expectation from our users on other types of zooms and axis sharing combinations.

  
@hubertkozik can you help in giving us right callbacks to run, I can take from there. Thanks!

hubertkozik:
@nishikantparmariam I am sorry, but it's starting to be too complicated for supporting via GitHub and demos, I believe there are a lot of possibilities, that you are trying to handle and I don't know them as well as you. To be honest it's a really custom thing.

### SyncingZooming into area having no data points
I suggest building your logic based on global `extremes` variable shared between all possible charts. Thanks to that, there will be a possibility to calculate a correct extremes on every `afterSetExtremes` event and you will always have an information, which extremes are right to set/update. 

### Infinite zooming
Yes, you are right `xAxis.minRange` is the property, that you need. I don't think it's needed to calculate it somehow, it depends on your project, I suggest setting it is manually: https://jsfiddle.net/BlackLabel/vjLkqwte/

### Optimization
You can write a logic, where everything will set redraw flag as `false` and after all possible operations you can loop through charts and fire `redraw` for each chart.

PR #17877, opened 2022-10-23
Update stacked area

mekhatria:
The stacked area demo displayed 125 instead of 100


TorsteinHonsi (changes requested):
Thanks!

Rather than setting the `yAxis.max`, I think we should use `stacking: 'percent'` instead of `normal`. Because it is semantically true.

Other suggestions for improving the readability:
* Include the `series-label` module to have the series name printed directly on the areas. Improves readability.
* Add percent units to the y axis labels. `format: '{value} %`.
* Add percent units to the tooltip. `valueSuffix: ' %'`
* Msis => MSIS

PR #17895, opened 2022-10-27
Enabled tooltip and mouse tracking

nasin:
Hi, I turned on the tooltip and mousetracking in the new homepage animation.


TorsteinHonsi (approved):
Thanks!

PR #17896, opened 2022-10-27
bugfix/17856-heatmap-default-linecolor

### LINKED ISSUE #17856: Heatmap cell borders disappear on hover (inconsistent with documentation)

pawellysy:
Fixed #17856, line color in hover state did not work as expected in heatmap


neillrobson:
Thank you for setting this up!

This code change is very small so I'm not concerned about the commit authorship changing. However, in the future, is there anything I can do as a contributor to maintain the blame/authorship annotation?

neillrobson:
Also, unfortunately, **this patch as currently written does not fix the issue**.

The `attr.stroke` is still set even when `lineColor` is undefined.

I would greatly appreciate us refraining from merging until the patch is changed to reflect the original fix.

neillrobson:
Hello! Since it has been two weeks since any activity has occurred on this PR, I'd like to circle back and ensure things don't go stale. Is there anything I can do to help move the process along?

Thank you!

raf18seb (approved):

PR #17905, opened 2022-10-31
Refactor: Inheritance of Lollipop series

hubertkozik:
Inheritance refactor of Lollipop series, #14103.


ts/Series/Lollipop/LollipopPointOptions.d.ts
```diff
@@ -14,15 +14,22 @@
  *
  * */
 
-import type DumbbellPointOptions from '../Dumbbell/DumbbellPointOptions';
+import type ScatterPointOptions from '../Scatter/ScatterPointOptions';
+import type ColorType from '../../Core/Color/ColorType';
+import DashStyleValue from '../../Core/Renderer/DashStyleValue';
```
TorsteinHonsi, line 19: Should be `import type`?

ts/Series/Lollipop/LollipopSeries.ts
```diff
@@ -104,8 +120,30 @@ class LollipopSeries extends DumbbellSeries {
...
+                 *
+                 * @type {number}
+                 */
+                y: void 0
+            },
+            pointRange: 1,
             tooltip: {
-                pointFormat: '<span style="color:{series.color}">●</span> {series.name}: <b>{point.y}</b><br/>'
+                headerFormat: '<span style="font-size: 10px">{point.key}</span><br/>',
+                pointFormat: '<span style="color:{series.color}">\u25CF</span> {series.name}: <b>{point.y}</b>'
```
TorsteinHonsi, line 146: Seems like we're inheriting scatter series options only to resett to default line/column options... Can we extend the `Series` or `ColumnSeries.defaultOptions` instead?

ts/Series/Lollipop/LollipopPointOptions.d.ts
```diff
@@ -14,15 +14,22 @@
...
  *
  * */
 
-export interface LollipopPointOptions extends DumbbellPointOptions {
+export interface LollipopPointOptions extends ScatterPointOptions {
+    connectorColor?: ColorType;
+    connectorWidth?: number;
+    dashStyle?: DashStyleValue;
+    low?: number;
+    lowColor?: ColorType;
```
TorsteinHonsi, line 32: Why `low`? Is this a rudiment from the dumbbell inheritance chain? In that case, I think we should change it. Looking at the `highcharts/demo/lollipop/demo.js`, it makes no sense to call the value `low`, when the chart in all other aspects works like a standard column chart.

ts/Series/Lollipop/LollipopSeries.ts
TorsteinHonsi: ... Or maybe it makes more sense that the whole dumbbell inherits `Series` directly? In some ways it is more closely related, it only removes the `drawGraph` and adds the connectors?

ts/Series/Lollipop/LollipopSeries.ts
```diff
@@ -129,6 +168,31 @@ class LollipopSeries extends DumbbellSeries {
...
+     * @param {Highcharts.Series} this The series of points.
+     *
+     */
+    public drawPoints(): void {
+        let series = this,
+            pointLength = series.points.length,
+            i = 0,
+            point;
+
+        this.seriesDrawPoints.apply(series, arguments as any);
```
TorsteinHonsi, line 187: Call `this.super` so we don't need this extra class method?

ts/Series/Lollipop/LollipopSeriesOptions.d.ts
```diff
@@ -15,17 +15,23 @@
...
  *
  * */
 
-export interface LollipopSeriesOptions extends DumbbellSeriesOptions {
-    lowColor?: undefined;
+export interface LollipopSeriesOptions extends ScatterSeriesOptions {
+    connectorColor?: ColorString;
+    connectorWidth?: number;
+    groupPadding?: number;
+    lowColor?: ColorType;
```
TorsteinHonsi, line 33: Deprecate it and use the inherited [marker.fillColor](https://api.highcharts.com/highcharts/series.line.marker.fillColor) instead.

TorsteinHonsi (changes requested):
Thanks, nice work!

In summary, I think we should deprecate the `point.low` and related options in favor of `point.y` that users would expect. Then consider inheriting basic Series instead of ScatterSeries.

ts/Series/Lollipop/LollipopPointOptions.d.ts
hubertkozik: Yes, this is a rudiment from the dumbbell inheritance. Sure, it would be more clear if we remove this property and leave it only with the possibility to use the y property - just like in the column charts.

ts/Series/Lollipop/LollipopPointOptions.d.ts
hubertkozik: I was looking for similar files with that kind of importing options and we are using `import type`.

ts/Series/Lollipop/LollipopPointOptions.d.ts
hubertkozik: But should we deprecate it (same as `lowColor`) and leave function `toYData` as it is now to keep it working with `low` property?

ts/Series/Lollipop/LollipopPointOptions.d.ts
TorsteinHonsi: Yes, I mean it should be `import type DashStyleValue` too?

ts/Series/Lollipop/LollipopSeries.ts
TorsteinHonsi: We could do it in this PR too, since it's closely related?

ts/Series/Lollipop/LollipopPointOptions.d.ts
hubertkozik: Oh, didn't notice. Thank you for the clarification!

TorsteinHonsi:
Status: Pending requested changes.

ts/Series/Lollipop/LollipopSeries.ts
TorsteinHonsi: @hubertkozik I made a typo. I meant "maybe it makes more sense that the whole _lollipop_ inherits Series directly?". Sorry for that. But for the lollipop, it seems to me that this is just a common Series with no graph, and connectors connecting the marker with the threshold. And then there's some ColumnSeries logic to handle how multiple points are laid out on the same x value.

hubertkozik: Ok, got it, thank you for your clarification. I feel, that it should inherit from `ScatterSeries` because it looks like a point with a line back to the threshold. Also, I think, that it makes sense to inherit default options from `Series.defaultOptions` and fill them with other necessary options for Lollipop. What do you think about it? I have pushed new changes (with inheriting from `ScatterSeries`).

ts/Series/Lollipop/LollipopSeries.ts
```diff
@@ -119,16 +147,32 @@ class LollipopSeries extends DumbbellSeries {
...
+        const series = this,
+            pointLength = series.points.length;
 
-    public toYData(point: LollipopPoint): Array<number> {
-        return [pick(point.y, point.low)];
-    }
+        let i = 0,
+            point;
+
+        super.drawPoints.apply(series, arguments as any);
```
TorsteinHonsi, line 167: Is the `any` needed?

ts/Series/Lollipop/LollipopSeries.ts
```diff
@@ -119,16 +147,32 @@ class LollipopSeries extends DumbbellSeries {
...
 
+        // Draw connectors
+        while (i < pointLength) {
+            point = series.points[i];
+            series.drawConnector(point as any);
```
TorsteinHonsi, line 172: Can we fix the typing of `drawConnector` so we don't need to cast to `any`?

TorsteinHonsi (changes requested):
Thanks so far!

ts/Series/Lollipop/LollipopSeries.ts
```diff
@@ -54,7 +66,7 @@ const {
  * @augments Highcharts.Series
  *
  */
-class LollipopSeries extends DumbbellSeries {
+class LollipopSeries extends ScatterSeries {
```
TorsteinHonsi, line 69: Although the Lollipop resembles a scatter series, it is more related to the Series. For example, the scatter has `jitter`, which does not make sense in the Lollipop. It works, but does not make sense. I suggest we extend Series directly, and by setting `lineWidth` to 0 by default (as it currently is), it looks like intended. It is also a simpler structure to not go via the ScatterSeries when we are not actually benefiting from it.

TorsteinHonsi (changes requested):

ts/Series/Lollipop/LollipopSeries.ts
```diff
@@ -104,9 +114,27 @@ class LollipopSeries extends DumbbellSeries {
                     halo: false
                 }
             },
-            tooltip: {
-                pointFormat: '<span style="color:{series.color}">●</span> {series.name}: <b>{point.y}</b><br/>'
-            }
+            /** @ignore-option */
+            stickyTracking: false,
+            /** @ignore-option */
+            trackByArea: false,
```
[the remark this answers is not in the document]
hubertkozik: Leftover from dumbbell.

ts/Series/Lollipop/LollipopSeries.ts
```diff
@@ -119,16 +147,32 @@ class LollipopSeries extends DumbbellSeries {
     public options: LollipopSeriesOptions = void 0 as any;
     public points: Array<LollipopPoint> = void 0 as any;
 
-    /* *
+    /**
+     * Extend the arearange series' drawPoints method by applying a connector
```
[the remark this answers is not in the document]
hubertkozik: Leftover from dumbbell.

ts/Series/Lollipop/LollipopSeries.ts
hubertkozik: Okey, got it! Thank you for the amazing overall clarification!

ts/Series/Lollipop/LollipopSeries.ts
```diff
@@ -104,9 +114,27 @@ class LollipopSeries extends DumbbellSeries {
...
-            }
+            /** @ignore-option */
+            stickyTracking: false,
+            /** @ignore-option */
+            trackByArea: false,
+            /** @ignore-option */
+            fillColor: 'none',
+            /** @ignore-option */
+            lineWidth: 0,
+            crisp: false,
```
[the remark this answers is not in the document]
hubertkozik: Dumbbell leftover.

ts/Series/Lollipop/LollipopSeries.ts
```diff
@@ -104,9 +114,27 @@ class LollipopSeries extends DumbbellSeries {
...
+                align: void 0,
+                verticalAlign: void 0,
+                /**
+                 * The y position offset of the label relative to the point in
+                 * pixels.
+                 *
+                 * @type {number}
+                 */
+                y: void 0
+            },
```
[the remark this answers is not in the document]
hubertkozik: This is needed for column data labels placement logic. Actually, `y: void 0` is not necessary now, so I will remove it and leave only `align` and `vercitalAlign`.

ts/Series/Lollipop/LollipopSeries.ts
```diff
@@ -104,9 +114,27 @@ class LollipopSeries extends DumbbellSeries {
                     halo: false
                 }
             },
-            tooltip: {
-                pointFormat: '<span style="color:{series.color}">●</span> {series.name}: <b>{point.y}</b><br/>'
-            }
+            /** @ignore-option */
+            stickyTracking: false,
```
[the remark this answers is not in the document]
hubertkozik: Currently `stickyTracking` is by default set to `false` (https://jsfiddle.net/BlackLabel/wseyf86g/) Should we enable `stickyTracking`?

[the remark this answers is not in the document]
TorsteinHonsi: Yes, why was it set to false in the first place? If not there's a good reason, we should follow the defaults. In the demo you provide, it works much better when `true`.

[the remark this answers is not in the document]
hubertkozik: To be honest, I don't know why it was set to false... I agree, that it works better with enabled stickyTracking!

ts/Series/Lollipop/LollipopSeries.ts
hubertkozik: To remove `any` I had to modify `drawConnector` and `getConnectorAttribs` methods from `DumbbellSeries.ts` file. Now everything is working correctly, is that a good approach to add a `LollipopPoint` type to the `Dumbbell` series file?

TorsteinHonsi: I think that is okay - as long as we only import types, the file size is not affected. What is the best organization is another matter. Ideally they should probably both import some ConnectorSeries types or something. But your suggestion is probably fine.

TorsteinHonsi (approved):
Thanks! Approved, just waiting for the CI passing the final commit.

PR #17908, opened 2022-11-01
Fixed #17907, stuck data label on ortho map

### LINKED ISSUE #17907: Data labels on Orthographic map not hidden
#### Expected behaviour
Data labels to hide when dragged behind horizon

#### Actual behaviour
They stick

[](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/maps/demo/topojson-projection/)

#### Product version
v10.1.0

A regression since https://github.com/highcharts/highcharts/commit/d2a3cc31a05d6a9c69bf0b2852f70b65dd57dfba

TorsteinHonsi:
Fixed #17907, a regression causing data labels on the far side on an orthographic world projection to stick.

### DIFF
--- a/ts/Core/Renderer/BBoxObject.d.ts
+++ b/ts/Core/Renderer/BBoxObject.d.ts
@@ -32,16 +32,6 @@ export interface BBoxObject extends PositionObject, SizeObject {
 
 /* *
  *
- *  Declatations
- *
- * */
-
-export interface BBoxObjectWithCenter extends BBoxObject {
-    centerX?: number;
-}
-
-/* *
- *
  *  Default Export
  *
  * */
--- a/ts/Core/Renderer/BBoxObject.d.ts
+++ b/ts/Core/Renderer/BBoxObject.d.ts
@@ -24,11 +24,20 @@ import type SizeObject from './SizeObject';
  * */
 
 export interface BBoxObject extends PositionObject, SizeObject {
-    // @todo: create custom type with centerX type.
     height: number;
     width: number;
     x: number;
     y: number;
+}
+
+/* *
+ *
+ *  Declatations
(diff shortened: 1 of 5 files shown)


ts/Core/Renderer/BBoxObject.d.ts
```diff
@@ -24,13 +24,22 @@ import type SizeObject from './SizeObject';
...
 
+/* *
+ *
+ *  Declatations
+ *
+ * */
+
+export interface BBoxObjectWithCenter extends BBoxObject {
+    centerX?: number;
+}
```
pawellysy, line 41: since this option is only used by x-range series, maybe the better approach would be moving this piece of code to the x-range module, as well as the assignment of `plotX` in the `DataLabel.ts`?

pawellysy (commented):
Looks good to me, just 2 small questions :)

ts/Core/Series/DataLabel.ts
```diff
@@ -236,6 +236,7 @@ namespace DataLabel {
             visible =
                 this.visible &&
                 point.visible !== false &&
+                defined(plotX) &&
```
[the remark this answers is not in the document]
TorsteinHonsi: It's not strictly necessary for that calculation, and we do check it further down as far as I can see.

ts/Core/Renderer/BBoxObject.d.ts
TorsteinHonsi: Yes, that was exactly my thought too! I tried it this morning, but modifying `point.plotX` in X-range made one of the other unit tests to fail. So I gave up for now. But feel free to see if you can resolve this eye-sore.

PR #17909, opened 2022-11-01
Used local sidebar from demo manager in dist-examples

goransle:
Used local sidebar from demo manager in `dist-examples` Gulp task

Example `git-ignore-me.properties`:
```properties
demos.path=../highcharts-demo-manager/tmp/demo
```


TorsteinHonsi (approved):
Thanks!

TorsteinHonsi (changes requested):
... but no thanks 😬 . I didn't notice that the CI failed.

TorsteinHonsi (approved):

PR #17910, opened 2022-11-01
Refactor: Removed stroke/stroke-width workaround for older browser bug

TorsteinHonsi:
Removed stroke/stroke-width workaround for older browser bug.

If this passes, we can simplify the fix in #17834.


hubertkozik (approved):
It's great, that this workaround could be removed and everything is working correctly! I will simplify #17834.

PR #17915, opened 2022-11-02
dashboard minimal demo

pawellysy:
Added new minimal dashboard demo.


TorsteinHonsi (commented):
Nice! One question, this may be implemented already. The mouseovers/tooltips between the pie and the column are synced. Can we do that with the grid too? So that hovering a row pops up the tooltip of the charts, and hovering the charts highlights the row in the grid?

samples/highcharts/studies/dashboard-minimal/demo.css
```diff
@@ -0,0 +1,8 @@
+@import "https://code.highcharts.com/css/highcharts.css";
+@import "https://code.highcharts.com/css/datagrid.css";
+
+#datagrid {
+    height: 402px;
+    width: 800px;
```
karolkolodziej, line 6: How about stretching that table for the whole width of the dashboard cell?

samples/highcharts/studies/dashboard-minimal/demo.mjs
```diff
@@ -0,0 +1,128 @@
...
+        tableAxisMap: {
+            Food: 'x',
+            'Vitamin A': 'value'
+        },
+        chartOptions: {
+            chart: {
+                type: 'pie'
+            }
+        },
+        /*dimensions: {
```
karolkolodziej, line 67: Unnececar

samples/highcharts/studies/dashboard-minimal/demo.mjs
```diff
@@ -0,0 +1,128 @@
...
+            width: 400,
+            height: 400
+        },*/
+        events: {
+            mount: function () {
+                // call action
+                console.log('dashboard-col-0 mount event');
```
karolkolodziej, line 74: Do we need this console.log?

karolkolodziej (changes requested):
The demo needs a bit of cleaning, but overall good job!

samples/highcharts/studies/dashboard-minimal/demo.css
karolkolodziej: Also, the height is slightly too big.

samples/highcharts/studies/dashboard-minimal/demo.html
```diff
@@ -0,0 +1,167 @@
...
+
+<script src="https://code.highcharts.com/dashboard.js"></script>
+<script src="https://code.highcharts.com/highcharts.js"></script>
+<!-- <script src="https://code.highcharts.com/modules/dashboard-plugin.js"></script> -->
+<div id="container" class="highcharts-dashboard">
+</div>
+
+<div id="datagrid" ></div>
+
+<pre id="csv" style="display:none">Food,Vitamin A
```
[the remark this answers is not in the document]
pawellysy: I can't, because its messes up the parsing of the CSV. 😅

pawelfus:
> Nice! One question, this may be implemented already. The mouseovers/tooltips between the pie and the column are synced. Can we do that with the grid too? So that hovering a row pops up the tooltip of the charts, and hovering the charts highlights the row in the grid?

Unfortunately, this won't work out of the box. We started work on DataGrid Component, so it should be possible soon. The demo will be updated with this feature, related [task](https://app.asana.com/0/1200521714849758/1203324513075811/f)

TorsteinHonsi (changes requested):
We need to fix the IE8 failure somehow. We are soon going to remove support for IE8 altogether, so even ignoring the dashboards for IE8 is a valid fix, as long as it doesn't provide an error.

TorsteinHonsi:
It may be a little early to remove the IE8 tests for _all_ Highcharts. Can you see if only the dashboard samples can be ignored? Or just fix the syntax error IE8 is complaining about?

pawellysy:
We are merging this PR to `dashboard` branch, so this changes don't go to master. We figured this would be acceptable since the development here will focus mainly on Dashboard itself. Isn't it sufficient?

.circleci/config.yml
```diff
@@ -681,13 +682,14 @@ workflows:
 #          requires: ["test-Win.Chrome"]
 #          browsers: "Mac.Firefox"
 #          notify_on_failure: true
-      - test_oldies:
-          name: "test-Win.IE8"
-          requires: ["test-Win.Chrome"]
-          browsers: "Win.IE8"
-          notify_on_failure: true
+      # - test_oldies:
```
goransle, line 685: If we go down this path we should  maybe use [a filter](https://circleci.com/docs/configuration-reference/#jobfilters) here instead. Then it could not potentially stop the tests on master if it got merged there accidentally. 

But I'll also have a quick looksee at the tests.

PR #17932, opened 2022-11-04
Refactor: simpler and animatable series invert

TorsteinHonsi:
Refactored and simplified series group inversion logic. It is now simpler and more lightweight, and allows animating the transition between a non-inverted and inverted chart. 

VML is thrown under the bus.

Preview: https://highcharts.github.io/highcharts-utils/samples/#gh/72f416abb2/sample/highcharts/demo/chart-update

To do
 - [x] Check if `SVGElement.inverted` and `parentInverted` can be removed. Shadows can probably read inversion state from `rotation`.

### DIFF
--- a/samples/highcharts/blog/map-europe-electricity-price/demo.js
+++ b/samples/highcharts/blog/map-europe-electricity-price/demo.js
@@ -1,6 +1,6 @@
 (async () => {
     const topology = await fetch(
-        "https://code.highcharts.com/mapdata/custom/europe.topo.json"
+        'https://code.highcharts.com/mapdata/custom/europe.topo.json'
     ).then(response => response.json());
 
     // Prepare demo data. The data is joined to map using value of 'hc-key'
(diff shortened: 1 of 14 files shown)


pawelfus (approved):
Looks great!

Extra question: How hard would it be to animate ["polar" <-> "plain"](https://highcharts.github.io/highcharts-utils/samples/#gh/72f416abb2/sample/highcharts/demo/chart-update)?

TorsteinHonsi:
Thanks!

> How hard would it be to animate ["polar" <-> "plain"](https://highcharts.github.io/highcharts-utils/samples/#gh/72f416abb2/sample/highcharts/demo/chart-update)?

I thin that would require a different approach, [something similar to the morphing study](https://highcharts.github.io/highcharts-utils/samples/#gh/31b17d28da/sample/highcharts/studies/morph), because we would be morphing the column between different types of SVG shapes. From `rect` to `path` and back. With inverted/non-inverted it was simpler because just the group transform changes.

pawelfus:
I was curious because `inverted` <-> `polar` already works 🤔

PR #17937, opened 2022-11-04
feature/17934-tiledwebmap-series

madepiet:
Added support for [tiled web maps](https://www.highcharts.com/docs/maps/tiledwebmap-series).
___
TO DO:

- [x] Check or even mail https://www.innerbody.com/ https://www.innerbody.com/anatomy/integumentary if we can use their images as tiles on the chart UPDATE: we cannot use these images -> the `human-anatomy-tiles` graphics will be changed


ts/Series/TiledwebMap/TiledwebMapSeries.ts
```diff
@@ -0,0 +1,221 @@
...
+/**
+ * The series type
+ *
+ * @private
+ * @class
+ * @name Highcharts.seriesTypes.tiledwebmap
+ *
+ * @augments Highcharts.Series
+ */
+class TiledwebMapSeries extends MapSeries {
```
TorsteinHonsi, line 55: Better call it `TiledWebMapSeries` etc. with a capital W, since they are separate words in natural language.

TorsteinHonsi (changes requested):
This should be in Draft mode, right?

ts/Extensions/GeoJSON.ts
```diff
@@ -574,7 +574,11 @@ function geojson(
         let pointOptions: (MapPointOptions|MapPointPointOptions|undefined);
 
         if (
-            (hType === 'map' || hType === 'mapbubble') &&
+            (
+                hType === 'map' ||
+                hType === 'mapbubble' ||
+                hType === 'tiledwebmap'
```
TorsteinHonsi, line 580: I don't think the tiledwebmap should relate to GeoJSON at all. As we can see from your demo, the tiledwebmap works also when we remove the mapData.

ts/Series/TiledWebMap/TiledWebMapSeries.ts
```diff
@@ -0,0 +1,308 @@
...
+
+        const replaceVariables = (
+            url: string,
+            x: number,
+            y: number,
+            zoom: number
+        ): string => url
+            .replace('{x}', x.toString())
+            .replace('{y}', y.toString())
+            .replace('{zoom}', zoom.toString());
```
TorsteinHonsi, line 201: I've seen some implementations using `{z}` for zoom, probably we should replace that too.

samples/maps/series-tiledwebmap/mappoint-on-twm/demo.js
```diff
@@ -0,0 +1,60 @@
...
+    },
+
+    mapNavigation: {
+        enabled: true
+    },
+
+    mapView: {
+        projection: {
+            name: 'WebMercator'
+        },
```
TorsteinHonsi, line 13: Isn't this obsolete? The TWM series is now able to force the projection.

samples/maps/series-tiledwebmap/mappoint-on-twm/demo.details
```diff
@@ -0,0 +1,10 @@
+---
+name: tiledwebmap
+authors:
+  - Kamil Musialowski
+js_wrap: b
+tags:
+  - Highcharts Maps demo
+categories:
+  - Series types
```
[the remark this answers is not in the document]
kamil-musialowski: Thank you, I had no idea about it :)

samples/maps/series-tiledwebmap/mappoint-on-twm/demo.js
kamil-musialowski: I left it as an indication for the users, to tell them which projection is being used to display the map points. But of course, at the current state of the TWM it is obsolete.

ts/Series/TiledWebMap/TiledWebMapSeries.ts
```diff
@@ -438,7 +447,8 @@ addEvent(MapView, 'beforeMapViewInit', function (e: any): boolean {
         const ProviderDefinition =
             TilesProvidersRegistry[twm.provider.type],
             def = new ProviderDefinition(),
-            providerProjectionName: string = def.getProjectionName();
+            providerProjectionName: string = def.getProjectionName(),
+            { minZoom, maxZoom } = def.getMinMaxZoom();
```
TorsteinHonsi, line 451: I'm wondering if this `getMinMaxZoom` function is necessary? It's identical for all the provider definitions, so in its current form it seems like a waste of space. Without it, we could just do `{ minZoom, maxZoom } = def;`.

samples/maps/series-tiledwebmap/europe-timezones-on-twm/demo.js
```diff
@@ -0,0 +1,105 @@
...
+                        }
+                    }
+                },
+                tooltip: {
+                    headerFormat: '',
+                    pointFormat: '{point.name}: <b>{series.name}</b>'
+                },
+                states: {
+                    hover: {
+                        borderWidth: 5
```
TorsteinHonsi, line 65: With this border width, it would look good with the new `linecap` option set to `round`

samples/maps/series-tiledwebmap/roman-empire-twm/demo.js
```diff
@@ -0,0 +1,67 @@
...
+        series: [{
+            type: 'tiledwebmap',
+            provider: {
+                type: 'Stamen',
+                theme: 'Watercolor'
+            },
+            showInLegend: false
+        }, {
+            name: 'Roman empire',
+            nullColor: 'rgb(255 0 0 / 0.5)',
```
KacperMadej, line 54: This is an unusual rgb definition. Why not a standard one (with commas)?

KacperMadej (commented):
Here are some notes mostly about the demos (most of them in code):

- [x] All the demos ending with `-twm` could be reduced due to the directory containing the series name already (`maps/series-tiledwebmap/`).

samples/maps/series-tiledwebmap/europe-timezones-twm/demo.js
```diff
@@ -0,0 +1,106 @@
...
+                descriptionFormat: 'Timezone {series.name} with {series.points.length} countries.'
+            },
+            point: {
+                valueDescriptionFormat: '{point.name}.'
+            }
+        },
+
+        legend: {
+            enabled: false
+        },
```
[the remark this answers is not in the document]
hubertkozik: Yes, it is disabled, because it would need extra logic to create dots in the legend items due to `color: transparent` and set only `borderColor` for the series. Do you think legend is necessary?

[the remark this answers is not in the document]
TorsteinHonsi: I think the appropriate way to do this would be to use a `mapline` series with a `color` setting instead of a `map` series with a `borderColor` and transparent `color`.

ts/Series/TiledWebMap/TiledWebMapSeries.ts
```diff
@@ -0,0 +1,698 @@
...
+                if (provider.theme && defined(def.themes[provider.theme])) {
+                    theme = def.themes[provider.theme];
+                } else {
+                    // if nothing set take first theme
+                    theme = def.themes[Object.keys(def.themes)[0]];
+                    error(
+                        'Missing option: Tiles provider theme cannot be ' +
+                        'reached, using standard provider theme.',
+                        false
+                    );
```
kamil-musialowski, line 244: Do we want to stick with the current generic error messages, or we want to be more specific about them? 
Example:
User sets provider theme to "foo"
Do we want to display a message that contains it? Like "Tiles provider "foo" theme cannot be reached..."

ts/Series/TiledWebMap/TiledWebMapSeries.ts
```diff
@@ -0,0 +1,698 @@
...
+ * */
+
+export default TiledWebMapSeries;
+
+/* *
+ *
+ *  API options
+ *
+ * */
+
```
magdalena-gut, line 611: It would be nice to include `series.tiledwebmap.data` specification

magdalena-gut, line 611: I tried to build API docs locally but got some errors. It might be an issue on my side, but left you a comment for you to double-check.

ts/Series/TiledWebMap/TiledWebMapSeries.ts
```diff
@@ -0,0 +1,698 @@
+/* *
+ *
+ *  (c) 2010-2022
+ *
```
magdalena-gut, line 4: No authors for this series? ;)

samples/maps/demo/tiledwebmap/demo.html
```diff
@@ -0,0 +1,5 @@
+
+<script src="https://code.highcharts.com/maps/highmaps.js"></script>
+<script src="https://code.highcharts.com/maps/modules/tiledwebmap.js"></script>
+
```
magdalena-gut, line 4: `accessibility `and `exporting ` modules lost

ts/Series/TiledWebMap/TiledWebMapSeries.ts
hubertkozik: Well, there is no `data` for TiledWebMap series - only providers.

magdalena-gut: 🤦‍♀️ofc, my bad

ts/Series/TiledWebMap/TiledWebMapSeries.ts
hubertkozik: @KacperMadej can you share your thoughts about this?

ts/Series/TiledWebMap/TiledWebMapSeries.ts
hubertkozik: I double-checked that and everything is working correctly.

ts/Series/TiledWebMap/TiledWebMapSeriesOptions.d.ts
```diff
@@ -0,0 +1,44 @@
...
+ * */
+
+interface TiledWebMapSeriesOptions extends MapSeriesOptions {
+    provider?: ProviderOptions
+}
+
+interface ProviderOptions {
+    type?: string,
+    theme?: string,
+    subdomain?: string,
```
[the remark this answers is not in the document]
hubertkozik: @KacperMadej can you share your thoughts about this?

ts/Series/TiledWebMap/TiledWebMapSeries.ts
```diff
@@ -0,0 +1,698 @@
...
+        if (provider && (provider.type || provider.url)) {
+            if (provider.type && !provider.url) {
+                const ProviderDefinition =
+                    TilesProvidersRegistry[provider.type];
+
+                if (!defined(ProviderDefinition)) {
+                    error(
+                        'Provider cannot be reached.',
+                        false
+                    );
```
KacperMadej, line 219: The error message suggests that a connection attempt was made but it looks like the situation is a bit different - could be a type in the config or someone forgot to add the provider to the registry. "Tile Provider '${provider.type}' not defined in the Provider Registry." seems to be a more truthful error message in my opinion.

ts/Series/TiledWebMap/TiledWebMapSeries.ts
KacperMadej: If the reason for the failure is clear then the message should also be clear. Please add the detailed reason for the error if we are sure why it failed. When using defaults - no need to throw for that. A better message would be to throw a warning when the set theme was not defined and it fell down to the default - something like "The Provider's Theme '${provider.theme}' is not defined in the Provider definition - falling back to '${theme}'." That way at least something works but looks not as expected and checking the console will give a helpful message. The same could be done for subdomain.

ts/Series/TiledWebMap/TiledWebMapSeriesOptions.d.ts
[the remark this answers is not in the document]
KacperMadej: Do not throw an error or warning when using the defaults. See more info in the comments by the code that throws the errors.

ts/Series/TiledWebMap/TiledWebMapSeries.ts
hubertkozik: Ok, so we will stick with warnings for using defaults.

TorsteinHonsi (changes requested):
Thanks so far! 

Before I do the actual review, there are some things we need to work on to make the UX work smoothly. All of which can be seen in for example http://localhost:3030/samples/#view/maps/series-tiledwebmap/europe-timezones . 

* **Visible joints** between the tiles. Between each tile we can see that the pixels are not perfectly joined next to each other (at least on my monitor). Resulting in a visible grid. See screenshot below. I tried to debug this in one of my early studies, but I am not sure why it happens. The tiles are as far as I can see rendered 1:1 on rounded pixels. I was able to partly work around it by rendering each tile slightly bigger than 256 pixels wide and high, but that too seems like a hack.
* **Animation during zoom**. Click the Plus button to zoom in one level. While the `mapline` representation of each country animates, the underlying tiles do not.
* **Visible loading**. Related to the animation. When we zoom in, the currently visible tiles are scrapped, and it takes time (especially on a slow connection) to load the next zoom level. Instead, this is how I think it should play out. a) Start loading the new tiles, but keep the old ones. b) Start animation of both "layers" into the new position. c) When all the new tiles have loaded (or errored), fade out and destroy the old tiles. This may happen any time before, during or after the animation.

Skjermbilde 2023-02-24 kl  13 22 36

hubertkozik:
Regarding the **Visible joints** - unfortunately, I am unable to invent any better solution than this one: https://jsfiddle.net/BlackLabel/8ndmwbhx/ It is based on rounding the width and height of each tile and adding a 1 px extra to overlap tiles each other. TBH I am unable to see any misalignment in this solution, @TorsteinHonsi what do you think about this approach?

TorsteinHonsi:
> TBH I am unable to see any misalignment in this solution, @TorsteinHonsi what do you think about this approach?

@hubertkozik As per our Slack discussion, let's go fot this approach for now. If unforeseen problems occur, we can always refactor it later.

TorsteinHonsi:
The zooming animation is awesomely smooth now 👍 

In addition to the tile gaps, there's one thing we need to fix. On http://localhost:3030/samples/#view/maps/series-tiledwebmap/mappoint, I can mousewheel out seemingly indefinately. Shouldn't we have a default `minZoom` ?

hubertkozik:
Well, we don't have any minZoom for only `MapPoint` series: https://jsfiddle.net/BlackLabel/4ep3k8ch/ so TWM is working only as a background - in my opinion it is intended behaviour.

TorsteinHonsi (changes requested):
> TWM is working only as a background - in my opinion it is intended behaviour.

I think we need some default `minZoom` mechanism, we can't ship it like this. We see the same in a pure TWM setup like http://localhost:3030/samples/#view/maps/series-tiledwebmap/basic-configuration.

KacperMadej (approved):

magdalena-gut (approved):

KacperMadej:
TODO
- [ ] improve docs and demos to un-draft this PR

TorsteinHonsi:
Thanks, getting really close now! Some more nitpicking:

* In the `twm-providers` demo, the choropleth map comes in the way of the underlying twm. If I zoom in on any place inside a country, this red layer is covering the map. Besides, it is very low resolution and doesn't look good when zooming it. I think it is better to remove the choropleth map altogether. Or perhaps if we want to show that we can combine a twm with choropleth, use it to highlight a specific region of interest instead of covering every country.
* In the europe-timezones demo, is there a specific reason for not using `mapline` series for the timezone outlines? In its current state, it looks like we're hacking a `map` series (with transparent background) in order to behave like a mapline.

hubertkozik:
Regarding europe-timezones demo: in an earlier conversation, we were trying to use `MapLine` instead of `Map` series with transparent background, but unfortunately it is not so easy to achieve -> we cannot create lines based on Europe topoJSON files :/

TorsteinHonsi:
Here's a mapline map using the Europe TopoJSON source: https://jsfiddle.net/highcharts/0awoy4p8/

TorsteinHonsi:
Here's a mapline map using the Europe TopoJSON source: https://jsfiddle.net/highcharts/0awoy4p8/

hubertkozik:
Ok... Got it. So the key is to use `Highcharts.geojson` method on topology. I will add changes to the demos.

hubertkozik:
Ok... Got it. So the key is to use `Highcharts.geojson` method on topology. I will add changes to the demos.

hubertkozik:
Ok... Got it. So the key is to use `Highcharts.geojson` method on topology. I will add changes to the demos.

hubertkozik:
Ok... Got it. So the key is to use `Highcharts.geojson` method on topology. I will add changes to the demos.

ts/Series/TiledWebMap/TiledWebMapSeries.ts
```diff
@@ -0,0 +1,934 @@
...
+     * @optionparent plotOptions.tiledwebmap
+     */
+
+    public static defaultOptions: TiledWebMapSeriesOptions = merge(MapSeries.defaultOptions, {
+        states: {
+            inactive: {
+                enabled: false
+            }
+        },
+        zIndex: -1
```
TorsteinHonsi, line 110: Is there a special reason for a fixed default z index while all other series types have `undefined`? I think this may cause problems for implementers trying to stack map layers the usual way, by either defining the order in the `series` array, or by the [index option](https://api.highcharts.com/highcharts/series.line.index).

TorsteinHonsi (commented):
Just _one_ comment left now, and we're good to go.

ts/Series/TiledWebMap/TiledWebMapSeries.ts
hubertkozik: The purpose of this was to place the TWM series always below any other series. For example, if a user would declare the TWM series as second, third, the last, the TWM won't be above the previous series. Sure, we can delete this default option and let the TWM series always be first in configuration.

Demo: https://jsfiddle.net/BlackLabel/7awn2y9g/

TorsteinHonsi: Ok, that's what I figured. I still think we should delete it so that the tiledwebmap series behaves like any other series. In addition, it may be beneficial to add a paragraph in the tiledwebmap docs article, that the `series` array can be thought of like _layers_ that are stacked on top of each other, and that the ordering of the layers are determined by source order, and can be overridden through the `index` option and the `zIndex` option.

On the same note - maybe we should set [legend.reversed](https://api.highcharts.com/highcharts/legend.reversed) in the demos that currently apply map layers, like the `twm-providers` demo? Then the layering in the legend is more in line with the actual z-index.

TorsteinHonsi (approved):
Congrats on the great work! 🎉  Approving this now, but not merging until we do the next minor release. It may be that we need to do a maintenance release from the master first.

oysteinmoseng (approved):
Awesome work! I just snuck in a minor a11y tweak.

kamil-musialowski (approved):

PR #17938, opened 2022-11-04
bugfix/17933-drilldown-xaxis-labels-styling

### LINKED ISSUE #17933: Drilldown xAxis labels underline
#### Actual behaviour
After drilling down on first xAxis category "Chrome", first three `xAxis.labels` are underlined. 
This error occurs on `chart.width` < 430px.

#### Internal note
Possible regression, last working version **9.3.3**.

https://jsfiddle.net/BlackLabel/rwhp1cbj/
Click on "Chrome" and observe the first three labels after drilldown.

#### Product version
Highcharts 10.3.1
modules/drilldown 10.3.1

hubertkozik:
Fixed #17933, x axis labels had incorrect styling after drilling down.

### DIFF
--- a/samples/unit-tests/drilldown/drilldown-general/demo.js
+++ b/samples/unit-tests/drilldown/drilldown-general/demo.js
@@ -235,6 +235,58 @@ QUnit.test('Drilldown methods', function (assert) {
         `Adding drill down series through mouseover should not
         generate errors in console, #16820.`
     );
+
+    chart.drillUp();
+    chart.update({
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (changes requested):
Can we find a way to do this without modifying the core Highcharts file? Maybe by having the drilldown module reverse the changes that it did to the labels?

TorsteinHonsi (approved):
Thanks!

Issue #17928, opened 2022-11-03
Projection explorer gets cut off when embedded in the docs


KacperMadej:
#### Expected behaviour
[The projection explorer demo in the docs](https://www.highcharts.com/docs/maps/mapview-and-projection#projection) looking good like in [the full sample view](https://www.highcharts.com/samples/maps/mapview/projection-explorer).

#### Actual behaviour
Often the text is cut off so not all the info is accessible.

#### Live demo with steps to reproduce
Extremes case when no description is visible at all. The left side is the full text. The right side is the general documentation site with that text completely cut off (Insets is the next docs chapter after the embedded explorer).
![image](https://user-images.githubusercontent.com/8960408/199696572-3a898047-d059-425b-a88b-ca5be192b1b6.png)

#### Product version
Documentation (+Highcharts Maps)

hubertkozik:
Closed as completed in [#17950](https://github.com/highcharts/highcharts/pull/17950).

PR #17950, opened 2022-11-08
bugfix/17928-clipped-projection-explorer

magdalena-gut:
Scrolling enabled at projection-explorer in documentation, touch #17928


TorsteinHonsi (approved):
Thanks! Approving this now, and we'll see if we can come up with a more general solution in the docs.

PR #17951, opened 2022-11-08
enhancement/17924-chart-reflow 

### LINKED ISSUE #17924: Chart reflow on container dimension change instead of window resize
Related to https://github.com/highcharts/highcharts/issues/6707.

# Description of the feature

Right now, charts gets reflowed on window resize. Because of that, there's no automatic reflow on layout shifts. It would be nice to get reflow on changes to the dimensions of a chart's container instead.

Highcharts

Instead of listening for window resize here:

https://github.com/highcharts/highcharts/blob/561e76dc35ee953fd9bf6baa75e932b792dbf651/ts/Core/Chart/Chart.ts#L1815-L1847

We could use a [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) and listen for dimensions changes of the `chart.renderTo` element.

----

karolkolodziej:
Added `ResizeObserver` to the chart so it could adjust reflow on container size change, closes #17924.

To do: 
- [ ] Add tests

Things to discuss/investigate:
- Without that `setTimeout` the `reflow` method is also fired on the first renderer which should not take place.
- Should we make some exceptions/limitations for the boosted charts?

### DIFF
--- a/ts/Core/Chart/Chart.ts
+++ b/ts/Core/Chart/Chart.ts
@@ -349,6 +349,7 @@ class Chart {
     public pointCount: number = void 0 as any;
     public pointer: Pointer = void 0 as any;
     public reflowTimeout?: number;
+    public resizeObserver?: ResizeObserver;
     public renderer: Chart.Renderer = void 0 as any;
     public renderTo: globalThis.HTMLElement = void 0 as any;
     public series: Array<Series> = void 0 as any;
@@ -362,7 +363,6 @@ class Chart {
     public time: Time = void 0 as any;
     public title?: SVGElement;
     public titleOffset: Array<number> = void 0 as any;
-    public unbindReflow?: Function;
     public userOptions: Partial<Options> = void 0 as any;
     public xAxis: Array<AxisType> = void 0 as any;
     public yAxis: Array<AxisType> = void 0 as any;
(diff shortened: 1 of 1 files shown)


karolkolodziej:
@TorsteinHonsi Could you please check `Things to discuss/investigate` and take a stand?

TorsteinHonsi:
@karolkolodziej 

> Without that setTimeout the reflow method is also fired on the first renderer which should not take place. Most likely caused by the ResizeObserver itself and the way browser works.

I think the `setTimeout` is an acceptable solution. But it may be worthwhile spending some time to investigate _why_ the resize observer is triggered. Does it also happen on later redraws of the chart? In that case we should try to eliminate the ground cause.

> Should we make some exceptions/limitations for the boosted charts?

I don't know... Do you think we need to? Currently we have this timeout that makes sure the chart is not redrawn more often than 100 ms. I am not aware of problems with boosted charts.

karolkolodziej:
As you can read in the [Resize Observer documentation ](https://w3c.github.io/csswg-drafts/resize-observer/#intro):
> Observation will fire when watched Element is inserted/removed from DOM.

Thus we have to make sure that each element is ready and here this `setTimeout` comes in handy.

When it comes to the boost, let's leave it as it is for now.

ts/Core/Chart/Chart.ts
```diff
@@ -1744,12 +1744,11 @@ class Chart {
...
-     * container is resized without a window resize event, this must be called
-     * explicitly.
+     * Reflows the chart to its container. By default, the Resize Observer is
+     * attached to the chart's div which allows to reflows the he chart
+     * automatically to its container, as per the
+     * [chart.reflow](https://api.highcharts.com/highcharts/chart.reflow)
+     * option.
      *
      * @sample highcharts/members/chart-reflow/
      *         Resize div and reflow
```
TorsteinHonsi, line 1754: This demo now demonstrates how to do it manually in two steps. Maybe we should just remove this sample now, because things work out of the box, and we don't need to show implementers how to reflow to the container?

TorsteinHonsi:
Now only the Safari error is left before we can merge this.

TorsteinHonsi:
Regarding the failing test in Safari - I am not sure what causes it because Safari 14.1 should support resize observer. Nonetheless, support was only added to Safari in March 2020. IE11 does not support it. For v11 of Highcharts we will mainly target browsers that are up to 3 years old, with polyfills for some older browsers.

So, basically, I think we need to have the old logic as a fallback. Inline for now, and in the near future we can move the fallback to a new planned module for polyfills.

TorsteinHonsi:
Thanks so far! Safari is still failing. Which makes sense because it _should_ support ResizeObserver in that version, so the cause is different.

TorsteinHonsi (changes requested):
Please look into the Safari issue

karolkolodziej:
I investigated this issue and it looks like the `ResizeObserver` goes into an infinite loop which gives an error in Safari (in Mozilla and Chrom it looks like they are silent).

According to the [thread](https://stackoverflow.com/a/58701523/12992689), I added a workaround.

But with that tests in karma are failing because it lacks the `requestAnimationFrame` there.

TorsteinHonsi:
Thanks! I'm now trying to distill what you have found out and make the code more compact and/or more efficient. But I am not sure it all works well, so I'm running it in another branch for CI to do its thing: https://github.com/highcharts/highcharts/pull/18437

TorsteinHonsi (approved):

PR #17953, opened 2022-11-08
Stock Tools GUI - restore inactive state for indicators button

### LINKED ISSUE #16487: Closing indicator popup and changing button state.
#### Expected behaviour
Change the button state after closing the indicator popup through the `x` button.

#### Actual behaviour
The button doesn't change the state after closing the indicator popup.

#### Product version
Highstock JS v9.2.2 (2021-08-24)

#### Affected browser(s)
All

bm64:
Fixed #16487, Indicators button in Stock Tools GUI was active after popup close.


TorsteinHonsi:
Pending review by @pawelfus

pawelfus (changes requested):
Instead of adding more `indicators` related code in Popup.ts, could you solve it in a different way? To me, it seems "X"-button fires the` closePopup` event, but not the `deselectButton` event (like "add"-button does) - could you check it, please?

pawelfus (approved):
Approved, thank you!

TorsteinHonsi (approved):

PR #17954, opened 2022-11-08
Fixed #17355, map zoom and pan a11y issues.

### LINKED ISSUE #17355: Allow zoom using keyboard only accessibility
# Zoom using Keyboard only
I'm wondering if the zoom features of Highcharts Map, when enabled, allow for keyboard functionality within the zoom buttons. While I'm able to navigate to the zoom buttons using the TAB key on the keyboard, pressing the ENTER key on the keyboard upon navigating successfully to these buttons does not zoom the map in nor zoom the map out - the buttons only work with the mouse. 

# Chart Type: Highcharts Map

# POC
I'm actually using the Highcharts R wrapper Highcharter, where I started an issue already, but realize this may actually be a feature that would need to be built into Highcharts: [https://github.com/jbkunst/highcharter/issues/756](https://github.com/jbkunst/highcharter/issues/756)

oysteinmoseng:
Fixed #17355, keyboard navigation for map zooming and panning no longer worked.

### DIFF
--- a/ts/Accessibility/Accessibility.ts
+++ b/ts/Accessibility/Accessibility.ts
@@ -486,7 +486,6 @@ namespace Accessibility {
      * @private
      */
     export function compose(
-        AxisClass: typeof Axis,
         ChartClass: typeof Chart,
         LegendClass: typeof Legend,
         PointClass: typeof Point,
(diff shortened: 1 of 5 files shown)


TorsteinHonsi (approved):
Thanks!

PR #17960, opened 2022-11-09
Tools: Restore support of JSDoc tags excluding and validvalue.

### LINKED ISSUE #17919: String literal option types loosened in 10.3.0
#### Expected behaviour

A lot of string literal types in the TypeScript typings as at 10.2.1, like `stacking?: OptionsStackingValue;` or `step?: OptionsStepValue;` have been widened to `string` as of 10.3.0. This is a regression that reduces compile-time type safety, since invalid configs like `step: 'foo'` will no longer throw a compile time error.

#### Actual behaviour

Types should be as narrow as possible to maximise type safety.

N/A

#### Product version

Broken in highcharts.d.ts @ 10.3.0/10.3.1/master (see https://github.com/highcharts/highcharts-dist/blob/master/highcharts.d.ts);
previously working in highcharts.d.ts @ 10.2.1 and below (see https://github.com/highcharts/highcharts-dist/blob/v10.2.1/highcharts.d.ts) 

#### Affected browser(s)

N/A

bre1470:
Fixed #17919

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,7 @@
         "@babel/core": "^7.19.1",
         "@babel/preset-env": "^7.19.1",
         "@highcharts/highcharts-declarations-generator": "github:highcharts/highcharts-declarations-generator#v1.1.26",
-        "@highcharts/highcharts-documentation-generators": "github:highcharts/highcharts-documentation-generators#v0.6.1",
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (approved):
Thanks!

PR #17972, opened 2022-11-14
Bugfix - incorrect cursor after plotOptions update

### LINKED ISSUE #17878: Updating cursor though plotOptions does not update each series
#### Expected behaviour
When updating the series `cursor` through the `plotOptions` each series should have that property as it is with `color`.

#### Actual behaviour
When updating the series `cursor` through the `plotOptions` only the first series has updated cursor.
This is not the case when you have column and line but is when having column and scatter.

https://jsfiddle.net/BlackLabel/qofx0uhg/

#### Product version

#### Affected browser(s)
All

bm64:
Fixed #17878, column series cursor was incorrect after `plotOptions` update.


TorsteinHonsi (approved):
Thanks!

PR #17174, opened 2022-04-11
New feature: blendColors for bubble series

madepiet:
Added new feature for a `bubble` series, [blendColors](https://api.highcharts.com/highcharts/plotOptions.bubble.blendColors). See [the bubble series docs article](https://www.highcharts.com/docs/chart-and-series-types/bubble-series#blend-colors) for more info.

PR #17978, opened 2022-11-15
enhancement/13394-blendcolors-module

madepiet:
Transforming [blendColors](https://github.com/highcharts/highcharts/pull/17174) into an additional modul to a `bubble` series.


TorsteinHonsi:
The blend colors work is terminated, we will pursue interpolated heatmap instead.

PR #17958, opened 2022-11-09
Tools: Added DTS distribution for ESM masters

bre1470:
Fixed #17952, added TypeScript support to package files in `es-module/masters`.

PR #17983, opened 2022-11-16
Tools: Fixed distribution of DTS files for ESM.

bre1470:
This fix is related to PR https://github.com/highcharts/highcharts/pull/17958 and Issue https://github.com/highcharts/highcharts-angular/issues/334.


TorsteinHonsi (approved):
Thanks!

PR #17984, opened 2022-11-16
X-property of stackLabels now functions

### LINKED ISSUE #17902: StackLabels formatter in stacked waterfall chart
Formatter should have correct properties values. At this moment, it is not possible to create a simple, conditional formatter like `this.x === 3`, since console logging the `x` values throws only 0.

Open up the console and see `this.x` values.
https://jsfiddle.net/BlackLabel/p8u6be1a/

#### Product version
Highcharts + Highcharts More 10.2.1 (latest)

MarkusBarstad:
Fixed #17902, in stacked waterfalls, the `x` value was not available for the context of `stackLabels.formatter`.


TorsteinHonsi (approved):
Thanks!

PR #17987, opened 2022-11-17
enhancement-DataGridComponent

karolkolodziej:
Added DataGrid Component.

To do:

- [x] TS
- [x] Tests
   - [x] Constructor
   - [x] Events
- [x] Fix demo
- [x] Sync handler


ts/Extensions/DashboardPlugin/DataGridComponent.ts
```diff
@@ -0,0 +1,281 @@
...
+ */
+class DataGridComponent extends Component<DataGridComponent.ChartComponentEvents> {
+
+    /* *
+     *
+     *  Static properties
+     *
+     * */
+
+    public static syncHandlers = HighchartsSyncHandlers;
```
goransle, line 56: Question here might be if we should build out these existing handlers to support the datagrid, or have a separate file for the DataGridSyncHandlers 🤔

goransle:
@karolkolodziej looks like a good setup. Don't know if I have too much to add to the todo list.

ts/Extensions/DashboardPlugin/DataGridComponent.ts
karolkolodziej: Synch handler is not yet implemented but that is actually a good question! And I think it depends on what we would like to synchronize. But I guess it should be a separate file as you suggested. @bre1470 @pawellysy

pawellysy: Separate files sound better IMO

ts/Dashboard/Component/SharedComponentState.ts
```diff
@@ -230,13 +232,20 @@ Serializable<SharedComponentState, SharedComponentState.JSON> {
     }
 
     public setHoverPoint(
-        point?: SharedComponentState.PresentationHoverPointType,
-        eventDetail?: {}
+        point?: SharedComponentState.PresentationHoverPointType | HTMLElement,
+        eventDetail?: any
```
bre1470, line 236: `any` should be fixed, as an object with `isDataGrid` property is used. At least it should become `AnyRecord`.

bre1470 (changes requested):

bre1470:
Otherwise it looks good.

bre1470:
Thank you! Looks good. 👍

bre1470 (approved):

goransle (approved):
Nice!

pawellysy (approved):
Looks good to me :D

PR #17990, opened 2022-11-17
Fixed HTMLComponent overflow

bre1470:


bre1470:
Regarding the visual test result we probably need a merge from master.

ts/Dashboard/Component/HTMLComponent.ts
```diff
@@ -196,6 +196,7 @@ class HTMLComponent extends Component<HTMLComponent.HTMLComponentEvents> {
     public render(): this {
         this.emit({ type: 'beforeRender' });
         super.render(); // Fires the render event and calls load
+        this.contentElement.style.overflow = 'auto';
```
karolkolodziej, line 199: I guess the scrollbar is ok but since it is not blocking anything we should let @sebastianbochan make a final call.

karolkolodziej (approved):

TorsteinHonsi:
What's up with this PR? There are 0 changed files, yet the bundle size has changed 🤔

karolkolodziej:
PR was created by reverting changes and applying them once again which is why there are no files changed. 
We did that to make the work more transparent and easier to review for the whole team.

bre1470:
This PR is for the scrum process, because the change went into the dashboard branch without a review.

TorsteinHonsi:
So... What do I do? Merge? Delete? Pretend it never existed?

karolkolodziej:
Let's wait for @sebastianbochan, please :)

sebastianbochan:
The reason of the problem is missing CSS for images. In the CSS we have declared `width: 100%`, but we missed the height. 

So the solution is simpler.

We should use 

```
.hd-component img {
    width: 100%;
    height: 100%;
}
```

As a result, we will avoid hardcoded CSS styles.

Sorry for my late review.

bre1470:
Correct me if I am wrong, but this will not fix the overflow issue, just prevent the issue from happening for images.

sebastianbochan:
You are right!

sebastianbochan (approved):

PR #17992, opened 2022-11-17
Climate demo loading

bre1470:

PR #17993, opened 2022-11-17
DataOnDemand

bre1470:


pawellysy (commented):
Looks fine, but is there any demo/test, where we could test this functionality?

ts/Data/Stores/DataStore.ts
```diff
@@ -60,7 +60,7 @@ abstract class DataStore<TEventObject extends DataStore.Event> implements DataEv
     private static readonly registry = {} as Record<string, StoreType>;
 
     /**
-     * Regular expression to extract the store name (group 1) from the
+     * Regular expression to extract the store type (group 1) from the
```
karolkolodziej, line 63: The name of the variable probably needs to be changed to match the comment.

karolkolodziej (changes requested):

karolkolodziej (approved):

bre1470:
@pawellysy I like to use it in this PR: https://github.com/highcharts/highcharts/pull/17992

PR #17994, opened 2022-11-17
Fixed #17917 docs clarification regarding chart-style

### LINKED ISSUE #17917: Global chart styles overwritten by defaults
#### Expected behaviour
Defined global chart style `fontSize` shouldn't be overwritten by a default value (all text elements).

#### Actual behaviour
Defined global chart style `fontSize` is overwritten by a default value (all text elements).

https://jsfiddle.net/BlackLabel/odm53zwt/

#### Product version
Highcharts v10.3.1

#### Workaround
Added defined `fontSize` property into condition (Changed line 7).
https://jsfiddle.net/BlackLabel/c9spjtfr/

MarkusBarstad:
Fixed #17917, docs added which clarifies `chart.style.fontSize` functionality.

### DIFF
--- a/tools/gulptasks/dist-release.js
+++ b/tools/gulptasks/dist-release.js
@@ -181,7 +181,11 @@ function copyFiles() {
         } = folder;
         getFilesInFolder(from, true)
             .filter(path => (
-                !(path.startsWith('es-modules') && path.endsWith('.d.ts')) &&
+                (
+                    path.startsWith('es-modules/masters') ||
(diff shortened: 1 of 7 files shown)


TorsteinHonsi (changes requested):
Thanks! But I don't think we should actually change this behaviour. See my [comment in the issue](https://github.com/highcharts/highcharts/issues/17917#issuecomment-1319749816). We can pivot this PR to change the docs instead.

TorsteinHonsi (approved):

PR #17996, opened 2022-11-18
bugfix/17918-inset-points-zoom

### LINKED ISSUE #17918: Zooming to points in inset not working correctly
#### Expected behaviour
If you want to make zoom to the point, which is the inset, it will zoom into that point in the inset.

#### Actual behaviour
If you zoom to a point, which is in the inset the zoom will be incorrectly calculated and will zoom into bad region of the map.

https://jsfiddle.net/BlackLabel/6stLkbd8/
1) Click the 'Zoom to Texas' button - everything works correctly.
2) Click the 'Zoom to Alaska' button - incorrect behaviour, the point is in the inset.
2) Click the 'Zoom to Hawaii' button - incorrect behaviour, the point is in the inset.

#### Product version
Highmaps v10.3.1

kamil-musialowski:
Fixed #17918, zooming to insets in TopoJSON maps was not working.


DJTechnoo (approved):
To me it looks good!

ts/Series/Map/MapPoint.ts
```diff
@@ -220,18 +220,53 @@ class MapPoint extends ScatterSeries.prototype.pointClass {
...
-        const chart = point.series.chart;
+        const point = this as (MapPoint&MapPoint.CacheObject),
+            chart = point.series.chart,
+            mapView = chart.mapView;
+
+        let bounds = point.bounds;
+
+        if (mapView && bounds) {
+
+            const inset = isNumber(point.insetIndex) &&
```
hubertkozik, line 238: I would remove this 'enter' / new line after the if statement line. I have checked other HC files and in 99% of cases, there is no new line after the if statement 😊

hubertkozik (changes requested):
Looks great to me! I have marked only editorial cases - not relevant to code logic.

madepiet (approved):

hubertkozik (approved):
Good job! Thanks!

magdalena-gut (approved):

kamil-musialowski:
@TorsteinHonsi, as you have asked in #17918, we checked for similar code implementations in the core, but we haven't found anything - so we decided to stick with your solution as it is.

Also, I have decided to change the map in the unit test (so that it's also usable for TopoJSON insets), so please take a look if you are okay with it.

Thank you!

TorsteinHonsi (approved):
Thanks, it looks good!

Regarding the unit tests, we have a simplified map for insets at http://localhost:3030/samples/#test/unit-tests/maps/mapview-inset, but I'm okay with changing the TopoJSON test too.

PR #17999, opened 2022-11-18
docs/map-drilldown-insets

KacperMadej:
Added insets clean-up in the demo.

---

In case someone would like to use this demo for a use case with the top map having no insets the lower level insets on drill-up were not cleared.


TorsteinHonsi (approved):
Thanks!

PR #18015, opened 2022-11-22
bugfix/17753-incorrect-symbol-in-attribute

### LINKED ISSUE #17753: SVG export: Incorrect escaping for HTML within legend label
#### Expected behaviour
When `useHTML: true`, a legend label whose HTML has a `<` symbol inside an attribute value produces valid SVG when exported.

#### Actual behaviour
The `<` symbol is not escaped on the exported SVG. Instead, it appears as a literal character within an SVG attribute value.

https://jsfiddle.net/y7vtpzb4/2

1. Add the `exporting` module:
    ```<script src="https://code.highcharts.com/modules/exporting.js"></script>```
2. Write JS code:
    ```
        legend: {
            useHTML: true,
            labelFormatter: () => `<span title="&lt;">hover on me</span>`,
        },
    
        series: [{
            data: [[1, 1]]
        }]
    });
    ```
3. Click the button at the corner of the chart and export to an SVG file
4. Open the SVG file in an editor and observe invalid XML:
    > error on line 1 at column 4222: Unescaped '<' not allowed in attributes values

#### Product version
Highcharts 10.1.0, 10.2.1

#### Affected browser(s)
Google Chrome on Windows version 105.0.5195.127
Safari on Mac OS version 15.6.1 (17613.3.9.1.16)

bm64:
Fixed #17753, removed not allowed `<` symbol from attributes.

### DIFF
--- a/ts/Core/Chart/Chart.ts
+++ b/ts/Core/Chart/Chart.ts
@@ -2807,7 +2807,10 @@ class Chart {
             // Make chart behave as an image with the title as alt text
             this.renderer.boxWrapper.attr({
                 role: 'img',
-                'aria-label': (title && title.element.textContent) || ''
+                'aria-label': (
+                    (title && title.element.textContent) || ''
+                // #17753, < is not allowed in SVG attributes
+                ).replace(/</g, '&lt;')
             });
 
             if (!(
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (commented):
To do: Tests for (at least) the two given cases.

PR #18020, opened 2022-11-23
Samples: Map demo of lightning strikes

TorsteinHonsi:
Added new maps/mapbubble demo. [See preview](https://highcharts.github.io/highcharts-utils/samples/#gh/ae788dda8e/sample/maps/demo/lightning).


KacperMadej (approved):
Looks great 👍 
It would be nice to load large data in another way than inserting HTML, but this is not a blocker.

TorsteinHonsi:
> It would be nice to load large data in another way than inserting HTML

@KacperMadej We'll have something smarter with the new demo project

PR #18023, opened 2022-11-24
Regression in tooltip class names for color index

TorsteinHonsi:
Tooltip border color was lost on styled mode charts.


MarkusBarstad:
Wow, I didn't catch that erroneous CSS-naming!

What is a way we could have catched this before? I did run unit-tests in utils while working on the new unit-tests related to the new tt-functionality, but perhaps I didn't configure utils to use my local repo properly as a base for the unit-tests.

PR #18028, opened 2022-11-25
bugfix/17995-measure-control-points-position

### LINKED ISSUE #17995: Controlpoint positioner wrong behaviour on measure annotation selectType x
#### Expected behaviour
Using the measure annotation, with selectType "x", should place the controlpoints midway between y-axis.max and y-axis.min, accounting for non-zero start of y-axis.

Meaning for a chart with y-axis of 60<->10 the controlpoints should be placed at y=35 (y=dY/2+y0)

Annotation is of "measure" type and selectType is "x".

#### Actual behaviour
Controlpoints for measure annotation and selectype "x" does not account for y-axis that is not 0, and follows the formula y=dY/2 for placement. 
Meaning for a chart with y-axis of 60<->10 the controlpoints are placed at y=25
Behavior holds true for chart that stretches between negative and positive y-numbers

Example_02
Example_01

[Example using positive numbers:](https://jsfiddle.net/jobbTvilde/es7tgcak/16/)
[Example using negative to positive numbers](https://jsfiddle.net/jobbTvilde/u8vhz1sr/12/)

#### Product version
Highcharts JS v10.3.1, Annotations module &  Annotation advanced module 

Similar behaviour in 
 Highcharts JS v8.0.4, Annotations module &  Annotation advanced module

#### Affected browser(s)
Both versions tested in :
Brave v1.45.127 w/ chromium 107.0.5304.110
Chrome v107.0.5304.107 (32-bit)
Chromium browser v 88.0.4324.190

Both fiddles are opened in
Firefox v 106.0.5 (64-bit)

bm64:
Fixed #17995, measure's control points were positioned incorrectly when `yAxis.min` wasn't equal to 0.

### DIFF
--- a/samples/unit-tests/annotations/annotations-measure/demo.js
+++ b/samples/unit-tests/annotations/annotations-measure/demo.js
@@ -6,6 +6,7 @@ QUnit.test('#13664 - annotation measure on yAxis', function (assert) {
                 height: '50%'
             },
             {
+                min: 3,
                 top: '50%',
                 height: '50%'
             }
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #18032, opened 2022-11-25
bugfix/17970-heatmap-update-point-wrong-color

### LINKED ISSUE #17970: Heatmap: wrong point color (nullColor) when update value from null to number
#### Expected behaviour
The updated point's color (with all valid options like x, y, value) should be different than `nullColor`

#### Actual behaviour
The point's color is `nullColor` even though it's not a null point

https://jsfiddle.net/BlackLabel/Ld1x9op3/

#### Product version
v10.3.1 (probably never worked)

hubertkozik:
Fixed #17970, wrong color on heatmap point when updating from `null` to a valid number.


TorsteinHonsi (approved):
Thanks!

PR #18036, opened 2022-11-28
Db/move demos to dashboard dir

pawellysy:
Moved dashboard demos to samples/dashboard dir.


bre1470 (approved):
Strava seems to not work, but the rest looks good. 👍

PR #18037, opened 2022-11-28
Demos/title and subtitle modifications

fanuntez:
Aligned title, subtitle to the left on some demos. When aligning, the initial screen width was set to 370px.


TorsteinHonsi (changes requested):
Thanks!

Titles looking good now, but some of the subtitles are still centered-aligned while the main title is left-aligned. The ones in question are the ones that I did not approve at https://vrevs.highsoft.com/pr/18037/review

TorsteinHonsi (approved):
Thanks!

PR #18038, opened 2022-11-28
bugfix/17948-tooltip-reversed-stacks-position

### LINKED ISSUE #17948: Display of shared tooltip incorrect when using `reversedStacks`
#### Expected behaviour
When using a shared tooltip in a stacked bar chart that is using the `reversedStacks` option on the `yAxis`, I expect the tooltip to be displayed at the end of the bar (if there is room). This expected behavior would be similar to how the shared tooltip is displayed when not using the `reversedStacks` option.

#### Actual behaviour
The shared tooltip is displayed in the middle of the bar, making it difficult to view the data:

Screenshot 2022-11-07 at 20 29 14

For comparison, the behaviour when not using the `reversedStacks` option:

Screenshot 2022-11-07 at 20 28 38

Using the `{ legend: { reversed: true } }` as a workaround works, but I can't use that because our series are ordered by size/number of results. Something like: '1', '2', '3', 'Other'.

https://jsfiddle.net/5y6mcjp9/

#### Product version
Reproducible on Highcharts, master.

#### Affected browser(s)
I don't think the browser is of influence, but I was able to reproduce it on:
- Microsoft Edge: 107.0.1418.28 (Official build) (arm64)
- Safari: 15.6.1 (17613.3.9.1.16)

hubertkozik:
Fixed #17948, the tooltip position was bad when `yAxis.reversedStacks` was set to `false`.

### DIFF
--- a/samples/unit-tests/tooltip/shared/demo.js
+++ b/samples/unit-tests/tooltip/shared/demo.js
@@ -113,8 +113,7 @@ QUnit.test(
 );
 
 QUnit.test(
-    'Click event was called for a wrong series (#5622)',
-    function (assert) {
+    'Click event was called for a wrong series (#5622)', assert => {
         var $container = $('#container'),
             chart = $container
                 .highcharts({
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (changes requested):
Thanks!

I'm afraid the proposed fix won't work because it modifies the tooltip position of _all_ points in a stack, causing wrong position when `tooltip.shared` is not true. [Local test fiddle](https://jsfiddle.net/highcharts/0tpLxurq/).

When `reversedStacks` is true (default), what is it that makes the tooltip stick to the top point of the stack in the first place? If you find that code, you can probably modify that instead.

TorsteinHonsi (approved):

PR #18041, opened 2022-11-28
Fixed #17998, linecap not being applied to legend and dashed series

### LINKED ISSUE #17998: Linecap combined with dashStyle not working
#### Expected behaviour
`series.linecap` property should be applied when `series.dashStyle` property is declared.

#### Actual behaviour
`series.linecap` property isn't applied when `series.dashStyle` property is declared.

https://jsfiddle.net/BlackLabel/jpt2wrx7/

#### Product version
Highcharts v10.3.1

MarkusBarstad:
Fixed #17998, setting `linecap` wasn't reflected in dashed series nor `legendItem`.

When linecap is applied to a `legendItem.line` it adds to the element's width, and it overflows into the `legendItem`'s name. What is a reasonable way we could go about changing it?

- Setting a proper width for the linecapped `legendItem` [ ]


TorsteinHonsi:
> When linecap is applied to a legendItem.line it adds to the element's width, and it overflows into the legendItem's name. What is a reasonable way we could go about changing it?

In general it's not a problem before the `lineWidth` becomes very bug, but yes it's a problem then. We probably need to subtract half the stroke width from both ends of the line.

Skjermbilde 2022-11-30 kl  12 23 11

ts/Core/Legend/LegendSymbol.ts
```diff
@@ -100,22 +100,31 @@ namespace LegendSymbol {
...
+            markerOptions = options.marker,
+            symbolOffset = 0,
+            lineSizer;
 
         // Draw the line
         if (!this.chart.styledMode) {
             attr = {
-                'stroke-width': options.lineWidth || 0
+                'stroke-width': options.lineWidth || 0,
+                'stroke-linecap': (options.linecap || 'round') as any
```
TorsteinHonsi, line 111: The `any` cast is needed here because [SVGAttributes['stroke-linecap']](https://github.com/highcharts/highcharts/blob/98e732a40b8737d6199615c181adc2a3b9675a91/ts/Core/Renderer/SVG/SVGAttributes.d.ts#L99) is typed as specific strings, while `series.options.linecap` [may be a string](https://github.com/highcharts/highcharts/blob/98e732a40b8737d6199615c181adc2a3b9675a91/ts/Core/Series/SeriesOptions.d.ts#L89). If we remove the string, and also make the override in the gauge series stricter, we don't need the `any` cast.

ts/Core/Legend/LegendSymbol.ts
```diff
@@ -100,22 +100,31 @@ namespace LegendSymbol {
...
+
             if (options.dashStyle) {
                 attr.dashstyle = options.dashStyle;
             }
         }
 
+
+        lineSizer = attr['stroke-width'] ?
+            Math.min(attr['stroke-width'] / 2, symbolWidth / 2) :
+            0;
```
TorsteinHonsi: In styled mode, the [result will be 0](https://jsfiddle.net/highcharts/2gLrap5t/) because the stroke-width attribute does not exist. This is a common pattern that you can see in for example [Chart.ts](https://github.com/highcharts/highcharts/blob/98e732a40b8737d6199615c181adc2a3b9675a91/ts/Core/Chart/Chart.ts#L2145-L2171). The idea is to 
1. Generate the line first without a path definition. Just an empty `.path()`, you can add class , attr and all as usual.
2. Then read `legendItem.line.strokeWidth()` and set the path.

MarkusBarstad:
Thank you for good feedback!

The `legendItem.line` will stay limited now in non-styledMode, I had to add a limit to "stroke-width" for non-styledMode, else a big enough value would still make it overflow into the item-text.

I have yet to look at `stroke-linecap`s typing but I will do that now and try to remove the "any"-cast as you suggest!

TorsteinHonsi:
Thanks, the `.strokeWidth()` thing looks good now. The only thing remaining now is the [unwanted any cast](https://github.com/highcharts/highcharts/pull/18041/files#diff-04b2fffc3835932f513dbc957213c547ea07966c0380a435147bf70eb3f92cfaR110).

TorsteinHonsi (approved):
Thanks! I'll approve this now, but it should not be merged until the last minute before the next release, because the gigantic visual diffs will mask other differences when testing.

MarkusBarstad:
> Thanks! I'll approve this now, but it should not be merged until the last minute before the next release, because the gigantic visual diffs will mask other differences when testing.

Yeah, that makes sense x)

TorsteinHonsi:
Status: On hold until next release. Do not merge.

PR #18044, opened 2022-11-28
Rotating legendSymbols in inverted charts

### LINKED ISSUE #17880: When inverting a chart, the marker symbol is rotated
#### Expected behaviour
When inverting a chart, the marker symbol is not rotated

#### Actual behaviour
When inverting a chart, the marker symbol is not rotated

The following example shows the problem. Notice that the legend shows the triangle up icon, whereas the chart itself shows the triangle rotated 90 degrees. 

https://codepen.io/htveit/pen/oNyvKqX

#### Product version
Highcharts latest (v 10.2.1)

#### Affected browser(s)
All

MarkusBarstad:
Fixed #17880, marker symbols rotated on inverted chart

### DIFF
--- a/samples/unit-tests/series/datasorting-animation/demo.js
+++ b/samples/unit-tests/series/datasorting-animation/demo.js
@@ -227,7 +227,7 @@ QUnit.test(
             // Old point
             assert.close(
                 pointC.plotX,
-                pointC.graphic.getBBox().x,
+                pointC.graphic.getBBox(false, 0).x,
                 2,
                 'Time 0 - point has not started moving'
             );
@@ -287,8 +287,8 @@ QUnit.test(
                 );
                 // New point
                 assert.strictEqual(
-                    pointD.graphic.getBBox().x > 10 &&
-                        pointD.graphic.getBBox().x < pointD.plotX,
+                    pointD.graphic.getBBox(false, 0).x > 10 &&
+                        pointD.graphic.getBBox(false, 0).x < pointD.plotX,
                     true,
                     'Time 200 - new point has continued'
                 );
@@ -327,7 +327,7 @@ QUnit.test(
                 );
                 // New point
                 assert.close(
-                    pointD.graphic.getBBox().x,
+                    pointD.graphic.getBBox(false, 0).x,
                     pointD.plotX,
                     2,
                     'Time 600 - new point should be placed in the final position'
--- a/samples/unit-tests/series/datasorting-animation/demo.js
+++ b/samples/unit-tests/series/datasorting-animation/demo.js
@@ -227,7 +227,7 @@ QUnit.test(
             // Old point
             assert.close(
                 pointC.plotX,
-                pointC.graphic.getBBox(false, 0).x,
+                pointC.graphic.getBBox().x,
                 2,
                 'Time 0 - point has not started moving'
             );
@@ -287,8 +287,8 @@ QUnit.test(
                 );
                 // New point
                 assert.strictEqual(
-                    pointD.graphic.getBBox(false, 0).x > 10 &&
-                        pointD.graphic.getBBox(false, 0).x < pointD.plotX,
+                    pointD.graphic.getBBox().x > 10 &&
+                        pointD.graphic.getBBox().x < pointD.plotX,
                     true,
                     'Time 200 - new point has continued'
                 );
@@ -327,7 +327,7 @@ QUnit.test(
                 );
                 // New point
                 assert.close(
-                    pointD.graphic.getBBox(false, 0).x,
+                    pointD.graphic.getBBox().x,
                     pointD.plotX,
                     2,
                     'Time 600 - new point should be placed in the final position'
--- a/samples/unit-tests/series/datasorting-animation/demo.js
+++ b/samples/unit-tests/series/datasorting-animation/demo.js
@@ -110,7 +110,7 @@ QUnit.test('Data sorting animation', function (assert) {
                 true,
                 'Time 300 - label has continued'
             );
-            //dataLabel
+            // dataLabel
             assert.notEqual(
                 dataLabel.translateX,
                 initialDataLabelsPos,
@@ -132,7 +132,7 @@ QUnit.test('Data sorting animation', function (assert) {
                 targetLabelPos,
                 'Time 600 - label has landed'
             );
-            //dataLabel
+            // dataLabel
             assert.close(
                 Math.round(dataLabel.translateX),
                 Math.round(getCalculatedXPos(pointA) - dataLabel.width / 2),
@@ -226,8 +226,8 @@ QUnit.test(
             dataLabelDInitPos = dataLabelD.y;
             // Old point
             assert.close(
-                pointC.plotX,
-                pointC.graphic.getBBox(false, 0).x,
+                series.yAxis.len - pointC.plotY,
+                pointC.graphic.getBBox().x,
                 2,
                 'Time 0 - point has not started moving'
             );
@@ -268,7 +268,8 @@ QUnit.test(
             setTimeout(function () {
                 // Old point
                 assert.strictEqual(
-                    pointC.plotX - pointC.graphic.getBBox().x > 50,
+                    (series.yAxis.len - pointC.plotY) -
+                    pointC.graphic.getBBox().x > 50,
                     true,
                     'Time 200 - point should be in a process of removing animation'
                 );
@@ -287,8 +288,8 @@ QUnit.test(
                 );
                 // New point
                 assert.strictEqual(
-                    pointD.graphic.getBBox(false, 0).x > 10 &&
-                        pointD.graphic.getBBox(false, 0).x < pointD.plotX,
+                    pointD.graphic.getBBox().x > 10 &&
+                        pointD.graphic.getBBox().x < pointD.plotX,
                     true,
                     'Time 200 - new point has continued'
                 );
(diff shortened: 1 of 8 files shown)


TorsteinHonsi (changes requested):
Thanks! The OP actuallly reports that the marker is rotated in the series plot itself, while the legend is the one that is correctly not-rotated. 

The problem becomes more apparent if we use the `triangle-down` symbol. In the series markers it is not pointing down as it should: https://jsfiddle.net/highcharts/rz0hs3v7/

MarkusBarstad:
I have a fix but I am not quite sure if its at the optimal location. Should this default behavior be enabled or disabled in `styledMode`?

I also changed a unit-test which was failing me [(`samples/unit-tests/datasorting-animation`)](https://github.com/highcharts/highcharts/pull/18044/commits/8ce8441a513d76bea9af6768212248dd1fd47323#diff-a84263437a38d0666f7c3f4152fe52d4d6f37c043aac806078bf882d0ab37681) to use `getBBox()` with rotation disabled, but I realize that we should perhaps instead test for the rotated positions right?

TorsteinHonsi:
[Looks like](https://vrevs.highsoft.com/pr/18044/review) we need the `rotationOriginX` after all.

Regarding the visual regressions in the timeline and heatmap series, we need a way to say that individual series types should not rotate the markers. I think it would make sense to add these rotation attributes to the `pointAttribs` function. But in order to avoid rotating it in the Legend and other cases where the underlying group is not rotated, we must add a parameter to the function. Like `parentRotated`, `applyRotation` or something. Maybe just `flip`. Then timeline and heatmap would have `pointAttrib` functions where rotation is not applied.

TorsteinHonsi (changes requested):
So far so good 👍 

It doesn't work [with image symbols](https://jsfiddle.net/highcharts/g3z7rn6L/) though. I am wondering if we should use `pointMarkerOptions.radius || (seriesMarkerOptions as any).radius` instead of `graphic.width` to calculate the rotation offset.

TorsteinHonsi:
@MarkusBarstad Actually `plotX` and `plotY` seem to work well for symbol markers, but not so well for image markers. It's a good start though:

```js
            rotationOptions = flip && point && point.graphic ?
                {
                    rotation: 90,
                    rotationOriginY: point.plotY,
                    rotationOriginX: point.plotX
                } : {};
```

MarkusBarstad:
@TorsteinHonsi 

I think its the "translateX/Y" attributes on a point's `graphic` that only image-symboled points receive. When one makes up for their values (both -15) it seems to be positioned correctly using `plotX` and `plotY`. Trying to find a nice solution now.

TorsteinHonsi:
@MarkusBarstad The number 15 comes from the size of the image (it's 30x30), so we can't hard code it.

MarkusBarstad:
It centers nicely when I change line [nr 1500 of the SVGRenderer](https://github.com/highcharts/highcharts/blob/cd0e4f5ac925be2a5afe8bd7ee012944223d6414/ts/Core/Renderer/SVG/SVGRenderer.ts#L1500) to this `((width || 0) + (imgSize * scale)) / 2`, however, I realize this may break other stuff. 

Should we have a boolean-option of some kind in `symbolOptions` indicating whether the chart is inverted and then translate images based on that? 

I have tried to refer to `charts[ren.chartIndex].inverted` but npx won't have it, even when I preface it like `charts[ren.chartIndex] && charts[ren.chartIndex].inverted`

TorsteinHonsi:
Check out https://jsfiddle.net/highcharts/tg82dqaw/ where I try to summarize the different scenarios.

I tried this code:
```ts
            graphic = point && point.graphic,
            rotationOptions = flip && point && graphic ?
                {
                    rotation: 90,
                    rotationOriginY: (point.plotY || 0) +
                        (graphic.imgwidth || 0) / 2,
                    rotationOriginX: (point.plotX || 0) +
                        (graphic.imgheight || 0) / 2
                } : {};
```

This works for three out of four cases, but it fails for the image where the size is not explicitly set. The reason for that is that the image has to be loaded async in order to know its size, and `pointAttribs` runs before that. So you are probably on the right track by modifying the [translation part](https://github.com/highcharts/highcharts/blob/17880-when-inverting-a-chart-the-marker-symbol-is-rotated/ts/Core/Renderer/SVG/SVGRenderer.ts/#L1499-L1502). Maybe we can set origin there too?

MarkusBarstad:
> Check out https://jsfiddle.net/highcharts/tg82dqaw/ where I try to summarize the different scenarios.
> 
> I tried this code:
> 
> ```ts
>             graphic = point && point.graphic,
>             rotationOptions = flip && point && graphic ?
>                 {
>                     rotation: 90,
>                     rotationOriginY: (point.plotY || 0) +
>                         (graphic.imgwidth || 0) / 2,
>                     rotationOriginX: (point.plotX || 0) +
>                         (graphic.imgheight || 0) / 2
>                 } : {};
> ```
> 
> This works for three out of four cases, but it fails for the image where the size is not explicitly set. The reason for that is that the image has to be loaded async in order to know its size, and `pointAttribs` runs before that. So you are probably on the right track by modifying the [translation part](https://github.com/highcharts/highcharts/blob/17880-when-inverting-a-chart-the-marker-symbol-is-rotated/ts/Core/Renderer/SVG/SVGRenderer.ts/#L1499-L1502). Maybe we can set origin there too?

Alright, I will look at your fiddle and try to move rotation there!

I am by the way able to achieve correct translation for the images when: 

- in an inverted chart, setting rotation-origins to plotX and plotY, 

- and then using addition instead of subtraction , i.e.

`((width || 0) +  imgSizeAndScale) / 2`
instead of 
`((width || 0) -  imgSizeAndScale) / 2`

TorsteinHonsi:
Nice catch, that's probably because the x scale is -1 in an inverted series group. Meaning the x translation is flipped. So back to yesterday's question, to detect whether the chart is inverted, you can look at the element's parent group:

`const inverted = this.parentGroup && this.parentGroup.rotation === 90;`

MarkusBarstad:
@TorsteinHonsi 
I am having an issue with setting `rotationOrigin` closer to the image translation (in the symbol-function):

- I assume we only want to set `rotationOrigin` when chart is inverted,
- `parentGroup` only exists for image-objects during the lifetime of the function (as far as I can tell)

Should we keep `rotationOriginX/Y` in `pointAttrib`s after all? 

- We can still translate the images correctly, 
- we can set rotation once for both image and non-image markers, 
- and we can avoid setting any rotation-options for the series which should not have them (like the `heatmap` and the `timeline`)

TorsteinHonsi:
> Should we keep rotationOriginX/Y in pointAttribs after all?

Yes, whatever works! I think it makes more sense that way too, now that your inverted scale fix turns out successful.

PS: With the current state of the branch, the image with an explicit size is not correct:
Skjermbilde 2022-12-06 kl  09 18 44

PPS: We can read the `scaleX` directly instead of going the detour around `rotation`:
```ts
                        if (!alignByTranslate) {
                            const scaleX = (
                                this.parentGroup && this.parentGroup.scaleX
                            ) || 1;
                            this.translate(
                                ((width || 0) - (imgSize * scale * scaleX)) / 2,
                                ((height || 0) - (imgSize * scale)) / 2
                            );
                        }
```

MarkusBarstad:
@TorsteinHonsi 
Images are sized correctly now! But I am having an issue with the explicitly sized images not yet having `parentGroup` when we wish to check for it. 

I could be wrong but it seems like it has to do with the image-object being given properties asynchronously, and that for explicitly set images they have yet to arrive.

Trying to find out why sized and not-sized images diverge now.

MarkusBarstad:
> @MarkusBarstad When we have an explicit size for an image, we don't have to wait for `onload` to read its natural size. Therefore it is handled synchronously. That's probably why we don't yet have the `parentGroup`.
@TorsteinHonsi 
Ah I see, that makes a lot of sense! 

I have been experimenting with an additional `markerOption` or `markerAttrib` but now I am trying to await the attribute instead.

TorsteinHonsi:
@MarkusBarstad Hmm. That would work, but then we mutate the user options. It's not like we never do that elsewhere, but it is not good practice. An alternative could be to modify the following code so that the element is first added to the parent group, then the symbol name is set:

https://github.com/highcharts/highcharts/blob/c2159eb0707d6ab60d7ddd0a4026150478e3abf8/ts/Core/Series/Series.ts#L2726-L2736

TorsteinHonsi:
This works, though it doesn't look so good :/

```ts
                        point.graphic = graphic = chart.renderer
                            .symbol(
                                symbol,
                                markerAttribs.x,
                                markerAttribs.y,
                                void 0,
                                void 0,
                                hasPointMarker ?
                                    pointMarkerOptions :
                                    seriesMarkerOptions
                            )
                            .add(markerGroup)
                            .attr({
                                width: markerAttribs.width,
                                height: markerAttribs.height
                            });
```

TorsteinHonsi:
This just in: We need to make our solution adapt to updated rotation too.

https://jsfiddle.net/highcharts/tg82dqaw/

MarkusBarstad:
@TorsteinHonsi 

- Updated rotation and image-markers with explicitly set dimensions function now
- I have had to alter a unit-test so it can find the new positions for markers, it works but not quite sure if we should go about it some other way

TorsteinHonsi:
There's still a lot of redundant changes from the first attempt (`rotationOrigin`). In cases like this, when we're at a dead end and decide to go with a completely different solution, it is often better to just delete the old PR and start from scratch.

PR #18053, opened 2022-11-29
bugfix/17238-map-inverted-panning

### LINKED ISSUE #17238: Highmaps panning on Y axis is inverted when using SVG paths as basemap
#### Expected behaviour
When I drag my zoomed map upwards the map should move upwards. When I drag downwars it should move downwards.
Like on any map service, eg Google Maps or previous versions of Highmaps.

#### Actual behaviour
When I drag my zoomed map upwards the map moving downwards. Inverted to the mouse movement.

The issue is only occuring when I use SVG paths to display the map.
For me the last known working version was 8.2.2.

https://jsfiddle.net/82oyfr7k/

1. Zoom in
2. Start panning on Y axis

#### Product version
10.1.0

#### Affected browser(s)
All

kamil-musialowski:
Fixed #17238, panning on y axis was inverted in maps using custom SVG path as map data.


samples/unit-tests/maps/custom-map-panning/demo.js
```diff
@@ -0,0 +1,83 @@
+QUnit.test('Custom SVG path map panning', assert => {
+    // eslint-disable-next-line max-len
+    const customMapData = 'M853,513L853,505L851,503L855,501L853,497L851,493L847,489L845,491L841,487L843,483L843,481L841,479L843,475L841,473L841,473L841,469L841,467L841,463L839,465L839,463L843,461L841,459L843,457L839,451L845,449L853,451L855,447L853,445L855,445L859,447L859,445L861,447L867,445L871,445L867,437L869,437L873,435L879,435L883,437L881,439L881,443L881,445L883,445L885,439L889,433L893,435L891,437L897,439L901,439L901,441L905,445L903,449L905,451L907,449L911,455L907,457L913,461L913,463L917,467L923,465L925,469L933,467L935,469L939,479L943,479L955,475L955,479L953,481L953,487L957,493L959,499L957,501L949,501L947,505L943,511L939,509L937,515L935,513L933,517L929,515L925,521L927,523L925,527L925,525L921,529L919,533L913,543L905,545L901,549L899,545L889,535L891,533L889,531L885,525L883,525L879,527L879,529L879,531L877,529L875,531L873,533L871,531L869,533L869,531L867,531L865,531L859,533L853,535L849,539L847,535L847,529L849,527L847,519L851,515z';
```
DJTechnoo, line 3: This is a lot of data, maybe we can limit this to a simple square?

DJTechnoo (approved):
It all looks good! I added a comment, not necessarily something that should be changed, but only a thought.

samples/unit-tests/maps/custom-map-panning/demo.js
TorsteinHonsi: Better yet, try to reuse one of the existing tests for `path`

ts/Maps/MapView.ts
```diff
@@ -926,16 +926,35 @@ class MapView {
...
+                            mouseDownCenterProjected[1]
+                        ]);
+                    } else if (options.chart && options.chart.panning &&
+                                    options.chart.panning.type === 'y') {
+                        newCenter = this.projection.inverse([
+                            mouseDownCenterProjected[0],
+                            mouseDownCenterProjected[1] -
+                                (mouseDownY - chartY) / scale * flipFactor
+                        ]);
+                    } else {
```
TorsteinHonsi, line 950: Do we need to support individual `x` and `y` panning for a map? The OP is only reporting the `flipFactor` issue...

kamil-musialowski: While testing it, I have noticed that the `chart.panning.type` no longer works, so I've decided to also fix it here. But it's your call whether we should just fix the `flipFactor` issue, or leave the  `x` and `y` `panning.type` aswell?

magdalena-gut (approved):

kamil-musialowski:
Following your suggestions I have deleted my tests and actually find out, the we do have the exact tests for this issue. But since the old test was done using `assert.notEqual`, it was returning `true` even on wrong panning direction.

Now I'm only waiting for @TorsteinHonsi suggestion about the `chart.panning.type` fix, since from what I saw, we do not have a test for that (and it's not working on master).

hubertkozik (approved):

DJTechnoo (approved):
Fantastic!

ts/Maps/MapView.ts
TorsteinHonsi: @kamil-musialowski Thanks! I don't think we should support `x` and `y` panning in Maps. I don't think it makes much sense on a geo map. Unless people are using it, we risk adding unnecessary bloat.

kamil-musialowski: You are right. When I think about it now, it doesn't make much sense now to limit the panning in maps. In that case, should we remove it from [Highmaps API](https://api.highcharts.com/highmaps/chart.panning.type)?

madepiet (approved):

ts/Maps/MapView.ts
TorsteinHonsi: Yes we should. I assume that description is just inherited from vanilla charts.

kamil-musialowski: @TorsteinHonsi Done and ready for merge!

TorsteinHonsi (approved):

PR #18055, opened 2022-11-30
Fixed internal FAQ link

goransle:
Could alternatively remove the whole list and rely on the ToC that Docusaurus creates.


TorsteinHonsi (approved):
Thanks!

PR #18064, opened 2022-11-30
Db/csv assets

bre1470:


ts/Data/DataOnDemand.ts
```diff
@@ -117,7 +117,7 @@ class DataOnDemand {
             return this.loadStore(sourceOptions);
         }
 
-        throw new Error('Source not found');
+        throw new Error(`Source not found. (${name})`);
```
karolkolodziej, line 120: How about?
```
`Source ${name} not found.
```

karolkolodziej (approved):

ts/Data/DataOnDemand.ts
bre1470: This would make it harder to have a string comparison agains a specific error message. That's why related values are put at the end.

PR #18075, opened 2022-12-02
db/overflow hd component

karolkolodziej:


sebastianbochan:
I've tested it and looks good to me 👍 

However I would test it again, when the ticket: **Size of content-dependent components is set too early** will be ready. 

@goransle

sebastianbochan (approved):

goransle (approved):

PR #18078, opened 2022-12-02
db/data-api

bre1470:
- Changed the `DataParser`s to `DataConverter`s
- Changed the `DataStore.save` to the `DataConverter.export`, changed demos
- Merged old `DataConverter` functionality into new `DataConverter`
- Simplified column order functions of `DataStore`
- Moved `DataCSSObject` to `Dashboard/CSSJSONObject`
- Removed unused `DataValueType`

### DIFF
--- a/ts/Data/Converters/CSVConverter.ts
+++ b/ts/Data/Converters/CSVConverter.ts
@@ -103,11 +103,14 @@ class CSVConverter extends DataConverter {
     /**
      * Creates a CSV string from the datatable on the store instance.
      *
+     * @param {DataStore} store
+     * Store instance to export from.
+     *
      * @param {CSVConverter.Options} [options]
-     * The options used for the export.
+     * Options used for the export.
      *
      * @return {string}
-     * A CSV string from the table.
+     * CSV string from the store table.
      */
     public export(
         store: DataStore,
--- a/ts/Data/Converters/HTMLTableConverter.ts
+++ b/ts/Data/Converters/HTMLTableConverter.ts
@@ -119,15 +119,14 @@ class HTMLTableConverter extends DataConverter {
      * Exports the datastore as an HTML string, using the options
      * provided on import unless other options are provided.
      *
-     * @param {HTMLTableStore.ExportOptions} [htmlExportOptions]
-     * Options that override default or existing export options.
+     * @param {DataStore} store
+     * Store instance to export from.
      *
-     * @param {DataEvent.Detail} [eventDetail]
-     * Custom information for pending events.
+     * @param {HTMLTableStore.ExportOptions} [options]
+     * Options that override default or existing export options.
      *
      * @return {string}
      * HTML from the current dataTable.
-     *
      */
     public export(
         store: DataStore,
--- a/ts/Data/Stores/DataStore.ts
+++ b/ts/Data/Stores/DataStore.ts
@@ -116,7 +116,7 @@ abstract class DataStore implements DataEvent.Emitter {
      * Returns a copy of the dataStore registry as record object with
      * DataStore type and their class.
      *
-     * @return {Record<string,DataStoreRegistryType>}
+     * @return {Highcharts.Dictionary<DataStoreRegistryType>}
      * Copy of the dataStore registry.
      */
     public static getAllStores(): Record<string, StoreType> {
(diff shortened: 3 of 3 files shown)


ts/Data/Converters/CSVConverter.ts
```diff
@@ -62,40 +66,139 @@ class CSVParser extends DataParser {
...
+     * Creates a CSV string from the datatable on the store instance.
+     *
+     * @param {CSVConverter.Options} [options]
+     * The options used for the export.
+     *
+     * @return {string}
+     * A CSV string from the table.
+     */
+    public export(
+        store: DataStore,
```
karolkolodziej, line 113: param description is missing

ts/Data/Converters/CSVConverter.ts
```diff
@@ -62,40 +66,139 @@ class CSVParser extends DataParser {
...
+
+            for (let rowIndex = 0; rowIndex < columnLength; rowIndex++) {
+                let cellValue = column[rowIndex];
+
+                if (!rowArray[rowIndex]) {
+                    rowArray[rowIndex] = [];
+                }
+
+                // Prefer datatype from metadata
+                if (columnDataType === 'string') {
```
karolkolodziej, line 165: How about this: 
`if (columnDataType === 'string' || typeof cellValue === 'string') { `

Without the last `else if`?

ts/Data/Converters/HTMLTableConverter.ts
```diff
@@ -0,0 +1,570 @@
...
+     *
+     * @param {DataEvent.Detail} [eventDetail]
+     * Custom information for pending events.
+     *
+     * @return {string}
+     * HTML from the current dataTable.
+     *
+     */
+    public export(
+        store: DataStore,
```
karolkolodziej, line 133: The jsdoc is not correct

karolkolodziej (changes requested):

ts/Dashboard/Layout/Layout.ts
```diff
@@ -1,5 +1,5 @@
 /* eslint-disable */
```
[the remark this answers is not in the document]
bre1470: I do not know

ts/Data/Converters/CSVConverter.ts
bre1470: Yes, the code can be improved, but not in this PR. I just copied it over and are glad that it finally works like before. 😄 If you like, you can go over the Converters and Stores and improve it after merge.

karolkolodziej (approved):

ts/Data/Stores/DataStore.ts
```diff
@@ -342,6 +326,20 @@ abstract class DataStore implements DataEvent.Emitter {
...
+     * The default save method, which fires the `afterSave` event
+     *
+     * @return {Promise<DataStore>}
+     * The saved store.
+     *
+     * @emits DataStore#afterSave
+     * @emits DataStore#saveError
+     */
+    public save(): DataPromise<this> {
+        fireEvent(this, 'saveError', { table: this.table });
```
sebastianbochan, line 339: Why do we use saveError instead of `afterSave` event?

ts/Data/Stores/DataStore.ts
```diff
@@ -342,6 +326,20 @@ abstract class DataStore implements DataEvent.Emitter {
...
+        return DataPromise.reject(new Error('Not implemented'));
```
sebastianbochan, line 340: Im not pretty sure why it is not resolve here, but reject

ts/Data/Stores/DataStore.ts
bre1470: There is no implementation for it in the basis class. Store implementations have to overwrite it to support it without error.

ts/Data/Stores/DataStore.ts
bre1470: Because we can not keep it an abstract class. So it becomes a regular class with some implementations missing. They get eventually overwritten by Store implementations.

pawellysy (approved):
👍

ts/Data/Stores/DataStore.ts
bre1470: Sorry, it is still an abstract class, but the default should be an error, when not implemented by the Store.

PR #18090, opened 2022-12-06
Db/dashboard object in window

pawellysy:
Added dashboard object to Dashboards global namespace.

### DIFF
--- a/ts/Dashboard/Globals.ts
+++ b/ts/Dashboard/Globals.ts
@@ -11,11 +11,14 @@
  *  - Wojciech Chmiel
  *  - Gøran Slettemark
  *  - Sophie Bremer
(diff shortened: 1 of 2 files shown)


ts/Dashboard/Globals.ts
```diff
@@ -69,6 +72,8 @@ namespace Globals {
         large: 'large'
     };
 
+    export const dashboards: Array<Dashboard|undefined> = [];
```
bre1470, line 75: If we allow `undefined` for destroyed Dashboards, we need to have a destroy function setting the value at the index to `void 0`.

bre1470 (approved):
Looks good. 👍

karolkolodziej (approved):

PR #18092, opened 2022-12-06
DB: Fixed error when syncing using global scripts

goransle:


bre1470 (commented):
I get a compile issue with TypeScript 4.9.3.

The return type `typeof this.syncHandlers` of `Component.handleSyncOptions` results in
```sh
ts/Dashboard/Component/Component.ts(192,15): error TS2304: Cannot find name 'this'.
```
After some testing TypeScript became more open:
```
Private or protected member 'syncHandlers' cannot be accessed on a type parameter.
```

`Component['syncHandlers']` as return type is working for me even though it is not as accurate.

karolkolodziej (approved):

bre1470 (approved):

PR #18106, opened 2022-12-07
Website/timeline fixes

nasin:
I fixed spacing issues with the flag series and updated the revenue and employee data


TorsteinHonsi (approved):
Thanks!

The chart goes into scroll mode on mobile, is that intended? Where is it used? Should we add some smarter responsive options to make the chart look good on small screens?

nasin:
Hi, this chart is used on the About Us page. Yes, the scroll mode is intended. When I made this version during our last website redesign, we decided to use the scroll on smaller screens. Should I tweak it more for mobile?

TorsteinHonsi (approved):

PR #18112, opened 2022-12-08
Db/component content size

goransle:


ts/Dashboard/Component/Component.ts
```diff
@@ -584,8 +584,12 @@ abstract class Component<TEventObject extends Component.EventTypes = Component.E
...
         if (!this.hasLoaded) {
             this.load();
-            // Call resize to fit to the cell
-            this.resizeTo(this.parentElement);
+            // Call resize to fit to the cell. Timeout was added to make sure
+            // that each element is added to the DOM and that we have the
+            // correct dimensions.
+            setTimeout((): void => {
+                this.resizeTo(this.parentElement);
+            }, 100);
```
bre1470, line 592: If you are waiting for the DOM refresh cycle, a timeout of `0` is enough. If you need more than that, the timeout concept is not the best solution and `MutationObserver` should be used instead.

bre1470 (changes requested):

ts/Dashboard/Component/Component.ts
karolkolodziej: Solved easier than expected.

bre1470 (approved):
Looks good! 👍

pawellysy (approved):
LGTM 👍

PR #18113, opened 2022-12-08
bugfix/16601-map-drilldown-animation

### LINKED ISSUE #16601: Map drilldown animation is not working correctly
#### Expected behaviour
The animation should work like in the 9.2.2 version when drilling down and drilling up independently on drilldown level.

#### Actual behaviour
Let's discuss the map drill down with level 1, level 2 and level 3.

1. In version 9.2.2, the animation was working fine for:
- level 1 -> level 2 (drilldown)
- level 2 -> level 3 (drilldown)
- level 2 -> level 1 (drillup)
but wasn't working correctly for:
- level 3 -> level 2 (drillup)

2. Since version 9.3.0, the animation is different. It's working fine enough for:
- level 1 -> level 2 (drilldown)
- level 2 -> level 3 (drilldown)
but isn't working correctly for:
- level 2 -> level 1 (drillup)
- level 3 -> level 2 (drillup)

v9.2.2: https://jsfiddle.net/BlackLabel/o4qsu76L/
v9.3.0: https://jsfiddle.net/BlackLabel/49rjb08h/

hubertkozik:
Fixed #16601, improved map drilldown and drillup animations.
___
### To do
 - [x] Find a way to block the mouse events while drilling down/up - there cannot be any hover effects on the tooltip, legend, colorAxis etc.


KacperMadej (changes requested):
Initial review ready - comments in code.

ts/Core/Chart/MapChart.ts
```diff
@@ -198,6 +207,249 @@ namespace MapChart {
...
+        }
+
+        if (chart.ddDupes) {
+            chart.ddDupes.length = 0; // #3315
+        } // #8324
+        // Fire a once-off event after all series have been drilled up (#5158)
+        fireEvent(chart, 'drillupall');
+    }
+
+    return false; // to prevent default function from fireEvent
```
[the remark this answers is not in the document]
hubertkozik: Actually, I cannot `return true` here due to the event `afterDrillUp` and additional `redraw`, which will be called from the "default" function and it shouldn't be called.

KacperMadej (approved):
LGTM 👍

madepiet (approved):

samples/maps/demo/map-drilldown-without-async/demo.details
```diff
@@ -0,0 +1,11 @@
+---
+name: Map drilldown without async maps loading
+authors:
+  - Hubert Kozik
+requiresManualTesting: true
+js_wrap: b
+tags:
+  - Highcharts Maps demo
```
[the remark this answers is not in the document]
hubertkozik: It is added as a sample demo for `drilldown.mapZooming` property - not the main page. As the main demo, we still use the United States drilldown demo.

ts/Core/Chart/MapChart.ts
```diff
@@ -198,6 +216,242 @@ namespace MapChart {
...
+                });
+            }
+        });
+    }
+    return false; // to prevent default function from fireEvent
+});
+
+// to prevent default function from fireEvent
+addEvent(MapChart, 'midDrillUp', function (): boolean {
+    return !this.options.isMapChart;
```
TorsteinHonsi, line 357: Check for `this instanceof MapChart` instead. Then we don't have to introduce the `isMapChart` property.

ts/Core/Chart/MapChart.ts
```diff
@@ -198,6 +216,242 @@ namespace MapChart {
...
+        }
+
+        // first zoomTo then crossfade series
+        chart.disableTransform = false;
+
+        let animOptions: boolean | Partial<AnimationOptions> | undefined = {};
+
+        if (chart.options.drilldown) {
+            animOptions = merge({}, chart.options.drilldown.animation);
+        }
```
TorsteinHonsi, line 252: Use the [animObject](https://github.com/highcharts/highcharts/blob/6ed5d3ae8c8fb9f30e78fc5ef16cf664cbf75a4c/ts/Core/Animation/AnimationUtilities.ts#L86-L95) function to avoid type checking. It takes animation options and always returns an object.

ts/Series/Map/MapSeries.ts
```diff
@@ -765,7 +780,11 @@ class MapSeries extends ScatterSeries {
             */
             const scale = svgTransform.scaleX;
             const flipFactor = svgTransform.scaleY > 0 ? 1 : -1;
-            if (renderer.globalAnimation && chart.hasRendered) {
+            if (
+                renderer.globalAnimation &&
+                chart.hasRendered &&
+                !chart.disableTransform
```
TorsteinHonsi, line 786: Extra properties on the already crowded Chart object should be avoided. Can we solve this another way? Suggestions:
* Set `globalAnimation` to false so that the _if_ fails? But that may have side effects.
* Set a property on MapView instead of Chart. But in that case, `disableTransform` is confusing because of double negation (`disableTransform = false` => better `allowTransform = true`). But it is not really the transform we're disabling/allowing is it? From the code it seems like it is the transform _animation_. So if I understand this correctly, I suggest `MapView.allowTransformAnimation`. Or maybe `MapSeries.allowTransformAnimation`.

TorsteinHonsi (changes requested):
Thanks so far! 

I also fixed the glitch that I mentioned earlier, with the map zooming out slightly after drilldown. It was the padding from the US map that was not overwritten.

But still, the animation is a bit long. In the `map-drilldown` demo, it looks like now the counties (drilldown map) is faded in first, _then_ the data labels. Can we have the data labels fade in simultaneously with the drilldown series? It may have to do with the `dataLabels.defer` or `dataLabels.animation.defer` option.

ts/Series/Map/MapSeries.ts
hubertkozik: I have changed it to `MapView.allowTransformAnimation` because if it would be on series then after removing series during drill down/up the flag would be also removed.

hubertkozik:
After consultation with Kacper, we decided to first hide data labels and then animate (fade in/out) groups of series, because the data labels animation is different from the animation of `MapSeries` due to an additional plugin/code in `MapSeries.ts` file (from line 778 to 890 on this branch). Map series transform groups need to have another animation than data labels due to flickering and these animations are not consistent with each other, they have different translations during animation. So now it is really hard to animate data labels during drill down. Maybe we should just speed up the drill-down animations by using `animation.duration` property?

TorsteinHonsi:
@hubertkozik 

If you look at the fade-in of the data labels, you'll notice it also moves slightly into place. Now if you change this line:

https://github.com/highcharts/highcharts/blob/312bde021fc7d6ab497b26772c09b82a22185e10/ts/Series/Map/MapSeries.ts#L533

... to 
`this.dataLabelsGroup && this.dataLabelsGroup.show();`

... you'll see that movement is just the end of a full animation. 

Do you know why the data labels animate that way? It seems that `drawMapDataLabels` is called tree times during the drilldown. I would assume that one time would be sufficient, and that would be in the final position. I tried to dig a bit in the code to see where this is coming from. I assume that if we reduce this to one `drawMapDataLabels` call at the end of the drilldown (when the final translation is known), we can just place the labels there without animation and fade them in.

TorsteinHonsi (changes requested):
Thanks, I agree the functionality looks good now! One minor detail is that the animation of the labels could be a bit earlier, so that they fade in simultaneously with the detailed map. But that's not a big deal.

Another thing before we go on to detailed code review - there has been added quite a bit of logic now, and I think we should try to move as much as possible to the drilldown module. The `highmaps.js` file is already quite bulky since it carries all the code for Highcharts. So all events on `MapChart` and the `animateDrilldown` and similar methods of MapSeries - I think that should be moved to the drilldown module on the same pattern as Chart, PieSeries and ColumnSeries.

hubertkozik:
@TorsteinHonsi I have moved map-drilldown related code to the drilldown module - just like it was for `PieSeries` so far.

TorsteinHonsi:
@hubertkozik Before the final detailed PR review, please fix the TS declarations error.

hubertkozik:
I have consulted this TS declaration problem with Kacper and Sophie and we didn't find a quick easy solution for that. I have created a copy of the actual branch and sent it to Sophie because she would like to check it on her own next week. At this moment to check if it is a map chart I added an if statement with checking if a map view exists.

bre1470:
The issue is basically that most of Highcharts Maps becomes part of two modules (= drilldown gets bloated) and when I fix the declaration detection and order it will have duplicate properties in the namespace. This will solve itself with a composition (arguments instead of imports) but until then a workaround based on Huberts approach is necessary: https://github.com/highcharts/highcharts/commit/38cc70a91987a02e10af50b743e2ae3ab870b603

samples/maps/demo/map-drilldown-without-async/demo.js
```diff
@@ -0,0 +1,84 @@
+(async () => {
+
+    const nordicTopology = await fetch(
+        'https://code.highcharts.com/mapdata/custom/nordic-countries-core.topo.json'
+    ).then(response => response.json());
+
+    const noTopology = await fetch(
+        'https://code.highcharts.com/mapdata/countries/no/no-all.topo.json'
+    ).then(response => response.json());
```
TorsteinHonsi, line 9: A demo with only one possible drill down is a bit limited... Especially since Norway covers most of the height of the basic map in the first place 🇳🇴 . What if we add the other Nordic countries to make it a bit more interesting?

ts/Extensions/Drilldown.ts
```diff
@@ -622,7 +624,22 @@ defaultOptions.drilldown = {
              */
             y: 10
         }
-    }
+    },
+
+    /**
+     * Enable or disable the zooming into the region of clicked map point you
```
TorsteinHonsi, line 630: ... disable zooming into a region of the clicket map point ...

ts/Maps/MapView.ts
```diff
@@ -748,49 +749,56 @@ class MapView {
...
-                // Adjust westwards
-                projectedCenter[0] += Math.min(x2 - width - x, x1 - x) / scale;
-            }
-
-            // Map smaller than plot area, center it
-            if (y2 - y1 < height) {
-                projectedCenter[1] = boundsCenterProjected[1];
+                ],
+                isDrilling = this.chart.series.some((series): boolean =>
+                    !!series.isDrilling);
```
TorsteinHonsi, line 754: Do we need the `!!`?

TorsteinHonsi: Changing the type to `boolean|undefined` may be enough for TS to be happy.

TorsteinHonsi (changes requested):

ts/Extensions/Drilldown.ts
hubertkozik: If `mapZooming` is false the drilldown/drillup animations only fade in/fade out without zooming to a specific MapPoint.

samples/maps/demo/map-drilldown-without-async/demo.js
KacperMadej: @TorsteinHonsi  If the goal is to have an interesting map then it could be done as a new demo because it would be a more efficient approach to load the other maps dynamically and this demo is mainly about drilldown without async. Should this sample go in both directions? More simplified through the usage of maps that are smaller for tests and with async for all the Nordic countries for the API sample. Should the full-Nordic be also one of the main demos in the "Dynamic" category? There's already the USA map showing drill down in there.

ts/Extensions/Drilldown.ts
```diff
@@ -1071,19 +1116,27 @@ Chart.prototype.drillUp = function (): void {
             if (level.resetZoomButton) {
                 chart.resetZoomButton = level.resetZoomButton;
             }
+
+            const shouldAnimate = level.levelNumber === levelNumber &&
+                isMultipleDrillUp;
+
+            fireEvent(
+                chart,
+                'finishDrillUp',
```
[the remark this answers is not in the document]
hubertkozik: To be honest, I have removed all events - I think it is not necessary now and it is better to keep all of the logic in the correct function.

ts/Extensions/Drilldown.ts
TorsteinHonsi: Okay, we should explain that in the doclet description.

TorsteinHonsi (approved):

Issue #18070, opened 2022-12-01
Build for evergreen browsers


TorsteinHonsi:
Implicitly, remove support for IE9-11 in the built JavaScript files.

 - [x] Create a legacy path, where all JavaScript files and modules are compiled into `/es5/` on code.highcharts.com and npm etc. This legacy path will use the existing, pre-v11 setup. Consider doing this early, in a maintenance time, to give implementers time to switch before v11.
 - [x] Early warning that IE11 will stop working with Highcharts v11 unless implementers load the `/es5` path. Blog article plus notice on compliance page.
 - [x] Now build the main files for evergreen browsers. A possible target for the TypeScript compiler is to allow features that have been supported for the last two years.
 - [x] Remove the MSPointer file/logic from the main build, and add it only to es5 builds. This will also serve as a POC for adding future polyfills in the es5 build.
 - [x] Update docs. Browser compatibility page especially.

TorsteinHonsi:
Testing `gulp dist` fails because of https://github.com/highcharts/highcharts/issues/18234 . Will revisit after that is fixed.

PR #18115, opened 2022-12-08
CSS Variables

### LINKED ISSUE #18071: Use CSS variables for the highcharts.css file
We currently use SASS, but CSS variables have been available in evergreen browsers since 2017. In accordance with our JavaScript [mainly targeting evergreen browsers](https://github.com/highcharts/highcharts/issues/18070), CSS should follow up.

 - [ ] Rewrite the `highcharts.css` file and some other CSS files with CSS variables. See the SASS files for how colors are used. Test with styled mode demos. If not successful, reconsider this issue.
 - [ ] Rewrite the part of the tooling where the palette is parsed from the SASS variables and applied to TS/JS.
 - [ ] From the tooling and build system, remove all remaining SASS related stuff. 
 - [ ] Docs and demos. [Dark mode with `prefers-color-scheme`](https://www.highcharts.com/blog/products/highcharts/dark-mode-in-highcharts/) should be demonstrated.

TorsteinHonsi:
Replaced SASS files for `styledMode` with updated CSS files. Highcharts theming is now easier by supporting CSS variables as well as `prefers-color-scheme`. See [style by CSS](https://www.highcharts.com/docs/chart-design-and-style/style-by-css).

### Upgrade note
If you're using `chart.styledMode`, and your end users have [prefers-color-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) set to `dark` in their browsers, they may now be presented with dark-themed charts. The exception is if your CSS overrides the default coloring.

___

We currently use SASS, but CSS variables have been available in evergreen browsers since 2017. In accordance with our JavaScript [mainly targeting evergreen browsers](https://github.com/highcharts/highcharts/issues/18070), CSS should follow up.

Closes #18071.

 - [x] Rewrite the `highcharts.css` file and some other CSS files with CSS variables. See the SASS files for how colors are used. 
   - [x] Test with styled mode demos. ~~If not successful, reconsider this issue.~~
   - [x] Test exporting from styled mode demos. Are the variables rendered properly? (Preliminary tests show it's working.)
 - [x] Rewrite the part of the tooling where the palette is parsed from the SASS variables and applied to TS/JS.
   - [x] Internally in gulp tasks
   - [x] The assembler also has `getPalette` and `printPalette` functions, marked deprecated, that now crash. Check if they are needed.
 - [x] From the tooling and build system, remove all remaining SASS related stuff.
 - [x] Docs and demos. 
   - [x]  Search docs for mentions of SASS and `scss`, and rewrite this.
   - [x] [Dark mode with prefers-color-scheme](https://www.highcharts.com/blog/products/highcharts/dark-mode-in-highcharts/) should be demonstrated.
   - [x] Use variables in the `highcharts/css/colors` demo.
   - [x] Go over the other `/css` demos and see if using variables makes sense.
 - [x] Remove [CSS variables that refers to color variables](https://github.com/highcharts/highcharts/blob/enhancement/css-variables/css/highcharts.css/#L83-L115) in highcharts.css and theme CSS files. Those do not work when web pages dynamically override the base variable. Besides it's not needed and over-complicated.
 - [x] Consider using `hsl` colors with individual variables for hue and saturation.
 - [x] Add dark mode variables to `highcharts.css`. Consider enabling `prefers-color-scheme` by default.
 - [x] Write tutorial (docs or article) for the release. Two sections - 1) do the easy theming by data colors, neutral colors and 
highlight colors, and 2) override individual CSS rules with their own colors / styles. Done in the `custom-themes-in-styled-mode` docs article.


css/highcharts.css
```diff
@@ -1,83 +1,130 @@
...
+    /* Pressed button, color axis min color. */
+    --highcharts-highlight-color-10: #e6ebf5;
+
+    /* Indicators */
+    --highcharts-positive-color: #06b535;
+    --highcharts-negative-color: #f21313;
+}
+
+@media (prefers-color-scheme: dark) {
+    :root:not(:has(.highcharts-light)) {
```
TorsteinHonsi, line 81: No can do. `:has` is too new.

css/highcharts.css
```diff
@@ -87,110 +135,137 @@ $negative-color: #f21313; // Negative indicator color
...
 .highcharts-credits {
     cursor: pointer;
-    fill: $neutral-color-40;
+    fill: var(--highcharts-neutral-color-40);
     font-size: 0.7em;
     transition: fill 250ms, font-size 250ms;
 }
+
 .highcharts-credits:hover {
     fill: black;
```
pawelfus, line 270: Could be changed to ` --highcharts-neutral-color-100`

css/highcharts.css
```diff
@@ -819,7 +1053,7 @@ text.highcharts-drilldown-data-label,
 .highcharts-no-data text {
     font-weight: bold;
     font-size: 12px;
-    fill: $neutral-color-60;
+    fill: var(--highcharts-neutral-color-60);
 }
 
 /* Drag-panes module */
```
pawelfus, line 1059: Line no 1062 (can't comment that line directly) could also use  `--highcharts-neutral-color-100` instead of hard-coded `black`, but no strong opinion here

docs/chart-design-and-style/custom-themes-in-styled-mode.md
```diff
@@ -1,180 +1,64 @@
...
-
-Remark
-
-Sass is an extension of CSS that allows you to use variables, nested rules and inline imports in stylesheet.
-
-In the [highcharts.scss](https://github.com/highcharts/highcharts/blob/master/css/highcharts.scss) file, there are declared variables which define the most important and popular elements of charts. Use them to easy style the chart in the customs theme. As a result, there is no need to overwrite all CSS classes.
-
-[View more about SASS](https://sass-lang.com/guide)
-
-Here are the list of all available variables:
```
pawelfus, line 98: Shall we keep this legacy solution in docs? Eg under "vX-v10" paragraph or something similar?

docs/chart-design-and-style/custom-themes-in-styled-mode.md
```diff
@@ -1,180 +1,64 @@
...
-     
-    .highcharts-color-0 {
-      fill: #f45b5b;
-      stroke: #f45b5b;
+
+    /* This will re-declare variables */
+    :root {
+      --highcharts-color-0: #f45b5b;
+    }
+
```
pawelfus, line 35: How about adding this example too? So we would have:
- all charts change: eg theme
- per chart change: eg different color base for Pie charts (gradient color change?) and a Multi-line chart (different colors per series?)
- change of variables on a class/element (the example below with axis-labels)
- classic CSS change (the example with highcharts-title below)

The example:
```
/* This will overwrite variables for different charts */
#container-1 {
  --highcharts-color-0: #f45b5b;
}

#container-2 {
  --highcharts-color-0: #dedede;
}
```

pawelfus (changes requested):
Looks great! 

Note: I did not review the palette for dark mode (eg [this demo](http://utils.highcharts.local/samples/#view/highcharts/css/series-cursor) has bad contrast for line series), because I realized it's not set (and because later I remember you said that the design overhaul is coming ;) )

Extra question: Do you also plan to change some other variables? For example `opacity` or `transition`? I don't find them necessary, but `--highcharts-transition-duration: 250ms` could come handy

docs/chart-design-and-style/custom-themes-in-styled-mode.md
TorsteinHonsi: I added a Legacy section with a link to the [old docs article](https://github.com/highcharts/highcharts/blob/v10.3.3/docs/chart-design-and-style/custom-themes-in-styled-mode.md).

TorsteinHonsi:
> Do you also plan to change some other variables?

Not really... The `opacity` varies a bit between elements. A transition of 250 is quite common, but we also have 50ms and 150ms various places. So it would have to bee something like fast, normal and slow duration. But for simplicity maybe we should just stick to the colors for now.

pawelfus (approved):
Looks good, thanks!

PR #18119, opened 2022-12-09
samples/jsfiddle-2-samples

sigrunnu:
Moved bar chart.


TorsteinHonsi (changes requested):
Thanks!

As for the other PR, I think demos for the blog should be in the `blog` folder.

TorsteinHonsi (approved):

PR #18120, opened 2022-12-09
Db/remove-heat

bre1470:


karolkolodziej (changes requested):
Looks ok but I'm not sure if the color axis is necessary.
I would remove it.

karolkolodziej (approved):

TorsteinHonsi:
Some improvements for the map (besides just acting as navigation)

* Colored markers for the cities. Tie it up to the `colorAxis` and give it a color that corresponds to the current temperature. Consider `mapbubble`, or bigger markers.
* Data labels could also print the current temperature to make the map more relevant. How about a bigger marker or bubble with the temperature inside? For inspiration: https://jsfiddle.net/highcharts/v284de5y/1/
* The legend, saying "World Map" and "Cities" is stating the obvious and can be removed.
* Preferably use the default projection.
* Add some color to the land mass, but that's probably up to the designer.

pawellysy (approved):
👍

PR #18124, opened 2022-12-11
Fixed rotating markers

MarkusBarstad:
Fixed https://github.com/highcharts/highcharts/issues/17880, marker symbols were rotated on inverted chart.

- Markers don't rotate now, including for images both with and without explicit size
- Their halos follow them as expected
- I've had to alter a unit-test slightly

Sorry about the other messy branch, @TorsteinHonsi, I didn't run `git diff` properly and thought I was comparing branches when I was only comparing my commit to the `master`. 

Should I delete my other messy branch(es) in cases like these?

EDIT: I forgot to check how the timeline series is doing now on a fresh branch, will check for that tomorrow.

### To do (Torstein)
 - [x] Identify more use cases for `Point.pos`
 - [x] Rewrite the [initial clipping animation for marker group](https://github.com/highcharts/highcharts/blob/a1c439e7410e995dd949db24b4e6e708d1bca2b6/ts/Core/Series/Series.ts#L2535-L2541)
 - [x] State marker position for inverted boost


TorsteinHonsi:
> Should I delete my other messy branch(es) in cases like these?

Yes, all branches and PRs that are dead ends should be deleted.

> I forgot to check how the timeline series is doing now on a fresh branch, will check for that tomorrow.

Sure. The visual tests already discovered that and some other issues.

MarkusBarstad:
Had some trouble with `utils` not tracking my highcharts-repo so I'm just pushing some commits here to check the output of the `Highcharts review tool.`

It seems like both series and markers inversion behave (more) properly now, and you were entirely correct, it had to do with the render-groups.

TorsteinHonsi:
Morning! As you can see from the visual diff, `!this._hasPointMarkers` is not a good flag to use for this. It only tells if there are individual point markers versus series-level point markers.

Instead, better fix those series that failed in the last run. For the timeline series, it may be just to flip the individual positions in `markerAttribs` the same way as the base series, by swapping x and y. For 3d column, find out why it is rendered into the marker group in the first place, and not the `series.group` like other columns are.

> Had some trouble with utils not tracking my highcharts-repo

Make sure the two repos are side-by-side on the file system.

TorsteinHonsi:
In the current state, the OP issue is not fixed.

TorsteinHonsi:
Note to self: The clipping animation logic for the marker group handles inversion. This must be rewritten:

https://github.com/highcharts/highcharts/blob/a1c439e7410e995dd949db24b4e6e708d1bca2b6/ts/Core/Series/Series.ts#L2535-L2541

TorsteinHonsi (approved):

PR #18128, opened 2022-12-12
Docs: Added defaults of allowed AST values.

### LINKED ISSUE #16521: Docs AST
#### Expected behaviour
Add to the documentation a list of `allowedAttributes`,`allowedReferences` and `allowedTags`.

https://api.highcharts.com/class-reference/Highcharts.AST#.allowedAttributes

bre1470:
Closed #16521

### DIFF
--- a/ts/Core/Renderer/HTML/AST.ts
+++ b/ts/Core/Renderer/HTML/AST.ts
@@ -104,26 +104,15 @@ class AST {
      * potentially harmful content from the chart configuration before adding to
      * the DOM.
      *
+     * @see [Source code with default values](
+     * https://github.com/highcharts/highcharts/blob/master/ts/Core/Renderer/HTML/AST.ts#:~:text=public%20static%20allowedAttributes)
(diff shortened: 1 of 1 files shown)


TorsteinHonsi:
I know you said "code linking is fragile", and I know I said "I think this belongs in the API". But I meant the details should be kept in the API rather than the general docs.

But listing all elements is also fragile, as we add new ones from time to time. At some point in the future one of us is guaranteed to forget updating the doclet. If all browsers were like Chrome and Safari we could link directly [to the code by text search](https://github.com/highcharts/highcharts/blob/master/ts/Core/Renderer/HTML/AST.ts#:~:text=public%20static%20allowedAttributes). Do you have other ideas for a semi-robust link to the code? Or ignore Firefox?

bre1470:
Firefox users have to scroll, but otherwise it looks like the best alternative.

TorsteinHonsi (approved):

PR #18129, opened 2022-12-12
db/cypress

karolkolodziej:


goransle (approved):
Looks good! Also it's no longer super slow in Chrome for me :)

PR #18130, opened 2022-12-12
bugfix/17086-mapline-in-non-topojson 

### LINKED ISSUE #17086: mapline geometry-based data on a non-TopoJSON
#### Expected behaviour

The mapline should be rendered as it works with the TopoJSON map: https://jsfiddle.net/BlackLabel/cuqax7zk/

#### Actual behaviour

The mapline series doesn't appear: https://jsfiddle.net/BlackLabel/bwtf4arL/

kamil-musialowski:
Fixed #17086, `mapline` didn't work with lon/lat coordinates on non-TopoJSON maps

### DIFF
--- a/samples/unit-tests/maps/mappoint/demo.js
+++ b/samples/unit-tests/maps/mappoint/demo.js
@@ -60,6 +60,36 @@ QUnit.test('MapPoint with LineWidth', function (assert) {
             'The proj4 library was loaded correctly from the chart.proj4 property'
         );
 
+        const data = [
+            [-0.1275, 51.507222],
+            [-1.893611, 52.483056]
+        ];
+
+        chart.series[1].update({
+            type: 'mapline',
+            data: [{
+                geometry: {
+                    type: 'LineString',
+                    coordinates: data,
+                    color: 'red'
+                }
+            }]
+        });
+
+        const projectedData = data.map(coords => {
+            const { x, y } = chart.mapView.lonLatToProjectedUnits({
+                lon: coords[0],
+                lat: coords[1]
+            });
+            return [x, y];
+        });
+
+        assert.deepEqual(
+            chart.series[1].data[0].geometry.coordinates,
+            projectedData,
+            'Mapline series should have correct position on non TopoJSON chart.'
+        );
+
     } finally {
         window.proj4 = proj4Script;
         TestUtilities.lolexUninstall(clock);
--- a/ts/Series/Map/MapPoint.ts
+++ b/ts/Series/Map/MapPoint.ts
@@ -19,7 +19,7 @@
 import type { GeoJSONGeometryMultiPoint } from '../../Maps/GeoJSON';
 import type MapPointOptions from './MapPointOptions';
 import type MapSeries from './MapSeries';
-import type { MapBounds } from '../../Maps/MapViewOptions';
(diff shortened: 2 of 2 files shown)


TorsteinHonsi:
Hi, this is how geometries work with the legacy GeoJSON files. Not because they are GeoJSON, but because they have the `crs` specifying the projection and the `hc-transform` attribute that defines how geometries should be translated. So basically, all geometries in the legacy files are pre-projected.

The `Point` geometry is an exception because [I made it work identically](https://github.com/highcharts/highcharts/blob/b3006dfca7b2b2ddba94e8d1ae32fe28c3358abd/ts/Series/MapPoint/MapPointSeries.ts#L139-L148) to the `lon` and `lat` point options. So if there is a bug, that is it. But if no-one is reporting it, I don't think we should waste more time on supporting the legacy maps.

In the OP demo, you can see that the fix only works with `LineString`, but not `MultiLineString`, `Polygon` etc.

Please revert the added code in this PR. But the question is if we should add some heads-up in the docs under the `geometry` option?

TorsteinHonsi (changes requested):
Hi, this is how geometries work with the legacy GeoJSON files. Not because they are GeoJSON, but because they have the `crs` specifying the projection and the `hc-transform` attribute that defines how geometries should be translated. So basically, all geometries in the legacy files are pre-projected.

The `Point` geometry is an exception because [I made it work identically](https://github.com/highcharts/highcharts/blob/b3006dfca7b2b2ddba94e8d1ae32fe28c3358abd/ts/Series/MapPoint/MapPointSeries.ts#L139-L148) to the `lon` and `lat` point options. So if there is a bug, that is it. But if no-one is reporting it, I don't think we should waste more time on supporting the legacy maps.

In the OP demo, you can see that the fix only works with `LineString`, but not `MultiLineString`, `Polygon` etc.

Please revert the added code in this PR. But the question is if we should add some heads-up in the docs under the `geometry` option?

KacperMadej:
@TorsteinHonsi 

> But the question is if we should add some heads-up in the docs under the geometry option?

That would be very nice. [The current description](https://api.highcharts.com/highmaps/series.mapline.data.geometry) mentions only geoJSON (by the way, for consistent should it be GeoJSON?). Should it also contain info about the geometry of 'Point' type working differently? or is that just a small inconsistency waiting to be addressed when reported?

TorsteinHonsi:
> That would be very nice. [The current description](https://api.highcharts.com/highmaps/series.mapline.data.geometry) mentions only geoJSON (by the way, for consistent should it be GeoJSON?).

Yes, should be GeoJSON. The description is technically correct. TopoJSON is missing because it unpacks into GeoJSON, but should probably be mentioned nonetheless. 

The destinction in the behaviour is not between GeoJSON and TopoJSON per se, but between sources that are pre-projected (like our legacy GeoJSON files are), and those that are not (like our TopoJSON files are). So I guess this should be described.

> Should it also contain info about the geometry of 'Point' type working differently? or is that just a small inconsistency waiting to be addressed when reported?

I think [the description](https://api.highcharts.com/highmaps/series.mappoint.data.geometry) of the `mappoint.data.geometry` is correct, so no need to elaborate there for the time being.

kamil-musialowski:
Moved to #18148

PR #18135, opened 2022-12-13
bugfix/17925-mobile-zooming

### LINKED ISSUE #17925: World map zooms in too much with single pinch even if there a limit on zoom
#### Expected behaviour
We have disabled zoom in with pinch and added only buttons to zoom in. Technically pinch should have no effect on the chart. 

#### Actual behaviour
Zooming in with button and then pinching zooms the chart in or out so much that it all the regions disappear. It is shown in the video here: https://www.veed.io/view/b6ce7fa0-b899-4664-a524-e652442a67ef?sharingWidget=true

#### Affected browser(s)
IPAD webview

magdalena-gut:
Fixed #17925, zooming issues with Maps on mobile.


TorsteinHonsi (changes requested):
Did you manage to add tests? The test controller has functions like `touchStart`, `touchMove` and `touchEnd` that emulates a mobile.

magdalena-gut:
> Did you manage to add tests? The test controller has functions like `touchStart`, `touchMove` and `touchEnd` that emulates a mobile.

I tried but wasn't able to reproduce the issue in Utils, however, I didn't know about these functions. I'll check them, thanks!

PR #18138, opened 2022-12-13
bugfix/18116-sunburst-central-dl-width

### LINKED ISSUE #18116: Sunburst chart: central data label width is too short
#### Expected behaviour
In the central disk of a sunburst chart, the data label width should be able to almost fill up the full diameter of that disk.

#### Actual behaviour
The central data label is too narrow.

I could also not manage to set its width manually through dataLabels style options. It is not so clear to me whether [`series.sunburst.data.dataLabels.style`](https://api.highcharts.com/highcharts/series.sunburst.data.dataLabels.style) or [`series.sunburst.levels.dataLabels.style`](https://api.highcharts.com/highcharts/series.sunburst.levels.dataLabels.style) should be able to override that width.

Here is a screenshot from the demo. Clearly, there is enough space to display "the world", but the label is constrained to roughly half of the available width:
image

From looking at the [code](https://github.com/highcharts/highcharts/blob/v10.3.2/ts/Series/Sunburst/SunburstSeries.ts#L121), the cause appears to be that rendering for this central data label follows the same logic as other non-central points. Though while the children compute their label width in reference to the radius of the point shape, the central data label can use *twice* that radius, since it is a full circle after all.

https://jsfiddle.net/uj43fnp2/

#### Product version
Highcharts 10.3.2

#### Affected browser(s)
Firefox, Chrome.

bm64:
Fixed #18116, the central data label in sunburst series was too short.

### DIFF
--- a/ts/Series/Sunburst/SunburstSeries.ts
+++ b/ts/Series/Sunburst/SunburstSeries.ts
@@ -55,6 +55,7 @@ const {
 import U from '../../Core/Utilities.js';
 import SunburstNode from './SunburstNode.js';
 const {
+    defined,
     error,
     extend,
     isNumber,
@@ -199,7 +200,12 @@ function getDlOptions(
                 ((point.outerArcLength as any) + point.innerArcLength) / 2
             );
         } else {
-            (options.style as any).width = shape.radius;
+            if (
+                !defined((options.style as any).width) &&
+                shape.radius
+            ) {
+                (options.style as any).width = 2 * shape.radius;
+            }
         }
 
         if (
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (changes requested):
See visual diffs. Just a tip, you can discover visual changes in the sample you're working on before you push to GitHub. Use the side-by-side feature in the utils to compare to the current stable release.

Skjermbilde 2022-12-13 kl  16 37 08

TorsteinHonsi (approved):

PR #18139, opened 2022-12-13
db/series points modifier

karolkolodziej:
Removed SeriesPointsModifier.


bre1470 (approved):
Thank you! Looks right. 👍

PR #18144, opened 2022-12-14
feature/mapmarker

TorsteinHonsi:
Added new symbol definition, `mapmarker`, to use with [map point series](https://www.highcharts.com/samples/maps/series-mappoint/mapmarker) or [data labels](https://www.highcharts.com/samples/maps/series-mappoint/datalabels-mapmarker).
___

The Internet leaves no doubt about [how it expects a map marker to look](https://www.google.com/search?q=map+marker+svg&source=lnms&tbm=isch&sa=X&ved=2ahUKEwjSsZmx2fj7AhXFXfEDHfrWAiQQ_AUoAXoECAIQAw&biw=1663&bih=1000&dpr=2).

Note: The links in the description is for the changelog and do not work until release. See the two new demos in the utils for previews.

### To do
 - [x] In the legend, the mapmarker should be correctly aligned vertically (on the map, the pin is in the center)
 - [ ] Off topic: map navigation should `alignTo` spacing box by default, not plot box.
 - [ ] Off topic: Map borders should have `stroke-linejoin: round`.
 - [x] Consider using the two samples in demos


TorsteinHonsi:
@KacperMadej Good point. So the `dataLabel` approach is better for this use case because it treats the marker and the text as one. Even with `allowOverlapping` set to true, it works well because the markers including the text are hidden behind each other.

So the question is - maybe this is not the best data set for showing `marker.symbol` at all? I am thinking that it would be better for cases where we don't have a value to display. Maybe we should just add mapmarkers to the [marker-cluster main demo](https://www.highcharts.com/demo/maps/marker-clusters)? Or come up with some other categorized data that doesn't require a value label. What do you think?

KacperMadej:
>  `allowOverlapping` set to true

1. This works very well for the dataLabels approach IMO.

2. The current marker clusters demo could look like this:
image
Used additional config:
```js

                color: Highcharts.getOptions().colors[0],
                marker: {
                    lineWidth: 1,
                    lineColor: '#000',
                    symbol: 'mapmarker'
                },
                dataLabels: {
                    y: -10
                },
```

3. Maybe with v11 default `marker.symbol` for `mappoint` could be `'mapmarker'`? It would require some additional changes as the legend icon and default dataLabels placement is a bit off:
image

Alternatively to the new default, some other demos could be adjusted via API options to use `'mapmarker'`, but it looks like [the lat/lon](https://www.highcharts.com/demo/maps/mappoint-latlon) is the only one that would use it.

TorsteinHonsi:
1. Agreed, the markers come out stacked.
2. Yes I think we should use this 
3. Agreed. I think these markers are the expected rendering of a map point. We can make the default data label alignment left-aligned, and the `radius` a bit larger.

PR #18150, opened 2022-12-14
bugfix/18145-grey-series-boost

### LINKED ISSUE #18145: Cannot render 50 or more series with specified colors and enabled boost
As suggested by [michal.f](https://www.highcharts.com/forum/memberlist.php?mode=viewprofile&u=40740&sid=d278d440396aaa4a62ceb26bcd979a73) in forum [Topic#50030](https://www.highcharts.com/forum/viewtopic.php?f=9&t=50030), I'm logging this bug issue. 

Chart points are not rendered when:
1. Boost module is enabled.
2. Color is specified in input.
3. Number of series is greater than or equal to 50.

Our business-use-case involves displaying more than 50 series with specific colors. Can you please fix the bug / give us some workaround / tell me if I'm doing anything wrong here?

#### Expected behaviour
All series with data points with "grey" color should be correctly rendered.

#### Actual behaviour
None of the data-points are rendered.

https://jsfiddle.net/bqghn513/

#### Product version
Highcharts version `10.3.2`

#### Affected browser(s)
Chrome

bm64:
Fixed #18145, boost module couldn't render series with colour set to grey.

### DIFF
--- a/ts/Extensions/Boost/NamedColors.ts
+++ b/ts/Extensions/Boost/NamedColors.ts
@@ -80,6 +80,7 @@ const defaultHTMLColorMap: Record<string, ColorString> = {
     gold: '#ffd700',
     goldenrod: '#daa520',
     gray: '#808080',
+    grey: '#808080',
     green: '#008000',
     greenyellow: '#adff2f',
     honeydew: '#f0fff0',


TorsteinHonsi (approved):
Thanks!

PR #18155, opened 2022-12-15
Samples/jsfiddle-2-samples-2

sigrunnu:
Move jsfiddle to samples


samples/highcharts/website/histogram-symmetrical-distribution/demo.html
```diff
@@ -0,0 +1,211 @@
+<script src="https://code.highcharts.com/highcharts.js"></script>
+<script src="https://code.highcharts.com/highcharts-more.js"></script>
+<script src="https://code.highcharts.com/modules/exporting.js"></script>
+<script src="https://code.highcharts.com/modules/accessibility.js"></script>
+
+<div id="container" style="width:350px;height:250px;margin:1.5em 1em;"></div>
+
+<script>
```
TorsteinHonsi, line 8: Usually we put JS code in `demo.js`... Is there a particular reason for putting it in `demo.html`?

samples/highcharts/website/map-resellers/demo.js
```diff
@@ -0,0 +1,923 @@
...
+        value: 213
+    },
+    {
+        'hc-key': 'kg',
+        value: 214
+    }
+];
+
+function infoFunction() {
+    var $div = $('<div></div>')
```
TorsteinHonsi, line 867: jQuery is missing...

TorsteinHonsi (changes requested):
Thanks!

* These demos are mainly used by the blog, right? Then they should be added to `samples/highcharts/blog`. The `website` folder is more intended for the website design element.
* Data URLs like `https://cdn.jsdelivr.net/gh/highcharts/highcharts/samples/data/worldbank.json` don't work yet, but they will after the next HC release. If we want them to work before that, we can specify the commit hash like `https://cdn.jsdelivr.net/gh/highcharts/highcharts@d8e5ad7eb774e4a6c799b6bb0915eb1261593047/samples/data/worldbank.json`.

samples/highcharts/website/histogram-symmetrical-distribution/demo.html
sigrunnu: No, I don't know why it is in the HTML file...I can move it to `demo.js`

sigrunnu: All the demos are used in the blog so I'll move them to /blog.

TorsteinHonsi (approved):

PR #18165, opened 2022-12-19
db/data series converter

karolkolodziej:
Moved DataSeriesConverter to the core.

### DIFF
--- a/ts/Core/DataSeriesConverter.ts
+++ b/ts/Core/DataSeriesConverter.ts
@@ -21,13 +21,13 @@
  *
  * */
 
-import type DataEvent from './DataEvent';
+import type DataEvent from '../Data/DataEvent';
 import type LineSeries from '../Series/Line/LineSeries';
 import type PointOptions from '../Core/Series/PointOptions';
 import type SeriesOptions from '../Core/Series/SeriesOptions';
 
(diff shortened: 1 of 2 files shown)


ts/Core/DataSeriesConverter.ts
```diff
@@ -21,13 +21,13 @@
...
-import type DataEvent from './DataEvent';
+import type DataEvent from '../Data/DataEvent';
 import type LineSeries from '../Series/Line/LineSeries';
 import type PointOptions from '../Core/Series/PointOptions';
 import type SeriesOptions from '../Core/Series/SeriesOptions';
 
-import DataTable from './DataTable.js';
-import U from '../Core/Utilities.js';
+import DataTable from '../Data/DataTable.js';
+import U from './Utilities.js';
```
bre1470, line 30: The converter should be placed in `ts/Core/Series/`.

bre1470 (changes requested):

pawellysy (commented):
Code looks good. Just one question:
Are there any other files, that reference this file?

bre1470:
Probably not. It will eventually be merged with the SeriesTableComposition.

bre1470 (approved):
Looks good. 👍

PR #18179, opened 2022-12-20
bugfix/17314-align-threshold-and-stacking

### LINKED ISSUE #17314: alignThreshold + stacking doesn't work
#### Expected behaviour
The series should be rendered in the correct place.

#### Actual behaviour
The Series are behaving as if they would be stacking even though they are on separate axes.

https://jsfiddle.net/BlackLabel/9b1p5vkn/169/

#### Product version
v10.1.0
#### Affected browser(s)
all

#### Workaround

call `chart.redraw()` in the load event: 
https://jsfiddle.net/BlackLabel/9b1p5vkn/170/

hubertkozik:
Fixed #17314, `chart.alignThresholds` and stacking were not working correctly together.


TorsteinHonsi (approved):
Thanks!

PR #18180, opened 2022-12-20
bugfix/18088-heikin-ashi-wrong-color

### LINKED ISSUE #18088: Heikin-Ashi wrong candles color
#### Expected behaviour
Candles with close lower than open should be colored as red (or any "down color").

#### Actual behaviour
Some candles have their color mixed.

Take a look at the Nov 7 candle
https://jsfiddle.net/BlackLabel/Lj5whnuc/

#### Product version
Highstock + Heikin-Ashi 10.3.2 (latest)

kamil-musialowski:
Fixed #18088, color for Heikin-Ashi candles was set incorrectly.

### DIFF
--- a/ts/Series/HeikinAshi/HeikinAshiPoint.ts
+++ b/ts/Series/HeikinAshi/HeikinAshiPoint.ts
@@ -24,6 +24,11 @@ const {
         prototype: {
             pointClass: CandlestickPoint
         }
+    },
+    hlc: {
+        prototype: {
+            pointClass: HLCPoint
+        }
     }
 } = SeriesRegistry.seriesTypes;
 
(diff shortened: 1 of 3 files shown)


ts/Series/HeikinAshi/HeikinAshiPoint.ts
```diff
@@ -35,6 +40,9 @@ const {
 
 class HeikinAshiPoint extends CandlestickPoint {
     // clone inheritence
+    public resolveColor(): void {
+        HLCPoint.prototype.resolveColor.call(this);
+    }
```
TorsteinHonsi, line 45: Instead of wrapping and calling the nested function with the same scope, you can just borrow it:

```ts
class HeikinAshiPoint extends CandlestickPoint {
    // clone inheritence
    public resolveColor = HLCPoint.prototype.resolveColor;
}
```

TorsteinHonsi (changes requested):
Thanks! One minor change request.

TorsteinHonsi (approved):

PR #18188, opened 2022-12-21
bugfix/18168-legend-proximate-layout

### LINKED ISSUE #18168: Legend item at incorrect position when layout set to "proximate" and "startOnTicks" set to false
#### Expected behaviour

I've defined a chart that would not `startOnTicks` (a.k.a. set to `false`) and with the legend layout set to `proximate`.
The legend item corresponding to the maximum value of all should be drawn on the same level as the first point of the series.
image

#### Actual behaviour

The legend item is way off (rather drawn at the top of the chart).
The behavior seems to be a result of the current implementation of the [`proximatePositions` method in line 1057](https://github.com/highcharts/highcharts/blob/4fd6ea3d81c5821912d2c86a550af6265b4b922f/ts/Core/Legend/Legend.ts#L1057).
Instead of `if (box.pos)` the code should rather ask `if (isNumber(box.pos))` as it would otherwise leave a box with the `pos` value of `0` unhandled.
image

Might be linked to https://github.com/highcharts/highcharts/issues/14809#issuecomment-746330827.

- Live demo of the bug: https://jsfiddle.net/ohaztj64/
- Live demo with the fix included: https://jsfiddle.net/jz8vya3h/4/

#### Product version
Latest

#### Affected browser(s)
All

bm64:
Fixed #18168, legend item was mispositioned if the related series was located at the top of the chart and proximate layout was enabled.

### DIFF
--- a/ts/Core/Legend/Legend.ts
+++ b/ts/Core/Legend/Legend.ts
@@ -1054,7 +1054,7 @@ class Legend {
 
         for (const box of distribute(boxes, chart.plotHeight)) {
             legendItem = box.item.legendItem || {};
-            if (box.pos) {
+            if (isNumber(box.pos)) {
                 legendItem.y = chart.plotTop - chart.spacing[0] + box.pos;
             }
         }


TorsteinHonsi (approved):
Thanks!

PR #18189, opened 2022-12-21
Tools: build for Evergreen browsers

raf18seb:
Added `/es5` path for legacy script files on [code.highcharts.com](https://code.highcharts.com) and in the npm package, see #18070.

First part of #18070:
Added tasks for making a copy of `code` folder into `code/es5` folder (but without any dts files).


raf18seb:
Q: Should I not create new `tools/build-es5` util but instead maybe add a parameter handled by the existing `tools/build.js`?

TorsteinHonsi:
> Q: Should I not create new tools/build-es5 util but instead maybe add a parameter handled by the existing tools/build.js

In general, if we end up with a solution where the same (or similar) tooling code is duplicated in two related tasks, it is not optimal. It is more secure and more maintainable if we don't repeat ourselves.

bre1470:
@TorsteinHonsi How different do you expect the `es5` pipeline to be in 6 months?

TorsteinHonsi:
> How different do you expect the es5 pipeline to be in 6 months?

@bre1470 Not very different... I don't expect us to spend too much resources on optimizing the legacy pipeline, but just make sure it works.

TorsteinHonsi (changes requested):
Thanks, I finally got to review this.

I noticed that the compiled JS files are not generated for the `es5` folder. Only `.src.js`. We should generated compiled files.

On an opposite note, the `es-modules` filed were generated for es5. Do we need those for the legacy path? Or can we assume that those who load `es-modules` require modern browsers? It would save us some complexity, and I think it's a fair assumption to make. 

@bre1470 FYI

bre1470:
That is fair assumption imho. I suggest to move to a version with Promise support out of the box, I believe it's `ES2016` as target. So we will need some extra CLI argument for the gulp task for `ES5` target.

raf18seb:
> I noticed that the compiled JS files are not generated for the es5 folder. Only .src.js. We should generated compiled files.

It turned out that the order of tasks was incorrect (first should be `scripts-es5` to create `es5/` folder, then `scripts-compile` to compile all the `.src.js` files). I wrote the whole new script before I figured out this small order issue 😅

The `es5/es-modules` folder is now simply being deleted at the end of `scripts-es5` run.

tools/build.js
```diff
@@ -50,7 +50,7 @@ const getBuildOptions = input => {
     );
     const type = ['classic'];
     const mapTypeToSource = {
-        classic: './code/es-modules',
+        classic: [output, 'es-modules'].join(''),
         css: './code/js/es-modules'
```
bre1470, line 54: Two thoughts here:
- `path.join` from `require('path')` is the proper way to connect path fragments
- Not related to your work, so you do not need to fix it. I just notice that the css path makes no sense. There is no `code/js` folder structure anywhere

tools/gulptasks/scripts-es5.js
```diff
@@ -0,0 +1,68 @@
...
+        logLib.message('Generating code...');
+
+        processLib.isRunning('scripts-js', true);
+
+        BuildScripts
+            // assemble JS files
+            .fnFirstBuild()
+            // deleting invalid masters DTS
+            .then(() => fsLib.getFilePaths('js/masters/', true).forEach(
+                path => path.endsWith('.d.ts') && fsLib.deleteFile(path)
```
bre1470, line 53: Invalid masters DTS can be ignored in our case as es5 does not provide DTS files, if I am not mistaken. This needs some investigation.

bre1470 (changes requested):
Some comments before the session tomorrow.

TorsteinHonsi (approved):
Thanks! Pending approval by @bre1470 .

bre1470 (approved):
Good job. 👍 `scripts-es5` itself does create `code/es5` but the `dist` task fails with
```
Error: Could not find tmp/demo/sidebar.html
  If demos are built elsewhere, the path can be specified in git-ignore-me.properties by the demos.path property.
    at getLocalSidebar (/Users/highsoft/GitHub/highcharts/tools/gulptasks/dist-examples.js:193:19)
```

Maybe it is unrelated, but it is not clear, if `code/es5` gets copied over into `../highcharts-dist`.
The `build/dist/highcharts/code/es5` is there were it should be.

TorsteinHonsi (approved):
Thanks!

The sidebar issue is being handled in https://github.com/highcharts/highcharts/issues/18234

PR #18193, opened 2022-12-22
db/fullscreen

karolkolodziej:
Added fullscreen support.

- [x] Change text
- [x] Scroll on fullscreen

### DIFF
--- a/ts/Dashboard/Dashboard.ts
+++ b/ts/Dashboard/Dashboard.ts
@@ -27,13 +27,10 @@ import type JSON from '../Core/JSON';
 import Bindings from './Actions/Bindings.js';
 import Globals from './Globals.js';
 import EditMode from './EditMode/EditMode.js';
-import H from '../Core/Globals.js';
 import Layout from './Layout/Layout.js';
 import Serializable from './Serializable.js';
-import U from '../Core/Utilities.js';
 import Fullscreen from './EditMode/Fullscreen.js';
-
-const { doc } = H;
+import U from '../Core/Utilities.js';
 const {
     merge,
     addEvent,
@@ -113,7 +110,7 @@ class Dashboard implements Serializable<Dashboard, Dashboard.JSON> {
         this.initEvents();
 
         // Add fullscreen support.
-        Fullscreen.compose(this);
+        this.fullscreen = new Fullscreen(this);
 
         this.index = Globals.dashboards.length;
         Globals.dashboards.push(this);
@@ -164,7 +161,7 @@ class Dashboard implements Serializable<Dashboard, Dashboard.JSON> {
 
         if (isString(renderTo)) {
             dashboard.container = renderTo =
-                doc.getElementById(renderTo as any) as any;
+                window.document.getElementById(renderTo as any) as any;
         }
 
         // Display an error if the renderTo is wrong
--- a/ts/Dashboard/Dashboard.ts
+++ b/ts/Dashboard/Dashboard.ts
@@ -28,10 +28,12 @@ import Bindings from './Actions/Bindings.js';
 import Globals from './Globals.js';
 import EditMode from './EditMode/EditMode.js';
 import H from '../Core/Globals.js';
-const { doc } = H;
 import Layout from './Layout/Layout.js';
 import Serializable from './Serializable.js';
 import U from '../Core/Utilities.js';
+import Fullscreen from './EditMode/Fullscreen.js';
+
+const { doc } = H;
 const {
     merge,
     addEvent,
@@ -110,6 +112,9 @@ class Dashboard implements Serializable<Dashboard, Dashboard.JSON> {
         // Init events.
         this.initEvents();
 
+        // Add fullscreen support.
+        Fullscreen.compose(this);
+
         this.index = Globals.dashboards.length;
         Globals.dashboards.push(this);
     }
@@ -120,15 +125,16 @@ class Dashboard implements Serializable<Dashboard, Dashboard.JSON> {
      *
      * */
 
-    public options: Dashboard.Options;
-    public layouts: Array<Layout>;
-    public mountedComponents: Array<Bindings.MountedComponentsOptions>;
     public container: globalThis.HTMLElement = void 0 as any;
+    public editMode?: EditMode;
+    public fullscreen?: Fullscreen;
     public guiEnabled: (boolean|undefined);
     public id: string;
     public index: number;
-    public editMode?: EditMode;
+    public layouts: Array<Layout>;
     public layoutsWrapper: globalThis.HTMLElement;
+    public mountedComponents: Array<Bindings.MountedComponentsOptions>;
+    public options: Dashboard.Options;
 
     /* *
      *
--- a/ts/Dashboard/EditMode/EditContextMenu.ts
+++ b/ts/Dashboard/EditMode/EditContextMenu.ts
@@ -47,7 +47,7 @@ class EditContextMenu extends Menu {
         parentElement: HTMLElement,
         options: EditContextMenu.Options,
         editMode: EditMode,
-        parent?: any
+        parent?: HTMLElement
     ) {
         super(
             editMode.dashboard.container,
--- a/ts/Dashboard/EditMode/EditContextMenu.ts
+++ b/ts/Dashboard/EditMode/EditContextMenu.ts
@@ -44,12 +44,15 @@ class EditContextMenu extends Menu {
     *
     * */
     constructor(
+        parentElement: HTMLElement,
+        options: EditContextMenu.Options,
         editMode: EditMode,
-        options?: EditContextMenu.Options|undefined
+        parent?: any
     ) {
         super(
             editMode.dashboard.container,
-            merge(EditContextMenu.defaultOptions, options || {})
+            merge(EditContextMenu.defaultOptions, options || {}),
+            editMode
         );
 
         this.editMode = editMode;
--- a/ts/Dashboard/EditMode/EditContextMenu.ts
+++ b/ts/Dashboard/EditMode/EditContextMenu.ts
@@ -25,13 +25,13 @@ class EditContextMenu extends Menu {
         items: ['editMode']
     };
 
-    public static items: Record<string, MenuItem.Options> = merge(Menu.items, {
+    public static items = merge(Menu.items, {
         editMode: {
             id: 'editMode',
             type: 'toggle',
             text: 'Edit mode',
             events: {
-                click: function (this: MenuItem, e: any): void {
+                click: function (this: MenuItem): void {
                     (this.menu as EditContextMenu).editMode.onEditModeToggle();
                 }
             }
--- a/ts/Dashboard/EditMode/Fullscreen.ts
+++ b/ts/Dashboard/EditMode/Fullscreen.ts
@@ -73,7 +73,7 @@ class Fullscreen {
     public open(): void {
         // eslint-disable-next-line no-console
         console.log('open');
-        this.isOpen = false;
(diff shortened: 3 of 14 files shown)


ts/Dashboard/EditMode/Fullscreen.ts
```diff
@@ -0,0 +1,131 @@
...
+     * Toggles displaying the dashboard in fullscreen mode.
+     */
+    public toggle(): void {
+        const fullscreen = this;
+
+        if (!fullscreen.isOpen) {
+            fullscreen.open();
+        } else {
+            fullscreen.close();
+        }
```
pawellysy, line 68:
```suggestion
       const fullscreen = this,
           isOpen = this.isOpen;
       fullscreen[isOpen ? 'close' : 'open']()
```

ts/Dashboard/EditMode/Menu/Menu.ts
```diff
@@ -86,7 +78,7 @@ class Menu {
 
     // itemsSchemas - default items definitions.
     public initItems(
-        itemsSchemas: Record<string, MenuItem.Options>,
+        itemsSchemas: any,
```
pawellysy, line 81: Wasn't previous type better?

pawellysy (changes requested):
Good job! Just a few comments/questions.

ts/Dashboard/EditMode/Fullscreen.ts
```diff
@@ -0,0 +1,131 @@
...
+                }
+            }
+        );
+
+        fullscreen.unbindFullscreenEvent = (): void => {
+            unbindChange();
+        };
+
+        // eslint-disable-next-line @typescript-eslint/no-floating-promises
+        dashboard.container.requestFullscreen()['catch']((): void => {
```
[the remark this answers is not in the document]
karolkolodziej: The eslint is complaining but I agree that without parentheses it looks better.

ts/Dashboard/EditMode/Menu/MenuItem.ts
```diff
@@ -185,7 +185,8 @@ class MenuItem {
 namespace MenuItem {
     export interface Options {
         id: string;
-        type?: string;
+        type?: 'addComponent'|'addLayout'|'horizontalSeparator'|'icon'|'input'|
+        'toggle'|'text'|'textarea'|'verticalSeparator';
```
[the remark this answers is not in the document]
karolkolodziej: I don't see any reason why it should be moved to a separate interface.

karolkolodziej:
@pawellysy I found a few things that might be added. I will let you know when it is ready ;)

ts/Dashboard/EditMode/EditContextMenu.ts
```diff
@@ -44,12 +44,15 @@ class EditContextMenu extends Menu {
     *
     * */
     constructor(
+        parentElement: HTMLElement,
+        options: EditContextMenu.Options,
         editMode: EditMode,
-        options?: EditContextMenu.Options|undefined
+        parent?: any
```
pawellysy, line 50: isn't parent a `HTMLElement`?

ts/Dashboard/EditMode/Fullscreen.ts
```diff
@@ -0,0 +1,148 @@
...
+            }
+        );
+
+        fullscreen.unbindFullscreenEvent = (): void => {
+            unbindChange();
+        };
+
+        const promise = dashboard.container.requestFullscreen();
+
+        // eslint-disable-next-line highcharts/quote-members
```
pawellysy, line 96: no need for this line

ts/Dashboard/EditMode/Menu/MenuItemBindings.ts
```diff
@@ -0,0 +1,35 @@
+import EditGlobals from '../EditGlobals.js';
```
pawellysy, line 1: Licence  comment at the beginning of the file is missing

ts/Dashboard/EditMode/Fullscreen.ts
```diff
@@ -0,0 +1,148 @@
...
+        this.setButtonText();
+    }
+
+    /**
+     * Set the correct text depending of the fullscreen is on or of.
+     */
+    public setButtonText(): void {
+        const editMode = this.dashboard.editMode,
+            contextMenu = editMode && editMode.tools.contextMenu,
+            button = contextMenu && contextMenu.items['viewFullscreen'];
```
pawellysy, line 133: isn't there a way, to define an interface for the `items` property, so that you don't have to use quoted members?

pawellysy (changes requested):
Nice work! A few minor comments. Additionally, is there a chance for creating a karma/cypress test?

ts/Dashboard/EditMode/EditContextMenu.ts
```diff
@@ -44,12 +44,15 @@ class EditContextMenu extends Menu {
...
     ) {
         super(
             editMode.dashboard.container,
-            merge(EditContextMenu.defaultOptions, options || {})
+            merge(EditContextMenu.defaultOptions, options || {}),
+            editMode
```
sebastianbochan, line 55: Why do we need it here?

ts/Dashboard/EditMode/Fullscreen.ts
```diff
@@ -0,0 +1,148 @@
...
+
+    /* *
+    *
+    *  Properties
+    *
+    * */
+
+    public dashboard: Dashboard;
+    public isOpen: boolean;
+    public unbindFullscreenEvent?: Function;
```
sebastianbochan, line 48: Is it not too general? I mean a type: `Function`

ts/Dashboard/EditMode/Fullscreen.ts
```diff
@@ -0,0 +1,148 @@
...
+        const fullscreen = this,
+            dashboard = fullscreen.dashboard;
+
+        // Handle exitFullscreen() method when user clicks 'Escape' button.
+        const unbindChange = addEvent(
+            dashboard.container.ownerDocument, // dashboard's document
+            'fullscreenchange',
+            function (): void {
+                if (fullscreen.isOpen) {
+                    fullscreen.isOpen = false;
```
sebastianbochan, line 81: You set this flag in the `fullscreen.close()` function, so do we need it here again?

ts/Dashboard/EditMode/Fullscreen.ts
```diff
@@ -0,0 +1,148 @@
...
+                    fullscreen.close();
+                } else {
+                    fullscreen.isOpen = true;
```
sebastianbochan, line 84: You can set the flag into `fullscreen.setButtonText()`

sebastianbochan (changes requested):

ts/Dashboard/EditMode/Fullscreen.ts
karolkolodziej: My eslint complains about that :(

ts/Dashboard/EditMode/Menu/MenuItemBindings.ts
karolkolodziej: I see that in other files it is also missing.
I added a task to asana for that

ts/Dashboard/EditMode/Fullscreen.ts
karolkolodziej: Any idea how we can make it more precise?

ts/Dashboard/EditMode/Fullscreen.ts
karolkolodziej: We need it in case the user hit the Esc button. Without that we loos track if the dashboard is in fullscreen or not.

ts/Dashboard/EditMode/Fullscreen.ts
karolkolodziej: I feel like `setButtonText` should do no more and no less than what the name states.

ts/Dashboard/EditMode/EditContextMenu.ts
karolkolodziej: We don't exactly need it here but I'm trying to unify the constructors of each class that extends the `Menu` class.

ts/Dashboard/EditMode/Fullscreen.ts
sebastianbochan: I've found the `Event` type, used in a few places.

@bre1470 what do you think?

pawellysy: `Event` type is the type is the object. Here could be 
```ts
  public unbindFullscreenEvent?: () => void;
```

pawellysy (approved):
LGTM 👍

ts/Dashboard/Dashboard.ts
```diff
@@ -28,10 +28,12 @@ import Bindings from './Actions/Bindings.js';
 import Globals from './Globals.js';
 import EditMode from './EditMode/EditMode.js';
 import H from '../Core/Globals.js';
-const { doc } = H;
 import Layout from './Layout/Layout.js';
 import Serializable from './Serializable.js';
 import U from '../Core/Utilities.js';
+import Fullscreen from './EditMode/Fullscreen.js';
+
+const { doc } = H;
```
bre1470, line 36: We should use `window.document` and not the Highcharts globals. Or at least use Dashboard globals.

ts/Dashboard/Dashboard.ts
```diff
@@ -28,10 +28,12 @@ import Bindings from './Actions/Bindings.js';
...
 const {
```
bre1470, line 37: The import deconstruct should be directly under each individual import.

ts/Dashboard/EditMode/Fullscreen.ts
```diff
@@ -0,0 +1,148 @@
...
+     *  Static Functions
+     *
+     * */
+
+    /**
+     * Prepares the dashboard class to support fullscreen.
+     * @param {typeof_Dashboard} DashboardClass
+     *        The dashboard class to decorate with fullscreen support.
+     */
+    public static compose(DashboardClass: Dashboard): void {
```
bre1470, line 18: The argument is misleading, because the compose function accepts a `Dashboard` instance, not the class prototype. I suggest to skip the compose and instead move the `DashboardClass.fullscreen = new Fullscreen(DashboardClass);` to the dashboard constructor. So it becomes there `this.fullscreen = new Fullscreen(this);`.

ts/Dashboard/EditMode/Fullscreen.ts
```diff
@@ -0,0 +1,148 @@
...
+        const fullscreen = this,
+            isOpen = this.isOpen;
+
+        fullscreen[isOpen ? 'close' : 'open']();
+    }
+
+    /**
+     * Display dashboard in fullscreen.
+     */
+    public open(): void {
```
bre1470, line 71: The function lacks a check, if it already is open.

ts/Dashboard/EditMode/Fullscreen.ts
bre1470: I would expect this to become protected or private, as consumers should use `open` / `close` functions and have the `isOpen` flag. Otherwise the typing is fine for me. Microsoft likes to see the `() => void` used, but `Function` is good enough as we do not care about the return type.

bre1470 (changes requested):
Good work! 👍 Please see comments for adjustments.

sebastianbochan (approved):

bre1470 (approved):
Thank you. Looks good! 👍

PR #18196, opened 2022-12-22
Enhancement/default theme redo

nasin:
Hi, I've been revising the TS files for the default theme. Major changes include a new color palette (including highlights and neutrals), new default font (Trebuchet is closest to our brand font), and tweaks concerning border radius, width, drop shadows, etc.  

WIth the neutral colors, I tried to soften them a bit to make the axes and non-data stuff blend in a bit more to accentuate the data...don't know if it's working. I was trying to keep eveyrthing (except series labels) AA compliant. I also un-bolded some things, like the legend items' normal state, for the same reason (to help the data stand out more.)

For the highlight colors, I made highlight 100 almost black since it's used for the export menu item rollover. For the color axis (used in the maps, for example), I adjusted the max color to highlight 60, so the maps have nice blues that match highcharts-color-0 (the bright blue I picked.)

Let me know what you think so far.


TorsteinHonsi (commented):
Thanks, so far I like what I see!

I don't have time to do a proper review before the holidays, but here are some first impressions/comments:
* I like the new data colors 
* I'm not sure about the fonts, I was expecting something more standard sans-seriif, maybe even Arial, Verdana or Helvetica. Something less characteristic that blends in better on any client website?
* We need to consider (at least experiment with) larger font sizes for some of the labels, like axis labels etc. But this is hard, because it also means less labels will be rendered because of anti-collision.
* "For the highlight colors, I made highlight 100 almost black since it's used for the export menu item rollover." Is that to preserve contrast? Then maybe we should use the neutral100 instead for the export menu? The idea with the highlight color is that it is colored to highlight active items, like the drilldown active labels.

On the technical side, note this comment in Palette.ts: https://github.com/highcharts/highcharts/blob/c04edf8654a2a829aa5c21f5d094d655da3a3879/ts/Core/Color/Palettes.ts#L3-L6

Furthermore, the `highcharts.scss` file will no longer exist with v11, as it is all ported to CSS variables with https://github.com/highcharts/highcharts/pull/18115 . I suggest you make a branch out of the CSS variables branch and create a draft pull request into that branch (instead of master). I didn't give you a heads up because  I was expecting sketches for the design first, not a PR. But I appreciate that it is easier to see the full impact of the changes in a PR, both for the designer and the reviewer.

TorsteinHonsi:
So to sum up the how you can work on the theme from here, first create a fork from the [CSS Variables branch].

Then the workflow will be
1. Make changes in `highcharts.css`
2. Run `gulp palette` to generate `Palette.ts`
3. Make changes in the `.ts` files if you want to redefined what palette colors are used (but do not change `Palette.ts`)
4. Run `gulp`
5. Test

TorsteinHonsi:
Closing this as a dead end, the new PR has colors applied correctly.

PR #18201, opened 2022-12-26
bugfix/16753-number-in-parameter-name

### LINKED ISSUE #16753: Popup: when adding custom indicator with parameter which name consists a number the popup doesn't display parameter name.
#### Expected behaviour
The custom indicator should have all of the properties listed in the popup. 

#### Actual behaviour
When adding the custom indicator with a parameter which name consists of a number, the popup in stock tools doesn't display its name

https://jsfiddle.net/BlackLabel/y49rxmw3/1/
open the indicators popup
check that one of the indicator's parameters don't have the label

#### Product version
v9.3.2

#### Affected browser(s)
all

bm64:
Fixes #16753, when adding a custom indicator, if the parameter name contained any digits, the stock tools pop-up did not display it.

### DIFF
--- a/ts/Extensions/Annotations/Popup/Popup.ts
+++ b/ts/Extensions/Annotations/Popup/Popup.ts
@@ -309,7 +309,7 @@ class Popup {
                 optionName
             );
 
-        if (!inputName.match(/\d/g)) {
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (approved):
Thanks!

PR #18203, opened 2022-12-27
bugfix/18025-gridlines-across-panes

### LINKED ISSUE #18025: gridline: clipping issue in stock
#### Actual behaviour
We are currently unable to remove the red line from the gridline options: 
https://api.highcharts.com/highstock/xAxis.gridLineWidth

https://jsfiddle.net/BlackLabel/8znmf1ew/

#### Workaround: 
https://jsfiddle.net/BlackLabel/c0swz1mL/

#### Product version
v10.3.1

kamil-musialowski:
Fixed #18025, gridlines were rendered across panes.

### DIFF
--- a/ts/Core/Axis/Axis.ts
+++ b/ts/Core/Axis/Axis.ts
@@ -989,42 +989,69 @@ class Axis {
             return x;
         }
 
-        const evt: Event = {
+        const evt: Partial<Event&Axis.PlotLinePathOptions> = {
             value: value,
             lineWidth: lineWidth,
             old: old,
             force: force,
             acrossPanes: options.acrossPanes,
             translatedValue: translatedValue
-        } as any;
-        fireEvent(this, 'getPlotLinePath', evt, function (e: Event): void {
+        };
+        fireEvent(this, 'getPlotLinePath', evt, function (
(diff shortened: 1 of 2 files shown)


ts/Core/Axis/Axis.ts
```diff
@@ -989,42 +989,69 @@ class Axis {
             return x;
         }
 
-        const evt: Event = {
+        const evt: Partial<Event&Axis.PlotLinePathOptions> = {
```
pawellysy, line 992: When I think about it, I think it should be just `Partial<Axis.PlotLinePathOptions>`Since there are no `Event`'s attributes, but your call.

ts/Core/Axis/Axis.ts
```diff
@@ -989,42 +989,69 @@ class Axis {
             return x;
         }
```
pawellysy, line 990: Now, when looking at it, the body of `between` function looks a lot like what we implemented. Could you maybe investigate, why didn't it work correctly?

pawellysy (changes requested):
Good job! Few comments

pawellysy (approved):
Looks good!

TorsteinHonsi (approved):
Thanks!

PR #18205, opened 2022-12-27
Enhancement/default theme redesign

nasin:
Hi, I hope I did this right. I'm using the css variables now and made other modifications, like the font. I'll keep working here.


TorsteinHonsi:
Thanks! Yes this is technically done right. Remember to run `gulp palette` after modifying `highcharts.css`.

PR #18206, opened 2022-12-28
Db/redraw title caption

goransle:
Test plan:
* Open Climate demo in utils
* Enable edit mode
* Select the ⚙️ on one of the chart components
* Change the title and caption and click apply
* The title and caption should now be updated

### DIFF
--- a/ts/Dashboard/Component/Component.ts
+++ b/ts/Dashboard/Component/Component.ts
@@ -81,7 +81,7 @@ abstract class Component<TEventObject extends Component.EventTypes = Component.E
         },
         sync: Sync.defaultHandlers,
         editableOptions: [
-            'id',
+            // 'id',
             'store',
             'style',
             'title',
--- a/ts/Dashboard/Component/Component.ts
+++ b/ts/Dashboard/Component/Component.ts
@@ -507,28 +507,30 @@ abstract class Component<TEventObject extends Component.EventTypes = Component.E
     }
 
     public setTitle(titleOptions: Component.TextOptionsType): void {
-        if (this.titleElement) {
-            this.titleElement.remove();
-        }
-
+        const old = this.titleElement;
         const titleElement =
             Component.createTextElement('h1', 'title', titleOptions);
 
         if (titleElement) {
             this.titleElement = titleElement;
+
+            if (old) {
+                old.replaceWith(this.titleElement);
+            }
         }
     }
 
     public setCaption(captionOptions: Component.TextOptionsType): void {
-        if (this.captionElement) {
-            this.captionElement.remove();
-        }
-
+        const old = this.captionElement;
         const captionElement =
             Component.createTextElement('div', 'caption', captionOptions);
 
         if (captionElement) {
             this.captionElement = captionElement;
+
+            if (old) {
+                old.replaceWith(this.captionElement);
+            }
         }
     }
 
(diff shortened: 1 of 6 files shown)


sebastianbochan:
We should not commit the `package-lock.json`

ts/Dashboard/Component/Component.ts
```diff
@@ -322,6 +324,16 @@ abstract class Component<TEventObject extends Component.EventTypes = Component.E
     }
 
     public setStore(store: Component.StoreTypes | undefined): this {
+        // Clean up old event listeners
+        if (this.tableEvents.length) {
+            while (this.tableEvents.length) {
```
sebastianbochan, line 329: Assign the length to the variable, as a result, the code does not have to calculate the length of the array multiple times.

ts/Dashboard/Component/Component.ts
```diff
@@ -489,29 +492,45 @@ abstract class Component<TEventObject extends Component.EventTypes = Component.E
...
 
+        if (redraw) {
+            this.redraw();
+        }
+
         return this;
     }
 
     public setTitle(titleOptions: Component.TextOptionsType): void {
+        const old = this.titleElement;
```
sebastianbochan, line 510: I think that the `old` is too general, can we replace it with something more meaningful? F.i. - `prevTitle` or `currentTitle`

sebastianbochan (changes requested):

ts/Dashboard/Component/Component.ts
```diff
@@ -322,6 +324,14 @@ abstract class Component<TEventObject extends Component.EventTypes = Component.E
     }
 
     public setStore(store: Component.StoreTypes | undefined): this {
+        // Clean up old event listeners
+        while (this.tableEvents.length) {
```
sebastianbochan, line 328: I would replace it with:

```
const tableEventsLength =  this.tableEvents.length;
while (tableEventsLength) {
```
As a result, the length will be calculated only one time, instead of each time when run a loop.

ts/Extensions/DashboardPlugins/HighchartsComponent.ts
```diff
@@ -231,8 +231,8 @@ class HighchartsComponent extends Component<HighchartsComponent.ChartComponentEv
     }
 
     public redraw(): this {
-        super.redraw();
-        return this.render();
+        return super.redraw();
+        // return this.render();
```
sebastianbochan, line 235: Do we need this commented line?

sebastianbochan (changes requested):

ts/Dashboard/Component/Component.ts
goransle: It needs to be calculated every time, otherwise it will keep on going ♾️

ts/Extensions/DashboardPlugins/HighchartsComponent.ts
goransle: probably not, will remove it

ts/Dashboard/Component/Component.ts
sebastianbochan: Ok!

sebastianbochan (approved):

bre1470 (commented):
🍔 Menu focus lets the dashboard duck.
⚙️ Panel does not come up in Safari.

bre1470 (approved):

goransle:
> 🍔 Menu focus lets the dashboard duck. ⚙️ ~Panel does not come up in Safari.~ No, my fault. Panel works.

🦆 ? I guess this should be a separate asana task?

bre1470:
LGTM :)

PR #18207, opened 2022-12-28
Refacto/rstack-label-position

pawellysy:
Refactored how the position of the stack label is calculated.

### DIFF
--- a/ts/Core/Axis/Stacking/StackItem.ts
+++ b/ts/Core/Axis/Stacking/StackItem.ts
@@ -18,7 +18,8 @@
 
 import type {
     AlignObject,
-    AlignValue
+    AlignValue,
+    VerticalAlignValue
 } from '../../Renderer/AlignObject';
 import type Axis from '../Axis';
 import type BBoxObject from '../../Renderer/BBoxObject';
@@ -38,10 +39,9 @@ import SeriesRegistry from '../../Series/SeriesRegistry.js';
 const { series: Series } = SeriesRegistry;
 import U from '../../Utilities.js';
 const {
-    defined,
     destroyObjectProperties,
-    isNumber,
-    pick
+    pick,
+    isNumber
 } = U;
 
 /* *
@@ -56,7 +56,6 @@ const {
  * @private
  */
 class StackItem {
-
     /* *
      *
      *  Constructor
@@ -66,17 +65,18 @@ class StackItem {
     public constructor(
         axis: StackingAxis,
         options: StackLabelOptions,
-        isNegative: boolean,
+        negativeValue: boolean,
         x: number,
         stackOption?: StackOverflowValue
     ) {
-
-        const inverted = axis.chart.inverted;
+        const inverted = axis.chart.inverted,
+            reversed = axis.reversed;
 
         this.axis = axis;
 
-        // Tells if the stack is negative
-        this.isNegative = isNegative;
+        // The stack goes to the left either if the stack has negative value
+        // or when axis is reversed. XOR operator.
+        const isNegative = (this.isNegative = !!negativeValue !== !!reversed);
 
         // Save the options to be able to style the label
         this.options = options = options || {};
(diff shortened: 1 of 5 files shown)


ts/Core/Axis/Stacking/StackItem.ts
```diff
@@ -208,86 +212,61 @@ class StackItem {
...
-
-            // Align the label to the box
-            label.align(this.alignOptions, null as any, stackBox);
+            // Calculate the adjusted Stack position, to take into consideration
+            // The size if the labelBox and vertical alignment as
+            // well as the text alignment. It's need to be done to work with
+            // default SVGLabel.align/justify methods.
+            const { x, y } = this.adjustStackPosition({
+                labelBox: bBox,
+                vAlign: verticalAlign as VerticalAlignValue,
```
TorsteinHonsi, line 249: Can we avoid the `as VerticalAlignValue`? What happens if `verticalAlign` is undefined?

ts/Core/Axis/Stacking/StackItem.ts
```diff
@@ -302,11 +281,14 @@ class StackItem {
                     stackBox
                 );
             }
+
+            // add attr to aviod the default animation of justifyDataLabel.
```
TorsteinHonsi, line 285: Follow the convention of Upper case first in comments

ts/Core/Axis/Stacking/StackItem.ts
```diff
@@ -302,11 +281,14 @@ class StackItem {
...
             label.attr({
                 x: label.alignAttr.x,
                 y: label.alignAttr.y
             });
 
+            // check if the dataLabel should be visible.
```
TorsteinHonsi, line 291: As above

ts/Core/Series/Point.ts
```diff
@@ -194,7 +194,7 @@ class Point {
      * @name Highcharts.Point#percentage
      * @type {number|undefined}
      */
-    public percentage?: number = void 0;
+    public percentage?: number|null = null;
```
TorsteinHonsi, line 197: Do we need the `null` value? In general, if there is no logical distintion between `null` and `undefined`, we should stick to `undefined`.

ts/Core/Series/Point.ts
```diff
@@ -229,7 +229,7 @@ class Point {
      * @name Highcharts.Point#total
      * @type {number|undefined}
      */
-    public total?: number = void 0;
+    public total: number|null = null;
```
TorsteinHonsi, line 232: As above.

ts/Core/Series/Point.ts
```diff
@@ -1605,8 +1605,8 @@ namespace Point {
         key?: number|string;
         series: Series;
         point: Point;
-        percentage?: number;
-        total?: number;
+        percentage?: number|null;
+        total: number|null;
```
TorsteinHonsi, line 1609: Reconsider `null`

ts/Core/Series/Series.ts
```diff
@@ -2235,60 +2235,62 @@ class Series {
...
-                        (isNumber(threshold) && threshold) as any,
-                        yAxis.min as any
-                    );
-                }
+                    if (yBottom === stackThreshold &&
+                        stackIndicator.key === stack[xValue].base
+                    ) {
+                        yBottom = pick(
+                            isNumber(threshold) ? threshold : yAxis.min,
+                            0
```
TorsteinHonsi, line 2260: Does 0 make sense, or is it just for TS compliance? If it is just for compliance, maybe we should opt out instead?

ts/Core/Series/Series.ts
```diff
@@ -2235,60 +2235,62 @@ class Series {
...
-                // variwide's translate function (#10962)
-                if (!(series as any).irregularWidths) {
-                    (pointStack as any).setOffset(
-                        series.pointXOffset || 0,
-                        series.barW || 0
-                    );
-                }
+                    point.total = point.stackTotal = pointStack.total;
+                    point.percentage =
+                        point.y && pointStack.total &&
```
TorsteinHonsi, line 2271: What if `point.y` is 0? Then `percentage` should be 0...

ts/Series/Line/LinePoint.d.ts
```diff
@@ -38,7 +38,7 @@ declare module '../../Core/Series/PointLike' {
         low?: number;
         negative?: boolean;
         options: PointOptions;
-        stackTotal?: number;
+        stackTotal: number|null;
```
TorsteinHonsi, line 41: Reconsider `null`

TorsteinHonsi (changes requested):
Thanks! That's a solid piece of work!

ts/Core/Series/Series.ts
pawellysy: Opt out of what? Should happen when `yAxis.min` is `null`? It can be according to ts.

ts/Core/Series/Series.ts
pawellysy: And it is 0: https://jsfiddle.net/BlackLabel/4k83dxc7/
Or you mean some other scenario?

ts/Core/Axis/Stacking/StackItem.ts
pawellysy: It cannot be, because it is set in the constructor. here: https://github.com/highcharts/highcharts/blob/20d15647c198106f74cc014e6ba0982902442e86/ts/Core/Axis/Stacking/StackItem.ts#L105-L114

TorsteinHonsi: Okay. It's not ideal to override the type checking like that, but it's not the only place we're doing it. I think the appropriate way of doing this would be that `StackItem.alignOptions` were a separate type where the properties were non-optional.

ts/Core/Series/Point.ts
TorsteinHonsi: I think it should be `undefined` when `point.y` is null. If it doesn't complicate the code too much.

ts/Core/Axis/Stacking/StackItem.ts
pawellysy: seems logical. I will create seperate interface 👍

TorsteinHonsi (approved):
Good job 👏

PR #18217, opened 2022-12-29
flowmap/demo-ship-route

KacperMadej:
Changed flowmap ship route demo.

### DIFF
--- a/samples/maps/demo/flowmap-ship-route/demo.js
+++ b/samples/maps/demo/flowmap-ship-route/demo.js
@@ -78,8 +78,8 @@
                     enabled: false
                 }
             },
-            nullColor: '#BD935D',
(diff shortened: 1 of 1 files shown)


samples/maps/demo/flowmap-ship-route/demo.js
```diff
@@ -49,10 +78,27 @@
                     enabled: false
                 }
             },
-            showInLegend: false
+            nullColor: '#BD935D',
+            borderColor: '#BD935D'
```
kamil-musialowski, line 82: Great idea for the demo, I have only one concern - which is the basemap color. I was thinking about something less 'punchy' and more in harmony with the arrows color, maybe something like `#8C8F8D`? Or is it too boring :)

PR #18218, opened 2022-12-29
db/climate-cities

bre1470:
- Added color / tooltip support for all data scopes in the map
- Improved loading dynamic (parallel instead of serial)
- Updated cities and timeline
- Updated map projection
- Removed heatmap data

### DIFF
--- a/samples/dashboard/demos/climate/demo.js
+++ b/samples/dashboard/demos/climate/demo.js
@@ -4,6 +4,13 @@ const color1 = Highcharts.Color.parse('#39F');
 const color2 = Highcharts.Color.parse('#6C0');
 const color3 = Highcharts.Color.parse('#F00');
 const dataPool = new Dashboard.DataOnDemand();
+const dataScopes = {
+    'FD' : 'Days with fog',
+    'ID' : 'Days with ice',
+    'RR1' : 'Days with rain',
+    'TN' : 'Average temperature',
+    'TX' : 'Maximal temperature'
+};
 
 let citiesData;
 let cityGrid;
@@ -17,6 +24,7 @@ async function buildDashboard() {
     citiesData = await buildCitiesData();
 
     const defaultCity = await dataPool.getStore('Tokyo');
+
     const dashboard = new Dashboard.Dashboard('container', {
         components: [{
             cell: 'time-range-selector',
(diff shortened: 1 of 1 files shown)


samples/dashboard/demos/climate/demo.js
```diff
@@ -424,37 +447,70 @@ function buildDates() {
 function buildDateTicks() {
     const dates = [];
 
-    for (let date = new Date(Date.UTC(2010, 0, 15)),
+    for (let date = new Date(Date.UTC(1951, 0, 15)),
```
karolkolodziej, line 450: Do we want such old data?

karolkolodziej (approved):
Just have one question. Otherwise it works flawlessly ;)

samples/dashboard/demos/climate/demo.js
bre1470: Yes, we want them all. :)

bre1470: It is climate data so we need a couple of centuries.

samples/dashboard/demos/climate/demo.js
```diff
@@ -4,66 +4,63 @@ const color1 = Highcharts.Color.parse('#39F');
 const color2 = Highcharts.Color.parse('#6C0');
```
sebastianbochan, line 4: Can we use a more meaningful name for variables? F.i. `color_black` or

samples/dashboard/demos/climate/demo.js
```diff
@@ -4,66 +4,63 @@ const color1 = Highcharts.Color.parse('#39F');
...
+                        click: async function (e) {
+                            worldDate = new Date(e.point.x);
+                            worldCities.setData(await buildCitiesMap());
                         }
                     },
                     tooltip: {
-                        footerFormat: void 0,
-                        headerFormat: void 0,
+                        footerFormat: '',
+                        headerFormat: '',
```
sebastianbochan, line 51: Why we use empty string instead of default `undefined` ?

sebastianbochan (changes requested):

samples/dashboard/demos/climate/demo.js
bre1470: Because this throws internal exceptions. Bug in maps.

samples/dashboard/demos/climate/demo.js
bre1470: I do not think that this is wise for gradient colors. The value might change and then you have to replace the code lines. Something like `colorMin`, `colorMid`, `colorMax` might be an alternative.

bre1470: It seems that it does not matter soon, because `styledMode` will make the whole concept obsolete.

goransle (approved):

PR #18220, opened 2022-12-30
bugfix/16841-stochastic-missing-D-line

### LINKED ISSUE #16841: Highstock's Slow Stochastic indicator missing %D line when both CL and HL are 0
#### Expected behaviour
The division of the CL and HL needs to be validated to prevent NaN value.

I have tested adding the line `K = K || 0;` to solve the issue temporarily in this function `StochasticIndicator.prototype.getValues`.
```
for (i = periodK - 1; i < yValLen; i++) {
  slicedY = yVal.slice(i - periodK + 1, i + 1);
  // Calculate %K
  extremes = AU.getArrayExtremes(slicedY, low, high);
  LL = extremes[0]; // Lowest low in %K periods
  CL = yVal[i][close] - LL;
  HL = extremes[1] - LL;
  K = CL / HL * 100;
  K = K || 0; // Added to solve the issue where the CL (Current Close - Lowest Low) / HL (Highest High - Lowest Low) is equal to NaN because both values are 0
  xData.push(xVal[i]);
...
```

#### Actual behaviour
The %D line went missing because the end result is NaN after the division.

Based on the screenshot below during debugging:
Highcharts_Stock_SlowStochastic_%D_missing

- CL: 0.59 - 0.59 = 0
- HL: 0.59 - 0.59 = 0
- CL / HL = NaN

[https://jsfiddle.net/jackytanjq/sceym2Lo/2/](https://jsfiddle.net/jackytanjq/sceym2Lo/2/)
Kindly refer to the datetime starting from 22-Dec-2021 at 03:55, to find the missing %D line and value in the Slow Stochastic.

#### Product version
Highcharts Stock v9.3.2

#### Affected browser(s)
N/A

bm64:
Fixed #16841, Slow Stochastic Indicator was missing `%D` line when the `K` value was `NaN`.

### DIFF
--- a/ts/Stock/Indicators/Stochastic/StochasticIndicator.ts
+++ b/ts/Stock/Indicators/Stochastic/StochasticIndicator.ts
@@ -206,7 +206,7 @@ class StochasticIndicator extends SMAIndicator {
             K = CL / HL * 100;
 
             xData.push(xVal[i]);
-            yData.push([K, null]);
+            yData.push([isNaN(K) ? yData[yData.length - 1][0] : K, null]);
 
             // Calculate smoothed %D, which is SMA of %K
             if (i >= (periodK - 1) + (periodD - 1)) {
--- a/ts/Stock/Indicators/Stochastic/StochasticIndicator.ts
+++ b/ts/Stock/Indicators/Stochastic/StochasticIndicator.ts
@@ -192,6 +192,10 @@ class StochasticIndicator extends SMAIndicator {
             return;
         }
 
+        // If the value of initial points is constant, wait until it changes
+        // to calculate correct Stochastic values
+        let constantValues = true,
+            j = 0;
         // For a N-period, we start from N-1 point, to calculate Nth point
         // That is why we later need to comprehend slice() elements list
         // with (+1)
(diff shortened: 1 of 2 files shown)


ts/Stock/Indicators/Stochastic/StochasticIndicator.ts
```diff
@@ -206,7 +206,7 @@ class StochasticIndicator extends SMAIndicator {
             K = CL / HL * 100;
 
             xData.push(xVal[i]);
-            yData.push([K, null]);
+            yData.push([isNaN(K) ? yData[yData.length - 1][0] : K, null]);
```
pawelfus, line 209: Can we add a few more conditions here?
1. When CL = 0, then we can render the line at 0 anyway
2. When it's a bug for the first point, then `yData[yData.length - 1]` will be undefined, so use simply `null`, but please check it (+test)

I couldn't find an official statement from the Stochastic's author in his paperwork about what should be done in this case, but repeating the value seems to be a reasonable solution (if the price is frozen, then the indicator will be too)

pawelfus (changes requested):
Thanks!
See my comments inline

+ missing test

TorsteinHonsi:
@bm64 Pending resolution to @pawelfus' change request

TorsteinHonsi:
Pending final review from @pawelfus

ts/Stock/Indicators/Stochastic/StochasticIndicator.ts
```diff
@@ -205,11 +209,28 @@ class StochasticIndicator extends SMAIndicator {
...
+            } else if (constantValues && !isNaN(K)) {
+                constantValues = false;
+            }
+
             xData.push(xVal[i]);
-            yData.push([K, null]);
+
+            if (isNaN(K)) {
+                yData.push([
+                    yData[yData.length - 1] &&
```
pawelfus, line 223: Micro opt: cache manually the variable. In some cases browsers nicely cache the value, however, I'm not sure about this particular case - we actually do change the length of the array inside the if-else block. You can get the value easily from `push()`: `length = yData.push( ... );` and use it: `yData[length - 1]` or even better: `length = yData.push( ... ) - 1;` (however, personally, I don't like this hanging `-1` style). Or use it in a line above when pushing to `xData`

Note: I couldn't find how`.length` getter works - [ecma](https://262.ecma-international.org/13.0/#sec-properties-of-array-instances-length) is enigmatic. Does it actually look for the last object on the array, then read the index, then increment it?

ts/Stock/Indicators/Stochastic/StochasticIndicator.ts
```diff
@@ -205,11 +209,28 @@ class StochasticIndicator extends SMAIndicator {
...
+                j++;
+                continue;
+            } else if (constantValues && !isNaN(K)) {
+                constantValues = false;
+            }
+
             xData.push(xVal[i]);
-            yData.push([K, null]);
+
+            if (isNaN(K)) {
```
pawelfus, line 221: Could you please add a comment with an explanation of what actually happens inside the if-else? The one we discussed a few weeks ago: imho the logic is not so self-explanatory

pawelfus (changes requested):
Some nit-picking again, sorry!
Overall looks (and works) well!

TorsteinHonsi:
Pending resolution to Pawel's change requests

pawelfus (approved):
Thanks!

TorsteinHonsi (approved):

PR #18222, opened 2022-12-30
bugfix/16274-custom-marker-size-update

### LINKED ISSUE #16274: Marker size doesn't update when updating height/width of marker
#### Expected behaviour
When you press the "Switch" button, the size of the rubik's cubes should toggle between 20x20 and 35x35. This is the behavior before 9.2.0.

#### Actual behaviour
When you press the "Switch" button, the size of the rubik's cubes doesn't change.

https://stackblitz.com/edit/react-2gsjn8?file=index.js

#### Affected browser(s)
Chrome

kamil-musialowski:
Fixed #16274, custom marker symbol size was not being updated.

### DIFF
--- a/samples/unit-tests/series/marker/demo.js
+++ b/samples/unit-tests/series/marker/demo.js
@@ -181,74 +181,99 @@ QUnit.test('Marker size and position', function (assert) {
         series.stateMarkerGraphic.attr('class'),
         '#5430: State marker should have class set'
     );
-});
 
-QUnit.test('visibility', assert => {
-    const axisHeight = 150;
-    const data = [85, 82, 84, 87, 92];
-    const {
-        series: [series1, series2],
-        yAxis: [, /* yAxis1 */ yAxis2]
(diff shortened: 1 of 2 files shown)


samples/unit-tests/series/marker/demo.js
```diff
@@ -181,6 +181,31 @@ QUnit.test('Marker size and position', function (assert) {
...
+            height: 25
+        }
+    });
+
+    series.update({
+        marker: {
+            width: 50,
+            height: 50
+        }
+    });
```
kamil-musialowski, line 198: In order to test this functionality we need to: 

- Have markers enabled, with a given width/height
- Update ONLY width and height
- Check if update got triggered

Changing the `enabled` property updated the series itself, which also changed the width/height. That's why there are two updates. We couldn't do it in one go if markers were previously disabled in previous tests.

TorsteinHonsi (approved):
Thanks!

PR #18224, opened 2023-01-02
db/climate-demo-datagrid

sebastianbochan:
- [x] - disable the edit option in the Datagrid
- [x] - sync with chart
- [x] - bug in es-modules
- [x] - problem with sync


goransle (approved):
Looks good!

karolkolodziej (changes requested):
Good idea with the whole `rowXIndex` concept ;) 

Could you please only see why the last few rows are not visible in the `dataGrid`? Perhaps add some CSS styles here :)

sebastianbochan:
The issue is caused by the commit: https://github.com/highcharts/highcharts/commit/79e10afad9614bd59052c3c2793c96c17b7dd6fa

bre1470:
I think the commit just triggers the issue. The datagrid is larger than the component space, because of the headline, therefor rows are hidden.

sebastianbochan:
Problem with overflow has been fixed.

@bre1470 @karolkolodziej

karolkolodziej (approved):

bre1470 (approved):
Behind the scenes there are still rows rendered outside the visible area. These need to eventually be removed or otherwise the last ones are never visible.

Screenshot 2023-01-19 at 12 20 03

sebastianbochan:
I think that its a completely different issue and it is a problem with datagrid inside. 

There is a few issues f.e elements positioned absolutely etc. So I think that we should refactor datagrid in the future.

PR #18225, opened 2023-01-02
bugfix/17114-datagrouping-error-plotsize-zero

### LINKED ISSUE #17114: Error: <path> attribute d: Unexpected end of attribute. Expected number, "M 0 " when resizing highstock line chart
Please extract files from [highcharts-bug.tar.gz](https://github.com/highcharts/highcharts/files/8326623/highcharts-bug.tar.gz).

This is an Angular 13 app with highstock line chart. Run `npm install` and `ng serve` to launch application.

You will see something like this:

app

Grab the right window border and move it towards the left border. At some point you will get this:

bug

NOTE: you may need to drag border back and forth a few times before you get this error message.

#### Expected behaviour

This error should not appear in browser's console.

#### Actual behaviour

The error appears

See the link to the source code at the very beginning.

#### Product version

10.0.0

#### Affected browser(s)

Chrome. Most likely other browsers too.

hubertkozik:
Fixed #17114, chart with data grouping had errors if the plot width was zero.

### DIFF
--- a/samples/unit-tests/series/datagrouping/demo.js
+++ b/samples/unit-tests/series/datagrouping/demo.js
@@ -118,6 +118,24 @@ QUnit.test('General dataGrouping options', function (assert) {
         calledWithNaN,
         'Empty series should not cause getTimezoneOffset to get called with NaN timestamp (#13247)'
     );
+
+    chart.update({
+        chart: {
+            width: 10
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #18226, opened 2023-01-02
db/climate-demo-kpi

sebastianbochan:
- [x] - redesign cells / rows
- [x] - upgrade bindings / PoC KPI
- [x] - add KPI for default series
- [x] - refresh KPI when change city
- [x] - KPI data
- [x] - interactions (click events)
   - [x] - navigator
   - [x] - map
   - [x] - chart
 - [x] - states
    - [x] - hover state
    - [x] - current state
 - [x] - navigator
 - [x] - TX as default


samples/dashboard/demos/climate/demo.js
```diff
@@ -199,30 +204,17 @@ async function setupDashboard() {
                             dataPool
                                 .getStore(city)
                                 .then(store => {
-                                    const data = store.table.modified.getRows(
-                                        void 0, void 0,
-                                        ['time', dataScope]
-                                    );
+                                    dataScope = 'TX';
```
karolkolodziej, line 207: If I'm looking at something else (for example rain) and I click on another city I don't get the data I needed (rain).
To me, it is a bit confusing. What do you think?

ts/Dashboard/Actions/Bindings.ts
```diff
@@ -101,16 +115,49 @@ class Bindings {
             fireEvent(component, 'mount');
         }
 
-            if (cell && component) {
-                component.setCell(cell);
-                cell.mountedComponent = component;
+        if (cell && component) {
```
karolkolodziej, line 118: Shouldn't we also add a `selected` state? IMO after clicking on some KPI, the selection color should be changed.

ts/Dashboard/Actions/Bindings.ts
```diff
@@ -133,6 +180,9 @@ class Bindings {
             case 'DataGrid': 
                 component = DataGridComponent.fromJSON(json as DataGridComponent.ClassJSON);
                 break;
+            // case 'kpi':
```
karolkolodziej, line 183: Remove?

karolkolodziej (changes requested):

ts/Dashboard/Actions/Bindings.ts
sebastianbochan: Its intentional to do not miss it. 

`toJSON` method should be added in `KPI component`. I will create a separate ticket in Asana.

ts/Dashboard/Actions/Bindings.ts
sebastianbochan: Its active state is added in the click event (a few lines) above.

bre1470 (approved):
KPI's look good. Something is not right with the map.

samples/dashboard/demos/climate/demo.js
sebastianbochan: We adjust it when the final design will come ;)

pawellysy (approved):
Looks good! Thanks!

karolkolodziej (approved):

PR #18227, opened 2023-01-02
bugfix/17190-indicators-linkedTo-update

### LINKED ISSUE #17190: series.linkedto - not all series are updated together with the series they are linked to
#### Expected behaviour
All series linked to the updated series are updated as well.

#### Actual behaviour
The following demo shows three series. The first one - ohlc is updated by adding points. The are two other series:
- ppo linked to the ohlc,
- ema linked to the ppo.

Issue: the ppo series receives new points like the first, updated series which is linked to, whereas ema series does not.

**Demo:**
https://jsfiddle.net/BlackLabel/5jy97g3u/

#### Product version
Highstock v10.0.0 

#### Affected browser(s)
N/A

bm64:
Fixes #17190, indicator linked to another indicator didn't update when data was added to main series.

### DIFF
--- a/samples/unit-tests/indicator-sma/recalculations/demo.js
+++ b/samples/unit-tests/indicator-sma/recalculations/demo.js
@@ -306,15 +306,15 @@ QUnit.test('Order of series and indicators, #15892.', function (assert) {
             enabled: false
         },
         series: [{
-            id: 'main',
-            data: [13, 14, 15, 13, 14, 15, 13, 14, 15]
-        },
-        {
             type: 'sma',
+            id: 'sma',
             linkedTo: 'main',
             params: {
                 period: 4
             }
+        }, {
+            id: 'main',
+            data: [13, 14, 15, 13, 14, 15, 13, 14, 15]
         }]
     });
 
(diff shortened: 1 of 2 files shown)


samples/unit-tests/indicator-sma/recalculations/demo.js
```diff
@@ -306,15 +306,15 @@ QUnit.test('Order of series and indicators, #15892.', function (assert) {
...
-        {
             type: 'sma',
+            id: 'sma',
             linkedTo: 'main',
             params: {
                 period: 4
             }
+        }, {
+            id: 'main',
+            data: [13, 14, 15, 13, 14, 15, 13, 14, 15]
```
pawelfus, line 317: Q: Why change the order of the series?

bm64: It's actually not related to the PR - I think I should have mentioned it in commit message. There was [a test](https://github.com/highcharts/highcharts/blob/d85eeae99daa78343ca26317a431dcd47291d68f/samples/unit-tests/indicator-sma/recalculations/demo.js#L324) that was supposed to check if the indicator declared before main series is initialized correctly. The thing is, old chart config had main series declared before indicator and I have changed the order so that the test makes sense now.

pawelfus: Nice catch! Thanks

pawelfus (approved):
Thanks!

PR #18228, opened 2023-01-02
 bugfix/18137-map-border-width-hover

### LINKED ISSUE #18137: Map border width not reset after hover state change to inactive
#### Expected behaviour
On a map where the borderWidth is set on hover state, the borderWidth should be set on hover and it should revert to the original width when no longer hovering over a country.

#### Actual behaviour
The hover borderWidth remains set after moving away from a country. Border colours when applied via the same method are unset when no longer hovering, the same behaviour is expected of borderWidth..  

https://jsfiddle.net/n0hrswpf/8/ 

#### Product version
10.3.2 (worked in 8.2.2)

#### Affected browser(s)
Chrome 108

hubertkozik:
Fixed #18137, border width was not resetting after the state change.


kamil-musialowski (approved):

magdalena-gut (approved):

samples/unit-tests/maps/series-map/demo.js
```diff
@@ -61,92 +61,99 @@ QUnit.test('Mapping of joinBy with data', assert => {
...
     );
+
+    series.points[0].setState('hover', false);
+    series.points[0].setState('normal', false);
+
+    assert.strictEqual(
+        series.points[0].graphic['stroke-width'],
+        undefined,
+        'Map border should be set to initial after changing state (#18137).'
+    );
```
KacperMadej, line 156: It would be good to check if the hover worked because if it fails the test will pass regardless of whether the actually tested functionality is working or not.

madepiet (approved):

samples/unit-tests/maps/series-map/demo.js
hubertkozik: Added test one more assert.

KacperMadej (approved):

TorsteinHonsi (changes requested):
Thanks for the PR! The animation seems to run smooth now.

However, with the new code there are a lot of calls to `stroke-widthSetter` with an `undefined` value. See for example http://localhost:3030/samples/#view/maps/demo/all-areas-as-null, or the OP chart for this issue. Previously we would not add `attr['stroke-width']` it was `undefined`. But now it is added, passed on to the setter, which will run `removeAttributes` unnecessarily even if it doesn't exist before. So we need to filter out this earlier to avoid unnecessary DOM traffic.

One way to do that could be that `pointAttribs` check if a numeric value already exists in `point.graphic` before returning it. Or simply check if it changed at all, regardless of type or value? If it doesn't change, we shouldn't have to return it as part of `attr`.

TorsteinHonsi (changes requested):
Unfortunately I think we have been barking up the wrong tree here. In MapSeries.ts there was an `animate` function that was removed, that dealt with exactly this situation. It stopped working after we fixed the issue with `stroke-width: 'inherit'` a couple of months ago. Your fix works, but by removing the `animate` function, animation is broken.

In the following patch (based on the master), I fixed the `animate` function to work with the new attributes. Can you see any problems with that? If not, we can probably keep your tests, but use that fix instead.

```diff
diff --git a/ts/Series/Map/MapSeries.ts b/ts/Series/Map/MapSeries.ts
index 0350e43673..a7cae2e586 100644
--- a/ts/Series/Map/MapSeries.ts
+++ b/ts/Series/Map/MapSeries.ts
@@ -695,9 +695,17 @@ class MapSeries extends ScatterSeries {
                     graphic.animate = function (params,
                         options, complete): SVGElement {
 
-                        let switchBack = false;
+                        const animateIn = (
+                                isNumber(params['stroke-width']) &&
+                                !isNumber(graphic['stroke-width'])
+                            ),
+                            animateOut = (
+                                isNumber(graphic['stroke-width']) &&
+                                !isNumber(params['stroke-width'])
+                            );
+
                         // When strokeWidth is animating
-                        if (params['stroke-width']) {
+                        if (animateIn || animateOut) {
 
                             const strokeWidth = pick(
                                     series.getStrokeWidth(series.options),
@@ -711,25 +719,23 @@ class MapSeries extends ScatterSeries {
                                         1
                                     )
                                 );
-                            // For animating from inherit,
-                            // .attr() reads the property as the starting point
-                            if (graphic['stroke-width'] === 'inherit') {
+                            // For animating from undefined, .attr() reads the
+                            // property as the starting point
+                            if (animateIn) {
                                 graphic['stroke-width'] = inheritedStrokeWidth;
                             }
-                            // For animating to inherit
-                            if (params['stroke-width'] === ('inherit' as any)) {
+                            // For animating to undefined
+                            if (animateOut) {
                                 params['stroke-width'] = inheritedStrokeWidth;
-                                switchBack = true;
                             }
                         }
                         const ret = animate.call(
                             graphic, params, options,
-                            switchBack ? function (this: SVGElement): void {
-                                // Switch back to "inherit" for zooming
-                                // to work with the existing logic + complete
-                                graphic.attr({
-                                    'stroke-width': ('inherit' as any)
-                                });
+                            animateOut ? function (this: SVGElement): void {
+                                // Remove the attribute after finished animation
+                                graphic.element.removeAttribute('stroke-width');
+                                delete graphic['stroke-width'];
+
                                 // Proceed
                                 if (complete) {
                                     complete.apply(this, arguments);
```

hubertkozik:
I don't see any problems with that fix. Sorry, I didn't notice, that the animation has stopped working. That fix looks awesome now...

TorsteinHonsi (approved):

PR #18229, opened 2023-01-02
bugfix/17888-plotLine-label-align

### LINKED ISSUE #17888: PlotLine labels do not appear when using align right with textAlign left
#### Expected behaviour
PlotLine label text is in DOM
PlotLine label appears to the right of the chart

#### Actual behaviour

PlotLine label text is not in DOM
PlotLine label is not appearing anywhere

* Add a y-axis plotline label, with `align: 'right'` and `textAlign: 'left'`
* plotline label does not appear
* will appear if you set `useHTML: true`, or use a different the textAlign value

Demo with a working plotline label, and a missing plotline label
https://jsfiddle.net/br8ag2np/

#### Product version
Highcharts 10.2.1

bm64:
Fixed #17888, plot line labels with `align: 'right'` and `textAlign: 'left'` weren't displayed.

### DIFF
--- a/ts/Core/Axis/PlotLineOrBand/PlotLineOrBand.ts
+++ b/ts/Core/Axis/PlotLineOrBand/PlotLineOrBand.ts
@@ -345,13 +345,12 @@ class PlotLineOrBand {
             width: arrayMax(xBounds) - x,
             height: arrayMax(yBounds) - y
         });
-
         if (!label.alignValue || label.alignValue === 'left') {
             label.css({
                 width: (
                     label.rotation === 90 ?
                         axis.height - (label.alignAttr.y - axis.top) :
-                        axis.width - (label.alignAttr.x - axis.left)
+                        axis.chart.chartWidth - (label.alignAttr.x - axis.left)
                 ) + 'px'
             });
         }
--- a/ts/Core/Axis/PlotLineOrBand/PlotLineOrBand.ts
+++ b/ts/Core/Axis/PlotLineOrBand/PlotLineOrBand.ts
@@ -265,7 +265,6 @@ class PlotLineOrBand {
             // apply defaults
             optionsLabel = merge({
                 align: horiz && isBand && 'center',
-                clip: false,
                 x: horiz ? !isBand && 4 : 10,
                 verticalAlign: !horiz && isBand && 'middle',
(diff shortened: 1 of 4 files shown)


TorsteinHonsi (changes requested):
Thanks for the PR!

Unfortunately it reintroduces #15777, where the idea was that labels should not extend past their plot area.

There seems to be a conflict of interest here, so what we probably need to do is to introduce an option, `label.clip`, that is false by default to preserve the current functionality. 

Also part of this PR, we should include tests for both cases, including #15777.

ts/Core/Axis/PlotLineOrBand/PlotLineOrBand.ts
```diff
@@ -914,6 +917,15 @@ export default PlotLineOrBand;
  * @apioption  xAxis.plotLines.label.align
  */
 
+/**
+ * Whether to hide labels that are outside the plot area.
+ *
+ * @type      {boolean}
+ * @default   false
+ * @since     next
+ * @apioption xAxis.plotLines.clip
```
TorsteinHonsi, line 926: It is on`xAxis.plotLines.labels.clip`

ts/Core/Axis/PlotLineOrBand/PlotLineOrBand.ts
```diff
@@ -345,13 +346,15 @@ class PlotLineOrBand {
             width: arrayMax(xBounds) - x,
             height: arrayMax(yBounds) - y
         });
-
         if (!label.alignValue || label.alignValue === 'left') {
+            const length = optionsLabel.clip ?
```
TorsteinHonsi, line 350: Better call it `width` because that's what it reflects

ts/Core/Axis/PlotLineOrBand/PlotLineOrBand.ts
```diff
@@ -265,6 +265,7 @@ class PlotLineOrBand {
             // apply defaults
             optionsLabel = merge({
                 align: horiz && isBand && 'center',
+                clip: false,
```
TorsteinHonsi, line 268: Is this necessary? If we leave it out, it will become `undefined` by default which will still work correctly on line 350.

TorsteinHonsi (changes requested):

TorsteinHonsi (approved):
Thanks!

PR #18230, opened 2023-01-02
db/climate-marker

bre1470:
If we go with this design, we should reduce the number of cities.


samples/dashboard/demos/climate/demo.js
```diff
@@ -120,10 +120,27 @@ async function setupDashboard() {
...
+                        verticalAlign: 'top',
+                        x: 5,
+                        y: -10
+                    }, {
+                        crop: false,
+                        enabled: true,
+                        formatter: function () {
+                            const point = this;
+
+                            return labelFormatter(point.y, true);
```
pawelfus, line 138: Nitpicking: unnecessary extra argument

pawelfus (approved):

TorsteinHonsi:
> If we go with this design, we should reduce the number of cities.

I don't think we need to do that. The data labels handle overlapping gracefully. Instead, I think we should
* Use true data label mapmarkers as seen in the `maps/demo/mappoint-datalabels-mapmarker` demo. The difference is that the marker _is_ the label itself, instead of the label being superimposed on the point marker. The advantage being handling of label overlapping and, optionally, nice stacking of labels if `allowOverlap` is true.
* Show Celcius instead of Kelvin. That way we have mostly two digits and can...
* Make the labels smaller.

See the [Netatmo Weathermap](https://weathermap.netatmo.com/) for inpiration. Overlapping of labels isn't a problem.

samples/dashboard/demos/climate/demo.js
bre1470: Good catch.

TorsteinHonsi (changes requested):
Remaining change requests:
* Show Celcius instead of Kelvin. That way we have mostly two digits and can...
* Make the labels smaller.

TorsteinHonsi:
This can be avoided by setting ordering the points by descending latitude.
Skjermbilde 2023-01-04 kl  13 38 09

karolkolodziej (changes requested):
After changing the navigator, the text inside the markers is not centered.
image

bre1470:
Yes, I noticed a lot is going on there.
- The styling fix unfortunately does not work for label shapes.
- The animation of label shapes has some position flaw, which makes them move with every redraw.
- In addition the border of the selected city is gone.
- Also the styling fix does not cover selected cities.

bre1470:
> * Also the styling fix does not cover selected cities.

Pending PR: https://github.com/highcharts/highcharts/pull/18247

samples/dashboard/demos/climate/demo.css
```diff
@@ -64,12 +67,19 @@
     stroke: #986;
 }
 
-#world-map .highcharts-series-1 .highcharts-data-label text {
+#world-map .highcharts-series-1 .highcharts-data-label:nth-child(odd) text {
```
sebastianbochan, line 70: We can use default style for text and use only `even` as extension. We do not need to declare odd and even.

sebastianbochan (approved):

Issue #14114, opened 2020-08-24
Highstocks: colorAxis marker color conflicts with styledMode


xy2iii:
#### Expected behaviour
When setting a `colorAxis` and styling markers with it, with `styledMode` on, we should expect the color to be set on each marker, since we are styling it within Highcharts. (Are there other ways to do this?)

#### Actual behaviour
With `styledMode` on, each marker has a color, done by setting a `fill` attribute on each marker. With `styledMode` on, the color for each marker is not set. 

#### Live demo with steps to reproduce
`styledMode` off: https://jsfiddle.net/xy2iii/1L0sgm8r/
`styledMode` on: https://jsfiddle.net/xy2iii/exyb7ko9/

#### Product version
Latest Highstocks (v8.2.0)

#### Affected browser(s)
Firefox, Chrome

Thank you for your time!

pawelfus:
Thank you for reporting this issue @xy2iii 

*Simplified demo:*
https://jsfiddle.net/BlackLabel/adz8kb6s/3/

**Workaround:**
Demo: https://jsfiddle.net/BlackLabel/adz8kb6s/4/
Plugin:
```js
(function(H) {
  H.addEvent(
    H.Series,
    'afterRender',
    function() {

      if (!this.chart.styledMode || !this.colorAxis) {
        return;
      }

      this.points.forEach(function(point) {
        if (point.graphic && point.color) {
          point.graphic.attr({
            fill: point.color
          });
        }
      })
    }
  );
})(Highcharts);
```

*Internal note:*
Check if Highcharts Maps or Heatmap/Treemap series have the same issue. If not, let's implement the same solution across the Highcharts & Stock products.

PR #18232, opened 2023-01-03
db/climate-city

karolkolodziej:


bre1470 (approved):
Looks good. 👍 One thought: If you remove the point border and increase the size slightly, it might blend better over the curve.

samples/dashboard/demos/climate/demo.js
```diff
@@ -309,10 +326,15 @@ async function setupDashboard() {
                 },
                 xAxis: {
                     type: 'datetime',
-                    visible: true,
+                    visible: false,
                     labels: {
                         format: '{value:%Y-%m-%d}'
                     }
```
TorsteinHonsi, line 332: Why hide the X axis labels? Without them we don't know what we're looking at. I find that making them visible (by default) and using the default `labels.format` works well for this chart. A custom labels format is not recommended because it hardwires it regardless of what time span we're looking at.
Skjermbilde 2023-01-04 kl  08 17 35

samples/dashboard/demos/climate/demo.js
```diff
@@ -272,19 +269,39 @@ async function setupDashboard() {
...
+
+                            // force point colors
+                            for (const point of this.series[0].points) {
+                                if (point.graphic && point.color) {
+                                    point.graphic.element.style.fill =
+                                        point.color;
+                                }
+                            }
+                        }
+                    },
```
TorsteinHonsi, line 291: Why is this hack necessary? It's used twice in this demo. Styled mode should work with `colorAxis` out of the box, like seen for example in http://localhost:3030/samples/#view/highcharts/css/grid-light-highmaps.

If it doesn't, it's a bug that should be reported and fixed. Maybe this is an issue with mappoint and scatter series types?

TorsteinHonsi (changes requested):

pawellysy (approved):
LGTM!

samples/dashboard/demos/climate/demo.js
karolkolodziej: That was the initial design, it's going to be changed anyway so I'm going to leave it as it is for now :)

samples/dashboard/demos/climate/demo.js
bre1470: #14114

bre1470: PR: https://github.com/highcharts/highcharts/commit/ea30b04d4f873cf575ca0587af217deec1c96eac

PR #18241, opened 2023-01-04
Fixed error when no property set in config, more verbose errors

goransle:
Seems that package-lock.json was out of sync as well. Let me know if that should be updated in another PR


TorsteinHonsi (approved):
That's fine, thanks!

PR #18247, opened 2023-01-05
Limited impact of styled ColorAxis to fill color.

bre1470:
Also restored undocumented change in marker animation.

### DIFF
--- a/ts/Core/Series/Series.ts
+++ b/ts/Core/Series/Series.ts
@@ -2751,16 +2751,23 @@ class Series {
                     }
 
                     // Presentational attributes
-                    if (graphic && (!styledMode || colorAxis)) {
-                        graphic[
-                            // #14114
-                            styledMode && colorAxis ? 'css' : verb
-                        ](
-                            series.pointAttribs(
-                                point,
-                                (point.selected && 'select') as any
+                    if (graphic) {
+                        const pointAttr = series.pointAttribs(
+                            point,
+                            (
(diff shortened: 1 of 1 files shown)


TorsteinHonsi:
Thanks! Can you explain some more what this PR does?

> Limited impact of styled ColorAxis to fill color

What other impacts were there?

> Also restored undocumented change in marker animation.

How did that materialize?

TorsteinHonsi:
Note that we now have this duplicate code, which I think was what the other PR tried to address by deleting the first one.

https://github.com/highcharts/highcharts/blob/288cd5fae85a87145d99ad07458c27626f77d0a9/ts/Core/Series/Series.ts#L2698-L2701

https://github.com/highcharts/highcharts/blob/288cd5fae85a87145d99ad07458c27626f77d0a9/ts/Core/Series/Series.ts#L2751-L2754

bre1470:
- It was taking all styling attributes of the point, not just fill color.
- Also options of marker states where winning over CSS because of point styling attributes.

> Note that we now have this duplicate code, which I think was what the other PR tried to address by deleting the first one.
I did not notice this and will undo mine. I did not saw any relation to the fix and animations in the climate map have some flaws.

bre1470:
> > Also restored undocumented change in marker animation.

> How did that materialize?

I did not test it, but I was speculating that it might have been related to random reposition animations when labels change.

TorsteinHonsi (approved):
Thanks!

PR #18253, opened 2023-01-05
bugfix/18166-mapline-stroke-width-zoom-scale

### LINKED ISSUE #18166: Map SVG line width set per path not scaling on zoom
#### Expected behaviour
SVG path line width should scale on zoom when defined for each line separately, as it does when the line width is set for the whole series.

#### Actual behaviour
When giving the line width for each SVG path individually, rather than for the whole series, the width doesn't scale on zoom. If the width is specified for the whole series it scales on zoom as expected. Further more, when the width is given for each path it does resize after zoom if the user views the tooltip.

https://jsfiddle.net/h6jkx8y2/2/

Zoom the map and the lines representing flights from Lerwick don't change size (width set for each line) while those for London flights do (set at series level). While zoomed, hover over a Lerwick line so the tooltip appears and then move away, the line will resize when the tooltip fades. Happens in reverse when zooming back out.

#### Product version
10.3.2

#### Affected browser(s)
Chrome 108

kamil-musialowski:
Fixed #18166, border or line width set on point was not scaling on zoom in `map` and `mapline` series.

### DIFF
--- a/samples/unit-tests/maps/mapline-linewidth/demo.js
+++ b/samples/unit-tests/maps/mapline-linewidth/demo.js
@@ -56,5 +56,17 @@ QUnit.test(
             0.0000000001,
             'Line width in mapline should have a value of 10 on hover state (#5201, #17105).'
         );
+
+        chart.series[1].points[0].update({
+            lineWidth: 15
+        });
+
+        const pointLineWidth = chart.series[1].points[0].graphic['stroke-width'];
+        chart.mapView.zoomBy(2);
+
+        assert.ok(
+            chart.series[1].points[0].graphic['stroke-width'] < pointLineWidth,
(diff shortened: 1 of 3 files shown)


DJTechnoo (approved):

magdalena-gut (approved):

madepiet (approved):

TorsteinHonsi (changes requested):
Awesome, it works great for both `map` and `mapline` series. With smooth animation and all 👏 

As a further improvement, I think we should refactor out the `series.point` loop so that we don't repeat ourselves. The logic inside is repeated twice in MapSeries.ts, the only difference is the `scale` or `scaleStep`.

TorsteinHonsi (approved):

PR #18256, opened 2023-01-06
Disabled an eslint error which might block a merge commit

goransle:
lint-staged picked this up on a merge when there were conflicting files.


TorsteinHonsi (approved):
Thanks!

PR #18257, opened 2023-01-06
Fixed tooltip font family

nasin:
I changed the font to web-safe Arial/Helvetica. The external tooltip font wasn't working.


TorsteinHonsi (approved):
Thanks!

PR #18261, opened 2023-01-09
enhancement/geoheatmap-series

magdalena-gut:
Geo Heatmap Series Type


TorsteinHonsi:
Hi @magdalena-gut , looking at the preliminary code here, I see it contains the code for markers. I suggest sticking with squares for now, ignoring markers, because I don't think it makes much sense on a geo heatmap.

I think the main thing that this series type needs to do is to handle `lon` and `lat` as input, then after `getCellAttributes` has run, translate the resulting rectangle into a geo polygon.

PR #18263, opened 2023-01-09
flowmap/migration-demo

KacperMadej:
Added flowmap demo for migration.

The demo preview: https://highcharts.github.io/highcharts-utils/samples/#gh/65bf0c09cc/sample/maps/demo/flowmap-migration


TorsteinHonsi (changes requested):
Hei, while working on a demo I just noticed I think the `weight` is wrong. [In this jsFiddle](https://jsfiddle.net/highcharts/x9yr245m/), the max width is set to 15, but the widest arrow seems to be more like 30 pixels. Maybe we forgot to divide by 2 somewhere and added the width on both sides of the arrow?

TorsteinHonsi (approved):

PR #18268, opened 2023-01-09
Db/html element helper

pawellysy:
Created serialization support for HTML Table element in HTMLTableStore.

### DIFF
--- a/ts/Dashboard/SerializeHelper/HTMLTableConverterHelper.ts
+++ b/ts/Dashboard/SerializeHelper/HTMLTableConverterHelper.ts
@@ -37,12 +37,13 @@ const { merge } = U;
  * @return {HTMLTableConverterHelper.JSON} Serialized object
  */
 function toJSON(obj: HTMLTableConverter): HTMLTableConverterHelper.JSON {
-    return {
+    const json: HTMLTableConverterHelper.JSON = {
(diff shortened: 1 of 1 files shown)


bre1470 (approved):
Looks good. 👍

ts/Dashboard/SerializeHelper/HTMLTableConverterHelper.ts
```diff
@@ -60,7 +59,10 @@ function jsonSupportFor(obj: unknown): obj is HTMLTableConverter {
  * @return {HTMLTableConverter} New Data Converter object created from serialized object
  */
 function fromJSON(json: HTMLTableConverterHelper.JSON): HTMLTableConverter {
-    return new HTMLTableConverter(json.options);
+    const id = json.tableElementId,
+        tableElement = id ? document.getElementById(id) : null;
+
+    return new HTMLTableConverter(json.options, tableElement);
```
sebastianbochan, line 65: What in the case when ID is null, does it return the HTMLTableConverter?

pawellysy: Yes, it does, it creates the instance of the `HTMLTableConverter`, but without the tableElement.

PR #18270, opened 2023-01-10
db/chart-update-store

karolkolodziej:
- [x] Basic functionality 
- [x] Restore focus on input 
- [x] Test

### DIFF
--- a/ts/Extensions/DashboardPlugins/DataGridComponent.ts
+++ b/ts/Extensions/DashboardPlugins/DataGridComponent.ts
@@ -193,6 +193,11 @@ class DataGridComponent extends Component<DataGridComponent.ChartComponentEvents
                     this.store.table.setColumns(e.table.getColumns());
                 }
             });
+
+            // Update the DataGrid when store changed.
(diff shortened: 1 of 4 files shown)


bre1470 (changes requested):
Please see comments.

pawellysy (approved):

ts/Extensions/DashboardPlugins/DataGridComponent.ts
```diff
@@ -193,6 +194,32 @@ class DataGridComponent extends Component<DataGridComponent.ChartComponentEvents
...
+                            const input =
+                                cell.childNodes[0] as HTMLInputElement;
+
+                            if (cell.dataset.columnName === e.columnName &&
+                                +input.value === e.cellValue
+                            ) {
+                                shouldUpdateTheGrid = false;
+                            }
+                        }
+                    });
```
bre1470, line 218: Good job. 👍 It is quite some overhead. Let's do this for now. Maybe in the future we should check, if we can save and check the `table.version` property.

bre1470 (approved):

PR #18272, opened 2023-01-10
Migration: EMAIndicator

marvin19:
Improved EMAIndicator. Also checked EMAOptions.d.ts, but found nothing there.


TorsteinHonsi (approved):
Thanks!

PR #18273, opened 2023-01-10
bugfix/17128-ordinal-toValue

### LINKED ISSUE #17128: Zoom with flags
#### Expected behaviour
The zoom function should not change time scope if zoom is made after a flag on a series.

#### Actual behaviour
If I have a time series and set out a flag attached to it. The zoom works before the flag  but not after the flag in time.

https://jsfiddle.net/rick822/p64obe01/

1. Make zoom 3 times before the flag on the time series e.g before 12:00. This works fine
2. Make zoom 3 times after the flag on the time series e.g after 12:00. This makes the time scope change. for example, zooming between 13:00-14:00 gives me a zoomed view. Then, zoom between 13:30-13:40 the time suddenly jumps to 14:00 and again zoom makes me end up in 15:00.
3. Add the "flags1" data to the series which adds a flag on the first point in the series and the same behaviour will now apply on points after 08:00.
4. remove all flags "flags1, flags2" and zoom  works on the series.

Seems like data after a flag on a series gets a different behaviour than data before a flag. This makes it impossible to examine values on a lower level with zoom if the dataset is much larger than the demo values in my example are.

Might have someting to do with datagrouping, however datagrouping is needed in my real case.

#### Product version
Highstock JS v10.0.0 (2022-03-07)

#### Affected browser(s)
Tested in chrome and safari

karolkolodziej:
Fixed #17128, zooming was inaccurate when `ordinal` was enabled with series like `flag` or `scatter`.

### DIFF
--- a/samples/unit-tests/axis/ordinal/demo.js
+++ b/samples/unit-tests/axis/ordinal/demo.js
@@ -493,7 +493,7 @@ QUnit.test('val2lin- unit tests', function (assert) {
     );
 });
 
-QUnit.test('Ordianl axis, data grouping and boost module, #14055.', assert => {
+QUnit.test('Ordinal axis, data grouping and boost module, #14055.', assert => {
     const chart = Highcharts.stockChart('container', {
         series: [{
             data: [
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #18283, opened 2023-01-12
bugfix/18246-mapline-colorbypoint-api

### LINKED ISSUE #18246: Map SVG line colour doesn't cycle through defined colours when set for individual lines
#### Expected behaviour
If there a more lines in the data than there are colours defined I expected it to cycle through the colours array and reuse colours, as it does for series colours.

#### Actual behaviour
The last colour in the array is used for all lines where the array index is greater than the number of array items. Only happens when the colours are set on individual lines, if set at series level it cycles as expected.

https://jsfiddle.net/god0Lpr5/28/ 

In the example there are 3 colours defined and 6 flight routes. I expected to see each colour used twice but red and green are used once and everything else is blue.

#### Product version
10.3.2

#### Affected browser(s)
Chrome 108

hubertkozik:
Fixed #18246, missing map and mapline colorByPoint API description.

### DIFF
--- a/samples/maps/plotoptions/mapline-colorbypoint-false/demo.css
+++ b/samples/maps/plotoptions/mapline-colorbypoint-false/demo.css
@@ -1,0 +1,6 @@
+#container {
+    height: 500px;
+    min-width: 310px;
+    max-width: 800px;
+    margin: 0 auto;
+}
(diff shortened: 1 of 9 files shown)


TorsteinHonsi (approved):
Thanks!

PR #18285, opened 2023-01-12
Db/fix default editable options

goransle:


ts/Dashboard/Component/Component.ts
```diff
@@ -800,7 +803,7 @@ namespace Component {
         // allow overwriting gui elements
         navigationBindings?: NavigationBindingsOptionsObject[];
         events?: Record<string, Function>;
-        editableOptions: Array<keyof EditableOptions>;
+        editableOptions: Array<EditableOptionsType>;
```
sebastianbochan, line 806: I'm not pretty sure why we declare it as a string and ... do not use it anywhere. 

TD;LR What is a reason of not using `Array<string>;`

goransle: Mostly for if we want to change it back to a stricter typing. But can probably remove it for now

karolkolodziej (approved):

sebastianbochan (approved):

PR #18289, opened 2023-01-12
db/navigator-colors

bre1470:
I struggle with the shapes of the handle elements in navigator, because these are SVGs. And the reposition of the scrollbar to the top calls also for a hack.

### DIFF
--- a/samples/dashboard/demos/climate/demo.css
+++ b/samples/dashboard/demos/climate/demo.css
@@ -9,24 +9,40 @@
 #time-range-selector {
     --color-axis: #646479;
     --color-background: #fafafa;
-    --color-grid: #fafafb;
+    --color-grid: #9a92ce;
     --color-handle: #5352be;
-    --color-handle-line: #dadae2;
-    --color-handle-mask: #41378233;
+    --color-mask: #41378233;
+    --color-mask-line: #9a92ce;
     --color-scroll: #9295ba;
     --color-series: #5352be;
+    --color-thumb: #5352be;
+    --color-thumb-line: #f7f7f7;
+}
+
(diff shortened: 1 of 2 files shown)


pawelfus:
> I struggle with the shapes of the handle elements in navigator, because these are SVGs

Created simple arrows for that: https://jsfiddle.net/BlackLabel/xqew7o6p/

> And the reposition of the scrollbar to the top calls also for a hack.

Please, skip that.

pawelfus (changes requested):
Looks good! Just some nitpicking:

- I don't believe I'm the one asking this.. 😅 but could you please move the right arrowhead by 1-2px to the right? The same distance from the mask border as the left one: 
image
- could you change the scrollbar's track background? As for the scrollbar buttons, let's keep them as they are or use the same colors as mask arrows. What do you think @nasin ? Right now it's not possible to disable them

samples/dashboard/demos/climate/demo.css
```diff
@@ -18,38 +64,86 @@
...
 
+/* *
+ *
+ *  KPI
+ *
+ * */
+
+.hd-cell.hd-cell-state-hover:hover > .hd-component,
+.hd-cell.hd-cell-state-active > .hd-component {
+    background-color: #46475d;
```
nasin, line 75: I was using this as the highlight color for table cells: #5352BE

nasin (approved):
All of this looks good. I just noted that the highlight color for the table rows is different from what I have in Figma.

sebastianbochan (approved):

pawelfus:
I don't see my requested changes..?

PR #18296, opened 2023-01-13
heatmap interpolation

MarkusBarstad:
Added the [heatmap.interpolation](https://api.highcharts.com/highcharts/plotOptions.heatmap.interpolation) option for creating smooth heatmaps.

TODO:
- [x] aspect ratio,
- [x] tooltip indicating what the data is,
- [x] legend label indicating what the data is,
- [x] hide xaxis,
- [x] denser data


ts/Series/Heatmap/HeatmapSeries.ts
```diff
@@ -426,18 +435,67 @@ class HeatmapSeries extends ScatterSeries {
...
+            if (canvas && ctx && colorAxis) {
+                const pixels =
+                    ctx.createImageData(canvas.width, canvas.height),
+                    image = heatmap.image;
+
+                let pixelPos = 0,
+                    color,
+                    rgb;
+
+                heatmap.data.forEach((p: HeatmapPoint): void => {
```
TorsteinHonsi, line 458: This `data.forEach` must be replaced by logic that finds the actual position of the data points on the heatmap. What this code currently does is to just fill in pixels in the canvas successively. In the current demo, when toggling the `Interpolation`, we can see that it is not correct.

That's because the data is not necessarily sorted, so when the points are applied to pixels, each point's pixel position must be determined individually. Use the `x` and `y` values and the `colsize` and `rowsize` options as well axis extremes.

ts/Series/Heatmap/HeatmapSeries.ts
```diff
@@ -449,14 +449,19 @@ class HeatmapSeries extends ScatterSeries {
...
+                    image = heatmap.image,
                     { min: xMin, max: xMax } = heatmap.getXExtremes(
                         heatmap.xData || []
                     ),
                     { dataMin: yMin, dataMax: yMax } = heatmap.getExtremes(),
                     xSpan = xMax - xMin,
                     ySpan = (yMax as number) - (yMin as number);
 
+                if (image) {
+                    image.destroy();
```
TorsteinHonsi, line 462: Are you sure? Is it efficient to destroy and rebuild every time the chart redraws?

MarkusBarstad: The thing is, I was a bit unsure if there is another way to update the image in case of new data. 

I guess destroying it on every redraw is a bit much, but do we have some internal property perhaps that tells us data has changed (thus an image needs a new source)?

Or can we just assign a new image to the heatmap-series, and the previous one will go out of scope?

TorsteinHonsi: I don't think this code executes at all if the data doesn't require updating. What I think is most efficient is to create the image only once, resize it only if a size change is needed (compare the existing width/height against the required width/height), and modify the pixels, which is probably always necessary at this point.

MarkusBarstad: Turns out it was not necessary to do anything with the content of the image - it reflects new data via the update-function, and as far as I can tell it does not create a new image. I also made it so it only resizes when needed as you describe.

ts/Series/Heatmap/HeatmapSeries.ts
```diff
@@ -517,10 +515,8 @@ class HeatmapSeries extends ScatterSeries {
                         plotHeight
                     ).add(heatmap.group);
                 }
+                Series.prototype.drawPoints.call(heatmap);
```
TorsteinHonsi, line 518: Back to drawing an SVG shape for each point? That will be slow for heatmaps with thousands of points. Have you considered using a k-d-tree for the tooltips? We already have the code for it in the boost module.

samples/graphics/heatmap-userdata-backgroundimage/exampleScreenshot.png
TorsteinHonsi, line 1: Please use kebab-case for example file names

samples/highcharts/series-heatmap/interpolation/demo.js
```diff
@@ -0,0 +1,130 @@
...
+        clusterRegion(y, x, -i, heat);
+        addIfWithin(y - i, x + i, Math.max(0, heat - 2));
+        addIfWithin(y + i, x - i, Math.max(0, heat - 2));
+    }
+};
+
+Highcharts.chart('container', {
+    chart: {
+        type: 'heatmap',
+        plotBackgroundImage: 'https://code.highcharts.com/samples/graphics/heatmap-userdata-backgroundimage/exampleScreenshot.png',
```
TorsteinHonsi, line 42: This will not work. See https://github.com/highcharts/highcharts/tree/master/samples#how-to-set-up-sample-data-and-graphics.

samples/highcharts/series-heatmap/interpolation/demo.js
```diff
@@ -0,0 +1,130 @@
...
+    addIfWithin(y, x, size);
+
+    for (let i = 1; i <= size; i++) {
+        const heat = size - i;
+        clusterRegion(y, x, i, heat);
+        clusterRegion(y, x, -i, heat);
+        addIfWithin(y - i, x + i, Math.max(0, heat - 2));
+        addIfWithin(y + i, x - i, Math.max(0, heat - 2));
+    }
+};
```
TorsteinHonsi, line 37: This may look confusing to users. In the real world the data set would be coming from a CSV or a database or similar. Please use static data in the form of CSV or JSON or similar. If it is so large that it steals focus from the real sample, it can be moved to ta data file or to a `<pre>` section within the `demo.html` file.

samples/highcharts/series-heatmap/interpolation/demo.js
```diff
@@ -0,0 +1,130 @@
...
+                            { x: xIndex, y: yIndex, value: heat }
+                        )
+                    )
+                ).flat(1);
+
+                this.addSeries({
+                    data: points,
+                    interpolation: true,
+                    opacity: 0.7
+                });
```
TorsteinHonsi, line 109: Is this too part of generating the random-ish data? If so, move it out of the sample, and use static data instead.

ts/Series/Heatmap/HeatmapSeries.ts
```diff
@@ -391,17 +399,30 @@ class HeatmapSeries extends ScatterSeries {
             }
 
         }
-
+        /**
+         * Make the heatmap render its datapoints as a interpolated image.
```
TorsteinHonsi, line 403: *data points

ts/Series/Heatmap/HeatmapSeries.ts
```diff
@@ -426,18 +447,137 @@ class HeatmapSeries extends ScatterSeries {
      * @private
      */
     public drawPoints(): void {
+        const
+            heatmap = this,
+            heatmapOptions = heatmap.options,
```
TorsteinHonsi, line 452: To keep this consistent with existing code (and easier for colleagues to read), these two variables should be called `series` and `options`.

ts/Series/Heatmap/HeatmapSeries.ts
```diff
@@ -426,18 +447,137 @@ class HeatmapSeries extends ScatterSeries {
...
+            interpolation = heatmapOptions.interpolation,
+
+            // In styled mode, use CSS, otherwise the fill used in the style
+            // sheet will take precedence over the fill attribute.
```
TorsteinHonsi, line 456: Is this comment relevant here? Seems to me it should be before line 578...

TorsteinHonsi (changes requested):
Thanks, great work!

* I think we should have this feature as part of the demo pages to. So move the `series-heatmap/interpolation` sample to `highcharts/demo/heatmap-interpolation`. Then do the requested changes. Additionally, add a meaningful series name so the tooltip makes sense.
* We should also have a demo of toggling the `interpolation` option, which we can refer to from the doclet for the option. This will make it clear to viewers what is the difference between interpolated and not. You can base that on the study, but simplify it a bit by removing the change data stuff.
* We also need a paragraph [in the docs article](https://www.highcharts.com/docs/chart-and-series-types/heatmap) about this feature.

samples/highcharts/series-heatmap/interpolation/demo.js
MarkusBarstad: Alright! I will probably go for a `<pre>` tag. 

One thing though: Is it ok if I use this to generate the random data or is there some other dataset we should use instead?

TorsteinHonsi: > Is it ok if I use this to generate the random data

Absolutely 👍

samples/highcharts/series-heatmap/interpolation/demo.js
MarkusBarstad: If i read the docs correct I do not need to replace the "@_sha_" with the commit hash every commit right?

It does not show up in utils for me, hashed or not, however I can see from the requests that it is found, so I assume its correctly placed! (I have tested it outside of utils just to make sure the points align with the screenshot as well, at multiple sizes)

TorsteinHonsi: It is there, but the heatmap series is covering it because it uses solid colors. A simple fix for that is to set an `opacity` on the series:

```ts
    series: [{
        name: 'User activity',
        data: JSON.parse(document.getElementById('data').innerText),
        interpolation: true,
        opacity: 0.4
    }]
```

Alternatively, you could defined the color stops as rgba colors. That would allow individual opacity for each stop.

MarkusBarstad: You are right I see it!

Regarding rgba:
I had some wrong assumptions about alpha in the `colorAxis` because I only tried hex-codes containing alpha-values, and not RGBA strings, so I "turned it off" both for the CTX and when handling the imageData.

I turned alpha back on now, and it is working somewhat now, but I am not sure it works quite as intended.

MarkusBarstad: @TorsteinHonsi 

Apparently alpha-values for the ImageData-object is pre-multiplied by a global value.

Still not totally a 100% on the math of this, but essentially this means that we cannot have individual opacity per pixel using the imageData-object. As far as I can tell this is a limit of the spec.

Should we switch back to constructing the image by drawing with ctx?

https://stackoverflow.com/questions/23497925/how-can-i-stop-the-alpha-premultiplication-with-canvas-imagedata

TorsteinHonsi: I think I need this spelled out in details. Does that mean we need alpha to be either 0 or 1?

> Should we switch back to constructing the image by drawing with ctx?

Isn't that [what we're doing](https://github.com/highcharts/highcharts/blob/heatmapinterpolation/ts/Series/Heatmap/HeatmapSeries.ts/#L478)?

TorsteinHonsi: I just tested with [our original POC](https://jsfiddle.net/highcharts/90qm1erd/), and modifying the alpha channel in the ImageData object works well there. On line 43, the alpha channel is set to 255, but you can see that the opacity of the output changes is you set it to, say, 128 or other semi-transparent values...

MarkusBarstad: Yes, you are entirely correct.

Essentially I forgot to take the alpha value and multiply it times 255, then I found those posts talking about a spec limit for alpha. Sorry about  that!

Latest push has different alphas for different stops [working in the demo](https://utils.highcharts.local/samples/#view/highcharts/demo/heatmap-interpolation) (not surte if this is the correct way to link to utils)

TorsteinHonsi (changes requested):
Thanks! We're closing in :). Now some nitpicking on the [main demo](http://localhost:3030/samples/view?path=highcharts/demo/heatmap-interpolation)

* The aspect ratio is broken. Can you experiment a bit with setting the `chart.height` to a percentage? It looks to me that around `'80%'` is an acceptable value at large screen sizes. But on the mobile it gets too low, maybe we need a responsive rule to deal with smaller screens.
* The tooltip: What do the numbers mean? Maybe we don't need to spell out the screen coordinates, just say something like "5 interactions in this region" or "5 clicks in this 10x10 px square"? 
* The legend: What do the numbers mean? Basically the same as the tooltip issue. What are we looking at? Clicks per square 10px? We can communicate that in a `legend.label`.
* The legend: The color scale is red from about half way up. Are we missing some data or should the scale be adjusted?
* The x axis: Do we need the numbers? I don't think so. We hid them on the y axis.
* We should probably make the higher-order colors semi-opaque too, so that we can see the page below. But not as transparent as the low-activity area.
* Lastly, can we generate sample data with a bit higher density? Currently we can see the blocks shining through - it looks like a scaled-up image, which it actually is. But if we have more data we can mask that effect.

TorsteinHonsi:
> Should we have a legend-title instead?

Yes, sorry a legend title was what I meant.

The data looks a bit weird with the noise. Like in the real world, users wouldn't click on or stare at the empty areas at all. Is it possible to just increase the resolution of the generator function?

MarkusBarstad:
> The data looks a bit weird with the noise. Like in the real world, users wouldn't click on or stare at the empty areas at all. Is it possible to just increase the resolution of the generator function?

Not entirely sure what you been by increasing the resolution of the generator function! 

I have removed the noise (my commit is wrongly titled "less random noise") and added a cluster for the "Try our venn diagram"-button present in the screenshot. 

There are still some things I can try to make it less blocky without raising the `turboThreshold`, as well as having some noise around the relevant parts, will look at it.

MarkusBarstad:
Looks hopefully a bit better now. 

It is actually less points than the previous but I suspect the clusters I was trying to create made the blocky look. 

Also no noise here, only relevant spots have data. If we should have some noise, feel free to give the word!

Also note:
I noticed that the tooltip felt a bit "trapped" to the plot area, noticeable when one was hovering the points towards the top, so I have simply added `followPointer` to the tooltip in the demo.

TorsteinHonsi (changes requested):
Thanks!

The positioning of the shapes seems to be a bit off, see https://jsfiddle.net/3ukm16gy/ . This also causes the tooltip not to report correctly.

MarkusBarstad:
> Thanks!
> 
> The positioning of the shapes seems to be a bit off, see https://jsfiddle.net/3ukm16gy/ . This also causes the tooltip not to report correctly.

@TorsteinHonsi 
I think I have found the cause of the discrepancy.

The extra row of space was caused by `endOnTick` being true for the `yAxis`. 

Since the interpolated image is just stretched to fit `plotBackground`, it does not receive any padding caused by `endOfTick`. 

I suspect this was not apparent in the other demos, because we got lucky with the last ticks.

I have turned `endOfTick` off for this demo, and the tooltip is now only using a formatter as you suggest!

TorsteinHonsi:
> I have turned endOfTick off for this demo

That's cheating 😉  We cannot assume `endOnTick` is false. Instead we must fix the interpolation logic so that positioning is correct regardless of the `endOnTick` and `startOnTick` options. Especially, I suspect you should look at axis `min` and `max` instead of `dataMin` and `dataMax` here:

https://github.com/highcharts/highcharts/blob/b8d7c7774ca0b9ce92d9b8fa6c1c001c16bdc77e/ts/Series/Heatmap/HeatmapSeries.ts#L484-L492

MarkusBarstad:
Hey @Torstein! 

I agree with not cheating 😆 however, the rounded min/max values are not present when the series first loads!

Is there a way we could trigger this calculation earlier for the heatmap-series, or should we just maybe do the same rounding calculation in the heatmap-series?

TorsteinHonsi:
Some remaining glitches, taking notes here as a reminder to check them again:
 - [ ] In [the main demo](http://localhost:3030/samples/#view/highcharts/demo/heatmap-interpolation), if I set `interpolation: false`, the positioning is different. Looks like it has something to do with responsive rules and redrawing.
 - [ ] In the [series-heatmap/interpolation](http://localhost:3030/samples/view?path=highcharts/series-heatmap/interpolation) demo, the canvas does not extend out to the edges, so the interpolated version appears narrower than the non-interpolated. It probably needs to be widened by half the `colsize` in each direction.
 - [ ] Reversed axis not considered, both for x and y axis: https://jsfiddle.net/highcharts/q3w9g12z/

MarkusBarstad:
Should have re-requested review on this one, sorry about that! 

- positioning is still different when responsive is present, without it both normally drawn points and interpolated align,
- [series-heatmap](http://localhost:3030/samples/view?path=highcharts/series-heatmap/interpolation) should be fixed,
- inverse & reverse (and any combination of them) should be functioning

I ended up basically building a function to build the interpolated image based on inversion/reversion/scale. Just tell me if we should have it be more conventional. My idea was that if we are to handle a lot of points, better to just make a function once, and try to limit conditional checks.

TorsteinHonsi:
> Should have re-requested review on this one, sorry about that!

Not yet, first we need to resolve the [remaining issues](https://github.com/highcharts/highcharts/pull/18296#issuecomment-1506471060). The two first check boxes are not resolved. The last one is partly resolved, but as you can see if you toggle the interpolation box, there is an offset when enabling interpolation.

TorsteinHonsi:
> positioning is still different when responsive is present, without it both normally drawn points and interpolated align

@MarkusBarstad I took a closer look at this, turns out it is the Y axis that jumps up one tick. So not related to interpolation. I set `endOnTick: false` in the demo.

MarkusBarstad:
> @MarkusBarstad I took a closer look at this, turns out it is the Y axis that jumps up one tick. So not related to interpolation. I set `endOnTick: false` in the demo.

Alright, good to know, @TorsteinHonsi !

About that fiddle however (and I could be missing something but); 

It links to an old commit `334f5f97f3`, more recent commits should have the image width be similar to normal points. When the axis has padding though, its off by a few pixels in each direction. Should probably use col as suggested.

It also has reversion (but it flips in the opposite direction, will fix it)

TorsteinHonsi:
> About that fiddle however

I updated it with the latest commit: https://jsfiddle.net/highcharts/q3w9g12z/1/ . What we see now when we toggle on and off Interpolation is that the heatmap is shifted one `colsize`. This happens when `xAxis.reversed` is true. 

https://user-images.githubusercontent.com/227836/233268670-b336fe90-6b23-4272-ab49-16f48d7b897f.mov

In addition, `chart.inverted` now fails, see https://jsfiddle.net/highcharts/q3w9g12z/2/. 

In short, all combinations of `chart.inverted` and x and y axis `reversed` must be tested specifically.

PR #18311, opened 2023-01-16
db/chart-type-dropdown

sebastianbochan:
- [x] - refactor renderSelect
- [x] - update langpack
- [x] - added select to sidebar (based on options)
- [x] - exclude map
- [x] - set chart's type in select
- [x] - merge chart options
- [x] - tests when redraw will be fixed

### DIFF
--- a/ts/Dashboard/EditMode/Sidebar.ts
+++ b/ts/Dashboard/EditMode/Sidebar.ts
@@ -978,12 +978,12 @@ class Sidebar {
                     ) {
                         (updatedSettings as any).chartOptions =
                             merge(
+                                (updatedSettings as any).chartOptions,
                                 {
                                     chart: {
                                         type: (chartType[0] as any).value
                                     }
-                                },
-                                (updatedSettings as any).chartOptions
+                                }
                             );
                     }
                 } catch {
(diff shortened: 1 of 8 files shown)


ts/Dashboard/EditMode/Sidebar.ts
```diff
@@ -927,6 +955,8 @@ class Sidebar {
             activeTab.contentContainer.querySelectorAll(
                 'input, textarea'
             ) || [];
+        const chartType = activeTab &&
```
goransle, line 958: `querySelector('#chartType option[selected]')` might make things simpler here

ts/Dashboard/EditMode/Sidebar.ts
```diff
@@ -939,6 +969,23 @@ class Sidebar {
                     (updatedSettings as any)[fieldId] = JSON.parse(
                         (formFields[i] as HTMLTextAreaElement).value
                     );
+
+                    if (
+                        fieldId === 'chartOptions' &&
+                        (updatedSettings as any).chartOptions &&
```
goransle, line 975: Looks like this evaluates to false currently. Might be due to chartOptions being before the chart type in the order?

ts/Dashboard/EditMode/Sidebar.ts
sebastianbochan: Makes sense, thanks!

PR #18325, opened 2023-01-19
db/climate-colors

bre1470:
More style adjustments in city chart, datagrid and KPI. It is not done, but we should have first some KPI fixes before continue with styles.


karolkolodziej (approved):

pawelfus (commented):
Looks good!

Two questions though:
- you mention datagrid in PR, but it seems quite different from the designs. Will there be a separate PR for this?
- ~will Gauge charts be part of the other PR as well? Or missing gauges are one of the KPI issues? (I don't see anything related to it in Asana)~ - yes, it is

samples/dashboard/demos/climate/demo.css
```diff
@@ -118,19 +209,106 @@
...
+.hd-component-kpi > *,
+.hd-component-kpi-value-wrap > * {
+    align-self: center;
+    flex: none;
+    margin: auto;
+}
+
+.hd-component-kpi {
+    background-color: var(--color-background) !important;
+    color: var(--color-text) !important;
```
sebastianbochan, line 229: Please do not use `!important` anywhere, you can add the extra parent class instead.

sebastianbochan (changes requested):

samples/dashboard/demos/climate/demo.css
bre1470: Why is the color set dependent on whether the component is in a cell or not? Shouldn't this play no role?

samples/dashboard/demos/climate/demo.css
```diff
@@ -118,19 +209,122 @@
...
+    background-color: var(--color-header-background);
+    border-color: var(--color-header-border);
+    color: var(--color-header-text);
+}
+
+#selection-grid .hc-dg-outer-container {
+    background-color: var(--color-background);
+    color: var(--color-row-text);
+    position: relative;
+    top: 0 !important;
```
sebastianbochan, line 301: The `!important` should be removed, you can change the structure of selector.

sebastianbochan (changes requested):

samples/dashboard/demos/climate/demo.css
sebastianbochan: It was the only suggestion and one of the possibilities. The most important thing is to not use `!important` at all ;)

bre1470: These all should work without important, but they do not. IMHO the default design should have simpler selectors when it comes to coloring.

sebastianbochan: I absolutely agree :)

samples/dashboard/demos/climate/demo.css
bre1470: It has to change an element style, so this is needed. In general I do not see a problem with `!important` flags in the demo as these are end-user CSS and not part of any distribution.

sebastianbochan: The demo is a kind of showcase that would be also visible to developers. 

I can not agree that we should ignore best practices here.

bre1470:
@pawelfus I made adjustments to the data grid. Please review.

samples/dashboard/demos/climate/demo.css
sebastianbochan: The demo is a kind of showcase that would be also visible to developers.

I can not agree that we should ignore best practices here.

sebastianbochan (changes requested):

samples/dashboard/demos/climate/demo.css
sebastianbochan: The demo is a kind of our show case and developers will have access to it too. 

Sorry, I cannot agree that we should use `!important` and break the basic rules of using CSS.

sebastianbochan: The demo is a kind of our show case and developers will have access to it too. 

Sorry, I cannot agree that we should use `!important` and break the basic rules of using CSS.

sebastianbochan (changes requested):

samples/dashboard/demos/climate/demo.js
```diff
@@ -1,12 +1,17 @@
...
-//     TX: 'Maximal temperature'
-// };
 const dataPool = new Dashboard.DataOnDemand();
+const dataScopes = {
+    FD: 'Days with Fog',
+    ID: 'Days with Ice',
+    RR: 'Days with Rain',
+    TN: 'Average Temperature',
+    TX: 'Maximal Temperature'
+};
```
pawelfus, line 9: I see you defined nice names - could we use them in DataGrid too? Instead of abbreviations? 

If it's something more time-consuming than ~15min, I will create a ticket, just let me know, please

pawelfus (changes requested):
Minor nitpicking, otherwise looks good!

samples/dashboard/demos/climate/demo.js
```diff
@@ -644,52 +647,56 @@ async function setupDashboard() {
...
+                    }],
+                    style: {
+                        height: '414px'
+                    }
                 }]
             }]
         }
     });
+
+    console.log(dashboard);
```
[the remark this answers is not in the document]
bre1470: I like to keep this for debugging until we are done with the demo.

samples/dashboard/demos/climate/demo.js
bre1470: Please create a ticket. This needs the DataGrid to take column options and we need to create column aliases in the table.

sebastianbochan (approved):

pawelfus (approved):
Thanks!

PR #18353, opened 2023-01-24
db/climate-loading

bre1470:
This PR also fixes 10 ghost series in the city chart.


pawellysy (approved):
Looks good!

samples/dashboard/demos/climate/demo.js
```diff
@@ -749,9 +745,43 @@ async function convertTemperature(city) {
...
+                citiesMap.setData(await buildCitiesMap());
+            }
+        }
+    }
+}
+
 async function main() {
     await setupDataPool();
     await setupDashboard();
+    await delay(750); // wait for animations to finish
```
sebastianbochan, line 783: 750 is an animation time from the chart? If yes, can we extract it from the chart options or set it as global?

bre1470: Not really. It is the timing of rendering plus animation and covers two charts - navigator and city. If we like to be exact then the follow up setupCitiesData needs to run after both chart complete callbacks. The best would be an animation finished callback though. I will create a promise wrap.

karolkolodziej (approved):
It looks good altho there is one issue- see the screen below (sorry for the quality- it is from the Performance tab).
There is a brief moment when the map does have data and the chart does not.
Should we do something to avoid this?

image

bre1470:
@karolkolodziej I am working on this. It is caused by the `afterSetExtremes` event emitting after initial chart renders.

samples/dashboard/demos/climate/demo.js
```diff
@@ -26,7 +26,7 @@ let cityScope = defaultCity;
 let citySeries;
 let dataScope = defaultData;
 let navigatorSeries;
-let worldDate = new Date(Date.UTC(2010, 11, 25));
+let worldDate = 1293235200000;
```
sebastianbochan, line 29: I think that the `Date.UTC(2010, 11, 25)` would be more readable.

bre1470: I write dates always as timestamps and divide it in my mind with 86400000. jk

It's a debug leftover. I will restore. Thanks!

sebastianbochan: I'm really impressed :D

sebastianbochan (approved):

pawelfus (changes requested):
Minor bug, I hope it's not cache or anything like this again 😅 
```
From github.com:highcharts/highcharts
 * branch                  db/climate-loading -> FETCH_HEAD
Already up to date.
➜  highcharts git:(db/climate-loading) npx gulp scripts                  
[17:40:52] Using gulpfile ~/Projects/highcharts/gulpfile.js
[17:40:53] Starting 'scripts'...
[17:40:53] ✓ Source code has not been modified since the last successful run.
[17:40:53] Hint: Run the `scripts-watch` task to watch the js and ts directories.
[17:40:53] Finished 'scripts' after 658 ms
```

Demo: http://utils.highcharts.local/samples/#view/dashboards/demos/climate

Steps: Open demo above and dev tools


Error: 
image

pawelfus (approved):
Thanks!

PR #18354, opened 2023-01-24
bugfix/18305-custom-marker-size-update

### LINKED ISSUE #18305: Marker with custom image symbol doesn't update when there's no initial width/height
#### Expected behaviour
Marker should update it's width and height, even when these properties are not specified initially.
Related bug: #18222

https://jsfiddle.net/BlackLabel/w9516epm/

#### Product version
Highcharts 10.3.2 (latest)

kamil-musialowski:
Fixed #18305, updating series marker size was not working if the initial size was not set.

Demo for testing: https://jsfiddle.net/BlackLabel/xjeLkv3u/

This bug only happened while updating from a button click. Updating outside of `addEventListener` worked fine, therefore I was not able to create tests for this one.

### DIFF
--- a/samples/unit-tests/series/marker/demo.js
+++ b/samples/unit-tests/series/marker/demo.js
@@ -60,9 +60,35 @@ QUnit.test('Marker should not move after hover(#4586)', function (assert) {
 });
 
 QUnit.test('Marker size and position', function (assert) {
-    var series = Highcharts.chart('container', {
+    let onlyOnce = true,
+        now = false;
+
+    const series = Highcharts.chart('container', {
         chart: {
-            animation: false
+            animation: false,
+            events: {
+                render() {
+                    if (onlyOnce && now) {
+                        onlyOnce = false;
+                        this.series[0].update({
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (changes requested):
Thanks! Although we must assume that the order of the picked elements was there for a reason, this changes seems legit and I can't think of a way it would break. 

We need a unit test for it though. Preferrably find an existing test that deals with markers, and add an `.update` there.

kamil-musialowski:
@TorsteinHonsi I had to add the update and test in a tricky way, since I was not able to do that with a simple `series.update`. 
[This demo](https://jsfiddle.net/BlackLabel/yw836hu0/) explains it a bit further, why a standard update does not throw an error in tests.

TorsteinHonsi (approved):
Thanks, it looks good. 

I don't fully understand why the test has to be inside the `render` callback. But I understand this problem that the image size is cached the first time the image is loaded, then the size is read from cache after that.

PR #18360, opened 2023-01-24
Add a line chart

mekhatria:
A line chart for a blog

### DIFF
--- a/samples/highcharts/blog/line-chart/demo.details
+++ b/samples/highcharts/blog/line-chart/demo.details
@@ -1,5 +1,5 @@
 ---
- name: Line chart to replace a chart race
+ name: Line chart with layout set to proximate
  authors:
    - Mustapha Mekhatria
  js_wrap: b
--- a/samples/highcharts/blog/line-chart/demo.details
+++ b/samples/highcharts/blog/line-chart/demo.details
@@ -1,0 +1,6 @@
+---
+ name: Line chart to replace a chart race
+ authors:
+   - Mustapha Mekhatria
+ js_wrap: b
+...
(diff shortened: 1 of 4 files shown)


samples/highcharts/blog/line-chart/demo.details
```diff
@@ -0,0 +1,6 @@
+---
+ name: Line chart to replace a chart race
```
TorsteinHonsi, line 2: What does "Line chart to replace a chart race" mean? This is a static line chart, not a race?

samples/highcharts/blog/line-chart/demo.js
```diff
@@ -0,0 +1,168 @@
...
+            {
+                condition: {
+                    maxWidth: 500
+                },
+                chartOptions: {
+                    legend: {
+                        layout: 'horizontal',
+                        align: 'center',
+                        verticalAlign: 'bottom'
+                    }
```
TorsteinHonsi, line 163: This responsive rule actually doesn't work. First of all, it needs to set `chart.marginBottom` to `undefined` in order to have space for the legend. But even then, it seems HC is not able to switch from `layout: 'proximate'` to `layout: 'horizontal'`. That's a bug, and I'll investigate.

My suggestion is that we remove the responsive rule. The legend looks fine on mobile with the initial settings too in my opinion.

TorsteinHonsi (changes requested):
Thanks!

samples/highcharts/blog/line-chart/demo.js
TorsteinHonsi: Related issue https://github.com/highcharts/highcharts/issues/18362

TorsteinHonsi (approved):
Thanks!

PR #18376, opened 2023-01-26
bugfix/18312-tooltip-split

### LINKED ISSUE #18312: Split tooltip is not supported when chart is inverted
According to the following snippet (from the `Tooltip.ts` file), the API docs should be updated to indicate that the `tooltip.split` is not supported when the chart is inverted.

`this.split = options.split && !chart.inverted && !chart.polar;`

jakubSzuminski:
Fixed #18312, added info to the API about `tooltip.split` not working for `polar` & `inverted` charts


pawellysy (approved):
Looks good!

TorsteinHonsi (approved):
Thanks!

PR #18381, opened 2023-01-26
bugfix/16675-custom-annotations-button

### LINKED ISSUE #16675: NavigationBindings: Cannot create a custom button
#### Expected behaviour
When creating the button as SVG shape it should work normally. 

#### Actual behaviour
There is an error, which is called by `bindings.button.className.indexOf('highcharts-disabled-btn') === -1`

https://jsfiddle.net/BlackLabel/by126xor/
Click on the button next to the exporting button

#### Product version
v9.3.1

#### Affected browser(s)
all

kamil-musialowski:
Fixed #16675, adding annotation from a custom button was not possible, if the button had a custom SVG symbol.

### DIFF
--- a/samples/unit-tests/annotations/annotations-events/demo.js
+++ b/samples/unit-tests/annotations/annotations-events/demo.js
@@ -4,6 +4,7 @@ QUnit.test('Annotations events - general', function (assert) {
         removeEventCalled = 0,
         closeEventCalled = 0,
         circleAfterUpdateCalled = 0,
+        customButtonClicked = 0,
         popupOptions,
         getShapes = function () {
             return [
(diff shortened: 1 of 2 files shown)


OldBuddy:
@kamil-musialowski maybe better to reuse [classList](https://developer.mozilla.org/en-US/docs/Web/API/Element/classList) instead?

kamil-musialowski:
@OldBuddy, thank you for your suggestion and contribution. It's a great idea! I've added the requested changes.

OldBuddy:
@kamil-musialowski thanks! what about reusing [existing utility method](https://github.com/highcharts/highcharts/blob/master/ts/Accessibility/Utils/HTMLUtilities.ts#L59)?

TorsteinHonsi (approved):
Thanks!

PR #18384, opened 2023-01-27
Migration: MACDIndicator

marvin19:
Improved MACDIndicator and MACDOptions, also checked MACDPoint.


bre1470 (changes requested):
Some class options have undefined type and therefor can be optional without `= void 0` assignment.

bre1470 (approved):

TorsteinHonsi (approved):

PR #18387, opened 2023-01-27
WIP component redraw

goransle:


ts/Dashboards/Component/Component.ts
```diff
@@ -518,6 +569,18 @@ abstract class Component<TEventObject extends Component.EventTypes = Component.E
 
     public setTitle(titleOptions: Component.TextOptionsType): void {
         const previousTitle = this.titleElement;
+
+        if (
+            !titleOptions || typeof titleOptions === 'string' ?
```
karolkolodziej, line 574: You are using it in two places, each has different formatting.
How about extracting that logic somewhere outside?

ts/Extensions/DashboardPlugins/HighchartsComponent.ts
```diff
@@ -214,6 +220,18 @@ class HighchartsComponent extends Component<HighchartsComponent.ChartComponentEv
         Component.addInstance(this);
     }
 
+    public setOptions(): void {
```
karolkolodziej, line 223: Please add docklet

ts/Extensions/DashboardPlugins/HighchartsComponent.ts
```diff
@@ -420,9 +439,9 @@ class HighchartsComponent extends Component<HighchartsComponent.ChartComponentEv
 
     private initChart(): Chart {
         if (this.chart) {
-            if (this.chart.series.length) {
-                return this.chart;
-            }
+            // if (this.chart.series.length) {
```
karolkolodziej, line 442: Remove if not using

karolkolodziej (changes requested):

samples/cypress/dashboards/dashboard-layout/demo.mjs
```diff
@@ -115,6 +116,7 @@ let dashboard = new Dashboard('container-nested-layout', {
         chartOptions: {
             chart: {
                 animation: false
+                // type: 'line'
```
sebastianbochan, line 119: I would remove this line, its unecessary

sebastianbochan (changes requested):

ts/Dashboards/Component/Component.ts
goransle: Think this will have to be for the backlog

karolkolodziej (approved):

pawellysy (approved):

PR #18388, opened 2023-01-27
db/kpi-design

bre1470:


bre1470:
There are two issues left afaik:
- The styledMode CSS prevents colored gauge because of `highcharts-color-0`
- I do not get the yAxis labels to be activated. My reference was this demo: https://www.highcharts.com/demo/gauge-solid

samples/dashboards/demos/climate/demo.css
```diff
@@ -209,31 +201,79 @@
...
+    stroke: var(--color-pane-line);
+}
+
+#kpi-layout .hd-component-kpi-chart-container .highcharts-series .highcharts-color-0 {
+    fill: var(--color-gauge);
+    stroke: none;
+}
+
+#kpi-layout .hd-component-kpi-chart-container .highcharts-title {
+    font-size: 13px;
```
karolkolodziej, line 244: I feel like the `font-size` might be slightly bigger. I can't see that in figma.

karolkolodziej (approved):

samples/dashboards/demos/climate/demo.css
bre1470: Probably, but we are out of space anyway.

sebastianbochan (approved):

pawelfus (approved):
Indeed, something strange with these labels. We go with what we have, thanks!

Please merge in 👍

PR #18390, opened 2023-01-27
db/chart-type-dropdown-rebased

goransle:
- [x] - refactor renderSelect
- [x] - update langpack
- [x] - added select to sidebar (based on options)
- [x] - exclude map
- [x] - set chart's type in select
- [x] - merge chart options
- [x] - tests when redraw will be fixed
- [x] - fix climate demo seriesType


sebastianbochan (approved):

ts/Dashboards/EditMode/Sidebar.ts
```diff
@@ -831,17 +831,45 @@ class Sidebar {
...
+                        const chartTypesEnum: Array<string> = [
+                            'column',
+                            'line',
+                            'scatter',
+                            'spline',
+                            'pie'
+                        ];
+
+                        const chartOpts =
+                            (currentComponent as any).chartOptions;
```
karolkolodziej, line 850: Can we check if the `currentComponent` is a Highcahts Component that has `chartOptions`

ts/Dashboards/EditMode/Sidebar.ts
```diff
@@ -831,17 +831,45 @@ class Sidebar {
...
                 activeTab && activeTab.content.container;
             let type;
+            let chartTypes = {};
 
             for (const key in componentSettings) {
                 if (componentSettings[key]) {
                     type = componentSettings[key].type;
 
+                    if (key === 'chartType') {
+                        const chartTypesEnum: Array<string> = [
```
karolkolodziej, line 841: Why not use `enum`?
Also, there is no need to declare that type here.

ts/Dashboards/EditMode/Sidebar.ts
```diff
@@ -955,6 +985,23 @@ class Sidebar {
                     (updatedSettings as any)[fieldId] = JSON.parse(
                         (formFields[i] as HTMLTextAreaElement).value
                     );
+
+                    if (
+                        fieldId === 'chartOptions' &&
+                        (updatedSettings as any).chartOptions &&
```
karolkolodziej, line 991: Can we get rid of this `any` casting?

karolkolodziej (changes requested):

ts/Dashboards/EditMode/Sidebar.ts
```diff
@@ -27,6 +27,8 @@ import EditRenderer from './EditRenderer.js';
 import Bindings from '../Actions/Bindings.js';
 import GUIElement from '../Layout/GUIElement.js';
 import Layout from '../Layout/Layout.js';
+import type Component from '../Component/Component.js';
```
karolkolodziej, line 30: Can be removed, also could you please separate the `type` imports from the other imports?

ts/Dashboards/EditMode/Sidebar.ts
```diff
@@ -838,22 +840,27 @@ class Sidebar {
...
                             'column',
                             'line',
                             'scatter',
                             'spline',
                             'pie'
                         ];
 
-                        const chartOpts =
-                            (currentComponent as any).chartOptions;
+                        // eslint-disable-next-line
```
karolkolodziej, line 851: How about instead of casting it to a certain type, check it like this.
```TS
                        if (currentComponent instanceof HighchartsComponent) {
                            ... your code
                        }
```

ts/Dashboards/EditMode/Sidebar.ts
```diff
@@ -988,13 +995,13 @@ class Sidebar {
 
                     if (
                         fieldId === 'chartOptions' &&
-                        (updatedSettings as any).chartOptions &&
+                        (updatedSettings as HighchartsComponent.ComponentOptions).chartOptions && // eslint-disable-line
```
karolkolodziej, line 998: Similar here

karolkolodziej (changes requested):

ts/Dashboards/EditMode/Sidebar.ts
sebastianbochan: It's surprising, because the linter skipped it :D 

Thanks! I will fix it!

ts/Dashboards/EditMode/Sidebar.ts
bre1470: If this is done, we have Highcharts in the Dashboard bundle. Maybe this should become an event somehow and HighchartsComponent should listen to the Sidebar to do whatever is necessary.

sebastianbochan: Events in the sidebar sound good, but I would like to refactor the sidebar in the next step when we will design all controllers inside.

Any suggestions which solution is better: use `instanceof` or left like it is.

bre1470 (approved):
A bigger refactoring will be done in another PR. So I am fine with the temporary type issues. Most important is to keep Highcharts out of Dashboards bundle.

ts/Dashboards/EditMode/Sidebar.ts
bre1470: For me it is fine as it is a temporary thing.

karolkolodziej:
Yes, this can be done in a separate PR ;)

ts/Dashboards/EditMode/Sidebar.ts
```diff
@@ -14,6 +14,8 @@
  *
  * */
 
+import type HighchartsComponent from '../../Extensions/DashboardPlugins/HighchartsComponent.js';
```
karolkolodziej, line 17: There is no need to add `.js` if you are importing just types.

karolkolodziej (approved):
One small thing otherwise looks good :)

pawellysy (approved):
LGTM!

PR #18391, opened 2023-01-27
db/fullscreen-incorrect-offsets

sebastianbochan:


pawellysy (approved):
LGTM!

ts/Dashboards/Actions/Resizer.ts
```diff
@@ -181,7 +181,7 @@ class Resizer {
         const minHeight = options.styles.minHeight;
         const snapWidth = this.options.snap.width || 0;
         const snapHeight = this.options.snap.height || 0;
-        const dashboardContainer = this.editMode.dashboard.container;
+        const dashboardsContainer = this.editMode.dashboard.container;
```
karolkolodziej, line 184: IMO this should be singular because we are referring to a specific instance of the Dashboards.

ts/Dashboards/Globals.ts
```diff
@@ -59,7 +59,8 @@ namespace Globals {
         cellHover: classNamePrefix + 'cell-state-hover',
         cellActive: classNamePrefix + 'cell-state-active',
         row: classNamePrefix + 'row',
-        layoutsWrapper: classNamePrefix + 'layouts-wrapper'
+        layoutsWrapper: classNamePrefix + 'layouts-wrapper',
+        dashboardsContainer: classNamePrefix + 'dashboards-wrapper'
```
karolkolodziej, line 63: Here it looks correct!

karolkolodziej (changes requested):
There are still some leftovers from the `highcharts-dashboards` class especially in CSS

ts/Dashboards/Actions/Resizer.ts
sebastianbochan: Good point, I've started it before renaming.

karolkolodziej (approved):

PR #18393, opened 2023-01-27
db/click-selected-city

karolkolodziej:
Added return when clicking on the selected city.

### DIFF
--- a/samples/dashboards/demos/climate/demo.js
+++ b/samples/dashboards/demos/climate/demo.js
@@ -239,6 +239,10 @@ async function setupDashboard() {
                             const point = e.point;
                             const city = point.name;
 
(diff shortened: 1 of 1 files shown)


sebastianbochan:
What is the reason for it? Any side effect?

karolkolodziej:
@sebastianbochan without that, if you click on the already selected city, the series in the navigator disappear.

sebastianbochan:
Does `e.preventDefault` or `stopPropagation` not solve the problem?

karolkolodziej:
No, in this case, `cityScope === city`, we don't want to run the callback at all.

sebastianbochan (approved):

pawellysy (approved):

PR #18397, opened 2023-01-27
bugfix/18238-vbp-errors-improvement

### LINKED ISSUE #18238: VBP Overlay Indicator Doesn't Work After Resize
#### Expected behaviour
I expect to be able to resize browser, trigger state changes, and re-render without app crashing

#### Actual behaviour
App crashes when resizing and re-rendering VBP indicator series

https://codesandbox.io/embed/highcharts-react-example-forked-9o6l9p?fontsize=14&hidenavigation=1&theme=dark

#### Product version
highcharts: 10.3.2
highcharts-react-official 3.1.0

#### Affected browser(s)
Chrome

kamil-musialowski:
Fixed #18238, VBP indicator was not throwing an error when volume series had no data, or didn't exist.

### DIFF
--- a/ts/Stock/Indicators/VBP/VBPOptions.d.ts
+++ b/ts/Stock/Indicators/VBP/VBPOptions.d.ts
@@ -42,7 +42,7 @@ export interface VBPOptions extends SMAOptions {
 
 export interface VBPParamsOptions extends SMAParamsOptions {
     ranges?: number;
-    volumeSeriesID?: string;
+    volumeSeriesID: string;
     // for inheritance
 }
 
--- a/samples/unit-tests/indicator-volume-by-price/recalculations/demo.js
+++ b/samples/unit-tests/indicator-volume-by-price/recalculations/demo.js
@@ -272,3 +272,76 @@ QUnit.test('Test algorithm on data updates.', function (assert) {
         'Indicator is removed after series remove.'
     );
 });
+
+QUnit.test('VBP series errors.', function (assert) {
+    function createChart(volumeSeries, vbpSeries) {
+        const chart = Highcharts.chart('container', {
+            series: [{
+                id: 'quotes',
+                type: 'candlestick',
+                data: [{
+                    x: 1515042000000,
+                    y: 99.6,
+                    high: 100,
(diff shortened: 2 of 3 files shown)


pawelfus (changes requested):
Please, add the test. Apart from that looks good 👍

ts/Stock/Indicators/VBP/VBPOptions.d.ts
```diff
@@ -42,7 +42,7 @@ export interface VBPOptions extends SMAOptions {
 
 export interface VBPParamsOptions extends SMAParamsOptions {
     ranges?: number;
-    volumeSeriesID?: string;
+    volumeSeriesID: string;
```
pawelfus, line 45: I'm curious about this one. This is true: to render the VbP indicator, `volumeSeriesID` must be set. The same for `linkedTo` option.

However, will that affect also `series.update({ ... });`? `volumeSeriesID` there is optional

pawelfus: What do you think @TorsteinHonsi ?

kamil-musialowski: My though process with this one was: 
Since we set [the default value](https://github.com/highcharts/highcharts/blob/d1a37ffe3a085206c6d7b42d13381e3d6d0339db/ts/Stock/Indicators/VBP/VBPIndicator.ts#L154) for `volumeSeriesID`, does it make sense to leave it as an optional property in this interface?

TorsteinHonsi: If TS is happy, I'm happy. The idea with `series.update()` and other update functions is that they tak `DeepPartial<SeriesOptions>`, which allows us to set non-optional properties on the default options. It is not consistently implemented yet, though.

pawelfus (approved):
Thanks! 👍

TorsteinHonsi (changes requested):
Looks like some related tests failed now

TorsteinHonsi (approved):

PR #18401, opened 2023-01-30
db/fix-climate

bre1470:


karolkolodziej:
Sync between `DataGrid` and `city-chart` is restored.
I presume this is going to change when the `cursor` is ready.

pawellysy (approved):
LGTM!

PR #18407, opened 2023-01-31
db/initial-cell-height

sebastianbochan:


ts/Dashboards/Layout/Cell.ts
```diff
@@ -124,7 +124,10 @@ class Cell extends GUIElement {
                 style: merge(
                     layoutOptions.style,
                     rowOptions.style,
-                    options.style
+                    options.style,
+                    {
+                        height: options.height
```
karolkolodziej, line 129: Is height the only property we would like to merge?
What about the `width`?

karolkolodziej (approved):

ts/Dashboards/Layout/Cell.ts
sebastianbochan: The width is added in the function below (reflow) because of
- responsive rules
- a requirement of translation width from 1/3 into 33.333% for instance.

pawellysy (approved):

bre1470 (approved):
Looks good. 👍

PR #18410, opened 2023-01-31
Tools: Validate samples for categories and tags outside demo

TorsteinHonsi:
Because it was easy for our developers to forget to remove meta information for the demo pages when copying samples for other purposes.


goransle (approved):
Looks good!

PR #18413, opened 2023-01-31
db/fix-kpi-sizing-bug

pawellysy:
Updating default charts' height in KPI Component.


pawellysy:
@bre1470 there are some further bugs with this topic, which I think need to overwrite this Highcharts hardcoded value: https://github.com/highcharts/highcharts/blob/5836683ab8d3a7fe3652d7546d5db51ac1aa3bd2/ts/Core/Chart/Chart.ts#L1425

If it will be kept the way it is currently, there are some problems with setting size by component's size by CSS. 

Any insights how to do it by default? My idea would be to move this to some global value, and overwrite it in KPI. 
Is that acceptable?

sebastianbochan (approved):

PR #18414, opened 2023-01-31
db/api-docs

bre1470:


bre1470:
- Run `npx gulp dashboards/api-docs`
- Open `build/api/dashboards-1` & `build/api/dashboards-2`
- `build/api/dashboards-2` has two defaults, you need to open the upper file path to find the class default
   (issue will not be solved, see `https://github.com/TypeStrong/typedoc/issues/1521`)

oysteinmoseng:
Well I found dashboards-1 significantly more usable, UI-wise.

bre1470:
In my opinion:
- TypeDoc is not optimized for default exports anymore, because it organize everything around the keywords/members.
- When class, interface, and namespace share the name, the namespace always wins in the sidebar navigation. You can change the priorities, but you can not have all parts visible in the sidebar.
- Since 0.20.x the option for a global namespace is also gone, so there is no workaround. 
- I agree with Øystein about the UI. IMHO it has become significantly worse since 0.19.0.

pawelfus:
Okay, a few questions:

During the meeting, you mentioned some issues regarding releasing the API docs. Do both solutions solve that issue? And will both provide tree/namespace JSON files?

D1 (aka `dashboards-1`):
- Could we configure Classes/Namespaces to get an overview/list like we currently have? See: https://api.highcharts.com/class-reference/Highcharts.Chart - for Dashboards it might not be a problem, however, for Highcharts we have classes with a long list (eg Chart). With Interfaces, it shouldn't be a problem, I guess?
- Alternatively, can we get a tree structure in the right-hand sidebar? Similar to what we already have
- does it have a built-in searchbox?


D2 (aka `dashboards-2`):
- it surprised me nicely with dark-theme out of the box - somehow it makes me more comfortable 😅 (I have by default dark mode in OS)
- the sidebar overview is a nice feature (see screenshot below)
- comes with a search out-of-the-box
- what is going on with these "default" links under the (sub)module name?
> When class, interface, and namespace share the name, the namespace always wins in the sidebar navigation. You can change the priorities, but you can not have all parts visible in the sidebar.

Do we have something like this in Dashboards or Highcharts? I mean a Class/Interface/Namespace common for all of them?


Screenshots:
D2 with a list of options:
Zrzut ekranu 2023-02-1 o 15 10 24

Current API with sidebar on the right (I needed this, because my brain was screaming from anxiety for a few minutes):
Zrzut ekranu 2023-02-1 o 15 14 26

bre1470:
@pawelfus 
> Okay, a few questions:
> 
> During the meeting, you mentioned some issues regarding releasing the API docs. Do both solutions solve that issue?

Sorry, I do not recall the issue.

> And will both provide tree/namespace JSON files?

Not really, something similar to a namespace JSON file at most. JSDoc needs a JSON theme. TypeDoc can generate it out of the box. Do we need this?

> D1 (aka `dashboards-1`):
> 
> * Could we configure Classes/Namespaces to get an overview/list like we currently have? See: https://api.highcharts.com/class-reference/Highcharts.Chart - for Dashboards it might not be a problem, however, for Highcharts we have classes with a long list (eg Chart). With Interfaces, it shouldn't be a problem, I guess?

This is not possible without changing the theme.

> * Alternatively, can we get a tree structure in the right-hand sidebar? Similar to what we already have

This is not possible without changing the theme.

> * does it have a built-in searchbox?

No.

> D2 (aka `dashboards-2`):
> 
> * it surprised me nicely with dark-theme out of the box - somehow it makes me more comfortable 😅 (I have by default dark mode in OS)
> * the sidebar overview is a nice feature (see screenshot below)
> * comes with a search out-of-the-box
> * what is going on with these "default" links under the (sub)module name?

See my [first comment](https://github.com/highcharts/highcharts/pull/18414#issuecomment-1410885229).

> > When class, interface, and namespace share the name, the namespace always wins in the sidebar navigation. You can change the priorities, but you can not have all parts visible in the sidebar.
> 
> Do we have something like this in Dashboards or Highcharts? I mean a Class/Interface/Namespace common for all of them?

Everywhere. We have to pack everything into the defaults because the Highcharts Assembler supports only default import/export.

bre1470:
BTW I personally find our class reference hard to navigate and not a good example of user navigation.

pawelfus:
> > During the meeting, you mentioned some issues regarding releasing the API docs. Do both solutions solve that issue?
>
> Sorry, I do not recall the issue.

It was about releasing thousands of files on S3. Someone said it might be useful to store API in the database even. 

> > And will both provide tree/namespace JSON files?
> 
> Not really, something similar to a namespace JSON file at most. JSDoc needs a JSON theme. TypeDoc can generate it out of the box. Do we need this?

I don't think we need this, but there are clients who use it. iOS and Android wrappers use those files too, but I guess Christer's project will resolve that in the future?

bre1470:
> > > During the meeting, you mentioned some issues regarding releasing the API docs. Do both solutions solve that issue?
> > 
> > Sorry, I do not recall the issue.
> 
> It was about releasing thousands of files on S3. Someone said it might be useful to store API in the database even.

Oh, that one. Yes, it is an issue for Highcharts, because we have so many different option sets. I do not see this as an issue here, because the options are just recursive. It looks like a lot, but we produce not many interfaces.

PR #18418, opened 2023-02-01
Migration: Breadcrumbs

bre1470:
- Added Breadcrumbs compose
- Moved Breadcrumbs class
- Moved Breadcrumbs options
- Removed internal namespace


TorsteinHonsi (approved):
Thanks!

PR #18419, opened 2023-02-01
db/dark-mode-climate

karolkolodziej:
Fixed dark mode in climate demo.


sebastianbochan (approved):

bre1470 (approved):
Thank you! Looks good. 👍 

BTW I noticed that the layout does not support the system dark mode, but that is another issue.

karolkolodziej:
@bre1470 could you please create a task with a description for that? Thanks ;)

PR #18421, opened 2023-02-01
Samples: combo-timeline demo updates

nasin:
Hi, I updated this demo using the styles from the version on the About Us page on highcharts.com.

#### To do
 - [ ] Torstein: Remove graphics for the flag backgrounds. Instead, define custom symbols. If that fails, use Annotations. 
 - [x] Remove the Google Trends series


TorsteinHonsi:
Thanks! Before I do a detailed review, can you update it with recent data? See http://localhost:3030/samples/#view/highcharts/studies/combo-timeline-updated . And we should fill in the years after 2019. HR can provide the numbers.

nasin:
Hi, I updated the data. I had revenue and employee data from HR already (for the About Us verson of this chart), and I added it, but you should look and see if I did it right because all the employees post August 2019 are nameless in the data I used.

Also, I need updated Google trends data. Should I ask Marketing?

samples/highcharts/demo/combo-timeline/demo.js
```diff
@@ -184,12 +204,87 @@ function onChartLoad() {
...
+        [])
+        .sort((a, b) => a.x - b.x)
+        .map(event => Object.assign(
+            event, {
+                y: (employees += event.accumulate)
+            }
+        ));
+}
+
+const moreEmployees = [
```
TorsteinHonsi, line 245: Looks like employees before 2019 are loaded from the HTML table then this is added for employees after 2019? That's not consistent. Add the new ones to the HTML table instead, if that is how we did it from before.

samples/highcharts/demo/combo-timeline/demo.js
```diff
@@ -445,22 +640,7 @@ const options = {
...
-            { x: Date.UTC(2012, 8, 1), y: 5, name: 'Erik left', image: null },
-            { x: Date.UTC(2012, 8, 15), y: 6, name: 'Anne Jorunn joined', image: 'AnneJorunn' },
-            { x: Date.UTC(2013, 0, 1), y: 7, name: 'Hilde T. joined', image: null },
-            { x: Date.UTC(2013, 7, 1), y: 8, name: 'Jon Arild joined', image: 'JonArild' },
-            { x: Date.UTC(2013, 7, 20), y: 9, name: 'Øystein joined', image: 'Oystein' },
-            { x: Date.UTC(2013, 9, 1), y: 10, name: 'Stephane joined', image: 'Stephane' },
-            { x: Date.UTC(2014, 9, 1), y: 11, name: 'Anita joined', image: 'Anita' },
-            { x: Date.UTC(2014, 10, 27), y: 11, name: ' ', image: null }
-        ]
+        data: getTurnover().concat(moreEmployees)
```
TorsteinHonsi, line 643: Do not concat `moreEmployees`, the newer employees should be added to the HTML table like the others.

TorsteinHonsi (changes requested):
Thanks so far! 

We also want to remove the Google Translate series.

sigrunnu:
I have no idea if I did it right, I think I might just changed it back to the way it was before Nancy's changes..
The shape of the flags are now back to `squarepin`, the only problem is that I did not manage to set the border-radius, so it doesn't look as good.

The 'award' flags are changed to annotations and I am not sure how to add tooltips to them. 
@TorsteinHonsi

TorsteinHonsi:
You mean this one?
> Torstein: Remove graphics for the flag backgrounds. Instead, define custom symbols. If that fails, use Annotations.

I added my name there because I thought I should do it :) But no worries, now I have a better starting point.

But you can still remove the `moreEmployees` array and add the newcomers into the HTML table with the rest.

sigrunnu:
Yes, that one, thought that meant you wrote it.. Well, hopefully the starting point is not worse  :)
I'll do the other stuff then!

sigrunnu:
> But you can still remove the moreEmployees array and add the newcomers into the HTML table with the res

HR mentioned that we are not allowed to use people's names in the demo. Do you still think it is better to update the table as  it won't add anything extra to it? @TorsteinHonsi

TorsteinHonsi (commented):
Thanks @sigrunnu !

This demo now has some features that we can't use in our demo pages, like hard graphics for the flags. And I don't have time to fix it until after v11. 

I think we should go back to the master and apply the updated data to that chart without touching the design in this round. Can you do that? Then we will close this PR and make a new one that only updates the data.

PR #18431, opened 2023-02-02
docs/18278-mapview-reference

### LINKED ISSUE #18278: MapChart or Chart doesn't have a property mapView with a MapView class
#### Expected behaviour
When set extremes for worldMap chart it should zoomIn
For example:
chart.yAxis[0].setExtremes(-7686.975480799574, -7751.509535225241);
chart.xAxis[0].setExtremes(1449.334671000368, 8229.656780260837);

#### Actual behaviour
It doesn't zoomIn.

http://jsfiddle.net/pawel_dalek/qhtdt7xd/

#### Product version
Latest (10.3). But it got broken since version 9.
I confirm it works on version 8.2.2

**UPDATE:** the description of the issue is not valid. Only TS typings and documentation issue. See the thread.

madepiet:
Fixed #18278, MapView/MapChart API.

### DIFF
--- a/ts/Core/Chart/MapChart.ts
+++ b/ts/Core/Chart/MapChart.ts
@@ -50,6 +50,14 @@ declare module './ChartLike'{
  */
 class MapChart extends Chart {
     /**
+     * The map view handles zooming and centering on the map, and various
+     * client-side projection capabilities.
+     *
+     * @name Highcharts.MapChart#mapView
+     * @type {Highcharts.MapView|undefined}
+     */
+
+    /**
      * Initializes the chart. The constructor's arguments are passed on
      * directly.
      *
--- a/ts/Core/Chart/MapChart.ts
+++ b/ts/Core/Chart/MapChart.ts
@@ -50,14 +50,6 @@ declare module './ChartLike'{
  */
 class MapChart extends Chart {
     /**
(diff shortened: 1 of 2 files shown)


ts/Core/Chart/MapChart.ts
```diff
@@ -49,6 +49,14 @@ declare module './ChartLike'{
...
  */
 class MapChart extends Chart {
+    /**
+     * The map view handles zooming and centering on the map, and various
+     * client-side projection capabilities.
+     *
+     * @name Highcharts.MapChart#mapView
+     * @type {Highcharts.MapView|undefined}
+     */
+
```
bre1470, line 59: This should be over the assignment. If JSDoc does not pick it up, then move the doclet to the end of `MapView.ts` followed by `'';`.

bre1470 (changes requested):

ts/Core/Chart/MapChart.ts
bre1470: I mean this assignment: https://github.com/highcharts/highcharts/blob/master/ts/Maps/MapView.ts#L1374

bre1470 (approved):
Thank you! Looks good. 👍

TorsteinHonsi (approved):
Thanks!

PR #18432, opened 2023-02-02
db/a11-module-extract

sebastianbochan:


ts/Dashboards/Accessibility/DashboardsAccessibility.ts
```diff
@@ -0,0 +1,88 @@
...
+ *  Authors:
+ *  - Sebastian Bochan
+ *
+ * */
+
+'use strict';
+
+import Dashboard from '../Dashboard.js';
+
+/* *
```
karolkolodziej, line 18: This comment should be above the import ;)

ts/Dashboards/Accessibility/DashboardsAccessibility.ts
```diff
@@ -0,0 +1,88 @@
...
+    *  Constructor
+    *
+    * */
+
+    public constructor(
+        dashboard: Dashboard
+    ) {
+        this.dashboard = dashboard;
+
+        this.addTabIndexToCells();
```
karolkolodziej, line 48: I presume this module might get bigger and bigger. What do you think of creating a public method `init` where all extra methods like `addTabIndexToCells` will be invoked?

ts/Dashboards/Dashboard.ts
```diff
@@ -26,6 +26,7 @@
 
 import type JSON from '../Core/JSON';
 
+import DashboardAccessibility from './Accessibility/DashboardsAccessibility.js';
```
karolkolodziej, line 29: Should the `DashboardsAccessibility` be a part of the `dashboards`  module or should it be a separate module to include if needed?

ts/Dashboards/Accessibility/DashboardsAccessibility.ts
```diff
@@ -0,0 +1,88 @@
...
+            if (cell && cell.container) {
+                cell.container.setAttribute('tabindex', -1);
+            }
+        }
+    }
+}
+
+namespace DashboardsAccessibility {
+    export interface Options {
+        addTabIndexToCells: Function
```
karolkolodziej, line 78: I don't understand why adding this method here to `Options`. Could you please explain?

karolkolodziej (changes requested):

ts/Dashboards/Accessibility/DashboardsAccessibility.ts
sebastianbochan: Your are right, at this moment its more as a PoC of a class that will be developed and redesign when we will add support for the Edit Mode and other parts.

ts/Dashboards/Dashboard.ts
sebastianbochan: Oystein suggested to keep it In the `Dashboard/Accessibility` directory, so I named it to be intuitive. I'm open for any suggestions.

ts/Dashboards/Accessibility/DashboardsAccessibility.ts
sebastianbochan: The type is unnecessary, so I removed the whole section and commented the `namespace` section for the future.

oysteinmoseng (approved):
No comments from me!

karolkolodziej (approved):

bre1470 (approved):
Looks good. 👍

PR #18437, opened 2023-02-03
Refactor chart.reflow

TorsteinHonsi:
Trying to simplify and optimize for #17951. Let's see how CI goes.


karolkolodziej (approved):
Looks much cleaner thanks ;)

PR #18438, opened 2023-02-03
bugfix/18406-labelposition-any-cast-fix

jakubSzuminski:
Changed `labelPosition.final` to `labelPosition.computed` for Pie & Funnel series + removed casting from one place that caused testing errors.


TorsteinHonsi (changes requested):
Thanks! Just update the failing unit test, then it should be okay. The visual diffs are not related.

TorsteinHonsi (approved):

PR #18440, opened 2023-02-03
bugfix/17790-fixed-duplicate-price-indicators

### LINKED ISSUE #17790: Price indicator is having duplicate indicators
#### Expected behaviour
When I drag the scrollbar in chart navigator, The last price indicator should remain in a static place.

#### Actual behaviour
Its not remaining into a place rather its creating duplicate indicator.

[](https://jsfiddle.net/Subhabrata/p729m65c/)

#### Product version
10.1.0

#### Affected browser(s)
In all browsers

kamil-musialowski:
Fixed #17790, `lastPrice` and `lastVisiblePrice` labels were not destroyed during redraw.

Demo for testing: https://jsfiddle.net/BlackLabel/Lz1pumsr/

### DIFF
--- a/samples/unit-tests/indicator-current-price/recalculations/demo.js
+++ b/samples/unit-tests/indicator-current-price/recalculations/demo.js
@@ -101,13 +101,13 @@ QUnit.test('Price indicator labels rendering and visibility, #14879, #17790.',
                 }
             },
             series: [{
-                data: [1, 2, 1, 2, 1, 2, 1, 2, 1, 2, 1],
+                compare: 'percent',
+                data: [100, 1, 1, 10, 1],
                 lastPrice: {
                     enabled: true,
-                    label: {
-                        enabled: true
-                    }
+                    color: 'red'
                 },
+                // label
                 lastVisiblePrice: {
                     enabled: true,
                     label: {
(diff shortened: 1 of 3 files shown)


samples/unit-tests/indicator-current-price/recalculations/demo.js
```diff
@@ -101,13 +101,13 @@ QUnit.test('The currentPriceIndicator label should be visible, #14879.',
                 }
             },
             series: [{
-                compare: 'percent',
-                data: [100, 1, 1, 10, 1],
```
pawelfus, line 105: This is very risky to change the config of the demo. In this particular case, perhaps `compare` and that dataset was necessary to reproduce the issue.

In general, I recommend using `chart.update()` after the last `assert()`, instead of alternating the initial config.
 
Just tested:
- https://jsfiddle.net/BlackLabel/b1e90d2x/1/ - with the old config: we can observe missing label - correct test
- https://jsfiddle.net/BlackLabel/b1e90d2x/2/ - with the new config: the label is visible - the purpose of the test is lost (the test is still testing something, but we leave the doors open for regressions)

kamil-musialowski: Sorry about that! I went too far with the unit tests simplification idea. I've restored the old test and added a series update like you suggested. Thank you for spotting that!

pawelfus (approved):
Nice, thank you!

TorsteinHonsi (approved):

PR #18442, opened 2023-02-06
Correct typo and container max-width

mekhatria:
Correct the subtitle title of [donut race chart](https://www.highcharts.com/samples/highcharts/demo/donut-race), and add a max-width for [line chart ](https://www.highcharts.com/samples/highcharts/blog/line-chart).


TorsteinHonsi (approved):
Thanks!

Issue #18433, opened 2023-02-02
Add point.percentage property for solidgauge chart type


vitaliy-kretsul:
This property is currently supported for stacked charts or pies. It would be convenient to have it for solid gauge chart type as well, considering that displaying gauge as a percentage is a pretty common use case.

karolkolodziej:
Hi @vitaliy-kretsul. 
Could you please elaborate on how the percentage value should be calculated for the `solidgauge` points?

In `pie` it is rather straightforward, you check the `point.value / sumOfPointsInPie` but I don't see how it should work in the `solidgauge` series.

TorsteinHonsi:
I imagine it would make sense to calculate the percentage on the `yAxis` like this: https://jsfiddle.net/highcharts/fcwL4r9s/

This is quite simple to implement, I'll mark it good for beginners. Make sure it is implemented in the `gauge` series types, that the solid gauge inherits from.

vitaliy-kretsul:
@karolkolodziej 
I am not sure what is the best way to calculate it, but here is what I am doing currently to get percentage: https://jsfiddle.net/Cortez/7vtx5kmn/
The 'percentage' property is present on 'point' object, but it is undefined. But for pie chart it is there, and it is very convenient to use it.

karolkolodziej:
Thank you @vitaliy-kretsul for additional feedback!

That is what I was hoping to see, now everything is clear!

hubertkozik:
Closing as completed in https://github.com/highcharts/highcharts/pull/18448.

PR #18448, opened 2023-02-06
feature/18433-gauge-percentage-property

Vazonik:
Added the `point.percentage` property for gauge charts, see #18433.

### DIFF
--- a/ts/Series/Gauge/GaugeSeries.ts
+++ b/ts/Series/Gauge/GaugeSeries.ts
@@ -46,7 +46,8 @@ const {
     extend,
     merge,
     pick,
-    pInt
+    pInt,
+    defined
 } = U;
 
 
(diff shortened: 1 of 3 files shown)


ts/Series/Gauge/GaugeSeries.ts
```diff
@@ -487,6 +487,11 @@ class GaugeSeries extends Series {
             // Positions for data label
             point.plotX = center[0];
             point.plotY = center[1];
+
+            if (point.y || point.y === 0) {
```
karolkolodziej, line 491: We have a utils method called `defined` which will fit perfectly here.

ts/Series/Gauge/GaugeSeries.ts
```diff
@@ -487,6 +487,11 @@ class GaugeSeries extends Series {
...
+                point.percentage =
+                    ((point.y - yAxis.min) / (yAxis.max - yAxis.min)) * 100;
```
karolkolodziej, line 493: You can also check if the range exists in order to avoid dividing by zero.

karolkolodziej (changes requested):
It looks good, but could you please add a test for that? Please reuse the already existing one for the gauge.

TorsteinHonsi:
In addition to Karol's comments, we also need a unit test. The cheapest approach is to find an existing gauge or solid gauge test and add an assertion for the percentage.

TorsteinHonsi (approved):
Thanks!

PR #18455, opened 2023-02-07
bugfix/10005-wrong-legend-value-series-mapping

### LINKED ISSUE #10005: Wrong legend value on series mapping
#### Expected behaviour
Would expect the legend to stay the same if the series mapping is changed

#### Actual behaviour
Legend value changes from "Column 1" to "Column 2"

http://jsfiddle.net/h6L7r2vw/
- At start legend is set to Column 1
- Remove label: 0 from series mapping
- Now legend is set to Column 2

#### Product version
Latest

hubertkozik:
Fixed #10005, wrong series name and legend item by using `data.seriesMapping` property.

### DIFF
--- a/samples/unit-tests/data/csv/demo.js
+++ b/samples/unit-tests/data/csv/demo.js
@@ -643,12 +643,10 @@ QUnit.test('startRow, endRow, startColumn, endColumn', function (assert) {
     });
 });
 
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks, looks good!

PR #18464, opened 2023-02-09
bugfix/18326-tooltip-blink

### LINKED ISSUE #18326: Tooltip sometimes blinks on iOS Safari on first touch
On iOS/Safari, on touching the chart, the tooltip sometimes flashes quickly before showing. Only on the first touch after page load. This happens only when the accessibility module is loaded. The intermediate `hide` call can be traced back to the keyboard navigation logic: 

https://github.com/highcharts/highcharts/blob/8b4774446dc53cd53bf05e1d0d617950e05ebf81/ts/Accessibility/Components/SeriesComponent/SeriesKeyboardNavigation.ts#L556-L558

https://jsfiddle.net/highcharts/0qmopktv/show
The issue can be reproduced in Chrome Dev tools with mobile preview. The flashing is so quick that it can't be seen, but in the jsFiddle we can verify that `Tooltip.hide` is called when it shouldn't. 

#### Product version
v10.3.2

#### Affected browser(s)
iOS/Safari

marvin19:
Fixed #18326, tooltip was blinking on iOS Safari on first touch.

### DIFF
--- a/ts/Accessibility/KeyboardNavigation.ts
+++ b/ts/Accessibility/KeyboardNavigation.ts
@@ -92,7 +92,6 @@ class KeyboardNavigation {
     public isClickingChart?: boolean;
     public keyboardReset?: boolean;
     public modules: Array<KeyboardNavigationHandler> = [];
-    public pointerIsOverChart?: boolean;
     public tabindexContainer: HTMLDOMElement = void 0 as any;
     public tabbingInBackwards?: boolean;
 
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (changes requested):
Thanks! The fix itself looks good to me, except maybe the `any` cast. But I don't have a good proposal for a fix.

For the description itself, see [Writing content for the changelog](https://github.com/highcharts/highcharts/tree/master/changelog#writing-content-for-the-changelog). Specifically:
* No. 7, describe the symptoms, not the cure, otherwise the readers of the changelog won't recognize it. The symptoms here is what the [issue](https://github.com/highcharts/highcharts/pull/18464) describes.
* No. 8, write either code formatting or plain English. But this probably doesn't apply anymore when you change the description according to no. 7.

TorsteinHonsi (approved):
Thanks!

PR #18468, opened 2023-02-09
enhancement/16658-mousewheel-zoom

DJTechnoo:
Added support for mouse wheel zooming through [chart.zooming.mouseWheel](https://api.highcharts.com/highcharts/chart.zooming.mouseWheel). This feature is embedded in the [Highcharts Stock bundle](https://code.highcharts.com/stock/highstock.js), but requires an [additional module file](https://code.highcharts.com/modules/mouse-wheel-zoom.js) for the [Highcharts bundle](https://code.highcharts.com/highcharts.js).


TorsteinHonsi (changes requested):
Thanks so far! In addition to the inline comments

* Please add at least one sample
* It should be built into Highcharts Stock. See how we did it with the DataGrouping module, which is also standalone for Highcharts but built into Stock. Should it be enabled by default? What do you think @pawelfus ?

More QA is needed. Try to break it with weird settings. For example
* In https://jsfiddle.net/highcharts/f1y35seb/, it fails to zoom above ca. 10 on the x axis
* With a reversed axis it gets all confused: https://jsfiddle.net/highcharts/f1y35seb/1/
* Inverted chart, the zooming does not keep the position under the mouse: https://jsfiddle.net/highcharts/f1y35seb/2/. (Pro tip: test with all combinations of inverted chart, and reversed x and y axes).
* Polar chart, it does react to mousewheel when we have first drag-zoomed in, then do the wheel. But it is not correct: https://jsfiddle.net/highcharts/f1y35seb/3/ . This one is not top priority though.

pawelfus:
> Should it be enabled by default? What do you think @pawelfus ?

No strong opinion here, but it would be nice to have it enabled by default. First I thought about it like `exporting`, `no-data-to-display`, or `series-label` modules, which are enabled by default. But at the same time, we have others that need to be configured first (eg `draggable-points`).

> Please add at least one sample

Can I ask for one too? I mean, one for Highcharts (with importing the module) and one for Stock (without loading the module).

TorsteinHonsi:
_Note to self_: Pending the requested changes.

DJTechnoo:
- [ ] In https://jsfiddle.net/9rvf0bp8/, it fails to zoom above ca. 10 on the x axis
- [ ] With a reversed axis it gets all confused: https://jsfiddle.net/cufr41gm/
- [ ] Inverted chart, the zooming does not keep the position under the mouse: https://jsfiddle.net/fpmjcbxn/. 
- [ ] (Bonus?)Polar chart, it does react to mousewheel ~~when we have first drag-zoomed in, then do the wheel. But it is not correct.~~ Now it zooms but looks broken after a deep zoom: https://jsfiddle.net/6urh3tas/ . This one is not top priority though. 
- [ ] Fix stock zoom out bug: https://jsfiddle.net/5s9p12oj/

DJTechnoo:
> Hi,
> 
> there seems to be a performance issue when using this type of zoom for grouped data: The performance gets worse, the more you zoom in. The first "zoom-steps" are fine / fast.
> 
> Live demo: https://jsfiddle.net/freseb/cu6mkf07/
> 
> Greetings

Hello,

Thank you for making us aware. Will discuss this internally and investigate.
Have a great weekend!

TorsteinHonsi (changes requested):
Thanks!

* The [API docs](https://github.com/highcharts/highcharts/blob/1b282147b2ec977478d71e3cac010de8b9b3e815/ts/Extensions/MouseWheelZoom/MouseWheelZoom.ts#L288-L289) should mention that this feature is included in Highcharts Stock, and that it is available a a module in Highcharts Core. Then add two `@sample` links to show it. The sensitivity demo is fine for Highcharts. But maybe we should have a more specialized demo for Stock?
* It doesn't work out of the box in Stock. It works if I explicitly set it to enabled in the chart config, but as per the docs and what we agreed on, that shouldn't be necessary.
* `zoomType: 'y'` apparently doesn't work. Tested on the sensitivity demo.

TorsteinHonsi (approved):
All good, thanks! 🎉 

I did a final change to the y axis post-zoom adjustment, to support animation of the axis ticks and grid lines.

Now only awaiting CI before merging.

PR #18472, opened 2023-02-10
db/editableOptions-serialization

karolkolodziej:
Removed store from `editableOptions` array.


sebastianbochan (approved):

goransle (approved):
Should be good for now, but this is something we should consider when implementing changing data sources via GUI

PR #18476, opened 2023-02-10
docs/18450-chart-animation-defaults

### LINKED ISSUE #18450: The chart update method does not have set animations to true by default.
#### Expected behaviour
As [API documentation](https://api.highcharts.com/class-reference/Highcharts.Chart#update) states, the animation should be enabled by default while performing the chart update.

#### Actual behaviour
In my case, I would like to disable all the animations and only perform them while updating the chart.
When performing chart updates with only specifying one argument, the chart does not animate.

https://jsfiddle.net/BlackLabel/h6dvsLf0/

#### Product version

#### Affected browser(s)
All

#### Suggested fix
```TS
    public update(
        options: Partial<Options>,
        redraw?: boolean,
        oneToOne?: boolean,
        animation?: (boolean|Partial<AnimationOptions>)
    ): void {
        animation = pick(animation, true);
```

Vazonik:
Fixed #18450, wrong chart animation defaults.

### DIFF
--- a/ts/Core/Chart/Chart.ts
+++ b/ts/Core/Chart/Chart.ts
@@ -815,7 +815,8 @@ class Chart {
      * @function Highcharts.Chart#redraw
      *
      * @param {boolean|Partial<Highcharts.AnimationOptionsObject>} [animation]
-     * If or how to apply animation to the redraw.
+     * If or how to apply animation to the redraw. When `undefined`, it applies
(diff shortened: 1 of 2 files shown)


karolkolodziej (approved):

TorsteinHonsi (approved):
Thanks!

PR #18481, opened 2023-02-13
bd/board-factory

karolkolodziej:
- [x] Rename `Dashboards.Dashboard` into `Dashboard.Board`
- [x] Implement `Dashboard.board` factory
- [x] Fix demos
- [x] Fix documentation
- [x] change `xxx.dashboard` to `xxx.board`


sebastianbochan (approved):

goransle (approved):
LGTM

ts/masters/dashboards.src.ts
```diff
@@ -53,9 +53,9 @@ declare global {
 const D = {
     ...Globals,
     ...Utilities,
-    _modules: (typeof _modules === 'undefined' ? {} : _modules),
+    Board: Dashboard,
+    board: Dashboard.board,
```
bre1470, line 57: The import should be renamed as well, so you can skip the colon thing.

samples/dashboards/demos/climate/demo.js
```diff
@@ -42,7 +42,7 @@ async function setupDashboard() {
     ).then(response => response.json());
     const mapPoints = await buildCitiesMap();
 
-    return new Promise(resolve => new Dashboards.Dashboard('container', {
+    return new Promise(resolve => new Dashboards.Board('container', {
```
bre1470, line 45: As a reference demo this should have the `Dashboards.board` call.

docs/dashboards/your-first-dashboard.md
```diff
@@ -29,9 +29,15 @@ Thus we have to import the main dashboard package, the Highcharts library and a
     * the div’s id where the dashboard should be placed (see point 1).
     * the object with dashboard options.
 
+    To do so, you can use factory function `Dashboards.board`:
 
     ```js
-        const dashboard = new Dashboards.Dashboard('container', {...})
+        Dashboards.board('container', {...})
+    ```
+    or the constructor to create a new instance of the dashboard:
```
bre1470, line 37: or the class `Dashboards.Board`…

bre1470 (changes requested):

pawellysy (approved):
LGTM! :D

bre1470 (changes requested):
I do not see the point done:
- Rename Dashboards.Dashboard into Dashboards.Board

karolkolodziej:
@bre1470 [isn't it that?](https://github.com/highcharts/highcharts/pull/18481/files#diff-af80584fc55104f51ff478328047f79b9630f972378d89ca846d93ecba4f1d8aR56)

bre1470:
Okay, I understand now. I am expecting the rename of the file as well. So, from `ts/Dashboards/Dashboard.ts` to `ts/Dashboards/Board.ts` plus the adjustments of imports.

bre1470:
@pawelfus
Should also `X.dashboard` properties (reference of `Board` class) be renamed?

karolkolodziej:
IMO it will be cleaner and shorter if we use `xxx.board` instead of `xxx.dashboard` when referring to the `Board` instance.

bre1470 (changes requested):
Cypress imports need adjustment.

pawelfus:
> @pawelfus Should also `X.dashboard` properties (reference of `Board` instance) be renamed?

Yes, please do 👍 Note: I wouldn't agree to change like `series.chart` => `series.c`, because that would break existing implementations, but in this case - just checked with Karol docs and there's no place we recommend this, so it's safe 👍

bre1470 (approved):
Thank you! Looks good. 👍

PR #18482, opened 2023-02-13
Removed pictorial doc from sidebar

goransle:
Assuming it should not be live yet as the samples are broken


TorsteinHonsi (approved):
Thanks!

TorsteinHonsi:
@goransle What about treegraph? It is also merged into the master but not released.

PR #18483, opened 2023-02-13
db/remove-dashboard

karolkolodziej:
The `destroy` dashboard did not work correctly.

Todo:
- [x] Fix errors in the console
- [x] Remove the container
- [x] Test

### DIFF
--- a/ts/Dashboards/Layout/Layout.ts
+++ b/ts/Dashboards/Layout/Layout.ts
@@ -277,9 +277,6 @@ class Layout extends GUIElement {
         layout.rows.forEach(function (row): void {
             row.destroy();
         });
-        // for (let i = 0, iEnd = layout.rows.length; i < iEnd; ++i) {
-        //     layout.rows[i].destroy();
-        // }
 
         if (layout.parentCell) {
             layout.parentCell.destroy();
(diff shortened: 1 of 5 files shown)


ts/Dashboards/Layout/Layout.ts
```diff
@@ -274,9 +274,9 @@ class Layout extends GUIElement {
         const layout = this;
 
         // Destroy rows.
-        for (let i = 0, iEnd = layout.rows.length; i < iEnd; ++i) {
-            layout.rows[i].destroy();
-        }
+        layout.rows.forEach(function (row): void {
+            row.destroy();
+        });
```
bre1470, line 279: It is slower. Please restore the regular loop.

bre1470 (approved):
Looks good with exception of the loop code.

ts/Dashboards/Layout/Layout.ts
sebastianbochan: I agree that the regular for (in the appropriate declaration - see in other components) is the fastest.

ts/Dashboards/Layout/Row.ts
```diff
@@ -291,13 +291,15 @@ class Row extends GUIElement {
             }
         }
 
-        row.layout.unmountRow(row);
-        const destroyLayout = layout.rows.length === 0;
+        if (row.layout) {
+            row.layout.unmountRow(row);
+            const destroyLayout = layout.rows.length === 0;
```
sebastianbochan, line 296: Do we need to declare this for one-time use only?

sebastianbochan (changes requested):

sebastianbochan (approved):

PR #18489, opened 2023-02-13
bugfix/18245-map-zoom-buttons-style

### LINKED ISSUE #18245: MapNavigation buttons' text should be cenetered in styled mode
#### Expected behaviour
When `styledMode: enabled`, text in the mapNaviation buttons should be centered vertically and horizontally.

#### Actual behaviour
Centered only vertically.

https://jsfiddle.net/BlackLabel/tw6f5v4r/

#### Workaround
https://jsfiddle.net/BlackLabel/vwhyts4u/
```
.highcharts-button.highcharts-zoom-in text {
	 transform: translateX(2px);
}
.highcharts-button.highcharts-zoom-out text {
	 transform: translateX(3.5px);
}
```

Vazonik:
Fixed #18245, centered map navigation buttons' text.

### DIFF
--- a/css/highcharts.scss
+++ b/css/highcharts.scss
@@ -740,6 +740,12 @@ input.highcharts-range-selector {
     font-weight: bold;
     text-align: center;
 }
+.highcharts-button.highcharts-zoom-in text {
+    transform: translateX(2.25px);
+}
+.highcharts-button.highcharts-zoom-out text {
+    transform: translateX(3.75px);
+}
 .highcharts-mapview-inset-border {
     stroke: $neutral-color-20;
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (changes requested):
Thanks! The change is good, but this file is going to disappear in v11, which is soon to be released. So merging it now will only cause a merge conflict and risk that the change gets lost. Can you create a PR based on the `v11` branch instead? Then you will have to apply the same changes to the `css/highcharts.css` file.

Vazonik:
PR based on the `v11` branch: #18495

PR #18493, opened 2023-02-14
Imported data layer updates

bre1470:


ts/Core/Utilities.ts
```diff
@@ -207,34 +207,45 @@ function merge<T>(): T {
...
         ret = {} as T;
-    const doCopy = function (copy: any, original: any): any {
-        // An object is replacing a primitive
-        if (typeof copy !== 'object') {
-            copy = {};
-        }
 
-        objectEach(original, function (value, key): void {
+    // describtive error stack:
+    const copyDepthError = new Error('Recursive copy depth > 100'),
```
TorsteinHonsi, line 214: Shouldn't we stick to our own [error handling practice](https://github.com/highcharts/highcharts/blob/c09c2bb8ddcd8bece69a074bc8d3e3fef7e7eb61/ts/Core/Utilities.ts#L83)?

TorsteinHonsi (changes requested):
Unless there's a specific reason, I think we should use consistent error handling - see inline comment.

ts/Core/Utilities.ts
bre1470: Good point. This is from Dashboards 2020 chart component and can be removed.

TorsteinHonsi (approved):
Looks good, thanks!

PR #18496, opened 2023-02-14
ts/migration-tooltip

bre1470:
- ~Moved tooltip to extensions.~
- Added tooltip composition.
- Added Pointer:afterInit event.
- Removed tooltip from a11y, annotations, pointer, series.


TorsteinHonsi:
Thanks so far!

The tooltip is now missing from `highcharts.js`. 

> Moved tooltip to extensions

I am wondering if there's a misunderstanding here - Tooltip should still be part of the core. What we want to do is to make it possible to create _custom_ builds where the tooltip is not included. As far as I understand, it is currently not possible to remove (and tree-shake off) the Tooltip because it is used in Chart.ts?

bre1470:
> I am wondering if there's a misunderstanding here - Tooltip should still be part of the core. What we want to do is to make it possible to create _custom_ builds where the tooltip is not included. As far as I understand, it is currently not possible to remove (and tree-shake off) the Tooltip because it is used in Chart.ts?

Yes, sorry. I will reintegrate it. `Tooltip` is/was a hard dependency (not type) in `Core/Pointer`, where the related chart received a new tooltip instance.

bre1470:
> ... and do we need the added code of making Tooltip a composition? As far as I can see, only Chart refers to `tooltip` directly. In Pointer, the `tooltip` is always read from `chart.tooltip` and checked for falsy before accessing it.

Pointer is creating the Tooltip instance on chart. The rest only deals with tooltip options and/or `chart.tooltip` property.

bre1470:
This will still need a composition to support async import with modifications of Pointer or Chart, but compose is not much overhead. The overhead comes mainly from assemblers imports with lengthy paths and internal `_modules`, because this can not be compressed. I can move the event handle into the composition, so we save some bytes.

Should I move the default tooltip options back into `Core/Defaults` or keep them in the new `TooltipDefaults`?

TorsteinHonsi:
All right 👍 

> Should I move the default tooltip options back into Core/Defaults or keep them in the new TooltipDefaults?

I think it should be in the core defaults, at least for now. So the point is we want to make it _possible_ for implementers to make a custom build without the Tooltip, but it is not the norm. Next we can explore the same with Pointer, Legend etc.

tools/gulptasks/jsdoc-classes.js
```diff
@@ -61,6 +60,7 @@ const SOURCE_GLOBS = [
     'js/Extensions/OfflineExporting/OfflineExporting.js',
     'js/Extensions/PatternFill.js',
     'js/Extensions/SeriesLabel/SeriesLabel.js',
+    'js/Extensions/Tooltip/Tooltip.js',
```
TorsteinHonsi, line 63: But it's now in the Core again, right?

ts/Core/Axis/Axis.ts
```diff
@@ -514,7 +514,10 @@ class Axis {
          */
         const crosshair = pick(
             options.crosshair,
-            splat((chart.options.tooltip as any).crosshairs)[isXAxis ? 0 : 1]
+            splat(
+                chart.options.tooltip &&
+                chart.options.tooltip.crosshairs
```
TorsteinHonsi, line 519: Can this change be reverted? Now that `options.tooltip` should always exist.

TorsteinHonsi (commented):
I added a couple of comments - there are several places now that test for the pattern `chartOptions.tooltip && chartOptions.tooltip.someProperty`, or `if (chart.options.tooltip)`. But `chart.options.tooltip` is now defined in `Chart/Defaults.ts`, so we know it exists even though TS doesn't. So I suggest we revert these changes and live with the `any`s until we get the Partial options handling sorted out?

tools/gulptasks/jsdoc-classes.js
bre1470: Thanks, good catch.

TorsteinHonsi:
As part of making the core stuff modular - do we have any tests where we can see how the file size of a minimal build is affected? I mean it is a goal in this work that it should be able to create a small, stripped build. So in order to pursue that goal we need to know a) that stuff still works for a minimal setup, and b) how it affects the build size (of `custom-highcharts.js`).

bre1470:
We could create a master bundle that only imports line series and maybe other files for compose and then use the `test/typescript-karma` to test it. Our regular karma always loads all bundles so that you never know, if an alternative is working unless you load it in a separate session from `code/` or `code/es-modules/masters/`.

TorsteinHonsi (approved):
Thanks! We can deal with the tests in another PR.

PR #18500, opened 2023-02-15
bugfix/18444-polar-bar-datalabels-error

### LINKED ISSUE #18444: Polar bar fails with data labels
#### Actual behaviour
A polar chart with type "bar" and data labels causes JavaScript error.

https://jsfiddle.net/highcharts/30fg6y8d/

#### Product version
v10.3.3

#### Affected browser(s)
All

#### Workaround
Set `type: 'column'` and `inverted: true`.

#### Suggested fix
The following statement goes wrong because it assumes that `Series` is two steps up in the inheritance chain:

https://github.com/highcharts/highcharts/blob/d8c12a4b062ccf79dde8ae8c37ed8c2ab7baa73b/ts/Series/PolarComposition.ts#L896-L905

However, with `bar`, and theoretically other custom series types, this assumption is not correct, and we get into an eternal loop. The fix is to apply it on the `Series` prototype directly: 

```ts
        Series.prototype.alignDataLabel.call(
            this, point, dataLabel, options, alignTo, isNew
        );
```

In addition to that statement, Series must be imported and the typing of `alignTo` must be resolved.

hubertkozik:
Fixed #18444, a polar chart with type "bar" and data labels caused JavaScript error.

### DIFF
--- a/samples/unit-tests/chart/polar/demo.js
+++ b/samples/unit-tests/chart/polar/demo.js
@@ -57,6 +57,21 @@ QUnit.test('Polar chart data', function (assert) {
         ),
         '#18147: All data labels should be visible'
     );
+
+    chart.update({
+        chart: {
+            inverted: true
+        }
+    }, false);
+    chart.series[0].update({
+        type: 'bar'
+    });
+
+    assert.ok(
+        true,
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Perfect, thanks!

PR #18507, opened 2023-02-15
Db/chartOptions-update

sebastianbochan:


bre1470 (approved):
See comment, looks good otherwise.

ts/Extensions/DashboardPlugins/HighchartsComponent.ts
```diff
@@ -233,14 +232,23 @@ class HighchartsComponent extends Component<HighchartsComponent.ChartComponentEv
...
 
+        hcComponent.emit({ type: 'beforeRender' });
+        super.render();
+        hcComponent.chart = hcComponent.initChart();
+        hcComponent.updateSeries();
+        hcComponent.sync.start();
+        hcComponent.emit({ type: 'afterRender' });
+        hcComponent.setupStoreUpdate();
+
+        addEvent(hcComponent.chart, 'update', function (e):void {
```
bre1470, line 245: It might be better to listen to `afterUpdate` instead, then special case are catched and you can access `this.options`.

ts/Extensions/DashboardPlugins/HighchartsComponent.ts
```diff
@@ -301,9 +309,21 @@ class HighchartsComponent extends Component<HighchartsComponent.ChartComponentEv
...
     }
+    /**
+     * Handles updating via options
+     * @param {Partial<Component.ComponentOptions>} options
+     * The options to apply
+     *
+     * @param {boolean} redraw
+     * The flag triggers the main redraw operation
+     */
+    private updateComponentOptions(options: Partial<HighchartsComponent.ComponentOptions>, redraw = true): void {
```
karolkolodziej, line 320: Shouldn't eslint complain here about the line length?

karolkolodziej (approved):

karolkolodziej (changes requested):
For this particular issue, we should introduce a test.

ts/Extensions/DashboardPlugins/HighchartsComponent.ts
bre1470: Because of generic typing it will not complaint.

sebastianbochan: No errors here ;)

bre1470: Can you please fix it and split over multiple lines. The eslint ignore is just for overlengthy generic types, which is not the case here.

test/typescript-karma/Dashboards/Component/tests.js
```diff
@@ -478,7 +480,34 @@ test('HighchartsComponent resizing', function(assert) {
...
+        }
+    }).render();
+
+    component.chart.update({
+        title: {
+            text: 'updated'
+        }
+    });
+
+    assert.deepEqual(component.options.chartOptions.title.text, 'updated');
```
karolkolodziej, line 509: Do we need that `deepEqual` assertion? Isn't `srictEqual` enough?

karolkolodziej (approved):

test/typescript-karma/Dashboards/Component/tests.js
sebastianbochan: Good point!

Long time ago, I created tests like this one :D

bre1470: It depends. `deepEqual` is not strict. `strictEqual` is not deep.

pawellysy (approved):
LGTM

PR #18510, opened 2023-02-16
bugfix/18502-heatmap-pointpadding-yaxis-reversed

### LINKED ISSUE #18502: Heatmap: pointPadding with yAxis.reversed is not working correctly
#### Expected behaviour
Points should have a correct 10px padding in all directions like in a normal chart: https://jsfiddle.net/BlackLabel/32h6mpor/
image

#### Actual behaviour
Points have wrong pointPadding when `yAxis.reversed: true`: https://jsfiddle.net/BlackLabel/fgyzkce8/
image

#### Product version
Highcharts 10.3.3

Worked fine in 8.0.4, broken since 8.1.0. Unable to bisect.

hubertkozik:
Fixed #18502, `pointPadding` for heatmap series wasn't working with reversed axes.

### DIFF
--- a/samples/unit-tests/series-heatmap/members/demo.js
+++ b/samples/unit-tests/series-heatmap/members/demo.js
@@ -1,7 +1,7 @@
 QUnit.test(
     'Heatmap point size shouldn\'t overflow plot area(#4530)',
     function (assert) {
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks, looking good!

PR #18511, opened 2023-02-16
KPI Serialization.

pawellysy:
Implemented serialization methods to KPIComponent.

### DIFF
--- a/ts/Dashboards/Actions/Bindings.ts
+++ b/ts/Dashboards/Actions/Bindings.ts
@@ -18,15 +18,17 @@ import type ComponentTypes from '../Component/ComponentType';
 import type GUIElement from '../Layout/GUIElement';
 import type HighchartsComponent from '../../Extensions/DashboardPlugins/HighchartsComponent';
 import type Serializable from '../Serializable';
-import type Cell from '../Layout/Cell.js';
-import type Layout from '../Layout/Layout.js';
-import type Row from '../Layout/Row.js';
+import type KPIComponent from '../Component/KPIComponent';
+import type DataStore from '../../Data/Stores/DataStore';
+import type Cell from '../Layout/Cell';
+import type Layout from '../Layout/Layout';
+import type Row from '../Layout/Row';
 
 import Component from '../Component/Component.js';
 import HTMLComponent from '../Component/HTMLComponent.js';
 import DataGridComponent from '../../Extensions/DashboardPlugins/DataGridComponent.js';
 import Globals from '../Globals.js';
-import KPIComponent from '../Component/KPIComponent.js';
+import DataTable from '../../Data/DataTable';
 import U from '../../Core/Utilities.js';
 const {
     fireEvent,
@@ -40,23 +42,18 @@ class Bindings {
      *  Functions
      *
      * */
-    private static getGUIElement(idOrElement: string): GUIElement | undefined {
-        const container =
-            typeof idOrElement === 'string' ?
-                document.getElementById(idOrElement) :
-                idOrElement;
+    private static getGUIElement(idOrElement: string): GUIElement|undefined {
+        const container = typeof idOrElement === 'string' ?
+            document.getElementById(idOrElement) : idOrElement;
 
         let guiElement;
 
         if (container instanceof HTMLElement) {
-            fireEvent(
-                container,
-                'bindedGUIElement',
-                {},
-                function (e: GUIElement.BindedGUIElementEvent): void {
-                    guiElement = e.guiElement;
-                }
-            );
+            fireEvent(container, 'bindedGUIElement', {}, function (
+                e: GUIElement.BindedGUIElementEvent
+            ): void {
+                guiElement = e.guiElement;
+            });
         }
 
         return guiElement;
@@ -71,7 +68,7 @@ class Bindings {
         const optionsEvents = options.events;
 
         cell = cell || Bindings.getCell(options.cell);
-        let component: ComponentTypes | undefined;
+        let component: ComponentTypes|undefined;
 
         // add elements to containers
         if (componentContainer) {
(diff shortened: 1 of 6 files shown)


ts/Dashboards/Actions/Bindings.ts
```diff
@@ -18,17 +18,15 @@ import type ComponentTypes from '../Component/ComponentType';
 import type GUIElement from '../Layout/GUIElement';
 import type HighchartsComponent from '../../Extensions/DashboardPlugins/HighchartsComponent';
 import type Serializable from '../Serializable';
-import type KPIComponent from '../Component/KPIComponent';
-import type DataStore from '../../Data/Stores/DataStore';
+import type Cell from '../Layout/Cell.js';
```
karolkolodziej, line 21: No need to add `.js` when importing type

ts/Dashboards/Layout/Cell.ts
```diff
@@ -26,11 +26,15 @@ import type Serializable from '../Serializable';
 
 import Globals from '../Globals.js';
 import GUIElement from './GUIElement.js';
-import Bindings from '../Actions/Bindings.js';
+import B from '../Actions/Bindings.js';
```
karolkolodziej, line 29: We usually call it `Bindings`

ts/Dashboards/Layout/Cell.ts
```diff
@@ -26,11 +26,15 @@ import type Serializable from '../Serializable';
...
 import U from '../../Core/Utilities.js';
 import { HTMLDOMElement } from '../../Core/Renderer/DOMElementType';
 import EditGlobals from '../EditMode/EditGlobals.js';
 
+const {
+    componentFromJSON
+} = B;
+const B2 = B;
```
karolkolodziej, line 37: What is that `B2` for?

karolkolodziej (changes requested):

ts/Dashboards/Actions/Bindings.ts
```diff
@@ -226,17 +225,35 @@ class Bindings {
 
     public static getCell(idOrElement: string): Cell|undefined {
         const cell = Bindings.getGUIElement(idOrElement);
-        return cell instanceof Cell ? cell : void 0;
+
+        if (!cell) {
+            return;
+        }
+        let isCellType = cell?.getType() === 'cell';
```
sebastianbochan, line 232: Please remove optional chaining everywhere.

ts/Dashboards/Actions/Bindings.ts
```diff
@@ -226,17 +225,35 @@ class Bindings {
...
+
+        return isCellType ? (cell as Cell) : void 0;
     }
 
     public static getRow(idOrElement: string): Row|undefined {
         const row = Bindings.getGUIElement(idOrElement);
-        return row instanceof Row ? row : void 0;
+
+        if (!row) {
```
sebastianbochan, line 240: The `return;` returns `undefined` and `return isLayoutType ? (layout as Layout) : void 0;` also when the type does not exist. As a result, you can simplify it. You could check the type in the first condition and avoid the inline condition in the last return.

ts/Dashboards/Component/KPIComponent.ts
```diff
@@ -349,11 +349,73 @@ class KPIComponent extends Component {
...
+            options: {
+                ...base.options,
+                value: this.options.value,
+                subtitle: JSON.stringify(this.options.subtitle),
+                title: JSON.stringify(this.options.title),
+                threshold: this.options.threshold,
+                style: JSON.stringify(this.options.style),
+                thresholdColors: this.options.thresholdColors,
+                chartOptions: JSON.stringify(this.options.chartOptions),
+                valueFormat: this.options.valueFormat
```
sebastianbochan, line 365: Why we cannot `stringify` the whole `this.options` object? The value can be string too.

sebastianbochan (changes requested):

ts/Dashboards/Component/KPIComponent.ts
bre1470: valueFormatter and callback functions in general should not be stringified for security reasons as it opens the door to inject code via JSON.

pawellysy: I can't stringify `valueFormatter` because it is a callback function.  `valueFormat` on the other hand is a string, so it can work, as you said :D

pawellysy: Additionally, I needed to merge the options with the base json.

ts/Dashboards/Component/KPIComponent.ts
```diff
@@ -349,11 +349,73 @@ class KPIComponent extends Component {
...
     export type ComponentType = KPIComponent;
 
+    export interface ClassJSON extends Component.JSON {
+        options: ComponentJSONOptions;
+
+    }
+
+    export interface ComponentJSONOptions extends Component.ComponentOptionsJSON {
+        title?: string;
+        chartOptions?: string;
```
bre1470, line 411: Why does chartOptions get stringified and not assigned directly with `Core/Options` type? `import type` is safe as it does not add to the Dashboard bundle.

pawellysy: I looked at how it is done in the `HighchartsComponent` and wanted to be consistent. I can change it though, if you like

bre1470: I see. This is probably related to the callback options to Highcharts. Let's keep it for now.

bre1470 (changes requested):
Please revert the type formatting everywhere.

ts/Dashboards/Actions/Bindings.ts
karolkolodziej, line 21: It is not resolved.

karolkolodziej (approved):

sebastianbochan (approved):

PR #18515, opened 2023-02-17
bugfix/18276-touchscreen-annotation-select

### LINKED ISSUE #18276: Touchscreen devices could not end drawing, select nor edit the drawings (eg. Rectangle, Fibonacci etc.)
#### Expected behaviour
The drawing's _(eg. Rectangle)_ second step/point could not be plotted and therefore could not end while using the Safari browser.

Using the Google Chrome browser can complete the drawing, but the drawing could not be selected to edit the points nor the options to edit/delete.

#### Actual behaviour
The drawings should behave the same as when using a mouse and keyboard.
Each step/point to end on each tap should behave similarly to a mouse click.

Draw a "Rectangle", then single tap on the drawn rectangle to attempt to select.
Highcharts was not able to pick up the single taps to select the drawings.

#### Product version

Highcharts Stock v10.3.2

#### Affected browser(s)
Safari and Chrome on touchscreen devices (Tested on iPadOS 15.7, iPad Air 4th Gen and also on Android)

kamil-musialowski:
Fixed #18276, annotations touch was not captured on mobile devices.

### DIFF
--- a/ts/Core/Renderer/SVG/SVGElement.ts
+++ b/ts/Core/Renderer/SVG/SVGElement.ts
@@ -29,6 +29,7 @@ import type {
 } from '../DOMElementType';
 import type FontMetricsObject from '../FontMetricsObject';
 import type GradientColor from '../../Color/GradientColor';
+import type PositionObject from '../PositionObject';
 import type RectangleObject from '../RectangleObject';
 import type ShadowOptionsObject from '../ShadowOptionsObject';
 import type SVGAttributes from './SVGAttributes';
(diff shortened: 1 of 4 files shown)


ts/Core/Renderer/SVG/SVGElement.ts
```diff
@@ -1741,14 +1742,47 @@ class SVGElement implements SVGElementLike {
...
+                    e.preventDefault();
+                }
+            };
+            element.onclick = function (e: MouseEvent): void {
+                if (!touchEventFired) {
+                    handler.call(handler, e);
+                }
+            };
+        } else {
+            (element as any)['on' + eventType] = handler;
```
kamil-musialowski, line 1784: I'm not happy about this `any` casting. 
@bre1470 Could you please have a look at this one?

pawelfus (changes requested):
I don't think we want to go back to a state where `on()` used to implement `click = touchStart + touchMove + touchEnd`. Could you confirm @TorsteinHonsi ?

If that's the case, then I see two alternative solutions:
1. In `NavigationBindings.ts` we have `selectAndShowPopup` assigned to `click` -> we can assign it also to `touchDown`. Sounds like the easiest solution - could you try it?
2. In `EventEmitter.ts` we have `addEvents()` method responsible for adding `click` callback. Perhaps we could move the logic there. But we go back to the question if we should implement `Annotation.options.event.click` as `touchStart + touchMove + touchEnd` too. If there was a reason to resign from `SVGElement.on('click')` = `touchClick`, perhaps we shouldn't repeat this in Annotations

TorsteinHonsi:
Confirmed. We've been there before. Better to go with one of Pawel's proposed solutions and keep this out of the core. Also see how we're handling touch events in other parts of the code base.

kamil-musialowski:
@pawelfus With the current commit, there is no `if` check, that would prevent calling the `touchend` event in case of a finger drag event. Should it be implemented here?

pawelfus (approved):
Looks good, thanks!

TorsteinHonsi (approved):

PR #18529, opened 2023-02-20
ts/migration-psarindicator

marvin19:
Improved PSARIndicator.


bre1470 (changes requested):
One thing can be done. Remove the lines here https://github.com/highcharts/highcharts/blob/08fd102498f2e94d460fe3cbad47bb9d6aae2fcd/ts/Stock/Indicators/PSAR/PSARIndicator.ts#L408-L410. This is only needed in the class structure. In class prototype (interface) the type definition is enough.

bre1470 (changes requested):
Sorry for the mess. I did not expect visual tests to fail. Please move the `nameComponents` property from interface to class.

bre1470 (approved):
Thank you! Looks good. 👍

TorsteinHonsi (approved):

PR #18530, opened 2023-02-20
bugfix/18429-tick-label-animation

### LINKED ISSUE #18429: Allow axis label animation to be disabled separately
# Background & reasoning
Currently the animations of a chart can be set via the `chart.animation` property. This setting enables or disables multiple independent animations at once. For example, whenever you toggle a series in the legend the following animations are triggered:
- The (de-)selected series is smoothly shown / removed from the chart area
- The old axis labels fly out and the new labels fly into the chart area

In my case the latter animation is triggered because the label formatter only incorporates data from the visible series and thus the texts change upon toggling a series. Take a look at this example:
https://jsfiddle.net/FeBe95/c1z7w6fj/

The percentages do change but it doesn't make sense to have the labels fly in and out. Currently there is no way to disable this animation separately and keep the animation of toggling a series. Setting `chart.animation` to `false` disables both animations. 

# Proposed solution
In the source code there already is a seperate check whether to perform the axis animation. Though it currently only depends on the `axis.tickInterval` and `axis.labels.step` properties:
https://github.com/highcharts/highcharts/blob/ac0b014a021c8ec7017bef2b19caf119418c32a3/ts/Core/Axis/Tick.ts#L216-L217

The new feature could be implement as follows:
- Introduce a new boolean `axis.labels.animation` property
- Take this property into account for the check, in addition to `axis.tickInterval` and `axis.labels.step`

# Further thoughts
I thought about some workarounds but they seem unnecessarily complicated and might not even work:
- Use HTML labels, a data-percentage attribute on the span and a CSS ::after pseudo selector to display the percentage. The text content of the label should stay the same and thus the animation should not be triggered (?).
- Use the annotation module and place the calculated percentages manually next to the labels. Update the annotations when any series is toggled.

Highcharts

----

Vazonik:
Fixed #18429, deleted tick label sort animation.


TorsteinHonsi:
Thanks! As far as I can see this leads to a change in how new labels are introduced in http://localhost:3030/samples/#view/highcharts/demo/column-comparison, but I don't see that as a problem.

@ppotaczek The label animation feature was introduced with data sorting (https://github.com/highcharts/highcharts/commit/b05785120e5e53572d06eb35372005ec2a0b599b). Do you see any obvious problems with the changes in this PR?

ppotaczek:
Hi guys, The idea of the animation was intended only for sliding out/in x-axis labels with data sorting feature for new and old points. We can see a really similar situation here: https://app.flourish.studio/@flourish/bar-chart-race. The code is totally fine if the new way of animation looks good for you.

Simplified difference:
- Branch **master**: http://jsfiddle.net/BlackLabel/j6nxagm9/
- Branch **bugfix/18429-tick-label-animation**: http://jsfiddle.net/BlackLabel/9xL120pw/

As an alternative we can use the animation in a more specific case: https://github.com/highcharts/highcharts/blob/97a76c3e8d904062b0e7f8599089be0d7b318963/ts/Core/Axis/Tick.ts#L216-L217

TorsteinHonsi (approved):
Thanks Pawel! I'm willing to live with the _slightly_ degraded animation for sorting and dynamic data. Especially since it had side effects, like we don't know for sure where to animate to and from. A simple fade in/out is sufficient I think.

PR #18533, opened 2023-02-20
bugfix/broken-links-changelog

LindaSve93:
Fixed some broken links in the changelog, due to bad SEO. I am not sure how to build new changelog etc, but I assume someone else can fix that part, maybe @TorsteinHonsi ?


oysteinmoseng (approved):
All good from me!

oysteinmoseng:
@TorsteinHonsi would be the one to say if something needs to change elsewhere due to the changelog generator

TorsteinHonsi (approved):
Thanks! I'll build and update the changelog. For the records, it's [number 5 here](https://github.com/highcharts/highcharts/tree/master/changelog#workflow-of-generating-the-changelog).

PR #18534, opened 2023-02-20
bugfix/18501-stack-labels-vertical-align-logarithmic

### LINKED ISSUE #18501: Stack labels' `verticalAlign` doesn't work for logarithmic yAxis.
#### Expected behaviour
Setting `stackLabels.verticalAlign` in the logarithmic axis to any value works as in the regular axis.

#### Actual behaviour
Setting `stackLabels.verticalAlign` in the logarithmic axis to `bottom` or `middle` makes the labels invisible.

https://jsfiddle.net/BlackLabel/fntvzh31/

#### Product version
Highcharts v10.3.3

hubertkozik:
Fixed #18501, vertical alignment of stack labels didn't work with logarithmic axes.

### DIFF
--- a/ts/Core/Axis/Stacking/StackItem.ts
+++ b/ts/Core/Axis/Stacking/StackItem.ts
@@ -355,9 +355,15 @@ class StackItem {
             y = axis.toPixels(totalStackValue),
             xAxis = stackBoxProps.xAxis || chart.xAxis[0],
             x = pick(defaultX, xAxis.toPixels(this.x)) + xOffset,
-            yZero = axis.toPixels(boxBottom ? boxBottom :
-                (axis.logarithmic && isNumber(axis.min) ?
-                    Math.pow(10, axis.min) : 0)),
+            yZero = axis.toPixels(
+                boxBottom ||
+                (
+                    isNumber(axis.min) &&
+                    axis.logarithmic &&
+                    axis.logarithmic.lin2log(axis.min)
+                ) ||
+                0
+            ),
             height = Math.abs(y - yZero),
             inverted = chart.inverted,
             neg = stackItem.isNegative;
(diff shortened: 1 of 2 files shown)


ts/Core/Axis/Stacking/StackItem.ts
```diff
@@ -355,7 +355,9 @@ class StackItem {
             y = axis.toPixels(totalStackValue),
             xAxis = stackBoxProps.xAxis || chart.xAxis[0],
             x = pick(defaultX, xAxis.toPixels(this.x)) + xOffset,
-            yZero = axis.toPixels(boxBottom ? boxBottom : 0),
+            yZero = axis.toPixels(boxBottom ? boxBottom :
+                (axis.logarithmic && isNumber(axis.min) ?
+                    Math.pow(10, axis.min) : 0)),
```
TorsteinHonsi, line 360: Use the `lin2log` function as usual. We can also do an _or_ statement. So something like

```ts
            yZero = axis.toPixels(
                boxBottom ||
                (isNumber(axis.min) && axis.logarithmic?.log2lin(axis.min)) ||
                0
            ),
            
```.

TorsteinHonsi (changes requested):
Thanks! In addition to the inline comment, I'm curious why `axis.min` must be run through `log2lin`. We are already doing that [in the setTickInterval function](https://github.com/highcharts/highcharts/blob/0722252494fed0d83e4a52c2ed2711c53707c044/ts/Core/Axis/Axis.ts#L1703), so it should be ready to use. Or is it too late?

hubertkozik:
Yes, `correctFloat` is working correctly, but it is rounding a number, which is a bad value for `yZero` variable. We need to have there a bottom value of the yAxis - in this demo: https://jsfiddle.net/BlackLabel/sreodwg6/ it is a 125, which is saved to `chart.yAxis[0].min` as 2.0969100130080562 and we need to recalculate it.

My mistake, I should use `lin2log`at first, sorry for that!

TorsteinHonsi (approved):

Issue #13338, opened 2020-04-17
x number of minor ticks per major tick


awall:
# Description of the feature
When minorTicks is set to "auto", we get 5 minor ticks per major tick. Ideally, I would like to set the major ticks using other properties, like tickPixelInterval or tickAmount, and set the minor ticks as "2 minor ticks per major tick", or any other arbitrary multiple. Why restrict it to exactly 5?

Add some property to set this value from 5 to any arbitrary integer.
Please make sure it works in logarithmic and date axes, too!

# Library related to the feature
Highcharts

You can vote for the feature by adding thumbs-up reaction for this post.

sebastianbochan:
Hi @awall,
Thank you for your suggestion. I think that it could be something like tickAmount, but minorTickAmount.

We could parametrize this line:
https://github.com/highcharts/highcharts/blob/f9dc90de6c19a4410bba40c56846cec283155e37/js/parts/Axis.js#L1484-L1487

Have you tried to use [minorTickInterval](https://api.highcharts.com/highcharts/xAxis.minorTickInterval) ?

awall:
Thanks @sebastianbochan. In my case, the minorTickInterval does not work, it does work correctly for logarithmic or date axes (I've logged separate bugs for these issues already).

KacperMadej:
It works for logarithmic (https://jsfiddle.net/BlackLabel/vem4q6yt/) but doesn't look good since the offset is critical. It works for datetime (https://jsfiddle.net/BlackLabel/vem4q6yt/1/) - the units are milliseconds.

awall:
@KacperMadej Logarithmic only works for very specific numbers, try changing that "0.1" to "0.5" and see what happens ;). I've logged this bug already as https://github.com/highcharts/highcharts/issues/13337

As for datetime, I'd really like to use the default behaviour for the major ticks, and then just "2 minor ticks per major tick". If I use tickInterval and minorTickInterval, I need to write an enormous amount of tricky code involving dates and the chart width... this code is already built-in to how highcharts determines the major ticks on a date axis, and it is very painful to have to reverse engineer and duplicate that code.

pawelfus:
See plugin posted in #13336.  Demo: https://jsfiddle.net/BlackLabel/oc70hqxm/

pawellysy:
unstale

ReillyBova:
[...]

PR #18540, opened 2023-02-20
feature/13338-minor-ticks-per-major

Vazonik:
Added `minorTicksPerMajor` option for axes, see #13338.


ts/Core/Axis/AxisOptions.d.ts
```diff
@@ -151,6 +151,7 @@ export interface AxisOptions {
     minorTickLength: number;
     minorTickPosition: AxisTickPositionValue;
     minorTicks?: boolean;
+    minorTicksPerMajor?: number;
```
TorsteinHonsi, line 154: Since we're giving this number in the defaults, it doesn't have to be optional. We can remove the question mark.

ts/Core/Axis/Axis.ts
```diff
@@ -1930,6 +1930,7 @@ class Axis {
             tickPositionsOption = options.tickPositions,
             tickPositioner = options.tickPositioner,
             minorTickIntervalOption = this.getMinorTickInterval(),
+            minorTicksPerMajor = this.options.minorTicksPerMajor || 5,
```
TorsteinHonsi, line 1933: No need for a constant here, we're only using it once, so we can use it directly on line 1958. Also, no need for the `|| 5` part, if we make the option non-optional. Plus, no need for `this.options` because we already have the local `options` constant. So basically, remove line 1933 and change line 1958 to `this.tickInterval / options.minorTicksPerMajor`.

samples/unit-tests/axis/minortickspermajor/demo.js
```diff
@@ -0,0 +1,67 @@
+QUnit.test('Minor ticks per major - linear axis', function (assert) {
```
TorsteinHonsi, line 1: In order to keep processing time down for our test suite, try thinking of how to reduce the number of new charts generated. 

In this case, we can reduce it to one `QUnit.test` and re-use the same chart, just run `chart.yAxis[0].update` in between them to check linear, log and datetime respectively. 

Also consider if we can go even further and piggy back one of the existing tests related to minor ticks.

TorsteinHonsi (changes requested):
Thanks! See the inline comments, all related to boiling the code down to its essence.

ts/Core/Axis/AxisDefaults.ts
```diff
@@ -1278,6 +1278,17 @@ namespace AxisDefaults {
          * @apioption xAxis.minorTicks
          */
 
+        /**
+         * The number of minor ticks per major tick. Works for `linear`,
+         * `logarithmic` and `datetime` axes.
+         *
+         * @sample {highcharts} highcharts/xaxis/minortickspermajor/
```
TorsteinHonsi, line 1285: The sample path says `xaxis` while the demo actually shows `yaxis`. Better to rename the sample.

samples/highcharts/xaxis/minortickspermajor/demo.js
```diff
@@ -0,0 +1,10 @@
+Highcharts.chart('container', {
+    yAxis: {
+        minorTicks: true,
+        minorTicksPerMajor: 3
```
TorsteinHonsi, line 4: This is technically correct, but can we use `minorTicksPerMajor: 2` instead? Since we now have minor ticks at 1/3 and 2/3 which is not a common real-live use case.

samples/highcharts/xaxis/minortickspermajor/demo.js
```diff
@@ -0,0 +1,10 @@
+Highcharts.chart('container', {
+    yAxis: {
```
TorsteinHonsi, line 2: Add a descriptive title to the demo. For example `<em>minorTicksPerMajor</em> is 2`.

TorsteinHonsi (changes requested):

TorsteinHonsi (approved):

PR #18543, opened 2023-02-21
db/native-css

karolkolodziej:
Moved styling to `.css` files from `.scss`.


css/dashboards/lightDarkTheme.css
```diff
@@ -5,8 +5,8 @@
  *
  * License: www.highcharts.com/license
  */
- :root {
```
TorsteinHonsi, line 8: I don't think this file should be in the repo. In its common form, it will be uploaded to code.highcharts.com.

If the intension is to use this in a single demo, add the CSS rule to `demo.css` instead.

If however the intension is to provide default dark theme functionality, it must use the [same pattern as we are introducing in Highcharts v11](https://github.com/highcharts/highcharts/blob/22ac18a7e6ba2f250ff910fe964602f5455b8969/css/highcharts.css#L9-L128).

TorsteinHonsi (changes requested):

sebastianbochan (approved):

css/dashboards/lightDarkTheme.css
karolkolodziej: We have it as a separate task but I guess we can do it here.

TorsteinHonsi: Thanks. Also notice the forced color modes in the main file. `.highcharts-dark` and `.highcharts-light`. In addition, the proposed change in `gui.css` now only sets the background color, but there's more to a dark theme than that. As a minimum, all background colors must be dark and all text colors be white.

karolkolodziej: These classes to force light/dark themes are fine and we can implement themes here.
When it comes to adding other elements like text, buttons, and background we should implement them separately in other PR- we have it for the next sprint.

sebastianbochan (approved):

goransle (approved):
👍

Issue #17080, opened 2022-03-10
Wrong position of dataLabels with textpath on inverted chart


pawellysy:
#### Expected behaviour
The `dataLabels` with `textPath` should be in the correct place

#### Actual behaviour
The `dataLabels` with `textPath` are positioned in a wrong position

#### Live demo with steps to reproduce

https://jsfiddle.net/BlackLabel/fw1uqsnh/1/

#### Product version
HC v10.0.0

#### Affected browser(s)
all

PR #18546, opened 2023-02-21
feature/18423-organization-chart-link-labels

### LINKED ISSUE #18423: Organization Chart link labels
I would like to add text on the "links" from one node to the next in an organization chart. I can't seem to find in the docs how to format the links with custom labels.

Thank you!

jakubSzuminski:
Implemented link data labels for OrganizationSeries. See #18423.

Closes #18423.

@TorsteinHonsi In arcdiagram, link labels look like this: https://jsfiddle.net/BlackLabel/9zwe78x1/
But arcdiagram renders link labels as SVG elements and it uses `linkTextPath`.

Organization chart `dataLabels` are rendered as HTML elements so the easiest way to achieve a similar effect would be to also use `linkTextPath` which only works if `useHTML = false`.

Then, the link dataLabels would look like here: https://jsfiddle.net/BlackLabel/pmtgenqb/
I think that otherwise it will be quite difficult to make them look well.

Should I proceed with this solution & add a short note to the API that `linkFormat` renders `dataLabel`s  for the Organization chart only if `dataLabels.useHTML = false`?

Or should I do something else about it?

### DIFF
--- a/samples/highcharts/series-organization/link-labels/demo.js
+++ b/samples/highcharts/series-organization/link-labels/demo.js
@@ -13,10 +13,7 @@ Highcharts.chart('container', {
         dataLabels: {
             linkFormat: 'from {point.from} to {point.to}',
             linkTextPath: {
-                enabled: true,
-                attributes: {
-                    startOffset: '80%'
-                }
+                enabled: true
             }
         }
     }]
--- a/samples/highcharts/series-organization/link-labels/demo.js
+++ b/samples/highcharts/series-organization/link-labels/demo.js
@@ -10,9 +10,6 @@ Highcharts.chart('container', {
             ['A', 'C'],
             ['C', 'D']
         ],
-        linkRadius: 0,
-        linkLineWidth: 2,
-        linkColor: '#ccc',
         link: {
             type: 'curved'
         },
--- a/samples/highcharts/series-organization/link-labels/demo.js
+++ b/samples/highcharts/series-organization/link-labels/demo.js
@@ -14,7 +14,9 @@ Highcharts.chart('container', {
             linkFormat: 'from {point.from} to {point.to}',
             linkTextPath: {
                 enabled: true,
-                startOffset: '80%'
+                attributes: {
+                    startOffset: '80%'
+                }
             }
         }
     }]
--- a/samples/highcharts/series-organization/link-labels/demo.js
+++ b/samples/highcharts/series-organization/link-labels/demo.js
@@ -13,7 +13,8 @@ Highcharts.chart('container', {
         dataLabels: {
             linkFormat: 'from {point.from} to {point.to}',
             linkTextPath: {
-                enabled: true
+                enabled: true,
+                startOffset: '80%'
             }
         }
     }]
--- a/samples/highcharts/series-organization/link-labels/demo.js
+++ b/samples/highcharts/series-organization/link-labels/demo.js
@@ -10,9 +10,6 @@ Highcharts.chart('container', {
             ['A', 'C'],
             ['C', 'D']
         ],
-        link: {
-            type: 'curved'
-        },
         dataLabels: {
             linkFormat: 'from {point.from} to {point.to}',
             linkTextPath: {
--- a/samples/highcharts/series-organization/link-labels/demo.js
+++ b/samples/highcharts/series-organization/link-labels/demo.js
@@ -13,9 +13,10 @@ Highcharts.chart('container', {
         linkRadius: 0,
         linkLineWidth: 2,
         linkColor: '#ccc',
+        link: {
+            type: 'curved'
+        },
         dataLabels: {
-            useHTML: false,
-            color: '#000000',
             linkFormat: 'from {point.from} to {point.to}',
             linkTextPath: {
                 enabled: true
--- a/samples/highcharts/series-organization/link-labels/demo.js
+++ b/samples/highcharts/series-organization/link-labels/demo.js
@@ -1,0 +1,25 @@
+Highcharts.chart('container', {
+    title: {
+        text: 'Organization chart with link labels'
+    },
+    series: [{
+        type: 'organization',
+        keys: ['from', 'to'],
+        data: [
+            ['A', 'B'],
+            ['A', 'C'],
+            ['C', 'D']
+        ],
+        linkRadius: 0,
+        linkLineWidth: 2,
+        linkColor: '#ccc',
+        dataLabels: {
+            useHTML: false,
+            color: '#000000',
+            linkFormat: 'from {point.from} to {point.to}',
+            linkTextPath: {
+                enabled: true
+            }
+        }
+    }]
+});
--- a/ts/Series/Organization/OrganizationDataLabelOptions.d.ts
+++ b/ts/Series/Organization/OrganizationDataLabelOptions.d.ts
@@ -23,6 +23,7 @@ import type {
     SankeyDataLabelFormatterContext,
     SankeyDataLabelOptions
 } from '../Sankey/SankeyDataLabelOptions';
+import type DataLabelTextPathOptions from '../../Core/Series/DataLabelOptions';
 
 /* *
  *
@@ -48,6 +49,7 @@ export interface OrganizationDataLabelFormatterContext extends SankeyDataLabelFo
 export interface OrganizationDataLabelOptions extends SankeyDataLabelOptions {
     nodeFormatter?: OrganizationDataLabelsFormatterCallbackFunction;
     linkFormat?: string;
+    linkFormatter?: OrganizationDataLabelsFormatterCallbackFunction;
 }
 
 export default OrganizationDataLabelOptions;
--- a/ts/Series/Organization/OrganizationDataLabelOptions.d.ts
+++ b/ts/Series/Organization/OrganizationDataLabelOptions.d.ts
@@ -23,7 +23,6 @@ import type {
     SankeyDataLabelFormatterContext,
     SankeyDataLabelOptions
 } from '../Sankey/SankeyDataLabelOptions';
-import type DataLabelTextPathOptions from '../../Core/Series/DataLabelOptions';
 
 /* *
  *
--- a/ts/Series/Organization/OrganizationDataLabelOptions.d.ts
+++ b/ts/Series/Organization/OrganizationDataLabelOptions.d.ts
@@ -47,6 +47,7 @@ export interface OrganizationDataLabelFormatterContext extends SankeyDataLabelFo
 
 export interface OrganizationDataLabelOptions extends SankeyDataLabelOptions {
     nodeFormatter?: OrganizationDataLabelsFormatterCallbackFunction;
+    linkFormat?: string;
 }
 
 export default OrganizationDataLabelOptions;
--- a/ts/Series/Organization/OrganizationPoint.ts
+++ b/ts/Series/Organization/OrganizationPoint.ts
@@ -21,7 +21,7 @@
 import type OrganizationPointOptions from './OrganizationPointOptions';
 import type OrganizationSeries from './OrganizationSeries';
 import type { OrganizationSeriesNodeOptions } from './OrganizationSeriesOptions';
-import SankeyPoint from './../Sankey/SankeyPoint';
+import type SankeyPoint from './../Sankey/SankeyPoint';
 import SeriesRegistry from '../../Core/Series/SeriesRegistry.js';
 const {
     seriesTypes: {
--- a/ts/Series/Organization/OrganizationPoint.ts
+++ b/ts/Series/Organization/OrganizationPoint.ts
@@ -21,7 +21,7 @@
 import type OrganizationPointOptions from './OrganizationPointOptions';
 import type OrganizationSeries from './OrganizationSeries';
 import type { OrganizationSeriesNodeOptions } from './OrganizationSeriesOptions';
-import type SankeyPoint from './../Sankey/SankeyPoint';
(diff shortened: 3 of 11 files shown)


pawellysy (approved):
Looks good! 👍

TorsteinHonsi (changes requested):
Thanks! Please add a visual demo so we can see the new feature demonstrated. Then add `@sample` reference to the appropriate doclet for implementers to see.

pawellysy (changes requested):
Let's try adding the `useHTML: false` option for links only, when linkTextPath is enabled.

samples/highcharts/series-organization/link-labels/demo.js
```diff
@@ -0,0 +1,26 @@
...
+        type: 'organization',
+        keys: ['from', 'to'],
+        data: [
+            ['A', 'B'],
+            ['A', 'C'],
+            ['C', 'D']
+        ],
+        linkRadius: 0,
+        linkLineWidth: 2,
+        linkColor: '#ccc',
```
pawellysy, line 15: not necessary in the demo

ts/Series/Organization/OrganizationDataLabelOptions.d.ts
```diff
@@ -47,6 +47,7 @@ export interface OrganizationDataLabelFormatterContext extends SankeyDataLabelFo
 
 export interface OrganizationDataLabelOptions extends SankeyDataLabelOptions {
     nodeFormatter?: OrganizationDataLabelsFormatterCallbackFunction;
+    linkFormat?: string;
```
pawellysy, line 50: how about `linkFormatter`? Shouldn't it also be included?

ts/Series/Organization/OrganizationPoint.ts
```diff
@@ -21,7 +21,7 @@
 import type OrganizationPointOptions from './OrganizationPointOptions';
 import type OrganizationSeries from './OrganizationSeries';
 import type { OrganizationSeriesNodeOptions } from './OrganizationSeriesOptions';
-import type SankeyPoint from './../Sankey/SankeyPoint';
+import SankeyPoint from './../Sankey/SankeyPoint';
```
pawellysy, line 24: SankeyPoint is used as a type in this file, so please restore this change here. 

(If the full import is used, it increases the bundle size :) )

ts/Series/Organization/OrganizationSeriesDefaults.ts
```diff
@@ -412,6 +412,20 @@ const OrganizationSeriesDefaults: OrganizationSeriesOptions = {
  * @apioption series.organization.nodes.description
  */
 
+/**
+ * The format string specifying what to show for *links* in the
+ * organization chart.
+ *
+ * Best to use with `linkTextPath` enabled.
```
pawellysy, line 419: The `linkTextPath` should also be added as an API option.

ts/Series/Sankey/SankeySeries.ts
```diff
@@ -335,8 +335,16 @@ class SankeySeries extends ColumnSeries {
     }
 
     public drawDataLabels(): void {
-        ColumnSeries.prototype.drawDataLabels.call(this, this.points);
-        ColumnSeries.prototype.drawDataLabels.call(this, this.nodes);
+        for (const link of this.points) {
+            const dlOptions = this.options.dataLabels;
```
pawellysy, line 339: this const is the same for all the links, so can be defined outside the scope of the for loop.

pawellysy (changes requested):
Great work, thanks!
Minor nitpicking here and there 😅

ts/Series/Sankey/SankeySeries.ts
```diff
@@ -335,8 +335,16 @@ class SankeySeries extends ColumnSeries {
     }
 
     public drawDataLabels(): void {
-        ColumnSeries.prototype.drawDataLabels.call(this, this.points);
-        ColumnSeries.prototype.drawDataLabels.call(this, this.nodes);
+        for (const link of this.points) {
```
pawellysy, line 338: This modification was supposed to be added to `OrganizationSeries.drawDataLabels` 😅 
Please restore the original Sankey method and move the correct class :)

ts/Series/Organization/OrganizationDataLabelOptions.d.ts
pawellysy: Also, `linkTextPath`. (For the pattern, see `TreegraphDataLabelOptions` in `TreegraphSeriesOptions.d.ts`)

pawellysy (changes requested):
Few additional comments

ts/Series/Organization/OrganizationDataLabelOptions.d.ts
jakubSzuminski: You're right, I've included `linkFormatter`. But should I include the `linkTextPath` if it exists on `SankeyDataLabelOptions`?

pawellysy: Right, Not necessary! Thanks!

ts/Series/Organization/OrganizationSeries.ts
```diff
@@ -410,6 +410,20 @@ class OrganizationSeries extends SankeySeries {
             shapeArgs.height;
     }
 
+    public drawDataLabels(): void {
+        const dlOptions = this.options.dataLabels;
+
+        if (dlOptions && dlOptions.linkTextPath &&
```
pawellysy, line 416:
```suggestion
        if (dlOptions.linkTextPath &&
```

DlOptions always is defined after your change in `OrganizationSeriesOptions.d.ts`

ts/Series/Organization/OrganizationSeriesDefaults.ts
```diff
@@ -424,6 +426,52 @@ const OrganizationSeriesDefaults: OrganizationSeriesOptions = {
...
+ *
+ * Defaults to `undefined`.
+ *
+ * @sample highcharts/series-organization/link-labels
+ *         Organization chart with link labels
+ *
+ * @type      {string}
+ * @product   highcharts
+ * @apioption series.organization.dataLabels.linkFormat
+ * @since 10.3.3
```
pawellysy, line 443: if we are adding the option now, it will be available from the next release.
```suggestion
 * @since next
```

ts/Series/Organization/OrganizationSeriesDefaults.ts
```diff
@@ -424,6 +426,52 @@ const OrganizationSeriesDefaults: OrganizationSeriesOptions = {
...
+ * Callback to format data labels for _links_ in the
+ * organization chart. The `linkFormat` option takes
+ * precedence over the `linkFormatter`.
+ *
+ * Defaults to `undefined`.
+ *
+ * @type      {OrganizationDataLabelsFormatterCallbackFunction}
+ * @product   highcharts
+ * @apioption series.organization.dataLabels.linkFormatter
+ * @since 10.3.3
```
pawellysy, line 456:
```suggestion
 * @since next
```

ts/Series/Organization/OrganizationSeriesDefaults.ts
```diff
@@ -424,6 +426,52 @@ const OrganizationSeriesDefaults: OrganizationSeriesOptions = {
...
+ *
+ * **Note:** Only SVG-based renderer supports this option.
+ *
+ * @sample highcharts/series-organization/link-labels
+ *         Organization chart with link labels
+ *
+ * @type { DataLabelTextPathOptions }
+ * @product highcharts
+ * @apioption series.organization.dataLabels.linkTextPath
+ * @since 10.3.3
```
pawellysy, line 472:
```suggestion
 * @since next
```

ts/Series/Organization/OrganizationSeriesDefaults.ts
```diff
@@ -28,6 +29,7 @@ import type Point from '../../Core/Series/Point';
 import type { SankeyDataLabelFormatterContext } from '../Sankey/SankeyDataLabelOptions';
 
 import { Palette } from '../../Core/Color/Palettes.js';
+import { DataLabelTextPathOptions } from '../../Core/Series/DataLabelOptions';
```
pawellysy, line 32: import as type

pawellysy (changes requested):
Just a small polishing left 😅

pawellysy (approved):
Great job! Thanks!

ts/Series/Organization/OrganizationSeries.ts
```diff
@@ -410,6 +410,19 @@ class OrganizationSeries extends SankeySeries {
             shapeArgs.height;
     }
 
+    public drawDataLabels(): void {
+        const dlOptions = this.options.dataLabels;
+
+        if (dlOptions.linkTextPath && dlOptions.linkTextPath.enabled) {
+            for (const link of this.points) {
+                link.options.dataLabels = merge(link.options.dataLabels,
+                    { useHTML: false });
```
TorsteinHonsi, line 419: This exception for `useHTML` - is that specific for org series? If not, we should refactor this so that it is handled for every series type that can use `textPath` for data labels, without repeating the code. It may be that it should be handled in the SVG renderer itself.

ts/Series/Organization/OrganizationSeriesDefaults.ts
```diff
@@ -424,6 +426,48 @@ const OrganizationSeriesDefaults: OrganizationSeriesOptions = {
...
+ * @apioption series.organization.dataLabels.linkFormatter
+ * @since next
+ */
+
+/**
+ * Options for a _link_ label text which should follow link
+ * connection. Border and background are disabled for a label
+ * that follows a path.
+ *
+ * **Note:** Only SVG-based renderer supports this option.
```
TorsteinHonsi, line 460: This can be removed, as the VML renderer is being removed.

TorsteinHonsi (changes requested):
Thanks! 

Looking at the demo, I noticed two things:
* The links are rendered as `curve`, while the default for org series is not curve. I think it is better to keep the default, and instead move the label to the end of the link: https://jsfiddle.net/m9f3hgLp/
* Most org charts are inverted, but in this case the link labels don't show up at all: https://jsfiddle.net/m9f3hgLp/1/ . It renders something when adjusting the `startOffset`, but not in the right place: https://jsfiddle.net/m9f3hgLp/2/

jakubSzuminski:
Thanks @TorsteinHonsi!

Link labels don't show up for an inverted organization chart, but it seems to be a problem with `textPath` for `chart.inverted`, not anything specific to Organization series. Here's an open issue: [#17080](https://github.com/highcharts/highcharts/issues/17080)

What do you think we should do about this? Should we proceed with this feature and then try to fix the problem of `textPath` for `chart.inverted`?

ts/Series/Organization/OrganizationSeries.ts
jakubSzuminski: We wanted to render the entire org series with `useHTML = true` by default, but if the user enables `linkTextPath` then we turn off the `useHTML` only for the links while the rest of the `dataLabels` are rendered with `useHTML = true`.

Do you think that whenever the `textPath.enabled = true` we should automatically disable `useHTML`? 

If yes, then maybe I'll create another issue for it & list all the names of the series which need to have it changed. This note: `Note: Only SVG-based renderer supports this option. Setting useHTML to true will disable this option.` should be removed from the docs for all of them.

TorsteinHonsi: Okay, I understand what you are thinking. Org charts are special because we encourage `useHTML`. Let's just keep it as it is then.

TorsteinHonsi: > Do you think that whenever the textPath.enabled = true we should automatically disable useHTML?

Can you follow that lead for a bit? If we can do that in a simple way, it would probably be the best way.

jakubSzuminski: > Do you think that whenever the textPath.enabled = true we should automatically disable useHTML?

I'm sorry I wasn't clear. I meant to ask if we should automatically disable `useHTML` if `textPath.enabled = true` for other series than organization chart. For example, for pie series, if `series.pie.dataLabels.textPath.enabled`, should we automatically set `series.pie.dataLabels.useHTML` to `false`? 

But you're right - in org charts we encourage `useHTML` and that's why I think it'll be fine if we do this exception for `useHTML` only here for now.

jakubSzuminski:
So we leave this `useHTML` exception only for the org series.

And [#17080](https://github.com/highcharts/highcharts/issues/17080) issue needs to be taken care of in the future to make the `chart.inverted` work well. 

But for now, what else is there that I should do/fix with this feature?

TorsteinHonsi:
> But for now, what else is there that I should do/fix with this feature?

@jakubSzuminski I think the only thing left is that the org chart link labels should have a sensible _default_ position. That means instead of setting the start offset in the demo, it should be in the series types options itself. Because org chart links typically have a problematic middle part. I think this would be good defaults:

```js
attributes: {
    startOffset: '95%',
    textAnchor: 'end'
}
```

TorsteinHonsi (changes requested):
> But for now, what else is there that I should do/fix with this feature?

@jakubSzuminski I think the only thing left is that the org chart link labels should have a sensible _default_ position. That means instead of setting the start offset in the demo, it should be in the series types options itself. Because org chart links typically have a problematic middle part. I think this would be good defaults:

```js
attributes: {
    startOffset: '95%',
    textAnchor: 'end'
}
```

TorsteinHonsi (approved):

PR #18549, opened 2023-02-21
bugfix/17589-multiple-datalabels-update

### LINKED ISSUE #17589: Multiple dataLabels update
#### Expected behaviour
Multiple `dataLabels` should be updated correctly.

#### Actual behaviour
Multiple `dataLabels` behave strangely when updating them. For example, when you try to disable the second label, the first one is disabled instead of it

https://jsfiddle.net/BlackLabel/fkjyepsx/

#### Product version
Highcharts 10.2.0

hubertkozik:
Fixed #17589, update of multiple data labels didn't work in specific conditions.


TorsteinHonsi (commented):
Looking at the changes, I am wondering if we can simplify this a bit.

We know that `point.dataLabels` should always be an array, and that `point.dataLabel` should always be the same as `point.dataLabels[0]`. If this is not the case, it should be considered a bug. (Or am I wrong? Check `DataLabel.ts`, line 563).

So say the assumption above is true, or we make it true. In that case, all the juggling between `point.dataLabel` and `point.dataLabels` and checking `i === 0` can be replaced with a structure like this:

```ts
for (let dataLabel of point.dataLabels) {
    // Do stuff to each data label. Destroy if that's what we want, but remember
    // to splice it out from `point.dataLabels`
}

// Write back to `point.dataLabel`. `point.dataLabels[0]` can now be either
// an SVGElement or undefined, both of which are correct.
point.dataLabel = point.dataLabels[0];
```

hubertkozik:
@TorsteinHonsi It seems to me, that it won't be so simple. In the place, where I have added a fix we are already in a loop through data labels for each point. In my opinion, to change the logic (to use `point.dataLabel = point.dataLabels[0];`) we would need to refactor this entire function (`drawDataLabels`). Unfortunately, the current code looks to me a bit like spaghetti...

PR #18551, opened 2023-02-21
bugfix/18544-boost-null-values-data-length

### LINKED ISSUE #18544: Boost: with seriesThreshold doesn't calculate null values as array elements.
#### Expected behaviour
The length of data array should be calculated with null values.

#### Actual behaviour
With `boost.seriesThreshold` enabled, doesn't calculate null values as array elements.

http://jsfiddle.net/BlackLabel/pgj0Lq89/

#### Product version
v10.3.3 (2023-01-20)

#### Affected browser(s)
All

hihajdus:
Fixed #18544, data length for null values with [boost.seriesThreshold](https://api.highcharts.com/highcharts/boost.seriesThreshold).

### DIFF
--- a/samples/unit-tests/boost/general/demo.js
+++ b/samples/unit-tests/boost/general/demo.js
@@ -117,7 +117,7 @@ QUnit[Highcharts.hasWebGLSupport() ? 'test' : 'skip'](
     }
 );
 QUnit[Highcharts.hasWebGLSupport() ? 'test' : 'skip'](
-    'Combination with non-boostable series types (#7634)',
+    'Combination with non-boostable series types and null values (#7634)',
     function (assert) {
         var chart = Highcharts.chart('container', {
             boost: {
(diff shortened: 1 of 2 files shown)


pawellysy (commented):
Thanks!

Not quite sure though, about this change. We are using Boost module, when there is a problem with performance, so maybe stripping null points is a feature, not a bug?
Apart from the fact, that the change is about deleting the check inside the `if` statement, which must have been added deliberately.

The decision is @TorsteinHonsi's :D

hihajdus:
Thanks, yes please comments I also not sure about this case.

TorsteinHonsi:
At least it doesn't seem to cause problems with how the Boost is rendered. But what's the rationale for this change? In what scenario do we need this data length to include null points?

hihajdus:
In the scenario: 
inside `tooltip.formatter` we want to know how many points are in the selected series and what the index of the selected point is.

TorsteinHonsi (approved):

PR #18552, opened 2023-02-21
Db/custom component

pawellysy:
Created A demo and article about creating custom components in Dashboards.

### DIFF
--- a/docs/dashboards/custom-component.md
+++ b/docs/dashboards/custom-component.md
@@ -15,11 +15,11 @@ The name of the class will automatically be the string, that will be used to ref
     }
 ```
 
-Then, depending on what the Component is expected to do, the options are limitless. In this example, one `iframe` element will be added, which will accept one attribute from options, which is `videoId`, and since the iframe element needs it size to be defined, the resize method is extended to update ths size of the element.
+Then, depending on what the Component is expected to do, the options are limitless. In this example, one `iframe` element will be added, which will accept one attribute from options, which is `videoId`, and since the iframe element needs it size to be defined, the resize method is extended to update the size of the element.
 
-At the end, the new YouTubeComponent class needs to be addded to the registry, using the `Component.addComponent` method.
+At the end, the new `YouTubeComponent` class needs to be addded to the registry, using the `Component.addComponent` method.
 
-Whole custom code looks like that:
+The custom code looks like below:
 
 ```js
 class YouTubeComponent extends Component {
--- a/docs/dashboards/custom-component.md
+++ b/docs/dashboards/custom-component.md
@@ -4,8 +4,8 @@ This article shows how to create custom Dashboards Component, in this example Yo
 
 Note, that to create the custom component we are using ES6, to use the `class` and `extends` keywords, which makes creating custom class much easier.
 
-We start by importing the default `Component` class from `Dashboards` namespace, and by creating the class, which will inherit from this class.
-The name of the class will automatically be the name, that will be used to reference this component type, e.g. the class `YouTubeComponent` will be referenced by name: `YouTube`.
+We start by importing the default `Component` class from `Dashboards` namespace. The next step is creating the class which will inherit from the imported `Component` class.
+The name of the class will automatically be the string, that will be used to reference this component type e.g. the class `YouTubeComponent` will be referenced by name: "`YouTube`".
 
 ```js
     const Component = Dashboards.Component;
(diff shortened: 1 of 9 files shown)


karolkolodziej (changes requested):
I'm unable to see it working locally due to CORS.

docs/dashboards/custom-component.md
```diff
@@ -0,0 +1,78 @@
+Custom Dashboards components
+===
+This article shows how to create custom Dashboards Component, in this example YouTube Component.
+
+Note, that to create the custom component we are using ES6, to use the `class` and `extends` keywords, which makes creating custom class much easier.
+
+We start by importing the default `Component` class from `Dashboards` namespace. The next step is creating the class which will inherit from the imported `Component` class.
+The name of the class will automatically be the string, that will be used to reference this component type e.g. the class `YouTubeComponent` will be referenced by name: "`YouTube`".
```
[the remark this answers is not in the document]
bre1470: You can choose any name. If it ends with Component, this part gets stripped away.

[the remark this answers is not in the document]
karolkolodziej: You are right! Then is the `this.type = 'YouTube';` needed?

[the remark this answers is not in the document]
bre1470: It allows identification without `instanceof`, but in this demo probably not.

docs/dashboards/custom-component.md
```diff
@@ -0,0 +1,78 @@
...
+
+```js
+    const Component = Dashboards.Component;
+
+    class YouTubeComponent extends Component {
+
+    }
+```
+
+Then, depending on what the Component is expected to do, the options are limitless. In this example, one `iframe` element will be added, which will accept one attribute from options, which is `videoId`, and since the iframe element needs it size to be defined, the resize method is extended to update ths size of the element.
```
sebastianbochan, line 18: `ths` - typo

docs/dashboards/custom-component.md
```diff
@@ -0,0 +1,78 @@
...
+
+At the end, the new YouTubeComponent class needs to be addded to the registry, using the `Component.addComponent` method.
```
sebastianbochan, line 20: Please wrap the YouTubeComponent into brackets, in will increase the readibility.

docs/dashboards/custom-component.md
```diff
@@ -0,0 +1,78 @@
...
+
+Whole custom code looks like that:
```
sebastianbochan, line 22: `The custom code looks like below` - sounds better IMO.

samples/dashboards/demos/custom-component/demo.js
```diff
@@ -0,0 +1,67 @@
...
+    constructor(options) {
+        super(options);
+        this.type = 'YouTube';
+        this.youTubeElement = document.createElement('iframe');
+        return this;
+    }
+
+    resize(width, height) {
+        super.resize.call(this, width, height);
+        this.youTubeElement.setAttribute('width', width - 10);
```
sebastianbochan, line 13: Is the `10` a padding?

Would be good to comment it ;)

samples/dashboards/demos/custom-component/demo.js
```diff
@@ -0,0 +1,67 @@
...
+    load() {
+        super.load();
+
+        this.youTubeElement.setAttribute(
+            'src',
+            'https://www.youtube.com/embed/' + this.options.videoId
+        );
+        this.youTubeElement.setAttribute('title', 'YouTube video player');
+        this.youTubeElement.setAttribute('frameborder', '0');
+        this.youTubeElement.allowfullscreen = true;
```
sebastianbochan, line 26: In this part we mix two solutions for applying params, so I would suggest to use the `setAttribute` here, too.

ts/Dashboards/Actions/Bindings.ts
```diff
@@ -62,62 +62,34 @@ class Bindings {
...
         cell?: Cell
-    ): ComponentTypes | undefined {
+    ): Component | undefined {
         const componentContainer = document.getElementById(options.cell);
         const optionsStates = options.states;
         const optionsEvents = options.events;
 
         cell = cell || Bindings.getCell(options.cell);
-        let component: ComponentTypes|undefined;
+        let component: Component|undefined;
```
sebastianbochan, line 71: I guess that there should be also extra space around the separator `|` like in other parts of the code.

sebastianbochan (changes requested):

samples/dashboards/demos/custom-component/demo.js
pawellysy: When I do it with `setAttribute` the error is thrown. Apart from that, that is only a demo :)

docs/dashboards/custom-component.md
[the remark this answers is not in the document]
pawellysy: It is done in other constructors, so I figured, that it can be added :)

karolkolodziej (approved):

sebastianbochan (approved):

PR #18557, opened 2023-02-22
db/dashboard-icons

pawellysy:
Added dashboard series types icons.


TorsteinHonsi (changes requested):
Thanks!

We want icons to appear consistent across our JS products. Looking at the [Stock Tools icons](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/stock/demo/stock-tools-gui), they are grayscale, while these new ones are colorized. Any thoughts on that?

pawelfus:
Good point. We have a small problem here then:
- greyscale icons don't fit into the Climate demo
- colorful icons may not fit into the Dashboards default theme

Would it make sense to ask Nancy to create greyscale icons and store both, colorful and greyscale on the server? If it doesn't make sense, we can store greyscale, and export colorized ones into base64 and keep them in the demo

@pawellysy - Is this possible? I mean, do we have an API option to change icons per demo?

pawellysy:
We could also use the [grayscale function](https://developer.mozilla.org/en-US/docs/Web/CSS/filter-function/grayscale) for default icons, and use the `filter: grayscale(0)` where we want to have a color. Then no need for storing extra icons

pawelfus:
Paweł just pointed out, we have other colorful icons in Dashboard anyway: 
image - which aren't in greyscale, neither in violet-ish scale

There is also another solution: resign from icons in this dropdown.

TorsteinHonsi:
So the conclusion is that these icons are _not_ specific to the climate demo? Just pointing out that there are two different concerns:

- icons belonging to the default design should be stored in `gfx/some-general-name`
- icons belonging to a specific sample (like the climate demo) should be stored in `gfx/some-specific-name-for-that-sample`

pawellysy:
@TorsteinHonsi  (CC @pawelfus and @nasin) In my opinion these icons are added based on the climate demo design, but there is no reason not to use them as a default design as well. This is why I am adding them to the general gfx, the same way we are using them in the stock tools. If we want to be consistent and not use the color icons, we can always use the grayscale modifier there. If there are some other ideas. 

Another idea is also adding both color and grayscale icons because they only take a couple hundred bytes 🤔 Let me know if there is anything to add/change here.

TorsteinHonsi (approved):
I agree, I'll add my approval.

PR #18558, opened 2023-02-22
Migration: Legend

marvin19:
Improved Legend.


ts/Core/Legend/Legend.ts
```diff
@@ -1758,6 +1767,42 @@ namespace Legend {
...
+    /**
+     * @private
+     */
+    export function compose(ChartClass: typeof Chart): void {
+        if (composedMembers.indexOf(ChartClass) === -1) {
+            composedMembers.push(ChartClass);
+            addEvent(ChartClass, 'rendering', function (): void {
+                const chart = this;
+                // Title
+                chart.setTitle();
```
TorsteinHonsi, line 1793: This looks a bit suspicious - the chart title is not related to the legend 🤔

bre1470: I thought that it might be related to the box calculations.

TorsteinHonsi: Yes, but the way it is implemented now, there will not be a chart title without the Legend module?

ts/Core/Chart/Chart.ts
```diff
@@ -2414,106 +2412,102 @@ class Chart {
...
-         * @type {Highcharts.Legend}
-         */
-        chart.legend = new Legend(chart, options.legend as any);
-
-        // Get stacks
-        if (chart.getStacks) {
-            chart.getStacks();
-        }
+            // Title
+            chart.setTitle();
```
bre1470, line 2418: Here is is called without Legend, @TorsteinHonsi .

TorsteinHonsi: So it's called _twice_ when the Legend is imported? 🤔

bre1470: Yes, correct.

bre1470 (approved):
Thank you! Looks good. 👍

ts/Core/Chart/Chart.ts
TorsteinHonsi: Okay, so the reason for this is that we wanted to wrap the _whole_ rendering logic inside one event handler, and this `setTitle` call was done before the legend, right?

If so, I think we should rather create an event call after setting the title but before `getMargins()`, where we could hook in the legend creation. 

In this PR, there's also an `afterRendering` event that is not used by any component.

bre1470: > Okay, so the reason for this is that we wanted to wrap the _whole_ rendering logic inside one event handler, and this `setTitle` call was done before the legend, right?

Yes, and it is needed. We tested with changed position and it started to fail.

> If so, I think we should rather create an event call after setting the title but before `getMargins()`, where we could hook in the legend creation.

Do you mean that we move `setTitle` in front of the default handler, or do we have no default handler at all?

> In this PR, there's also an `afterRendering` event that is not used by any component.

`render` and `afterRender` are hijacked by redraw and update. Therefor we had to go with `rendering` / `afterRendering`.

TorsteinHonsi: > do we have no default handler at all?

Yes, that's what I think. In this case we should have a detailed event that is called at just the right moment. Like we do for example with `beforeRender`

https://github.com/highcharts/highcharts/blob/a905c72df5599de95cf9a8fd040d59338dde530c/ts/Core/Chart/Chart.ts#L2707-L2711

TorsteinHonsi (approved):

PR #18559, opened 2023-02-22
bugfix/16516-legend-symbol-height-zero-value

### LINKED ISSUE #16516: Legend.symbolHeight do not accept 0
#### Expected behaviour
No visible symbol when `symbolHeight` is set to `0`.

#### Actual behaviour
https://github.com/highcharts/highcharts/blob/ff6ddd41f6df9e08ffe9f19471c5a2ef91f7774b/ts/Core/Legend/Legend.ts#L745-L746

https://jsfiddle.net/BlackLabel/5u18nvym/

#### Workaround
Use a value close to `0` instead - so something like `0.001` will be ok.
https://jsfiddle.net/BlackLabel/5u18nvym/1/

hubertkozik:
Fixed #16516, legend symbol height didn't accept 0 as a value.

### DIFF
--- a/ts/Core/Legend/Legend.ts
+++ b/ts/Core/Legend/Legend.ts
@@ -40,9 +40,7 @@ import F from '../FormatUtilities.js';
 const { format } = F;
 import H from '../Globals.js';
 const {
-    isFirefox,
-    marginNames,
-    win
+    marginNames
 } = H;
 import Point from '../Series/Point.js';
 import R from '../Renderer/RendererUtilities.js';
@@ -61,8 +59,7 @@ const {
     pick,
     relativeLength,
     stableSort,
-    syncTimeout,
-    wrap
+    syncTimeout
 } = U;
 
 /* *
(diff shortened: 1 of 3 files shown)


ts/Core/Legend/Legend.ts
```diff
@@ -738,7 +738,8 @@ class Legend {
                 label.attr('y', legend.baseline);
 
                 legend.symbolHeight =
-                    options.symbolHeight || legend.fontMetrics.f;
+                    isNumber(options.symbolHeight) ?
```
TorsteinHonsi, line 741: We can use `pick` here. 

```ts
legend.symbolHeight = pick(options.symbolHeight, legend.fontMetrics.f);
```

TorsteinHonsi:
A related issue - I recently wanted to remove the legend item for a `proximate` layout. My gut feeling was to set `symbolWidth` to 0, and that this would remove the legend item. But that didn't work, there is still a small dot. See https://jsfiddle.net/highcharts/xk7t6nu8/. 

So can we extend this PR to also handle `symbolWidth: 0`, which should result in no rendered symbol?

hubertkozik:
Sure! Let's extend it! I will take a look into it.

PR #18566, opened 2023-02-23
demo/tooltip-follow-pointer

jakubjanuchta:
Updated tooltip follow pointer demo and added to API.


TorsteinHonsi (approved):
Nice one, thanks!

PR #18567, opened 2023-02-23
bugfix/18528-lastPrice-label-calculation

### LINKED ISSUE #18528: Highstocks: Current Price Indicator Label show the wrong point after changing chart type.
#### Expected behaviour
Price indicator should show price form last visible point.

#### Actual behaviour
After changing chart type, current price indicator show a label for the wrong point (before last instead of last).

https://user-images.githubusercontent.com/22296099/220066853-eaed1684-79f6-4b41-8da1-409e6c7a756d.mov

https://jsfiddle.net/BlackLabel/f7sc854n/

#### Product version
Highstock JS v10.3.3 (2023-01-20)

#### Affected browser(s)
All

kamil-musialowski:
Fixed #18528, `lastVisiblePrice` sometimes got renderer on a second to last visible point.

### DIFF
--- a/ts/Extensions/PriceIndication.ts
+++ b/ts/Extensions/PriceIndication.ts
@@ -15,6 +15,7 @@ import type SVGElement from '../Core/Renderer/SVG/SVGElement';
 import Series from '../Core/Series/Series.js';
 import U from '../Core/Utilities.js';
 import ColorType from '../Core/Color/ColorType';
+import Point from '../Core/Series/Point';
 const {
     addEvent,
     isArray,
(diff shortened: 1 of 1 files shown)


ts/Extensions/PriceIndication.ts
```diff
@@ -372,14 +373,25 @@ addEvent(Series, 'afterRender', function (): void {
         }
 
         if (lastVisiblePrice && lastVisiblePrice.enabled && pLength > 0) {
-            crop = (points[pLength - 1].x === x) || pointRange === null ? 1 : 2;
```
pawelfus, line 375: A few lines above you can remove `pointRange` definition (line 311)

We need to clean up the leftovers manually (it is disabled in eslinter afaik), sorry

pawelfus: PS: What happens when `pointRange === null` in your solution? Does it even matter?

pawelfus: That reminds me of `cropShoulder` property. Could you check this for me? 
`cropShoulder = 1`  for line-based series and `cropShoulder = 0` for column-based series. That property defines how many extra points are rendered. My guess is we should use this value to determine the "last point" when data is cropped (`series.cropped` property). That would be faster than looping over all points.

I mean, otherwise instead of looping over, why won't we simply do this:
```ts
const lastPoint = points[length - 1].isInside ? points[length - 1] : points[length - 2]
```

pawelfus (changes requested):
Q:
Was the previous test redundant? To me, it seems the old test used to test if the label is visible. The bug you are fixing right now is more about the content of the label, isn't it? 🤔

ts/Extensions/PriceIndication.ts
kamil-musialowski: I've tested the `pointRange` property while implementing this solution, and it seems like everything works fine with the `pointRange === null` and with other numeric values.

TorsteinHonsi (approved):

PR #18580, opened 2023-02-27
bugfix/18520-legend-item-font-size-issue

### LINKED ISSUE #18520: Legend item does not scale correctly when using font style with rem units
#### Expected behaviour
I would expect the legend item icons to scale proportionally to the legend text size when using a style like e.g.
```
font: "1rem Arial, \"Liberation Sans\", Freesans, sans-serif",
```

#### Actual behaviour
The legend icon keeps its default size unless we also specify:

```
fontSize: '1rem'
```

https://jsfiddle.net/846tfykc/

#### Product version
Might affect more but at least affects versions 8+

#### Affected browser(s)
Chromium (also Chrome, Edge) and Firefox

I think the issue is that in the absense of a `fontSize` option in the legend `itemStyle`, it uses a default of 12px font size to calculate the size of the legend icons.

jakubSzuminski:
Fixed #18520:

If `fontSize` property is not defined by the user, but the `font` property is, the `legend.fontMetrics` are given undefined for the first argument (to inherit `fontSize` from DOM). 

If neither `fontSize` nor `font` are defined by the user, we pass the default `fontSize` to the `legend.fontMetrics`.

### DIFF
--- a/ts/Core/Legend/Legend.ts
+++ b/ts/Core/Legend/Legend.ts
@@ -729,8 +729,25 @@ class Legend {
             // Get the baseline for the first item - the font size is equal for
             // all
             if (!legend.baseline) {
+                const legendUserOptions = legend.chart.userOptions.legend;
+                let fontSize: string | number | undefined;
+
+                if (legendUserOptions &&
+                    legendUserOptions.itemStyle &&
+                    legendUserOptions.itemStyle.fontSize) {
+                    // if the fontSize is defined by the user
+                    fontSize = legendUserOptions.itemStyle.fontSize;
+                } else if (!legendUserOptions ||
+                    !legendUserOptions.itemStyle ||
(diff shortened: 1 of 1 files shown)


TorsteinHonsi:
Thanks!

The problem here lies on a deeper level, so even if we accepted this fix, we would have to provide similar solutions to other places the label metrics are used.

However, I have been refactoring the whole `fontMetrics` because we want to use `rem` and `em` by default for the next Highcharts version. Now it doesn't look at the options at all, only at the actual size of the rendered object. See https://github.com/highcharts/highcharts/pull/18562 . 

This fixes the OP chart 👇 
Skjermbilde 2023-02-27 kl  13 10 09

The reason for all the other labels becoming bigger is that all default font sizes are now given as `em`, and the OP chart uses `50px` font size on the root element.

PR #18589, opened 2023-02-28
Tools: Skip test_oldies in Circle

TorsteinHonsi:
We don't need it and it's causing BrowserStack hiccups


bre1470 (approved):

bre1470:
Nice! That will reduce the load a bit and maybe help with the disconnect issue. 🤞

PR #18603, opened 2023-03-02
db/eslint-dashboards

karolkolodziej:
Added a separate eslint config for dashboards.

### DIFF
--- a/ts/Dashboards/.eslintrc
+++ b/ts/Dashboards/.eslintrc
@@ -1,0 +1,10 @@
+{
+    "extends": "../.eslintrc",
+    "rules": {
+      "jsdoc/require-description": 0,
+      "jsdoc/require-jsdoc": 0,
+      "jsdoc/require-param": 0,
+      "jsdoc/require-param-description": 0,
+      "jsdoc/require-param-type": 0
+    }
+}
--- a/ts/Dashboards/.eslintrc
+++ b/ts/Dashboards/.eslintrc
@@ -1,10 +1,10 @@
 {
     "extends": "../.eslintrc",
     "rules": {
-      "jsdoc/require-description": 0,
-      "jsdoc/require-jsdoc": 0,
       "jsdoc/require-param": 0,
       "jsdoc/require-param-description": 0,
-      "jsdoc/require-param-type": 0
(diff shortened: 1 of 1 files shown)


goransle (commented):
If we're using tsdoc, should we add https://www.npmjs.com/package/eslint-plugin-tsdoc as well here?

ts/Dashboards/.eslintrc
```diff
@@ -0,0 +1,10 @@
+{
+    "extends": "../.eslintrc",
+    "rules": {
+      "jsdoc/require-description": 0,
+      "jsdoc/require-jsdoc": 0,
```
bre1470, line 5: These still have to be warnings as the source code has to be documented.

bre1470:
> If we're using tsdoc, should we add https://www.npmjs.com/package/eslint-plugin-tsdoc as well here?

Does TypeDoc use TSDoc?

goransle:
> > If we're using tsdoc, should we add https://www.npmjs.com/package/eslint-plugin-tsdoc as well here?
> 
> Does TypeDoc use TSDoc?

Looks like yes https://typedoc.org/guides/doccomments/#tsdoc-compliance. But I was actually confused by the similar names 😓

bre1470:
> > > If we're using tsdoc, should we add https://www.npmjs.com/package/eslint-plugin-tsdoc as well here?
> > 
> > 
> > Does TypeDoc use TSDoc?
> 
> Looks like kind of sort of yes https://typedoc.org/guides/doccomments/#tsdoc-compliance. But I was actually confused by the similar names 😓

Hmm, they write: "TypeDoc aims to be compliant with the TSDoc standard, but does not enforce it. This means that while TypeDoc should be able to parse all (or nearly all) TSDoc-complaint comments, it does not require that your comments follow the standard."

For flexibility we should keep the setup simple. It is a temporary solution anyway.

pawellysy (commented):
Not relevant maybe we could: 
```json
"max-len": {
    "code": 120
    }
 ```
 ? 😅

bre1470 (approved):
Thank you! Looks good. 👍

pawellysy (approved):
LGTM!

PR #18604, opened 2023-03-02
bugfix/17280-zero-z-bubbles-disappear

### LINKED ISSUE #17280: Bubble series bounded by z=0 causes points to disappear
#### Expected behaviour
Points in bubble series with z = 0 should be handled as any other point when computing the range of z values used for rendering.

#### Actual behaviour
When plotting multiple bubble series, a series whose minimum z value is 0 causes the computation of the global `z` scale to fail. This results in points falling out of the range and not being plotted at all.

I believe the issue is in those lines: https://github.com/highcharts/highcharts/blob/v10.1.0/ts/Series/Bubble/BubbleSeries.ts#L469-L470  If the running `zMin` or `zMax` is 0, the running minimum/maximum is forgotten and replaced with the current series' minimum/maximum in the loop.

https://jsfiddle.net/5ns2ozma/

#### Product version
Highcharts v10.1.0

#### Affected browser(s)
Firefox, Chrome

Vazonik:
Fixed #17280, bubbles with `z: 0` caused points to disappear.

### DIFF
--- a/samples/unit-tests/series-bubble/extremes/demo.js
+++ b/samples/unit-tests/series-bubble/extremes/demo.js
@@ -91,36 +91,3 @@ QUnit.test('Setting X axis extremes on bubble (#5167)', function (assert) {
     assert.ok(chart.yAxis[0].min < 102.9, 'DE point is within range');
     assert.ok(chart.yAxis[0].max > 102.9, 'DE point is within range');
 });
-
-QUnit.test('Calculation of Z extremes when Z = 0 (#17280)', function (assert) {
-    const chart = Highcharts.chart('container', {
-        chart: {
-            type: 'bubble',
-            animation: false
-        },
-
-        series: [{
-            data: [
-                [2, 2, 0]
-            ]
-        }, {
-            data: [
-                [1, 1, 20]
-            ]
-        }]
-    });
-
-    assert.strictEqual(
-        chart.bubbleZExtremes.zMin,
-        0,
-        'The lower Z extreme equals 0'
-    );
-
-    chart.series[1].data[0].update({ z: -20 });
-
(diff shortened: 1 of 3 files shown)


pawellysy (approved):
Looks good! Thanks

raf18seb (changes requested):
Nice, the fix is nice and clean, but please add also:
- [x] regression test
- [ ] comment with a small description and issue number so it's easy to find info in the future if necessery

samples/unit-tests/series-bubble/extremes/demo.js
```diff
@@ -91,3 +91,36 @@ QUnit.test('Setting X axis extremes on bubble (#5167)', function (assert) {
     assert.ok(chart.yAxis[0].min < 102.9, 'DE point is within range');
     assert.ok(chart.yAxis[0].max > 102.9, 'DE point is within range');
 });
+
+QUnit.test('Calculation of Z extremes when Z = 0 (#17280)', function (assert) {
+    const chart = Highcharts.chart('container', {
```
raf18seb, line 96: Thanks for the test! Is there a chance to use an already existing test instead of creating a whole new separate test?
Word of explanation: a separate new test creates the whole chart from scratch. When having thousands of tests, we would have to create thousands of charts, and that would be a performance kill (running tests already takes at least a few minutes). Instead, you can use the existing test/chart and use the `update()` method (very often without `redraw` if not needed) - it's much more efficient for a longer period of time.

raf18seb (changes requested):
Thanks! One more comment, please see below

TorsteinHonsi (approved):
Thanks! Approved, waiting for the CI to approve.

PR #18606, opened 2023-03-02
db/board-api

karolkolodziej:
Added API documentation for the Board class.


ts/Dashboards/Board.ts
```diff
@@ -413,37 +503,105 @@ namespace Board {
...
         medium: number;
+        /**
+         * Value in px to test the dashboard is in large mode.
+         **/
         large: number;
     }
 
     export interface GUIOptions {
+        /**
+         * Whether the GUI is enabled or not.
```
karolkolodziej, line 584: Does it make sense to have that `enable` here? When can we use this property?

IMO it is redundant, what is your opinion?

ts/Dashboards/Board.ts
```diff
@@ -197,25 +281,12 @@ class Board implements Serializable<Board, Board.JSON> {
...
-     * The DOM element to render to, or its id.
-     *
-     * @param  {Board.Options} options
-     * The options for the dashboard.
+     * Creates a new layouts and adds it to the dashboard based on the options.
+     * @internal
      *
-     * @return {Board}
-     * The new dashboard.
+     * @param guiOptions THE GUI options for the layout.
```
bre1470, line 287: Add a line break.

ts/Dashboards/Board.ts
```diff
@@ -229,7 +300,13 @@ class Board implements Serializable<Board, Board.JSON> {
         }
     }
 
-    public setLayoutsFromJSON(json: Array<Layout.JSON>): void {
+    /**
+     * Set the layouts from JSON.
+     * @internal
+     *
+     * @param json An array of layout JSON objects.
```
bre1470, line 307: Add a line break.

ts/Dashboards/Board.ts
```diff
@@ -243,18 +320,48 @@ class Board implements Serializable<Board, Board.JSON> {
...
-    public addLayoutFromJSON(json: Layout.JSON): Layout|undefined {
-        return Layout.fromJSON(json, this);
-    }
-
-    public setComponents(
+    /**
+     * Set the components from options.
+     * @internal
+     *
+     * @param components An array of component options.
```
bre1470, line 327: Add a line break.

ts/Dashboards/Board.ts
```diff
@@ -292,31 +399,20 @@ class Board implements Serializable<Board, Board.JSON> {
         );
     }
 
+    /**
+     * Import the dashboard's layouts from the local storage.
+     *
+     * @param id The id of the layout to import.
```
bre1470, line 405: Add a line break.

ts/Dashboards/Board.ts
```diff
@@ -345,11 +441,9 @@ class Board implements Serializable<Board, Board.JSON> {
     /**
      * Converts the given JSON to a class instance.
      *
-     * @param {Serializable.JSON} json
-     * JSON to deserialize as a class instance or object.
+     * @param json JSON to deserialize as a class instance or object.
```
bre1470, line 444: Add a line break.

ts/Dashboards/Board.ts
bre1470: I agree. It seems to be redundant. Maybe it is supposed to control visibility?

bre1470 (changes requested):
Nice job! Some changes requested...

ts/Dashboards/Board.ts
bre1470: I agree. It seems to be redundant. Maybe it is supposed to control visibility?

bre1470 (approved):
Thank you! Looks good. 👍

ts/Dashboards/Board.ts
```diff
@@ -126,24 +134,104 @@ class Board implements Serializable<Board, Board.JSON> {
...
+    /**
+     * Flag to determine if the GUI is enabled.
+     * @internal
+     * */
     public guiEnabled: (boolean|undefined);
+
+    /**
+     * The unique id of the dashboard, it is generated automatically.
+     * */
     public id: string;
```
pawellysy, line 203:
```suggestion
    public readonly id: string;
```

ts/Dashboards/Board.ts
```diff
@@ -126,24 +134,104 @@ class Board implements Serializable<Board, Board.JSON> {
...
+
+    /**
+     * Index of the board in the global boards array. Allows to access the
+     * specific one when having multiple dashboards.
+     * */
     public index: number;
```
pawellysy, line 209:
```suggestion
    public readonly index: number;
```

pawellysy (changes requested):
Great job! Small comments tho

sebastianbochan (approved):

sebastianbochan:
Looks good 👍

pawellysy (approved):
LGTM

ts/Dashboards/Board.ts
```diff
@@ -413,37 +522,107 @@ namespace Board {
...
+     **/
     export interface Options {
+        /**
+         * Options for the GUI. Allows to define graphical elements and its
+         * layout.
+         **/
         gui?: GUIOptions;
+        /**
+         * Options for the edit mode. Can be used to enable the edit mode and
+         * define al things related to it like the context menu.
```
TorsteinHonsi, line 536: Typo: "al things"

ts/Dashboards/Board.ts
```diff
@@ -413,37 +522,107 @@ namespace Board {
...
+         * General options for the components.
+         **/
         componentOptions?: Partial<Bindings.ComponentOptions>;
+        /**
+         * A list of serialized layouts to add to the board.
+         **/
         layoutsJSON?: Array<Layout.JSON>;
-        respoBreakpoints?: RespoBreakpoints;
+        /**
+         * Responsive breakpoints for the board- small, medium and large.
```
TorsteinHonsi, line 552: Space before the dash

ts/Dashboards/Board.ts
TorsteinHonsi: If we keep it, it should be true by default.

TorsteinHonsi (changes requested):
Thanks! Just some typos.

ts/Dashboards/Board.ts
```diff
@@ -52,10 +52,8 @@ const {
  * */
 
 /**
- * Creates a dashboard with components like charts, tables, and HTML.
+ * Class that represents a dashboard.
```
goransle, line 55: Should this have an `@example` (or similar) like the [`Chart` class](https://github.com/highcharts/highcharts/blob/7ca7702a41215c6c9696f7c555df6cef1b4f384b/ts/Core/Chart/Chart.ts#L189-L199) has in Highcharts?

pawelfus (changes requested):
Looks good! A few questions though:

1. Do we need properties like `editMode` and `fullscreen`? I am also curious about `fromJSON` method

image

2.  Shouldn't `layoutsJSON` be inside `gui` options? 

3. layout options are not described:
image

4. What is the difference between `options` and `optionsJSON`? 🤔

karolkolodziej:
1. IIMO users can make use of the `editMode` and `fullscreen` instances thus I would leave it as it is.
3. I will add that description.

When it comes to number 2/4 and generally JSON-related stuff maybe we should discuss it internally because, to be honest, I don't fully understand the whole concept/flow between. What do you think @pawelfus?

pawelfus:
> 1. IIMO users can make use of the `editMode` and `fullscreen` instances thus I would leave it as it is.

That means we need to describe the EditMode class. Isn't that similar to `RangeSelector`, where the class is not exposed, yet available (unofficially)? 

> 2. I will add that description.

I assume it's the answer to 3)? Thanks for the better description! However, I wanted to say that I don't know how to configure layouts. It says options, but without proper doclets:

Zrzut ekranu 2023-03-15 o 13 58 25


> When it comes to number 2/4 and generally JSON-related stuff maybe we should discuss it internally because, to be honest, I don't fully understand the whole concept/flow between. What do you think @pawelfus?

Could you organize it? I believe @sebastianbochan should join too. To prevent blocking this PR, could you please remove it?

karolkolodziej:
1. As I think about it, it looks like the case with the RangeSelector. Where users can interact with it but the instance is not exposed. I marked them as `internal` for now! 

I also marked each JSON-related thing, as `internal` for now. I will organize a meeting where we can discuss how to approach it.

ts/Dashboards/Layout/Cell.ts
```diff
@@ -501,27 +505,62 @@ class Cell extends GUIElement {
...
+     * Options for each cell.
+     **/
     export interface Options {
+        /**
+         * Unique cell id.
+         **/
         id: string;
-        width?: string; // eg 50%, 1/2
+        /**
+         * With of the cell. Can be a percentage value or a fraction.
```
pawelfus, line 532: Typo: `width`

pawelfus (changes requested):
Thanks!
- Please resolve conflicts
- About layouts:
image

Is that correct? `Layout` has the option to set a `cellClassName` and `rowClassName`? But Cell doesn't have a such option? If it's `layout.cellClassName` which cell will get that classname? All cells within a layout? Or it's something different? 🤔 If it's not a bug in docs, then a bit more wide description would be nice

image

ts/Dashboards/Layout/Cell.ts
pawelfus: I would add examples here, in the description. Simple `For example '50%' or '1/3'` would be enough, what do you think?

karolkolodziej:
In the `layout` demo we have that
https://github.com/highcharts/highcharts/blob/daa4c24380ae5ee65bb057da2f7e6e2058aa890e/samples/dashboards/demos/layout/demo.js#L82-L83

Two conclusions here:
1. The `cellClassName` is applied to each cell that is that particular layout
2. The  `rowClassName` is applied to each row of that layout but also to the layout container

Is that correct behaviour @sebastianbochan?

karolkolodziej:
@pawelfus maybe to don't block this issue we could leave it as it is and discuss the `cellClassName` and `rowClassName` in a separate issue?

pawelfus:
I'm okay with these options here (quite convenient to be honest). Just improve the description, for example this one:
> Two conclusions here:
> 
> The cellClassName is applied to each cell that is that particular layout
> The rowClassName is applied to each row of that layout but also to the layout container

Is much more detailed, than simple:

> The class name of the XYZ container.

pawelfus (approved):
Thanks!

PR #18607, opened 2023-03-02
bugfix/17257-item-point-color-update

### LINKED ISSUE #17257: Series Color is not updating in item series charts
In item series chart, on changing a data point's color, only the legend's symbol and connector's colors have changed.

JS Fiddle Reference :  https://jsfiddle.net/zmcr48q7/1/

Vazonik:
Fixed #17257, item chart points color wasn't updating.

### DIFF
--- a/ts/Series/Item/ItemSeries.ts
+++ b/ts/Series/Item/ItemSeries.ts
@@ -312,7 +312,7 @@ class ItemSeries extends PieSeries {
                         y: y,
                         width: width,
                         height: height,
-                        fill: point.color
+                        fill: point.color // Circles color update (#17257)
                     };
                     if (typeof r !== 'undefined') {
                         attr.r = r;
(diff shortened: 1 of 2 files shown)


ts/Series/Item/ItemSeries.ts
```diff
@@ -311,7 +311,8 @@ class ItemSeries extends PieSeries {
                         x: x,
                         y: y,
                         width: width,
-                        height: height
+                        height: height,
+                        fill: point.color // Circles color update (#17257)
```
TorsteinHonsi, line 315: This works for this case, but doesn't work for changes in the stroke color, stroke width etc. In short, those properties returned by `pointAttribs`. These attributes are only applied to new graphics:

https://github.com/highcharts/highcharts/blob/a905c72df5599de95cf9a8fd040d59338dde530c/ts/Series/Item/ItemSeries.ts#L326-L328

So we need to change that so they are applied to existing graphics as well.

TorsteinHonsi (changes requested):

TorsteinHonsi (approved):
Thanks!

PR #18609, opened 2023-03-02
bugfix/18590-topojson-from-highcharts-maps-array

### LINKED ISSUE #18590: TopoJSON maps not taken from `Highcharts.maps` when `chart.map` set to a `string`
#### Expected behaviour
If `Highcharts.maps['myMap']` is set to a TopoJSON map object, it should work when `chart.map` is set to `'myMap'`.

#### Actual behaviour
Even if we add a TopoJSON map object to the `Highcharts.maps` array it is not detected in `chart.map` by string. 

Doesn't work: `chart.map: 'myMap'`
Does work: `chart.map: Highcharts.maps['myMap']`

Even though the API (https://api.highcharts.com/highmaps/chart.map) states: "If set to a string, it functions as an index into the Highcharts.maps array."

See that the map works with GeoJSON: https://jsfiddle.net/BlackLabel/3vzg9wmd/
See that it doesn't work with TopoJSON: https://jsfiddle.net/BlackLabel/q9kuta2v/

#### Product version
Highcharts 10.3

jakubSzuminski:
Fixed #18590, TopoJSON maps didn't work correctly when strings were passed to `chart.map`. 

The problem was that in the `getGeoMap` function we assumed that when there was a string passed to `chart.map` that `Highcharts.maps[map]` was a GeoJSON. Adding a `topo2geo` conversion for `TopoJSON` maps solves this problem.


ts/Maps/MapView.ts
```diff
@@ -415,7 +415,13 @@ class MapView {
 
     public getGeoMap(map?: MapDataType): GeoJSON|undefined {
         if (isString(map)) {
-            return maps[map];
+            const mapFromArray = maps[map];
+
+            if (mapFromArray && mapFromArray.type === 'Topology') {
+                return topo2geo(mapFromArray);
+            }
```
hubertkozik, line 422: Do we need to create an extra variable for `maps[map]`? In my opinion, we don't need that declaration.

hubertkozik (changes requested):
Good job with finding a cause! We need also a test for this issue, can you add it?

TorsteinHonsi:
Good job! I second Hubert's requests.

hubertkozik (approved):
LGTM!

PR #18612, opened 2023-03-02
bugfix/18303-point-visibility-drillup

### LINKED ISSUE #18303: Default invisible point in the ignoreHiddenPoint: false series bugs out during drillup pie chart
#### Expected behaviour
When drilling up to the main state of the chart, which has `ignoreHiddenPoint` disabled and one of its points is with the parameter `visible: false` set, **the point is not visible, as when the chart is first rendered**.

#### Actual behaviour
When drilling down to the main state of the chart, which has `ignoreHiddenPoint` disabled and one of its points is with the parameter `visible: false` set, **the point is visible even though it is marked as disabled on the legend**.

1. Drilldown to a point called _Drilldown Here_
2. Drillup to main

https://jsfiddle.net/BlackLabel/43mtrxqL/

#### Product version
Highcharts v10.3.2

Vazonik:
Fixed #18303, a point with `visible: false` was shown after drill-up.

### DIFF
--- a/ts/Extensions/Drilldown.ts
+++ b/ts/Extensions/Drilldown.ts
@@ -1533,7 +1533,10 @@ ColumnSeries.prototype.animateDrillupTo = function (init?: boolean): void {
                         dataLabel = point.dataLabel;
 
 
-                    if (point.graphic && point.visible) { // #3407
+                    if (
+                        point.graphic && // #3407
+                        point.visible // Don't show if invisible (#18303)
+                    ) {
                         point.graphic[verb](inherit);
                     }
 
--- a/ts/Extensions/Drilldown.ts
+++ b/ts/Extensions/Drilldown.ts
@@ -1533,7 +1533,7 @@ ColumnSeries.prototype.animateDrillupTo = function (init?: boolean): void {
                         dataLabel = point.dataLabel;
 
 
-                    if (point.graphic) { // #3407
+                    if (point.graphic && point.visible) { // #3407
                         point.graphic[verb](inherit);
                     }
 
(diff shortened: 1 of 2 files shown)


ts/Extensions/Drilldown.ts
```diff
@@ -1533,7 +1533,10 @@ ColumnSeries.prototype.animateDrillupTo = function (init?: boolean): void {
                         dataLabel = point.dataLabel;
 
 
-                    if (point.graphic) { // #3407
+                    if (
+                        point.graphic && // #3407
+                        point.visible // Don't show if invisible (#18303)
```
TorsteinHonsi, line 1538: I didn't test this, but I think `point.visible` can be `undefined`, in which case the point is actually visible. In other parts of the code base we test for an explicit `false`:

https://github.com/highcharts/highcharts/blob/1f4eea0567fa6ba26fde40694a22441b8d52f822/ts/Core/Series/DataLabel.ts#L234

Please check if this may be the case here too.

TorsteinHonsi (changes requested):

ts/Extensions/Drilldown.ts
Vazonik: Thanks! From what I can see, `visible: undefined` for line and scatter chart's points works as you described, but for column and pie, they are still invisible. Moreover, in the case of the pie chart, the legend item is locked then.

Demo: https://jsfiddle.net/BlackLabel/w8zcvgbt/

TorsteinHonsi: Yes, but there's a difference between the options and the `Point` properties. Anyway, it seems that it is legit to use `point.visible` directly. See https://jsfiddle.net/highcharts/r9m56cgb/

TorsteinHonsi (approved):
Looks good 👍  Awaiting CI tests.

PR #18615, opened 2023-03-03
db/doclets-datagrid

bre1470:


pawellysy (approved):
LGTM!

tools/gulptasks/dashboards/api-docs/typedoc/assets/custom.css
```diff
@@ -0,0 +1,3 @@
+.tsd-navigation.primary .tsd-kind-namespace.tsd-parent-kind-module {
```
karolkolodziej, line 1: 🙏 thanks for that!

ts/DataGrid/DataGrid.ts
```diff
@@ -49,6 +54,9 @@ const {
  *
  * */
 
+/**
+ * Creates a grid structure with scrollable data cells.
```
karolkolodziej, line 58: ... scrollable and editable data cells?

ts/DataGrid/DataGrid.ts
```diff
@@ -57,45 +65,86 @@ class DataGrid {
...
      *
      *  Functions
      *
      * */
 
+    /**
+     * Creates an instance of DataGrid.
+     *
+     * @param container
+     * Container to create the grid structure into.
```
karolkolodziej, line 143: or its `id`

ts/DataGrid/DataGrid.ts
```diff
@@ -279,9 +340,7 @@ class DataGrid {
     /**
      * Get a reference to the underlying DataTable from options, or create one
      * if needed.
-     *
-     * @return {Highcharts.DataTable}
-     * Table reference
+     * @internal
```
karolkolodziej, line 343: Why removing the return description?

ts/DataGrid/DataGrid.ts
```diff
@@ -130,6 +179,7 @@ class DataGrid {
     }
 
 
+    /** @internal */
```
karolkolodziej, line 182: Please add a description and check other places where lint complains about missing one.

ts/DataGrid/DataGridOptions.d.ts
```diff
@@ -21,21 +27,82 @@ import type DataTable from '../Data/DataTable';
...
  * */
 
+/**
+ * Options to control the content and the user expierence of a grid structure.
+ */
 export interface DataGridOptions {
+
+    /**
+     * Height of each grid cell. This is used to calculate the amount of visible
+     * cells in a container and the size of the scrollbar.
```
karolkolodziej, line 37: I wouldn't say it controls the size of the scrollbar.
IMO it should be:
`cells in a container and the size of the scrollable container.`

ts/DataGrid/DataGridOptions.d.ts
```diff
@@ -21,21 +27,82 @@ import type DataTable from '../Data/DataTable';
...
+     *
+     * @default 25
+     */
     cellHeight: number;
```
karolkolodziej, line 41: Can we make it optional?

ts/DataGrid/DataGridOptions.d.ts
```diff
@@ -21,21 +27,82 @@ import type DataTable from '../Data/DataTable';
...
+    /**
+     * Switch to make the column sizes editable ('true') or fixed ('false').
+     *
+     * @default true
+     */
     resizableColumns: boolean;
 }
 
+/**
+ * Contains options for column headers.
```
karolkolodziej, line 76: Do we have to repeat the description here? What if they mismatch?

ts/Extensions/DashboardPlugins/DataGridComponent.ts
```diff
@@ -153,6 +178,7 @@ class DataGridComponent extends Component<DataGridComponent.ChartComponentEvents
      *
      * */
 
+    /** @internal */
```
karolkolodziej, line 181: Can you please add a block descriptions everywhere or change the rules for that folder?

ts/Extensions/DashboardPlugins/DataGridComponent.ts
```diff
@@ -342,38 +377,89 @@ class DataGridComponent extends Component<DataGridComponent.ChartComponentEvents
...
 
+    /**
+     * Options to control the DataGrid component. These options can be changed
+     * in edit mode of a dashboard.
+     */
     export interface EditableOptions extends Component.EditableOptions {
-        dataGridOptions?: DataGridOptions;
+
+        /**
+         * Generic options to adjust behavor and styling of the rendered data
```
karolkolodziej, line 431: I think @sebastianbochan already started working on describing `DataGridComponent`.
Please check with him in order to avoid conflicts.

karolkolodziej (changes requested):

ts/DataGrid/DataGrid.ts
bre1470: It is obvious from the description and the type can be reflected.

ts/DataGrid/DataGridOptions.d.ts
bre1470: I find your suggestion too technical. In the implementation it might be also related to the container, but this is only done to have a proper size of the scrollbar toggle.

ts/DataGrid/DataGridOptions.d.ts
bre1470: What do you have in mind?

karolkolodziej: In lines 43-46, you have a description for the `  columnHeaders?: ColumnHeaderOptions;` and also the same description is in lines 75-77. I was wondering which is truly picked up by the typedoc for documentation.

But as far as I can see we might have a different description in both places.
You can ignore this request ;)

ts/DataGrid/DataGridOptions.d.ts
```diff
@@ -21,21 +27,82 @@ import type DataTable from '../Data/DataTable';
  *
  * */
 
+/**
+ * Options to control the content and the user expierence of a grid structure.
```
karolkolodziej, line 31: `expierence` typo

karolkolodziej (approved):

ts/DataGrid/DataGridOptions.d.ts
bre1470: What typo do you see? 🤔

karolkolodziej: `expierence` should be `experience`

bre1470: Right, thanks. 👍

TorsteinHonsi (approved):

sebastianbochan:
Right now we have doclets for the DataGridComponent, so the question is, should we overwrite it by this PR or not?

@pawelfus @bre1470

bre1470:
1. Maybe the doclet over the namespace shows up at that place.
2. Yes, I do not know why we have two interfaces for options.

pawelfus (approved):
Thanks!

PR #18621, opened 2023-03-06
db/component-directory-rename

karolkolodziej:
Renamed the components directory and the classes to match the `Dashboards` name.


bre1470 (approved):
Good catch with `$class` strings. Looks good. 👍

pawellysy (approved):

PR #18614, opened 2023-03-03
Migration: Pointer and MSPointer

marvin19:
Added compose to Pointer and MSPointer.

PR #18624, opened 2023-03-06
ES5 master with MSPointer.

bre1470:
Part of #18070, #18614

To-do:
- [x] Highcharts Assembler ignores eventually base option and takes regular `master`
- [x] `lint-ts` task is picky with `MSPointer`, because excluded in `tsconfig.json`


ts/tsconfig.json
```diff
@@ -28,6 +28,10 @@
             "trusted-types"
         ]
     },
+    "exclude": [
+        "./Core/MSPointer.ts",
```
bre1470, line 32: @TorsteinHonsi Should we keep MSPointer in `code/es-modules`?

TorsteinHonsi: Yes, I guess we still need the file, just not include it in the default build?

bre1470:
Thanks, @ppotaczek! Looks good. 👍

TorsteinHonsi:
@bre1470 This branch sees to be based on the master, but for the new es5/es6 stuff we need to work on the `v11` branch. I tried to set a new base, but that doesn't seem to be the intension. Please advice. Should we merge the master into `v11` to get the latest Pointer compositions, then merge this one into `v11` after that?

TorsteinHonsi (approved):
Approved, now awaiting CI

PR #18625, opened 2023-03-07
enhancement/popup-refactor

karolkolodziej:
Refactored Popup.

- [ ] Add `BasePopup`
- [ ] Refactor `Popup`
- [ ] Refactor `PopupAnnotation`
- [ ] Refactor `PopupComposition`
- [ ] Refactor `PopupIndicators`
- [ ] Refactor `PopupTabs`

### DIFF
--- a/ts/Extensions/Annotations/Popup/BasePopup.ts
+++ b/ts/Extensions/Annotations/Popup/BasePopup.ts
@@ -1,0 +1,222 @@
+/* *
+ *
+ *  Base popup.
+ *
+ *  (c) 2009-2021 Karol Kolodziej
+ *
+ *  License: www.highcharts.com/license
+ *
+ *  !!!!!!! SOURCE GETS TRANSPILED BY TYPESCRIPT. EDIT TS FILE ONLY. !!!!!!!
+ *
+ * */
+
+'use strict';
+
+/* *
+ *
+ *  Imports
+ *
+ * */
+
+import type Chart from '../../../Core/Chart/Chart';
+
+import AST from '../../../Core/Renderer/HTML/AST.js';
+import D from '../../../Core/Defaults.js';
+const { getOptions } = D;
+import H from '../../../Core/Globals.js';
+const { doc } = H;
+import U from '../../../Core/Utilities.js';
+const {
+    addEvent,
+    createElement,
+    fireEvent
+} = U;
+
+/* *
+ *
+ *  Declarations
+ *
+ * */
+
+
+export interface PopupFieldsObject {
+    actionType: string;
+    fields: PopupFieldsTree;
+    linkedTo?: string;
+    seriesId?: string;
+    type?: string;
+}
+
+export interface PopupFieldsTree {
+    [key: string]: (string | PopupFieldsTree);
+}
+
+/* *
+ *
+ *  Class
+ *
+ * */
+
+class BasePopup {
+
+    /* *
+     *
+     *  Constructor
+     *
+     * */
+
+    public constructor(
+        parentDiv: HTMLElement,
+        iconsURL: string,
+        chart?: Chart
+    ) {
+        this.chart = chart;
+        this.iconsURL = iconsURL;
+        this.lang = (getOptions().lang.navigation as any).popup;
+
+        // Create popup div.
+        this.container = createElement(
+            'div',
+            {
+                className: 'highcharts-popup highcharts-no-tooltip'
+            },
+            void 0,
+            parentDiv
+        );
+
+        addEvent(this.container, 'mousedown', (): void => {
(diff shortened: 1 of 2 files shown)


css/annotations/popup.scss
```diff
@@ -38,6 +38,7 @@ $highlight-color-80: #335cad;
   color: $neutral-color-60;
   display: none;
   font-size: 0.876em;
+  min-height: 40px;
```
pawellysy, line 41: Is that necessary? Maybe set the height via flexbox?

pawellysy (changes requested):
Great job! Few comments tho

ts/Extensions/BasePopup.ts
```diff
@@ -0,0 +1,189 @@
...
+        const closeButton = createElement(
+            'div',
+            {
+                className: 'highcharts-popup-close'
+            },
+            void 0,
+            this.container
+        );
+
+        closeButton.style['background-image' as any] = 'url(' +
```
bre1470, line 114: Please solve this `any`.

bre1470 (changes requested):

bre1470:
Maybe it is time to have a `ts/Shared` folder with shared code between Highcharts and Dashboards? - like `ts/Shared/Data` and `ts/Shared/UI`.

pawellysy (approved):
LGTM!

ts/Extensions/BasePopup.ts
```diff
@@ -0,0 +1,189 @@
...
+    createElement
+} = U;
+
+/* *
+ *
+ *  Class
+ *
+ * */
+
+abstract class BasePopup {
```
pawelfus, line 34: I remember we started the discussion about the name. How about `Form`/`BaseForm`? Then changing `formType` property to a simple `type`

Bringing @bre1470 and @sebastianbochan to the discussion :)

ts/Extensions/BasePopup.ts
```diff
@@ -0,0 +1,189 @@
...
+     */
+    private addCloseButton(): HTMLElement {
+        const popup = this,
+            iconsURL = this.iconsURL;
+
+        // Create close popup button.
+        const closeButton = createElement(
+            'div',
+            {
+                className: 'highcharts-popup-close'
```
pawelfus, line 108: It would be nice to set `className` as an argument - similar to `createPopupContainer`

pawelfus (changes requested):
Some minor changes

ts/Extensions/BasePopup.ts
karolkolodziej: Changed to `BaseForm` and `type`

pawellysy: Sounds good!

bre1470 (approved):
Thank you! Looks good. 👍

ts/Extensions/BasePopup.ts
sebastianbochan: Unfortunately I cannot agree, the `BaseForm` means nothing for me. 

For me the `MockPopup` or `Popup` as a base sounds more intuitive. Then, we can extend the class by a specified form of popup like `AnnotationPopup`.

TL:LR;
I would craete an abstract class as `Popup` and extend it by a specific one like `AnnotationPopup` / `ConfirmationPopup` /  `DashboardPopup` / `DataPopup` etc.

ts/Shared/BaseForm.ts
```diff
@@ -0,0 +1,196 @@
+/* *
+ *
+ *  Base popup.
```
sebastianbochan, line 3: The name should be as a class name.

sebastianbochan (changes requested):

ts/Extensions/BasePopup.ts
karolkolodziej: For now, we decided only to extract that abstract class. 
Creating separate classes for each file will not be part of the PR.

Regarding naming, I think `MockPopup` sounds like it is something for temporary use.
`BaseForm`/`BasePopup` works for me :)

pawelfus: I agree with Karol, and I'm okay with both: `BasePopup` and `BaseForm` 👍

pawelfus (approved):
Apart from an ongoing name discussion - looks good 👍

sebastianbochan (approved):

ts/Extensions/BasePopup.ts
pawelfus: Seeing all green marks.. I guess the decision is up to @karolkolodziej ;)

karolkolodziej:
@TorsteinHonsi we are good to go 🚀

PR #18628, opened 2023-03-07
Migration: PivotPointsIndicator

marvin19:
Improved PivotPointsIndicator.


bre1470 (changes requested):
Looks good.

There are some more optimizations possible. The prototype calls can be replaced with shorter super calls as the class inherits from SMA. This is also possible in the point class, when you move the external function as a private function inside the class.

bre1470 (approved):
Thank you! Looks `super`. 👍

PR #18632, opened 2023-03-07
db/doclets-component-class

sebastianbochan:


karolkolodziej (approved):

ts/Dashboards/Components/Component.ts
```diff
@@ -101,57 +148,138 @@ abstract class Component<TEventObject extends Component.EventTypes = Component.E
...
+    /**
+     * The options for the component.
+     * */
     public options: Component.ComponentOptions;
+    /**
+     * The type of component like: `HTML`, `KPI`, `Highcharts`, `DataGrid`.
+     */
     public type: string;
+    /**
+     * Sets an ID for the component's `div`.
```
bre1470, line 193: This needs a bit more clarifiaction as a component can have multiple `div`.

ts/Dashboards/Components/HTMLComponent.ts
```diff
@@ -147,9 +188,10 @@ class HTMLComponent extends Component<HTMLComponent.HTMLComponentEvents> {
 
     /* *
      *
-     *  Class methods
+     *  Functions
```
bre1470, line 191: Nice catch.

bre1470 (changes requested):

TorsteinHonsi (approved):

ts/Dashboards/Components/Component.ts
sebastianbochan: The component's id is for the main component div that is unique.

ts/Dashboards/Components/KPIComponent.ts
```diff
@@ -36,9 +44,70 @@ const {
...
+        );
+    }
+
+    /* *
+     *
+     *  Static properties
+     *
+     * */
+
+    public static charter?: typeof G;
```
pawelfus, line 107: Do we need this?

image

ts/Dashboards/Components/Component.ts
```diff
@@ -639,10 +828,10 @@ abstract class Component<TEventObject extends Component.EventTypes = Component.E
     }
 
     /**
-     * Handles setting things up on initial render
+     * Handles setting things up on initial render.
      *
-     * @return {this}
-     * The component for chaining
+     * @returns
+     * The component for chaining.
```
pawelfus, line 834: Do we need this method in API?

ts/Extensions/DashboardPlugins/HighchartsComponent.ts
```diff
@@ -82,6 +81,9 @@ class HighchartsComponent extends Component<HighchartsComponent.ChartComponentEv
 
     public static syncHandlers = HighchartsSyncHandlers;
```
pawelfus, line 82: Do we need this property in API? And some others?

image

ts/Dashboards/Components/KPIComponent.ts
pawelfus, line 1: Generally about API for KPI:

image

Do we need all of these to be public? For example:
- sync
- valueWrap
- charter
- load method
- fromJSON method

ts/Dashboards/Components/KPIComponent.ts
```diff
@@ -89,18 +167,62 @@ class KPIComponent extends Component {
...
+    /**
+     * HTML element where the value is created.
+     */
     public value: HTMLElement;
+    /**
+     * The HTML element where the subtitle is created.
+     */
     public subtitle: HTMLElement;
+    /**
+     * HTML element where the chart is created.
```
pawelfus, line 193: Shouldn't `chartContainer` be optional since `chart` is optional?

ts/Dashboards/Components/HTMLComponent.ts
pawelfus, line 1: Do we need all of these to be public?

image

For example methods:
- autoScale
- load
- render & redraw (wouldn't just one of them be enough?)
- scaleText

And property:
- sync

ts/Dashboards/Components/Component.ts
pawelfus, line 1: Do we need to define all of this?

image


For example:
- Variables (?)
- relayMessage
- getName

sebastianbochan:
> See comments below
> 
> Additionally, `SyncOptions` seems to be a bit undocumented, see: Zrzut ekranu 2023-03-14 o 13 48 42

This is part of ComponentOptions and will be fixed in another PR.

ts/Dashboards/Components/Component.ts
sebastianbochan: Fixed.

pawelfus (approved):
Thanks!

PR #18633, opened 2023-03-07
bugfix/17142-venn-sets-comma-issue

### LINKED ISSUE #17142: Venn: Uncaught TypeError Cannot read properties of undefined
#### Expected behaviour

Simple Venn diagram with no overlaps should show as discrete circles. 

Live working demo: https://jsfiddle.net/abhitalks/tf9z24ur/

#### Actual behaviour

Simple Venn diagram exactly the same as above, but with three data points throws a console error:

```
venn.js:24 Uncaught TypeError: Cannot read properties of undefined (reading '<_insert first string in set_>')
    at b.reduce.loss (venn.js:24:85)
    at Array.reduce (<anonymous>)
    at venn.js:24:36
    at Object.layoutGreedyVenn (venn.js:23:502)
    at Function.f.layout (venn.js:31:347)
    at f.translate (venn.js:35:81)
    at highcharts.js:393:509
    at a.renderSeries (highcharts.js:393:487)
```

template used [^1]

Demo to repro with console error: https://jsfiddle.net/abhitalks/sbyc0auf/

#### Product version

Latest from CDN: `https://code.highcharts.com/...`

#### Affected browser(s)
Chrome Version 99.0.4844.84, 
Edge Version 99.0.1150.55.

On Firefox v92.0.1 the console error is: `Uncaught TypeError: h is undefined H venn.src.js:1115`

Other browsers, not tested.

[^1]: (based on official Venn template here: https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/demo/venn-diagram)

jakubSzuminski:
Fixed #17142, Venn diagram did not work when there was a comma in the string in the `series.data.sets` array.

**Example** of the problem: 
`['A,']` and `['B']` = `'A,,B'` `(combination)` which caused issues when `combination.split(',')`

**Solution:**
I join the elements of the array with a string `SPLIT//COMBINATIONS` which is unlikely to occur in user's code, but maybe we would want to join the sets with another unique string.

### DIFF
--- a/ts/Series/Venn/VennUtils.ts
+++ b/ts/Series/Venn/VennUtils.ts
@@ -815,8 +815,11 @@ function processVennData(
                 return validSets.indexOf(set) === -1;
             })
         ) {
-            mapOfIdToRelation[(relation.sets as any).sort().join()] =
-                relation as any;
+            mapOfIdToRelation[
+                (relation.sets as any)
+                    .sort()
+                    .join('SPLIT//COMBINATIONS')
+            ] = relation as any;
         }
         return mapOfIdToRelation;
     }, {});
@@ -830,13 +833,13 @@ function processVennData(
         const remaining = arr.slice(i + 1);
 
         remaining.forEach(function (set2: string): void {
-            combinations.push(set + ',' + set2);
+            combinations.push(set + 'SPLIT//COMBINATIONS' + set2);
         });
         return combinations;
     }, []).forEach(function (combination: string): void {
         if (!mapOfIdToRelation[combination]) {
             const obj: Highcharts.VennRelationObject = {
-                sets: combination.split(','),
+                sets: combination.split('SPLIT//COMBINATIONS'),
                 value: 0
             } as any;
 
--- a/ts/Series/Venn/VennUtils.ts
+++ b/ts/Series/Venn/VennUtils.ts
@@ -91,26 +91,31 @@ function addOverlapToSets(
     relations: Array<Highcharts.VennRelationObject>
 ): Array<Highcharts.VennRelationObject> {
     // Calculate the amount of overlap per set.
-    const mapOfIdToProps = relations
+    const mapOfIdToProps: Record<string, Highcharts.VennPropsObject> = {};
+
+    relations
         // Filter out relations consisting of 2 sets.
         .filter((relation): boolean => (relation.sets.length === 2))
         // Sum up the amount of overlap for each set.
-        .reduce(
-            (map, relation): Record<string, Highcharts.VennPropsObject> => {
-                relation.sets.forEach((set, i, arr): void => {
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (changes requested):
Thanks! Please add a test too. Preferably modify an existing venn test so that we don't add to the test processing time.

Besides, there are some `any` casts in the code that you modified - since you're working on this code now and have an understanding, maybe you can take the opportunity to fix the typing?

Not so important, and don't spend too much time on this: Why are we using a string that we concatenate and split, instead of an array? I understand it if it's used for indexing an object, otherwise an array would probably be cleaner.

pawellysy (approved):
Thanks!

jakubSzuminski:
1. I've added a test case for `,` in strings

2. When it comes to the types, I recognized that there are a few stages there:
- 1 - `relations` are created in `VennUtils.processVennData()` which is an array of objects with only `sets` and `value` properties (even though the `VennRelationObject` also had `circle, overlapping, totalOverlap` as required properties)
- 2 - after `VennSeries.layout()` the `circle` is added to each of the `relations` so I decided to make `circle` optional (because it doesn't appear in the first step, the `relations` are "extended")
- 3 - then, the `overlapping` and `totalOverlap` are added in `addOverlapToSets()` so I also decided to make them optional (as they do not appear in the first and second step)
Then, I fixed `any` casting in a few places by adding some `if` conditions.

3. In this case strings (which come from `Array.join(splitter)`) are used for object indexing so I did not change the method for now (but of course I'll do if you decide that we should). 
Also, I've added a `splitter` as a `public static` variable of `VennSeries` so that the user might easily override the `splitter` string to something else if he happens to use the `highcharts-split` in his `series.venn.sets`

ts/Series/Venn/VennUtils.ts
```diff
@@ -91,26 +91,31 @@ function addOverlapToSets(
...
+                    };
+                }
+
+                mapOfIdToProps[set] = {
+                    totalOverlap: (mapOfIdToProps[set].totalOverlap || 0) +
+                        relation.value,
+                    overlapping: {
+                        ...(mapOfIdToProps[set].overlapping || {}),
+                        [arr[1 - i]]: relation.value
                     }
```
jakubSzuminski, line 115: I ran into a strange TypeScript issue here. I've tried to check if `typeof mapOfIdToProps[set].totalOverlap !== 'undefined'` and then add `relation.value` to it, but it did not seem to work as the TypeScript always threw 2532 error - `Object is possibly 'undefined'` (no matter how many conditions did I check that it is not `undefined`).

TorsteinHonsi (approved):
Okay, thanks for the clean up! Let's keep i like this.

PR #18637, opened 2023-03-08
Added pushUnique function to Utilities.

bre1470:


ts/Core/Utilities.ts
```diff
@@ -462,6 +462,29 @@ function erase(arr: Array<unknown>, item: unknown): void {
...
+ */
+function pushUnique(
+    array: Array<unknown>,
+    item: unknown
+): boolean {
+    if (array.indexOf(item) === -1) {
+        array.push(item);
+        return true;
+    }
+    return false;
```
TorsteinHonsi, line 485: Maybe we can save a few bytes by typing the function `true|undefined` and skip this line?

TorsteinHonsi (approved):
Thanks!

PR #18638, opened 2023-03-08
 bugfix/17168-inverted-sunburst

### LINKED ISSUE #17168: Sunburst chart wrong position in inverted chart
#### Expected behaviour
The sunburst series should be positioned in the center even if the chart is inverted

#### Actual behaviour
The series is positioned so that some of the circle is invisible

https://jsfiddle.net/BlackLabel/9b1p5vkn/155/

#### Product version
v10.0.0

#### Affected browser(s)
all

hubertkozik:
Fixed #17168, the Sunburst series didn’t work properly when the chart was inverted.

### DIFF
--- a/samples/unit-tests/series-sunburst/color/demo.js
+++ b/samples/unit-tests/series-sunburst/color/demo.js
@@ -1,5 +1,5 @@
 QUnit.test('series.color: default to series.colors[0].', function (assert) {
-    var H = Highcharts,
+    const H = Highcharts,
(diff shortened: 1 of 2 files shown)


hubertkozik:
The review tool has shown some changes in samples and file differences are showing changes from `clipPath` to `path`, in my opinion, they are intended changes, because sunburst shouldn't be clipped - just like pie.

TorsteinHonsi (approved):
Thanks!

PR #18639, opened 2023-03-09
bugfix/16940-multiple-axis-cross-label

### LINKED ISSUE #16940: Current price label (price-indicator) does not work with multiple axis
#### Expected behaviour

If you have say four axis, two on the right, two on the left, corresponding to 4 series, then the current price indicator would show correctly on the axis corresponding to that series.

#### Actual behaviour

The price indicators show on the same axis on the right, and same axis on the left. See this screenshot from our client:

Screen Shot 2022-01-31 at 11 13 15 AM

Can't create at this time

#### Product version

    "highcharts": "^9.1.1",
    "highcharts-react-official": "^3.0.0",

#### Affected browser(s)

All

jakubjanuchta:
Fixed #16940, wrong cross label positions on multiple axes.

I positioned the crossLabels on the corresponding axis.

### DIFF
--- a/samples/unit-tests/axis/crosshairs/demo.js
+++ b/samples/unit-tests/axis/crosshairs/demo.js
@@ -101,27 +101,25 @@ QUnit.test('Crosshair on multiple axes (#4927)', function (assert) {
         }, {
             ...yAxisCrosshair,
             opposite: true
-        }]
+        }],
+        series: [
+            {
+                data: sampleData
+            },
+            {
+                yAxis: 1,
+                data: sampleData
+            },
+            {
+                yAxis: 2,
+                data: sampleData
+            },
+            {
+                yAxis: 3,
+                data: sampleData
+            }
+        ]
     }, true, true);
-
-
-    chart.series[0].update({
-        data: sampleData
-    });
-
-    chart.series[1].update({
-        data: sampleData
-    });
-
-    chart.addSeries({
-        yAxis: 2,
-        data: sampleData
-    });
-
-    chart.addSeries({
-        yAxis: 3,
-        data: sampleData
-    });
 
     chart.series[0].points[0].onMouseOver();
 
--- a/samples/unit-tests/axis/crosshairs/demo.js
+++ b/samples/unit-tests/axis/crosshairs/demo.js
@@ -1,5 +1,14 @@
 QUnit.test('Crosshair on multiple axes (#4927)', function (assert) {
-    var chart = Highcharts.chart('container', {
+    const sampleData = [20, 12, 50, 40],
+        yAxisCrosshair = {
+            crosshair: {
+                label: {
+                    enabled: true
+                }
+            }
+        };
+
+    var chart = Highcharts.stockChart('container', {
         yAxis: [
             {
                 id: 'primary',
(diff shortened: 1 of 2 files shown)


samples/unit-tests/axis/crosshairs/demo.js
```diff
@@ -77,6 +86,63 @@ QUnit.test('Crosshair on multiple axes (#4927)', function (assert) {
...
+
+    chart.addSeries({
+        yAxis: 2,
+        data: sampleData
+    });
+
+    chart.addSeries({
+        yAxis: 3,
+        data: sampleData
+    });
```
pawelfus, line 124: Two things here:

1. Instead of multiple updates, prefer just one `chart.update()` with all options
2. Whenever you run multiple updates, prefer setting `redraw` argument to `false`. Just the last one should have `redraw=true` (or after the last one use `chart.redraw()`). It is faster this way (multiple redraws vs one)

ts/Core/Chart/StockChart.ts
```diff
@@ -760,7 +761,11 @@ addEvent(Axis, 'afterDrawCrosshair', function (
         posx = snap ? (point.plotX || 0) + left : e.chartX;
         posy = top + (opposite ? 0 : this.height);
     } else {
-        posx = opposite ? this.width + left : 0;
+        if (opposite) {
+            posx = left + width + this.offset;
+        } else {
+            posx = left + this.offset;
+        }
```
pawelfus, line 768: This could be a little bit simplified:
```ts
posx = left + this.offset + (opposite ? width : 0);
```

ts/Core/Chart/StockChart.ts
```diff
@@ -800,6 +805,14 @@ addEvent(Axis, 'afterDrawCrosshair', function (
     crossBox = crossLabel.getBBox();
 
     // now it is placed we can correct its position
+    if (isNumber(crossLabel.x)) {
+        if (!horiz) {
+            if (!opposite) {
```
pawelfus, line 810: There are not if-else cases, so you can put all conditions into one `if()`

samples/unit-tests/axis/crosshairs/demo.js
jakubjanuchta: Yes, thanks. It makes sense to put everything in one `chart.update()` and I will remember about redrawing the chart only once in the future, thanks!

ts/Core/Chart/StockChart.ts
jakubjanuchta: It looks much better when simplified, thanks!

ts/Core/Chart/StockChart.ts
```diff
@@ -800,6 +801,12 @@ addEvent(Axis, 'afterDrawCrosshair', function (
     crossBox = crossLabel.getBBox();
 
     // now it is placed we can correct its position
+    if (isNumber(crossLabel.x)) {
+        if (!horiz && !opposite) {
```
TorsteinHonsi, line 805: Combine the conditions

TorsteinHonsi (changes requested):
Thanks! Just a tiny optimization requested inline.

ts/Core/Chart/StockChart.ts
jakubjanuchta: Combined, thanks!

pawelfus (approved):
Thanks!

PR #18641, opened 2023-03-09
bugfix/18017-mousewheel-animation

### LINKED ISSUE #18017: Maps zoom displaces map points
#### Expected behaviour
Map points should not change their position (relative to the map) while quick scrolling. It should be simultaneous (or at least go back to the correct position after zoom). It's easier to reproduce it on the touchpad. 

1. Scroll quickly to see how points and labels can't keep up with the main map zoom level.
https://jsfiddle.net/BlackLabel/07f8g6pL/

image (3)
image (4)

#### Product version
Highmaps 10.3.1 (latest)

madepiet:
Fixed #18017, animation should not change the positions of map points while quick scrolling on the touchpad.


ts/Maps/MapPointer.ts
```diff
@@ -44,6 +44,7 @@ declare global {
             onContainerMouseWheel(e: PointerEvent): void;
         }
         interface MapPointerChart extends Chart {
+            isAnimating: boolean;
```
TorsteinHonsi, line 47: Please set this flag on the MapPointer interface, so we don't pollute the omnipresent Chart object.

ts/Maps/MapPointer.ts
```diff
@@ -160,6 +169,12 @@ extend<Pointer|Highcharts.MapPointer>(Pointer.prototype, {
                 Math.abs(delta) < 1 ? false : void 0
             );
         }
+
+        // Set a timeout to reset the animation flag after the animation
+        // finishes
+        setTimeout(function (): void {
+            chart.isAnimating = false;
+        }, 50);
```
TorsteinHonsi, line 177: When zooming with the trackpad, it is much better than it was, but it is still a jerky movement. It's not super smooth like for example Google Maps. I tried to lower the timeout to 25 milliseconds. Then the animation is much smoother, but if we zoom quickly (with the trackpad), it sometimes jumps around. Could you have a look at this?

TorsteinHonsi (changes requested):
Thanks, it's much better!

In addition to the inline comments, also make sure the PR description is in accordance with the guide in [Writing for the changelog](https://github.com/highcharts/highcharts/tree/master/changelog#writing-content-for-the-changelog). Currently it only describes some internal code that makes no sense to the readers of the changelog.

ts/Maps/MapPointer.ts
madepiet: Yes, this is the limitation with the `isAnimating` flag that I also came across and discussed in a private message. I'll try to experiment with a different approach, but here what worked was temporarily disabling or blocking the animation for these fast non-standard movements.

madepiet:
> @madepiet What is the status of this PR? Still draft or should we mark it Ready for review?

@TorsteinHonsi As we chat - we changed our approach by going a little deeper. The presented solution should help in synchronizing the animations.

TorsteinHonsi:
It looks promising in the current state. To do:
* Tests if feasible
* Clean up the code. The merge step is probably not necessary, I think we can work on `globalAnimation` directly. The `animator: 1` doesn't seem to be used.

TorsteinHonsi (changes requested):
To do:
* Create tests if possible. Not sure if it is easy to do though.
* Clean up the code. The merge step is probably not necessary, I think we can work on globalAnimation directly. The animator: 1 doesn't seem to be used.

ts/Series/Map/MapSeries.ts
```diff
@@ -792,7 +790,8 @@ class MapSeries extends ScatterSeries {
             // When dragging or first rendering, animation is off
             } else {
                 transformGroup.attr(merge(
-                    svgTransform, { 'stroke-width': strokeWidth / scale }
+                    svgTransform,
+                    { 'stroke-width': strokeWidth / scale, animator: 1 }
```
TorsteinHonsi, line 794: What does `animator` do? The only place `animator` is in use that can make a difference, is in the [animation that starts on line 776](https://github.com/highcharts/highcharts/blob/6e1d9f13a02172a0a226a698a4884c2b14ad5160/ts/Series/Map/MapSeries.ts#L776-L786). So when we set `animator` to 1, does it mean that that animation is still running, and we're skipping to the end? In that case, the appropriate way of stopping it would be to call the [stop function](https://api.highcharts.com/class-reference/Highcharts#.stop).

TorsteinHonsi (changes requested):
In addition to the requested changes, please add a sample with simple manual test steps.

TorsteinHonsi (changes requested):
Thanks! For the test:
* Since this is not a unit test, it should not be in the `unit-test` folder. Better place it in `maps/series/mappoint-animation` or something.
* Use a TopoJSON map and remove proj4
* Is highcharts-more.js needed?
* Instead of the commented code, create a button that runs this code. In the test note, describe that the tester should click the button and verify that the animation looks good.

samples/maps/series/mappoint-animation/demo.js
```diff
@@ -0,0 +1,84 @@
...],[38,-20],[-67,-31],[-3,-18],[-93,-51],[46,-21],[55,7]],[[4847,4048],[74,35],[74,9],[-21,-48],[11,-35],[41,-33],[-63,-86],[-70,-49],[-95,-46],[-17,-31],[-74,26],[-71,-20],[43,56],[22,88],[92,51],[54,83]],[[2627,4694],[149,-24],[40,25],[34,60],[10,54],[65,-19],[105,4],[93,28],[57,2],[78,34],[48,-15],[32,13],[100,-32],[78,10],[48,-9],[52,-41],[-20,-86],[71,-19],[-10,-66],[47,-10],[82,-79],[23,-58],[55,-14],[-52,35],[38,11],[40,-29],[7,-54],[-29,-34],[-94,-26],[-67,-62],[25,-25],[46,40],[75,28],[44,-10],[97,10],[31,-20],[19,-55],[29,-26],[8,-49],[28,-46],[-21,-68],[-34,-48],[-61,58],[24,78],[-24,38],[-55,35],[-47,9],[-12,-37],[74,-19],[0,-69],[-29,-56],[-39,-26],[106,29],[23,-17],[-2,-61],[-83,-64],[-19,16],[-83,0],[-68,-14],[-4,-64],[-37,-65],[-104,-50],[-109,35],[-10,17],[-69,5]]],"bbox":[-10.461181018243009,49.95912154352611,1.77056437025751,60.845036413947454],"transform":{"scale":[0.0012232968685369056,0.0010887003570778423],"translate":[-10.461181018243009,49.95912154352611]},"title":"British Isles","version":"2.0.1","copyright":"Copyright (c) 2022 Highsoft AS, Based on data from Natural Earth","copyrightShort":"Natural Earth","copyrightUrl":"http://www.naturalearthdata.com"}';
```
TorsteinHonsi, line 1: Any reason why we are not just loading the TopoJSON from our maps server? Like other demos:

https://github.com/highcharts/highcharts/blob/fa98f7ebc8d618179c4337a03356712b42a78f20/samples/maps/accessibility/accessible-data-classes/demo.js#L3-L5

madepiet: Right, it wasn't necessary for the manual test.

TorsteinHonsi (approved):
Approved, now pending CI before merge. Thanks!

PR #18642, opened 2023-03-09
DataTable demo

bre1470:


pawellysy (changes requested):
Could you please add a description, how to use this demo?

goransle (commented):
Think I found a bug: the second time I add a column it also adds an unnamed column
image

ts/masters/modules/data-tools.src.ts
```diff
@@ -0,0 +1,29 @@
+/**
+ * @license Highcharts JS v@product.version@ (@product.date@)
+ * @module highcharts/modules/data-tools
```
karolkolodziej, line 3: Is it temporary or is it going to be an official module?

karolkolodziej (approved):

ts/masters/modules/data-tools.src.ts
bre1470: Temporary.

pawelfus (approved):
Thanks!

Q: `setRow()` does not add the row somewhere in the middle - always at the end? Even for `modified` table? For example I have `sort` modifier -> add a row -> do I need to sort again?

bre1470:
@pawelfus Good point. It should add the row to the end of the original table, and then the sorted one should sort again. I forgot to trigger the modifier.

pawellysy (approved):
LGTM!

TorsteinHonsi (approved):
Thanks!

goransle (approved):

PR #18644, opened 2023-03-09
bugfix/18619-yaxis-update-nans

### LINKED ISSUE #18619: YAxis updates break chart after v10.2.1
#### Expected behaviour
Updating yAxis config should not result in errors being logged and chart not rendering correctly. 

#### Actual behaviour
When updating yAxis an error is logged and chart breaks sometimes.

Open the demo https://jsfiddle.net/xhygej1v/2/ and look at the errors being logged in the console.

Demo with 10.2.1 https://jsfiddle.net/xhygej1v/3/

After some debugging i think the problem has been caused by the changes made in [this](https://github.com/highcharts/highcharts/commit/225ec4f469258d270671fd81c41c85bb41589e7a) commit.
The logic in the ts/Core/Renderer/SVG/SVGElement.ts updateTransform has been changed from:
```javascript
    public updateTransform(): void {
        const wrapper = this,
            scaleX = wrapper.scaleX,
            scaleY = wrapper.scaleY,
            inverted = wrapper.inverted,
            rotation = wrapper.rotation,
            matrix = wrapper.matrix,
            element = wrapper.element;

        let translateX = wrapper.translateX || 0,
            translateY = wrapper.translateY || 0;

        // Flipping affects translate as adjustment for flipping around the
        // group's axis
        if (inverted) {
            translateX += wrapper.width;
            translateY += wrapper.height;
        }
```
to:
```javascript
    public updateTransform(): void {
        const {
            element,
            matrix,
            rotation = 0,
            scaleX,
            scaleY,
            translateX = 0,
            translateY = 0
        } = this;
```
This actually looks fine, except, that translateY can be NaN. Which can occur if the [animate](https://github.com/highcharts/highcharts/blob/44e193353ba0fa46891220b9d996b24b578caf79/ts/Series/Column/ColumnSeries.ts#L189) function of the ColumnSeries computes the value of the [yAxis.pos](https://github.com/highcharts/highcharts/blob/44e193353ba0fa46891220b9d996b24b578caf79/ts/Core/Axis/Axis.ts#L270) is still undefined.

May be related to #18252

#### Product version

Highcharts v10.3.3

#### Affected browser(s)
*

Vazonik:
Fixed #18619, errors about NaN translate attributes sometimes appeared in the console. It happened when updating a chart without redrawing, before it finished the initial animation.

~Fixed #18619, translate attributes could be NaN.~

### DIFF
--- a/samples/unit-tests/series-column/animation-cancel/demo.css
+++ b/samples/unit-tests/series-column/animation-cancel/demo.css
@@ -1,0 +1,4 @@
+#container {
+    width: 800px;
+    height: 300px;
+}
--- a/samples/unit-tests/series-column/animation-cancel/demo.details
+++ b/samples/unit-tests/series-column/animation-cancel/demo.details
@@ -1,0 +1,5 @@
+---
+ resources:
+   - https://code.jquery.com/qunit/qunit-2.0.1.js
+   - https://code.jquery.com/qunit/qunit-2.0.1.css
+...
--- a/samples/unit-tests/series-column/animation-cancel/demo.html
+++ b/samples/unit-tests/series-column/animation-cancel/demo.html
@@ -1,0 +1,6 @@
+<script src="https://code.highcharts.com/highcharts.js"></script>
+
+<div id="qunit"></div>
+<div id="qunit-fixture"></div>
+
+<div id="container"></div>
--- a/samples/unit-tests/series-column/animation-cancel/demo.js
+++ b/samples/unit-tests/series-column/animation-cancel/demo.js
@@ -1,0 +1,32 @@
+QUnit.test(
+    'Updating the chart while the initial animation is running (#18644)',
+    function (assert) {
+        const clock = TestUtilities.lolexInstall(),
+            chart = Highcharts.chart('container', {
+                chart: {
+                    type: 'column'
+                },
+                series: [{
+                    animation: {
+                        duration: 10
+                    },
+                    data: [3]
+                }]
+            });
+
+        chart.update({
+            yAxis: {
+                min: -5
+            }
+        }, false);
+
+        setTimeout(() => {
+            assert.notOk(
+                chart.container.innerHTML.includes('NaN'),
+                'The innerHtml of the chart container should not contain NaNs'
+            );
+        }, 20);
+
+        TestUtilities.lolexRunAndUninstall(clock);
+    }
+);
--- a/ts/Core/Renderer/SVG/SVGElement.ts
+++ b/ts/Core/Renderer/SVG/SVGElement.ts
@@ -2382,15 +2382,14 @@ class SVGElement implements SVGElementLike {
      */
     public updateTransform(): void {
         const {
-                element,
-                matrix,
-                scaleX,
-                scaleY
-            } = this,
-            // Protection against assigning NaNs to the variables (#18619)
-            rotation = this.rotation || 0,
-            translateX = this.translateX || 0,
-            translateY = this.translateY || 0;
+            element,
+            matrix,
+            rotation = 0,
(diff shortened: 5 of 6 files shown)


TorsteinHonsi (changes requested):
Thanks! Please add a test. Also, for the issue description, please describe the symptoms of the bug rather than how it was fixed. So that users reading the changelog can recognize their issue.

TorsteinHonsi (commented):
Also, could you go one step further down and see why those NaN values were set on `this` in the first place? Maybe we can fix the problem by its root.

pawellysy:
@TorsteinHonsi together with @Vazonik we took a look into the issue, and there are several issues here. This example is an edge case, where the `chart.redraw()` is called from `requestAnimationFrame`, and the errors are thrown during animation (which also calls `requestAnimationFrame` itself). Also, some really similar configs do throw the error, while others do not. 

Error - https://jsfiddle.net/BlackLabel/q2nk4jus/3/
No Errors:
 - https://jsfiddle.net/BlackLabel/q2nk4jus/4/ (no previous update)
 - https://jsfiddle.net/BlackLabel/q2nk4jus/6/ (after animation finishes)
 - https://jsfiddle.net/BlackLabel/q2nk4jus/7/ (no request animation frame)


The only root of the problem we identified was in `ColumnSeries.prototype.animate` where `yAxis.pos` was sometimes `undefined`.

Maybe we could include a kind of redundancy here, where we stop the `yAxis.pos` from being undefined, as well as 
 translateX/Y from being `NaN` in `SVGElement.prototype.UpdateTranform`?
 
 The test is a bit problematic as well. How to check if the error isn't thrown during the animation?
 Thanks!

TorsteinHonsi:
@pawellysy The issue occurs when calling `chart.update` with `redraw` = false, while the chart is still in the initial animation.  See simplified demo https://jsfiddle.net/highcharts/xzq1dhbs/. 

The simplest fix I think is to use a local variable for `yAxis.pos` so that we don't read it from the destroyed axis. This patch (on the master) works:

```diff
diff --git a/ts/Series/Column/ColumnSeries.ts b/ts/Series/Column/ColumnSeries.ts
index 272ace6fa1..56c7ff0b51 100644
--- a/ts/Series/Column/ColumnSeries.ts
+++ b/ts/Series/Column/ColumnSeries.ts
@@ -147,6 +147,7 @@ class ColumnSeries extends Series {
     public animate(init: boolean): void {
         const series = this,
             yAxis = this.yAxis,
+            yAxisPos = yAxis.pos,
             options = series.options,
             inverted = this.chart.inverted,
             attr: SVGAttributes = {},
@@ -160,8 +161,8 @@ class ColumnSeries extends Series {
             attr.scaleY = 0.001;
             translatedThreshold = clamp(
                 yAxis.toPixels(options.threshold as any),
-                yAxis.pos,
-                yAxis.pos + yAxis.len
+                yAxisPos,
+                yAxisPos + yAxis.len
             );
             if (inverted) {
                 attr.translateX = translatedThreshold - yAxis.len;
@@ -187,7 +188,7 @@ class ColumnSeries extends Series {
                     step: function (val: any, fx: any): void {
                         if (series.group) {
                             attr[translateProp] = translateStart +
-                                fx.pos * (yAxis.pos - translateStart);
+                                fx.pos * (yAxisPos - translateStart);
                             series.group.attr(attr);
                         }
                     }
```

For the test, you can set up something like my jsFiddle, then use Lolex to gain control over the timers. Apply it to an existing test if possible. Assert that `chart.container.innerHTML` does not contain the string `NaN`, that should be enough.

TorsteinHonsi (approved):

PR #18645, opened 2023-03-09
Updated Highcharts Core animation

nasin:
Hi, I redid the Highcharts JS (Core) product animation since we renamed the it Core. Marketing has approved the concept.


TorsteinHonsi (changes requested):
Thanks, looking good! When looking at it in the utils the chart title is visible though...

nasin:
Thanks! I removed the title from the chart configuration.

TorsteinHonsi (approved):

PR #18650, opened 2023-03-10
DataCursor demo

bre1470:
Based on PR #18642


samples/highcharts/data-tools/datacursor-html/demo.js
```diff
@@ -0,0 +1,120 @@
...
+    const tbody = this; // The function is binded to one of the HTML tables
+
+    for (const tr of tbody.getElementsByTagName('tr')) {
+        // We search for the correct HTML row
+        if (tr.dataset.rank === rank) {
+            tr.classList.add('Focus');
+        } else {
+            tr.classList.remove('Focus');
+        }
+        for (const td of tr.getElementsByTagName('td')) {
```
goransle, line 97: Not sure if it's better, but we could use some querySelector magicks here

goransle (approved):
Looks pretty clean to me

samples/highcharts/data-tools/datacursor-html/demo.js
```diff
@@ -0,0 +1,120 @@
...
+    container.innerHTML = html.join('\n');
+}
+
+renderTable(container1, table);
+renderTable(container2, tableModified);
+
+
+// Synchronize MouseOver
+
+const cursor = new Highcharts.DataCursor();
```
karolkolodziej, line 60: Is it going to be on Highcharts names as it is now or will it be part of the dashboards?

karolkolodziej (approved):

pawelfus (approved):
Thanks!

Shall we bring this demo to the next workshop session, too? 🤔 It's the first time I see Cursor in usage

TorsteinHonsi (commented):
Looks good! Maybe this is covered by the documentation, but for these advanced demos I am  missing a short explanation on what the purpose of the demo is. What are we looking at, and why does it matter? This would be useful for people finding this demo randomly from the internet, who are not directed here from the docs. 

PS: Is there a specific reason for not using the data grid component here?

bre1470:
@TorsteinHonsi 
I was heading for the simplest demo possible, and for now DataGrid is also part of Dashboard only.

samples/highcharts/data-tools/datacursor-html/demo.css
```diff
@@ -0,0 +1,37 @@
...
+
+table thead th {
+    background-color: #ddd;
+}
+
+table tbody td {
+    background-color: #eee;
+}
+
+tr.Focus td {
```
sebastianbochan, line 31: I have a problem with the class name because of two reasons
- capitalic at the first letter
- use focus which is almost the same like `:focus` selector.

My suggestion is to use another name such as `current`.

sebastianbochan (changes requested):

sebastianbochan (approved):

samples/highcharts/data-tools/datatable-html/demo.html
TorsteinHonsi, line 1: Please add a description. What are we looking at? What is the purpose of the demo and how do we use it?

TorsteinHonsi (changes requested):

TorsteinHonsi (approved):

PR #18651, opened 2023-03-10
bugfix/18636-waterfall-missing-lines

### LINKED ISSUE #18636: Line connecting the columns of the waterfall series missing
#### Expected behaviour
If any series contain null data then the connecting line should work by skiping the null column only.

#### Actual behaviour
Line connecting the columns of the waterfall series doesn't plot if series contain any 'null' value. It was working fine before 9.2.0 release.

https://jsfiddle.net/ehtashamulhaq/09ykvopL/1/

#### Product version
Highcharts 9.2.0 and higher

#### Affected browser(s)
N/A

Vazonik:
Fixed #18636, waterfall lines missing when there was a `null` point.

### DIFF
--- a/ts/Series/Waterfall/WaterfallSeries.ts
+++ b/ts/Series/Waterfall/WaterfallSeries.ts
@@ -639,8 +639,8 @@ class WaterfallSeries extends ColumnSeries {
     ): SVGPath {
 
         let data = this.data.filter((d): Boolean =>
-                // Skip points where Y is null (#18636)
-                d.y !== null
+                // Skip points where Y is not a number (#18636)
+                isNumber(d.y)
             ),
             yAxis = this.yAxis,
             length = data.length,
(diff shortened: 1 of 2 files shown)


ts/Series/Waterfall/WaterfallSeries.ts
```diff
@@ -638,7 +638,10 @@ class WaterfallSeries extends ColumnSeries {
         this: WaterfallSeries
     ): SVGPath {
 
-        let data = this.data,
+        let data = this.data.filter((d): Boolean =>
+                // Skip points where Y is null (#18636)
+                d.y !== null
```
TorsteinHonsi, line 643: We can do this safer by testing for `isNumber`. That will also rule out things like `undefined`, `NaN` and `Infinity`.

```ts
let data = this.data.filter(isNumber)
```

TorsteinHonsi (changes requested):
Thanks! In addition to the inline comment, I think we should respect the [connectNulls](https://api.highcharts.com/highcharts/series.waterfall.connectNulls) option. 

If `connectNulls` is true => filter out the null point, like you currently do.
If `connectNulls` is false => insert a gap by making the next segment a move-to segment instead of line-to.

TorsteinHonsi (approved):
Awesome, thanks!

PR #18654, opened 2023-03-11
demo/combo-timeline-data

sigrunnu:
Updated the demo data up until 2022.


TorsteinHonsi (changes requested):
Thanks!

We also need updated data for the awards and the releases. 

And the anniversary badge must be changed. Maybe replace the number 5 with the actual number of years since 27 November 2009 (not hard coded, but computed). Then replace 2014 with the current year. Something like this:
```text
13
years and counting
2009-2023
```

TorsteinHonsi (approved):
Thanks!

PR #18657, opened 2023-03-13
bugfix/18647-tilemap-single-points

### LINKED ISSUE #18647: Negative radius for circle and diamond shapes in TileMap
#### Expected behaviour
When there is a single data point in the series, circle and diamond shape tiles should render on the chart with out error

#### Actual behaviour
Does not render the shape in the chart, with following error in browser console
Error: <circle> attribute r: A negative value is not valid. ("-2")

https://jsfiddle.net/nwtquxfm/2/

1. Open the above link
2. Right click and go to browser console.
3. You can see the above mentioned error
4. Same error happens when tileShape is changed to diamond

#### Product version
Latest Stable

#### Affected browser(s)
All browsers

Screen Shot 2023-03-10 at 2 54 47 PM

hubertkozik:
Fixed #18647, when there was a single data point in the TileMap series, circle and diamond shape tiles didn't render properly.

### DIFF
--- a/samples/unit-tests/series-tilemap/coloraxis/demo.js
+++ b/samples/unit-tests/series-tilemap/coloraxis/demo.js
@@ -1,5 +1,5 @@
 QUnit.test('Tilemap and ColorAxis', function (assert) {
-    var chart = Highcharts.chart('container', {
+    const chart = Highcharts.chart('container', {
             chart: {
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Perfect, thanks!

PR #18660, opened 2023-03-14
Tools: Changes to build against ES2016

bre1470:
Touches #18070


TorsteinHonsi:
No difference found for the file sizes... Shouldn't we expect the builds to be marginally smaller now? Or is something not set up right?

TorsteinHonsi:
I `gulp`-ed this branch and inspected `highcharts.src.js` => still no `const` or `let`, arrow functions or deconstructors...

TorsteinHonsi (changes requested):
There are now some test failures

ts/Core/Series/Point.ts
```diff
@@ -417,7 +417,7 @@ class Point {
 
             for (const prop in point) { // eslint-disable-line guard-for-in
```
bre1470, line 418: `for in` is not sufficient for classes anymore, because it does not catch inherited functions.

samples/highcharts/blog/rotating-3d-scatter/demo.js
```diff
@@ -9,7 +9,7 @@ Highcharts.getOptions().colors = Highcharts.getOptions().colors.map(
             },
             stops: [
                 [0, color],
-                [1, Highcharts.Color(color).brighten(-0.2).get('rgb')]
+                [1, new Highcharts.Color(color).brighten(-0.2).get('rgb')]
```
TorsteinHonsi, line 12: So this hack is no longer working?

https://github.com/highcharts/highcharts/blob/f39c2d61d63b5ef36eebdcdd5985eea193790bf8/ts/Core/Color/Color.ts#L133-L136

If so, it is ok. But then we can remove those lines.

ts/Core/Utilities.ts
```diff
@@ -890,6 +893,10 @@ function wrap<T, K extends FunctionNamesOf<T>>(
 ): void {
     const proceed = obj[method] as T[K]&ArrowFunction;
 
+    if (typeof proceed !== 'function') {
+        throw new Error(`WrapError: ${method} is not a function`);
+    }
```
TorsteinHonsi, line 898: Is this necessary? What changed? And if necessary, should we use an error code?

TorsteinHonsi (commented):
Awesome 🎉  Please see my inline comments.

ts/Core/Utilities.ts
TorsteinHonsi: PS: In v10, and previous, this silently fails. As can be seen in the wrap-drawgraph function. I prefer that, or a warning, over throwing an error.

samples/highcharts/blog/rotating-3d-scatter/demo.js
ppotaczek: Yes, it is no longer working. It is not possible to call class constructor without new in ES6. Error: `Uncaught TypeError: Class constructor Color cannot be invoked without ‘new’`. I will remove the old hack.

ts/Core/Utilities.ts
bre1470: It helped to identify the errors with ES6 prototype. I can remove it.

bre1470:
Should the PR merge with `master` or rather `v11` branch?

bre1470:
Except for visuals I do not see other problems. I can test IE with ES5 build on Monday.

pawelfus (approved):
Annotations looks good, thanks!

bre1470:
`code/es5` works with IE-mode of Edge. 👍

PR #18662, opened 2023-03-14
bugfix/17565-stock-tooltip-split

### LINKED ISSUE #17565: Tooltip formatter not working properly from 9.3.0 onwards
#### Expected behaviour
Tooltip should be visible for both series.

#### Actual behaviour
Initially, the tooltip for line series works fine. However, if the user hovers over flag series item, the tooltip for the line series stops working. This works fine with version 9.2.2.

https://jsfiddle.net/2pch8gyz/

#### Product version
Highstock 9.3.0

#### Affected browser(s)
Tested with Firefox on Linux, Edge on Windows

jakubSzuminski:
Fixed #17565, stock tooltip issue after first hovering over a flag and then hovering over a series (problem with going from non-split to split tooltip).

### DIFF
--- a/ts/Core/Tooltip.ts
+++ b/ts/Core/Tooltip.ts
@@ -328,9 +328,11 @@ class Tooltip {
         if (this.label) {
             this.label = this.label.destroy();
         }
-        if (this.split && this.tt) {
+        if (this.split) {
             this.cleanSplit(true);
-            this.tt = this.tt.destroy();
+            if (this.tt) {
+                this.tt = this.tt.destroy();
+            }
         }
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (changes requested):
Thanks! The fix looks good, but please add tests. The cheapest approach is probably to modify the flag tooltip test https://github.com/highcharts/highcharts/blob/de1a1a699c1687da99097602b0d02091c4e559ed/samples/unit-tests/series-flags/tooltip/demo.js

TorsteinHonsi (approved):

PR #18666, opened 2023-03-14
bugfix/18542-map-panning-update-zoom

### LINKED ISSUE #18542: Zooming and Panning in Orthographic Projection
#### Current behavior
Panning after zooming resets the zoom to the default. The issue only occurs with orthographic projection.

Zoom the chart and then rotate the globe.
https://jsfiddle.net/BlackLabel/8wu2167a/

#### Expected behavior
https://jsfiddle.net/BlackLabel/hqzu9sLo/

#### Last working version
10.1.0

hubertkozik:
Fixed #18542, panning in zoomed-in Orthographic projection didn't work.

### DIFF
--- a/samples/unit-tests/maps/map-navigation/demo.js
+++ b/samples/unit-tests/maps/map-navigation/demo.js
@@ -265,6 +265,7 @@ QUnit.test('Orthographic map rotation and panning.', assert => {
 
         mapView: {
             maxZoom: 30,
+            zoom: -1,
             projection: {
                 name: 'Orthographic',
                 rotation: [0, -90]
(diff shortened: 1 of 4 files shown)


magdalena-gut (approved):

madepiet (approved):

TorsteinHonsi (changes requested):
Thanks!

There is an issue we need to deal with: In https://jsfiddle.net/highcharts/8nkdz1oL/, grab Spain and drag straight right out of the chart => console complains about NaN, and the view is frozen.

In the current official release it actually also freezes, but not until dragging all the way of of the chart and releasing the mouse outside.

KacperMadej:
@TorsteinHonsi 
There is or was a concept of a change between pan-rotate and pan-translate in Orthographic projection based on the current zoom and not what is the entire data set. With this, if you zoom in enough, you can pan-translate. The pan-rotate should also be disabled if only a part of the world is loaded (e.g. only 1 continent). Should all of this still be the case?

KacperMadej (approved):

TorsteinHonsi (approved):

PR #18671, opened 2023-03-15
db/confirmation-popup-refactor

karolkolodziej:
Refactored confirmation popup to use the `BaseForm`.

### DIFF
--- a/css/dashboards/gui.css
+++ b/css/dashboards/gui.css
@@ -6,9 +6,6 @@
  * License: www.highcharts.com/license
  */
 
-/* Popup */
-@import url("../annotations/popup.css");
-
 :root,
 .highcharts-dashboards-light {
     /* Base colors */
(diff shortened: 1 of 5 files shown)


css/dashboards/gui.css
```diff
@@ -749,28 +746,59 @@
...
     width: 200px;
     left: calc(50% - 150px);
+    display: none;
+    background-color: var(--highcharts-dashboards-main-background-color);
+    color: var(--highcharts-neutral-color-60);
+    font-size: 0.876em;
+    max-height: 90%;
+    top: 5%;
+    overflow-x: hidden;
+    overflow-y: hidden;
```
sebastianbochan, line 762: Replace
```
overflow-x: hidden;
overflow-y: hidden;
```
with
```
overflow:hidden
```

css/dashboards/gui.css
```diff
@@ -749,28 +746,59 @@
...
+    min-width: 300px;
+    max-width: 600px;
```
sebastianbochan, line 764: In case of `width: 200px;` the `max-width` will not be used anymore.

css/dashboards/gui.css
```diff
@@ -749,28 +746,59 @@
...
     left: calc(50% - 150px);
+    display: none;
+    background-color: var(--highcharts-dashboards-main-background-color);
+    color: var(--highcharts-neutral-color-60);
+    font-size: 0.876em;
+    max-height: 90%;
+    top: 5%;
+    overflow-x: hidden;
+    overflow-y: hidden;
+    min-width: 300px;
```
sebastianbochan, line 763: The `min-width` is bigger than `width`. We should use the only one.

css/dashboards/gui.css
```diff
@@ -749,28 +746,59 @@
...
 
-.highcharts-popup .highcharts-dashboards-edit-popup-content {
+.highcharts-dashboards-edit-popup-close {
+    background-repeat: no-repeat;
+    background-position: 50% 50%;
+    width: 20px;
+    height: 20px;
+    cursor: pointer;
+    position: absolute;
+    padding: 10px;
```
sebastianbochan, line 779: In case of background-position 50/50%, you can declare only width/height. The padding is unecessary.

sebastianbochan (changes requested):

css/dashboards/gui.css
```diff
@@ -749,28 +746,59 @@
     opacity: 0.3;
 }
 
-.highcharts-popup {
+/* Confirmation popup */
+
+.highcharts-dashboards-edit-popup {
     min-height: 160px;
     width: 200px;
     left: calc(50% - 150px);
```
pawellysy, line 754: maybe instead of this left, we could go with:
```css
left: 50%
transform: translate(-50%, 0)
```

pawellysy (changes requested):

ts/Dashboards/EditMode/EditMode.ts
```diff
@@ -99,8 +99,9 @@ class EditMode {
         this.createTools();
 
         this.confirmationPopup = new ConfirmationPopup(
-            this,
             board.container,
+            EditGlobals.iconsURL,
```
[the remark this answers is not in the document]
karolkolodziej: Any suggestions?

sebastianbochan (approved):

ts/Dashboards/EditMode/EditMode.ts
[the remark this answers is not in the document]
pawellysy: Can be referenced as a value in the Class IMO.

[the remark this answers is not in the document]
pawellysy: nvm, leave as it is

pawellysy (approved):

PR #18681, opened 2023-03-16
bugfix/18635-outside-tooltip-inverted-scrollable

### LINKED ISSUE #18635: Outside tooltip on bar charts scroll offset not calculated correctly
#### Actual behaviour

Here's a fiddle I was able to reproduce with https://jsfiddle.net/runfaj/wx2cabpv/1/. Essentially in certain scroll situations, it seems like inverted charts don't calculate the horizontal scroll offset correctly. In this fiddle, if you size the preview pane to about 700 px, so the tiles on the far right need to be scrolled to, then you hover, you'll see the tooltips no longer line up on the points.
image

If you shrink that area so the scroll increased, the problem is exacerbated to where the tooltips are completely out of view.
image

I added a couple examples the the far right and far bottom and it seems there's no issues with non-inverted charts or vertical positioning. 

Upon digging in the code, it appears the defaultPositioner in the tooltip is calculated quite differently than what is provided for the getPosition function that is available for the tooltip.positioner. You'll see if you define a positioner in this same fiddle:
```
positioner(boxWidth, boxHeight, point) {
  const pos = this.getPosition(boxWidth, boxHeight, point);
  return pos;
},
```
Then the tooltips don't even calculate against the chart container's offset anymore, so the tooltips go haywire on all the tiles. 

Since it doesn't appear that the defaultPositioner is working and it also appears to be a completely private method, I don't see any way to fix this issue.

#### Expected behaviour

Tooltips should always align to the top of the point for columns or to the right of the point for bars (at least in this example). This would be a fix for the defaultPositioner.

Also, it would be helpful potentially to also update the tooltip.positioner call:
- so the provided point receives the same inputs that the defaultPositioner has available (like the anchorX/Y).
- update so getPosition matches the same offsetting behavior as the defaultPositioner.

#### Product version
highcharts 10.3.3

#### Affected browser(s)
mostly tested in chrome

hubertkozik:
Fixed #18635, bad tooltip position in an inverted chart inside a scrollable container when `tooltip.outside` was true.
____
Unfortunately, I cannot add a test for this fix, it would need to add a scroll on the window container, which is unpredictable due to tests in the browser and then tests in a terminal (see the original issue for reference).

### DIFF
--- a/ts/Core/Tooltip.ts
+++ b/ts/Core/Tooltip.ts
@@ -628,23 +628,24 @@ class Tooltip {
             // Don't use h if chart isn't inverted (#7242) ???
             h = (chart.inverted && (point as any).h) || 0, // #4117 ???
             outside = this.outside,
+            { body, documentElement } = doc,
             outerWidth = outside ?
                 // substract distance to prevent scrollbars
                 Math.max(
-                    doc.body.scrollWidth,
-                    doc.documentElement.scrollWidth,
-                    doc.body.offsetWidth,
-                    doc.documentElement.offsetWidth,
-                    doc.documentElement.clientWidth
+                    body.scrollWidth,
+                    documentElement.scrollWidth,
+                    body.offsetWidth,
+                    documentElement.offsetWidth,
+                    documentElement.clientWidth
                 ) - 2 * distance :
                 chart.chartWidth,
             outerHeight = outside ?
                 Math.max(
-                    doc.body.scrollHeight,
-                    doc.documentElement.scrollHeight,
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (changes requested):
Thanks! In the `getPosition` function we are reading `doc.body` four times and `doc.documentElement` six times. I assume it would produce shorter code to store them in local variables?

TorsteinHonsi (approved):

TorsteinHonsi:
All good, pending CI

PR #18682, opened 2023-03-16
bugfix/17614-timeline-styled-mode-default-colors

### LINKED ISSUE #17614: Timeline chart: Wrong point-to-labelbox path color in styled mode
#### Expected behaviour
In the [demo example for timeline chart](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/series-timeline/datetime-axis), the axis point matches the color of the path going to the label box. I would expect the same in styled mode.
image

#### Actual behaviour
In styled mode the path color differs from the point color (it inherits from `.highcharts-color-0`). While there are color classes for the paths (e.g. `highcharts-data-label-color-1`), there are no style rules for these. The same issue is with bullet colors in the label box.
image

https://jsfiddle.net/scebotari66/1btx2wd3/

#### Product version

#### Affected browser(s)
All

jakubjanuchta:
Fixed #17614, timeline styled mode default colors


samples/unit-tests/series-timeline/datalabels/demo.js
```diff
@@ -168,3 +168,71 @@ QUnit.test('Timeline: General tests.', function (assert) {
...
+
+        assert.ok(
+            point.dataLabel.element.innerHTML.includes(
+                '<tspan class="highcharts-color-' + colorIndex +
+                '">● </tspan>'),
+            'Datalabel dot has class with the same colorIndex as the point'
+        );
+    });
+
+    setTimeout(() => {
```
jakubjanuchta, line 230: I had to put this test in setTimeout: 0 because in Karma the stroke value without the setTimeout was 'none' and in utils, value was correct and the test passed.

TorsteinHonsi (approved):
Thanks!

PR #18685, opened 2023-03-16
Db/demos-resize-flag

sebastianbochan:


karolkolodziej (changes requested):
Waiting for the main branch to be fixed and for CI.

In the meantime could you please explain why removing it?

sebastianbochan:
The `isResizable` param is only in options and does not cover the real functionality in the code ;) 

Instead of adding this option, Pawel decided to remove it from demos.

karolkolodziej:
If so, please remove it from each file including `.test.js`, `.mjs` and `.md`

pawellysy (approved):
LGTM!

karolkolodziej (approved):

PR #18687, opened 2023-03-16
db/board-datapool

bre1470:


karolkolodziej (approved):

goransle (approved):
Looks good. I see we have a demo task in the backlog, should we also update/add to the docs?

bre1470:
No, the demos are Highcharts intern and rely on temporary module.

PR #18691, opened 2023-03-17
bugfix/18683-stock-flags-demo-bug

### LINKED ISSUE #18683: Flags demo doesn't work properly
#### Expected behaviour
Show flags on the series.

#### Actual behaviour
Flags are not added to the series properly.

https://www.highcharts.com/demo/stock/flags-general

#### Product version

#### Affected browser(s)
All

jakubSzuminski:
Fixed #18683, flags demo did not work. 

This problem probably stemmed from the difference in data length imported with `Highcharts.getJSON()` method in `Highcharts-Utils` and `jsfiddle` (I don't know why but `Highcharts-Utils` have data till 2021 while in JSFiddle the data ends in 2017).

I simply positioned the flags in 2017, which is included in the data imported in JSFiddle.

### DIFF
--- a/samples/highcharts/css/flags/demo.js
+++ b/samples/highcharts/css/flags/demo.js
@@ -1,5 +1,5 @@
 Highcharts.getJSON(
-    'https://cdn.jsdelivr.net/gh/highcharts/highcharts@v7.0.0/samples/data/usdeur.json',
+    'https://cdn.jsdelivr.net/gh/highcharts/highcharts@v10.3.3/samples/data/usdeur.json',
     function (data) {
 
         var year = new Date(data[data.length - 1][0]).getFullYear(); // Get year of last data point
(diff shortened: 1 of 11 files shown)


TorsteinHonsi (changes requested):
> I don't know why but Highcharts-Utils have data till 2021 while in JSFiddle the data ends in 2017

It's because in jsFiddle and on our website, the demo loads the content from the v7.0.0 of the data file:

https://github.com/highcharts/highcharts/blob/4197c8366e830036240bf83a30c633005e7b0c76/samples/stock/demo/flags-general/demo.js#L1

In the utils however, this is rewritten to load from the local file.

To fix it, please update the jsdelivr URL to work with a more recent Highcharts tag:

`https://cdn.jsdelivr.net/gh/highcharts/highcharts@v10.3.3/samples/data/usdeur.json`

In addition to that, please search our samples for other occurences of linking to the outdated `usdeur.json` file.

TorsteinHonsi (approved):
Thanks!

PR #18695, opened 2023-03-20
bugfix/18689-linked-indicator-update

### LINKED ISSUE #18689: Adding points to series with linked indicator
#### Expected behaviour
If a main data series is not declared on a first position in the `series` array, adding multiple point's to it should work without any additional redraws.

#### Actual behaviour
If you add a point to the main series, which is not on the first index of the `series` array, you can only add one. Rest of them are added after calling a redraw event (i.e by dragging the navigator or resizing the chart).

https://jsfiddle.net/BlackLabel/frvdn79c/

1. Click on the `add` button once - see how points of value 16 get's added to the chart.
2. Click on it again - nothing happens. (also click it multiple times)
3. Trigger a redraw event by resizing the chart, or dragging the navigator - the point gets added.

Everything works fine when the main data is declared before indicator -> https://jsfiddle.net/BlackLabel/b5twqua6/

#### Product version
Highstock + Indicators v10.3.3

#### Internal note
Regression -> Last working version 10.3.2, broken since 10.3.3

kamil-musialowski:
Fixed #18689, indicator linked to another indicator should update after adding points.

### DIFF
--- a/ts/Stock/Indicators/SMA/SMAIndicator.ts
+++ b/ts/Stock/Indicators/SMA/SMAIndicator.ts
@@ -482,10 +482,8 @@ class SMAIndicator extends LineSeries {
             indicator.redraw();
         }
 
-        indicator.isDirtyData = !!indicator.linkedSeries.length;
-        indicator.linkedSeries.forEach((): void => { // #18689
-            fireEvent(indicator, 'updatedData');
-        });
+        indicator.isDirtyData = !!indicator.linkedSeries.length;// #18689
+        fireEvent(indicator, 'updatedData');
     }
 
     /**
--- a/ts/Stock/Indicators/SMA/SMAIndicator.ts
+++ b/ts/Stock/Indicators/SMA/SMAIndicator.ts
@@ -31,6 +31,7 @@ const {
 import U from '../../../Core/Utilities.js';
 const {
     addEvent,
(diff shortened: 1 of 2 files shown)


kamil-musialowski:
I decided to not add test to this one, since we already have a good test for it (it simply didn't work properly before this fix because the `series.points` were not updated, therefore it didn't fail).

https://github.com/highcharts/highcharts/blob/eba0e62fe57d25eff09e5bb2e010b955a5195b66/samples/unit-tests/indicator-sma/recalculations/demo.js#L343-L349

ts/Stock/Indicators/SMA/SMAIndicator.ts
```diff
@@ -481,7 +482,10 @@ class SMAIndicator extends LineSeries {
             indicator.redraw();
         }
 
-        indicator.isDirtyData = !!indicator.linkedSeries;
+        indicator.isDirtyData = !!indicator.linkedSeries.length;
```
kamil-musialowski, line 485: As a minor improvement to #17190, I added the `length` property, since `linkedSeries` could be an empty array.

Case: `!![] === true`

pawelfus (approved):
Thanks!

TorsteinHonsi (approved):

Issue #18478, opened 2023-02-11
Pre-release regression with immediate redraw


TorsteinHonsi:
#### Expected behaviour
The chart should render once

#### Actual behaviour
When the container has no specific height set, the chart is rendered with 400px height, and the container triggers a resize event. Subsequently the ResizeObserver is triggered and the chart redraws.

#### Live demo with steps to reproduce
https://jsfiddle.net/highcharts/h1wL2bxq/

#### Product version
Pre-release

#### Affected browser(s)
Presumably all
___

@karolkolodziej I'm assigning this to you. I probably caused it myself when I tried to simplify your proposal for #17951, but would appreciate if you had a look at it.

TorsteinHonsi:
Proposed solution: Since this resize event happens because the SVG changes the size of the container - maybe we can avoid it by simply delaying the whole setting of the ResizeObserver until after the SVG has been inserted?

karolkolodziej:
I went throught my PR. I spent some time exploring bd74c3465a85bc722a210fe9b99ea5f5db403ceb and it looks like the problem you are mentioning was neatly masked by `setTimeout` but it was still there.


You suggestion to move the initialization of the `ResizeObserver` seems good- do you have any particular place in mind?
As far as I can see the flow looks like this: `onload()` -> `setReflow()` -> `new ResizeObserver()`

Is there any place/event where we are certain that the size is set? 
Beacause setting fixed time in `setTimeout` doesn't look like a valid solution?

TorsteinHonsi:
@karolkolodziej Sorry I forgot to follow up here.

On inspection, it looks like it is already done this way, it is added after the SVG is inserted:

https://github.com/highcharts/highcharts/blob/3542a8e9e14a8f26549f241633daa5f197f878de/ts/Core/Chart/Chart.ts#L2768-L2771

Can you do some more debugging and see what happens?

TorsteinHonsi:
Still happens if the chart has an explicit height: https://jsfiddle.net/highcharts/wn51zfLe/

PR #18697, opened 2023-03-20
bugfix/18658-sunburst-labels-blinking

### LINKED ISSUE #18658: Sunburst: data labels re-animate when points are programmatically updated
#### Expected behaviour
When updating a point programmatically using `point.update(...)` (e.g., to make it sliced), I expect minimal animation.

#### Actual behaviour
The data labels disappear and fade back in when I perform that update.

https://jsfiddle.net/cedricbaptk/f0xhqpuk/

Click the "toggle sliced" button. This causes one of the point to toggle its `sliced` option. The slice correctly animates, but all data labels also disappear and fade back in when the button is clicked.

This can also be triggered simply by calling `series.redraw()`.

Presumably, this is caused by the `hackDataLabelAnimation` flag being `true` here (v10.3.3):
https://github.com/highcharts/highcharts/blob/2bc6062f482bd649a52aa625bc53116c0bc81a4b/ts/Series/Sunburst/SunburstSeries.ts#L839-L844

Indeed, I can work around the issue by temporarily setting `series.idPreviousRoot` to `series.rootNode`. Example workaround to apply to the jsfiddle above:
```diff
+      const s = p.series;
+      const tmp = s.idPreviousRoot;
+      s.idPreviousRoot = s.rootNode;
       p.update({sliced: !p.sliced});
+      s.idPreviousRoot = tmp;
```

#### Product version
Highcharts v10.3.3

#### Affected browser(s)
Latest Chrome

Vazonik:
Fixed #18658, all sunburst's data labels were disappearing and showing back after one point update.

### DIFF
--- a/samples/unit-tests/series-sunburst/allowtraversingtree/demo.js
+++ b/samples/unit-tests/series-sunburst/allowtraversingtree/demo.js
@@ -16,9 +16,6 @@ QUnit.test('Drill to node by click events', function (assert) {
                             id: 'level-3',
                             parent: 'level-2',
                             value: 1
-                        },
-                        {
-                            value: 1
                         }
                     ]
                 }
@@ -40,6 +37,30 @@ QUnit.test('Drill to node by click events', function (assert) {
         level1 = findPointById('level-1'),
         level2 = findPointById('level-2'),
         level3 = findPointById('level-3');
+
+    click(level1);
+    assert.strictEqual(
+        series.rootNode,
+        '',
+        'should have series.rootNode equal "" after clicking level1. (#18658)'
+    );
+
+    click(level2);
+    assert.strictEqual(
+        series.rootNode,
+        'level-2',
+        'should have rootNode equal "level-2" after clicking level2. (#18658)'
+    );
+
+    click(level2);
+    assert.strictEqual(
+        series.rootNode,
+        '',
+        'should have rootNode equal "" after clicking level2 twice. (#18658)'
+    );
+
+    series.addPoint({ value: 1 }, false);
+    series.drillUp();
 
     assert.strictEqual(
         series.rootNode,
--- a/ts/Series/Sunburst/SunburstSeries.ts
+++ b/ts/Series/Sunburst/SunburstSeries.ts
@@ -841,31 +841,8 @@ class SunburstSeries extends TreemapSeries {
                 y: positions[1]
             },
             innerR = positions[3] / 2,
-            renderer = series.chart.renderer,
-            animateLabels: (Function|undefined),
-            animateLabelsCalled = false,
-            addedHack = false,
-            hackDataLabelAnimation = !!(
-                animation &&
-                hasRendered &&
-                idRoot !== idPreviousRoot &&
-                series.dataLabelsGroup
-            );
+            renderer = series.chart.renderer;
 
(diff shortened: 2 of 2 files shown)


TorsteinHonsi (changes requested):
@Vazonik When removing code, it is important to consider closely what is lost. One thing I can see is that the data labels now suddenly appear initially, instead of easing/fading in. This is what should be controlled by the [dataLabels.defer](https://api.highcharts.com/highcharts/series.line.dataLabels.defer) option, and in the current state of this PR, it behaves like `defer` is always false.

I recommend taking a closer look at what is going on in that code block, hopefully we can do it in a more consistent way. Maybe we can implement the `defer` option in another place, more consistent with other series types. If things were done correctly, it shouldn't be necessary with series-specific overrides like this.

Vazonik:
@TorsteinHonsi Thanks! Ignoring the defer option was caused by some other commit that had already been merged into the master, but I haven't identified which one of it yet. This regression applies to data labels regardless of series type. Should I fix it in this pull request or should I make a separate issue for it?

My change causes, for example, removing the fadein-fadeout animation of data labels during drill down, but with @pawellysy we found that it could also be unnecessary.

TorsteinHonsi:
> Ignoring the defer option was caused by some other commit that had already been merged into the master

In that case it is important that we find it and fix it before the next release. Do you have a demo of this? It seems to work well here (on the master): https://jsfiddle.net/highcharts/rz82ywtp/

Vazonik:
> Do you have a demo of this?

@TorsteinHonsi Yes:
- master: https://jsfiddle.net/BlackLabel/t23qonpy/
- v10.3.3.: https://jsfiddle.net/BlackLabel/2b6jzk3g/

TorsteinHonsi:
Okay, it's caused by this regression: https://github.com/highcharts/highcharts/issues/18478 . If we set a container height in the CSS, we fix the redraw issue and the data labels behave like they should.

Vazonik:
Ok, so now it works like this: https://jsfiddle.net/BlackLabel/Lg5c0nwu/

TorsteinHonsi (approved):
Yes, seems good now!

TorsteinHonsi (changes requested):
I now remembered what the `hackDataLabelAnimation` is about. It is for traversing the sunburst. See http://localhost:3030/samples/#view/highcharts/demo/sunburst and click Europe. Without the hack, the data labels animate in a chaotic way. That's why we decided to just temporary hide them and fade back in instead of trying to animate between different rotations and text paths. I think what you need to do is to keep the logic is it was, but rather inspect when the `hackDataLabelAnimation` is activated. So that it does not get activated when just doing a simple point update. In fact, looking at the previous line 851, I would expect it to only happen when the root changed, which is does not do in this case.

TorsteinHonsi:
Thank you! While we're in this part of the code, I want to ask you to to check out something I have been noticing for a while. See https://jsfiddle.net/highcharts/6eka4tmn/. If you click the middle element, "Oceanic", you also get the blinking of the data labels. I would expect that in this case the root id was identical to the previous root id, since we're not allowed to traverse further up, and subsequently the labels shouldn't blink. Can you have a look at that?

TorsteinHonsi (approved):
Thanks!

PR #18703, opened 2023-03-21
db/iconURLPrefix

karolkolodziej:
Added `iconURLPrefix`.


pawellysy (approved):
Looks good!

ts/Dashboards/EditMode/EditMode.ts
```diff
@@ -123,6 +124,7 @@ class EditMode {
 
     private active: boolean = false;
     public options: EditMode.Options;
+    public iconsURLPrefix: string;
```
sebastianbochan, line 127: Why do not place the default value here? In terms of it, we can simplify the condition

```
options && options.iconsURLPrefix ||
            'https://code.highcharts.com/@product.version@/gfx/dashboard-icons/';
```

sebastianbochan (changes requested):

PR #18705, opened 2023-03-21
GoogleSheetsConnector demo 

bre1470:


pawelfus (approved):
Thanks!

samples/highcharts/data-tools/googlesheets-html/demo.html
```diff
@@ -0,0 +1,25 @@
+<script src="https://code.highcharts.com/highcharts.js"></script>
+<script src="https://code.highcharts.com/modules/data-tools.js"></script>
+<script src="https://code.highcharts.com/modules/accessibility.js"></script>
+
+<p>
+    Use the fields below to configure the GoogleSheetsConnector. It will load
```
karolkolodziej, line 6: Is it ok to share this sheet so it's easily accessible and we see how it is structured?
https://docs.google.com/spreadsheets/d/1U17c4GljMWpgk1bcTvUzIuWT8vdOnlCBHTm5S8Jh8tw/edit#gid=0

karolkolodziej (approved):

samples/highcharts/data-tools/googlesheets-html/demo.html
bre1470: We can't do this, but it has a form, so you can test with your own Google Spreadsheets.

pawellysy (approved):
LGTM!

goransle (approved):

samples/highcharts/data-tools/googlesheets-html/demo.css
```diff
@@ -0,0 +1,41 @@
+:root {
+    --color-background: #fff;
+    --color-foreground: #000;
```
sebastianbochan, line 3: As I see it is used by the `color` param which means text. I suggest to use a name like `--color-text` or something similar which is more clear.

sebastianbochan (changes requested):

sebastianbochan (approved):

PR #18706, opened 2023-03-21
Fixed async of ChainModifier.modify.

bre1470:
Fixed error handling and async processing, which gets triggered by `DataTable.setModifier`.


karolkolodziej (approved):

pawellysy (approved):
LGTM!

ts/Data/Modifiers/ChainModifier.ts
```diff
@@ -151,6 +151,56 @@ class ChainModifier extends DataModifier {
...
+                modifier.modify(chainTable.modified as T, eventDetail)
+            );
+        }
+
+        promiseChain = promiseChain.then((chainTable): T => {
+            table.modified = chainTable.modified;
+            return table;
+        });
+
+        promiseChain = promiseChain['catch']((error): Promise<T> => {
```
goransle, line 192: I might be missing something, but could this not be a `.catch` on the assignment above?

sebastianbochan (approved):

ts/Data/Modifiers/ChainModifier.ts
bre1470: Kind of, but I have to use brackets for IE and then I can not make proper alignments and it becomes ugly code. So this is an aesthetical thing to make the code more readable.

goransle: Ah, IE 😢

goransle (approved):

PR #18712, opened 2023-03-22
bugfix/18183-legend-item-cut

### LINKED ISSUE #18183: First line of legend with navigation is cut
#### Expected behaviour
The line shouldn't be cut

#### Actual behaviour
The line is cut

The bug exists in your demo for legends with navigation:
* link to doc: https://api.highcharts.com/highcharts/legend.navigation.style
* link to demo: https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/legend/navigation/
you can see that circle is cut a little bit:

#### Product version

v10.3.2

#### Affected browser(s)
Chrome

jakubSzuminski:
Fixed #18183, first line of legend with navigation was cut.

Simplified **demo** of the problem: https://jsfiddle.net/BlackLabel/9rypt16j/

**Cause** of the problem: it seems to me that the `clipRect` `y` used for legend was too high. 

**Solution**: Decreasing `clipRect y` by 2 pixels solves the problem (it does not cut the legend anymore).

But I'm not sure if that's the best solution out there.

### DIFF
--- a/ts/Core/Legend/Legend.ts
+++ b/ts/Core/Legend/Legend.ts
@@ -1358,7 +1358,7 @@ class Legend {
             // defines the scroll top for each page (#2098)
             allItems.forEach((item, i): void => {
                 legendItem = item.legendItem || {};
-                const y = legendItem.y ? legendItem.y - 2 : 0,
+                const y = legendItem.y || 0,
                     h = Math.round(
                         (legendItem as any).label.getBBox().height
                     );
@@ -1398,7 +1398,7 @@ class Legend {
             // PDF export (#1787)
             if (!clipRect) {
                 clipRect = legend.clipRect =
-                    renderer.clipRect(0, padding, 9999, 0);
+                    renderer.clipRect(0, padding - 2, 9999, 0);
                 legend.contentGroup.clip(clipRect);
             }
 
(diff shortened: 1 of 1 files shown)


ts/Core/Legend/Legend.ts
```diff
@@ -1398,7 +1398,7 @@ class Legend {
             // PDF export (#1787)
             if (!clipRect) {
                 clipRect = legend.clipRect =
-                    renderer.clipRect(0, padding, 9999, 0);
+                    renderer.clipRect(0, padding - 2, 9999, 0);
                 legend.contentGroup.clip(clipRect);
```
jakubSzuminski, line 1402: This creates the clipping rectangle **only** when the legend has too many elements and the navigation is added. It is not created for a legend with just a few elements.

The `x, width, height` for this `clipRect` are set correctly, but `y` is not. From what I deduced from the code `y` is not calculated anywhere, `clipRect.y` is simply `padding` and it's only set in the `1401` line, nowhere else. So reducing it by a few pixels here solves the problem.

I know it's still hardcoding a value, but I believe it's a better solution than the previous one because we only affect the clipping rectangle which is only created when there is `legend.nav`

TorsteinHonsi: Yes I see. I think the two pixels account for the symbol (circle) extending above the text. So a perfect solution would be to subtract the actual symbol height from the line height. But I think this is good enough.

TorsteinHonsi (approved):

PR #18719, opened 2023-03-23
db/JSON-api

karolkolodziej:
Marked some of the JSON properties and methods as internal.


ts/Dashboards/Components/Component.ts
```diff
@@ -918,7 +922,10 @@ namespace Component {
         sync: SyncOptions;
     }
 
-    // JSON compatible options for export
+    /**
+     * // JSON compatible options for export
```
sebastianbochan, line 926: Do we need `//` here? :)

sebastianbochan (changes requested):

sebastianbochan (approved):

pawellysy (approved):

PR #18721, opened 2023-03-23
bugfix/18074-dragdrop-mappoint

### LINKED ISSUE #18074: modules/draggable-points no longer works with mappoints
#### Expected behaviour

Dragging map points.

It worked, I used it in my app and there's a demo someone posted to StackOverlow. I think it broke with version 10.

#### Actual behaviour

Mappoints can't be dragged. There is no error.

https://jsfiddle.net/BlackLabel/0cq3a9h4/1/

#### Product version

latest from CDN, Highmaps JS v10.3.2 (2022-11-28)

#### Affected browser(s)

confirmed firefox and vivaldi

hubertkozik:
Fixed #18074, the Draggable Points module didn't work for `mappoint` series.


samples/unit-tests/dragdrop/mappoint/demo.js
```diff
@@ -0,0 +1,64 @@
...
+        }
+    });
+
+    oldPointPlotX = point.plotX;
+    oldPointPlotY = point.plotY;
+    controller.mouseDown(x + 100, y + 100, void 0, true);
+    controller.mouseMove(x, y, void 0, true);
+    controller.mouseUp(x, y, void 0, true);
+
+    assert.ok(
```
magdalena-gut, line 55: What about combining these two checks to  `assert.ok(point.plotX < oldPointPlotX && point.plotY < oldPointPlotY)` ? Similar situation to `assert.strictEqual` above.

KacperMadej (approved):

TorsteinHonsi (changes requested):
How does a draggable map area work? We need a visual sample too in addition to the unit test.

hubertkozik:
Drag drop has never worked for MapSeries in Highcharts Maps. Furthermore, MapSeries doesn't have simple lon / lat positions to change it with a mouse move. In this bugfix, I focused on removing the regression about DragDrop for MapPoints, which was present after some major changes in Highcharts Maps v9.3.3. Adding a DragDrop for MapSeries would be a totally new feature and it would require a lot more time to implement it (from scratch).

TorsteinHonsi (approved):
Awesome, thanks!

TorsteinHonsi (changes requested):
... but please resolve the conflict :/

TorsteinHonsi (approved):

PR #18597, opened 2023-03-01
db/doclets-component-options

sebastianbochan:

PR #18722, opened 2023-03-23
db/doclets-component-options-namespace

sebastianbochan:
Due to issues in mergeing [PR](https://github.com/highcharts/highcharts/pull/18597) with `dashboards/main`, I was forced to create a new PR.


ts/Dashboards/Components/Component.ts
```diff
@@ -1228,6 +1244,7 @@ namespace Component {
         caption?: TextOptionsType;
     }
 
+    /** @internal */
```
pawelfus, line 1247: I think this one should be external, otherwise, I wouldn't know what options I have:

image

ts/Dashboards/Components/Component.ts
```diff
@@ -1144,6 +1154,7 @@ namespace Component {
             detail?: AnyRecord;
         } & EventRecord;
 
+    /** @internal */
     export type SyncOptions = Record<string, boolean | Partial<Sync.OptionsEntry>>;
```
pawelfus, line 1158: Shouldn't be `SyncOptions` available?

image

pawelfus (changes requested):
A few minor changes, but overall look good!

Extra question:
What's the difference between ComponentOptions and EditableOptions?
image

karolkolodziej (changes requested):
I agree with @pawelfus, the `Editable Options` should not be listed here.

Also, I'm not able to see the component options by navigating:
`Highcharts Dashboards -> Dashboards -> Board -> default -> Options -> components`

ts/Dashboards/Components/Component.ts
karolkolodziej: Might be connected: https://github.com/highcharts/highcharts/pull/18723
@goransle

karolkolodziej (approved):

sebastianbochan:
@karolkolodziej 
It looks like a wrong reference in the `Board.ts` file (to `Bindings.ComponentOptions` instead of `Component.ComponentOptions`). Will be checked again, when we will apply the refactor.

pawellysy (approved):

TorsteinHonsi (approved):

PR #18723, opened 2023-03-23
Db/synchandler doclets

goransle:


ts/Dashboards/Components/Sync/Emitter.ts
```diff
@@ -18,41 +18,78 @@ import type ComponentTypes from '../ComponentType';
...
+     * Registry for reusable emitter.
+     * The emitter is stored by ID.
      */
     public static registry: Record<string, SyncEmitter> = {};
 
-    public static register(handler: SyncEmitter): void {
-        const { id } = handler;
-        this.registry[id] = handler;
+    /**
+     * Adds an emitter to the emitter regisitry.
```
karolkolodziej, line 32:
```suggestion
     * Adds an emitter to the emitter registry.
```

karolkolodziej (changes requested):

karolkolodziej:
Looks good but I'd like to know if we should show all of that.

IMO the `OptionsEntry` should be enough for now especially after we establish that there should not be any standalone components.

pawelfus (commented):
+1 to what Karol said: how much of it is necessary to build a dashboard? Which other part is necessary to build custom components and create sync between those and existing ones?

sebastianbochan (approved):

goransle:
> +1 to what Karol said: how much of it is necessary to build a dashboard? Which other part is necessary to build custom components and create sync between those and existing ones?

`OptionsEntry` should cover it it for most use cases.

pawelfus (approved):
Thanks!

bre1470 (approved):
Looks good. 👍

pawellysy (approved):

karolkolodziej (approved):

PR #18727, opened 2023-03-24
bugfix/17297-maps-remove-point

### LINKED ISSUE #17297: Maps, remove point: point is not removed from the series.points array
#### Expected behaviour
Let's discuss a map chart with 3 data points. After `point.remove()` I expect the `series.points` array to be length of 2.

#### Actual behaviour
After `point.remove()`, the removed point's reference is still in the `series.points` array. The `series.points.length` is still 3

https://jsfiddle.net/BlackLabel/2udxkv3j/

#### Product version
Highcharts Maps 10.0.0 (worked fine in 9.3.3)
Regression after https://github.com/highcharts/highcharts/commit/91f9740ba172a3e48caac5963b1f7b6a66a645df

@TorsteinHonsi was that intended? (I think no)

magdalena-gut:
Fixed #17297, incorrect `series.points` array's length after `point.remove()` in MapBubbleSeries.

### DIFF
--- a/samples/unit-tests/maps/mapbubble/demo.js
+++ b/samples/unit-tests/maps/mapbubble/demo.js
@@ -90,4 +90,14 @@ QUnit.test('MapBubble', function (assert) {
         `When hovering over mapbubble series with shared tooltip,
         there should be no errors in the console.`
     );
+
+    // Remove one point from the mapbubble series (#17297).
+    const origLength = chart.series[1].points.length;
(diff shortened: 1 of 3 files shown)


KacperMadej (approved):

hubertkozik (approved):
LGTM :)

TorsteinHonsi (approved):
Thanks!

PR #18746, opened 2023-03-28
db/cy-board-fix

goransle:
changed `cy.board` to use built-in cypress functionality, which includes automatic retries. Also added `cy.boardRendered`, which waits for components (and especially chart components) to have rendered.


sebastianbochan (approved):

pawellysy (approved):
LGTM

PR #18749, opened 2023-03-28
bugfix/18720-stack-y-float-issue

### LINKED ISSUE #18720: Incorrect value of stackY
#### Expected behaviour
As I understood stackTotal is a total value of the stack, stackY is a total value from the bottom point to the specific point of the stack. So I expect that if I have stack with values (10.5, 20, 15) I will get stackTotal = 45.5 and stackY for "20" value = 10.5 + 20 = 30.5. For "15" value I should get stackY = 10.5 + 20 + 15 = 45.5

#### Actual behaviour
So sometimes for some reason I get incorrect value of stackY. It has redundant numbers

I output stackTotal & stackY of the last point of the first series in console. I expect to get stackTotal = 691.6 and stackY = 691.6. But stackY = 691.5999999999999.
You could see that I'm using data values with only one digit after the decimal point, so value 691.5999999999999 is impossible.

```
    chart: {
        type: 'column',
        events: {
            load: function() {
                console.log(this.series[0].points[2].stackTotal, "stackTotal");
                console.log(this.series[0].points[2].stackY, "stackY");
            },
        },
    },
    xAxis: {
        categories: ['Arsenal', 'Chelsea', 'Liverpool']
    },
    plotOptions: {
        column: {
            stacking: 'normal',
        }
    },
    series: [
    {
        name: "Year 1990",
        data: [63.1, 72.7, 320.2],
    },
    {
        name: "Year 2000",
        data: [81.4, 84.1, 371.4],
    },
]
});
```

template: https://jsfiddle.net/fuL6gaxh/8/

#### Product version

"highcharts": "9.3.3",
"highcharts-angular": "3.0.0",

#### Affected browser(s)
Chrome

jakubSzuminski:
Fixed #18720, the `correctFloat` method was not applied to `stackY` which was throwing strange values like `12.00013...`  instead of `12`.

### DIFF
--- a/samples/unit-tests/series-column/stacking/demo.js
+++ b/samples/unit-tests/series-column/stacking/demo.js
@@ -1,27 +1,25 @@
 QUnit.test(
     'Negative stack with just one point should be also calculate (#4979)',
     function (assert) {
-        var chart = $('#container')
-                .highcharts({
-                    chart: {
-                        type: 'column'
+        const chart = Highcharts.chart('container', {
+                chart: {
+                    type: 'column'
+                },
+                plotOptions: {
+                    column: {
+                        borderWidth: 0,
+                        stacking: 'percent'
+                    }
+                },
+                series: [
+                    {
+                        data: [-10]
                     },
-                    plotOptions: {
-                        column: {
-                            borderWidth: 0,
-                            stacking: 'percent'
-                        }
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #18751, opened 2023-03-28
bugfix/16556-firefox-datepicker

### LINKED ISSUE #16556: data range in Firefox browser it is impossible to select month or year because of closing calendar popup
#### Expected behaviour
Datapicker doesn't close

#### Actual behaviour
Problem in Calendar picker in Firefox browser it is impossible to select month or year because of closing calendar popup.

Link to random example:
[Link](http://jsfiddle.net/5CRTa/1/)

#### Product version
highcharts: 9.3.0,
highcharts-react-official: 3.1.0,

#### Affected browser(s)
Firefox Browser

kamil-musialowski:
Fixed #16556, `datepicker` for `rangeSelector` input was not working in Firefox browsers.

### DIFF
--- a/ts/Stock/RangeSelector/RangeSelector.ts
+++ b/ts/Stock/RangeSelector/RangeSelector.ts
@@ -1022,17 +1022,21 @@ class RangeSelector {
         };
 
         // Hide away the input box
-        input.onblur = (): void => {
+        input.onblur = (e: any): void => {
             // update extermes only when inputs are active
             if (input === H.doc.activeElement) { // Only when focused
                 // Update also when no `change` event is triggered, like when
                 // clicking inside the SVG (#4710)
                 updateExtremes();
             }
-            // #10404 - move hide and blur outside focus
-            rangeSelector.hideInput(name);
-            rangeSelector.setInputValue(name);
-            input.blur(); // #4606
+            // Firefox calls the blur event on the datepicker click. Don't hide
+            // the input box if datepicker is clicked, #16556
+            if (!(H.isFirefox && e.target === e.explicitOriginalTarget)) {
+                // #10404 - move hide and blur outside focus
+                rangeSelector.hideInput(name);
+                rangeSelector.setInputValue(name);
+                input.blur(); // #4606
+            }
         };
 
         let keyDown = false;


ts/Stock/RangeSelector/RangeSelector.ts
```diff
@@ -1022,17 +1022,21 @@ class RangeSelector {
...
                 // clicking inside the SVG (#4710)
                 updateExtremes();
             }
-            // #10404 - move hide and blur outside focus
-            rangeSelector.hideInput(name);
-            rangeSelector.setInputValue(name);
-            input.blur(); // #4606
+            // Firefox calls the blur event on the datepicker click. Don't hide
+            // the input box if datepicker is clicked, #16556
+            if (!(H.isFirefox && e.target === e.explicitOriginalTarget)) {
```
pawelfus, line 1034: I'm not sure about this. Never heard of [`explicitOriginalTarget`](https://developer.mozilla.org/en-US/docs/Web/API/Event/explicitOriginalTarget) and it seems to be a non-standard feature. Is there any other way to check this? What do you think about this @TorsteinHonsi?

Is that expected behaviour for Firefox? Since other browsers don't do that, perhaps we should report it on bugzilla?

TorsteinHonsi: Agreed. It looks a bit weird and we don't like browser-specific checks. But if this is well tested across browsers and we don't have a better way of detecting it, I am okay with it.

kamil-musialowski: Thank you for your insights! I did more digging, but unfortunately, I couldn't came up with any other solution that the one proposed above.

But, there is a good news! This was actually reported as a bug in Firefox and just got fixed. Right now it is only available in beta versions, but I've tested it and everything works fine there.

> The issue is fixed on Firefox 112.0b9 and Firefox 113.0a1 (2023-04-03).
https://bugzilla.mozilla.org/show_bug.cgi?id=1812041

Based on their update schedule, the fix should be available for everyone with `112.0` release on 9th May, 2023.

In my opinion we can close this pull request and mark the original issue as 'Not a bug'.

pawelfus: Ohh, I see. Nice catch!

Could you please create a simple workaround for the clients and post it in #16556? Even though it's a bug in a browser, we can solve it

kamil-musialowski:
Issue fixed with 112.0 Firefox update.

PR #18758, opened 2023-03-29
bugfix/18584-vertical-line-local-storage-update

### LINKED ISSUE #18584: The location of the vertical labels is not updated in the local storage
#### Expected behavior
The location of the **vertical labels** after drag & drop should be updated in the local storage.

#### Actual behavior
In local storage, only the first location of vertical labels is kept.

Demo: https://www.highcharts.com/demo/stock/stock-tools-gui
Add the vertical label. Drag and drop the label. Save the chart.
Check the local storage - the vertical label was saved in the initial place when it was dropped first.
image

#### Product version
10.3.3

#### Affected browser(s)
N/A

jakubjanuchta:
Fixed #18584, vertical line annotation position was incorrect when saved to options after drag and drop.

### DIFF
--- a/samples/unit-tests/stock-tools/bindings/demo.js
+++ b/samples/unit-tests/stock-tools/bindings/demo.js
@@ -70,6 +70,46 @@ QUnit.test('Bindings general tests', function (assert) {
         );
     }
 
+    const verticalAnnotation = chart.addAnnotation({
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #18761, opened 2023-03-29
db/doclets-pluginHandler

sebastianbochan:


ts/Dashboards/PluginHandler.ts
```diff
@@ -39,21 +38,35 @@ namespace PluginHandler {
      * */
 
     export interface DashboardPlugin<T = (AnyRecord|undefined)> {
+        /** @internal */
         custom: T;
+        /**
+         * Maximal version of plugin that is compatible with dashboard
```
karolkolodziej, line 44: Do we even use it anywhere?

karolkolodziej (changes requested):

bre1470 (approved):
Looks good! 👍

ts/Dashboards/PluginHandler.ts
sebastianbochan: @bre1470 ?

bre1470: I do not understand the question either. It is not used by a plugin, because there is no discontinued plugin yet. The property is used though.

pawellysy (approved):

pawelfus (approved):
Thanks!

karolkolodziej (approved):

PR #18765, opened 2023-03-30
bugfix/14080-bubblelegend-prevent-default-click

### LINKED ISSUE #14080: BubbleLegends hides when clicking the last legend, even if I prevent the event from happening
I prevent the event on `legendItemClick` from happening and I manage externally when to hide the legends.

#### Expected behaviour
When clicking on the last legend, the BubbleLegend remains showing because I'm controlling externally the legends show/hide.

#### Actual behaviour
When clicking on the last legend, the BubbleLegend hides

https://jsfiddle.net/kzqhfa0r/6/
I'm preventing the Series 1 from being hidden

1. Click on Series 2 to hide it.
2. Click on Series 1 - It won't hide the Series, but the BubbleLegend will do

#### Product version
Reproduced using https://code.highcharts.com/highcharts-more.js

#### Affected browser(s)

At least Google Chrome Version 84.0.4147.105 (Official Build) (64-bit)

I suppose this happens because BubbleLegend is listening to `legendItemClick`, assuming the legend will toggle regardless.

https://github.com/highcharts/highcharts/blob/4d2a3690afdfe3b9dc4e62929b653086c7d3b25c/ts/Series/Bubble/BubbleLegend.ts#L1267

edit:

Actually the same happens with legend text is dimmed off(?), suppose is something similar.

I'm workarounding this issue by doing:
```js
legendItemClick(e: any) {
    e.preventDefault();

    // Workaround to prevent the bubbleLegend from being hidden:
    this.chart.legend.update({ bubbleLegend: { enabled: true }});
    this.chart.legend.bubbleLegend.visible = true;
    // for the label, I'm doing something like this, but somewhere else
    if (this.visible) {
        this.show();
     }
}
```

hubertkozik:
Fixed #14080, bubble legend didn't work correctly if the `legendItemClick` event for one series was prevented.

### DIFF
--- a/ts/Core/Foundation.ts
+++ b/ts/Core/Foundation.ts
@@ -83,7 +83,9 @@ namespace Foundation {
 
                     if (isFunction(event)) {
                         component.eventOptions[eventType] = event;
-                        addEvent(component, eventType, event);
+                        addEvent(component, eventType, event, {
+                            order: 0 // #14080 fire those events as firsts
+                        });
                     }
                 }
             }
--- a/samples/unit-tests/bubble-legend/layout/demo.js
+++ b/samples/unit-tests/bubble-legend/layout/demo.js
@@ -80,6 +80,48 @@ QUnit.test('Bubble legend ranges', function (assert) {
         true,
         'Bubble legend was properly disabled with the legend'
     );
+
+    chart.addSeries({
+        type: 'bubble',
+        data: [
+            [1, 4, 4],
+            [2, 5, 5]
+        ],
+        events: {
+            legendItemClick(e) {
+                e.preventDefault();
+            }
+        }
+    }, false);
+
+    chart.legend.update({
+        enabled: true,
+        floating: false
+    });
+
+    chart.series[1].legendItem.group.element.dispatchEvent(new Event('click'));
+    chart.series[0].legendItem.group.element.dispatchEvent(new Event('click'));
+
+    assert.ok(
+        true,
+        `There shouldn't be any error in the console, when one series
+        legendItemClick has prevented the event and we click both legend
+        items (#14080).`
+    );
+
+    assert.notOk(
+        isNaN(chart.legend.bubbleLegend.legendItem.labelHeight),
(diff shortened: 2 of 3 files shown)


ts/Core/Foundation.ts
```diff
@@ -83,7 +83,9 @@ namespace Foundation {
 
                     if (isFunction(event)) {
                         component.eventOptions[eventType] = event;
-                        addEvent(component, eventType, event);
+                        addEvent(component, eventType, event, {
+                            order: 0 // #14080 fire those events as firsts
```
TorsteinHonsi, line 87: I'm not sure I understand this correctly - but wouldn't it be better to set a higher order to the event in `BubbleLegendComposition.ts` instead of updating the core functionality?

hubertkozik: That was my first idea: change the order of that event in `BubbleLegendComposition.ts`, but unfortunately, it is not working. Events, which are added by users from chart configuration are added with order equal to `Infinity` - that's why, even if I add some order to an event in `BubbleLegendComposition.ts` it won't work, because there is no number bigger than `Infinity`... As a consequence of that, I decided to add `order: 0` to all "user's event functions" and it solves this problem. I think, that there may be more such problems, maybe not reported by users, or in the future, if someone would like to add an event, which will be connected also to events, which are added by users from chart configuration.

TorsteinHonsi: I assume we chose to set `Infinity` to default handlers so that any user-defined handlers could prevent the default, right? So now with this change, we still have Infinity on _default handlers_, but we have `order: 0` on any event that is defined in the _chart configuration_. Is that correctly understood? If so, I think it makes sense.

hubertkozik: Yes, that's true. 
Default handlers - `Infinity`to able to prevent. 
Events defined in the chart configuration -`order: 0` to fire them as firsts.

TorsteinHonsi (approved):

PR #18773, opened 2023-03-31
bugfix/18753-area-chart-legend-marker

### LINKED ISSUE #18753: Stacked Area Chart Legend is Not Correct
#### Expected behaviour
We would expect the stacked area chart legend to match the markers on the lines like it does in line chart. 

#### Actual behaviour
The color is correct in the legend, but the marker is always listed as a circle, putting it out of sync with the markers on the lines. This is even a problem in the stacked area chart demo on your site.

https://www.highcharts.com/demo/area-stacked

#### Product version
Current Version of Highcharts

#### Affected browser(s)
NA

jakubSzuminski:
Fixed #18753, area chart legend marker was circle for all series.


pawellysy (approved):
LGTM!

TorsteinHonsi:
Before going further with this, I think we should make it more flexible, but yet with less impact on the file size.

* The `legendSymbol` option should be an string, with possible values `lineMarker` and `rectangle`. 
* Now the functions in `LegendSymbol` should be renamed to match that. Then the `Series.drawLegendSymbol` function can be made much simpler by just running `LegendSymbol[this.options.legendSymbol].call(this, legend, item);`
* We now have a pluggable solution where implementers can add their own custom legend symbol function and apply it through the option. Make a demo for that (for example an area chart with an actual area for symbol?), and refer to it in `@samples` under the new `legendSymbol` option.

TorsteinHonsi (commented):
Awesome, implementing a new feature _and_ reducing file size 🎉 

I modified some types and the `drawLegendSymbol` function to avoid `any` cast. With this line

`LegendSymbol[this.options.legendSymbol || 'rectangle']`

jakubSzuminski:
Okay, great, I'm on it!

jakubSzuminski:
I think these were the only places where I could remove the default `legendSymbol: 'rectangle'`, because other series which have the `rectangle` as default inherit from Line, Scatter, or other series that has `lineMarker` as default.

TorsteinHonsi (approved):

PR #18774, opened 2023-03-31
db/compare-library-pr

sebastianbochan:


tools/gulptasks/pr-size-table.js
```diff
@@ -19,7 +19,9 @@ const files = argv.files ? argv.files.split(',') : [
     'modules/data.src.js',
     'modules/exporting.src.js',
     'modules/heatmap.src.js',
-    'modules/offline-exporting.src.js'
+    'modules/offline-exporting.src.js',
+    'dashboards.src.js',
+    'datagrid.src.js'
```
karolkolodziej, line 24: How about `dashboards-plugins`?

karolkolodziej (changes requested):

tools/gulptasks/pr-size-table.js
sebastianbochan: It is intentional because we also skip DataGrid or Highcharts plugins.

@pawelfus @goransle  do you think that we should add plugins to comparer too?

pawellysy: I'd go with comparing all files that we bundle. Apart from that, looks good!

pawellysy (approved):

tools/gulptasks/pr-size-table.js
sebastianbochan: Ok so let's me add them

pawelfus: I would add `dashboards-plugins` as a safety net - this module shouldn't load whole products (like `highcharts.js` or `datagrid.js`).

goransle (approved):

PR #18777, opened 2023-03-31
bugfix/18741-dragging-negative-stacked-bar

### LINKED ISSUE #18741: Incorrect guideBox behaviour when dragging negative bar in stacked column chart (or bar chart)
#### Expected behaviour
When we set the `series.bar.dragDrop.liveRedraw` option to `false`, the `guideBox` "shadow" should cover the area from "zero" up to the point being dragged.

#### Actual behaviour
When dragging a **negative** bar, the `guideBox` "shadow" is capped and has incorrect height: being equal to only the small dragged part of the whole stack, instead of the accumulated stack value up to that particular point.
The problem persists even when we "fake" the negative columns - adding them to a separate `reversed` yAxis, as shown on the second demo attached. In this case the guide box looks like a single trembling line.

IMPORTANT: I observed the same behaviour for both types of chart: `'column'` AND `'bar'`.

_NOTE: The `guideBox` works correctly for positive bars._

Example with negative stacked columns: https://jsfiddle.net/tge1a9mb/
Example with "fake" negative columns (with second `reversed` yAxis): https://jsfiddle.net/7j2x3dbo/2/ 

#### Product version
Using the **latest** version of both: `highcharts.js` and `modules/draggable-points.js`, see the live demos attached.
```
```

#### Affected browser(s)
Tested using Mozilla Firefox 111.0 (64-bit) for Manjaro Linux.

#### Solution(s)???
I guess that's a bug which needs to be fixed, but in the meantime: is there a way to workaround this behaviour?
For example, can I modify the [guide box](https://api.highcharts.com/highcharts/series.bar.dragDrop.guideBox) somehow to work correctly?

Any help appreciated.

Thanks!
Piotr

Vazonik:
Fixed #18741, incorrect guide box behaviour while dragging stacked bars.

### DIFF
--- a/samples/unit-tests/dragdrop/dynamic/demo.js
+++ b/samples/unit-tests/dragdrop/dynamic/demo.js
@@ -337,3 +337,131 @@ QUnit.test('Dragdrop with boost', function (assert) {
         'Dragdrop should work with boost (#11156).'
     );
 });
+
+QUnit.test('Dragdrop with stacked columns (#18741)', function (assert) {
+    const chart = Highcharts.chart('container', {
+            chart: {
+                type: 'column'
+            },
+            plotOptions: {
+                series: {
+                    stacking: 'normal',
+                    dragDrop: {
+                        draggableY: true,
+                        liveRedraw: false
+                    }
+                }
+            },
+            series: [{
+                data: [2, -3]
+            }, {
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #18779, opened 2023-04-02
Semgrep scan

MarkusBarstad:
Implemented a semgrep scan into the CircleCI pipeline

A few notes:
- It scans the ts-folder for the OWASP top 10 web app vulnerabilities
- Our `ts/`-folder passes the scan 100% as-is
- I have *not* tested this `config.yml` because I didn't want to mess with our CircleCI account and pipelines
- But I have used a totally separate throwaway CircleCI account to test and reason about CircleCI configs in general

I do have a decently sized scan-report done with way more rules (from running `semgrep` with the `--config auto` flag) in case we want more rules, and I have also learnt to make custom rules (`semgrep` has a decent tutorial) in case we ever need anything custom!

EDIT: 
Also feel free to say if we should run the scan somewhere else in the pipeline! I assume that the pipeline in CircleCI is separate from the pre-commit scans, but if not, should we have this scan there in stead?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -55,9 +55,6 @@ executors:
   node_image:
     docker:
       - image: cimg/node:16.3.0-browsers
-  semgrep_image:
-    docker:
-      - image: returntocorp/semgrep
 
 commands:
   generate_references_command:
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (approved):
Looks good to me 👍 

Adding Øystein and Christer for review.

oysteinmoseng (approved):
Looks great! If it breaks something we can just revert

PR #18780, opened 2023-04-03
db/nested-layouts-serialization

pawellysy:
Implemented nested layout serialization.

### DIFF
--- a/test/typescript-karma/Dashboards/Layout/tests.js
+++ b/test/typescript-karma/Dashboards/Layout/tests.js
@@ -182,3 +182,77 @@ test('Components and rows in layout with set height', function (assert) {
 
     layouts[0].style = {}
 })
+
+test('Nested layouts serialization.', function (assert) {
+    const container = setupContainer();
+
+    const chartComponentOptions = {
+        type: 'Highcharts',
+        chartOptions: {
+            type: 'line',
+            series: [{
+                name: 'Series from options',
+                data: [1, 2, 3, 4]
+            }],
+            chart: {
+                animation: false
+            }
+        }
+    };
+    const board = new Board(container.id, {
+        editMode: {
+            enabled: true,
+            contextMenu: {
+                enabled: true,
+                items: ['editMode']
+            }
+        },
+        gui: {
+            layouts: [{
+                id: 'layout-in-1',
+                rows: [{
+                    cells: [{
+                        id: 'dashboard-col-nolayout-0'
+                    }, {
+                        id: 'dashboard-col-layout-0',
+                        layout: {
+                            rows: [{
+                                cells: [{
+                                    id: 'dashboard-col-layout-1',
+                                }, {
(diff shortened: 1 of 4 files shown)


pawellysy:
TODO:

- [ ] Test

ts/Dashboards/Layout/Cell.ts
```diff
@@ -599,6 +648,10 @@ namespace Cell {
          * To create a nested layout, cell might contain an additional layouts.
          **/
         layout?: LayoutType.Options;
+        /**
+         * To create nested layout from JSON config.
```
karolkolodziej, line 652: Please mark as `internal` just in case

karolkolodziej (approved):

test/typescript-karma/Dashboards/Layout/tests.js
```diff
@@ -182,3 +182,77 @@ test('Components and rows in layout with set height', function (assert) {
...
+    const exportedRows = layoutToExport.rows;
+    const exportedRowsLength = layoutToExport.rows.length;
+    const exportedCellsLength = exportedRows[0].cells.length;
+    const numberOfMountedComponents = board.mountedComponents.length;
+    layoutToExport.exportLocal();
+    layoutToExport.destroy();
+    const importedLayout = board.importLayoutLocal(exportedLayoutId);
+
+
+    assert.equal(importedLayout.rows.length, exportedRowsLength, 'The imported layout has an equal number of rows as exported one.')
```
sebastianbochan, line 254: You can use a dedicated assertion for checking the length.

`importedLayout.rows.length.should('have.length', exportedRowsLength)`

ts/Dashboards/Layout/Layout.ts
```diff
@@ -282,6 +285,12 @@ class Layout extends GUIElement {
     public destroy(): void {
         const layout = this;
 
+        for (let i = layout.board.layouts.length - 1; i >= 0; i--) {
+            if (layout.board.layouts[i] === layout) {
```
sebastianbochan, line 289: Can we compare layouts by id? It will save us comparing the whole structure of layouts.

sebastianbochan (changes requested):

ts/Dashboards/Layout/Layout.ts
pawellysy: When you are comparing the objects in JS it is done by checking the reference of the object, so IMO it is safer to check if this is the same object.

test/typescript-karma/Dashboards/Layout/tests.js
pawellysy: it is not cypress but karma, there is no such assertion here, or am I wrong?

sebastianbochan: Sorry, my mistake. I was pretty sure that is cypress test.

sebastianbochan (approved):

PR #18783, opened 2023-04-03
bugfix/14398-networkgraph-datalabels-defer

### LINKED ISSUE #14398: dataLabels.animation.defer in networkgraph is not respected
#### Expected behaviour
`dataLabels.animation.defer` should defer rending dataLabels like for other series.

#### Actual behaviour
`dataLabels.animation.defer` is ignored, dataLabels are rendered right away.

https://jsfiddle.net/highcharts/18vrf9su/

*Internal note:*
By design, networkgraph disables all animations. Everything happens in `rAF`, where every node's position is calculated. In every simulation step, labels are checked for overlapping and rendered. Perhaps we can simply delay this overlapping logic by the `animation.defer` value.

jakubSzuminski:
Fixed #14398, `dataLabels.animation.defer` wasn't working for the `networkgraph` series type.

Here's a simplified **demo:** https://jsfiddle.net/BlackLabel/dwt25unp/

**Problem:** 
- `drawDataLabels()` method is being called out each frame in `Networkgraph` which is not the case for other series
- therefore, the `dataLabels.animation.defer` is overriden by the next call of `drawDataLabels()` for which `series.hasRendered = true` (during the second call the `dataLabels` are drawn so the `animation.defer` property has no effect)

**Solution (first draft):**
- call `drawDataLabels()` for the first time after `dataLabels.animation.defer` time (2.5 seconds by default) - the nodes should be in correct positions by this time,
- call `drawDataLabels()` for the second time after 150 miliseconds to allow the fade-in animation to take effect
- after first 2 calls of `drawDataLabels()` perform everything as usual

### DIFF
--- a/ts/Core/Series/DataLabel.ts
+++ b/ts/Core/Series/DataLabel.ts
@@ -458,9 +458,7 @@ namespace DataLabel {
         const series = this,
             chart = series.chart,
             seriesOptions = series.options,
-            hasRendered = series.type === 'networkgraph' ?
-                ((series as any).firstDlDraw || 0) :
-                (series.hasRendered || 0),
+            hasRendered = (series.hasRendered || 0),
             renderer = chart.renderer,
             { backgroundColor, plotBackgroundColor } = chart.options.chart,
             contrastColor = renderer.getContrast(
--- a/ts/Core/Series/DataLabel.ts
+++ b/ts/Core/Series/DataLabel.ts
@@ -474,8 +474,7 @@ namespace DataLabel {
     }
 
     /**
-     * Init the data labels
-     * with the correct animation
+     * Init the data labels with the correct animation
      * @private
      */
     function initDataLabels(
--- a/ts/Core/Series/DataLabel.ts
+++ b/ts/Core/Series/DataLabel.ts
@@ -91,7 +91,7 @@ declare module './SeriesLike' {
             alignTo: BBoxObject,
             isNew?: boolean
         ): void;
-        drawDataLabels(points?:Array<Point>): void;
+        drawDataLabels(points?:Array<Point>, shouldSkipOpacity?: boolean, shouldAnimate?: boolean): void;
         justifyDataLabel(
             dataLabel: SVGElement,
             options: DataLabelOptions,
@@ -453,7 +453,9 @@ namespace DataLabel {
      */
     function drawDataLabels(
         this: Series,
-        points: Array<Point> = this.points
+        points: Array<Point> = this.points,
+        shouldSkipOpacity: boolean = false,
+        shouldAnimate: boolean = false
     ): void {
         const series = this,
             chart = series.chart,
@@ -499,28 +501,56 @@ namespace DataLabel {
             (seriesDlOptions as any).enabled ||
             series._hasPointLabels
         ) {
-
             // Create a separate group for the data labels to avoid rotation
-            dataLabelsGroup = series.plotGroup(
-                'dataLabelsGroup',
-                'data-labels',
-                !hasRendered ? 'hidden' : 'inherit', // #5133, #10220
-                (seriesDlOptions as any).zIndex || 6
-            );
+            if (series.type !== 'networkgraph') {
+                dataLabelsGroup = series.plotGroup(
+                    'dataLabelsGroup',
+                    'data-labels',
+                    !hasRendered ? 'hidden' : 'inherit', // #5133, #10220
+                    (seriesDlOptions as any).zIndex || 6
+                );
 
-            dataLabelsGroup.attr({ opacity: +hasRendered }); // #3300
-            if (!hasRendered) {
-                const group = series.dataLabelsGroup;
-                if (group) {
-                    if (series.visible) { // #2597, #3023, #3024
-                        dataLabelsGroup.show();
+                dataLabelsGroup.attr({ opacity: +hasRendered }); // #3300
+
+                if (!hasRendered) {
+                    const group = series.dataLabelsGroup;
+                    if (group) {
+                        if (series.visible) { // #2597, #3023, #3024
+                            dataLabelsGroup.show();
+                        }
+                        (group[
+                            seriesOptions.animation ? 'animate' : 'attr'
+                        ] as any)(
+                            { opacity: 1 },
+                            animationConfig
+                        );
                     }
-                    (group[
-                        seriesOptions.animation ? 'animate' : 'attr'
-                    ] as any)(
-                        { opacity: 1 },
-                        animationConfig
-                    );
+                }
+            } else if (shouldAnimate) {
+                dataLabelsGroup = series.plotGroup(
+                    'dataLabelsGroup',
+                    'data-labels',
+                    !hasRendered ? 'hidden' : 'inherit', // #5133, #10220
+                    (seriesDlOptions as any).zIndex || 6
+                );
+
+                if (!shouldSkipOpacity) {
+                    dataLabelsGroup.attr({ opacity: +hasRendered }); // #3300
+                }
+
+                if (!hasRendered) {
+                    const group = series.dataLabelsGroup;
+                    if (group) {
+                        if (series.visible) { // #2597, #3023, #3024
+                            dataLabelsGroup.show();
+                        }
+                        (group[
+                            seriesOptions.animation ? 'animate' : 'attr'
+                        ] as any)(
+                            { opacity: 1 },
+                            animationConfig
+                        );
+                    }
                 }
             }
 
--- a/ts/Core/Series/DataLabel.ts
+++ b/ts/Core/Series/DataLabel.ts
@@ -459,18 +459,12 @@ namespace DataLabel {
      * @private
      */
     function initDataLabelsGroup(this: Series): SVGElement {
-        const series = this,
-            hasRendered = (series.hasRendered || 0),
-            seriesDlOptions = series.options.dataLabels;
-
-        const dataLabelsGroup = series.plotGroup(
+        return this.plotGroup(
             'dataLabelsGroup',
             'data-labels',
-            !hasRendered ? 'hidden' : 'inherit', // #5133, #10220
-            (seriesDlOptions as any).zIndex || 6
+            this.hasRendered ? 'inherit' : 'hidden', // #5133, #10220
+            (this.options.dataLabels as any).zIndex || 6
         );
-
-        return dataLabelsGroup;
     }
 
     /**
@@ -482,25 +476,20 @@ namespace DataLabel {
         animationConfig: Partial<AnimationOptions>
     ): SVGElement {
         const series = this,
-            hasRendered = (series.hasRendered || 0);
+            hasRendered = series.hasRendered || 0;
 
         // Create a separate group for the data labels to avoid rotation
-        const dataLabelsGroup = this.initDataLabelsGroup();
+        const dataLabelsGroup = this.initDataLabelsGroup()
+            .attr({ opacity: +hasRendered }); // #3300
 
-        dataLabelsGroup.attr({ opacity: +hasRendered }); // #3300
-
-        if (!hasRendered) {
-            const group = series.dataLabelsGroup;
-            if (group) {
-                if (series.visible) { // #2597, #3023, #3024
-                    dataLabelsGroup.show();
-                }
-                (group[
-                    series.options.animation ? 'animate' : 'attr'
-                ] as any)(
-                    { opacity: 1 },
-                    animationConfig
-                );
+        if (!hasRendered && dataLabelsGroup) {
+            if (series.visible) { // #2597, #3023, #3024
+                dataLabelsGroup.show();
+            }
+            if (series.options.animation) {
+                dataLabelsGroup.animate({ opacity: 1 }, animationConfig);
+            } else {
+                dataLabelsGroup.attr({ opacity: 1 });
             }
         }
 
--- a/ts/Core/Series/DataLabel.ts
+++ b/ts/Core/Series/DataLabel.ts
@@ -26,6 +26,7 @@ import type Series from './Series';
 import type SVGAttributes from '../Renderer/SVG/SVGAttributes';
 import type SVGElement from '../Renderer/SVG/SVGElement';
 import type SVGLabel from '../Renderer/SVG/SVGLabel';
+import type AnimationOptions from '../Animation/AnimationOptions';
 
 import A from '../Animation/AnimationUtilities.js';
 const { getDeferredAnimation } = A;
@@ -84,6 +85,10 @@ declare module './SeriesLike' {
         _hasPointLabels?: boolean;
         dataLabelsGroup?: SVGElement;
         dataLabelPositioners?: DataLabel.PositionersObject;
+        initDataLabelsGroup(): SVGElement;
+        initDataLabels(
+            animationConfig?: Partial<AnimationOptions>
+        ): SVGElement;
         alignDataLabel(
             point: Point,
             dataLabel: SVGElement,
@@ -91,7 +96,7 @@ declare module './SeriesLike' {
             alignTo: BBoxObject,
             isNew?: boolean
         ): void;
-        drawDataLabels(points?:Array<Point>, shouldSkipOpacity?: boolean, shouldAnimate?: boolean): void;
+        drawDataLabels(points?:Array<Point>): void;
         justifyDataLabel(
             dataLabel: SVGElement,
             options: DataLabelOptions,
@@ -439,6 +444,8 @@ namespace DataLabel {
         if (U.pushUnique(composedMembers, SeriesClass)) {
             const seriesProto = SeriesClass.prototype;
 
+            seriesProto.initDataLabelsGroup = initDataLabelsGroup;
+            seriesProto.initDataLabels = initDataLabels;
             seriesProto.alignDataLabel = alignDataLabel;
             seriesProto.drawDataLabels = drawDataLabels;
             seriesProto.justifyDataLabel = justifyDataLabel;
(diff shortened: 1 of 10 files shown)


ts/Core/Series/DataLabel.ts
```diff
@@ -458,7 +458,9 @@ namespace DataLabel {
         const series = this,
             chart = series.chart,
             seriesOptions = series.options,
-            hasRendered = series.hasRendered || 0,
+            hasRendered = series.type === 'networkgraph' ?
```
jakubSzuminski, line 461: I don't like this exception here, but I did not yet come up with a better solution (other than overriding the entire `drawDataLabels` function in `NetworkgraphSeries.ts`.

ts/Series/Networkgraph/NetworkgraphSeries.ts
```diff
@@ -297,9 +324,41 @@ class NetworkgraphSeries extends Series {
...
+            dlOptions.animation.defer
+        ) {
+            deferTime = dlOptions.animation.defer;
+        }
+
+        // @todo: fix casting here
+        // even if the defer is set by the user,
+        // networkgraph will use a custom defer method (with setTimeout)
+        // so we need to set it automatically to a very small number
+        ((dlOptions as any).animation as any).defer = 150;
```
jakubSzuminski, line 353: In the `Series.prototype.drawDataLabels()` there's a function `getDeferredAnimation` which sets the `animation = { defer, duration }`. If the defer is not set, the fade-in animation will not play. 

If the defer is set to a small value like `150` then the `animationConfig = { defer: 0, duration: 150 }`. Then, we achieve a desired result (_fade-in_ animation which takes 150 miliseconds with no defer time [defer is the `setTimeout`). 

I know hard-coding values like that is not good so I'll look into a better solution for this issue.

jakubSzuminski:
Also, another problem I see with this is when the value of `dataLabels.animation.defer` is too small or too large this solution doesn't perform well. Maybe, the fading animation of the datalabels would be worth sacrificing for - getting rid of this "freeze" time between the 1st and 2nd call of the `drawDataLabels()` would make a much higher range of `dataLabels.animation.defer` values work.

This problem stems from the fact that if `dataLabels.animation.defer = 2500` and the `Networkgraph` takes 5 seconds to reach the correct layout then the difference between position in dataLabels in `t = 2.5s` and `t=2.65s` can be huge (high speed of nodes).

Same goes for a too small value.

ts/Core/Series/DataLabel.ts
pawelfus: We can't leave it like this. We already override `drawDataLabels` in `NetworgraphSeries.ts` so it's not a problem

ts/Core/Series/SeriesDefaults.ts
```diff
@@ -1653,13 +1653,9 @@ const seriesDefaults: PlotOptionsOf<Series> = {
...
          * data label immediately. If set to `true` inherits the defer
          * time set in [plotOptions.series.animation](#plotOptions.series.animation).
-         * If set to a number, a defer time is specified in milliseconds.
-         *
-         * @sample highcharts/plotoptions/animation-defer
-         *         Set defer time
          *
          * @since     4.0.0
-         * @type      {boolean|number}
+         * @type      {boolean}
```
pawelfus, line 1658: Do you have a jsFiddle demo where it doesn't work?

ts/Core/Series/DataLabel.ts
pawelfus: Hint:
You can set temporarily `hasRendered` and then restore it:
```ts
drawDL() {
  const hasRendered = this.hasRendered;

  this.hasRendered = true;
  super();
  this.hasRendered = hasRendered;
}

```

ts/Series/Networkgraph/NetworkgraphSeries.ts
```diff
@@ -297,9 +324,41 @@ class NetworkgraphSeries extends Series {
...
+        const dlOptions = this.options.dataLabels;
+
+        // time from init() to the time when dataLabels show up #14398
+        // default set to 2.5 seconds
+        let deferTime = 2500;
+
+        // PROBLEM: if this time is too high
+        // then the dataLabels will never draw because the drawDataLabels()
+        // will not be called after the end of simulation
+        // but how to take care of this edge case?
```
pawelfus, line 338: In `ReingoldFruhtermanLayout` file you can find `step()` method. If you fire an event there, about a completed simulation, then you can listen to that event in Networgraph. I'm suggesting an event because it wouldn't be the first time we/users need this. We could make this an official API option, listing it among [other events](https://api.highcharts.com/highcharts/series.networkgraph.events).

ts/Series/Networkgraph/NetworkgraphSeries.ts
```diff
@@ -297,9 +324,41 @@ class NetworkgraphSeries extends Series {
...
+        // if dataLabels.animation.defer set by the user, use this value
+        if (dlOptions &&
+            dlOptions.animation &&
+            typeof dlOptions.animation !== 'boolean' &&
+            dlOptions.animation.defer
+        ) {
+            deferTime = dlOptions.animation.defer;
+        }
+
+        // @todo: fix casting here
```
pawelfus, line 349: In general, please avoid adding any's

ts/Series/Networkgraph/NetworkgraphSeries.ts
```diff
@@ -297,9 +324,41 @@ class NetworkgraphSeries extends Series {
...
+        // will not be called after the end of simulation
+        // but how to take care of this edge case?
+
+        // if dataLabels.animation.defer set by the user, use this value
+        if (dlOptions &&
+            dlOptions.animation &&
+            typeof dlOptions.animation !== 'boolean' &&
+            dlOptions.animation.defer
+        ) {
+            deferTime = dlOptions.animation.defer;
```
pawelfus, line 346: Could you try using `getDeferredAnimation()` method?

pawelfus (changes requested):
I don't see the fade-in effect on the first step: https://jsfiddle.net/BlackLabel/dwt25unp/2/

ts/Core/Series/SeriesDefaults.ts
jakubSzuminski: I think here: https://jsfiddle.net/BlackLabel/19cysqpx/

Plus if you take a look at the `interface DataLabelOptions` in `DataLabelOptions.d.ts` its type is `boolean`, not `boolean | number`.

ts/Core/Series/DataLabel.ts
jakubSzuminski: You're right! Thanks!

pawelfus (changes requested):
Overall looks good, but fade-in effect isn't applied: https://jsfiddle.net/BlackLabel/dwt25unp/11/

pawelfus (changes requested):
See comments inline

ts/Series/Networkgraph/NetworkgraphSeries.ts
```diff
@@ -208,26 +206,34 @@ class NetworkgraphSeries extends Series {
...
-            shouldAnimate = series.dlShouldAnimate;
+            dlOptions = series.options.dataLabels,
+            shouldAnimate = series.shouldAnimateDataLabels;
+
+        let fadeDuration = 500;
+        if (dlOptions &&
+            dlOptions.animation &&
+            dlOptions.animation.fadeDuration) {
+            fadeDuration = dlOptions.animation.fadeDuration;
+        }
```
TorsteinHonsi, line 217: From v11 we can use ES6 syntax, so these 5 lines can be written as 

`const fadeDuration = dlOptions?.animation?.fadeDuration ?? 500;`

jakubSzuminski:
**Demos:**
- with `animation: false` only `style.transition` applies (but could also be disabled): https://jsfiddle.net/BlackLabel/jgr5m24q/
- with both `animation.defer` and `style.transition` set: https://jsfiddle.net/BlackLabel/dyv8294c/
- with `animation.defer` longer than the time of simulation: https://jsfiddle.net/BlackLabel/54791dto/

ts/Series/PackedBubble/PackedBubbleDataLabelOptions.d.ts
```diff
@@ -47,6 +51,7 @@ export interface PackedBubbleDataLabelOptions extends DataLabelOptions {
     );
     parentNodeTextPath?: DataLabelTextPathOptions;
     textPath?: DataLabelTextPathOptions;
+    animation?: PackedBubbleDataLabelAnimationObject;
```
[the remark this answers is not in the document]
jakubSzuminski: You're right - I applied this interface because `animation.defer` for `PackedBubble` was in the docs and I had to define an interface for `animation` because of a TypeScript issue that arised (`NetworkgraphSeries` is not assignable to `BubbleSeries`) because I used an additional `fadeDuration` property. 

Now, I've removed this property (as we'll use `style.transition`) so I could also get rid of this interface - the `PackedBubble` series is almost untouched.

However, the `animation.defer` didn't work & doesn't seem to be working now (for the same reason it did not work for `Networkgraph`). Should I implement a similar logic for the `PackedBubble` as I did for `Networkgraph`?

ts/Series/Networkgraph/NetworkgraphSeriesOptions.d.ts
```diff
@@ -19,6 +19,7 @@
 import type AnimationOptions from '../../Core/Animation/AnimationOptions';
 import type ColorType from '../../Core/Color/ColorType';
 import type DashStyleValue from '../../Core/Renderer/DashStyleValue';
+import { EventCallback } from '../../Core/Callback';
```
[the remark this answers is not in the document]
jakubSzuminski: You're right, fixed it.

ts/Series/Networkgraph/NetworkgraphSeries.ts
```diff
@@ -297,9 +379,29 @@ class NetworkgraphSeries extends Series {
...
     ): NetworkgraphSeries {
-
         super.init(chart, options);
 
+        const dlOptions = this.options.dataLabels;
+
+        // time from init() to the time when the drawDataLabels()
+        // is called for the first time, #14398
+        // default set to 2.0 seconds
+        let deferTime = 2000;
```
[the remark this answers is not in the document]
jakubSzuminski: You're right, that's better. I'm now using this `animObject`.

ts/Series/Networkgraph/NetworkgraphSeries.ts
```diff
@@ -193,23 +200,100 @@ class NetworkgraphSeries extends Series {
...
+    public initDataLabels(this: NetworkgraphSeries): SVGElement {
+        const series = this,
+            dlOptions = series.options.dataLabels,
+            shouldInitialize = series.shouldInitializeDataLabels;
+
+        // if it's the first drawDataLabels() call
+        // then we'll create the SVG group for the first time
+        // & setup the animation from opacity 0 to opacity 1
+        // which lasts based on style.transition property
+        if (shouldInitialize) {
```
pawelfus, line 218: Following the idea of removing `shouldInitializeDataLabels` property: could you check if we can simply check if `series.dataLabelsGroup` exist? Then simple:
```ts
if (!group) {
  // create group;
  // set opacity;
  // return group;
} else {
  // do something
  // return group;
}
```

Should do the trick

ts/Series/Networkgraph/NetworkgraphSeries.ts
```diff
@@ -314,6 +406,12 @@ class NetworkgraphSeries extends Series {
             });
         });
 
+        addEvent(this, 'afterSimulation', function (): void {
+            this.deferDataLabels = false;
+            this.drawDataLabels(); // first time initialization with opacity = 0
+            this.drawDataLabels(); // second time for setting the opacity = 1
```
pawelfus, line 412: I don't think you need both, could you double-check it? I mean networkgraph has one group for links and nodes, so `drawDataLabels` is called "twice" anyway

ts/Series/Networkgraph/NetworkgraphSeries.ts
```diff
@@ -193,23 +200,100 @@ class NetworkgraphSeries extends Series {
...
+            }
+
+            // initialize the opacity of the group to 0 (start of animation)
+            dataLabelsGroup.attr({ opacity: 0 });
+
+            const group = series.dataLabelsGroup;
+            if (group) {
+                if (series.visible) { // #2597, #3023, #3024
+                    dataLabelsGroup.show();
+                }
```
pawelfus, line 238: Could be simplified to `if (series.visible ) { ... }` (the group and variable is available, right?)

ts/Series/Networkgraph/NetworkgraphSeries.ts
```diff
@@ -193,23 +200,100 @@ class NetworkgraphSeries extends Series {
...
+        if (dlOptions && dlOptions.linkTextPath) {
+            // if linkTextPath is set, render link labels with linkTextPath
+            dlOptions.textPath = dlOptions.linkTextPath;
+        }
+
         Series.prototype.drawDataLabels.call(this, this.data);
 
-        (this.options.dataLabels as any).textPath = textPath;
+        // go back to textPath for nodes
+        if (dlOptions && dlOptions.textPath) {
```
pawelfus, line 288: Could be simplified as above

ts/Series/Networkgraph/NetworkgraphSeries.ts
```diff
@@ -193,23 +200,100 @@ class NetworkgraphSeries extends Series {
...
+        // we defer drawing the dataLabels
+        // until dataLabels.animation.defer time passes
+        if (this.deferDataLabels) {
+            return;
+        }
+
+        const dlOptions = this.options.dataLabels;
+
+        let textPath;
+        if (dlOptions && dlOptions.textPath) {
```
pawelfus, line 272: Could be simplified as below

ts/Core/Series/DataLabel.ts
```diff
@@ -447,6 +454,60 @@ namespace DataLabel {
...
+            !hasRendered ? 'hidden' : 'inherit', // #5133, #10220
+            (seriesDlOptions as any).zIndex || 6
+        );
+
+        return dataLabelsGroup;
+    }
+
+    /**
+     * Init the data labels
+     * with the correct animation
```
pawelfus, line 478: Feel free to use 80 chars in a line :)

ts/Series/Networkgraph/NetworkgraphSeries.ts
jakubSzuminski: The problem is this:
- when the layout is stable the `drawDataLabels` method is not called each frame
- if the `animation.defer > simulationTime` then the `drawDataLabels` will never be called (that's why we force it here)
- the first call of `drawDataLabels` (therefore also the first call of `initDataLabels`) initializes the SVG group with an `opacity = 0`
- if we leave just one call of `drawDataLabels` in the `afterSimulation` event then they will always have `opacity = 0`
- during the second call of `initDataLabels` we see that the `dataLabels` have already been initialized so we set the `opacity = 1` and the animation starts

If I leave only one call of `drawDataLabels` the labels will always have opacity of `0`. That's why I had to add the second call.

However, I think now that instead of calling the `drawDataLabels` for the second time we could just try to set opacity of the SVG group with `this.dataLabelsGroup.attr({ opacity: 1 })` and that should work.

What do you think?

ts/Series/Networkgraph/NetworkgraphSeries.ts
```diff
@@ -193,23 +200,100 @@ class NetworkgraphSeries extends Series {
...
-        (this.options.dataLabels as any).textPath = textPath;
+        // go back to textPath for nodes
+        if (dlOptions && dlOptions.textPath) {
+            dlOptions.textPath = textPath;
+        }
+
+        // We should not initiate the animation anymore
+        // after the first call of drawDataLabels() method
+        if (this.shouldInitializeDataLabels) {
+            this.shouldInitializeDataLabels = false;
```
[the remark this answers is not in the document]
jakubSzuminski: You're right, it is redundant - we can recognize the first call of `initDataLabels` by checking whether the `series.dataLabelsGroup` exists. I've changed that.

ts/Series/Networkgraph/NetworkgraphSeries.ts
pawelfus: Yes, I saw the comments.
What I wanted to say is `drawDataLabels` in Networkgraph calls `Series.drawDataLabels` twice (at this moment lines: 264 and 272 in `NetworkgraphSeries.ts`), which calls `initDataLabels` twice, thus no need to call it twice here.

ts/Core/Series/SeriesDefaults.ts
pawelfus: FYI @TorsteinHonsi

pawelfus (approved):
Thanks!

ts/Series/PackedBubble/PackedBubbleDataLabelOptions.d.ts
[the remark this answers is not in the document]
TorsteinHonsi: Yes please, those two series types should work in the same way.

jakubSzuminski:
**Demo** for _PackedBubble_: https://jsfiddle.net/BlackLabel/kgr92L87/

I tried to re-use the logic from Networkgraph with `NetworkgraphSeries.prototype.initDataLabels`, but I finally just rewrote it into `PackedBubbleSeries.ts` as the `NetworkgraphSeries` requires importing the additional `networkgraph.js` module so these functions could not be used directly (from what I know).

ts/Series/Networkgraph/NetworkgraphSeriesDefaults.ts
```diff
@@ -216,6 +216,11 @@ const NetworkgraphSeriesDefaults: NetworkgraphSeriesOptions = {
         },
         style: {
             transition: 'opacity 2000ms'
+        },
+
+        defer: true,
+        animation: {
+            defer: 2000
```
TorsteinHonsi, line 223: I am thinking 2 seconds is a bit much? On the [main demo](http://localhost:3030/samples/#view/highcharts/demo/network-graph), I would say 1000 works well.

ts/Series/PackedBubble/PackedBubbleSeries.ts
```diff
@@ -479,12 +484,51 @@ class PackedBubbleSeries extends BubbleSeries {
         seriesProto.destroy.apply(this, arguments as any);
     }
 
+    /**
+     * PackedBubble needs a custom initDataLabels function
+     * to initiate the animation in a different way
+     * than for other series (drawDataLabels is called each frame)
+     * @private
+     */
+    public initDataLabels(this: PackedBubbleSeries): SVGElement {
```
TorsteinHonsi, line 493: As far as I can see this implementation is identical to the one in NetworkgraphSeries.ts, which means we're violation the [DRY principle](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself). Sooner or later someone will fix something in one place, not knowing about the other. So we should put this function in a module and import it into both series types. I am not sure if we have a fitting module for sharing stuff between packed bubble and network graph, otherwise we can create a new one. @pawelfus please advice.

TorsteinHonsi (changes requested):
Thanks, it's working really well now I think! Some refactoring then we're good to go.

ts/Series/PackedBubble/PackedBubbleSeries.ts
pawelfus: We have files that are being shared between those series (layout-related ones), but `initDataLabels` doesn't suit there. I think a simple composition similar to `DragNodesComposition.ts` or `GraphLayoutComposition.ts` should be enough. Unless we can use something that fits better (like 2010' mixins)?

pawelfus: Or something similar to `DrawPointUtilities.ts`? 🤔

TorsteinHonsi: Yes, I don't think we need to make a composition out of it, just some module that shares that function. It can even export _only_ that function (as `export default`). In the future, if we identify more shared needs, we can just refactor it.

ts/Series/RAFDeferDataLabels.ts
```diff
@@ -0,0 +1,87 @@
...
+import type PackedBubbleSeries from './PackedBubble/PackedBubbleSeries';
+import type SVGElement from '../Core/Renderer/SVG/SVGElement';
+
+import U from '../Core/Utilities.js';
+const { syncTimeout } = U;
+
+import A from '../Core/Animation/AnimationUtilities.js';
+const { animObject } = A;
+
+type RAFSeries = (NetworkgraphSeries | PackedBubbleSeries);
```
TorsteinHonsi, line 24: What is RAF? 🤔

jakubSzuminski: Request Animation Frame. I tried to find a name specific to simulated series, but struggled to find anything better.

TorsteinHonsi: SimulationSeries? These series run a physical simulation instead of plain animation.

jakubSzuminski: Okay, I'll rename the type to `type SimulationSeries = (NetworkgraphSeries | PackedBubbleSeries)`.

Should the name of the file stay the same? Or should it also include _SimulationSeries_?

TorsteinHonsi: It should also include SimulationSeries. I think `SimulationSeriesUtilities.ts` would be a descriptive name.

TorsteinHonsi (approved):

PR #18784, opened 2023-04-04
db/datagrid-new-api-options

karolkolodziej:
- [x] show/hide
- [x] header format
- [x] cell format
- [x] test
- [x] docklet
- [x] editing cell format


ts/DataGrid/DataGrid.ts
```diff
@@ -397,7 +400,25 @@ class DataGrid {
...
+     * column `show` option.
+     * @internal
+     */
+    private getColumnsToDisplay(): Array<string> {
+        const columnsOptions = this.options.columns,
+            tableColumns = this.dataTable.getColumnNames();
+
+        return tableColumns.filter(function (columnName): boolean {
+            const column = columnsOptions[columnName];
+            if (column && defined(column.show)) {
```
sebastianbochan, line 413: Do we need defined here?

ts/DataGrid/DataGrid.ts
```diff
@@ -397,7 +400,25 @@ class DataGrid {
...
-     * Determs whether a column is editable or not.
+     * Check which columns should be displayed based on the individual
+     * column `show` option.
+     * @internal
+     */
+    private getColumnsToDisplay(): Array<string> {
+        const columnsOptions = this.options.columns,
+            tableColumns = this.dataTable.getColumnNames();
+
+        return tableColumns.filter(function (columnName): boolean {
```
sebastianbochan, line 411: AFAIK we should not use `.filter`, `.map` etc functions due to performance reasons.

ts/DataGrid/DataGrid.ts
```diff
@@ -517,13 +538,15 @@ class DataGrid {
 
 
             for (let k = 0; k < columnsInPresentationOrder.length; k++) {
```
sebastianbochan, line 540: Would be better to replace it with `let k = 0; kEnd = columnsInPresentationOrder.length; k < kEnd` - performance reasons.

ts/DataGrid/DataGrid.ts
```diff
@@ -784,6 +818,60 @@ class DataGrid {
...
+     * @internal
+     *
+     * @param columnName
+     * Column name to format.
+     */
+    private formatHeaderCell(columnName: string): string {
+        const options = this.options,
+            columnOptions = options.columns[columnName],
+            headerFormat = columnOptions && columnOptions.headerFormat,
+            ctx = { text: columnName };
```
sebastianbochan, line 833: The variable is declared for use it once, I suggest to remove it from here.

sebastianbochan (changes requested):

pawellysy (approved):

ts/DataGrid/DataGrid.ts
karolkolodziej: Yes, we need it ;) 
False is a valid value here, thus we have to use `define`.

ts/DataGrid/DataGrid.ts
karolkolodziej: @bre1470 ?

goransle (approved):
Approved pending other requested changes

PR #18791, opened 2023-04-05
bugfix/18643-indicator-error

### LINKED ISSUE #18643: range selector with data grouping breaks chart when volume based indicator is used
#### Expected behaviour
On Highstocks chart user should be able to use range selector buttons with dataGrouping on a chart which includes volume and volume based indicators (e.g. OBV).

#### Actual behaviour
Volume columns and indicator disappear after clicking on range selector button. 
In console the following error is visible: _Uncaught Error: Series aapl-volume not found! Check `volumeSeriesID`._

I guess this unresolved error: https://github.com/highcharts/highcharts/issues/18409 could be caused by the same problem.

https://jsfiddle.net/72dghy9z/2/

Steps to reproduce:
1. Open the demo and add OBV indicator
1. Click on any range selector button

#### Product version
Highstock - latest version (v10.3.3 at the time of reporting).
The problem seems to exist in v9.2.1 but v8.2.2 seems to work fine.

#### Affected browser(s)
Chrome, Firefox

jakubjanuchta:
Fixed #18643, indicator threw `volumeSeriesID` error on range selector click.

### DIFF
--- a/ts/Stock/Indicators/SMA/SMAIndicator.ts
+++ b/ts/Stock/Indicators/SMA/SMAIndicator.ts
@@ -300,66 +300,66 @@ class SMAIndicator extends LineSeries {
             function ({ isUpdating }: AnyRecord): void {
                 // #18643 indicator shouldn't recalculate
                 // values while series updating.
-                if (!isUpdating) {
-                    const hasEvents = !!indicator.dataEventsToUnbind.length;
+                if (isUpdating) {
+                    return;
+                }
+                const hasEvents = !!indicator.dataEventsToUnbind.length;
 
-                    if (indicator.linkedParent) {
-                        if (!hasEvents) {
-                            // No matter which indicator, always recalculate
-                            // after updating the data.
+                if (indicator.linkedParent) {
+                    if (!hasEvents) {
+                        // No matter which indicator, always recalculate after
+                        // updating the data.
+                        indicator.dataEventsToUnbind.push(
+                            addEvent(
+                                indicator.linkedParent,
+                                'updatedData',
+                                function (): void {
+                                    indicator.recalculateValues();
+                                }
+                            )
+                        );
+
+                        // Some indicators (like VBP) requires an additional
+                        // event (afterSetExtremes) to properly show the data.
+                        if (indicator.calculateOn.xAxis) {
                             indicator.dataEventsToUnbind.push(
                                 addEvent(
-                                    indicator.linkedParent,
-                                    'updatedData',
+                                    indicator.linkedParent.xAxis,
+                                    indicator.calculateOn.xAxis,
                                     function (): void {
                                         indicator.recalculateValues();
                                     }
                                 )
                             );
-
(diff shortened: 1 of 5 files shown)


ts/Stock/Indicators/SMA/SMAIndicator.ts
```diff
@@ -297,64 +297,69 @@ class SMAIndicator extends LineSeries {
...
-                            )
-                        );
-
-                        // Some indicators (like VBP) requires an additional
-                        // event (afterSetExtremes) to properly show the data.
-                        if (indicator.calculateOn.xAxis) {
+            function ({ isUpdating }: AnyRecord): void {
+                // #18643 indicator shouldn't recalculate
+                // values while series updating.
+                if (!isUpdating) {
```
pawelfus, line 303: Instead of:
```ts
if () {
  if () {
    // code
  }
}
```
You can do:
```ts
if () {
  return;
}

if () {
  // code
}
```

pawelfus (changes requested):

pawelfus:
A short explanation to @TorsteinHonsi :

Tech Indicators are calculated after series are linked to each other (thanks to that, we don't care about the order of series in `series` array).

When you configure RangeSelector to set `dataGrouping` on click, then under the hood you call `series.update()`. Which means `series.remove` + `series.init`. And both of these methods cause linking series. After linking in `remove()`, the linked parent series is missing, thus an error in the console.

To solve it, we pass on `keepEvents`/`isUpdating` flag from `series.update()` method to `recalculateValues()`, so when linking happens, we simply skip calculating values.

Note: Updating DG from RangeSelector is just one scenario, others can happen too (simple example: https://jsfiddle.net/BlackLabel/8kjd5g71/ = two consequent updates, indicator without a redraw)

ts/Stock/Indicators/SMA/SMAIndicator.ts
```diff
@@ -297,7 +297,12 @@ class SMAIndicator extends LineSeries {
         const linkedSeriesUnbiner = addEvent(
             Chart,
             'afterLinkSeries',
-            function (): void {
+            function ({ isUpdating }: AnyRecord): void {
```
pawelfus, line 300: Probably type:
```ts
function (event: { isUpdating: boolean })
```
 would be better. Or perhaps
 
```ts
function (event: Event&{ isUpdating: boolean })
```

Or shall we create a separate interface for it? 🤔 What do you think @bre1470

pawelfus (approved):
Thanks!

Extra request: VbP also has a bit different solution to overcome this issue. Could you try removing logic in `VbP.init` and see if that helps?

jakubjanuchta:
> Thanks!
> 
> Extra request: VbP also has a bit different solution to overcome this issue. Could you try removing logic in `VbP.init` and see if that helps?

Thanks @pawelfus!
As discussed, it cannot be checked because of another issue related to VBP series

PR #18792, opened 2023-04-05
bugfix/17007-vbp-indicator-redraw-issue

### LINKED ISSUE #17007: VBP indicator doesn't recalculate values on redraw
This is a continuation of https://github.com/highcharts/highcharts/issues/16397

The above ticket solved the issue when the VBP indicator didn't recalculate on data update. But there's still issue when we not only update data, but update params or set `series.visible(true)`.

https://jsfiddle.net/BlackLabel/rpfxv16L/
All updates should work without any console errors. For now, only the third update (`series.update({ visible: false })`) is working fine.

#### Product version
9.3.3

Workaround: update options first without redraw, then recalculate data manually, and then redraw the series.
Demo with a workaround: https://jsfiddle.net/BlackLabel/23ojL10w/

FYI @karolkolodziej

jakubSzuminski:
Fixed #17007, VBP indicator did not update correctly.

**Problem:**
_the problem was that the values, for example_ `volumeDataArray`, _for the indicator series were not updated after_ `series.update`

**Solution:**
_after each update, trigger_ `series.recalculateValues()`

### DIFF
--- a/samples/unit-tests/indicator-volume-by-price/recalculations/demo.js
+++ b/samples/unit-tests/indicator-volume-by-price/recalculations/demo.js
@@ -161,9 +161,9 @@ QUnit.test('Test algorithm on data updates.', function (assert) {
     });
 
     assert.strictEqual(
-        VBPIndicator.options.params.ranges,
+        VBPIndicator.volumeDataArray.length,
         6,
-        'We should be able to update VBP params with update(), #17007.'
+        'VBP params should be updated after update(), #17007.'
     );
 
     VBPIndicator.update({
--- a/samples/unit-tests/indicator-volume-by-price/recalculations/demo.js
+++ b/samples/unit-tests/indicator-volume-by-price/recalculations/demo.js
@@ -154,6 +154,24 @@ QUnit.test('Test algorithm on data updates.', function (assert) {
 
     // End of #16277 testing, start other tests
 
+    VBPIndicator.update({
+        params: {
+            ranges: 6
+        }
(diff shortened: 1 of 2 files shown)


samples/unit-tests/indicator-volume-by-price/recalculations/demo.js
```diff
@@ -154,6 +154,24 @@ QUnit.test('Test algorithm on data updates.', function (assert) {
...
+    VBPIndicator.update({
+        params: {
+            ranges: 6
+        }
+    });
+
+    assert.strictEqual(
+        VBPIndicator.options.params.ranges,
+        6,
+        'We should be able to update VBP params with update(), #17007.'
```
kamil-musialowski, line 166: The test looks absolutely fine, just two minor suggestions:

- Maybe we also should check for the `volumeSeriesData.length` since it was causing the problem?
`volumeSeriesData.length === VBPIndicator.options.params.ranges` should do the job.

- I would suggest changing the description to something more consistent with other tests, i.e:
"VBP params should be updated after update(), #17007"

kamil-musialowski (changes requested):
Other than a one small comment regarding the test, everything looks good to me!

pawelfus (changes requested):
Please, investigate the other solution first (without adding a new event)

pawelfus (approved):
Nice, thanks!

TorsteinHonsi (approved):
Thanks!

Issue #17261, opened 2022-05-06
Maps: orthographic projection dataLabels placement


KacperMadej:
#### Behaviour
dataLabels for map features that are not visible are placed on the world edge and other labels sometimes have a strange offset - e.g. see Norway in the below demo after some globe rotations.

#### Live demo with steps to reproduce
https://jsfiddle.net/BlackLabel/ofwqnt9b/

#### Product version
Highmaps v9.3+

KacperMadej:
Resolved by https://github.com/highcharts/highcharts/pull/18795

PR #18795, opened 2023-04-06
bugfix/18004-map-datalabels-misplaced

### LINKED ISSUE #18004: Maps: dataLabels placement when the shape loops around the world
#### Expected behaviour
Placement is to be based on the latitude and longitude corresponding to the center of the region on the map.
Get the lat/long out of the `hc-middle-x`/`hc-middle-y` and use that for the dataLabels instead of the current behavior.

#### Actual behaviour
Placement is based on the bounding box - which doesn't work well for some projections or projection options.

Asia is wrong in the default demo - https://jsfiddle.net/gh/get/jquery/1.11.0/highslide-software/highcharts.com/tree/master/samples/mapdata/custom/world-continents

After some changes, see Europe, Asia in the wrong places - https://jsfiddle.net/BlackLabel/5w61mt4y/

#### Product version
Highcharts Maps v9.3+

hubertkozik:
Fixed #18004 and #17261, map data labels were misplaced for some shapes.

----
This PR is related to PR in map-collection: https://github.com/highcharts/map-collection/pull/32


ts/Series/Map/MapSeries.ts
```diff
@@ -532,6 +532,17 @@ class MapSeries extends ScatterSeries {
...
+        this.points.forEach((point): void => {
+            if (point.dataLabel) {
+                if (
+                    point.projectedPath &&
+                    !point.projectedPath.length
+                ) {
+                    point.dataLabel.hide();
+                }
+            }
+        });
```
TorsteinHonsi, line 545: I'm wondering about this code - what if we didn't store `point.projectedPath` in the first place when it is `[]`? Would that stop the data label from being added so that we didn't have to loop over the points and hide conditionally?

hubertkozik: I have added this condition to hide the data label when the point is not visible in the current range of the map view (sometimes e.g. Brazil data label was shown, but the shape of Brazil wasn't visible due to the current earth's rotation [Orthographic Projection]). Unfortunately, deleting the `point.projectedPath` when it is empty won't stop drawing the data label, but I came up with another idea.

We can use the loop through points inside MapSeries `translate` method, check if `projectedPath` exists and set a proper visible flag. Thanks to that, the data label won't be shown due to the basic logic for drawing data labels and we won't have additional loops.

TorsteinHonsi (approved):

PR #18800, opened 2023-04-07
demo/axis-position-properties

jedrzejruta:
Added API demo to `xAxis.top`, `xAxis.left`, `xAxis.height` and `xAxis.width`.


jedrzejruta:
Sure! I’ll get to it as soon as I’m back in the office on Monday.

PR #18801, opened 2023-04-07
bugfix/18768-missing-legend-page

### LINKED ISSUE #18768: Legend is missing the last page when legend.maxHeight is smaller than 43
#### Expected behaviour

Legend has the 3rd page containing series 9 for small legend.maxHeight. (in the attached JSFiddle case)

#### Actual behaviour

Legend does not have the 3rd page when legend.maxHeight is less than 43.

https://jsfiddle.net/386czx9f/1/

#### Product version

Highcharts v10.3.3 (to at least v8.1.1)

#### Affected browser(s)
Chrome, Safari, Firefox, Edge

Vazonik:
Fixed #18768, missing the last item in a legend when it was alone on a page.

### DIFF
--- a/samples/unit-tests/legend/navigation/demo.js
+++ b/samples/unit-tests/legend/navigation/demo.js
@@ -19,5 +19,24 @@ QUnit.test(
             4,
             'There should be enough pages to fully fit all elements (#13683)'
         );
+
+        chart.update({
+            chart: {
+                width: 400
+            },
+            legend: {
+                maxHeight: 40,
+                itemStyle: {
+                    fontSize: '12px'
+                }
+            }
+        }, false);
+        chart.addSeries({});
+
+        assert.strictEqual(
+            chart.legend.pages.length,
+            3,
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):

TorsteinHonsi:
@Vazonik Since merging this, `npm test` fails on the master. Can you check it out?

Vazonik:
I ran tests on Windows (node v18.12.1) and two Macs, with reinstalled modules and in all cases they pass. Does the problem still occur? @TorsteinHonsi

TorsteinHonsi:
You're probably running your tests in ChromeHeadless. The test issue can be reproduced by opening http://localhost:3030/samples/#test/unit-tests/legend/navigation in Firefox on the Mac. With a disclaimer that it may depend on how the system handles fonts.

Vazonik:
@TorsteinHonsi That's right, thanks! Fixed it here: #18820.

PR #18803, opened 2023-04-11
Db/ci visual order

goransle:


karolkolodziej:
This https://github.com/highcharts/highcharts/pull/18806 should fix the visual comparison.

.circleci/config.yml
```diff
@@ -461,6 +450,12 @@ jobs:
           name: Comment on PR and upload visual test images if any produced
           command: echo ${CIRCLE_PULL_REQUEST##*/} | xargs -I{} npx gulp update-pr-testresults-cypress --fail-silently --user circleci-mu --pr {}
           when: always
+      - run:
```
sebastianbochan, line 453: In this case I would suggest to move

```
- run:
          name: Write file sizes at master
          command: npx gulp write-file-sizes --filename master.json
```
above moved lines instead of leaving them on top.

@goransle what do you think?

sebastianbochan (changes requested):

.circleci/config.yml
goransle: as in doing another round of checkouts at the end of the job?

sebastianbochan: I mean, call "writes size to master.json" before the file comparison instead of running write first, then call some other things and then file comparison.

pawellysy (approved):
LGTM!

karolkolodziej (approved):

goransle:
@sebastianbochan ok to merge?

PR #18787, opened 2023-04-04
DataConnector with table options

bre1470:

PR #18810, opened 2023-04-11
DataPool demo

bre1470:
Builds on #18787


goransle (approved):
Looking good

karolkolodziej (approved):

pawellysy (approved):
LGTM!

TorsteinHonsi (approved):
Looks good!

samples/highcharts/data-tools/datapool-html/demo.css
```diff
@@ -0,0 +1,65 @@
+:root {
+    --color-background: #fff;
+    --color-foreground: #000;
```
sebastianbochan, line 3: IMO would be better to use --color-text, because of using for param `color:` that is applied for the font.

sebastianbochan (changes requested):

PR #18814, opened 2023-04-12
Adjustments for stricter scans

MarkusBarstad:
Adjusted code in the repo to be in accordance with `semgreps` default scan.

~~Beware: I have so far trusted the unit-tests as verification of this branch functioning.~~

- [ ] javascript.browser.security.insecure-document-method.insecure-document-method
- [ ] javascript.lang.security.audit.detect-non-literal-regexp.detect-non-literal-regexp
~~- [ ] javascript.lang.security.audit.incomplete-sanitization.incomplete-sanitization~~ (only oldie flagged)
- [ ] javascript.lang.security.audit.prototype-pollution.prototype-pollution-loop.prototype-pollution-loop

~~Edit: I realize I might have been a tad crude in using search/replace as I review my initial commit x)~~


MarkusBarstad:
> Maybe we can use comments or something to indicate to the scanner that this is deliberate and we know what we're doing?

Yeah, and going through the other flagged code, I think this the case with a lot if not all it has found in the `ts/`-folder.

The places where it warns us against non-literal regex-usage, our regexes are still quite controlled, the places where it warns us against prototypes being changed, I cannot tell how that should be able to happen etc.

I will try to configure the scan to not warn us about these specific cases + revert the erroneous replacements of "innerHTML" in this branch.

MarkusBarstad:
@TorsteinHonsi 
I have reverted and only kept the `textContent`-changes for when we clean an element of its contents.

Regarding the other rules which were flagged, essentially I do not think they pose a problem, because user input cannot, as far as I can tell:
- Make Highcharts read custom regexes which takes a very long time to complete
- Make Highcharts alter prototypes (more on this in the scan-doc)

A possible exception for RegExes: 
I am a bit unsure about the [indicator-search-box](http://localhost:3030/samples/#view/stock/demo/stock-tools-gui), eventhough the string is filtered, is it conceivable that an attacker could make the filtering slow stuff down a lot?

PR #18816, opened 2023-04-12
db/rwd-buttons

karolkolodziej:
Added default styling to the rwd buttons, and added select state.

- [x] styling 
- [x] select
- [x] context menu style
- [x] unselect
- [x] test


css/dashboards/gui.css
```diff
@@ -416,21 +416,17 @@
...
-.highcharts-dashboards-edit-context-menu-btn div {
-    height: 4px;
-    background-color: var(--highcharts-dashboards-neutral-color-50);
-    margin: 0 0 4px;
+    cursor: pointer;
 }
 
 .highcharts-dashboards-edit-context-menu-btn:hover {
-    border: 1px solid var(--highcharts-dashboards-neutral-color-80);
+    background-color: var(--highcharts-dashboards-neutral-color-50) !important;
```
pawellysy, line 428: Please use more accurate selector and avoid using `!important`

pawellysy (changes requested):
One minor comment, otherwise good!

bre1470 (approved):
Looks good. 👍

css/dashboards/gui.css
sebastianbochan: Yes, no `!important` anywhere, please!

ts/Dashboards/EditMode/EditMode.ts
```diff
@@ -622,12 +622,31 @@ class EditMode {
...
-                            // reflow elements
+                        callback: (e: PointerEvent): void => {
+                            const button = e.target as HTMLElement,
+                                isSelected =
+                                    button.classList.contains('selected');
+
+                            // Deselect given button and reset board width.
+                            if (isSelected) {
+                                button.classList.remove('selected');
+                                this.board.layoutsWrapper.style.width = '100%';
```
sebastianbochan, line 633: You can remove the width from the style instead of applying 100%. It is by default in our CSS.

sebastianbochan (approved):

pawellysy (changes requested):
image
On Darkmode, the context menu looks wrong 😢 

image


So does the rwd buttons

pawellysy (approved):
LGTM!

PR #18817, opened 2023-04-12
bugfix/18693-tooltip-stickoncontact-followpointer

### LINKED ISSUE #18693: Tooltip `stickOnContact: true` throws error on hover from `followPointer: false` to `true`
#### Works almost, for the first few seconds, but fails when doing mouseover / and inspecting tooltips
txt file uploaded must be renamed to html, then open by Chrome

https://jsfiddle.net/eriksoeberg/z4fr87p5/1/
and 
similar: https://jsfiddle.net/BlackLabel/wxgyuk35/

#### Product version

latest Highmaps

#### Affected browser(s)  Chrome and edge
[map.txt](https://github.com/highcharts/highcharts/files/11009023/map.txt)

hubertkozik:
Fixed #18693, `tooltip.stickOnContact` threw errors on hover between one series with `followPointer` set to true and the second series `followPointer` set to false.

### DIFF
--- a/samples/unit-tests/tooltip/stickoncontact/demo.js
+++ b/samples/unit-tests/tooltip/stickoncontact/demo.js
@@ -158,6 +158,50 @@ QUnit.test(
                     { x: tooltipPosition1.x + 1, y: tooltipPosition1.y + 1 },
                     'Tooltip should move with pointer movement.'
                 );
+
+                chart.update({
+                    tooltip: {
+                        followPointer: false
+                    }
+                }, false);
+
+                chart.series[0].update({
+                    type: 'column',
+                    tooltip: {
+                        followPointer: true
+                    }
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (changes requested):
Thanks for the PR!

So the error happens because we already destroyed `tooltip.tracker` and we're now trying to do stuff with `tooltip.tracker` post mortem?

In that case the pattern we usually use is to unset the reference in the same operation as we destroy the element. Then you don't need to check the length of the keys. This modification passes your tests:

```ts
        if (!this.shouldStickOnContact()) {
            if (tooltip.tracker) {
                tooltip.tracker = tooltip.tracker.destroy();
            }
            return;
        }
```

TorsteinHonsi (approved):

PR #18820, opened 2023-04-13
18801-legend-pages-mac-firefox

Vazonik:
Fixed test added in #18801 that didn't work in Firefox on Mac due to specific font handling.


TorsteinHonsi (approved):
Thanks!

PR #18822, opened 2023-04-13
MathModifier

bre1470:
Builds on #18701


samples/highcharts/data-tools/mathmodifier-formula/demo.details
```diff
@@ -0,0 +1,7 @@
+---
+ name: MathModifier Formula demo
+ authors:
+   - Sophie Bremer
+ js_wrap: b
+ skipTest: true
```
sebastianbochan, line 6: Should we declare skipTest here?

samples/highcharts/data-tools/mathmodifier-formula/demo.js
```diff
@@ -0,0 +1,163 @@
...
+table.setModifier(new MathModifier());
+
+renderTable(container1, table, true);
+renderTable(container2, table.modified);
+
+
+// Add Change Listener
+
+function onChange(e) {
+    console.log(e);
```
sebastianbochan, line 31: Remove `console.log()`

samples/highcharts/data-tools/mathmodifier-formula/demo.js
```diff
@@ -0,0 +1,163 @@
...
+    const tdHighlight = trHighlight.querySelectorAll('td')[cursor.column];
+
+    for (const tr of tbody.querySelectorAll('tr')) {
+        // We search for the correct HTML row
+        if (tr === trHighlight) {
+            tr.classList.add('highlight');
+        } else {
+            tr.classList.remove('highlight');
+        }
+        for (const td of tr.querySelectorAll('td')) {
```
sebastianbochan, line 114: The same situation like in `tr`

samples/highcharts/data-tools/mathmodifier-formula/demo.js
```diff
@@ -0,0 +1,163 @@
...
+            // We search for the correct HTML cell
+            if (td === tdHighlight) {
+                td.classList.add('highlight');
+            } else {
+                td.classList.remove('highlight');
```
sebastianbochan, line 119: You can use `toggle()` instaed of add/remove. It reduces amount of conditions.

test/typescript-karma/Data/Formula/Functions/IF.test.js
```diff
@@ -0,0 +1,21 @@
...
+    assert.strictEqual(
+        Formula.processFormula(formula1, table),
+        2,
+        'IF test should succeed and AVERAGE should calculate first range.'
+    );
+
+    assert.strictEqual(
+        Formula.processFormula(formula2, table),
+        5,
+        'IF test should fail and AVERAGE should calculate second range.'
```
sebastianbochan, line 19: `second range` means?

I have a general question about `expected value.` or `second range`. In my opinion its too general.

tools/code.js
```diff
@@ -162,8 +162,8 @@ function processVariables(content) {
             variables.split(/\r\n|\r|\n/gu).map(function (line) {
```
sebastianbochan, line 162: Please remove the `u` flag because of errors in minified build :)

ts/Data/Formula/FormulaParser.ts
```diff
@@ -0,0 +1,448 @@
...
+/**
+ * @private
+ */
+const decimal1RegExp = /^[+-]?\d+(?:\.\d+)?(?:e[+-]\d+)?/;
+
+
+/**
+ * @private
+ */
+const decimal2RegExp = /^[+-]?\d+(?:,\d+)?(?:e[+-]\d+)?/;
```
sebastianbochan, line 75: What is the difference between `decimal2RegExp` and `decimal1RegExp` ? Add comment please.

ts/Data/Formula/FormulaProcessor.ts
```diff
@@ -0,0 +1,470 @@
...
+    operator: Operator,
+    x: Value,
+    y: Value
+): Value {
+
+    switch (operator) {
+        case '=':
+            return x == y ? 1 : 0; // eslint-disable-line eqeqeq
+        case '<':
+            return x < y ? 1 : 0;
```
sebastianbochan, line 120: You can also use Number(x < y)

ts/Data/Formula/Functions/COUNT.ts
```diff
@@ -0,0 +1,110 @@
...
+    for (
+        let i = 0,
+            iEnd = values.length,
+            value: (Value|Array<Value>);
+        i < iEnd;
+        ++i
+    ) {
+        value = values[i];
+
+        switch (typeof value) {
```
sebastianbochan, line 77: Do we need switch here? We have two cases like in IF / ELSEIF

ts/Data/Formula/Functions/NOT.ts
```diff
@@ -0,0 +1,93 @@
...
+): (boolean|number) {
+    let value = getArgumentValue(args[0], table);
+
+    if (typeof value === 'object') {
+        value = value[0];
+    }
+
+    switch (typeof value) {
+        case 'boolean':
+        case 'number':
```
sebastianbochan, line 68: Should not we include string too?

samples/highcharts/data-tools/mathmodifier-formula/demo.details
bre1470: `data-tools` is not an official module. This is for internal use to make you understand the data layer.

samples/highcharts/data-tools/mathmodifier-formula/demo.js
bre1470: This is a different situation as it selects only `td`s from the current `tr`.

tools/code.js
bre1470: This is in a tool, not in sources.

test/typescript-karma/Data/Formula/Functions/IF.test.js
bre1470: It is describing the IF function in formula 2. I am not sure how this could be more specific.

ts/Data/Formula/FormulaProcessor.ts
bre1470: These need to be booleans. So this needs a different form of fix.

ts/Data/Formula/Functions/NOT.ts
bre1470: No, it has to work like in Excel.

ts/Data/Formula/Functions/COUNT.ts
bre1470: You safe one typeof.

tools/code.js
sebastianbochan: Im to sensitive after last problems :D

ts/Data/Formula/FormulaTypes.ts
```diff
@@ -0,0 +1,287 @@
...
+ *
+ * @return {number}
+ * Number value. `NaN` if not convertable.
+ */
+function asNumber(
+    value: Value
+): number {
+    switch (typeof value) {
+        case 'boolean':
+            return value ? 1 : 0;
```
sebastianbochan, line 133: In terms of boolean type, the `value` is enough

sebastianbochan (changes requested):

ts/Data/Formula/FormulaTypes.ts
bre1470: Please pay more attention to typing.

karolkolodziej (approved):
Good job!

test/typescript-karma/Data/Formula/Functions/AVERAGEA.test.js
```diff
@@ -0,0 +1,16 @@
+import DataTable from '/base/code/es-modules/Data/DataTable.js';
+import Formula from '/base/code/es-modules/Data/Formula/Formula.js';
+
+
+QUnit.test('Formula.processorFunctions.AVERAGEA', function (assert) {
+    const table = new DataTable({
+            columns: { values: [1, 3, 4, 5, 6, 7, false, true, null, '7'] }
```
karolkolodziej, line 7: What am I missing here?
image

pawellysy (commented):
Everything works as expected, just one question to the design:
In excel, when there is a syntax error in the cell, there is a message displayed: `#ERROR!`
In our implementation, there is a last valid value displayed, and the error is added to the cosole. Should we change that, or leave as it is?

test/typescript-karma/Data/Formula/Functions/AVERAGEA.test.js
bre1470: FALSE and NULL get skipped. Strings and TRUE are counted as `1`.

bre1470:
@pawellysy `NaN` should become the outcome for an error. Can you give the cell and term that should fail?

ts/Data/Formula/FormulaTypes.ts
```diff
@@ -0,0 +1,287 @@
...
+
+
+/**
+ * Converts non-number types to number.
+ *
+ * @param {Highcharts.FormulaValue} value
+ * Value to convert.
+ *
+ * @return {number}
+ * Number value. `NaN` if not convertable.
```
pawelfus, line 126: From the description, it should convert, however, it converts strings to 0. Is this expected? And never returns NaN: (except when sending NaN)

Tested with:
```ts
> Highcharts._modules['Data/Formula/FormulaTypes.js'].asNumber('103')
> 0

// NaN tests:
> Highcharts._modules['Data/Formula/FormulaTypes.js'].asNumber(undefined);
> undefined

> Highcharts._modules['Data/Formula/FormulaTypes.js'].asNumber(NaN)
> NaN
```

pawelfus: Edit:
Just realized it's a part of AVERAGEA implementation - the name is a bit misleading (especially we have another `asNumber` method in DataConverter). And this is used in `basicOperation`, which seems to be incorrect, this is g-spreadsheet:
image
In our implementation we treat strings as 0's, so results will be: `1`, `FALSE` and `3`, right?

pawelfus (changes requested):
Just a small nitpicking! Overall looks good

Extra question:
How to use it in API? It's great to have an internal example in `data-tools`, but how devs can setup those modifiers in Dashboards?

TorsteinHonsi (commented):
Thanks! Looks awesome. 

The only concern I have is about the bundling. Adding this to `data-tools.js` increased the file size from 41 to 56 kB (compiled, not gzipped). What is the plan for bundling with the Dashboard core file(s)? I imagine this is a feature that is not always required, and perhaps required more often with the data grid than with a vanilla dashboard? So perhaps it would be better to bundle it as a standalone module file?

pawellysy (approved):
LGTM!

bre1470:
@pawel
Sorry. This PR is for master. The climate demo will demonstrate it for Dashboards. Or do you like to see the modifier demo expanded?

pawelfus:
> The climate demo will demonstrate it for Dashboards. Or do you like to see the modifier demo (samples/highcharts/data-tools/datamodifier-tree) expanded?

I think both. Climate should promote the right way to use it (instead of workarounds) and we need a simple demo presenting it, so we can link it to API/docs.

bre1470:
@pawelfus 
This will require a new PR. All the data layer demos are not for API docs as this would requires a public bundle in master, which `data-tools` was decided to not to be.

test/typescript-karma/Data/Formula/Functions/IF.test.js
bre1470: Please take another look, @sebastianbochan.

samples/highcharts/data-tools/mathmodifier-formula/demo.js
```diff
@@ -0,0 +1,159 @@
...
+    const trs = tbody.querySelectorAll('tr');
+    const trHighlight = trs[cursor.row];
+    const tdHighlight = trHighlight.querySelectorAll('td')[cursor.column];
+
+    for (const tr of trs) {
+        // We search for the correct HTML row
+        if (tr === trHighlight) {
+            tr.classList.add('highlight');
+        } else {
+            tr.classList.remove('highlight');
```
sebastianbochan, line 112: You can use `toggle()`

sebastianbochan (approved):

TorsteinHonsi (approved):

PR #18825, opened 2023-04-14
bugfix/18066-categories-xaxis-unwanted-label

### LINKED ISSUE #18066: -1 on xAxis when legends filtered out
#### Expected behaviour
When Legend filter is used, x and y axis should be readjusted automaticaly.

#### Actual behaviour
"-1" appears on the xAxis. [-1 on xAxis](https://i.stack.imgur.com/kOlnp.png)
https://stackoverflow.com/questions/74628650/highcharts-1-on-xaxis-when-legends-filtered-out?noredirect=1#comment131732957_74628650

https://jsfiddle.net/ag_test/pe9t4vxu/1/
- add a new data series to the chart (by clicking the add series button) 
- filter out the old series by clicking the legends of them.

#### Product version

 Highcharts JS v10.3.2

#### Affected browser(s)
Chrome 108.0.5359.71

hubertkozik:
Fixed #18066, after updating the series data to one point with category xAxis there were unwanted labels.

### DIFF
--- a/samples/unit-tests/series/softthreshold/demo.js
+++ b/samples/unit-tests/series/softthreshold/demo.js
@@ -138,6 +138,20 @@ QUnit.test('Soft threshold', function (assert) {
         [0, 10],
         'Hard max and flat data and minRange'
     );
+
+    chart.xAxis[0].update({
+        categories: ['cat1', 'cat2']
+    }, false);
+    chart.series[0].points[2].remove(false);
+    chart.series[0].points[1].remove();
+
+    assert.strictEqual(
+        chart.xAxis[0].tickPositions.length,
+        1,
+        `There shouldn't be visible unwanted ticks after updating the series,
+        minRange should be updated based on data if not set in chart
+        configuration (#18066).`
+    );
 });
 
 QUnit.test('Soft threshold = false', function (assert) {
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (commented):
As far as I understand, the line you removed works in tandem with this line:

https://github.com/highcharts/highcharts/blob/a68150996095c1438d5d00b2ee097ce002710225/ts/Core/Axis/Axis.ts#L1253

Together, they make sure that we don't end up running the same expesive loop over all series xData for every axis redraw. Can you explore this a bit more? Why does #18066 happen in the first place? Maybe we should delete `axis.minRange` on adding series or updating data if that's the root problem?

hubertkozik:
I rethought the whole flow and now think I found a better place to apply the fix. As you suggested we should delete `axis.minRange` if there was a series add/update event (or changing the visibility of the series), so really, if the series is dirty. In my opinion, it shouldn't be deleted in the series update/redraw method, because there are too many places where the fix should be applied. I think, that the correct place to do this is inside Axis `setScale` method - thanks to that, we won't make unnecessary expensive loops. what do you think about it now?

Btw. I have also added one more test for this fix to cover it more precisely.

PR #18827, opened 2023-04-14
bugfix/18103-y-zoom-inverted-issue

### LINKED ISSUE #18103: Zooming issue in on inverted chart
#### Expected behaviour
I am using an inverted chart to create a timeline. Zooming is enabled and works most of the time, however when want to zoom and place my cursor on the line of the first category of the y-axis (actually the x-axis since its inverted), and drag to zoom the blue zoom effect is shown but the graph does not zoom in. 

This bug occurs when you start zooming in between the red lines.
image

The GIF below shows the problem. The first two zooms work, the third zoom shows the problem. Another strange thing is that when we zoom in normally, the zoom button shows. But if you then zoom in by starting at the first category, the zoom button disappears.
timeline

#### Actual behaviour
Zoom should work as expected.

See JS fiddle below. Note that I've removed my data, and behavior stays the same.
https://jsfiddle.net/leonardomathon/58ojyrqm/3/

#### Product version
Latest version of Highcharts from CDN

#### Affected browser(s)
All major browsers, latest versions (Chrome, Edge, Firefox)

jakubjanuchta:
Fixed #18103, Y zoom while panning on top of the plot when chart inverted was not working.

### DIFF
--- a/samples/unit-tests/chart/zoomtype/demo.js
+++ b/samples/unit-tests/chart/zoomtype/demo.js
@@ -2,7 +2,7 @@ QUnit.test('Zoom type', function (assert) {
     var chart = Highcharts.chart('container', {
         chart: {
             type: 'line',
-            zoomType: 'x',
+            zoomType: 'y',
             width: 600,
             height: 300
         },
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #18831, opened 2023-04-14
Website/small demos corner fix

nasin:
Hi, I removed the hard coded border radius settings from the small gantt demos (that appear on the product page.)


TorsteinHonsi (approved):
Thanks!

PR #18832, opened 2023-04-14
DB: Refactored classnames and styles, added langs to the sidebar.

sebastianbochan:


pawellysy (commented):
Now, the first level of accordeon is expanded by default. Also, it takes 2 clicks to expand collapse the 2 level. Could you add the collapse behavior to class as well? It is handled by inline styles right now, and I don't quite know, where to put it, so that the display value is correctly set. 😊

pawellysy (approved):

PR #18837, opened 2023-04-17
db/accordeon-ts-fixes

sebastianbochan:


ts/Dashboards/EditMode/AccordeonMenu.ts
```diff
@@ -185,11 +188,11 @@ class AccordeonMenu {
         let value = component.options as any;
 
         for (let i = 0, end = propertyPath.length; i < end; i++) {
-            if (!value) {
-                return;
+            if (value) {
+                value = value[propertyPath[i]];
             }
```
pawellysy, line 193: If there is no value, the function should return undefined. Now, it will continue to do a loop, and it might return an object. 
IMO, it should be the way, it was implemented previously

PR #18840, opened 2023-04-18
db/sidebar-add-new-component

sebastianbochan:


css/dashboards/gui.css
```diff
@@ -682,19 +682,23 @@
     display: block;
 }
 
+.highcharts-dashboards-edit-grid-items {
+    margin-top: 20px;
+}
+
 .highcharts-dashboards-edit-grid-items div {
-    width: calc(50% - 20px);
+    width: calc(100% - 20px);
```
pawellysy, line 690: maybe there is a way to avoid using calc?
Why don't we go for flexbox and, something like `align-items: streach` + margins on the surrounding div?

pawellysy (changes requested):
One comment to styling, otherwise looks good!

pawellysy (approved):
LGTM!

PR #15960, opened 2021-06-24
bugfix/14143-14434-stickOnContact-useHTML-outside

khlieng:
Fixed #14143, #14434, #13310, tooltip with `stickOnContact` and `useHTML` or `outside` enabled did not work properly.

Replaces #14473.
Closes #14434.
Closes #13310.

PR #18842, opened 2023-04-18
Fixed #15960, tooltip with stickOnContact and outside

TorsteinHonsi:
Fixed #15960, a regression causing tooltip with `stickOnContact` and `outside` to not work correctly.


khlieng (approved):
Seems to work 👍

PR #18850, opened 2023-04-19
db/fix-context-menu-buttons

pawellysy:


ts/Dashboards/EditMode/Menu/MenuItem.ts
```diff
@@ -133,25 +112,32 @@ class MenuItem {
...
             name: options.id,
             title: options.collapsable ? '' : options.text || '',
-            callback,
             item: this,
-            nestedOptions: options.nestedOptions,
             icon: options.icon,
             mousedown: options.events && options.events.onmousedown,
             value: options.value || '',
-            onchange: options.events && options.events.change,
+            onchange: callback,
```
sebastianbochan, line 140: WIll not break renderInput where onChange is used? So when we're typing the "new chart's title", all is correct?

sebastianbochan (approved):

PR #18854, opened 2023-04-20
bugfix/17006-datalabels-series-opacity

### LINKED ISSUE #17006: Interacting with series that has opacity changes the dataLabels opacity.
#### Expected behaviour
Decide if the series `opacity` should have an impact on the `dataLabels`.

#### Actual behaviour
When series `opacity` is set, hovering over the data label that is inside the column causes changing the label text color.

Hover over the column and see how the color of the labels is changing.
https://jsfiddle.net/BlackLabel/kcq0f3db/

#### Product version
 Highcharts JS v9.3.3 (2022-02-01)

#### Affected browser(s)
All

hubertkozik:
Fixed #17006, data labels should inherit the series opacity.

### DIFF
--- a/samples/unit-tests/datalabels/multiple-labels/demo.js
+++ b/samples/unit-tests/datalabels/multiple-labels/demo.js
@@ -192,4 +192,17 @@ QUnit.test('Multiple data labels general tests.', function (assert) {
         result,
         'All data labels should be visible when chart is inverted (#12370).'
     );
+
+    const newOpacity = 0.4;
+
+    chart.series[0].update({
+        opacity: newOpacity
+    });
+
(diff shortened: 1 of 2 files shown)


TorsteinHonsi:
I think it makes sense, but:
* What's up with the visual diff? It looks like the data labels are semi-transparent in a lot of series types, even though the main series is not. Does that mean that we are handling the `opacity` option wrong for these series?
* What if a user wants semi-transparency for the series elements but not for the data labels? Then we need a `dataLabels.opacity` option?
* Does it work with `series.update({ opacity })` in its current form?

hubertkozik:
Closing this PR. Future coding should be done in the new PR.

PR #18862, opened 2023-04-21
Enhancement/sonification rewrite

oysteinmoseng:
Sonification module v2 with many new features for audio charts, now part of the official API.

**Upgrade notes**

The new version of the sonification module is not backwards compatible with the old, experimental module.
___
**API HIGHLIGHTS**
- Still builds the `sonification.js` module - no dependencies other than core.
- Adds `chart.sonify()`, `chart.toggleSonify()`, `series.sonify()`, `point.sonify()`. Also adds `chart.sonification` with the rest of the functionality.
- Adds a top level `sonification` API option, as well as one for series.
- Adds a global `Highcharts.sonification` object with access to classes that should be available to users.
- New functionality highlights include:
    - Built-in lightweight synthesizer - much improved performance, sound & flexibility over old architecture
    - Speech support
    - Navigation support, incl. features like timeline filters and scrubbing
    - Extended & more flexible mapping support incl. tremolo, audio filters, delays, and speech parameters
    - Support for human readable note names
    - Mapping to musical scales
    - Multi-axis & logarithmic axis support
    - Context tracks / cues
    - Conditional tracks
    - Play marker / cursor
    - Point grouping support (separate from visual data grouping)
    - MIDI export (however not using MIDI internally)
- Lost functionality from old module:
    - Real-time mapping functions (maybe an addition for the future if requested, but usually better ways to achieve goals)
    - Define custom order & groupings of series to sonify together (addition for future if requested)

**THINGS TO NOTE:**
- Made `alt` attributes allowed in AST (figured this would be a common use case for icons). This is pretty much the only notable change to core code.
- One or two related a11y module fixes are baked in.
- Any compatibility with the old sonification module is more or less accidental - this is a total rewrite with a different API & architecture.
- `demo/sonification-music`: This demo has some icons, should probably be served from us instead of Wikimedia? License-wise should be fine to grab, assuming we don't want to make our own icons.
- Demos in general could maybe use a touchup from an actual designer at some point...
- Renamed "Accessible charts" to "Accessibility" for demo page to reduce confusion about which charts are accessible
- Added new demo page category "Audio charts" for Highcharts demos, to separate it from accessibility. Just placed in "General" for other products, not sure what the plans are with new demo pages.
- We can consider splitting some instrument presets into a separate module, especially if we eventually add more (maybe from someone who actually deals with programming synths more than once a year...)

**DEMO LIST:**
- `highcharts/demo/all-instruments`
- `highcharts/demo/audio-boxplot`
- `highcharts/demo/plotline-context`
- `highcharts/demo/sonification-music`
- `highcharts/demo/sonification-navigation`
- `highcharts/demo/sonified-function`
- `maps/demo/audio-map`
- `maps/demo/map-trajectory`
- `stock/demo/soundscape`

In addition see samples under `highcharts/sonification/*` for docs/API use.

Demos should give a pretty good idea of the API as well, in case we need to reconsider some options naming or any unintuitive concepts. Would be good to get a 2nd set of eyes on all of this!

**KNOWN TODOS** (can be done after merge to master):
- ~Doclets / API documentation~ (done, but track structure could be better in the API if there is a way, bit confusing mixing instrument and speech tracks in the same object)
- A high-level TypeScript review would be nice (so adding Sophie as reviewer too)
- Docs / Tutorials

**LONGER TERM PLANNED:**
- Smart defaults (high impact, low effort)
- Built-in accessible navigation integration with a11y module (dependent on a11y module project)


oysteinmoseng:
@bre1470 If you have time to take a look at high-level TypeScript stuff, `ts/Extensions/Sonification/Options.ts`,  `ts/Extensions/Sonification/Sonification.ts` and `ts/Extensions/Sonification/SonificationTimeline.ts` might be the most important.

TorsteinHonsi (approved):
Thanks for the awesome work and beautiful demos!

PR #18868, opened 2023-04-24
Website/homepage 2023

nasin:
Hi, I'm working on creating new homepage animations. Marketing asked for a "randomizer" that will load a random chart on page load. I have 5 charts in demo.js and a random number generator to make this happen.


TorsteinHonsi (approved):
Thanks!

PR #18870, opened 2023-04-25
db/climate-dataGrid-header

karolkolodziej:
Changed and hidden some `dataGrid` header names.

### DIFF
--- a/samples/dashboards/demos/climate/demo.css
+++ b/samples/dashboards/demos/climate/demo.css
@@ -366,6 +366,7 @@ div.highcharts-dashboards-wrapper .highcharts-dashboards-edit-context-menu-btn {
     border-color: var(--color-header-border);
     color: var(--color-header-text);
     padding: 0 12px;
+    font-size: 12px;
 }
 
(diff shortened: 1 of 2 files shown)


sebastianbochan (approved):

pawelfus:
> Header cells should have some padding, so text does not get cut off.

I don't think it will help. It already has padding, but the text is too long, so it's "ignored". Just like other cells:
image

> I also get into scrolling troubles when scrolling while the pointer is over the header.

Could you describe those troubles? Works "fine" for me (osx + touchpad) - or at least works the same way as on the main branch..

Personally, I believe both issues deserve their own tickets. The first one could be worked around by setting shorter names, but the issue will go back when testing on mobile devices, where screens are even smaller.

bre1470:
I recorded a video in the new ticket.

In general you can fix most (if not all) issues by using HTML table. I do not see an advantage in using divs.

bre1470 (approved):
We agreed on follow-up tickets.

pawellysy (approved):

PR #18872, opened 2023-04-25
bugfix/map-datalabels-insets

hubertkozik:
Fixed a map data labels placement issue for insets.


KacperMadej (changes requested):
Looks good.
Because this won't make it to v11 there is some time to add a test for new map collection map to avoid the problem that caused this PR. It could be done with a mocked map so there is no dependency on the new map collection version.

PR #18875, opened 2023-04-25
db/dataPool-connector-names

karolkolodziej:
Added method that returns names of connectors.

### DIFF
--- a/ts/Data/DataPool.ts
+++ b/ts/Data/DataPool.ts
@@ -120,6 +120,20 @@ class DataPool {
     }
 
     /**
+     * Returns the names of all connectors.
+     *
+     * @private
+     *
+     * @return {Array<string>}
+     * Names of all connectors.
+     */
+    public getConnectorsNames(): Array<string> {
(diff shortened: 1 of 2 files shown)


ts/Data/DataPool.ts
```diff
@@ -119,6 +119,20 @@ class DataPool {
...
+    /**
+     * Returns the names of all connectors.
+     *
+     * @private
+     *
+     * @return {Array<string>}
+     * Names of all connectors.
+     */
+    public getConnectorsNames(): Array<string> {
+        return this.options.connectors.map(
```
sebastianbochan, line 131: We should not use `.map`, `.filter` due to performance reasons.

sebastianbochan (changes requested):

pawellysy (changes requested):
Could you add Unit test to that? 😅

sebastianbochan (approved):

bre1470 (approved):
Looks good. 👍

PR #18877, opened 2023-04-25
bugfix/18856-boost-last-point-hover

### LINKED ISSUE #18856: When boost mode enable, last point is not accessible via hover
#### Expected behaviour
Last point of the series must be highlight when the mouse is at the right, only work when boost mode is disabled:
image

#### Actual behaviour
When boost mode is enabled we have to zoom to reach the last point while hovering the series.
image

Look after `boostThreshold` param
http://jsfiddle.net/7zg9L5n0/1/

#### Product version

#### Affected browser(s)
Tested on Chrome and Firefox

Vazonik:
Fixed #18856, the last point wasn't displayed while hovering in the boost module.


TorsteinHonsi (approved):
Thanks, looks good 👍

PR #18885, opened 2023-04-26
Fixed arc diagram for mobile and streamgraph data

nasin:
I fixed the arc diagram. It now inverts for mobile. I fixed the streamgraph data series names to reflect the actual data.


nasin:
I created the new directory called homepage-23. I made changes to the arc diagram and fixed some other small issues with the other charts.

In homepage-2023, I removed the arc diagram from the rotation and fixed the streamgraph data so that it looks good on mobile.

PR #18886, opened 2023-04-26
Fixed pixel issues after latest merges.

bre1470:


TorsteinHonsi:
If you use `assert.close` for the two first ones, it will pass for the rest of us too.

TorsteinHonsi (approved):

PR #18888, opened 2023-04-26
feature/locator-map-demo'

magdalena-gut:
Locator Map Demo for demo pages.

TO DO:
---
- [x] - Instead of putting the initiating code in the chart constructor callback, use events.
- [x] - Remove the locator div from the HTML. Instead, this can be generated as part of the snippet.
- [x] - optional toggle.
- [x] - update path when redrawing


TorsteinHonsi:
Thanks so far! Also, let me elaborate on this:

> Gather all the related code in one place

The idea is that we get a clear separation of the code that is related to the _locator_, and the code that is the main chart (in this case Italy). The design goal is that we have all the locator code assembled in _one_ block of code, so that we can copy-paste that block into another demo, and then a locator map appears. Take any of the standard demos, [like Africa](https://jsfiddle.net/gh/get/jquery/1.11.0/highslide-software/highcharts.com/tree/master/samples/mapdata/custom/africa) for example.

magdalena-gut:
> Thanks so far! Also, let me elaborate on this:
> 
> > Gather all the related code in one place
> 
> The idea is that we get a clear separation of the code that is related to the _locator_, and the code that is the main chart (in this case Italy). The design goal is that we have all the locator code assembled in _one_ block of code, so that we can copy-paste that block into another demo, and then a locator map appears. Take any of the standard demos, [like Africa](https://jsfiddle.net/gh/get/jquery/1.11.0/highslide-software/highcharts.com/tree/master/samples/mapdata/custom/africa) for example.

Thank you for clarifying and pointing in the right direction. I moved the code to the one block as you requested. I added the update of the pointer when zooming as well. 

I tested the locator with some other maps and unfortunately, it doesn't work perfectly. (North America is a good example: https://jsfiddle.net/BlackLabel/wvfsg7Lx/)

I was wondering about modifying the calculations of the pointer(locator). For example, based on the bounding box of the main map (or continent only). It has some limitations when zooming, but I thought about counting the size and shift of the pointer via zoom factor and the center of the main map. What do you think?

TorsteinHonsi:
Thanks!

> I tested the locator with some other maps and unfortunately, it doesn't work perfectly. (North America is a good example

It looks correct though, except it is not centered. So in addition to finding the edges of the frame you probably just need to locate the center point and set the [projection.rotation](https://api.highcharts.com/highmaps/mapView.projection.rotation) of the locator map.

> I was wondering about modifying the calculations of the pointer(locator)

First of all the term "pointer" is a bit confusing in Highcharts, it usually means the pointing device (mouse or touch). Maybe you can call it "frame" or something in your code?

>  I thought about counting the size and shift of the pointer via zoom factor and the center of the main map. What do you think?

As said above, I think it looks correct, except we should rotate the globe.

magdalena-gut:
> As said above, I think it looks correct, except we should rotate the globe.

Changes added. Referring to the previous one about the calculations - I was thinking about the curvature as well - I'm not sure if the createPoints() function is necessary here. Two cases:
https://jsfiddle.net/BlackLabel/5xwuah6g/
https://jsfiddle.net/BlackLabel/gukpz1d8/

TorsteinHonsi:
> I'm not sure if the createPoints() function is necessary here. Two cases:

The first of those two cases looks correct to me, right?

But I agree North America looks wrong: https://jsfiddle.net/BlackLabel/wvfsg7Lx/ . It seems you created only the bottom points and the top points, but you need to create the intermediate points all around the frame, also along the top and the bottom. Otherwise Highcharts maps will assume the [shortest line on the globe](https://jsfiddle.net/BlackLabel/gukpz1d8/).

TorsteinHonsi (changes requested):
Thanks! I did some minor optimizations and code organization. However I noticed there appears to be still some quirk with the edge algorithm:
* The top and bottom edges seem off in many cases. For example, when I apply the locator snippet to [Australia](https://jsfiddle.net/gh/get/jquery/1.11.0/highslide-software/highcharts.com/tree/master/samples/mapdata/countries/au/au-all), the northern tip is cut off, and there is too much space south of Tasmania, compared to the main chart: https://jsfiddle.net/highcharts/yjcuhd28/ . Edit: it seems that the margins are not handled - when I set `chart.margin: 0` it works: https://jsfiddle.net/highcharts/yjcuhd28/2/ . Which means that `chart.plotTop` and `chart.plotLeft` must be considered when calculating the frame.
* For the full world map (stress testing, I know...), the shape translation goes NaN: https://jsfiddle.net/highcharts/z9350xtc/

magdalena-gut:
> Thanks! I did some minor optimizations and code organization. However I noticed there appears to be still some quirk with the edge algorithm:
> 
> * The top and bottom edges seem off in many cases. For example, when I apply the locator snippet to [Australia](https://jsfiddle.net/gh/get/jquery/1.11.0/highslide-software/highcharts.com/tree/master/samples/mapdata/countries/au/au-all), the northern tip is cut off, and there is too much space south of Tasmania, compared to the main chart: https://jsfiddle.net/highcharts/yjcuhd28/ . Edit: it seems that the margins are not handled - when I set `chart.margin: 0` it works: https://jsfiddle.net/highcharts/yjcuhd28/2/ . Which means that `chart.plotTop` and `chart.plotLeft` must be considered when calculating the frame.
> * For the full world map (stress testing, I know...), the shape translation goes NaN: https://jsfiddle.net/highcharts/z9350xtc/

Thank you, I implemented the margin improvements you suggested.  When it comes to NaN's, it seems to be an Equal Earth case only. I added NaN's exceptions for the time being, but it seems to be challenging to address this issue effectively. Do you have any ideas on how to tackle this?

TorsteinHonsi (approved):

PR #18892, opened 2023-04-27
bugfix/18884-sunburst-circular-labels

### LINKED ISSUE #18884: Sunburst chart: Circular labels don't show when the numbers are large and the difference between parent and child is very small
#### Expected behaviour
When dataLabels are enabled and circularMode is enabled, the labels should always show.

#### Actual behaviour
When the difference between parent node and child node is very small, e.g. root.value: 6284, child node value: 6283, the label for the child does not show.

https://jsfiddle.net/L37wze5h/1/

#### Product version
Highcharts 10.2.1

#### Affected browser(s)
Microsoft Edge, Firefox, Chrome

jakubSzuminski:
Fixed #18884, sunburst circular labels did not work when the difference in value was small..

**Demo of the problem:** https://jsfiddle.net/BlackLabel/4hqxa3s0/
**After fix:** https://jsfiddle.net/BlackLabel/fjrus9pw/

### DIFF
--- a/ts/Series/Sunburst/SunburstPoint.ts
+++ b/ts/Series/Sunburst/SunburstPoint.ts
@@ -104,11 +104,17 @@ class SunburstPoint extends TreemapPoint {
             end = -Math.PI / 360;
             upperHalf = true;
         }
-        // Check if dataLabels should be render in the
-        // upper half of the circle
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (approved):
Thanks!

PR #18895, opened 2023-04-27
bugfix/18876-tooltip-update-userOptions

### LINKED ISSUE #18876: Updating the tooltip.pointFormat does not work
#### Expected behaviour
The pointFormat should update.

#### Actual behaviour
The `tooltip.pointFormat` does not update, when is defined in `chart.update()`

https://jsfiddle.net/BlackLabel/zd7t1xgL/

#### Product version
v10.3.3

### Workaround
Define the empty tooltip object in configuration. 
https://jsfiddle.net/BlackLabel/9b1p5vkn/617/

hubertkozik:
Fixed #18876, updating the `tooltip` didn't work when wasn't declared in the chart configuration.

_______

I have also removed some castings to `any` around the code I have added a fix.

### DIFF
--- a/samples/unit-tests/tooltip/pointformat/demo.js
+++ b/samples/unit-tests/tooltip/pointformat/demo.js
@@ -11,23 +11,25 @@ QUnit.test('Repetetive formats', function (assert) {
             width: 200,
             height: 200
         },
+        series: [{
+            data: [1.11]
+        }]
+    });
+
+    chart.update({
         tooltip: {
             headerFormat: '',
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (changes requested):
Awesome! I like that you use `cleanRecursively` instead of the `userOptions`.

As the final step, we can now apply optional chaining for those long && expressions.

TorsteinHonsi (approved):
Thanks!

PR #18896, opened 2023-04-27
bugfix/14602-arearange-label-on-area-true

### LINKED ISSUE #14602: Arearange serie label is not displayed when onArea:true
#### Expected behaviour
On an arearange serie, a label with onArea:true is displayed inside the area.

#### Actual behaviour
The label is not displayed

https://jsfiddle.net/v73pu4z5/

#### Product version
Highcharts 8.2.2

#### Affected browser(s)
Tested on Chrome 86

jakubjanuchta:
Fixed #14602, arearange `series.label.onArea` was not implemented.

### DIFF
--- a/samples/unit-tests/series-arearange/arearange/demo.html
+++ b/samples/unit-tests/series-arearange/arearange/demo.html
@@ -1,5 +1,6 @@
 <script src="https://code.highcharts.com/stock/highstock.js"></script>
 <script src="https://code.highcharts.com/stock/highcharts-more.js"></script>
+<script src="https://code.highcharts.com/modules/series-label.js"></script>
 
 <div id="qunit"></div>
 <div id="qunit-fixture"></div>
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (changes requested):
Thanks, looks good!

For the unit test, can you add the series labels to one of the existing charts instead of creating a new `test` with a new chart? It's for keeping the performance of the test unit on a decent level.

jakubjanuchta:
> Thanks, looks good!
> 
> For the unit test, can you add the series labels to one of the existing charts instead of creating a new `test` with a new chart? It's for keeping the performance of the test unit on a decent level.

I wanted to extract this tests to a new chart for better visibility, but I understand that performance is more important in that case. I moved them to another test unit.

TorsteinHonsi (approved):
> I wanted to extract this tests to a new chart for better visibility, but I understand that performance is more important in that case. I moved them to another test unit.

@jakubjanuchta Sure, sometimes it is better to create a new to avoid big test monsters.

Issue #7346, opened 2017-11-02
Bars outside xAxis visible region when one bar per category


gedrox:
When we have 3 categories with one data sample in each one in different "x" points, the bars in the sides are not visible. E.g. 
* CatA has value in x=1
* CatB has value in x=2
* CatC has value in x=3

In this case CatA and CatC bars will be outside the chart visible area.

#### Expected behaviour
Three bars should be shown.

#### Actual behaviour
Center bar is visible, bars on sides are outside the visible area.

#### Live demo with steps to reproduce
https://jsfiddle.net/gcgayzzu/2/

#### Affected browser(s)
All

pawelfus:
Hi @gedrox 

Thanks for reporting the issue. It's caused by using separate series: each series has just one point, so we are unable to calculate distance required per one series. This can be fixed it two ways:
- if you have categories, use `yAxis.type="category"`: https://jsfiddle.net/gcgayzzu/3/
- if know the distance, use `series.pointRange=value`: https://jsfiddle.net/gcgayzzu/4/

Just a side note: I'm not sure if this is expected, but in your demo columns are translated to make space for another columns within the same series. If you have only one point per series, then you may want to set `series.grouping=false`, please take a look: https://jsfiddle.net/gcgayzzu/5/

gedrox:
In fact we have `datetime` xAxis type (with HighStack option xAxis.ordinal) and this is just a data corner case we have problems with; usually there will be values for all categories.

pawelfus:
Thank you for the details!

gedrox:
This is the example for that: https://jsfiddle.net/gcgayzzu/6/.

pawelfus:
Thanks again. We really appreciate this. Just in case, setting `pointRange` resolves the issue: https://jsfiddle.net/gcgayzzu/7/ 

*Internal note:*
Simple idea to concat data from all series, then calculate closest range is not that simple: https://jsfiddle.net/3f57soLr/ (panning is broken).

Issue #7220, opened 2017-10-03
x-axis min and max are not respected for a datetime column series containing a single value


andiger:
#### Expected behaviour
The x axis should show 2016 to 2020 with a column only covering the 2016 tick.

#### Actual behaviour
The x axis seems to start at 2005 and end at 2029 (see console in demo). The column covers multiple years.

#### Live demo with steps to reproduce
http://jsfiddle.net/adgadg/6xfgc2rd/

Uncommenting the dummy data value in the total-series gives a much more reasonable start and end dates for the x-axis and the column only covers the 2016 tick.

pawelfus:
Hi @andiger 

This is caused, as you noticed, by using just one column. With just one column, we don't know if you want to represent a year, a day, or maybe it's a value for an hour and we can only guess the best range and size for a point. 

In that cases, please use `series.pointRange` to inform Highcharts, what exactly that points represents, take a look: http://jsfiddle.net/6xfgc2rd/11/

andiger:
Oh, I see. That solves my problem. Thank you.

Issue #4112, opened 2015-04-17
pointPlacement isn't working if there is no column type series


KacperMadej:
If pointPlacement is set, but chart has no column series, then pointPlacement is not applied.

Example with column type series: http://jsfiddle.net/mcLoxpur/1/
Without: http://jsfiddle.net/mcLoxpur/

highslide-software:
It requires the `pointRange` to be known. When a column is present, the pointRange is computed. http://jsfiddle.net/highcharts/mcLoxpur/2/.

This information is now added to the docs, http://api.highcharts.com/highcharts#plotOptions.series.pointPlacement

drmrbrewer:
But it seems that pointRange cannot be set per series?  See http://jsfiddle.net/drmrbrewer/pggdnx35/1/ and the comments therein.

highslide-software:
Why would you want different point ranges? The point range only dictates the width of each column, and the position when using pointPlacement. The X values are not affected.

drmrbrewer:
I have a situation where I have a datetime x axis, and am plotting points that start at 1 hour spacing (between this point and the previous), then move to 2, then 3, etc... so the spacing between points in datetime is not constant.  So if I want to plot the points at pointPlacement of -0.5 (i.e. in the middle of the range for the point concerned), then I don't want a single pointRange that applies across the whole chart, for all series.

Initially I figured I'd have to divide my points into different series, according to point spacing.  But that gets awfully complicated, and I'm not sure how to join the line from one series to the next (bridge the gap).  Even worse if it's not a neat progression from 1hr to 2hr etc, but instead the spacings are somewhat more random.

So what I figure now is just to keep them all in a single series, but then I can't use pointPlacement to place my points correctly because of the variable spacing, so instead I am actually plotting them at intermediate positions (i.e. setting x explicitly to the mid point), and deal with the shared tooltip separately (to ensure the tooltip doesn't fire at the intermediate positions).

Ultimately, I would suggest that it would be really neat to be able to set pointRange *per point* or *per series*, so that pointPlacement will look first at pointRange for the point, and if not defined then look at pointRange for the series, and if not defined then look at pointRange for the chart.

That way you can be flexible with pointPlacement when you have data with variable spacing along the x axis.

highslide-software:
I guess the `pointRange` wasn't designed to work like that. You could try resolving it by using categories instead of a datetime axis.

drmrbrewer:
[...]

Issue #3071, opened 2014-05-19
Column gets cropped off if data point is max of X-Axis


purshi-k:
hi,
I am facing issue where Column gets cropped off if data point is max of X-Axis.
Example is @ "http://jsfiddle.net/B4Tj2/2/". where it is necessary we fix our x-axis range by providing min & max but the second series doesn't just show up on chart.
I dont want to use " clip: false," because that doesn't render whole as a chart. I know maxPadding wont work as i am setting max of x-axis. Anything else that i can work around to fix this issue?
Any Help is much appreciated. 

Thanks,
Purush

PR #18897, opened 2023-04-27
bugfix/17791-wrong-extremes-one-point-multiple-series

### LINKED ISSUE #17791: Wrong extremes for multiple column series with one point
#### Expected behaviour

Extremes will be calculated properly for multiple column series with one point.

#### Actual behaviour

Two columns are not visible, showing/hiding the series by legend item click is also broken.

https://jsfiddle.net/BlackLabel/e9osLb2h/

#### Product version
Highcharts v10.2.1

#### Workaround
Set proper `minPadding` and `maxPadding`  values.
https://jsfiddle.net/BlackLabel/s1fqyd73/

hubertkozik:
Fixed #17791, `pointRange` wasn't calculated for multiple series with single points.

____

This was an issue, which was reported by users many times, e.g. 
- https://github.com/highcharts/highcharts/issues/7220
- https://github.com/highcharts/highcharts/issues/7346
- https://github.com/highcharts/highcharts/issues/4112
- https://github.com/highcharts/highcharts/issues/3071
- https://github.com/highcharts/highcharts/issues/712
- https://github.com/highcharts/highcharts/issues/987

I think it would be good to make native support in HC Core for this problem.


TorsteinHonsi (changes requested):
Thanks! 

Since we're now in the core of the core, I was hoping for a more compact solution. We already have the `Axis.getClosest` function that is supposed to return the distance between the closest points. A modification of that function shows promising results:

```ts
    public getClosest(): number {
        let ret: any,
            closestSingle = Infinity;

        if (this.categories) {
            ret = 1;
        } else {
            const singleXs: number[] = [];
            this.series.forEach(function (series): void {
                const seriesClosest = series.closestPointRange,
                    visible = series.visible ||
                        !series.chart.options.chart.ignoreHiddenSeries;

                if (series.xData?.length === 1) {
                    singleXs.push(series.xData[0]);
                } else if (
                    !series.noSharedTooltip &&
                    defined(seriesClosest) &&
                    visible
                ) {
                    ret = defined(ret) ?
                        Math.min(ret, seriesClosest) :
                        seriesClosest;
                }
            });
            if (singleXs.length) {
                singleXs.sort((a, b): number => a - b);
                let i = singleXs.length;
                while (--i) {
                    const distance = singleXs[i] - singleXs[i - 1];
                    if (distance > 0 && distance < closestSingle) {
                        closestSingle = distance;
                    }
                }
            }
        }
        return Math.min(closestSingle, ret || Infinity);
    }
```
I tested this successfully on the OP demo, but not on your unit tests, so there may be edge cases I didn't think of. If this is a viable way forward, we should also
* Clean up the use of any/Inifinity/defined in the logic. We can probably just start out with Infinity for both variants, then return the number only if it is < Infinity.
* Assess whether it is better to refactor the get-smallest-distance code into a utility function to avoid repeating ourselves. It is now used in [Series.ts](https://github.com/highcharts/highcharts/blob/6af547daa0aaecae9bb1ed1b0b59453e600053b6/ts/Core/Series/Series.ts#L1690-L1718), [Axis.ts](https://github.com/highcharts/highcharts/blob/6af547daa0aaecae9bb1ed1b0b59453e600053b6/ts/Core/Axis/Axis.ts#L1261-L1274) and maybe other places, plus now this new logic.

hubertkozik:
@TorsteinHonsi I have decided to not refactor code in [Series.ts](https://github.com/highcharts/highcharts/blob/6af547daa0aaecae9bb1ed1b0b59453e600053b6/ts/Core/Series/Series.ts#L1690-L1718) because it is the only place in HC, where `closestPointRange` is calculated for only one series. The other places are calculating the `closestPointRange` based on all series on one axis - that's why I have moved the code to a utility function. By the way, changes in `Axis.getClosest` method are enough to fix the main bug.

TorsteinHonsi (approved):

PR #18898, opened 2023-04-27
docs/18881-stacked-area-connectnulls

### LINKED ISSUE #18881: Docs: add information about connectNulls behavior for stacked area
From forum support I learned that _"The connectNulls option connects the extreme points between nulls by default, but when using stacking on an area chart, it has a special effect. It treats nulls similar to zeros and instead of sharply breaking the series, they interpolate to 0 for the next tick."_
I think this information should be reflected in docs.

jedrzejruta:
Fixed #18881, added information about different `connectNulls` behaviour in stacked area.


TorsteinHonsi (approved):
Thanks!

PR #18904, opened 2023-04-28
Bugfix/a11y tweaks

oysteinmoseng:
Fixed too low contrast for credits in High Contrast theme.
___
Also improved docs for lang options.


TorsteinHonsi (approved):
Thanks!

PR #18905, opened 2023-04-28
bugfix/18891-treegraph-hidden-labels-hover

### LINKED ISSUE #18891: Treegraph collapsed nodes hover shows tooltips for its hidden children.
#### Expected behaviour
When a node in a tree chart is collapsed, its hidden children should not respond to hover and their tooltip should not be displayed at any time.

#### Actual behaviour
When we hover and move the mouse over the label of a collapsed parent, the tooltips of its hidden children will appear.

https://www.highcharts.com/demo/treegraph-chart

1. Collapse the `Proto Indo-European` node.
2. Hover and move the mouse over its label.

image

#### Product version
Highcharts v11.0.0

Vazonik:
Fixed #18891, invisible data labels in treegraph series reacted to hover.

### DIFF
--- a/samples/unit-tests/series-treegraph/datalabels/demo.css
+++ b/samples/unit-tests/series-treegraph/datalabels/demo.css
@@ -1,0 +1,4 @@
+#container {
+    width: 600px;
+    margin: 0 auto;
+}
--- a/samples/unit-tests/series-treegraph/datalabels/demo.details
+++ b/samples/unit-tests/series-treegraph/datalabels/demo.details
@@ -1,0 +1,8 @@
+---
+ name: Highcharts Demo
+ resources:
+    - https://code.jquery.com/qunit/qunit-2.0.1.js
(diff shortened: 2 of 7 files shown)


TorsteinHonsi (changes requested):
Thanks for the PR!

As far as I can see this issue is specific to treegraph, so I don't think we should do the fix in the core. The problem seems to be that the treegraph series only sets the opacity of the data labels:

https://github.com/highcharts/highcharts/blob/971f85e5567c866526570ddf4a41ffc9d86d7514/ts/Series/Treegraph/TreegraphSeries.ts#L481-L482

Elements with `opacity: 0` will still capture pointer events, so we need to hide them in addition. Try adding `show` and `hide` calls to the fade in/out logic. The hide operation must be added in an animation `complete` callback, otherwise the actual fade will not be visible because the label is already hidden.

TorsteinHonsi (approved):
Thanks!

TorsteinHonsi:
Pending CI

PR #18909, opened 2023-04-29
Added support for axis offset in percent

TorsteinHonsi:
Added support for `xAxis.offset` and `yAxis.offset` in percent.

___

Makes it easier to make centered axes without knowing the pixel size of the chart
Skjermbilde 2023-04-29 kl  08 14 09


TorsteinHonsi:
If https://github.com/highcharts/highcharts/pull/18911 tests are okay, it is a better solution to the same problem, and we can close this.

PR #18910, opened 2023-04-29
Fixed sonification resume issue.

oysteinmoseng:
Fixed sonification issue where resuming after pause ignored the original `onEnd` argument.


TorsteinHonsi (approved):
Thanks!

PR #18918, opened 2023-05-02
Added docs for editable options.

pawellysy:
Added docs for editable options.


karolkolodziej (changes requested):
Questions:
1. Do we need to show the whole class in the documentation?
2. Isn't there an interface for the `item` in `items`? It looks strange:
image

karolkolodziej (approved):

ts/Dashboards/Components/EditableOptions.ts
```diff
@@ -14,19 +14,48 @@
...
         detailedOptions?: Array<DetailedOptions>
+        /**
+         * Relative path to the option, that should be changed in the component.
+         */
         propertyPath?: Array<string>
-        value?: any;
+        /**
+         * Items that should be included in the select element.
+         */
+        items?: Array<ItemOptions>;
```
bre1470, line 45: Naming should be more specific, if it is an option for select elements.

bre1470 (changes requested):

ts/Dashboards/Components/EditableOptions.ts
```diff
@@ -14,19 +14,48 @@
...
         name: string;
+        /**
+         * Type of the editable element.
+         */
         type: ElementType;
+        /**
+         * Detailed options that should be included in the accorion menu.
+         * Available for `nested` type.
+         */
         detailedOptions?: Array<DetailedOptions>
```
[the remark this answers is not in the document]
pawellysy: any suggestion?

ts/Dashboards/Components/EditableOptions.ts
```diff
@@ -98,5 +57,92 @@ class EditableOptions {
...
+         * Type of the editable element.
+         */
+        type: ElementType;
+        /**
+         * Detailed options that should be included in the accordion menu.
+         * Available for `nested` type.
+         */
+        detailedOptions?: Array<DetailedOptions>
+        /**
+         * Relative path to the option, that should be changed in the component.
```
sebastianbochan, line 82: I would add an example of path.

sebastianbochan (changes requested):

ts/Dashboards/Components/EditableOptions.ts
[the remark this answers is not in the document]
pawellysy: maybe `nestedOptions`?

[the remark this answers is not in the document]
pawellysy: let's go with `nestedOptions` 👍

pawelfus (changes requested):

ts/Dashboards/Components/EditableOptions.ts
```diff
@@ -98,5 +57,93 @@ class EditableOptions {
...
+     */
+    export interface NestedOptions {
+        /**
+         * Name of the option that should be displayed.
+         */
+        name: string;
+        /**
+         * whether the option should have a toggle to be enabled or disabled.
+         */
+        allowEnabled?: boolean;
```
pawelfus, line 129: Just realized:
- upper case
- what exactly does this option do?

pawellysy: This option adds a toggle to a header option. 
Screenshot 2023-05-12 at 08 45 35

pawelfus: Hm.. how about "showToggle" name then? 🤔

bre1470 (approved):

ts/Dashboards/Components/EditableOptions.ts
pawellysy: Great name :D

ts/Dashboards/Components/EditableOptions.ts
```diff
@@ -98,5 +57,93 @@ class EditableOptions {
...
     }
 }
+namespace EditableOptions {
+
+    /**
+     * Configuration for a single option in editable options. If type is
+     * `nested` the options are rendered in the accordion menu, with rest of the
+     * options defined in the detailed options.
+     */
+    export interface Configuration {
```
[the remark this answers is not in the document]
pawellysy: Any suggestions for the new name of the `Configuration` interface?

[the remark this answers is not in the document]
pawelfus: > The EditableOptions interface which is in the Component.ts namespace is a bit missleading, or even unnecessary. 

Could you please check it? If it's unnecessary, it should be removed.

> We could move all the contents of this interface to the ComponentOptions interface. IMO, we could delete it from component and leave only the EditableOptions class. (I think it is also a bit redundant, but since it was there in the original code, I left it, when refactoring the Sidebar).

The first part sounds good, the second part confuses me a little, sorry. Do you mean to move the content to ComponentOptions and then delete it?

I know it might not be part of this task, but naming is a bit confusing here. My first thought was `EditableOptions` doesn't sound like the correct name for the Class. Then we could go with simple `NewClassName.Options`, just like in the other Classes

PR #18921, opened 2023-05-02
Added data store selector to the sidebar.

pawellysy:


samples/dashboards/demos/minimal/demo.js
```diff
@@ -1,115 +1,141 @@
...
+    const connector = await board.dataPool.getConnector('Vitamin');
 
-Dashboards.board('container', {
-    connector,
-    editMode: {
-        enabled: true,
-        contextMenu: {
-            enabled: true,
-            items: ['editMode']
+    board.setComponents([{
```
karolkolodziej, line 60: Is that the only way for now to use the data connectors with the board?

karolkolodziej (changes requested):
Todo:

- [ ] Tests if the list is correctly generated
- [ ] This should not be called `store` but `Data connector`
image
- [ ] By default some connector should be selected

ts/Dashboards/Components/EditableOptions.ts
```diff
@@ -95,7 +96,20 @@ class EditableOptions {
     }
 
     public getOptions(): (Array<EditableOptions.Configuration>) {
-        return this.component.options.editableOptions;
+        const options = this.component.options.editableOptions;
+        for (let i = 0; i < options.length; i++) {
```
sebastianbochan, line 100: Please assign length into variable, it will increase performance.

Eg: `for (let i = 0, iEnd = options.length; i < iEnd; ++i) {`

samples/dashboards/demos/minimal/demo.js
pawellysy: For now, yes 😓

sebastianbochan (approved):

bre1470 (approved):

PR #18923, opened 2023-05-02
bugfix/18922-scrollbar-track-click-event

### LINKED ISSUE #18922: Transparent scrollbar track in V11 doesn't fire click events on it
Since V11 update, the scrollbar track now has a transparent background instead of a white one, which makes it impossible for it to track click events on it.

#### Expected behaviour
Clicking on scrollbar track should fire the click event, and scroll the scrollbar.

#### Actual behaviour
Clicking on scrollbar track is not registered. Chart background click event is fired instead.

https://jsfiddle.net/BlackLabel/54g60mcd/

#### Product version
Stock v11
Last working version 10.3.3

kamil-musialowski:
Fixed #18922, scrollbar track did not catch click events.

### DIFF
--- a/css/highcharts.css
+++ b/css/highcharts.css
@@ -829,7 +829,7 @@ circle.highcharts-legend-nav-inactive { /* tracker */
 }
 
 .highcharts-scrollbar-track {
-    fill: none;
+    fill: rgba(255, 255, 255, 0.001);
     stroke: var(--highcharts-neutral-color-20);
     stroke-width: 1px;
 }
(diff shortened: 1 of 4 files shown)


TorsteinHonsi:
#### Workaround

Set `trackBackgroundColor` to 0.001.

[Live jsFiddle demo](https://jsfiddle.net/highcharts/hwfc2vq1/)

TorsteinHonsi (changes requested):
Thanks for the quick handling! Instead of making a bad choice of default value and then throw in a workaround for it, I think we should make a good choice for the default value in the first place. How about this?

https://github.com/highcharts/highcharts/blob/df9b7029f0d6b2cab574e9a50cea2cd69bd30a13/ts/Stock/Scrollbar/ScrollbarDefaults.ts#L227

👇 

```ts
trackBackgroundColor: 'rgba(255,255,255,0.001)',
```

TorsteinHonsi (approved):

PR #18924, opened 2023-05-02
Docs: Fixed conflicts

bre1470:
- Fixed conflicts with product defaults.
- Fixed conflicts with reflected options types.


TorsteinHonsi (approved):
Thanks!

PR #18926, opened 2023-05-02
db/fix-demos-typing

bre1470:
CI tests fail because of failure in main branch. This branch fixes it.


karolkolodziej (approved):

karolkolodziej:
The code looks fine but some CI tasks are still failing, is it related?

bre1470:
Yes, I think so. I tested locally the commands and its fine. So the fail comes from main branch and comparison can not run through.

pawellysy (approved):

PR #18930, opened 2023-05-03
Fixed #18928, tooltip not showing when one chart hidden

### LINKED ISSUE #18920: Multiple SVG filters with the same id when multiple charts
#### Actual behaviour
Each drop-shadow filter has the same id (when we create multiple charts we will get multiple elements with the same ID).

**Demo:** https://jsfiddle.net/BlackLabel/1qo079ue/
Two charts, each one of them has a filter with the same id (see console log).

#### Product version
Highcharts 11

Reported by the forum user in this topic: https://www.highcharts.com/forum/viewtopic.php?f=9&t=50781&p=185907#p185907

### LINKED ISSUE #18928: Tooltip does not show if there are multiple charts and one is hidden
#### Actual behaviour
If there is more than one chart, and `display: none` is set for one of the charts, the tooltip shows only for the first chart.

**Demo:** https://jsfiddle.net/BlackLabel/jr82ognp/
1. Switch tabs
2. Hover over a series (there's no tooltip).

#### Product version
Regression from Highcharts 11 (everything worked well for <= 10.3.3)

TorsteinHonsi:
Fixed #18928, a regression in v11 causing the tooltip to not display in subsequent charts when the first chart in the web page was hidden.
___
Closes #18920

### DIFF
--- a/samples/unit-tests/svgrenderer/shadows/demo.js
+++ b/samples/unit-tests/svgrenderer/shadows/demo.js
@@ -40,7 +40,7 @@ QUnit.test('Series shadows', function (assert) {
 
     assert.strictEqual(
         chart.series[0].graph.attr('filter'),
-        'url(#drop-shadow-0)',
+        `url(#highcharts-drop-shadow-${chart.index})`,
         'Shadows should be updated when old options defined as object and new as boolean (#12091).'
     );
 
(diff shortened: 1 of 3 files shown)


jakubSzuminski (approved):
Looks good to me! Everything works now as expected: https://jsfiddle.net/BlackLabel/x3Lwtja8/

I'll reply to the forum users who struggled with this issue (3 people reported this problem) that the fix can be found here.

TorsteinHonsi:
Thanks @jakubSzuminski ! We will also release a maintenance release soon. Probably next week.

PR #18939, opened 2023-05-04
Fixed #18001, area opacity for HTML colors.

### LINKED ISSUE #18001: area charts/series does not accept named colors
#### Expected behaviour
I expect no difference between color 'red' and '#FF0000' when using a 'filled area' series

#### Actual behaviour
Using color red fills the area without using the fillOpacity, #FF0000 acts as expected.

[See my fiddle](https://jsfiddle.net/HansR/w6gjt1rf/)
Please change between the commented 'red' and the un commented hex value.

#### Product version
Highcharts "10.3.1" (see log of the fiddle)
NOTE: it used to work before with the color names and the hex values hence the bugreport: a sudden change means I have to change all charts / graphs / series. It is very confusing.

#### Affected browser(s)
Everywhere.

jakubSzuminski:
Fixed #18001, `fillOpacity` did not work for the area when the color was not HEX or RGBA, but an HTML color like _red, blue_, etc.

**Before fix:** https://jsfiddle.net/BlackLabel/8s95Lamo/
**After fix:** https://jsfiddle.net/BlackLabel/jvykotx1/

I've added an `if` which is checking whether the `color(series.color).setOpacity(fillOpacity) == series.color` because if that's true that means that the `series.color` is an HTML color (if it's HEX or RGBA, there will be a difference after setting the opacity).

I've added 2 tests which are checking this case, but unfortunately, I had to add an if statement outside one assertion as the test would pass in Utils, but not in Karma (Karma would have `fill: rgba(0, 0, 255, 0.25)` instead of `fill: blue, fill-opacity: 0.25` which is the case for Utils).

### DIFF
--- a/samples/unit-tests/series-area/fillopacity/demo.js
+++ b/samples/unit-tests/series-area/fillopacity/demo.js
@@ -11,6 +11,11 @@ QUnit.test('Fill opacity zero (#4888)', function (assert) {
         series: [
             {
                 data: [1, 3, 2, 4]
+            },
+            {
+                data: [3, 2, 5, 2],
+                color: 'blue',
+                fillOpacity: 0.25
             }
         ]
     });
@@ -20,4 +25,18 @@ QUnit.test('Fill opacity zero (#4888)', function (assert) {
         'rgba(44,175,254,0)',
         'Fill opacity should be set'
     );
+
+    const areaFill = chart.series[1].area.element.getAttribute('fill').replace(/ /g, '');
+    assert.ok(
+        areaFill === 'blue' || areaFill === 'rgba(0,0,255,0.25)',
+        'There should be support for HTML color names'
+    );
+
+    if (areaFill === 'blue') {
+        assert.strictEqual(
+            chart.series[1].area.element.getAttribute('fill-opacity'),
+            '0.25',
+            'There should be opacity set for the HTML color names'
+        );
+    }
 });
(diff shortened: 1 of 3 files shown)


TorsteinHonsi:
As noted in the issue report, this was actually a regression in v5.0.0, as it did work in v4.2. I tried to find out why it stopped working. Turns out that the current solution (with `setOpacity`) was used until v4.2.0, when it was [replaced by setting the fill-opacity](https://github.com/highcharts/highcharts/commit/bbcfdfe766abffc5f88610e24128dddfdabdb8d5#diff-0c4f572e04a7d190f38166608672d43079ae5fab74158d0c084ed59b520105b4R16251). The commit message states that part of the mission with the commit was to allow semi-transparent fill with named colors. Which is exactly what #18001 is about. 

So the question is why this was broken again in v5.0.0. I tried to do a `git bisect`, and I am not able to fully pinpoint it. But I suspect this was simply a merge error that brought back the old functionality. 

The conclusion is, we should do this simple and just set `fill-opacity` in any circumstance - named colors or not. At least let's try it and see how the visual tests go.

jakubSzuminski:
Okay, thanks! I'm on it.

jakubSzuminski:
@TorsteinHonsi I think I commited the changes as you requested (`fill-opacity` is always added if no `fillColor` on series is specified). But something strange shows up in the Highcharts Review Tool: "Uh oh, something went wrong. The specified bucket does not exist". What should I do about this? Should I change something in my code?

PR #18951, opened 2023-05-05
bugfix/18925-map-drilldown-geojson

### LINKED ISSUE #18925: Custom GeoJSON Drillable Maps Do Not Render Correctly
I believe this is caused maybe by mapView not being set or not having a default? This has not been an issue previously and I would imagine that the lack of default mapView would break even the initial map. However, the initial map loads fine. We are using custom ArcGIS maps to handle specific state-defined areas as well as including US Territories and their sub-state areas that the HighCharts map collection does not contain.

#### Expected behaviour
Using custom GeoJSON maps created from ArcGIS clicking on a drillable map point should load another map of that region selected with higher detail. The parent and child maps are GeoJSON created from our ArcGIS system. An example would be loading the USA national map and clicking on Texas. This causes the Texas county level map to load and show data as a choropleth.

#### Actual behaviour
Using custom GeoJSON maps and clicking on Texas shows a quick animated zoom to the county level map of Texas then it immediately zooms back out to the national view with Texas broken up by counties but no choropleth. Clicking on the drill up navigation completely breaks the view of the USA map and no further action can be made in the map (no choropleth data and no clickability).
After clicking on TX:
image

Unbale to provide demo that works as ArcGIS maps are not available externally.
However, code is provided here:
https://jsfiddle.net/wergeld/uaowedj7/

#### Product version
HighCharts v11
HighMaps v11

#### Affected browser(s)
Chrome, Edge.

hubertkozik:
Fixed #18925, map drill down was broken for GeoJSON maps.


TorsteinHonsi (approved):
Thanks!

PR #18969, opened 2023-05-09
bugfix/18863-last-point-ordinal-axis

### LINKED ISSUE #18863: Unable to zoom till the very last date point
#### Expected behaviour
The very last date point, which is `31 March` in this case should be visible when zooming to very right-end.

#### Actual behaviour
That very last date point is dropped in the zoomed view.

https://jsfiddle.net/jpzn082e/

#### Product version
Highstock `v10.3.3`

#### Affected browser(s)
Chrome

kamil-musialowski:
Fixed #18863, last point in zoomed ordinal axis was not visible.

### DIFF
--- a/samples/unit-tests/axis/ordinal/demo.js
+++ b/samples/unit-tests/axis/ordinal/demo.js
@@ -46,6 +46,36 @@ QUnit.test('Ordinal general tests.', function (assert) {
             'ordinal is enabled (#12716).'
     );
 
+    assert.strictEqual(
+        chart.xAxis[0].toValue(
+            chart.series[0].data[chart.series[0].data.length - 1].plotX,
+            true
+        ),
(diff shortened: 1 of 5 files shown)


kamil-musialowski:
Due to the fact, that the `returnValue` might be not be a whole number, after discussing the case with @pawelfus, I have decided to change the test to `assert.close` instead of fighting with the 0.0002 px return difference.

https://github.com/highcharts/highcharts/blob/55b93b7dc89f7a0b4c1c2faa9d52cc57c5c31c34/ts/Core/Axis/Axis.ts#L874

pawelfus (approved):
Thanks!

Yes, I think it's not worth sacrificing the performance to use `correctFloat()` in one of the most used methods in core (`axis.translate()`), just to pass the test. If there will show up a real-case issue, that forces us to add it, we will add it.

PR #18973, opened 2023-05-10
bugfix/18960-dumbbell-data-labels

### LINKED ISSUE #18960: Data labels for Dumbell of multiple series are collecting at the same serie.
Am not sure what has changed internally here, but it looks like an issue which could be more general than just this case I observed.

Highcharts GPT was particularly useful in creating an MRE!

#### Expected behaviour
Data labels are attached to their respective series

#### Actual behaviour
All data labels are attached to the same series

You can downgrade the imported scripts from 11 to 10 to see the previous behavior.

https://jsfiddle.net/kluvin/a1Lg4k75/

#### Product version
11.0.0 >
 
#### Affected browser(s)
FF, not tested in Chrome.

jakubSzuminski:
Fixed #18960, a regression in v11, wrong placement for data labels in dumbbell series.

The issue was the change of `ColumnRangeSeries.prototype.translate()` function which was used in `DumbellSeries` as a `setShapeArgs` function.

**Demo before fix:** https://jsfiddle.net/kluvin/a1Lg4k75/
**Demo after fix:** https://jsfiddle.net/BlackLabel/e9fa0ck8/

### DIFF
--- a/samples/highcharts/demo/dumbbell/demo.js
+++ b/samples/highcharts/demo/dumbbell/demo.js
@@ -93,7 +93,10 @@ Highcharts.chart('container', {
 
     series: [{
         name: 'Life expectancy change',
-        data: data
+        data: data,
+        dataLabels: {
+            enabled: true
+        }
     }]
 
 });
--- a/samples/highcharts/demo/dumbbell/demo.js
+++ b/samples/highcharts/demo/dumbbell/demo.js
@@ -93,10 +93,7 @@ Highcharts.chart('container', {
 
     series: [{
         name: 'Life expectancy change',
-        data: data,
-        dataLabels: {
-            enabled: true
-        }
+        data: data
     }]
 
 });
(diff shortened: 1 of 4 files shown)


karolkolodziej (approved):
Please add test as Torstein mentioned

jakubSzuminski:
This dumbbell sample should pick up on this issue as the `dataLabels` are now shown.

samples/highcharts/demo/dumbbell/demo.js
TorsteinHonsi, line 1: This is the main demo that we display in our web pages, https://www.highcharts.com/demo/dumbbell

I don't think we should add data labels here, as it kind of clutters up the nice clean appearance of the chart. A Google Images search confirms that this chart type is more often used without data labels. We have a sample for `highcharts/series-dumbbell/styled-mode-dumbbell` that is not part of the website demos, we could add the labels there instead.

TorsteinHonsi (changes requested):

samples/highcharts/demo/dumbbell/demo.js
jakubSzuminski: You're right. I'm sorry, my mistake. I've now enabled dataLabels for the `highcharts/series-dumbbell/styled-mode-dumbbell` sample.

PR #18975, opened 2023-05-10
Db/kpi default style

pawellysy:
Updated the default style of KPI Component.

### DIFF
--- a/ts/Dashboards/Components/KPIComponent.ts
+++ b/ts/Dashboards/Components/KPIComponent.ts
@@ -115,7 +115,8 @@ class KPIComponent extends Component {
             ].join(' '),
             style: {
                 boxSizing: 'border-box',
-                textAlign: 'center'
+                textAlign: 'center',
+                'min-height': '150px'
             },
             thresholdColors: ['#f45b5b', '#90ed7d']
         }
--- a/ts/Dashboards/Components/KPIComponent.ts
+++ b/ts/Dashboards/Components/KPIComponent.ts
@@ -118,6 +118,7 @@ class KPIComponent extends Component {
                 textAlign: 'center',
                 'min-height': '150px'
             },
+            minFontSize: 30,
             thresholdColors: ['#f45b5b', '#90ed7d']
         }
     );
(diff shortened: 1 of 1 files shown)


sebastianbochan (changes requested):

ts/Dashboards/Components/KPIComponent.ts
```diff
@@ -339,8 +357,19 @@ class KPIComponent extends Component {
      */
     private updateSize(width: number, height: number): void {
         this.updateTitleSize(width, height);
-        this.value.style.fontSize = 0.2 * Math.min(width, height) + 'px';
-        this.subtitle.style.fontSize = 0.08 * Math.min(width, height) + 'px';
+        // If there is no chart, make the font size  bigger.
+        const noChartMultiplier = (this.chart ? 1 : 1.8);
```
karolkolodziej, line 361: Same here, any reason behind these numbers?

karolkolodziej (changes requested):
Is that all that should be fixed in this PR? What about the `square component`?

ts/Dashboards/Components/KPIComponent.ts
```diff
@@ -322,11 +324,27 @@ class KPIComponent extends Component {
      */
     private updateTitleSize(width: number, height: number): void {
         if (this.titleElement) {
-            this.titleElement.style.fontSize =
-                0.1 * Math.min(width, height) + 'px';
+            this.titleElement.style.fontSize = this.getFontSize(
+                width,
+                height,
+                0.08 * (this.chart ? 1 : 1.7)
```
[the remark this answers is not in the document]
pawellysy: those variables were there previously. For now, we need a design. 
The square component is not that easy to be added. Right now, I've added some arbitrary values, that I figured out randomly. When designs will be there, I'll set the values to be better defined.

samples/dashboards/demos/component-kpi/demo.js
```diff
@@ -145,6 +145,6 @@ function setValues() {
 }
 
 // Update the data every second
-setInterval(() => {
-    setValues();
-}, 1000);
+// setInterval(() => {
```
karolkolodziej, line 148: Please restore

karolkolodziej (approved):

sebastianbochan (approved):

PR #18976, opened 2023-05-10
bugfix/17861-zoomtype-update

### LINKED ISSUE #17861: Setting zoomType in update broke in version 10.2.1
#### Expected behaviour
The zoomType is updated and zoom is enabled.

#### Actual behaviour
The update doesn't seem to have any effect.

https://jsfiddle.net/3b1eqpkw/

#### Product version
10.2.1

#### Affected browser(s)
All? Only tested in chromium based browsers.

I understand that zoomType is now deprecated in favor of zooming (and switching to that property is a workaround), but this is a regression nonetheless.

jakubSzuminski:
Fixed #17861, updating the deprecated properties for zoom didn't work (`chart.zoomType`, `chart.zoomKey`, `chart.resetZoomButton`, `chart.pinchType`, `chart.singleTouch`).

**Demo after fix:** https://jsfiddle.net/BlackLabel/61zrypqv/

### DIFF
--- a/ts/Core/Chart/ChartDefaults.ts
+++ b/ts/Core/Chart/ChartDefaults.ts
@@ -492,7 +492,7 @@ const ChartDefaults: ChartOptions = {
      *
      * @type       {string}
      * @default    {highcharts} undefined
-     * @default    {highstock} x
+     * @default    {highstock} undefined
      * @since      3.0
      * @product    highcharts highstock gantt
      * @deprecated
@@ -1028,7 +1028,6 @@ const ChartDefaults: ChartOptions = {
      * @product    highcharts highstock gantt
      * @deprecated
      */
-    zoomBySingleTouch: false,
 
     /**
      * Chart zooming options.
--- a/samples/unit-tests/chart/zoomtype/demo.js
+++ b/samples/unit-tests/chart/zoomtype/demo.js
@@ -372,7 +372,7 @@ QUnit.test('Zooming scatter charts', function (assert) {
     });
 
     assert.strictEqual(
-        chart.options.chart.zooming.type,
+        chart.zooming.type,
         'xy',
         'There should be support for updating the deprecated zoomType (#17861)'
     );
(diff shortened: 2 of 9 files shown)


pawellysy (changes requested):
Looks good! Test is missing though 😅

TorsteinHonsi (changes requested):
Thanks! I'm concerned about the amount of added code though. All because we were reorganizing the option names. 

The problem is that we are rewriting `options.chart.zooming`, so in the next run it takes precedence over the existing options. So I think we can fix this more cheaply by not modifying the options.

Instead, can you try adding a property directly on the chart instance, `chart.zooming`? In the `init` function, this property would always be set (whether it is a new chart or updated). And all places where we currently read `chart.options.zooming.xxxx`, it would be changed to `chart.zooming.xxx`.

TorsteinHonsi:
Note to self: Pending refactor.

ts/Core/Chart/ChartDefaults.ts
```diff
@@ -492,7 +492,7 @@ const ChartDefaults: ChartOptions = {
      *
      * @type       {string}
      * @default    {highcharts} undefined
```
jakubSzuminski, line 494: I've removed the default values for the deprecated options so that the `setZoomOptions` function is much shorter. Now, `chart.options.zoomType` is set only when the users set it (it is not always taken from `ChartDefaults`).

ts/Core/Chart/ChartDefaults.ts
```diff
@@ -492,7 +492,7 @@ const ChartDefaults: ChartOptions = {
...
-     * @default    {highstock} x
+     * @default    {highstock} undefined
```
jakubSzuminski, line 495: I've removed the default value for this deprecated option so that the setZoomOptions function is much shorter. Now, chart.options.zoomType, chart.options.pinchType is set only when the users set it (it is not always taken from ChartDefaults).

TorsteinHonsi (approved):

Issue #18839, opened 2023-04-17
Allow rounded corners in pyramid charts


bcmrcool:
# Description of the feature

We'd like to have rounded corners in highchart's pyramid chart as in the following image:

![Screen Shot 2023-04-17 at 3 41 59 PM](https://user-images.githubusercontent.com/2309225/232593482-4d9bf8ff-9d28-4ff0-b67e-2d75567b372b.png)

It currently looks like:
![Screen Shot 2023-04-17 at 3 43 18 PM](https://user-images.githubusercontent.com/2309225/232593744-964fd9e9-2623-4e21-8922-bdf5e07fd27d.png)

I've looked around and reached out to support to no avail.

# Library related to the feature

Highcharts
# Proof of Concept/Live example for the feature

![Screen Shot 2023-04-17 at 3 41 59 PM](https://user-images.githubusercontent.com/2309225/232593482-4d9bf8ff-9d28-4ff0-b67e-2d75567b372b.png)

----

**You can vote for this feature by adding a thumbs-up reaction to this post.**

pawellysy:
Hello @bcmrcool, thanks for your input!

As a workaround you can use the following drop-in modification: https://jsfiddle.net/BlackLabel/aqd2yn0v/

bcmrcool:
> Hello @bcmrcool, thanks for your input!
> 
> As a workaround you can use the following drop-in modification: https://jsfiddle.net/BlackLabel/aqd2yn0v/

Hi @pawellysy , it doesn't seem to work in my case. Do you know of a resource to help me better understand `shapeArgs`?

An entire section of my pyramid is chopped off. For what it's worth I do have an additional letter `M` in my shape args however I'm not really sure what that translates to.

bcmrcool:
hi @pawellysy, for reference here I have a jsfiddle here which is exactly how I've got my chart setup: https://jsfiddle.net/osdm2je8/74/, really appreciate the help

pawellysy:
Hi @bcmrcool, thanks for your feedback!
In highcharts, `shapeArgs` property defines the shape, that each point should have. In case of paths (like most points are created) the `d` property comes as a SVG Path line command. The docs of the SVG Path [you can find here](https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorial/Paths).

In case of your example, the snippet I shared is only a PoC, which shows that this is achievable, but not really a fully working implementation.  We'll look into it though

bcmrcool:
[...]

PR #18979, opened 2023-05-11
feature/18839-funnel-pyramid-rounded-corners

Vazonik:
Added `borderRadius` option support for funnel and pyramid series. See #18839.
___
TODO:
- [x] Add the ability to round the boundaries of funnel points.
- [x] Improve rounding to exact circles inscribed in the angles.
- [x] Handle rounding scope for series edges only (the `borderRadius.scope: 'stack'` option).
- [x] Add tests & samples.

### DIFF
--- a/samples/highcharts/plotoptions/funnel-border-radius/demo.css
+++ b/samples/highcharts/plotoptions/funnel-border-radius/demo.css
@@ -1,0 +1,39 @@
+#outer-container {
+    min-width: 300px;
+    max-width: 600px;
+    margin: 1em auto;
+}
+
+#outer-container .button-group {
+    padding-right: 2em;
+    padding-bottom: 4px;
+}
+
+#outer-container button {
+    padding: 1em;
+    border: 0;
+    border-radius: 2px;
+    font-size: 12px;
+    cursor: pointer;
+}
+
+#container {
+    height: 400px;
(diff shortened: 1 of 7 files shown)


TorsteinHonsi (changes requested):
Thanks, nice work! I bet you had some great fun with that math @Vazonik :)

While the math looks perfect, we need to make the options consistent with [column and pie borderRadius](https://api.highcharts.com/highcharts/series.column.borderRadius).

* Instead of a separate `borderRadiusMode` option, use [borderRadius.scope](https://api.highcharts.com/class-reference/Highcharts.BorderRadiusOptionsObject#scope). Actually I think the existing `point` and `stack` values make perfect sense also for funnels and pyramids, even though the stack here is technically the series. But in order not to complicate things, I think we should just use `point` and `stack`.
* Percentage values for `borderRadius` breaks the funnel/pyramid. It should either fail gracefully, or work. I don't think it is much work to get the percentages working. We just need to decide what 100% means. The total width of the pyramid/funnel make sense, alternatively the least of the width and the height.
* And obviously `borderRadius.radius` must work. See for example the [SolidGaugeSeries](https://github.com/highcharts/highcharts/blob/b710960120e76346bec103b8bb7210dc86aee9be/ts/Series/SolidGauge/SolidGaugeSeries.ts#L218-L222) for how we work with the object and shorthand option.

TorsteinHonsi (approved):

PR #18981, opened 2023-05-11
bugfix/18110-softmin-max-padding

### LINKED ISSUE #18110: softMin causes maxPadding to be ignored for certain cases
#### Expected behaviour

There should be a padding between the serie maximum and the chart maximum that can be controlled with `maxPadding`

#### Actual behaviour

`maxPadding` is not respected, causing the serie to end up to close to the chart edge

https://jsfiddle.net/zjx86fvm/50/

- If `softMin` is replaced with `min`, the padding is added
- If the series have a higher variance (e.g replace the first data point with 220), the padding is added

#### Product version
Highcharts 10.2.0

jakubjanuchta:
Fixed #18110, `yAxis.maxPadding` was ignored when `yAxis.softMin` was set.


TorsteinHonsi (approved):
Thanks!

PR #18990, opened 2023-05-15
db/datagrid-default-theme

sebastianbochan:


bre1470 (approved):

ts/DataGrid/DataGridDefaults.ts
```diff
@@ -32,7 +32,7 @@ import type DataGridOptions from './DataGridOptions';
 /** @internal */
 const DataGridDefaults = {
     /** @internal */
-    cellHeight: 25,
+    cellHeight: 49,
```
karolkolodziej, line 35: As I see in the design only the header row should have a more significant height.
image

Also, could you explain why 49?

karolkolodziej (changes requested):

ts/DataGrid/DataGridDefaults.ts
sebastianbochan: 49 is applied only for header's cells.

That works from the design of data-grid, so I did not refactor the core of it.

css/datagrid.css
```diff
@@ -129,9 +129,9 @@
 }
 
 .hc-dg-cell-input {
-    width: calc(100% + 16px);
-    margin: 0 -8px;
-    padding: 0 6px;
+    width: calc(100% + 24px);
```
pawellysy, line 132: Could we replace this `width: calc(...)` with `flex: 1`?

pawellysy (changes requested):
Minor request

css/datagrid.css
sebastianbochan: Not really, it causes problems with width and highlighting input.

pawellysy (approved):

PR #19003, opened 2023-05-16
db/dashboards-math

bre1470:
- [x] update demo
- [x] fix DataGrid
- [x] fix DataPool
- ~[ ] integrate dashboard gulp tasks~ later


samples/dashboards/demos/mathmodifier-formula/demo.js
```diff
@@ -0,0 +1,154 @@
+const container1 = document.querySelector('#container');
+const container2 = document.querySelector('#container2');
+const cursor = new Dashboards.DataCursor();
+const MathModifier = Dashboards.DataModifier.types.Math;
```
pawelfus, line 4: I don't think we want to show these methods and Classes, do we? When I said we need a public demo for MathModifier, I meant MathModifier in Dashboards/DataGrid, not a copy of the demo from internal data-tools.

pawelfus (changes requested):

samples/dashboards/demos/mathmodifier-formula/demo.js
bre1470: Then I suggest to extend one of the existing Dashboards/DataGrid demos. Or do you like to have a new Dashboards demo?

samples/dashboards/demos/mathmodifier-formula/demo.html
```diff
@@ -0,0 +1,17 @@
+<script src="https://code.highcharts.com/dashboards.js"></script>
+<script src="https://code.highcharts.com/dashboards-math.js"></script>
```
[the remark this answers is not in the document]
TorsteinHonsi: This is fine by me: 
`<script src="https://code.highcharts.com/dashboards/modules/math-modifier.js"></script>`

It is consistent with other packages like Stock and Maps:
`https://code.highcharts.com/stock/modules/exporting.js`

> How about a plugins folder? modules naming has become confusing with modern JavaScript, because the modules are not ES modules.

Yes, that's a good point. On the flip side it will not be consistent with Highcharts, Stock and Maps then. Also it is debatable whether it is a plugin - these modules are actually part of the product, while a plugin typically is either third party or at least not part of the product. Add-ons has the same problem.

pawelfus (changes requested):
A few things about the demo:
- could you merge main branch in, and use connector options in Dashboards?
- Drag&drop + editable Grid is  broken, see gif below:
dragdrop

pawelfus (approved):
Approved 👍

TorsteinHonsi (approved):
Thanks!

TorsteinHonsi:
@bre1470 There's one failing test though...

oysteinmoseng (approved):

PR #19006, opened 2023-05-16
Tools: Changed closure compiler utilization

bre1470:
There are significant changes in file sizes.

To-do:
- [x] Update Google Closure Compiler
- [x] Replace `scripts-compile`
- [x] Test minified `code/highcharts.js`


TorsteinHonsi:
What happened here? This is the opposite of what we expected 🤔 
Skjermbilde 2023-05-23 kl  18 58 06

bre1470:
I was hoping that you might have an idea. When comparing both builds, I noticed that many of the shortcut variables got one char longer. So maybe it does not work as well anymore?

TorsteinHonsi:
This PR now does two things:
* Changes the TS target from ES2016 to ES2020
* Changes the version of Closure Compiler from 2020 to 2023.

Here's a breakdown of some file size changes for each step. These are the file sizes reported by Finder in the file system:

highcharts.js
* TS target = ES2016, CC version 20200504: 286 948 bytes
* TS target = ES2016, CC version 20230502: 289 859 bytes
* TS target = ES2020, CC version 20230502: 288 979 bytes

highcharts-gantt.js
* TS target = ES2016, CC version 20200504: 394 495 bytes
* TS target = ES2016, CC version 20230502: 397 981 bytes
* TS target = ES2020, CC version 20230502: 397 045 bytes

modules/data.js
* TS target = ES2016, CC version 20200504: 20 176 bytes
* TS target = ES2016, CC version 20230502: 16 130 bytes
* TS target = ES2020, CC version 20230502: 16 130 bytes

The `data.js` changes are not representative. Apparently the great reduction in file size is because of a polyfill, probably because we were using some language feature that is not supported by ES2016.

That leaves `highcharts.js` and `highcharts-gantt.js`. We see that upgrading the Closure Compiler makes the bundle size go up by ~3 kB. Then by changing the build target to ES2020 reduces the bundle by ~1 kB. This is not fully in line with the file size changes reported by CI, but nevertheless there is something funny with Closure Compiler. It's not a big deal, but I still wonder if there are more optimizations we can do to make the 2023 version compress on par with the 2020 version...

bre1470:
@TorsteinHonsi Thank you for the testing!

I checked now every documented option with before/after JSON table, and added in the last commit another option to safe some bytes in annotations.

Next I will take another look into CI file size comparison. It looks like it is not producing reliable numbers after changes.

bre1470:
If the out language is not defined, it safes us about 500 kB compared to the previous changes. The old closure compiler is still unbeaten by about 0.5-2.5 kB less. So my suggestion is to go with the old closure compiler and ES2016 for now, but keep the changes in the gulp tasks to have it a bit faster.

bre1470:
I had to update the TypeDoc dependency and now the theme has changed. Please review.

bre1470:
Screenshot:
Screenshot 2023-06-23 at 12 52 01

bre1470:
Screenshot:
Screenshot 2023-06-23 at 12 53 34

pawelfus (approved):
To be honest, I prefer this much more. It looks more like Highcharts, I would even say it has a feature I was missing (and I didn't know until now): Class and Options are available at the same level (like you have: Board Namespace and Board Class).

Note: Reviewed only Dash docs

oysteinmoseng (approved):
Looks good to me!

Edit: I do notice there are some descriptions missing, but the API docs are not fully done yet right?

TorsteinHonsi (approved):

PR #19015, opened 2023-05-17
db/fix-import-export-dashboard

sebastianbochan:
- [x] - Maximum call on stack when serialize HTML Component
- [ ] - Replace `Dashboard.importLocal()` with `Dashboard.Board.importLocal()`


sebastianbochan:
Closed, because it is not still valid after refactoring of connectors.

PR #19018, opened 2023-05-18
db/fixed-lang-options

sebastianbochan:


pawellysy (approved):
Looks Good!

karolkolodziej (approved):
Looks good but I have the same questions as Pawel.
Is the demo going to be a separate PR from someone else?

bre1470 (approved):

samples/dashboards/cypress/options-lang/demo.html
```diff
@@ -0,0 +1,7 @@
+<!-- <script src="https://code.highcharts.com/dashboards.js"></script>
+<script src="https://code.highcharts.com/highcharts.js"></script>
```
karolkolodziej, line 2: Remove?

karolkolodziej (approved):
Shouldn't we also have a demo with that for the distribution? I mean not only the `.mjs`

samples/dashboards/cypress/options-lang/demo.html
sebastianbochan: Done, but it was not a part of this task ;)

sebastianbochan:
> Shouldn't we also have a demo with that for the distribution? I mean not only the `.mjs`

We can create a separate ticket for that.

PR #19020, opened 2023-05-19
bugfix/12063-packedbubble-bad-positions-after-width-update

### LINKED ISSUE #12063: Bubbles get stuck in packed bubble chart
#### Expected behaviour
All chart bubbles should be moving along with outer big bubble in UI.

#### Actual behaviour
This UI issue is random in nature.
image

#### Product version

Highcharts version 7.2.0

hubertkozik:
Fixed #12063, packed bubble parent nodes had a bad position after width update.

### DIFF
--- a/ts/Series/PackedBubble/PackedBubbleSeries.ts
+++ b/ts/Series/PackedBubble/PackedBubbleSeries.ts
@@ -554,9 +554,9 @@ class PackedBubbleSeries extends BubbleSeries {
         this.calculateParentRadius();
         if (
             this.parentNode &&
-            this.parentNode.plotX &&
-            this.parentNode.plotY &&
-            this.parentNodeRadius
+            defined(this.parentNode.plotX) &&
+            defined(this.parentNode.plotY) &&
+            defined(this.parentNodeRadius)
         ) {
             parentAttribs = merge({
                 x: this.parentNode.plotX -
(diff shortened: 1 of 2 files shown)


ts/Series/PackedBubble/PackedBubbleSeries.ts
```diff
@@ -539,35 +539,40 @@ class PackedBubbleSeries extends BubbleSeries {
...
-            height: (this.parentNodeRadius as any) * 2
-        }, parentOptions);
-        if (!(this.parentNode as any).graphic) {
-            this.graph = (this.parentNode as any).graphic =
-                chart.renderer.symbol((parentOptions as any).symbol)
-                    .add(this.parentNodesGroup);
+        if (
+            this.parentNode &&
+            this.parentNode.plotX &&
+            this.parentNode.plotY &&
```
pawelfus, line 558: When checking the number-type value, always use `H.defined()`, so `0`-values are correctly handled. For example `plotX = 0` and `plotY = 0` (top left corner) would be skipped forever if that happens (unlikely, but still)

pawelfus (changes requested):
Nice, minor change only

pawelfus (approved):
Thanks!

TorsteinHonsi (approved):
Thanks!

PR #19022, opened 2023-05-19
db/editMode-docs

pawellysy:
Added Initial docs to EditMode.


pawellysy:
Will be continued here: https://github.com/highcharts/highcharts/pull/19058

PR #19023, opened 2023-05-19
bugfix/18974-cumulative-datagrouping-approximation

### LINKED ISSUE #18974: Highstock cumulative line series y axis labels are wrong
#### Expected behaviour
If i have lots of line data (1000+ data points) and want to view the cumulative line plot, the y axis labels should be correct independent of the chart width.

#### Actual behaviour
The y axis value is displayed incorrectly if the chart is not wide enough.

This example demonstrates the y axis labels incorrectly displayed for cumulative line data with values 1 to 1000.

It shows three charts rendering the same data  but with different widths.
The top y axis labels should all be about 500k (sum of 1 to 1000) but only the first chart (2000px wide) actually displays the correct label.
The second and third chart (1000px and 200px wide) display incorrect y axis labels.

https://jsfiddle.net/tgaepxqy/1/

#### Product version
Highstock JS v11.0.1 (2023-05-08)

#### Affected browser(s)
Chrome (112.0.5615.137), Firefox (112.0.2), Safari (16.4)

jedrzejruta:
Fixed #18974, wrong approximation set for cumulative sum when `dataGrouping` was enabled.


pawelfus (approved):
Thanks!

PR #19029, opened 2023-05-22
db/dragDrop-enabled

karolkolodziej:
DragDrop did not work when empty options.

DradDrop is tested in the cypress. No need for additional tests.

### DIFF
--- a/ts/Dashboards/Actions/DragDrop.ts
+++ b/ts/Dashboards/Actions/DragDrop.ts
@@ -71,6 +71,8 @@ class DragDrop {
         this.editMode = editMode;
         this.options = merge(DragDrop.defaultOptions, options);
 
+        this.enabled = !!this.options.enabled;
+
         this.mockElement = createElement(
             'div',
             { className: EditGlobals.classNames.dragMock },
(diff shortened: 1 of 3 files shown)


sebastianbochan (changes requested):

ts/Dashboards/Actions/DragDrop.ts
```diff
@@ -71,6 +71,8 @@ class DragDrop {
         this.editMode = editMode;
         this.options = merge(DragDrop.defaultOptions, options);
 
+        this.enabled = this.options.enabled;
```
pawellysy, line 74: this property is not used at the moment, so there is no need in adding it.

ts/Dashboards/Actions/DragDrop.ts
```diff
@@ -104,6 +106,11 @@ class DragDrop {
      */
     public editMode: EditMode;
 
+    /**
+     * Whether the drag drop is enabled.
+     */
+    public enabled: boolean;
+
```
pawellysy, line 113: Same as above

pawellysy (changes requested):
Test is missing 😅

bre1470 (approved):

sebastianbochan (approved):

karolkolodziej:
@pawellysy  The drag-drop is already tested in Cypress - I added a small change to the demo to fully test that. But since there is no way to update the EditMode (we can talk about that in a dedicated issue in Assanay) we have to leave it as it is.

Issue #9728, opened 2018-12-19
Bubble Chart StyledMode incorrectly handles Negative z values


DannyLingard:
#### Expected behaviour
As with styledMode = false

#### Actual behaviour
.highchart-negative class added only to points based on y value instead of z. Have tested setting relevant displayNegative, negativeColor and zThreshold at both the series{} and bubble{} within plotOptions.

#### Live demo with steps to reproduce
http://jsfiddle.net/9b328kv0/1/

#### Product version
Highcharts v7.0.0

#### Affected browser(s)
Chrome

TorsteinHonsi:
Thanks for reporting! 

Demo with the fix applied: http://jsfiddle.net/highcharts/5nthykuo/

PR #19030, opened 2023-05-22
bugfix/19028-columns-point-negative

### LINKED ISSUE #19028: Border radius of columns applies to the top for negative values when `series.column.zoneAxis = 'x'`
#### Actual behaviour
Changing the `series.column.zoneAxis` property to `x` makes the `series.column.borderRadius` not apply correctly for the columns of negative value.

**Demo:** https://jsfiddle.net/BlackLabel/edqbfunh/
1. Open the demo
2. See how the border radius is applied
3. Comment the `zoneAxis: 'x'` line

#### Product version
Highcharts 11

#### Affected browser(s)
All?

hubertkozik:
Fixed #19028, border radius was wrongly applied for negative column points when the `zoneAxis` option was `x`.

### DIFF
--- a/ts/Series/Bubble/BubbleSeries.ts
+++ b/ts/Series/Bubble/BubbleSeries.ts
@@ -727,8 +727,11 @@ class BubbleSeries extends ScatterSeries {
     }
 
     public translateBubble(): void {
-        const { data, radii } = this;
-        const { minPxSize } = this.getPxExtremes();
+        const { data, radii } = this,
+            { minPxSize } = this.getPxExtremes(),
+            options = this.options,
+            threshold = options.threshold,
+            zoneAxis = this.zoneAxis || 'y';
 
         // Set the shape type and arguments to be picked up in drawPoints
         let i = data.length;
@@ -736,6 +739,13 @@ class BubbleSeries extends ScatterSeries {
         while (i--) {
             const point = data[i];
             const radius = radii ? radii[i] : 0; // #1737
+
+            // Negative points means negative z values (#9728)
+            point.negative = (point as any)[zoneAxis] < (
+                (options as any)[zoneAxis + 'Threshold'] ||
+                threshold ||
+                0
+            );
 
             if (isNumber(radius) && radius >= minPxSize / 2) {
                 // Shape arguments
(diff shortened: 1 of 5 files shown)


ts/Series/Column/ColumnSeries.ts
```diff
@@ -535,6 +535,9 @@ class ColumnSeries extends Series {
                 barX = plotX + seriesXOffset,
                 barW = seriesBarW;
 
+            point.negative = defined(point.y) &&
+                point.y < (series.options.threshold || 0);
```
TorsteinHonsi, line 539: I think instead we should try to understand why `negative` was wrong in the first place. Consider this:

https://github.com/highcharts/highcharts/blob/9ff918be0104d77d0b1981e11e0daaa2c4c89ecd/ts/Core/Series/Series.ts#L2320-L2326

Judging by the comment, we added the `zoneAxis` stuff because we wanted bubbles to be negative when z was negative. As a side effect, it breaks when `zoneAxis` is `x`. It looks like we didn't think of that. There is no `options.xThreshold` property. So I think we should just rewrite the fix for #9728. Maybe check spesificly for `zoneAxis: 'z'`, or better still, move that part out to the Bubble series. It has a `translateBubble` function where we can set `negative`.

ts/Series/Bubble/BubbleSeries.ts
```diff
@@ -737,6 +740,13 @@ class BubbleSeries extends ScatterSeries {
             const point = data[i];
             const radius = radii ? radii[i] : 0; // #1737
 
+            // Negative points means negative z values (#9728)
+            point.negative = (point as any)[zoneAxis] < (
+                (options as any)[zoneAxis + 'Threshold'] ||
+                threshold ||
+                0
+            );
```
TorsteinHonsi, line 748: Can we take this opportunity to do this block typesafe? We learned that the zone axis check is only for the case of `z`, not for `x`. Besides, we already have `point.negative` computed for zoneAxis = 'y'. So I believe we can do something like this:

```ts
if (zoneAxis === 'z') {
    point.negative = (point.z || 0) < (options.zThreshold || 0);
}
```

TorsteinHonsi (changes requested):

ts/Series/Bubble/BubbleSeries.ts
hubertkozik: Ahh, ok! I didn't get it the first time. Will change it - thank you!

TorsteinHonsi (approved):

PR #19031, opened 2023-05-22
bugfix/18961-map-drilldown-multiple-series

### LINKED ISSUE #18961: HC Maps + Drilldown module - Series are not removed after drilling into one of them
[#18925](https://github.com/highcharts/highcharts/issues/18925), map drill down was broken for GeoJSON maps.
Was first iteration of this issue.
I cannot re-open that ticket.

#### Expected behaviour
Clicking on state should drill down to sub-state area type view of data.

#### Actual behaviour
Drilldown occurs then map zooms back out to national view with sub-state delineation visible for state clicked on.
Repeated steps of this drilldown/drillup result in map data breaking.

Let me know if this should be moved to a new issue. I think this is the same issue as the behavior has not changed at all.

Tested under v11.0.1 and issue persists. Fiddle to reproduce: https://jsfiddle.net/wergeld/3djeg6ov/
Click on Florida as it has data (other states will just have null value counties).
Map does drill down to the state view of the counties then zooms back out to national view.
Click "Sample National" breadcrumb to drill back up to national view
Click on any other state - same issue occurs, but this time the Florida county-level map is visible again.
Click on "Sample National" breadcrumb to drill back up.
Click on another state and "Sample National" breadcrumb to drill back up.
Now map is empty - no longer a choropleth, just has the state outlines.

#### Product version
HighMaps v11.0.1

#### Affected browser(s)
Chrome, Edge, Safari all latest versions.

hubertkozik:
Fixed #18961, map drilldown didn't work with multiple series.

### DIFF
--- a/samples/unit-tests/maps/drilldown/demo.js
+++ b/samples/unit-tests/maps/drilldown/demo.js
@@ -28,6 +28,13 @@ QUnit.test('Map drilldown animation', assert => {
                         value: 1,
                         drilldown: 'germany'
                     }]
+                }, {
+                    mapData: europeTopo,
+                    name: 'second',
+                    data: [{
+                        'hc-key': 'pl',
+                        value: 2
+                    }]
                 }],
                 drilldown: {
                     animation: {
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #19032, opened 2023-05-22
bugfix/19017-data-module-spaces-number-parsing

### LINKED ISSUE #19017: A negative number with space is treated as text in data module.
#### Expected behaviour
Text describing a number containing a space as a thousand separator should be treated by the data module as a number regardless of whether it is positive or negative.

#### Actual behaviour
Text describing a number containing a space as a thousand separator is not treated by the data module as a number when it is negative.

https://jsfiddle.net/BlackLabel/9rh4uz2v/

#### Product version
Highcharts v11.0.1

#### Workaround
https://jsfiddle.net/BlackLabel/v5ndkfgh/

Vazonik:
Fixed #19017, parsing negative numbers with spaces in the data module.

### DIFF
--- a/samples/unit-tests/data/general/demo.js
+++ b/samples/unit-tests/data/general/demo.js
@@ -32,7 +32,7 @@ QUnit.test(
 );
 
 QUnit.test(
-    'Data module with decimapPoint and negative numbers (#4749)',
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #19052, opened 2023-05-25
bugfix/18985-enableMouseTracking-update

### LINKED ISSUE #18985: Updating `enableMouseTracking` for Column Series does not work
#### Expected behaviour
Updating `enableMouseTracking` for Column Series (from `true` -> `false`) should work.

**Demo:** https://jsfiddle.net/BlackLabel/0jnuqref/

Even though the `enableMouseTracking` property was updated to `false` for the column series it is still enabled.

#### Product version
Highcharts 11

#### Affected browser(s)
All?

hubertkozik:
Fixed #18985, updating `enableMouseTracking` from `true` to `false` didn't work.

### DIFF
--- a/ts/Core/Series/Series.ts
+++ b/ts/Core/Series/Series.ts
@@ -3318,12 +3318,28 @@ class Series {
             series.redrawPoints();
         }
 
-        // draw the mouse tracking area
-        if (
-            series.drawTracker &&
-            series.options.enableMouseTracking !== false
-        ) {
-            series.drawTracker();
+        // Draw the mouse tracking area
+        if (series.drawTracker) {
+            if (series.options.enableMouseTracking !== false) {
+                series.drawTracker();
+            } else {
+                series.trackerGroups?.forEach((key: string): void => {
+                    const tracker = (series as any)[key];
+                    if (tracker) {
+                        tracker.removeClass(
+                            options.trackByArea ?
+                                'highcharts-tracker-area' :
+                                'highcharts-tracker-line'
+                        );
+                        removeEvent(tracker.element, 'mouseover');
+                        removeEvent(tracker.element, 'mouseout');
+
+                        if (hasTouch) {
+                            removeEvent(tracker.element, 'touchstart');
+                        }
+                    }
+                });
+            }
         }
 
         // Run the animation
--- a/samples/unit-tests/series-column/double-mouseover-column/demo.js
+++ b/samples/unit-tests/series-column/double-mouseover-column/demo.js
@@ -1,24 +1,22 @@
 QUnit.test(
-    'Hover on column should call mouseOver only once.(#4448)',
+    'Hover on column should call mouseOver only once (#4448).',
     function (assert) {
-        var iter = 0,
-            chart = $('#container')
-                .highcharts({
-                    series: [
-                        {
-                            type: 'column',
-                            data: [107, 31, 635, 203, 2],
-                            point: {
-                                events: {
-                                    mouseOver: function () {
-                                        iter++;
-                                    }
-                                }
+        let iter = 0;
+        const chart = Highcharts.chart('container', {
+            series: [
+                {
+                    type: 'column',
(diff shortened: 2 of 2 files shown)


TorsteinHonsi (changes requested):
Thanks!

I think we can simplify this a lot by reading the `enableMouseTracking` option on runtime instead of on the series creation and then deal with it on update.

With the following patch, the check is moved to the event handlers:

```ts
diff --git a/ts/Core/Series/Series.ts b/ts/Core/Series/Series.ts
index bacf661050..17a1cc94aa 100644
--- a/ts/Core/Series/Series.ts
+++ b/ts/Core/Series/Series.ts
@@ -3318,13 +3318,8 @@ class Series {
             series.redrawPoints();
         }
 
-        // draw the mouse tracking area
-        if (
-            series.drawTracker &&
-            series.options.enableMouseTracking !== false
-        ) {
-            series.drawTracker();
-        }
+        // Draw the mouse tracking area
+        series.drawTracker();
 
         // Run the animation
         if (series.animate && animDuration) {
@@ -3666,7 +3661,10 @@ class Series {
             snap = (chart.options.tooltip as any).snap,
             tracker = series.tracker,
             onMouseOver = function (e: PointerEvent): void {
-                if (chart.hoverSeries !== series) {
+                if (
+                    series.options.enableMouseTracking &&
+                    chart.hoverSeries !== series
+                ) {
                     series.onMouseOver();
                 }
             },
diff --git a/ts/Series/Column/ColumnSeries.ts b/ts/Series/Column/ColumnSeries.ts
index 71ee34fc64..a5b612d78b 100644
--- a/ts/Series/Column/ColumnSeries.ts
+++ b/ts/Series/Column/ColumnSeries.ts
@@ -841,7 +841,10 @@ class ColumnSeries extends Series {
                 const point = pointer.getPointFromEvent(e);
 
                 // undefined on graph in scatterchart
-                if (typeof point !== 'undefined') {
+                if (
+                    typeof point !== 'undefined' &&
+                    series.options.enableMouseTracking
+                ) {
                     pointer.isDirectTouch = true;
                     point.onMouseOver(e);
                 }
```

A possible downside now is that the tracker elements and the event handlers are always created, also when `enableMouseTracking` is false. But I can't see any major problem with that. Besides, the majority of implementations are likely to have this option to true, so it will only change for the edge cases that turn it off.

hubertkozik:
Our current solution with `enableMouseTracking` is based on the condition, that it is `not false`, so even if we have this property as `undefined` the code will be fired. That's why I used this condition also in the places, which it is needed.

TorsteinHonsi (changes requested):
> Our current solution with enableMouseTracking is based on the condition, that it is not false, so even if we have this property as undefined the code will be fired. That's why I used this condition also in the places, which it is needed.

Okay. Now since we're doing the `!-- false` check three times, I think it is better if we [set it to true in SeriesDefaults.ts](https://github.com/highcharts/highcharts/blob/26e1d2060194dbf865a2ff04ff9ac39063f69189/ts/Core/Series/SeriesDefaults.ts#L418-L431), so that we can just do the truthy check which is more intuitive.

Also, I was about to say that we should draw the tracker anyway, also when it is disabled. But there may be cases with thousands of points in a line chart where the performance hit is noticeable. So let's keep the check.

A last small optimization - in the Series.ts file we have a local `options` variable in both affected functions, so we should use that.

TorsteinHonsi (approved):

ts/Core/Series/Series.ts
```diff
@@ -3318,10 +3318,10 @@ class Series {
             series.redrawPoints();
         }
 
-        // draw the mouse tracking area
+        // Draw the mouse tracking area
         if (
             series.drawTracker &&
-            series.options.enableMouseTracking !== false
+            options.enableMouseTracking
```
deser, line 3324: Maaaan, this is breaking change...

PR #19055, opened 2023-05-25
bugfix/16068-panning-on-ordinal-axis

### LINKED ISSUE #16068: Panning on ordinal Axis is broken
#### Expected behaviour
When panning, everything should be smooth. 

#### Actual behaviour
Panning causes some strange behaviour, including fireing the dataGrouping. 

http://jsfiddle.net/BlackLabel/d1ps4n3g/
Try to pan to the right (towards higher values)
#### Product version
v9.1.2
#### Affected browser(s)
all

kamil-musialowski:
Fixed #16068, panning on ordinal axis was inconsistent.

### DIFF
--- a/samples/unit-tests/axis/ordinal/demo.js
+++ b/samples/unit-tests/axis/ordinal/demo.js
@@ -317,6 +317,18 @@ QUnit.test('Panning ordinal axis on mobile devices- lin2val calculation, #13238'
         `After adding the point, the extendedOrdinalPositions array
         should be recalculated, #16055.`
     );
+    // #16068
+    chart.xAxis[0].setExtremes(1585665128355, 1586026260000);
+    const controller = new TestController(chart),
+        visiblePoints = chart.series[0].points.filter(p => p.isInside);
+
+    controller.pan([20, 100], [chart.xAxis[0].len, 100]);
+
+    assert.strictEqual(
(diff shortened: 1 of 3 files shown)


pawelfus:
I need a second opinion on this, could you help @TorsteinHonsi ? Personally, I prefer "smooth" zooming in the ordinal axis. To compare see gifs below. 
The idea behind this change is to move always by N-ordinal units, where N is an integer (-1, 2, 3). In the current version, we use rational numbers (-1.5, 2.1, 3.5). I don't see any problems with this solution, how about you? Thanks to that, we can remove the workaround with 30% gap-detection (we still need to make sure we don't accidentally get out of the extremes though)

- v11.0, demo: http://jsfiddle.net/BlackLabel/d1ps4n3g/9/ - you can observe a "lag"/"jump" when reaching 05:00 on the left side:
pan-11 0
Sometimes you can change the zoomed pane:
pan-11


- PR, demo: http://jsfiddle.net/BlackLabel/d1ps4n3g/8/ 
pan-PR

pawelfus:
Okay, thank you!

@kamil-musialowski could you add a test to this PR?

pawelfus (approved):
Thanks!

TorsteinHonsi (approved):

PR #19056, opened 2023-05-26
docs/19034-point-position

### LINKED ISSUE #19034: `shapeArgs`, `plotX`, `plotY` not calculated correctly for inverted charts
#### Expected behaviour
`point.shapeArgs`, `point.plotX`, `point.plotY` are not calculated properly when the `chart.inverted = true`.

**Demo for line series:** https://jsfiddle.net/BlackLabel/2ytor9x1/
**Demo for organization chart:** https://jsfiddle.net/BlackLabel/e3p4xoj5/
The circles should be rendered "on points", but are rendered in strange places. Comment out the `chart.inverted` to see that circles are rendered correctly.

#### Product version
Highcharts 11

#### Affected browser(s)
All?

jakubSzuminski:
Fixed #19034, `Highcharts.Point#pos` will now be a public function available in the API.

---
TODO
- [x] Expose `Point.pos` in the docs (remove `@private`)
- [x] Provide a link from `plotX` and `plotY` to `Point.pos`
- [x] Create a `@sample` for `Point.pos`
- [ ] Add correct params & return to `Highcharts.Point#pos`
- [ ] Improve the information about `plotX` and `plotY` (what happens when `chart.inverted`)

### DIFF
--- a/ts/Core/Series/Point.ts
+++ b/ts/Core/Series/Point.ts
@@ -820,8 +820,13 @@ class Point {
 
     /**
      * Get the pixel position of the point relative to the plot area.
-     * @private
-     * @function Highcharts.Point#pos
+     *
+     * @sample highcharts/point/position
+     *         Get point's position in pixels.
+     *
+     * @param {boolean} chartCoordinates
+     * Whether to return the position relative to the plot area of the chart.
+     *
      */
     public pos(
         chartCoordinates?: boolean,
--- a/samples/highcharts/point/position/demo.css
+++ b/samples/highcharts/point/position/demo.css
@@ -1,0 +1,5 @@
+#container {
+    min-width: 300px;
+    max-width: 600px;
+    margin: 0 auto;
+}
--- a/samples/highcharts/point/position/demo.details
+++ b/samples/highcharts/point/position/demo.details
@@ -1,0 +1,6 @@
+---
+ name: Point Position Example
+ authors:
+   - Jakub Szumiński
+ js_wrap: b
+...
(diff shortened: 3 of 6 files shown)


jakubSzuminski:
**Demo** added as a sample of the `point.pos` function: https://jsfiddle.net/BlackLabel/cn0d7sLb/

jakubSzuminski:
@TorsteinHonsi I'm not really sure about the descriptions I've written of the:
- function arguments (`chartCoordinates`, `plotY`)
- the additional information I've added about the `plotX` and `plotY` of the point 

Could you please take a look and help me change those texts so that they are more clear for the user? I tried, but was unable to formulate my thoughts well enough.

ts/Core/Series/Point.ts
```diff
@@ -820,8 +820,19 @@ class Point {
...
     /**
      * Get the pixel position of the point relative to the plot area.
-     * @private
      * @function Highcharts.Point#pos
+     *
+     * @sample highcharts/point/position
+     *         Get point's position in pixels.
+     *
+     * @param {boolean} chartCoordinates
+     * Whether to return the position relative to the plot area of the chart.
```
TorsteinHonsi, line 829: Actually the other way around.

"If true, the returned position is relative to the full chart area. If false, it is relative to the plot area determined by the axes."

TorsteinHonsi (changes requested):
Sorry for the delay, here's my review.

TorsteinHonsi (approved):

PR #19057, opened 2023-05-26
db/datagrid-height

karolkolodziej:
Fixed the double scroll issue for `DataGrid` in the `climate` demo.


samples/dashboards/demos/climate/demo.css
```diff
@@ -267,6 +267,10 @@
     overflow: auto;
 }
 
+#selection-grid .hc-dg-outer-container {
+    height: 330px !important;
```
pawellysy, line 271: Can we avoid using `!important`?  Maybe the design of the DataGrid is wrong, to set inner style on the element?

pawellysy: The size is set by dataGridComponent. maybe we should add some margin there?

pawellysy (changes requested):

samples/dashboards/demos/climate/demo.css
sebastianbochan: Maybe we should set this value from the JS?

bre1470 (approved):
Looks good. 👍

sebastianbochan (changes requested):

sebastianbochan (approved):

Issue #19038, opened 2023-05-23
Treegraph chart does not fill entire width when a few nodes are collapsed


Ritesh717:
When I create a tree graph with collapsed nodes, the chart does not occupy the full width of the container. This is because the chart is still taking up space for the collapsed nodes.

#### Expected behaviour
When a few nodes in a tree graph are collapsed, the chart should fill the full width of the container. If a user expands a collapsed node, the chart should scroll to the right to make room for the expanded node.

#### Actual behaviour
When a few nodes in a tree graph are collapsed, the chart does not fill the full width of the container.

#### Live demo with steps to reproduce

https://jsfiddle.net/riteshGangam/fe6sL5bq/
https://jsfiddle.net/riteshGangam/pez5du3L/35/

Without Collapsable nodes
<img width="960" alt="image" src="https://github.com/highcharts/highcharts/assets/40355220/2609dc7e-1ebe-4115-a79f-62d987ab8e12">

With collapsable nodes
<img width="959" alt="image" src="https://github.com/highcharts/highcharts/assets/40355220/778eae11-fc0d-4672-bbf7-63ebb99c8125">

#### Product version

Highcharts JS v11.0.1 (2023-05-08)
#### Affected browser(s)
Chrome

pawellysy:
Hi @Ritesh717, thanks for your suggestion. This behaviour is working according to the design, but we might want to add some property, which will allow to do what you want to achieve 👍

pawellysy:
Here is the workaround: https://jsfiddle.net/BlackLabel/h42f9rzg/

Ritesh717:
Hi @pawellysy 

Thanks for the workaround you provided to fix the issue of rendering the chart with nodes collapsed. However, expanding the collapsed nodes makes it difficult to read the data. 
Is there a workaround to add a scroll option?
https://jsfiddle.net/riteshGangam/fshjxynu/1/

PR #19061, opened 2023-05-29
feature/19038-treegraph-fill-x-option

Vazonik:
Added the [fillSpace](https://api.highcharts.com/highcharts/series.treegraph.fillSpace) option for the treegraph to fill the whole plot area in X by all visible points. See #19038.


TorsteinHonsi (changes requested):
Thanks!

* First of all, we need a visual `@sample` to accompany the doclet for the new option.
* In the PR description, add a link to the API option for `fillX`. It does not yet exist, but it will do after release.

Another thing to consider - how does this work with an inverted treegraph?

Vazonik:
@TorsteinHonsi Thanks for the review!
After advice from the team, I decided to rename the option from `fillX` to `fillSpace`.

> Another thing to consider - how does this work with an inverted treegraph?

With both inverted chart and reversed series it seems to work fine. (I found one bug, but it's unrelated to this feature and I reported it here: #19068.)

samples/highcharts/series-treegraph/fillspace-option/demo.css
TorsteinHonsi, line 1: Just `series-treegraph/fillspace` will do for the file path. There are lots of options, we don't need `-option` as part of the path.

samples/highcharts/series-treegraph/fillspace-option/demo.html
```diff
@@ -0,0 +1,10 @@
+<script src="https://code.highcharts.com/highcharts.js"></script>
+<script src="https://code.highcharts.com/modules/treemap.js"></script>
+<script src="https://code.highcharts.com/modules/treegraph.js"></script>
+<script src="https://code.highcharts.com/modules/exporting.js"></script>
+<script src="https://code.highcharts.com/modules/accessibility.js"></script>
+
+<div id="container1"></div>
+<div id="container2"></div>
+<button id="toggleCollapseNodeA">Toggle collapse of nodes A</button>
+<button id="toggleCollapseNodeB">Toggle collapse of nodes B</button>
```
TorsteinHonsi, line 10: We use kebab-case for class names and id's in our demos.

TorsteinHonsi (changes requested):
Thanks, nice demo! Just a little nitpicking now.

TorsteinHonsi (approved):
Thanks! Now only awaiting CI.

PR #19062, opened 2023-05-29
db/error-handler

sebastianbochan:


karolkolodziej (changes requested):
- [ ] Test/demo missing
- [ ] Don't we need some default styling for that error component or at least the `className` so it can be configured if needed?

samples/dashboards/demos/component-error-handler/demo.html
```diff
@@ -0,0 +1,7 @@
+<script src="https://code.highcharts.com/dashboards.js"></script>
+<script src="https://code.highcharts.com/highcharts.js"></script>
+<script src="https://code.highcharts.com/highcharts-more.js"></script>
```
karolkolodziej, line 3: We don't need `highcharts-more` and `solid-gauge`

karolkolodziej (approved):

goransle (approved):

ts/Extensions/DashboardPlugins/HighchartsComponent.ts
```diff
@@ -698,7 +698,11 @@ class HighchartsComponent extends Component {
         if (this.chartConstructor !== 'chart') {
             const factory = charter[this.chartConstructor] || G.chart;
             if (factory) {
-                return factory(this.chartContainer, this.chartOptions);
+                try {
+                    return factory(this.chartContainer, this.chartOptions);
+                } catch {
+                    // returns errors from factory by default
```
bre1470, line 704: Silent catch is bad design. At least log the error.

bre1470 (changes requested):

ts/Extensions/DashboardPlugins/HighchartsComponent.ts
sebastianbochan: Fixed.

bre1470 (approved):

PR #19065, opened 2023-05-31
bugfix/18459-moving-ordinal-annotations

### LINKED ISSUE #18459: Moving Annotations Bug
#### Expected behaviour
After drawing an annotation (such as a trend line) on a chart it should be possible to edit it to change the position by clicking it and moving it with the mouse.

#### Actual behaviour
In the following situation when trying to move an annotation with a mouse the movement of the annotation doesn't track the position of the pointer:
- Set up a stock chart with a reasonably long timeseries of data
- Set the view to less than the full range - eg 1 Year
- Add an indicator - in my testing I've used SMA and EMA indicators
- Draw a simple annotation such as a line
- If you try to move this annotation it won't track the mouse

This bug only seems to appear under those specific conditions. If the technical indicator is removed or the zoom level is set to view the full timeseries then the annotation can be moved as expected.

It is possible to reproduce this with the standard Stock Tools GUI example chart:

#### Product version
 Highstock JS v10.3.3 (2023-01-20)

#### Affected browser(s)
Firefox 109.0.1 on Fedora (64-bit)
Chromium Version 109.0.5414.74 (Official Build) Fedora Project (64-bit)

kamil-musialowski:
Fixed #18459, dragging annotations on chart with ordinal axes was inconsistent.

### DIFF
--- a/samples/unit-tests/axis/ordinal/demo.js
+++ b/samples/unit-tests/axis/ordinal/demo.js
@@ -615,34 +615,42 @@ QUnit.test('Ordinal axis, data grouping and boost module, #14055.', assert => {
 });
 
 QUnit.test('Circular translation, #17128.', assert => {
+    const data = [{
+        type: 'line',
+        data: [
+            [548935806499, 95.82],
+            [1548936121889, 95.84],
+            [1548936895949, 95.75],
+            [1548937941785, 95.48],
+            [1548938881593, 95.6],
+            [1548939834796, 95.37],
+            [1548940821273, 95.16],
+            [1548941760541, 95.15],
+            [1548942617180, 94.9],
+            [1548943265472, 95.04],
+            [1548943953574, 94.93],
+            [1548944604420, 94.94],
+            [1548945157396, 95.19],
+            [1548945448867, 94.92],
+            [1548946059662, 94.98],
+            [1548946666809, 95.17],
+            [1548947190658, 95.38]
+        ],
+        showInNavigator: true
+    }, {
+        type: 'scatter',
+        data: [
+            [1548936121889, 90],
+            [1548938881593, 95]
+        ],
+        showInNavigator: false
+    }];
+
     const chart = Highcharts.stockChart('container', {
-            series: [{
-                data: [
-                    [548935806499, 95.82],
-                    [1548936121889, 95.84],
-                    [1548936895949, 95.75],
-                    [1548937941785, 95.48],
-                    [1548938881593, 95.6],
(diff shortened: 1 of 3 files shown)


ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -1407,26 +1407,31 @@ namespace OrdinalAxis {
...
+            axis.series.forEach((series): void => {
+                const firstPoint = series.points?.[0];
+
+                if (
+                    defined(firstPoint?.plotX) &&
+                    hasPointsInside(series) &&
+                    (
+                        !defined(firstPointX) ||
+                        firstPoint.plotX < firstPointX
+                    )
```
pawelfus, line 1427: Could you please change the order of conditions? In general, the most consuming ones should be the last ones (they are evaluated only if others are truthy). I mean `hasPointsInside` imho should be the last one

pawelfus (changes requested):
Just a minor polishing, nice!

ts/Core/Axis/OrdinalAxis.ts
kamil-musialowski: Absolutely right, thank you for spotting that!

pawelfus (approved):
Thanks!

samples/unit-tests/axis/ordinal/demo.js
```diff
@@ -355,9 +355,11 @@ QUnit.test('lin2val- unit test for values outside the plotArea.', function (asse
         },
         series: [{
             points: [{
+                isInside: true, // #18459
                 x: 3,
                 plotX: -20
             }, {
+                isInside: true, // #18459
```
TorsteinHonsi, line 362: There is no such option... Only the internal, computed property `Point.isInside`. So is this test realistic, or are we cheating by forcing a property that should be computed?

TorsteinHonsi (commented):
Just one question

samples/unit-tests/axis/ordinal/demo.js
kamil-musialowski: Sure thing, there is no such property - but since we don't even create a chart there, and use fake points, this property won't get computed (right?). We test `lin2val` outside the chart environment. Therefore I decided to manually put it in there, to ensure that a new logic will treat those points as they are inside the chart.

In fact, I think line 358 should be removed here.

TorsteinHonsi: Ok, I didn't realize it was mock points.

PR #19067, opened 2023-05-31
bugfix/19042-annotation-label-edit

### LINKED ISSUE #19042: Annotations label edit makes the text dissapear
#### Actual behaviour
After adding the label annotation and editing it through annotations GUI (even without changing any parameters) makes the text inside the label disappear.

Click on the `TXT` icon to add the simple label to the chart
Click on the label
Click the pencil icon to edit it
Click save button (you don't have to change any data)

#### Product version
Highstock 11.0.1 (latest)

kamil-musialowski:
Fixed #19042, text of annotation label was missing after updating it.


TorsteinHonsi (approved):
Thanks!

PR #19071, opened 2023-06-01
bugfix/19070 heatmap interpolation update

### LINKED ISSUE #19070: Heatmap interpolation doesn't redraw on changed data
#### Expected behaviour
Redraw (on the same `image`) when data changes

#### Actual behaviour
It only renders once, after that only responds to size changes. The logic where the image data is computed and rendered onto the `series.image` is inside a block checking for `!image`. Instead, if the image exists already, the updated image data should be rendered to the pre-existing image. To prevent redundant processing, the `series.isDirtyData` flag can be checked.

https://jsfiddle.net/highcharts/30jx9yvc/ . Note the difference between interpolation true and false.

#### Product version
Pre-release of v11.1

#### Affected browser(s)
All

TorsteinHonsi:
- Fixed #19070, pre-release issue with heatmap update

### DIFF
--- a/docs/maps/getting-started.md
+++ b/docs/maps/getting-started.md
@@ -1,7 +1,7 @@
 Getting started with Highcharts Maps
 ===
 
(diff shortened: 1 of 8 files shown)


MarkusBarstad (commented):
Looks good!

MarkusBarstad (approved):

PR #19072, opened 2023-06-01
bugfix/19053-item-series-update

### LINKED ISSUE #19053: Item chart update not working properly
#### Expected behaviour

Item chart update should work (ref. [Item chart docs](https://www.highcharts.com/docs/chart-and-series-types/item-chart))

#### Actual behaviour

Update not working properly. For example, in the doc demo above, try switching between Britain and Norway. Series markers are "stuck" when updating. Also, see demo links below. 

Demo - working (v10.2.1): https://jsfiddle.net/BlackLabel/z6yo2x1q/   :white_check_mark:
Demo - not working (v10.3.0): https://jsfiddle.net/BlackLabel/6eg859p0/  :x:

#### Product version
Highcharts Core, v.10.3.0
Probable culprit: looking at the changelog, there is a bugfix #17315 related to item series markers.

#### Affected browser(s)
All

jakubSzuminski:
Fixed #19053, a regression since 10.3 causing item series update not to work correctly. 

**Demo of the problem:** https://jsfiddle.net/BlackLabel/1x2L70h5/
_Some of the graphics from before-update were still there even though they shouldn't be._ 

**Demo after fix:** https://jsfiddle.net/BlackLabel/u6m4rks8/

### DIFF
--- a/ts/Series/Item/ItemSeries.ts
+++ b/ts/Series/Item/ItemSeries.ts
@@ -345,18 +345,21 @@ class ItemSeries extends PieSeries {
                     i++;
                 }
             }
-            graphics.forEach((graphic, i): void => {
+
+            for (let j = 0; j < graphics.length; j++) {
+                const graphic = graphics[j];
                 if (!graphic) {
                     return;
                 }
 
                 if (!graphic.isActive) {
                     graphic.destroy();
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (approved):
Perfect, thanks!

Issue #2438, opened 2013-11-07
cannot select first point with rangeSelector input after moving it past (Highstock)


simpixelated:
Example: http://jsfiddle.net/48zYU/3/
- using the date input, move the date at least one day forward (e.g. 2013-07-02)
- using the date input, try to move the date back to the original date (2013-07-01)
- notice it resets the date to the one you've entered and will not update the navigator

Possibly related to issue #433

PR #19073, opened 2023-06-01
bugfix/18251-range-selector-multiple-series

### LINKED ISSUE #18251: Gantt Chart Range selector not consistent
Hi Team,
I'm trying to check Gantt chart meet our requirements. Range selector didn't allow me to select valid date from date picker. it is going to first row date range.

2023-01-05_15h11_01

When i select Jan 22nd. it is going to Jan 19 (first row date range). Please check on live demo
Live Demo : [Project Management - JSFiddle - Code Playground](https://jsfiddle.net/shihabkaniyarath/y0r496ct/2/)

#2 When I try to export with collapsed, collapsed items are not showing the result. And navigator showing the result

jakubjanuchta:
Fixed #18251, range selector should consider all series extremes.

### DIFF
--- a/samples/unit-tests/rangeselector/input-range/demo.js
+++ b/samples/unit-tests/rangeselector/input-range/demo.js
@@ -375,10 +375,33 @@ QUnit.test('Set extremes on inputs blur (#4710)', function (assert) {
                         [Date.UTC(2007, 8, 20), 0.7107],
                         [Date.UTC(2007, 8, 21), 0.7097],
                         [Date.UTC(2007, 8, 24), 0.7098],
-                        [Date.UTC(2007, 8, 25), 0.7069],
-                        [Date.UTC(2007, 8, 26), 0.7078],
-                        [Date.UTC(2007, 8, 27), 0.7066],
-                        [Date.UTC(2007, 8, 28), 0.7006]
+                        [Date.UTC(2007, 8, 25), 0.7069]
+                    ]
+                },
+                {
+                    data: [
+                        [Date.UTC(2007, 8, 3), 0.8342],
+                        [Date.UTC(2007, 8, 4), 0.8349],
+                        [Date.UTC(2007, 8, 5), 0.8326],
+                        [Date.UTC(2007, 8, 6), 0.8306],
+                        [Date.UTC(2007, 8, 7), 0.8263],
+                        [Date.UTC(2007, 8, 10), 0.8247],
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

TorsteinHonsi (commented):
I can't break this, but still I am wondering about the fix. First, the comment "Validate the extremes. If it goes beyound the data min or max, use the actual data extreme (#2438)." doesn't seem to match the new logic. Yet, it still apparently works for the #2438 case.

Next, I am wondering why `dataAxis.dataMin` and `dataMax` do not work with multiple series. They should, because the `Axis.getSeriesExtremes` function should consider all the axis' series. So maybe that's where the real bug is?

pawelfus:
I can explain the problem a  little deeper:
In the demo, we have the main x-axis with two series and a navigator x-axis with just one series. `dataMax` in the navigator reflects data extremes taken from the shorter series (which is correct). `max` is taken from all series in the chart (from all x-axes) using `modifyNavigatorAxisExtremes` method (which is well... correct too). 

There is another solution, instead of getting `scroller.max`: use `getUnioneExtremes()` (like all other methods in RS do). Aaaan when writing a comment, I realized how to break it: `max` < `dataMax`:  [current fiddle](https://jsfiddle.net/BlackLabel/zL843r7x/) vs [PR fiddle](https://jsfiddle.net/BlackLabel/zL843r7x/4/). Select in max-input 30-09-2006 date, nothing happens in PR fiddle.

jakubjanuchta:
Thanks for following up @TorsteinHonsi and @pawelfus.

Regarding to what @pawelfus pointed out about breaking the fix when `max` < `dataMax`, should I first apply the logic to check which of them is bigger in Range Selector, or do you guys think there is a better solution for that?

TorsteinHonsi:
> should I first apply the logic to check which of them is bigger in Range Selector, or do you think that there is a better solution for that?

@pawelfus What do you advice?

jakubjanuchta:
> I would start with using `getUnionExtremes` - it already has this logic implemented (max for `dataMax` and `max`)

Okey, thanks for the tip @pawelfus , I'll do that!

TorsteinHonsi (approved):
Thanks! Pending final approval by Pawel.

pawelfus (approved):
Looks good, thanks!

PR #19074, opened 2023-06-01
db/hc-component-columnKeyMap

sebastianbochan:


karolkolodziej (approved):

bre1470 (approved):
We need also an update on documentation to point out that you can use `null` to keep columns selectively out of the chart.

PR #19081, opened 2023-06-02
db/fix-import-dashboard

sebastianbochan:


samples/dashboards/demos/layout/demo.js
```diff
@@ -17,7 +17,7 @@ let board = Dashboards.board('container', {
                 text: 'Import saved dashboard',
                 events: {
                     click: function () {
-                        board = Dashboards.importLocal();
+                        board = Dashboards.Board.importLocal();
```
karolkolodziej, line 20: There are multiple places where the `Import saved dashboard` is being used.

Also please add test if possible.

karolkolodziej (changes requested):

samples/dashboards/demos/layout/demo.js
sebastianbochan: I was not able to another demos. We have some cypress tests, based on modules and there is a reference to `Board.importLocal()` which is correct.

I'm not pretty sure how this test could look like, it's a simple reference to a function. In the case of tests, we should have many tests that cover using all references from classes.

karolkolodziej: You are right with the first one!

For now, we can skip tests ;)

karolkolodziej (approved):

goransle (approved):

PR #19082, opened 2023-06-02
Tooltip not showing up for boosted interpolated heatmap

MarkusBarstad:
Fixed, directTouch was only true for non-boost


TorsteinHonsi (changes requested):
Thanks! Remember to remove the now unused `notBoosting` constant too.

TorsteinHonsi:
Done. Pending CI.

TorsteinHonsi (approved):

PR #19083, opened 2023-06-02
bugfix/15569-zoom-by-drag-improvement 

### LINKED ISSUE #15569: Chart zoom is not smooth
#### Expected behaviour
Chart zoom should be smooth 

#### Actual behaviour
We have encountered this issue mostly with stock charts with large dataset. When we zoom into the chart with a relatively faster mouse movement, the selection often gets stuck.

Right now, mouse-zooming charts is done by dragging the mouse inside the axes area. Unfortunately, dragging beyond the edge of that area stops the JS code from receiving events (or something like that) and it's extremely hard to get the selection to go all the way to the left or right edge. When we move the mouse slowly, we are able to select that. Refer the following gif:

gif

https://jsfiddle.net/o3pnxkjw/

#### Product version
Highstock 9.0.1

#### Affected browser(s)
Browser

kamil-musialowski:
Fixed #15569, at rapid mouse movements outside the chart, zoom didn't work properly.

### DIFF
--- a/samples/unit-tests/chart/zoomtype/demo.css
+++ b/samples/unit-tests/chart/zoomtype/demo.css
@@ -3,7 +3,3 @@
     height: 400px;
     margin: 0 auto;
 }
-
-.flex-container {
-    display: flex;
-}
--- a/samples/unit-tests/chart/zoomtype/demo.css
+++ b/samples/unit-tests/chart/zoomtype/demo.css
@@ -3,3 +3,7 @@
     height: 400px;
     margin: 0 auto;
 }
+
+.flex-container {
+    display: flex;
+}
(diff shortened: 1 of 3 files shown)


samples/unit-tests/chart/zoomtype/demo.css
```diff
@@ -3,3 +3,7 @@
     height: 400px;
     margin: 0 auto;
 }
+
+.flex-container {
+    display: flex;
+}
```
pawelfus, line 9: This will not work in Karma - CSS file (all together with HTML file) is not loaded

pawelfus (changes requested):
Minor polishing in tests, otherwise good!

pawelfus (approved):
Looks good, works fine - should be ok, but let's get feedback from devs participating in #15569

TorsteinHonsi (approved):

PR #19084, opened 2023-06-03
Fixed #19075: error on pie drilldown

### LINKED ISSUE #19075: Pie chart drilldown errors on V11
We have recently upgraded Highcharts from version 10 to 11 and observed an error being logged in the console when clicking to view drilldown data in a pie chart.

#### Expected behaviour
When clicking to view drilldown data in a pie chart, no errors are logged in the console.

#### Actual behaviour
When clicking to view drilldown data in a pie chart, the following error is logged in the console:
```
Uncaught TypeError: Cannot read properties of undefined (reading 'chart')
at c.select (highcharts-<fingerprint>.js:7904:19)
    at c.d (highcharts-<fingerprint>.js:7689:29)
    at e (highcharts-<fingerprint>.js:269:37)
    at c.firePointEvent (highcharts-<fingerprint>.js:7691:11)
    at F.onContainerClick (highcharts-<fingerprint>.js:8478:35)
```
Note: the error does not prevent the drilldown data from being viewed.

https://jsfiddle.net/chrisgilbert/n3Lvy6pd/

#### Product version
Highcharts version 11

#### Affected browser(s)
Google Chrome Version 113.0.5672.126 (Official Build) (arm64)

Kogoshvili:
Fix for issue #19075
It seems that defaultFunction is called after point gets destroyed because of drilldown action, so I added check for the point.

### DIFF
--- a/ts/Core/Series/Point.ts
+++ b/ts/Core/Series/Point.ts
@@ -524,7 +524,7 @@ class Point {
             defaultFunction = function (event: MouseEvent): void {
                 // Control key is for Windows, meta (= Cmd key) for Mac, Shift
                 // for Opera.
-                if (point.select) { // #2911
+                if (!point.destroyed && point.select) { // #2911, #19075
                     point.select(
                         null as any,
                         event.ctrlKey || event.metaKey || event.shiftKey


TorsteinHonsi:
Thanks! Taken over by #19101 so that we can run our CI on it.

PR #19100, opened 2023-06-06
db/docs-review-improvements

sebastianbochan:


ts/Data/README.md
```diff
@@ -162,8 +162,8 @@ const chart = new Highcharts.chart('container', {
 
 DataGrid shows and optionally modifies cell content in a table. DataGrid can
 also change the order of cells, but DataTable provides only limited information
-about the original order of a source. Therefor a DataConnector might be needed
-in addition to retrieve the original order.
+about the original order of a source. Therefore, a DataConnector might be needed
```
bre1470, line 165: There should be no comma.

ts/Data/README.md
```diff
@@ -211,8 +211,8 @@ either an URL or a local source.
 
 ### DataConnector Registry
 
-DataConnector types can be directly loaded via import. In case of bundles
-connectors can also accessed via registry, as the registry gets updated with
+DataConnector types can be directly loaded via import. In case of bundles,
```
bre1470, line 214: There should be no comma. Instead continue with ...the connectors...

docs/dashboards/your-first-dashboard.md
```diff
@@ -1,24 +1,24 @@
...
 1. Add a div to your webpage with a specific id.
 
     ```html
         <div id="container">
     ```
 
-2. In this basic example we are going to create a dashboard with two cells. One will contain a static HTML and the other will contain a chart created with the Highcharts.
-Thus we have to import the main dashboard package, the Highcharts library and a plugin that allows implementing the Highcharts chart inside the dashboard.
+2. In this basic example we are going to create a dashboard with two cells. One will contain a static HTML and the other will contain a chart created with Highcharts.
+Thus, we have to import the main dashboard package, the Highcharts library and a plugin that allows implementing the Highcharts chart inside the dashboard.
```
bre1470, line 13: There should be no comma.

docs/dashboards/your-first-dashboard.md
```diff
@@ -85,4 +85,4 @@ Thus we have to import the main dashboard package, the Highcharts library and a
         }]
     ```
 
-6. With that config, you dashboard should look like the one in the [demo](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/samples/dashboard/demos/your-first-dashboard).
\ No newline at end of file
+6. With that config, your dashboard should look like the one in the [demo](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/samples/dashboard/demos/your-first-dashboard).
```
bre1470, line 88: There should be no comma.

bre1470 (changes requested):
It would be good to avoid editing data layer in dashboards, as it should go over the master. Some commas er not necessary.

docs/dashboards/your-first-dashboard.md
sebastianbochan: It's the input from Lars Helge.

ts/Data/README.md
sebastianbochan: It's the input from Lars Helge.

pawelfus (approved):
Looks good to me 👍 
Regarding comma-discussions, summoned Lars Helge to join the discussion

karolkolodziej (approved):

ts/Data/README.md
LB0ge: I think there should be a comma here, it's more clear. Also, perhaps add a _the_ to the introductory phrase ("In the case of bundles").

docs/dashboards/your-first-dashboard.md
LB0ge: Same as for _therefore_ at the beginning of a sentence.

docs/dashboards/your-first-dashboard.md
LB0ge: Comma after the introductory phrase is better in my opinion.

PR #19116, opened 2023-06-07
Fixed DTS for mouse-wheel-zoom bundles

### LINKED ISSUE #19103: Mousewheel zoom module missing d.ts file
#### Expected behaviour

The mouse-wheel-zoom module should have a .d.ts type declaration file (like other modules) so that the module can be imported in TypeScript apps.

#### Actual behaviour

The file is missing.

Try to import the module in a TypeScript app

#### Product version

11.1.0

#### Affected browser(s)

N/A

bre1470:
Fixed #19103


TorsteinHonsi (approved):
Thanks!

PR #19119, opened 2023-06-08
db/context-menu-button

karolkolodziej:
Fixed context menu button click.

### DIFF
--- a/ts/Dashboards/EditMode/EditContextMenu.ts
+++ b/ts/Dashboards/EditMode/EditContextMenu.ts
@@ -14,13 +14,13 @@
  *
  * */
 
+import type { HTMLDOMElement } from '../../Core/Renderer/DOMElementType';
+
 import EditGlobals from './EditGlobals.js';
-import U from '../../Core/Utilities.js';
 import MenuItem from './Menu/MenuItem.js';
 import Menu from './Menu/Menu.js';
 import EditMode from './EditMode.js';
-import { HTMLDOMElement } from '../../Core/Renderer/DOMElementType.js';
-
+import U from '../../Core/Utilities.js';
 const {
     addEvent,
     merge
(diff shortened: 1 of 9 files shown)


ts/Dashboards/EditMode/EditContextMenu.ts
```diff
@@ -47,21 +47,7 @@ class EditContextMenu extends Menu {
...
-            },
-            langKey: 'editMode',
-            events: {
-                click: function (this: MenuItem): void {
-                    (this.menu as EditContextMenu).editMode.onEditModeToggle();
-                }
-            }
-        }
-    });
+    public static items = merge(Menu.items, {});
```
sebastianbochan, line 50: Do we need merge here?

ts/Dashboards/EditMode/EditRenderer.ts
```diff
@@ -486,6 +482,14 @@ function renderText(
             {},
             parentElement
         );
+
+        if (options.click) {
+            textElem.addEventListener('click', (): void => {
+                if (options.click instanceof Function) {
+                    options.click();
```
sebastianbochan, line 489: When the toggle is created, the click event is also added. So, the click event will be triggered twice on the toggle button.

sebastianbochan (changes requested):

bre1470 (commented):
If the button is supposed to work out of the box, then it fails for me in Safari on the `dashboards/demos/layout` demo.

ts/Dashboards/EditMode/EditContextMenu.ts
karolkolodziej: `merge` create a copy, IMO it's safer to work on a fresh copy than accidentally modify an existing object.

ts/Dashboards/EditMode/EditRenderer.ts
karolkolodziej: Tested this and everything works the same as in `dashboards/main`. 
When I tried to simplify it the toggle in the sidebar fail. I guess it needs to stay as it is.

ts/Dashboards/EditMode/EditRenderer.ts
```diff
@@ -486,6 +482,14 @@ function renderText(
...
+                }
+            });
```
pawellysy, line 491:
```suggestion
        if (options.click && (options.click instanceof Function)) {
            textElem.addEventListener('click', options.click);
```

pawellysy (changes requested):
Minor polishing

PR #19120, opened 2023-06-08
Heatmap interp boostbug

MarkusBarstad:
Fixed issue with missing points in boosted heatmap with interpolation set to true

---
  - https://app.asana.com/0/0/1204905562567971
  - https://app.asana.com/0/0/1205169853510983


TorsteinHonsi (changes requested):
Thanks! 

The description (for the changelog) should be more, well, _descriptive_. It doesn't currently say it applies to the heatmap series, nor that it applies only when interpolation is true.

Also, please fix the conflicts.

I will follow up with a more detailed review.

TorsteinHonsi:
There's something funny going on around the 15th-16th of may in the http://localhost:3030/samples/view?path=highcharts/studies/heatmap-interpolation demo. You can see it in the [visual diffs](https://vrevs.highsoft.com/pr/19120/review) too. Looks like a whole column is missing or something.

ts/Series/Heatmap/HeatmapSeries.ts
```diff
@@ -608,42 +568,41 @@ class HeatmapSeries extends ScatterSeries {
...
                                 ) +
                                 xPlacement(
-                                    scaleToImg(
+                                    resizeFromRange(
                                         x,
                                         xMin,
-                                        xMax,
-                                        0,
+                                        xDelta,
                                         widthLastIndex
```
TorsteinHonsi, line 583: If I change this from `widthLastIndex` to `canvasWidth`, the `studies/heatmap-interpolation` is correct. But then the other two demos break 🤷

MarkusBarstad:
Note: problem with reversion/inversion in last commit, should not be a major issue

TorsteinHonsi:
Putting this back in draft until it's considered read for review.

TorsteinHonsi (approved):
Thanks, the math looks good now! As far as I can see the interpolation is perfect in all cases.

PR #19124, opened 2023-06-08
feature/geoheatmap-interpolation

hubertkozik:
Added support for GeoHeatmap interpolation.

----

### DIFF
--- a/ts/Series/GeoHeatmap/GeoHeatmapSeries.ts
+++ b/ts/Series/GeoHeatmap/GeoHeatmapSeries.ts
@@ -21,6 +21,11 @@
 import type GeoHeatmapSeriesOptions from './GeoHeatmapSeriesOptions.js';
 import SeriesRegistry from '../../Core/Series/SeriesRegistry.js';
 import GeoHeatmapPoint from './GeoHeatmapPoint.js';
+import type SVGElement from '../../Core/Renderer/SVG/SVGElement';
+import H from '../../Core/Globals.js';
+const {
+    doc
+} = H;
 
 const {
     seriesTypes: {
@@ -32,7 +37,9 @@ import U from '../../Core/Utilities.js';
 
 const {
     extend,
-    merge
+    merge,
+    pick,
+    defined
 } = U;
 
 /* *
@@ -159,6 +166,10 @@ class GeoHeatmapSeries extends MapSeries {
 
     public points: Array<GeoHeatmapPoint> = void 0 as any;
 
+    public canvas?: HTMLCanvasElement = void 0 as any;
+
+    public context?: CanvasRenderingContext2D = void 0 as any;
+
     /* *
      *
      *  Functions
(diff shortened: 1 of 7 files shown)


MarkusBarstad (approved):
Good work! 

Just adding in my 2 cents in terms of improving performance, take anything I say with a major pinch of salt, I have only recently started reading more up on JS performance :)

Initially I tried to see if caching property-access could increase performance but as I went through the code, trying this hypothesis, I found that this did not really matter that much.

Looking in DevTools revealed that the most taxing function was "getImageData()" from the Canvas API. Other than that "getProjectedImageData()" is the most taxing function.

There could be other ways of scaling the image-data such (maybe there is some way were ctx.scale could help us avoid ctx.drawImage, or perhaps an ArrayBuffer would make it so that we would not have to call getImageData()). 

Other possible optimizations could be
Optimizing getProjectedImageData (caching all property accesses, limiting the for-loop there as hard as we can) and using 32 bit unsigned ints instead of uint8 arrays for colors could probably help to an extent as well, but I am not sure how much, and this could also make the code less readable.

All in all, good job, it was a joy to read and learn from this!

hubertkozik:
I tried to optimize `getProjectedImageData`, I have pushed new changes.

Regarding 32-bit unsigned ints instead of unit8 clamped array for colors, I have done a lot of tests and it seems, that for our purpose (calculating color from colorAxis during the loop through pixels) it is not efficient as using a simple loop.

Demos (console times in the console):
`Uint32Array`: https://jsfiddle.net/BlackLabel/67ypo13L/
`Unit8ClampedArray`: https://jsfiddle.net/BlackLabel/enLd9hos/

MarkusBarstad (approved):
Cool! Zooming seems better now!

ts/Series/GeoHeatmap/GeoHeatmapSeries.ts
```diff
@@ -29,12 +40,146 @@ const {
...
+        (canvasWidth * (canvasHeight - 1 - (lat + 90) / rowsize)) +
+        ((lon + 180) / colsize)
+    );
+}
+
+/**
+ * Project ImageData to actual mapView projection used on a chart.
+ * @private
+ */
+function getProjectedImageData(
```
KacperMadej, line 92: Functions defined like this (scope) are not wrappable or overwritable from outside making any hotfix or any potential extension / wrap longer. getProjectedImageData could be a method of GeoHeatmapSeries instead.

ts/Series/GeoHeatmap/GeoHeatmapSeries.ts
```diff
@@ -175,9 +381,314 @@ class GeoHeatmapSeries extends MapSeries {
...
+
+                    const upscaledWidth = ~~(canvasWidth * blurFactor),
+                        upscaledHeight = ~~(canvasHeight * blurFactor),
+                        projectedWidth = ~~dimensions.width,
+                        projectedHeight = ~~dimensions.height;
+
+                    canvas.width = upscaledWidth;
+                    canvas.height = upscaledHeight;
+
+                    const img =
```
KacperMadej, line 502: Looks like these lines could be cleaned-up - e.g. less const, less empty lines, etc.

ts/Series/InterpolationUtilities.ts
```diff
@@ -0,0 +1,113 @@
...
+/* *
+ *
+ *  Imports
+ *
+ * */
+
+import type HeatmapSeries from './Heatmap/HeatmapSeries';
+import type GeoHeatmapSeries from './GeoHeatmap/GeoHeatmapSeries';
+import type Point from '../Core/Series/Point';
+import type ColorAxis from '../Core/Axis/Color/ColorAxis';
```
KacperMadej, line 22: 'ColorAxis' is declared but its value is never read.

ts/Series/InterpolationUtilities.ts
```diff
@@ -0,0 +1,113 @@
...
+        }
+        return rgba;
+    }
+    return [0, 0, 0, 0];
+}
+
+/**
+     * Method responsible for creating a canvas for interpolation image.
+     * @private
+     */
```
KacperMadej, line 86: Indentation

KacperMadej (changes requested):

KacperMadej (approved):

samples/maps/demo/geoheatmap-earth-statistics/demo.details
```diff
@@ -0,0 +1,15 @@
+---
+name: Earth Statictics August 2022 - GeoHeatmap interpolation
```
TorsteinHonsi, line 2: The name should be shortened - this is what is used for the titles on the [demo page](https://www.highcharts.com/demo/maps#highcharts-maps-demo-general).

samples/maps/demo/geoheatmap-earth-statistics/demo.js
```diff
@@ -0,0 +1,223 @@
+(async () => {
+
+    const topology = await fetch(
+            'https://code.highcharts.com/mapdata/custom/world-continents.topo.json'
+        ).then(response => response.json()),
+        landDayData = await fetch(
+            'https://cdn.jsdelivr.net/gh/highcharts/highcharts@a4b25b3845/samples/data/geoheatmap-land-sea-day-temp-august_2022.json'
+        ).then(response => response.json()),
+        data = [{
```
TorsteinHonsi, line 9: Can we rename this consant to `datasets` to avoid confusion with the often-unsed `series.data` option?

samples/maps/series-geoheatmap/turkey-fire-areas/demo.js
```diff
@@ -0,0 +1,709 @@
+(async () => {
+
+    const topology = await fetch(
+        'https://code.highcharts.com/mapdata/custom/world.topo.json'
+    ).then(response => response.json());
```
TorsteinHonsi, line 5: Is this used at all? This demo has only a twm and a geoheatmap series, so we probably shouldn't load the topoJSON.

test/karma-conf.js
```diff
@@ -86,7 +86,7 @@ function resolveJSON(js) {
 
         // Look for sources that can be matched to samples/data
         innerMatch = src.match(
-            /^(https:\/\/cdn.jsdelivr.net\/gh\/highcharts\/highcharts@[a-z0-9\.]+|https:\/\/www.highcharts.com)\/samples\/data\/([a-z0-9\-\.]+$)/
+            /^(https:\/\/cdn.jsdelivr.net\/gh\/highcharts\/highcharts@[a-z0-9\.]+|https:\/\/www.highcharts.com)\/samples\/data\/([a-z0-9\-_\.]+$)/
```
TorsteinHonsi, line 89: Can we change the file names instead? All the other files are kebab-case...

ts/Series/GeoHeatmap/GeoHeatmapSeries.ts
```diff
@@ -143,7 +204,45 @@ class GeoHeatmapSeries extends MapSeries {
...
+             * @sample maps/series-geoheatmap/turkey-fire-areas
+             *         Simple demo of GeoHeatmap interpolation
+             *
+             * @type      {number}
+             * @default   1
+             * @since     @next
+             * @product   highmaps
+             * @apioption plotOptions.geoheatmap.interpolationBlur
+             */
+            interpolationBlur: 1
```
TorsteinHonsi, line 245: The intension when we planned the `heatmap.interpoation` option, was that we first made it a simple boolean, but with the long-term possibility of expanding it as an object later. So instead of adding future options like `interpolationBlur`, `interpolationAlgorithm`, `interpolationSomething...`, we should now split it into two types.

A boolean enables or disables interpolation with default settings.

An object is used for enhanced control
```js
interpolation: {
    enabled: true,
    blur: 1
}
```

ts/Series/GeoHeatmap/GeoHeatmapSeries.ts
```diff
@@ -167,6 +272,22 @@ class GeoHeatmapSeries extends MapSeries {
...
 
+    public init(): void {
+        const series = this;
+
+        super.init.apply(series, arguments);
+
+        if (
+            Object.keys(pick(series.hcEvents, {}))
+                .indexOf('afterDataClassLegendClick') === -1
+        ) {
```
TorsteinHonsi, line 283: Instead of adding this event on the first init (and check if it is actually the first), I think it would be cleaner to add it to the class itself. See for example [how we do it wth the AreaRangeSeries](https://github.com/highcharts/highcharts/blob/46db0ddb81fc77e852fa1f1b19ccce1f04a7595a/ts/Series/AreaRange/AreaRangeSeries.ts#L649).

TorsteinHonsi (changes requested):
Great work! Added some request changes.

samples/maps/series-geoheatmap/turkey-fire-areas/demo.js
hubertkozik: Yes, it is used to create outlines of continents.

samples/maps/demo/geoheatmap-earth-statistics/demo.js
```diff
@@ -0,0 +1,224 @@
...
+            }
+        },
+
+        colorAxis: {},
+
+        tooltip: {
+            headerFormat: '<span style="font-size: 11px">Lon: {point.point.lon}° Lat: {point.point.lat}°</span><br/>',
+            pointFormatter() {
+                return `Value: ${this.options.value.toFixed(2)}`;
+            }
```
TorsteinHonsi, line 194: Avoid unnecessary use of formatter callbacks. Use this instead:

`pointFormat: 'Value: {point.value:.2f}'`

TorsteinHonsi (changes requested):

ts/Series/GeoHeatmap/GeoHeatmapSeries.ts
```diff
@@ -175,11 +281,418 @@ class GeoHeatmapSeries extends MapSeries {
...
+        // Create the extended object out of the boolean
+        options.interpolation = merge(
+            true,
+            GeoHeatmapSeries.defaultOptions.interpolation,
+            {
+                enabled: options.interpolation
+            }
+        );
+    }
+});
```
hubertkozik, line 694: I have created the extended object out of the boolean by using this event. I am not sure if that's an acceptable solution @TorsteinHonsi, but works pretty well.

TorsteinHonsi: This doesn't seem to work well with `series.update`. In the earth statistics demo, try adding this after chart creation:

```js
    setTimeout(function () {
        chart.series[0].update({
            interpolation: true
        });
    }, 1234);
```

I see that the `update` function merges the new, raw options into the existing options, so there you will get the boolean for of interpolation. Maybe you should just have a `getInterpolation` method that handles this, so you can do `this.getInterpolation().enabled` instead of `this.options.interpolation.enabled`. Performance wise it wouldn't matter since it's only called 2 or 3 times per render.

Another thing, when doing `merge` with the first argument set to true, we actually modify the default options, which is not what we want. Once we remove the `true` argument, I think this merge is okay.

hubertkozik: Added `getInterpolation` method - looks solid for me now.

TorsteinHonsi (approved):
🎉 🎉  Nice work!

PR #19126, opened 2023-06-09
bugfix/19121-measure-annotation-update

### LINKED ISSUE #19121: Measure Annotation moves back to initial position after customizing settings
#### Expected behaviour
The position of the `measure` annotation should not be altered by updating things unrelated to its position (like `lineWidth`).

**Demo:** https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/stock/stocktools/stocktools-thresholds
**Steps to reproduce:**
1. Add a measure annotation
2. Drag it to a different position
3. Change its `lineWidth` from 10 to 11
4. Click "Save"
The position went back to initial one.

This issue was reported in this forum topic: https://www.highcharts.com/forum/viewtopic.php?f=12&t=50985 where there's a nice GIF showing what's wrong.

#### Product version
Highcharts 11.1.0

#### Affected browser(s)
All?

jakubSzuminski:
Fixed #19121, `Measure Annotation` went back to its initial position after update.

**GIF of the problem is here:** https://www.highcharts.com/forum/viewtopic.php?f=12&t=50985

**The source of the problem:**
1. When we create a new `Measure Annotation` its initial position is set in the `annotation.userOptions`
2. When we drag the `Measure Annotation` we update its coordinates in `annotation.options`
3. When we update the annotation - `Annotation.update(userOptions, redraw)` - we initialize the annotation with `userOptions` (set initially), not with `options` (which were updated after dragging)

**Simple solution:** updating both `options` and `userOptions` after dragging.

### DIFF
--- a/samples/unit-tests/annotations/annotations-measure/demo.js
+++ b/samples/unit-tests/annotations/annotations-measure/demo.js
@@ -124,8 +124,9 @@ QUnit.test('#13664 - annotation measure on yAxis', function (assert) {
         }
     });
 
-    assert.close(
-        bbox.x, chart.annotations[0].shapesGroup.getBBox().x, 1,
+    assert.equal(
+        bbox.x,
+        chart.annotations[0].shapesGroup.getBBox().x,
         'The annotation should stay in the same place after update, #19121.'
     );
 });
--- a/samples/unit-tests/annotations/annotations-measure/demo.js
+++ b/samples/unit-tests/annotations/annotations-measure/demo.js
@@ -110,7 +110,6 @@ QUnit.test('#13664 - annotation measure on yAxis', function (assert) {
     // drag the annotation to the left
     controller.mouseDown(bbox.x + bbox.width / 2, bbox.y + bbox.height / 2);
     controller.mouseMove(bbox.x - 50, bbox.y);
-    controller.mouseMove(bbox.x - 50, bbox.y);
     controller.mouseUp();
 
     bbox = chart.annotations[0].shapesGroup.getBBox();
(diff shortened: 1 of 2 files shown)


jakubSzuminski:
Alternative solution could be adding a `Measure.update()` method, which would take care of this case (I'm not sure which solution's better).

```
* Update the measure annotation.
* @private
*/
public update(
     userOptions: DeepPartial<AnnotationOptions>,
     redraw?: boolean | undefined
 ): void {
     // Need to update annotation coordinates after dragging, #19121.
     userOptions.typeOptions = {
         point: {
             x: this.startXMin,
             y: this.startYMin
         },
         ...userOptions.typeOptions
     };
     super.update(userOptions, redraw);
 }
```

jakubSzuminski:
**Demo before fix:** https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/stock/stocktools/stocktools-thresholds
**Demo after fix:** https://jsfiddle.net/BlackLabel/Lp2gqm4o/

samples/unit-tests/annotations/annotations-measure/demo.js
```diff
@@ -102,4 +104,29 @@ QUnit.test('#13664 - annotation measure on yAxis', function (assert) {
         `Annotation's control points should be positioned in the middle of yAxis
         #17995`
     );
+
+    bbox = chart.annotations[0].shapesGroup.getBBox();
+
+    // drag the annotation to the left
+    controller.mouseDown(bbox.x + bbox.width / 2, bbox.y + bbox.height / 2);
+    controller.mouseMove(bbox.x - 50, bbox.y);
+    controller.mouseMove(bbox.x - 50, bbox.y);
```
pawelfus, line 113: As discussed on Slack: this means we need another solution

pawelfus (changes requested):

PR #19131, opened 2023-06-10
bugfix/incorrect-outline-offline-exporting

### LINKED ISSUE #17170: Wrongly positioned shadows while using the offline export.
#### Expected behaviour
Properly position the shadows while using offline exporting.

#### Actual behaviour
After exporting the chart to PDF via offline exporting, `dataLabels` are shifted and shadows are incorrectly positioned.

Export the chart to PDF and compare the result with the chart.
https://jsfiddle.net/BlackLabel/23ks15d0/

Workaround (disabling the shadows): https://jsfiddle.net/BlackLabel/jtr6382u/

#### Product version
 Highcharts JS v10.0.0 (2022-03-07)

#### Affected browser(s)
All

bm64:
Fixed #17170, data labels with outline or shadow were misplaced in PDF files exported using the offline exporting module.

### DIFF
--- a/samples/highcharts/exporting/offline-download-pdffont/demo.js
+++ b/samples/highcharts/exporting/offline-download-pdffont/demo.js
@@ -16,7 +16,13 @@ Highcharts.chart('container', {
         data: [1, 3, 2, 4],
         type: 'column',
         colorByPoint: true,
-        name: 'Αυτή είναι η σειρά'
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (approved):
Thanks! I did the visual test less intrusive, now only pending the CI before we can merge it.

PR #19133, opened 2023-06-12
Dashboards - Release Candidate

sebastianbochan:
- [x] - a fork from the `master`
- [x] - merge with `dashboards/main`
- [x] - demos
- [x] - lint
- [x] - lint demos
- [x] - cypress tests run
- [x] - karma tests


sebastianbochan:
Should we convert this branch into `master-dashboards` ?

karolkolodziej (approved):
HTMLElement looks weird:
image

test/karma-files.json
```diff
@@ -89,6 +90,7 @@
 	"samples/data/usdeur.js",
 	"samples/data/ohlc.js",
 	"samples/data/three-series-1000-points.js",
-	"samples/data/large-dataset.js"
+	"samples/data/large-dataset.js",
 
+	"code/dashboard.src.js"
```
pawelfus, line 95: Q: Shouldn't be loaded just one?

tools/gulptasks/jsdoc-classes.js
```diff
@@ -40,6 +40,8 @@ const SOURCE_GLOBS = [
     'js/Core/Time.js',
     'js/Core/Tooltip.js',
     'js/Core/Utilities.js',
+    'js/Dashboards/Board.js',
+    'js/Data/DataTable.js',
```
pawelfus, line 44: Shall we remove it from JSDoc task since we use TypeDoc?

tools/gulptasks/jsdoc-namespace.js
```diff
@@ -18,10 +18,11 @@ const TREE_FILE = 'tree-namespace.json';
 const TSCONFIG_FILE = path.join(LINT_DIRECTORY, 'tsconfig.json');
 
 const TARGET_DIRECTORIES = [
+    'dashboard',
```
pawelfus, line 21: Shall we remove it?

pawelfus (commented):
Just a few questions before the greenlight (see comments inline).

One more:
I see some changes in Highcharts-related files. Is that because `master` has already changed? Could you please double-check them @sebastianbochan ? For example `ts/Core/Chart/Chart.ts`, `ts/Core/Globals.ts` and `ts/Core/Utilities.ts`

bre1470 (changes requested):

ts/Extensions/DashboardPlugins/.eslintrc
```diff
@@ -0,0 +1,10 @@
+{
+    "extends": "../../.eslintrc",
```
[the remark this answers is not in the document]
pawelfus: Yes, nice catch!

test/karma-files.json
sebastianbochan: I have a problem with that because it does not matter in which direction we merge, (`rc` to `dashboards/main` or `dashboards/main` to `rc`) it always overwrites HC files with `dashboards/main`. Any ideas, how can we proceed it? I cannot imagine to get through all files and checking each difference

bre1470: Maybe @pawelfus can simulate after the requests a release procedure for Dashboards and I can do the same for Highcharts. This should give us an idea how many hidden issues we have.

ts/Data/Modifiers/RangeModifier.ts
```diff
@@ -194,6 +194,40 @@ class RangeModifier extends DataModifier {
...
 
+
+    /**
+     * Utility function that returns the first row index
+     * if the table has been modified by a range modifier
+     * @param {DataTable} table the table to get the offset from
+     *
+     * @return {number} The row offset of the modified table
+     */
+    public getModifiedTableOffset(table: DataTable): number {
```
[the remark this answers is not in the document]
sebastianbochan: In the current `dashboards/main` we have it here, so I guess that it will be a part of another PR or commit.

Please correct me if Im wrong ;)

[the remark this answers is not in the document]
bre1470: I will not correct you and instead create a ticket. :)

ts/Extensions/DashboardPlugins/.eslintrc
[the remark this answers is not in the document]
sebastianbochan: Should we do it in this PR as this task or during "adjust dashboards build / prod" task

ts/Data/Modifiers/RangeModifier.ts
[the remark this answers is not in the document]
sebastianbochan: Thanks! ;)

ts/Extensions/DashboardPlugins/.eslintrc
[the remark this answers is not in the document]
pawelfus: If it's simple, I would say just move it

test/karma-files.json
bre1470: I am talking about testing the distribution process. So not just looking at code, but transpiling and bundling it. I will give it a shot.

bre1470: I was not able to repeat the steps to create distributions for Highcharts or Dashboards. So we might need some adjustments later on that part of scripts.

bre1470 (approved):
LGTM

bre1470:
There is a change in `Core/Utilities` that looks like a fix for recursive issues in Dashboards, when class instances with references are used as values in options. This might be something that should go into `Dashboards/Utilities` instead.

pawelfus (changes requested):
See the discussion above (about the changes in Core/Utilities)

pawelfus (approved):
Thanks!

TorsteinHonsi (approved):
Looks good to me!

PR #19137, opened 2023-06-12
bugfix/17820-clip-rect-navigator-boost

### LINKED ISSUE #17820: Navigator with dataGrouping: false doesn't work with Boost
With Boost module imported, setting the dataGrouping to false on navigator results in a blank space inside navigator itself.

Last working version - **Highstock 8.2.2, Boost 8.2.2**

https://jsfiddle.net/BlackLabel/orwukp4h/

#### Workaround
Set the `navigator.series.boostThreshold` to `0`

#### Product version
Highstock 10.2.1
Boost 10.2.1

kamil-musialowski:
Fixed #17820, clip rect did not take into account boosted navigator.

Demo `.com`: https://jsfiddle.net/BlackLabel/g12drxjo/
Demo from branch: https://jsfiddle.net/BlackLabel/zm1bsdqv/

### DIFF
--- a/samples/unit-tests/boost/general/demo.js
+++ b/samples/unit-tests/boost/general/demo.js
@@ -329,5 +329,22 @@ QUnit[Highcharts.hasWebGLSupport() ? 'test' : 'skip'](
             `After setting the axis position manually, the boost clip-path
             shouldn\'t be bigger than the axis size.`
         );
+        chart.update({
+            chart: {
+                inverted: false
+            },
+            navigator: {
+                enabled: true,
+                height: 80,
+                series: {
+                    boostThreshold: 1,
+                    color: 'red',
+                    type: 'line',
+                    dataGrouping: {
+                        enabled: false
+                    }
+                }
+            }
+        });
     }
 );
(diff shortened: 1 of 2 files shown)


pawelfus:
Thanks!
The problem with this solution is setting min on the x-axis, compare:
- current: https://jsfiddle.net/BlackLabel/orwukp4h/3/
- PR: https://jsfiddle.net/BlackLabel/orwukp4h/2/

What do you think about this @TorsteinHonsi ?

pawelfus:
Ping @TorsteinHonsi

TorsteinHonsi:
@pawelfus There's also the case of multiple panes: https://github.com/highcharts/highcharts/commit/414ff2a56943c7eb4efdbeab46d8cd7fa071560e

If all the series - including the navigator - are rendering to the same canvas, I think we have no other choice than to extend the clip rectangle around all series.

If they are separate canvases, we are probably covered by the fix for multiple panes.

pawelfus:
Thanks!
It is almost perfect with multiple panes: https://jsfiddle.net/BlackLabel/orwukp4h/7/ - resize window/navigator to get the series in the navigator. Will do

pawelfus:
> If all the series - including the navigator - are rendering to the same canvas, I think we have no other choice than to extend the clip rectangle around all series.

We can introduce canvas per pane (yAxis?). It will drastically reduce the number of canvases for multi-series and multi-pane charts. It can complicate code a little: adds new target=Axis to other targets (series and chart). Let's pay close attention to similar issues in the future (navigator + boost + multi-pane).

kamil-musialowski:
Fix moved to #19880.

PR #19138, opened 2023-06-13
Update your-first-chart.md

peaceabu:


TorsteinHonsi:
Thanks for the PR! However this change is not consistent with the content of the rest of our pages, so we'll leave it for now.

PR #19142, opened 2023-06-13
Db/datapool connector rename

pawellysy:


karolkolodziej (changes requested):
Update Cypress demos.

karolkolodziej (approved):

sebastianbochan (approved):

ts/Data/DataPool.ts
```diff
@@ -181,17 +181,17 @@ class DataPool implements DataEvent.Emitter {
...
-     * @param {string} name
+     * @param {string} id
      * Name of the connector.
      *
      * @return {Promise<Data.DataTable>}
      * Returns the connector table.
      */
     public getConnectorTable(
-        name: string
+        id: string
```
bre1470, line 191: It is better to write here `connectorId` as each DataTable has also an `id`.

bre1470 (approved):
One thing to change, otherwise it looks good. 👍

PR #19144, opened 2023-06-13
Improved performance of climate demo.

bre1470:


karolkolodziej (changes requested):
Conversion from C to F still takes significant time, is there anything to improve?

bre1470:
It looks like it is a bug in the click event. The toggle triggers the click event three times.

karolkolodziej (approved):
@bre1470 You are right, let's treat it in a separate PR ;)

samples/dashboards/demos/climate/demo.js
```diff
@@ -81,10 +81,13 @@ async function setupBoard() {
                         text: 'Fahrenheit',
                         events: {
                             click: function () {
+                                console.log('CLICK');
```
pawellysy, line 84:
```suggestion

```

pawellysy (approved):
Minor change, Apart from that LGTM!

pawelfus (approved):
Thanks!

sebastianbochan (approved):

PR #19145, opened 2023-06-13
bugfix/19069-stacked-column-bar-stroke

### LINKED ISSUE #19069: Inconsistent stacked column chart behavior after upgrading Highcharts from v10 to v11 - 0 value bars displayed
We have recently upgraded Highcharts from version 10 to 11 and observed an issue with stacked column charts which are inconsistent with version 10.

#### Expected behaviour
The bars corresponding to the 0 value don't display in the stacked columns chart.

#### Actual behaviour
When we use stacked columns with `chart.styledMode === true && yAxis.reversedStacks === false`, and on the top is a point that has 0 value then it's displayed anyway, even if explicitly assigning `plotOptions.minPointLength` to 0.

The following two reproductions both use the same options.
(Bug occurring) V11: https://jsfiddle.net/027nxk3z/
(Expected) V10: https://jsfiddle.net/027nxk3z/1/

#### Product version
Highcharts v11.0.1

#### Affected browser(s)
Chrome 113.0.5672.126 (Official Build) (x86_64)

jakubjanuchta:
Fixed #19069, columns and bars had unnecessary stroke in styled mode.

### DIFF
--- a/css/highcharts.css
+++ b/css/highcharts.css
@@ -484,8 +484,9 @@ text.highcharts-data-label {
     stroke: var(--highcharts-neutral-color-100);
 }
 
-.highcharts-column-series rect.highcharts-point {
-    /* rect to prevent stroke on 3D columns */
+.highcharts-column-series path.highcharts-point,
+.highcharts-bar-series path.highcharts-point {
+    /* path to prevent stroke on 3D columns and bars */
     stroke: var(--highcharts-background-color);
 }
 


TorsteinHonsi (approved):
Thanks!

TorsteinHonsi:
Pending CI before merging

rainx:
Hi team, any plan for releasing this bug fix?

PR #19146, opened 2023-06-13
Db/cell options reponsive docs

pawellysy:


ts/Dashboards/Layout/Cell.ts
```diff
@@ -584,9 +584,6 @@ class Cell extends GUIElement {
 
 namespace Cell {
 
-    /**
-     * @internal
-     **/
     export interface CellResponsiveOptions {
         width: (string|number);
```
karolkolodziej, line 588: Width needs to be described, can `visible` be removed?

karolkolodziej (changes requested):

ts/Dashboards/Layout/Cell.ts
pawellysy: Sure :D

karolkolodziej (approved):

sebastianbochan:
Do we have an example of defining `responsiveOptions` somewhere? If not, would be good to add it.

sebastianbochan (approved):

PR #19147, opened 2023-06-13
Db/climate dg columns hidden

pawellysy:


bre1470 (approved):

samples/dashboards/demos/climate/demo.js
```diff
@@ -808,13 +809,15 @@ async function setupBoard() {
                         headerFormat: 'Average Temperature °C'
                     },
                     TNF: {
-                        headerFormat: 'Average Temperature °F'
+                        headerFormat: 'Average Temperature °F',
+                        show: false
```
sebastianbochan, line 813: I would suggest to replace `show` with `visibile` or `visibility` like other HC parameters are.

Another option use `isHidden`.

sebastianbochan (changes requested):

karolkolodziej (approved):

samples/dashboards/demos/climate/demo.js
pawellysy: Well, me might do it, but it is not a part of this PR. Here I just use the API option. If you'd like to see it renamed, please, report it seperately.

sebastianbochan (approved):

PR #19152, opened 2023-06-14
db-custom-html-component

karolkolodziej:
Added custom HTML component with AST.

### DIFF
--- a/docs/dashboards/custom-component.md
+++ b/docs/dashboards/custom-component.md
@@ -1,5 +1,7 @@
 Custom Dashboards components
 ===
+
+## Custom YouTube Component
 This article shows how to create custom Dashboards Component, in this example YouTube Component.
 
 Note, that to create the custom component we are using ES6, to use the `class` and `extends` keywords, which makes creating custom class much easier.
@@ -78,3 +80,32 @@ Dashboards.board({
 });
 ```
 [The live example can be found here](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/samples/dashboards/demos/custom-component).
+
+---
+## Custom HTML Component
+The basic HTML component is described in the [Types of Components](https://www.highcharts.com/docs/chart-and-series-types/dashboards/types-of-components) page. The basic one is very flexible, but requires a lot of configuration. In this example, we will create a custom HTML component, which will be more flexible and easier to use.  
+To do so, we will use the `Highcharts.AST` class, which will be used to parse the string type HTML into the AST-like object.
+
+First we can declare our custom HTML in HTML file as shown in the code snippet below or you can declare that as a string in the code.
+``` HTML
+<div id="custom-html">
+    <h1>Your custom HTML</h1>
+    <p>Here you can add any HTML you like.</p>
+</div>
+```
+Then we can reference it in the code and parse it into the AST object like below:
+
+```JS
+const customHTML = document.getElementById('custom-html').outerHTML,
+    AST = Highcharts.AST;
+
+Dashboards.board('container', {
+    ...
+    components: [{
+        type: 'HTML',
+        cell: 'dashboard-col-0',
+        elements: new AST(`${customHTML}`).nodes
+    }, ....]
+});
+```
+Check out the [live example here](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/samples/dashboards/demos/custom-html-component).
--- a/docs/dashboards/types-of-components.md
+++ b/docs/dashboards/types-of-components.md
@@ -28,6 +28,7 @@ Code snippet:
         }]
     }
 ```
+Also please check the [Custom Component](http://www.highcharts.com/docs/dashboards/custom-component) section below, where you can find a more flexible way to define HTML components.
 
 ### Highcharts Component
 The option to include a Highcharts chart in one of the components is available out of the box. Here is the set of files that need to be included to make the Highcharts component work.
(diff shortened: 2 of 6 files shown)


samples/dashboards/demos/custom-html-component/demo.js
```diff
@@ -0,0 +1,30 @@
...
+                    id: 'dashboard-col-1'
+                }]
+            }]
+        }]
+    },
+    components: [{
+        type: 'HTML',
+        cell: 'dashboard-col-0',
+        elements: new AST(`${customHTML}`).nodes
+    }, {
```
pawellysy, line 21: It's not really custom component. Why don't we create a **real** custom component, where instead of adding `element` you add new `type: 'HTMLString'` and 
```html: myHtml```

**Review by pawellysy — CHANGES_REQUESTED** [review]

**Review by sebastianbochan — APPROVED** [review]

File: docs/dashboards/custom-component.md [hunk 2]
```diff
@@ -77,4 +79,66 @@ Dashboards.board({
...
+        this.getCustomHTML();
+        return this;
+    }
+
+    getCustomHTML() {
+        const options = this.options;
+        if (options.id) {
+            const customHTML = document.getElementById(options.id).outerHTML;
+
+            this.options.elements = new AST(`${customHTML}`).nodes;
```
pawellysy, line 110: Can't it just be?

```suggestion
            this.options.elements = new AST(customHTML}).nodes;
```

docs/dashboards/custom-component.md
```diff
@@ -77,4 +79,66 @@ Dashboards.board({
...
+        } else if (options.html) {
+            this.options.elements = new AST(options.html).nodes;
+        }
+    }
+}
+
+ComponentRegistry.registerComponent(CustomHTML);
+```
+
+This custom component can generate HTML elements to you dashboards by:
```
pawellysy, line 120:
```suggestion
This custom component can generate HTML elements in the dashboards by:
```

docs/dashboards/custom-component.md
```diff
@@ -77,4 +79,66 @@ Dashboards.board({
     }]
 });
 ```
-[The live example can be found here](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/samples/dashboards/demos/custom-component).
\ No newline at end of file
+[The live example can be found here](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/samples/dashboards/demos/custom-component).
+
+---
+## Custom HTML Component
+The basic HTML component described in the [Types of Components](https://www.highcharts.com/docs/chart-and-series-types/dashboards/types-of-components) page is very flexible, but requires a lot of configuration. In this example, we will create a custom HTML component, which will be more flexible and easier to use.
```
pawellysy, line 86: Is it more flexible? IMO it is less flexible, but it is easier to use 🤔

docs/dashboards/custom-component.md
```diff
@@ -77,4 +79,66 @@ Dashboards.board({
...
+
+This custom component will be extended from the basic HTML component, so we will need to import the `HTMLComponent` class. The easiest way to do so is through the `ComponentRegistry` as shown below. We also will use the `Highcharts.AST` class, which will be used to parse the string type HTML into the AST-like object.
```
pawellysy, line 88:
```suggestion
This custom component will extend basic HTML component, so we need to import the `HTMLComponent` class. The easiest way to do so is through the `ComponentRegistry` as shown below. We also will use the `Highcharts.AST` class, which will be used to parse the string type HTML into the AST-like object.
```

docs/dashboards/custom-component.md
```diff
@@ -77,4 +79,66 @@ Dashboards.board({
...
+
+Then we can create our custom class, which will inherit from the `HTMLComponent` class. The only thing we need to do is to create a method to extract the HTML from the options and parse it into the AST-like object. In the example this method is called `getCustomHTML` and it assigns the element generated by the AST to the `elements` property. The only thing left is to register the new component in the `ComponentRegistry` and we are ready to use it.
```
pawellysy, line 90:
```suggestion
Then we can create our custom class, which extends the `HTMLComponent` class. The only thing we need to do, is to create a method to extract the HTML from the options and parse it into the AST-like object. In the example this method is called `getCustomHTML` and it assigns the element generated by the AST to the `elements` property. The only thing left is to register the new component in the `ComponentRegistry` and we are ready to use it.
```

docs/dashboards/custom-component.md
```diff
@@ -77,4 +79,66 @@ Dashboards.board({
...
+}
+
+ComponentRegistry.registerComponent(CustomHTML);
+```
+
+This custom component can generate HTML elements to you dashboards by:
+- referencing the HTML element by its `id` attribute
+- passing the HTML as a string to the `html` property
+
+Use of this component is shown below:
```
pawellysy, line 124: Maybe, to be consistent across the docs, both YouTube and CustomHTML components should have the same structure and text?

docs/dashboards/types-of-components.md
```diff
@@ -28,6 +28,7 @@ Code snippet:
         }]
     }
 ```
+Also please check the [Custom Component](http://www.highcharts.com/docs/dashboards/custom-component) section below, where you can find a more flexible way to define HTML components.
```
pawellysy, line 31: Is it more flexible tho?

pawellysy (changes requested):
Minor polishing

pawellysy (approved):
LGTM

PR #19153, opened 2023-06-14
bugfix/19078-proximate-legend-enabling

### LINKED ISSUE #19078: Legend with layout 'proximate' issues with enabling/disabling
#### Actual behaviour
**Legend initially enabled** - disabling it through update throws error to the console
**Legend initially disabled** - enabling it positions all of the legend items in the same place 

Change the initial value of `legend.enabled` and enable/disable it. Watch the console and legend layout.
https://jsfiddle.net/BlackLabel/jkxbwsep/

#### Product version
Highcharts v11.0.1 (latest)

jakubSzuminski:
Fixed #19078, proximate legend did not work after updating the `legend.enabled` property.

**Demo of the problem:** https://jsfiddle.net/BlackLabel/1cmx87k9/
**Demo after fix:** https://jsfiddle.net/BlackLabel/wftjvga3/

**Source of the problem:**
- `proximatePositions()` was called even when `legend.enabled = false` which caused multiple errors
- `proximatePositions()` and `positionItems()` functions were called only in the initial chart render, not after updating

### DIFF
--- a/samples/unit-tests/legend/layout/demo.js
+++ b/samples/unit-tests/legend/layout/demo.js
@@ -1,6 +1,7 @@
 QUnit.test('Legend layout', function (assert) {
     var chart = Highcharts.chart('container', {
         legend: {
+            enabled: false,
             layout: 'proximate',
             align: 'right'
         },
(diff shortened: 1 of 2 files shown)


hubertkozik (changes requested):
One comment on the current solution below. Also, please add a unit test to cover this case.

TorsteinHonsi (approved):
Thanks!

PR #19156, opened 2023-06-14
Website/hp animations

nasin:
Hi, we're redesigning our homepage hero layout for the dashboard launch, so I changed some of the homepage animations and added a few more for the launch.


TorsteinHonsi (approved):
Thanks!

PR #19159, opened 2023-06-15
db/fix-ts-karma-tests

sebastianbochan:


test/typescript-karma/Dashboards/Component/tests.js
```diff
@@ -76,7 +76,8 @@ test('Board without data connectors and HighchartsComponent update', async funct
                 type: 'Highcharts',
                 chartOptions: {
                     title: {
-                        text: void 0
+                        text: void 0,
+                        style: {}
```
karolkolodziej, line 80: Why this one is required?

karolkolodziej (approved):

test/typescript-karma/Dashboards/Component/tests.js
sebastianbochan: Because of Highcharts changes. Right now the style param is required in the TS type.

We discussed it with @bre1470, and adding style is the simplest way of solving problem. However in the future we should consider using optional `style` in HC, like it was in the past.

ts/Data/DataPoolOptions.ts
```diff
@@ -34,7 +34,7 @@ export interface DataPoolOptions {
 export interface DataPoolConnectorOptions
 <T extends keyof DataConnectorTypes = keyof DataConnectorTypes> {
     id: string;
-    options: DataConnectorTypes[T]['prototype']['options'];
+    options: DeepPartial<DataConnectorTypes[T]['prototype']['options']>;
```
bre1470, line 37: Please use `DeepPartial<DataPoolConnectorOptions>`.

bre1470 (changes requested):
One type issue, otherwise it looks good.

ts/Data/DataPoolOptions.ts
sebastianbochan: It causes some errors.

bre1470: Maybe just remove it. I tested locally and everything works without DeepPartial. What was the reason to add DeepPartial?

test/typescript-karma/Dashboards/Component/tests.js
```diff
@@ -30,7 +30,7 @@ const eventTypes = [
 ];
 
 const registeredEvents = [];
-window.registeredEvents = registeredEvents;
+// window.registeredEvents = registeredEvents;
```
karolkolodziej, line 33: This was added for internal test reasons, can be removed ;)

karolkolodziej (approved):

bre1470 (approved):
Thank you! Looks good. 👍

PR #19160, opened 2023-06-15
bugfix/19093-svg-filter-shadow-issue

### LINKED ISSUE #19093: Line and shadow issue when points have same `y` value and `shadow` is enabled
#### Expected behaviour
Line is rendered on the chart as well as the shadow

#### Actual behaviour
The line between points isn't rendered, when all points have same `y` value

[https://jsfiddle.net/BlackLabel/t6hypbn7/](https://jsfiddle.net/BlackLabel/t6hypbn7/)
Switch `shadow` to false, the line will render properly
#### Product version
since Higcharts v11.0.0
#### Affected browser(s)
Google Chrome 113.0.5672.129
Microsoft Edge 114.0.1823.37
Firefox 113.0.2 (line is rendered, the shadow isn't)
Safari on iOS 16.5

jedrzejruta:
Fixed #19093, SVG `<filter>` issue with shadow enabled on line series when points had the same `x` or `y` value.

### DIFF
--- a/ts/Series/Line/LineSeries.ts
+++ b/ts/Series/Line/LineSeries.ts
@@ -184,16 +184,13 @@ class LineSeries extends Series {
                 // zone (1) #3932
                     .shadow(
                         (i < 2) &&
-                        (options.shadow ?
-                            // If shadow is defined, call function with
-                            // `filterUnits: 'userSpaceOnUse'` to avoid known
-                            // SVG filter bug (#19093)
-                            (isObject(options.shadow) ?
-                                merge(
-                                    options.shadow,
-                                    { filterUnits: 'userSpaceOnUse' }
-                                ) : { filterUnits: 'userSpaceOnUse' }
-                            ) : options.shadow
+                        options.shadow &&
+                        // If shadow is defined, call function with
+                        // `filterUnits: 'userSpaceOnUse'` to avoid known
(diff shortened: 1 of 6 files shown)


ts/Series/Line/LineSeries.ts
```diff
@@ -179,9 +180,22 @@ class LineSeries extends Series {
...
+                            // `filterUnits: 'userSpaceOnUse'` to avoid known
+                            // SVG filter bug (#19093)
+                            (isObject(options.shadow) ?
+                                merge(
+                                    options.shadow,
+                                    { filterUnits: 'userSpaceOnUse' }
+                                ) : { filterUnits: 'userSpaceOnUse' }
+                            ) : options.shadow
+                        )
+                    );
```
TorsteinHonsi, line 198: I'm wondering if this can be shortened. Didn't test it, but something like this?
```js
.shadow(
    (i < 2) &&
    options.shadow &&
    merge(
        { filterUnits: 'userSpaceOnUse' },
        isObject(options.shadow) ? options.shadow : {}
    )
);
```

jedrzejruta: Indeed, can be shortened; tested in local environment and works well.

TorsteinHonsi (approved):

PR #19163, opened 2023-06-15
bugfix/19134-drillup-color-change

### LINKED ISSUE #19134: Series color changes on drillup when colorByPoint is false
#### Expected behaviour
When returning to the top level series after drilldown the color of the top level series should be the same as the initial color.

#### Actual behaviour
When returning to the top level series after drilldown the color of the top level series cycles to the next color. This occurs when the `colorByPoint` option is not set or `false`.

https://jsfiddle.net/cxywousb/

#### Product version
Highcharts 11.1.0

#### Affected browser(s)
Chrome, Firefox, Safari

Vazonik:
Fixed #19134, color and symbol changed constantly for the same series after every drill-up.

### DIFF
--- a/samples/unit-tests/drilldown/drilldown-general/demo.js
+++ b/samples/unit-tests/drilldown/drilldown-general/demo.js
@@ -330,8 +330,17 @@ QUnit.test('Chart type update after drilldown', function (assert) {
         }
     });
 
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (approved):
Thanks!

PR #19165, opened 2023-06-16
Docs: Improved API to support URL history.

bre1470:
This only works online and locally with something like `python3 -m http.server -b 127.0.0.1` because of iframe CORS.

How to test on MacOS:
- `npx gulp dashboards/api-docs`
- `python3 -m http.server -b 127.0.0.1`
- `open http://127.0.0.1:8000/build/api/dashboards/index.html`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,7 +18,7 @@
         "@babel/preset-env": "^7.20.2",
         "@highcharts/highcharts-assembler": "github:highcharts/highcharts-assembler#v1.5.1",
         "@highcharts/highcharts-declarations-generator": "github:highcharts/highcharts-declarations-generator#v1.1.26",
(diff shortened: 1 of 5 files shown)


karolkolodziej (approved):
Perhaps it's worth mentioning the Python server somewhere in the `CONTRIBUTING.md` file?

bre1470:
It is not like it is breaking, when opened via `file://`. Just the outer frame hash history gets not updated because of cross-origin restrictions.

pawelfus (commented):
Thanks!

Can we store the hash too? For example:

- visit page: http://127.0.0.1:8000/build/api/dashboards/index.html#interfaces/Extensions_DashboardPlugins_HighchartsComponent.HighchartsComponent.Options
- click `cell` option
- result: url is missing `#cell`

pawelfus (approved):
Thanks!

PR #19166, opened 2023-06-16
feature/typed arrays

### LINKED ISSUE #14119: Splitting data x and y arrays
TypedArrays are supported with syntax:
```js
{
  data: new Uint8Array([0, 5, 3, 5])
}
```
And the following works too for adding X values to the data:
```js
{
  data: [[3, 0], [4, 5], [5, 3], [6, 5]]
}
```
It'd be nice if we could do: 
```js
{
  dataX: new Uint32Array([3, 4, 5, 6])  // Unix timestamp column to be formatted / parsed as date
  dataY: new Uint8Array([0, 5, 3, 5])  // Value
}
```
Columnar formats (like Apache Arrow) make it easy to retrieve data column by column as a fixed length, immutable array.
This way we'd be able to skip `zip()` or other helper data structures, I assume Highcharts could work on TypedArrays directly.

TorsteinHonsi:
Added support for typed data arrays.

Closes #14119

### DIFF
--- a/samples/highcharts/studies/typed-arrays/demo.css
+++ b/samples/highcharts/studies/typed-arrays/demo.css
@@ -1,0 +1,5 @@
+#container {
+    max-width: 800px;
+    height: 400px;
+    margin: 1em auto;
+}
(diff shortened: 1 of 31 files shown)


TorsteinHonsi:
Closing this, we won't add an object form of `series.options.data`. Instead, typed arrays will be supported through the [new data table approach](https://github.com/highcharts/highcharts/pull/20530).

PR #19172, opened 2023-06-19
feature/a11y-aria-atomic

kamil-musialowski:
Added `aria-atomic: true` for automated test tools.

### DIFF
--- a/ts/Accessibility/Utils/Announcer.ts
+++ b/ts/Accessibility/Utils/Announcer.ts
@@ -113,7 +113,8 @@ class Announcer {
 
         attr(div, {
             'aria-hidden': false,
-            'aria-live': type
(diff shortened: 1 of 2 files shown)


oysteinmoseng (approved):
Thanks!

TorsteinHonsi:
Thanks! But... does it work? This is not part of the `AST.allowedAttributes`, so either it is filtered out, or the AST is not doing what it is supposed to do. Or something else is happening that I don't understand.

oysteinmoseng:
It does indeed work.. The code just creates a div and then uses `attr` on it directly, so perhaps the AST functionality doesn't get to shine here? Not sure if the AST is even needed for this regardless, we just want to create an HTML element.

PR #19176, opened 2023-06-19
demo/19162-wrong-timeline-docs

### LINKED ISSUE #19162: Timeline chart points default at y: 1 even when explicitly stating y value in series data
#### Expected behaviour
Timeline chart points should be defaulted to y: 0 
Timeline chart points should respect the y value in the initial data series

This is especially an issue when combining charts e.g. line chart with timeline chart

#### Actual behaviour
Timeline chart data defaults to to y: 1 even when explicitly stating y: 0 in the data series

After updating the chart using chart.update({series: ...}) the stated y value is respected

image

Chart below shows the issue - I've left a commented out example of the update function 'fixing' the issue
https://jsfiddle.net/jso2nwea/40/

Highcharts ver: v11.1.0

jedrzejruta:
Fixed #19162, `timeline` docs refered to non-existing `point.y` property, last demo on page was not working.

### DIFF
--- a/docs/chart-and-series-types/timeline-series.md
+++ b/docs/chart-and-series-types/timeline-series.md
@@ -85,8 +85,8 @@ Use the `alternate` property in the `dataLabels` configuration, for placing the 
 Other configuring options
 -------------------------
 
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (approved):
Thanks!

PR #19177, opened 2023-06-19
db/climate-update

bre1470:


goransle (approved):

samples/dashboards/demos/climate/demo.js
```diff
@@ -179,32 +185,11 @@ async function setupBoard() {
...
-                                    date.getUTCMonth(),
-                                    date.getUTCDate() + 10
-                                ));
-                        }
-
-                        return dates;
-                    }()),
-                    showInNavigator: false,
+                    data: [
+                        [Date.UTC(1951, 0, 5), 0],
```
karolkolodziej, line 190: how about using something similar to `activeSelection` but like `dataExtremes`?

samples/dashboards/demos/climate/demo.js
```diff
@@ -160,6 +160,12 @@ async function setupBoard() {
         components: [{
             cell: 'time-range-selector',
             type: 'Highcharts',
+            // connector: {
```
karolkolodziej, line 163: Is that a placeholder for the future or can be removed?

karolkolodziej (approved):

pawellysy (approved):

samples/dashboards/demos/climate/demo.js
```diff
@@ -844,6 +829,7 @@ async function setupBoard() {
                 Date: null
             },
             sync: {
+                // extremes: true,
```
sebastianbochan, line 832: To remove

sebastianbochan (changes requested):

samples/dashboards/demos/climate/demo.js
bre1470: It is not working, so I guess it needs to be removed. :/

samples/dashboards/demos/climate/demo.js
bre1470: It is now `activeTimeRange` because of the component-id `time-range-selector`.

karolkolodziej (approved):

sebastianbochan (approved):

PR #19180, opened 2023-06-20
bugfix/19035-tooltip-pointer-events

### LINKED ISSUE #19035: Tooltip tap/click stops working after Pinch Zoom on Safari/iPad
#### Expected behaviour

The tooltip should work after pinch zoom just like it's working after zooming with the buttons or in the web browser.

#### Actual behaviour
On mobile browsers, after the pinch zoom, the click event on the tooltip is not registered. Please run the attached code.

Attached

#### Product version
Highcharts JS v10.2.0
 Highmaps JS v11.0.1 

#### Affected browser(s)
Safari on IPAD
[highmaps issue.zip](https://github.com/highcharts/highcharts/files/11541572/highmaps.issue.zip)

jakubSzuminski:
Fixed #19035, tooltip tap or click stopped working after pinch zoom in iOS.

**Demo** of the problem: https://jsfiddle.net/BlackLabel/hospxk7e/
1. Hover over one of the points till the tooltip is shown
2. Hover under the chart so that the tooltip hides
3. Notice that the entire area under the tooltip is "non-hoverable" once it was covered by the tooltip

### DIFF
--- a/ts/Core/Tooltip.ts
+++ b/ts/Core/Tooltip.ts
@@ -408,13 +408,7 @@ class Tooltip {
         const tooltip = this,
             styledMode = this.chart.styledMode,
             options = this.options,
-            doSplit = this.split && this.allowShared,
-            pointerEvents = (
-                options.style.pointerEvents ||
-                (
-                    this.shouldStickOnContact() ? 'auto' : 'none'
-                )
-            );
+            doSplit = this.split && this.allowShared;
 
         let container: globalThis.HTMLElement,
             renderer: SVGRenderer = this.chart.renderer;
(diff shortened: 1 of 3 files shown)


jakubSzuminski:
**Demo after fix:** https://jsfiddle.net/BlackLabel/az1nqj7g/

ts/Core/Tooltip.ts
```diff
@@ -449,7 +449,7 @@ class Tooltip {
                 css(container, {
                     position: 'absolute',
                     top: '1px',
-                    pointerEvents,
+                    pointerEvents: 'none',
```
hubertkozik, line 452: I would add a short description with the issue number to describe, why `'none'` is needed here.

hubertkozik (approved):
Good job!

TorsteinHonsi (approved):

PR #19182, opened 2023-06-20
db/adjust-demos-integration

sebastianbochan:
- [x] - replace `demos` with `demo`
- [x] - update paths in cypress tests
- [x] - update docs
- [x] - update details files


samples/dashboards/demo/climate/demo.js
```diff
@@ -390,7 +390,8 @@ async function setupBoard() {
                 },
                 tooltip: {
                     enabled: true,
-                    positioner: function (_width, _height, axisInfo) {
+                    distance: -60,
```
karolkolodziej, line 393: Shouldn't be part of this commit ;) 
But if you would like to leave it, remove the positioner.

karolkolodziej (changes requested):
- [ ] I think I tested sth and accidentally added more items to CSV for the `dashboards/cypress/component-datagrid`- the `Tuna` should be the last one 🙈

pawelfus (approved):
Thanks!

karolkolodziej (approved):

sebastianbochan:
> * [x]  I think I tested sth and accidentally added more items to CSV for the `dashboards/cypress/component-datagrid`- the `Tuna` should be the last one 🙈

It looks like breaks the cypress tests.

pawellysy (approved):

bre1470 (approved):
Looks good. 👍

goransle:
The visual test fails as it can no longer find the reference image because of the changing paths. Unless we want to write a script that tries to find path changes in the git log and do some magic with that‚ I suggest doing a manual comparison using the [test artifacts](https://app.circleci.com/pipelines/github/highcharts/highcharts/22170/workflows/da353831-6e4b-45bb-a7ce-d0ab65d0fb38/jobs/154939/artifacts)

goransle (approved):

PR #19183, opened 2023-06-20
db/datagrid-default-height

karolkolodziej:
Changed how the height is set in the datagrid.

### DIFF
--- a/ts/DataGrid/DataGrid.ts
+++ b/ts/DataGrid/DataGrid.ts
@@ -794,19 +794,13 @@ class DataGrid {
     public getDataGridSize(): number {
         const grid = this,
             options = grid.options,
-            { height } = grid.container.getBoundingClientRect(),
-            extraPixelsForBorders = 2;
+            { height } = grid.container.getBoundingClientRect();
 
         // If the container has a height declared in CSS, use that.
         if (height > 2) {
             return height;
         }
         // Use the default height if the container has no height declared in CSS
-        // Check if the column header is enabled.
-        if (options.columnHeaders.enabled) {
-            return options.defaultHeight +
-                options.cellHeight + extraPixelsForBorders;
-        }
         return options.defaultHeight;
     }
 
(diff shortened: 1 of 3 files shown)


pawellysy (changes requested):
- Go to `/dashboards/demos/minimal` demo. 
- Open Edit Mode. 
- Change the height of DG to show only 2 rows.
- try scrolling

The DG Won't scroll

ts/DataGrid/DataGrid.ts
```diff
@@ -1191,16 +1185,44 @@ class DataGrid {
...
+    /**
+     * If the grid is in the parent container that has margins, calculate the
+     * height of the margins.
+     * @internal
+     *
+     * @param  height
+     * The height of the parent container.
+     */
+    private getMarginHeight(height: number): number {
+        return height - this.gridContainer.getBoundingClientRect().height;
```
sebastianbochan, line 1216: How about using `getStyle()` with the margin or padding param?

sebastianbochan (changes requested):

ts/DataGrid/DataGrid.ts
karolkolodziej: I can't get the correct values using the `getStyle` method.  I'm afraid it has to stay as it is.

pawellysy (approved):
LGTM

ts/DataGrid/DataGrid.ts
```diff
@@ -1191,16 +1185,48 @@ class DataGrid {
...
+    }
+
+    /**
+     * Internal method used to get the height of the header.
+     * @internal
+     */
+    private getHeaderHeight(): number {
+        return this.options.cellHeight;
```
sebastianbochan, line 1228: Do we need this getter? it returns the reference only.

sebastianbochan (changes requested):

ts/DataGrid/DataGrid.ts
```diff
@@ -1191,16 +1185,40 @@ class DataGrid {
         height?: number | string | null
     ): void {
         if (width) {
-            this.innerContainer.style.width = width + 'px';
+            if (typeof width === 'string') {
+                this.innerContainer.style.width = width;
+            } else {
+                this.innerContainer.style.width = width + 'px';
```
sebastianbochan, line 1191: I suggest using `this.innerContainer.style.width = parseFloat(width)`;

sebastianbochan (approved):

Issue #14024, opened 2020-08-10
Allow developers to specify different symbols for upper/lower markers in dumbbell charts


hardingalexh:
# Description of the feature
[Dumbbell series](https://api.highcharts.com/highcharts/series.dumbbell) allow a developer to choose different colors for their low and high markers as `color` and `lowColor`. However, there is not an equivalent for `marker`, which allows the developer to choose which symbol is rendered at the ends of the dumbbell. Allowing a developer to alter `lowMarker` using identical syntax to `marker` would allow new features like directional arrows (circle to right arrow) or other custom features for display.

# Library related to the feature
Highcharts, Highcharts-more

# Proof of Concept/Live example for the feature
For example, the ability to render a [chart like this](https://imgur.com/kXf8jK6).

----

**You can vote for this feature by a adding thumbs-up reaction to this post.**

sebastianbochan:
Hi @hardingalexh,
Thank you for the idea!

FYI @raf18seb

pawellysy:
unstale

gpspake:
Bump. This is needed for accessibility. Is there no way to accomplish this currently 🤔? 
We have a dumbell chart that shows an increase or decrease over two periods. The point for the initial period is a black circle. If the amount has increased, the connector and the upper point are green. If it has decreased, the connector and the lower point are red.
What we want to do is use triangle and upside down-triangle for the second period point to indicate whether it has increased or decreased but we want to keep the initial point a circle.

raf18seb:
Hi @gpspake, thanks for popping this up!

This feature request is getting more popular. Imo the Highcharts dumbbell series should handle a case like this in an easy way. I personally came across the same problem with a very basic chart like this one
![image](https://user-images.githubusercontent.com/23310742/233168084-6d76a49f-08a1-4ed8-8475-968cf62af750.png)


We'll discuss adding this feature and provide more info soon :)

raf18seb:
We'll add it in one of the next releases. No ETA yet, but I'm aiming for the end of Q2 2023

karolkolodziej:
This feature is available in v11.2.0.

PR #19184, opened 2023-06-20
feature/14024-dumbbell-customization

magdalena-gut:
Added requested feature, #14024 .

FYI: Due to the implementation specifics of the Dumbbell series, a feature has been added to the AreaRange series from which it inherits.

---
TO DO:
- [x] - add lowMarker option
- [x] - add lowMarker logic
- [x] - update API
- [x] - update Docs
- [x] - tests for AreaRange
- [x] - additional tests for Dumbbell
- [x] - demo for AreaRange
- [x] - demo for Dumbbell
- [x] - fix update

TO FIX:
- [x] - when either the lowMarker or the marker is enabled while the other is disabled, it retains its symbol when hovered over
- [x] -  #19756 (reported)


docs/chart-and-series-types/range-series.md
```diff
@@ -20,6 +20,7 @@ Options related to Range series
 |Option|Description|
 |------|-----------|
 |[dataLabels](https://api.highcharts.com/highcharts/plotOptions.arearange.dataLabels)|Since both the low values and high values for a series need a separate data label, we added a new set of options, "xLow", "xHigh", "yLow" and "yHigh". With these options, the relative position can be altered.|
+|[lowMarker](https://api.highcharts.com/highcharts/plotOptions.arearange.lowMarker)|The new "lowMarker" option helps to visualize better the low and high values by customizing their symbols separately.
```
TorsteinHonsi, line 23: > The new "lowMarker" option...

Remove the word "new". This description will likely not be changed in a while.

samples/highcharts/demo/arearange-lowmarker/demo.css
TorsteinHonsi, line 1: Move it to highcharts/series-arearange/lowmarker

samples/highcharts/demo/arearange-lowmarker/demo.details
```diff
@@ -0,0 +1,9 @@
+---
+name: Temperature variation represented by lower and upper symbols
+authors:
+  - Magdalena Gut
+js_wrap: b
+tags:
+  - Highcharts demo 
+  - Area charts
```
TorsteinHonsi, line 8: Remove from demos

samples/highcharts/demo/dumbbell-lowmarker/demo.css
TorsteinHonsi, line 1: Move to `highcharts/series-dumbbell/lowmarker`

samples/highcharts/demo/dumbbell-lowmarker/demo.html
```diff
@@ -0,0 +1,14 @@
+<script src="https://code.highcharts.com/highcharts.js"></script>
+<script src="https://code.highcharts.com/modules/data.js"></script>
+<script src="https://code.highcharts.com/highcharts-more.js"></script>
```
TorsteinHonsi, line 3: The demo doesn't work, must add `modules/dumbbell.js`

ts/Core/Series/Series.ts
```diff
@@ -663,6 +663,29 @@ class Series {
         );
     }
 
+    /**
+     * Determine whether a marker in a series has changed.
```
TorsteinHonsi, line 667: ... the marker ...

ts/Series/AreaRange/AreaRangeSeries.ts
```diff
@@ -856,6 +889,30 @@ export default AreaRangeSeries;
  * @apioption series.arearange.fillOpacity
  */
 
+/**
+ * Options for the lower markers of the arearange-like series. When lowMarker is
```
TorsteinHonsi, line 893: > When lowMarker is....

Put `lowMarker` in backticks.

ts/Series/AreaRange/AreaRangeSeries.ts
```diff
@@ -856,6 +889,30 @@ export default AreaRangeSeries;
...
+ * @extends   plotOptions.series.marker
+ * @default   undefined
+ * @product   highcharts highstock
+ * @apioption plotOptions.arearange.lowMarker
+ */
+
+/**
+ *
+ * @sample {highcharts} highcharts/demo/arearange-lowmarker/
+ *         Area range chart with lowMarker option
```
TorsteinHonsi, line 908: Put `lowMarker` in backticks

ts/Series/Dumbbell/DumbbellSeries.ts
```diff
@@ -613,6 +619,28 @@ export default DumbbellSeries;
  * @product   highcharts highstock
  * @apioption series.dumbbell.data
  */
+/**
+ * Options for the lower markers of the dumbbell-like series. When lowMarker is
```
TorsteinHonsi, line 623: Backticks

TorsteinHonsi (changes requested):
Thanks, nice work! 

In addition to the comments, I think we should do some more work on the main dumbbell-lowmarker demo. We now have to tools to create something with directional arrows, like put forward in [this internal conversation](https://highsoft.slack.com/archives/C0GV77GCD/p1678370011472449).

We don't need to go all that way, but I think we should at least make the demo more useful by finding a data set that has both increasing and decreasing data. We can separate those into two series, with their specific markers. POC: https://jsfiddle.net/zfo0cnvp/

TorsteinHonsi (changes requested):
Thanks!

I still think we should do more with the dumbbell sample though. Please see my [previous comment](https://github.com/highcharts/highcharts/pull/19184#pullrequestreview-1582408533). The sample data set still has only rising values.

magdalena-gut:
> Thanks!
> 
> I still think we should do more with the dumbbell sample though. Please see my [previous comment](https://github.com/highcharts/highcharts/pull/19184#pullrequestreview-1582408533). The sample data set still has only rising values.

Certainly, the demo is almost ready, however I found a bug in the meantime - I work on it.

TorsteinHonsi:
@magdalena-gut Any progress on the final pieces?

magdalena-gut:
> 

I got stuck on this bug a bit, however, I hope tomorrow's PP session will help to sort it out. I just pushed the improved demo.

samples/highcharts/series-dumbbell/lowmarker/demo.details
TorsteinHonsi, line 1: I think we should add this sample to the demo pages, as it represents something that we don't have other demos for. So
* Move it to `highcharts/demos/dumbbell-markers`
* Pick a shorter name in demo.details. This will show up in the demo page.
* Set a category in demo.details (see existing demos for template).

samples/highcharts/series-dumbbell/lowmarker/demo.js
```diff
@@ -0,0 +1,204 @@
+// Define custom SVG symbols for left and right triangles
+Highcharts.SVGRenderer.prototype.symbols.triangleLeft = function (x, y, w, h) {
+    return ['M', x, y + h / 2, 'L', x + w, y, 'L', x + w, y + h, 'Z'];
+};
+
+Highcharts.SVGRenderer.prototype.symbols.triangleRight = function (x, y, w, h) {
+    return ['M', x + w, y + h / 2, 'L', x, y, 'L', x, y + h, 'Z'];
+};
```
TorsteinHonsi, line 8: Prefer arrow functions

samples/highcharts/series-dumbbell/lowmarker/demo.js
```diff
@@ -0,0 +1,204 @@
...
+    {
+        name: 'Decrease',
+        data: decreasingData,
+        marker: {
+            enabled: false
+        },
+        lowMarker: {
+            enabled: true,
+            symbol: 'triangleLeft',
+            fillColor: Highcharts.getOptions().colors[5]
```
TorsteinHonsi, line 200: Why is this necessary? By default I would expect the low-marker to get fill by the `point.color || series.color` just like the main marker.

samples/highcharts/series-dumbbell/lowmarker/demo.js
```diff
@@ -0,0 +1,204 @@
...
+    current: 3
+}].map(dataPoint => {
+    const isIncrease = dataPoint.previous < dataPoint.current;
+    return {
+        ...dataPoint,
+        low: isIncrease ? dataPoint.previous : dataPoint.current,
+        high: isIncrease ? dataPoint.current : dataPoint.previous,
+        color: isIncrease ?
+            Highcharts.getOptions().colors[2] :
+            Highcharts.getOptions().colors[5]
```
TorsteinHonsi, line 107: Since all the increasing points are assigned to the same series, and all the decreasing ones to another, I assume it would be better to set the color on the series rather than on each individual point.

samples/highcharts/series-dumbbell/lowmarker/demo.css
```diff
@@ -0,0 +1,5 @@
+#container {
+    height: 500px;
+    max-width: 600px;
+    margin: 0 auto;
```
TorsteinHonsi, line 4: Metter apply max-width and margin to the figure element, so that the description is aligned with the chart.

ts/Core/Series/Series.ts
```diff
@@ -663,6 +663,29 @@ class Series {
...
+     */
+    public hasMarkerChanged(
+        options: DeepPartial<SeriesOptions>,
+        oldOptions: DeepPartial<SeriesOptions>
+    ): boolean | undefined {
+        const series = this,
+            marker = options.marker,
+            oldMarker = oldOptions.marker || {};
+
+        // This is a workaround for update lowMarker symbol #14024
```
TorsteinHonsi, line 680: It's not really a "workaround", more a refactoring to allow extendibility. I think this comment can just be removed.

ts/Series/AreaRange/AreaRangeSeries.ts
```diff
@@ -560,14 +560,32 @@ class AreaRangeSeries extends AreaSeries {
...
 
+        if (series.options.lowMarker) {
+            const { options: { marker, lowMarker } } = series;
+
+            series.options.marker = merge(marker, lowMarker);
+
+            if (lowMarker.symbol) {
+                series.symbol = lowMarker.symbol;
+            }
+        }
```
TorsteinHonsi, line 578: It seams to me that this code block, and the restore block below, are repeated between the AreaRangePoint.ts and AreaRangeSeries.ts files. In adherence with the [DRY principle](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself), we should refactor it out to a function. The function should be placed on the AreaRangeSeries class since it handles stuff on the series level.

One way to resolve it is to do the two operations in the same function. So the first time it runs (without arguments), it returns the original settings for marker and symbol. The second time it is called with the original settings as an argument, and then those properties are restored.

TorsteinHonsi (changes requested):
Thanks! Some more comments. 

The dumbbell demo looks great now, I think we should add this to our demo pages.

samples/highcharts/series-dumbbell/lowmarker/demo.js
magdalena-gut: In dumbbell series there is also the lowColor option which is set by default and takes the priority over the series color. I didn't change it due to backward compatibility. Should I change it?

TorsteinHonsi: Okay, I didn't know that. Let's keep it.

However, I notice that if I set `lowColor` to `undefined`, the marker takes the color of the series. That is intuitive and makes sense to me. So I think we should do two things with that:
* Update the [API docs](https://api.highcharts.com/highcharts/series.dumbbell.lowColor) with that information.
* Change your demo to use `lowColor: undefined` instead of hard-setting `lowMarker.fillColor`.

TorsteinHonsi:
Now pending change requests:
* Use `lowColor: undefined` https://github.com/highcharts/highcharts/pull/19184#discussion_r1333965131
* Don't repeat code https://github.com/highcharts/highcharts/pull/19184#discussion_r1327059636

TorsteinHonsi (approved):
Thanks! 🎉

PR #19189, opened 2023-06-22
Tools: Dashboards dist preparations.

bre1470:
To keep PR size small, I will leave some gulp tasks out and put them in a new PR asap.

- `npx gulp dashboards/scripts` for now to produce `code/dashboards/**`
- `npx gulp dashboards/dist-minify` for compressed variant without `.js` instead of `.src.js`
- Adjustments to demos and tests
- Requires latest update of `highcharts-assembler`
- Requires latest update of `highcharts-utils`
- `npx gulp dashboards/scripts-esx` for some time in the future (`ts/` -> `code/` without `js/`)


goransle (approved):

docs/dashboards/installation.md
```diff
@@ -15,22 +15,22 @@ Then import the package in your project:
...
 ```
 ### Load additional modules
 There are some additional modules that you might want to include in your webpage that are not included in the basic `dashboards.js` module. The basic module lets you create the elementary dashboard with `HTMLComponents`, but if you would like to use some predefined component types, you need to add the necessary modules. E.g. to use a Highcharts component, you need to add the following scripts:
 ```html
-    ../dashboards.js
-    ../dashboards-plugin.js
-    ../highcharts.js
+    ../dashboards/dashboards.js
+    ../dashboards/modules/dashboards-plugin.js
+    ../highcharts/highcharts.js
```
pawelfus, line 33: should be just `highcharts.js`

pawellysy (approved):

karolkolodziej (approved):
Don't we need any `.md` file describing that?

pawelfus (approved):
One tiny change, otherwise good ✅ 
I have very limited knowledge of those tasks though.. but I couldn't break anything in the HC either, nice! 👍

PR #19193, opened 2023-06-22
bugfix/19175-map-selection-panning

### LINKED ISSUE #19175: Regression: Zooming on map with selections causes selections to not show as selected
#### Expected behaviour

Zooming on a map with selections will continue to highlight the selected areas on zoom. This is what was experienced with version 10.3.3.
working-10 3 3

Here I selected Minnesota. Next, I scrolled while on on Minnesota and South Dakota to show it working. The state stays selected, which you can see at the top of the screen capture.

#### Actual behaviour

When zooming or panning, the selected areas do not redraw correctly. When the mouse is over a selected area, that area is shown correctly. When the mouse is not, that area is drawn as though it is not selected. With with multiple selections active, only the one where the mouse is hovering will be drawn correctly.
fail-11 0 0

Same sort of scenario as above.

https://jsfiddle.net/g3oj507x/

#### Product version
Working in Highcharts 10.3.3, failing in 11.0.0, 11.0.1, 11.1.0.

#### Affected browser(s)
Chrome.

hubertkozik:
Fixed #19175, selected point was lost after map zomming or panning.

### DIFF
--- a/samples/unit-tests/maps/interaction/demo.js
+++ b/samples/unit-tests/maps/interaction/demo.js
@@ -4,8 +4,12 @@ QUnit.test('Hover color', function (assert) {
     Highcharts.Color.names = {};
 
     var chart = Highcharts.mapChart('container', {
+            mapNavigation: {
+                enabled: true
+            },
             series: [
                 {
+                    allowPointSelect: true,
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #19194, opened 2023-06-22
demo/column-basic-data-update

fanuntez:
Have updated to the given data set provided here: [Asana task](https://app.asana.com/0/1203542399740129/1203553273456339/f)


samples/highcharts/demo/column-basic/demo.js
```diff
@@ -3,67 +3,45 @@ Highcharts.chart('container', {
         type: 'column'
     },
     title: {
-        text: 'Monthly Average Rainfall'
+        text: 'Corn vs wheat estimated production for 2020 (1000 MT)'
```
TorsteinHonsi, line 6: It looks a bit crammed in mobile view, so I think we should
* Remove the "(1000 MT)" part. The unit is given in the y-axis title so we don't need it in the main title.
* Left-align the title and subtitle to make it appear more tidy.

samples/highcharts/demo/column-basic/demo.js
```diff
@@ -3,67 +3,45 @@ Highcharts.chart('container', {
...
     },
     tooltip: {
-        headerFormat: '<span style="font-size:10px">{point.key}</span><table>',
-        pointFormat: '<tr><td style="color:{series.color};padding:0">{series.name}: </td>' +
-            '<td style="padding:0"><b>{point.y:.1f} mm</b></td></tr>',
-        footerFormat: '</table>',
-        shared: true,
-        useHTML: true
+        valueSuffix: ' (1000 MT)',
+        stickOnContact: true,
```
TorsteinHonsi, line 28: Any particular reason for using `stickOnContact`? We usually enable that when there is clickable links in the tooltip, so in this case I don't think it is needed.

samples/highcharts/demo/column-basic/demo.js
```diff
@@ -3,67 +3,45 @@ Highcharts.chart('container', {
...
+        backgroundColor: 'rgba(255, 255, 255, 0.93)'
```
TorsteinHonsi, line 29: Better leave out the background color I think, otherwise this doesn't work well with themes (Dark Unica in particular).

TorsteinHonsi (changes requested):
Thanks! Some minor changes requested.

TorsteinHonsi (approved):

PR #19201, opened 2023-06-23
Getjson to fetch v2

andreastistel:
In demo.js:
- Added async wrapper to the demos that did not have it.
- Replaced getJSON with fetch.
In demo.html.
- Removed the script /modules/data.js


TorsteinHonsi (approved):
Thanks, looking good! Pending the CI check.

PR #19209, opened 2023-06-23
bugfix/18718-aria-expanded-state-sync

### LINKED ISSUE #18718: Accessibility: Visually hidden "View as data table" button's expanded/collapsed state gets out of sync with hamburger menu item
#### Expected behaviour
Element ID: "hc-linkto-highcharts-data-table-0"
aria-expanded="true" should change to "false" when the data table is closed by the hamburger menu's "hide data table" item

#### Actual behaviour
Element ID: "hc-linkto-highcharts-data-table-0"
aria-expanded="true" remains when the data table is closed by the hamburger menu's "hide data table" item

https://www.highcharts.com/demo/accessible-line

Step 1 (Works as expected):
Using a screen reader, activate the visually hidden button "View as data table", the data table is shown and the visually hidden button state reads "expanded, button", the item in the hamburger menu changes to "Hide data table"

Step 2 (Does not work as expected):
Using the hamburger menu's "Hide data table" the visually hidden button state still reads as "expanded, button"

#### Product version
10.3.3

#### Affected browser(s)
Voice Over in Safari
NVDA in Chrome
Likely others

marvin19:
Fixed #18718 , aria-expanded state for table was out of sync when viewing or hiding the data table.

### DIFF
--- a/ts/Accessibility/Components/InfoRegionsComponent.ts
+++ b/ts/Accessibility/Components/InfoRegionsComponent.ts
@@ -271,6 +271,13 @@ class InfoRegionsComponent extends AccessibilityComponent {
             }
         });
 
+        this.addEvent(chart, 'afterHideData', function (): void {
+            if (component.viewDataTableButton) {
+                component.viewDataTableButton
+                    .setAttribute('aria-expanded', 'false');
+            }
+        });
+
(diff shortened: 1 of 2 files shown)


oysteinmoseng (approved):
Thanks!

PR #19211, opened 2023-06-23
bugfix/19178-mouse-wheel-zoom-yaxis-options-gone

### LINKED ISSUE #19178: mouse-wheel-zoom: options.id desappear
I use new feature "modules/mouse-wheel-zoom" but I have a probleme, when I use the mouse wheel, the options.id in yAxis desappear.

#### Live demo :
https://jsfiddle.net/bxsyqwmh/9/

#### Product version
Highcharts v 11.1.0

jedrzejruta:
Fixed #19178, some `yAxis` options were dissapearing after mouse wheel zoom.

### DIFF
--- a/ts/Extensions/MouseWheelZoom/MouseWheelZoom.ts
+++ b/ts/Extensions/MouseWheelZoom/MouseWheelZoom.ts
@@ -149,11 +149,19 @@ const zoomBy = function (
             }
 
             if (startOnTick || endOnTick) {
-                yAxis.setOptions({ startOnTick: false, endOnTick: false });
+                yAxis.setOptions(
+                    // Merge with y-axis options so `yAxis.id` isn't
+                    // overwritten during wheel zoom, #19178
+                    merge(
+                        yAxis.options,
+                        { startOnTick: false, endOnTick: false }
+                    )
+                );
             }
             wheelTimer = setTimeout((): void => {
                 if (originalOptions) {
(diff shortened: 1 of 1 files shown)


test/cypress/integration/Highcharts/mouse-wheel-zoom/zoom-chart.cy.js
```diff
@@ -0,0 +1,41 @@
...
+            const yAxis = chart.yAxis[0],
+                { min, max } = yAxis.getExtremes();
+
+            cy.get('#container')
+                .trigger("wheel",
+                    {
+                        deltaY: -66.666666, wheelDelta: 120, wheelDeltaX: 0,
+                        wheelDeltaY: 120, bubbles: true
+                    }
+                ).then(() => {
```
hubertkozik, line 19: I think we should use a different code format, like e.g. here: https://github.com/highcharts/highcharts/blob/bcfc1ad10b53ab5d90bf64b76a8228713d66716e/ts/Core/Tooltip.ts#L1131-L1133
(brackets `{` positions)

hubertkozik (changes requested):
One minor comment, besides that fix looks good to me!

hubertkozik (approved):
LGTM!

TorsteinHonsi (approved):
Thanks!

PR #19230, opened 2023-06-27
bugfix/19179-annotation-outside-plot

### LINKED ISSUE #19179: Unexpected behavior of annotation
The expected behavior should be for this annotation to remain within the plot area

When we add text annotation and drag it outside the chart and then it has been removed/Hide.

(https://jsfiddle.net/BlackLabel/hmn3k12r/)

https://github.com/highcharts/highcharts/assets/135982085/a1a057b8-9d49-4656-a274-d41ff9449bc7

jakubSzuminski:
Fixed #19179, changing the crop property on annotations through `annotationOptions` in the chart config did not work.

---
As stated in the API (https://api.highcharts.com/highcharts/navigation.annotationsOptions.labelOptions.crop):

> crop: boolean, Whether to hide the annotation's label that is outside the plot area.

So maybe that's the desired behavior and we should not change it?

---
However, it was impossible to set `labels.crop` to `false` in `annotationOptions` (https://jsfiddle.net/BlackLabel/od3jxtkL/) and this fix solved it so now the users will be able to easily modify this behavior by simply setting `annotationOptions.labelOptions.crop = false`.

### DIFF
--- a/samples/highcharts/cypress/annotations-gui/demo.js
+++ b/samples/highcharts/cypress/annotations-gui/demo.js
@@ -6,6 +6,12 @@ Highcharts.chart('container', {
         text: 'Use the dropdown to add or edit annotations'
     },
     navigation: {
+        annotationsOptions: {
+            crop: false,
+            labelOptions: {
+                crop: false
+            }
+        },
         events: {
             selectButton: function (event) {
                 var newClassName = event.button.className + ' highcharts-active',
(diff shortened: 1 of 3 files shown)


pawelfus (approved):
> As stated in the API (https://api.highcharts.com/highcharts/navigation.annotationsOptions.labelOptions.crop):
> 
> > crop: boolean, Whether to hide the annotation's label that is outside the plot area.
> 
> So maybe that's the desired behavior and we should not change it?

That's correct, it is an option to set behaviour like this. Please keep it.

I agree with the fix: let devs configure it. Could you please change the description of the fix, so it's clear in the changelog what we have fixed?

TorsteinHonsi (approved):
Thanks!

PR #19232, opened 2023-06-27
bugfix/19173-treegraph-custom-marker

### LINKED ISSUE #19173: Treegraph custom marker symbol error
#### Expected behaviour
Treegraph custom _url_ marker symbol should be properly rendered.

#### Actual behaviour
Treegraph custom _url_ marker symbol ends with a console error, the custom symbol is not rendered.

https://jsfiddle.net/BlackLabel/jtv17zxu/

#### Product version
v11.1.0

Vazonik:
Fixed #19173, could not set a custom image as marker symbol in treegraph series.

---
TO DO:
- [x] Add rendering images when the image URL is set in the `symbol` option.
- [x] Add regression tests.
- [x] Update & fix docs (missing `marker` property in `levels`, `levels` as an object instead of an array).

### DIFF
--- a/samples/unit-tests/series-treegraph/treegraph/demo.js
+++ b/samples/unit-tests/series-treegraph/treegraph/demo.js
@@ -12,7 +12,13 @@ QUnit.test('Treegraph series',
                     }],
                     dataLabels: {
                         pointFormat: '{point.id}'
-                    }
+                    },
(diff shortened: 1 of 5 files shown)


TorsteinHonsi (approved):
Thanks!

PR #19234, opened 2023-06-28
db/DataPoolConnectorOptions

karolkolodziej:
Added documentation for `DataPoolConnectorOptions`.


ts/Data/DataPoolOptions.ts
```diff
@@ -33,8 +33,22 @@ export interface DataPoolOptions {
 
 export interface DataPoolConnectorOptions
 <T extends keyof DataConnectorTypes = keyof DataConnectorTypes> {
+    /**
+     * The unique identifier of the connector. Used later when referencing
+     * the connector in the component where it is used.
+     **/
     id: string;
+    /**
+     * The options of the given connector type.
```
sebastianbochan, line 42: Would be good to add an example of using it.

ts/Data/DataPoolOptions.ts
```diff
@@ -33,8 +33,22 @@ export interface DataPoolOptions {
...
+     **/
     options: DataConnectorTypes[T]['prototype']['options'];
+    /**
+     * The type of the connector, depends on your data source.
+     * Possible values are:
+     * - `CSV`
+     * - `GoogleSheets`
+     * - `HTMLTable`
```
pawellysy, line 50: Isn't the bundler show all available types?

pawellysy (approved):
LGTM! One minor comment tho

ts/Data/DataPoolOptions.ts
karolkolodziej: No, there is no registry for that.

bre1470: `DataConnector.types` is it.

ts/Data/DataPoolOptions.ts
```diff
@@ -33,8 +33,34 @@ export interface DataPoolOptions {
...
+     *       }]
+     * },
+     **/
     options: DataConnectorTypes[T]['prototype']['options'];
+    /**
+     * The type of the connector, depends on your data source.
+     * Possible values are:
+     * - `CSV`
+     * - `GoogleSheets`
+     * - `HTMLTable`
```
bre1470, line 62: Please replace with something like
> The available types can be found in the `DataConnector.types` registry.

bre1470 (changes requested):

sebastianbochan (approved):

pawelfus (changes requested):
Asana: "... and Connector options" 😬 

Zrzut ekranu 2023-06-28 o 17 00 50

Question: Can we define it? So we have access to `GoogleSheetsConnector.Options`, `HTMLTableConnector.Options` and `CSVConnector.Options`?

bre1470:
We can link them in the doclet of the class property.

pawelfus (approved):
Looks good (except the part with `DataPoolConnectorOptions.options`)

ts/Data/DataPoolOptions.ts
pawelfus: +1

ts/Data/DataPoolOptions.ts
```diff
@@ -28,13 +31,46 @@ import type { DataConnectorTypes } from './Connectors/DataConnectorType.js';
  * */
 
 export interface DataPoolOptions {
+    /**
+     * Link list to connector options + mention DataConnector.types
+     *
```
bre1470, line 36: This was text for your eyes only and should be replaced. :)

ts/Data/DataPoolOptions.ts
```diff
@@ -28,13 +31,46 @@ import type { DataConnectorTypes } from './Connectors/DataConnectorType.js';
...
+     * Available connectors and its options:
+     * {@link CSVConnectorOptions | CSVConnector}
+     * {@link GoogleSheetsConnectorOptions | GoogleSheetsConnector}
+     * {@link HTMLTableConnectorOptions | HTMLTableConnector}
```
bre1470, line 40: This should be formatted as a list.

ts/Data/Connectors/DataConnector.ts
```diff
@@ -367,8 +367,17 @@ namespace DataConnector {
...
+         * Option of the modifier that is used to modify the data.
+         */
         dataModifier?: DataModifierTypeOptions;
+        /**
+         * Options for the data table.
+         */
         dataTable?: DataTable.Options;
+        /**
+         * Your custom metadata.
+         */
```
bre1470, line 380: This one should also become a separate file.

bre1470 (changes requested):

ts/Data/DataPoolOptions.ts
karolkolodziej: 🙈

ts/Data/DataPoolOptions.ts
bre1470: DataConnector registry is not ready for API docs. So we leave it for now.

ts/Data/Connectors/DataConnectorOptions.d.ts
```diff
@@ -0,0 +1,70 @@
...
+    /**
+     * Option of the modifier that is used to modify the data.
+     */
+    dataModifier?: DataModifierTypeOptions;
+    /**
+     * Options for the data table.
+     */
+    dataTable?: DataTable.Options;
+    /**
+     * Your custom metadata.
```
bre1470, line 42: This needs more explanation. Something like
> Contains custom metadata related to the connector source, for example the column order on the source side.

ts/Data/Connectors/DataConnectorOptions.d.ts
```diff
@@ -0,0 +1,70 @@
...
+     */
+    metadata?: Metadata;
+}
+
+/**
+ * Metadata
```
bre1470, line 48: Same here.

ts/Data/Connectors/DataConnectorOptions.d.ts
```diff
@@ -0,0 +1,70 @@
...
+ */
+export interface Metadata {
+    columns: Record<string, MetaColumn>;
+}
+
+/**
+ * Metadata entry containing the name of the column and a metadata object.
```
bre1470, line 55: This is misplaced and should go to `Metadata.columns`. Here maybe something like:
> Metadata for a specific column in the connector source.

bre1470 (changes requested):

bre1470 (approved):
Thank you! Looks good. 👍

goransle (approved):

PR #19247, opened 2023-06-29
dash/json-serialization-demo

sebastianbochan:


karolkolodziej (approved):
I would add a `.md` file where you explain how it works.
The same as we have in `dashboards/demo/light-dark-theme`

pawellysy (approved):
LGTM!

PR #19249, opened 2023-06-29
Getjson to fetch v8

andreastistel:
In demo.js:
* Added async wrapper to the demos that did not use it.
*  Replaced getJSON with fetch.
In demo.html.
*  Removed the script /modules/data.js


samples/stock/demo/compare/demo.js
```diff
@@ -1,73 +1,76 @@
...
+        .then(data => success('msft', data));
+
+    await fetch('https://cdn.jsdelivr.net/gh/highcharts/highcharts@v7.0.0/samples/data/aapl-c.json')
+        .then(response => response.json())
+        .then(data => success('aapl', data));
+
+    await fetch('https://cdn.jsdelivr.net/gh/highcharts/highcharts@v7.0.0/samples/data/goog-c.json')
+        .then(response => response.json())
+        .then(data => success('goog', data));
+
```
TorsteinHonsi: The idea of this code block is to successively load the three JSON files, then create a series for each of them. I think with Promises it would be better to rewrite this. Proof of concept using `Promise.all`:

```ts
(async () => {
    const promises = ['msft', 'aapl', 'goog'].map(name => new Promise(async (resolve, reject) => {
        const data = await fetch(`https://cdn.jsdelivr.net/gh/highcharts/highcharts@v7.0.0/samples/data/${name}-c.json`)
            .then(response => response.json());
        resolve({
            name,
            data
        });
    }));

    const series = await Promise.all(promises);

    console.log(series);
})();
```

On jsFiddle: https://jsfiddle.net/highcharts/0m6ebufs/

There was also at least one other place this pattern was used, that should be updated too.

TorsteinHonsi (changes requested):
Thanks! One code pattern should be improved now that we rely on Promise support.

TorsteinHonsi (approved):

PR #19259, opened 2023-06-30
dash/exclude-datagrid-from-plugin-build

pawellysy:
Deleted DataGrid from dashboards-plugin build.

### DIFF
--- a/ts/masters-dashboards/modules/dashboards-plugin.src.ts
+++ b/ts/masters-dashboards/modules/dashboards-plugin.src.ts
@@ -19,11 +19,14 @@ const G: AnyRecord = Dashboards;
 G.DataGridPlugin = DataGridPlugin;
 G.HighchartsPlugin = HighchartsPlugin;
 
-if (G.win.Dashboards && G.win.Highcharts) {
+if (G.win.Dashboards) {
+    if (G.win.Highcharts) {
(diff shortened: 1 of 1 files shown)


ts/masters-dashboards/modules/dashboards-plugin.src.ts
```diff
@@ -25,6 +24,6 @@ if (G.win.Dashboards && G.win.Highcharts) {
     HighchartsPlugin.custom.connectHighcharts(G.win.Highcharts);
     G.win.Dashboards.PluginHandler.addPlugin(HighchartsPlugin);
 
-    DataGridPlugin.custom.connectDataGrid(DataGrid);
+    DataGridPlugin.custom.connectDataGrid(G.win.DataGrid.DataGrid);
```
karolkolodziej, line 27: What if `G.win.DataGrid` is undefined :)

karolkolodziej (changes requested):

bre1470 (approved):

bre1470:
Not directly related, but could you fix the path to the plugin module in `./tools/gulptasks/pr-size-table.js` with this PR? Then we might have a chance that the visual comparison job succeeds again.

karolkolodziej (approved):

sebastianbochan (approved):

PR #19260, opened 2023-06-30
Use prefered OSM tile URL

otbutz:
see https://github.com/openstreetmap/operations/issues/737


TorsteinHonsi:
@hubertkozik Do you agree to this change?

hubertkozik (approved):
Seems, that we have to update that URLs to new standard :) After small modifications, everything seems good to me.

TorsteinHonsi (approved):

PR #19262, opened 2023-06-30
bugfix/16784-wrong-extended-positions-calculations

### LINKED ISSUE #16784: Annotation line changes its angle when dragging.
#### Expected behaviour
Annotation angle should remain the same when dragging.

#### Actual behaviour
Annotation angle changes when dragging.

https://jsfiddle.net/BlackLabel/o5ewf4cq/1/

1. Drag the annotation.

#### Product version
Highcharts Stock 9.3.2
#### Affected browser(s)
N/A

#### Workaround
Set `xAxis.ordinal` to `false`.

kamil-musialowski:
Fixed #16784, calculations in extended ordinal position were wrong.

Test here: https://jsfiddle.net/BlackLabel/w4b2qjxg/
Grab the annotation, move it to the left (outside of the chart), try to bring it back. It shouldn't change it's length.

TODO: 

- [x] Test
- [x] Dragging annotation outside the dataset (i.e far left)


ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -808,13 +801,10 @@ namespace OrdinalAxis {
...
-            if (!extendedOrdinalPositions) {
-                extendedOrdinalPositions =
-                    ordinal.getExtendedPositions &&
-                    ordinal.getExtendedPositions();
-                ordinal.extendedOrdinalPositions = extendedOrdinalPositions;
-            }
+            extendedOrdinalPositions =
+                ordinal.getExtendedPositions &&
+                ordinal.getExtendedPositions();
+            ordinal.extendedOrdinalPositions = extendedOrdinalPositions;
```
pawelfus, line 807: Unnecessary line + I think you can remove from definitions `Ordinal.extendedOrdinalPositions` too

pawelfus (changes requested):

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -1296,6 +1287,8 @@ namespace OrdinalAxis {
                     fakeAxis.series.push(fakeSeries);
 
                     series.processData.apply(fakeSeries);
+
+                    fakeSeries.applyGrouping.apply(fakeSeries);
```
pawelfus, line 1291: Waaait. The idea with `Axis.applyGrouping` (the method that calls `Series.applyGrouping`) is to use this method **after** each series has been processed (after `processData()`). Could you try it?

samples/unit-tests/axis/ordinal/demo.js
```diff
@@ -804,6 +815,19 @@ QUnit.test('Moving annotations on ordinal axis, #18459', assert => {
...
         0.1,
-        'Annotation dragged on ordinal axis charts should follow mouse pointer.'
+        'Annotation dragged on ordinal axis charts should follow mouse pointer, #18459.'
+    );
+
+    chart.xAxis[0].setExtremes(1622813400000, 1623245400000);
+
+    const val = chart.xAxis[0].toValue(-150, true);
+    const pixels = chart.xAxis[0].toPixels(val, true);
+    console.log(pixels);
```
pawelfus, line 825: Unnecessary console

samples/unit-tests/axis/ordinal/demo.js
```diff
@@ -767,6 +775,9 @@ QUnit.test('Moving annotations on ordinal axis, #18459', assert => {
 
     const chart = Highcharts.stockChart('container', {
         series: [{
+            dataGrouping: {
+                forced: true
+            },
```
pawelfus, line 780: Please, keep the original test and update series for your scenario. I don't know if the `Annotation dragged on ordinal axis charts should follow mouse pointer` assertion will be fine if we force grouping.

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -851,6 +825,11 @@ namespace OrdinalAxis {
                 ordinalIndex = getIndexInArray(extendedOrdinalPositions, val) -
                     originalPositionsReference;
             } else {
+                if (!toIndex) {
+                    // If the index is outside positions array,
```
pawelfus, line 829: Index? 🤔

pawelfus (changes requested):
Minor polishing needed 👍

pawelfus (approved):
Thanks!

TorsteinHonsi (changes requested):
Can you check out the CI error complaining about `this.is()`?

kamil-musialowski:
@TorsteinHonsi 
This error comes from the point and figure JS demo (which I have created), and it only checks the series type. Local tests show nothing and honestly, I have no idea what could be causing this to fail on this particular branch (it's the first time I see an error regarding this demo).

https://github.com/highcharts/highcharts/blob/9360ca8eef784c8b9cd4238b0e0c5f302be5578e/samples/highcharts/studies/pointandfigure/demo.js#L181

Should we ignore it, or add some checks like this.is && this.is('...')?

TorsteinHonsi:
@kamil-musialowski I can reproduce it in the utils:

Skjermbilde 2023-09-12 kl  10 07 51

Rather than adding a test for `this.is` in the study, at least first we should try to understand why it happens. From the stack trace we can see that it origins from the `getExtendedPositions` function, which indicates that the connection to this PR is real.

kamil-musialowski:
@TorsteinHonsi Thank you for the help, I can see it in the utils now.

The error comes from the fact, that the overwritten `groupData` method is also called on `fakeSeries`, which does not have a `fakeSeries.is()` method. I'm not sure if it's something that should be handled inside Core. Do we need this `is()` method on fakeSeries?

If not, I think we should be safe with checking if series is fake i.e `if (series.yData)` or with the `this.is && this.is()` approach.

TorsteinHonsi:
Okay, I understand. It's probably okay just to change it in the pointandfigure study then.

TorsteinHonsi (approved):

PR #19275, opened 2023-07-03
Dash/dist-examples

goransle:
Adds new gulp task can be run with`npx gulp dashboards/dist-examples`. Requires [demo-manager](https://github.com/highcharts/highcharts-demo-manager).

Note that I did not have too much luck running the samples from the filesystem, which is why the output links to highcharts.com/samples.


tools/gulptasks/dist-examples.js
```diff
@@ -204,12 +205,16 @@ async function createExamples(title, sourcePath, targetPath, template) {
...
+        sourcePath
+            .replaceAll('samples/', '')
+            .replaceAll('/demo', '')
+    );
 
     LogLib.success('Created', targetPath);
     const indexContent = localsidebar
-        .replace(/style=\"display:none;\"/g, '') // remove hidden style
-        .replace(/(?!href= ")(\.\/.+?)(?=")/g, 'examples\/$1\/index.html'); // replace links
+        .replaceAll('style="display:none;"', '') // remove hidden style
```
sebastianbochan, line 216: Is it not saver to use regexp and set style as optional?

goransle: Should be equivalent in this case. For context, this is the kind of file being replaced: https://www.highcharts.com/demo/sidebar.html

TorsteinHonsi (approved):
Thanks!

sebastianbochan (approved):

bre1470 (commented):
I still not manage to test it. What target folder do I have to use? Can you maybe add a CLI help for `dist-examples`?

bre1470:
@goransle helped with testing. Looks good. 🎉

bre1470 (approved):

PR #19278, opened 2023-07-03
bugfix/19200-annotation-label-html-style

### LINKED ISSUE #19200: Styles won't be applied to annotation label if useHTML is enabled
#### Expected behaviour
With useHTML enabled for annotations, you should be able to apply styles to the annotation text. E.g., change the font size.

#### Actual behaviour
When useHTML is enabled, style properties are ignored. If it helps, this issue didn't exist in version 11.0.0

[Demo](https://jsfiddle.net/night50/nmxqszep/17/)

1. View the demo and note the size of the font is set to 28px and useHTML is false. 
2. Change useHTML to true and run.
3. Note the font is no longer rendered at 28px

#### Product version
Highcharts version 11.1.0 (the bug was not present in version 11.0.0)

#### Affected browser(s)
MS Edge, Chrome, and FireFox

Vazonik:
Fixed #19200, styles were not applied to the annotation label when `useHTML` was enabled.

### DIFF
--- a/ts/Core/Renderer/HTML/HTMLRenderer.ts
+++ b/ts/Core/Renderer/HTML/HTMLRenderer.ts
@@ -314,7 +314,7 @@ class HTMLRenderer extends SVGRenderer {
                             // updating the shadow div counterpart with the same
                             // style.
                             css: function (styles: CSSObject): HTMLElement {
-                                wrapper.css.call(parentGroup, styles);
+                                wrapper.css(styles);
                                 (
                                     [
                                         // #6794
--- a/ts/Core/Renderer/HTML/HTMLRenderer.ts
+++ b/ts/Core/Renderer/HTML/HTMLRenderer.ts
@@ -240,7 +240,8 @@ class HTMLRenderer extends SVGRenderer {
                     // Ensure dynamically updating position when any parent
                     // is translated
                     parents.reverse().forEach(function (parentGroup): void {
-                        const cls = attr(parentGroup.element, 'class');
+                        const cls = attr(parentGroup.element, 'class'),
+                            parentProtoCss = parentGroup.css;
 
                         /**
                          * Common translate setter for X and Y on the HTML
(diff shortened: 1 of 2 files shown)


ts/Core/Renderer/HTML/HTMLRenderer.ts
```diff
@@ -314,7 +314,7 @@ class HTMLRenderer extends SVGRenderer {
                             // updating the shadow div counterpart with the same
                             // style.
                             css: function (styles: CSSObject): HTMLElement {
-                                wrapper.css.call(parentGroup, styles);
+                                wrapper.css(styles);
```
TorsteinHonsi, line 317: This is risky - currently it applies the change to the group's div counterpart, but now it is changed to apply to the actual wrapper. I can't find a fail case for it now, and the visual tests won't pick it up because they don't include the HTML content. But I don't think we can do it this way, we must assume there's a good reason why it currently applies the style to `parentGroup`.

TorsteinHonsi (changes requested):

ts/Core/Renderer/HTML/HTMLRenderer.ts
TorsteinHonsi: I see now that that code was new with https://github.com/highcharts/highcharts/commit/809fbbf37e321a63c73ae86e136774d056499400 . Did you bisect this regression to that commit? I'll have to do some more testing before approving.

Vazonik: @bm64 bisected it [here](https://github.com/highcharts/highcharts/issues/19200#issuecomment-1608515450). This bug appeared exactly when adding the `css` method in this commit. It seemed to substitute the `css` method call for the wrapper and it was probably missing, so I called it that way.

TorsteinHonsi: Thanks, I found out what's wrong. It doesn't take into account that there's a difference between SVGLabel.css and SVGElement.css. I'll upload a fix soon.

TorsteinHonsi (approved):

PR #19281, opened 2023-07-03
bugfix/14108-exporting-xrange-csv-data

### LINKED ISSUE #14108: Exporting x-range data to CSV
#### Expected behaviour
Default options for exporting x-range data to CSV.

#### Actual behaviour
It looks like it only includes start dates, which isn't very helpful.
You have to customize and build your own table in the getDataRows function each time.

https://jsfiddle.net/BlackLabel/ytL517mo/

#### Product version
v8.2.0

madepiet:
Fixed #14108, start and end dates in exported CSV data for X-range series.


TorsteinHonsi (changes requested):
Thanks!

I think rather than concatenating it into the x column, we should treat x2 as a separate column. Because it is actually a separate data value, which can be handled in spreadsheets etc. Probably we need to prepend it to the local [pointArrayMap](https://github.com/highcharts/highcharts/blob/dfbb2302dea0a320329eef630e9f2ca3c8c2d93a/ts/Extensions/ExportData/ExportData.ts#L436-L438), then make sure it is formatted like a date when printed.

Also make sure this works with the gantt series. Gantt points use `start`, `end` instead of `x`, `x2`.

madepiet:
> Thanks!
> 
> I think rather than concatenating it into the x column, we should treat x2 as a separate column. Because it is actually a separate data value, which can be handled in spreadsheets etc. Probably we need to prepend it to the local [pointArrayMap](https://github.com/highcharts/highcharts/blob/dfbb2302dea0a320329eef630e9f2ca3c8c2d93a/ts/Extensions/ExportData/ExportData.ts#L436-L438), then make sure it is formatted like a date when printed.
> 
> Also make sure this works with the gantt series. Gantt points use `start`, `end` instead of `x`, `x2`.

For X-range we currently have the following cases:
- [With xAxis in datatime format](https://jsfiddle.net/BlackLabel/uktws5e0/)
- [And standard](https://jsfiddle.net/BlackLabel/qvfgm2n9/)

However, for Gantt we only have [datatime](https://jsfiddle.net/BlackLabel/jho1tqdc/) because for other data it takes the [date format anyway](https://jsfiddle.net/BlackLabel/kz7n1328/).

TorsteinHonsi:
I pushed a commit that makes more use of existing structures in the export-data module, so that we don't have to make special cases for the XRange series. Let's see how the CI turns out.

TorsteinHonsi (approved):

PR #19285, opened 2023-07-04
Tools: dashboards visual test fix

goransle:
Adds an extra step which installs highcharts-utils. This must have been cached on the dev branch.


TorsteinHonsi (approved):
Thanks!

PR #19292, opened 2023-07-04
Samples/highcharts/demo - Updated keywords

andreastistel:
Updated keyword var to const/let in all the appropriate places in samples/highcharts/demo. 
For example see 'samples/highcharts/demo/combo-timeline/demo.js' line 10 to 21.


TorsteinHonsi (approved):
Thanks!

PR #19293, opened 2023-07-04
Samples/maps/demo - Updated keywords

andreastistel:
Updated keyword var to const/let in all the appropriate places in samples/maps/demo. 

This was done mostly with ESlint fix.

Also fixed a typo in 'samples/maps/demo/data-class-two-ranges/readme.md'.


TorsteinHonsi (approved):
Thanks!

PR #19295, opened 2023-07-04
Samples/stock/demo - Updated keywords

andreastistel:
Updated keyword var to const/let in all the appropriate places in samples/stock/demo. 
Also fixed the iterator in some for-loops where the dummy-variable were declared ahead of the loop.


TorsteinHonsi (approved):
Thanks!

PR #19296, opened 2023-07-04
Samples/gantt/demo - Updated keywords

andreastistel:
Updated keyword var to const/let in all the appropriate places in samples/gantt/demo.


TorsteinHonsi (approved):
Thanks!

PR #19298, opened 2023-07-04
Samples/column rotated labels responsive

fanuntez:
Made labels more responsive.

[Asana task](https://app.asana.com/0/1203542399740129/1203592881281866/f)


TorsteinHonsi (changes requested):
Thanks!

I think we can simplify this a bit:
* We don't really need to keep abbreviated names when the rotation is -90, because there is vertical space for the full name. As a consequence, we can go back to the original, simpler data structure.
* We can simply set `autoRotation: [-45, -90]` to achieve responsive handling of the label rotation. See https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/demo/column-rotated-labels .

fanuntez:
> Thanks!
> 
> I think we can simplify this a bit:
> 
> * We don't really need to keep abbreviated names when the rotation is -90, because there is vertical space for the full name. As a consequence, we can go back to the original, simpler data structure.
> * We can simply set `autoRotation: [-45, -90]` to achieve responsive handling of the label rotation. See https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/demo/column-rotated-labels .

Should the dataLabels changes also be reverted back as on the original sample?

TorsteinHonsi (approved):

PR #19306, opened 2023-07-05
Dash/demos improvements v3

pawellysy:
Improved responsive options for main dashboards demos for small screens.


TorsteinHonsi (changes requested):
Thanks! The main cell layout looks good now, but there are some other issues to address:

* Climate demo: temperature graph markers are hidden on mobile. Looks like the `enabledThreshold` should be lower, or removed. Actually setting `marker.enabled` to true should make them always visible.  But I can't make it work for some reason.
* Climate demo: All the kpi widgets have ugly overlapping text. The text size should be reduced responsively to avoid line breaking where it is not wanted.
* Light-dark-theme: Should also be 100% width, the chart is so narrow now that not all categories are visible. The height can be reduced though.
* Demo/minimal: the main title is clipped on mobile. If possible, it would be better with `height: auto`  on this cell (in all screen sizes). IMO it would also be better if this cell had no background and border, because it is not a "cell" per se, it is only the dashboard title.

pawellysy:
@TorsteinHonsi I'll take a `light-dark-theme` demo, and `minimal` 👍

pawelfus (changes requested):
Great work!

Just a couple minor things:

- minimal-demo has strange border/shadow in the title
Zrzut ekranu 2023-07-6 o 13 26 09
- could we also change `allowConnectorUpdate: true` etc titles? Perhaps replace title with subtitle, and remove the subtitle? (datagrid-sync demo)
- datacursor-sync: could we duplicate behaviour, to be the same as in sync-extremes? I mean narrow container will change width for the components to 100%

samples/dashboards/demo/climate/demo.css
```diff
@@ -184,6 +189,15 @@
     cursor: pointer;
 }
 
+#kpi-layout .highcharts-dashboards-component-title {
+    font-size: 1em !important;
```
karolkolodziej, line 193: If @TorsteinHonsi added here `!important` we can assume that other users will do it in a similar fashion...
IMO we should remove that dynamic `fontSize` for KPIComponent

samples/dashboards/demo/climate/test-notes.md
```diff
@@ -1 +1 @@
-Click on the world map
+Click on a city marker in the world map
```
karolkolodziej, line 1: Also, click on the KPI?

karolkolodziej (approved):

samples/dashboards/demo/climate/test-notes.md
bre1470: Good idea! And switching the temperature scale.

samples/dashboards/demo/climate/demo.css
TorsteinHonsi: Probably. My first thought was "why is this font size hard coded"? At least, an option to avoid dynamic font size would be appropriate.

TorsteinHonsi:
> minimal-demo has strange border/shadow in the title

@pawelfus @karolkolodziej Why not remove the first cell altogether and use a HTML h1 or h2 header instead?

samples/dashboards/demo/climate/demo.css
sebastianbochan: Being honest, I'm not convinced to use `!important`, I suggest to use more specific selector or inline style.

sebastianbochan, line 193: To be honest, I'm not convinced to use `!important`, I suggest using a more specific selector or inline style.

sebastianbochan (changes requested):

samples/dashboards/demo/climate/demo.css
pawellysy: @karolkolodziej @TorsteinHonsi, for now this is a part of discussion, so I'll just delete the `!important` flag, and we will come back to this topic in the future.

pawelfus (changes requested):
> could we also change allowConnectorUpdate: true etc titles? Perhaps replace title with subtitle, and remove the subtitle? (datagrid-sync demo)

Sorry for being unclear, I mean these - just keep the title with description (no need for "`allowupdate ... `" part):

Zrzut ekranu 2023-07-10 o 13 46 02

pawelfus (approved):
Amazing, thanks!

sebastianbochan (approved):
Please fix conflicts

bre1470 (approved):
Some linting issues after `npm run dtest`, but not related.

TorsteinHonsi (approved):
Thanks, now merged into https://github.com/highcharts/highcharts/pull/19391 where we continue the review.

PR #19307, opened 2023-07-05
dash/getOptions

sebastianbochan:
- [x] - add getOptions on each level
- [x] - fixed demo
- [x] - doclets
- [x] - update tests


ts/Dashboards/Board.ts
```diff
@@ -623,6 +623,39 @@ class Board implements Serializable<Board, Board.JSON> {
         };
     }
 
+    /**
+     * Get the board's options.
+     * @return {DeepPartial<Board.Options>}
```
karolkolodziej, line 628: No need to put type here, typedoc will pick it up from the method declaration.

ts/Dashboards/Board.ts
```diff
@@ -623,6 +623,39 @@ class Board implements Serializable<Board, Board.JSON> {
         };
     }
 
+    /**
+     * Get the board's options.
```
karolkolodziej, line 627: That is a public method. IMO needs better description what it returns, what are the limitations and what you can do with it ;)

ts/Dashboards/Components/HTMLComponent.ts
```diff
@@ -383,6 +384,21 @@ class HTMLComponent extends Component {
 
         return json;
     }
+
+    /**
+     * Get the HTML component's options.
+     * @return {Partial<HTMLComponent.ComponentOptions>}
```
karolkolodziej, line 390: Same here.

ts/Dashboards/Components/KPIComponent.ts
```diff
@@ -616,6 +617,21 @@ class KPIComponent extends Component {
 
         return json;
     }
+
+    /**
+     * Get the KPI component's options.
+     * @return {Partial<KPI.ComponentOptions>}
```
karolkolodziej, line 623: Same here

ts/Dashboards/Layout/Cell.ts
```diff
@@ -347,6 +347,17 @@ class Cell extends GUIElement {
         };
     }
 
+    /**
+     * Get the cell's options.
+     * @return {DeepPartial<Cell.Options>}
```
karolkolodziej, line 352: Same

ts/Dashboards/Layout/Layout.ts
```diff
@@ -418,6 +418,28 @@ class Layout extends GUIElement {
             }
         };
     }
+
+    /**
+     * Get the layout's options.
+     * @return {DeepPartial<Cell.Options>}
```
karolkolodziej, line 424: Same everywhere 😅

karolkolodziej (changes requested):

karolkolodziej:
Also TS declaration task fails

karolkolodziej (approved):

ts/Dashboards/Components/Component.ts
```diff
@@ -1066,6 +1067,18 @@ abstract class Component {
...
+     * Get the component's options.
+     * @returns
+     * The JSON of component's options.
+     *
+     * @internal
+     *
+     */
+    public getOptions(): Partial<Component.ComponentOptions> {
+        return diffObjects(this.options, Component.defaultOptions);
+    }
```
pawellysy, line 1080: Since all of the inheriting classes are implementing their own version of this method, I'd use the abstract method here, and force those classes to implement their version.

```suggestion
    /**
     * Get the component's options.
     * @returns
     * The JSON of component's options.
     *
     * @internal
     *
     */
    public abstract getOptions(): Partial<Component.ComponentOptions>;
```

ts/Dashboards/Layout/Row.ts
```diff
@@ -342,6 +342,27 @@ class Row extends GUIElement {
...
+        const row = this,
+            cells = [];
+
+        for (let i = 0, iEnd = row.cells.length; i < iEnd; ++i) {
+            cells.push(row.cells[i].getOptions());
+        }
+
+        return {
+            cells
+        };
```
pawellysy, line 363: Maybe let's serialise some other options of row, like `id` or `style` (it is still in the API)?

ts/Dashboards/Layout/Layout.ts
```diff
@@ -418,6 +418,28 @@ class Layout extends GUIElement {
...
+            rows = [];
+
+        // Get rows JSON.
+        for (let i = 0, iEnd = layout.rows.length; i < iEnd; ++i) {
+            rows.push(layout.rows[i].getOptions());
+        }
+
+        return {
+            rows
+        };
```
pawellysy, line 441: maybe let's serialise some options as well, like `id`, `rowClassName`, `cellClassName`, `layoutClassName` (they are in the API and in the interface)?

pawellysy (changes requested):

ts/Dashboards/Components/Component.ts
sebastianbochan: It would require refactoring KPI parts and ThresholdComponent, so I don't think that is valuable at this moment.

pawellysy: Your call

pawellysy (approved):

bre1470 (approved):

PR #19310, opened 2023-07-05
Improved Dashboards DTS

bre1470:
Improved TypeScript declarations to support classic bundles minified and as source.


ts/Dashboards/Plugins/HighchartsTypes.ts
```diff
@@ -0,0 +1,251 @@
...
+ *
+ *  Declarations
+ *
+ * */
+
+
+export interface AnimationOptions {
+}
+
+export interface Axis {
```
karolkolodziej, line 38: Can't we get the reference to these types instead of writing them once again?

karolkolodziej (approved):

ts/Dashboards/Plugins/HighchartsTypes.ts
bre1470: That is essential to keep the Highcharts Core out. We need the API definition for Plugin implementation, but not the types in the distribution, except Options. It would be easier to keep it out, if we had our own repository, because then we could reference highcharts as a package.

bre1470: Now thinking about we could simulate it with the tsconfig. It is just a question how this then maps to the classic bundle.

sebastianbochan (approved):

pawellysy (approved):
LGTM!

ts/Dashboards/Plugins/HighchartsTypes.ts
bre1470: No luck with tsconfig. 😞 We might change the repository structure and the we can try ones more.

karolkolodziej: let's leave it as it is now ;) thanks!

PR #19312, opened 2023-07-05
bugfix/19233-moving-annotations-with-null-points

### LINKED ISSUE #19233: Insertion and drag of stock annotation failure when there are null points.
#### Expected behaviour
The annotation is inserted exactly at the mouse click, and the annotation does not "escape" when dragging.

#### Actual behaviour
The annotation is added slightly next to the place of clicking the mouse on the chart, the sliding causes the annotation to "escape" to the left.

https://jsfiddle.net/BlackLabel/5a379snx/

1. Draw a circle (Use the *Add Circle* button)
2. Move the circle

#### Product version
Highcharts v11.1.0

---
The issue is closely related to regression reported as #18459, but it seems that it has not been fully solved. Interestingly, [this workaround](https://github.com/highcharts/highcharts/issues/18459#issuecomment-1551424002) solves the problem.

**Workaround demo:** https://jsfiddle.net/BlackLabel/dyhf4csx/

kamil-musialowski:
Fixed #19233, dragging annotations on chart with null points was inconsistent.

_Internal note:_ Previous refactor of `getIndexOfPoint` ignored null points, because they have a property `isInside` set to false.
Demo: https://jsfiddle.net/BlackLabel/zuhw5tjr/

### DIFF
--- a/ts/Core/Axis/OrdinalAxis.ts
+++ b/ts/Core/Axis/OrdinalAxis.ts
@@ -1410,16 +1410,15 @@ namespace OrdinalAxis {
 
             // Check whether the series has at least one point inside the chart
             const hasPointsInside = function (series: Series): boolean {
-                return series.points.some(function (point): boolean {
-                    let hasPointsInside = false;
+                const { min, max } = axis;
 
-                    if (axis.min && axis.max) {
-                        hasPointsInside =
-                            point.x > axis.min && point.x < axis.max;
-                    }
+                if (defined(min) && defined(max)) {
+                    return series.points.some((point): boolean =>
+                        point.x > min && point.x < max
+                    );
+                }
 
-                    return hasPointsInside;
-                });
+                return false;
             };
 
             let firstPointX: number;
--- a/ts/Core/Axis/OrdinalAxis.ts
+++ b/ts/Core/Axis/OrdinalAxis.ts
@@ -1410,7 +1410,16 @@ namespace OrdinalAxis {
 
             // Check whether the series has at least one point inside the chart
             const hasPointsInside = function (series: Series): boolean {
-                return series.points.some((point): boolean => !!point.isInside);
+                return series.points.some(function (point): boolean {
+                    let hasPointsInside = false;
+
+                    if (axis.min && axis.max) {
+                        hasPointsInside =
+                            point.x > axis.min && point.x < axis.max;
+                    }
+
+                    return hasPointsInside;
+                });
             };
 
             let firstPointX: number;
--- a/ts/Core/Axis/OrdinalAxis.ts
+++ b/ts/Core/Axis/OrdinalAxis.ts
@@ -1414,7 +1414,7 @@ namespace OrdinalAxis {
 
                 if (defined(min) && defined(max)) {
                     return series.points.some((point): boolean =>
-                        point.x > min && point.x < max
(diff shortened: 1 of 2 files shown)


ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -1410,7 +1410,16 @@ namespace OrdinalAxis {
 
             // Check whether the series has at least one point inside the chart
             const hasPointsInside = function (series: Series): boolean {
-                return series.points.some((point): boolean => !!point.isInside);
+                return series.points.some(function (point): boolean {
+                    let hasPointsInside = false;
+
+                    if (axis.min && axis.max) {
```
pawelfus, line 1416: Nitpicking: move this condition outside of the `some()` method. There's no need to check this condition for every point, right?

pawelfus (changes requested):

ts/Core/Axis/OrdinalAxis.ts
kamil-musialowski: That's a good point, thank you!

I have also simplified the code to improve the readability.

kamil-musialowski: That's a good point, thank you

I have also simplified the code to improve readability.

kamil-musialowski: Also, I have added the `isNumber` check, since if `axis.min` would be 0, previously it would fail.

pawelfus (approved):
Looks good, waiting for tests 👍

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -1410,7 +1410,15 @@ namespace OrdinalAxis {
...
-                return series.points.some((point): boolean => !!point.isInside);
+                const { min, max } = axis;
+
+                if (defined(min) && defined(max)) {
+                    return series.points.some((point): boolean =>
+                        point.x >= min && point.x <= max
+                    );
+                }
+
+                return false;
```
TorsteinHonsi, line 1421: So this change is basically done because `point.isInside` isn't reliable, right? Do we know why `point.isInside` isn't up-to-date? Should we look into a fix that fixes that instead? That may fix other side-effects too, that we are not yet aware of.

TorsteinHonsi (changes requested):

ts/Core/Axis/OrdinalAxis.ts
kamil-musialowski: Yes, the `point.isInside` is failing, because for null points `plotY` property is set to `undefined`.
So it fails right there: https://github.com/highcharts/highcharts/blob/3f75b6145731b0624348c823872215d3e041143e/ts/Core/Series/Series.ts#L3649

As a fix, for `point.isNull === true` we could omit this check, and maybe check only for `plotX` properties?

For example, we could set the compare value of `point.plotY` to 0 if `point.isNull === true`?  Like -> `point.plotY || 0`?

TorsteinHonsi: Okay, as long as there's a good reason for it, I think we should just leave it like it is. As far as I understand, the `Point.isInside` flag is correct. Null points are not inside, as expected.

TorsteinHonsi (approved):

PR #19313, opened 2023-07-05
Samples: Friendlier geo point notation in twm demo

TorsteinHonsi:
More human friendly geo point notation in the twm demo. The `geometry` notation is JSON compatible and good for machines, but `lon` and `lat` options are convenient shorthands that are easier to read and write.


hubertkozik (approved):
Looks better and few lines fewer :)

PR #19316, opened 2023-07-06
docs/fixed-link-gantt-demo-main-doclects

madepiet:
Docs: Fixed link to gantt demo in main doclects.

### DIFF
--- a/docs/gantt/getting-started-gantt.md
+++ b/docs/gantt/getting-started-gantt.md
@@ -47,4 +47,4 @@ Simple example
 
 _See below the simple live example of a Gantt chart_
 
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (changes requested):
Thanks! Actually we are supposed to load iframes from our own website, not jsFiddle. So it should be `https://www.highcharts.com/samples/embed/gantt/gantt/simple-gantt-chart`.

If you search for "<iframe" in the docs folder you'll notice the pattern. Actually there are a few remaining cases of jsfiddle links, so maybe we should change those too in this PR?

docs/chart-and-series-types/histogram-series.md
```diff
@@ -57,4 +57,4 @@ Use the following setting to set up a histogram chart with a column chart:
         }
     }
 
-<iframe width="100%" height="470" style="null" src=https://jsfiddle.net/mushigh/18b1vun9/5/embedded/result allow="fullscreen"></iframe>
+<iframe width="100%" height="470" style="null" src=https://www.highcharts.com/samples/embed/highcharts/demo/histogram-column allow="fullscreen"></iframe>
```
TorsteinHonsi, line 60: Samples that are added to `{product}/demo/***` are the ones that appear on our website demo section. We already have a histogram demo and we haven't planned another.

Instead, we should move this sample to `highcharts/series-histogram/column`, for use only by reference from the docs.

TorsteinHonsi (changes requested):

TorsteinHonsi (approved):

TorsteinHonsi (approved):

PR #19317, opened 2023-07-06
Dash/api docs cleanup

pawellysy:
Fixed API docs links.


sebastianbochan (approved):

ts/Dashboards/Board.ts
```diff
@@ -662,15 +662,15 @@ namespace Board {
          *
          * Try it:
          *
-         * {@link https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/dashboards/demo/component-highcharts/  | Highcharts component}
+         * {@link https://www.highcharts.com/samples/embed/dashboards/components/component-highcharts | Highcharts component}
```
goransle, line 665: Sorry for being the samples nitpicker today, but here they should probably be without `embed` or `nonav`

pawellysy: So what the URL should be? 😅

karolkolodziej, line 665: What with the other `.ts` files in the `Dashboards` direcotry? Shouldn't we also change them?

pawelfus: I think we should keep jsFiddle link 😅 Just like 99.99% (there is one exception) Highcharts API links redirect to jsFiddle, not to our demo pages. Unless there's a plan to change that? 🤔

ts/Dashboards/Board.ts
```diff
@@ -662,15 +662,15 @@ namespace Board {
...
          *
-         * {@link https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/dashboards/components/component-html/ | HTML component}
+         * {@link https://www.highcharts.com/samples/embed/dashboards/components/component-html | HTML component}
          *
-         * {@link https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/dashboards/demo/component-kpi/ | KPI component}
+         * {@link https://www.highcharts.com/samples/embed/dashboards/components/component-kpi | KPI component}
          *
-         * {@link https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/dashboards/components/custom-component/ | Custom component}
+         * {@link https://www.highcharts.com/samples/embed/dashboards/components/custom-component | Custom component}
```
bre1470, line 671: This demo looks not right.

bre1470 (approved):

ts/Dashboards/Board.ts
goransle: If we want to use highcharts.com/samples it should be https://www.highcharts.com/samples/dashboards/components/component-highcharts. (they're pretty similar, but `/embed` has some extra magic for being placed in iframes)

pawellysy: Ok, let's just go with jsfiddle then 😇

PR #19321, opened 2023-07-07
Samples/highcharts - Updated keywords

andreastistel:
Updated keyword var to const/let in all the appropriate placed in samples/highcharts. 

For highcharts/boost/treemap: [https://eslint.org/docs/latest/rules/no-prototype-builtins]
ESlint did not like “if (data.hasOwnProperty(region)”.
Fixed with: “if ({}.hasOwnProperty.call(data, region))”


TorsteinHonsi (approved):
Thanks!

PR #19330, opened 2023-07-10
dash/columnAssignment-datagrid

karolkolodziej:
Added the `visibleColumns` option which allows filtering the columns that should be visible in the `DataGridComponent`.

### DIFF
--- a/samples/dashboards/cypress/component-datagrid/demo.mjs
+++ b/samples/dashboards/cypress/component-datagrid/demo.mjs
@@ -51,7 +51,8 @@ Dashboards.board('container', {
             },
             columnAssignment: {
                 Food: 'x',
-                'Vitamin A': 'y'
+                'Vitamin A': 'y',
+                notVisible: null
             },
             chartOptions: {
                 xAxis: {
@@ -83,9 +84,7 @@ Dashboards.board('container', {
                 highlight: true,
                 extremes: true
             },
-            columnAssignment: {
-                notVisible: null
-            }
+            columnsToShow: ['Food', 'Vitamin A']
         }
     ]
 }, true);
--- a/samples/dashboards/cypress/component-datagrid/demo.mjs
+++ b/samples/dashboards/cypress/component-datagrid/demo.mjs
@@ -82,6 +82,9 @@ Dashboards.board('container', {
             sync: {
                 highlight: true,
                 extremes: true
+            },
+            columnAssignment: {
+                notVisible: null
             }
         }
     ]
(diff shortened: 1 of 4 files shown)


pawellysy (changes requested):
I wonder, if this option should be named the same as in the HighchartsComponent. 
In the HighchartsComponent, this option transforms each of the columns to one of the values in the chart, so there are many options like `x`, `y`, `value`, `null` etc, and here there is only one option, `null`. So maybe we should create something like `ignoredColumns: Array<string> where all ignored columns are listed` or `visibleColumns` where the is other way around, or just delete this option from the API and just go with `columns.show = false`? 

Summoning @pawelfus

pawelfus:
Discussion summary: Green light for a different option, instead of `columnAssignment` one

pawellysy (approved):
LGTM!

ts/Dashboards/Plugins/DataGridComponent.ts
```diff
@@ -378,6 +381,39 @@ class DataGridComponent extends Component {
...
+     *
+     * @internal
+     */
+    private filterColumns(): DataTable|undefined {
+        const table = this.connector?.table.modified,
+            columnsToShow = this.options.columnsToShow;
+
+        if (table) {
+            const columnsToDelete = table.getColumnNames().filter((columnName): boolean => {
+                if (columnsToShow?.length) {
```
goransle, line 395: The method can probably just return `table` earlier `if( !columnsToShow.length )`.

bre1470 (commented):
🙂

bre1470 (approved):

samples/dashboards/cypress/component-datagrid/demo.mjs
```diff
@@ -82,7 +83,8 @@ Dashboards.board('container', {
             sync: {
                 highlight: true,
                 extremes: true
-            }
+            },
+            columnsToShow: ['Food', 'Vitamin A']
```
sebastianbochan, line 87: IMO `visibleColumns` or `displayColumns` sound more intuitive.

sebastianbochan (approved):

samples/dashboards/cypress/component-datagrid/demo.mjs
sebastianbochan: Similar suggestion to `notVisible` param, IMO `hidden` sounds more user-friendly.

goransle (approved):

PR #19336, opened 2023-07-11
Dash/implement gulp dashboards tests

pawellysy:
Implemented gulp tasks for Dashboards.

### DIFF
--- a/tools/gulptasks/dashboards/lint.js
+++ b/tools/gulptasks/dashboards/lint.js
@@ -28,7 +28,7 @@ async function task() {
     const processLib = require('../lib/process');
     const logLib = require('../lib/log');
     const SOURCE_GLOB = './**/*.ts';
-    logLib.message('Linting [.ts/Dashboards/**/*.ts]...');
+    logLib.message('Linting [./ts/Dashboards/**/*.ts]...');
 
     try {
         await processLib.exec('npx eslint --quiet "' + SOURCE_GLOB + '"', {
--- a/tools/gulptasks/dashboards/lint.js
+++ b/tools/gulptasks/dashboards/lint.js
@@ -10,7 +10,6 @@ const gulp = require('gulp');
  *
  * */
 
-const SOURCE_GLOB = './**/*.ts';
 
 /* *
  *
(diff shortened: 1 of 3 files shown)


.husky/pre-commit
```diff
@@ -6,5 +6,5 @@ npx gulp lint-ts && \
 npx gulp scripts && \
 npx ts-node ./test/ts-node-unit-tests && \
 npx gulp test && \
-npx gulp test-ts && \
-npx gulp dashboards/test
+npx gulp test-ts # && \
```
sebastianbochan, line 9: Do we need the commented line?

pawellysy: We don't,  I've just left for the future, when the task with implementing conditional  pre-commit hooks is ready

sebastianbochan (approved):

karolkolodziej (approved):

tools/gulptasks/dashboards/lint.js
```diff
@@ -0,0 +1,44 @@
...
+
+    const processLib = require('../lib/process');
+    const logLib = require('../lib/log');
+
+    return new Promise((resolve, reject) => {
+
+        logLib.message('Linting [./Dashboards/**/*.ts]...');
+
+        processLib
+            .exec('cd ts/Dashboards && npx eslint --quiet "' + SOURCE_GLOB + '"')
```
bre1470, line 37: Use the `cwd` option of `processLib.exec` instead of the `cd` command.
https://nodejs.org/api/child_process.html#child_processexeccommand-options-callback

You can also integrate the constant value here. It is not used anywhere else.

tools/gulptasks/dashboards/lint.js
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * @return {Promise<string>}
+ *         Promise to keep with console output
+ */
+function task() {
+
+    const processLib = require('../lib/process');
+    const logLib = require('../lib/log');
+
+    return new Promise((resolve, reject) => {
```
bre1470, line 32: Make the task async and simplifier with await like in the other task.

ts/Dashboards/Plugins/HighchartsComponent.ts
```diff
@@ -27,7 +27,7 @@ import type {
     AxisOptions,
     Chart,
     Options as ChartOptions,
-    Highcharts,
+    Highcharts as HighchartsConstructor,
```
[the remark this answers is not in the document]
bre1470: It is about linting rules.

But Highcharts is a namespace, not a constructor. `H` should do it.

ts/Dashboards/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+    "extends": "../tsconfig.json",
```
bre1470, line 2: We have the dashboard config in `ts/masters-dashboards`.

Why do we need this `tsconfig.json`?

bre1470 (changes requested):

ts/Dashboards/tsconfig.json
bre1470: Maybe you can replace it with a `tslint.json`? https://palantir.github.io/tslint/usage/configuration/

pawellysy: Hm, but the tsconfig will be added here -> https://github.com/highcharts/highcharts/pull/19355
So does it ultimately matter not to add it?

bre1470 (approved):
Thank you! Looks good. 👍

PR #19342, opened 2023-07-11
bugfix/19169-focus-border-remove

### LINKED ISSUE #19169: Focus border persists even after chart loses focus
#### Expected behaviour

The focus border should disappear when the chart no longer has focus.

#### Actual behaviour

The focus border persists even when another element has focus:

image

https://jsfiddle.net/onfe21zq/

- Click into the chart area and press Tab once to focus the first column
- Click into the input below the chart to focus that instead
- Notice that the focus border in the chart persists

#### Product version
11.1.0

#### Affected browser(s)
Edge (Chromium), haven't tested others

jakubSzuminski:
Fixed #19169, the `focusBorder` is now removed from `SVGElement` after the element has been unfocused.

**Demo before fix:** https://jsfiddle.net/BlackLabel/2zq1cgov/
**Demo after fix:** https://jsfiddle.net/BlackLabel/19wcmae8/

---
  - https://app.asana.com/0/0/1204915984120493


TorsteinHonsi (approved):
Thanks!

TorsteinHonsi:
Pending reviews from @oysteinmoseng and @marvin19

oysteinmoseng:
Hm the focus border is not necessarily drawn around the element that actually has focus, so not sure we can rely on `focusout` here. Did you investigate why [this code](https://github.com/highcharts/highcharts/blob/master/ts/Accessibility/KeyboardNavigation.ts#L297-L325) doesn't do what it is supposed to? In your demo it works if you tab to the legend before clicking the input field.

jakubSzuminski:
@oysteinmoseng sorry for the late response.

I've followed your suggestion and:

1. The `keyboardReset` variable is described as: `Used for resetting nav state when clicking outside chart`.

2. When clicking outside the chart with a mouse, the `keyboardReset` is always `true`.

3. So, from what I've seen here:
https://github.com/highcharts/highcharts/blob/e302e91a89034d83a9529043a07996940a22c9db/ts/Accessibility/KeyboardNavigation.ts#L297-L325

   The `focusBorder` is never removed because the condition in the most external if is never passed - the `keyboardReset` is always true when clicking outside of the chart.

4. I'm not sure what would be the best solution to this, but I was thinking about adding an additional if:
```JavaScript
const insideChartClick = chart.container.contains(e.target as HTMLElement);

if (!insideChartClick && chart.focusElement) {
   chart.focusElement.removeFocusBorder();
   delete chart.focusElement; 
}
```

This seems to solve this problem and passes all the tests. 

Should I proceed with this? Or would you like me to solve this in a different way?

TorsteinHonsi:
Pending review from @oysteinmoseng

oysteinmoseng:
Aha, well let me explain a bit, sounds like there is a very simple fix then.

The `keyboardReset` variable is just there so that we don't reset multiple times. So whenever we are keyboard navigating in the chart, it should be `false`, and whenever the keyboard navigation is "reset", and focus is somewhere else on the page, it should be set to `true`.

Meaning, it should be set to `false` immediately when we start keyboard navigation. I thought this would happen [here](https://github.com/highcharts/highcharts/blob/master/ts/Accessibility/KeyboardNavigation.ts#L353) on every keydown, but maybe on initializing that function is not called?

jakubSzuminski:
"If the initial focus is outside the element where the keydown event is bound, pressing "Tab" will move the focus to the next focusable element, and if that element is the one with the event listener, it won't have the listener triggered until the next "Tab" press. This is because the element gains focus only after the first "Tab" press is completed." - I believe that's the reason why the `keyboardReset` was not properly initialized as the `onKeyDown` function was not called after we focused on the first element in our container, only after the second.

According to this 👆 and your point @oysteinmoseng, I decided to add the `this.keyboardReset = false` after the `onFocus` event has been called on the `tabindexContainer` which seems to solve the problem.

oysteinmoseng (approved):
Seems sensible, thanks for the investigating! 🕵️

PR #19347, opened 2023-07-12
demo/line-labels update

andreastistel:
Updated the dataset to one with a bit more diversity.
Updated the color to be more vibrant.
[Asana task](https://app.asana.com/0/1203542399740129/1203542400258104/f)


TorsteinHonsi:
Closing this. Rainfall is not suitable data for line charts, as it is non-continuous and commonly depicted as columns.

PR #19354, opened 2023-07-12
bugfix/16368-stock-axis-breaks

### LINKED ISSUE #16368: xAxis breaks make OHLC/CandleStick series overlap.
#### Expected behaviour
No overlapping of individual candlesticks/ohlc points.

#### Actual behaviour
OHLC or CandeStick seriespoints are overlapping when making use of xAxis breaks to skip weekends.

https://jsfiddle.net/stenTT/Lxa0pmvo/25/

Uncomment the xAxis breaks in this jsfiddle to get the series to overlap.

#### Product version
Highstock JS v9.2.2 (2021-08-24)

#### Affected browser(s)
Firefox

kamil-musialowski:
Fixed #16368, axis breaks caused points to overlap in stock charts.

TODO:

- [x] Check why this one is not working https://jsfiddle.net/BlackLabel/p7x651e0/


TorsteinHonsi (approved):
Thanks!

pawelfus (approved):
Thanks!

TorsteinHonsi:
All good, now just waiting for the CI tests after merge

TorsteinHonsi (changes requested):
The changes to the `intraday-breaks` demo don't look legit... Please advice.

kamil-musialowski:
Indeed, it doesn't look right. It did look the same previously (once we disable the ordinal, since it was preventing the `breaks` from working) https://jsfiddle.net/BlackLabel/8sptbhr6/ , but the ticks were fine back then. I'll investigate it.

kamil-musialowski:
I think I got it, however one scenario has to be consulted first @pawelfus .

How do we want to treat the `breaks: [{ }]` config?

Since we check the breaks by `hasBreaks = (isArray(breaks) && !!breaks.length)`, then adding an empty object to the `breaks` array would result in `!!breaks.length === true`. 

If we want to prevent it, then I'd suggest adding the `!!Object.keys(breaks[0]).length` to the if check at https://github.com/highcharts/highcharts/blob/815d2def69770e22827569c6d5a65cd606390863/ts/Core/Axis/BrokenAxis.ts#L711 (imo it is enough to only check for an empty object at the first position, since configs like `[...normal breaks, { }, ...normal breaks]` work just fine)


...however, there is a test that checks this scenario and it expects `hasBreaks === true`
https://github.com/highcharts/highcharts/blob/815d2def69770e22827569c6d5a65cd606390863/samples/unit-tests/axis/breaks/demo.js#L600-L604

If we don't handle the [{ }] scenario, then everything else works fine, however we should be aware that such demos will be broken: https://jsfiddle.net/BlackLabel/o71pu86k/

pawelfus (changes requested):
See the comment above

kamil-musialowski:
Thank you, done!

Still need to check this one: https://jsfiddle.net/BlackLabel/v9gyuxaL/

kamil-musialowski:
Done, manually tested for all of the previous demos that were failing on the master.

Added the `(!hasBreaks)` check to prevent `ordinal.slope` and `ordinal.offset` calculation for broken axis, as it was interfering with the `getColumnMetrics` calculations.

pawelfus (approved):
Thanks!

TorsteinHonsi (approved):

PR #19358, opened 2023-07-13
demo/combo-regression update

andreastistel:
Updated the dataset and added a function that calculates the trend-line instead of it being hard coded.
Renamed "regression" to "trend".  

[Asana task](https://app.asana.com/0/1203542399740129/1203673412610690/f)


TorsteinHonsi (approved):
Thanks!

PR #19361, opened 2023-07-13
bugfix/19264-composition-on-spline-series

### LINKED ISSUE #19264: Incorrect flag position for spline series.
#### Expected behaviour
Flags placed on the spline series should be positioned correctly.

#### Actual behaviour
Flags on series where the line is not a straight line connecting adjacent points, but a curve, such as in the spline series, are positioned incorrectly.

https://jsfiddle.net/BlackLabel/jgnqpkxd/

#### Product version
Highcharts v11.1.0

Vazonik:
Fixed #19264, flag positioning on the spline series.

### DIFF
--- a/ts/Series/OnSeriesComposition.ts
+++ b/ts/Series/OnSeriesComposition.ts
@@ -20,6 +20,7 @@ import type CoreSeriesOptions from '../Core/Series/SeriesOptions';
 import type Point from '../Core/Series/Point';
 
 import ColumnSeries from './Column/ColumnSeries.js';
+import SplinePoint from './Spline/SplinePoint';
 const { prototype: columnProto } = ColumnSeries;
 import Series from '../Core/Series/Series.js';
 const { prototype: seriesProto } = Series;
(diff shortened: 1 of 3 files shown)


ts/Series/OnSeriesComposition.ts
```diff
@@ -183,17 +184,97 @@ namespace OnSeriesComposition {
...
+                                        pixelThreshold = 0.25,
+                                        maxIterations = 100,
+                                        calculateCoord = (
+                                            t: number,
+                                            key: 'plotX'|'plotY'
+                                        ): number | null => {
+                                            if (t < 0 || t > 1) {
+                                                return null;
+                                            }
+                                            return Math.pow(1 - t, 3) *
```
pawelfus, line 217: I would add a reference or short description that's the cubic bezier explicit form equation

pawelfus (changes requested):
Great job! I like the idea of the bisecting approach to find the correct t-value. I'm not sure if 100 iterations aren't too much, but in most scenarios, we won't cross 10 (for 0.25px precision, with 10 iterations, we get 256px distance between left and right points) so let's keep it. 

Requesting some minor improvements, though.

pawelfus (approved):

ts/Series/Spline/SplineSeries.ts
```diff
@@ -175,58 +175,77 @@ class SplineSeries extends LineSeries {
...
-                .attr({
-                    stroke: 'red',
-                    'stroke-width': 2,
-                    zIndex: 9
-                })
-                .add();
+            point.rightCont = {
+                plotX: rightContX,
+                plotY: rightContY
+            };
```
TorsteinHonsi, line 246: There's always a balance between readable and compact code, but since the concept of using tuples for coordinates is so common, I suggest we use `[x, y]` here. Also, to avoid two so similar properties on the point, I think we should just add one. Besides, the words _left_ and _right_ only make sense on a non-reversed, horizontal axis. On a vertical axis, or a reversed one, it is not semantically correct. So I suggest something like this:

```js
point.controlPoints = {
    low: [leftContX, leftContY],
    high: [rightContX, rightContY]
}
```

TorsteinHonsi (changes requested):

TorsteinHonsi (approved):
Thanks!

PR #19365, opened 2023-07-13
bugfix/13240-bubble-aligndatalabel-beyond-plotarea

### LINKED ISSUE #13240: Bubble Chart: Problem with positioning dataLabels when bubble is beyond plotArea
#### Expected behavior
Possibility to set dataLabels position in the middle of the bubble with available API options.

#### Actual behavior
In the case of this chart configuration where the bubble goes beyond plotArea, it is not possible to set dataLabels correctly, see demo: https://jsfiddle.net/BlackLabel/dutzanxh/ 

The documentation would indicate that with the options I used in the above demo it should be possible: https://api.highcharts.com/highcharts/series.bubble.dataLabels.overflow 

##### Internal notes
This issue may be related to this case: https://github.com/highcharts/highcharts/issues/8923

#### Product version
Highcharts v8.0.4

#### Affected browser(s)
Chrome 80.0.3987.149

madepiet:
Fixed #13240, using default data labels alignment for bubble series.


TorsteinHonsi (changes requested):
By removing that function we can no longer align the data label within the bubble:

https://jsfiddle.net/mnkgbo06/

I think instead we need to look into how the `overflow` setting work. Does it only work to horizontal alignment?

TorsteinHonsi:
@madepiet Bump

madepiet:
> By removing that function we can no longer align the data label within the bubble:
> 
> https://jsfiddle.net/mnkgbo06/
> 
> I think instead we need to look into how the `overflow` setting work. Does it only work to horizontal alignment?

Yes, I started investigating the issue of how it behaves after changing axisBeforePadding. I'm back to work on it now so I'll give an update

TorsteinHonsi (approved):

PR #19370, opened 2023-07-14
dash/productsjs

bre1470:


tools/gulptasks/dist-productsjs.js
```diff
@@ -26,45 +24,53 @@ const TARGET_FILE = Path.join('build', 'dist', 'products.js');
...
+        (
+            'var products = ' + JSON.stringify({
+                Highcharts: { date, nr },
+                'Highcharts Stock': { date, nr },
+                'Highcharts Maps': { date, nr },
+                'Highcharts Gantt': { date, nr },
+                'Highcharts Dashboards': dashboardsProduct
+            }, void 0, '    ') + '\n'
+        )
+    );
```
goransle, line 71: I am not sure what it's used for, but won't this overwrite the date for the Highcharts(& family) release when doing a Dashboards release?

goransle: Oh wait, there's two tasks

sebastianbochan (approved):

tools/gulptasks/dist-productsjs.js
bre1470: It is a bit messy. Maybe it is better to download the latest `products.js` from code.highcharts.com and modify it?

karolkolodziej (approved):

goransle (approved):

tools/gulptasks/dist-productsjs.js
bre1470: What do you think about fetching the latest `products.js`?

goransle: Not sure. I guess it could help avoid problems with the local dist-repo not being updated?

bre1470: Yes, that was the idea. The issue though can be, that the cache on the server is not immediately updated. So releases of several products would need a time buffer in-between. Or the cache purge needs to be automated.

sebastianbochan: Are we able to clear cache after release or only Sam can do it?

bre1470: I know at least three more people that can do it.

pawellysy (approved):

PR #19371, opened 2023-07-14
bugfix/19046-pareto-with-datasorting

### LINKED ISSUE #19046: Pareto doesn't work for series with `dataSorting` enabled.
#### Expected behaviour
Pareto series works also for the series with option `dataSorting.enabled: true`.

#### Actual behaviour
When the base series has the option `dataSorting.enabled: true`, pareto series isn't visible.

https://jsfiddle.net/BlackLabel/Ltru9fso/

#### Product version
Highcharts v11.0.1

jszuminski:
Fixed #19046, Pareto series did not work for series with `dataSorting` enabled.

**Demo of the issue:** https://jsfiddle.net/BlackLabel/75s1cak0/
**Demo after fix:** https://jsfiddle.net/BlackLabel/skjgtyLu/

**Tasks:**
- [x] Fix the issue
- [x] Add a test

### DIFF
--- a/samples/unit-tests/series-pareto/pareto/demo.js
+++ b/samples/unit-tests/series-pareto/pareto/demo.js
@@ -123,17 +123,20 @@ QUnit.test('Pareto', function (assert) {
 });
 
 QUnit.test(
-    'Pareto wasnt working with baseSeries set to 0 - #10471',
+    'Pareto with baseSeries set to 0 (#10471) and with dataSorting (#19046)',
     function (assert) {
         var chart = Highcharts.chart('container', {
             series: [
                 {
                     type: 'column',
-                    data: [155, 55, 231, 22, 72, 51, 36, 10]
+                    data: [155, 55, 231, 22, 72, 51, 36, 10],
+                    dataSorting: {
+                        enabled: true
+                    }
                 },
                 {
                     type: 'column',
-                    data: [755, 222, 151, 86, 72, 51, 36, 10]
+                    data: [1, 2, 3, 4, 5]
                 },
                 {
                     type: 'pareto',
@@ -146,6 +149,21 @@ QUnit.test(
             chart.series[2].points.length,
             chart.series[0].points.length,
             'Number of points in pareto series should be equal amount of point in assigned series'
+        );
+
+        assert.deepEqual(
+            chart.series[2].yData,
+            [
+                36.550632911392,
+                61.075949367089,
+                72.46835443038,
+                81.170886075949,
+                89.240506329114,
+                94.936708860759,
+                98.417721518987,
+                100
+            ],
+            'Pareto series should work if the base series has dataSorting enabled, #19046'
         );
     }
 );
--- a/ts/Series/ParetoSeries/ParetoSeries.ts
+++ b/ts/Series/ParetoSeries/ParetoSeries.ts
@@ -167,14 +167,44 @@ class ParetoSeries extends LineSeries {
      * @requires modules/pareto
      */
     public setDerivedData(): void {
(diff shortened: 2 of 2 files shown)


TorsteinHonsi:
@jakubSzuminski Thanks for the heads up!

I think we should look into all linked series. Otherwise we will just end up getting reports for this over and over, and creating repeated fixes for it. Let's see what it looks like with a general fix, then we'll see if we can optimize it somehow.

jakubSzuminski:
@TorsteinHonsi, me and @pawellysy looked into making it work for all the linked series. However, we'd like to ask you a few questions before we proceed with coding (as we encountered a few problems on the way and I think we need some of your guidance):

1. In the `sortData` method we override the `point.x` value. This does not seem to be a problem with the `xAxis.category`, but it does seem to be a problem with `xAxis.type = 'datetime'` or when, simply, we have a certain `pointInterval` set. Here's a **[demo](https://jsfiddle.net/BlackLabel/74hmLjq1/)** of the problem (the ticks of the `datetime` axis are not as expected). 
https://github.com/highcharts/highcharts/blob/d4da09b934f98a7b41e6d8a06e96e820ba52e2d3/ts/Core/Series/Series.ts#L1539-L1542
Should it work this way or should we modify its behavior? If yes, should we include an additional note in the docs that `dataSorting` works only for the `xAxis.type = 'category' | 'linear` and it ignores `pointInterval` (and `data.x`)? 

2. Is this **[behavior](https://jsfiddle.net/BlackLabel/yw3a1hm5/)** desired? Should we draw the line based on the order given by the user? Or should the behavior of `dataSorting` be the same as when the user sorts the data himself like in this **[demo](https://jsfiddle.net/BlackLabel/yw3a1hm5/2/)**?
A lot of the tests related to `dataSorting` should me modified if we get rid of the current assumption and start rendering the points based on the sorted order (mostly related to `xData` and `yData`).

3. The other "subproblem" that is there to solve is related to calling `linkSeries()` first and then calling the `setSeriesData()` method: https://github.com/highcharts/highcharts/blob/d4da09b934f98a7b41e6d8a06e96e820ba52e2d3/ts/Core/Chart/Chart.ts#L2665-L2666
which makes the sorted data inaccessible in the linked series, but that is something we think we know how to solve. Our main consternation/difficulty relates to the points (1) and (2).

TorsteinHonsi:
Thanks for asking!

1. So data sorting doesn't work when we have _explicit_ x data, whether it be set through `pointInterval`, `pointStart` and similar, or as an `x` value on each individual data point. I think that makes sense, and I am not sure how we should deal with sorting without this. So I suggest we find out how we can check for explicit x values as cheaply as possible, then opt out of sorting in that case. 

2. I would say the current line makes sense. I suggest we do nothing, at least until we get a request or issue report on it.

jakubSzuminski:
1. Fully understood.
2. If we decide to stay with the current approach of sorting data, fixing the `linkSeries()` and `setSeriesData()` methods will not solve the problem for line series (like Pareto). Example: https://jsfiddle.net/BlackLabel/e6gtua9r/ (even if the data is available in the `setDerivedData` methods on the child series, the points will not be drawn in the correct order)

---
The two problems when the `baseSeries.enabledDataSorting` are:
1. the `baseSeries.xData, baseSeries.yData` is not available in the `setDerivedData()` called on the child series
2. the `baseSeries.xData, baseSeries.yData` is not given in an order that would allow to draw a 'proper' line series

I think that the first problem can be solved easily, but the second one would require to somehow modify the `sortData` method. Otherwise, we would encounter issues like this: https://jsfiddle.net/BlackLabel/e6gtua9r/

Do you have any ideas what could be done with it?

PR #19334, opened 2023-07-11
bugfix/18988-inverted-sankey-offsets

hubertkozik:
Fixed #18988 and #17594, vertical and horizontal offsets for inverted Sankey and Organization charts didn't work correctly.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1204996793187160

PR #19385, opened 2023-07-17
samples/vertical-sankey

hubertkozik:
Added new option, [sankey.linkColorMode](https://api.highcharts.com/highcharts/series.sankey.linkOpacity).

----

Added a Vertical Sankey demo.

### Important
This branch is based on a fix from [this PR](https://github.com/highcharts/highcharts/pull/19334). This branch should be merged after the one with a fix. 

---
  - https://app.asana.com/0/0/1203935305712004


marvin19 (approved):

KacperMadej (approved):

MarkusBarstad (approved):

ts/Series/Sankey/SankeySeriesDefaults.ts
```diff
@@ -213,6 +213,28 @@ const SankeySeriesDefaults: PlotOptionsOf<SankeySeries> = {
...
+     *
+     * @sample highcharts/demo/vertical-sankey
+     *         Vertical sankey diagram with gradients
+     * @sample highcharts/series-sankey/link-color-mode
+     *         Sankey diagram with gradients and explanation
+     *
+     * @type      {('from'|'gradient'|'to')}
+     * @since     @next
+     * @default   from
+     * @apioption plotOptions.sankey.levels.linkOpacity
```
TorsteinHonsi, line 234: The `@default` can be removed as it is inferred from the code. The `@apioption` is just wrong, isn't it?

samples/highcharts/demo/vertical-sankey/demo.js
```diff
@@ -0,0 +1,300 @@
...
+                rotation: -80
+            }
+        }, {
+            id: 'Electric water-related applications',
+            offsetHorizontal: '-250%',
+            color: 'rgba(255, 141, 100, 0.8)',
+            dataLabels: {
+                rotation: -90,
+                x: -10,
+                format: '{addBR point.name}'
```
TorsteinHonsi, line 137: We can achieve word-wrap by setting a width to the data label. Which I think is better because we don't have to add a custom template helper, and can achieve what we want with pure configuration.

```
style: {
     width: '90px'
}
```

TorsteinHonsi (changes requested):

samples/highcharts/demo/vertical-sankey/demo.js
```diff
@@ -0,0 +1,300 @@
...
+Highcharts.Templating.helpers.addBR = value => value.split(' ').join('<br/>').toUpperCase();
+
+Highcharts.chart('container', {
+    chart: {
+        type: 'sankey',
+        inverted: true,
+        height: 1000
+    },
+    title: {
+        text: 'Evaluating the energy consumed for water use in the United States in 2010'
```
TorsteinHonsi, line 11: Can we find some recent data? Using data from 2010 sends the unconscious message that we created this demo over a decade ago. Alternatively just remove the year from the title.

TorsteinHonsi (changes requested):

samples/highcharts/demo/vertical-sankey/demo.js
hubertkozik: Sadly, there is no newer data, so I think removing the year would be good for that case.

ts/Series/Sankey/SankeySeriesDefaults.ts
```diff
@@ -213,6 +213,27 @@ const SankeySeriesDefaults: PlotOptionsOf<SankeySeries> = {
...
+     * - `to` color of the sankey link will be same as the 'to node'.
+     *
+     * @sample highcharts/demo/vertical-sankey
+     *         Vertical sankey diagram with gradients
+     * @sample highcharts/series-sankey/link-color-mode
+     *         Sankey diagram with gradients and explanation
+     *
+     * @type      {('from'|'gradient'|'to')}
+     * @since     @next
+     * @apioption plotOptions.sankey.levels.linkOpacity
```
TorsteinHonsi, line 233: The `@apioption` tag is still wrong...

TorsteinHonsi (changes requested):

ts/Series/Sankey/SankeySeriesDefaults.ts
```diff
@@ -213,6 +213,27 @@ const SankeySeriesDefaults: PlotOptionsOf<SankeySeries> = {
...
+     * - `to` color of the sankey link will be same as the 'to node'.
+     *
+     * @sample highcharts/demo/vertical-sankey
+     *         Vertical sankey diagram with gradients
+     * @sample highcharts/series-sankey/link-color-mode
+     *         Sankey diagram with gradients and explanation
+     *
+     * @type      {('from'|'gradient'|'to')}
+     * @since     @next
+     * @apioption plotOptions.sankey.levels.linkColorMode
```
TorsteinHonsi, line 233: The `linkColorMode` option is not only on `levels`. It is primarily a top-level option on the sankey series. This `@apioption` tag can probably be removed, since all the members on this object literal belong to the series options. As you can see with other options like `linkOpacity` and `opacity`. But I highly recommend you build the API (`gulp jsdoc`) and verify that the option is documented at the right place of the tree.

TorsteinHonsi (changes requested):

ts/Series/Sankey/SankeySeriesDefaults.ts
hubertkozik: Yes, I am currently pushing new changes, I did another oversight during the previous commit... I am really sorry for that. Totally understand, that this option is not on `levels`.

TorsteinHonsi (approved):

PR #19386, opened 2023-07-17
feature/dashboards-redraw-event

sebastianbochan:
Added user-defined `redraw` event for dashboard components.

### DIFF
--- a/test/cypress/integration/Dashboards/Components/events.cy.js
+++ b/test/cypress/integration/Dashboards/Components/events.cy.js
@@ -1,0 +1,49 @@
+function dropComponent(elementName) {
+  cy.get(elementName).first().trigger('mouseenter', {force: true});
+  cy.get(elementName).first().trigger('mousemove', 'right', {force: true});
+  cy.get(elementName).first().trigger('mouseup', 'right', {force: true});
+}
+
+describe('Component events', () => {
+  before(() => {
+    cy.visit('/dashboards/cypress/chart-interaction');
+  })
+
+  it('Mount event should be triggered on init', () => {
+    cy.get('#mount').should('have.value', 'mount');
+  });
+
+  it('Resize event should be triggered after setting init size', () => {
+    cy.get('#resize').should('have.value', 'resize');
+  });
+
+  // should be uncomment when unmount bug will be fixed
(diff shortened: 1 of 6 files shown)


ts/Dashboards/Components/Component.ts
```diff
@@ -730,6 +730,10 @@ abstract class Component {
             width,
             height
         });
+
+        fireEvent(this, 'redraw', {
+            component: this
+        });
```
pawellysy, line 736: IDK if this is accurate. it is added in the `resize` method. Apart from that, there already is a `redraw` event in the Component Class. 
https://github.com/highcharts/highcharts/blob/4181cfcd97f85e7de70fb514333434ce1c49f7b0/ts/Dashboards/Components/Component.ts#L966

pawellysy (changes requested):

ts/Dashboards/Components/Component.ts
karolkolodziej: Perhaps this should be changed?

pawellysy (approved):
So there is no changes in the core, just few tests added, right? If so, Looks good to me :D

test/cypress/integration/Dashboards/Components/events.cy.js
```diff
@@ -0,0 +1,49 @@
...
+        .trigger('mousemove', { clientX: 300 })
+        .trigger('mouseup');
+
+    cy.get('#resize').should('have.value', 'resize');
+  });
+
+  it('Update event should be triggered after sidebar update', () => {
+    cy.get('.highcharts-dashboards-component').first().click();
+    cy.get('.highcharts-dashboards-edit-toolbar-cell > .highcharts-dashboards-edit-toolbar-item:nth-child(2)').click();
+    cy.contains('Confirm').click();
```
karolkolodziej, line 47: You forgot to check the update event.

karolkolodziej (approved):

karolkolodziej (approved):

test/cypress/integration/Dashboards/Components/events.cy.js
sebastianbochan: Good point!

PR #19387, opened 2023-07-17
bugfix/19368-treegraph-collapsebutton-opacity

### LINKED ISSUE #19368: TreeGraph chart with collapsed levels shows initially one extra node expander with tooltip for hidden child
#### Expected behaviour
When I initialize TreeGraph chart with certain levels collapsed it should render with that levels collapsed.
image

#### Actual behaviour
When I initialize TreeGraph chart with certain levels collapsed it renders extra collapse node and when you hover over that node the tooltip for hidden child is shown.
image

https://jsfiddle.net/80kp4fwb

#### Product version

Highcharts 11.1.0

#### Affected browser(s)
Chrome, Firefox

madepiet:
Fixed #19368, `TreeGraph` chart displayed extra node expander for hidden child when levels collapsed. Also fixed `collapseButton` visibility.
___
**Internal note:**
- [x] Handle animation collapse button
- [x] Improvement test


ts/Series/Treegraph/TreegraphPoint.ts
```diff
@@ -154,6 +154,10 @@ class TreegraphPoint extends TreemapPoint {
             if (btnOptions.onlyOnHover && !point.collapsed) {
                 point.collapseButton.attr({ opacity: 0 });
             }
+
+            point.collapseButton.attr({
+                opacity: point.visible ? 1 : 0
```
sebastianbochan, line 159: You can use `+point.visible` instead

TorsteinHonsi:
Thanks!

1. As you can see in the visual diffs, the buttons now appear for all items, not only the first one that is hovered. Do you know why? This doesn't happen when testing the same charts in the utils, but we need to understand why it happens in the screenshots.
2. The discription for the changelog needs to be more descriptive: "Fixed #19368, treegraph collapseButton visibility."
   - What about the collapse button visibility? We must write so that the readers recognize the problem. A good starting point is the issue title.
   - Either write code (in backticks), or plain English. So either `collapseButton` or collapse button.

TorsteinHonsi (approved):

PR #19389, opened 2023-07-18
Dash/datagrid edit modified

pawellysy:
Implemented disabling of editing of the columns, which are created by Math Modifier.


sebastianbochan (approved):

ts/Dashboards/Plugins/DataGridComponent.ts
```diff
@@ -24,12 +24,16 @@ import type DataGrid from '../../DataGrid/DataGrid';
 import type DataTable from '../../Data/DataTable';
 import type DataGridOptions from '../../DataGrid/DataGridOptions';
 import type BaseDataGridOptions from '../../DataGrid/DataGridOptions';
```
karolkolodziej, line 26: Do we need these two imports? Looks like we import the same thing

ts/Dashboards/Plugins/DataGridComponent.ts
```diff
@@ -227,6 +236,36 @@ class DataGridComponent extends Component {
...
+     * @param connector
+     * Attached connector
+     */
+    private disableEditingModifiedColumns(connector: DataConnectorType) {
+        const dataModifier = connector.options.dataModifier;
+
+        if (!dataModifier || dataModifier.type !== 'Math') {
+            return;
+        }
+        const modifierColumns = (dataModifier as MathModifierOptions).columnFormulas;
```
karolkolodziej, line 252: Most likely we don't need this casting.

ts/Dashboards/Plugins/HighchartsComponent.ts
```diff
@@ -602,7 +604,8 @@ class HighchartsComponent extends Component {
                     .modifyTable(this.presentationTable).modified;
             }
 
-            const table = this.presentationTable;
+            const table = this.presentationTable,
+                modifierOoptions = table.getModifier()?.options;
```
karolkolodziej, line 608: typo

karolkolodziej (changes requested):

ts/Dashboards/Plugins/DataGridComponent.ts
```diff
@@ -22,7 +22,6 @@
 import type Cell from '../Layout/Cell';
 import type DataGrid from '../../DataGrid/DataGrid';
 import type DataTable from '../../Data/DataTable';
-import type DataGridOptions from '../../DataGrid/DataGridOptions';
 import type BaseDataGridOptions from '../../DataGrid/DataGridOptions';
```
karolkolodziej, line 25: I'm voting in favour of `DataGridOptions` instead of `BaseDataGridOptions`

pawellysy: There is a name conflict, so maybe let's leave the `BaseDataGridOptions` :)

karolkolodziej: ah, i didn't see that ;)

karolkolodziej (approved):

samples/dashboards/demo/datagrid-mathmodifier/demo.html
```diff
@@ -7,12 +7,12 @@
...
-16, 23, 1.0881
-17, 15, 1.0829
-18, 27, 1.0813
-19, 13, 1.0808</pre>
+<pre id="csv">Day,EUR,Rate
+15,11,1.0876
+16,23,1.0881
+17,15,1.0829
+18,27,1.0813
+19,13,1.0808</pre>
```
pawelfus, line 15: Why changing this? 🤔

ts/Dashboards/Plugins/HighchartsComponent.ts
```diff
@@ -656,9 +659,22 @@ class HighchartsComponent extends Component {
...
+                        (formula) => formula.column == seriesName
+                    )
+                );
+
                 return chart.addSeries({
                     name: seriesName,
-                    id: `${storeTableID}-series-${index}`
+                    id: `${storeTableID}-series-${index}`,
+                    dragDrop: {
+                        draggableY: shouldBeDraggable
```
pawelfus, line 676: We need this info in API - any thoughts on where we should place it? Not in Highcharts API obviously.

Perhaps in the general description of the Chart component? We could list default options for the series (and the chart) there.

pawelfus (changes requested):

samples/dashboards/demo/datagrid-mathmodifier/demo.html
pawellysy: with spaces, the name of the series will be ` EUR` with the extra space in the beginning, which may cause some bugs, when we will try to test something in the future. I thought that a good idea would be to fix it here by the way.

pawelfus (approved):
LGTM

PR #19390, opened 2023-07-18
Usdeur.json with fetch 2

andreastistel:
Added async wrapper and usdeur.json with fetch to demo.js.
Deleted usdeur.js from demo.html


TorsteinHonsi (approved):
Thanks!

PR #19391, opened 2023-07-18
Dashboards/demo styling

nasin:
Style suggestions for dashboard demos


bre1470 (changes requested):
Looks awesome! One thing in minimal demo: The RDA label of the column chart is better placed on the right side, otherwise the contrast is not strong enough.

samples/dashboards/demo/datacursor-sync/demo.js
```diff
@@ -40,12 +40,40 @@ Dashboards.board('container', {
...
+
     return {
         chart: {
+            spacing: [20, 20, 0, 20],
+            margin: [50, 50, 80, 100],
             events: {
                 load: function () {
                     const chart = this;
                     const series = chart.series[0];
+                    console.log(series);
```
sebastianbochan, line 53: To remove

samples/dashboards/demo/datacursor-sync/demo.js
```diff
@@ -40,12 +40,40 @@ Dashboards.board('container', {
...
+                                spacing: 30
+                            },
+                            legend: {
+                                enabled: true,
+                                layout: 'vertical',
+                                align: 'center',
+                                x: -200,
+                                verticalAlign: 'middle'
+                            }
+                        });
```
sebastianbochan, line 68: Replace `});` with `}, false)` to avoid double redraw.

samples/dashboards/demo/sync-extremes/demo.css
```diff
@@ -2,6 +2,28 @@
 @import url("https://code.highcharts.com/css/datagrid.css");
 @import url("https://code.highcharts.com/css/dashboards.css");
 
+.highcharts-background-color {
+    fill: #f0f0f0;
+}
+
+/* .highcharts-dashboards-wrapper {
+    background-color: #46465c;
+} */
```
sebastianbochan, line 11: To remove

sebastianbochan (changes requested):

pawellysy (commented):
Looks good to me, just a minor comment -> 
There is a bunch of changes in this PR https://github.com/highcharts/highcharts/pull/19306
which are waiting for Torstein's approval, and there is a bit of overlap on the changes.
Should we merge the changes here, or just leave other PR until it gets reviewed?

@pawelfus

pawelfus:
Let's include Torstein in this PR too @pawelfus

samples/dashboards/demo/sync-extremes/demo.css
nasin: done

samples/dashboards/demo/datacursor-sync/demo.js
nasin: done

bre1470 (approved):
Thanks! Looks good. 👍

samples/dashboards/demo/sync-extremes/demo.js
```diff
@@ -97,10 +162,184 @@ Dashboards.board('container', {
...
+                },
+                enabled: true,
+                padding: 0,
+                floating: true,
+                y: -30,
+                navigation: {
+                    enabled: false
+                }
+            },
+            series: [{
```
pawelfus, line 258: We need to remove all `series` options - it should be possible to get the same effect without `series`. 

Something like this: https://jsfiddle.net/BlackLabel/9zd4aLs1/3/ (in JS tab: modified config for the component, in CSS disabled inactive state at the bottom)

pawelfus (changes requested):
Just one tiny change, other than that it looks good!

samples/dashboards/demo/datagrid-sync/demo.js
```diff
@@ -1,18 +1,52 @@
...
+        }
+    },
     chart: {
         animation: false,
-        type: 'column'
+        type: 'column',
+        events: {
+            load: function () {
+                const chart = this;
+                if (chartCount === 1) {
```
sebastianbochan, line 20: You don't need to use counter. You can remove the whole load event and use spread operator in component chartOptions.

See the comment below.

samples/dashboards/demo/datagrid-sync/demo.js
```diff
@@ -60,7 +94,7 @@ const board = Dashboards.board('container', {
                 'Vitamin A': 'y'
             },
             title: {
-                text: 'allowConnectorUpdate: true'
+                text: 'Vitamin A'
             },
             chartOptions
```
sebastianbochan, line 99:
```
chartOptions: {
    ...chartOptions,
   {
        chart: {
           events: {
               load: function () {
                   // call update
               }
           }
        }
   }
}
```

sebastianbochan (changes requested):

pawellysy (approved):

sebastianbochan (approved):
Looks great!

pawelfus (approved):
Thanks!

samples/dashboards/demo/minimal/demo.css
TorsteinHonsi, line 1: This demo has several cases of overlapping and clipped text on small screen widths. Particularly in mobile view, where the three columns in the top probably should be resolved to 100% width.

samples/dashboards/demo/minimal/demo.js
TorsteinHonsi, line 1: Instead of a separate cell for the title, we should consider a good old `h1` or `h2` here. The frame around the title looks out of place, besides the title text is clipped on small screens.

samples/dashboards/demo/minimal/demo.js
```diff
@@ -89,11 +91,44 @@ Dashboards.board('container', {
             xAxis: {
                 type: 'category'
             },
+            yAxis: {
+                title: {
+                    text: 'mcg'
+                },
+                plotLines: [{
+                    value: 900,
+                    zIndex: 20,
```
TorsteinHonsi, line 100: A z index of 20 makes the tooltip render below the plot line. Which looks buggy.

samples/dashboards/demo/sync-extremes/demo.js
TorsteinHonsi, line 1: It took me a while to make the connection that all the columns represented different data of the same city. Big color legends are not very (cognitively) accessible, so it is hard to read and compare the charts. Instead of doing this split, can we try to make the font size smaller for the city names, and maybe force a vertical rotation?

TorsteinHonsi, line 1: This demo too lacks reponsive rules. On mobile screens, the three charts in the top are too narrow. If possible, we should set the breakpoint so high that the axis labels are always readable.

TorsteinHonsi: PS: I may be that those charts would be better off as bar charts (with vertical drag-zoom). Then the labels would always be readable, and as a bonus each city would always be on the same line.

samples/dashboards/demo/datacursor-sync/demo.js
```diff
@@ -10,8 +10,18 @@ Dashboards.board('container', {
             id: 'dashboards-layout-1',
             rows: [{
                 cells: [{
+                    responsive: {
+                        small: {
```
oysteinmoseng, line 14: Responsive options (going from small to large) don't seem to take effect unless the page is reloaded, is this a bug or config?

samples/dashboards/demo/minimal/demo.css
```diff
@@ -6,6 +6,67 @@
...
+#kpi-iron .highcharts-dashboards-component-kpi-value {
+    font-size: 4em;
+    font-weight: normal;
+    line-height: 2.4rem;
+    margin-top: 10px;
+    color: var(--highcharts-color-0);
+}
+
+#kpi-iron .highcharts-dashboards-component-kpi-value {
+    color: var(--highcharts-color-1);
```
oysteinmoseng, line 50: The iron value color is very low contrast against the background in dark mode. We should probably not try to use the same text colors between light/dark, as that is hard to make work.

samples/dashboards/demo/minimal/demo.css
```diff
@@ -6,6 +6,67 @@
...
+    font-size: 1rem;
+}
+
+.highcharts-plot-line {
+    stroke-dasharray: 2px;
+    stroke: #b73c28;
+}
+
+.highcharts-plot-line-label {
+    fill: #b73c28;
```
oysteinmoseng, line 71: This one is low contrast too in dark mode.

samples/dashboards/demo/minimal/demo.js
```diff
@@ -21,18 +21,45 @@ Dashboards.board('container', {
         layouts: [{
             rows: [{
                 cells: [{
-                    id: 'title'
-                }]
-            }, {
-                cells: [{
+                    responsive: {
```
oysteinmoseng, line 24: Is the data grid a bit tall for the content?

Screen Shot 2023-08-19 at 15 19 43

oysteinmoseng (changes requested):
Nice work! Looks a lot cleaner. Added some comments, but I'm not sure how many are relevant to the styling changes, so if any should be separate tasks instead feel free to just add them in Asana.

samples/dashboards/demo/datagrid-sync/demo.js
```diff
@@ -4,15 +4,30 @@ const chartOptions = {
...
     },
+    credits: {
+        enabled: false
+    },
     title: {
+        text: 'allowConnectorUpdate: true'
+    },
+    subtitle: {
         text: 'Drag points to update the data grid'
     },
```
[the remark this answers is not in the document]
pawellysy: IMO centered title/subtitle looks good. Maybe we should change the text itself?

samples/dashboards/demo/axes-sync/demo.js
```diff
@@ -83,6 +77,18 @@ Dashboards.board('container', {
...
+        {
+            cell: 'title',
+            type: 'HTML',
+            elements: [{
+                tagName: 'h1',
+                textContent: 'Polio (Pol3) immunization coverage'
+            },
+            {
+                tagName: 'p',
+                textContent: ' Among 1-year-olds (%) | Source: https://apps.who.int/gho/data/'
```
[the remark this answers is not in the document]
pawellysy: I guess this is a bug of HTMLComponent.

samples/dashboards/demo/datacursor-sync/demo.js
pawellysy: I think this is a bug

samples/dashboards/demo/minimal/demo.js
pawellysy: that's something that happens to the DataGridComponent by default. Something that we might want to add to backlog.

oysteinmoseng (changes requested):
Nice improvements 👍  I will add the relevant non-styling issues to the backlog. 

Couple more things:
- Subtitle in axes-sync demo needs some bottom padding (or remove the background color for the component)
- Climate demo still struggles on mobile (see pic)
Screen Shot 2023-08-22 at 12 22 17

TorsteinHonsi (approved):
Approved my myself. Now pending @oysteinmoseng's approval.

PR #19395, opened 2023-07-19
demo/pie-gradient update

andreastistel:
Updated the dataset to distinguish it from other pie charts.
Changed the connector style.
[Asana task](https://app.asana.com/0/1203542399740129/1203592881281923/f)


TorsteinHonsi (approved):
Thanks!

PR #19401, opened 2023-07-19
dash/fix-dts

bre1470:
Improved TypeScript-support of Dashboards.

### DIFF
--- a/ts/Data/Connectors/CSVConnectorOptions.d.ts
+++ b/ts/Data/Connectors/CSVConnectorOptions.d.ts
@@ -33,25 +33,25 @@ export interface CSVConnectorOptions extends DataConnectorOptions {
     /**
      * Data in CSV format passed directly to connector as a string.
      */
-    csv: string;
+    csv?: string;
     /**
      * The URL to a remote CSV dataset
      */
-    csvURL: string;
+    csvURL?: string;
     /**
      * The rate in seconds for polling for live data.
      * Note that polling requires the option `enablePolling` to be true.
      */
-    dataRefreshRate: number;
+    dataRefreshRate?: number;
     /**
      * Whether to enable polling for live data.
      */
-    enablePolling: boolean;
+    enablePolling?: boolean;
     /**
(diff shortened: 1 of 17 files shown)


ts/Data/Connectors/CSVConnectorOptions.d.ts
```diff
@@ -33,25 +33,25 @@ export interface CSVConnectorOptions extends DataConnectorOptions {
...
      * Whether to enable polling for live data.
      */
-    enablePolling: boolean;
+    enablePolling?: boolean;
     /**
      * Whether to treat the first row of the data set as series names.
      * @default true
      */
-    firstRowAsNames: boolean;
+    firstRowAsNames?: boolean;
```
sebastianbochan, line 54: What is the reason for replacing all params with optional?

`Partial<>` is not enough then?

bre1470: It is enough, but non-optional would indicate a default value, which is incorrect.

ts/Shared/Types.ts
```diff
@@ -0,0 +1,45 @@
+/* *
+ *
+ *  (c) 2010-2021 Torstein Honsi
```
karolkolodziej, line 3: Are you sure?

ts/Shared/Types.ts
```diff
@@ -0,0 +1,45 @@
...
+
+'use strict';
+
+/* *
+ *
+ *  Declarations
+ *
+ * */
+
+namespace Types {
```
karolkolodziej, line 19: I'm not convinced do the naming, how about `SharedTypes`  or `GlobalTypes`?

karolkolodziej (approved):

ts/Shared/Types.ts
bre1470: This seems to be redundant as it already is in the `Shared`. `Global` would be not accurate as it is not part of the global namespace.

PR #19407, opened 2023-07-20
bugfix/19203-update-chart-plotoptions-pointstart

### LINKED ISSUE #19203: pointStart in plotOptions doesn't update when set alone
#### Expected behaviour
When setting pointStart in plotOptions I expect it to be updated when the property changes.
I want to be able to set pointStart only and not be forced to also set pointINterval in the plotOptions.

#### Actual behaviour
When *only* pointStart is set, it updates once and then never again.

Only if both pointStart and pointInterval is set it works how its supposed to work
Here is an example: https://stackblitz.com/edit/react-uu6kug?file=components%2FContainer.jsx

Looking at this old issue of yours it's supposed to work: https://github.com/highcharts/highcharts/pull/13949. But it does not.

 https://stackblitz.com/edit/react-uu6kug?file=components%2FContainer.jsx

#### Product version
"highcharts": "^11.1.0",
 "highcharts-react-official": "^3.2.0",
 "react": "^17.0.2",

#### Affected browser(s)
I tested in Chrome.

jedrzejruta:
Fixed #19203, `plotOptions.series.pointStart` couldn't be updated more than once.

Added tests for plot options updates.


ts/Core/Series/Series.ts
```diff
@@ -4104,14 +4112,20 @@ class Series {
...
             typeof options.pointInterval !== 'undefined' ||
             typeof options.relativeXValue !== 'undefined' ||
+
+            // Check if below `plotOptions.series` props are defined (#19203)
+            typeof plotOptions.series?.dataGrouping !== 'undefined' ||
+            typeof plotOptions.series?.pointStart !== 'undefined' ||
+            typeof plotOptions.series?.pointInterval !== 'undefined' ||
+            typeof plotOptions.series?.pointIntervalUnit !== 'undefined' ||
+            typeof plotOptions.series?.keys !== 'undefined' ||
+
```
hubertkozik, line 4122: I think those lines are not necessary for that fix. It is checked in `hasOptionsChanged` method (`defined`) and also currently (in these lines) you are not checking plot options for exact series type (e.g. column). After removing these lines all tests pass.

ts/Core/Series/Series.ts
```diff
@@ -4339,22 +4353,21 @@ class Series {
...
             plotOptions = chart.options.plotOptions,
-            oldOption = (this.userOptions as any)[optionName];
+            oldOption = this.userOptions[optionName as keyof DeepPartial<SeriesOptions>],
+            plotOptionsOption = pick(
+                plotOptions?.[this.type]?.[optionName as keyof Omit<SeriesOptions, NonPlotOptions>],
+                plotOptions?.series?.[optionName as keyof Omit<SeriesOptions, NonPlotOptions>]
+            );
 
-        if (oldOption) {
+        if (oldOption && !defined(plotOptionsOption)) {
```
hubertkozik, line 4364: It would be good to add a comment here with the issue number and a short description of why it is here.

hubertkozik (changes requested):
Good job! Two comments below.

hubertkozik (approved):
LGTM!

TorsteinHonsi (approved):
Thanks!

PR #19408, opened 2023-07-20
docs/19366-restricted-draggable-points

### LINKED ISSUE #19366: [draggable-points] Restrict which points can be dragged
Thank you for this excellent library.

# Description of the feature

Our application renders a chart that the user can modify by dragging and dropping points (in the y-direction only). Here's an example of a curve with a single manual adjustment:

image

The chart displays ~50 _control points_, which are denoted by the circular markers. The total number of data points in the series is much greater, e.g. 50*8 = 400. Rendering _intermediate points_ in between the control points is necessary for the curve to appear smooth, and to ensure that the graph has a high enough "resolution" to capture short-lived fluctuations in the data.

**We want to make it so the control points are draggable, without also making the intermediate points draggable.** I have searched the draggable-points docs and cannot find any way to restrict which points are draggable.

This would resolve the following issues with the current implementation where all points are draggable:

1. Currently, the user can drag an intermediate point, but then we reset the point's position when they drop it. This might confuse the user and make them wonder why they were able to drag the point if it did not end up influencing the curve.
2. Currently, you have to be very precise with your mouse movements when dragging a control point (in the y-direction). If you move your mouse very slightly to the left or right, you will switch to dragging an intermediate point. The UX is very poor. To mitigate this problem, we have to limit the number of intermediate points between each control point to 3 or fewer. This results in a somewhat jagged and low-resolution curve.

# API Design Ideas

I can think of two different ways to design the API for this feature.

## Idea 1 (Probably the most user-friendly)

When defining the data points for a series, allow passing an `isDraggable` boolean for each point. E.g.

```ts
const data = [
  { x: 0, y: 0, isDraggable: false },
  { x: 1, y: 0, isDraggable: true },
  /* ... */
]

const series = { 
  data,
  /* ... */
}
```

## Idea 2

In `plotOptions.series.dragDrog`, allow passing in an `isDraggable` function with the signature `(point: Highcharts.Point) => boolean`. Internally, Highcharts would call this function for each point to determine if it is draggable.

Highcharts, draggable-points module.

N/A

----

Vazonik:
Added demo showing how to disable drag & drop for specific points. Closes #19366.

### DIFF
--- a/samples/highcharts/dragdrop/undraggable-points/demo.js
+++ b/samples/highcharts/dragdrop/undraggable-points/demo.js
@@ -25,14 +25,10 @@ Highcharts.chart('container', {
         point: {
             events: {
                 drag: function () {
-                    if (this.x === 2) {
-                        return false;
-                    }
+                    return this.className !== 'undraggable';
                 },
                 drop: function () {
-                    if (this.x === 2) {
-                        return false;
-                    }
+                    return this.className !== 'undraggable';
                 }
             }
         },
@@ -40,9 +36,7 @@ Highcharts.chart('container', {
         dataLabels: {
             enabled: true,
             y: -6,
-            formatter: function () {
-                return this.x === 2 ? 'undraggable' : '';
-            }
+            format: '{point.className}'
         },
 
         cursor: 'ns-resize',
@@ -53,7 +47,8 @@ Highcharts.chart('container', {
         }, {
             x: 2,
             y: 1,
-            color: '#f33'
+            color: '#f33',
+            className: 'undraggable'
         }, {
             x: 3,
             y: 25
--- a/samples/highcharts/dragdrop/undraggable-points/demo.js
+++ b/samples/highcharts/dragdrop/undraggable-points/demo.js
@@ -1,0 +1,62 @@
+Highcharts.chart('container', {
+    yAxis: {
+        softMin: -200,
+        softMax: 200
+    },
+
+    xAxis: {
+        min: 0.5,
+        max: 3.5
+    },
+
+    tooltip: {
+        enabled: false
+    },
+
+    title: {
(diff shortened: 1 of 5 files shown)


samples/highcharts/dragdrop/undraggable-points/demo.js
```diff
@@ -0,0 +1,62 @@
...
+                }
+            }
+        },
+
+        dataLabels: {
+            enabled: true,
+            y: -6,
+            formatter: function () {
+                return this.x === 2 ? 'undraggable' : '';
+            }
```
TorsteinHonsi, line 45: Prefer using `format` instead. To limit the data label to only the undraggable point, it can be defined directly on the `data` item. Alternatively use [templating](https://www.highcharts.com/docs/chart-concepts/templating).

samples/highcharts/dragdrop/undraggable-points/demo.js
```diff
@@ -0,0 +1,62 @@
...
+                drag: function () {
+                    if (this.x === 2) {
+                        return false;
+                    }
+                },
+                drop: function () {
+                    if (this.x === 2) {
+                        return false;
+                    }
+                }
```
TorsteinHonsi, line 36: Checking for the `x` value it totally valid, but may not be practical. What if we want multiple points to be undraggable? What if the points also are `draggableX` so that x is not known?

I think it would be better and more extendable to mark the point with either a custom option or a `className`. 

```js
{
            x: 2,
            y: 1,
            color: '#f33',
            className: 'undraggable'
        }, 
```

TorsteinHonsi (changes requested):
Thanks! Some change requests added.

TorsteinHonsi (approved):

PR #19410, opened 2023-07-20
bugfix/19341-boost-tooltip-shows-null-value

### LINKED ISSUE #19341: Boost: Tooltip shows null value
#### Expected behaviour
Tooltip: In normal mode, only values != null are displayed.

#### Actual behaviour
The boost mode does show the tooltip also when the value is null.
This was not the case in version 10.3.3 so the behaviour has changed.

https://jsfiddle.net/BlackLabel/wgujxm7k/

#### Product version
v11.0.0

Internal note:
Bisected to https://github.com/highcharts/highcharts/commit/f0977cef5502272a87dca24ee230d56b92020f5f

Vazonik:
Fixed #19341, null points could be hovered in the boost module.
___
_**Internal note:** This bug was introduced in #18551. I can't see any reason why null points should also be added to KD-tree. In the case of https://github.com/highcharts/highcharts/pull/18551#issuecomment-1439109364, the number of points in a series can be checked by referring to e.g. `series.processedXData`; point index also from this, or e.g. from the `point.i` parameter._

### DIFF
--- a/samples/unit-tests/boost/general/demo.js
+++ b/samples/unit-tests/boost/general/demo.js
@@ -166,8 +166,8 @@ QUnit[Highcharts.hasWebGLSupport() ? 'test' : 'skip'](
 
         assert.strictEqual(
             chart.series[3].points.length,
-            4,
-            'array length should include values with null'
+            2,
+            'Null points should not be added to the series\' kd-tree (#19341)'
         );
     }
 );
--- a/ts/Extensions/Boost/BoostSeries.ts
+++ b/ts/Extensions/Boost/BoostSeries.ts
@@ -1055,7 +1055,7 @@ function seriesRenderCanvas(this: Series): void {
                 isYInside = (y || 0) >= yMin && y <= yMax;
             }
 
-            if (x >= xMin && x <= xMax && isYInside) {
+            if (y !== null && x >= xMin && x <= xMax && isYInside) {
 
                 clientX = xAxis.toPixels(x, true);
 


Vazonik:
> Does this mean that #18544 is effectively reintroduced? I am not sure why it was perceived as a bug, but someone reported it so if it is reintroduced it will probably create problems.

I'm pretty sure that #18544 was incorrectly identified as a bug. It was reported and _fixed_ by the same person. I agree with @pawellysy comment (https://github.com/highcharts/highcharts/pull/18551#pullrequestreview-1307813744) and I don't agree with the only argumentation (https://github.com/highcharts/highcharts/pull/18551#issuecomment-1439109364) why it would be a bug, it's rather due to the specificity of Boost and it is not at all problematic to get the number of points on the chart and their ID.

> If so, it is possible to separate the two, so that null points are parts of the `series.points` array, but not part of the k-d-tree?

The `series.points` array in the boost is of the `BoostPointMockup` type, which is closely related to the position of the point in pixels and added only when the point is in the range of extremes and in the appropriate position in pixel units. I don't think it would be too problematic to allow the `BoostPointMockup.plotY` parameter to be `null` or `undefined`, but that would require some changes, I think, against the purpose of the array. Especially since in the Boost, we have access to `series.processedXData` and `series.processedYData`.

But of course if you @TorsteinHonsi  feel it needs to be rebuilt I can implement it.

TorsteinHonsi (approved):
> I'm pretty sure that https://github.com/highcharts/highcharts/issues/18544 was incorrectly identified as a bug

Okay, let's merge this as it is then.

PR #19411, opened 2023-07-20
demo/arearange update

andreastistel:
Added color gradient to the arearange chart. Goes from blue to red based on temp value
[Asana task](https://app.asana.com/0/1203542399740129/1203553273454181/f)


TorsteinHonsi (approved):
Thanks!

PR #19415, opened 2023-07-21
demo/dynamic-update Updated

andreastistel:
Changed to non-default colors and tried to add pulsing animation to the random data points.
This was done with help from ChatGPT.
May need to change the colors to something more appropriate. 

[Asana task](https://app.asana.com/0/1203542399740129/1203712551516194/f)


TorsteinHonsi (approved):
Thanks!

PR #19416, opened 2023-07-21
bugfix/19412-reset-zoom-button

### LINKED ISSUE #19412: Reset Zoom Button position settings are overwritten by the deprecated parameter
#### Expected behaviour
In Highcharts.
According to documentation the position of Reset Zoom Button should be set as follows:
`chart: {
    zooming: {
         resetButton: {
               position: {
                   x: [customXValue],
                   y: [customYValue]
               }
`

#### Actual behaviour
The values above are overwritten by default values from now deprecated parameter: 
`chart: {
      resetZoomButton: {
           position: {
.....
      }             
`
In the source code this happens in the private function **setZoomOptions()**, as below:
`resetButton: merge(zooming.resetButton, options.resetZoomButton)`

The workaround is passing the following in the options:
`resetZoomButton: null`

I don't have a live demo.

#### Product version
Works correctly in: v10.3.3
Works **Incorrectly** in: v11.1.0

#### Affected browser(s)
This is a general behaviour

jakubSzuminski:
Fixed #19412, `chart.zooming.resetZoom` position settings were overwritten by the deprecated parameter.

~~Fixed #19412, setting chart.zooming.resetZoom now works.~~

**Demo after fix:** https://jsfiddle.net/BlackLabel/krx3jys0/

**Source of the problem:**
This happened because we're merging `merge(chart.options.zooming.resetZoom, chart.options.resetZoomButton)` while the defaults for the `resetZoomButton` are still there so any option we set in `zooming.resetZoom` is overriden by the default one of the deprecated option.

### DIFF
--- a/ts/Core/Chart/ChartDefaults.ts
+++ b/ts/Core/Chart/ChartDefaults.ts
@@ -619,7 +619,8 @@ const ChartDefaults: ChartOptions = {
 
     /**
      * The button that appears after a selection zoom, allowing the user
-     * to reset zoom.
+     * to reset zoom. This option is deprecated in favor of
+     * [zooming](#chart.zooming).
      *
      * @since      2.2
      * @deprecated 10.2.1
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (approved):
Thanks!

PR #19419, opened 2023-07-24
demo/line-basic renamed to line-chart

andreastistel:
Renamed from line-basic to line-chart. Also changed name in all other places I could find.
[Asana task](https://app.asana.com/0/1203542399740129/1203542399740132/f)


TorsteinHonsi (approved):
Thanks!

PR #19428, opened 2023-07-25
demo/3d-pie update

andreastistel:
Placed datalabels on top of slices. Made fontsize a bit bigger.
[Asana task](https://app.asana.com/0/1203542399740129/1203712551516239/f)


TorsteinHonsi:
Closing this, as there turns out to be an issue with label placement on 3d pies. When setting the `distance` to a negative number, the labels are not always on top of the slices. Using individual adjustments for this in a demo is bad practice. Issue recorded here: https://github.com/highcharts/highcharts/issues/19615

PR #19429, opened 2023-07-25
API: Fixed vector plotOptions.

bre1470:


TorsteinHonsi (changes requested):
To avoid the `any` cast, can we make it `void 0` like [we do with the tree map](https://github.com/highcharts/highcharts/blob/2435a8eeeca4574068d0ecf9339463454f5b5f24/ts/Series/Treemap/TreemapSeries.ts#L246)?

TorsteinHonsi (approved):
Thanks!

PR #19430, opened 2023-07-25
Dash/datagrid sync visibility

pawellysy:
Implemented visibility sync handler.

### DIFF
--- a/samples/dashboards/cypress/component-datagrid/demo.mjs
+++ b/samples/dashboards/cypress/component-datagrid/demo.mjs
@@ -47,6 +47,7 @@ Dashboards.board('container', {
             type: 'Highcharts',
             sync: {
                 highlight: true,
+                visibility: true,
                 extremes: true
             },
             columnAssignment: {
@@ -82,6 +83,7 @@ Dashboards.board('container', {
             editable: true,
             sync: {
                 highlight: true,
+                visibility: true,
                 extremes: true
             },
             visibleColumns: ['Food', 'Vitamin A']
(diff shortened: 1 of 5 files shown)


samples/dashboards/cypress/component-datagrid/demo.mjs
```diff
@@ -82,6 +83,7 @@ Dashboards.board('container', {
             editable: true,
             sync: {
                 highlight: true,
+                visibility: true,
```
sebastianbochan, line 86: In HC we use `visible` so I would expect to have the same name here.

ts/Dashboards/Plugins/DataGridSyncHandlers.ts
```diff
@@ -210,13 +210,72 @@ const configs: {
                 this.on('setConnector', (): void => unregisterCursorListeners());
                 this.on('afterSetConnector', (): void => registerCursorListeners());
             }
+        },
+        visibilityHandler: function(this: DataGridComponent): void {
+            const component = this;
+            const { board } = this;
```
sebastianbochan, line 216: Would be better to use component here.

`const { board } = component;`

It will be more clearful.

samples/dashboards/cypress/component-datagrid/demo.mjs
karolkolodziej: Everywhere in Dash, we use `visibility`, @pawelfus should we change that? 
IMO the `visible` might be confusing because it looks like the `sync` might or might not be visible.

karolkolodziej (approved):

PR #19433, opened 2023-07-25
demo/scatter-boost update

andreastistel:
Modified the for-loop so that the datapoints makes a tangent wave pattern.
[Asana task](https://app.asana.com/0/1203542399740129/1203673412610654/f)


TorsteinHonsi (approved):
Thanks!

PR #19436, opened 2023-07-25
bugfix/17988-dataLabels-type

### LINKED ISSUE #17988: Docs: series.scatter.data.dataLabels is declared as object but also accepts an array
#### Expected behaviour

As mentioned the input is the same as for plotOptions.series.dataLabels. But the "{" in the properties overview indicates that for series.scatter.data.dataLabels only a single object specifying a single label is allowed.

#### Actual behaviour

series.scatter.data.dataLabels is displayed as "[{" in the properties menu.

#### Product version
10

image

magdalena-gut:
Fixed #17988, wrong `data.dataLabels` type in API.


bre1470 (approved):
Thank you! Looks good. 👍

PR #19437, opened 2023-07-25
Demos: Increased defer in network graph.

pawelfus:


TorsteinHonsi (approved):
Thanks! However I think the delay was too great now, it is okay to show the label before the simulation has settled, just not interfere with the initial explosion.

PR #19438, opened 2023-07-25
bugfix/19190-map-drag-NaN

### LINKED ISSUE #19190: Highcharts maps demo interaction: Map inverts after zooming and dragging.
#### Expected behaviour
After zooming and dragging the map to the top. The map stops at its border.

#### Actual behaviour
After zooming and dragging the map to the top. The map inverts and gets stuck to the top-left corner of the demo. This happens on all the demos with the same map type.

https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/maps/demo/doubleclickzoomto

Video:
https://github.com/highcharts/highcharts/assets/137054126/8855e9b9-b293-4eb3-9de6-42584bcab37f

#### Product version
Highcharts v11.1.0
Highmaps  v11.1.0

#### Affected browser(s)
Tested on Firefox and Chrome.

KacperMadej:
Fixed #19190, prevented map display error on drag out of valid coordinates.

---
  - https://app.asana.com/0/0/1204915984120491

### DIFF
--- a/samples/unit-tests/maps/map-navigation/demo.html
+++ b/samples/unit-tests/maps/map-navigation/demo.html
@@ -1,6 +1,7 @@
 <script src="http://cdnjs.cloudflare.com/ajax/libs/proj4js/2.3.6/proj4.js"></script>
 <script src="https://code.highcharts.com/maps/highmaps.js"></script>
 <script src="https://code.highcharts.com/mapdata/countries/gb/gb-all.js"></script>
+<script src="https://code.highcharts.com/mapdata/custom/world.js"></script>
 
 <div id="qunit"></div>
 <div id="qunit-fixture"></div>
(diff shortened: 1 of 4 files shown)


samples/unit-tests/maps/map-navigation/demo.html
```diff
@@ -1,6 +1,7 @@
 <script src="http://cdnjs.cloudflare.com/ajax/libs/proj4js/2.3.6/proj4.js"></script>
 <script src="https://code.highcharts.com/maps/highmaps.js"></script>
 <script src="https://code.highcharts.com/mapdata/countries/gb/gb-all.js"></script>
+<script src="https://code.highcharts.com/mapdata/custom/world.js"></script>
```
hubertkozik, line 4: I think we don't need to load those maps now, after using simplified topoJSON in `json-sources.js` file.

hubertkozik (changes requested):

samples/unit-tests/maps/map-navigation/demo.html
KacperMadej: True - removed.
Just to clarify, the simplified TopoJSON from `json-srouces.js` is used for automated karma testing while the scripts in `demo.html` are used for utils in-browser testing. None of them uses the maps in `.js` format because TopoJSON maps are used instead. Mapped locally to `json-soruces.js` (karma) or fetched from CDN (utils).

hubertkozik (approved):

TorsteinHonsi (approved):
Thanks!

PR #19440, opened 2023-07-25
bugfix/19426-extending-default-formatter-demo

### LINKED ISSUE #19426: "Extending default formatter" sample code doesn't work
#### Expected behaviour
Sample code does what it says it does

#### Actual behaviour
Sample code does nothing

https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/tooltip/formatter-conditional-default/
1. Add `console.log(this.point)`
2. Note that `isNull` is never true

#### Fix:
https://github.com/magnusanderson-wk/highcharts

#### Product version
highcharts 11.1.0

#### Affected browser(s)
All?

Vazonik:
Fixed #19426, extending default formatter demo.

### DIFF
--- a/samples/highcharts/tooltip/formatter-conditional-default/demo.js
+++ b/samples/highcharts/tooltip/formatter-conditional-default/demo.js
@@ -24,7 +24,7 @@ Highcharts.chart('container', {
 
     tooltip: {
         formatter: function (tooltip) {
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (approved):
Thanks!

PR #19441, opened 2023-07-25
Connector with modifier options

bre1470:
- Moved async modifier attachment from pool to connectors.
- Updated climate demo.
- Updated doclets.
- Removed API duplicates


samples/dashboards/demo/climate/demo.js
```diff
@@ -977,20 +976,29 @@ async function updateBoard(board, city, column, scale, newData) {
...
     const timeRangeMax = timeRangeSelector.chart.axes[0].max;
     const timeRangeMin = timeRangeSelector.chart.axes[0].min;
-    await selectionTable.setModifier(new RangeModifier({
-        modifier: 'Range',
-        ranges: [{
+    const selectionModifier = selectionTable.getModifier();
+
+    if (
+        !selectionModifier ||
+        selectionModifier.options.ranges[0].maxValue !== timeRangeMax ||
```
sebastianbochan, line 991: Shouldn't we check that `selectionModifier.options` exist?

sebastianbochan (approved):

samples/dashboards/demo/climate/demo.js
bre1470: `DataModifier.options` always exists. `ranges` is a default. The only thing that could fail is accessing index 0, but we know that we have set it.

samples/dashboards/demo/climate/demo.js
```diff
@@ -977,20 +976,29 @@ async function updateBoard(board, city, column, scale, newData) {
...
-    selectionTable.setColumns(cityTable.modified.getColumns(), 0);
     const timeRangeMax = timeRangeSelector.chart.axes[0].max;
     const timeRangeMin = timeRangeSelector.chart.axes[0].min;
-    await selectionTable.setModifier(new RangeModifier({
-        modifier: 'Range',
-        ranges: [{
+    const selectionModifier = selectionTable.getModifier();
+
+    if (
+        !selectionModifier ||
```
karolkolodziej, line 990: image
I get that so i presume the condition `!selectionModifier ||` isn't ideal

karolkolodziej (changes requested):

pawellysy (approved):
Looks good! 👍

samples/dashboards/demo/climate/demo.js
```diff
@@ -965,6 +969,7 @@ async function updateBoard(board, city, column, scale, newData) {
         selectionGrid,
         cityChart
     ] = board.mountedComponents.map(c => c.component);
+    console.log(await dataPool.getConnector('Range Selection'));
```
karolkolodziej, line 972: do we need that?

karolkolodziej (approved):

samples/dashboards/demo/climate/demo.js
bre1470: No 🙂

PR #19448, opened 2023-07-26
dash/edit-mode-design

karolkolodziej:
Improved the edit mode sidebars and popup design.

Waiting for confirmation on how the layout should be styled, which will be in a separate PR.


tools/gulptasks/dashboards/scripts.js
```diff
@@ -70,7 +70,7 @@ async function dashboardsScripts() {
                 ),
                 namespace: 'Dashboards',
                 output: bundleTargetFolder,
-                version: (argv.release || '0.9.9')
```
sebastianbochan, line 73: Which version should be the default? Should we update it each time?

sebastianbochan (approved):

tools/gulptasks/dashboards/scripts.js
karolkolodziej: I've added this only to test the designs with the proper icons. 
Finally, it is going to be replaced in the `highcharts-utils` we have dedicated issue in the Assana for that.

pawelfus (changes requested):
There are a few missing changes (number in brackets reflects number on screenshots):
- Hamburger menu:
  - wrong toggle color (1)
- Edit Cell/Row:
  - wrong handles (2)
  - wrong  background pattern (3)
- Sidebar:
  - wrong input/dropdown backgrounds (4)
  - wrong toggle button colors (5)
  - wrong overlay background (6)
  - increase the Confirm/Cancel buttons corners (radius) (7)
  - replace triangles with arrows (8)
- Adding a new component:
  - The implementation is missing the transition as described in [Asana](https://app.asana.com/0/1203234853779882/1204307015727101/f) 
  - Could we also animate the hiding of the AddComponent sidebar? It vanishes right now


Zrzut ekranu 2023-07-28 o 11 15 03

Zrzut ekranu 2023-07-28 o 11 12 40
Zrzut ekranu 2023-07-28 o 11 10 16

samples/dashboards/demo/minimal/demo.js
```diff
@@ -136,4 +136,6 @@ Dashboards.board('container', {
             highlight: true
         }
     }]
-}, true);
+}, true).then(dashboard => {
+    dashboard.editMode.activate();
```
pawelfus, line 140: Cleanup needed ;)

gfx/dashboards-icons/resize-handle.svg
pawelfus, line 1: Q: It looks a bit strange 🤔 Do you know by any chance if is it because of the GitHub review tool?

pawelfus (changes requested):
Looks great, thank you!

In addition to inline comments:

1. Could you try a white stroke instead of black? I think it will be more visible:
image

gfx/dashboards-icons/resize-handle.svg
karolkolodziej: Hmm you are right. It looks a bit strange but I have no idea why it might look like this 🤷‍♂️

pawelfus (approved):
Looks good, thanks!

(Approving, but I support Sebastian's suggestions)

pawellysy (approved):
LGTM

karolkolodziej:
@sebastianbochan 
1. Let's leave it as it is, I added a separate task to Assana about `scaleElements` to discuss what to do with it.
2. That is how it was designed, let's leave it.
3. Label was added but there are still some other issues that I reported in Assana. Please resolve this in a separate PR.
4. FIxed.
5. Fixed.

sebastianbochan:
I'm not convinced we should leave bugs like in point 1 or 3.

sebastianbochan (changes requested):
I'm not convinced that we should leave bugs like in point 1 or 3.
@pawelfus @oysteinmoseng

LB0ge (approved):

sebastianbochan (approved):

PR #19450, opened 2023-07-26
bugfix/16445-gantt-ticks-interval

### LINKED ISSUE #16445: Gantt: Correct ticks not showing for set of data
#### Expected behaviour
The ticks on both axes should change their interval to weeks/days etc

#### Actual behaviour
The ticks of the xAxes are constant, and don't scale

https://jsfiddle.net/BlackLabel/tx6430e5/3/

#### Product version
Highcharts Gantt v9.2.2.
#### Affected browser(s)
all

#### Example workaround

```js
(function(H) {
  H.wrap(H.Series.prototype, 'processData', function(proceed, points) {
    proceed.call(this);
    this.closestPointRange = this.basePointRange = 259200000;
  });
}(Highcharts));
```

demo with a workaround: 
https://jsfiddle.net/BlackLabel/tx6430e5/4/

magdalena-gut:
Fixed  #16445, allow putting ticks closer than their data points.


ts/Core/Axis/Axis.ts
```diff
@@ -1905,6 +1905,7 @@ class Axis {
             options.minTickInterval,
             // In datetime axes, don't go below the data interval, except when
             // there are scatter-like series involved (#13369).
+            !axis.options.grid &&
```
TorsteinHonsi, line 1908: Instead of creating an exception for the grid axis in the core, can we just set the default `minTickInterval` to 0 or 1 in the [grid axis defaults](https://github.com/highcharts/highcharts/blob/815d2def69770e22827569c6d5a65cd606390863/ts/Core/Axis/GridAxis.ts#L173-L200)?

It seems to work as intended when setting the option explicitly: https://jsfiddle.net/2r83tpuo/

TorsteinHonsi (changes requested):

TorsteinHonsi (approved):

PR #19451, opened 2023-07-26
bugfix/19362-xrange-tooltip-default-formatting

### LINKED ISSUE #19362: X-range: Tooltip loses default data formatting
#### Expected behaviour
After adding [API-compliant](https://api.highcharts.com/highcharts/series.xrange.data) properties like `name` the tooltip should not lose its default datatime formatting.

#### Actual behaviour
As you can see in the demo below, for data with a `name` set, the tooltip does not recognize `datatime` formatting, and for data without `name`, it formats the data correctly. It would be necessary to standardize this behavior also for the axis standard data type in the X-range series.

https://jsfiddle.net/BlackLabel/ohxp34qt/

#### Product version
7.0.0

madepiet:
Fixed #19362, x-range series tooltip lost its correct default date and time formatting.

### DIFF
--- a/samples/unit-tests/series-xrange/xrange/demo.js
+++ b/samples/unit-tests/series-xrange/xrange/demo.js
@@ -331,9 +331,9 @@ QUnit.test('Partial fill reversed', assert => {
 
     assert.close(
         chart.series[0].points[0].graphic.rect.attr('x') +
-            chart.series[0].points[0].graphic.rect.attr('width'),
+        chart.series[0].points[0].graphic.rect.attr('width'),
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (changes requested):
Thanks!

The code is good, just a little work is needed on the changelog text:

"Fixed #19362, X-range series tooltip retains the correct default formatting."

https://github.com/highcharts/highcharts/tree/master/changelog#writing-content-for-the-changelog

Specifically point 6.

TorsteinHonsi (approved):

PR #19453, opened 2023-07-27
bugfix/18815-one-series-boost-threshold-works-for-all

### LINKED ISSUE #18815: `boostThreshold` for one series makes all series boosted.
#### Expected behaviour
The `boostThreshold` option enables boost only for series that have more or equal points amount than the threshold, and series that have less are displayed by default.

#### Actual behaviour
If there is a series that has more points than the set `boostThreshold`, all series are displayed in boost mode until one series is toggled off and on, using the legend.

https://jsfiddle.net/BlackLabel/3beqs7ch/

1. *All series are displayed in boost mode.*
2. Double-click any legend item twice to disable and enable a series.
3. *The scatter series is displayed as it should have been from the beginning.*

#### Product version
Highcharts v10.3.3

Vazonik:
Fixed #18815, `boostThreshold` set for one series boosted all series.

### DIFF
--- a/samples/unit-tests/boost/general/demo.js
+++ b/samples/unit-tests/boost/general/demo.js
@@ -331,3 +331,45 @@ QUnit[Highcharts.hasWebGLSupport() ? 'test' : 'skip'](
         );
     }
 );
+
+QUnit[Highcharts.hasWebGLSupport() ? 'test' : 'skip'](
+    'Boost handling of individual boostable series',
+    function (assert) {
+        const chart = Highcharts.chart('container', {
+            plotOptions: {
+                series: {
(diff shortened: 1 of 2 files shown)


TorsteinHonsi:
@cvasseng Do you remember if there was a good and deliberate reason for making one boost-series force all others to boost? I feel there was, but I can't remember the scenario.

cvasseng:
I can't totally recall either. 🤔 may have been related to weirdness when combining boosted and not boosted series, which may or may not still be a thing.

TorsteinHonsi (approved):
Okay, let's do this then.

PR #19456, opened 2023-07-27
dash/datacursor-group

bre1470:


samples/highcharts/data-tools/datacursor-group/demo.css
```diff
@@ -0,0 +1,38 @@
+#wrapper {
+    display: flex;
+    justify-content: stretch;
```
karolkolodziej, line 3: Can we please add for example `gap: 20px`?
At first I though it is one big table ;)

samples/highcharts/data-tools/datacursor-group/demo.js
```diff
@@ -0,0 +1,148 @@
...
+// Main table with cursors
+const table1 = new Highcharts.DataTable({
+    columns: {
+        Rank: [1, 2, 3, 4, 5, 6],
+        City: ['Sydney', 'Melbourne', 'Brisbane', 'Perth', 'Adelaide', 'Canberra'],
+        Population: [5259764, 4976157, 2568927, 2192229, 1402393, 453558]
+    }
+});
+// Resorted tables
+const sort2 = new Highcharts.DataModifier.types.Sort({
```
karolkolodziej, line 13: Wouldn't it make more sense if we have 3 different tables with the same column names but different values of Population? 
Each table might represent one year. IMO this might be more natural than sorted values.
Then instead of the `Rank` We can look for the specific city?

karolkolodziej (changes requested):

samples/highcharts/data-tools/datacursor-group/demo.js
bre1470: If we talk about sense then we should have something that is applicable to the group concept. So some other demo data is needed. Though I think the time is spent better somewhere else because this is an internal demo for the technical part.

karolkolodziej: Sure!

karolkolodziej (approved):

samples/highcharts/data-tools/datacursor-group/demo.js
bre1470: But I agree with your comment. It is not ideal.

sebastianbochan (approved):

pawelfus (approved):
Thanks!

pawellysy (approved):
LGTM!

PR #19464, opened 2023-07-28
feature/19367-fibonacci-annotation-reverse

### LINKED ISSUE #19367: navigation.bindings.fibonacci calculation vs Annotation module sample
both calculation should be consistent. if you try the annotation module example, it uses end-start as difference (which is correct) navigation.bindings.fibonacci (Fibonacci class) should use the same formula.

Actual behavior: in stockTools' GU if you draw Fibonacci retracement --> depending on the direction it uses 0(0%) for the first click point(first candle) which is not correct.

[https://jsfiddle.net/sLn4hz1j/4/](url)

steps:
1. draw the retracement from some where from top of the chart to bottom..(opposite direction of existing one diagonal line)
2. draw from bottom to top.. 
in both direction first candle(first point) should be at 1 (100%)

version --
11.1

jakubSzuminski:
Added `fibonacci.reverse` option. Closes #19367.

- [x] Add a `fibonacci.reverse` option
- [x] Improve the docs about `fibonacci` and this option (`navigation.bindings.fibonacci.annotationsOptions.typeOptions.reverse`) 
- [x] Add a visual test


TorsteinHonsi:
The convention says `reversed`

Skjermbilde 2023-08-21 kl  08 52 01

On the rationale that _reverse_ is a verb, but as an option it is more natural to use the noun, _reversed_. I realize we don't follow this principle 100%, but in this case we are already using the same word for other options, so then we should stick to one form.

PR #19471, opened 2023-07-31
dash/datagrid-dragging-point

sebastianbochan:
Fixed missing reference to DOM element in the dataGrid, after dragging a point.

-----

- [x] - bugfix
- [x] - test

### DIFF
--- a/test/cypress/integration/Dashboards/Components/DataGrid/DataGridComponent.cy.js
+++ b/test/cypress/integration/Dashboards/Components/DataGrid/DataGridComponent.cy.js
@@ -65,13 +65,11 @@ describe('layout resize on window changes', () => {
             .trigger('mousemove', 300, 100)
             .trigger('mouseup');
 
-
         cy.get('.highcharts-datagrid-outer-container').then($container =>{
             assert.ok(
                 $container.scrollTop() > containerTop,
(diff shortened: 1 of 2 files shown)


test/cypress/integration/Dashboards/Components/DataGrid/DataGridComponent.cy.js
```diff
@@ -81,4 +81,32 @@ describe('layout resize on window changes', () => {
         cy.get('.highcharts-legend-item').eq(0).click();
         cy.get('.highcharts-datagrid-column-headers').children().should('have.length', 2);
     });
+
+    it('DataGridComponent should not throw error when dragging point on chart.', () => {
+        cy.chart().then(chart => {
```
karolkolodziej, line 86: Why don't you use the `board.mountedComponents` instead of the `chart` you get there?

karolkolodziej (changes requested):

test/cypress/integration/Dashboards/Components/DataGrid/DataGridComponent.cy.js
sebastianbochan: Nice catch! thanks!

PR #19472, opened 2023-07-31
bugfix/19373/map-drilldown-animation-false

### LINKED ISSUE #19373: Drilldown.animation = false causes Uncaught RangeError: Maximum call stack size exceeded
#### Expected behaviour
When setting the drilldown.animation to 'false', the chart should function as expected when user attempts to drilldown. As described: https://api.highcharts.com/highmaps/drilldown 

#### Actual behaviour
The chart hangs for a few seconds and then an error is logged to console: Uncaught RangeError: Maximum call stack size exceeded

https://jsfiddle.net/f1wv8ys3/6/

Notice line 139. 
* Open console 
* Click on a state in the chart
* Wait for error in console 
* Change line 139 to be true
* Save/run
* Click on a state in the chart
* Observe glorious drill-down

#### Product version

I'm seeing this in 11.1.0, the only version I've tried. 

#### Affected browser(s)
All that I've tried, which is just Edge.

hubertkozik:
Fixed #19373, map drilldown animation was impossible to disable.

---
  - https://app.asana.com/0/0/1205072268306489

### DIFF
--- a/samples/unit-tests/maps/drilldown/demo.js
+++ b/samples/unit-tests/maps/drilldown/demo.js
@@ -14,7 +14,7 @@ QUnit.test('Map drilldown animation', assert => {
             germanyTopo = JSON.parse(germanyTopojsonText),
             europeGeo = JSON.parse(europeGeojsonText),
             germanyGeo = JSON.parse(germanyGeojsonText),
-            duration = 400,
+            duration = 350,
             chart = Highcharts.mapChart('container', {
                 chart: {
                     animation: {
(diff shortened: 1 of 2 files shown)


MarkusBarstad (approved):

KacperMadej (approved):
The relevant problem looks resolved 👍 

Could you check if the duration of the unit-test could be reduced?
There's `duration = 400` but the tests are failing in browser when reduced to e.g. `40` or even `200`.

TorsteinHonsi (approved):
Thanks!

PR #19474, opened 2023-08-01
dash/hccColumnAssignment-nulls

sebastianbochan:
Replaced blacklist with whitelist in columnAssignment option.

--- 

- [x] - fix
- [x] - update demos

### DIFF
--- a/samples/dashboards/demo/climate/demo.js
+++ b/samples/dashboards/demo/climate/demo.js
@@ -789,19 +789,6 @@ async function setupBoard() {
             connector: {
                 id: 'Range Selection'
             },
-            columnAssignment: {
-                time: null,
-                FD: null,
-                ID: null,
-                RR1: null,
-                TN: null,
-                TX: null,
-                TNC: null,
-                TNF: null,
-                TXC: null,
-                TXF: null,
-                Date: null
-            },
             sync: {
                 highlight: true
             },
@@ -1077,16 +1064,13 @@ async function updateBoard(board, city, column, scale, newData) {
             },
             columnAssignment: {
                 time: 'x',
-                FD: column === 'FD' ? 'y' : null,
-                ID: column === 'ID' ? 'y' : null,
-                RR1: column === 'RR1' ? 'y' : null,
-                TN: null,
-                TNC: column === 'TNC' ? 'y' : null,
-                TNF: column === 'TNF' ? 'y' : null,
-                TX: null,
-                TXC: column === 'TXC' ? 'y' : null,
-                TXF: column === 'TXF' ? 'y' : null,
-                Date: null
+                FD: column === 'FD' ? 'y' : void 0,
+                ID: column === 'ID' ? 'y' : void 0,
+                RR1: column === 'RR1' ? 'y' : void 0,
+                TNC: column === 'TNC' ? 'y' : void 0,
+                TNF: column === 'TNF' ? 'y' : void 0,
+                TXC: column === 'TXC' ? 'y' : void 0,
+                TXF: column === 'TXF' ? 'y' : void 0
             }
         });
     }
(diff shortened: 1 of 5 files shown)


samples/dashboards/demo/climate/demo.js
```diff
@@ -1064,13 +1065,13 @@ async function updateBoard(board, city, column, scale, newData) {
...
-                TNF: column === 'TNF' ? 'y' : void 0,
-                TXC: column === 'TXC' ? 'y' : void 0,
-                TXF: column === 'TXF' ? 'y' : void 0
+                FD: column === 'FD' && 'y',
+                ID: column === 'ID' && 'y',
+                RR1: column === 'RR1' && 'y',
+                TNC: column === 'TNC' && 'y',
+                TNF: column === 'TNF' && 'y',
+                TXC: column === 'TXC' && 'y',
+                TXF: column === 'TXF' && 'y'
```
bre1470, line 1074: This is not type safe. You should settle on something regarding the negative flag. At the moment it is `null`, previously it was `undefined`, here it is `false`.

samples/dashboards/demo/climate/demo.js
```diff
@@ -1083,13 +1084,13 @@ async function updateBoard(board, city, column, scale, newData) {
...
-            TNF: column === 'TNF' ? 'y' : void 0,
-            TXC: column === 'TXC' ? 'y' : void 0,
-            TXF: column === 'TXF' ? 'y' : void 0
+            FD: column === 'FD' && 'y',
+            ID: column === 'ID' && 'y',
+            RR1: column === 'RR1' && 'y',
+            TNC: column === 'TNC' && 'y',
+            TNF: column === 'TNF' && 'y',
+            TXC: column === 'TXC' && 'y',
+            TXF: column === 'TXF' && 'y'
```
bre1470, line 1093: Same issue here.

ts/Dashboards/Plugins/HighchartsComponent.ts
```diff
@@ -283,7 +284,7 @@ class HighchartsComponent extends Component {
                         'chartConfig'
                     ]
                 }),
-            columnAssignment: {}
+            columnAssignment: null
```
bre1470, line 287: This is not type safe as well. Use `void 0` or remove it if possible.

ts/Dashboards/Plugins/HighchartsComponent.ts
```diff
@@ -618,11 +619,11 @@ class HighchartsComponent extends Component {
...
                         true;
 
-                    if (!isVisible && !columnAssignment[name]) {
-                        return false;
+                    if (!defined(this.options.columnAssignment)) {
+                        return true;
                     }
 
-                    if (columnAssignment[name] === null) {
+                    if (!isVisible || !columnAssignment[name]) {
```
bre1470, line 626: This will have issues with empty strings.

bre1470 (changes requested):

karolkolodziej (changes requested):
Please update the API for `columnAssignment`

samples/dashboards/demo/climate/demo.js
sebastianbochan: You're right, I optimized it, too much. Let's leave nulls there.

samples/dashboards/demo/climate/demo.js
sebastianbochan: You're right, I optimized it, too much. Let's leave nulls there.

ts/Dashboards/Plugins/HighchartsComponent.ts
bre1470: I will create a ticket.

ts/Dashboards/Plugins/HighchartsComponent.ts
```diff
@@ -993,8 +994,8 @@ namespace HighchartsComponent {
          */
         chartID?: string;
         /**
-         * Names / aliases that should be mapped to xAxis values. You can use
-         * null to keep columns selectively out of the chart.
+         * Names / aliases that should be mapped to xAxis values.
```
karolkolodziej, line 997: Sorry for being picky but I would describe how it works. What is going to happen if you don't declare them etc.

karolkolodziej (changes requested):

ts/Dashboards/Plugins/HighchartsComponent.ts
sebastianbochan: I tested this scenario and all should work properly. If you have data or scenario when not, please let me know.

pawellysy (approved):
LGTM

karolkolodziej (approved):

PR #19452, opened 2023-07-27
CSV demo with modifiers

bre1470:

PR #19478, opened 2023-08-01
dash/fix-datapool

bre1470:
Fixed insufficient waiting for loading connectors on second request.

Discovered via #19452

### DIFF
--- a/ts/Data/DataPool.ts
+++ b/ts/Data/DataPool.ts
@@ -13,11 +13,13 @@
 
 'use strict';
 
+
 /* *
  *
  *  Imports
  *
  * */
+
 
 import type DataEvent from './DataEvent';
 import type {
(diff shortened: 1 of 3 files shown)


pawellysy (approved):
Seems to work alright :D

ts/Data/DataPool.ts
```diff
@@ -120,20 +149,44 @@ class DataPool implements DataEvent.Emitter {
...
+            return this
+                .loadConnector(connectorOptions)
+                .then((connector): DataConnector => {
+                    delete this.waiting[name];
+
+                    window.setTimeout((): void => {
+                        for (let i = 0, iEnd = waiting.length; i < iEnd; ++i) {
+                            waiting[i](connector);
+                        }
+                    }, 1);
```
karolkolodziej, line 180: Do we need this 1ms here?

karolkolodziej (approved):

ts/Data/DataPool.ts
bre1470: Yes. :) So everything gets resolved in order.

PR #19484, opened 2023-08-02
bugfix/19465-numberformat-api-demo-typo

### LINKED ISSUE #19465: HighMaps - mappoint Series Tooltip Ignores numberFormat
#### Expected behaviour
Tooltip should obey numberFormat settings for mappoint series.

#### Actual behaviour
Tooltip does not obey numberFormat settingsmappoint series.

https://jsfiddle.net/wergeld/pyuhazs1/5/
Hover over the top-left most map marker for Reykjavik. The tooltip should show "10,000.000".
I have added setOptions and also explicitly set the the Highcharts.numberFormat on the tooltip formatter. In either case the desire to use the "," as the thousands separator is ignored.

This does work as expected for other chart series, just not mappoint.

#### Product version
HighMaps v11.1

#### Affected browser(s)
Chrome, Firefox, Edge.

jedrzejruta:
Fixed #19465, wrong property in `numberFormatter` API demo.

### DIFF
--- a/samples/highcharts/members/highcharts-numberformat/demo.js
+++ b/samples/highcharts/members/highcharts-numberformat/demo.js
@@ -16,7 +16,7 @@ var converters = {
 Highcharts.setOptions({
     lang: {
         decimalPoint: '\u066B',
-        thousandsSeparator: '\u066C'
+        thousandsSep: '\u066C'
     }
 });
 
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (approved):
Thanks!

PR #19485, opened 2023-08-02
bugfix/19457-multiple-dataLabels-enabled/disabled-threw-render-error

### LINKED ISSUE #19457: Error when first dataLabel is disabled and second dataLabel is enabled
#### Expected behaviour
The second dataLabel from config should work fine or at least no error in the console occurs (chart should be rendered correctly).

#### Actual behaviour
When disabling the first dataLabel and trying to enable the second one, the error in the console occurs and the chart isn't rendering.

Also, in Highcharts it is possible to define multiple dataLabels (especially in multiple values series like columnrange), but the API suggests that only an object is valid (it is possible to define multiple dataLabels in an array).

https://jsfiddle.net/BlackLabel/sz2e7q40/

#### Product version
Current is 11.1.0

Internal notes:
It looks simple at first glance thus tagging as good for beginners, but it needs to be debugged and investigated first.

KamilKubik:
Fixed #19457, multiple `dataLabels` enabled/disabled threw render (console) error.

### DIFF
--- a/samples/unit-tests/series/datalabels/demo.js
+++ b/samples/unit-tests/series/datalabels/demo.js
@@ -116,9 +116,9 @@ QUnit.test('Top -90', function (assert) {
         Math.abs(
             chart.series[0].points[0].dataLabel.element.getBoundingClientRect()
                 .top -
-                chart.series[0].points[0].graphic.element
-                    .getBoundingClientRect()
-                    .top
+            chart.series[0].points[0].graphic.element
+                .getBoundingClientRect()
+                .top
         ) < 12,
         'Label is top aligned to element'
     );
@@ -126,9 +126,9 @@ QUnit.test('Top -90', function (assert) {
         Math.abs(
             chart.series[0].points[1].dataLabel.element.getBoundingClientRect()
                 .top -
-                chart.series[0].points[1].graphic.element
-                    .getBoundingClientRect()
-                    .top
+            chart.series[0].points[1].graphic.element
+                .getBoundingClientRect()
+                .top
         ) < 12,
         'Label is top aligned to element'
     );
@@ -202,9 +202,9 @@ QUnit.test('Bottom 90', function (assert) {
         Math.abs(
             chart.series[0].points[0].dataLabel.element.getBoundingClientRect()
                 .bottom -
-                chart.series[0].points[1].dataLabel.element
-                    .getBoundingClientRect()
-                    .bottom
+            chart.series[0].points[1].dataLabel.element
+                .getBoundingClientRect()
+                .bottom
         ) < 12,
         'Labels are equally bottom aligned'
     );
@@ -242,11 +242,11 @@ QUnit.test('Top 90', function (assert) {
                     .getBoundingClientRect()
                     .top
             ) -
-                Math.round(
-                    chart.series[0].points[0].graphic.element
-                        .getBoundingClientRect()
-                        .top
-                )
+            Math.round(
+                chart.series[0].points[0].graphic.element
+                    .getBoundingClientRect()
+                    .top
+            )
         ) < 12,
         'Label is top aligned to element'
     );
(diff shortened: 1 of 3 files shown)


samples/unit-tests/series/datalabels/demo.js
```diff
@@ -381,6 +386,19 @@ QUnit.test('Datalabels overlap in hidden series (#3866)', function (assert) {
...
+        'No console error should be displayed (#19457).'
+    );
+
+    assert.strictEqual(
+        chart.options.plotOptions.series.dataLabels.filter(({ enabled }) =>
+            enabled
+        ).length,
+        chart.series[0].points[0].dataLabels.length,
+        'Only the second point\'s data label should be rendered (#19457).'
+    );
```
hubertkozik, line 401: It would be better to move these tests to test, which is responsible for multiple data labels: `unit-tests/datalabels/multiple-labels`. There are already declared data labels as an array, so one update there should be fine.

hubertkozik (changes requested):
One thing about tests is below.

samples/unit-tests/series/datalabels/demo.js
KamilKubik: Agree, it's cleaner now as there's no need to add any extra chart config, thanks!

hubertkozik (approved):

ts/Core/Series/DataLabel.ts
```diff
@@ -814,10 +814,10 @@ namespace DataLabel {
                 // Destroy and remove the inactive ones
                 let j = dataLabels.length;
                 while (j--) {
-                    if (dataLabels[j].isActive) {
+                    if (dataLabels[j]?.isActive) {
```
TorsteinHonsi, line 817: According to TypeScript, `dataLabels` is `(SVGElement | SVGLabel)[]`. So somewhere the typing have failed us by allowing an undefined value in the array. Can we dig a bit deeper and find out why/where that happens? It may prevent future issues from the same cause.

TorsteinHonsi (changes requested):

ts/Core/Series/DataLabel.ts
KamilKubik: The problem here is that we map through the [pointOptions](https://github.com/highcharts/highcharts/blob/dcf0ac67b30ed690429aac019c5eceb3a1139aaf/ts/Core/Series/DataLabel.ts#L561). Then, we assign the `dataLabels` to an array by the mapping index, but only when the `dataLabel` is enabled. In effect, when e.g. the first `dataLabel` isn't enabled, and the second one is, we have an empty element inside the array, so the `dataLabels[j].isActive` breaks. On the other hand, as you mentioned, we assume that all elements are defined. 

What's more, I think we need to be aware of two cases, new labels, and the existing ones we want to update. Therefore, we should use the mapping indices we have now to properly add or update `dataLabels`, and create elements even if they aren't enabled. I think we need these only in the scope of mapping, and this [check](https://github.com/highcharts/highcharts/blob/dcf0ac67b30ed690429aac019c5eceb3a1139aaf/ts/Core/Series/DataLabel.ts#L815) can base on formatted (enabled only) `dataLabels`.

TorsteinHonsi: What if we have only one `dataLabel` setting, and that has `enabled: false`? In that case we don't create an element, right? And for a good reason - turning off the data labels will increase performance. So we need to consider allowing these undefined places in the `dataLabels` array. Like it currently works, just with correct typing?

TorsteinHonsi: Yes, I absolutely agree to this. So we allow `undefined` only in the local scope.

Now related to your latest commit, I think the `// Write back` block can be simplified. Setting `point.dataLabel` to `point.dataLabels[0]` should be save because it is either undefined or an element. For `point.dataLabels` we should be able to do something like this:

`point.dataLabels = dataLabels.filter(defined)`

Or alternatively do the handling further up, so that `dataLabels` is either undefined or filtered.

KamilKubik: Yes, I think everything is properly typed now. Thanks for the refactoring suggestion, I pushed the changes.

TorsteinHonsi (changes requested):
Thanks again! 

I did some simplification of the code - we already had this case where we spliced out inactive data labels from the array, so I am trying to combine that with the undefined ones.

I think there's a remaining problem though - when experimenting with the code, it seems that your changes to the unit test doesn't actually catch the issue. I had to go back to the OP demo to see if my changes were working. Can you take a look at that? It looks like the unit test doesn't actually have a disabled data label followed by an enabled one.

TorsteinHonsi (approved):
All right, thanks!

PR #19490, opened 2023-08-03
dash/demo-crossfilter

bre1470:
Added NavigatorComponent for crossfilter of a connector table.


goransle (approved):
Looks good!

samples/dashboards/demo/crossfilter/demo.js
```diff
@@ -0,0 +1,226 @@
...
+                    valueSuffix: '%'
+                },
+                yAxis: {
+                    visible: false
+                },
+                xAxis: {
+                    type: 'category',
+                    minRange: 0.8,
+                    labels: {
+                        rotation: 90
```
karolkolodziej, line 156: Is it just me or is it easier to tilt my head to the left instead of right? 😅

samples/dashboards/demo/crossfilter/demo.js
```diff
@@ -0,0 +1,226 @@
...
+        }, {
+            cell: 'Bottom',
+            type: 'DataGrid',
+            connector: {
+                id: 'Economy'
+            }
+        }]
+    });
+
+    console.log(board);
```
karolkolodziej, line 224: Remove

karolkolodziej (changes requested):
Few things I've spotted:
1. If you select too much of a range, the right chart is blank:
image
2. I don't see any reason for using a line series. What do Chad, Chile and China have in common?
image
3. Should the second chart (timeline) zoom in when we zoom the first one?
4. It might be useful to have panning enabled.

samples/dashboards/demo/crossfilter/demo.js
bre1470: For me it is easier as well, but then it becomes harder to read.

bre1470:
Thanks, @karolkolodziej .
1. Please create a bug report.
2. They share the planet. The lines are related to the economic categories.
3. Filters should be independent from each other.
4. What is the use-case?

pawelfus (changes requested):
The demo looks good, however, I didn't expect `setCrossfilter` method here.
Like when we use Navigator and sync x-extremes -> we use a built-in sync to match other components' extremes.

Let's consider this demo: https://www.highcharts.com/demo/highcharts/parallel-coordinates  - Imagine each vertical axis as a separate chart: x-value would be the category axis (years, miles , training time) and y-value would be a number of occurrences. 

Now, each of the charts has xAxis zoom enabled. When we select 2003-2004 range, other charts will display data relevant to that period. Next, when we select the other chart extremes, let's say miles, to be in range <8, 16>, then again, data will be filtered out.

karolkolodziej:
1. That is what I'm doing 😉 IMO `minRange` for the timeline should be less than 1.
2. Hmm, I disagree. Countries should not be connected by the line series. But I will leave this call to you.
3. Ok
4. With panning it is easier to navigate through the `timeline` series.

pawellysy (approved):
Looks good!

bre1470:
1. I think it is set that way, because otherwise it disappears sometimes.
2. I will look how I can come up with something else. Maybe crossfilter agains the name is not useful as a range.
4. Okay, I will activate it.

goransle (commented):
image
Titles in dark mode are a bit tough to read, not sure if that is covered by other PRs/tasks though.

Also the content of the navigator components seems to load a bit inconstant for me. The last two loads on first interaction, but the first one only sometimes loads
image

sebastianbochan (changes requested):
I have the same observations about navigators. When we run the demo, all of them are empty, moreover, the first navigator is always empty.
Screenshot 2023-08-18 at 11 25 15

The solutions is using `columnAssignment` with y, instead of custom name.

**Suggestion:**
```
columnAssignments: {
        Agriculture: 'y'
},
```

I am not pretty sure about the impact on other parts.

oysteinmoseng:
This demo fails for me - build issue?

Screen Shot 2023-08-19 at 13 12 47

Screen Shot 2023-08-19 at 13 18 22

bre1470:
I use `npm run dtest` as a shortcut.

sebastianbochan (approved):

sebastianbochan:
Looks good to me 👍

ts/Dashboards/Plugins/NavigatorComponent.ts
```diff
@@ -0,0 +1,690 @@
...
+
+    /* *
+     *
+     *  Functions
+     *
+     * */
+
+
+    /** @private */
+    private adjustNavigator(): void {
```
karolkolodziej, line 447: I'm unable to set the height of the chart by `chartOptions.chart.height`

ts/Dashboards/Plugins/NavigatorComponent.ts
```diff
@@ -0,0 +1,690 @@
...
+        for (let i = 0, iEnd = timeouts.length; i < iEnd; ++i) {
+            clearTimeout(timeouts[i]);
+        }
+
+        timeouts.length = 0;
+
+        timeouts.push(setTimeout((): void => {
+            this.adjustNavigator();
+            this.chart.redraw();
+        }, 50));
```
karolkolodziej, line 560: Why 50ms? It doesn't look good on the initial render

karolkolodziej (changes requested):
A couple of things:
- Comments
- When trying to edit the `NavigatorComponent` through the edit mode I get errors in the console.
- API docs are not generated
- Documentation about the component is missing

ts/Dashboards/Plugins/NavigatorComponent.ts
```diff
@@ -0,0 +1,690 @@
...
+            NavigatorComponent.charter ||
+            Globals.win.Highcharts as unknown as typeof H
+        );
+
+        this.chartContainer = Globals.win.document.createElement('div');
+        this.chart = charter.chart(
+            this.chartContainer,
+            (this.options.chartOptions || {})
+        );
+        this.chartContainer.classList.add('navigator-container');
```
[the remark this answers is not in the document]
sebastianbochan: It would be good to extract class name to Globals, as we have in other parts of the Dashboards.

ts/Dashboards/Plugins/NavigatorComponent.ts
bre1470: Navigator animation can not be turned off. I will try an increase.

pawellysy (approved):
LGTM

karolkolodziej (approved):

Vazonik (approved):

oysteinmoseng (changes requested):
Nice work! 🎉 

I would expect filtering one of the navigators to also affect the content of the other navigators, not just the data grid. Is this a quick config, or should we create a follow up task? 

Also needs tutorial docs, but can add a new task for this if we want this merged quickly.

bre1470:
Okay, then these have to be follow up tasks. I can work on it, but it will not make it for the next release.

oysteinmoseng (approved):
Approving this then so we can merge, and then will create follow up tasks

PR #19498, opened 2023-08-05
Fixed #18490, tooltip.outside failed to clean up container

### LINKED ISSUE #18490: Tooltip with useHTML and outside does not cleanup tooltip container
#### Expected behaviour

Tooltips with outside: true and useHTML: true should hide page overflow by default and also cleanup their container when the chart is not actively hovered.

#### Actual behaviour

I noticed a scroll issue bug already reported here: https://github.com/highcharts/highcharts/issues/17561. I was able to work-around that by setting a custom tooltip class that hides overflow. However, highcharts doesn't appear to remove the container after hover every time. The example straight from the docs for useHTML has this issue:
https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/tooltip/fullhtml/

You can see if I hover quickly, then move out of the preview area, sometimes the container is left there:
image

This isn't super noticable, but when you have pages with scroll overflows and fixed positioning rules, this can cause all sorts of undesired page scrolling behavior. This also compounds with more complex charts, or multiple charts on the page.
image

This is especially problematic on pages that shouldn't be overflowing when you've got taller tooltips, because it can cause a ton of extra whitespace scrolling on the page that isn't desired. Here's an example that seems outlandish, but does come up occasionally.
image

Also for context, because of this overflow and scrolling issue, I essentially had to add the following overrides to the tooltip container to prevent a lot of layout issues I was running into.
```
max-height: 100vh!important;
overflow-y: hidden!important;
position: fixed!important;
z-index: 10000!important;
```

see link above

#### Product version
highcharts 10.3.3, but happened on previous versions also

#### Affected browser(s)
any. Mostly tested in chrome

### LINKED ISSUE #19325: Tooltip with 'outside' and 'useHTML' lingers after leaving the hover area instead of being removed.
#### Expected behaviour
Leaving the chart hover area should remove the tooltip from the HTML.

#### Actual behaviour
Leaving the chart hover area does not remove the tooltip from the HTML. After hovering the tooltip and resizing the browser, the lingering tooltip causes unwanted whitespace in the outer container.

http://jsfiddle.net/bhmgv7rx/

Steps to Reproduce:

1. Hover over the 2nd (right-most) chart to show a tooltip. This chart must use `outside: true` and `useHTML: true`.
2. Resize the outer container so the width is less < 500px (see the CSS media query). The charts will wrap and take up 100% width as expected.
3. Use a trackpad or mouse-wheel to scroll to the right. Extra whitespace is present on the right side of the outer container.

Steps to Reproduce (video):
https://github.com/highcharts/highcharts/assets/1767135/590e6081-f993-4a3f-9257-42e4f84d0b96

Note: There's an [existing issue](https://github.com/highcharts/highcharts/issues/18490) that seems related, but the suggested 'fix' is basically a hack. [That bug](https://github.com/highcharts/highcharts/issues/18490) (which might be fixed) is that the tooltip stays over the screen. In this bug (19325) the tooltip isn't visible on screen, but causes extra whitespace.

#### Product version
Highcharts v11.1.0 (Our app uses v10.3.3. This bug is present in both versions).

#### Affected browser(s)
Chrome, Safari, possibly others.

TorsteinHonsi:
Fixed #18490, tooltip container was not cleaned up from DOM when `tooltip.outside` applied.
___ 
Closes #19325

### DIFF
--- a/samples/unit-tests/tooltip/outside/demo.js
+++ b/samples/unit-tests/tooltip/outside/demo.js
@@ -41,14 +41,16 @@ QUnit.test('Outside tooltip styling and correct position', function (assert) {
                 }
             },
             tooltip: {
-                outside: true
+                outside: true,
+                hideDelay: 0
             },
             series: [{
                 data: [1, 3, 2, 4]
             }]
         }),
         point = chart.series[0].points[0],
-        tooltip = chart.tooltip;
+        tooltip = chart.tooltip,
+        controller = new TestController(chart);
 
     // Set hoverPoint
     point.onMouseOver();
@@ -104,6 +106,20 @@ QUnit.test('Outside tooltip styling and correct position', function (assert) {
         Math.round(pointX),
         'Tooltip position should appear at point with sets margin for chart and container'
     );
+
+    assert.strictEqual(
+        chart.tooltip.container.parentNode.nodeName,
+        'BODY',
+        'The outside tooltip should be part of the body'
+    );
+
+    controller.moveTo(-1, -1);
+    assert.strictEqual(
+        chart.tooltip.container.parentNode,
+        null,
+        'When hiding the tooltip, the container should be removed from DOM (#18490)'
+    );
+
 });
 
 QUnit.test('Tooltip when markers are outside, #17929.', function (assert) {
--- a/ts/Core/Renderer/SVG/SVGElement.ts
+++ b/ts/Core/Renderer/SVG/SVGElement.ts
@@ -1318,29 +1318,6 @@ class SVGElement implements SVGElementLike {
     }
 
     /**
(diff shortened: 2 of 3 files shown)


karolkolodziej (approved):

KacperMadej (commented):
> I assume label.animate() can simply be used instead 🤔

Yes, e.g. the below code works fine and has no memory leak.
```
                    label.animate({
                    	opacity: 0
                    }, {
                    	complete: label.destroy
                    });
```

> ...breaking some tests, so I had to...

The related change is insignificant, so I don't see a point in investing any more time into a change or investigation, but in general we can change tests - especially if they are false negatives.

Nothing more from me - everything else is ok ✅

PR #19501, opened 2023-08-07
Changed dataset

tordvv:
Changed the dataset of this demo to a more positive set. The demo might not be as visually impressive as before, but the functionality is identical.


TorsteinHonsi:
Closing this, the data set has less variation

PR #19502, opened 2023-08-07
Added upload of truncated versions

goransle:


tools/gulptasks/dashboards/dist-upload.js
```diff
@@ -288,6 +292,22 @@ async function uploadZips(
...
+ */
+function getTruncatedVersions(fullVersion) {
+    const [major, minor] = fullVersion.split('.');
+
+    return [
+        `${major}.${minor}`,
+        major
+    ];
+}
+
```
bre1470, line 310: Can we merge this down with the other array? Seems to be a lot of code for one call and two array items.

bre1470 (changes requested):

bre1470 (approved):
Thank you! Looks good. 👍

PR #19503, opened 2023-08-07
Moved getModifiedTableOffset to syncHandler file

goransle:
Removed method `getModifiedTableOffset` from RangeModifier class


karolkolodziej (approved):
Will this go to the changelog? If so, please add the label and PR message.

ts/Dashboards/Plugins/HighchartsSyncHandlers.ts
```diff
@@ -27,15 +27,59 @@ import type {
...
+ * The row offset of the modified table.
+ */
+function getModifiedTableOffset(
+    table: DataTable,
+    modifierOptions: RangeModifierOptions
+): number {
+    const { ranges } = modifierOptions;
+
+    if (ranges) {
+        const minRange = ranges.reduce(
```
sebastianbochan, line 60: How about the performance, should not we use a regular loop ?

sebastianbochan (approved):

pawellysy (approved):

PR #19504, opened 2023-08-07
bugfix/18858-boost-point-colors

### LINKED ISSUE #18858: Invisible boost series when point color is set.
#### Expected behaviour
When there is more than one boosted series, setting the color for individual points causes the points to have the same color as in the options for all the series.

#### Actual behaviour
When there is more than one boosted series, setting the color for individual points causes the points to have the same color as in the options for the first series only, the subsequent ones are invisible (they have transparent points).

https://jsfiddle.net/BlackLabel/10s58omu/

#### Product version
Highcharts v10.3.3

Vazonik:
Fixed #18858, invisible series when point color was set in the chart boost mode.

### DIFF
--- a/samples/highcharts/boost/line-data-color/demo.css
+++ b/samples/highcharts/boost/line-data-color/demo.css
@@ -1,0 +1,5 @@
+#container {
+    height: 400px;
+    max-width: 800px;
+    margin: 0 auto;
+}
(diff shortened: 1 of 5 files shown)


TorsteinHonsi (approved):
Thanks!

PR #19507, opened 2023-08-07
Demos/sankey diagram dataset update

tordvv:
Changed the dataset of the demo to match the dataset from http://localhost:3030/samples/#view/highcharts/chartchooser/categorical-flow-sankey, and added source


TorsteinHonsi (approved):
Thanks!

PR #19508, opened 2023-08-07
dash/default-connector

sebastianbochan:
Applied default connector to the dragged new element.

------
- [x] - DataGrid
- [x] - Highcharts
- [x] - KPI
   - add value to editableOptions
- [x] - tests


ts/Dashboards/Plugins/HighchartsComponent.ts
```diff
@@ -719,6 +717,36 @@ class HighchartsComponent extends Component {
...
+    private getDefaultColumnAssignment(
+        columnNames: Array<string> = []
+    ): Record<string, string | null> {
+        const defaultColumnAssignment:Record<string, string> = {};
+
+        for (let i = 0, iEnd = columnNames.length; i < iEnd; ++i) {
+            defaultColumnAssignment[columnNames[i]] = 'y';
+            if (i === 0) {
+                const firstColumnValues =
+                    this.presentationTable?.getColumn(columnNames[i]);
```
bre1470, line 739: If you just test the first cell of the column, then it is more efficient to use `DataTable.getCell`.

bre1470: Otherwise you can also use the asReference argument of `DataTable.getColumn`, so that you directly access the actual column instead of a slice copy.

ts/Dashboards/EditMode/EditRenderer.ts
```diff
@@ -572,7 +572,7 @@ function renderInput(
             id: options.id || '',
             name: options.name || '',
             value: (
-                options.value && options.value.replace(/\"/g, '') ||
+                options.value && options.value.toString().replace(/\"/g, '') ||
```
bre1470, line 575: Something like `isString(options.value) ? ...replace... : value` might be better, because you will miss zero otherwise.

ts/Dashboards/Components/KPIComponent.ts
```diff
@@ -123,7 +123,15 @@ class KPIComponent extends Component {
                 `${Component.defaultOptions.className}-kpi`
             ].join(' '),
             minFontSize: 20,
-            thresholdColors: ['#f45b5b', '#90ed7d']
+            thresholdColors: ['#f45b5b', '#90ed7d'],
```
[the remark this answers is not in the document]
sebastianbochan: It's not a part fo this PR / task. So, please create a separate ticket.

ts/Dashboards/EditMode/SidebarPopup.ts
```diff
@@ -103,22 +107,30 @@ class SidebarPopup extends BaseForm {
...
                             }
                         }
-                    });
+                    } as Partial<HighchartsComponent.Options>;
+
+                    if (connectorsIds.length) {
+                        options = {
+                            ...options,
+                            connector: {
+                                id: connectorsIds[0]
```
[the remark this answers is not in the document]
sebastianbochan: Its not a bug, because we have 3 columns in our CSV, so Highcharts creates 3 series. We can change the default pie into columns.

ts/Dashboards/EditMode/EditRenderer.ts
sebastianbochan: Good suggestion will fix it!

sebastianbochan: Good suggestion, I used `defined()` from Utils and it covers the problem.

ts/Dashboards/EditMode/SidebarPopup.ts
```diff
@@ -162,12 +186,12 @@ class SidebarPopup extends BaseForm {
...
                 if (sidebar && dropContext) {
-                    return sidebar.onDropNewComponent(dropContext, {
+                    const options = {
                         cell: '',
                         type: 'KPI',
-                        title: 'Example KPI',
-                        value: 70
-                    });
+                        title: 'Example KPI'
+                    } as Partial<KPIComponent.ComponentOptions>;
```
bre1470, line 193: It would be better if you have full type check via
`const options: Partial<KPIComponent.ComponentOptions> =...`
than to force the type with fewer type checks
`... as Partial<KPIComponent.ComponentOptions>`.

bre1470 (approved):

bre1470:
One comment, otherwise LGTM. 👍

karolkolodziej (approved):

ts/Dashboards/EditMode/SidebarPopup.ts
```diff
@@ -379,6 +403,13 @@ class SidebarPopup extends BaseForm {
                                 components[i].onDrop(sidebar, dropContext);
 
                             if (newCell) {
+                                const mountedComponent =
+                                    newCell.mountedComponent;
+                                // skip init connecter when is not defined by
```
pawellysy, line 408:
```suggestion
                                // skip init connector when is not defined by
```

ts/Dashboards/EditMode/SidebarPopup.ts
sebastianbochan: Set type implicited, is better than casting, I agree

pawellysy (approved):

PR #19509, opened 2023-08-07
dash/input-focus

karolkolodziej:
Fixed, when editing a data grid cell, the input focus was lost.


sebastianbochan (approved):

bre1470:
This PR can be merged before release as it is a RC fix.
It should not go into changelog afaik, as it fix an internal regression instead of a bug from last version.

PR #19510, opened 2023-08-07
Fixed sync-aliases demo, added trim of headers

goransle:
Fixed whitespaces not being removed from names when using CSVConnector with `firstRowAsNames` enabled


bre1470 (commented):
The DataTable looks correct, but the result of the assignment does not look right. Date should be the axis and not the customName series.

bre1470 (approved):
Maybe this is something for another fix in components. Visually the demo looks good.

karolkolodziej (approved):
The test is missing but we can add it in a separate PR 😉

pawellysy (approved):

sebastianbochan:
Would be good to merge it before the release.

PR #19511, opened 2023-08-07
demo/remove-flags-general

kamil-musialowski:
Removed flags general demo.

_Internal note: https://app.asana.com/0/1203542399740129/1204260211115426/f_


pawelfus (changes requested):
For now, let's keep this demo

kamil-musialowski:
Sure thing. Should we keep the PR as a draft (in case we want to remove the demo in the future?)

pawelfus:
Yes, let's keep it

PR #19512, opened 2023-08-08
Changed posittions of lables

tordvv:
Experimented with different positions for the labels, and found i liked it best when the labels surrounds the pie chart.
Screenshot 2023-08-08 at 11 17 46


TorsteinHonsi:
Thanks for the PR!

I'm afraid we can't do it like this in the demo, with hard coded rotation values for each point. That would mean that when implementers exchange our demo data with their own, they would also have to figure out the best rotation for each item. Which would basically mean writing a function to do it, because their data is usually dynamic.

PR #19520, opened 2023-08-09
Created demo with heatmap and bubble combination

tordvv:
Created a new demo which uses a combination of heatmap and bubble chart to display an amount of data inside categories of varying preferability. 

[Internal Asana reference](https://app.asana.com/0/1203542399740129/1203927121512823)

I am not very good at writing descriptions, so those might need a bit of work.
Screenshot 2023-08-09 at 10 26 10


samples/highcharts/demo/heatmap-bubbles/demo.details
```diff
@@ -0,0 +1,12 @@
+---
+name: Heat map
```
TorsteinHonsi, line 2: Heat map and bubble

samples/highcharts/demo/heatmap-bubbles/demo.details
```diff
@@ -0,0 +1,12 @@
...
+authors:
+  - Torstein Hønsi
```
TorsteinHonsi, line 4: 🤔

samples/highcharts/demo/heatmap-bubbles/demo.details
```diff
@@ -0,0 +1,12 @@
...
+compareTooltips: true
+js_wrap: b
+tags:
+  - Highcharts demo
+categories:
+  - Heat and tree maps:
+      priority: 1
```
TorsteinHonsi, line 11: Remove the priority

samples/highcharts/demo/heatmap-bubbles/demo.css
TorsteinHonsi, line 1: I think we should change the path to `heatmap-bubble`, since "bubble" in singular is the name we use for the series type.

samples/highcharts/demo/heatmap-bubbles/demo.js
```diff
@@ -0,0 +1,106 @@
...
+        borderWidth: 1,
+        color: '#FFFFFF',
+        data: [[0, 4, 3], [1, 4, 2], [2, 4, 1], [3, 4, 0], [4, 4, 0],
+            [0, 3, 4], [1, 3, 3], [2, 3, 2], [3, 3, 1], [4, 3, 0],
+            [0, 2, 6], [1, 2, 5], [2, 2, 4], [3, 2, 2], [4, 2, 1],
+            [0, 1, 8], [1, 1, 7], [2, 1, 6], [3, 1, 5], [4, 1, 3],
+            [0, 0, 10], [1, 0, 10], [2, 0, 8], [3, 0, 6], [4, 0, 5]],
+        dataLabels: {
+            enabled: false
+        },
```
TorsteinHonsi, line 68: I don't think this is necessary, as data labels are disabled by default.

samples/highcharts/demo/heatmap-bubbles/demo.js
```diff
@@ -0,0 +1,106 @@
...
+        tooltip: {
+            enabled: false
+        },
```
TorsteinHonsi, line 71: Disabling the tooltip is not necessary because we have the `enableMouseTracking` option set to false.

samples/highcharts/demo/heatmap-bubbles/demo.js
```diff
@@ -0,0 +1,106 @@
...
+
+    responsive: {
+        rules: [{
+            condition: {
+                maxWidth: 500
+            },
+            chartOptions: {
+                yAxis: {
+                    labels: {
+                        format: '{substr value 0 1}'
```
TorsteinHonsi, line 99: This causes the labels to disappear altogether in mobile view, because we don't have a built-in templating function called `substr`. Actually I don't think we need the responsive option structure at all, because there is room enough for the full category names.

samples/highcharts/demo/heatmap-bubbles/demo.js
```diff
@@ -0,0 +1,106 @@
...
+
+    yAxis: {
+        categories: ['Insignificant', 'Minor', 'Moderate', 'Major', 'Critical'],
+        title: null,
+        min: 0,
+        max: 4
+    },
+
+    chart: {
+        marginTop: 40,
```
TorsteinHonsi, line 25: The `marginTop` setting makes the plot area touch the export icon in dark mode. Also, if a user wants to use this config with a longer title that wraps, that will not work well with an explicit top margin. Usually it is better to leave this out of the config and let the defaults work out the position.

TorsteinHonsi (changes requested):
Thanks! That's a good data set. Some more nitpicking comments related to the configuration. When we get those sorted out, we're good to go.

TorsteinHonsi (approved):

PR #19523, opened 2023-08-09
demo/stock-technical-analysis

kamil-musialowski:
Added a technical analysis demo with annotations.

_Asana:_ https://app.asana.com/0/1203542399740129/1204260211115403/f


kamil-musialowski:
@TorsteinHonsi Since we have to use a static dataset for this one, do you have any recommendations for that instead of clogging the HTML with 1000 lines? (moved to HTML to make the JS more readable) Or should I leave it as it is?

pawelfus (changes requested):
Nice! A couple of minor things:

- Could you add also `id` to series? It seems without it we can't add indicators from GUI
- Could you add names or `langKey` to added annotations? Otherwise, the name is missing in the popup
- ~It's not possible to scroll in the Indicators Popup~ Edit, reported:    https://github.com/highcharts/highcharts/issues/19736

kamil-musialowski:
Merged master to fix visuals & Chrome.

kamil-musialowski:
@TorsteinHonsi Could you please suggest what could we do in such situation, where a static dataset is needed? 
The approach with HTML does not pass the Visuals check, so should I simply move it to JS, or maybe there is anything better that could be done here? Thanks :)

TorsteinHonsi:
I think you should put it in `/samples/data`. See [the description](https://github.com/highcharts/highcharts/tree/master/samples#:~:text=Using%20data%20from%20/samples/data).

pawelfus (approved):
Thanks!

@goransle - any thoughts about the height of the container? Will the thumbnail look good?

goransle:
Thumbnail looks good to me:

thumbnail

TorsteinHonsi (approved):

PR #19527, opened 2023-08-10
Changed data and corresponding info

tordvv:
Changed the date of the data from 13th of june 2022 to 30th of december 2015. Changed tehe information around this data as well, and added more plot bands.


TorsteinHonsi (approved):
Thanks!

Issue #19271, opened 2023-07-03
Missing data in CSV export for scatter plot


foxcodes:
#### Expected behaviour
All data is included in the exported CSV file

#### Actual behaviour
Some data is omitted when 2 different series have data with the same x value

#### Live demo with steps to reproduce
JSFiddle: https://jsfiddle.net/ekuLfg4y/

* Click on the export menu and Download as CSV
* Open the downloaded CSV
* Observe that only 5 rows are present rather than 8

#### Product version
11.1.0

#### Affected browser(s)
Chrome

foxcodes:
from my investigation seems that issue stems from `ExportData.ts` line 564 where it checks for `rows[key]`. If two data series with the same `x` value are there then it overwrites the original data with the new series one.

bm64:
Hello @foxcodes and thanks for reporting the issue.

foxcodes:
any ETA on when the bug might be approached?

bm64:
Hi @foxcodes. Unfortunately, we don't have an ETA yet, but I'll let you know as soon as we do.

oysteinmoseng:
Related: https://github.com/highcharts/highcharts/issues/10743

foxcodes:
Hi,

I can see the issue is closed and that a branch was raised, but when I get the latest version of Highcharts 11.4.0, the issue still exists and I can't see the code changes from the PR: https://github.com/highcharts/highcharts/commit/6d29948fa877df02f5a54ca1695a71758222431d?diff=unified&w=0 , being part of the code,

Can I get an update on what has happened with the issue that was reported?

hubertkozik:
Hi @foxcodes!
I've just tested the OP demo, which was provided in first post here: https://jsfiddle.net/BlackLabel/5e7bphud/ and everything is working correctly. Exported CSV contains 8 elements, for two series (ss attached).
<img width="423" alt="image" src="https://github.com/highcharts/highcharts/assets/62724907/7799eb00-19d9-4d18-b77b-be5222d7bca1">
Can you double-check? Maybe you have some issues with the cache?

foxcodes:
Hi @hubertkozik,

Thank you for your reply, I checked it again, and removed the overwrite we had for the `getDataRows` and I can see the fix working. 

Thank you for resolving the original issue!

PR #19532, opened 2023-08-10
bugfix/19271-added-uniqueNames-option-for-csv-export

KamilKubik:
Added `uniqueNames` option for csv export, see #19271.


TorsteinHonsi:
Thanks for the PR, it works great! One thing to consider though: The OP demo was a scatter chart. A fundamental difference between scatter data and the (default) line-type data, is that scatter data is not necessarily sorted on `x`, like the line series are. Would it make sense to check the [requireSorting](https://github.com/highcharts/highcharts/blob/dcf0ac67b30ed690429aac019c5eceb3a1139aaf/ts/Series/Scatter/ScatterSeries.ts#L165) flag for each series, when we decide whether to render x data separately? So when `requireSorting` is false, it would behave like `uniqueNames: true`. Maybe we don't need a new option for this in the first place?

A problem with that approach is what to do with multiple series. Maybe we just need more rows if (and only if) there are multiple occurences of the same x value.

KamilKubik:
> Thanks for the PR, it works great! One thing to consider though: The OP demo was a scatter chart. A fundamental difference between scatter data and the (default) line-type data, is that scatter data is not necessarily sorted on `x`, like the line series are. Would it make sense to check the [requireSorting](https://github.com/highcharts/highcharts/blob/dcf0ac67b30ed690429aac019c5eceb3a1139aaf/ts/Series/Scatter/ScatterSeries.ts#L165) flag for each series, when we decide whether to render x data separately? So when `requireSorting` is false, it would behave like `uniqueNames: true`. Maybe we don't need a new option for this in the first place?

Thanks for the clarification on that! My thinking was that the main issue was to render all the points separately, as that's what the user wanted to achieve. I imagine it can be some datasets that can be more useful if we render them separately. If I understood the idea correctly, we can set the condition to e.g.: `!series.requireSorting || uniqueNames`, we even have a property for the `!series.requireSorting` already defined, but I suppose we won't leave an option for the user whether to render points separately or not then. Of course, it will happen only for a few series as I can see (scatter, pie, x-range), but I think it's more about a business decision if we want to always "merge" points with the same `x` values for these series or not. 

> A problem with that approach is what to do with multiple series. Maybe we just need more rows if (and only if) there are multiple occurences of the same x value.

Maybe I'm wrong here but don't we always generate the points keys based on the `x` values, and conditionally multiple occurrences of these `x` values? I mean, is there any example of when multiple rows will be rendered and there won't be any multiple `x` values occurrences? 

To sum up my thinking, I wanted users to always decide whether to render points separately or not, no matter what the series is, but I agree it may be a good idea to use the `requireSorting` option, but I assume it's more about the business decision. I was also wondering if we could e.g. add some options in the future to decide how the "merged" points should behave. 

Of course, there's a chance I misunderstood the idea, so please, feel free to correct me.

TorsteinHonsi:
Thinking about it, maybe we don't need to consider `requireSorting` or `uniqueNames` at all. If two different points have the same x value, we would expect them to render on separate rows regardless, right? I don't think it is more complicated than that. 

Then in a multiple series scenario, if Series 1 has two points on a certain x value, and Series 2 has three points on that same x value, then we need three rows. Basically, any given x value requires as many rows as the maximum amount of points in it, per series.

KamilKubik:
> Thinking about it, maybe we don't need to consider `requireSorting` or `uniqueNames` at all. If two different points have the same x value, we would expect them to render on separate rows regardless, right? I don't think it is more complicated than that.

Yes, I agree that points with the same x value (in the scope of the same series) should be always rendered on separate rows regardless. 

> Then in a multiple series scenario, if Series 1 has two points on a certain x value, and Series 2 has three points on that same x value, then we need three rows. Basically, any given x value requires as many rows as the maximum amount of points in it, per series.

I tried to figure out any practical example (reason) for keeping points separated but the best I did seems to be a bit unlogical to me: https://jsfiddle.net/BlackLabel/60ojkx2a/ - I'd probably expect the _Person 1_ to be rendered. Let me also go through the merging functionality. Please, take a look at this example: https://jsfiddle.net/BlackLabel/1j2Ld5ts/. The presented data on a chart doesn't reflect much of the real ones, but the _Person 2_ is completely missing in the table. Would you think this should be also considered?

TorsteinHonsi:
Good job on the cases! I must admit I wasn't aware we handled `name` and `x` separately like this. But that obviously makes sense. 

So when `name` is present, it takes over for `x` as the header column (delete all names, and `x` will be bold, in the `th` cells).

I think we need to consider both `name` and `x`.

* Same `name` and `x`: Separate rows
* Same `name`, different `x`: Separate rows
* Different `name`, same `x`: Separate rows

In short, if I get this right:
> I'd probably expect the Person 1 to be rendered.
Yes, me too
> Person 2 is completely missing in the table. Would you think this should be also considered?
Yes

KamilKubik:
Thanks for your answer!  

I think I've faced a small blocker here, but it probably shows us a deeper problem, and maybe a need to use an external option, like `uniqueNames`. Please, consider these two examples: https://jsfiddle.net/BlackLabel/xc7bu6m1/, https://jsfiddle.net/BlackLabel/0u52tab9/. The first one is based on our [tests](https://github.com/highcharts/highcharts/blob/dcf0ac67b30ed690429aac019c5eceb3a1139aaf/samples/unit-tests/accessibility/datatable/demo.js#L62) in which we expect rows to merge, but in the second example, we expect rows not to be merged. So we basically have two cases with the same data (same `x`, different `name`), but we expect two different behaviors. 

I think this case from tests is a bit specific, as at first, I'd think of merging only when we have the same `x` and `name` values (two different series).

TorsteinHonsi:
Sorry for the delay!

> I think this case from tests is a bit specific, as at first, I'd think of merging only when we have the same x and name values (two different series)

That is what I am thinking too. If the name and the x are identical, use the same row.

In the [second demo above](https://jsfiddle.net/BlackLabel/0u52tab9/), I think it should be two rows because the names are different?

KamilKubik:
> Sorry for the delay!
> 
> > I think this case from tests is a bit specific, as at first, I'd think of merging only when we have the same x and name values (two different series)
> 
> That is what I am thinking too. If the name and the x are identical, use the same row.
> 
> In the [second demo above](https://jsfiddle.net/BlackLabel/0u52tab9/), I think it should be two rows because the names are different?

My apologies for the delay too. 

Yes, I think we should always keep separate rows for different names.

I've ended up just depending on both **x** and **name** values as you suggested. So to sum up everything, the rows are being merged only when the **x** and **name** values are the same (in different series). This solution covers all the examples mentioned by us in this thread, and also automatically fixes the customer's [issue](https://github.com/highcharts/highcharts/issues/19271). It's simple, but I think it's probably the most appropriate.

The only collision was with this mentioned [test](https://github.com/highcharts/highcharts/blob/dcf0ac67b30ed690429aac019c5eceb3a1139aaf/samples/unit-tests/accessibility/datatable/demo.js#L62). I've removed it for now - I think we both somehow agreed to that? Please let me know if so and what you think about the current solution. If we both agree, I'll add new replacement tests for the removed one to cover the current solution.

TorsteinHonsi (approved):
Thanks!

PR #19537, opened 2023-08-11
Changed labels and added legend to popup

tordvv:
Removed the names from the data labels so that just the percentages is displayed next to the pie charts. Added a legend to display the names of the candidates instead


TorsteinHonsi (approved):
Thanks, looks good!

PR #19544, opened 2023-08-11
bugfix/19497-boost-category-axis-disappearing-points

### LINKED ISSUE #19497: Boost mode with categories causes points inside plot to disappear
#### Expected behaviour
Points inside the plot area should always be rendered

#### Actual behaviour
While scrolling after zoom, the points get clipped before they reach the edge of the plot.

The series needs to be boosted and `xAxis.categories` must be set.

Zoom in and start scrolling horizontally. Notice that some points disappear early than we expect, and eventually most disappear before the category is off the screen.
https://jsfiddle.net/9g7Lmqph/1/

image
image

#### Product version
Latest

#### Affected browser(s)
All

Vazonik:
Fixed #19497, some points disappeared when changing extremes of the category axes in boost.


TorsteinHonsi (approved):
Thanks!

PR #19548, opened 2023-08-11
bugfix/16769-tunnel-annotation-log-axis

### LINKED ISSUE #16769: parallel channel in Highstock does not work in logarithmic mode
#### Expected behaviour
when drawing a parallel channel the chart should take into accoun the logarithmic scale and appear parallel to end user even in that case.. that is how stock trader would use

#### Actual behaviour
when in logarithmic mode, it gets skewed which is incorrect

https://jsfiddle.net/anoosurf/uzwvrhjt/3/

#### Product version
highcharts /highstock 9.3.2

#### Affected browser(s)
image

kamil-musialowski:
Fixed #16769, tunnel annotations had incorrect height on logarithmic axes.


TorsteinHonsi (changes requested):
Thanks!

There's an error while building images for the visual comparison. CI fails on the following command, try running this to debug:

`npx karma start test/karma-conf.js --tests stock/*/* --reference --browsercount 2 --no-fail-on-empty-test-suite`

You can narrow in on the problematic code by modifying the `--tests` argument and reduce the `--browsercount`.

pawelfus (approved):
Look good, thanks!

PR #19553, opened 2023-08-14
demo/19322-negative-logarithm-tick-interval

### LINKED ISSUE #19322: tickInterval is not work currectly in negative logarithm
I'm using https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/yaxis/type-log-negative/ demo

But tickInterval is not same when I set allowNegativeLog to before

#### Expected behaviour

https://jsfiddle.net/ecn7sb3z/8/

It should be line on in yAxis in `1, 2, 4, 6, 8, 10, 20`

image

#### Actual behaviour

https://jsfiddle.net/pxgcn79a/2/

It only show `10, 20` lines

image

https://jsfiddle.net/pxgcn79a/2/

#### Product version
Last version

Vazonik:
Added the ability to set a custom `tickInterval` to the [negative logarithmic axis demo](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/yaxis/type-log-negative/). Closes #19322.

### DIFF
--- a/samples/highcharts/yaxis/type-log-negative/demo.js
+++ b/samples/highcharts/yaxis/type-log-negative/demo.js
@@ -4,6 +4,10 @@
  * logarithmic axis never reaches or crosses zero.
  */
 (function (H) {
+    /* eslint-disable no-underscore-dangle */
+    const logAdditions =
+        H._modules['Core/Axis/LogarithmicAxis.js'].Additions.prototype;
+
     H.addEvent(H.Axis, 'afterInit', function () {
         const logarithmic = this.logarithmic;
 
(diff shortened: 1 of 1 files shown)


TorsteinHonsi:
Thanks! The demo is now re-implementing a lot of code from the library just to change some small part of it. Which makes it prone to errors for example if we refactor the code in the library, or if we fix a bug that is not fixed in the demo etc. In short, we repeat ourselves in the demo.

Not that we want to take this too far, but how would it look if we did the change in the core itself? You can post the changes as a diff.

Vazonik:
> Not that we want to take this too far, but how would it look if we did the change in the core itself? You can post the changes as a diff.

The only thing that comes to my mind is to extract the piece of code into a wrappable method (in core): 
https://github.com/highcharts/highcharts/blob/d56249fd87b3a6c56438bdceac78af25c8b2be6c/ts/Core/Axis/LogarithmicAxis.ts#L236-L251

And then in the demo, just wrapping it. However, it still seems to me quite unnecessary to interfere in the core in order to improve the demo, which presents a strongly customized and basically mathematically incorrect logarithmic scale.

This fix was a response to a problem that occurred when a user introduced a custom `tickInterval` to this demo. Without any interference, this demo works properly, so I had doubts about whether such a change is necessary not only in the core but even in the demo. @TorsteinHonsi

samples/highcharts/yaxis/type-log-negative/demo.js
```diff
@@ -37,6 +41,90 @@
...
+        max,
+        minor
+    ) {
+        if (
+            !this.axis.options.custom.allowNegativeLog ||
+            interval >= 0.5 ||
+            interval < 0.08
+        ) {
+            return proceed.call(this, interval, min, max, minor);
+        }
```
Vazonik, line 59: Anyway, I've slightly improved this solution using a wrapper, removing as much core code repetition as possible and reducing the error-proneness associated with future Highcharts updates.

TorsteinHonsi (approved):
Ok, thanks! Let's do it like this.

PR #19554, opened 2023-08-14
Music format revenue race chart demo.

DJTechnoo:
Race area chart showing the revenue by format in the US for the past 50 years.


samples/highcharts/demo/music-race-chart/demo.details
TorsteinHonsi, line 1: Better rename it to `area-race-chart` I think, it's more likely what users are googling for.

samples/highcharts/demo/music-race-chart/demo.details
```diff
@@ -0,0 +1,10 @@
+---
+name: Music Format Area Race Chart
```
TorsteinHonsi, line 2: Shorten to "Area Race Chart"

samples/highcharts/demo/music-race-chart/demo.js
```diff
@@ -0,0 +1,196 @@
...
+    startYear = 1973,
+    endYear = 2021;
+
+// General helper functions
+const getRevenue = (record, year) => {
+    const revenue = record[year];
+    if (typeof revenue !== 'undefined') {
+        return revenue;
+    }
+    return 0;
```
TorsteinHonsi, line 12: If `record[year]` is either a number or undefined, this could be shortened to 

`const getRevenue = (record, year) => record[year] || 0`

In that case, maybe we don't need a helper function for it either?

samples/highcharts/demo/music-race-chart/demo.js
```diff
@@ -0,0 +1,196 @@
...
+    // Reset slider at the end
+    if (input.value > endYear) {
+        input.value = startYear;
+    }
+    button.title = 'pause';
+    button.className = 'fa fa-pause';
+    chart.sequenceTimer = setInterval(function () {
+        update();
+    }, 700);
+}
```
TorsteinHonsi, line 183: On the contrary to our other race chart demos, this one has no visible data to start with. So I think it would be better to auto-play the race, otherwise users will be looking at an empty canvas until they find the play button.

samples/highcharts/demo/music-race-chart/demo.js
```diff
@@ -0,0 +1,196 @@
...
+    button.title = 'play';
+    button.className = 'fa fa-play';
+    clearTimeout(chart.sequenceTimer);
+    chart.sequenceTimer = undefined;
+}
+
+function update() {
+    chart.update({
+        subtitle: {
+            text: 'Year ' + input.value
```
TorsteinHonsi, line 128: The year kind of disappears. Can we put it in a big fat font somewhere on top of the chart, like we did with the [bar race](https://www.highcharts.com/demo/bar-race/brand-dark)?

samples/highcharts/demo/music-race-chart/demo.js
```diff
@@ -0,0 +1,196 @@
...
+                fontSize: '0.5em'
+            }
+        }
+    }, {
+        label: {
+            style: {
+                fontSize: '0.5em'
+            }
+        }
+    }]
```
TorsteinHonsi, line 114: Since we already needed to do some hard-coding here, I am wondering if series-labels is the right tool for the job in the first place. The way the labels move during animation to adapt to the available area, is a bit distracting and appears buggy. 

Can you do an attempt at using annotations instead? With an [annotation label](https://api.highcharts.com/highcharts/annotations.labels) we could hard code the label to a specific [x, y] value, which would keep it in place (relative to the axes) as the data comes in. Not sure if rotation is supported, but if it is, we could use that for narrow, rising bands like the Cassette. Probably we need to add the annotation dynamically as the content is revealed. We can also consider updating the font size and/or rotation depending on the chart width or the y axis translation.

samples/highcharts/demo/music-race-chart/demo.js
```diff
@@ -0,0 +1,196 @@
...
+        min: startYear,
+        max: endYear
+    },
+    yAxis: {
+        title: {
+            text: 'Revenue in the U.S.'
+        },
+        labels: {
+            format: '${text} B'
+        }
```
TorsteinHonsi, line 56: On the mobile, it would buy us some plotting space if we reorganized this. Add a responsive rule, then put the labels on top of the grid lines. The title could probably be placed on top of the axis.

TorsteinHonsi (changes requested):
Thanks, great work! Some thoughts added.

samples/highcharts/demo/music-race-chart/demo.details
TorsteinHonsi, line 1: We now have a `music-race-chart` and an `area-race-chart` demo. That was not the intension, right? I assume the music-race-chart should be deleted.

samples/highcharts/demo/area-race-chart/demo.js
```diff
@@ -0,0 +1,321 @@
...
+                    });
+            }
+        },
+        type: 'area',
+        marginTop: 100,
+        animation: {
+            duration: 700,
+            easing: function (t) {
+                return t;
+            }
```
TorsteinHonsi, line 48: Could be simplified to arrow function: `easing: t => t`

samples/highcharts/demo/area-race-chart/demo.js
```diff
@@ -0,0 +1,321 @@
...
+    },
+    tooltip: {
+        pointFormat:
+            '{series.name} had a revenue of <b>${point.y:,.1f} B</b><br/> in {point.x}'
+    },
+    plotOptions: {
+        area: {
+            stacking: 'normal',
+            pointStart: startYear,
+            pointEnd: endYear,
```
TorsteinHonsi, line 95: No such property in our API...

samples/highcharts/demo/area-race-chart/demo.js
```diff
@@ -0,0 +1,321 @@
...
+        play(this);
+    }
+});
+
+play(btn);
+
+// Trigger the update on the range bar click.
+input.addEventListener('input', function () {
+    update();
+});
```
TorsteinHonsi, line 321: Could be simplified to `input.addEventListener('input', update);` as far as I can see

samples/highcharts/demo/area-race-chart/demo.js
```diff
@@ -0,0 +1,321 @@
...
+            easing: function (t) {
+                return t;
+            }
+        }
+    },
+    title: {
+        text: 'Music revenue race chart'
+    },
+    subtitle: {
+        useHTML: true,
```
TorsteinHonsi, line 55: Why useHTML? It should only be used if there is formatting that we can't do in SVG.

TorsteinHonsi (changes requested):
Thanks, this demo looks great now. Great idea with making the animation easing linear!

Some things to address:

* The annotation labels capture the mouse events, effectively hindering the tooltip when hovering the label. Can be fixed by setting `pointerEvents: 'none'` to the `labelOptions.style`.

* The long empty part of modern media like download and stream add a line on top of the chart. We should replace those long segments to add `null` values instead of 0. But keep the last 0 before the actual data begins, otherwise we will get a cliff. 
Skjermbilde 2023-09-12 kl  08 50 27

* After fixing that, check the tooltip over CD. I am not sure what makes it stick to other series. From what I remember, it should show the data of the area it is hovering. Also consider a `split` tooltip. In this case it would make it easier for the viewer to compare different series against each other at a specific point in time.

* If we hide items from the legend during the animation, something goes wrong. 
Skjermbilde 2023-09-12 kl  08 55 06

TorsteinHonsi (approved):

PR #19555, opened 2023-08-15
dashboards/cypress-lighthouse

goransle:
Adds a setup for running Lighthouse tests via Cypress and GH actions

Test can be run locally with

```sh
npx cypress run --browser chrome --config specPattern="test/cypress/**/performance/*.cy.js"
```

The current climate-demo test will fail if the lighthouse performance score dips below 46 (currently it's ~47-50 in GH actions). Some selected metrics are available in the action summary, and the full report is available in the job artefacts.

Comparison with master branch will follow in a separate PR.

### DIFF
--- a/.github/workflows/dashboards-lighthouse-test.yml
+++ b/.github/workflows/dashboards-lighthouse-test.yml
@@ -118,11 +118,14 @@ jobs:
           config: specPattern=**/performance/*.cy.js
           install: false
 
-      - name: Summary
+      - name: Add to summary
+        if: ${{ always() }}
         run: |
-          node test/cypress/lighthouseCompare.mjs  --proposed ./tmp/lighthouse.json > $GITHUB_STEP_SUMMARY
+          echo "## Lighthouse metrics" >> $GITHUB_STEP_SUMMARY
+          node test/cypress/lighthouseCompare.mjs  --proposed ./tmp/lighthouse.json >> $GITHUB_STEP_SUMMARY
 
       - name: Store artifacts
+        if: ${{ always() }}
         uses: actions/upload-artifact@v3
         with:
           name: post-tests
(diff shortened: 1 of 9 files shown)


.github/workflows/dashboards-test.yml
```diff
@@ -27,7 +27,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [lts]
+        node-version: [18.x]
```
bre1470, line 30: Why the change?

goransle: In the initial commit it gave the following error: 

```sh
Run actions/setup-node@v3
Attempting to download lts...
Not found in manifest. Falling back to download directly from Node
Error: Unable to find Node version 'lts' for platform linux and architecture x64.
```

Which I took to mean that the `node-version` does not support just `lts` (but according to the docs it _should_ support `lts/{codename}`. I see now that this workflow does not use the setup-node action, so that would explain why it did not fail previously.

karolkolodziej (approved):

sebastianbochan (approved):

.github/workflows/dashboards-lighthouse-test.yml
```diff
@@ -0,0 +1,133 @@
...
+      - 'ts/Dashboards/**'
+      - 'ts/Data/**'
+      - 'css/**'
+      - '.github/workflows/**'
+
+permissions:
+  contents: read # to fetch code (actions/checkout)
+
+jobs:
+  # cypress_lighthouse_base:
```
sebastianbochan, line 19: Why do we have commented lighthouse here?

goransle: That's for the next PR, which should add the comparison with master. It requires the cypress test to be on master to work (without any hacks at least), so this PR should be merged before activating the `cypress_lighthouse_base ` job

oysteinmoseng (approved):
Nice!

PR #19556, opened 2023-08-15
bugfix/19505-missing-label-connector

### LINKED ISSUE #19505: Disappearing annotation connector
#### Expected behaviour
The connector does not disappear for the added annotation.

#### Actual behaviour

https://github.com/highcharts/highcharts/assets/39968065/cbcfaecb-18f7-45c4-aa5a-e9d0fb1b094f

https://jsfiddle.net/BlackLabel/92q5hvne/

#### Product version
All

kamil-musialowski:
Fixed #19505, label connector for the `callout` shape was not rendered in certain label positions.


TorsteinHonsi:
> Originally, the connector is rendered from the side of a label, so should it be rendered from the bottom left corner in that case?

@kamil-musialowski  Yes. That would prevent the connector from crossing over the labels. Just like the chevrons, they are added to whatever side is closest to the anchor. Not only right or left, but also top or bottom.

TorsteinHonsi (changes requested):
Thanks! We now have a situation where the split tooltip is not working well, because the connector connects to the top or bottom of the rectangle. 

Consider the `highcharts/members/renderer-callout` demo, updated in my latest commit.

When the anchor is to the left or to the right of the callback, we want it to to be connected to the side. This is how it currently works in the master, so that should not change. Screenshot from the master:

Skjermbilde 2023-08-30 kl  09 50 11

The only situation we want to fix is when the anchor is directly above or below the rounded corner. This is the case when the connector disappears in the master. In this case only, it should connect to the top or bottom side. Screenshot from the current state of this PR:

Skjermbilde 2023-08-30 kl  09 57 49

kamil-musialowski:
Got it, thank you. Now it looks and behaves even better imo.

TorsteinHonsi (approved):

PR #19566, opened 2023-08-16
dash/remove-parentElement

sebastianbochan:
Removed parentNode from the component.


pawellysy (changes requested):
We want to delete the API Option `parentElement`, not the property itself.
https://github.com/highcharts/highcharts/blob/d56249fd87b3a6c56438bdceac78af25c8b2be6c/ts/Dashboards/Components/Component.ts#L1233-L1239
This one

bre1470:
We can still access `parentElement` still via `cell.container`. IMHO the property is redundant.

karolkolodziej (approved):

PR #19567, opened 2023-08-16
Updated timeline data

nasin:
Hi, Marketing asked me to update the timeline on the About Us page to include YTD data.


TorsteinHonsi (changes requested):
Thanks! While we're at it, can you apply the same changes to `highcharts/demo/combo-timeline`?

PR #19571, opened 2023-08-16
samples/stellar-chart

hubertkozik:
Added a Stellar Chart demo.


MarkusBarstad (commented):
Looks beautiful to me! Only one thing I don't understand (and this might be a very nitpick thing, just trying to give more substantive reviews here)

Min width is set as 310px but middle text overflows before then 🤔 I think around 460px or something on my end.

samples/highcharts/demo/stellar-chart/demo.js
```diff
@@ -0,0 +1,224 @@
...
+        csv: document.getElementById('csv').innerHTML,
+        seriesMapping: [{
+            name: 0,
+            x: 1,
+            y: 2,
+            z: 3,
+            lightYears: 4,
+            planetMass: 5,
+            stellarMagnitude: 6,
+            discoveryDate: 7
```
KacperMadej, line 27: Custom stuff should be in [`custom`](https://api.highcharts.com/highcharts/series.bubble.data.custom).

samples/highcharts/demo/stellar-chart/demo.js
```diff
@@ -0,0 +1,224 @@
...
+    Planets discovered</span><br><span class='percentage'>
+    ${percentage.toFixed(2)}<sup style='font-size: 0.5em;'>%</sup></span>`;
+
+    if (!customLabel) {
+        customLabel = chart.renderer.label(labelText, 0, 0, void 0, void 0,
+            void 0, true
+        ).css({
+            color: '#000',
+            textAlign: 'center',
+            'pointer-events': 'none'
```
KacperMadej, line 212: For this to work it needs to be `pointerEvents` because of `useHTML` argument set to `true` in the constructor. Not sure if this is a bug. The style is set only for the empty group in SVG and has no effect on the HTML elements under the chart's SVG.

KacperMadej (changes requested):
- [ ] Tooltip decimal place inconsistency: 
image
(or is this really over 1k Jupiters?)

- What's the data source? If the license doesn't require this info to be visible it still would be nice to have it somewhere for maintenance.

MarkusBarstad (approved):
Feels solid!

samples/highcharts/demo/stellar-chart/demo.js
hubertkozik: I am not sure if that's a bug, we are mixing HTML and SVG elements, which overlaid with each other, so there is a good explanation for adding `pointEvents: none`.

hubertkozik:
Sadly, the data and its source are not available. I read about the licence before downloading and everything was allowed.

samples/highcharts/demo/stellar-chart/demo.js
KacperMadej: `'pointer-events': 'none'` doesn't work for me. Could you verify and try using  `pointerEvents: 'none'` instead?

samples/highcharts/demo/stellar-chart/demo.js
```diff
@@ -0,0 +1,232 @@
...
+        outside: true,
+        hideDelay: 20,
+        formatter: function (tooltip) {
+        // Hide tooltip for pie series
+            if (this.series.options.type === 'pie') {
+                return false;
+            }
+            // If not null, use the default formatter
+            return tooltip.defaultFormatter.call(this, tooltip);
+        }
```
TorsteinHonsi, line 93: This formatter can probably be omitted by just disabling tooltip for pies. `plotOptions.pie.enableMousetracking: false`.

samples/highcharts/demo/stellar-chart/demo.js
```diff
@@ -0,0 +1,232 @@
...
+                    halo: 0
+                }
+            }
+        }
+    },
+    series: [{
+        colorKey: 'custom.stellarMagnitude',
+        maxSize: 14,
+        minSize: 3,
+        tooltip: {
```
TorsteinHonsi, line 115: I think we should do `useHTML` and add a table here. Since we have four different dimensions, a data table would be more easiliy readable.

samples/highcharts/demo/stellar-chart/demo.js
```diff
@@ -0,0 +1,232 @@
...
+                        bubbleSeries = chart.series[0];
+
+                    bubbleSeries.points.forEach(point => {
+                        if (
+                            point.options.custom.discoveryDate < minDate ||
+                            point.options.custom.discoveryDate >= maxDate
+                        ) {
+                            point.graphic.attr({
+                                opacity: 0.2
+                            });
```
TorsteinHonsi, line 145: Does this work? I don't see any change in the opacity of the bubbles?

samples/highcharts/demo/stellar-chart/demo.js
```diff
@@ -0,0 +1,232 @@
...
+                maxDate: 2030
+            }
+        }
+        ]
+    }]
+});
+
+function fillCenter(percentage, info, chart, customLabel) {
+    const labelText = `<span style='font-size: 2em'>☉</span><span
+    class='uppercase'><br><span class='decade'>${info}</span><br>
```
TorsteinHonsi, line 210: Can we leave the decade out of the uppercase class? Decades are commonly written 1900s, 2000s, so the upper case S makes it kind of unfamiliar.

TorsteinHonsi (changes requested):
Thanks, great work! Besides the comments below, I noticed there's an empty grid line outside all the others. Is that intentional? If not, it probably comes from the default `yAxis.maxPadding` or `yAxis.endOnTick`.

samples/highcharts/demo/stellar-chart/demo.css
```diff
@@ -0,0 +1,28 @@
+.highcharts-figure {
+    min-width: 470px;
```
TorsteinHonsi, line 2: We need to find a way to make it work with mobile. Aim for a `min-width` of 360px. I suggest removing the outer, empty grid line (see my other comment), and create a responsive rule for the font size of the center text.

TorsteinHonsi (changes requested):

samples/highcharts/demo/stellar-chart/demo.js
hubertkozik: Not working, because of a small mistake with using `minDate` and `maxDate` from the pie series after the requested changes from Kacper. Gonna fix it.

samples/highcharts/demo/stellar-chart/demo.js
hubertkozik: But then we will lose the possibility to use `mouseOver` and `mouseOut` events on pie slices to update custom text inside the circle and opacity of bubbles...

samples/highcharts/demo/stellar-chart/demo.css
hubertkozik: Achieved with some font-size adjustments and `responsive.rule`!

TorsteinHonsi (approved):

PR #19580, opened 2023-08-17
dash/highcharts-types

bre1470:
Basically we will expect highcharts as an NPM package present in the Dashboards declarations. If we manage to avoid anything related to the plugin, it should only be a dependency when actually using the plugin.

We could create a beta branch in `dashboards-dist` to test it with the wrapper examples.

### DIFF
--- a/ts/Dashboards/Components/Component.ts
+++ b/ts/Dashboards/Components/Component.ts
@@ -29,7 +29,6 @@ import type {
     ComponentType,
     ComponentTypeRegistry
 } from './ComponentType';
-import type Globals from '../Globals';
 import type JSON from '../JSON';
 import type Serializable from '../Serializable';
 import type DataModifier from '../../Data/Modifiers/DataModifier';
@@ -45,6 +44,7 @@ const {
 } = DG;
 import DataTable from '../../Data/DataTable.js';
 import EditableOptions from './EditableOptions.js';
+import Globals from '../Globals.js';
 import U from '../../Core/Utilities.js';
 const {
     createElement,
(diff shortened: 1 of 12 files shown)


karolkolodziej (approved):
Looks promising thanks!

ts/Dashboards/EditMode/SidebarPopup.ts
```diff
@@ -138,7 +138,8 @@ class SidebarPopup extends BaseForm {
                         chartOptions: {
                             chart: {
                                 animation: false,
-                                type: 'column'
+                                type: 'column',
+                                zooming: {}
```
sebastianbochan, line 142: Do we need that?

sebastianbochan (approved):
Looks really great!

ts/Dashboards/EditMode/SidebarPopup.ts
bre1470: It is a required option in Highcharts.

goransle (approved):
LGTM

pawellysy (approved):
LGTM

ts/Dashboards/Components/Component.ts
```diff
@@ -45,6 +44,7 @@ const {
 } = DG;
 import DataTable from '../../Data/DataTable.js';
 import EditableOptions from './EditableOptions.js';
+import Globals from '../Globals.js';
```
Vazonik, line 47: Is it necessary to import the default twice from `'../Globals.js'` as `Globals` and `DG`? I guess the `DG` can be removed and `classNamePrefix` can be taken from `Globals`. I mean like:
```ts
import Globals from '../Globals.js';
const {
    classNamePrefix
} = Globals;
```

Vazonik (approved):
Looks good, I found only one detail to be fixed, one redundant import.

TorsteinHonsi (approved):
Thanks!

PR #19581, opened 2023-08-17
dash/loading-indicator

Vazonik:
Added loading indicator which turns on for a cell until a connector is loaded.

---
demo: https://jsfiddle.net/BlackLabel/yvz4m9bt/

### DIFF
--- a/ts/Dashboards/Components/Component.ts
+++ b/ts/Dashboards/Components/Component.ts
@@ -391,12 +391,18 @@ abstract class Component {
             this.options.connector?.id &&
             this.connectorId !== this.options.connector.id
         ) {
+            // Show the loading indicator
+            this.cell.setLoadingState(true);
+
             const connector = await this.board.dataPool
                 .getConnector(this.options.connector.id);
 
             this.setConnector(connector);
             this.shouldRedraw = true;
             this.redraw();
+
+            // Hide the loading indicator
+            this.cell.setLoadingState(false);
         }
         return this;
     }
--- a/ts/Dashboards/Components/Component.ts
+++ b/ts/Dashboards/Components/Component.ts
@@ -398,14 +398,20 @@ abstract class Component {
                 this.cell.setLoadingState();
             }
 
-            const connector = await this.board.dataPool
-                .getConnector(this.options.connector.id);
+            try {
+                const connector = await this.board.dataPool
+                    .getConnector(this.options.connector.id);
 
-            this.setConnector(connector);
-            this.shouldRedraw = true;
-            this.redraw();
+                this.setConnector(connector);
+                this.shouldRedraw = true;
+                this.redraw();
+            } catch (error) {
+                // Hide the loading indicator when catched an error.
+                this.cell.setLoadingState(false);
+                return Promise.reject(error);
+            }
 
-            // Hide the loading indicator.
+            // Hide the loading indicator when finished loading the connector.
             this.cell.setLoadingState(false);
         }
         return this;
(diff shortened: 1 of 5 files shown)


ts/Dashboards/Components/Component.ts
```diff
@@ -391,12 +391,18 @@ abstract class Component {
             this.options.connector?.id &&
             this.connectorId !== this.options.connector.id
         ) {
+            // Show the loading indicator
+            this.cell.setLoadingState(true);
```
sebastianbochan, line 395: I would remove true from here and make this flag as default.

ts/Dashboards/Layout/Cell.ts
```diff
@@ -590,6 +590,16 @@ class Cell extends GUIElement {
         }
     }
 
+    public setLoadingState(enabled: boolean): void {
```
sebastianbochan, line 593: `enabled: boolean = true` enabled default value.

ts/Dashboards/Layout/Cell.ts
```diff
@@ -590,6 +590,16 @@ class Cell extends GUIElement {
         }
     }
 
+    public setLoadingState(enabled: boolean = true): void {
```
karolkolodziej, line 593: - method description is missing
- IMO it should be an internal method ( use `@internal` to don't list it in the API)
- it toggles the class name so it should be called `toggleLoadingState`

karolkolodziej (changes requested):
When there is something wrong with the component, the spinner should not be shown- [demo](https://jsfiddle.net/BlackLabel/j0fuyngk/).

ts/Dashboards/Layout/Cell.ts
Vazonik: > it toggles the class name so it should be called toggleLoadingState

It toggles but always forces value (in default `true`). Should it be called `toggleLoadingState` anyway?

karolkolodziej: hmm, in that case I guess you can leave it as it is 😉

pawellysy (changes requested):
Great work!
one minor comment tho.
If the error occurs, we need to delete the spinner as well.  https://jsfiddle.net/BlackLabel/5nzsymu6/1/

ts/Dashboards/Components/Component.ts
```diff
@@ -398,14 +398,20 @@ abstract class Component {
...
+                const connector = await this.board.dataPool
+                    .getConnector(this.options.connector.id);
+
+                this.setConnector(connector);
+                this.shouldRedraw = true;
+                this.redraw();
+            } catch (error) {
+                // Hide the loading indicator when catched an error.
+                this.cell.setLoadingState(false);
+                return Promise.reject(error);
```
sebastianbochan, line 409: Do we need promise here? You can return the `error()` from HC.

sebastianbochan (changes requested):

ts/Dashboards/Layout/Cell.ts
```diff
@@ -590,6 +590,21 @@ class Cell extends GUIElement {
...
     }
 
+    /**
+     * Enables or disables the loading indicator in the cell.
+     *
+     * @internal
+     */
+    public setLoadingState(enabled: boolean = true): void {
+        const cell = this;
+        if (cell.container) {
```
sebastianbochan, line 600: The cell.container is always available (I committed it a few days ago). Please check if `cell.container.classList` is fine, otherwise `cell.container?.classList` will be enough.

sebastianbochan (changes requested):

karolkolodziej (approved):

sebastianbochan (approved):

goransle (approved):
Nice

pawellysy (approved):
LGTM!

PR #19585, opened 2023-08-18
bugfix/18940-sticky-bubble

### LINKED ISSUE #18940: Sticky tracking doesn't work on bubble
#### Expected behaviour
Sticky tracking should work by default

#### Actual behaviour
The tooltip stays open indefinitely, but does not move to the nearest bubble.

https://jsfiddle.net/jqawb92r/1/

When hovering over a bubble and then moving away, the tooltip stays in the same position indefinitely and does not move to the bubble nearest the cursor. If you were to change `type: "scatter"`, sticky tracking starts working again. 

#### Product version
Highcharts v10.3.3

#### Affected browser(s)
All

KacperMadej:
Fixed #18940, sticky tracking was not working for bubble series.

### DIFF
--- a/samples/unit-tests/pointer/members/demo.js
+++ b/samples/unit-tests/pointer/members/demo.js
@@ -85,6 +85,8 @@ QUnit.test(
             0,
             'mousemove to 0.0: no unexpected events'
         );
+        // Clear to mark only failed tests.
+        events.length = 0;
 
         // New point, and new series.
         // NOTICE Qunit has added new content to the page,
@@ -122,6 +124,8 @@ QUnit.test(
             0,
             'mousemove 30px below 1.0: no unexpected events'
         );
+        // Clear to mark only failed tests.
+        events.length = 0;
 
         // New point, same series
         // NOTICE Qunit has added new content to the page,
@@ -159,6 +163,9 @@ QUnit.test(
             0,
             'mousemove to 1.2: no unexpected events'
         );
+        // Clear to mark only failed tests.
+        events.length = 0;
+
         // Same point, same series.
         controller.setPosition(
             chart.plotLeft + chart.series[1].points[2].plotX,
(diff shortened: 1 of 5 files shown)


MarkusBarstad (approved):
Works!

samples/unit-tests/pointer/members/demo.js
```diff
@@ -173,6 +180,48 @@ QUnit.test(
...
+            chart.plotTop + chart.series[2].points[0].plotY
+        );
+        // Mouse move between the bubbles should trigger tooltip only for the
+        // first point at the start and the last point close to the end of the
+        // mouse move event due to searchKDTree including radius of the marker.
+        controller.moveTo(
+            chart.plotLeft + chart.series[2].points[0].plotX + bigRadius + 7,
+            chart.plotTop + chart.series[2].points[0].plotY + 70,
+            {},
+            true
```
hubertkozik, line 209: I am not sure if that flag works in headless tests, but I think we should remove drawing the mouse in tests?

ts/Core/Series/Series.ts
```diff
@@ -4792,7 +4791,7 @@ interface Series extends SeriesLike {
     directTouch: boolean;
     hcEvents?: Record<string, Array<U.EventWrapperObject<Series>>>;
     isCartesian: boolean;
-    kdAxisArray: Array<string>;
+    kdAxisArray: Array<'clientX' | 'plotY'>;
```
hubertkozik, line 4794: I am not sure if we should limit the type of this property that strictly. For example, in GeoHeatmap interpolation I [build kd-tree based on lon, lats value](https://github.com/highcharts/highcharts/blob/4b54538e0e35fc614b7925eccb5f01c00c69a92e/ts/Series/GeoHeatmap/GeoHeatmapSeries.ts#L703). With that change it won't be clear enough for me (if we merge GeoHeatmap logic as it is).

hubertkozik (changes requested):

ts/Core/Series/Series.ts
KacperMadej: Should be fixed now but I might be safer to wait for your branch to be merged first so I'll probably wait with making this PR not a draft until then.

hubertkozik (approved):

PR #19587, opened 2023-08-20
bugfix/18427-fix-local-build

### LINKED ISSUE #18427: Tools: dist-examples task breaks local and in CI
#### Expected behaviour
`npx gulp dist` should run until end without issues.

#### Actual behaviour
It breaks, either with
- `Error: ENOENT: no such file or directory, open './git-ignore-me.properties'`
- `Error: Could not find tmp/demo/sidebar.html` `If demos are built elsewhere, the path can be specified in git-ignore-me.properties by the demos.path property.`

Every nightly building process breaks. Otherwise run `npx gulp dist` or `npx gulp dist-examples` in the repository.

#### Product version
Repository

#### Affected browser(s)
Not related

mccahan:
See #18427 - freshly-cloned copies of the repository were failing while running `npx gulp dist`.

Caused by two issues:
- A missing `/git-ignore-me.properties` which could contain settings. If the file is missing, return an empty object so the dependencies can use their defaults.
- A missing `sidebar.html` used when running `dist-examples`. If the file is missing, attempt to generate the demos with an empty file.

I couldn't for the life of me figure out where the sidebar.html file was supposed to come from. I see a few references to parameters that come from `git-ignore-me.properties`, including the path where the demos should be created. Absent a sample default file, I return an empty object for downstream processes to fall back to their own defaults.

Fixed #18427

### DIFF
--- a/tools/gulptasks/dist-examples.js
+++ b/tools/gulptasks/dist-examples.js
@@ -204,11 +204,18 @@ async function createExamples(title, sourcePath, targetPath, template) {
 
     LogLib.success('Created', targetPath);
 
-    const localsidebar = getLocalSidebar(
-        sourcePath
-            .replaceAll('samples/', '')
-            .replaceAll('/demo', '')
-    );
+    let localsidebar;
+    try {
+        localsidebar = getLocalSidebar(
+            sourcePath
+                .replaceAll('samples/', '')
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks, looks legit!

PR #19590, opened 2023-08-21
dashboards/cypress-lighthouse

goransle:
Adds comparison of  between master and PR branch. See the [run summary](https://github.com/highcharts/highcharts/actions/runs/5950509225?pr=19590) for example output.

### DIFF
--- a/test/cypress/performance/minimal.cy.js
+++ b/test/cypress/performance/minimal.cy.js
@@ -22,7 +22,7 @@ describe('Lighthouse tests', () => {
         cy.visit('/dashboards/demo/minimal');
         cy.lighthouse(
             {
-                performance: 42
+                performance: 0
             },
             desktopConfig
         )
--- a/test/cypress/performance/minimal.cy.js
+++ b/test/cypress/performance/minimal.cy.js
@@ -18,8 +18,8 @@ const desktopConfig = {
 };
 
 describe('Lighthouse tests', () => {
-    it('Climate demo', () => {
-        cy.visit('/dashboards/demo/climate');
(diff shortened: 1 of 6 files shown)


test/cypress/performance/minimal.cy.js
```diff
@@ -18,11 +18,11 @@ const desktopConfig = {
...
 
 describe('Lighthouse tests', () => {
-    it('Climate demo', () => {
-        cy.visit('/dashboards/demo/climate');
+    it('Minimal demo', () => {
+        cy.visit('/dashboards/demo/minimal');
         cy.lighthouse(
             {
-                performance: 42
+                performance: 0
```
karolkolodziej, line 25: Is this correct?

goransle: Kind of, seems you have to set something other than `undefined` there to only run the performance tests. (Could not provide the whole object, but then other categories are also run).

The test will fail [if this comparison](https://github.com/highcharts/highcharts/pull/19590/files#diff-cf22095590be315f44cba8f347076fdf3036fb674ed60d85422d762f7085d9c0R50) is not met, so this threshold is not really needed.

sebastianbochan (approved):

pawellysy (approved):

pawellysy:
Looks good! Just one comment. Shouldn't the results be posted on the PR?

goransle:
> Looks good! Just one comment. Shouldn't the results be posted on the PR?

The results are currently only in the actions summary, as we don't have tokens for commenting in the GH secrets quite yet.

PR #19591, opened 2023-08-21
Advanced polar demo

MarkusBarstad:
Added advanced polar demo.

~~Key things I am not quite sure about:~~
~~- What more should we display on point hover/display differently?~~
~~- Is the formatting-strings okay or would it be better to make some actual css-classes?~~
~~- Proper Highcharts syntax (in what order should one define properties?)~~
~~- Resizing bugs could be aplenty~~

- [x] Decent comments (proper quantity and proper quality)
~~- [x] Subtitle-text collides into series when chart width is small~~

---
  - https://app.asana.com/0/0/1203935305712037


samples/highcharts/demo/polar-advanced/demo.js
```diff
@@ -0,0 +1,351 @@
...
+            backgroundColor: '#E2F4F8',
+            borderColor: '#D2B8FA',
+            borderRadius: 32,
+            floating: true,
+            layout: 'vertical',
+            verticalAlign: 'top',
+            squareSymbol: true,
+            borderWidth: 4,
+
+            itemStyle: { color: '#333333', fontSize: '0.86rem', letterSpacing: '0.032rem' },
```
KacperMadej, line 275: Empty line and a too long line (shouldn't linter pick this up?).

samples/highcharts/demo/polar-advanced/demo.js
```diff
@@ -0,0 +1,351 @@
...
+                        '<b>Top earner: </b><span>{point.topEarner}</span></span>'
+                    ),
+
+                    footerFormat: '<i style=\"margin:0.3rem; padding:0.3rem;width: 10rem;text-align:center; border-top: 0.1rem solid {point.color};\">Week {point.week}</i></span>'
+                }
+            }
+        ]
+    }),
+    resizeObserver = new ResizeObserver(() => {
+        chart.legend.update({ y: chart.chartHeight / 12 });
```
KacperMadej, line 348: legend.y as string with % - e.g. `y: '8%'` could remove the need for the observer, right? That could be a feature request - add support for percentage based values for the legend y position.

KacperMadej (changes requested):
1. There are few things different here than in other demos: readme.md file, HTML includes text with `highcharts-description` class, CSS file content.

2. Support for a11y and exporting is a standard thing in the demos so the related module should be added.

3. Weekly and/or daily alignment seems to be off for the week labels. Maybe the padding for x axis could be added so the bubbles are not on the edges (so not partially outside). 
image

samples/highcharts/demo/polar-advanced/demo.js
hubertkozik: Linter is not counting strings and probably that's why there wasn't any error here.

MarkusBarstad: > Linter is not counting strings and probably that's why there wasn't any error here.

Ah I see! I had no idea!

samples/highcharts/demo/polar-advanced/demo.details
```diff
@@ -0,0 +1,10 @@
+---
+name: Advanced polar chart
+authors:
+  - Markus Barstad
+js_wrap: b
+tags:
```
hubertkozik, line 6: We should also add `alt_text` as it will be the demo on the main demo page.

samples/highcharts/demo/polar-advanced/demo.html
```diff
@@ -0,0 +1,94 @@
+<script src="https://code.highcharts.com/highcharts.js"></script>
+<script src="https://code.highcharts.com/modules/coloraxis.js"></script>
+<script src="https://code.highcharts.com/modules/pattern-fill.js"></script>
+<script src="https://code.highcharts.com/highcharts-more.js"></script>
```
hubertkozik, line 4: All of our demos have first imports of `highcharts` and `highcharts-more` files, then additional modules and like Kacper said we should use also exporting and accessibility modules (please check also if exporting works fully without any bugs).

samples/highcharts/demo/polar-advanced/demo.js
```diff
@@ -0,0 +1,351 @@
...
+            max: 2
+        },
+        {
+            minColor: '#A0FAB4',
+            maxColor: '#ACB7FA',
+            showInLegend: false,
+            ...monthExtremes
+        }
+        ],
+        pane: [{
```
hubertkozik, line 158: Mixed indentation (should be `}, {`).

samples/highcharts/demo/polar-advanced/demo.js
hubertkozik: Can't we just set a static value in pixels in `legend.y` property?

samples/highcharts/demo/polar-advanced/demo.js
```diff
@@ -0,0 +1,351 @@
...
+                centerInCategory: true,
+                enableMouseTracking: false
+            }, {
+                name: 'Total',
+                type: 'columnrange',
+                data: scoreData,
+                xAxis: 1,
+                yAxis: 1,
+
+                shadow: true,
```
hubertkozik, line 315: Unnecessary empty line :)

samples/highcharts/demo/polar-advanced/demo.js
```diff
@@ -0,0 +1,351 @@
...
+                    letterSpacing: '0.1rem',
+                    color: '#444444'
+                },
+                text: 'Teams'
+            }
+        },
+
+        series: [
+            ...teamSeries,
+            {
```
hubertkozik, line 295: Unnecessary enter `, {`.

samples/highcharts/demo/polar-advanced/demo.js
```diff
@@ -0,0 +1,351 @@
...
+            borderWidth: 4,
+
+            itemStyle: { color: '#333333', fontSize: '0.86rem', letterSpacing: '0.032rem' },
+            y: this.plotSizeY,
+            width: '10%',
+            padding: 16,
+            maxHeight: '18%',
+
+            title:
+            {
```
hubertkozik, line 282: Remove one enter (`title: {`).

hubertkozik (changes requested):
I think we should also create a readme file ([example](https://github.com/highcharts/highcharts/blob/master/samples/highcharts/demo/polar/readme.md)).

samples/highcharts/demo/polar-advanced/demo.js
MarkusBarstad: Thanks for good feedback, Hubert!

The reason I have not done so is that a pixel-value will not work for all sizes, as the chart shrinks/grows the legend becomes missplaced. The legends Y-property does not have a percentage (yet but i will make a feature request as Kacper suggests) thus I chose to do it with the ResizeObserver

Originally I wanted to do it with a callback using HCs responsive system but since I want this rule on every size, I had to put in no conditions, which caused recursion error for some reason :S

Edit:
Putting a minimum condition that all charts satisfy (such as minWidth: 1) would cause the chart not to update the legends position on resize so that too was not an option.

MarkusBarstad: Temp fix until ive made a feature req for legend.y as percentage

nasin:
Hi, I added my styling to the demo.

samples/highcharts/demo/polar-advanced/demo.html
```diff
@@ -0,0 +1,104 @@
...
+<!-- <script src="https://code.highcharts.com/themes/brand-light.js"></script> -->
+
+<figure class="highcharts-figure center">
+    <div id="container"></div>
+    <p class="highcharts-description">
+        An advanced polar chart, showing the performance of three sales teams throughout a month
+    </p>
+</figure>
+
+<pre id="data" style="display: none">
```
KacperMadej, line 17: This inline style throws an error in utils demo view (similar works fine in JSFiddle so I do not think we should care) and importantly, CSS already sets this style so a duplicate.

samples/highcharts/demo/polar-advanced/demo.html
```diff
@@ -0,0 +1,104 @@
+<script src="https://code.highcharts.com/highcharts.js"></script>
+<script src="https://code.highcharts.com/highcharts-more.js"></script>
+<script src="https://code.highcharts.com/modules/coloraxis.js"></script>
+<script src="https://code.highcharts.com/modules/pattern-fill.js"></script>
+<script src="https://code.highcharts.com/modules/exporting.js"></script>
+<script src="https://code.highcharts.com/modules/export-data.js"></script>
+<script src="https://code.highcharts.com/modules/accessibility.js"></script>
+<!-- <script src="https://code.highcharts.com/themes/brand-light.js"></script> -->
```
KacperMadej, line 8: I support removing theme because there's an option to view the demos with selected themes for the users making default and the used theme (brand light) the same.
image

KacperMadej (changes requested):
Looks better and better 👍 
Assuming there will be a clean up after the design changes only minor notes from me.

MarkusBarstad:
> Hi, I added my styling to the demo.

Looks awesome!

hubertkozik (changes requested):
I have tested this chart and it seems to me, that we cannot use it on mobile devices. The chart is no scaling below 500px. I believe it should be responsive for mobile devices.
Zrzut ekranu 2023-09-4 o 11 46 38

MarkusBarstad:
> I have tested this chart and it seems to me, that we cannot use it on mobile devices. The chart is no scaling below 500px. I believe it should be responsive for mobile devices. Zrzut ekranu 2023-09-4 o 11 46 38

You are right, @hubertkozik , it is a problem related to panes I believe (not sure though, and not only panes, since your chart worked out). 

Scaling now works because I specifically use responsive units in `demo.css` setting the dimensions of the outer `#container` you can see the difference in the latest commit.

nasin (approved):

samples/highcharts/demo/polar-advanced/demo.html
```diff
@@ -0,0 +1,103 @@
...
+{"x": 22, "low": 0, "high": 157, "week": 4, "avg": 52, "highscore": 94, "topEarner": "Sofia" } ,
+{"x": 23, "low": 0, "high": 112, "week": 4, "avg": 37, "highscore": 63, "topEarner": "Sofia" } ,
+{"x": 24, "low": 0, "high": 158, "week": 4, "avg": 52, "highscore": 67, "topEarner": "Sofia" } ,
+{"x": 25, "low": 0, "high": 239, "week": 4, "avg": 79, "highscore": 96, "topEarner": "Ulambaator" } ,
+{"x": 26, "low": 0, "high": 175, "week": 4, "avg": 58, "highscore": 82, "topEarner": "Ulambaator" }
+]]
+</pre>
+
+
+
```
hubertkozik, line 103: There are too many enters here :)

samples/highcharts/demo/polar-advanced/demo.js
```diff
@@ -0,0 +1,479 @@
...
+        pane: 2,
+        tickInterval: 0.25,
+        gridLineWidth: 0,
+        gridLineColor: hcColor1.brighten(0.05).get('rgba'),
+        min: -3,
+        max: 1,
+        title: null,
+        ...noLabelProp
+    }],
+    legend: {
```
hubertkozik, line 286: When I'm hovering any legend item it applies the default `itemHoverStyle` with black font color, I think we should change it.

hubertkozik (changes requested):
Other comments below. Also, I discovered some differences in "Week" labels positions: 'Week 1' is closer to the edge than 'Week 3'.

samples/highcharts/demo/polar-advanced/demo.js
```diff
@@ -0,0 +1,479 @@
...
+        chart.subtitle.element.style.opacity = 1;
+    },
+    colors = Highcharts.getOptions().colors,
+    asHcColor = hexval => new Highcharts.Color(hexval),
+    // Declaring colors used more than once
+    hcColor0 = asHcColor(colors[0]),
+    hcColor1 = asHcColor(colors[1]),
+    hcColor5 = asHcColor(colors[5]),
+    hcColor8 = asHcColor(colors[8]),
+    hcColor9 = asHcColor(colors[9]),
```
TorsteinHonsi, line 34: This is kind of subjective, but instead of a list of named constants with numbers in them, in this case I would just do a map into a modified array.

```js
colors = Highcharts.getOptions().colors.map(color => new Highcharts.Color(color));
```

Or, if I am not mistaken, we can shorten it using `Color.parse`:

```js
colors = Highcharts.getOptions().colors.map(Highcharts.Color.parse);
```

samples/highcharts/demo/polar-advanced/demo.js
```diff
@@ -0,0 +1,479 @@
...
+                'font-size: 2em; padding: 0.8rem;' +
+                'border: 0 outset {series.color}; border-block-end:' +
+                '0 outset {series.color};"><b>{series.name}</b></span>'
+            ),
+            pointFormat: (
+                '<span style="margin-top:11em; position: absolute;' +
+                'font-size: 1em;"><span style="width:100%;' +
+                'text-align:center;">Daily Sales:</span></br>' +
+                '<span style="line-height:3em;width:100%;font-size:2em;' +
+                'text-align:center;">{point.z}</span>'
```
TorsteinHonsi, line 78: Replace the inline styles with CSS class names perhaps? It would make these format strings look friendlier.

samples/highcharts/demo/polar-advanced/demo.js
```diff
@@ -0,0 +1,479 @@
...
+                        dy: '2.8%'
+                    }
+                }
+            },
+            x: index + 1,
+            y: 1.5
+        })),
+    asColFieldStr = str => (
+        '<span class=\"col-display-fieldwrap\">' +
+        '<span style=\"color:{point.color}; font-size: 1em;\">●</span> ' +
```
TorsteinHonsi, line 122: Escaped double quotes not needed in a single-quoted string

TorsteinHonsi (changes requested):
Very cool demo!

* Do these two outermost grid lines have a function? Or are they just a result of the y axis settings?
Skjermbilde 2023-09-08 kl  15 27 16

* On the mobile, the font-size for the legend is so small that it is hardly readable. But there is actually room for the legend to be a bit bigger and spill out over the purple and blue band. If we make the legend background semi-opaque, one can also see the bands through. Besides, there is empty space to the right of Ulanbator, so it should be possible to use larger fonts.

* The text is sometimes spilling over the center ring: 
Skjermbilde 2023-09-08 kl  15 31 00

* And of course, we need comments in order to explain what we're doing. In general, the code for a demo doesn't need to be tight and super-optimized (DRY) like the Highcharts source code should be. The purpose of the Highcharts source code is to be fast and compact. The purpose of the demo is to educate our users, and to demonstrate that it is easy to create cool visualizations with our product.

samples/highcharts/demo/polar-advanced/demo.js
MarkusBarstad: I agree with your preference here, but, if i recall correctly, I had problems with the constructor when putting them all in an array using the "map" function.

I will try your second suggestion!

samples/highcharts/demo/polar-advanced/demo.js
MarkusBarstad: True! I was a bit unsure about creating classes for the css-configurations which only occurred once, but I agree, it will look a lot nicer.

MarkusBarstad:
> Very cool demo!
> 
> In addition to the inline comments:
> 
> * Do these two outermost grid lines have a function? Or are they just a result of the y axis settings?
Just an artifact from originally trying to fit the bubbles in, but should be doable with pane-sizing instead.
Also a weird artifact is the needlessly big units of y-value for the bubble series, I will also change that.

>   Skjermbilde 2023-09-08 kl  15 27 16
> * On the mobile, the font-size for the legend is so small that it is hardly readable. But there is actually room for the legend to be a bit bigger and spill out over the purple and blue band. If we make the legend background semi-opaque, one can also see the bands through. Besides, there is empty space to the right of Ulanbator, so it should be possible to use larger fonts.
> * The text is sometimes spilling over the center ring:
>   Skjermbilde 2023-09-08 kl  15 31 00
Good catches! 👍 I will adjust the `width`-property as far as I can without hindering the visual datapoints

> * And of course, we need comments in order to explain what we're doing. In general, the code for a demo doesn't need to be tight and super-optimized (DRY) like the Highcharts source code should be. The purpose of the Highcharts source code is to be fast and compact. The purpose of the demo is to educate our users, and to demonstrate that it is easy to create cool visualizations with our product.
That makes sense and is a good heuristic to code after for the demos. 

I will admit I might have coded myself blind a tad with this demo but I will try to have some workmates read my code + comments once I have some comments!

MarkusBarstad:
> Other comments below. Also, I discovered some differences in "Week" labels positions: 'Week 1' is closer to the edge than 'Week 3'.

Good catch!

samples/highcharts/demo/polar-advanced/demo.js
MarkusBarstad: Also, @TorsteinHonsi , regarding this:
I do not see a simple way to avoid the style attribute when using HCs format-string variables (`series.color`, `point.x.color` etc), so I assume its okay to keep styles using colors (there are a couple of cases involving `border-*`, there is the general `color` and `background-color`)

TorsteinHonsi: Yes, that's fine!

samples/highcharts/demo/polar-advanced/demo.js
MarkusBarstad: Done!

samples/highcharts/demo/polar-advanced/demo.js
MarkusBarstad: It did work!

MarkusBarstad:
> Other comments below. Also, I discovered some differences in "Week" labels positions: 'Week 1' is closer to the edge than 'Week 3'.

I have tried to adjust it, but I admit my logic looks a tad dodgy, and I am not sure it is 100% exact!

Any ideas x)? I guess I can spin up some sliders + a grid with the `renderer` and try to adjust them like that!

TorsteinHonsi (changes requested):
Looks great! Two comments regarding responsive rules.

samples/highcharts/demo/polar-advanced/demo.js
```diff
@@ -0,0 +1,553 @@
...
+                condition: {
+                    minWidth: 0
+                },
+                chartOptions: {
+                    legend: {
+                        borderRadius: 18,
+                        borderWidth: 0.7,
+                        padding: 4,
+                        symbolPadding: 2,
+                        symbolHeight: 1,
```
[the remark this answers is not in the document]
MarkusBarstad: Yeah, my mistake there, I see what you are saying! I will eliminate dupes and probably just keep the `symbolHeight >= 8`

samples/highcharts/demo/polar-advanced/demo.js
```diff
@@ -0,0 +1,553 @@
...
+            custom: {
+                textSizeClass: 'small-size'
+            }
+        }
+    },
+    responsive: {
+        rules: [
+            {
+                condition: {
+                    minWidth: 0
```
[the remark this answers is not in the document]
MarkusBarstad: Will do! 

I originally did do this, however (if i recall correctly) it seemed like those styles were not being re-applied when resizing up, then down again 🤔 

I will look into i again though!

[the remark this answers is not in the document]
TorsteinHonsi: > it seemed like those styles were not being re-applied when resizing up, then down again

Ok, if that's true it would be a bug. Let me know what you find out.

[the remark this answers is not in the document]
TorsteinHonsi: I did a basic test with `legend.borderWidth`, it seems to work correctly: https://jsfiddle.net/highcharts/z4m369kx/

MarkusBarstad:
@TorsteinHonsi Alright, let me know if it is too messy now! 

There were some props for the legend i had forgotten to play with but which could have suited my purposes (namely margin and lineheight), I used some of lineHeight, but only for the 2 smallest sizes.

Also, in regards to responsive styles not being displayed when scaling down, as i mentioned earlier,
this is what i mean:

Now, when we start at some bigger size, then scale down to 360px (the page being 360px that is) it looks like this:
Skjermbilde 2023-09-18 kl  13 55 44


But if we start from the same size (by hitting CTRL + R) you can see that the responsive styles are being properly applied:
Skjermbilde 2023-09-18 kl  13 56 03

TorsteinHonsi:
Weird. I get a feeling it has something to do with the multiple levels of responsive rules, but I am not able to reproduce it in a controlled demo. This is what I am trying: https://jsfiddle.net/highcharts/z4m369kx/1/

TorsteinHonsi:
@MarkusBarstad In your fiddle the chart width is still 400, so the responsive rules don't change. Removing `chart.width` makes it behave like you expected: https://jsfiddle.net/highcharts/Lz5swae4/

TorsteinHonsi:
I isolated the responsive issue to something related to the color axis. Reported here: https://github.com/highcharts/highcharts/issues/19777

MarkusBarstad:
Alright @TorsteinHonsi , master is merged into the branch

TorsteinHonsi:
Nice! Now pending approvals from @hubertkozik and @KacperMadej

samples/highcharts/demo/polar-advanced/demo.js
```diff
@@ -0,0 +1,527 @@
...
+        '<span class="symbolSize" ' +
+        'style="color:{point.color};">●</span> ' + str + '</span>'
+    ),
+
+    // We create our teams, 1 serie per team.
+    teamNames = ['Ulambaator', 'Sofia', 'Asmara'],
+    teamColors = [
+        colors[9].tweenTo(colors[0], 0.25),
+        colors[9].tweenTo(colors[8], 0.65),
+        colors[9].tweenTo(colors[3], 0.85)
```
KacperMadej, line 49: `colors[9]` fails with an error for themes "Brand Light" and "Brand Dark" as there are only 8 colors for each theme.

For getting the colors array, something like this works:
`[...Highcharts.getOptions().colors, ...(Highcharts.getOptions().colors.length < 9 ? new Array(9 - Highcharts.getOptions().colors.length).fill('#bada55') : [])]` (and then map it with `parse` as in the 4th line) but readability could be improved and the place holder color picked so the chart looks good.

KacperMadej (changes requested):
One error on theme change is explained with more details in the code.
Other than that - LGTM 👍

samples/highcharts/demo/polar-advanced/demo.html
hubertkozik: Still too many enters? :D

hubertkozik (changes requested):
There is a wrong hover effect while hovering the legend items (font size is changing from bigger to lower).

samples/highcharts/demo/polar-advanced/demo.js
TorsteinHonsi: I'd prefer that we just add more colors to those themes.

KacperMadej: @TorsteinHonsi Should this be resolved in this PR or will it be handled elsewhere?

KacperMadej (approved):

hubertkozik (approved):
Also approving, but I think we should handle the issue with the colors array also.

samples/highcharts/demo/polar-advanced/demo.js
TorsteinHonsi: Doing it in this PR is fine by me.

TorsteinHonsi: @KacperMadej On second thought, I don't think we should change the theme. Chances are there are themes out there with any number of colors, so our demo should support that. We use the rest operator as safeguard, something like this:

`colors[9 % colors.length].tweenTo(n)`

PR #19593, opened 2023-08-21
Initial benchmark setup

goransle:
See [README file](https://github.com/highcharts/highcharts/blob/a1c3cfd4b4bdad664c6fb54eb2665e5ae9aae106/test/ts-node-unit-tests/readme.md) for instructions on how to run locally.


test/ts-node-unit-tests/benchmarks/DataPool-CSV-constructor.bench.ts
```diff
@@ -0,0 +1,66 @@
...
+            }
+            csv += '\n';
+        }
+
+        return csv;
+    }
+
+    return {
+        fileName: `${size}-rows.csv`,
+        func: generateCSV.bind(undefined, [size, 5])
```
karolkolodziej, line 31: Shouldn't we also test it with other column values?

karolkolodziej (approved):

test/ts-node-unit-tests/benchmarks/DataPool-CSV-constructor.bench.ts
goransle: Yes and no. For now this test is mostly to confirm that the DataPool constructor doesn't do any loading, so rows vs columns should not matter too much. But it might be neater with a square dataset 🤔 

I think I will add a test for the DataTable with increasing # of columns though 👍

bre1470 (approved):
Nice! 👍

PR #19596, opened 2023-08-21
dash/bugfix-mousewheel-climate-demo

sebastianbochan:
Fixed mousewheel problem in the climate demo.

------

I debugged the problem that is caused by the stock.js module. In the stock, we have the MouseWheelZoom module, enabled by default.

The solution is to disable this option by:

```
zooming: {
    mouseWheel: {
        enabled: false
    }
}
```


karolkolodziej (approved):

pawellysy (approved):
LGTM

Vazonik (approved):
LGTM

bre1470 (approved):

PR #19600, opened 2023-08-22
enhancement/karma-prepareshot-mouseover

kamil-musialowski:
During the visual comparison, on some charts tooltip was displayed on point that was not inside the plot.

### DIFF
--- a/test/karma-setup.js
+++ b/test/karma-setup.js
@@ -513,6 +513,7 @@ Highcharts.prepareShot = function (chart) {
                 if (
                     points[i] &&
                     !points[i].isNull &&
+                    points[i].isInside &&
                     !( // Map point with no extent, like Aruba
                         points[i].shapeArgs &&
                         points[i].shapeArgs.d &&


kamil-musialowski:
@TorsteinHonsi What do you think about such addition? Of course the visual tests have changed, but it seems like they didn't work properly before.

kamil-musialowski:
@TorsteinHonsi We are down to 196 changes from 496, however most of them are still quite significant. What do you think about the whole idea? Should we try to figure something out for that, or just abandon it and remove the PR?

TorsteinHonsi:
@kamil-musialowski I think the idea is good, and we should follow through. But I won't merge right away. Instead, we should merge this immediately before the next release, so that the benchmark for the next release is based on these changes.

There's another problem though. Most cartesian charts are improved with the new logic. But in some cases, like polar charts and wind barbs, the tooltip is gode, because the `isInside` flag is not used (why doesn't it work with polar charts?).

Anyway, I think we should try testing that the point's x value (if any) is between the x-axis (if any) min and max. That would probably work with polar and wind barbs too. Series like pies don't have an x-axis, so such cases all points must pass. Something like this (untested):

`typeof points[i].x !== 'number' || !series.xAxis || (points[i].x >= series.xAxis.min && points[i].x <= series.xAxis.max)`

kamil-musialowski:
@TorsteinHonsi I've commited a variation of your code, and now we don't have any visual differences - and I'm kinda confused why, since all of the if checks seem to be logical.

TorsteinHonsi:
That is confusing. Anyway, I'll make a note to merge this PR directly before next release.

TorsteinHonsi (approved):

PR #19607, opened 2023-08-22
bugfix/19606-for-export-property

### LINKED ISSUE #19606: Property `forExport` still not visible in the TypeScript declaration of `SVGRenderer`
#### Expected behaviour

For the property [`forExport`](https://github.com/highcharts/highcharts/blob/063e4643872af29c7cf7299587b3d287e710eac9/ts/Core/Renderer/SVG/SVGRenderer.ts#L208) to be visible on the TypeScript declaration for the class `SVGRenderer`

#### Actual behaviour

It is still not visible, despite #19578

#### Product version
highcharts 11.1.0

### LINKED ISSUE #19612: property `forExport` on class `SVGRenderer` is not included in the generated .d.ts files
#### Expected behaviour

If I check out the code and run `npm run build`, I expect the property `forExport` of the class `SVGRenderer` to be present in the generated `code/highcharts.d.ts` file.

#### Actual behaviour

It is not.

#### Product version
Highcharts, latest.

#### Additional remarks

An attempt was made to fix this issue in #19578 , but the build output could not be checked. However, since [this commit](https://github.com/highcharts/highcharts/commit/f17638d273f121df92aab387445e038e1c8430d4), the build output could be checked. It now appeared that #19578 had not fixed the issue.

emilefokkema:
This fixes #19612 . (Though I have no idea why.)


karolkolodziej (approved):

TorsteinHonsi:
@bre1470 Any idea why this works? Preferably I would want to avoid setting it to `void 0` because it is a waste of bytes.

bre1470:
Yes, for now we do not use native declarations and instead use JSDoc to extract doclets with their codelines from the JS file. Though in JS files TypeScript uses optimizations to remove any codeline with preceding doclet, that does not has an immediate impact on the JS code. We solve this issue by moving the doclet to the actual assignment which is also part of the JS code, usually found in some `init` function.

So in this PR it would be better to move the doclet to [this line](https://github.com/highcharts/highcharts/blob/14b3265319c8fa071eee2065d6427d36b3f4afa4/ts/Core/Renderer/SVG/SVGRenderer.ts#L319) and keep the property definition untouched.

emilefokkema:
Let me see if it works and if so, I'll update the pr accordingly.

emilefokkema:
@bre1470 Excellent, it works! I updated the pr.

TorsteinHonsi (approved):

bre1470:
@emilefokkema Thank you!

PR #19611, opened 2023-08-23
bugfix/19253-boost-offline-exporting

### LINKED ISSUE #19253: Offline exporting fails when boost is active
#### Expected behaviour
When exporting PDF offline, I expect a file with the same data as the displayed trend.

#### Actual behaviour
Once the Boost module is active, the exported PDF-chart is empty. If the boost module is not active, the export basically works. However, if a PDF is exported, an error message appears in the Chrome log for internal parameters that were not found.

Standard: https://jsfiddle.net/PadWorks/stn6w9a7/26/
Boost Active: https://jsfiddle.net/PadWorks/759csfgk/9/

#### Product version
Highcharts JS v11.1.0

#### Affected browser(s)
Chrome

jakubSzuminski:
Fixed #19253, Offline Export did not work for boosted series.

---
**Demo of the problem:** https://jsfiddle.net/BlackLabel/3y6uwf97/


TorsteinHonsi (commented):
Thanks!

A couple of questions:

Where does `fedropshadow` come from? Do you know why we need both `feDropShadow` and `fedropshadow`?  Or to put it a different way - can we eliminate the lowercase version somewhere?

This change from using `setAttributeNS` to plain `setAttribute` - did you manually test other demos for offline exporting?

jakubSzuminski:
1. There's no `fedropshadow` in our code, but to export PDFs we're using 2 external libraries (`svg2pdf.js` and `jspdf.js`). My guess is that they're performing some kind of DOM operations and that's where the `fedropshadow` comes from (as it's not an issue when exporting to other formats when no external libraries are used). Maybe I should add the `fedropshadow` to `AST.allowedAttributes` before exporting to pdf and remove it after export to PDF? 

2. `setAttributeNS` and `getAttributeNS` are used only for images there so it should not impact any other elements while performing an offline export. From what I've tested, these changes do not cause any troubles as they do not have impact on elements other than images: 
   - https://jsfiddle.net/BlackLabel/odkb4yv2/
   - https://jsfiddle.net/BlackLabel/46d1y8eu/
   - I've ran all the examples listed in `./samples/highcharts/exporting/offline-...` and everything seems to work as expected

TorsteinHonsi:
> Maybe I should add the fedropshadow to AST.allowedAttributes before exporting to pdf and remove it after export to PDF?

At least I think we should leave it out of the core. Removing it is not important though. I see two options:
* Have the offline-exporting module add it to the allowed attributes. Or:
* Have the offline-exporting module to an `svg.replace(/fedropshadow/g, 'feDropShadow')` on the results of the external library.

TorsteinHonsi (approved):

PR #19614, opened 2023-08-23
bugfix/19314-tooltip-outside-wrong-size

### LINKED ISSUE #19314: Tooltip outside on column chart cut off
#### Actual behaviour

When setting tooltip.outside to true on a column chart, the bottom of the tooltip is cutoff by 1 pixel.

https://jsfiddle.net/ty3vgpnr/

#### Product version
Highcharts 11+

#### Affected browser(s)
Chrome, Firefox

jedrzejruta:
Fixed #19314, tooltip chevron was cut off when `outside` was enabled.

Shadow `offsetX` and `offsetY` wasn't calculated in Tooltip's size, which resulted in cutting it off.


TorsteinHonsi (approved):
Thanks!

I did some simplification of the logic to keep the complexity and file size down. Non-default shadow offsets are quite rare in my experience, so I added hard-coded values instead. We can reconsider in the future if users report it.

Now pending CI tests.

jedrzejruta:
> I did some simplification of the logic to keep the complexity and file size down. Non-default shadow offsets are quite rare in my experience, so I added hard-coded values instead. We can reconsider in the future if users report it.

Of course - if users start reporting it, we can always come back to this issue and have a discussion, Thanks!

PR #19625, opened 2023-08-25
bugfix/19524-treegraph-set-data-links

### LINKED ISSUE #19524: Treegraph leaves unnecessary links after series.setData
Hello!

I met the problem that if we dynamically set data for a treegraph series by `series.setData` method, the chart leaves links that must be deleted, or they are links between unexisting nodes.
If we do that by `series.update({ data })` all works fine but links lost animation.

#### Expected behaviour
The chart leaves only links between existing nodes if data is updated by `series.setData` method.
Or another way if data is updated by `series.update` method, keep links animation.

#### Actual behaviour
After updating data by `series.setData` method, the chart leaves unnecessary links. And after some updates throw error:
```
highstock.src.js:45857 
Uncaught TypeError: Cannot read properties of undefined (reading 'y')
    at y.g (highstock.src.js:45857:1)
    at highstock.src.js:1748:1
    at f (highstock.src.js:1745:1)
    at y.generatePoints (highstock.src.js:32552:1)
    at y.t (highstock.src.js:47125:1)
    at y.<anonymous> (highstock.src.js:914:1)
    at y.h (boost.src.js:2976:1)
    at b.<computed> (highstock.src.js:912:1)
    at y.translate (highstock.src.js:32693:1)
```

[Reproducing the bug](https://jsfiddle.net/0ob571Lz/18/)

image
image

#### Product version
Highcharts: 11.1.0

jakubjanuchta:
Fixed #19524, treegraph links weren't dissappearing after calling `setData`.


TorsteinHonsi (approved):
Thanks, looks good!

CI failed because of master error. I merged in now, pending successful CI run.

PR #19630, opened 2023-08-25
dash/docs-datagrid-component

Vazonik:
Added additional DataGrid Component docs article.


docs/dashboards/datagrid-component.md
```diff
@@ -0,0 +1,104 @@
+DataGrid Component
+===
+
+The DataGrid Component can be placed inside a data cell to allow the data visualization in the form of an editable data grid.
```
karolkolodziej, line 4: `a data cell` we don't use this phrase. It's just `a cell` or if you would like to be specific it's `a dashboard's cell`.

docs/dashboards/datagrid-component.md
```diff
@@ -0,0 +1,104 @@
...
+
+<iframe style="width: 100%; height: 470px; border: none;" src="https://www.highcharts.com/samples/nonav/dashboards/datagrid-component/datagrid-options" allow="fullscreen"></iframe>
```
karolkolodziej, line 6: The link is incorrect.

docs/dashboards/datagrid-component.md
```diff
@@ -0,0 +1,104 @@
...
+            type: 'CSV',
+            options: {
+                csvURL: 'https://demo-live-data.highcharts.com/updating-set.csv'
+            }
+        }]
+    }
+    ```
+    Click [here](https://www.highcharts.com/docs/dashboards/data-handling) to read more about the data handling.
+
+5. Once you have the data and where to place your component, you can define it as below. In the `cell` option, we're passing the ID of the cell, which we defined in point 3, and the ID of the connector with the data in the `connector.id` option (point 4). In order for the component to be created as a DataGrid, it was set with the `type` option to `'DataGrid'`.
```
karolkolodziej, line 62: What if I don't have a connector or I don't want to use it? How to create a simple grid without?

karolkolodziej (changes requested):
A few things missing:
- API reference, 
- math modifiers
- input
- styling and cell formating
- sync

docs/dashboards/datagrid-component.md
sebastianbochan: `allow user to visualize the data in the editable table` ?

Just a suggestion, because for me, the description is too complicated.

docs/dashboards/datagrid-component.md
```diff
@@ -0,0 +1,104 @@
...
+            rows: [{
+                cells: [{
+                    id: 'dashboard-col-0'
+                }]
+            }]
+        }]
+    }
+    ```
+
+4. You will also need some data that you could display in your DataGrid component. For this purpose, you can, for example, define a connector that, in this case, downloads CSV data from the example URL.
```
sebastianbochan, line 48: IMO, would be better to show user how to use in the simplest way and then add extra example with the connector.

(You can hardcode the CSV value).

sebastianbochan (changes requested):

docs/dashboards/datagrid-component.md
```diff
@@ -0,0 +1,104 @@
...
+    ```ts
+    import * as Dashboards from '@highcharts/dashboards';
+    import DataGrid from '@highcharts/dashboards/es-modules/DataGrid/DataGrid';
+    import DataGridPlugin from '@highcharts/dashboards/es-modules/Dashboards/Plugins/DataGridPlugin';
+
+    DataGridPlugin.custom.connectDataGrid(DataGrid);
+    Dashboards.PluginHandler.addPlugin(DataGridPlugin);
+    ```
+
+2. You also need to import the styles.
```
goransle, line 28: Wondering if we should specify "in your main CSS file" here? 🤔

goransle: Also applies to the HTML and JS I guess, but methinks it's easier to confuse the CSS import for JS

docs/dashboards/datagrid-component.md
Vazonik: Scheduled after bug fix with the [dataTable](https://api.highcharts.com/dashboards/#interfaces/DataGrid_DataGridOptions.DataGridOptions-1#dataTable) option, which doesn't work for the DataGrid Component.

docs/dashboards/datagrid-component.md
Vazonik: Scheduled after bug fix in the [dataTable](https://api.highcharts.com/dashboards/#interfaces/DataGrid_DataGridOptions.DataGridOptions-1#dataTable) option, which doesn't work for the DataGrid Component. At the moment changed from `csvURL` to `csv`.

pawellysy (approved):

sebastianbochan (approved):

karolkolodziej (approved):

PR #19631, opened 2023-08-25
tools/test-docs-win-path

KacperMadej:
Changed path to relative so it would work on Windows too.


KacperMadej:
Closing in favor of the https://github.com/highcharts/highcharts/pull/19637

PR #19639, opened 2023-08-28
bugfix/19148-NaN-x-value

### LINKED ISSUE #19148: Area charts with NaN value do not render actual area (just line is rendered)
I'm upgrading from 7.2.0 to the latest (11.1.0). One issue appeared, which seems to me to be buggy behavior. 
I know, NaN is unexpected value type. But the issue happens only in area chart type, all other charts, even having NaN in series data, are rendered correctly.

@TorsteinHonsi , the behavior has been changed here - https://github.com/highcharts/highcharts/commit/e70baaba67d36aa89b2f724fcf7b9c1e04b0b02f#diff-781d34161016cd83d4211d594c56c68a8c1802de3c17eae5158ebef0f2b5e881L1888 (line 1888 if it doesn't scroll automatically).
I just want to make sure that the change is intentional. Or probably the bug has been introduced there.  

#### Expected behaviour
Area charts containing NaN values render area

#### Actual behaviour
Only line is rendered in case there's NaN in data

https://jsfiddle.net/vpe87cx1/40/
There are 2 identical charts there with 1 difference: last series data point varies from NaN to null. 

#### Product version
11.1.0

#### Affected browser(s)
Latest

KamilKubik:
Fixed #19148, NaN x value disabled area chart graphic elements.

@TorsteinHonsi the fix provided here #19148 seems to be working fine. I added a check for the date object only and moved the `isNull` assignment below the auto incrementation for the `undefined` x values, which I think was crucial to always base on defined ones.

### DIFF
--- a/samples/unit-tests/point/point/demo.js
+++ b/samples/unit-tests/point/point/demo.js
@@ -795,3 +795,24 @@ QUnit.test('#14623: colorIndex Series.update()', assert => {
         'Point.colorIndex should be correct'
     );
 });
+
+QUnit.test('NaN x value (#19148).', assert => {
+    const chart = Highcharts.chart('container', {
+        series: [
+            {
+                type: 'area',
+                data: [1, 2, [NaN, 3], 4, 5]
+            }
+        ]
+    });
+
+    chart.series[0].points.forEach(point => {
+        if (Number.isInteger(point.x)) {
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks, looks good!

PR #19640, opened 2023-08-28
bugfix/19621-a11y-broken-legend-kbd-navigation

### LINKED ISSUE #19621: Navigating beyond the end of legend items with arrow keys throws an error.
#### Expected behaviour

No error should be thrown.

#### Actual behaviour

The console in the DevTools shows this output:
```
Uncaught TypeError: Cannot read properties of undefined (reading 'legendItem')
    at ea.b [as highlightLegendItem] (accessibility.src.js:3846:65)
    at G.onKbdArrowKey (accessibility.src.js:3778:29)
    at u.keyCodeMap (accessibility.src.js:3741:40)
    at q.run (accessibility.src.js:1302:32)
    at a.onKeydown (accessibility.src.js:3217:38)
    at HTMLDivElement.<anonymous> (accessibility.src.js:3060:71)
```

https://jsfiddle.net/v8fp7z34/
Navigate through the legend items using the arrow keys and press the right or downwards arrow key when focusing the last legend item or press the left or upwards arrow key when focusing the first legend item.

#### Product version

Highcharts 11.1.0

#### Affected browser(s)

Firefox, Chrome, Edge

MarkusBarstad:
Fixed #19621, errors instead of expected behavior when using arrow-keys to navigate first and last items in legend.

---
  - https://app.asana.com/0/0/1205344683223003


ts/Accessibility/Components/LegendComponent.ts
```diff
@@ -535,21 +535,32 @@ class LegendComponent extends AccessibilityComponent {
...
+                rawDirectionalInput < numItems
+            ) ? direction : (
+                    wrapAround && (
+                        (numItems - 1) * (direction > 0 ? -1 : 1)
+                    ) || 0
+                )
+            ),
+            res = chart.highlightLegendItem(
+                highlightedLegendItemIx + adjustedDirection
+            ) || void 0;
```
hubertkozik, line 556: I think we don't need that `OR` with `undefined` value. After removing that everything works as expected.

hubertkozik (changes requested):
Can you add a unit test also? If it won't be possible to simulate arrows in QUnit, you can make a test of a function (`Highcharts._modules['Accessibility/Components/LegendComponent.js'].prototype.onKbdArrowKey`) with proper arguments and check the result of it. Also, I think it would be good to ask Oystein for CR, as he is Specialist in Accessibility :D

ts/Accessibility/Components/LegendComponent.ts
```diff
@@ -535,21 +535,32 @@ class LegendComponent extends AccessibilityComponent {
...
             return response.success;
         }
 
-        if (
-            numItems > 1 &&
-            a11yOptions.keyboardNavigation.wrapAround
-        ) {
+        if (wrapAround && numItems > 1) {
             keyboardNavigationHandler.init(direction);
             return response.success;
```
KacperMadej, line 565: Nitpick:
The code structure could be improved - DRY. Now there's:
```
        if (res) {
            this.highlightedLegendItemIx += adjustedDirection;
            return response.success;
        }

        if (wrapAround && numItems > 1) {
            keyboardNavigationHandler.init(direction);
            return response.success;
        }

        return response.success;
```

Always `return response.success`. How about something like `if(A) { B; } else if (C) { D; } return S;`

ts/Accessibility/Components/LegendComponent.ts
MarkusBarstad: Good catch!

ts/Accessibility/Components/LegendComponent.ts
```diff
@@ -535,21 +535,32 @@ class LegendComponent extends AccessibilityComponent {
...
+                keyCode === keys.left || keyCode === keys.up
+            ) ? -1 : 1,
+            highlightedLegendItemIx = this.highlightedLegendItemIx,
+            rawDirectionalInput = highlightedLegendItemIx + direction,
+            adjustedDirection = ((
+                rawDirectionalInput > -1 &&
+                rawDirectionalInput < numItems
+            ) ? direction : (
+                    wrapAround && (
+                        (numItems - 1) * (direction > 0 ? -1 : 1)
```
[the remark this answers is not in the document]
MarkusBarstad: Nice catch x)!

ts/Accessibility/Components/LegendComponent.ts
MarkusBarstad: Good suggestion!

Also regarding DRY-principle, a thought I had while solving this bug; 

Just as a basic example, I believe in this function, `chart` is referenced twice during var-declaration at the start of the function, but now with the de-structuring syntax, we could easily limit this to one such reference. 

This will really break the old syntactic style of var-declaration, but it will limit property lookup, and thus, in theory improve performance if one can look up a property only once. 

Since there is a lot of property lookups in a big project like HC, should we move towards destructuring?

MarkusBarstad:
> Can you add a unit test also? If it won't be possible to simulate arrows in QUnit, you can make a test of a function (`Highcharts._modules['Accessibility/Components/LegendComponent.js'].prototype.onKbdArrowKey`) with proper arguments and check the result of it. Also, I think it would be good to ask Oystein for CR, as he is Specialist in Accessibility :D

Wow I seem to have scrolled past your comment, sorry about that! Good ideas, will do so tomorrow!

MarkusBarstad:
Alright guys, @oysteinmoseng  and @hubertkozik (Kacper is on hollidays 🏖️ ) 

Unit-test have been added, checking both wrap-around and non-wrap-around

hubertkozik (changes requested):
The test looks great to me! I am only thinking, do we need to create a separate test for that? I think, that we are trying to reduce the number of tests (creation of charts). Maybe it will be better if we move it somewhere to the existing test with an existing chart and just make some `assert`s there?

MarkusBarstad:
> The test looks great to me! I am only thinking, do we need to create a separate test for that? I think, that we are trying to reduce the number of tests (creation of charts). Maybe it will be better if we move it somewhere to the existing test with an existing chart and just make some `assert`s there?

Good point! Do you think [this one](https://github.com/highcharts/highcharts/tree/bugfix/19621-a11y-kbd-arrows/samples/unit-tests/accessibility/accessibility-options) is ok to use?

marvin19 (approved):

KacperMadej (approved):

oysteinmoseng (changes requested):
It works! But I do think we could simplify the fix quite a lot actually, so will invite to a quick pair programming session to look at it together

oysteinmoseng (approved):
🎉

TorsteinHonsi (approved):

PR #19645, opened 2023-08-30
dash/kpi-render-event-bug

Vazonik:
Added render event emitters to the KPI Component.

### DIFF
--- a/ts/Dashboards/Components/KPIComponent.ts
+++ b/ts/Dashboards/Components/KPIComponent.ts
@@ -332,7 +332,6 @@ class KPIComponent extends Component {
 
 
     public render(): this {
-        this.emit({ type: 'beforeRender' });
         super.render();
         this.updateElements();
 
--- a/ts/Dashboards/Components/KPIComponent.ts
+++ b/ts/Dashboards/Components/KPIComponent.ts
@@ -332,6 +332,7 @@ class KPIComponent extends Component {
 
 
     public render(): this {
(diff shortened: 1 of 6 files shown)


ts/Dashboards/Components/KPIComponent.ts
```diff
@@ -332,6 +332,7 @@ class KPIComponent extends Component {
 
 
     public render(): this {
+        this.emit({ type: 'beforeRender' });
```
bre1470, line 335: Not strictly related, but that is something, I already wondered in other components. We have `load` / `afterLoad`, `update` / `afterUpdate`, etc.; but here it is `beforeRender` / `afterRender`. We should be consistent and I think the best way to address this is actually to move the event as `render` into the base component class, so it gets emitted with the super call below like with the rest of events.
Any thoughts? @Vazonik , @sebastianbochan, @karolkolodziej , @pawellysy , @goransle

pawellysy (approved):
LGTM! Let's implement what @bre1470 suggested here as well tho.

ts/Dashboards/Components/KPIComponent.ts
karolkolodziej: I agree :)

bre1470 (approved):
Nice!

karolkolodziej (approved):

goransle (approved):

sebastianbochan (approved):

PR #19650, opened 2023-08-30
Dash/data modifiers article

pawellysy:
Created an article about data modifiers.

### DIFF
--- a/docs/dashboards/data-modifiers.md
+++ b/docs/dashboards/data-modifiers.md
@@ -74,6 +74,7 @@ dataModifier: {
 }
 ```
 
+### Example demos
 You can check the modifiers in work in given demos:
 * [DataGrid with MathModifier](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/dashboards/demo/datagrid-mathmodifier/)
 * [Parsing the data in CSV using different modifiers](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/data-tools/csv-modifiers)
--- a/docs/dashboards/data-modifiers.md
+++ b/docs/dashboards/data-modifiers.md
@@ -45,7 +45,7 @@ dataModifier: {
     }]
 }
 ```
-* Range - Allows to filter the data, which are exeeding the min and max properties. It works both for numbers as well as alphanumerical values. the `column` option specifies, from which column the values should be used to perform a filtering, and the `minValue` and `maxValue` specify the range limits. Here is the example:
+* Range - Allows filtering the data, which are exceeding the min and max properties. It works both for numbers as well as alphanumerical values. the `column` option specifies, from which column the values should be used to perform a filtering, and the `minValue` and `maxValue` specify the range limits. Here is an example:
 ```javascript
 dataModifier: {
     type: 'Range',
--- a/docs/dashboards/data-modifiers.md
+++ b/docs/dashboards/data-modifiers.md
@@ -29,7 +29,7 @@ connector: {
     }
 }
 ```
-The modifier is modifing the `table.modified` property of the connector, which leaves the original data unchanged, so that there is a way of restoring the original data if needed.
+The modifier modifies the `table.modified` property of the connector, which leaves the original data unchanged so that there is a way of restoring the original data if needed.
 
 ### What does each of the modifiers do?
 
--- a/docs/dashboards/data-modifiers.md
+++ b/docs/dashboards/data-modifiers.md
@@ -1,7 +1,7 @@
 Data Modifiers
 ===
 
-The data comes in different shapes and sizes from all over the internet, so there needs to be a way of adjusting its format so that it is accessible to the Dashboards. Instead of manually parsing the data, there is an option to use the `DataModifiers`. Those modifiers allow the user to modify the data on the initial load, as well as the end user to modify visible data through the GUI. Each of the modifiers is accessible through the main Dashboards package, except for the `MathModifier` which is the most complex one, and requires loading a separate module from `https://code.highcharts.com/dashboards/modules/math-modifier.js`
+The data comes in different shapes and sizes from all over the internet, so there needs to be a way of adjusting its format so that it is accessible to the Dashboards. Instead of manually parsing the data, there is an option to use the `DataModifiers`. Those modifiers allow the user to modify the data on the initial load, as well as the end user to modify visible data through the GUI. Each of the modifiers is accessible through the main Dashboards package, except for the `MathModifier` which is the most complex one, and requires loading a separate module from `https://code.highcharts.com/dashboards/modules/math-modifier.js`. Each of the modifiers is available to connect to every type of connector.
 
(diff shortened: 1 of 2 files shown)


docs/dashboards/data-modifiers.md
```diff
@@ -0,0 +1,80 @@
+Data Modifiers
+===
+
+The data comes in different shapes and sizes from all over the internet, so there needs to be a way of adjusting its format, so that it is accessible to the Dashboards. Instead of manually parsing the data, there is an option to use the `DataModifiers`. Those modifiers allow the user to modify the data on the initial load, as well as the end user to modify visible data through the GUI. Each of the modifiers is accessible through the main Dashboards package, except of the `MathModifier` which is most complex one, and requires loading a seperate module from `https://code.highcharts.com/dashboards/modules/math-modifier.js`
```
karolkolodziej, line 4:
```suggestion
The data comes in different shapes and sizes from all over the internet, so there needs to be a way of adjusting its format so that it is accessible to the Dashboards. Instead of manually parsing the data, there is an option to use the `DataModifiers`. Those modifiers allow the user to modify the data on the initial load, as well as the end user to modify visible data through the GUI. Each of the modifiers is accessible through the main Dashboards package, except for the `MathModifier` which is the most complex one, and requires loading a separate module from `https://code.highcharts.com/dashboards/modules/math-modifier.js`
```

docs/dashboards/data-modifiers.md
```diff
@@ -0,0 +1,80 @@
+Data Modifiers
+===
+
```
karolkolodziej, line 3: I would add why it's beneficial to use the modifiers instead of parsing everything on your own.

docs/dashboards/data-modifiers.md
```diff
@@ -0,0 +1,80 @@
...
+    options: {
+        //... other options
+        dataModifier: {
+            type: 'Math' // or 'Invert'|'Math' etc,
+            // additional modifier options
+        }
+    }
+}
+```
+The modifier is modifing the `table.modified` property of the connector, which leaves the original data unchanged, so that there is a way of restoring the original data if needed.
```
karolkolodziej, line 32:
```suggestion
The modifier modifies the `table.modified` property of the connector, which leaves the original data unchanged so that there is a way of restoring the original data if needed.
```

docs/dashboards/data-modifiers.md
```diff
@@ -0,0 +1,80 @@
+Data Modifiers
+===
+
+The data comes in different shapes and sizes from all over the internet, so there needs to be a way of adjusting its format, so that it is accessible to the Dashboards. Instead of manually parsing the data, there is an option to use the `DataModifiers`. Those modifiers allow the user to modify the data on the initial load, as well as the end user to modify visible data through the GUI. Each of the modifiers is accessible through the main Dashboards package, except of the `MathModifier` which is most complex one, and requires loading a seperate module from `https://code.highcharts.com/dashboards/modules/math-modifier.js`
+
+### Where the modifier is defined?
+The modifier is attached to the connector that is defined in the [dataPool](https://www.highcharts.com/docs/dashboards/data-handling), in its options.
+
+### Types of modifiers
+The types of modifiers that are available are:
```
karolkolodziej, line 10: You can add that each one is available for every connector

docs/dashboards/data-modifiers.md
```diff
@@ -0,0 +1,80 @@
...
+
+* Chain (Allows to combine multiple modifiers in the same connector)
```
karolkolodziej, line 12: You're describing everything bellow, I'm not sure if there is a need to duplicate it here.

docs/dashboards/data-modifiers.md
```diff
@@ -0,0 +1,80 @@
...
+dataModifier: {
+    type: 'Chain',
+    chain: [{
+     // ... Another modifier options
+    }, {
+     // ... Another modifier options
+    }]
+}
+```
+* Range - Allows to filter the data, which are exeeding the min and max properties. It works both for numbers as well as alphanumerical values. the `column` option specifies, from which column the values should be used to perform a filtering, and the `minValue` and `maxValue` specify the range limits. Here is the example:
```
karolkolodziej, line 48:
```suggestion
* Range - Allows filtering the data, which are exceeding the min and max properties. It works both for numbers as well as alphanumerical values. the `column` option specifies, from which column the values should be used to perform a filtering, and the `minValue` and `maxValue` specify the range limits. Here is an example:
```

docs/dashboards/data-modifiers.md
```diff
@@ -0,0 +1,80 @@
...
+    }]
+}
+```
+* Invert - Allows to flip the data, and replace the columns with rows. This comes in handy, when your data is structured in a wrong way, and you want to adjust the data to the format acceptable by Dashboards. No other options need to be specified here. Example:
+```javascript
+dataModifier: {
+    type: 'Invert'
+}
+```
+* Math - This is the only modifier, that comes in the seperate module. It allows to perform the complex math calculations, and add columns of values that are calculated based on the existing values. It uses the commands similar to the those, that you can find in spreadsheets like Excel. You can read more in the [MathModifier Article](<link to mathmodifier article>).
```
karolkolodziej, line 65:
```suggestion
* Math - This is the only modifier, that comes in the separate module. It allows performing complex math calculations and adding columns of values that are calculated based on the existing values. It uses commands similar to those, that you can find in spreadsheets like Excel. You can read more in the [MathModifier Article](<link to mathmodifier article>).
```

karolkolodziej (changes requested):

docs/dashboards/data-modifiers.md
```diff
@@ -0,0 +1,80 @@
...
+    type: 'Math',
+    columnFormulas: [{
+        column: 'USD',
+        formula: 'B1*C1' // Multiply EUR (B1) with the rate (C1)
+    }]
+}
+```
+
+### Example demos
+You can check the modifiers in work in given demos:
```
goransle, line 78: Suggestion: "You can check the modifiers at work in the following demos"

Vazonik (approved):

sebastianbochan (approved):

docs/dashboards/data-modifiers.md
```diff
@@ -0,0 +1,80 @@
+Data Modifiers
+===
+
+The data comes in different shapes and sizes from all over the internet, so there needs to be a way of adjusting its format so that it is accessible to the Dashboards. Instead of manually parsing the data, there is an option to use the `DataModifiers`. Those modifiers allow the user to modify the data on the initial load, as well as the end user to modify visible data through the GUI. Each of the modifiers is accessible through the main Dashboards package, except for the `MathModifier` which is the most complex one, and requires loading a separate module from `https://code.highcharts.com/dashboards/modules/math-modifier.js`. Each of the modifiers is available to connect to every type of connector.
```
bre1470, line 4: Data without "the" article gives a better reading flow.

bre1470 (approved):
Looks good. Some suggestions.

karolkolodziej (approved):

PR #19651, opened 2023-08-30
Dash/replace csv with jsonconnector

pawellysy:
Replaced `CSVconnector` with `JSONConnector` in the demos visible on main page.


karolkolodziej (approved):
Looks good!

goransle (approved):

sebastianbochan:
Can we change it in cypress test as well?

samples/dashboards/demo/datagrid-sync/demo.js
```diff
@@ -38,37 +45,47 @@ const chartOptions = {
...
+                            },
+                            {
+                                id: 'dashboard-col-2'
+                            }
+                        ]
+                    }
+                ]
+            }
+        ]
     },
```
Vazonik, line 89: This results in such long, in my opinion, unnecessary lines of code.

samples/dashboards/demo/datagrid-sync/demo.js
```diff
@@ -108,7 +126,8 @@ Dashboards.board('container', {
                     text: 'Dragging points <em>will not update</em> the grid'
                 }
             })
-        }, {
+        },
+        {
```
Vazonik, line 130: Similarly in these cases

Vazonik (approved):
Just a small note, but was breaking the blocks of brackets/braces on purpose? I think it would be better to restore it how it was before, i.e.:
```js
connectors: [{
    ...
}]
```
instead of 
```js
connectors: [
    {
        ...
    }
]
```

sebastianbochan (approved):

PR #19669, opened 2023-08-31
dash/editmode-event

sebastianbochan:
Added events to the EditMode state.


ts/Dashboards/EditMode/EditMode.ts
```diff
@@ -558,6 +560,8 @@ class EditMode {
 
         editMode.active = true;
         editMode.isContextDetectionActive = true;
+
+        eventsOptions?.activate.call(editMode);
```
Vazonik, line 564: When enabling edit mode, an error is thrown on this line and edit mode does not activate. `eventsOptions.activate` seems to be undefined; adding optional chaining to `activate` does not solve the problem.
image

Vazonik (changes requested):
For some reason, this feature doesn't work in my case, switching to edit mode throws an error.

ts/Dashboards/EditMode/EditMode.ts
sebastianbochan: By mistake, I set `activated` instead of `activate` in the demo.

Thanks!

Vazonik, line 564:
```suggestion
        eventsOptions?.activate?.call(editMode);
```

ts/Dashboards/EditMode/EditMode.ts
```diff
@@ -602,6 +607,8 @@ class EditMode {
 
         this.editCellContext = void 0;
         this.potentialCellContext = void 0;
+
+        eventsOptions?.deactivate.call(editMode);
```
Vazonik, line 611:
```suggestion
        eventsOptions?.deactivate?.call(editMode);
```

ts/Dashboards/EditMode/EditMode.ts
```diff
@@ -1061,6 +1068,14 @@ namespace EditMode {
...
          */
         tools?: Tools;
+         /**
+         * Events attached to the component : `activated`, `deactivated`.
+         *
+         * Try it:
+         *
+         * {@link https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/dashboards/edit-mode/events/ | Activated / Deactivated event }
+         */
+         events?: Record<string, Function>;
```
Vazonik, line 1078: + Too many spaces here. :)
```suggestion
        /**
        * Events attached to the component : `activated`, `deactivated`.
        *
        * Try it:
        *
        * {@link https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/dashboards/edit-mode/events/ | Activated / Deactivated event }
        */
        events?: Record<string, Function>;
```

Vazonik (changes requested):
Thanks, now it works! One more thing: Options can be unused, in which case they are undefined and an error is thrown while toggling the edit mode. We should check if they exist before calling callbacks. [demo](https://jsfiddle.net/BlackLabel/vknxm7cs/)

Vazonik (approved):

samples/dashboards/edit-mode/events/demo.html
```diff
@@ -0,0 +1,7 @@
+<script src="https://code.highcharts.com/dashboards/dashboards.js"></script>
+<script src="https://code.highcharts.com/highcharts.js"></script>
+<script src="https://code.highcharts.com/dashboards/datagrid.js"></script>
```
karolkolodziej, line 3: It isn't used.

samples/dashboards/edit-mode/events/demo.css
```diff
@@ -0,0 +1,11 @@
+@import url("https://code.highcharts.com/css/highcharts.css");
+@import url("https://code.highcharts.com/css/datagrid.css");
+@import url("https://code.highcharts.com/css/dashboards.css");
+
+#csv {
```
karolkolodziej, line 5: You don't have csv

samples/dashboards/edit-mode/events/demo.css
```diff
@@ -0,0 +1,11 @@
+@import url("https://code.highcharts.com/css/highcharts.css");
+@import url("https://code.highcharts.com/css/datagrid.css");
```
karolkolodziej, line 2: not nececary

karolkolodziej (changes requested):
Test is missing

karolkolodziej (approved):

ts/Dashboards/EditMode/EditMode.ts
```diff
@@ -558,6 +560,8 @@ class EditMode {
 
         editMode.active = true;
         editMode.isContextDetectionActive = true;
+
+        eventsOptions?.activate?.call(editMode);
```
pawellysy, line 564: I'd do that with `addEvent`  / `fireEvent`. If users will like to listen to that event from out of the API, there should be such way. So here, should just be `fireEvent` called, and in the constructor the event should be registered.

pawellysy (changes requested):

ts/Dashboards/EditMode/EditMode.ts
sebastianbochan: My first thought was the same, but I was wondering if adding the event to the editmode object makes sense. Im not convinced.

karolkolodziej: I would certainly unify them. As I look at the first one `mount` it's done by `fireEvent` so we should stick with that 😉

pawellysy: I'd add the event to the Board class.

pawellysy (approved):

PR #19674, opened 2023-09-01
tools/prods-change-detection

KacperMadej:
The node scripts could be tested using:
`node -e 'const { checkProduct } = require(\"./tools/gulptasks/lib/test.js\"); checkProduct(\"Core\");'`
(Escaping `"` might not be needed)

Expect an error when nothing is staged.

---
  - https://app.asana.com/0/0/1204891589365851


bre1470 (approved):
Nice! Looks good. 👍

TorsteinHonsi (approved):
Thanks!

Issue #14248, opened 2020-09-20
indicators Left and right will automatically fetch the day after or the day before if the data is empty


notbucai:
**Chinese translation, excuse me** [中文翻译，见谅]

indicators Left and right will automatically fetch the day after or the day before if the data is empty. [如果数据为空，左侧和右侧的指示器会在次日或前一天自动取数据。]

#### Expected behaviour
Expected data such as left and right null should be ignored [期望当前数据为空自动忽略而不是连接下一天数据] 

![image](https://user-images.githubusercontent.com/32786133/93713273-497b4f80-fb8d-11ea-808e-5e445a80dcba.png)

#### Actual behaviour
【实际会链接下一天】
![image](https://user-images.githubusercontent.com/32786133/93713269-3ff1e780-fb8d-11ea-93b6-56a571c0b080.png)

#### Live demo with steps to reproduce

template: https://jshare.com.cn/indicator/RSdgr8/15
#### Product version

Highstock JS v8.1.2 

#### Affected browser(s)
all

pawelfus:
Hi @notbucai 

Thank you for reporting the issue. Unfortunately jshare is not working right now (502 Bad gateway). I believe your demo is the same as this one: https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/stock/indicators/sma/ ?

This is expected behaviour, line series render one point before and after the visible range. For example simple line series: https://jsfiddle.net/BlackLabel/v14wt0qj/1/

If you don't want this, use this workaround:
```js
Highcharts.seriesTypes.sma.prototype.cropShoulder = 0;
```
Demo: https://jsfiddle.net/BlackLabel/v14wt0qj/2/

notbucai:
Thank you. This solves my problem

Issue #13985, opened 2020-08-02
Errors appear in console and bars are blinking with horizontal scroll


lifemaster-softserve:
#### Expected behaviour
Scroll shouldn't throw errors in console and bars shouldn't blinking while scrolling.

#### Actual behaviour
1. A lot of errors appear in console:
`Error: <rect> attribute y: Expected length, "NaN".`
`Error: <rect> attribute height: Expected length, "NaN".`

2. Bars are blinking while scrolling

#### Live demo with steps to reproduce
https://plnkr.co/edit/kZ40poSjW0Hm8UMD?open=lib%2Fapp.ts&deferRun=1&preview

1. Open console;
2. Just scroll horizontal scroll;

Bars are blinking and there are a lot of errors in console.

#### Product version
Errors are thrown since version 8.0.0
Bars are blinking since version 7.0.3

Till version 7.0.2 everything is fine.

#### Affected browser(s)
Google Chrome 84.0.4147.105 (x64)
Mozilla Firefox 79.0 (x64)

pawelfus:
Thank you for reporting the issue @lifemaster-softserve 

*Internal note:*
Bisected to commit fde683209c - caused by cropping points that are outside the visible xAxis extremes. In this particular case, that means we may have zoomed range that contains zero points (range = 0.5) which hides yAxis and causes these rendering errors and artifacts.

lifemaster-softserve:
Your workaround works fine but I have problem to do it in Angular with TypeScript:

`Property 'seriesTypes' does not exist on type 'typeof import("highcharts/highcharts")'. Did you mean 'seriesType'?`

So, `(Highcharts as any).seriesTypes.column.prototype.cropShoulder = 1;` works fine. Hope you will fix this soon.

lifemaster-softserve:
@pawelfus I have met one more issue with already applied workaround you suggested. Please, follow this link and try to use horizontal scroll. We can see the same issue again.

https://jsfiddle.net/oe49pa0n/

pawelfus:
Thank you for posting this issue :) 

It's not a perfect solution, but for now, the best would be to use [`series.getExtremesFromAll`](https://api.highcharts.com/highcharts/series.column.getExtremesFromAll) option: https://jsfiddle.net/BlackLabel/qL4gzp2d/ - downside is that yAxis extremes are always calculated according to the entire dataset, not using the visible points only.

lifemaster-softserve:
Thank you, @pawelfus It works for me.

[...]

PR #19678, opened 2023-09-04
Fixed #14182, points disappeared after zooming on variwide series

### LINKED ISSUE #3136: cropThreshold doesn't work

### LINKED ISSUE #14182: Variwide within range - zoom not working properly
#### Expected behaviour
After zooming between 21.25 and 22.00  should render properly.
https://jsfiddle.net/raghava_2121/rq3mfc4L/9/

7138A6F8-569E-4B19-8AC1-DEF6C8903E6F_1_105_c

#### Actual behaviour
After zoomin graph doesn't display the selected range with colour.
 
https://jsfiddle.net/raghava_2121/rq3mfc4L/13/

#### Product version
Highstock 

#### Affected browser(s)
google chrome

TorsteinHonsi:
Fixed #14182, points disappeared after zooming on variwide series

___
Removed the existing `cropThreshold` series property, since it seems that there is no case when we want it to be 0 rather than 1. Issues #14182 and #3136 shows that column-like series also should have one graphic generated outside the zoomed area.

Closes #3136.

---
  - https://app.asana.com/0/0/1205293302215875


TorsteinHonsi:
Most of the visual diff is caused by karma now picking the shoulder point for tooltip comparison. Will probably be fixed by #19600.

TorsteinHonsi:
@pawelfus As of now there are 15 diffing visual tests. As far as I can see, all of those come from the fact that the data grouping algorithm uses only data that is within the visible range. Except now it also includes one point outside. If I apply `groupAll` to those samples, it is not different from before. So it is no more incorrect now than it used to be, so to speak. 

Any thoughts?

pawelfus:
There is not much to add, I don't know any specific case where shoulder=1 would cause more trouble than shoulder=0 for column series. If we get such a bug report, we can revert it and set it only for variwide series.

pawelfus (approved):

pawelfus:
Worth checking if those are also fixed with the commit above: #13985

Note: Just realized there was a case when someone didn't want extra points outside the visible chart: #14248

PR #19687, opened 2023-09-05
dash/fix-component-loading-state

bre1470:
Limited loading state to components with (new) connector.


karolkolodziej (changes requested):
Looks good but can we also get rid of the error `Uncaught (in promise) TypeError: component.load(...).catch is not a function` by probably making the `load` as async?

bre1470:
Sorry, I do not understand. For me both function are async. `Component.load` and `Component.initConnector`.

bre1470:
Oh, I understand now. The custom YouTube component lacks async. I will fix. 👍

karolkolodziej (approved):

goransle (approved):

PR #19690, opened 2023-09-05
dash/datagrid-mobile-responsiveness

Vazonik:
Added updating the data grid while container resizing, fixed overlapping headers and some minor bugs.


ts/DataGrid/DataGrid.ts
```diff
@@ -514,6 +520,10 @@ class DataGrid {
         this.container.addEventListener('mouseover', (e): void => {
             this.handleMouseOver(e);
         });
+
+        new ResizeObserver((): void => {
```
goransle, line 524: Should we add this as a class prop, so DataGridComponent can call `observer.disconnect` on `destroy()`? (Possibly we should also have destroy on the DataGrid, but that's a larger matter)

ts/DataGrid/DataGrid.ts
```diff
@@ -923,7 +946,8 @@ class DataGrid {
         }
 
         const headerEl = makeDiv(className);
-        headerEl.style.height = this.options.cellHeight + 'px';
+        headerEl.style.minHeight = this.options.cellHeight + 'px';
+        headerEl.style.maxHeight = this.options.cellHeight * 2 + 'px';
```
karolkolodziej, line 950: Please add the description in the `cellHeight` API that the header height might increase its size in certain cases.

ts/DataGrid/DataGrid.ts
```diff
@@ -956,6 +980,89 @@ class DataGrid {
...
+
+        this.updateColumnHeaders();
+    }
+
+
+    /**
+     * Updates the column headers of the table.
+     * @internal
+     */
+    private updateColumnHeaders(): void {
```
karolkolodziej, line 992: I'm not sure if this only updated the header. I would change the name and would try to split it into smaller methods responsible for a given element.

karolkolodziej (changes requested):

ts/DataGrid/DataGrid.ts
bre1470: BTW Do we have any plans for row headers?

karolkolodziej (approved):
Looks good I'm only wondering why we don't get any visual changes. 🤔

test/cypress/integration/DataGrid/cell-header-format.cy.js
```diff
@@ -53,11 +53,21 @@ describe('Remove the dashboard.', () => {
...
                 assert.closeTo(
                     finalWidth,
-                    500,
+                    200,
                     10,
                     'The width should be close to 500px.'
                 )
             });
+
+            cy.wait(50);
```
sebastianbochan, line 62: What is the reason of `cy.wait` here?

ts/DataGrid/DataGrid.ts
```diff
@@ -956,6 +979,115 @@ class DataGrid {
...
+        const headersContainer = this.columnHeadersContainer;
+        if (!headersContainer) {
+            return;
+        }
+
+        // Handle overflowing text in headers.
+        this.columnNames.forEach((columnName, i): void => {
```
pawellysy, line 1008: We avoid using `forEach` in the dashboards for performance reasons. Could you use the basic `for` loop?

pawellysy (changes requested):

sebastianbochan (approved):
Please fix Pueblo suggestions.

bre1470:
@goransle Is it possible to round the benchmark results to two fraction digits?

bre1470 (approved):
Nice!

pawellysy (approved):

PR #19694, opened 2023-09-05
Tools: Fixed Chrome GH workflow, Added BrowserStack, Safari

bre1470:
I structured the testing, so that it spends as little time as possible. BrowserStack is a separate job though, because of the potential timeout. Safari headless is not starting fast enough, because of limited resources; XL resource is not available without billing setup.

References:
- [GHA runner resources doc](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources)
- [GHA XL runners doc](https://docs.github.com/en/actions/using-github-hosted-runners/about-larger-runners/about-larger-runners)
- [GHA billing doc](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#per-minute-rates)


TorsteinHonsi (approved):
Thanks!

PR #19695, opened 2023-09-05
feature/split-utils

pawellysy:
Splitted `ts/Core/Utilities.ts` into several files, located in the `ts/Shared` directory.


TorsteinHonsi:
A little status update here. This PR was put on hold because with the current bundler, splitting files increases the produced bundle size significantly. We are now in the process of implementing Webpack as the bundler for Highcharts and Dashboards. Eventually it will change this PR or make it redundant. Here's the way forward:

1. Implement Webpack bundling into the master.

2. After Webpack has been implemented, we can rewrite our modules to use direct import instead of importing the whole Utilities.

This pattern
```ts
import U from './Utilities.js';
const {
    defined,
    extend
} = U;
```
... then becomes
```ts
import { defined, extend } from './Utilities';
```
At this point, the file size should be reduced because
a) the utility functions are lifted to the top scope and referenced directly instead of every module having to load and unpack `U`, and
b) utility functions that are not used in a specific primary bundle (read `dashboards.js`), are shaken off.

3. After this, we can reconsider splitting the Utilities.ts like we did in this PR, but now only for our convenience. At this point, it shouldn't affect the file size of the bundle.

PR #19697, opened 2023-09-06
Fixed #19692, mousewheel zoom was enabled even when general zoom was not

### LINKED ISSUE #19692: Mouse wheel zoom is enabled even when general zoom is not
#### Expected behaviour
When `zooming.type` is not defined, mouse wheel zoom should not be enabled. When `zooming.type` is enabled, mouse wheel zoom should be enabled (unless explicitly disabled).

#### Actual behaviour
It is enabled by default whenever the mouse-wheel-zoom module is loaded. This blocks page scrolling.

[](https://jsfiddle.net/highcharts/01tman8z/)

#### Product version
v11.x

#### Affected browser(s)
All

TorsteinHonsi:
Fixed #19692, mouse wheel zoom was enabled even when general zoom was not

We should have a test, but I didn't manage to get cypress right. For now I'm using this: https://jsfiddle.net/highcharts/01tman8z/ . Note that when zooming is enabled, and you try to zoom out past the current extremes, browser scroll kicks in because there is nothing left to zoom in the chart.

### DIFF
--- a/test/cypress/readme.md
+++ b/test/cypress/readme.md
@@ -27,6 +27,6 @@ The Cypress docs has a good [article](https://docs.cypress.io/guides/references/
 
 ## Running tests locally
 
-Requires `highcharts-utils` to be running (`npx highcharts-utils` should do the trick). Cypress can then be run in the CLI using `npx cypress run`. Run `npx cypress info` to get a list of available browsers.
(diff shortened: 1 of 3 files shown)


DJTechnoo (approved):
Looking good!

bre1470 (approved):
Tested with reduced defaults in Dashboards and works nicely. 👍

PR #19698, opened 2023-09-06
docs/19623-esm-docs-improvements

### LINKED ISSUE #19623: Many console errors for basic chart following the installation instructions
#### Details
I have followed the most basic of tutorials in the [docs](https://www.highcharts.com/docs/getting-started/installation-with-esm#creating-a-custom-bundle-es6-module) and the console is full of errors already. On page load there is an error and when I mouse over anything I get a stream of other errors (see screenshot). Having ran a debugger I can see that [here](https://github.com/highcharts/highcharts/blob/0043bc371ed03f081616a961453d7d765ec813a4/ts/Core/Series/Series.ts#L4381) the `pointer` variable is being set to `undefined`.

This also happens in Column charts.

#### My code

```
import Chart from 'highcharts/es-modules/Core/Chart/Chart.js';
import LineSeries from 'highcharts/es-modules/Series/Line/LineSeries.js';
// Example to create a simple line chart in a div#container:
const myChart = new Chart('container', {
    series: [
        {
            type: 'line',
            data: [1, 2, 3]
        }
    ]
});
```

I feel like maybe I'm supposed to get a tooltip on hover but as this is my first encounter with the product I'm not sure. But I don't get a tooltip, I just get errors.

#### Expected behaviour
No console errors

#### Actual behaviour
Many console errors
Screenshot 2023-08-25 at 10 41 24

#### Product version
highcharts: 11.1.0

#### Affected browser(s)
Chrome: 115.0.5790.170 MacOs

bm64:
Fixed #19623, added missing pointer imports and minor improvements.


karolkolodziej (approved):

TorsteinHonsi:
Did it fail when the Pointer class was not imported? Or is it just that we want the tooltip in the resulting bundle? I mean, if it failed without the Pointer then that would be a bug in itself.

karolkolodziej:
There is an error in the console without the `Pointer`,[ see it in the demo.](https://jsfiddle.net/BlackLabel/tLf82wca/)

TorsteinHonsi:
Then the bug is that Chart uses `pointer` without knowing if it exists. @bre1470 how do you think we should resolve that?

There are 86 occurrences throughout the product where we read `chart.pointer.someProperty`. We could make `Chart.pointer` optional and change the calls to `chart.pointer?.someProperty`. 

Alternatively we could instanciate the Pointer inside Chart.ts, but then it will not be possible for custom builds to build without the Pointer, and the smallest possible bundle size will increase.

TorsteinHonsi:
> Otherwise optional-chains sounds fine as well, but will increase the size because of transpiling, as added [browser support](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) is not that long ago.

@bre1470 It was added to the evergreen browsers more than three years ago, so there is sufficient support as per our three-year rule. We tried to fix this earlier this summer, but it didn't work out fully - do you remember what failed? Currently optional chaining is transpiled to a long and uneffective chain of logic.

TorsteinHonsi:
We now have support for optional chaining, so I think we should make the change to make `Chart.pointer` optional. Then the docs changes will no longer be relevant. @bm64

PR #15428, opened 2021-03-31
enhancement/12455-dataGrouping-x-coordinates

karolkolodziej:
Feature #12455, added an option to control the x-coordinates of the grouped point.

#12455 Upgrade note
The data grouping `smoothed` property has been deprecated and has been changed to `firstAnchor`, `anchor` and `lastAnchor`.

#### To do list:
- [x] - implementation
- [x] - deprecate `smoothed` option + implemented fallback to the new API (e.g. Navigator series use `smoothed` option by default)
- [x] - tests
- [x]  - demos:
  - [x] - candlestick with static data to show `anchor` in action
  - [x] - line series with "live data" to show `lastAnchor` using `lastPoint` and `firstAnchor` using `firstPoint`
- [x] - API (https://api.highcharts.com/highstock/plotOptions.series.dataGrouping)
- [x] - update docs (https://www.highcharts.com/docs/stock/data-grouping)

Bugs:
- [x] `anchor: 'end'` last point missing- extremes
- [x] anchors in navigator
- [x] accept visual differences

PR #19699, opened 2023-09-06
bugfix/17212-navigator-gap

### LINKED ISSUE #17212: StockChart set rangeSelector: enabled: false, navigator Left a small gap
$.getJSON('https://data.jianshukeji.com/stock/history/000001', function (data) {
	if(data.code !== 1) {
		alert('读取股票数据失败！');
		return false;
	}
	data = data.data;
	// 去掉多余的数据
	Highcharts.each(data, function(d) {
		d.length = 2;
	});
	Highcharts.stockChart('container', {
		rangeSelector: {
			enabled: false
		},
		title: {
			text: '平安银行历史股价'
		},
		plotOptions: {
			series: {
				showInLegend: true
			}
		},
		xAxis: {
			min:data[0][0],
			max:data[0][0]+10000000000,
		},
		tooltip: {
			split: false,
			shared: true
		},
		series: [{
			// type: 'line',
			id: '000001',
			name: '平安银行',
			data: data
		}]
	});
});
image

kamil-musialowski:
Fixed #17212, first grouped point in `navigator` was incorrect.

_Internal:_
Original issue is caused by #15428. When `dataGrouping.smoothed` (true by default for navigator) was not deprecated, first main series point was assigned to be a first grouped point in navigator data.
https://github.com/highcharts/highcharts/pull/15428/files#diff-38953316589f5604be28dd779a25430f3178ae7525fc6fbc1864ff1fb3f7aad6L777

After the aforementioned commit, this logic got replaced by `anchorPoints()`, but it's internal if check was not passed by navigator.
https://github.com/highcharts/highcharts/blob/0a045faee69f61563ddbbefb86d435ec1975c667/ts/Extensions/DataGrouping/DataGroupingSeriesComposition.ts#L208

### DIFF
--- a/ts/Stock/Navigator/Navigator.ts
+++ b/ts/Stock/Navigator/Navigator.ts
@@ -1558,6 +1558,7 @@ class Navigator {
                 showInLegend: false,
                 stacking: void 0, // #4823
                 isInternal: true,
+                isNavigator: true, // #17212
                 states: {
                     inactive: {
                         opacity: 1
--- a/ts/Stock/Navigator/Navigator.ts
+++ b/ts/Stock/Navigator/Navigator.ts
@@ -1558,7 +1558,6 @@ class Navigator {
                 showInLegend: false,
                 stacking: void 0, // #4823
                 isInternal: true,
-                isNavigator: true, // #17212
                 states: {
                     inactive: {
                         opacity: 1
--- a/samples/unit-tests/navigator/navigator/demo.js
+++ b/samples/unit-tests/navigator/navigator/demo.js
@@ -425,7 +425,7 @@ QUnit.test('Scrollbar without navigator (#5709).', function (assert) {
     );
 });
 
-QUnit.test('Missing points using navigator (#5699)', function (assert) {
+QUnit.test('Missing points using navigator (#5699, #17212)', function (assert) {
     const data = new Array(3000).fill(1).map((item, i) => [
         Date.UTC(2010, 0, 1) + i * 24 * 36e5,
         Math.random()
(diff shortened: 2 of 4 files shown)


ts/Stock/Navigator/Navigator.ts
```diff
@@ -1558,6 +1558,7 @@ class Navigator {
                 showInLegend: false,
                 stacking: void 0, // #4823
                 isInternal: true,
+                isNavigator: true, // #17212
```
TorsteinHonsi, line 1561: I think you can just use the `isInternal` property, instead of adding a new property. It is mostly used by the navigator.

TorsteinHonsi (changes requested):

ts/Stock/Navigator/Navigator.ts
kamil-musialowski: Thank you, I had no idea about it :)

TorsteinHonsi:
Pending review from @pawelfus

pawelfus (changes requested):
I'm afraid this is just the tip of the iceberg.. 

The condition `if (dataGroupingOptions && series.xData && totalRange && series.groupMap) {` is always falsy on the initial rendering (for the first series). `series.totalRange` and `series.groupMap` properties are defined after `anchorPoints()`. Perhaps we should just move `anchorPojnts()` call just a few lines later, after properties are set? Could you try it, please?

kamil-musialowski:
Thank you @pawelfus , I have implemented it - and now it looks like the issue is fixed, without any additional code :)

TorsteinHonsi (approved):

PR #19710, opened 2023-09-07
dash/cellFormat-datagrid

sebastianbochan:
Added cellFormatter to the dataGrid module.

### DIFF
--- a/test/cypress/integration/DataGrid/cell-format.cy.js
+++ b/test/cypress/integration/DataGrid/cell-format.cy.js
@@ -3,7 +3,7 @@ describe('Remove the dashboard.', () => {
         cy.visit('/dashboards/cypress/datagrid-cell-formatter');
     });
 
-    it('Dates un cells should be formatted.', () => {
+    it('Dates in cells should be formatted.', () => {
         cy.get('.highcharts-datagrid-cell').eq(0).should('have.text', '2022-01-01');
     });
 });
--- a/test/cypress/integration/DataGrid/cell-format.cy.js
+++ b/test/cypress/integration/DataGrid/cell-format.cy.js
@@ -1,0 +1,9 @@
+describe('Remove the dashboard.', () => {
+    before(() => {
+        cy.visit('/dashboards/cypress/datagrid-cell-formatter');
+    });
+
+    it('Dates un cells should be formatted.', () => {
+        cy.get('.highcharts-datagrid-cell').eq(0).should('have.text', '2022-01-01');
+    });
+});
--- a/ts/DataGrid/DataGridOptions.d.ts
+++ b/ts/DataGrid/DataGridOptions.d.ts
@@ -118,9 +118,9 @@ export interface ColumnOptions {
      * Extendable method for formatting each cell's in DataGrid.
      *
      * @return {string}
-     *         A string to be concatenated in to the common cell's text.
+     * A string to be concatenated in to the common cell's text.
      */
-    cellFormatter?: Function;
+    cellFormatter?: cellFormatterCallback;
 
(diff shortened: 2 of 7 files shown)


ts/DataGrid/DataGridOptions.d.ts
```diff
@@ -114,6 +114,14 @@ export interface ColumnOptions {
      */
     cellFormat?: string;
 
+    /**
+     * Extendable method for formatting each cell's in DataGrid.
+     *
+     * @return {string}
+     *         A string to be concatenated in to the common cell's text.
+     */
+    cellFormatter?: Function;
```
karolkolodziej, line 123: Please add a dedicated interface for this callback.

karolkolodziej (changes requested):
Test is missing that's for sure. 
Also, one thing to consider if we should add the `headerFormatter`

ts/DataGrid/DataGridOptions.d.ts
```diff
@@ -153,6 +153,14 @@ export interface ColumnOptions {
     show?: boolean;
 }
 
+export interface cellFormatterCallback {
```
karolkolodziej, line 159: Conventionally interfaces should start from the capital letter.

ts/DataGrid/DataGridOptions.d.ts
```diff
@@ -153,6 +153,14 @@ export interface ColumnOptions {
...
+    (this: cellValue): string;
+}
+
+export interface cellValue {
```
karolkolodziej, line 166: same

karolkolodziej (changes requested):

ts/DataGrid/DataGridOptions.d.ts
```diff
@@ -145,6 +153,20 @@ export interface ColumnOptions {
...
+    (this: CellValue): string;
+}
+
+/**
+ * Value to convert
+ */
+export interface CellValue {
+    value: number|string
+}
+
```
bre1470, line 169: Shouldn't this interface be the same as DataTable's `CellType` types?

ts/DataGrid/DataGridOptions.d.ts
```diff
@@ -145,6 +153,20 @@ export interface ColumnOptions {
     show?: boolean;
 }
 
+/**
+ * Returns a formatted call's string.
+ */
+export interface CellFormatterCallback {
+    (this: CellValue): string;
```
bre1470, line 160: The function scope makes it impossible to use arrow functions as callbacks. I would expect that the table value comes as the first or second argument.

sebastianbochan: We do not support arrow functions in Highcharts or Dashboards events and formatted.

Eg. https://jsfiddle.net/2g7edxvf/1/

Knowing that the formatter is a temporary solution (like we discussed during the last meeting), I would skip arrow function support.

test/cypress/integration/DataGrid/cell-format.cy.js
```diff
@@ -0,0 +1,9 @@
+describe('Remove the dashboard.', () => {
+    before(() => {
+        cy.visit('/dashboards/cypress/datagrid-cell-formatter');
+    });
+
+    it('Dates un cells should be formatted.', () => {
```
goransle, line 6: Minor typo here

goransle (approved):
LGTM (small typo in test description aside)

bre1470 (approved):

karolkolodziej (approved):

Vazonik (approved):
LGTM

PR #19712, opened 2023-09-07
bugfix/19671-treegraph-legend-symbol-line-marker

### LINKED ISSUE #19671: Treegraph chart throws error when setting legendSymbol to 'lineMarker'.
#### Expected behaviour
When `treegraph.legendSymbol` is set to `lineMarker`, the chart doesn't break and there is no error in the console.

#### Actual behaviour
When `treegraph.legendSymbol` is set to `lineMarker`, the chart breaks and `TypeError: Cannot read properties of undefined (reading 'indexOf')` error is thrown.

https://jsfiddle.net/BlackLabel/5avsd4hu/

#### Product version
Highcharts v11.1.0

#### Affected browser(s)
All

jakubjanuchta:
Fixed #19671, treegraph was throwing an error when setting `legendSymbol` to `lineMarker`.

### DIFF
--- a/ts/Series/Treegraph/TreegraphSeries.ts
+++ b/ts/Series/Treegraph/TreegraphSeries.ts
@@ -532,40 +532,47 @@ class TreegraphSeries extends TreemapSeries {
         state?: StatesOptionsKey
     ): SVGAttributes {
         const series = this,
-            levelOptions =
+            levelOptions = point &&
                 (series.mapOptionsToLevel as any)[point.node.level || 0] || {},
-            options = point.options,
+            options = point && point.options,
             stateOptions =
                 (levelOptions.states &&
                     (levelOptions.states as any)[state as any]) ||
                 {};
-        point.options.marker = merge(
-            series.options.marker,
-            levelOptions.marker,
-            point.options.marker
-        );
+
+        if (point) {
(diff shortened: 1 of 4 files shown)


ts/Series/Treegraph/TreegraphSeries.ts
```diff
@@ -532,40 +532,49 @@ class TreegraphSeries extends TreemapSeries {
         state?: StatesOptionsKey
     ): SVGAttributes {
         const series = this,
-            levelOptions =
+            levelOptions = point &&
```
TorsteinHonsi, line 535: Now that we have a situation where `point` may be undefined, the typing must reflect it:

https://github.com/highcharts/highcharts/blob/a05042ac5298b093986527f166a9b92e548e472d/ts/Series/Treegraph/TreegraphSeries.ts#L530-L533

Like in the base series, where the `point` argument is optional:

https://github.com/highcharts/highcharts/blob/a05042ac5298b093986527f166a9b92e548e472d/ts/Core/Series/Series.ts#L2811-L2814

TorsteinHonsi (changes requested):
Thanks! Just a tiny little question mark missing, otherwise perfect.

TorsteinHonsi (approved):

PR #19717, opened 2023-09-08
samples/15643-scrollable-legend-demos

### LINKED ISSUE #15643: support scroll-bar in legend
Highcharts in the past has provided the legend.style API, which provided an option to use scrollbar instead of the arrows navigation thats supported today

In our product, UX team requires to have a **scrollbar if there is not enough space for all the legends.**
So I would like to have an option to add the scroller manually or (better option) to have a proper configuration on the legend
options

**Example:** legend.scroller API/ legend.navigator.scrollerEnable

**Mock:**
image

hubertkozik:
Added two demos with a custom plugin for the scrollable legend, closes #15643.

---
  - https://app.asana.com/0/0/1204962059999285


MarkusBarstad (approved):
Works well and code is concise!  ⭐ 

Not sure if it matters but you could shorten property lookups without hindering readability for `scrollableLegendArea.minHeight` and `scrollableLegendArea.minWidth` by adding a simple destructuring statement above the two ifs like `{ minWidth, minHeight } = scrollableLegendArea;` 

I must admit I am not totally sure on what the policy around property-lookups are for demos, but I assume readability is the most important thing so maybe it could also be good to have some property lookups 🤔 .

KacperMadej (changes requested):
- [ ] The default design clips the legend's scrollbar on Win
image

- [ ] An export attempt causes error.
- [ ] Horizontal: start with a big window and resize to a small one - doesn't look correct (vertical scollbar, 3 lines of legend items limited in layout by the chart's width).
image

- [ ] Horizontal: start small and go big - items clipped, no visible scrollbar (it probably lands under the view box).
- [ ] It would be nice to have a reference to the [legend.navigation](https://api.highcharts.com/highcharts/legend.navigation) in the edited API reference in case someone would be interested in the build-in alternative.
- Could one of the demos be more styled to demonstrate more advanced styling capabilities?

hubertkozik:
Can you check the default design on Windows with the current code?

Regarding one more styled demo, we had a short conversation: https://highsoft.slack.com/archives/C05D7UNHBN0/p1694160518700729?thread_ts=1694006524.794019&cid=C05D7UNHBN0

TorsteinHonsi (approved):
Thanks, looks good!

PR #19720, opened 2023-09-08
dash/fix-climate-demo

bre1470:
- Fixed KPI switch in climate demo
- Fixed initial range (= real 365 days)
- Reduced data pool requests
- Reduced chart redraws
- Removed selection timeout


bre1470:
On the old PC, the navigator selection visually jumps, but the mouse cursor moves smooth, so imho this is tolerable.

samples/dashboards/demo/climate/demo.js
```diff
@@ -284,26 +286,23 @@ async function setupBoard() {
...
-                                    );
-                                }
-                            }, 50);
+                        afterSetExtremes: async function (e) {
+                            if (
+                                activeTimeRange[0] !== Math.round(e.min) ||
+                                activeTimeRange[1] !== Math.round(e.max)
+                            ) {
+                                activeTimeRange = [
+                                    Math.round(e.min),
```
sebastianbochan, line 295: Would not be faster if the `Math.round(e.min)` were assigned to a variable? We avoid double calculations.

samples/dashboards/demo/climate/demo.js
```diff
@@ -1038,69 +1035,56 @@ async function updateBoard(board, city, column, scale, newData) {
     }
 
     const rangeTable = selectionTable.modified;
+    const rangeEnd = rangeTable.getRowCount() - 1;
 
     // Update world map
+    const mapPoints = worldMap.chart.series[1].data;
+    const lastTime = rangeTable.getCellAsNumber('time', rangeEnd);
+
+    for (const mapPoint of mapPoints) {
```
sebastianbochan, line 1044: I checked it and it turned out that the regular loop is 2x faster then for/of.
- https://jsfiddle.net/mxcfyvkw/2/
(The newest chrome).

I think that in the old browsers or CPU, the difference will be more significant.

sebastianbochan (changes requested):

samples/dashboards/demo/climate/demo.js
bre1470: Good (map)point. 👍

sebastianbochan (approved):

karolkolodziej (approved):
That is how it is supposed to work! Great work! 
Can you please check why the visual changes are failing? Especially in the `minimal` demo?

bre1470:
It seems to be a configuration issue in a GitHub job or action.

bre1470:
Additional tests show decent performance for a PC from 2020, and okaisch performance for a PC from 2013. We might consider a short timeout in further valuation.

PR #19723, opened 2023-09-08
tools/cypress-multi-config

goransle:
Separated dashboards lighthouse and visual tests into separate configs, to avoid setting up plugins for other tests. 

Note that comparison tests using Cypress will fail since the config files are not on master.


TorsteinHonsi (approved):
Thanks!

oysteinmoseng (approved):
Thanks! No comments from me 👍

PR #19732, opened 2023-09-11
Optimalized imports in Dashboards.

pawellysy:


ts/Shared/Utilities.ts
```diff
@@ -16,22 +16,21 @@
...
-
-import H from './Globals.js';
+} from '../Core/Renderer/DOMElementType';
+import type EventCallback from '../Core/EventCallback';
+import type HTMLAttributes from '../Core/Renderer/HTML/HTMLAttributes';
+import type Series from '../Core/Series/Series';
+import type SVGAttributes from '../Core/Renderer/SVG/SVGAttributes';
+import type Time from '../Core/Time';
+
+import H from '../Core/Globals.js';
```
bre1470, line 32: It does not help to just move the file, if the dependencies are kept. `Shared`-files should not have imports at all, and if really necessary these should be in `Shared` as well. The best `Shared`-files are atomic (self-containing) ones.

bre1470 (changes requested):
IMHO Utilities needs to be splitted in two files; one in Core and the other in Shared. Otherwise we continue to have polluted declarations in Dashboards.

bre1470:
Maybe it is better to move only the functions into `Shared/Utilities` that are used by Dashboards - and the rest is left in `Core/Utilities`. That reduces the amount of changes all over the places.

PR #19737, opened 2023-09-12
samples/geoheatmap-demos

hubertkozik:
Changed demos for the GeoHeatmap series.

---
  - https://app.asana.com/0/0/1205395062434666
  - https://app.asana.com/0/0/1205383740754138


KacperMadej (approved):

TorsteinHonsi (changes requested):
Thanks!

I have a change request - I'm not sure if this belongs to the demo or if it is an issue with the interpolation itself. 

If you open the utils for the Europe demo and change the theme to BrandDark, you'll notice that the cells fade to transparent via a light color rather than a dark one. I am guessing that they fade to `rgba(256, 256, 256, 0)`. I'm not sure how to fix this, but they should probably fade to the zero-opacity version of the main color instead.

Feel free to take that problem out to a separate issue if it doesn't belong in this PR.

Skjermbilde 2023-10-03 kl  06 35 24

hubertkozik:
@TorsteinHonsi I believe it is correct behaviour. I did some research in our code and I came up with some conclusions. BrandDark theme and other "dark" themes are changing the background color of the chart to some dark color. That's why we see that color axis bar has a white color from the left side. 

If we add dark background color to a normal chart, for example, in this demo: https://jsfiddle.net/BlackLabel/cf2wz1t8/ we will also see "white-y" color on columns - same as for interpolated image in Geoheatmap. If we would like to change it to take into account opacity instead of white color: https://jsfiddle.net/BlackLabel/kcza0m34/ we get the desired behaviour for Geoheatmap series - fade to transparent via a light color.

I think there is nothing to change in the code, it can be fully controlled by `colorAxis.minColor` and `colorAxis.maxColor`. What do you think about my understanding of this problem?

TorsteinHonsi:
@hubertkozik That sounds good!

I guess that means we should change the `minColor` to a transparent one in the demo then? And not change anything in the product.

TorsteinHonsi:
> but we will lose the possibility of changing those colors by theme.

@hubertkozik Okay. Probably not what we want. Can you check out setting a default `minColor` and `maxColor` in the BrandDark theme then?

TorsteinHonsi (changes requested):
Thanks! 

I'm afraid we need to revert the last change, because simply by adding a `colorAxis` object to the config, it will enable a color axis in all charts. You'll see it if you load a vanilla chart, a theme and include the map module. This is a design flaw that we need to overcome in order to use `colorAxis` in the themes, but that's another issue. Let's just skip the gradient issue for now and get this merged.

TorsteinHonsi (approved):

Issue #16277, opened 2021-08-31
Indicator Volume by Price not working correctly in relative ordinate


robocop98:
#### Expected behaviour
Points of volume-by-price indicator should be drawn with relative values when using compare: 'percent'.

When switching between absolute and relative ordinate the volume-by-price bars should remain on the same relative position to its underlying series line, like simple indicators (SMA, ...) and multi-line indicator (Bollinger Bands) do.

#### Actual behaviour
volume-by-price is always drawn with absolute values, so not shown if switching to relative ordinates (compare: 'percent'),

Exception: indicator can be seen randomly, if its underlying series has values near the percentual ones.

EDIT: vbp Indicator is also not working on logarithmic ordinates, but not sure if it makes sense to use it there.

#### Live demo with steps to reproduce
Absolute ordinate works: https://jsfiddle.net/46d28qxb/1/
Relative ordinate does not work, volume-by-price indicator is outside of the view: https://jsfiddle.net/46d28qxb/

Difference between both fiddles:
- at main series and vbp series: compare: 'percent'
- at vbp series: compareToMain: true

#### Product version
Highstocks 9.2.0 - 9.2.2

#### Affected browser(s)
Tested in Chrome and Firefox, but issue should not depend on browser.

Thank you for your constant support @ Highcharts team

raf18seb:
Hi @robocop98, thanks for creating the ticket!

The VBP indicator is slightly different from the others - it's not a simple line, the values are not distributed along the x-axis and the standard `compare` feature wasn't developed to cover this case.
How would you like to compare the VBP columns values (green and red)? Could you describe a simple case?

robocop98:
Hi @raf18seb , thanks for picking up this topic.

The use cases we currently have in our setup:
- in a stock chart with one instrument series we can manually switch the y-axis between absolute and relative coordinates
-  either see the

raf18seb:
@robocop98, thanks for clarifying!

> The indicators y-coordinates are correct in absolute view, but would need re-calculation for the relative view (compare='percent'), so that it is shown at "same location" like in absolute view.

Yes, this makes sense. In other words, the candlestick series would be calculated in the `compare` mode and the VbP indicator would be translated to the correct place, so it fits new y-axis extremes.

However, could you clarify the options you are describing? Especially `compareToMain`.
Did you mean that:
- if we set `compareToMain: true` on the VbP series, then the VbP would be calculated on the main candlestick values, so the VbP would stay the same (as it works now), and
[...]

PR #19738, opened 2023-09-12
bugfix/16686-vbp-yaxis-extremes 

### LINKED ISSUE #16686: VBP with compare - strange yAxis extremes when moving the navigator
Like in the title.

vbp-bug

Take an official VBP indicator demo, set `compare: 'percent'` on the VBP series and then move the navigator
https://jsfiddle.net/zcy6vp8e/

#### Product version
Highcharts Stock v9.3.1

kamil-musialowski:
Fixed #16686, yAxis extremes were incorrect with VBP indicator compare.

Although it looks like a hacky solution, VBP (with compare) recalculations should not affect the yAxis extremes calculations, since due to the VBP position and rotation, `yData` does not affect the `yAxis` ticks.

### DIFF
--- a/ts/Stock/Indicators/VBP/VBPIndicator.ts
+++ b/ts/Stock/Indicators/VBP/VBPIndicator.ts
@@ -22,6 +22,7 @@ import type AxisType from '../../../Core/Axis/AxisType';
 import type Chart from '../../../Core/Chart/Chart';
 import type ColumnSeries from '../../../Series/Column/ColumnSeries';
 import type CSSObject from '../../../Core/Renderer/CSSObject';
+import type DataExtremesObject from '../../../Core/Series/DataExtremesObject';
 import type IndicatorValuesObject from '../IndicatorValuesObject';
 import type LineSeries from '../../../Series/Line/LineSeries';
 import type SVGAttributes from '../../../Core/Renderer/SVG/SVGAttributes';
@@ -46,7 +47,6 @@ const {
 } = SeriesRegistry.seriesTypes;
 import U from '../../../Core/Utilities.js';
 import StockChart from '../../../Core/Chart/StockChart.js';
-import DataExtremesObject from '../../../Core/Series/DataExtremesObject';
 const {
     addEvent,
     arrayMax,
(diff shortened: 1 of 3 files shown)


pawelfus (changes requested):
A bit too hacky :) Let's keep VBP Indicator code inside its module. `compare` (and `cumulative`) options should be ignored by VBP. Perhaps it's possible to remove/disable these options in `init()`?

kamil-musialowski:
After #16277 I think it should be possible to use it in VBP, so I don't think we should get rid of the compare option, what do you think?

pawelfus:
Good point! We should keep it. Now I hope for an implementation similar to this workaround: https://jsfiddle.net/BlackLabel/c2dohrta/ - but again, it looks a little hacky.

The best would be to simply.. remove the attached event for this particular series. Something like this:
```ts
// VBPIndicator.ts
removeEvent(VbPSeries, 'afterGetExtremes', DataModifyComposition.afterGetExtremes);
```

But I'm not sure if something similar is supported or should be avoided (because it would load the whole DataModify composition into the VbP indicator). Perhaps @bre1470 or @TorsteinHonsi can bring the answer to light

pawelfus:
Let's try it and see how many bytes will be added to `indicators-all.js`. Could you try it @kamil-musialowski?

kamil-musialowski:
Sure thing, I'll try it :)

kamil-musialowski:
After discussing the case with @pawelfus, we have discovered that it won't be possible to remove the event (since it's added on the Series prototype, not on the VBP prototype). Therefore, we have agreed to try the solution with overwriting the `getExtremes` method in VBP class.

ts/Stock/Indicators/VBP/VBPIndicator.ts
```diff
@@ -46,6 +46,7 @@ const {
 } = SeriesRegistry.seriesTypes;
 import U from '../../../Core/Utilities.js';
 import StockChart from '../../../Core/Chart/StockChart.js';
+import DataExtremesObject from '../../../Core/Series/DataExtremesObject';
```
pawelfus, line 49: Import type only

pawelfus (changes requested):
Tiny nitpicking, we can go with this solution, thanks!

ts/Stock/Indicators/VBP/VBPIndicator.ts
kamil-musialowski: Oh, I must have missed this one - thanks!

pawelfus (approved):
Thanks!

TorsteinHonsi (approved):

PR #19747, opened 2023-09-13
Simplified dashboards visual test workflow

goransle:
Combined the two job into one job. Fixes an issue with the visual-regression plugin saves files relatively, which was not picked up by `save-artifact`


bre1470 (approved):
Works in my branch. 🎉

karolkolodziej (approved):

Vazonik (approved):

PR #19748, opened 2023-09-13
bugfix/19725-drilldown-cartesian-to-non-cartesian

### LINKED ISSUE #19725: WordCloud does not work with drilldown
#### Actual behaviour
The `Wordcloud` series does not work with drilldown. After drilling up with breadcrumbs, the parent series is not displayed as it initially was (specifically - the yAxis labels are missing). 

**Demo:** https://jsfiddle.net/BlackLabel/h51a7yw4/2/

#### Product version
Highcharts 11

#### Affected browser(s)
Chrome, probably all?

hubertkozik:
Fixed #19725, drilling between cartesian and non-cartesian series didn’t work properly.

---
  - https://app.asana.com/0/0/1205463053195102


TorsteinHonsi (approved):
Thanks, looks good!

PR #19758, opened 2023-09-14
dash/confirmation-popup-close

karolkolodziej:
Fixed, it was impossible to change the confirmation popup close button.

### DIFF
--- a/ts/Dashboards/EditMode/ConfirmationPopup.ts
+++ b/ts/Dashboards/EditMode/ConfirmationPopup.ts
@@ -64,6 +64,11 @@ class ConfirmationPopup extends BaseForm {
         editMode: EditMode,
         options?: ConfirmationPopup.Options
     ) {
+        iconsURL =
+            options && options.close && options.close.icon ?
+                options.close.icon :
+                iconsURL;
+
         super(parentDiv, iconsURL);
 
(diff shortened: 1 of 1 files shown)


bre1470 (approved):
Looks good! 👍

ts/Dashboards/EditMode/ConfirmationPopup.ts
```diff
@@ -64,6 +64,11 @@ class ConfirmationPopup extends BaseForm {
         editMode: EditMode,
         options?: ConfirmationPopup.Options
     ) {
+        iconsURL =
+            options && options.close && options.close.icon ?
+                options.close.icon :
+                iconsURL;
```
sebastianbochan, line 70: `options?.close?.icon || iconsURL;` should be more transparent

sebastianbochan (changes requested):

ts/Dashboards/EditMode/ConfirmationPopup.ts
bre1470: But increases transpiled code. We looking into proper support for optionals, but at the moment we should not actively change conditions imho.

sebastianbochan (approved):

pawellysy (approved):

PR #19761, opened 2023-09-14
Tree.ts now doesnt throw error on missing parents

### LINKED ISSUE #15196: removing parent point in gantt chart does not work as expected
#### Expected behaviour
I expect the row for the data point to be removed and everything else shifted up/redrawn correctly.

#### Actual behaviour
It depends on how you remove the data point.

- If you do a `chart.update(newOptions);` the row is removed, but the collapse icon (the small triangle in the first column) is still shown on a non-parent data point.
- If you remove the point using any other method I could find (`series.removePoint()`, `point.remove()`, `series.setData()`, and `series.update()`), the rectangle in the gantt chart is removed, but the rest of the row's data, including the dependency arrows, remains.

I have commented out the various other methods I have tried.

https://codepen.io/brettk-dev/pen/yLVzvxB

#### Product version

Highgantt 9.0.1

#### Affected browser(s)

- Firefox 85.0.2
- Chrome 88.0.4324.182 (Official Build) (x86_64)
- Safari 14.0.3 (16610.4.3.1.4)

MarkusBarstad:
Fixed #15196, gantt chart threw error when nodes in the data set were orphaned.

---
  - https://app.asana.com/0/0/1205432171194783


TorsteinHonsi (changes requested):
Thanks, the fixes look good!

It now fixes part 1 of the issue, which was to render correctly in a static chart when a parent node was missing.

However part 2 still fails - the case when a parent node is dynamically removed. See https://jsfiddle.net/highcharts/ezufy4qc/ . 

I'd say we _can_ merge this right now, as it clearly fixes a bug, but it doesn't fully fix #15196.

ts/Gantt/Tree.ts
```diff
@@ -71,37 +71,40 @@ const getListOfParents = function (
...
-            delete (list as any)[parent];
+    const root = '';
+
+    Object.keys(listOfParents).forEach(
+        (node: string): void => {
+            if ((node !== root) && (ids.indexOf(node) === -1)) {
+                listOfParents[root] = [
+                    ...listOfParents[root],
+                    ...listOfParents[node]
+                ];
```
goransle, line 101: `listOfParents[root].push(...listOfParents[node])` [might be marginally quicker](https://www.measurethat.net/Benchmarks/Show/4223/0/array-concat-vs-spread-operator-vs-push#latest_results_block)

MarkusBarstad: Cool to know, we will push and not spread in that case! 

On a sidenote, I love that site, but I wish it had a more structured way of browsing it instead of just "the latest" :/

TorsteinHonsi:
Yes it does run without errors, but there are still issues. Apparently related to the ones in the OP, though the OP demo is lost. In your fiddle, after the update, we can see the following issues in the tree grid:

* There's a dead dependency triangle next to "Launch" 
* Clicking "New product launch" does nothing
* On mouseover and mouseout on the triangle next to "New product launch", there are JS errors
* The three orphaned categories do not receive bold font style like root elements should

For clarity, this is how it currently looks after the update...
Skjermbilde 2023-09-15 kl  19 05 59

... while we would expect it to look and function identical to what we get when we start out with the `relocate` node deleted:
Skjermbilde 2023-09-15 kl  19 06 59

MarkusBarstad:
Just an update here, its the TreeGrid-axis code.

TorsteinHonsi:
I put this back to draft mode to get it out of my "review" canban column. Please set it to Ready to review when you're done.

MarkusBarstad:
Another update here, (just to log progress on this bug):
`GridNode.depth` is not being updated which makes it so that `highcharts-treegrid-level-` does not reflect changes to the parent-child structure (at least not in my working case based off of the project-manager demo)

MarkusBarstad:
Note to self regarding the orphaned nodes not showing up on the `yAxis` (using the project-management `demo`):
- Categories are handled ok but;
- Nodes receive a `pos` outside of the `yAxis`-range and `yAxis` does not react to the changes.

Will deal with `labels` and `yAxis` first;
- [ ] Need to make sure the `yAxis` is updated to hold the labels, 
- [ ] Maybe we want to alter where orphaned nodes show up on the `yAxis`.

goransle (commented):
Looks good to me. Can we add a karma/cypress test for this?

goransle (approved):
Seems good to me 👍

TorsteinHonsi:
Thanks so far! The changes look good in general. Maybe you're aware of it, but I noticed a glitch here: https://jsfiddle.net/highcharts/kzhots93/

1. Click the Update button to re-id the Relocate node
2. Click the "New product launch" grid item => nothing happens

samples/unit-tests/gantt/gantt/demo.js
```diff
@@ -411,6 +411,16 @@
...
             'Collapsed graphics should not be rendered (#12617)'
         );
+
+        // Changing id of a parent
+        points[1].update({ id: '_moving' });
+
+        assert.strictEqual(
+            chart.series[0].yAxis.treeGrid.tree.children.length,
+            5,
+            'Orphaned nodes should appear as direct children of root'
```
hubertkozik, line 421: It would be good to add GH issue number if someone would like to get more information about this test.

ts/Core/Axis/TreeGrid/TreeGridAxis.ts
```diff
@@ -328,12 +328,16 @@ function getTreeGridFromData(
             start: number,
             result: Record<string, GridNode>
         ): Record<string, GridNode> {
-            const nodes = gridNode.nodes,
+
+            const
+                nodes = gridNode.nodes,
```
hubertkozik, line 333: Just indentation. I believe we are not making enters (new lines) after `const` or `let`.

hubertkozik (changes requested):
Good job! It's good, that there is no error in the console now. I think the bug mentioned by Torstein should be also fixed in that branch.

samples/unit-tests/gantt/gantt/demo.js
MarkusBarstad: True! Also I am not completely sure this one unit-test "covers it". 

I suppose visual tests should pick up on this stuff since the stuff in this fix deals with visual errors, but still idk what the praxis is here 😬

ts/Core/Axis/TreeGrid/TreeGridAxis.ts
MarkusBarstad: Alright, but I am afraid to tell you, Hubert 😬 , I think I may have been doing this for a while 😆

MarkusBarstad:
> Thanks so far! The changes look good in general. Maybe you're aware of it, but I noticed a glitch here: https://jsfiddle.net/highcharts/kzhots93/
> 
>     1. Click the Update button to re-id the Relocate node
> 
>     2. Click the "New product launch" grid item => nothing happens

Good catch!

TorsteinHonsi (changes requested):
Thanks, it works well now!

I would like to request that Hubert's initial comments were adressed:
* GH issue number reference in the tests (and perhaps in code?) will potentially save us time down the line.
* Go over and change back unnecessary whitespace changes. Newline after `let` and `const`, and there are also new empty lines, presumably remnants of inserting debug and console statements.

TorsteinHonsi (approved):

MarkusBarstad (commented):
LGTM 🚀

PR #19764, opened 2023-09-15
Added visual comparison gh action

goransle:
Replaces CircleCI visual comparison test. Also separates out file size comparison


TorsteinHonsi (approved):
Awesome 👏

bre1470 (approved):
Looks good. 👍 

I guess we can test it only after merge. It would be good to force a failing test then to check things out.

oysteinmoseng (approved):
🎉

PR #19774, opened 2023-09-18
docs/negative-log-axis-docs

hubertkozik:
Added information about plotting negative values on the logarithmic axis in docs.

---
  - https://app.asana.com/0/0/1205441089005974


TorsteinHonsi (approved):
Thanks, great explanation! Now we have a good resource to point to the next time this question comes up.

PR #19776, opened 2023-09-18
dash/a11y-light-dark-demo

karolkolodziej:
Improved the Light-Dark Demo accessibility.


samples/dashboards/demo/light-dark-theme/demo.css
```diff
@@ -38,3 +38,8 @@
         fill-opacity: 0.9;
     }
 }
+
+label {
```
sebastianbochan, line 42: Instead of overwriting global rules for labels, I would suggest to use `.inputs-wrapper label { }`

sebastianbochan (changes requested):

marvin19 (approved):

sebastianbochan (approved):

PR #19781, opened 2023-09-19
Fixed a few typos in the docs

bnzone:
Just fixing a few typos in the docs


docs/dashboards/layout-description.md
```diff
@@ -3,7 +3,7 @@ Layout
 
 An essential part of any dashboard is the positioning of its components. The layout can initially be defined from the configuration object, and then changed when Edit mode is enabled. Here are the building blocks of a dashboard layout.
 
-![layout-descripton-img.png](layout-descripton-img.png)
+![layout-description-img.png](layout-descripton-img.png)
```
TorsteinHonsi, line 6: This is a file path, so I'm afraid we can't change it without changing the file name as well. https://www.highcharts.com/docs/dashboards/layout-description

TorsteinHonsi (changes requested):
Thanks, much appreciated! One little change request.

docs/dashboards/layout-description.md
bnzone: No problem!

I believe the file path is what's in the `()` brackets and `[]` contain the alt text.

But I've updated both as per your suggestion for consistency 👍🏻

TorsteinHonsi (approved):

PR #19787, opened 2023-09-19
feature/19675-missing-annotations-drag-event

KamilKubik:
Added missing API reference on the `annotations.events.drag` event.


TorsteinHonsi (approved):
Thanks!

PR #19791, opened 2023-09-20
dash/add-component-button-disabled

Vazonik:
Disabled add component button when the cell toolbars are disabled. Added options for disabling the add component and rwd buttons.


sebastianbochan (approved):

karolkolodziej (approved):
I only think we should add a description of this behaviour to the API documentation.

bre1470 (approved):

ts/Dashboards/EditMode/EditMode.ts
```diff
@@ -737,27 +737,29 @@ class EditMode {
         // Create rwd menu
         this.createRwdMenu();
 
-        // Create add button
-        const addIconURL = options?.tools?.addComponentBtn?.icon;
+        // Create add component button
+        if (options.toolbars?.cell?.enabled) {
+            const addIconURL = options?.tools?.addComponentBtn?.icon;
```
sebastianbochan, line 742: `options` exists here, because of the reference in the IF.

sebastianbochan (changes requested):

ts/Dashboards/EditMode/EditMode.ts
```diff
@@ -1082,29 +1091,73 @@ namespace EditMode {
...
+        * Add Component button options.
+        */
+        addComponentBtn?: AddComponentBtn;
+        /**
+         * Whether the RWD buttons should be visible.
+         *
+         * @default true
+         *
+         */
+        enableRwdButtons?: boolean;
```
karolkolodziej, line 1107: As I think about it now, instead of `enableRwdButtons`  I would add an interface and option like 
`rwdButtons: RwdButtons` and inside place the `enabled` property and the `icons` as it is now with `addComponentBtn` 

Pseudo code:
```ts
RwdButtons {
   enabled?: boolean,
   icons: {
       ...
   }
}
```

It will be easier to scale this in the future if there is such a need.

karolkolodziej (changes requested):

ts/Dashboards/EditMode/EditMode.ts
Vazonik: Right, I didn't want to make changes to an already existing options in API, but I agree with that. I'll fix it.

karolkolodziej (approved):

sebastianbochan (approved):

PR #19792, opened 2023-09-20
dash/fix-es-master-dts

bre1470:
Fixed never types in es-modules master of Dashboards.

### DIFF
--- a/ts/masters-dashboards/dashboards.src.ts
+++ b/ts/masters-dashboards/dashboards.src.ts
@@ -34,11 +34,10 @@ import Utilities from '../Dashboards/Utilities.js';
 // Fill registries
 import '../Dashboards/Components/HTMLComponent.js';
 import '../Dashboards/Components/KPIComponent.js';
-import '../Data/Connectors/JSONConnector.js';
 import '../Data/Connectors/CSVConnector.js';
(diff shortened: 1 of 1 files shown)


sebastianbochan (approved):

ts/masters-dashboards/dashboards.src.ts
```diff
@@ -32,6 +32,9 @@ import Sync from '../Dashboards/Components/Sync/Sync.js';
 import Utilities from '../Dashboards/Utilities.js';
 
 // Fill registries
+import '../Dashboards/Components/HTMLComponent.js';
+import '../Dashboards/Components/KPIComponent.js';
+import '../Data/Connectors/JSONConnector.js';
 import '../Data/Connectors/CSVConnector.js';
 import '../Data/Connectors/JSONConnector.js';
```
khlieng, line 39: Its imported twice I think 🤔

karolkolodziej (approved):

ts/masters-dashboards/dashboards.src.ts
bre1470: Oops, C&P failure.

khlieng (approved):

PR #19800, opened 2023-09-20
bugfix/19736-scrollable-indicators

### LINKED ISSUE #19736: Scrollable list of Tech indicators in Stock Tools
#### Expected behaviour
It should be possible to scroll through list of available tech indicators in GUI

#### Actual behaviour
Chart zooms in & out instead 

#### Product version
Stock v11.1.0+, version v11.0.0+ are fine

#### Affected browser(s)
All

DJTechnoo:
Fixed #19736 , `mouseWheel` zooming prevented scrolling in Stock Tools popups.

### DIFF
--- a/ts/Extensions/MouseWheelZoom/MouseWheelZoom.ts
+++ b/ts/Extensions/MouseWheelZoom/MouseWheelZoom.ts
@@ -24,6 +24,7 @@ import type BBoxObject from '../../Core/Renderer/BBoxObject';
 import type { YAxisOptions } from '../../Core/Axis/AxisOptions';
 
 import U from '../../Core/Utilities.js';
+import DOMElementType from '../../Core/Renderer/DOMElementType';
 const {
     addEvent,
     isObject,
@@ -259,7 +260,6 @@ const zoomBy = function (
     }
 };
 
-
 /**
  * @private
  */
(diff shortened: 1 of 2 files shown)


ts/Extensions/MouseWheelZoom/MouseWheelZoom.ts
```diff
@@ -24,6 +24,7 @@ import type BBoxObject from '../../Core/Renderer/BBoxObject';
 import type { YAxisOptions } from '../../Core/Axis/AxisOptions';
 
 import U from '../../Core/Utilities.js';
+import DOMElementType from '../../Core/Renderer/DOMElementType';
```
pawelfus, line 27: I believe you can simply import `DOMElementType` type 🤔

pawelfus (changes requested):
Minor improvements, otherwise looks good!

kamil-musialowski (changes requested):
Looks good and works good! 

The original issue is resolved, however one still remains:
- add any label, line, annotation (whatever) to the chart,
- click on it, then click on the pencil icon to edit it,
- try to scroll the edit popup

I don't think it makes any sense to create a separate issue for that, so could you also look into that @DJTechnoo?

TorsteinHonsi (approved):
Thanks!

pawelfus (changes requested):
Looks good now, but if you could check the suggestion by Kamil, that would be great. If it's a simple fix, let's include it in this PR

DJTechnoo:
I am adding a fix very soon

pawellysy (approved):

pawelfus (approved):
Thanks!

PR #19812, opened 2023-09-21
demo/navigator-set-maskinside-false

kamil-musialowski:
Moved `navigator.maskInside` sample to the main demo page.

Asana: https://app.asana.com/0/1203542399740129/1204308269116701/f


pawelfus (changes requested):
Thanks, looks good!

Could we also change min-width to 310px? To match other demos. Sorry for missing this in the first review!

TorsteinHonsi (approved):
Thanks!

Issue #19205, opened 2023-06-23
Cannot find module 'highcharts' or its corresponding type declarations. ts(2307)


LuckyWindsck:
#### Expected behaviour
Being able to find module 'highcharts' or its corresponding type declarations.

#### Actual behaviour
I used `create-vue` to create a `Vue3` project with `Vite` and `TypeScript` support.
When I try to import `highcharts` in a Vue [Single-File Components](https://vuejs.org/guide/scaling-up/sfc.html), VSCode shows me the following ts error:
> Cannot find module 'highcharts' or its corresponding type declarations. ts(2307)

#### Live demo with steps to reproduce
I made a [minimal reproducible example](https://github.com/LuckyWindsck/minimal-vue-typescript-highcharts/).

1. Clone the repository, install dependencies, and open VSCode
```sh
$ git clone https://github.com/LuckyWindsck/minimal-vue-typescript-highcharts.git
$ cd minimal-vue-typescript-highcharts
$ yarn install
$ code . # open with visual studio code
```
2. Open `src/App.vue` and hover your mouse pointer over the `'highcharts'` in `import highcharts from 'highcharts'`, then VSCode will show ts error.

![importing-highcharts-in-vue-sfc-shows-ts-error](https://github.com/highcharts/highcharts/assets/24631178/542536fd-f51a-4f1d-8618-fc9280005520)

Btw, there's no problem with other dependencies like `axios` in the minimal reproducible example.

![importing-axios-in-vue-sfc-shows-no-error](https://github.com/highcharts/highcharts/assets/24631178/89022e67-1a8b-47fa-a410-fecbadced186)

Also, this problem is only found in Vue SFC. I can import `highcharts` without any error in normal `.ts` file

![importing-highcharts-in-ts-file-shows-no-error](https://github.com/highcharts/highcharts/assets/24631178/beb28057-6c71-48fe-9457-2b7120aa217a)

##### Visual Studio Code
```sh
$ code --version
1.79.1
4cb974a7aed77a74c7813bdccd99ee0d04901215
x64
$ code --list-extensions --show-versions | grep Vue
Vue.volar@1.8.1
Vue.vscode-typescript-vue-plugin@1.8.2
```

##### Dependencies
```json
{
  "dependencies": {
    "axios": "^1.4.0",
    "highcharts": "11.1.0",
    "vue": "3.3.4"
  },
  "devDependencies": {
    "@vue/tsconfig": "0.4.0",
    "typescript": "5.0.4"
[...]

Issue #19063, opened 2023-05-30
highcharts.d.ts is not a module


kkesidis:
I upgraded Highcharts from version `10.3.2` to `11.0.1` and i have issues with typescript not recognizing the `highcharts.d.ts` as a module. I was able to reproduce this error on a newly created application, too

#### Expected behaviour
Being able to import the library on a typescript project
```ts
import Highcharts from 'highcharts'
import type {Chart, Options} from 'highcharts';
```

#### Actual behaviour
Getting an error
> File '<local path>/node_modules/highcharts/highcharts.d.ts' is not a module

I can still use the library, but typescript is not working

#### Live demo with steps to reproduce
[Github repository](https://github.com/kkesidis/highcharts-11-reproduction)

The above repository is a freshly created app using Vite and the vue-ts template

#### Product version
v11.0.x

#### Affected browser(s)
None

karolkolodziej:
Hi @kkesidis,

In order to omit that error, You can import Highcharts from a master directory like:
```
import Highcharts from 'highcharts/es-modules/masters/highcharts.src'
```

@bre1470 any other insights?

kkesidis:
Thank you, this works. Closing this issue, then

PR #19814, opened 2023-09-21
docs/vue-volar

bre1470:
Docs: Added note for Volar setting as seen in #19063, #19205.

### DIFF
--- a/docs/dashboards/dashboards-with-vue.md
+++ b/docs/dashboards/dashboards-with-vue.md
@@ -33,9 +33,11 @@ To create a dashboard with Vue, please follow the steps below: <br>
     ```
 
     __Please Note:__ If you are using the Visual Studio Code editor with the
-    Volar extension, you should change the editor setting
+    Volar extension, you should change the extension setting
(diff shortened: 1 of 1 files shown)


jakubSzuminski:
LGTM.

This is not crucial, but I would consider including an additional sentence which would instruct how to increase the max file size (go to: `File` -> `Preferences` -> `Settings` and change the `Vue: Max File Size`) if you think that users find it helpful.

jakubSzuminski (approved):
LGTM!

This is not crucial, but I would consider including an additional sentence which would instruct how to increase the max file size (go to: `File` -> `Preferences` -> `Settings` and change the `Vue: Max File Size`) if you think that users may find it helpful.

karolkolodziej (approved):

Vazonik (approved):

bre1470:
@jakubSzuminski I added a second sentence to describe the menu path. Please check and feel free to change in a commit, if it needs improvement.

TorsteinHonsi (approved):

PR #19815, opened 2023-09-21
bugfix/19762-offline-export-pdf-scale

### LINKED ISSUE #19762: Scale doesn't work for PDF offline exporting
#### Actual behaviour
When I use the offline-exporting module, I noticed that: the option `exporting.scale` didn't work for PDF type. The PDF file always has width/height same as parent container, without scale.
Other types (PNG/JPEG) works properly.

#### Expected behaviour
The option `exporting.scale` should work for PDF type in offine-exporting, same as other types.

[https://jsfiddle.net/37zb2h68/9/](https://jsfiddle.net/37zb2h68/9/)

#### Product version
Highcharts 10.x

#### Affected browser(s)
All browsers

KamilKubik:
Fixed #19762, offline export PDF documents didn't scale.

### DIFF
--- a/samples/highcharts/exporting/offline-download-scale/demo.css
+++ b/samples/highcharts/exporting/offline-download-scale/demo.css
@@ -1,0 +1,5 @@
+#container {
+    max-width: 800px;
+    height: 400px;
+    margin: 0 auto;
+}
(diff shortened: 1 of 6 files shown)


TorsteinHonsi (approved):
Thanks!

PR #19825, opened 2023-09-22
bugfix/19552-treegraph-export-parent-node-lost

### LINKED ISSUE #19552: Treegraph loses root node at exporting to svg, png and etc.
Good day!
Found that we lose root node at exporting to SVG, png if data is updated by both `series.setData` or `series.update`.
We have a chart with dynamically adding data [here](https://jsfiddle.net/0d8akt72/9/)

#### Expected behaviour
Expect to have a similar tree in the exported png.
image

#### Actual behaviour
Currently, the chart loses the root node at exporting to SVG or png and we have this result:
image

https://jsfiddle.net/0d8akt72/9/

#### Product version
Highcharts: 11.1.0

#### Workaround
I did some research and found that the chart use `series.userOptions.data` in `getSVG` method.
And figured out that the collection `series.userOptions.data` is broken. Possible, it is related with [this bug](https://github.com/highcharts/highcharts/issues/19524).
But I made a workaround that fixes exporting:
```javascript
const getSVG = (H) => {
  H.wrap(H.Chart.prototype, 'getSVG', function (proceed, chartOptions) {
    this.series.forEach((series) => {
      if (series.type !== 'treegraph') {
        return;
      }

      series.userOptions.data = series.data.map((point) => ({
        ...point.options,
      }));
    });
    return proceed.call(this, chartOptions);
  });
};
```

You can uncomment a line with using the workaround in the above link or use this [link](https://jsfiddle.net/0d8akt72/10/)

KamilKubik:
Fixed #19552, treegraph was losing node parent during export.

We're updating the point to keep the _collapsed_ property updated. Previously, the _update_ was on the _linkToParent_ which was a bit misleading because the _collapsed_ property exists on point (not link) which was updated anyway, instead of the link (that's the correct behaviour, the main purpose is to update a point). That was causing the issue because we were updating points but using the indices from links, so the parent node was being overwritten. The issue was visible only in exported files because in the provided example (in the [issue](https://github.com/highcharts/highcharts/issues/19552)), the data is generated in intervals, and that overwrite happens after rendering the parent node.

The existing _update_ method was used only inside the _getLinks method,_ so it should be safe to remove it.


pawellysy:
The fix itself makes sense. Could you add some unit test tho?

pawellysy (changes requested):
The fix itself looks correct. Could you add the unit-test tho?

KamilKubik:
> The fix itself looks correct. Could you add the unit-test tho?

Thanks, I added a test that checks both `id` and `parent` to verify the data is rendered correctly.

pawellysy (approved):
LGTM!

TorsteinHonsi (changes requested):
Thanks, it looks good!

Can you fix the merge conflicts too?

TorsteinHonsi (approved):

TorsteinHonsi:
_Internal note_

This fix created a regression in tree graphs: cannot collapse two nodes at the same time. Reverting and awaiting fix for the fix.

PR #19827, opened 2023-09-25
bugfix/17014-boostTreshold-null-point-error

### LINKED ISSUE #17014: Boost error: Cannot read properties of null (reading '0')
#### Expected behaviour
No error

#### Actual behaviour
Error out with Cannot read properties of null (reading '0')

https://jsfiddle.net/soetji/j96dmf45/

Open the browser console to see the error message.

It will work if you set in the code:
opts.boost.enabled = false;

#### Product version
Latest

KamilKubik:
Fixed #17014, `boostTreshold` with null point threw an error.


TorsteinHonsi (approved):
Thanks!

PR #19831, opened 2023-09-25
Migration: GeoJSON

bre1470:
- Moved GeoJSON as composition to Maps
- Fixed BrowserStack Safari test in GHA


TorsteinHonsi (approved):
Thanks!

PR #19834, opened 2023-09-26
Migration: Maps/TilesProviders

bre1470:
Requires merge of #19833


bre1470:
@TorsteinHonsi Please note that I corrected the name of the TilesProviderRegistry. Before it was TilesProvider__s__Registry which is inconsistent with the rest of registries and as it is not an official property only a few demos are affected.

TorsteinHonsi (changes requested):
Thanks!

Now (probably after merging #19833) we have many conflicts.

Also, the visual diff reports two differences, but https://vrevs.highsoft.com/pr/19834/review reports some server side problem. Do we know what's causing that? I've seen it on other recent test runs too. I think it's a bit risky to merge before we can see the visual diff, it may come back to bite us later.

TorsteinHonsi (approved):

PR #19837, opened 2023-09-27
Migration: Gantt/Pathfinder

bre1470:
Requires merge of #19837


TorsteinHonsi (changes requested):
Now there are conflicts since I merged https://github.com/highcharts/highcharts/pull/19837 . 

Actually this PR contains the changes from the other one, so the note that this _requires_ the other one is a bit confusing. Maybe in cases like this, when one PR builds upon another, it would be better to delete the first one? I mean, apparently merging this one would be enough to get the changes from both.

bre1470:
I thought that this is easier for you to review. What puzzles me is that these conflicts occur at all. It should just continue at the same commits on master.

TorsteinHonsi:
> What puzzles me is that these conflicts occur at all. It should just continue at the same commits on master.

@bre1470 Yes that puzzles me too. If it behaved like that it would indeed be easier to review.

TorsteinHonsi (approved):

PR #19850, opened 2023-09-28
Migration: Extensions/BoostCanvas

bre1470:


bre1470:
Boost canvas is not part of boost module anymore (which seemed to happen by accident). Now it only is a fallback if 'boost-canvas' is loaded before `boost`.

TorsteinHonsi (approved):
Thanks!

PR #19854, opened 2023-09-29
Migration: Maps/MapView

bre1470:


ts/Maps/MapViewDefaults.ts
```diff
@@ -0,0 +1,363 @@
...
+
+    /**
+     * Generic options for the placement and appearance of map insets like
+     * non-contiguous territories.
+     *
+     * @since        10.0.0
+     * @product      highmaps
+     * @optionparent mapView.insetOptions
+     */
+    mapViewInsets: {
```
TorsteinHonsi, line 212: Since this structure is resolved to `mapView.insetOptions` (as seen in the `@optionparent`) tag and [in the API](https://api.highcharts.com/highmaps/mapView.insetOptions), wouldn't it make more sense to place it in that position in the `MapViewDefaults` file too? Then we wouldn't need the `@optionparent` tag, and the top-level `mapView` property?

TorsteinHonsi (commented):
Thanks! See inline comment.

ts/Maps/MapViewDefaults.ts
bre1470: I tried it and it was failing, because the defaults contains the properties for `mapView.insets` and not `mapView.insetOptions`. So the doclets are only partly matching. We could reshuffle the doclets, but I am unsure if this would stop the confusion.

TorsteinHonsi (approved):

PR #19689, opened 2023-09-05
Added accessibility support for navigator handles.

oysteinmoseng:
Added accessibility support for navigator handles.
___
Keyboard nav and basic screen reader support should now work for most screen readers/browsers. Tweaks can be done from this point as we collect user feedback.

Proxies the handles using `<input type="range">` elements, indicating axis percentage from 0 to 100. Changing the values updates the navigator through simulating mouse drag, which is not ideal, but will do for now until Navigator is refactored into its own module. At that point we should define a two-way API for the Navigator.

Closes #18216.

PR #19856, opened 2023-10-01
bugfix/18716-export-menu-redraw-a11y

### LINKED ISSUE #18716: Export menu stops working with accessibility module with frequent chart redraws
Our app's use case involves updating charts many times per second as a real-time simulation runs and data is gathered.  Static charts do not experience this issue.  I hesitated a bit on submitting this as a bug as it's probably not the typical usage for highcharts, but since we do get the behavior we want without accessibility enabled it seemed close enough to submit it and see.

#### Expected behaviour

Without accessibility enabled we have the triple-bar menu clickable at all times with options to export images or CSV data.

#### Actual behaviour

With accessibility enabled the triple-bar menu doesn't respond most of the time.

[Here is a demo that shows the issue](https://jsfiddle.net/el_tranquilo/2z9pf8qo/27/).  If you comment out the accessibility script in the HTML you can see the export button works again while the data is being generated.

#### Product version

Highcharts 10.3.1 and 10.3.3 that I've tested.

#### Affected browser(s)

Chrome and Firefox.

jedrzejruta:
Fixed #18716, stop export menu proxy button reposition on data change.


oysteinmoseng (changes requested):
Hm, I wonder if this will cause issues if we e.g. add one more series, and the legend buttons then need to change positions?

Also in the MenuComponent we are redrawing the export button proxy anyway on every chart render: https://github.com/highcharts/highcharts/blob/master/ts/Accessibility/Components/MenuComponent.ts#L189

Could you investigate and see if there is another way of fixing this, where we still catch updates for other proxy elements, and where we still support for example calling `chart.update` and changing the position of the menu button?

jedrzejruta:
I'll investigate it further to find possible issues in this approach. 

I'm aware of MenuComponent redraw on each chart render, but from what I see it doesn't affect the mentioned issue (at least on master branch, since #19689 PR). Before this PR, actually preventing the redraw on MenuComponent seemed to solve the problem.

I'll try to find another way of fixing this and will keep you updated on the progress.

jedrzejruta:
@oysteinmoseng any thoughts on this PR?

oysteinmoseng (changes requested):
Good! Minor change request only.

Now, the current solution will still cause issues if people do
```javascript
    chart.update({
        // some a11y options or menu placement options
    })
```
but I think that's an edge case, and we can ignore it for now.

Ideally we just want to skip handling the re-renders that are caused because of new data, and nothing else, but to my knowledge there is no easy way to detect that currently (@TorsteinHonsi any ideas?)

Maybe we could hook into the new data event, and also the update event, look at options and set flags accordingly, but it sounds a bit messy, so I am inclined to wait unless there is an easier way.

EDIT: Or maybe we should do the handling on update, and the handling on render separately 🤔 Doing the handling on update always, and on render if resized/first render. Perhaps investigate this a bit further, there might be a simple solution.

TorsteinHonsi:
> Ideally we just want to skip handling the re-renders that are caused because of new data, and nothing else, but to my knowledge there is no easy way to detect that currently (@TorsteinHonsi any ideas?)

We have the `updatedData` event. I see we have code already that [checks for the trigger](https://github.com/highcharts/highcharts/blob/312dd7dd9cf74c72bdadcb0ebe6ee31838f26e3d/ts/Stock/RangeSelector/RangeSelector.ts#L476). But I'm not sure if that is consistently implemented. If not, we could add an event listener for `updatedData` that sets some flag, then reset the flag when the redraw is gone.

oysteinmoseng:
Looking at it, I think the simplest solution is to recreate the proxies on chart update in addition to the current code, where we update on resize & initial load. There is already an inherited `onChartUpdate` function that can be overridden in the `MenuComponent` class. This function should not be called on new data, I think? But let's double check that..

jedrzejruta:
Thanks for all the insights regarding this issue. I must've missed your comments, getting back to this PR soon.

PR #19866, opened 2023-10-03
tools/github-actions-nightly-dist

goransle:
* Added job for distributing nightly branch on highcharts-dist
* Browserstack tests are now only triggered nightly
* Added Cypress and ts-node tests to highcharts-headless workflow

<details>
<summary><h2>ToDone</h2></summary>

- [x] Add missing tests to nightly workflow
- [x] Separate out dispatch to job that requires tests to pass


bre1470 (approved):
I suggest to remove the climate demo from any visual test to make test results reliable. Otherwise chances increase that real errors get missed, because one gets used to the workflow failing.

TorsteinHonsi (approved):
Thanks!

PR #19875, opened 2023-10-04
dash/19707-hide-settings-button

### LINKED ISSUE #19707: Highcharts Dashboards - Need to remove settings option from HighCharts Dashboard in edit mode
# Description of the feature
Need to remove Setting option from highchart dashboard when its in "Editmode"

Highcharts Dashboard

1) How to remove options in the settings menu in Highchart dashboard
image

2) How to remove settings icon while enable the 'Edit mode'
image

Vazonik:
Added the `settings.enabled` option. Closes #19707.


ts/Dashboards/EditMode/Toolbar/CellEditToolbar.ts
```diff
@@ -87,36 +88,39 @@ class CellEditToolbar extends EditToolbar {
...
-                                    popup.closePopup();
-                                }
-                            },
-                            text: editMode.lang.confirmDestroyCell
-                        });
-                    }
+            });
+        }
+
+        items.push({
```
karolkolodziej, line 94: At this point, we can implement in a similar fashion the `destroy` button 😉 
@oysteinmoseng do you think it makes sense?

ts/Dashboards/EditMode/EditMode.ts
```diff
@@ -108,7 +108,8 @@ class EditMode {
                     },
                     row: {
                         enabled: true
-                    }
+                    },
+                    settingsButtonEnabled: true
```
karolkolodziej, line 112: I'm not convinced if this should be here...

IMO in the `toolbar` we have options for the toolbar itself.
The setting button, at least for me, is something like the `dragDrop` or `resize` and it should have a separate config object like:
```js
                dragDrop: {
                    enabled: true
                },
                resize: {
                    enabled: true
                },
                settings: {
                    enabled: true
                },
```

Altho I can imagine in the future someone would like to disable separately for the cell/row...
But let's not get too far 😅

karolkolodziej (changes requested):

karolkolodziej (approved):

sebastianbochan (approved):

PR #19890, opened 2023-10-05
dash/remove-threshold-component

Vazonik:
Removed threshold component.


samples/dashboards/components/custom-threshold-component/demo.js
```diff
@@ -0,0 +1,166 @@
...
+            options: {
+                subtitle: `Options from previous thresholds, if not limited by
+                the 'min' option or overwritten, are passed to the next
+                thresholds.`
+            }
+        }, {
+            min: 80,
+            options: {
+                style: {
+                    background: 'blue'
```
karolkolodziej, line 130: This colour is super bright 😅 can we please use something a bit more pale?

karolkolodziej (approved):

samples/dashboards/components/custom-threshold-component/demo.js
sebastianbochan: I prefer to use hex instead of "names" of colors.

docs/dashboards/custom-component.md
```diff
@@ -163,4 +164,130 @@ Dashboards.board('container', {
     }]
 });
 ```
-Check out the [live example here](https://www.highcharts.com/samples/embed/dashboards/components/custom-html-component).
\ No newline at end of file
+Check out the [live example here](https://www.highcharts.com/samples/embed/dashboards/components/custom-html-component).
+
+## Custom Threshold Component
+In some cases, you may want to create a component that behaves as if it changes its type or/and options depending on certain conditions. Such a condition may be, for example, value. The example below shows how to program a custom so-called Threshold Component, which is presented below.
```
sebastianbochan, line 170: We should not use "behaves" for artificial elements. Maybe "act" or "works"?

docs/dashboards/custom-component.md
```diff
@@ -163,4 +164,130 @@ Dashboards.board('container', {
...
+        // Selecting appropriate options and components based on thresholds
+        // and given value.
+        if (thresholds && U.isNumber(value)) {
+            for (let i = 0; i < thresholds.length; i++) {
+                const threshold = thresholds[i];
+
+                if (U.isNumber(threshold.min) && value < threshold.min) {
+                    continue;
+                }
+                if (U.isNumber(threshold.max) && value > threshold.max) {
```
sebastianbochan, line 227: Instead of 2 conditions, you can use "or".

docs/dashboards/custom-component.md
```diff
@@ -163,4 +164,130 @@ Dashboards.board('container', {
...
+                    continue;
+                }
+
+                componentOptions = U.merge(componentOptions, threshold.options);
```
sebastianbochan, line 231: Instead of reference to "U" in some places, use the destruction. like

```
const { merge, isNumber } = U;
```

sebastianbochan (changes requested):

sebastianbochan (approved):

PR #19896, opened 2023-10-06
bugfix/19604-ticks-calculation-after-update

### LINKED ISSUE #19604: Ambiguous behavior of the axis labels/ticks calculation procedure
#### Expected behaviour
Axis labels and ticks should not change unless the values that affect their calculation (chart size, font size, axis range etc..) remain constant.

For user/programmer it's expected that the tick interval calculation is provided by an internal Highcharts' procedure which gives an unkwonw in advance but the same distance between the ticks. That is, this distance must always be the same no matter in which way the axis is drawn - whether as a result of the initial chart appearance or after its update.
In other words, using the programmer's lexicon - the calculation procedure should be a "pure function" (one that always returns the same result with the same input arguments). 

#### Actual behaviour
Vertical axis labels and ticks changed.

[demo:](https://jsfiddle.net/htk8oa20/2/)
1. Run the demo.
2. Note the vertical axis labels and ticks. They are **25/50/75**
image
4. Click the "Update" button to change series color.
5. Series color is changed from blue to green, but the vertical axis labels and ticks are also changed (recalculated). Now they are **20/40/60**
image

#### Product version
11.1.0

#### Affected browser(s)
Chrome 115.0

hubertkozik:
Fixed #19604, ticks were sometimes differently calculated after a series update.

---
  - https://app.asana.com/0/0/1205344683223009


TorsteinHonsi:
Thanks @hubertkozik !

Before continuing on this fix, I'd like to share some background. The reason why we are doing first the temporary plot height, then redo horizontal/vertical logic, is that at the time of rendering the vertical axis, we don't know how much space the labels of the horizontal axis takes. Now since the horizontal axis labels may or may not be rotated (depending on its options), the vertical axis may be shorter than first assumed. As a result we may have to recalculate the tick interval in order for the `tickPixelInterval` setting and other layout concerns to be respected. 

However, this process is quite organic - the x-axis labels affect the length of the y-axis, which affects the y-axis label positioning, which in turn may affect the x-axis length, leading to new label positioning, and so forth. The cost of this process is that each iteration slows down the performance of the rendering. Each iteration requires DOM traffic in order for the labels to be rendered and measured. For that reason, we stop at two iterations, and we do the second iteration only if the length of the axis changes by a factor of more than 1.05. 

Now to this issue. For the x-axis, in most cases, it is laid out with a single row (`staggerLines`) of labels which are not rotated (`autoRotation` and `rotation`) settings. For that reason, we can make an assumption that the x-axis labels will take up a certain amount of pixels. That's what we are doing here:

https://github.com/highcharts/highcharts/blob/b5f5104e1bad58c8d3be42f2c4ca546906847eb6/ts/Core/Chart/Chart.ts#L2414-L2415

Now the **first problem** is that we changed the axis label `distance` in v11, so the default correction is no longer 21. According to my tests in Chrome, it is more like 30 in v11.

The **second problem** is that it is not aware of various axis settings. In the OP demo, the implementer sets `reserveSpace` to `false`, which effectively throws off our assumption of 21 (which now should. be 30). Other settings would throw it off too, like changing the font size of the labels, or the `distance`. 

I think we should look into a slightly more dynamic way of finding the correction. Maybe we should rename `correction` to `expectedSpace`, to make it more comprehensive what we're dealing with. It should be possible to compute a better approximation by taking some options into account, without it becoming a bloat. Correcting for `reserveSpace`, `distance` and `style.fontSize` would be a good start. To get a true approximation for `fontSize` we will have to temporarily create a text element and measure it, otherwise CSS and relative sizes won't apply. Maybe creating a dummy `Tick` instance would take care of the whole thing?

hubertkozik:
@TorsteinHonsi I have added changes to calculate the `expectedSpace` dynamically. After those changes, we've got 54 visual differences. Most of them seem legit to me, but there are also strange differences with random pixels - don't know what is it specifically. 

Regarding legit visual differences: through all of this time, it was a bug in these charts, that after `series.update` ticks were changed (relative to the initial render). You can see it e.g. in this demo: https://jsfiddle.net/BlackLabel/btgp5so8/ update is changing y-axis ticks -> bug. The rest of the visual differences behave the same as this one.

TorsteinHonsi (commented):
Nice work!

> What do you think about the current solution? Would you add some more conditions or something else?

It's exactly what I had in mind. We can probably trim it a bit. Simplify some stuff and maybe add some conditions, but overall it's a great starting point.

I agree with the visual diffs, they were buggy earlier but now are good. I will look more into detail for the final review. There are some strange cases with polar charts and axis breaks too, where the ticks are now denser. While in the regular cartesian diffs, the ticks are now more sparse. Maybe those were also buggy in the first place.

Some things to address except the inline comments:
* Are the unit tests valid for all cases? Since this recalculation of the axis only happens on the first chart rendering, maybe we need to create new chart instances for every combination of options that we want to test? And I do think we should have tests for all known cases, otherwise we risk repeating the same mistake as I did when I changed the default `distance` without changing the `correction`.
* On an x-axis with no categories, the `tickLength` option affects the plot height. But apparently only when it is larger than 10, or when `reserveSpace` is false. 
* If the `offset` option is positive, it affects the expected space. If it is bigger than ~20, it triggers recalculation.
* If we set `style.fontSize` to 3em, the prediction is 11px off in my browser. Not enough to trigger a recalculation, but still. This may be the same issue as the fact that...
* In styled mode, the bounding box size is not detected correctly. If [I increase the font size](https://jsfiddle.net/highcharts/9k27nLxa/) to 3em, the prediction is 35 px off. Maybe the mock tick is not inserted in the correct SVG structure, so the CSS doesn't cascade correctly?

ts/Core/Chart/Chart.ts
```diff
@@ -2405,20 +2403,81 @@ class Chart {
         const tempWidth = chart.plotWidth;
 
         axes.some(function (axis: Axis): (boolean|undefined) {
```
KacperMadej, line 2405: Since the calculation of the `expectedSpace` are more elaborate now maybe instead of `axes.some` there should be done a search for a max `expectedSpace` across all axes? There might be more than 1 axis that will cause `expectedSpace` to be calculated and that 1 might not produce the biggest value causing a redo or unstable render.

ts/Core/Chart/Chart.ts
```diff
@@ -2405,20 +2403,81 @@ class Chart {
...
             ) {
-                // 21 is the most common correction for X axis labels
-                correction = 21;
-                return true;
+                // Create the axisGroup and gridGroup elements on first
+                // iteration
+                if (!axis.axisGroup) {
+                    if (axis.coll === 'colorAxis') {
+                        axis.axisParent =
+                            axis.legendItem && axis.legendItem.group;
```
KacperMadej, line 2429: Nitpick: `axis.legendItem?.group`

ts/Core/Chart/Chart.ts
hubertkozik: I am not sure if we should address this issue at that moment. After my investigation, calculating `expectedSpace` for multiple axes is really complicated and probably will need some time/performance to calculate everything correctly. Also, I think, if we would like to calculate it, `expectedSpace` should be the sum of all expected spaces for every axis, but keeping in mind options like `axis.offset`/`axis.opposite`. Now it is handled by the additional trigger of recalculation (`redoVertical`).

KacperMadej (approved):

ts/Core/Chart/Chart.ts
```diff
@@ -2405,20 +2403,80 @@ class Chart {
...
+                    axis.axisGroup = axis.createGroup(
+                        'axis',
+                        '',
+                        axis.options.zIndex
+                    );
+                    axis.labelGroup = axis.createGroup(
+                        'axis-labels',
+                        '-labels',
+                        axis.options.labels.zIndex
+                    );
```
TorsteinHonsi, line 2445: This code block is in principle identical to this:

https://github.com/highcharts/highcharts/blob/c286688e549da6aa1080ba649d9b1af7d16b54b6/ts/Core/Axis/Axis.ts#L3478-L3494

It should be refactored out so that the `Axis.createGroup` function takes care of adding all the groups, and we just have to call it once. Then just one single line will be repeated instead of the whole block.

ts/Core/Chart/Chart.ts
```diff
@@ -2405,20 +2403,80 @@ class Chart {
         const tempWidth = chart.plotWidth;
 
         axes.some(function (axis: Axis): (boolean|undefined) {
+            // Left side must be align: right and right side must
+            // have align: left for labels
+            axis.reserveSpaceDefault = (
+                axis.side === 0 ||
+                axis.side === 2 ||
+                { 1: 'left', 3: 'right' }[axis.side] === axis.labelAlign
+            );
```
TorsteinHonsi, line 2412: I'm not sure if this was answered, but do we need repeating this code block? Since the whole logic of expected space only applies to horizontal axes (`axis.side` is always 0 or 2), this condition will always resolve to `true`.

TorsteinHonsi (changes requested):
Thanks! A couple more comments.

ts/Core/Chart/Chart.ts
hubertkozik: I felt, that I had forgotten about something... I agree that's true and shouldn't be here.

TorsteinHonsi (approved):

PR #19901, opened 2023-10-09
docs/19480-plotlineorband-missing-properties

### LINKED ISSUE #19480: Typings in YAxisPlotLinesLabelOptions.formatter not correct
#### Expected behaviour
_this_ should be of a type containing all properties which are available at runtime, namely _axis_, _id_, _options_, svgElem

#### Actual behaviour
Type is PlotLineOrBand which contains _label_ and _svgElem_

https://jsfiddle.net/stormtroopa/cb2tnfem/2/

#### Product version
Highstock 11.1.0

#### Affected browser(s)
affects VS Code

jedrzejruta:
Fixed #19480, added missing TypeScript definitions for `PlotLineOrBand` properties.
Removed `dts` script from `package.json`, as the defined commands weren't present in any of the gulp scripts.


TorsteinHonsi (approved):
Thanks!

PR #19902, opened 2023-10-10
Feature/stock benchmark demos

pawellysy:
Implemented stock benchmark tests.


.github/workflows/stock-benchmark.yml
```diff
@@ -0,0 +1,82 @@
+name: Stock benchmark
+on:
+  pull_request:
+    branches:
+      - 'master'
+    paths:
+      - 'ts/Stock/**'
+      - 'ts/Data/**'
```
pawelfus, line 8: No need for data-modules, right?

package.json
```diff
@@ -127,5 +127,8 @@
       "git add"
     ]
   },
-  "license": "SEE LICENSE IN <license.txt>"
+  "license": "SEE LICENSE IN <license.txt>",
+  "dependencies": {
+    "latest": "^0.2.0"
```
pawelfus, line 132: Hm?

test/ts-node-unit-tests/benchmarks/Stock/Stock-DataGrouping-Ordinal.bench.ts
```diff
@@ -0,0 +1,88 @@
+import type { BenchmarkContext, BenchmarkResult } from '../../benchmark';
+import { performance } from 'node:perf_hooks';
+import { join } from 'node:path';
+
+
+export const config = {
+    sizes: [1000, 10_000, 100_000]
+};
+
+export function generateOHLC(size: number){
```
pawelfus, line 10: Shall we move it to a separate file, with data generators? Instead of other tests depending on this test

pawelfus (changes requested):
Looks good, just a few things

goransle (approved):
Looks good, except for other requested changes

.github/workflows/stock-benchmark.yml
pawellysy: idk, we might wanna remove it, if you think so.

pawelfus: I would say keep dependencies at a minimum

DJTechnoo (approved):
Excellent job!

pawelfus (approved):
Thanks!

kamil-musialowski (approved):

TorsteinHonsi (approved):

PR #19903, opened 2023-10-10
Added min-height and min-width to flags

goransle:
## Test plan:

Run the following commands 10-15 times:
```sh
npx karma start test/karma-conf.js --tests "highcharts/demo/column-comparison" --single-run --splitbrowsers Chrome --reference
npx karma start test/karma-conf.js --tests "highcharts/demo/column-comparison" --single-run --splitbrowsers Chrome --visualcompare

```


goransle:
Following the test plan again after committing I got a pixel diff 😢

PR #19906, opened 2023-10-10
bugfix/19475-mouse-wheel-zoom-empty-space

### LINKED ISSUE #19475: Mouswheel zooming module does not respect customized xAxis range
#### Expected behaviour
The mouse wheel zooming should use the xAxis range instead of the data range for zooming into the the chart.

#### Actual behaviour
Currently the zooming module uses the data min/max range. This will cause a jumping behaviour between the data min/max and xAxis min/max range.

https://jsfiddle.net/mb3k6dpc/

#### Product version
Highcharts/Highstock v11.1.0

#### Affected browser(s)
Edge (Windows, 115.0.1901.188 64bit)

### LINKED ISSUE #19779: Mousewheel zoom stops working if any axis has no data points visible at the current zoom level
#### Expected behaviour
You should be able to zoom in or out with a mousewheel, irrespective of what data is visible on any given axis. If a given axis has no data visible, it should not break subsequent mousewheel zooming, on it or other axes.

#### Actual behaviour
If you've mousewheel zoomed to a point where one axis has no data visible, mousewheel zooming stops working entirely. You can no longer zoom in or out using the mousewheel when hovering over any axis.

This bug is mentioned in this comment https://github.com/highcharts/highcharts/issues/19475#issuecomment-1671782103 but is a separate issue than the original report there (though perhaps related).

https://jsfiddle.net/0znyexpq/

To reproduce, simply 

1. Zoom in with the mousewheel until either the top or bottom axis has no more data visible
2. Notice that moving the mousewheel up or down no longer affects zoom on any axis. Mousewheel zoom is unexpectedly frozen/stuck.

#### Product version
11.1.0

#### Affected browser(s)
Chrome and Firefox (presumably others)

jakubjanuchta:
Fixed #19779, mouse wheel zoom was not working when zooming on empty space without any points.

Fixed #19475

**Demos to test:**
https://jsfiddle.net/BlackLabel/ku90z47h/
https://jsfiddle.net/BlackLabel/7rao209e/


ts/Extensions/MouseWheelZoom/MouseWheelZoom.ts
```diff
@@ -173,91 +213,96 @@ const zoomBy = function (
...
+        centerYArg = yAxis.toValue(mouseX);
+
+        // Swapping x and y for simplicity when chart is inverted.
+        const tmp = mouseX;
+        mouseX = mouseY;
+        mouseY = emulateRoof - tmp + yAxis.pos;
+    }
+
+    const hasZoomedX = zoomOnX(chart, zoomX, mouseX, howMuch, centerXArg),
+        hasZoomedY = zoomOnY(chart, zoomY, mouseY, howMuch, centerYArg),
```
pawellysy, line 300: If you change the `zoomOnX` and `zoomOnY` to be the same method,  you can use them like that
```suggestion
    const hasZoomedX = zoomOnDir(chart, AzoomX, mouseX, howMuch, centerXArg),
        hasZoomedY = zoomOnDir(chart, zoomY, mouseY, howMuch, centerYArg),
```

you probably need to pass the correct axis as well 🤔🤨

pawellysy (changes requested):
Good job! Some comments tho

ts/Extensions/MouseWheelZoom/MouseWheelZoom.ts
```diff
@@ -113,27 +96,84 @@ let wheelTimer: number,
...
+    howMuch: number,
+    centerXArg: number
 ): boolean {
-    const xAxis = chart.xAxis[0],
-        yAxis = chart.yAxis[0],
-        yOptions = yAxis.options,
-        type = pick(options.type, chart.zooming.type, ''),
-        zoomX = /x/.test(type),
-        zoomY = /y/.test(type);
+    const xAxis = chart.xAxis[0];
```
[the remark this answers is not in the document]
jakubjanuchta: Thanks for the suggestion, support for mouse wheel zooming on multiple axis has been added.

ts/Extensions/MouseWheelZoom/MouseWheelZoom.ts
```diff
@@ -113,27 +96,84 @@ let wheelTimer: number,
 /**
  * @private
  */
-const zoomBy = function (
+const zoomOnX = function (
```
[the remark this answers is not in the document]
jakubjanuchta: zoomOnX and zoomOnY had been replaced with zoomOnDirection

ts/Extensions/MouseWheelZoom/MouseWheelZoom.ts
```diff
@@ -68,42 +69,24 @@ const optionsToObject = (
...
+    outerStart:number,
+    outerWidth:number,
+    innerStart:number,
+    innerWidth: number
+): {
+    rangeStart: number,
+    rangeWidth: number
+} => {
+    const outerEnd = outerStart + outerWidth,
+        rangeStart = U.clamp(innerStart, outerStart, outerEnd),
```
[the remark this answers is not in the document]
jakubjanuchta: As discussed, removing this logic and instead changing the existing one.

pawellysy (changes requested):
When testing this demo -> https://jsfiddle.net/BlackLabel/m2rcg91t/3/

I stumbled upon the issue, that when zoom is enabled only on yAxis, the xAxis length "jumps around" like in the video below: 

https://github.com/highcharts/highcharts/assets/26306846/c8a55d26-c353-4469-881c-b4e193dc31a5

ts/Extensions/MouseWheelZoom/MouseWheelZoom.ts
```diff
@@ -68,196 +72,211 @@ const optionsToObject = (
...
-    let hasZoomed = false;
+    if (defined(axis.max) && defined(axis.min) &&
+        defined(axis.dataMax) && defined(axis.dataMin)) {
 
-    if (defined(xAxis.max) && defined(xAxis.min) &&
-        defined(yAxis.max) && defined(yAxis.min) &&
-        defined(xAxis.dataMax) && defined(xAxis.dataMin) &&
-        defined(yAxis.dataMax) && defined(yAxis.dataMin)) {
+        if (isXAxis) {
+            fixToDir = mousePos ? ((mousePos - axis.pos) / axis.len) : 0.5;
```
pawellysy, line 130: I'd go for splitting xAxis- and yAxis-specific functionalities out of this function. The fixToDir value gets calculated a bit differently, so I'd leave it out of the scope of this function. It will be also a bit easier to read.

pawellysy (changes requested):

jakubjanuchta:
> When testing this demo -> https://jsfiddle.net/BlackLabel/m2rcg91t/3/
> 
> I stumbled upon the issue, that when zoom is enabled only on yAxis, the xAxis length "jumps around" like in the video below:
> 
>  Screen.Recording.2023-10-16.at.08.33.42.mov

This behavior is caused by the yAxis labels changing the width of xAxis and it has been there before.

ts/Extensions/MouseWheelZoom/MouseWheelZoom.ts
```diff
@@ -66,198 +70,257 @@ const optionsToObject = (
...
+ * Beggining of bounds range.
+ * @param {Highcharts.Axis} outerWidth
+ * Width of bounds range.
+ * @param {Highcharts.Chart} innerStart
+ * Beggining of new extremes.
+ * @param {Highcharts.Axis} innerWidth
+ * Width of new extremes.
+ * @return {Object}
+ * Object containing rangeStart and rangeWidth.
  */
```
pawellysy, line 85: Not true really, because this function does not need any axis knowledge, it just fits a segment inside some range. and the params have the wrong type, those are just numbers.

ts/Extensions/MouseWheelZoom/MouseWheelZoom.ts
```diff
@@ -66,198 +70,257 @@ const optionsToObject = (
...
+        }
+    }, 400);
+};
+
+/**
+* Calulate the ratio of mouse position on the axis to it's length. If mousePos
+* doesn't exist, returns 0.5;
+* @private
+*/
+const calucateFixToDirection = function (
```
pawellysy, line 171: if this function calculates the ratio of mouse position, then I'd name the method something like `getMouseAxisRatio`, because the current name doesn't say anything.

ts/Extensions/MouseWheelZoom/MouseWheelZoom.ts
```diff
@@ -66,198 +70,257 @@ const optionsToObject = (
...
+    return fixToDir;
+};
 
-    if (defined(xAxis.max) && defined(xAxis.min) &&
-        defined(yAxis.max) && defined(yAxis.min) &&
-        defined(xAxis.dataMax) && defined(xAxis.dataMin) &&
-        defined(yAxis.dataMax) && defined(yAxis.dataMin)) {
+/**
+* Check if zoom is out of the current axis, if not, change the extremes to
+* previously calulated values based on how much zoom should be applied.
```
pawellysy, line 191: The method description should focus more on WHAT the method is doing not on HOW it is doing it. So just saying something like `Perform the zooming on the passed axis` should be enough ;)

pawellysy (changes requested):

ts/Extensions/MouseWheelZoom/MouseWheelZoom.ts
jakubjanuchta: Thanks for the suggestion!

ts/Extensions/MouseWheelZoom/MouseWheelZoom.ts
jakubjanuchta: I agree, this doesn't say anything. I changed the method name and also the fixToDir variable name.

pawellysy (approved):

ts/Extensions/MouseWheelZoom/MouseWheelZoom.ts
```diff
@@ -66,198 +70,256 @@ const optionsToObject = (
...
-    const xAxis = chart.xAxis[0],
-        yAxis = chart.yAxis[0],
-        yOptions = yAxis.options,
-        type = pick(options.type, chart.zooming.type, ''),
-        zoomX = /x/.test(type),
-        zoomY = /y/.test(type);
+    axis: Axis,
+    mousePos: number
+) : number {
+    const mouseAxisRatio = mousePos ? ((mousePos - axis.pos) / axis.len) : 0.5,
```
pawellysy, line 176: this check will return false, if `mousePos` is 0, and I think this is not a desired behaviour :D 
for the check use `U.defined` method, and if the mousePos can be undefined, let's reflect it in the parameter type either by making it optional, or by piping it with the `undefined` type.

pawellysy: Btw, it can be simplified to just 
```suggestion
   if (!defined(mousePos)) {
   return 0.5;
   }
    const mouseAxisRatio = ((mousePos - axis.pos) / axis.len);
```

pawellysy (changes requested):
Actually I noticed one more thing 😓

ts/Extensions/MouseWheelZoom/MouseWheelZoom.ts
jakubjanuchta: Good catch, thanks!

TorsteinHonsi (approved):
Thanks, nice work!

PR #19913, opened 2023-10-11
Added workflow for notifying #CI channel when docs changes

goransle:


MarkusBarstad (approved):
Alright, think I get it, looks pretty clear and concise 🚀 ! 

Just to make sure I understand; 

The `concurrency` definition here will halt running workflows for commits to master (those which alter docs), to prevent updating slack with the same ref many times in a row right?

goransle:
> The `concurrency` definition here will halt running workflows for commits to master (those which alter docs), to prevent updating slack with the same ref many times in a row right?

Yes, more or less. `${{ github.workflow }}-${{ github.ref }}` means it will group by this workflow and the branch it was triggered on. So more or less `docs-notifier-master`.

TorsteinHonsi (approved):
Thanks!

PR #19916, opened 2023-10-11
bugfix/19885-breaks-with-zooming

### LINKED ISSUE #19885: BrokenAxis: `yAxis.breaks` does not work with zooming
#### Actual behavior
The[ **demo**](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/axisbreak/break-visualized/) included in the [**API**](https://api.highcharts.com/highcharts/yAxis.breaks) does not work after adding `chart.zoomType`. 

After zooming into a part of the axis which is not broken, the `pointBreakColumn` will not be called; thus, the position of the 'point breaks' will remain the same (which is not correct).

**Demo:** https://jsfiddle.net/BlackLabel/6btj37sL/
Zoom the chart to see that the 'point breaks' are not updated.

#### Product version
Highcharts v11

#### Affected browser(s)
All.

jakubSzuminski:
Fixed #19885, `axis.breaks` works now with zooming.

---
**Tasks:**
- [ ] Add an event that will trigger after the axis has been zoomed-in and the break is no longer visible
- [ ] Add a unit test to cover this case
- [ ] Refactor the `BrokenAxis.ts`*
- [ ] Update the **[demo](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/axisbreak/break-visualized/)**
- [ ] Add the event to the API documentation

### DIFF
--- a/ts/Core/Axis/BrokenAxis.ts
+++ b/ts/Core/Axis/BrokenAxis.ts
@@ -293,11 +293,36 @@ namespace BrokenAxis {
                 threshold = axis.isXAxis ?
                     axis.min :
                     pick(series.options.threshold, axis.min);
+
+                // array of breaks that have been "zoomed-out"
+                // which means that they were shown previously, but now
+                // after zoom, they are not (#19885).
+                const breaksOut = axis?.options?.breaks?.filter(
+                    function (brk): boolean {
+                        let isOut = false;
+
+                        // iterate to see if "brk" is visible
+                        for (let i = 0; i < breaks.length; i++) {
+                            const otherBreak = breaks[i];
+                            if (
+                                otherBreak.from === brk.from &&
+                                otherBreak.to === brk.to
+                            ) {
+                                isOut = false;
+                                break;
+                            }
+                        }
+
+                        return isOut;
+                    }
+                );
+
                 points.forEach(function (point: Point): void {
                     y = pick(
                         (point as any)['stack' + key.toUpperCase()],
                         (point as any)[key]
                     );
+
                     breaks.forEach(function (brk: AxisBreakObject): void {
                         if (isNumber(threshold) && isNumber(y)) {
 
@@ -326,6 +351,12 @@ namespace BrokenAxis {
                             }
                         }
                     });
+
+                    breaksOut?.forEach(
+                        function (brk: AxisBreakOptions | undefined): void {
+                            fireEvent(axis, 'pointBreakOut', { point, brk });
+                        }
+                    );
                 });
             });
         }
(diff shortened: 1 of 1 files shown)


jakubSzuminski:
@karolkolodziej could you please check if the idea of this `pointBreakOut` event even makes sense? The problem that I was trying to solve was that there was no way to delete the "point breaks visualized" from this **[demo](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/axisbreak/break-visualized/)** if a user added the `chart.zoomType` option.

My questions are:
1. Does this event even makes sense (like in this [demo](https://jsfiddle.net/BlackLabel/ok2c5qs8/))? Or should I aim to solve this problem in a different way?
2. Should I add a demo showing how the `pointBreakOut` event works?

ts/Core/Axis/BrokenAxis.ts
```diff
@@ -334,11 +343,18 @@ namespace BrokenAxis {
...
+
                             if (eventName) {
                                 fireEvent(axis, eventName, { point, brk });
                             }
                         }
                     });
+
+                    breaksOut?.forEach(
+                        function (brk: AxisBreakOptions | undefined): void {
+                            fireEvent(axis, 'pointBreakOut', { point, brk });
```
karolkolodziej, line 355: Looks and works well but I would stick to the naming we already have and change it to something like `pointOutsideOfBreak`

karolkolodziej (approved):

ts/Core/Axis/BrokenAxis.ts
```diff
@@ -292,30 +282,49 @@ namespace BrokenAxis {
...
             keys.forEach(function (key: string): void {
-                breaks = brokenAxis && brokenAxis.breakArray || [];
+                breaks = brokenAxis?.breakArray || [];
                 threshold = axis.isXAxis ?
                     axis.min :
                     pick(series.options.threshold, axis.min);
+
+                // array of breaks that have been "zoomed-out"
+                // which means that they were shown previously, but now
+                // after zoom, they are not (#19885).
```
TorsteinHonsi, line 298: Style guide: Upper case first in comments, and fill the line out until max-len

ts/Core/Axis/BrokenAxis.ts
```diff
@@ -292,30 +282,49 @@ namespace BrokenAxis {
...
+                const breaksOut = axis?.options?.breaks?.filter(
+                    function (brk): boolean {
+                        let isOut = true;
+
+                        // iterate to see if "brk" is visible
```
TorsteinHonsi, line 303: Upper case first

TorsteinHonsi (changes requested):
Thanks! Just a couple of cosmetic change requests. And I second Karol's comment about the naming of the `breaksOut` array.

TorsteinHonsi (approved):
Thanks!

PR #19918, opened 2023-10-11
bugfix/19917-arialabel-empty-series

### LINKED ISSUE #19917: Setting options in a series without `series.data` is adding default aria-label
#### Expected behaviour
I would expect VoiceOver to read the correct aria-label for the series and not create area-label for the empty series (if series is not empty on purpose).

#### Actual behaviour
When using `dashStyle` and fetching data with `csvURL`, VoiceOver tells me that there are 5 lines, but reads out duplicates in this order:
- Series 1, Line 1 of 5 with 0 data points
- Android, Line 1 of 5 with 218 data points

https://jsfiddle.net/maritavindedal/tL0bf6yv/5/

`console.log()`:
Screenshot 2023-10-11 at 15 33 35

marvin19:
Fixed #19917, setting options in a series without `series.data` would add default aria-label.

Removed aria-label for series without data. Added a check to see if `series.data` is empty and only describe if `series.data` has value. This makes styling with e.g `dashStyle` and nothing else in the series possible without describing the object.


oysteinmoseng (changes requested):
Nice and simple, I think that's fine (until we get complaints) - and if people want the series described it needs to have null points.

Would be nice to have this covered by a unit test, though. Probably an existing one you can add to.

oysteinmoseng (approved):
🎉

TorsteinHonsi (approved):
Thanks!

PR #19920, opened 2023-10-12
Migration: Series part 2

bre1470:


ts/Series/Item/ItemSeries.ts
```diff
@@ -189,50 +123,54 @@ class ItemSeries extends PieSeries {
...
+                    opacity: 1
+                }, series.options.animation);
+            }
         }
     }
 
     public drawDataLabels(): void {
-        if (this.center && this.slots) {
-            H.seriesTypes.pie.prototype.drawDataLabels.call(this);
+        const series = this;
```
TorsteinHonsi, line 148: A small detail - I wouldn't bother making a local constant alias for the `this` context in cases like this. In fact I often change it back to `this` when I am working on code with this pattern. Historically we did this when we needed to access the top scope inside nested functions, but that doesn't apply anymore with arrow functions.

TorsteinHonsi (changes requested):
Thanks! Seems like there is a regression in the item series, ref the visual tests.

ts/Series/Item/ItemSeries.ts
TorsteinHonsi: @bre1470 
> understanding the code.
Yes, but arguably only if the function is so large that we lose track of the context. In short and simple functions I find it more tidy to use `this` directly. A matter of taste I guess.

> helps with compression
Actually not, Closure Compiler makes that decision for us. For example this block
```js
function Person(name) {
  this.name = name;
}
Person.prototype.print = function () {
  const p = this;
  console.log(p.name);
}
const t = new Person('Torstein');
```

Compiles to

```js
function Person(a){this.name=a}Person.prototype.print=function(){console.log(this.name)};var t=new Person("Torstein");
```

My point is we shouldn't spend too much time on changing this either one way or the other, as it is subjective and we don't have a fixed style guide decision on it.

TorsteinHonsi (approved):

PR #19935, opened 2023-10-16
compare/bugfix/19934-safari-csv-export

### LINKED ISSUE #19934: CSV export is broken in Safari
#### Expected behaviour
CSV export should work in Safari

#### Actual behaviour
CSV export fails with the following error:

```
[Error] Error: Failed to convert to blob
	(anonymous function) (export-data.js:19:355)
	c (export-data.js:20:475)
	onclick (export-data.js:38)
	onclick (exporting.js:32:84)
```

https://jsfiddle.net/etp6wvd3/
* Hamburger --> Download CSV

#### Product version
(At least) Highcharts 10.2 - current master

#### Affected browser(s)
Safari 17.0

goransle:
Fixed #19934, CSV export was broken in Safari.


TorsteinHonsi (approved):
Thanks!

Issue #19832, opened 2023-09-26
CSV export performance issue


KamilKubik:
#### Expected behaviour
The CSV export should be performed almost immediately.

#### Actual behaviour
The CSV export takes approximately 5-10 seconds.

#### Live demo with steps to reproduce
Stackblitz is the only editor, where we can run the newest HC version (11.1.0) and have that amount of data without lagging the whole editor. It probably needs more investigation, and maybe it's not because of the amount but a specific data set.

https://stackblitz.com/edit/js-u8a9rx?file=package.json,chart-options.js,index.js,index.html

#### Product version
Affected from: v10.3.0
Last working: v10.2.1

Bisected to: 426f0660927742380945314cba2494dca66d70b4

sergei-lobanov:
Any updates on this issue?

karolkolodziej:
Hi @sergei-lobanov! You can check the progress in the PR linked above.
We will do our best to include it in the upcoming release.

sergei-lobanov:
<<Actual behaviour
The CSV export takes approximately 5-10 seconds.

On our production env it may take 1-2 minutes to export all data from the chart to csv and before it was immediately or a few seconds.

PR #19937, opened 2023-10-16
enhancement/export-data-loading

goransle:
Adds a loading message when running data export. Also adds a `getDataRows` benchmark test, will add this to CI in another PR.

https://github.com/highcharts/highcharts/assets/7478772/1f072c45-0d02-4671-b6a0-1cf5fbc748be

[jsfiddle](https://jsfiddle.net/goransle/0gzkr24f/)

Related to https://github.com/highcharts/highcharts/issues/19832

### DIFF
--- a/test/ts-node-unit-tests/benchmarks/Highcharts/ExportData-getDataRows.bench.ts
+++ b/test/ts-node-unit-tests/benchmarks/Highcharts/ExportData-getDataRows.bench.ts
@@ -24,7 +24,7 @@ export default async function benchmarkTest(
     }: BenchmarkContext
 ): Promise<BenchmarkResult> {
 
-  const { win,  el} = setupDOM();
+  const { win,  el } = setupDOM();
 
   const hc = require(join(CODE_PATH, '/highcharts.src.js'))(win);
   require(join(CODE_PATH, '/modules/export-data.src.js'))(hc);
--- a/test/ts-node-unit-tests/benchmarks/Highcharts/ExportData-getDataRows.bench.ts
+++ b/test/ts-node-unit-tests/benchmarks/Highcharts/ExportData-getDataRows.bench.ts
@@ -1,0 +1,49 @@
+import type { BenchmarkContext, BenchmarkResult } from '../../benchmark';
+import { performance } from 'node:perf_hooks';
+import { join } from 'node:path';
+import { setupDOM } from '../../test-utils';
+import { generateColumnData } from '../../data-generators';
+
+export const config = {
+    sizes: [10_000, 100_000, 500_000, 1000_000]
+};
+
+export function before(size: number) {
+
+    return {
+        fileName: `${size}-rows.json`,
+        func: () => generateColumnData(size, 1)
+    };
+}
+
+export default async function benchmarkTest(
+    {
+        size,
+        CODE_PATH,
+        data
+    }: BenchmarkContext
+): Promise<BenchmarkResult> {
+
+  const { win,  el} = setupDOM();
+
+  const hc = require(join(CODE_PATH, '/highcharts.src.js'))(win);
+  require(join(CODE_PATH, '/modules/export-data.src.js'))(hc);
+
+
+  const chart = hc.chart(el, {
+      accessibility:  {
+        enabled: false
+      },
+      series: [{
+          data: data,
+      }]
+  });
+
+  performance.mark('Start');
+
+  chart.getDataRows();
+
+  performance.mark('End');
+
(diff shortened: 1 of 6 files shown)


ts/Extensions/ExportData/ExportDataDefaults.ts
```diff
@@ -265,7 +265,9 @@ const lang: LangOptions = {
      * @since 8.2.0
      * @requires modules/export-data
      */
-    hideData: 'Hide data table'
+    hideData: 'Hide data table',
+
+    exportInProgress: 'Exporting...'
```
hubertkozik, line 270: I think, that kind of labels (`Exporting...`) should be in `lang` object, same as e.g. [loading label](https://api.highcharts.com/highcharts/lang.loading).

ts/Extensions/ExportData/ExportDataOptions.d.ts
```diff
@@ -31,6 +31,7 @@ declare module '../../Core/Options' {
         exportData?: ExportDataLangOptions;
         viewData?: string;
         hideData?: string;
+        exportInProgress?: string;
```
hubertkozik, line 34: Possibly not needed due to `lang` object.

ts/Extensions/ExportData/ExportDataOptions.d.ts
```diff
@@ -79,6 +80,7 @@ export interface LangOptions {
     exportData?: ExportDataLangOptions;
     viewData?: string;
     hideData?: string;
+    exportInProgress?: string;
```
hubertkozik, line 83: Possibly not need due to `lang` object.

hubertkozik (changes requested):
Great idea!

Besides inline comments, what do you think about the possibility of disabling that exporting label? Maybe it would be good to add logic with a boolean variable to enable/disable it?

bre1470:
Sorry for the branch confusion.

ts/Extensions/ExportData/ExportDataDefaults.ts
goransle: I think it kind of is? I see that `hideData` ends up in the same API location and it is defined in the same place 🤔

test/ts-node-unit-tests/benchmarks/Highcharts/ExportData-getDataRows.bench.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export default async function benchmarkTest(
+    {
+        size,
+        CODE_PATH,
+        data
+    }: BenchmarkContext
+): Promise<BenchmarkResult> {
+
+  const { win,  el} = setupDOM();
```
MarkusBarstad, line 27: `{ win,  el }`* 8)

ts/Extensions/ExportData/ExportDataOptions.d.ts
goransle: It's not necessarily defined in the sense that it requires the export-data module to be loaded for the end user.

MarkusBarstad (approved):

marvin19 (approved):

ts/Extensions/ExportData/ExportDataDefaults.ts
hubertkozik: Ahh sorry! My bad!

hubertkozik (approved):

sergei-lobanov:
I've a question how this help to resolve https://github.com/highcharts/highcharts/issues/19832 ?
The issue not only that it may freeze your browser during export, but also that before last changes the process was much faster. When you doing export to csv for a lot of data - last time it took a few seconds and now 1-2 minutes.

I think need to check what is the real issue with the performance.

TorsteinHonsi (commented):
Thanks! This is a good idea for UX improvements. Instead of wrapping the function calls in the onclicks, I am thinking it would be cleaner to wrap the contents of the functions themselves. Then we don't need to change the simple format of the `onclick` handler.

TorsteinHonsi:
> I've a question how this change can help to resolve https://github.com/highcharts/highcharts/issues/19832 ?

You're right it doesn't resolve it. As per the OP's comment, it is _related_ to it. We should look into the performance of the [modified code](https://github.com/highcharts/highcharts/blob/aef9d2afc53a5d7f1fa1338bb2366807193b014e/ts/Extensions/ExportData/ExportData.ts#L528-L544) in addition.

TorsteinHonsi (approved):
Thanks!

PR #19938, opened 2023-10-16
Samples: Occasionally failing column-comparison

TorsteinHonsi:
The previous change needed `display: inline-block` to work, and `vertical-align` for pixel perfection.


goransle (approved):
🤞

PR #19941, opened 2023-10-16
Tools: test ES2020

TorsteinHonsi:
Test upgrading builds to ES2020. A continuation from #19006. 

We now use more optional chaining and spread operator in the sources. Would be nice to see the more compact code pay off.

#### To do
 - [x] Ensure the `/es5` builds get transpiled down. Use legacy Closure Compiler if necessary. Edit: TypeScript transpiles it to the correct ES version. Then swc just minifies that syntax without using newer language features.
 - [x] Refactor the [distMinify](https://github.com/highcharts/highcharts/blob/e5cc284174b55b676e48aa2c9719ec72d5725b73/tools/gulptasks/dashboards/dist-minify.js) function for Dash. Currently it is a copy of scripts-compile with only minor differences. Call scripts-compile instead. DRY.
 - [x] Test run the minified files in the Utils (by setting `"useSourceCode": false` in config.json).
 - [x] Check the remaining references to Closure Compiler in the repo.


bre1470 (approved):
All demos work for me.

bre1470:
It also seems to compile faster than CC. 🚀

karolkolodziej (approved):
Tested with Dashboards- everything works as expected 😉

PR #19945, opened 2023-10-17
Enhancement/fix benchmarks

pawellysy:
Fixed the benchmark comparison scripts.

### DIFF
--- a/.github/workflows/stock-benchmark.yml
+++ b/.github/workflows/stock-benchmark.yml
@@ -5,6 +5,8 @@ on:
       - 'master'
     paths:
       - 'ts/Stock/**'
+      - 'ts/Core/Axis/OrdinalAxis.ts'
+      - 'ts/Extensions/Annotations/**'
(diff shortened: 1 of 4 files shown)


kamil-musialowski (approved):

.github/workflows/stock-benchmark.yml
```diff
@@ -5,6 +5,8 @@ on:
       - 'master'
     paths:
       - 'ts/Stock/**'
+      - 'ts/Core/Axis/OrdinalAxis.ts'
+      - 'ts/Extensions/Annotations/**'
```
pawelfus, line 9: + `StockChart.ts` and `BrokenAxis.ts` please

pawelfus (changes requested):
While you are at this.. 😬

pawelfus:
How hard would it be to add a column with %-change? Is 100ms too long? It depends if the test takes 500ms or 5000ms

DJTechnoo (approved):

pawelfus (approved):
Thanks!

PR #19951, opened 2023-10-17
bugfix/19826-lollipop-grouping

### LINKED ISSUE #19826: Series with the same index overlap each other on a lollipop chart with the value inverted: true
#### Expected behaviour
In highcharts 10.0.0, a lollipop chart with the value inverted: true had indents between series with the same values.
For example, in series 1 there was data: [[0, 67]], in series 2 there was [[0, 55]]
But on the graph there was an indent between these two connectors

#### Actual behaviour
Now connectors overlap each other.

https://jsfiddle.net/68bycojd/19/

#### Product version
Highcharts version 11.1.0

#### Affected browser(s)
All

hubertkozik:
Fixed #19826, Lollipop grouping didn't work correctly.

### DIFF
--- a/samples/highcharts/series-lollipop/disabled-grouping/demo.css
+++ b/samples/highcharts/series-lollipop/disabled-grouping/demo.css
@@ -1,0 +1,6 @@
+#container {
+    min-width: 310px;
+    max-width: 800px;
+    height: 400px;
+    margin: 0 auto;
+}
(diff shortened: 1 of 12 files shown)


goransle (approved):
Looks good 👍

TorsteinHonsi (approved):
Thanks!

PR #19952, opened 2023-10-17
bugfix/19793-trendline-duplicate-x-values

### LINKED ISSUE #19793: Trendline functionality issue
#### Expected behaviour
Trendline is supposed to be a straight line but it's not. 

#### Actual behaviour
image

https://jsfiddle.net/4cwobzrv/1

#### Product version
11.1.0 (It was ok in the 10.3.3)

#### Affected browser(s)
Any borwser

kamil-musialowski:
Fixed #19793, trendline was not working for duplicate `x` values.

### DIFF
--- a/ts/Stock/Indicators/TrendLine/TrendLineIndicator.ts
+++ b/ts/Stock/Indicators/TrendLine/TrendLineIndicator.ts
@@ -113,7 +113,7 @@ class TrendLineIndicator extends SMAIndicator {
             LR: Array<Array<number>> = [],
             xData: Array<number> = [],
             yData: Array<number> = [],
-            index: number = params.index || 3;
+            index: number = params.index;
 
         let numerator = 0,
             denominator = 0,
--- a/ts/Stock/Indicators/TrendLine/TrendLineIndicator.ts
+++ b/ts/Stock/Indicators/TrendLine/TrendLineIndicator.ts
@@ -28,8 +28,7 @@ import U from '../../../Core/Utilities.js';
 const {
     extend,
     merge,
-    isArray,
-    pushUnique
+    isArray
 } = U;
 
 /* *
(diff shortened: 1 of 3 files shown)


kamil-musialowski:
@pawelfus I went with the approach for sum of least squares. Everything works just fine imo, however I feel like the code could be a bit cleaner/shorter. If you have any suggestions for further refactoring of that solution, I'm more than happy to implement them.

pawelfus (changes requested):
Perf

DJTechnoo (approved):
Well done!

ts/Stock/Indicators/TrendLine/TrendLineIndicator.ts
```diff
@@ -107,48 +107,55 @@ class TrendLineIndicator extends SMAIndicator {
...
-            alpha: number,
-            i: number,
-            y: number;
-
-        // Get sums:
-        for (i = 0; i < xValLength; i++) {
-            y = isArray(yVal[i]) ? yVal[i][index] : (yVal[i] as any);
-            sumY += y;
-            sumXY += i * y;
+            index: number = params.index || 3;
```
pawelfus, line 116: No need for that. If index can be undefined, then it means TS is incorrect (eg DeepPartial is not used)

ts/Stock/Indicators/TrendLine/TrendLineIndicator.ts
```diff
@@ -107,48 +107,55 @@ class TrendLineIndicator extends SMAIndicator {
...
+        let numerator = 0,
+            denominator = 0,
+            xValSum = 0,
+            yValSum = 0,
+            counter = 0;
+
+        // Create an array of consecutive xValues, (don't remove duplicates)
+        for (let i = 0; i < orgXVal.length; i++) {
+            if (i === 0 || orgXVal[i] !== orgXVal[i - 1]) {
+                counter++;
```
pawelfus, line 127: Just double checking: it means we start the array from x=1, is this expected? Previous solution started from x=0.

pawelfus (changes requested):
Check comments inline. This solution looks promising!

ts/Stock/Indicators/TrendLine/TrendLineIndicator.ts
kamil-musialowski: Yes, exactly - by default HC xData array starts from 1 (if no value is given), and that what we want to achieve here.

ts/Stock/Indicators/TrendLine/TrendLineIndicator.ts
kamil-musialowski: @pawelfus What would you say about the solution in [Added index to TrendLine interface.](https://github.com/highcharts/highcharts/pull/19952/commits/83f26f7fa6ba6cbc1a0c35bf8f10cd0b7b596832) commit?

pawelfus: Looks good!

bm64 (approved):
LGTM 👍

pawelfus (approved):
Thanks!

TorsteinHonsi (approved):

PR #19954, opened 2023-10-17
bugfix/19545-legend-symbols-linecap

### LINKED ISSUE #19545: Linecap is not working for bar, area, column and pie chart
#### Actual behaviour
There was an example in documentation:
https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/series-line/linecap/

But, linecaps are working for line chart. In legend that linecaps was changed to square and round accordingly. 
If we add 
```
chart: {
    	type: 'area'
    },
```
After that linecaps in legend became round and they size was changed.

https://jsfiddle.net/DmytroBruslik/4kfgdvxL/1/

Also, changing type for concrete series gives nothing.
Another example:
https://jsfiddle.net/DmytroBruslik/4kfgdvxL/3/

#### Product version

"highcharts": "^10.3.3",
 "highcharts-react-official": "^3.1.0",
#### Affected browser(s)
Chrome Version 115.0.5790.171 (Official Build) (64-bit)
OS: Windows 11

KamilKubik:
Fixed #19545, the series linecap wasn't updating the legend symbol.

### DIFF
--- a/samples/unit-tests/legend/legend-general/demo.js
+++ b/samples/unit-tests/legend/legend-general/demo.js
@@ -180,3 +180,31 @@ QUnit.test(
         );
     }
 );
+
+QUnit.test('Legend items symbols using linecap (#19545).', function (assert) {
+    const chart = Highcharts.chart('container', {
+        chart: {
+            type: 'area'
+        },
+        series: [{
+            data: [3, 1, 4, 5, 2]
+        }, {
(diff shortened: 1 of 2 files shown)


TorsteinHonsi:
As per my [comment in the original issue](https://github.com/highcharts/highcharts/issues/19545#issuecomment-1772415323), I think this PR should be closed as not a bug.

PR #19956, opened 2023-10-17
dash/a11y-datagrid-sync-demo

Vazonik:
Improved accessibility in the Datagrid Sync demo.

### DIFF
--- a/samples/dashboards/demo/datagrid-sync/demo.js
+++ b/samples/dashboards/demo/datagrid-sync/demo.js
@@ -139,7 +139,7 @@ Dashboards.board('container', {
                 text: 'Dragging points <em>will not update</em> the grid'
             },
             accessibility: {
-                description: `The second bar chart uses some random data to
+                description: `The second bar chart uses some example data to
                 show disabled editing of connector values by manually changing
                 the height of bars in a series; allowConnectorUpdate option set
                 to false.`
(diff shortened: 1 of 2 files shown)


karolkolodziej (approved):

samples/dashboards/demo/datagrid-sync/demo.js
```diff
@@ -24,9 +21,29 @@ const chartOptions = {
...
+        connector values by manually changing the height of the bars in the 
+        series, which is possible with allowConnectorUpdate option set to true.`
+    },
+    tooltip: {
+        stickOnContact: true
+    },
+    xAxis: {
+        accessibility: {
+            description: 'categories of food',
+            rangeDescription: ''
```
sebastianbochan, line 38: Do we need it?

Vazonik: It seems to me that without this, the description of the axis is quite ambiguous, i.e. just "categories". Do you think this is enough or is it better to leave "categories of food"? @marvin19

marvin19: I do not think we need it in this case because the main point of this chart is showing features

marvin19 (changes requested):
Looks good! I would also add unit to the tooltip, so that the user knows it is mcg and xAxis.accessibility.description can be removed.

sebastianbochan (approved):

marvin19 (approved):

PR #19959, opened 2023-10-18
Bugfix/19742 anchor pos moved twice

### LINKED ISSUE #19742: Anchor position is moved twice when having a single group in dataGrouping mode
#### Expected behaviour
With the attached demo, that has groups configured for 2-hour periods, the groups should be positioned at:
anchor 'start': 00:00
anchor 'middle': 01:00
anchor 'end': 02:00

#### Actual behaviour
With the attached demo, the groups are positioned at:
anchor 'start': 00:00
anchor 'middle': 02:00
anchor 'end': 04:00

This is because the anchorPoints function from ts/Core/Axis/Axis.ts will make the the change twice, first on the firstAnchor condition, and then on the lastAnchor one.

https://jsfiddle.net/qfaw4tsg/1/
- This is the official demo for the anchor property, with a single group and using columns for better visibility
- Each group is for a 2 hours period.

#### Product version
Highstock [v11.1.0](https://github.com/highcharts/highcharts/releases/tag/v11.1.0)

#### Affected browser(s)
*

### LINKED ISSUE #19743: dataGrouping: Anchor property is not applied to the last group when xAxis.max is out of the last group interval
#### Expected behaviour
The anchor property is not applied to the last group

#### Actual behaviour
The anchor property (or lastAnchor) should be applied to the last group

https://jsfiddle.net/nfz5cyx6/1/
Note that the last 2 series have one group less that the first one, so the 'groupedXData[groupedDataLength] >= xMax - totalRange' condition is false, and it doesn't apply the anchor property on the anchorPoints function from ts/Core/Axis/Axis.ts

#### Product version
Highstock [v11.1.0](https://github.com/highcharts/highcharts/releases/tag/v11.1.0)

pawellysy:
Fixed #19743 and #19742, `DataGrouping` anchor positions weren't working correctly if there was only 1 point. Also Tooltip didn't take into consideration the anchors.

Closes #19742

### DIFF
--- a/ts/Extensions/DataGrouping/DataGrouping.ts
+++ b/ts/Extensions/DataGrouping/DataGrouping.ts
@@ -27,10 +27,12 @@ import DataGroupingSeriesComposition from './DataGroupingSeriesComposition.js';
 import F from '../../Core/Templating.js';
 const { format } = F;
 import U from '../../Core/Utilities.js';
+import Point from '../../Core/Series/Point';
 const {
     addEvent,
     extend,
-    isNumber
+    isNumber,
+    pick
 } = U;
 
 /* *
@@ -76,10 +78,12 @@ function onTooltipHeaderFormatter(
     this: Tooltip,
     e: Event&AnyRecord
 ): void {
+
     const chart = this.chart,
         time = chart.time,
         labelConfig = e.labelConfig,
         series = labelConfig.series as Series,
+        point = labelConfig.point as Point,
         options = series.options,
         tooltipOptions = series.tooltipOptions,
         dataGroupingOptions = options.dataGrouping,
@@ -132,12 +136,22 @@ function onTooltipHeaderFormatter(
             );
         }
 
-        // now format the key
-        formattedKey = time.dateFormat(xDateFormat as any, labelConfig.key);
+        const groupStart = pick(
+                series &&
+                    series.groupMap &&
+                    series.groupMap[point.index].startX,
+                labelConfig.key
+            ),
+            groupEnd = groupStart + currentDataGrouping?.totalRange - 1;
+
+        formattedKey = time.dateFormat(
+            xDateFormat as any,
+            groupStart
+        );
         if (xDateFormatEnd) {
             formattedKey += time.dateFormat(
                 xDateFormatEnd,
-                labelConfig.key + (currentDataGrouping as any).totalRange - 1
+                groupEnd
             );
         }
 
--- a/ts/Extensions/DataGrouping/DataGrouping.ts
+++ b/ts/Extensions/DataGrouping/DataGrouping.ts
@@ -137,9 +137,7 @@ function onTooltipHeaderFormatter(
         }
 
         const groupStart = pick(
-                series &&
-                    series.groupMap &&
-                    series.groupMap[point.index].groupStart,
+                series.groupMap?.[point.index].groupStart,
                 labelConfig.key
             ),
             groupEnd = groupStart + currentDataGrouping?.totalRange - 1;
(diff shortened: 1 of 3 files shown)


DJTechnoo (approved):

pawelfus (changes requested):
This doesn't fix the issue: https://jsfiddle.net/BlackLabel/tspwxfr5/1/ (and current: https://jsfiddle.net/BlackLabel/tspwxfr5/1/ ) - both double the offset

ts/Extensions/DataGrouping/DataGroupingSeriesComposition.ts
```diff
@@ -309,150 +319,151 @@ function applyGrouping(
...
+            gapSize = 0;
 
-            // The smoothed option is deprecated, instead, there is a fallback
-            // to the new anchoring mechanism. #12455.
-            if (
-                dataGroupingOptions &&
+        // The smoothed option is deprecated, instead, there is a fallback
+        // to the new anchoring mechanism. #12455.
+        if (
+            dataGroupingOptions &&
```
pawelfus, line 398: indent

ts/Extensions/DataGrouping/DataGroupingSeriesComposition.ts
```diff
@@ -202,73 +203,82 @@ function anchorPoints(
...
-
-            groupedXData[groupedDataLength] = ({
-                middle: groupedXData[groupedDataLength] + 0.5 * totalRange,
-                end: groupedXData[groupedDataLength] + totalRange,
-                firstPoint: lastGroupStart && series.xData[lastGroupStart],
-                lastPoint: series.xData[series.xData.length - 1]
-            } as AnchorChoiceType)[lastAnchor];
+            groupedXData[groupedDataLastIndex] >= xMax - totalRange
+    ) {
+        i--;
```
pawelfus, line 254: I suggested a more descriptive name for `i` variable. It is easy to miss it and it is not obvious why you decrease the iterator here. Previously `i` was used close to its assignment so it was more readable

pawelfus (changes requested):
Nice work!

Before accepting, I would like to see the perf results

bm64 (approved):
lgtm

kamil-musialowski (approved):

ts/Extensions/DataGrouping/DataGroupingSeriesComposition.ts
```diff
@@ -202,73 +203,81 @@ function anchorPoints(
...
-    let i;
-
-    // DataGrouping x-coordinates.
-    if (dataGroupingOptions && series.xData && totalRange && series.groupMap) {
-        const groupedDataLength = groupedXData.length - 1,
-            anchor = dataGroupingOptions.anchor,
-            firstAnchor = pick(dataGroupingOptions.firstAnchor, anchor),
-            lastAnchor = pick(dataGroupingOptions.lastAnchor, anchor);
-
-        // Anchor points that are not extremes.
```
pawelfus, line 214: This comment was misplaced, it shouldn't be removed :)

pawelfus (changes requested):
Just a minor thingy, otherwise look good!

pawelfus (approved):
Thanks!

ts/Extensions/DataGrouping/DataGrouping.ts
```diff
@@ -132,12 +136,22 @@ function onTooltipHeaderFormatter(
             );
         }
 
-        // now format the key
-        formattedKey = time.dateFormat(xDateFormat as any, labelConfig.key);
+        const groupStart = pick(
+                series &&
+                    series.groupMap &&
+                    series.groupMap[point.index].groupStart,
```
TorsteinHonsi, line 142: Use optional chaining: 
`series?.groupMap?.[point.index].groupStart`

TorsteinHonsi (changes requested):
Thanks! Just a minor change request to save a few bytes of weight.

TorsteinHonsi (approved):

PR #19966, opened 2023-10-19
tools/scripts-remove-scripts-es5

goransle:
The es5 output does not seem to be required by tests or other Gulp tasks.


TorsteinHonsi:
The way I understand it @bre1470 is that we can build the es5 bundles in dist, but we don't need them on every commit. Right?

bre1470:
@TorsteinHonsi Right, good point. It is essentially for distribution, but it is also a compiling test against modern code pattern that might be incompatible with an ES5 target. So we should test it somewhere on a regular basis, like in the nightly.

TorsteinHonsi:
@bre1470 
> So we should test it somewhere on a regular basis, like in the nightly.

I think we should test in in CI. But we don't need to test it pre-commit.

goransle:
> @bre1470
> 
> > So we should test it somewhere on a regular basis, like in the nightly.
> 
> I think we should test in in CI. But we don't need to test it pre-commit.

`scripts-es5` currently runs via `gulp dist` in the [distribution job of the nightly workflow](https://github.com/highcharts/highcharts/blob/a6ec95aca680f7e48c0d90078cadc26c3bcf3fb8/.github/workflows/nightly.yml#L127). 

We could also add some tests there, maybe as simple as asserting that the `es5` folder exists to start with?

TorsteinHonsi:
> We could also add some tests there, maybe as simple as asserting that the es5 folder exists to start with?

Yes, that's fine 👍

TorsteinHonsi (approved):
Thanks!

PR #19967, opened 2023-10-19
dash/simplified-climate-demo

sebastianbochan:
- [x] - remove custom handles from the navigator
- [x] - optimize KPI options
- [x] - remove light/dark mode toggle
- [x] - leave 4 KPI
- [x] - remove click from the KPI
- [x] - 3 series on the chart
- [x] - leave only 6 markers on the map
- [x] - a11y


Vazonik (approved):
LGTM

oysteinmoseng (approved):
Very nice! I was going to suggest to keep the original so we could still test against it and use it for the 97% goal, but actually this one is still quite complex, so it should be enough as is, and we can still keep trying to simplify it.

sebastianbochan:
@oysteinmoseng 
The "full" demo is also available, but in the other directory. 

https://github.com/highcharts/highcharts/pull/19933

marvin19 (changes requested):
@sebastianbochan See Slack for comments

marvin19 (approved):

Issue #19928, opened 2023-10-15
Chart does not draw all flags when there are many flags


godfreya:
#### Expected behaviour
The chart should draw all flags in the series given to it.

#### Actual behaviour
The chart will stop drawing flags after a certain point, even if there is space for them. In this instance it only draws the first 10 in the blue series. Apparently this is due to certain flags, as deleting a few of the first 10 allows the rest of the blue series to be drawn.

#### Live demo with steps to reproduce
https://jsfiddle.net/5to7up3L/

#### Product version
Highstock latest

#### Affected browser(s)
Google Chrome

The chart should look something akin to this screenshot. As seen in the JsFiddle however, only the first 10 blue flags are drawn due to a bug. No error message is shown.

![Screenshot 2023-10-13 045251](https://github.com/highcharts/highcharts/assets/14304377/810fc785-453b-4d19-b7ee-693704c3d637)

karolkolodziej:
Hi @godfreya! 

Thank you for creating this issue! I'm going to close it because [a similar one has been reported.](https://github.com/highcharts/highcharts/issues/10073)
You can try to apply the workaround provided by Pawel.

However, I will leave a simplified demo of yours because it is a very useful real-use case of the flag series- [demo](https://jsfiddle.net/BlackLabel/nusbkxz6/).

PR #19973, opened 2023-10-20
Fixed #10073, distribution and maxDistance

### LINKED ISSUE #10073: stock chart flag rendering problem with allowOverlapX: false
#### Expected behaviour
no rendering issue

#### Actual behaviour
rendering issue when lots of flags and time period is large

https://jsfiddle.net/kdragon/0zcg7qn4/6/ 

run the above fiddle

either select "All" time range or drag the from range to the far left and notice the flags get all clumpy and don't represent the actual flags on the actual dates

if you uncomment: allowOverlapX: true, (which highsoft support recommended as a work around), it no longer has the rendering issues, but will obviously overlap, which may not be desired. they think the bug is "Highcharts distributing algorithm seems to work incorrectly"

#### Product version

highstock 6 and 7.02

#### Affected browser(s)
latest chrome: Version 71.0.3578.98 (Official Build) (64-bit)

TorsteinHonsi:
Fixed #10073, flags were clumped together on the left side of the chart when `allowOverlapX` was false and the flags were dense.

### DIFF
--- a/ts/Core/Renderer/RendererUtilities.ts
+++ b/ts/Core/Renderer/RendererUtilities.ts
@@ -53,7 +53,6 @@ namespace RendererUtilities {
     }
 
     export interface DistributedBoxObject extends BoxObject {
-        del?: boolean;
         pos?: number;
     }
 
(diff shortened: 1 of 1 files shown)


TorsteinHonsi:
Demo from #19928.

* With fixed width, v11.1.0: https://jsfiddle.net/highcharts/L8ngp9jv/
* With the fix applied: https://jsfiddle.net/highcharts/L8ngp9jv/1/

Side effect, pie chart label placement
* With v11.1.0: https://jsfiddle.net/highcharts/tcfnd4hv/
* With the fix applied: https://jsfiddle.net/highcharts/tcfnd4hv/1/

PR #19980, opened 2023-10-23
bugfix/pattern-fill-scaling

### LINKED ISSUE #19551: Incorrect map color pattern display
#### Expected behaviour
1. GeoJSON format:
- Pattern should be displayed.

2. TopoJSON format:
- Pattern shouldn't be overscaled.

#### Actual behaviour
1. GeoJSON format:
- Pattern is displayed as a normal color (no pattern visible). Looks like **regression** - works on v8.

2. TopoJSON format:
- Pattern is displayed but is overscaled.

1. GeoJSON format:
- https://jsfiddle.net/BlackLabel/ox4dwqtb/

2. TopoJSON format:
- https://jsfiddle.net/BlackLabel/e6nxutbo/

#### Product version

Highmaps JS v11.1.0

Xillez:
Fixed #19551, patterns got scaled with their assigned map areas causing inconsistent rendering.

---

  - https://app.asana.com/0/0/1205291551309235


Xillez:
@TorsteinHonsi There are some minor visual differences and I was hoping you could take a look to see if they're ok? (or maybe someone else if you don't have the time).
It comes from scaling of the built in pattern definitions to remove the patternTransform attribute, so it possibly won't interfere with the fix.

@hubertkozik Made some fiddles and it looks like patterns got inadvertently scaled up a bit somewhere since 9.0.0:
v9.0.0: https://jsfiddle.net/BlackLabel/gashvrzw/
current fix: https://jsfiddle.net/BlackLabel/ekbz4s75/

ts/Extensions/PatternFill.ts
```diff
@@ -300,23 +302,23 @@ SVGRenderer.prototype.addPattern = function (
     animation?: (boolean|Partial<AnimationOptions>)
 ): (SVGElement|undefined) {
     let pattern: (SVGElement|undefined),
-        animate = pick(animation, true),
-        animationOptions = animObject(animate),
         path: SVGAttributes,
+        attribs: SVGAttributes,
+        id = options.id;
```
KacperMadej, line 307: It is recommended to keep variables sorted in the ascending order as long as this is not causing any initialization problems (if you have something like `let x = this, a = x.a;` then it is going to stay as is).

hubertkozik (changes requested):
Let's discuss these 3 demos:
- TopoJSON from the actual branch: https://jsfiddle.net/BlackLabel/1b69aqdm/
- GeoJSON from the actual branch: https://jsfiddle.net/BlackLabel/bosm3h4a/
- GeoJSON with v9.2.2: https://jsfiddle.net/BlackLabel/bs3w7oue/

Actual TopoJSON and GeoJSON looks almost the same, minor differences in pattern position - I wouldn’t classify it as an issue. Just a small difference between topo and geo jsons.

Actual TopoJSON and actual GeoJSON are really similar to map with GeoJSON from v9.2.2 (the last working version before regression). To me, this fix looks good now. It applies only for Highcharts Maps and now we don't have any differences for normal series.

- [x] The last missing thing in the code is a unit-test. We've got tests for pattern fill in HC Maps, so I think checking the scale of the pattern in map series would be enough.

- [ ] Also, some notes about the description itself. See [Writing content for the changelog](https://github.com/highcharts/highcharts/tree/master/changelog#writing-content-for-the-changelog). We should follow the rules about assigning yourself to PR, correct labelling and linking the main issue to PR etc.

KacperMadej (changes requested):
- [ ] change the PR to draft until it passes the internal review
- [ ] add test(s)
- [ ] the [pattern-fill-map](https://www.highcharts.com/demo/maps/pattern-fill-map) seems to be broken [now](http://utils.highcharts.local/samples/#view/maps/demo/pattern-fill-map). Maybe just that demo code needs to be adjusted - it should look like before the fix or better. Maybe the fix should be excluded from being applied to patterns that use images.
image

KacperMadej (changes requested):
[The tests](http://utils.highcharts.local/samples/#test/unit-tests/color/pattern-fill) are not passing in a browser or this is happening only for me? I even merged the master branch locally to this branch to double-check (because there were 16 new commits there) but it still shows as 55/61 (or 58/61 sometimes).

Xillez:
> [The tests](http://utils.highcharts.local/samples/#test/unit-tests/color/pattern-fill) are not passing in a browser or this is happening only for me? I even merged the master branch locally to this branch to double-check (because there were 16 new commits there) but it still shows as 55/61 (or 58/61 sometimes).

Yeah I'm not sure. In firefox locally some tests are failing (none fail with chrome) but it's not the same tests that fail in the CI, and for the most part all the tests that do fail seems to be unrelated with pattern-fill as far as I can see. So not really sure what's going on. Should I try to rerun the tests to see if it's just a coincidence or something happened to not work at the time? @KacperMadej

hubertkozik (approved):

ts/Extensions/PatternFill.ts
```diff
@@ -794,6 +801,116 @@ function wrapSeriesGetColor(
...
+        // Filter for points which have patterns that don't use images assigned
+        // and has a group scale available.
+        series.points.filter(function (p: Point): boolean {
+            return (
+                p.graphic?.element.hasAttribute('fill') ||
+                p.graphic?.element.hasAttribute('color') ||
+                p.graphic?.element.hasAttribute('stroke')
+            ) &&
+            !(p?.options.color as any)?.pattern?.image &&
+            (p as any).group?.scaleX &&
```
TorsteinHonsi, line 872: Can we avoid the `any` cast here by casting to the `MapPoint` type`
`(p as MapPoint).group?.scaleX`

TorsteinHonsi (changes requested):
Thanks, this looks really solid!

**Zooming**. See this demo: https://jsfiddle.net/highcharts/fq0w7teL/ . 
* Click the plus button. Patterns don't scale.
* Then click the minus button. Patterns scale at the beginning of the animation, but are not corrected during or after the animation.
* Then click the minus button again. Patterns are now correctly scaled.

**Demo**. I realised the only demo we have for patterns on maps is the one with flags. Can we take this demo https://jsfiddle.net/highcharts/fq0w7teL/ and add it to `maps/accessibility/pattern-fills` ? It will also help us catch regressions in the future.

ts/Extensions/PatternFill.ts
```diff
@@ -24,7 +24,7 @@ import type BBoxObject from '../Core/Renderer/BBoxObject';
 import type Chart from '../Core/Chart/Chart.js';
 import type ColorString from '../Core/Color/ColorString';
 import type Point from '../Core/Series/Point.js';
-import type Series from '../Core/Series/Series.js';
+import Series from '../Core/Series/Series.js';
```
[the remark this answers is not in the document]
Xillez: I did get an error related to it but not anymore 🤔

ts/Extensions/PatternFill.ts
```diff
@@ -794,6 +801,116 @@ function wrapSeriesGetColor(
...
+ * Pattern element to scale.
+ *
+ * @param {number} scaleX
+ * X scale to apply on the element.
+ *
+ * @param {number} scaleY
+ * Y scale to apply on the element.
+ */
+function setPatternScale(
+    element: SVGDOMElement,
```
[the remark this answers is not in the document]
TorsteinHonsi: Yes, I think that's fine! A few extra bytes of code in the core, but just a handful.

Xillez:
Managed to find a fix for the zoom animation thing but it sort of looks like the pattern shifts inside the shape:
https://github.com/highcharts/highcharts/assets/5503724/0bb58529-db89-46fe-a901-d8d19accbf8e

Is this a wanted result or should I try to get rid of the "sliding" of the pattern if possible?

TorsteinHonsi:
> Is this a wanted result or should I try to get rid of the "sliding" of the pattern if possible?

I fear this might cause some seizures out there 😬 . With the current solution, are you updating the pattern on every `step` of the animation? To avoid the flickering, maybe it would be better to only apply the transformation on animation `complete`?

samples/unit-tests/color/pattern-fill/demo.js
```diff
@@ -708,4 +708,143 @@ QUnit.test('Destroy and recreate chart', function (assert) {
...
+        'custom pattern; patternTransform attribute should contain scale'
+    );
+    assert.ok(
+        defaultPattern.hasAttribute('patternTransform'),
+        'custom pattern element doesn\'t have patternTransform attribute'
+    );
+    assert.ok(
+        defaultPattern.getAttribute('patternTransform').includes('scale'),
+        'default pattern; patternTransform attribute should contain scale'
+    );
```
TorsteinHonsi, line 737: We have some unwritten conventions for assertion messages:
* Upper case first (real sentence)
* Always include the word "should" in some form. It ensures that we describe what whe _expect_ to see, not what we are actually looking at, which will be confusing depending on whether the test fails or passes, or if the assertion is `ok` or `notOk`.

TorsteinHonsi (changes requested):
Awesome, works like a charm! Just some style guide related comments now and it's ready to merge.

TorsteinHonsi (approved):

PR #19984, opened 2023-10-23
tools/post-release-gh-action

goransle:
Partially tested here: https://github.com/goransle/forkcharts/actions/runs/6623715761


TorsteinHonsi (approved):
Thanks!

PR #19986, opened 2023-10-23
dash/datagrid-a11y-quickfix

Vazonik:
Added accessibility quick fix for data grid.

### DIFF
--- a/ts/DataGrid/DataGrid.ts
+++ b/ts/DataGrid/DataGrid.ts
@@ -269,7 +269,6 @@ class DataGrid {
 
         this.gridContainer.style.height = this.getDataGridSize() + 'px';
         this.gridContainer.role = 'figure';
-        this.gridContainer.tabIndex = 0;
 
         // Init data table
         this.dataTable = this.initDataTable();
--- a/ts/DataGrid/DataGrid.ts
+++ b/ts/DataGrid/DataGrid.ts
@@ -268,6 +268,8 @@ class DataGrid {
         this.options = merge(DataGrid.defaultOptions, options);
 
         this.gridContainer.style.height = this.getDataGridSize() + 'px';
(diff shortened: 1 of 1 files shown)


sebastianbochan (approved):

ts/DataGrid/DataGrid.ts
```diff
@@ -268,6 +268,8 @@ class DataGrid {
         this.options = merge(DataGrid.defaultOptions, options);
 
         this.gridContainer.style.height = this.getDataGridSize() + 'px';
+        this.gridContainer.role = 'figure';
+        this.gridContainer.tabIndex = 0;
```
sebastianbochan, line 272: We should avoid tabIndex AFAIK, @marvin19 could you confirm?

ts/DataGrid/DataGrid.ts
```diff
@@ -518,6 +520,12 @@ class DataGrid {
         }
 
         this.updateGridElements();
+
+        this.gridContainer.ariaLabel = `Grid with ${
```
karolkolodziej, line 524: Shouldn't it be configurable?

karolkolodziej (changes requested):

ts/DataGrid/DataGrid.ts
sebastianbochan: I think that the whole a11y support should be discussed during the DataGrid "new vision". Right now, it is a temporary solution.

karolkolodziej (approved):

ts/DataGrid/DataGrid.ts
marvin19: Can be removed

marvin19 (changes requested):

marvin19 (approved):

PR #19987, opened 2023-10-23
Tools: Product-specific cypress tests

bre1470:


bre1470:
Master is now failing because of old folder structure in `test/cypress`. So no comparisons this time, but at least downgraded utils works for the tests. It is a hot fix and we will have to work on a substantial fix to make the latest highcharts-utils run in CI again.

goransle (changes requested):
The `dashboards/cypress` gulp task needs an update (or an 🪓).

~~Also, I get why it the comparison-workflows fails when working towards master, but why does the `lighthouse_comparison` job give this error when it is on the merge-commit? 🤔~~

bre1470:
Because lighthouse needs the reference numbers.

goransle:
> Because lighthouse needs the reference numbers.

Doh, it looked like the no-specs found error to me 🙈

karolkolodziej (changes requested):
Please also change the `tools/gulptasks/dashboards/cypress.js` config file as Goran suggested.

karolkolodziej (changes requested):
Still valid 🙈

goransle (approved):
Approved, pending Karol's request 😬

cypress.config.mjs
```diff
@@ -1,10 +1,31 @@
 import { defineConfig } from 'cypress';
+import { getProducts } from './tools/gulptasks/lib/test.js';
+
+const products = new Set();
+
+for (const product of getProducts()) {
+    switch (product) {
+        case 'Dashboards':
```
karolkolodziej, line 8: I guess we don't need to distinguish here between the products because later on you overwrite the `specPattern` in the `test/cypress/dashboards/config.mjs` file.

karolkolodziej: you are right! you can ignore it 😉

karolkolodziej (approved):

.github/workflows/dashboards-lighthouse-test.yml
```diff
@@ -49,7 +49,7 @@ jobs:
         run: npx gulp dashboards/scripts
 
       - name: Install utils
-        run: npm i github:highcharts/highcharts-utils
+        run: npm i github:highcharts/highcharts-utils#b46e7588e3aaa004d6148f4d85a9a419949bcd7c
```
goransle, line 52: This can be rolled back now

TorsteinHonsi (approved):
Thanks!

TorsteinHonsi (changes requested):
Two change requests
- Fix @goransle 's comment
- Resolve conflict

TorsteinHonsi (approved):

PR #19988, opened 2023-10-24
bugfix/19892-rangeSelector-with-heikinashi

### LINKED ISSUE #19892: StockTools rangeSelector after changing series throws an error
#### Expected behaviour
Range selector should work normally after changing the series type.

#### Actual behaviour
After changing the series to i.e Heikin-Aski and then clicking 1m/3m on rangeSelector chart is gone and error is shown in the console.

#### Product version
Stock 11.1.0

_Internal note: Check other series types after the fix._

kamil-musialowski:
Fixed #19892, HeikinAshi and Hollow Candlestick were not working with certain `dataGrouping` periods.

- [x] Add tests for both HeikinAshi and HollowCandlestick

### DIFF
--- a/ts/Series/HollowCandlestick/HollowCandlestickSeries.ts
+++ b/ts/Series/HollowCandlestick/HollowCandlestickSeries.ts
@@ -169,31 +169,26 @@ class HollowCandlestickSeries extends CandlestickSeries {
      */
     public getPriceMovement(): void {
         const series = this,
-            // procesed and grouped data
+            // Processed and grouped data
             processedYData = series.allGroupedData || series.yData,
             hollowCandlestickData = this.hollowCandlestickData;
 
-        if (
-            !hollowCandlestickData.length &&
-            processedYData &&
-            processedYData.length
-        ) {
+        hollowCandlestickData.length = 0;
 
-            // First point is allways bullish (transparent).
-            hollowCandlestickData.push({
-                isBullish: true,
-                trendDirection: 'up'
-            });
+        // First point is always bullish (transparent).
+        hollowCandlestickData.push({
+            isBullish: true,
+            trendDirection: 'up'
+        });
 
-            for (let i = 1; i < processedYData.length; i++) {
-                const dataPoint: any = processedYData[i],
-                    previousDataPoint: any = processedYData[i - 1];
+        for (let i = 1; i < processedYData.length; i++) {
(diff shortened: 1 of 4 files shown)


ts/Series/HollowCandlestick/HollowCandlestickSeries.ts
```diff
@@ -169,31 +169,26 @@ class HollowCandlestickSeries extends CandlestickSeries {
...
-            for (let i = 1; i < processedYData.length; i++) {
-                const dataPoint: any = processedYData[i],
-                    previousDataPoint: any = processedYData[i - 1];
-
-                hollowCandlestickData.push(series.isBullish(
-                    dataPoint,
-                    previousDataPoint
-                ));
-            }
+        hollowCandlestickData.length = 0;
```
pawelfus, line 176: Why do we need to clear this always? I think we have cleared this in `updatedData` event 🤔 If not, perhaps we should add `updated` event too? Updating this on each chart's redraw seems a bit too aggressive.

pawelfus (changes requested):

ts/Series/HollowCandlestick/HollowCandlestickSeries.ts
kamil-musialowski: Previously, it didn't work at all. Similar scenario to heikinashi, `hollowCandlestickData` had a wrong length.
Not working on master: https://jsfiddle.net/BlackLabel/6um2gctx/

pawelfus: Yes, I know, I know. Just wanted to clarify if we need to clear that array length on each redraw. It seems heikinashi has the same problem. We clear and build the dataset on each redraw (aaand we have event on updated data which is redundant now, isn't it?). Let's keep this solution for now, however adding this to the Performance profile

kamil-musialowski: Sounds fair! I'll add the test, and we are good to go. We can discuss it later F2F, since it's a good case to work on perf optimization.

DJTechnoo (approved):
Good stuff!

bm64 (approved):

pawelfus (approved):
Thanks!

TorsteinHonsi (approved):

PR #19990, opened 2023-10-24
Tools: Product-specific visual comparison

bre1470:


goransle (approved):
Looks good!

TorsteinHonsi (changes requested):
Okay, let's try this 🤞 . But first, there is a conflict to resolve.

goransle (commented):
Looks like the Cypress workflow needs the fetch-depth setting as well. Also the dashboards/script is gone from master section of the filesize job?

bre1470:
I moved file size comparison into its own workflow, so that we can keep both, Highcharts and Dashboards, in that regard together while visuals are separate.

tools/gulptasks/lib/test.js
```diff
@@ -17,79 +17,65 @@
...
+    const ChildProcess = require('node:child_process');
+
+    const paths = Array.from(new Set([
+            ...ChildProcess
+                .execSync('git diff --cached --name-only --diff-filter=ACM')
+                .toString()
+                .split('\n')
+                .filter(match => !!match),
+            ...ChildProcess
+                .execSync('git diff refs/heads/master HEAD --name-only --diff-filter=ACM')
```
goransle, line 36: Still see some git weirdness here.
image

Locally this worked for me on a single-branch checkout:
```
➜  shallowcharts git:(tools/product-visual-ci) git remote set-branches origin 'master'
➜  shallowcharts git:(tools/product-visual-ci) git fetch -v --depth=1
➜  shallowcharts git:(tools/product-visual-ci) git diff HEAD origin/master
```

PR #19995, opened 2023-10-25
bugix/19946-organization-series-node-alignment

### LINKED ISSUE #19946: Organization chart alignment node with specified node width
#### Expected behaviour
After setting a specific width for node, that node should be centred.

#### Actual behaviour
After setting a specific width for node, that node is not in the center, it is moved to the side.

https://jsfiddle.net/BlackLabel/d87kzcfr/

#### Product version
Highcharts v11.1.0

#### Workaround
You can use the custom IIFE plugin from the demo below.
https://jsfiddle.net/BlackLabel/vrmdf3bx/

hubertkozik:
Fixed #19946, nodes in the organization chart weren’t centered while using node width or height options.

---
  - https://app.asana.com/0/0/1205737378103696

### DIFF
--- a/samples/unit-tests/series-organization/organization/demo.js
+++ b/samples/unit-tests/series-organization/organization/demo.js
@@ -1,5 +1,5 @@
 QUnit.test('Organization data', assert => {
-    let chart = Highcharts.chart('container', {
+    const chart = Highcharts.chart('container', {
         series: [
             {
                 type: 'organization',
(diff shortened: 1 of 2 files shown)


hubertkozik:
There are some visual differences, but to me, it is intended differences, due to improved positioning of nodes in Organization charts.

marvin19 (changes requested):
In some of the visuals, the context is lost because the node is then overlapping the link between them. See for example [this demo](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/series-organization/datalabels-nodeformatter) and the nodes linked to CTO.

marvin19 (approved):

TorsteinHonsi (approved):
Thanks!

PR #19996, opened 2023-10-25
tools/changelog-generator-treejson-error

hubertkozik:
Added error message if tree.json doesn't exist during changelog generation.

---
  - https://app.asana.com/0/0/1205746693059627

### DIFF
--- a/changelog/generate.js
+++ b/changelog/generate.js
@@ -1,5 +1,6 @@
 /* eslint-env node, es6 */
-/* eslint func-style: 0, valid-jsdoc: 0, no-console: 0, require-jsdoc: 0 */
+/* eslint func-style: 0, valid-jsdoc: 0, no-console: 0, require-jsdoc: 0
+consistent-return: 0 */
 
 /**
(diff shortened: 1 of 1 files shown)


changelog/generate.js
```diff
@@ -43,9 +44,18 @@ const getFile = url => new Promise((resolve, reject) => {
...
     var fs = require('fs'),
-        path = require('path'),
+        path = require('path');
+
+    var tree;
+
+    try {
         // eslint-disable-next-line node/no-missing-require
         tree = require('../tree.json');
+    } catch {
```
goransle, line 54: I guess there _could_ be some other error with the require, so maybe we should add a check in the catch block for the right error type/message?

Alternatively, checking with `fs.existsSync` before the require could also do the trick.

hubertkozik: Changed to `fs.existsSync` :D

goransle (approved):
Looks good!

TorsteinHonsi (approved):
Thanks!

PR #20013, opened 2023-10-26
bugfix/19832-CSV-export-performance

### LINKED ISSUE #19832: CSV export performance issue

bm64:
Fixed #19832, improved CSV export performance.

### DIFF
--- a/ts/Extensions/ExportData/ExportData.ts
+++ b/ts/Extensions/ExportData/ExportData.ts
@@ -259,7 +259,7 @@ function chartGetCSV(
         while (j--) {
             val = row[j];
             if (typeof val === 'string') {
-                val = '"' + val + '"';
+                val = `"${val}"`;
             }
             if (typeof val === 'number') {
                 if (decimalPoint !== '.') {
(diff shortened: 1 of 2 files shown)


TorsteinHonsi:
Thank you @bm64 !
Can you throw in some comments too? This code is quite hard to grasp, and actually some explanatory comments were removed in the PR 😬 

Also, I am curious what your benchmark show? How much of a speed improvement is this?

ts/Extensions/ExportData/ExportData.ts
```diff
@@ -562,26 +544,41 @@ function chartGetDataRows(
...
+                        arr[i] = 0;
+                    }
+
+                    rows[key].pointers = arr;
+                    rows[key].pointers[series.index] = 1;
+                } else {
+                    const modifiedKey = `${key},${rows[key].pointers[series.index]}`,
+                        originalKey = key;
+
+                    if (rows[key].pointers[series.index]) {
```
pawelfus, line 561: It took me a minute to get it. It would be nice to have a comment explaining we check here if it's a duplicate.

ts/Extensions/ExportData/ExportData.ts
```diff
@@ -562,26 +544,41 @@ function chartGetDataRows(
...
                     rows[key] = [];
-                    // Contain the X values from one or more X axes
                     rows[key].xValues = [];
+
+                    const arr = [];
+                    for (let i = 0; i < series.chart.series.length; i++) {
+                        arr[i] = 0;
+                    }
+
+                    rows[key].pointers = arr;
```
pawelfus, line 555: I would add a comment here explaining briefly the solution with pointers

pawelfus (approved):
Looks good, I don't see any logic errors, good job 👍 

PS: I would add some comments though, so it's easier to maintain in the future

bm64:
@TorsteinHonsi comments added. When it comes to the benchmark, I created a demo with challenging data set for this specific issue, and the export times went from few seconds (~10s in my case) to instant.
Before: https://stackblitz.com/edit/web-platform-thrrjk?file=index.html
After: https://stackblitz.com/edit/web-platform-zju7vz?file=index.html

TorsteinHonsi (approved):

PR #20023, opened 2023-10-28
Theming preparations

TorsteinHonsi:
Added [global.buttonTheme](https://api.highcharts.com/highcharts/global.buttonTheme) option to allow general button themes.

This PR also improves the Palette Helper demo.

### To do
 - [x] Pull out `xAxis`, `yAxis` and `colorAxis` defaults so they can be reachable without `HighchartsModuleLoaded` hack.
 - [x] Check out why marker default options are duplicated for all series types. All series default options are merged by default line series options, then merged again in `setOptions`. Seems like redundant merging.


KacperMadej (changes requested):
- paletter-helper demo:
image

Dragging on the color picker is very sluggish. Maybe there could be some timeout before running the code for a smoother UX?
It looks better now, so it is a shame it is not linked in any docs (link available only in a comment of .css files).

- [ ] map navigation buttons get bigger on hover and stay bigger
image

- [ ] global/buttontheme demo throws an error on the button click

samples/highcharts/css/palette-helper/demo.js
```diff
@@ -170,6 +182,50 @@ const chartPreview = async theme => {
...
+            'pale'
+        ][Math.floor(Math.random() * 5.999)],
+        distance = Math.random();
+
+    console.log(`
+        from_hue: ${fromHue},
+        scheme: ${scheme},
+        variation: ${variation},
+        distance: ${distance}
+    `);
```
KacperMadej, line 211: Removing not needed `console.log` makes `fromHue` not used.

PR #20036, opened 2023-10-31
bugfix/19676-drillup-setview

### LINKED ISSUE #19676: mapView.setView called by drillUp method doesn't work when chart.width is set
#### Expected behaviour
Map view should fit to natural bounds after setting zoom to 1
https://jsfiddle.net/BlackLabel/sntrzbg1/

FYI @hubertkozik - It probably has something to do with drilldown animation

#### Product version
Maps 11.1.0

hubertkozik:
Fixed #19676, after drilling up in certain cases the map zoom was inconsistent compared with the initial zoom.
_______
I have refactored unit tests for maps drilldown because previous tests were created based on many `setTimeout`s. The readability of that tests was really low and It was hard to add another test in that place. I think, it is better to have few QUnits instances instead of many `setTimeout` with `chart.update`s in that case.

---
  - https://app.asana.com/0/0/1205415346697462


KacperMadej (approved):

TorsteinHonsi (approved):
Thanks!

PR #20039, opened 2023-10-31
Fix: improper place to define variable causing errors in some edge cases

KubaGorka:
This pull request introduces a small code change, moving the declaration of the legendItem variable to where it is needed, preventing an error in some edge case when itemToHighlight might be undefined. This modification has no functional impact on the code's behavior, other than fixing this small bug.

### DIFF
--- a/ts/Accessibility/Components/LegendComponent.ts
+++ b/ts/Accessibility/Components/LegendComponent.ts
@@ -682,10 +682,10 @@ namespace LegendComponent {
         const items = this.legend.allItems;
         const oldIx = this.accessibility &&
                 this.accessibility.components.legend.highlightedLegendItemIx;
-        const itemToHighlight = items[ix],
-            legendItem = itemToHighlight.legendItem || {};
-
+        const itemToHighlight = items[ix];
+        
         if (itemToHighlight) {
+            const legendItem = itemToHighlight.legendItem || {};
             if (isNumber(oldIx) && items[oldIx]) {
                 setLegendItemHoverState(false, items[oldIx]);
             }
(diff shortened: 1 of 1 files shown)


pawellysy (commented):
Hello, thank you for your contribution. Could you provide such a demo, where the issue is reproduced, and there is some error in the code?

ts/Accessibility/Components/LegendComponent.ts
```diff
@@ -682,10 +682,9 @@ namespace LegendComponent {
         const items = this.legend.allItems;
         const oldIx = this.accessibility &&
                 this.accessibility.components.legend.highlightedLegendItemIx;
-        const itemToHighlight = items[ix],
-            legendItem = itemToHighlight.legendItem || {};
-
-        if (itemToHighlight) {
+        const itemToHighlight = items[ix], legendItem = itemToHighlight?.legendItem || {};;
+        
+        if (itemToHighlight && legendItem) {
```
pawellysy, line 687: If you are setting the `legendItem` to be equal to `{}`, then this value will always be true, so this check is redundant.

KubaGorka:
Hello @pawellysy,

Unfortunatelly I don't have a small demo to provide you, because I encountered this issue while working on a bigger project. What I can do is describe when I started seeing this issue. 

For reference, I'm using [highcharts-react](https://github.com/highcharts/highcharts-react) and highcharst v11.1.0.

The error started happening when I added a support for multiple axes. It is not consistent (I would say it happens about 10% of the time), but sometimes when an action to show/hide series is triggered by clicking on it in the legend, the error pops out. It does not break the behaviour of the chart completely, because series is shown/hidden correctly, but if this action would cause the labels of axis to show/hide, this part is not updated.

ts/Accessibility/Components/LegendComponent.ts
KubaGorka: I removed the assignment to `{}`. If you think it would be better to keep it, and remove the check in line 687 let me know.

KubaGorka:
I see that this exact change was fixed in scope of a different PR: https://github.com/highcharts/highcharts/pull/19640
Closing this one

PR #20041, opened 2023-10-31
bugfix/19908-zoom-reset-export-menu-overlap

### LINKED ISSUE #19908: Issues related to zoom Reset button, export icon, and title
#### Expected behaviour
- The zoom button obeys `align` property 
- The zoom button and export icon stack appropriately, even without a title.

#### Actual behaviour
- Zoom reset button ignores any text in `align` property.
- zoom reset button shifts over the export icon when there is no title

https://jsfiddle.net/kdavidenkovmware/5zygms3h/6/
Steps to Reproduce: 
- Demo should already be set up to display error. Just zoom in.

#### Product version
Highcharts 10.3.3 and 11.1.0

#### Affected browser(s)
Chrome Version 117.0.5938.92 (Official Build) (x86_64)

madepiet:
Fixed #19908, the zoom reset button overlapped with the export menu icon when no title was present.


TorsteinHonsi (commented):
Thanks!

First, if we want to do this we must move the logic to the exporting module so that we don't pollute the core with non-core concerns.

Second, I am not sure if we should do this at all. The OP case is a non-default edge case with no title. It can easily be prevented by aligning the zoom button to the left instead. If we start looking at edge cases with non-default settings, there are many ways to get an unintended overlap:

* Export button and zoom button aligned left: https://jsfiddle.net/highcharts/5satx49e/
* Zoom button vertically aligned to bottom: https://jsfiddle.net/highcharts/5satx49e/1/
* Floating title, centered zoom button: https://jsfiddle.net/highcharts/5satx49e/2/

My point is, once we start changing the defaults, there are all kinds of ways we can make stuff overlap. Maybe it is better to describe this in the docs for the zoom button and/or the export button intead?

@KacperMadej What do you think?

madepiet:
> Thanks!
> 
> First, if we want to do this we must move the logic to the exporting module so that we don't pollute the core with non-core concerns.
> 
> Second, I am not sure if we should do this at all. The OP case is a non-default edge case with no title. It can easily be prevented by aligning the zoom button to the left instead. If we start looking at edge cases with non-default settings, there are many ways to get an unintended overlap:
> 
> * Export button and zoom button aligned left: https://jsfiddle.net/highcharts/5satx49e/
> * Zoom button vertically aligned to bottom: https://jsfiddle.net/highcharts/5satx49e/1/
> * Floating title, centered zoom button: https://jsfiddle.net/highcharts/5satx49e/2/
> 
> My point is, once we start changing the defaults, there are all kinds of ways we can make stuff overlap. Maybe it is better to describe this in the docs for the zoom button and/or the export button intead?
> 
> @KacperMadej What do you think?

Thanks @TorsteinHonsi! I agree, I also think it would be better to describe it, especially since we have good options that allow us to position the button freely according to a specific configuration.

TorsteinHonsi (approved):

Issue #3802, opened 2015-02-11
setOptions - yAxis.opposite


pawelfus:
Example: http://jsfiddle.net/6rj2s4xh/

In Highstock, `opposite` option doesn't change placement for main yAxis - however, we can see applied theme in the navigator.

sebastianbochan:
+1

PR #20044, opened 2023-11-01
Refactored Axis defaults to use global defaultOptions

TorsteinHonsi:
Refactored and simplified handling of `Axis` default options.

An attempt at cleaning up how default options, axis type specific options and user options are handled by the axes. The `xAxis`, `yAxis` and `colorAxis` defaults are now set directly on `Highcharts.defaultOptions`.

#### Todo
 - [x] Correct typing so that `defaultOptions.xAxis`, `yAxis` etc. is a full `AxisOptions` object (not deep partial and not an array).


ts/Core/Axis/Axis3DComposition.ts
```diff
@@ -447,7 +449,7 @@ class Axis3DAdditions {
         Tick3D.compose(TickClass);
 
         if (U.pushUnique(composedMembers, AxisClass)) {
-            merge(true, AxisClass.defaultOptions, Axis3DDefaults);
+            merge(true, defaultOptions.xAxis, Axis3DDefaults);
```
bre1470, line 452: This might be replaceable with `D.setOptions`.

ts/Core/Axis/ZAxis.ts
```diff
@@ -29,6 +33,11 @@ const {
     splat
 } = U;
 
+defaultOptions.zAxis = merge(xAxis, {
```
bre1470, line 36: This should go into a `compose` or otherwise default options change uncontrolled during imports.

ts/Core/Chart/GanttChart.ts
```diff
@@ -147,27 +146,29 @@ class GanttChart extends Chart {
...
         });
 
-        // apply Y axis options to both single and multi y axes
+        // Apply Y axis options to both single and multi y axes
         options.yAxis = (splat(userOptions.yAxis || {})).map((
             yAxisOptions: YAxisOptions
         ): YAxisOptions => merge(
-            defaultOptions.yAxis, // #3802
-            { // defaults
+            // Defaults
```
bre1470, line 170: Without `defaultOptions.yAxis` this now produces a regression on #3802:
Example https://jsfiddle.net/6rj2s4xh/ where second and third chart should have axis to the left, in both chart and navigator.

ts/Core/Chart/GanttChart.ts
```diff
@@ -147,27 +146,29 @@ class GanttChart extends Chart {
                 defaultLinkedTo = 0;
             }
             return merge(
-                defaultOptions.xAxis,
-                { // defaults
+                // Defaults
```
bre1470, line 149: Not sure, if this should have defaultOptions.xAxis restored as well, for similar cases like #3802.

ts/Core/Axis/AxisDefaults.ts
```diff
@@ -3157,47 +3149,6 @@ namespace AxisDefaults {
      * @apioption zAxis
      */
```
bre1470, line 3151: This is now followed by a closing curly bracket and so is removed by TypeScript before JSDoc and DTS generator can pick it up. You can move it outside to the root level followed by `''; // keeps doclets above in JS file` / or to the top inside the scope.

bre1470 (changes requested):
I get now one unit test failure in Firefox (v119):
```
Expected space for ticks, calculation of chart plot height (#19896). FAILED
        Plot height with {"xAxis":{"labels":{"style":{"fontSize":11}}}} should
        Debug: http://utils.highcharts.local/samples/#test/unit-tests/axis/ticks
                        stay the same after updating series (#19604).
        Expected: 236.5
        Actual: 237
```
... and some changes need attention; see comments.

ts/Core/Chart/GanttChart.ts
TorsteinHonsi: Ok. This probably requires that we change the (actual) default value of `opposite` to false, so that we can check against `defaultOptions` here: 

https://github.com/highcharts/highcharts/blob/b74c01fe0b1df5fe6bb4e424ffbb415db9a614aa/ts/Core/Chart/StockChart.ts#L135

ts/Core/Chart/GanttChart.ts
TorsteinHonsi: Strictly speaking, yes. Nice catch! But on the contrary to Stock charts, this one never worked: https://jsfiddle.net/0L3fvyhb/ . I'll see if we can resolve it the same way as for Stock.

ts/Core/Axis/Axis3DComposition.ts
TorsteinHonsi: I think this one is simpler because it goes directly to the xAxis sub-object.

PR #20047, opened 2023-11-01
Tools/ci enhancements

goransle:


.github/workflows/highcharts-headless.yml
```diff
@@ -17,12 +17,15 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
         with:
-          node-version: lts/*
+          node-version: 18
```
bre1470, line 20: Why not the latest LTS (v20)?

bre1470: I see now the `ts-node-unit-tests` issue. 👍

.circleci/config.yml
```diff
@@ -11,3 +11,13 @@ jobs:
...
+  security_scan_cron:
+    triggers:
+      - schedule:
+          cron: "0 0 31 2 0" # February 31st
+          filters:
+            branches:
+              only:
+                - chore/scan-security
+    jobs:
+      - security_scan
```
bre1470, line 23: The setup for security scans needs to be created via repository settings.

bre1470 (approved):
Looks good. 👍

.circleci/config.yml
goransle: To be clear: this bit is the workaround for the CircleCI "no workflows" error. (The scheduled date + the branch filter should ensure it the workflow never runs)

TorsteinHonsi (approved):
Thanks!

PR #20050, opened 2023-11-01
bugfix/20048-newdataannouncer-not-announcing

### LINKED ISSUE #20048: NewDataAnnouncer not announcing
#### Expected behaviour
I would expect announcementFormatter to announce points above 10 to VoiceOver and announcer-container in the DOM.

#### Actual behaviour
Nothing is happening when a point is above 10.

https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/accessibility/custom-dynamic

#### Product version
Broken since 9.3.0

#### Affected browser(s)
Chrome

marvin19:
Fixed #20048, `NewDataAnnouncer` was not announcing.


goransle (approved):
Looks good!

oysteinmoseng (changes requested):
Nice 🎉 Could we find a way to unit test that something is added to the announce region `div`, though? Would be nice not to have to test this manually.

TorsteinHonsi:
Pending review by @oysteinmoseng

samples/unit-tests/accessibility/newdataannouncer/demo.js
```diff
@@ -0,0 +1,71 @@
...
+        assert.ok(
+            announcerDiv.innerHTML,
+            'The announceRegion should contain text content after adding a point'
+        );
+    }, 1);
+
+    assert.strictEqual(
+        announcerDiv.innerHTML,
+        '',
+        'The text in the announceRegion should be removed after a short while'
```
oysteinmoseng, line 51: Should this check be moved into the 2nd setTimeout, so that it runs after the announcement text has first been added, and then removed again?

oysteinmoseng (changes requested):
Nice, small question to the test only!

oysteinmoseng (approved):
🎉

PR #20059, opened 2023-11-02
Tools/fix unit tests node20

goransle:
---
  - https://app.asana.com/0/0/1205842783926761


pawellysy (commented):
Changes seem reasonable, but could we trigger the Stock/Dashboards benchmark for this PR? Maybe let's add a manual event dispatch for each of the GH action 🤔

goransle:
> Changes seem reasonable, but could be trigger the Stock/Dashboards benchmark for this PR? Maybe let's add a manual event dispatch for each of the GH action 🤔

Good point. Will add that!

pawellysy (approved):
Stock Benchmark passes, Looks good to me!

PR #20060, opened 2023-11-02
tools/nightly-changelog-review

goransle:
Changelog ends up here: https://vrevs.highsoft.com/api/assets/changelog/nightly-changelog.html

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -156,6 +156,7 @@ jobs:
 
       - name: Generate and upload nightly changelog
         run: |
+          npx gulp jsdoc-options
           node changelog/generate --since $(git describe --abbrev=0 --match 'v*') --review
           aws s3 cp ./changelog/review.html s3://${HIGHCHARTS_VISUAL_TESTS_BUCKET}/changelog/nightly-changelog.html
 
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly.yml
```diff
@@ -141,9 +141,29 @@ jobs:
...
+      - run: npm i
+
+      - name: Build tree.json
+        run: |
+          npx gulp scripts
+          npx gulp jsdoc-options
+
+      - name: Generate and upload nightly changelog
+        run: |
+          node changelog/generate --since $(git describe --abbrev=0 --match 'v*') --review
```
TorsteinHonsi, line 164: Do you need the `--since` argument? It defaults to the latest release. I see that the description on line 10 is wrong, it says it defaults to the latest commit, which makes no sense.

TorsteinHonsi (changes requested):
Awesome! One comment inline. Also, is it easy to post this to Slack? Then we can tag each other with things to fix, best practices etc. (Read: I can pretend school teacher 🤓 )

.github/workflows/nightly.yml
goransle: Looks like no (and since octokit also prefers annotated tags we don't have to filter out Dashboards releases)

goransle:
> Awesome! One comment inline. Also, is it easy to post this to Slack? Then we can tag each other with things to fix, best practices etc. (Read: I can pretend school teacher 🤓 )

Link is added to the latest post in the `#CI` channel. Could possibly also add the changed content from the markdown files

TorsteinHonsi (approved):

PR #20062, opened 2023-11-02
Bugfix: Scatter boost exit

### LINKED ISSUE #18804: Zooming-in below `boostThreshold` doesn't disable the `Boost`
#### Expected behavior
As written in the Boost docs (https://www.highcharts.com/docs/advanced-chart-features/boost-module):

> The intended way of using the module, is to set thresholds in such a way that the SVG-renderer “takes over” rendering when zooming in. This approach gives the expected interactivity when the points are less dense, coupled with high performance when the point density is high.

But if we zoom on a boosted series so that the number of visible points on the chart is less than `series.boostThreshold` the series is still boosted (and, for example, the `dataLabels` do not show up).

https://jsfiddle.net/BlackLabel/yke96xbj/

1. Try zooming in so that there are less than 100 points (less than `series.boostThreshold`)
2. The `dataLabels` do not show up and the series is still boosted

bre1470:
Fixed #18804, zooming in below `boostThreshold` didn't disable boost on scatter series.


ts/Extensions/Boost/BoostSeries.ts
```diff
@@ -849,6 +845,76 @@ function getPoint(
...
+        } else {
+            cropped = true;
+        }
+    }
+
+    series.cropped = cropped;
+    series.cropStart = 0;
+    series.processedData = processedData;
+    series.processedXData = processedXData;
+    series.processedYData = processedYData;
```
TorsteinHonsi, line 913: Do we need all these properties? As far as I can see, cropping don't make sense in this scenario.

Also, we shouldn't have to keep both `processedXData`/`processedYData` and `processedData`. In fact, the code in `seriesRenderCanvas` shows that `processedData` is not used whenever `processedXData` is around (through the statement `useRaw = !xData`). I suggest we remove `processedData` and focus on efforts on `proxessedXData` and `processedYData`.

ts/Extensions/Boost/BoostSeries.ts
```diff
@@ -849,6 +845,76 @@ function getPoint(
...
+        return;
+    }
+
+    if (!series.boost) {
+        series.data.length = 0;
+        series.isDirty = true;
+        return;
+    }
+
+    series.yAxis.setTickInterval();
```
TorsteinHonsi, line 865: What's the purpose of this? It may potentially drag in a hairball of loops and stuff... If legit, we need a comment.

bre1470: It makes sure that I get proper extremes in the following lines.

ts/Extensions/Boost/BoostSeries.ts
bre1470: This is for `Series.generatePoints` to pick up the right amount points after the boost exit. So this covering both situations: `enterBoost` with `Series.renderCanvas` and `exitBoost` with `Series.render`.

goransle (approved):
Seems wholesome to me 👍

ts/Extensions/Boost/BoostSeries.ts
```diff
@@ -670,6 +667,20 @@ function enterBoost(
...
+
+    // Destroy existing points after zoom out
+    if (
+        series.type === 'scatter' &&
+        series.data.length
+    ) {
+        for (const point of series.data) {
+            if (point && point.destroy) {
+                point.destroy();
+            }
```
TorsteinHonsi, line 679: Shorter: `point?.destroy?.()`

ts/Extensions/Boost/BoostSeries.ts
```diff
@@ -670,6 +667,20 @@ function enterBoost(
     if (series.labelBySeries) {
         series.labelBySeries = series.labelBySeries.destroy();
     }
+
+    // Destroy existing points after zoom out
+    if (
+        series.type === 'scatter' &&
```
TorsteinHonsi, line 673: In other parts of HC, we consistently use `series.is(type)` for series type checking, otherwise we will not capture series types that are inherited. In Boost we haven't done that change, mainly because we haven't had issues with it yet. But I think we should go for `series.is('scatter')`.

ts/Extensions/Boost/BoostSeries.ts
```diff
@@ -849,6 +859,111 @@ function getPoint(
...
+ */
+function scatterProcessData(
+    this: Series,
+    force?: boolean
+): (boolean|undefined) {
+    const series = this,
+        options = series.options,
+        xAxis = series.xAxis,
+        yAxis = series.yAxis;
+    (window as any).TheSeries = series;
```
TorsteinHonsi, line 873: 👀

ts/Extensions/Boost/BoostSeries.ts
```diff
@@ -849,6 +859,111 @@ function getPoint(
...
 
+/**
+ * @private
+ */
+function scatterProcessData(
+    this: Series,
+    force?: boolean
+): (boolean|undefined) {
+    const series = this,
+        options = series.options,
```
TorsteinHonsi, line 870: Use deconstructor?

ts/Extensions/Boost/BoostSeries.ts
```diff
@@ -849,6 +859,111 @@ function getPoint(
...
+    if (
+        !series.isDirty &&
+        !xAxis.isDirty &&
+        !yAxis.isDirty &&
+        !force
+    ) {
+        return false;
+    }
+
+    // Required to get tick-based zoom ranges
```
TorsteinHonsi, line 885: A little more elaboration? I _think_ the reason why we need to run the whole `setTickInterval` function is because it takes axis settings like `minPadding`, `maxPadding`, `startOnTick` and `endOnTick` into account. It is hard to account for these without actually running the whole function.

ts/Extensions/Boost/BoostSeries.ts
```diff
@@ -849,6 +859,111 @@ function getPoint(
...
+        yExtremes = yAxis.getExtremes(),
+        yMax = yExtremes.max,
+        yMin = yExtremes.min;
+
+    // Skip processing in non-boost zoom
+    if (
+        !series.boosted &&
+        xAxis.old &&
+        yAxis.old &&
+        xMin >= (xAxis.old.min || 0) &&
```
TorsteinHonsi, line 905: This looks suspicious: `(xAxis.old.min || 0)`. What if we're dealing with subzero values? Since the max falls back to `Number.MAX_VALUE`, I suspect the min should fall back to `-Number.MAX_VALUE`.

Also, a simple `||` operator may not be safe, because if `min` or `max` are the number 0, it will count as falsy. I think what we want to use is the `??` operator.

ts/Extensions/Boost/BoostSeries.ts
```diff
@@ -849,6 +859,111 @@ function getPoint(
...
+            cropped = true;
+        }
+    }
+
+    // Set properties as base processData
+    series.cropped = cropped;
+    series.cropStart = 0;
+    series.processedData = processedData; // For un-boosted points rendering
+    series.processedXData = processedXData; // For boosted points rendering
+    series.processedYData = processedYData;
```
TorsteinHonsi, line 962: Do we need to set both these varieties? It will double the memory use. I don't remember if there was a legit reason for storing both `processedData` and `processedXData`/`processedYData` at the same time. But if there is, we need a proper explanation in a comment.

TorsteinHonsi (changes requested):
Thanks, great work! Some more comments in line. All in all, I'm requesting thorough comments for everything that looks suspicious, otherwise we will find ourselves asking the same questions over and over.

ts/Extensions/Boost/BoostSeries.ts
TorsteinHonsi: PS: A trick I use to stop myself from committing temporary debug code is to write it in a way that doesn't pass lint. That way (and it happens often), I won't be able to commit until I've removed it.

ts/Extensions/Boost/BoostSeries.ts
TorsteinHonsi: Note the difference between `Number.MIN_VALUE` and `-Number.MAX_VALUE`. `Number.MIN_VALUE` is the least possible number _above zero_.

bre1470: Thank you. It's always a trap for me.

TorsteinHonsi (approved):
Thanks 🎉

PR #20067, opened 2023-11-03
dash/toggle-edit-mode-a11y

karolkolodziej:
Improved accessibility by changing the context menu toggle into a button.

---
  - https://app.asana.com/0/0/1205518309446628


ts/Dashboards/EditMode/EditGlobals.ts
```diff
@@ -359,7 +362,9 @@ namespace EditGlobals {
     }
 
     export interface LangAccessibilityOptions {
+        [key: string]: any;
```
sebastianbochan, line 365: Do we need `any` here? Maybe the generic type would be useful?

sebastianbochan (changes requested):

Vazonik (approved):

marvin19 (approved):

PR #20068, opened 2023-11-03
docs/20061-api-improvement-jitter-boost

### LINKED ISSUE #20061: Missing information in API about `jitter` not available in Boost module
#### Expected behaviour

There is an information in API about `jitter` not available when using Boost module.

#### Actual behaviour
There is no information in API about `jitter` not being available for Boost module.

[https://api.highcharts.com/highcharts/series.scatter.jitter](https://api.highcharts.com/highcharts/series.scatter.jitter)

jedrzejruta:
Fixed #20061, added API info about `jitter` not being supported when using Boost mode.


TorsteinHonsi (approved):
Thanks!

PR #20069, opened 2023-11-03
docs/20012-fix-wrong-histogram-dependency

### LINKED ISSUE #20012: Wrong module dependency for histogram series in API
Should point to [modules/histogram-bellcurve](https://code.highcharts.com/modules/histogram-bellcurve.js).

https://github.com/highcharts/highcharts/blob/master/ts/Series/Histogram/HistogramSeriesDefaults.ts#L41C18-L41C35 

Also check if there are other instances of wrongly pointing to the non-existent `modules/histogram` anywhere.

madepiet:
Fixed #20012, changed the histogram series to correctly point to `modules/histogram-bellcurve` instead of the non-existent module.


TorsteinHonsi (approved):
Thanks!

PR #20075, opened 2023-11-06
Removed any cast, added property check

goransle:


hubertkozik (changes requested):
Fix looks great! Can you add a `unit-test` to be sure that this bug won't exist in the future? :)

hubertkozik (approved):

TorsteinHonsi (changes requested):
Thanks!

Regarding the unit test - whenever it makes sense we try to hook new assertion on to existing charts instead of creating a new test with a new chart for each bug fix. Because the number of charts is the main reason why the local tests take time. In this case for example, it should be possible to take one of the tests in `unit-tests/boost/general`, and and remove a series and do the assertion.

TorsteinHonsi (approved):

Issue #18062, opened 2022-11-30
Pinch to zoom stops working after interacting with legend or title


ludvikbrodl:
#### Expected behaviour

The pinch to zoom should keep working as normal even after clicking on other elements on the chart.

#### Actual behaviour

The pinch to zoom stops working until the chart is double tapped

#### Live demo with steps to reproduce

http://jsfiddle.net/hL7ebz5q/6/

Reproduceable on atleast
* Android 13 Samsung S21 G5 - chrome 107.0.5304.105 and firefox 107.1.0
* Android 9 Samsung Galaxy S8 - chrome 107.0.5304.141

1) Open the jsfiddle on android,
2) Single click the title or toggle the dataline by clicking the legend item twice.
3) Try to zoom with pinch
4) No zoom is applied after the pinch is released
5) double tap inside the chart
6) pinch to zoom now works again as intended
7) click on the chart title or the legenditem twice
8) the zoom stops working again

#### Product version
https://github.highcharts.com/master/highcharts.js
and
10.3.2
#### Affected browser(s)
* Android 13 Samsung S21 G5 - chrome 107.0.5304.105 and firefox 107.1.0
* Android 9 Samsung Galaxy S8 - chrome 107.0.5304.141
* Tested on 2 Iphones, neither experience this bug in chrome or safari.

raf18seb:
Hi @ludvikbrodl, thank you for reporting the issue!
Confirmed on S22 Ultra, Chrome 108.0.5359.61 (working fine on iPhone)

`Internal note:`
Works fine on Android in v10.2.1, broken since 10.3.0 (of course, unable to bisect on mobile)

jacob-haskins:
This is also happening for me. Is there a workaround or fix in the works?

karolkolodziej:
Thanks @jacob-haskins  for bringing our attention to this issue! In the upcoming weeks, we are going to take a closer look at this issue.
We will keep you posted 😉

hubertkozik:
`Internal note:`
Bisected to eafc0873453de2df9f3ae13152708908f9e94b3b.

hubertkozik:
Hello guys!
I have prepared a workaround for that issue. You can check it in the demo below. If you want to use it in your project now, just copy-paste the IIFE function with the plugin.

**Live demo:** https://jsfiddle.net/BlackLabel/qo0wrz6s/

jacob-haskins:
[...]

Issue #6804, opened 2017-06-06
Mobile Browser - CHart with zoom not able to scroll the page


olivierscheffler:
#### Expected behaviour
When I see a zoom's chart, I should be able to still scroll the page where the chart is and not be in conflict with the charts event

#### Actual behaviour
I'm not able to scroll the page and some events are live on the charts.
More precision : on beginning it seems to be able to scroll the page but if you zoom once and return to default view you will not be able.
In my case i'm never allow to scroll the page when touching the chart on mobile

#### Live demo with steps to reproduce
[DEMO](http://jsfiddle.net/z70vzor9/)

#### Affected browser(s)
Chrome on Android and Safari on iOs

KacperMadej:
Hi @SojeSkormdark ,

Thank you for reporting about the problem. However, I am not able to recreate the problem on Chrome on Android. Could you provide more detailed steps required to replicate the problem on the demo you have provided?

sebastianbochan:
I was also cannot be able to replicate it on the iOS.

SojeSkormdark:
@KacperMadej @sebastianbochan ,

Hi,

I try another time on my chrome in android (last version).
If you run the jsfiddle, try to scroll the right part from the chart itself you will be able to scroll the entire page with no trouble.
But if you zoom on the chart and try to scroll the page from the chart you will not be able. ANd if you click on Reset zoom you will not be able to scroll the page from the chart. (When I say from the chart, I mean put your finger on the chart and try to scroll the right section)

Thank you,

sebastianbochan:
@SojeSkormdark could you prepare a simple clip from your device?

SojeSkormdark:
@sebastianbochan  You are not able to replicate?
Here a screen recorder from my android with touchpoint visible.
[screen-recorder.zip](https://github.com/highcharts/highcharts/files/1057813/screen-recorder.zip)
Hope you will be able to look at it with no trouble!

Thx

KacperMadej:
@SojeSkormdark Thank you for the video. I was able to recreate the problem.

*Internal note:* After a pinch triggering zoom, drag on plot area are captured and interpreted as hover over chart - only tooltip is triggered and page scroll is disabled.

SojeSkormdark:
On my version I use for a web site (4.1.9), I'm never able to scroll over the zoom's chart but in this version I'm able at the beginning.. I don't know if it's a version problem or not but this is a strange behaviour..

[...]

PR #20077, opened 2023-11-06
Pinch zoom refactor

### LINKED ISSUE #5270: Panning on touch devices doesn't move the axis and data
When panning on touch screens devices I expected to see a moving axis and data.

Actually I see a not moving static axis until touchEnd Event and white Space instead of data

http://jsfiddle.net/gh/get/jquery/1.7.2/highslide-software/highcharts.com/tree/master/samples/highcharts/chart/pankey/

On touch and non-touch devices I use Chrome.

### LINKED ISSUE #19872: Panning does not respect axis min/max values
#### Expected behaviour
After zooming into a chart with specified `yAxis.min` and `yAxis.max` values, the user should freely pan the chart with respecting min/max values.

#### Actual behaviour
After zooming into a chart with specified `yAxis.min` and `yAxis.max` values, panning does not respect min/max values.

https://jsfiddle.net/BlackLabel/vjhzw71a/
1) Zoom into the bottom part of the points.
2) Click the shift key and pan in y direction -> yAxis.min and yAxis.max are not respected.

#### Product version
Highcharts v11.1.0

### LINKED ISSUE #19976: Jumping of the time period when we zoom in and out or when we shift to the right
#### Expected behaviour
When we pinch in or out it should be work well.

#### Actual behaviour
Jumping of the time period when we zoom in and out or when we shift to the right

https://jsfiddle.net/KrupaliClavier/7vea0Lfg/13/

#### Affected browser(s)
Web view in reactnative

https://github.com/highcharts/highcharts/assets/140229103/a21f52cd-9e99-4df2-97ff-20ecaaae1627

[jumping_to_right.webm](https://github.com/highcharts/highcharts/assets/140229103/7a3229bb-acae-431f-b1c9-7a356a380dab)

### LINKED ISSUE #20070: Panning does not respect minPadding and maxPadding
#### Expected behaviour
Panning should respect minPadding and maxPadding when set.

#### Actual behaviour
When panning and minPadding or maxPadding are set, panning range is limited to dataMin and dataMax and not allowing to pan further.

https://jsfiddle.net/BlackLabel/u396h8ad/

1. Zoom in
2. Try to pan beyond dataMin and dataMax

#### Product version
Highcharts v11.2.0

#### Affected browser(s)
All

TorsteinHonsi:
Refactored the touch handling and pinch zoom logic to redraw the axes and series continuously, instead of the previous scaling preview then setting extremes at the end. Also a deeper refactor and unification of the whole zoom, pan and mousewheel logic.
___

Closes #5270, closes #19872, closes #19976, closes #20070.

#### Preview
Live preview as of 2023-11-06: https://highcharts.github.io/highcharts-utils/samples/#gh/e69d358214/sample/highcharts/demo/line-time-series
___
#### Background
Previously, touch zooming in Highcharts would transform-scale the series groups to provide a feedback to the user how the zooming would affect the chart. But the actual zooming wasn't performed until the touch ended. This approach was chosen because mobile phone hardware at the time (~2011) could't handle actual live rescaling of the axes. The downside is that the preview does not perfectly represent what the actual zooming would look like, and the axis ticks and labels are not updated.

The old approach was also a bit over-complicated as it emulated a selection marker and did all the computation through that.

#### New approach
The new approach is to rescale the axes and redraw the chart on each touch move.

#### To do 
 - [x] Re-implement map zooming
 - [x] Do not allow zooming in with one finger when `tooltip.followTouchMove` is `false` and we're panning close to the edge. Previously handled in [pinchTranslateDirection](https://github.com/highcharts/highcharts/blob/dd6879c0826b89d96617e0c117e6f45b6a96f8e5/ts/Core/Pointer.ts#L1576C1-L1582C14).
 - [x] Apply correction for padding of column series (`pointRange`) etc.
 - [x] Handle `startOnTick` and `endOfTick`, like we [currently do in the mouse wheel zoom logic](https://github.com/highcharts/highcharts/blob/e69d35821438ba85354d5859b0d17e4a1cde684d/ts/Extensions/MouseWheelZoom/MouseWheelZoom.ts#L120-L164). Share the logic if possible.
 - [x] Zooming on `xy` is wobbly. Probably the `suppressEndOnTick` logic must be global (or the `trigger` fixed), otherwise setting the extremes on the x-axis will redraw the y-axis with `endOnTick`.
 - [x] Fix reversed x-axis and y-axis move direction, probably related.
 - [x] Handle smooth zooming on ordinal axes. This problem too is shared with the mouse wheel zoom logic, except we haven't resolved it. Look for a common solution. Edit: works well with touch-zoom.
 - [x] Unify the logic of mouse-pan, touch-pan and zoom, and mousewheel zoom. All three ultimately come down to a) moving the pointer [x, y] pixels and b) zooming in or out. Much like `MapView.zoomBy`.
 - [x] Check if `Chart.zoom` and the `selection` event can use `Chart.transform`.
 - [x] Manually check all the combination of options related to touching and zooming. Check the conditions and comments of the removed code. Look for edge cases.
 - [x] Read through Pointer code and zoom/pan-related code in Chart and Axis for workarounds related to the previous touch-behaviour.
   - [x] Why `allowZoomOutside`?
   - [x] Check `hasPinched` and the #18062 fix.
 - [x] Simplify `panningState`. Possibly a separate PR. It seems overcomplicated. In essence it is just a placeholder for the full-range `dataMin` and `dataMax`. Later, `minPadding` and `maxPadding` has been attached to the logic, but should be unrelated. Consider simple `Axis.dataMinAll` and `dataMaxAll` properties. Consider refactoring out min/max-padding and threshold, softThreshold handling from `Axis.setTickInterval` and reuse in the `Chart.transform` function. Related to #20105.
 - [x] Consider special handling in Boost mode. This may be the case where live redraws are still too slow. In that case, re-implement SVG transforms, but do it more directly without going through the selection marker emulation.
    - [x] Correct clipping. Make sure the image is nested inside a clipping group, so that we don't have to compute clipping. Like the map series.
    - [x] Pinch-zooming - translation is off when moving the two fingers.
    - [x] Why is the y-axis re-computing during x-axis zoom? Look for the possible performance sink.
    - [x] What about global rendering target?
 - [x] Consider a Zooming or Interaction composition that would gather all the zoom-related functionality from Chart, Pointer and Axis. MouseWheelZoom could be an extension to that. The current `chart.zooming` property could be repurposed to point directly to the composition. 
 - [x] The `suppressEndOnTick` code could be unified, equal for mouse-pan, touch-zoom/pan and mousewheel.
 - [x] When to show reset zoom button? Investigate and run some combinations.
 - [x] Mousewheeling on stock => Reset zoom does nothing
 - [x] Check out if minPadding, maxPadding and threshold can be refactored out and  shared with `setTickPositions`.
 - [x] On multiple panes, make sure only zoomed axes are affected. Check target or reference position (center point?) against axis position. Add/modify tests.
 - [x] Why is `mouseDownX` and `mouseDownY` registered on both Chart and Pointer?
 - [x] Regression in overscroll: Panning not working, goes to max when dropped first time. Depends on https://github.com/highcharts/highcharts/issues/20430
 - [x] Regression: Boosted scatter only zooms x-axis (with mouse-drag): http://localhost:3030/samples/#view/highcharts/boost/scatter
 - [x] Verify that https://github.com/highcharts/highcharts/issues/6804 is not broken


samples/stock/demo/area/demo.html
```diff
@@ -1,5 +1,5 @@
 
-<script src="https://code.highcharts.com/stock/highstock.js"></script>
+<script src="https://code.highcharts.com/highstock.js"></script>
```
KacperMadej, line 2: This change will break [the online demo](https://www.highcharts.com/demo/stock/area) because the URL is not working for the CDN.

ts/Core/Pointer.ts
```diff
@@ -2061,13 +1905,14 @@ class Pointer {
                 // however, the touchmove doesn't fire unless the finger moves
                 // more than ~4px. So we emulate this behaviour in Android by
                 // checking how much it moved, and cancelling on small
-                // distances. #3450.
+                // distances. #3450. Tested and still relevant as of 2023.
```
KacperMadej, line 1908: or maybe 2024 even? :)

KacperMadej (commented):
This might be an edge case or a hidden feature:
Enable zoomType:'xy' and use a touch device. Press 2 fingers on an imaginary horizontal line; keep one finger steady; move the other finger up and down crossing the imaginary line; notice the y axis range is getting smaller and smaller - after a few moves I can go from 0-100 to numbers like 50.0001-50.0002. This also applies to x axis if done vertically. This doesn't happen on a map.

ts/Extensions/MouseWheelZoom/MouseWheelZoom.ts
```diff
@@ -62,282 +63,73 @@ const optionsToObject = (
...
+    );
+
+    let axes: Array<Axis> = [];
+    if (type === 'x') {
+        axes = xAxis;
+    } else if (type === 'y') {
+        axes = yAxis;
+    } else if (type === 'xy') {
+        axes = chart.axes;
     }
```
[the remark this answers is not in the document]
TorsteinHonsi: @DJTechnoo Thanks, I tried, but the switch actually produces more verbose code and increases the compiled and minified file size.

KacperMadej (approved):

DJTechnoo (approved):
Amazing!

Issue #17582, opened 2022-08-08
Multiple redraw events fired during annotations line drawing


kamil-musialowski:
# Description of the feature
Function responsible for creating a `segment line` could be improved. Right now, redraw event is fired on every mouse move, during line drawing which is very performance heavy (the same issue is not happening when updating or dragging a line).
Since extremes are not changed, is there a point to redraw all points on the chart?

# Library related to the feature
Highcharts Stock with Stock Tools GUI

# Proof of Concept/Live example for the feature
DEMO: https://jsfiddle.net/BlackLabel/rdLu86z4/

0. Open up the console.
1. Click the 3rd option from the top (draw a line).
2. Click the mouse on the chart **once**, to create a starting point.
3. Move mouse around and watch how many redraw events are called.

Also, watch the console while updating the same line.

----

**You can vote for this feature by adding a thumbs-up reaction to this post.**

pawelfus:
**Workaround:**
Replace `chart.redraw()` with `chart.drawAnnotations()` in the `Annotation.update()` method.
Demo: https://jsfiddle.net/BlackLabel/rdLu86z4/1/
Snippet:
```js
Highcharts.Annotation.prototype.update = function(userOptions, redraw) {
  var chart = this.chart,
    labelsAndShapes = this.getLabelsAndShapesOptions(this.userOptions, userOptions),
    userOptionsIndex = chart.annotations.indexOf(this),
    options = Highcharts.merge(true, this.userOptions, userOptions);
  options.labels = labelsAndShapes.labels;
  options.shapes = labelsAndShapes.shapes;
  this.destroy();
  this.constructor(chart, options);
  // Update options in chart options, used in exporting (#9767):
  chart.options.annotations[userOptionsIndex] = options;
  this.isUpdating = true;
  if (Highcharts.pick(redraw, true)) {
    chart.drawAnnotations();
  }
  Highcharts.fireEvent(this, 'afterUpdate');
  this.isUpdating = false;
};
```
___
*Internal note:*
In fact, we should fix performance issues here:

1. Unnecessary grouped data destruction. Regression after #10290
![Zrzut ekranu 2022-08-8 o 14 18 15](https://user-images.githubusercontent.com/1453926/183416332-e6e6ae0c-2a2e-41fa-a614-efe4c6b19467.png)
[...]

PR #20080, opened 2023-11-06
bugfix/19861-search-point-undefined

### LINKED ISSUE #19861: `series.searchPoint` returning undefined + Discussion on editing existing line annotation so that end-points attach to nearest series point
#### Expected behaviour

`Series.searchPoint` should always work.

#### Actual behaviour

`Series.searchPoint` is randomly returning `undefined` sometimes.

**Live - https://jsfiddle.net/5x7g6vwf/4/**

My use-case is to add line annotation such that the end points of line attach to the nearest series point. I am using custom bindings to solve this. It is working correctly when annotation is first added, but in second step, when user clicks again to finalize the second end-point, I want the second point also to attach to the nearest series point - which does not seem to happen because `series.searchPoint` is returning `undefined` for all series.

searchpoint-bug

#### Product version
Latest Highcharts

#### Affected browser(s)
Chrome

kamil-musialowski:
Fixed #19861, `kdTree` was deleted on annotation redraw.

Demo for testing: https://jsfiddle.net/BlackLabel/jetn459p/

_Internal note:_
The problem was in the annotation update (from the OP demo), not in the search point itself. KD Tree was deleted every time annotation was updated (which called `chart.redraw()`). https://github.com/highcharts/highcharts/blob/acaa13a09f1bd7e463b30af984a07ce396bdc1d6/ts/Core/Series/Series.ts#L3451-L3460

Connected to #17582.

---
  - https://app.asana.com/0/0/1205657626316268


bm64 (approved):

pawelfus (changes requested):
Nice catch! Missing test though 🤔 

___
I'm fine with this solution, however I was curious why `redraw()` destroys the kdTree? It seems like something we could improve in Core. 

Just made a simple demo: https://jsfiddle.net/BlackLabel/Lozm98kd/ - two charts, just bit different datasets (one point less in the second chart), and we get different behaviour of `chart.redraw()`. 

Debugged a little:
- charts look similar, however xAxis labels have different tickInterval. With interval=1, logic to animate labels (move them like categorized) jumps in
- this sorting? animating? implementation sets [axis.isDirty](https://github.com/highcharts/highcharts/blob/dd6879c0826b89d96617e0c117e6f45b6a96f8e5/ts/Core/Axis/Tick.ts#L1093-L1120) flag to true
- dirty axis = dirty series => let's destroy the kdTree 🤷 

I'm not sure if this is something that we should improve right now. tickInterval=1 is unlikely to happen in Stock charts, and with the workaround/fix by Kamil, we can completely ignore the issue.

FYI @TorsteinHonsi

kamil-musialowski:
@pawelfus Thank you! I've added a test, however it was not possible to simply call `annotation.update()`, I had to go with `annotation.update( { shapes: [{}] } )`. Are we cool with it, or it should be reported as a bug?

pawelfus:
Yes, we are cool with that

DJTechnoo (approved):

bm64 (approved):

pawelfus:
The remaining thing is a green light from Torstein: are we okay with breaking a non-written rule that `redraw=boolean` parameter means `chart.redraw()`? 

As of now, it is [an undescribed param](https://api.highcharts.com/class-reference/Highcharts.Chart#addAnnotation).

TorsteinHonsi:
@pawelfus 

> The remaining thing is a green light from Torstein: are we okay with breaking a non-written rule that redraw=boolean parameter means chart.redraw()?

Yes, I think that's fine. Since we don't have to redraw the whole chart in order to render this annotation. The question is - if we call `addAnnotation` with redraw = false, then run `chart.redraw()` - will that work as expected?

pawelfus:
> Yes, I think that's fine. Since we don't have to redraw the whole chart in order to render this annotation. The question is - if we call `addAnnotation` with redraw = false, then run `chart.redraw()` - will that work as expected?

Yes, it will work fine. The redraw of all annotations is attached to `chart.redraw` event, so changes made by `addAnnotation()` will be picked up.

I just realized I mentioned the wrong method. I wanted to discuss [`annotation.update()`](https://api.highcharts.com/class-reference/Highcharts.Annotation#update), not `addAnnotation()` 🤦. Anyway, the answer remains the same: the `chart.redraw()` method will apply changes made in `annotation.update(options, redraw=false)`.

TorsteinHonsi (approved):

PR #20082, opened 2023-11-06
dash/kpi-chart-value-syncing

Vazonik:
Added `valueInChart` option to the KPI Component.


karolkolodziej (commented):
Wouldn't it be best to have all the sync related options in the sync config?
Like for example:
```js
    sync: {
      extremes: {
       	enabled: true,
        seriesToSync: 'series-id / series index',
        pointToSync: 'point-id / point index'
      }
    },
```

I can imagine that this might be useful in the future for syncs like `highlight`:
```js
    sync: {
      highlight: {
       	enabled: true,
        showTooltio: false,
        showCrosshair: true
      }
    },
```

karolkolodziej:
I would use it and name it like this: 
```js
    linkedValueTo: {
      enabled: true,
      seriesIndex: 0,
      pointIndex: 0
    },
```

ts/Dashboards/Plugins/KPIComponent.ts
```diff
@@ -440,11 +446,56 @@ class KPIComponent extends Component {
...
             this.prevValue = prevValue;
         }
     }
 
+    /**
+     * Handles updating chart point value.
+     *
+     * @internal
+     */
+    public setChartValue(
```
karolkolodziej, line 460: Maybe let's call it `linkValueToChart`

karolkolodziej (approved):

ts/Dashboards/Plugins/KPIComponent.ts
```diff
@@ -314,6 +318,8 @@ class KPIComponent extends Component {
         this.contentElement.style.display = 'flex';
         this.contentElement.style.flexDirection = 'column';
 
+        this.setChartValue();
```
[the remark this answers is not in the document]
Vazonik: I tested it and came to the conclusion that it's better to execute this method in `load`. When updating the component, `setValue` and `render` are called at once, so the chart would be updated too many times unnecessarily. Also, I can't do it only in `render`, because the extremes sync executes only `setValue`, without `render`.

docs/dashboards/kpi-component.md
```diff
@@ -87,6 +90,10 @@ Dashboards.board('container', {
 });
 ```
 
+By default, the KPI value is synchronized with the Y value of the first point in the first series. To turn off the synchronisation, disable the [linkedValueTo](https://api.highcharts.com/dashboards/#types/Dashboards_Plugins_KPIComponent.KPIComponent.LinkedValueToOptions) option as in the example above.
```
sebastianbochan, line 93: Should be `synchronization`

ts/Dashboards/Plugins/KPIComponent.ts
```diff
@@ -708,6 +759,26 @@ namespace KPIComponent {
          * Callback function to format the text of the value from scratch.
          */
         valueFormatter?: ValueFormatterCallbackFunction;
+        /**
+         * This option lets you toggle KPI value connection with the chart and
```
sebastianbochan, line 763: I would replace `lets you tobble` with `allows user to toggle`

sebastianbochan (approved):

PR #20085, opened 2023-11-06
bugfix/19922-range-selector-all-state

### LINKED ISSUE #19922: RangeSelector "All" not highlighted when the selected button is not enabled
#### Expected behaviour

When the range corresponding to the `selected` button index is wider than the full dataset, the "All" button should be selected.

#### Actual behaviour

The "All" button is shown as disabled. Selecting a narrower range then brings the "All" button back to being enabled.

https://jsfiddle.net/x6gqr9pt/ (modified from [this official demo](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/stock/rangeselector/allbuttonsenabled-true/))

#### Product version
Highstock v11.1.0

#### Affected browser(s)

Chrome 117

bm64:
Fixed #19922, 'all' option now auto-selects if data range is smaller than selected range.

### DIFF
--- a/samples/unit-tests/rangeselector/buttons-range-state/demo.js
+++ b/samples/unit-tests/rangeselector/buttons-range-state/demo.js
@@ -149,7 +149,7 @@ QUnit.test('Range selector buttons states (#3375)', function (assert) {
                     }
                 ],
                 inputEnabled: false,
-                selected: 0
+                selected: 2
             },
             series: [
                 {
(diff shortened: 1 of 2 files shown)


pawelfus (changes requested):
Looks good. Could you also update API's `.select` description? Informing about this behaviour.

pawelfus:
Wait, any idea why it selects now YTD and All?

DJTechnoo (changes requested):
This looks great! One small observation: When testing the solution on the original demo, if you drag the first marker to the right, the 'All' button is still highlighted. It works fine again after pressing 'All' and repeating the step.

kamil-musialowski (approved):
No questions other than my colleagues comments, good job!

DJTechnoo (approved):
Looks great!

pawelfus (approved):
Thanks!

PR #20095, opened 2023-11-07
Enabled codescanning through GH Actions.

oysteinmoseng:
As discussed to pull plug on CircleCI.

---
  - https://app.asana.com/0/0/1205770233445941


bre1470 (approved):
Looks good. 👍

oysteinmoseng:
Looks like lots of issues detected, so we should fix that before merge. Will add a task for the team.

PR #20096, opened 2023-11-07
Tools/actions chrome version

goransle:
Fixes chrome and chromedriver version mismatch

---
  - https://app.asana.com/0/0/1205896333519740

### DIFF
--- a/.github/workflows/highcharts-headless.yml
+++ b/.github/workflows/highcharts-headless.yml
@@ -74,17 +74,13 @@ jobs:
           cache: 'npm'
       # - uses: browser-actions/setup-chrome@latest
       #   with:
-      #     chrome-version: stable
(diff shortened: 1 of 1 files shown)


.github/workflows/highcharts-headless.yml
```diff
@@ -89,14 +95,11 @@ jobs:
       ### TASKS ###
       - name: Build Highcharts
         run: npx gulp scripts
-      - name: Check Chrome
```
bre1470, line 92: Please keep this in, as it is important help to debug issues with tools.

bre1470 (changes requested):

marvin19 (approved):

bre1470 (approved):
Nice!

PR #20099, opened 2023-11-07
Website/countdown clock

nasin:
Hi,

For our Cyber Monday promotion, I created a couple of countdown clocks. I tried to keep them all in one demo, but it was too laborious with all the CSS, etc., so I split them into two demos.

"Cyber Clock" contains a countdown gauges for a strip we're placing above out main site navigation and gauges for the homepage hero animation. Here's a proof of concept: [https://www.goodwiththat.com/cyber-homepage/cyber-home.html](https://www.goodwiththat.com/cyber-homepage/cyber-home.html).

"Cyber Page Hero" is for the Cyber Week landing page. It has the same gauges and a more animated background. Here's a proof of concept: [https://www.goodwiththat.com/cyber-week/hero.html](https://www.goodwiththat.com/cyber-week/hero.html).

I'm not entirely sure if my countdown code is correct, but it seems to be working. I'm sure there's a more efficient way to set it up, so any ideas to improve what I have are appreciated. Thanks!


TorsteinHonsi:
Thanks! 

Since these two demos are so specific to the upcoming date, and they load so many custom graphics and fonts, I am a bit reluctant to add it to the Highcharts product repo. If we merge it now, is it okay if we remove it after Cyber Monday? Next year if we want to use it we will find it in this PR.

nasin:
Yes, I don't see a problem with removing them once the promotion is over.

TorsteinHonsi (approved):

PR #20102, opened 2023-11-08
dash/columnAssignment-point

sebastianbochan:
Added parsing columns to points params in the columnAssignment option.

----
Demo for tests: https://jsfiddle.net/BlackLabel/gqfxhc3a/


bre1470 (commented):
Some questions and remarks in sources. Another question: Does HighchartsComponent support multiple series and is the key mapping supposed to support multiple series?

ts/Dashboards/Plugins/HighchartsComponent.ts
```diff
@@ -1029,7 +1081,13 @@ namespace HighchartsComponent {
          * }
          * ```
          */
-        columnAssignment?: Record<string, string | null>;
+        columnAssignment?: Record<string, string|seriesColumnMap>;
+    }
+
+    /** @private */
+    export interface seriesColumnMap {
```
karolkolodziej, line 1088: IMO it should be public- please add a description and demo with how it works.

ts/Dashboards/Plugins/HighchartsComponent.ts
```diff
@@ -604,13 +605,28 @@ class HighchartsComponent extends Component {
             const columnNames = table.modified.getColumnNames();
             const columnAssignment = this.options.columnAssignment ||
                 this.getDefaultColumnAssignment(columnNames);
+            const seriesColumnMap = columnAssignment.seriesColumnMap;
+            // eslint-disable-next-line max-len
+            const pointColumnMap = (seriesColumnMap as HighchartsComponent.seriesColumnMap)?.pointColumnMap;
```
karolkolodziej, line 610: Why don't you cast it earlier and remove the redundant casting later.

ts/Dashboards/Plugins/HighchartsComponent.ts
```diff
@@ -629,20 +645,37 @@ class HighchartsComponent extends Component {
                     return true;
                 });
 
+            // create empty series for OHLC data
```
karolkolodziej, line 648: It doesn't have to be only OHLC- area range, x-range also works.
I would change the comment to something more generic.

karolkolodziej (changes requested):

ts/Dashboards/Plugins/HighchartsComponent.ts
sebastianbochan: Because the object can be undefined or string (because of using RECORD), so rest parts of code still should verify the type.

test/typescript-karma/Dashboards/Component/tests.js
```diff
@@ -535,3 +535,311 @@ test('DataGrid component with dataTable', async function (assert) {
...
+            cell: 'dashboard-col-3',
+            type: 'Highcharts',
+            connector: {
+              id: 'micro-element'
+            },
+            columnAssignment: {
+              'x': 'x',
+              'mySeries1': 'value',
+              'mySeries2': 'value',
+              seriesColumnMap: {
```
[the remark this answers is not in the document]
sebastianbochan: We can discuss it during the API meeting on friday.

test/typescript-karma/Dashboards/Component/tests.js
```diff
@@ -535,3 +535,311 @@ test('DataGrid component with dataTable', async function (assert) {
...
+              yAxis: [{}, {
+                opposite: true
+              }],
+              series: [{
+                name: 'EUR',
+                yAxis: 1
+              }]
+            }
+          },
+          {
```
Vazonik, line 648: Let's stick to one unified way of grouping brackets. :D I propose
```suggestion
          }, {
```

test/typescript-karma/Dashboards/Component/tests.js
```diff
@@ -535,3 +535,311 @@ test('DataGrid component with dataTable', async function (assert) {
...
+              }, {
+                name: 'fake trend',
+                data: [
+                  [1691971200000, 22],
+                  [1692316800000, 22]
+                ]
+              }]
+            }
+          }
+        ]
```
Vazonik, line 746: It seems like a similar situation here.

Vazonik (approved):
Looks good! Just pay attention to the brackets.

karolkolodziej (approved):

test/typescript-karma/Dashboards/Component/tests.js
```diff
@@ -535,3 +535,309 @@ test('DataGrid component with dataTable', async function (assert) {
...
+            cell: 'dashboard-col-3',
+            type: 'Highcharts',
+            connector: {
+                id: 'micro-element'
+            },
+            columnAssignment: {
+                'x': 'x',
+                'mySeries1': 'value',
+                'mySeries2': 'value',
+                seriesColumnMap: {
```
oysteinmoseng, line 657: I think maybe this should be an array, like `seriesMapping` in Highcharts - otherwise it gets very verbose, but let's discuss that in the team.

oysteinmoseng (changes requested):
Good suggestion!
I think we should discuss the API in the team and make a decision there.

test/typescript-karma/Dashboards/Component/tests.js
oysteinmoseng:
```
 columnAssignment: {
                'x': 'x',
                'mySeries1': 'value',
                'mySeries2': 'value',
                'mySeriesName': { 
                        open: 'something',
                        high: 'somethingElse'
                }
```

oysteinmoseng (changes requested):
Nice, but also docs? Explain the idea of `columnAssignment` in [Highcharts Component docs](https://www.highcharts.com/docs/dashboards/highcharts-component) at least?

oysteinmoseng (approved):
🎉

PR #20107, opened 2023-11-09
dash/project-demo

karolkolodziej:
Added the project management demo.


samples/dashboards/demo/project-management/demo.css
```diff
@@ -0,0 +1,70 @@
...
+    stroke: var(--gray);
+}
+
+#dashboard-chart-1 .highcharts-point.highcharts-color-0 {
+    fill: var(--gray);
+    stroke: var(--gray);
+}
+
+#dashboard-kpi-3 .highcharts-dashboards-component .highcharts-point.highcharts-color-0 {
+    fill: black;
```
sebastianbochan, line 45: Please use hex instead.

Vazonik (approved):
Looks good! 🎉

samples/dashboards/demo/project-management/demo.html
```diff
@@ -0,0 +1,16 @@
...
+<script src="https://code.highcharts.com/highcharts.js"></script>
+<script src="https://code.highcharts.com/dashboards/modules/dashboards-plugin.js"></script>
+<script src="https://code.highcharts.com/gantt/modules/gantt.js"></script>
+<script src="https://code.highcharts.com/modules/accessibility.js"></script>
+
+<div id="container"></div>
+
+<p class="highcharts-description">
+    This demo shows how to use the Highcharts dashboards to create a project
+    management dashboard. <br>
```
TorsteinHonsi, line 11: Prefer using paragraphs, not single breaks. https://github.com/highcharts/highcharts/blob/master/repo-guidelines.md#1-avoid-line-breaks-without-paragraphs

samples/dashboards/demo/project-management/demo.js
```diff
@@ -0,0 +1,455 @@
...
+        linkedValueTo: {
+            enabled: false
+        }
+    }, {
+        cell: 'dashboard-kpi-4',
+        type: 'Highcharts',
+        title: 'Tasks by status',
+        chartOptions: {
+            xAxis: {
+                categories: ['Done', 'To Do', 'In Progress', 'Blocked']
```
TorsteinHonsi, line 175: Pie series don't use `xAxis` 🤔

TorsteinHonsi (commented):
Nice work! I have some comments inline, and some design related.

1. The dashboard title ("Current sprint") has its own card, which looks a bit weird. I think it would look better if it was either outside the cell structure, or the cell background and shadow was removed. I see we have solved this in different ways in the other demos.

2. The #current-sprint div has its own background color, making the background of the top row darker than the rest of the dashboard. Is that intended? It looks a bit unmotivated.

3. On the mobile, we use almost all the screen estate above the fold to show two numbers; completed tasks and incomplete tasks. Maybe it would be better to shrink the two cells to 50% and adjust the font sizes accordingly?

Skjermbilde 2023-11-29 kl  10 51 19

sebastianbochan (changes requested):

samples/dashboards/demo/project-management/demo.js
```diff
@@ -0,0 +1,455 @@
...
+        title: 'Timeline',
+        chartConstructor: 'ganttChart',
+        chartOptions: {
+            chart: {
+                marginLeft: 10
+            },
+            xAxis: [{
+                plotLines,
+                dateTimeLabelFormats: {
+                    day: '%e<br><span style="opacity: 0.5; font-size: 0.7em">%a</span>'
```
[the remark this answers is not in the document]
TorsteinHonsi: > use `<div>`

I don't think the SVG TextBuilder supports that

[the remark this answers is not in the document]
TorsteinHonsi: It doesn't: https://jsfiddle.net/highcharts/df0v6sn4/

samples/dashboards/demo/project-management/demo.js
karolkolodziej: leftover 😅

karolkolodziej:
3. It Seems valid but responsive options don't support the height. I added that to the backlog and we will change the demo after it is implemented.

TorsteinHonsi (changes requested):
Overall it looks great!

The contrast between the series types is too low (they are all very dark), but I know other team members are working on that. 

One thing that I think should be fixed is the tooltip for the Gantt chart. Both start and end date shows "Week from...", which is not very specific. I assume this happens because both dates are defined on the Monday midnight. There are two ways of resolving it; either set an explicit time format for the tooltip, or change the dates. Actually in this case I think it would make sense to reduce the length of the tasks. What about starting Monday 9:00 and ending them Friday at 17:00? That is a more realisting work week, and will probably resolve the tooltip.

TorsteinHonsi:
Thanks! Not a big deal, but wouldn't it be more realistic if the tasks end on Friday 17:00 instead of the next Monday 17:00? Like fill out the work week and keep the weekend off?

samples/dashboards/demo/project-management/demo.css
```diff
@@ -0,0 +1,123 @@
...
+.highcharts-plot-line {
+    stroke-width: 3px;
+    stroke: var(--highcharts-neutral-color-100);
+}
+
+.highcharts-plot-line-label {
+    fill: var(--highcharts-neutral-color-100);
+    font-size: 1em;
+    font-weight: bold;
+}
```
sebastianbochan, line 123: So many duplicates in declarations, you can optimize it.

nasin:
Hi, I think it looks great. 
If possible, can you increase the margin a bit around the component header like in the screenshot?
image

Also, I'm not sure if the "done" color (that darker green color) should be used for the timeline, unless all those tasks are "done." If the timeline is just a timeline, I would use the almost black color like you have in the column chart. And maybe make the "today" line a distinct color from everything else so it stands out more.
image

marvin19 (approved):

TorsteinHonsi (changes requested):
We need to do something with the colors of the area chart. I know we did this for a11y compliance, but according to common sense it is not really usable when the legend color is brown and the area color is yellow. Yes, the line color is brown, but that's barely distinguishable from the other lines.

As far as I know, the a11y aspect is covered since we have text directly on the area, so can we just use normal colors?

Skjermbilde 2023-12-11 kl  12 46 32

karolkolodziej:
@TorsteinHonsi I'm ok with that- it seems more logical 😉

TorsteinHonsi (approved):

PR #20109, opened 2023-11-09
Website/cyber clock updates

nasin:
Hi,

Based on feedback, I made a few design changes to the gauges and incorporated some a11y funcitonality.

Thanks


TorsteinHonsi (approved):
Thanks!

Issue #19794, opened 2023-09-20
Chart changes y axis grid lines min max , after updating a point/serie or redraw


dragoshzava:
#### Expected behaviour
Should not change the min/max of the grid line.

#### Actual behaviour
Changes after update a series or point.

#### Live demo with steps to reproduce
https://jsfiddle.net/Resems/2q61jt97/2/

Open
Click update point

#### Product version
Latest

#### Affected browser(s)
All

#### Attached Video 
[Spline with symbols - JSFiddle - Code Playground (1).webm](https://github.com/highcharts/highcharts/assets/3819194/7e709623-27db-4fc3-bf53-1767241df334)

#### Additional Comment
It happens on both redraw, and both on series.update, and only on specific chart sizes.
We use redraw, after the end user changes the configuration of the chart, and the grid lines changes, which is unexpected result for an end user as the data did not change.

bm64:
Hello @dragoshzava, thanks for reporting the issue.

The issue appears to have started with v11.0.0. It works as expected in v10.3.3: https://jsfiddle.net/BlackLabel/dx6eLb8c/

I can't provide a workaround at the moment. Additionally, series.update and chart.redraw behave differently based on the config. For example, in this case, only series.update alters the axis ticks: https://jsfiddle.net/BlackLabel/woukzn73/

hubertkozik:
_Internal note:_ 
This issue is caused by autorotating the x-axis labels, which cannot be simply calculated before the actual rendering of labels and checking space for them. This issue is **not** fixed by [this Pull Request](https://github.com/highcharts/highcharts/pull/19896).

hubertkozik:
This issue is not regression, in v10.3.3 the styles of the chart were different and the issue was not reproduced.
**Latest release with old styles:** https://jsfiddle.net/BlackLabel/tqzrympx/

TorsteinHonsi:
Agree with Hubert, it is not a regression, and this behaviour is well known. We have traditionally rejected it as not a bug, allthough #19896 fixed most cases. In the API for [tickPixelInterval](https://api.highcharts.com/highcharts/xAxis.tickPixelInterval) we state that it defines the _approximate_ pixel distance between ticks. That's what comes to play here.

[...]

PR #20110, opened 2023-11-10
bugfix/19794-overheatprotector-redraw

hubertkozik:
Added new option `chart.axisLayoutRuns`. It allows more predictable axis tick and label layout at the cost of performance. See #19794.

---
  - https://app.asana.com/0/0/1205674704253590

### DIFF
--- a/ts/Core/Chart/Chart.ts
+++ b/ts/Core/Chart/Chart.ts
@@ -2377,7 +2377,7 @@ class Chart {
             axes = chart.axes,
             colorAxis = chart.colorAxis,
             renderer = chart.renderer,
-            { overheatProtector } = chart.options.chart,
+            axesLayoutRuns = pick(chart.options.chart.axesLayoutRuns, 1),
             renderAxes = (axes: Array<Axis>): void => {
                 axes.forEach((axis): void => {
                     if (axis.visible) {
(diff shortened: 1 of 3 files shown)


ts/Core/Chart/Chart.ts
```diff
@@ -2445,30 +2443,39 @@ class Chart {
...
 
         // If the plot area size has changed significantly, calculate tick
         // positions again
-        const redoHorizontal = tempWidth / chart.plotWidth > 1.1,
-            // Height is more sensitive, use lower threshold
-            redoVertical = tempHeight / chart.plotHeight > 1.05;
-
-        if (redoHorizontal || redoVertical) {
+        let redoHorizontal = tempWidth !== chart.plotWidth,
+            redoVertical = tempHeight !== chart.plotHeight,
```
TorsteinHonsi, line 2456: A notable change here is that previously we accepted a failure margin (> 1.1, > 1.05 respectively). I think we should keep that acceptance margin on these lines, but if the `overheatProtector` is > 1, lines 2476 and 2477 come into play, and those require an exact match.

ts/Core/Chart/ChartDefaults.ts
```diff
@@ -426,6 +426,21 @@ const ChartDefaults: ChartOptions = {
...
+     * intervals. Use this property if the axes change after you redraw the
+     * chart.
+     *
+     * **Note:** Changing that option to higher than the default might decrease
+     * performance significantly, especially with bigger sets of data.
+     *
+     * @type      {number}
+     * @default   1
+     * @since     @next
+     * @apioption chart.overheatProtector
```
TorsteinHonsi, line 441: We need a more descriptive name. Like `chart.axisFittingRuns`? Any better idea?

KacperMadej: Some other ideas: `initFittingMaxIterations`, `axesLayoutLoopLimit`, `maxReflowTries`.

hubertkozik: What do you think about `axesLayoutRuns`?

TorsteinHonsi (approved):

PR #20125, opened 2023-11-13
bugfix/16920-data-module-columntypes

### LINKED ISSUE #16920: Data module removing 0's in fips codes
#### Expected behaviour
The data module does not remove the first 0 in the fips code and the map renders correctly

#### Actual behaviour
The data module removes the first 0 in the fips code and the map does not render any of its points.

An example of a fips code in the Alabama map: 01007

The data module seems to be removing the first 0 in the codes, and so Highmaps cant find the right code to link to.

If you log mapKey and mapMap in the line I've linked below, you'll see that the codes do not match. I assume the data module is doing some custom parsing before it gets to this point and the 0 is getting removed because of it.

https://github.com/highcharts/highcharts/blob/master/ts/Series/Map/MapSeries.ts#L1057

https://jsfiddle.net/vf2yeL5g/

#### Product version
Latest

hubertkozik:
Fixed #16920, formalized and fixed the `data.columnTypes` option.

---
  - https://app.asana.com/0/0/1205897590371460

### DIFF
--- a/samples/unit-tests/data/general/demo.js
+++ b/samples/unit-tests/data/general/demo.js
@@ -244,7 +244,7 @@ QUnit.test('Combination charts and column mapping', function (assert) {
     chart = Highcharts.chart('container', {
         data: {
             csv: 'code;value\n01007;58\n01003;10\n01041;32\n01085;84\n01133;12',
-            columnTypes: ['string', 'float']
+            columnTypes: ['string', 'number']
         }
     });
 
--- a/samples/unit-tests/data/general/demo.js
+++ b/samples/unit-tests/data/general/demo.js
@@ -240,6 +240,20 @@ QUnit.test('Combination charts and column mapping', function (assert) {
         `Name should be mapped correctly from CSV for packed bubble series,
         which is non-cartesian (#19143).`
     );
+
+    chart = Highcharts.chart('container', {
+        data: {
+            csv: 'code;value\n01007;58\n01003;10\n01041;32\n01085;84\n01133;12',
+            columnTypes: ['string', 'float']
+        }
+    });
+
+    assert.strictEqual(
+        chart.xAxis[0].options.type,
+        'category',
+        `After setting first column as string in data.columnTypes, x-axis should
+        be configured as category axis (#16920).`
+    );
 });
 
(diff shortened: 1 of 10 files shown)


KacperMadej (changes requested):
The column-types-map demo reports an error in Utils that is now tracked [here](https://github.com/highcharts/map-collection-dist/issues/134).

samples/unit-tests/data/general/demo.js
```diff
@@ -240,6 +240,20 @@ QUnit.test('Combination charts and column mapping', function (assert) {
         `Name should be mapped correctly from CSV for packed bubble series,
         which is non-cartesian (#19143).`
     );
+
+    chart = Highcharts.chart('container', {
+        data: {
+            csv: 'code;value\n01007;58\n01003;10\n01041;32\n01085;84\n01133;12',
+            columnTypes: ['string', 'float']
```
KacperMadej, line 247: Is `'float'` supported? or should this be `'number'`?

KacperMadej (approved):
One test related comment below, but other than that LGTM 👍

samples/unit-tests/data/general/demo.js
hubertkozik: `float` is working fine, but I think we should support `number` only.

ts/Extensions/Data.ts
```diff
@@ -2450,6 +2457,22 @@ export default Data;
  * @apioption data.columns
  */
 
+/**
+ * An array option that specifies the data type for each column in the series
+ * loaded within the data module.
+ *
+ * Possible values: `"string"`, `"number"`, `"date"`.
```
TorsteinHonsi, line 2464: To help our future Editor (and other GUI generators), we should add:

* A stricter [type](https://github.com/highcharts/highcharts/blob/e6cec31e422a680d9e2a6cc22b10f321d0c90d98/ts/Extensions/Data.ts#L87): `columnTypes?: Array<'string'|'number'|'date'>;
* A `@validvalue` doclet tag, even though this one will soon be redundant as we create the TS-driven API.

TorsteinHonsi (changes requested):
Thanks!

TorsteinHonsi (approved):

PR #20126, opened 2023-11-13
bugfix/20098-plot-options-data-labels-enabled 

### LINKED ISSUE #20098: Two or more `dataLabels` defined for `bullet` series are not displayed
#### Expected behaviour

When setting an array of `dataLabels` for a series, all of the defined data labels should be visible.

_Per API:_ 

> Since v6.2.0, multiple data labels can be applied to each single point by defining them as an array of configs.

#### Actual behaviour

There are no data labels rendered at all in bullet chart.

[https://jsfiddle.net/BlackLabel/9pvr27hs/](https://jsfiddle.net/BlackLabel/9pvr27hs/)

#### Product version
Highcharts v11.2.0

jakubjanuchta:
Fixed #20098, multiple data labels were not rendered when enabled in `plotOptions`.


TorsteinHonsi (changes requested):
Thanks!

We're now repeating ourselves, having exactly the same block of code in the `hasDataLabels` function as we have in `@drawDataLabels`.

https://github.com/highcharts/highcharts/blob/f9b8d9d0dd9282b342571eb51fd46b5d601a36e8/ts/Core/Series/DataLabel.ts#L587-L593

Can you try refactoring this out to a local function of the DataLabels.ts module? Something like `mergedDataLabelOptions(series)` that can be called from both places.

TorsteinHonsi (approved):
Awesome! Now awaiting a small push then CI success.

PR #20132, opened 2023-11-14
dash/aws-demo

sebastianbochan:


ts/DataGrid/DataGridOptions.d.ts
```diff
@@ -170,6 +175,13 @@ export interface CellValue {
     value: DataTable.CellType
 }
 
+/**
+ * Contains events for row
+ */
+export interface DataGridEvents {
+    row?: Record<string, Function>
```
karolkolodziej, line 182: I would add an extra interface for the row events and IMO each event should have its own, public event callback interface.

samples/dashboards/demo/aws-management/demo.css
```diff
@@ -0,0 +1,124 @@
...
+
+#instance-details #instance .highcharts-dashboards-component-content h2 {
+    background:url('data:image/svg+xml,<svg class="w-6 h-6" height="40" width="40" xmlns="http://www.w3.org/2000/svg"><defs><linearGradient x1="0%" y1="100%" x2="100%" y2="0%" id="Arch_Amazon-EC2_32_svg__a"><stop stop-color="%23C8511B" offset="0%"></stop><stop stop-color="%23F90" offset="100%"></stop></linearGradient></defs><g fill="none" fill-rule="evenodd"><path d="M0 0h40v40H0z" fill="url(%23Arch_Amazon-EC2_32_svg__a)"></path><path d="M26.052 27L26 13.948 13 14v13.052L26.052 27zM27 14h2v1h-2v2h2v1h-2v2h2v1h-2v2h2v1h-2v2h2v1h-2v.052a.95.95 0 01-.948.948H26v2h-1v-2h-2v2h-1v-2h-2v2h-1v-2h-2v2h-1v-2h-2v2h-1v-2h-.052a.95.95 0 01-.948-.948V27h-2v-1h2v-2h-2v-1h2v-2h-2v-1h2v-2h-2v-1h2v-2h-2v-1h2v-.052a.95.95 0 01.948-.948H13v-2h1v2h2v-2h1v2h2v-2h1v2h2v-2h1v2h2v-2h1v2h.052a.95.95 0 01.948.948V14zm-6 19H7V19h2v-1H7.062C6.477 18 6 18.477 6 19.062v13.876C6 33.523 6.477 34 7.062 34h13.877c.585 0 1.061-.477 1.061-1.062V31h-1v2zM34 7.062v13.876c0 .585-.476 1.062-1.061 1.062H30v-1h3V7H19v3h-1V7.062C18 6.477 18.477 6 19.062 6h13.877C33.524 6 34 6.477 34 7.062z" fill="%23FFF"></path></g></svg>') 10px 50% no-repeat;
```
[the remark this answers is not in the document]
sebastianbochan: Nope.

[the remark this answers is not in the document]
sebastianbochan: Only by upload images on our server.

sebastianbochan:
At this moment we need to discuss the data structure, if we agree with the suggestion we will use the polling option and remove `setInterval`s from the demo.

nasin:
Looks great!

How about using a neutral color (like neutral-color-10) on the data grid highlight for current/hover?

And can you make the red warning icon a different shape? Then all three icons will be distinct.

image

And then, if possible, rounding the borders of the icons up top might look nice:
image

nasin (approved):

sebastianbochan:
@nasin thank you for the suggestions.

I have two extra questions:
- compatibility color palette with the a11y (and contrast)
- colors in the dark mode

samples/dashboards/demo/aws-management/demo.details
```diff
@@ -0,0 +1,12 @@
+---
+name: AWS Management
```
karolkolodziej, line 2: I'm wondering if we even should use the `aws` keyword here 🤔 
Maybe all should be more general like `Server monitoring dashboard`

samples/dashboards/demo/aws-management/demo.details
```diff
@@ -0,0 +1,12 @@
...
+authors:
+  - Sebastian Bochan
+js_wrap: b
+tags:
+  - Highcharts Dashboards demo
+categories:
+  - General:
+        priority: 12
```
karolkolodziej, line 10: Maybe it should go slightly higher in the hierarchy?

samples/dashboards/demo/aws-management/demo.js
```diff
@@ -0,0 +1,949 @@
+let board = void 0;
+const pollingCheckbox = document.getElementById('enablePolling');
+
+pollingCheckbox.onchange = () => {
```
karolkolodziej, line 4: Let's figure out the data for this demo to implement and test it properly.

samples/dashboards/demo/aws-management/demo.js
```diff
@@ -0,0 +1,949 @@
...
+            UsedGB: 15,
+            FreeGB: 15,
+            MediaGB: 5,
+            RootGB: 10,
+            Documents: 10,
+            Downloads: 5
+        }
+    },
+    Details: [{
+        timestamp: 1640995200000, // JavaScript timestamp for '2022-01-01T00:00:00Z'
```
karolkolodziej, line 70: I'm not sure if we need the comments

karolkolodziej (changes requested):
1. I would use more general naming like mentioned in the comment `Server monitoring dashboard` 
2. Lint is failing
3. Move data from the JS and implement the polling.
4. I would add sync between the charts, they share the same x values- it might be helpful to understand what is going on at a given moment. 
5. The table in dark mode looks strange:
image

samples/dashboards/demo/aws-management/demo.js
sebastianbochan: When the general concept of the demo is accepted, I will speak with Goran and upload the data on our public server.

samples/dashboards/demo/aws-management/demo.js
sebastianbochan: We can simplify it, I agree.

samples/dashboards/demo/aws-management/demo.details
sebastianbochan: On the one hand, it makes sense, however on the other hand it's the real use case.
@oysteinmoseng what do you think?

samples/dashboards/demo/aws-management/demo.details
sebastianbochan: Any suggestion?

goransle: Perhaps next to the Project Management demo as they are somewhat similar? 

Maybe also worth considering if the demos could be split into separate catagories? I see there's several with 'sync' in the title that perhaps could go into a "Synchonization" (/Synchronisation?) category.

karolkolodziej: good point! added to an internal discussion!

sebastianbochan: Just set the priority as 11, until end of subcategories discussion

TorsteinHonsi:
The design and layout look great so far!

When selecting different instances, there is an unwanted redraw artifact in all the cartesian charts. For example, the "Disk usage" chart very quickly jumps to a greater height, then back. This causes a flickering effect. It looks to me that the charts are first rendered into a higher container, then get redrawn to their intended sizes.

Furthermore, when trying to follow this lead, I see that `Highcharts.charts` initially contains 7 charts. After switching instances, there are 21 charts. Switching again, the number increases to 35. So it seems like for every switching rows in the Instances, another 14 charts are created while the existing ones are not destroyed.

I would expect one of two things when switching the Instances:
* The existing charts within each component were updated by `chart.update`, which would also include animation of the series. Or is that to risky, considering the config may change too much?
* Otherwise, the charts should be replaced cleanly. Highcharts automatically takes care of destroying existing charts when rendering to the same container, see https://jsfiddle.net/highcharts/LLExL/14550/

Generally, if we know that the data structure is the same, I think we should prefer `chart.update` internally.

sebastianbochan:
@TorsteinHonsi The issue with Highcharts was more complex than I expected. It turned out that we removed only containers instead of components, which was even more complicated in nested layouts. Thanks for pointing me out.

TorsteinHonsi:
@sebastianbochan Thanks!

Are there any plans to refine the animation UX on dynamic data? When enabling polling, I expected the values of the charts to change smoothly like [they do with the data module](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/data/livedata-columns/). [Updating line demo here](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/data/livedata-rows/). This seems to work correctly with the KPI gauges, but the four cartesian charts brutally redraw to their new values. There also seems to be some remnant bars in the bar chart.

samples/dashboards/demo/cloud-monitoring/demo.css
```diff
@@ -0,0 +1,144 @@
+@import url("https://code.highcharts.com/css/highcharts.css");
+@import url("https://code.highcharts.com/css/datagrid.css");
```
karolkolodziej, line 2: Wrong path

samples/dashboards/demo/cloud-monitoring/demo.html
```diff
@@ -0,0 +1,26 @@
...
+  <div class="cloud-monitoring-data-controls">
+    <label for="enablePolling">
+      <input type="checkbox" id="enablePolling" name="enablePolling" />
+      <span>Enable data polling</span>
+    </label>
+  </div>
+  <div id="container"></div>
+</div>
+<p class="highcharts-description">
+  The demo shows how you can visualize AWS server data on a dashboard.<br/>
```
karolkolodziej, line 19: As far as I remember we shouldn't use `<br/>` here

karolkolodziej (changes requested):
Good job! 

Altho I have a few things:
1. From time to time when the pooling enabled, I see multiple points on the bar chart
image
2. Background and the data grid rows color in the dark mode are off.
3. I agree with Torstein that removing and creating the chart all over again isn't ideal.
4. Errors in CI- linting.

samples/dashboards/demo/cloud-monitoring/demo.html
sebastianbochan: In the minimal demo we use it as well.

sebastianbochan:
@karolkolodziej 
1. The polling is not perfect and has some limitations It's a more complex issue, so I think that it's better to disable polling on it. Especially since it's not important in this demo.
2. Fixed
3. At this moment that's the way how the polling option (in connectors)  works. We can use a regular async fetching and setIntervals, however, I would prefer to use the build-in option.
5. Fixed

karolkolodziej (changes requested):
I'm ok with disabling pooling for now.

But still, there are some errors in the CI tests.

sebastianbochan:
@karolkolodziej I asked @goransle about it, just waiting for the feedback.

jomar-honsi (commented):
Very nice demo, certainly future demos will borrow a let from it. I only have on observation (rendered with mobile view): 
image

sebastianbochan:
Fixed karma tests, however, the one test left that is broken on master too.

karolkolodziej (approved):

samples/dashboards/demo/cloud-monitoring/demo.css
```diff
@@ -0,0 +1,144 @@
...
+
+:root,
+.highcharts-light {
+    /* Colors for data series and points */
+    --highcharts-color-0: #33a29d;
+    --highcharts-color-2: #fe9d00;
+
+    /* Extra colors */
+    --bck-gray: #f1f5f9;
+    --dashboard-bck-gray: #e4eaf1;
```
Vazonik, line 23: 1. Do we need two variables here for almost the same color?
2. IMO it is not good practice to name variables in CSS by specifying a color in its name. Better to use something like `--bcg-color`.
3. I think that this color should have a darker counterpart in dark mode (e.g. `#252525`). A light background and dark components create too much contrast.

samples/dashboards/demo/cloud-monitoring/demo.css
```diff
@@ -0,0 +1,144 @@
...
+
+#instances-table .highcharts-datagrid-container {
+    border: none;
+    font-weight: 100;
+}
+
+#instances-table .highcharts-datagrid-column-header,
+#instances-table .highcharts-datagrid-row {
+    background-color: var(--highcharts-neutral-color-0);
+}
```
Vazonik, line 88: It doesn't look good in dark mode, the data grid background is completely black.

Vazonik (approved):
Good job! A few comments about the appearance in dark mode and styling in general. But I approve so as not to block you further.

jomar-honsi (approved):

jomar-honsi:
> Fixed karma tests, however, the one test left that is broken on master too.
> 
> @jomar-honsi we can use `word-break: break-all` in CSS, however, I'm not a big fan of it.

@sebastianbochan  Is it possible to shrink the font for mobile view? (I also have this problem in the live-weather demo).

sebastianbochan:
@jomar-honsi I think that it's a more complex issue and we should rethink it during the DataGrid new concept. We can add a feature that disables some columns or uses CSS styles etc.

At this moment we can only reduce the `font-size` or use `word-wrap` which are not perfect as well.

TorsteinHonsi:
@sebastianbochan @karolkolodziej 
> 1. The polling is not perfect and has some limitations It's a more complex issue, so I think that it's better to disable polling on it. Especially since it's not important in this demo.

Actually I intended on asking that we enable polling by default... I think it makes the demo more interesting. But then of course the animation must be smooth:

> 3. At this moment that's the way how the polling option (in connectors) works. 

Can we fix it? Animation is a crucial part of the UX experience, along with the graphical design, and we have spent considerable resources on getting animation on data updates to run smoothly in Highcharts. Dashboards is the perfect use case for this. If we don't adhere to that in Dashboards, we are going against our own recommendations. Ref [Column series](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/data/livedata-columns/) and [line series with shifting time axis](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/data/livedata-rows/).

sebastianbochan:
1. Due to accessibility it's triggered by the checkbox.
2. I would ask @bre1470 about it, because only what I can do it's using polling or `setInterval` / `async`.

stitot (commented):
Looks great. Just some quick comments on UX:

1. Possible with smoother transitions when polling is enabled? Believe I've seen this on individual charts? Probably a biggie.
2. ID column in instances should have a no-wrap or something to avoid line-break.
3. Could the Disk usage bar chart have labels instead of legend? Believe it would be easier to read.
4. CPU utilization doesn't need legend if only one CPU.  Remove clutter.
5. When hovering over Instances table the cursor changes to pointer, but nothing to click/click doesn't work. Perhaps move this table closer to the top, since it is a navigational element?

TorsteinHonsi:
Regarding the animation issue, it is also a general performance issue. I isolated it and created a separate ticket for it here: https://github.com/highcharts/highcharts/issues/20483 . I think we should get it fixed before publishing the cloud monitoring demo.

samples/dashboards/demo/cloud-monitoring/demo.html
```diff
@@ -0,0 +1,26 @@
...
+<script src="https://code.highcharts.com/modules/accessibility.js"></script>
+<script src="https://code.highcharts.com/highcharts-more.js"></script>
+<script src="https://code.highcharts.com/modules/solid-gauge.js"></script>
+<script src="https://code.highcharts.com/dashboards/datagrid.js"></script>
+<script src="https://code.highcharts.com/dashboards/modules/dashboards-plugin.js"></script>
+
+<div class="cloud-monitoring-dashboard-container">
+  <div class="cloud-monitoring-data-controls">
+    <label for="enablePolling">
+      <input type="checkbox" id="enablePolling" name="enablePolling" />
```
TorsteinHonsi, line 12: The polling feature works so well now, and adds a wow factor to this demo. It should be enabled by default. It's too good to hide behind an opt-in. At the same time, we could consider moving this checkbox to a less prominent place.

samples/dashboards/demo/cloud-monitoring/demo.js
```diff
@@ -0,0 +1,724 @@
...
+            connector: {
+                id: 'instanceDetails'
+            },
+            columnAssignment: {
+                index: 'x',
+                MediaGB: 'y',
+                RootGB: 'y',
+                Documents: 'y',
+                Downloads: 'y'
+            },
```
TorsteinHonsi, line 338: Like [Stian commented](https://github.com/highcharts/highcharts/pull/20132#pullrequestreview-1831791062), this chart would be better with labels instead of a legend. Color matching alone is not a11y compliant - it's hard for the vision impaired, and adds another layer of cognitive load for the rest of us. We could use a single series with categories instead: https://jsfiddle.net/highcharts/y0rbwdpa/

PS: We could also consider including data labels, that would further improve accessibility.

stitot (commented):
Looks very nice now! Three, hopefully small, things:

1. Possible to move Instances further up? It's a navigational element, and at least on mobile it's confusing that things change both above and below that element. 

2. Possible to keep data polling going when you change Instance?

3. There's something fishy with CPU utilization in mobile view. Sometimes it's empty. Able to reproduce by changing Instance, but not always. See image. Only happens in mobile view.

Skjermbilde 2024-02-08 kl  08 23 34

samples/dashboards/demo/cloud-monitoring/demo.js
```diff
@@ -0,0 +1,724 @@
...
+                id: 'charts'
+            },
+            columnAssignment: {
+                timestamp: 'x',
+                cpuUtilization: 'y'
+            },
+            sync: {
+                highlight: true
+            },
+            chartOptions: {
```
TorsteinHonsi, line 391: As Stian commented, this chart doesn't need a legend as it has only one series.

samples/dashboards/demo/cloud-monitoring/demo.js
```diff
@@ -0,0 +1,724 @@
...
+                    }
+
+                },
+                events: {
+                    row: {
+                        click: async function (e) {
+                            if (pollingCheckbox.checked) {
+                                // stop polling when is enabled
+                                await pollingCheckbox.click();
+                            }
```
TorsteinHonsi, line 691: Also rephrasing Stian. It feels like a bug that the polling stops when I switch instances. Can we keep it running? If that causes problems, maybe just a temporary stop then restart after `setupDashboard` has run?

samples/dashboards/demo/cloud-monitoring/demo.js
```diff
@@ -0,0 +1,724 @@
...
+            cell: 'network-opt',
+            type: 'Highcharts',
+            title: 'Network (bytes)',
+            connector: {
+                id: 'charts'
+            },
+            columnAssignment: {
+                timestamp: 'x',
+                networkIn: 'y',
+                networkOut: 'y'
```
TorsteinHonsi, line 581: In both the _Network (bytes)_ chart and the _Disk operations_ chart, the column names are used for `series.name`. Is it possible to add more human readable series names (like "Network In", "Read operations") instead of the camelCase column headers? 

This may be a bigger issue though - according to the docs, `columnAssignment` matches series by their `name`. It would be preferable if they were matched by `series.id` instead, in consistence with other Highcharts matching (like `series.xAxis`, `series.linkedTo` etc). In general, the `name` property is intended for human readers, while the `id` is intended for computing. If this is not currently possible, it's not a blocker for this demo, but we should add an enhancement request for it.

samples/dashboards/demo/cloud-monitoring/demo.html
TorsteinHonsi: I agree with Karol. Use nested paragraphs instead. See https://github.com/highcharts/highcharts/blob/master/repo-guidelines.md#1-avoid-line-breaks-without-paragraphs .

samples/dashboards/demo/cloud-monitoring/demo.js
```diff
@@ -0,0 +1,724 @@
...
+                                width: '100%'
+                            },
+                            large: {
+                                width: '50%'
+                            }
+                        }
+                    }]
+                }, {
+                    cells: [{
+                        id: 'instances-table',
```
TorsteinHonsi, line 226: The _Instances_ data grid is apparently 400px high in mobile view, causing a large padding below even when there are no other cells on the same row. Do we have a simple fix for that now? Or wait until v2 maybe?

TorsteinHonsi (changes requested):
Awesome, the polling looks great now! See the inline comments - some reiterations of previous comments, some new findings.

samples/dashboards/demo/cloud-monitoring/demo.js
sebastianbochan: Until v2, let's leave it like it is. In the next version, we can use media queries.

samples/dashboards/demo/cloud-monitoring/demo.js
sebastianbochan: You are right, it should be matched by ID.

Meanwhile I used a simple snippet in the labelFormatter.

stitot (commented):
Looks very good now, but selection of other instances have stopped working? Nothing happens when I click in the table. Other than that it's good to go!

sebastianbochan:
@stitot which browser?
In the chrome (macosx) it works like a charm.

stitot:
@sebastianbochan : hm, today it works perfect. Could be something with the demo-viewer I messed up yesterday.

TorsteinHonsi (approved):
Well done, works like a charm! This was quite a long review process, but we also discovered and fixed a couple of bugs along the way, which is a positive side effect of creating advanced demos.

sebastianbochan:
@TorsteinHonsi I appreciate it :)

goransle (approved):
Looking good!

PR #20138, opened 2023-11-15
Fixed some a11y issues

nasin:
Hi, I disabled the animated elements entirely for people with reduce-motion preference and fixed some responsiveness issues.


TorsteinHonsi (approved):
Thanks!

PR #20144, opened 2023-11-16
tools/dist-ignore-files

goransle:


tools/gulptasks/dist-release.js
```diff
@@ -307,6 +321,11 @@ async function release() {
     await removeFilesInFolder(pathToDistRepo, keepFiles);
     log.message('Successfully removed content of ' + pathToDistRepo);
 
+    // Rerun compile with sourceMaps disabled
+    await scriptsCompile(void 0, {
+        skipSourceMaps: true
+    });
+
```
bre1470, line 328: Why is a recompile necessary?

goransle: I assumed we still want the sourceMaps on code.highcharts.com. Otherwise we can just set it false in swc

bre1470:
Looks good! 👍

TorsteinHonsi (approved):
Thanks!

PR #20148, opened 2023-11-16
bugfix/19845-gantt-axis-alignment

### LINKED ISSUE #19845: Gantt chart display issue
#### Expected behaviour
Top left corner border line should align properly.
Last row label is vertically centered similar to rest of the labels.

#### Actual behaviour
Top left corner border line does not align properly.
Last row label is not vertically centered.

https://jsfiddle.net/0kope62s/

#### Product version
highcharts/modules/gantt version 11.1.0

#### Affected browser(s)
All

hubertkozik:
Fixed #19845, grid axes had open-ended cells in some cases.

---
  - https://app.asana.com/0/0/1205674704253584


hubertkozik:
After a few tries and code changes: there is only one visual difference, which is expected - there shouldn't be any shift of label cell in that demo.

KacperMadej (approved):

TorsteinHonsi (commented):
Thanks!

Can you give a little explanation of what is going on?

In the [OP demo](https://jsfiddle.net/0kope62s/), the minimum of the x-axis is set to January 31, 00:00. Which means that by moving the tick two pixels to the left, we actually include the last day of January, making it look like it were part of February. 

In the one visual diff, at least on my machine (Central European Time), the situation appears to be similar. The `xAxis.min` is on the previous Sunday at 17:00 hours. So the two pixels on the left actually belong to the week before, and it's misleading to move the tick to the end of the axis.

In https://jsfiddle.net/highcharts/hp9xczn0/ I have added a slider to show what happens when we add or remove 6 hours from the min and max. In both ends we see the same issue - the final line doesn't show up if the first or last tick is very close to the end.

Judging from the comments in the code, I would suspect the problem lies somewhere in this function:

https://github.com/highcharts/highcharts/blob/fbee6f0acf7b2af5dc60390f105d631072f92b32/ts/Core/Axis/GridAxis.ts#L517-L524

hubertkozik:
I have overlooked some functions in the `GridAxis.ts` file, the previous fix was bad. I added new changes now, which look much better.

ts/Core/Axis/GridAxis.ts
```diff
@@ -1246,38 +1246,48 @@ function onTickLabelFormat(ctx: AxisLabelFormatterContextObject): void {
...
+        firstPos = tickPositions[0],
+        secondPos = tickPositions[1],
+        lastPos = tickPositions[tickPositions.length - 1],
+        linkedMin = axis.linkedParent && axis.linkedParent.min,
+        linkedMax = axis.linkedParent && axis.linkedParent.max,
+        min = linkedMin || axis.min,
+        max = linkedMax || axis.max,
+        tickInterval = axis.tickInterval,
+        endMoreThanMin = (
+            min &&
```
TorsteinHonsi, line 1263: If `min` is 0, `endMoreThanMin` will end up as falsy. Not very likely that `min` is on January 1st 1970, but nonetheless. Use `isNumber(min)` instead.

TorsteinHonsi (changes requested):
Thanks! The visual diffs now resolve some of the cases that I always found strange, but weren't sure were a bug.

Since we've fixed this case for the low end of the axes, we should also apply the same logic to the high end. I consider those the same issue:

Check out https://jsfiddle.net/highcharts/hp9xczn0/1/
* For slider values 1 and 2, the last week box is open ended.
* For slider values from 13 up until 24, the January box is open ended.

ts/Core/Axis/GridAxis.ts
TorsteinHonsi: This also applies to the two more checks for `min` and `max` down below.

TorsteinHonsi (changes requested):
Awesome! I found one more glitch that I think we should address before we leave this subject. In the `gantt/demo/inverted` sample, the week cells have open ends both at min and max.

TorsteinHonsi (approved):

PR #20150, opened 2023-11-16
bugfix/20086-mapbubble-wrong-hover-point

### LINKED ISSUE #20086: Mapbubble series wrong point on hover
#### Behaviour
After recent changes in bubble series, due to the lack of clientX values for the mapbubble points, point search on hover fails to deliver the correct points.

https://jsfiddle.net/BlackLabel/sp1w7qyx/1/

#### Product version
Highcharts Maps v11.2.0

#### Workaround
https://jsfiddle.net/BlackLabel/sp1w7qyx/

```js
Highcharts.seriesTypes.mapbubble.prototype.kdAxisArray = [
	'plotX', 'plotY'
];

Highcharts.seriesTypes.mapbubble.prototype.searchPoint = function (e, compareX) {
                return this.searchKDTree({
                    plotX: e.chartX - this.chart.plotLeft,
                    plotY: e.chartY - this.chart.plotTop
                }, compareX, e);
            };
```

madepiet:
Fixed #20086, a bug where hovering over `mapbubble` series points would fail to show the correct point.

### DIFF
--- a/samples/unit-tests/maps/mapbubble/demo.js
+++ b/samples/unit-tests/maps/mapbubble/demo.js
@@ -42,6 +42,38 @@ QUnit.test('MapBubble', function (assert) {
         }]
     });
 
+    const done = assert.async();
+
+    setTimeout(() => {
+        const point = chart.series[1].points[0],
+            { plotX, plotY } = point,
+            chartPosition = Highcharts.offset(chart.container),
+            chartX = chart.plotLeft + plotX + chartPosition.left,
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (changes requested):
Thanks, the solution looks good!

Can you provide a unit test too? See if we can hook this up to an existing test so we don't add another chart instance. Use the test-controller to simulate mouse move close to the bubble.

TorsteinHonsi (approved):

PR #20152, opened 2023-11-17
bugfix/17545-funnel-data-labels-inside 

### LINKED ISSUE #17545: Funnel/Pyramid Chart - Data Labels appear inside indicator even when inside is set to false
#### Expected behaviour
Setting dataLabels > inside : false should display data labels outside the Indicator on full screen view

#### Actual behaviour
The data labels appear inside Funnel/Pyramid indicator when viewed in 'View in full screen'. This also happens at time on when loading the Indicator itself.

https://jsfiddle.net/obpLf8js/1/
Funnel_Fullscreen_InsideFalse_DataLabelsAppearInside

#### Product version
Highcharts JS v10.1.0 (2022-04-29)

#### Affected browser(s)
Chrome. Edge

Note: This does not happen all the time but sporadically happens. I am not able to tell the pattern as it is very random. Setting  allowOverlap: false resolves this issue but does not display the lower value data labels on load unless mouse hover. The missing data values affect when exported as image/pdf

jakubjanuchta:
Fixed #17545, misplaced funnel/pyramid dataLabels after legend item click when `dataLabels.inside` set to false.

### DIFF
--- a/ts/Extensions/OverlappingDataLabels.ts
+++ b/ts/Extensions/OverlappingDataLabels.ts
@@ -263,7 +263,7 @@ function hideOrShow(label: SVGElement, chart: Chart): boolean {
 
             // Make sure the label is completely hidden to avoid catching clicks
             // (#4362)
-            if (label.alignAttr && label.placed) { // data labels
+            if (label.alignAttr && label.placed) { // Data labels
                 label[
                     newOpacity ? 'removeClass' : 'addClass'
                 ]('highcharts-data-label-hidden');
@@ -284,7 +284,7 @@ function hideOrShow(label: SVGElement, chart: Chart): boolean {
                     complete
                 );
                 fireEvent(chart, 'afterHideOverlappingLabel');
-            } else { // other labels, tick labels
+            } else { // Other labels, tick labels
                 label.attr({
                     opacity: newOpacity
                 });
--- a/ts/Extensions/OverlappingDataLabels.ts
+++ b/ts/Extensions/OverlappingDataLabels.ts
@@ -278,9 +278,11 @@ function hideOrShow(label: SVGElement, chart: Chart): boolean {
                 isLabelAffected = true;
 
                 // Animate or set the opacity
-                label.alignAttr.opacity = newOpacity;
+                const labelAttr = label.dataLabelPosition?.posAttribs ||
+                    label.alignAttr;
+                labelAttr.opacity = newOpacity;
                 label[label.isOld ? 'animate' : 'attr'](
-                    label.alignAttr,
+                    labelAttr,
                     null as any,
                     complete
                 );
--- a/ts/Extensions/OverlappingDataLabels.ts
+++ b/ts/Extensions/OverlappingDataLabels.ts
@@ -278,11 +278,9 @@ function hideOrShow(label: SVGElement, chart: Chart): boolean {
                 isLabelAffected = true;
 
                 // Animate or set the opacity
-                const labelAttr = label.dataLabelPosition?.posAttribs ||
-                    label.alignAttr;
-                labelAttr.opacity = newOpacity;
+                label.alignAttr.opacity = newOpacity;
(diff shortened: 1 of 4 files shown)


ts/Extensions/OverlappingDataLabels.ts
```diff
@@ -278,9 +278,11 @@ function hideOrShow(label: SVGElement, chart: Chart): boolean {
                 isLabelAffected = true;
 
                 // Animate or set the opacity
-                label.alignAttr.opacity = newOpacity;
+                const labelAttr = label.dataLabelPosition?.posAttribs ||
+                    label.alignAttr;
+                labelAttr.opacity = newOpacity;
                 label[label.isOld ? 'animate' : 'attr'](
-                    label.alignAttr,
+                    labelAttr,
```
TorsteinHonsi, line 285: Is a change in the core needed? Do we know why `label.alignAttr` is not updated for the funnel? Is it something specific to the funnel series, or can the same situation appear in a pie series?

TorsteinHonsi (commented):
Thanks! In addtion to my inline question, see https://github.com/highcharts/highcharts/tree/master/changelog#writing-content-for-the-changelog, point 8 about using code backticks.

ts/Extensions/OverlappingDataLabels.ts
jakubjanuchta: Thanks for the backticks suggestions, I changed the PR description. 

Funnel utilizes Pie drawDataLabels only when inside: false. 
https://github.com/highcharts/highcharts/blob/12a5a01b32a09882151b806e86021fe269b71019/ts/Series/Funnel/FunnelSeries.ts#L248-L254
Pie dataLabels are positioned based on `dataLabelPosition.posAttribs` and the `alignAttr` for pie label remains undefined (which means that it's position is never updated in `OverlappingDataLabels`) because alignAttr is only updated by `SVGElement.align` in `alignDataLabel` which is not utilized by pie labels but it is by funnel with `dataLabels.inside` = false and that's why the incorrect position is applied in `OverlappingDataLabels`.
https://github.com/highcharts/highcharts/blob/12a5a01b32a09882151b806e86021fe269b71019/ts/Series/Funnel/FunnelSeries.ts#L216-L226

jakubjanuchta: Thanks for the backticks suggestions @TorsteinHonsi , I changed the PR description. 

Funnel utilizes Pie drawDataLabels only when inside: false. 
https://github.com/highcharts/highcharts/blob/12a5a01b32a09882151b806e86021fe269b71019/ts/Series/Funnel/FunnelSeries.ts#L248-L254
Pie dataLabels are positioned based on `dataLabelPosition.posAttribs` and the `alignAttr` for pie label remains undefined (which means that it's position is never updated in `OverlappingDataLabels`) because alignAttr is only updated by `SVGElement.align` in `alignDataLabel` which is not utilized by pie labels but it is by funnel with `dataLabels.inside` = false and that's why the incorrect position is applied in `OverlappingDataLabels`.
https://github.com/highcharts/highcharts/blob/12a5a01b32a09882151b806e86021fe269b71019/ts/Series/Funnel/FunnelSeries.ts#L216-L226

TorsteinHonsi: So can we override this somehow from the funnel series module so we don't need to make an exception in the overlapping labels module?

jakubjanuchta: Sure! Although, I haven't found a right place to overwrite the `alignAttr` in the funnel module after `posAttribs` is calculated and before utilizing `hideOrShow` where the position is updated so I had to update it in PieDataLabels.

TorsteinHonsi: I found some inconsistencies and moved the fix to the funnel series (and simplified overlappping labels along the way). Thanks!

TorsteinHonsi (approved):

PR #20159, opened 2023-11-18
feature/standalone-navigator

bm64:
Added a Standalone Navigator component, allowing for range synchronization and manipulation in bound charts and axes.


ts/Stock/Navigator/NavigatorComposition.ts
pawelfus, line 1: In this file we have many unnecessary imports and variables, could you double-check it?

ts/Stock/Navigator/StandaloneNavigator.ts
```diff
@@ -0,0 +1,330 @@
...
+} = U;
+
+declare module '../../Core/GlobalsLike.d.ts' {
+    interface GlobalsLike {
+        navigators: Array<StandaloneNavigator>;
+    }
+}
+
+type StandaloneNavigatorOptions = {
+    navigator: NavigatorOptions;
```
pawelfus, line 41: I'm not sure if we can reuse the whole interface. Some options (`baseSeries`, `enabled` and `opposite`) shouldn't be available in Standalone. How about split into two? Like this:

```ts
interface BaseNavO {
  ...
};

// Current, in chart options:
interface NavigatorOptions extends BaseNavO {
  enabled: ...
  baseSeries: ...
};

// New interface for standalone nav:
interface StandaloneNavigatorOptions extends BaseNavO {
  ...
};
```

pawelfus: In general, the structure of options seems to be a bit off: `Highcharts.navigator('container', options)` -> options should be options for the navigator, not the whole ChartOptions

ts/Stock/Navigator/StandaloneNavigator.ts
```diff
@@ -0,0 +1,330 @@
...
+        });
+
+        // Set extremes to match the navigator's extremes
+        axis.setExtremes(min, max);
+    }
+
+    /**
+     * Unbinds an axis from the standalone navigator.
+     *
+     * @param {Chart | Axis | undefined} axisOrChart
```
pawelfus, line 170: We need to describe all three params. Eg we need to add info `undefined` removes all charts and axes from nav

ts/Stock/Navigator/StandaloneNavigator.ts
pawelfus, line 1: Missing `Navigator.redraw()` Method

pawelfus (changes requested):
Looks very promising!

In addition to inline comments:
- missing doc page
- official demo
- a couple of samples in API
- is standalone module loaded to karma?

ts/Stock/Navigator/StandaloneNavigator.ts
```diff
@@ -0,0 +1,330 @@
...
+            axisOrChart.xAxis[0] :
+            axisOrChart;
+
+        if (!(axis instanceof Axis)) {
+            return;
+        }
+
+        const { min, max } = this.navigator.xAxis;
+
+        this.boundAxes.push(axis);
```
pawelfus, line 154: We should bind to `setRange` event here. Instead of creating and maintaining `boundAxes` array. The same in `unbind` - just remove `setRange` event added here.

Then in `initNavigator()` you can remove the artificial event added to the chart's empty xAxis `setExtremes`

pawelfus (changes requested):
One more thing:

ts/Stock/Navigator/StandaloneNavigator.ts
pawellysy: Imo this solution is easier, because when you are keeping the reference to the correct object. The unbind is easier, because you are just deleting the correct object from the array. When unbinding the chart/axis, and passing it as an argument, how to find the correct event to remove, without mutating the bound axis?

ts/masters/modules/navigator.src.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * @license Highcharts JS v@product.version@ (@product.date@)
+ * @module highcharts/modules/navigator
+ * @requires highcharts
+ *
+ * Standalone navigator module
+ *
+ * (c) 2009-2021 Torstein Honsi
```
pawellysy, line 8:
```suggestion
 * (c) 2009-2024 Mateusz Bernacik
```

pawellysy (changes requested):
Great job! Some minor things to improve

samples/stock/standalone-navigator/getrange/demo.js
```diff
@@ -0,0 +1,42 @@
...
+            100,
+            100
+        )
+            .attr({
+                fill: '#FCFFC5',
+                zIndex: 8
+            })
+            .add();
+
+        // e.target.disabled = true;
```
kamil-musialowski, line 40: Do we need this one here or it's a minor development leftover?

If it should stay like that in the demo, then I would suggest adding a short comment explaining why it's here

ts/Stock/Navigator/Navigator.ts
```diff
@@ -2036,6 +2036,19 @@ class Navigator {
...
+            ),
+            addEvent(
+                Navigator,
+                'setRange',
+                function (this: Navigator, e: SetRangeEvent):void {
+                    this.chart.xAxis[0].setExtremes(
+                        e.min,
+                        e.max,
+                        e.redraw,
+                        e.animation,
```
kamil-musialowski, line 2048: If the `e.animation` is undefined, that will result in triggering the animation on setExtremes, which does not look good and decreases the performance. (This does not happen in normal navigator).

kamil-musialowski (changes requested):
Fantastic work, just a minor stuff!

ts/Stock/Navigator/StandaloneNavigator.ts
pawelfus: With `addEvent` and `removeEvent` you don't need a bound axis at all. I was thinking about something like this:

```js
bind(axis) {
  addEvent(this, 'setRange', axis.setExtremes);
}

unbind(axis) {
  removeEvent(this, 'setRange', axis.setExtremes);
}
```

But it seems I made a mistake in the draft, where I assumed compatible interfaces, but the Event interface is an object, not an Array. We could extend Axis prototype with an adapter:
```js
Axis.proto.setRange = function () {
  return this.setExtremes(...args);
}
```

And use it like I said above. Then we can get rid of this part of code: https://github.com/highcharts/highcharts/pull/20159/files#diff-900ccb90f094a9e5757b539c60572a4d491791cadcd1c824d1a56f4ac45a38b9R279-R291

What do you think?

pawellysy: Hm, I didn't think of implementing it this way. Indeed it seems much simpler. Let's go for that

docs/stock/standalone-navigator.md
```diff
@@ -0,0 +1,20 @@
...
+
+The `bind` method binds chart or axis to the Standalone Navigator. Once a chart is bound, any range adjustments made using the navigator are automatically reflected in the bound chart. See the [bind method documentation](https://api.highcharts.com/class-reference/Highcharts.StandaloneNavigator.bind) for more details.
+
+The `unbind` method is used to detach a chart or axis from the Standalone Navigator. This method can unbind either a single chart or all charts bound to the navigator simultaneously. Detaching charts from the navigator stops it from influencing their range. See the [unbind method documentation](https://api.highcharts.com/class-reference/Highcharts.StandaloneNavigator.unbind) for additional information.
+
+## Managing Navigator Range
+
+The `setRange` method allows programmatic control of the navigator's range. This method not only updates the navigator's range but also synchronizes this change with all bound charts. Additionally, it triggers the `setRange` event.
+
+The `getRange` method retrieves the current range of the Standalone Navigator, enabling users to view or use the navigator's existing range settings.
```
pawelfus, line 16: I would add a link to docs to `getRange` too

docs/stock/standalone-navigator.md
```diff
@@ -0,0 +1,20 @@
...
+
+## Binding Charts to the Navigator
+
+The `bind` method binds chart or axis to the Standalone Navigator. Once a chart is bound, any range adjustments made using the navigator are automatically reflected in the bound chart. See the [bind method documentation](https://api.highcharts.com/class-reference/Highcharts.StandaloneNavigator.bind) for more details.
+
+The `unbind` method is used to detach a chart or axis from the Standalone Navigator. This method can unbind either a single chart or all charts bound to the navigator simultaneously. Detaching charts from the navigator stops it from influencing their range. See the [unbind method documentation](https://api.highcharts.com/class-reference/Highcharts.StandaloneNavigator.unbind) for additional information.
+
+## Managing Navigator Range
+
+The `setRange` method allows programmatic control of the navigator's range. This method not only updates the navigator's range but also synchronizes this change with all bound charts. Additionally, it triggers the `setRange` event.
```
pawelfus, line 14: I would add a link to docs to `setRange` too

docs/stock/standalone-navigator.md
```diff
@@ -0,0 +1,20 @@
+Standalone Navigator
+====================
+
+The Standalone Navigator is a versatile Highcharts component, designed to synchronize and manipulate the viewing range across multiple charts. This tool enables users to easily control the range of one or more charts by interacting with a single navigator interface. It's particularly useful in dashboards and data analysis applications where viewing different data sets in parallel is essential.
+
+## Binding Charts to the Navigator
```
pawelfus, line 6: I suggest adding one more paragraph here: how to create the navigator. And information which files should be loaded.

ts/Stock/Navigator/Navigator.ts
```diff
@@ -84,6 +83,18 @@ function numExt(
     }
 }
 
+interface SetRangeEvent {
+    min: number;
+    max: number;
+    redraw: boolean;
+    animation?: boolean;
+    options: {
```
pawelfus, line 91: `options` name is too generic, I think we need a bit more descriptive name here

pawelfus (changes requested):
Great job! We are one step closer to the final solution 

Bugs:
- destroying chart throws errors upon navigating: https://jsfiddle.net/BlackLabel/4zbrjs8m/ 
- the same when removing xAxis: https://jsfiddle.net/BlackLabel/4zbrjs8m/1/ 
- `update()` seems to be broken: https://jsfiddle.net/BlackLabel/4zbrjs8m/2/ 

Additional samples:
- `navigator.update()` would be great
- a sample with some options configured in nav also would be great (eg styling masks, scrollbar and series) - it will serve as a test additionally

ts/Stock/Navigator/StandaloneNavigator.ts
```diff
@@ -0,0 +1,401 @@
...
+     *
+     * */
+
+    constructor(
+        element: (string | globalThis.HTMLElement),
+        userOptions: StandaloneNavigatorOptions
+    ) {
+        this.userOptions = userOptions;
+        this.chartOptions = merge(
+            (G as any).getOptions(),
```
pawelfus, line 116: Hm.. could you try using Defaults instead of Globals? Like this:
```
import D from '../Defaults.js';
const { getOptions } = D;
```

Perhaps we can get rid of this as any casting

samples/stock/standalone-navigator/update/demo.js
pawelfus, line 1: Could you also bind nav and chart? 😬

docs/stock/standalone-navigator.md
```diff
@@ -0,0 +1,50 @@
...
+## Creating the Standalone Navigator
+
+To initialize the Standalone Navigator, use the `navigator` constructor with the container reference and configuration options:
+
+```javascript
+Highcharts.navigator('container', {
+    // Standalone Navigator configuration options
+});
+```
+The configuration options for the Standalone Navigator are the same as those for a chart navigator, and can be found in the [API documentation](https://api.highcharts.com/highstock/navigation).
```
pawelfus, line 15: wrong url

pawelfus (changes requested):
Just some minor polishing, looks really good!

I was playing with it, when realized bind does work only one way: extremes from navigator are set on charts. However panning and zooming on charts don't update the navigator, demo: http://localhost:3030/samples/#view/stock/demo/standalone-navigator

I know it wasn't part of the initial draft, but how hard would it be to add it?

What do you think @kamil-musialowski @DJTechnoo @pawellysy ?

kamil-musialowski:
> Just some minor polishing, looks really good!
> 
> I was playing with it, when realized bind does work only one way: extremes from navigator are set on charts. However panning and zooming on charts don't update the navigator, demo: http://localhost:3030/samples/#view/stock/demo/standalone-navigator
> 
> I know it wasn't part of the initial draft, but how hard would it be to add it?
> 
> What do you think @kamil-musialowski @DJTechnoo @pawellysy ?

If it is not hard to implement, then sure - that would be great to add it now. If now, then this definitely has to be added later 👍

pawellysy:
> I know it wasn't part of the initial draft, but how hard would it be to add it?

Maybe it wouldn't be hard to do, each of the axes just needs to add an event on extreme change, and we need to make sure, that no infinite loops are going on. 
However, I thought that the standalone navigator would be a single source of changing the extremes for all the charts. Binding both ways would be just a sync, similar to the one implemented in dashboards, and in my opinion, it shouldn't be implemented as a part of this feature.

ts/Stock/Navigator/StandaloneNavigator.ts
bm64: I've tried it but still results in a problem with types.

pawellysy (approved):

kamil-musialowski (approved):

pawelfus (changes requested):
Tiny bug: http://localhost:3030/samples/#view/stock/standalone-navigator/setrange - click on button -> error in console

pawelfus (approved):
Amazing, thank you!

docs/stock/standalone-navigator.md
```diff
@@ -0,0 +1,50 @@
...
+```
+
+- As a bundled package that includes both Highcharts and the Navigator:
+```html
+<script src="https://code.highcharts.com/standalone-navigator.js"></script>
+```
+
+## Binding Charts to the Navigator
+
+The `bind` method binds chart or axis to the Standalone Navigator. Once a chart is bound, any range adjustments made using the navigator are automatically reflected in the bound chart. See the [bind method documentation](https://api.highcharts.com/class-reference/Highcharts.StandaloneNavigator.bind) for more details.
```
TorsteinHonsi, line 36: ... binds _a_ chart or _an_ axis ...

TorsteinHonsi (changes requested):
Looks perfect, great work! Just a small grammar error in the docs, nothing else to nitpick on.

TorsteinHonsi (approved):

PR #20163, opened 2023-11-20
Fixed #19726 Docs did not adequately explain tickPlacement for category axes

### LINKED ISSUE #19726: `tickmarkPlacement: 'between'` does not work for Heatmap
#### Actual behaviour
The `xAxis.tickmarkPlacement: 'between'` does not work for the Heatmap series (but it seems to work fine for other series). 

- a strange grid line appears on the left side of the chart (outside of plot area) when this option is set
- the heatmap points are not "between" the ticks, but are rendered "on the ticks"
- the `xAxis.tickmarkPlacement: 'on'` seems to act like the `between` should

**Demo:** https://jsfiddle.net/BlackLabel/h68rsczL/

#### Product version
Highcharts 11

#### Affected browser(s)
Chrome, but probably all?

#### Workaround
While the `'between'` option does not work, setting the `xAxis.tickmarkPlacement` to `'on'` acts like the `'between'` option should. **Demo:** https://jsfiddle.net/BlackLabel/9mnjteyq/

MarkusBarstad:
Fixed #19726, Docs did not explain that for category axes, `tickWidth` must be set in order to the see the effects of `tickPlacement`.

NB: I have an issue building docs, please somebody test that this does not format wierdly! 😆

### DIFF
--- a/ts/Core/Axis/AxisDefaults.ts
+++ b/ts/Core/Axis/AxisDefaults.ts
@@ -1744,7 +1744,8 @@ namespace AxisDefaults {
          * For categorized axes only. If `on` the tick mark is placed in the
          * center of the category, if `between` the tick mark is placed between
          * categories. The default is `between` if the `tickInterval` is 1, else
-         * `on`.
+         * `on`. In order to render tick marks on a category axis it is necessary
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (approved):
Thanks!

PR #20170, opened 2023-11-22
Fixed #19657, gantt grid axis title shortening inconsistent

### LINKED ISSUE #19657: Gantt Chart grid column headers don't display correctly sometimes when there are only 1 or sometimes 2 rows
#### Expected behaviour
Column headers should display correctly and consistently whether there is 1 row or 1000 rows

#### Actual behaviour
When there is only 1 row, and sometimes 2 or 3 rows, the column headers wrap when they shouldn't or go to ellipsis too early in the string. It appears that the code that tells the column headers how much room they have is telling them that they have much less horizontal space than they actually do.

This example demonstrates the bug with 1 row: 
https://jsfiddle.net/vjrxkbhd/2/

This particular example works correctly with 2 rows that are identical to the rows in the previous example, so we can rule out changes due to the data itself: 
https://jsfiddle.net/vjrxkbhd/4/

#### Product version
Highcharts (version not specified in script tags, so latest is assumed)

#### Affected browser(s)
Chrome, Firefox, Edge, and maybe more

TorsteinHonsi:
Fixed #19657, grid axis column headers displayed inconsistently depending on how many rows were in the chart.

### DIFF
--- a/samples/gantt/grid-axis/vertical/demo.js
+++ b/samples/gantt/grid-axis/vertical/demo.js
@@ -12,14 +12,20 @@ Highcharts.ganttChart('container', {
                     text: 'Tasks',
                     rotation: 45,
                     y: -15,
-                    x: -15
+                    x: -15,
+                    style: {
+                        width: '100px'
+                    }
                 }
             }, {
                 title: {
                     text: 'Assignee',
                     rotation: 45,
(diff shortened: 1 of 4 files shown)


bm64 (approved):
Looks good!

PR #20176, opened 2023-11-22
feature/live-candlestick-demo

kamil-musialowski:
Added a new Stock demo with real-time updating candlesticks.

---
  - https://app.asana.com/0/0/1203935305712000


samples/stock/demo/live-candlestick/demo.js
```diff
@@ -0,0 +1,1025 @@
...
+        newClose >= updatedLastPoint[2] ? newClose : updatedLastPoint[2],
+        newClose <= updatedLastPoint[3] ? newClose : updatedLastPoint[3],
+        newClose
+    ];
+}
+
+Highcharts.stockChart('container', {
+
+    chart: {
+        height: 800,
```
pawelfus, line 947: I think the default height is good enough. Then the size of the demo's thumbnail will match other demos. The same in the CSS file

pawelfus (changes requested):
Just a tiny change

samples/stock/demo/live-candlestick/demo.js
kamil-musialowski: I thought about it but I was afraid that the update of the candle won't be visible enough - but I'll change that 👍

samples/stock/demo/live-candlestick/demo.js
```diff
@@ -0,0 +1,1024 @@
...
+        377.16,
+        377.79,
+        377.01,
+        377.4512
+    ]
+];
+
+// Imitate getting point from backend:
+function getNewPoint(i) {
+    const lastPoint = data[data.length - 1];
```
DJTechnoo, line 916: Great job!  If we use this function on an empty dataset it will break, so maybe there should be something to handle that case or prevent any errors.

DJTechnoo (changes requested):
Demo looks great! Just a small thing I mentioned

pawelfus (approved):
Thanks!

samples/stock/demo/live-candlestick/demo.js
kamil-musialowski: Thank you! I agree with your point here, that it won't work without the data - but since this function is created to mimic getting point from the backend, I don't think we should include any no-data prevention here. In real world scenario the implementation would be drastically different since we would work on real values, not values based on previous points.

With such a `if(data)` check, we couldn't even create a new point, because then it would have to have random values (since now they are based on previous point).

But I'll leave the final decision to @pawelfus 😅

pawelfus: No strong opinion here, you are both right. Since it's a demo, then I would say less code is better

bm64 (approved):
LGTM!

DJTechnoo (approved):

TorsteinHonsi (approved):

PR #20186, opened 2023-11-23
bugfix/19857-disappearing-auto-rotation-labels

### LINKED ISSUE #19857: X Axis labels are not rendered on very specific widths, but on larger/smaller they do.
#### Expected behaviour
Render the labels.

#### Actual behaviour
Does not

Does not display labels
https://jsfiddle.net/Resems/atb5vf8j/

Does display labels 
https://jsfiddle.net/Resems/atb5vf8j/1.
https://jsfiddle.net/Resems/atb5vf8j/2/

#### Product version
Latest , v11

#### Affected browser(s)
All

Any workaround for this?

jakubjanuchta:
Fixed #19857, short labels disappeared when `labels.autoRotation` set.

**Demos to test:**
- User's demo: https://jsfiddle.net/BlackLabel/ht2b1fua/
- Adjustable width demo: https://jsfiddle.net/BlackLabel/dhpxj9q1/

### DIFF
--- a/samples/unit-tests/axis/labels/demo.js
+++ b/samples/unit-tests/axis/labels/demo.js
@@ -1,292 +1,292 @@
-// QUnit.test(
-//     'StaggerLines on opposite xAxis should be placed ' +
-//         'between title and axis line. (#4694)',
-//     function (assert) {
-//         var chart = $('#container')
-//                 .highcharts({
-//                     chart: {
-//                         marginTop: null
-//                     },
-//                     xAxis: {
-//                         categories: [],
-//                         opposite: true,
-//                         labels: {
-//                             staggerLines: 3
-//                         }
-//                     },
-//                     series: [
-//                         {
-//                             data: [1, 2, 3, 1, 2, 3]
-//                         }
-//                     ]
-//                 })
-//                 .highcharts(),
-//             labelsBox = chart.xAxis[0].labelGroup.getBBox(),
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (changes requested):
Thanks! 

This is a very marginal case that happens only when the label width is smaller than the line height. In practice as far as I can see only in a category axis when the categories consist of two characters.

I've proposed a cheaper solution at https://github.com/highcharts/highcharts/pull/20190, by just making sure the slot width never becomes smaller than the line height. Can you have a look at that? If you agree, you can bring those changes over to this PR and keep the unit test. 

There are a few visual differences, which I think are legit. There is now more space around the auto-rotated labels which makes them easier to read.

jakubjanuchta:
> Thanks!
> 
> This is a very marginal case that happens only when the label width is smaller than the line height. In practice as far as I can see only in a category axis when the categories consist of two characters.
> 
> I've proposed a cheaper solution at #20190, by just making sure the slot width never becomes smaller than the line height. Can you have a look at that? If you agree, you can bring those changes over to this PR and keep the unit test.
> 
> There are a few visual differences, which I think are legit. There is now more space around the auto-rotated labels which makes them easier to read.

Thanks for following up!

I definitely agree that your solution is much cheaper, I brought your changes to this PR.

TorsteinHonsi (approved):

PR #20187, opened 2023-11-23
dash/error-add-chart

### LINKED ISSUE #20046: Highchart Dashboard : Add Component - Unable to create the chart using add component
#### Expected behaviour
We need to add new chart into the dashboard using "Add Component."

#### Actual behaviour
Unfortunately, I am not able to do that. Kindly refer the screen
image
Even though I use same chart configuration for the "Add component chart" which i used in "Chart 1"
Title only get changed
image

Kindly refer the previous tickets (#19970 , #20004 )

[](https://jsfiddle.net/0bk63vc9/2/)

#### Product version
Highchart Dashboard

karolkolodziej:
Fixed #20046, there was an error when adding a new chart component.

### DIFF
--- a/cypress.config.mjs
+++ b/cypress.config.mjs
@@ -25,7 +25,7 @@ export default defineConfig({
     fixturesFolder: 'test/cypress/fixtures',
     e2e: {
         baseUrl: 'http://localhost:3030/samples/view?mobile=true&path=/',
-        specPattern: `test/cypress/${productsPattern}/integration/*.cy.{js,jsx,ts,tsx}`,
+        specPattern: `test/cypress/${productsPattern}/integration/**/*.cy.{js,jsx,ts,tsx}`,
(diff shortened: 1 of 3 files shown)


cypress.config.mjs
```diff
@@ -25,7 +25,7 @@ export default defineConfig({
     fixturesFolder: 'test/cypress/fixtures',
     e2e: {
         baseUrl: 'http://localhost:3030/samples/view?mobile=true&path=/',
-        specPattern: `test/cypress/${productsPattern}/integration/*.cy.{js,jsx,ts,tsx}`,
+        specPattern: `test/cypress/${productsPattern}/integration/**/*.cy.{js,jsx,ts,tsx}`,
```
karolkolodziej, line 28: @bre1470 @goransle It won't break anything in the Core, right?

bre1470: Looks good. 👍

PR #20195, opened 2023-11-24
tools/product-test-mapping

goransle:
Usage:
To test via detection of staged files `npm run test:precommit` or `npx gulp test --modified`.

To test a specific set of product tests: `npx gulp test --product Core`.

---
  - https://app.asana.com/0/0/1205498509080370


bre1470:
- `npx gulp test --modified` is running all tests if no modification is done. I'm fine with this, but not sure if it is intended.
- `npx gulp test --product` is failing hard when using an unknown product. A nice error message with available Object.keys would be helpful here.

bre1470 (approved):
Otherwise it is working smoothly. 👍

KacperMadej (approved):

PR #20201, opened 2023-11-24
bugfix/20183-animation-redraw

### LINKED ISSUE #20183: Series animation is not respected from user options when updating the series
#### Actual behaviour
When the series animation is set to true from options, it is triggered every second time during redraw, rather than every time.

https://jsfiddle.net/BlackLabel/up26csfk/

#### Product version
Since HC v7.0.0

#### Workaround
Modify the series update function:
https://github.com/highcharts/highcharts/blob/0718d7d251d19c0b7272199830de183f30672034/ts/Core/Series/Series.ts#L4188

with the following:
```     
animation = series.finishedAnimating && {animation: series.userOptions?.animation || false},

```
https://jsfiddle.net/BlackLabel/udohp5n1/

magdalena-gut:
Fixed #20183, `Series.update()` was acting inconsistently when animation set in options.


TorsteinHonsi:
I am not sure if this is how it is supposed to work. The general idea is that [series.animation](https://api.highcharts.com/highcharts/plotOptions.series.animation) should "Enable or disable the initial animation when a series is displayed". Meaning the very first animation, not on updates. On an update, the [chart.animation](https://api.highcharts.com/highcharts/chart.animation) should apply. 

There is still something buggy going on, since the initial animation runs every other time.

In my opinion, the `animation` setting should just be ignored when updating the series, so it behave like https://jsfiddle.net/highcharts/9n2d4vLk/. On the other hand, there might be situations where the implementer would like to run the initial animation again, and then updating with an `animation` property would be the natural way of achieving it... I'm not sure. I don't think it has ever been requested, so I guess the simplest fix would be just to make sure chart animation runs consistently on every update.

magdalena-gut:
> I am not sure if this is how it is supposed to work. The general idea is that [series.animation](https://api.highcharts.com/highcharts/plotOptions.series.animation) should "Enable or disable the initial animation when a series is displayed". Meaning the very first animation, not on updates. On an update, the [chart.animation](https://api.highcharts.com/highcharts/chart.animation) should apply.
> 
> There is still something buggy going on, since the initial animation runs every other time.
> 
> In my opinion, the `animation` setting should just be ignored when updating the series, so it behave like https://jsfiddle.net/highcharts/9n2d4vLk/. On the other hand, there might be situations where the implementer would like to run the initial animation again, and then updating with an `animation` property would be the natural way of achieving it... I'm not sure. I don't think it has ever been requested, so I guess the simplest fix would be just to make sure chart animation runs consistently on every update.

I reported this bug based on the following Stack Overflow topic: https://stackoverflow.com/questions/77501369/highcharts-redraw-issue-line-not-consistently-drawing-on-update?noredirect=1#comment136664162_77501369 .

However, if this is not the expected behavior, finding a solution for every second animation might be more relevant. Setting options.animation = false in the series update method,
seems to be working fine, but I'm not sure if it's not too forceful. What do you think?

TorsteinHonsi:
@magdalena-gut The SO question seems to focus around the animation running _inconsistently_. So I think we should ignore `options.animation` for now and make sure it adheres consistently to the current global animation.

magdalena-gut:
@TorsteinHonsi This happen only when the animation property is set to true during the update call. I debugged it again and the inconsistently comes from the Series.ts update method. On every second update, diffObjects() finds no changes and returns an empty object, causing the merge function to inherit from a prior condition (`animation = series.finishedAnimating && { animation: false }`) which always returns `{animation: false}` . On the next update, diffObjects() does find a change since the animation flag has been toggled, and merge receives `{ animation: true }`. This cause altering between on and off animation with each update. 

If the series animation should be ignored on update, one of my ideas was to force options.animation to false, however I went for moving the animation variable from the second to the last argument during merging options. What do you think about this apporach?

TorsteinHonsi:
> I went for moving the animation variable from the second to the last argument during merging options

@magdalena-gut That's a good solution! Forced options last in the merge, a known common pattern in our code base.

TorsteinHonsi (approved):

PR #20205, opened 2023-11-27
Treegrid columns

### LINKED ISSUE #13618: How to use category type and treegrid type together in highcharts gantt?
I want to have a Gantt chart like the below picture with highcharts. Actually I want to category the tree grids in my Gantt chart. if the tree grids are toggled, the related category will be stretched. I would appreciate it if there is some feature to do this!

bVP69

TorsteinHonsi:
Added support for `grid.columns` on tree-grid axes. 

Demo with the enhancement applied: https://highcharts.github.io/highcharts-utils/samples/#gh/d8a6d112c5/sample/gantt/demo/treegrid-columns

Closes #13618.

#### To do
 - [x] Figure out how to put the main axis (with the collapse/expand arrows) to the left. 
 - [x] The bottom tick label on the linked axis is moved up half a category. The bottom line is missing.
 - [x] Look into padding in the column. Looks like it reserves space for tree collapse/expand arrows.
 - [x] Create a new demo for the demo pages
 - [x] API and docs


TorsteinHonsi:
Setting back to draft, I need to check the CI tests before ready for review.

hubertkozik (approved):
The code looks great! Nice and smooth. I have one question/suggestion, maybe we should forbid to use of this new feature for inverted charts? For example this demo: https://jsfiddle.net/BlackLabel/sjg2du6b/

TorsteinHonsi:
> maybe we should forbid to use of this new feature for inverted charts?

@hubertkozik Fair point! However I don't think the new columns in particular are the problem. The pre-existing tree grid also doesn't work very well inverted. Probably mostly because of `staticScale`, but there are also some animation quirks. Besides the connector paths are off. 

https://jsfiddle.net/highcharts/6ygk9n4h/

hubertkozik:
Yes, that's true. Agree, that the new feature doesn't impact inverted charts.

PR #20206, opened 2023-11-27
Fixed link

nasin:
The link in the venn diagram to the dashboard page was loading into the demo iframe. I fixed it so it loads into the main window.


TorsteinHonsi (approved):
Thanks!

PR #20212, opened 2023-11-27
bugfix/20078-17001-inverted-polar-positions-misaligned

### LINKED ISSUE #17001: Polar inverted: wrong tooltip position for some series types
#### Expected behaviour
Tooltip displayed in the correct place.

#### Actual behaviour
Tooltip works fine for the column series type, but the position is incorrect for line (scatter) series types (with markers in general).

https://jsfiddle.net/BlackLabel/naz4o31L/

#### Product version
9.3.3
Never worked (radial bar chart was added in 8.0.0)

`Internal notes:`
The problem is with the `Tooltip.prototype.getAnchor` method that returns wrong tooltip's position (the `y` value is negative and the tooltip is above the container).
Probably the `inverted` calculations are wrong.

image

https://github.com/highcharts/highcharts/blob/7aebc28faf5b755de9ea8481ba00e23d14745c86/ts/Core/Tooltip.ts#L352-L432

### LINKED ISSUE #20078: Inverted polar chart incorrect points positions
#### Expected behaviour
Points should be displayed in the correct positions.

#### Actual behaviour
Points move (animate) to the top (beyond the chart).

https://jsfiddle.net/BlackLabel/2bgar570/

#### Product version
Affected from: v10.3.3
Last working: v10.3.2

Bisected to: 3510f61bd7cd3f7f1745e5a7d79265ed3e3e7554

KamilKubik:
Fixed #20078, misaligned tooltip on inverted polar chart.

Fixed #17001.

I discovered the #20078 issue during debugging #17001. Since these issues were strictly related, I decided to cover them in the same branch.


ts/Core/Series/Point.ts
TorsteinHonsi, line 1: I think we should move the changes to `Point.pos` out of the core and into the PolarComposition. Create a wrap for `Point.pos` that returns the `chartCoordinates` based on `plotLeft` and `plotTop`, then if the chart is not polar, proceed to the regular function. This way we avoid bloating the core with logic that deals with an add-on.

ts/Core/Series/Series.ts
TorsteinHonsi, line 1: The same here, we should avoid bloating the core. The `PolarComposition.ts` already has overrides for `searchPoint`, so we can just hook up to that and deal with polar & inverted there.

ts/Core/Tooltip.ts
```diff
@@ -664,7 +666,7 @@ class Tooltip {
         const preferFarSide = !this.followPointer &&
             pick(
                 point.ttBelow,
-                !chart.inverted === flipped
+                polar ? (!polar && inverted) : !inverted === flipped
```
TorsteinHonsi, line 669: This is quite a mind bender - what does it mean? What fails if we leave it like it was?

TorsteinHonsi (changes requested):
Thanks! 

I propose to move the logic changes of Point and Series into the polar composition, in order to keep the core clean. For the changes to Tooltip however, it is harder to split them out, so I think we should keep those in the core. They are also more light weight.

ts/Core/Series/Point.ts
KamilKubik: Thanks for the suggestion and clarification on that. I must admit that I wasn't entirely aware of the wrap implementation for the add-on functionalities to the core. I changed it as suggested.

ts/Core/Series/Series.ts
KamilKubik: Same as the above - changed as suggested.

ts/Core/Tooltip.ts
KamilKubik: Yes, I agree. When I'm looking at it now, it doesn't even make the logical sense (but it probably did before to me 😄). It's for the **inverted**, **polar**, and **yAxis.reversed** configuration (reversed y-axis basically). I changed it to `polar ? false : !inverted === flipped`. Thanks!

TorsteinHonsi (approved):

PR #20216, opened 2023-11-28
samples/cyber-assets

goransle:
Looks like jsdelivr is on some popular block-lists

(https://highsoft.slack.com/archives/C3LGP1XQ9/p1701162782795489)


TorsteinHonsi:
Hmm.. We use jsdelivr a lot in our samples. Do you think we should change all those cases too?

goransle:
> Hmm.. We use jsdelivr a lot in our samples. Do you think we should change all those cases too?

I'd say we should probably avoid it in demos and samples used on the homepage. I'll try to include the rest of the `samples/graphics`-URLs.

 As for `samples/data`-assets, it looks like we do do a replace on those in utils, so they might be safe to replace as well?

TorsteinHonsi:
>  it looks like we do do a replace on those in utils, so they might be safe to replace as well?

I think so, but it depends on those assets actually being uploaded to the web page.

TorsteinHonsi (approved):

PR #20219, opened 2023-11-28
Website/cyber venn fixes

nasin:
Hi,  Sales requested some fixes to the Venn diagram, and there was a typo.

Mainly, I reduced the size of the tooltips and wrote some JS to make the tooltips go away (they weren't disappearing I think because the chart margin is 0.) And I created a custom fake tip for the "see terms" link.


TorsteinHonsi (approved):
Thanks!

PR #20220, opened 2023-11-28
dash/accounting-demo

Vazonik:
Added accounting demo.

---
  - https://app.asana.com/0/0/1204432321765571


samples/dashboards/demo/accounting/demo.details
```diff
@@ -0,0 +1,13 @@
+---
+name: Accounting Dashboard
+authors:
+  - Dawid Dragula
+requiresManualTesting: true
```
karolkolodziej, line 5: Does it? If so, add the test description.

karolkolodziej (commented):
1. Budget color doesn't have enough contrast against the background, especially in dark mode
2. NOK's are ok but maybe we should use sth more global like $ or €?
3. It isn't intuitive that you have to click on the chart to see the table, maybe there is a better way to indicate that 🤔 ?
4. If the KPI has a fixed value it will be nice to indicate which period it's about.
5. Also I don't understand what the KPI values are:
image
The gauge says we are 82% in target but next to it, we have an element that says 98% which is which? We have data for 4 years but about which one is it?

jomar-honsi (approved):

samples/dashboards/demo/accounting/demo.html
```diff
@@ -0,0 +1,15 @@
+<script src="https://code.highcharts.com/dashboards/dashboards.js"></script>
+<script src="https://code.highcharts.com/dashboards/modules/math-modifier.js"></script>
+<script src="https://code.highcharts.com/dashboards/datagrid.js"></script>
+<script src="https://code.highcharts.com/stock/highstock.js"></script>
+<script src="https://code.highcharts.com/highcharts-more.js"></script>
+<script src="https://code.highcharts.com/modules/accessibility.js"></script>
+<script src="https://code.highcharts.com/dashboards/modules/dashboards-plugin.js"></script>
+
+<div id="container"></div>
+<div id="popup">
```
sebastianbochan, line 10: In the dashboards you have the build-in popup, was it not useful in this demo?

samples/dashboards/demo/accounting/demo.js
```diff
@@ -0,0 +1,637 @@
...
+    [Date.UTC(2024, 3), null, null, null, 81e5, 84e5, 74e5, 92e5],
+    [Date.UTC(2024, 4), null, null, null, 79e5, 82e5, 72e5, 92e5],
+    [Date.UTC(2024, 5), null, null, null, 81e5, 82e5, 73e5, 90e5],
+    [Date.UTC(2024, 6), null, null, null, 82e5, 83e5, 74e5, 91e5],
+    [Date.UTC(2024, 7), null, null, null, 84e5, 85e5, 75e5, 92e5],
+    [Date.UTC(2024, 8), null, null, null, 83e5, 85e5, 74e5, 93e5],
+    [Date.UTC(2024, 9), null, null, null, 82e5, 82e5, 75e5, 93e5],
+    [Date.UTC(2024, 10), null, null, null, 83e5, 83e5, 74e5, 92e5],
+    [Date.UTC(2024, 11), null, null, null, 81e5, 84e5, 75e5, 93e5]
+];
```
sebastianbochan, line 82: We can push data to the public HC server, like in AWS. We will minify the demo then.

You can discuss it with @goransle

samples/dashboards/demo/accounting/demo.js
```diff
@@ -0,0 +1,637 @@
...
+                useHTML: true,
+                format: '${y}M'
+            }
+        }
+    }
+};
+
+const commonColumnOptions = {
+    chart: {
+        className: 'column',
```
sebastianbochan, line 136: Can we use `highcharts-column-chart` instead? We will keep the namespace.

samples/dashboards/demo/accounting/demo.js
```diff
@@ -0,0 +1,637 @@
...
+    [Date.UTC(2024, 11), null, null, null, 81e5, 84e5, 75e5, 93e5]
+];
+
+const currentYear = new Date(currentMonth).getFullYear();
+
+const commonGaugeOptions = {
+    chart: {
+        height: 150,
+        type: 'gauge',
+        className: 'gauge'
```
sebastianbochan, line 90: Can we use `highcharts-gauge-chart` instead? We will keep the namespace.

samples/dashboards/demo/accounting/demo.js
```diff
@@ -0,0 +1,637 @@
...
+                    from: 83,
+                    to: 92,
+                    className: 'opt-band'
+                }, {
+                    from: 92,
+                    to: 102,
+                    className: 'high-band'
+                }]
+            },
+            series: [{}]
```
sebastianbochan, line 305: Do we need it?

samples/dashboards/demo/accounting/demo.js
```diff
@@ -0,0 +1,637 @@
...
+        },
+        columnAssignment: {
+            Date: 'x',
+            Budget: 'y',
+            Revenue: 'y'
+        },
+        chartOptions: Highcharts.merge(commonColumnOptions, {
+            title: {
+                text: 'Revenue'
+            }
```
sebastianbochan, line 396: You can simplify it by:
```
chartOptions: {
   ...commonColumnOptions,
   {
       title: {
                text: 'Revenue'
       }
   }
}
```

sebastianbochan (changes requested):

sebastianbochan:
It would be good to add the information that columns are clickable. Otherwise is impossible to know about it.

samples/dashboards/demo/accounting/demo.html
Vazonik: If you mean the possibility of creating a popup based on the abstract `BaseForm` class, then it is possible, but in my opinion it is style over substance. I think my idea needs much less code that may cause confusion for users.

samples/dashboards/demo/accounting/demo.js
Vazonik: This is because I'm updating the KPI series instead of adding it, but it's a good point since it won't be a lively updated dashboard anyway. I'll change it.

samples/dashboards/demo/accounting/demo.js
Vazonik: I'm just wondering if it would be enough to change the data format to csv and transfer it to HTML. `demo.js` will then be clean. Unless it's better to host it?

samples/dashboards/demo/accounting/demo.js
```diff
@@ -0,0 +1,658 @@
...
+        costYearlyForecast = costYTD;
+    for (let i = lastRowID + 1; i <= forecastRowID; i++) {
+        revYearlyForecast += table.RevPredA[i] / 1e6;
+        costYearlyForecast += table.CostPredA[i] / 1e6;
+    }
+
+    revKPI.chart.addSeries({ data: [revYTD] });
+    revKPI.chart.subtitle.update({
+        text: `${Math.round(revYTD / revTarget * 100)}% of annual target`
+    });
```
sebastianbochan, line 522: Instead of 2 redraws you can call one.

```
revKPI.chart.addSeries({ data: [revYTD] }, false);
revKPI.chart.subtitle.update({...});
```

sebastianbochan (approved):

samples/dashboards/demo/accounting/demo.html
sebastianbochan: Being honest, I'm not convinced, however if its simpler, we can leave like it is.

samples/dashboards/demo/accounting/demo.js
Vazonik: I don't think I can do it in this case because `subtitle.update` doesn't redraw the chart, it only updates the subtitle `SVGElement`.

See: https://jsfiddle.net/BlackLabel/h53uryze/

However, when I look at the implementation of this method, now I am not sure whether this is the correct behavior, cause the redraw argument is passed only to the [`layOutTitles`](https://github.com/highcharts/highcharts/blob/5132ff9c646ad99c494a5d6d82f4d843e1ba17d2/ts/Core/Chart/Chart.ts#L1263) method, in which it is unused.

Anyway, because I'm not sure is this a bug in Highcharts and whether it will be fixed once, I'll change the order to:
1. subtitle with the second argument false
2. second - `addSeries`.

karolkolodziej (approved):

stitot (commented):
Looks great, but one comment. Big one, and perhaps not something we should fix in this particular demo:

Cost vs Revenue charts doesn't provide an easy and quick way to see results month by month, which I guess is an interesting metric. You have to "manually" compare columns. Could this be one chart?

E.g. Cost and Revenue as columns pairs, and Budget result and Final result as lines on extra axis to the right?

marvin19 (approved):

nasin:
Hi, sorry for the delay on the feedback. i had a lot of other things to do. The only thing I would adjust are the KPI cards to the right of the gauges. I've attached an image with what I was thinking.

accounting-dashboard

nasin (approved):

jomar-honsi (approved):

stitot (commented):
Looks really nice, but would look even better if font-sizes and vertical alignment (center) in the 6 KPI-boxes where more in line with the @nasin screendump. Slightly too big now, and doesn't look "right". Other than that perfect!

samples/dashboards/demo/accounting/demo.css
```diff
@@ -0,0 +1,194 @@
+@import url("https://code.highcharts.com/css/highcharts.css");
+@import url("https://code.highcharts.com/dashboards/css/dashboards.css");
+@import url("https://code.highcharts.com/datagrid/css/datagrid.css");
+
+/* Dashboards */
+
+.highcharts-dashboards-component-title {
```
stitot, line 7: Could be even smaller:

font-size: 0.9rm
margin-bottom: 8px
color: var(--highcharts-neutral-color-100);

samples/dashboards/demo/accounting/demo.css
```diff
@@ -0,0 +1,194 @@
...
+    margin-bottom: 4px;
+    color: var(--highcharts-neutral-color-60);
+}
+
+.highcharts-dashboards-component-subtitle {
+    font-weight: 100;
+    color: var(--highcharts-neutral-color-60);
+}
+
+.highcharts-dashboards-component-kpi-value {
```
stitot, line 20: Could be even smaller and bolder:

font-size: 1.4rem
font-weight: 600

stitot (commented):
Just some more minor tweaks needed to adhere closer to proposed design. See inline comments.

stitot (approved):
Yes! Minor changes, but believe it's worth it. Thanks.

samples/dashboards/demo/accounting/demo.js
```diff
@@ -0,0 +1,657 @@
...
+    tooltip: {
+        formatter: function () {
+            const { x, y, colorIndex: color, series } = this;
+            const date = Highcharts.dateFormat('%B %Y', x);
+
+            return `<span style="font-size: 10px">${date}</span><br>
+                <span class="highcharts-color-${color}">&#9679;</span>&nbsp;
+                ${series.name}: $${(y / 1e6).toFixed(2)}M
+            `;
+        }
```
TorsteinHonsi, line 87: This should be doable with a `tooltip.format`, which is preferred over callbacks because it is serializable and XSS-secure. Off the top of my head, something like this:
```js
format: `<span style="font-size: 10px">{x:%B %Y}</span><br>
                <span class="highcharts-color-{colorIndex}">&#9679;</span>&nbsp;
                {series.name}: ${(divide y 1000000):.2f}M
            `
```

TorsteinHonsi (changes requested):
Thanks, nice work!

In addition to the inline comments, we need to work on responsiveness:
Skjermbilde 2024-02-21 kl  15 12 36

Vazonik:
Thank you @TorsteinHonsi for the review!

>In addition to the inline comments, we need to work on responsiveness

Since responsiveness is based on CSS media queries (v2) instead of the old JS option, mobile view in utils does not correctly display what it looks like in reality on a narrow screen. If you reduce the window width manually, it should be responsive.

stitot:
Good point from @TorsteinHonsi. Media queries is perfectly fine for fullscreen dashboards, but if placed inline in a narrower container it's potentially not optimal.  With the new custom layout possibilities devs can of course do as they please on this, but when it comes to `gui` this is a topic for discussion.

samples/dashboards/demo/accounting/demo.css
```diff
@@ -43,15 +47,15 @@
     color: #aaa;
 }
 
-@media (max-width: 965px) {
+@container dashboard (max-width: 975px) {
```
sebastianbochan, line 50: Can we use the width, used in other media queries? We will keep the consistency.

```
/* LARGE */
@media (max-width: 1200px) {

}

/* MEDIUM */
@media (max-width: 992px) {

}

/* SMALL */
@media (max-width: 576px) {

}
```

TorsteinHonsi (changes requested):
It seems like something is wrong with the width of the gauges now. Regardless of screen size, I can only see the center top part of each gauge.

Skjermbilde 2024-03-06 kl  09 01 04

jomar-honsi (approved):

TorsteinHonsi (approved):

PR #20223, opened 2023-11-29
dash/first-dash-docs

karolkolodziej:
Attempt to fix the first dashboard docs.
Looks like the layout is broken: https://www.highcharts.com/docs/dashboards/your-first-dashboard

### DIFF
--- a/docs/dashboards/your-first-dashboard.md
+++ b/docs/dashboards/your-first-dashboard.md
@@ -28,7 +28,7 @@ If you want to use Highcharts charts, you should also import the charts styles, 
     ```
 
 4. With that, we are ready to start working on an actual dashboard config. Inside the script tag, create a new instance of the dashboard.  
-To do so, you have to pass two arguments:
+To do so, you have to pass two arguments:  
     * the div’s id where the dashboard should be placed (see point 1).
     * the object with dashboard options.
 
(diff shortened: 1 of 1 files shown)


docs/dashboards/your-first-dashboard.md
```diff
@@ -27,8 +27,8 @@ If you want to use Highcharts charts, you should also import the charts styles,
     @import url("https://code.highcharts.com/dashboards/css/dashboards.css");
     ```
 
-4. With that, we are ready to start working on an actual dashboard config. Inside the script tag, create a new instance of the dashboard. <br>
-To do so, you have to pass two arguments:
+4. With that, we are ready to start working on an actual dashboard config. Inside the script tag, create a new instance of the dashboard.  
+To do so, you have to pass two arguments:
```
sebastianbochan, line 31: Remove extra spaces.

sebastianbochan (changes requested):

Vazonik (approved):
Looks ok, except the extra spaces.

Vazonik (approved):
Looks ok, except the extra spaces.

karolkolodziej:
With the double space, I'm trying to fix the issue visible here:
image

I don't see this in VSCode, it is only visible in the released version thus I'm experimenting here slightly 😅

sebastianbochan (approved):

PR #20238, opened 2023-12-03
Refactor: Got rid of SVGElement init function

TorsteinHonsi:
Replaced `SVGElement.init` with `constructor`.


bre1470 (approved):
`-2 any`. Awesome! 👍

PR #20241, opened 2023-12-04
Named type for pdfFont

TorsteinHonsi:
Avoid inline types


cvasseng (approved):
Looks like it takes care of the typing issue in the generator. :)

PR #20242, opened 2023-12-04
bugfix/19730-inverted-centerInCategory-order

### LINKED ISSUE #19730: Center In Category + Inverted chart displays series in reverse
#### Expected behaviour
Independent of settings, series should be shown in the order of series added.

#### Actual behaviour
With inverted chart enabled, categories are shown in the correct order (ex series 1,2,3) but when 'centerInCategory' is enabled in an inverted chart, series are displayed in the opposite order (3,2,1). 

1) Navigate to https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/series-column/centerincategory/ , note on page load, series are displayed in the correct order 
2) Enable 'inverted' under the chart settings, and press 'Run'
The series will now be displayed in the incorrect order.

#### Product version
Highcharts v11.1.0

#### Affected browser(s)
All.

bm64:
Fixed #19730, `centerInCategory` on inverted column chart had incorrect series order.

### DIFF
--- a/samples/unit-tests/series/centerincategory/demo.js
+++ b/samples/unit-tests/series/centerincategory/demo.js
@@ -312,4 +312,17 @@ QUnit.test('series.centerInCategory', function (assert) {
         2,
         '#17764: Points should be evenly spaced, null point between'
     );
+
+    chart.update({
+        chart: {
+            inverted: true
+        }
+    });
+
+    assert.ok(
+        chart.series[0].points[0].barX <
+        chart.series[1].points[0].barX <
+        chart.series[2].points[0].barX,
+        'Points should have correct order in inverted chart, #19730'
+    );
 });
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (changes requested):
Thanks!

I think it actually more a question of whether the x-axis is `reversed`. In an inverted chart, the x-axis is reversed by default, that's why this works. However if we look at a chart where the x-axis is reversed, but the chart not inverted, we'll see that there is a difference. 

https://jsfiddle.net/highcharts/t8aowh6x/2/

If `centerInCategory` is false, "Series 1" is to the right within its category. That should also happen when `centerInCategory` is true.

TorsteinHonsi:
@bm64 Bump

TorsteinHonsi (approved):

PR #20244, opened 2023-12-04
Added stale action

goransle:
---
  - https://app.asana.com/0/0/1205990323776997


TorsteinHonsi (approved):
Thanks!

PR #20249, opened 2023-12-05
tools/bs-safari-version

goransle:
Forgot that Safari is dependent on the correct OS version. Used https://www.browserstack.com/automate/capabilities to get the correct version.


TorsteinHonsi (approved):
Thanks!

PR #20257, opened 2023-12-06
tools/bs-safari-version

goransle:
Seems Safari 16.5 got stuck in fullscreen mode at some point, which threw quite a few tests off. Latest Safari + MacOS works better, see [this test-run](https://github.com/highcharts/highcharts/actions/runs/7111802751)


TorsteinHonsi (approved):
Thanks!

PR #20263, opened 2023-12-06
Tools: Migration to global composed registry.

bre1470:


bre1470:
@TorsteinHonsi It looks like `pushUnique` will save as more bytes than a registry.

PR #20266, opened 2023-12-07
bugfix/20191-beforeunload-export

### LINKED ISSUE #20191: beforeunload is triggered on download a chart
Hi,
I have a chart with download button and a form on the same page. The trigger of beforeunload on download is causing unnecessary prompts for leaving the page 

#### Expected behaviour
beforeunload event should not be triggered by exporting a chart

#### Actual behaviour
beforeunload event is triggered by exporting a chart

https://jsfiddle.net/o0cx86qL/6/

#### Product version
Highcharts 11.2.0

#### Affected browser(s)
Chrome

goransle:
Fixed #20191, exporting a chart using the export module caused `beforeunload` event to fire. 

The export function now uses `fetch` internally. As a result, the `exporting.formAttributes` option has been deprecated. `exporting.fetchOptions` has been introduced  to allow customisation of the `fetch` call.

### Upgrade note
The `exporting.formAttributes` option can for some purposes be replaced with `exporting.fetchOptions`. See [the fetch article on MDN](https://developer.mozilla.org/en-US/docs/Web/API/fetch) for more information about possible options.

Requires https://github.com/highcharts/highcharts-utils/pull/80 for testing

~~Sidenote: should we consider marking https://api.highcharts.com/highcharts/exporting.formAttributes as deprecated?~~

---
  - https://app.asana.com/0/0/1206030888370566

### DIFF
--- a/samples/highcharts/exporting/export-fetch/demo.css
+++ b/samples/highcharts/exporting/export-fetch/demo.css
@@ -1,0 +1,11 @@
+.container {
+    max-width: 600px;
+    min-width: 320px;
+    margin: 0 auto;
+}
+
+#buttonrow {
+    max-width: 600px;
+    min-width: 320px;
+    margin: 0 auto;
+}
--- a/samples/highcharts/exporting/export-fetch/demo.css
+++ b/samples/highcharts/exporting/export-fetch/demo.css
@@ -1,4 +1,4 @@
-.container {
+#container {
     max-width: 600px;
     min-width: 320px;
     margin: 0 auto;
--- a/samples/highcharts/exporting/export-fetch/demo.details
+++ b/samples/highcharts/exporting/export-fetch/demo.details
@@ -1,0 +1,7 @@
+---
+ name: Highcharts Demo
+ authors:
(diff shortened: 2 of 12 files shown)


bre1470 (approved):
No `beforeunload` with the issue example. Nice! 👍

KacperMadej (commented):
Shouldn't tests be in test files and not in a sample used in docs? From the perspective of someone trying only to check out the sample code with the usage explained in the docs the demo is polluted with irrelevant code.

goransle:
> Shouldn't tests be in test files and not in a sample used in docs? From the perspective of someone trying only to check out the sample code with the usage explained in the docs the demo is polluted with irrelevant code.

Good point. I added it to this sample as it already had a `requiresManualTesting` flag, but did not check if it was used in the documentation.

TorsteinHonsi (commented):
The `fetch` API has been supported in evergreen browsers since 2017, which means we can throw out the old form submit feature and replace it with fetch. 

IE11 implementations can use a polyfill.

TorsteinHonsi (approved):

Issue #17870, opened 2022-10-20
highmaps colorAxis legend width truncates rather than re-adjusts


runfaj:
#### Expected behaviour

It appears if you have a floating colorAxis legend for a map chart, the legend just truncates rather than re-adjusting to fit the given width. I would expect the legend/axis to size within the given width.

#### Actual behaviour

https://jsfiddle.net/rd3607s5/

You can see in this fiddle that the the last label is not at the edge of the axis/legend. Upon some element inspection, it appears that the full legend width is still drawn instead of actually sizing to the approropriate given width, so it appears cut off.
<img width="760" alt="image" src="https://user-images.githubusercontent.com/3292053/196988745-e0595b61-5679-4a6e-8166-d6ae2de198d9.png">

Some additional context: I have map charts that will be displayed within different sizes of containers. Because there isn't a maxWidth option for legends (which I also wish was added), I have to do something like this:
```
  let legendWidth = chartWidth / 3;
  if (legendWidth > 200) legendWidth = 200;
  if (legendWidth < 130) legendWidth = 130;
```
This issue happens with both pixel and percentage values. (tested with the above block adding this line: ```legendWidth = `${(legendWidth / chartWidth) * 100}%`;```)

#### Product version
highcharts 10.2.1 with maps module

#### Affected browser(s)
any, mostly tested in chrome latest

KacperMadej:
Hi @runfaj 

Thank you for reporting.

This can be simplified to a Highcharts problem - https://jsfiddle.net/BlackLabel/8gefdt9u/

Both (the simplified and the originally reported problems) can be resolved by setting [`symbolWidth`](https://api.highcharts.com/highcharts/legend.symbolWidth).
Demo: https://jsfiddle.net/BlackLabel/198g7Lhd/

There is no auto adjustment for the legend item nor color axis width after the legend has its width changed and symbolWidth should be used. Not sure if this is a bug or if should this be more like a feature request. Anyway, this is not a Maps-specific problem.

runfaj:
@KacperMadej symbolWidth works great. However, I would never have thought to look at that for this scenario, especially given the description for symbolWidth in the docs. Maybe this one doesn't need any code changes, but could benefit from some additional documentation on both width and symbolWidth for this type of scenario.

oysteinmoseng:
Internal note: We will create new options for `colorAxis.width` & `colorAxis.height`, and also explore supporting % sizes like other axes.

hubertkozik:
[...]

PR #20274, opened 2023-12-07
enh/17870-coloraxis-width-height

hubertkozik:
Added new options [colorAxis.width](https://api.highcharts.com/highcharts/colorAxis.width) and [colorAxis.height](https://api.highcharts.com/highcharts/colorAxis.height), allowing absolute or relative size. See #17870.

---
  - https://app.asana.com/0/0/1205978129852336


ts/Core/Axis/Color/ColorAxisOptions.d.ts
```diff
@@ -43,13 +43,15 @@ export interface ColorAxisMarkerOptions {
 export interface ColorAxisOptions extends AxisOptions {
     dataClassColor?: string;
     dataClasses?: Array<ColorAxisDataClassesOptions>;
+    height?: number;
```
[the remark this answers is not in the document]
hubertkozik: Agree, changes here not needed.

KacperMadej (approved):
👍

samples/highcharts/coloraxis/width-and-height/demo.details
```diff
@@ -0,0 +1,6 @@
+---
+ name: Color axis with specified width and height
```
TorsteinHonsi, line 2: A bit long for the name. It will show on the tab when viewing the sample standalone. "Color axis width and height" would be sufficient.

samples/highcharts/coloraxis/width-and-height/demo.js
```diff
@@ -0,0 +1,11 @@
+Highcharts.chart('container', {
+    colorAxis: {
+        width: '50%',
+        height: 80
+    },
+
```
TorsteinHonsi, line 6: Add a `title` to, so random viewers get some more information. "Color axis with specified width and height" would be a nice fit here.

ts/Core/Axis/Color/ColorAxis.ts
```diff
@@ -293,10 +294,10 @@ class ColorAxis extends Axis implements AxisLike {
      * @private
      */
     public setAxisSize(): void {
-        const axis = this;
-        const symbol = axis.legendItem && axis.legendItem.symbol;
-        const chart = axis.chart;
-        const legendOptions = chart.options.legend || {};
+        const axis = this,
+            chart = axis.chart,
+            symbol = axis.legendItem && axis.legendItem.symbol,
```
TorsteinHonsi, line 299: `axis.legendItem?.symbol`

ts/Core/Defaults.ts
```diff
@@ -1572,6 +1572,10 @@ const defaultOptions: DefaultOptions = {
          * The pixel height of the symbol for series types that use a rectangle
          * in the legend. Defaults to the font size of legend items.
          *
+         * Note: This option is a default source of color axis height, if
```
TorsteinHonsi, line 1575: ... if _the_ `colorAxis.height` option is not...

ts/Core/Defaults.ts
```diff
@@ -1609,6 +1613,10 @@ const defaultOptions: DefaultOptions = {
          * The pixel width of the legend item symbol. When the `squareSymbol`
          * option is set, this defaults to the `symbolHeight`, otherwise 16.
          *
+         * Note: This option is a default source of color axis width, if
```
TorsteinHonsi, line 1616: ... if _the_ `colorAxis.width` option is not...

TorsteinHonsi (changes requested):
Thanks!

The implementation looks very good, except the percentage size is not responding to changes in the chart size:

Open http://localhost:3030/samples/#view/highcharts/coloraxis/width-and-height and toggle the mobile button to reproduce.

Also I am wondering what makes more sense for the percentage size - the plot area size or the full chart size? I am thinking maybe the full chart size makes more sense, if the user wants to position the color axis exactly relative to the full available space. Given that the legend itself is positioned relative to the full chart, this may be more consistent.

hubertkozik:
Definitely, it makes more sense, that the percentage of color axis should be calculated based on the full chart size. During implementation, I based on normal axes, which are calculated based on plot height, but in this case it should be consistent with legend.

TorsteinHonsi (changes requested):
Looks like it still doesn't respond to changes in the chart size?

https://github.com/highcharts/highcharts/assets/227836/49a0f5b8-7108-416e-b75c-b3a538a65599

hubertkozik:
Ah, I've just refreshed the page and didn't check the update after the change of dimensions... I will dig a little bit more.

ts/Core/Axis/Color/ColorAxis.ts
```diff
@@ -401,18 +414,23 @@ class ColorAxis extends Axis implements AxisLike {
...
             padding = legend.padding,
             legendOptions = legend.options,
             labelOptions = axis.options.labels,
             itemDistance = pick(legendOptions.itemDistance, 10),
             horiz = axis.horiz,
             width = pick(
-                legendOptions.symbolWidth,
+                this.options.width ?
+                    relativeLength(this.options.width, chart.chartWidth) :
+                    legendOptions.symbolWidth,
```
TorsteinHonsi, line 427: We now have two places in the code base that does exactly this, and one that is quite similar on like 311. Time for a refactor? What about a function like `getSize()`. 

Regarding line 311, currently it looks like this:

```
relativeLength(
                this.options.width ? this.options.width :
                    symbol.attr('width'),
                chart.chartWidth
            );
```

But I think this can be changed to work more like the two others because `symbol.attr('width'`) can never be a percentage. That allows us to create a function `getSize(width, height)` where the parameters are the default values if an option is not set.

TorsteinHonsi (changes requested):

ts/Core/Axis/Color/ColorAxis.ts
```diff
@@ -837,6 +855,18 @@ class ColorAxis extends Axis implements AxisLike {
...
 }
 
+addEvent(ColorAxis, 'afterSetScale', function (): void {
+    if (this.chart.legend.options.enabled) {
+        const symbolBox = this.legendItem?.symbol?.getBBox(),
+            actualLen = this.horiz ? symbolBox?.width : symbolBox?.height;
+        // Fire update if dimensions have changed, but in another place, than by
+        // updating color axis
+        if (actualLen && actualLen !== this.len) {
+            this.update({});
```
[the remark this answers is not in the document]
hubertkozik: I figured out, that my idea about the event could be bad and we do not need any logic in special events. Instead of this, we can simply overwrite attributes of the existing color axis symbol.

ts/Core/Axis/Color/ColorAxis.ts
hubertkozik: I refactored it, but I think, that I could overthink it, so please feel free to check it deeply.

TorsteinHonsi (approved):
Looks great, thanks!

PR #20276, opened 2023-12-07
bugfix/20006-treegraph-collapsed-export

### LINKED ISSUE #20006: Hichcharts treegraph export to png, crops parent nodes after interacting with it
#### Expected behaviour
After a user interacts with the treegraph, f.e say minimizing some nodes, and then chose to export it (in our case to png), the expected result image, should match the one he created by interacting with the treegraph.

#### Actual behaviour
After a user interacts with the treegraph, f.e say minimizing some nodes, and then chose to export it (in our case to png), the exported image is cropped missing the first parent node, and also the minimized ones are expanded, so the user's actions get ignored

the issue was reproduced straight at your site [https://www.highcharts.com/docs/chart-and-series-types/treegraph-chart]
and it is really easy to reproduce.

Steps:
1. open the link above
2. minimize some random nodes
3. try exporting to png. you ll see that the saved image is cropped.

#### Product version

"highcharts": "^11.1.0",
"@highcharts/map-collection": "^2.0.1",
"highcharts-react-official": "^3.2.1",

#### Affected browser(s)

all browsers

KacperMadej:
Fixed #20006, dynamically collapsed nodes were not exported correctly for treegraph.

---
  - https://app.asana.com/0/0/1205846970443550

### DIFF
--- a/samples/highcharts/demo/treegraph-chart/demo.js
+++ b/samples/highcharts/demo/treegraph-chart/demo.js
@@ -1,7 +1,8 @@
 Highcharts.chart('container', {
     chart: {
         spacingBottom: 30,
-        marginRight: 120
+        marginRight: 120,
+        height: 1200
     },
     title: {
         text: 'Phylogenetic language tree'
--- a/samples/unit-tests/series-treegraph/treegraph/demo.html
+++ b/samples/unit-tests/series-treegraph/treegraph/demo.html
@@ -1,6 +1,7 @@
 <script src="https://code.highcharts.com/highcharts.js"></script>
 <script src="https://code.highcharts.com/modules/treemap.js"></script>
 <script src="https://code.highcharts.com/modules/treegraph.js"></script>
+<script src="https://code.highcharts.com/modules/exporting.js"></script>
 
 <div id="qunit"></div>
 <div id="qunit-fixture"></div>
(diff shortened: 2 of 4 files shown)


bre1470 (approved):
Nice. Looks good! 👍

hubertkozik (approved):
Amazing idea for testing exported chart!

TorsteinHonsi (commented):
Thanks! 

The code works great. I was just wondering, maybe we should just pass this triple updating over to `Point.update`, which usually deals with stuff like this?

```ts
    public toggleCollapse(state?: boolean): void {
        const series = this.series;

        this.update({
            collapsed: state ?? !this.collapsed
        }, false, void 0, false);

        fireEvent(series, 'toggleCollapse');
        series.redraw();
    }
```

We reduce repetition, and it will also serve as a template for future similar operation on points. It also kind of makes sense to treat `toggleCollapse` as a thin proxy to `Point.update`.

KacperMadej:
@TorsteinHonsi 
Applied the suggested improvement 👍 
I based my initial code on [PiePoint's setVisible](https://github.com/highcharts/highcharts/blob/a046a229650799da262905b547dce65a34e0f28f/ts/Series/Pie/PiePoint.ts#L189-L194) - in case that code could or should be also changed.

karakostasi:
yo team!  thanks a lor for resolving this issue. this was a minor or major version upgrade? Asking so we know  what is the version we need to update to. thanks in advance :)

PR #20278, opened 2023-12-08
tools/scripts-css-ignore-dashboards

goransle:


TorsteinHonsi:
Maybe this is not the right time to ask, but why do we have a `scripts-css.js` file for Highcharts and a separate one for Dashboards? We should strive for having the same toolchain, only parametrized, (DRY), otherwise we risk ending up with two similar chains diverging over time, require double the maintenance cost. 

I discovered the same pattern with the `dist-minify` task some time ago, and refactored it to avoid repetition: https://github.com/highcharts/highcharts/commit/dd404704ff64c9cccc8eb94193a6e88443ada9df

goransle:
> Maybe this is not the right time to ask, but why do we have a `scripts-css.js` file for Highcharts and a separate one for Dashboards? We should strive for having the same toolchain, only parametrized, (DRY), otherwise we risk ending up with two similar chains diverging over time, require double the maintenance cost.
> 
> I discovered the same pattern with the `dist-minify` task some time ago, and refactored it to avoid repetition: [dd40470](https://github.com/highcharts/highcharts/commit/dd404704ff64c9cccc8eb94193a6e88443ada9df)

I don't really know, but it might have stemmed from when there were some talk of a separate repository 🤔

 I'll have a try at combining the the script-css's

TorsteinHonsi:
Okay, perfect! Thank you.

TorsteinHonsi (approved):

Issue #20151, opened 2023-11-16
Public access to `linkedSeries` and `linkedParent` Series properties


mreifman:
# Description of the feature
Making `linkedSeries` and `linkedParent` public properties of `Series` objects would allow clients to use those properties in a safe, supported way. There are workaround to accessing private properties, such as by [extending a type](https://github.com/highcharts/highcharts-angular#property-xxx-does-not-exist-on-type-yyy), but doing so is unreliable as the non-public aspects of the Series object may change in the future without backwards compatibility.

Originally discussed in https://www.highcharts.com/forum/viewtopic.php?f=12&t=51674.

# Proof of Concept/Live example for the feature

This came up in a real world example in a chart where multiple line series each have an associated area range series. The chart also has a toggle for showing and hiding the area range series. Being able to access `linkedSeries` and `linkedParent` allows for supporting different permutations of showing an hiding the line series and the linked area range series. 

The visibility of the linked series should always match the visibility of its parent, which means that: 

1. When the area range toggle goes from turned off to turned on, an area range series should only be made visible if its parent series is visible and 
2. When a series goes from not visible to visible via the legend, its linked series should only be shown if the area range toggle is turned on.

https://jsfiddle.net/mreifman/h65a7y9q/7/

hubertkozik:
Done https://github.com/highcharts/highcharts/pull/20280.

PR #20280, opened 2023-12-08
feature/20151-public-linkedseries-linkedparent

hubertkozik:
Converted linkedSeries and linkedParent properties to public #20151.

---
  - https://app.asana.com/0/0/1205971902524722

### DIFF
--- a/ts/Core/Chart/Chart.ts
+++ b/ts/Core/Chart/Chart.ts
@@ -2326,35 +2326,37 @@ class Chart {
         chartSeries.forEach(function (series): void {
             const { linkedTo } = series.options;
 
-            if (linkedTo && isString(linkedTo)) {
-                let seriesLinkedTo: Series | undefined;
+            if (isString(linkedTo)) {
+                let linkedParent: Series | undefined;
                 if (linkedTo === ':previous') {
-                    seriesLinkedTo = chart.series[series.index - 1];
+                    linkedParent = chart.series[series.index - 1];
                 } else {
-                    seriesLinkedTo = chart.get(linkedTo) as Series | undefined;
+                    linkedParent = chart.get(linkedTo) as Series | undefined;
                 }
                 // #3341 avoid mutual linking
                 if (
-                    seriesLinkedTo &&
-                    seriesLinkedTo.linkedParent !== series
+                    linkedParent &&
+                    linkedParent.linkedParent !== series
                 ) {
-                    seriesLinkedTo.linkedSeries.push(series);
+                    linkedParent.linkedSeries.push(series);
                     /**
-                     * Parent series of the current series.
+                     * The parent series of the current series, if the current
+                     * series has a [linkedTo](https://api.highcharts.com/highcharts/series.line.linkedTo)
+                     * setting.
                      *
                      * @name Highcharts.Series#linkedParent
                      * @type {Highcharts.Series}
                      * @readonly
                      */
-                    series.linkedParent = seriesLinkedTo;
+                    series.linkedParent = linkedParent;
 
-                    if (seriesLinkedTo.enabledDataSorting) {
+                    if (linkedParent.enabledDataSorting) {
                         series.setDataSortingOptions();
                     }
 
                     series.visible = pick(
                         series.options.visible,
-                        seriesLinkedTo.options.visible,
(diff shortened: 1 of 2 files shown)


ts/Core/Chart/Chart.ts
```diff
@@ -2324,26 +2324,35 @@ class Chart {
...
+                if (
+                    seriesLinkedTo &&
+                    seriesLinkedTo.linkedParent !== series) {
+                    seriesLinkedTo.linkedSeries.push(series);
+                    /**
+                     * Parent series of the current series.
+                     *
+                     * @name Highcharts.Series#linkedParent
+                     * @type {Highcharts.Series}
+                     */
```
goransle, line 2346: Looks like this  needs an `@apioption` tag to show up on api.highcharts.com (`npm run jsdoc` to test).

I also feel like this should be in `Series.ts`,  but perhaps there's a reason it's here?

hubertkozik: Hmm, it is working correctly for me: http://localhost:9005/class-reference/Highcharts.Series#linkedParent It should show in the class members part of the documenation because it is not an API option, but just a class member.

hubertkozik: When it comes to a place of those lines, I followed the `series.index` property, described here: 
https://github.com/highcharts/highcharts/blob/d3aaa01d7fd79083d503ff2041f70c945e2821f5/ts/Core/Chart/Chart.ts#L690-L697

And here: https://api.highcharts.com/class-reference/Highcharts.Series#index

It seems to me, that the description of that kind of property is before first usage of it.

goransle: Ah, ok. If that's the precedent we should probably stick with that.

goransle (approved):

ts/Core/Chart/Chart.ts
```diff
@@ -2324,26 +2324,37 @@ class Chart {
 
         // Apply new links
         chartSeries.forEach(function (series): void {
-            let linkedTo = series.options.linkedTo;
+            const { linkedTo } = series.options;
 
-            if (isString(linkedTo)) {
+            if (linkedTo && isString(linkedTo)) {
```
TorsteinHonsi, line 2329: Do we need the extended checked for `if (linkedTo` ? This will fail if `linkedTo` is an empty string, but is it necessary to change how it was?

ts/Core/Chart/Chart.ts
```diff
@@ -2324,26 +2324,37 @@ class Chart {
...
+                let seriesLinkedTo: Series | undefined;
```
TorsteinHonsi, line 2330: Instead of calling the local variable `seriesLinkedTo`, I think it should be called `linkedParent` in consistency with the series property.

ts/Core/Chart/Chart.ts
```diff
@@ -2324,26 +2324,37 @@ class Chart {
...
-                if (linkedTo && (linkedTo as any).linkedParent !== series) {
-                    (linkedTo as any).linkedSeries.push(series);
-                    series.linkedParent = linkedTo as any;
+                if (
+                    seriesLinkedTo &&
+                    seriesLinkedTo.linkedParent !== series
+                ) {
+                    seriesLinkedTo.linkedSeries.push(series);
+                    /**
+                     * Parent series of the current series.
```
TorsteinHonsi, line 2343: Should be more elaborated:

The parent series of the current series, if the current series has a [linkedTo](https://api.highcharts.com/highcharts/series.line.linkedTo) setting.

ts/Core/Series/Series.ts
```diff
@@ -432,8 +432,15 @@ class Series {
         const options = series.options,
             visible = options.visible !== false;
 
+        /**
+         * All child series belonging to the current series.
```
TorsteinHonsi, line 436: Should be more elaborated:

All child series that are linked to the current series through the [linkedTo](https://api.highcharts.com/highcharts/series.line.linkedTo) option.

TorsteinHonsi (changes requested):
Thanks! Some comments inline.

TorsteinHonsi (approved):
Thanks!

Issue #2342, opened 2013-10-09
Bug in rendering of dates


anber500:
We have a version of highstock with a patch applied as suggested here:
https://github.com/highslide-software/highcharts.com/issues/1651#issuecomment-25227770

As seen on the images, the dates does not match the dates on the navigator. I assume it's a bug within highstock. We are starting our final TC soon and would like this bug fixed as the client has already indicated that he won't be signing off on it :(

![wrongdates](https://f.cloud.github.com/assets/2919955/1295924/eae432c8-30bb-11e3-8bee-7a9a4d4659db.png)

PR #20282, opened 2023-12-08
Fixed Ordinal with scatter series.

### LINKED ISSUE #19243: Last scatter point not visible
Expected behaviour: 
All scatter point should be visible when chart is loaded.
Please check: 
https://www.highcharts.com/forum/viewtopic.php?f=12&t=51074&e=1&view=unread#unread

Actual behaviour:
Last scatter point is not getting visible when graph is loaded.

Affected browser: chrome, not sure about other browser.

pawellysy:
Fixed #19243, If the last point in the dataset was from scatter series, it was invisible.

### DIFF
--- a/samples/unit-tests/axis/ordinal/demo.js
+++ b/samples/unit-tests/axis/ordinal/demo.js
@@ -667,6 +667,9 @@ QUnit.test('Circular translation, #17128.', assert => {
 
     const chart = Highcharts.stockChart('container', {
             series: data,
+            boost: {
+                enabled: false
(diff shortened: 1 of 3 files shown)


kamil-musialowski (approved):
Wow, nice!

pawelfus (approved):
Thanks!

PS: Tested also #2342: https://jsfiddle.net/BlackLabel/xnactgrd/1/ - works fine

TorsteinHonsi (approved):

PR #20285, opened 2023-12-08
docs/18744-trackbyarea-covered-series

### LINKED ISSUE #18744: Problems with `areaspline` series that is covered by another series
#### Expected behaviour
1. The `series.point.events.click` event does not trigger when the `areaspline` series is covered by another `areaspline` series.
2. When `stickyTracking = false` then the `tooltip` never shows up after hovering over a covered series. 

https://jsfiddle.net/BlackLabel/h3a7pbcr/

1. Try clicking on the point on a blue series 
2. Turn off `stickyTracking` and hover over a point on a blue series 

#### Product version
Highcharts 10.3.3

hubertkozik:
Fixed #18744, the covered area series should be fully clickable.

---
  - https://app.asana.com/0/0/1206054710674049

### DIFF
--- a/samples/unit-tests/series-area/stacked-area-threshold/demo.js
+++ b/samples/unit-tests/series-area/stacked-area-threshold/demo.js
@@ -111,7 +111,6 @@
                     events: {
                         click() {
                             coveredSeriesClicked = true;
-                            console.log(1);
                         }
                     }
                 }
--- a/samples/unit-tests/series-area/stacked-area-threshold/demo.js
+++ b/samples/unit-tests/series-area/stacked-area-threshold/demo.js
@@ -12,105 +12,142 @@
         );
     }
 
-    QUnit.test('Positive (#5280)', function (assert) {
-        var chart = Highcharts.chart('container', {
-            chart: {
-                type: 'area'
-            },
-            plotOptions: {
-                area: {
-                    stacking: 'normal',
-                    threshold: -10
-                }
-            },
-            series: [
-                {
-                    data: [1, 1, 0, 1, 1]
-                },
-                {
-                    data: [1, 1, 0, 1, 1]
-                }
-            ]
-        });
+    // QUnit.test('Positive (#5280)', function (assert) {
+    //     var chart = Highcharts.chart('container', {
+    //         chart: {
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (commented):
There was something not right about this, and now I'm having second thoughts. It just doesn't make sense that `stickyTracking` should affect what happens on clicking through an area. And `trackByArea` immediately sounds like it should do the opposite; when `true`, the area should capture the click for the series that the area belongs to.

In the demo, the most visible inconsistency is that the tooltip for the underlying series is displayed, but when clicking the point, it doesn't click. I think that is because `stickyTracking` causes [k-d-tree search on both series](https://github.com/highcharts/highcharts/blob/d3aaa01d7fd79083d503ff2041f70c945e2821f5/ts/Core/Pointer.ts#L850-L857), as opposed to non-sticky.

Then when clicking, the click is [cancelled if the click doesn't happen on a tracker](https://github.com/highcharts/highcharts/blob/d3aaa01d7fd79083d503ff2041f70c945e2821f5/ts/Core/Pointer.ts#L1091-L1094) element. That logic looks to me like a danger zone, so I am reluctant to change it. 

The best solution I think would be just to disable pointer-events on the area, so that end users can click right through it, just like the tooltip k-d-tree doesn't mind areas. What do you think @hubertkozik ?

```diff
diff --git a/ts/Series/Area/AreaSeries.ts b/ts/Series/Area/AreaSeries.ts
index e07551f685..0467dad3b2 100644
--- a/ts/Series/Area/AreaSeries.ts
+++ b/ts/Series/Area/AreaSeries.ts
@@ -310,6 +310,9 @@ class AreaSeries extends LineSeries {
                     attribs.fill = owner.color || this.color;
                     attribs['fill-opacity'] = pick(options.fillOpacity, 0.75);
                 }
+                area.css({
+                    pointerEvents: this.stickyTracking ? 'none' : 'auto'
+                });
             }
 
             area[verb](attribs);
```

hubertkozik:
@TorsteinHonsi totally agree! `pointer-events` in that place works as a charm.

TorsteinHonsi (approved):

PR #20286, opened 2023-12-09
Replace init with constructor

TorsteinHonsi:
Refactor, replaced init with constructor in some central classes. 

In other classes, like `Axis` and `Legend`, it makes sense to have a separate `init` structure. The `init` function is called again from from the `update` function in these cases. In ES6 classes, it is not possible to call the constructor, therefore a separate function is needed.


bre1470 (approved):
Nice! Looks good. 👍

PR #20288, opened 2023-12-10
Typesafing PlotLineOrBand

TorsteinHonsi:


ts/Core/Axis/PlotLineOrBand/PlotLineOrBand.ts
```diff
@@ -132,36 +142,21 @@ class PlotLineOrBand {
     public render(): (PlotLineOrBand|undefined) {
```
bre1470, line 142: I do not see much value in this chain concept when `undefined` is involved.

ts/Core/Axis/PlotLineOrBand/PlotLineOrBand.ts
```diff
@@ -219,68 +202,74 @@ class PlotLineOrBand {
...
-            path = axis.getPlotBandPath(from, to, options);
+            });
+        } else if (defined(from) && defined(to)) { // Plot band
+            path = axis.getPlotBandPath(
+                logarithmic?.log2lin(from) ?? from,
+                logarithmic?.log2lin(to) ?? to,
+                options
+            );
         } else {
             return;
```
bre1470, line 232: Probably this return lacks the `this`.

bre1470 (approved):
Looks good! 👍

Some half-related comments...

ts/Core/Axis/PlotLineOrBand/PlotLineOrBand.ts
TorsteinHonsi: It seems that the use case is to return `undefined` when the options are not valid. I'll update the docs.

TorsteinHonsi:
Pending checks

PR #20293, opened 2023-12-11
docs/20252-tooltip-bordercolor

### LINKED ISSUE #20252: Highchart tooltip borderColor not working
#### Expected behaviour
There should be border in tooltip

#### Actual behaviour
setting borderColor is doing nothing

#### Live demo with steps to reproduce (change color and see)
https://jsfiddle.net/70ezuvrx/
#### Product version
V11.2.0. running with angular 14

madepiet:
Fixed #20252 - Updated tooltip.borderColor API documentation to clarify visibility conditions. Now it clearly states that the border color is only visible if a `borderWidth` is set, addressing the issue where `borderColor` was not showing when `borderWidth` was `0`.

### DIFF
--- a/ts/Core/Defaults.ts
+++ b/ts/Core/Defaults.ts
@@ -1893,22 +1893,23 @@ const defaultOptions: DefaultOptions = {
      */
     tooltip: {
         /**
-         * The color of the tooltip border. Note that the border color is only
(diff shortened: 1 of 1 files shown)


TorsteinHonsi (approved):
Thanks!

PR #20294, opened 2023-12-11
bugfix/16626-secondary-grid-axis-units 

### LINKED ISSUE #16626: Gantt, with multiple xAxis and different units, breaks depending on xAxis order, or doesn't display correct the set xAxis units.
#### Expected behaviour
* When setting xAxis units, it should stick to the set units
* When setting xAxis month/year, the chart should work

#### Actual behaviour
* When setting xAxis units, month/year it stops working, breaks due to js error.
* When setting xAxis year/month, the xAxis doesn't stick to month, instead display both as years.

#### Use case.
We want to allow users to be able to set the format, which is would be usually 
Year
Quarter/Year ( Year optional )
Month/Quarter/Year ( Year optional )
Day/Month/Year ( Year optional )
Dynamic - Default Gantt Logic

https://jsfiddle.net/Resems/fhktLr3s/30/ - Base version with quarters, works ( ignore the `1`, was used for testing purpose )
https://jsfiddle.net/Resems/fhktLr3s/27/ Month/Year xAxis case, it is breaking.
https://jsfiddle.net/Resems/fhktLr3s/29/ Year/Month xAxis case, not displays months.

#### Product version
Gantt, 9.3.0

#### Affected browser(s)
N/A

jakubjanuchta:
Fixed #16626, secondary grid axis ignored `xAxis.units` set.

### DIFF
--- a/samples/unit-tests/gantt/grid-axis/demo.js
+++ b/samples/unit-tests/gantt/grid-axis/demo.js
@@ -1862,6 +1862,46 @@ QUnit.test(
             '2019',
             '#15692: Primary axis should show years'
         );
+
+        chart.setSize(800);
+
+        chart.update({
+            series: [{
+                data: [{
+                    start: Date.UTC(2018, 6, 2),
+                    end: Date.UTC(2019, 6, 1)
+                }]
+            }],
+            xAxis: [{
+                units: [['month', [6]]]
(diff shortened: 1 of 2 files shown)


TorsteinHonsi (approved):
Thanks!

PR #20295, opened 2023-12-11
docs/20272-lang-numericSymbols

jedrzejruta:
Fixed typing in `lang.numericSymbols` API from `null` to `undefined`.

### DIFF
--- a/ts/Core/Defaults.ts
+++ b/ts/Core/Defaults.ts
@@ -232,7 +232,7 @@ const defaultOptions: DefaultOptions = {
          * [Metric prefixes](https://en.wikipedia.org/wiki/Metric_prefix) used
          * to shorten high numbers in axis labels. Replacing any of the
          * positions with `null` causes the full number to be written. Setting
-         * `numericSymbols` to `null` disables shortening altogether.
+         * `numericSymbols` to `undefined` disables shortening altogether.
          *
          * @sample {highcharts} highcharts/lang/numericsymbols/
          *         Replacing the symbols with text


TorsteinHonsi (changes requested):
Thanks!

I think the type must also be changed to `Array<string> | undefined` in order for the type definitions to be correct:

https://github.com/highcharts/highcharts/blob/af97c35dc858a955d5db19c8e60a3a31cf6cff68/ts/Core/Options.d.ts#L39

The implementation shows that we are already handling the case that it can be undefined:

https://github.com/highcharts/highcharts/blob/af97c35dc858a955d5db19c8e60a3a31cf6cff68/ts/Core/Axis/Axis.ts#L650

jedrzejruta:
> I think the type must also be changed to `Array<string> | undefined` in order for the type definitions to be correct:

After updating the type definition, there had to be added a check if `numericSymbols` is defined, so there isn't a TS error thrown.

TorsteinHonsi (changes requested):
> After updating the type definition, there had to be added a check if numericSymbols is defined

Excellent, that proves the typing works as intended. 

We don't have to do the full `defined(numericSymbols)`. Since it's either an array or undefined, a truthy check is sufficient, and slightly terser:

```ts
        } else if (
            i &&
            numericSymbols &&
            numericSymbolDetector >= 1000
        ) {
```

jedrzejruta:
You're right, just checking for `numericSymbols` is sufficient for this case. I've updated the solution, thanks!

TorsteinHonsi (approved):

PR #20300, opened 2023-12-12
bugfix/19808-range-selector-states

### LINKED ISSUE #19808: Range selector week button weird behavior after using the scrollbar.
#### Expected behaviour
When clicking the week button, and scrolling with the scrollbar, the week button should remain active, as the range doesn't change.

#### Actual behaviour
When clicking the week button, and scrolling with the scrollbar, the week button doesn't remain active.

When I change the data range to a smaller one, at some point, the problem doesn't occur. Moreover, I noticed that even when I change the width of the chart, the result might change, see: 

https://github.com/highcharts/highcharts/assets/47400438/f999e071-8e98-4698-a5f2-cdee415d565f

When chart width is 800, everything works correctly, but when I change it to 700, we can see that the button doesn't remain active after scrolling. 

https://jsfiddle.net/BlackLabel/w9y1u8o0/

https://jsfiddle.net/BlackLabel/sp5fLr0g/

#### Product version
Highstock v11.1.0

#### Affected browser(s)
All

kamil-musialowski:
Fixed #19808, the range selector week button behaved weirdly after using the scrollbar.

~~Fixed #19808, week/month buttons state was incorrect.~~

- [x] Add test once agreed on a final solution.

---
  - https://app.asana.com/0/0/1205568209738908

### DIFF
--- a/ts/Stock/RangeSelector/RangeSelector.ts
+++ b/ts/Stock/RangeSelector/RangeSelector.ts
@@ -527,7 +527,6 @@ class RangeSelector {
         ): void => {
             const range = rangeOptions._range,
                 type = rangeOptions.type,
-                count = rangeOptions.count || 1,
                 button = buttons[i],
                 offsetRange =
                     (rangeOptions._offsetMax as any) -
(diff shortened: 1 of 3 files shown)


kamil-musialowski:
@pawelfus Pushing the solution draft, let me know what you think

pawelfus (changes requested):
The solution looks fine, but I'm not convinced about the numbers (0.99, 1.01 etc). I think it will be better to keep the strict number of days (5, 28, 365 etc) instead of the percentage of the range.

kamil-musialowski:
@pawelfus Agreed, thank you!

I didn't go with the approach i.e `range > 28 && range < 34` since I wanted this solution to work with buttons that have a `count > 1`, i.e 3M  (months) rangeSelector button.

ts/Stock/RangeSelector/RangeSelector.ts
```diff
@@ -527,7 +527,7 @@ class RangeSelector {
         ): void => {
             const range = rangeOptions._range,
                 type = rangeOptions.type,
-                count = rangeOptions.count || 1,
+                day = 864e5,
```
pawelfus, line 530: Variable already defined in the outer scope

pawelfus (changes requested):
Looks promising!
Missing test + api info

kamil-musialowski:
@pawelfus I've changed the logic a bit - my mistake. Previously I was checking if `range is < > than actualRange +/- few days`, while it should go, `is actualRange < > than a range +/- few days.`

pawellysy (approved):
Great job! thanks!

bm64 (approved):
Looks good!

pawelfus (approved):
Thanks!

TorsteinHonsi:
Are we sure we want the button matching to be sloppier? As noted in the issue, this is a regression after a commit dealing with `fixedRange` on an ordinal axis. Could it be that the solution should be related to the ordinal axis rather than the range selector? Maybe what we need is a more granular fix for 8ec07aca85f67a48c0023da2214f927b961dc4b2 that retains the button state?

kamil-musialowski:
New PR with a new solution can be found here: https://github.com/highcharts/highcharts/pull/20522

PR #20309, opened 2023-12-14
Implemented native handling of the `time.timezone` option

### LINKED ISSUE #16070: Replace moment.js with something more modern.
# Description of the feature
moment.js is a bit [eol-ish](https://momentjs.com/docs/#/-project-status/).
Have you considered upgrading to something a bit more modern?

moment.js

----

TorsteinHonsi:
Implemented native handling of the `time.timezone` option, making the dependency on moment.js redundant. 

Fixes #16070.

### To do
 - [x] Graceful failure when the browser doesn't know the timezone name. Throw a warning and fall back to 0 offset.
 - [x] Test across browsers.
 - [x] Update docs for `time.timezone`. Inform about browser capabilities. Advice about polyfill?
 - [x] Remove `moment` option and properties.
 - [x] Remove moment.js from vendor folder if applicable. Edit: No can do. Used in tests.
 - [x] Figure out the regression, http://localhost:3030/samples/#test/stock/time/individual . Edit: due to moment-timezone missing in the old image.


KacperMadej (changes requested):

ts/Core/Time.ts
```diff
@@ -409,19 +408,24 @@ class Time {
...
-                        timestamp,
-                        options.timezone
-                    ).utcOffset() * 60000;
-                };
-            }
+            return (timestamp: number | Date): number => {
+
+                try {
+                    // eslint-disable-next-line new-cap
+                    const hourOffset = Intl.DateTimeFormat('en', {
```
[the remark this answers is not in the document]
TorsteinHonsi: Good catch! Didn't think of testing the half-hour time zones, even though it is not the first time we have issues with them.

ts/Core/Time.ts
```diff
@@ -409,19 +408,30 @@ class Time {
...
+                        timeZone: options.timezone,
+                        timeZoneName: 'shortOffset'
+                    })
+                        .format(timestamp)
+                        .split('GMT')[1]
+                        .replace(':30', '.5');
+
+                    // Make sure only valid numbers are returned, otherwise fall
+                    // back to 0. Safety vent for avoiding full crash in future
+                    // cases that we don't yet know about.
```
KacperMadej, line 425: When checking the [list](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones), it seems like there might be an issue with [Nepal](https://en.wikipedia.org/wiki/UTC%2B05:45) (`Asia/Katmandu`) and possibly with some historical stuff (e.g. [this](https://en.wikipedia.org/wiki/UTC%2B01:24) or [that](https://en.wikipedia.org/wiki/UTC%2B04:51)).

TorsteinHonsi: Thanks for taking the time to test thoroughly! I have a fix for Nepal. The two last ones don't seem to be known by browsers.

KacperMadej (approved):
👍

PR #20312, opened 2023-12-14
dash/json-columnNames

karolkolodziej:
Added the possibility to define a custom path in `columnNames` for JSON connector to parse complex data.


ts/Data/Connectors/JSONConnectorOptions.d.ts
```diff
@@ -57,9 +58,12 @@ export interface JSONConnectorOptions extends DataConnectorOptions {
      */
     orientation?: 'columns'|'rows';
     /**
-     * Order of columns.
+     * If JSON data is column oriented, these options defines keys
```
TorsteinHonsi, line 61: ... these options define*

TorsteinHonsi (commented):
Just a question - for accessing deep nested elements, in Highcharts we use dot notation. Examples can be seen for the [series.keys](https://api.highcharts.com/highcharts/series.line.keys) option, and in template formats as well. Array indexing is done also with dot notation: `someArray.items.0`. 

Did you consider this for the `columnNames` option?

The benefits are
* Simpler typing. Only strings allowed.
* Which in turn makes it easier for abstract wrappers. Consider for example a GUI, where an end user can type in a string directly regardless of deep nesting or not. 

I am not sure these argument apply to the `columnNames` case, but I thought I'd mention it. We have the `getNestedProperty` utility function to resolve it:

https://github.com/highcharts/highcharts/blob/924d5970f67602255ac1d30709327276d0f49403/ts/Core/Utilities.ts#L1369-L1425

sebastianbochan:
Using the array is safer because some fields in JSON could have dots in names or IDs. We also use this way of parsing in other parts of code (creating Editable options in the Sidebar), so we keep the consistency then.

ts/Data/Converters/CSVConverter.ts
```diff
@@ -636,12 +630,12 @@ namespace CSVConverter {
      * Options that are not compatible with ClassJSON
      */
     export interface SpecialOptions {
-        beforeParse?: DataBeforeParseCallbackFunction;
+        beforeParse?: BeforeParseCallbackFunction;
```
sebastianbochan, line 633: Does it work when dataPolling is enabled?

test/typescript-karma/Dashboards/Component/tests.js
```diff
@@ -869,4 +869,136 @@ test('Data columnAssignment', async function (assert) {
...
+
+    const mountedComponents = dashboard.mountedComponents;
+
+    assert.deepEqual(
+        mountedComponents[0].component.chart.series[0].yData,
+        [30, 20, 50],
+        'Each server instance should be rendered as a column.'
+    );
+
+    assert.deepEqual(
```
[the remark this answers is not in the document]
karolkolodziej: I think we should, if I change that to `assert.ok` I get:
```
This condition will always return 'false' since JavaScript compares objects by reference, not value.ts(2839)
```

ts/Data/Converters/CSVConverter.ts
karolkolodziej: demo added

ts/Data/Converters/JSONConverter.ts
```diff
@@ -179,14 +183,32 @@ class JSONConverter extends DataConverter {
...
+                                converter.headers[columnIndex] ||
+                                    columnIndex.toString(),
+                                converter.columns[columnIndex]
+                            );
+                        }
                     }
                 } else {
+                    const columnNames = converter.headers;
+
+                    if (columnNames && !(columnNames instanceof Array)) {
```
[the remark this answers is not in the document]
karolkolodziej: no, the `this.table.setRows([row], rowIndex);` needs to be outside of the if statement

TorsteinHonsi (approved):
Looks good!

Vazonik (approved):

test/typescript-karma/Dashboards/Component/tests.js
```diff
@@ -869,4 +869,136 @@ test('Data columnAssignment', async function (assert) {
...
+                cells: [{
+                id: 'dashboard-col-1',
+                }]
+            }]
+            }]
+        },
+        components: [{
+            cell: 'dashboard-col-1',
+            connector: {
+            id: 'micro-element'
```
sebastianbochan, line 972: I see a few lines of problems with indends.

sebastianbochan (approved):

PR #20313, opened 2023-12-14
bugfix/19715-dataGrouping-dataSorting-chart-update-fails

### LINKED ISSUE #19715: `dataSorting` and `dataGrouping` options make updating series impossible.
#### Expected behaviour
While `dataSorting` and `dataGrouping` are enabled, the `update` method for the chart and series works correctly.

#### Actual behaviour
While `dataSorting` and `dataGrouping` are enabled, the `chart.update` and `series.update` methods cause the series to not be displayed.

https://jsfiddle.net/BlackLabel/tg27y3q4/

#### Product version
Highcharts v11.1.0

jedrzejruta:
Fixed #19715, chart update fails with `dataGrouping` & `dataSorting`.

### DIFF
--- a/ts/Core/Series/Series.ts
+++ b/ts/Core/Series/Series.ts
@@ -4375,6 +4375,8 @@ class Series {
 
         series.initialType = initialType;
         chart.linkSeries(); // Links are lost in series.remove (#3028)
+        // setData isn't called with dataSorting & dataGrouping enabled (#19715)
+        chart.setSeriesData();
 
         // #15383: Fire updatedData if the type has changed to keep linked
         // series such as indicators updated


ts/Core/Series/Series.ts
```diff
@@ -4375,6 +4375,8 @@ class Series {
 
         series.initialType = initialType;
         chart.linkSeries(); // Links are lost in series.remove (#3028)
+        // setData isn't called with dataSorting & dataGrouping enabled (#19715)
```
TorsteinHonsi, line 4378: The comment is misleading - what does it means that it isn't called? Describe what happens and why we do it, not what failed prior to the fix.

pawelfus (changes requested):
Missing test :)

TorsteinHonsi (changes requested):
Thanks! I think the fix is legit, but some I was a bit confused at first.

The `chart.setSeriesData()` function - it sounded to me that this would open a pandora's box with setting all chart series' data all over again. But looking at the function, I see it only applies in narrow cases where sorting is enabled and data is not yet initialized. So I think we should rename that function to something more descriptive and specific, like `setSortedData`.

We also need a test. We can probably take an existing test and add an update call to it.

TorsteinHonsi (approved):

pawelfus (approved):
Thanks, looks good!

PR #20315, opened 2023-12-15
dash/pm-demo-css

karolkolodziej:
Changed CSS path.

### DIFF
--- a/samples/dashboards/demo/project-management/demo.css
+++ b/samples/dashboards/demo/project-management/demo.css
@@ -1,6 +1,6 @@
 @import url("https://code.highcharts.com/css/highcharts.css");
 @import url("https://code.highcharts.com/css/datagrid.css");
-@import url("https://code.highcharts.com/css/dashboards.css");
+@import url("https://code.highcharts.com/dashboards/css/dashboards.css");
 
 :root,
 .highcharts-light {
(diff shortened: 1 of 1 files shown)


sebastianbochan (approved):

samples/dashboards/demo/project-management/demo.css
```diff
@@ -1,6 +1,6 @@
 @import url("https://code.highcharts.com/css/highcharts.css");
 @import url("https://code.highcharts.com/css/datagrid.css");
```
goransle, line 2: Should probably do this one as well ( --> https://code.highcharts.com/datagrid/css/datagrid.css). https://code.highcharts.com/css/datagrid.css is still there, but seems to be an older version

sebastianbochan: I do not remember that we use  https://code.highcharts.com/datagrid/css/datagrid.css anywhere, should we?

goransle: Looks like https://code.highcharts.com/dashboards/css/datagrid.css is also still an option, but maybe /datagrid would be better for future-proofing? cc @oysteinmoseng

karolkolodziej: Good point but actually we don't use the DataGrid in this demo so it can be removed 😉

goransle: 😆

goransle (approved):

PR #20317, opened 2023-12-15
Refactor: PiePoint.setVisible relay to Point.update

TorsteinHonsi:
Looks like we can strip out a block of redundant code from `PieSeries.setVisible` because `Point.update` already handles the same concerns. 

Based on input from https://github.com/highcharts/highcharts/pull/20276#issuecomment-1857950257


KacperMadej (approved):
👍

PR #20326, opened 2023-12-19
bugfix/17148-xrange-align-datalabel

### LINKED ISSUE #17148: I have an issue with the display name alignment on the plots.
#### Expected behaviour
the name should be aligned with the plot.

#### Actual behaviour
When i collapse the grouping, the name sometimes miss align with the plot.

Demo Video: https://www.screencast.com/t/bYNmEW619lpm

#### Product version

highcharts: 9.2.2
highcharts-angular: 2.10.0

#### Affected browser(s)
Google Chrome
Version 100.0.4896.60 (Official Build) (64-bit)

marvin19:
Added width to data label in x-range and gantt series, allowing implementers to easily apply word-wrap or ellipsis within the tasks. See #17148.

### DIFF
--- a/ts/Series/XRange/XRangeSeries.ts
+++ b/ts/Series/XRange/XRangeSeries.ts
@@ -259,6 +259,14 @@ class XRangeSeries extends ColumnSeries {
     public alignDataLabel(point: XRangePoint): void {
         const oldPlotX = point.plotX;
         point.plotX = pick(point.dlBox && point.dlBox.centerX, point.plotX);
+
+        if (point.dataLabel && point.shapeArgs && point.shapeArgs.width) {
+            const width = point.shapeArgs.width;
+            point.dataLabel.css({
+                width: `${width}px`
+            });
+        }
+
         super.alignDataLabel.apply(this, arguments);
         point.plotX = oldPlotX;
     }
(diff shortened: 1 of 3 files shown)


ts/Series/XRange/XRangeSeries.ts
```diff
@@ -259,6 +259,14 @@ class XRangeSeries extends ColumnSeries {
     public alignDataLabel(point: XRangePoint): void {
         const oldPlotX = point.plotX;
         point.plotX = pick(point.dlBox && point.dlBox.centerX, point.plotX);
+
+        if (point.dataLabel && point.shapeArgs && point.shapeArgs.width) {
+            const width = point.shapeArgs.width;
+            point.dataLabel.css({
+                width: `${width}px`
```
TorsteinHonsi, line 266: This can be a little bit more compact if we skip the `const` and just enter `point.shapeArgs.width` directly.

TorsteinHonsi (changes requested):
Thanks!

In addition to the inline nitpicking, I think we should change one or two of the samples to show that we can use `textOverflow: 'ellipsis'` to make the text stay within the task. If we take one of the three samples that failed yesterday and configure it with ellipsis, that would serve as a good demo and future regression test.

TorsteinHonsi (approved):

PR #20331, opened 2023-12-19
Support y prop on points connected to tree grid

### LINKED ISSUE #14400: Highcharts gantt-chart: Not supporting toggle feature on Y axis with categories declaration on YAxis configuration
#### Expected behaviour
I want to implement "drag and drop" feature vertically on gantt chart. I am not able to achieve it without defining categories on YAxis. But if I declare categories on YAxis then the parent feature stops working.

#### Actual behaviour
Only supports the horizontal "drag and drop" feature. vertical "drag and drop" feature is not working. 

Want to implement vertical drag and drop functionality along with toggle functionality. 
https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/gantt/demo/subtasks

#### Product version

highcharts-gantt: v8.2.0
draggable-points: v8.2.0

#### Affected browser(s)
Chrome.

TorsteinHonsi:
Fixed #14400, drag and drop of tasks in the vertical direction was not supported with the default treegrid axis.

Demo: https://jsfiddle.net/highcharts/Lu91zavj/

Edit: It is probably more complicated that just allowing the `y` property. Since the y-axis categories are picked from the point values, those need to be updated too.

### DIFF
--- a/ts/Core/Axis/TreeGrid/TreeGridAxis.ts
+++ b/ts/Core/Axis/TreeGrid/TreeGridAxis.ts
@@ -336,8 +336,8 @@ function getTreeGridFromData(
                 const data = node.data;
 
                 if (isObject(data, true)) {
-                    // Update point
-                    data.y = start + (data.seriesIndex || 0);
+                    // Update point. Allow user-defined y (#14400).
+                    data.y ??= start + (data.seriesIndex || 0);
                     // Remove the property once used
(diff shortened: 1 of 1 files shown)


TorsteinHonsi:
Merging this, thoug it doesn't fully address the issue, it goes half way.

PR #20335, opened 2023-12-20
bugfix/20175-incorrect-annotation-label-position

### LINKED ISSUE #20175: Inconsistency while updating text annotation using `annotation.update`
#### Expected behaviour
For an annotation while doing `annotation.update()` if an option is not passed, it should remain as is, and provided options should be updated. (Let me know if this is incorrect expectation from the API)

#### Actual behaviour
In below reproducer, my use-case is to update just the color of text annotation without changing any of its options (including position), but when I do -

 `chart.annotations[0].update({labels: [{style: {color: 'yellow'}}]})` 

the label annotation moves to its original position if I had dragged it before. Weirdly other annotation properties like background color etc. remain as is.

A workaround I found is to merge all existing options and then update - 

`chart.annotations[0].update({labels: [_.merge({}, chart.annotations[0].labels[0].options, {style: {color: 'yellow'}})]})`

but wanted to check if this is a bug or is there better way to do this. In general, I want to update any option under `labels[0]`.

https://jsfiddle.net/xmou0463/

Reproducer:

text-annotation-position

#### Product version
11.2.0 (Latest)

#### Affected browser(s)
Chrome

KamilKubik:
Fixed #20175, incorrect annotation label position after update.


TorsteinHonsi (commented):
Thanks!

Do you know why we can't use `this.options` for this fix? In general, `this.options` should always contain the default options plus the user options. 

The `userOptions` should only contain the actual user-set options. Reading `userOptions` should only be necessary when we are actually comparing what the user set against the defaults - in other words it should not be necessary here. This may be a mistake in the Annotation module, possibly in the update method.

TorsteinHonsi:
I think the clue is this:

> In general, this.options should always contain the default options plus the user options.

In [BasicAnnotation](https://github.com/highcharts/highcharts/blob/bc076fb24ea60a8f64945d4c58ff81a525211cee/ts/Extensions/Annotations/Types/BasicAnnotation.ts#L80-L81), we are modifying the `userOptions`, but not the options. So after this drag event, the options are no longer correct. That violates the design principle of `userOptions` vs `options`.

KamilKubik:
> I think the clue is this:
> 
> > In general, this.options should always contain the default options plus the user options.
> 
> In [BasicAnnotation](https://github.com/highcharts/highcharts/blob/bc076fb24ea60a8f64945d4c58ff81a525211cee/ts/Extensions/Annotations/Types/BasicAnnotation.ts#L80-L81), we are modifying the `userOptions`, but not the options. So after this drag event, the options are no longer correct. That violates the design principle of `userOptions` vs `options`.

Yes, I agree, Thank you! Let me go through the changes we have now.

1. `BasicAnnotation.ts` - I added the reference to the `options` as above.
2. `EventEmitter.ts` - I found a bug referencing a minus array position on drag.
3. I'm still thinking about the annotation `update` method. If you could take a look [here](https://jsfiddle.net/BlackLabel/6wphk13r/), after performing the translation, the position during creating an annotation (`update` method) is lost. That's because we base on the `annotation.userOptions` when creating it ([reference](https://github.com/highcharts/highcharts/blob/master/ts/Extensions/Annotations/Annotation.ts#L859)), but the only place where the correct position is saved is `annotation.labels`. Nevertheless, would you classify it as an issue (should the translation be considered during an update)?

TorsteinHonsi:
> On one hand, we create an annotation based on it's options provided by the user which is of course logical but shouldn't we also consider its position when there's a possibility to change it?

I don't grasp the full consequence, but still the pattern stands 
* `userOptions` should hold options provided by the user (either on init, update or GUI operations like dragging, that presumably call `update` under the hood).
* `options` should hold the default options extended by the user options. So in principle, in any Highcharts class, ideally this pattern should apply. In pseudo code:

```ts
// Create
this.userOptions = userOptions;
this.options = merge(defaultOptions, userOptions);

// After an update
merge(true, this.userOptions, updatedOptions);
merge(true, this.options, updateOptions);
```

Vazonik:
The problem resulted from incorrect implementation of the `drag` function by the user in the demo. The fixed demo is [here](https://jsfiddle.net/BlackLabel/85u4yd31/).

During updating, the annotation is destroyed and rendered from scratch, and `options` are reset, only `userOptions` is taken into account. `userOptions` is edited on drag in many places in the code, we could consider a refactor, but it seems to require more effort.

https://github.com/highcharts/highcharts/blob/13b8a578aba1b1f16a0895898ccd0d42f1fc414e/ts/Extensions/Annotations/Types/BasicAnnotation.ts#L80-L82

PR #20337, opened 2023-12-20
dash/improve-sync-options

Vazonik:
Updated sync options typing. Added additional options to highlight sync.

---
  - https://app.asana.com/0/0/1206093271067169

### DIFF
--- a/samples/dashboards/sync/sync-highlight-options/demo.html
+++ b/samples/dashboards/sync/sync-highlight-options/demo.html
@@ -6,7 +6,7 @@
 <div class="inputs">
     <span>Chart Highlight Sync Options: </span>
 
-    <input type="checkbox" class="option" id="enabled" name="showTooltip" checked>
+    <input type="checkbox" id="enabled" name="showTooltip" checked>
     <label for="enabled">enabled</label> /
 
     <input type="checkbox" class="option" id="showTooltip" name="showTooltip" checked>
--- a/samples/dashboards/sync/sync-highlight-options/demo.html
+++ b/samples/dashboards/sync/sync-highlight-options/demo.html
@@ -7,16 +7,16 @@
     <span>Chart Highlight Sync Options: </span>
 
     <input type="checkbox" class="option" id="enabled" name="showTooltip" checked>
-    <label for="enabled"> enabled</label>
+    <label for="enabled">enabled</label> /
 
     <input type="checkbox" class="option" id="showTooltip" name="showTooltip" checked>
-    <label for="showTooltip"> showTooltip</label>
+    <label for="showTooltip">showTooltip</label> /
 
     <input type="checkbox" class="option" id="showMarker" name="showMarker" checked>
-    <label for="showMarker"> showMarker</label>
+    <label for="showMarker">showMarker</label> /
 
     <input type="checkbox" class="option" id="showCrosshair" name="showCrosshair" checked>
-    <label for="showCrosshair"> showCrosshair</label>
+    <label for="showCrosshair">showCrosshair</label>
 </div>
 
 <div id="container"></div>
--- a/samples/dashboards/sync/sync-highlight-options/demo.html
+++ b/samples/dashboards/sync/sync-highlight-options/demo.html
@@ -6,17 +6,17 @@
 <div class="inputs">
     <span>Chart Highlight Sync Options: </span>
 
-    <input type="checkbox" id="enabled" name="showTooltip" checked>
-    <label for="enabled">enabled</label> /
+    <input type="checkbox" class="option" id="cbx-enabled" data-value="enabled" checked>
+    <label for="cbx-enabled">enabled</label> /
 
-    <input type="checkbox" class="option" id="showTooltip" name="showTooltip" checked>
-    <label for="showTooltip">showTooltip</label> /
+    <input type="checkbox" class="option" id="cbx-showTooltip" data-value="showTooltip" checked>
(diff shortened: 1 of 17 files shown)


ts/Dashboards/Components/Component.ts
```diff
@@ -828,6 +827,35 @@ abstract class Component {
 
     }
 
+    /**
+     * Fixes the sync options to be in the correct format.
+     *
+     * @internal
+     */
+    protected fixSyncOptions(): void {
```
karolkolodziej, line 835: It took me a while to figure out what you are trying to `fix` here.
Wouldn't the `convertToSyncOptions` or simply `setSyncOptions`  be better?

ts/Dashboards/Components/Sync/Sync.ts
```diff
@@ -325,6 +319,99 @@ namespace Sync {
...
+    /**
+     * Highlight sync options.
+     *
+     * Example:
+     * ```
+     * {
+     *  enabled: true,
+     *  showTooltip: false,
+     *  showMarker: false,
+     *  showCrosshair: false
```
karolkolodziej, line 365: What is the point of enabling the sync by disabling all other options 🤔. Isn't it disabled then?

ts/Dashboards/Components/Sync/Sync.ts
```diff
@@ -325,6 +319,99 @@ namespace Sync {
...
+         */
+        showTooltip?: boolean;
+        /**
+         * Whether the marker should be synced. When hovering over a point in
+         * other component in the same group, the 'hover' state is enabled at
+         * the corresponding point in this component.
+         *
+         * @default true
+         */
+        showMarker?: boolean;
```
karolkolodziej, line 385: This option isn't that intuitive for the column/bar chart where there is no marker.
Options like `markPoint` or `highlightPoint` would be more generic.
Second though- `enableHoverState`?

karolkolodziej (changes requested):
Good job!

samples/dashboards/sync/sync-highlight-options/demo.html
```diff
@@ -0,0 +1,30 @@
+<script src="https://code.highcharts.com/dashboards/dashboards.js"></script>
+<script src="https://code.highcharts.com/dashboards/datagrid.js"></script>
+<script src="https://code.highcharts.com/highcharts.js"></script>
+<script src="https://code.highcharts.com/dashboards/modules/dashboards-plugin.js"></script>
+
+<div class="inputs">
+    <span>Chart Highlight Sync Options: </span>
+
+    <input type="checkbox" id="enabled" name="showTooltip" checked>
```
sebastianbochan, line 9: 1. The name is not unique, you have another input with the same name.
2. Can we change the name of id? `enabled` sounds not good.

samples/dashboards/sync/sync-highlight-options/demo.js
```diff
@@ -0,0 +1,92 @@
+const board = Dashboards.board('container', {
+    dataPool: {
+        connectors: [{
+            id: 'VegeTable',
```
sebastianbochan, line 4: Small details, but essentially its: `vegatables`. Do we need capitalics here?

samples/dashboards/sync/sync-highlight-options/demo.js
```diff
@@ -0,0 +1,92 @@
...
+const enabledCbx = document.querySelector('#enabled');
+const optionsCbx = document.querySelectorAll('.option');
+
+enabledCbx.addEventListener('change', () => {
+    const { checked } = enabledCbx;
+    const options = {};
+
+    optionsCbx.forEach(checkbox => {
+        checkbox.disabled = !checked;
+        options[checkbox.id] = checkbox.checked;
```
sebastianbochan, line 67: Do we need to refer to checked? When its enabled its true, otherwise is false.

test/cypress/dashboards/integration/sync/highlight-options.cy.js
```diff
@@ -0,0 +1,94 @@
...
+                'When hovering over DataGrid, chart should not have tooltip.'
+            );
+
+            assert.notOk(
+                chart.series[0].points[0].state === 'hover',
+                'When hovering over DataGrid, chart should not have marker.'
+            );
+
+            assert.notOk(
+                chart.xAxis[0].cross && chart.yAxis[0].cross.opacity === 1,
```
sebastianbochan, line 27: You can use optional chaining here. `chart.yAxis[0].cross?.opacity === 1` will be enough.

ts/Dashboards/Components/Component.ts
```diff
@@ -820,6 +818,7 @@ abstract class Component {
         }
 
         this.options = merge(this.options, newOptions);
+        this.fixSyncOptions();
```
sebastianbochan, line 821: Why do we need to call it twice?

```
this.options = merge(this.options, newOptions);
this.fixSyncOptions();
```

sebastianbochan: Can we replace `fixSyncOptions` with something that is more meaningful?

sebastianbochan (changes requested):

samples/dashboards/sync/sync-highlight-options/demo.js
Vazonik: I've just copied the connector & data from different demo, I can fix that.

karolkolodziej (approved):

sebastianbochan (approved):

PR #20338, opened 2023-12-20
bugfix/20291-3d-crosshair

### LINKED ISSUE #20291: Y-axis crosshair is not visible for 3D-column or 3D-bar chart.
#### Expected behaviour
Both X and Y-axis crosshair should be visible when crosshairs : [true, true] is set in tooltip of 3D-column or 3D-bar chart.

#### Actual behaviour
Only X-axis crosshair is visible and not of Y-axis.

Please see in below jsfiddle code, I have set crosshairs: [true, true] in toolip. When you hover on any column, only X-axis crosshair will be visible and not of Y-axis.
https://jsfiddle.net/saurabhkolhe/p8ou1q92/

#### Product version
Highcharts JS v11.2.0

#### Affected browser(s)
All browsers

KacperMadej:
Fixed #20291, y axis crosshair was missing for 3d column.

### DIFF
--- a/ts/Core/Chart/Chart3D.ts
+++ b/ts/Core/Chart/Chart3D.ts
@@ -368,10 +368,7 @@ namespace Chart3D {
          * Whether it is a 3D chart.
          */
         chartProto.is3d = function (): boolean {
-            return Boolean(
-                this.options.chart.options3d &&
-                this.options.chart.options3d.enabled
-            ); // #4280
+            return !!this.options.chart.options3d?.enabled;
         };
 
         chartProto.propsRequireDirtyBox.push('chart.options3d');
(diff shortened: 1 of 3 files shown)


ts/Core/Chart/Chart3D.ts
```diff
@@ -368,10 +368,7 @@ namespace Chart3D {
          * Whether it is a 3D chart.
          */
         chartProto.is3d = function (): boolean {
-            return Boolean(
-                this.options.chart.options3d &&
-                this.options.chart.options3d.enabled
-            ); // #4280
+            return !!this.options.chart.options3d?.enabled;
```
hubertkozik, line 371: I think we're not using double negation anywhere. Wouldn't it be better to use a comparison to `true` or `|| false`?

hubertkozik (commented):
One question below.

ts/Core/Chart/Chart3D.ts
KacperMadej: I can find many `!!(` in the code. My code change is 1 to 1 when it comes to the results. Changing to any of the 2 suggested will still be correct according to the docs and TS but might give different results for non-standard usage.

hubertkozik (approved):

TorsteinHonsi (approved):

Issue #20301, opened 2023-12-13
Long rotated data labels overlap in the Stacked Column chart


hubertkozik:
#### Expected behaviour
Long rotated data labels do not overlap in the Stacked Column chart, when `allowOverlap` is the default.

#### Actual behaviour
Long rotated data labels overlap in the Stacked Column chart,  when `allowOverlap` is the default.

#### Live demo with steps to reproduce
https://jsfiddle.net/BlackLabel/d0wpe31h/

1. Click "4" series legend item to hide that series.
2. Data labels of series "2" and series "3" overlap, but shouldn't.

#### Product version
Highcharts v11.2.0

#### Workaround
Set `useHTML` to `true` and if needed add styles to simulate HC data labels looks.

**Live demo with a workaround:** https://jsfiddle.net/BlackLabel/g7ujkdwx/

PR #20345, opened 2023-12-22
bugfix/20307-rotation-align-verticalalign

### LINKED ISSUE #20307: In bar charts, data label `rotation` does not seem to work well.
# Description of the feature
Rotated data labels works as expected in column charts. However, in a bar chart, it doesn't work. The `align` option aligns the text in the label (I assume), not the label in the bar (as works for column with `verticalAlign`).

Highcharts

https://jsfiddle.net/magnusandersonwk/16ukthd2/
Shows bar charts with/without rotation (not working), then column charts with/without rotation (does work).

----

hubertkozik:
Fixed #20307 and #20301, alignment and overlap detection for rotated data labels didn't work correctly.

---
  - https://app.asana.com/0/0/1206178840022924
  - https://app.asana.com/0/0/1206157168941914

### DIFF
--- a/samples/highcharts/chartchooser/categorical-composition-stacked-bar/demo.js
+++ b/samples/highcharts/chartchooser/categorical-composition-stacked-bar/demo.js
@@ -39,6 +39,8 @@ Highcharts.chart('container', {
                 enabled: true,
                 format: '{series.name}: {y}%',
                 color: 'black',
+                crop: false,
+                allowOverlap: true,
                 rotation: -55,
                 y: -100,
                 x: 30
--- a/samples/highcharts/css/series-datalabels/demo.js
+++ b/samples/highcharts/css/series-datalabels/demo.js
@@ -27,7 +27,6 @@ Highcharts.chart('container', {
         data: [{
             y: 100,
             dataLabels: {
-                align: 'right',
                 rotation: 45,
                 shape: null
             }
--- a/samples/highcharts/css/sunburst/demo.css
+++ b/samples/highcharts/css/sunburst/demo.css
@@ -12,17 +12,17 @@
 }
 
 /* General data label style */
-text.highcharts-data-label {
+.highcharts-data-label text {
     fill: black;
     text-shadow: 1px 1px white, 1px -1px white, -1px -1px white, -1px 1px white;
 }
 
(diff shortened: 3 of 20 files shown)


KacperMadej:
Existing changes are resolving https://github.com/highcharts/highcharts/issues/20301 because for dataLabels with rotation placed is set to true and alignAttr is set for dataLabels (in dataLabel.align). As long as those, or equivalent, changes will be done we should mark this PR as resolving https://github.com/highcharts/highcharts/issues/20301.

hubertkozik:
Current visual differences sound ok to me, due to many changes in the alignment of data labels.

MarkusBarstad (approved):
🚀

TorsteinHonsi (commented):
Very nice piece of work! I set up this demo to see how why the labels moved in the diff, it turns out the new logic greatly improved on the current: https://jsfiddle.net/highcharts/u8bot6mp/1/

Will do a proper review tomorrow.

TorsteinHonsi (changes requested):
Rock solid 💪 .

While working on overlapping detection of rotated labels, maybe you can take a look at an issue that has been bugging me for a while?

In the x-axis of variwide charts, we simply skipped overlapping detection because we lacked support for rotation. See https://www.highcharts.com/demo/highcharts/variwide, where the first few and especially the last few labels overlap.

Skjermbilde 2024-01-12 kl  09 16 18

Peeking at the source code, I was expecting everything to fall into place when I removed the exception for `this.horiz` at this line:

https://github.com/highcharts/highcharts/blob/4ad22a6fc5109fc5e562097f45b5a7f64e655cd8/ts/Series/Variwide/VariwideComposition.ts#L116

But it doesn't work well. The overlapping detection becomes too aggressive and removes more labels than it needs to. Can you look into that? Maybe [bring the commented box preview code](https://github.com/highcharts/highcharts/blob/4ad22a6fc5109fc5e562097f45b5a7f64e655cd8/ts/Core/Series/DataLabel.ts#L362-L381) over to OverlappingDataLabels to see what's going on.

hubertkozik:
Sure! I will take a look and try to fix.

TorsteinHonsi (approved):

PR #20350, opened 2023-12-27
bugfix/20327-rangeselector-inactive-button-click 

### LINKED ISSUE #20327: Zooming issues when rangeSelector.selected set to disabled button
#### Expected behaviour
Panning should preserve the current range

#### Actual behaviour
When the `rangeSelector.selected` option is set to a button that is disabled due to the range exceeding the data range, panning the chart will extend the range to the button instead of limiting it to the data range. 

https://jsfiddle.net/highcharts/sm1o95uf/
* Drag the navigator => x-axis range is extended way outside data range

Or
* Zoom in by mousewheel
* Drag the navigator to the right => poof

Interestingly, if we use the navigator handles to zoom instead of the mouse wheel, then it works as intended

#### Product version
11.2

This issue is forked from https://github.com/highcharts/highcharts/issues/19976

kamil-musialowski:
Fixed #20327, `rangeSelector` allowed to set range larger than possible extremes.

---
  - https://app.asana.com/0/0/1206202876057667

### DIFF
--- a/ts/Stock/RangeSelector/RangeSelector.ts
+++ b/ts/Stock/RangeSelector/RangeSelector.ts
@@ -362,6 +362,18 @@ class RangeSelector {
             this.dropdown.selectedIndex = i + 1;
         }
 
+        const resetFixedRange = (): void => {
+            // If rangeSelector.selected is larger than actual range, reset
+            // chart fixed range (#20327)
+            const xAxis = chart.xAxis[0];
+            if (xAxis.dataMax && xAxis.dataMin && rangeOptions._range) {
+                const dataRange = xAxis.dataMax - xAxis.dataMin;
+                if (rangeOptions._range > dataRange) {
+                    chart.fixedRange = void 0;
+                }
+            }
+        };
+
         // Update the chart
         if (!baseAxis) {
             // Axis not yet instanciated. Temporarily set min and range
@@ -372,6 +384,7 @@ class RangeSelector {
             minSetting = baseXAxisOptions.min;
             baseXAxisOptions.min = rangeMin;
             addEvent(chart, 'load', function resetMinAndRange(): void {
+                resetFixedRange();
                 baseXAxisOptions.range = rangeSetting;
                 baseXAxisOptions.min = minSetting;
             });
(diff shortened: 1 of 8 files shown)


bm64 (approved):

pawellysy (approved):

ts/Stock/RangeSelector/RangeSelector.ts
```diff
@@ -362,6 +362,18 @@ class RangeSelector {
             this.dropdown.selectedIndex = i + 1;
         }
 
+        const resetFixedRange = (): void => {
+            // If rangeSelector.selected is larger than actual range, reset
+            // chart fixed range (#20327)
+            const xAxis = chart.xAxis[0];
+            if (xAxis.dataMax && xAxis.dataMin && rangeOptions._range) {
```
pawelfus, line 369: When checking numbers, be careful with truthy values (`dataMax` or `dataMin` can be 0)

pawelfus (changes requested):
In addition to the comment below:

Did you consider **not** setting `fixedRange` property when the conditions are not met? Instead of setting it + removing it afterward? 

I'm still worried that the current solution in `toFixedRange` can cause leaks - I couldn't come up with any demo though.

Note: Current master branch delivers a bit different behaviour. When selecting a disabled button, we try to fall back to the "All" button, meaning we shouldn't necessarily clear that value, but set it to the correct one.

ts/Stock/RangeSelector/RangeSelector.ts
kamil-musialowski: Rookie mistake, thanks 😬

kamil-musialowski:
@pawelfus Thank you for the review, I did consider not setting it at all, however at the initial chart load, when the `RangeSelector.clickButton()` is fired we don't have any extremes, dataMin/Max, nothing to compare the `fixedRange` with.

To the note: I think it's a good idea, I will change the logic and see if everything works like it should.

pawelfus (changes requested):
As discussed:
- move the function to prototype + rename
- refactor setting fixedRange

pawelfus (approved):
Thanks!

TorsteinHonsi (approved):
Thanks!

Issue #20275, opened 2023-12-07
xAxis.overscroll in pixel values


kamil-musialowski:
# Description of the feature
Create a possibility to input xAxis.overscroll in pixel values i.e `"300px"` to achieve a consistent margin between data and right side of the chart.

# Library related to the feature
Highcharts Stock

----

**You can vote for this feature by adding a thumbs-up reaction to this post.**

raf18seb:
Done in https://github.com/highcharts/highcharts/pull/20360, closing.

raf18seb:
Done in https://github.com/highcharts/highcharts/pull/20360, closing.

PR #20360, opened 2023-12-29
feature/20275-xaxis-overscroll-pixel-percentage

jakubjanuchta:
Added #20275, possibility to set `xAxis.overscroll` in pixels or percentage.


ts/Core/Axis/AxisDefaults.ts
```diff
@@ -1539,15 +1539,27 @@ namespace AxisDefaults {
 
         /**
          * Additional range on the right side of the xAxis. Works similar to
-         * `xAxis.maxPadding`, but value is set in milliseconds. This property
-         * should be set to the same value for both the main `xAxis` and the
+         * `xAxis.maxPadding`, but value is set in milliseconds, percentage or pixels.
+         *
+         * If it's a number, it is interpreted as miliseconds.
```
pawelfus, line 1544: typo `milli`

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -1481,6 +1492,54 @@ namespace OrdinalAxis {
...
+        /**
+         * If overscroll is pixel or pecentage value, convert it to axis range.
+         *
+         * @private
+         * @param {number | string} overscroll
+         * Overscroll value in axis range, pixels or percentage value.
+         * @return {number}
+         * Overscroll value in axis range.
+         */
+        public convertOverscroll(overscroll : number | string = 0) : number {
```
pawelfus, line 1504: I'm curious about the placement of this method. It is not really part of the Ordinal Axis. I don't want to pollute Core Axis with Stock code but we don't have a separate StockAxis extension yet. Perhaps it is time to extract Overscroll option to a new file? What do you think @TorsteinHonsi

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -1481,6 +1492,54 @@ namespace OrdinalAxis {
...
+                overscrollValue = parseInt(overscroll as any, 10);
+
+            if (!defined(axis.max) || !defined(axis.min)) {
+                return 0;
+            }
+
+            if (/%$/.test(overscroll as any)) {
+                // If overscroll is percentage
+                return (axis.max - axis.min) *
+                    overscrollValue / 100;
```
pawelfus, line 1516: I thought the percentage was supposed to work in relation to the length of the axis, not the range 🤔 Could you confirm @TorsteinHonsi ?

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -1481,6 +1492,54 @@ namespace OrdinalAxis {
...
+            }
+
+            if (/px/.test(overscroll as any)) {
+                // If overscroll is pixels
+
+                // overscroll is limited to 90% of the axis width
+                // to prevent division by zero
+                const limitedOverscrollValue = Math.min(
+                        overscrollValue,
+                        axis.width * 0.9
```
pawelfus, line 1526: Use `axis.len`. In inverted chart width and height are swapped, but `axis.len` always refers to the length of the axis

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -1481,6 +1492,54 @@ namespace OrdinalAxis {
...
+         * @private
+         * @param {number | string} overscroll
+         * Overscroll value in axis range, pixels or percentage value.
+         * @return {number}
+         * Overscroll value in axis range.
+         */
+        public convertOverscroll(overscroll : number | string = 0) : number {
+            const ordinal = this,
+                axis = ordinal.axis,
+                overscrollValue = parseInt(overscroll as any, 10);
```
pawelfus, line 1507: Please get rid of `as any` castings

ts/Core/Axis/OrdinalAxis.ts
pawelfus: If it should be 50% of the range, then somehow we have to resolve the problem of mismatching overscroll in navigator and base Axis: https://jsfiddle.net/BlackLabel/t9mpac14/4/ - try panning in the middle chart

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -1481,6 +1492,54 @@ namespace OrdinalAxis {
...
+                // If overscroll is pixels
+
+                // overscroll is limited to 90% of the axis width
+                // to prevent division by zero
+                const limitedOverscrollValue = Math.min(
+                        overscrollValue,
+                        axis.width * 0.9
+                    ),
+                    axisRange = axis.max - axis.min,
+                    rangeWithOverscroll = (axisRange * axis.width) /
```
pawelfus, line 1529: Replace `axis.width` with `axis.len`

pawelfus (changes requested):
Thanks! I have some questions though, see comments inline

ts/Core/Axis/OrdinalAxis.ts
pawelfus: PS: We have in Utils `relativeLength` method, but I don't see at this moment a huge help from that method

ts/Core/Axis/OrdinalAxis.ts
TorsteinHonsi: Sure, if you prefer so. And I agree, if the `overscroll` feature also works with a non-ordinal axis, it strictly doesn't belong in the OrdinalAxis.ts file.

ts/Core/Axis/OrdinalAxis.ts
TorsteinHonsi: > I thought the percentage was supposed to work in relation to the length of the axis, not the range

That was what I was thinking too. But I think you guys should decide what makes more sense. My initial thought was that a percentage setting would make it more predictable in responsive charts, so I guess percentage of the length of the axis would make sense in that case.

ts/Core/Axis/OrdinalAxis.ts
jakubjanuchta: I'll do, thanks!

pawellysy (approved):
Apart from Torstein's and Pawel's comments, everything seems fine! Good job

ts/Core/Axis/OrdinalAxis.ts
pawelfus: To avoid further delay, we decided to keep it here for now. Will be taken care of when working on other related issues/bug/feature.

samples/stock/xaxis/overscroll-pixel/demo.js
```diff
@@ -0,0 +1,39 @@
...
+});
+
+// Create the chart
+Highcharts.stockChart('container', {
+    title: {
+        text: 'Overscroll set in pixels'
+    },
+
+    xAxis: {
+        overscroll: '600px'
```
pawelfus, line 14: Q: What do you think about a lower value? Like 50px? I think it's a more realistic value (a value to reserve some space for eg labels, annotations, predictions/forecastings)

samples/stock/xaxis/overscroll-pixel/demo.js
```diff
@@ -0,0 +1,39 @@
+Highcharts.setOptions({
+    global: {
+        useUTC: false
```
pawelfus, line 3: I think it's unnecessary for this demo

samples/stock/xaxis/overscroll-percent/demo.js
```diff
@@ -0,0 +1,39 @@
+Highcharts.setOptions({
+    global: {
+        useUTC: false
```
pawelfus, line 3: I think it's unnecessary config for this demo

pawelfus (changes requested):
Looks promising! And it's great to see it didn't affect the perf at all!

samples/stock/xaxis/overscroll-pixel/demo.js
jakubjanuchta: When I added a big pixel value, I thought that it could show the user that the overscroll is limited to 90% of the axis width at some point, but you're right - 50px is a more realistic value for an overscroll.

pawelfus (approved):
Looks good, thank you!

samples/stock/xaxis/overscroll-percent/demo.css
```diff
@@ -0,0 +1,4 @@
+#container {
+    height: 400px;
+    min-width: 600px;
```
TorsteinHonsi, line 3: Unnecessary restriction, I confirmed it works well on the mobile. Set `min-width: 320px` as usual.

samples/stock/xaxis/overscroll-pixel/demo.css
```diff
@@ -0,0 +1,4 @@
+#container {
+    height: 400px;
+    min-width: 600px;
```
TorsteinHonsi, line 3: Unnecessary restriction, I confirmed it works well on the mobile. Set `min-width: 320px` as usual.

ts/Core/Axis/AxisDefaults.ts
```diff
@@ -1538,15 +1538,29 @@ namespace AxisDefaults {
 
         /**
          * Additional range on the right side of the xAxis. Works similar to
-         * `xAxis.maxPadding`, but value is set in milliseconds. This property
-         * should be set to the same value for both the main `xAxis` and the
+         * `xAxis.maxPadding`, but value is set in milliseconds, percentage or pixels.
```
TorsteinHonsi, line 1541: This doclet mentions milliseconds, which is true for datetime axes (as is the case in Stock). But the more precise term is _axis values_. It may be better to use that term in case this works with linear axes too, now or in the future.

> Works similar to `xAxis.minPadding`, but the value is set in terms of axis values, percentage or pixels.

ts/Core/Axis/AxisDefaults.ts
```diff
@@ -1538,15 +1538,29 @@ namespace AxisDefaults {
...
+         *
+         * If it's a number, it is interpreted as milliseconds.
```
TorsteinHonsi, line 1543:
```suggestion
         * If it's a number, it is interpreted as axis values, which in a datetime axis equals milliseconds.
```

ts/Core/Axis/AxisDefaults.ts
```diff
@@ -1538,15 +1538,29 @@ namespace AxisDefaults {
...
+         *
+         * If it's a percentage string,
+         * it is interpreted as percentages of axis length.
+         * An overscroll of 50% will make a 100px axis 50px longer.
```
TorsteinHonsi, line 1547: Fill out the lines to 80 chars

ts/Core/Axis/AxisDefaults.ts
```diff
@@ -1538,15 +1538,29 @@ namespace AxisDefaults {
...
+         *
+         * If it's a pixel string, it is interpreted as fixed pixel value,
```
TorsteinHonsi, line 1549:
```suggestion
         * If it's a pixel string, it is interpreted as a fixed pixel value,
```

ts/Core/Axis/AxisDefaults.ts
```diff
@@ -1538,15 +1538,29 @@ namespace AxisDefaults {
...
+         * but limited to 90% of axis width.
```
TorsteinHonsi, line 1550:
```suggestion
         * but limited to 90% of the axis width.
```

ts/Core/Axis/AxisDefaults.ts
```diff
@@ -1538,15 +1538,29 @@ namespace AxisDefaults {
...
+         *
+         * This property should be set to the same value for both the main `xAxis` and the
          * navigator's `xAxis` to maintain consistent behavior during
          * interactions such as dragging or resizing the navigator.
```
TorsteinHonsi, line 1554: Did we consider having the navigator x-axis inherit the `overscroll` of the parent by default? This seems like ouble work for a very common situation. @pawelfus

ts/Core/Axis/OrdinalAxis.ts
```diff
@@ -1487,6 +1517,62 @@ namespace OrdinalAxis {
...
+                    // If overscroll is percentage
+
+                    return calculateOverscroll(overscrollValue / 100);
+                }
+
+                if (/px/.test(overscroll)) {
+                    // If overscroll is pixels
+
+                    // overscroll is limited to 90% of the axis length
+                    // to prevent division by zero
```
TorsteinHonsi, line 1557: Tidy up the comment

ts/Stock/Navigator/NavigatorDefaults.ts
```diff
@@ -451,9 +451,20 @@ const NavigatorDefaults: NavigatorOptions = {
...
+         *
+         * If it's a number, it is interpreted as milliseconds.
+         *
+         * If it's a percentage string,
+         * it is interpreted as percentages of axis length.
+         * An overscroll of 50% will make a 100px axis 50px longer.
+         *
+         * If it's a pixel string, it is interpreted as fixed pixel value,
+         * but limited to 90% of axis width.
+         *
```
TorsteinHonsi, line 464: This doclet looks like the same as the above, so the same corrections would apply here.

TorsteinHonsi (changes requested):
Thanks, great works! I have some comments, mostly proofreading and style guide-related.

ts/Core/Axis/AxisDefaults.ts
pawelfus: Good point. Yes, we did and resigned for some (unknown) reason. Found the related commit: https://github.com/highcharts/highcharts/commit/59108d0b219a7eae081f8f061142ec5ce9a886e6 - no test, so we can only guess and test it manually before implementing

pawelfus: @jakubjanuchta while resolving other issues, could you test it? It should be as simple as inheriting this option in Navigator.ts in `init` (where new Axis is being created)

TorsteinHonsi (changes requested):
Thanks! 

One final change request - now that we made the `navigator.xAxis.overscroll` inherit from `navigator.xAxis`, we have four samples where we set the navigator overscroll unnecessarily. Can you remove those too, and test that everything works as intended?

Skjermbilde 2024-02-29 kl  08 52 46

TorsteinHonsi (approved):

PR #20362, opened 2023-12-29
bugfix/20340-histogram-incorrect-bins-x2-range

### LINKED ISSUE #20340: Histogram - incorrect bins range in tooltip
#### Expected behaviour
The bins range in the tooltip should be correct: 2-3, 3-4, 4-5

#### Actual behaviour
The bins range is incorrect: 2-6, 3-6, 4-5

11.2.0 (current): https://jsfiddle.net/BlackLabel/vbns897g/

Works fine in 11.1.0: https://jsfiddle.net/BlackLabel/4arczejg/

jedrzejruta:
Fixed #20340, histogram bins `x2` value was wrongly calculated.

### DIFF
--- a/samples/unit-tests/series-histogram/histogram/demo.js
+++ b/samples/unit-tests/series-histogram/histogram/demo.js
@@ -473,6 +473,12 @@ QUnit.test('Histogram', function (assert) {
         'Bins frequencies are calculated correctly'
     );
 
+    assert.deepEqual(
+        histogram.data.map(point => point.x2),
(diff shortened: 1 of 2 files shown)


TateGunning (commented):
Hello happy new year. Working right now celebrating 2024. But from the phone screen is tiny. Noticed the changes appeared in line. Couldn't tell though: Are these mergable?

TorsteinHonsi (approved):
Thanks!

PR #20255, opened 2023-12-05
dash/kpi-chart-update

Vazonik:
Fixed KPI component updating when new options include chart options.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1205989776471959

PR #20374, opened 2024-01-04
Fixed #20255, pie data labels overlapped after hovering

### LINKED ISSUE #20355: Pie chart dataLabels overlap when distance is set to "-50%"
#### Expected behaviour
Pie chart should not change the amount of rendered dataLabels upon hover.

#### Actual behaviour
Once you hover over a chart and then go back, all dataLabels are rendered which causes them to overlap.

https://jsfiddle.net/BlackLabel/cnvrm9f5/
1. Visually check the amount of dataLabels
2. Mouse over the chart
3. Mouse out
4. All labels are now rendered

#### Product version
HC 11.2.0

_Internal note: Worked in 11.1, however all dataLabels were rendered initially so I'm not sure if that was a correct behaviour_

TorsteinHonsi:
Fixed #20355, pie data labels inside the pie overlapped after interacting with the chart.

### DIFF
--- a/ts/Extensions/OverlappingDataLabels.ts
+++ b/ts/Extensions/OverlappingDataLabels.ts
@@ -255,9 +255,11 @@ function hideOrShow(label: SVGElement, chart: Chart): boolean {
 
         if (label.oldOpacity !== newOpacity) {
 
-            // Make sure the label is completely hidden to avoid catching clicks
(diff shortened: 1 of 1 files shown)


kamil-musialowski (approved):
Works great now, thanks for the quick fix!

PR #20380, opened 2024-01-04
Climate demo tooltip error

jomar-honsi:
Fixed issue with missing elevation number in map tooltips.
Fixed issue with temperatures getting updated after clicking on any map marker.
Ensured that all charts are updated on temperature scale change.

### DIFF
--- a/samples/dashboards/demo/climate/demo.js
+++ b/samples/dashboards/demo/climate/demo.js
@@ -762,6 +762,7 @@ async function updateBoard(board, city, column, scale, newData) {
     );
     const selectionTable = await dataPool.getConnectorTable('Range Selection');
     const cityTable = await dataPool.getConnectorTable(city);
+
     const [
         timeRangeSelector,
         worldMap,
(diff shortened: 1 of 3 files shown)


samples/dashboards/demo/climate/test-notes.md
```diff
@@ -1 +1,28 @@
-Click on a city marker in the world map. You can also switch the temperature scale in the context menu.
+
+# Manual tests
+
+Visual test of the initial state of the application, followed the observing the effect of user interactions.
+
+## Initial state
```
karolkolodziej, line 6: The `Initial state` is covered with the visual test, there is no need to describe it here

samples/dashboards/demo/climate/demo.js
```diff
@@ -684,14 +684,20 @@ async function setupBoard() {
         });
     }
 
-    // Load initial city
-    await setupCity(board, activeCity, activeColumn, activeScale);
+    // Load cities
+    for (let i = 0, iEnd = cityRows.length; i < iEnd; ++i) {
```
karolkolodziej, line 688: Previously we had some sort of a `lazy loading` but right now we wait for each city.
I would revert this change to have a better initial performance.

karolkolodziej (changes requested):
Good job!

samples/dashboards/demo/climate/demo.js
jomar-honsi: Is it safe to call updateBoards() before all cities are loaded? Placing updateBoards() before or after city loading has an effect on the initial temperatures shown.

karolkolodziej: Hmm, the `updateBoard` should not affect the initial temperatures because we already should have the initial data.
With the current state everything works for me as expected.

karolkolodziej (approved):

samples/dashboards/demo/climate/demo.js
jomar-honsi: This is strange: On my setup, I observe that following (https://www.highcharts.com/demo/dashboards/climate):

1. After loading the temperature is, say, Dublin 12, Bangkok 25.
2. When click on any map marker and the temperatures of all but New York have changed.

I see the same behaviour on Windows 11(Chrome, Edge) and Ubuntu 20.4 (Chrome, Firefox).

jomar-honsi:
A new branch and PR will be created for implementing the changes in the 'climate-all' sample.

Issue #14, opened 2010-07-14
Error in extend function in IE8


TorsteinHonsi:
See http://jsfiddle.net/K4c2R/1/. The error is probably caused by an invalid style value.

PR #20383, opened 2024-01-05
dash/data-connector

bre1470:
Added DataPool.isNewConnector and fixed connector replacements in component.

### DIFF
--- a/ts/Data/DataPool.ts
+++ b/ts/Data/DataPool.ts
@@ -216,19 +216,19 @@ class DataPool implements DataEvent.Emitter {
      *
      * @private
      *
-     * @param {string} id
+     * @param {string} connectorId
      * Name of the connector.
      *
      * @return {DataPoolConnectorOptions|undefined}
      * Returns the options of the connector, or `undefined` if not found.
      */
     protected getConnectorOptions(
-        id: string
+        connectorId: string
     ): (DataPoolConnectorOptions|undefined) {
         const connectors = this.options.connectors;
 
         for (let i = 0, iEnd = connectors.length; i < iEnd; ++i) {
-            if (connectors[i].id === id) {
+            if (connectors[i].id === connectorId) {
                 return connectors[i];
             }
         }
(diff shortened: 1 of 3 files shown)


ts/Data/DataPool.ts
```diff
@@ -255,6 +255,22 @@ class DataPool implements DataEvent.Emitter {
     }
 
 
+    /**
+     * Tests whether the connector has never been requested.
+     *
+     * @param {string} connectorId
+     * Name of the connector.
```
karolkolodziej, line 262:
```suggestion
     * ID of the connector.
```

ts/Data/DataPool.ts
```diff
@@ -346,6 +363,11 @@ class DataPool implements DataEvent.Emitter {
             }
         }
 
+        if (instances[options.id]) {
+            instances[options.id].stopPolling();
```
karolkolodziej, line 367: Why do we stop pooling here? If declared, is it initialized once again? 
Maybe it's worth checking if the `polling` is even enabled first?

karolkolodziej (changes requested):
Please add the PR message.

ts/Data/DataPool.ts
bre1470: Do you call it now connector ID in Dashboards?

ts/Data/DataPool.ts
bre1470: Yes, here the connector gets replaced, and when it gets disposed from the data pool, then there is no point in polling anymore. Also `stopPolling` covers all conditions, no need for an extra check.

ts/Data/DataPool.ts
karolkolodziej: The argument is called `connectorId` so I thought the description should be the same

karolkolodziej (approved):

sebastianbochan:
I’m testing the PR and just found some errors in HC.

Running the https://jsfiddle.net/BlackLabel/837zwjpo/ return HC #14.  The points are not mapped properly, its stange, because it is the form of data like during the init.

The problem is caused by missing "orientation" param. When I added it, it worked https://jsfiddle.net/BlackLabel/0Lkwyqxo/.

bre1470:
@sebastianbochan Okay, so I guess, some merge of defaults is missing for setConnectorOptions.

sebastianbochan:
After some debugging, It turned out that was my stupid mistake. I updated the flag: `firstRowAsNames` from false to true, so it returned an issue. It was not the `orientation` problem.

@bre1470 sorry for it.

PR #20386, opened 2024-01-06
feature/replace-span-with-buttons

pawellysy:
Improved semantics of Stock Tools and popups to for better accessibility, specifically use `<button>` elements rather than spans.

TODO: 

- [x] Keep the separator in the stock tools to be a span


ts/Extensions/Annotations/Popup/PopupAnnotations.ts
```diff
@@ -154,14 +154,25 @@ function addToolbar(
...
-        ))
+    let label = createElement(
+        'p',
+        {
+            className: 'highcharts-annotation-label'
+        },
+        void 0,
+        popupDiv
+    );
+    label.setAttribute('aria-label', 'Annotation type');
```
TorsteinHonsi, line 165: The `createElement` function takes attributes as the second argument. It should work with the `aria-label` attribute too...

TorsteinHonsi (commented):
Looks good except the inline comment, but I suspect you already tried that since you're obviously aware of the attributes argument.

ts/Extensions/Annotations/Popup/PopupAnnotations.ts
pawellysy: Yes, I've tried with the second argument, and something was broken. This is one point, which I have to test out, to see, how exactly can I add the `aria-label` to the correct element. Thanks!

pawelfus (changes requested):
Just one thing: the mini overlay changed too much (wrong order of buttons, missing padding, wrong background colors):
Current:
Zrzut ekranu 2024-01-17 o 13 02 14
PR:
Zrzut ekranu 2024-01-17 o 13 02 20

kamil-musialowski (approved):

bm64 (approved):

pawelfus (changes requested):
Almost there, just the last thing, the cursor: pointer is missing

hover

DJTechnoo (approved):

pawelfus (approved):
Thanks!

TorsteinHonsi (changes requested):
Thanks! There's a failing test in the headless browser tests, that is related to Stock Tools. Can you look into that? Otherwise all good 👍

TorsteinHonsi (approved):

PR #20387, opened 2024-01-08
bugfix/17320-markers-in-boost

### LINKED ISSUE #17320: Using Accessibility module with boost module causes all markers to show on zoom out
#### Expected behaviour

Zooming in and zooming out should not cause all the markers to show on the graph when markers are disabled.

```
 plotOptions: {
        series: {
            marker: { enabled: false },
        }        
  },
```
#### Actual behaviour
Zooming in and zooming out causes all the points to have markers show on the zoomed out graph. 

Zoom in:
image
Once you zoom out, all the markers will show.
Zoom out:
image

https://jsfiddle.net/bhurlston/wxc8q2bv/3/

#### Product version
10.1.0. This also seems to be an issue in 9.3.3. Most likely affects a great amount of versions.

#### Affected browser(s)
Tested in Chrome, latest version

marvin19:
Fixed #17320, using the accessibility module with the boost module caused all markers to show on zoom out

~~Fixed #17320, disabled markers for boosted series that had forced marker.~~


bre1470 (commented):
Looks good. Some form of test would be nice.

ts/Accessibility/Components/SeriesComponent/ForcedMarkers.ts
```diff
@@ -91,6 +91,20 @@ namespace ForcedMarkersComposition {
                 'afterRender',
                 seriesOnAfterRender
             );
+            addEvent(
+                SeriesClass as typeof SeriesComposition,
+                'renderCanvas',
+                function (): void {
```
oysteinmoseng, line 97: Would suggest adding as a `seriesOnRenderCanvas` function, for consistency

oysteinmoseng: Also perhaps add a comment on why this is necessary (incl. issue #)

oysteinmoseng (changes requested):
Thanks, minor thing only!
+ tests as discussed

TorsteinHonsi (approved):
Thanks!

oysteinmoseng (approved):
🎉

PR #20392, opened 2024-01-08
bugfix/linting-unused-vars

Xillez:
Fixed linting issues and enables @typescript-eslint/no-unused-vars linting rule.

Issue: https://app.asana.com/0/1204848917817812/1206033804010331/f


ts/Series/OHLC/OHLCSeries.ts
```diff
@@ -126,6 +126,7 @@ class OHLCSeries extends HLCSeries {
 
     public static compose(
         SeriesClass: typeof Series,
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
bre1470, line 129: You should try `varsIgnorePattern` with `^_` pattern instead. https://eslint.org/docs/latest/rules/no-unused-vars#varsignorepattern, so that parameters with underscore get ignored.

ts/Series/Timeline/TimelineSeries.ts
```diff
@@ -110,8 +110,10 @@ class TimelineSeries extends LineSeries {
     public alignDataLabel(
         point: TimelinePoint,
         dataLabel: SVGLabel,
+        /* eslint-disable @typescript-eslint/no-unused-vars */
         _options: TimelineDataLabelOptions,
         _alignTo: BBoxObject
+        /* eslint-enable @typescript-eslint/no-unused-vars */
```
bre1470, line 116: See the other comment about https://eslint.org/docs/latest/rules/no-unused-vars#varsignorepattern.

bre1470 (changes requested):

bre1470 (approved):

ts/Accessibility/Components/MenuComponent.ts
```diff
@@ -499,6 +499,7 @@ namespace MenuComponent {
      * @private
      */
     export function compose(
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
TorsteinHonsi, line 502: This looks like a mistake - we pass ChartClass, but the implementation actually uses the global Chart. @bre1470 do you know what happened here?

ts/Accessibility/Components/RangeSelectorComponent.ts
```diff
@@ -690,6 +690,7 @@ namespace RangeSelectorComponent {
      */
     export function compose(
         ChartClass: typeof Chart,
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
TorsteinHonsi, line 693: As above, this one looks like a mistake. Why pass RangeSelectorClass when it's not used?

ts/Core/Animation/Fx.ts
```diff
@@ -411,6 +411,7 @@ class Fx {
          */
         function append(
             arr: SVGPath,
+            // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
TorsteinHonsi, line 414: And again

ts/Core/Axis/Axis.ts
```diff
@@ -629,6 +629,7 @@ class Axis {
      */
     public defaultLabelFormatter(
         this: AxisLabelFormatterContextObject,
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
TorsteinHonsi, line 632: And again

ts/Core/Time.ts
```diff
@@ -410,6 +410,7 @@ class Time {
             return (timestamp: number | Date): number => {
 
                 try {
+                    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
TorsteinHonsi, line 413: This is a case when there are unused vars before the ones that are used. Can we use the [after-used](https://eslint.org/docs/latest/rules/no-unused-vars#:~:text=for%20variables%20and-,after%2Dused,-for%20arguments.) flag so that we don't need an exception for this case?

ts/Dashboards/Actions/Bindings.ts
```diff
@@ -183,6 +183,7 @@ namespace Bindings {
     /** @internal */
     export function componentFromJSON(
         json: Component.JSON,
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
TorsteinHonsi, line 186: Better to remove the unused parameter?

TorsteinHonsi (changes requested):
Thanks, nice work!

In general I'm wondering about all the exceptions (`eslint-disable-next-line`). It looks to me like in most cases we should just remove those parameters?

ts/Accessibility/Components/MenuComponent.ts
bre1470: These are probably overlooked, but I recently tried to fix arguments for webpack and with some of these proper compose functions the tests start to fail. So the arguments should be fixed as long as test results are ok; and for the rest we have to fix the actual implementation to make it work.

ts/Dashboards/Actions/Bindings.ts
Xillez: this one is marked with @todo so i assumed it was to be used eventually, but I can remove it 👍

ts/Core/Time.ts
Xillez: hmm, the "" rule doesn't support support after-used on deconstructions as far as I could find, so proposing a minor refactor to:
```
const [hours, minutes = 0] =
    // eslint-disable-next-line new-cap
    Intl.DateTimeFormat('en', {
        timeZone: options.timezone,
        timeZoneName: 'shortOffset'
    })
        .format(timestamp)
        .split(/(GMT|:)/)
        .map(Number)
        // Filter NaNs (string values like "GMT" and ":")
        .filter(function (predicate): boolean {
            return !isNaN(predicate);
        }),
    offset = -(hours + minutes / 60) * 60 * 60000;
```

ts/Accessibility/Components/MenuComponent.ts
Xillez: Tests seems to run fine, so I'll remove 👍

ts/Core/Time.ts
```diff
@@ -410,16 +410,19 @@ class Time {
...
                             timeZoneName: 'shortOffset'
                         })
                             .format(timestamp)
                             .split(/(GMT|:)/)
-                            .map(Number),
+                            .map(Number)
+                            // Filter NaNs (string values like "GMT" and ":")
+                            .filter(function (predicate): boolean {
+                                return !isNaN(predicate);
+                            }),
```
TorsteinHonsi, line 425: I don't think we should add logic just to comply with shortcomings of the linter, so I am happy with the original use of `// eslint-disable-next-line` here.

ts/Dashboards/Layout/Cell.ts
```diff
@@ -289,7 +289,7 @@ class Cell extends GUIElement {
             json.options.parentElement = cell.id;
         }
 
-        const component = componentFromJSON(json, cellContainer);
+        const component = componentFromJSON(json);
```
TorsteinHonsi, line 292: GitHub Actions report an unused variable on line 284. I think it is bad practice leaving unimplemented code with a `// @todo` comment. Mostly `@todo`s like this get forgotten about anyway. So I think we should remove the business code, and either keep a note somewhere else, or leave it commented out.

TorsteinHonsi (changes requested):
Thanks! A couple of comments that may allow us to squeeze out some more bytes.

TorsteinHonsi:
@sebastianbochan I requested a review from you because we changed some redundant code in https://github.com/highcharts/highcharts/pull/20392/commits/b07e10af37ebd086ad1ece3a2029b8b05a071e1c . It may interfere with your plans for DOM containers, could you take a look at it?

TorsteinHonsi (approved):

sebastianbochan:
We are using `getOptions` in most cases so the JSON functionality is something that we will refactor shortly anyway.

sebastianbochan (approved):

PR #20396, opened 2024-01-08
Disabled interfering stock test and added QC to nightly.

bre1470:
Fixes run of `npx gulp test --browsercount 1`.
---
  - https://app.asana.com/0/0/1206299242772887

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -78,6 +78,36 @@ jobs:
         env:
           SLACK_WEBHOOK_URL: ${{ secrets.SLACK_NIGHTLY_ERROR_WEBHOOK }}
 
+  nightly_tests_test:
+    name: Unit-tests interference
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


pawelfus (commented):
I'm not sure what to review here. Why disable the test? It seems the test is 6yo, so why does it interfere now?

bre1470:
@pawelfus 
The test is not necessarily the problem, it is just the trigger for other fails when all tests run in the same browser session:

- When scrollbar on the left, the axis should have an additional margin
   `unit-tests/scrollbar/positions`
- The position of the tooltip is wrong
   `unit-tests/series-column/axis-multiple`
- Tooltip for number axis is correctly formatted
   `unit-tests/series-xrange/xrange`
- X axis should be rendered for the first series
   `unit-tests/series/data`
- Tooltip should have label text of first series
   `unit-tests/tooltip/stickoncontact`

 I was hoping you might know if something recently changed affecting it.

pawelfus:
Unfortunately, nothing comes to my mind

PR #20398, opened 2024-01-09
Added data labels contrast background

TorsteinHonsi:


docs/chart-concepts/series.md
```diff
@@ -150,11 +150,11 @@ Allows the cursor to change appearance to indicate that points and series are cl
...
+Allows data labels to be displayed for each point of data in a series.
 
-![datalabels_demo.png](datalabels_demo.png)
-
-[Try it here](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/demo/line-labels/)
+Rendering labels against a mixed color background requires that readability is ensured. Highcharts handles this by letting the `dataLabels.style.color` default to `contrast`, which renders the label either as black or white depending on the background. In addition, the contrast is enhanced using two strategies:
+* Text outline. By default, the `dataLabels.style.textOutline` is set to `1px contrast`, which provides a contrast outline around the text. With this setting, a white text label will get a 1x black contour around it. The text outline can be disabled by setting it to `none`.
+* Contrast background. In addition, implementers can opt to set the `backgroundColor` to `contrast`. This will provide a background fill that contrasts the text, with just enough opacity to ensure the accessibility contrast requirements are always met. The text outline can be enabled or disabled independently.
```
oysteinmoseng, line 158: Would be neat with a minimal demo showing each strategy visually maybe?

docs/chart-concepts/series.md
```diff
@@ -168,10 +168,7 @@ plotOptions: {
 },
 ```
 
-
-Note: You may wish to disable mouse tracking, which highlights the series and points the mouse hovers over (tooltips will not show if mouse tracking is disabled).
-
-The text displayed on datalabels may also be customized by using the formatter option. See [API reference](https://api.highcharts.com/highcharts/plotOptions.series.dataLabels) for more options.
+The text displayed on data labels may also be customized by using the formatter option. See [API reference](https://api.highcharts.com/highcharts/plotOptions.series.dataLabels) for the full set of options.
```
oysteinmoseng: Maybe we should market the `format` option here, rather than formatter?

oysteinmoseng (changes requested):
Nice, see comments for a couple of suggestions. Also wonder if these demos could benefit from the contrast background, worth checking maybe:
- demo/column-rotated-labels
- demo/sankey-diagram
- demo/waterfall

And if we should mention it in the a11y docs, perhaps here: https://www.highcharts.com/docs/accessibility/patterns-and-contrast

docs/chart-concepts/series.md
TorsteinHonsi: Agreed, and we should also make this text more AIO with recommendations for best readability.

PR #20410, opened 2024-01-10
Demo: different time resolution

MarkusBarstad:
Added demo with series displaying time at different intervals

---
  - https://app.asana.com/0/0/1203875355775326


hubertkozik (changes requested):
The Pull Request should contain a full working demo, not only PoC. I think that's why Torstein changed PR's name. So, I would wait with my Code Review until that.

samples/highcharts/demo/combo-data-resolution/demo.details
```diff
@@ -0,0 +1,10 @@
+---
+name: 'Dual axes, line and column'
+authors:
+  - Torstein Hønsi
```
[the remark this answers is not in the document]
MarkusBarstad: Torstein heavily altered the PoC so I guess me and him are both authors xD

Also, maybe it was a bit premature of me to make a PR for merely a PoC, I guess I should keep this as PR for the demo as a whole right?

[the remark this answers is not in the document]
hubertkozik: Yes, PR for the demo as a whole.

samples/highcharts/demo/combo-data-resolution/demo.details
```diff
@@ -0,0 +1,10 @@
+---
+name: 'Same data, different intervals'
+authors:
+  - Markus Barstad
+js_wrap: b
+tags:
```
hubertkozik, line 6: I think, for Demo pages demos we need also `alt_text` option. I know, that most demos don't have it, but we should add it to new demos.

samples/highcharts/demo/combo-data-resolution/demo.js
```diff
@@ -0,0 +1,307 @@
+const data = [
+    [1451520000000, 0.9363],
+    [1451520000000, 0.99],
```
hubertkozik, line 3: Let's move that data to an HTML file (maybe as CSV? - just thought). If the user would like to check the config of the chart he won't have to scroll through a lot of lines of data.

samples/highcharts/demo/combo-data-resolution/demo.js
```diff
@@ -0,0 +1,307 @@
...
+        pointWidth: 30
+    }, {
+        data: data,
+        pointInterval: 36e5
+    }],
+    yAxis: {
+        tickInterval: 0.5,
+        endOnTick: true,
+        max: 1
+    },
```
hubertkozik, line 303: What do you think about removing the title of the y-axis? If you think it's important, the y-axis title should be specified imo.

samples/highcharts/demo/combo-data-resolution/readme.md
hubertkozik, line 1: The empty file - should be filled with some description.

hubertkozik (commented):
Good idea with the dataset! Some comments below.

samples/highcharts/demo/combo-data-resolution/demo.details
```diff
@@ -0,0 +1,16 @@
...
+name: 'Same data, different intervals'
+authors:
+  - Markus Barstad
+js_wrap: b
+alt_text: >-
+  A chart of monthly mean temperature anomalies from 1880 to 2016.
+  It employs two series, one column and one line,
+  to convey the same data at different resolutions.
+  There is only one column per thirty years,
+  but the line series have a datapoint per year.
```
hubertkozik, line 11: In other demos it is meant to be one sentence of description, e.g. `samples/highcharts/demo/line-custom-entrance-animation/demo.details`.

samples/highcharts/demo/combo-data-resolution/demo.js
```diff
@@ -0,0 +1,36 @@
...
+            but the line series have a datapoint per year.
+        `
+    },
+    series: [{
+        type: 'column',
+        data: data.map((d, i) => (i % 30 === 0) && d),
+        pointWidth: 30
+    }, {
+        data: data,
+        pointInterval: 36e5
```
TorsteinHonsi, line 22: The `pointInterval` shouldn't be necessary, because the data already includes x values.

samples/highcharts/demo/combo-data-resolution/demo.js
```diff
@@ -0,0 +1,36 @@
...
+            A chart of monthly mean temperature anomalies from 1880 to 2016.
+            It employs two series, one column and one line,
+            to convey the same data at different resolutions.
+            There is only one column per thirty years,
+            but the line series have a datapoint per year.
+        `
+    },
+    series: [{
+        type: 'column',
+        data: data.map((d, i) => (i % 30 === 0) && d),
```
TorsteinHonsi, line 18: I'd prefer if we used static data for the columns too. The purpose of this demo is to show how simple it is to set up Highcharts, so we should avoid too much fancy code. Especially since this is only about the demo data.

hubertkozik (approved):

marvin19 (changes requested):
Looks good! A couple of requests:
- Add `yAxix.accessibility.description`, and give a name for your `yAxis` (temperature?)
- Specify units for celsius in tooltip and on `yAxis` labels
- Give both series names(will also look better in legend) 
- Give a `series.accessibility.description` to the second series. It has over 200 data points and screen reader will only take the series description, not each individual point

marvin19 (approved):

samples/highcharts/demo/combo-data-resolution/demo.js
```diff
@@ -0,0 +1,44 @@
...
+            A chart of monthly mean temperature anomalies from 1880 to 2016.
+            It employs two series, one column and one line,
+            to convey the same data at different resolutions.
+            There is only one column per thirty years,
+            but the line series have a datapoint per year.
+        `
+    },
+    series: [{
+        type: 'column',
+        pointWidth: 30,
```
TorsteinHonsi, line 16: I don't think we should have a fixed point width. In this case it leads to overlapping on mobile devices. Better leave it dynamic.

samples/highcharts/demo/combo-data-resolution/demo.js
```diff
@@ -0,0 +1,44 @@
...
+        name: 'Column Series',
+        data: JSON.parse(document.getElementById('columnData').textContent)
+    }, {
+        name: 'Line Series',
```
TorsteinHonsi, line 20: Use a proper name "Yearly average"?

samples/highcharts/demo/combo-data-resolution/demo.js
```diff
@@ -0,0 +1,44 @@
+Highcharts.chart('container', {
+    title: {
+        text: 'Global temperature mean anomalies from 1880 to 2016'
```
TorsteinHonsi, line 3: Long titles looks better left-aligned on mobile, when the words wrap

samples/highcharts/demo/combo-data-resolution/demo.js
```diff
@@ -0,0 +1,44 @@
...
+    }, {
+        name: 'Line Series',
+        data: JSON.parse(document.getElementById('lineData').textContent),
+        description: `
+            A data series illustrating the same data as the first,
+            but at different time intervals
+        `
+    }],
+    tooltip: {
+        valueSuffix: '°C'
```
TorsteinHonsi, line 28:
```suggestion
        valueSuffix: '°C',
        valueDecimals: 2
```
Five counting digits is a bit over-confident

samples/highcharts/demo/combo-data-resolution/demo.js
```diff
@@ -0,0 +1,44 @@
...
+    },
+    yAxis: {
+        description: 'temperature',
+        max: 1,
+        min: -0.5,
+        name: 'Temperature anomalies',
```
TorsteinHonsi, line 34: A label format that adds the celsius symbol would be nice here

samples/highcharts/demo/combo-data-resolution/demo.js
```diff
@@ -0,0 +1,44 @@
...
+        title: {
+            enabled: false
+        }
+    },
+    xAxis: {
+        endOnTick: true,
+        min: Date.UTC(1880, 0, 0),
```
TorsteinHonsi, line 41: Do we need a hard coded min? It creates trouble when readers start adding their own data instead of ours.

samples/highcharts/demo/combo-data-resolution/demo.js
```diff
@@ -0,0 +1,44 @@
...
+            but at different time intervals
+        `
+    }],
+    tooltip: {
+        valueSuffix: '°C'
+    },
+    yAxis: {
+        description: 'temperature',
+        max: 1,
+        min: -0.5,
```
TorsteinHonsi, line 33: Do we need a hard coded extremes? It creates trouble when readers start adding their own data instead of ours. If what we want to achieve is to use the full height of the plot area, it is better to use dynamic settings:
```suggestion
        minPadding: 0,
        maxPadding: 0,
```

samples/highcharts/demo/combo-data-resolution/demo.details
```diff
@@ -0,0 +1,14 @@
+---
+name: 'Same data, different intervals'
```
TorsteinHonsi, line 2: Technically not the same data (though it's aggregated from the same source). A more descriptive title would be "Same x-axis, different intervals"

samples/highcharts/demo/combo-data-resolution/demo.html
```diff
@@ -0,0 +1,316 @@
...
+        ],
+        [
+            -2335305600000,
+            -0.0974
+        ]
+    ]
+</pre>
+
+<pre id="lineData">
+    [
```
TorsteinHonsi, line 40: This data is sorted descendingly, causing an Highcharts warning 12. Also, it seems to be repeating two by two values.

samples/highcharts/demo/combo-data-resolution/demo.js
```diff
@@ -0,0 +1,47 @@
...
+        description: `
+            A chart of monthly mean temperature anomalies from 1880 to 2016.
+            It employs two series, one column and one line,
+            to convey the same data at different resolutions.
+            There is only one column per thirty years,
+            but the line series have a datapoint per year.
+        `
+    },
+    series: [{
+        type: 'column',
```
TorsteinHonsi, line 16: The columns appear with a fixed width, not reflecting their interval. I am not sure why, but I think it's because the other series is not sorted.

Anyway, we must supply a `pointRange` of 30 years in order for this to render like we expect. I think we should also remove the padding because the 30-year period extends to the whole time span. Something like this:

```js
,
        pointRange: 30 * 365 * 24 * 36e5,
        groupPadding: 0,
        pointPadding: 0,
        pointPlacement: -0.5
```

The `pointPlacement` places the column to the left of the value. So if x is on new-year 2020, the column will span from 1990-2020. Which leads to another issue, namely that the columns are registered 5 years prior, in 2015 etc.

samples/highcharts/demo/combo-data-resolution/demo.js
```diff
@@ -0,0 +1,47 @@
...
+        name: 'Average per thirty years',
+        data: JSON.parse(document.getElementById('columnData').textContent)
```
TorsteinHonsi, line 18: The column series should also have a tooltip format to reflect its time span. Use a format string to render something like `1990-2020`. Use the built-in `subtract` function of the Templating.

samples/highcharts/demo/combo-data-resolution/demo.js
```diff
@@ -0,0 +1,47 @@
...
+    }, {
+        name: 'Monthly average',
+        data: JSON.parse(document.getElementById('lineData').textContent),
+        description: `
+            A data series illustrating the same data as the first,
+            but at different time intervals
+        `
+    }],
+    tooltip: {
```
TorsteinHonsi, line 27: The tooltip for the monthly average shows dates like `31 May, 2019`. Check the timezone. Make sure that the values are actually set on midnight on the first of each month. The resulting tooltip should show something like 'May 2019` with no date or hour specified.

TorsteinHonsi (changes requested):

samples/highcharts/demo/combo-data-resolution/demo.js
MarkusBarstad: In order to subtract 30 years from a timestamp its necessary to do some calculations. It does function now, however the operation looks a bit messy:

```
tooltip: {
    headerFormat: '<span>{point.x:%Y} ' +
        '{(subtract point.x (multiply 30 ' +
        '(multiply 60 (multiply 1000' +
        '(multiply 60 (multiply 365.25 24)))))):%Y}</span>'
}
```

Can and should we use custom variables in templating to make this more readable?

TorsteinHonsi: I think we can just use the number of milliseconds and compensate with a comment:
```js
            // Subtract 30 years from the year to get the start of the period
            headerFormat:
                '<span>{(subtract point.x 946684800000):%Y} - {point.x:%Y}<br>'
```

TorsteinHonsi (changes requested):
Almost ready to merge now. But what's the 0 level of the y-axis? Is it the pre-industrial average? We need to add that in the y-axis title because otherwise the columns don't make sense.

TorsteinHonsi (approved):

PR #20411, opened 2024-01-10
bugfix/20231-map-point-name

### LINKED ISSUE #20231: The `point.update` method doesn't affect the point's name
#### Expected behaviour
The `point.update` method should update the `name` property on the point's instance.

#### Actual behaviour
The `point.update` method doesn't update the `name` property on the point's instance.

https://jsfiddle.net/BlackLabel/n15oxqay/

#### Workaround
Refer to the `point.options.name` property instead.
https://jsfiddle.net/BlackLabel/03ft91gy/

#### Product version
Highmaps JS v11.2.0

magdalena-gut:
Fixed #20231, if `mapPoint.name` was given, `point.name` didn't take effect

### DIFF
--- a/samples/unit-tests/maps/series-map/demo.js
+++ b/samples/unit-tests/maps/series-map/demo.js
@@ -202,15 +202,30 @@ QUnit.test(
 
 QUnit.test('Null points', function (assert) {
     var chart = Highcharts.mapChart('container', {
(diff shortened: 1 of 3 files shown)


TorsteinHonsi (approved):
Thanks!

PR #20414, opened 2024-01-10
bugfix/20166-columns-gridLines-and-pointPlacement

### LINKED ISSUE #20166: Plot-background and x/y axis groups are misaligned for column series
#### Expected behaviour
`plotBorder` should encapsulate all `gridLines` by default.

#### Actual behaviour
For column series, when one creates a `plotBorder`, a `gridLine` outside of plot space can be seen. This does, however, only appear in utils, not in jsfiddle (there it does not appear in the DOM either)

For the `yAxis` it appears by default, 
for the `xAxis` it only appears when `pointPlacement` is set.

From looking at the values in the DOM it seems like `minPointOffset` or `minPixelPadding` should apply somewhere,
and it seems like x and y axis `groups` are not positioned correctly compared to `plotBackground.`

https://jsfiddle.net/mkb93/qm1v4nxu/

#### Product version
Highcharts 11.2

#### Affected browser(s)
Chromium-based, Firefox, maybe more

MarkusBarstad:
Fixed #20166, columns with `pointPlacement` had a superfluous tick to the left of the chart, visible with `gridLineWidth`.

Note: Firefox headless broke at me so this was committed with `--no-verify` so I could review github-side checks 😬 

This could be a messy addition to otherwise clean code but I am unsure if i should:
- Create a variable at the top of the `axis.render` function which checks for column series with `pointPlacement`
- Create it closer to the check
- Handle it some other way

---
  - https://app.asana.com/0/0/1206008636612601

### DIFF
--- a/ts/Core/Axis/Axis.ts
+++ b/ts/Core/Axis/Axis.ts
@@ -981,7 +981,6 @@ class Axis {
             cHeight = (old && chart.oldChartHeight) || chart.chartHeight,
             cWidth = (old && chart.oldChartWidth) || chart.chartWidth,
             transB = axis.transB;
-
         let translatedValue = options.translatedValue,
             force = options.force,
             x1: number,
@@ -989,7 +988,6 @@ class Axis {
             x2: number,
             y2: number,
             skip: boolean;
-
         // eslint-disable-next-line valid-jsdoc
         /**
          * Check if x is between a and b. If not, either move to a/b
@@ -1030,9 +1028,16 @@ class Axis {
 
             x1 = x2 = Math.round(translatedValue + transB);
             y1 = y2 = Math.round(cHeight - translatedValue - transB);
-            if (!isNumber(translatedValue)) { // No min or max
+
+            if (
+                !axis.horiz ?
+                    (y1 < axisTop || y1 > axisTop + axis.height) :
+                    (x1 < axisLeft || x1 > axisLeft + axis.width) ||
+                // #7175, don't force it when path is invalid
+                !isNumber(translatedValue)
+            ) {
                 skip = true;
-                force = false; // #7175, don't force it when path is invalid
+                force = false;
             } else if (axis.horiz) {
                 y1 = axisTop;
                 y2 = cHeight - axis.bottom;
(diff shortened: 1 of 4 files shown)


bre1470:
Good news: It works here with Firefox ESR (v115).

ts/Core/Axis/Axis.ts
```diff
@@ -3934,10 +3934,20 @@ class Axis {
...
                 });
+
                 // In a categorized axis, the tick marks are displayed
                 // between labels. So we need to add a tick mark and
                 // grid line at the left edge of the X axis.
-                if (tickmarkOffset && (axis.min === 0 || axis.single)) {
+                if (
+                    !axis.series.find( // #20166
+                        (series): string | number | false | undefined => (
+                            series.is('column') &&
```
bre1470, line 3944: You should fix this for all cartesian series types. Instead of `series.is(...` check `series.xAxis`.

bre1470 (changes requested):
Looks good so far. This fixes the second issue example.

Things that should be improved:
- broaden the scope of the fix
- a visual test is needed, this properly can be part of an existing sample

ts/Core/Axis/Axis.ts
MarkusBarstad: Good catch! 

It now checks for `series.xAxis` and unit-test has been altered to check for this as well.

Might not have come out clearly in the issue but this is the only example which I was able to replicate both in jsfiddle as well as utils.

I have however uncovered a similar issue for the bar series, but it does seem to arise from the same conditions (`tick[-1]` being rendered for cartesian series using `pointPlacement`)

Should I create a seperate issue for the bar series bug?

bre1470: The bar series is cartesian, but it has inverted axis. This needs taken into account and is probably related.
See https://api.highcharts.com/highcharts/chart.inverted

TorsteinHonsi:
Instead of looking for series with the `pointPlacement` option, I think we should just check that the tick position is inside the axis. That would be simpler and safer.

For the grid line, we have a `force` option for the `getPlotLinePath` function. For some reason it is [set to `pass` for the grid lines](https://github.com/highcharts/highcharts/blob/bc076fb24ea60a8f64945d4c58ff81a525211cee/ts/Core/Axis/Tick.ts#L914). If we comment out that forde parameters, the problematic grid line is not rendered. So I think we should find out why and in what cases we want to force it to render.

For the tick mark we don't seem to have that kind of overflow protection, but we should probably have something similar. Anything that is not between `axis.pos` and `axis.pos + axis.length` should not be rendered.

bre1470:
When running `npx gulp test --browsercount 1` there are now 11 instead of 6 (master) tests failing.

MarkusBarstad:
@TorsteinHonsi 
A couple of things:

So checking that positons are within axis bounds seem to work fine, but we are left with some strange boolean logic for the -1 ticks;

```
                    tickmarkOffset &&
                    (isNumber(axis.max) && -1 <= axis.max) && 
                    (isNumber(axis.min) && -1 >= axis.min) &&
                    (axis.min === 0 || axis.single)
```
namely we check for -1 being more than or equal to axis min, and we check for axis.min being zero or axis.single being true
Any ideas on how to better this logic or is this fine?

Also, as long as the ticks are not rendered in strange places, no strange `gridLine` appears and no "rogue" tickMarks are rendered, so getting the tick-checks right should help us at least a bit with those other artifacts.

> When running `npx gulp test --browsercount 1` there are now 11 instead of 6 (master) tests failing.

Yeah you are right, I think gulp "got stale" or something and did not actually track my changes, allowing a failing commit.

hubertkozik:
The work will be continued in this PR: https://github.com/highcharts/highcharts/pull/20467 Due to many `no-verify` commits here, which can cause issues with future bisecting.

PR #20415, opened 2024-01-10
dash/add-message-on-visual-diff

goransle:
---
  - https://app.asana.com/0/0/1206286261825080

### DIFF
--- a/test/cypress/dashboards/visual.config.mjs
+++ b/test/cypress/dashboards/visual.config.mjs
@@ -55,32 +55,36 @@ export default defineConfig({
 
                     const diffFiles = await walkDir(diffDir);
 
-                    console.log(diffFiles);
-                    console.log(baseFiles);
+                    if (diffFiles.length){
+
+                        console.log(diffFiles);
+                        console.log(baseFiles);
 
 
-                    const diffFilesData = [];
-                    for (const file of diffFiles){
-                        const baseFileName = file.replace('diff', 'base');
-                        const actualFileName = file.replace('diff', 'actual');
+                        const diffFilesData = [];
+                        for (const file of diffFiles){
+                            const baseFileName = file.replace('diff', 'base');
+                            const actualFileName = file.replace('diff', 'actual');
 
-                        if (baseFiles.includes(baseFileName)) {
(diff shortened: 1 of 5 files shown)


test/cypress/dashboards/visual.config.mjs
```diff
@@ -24,6 +42,60 @@ export default defineConfig({
...
+                            ...commonPath
+                        )
+                    ).catch(() => []);
+
+                    const diffFiles = await walkDir(diffDir)
+                        .catch(() => []);
+
+                    if (diffFiles.length){
+
+                        console.log(diffFiles);
```
karolkolodziej, line 66: Do we need it?

test/cypress/dashboards/visual.config.mjs
```diff
@@ -24,6 +42,60 @@ export default defineConfig({
...
             getCompareSnapshotsPlugin(on, config);
+
+            if(config.env.type === 'actual') {
+                on('after:run', async (results) => {
+                    // create json of files in the diff directoryA
+                    const commonPath = ['test', 'cypress', 'dashboards',  'visual'];
+
+                    const diffDir = join(config.env.SNAPSHOT_DIFF_DIRECTORY, ...commonPath);
+
+                    // TODO: try to figure out why these files are saved to this bonkers place
```
karolkolodziej, line 53: Is it going to be solved here or in the future?

karolkolodziej (approved):

karolkolodziej:
@goransle any idea why we get the diffs in this PR?

goransle:
> @goransle any idea why we get the diffs in this PR?

Likely due to setting the viewport sizes within the tests (and then not waiting for the dashboard to adjust). Might be better for stability to use [this approach](https://docs.cypress.io/api/commands/viewport#Set-viewport-in-the-test-configuration), but then we might miss testing dynamic responsiveness 🤔 If we want to keep that aspect we will have to find a way to wait for the dashboard to resize.

karolkolodziej:
We are going to keep an eye on that 😉 thanks!

goransle:
Note that the comparison is failing on this PR as the config and plugin version differs from master (aka the path to the base snapshot has changed).

PR #20418, opened 2024-01-11
Refactored scrollablePlotArea

TorsteinHonsi:
Refactored `scrollablePlotArea` to use a class rather than extending the Chart.

#### Rationale
* Avoid polluting the Chart object with module-specific properties and methods
* Clean up typing
* Some smarter ways of doing things

 #### To do
 - [x] Find a true indication of `firstTime`
 - [x] See if `scrollablePlotBox` can be moved to the ScrollablePlotArea class
 - [x] Why is `scrollablePlotBox` set on the SVGRenderer?
 - [x] In the `afterSetSize` static function, see if the block modifying scrollablePlotBox, plotBox and clipBox should be moved to `applyFixed`.
 - [x] In the `afterSetSize` static function, see if the `Axis.getPlotLinePath` override can be removed and `getPlotLinePath` simply use `chart.scrollablePixelsX` / `…Y`
 - [x] Rename the properties starting with `scrolling…`. The prefix is no longer needed now that we have context.

### DIFF
--- a/samples/unit-tests/scrollable-plotarea/dynamics/demo.js
+++ b/samples/unit-tests/scrollable-plotarea/dynamics/demo.js
@@ -27,7 +27,9 @@ QUnit.test('Test dynamic behaviour of Scrollable PlotArea', function (assert) {
     chart.setTitle({ text: 'New title' });
 
     assert.ok(
-        chart.fixedRenderer.box.contains(chart.title.element),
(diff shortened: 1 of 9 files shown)


bre1470 (approved):
Fantastic! Looks good. 👍

One design suggestion: If you add the external event listeners as static class methods, you add a workaround ability to replace them before calling compose.

PR #20421, opened 2024-01-11
demo/all-maps-demo-upgrade

hubertkozik:
Upgraded all-maps demo, removed jQuery, used datalist element instead of jQuery combobox.

### DIFF
--- a/samples/maps/demo/all-maps/demo.js
+++ b/samples/maps/demo/all-maps/demo.js
@@ -2,178 +2,108 @@
  * This is a complicated demo of Highcharts Maps, not intended to get you up to
  * speed quickly, but to show off some basic maps and features in one single
  * place. For the basic demo, check out
- * https://www.highcharts.com/maps/demo/geojson instead.
+ * https://www.highcharts.com/demo/maps/tooltip instead.
  *
- * @todo
- * - Remove jQuery where not necessary
  */
 
+// Get all HTML elements
+const datalist = document.getElementById('maps'),
+    input = document.getElementById('map-datalist-input'),
+    prevMapButton = document.getElementById('prev-map-btn'),
+    nextMapButton = document.getElementById('next-map-btn'),
+    dataLabelsCheckbox = document.getElementById('datalabels-checkbox'),
+    mapNameHeader = document.getElementById('map-name-header'),
+    cleanDemoButton = document.getElementById('clean-demo-btn'),
+    svgLink = document.getElementById('svg-link'),
+    geojsonLink = document.getElementById('geojson-link'),
+    topojsonLink = document.getElementById('topojson-link'),
+    javascriptLink = document.getElementById('javascript-link');
+
 // Base path to maps
-const baseMapPath = 'https://code.highcharts.com/mapdata/';
+const baseMapPath = 'https://code.highcharts.com/mapdata/',
+    options = [], // Options elements
+    allMaps = {};
 
-let showDataLabels = false, // Switch for data labels enabled/disabled
-    mapCount = 0,
-    mapOptions = '';
-
-// Populate dropdown menus and turn into jQuery UI widgets
-$.each(Highcharts.mapDataIndex, (mapGroup, maps) => {
+// Populate dropdown options
+for (const [mapGroup, maps] of Object.entries(Highcharts.mapDataIndex)) {
     if (mapGroup !== 'version') {
-        mapOptions += `<option class="option-header">${mapGroup}</option>`;
-        $.each(maps, (desc, path) => {
-            mapOptions += `<option value="${path}">${desc}</option>`;
-            mapCount += 1;
-        });
+        Highcharts.merge(true, allMaps, maps);
     }
-});
-const searchText = `Search ${mapCount} maps`;
-mapOptions =
-    `<option value="custom/world.js">${searchText}</option>${mapOptions}`;
-$('#mapDropdown').append(mapOptions).combobox();
+}
 
-// Change map when item selected in dropdown
-$('#mapDropdown').on('change', async function () {
-    const $selectedItem = $('option:selected', this),
-        mapDesc = $selectedItem.text(),
-        mapKey = this.value.slice(0, -3),
-        svgPath = baseMapPath + mapKey + '.svg',
-        geojsonPath = baseMapPath + mapKey + '.geo.json',
-        topojsonPath = baseMapPath + mapKey + '.topo.json',
-        javascriptPath = baseMapPath + this.value,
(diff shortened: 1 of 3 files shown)


samples/maps/demo/all-maps/demo.js
```diff
@@ -2,180 +2,234 @@
...
+                type: 'MultiPoint',
+                coordinates: [
+                    // Alaska west
+                    [-164, 54],
+                    // Greenland north
+                    [-35, 84],
+                    // New Zealand east
+                    [179, -38],
+                    // Chile south
+                    [-68, -55]
```
KacperMadej, line 218: This was copied over but doesn't look like it was working or now is not working correctly as the initial map seems to be clipped on the side edges.

KacperMadej (changes requested):
Great improvement 🎉 

There are two potential improvements for index.js that would make this demo work even better - https://github.com/highcharts/map-collection-dist/issues/138, https://github.com/highcharts/map-collection-dist/issues/139 - but those could be fixed later as they don't look like critical issues.

samples/maps/demo/all-maps/demo.js
hubertkozik: Hmm. funny thing: the current state on the demo pages: https://www.highcharts.com/demo/maps/all-maps The initial map name is `World, Miller projection, medium resolution`, but it is presented in Equal Earth projection, that's why `fitToGeometry` was used. I removed `fitToGeometry` from the code, to keep the cohesion with map name from Map Collection.

KacperMadej (approved):

TorsteinHonsi (changes requested):
Thanks!

The code quality is top notch, so no comments there. Some comments on the actual GUI result. Comparing to the [original demo](https://www.highcharts.com/demo/maps/all-maps).

- [x] * If I click a country to drill down, the World map in the drilldown is titled "World, Miller projection, medium resolution". I think that's a bit overkill, we should simply call it "World" like in the original. The projection and resolution is not so interesting in this context, right?
- [x] * Related, click a country then drill up again to the world. Now click the right arrow next to the combo box. There are a number of maps (world maps with specific projection) that don't exist as TopoJSON (because TopoJSON isn't projected). Those should be skipped.
- [ ] * In the original, the right-side panel isn't visible until users start interacting with the map. Maybe we should keep that idea to keep the first impression clean?
- [ ] * When toggling data labels, bad animation of the chart occurs. I reportad an issue on that (#20489), but if we don't prioritize that we can still in the demo just run the update without animation like we used to do.
- [ ] * Maps with built-in projections. When I click Norway, the Norway/Svalbard map loads with correct projection. However when I search for Norway, the embedded projection is not applied. It is too top heavy. looks like the World map's Miller projection still applies.
- [ ] * When I click on Canada, the North America map appears.
- [ ] * In the original, when I search for a country, the navigation still works (arrow up to World). In the PR, this functionality is broken. When I search for a country, only the arrow left and right are presented. Don't know if this is crucial, but I think it's a navigation aid. Maybe the search should actually trigger a drilldown if it matches a country?
- [x] * Styling of the "View clean demo" button - should be styled according to our website design manual.
- [x] * Set `breadcrumbs.floating` to true to avoid too much going on in the animation?

hubertkozik:
Regarding a few things:
- overkill map names -> I agree, projection is not interesting, but I would leave the resolution to present, that we offer more detailed world maps, what do you think? If you don't agree we can add those maps to `mapsToSkip` and leave one map for the world.
- I would show the right panel at the start, for me (from a user perspective) it is clearer to see that UI from the beginning, also, if someone would like to check the world map in different formats (SVG, etc.) he can do it at the start.

TorsteinHonsi (changes requested):
Thanks so far! Some more issues:

* Projection. There are still quirks. When clicking Norway, the projection is still wrong (still Miller). However when clicking Canada, it uses good projection. Search for Antarctica, it now also uses the wrong projection compared to the online version of `all-maps`. I think the problem is the way the projection is applied in the demo. In Highcharts, if the map doesn't include a recommended projection, it has a function to [pick one that is suitable](https://github.com/highcharts/highcharts/blob/f25c9c45bffa3bf99ec46b8ce0edaa67a09b5938/ts/Maps/MapView.ts#L314-L329). In short, that is EqualEarth for views close to the whole world, and LCC for smaller parts of the world. As far as possible, the demo should avoid being deterministic and leave it up the the MapView to pick the projection. That probably implies that the existing projection must be removed before we add the drilldown series.
* All data labels are blue and underlines as if they were clickable, even if they are not. It looks like the drilldown module styles everything like a clickable point? If this is too hard to debug, we should just remove the blue color and the underline for all labels, drillable or not.
* The "View clean demo" button should have a `target="_blank"`. This problem exists also in the production version of the demo, that now also runs inside an iframe.
* Mobile view. The production demo has a responsive layout where the controls are rendered below the chart on mobile. Can also consider putting the color axis horizontally below the world map.

hubertkozik:
Regarding:

- projection: The Code responsible for suitable projection is not fired up after drill down, that's why the maps after drilldown and the map for map-collection (clean demo) look a little bit different. [Picking suitable projection](https://github.com/highcharts/highcharts/blob/f25c9c45bffa3bf99ec46b8ce0edaa67a09b5938/ts/Maps/MapView.ts#L314-L329) code is fired only when the map view is being initialized (constructor). I think, to make it work after drilldown, I should copy-paste calculating geobounds and picking projection to the demo code, but then the code of the demo would be longer. What do you think, should I copy-paste that code to drilldown function in the demo?
- Mobile view. I have added CSS media queries, to make it work exactly as on the current stage (production), but I don't know why CSS media queries are not working in Mobile View in Utils. You can check it here: https://jsfiddle.net/BlackLabel/6uwend93/ that it works correctly.

TorsteinHonsi:
@hubertkozik 

> [Picking suitable projection](https://github.com/highcharts/highcharts/blob/f25c9c45bffa3bf99ec46b8ce0edaa67a09b5938/ts/Maps/MapView.ts#L314-L329) code is fired only when the map view is being initialized (constructor).

I think this is a case when our own demos reveal a weakness in the product. I would expect that either the recommended projection were updated, or at least that we could access this code and apply it somehow. What do you think? What if we refactor out that the recommended projection logic so we can call it from the demo? Or should it be fully automatic, so that when we do a drilldown or swap series, it recomputes the recommended projection?

TorsteinHonsi:
> I would say, that it should be fully automatic. 

Perfect, I agree. Feel free to isolate it into a separate issue so that we don't bloat this PR.

hubertkozik:
This Pull Request is blocked by https://github.com/highcharts/highcharts/pull/20609.

PR #20434, opened 2024-01-12
bugfix/improve-fix-toogling-inverted-packedbubble

hubertkozik:
Improved `series.invertible` logic due to failed tests in https://github.com/highcharts/highcharts/pull/20328.

### DIFF
--- a/ts/Core/Series/Series.ts
+++ b/ts/Core/Series/Series.ts
@@ -313,8 +313,6 @@ class Series {
 
     public initialType?: string;
 
-    public invertible: boolean = true;
-
     public isDirty?: boolean;
 
     public isDirtyData?: boolean;
@@ -4172,7 +4170,6 @@ class Series {
             preserve = [
                 'colorIndex',
                 'eventOptions',
-                'invertible',
                 'navigatorSeries',
                 'symbolIndex',
                 'baseSeries'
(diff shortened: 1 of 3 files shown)


ts/Core/Series/Series.ts
```diff
@@ -4300,6 +4300,12 @@ class Series {
                 // #20264: Re-detect a certain chart properties from new series
                 chart.propFromSeries();
 
+                series.invertible = seriesTypes[newType].prototype.invertible;
+
+                if (!series.invertible) {
+                    preserve.splice(preserve.indexOf('invertible'), 1);
+                }
+
```
TorsteinHonsi, line 4308: I don't quite get this. Now that you correctly moved the `invertible` property to the prototype instead of the instance, I expected this to sort itself out. I expected that we could remove `invertible` from the `preserve` array, and every series regardless of update, remove or destroy would have the correct value of `invertible` from its prototype....

hubertkozik: I've added this code for this kind of situation: https://jsfiddle.net/BlackLabel/ya9sjxLv/ Init load bar chart (inverted) and then update the series to the packedbubble and then update chart to inverted. If you remove those lines, then the invertible in `preserve` variable will take over and the logic will be broken: https://jsfiddle.net/BlackLabel/06zjL4wo/

hubertkozik: And if we remove the invertible from `preserve` array, then the `[unit-tests/series-xrange/xrange](http://utils.highcharts.local/samples/compare-view?path=unit-tests/series-xrange/xrange&dummy=1705330261509)` is failing.

TorsteinHonsi (approved):

PR #20436, opened 2024-01-12
bugfix/20318-navigator-series-datasorting-empty-data

### LINKED ISSUE #20318: `Navigator` series not rendered on initial load with `dataSorting` enabled
#### Expected behaviour
Navigator series is displayed on initial load with `dataSorting` enabled.

#### Actual behaviour
Navigator series is not displayed on initial load with `dataSorting` enabled.

[https://jsfiddle.net/BlackLabel/p03895tu/](https://jsfiddle.net/BlackLabel/p03895tu/)

#### Product version
Highcharts Stock since v8 (since adding the data sorting)

jedrzejruta:
Fixed #20318, `Navigator` series data wasn't set with `dataSorting` enabled on initial run.

### DIFF
--- a/samples/unit-tests/series/datasorting/demo.js
+++ b/samples/unit-tests/series/datasorting/demo.js
@@ -3,6 +3,9 @@ QUnit.test('Data sorting ', function (assert) {
             chart: {
                 animation: false
             },
+            navigator: {
+                enabled: true
+            },
             series: [
                 {
                     type: 'column',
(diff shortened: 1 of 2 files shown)


pawelfus (commented):
Q: Why can't we reuse `setSortedData()` method? Or to put it in different words, why that method does not set data for series with 'sorted' option in Navigator?

jedrzejruta:
> Q: Why can't we reuse `setSortedData()` method? Or to put it in different words, why that method does not set data for series with 'sorted' option in Navigator?

When you mentioned using `setSortedData()`, it seems that it's better to reuse this method in this case.

pawelfus (approved):
Looks good. I think it shouldn't affect the performance - the navigator needs this, and other series should already have `points` defined

TorsteinHonsi (approved):
Thanks!

PR #20437, opened 2024-01-14
refactor/htmlelement

TorsteinHonsi:
Refactored the HTMLElement to a true class pattern.

#### Rationale
* It was not really a class. It was defined as a class, but in practice just copied functions over to the SVGElement.
* The HTMLRenderer was not really a class. Like HTMLElement, it just copied properties to SVGRenderer. Deleted it and moved the composition over to HTMLElement.
* The HTMLRenderer used to set functions on the HTML _instances_. Moved these over to the HTMLElement class instead.
* There were some deeply nested functions.
* Prepare the ground for an upcoming implementation of `foreignObject`


bre1470 (approved):
Looks good. 👍

PR #20447, opened 2024-01-16
bugfix/20329-accessibility-memory-leak

### LINKED ISSUE #20329: Accessibility | Memory leak when accessibility module is linked
Use case: 
We have a page with 15-20 charts that are being refreshed every 3seconds
We are using the official "highcharts-react-official" react wrapper (but I reproduced this with vanilla HC as well)

#### Expected behavior
Dashboard should continue working for a long period of time (>12h)

#### Actual behavior
Dashboard crashes after 30 minutes (become not responsive even earlier)
CPU @ 100%, Dom nodes count is around 400000, JS heap ~ 1GB

I will not post JSFiddle here because it loads the result inside iframe and its harder to trace
Take this code, and paste it inside a HTML page

HTML head
```
      <script src="https://code.highcharts.com/10.0.0/highcharts.js"></script>
      <script src="https://code.highcharts.com/10.0.0/modules/accessibility.js"></script>
      <script src="https://code.highcharts.com/modules/accessibility.js"></script>

```

HTML body
   
```
   <div id="charts-container"></div>

    <script type="text/javascript">
        const random = Math.random;

        for (let i = 0; i < 100; i++) {
            const chartDiv = document.createElement('div');
            document.getElementById('charts-container').appendChild(chartDiv);

            Highcharts.chart(chartDiv, {

                chart: {
                    plotBorderWidth: 5
                },

                series: [{
                    type: 'pie',
                    data: [2, 5, 2, 3, 6]
                }],
                accessibility: {
                    enabled: true
                }

            }, chart => {
                setInterval(function() {
                    chart.series[0].setData([random(), random(), random(), random(), random()]);
                    chart.update({
                        chart: {
                            plotBorderColor: '#' + Math.floor(random() * 16777215).toString(16)
                        }
                    });
                }, 1000)
            });
        }
    </script>
```

**let it run for about 10 minutes**
you will see that the number of nodes is growing rapidly, and the heap and CPU are growing as well.
and after about 30 minutes (number of nodes reaching 2 million, CPU @ 100% ) **the page will crash!**

After investigating this, I saw that there are lots of detached divs with this text "End of interactive chart" staying in the memory.
I guess things like this are the source of the memory leak.

I you remove Accessibility module, everything is working fine for days

#### Product version
 Highcharts 11.2
 

#### Affected browser(s)
Chrome, Edge, FF

hubertkozik:
Fixed #20329, detached elements and events in the Accessibility module produced a memory leak.

---
  - https://app.asana.com/0/0/1206271597548466


marvin19 (approved):

ts/Accessibility/ProxyProvider.ts
```diff
@@ -288,6 +288,24 @@ class ProxyProvider {
     public removeGroup(groupKey: string): void {
         const group = this.groups[groupKey];
         if (group) {
+            const { elements } = this.domElementProvider,
```
oysteinmoseng, line 291: The purpose of the `domElementProvider` is exactly to keep track of elements to be removed, so this functionality should probably be on a new `domElementProvider.removeElement` method, if we need to remove one element, but not all.

ts/Accessibility/KeyboardNavigation.ts
```diff
@@ -460,62 +460,71 @@ class KeyboardNavigation {
...
+
+        // Do not add event to the element if event is already added to prevent
+        // memory leak (#20329).
+        if (
+            (element as any).hcEvents &&
+            !(element as any).hcEvents[eventName]
+        ) {
+            this.eventProvider.addEvent(
+                element,
+                eventName,
```
oysteinmoseng, line 474: I think we should aim to fix this earlier, and not make an element that is already an exit anchor into an exit anchor again. So in `makeElementAnExitAnchor`, do a check if the element has already been assigned the class `highcharts-exit-anchor`, and if so, abort. Untested, but would be cleaner if that works.

oysteinmoseng (changes requested):
Nice investigation, just some comments on the fix

ts/Accessibility/KeyboardNavigation.ts
hubertkozik: That one check if the element has already been assigned the class `highcharts-exit-anchor` didn't work well... I've changed solution a little bit. I am unhappy with `any`s, but it would easy to leave it as `any`.

ts/Accessibility/ProxyProvider.ts
hubertkozik: Added `removeElement` method to `domElementProvider.removeElement` 👍

ts/Accessibility/KeyboardNavigation.ts
```diff
@@ -445,8 +446,23 @@ class KeyboardNavigation {
...
-            this.exitAnchor.parentNode.removeChild(this.exitAnchor);
+        // Remove event from element and from eventRemovers array to prevent
+        // memory leak (#20329).
+        if (this.exitAnchor) {
+            if (defined((this.exitAnchor as any).focusEventRemover)) {
+                const i =
+                this.eventProvider.eventRemovers.indexOf(
+                    (this.exitAnchor as any).focusEventRemover);
+
+                this.eventProvider.eventRemovers[i]();
```
oysteinmoseng, line 457: We should just add an `eventProvider.removeEvent()` function that handles this I think, like we have `domElementProvider.removeElement()`.

oysteinmoseng (changes requested):
All good now I think, just a cleanup thing

TorsteinHonsi (changes requested):
Thanks! Can get rid of the any casts on `focusEventRemover`? 

I propose either that we extend the HTMLElement definition to include this property, which is generally not good practice, or that we create some lookup table between the elements and the focus removers. Or maybe other proposals?

PR #20448, opened 2024-01-16
dash/live-weather-demo

jomar-honsi:
Dashboard with map, KPI, spline and datagrid.
Actual measurements and forecast for selected weather stations.


samples/dashboards/demo/live-weather/demo.js
```diff
@@ -0,0 +1,969 @@
...
+    }
+};
+
+// Geolocation info: https://www.maps.ie/coordinates.html
+
+// Three sets of locations are available:
+//  - worldLocations, testLocations, euroLocations
+//  - selected in "weatherStations.location"
+
+// eslint-disable-next-line no-unused-vars
```
karolkolodziej, line 67: We are using the `worldLocations` so there is no need to disable this. 
But if there is no dynamic toggle between the sets I would remove the rest.

karolkolodziej (changes requested):
Nice demo but a few comments:
1. KPI isn't resized properly:
image
2. Why time column have a different font?
3. You should be able to click twice on the same mapmarker.
4. Why this color set? It doesn't work well in the dark mode.

samples/dashboards/demo/live-weather/demo.js
jomar-honsi: I was in two minds here; remove the euro and test data sets, or add a dropdown to select one of them. "Cheapest" option is to leave only the world set, perhaps with a few extra cities.

jomar-honsi:
> Nice demo but a few comments:
> 
> 1. KPI isn't resized properly:
>    image
> 2. Why time column have a different font?
> 3. You should be able to click twice on the same mapmarker.
> 4. Why this color set? It doesn't work well in the dark mode.

1. I will fix it.
2. The time value overflows on mobile view. A way to make the font-size responsive?
3. I will investigate
4. I can revert to the colours used in the climate-all sample.

Vazonik (changes requested):
Nice! Few comments:
1. It would be good if the map chart had a title that was updated after clicking on the KPI, so that it is immediately clear what the numbers displayed in the points mean.
2. Could you add a pointer cursor to all clickable components so that they are known to be clickable?

samples/dashboards/demo/live-weather/demo.js
jomar-honsi: > We are using the `worldLocations` so there is no need to disable this. But if there is no dynamic toggle between the sets I would remove the rest.

Removed.

jomar-honsi:
> Nice! Two comments:
> 
> 1. It would be good if the map chart had a title that was updated after clicking on the KPI, so that it is immediately clear what the numbers displayed in the points mean.
> 2. Could you add a pointer cursor to all clickable components so that they are known to be clickable?

@Vazonik Thanks for the suggestions, both are implemented.

samples/dashboards/demo/live-weather/demo.css
```diff
@@ -0,0 +1,394 @@
...
+    display: inline-block;
+    margin-right: 0.25em;
+    vertical-align: middle;
+}
+
+#kpi-data .highcharts-dashboards-component-kpi-value::before {
+    content: url(data:image/svg+xml;base64,PHN2ZyB3aWR0aD0iMzIiIGhlaWdodD0iMzMiIHZpZXdCb3g9IjAgMCAzMiAzMyIgZmlsbD0ibm9uZSIgeG1sbnM9Imh0dHA6Ly93d3cudzMub3JnLzIwMDAvc3ZnIj4KPHBhdGggZD0iTTMxLjc0NTggMjkuNDE4OEwxNy4zNDYgMS40MTg3NUMxNy4wNTE1IDAuODQ2MjUgMTYuNTQ0NSAwLjUgMTYgMC41QzE1LjQ1NTUgMC41IDE0Ljk0ODUgMC44NDYyNSAxNC42NTQgMS40MTg3NUwwLjI1NDE1NSAyOS40MTg4QzAuMDk4NzU4MSAyOS43MjA5IDAuMDExMzIwOCAzMC4wNjk4IDAuMDAxMDI3MTcgMzAuNDI4N0MtMC4wMDkyNjY0NiAzMC43ODc3IDAuMDU3OTYzNSAzMS4xNDM1IDAuMTk1NjU1IDMxLjQ1ODdDMC4zMzM0MTEgMzEuNzczOSAwLjUzNjUxOCAzMi4wMzcgMC43ODM2NDggMzIuMjIwMkMxLjAzMDc4IDMyLjQwMzQgMS4zMTI4MSAzMi41IDEuNjAwMTQgMzIuNUgzMC4zOTk5QzMwLjk4NTQgMzIuNSAzMS41MjM4IDMyLjEwMDYgMzEuODA0MyAzMS40NTgxQzMxLjk0MiAzMS4xNDMgMzIuMDA5MyAzMC43ODczIDMxLjk5OSAzMC40Mjg0QzMxLjk4ODcgMzAuMDY5NSAzMS45MDEyIDI5LjcyMDggMzEuNzQ1OCAyOS40MTg4Wk0xNiA2LjE5ODc1TDIwLjI2OTUgMTQuNUgxNkwxMi44IDE4LjVMMTAuODk3MSAxNi4xMjEzTDE2IDYuMTk4NzVaIiBmaWxsPSIjOUE5MkNFIi8+Cjwvc3ZnPgo=);
```
sebastianbochan, line 245: The same like above

samples/dashboards/demo/live-weather/demo.html
```diff
@@ -0,0 +1,44 @@
+<script src="https://code.highcharts.com/dashboards/dashboards.js"></script>
+<script src="https://code.highcharts.com/dashboards/datagrid.js"></script>
+<script src="https://code.highcharts.com/highcharts.js"></script>
+<script src="https://code.highcharts.com/highcharts-more.js"></script>
+<script src="https://code.highcharts.com/maps/modules/map.js"></script>
+<script src="https://code.highcharts.com/modules/solid-gauge.js"></script>
+<script src="https://code.highcharts.com/modules/timeline.js"></script>
```
sebastianbochan, line 7: Do we need this series?

samples/dashboards/demo/live-weather/demo.js
```diff
@@ -0,0 +1,917 @@
...
+                            for (const [name, item] of Object.entries(weatherStations.location.points)) {
+                                ret.push([name, item.lat, item.lon, item.alt]);
+                            }
+                            return ret;
+                        }
+                    }
+                }
+            ]
+        },
+        // Copied from https://www.highcharts.com/demo/dashboards/climate)
```
sebastianbochan, line 249: We can remove it ;)

samples/dashboards/demo/live-weather/demo.js
```diff
@@ -0,0 +1,917 @@
...
+                    time: 'x',
+                    temperature: 'y'
+                },
+                sync: {
+                    highlight: true
+                },
+                chartOptions: {
+                    chart: {
+                        spacing: [40, 40, 40, 10],
+                        styledMode: true,
```
sebastianbochan, line 653: styledMode is enabled by default in all `HighchartsComponents`

samples/dashboards/demo/live-weather/demo.js
```diff
@@ -0,0 +1,917 @@
...
+                        y: 5
+                    },
+                    tooltip: {
+                        enabled: true,
+                        stickOnContact: true,
+                        formatter: function () {
+                            const point = this.point;
+
+                            // Date + value
+                            const hdr = Highcharts.dateFormat('%Y-%m-%d %H:%M<br />', point.x);
```
sebastianbochan, line 694: Do we need this extra variable? Its used only once.

sebastianbochan (changes requested):
What worries me, it's a lot of custom code. Our purpose was to create a simplified demo as much as possible.

samples/dashboards/demo/live-weather/demo.html
jomar-honsi: This series contains the geographical information (index 1 on the map) so I can't see how it can be removed. I could however do away with the JSON format and put the info directly in the data and thus save some code.

samples/dashboards/demo/live-weather/demo.js
jomar-honsi: Fixed.

samples/dashboards/demo/live-weather/demo.css
```diff
@@ -0,0 +1,394 @@
...
+#kpi-layout .highcharts-title {
+    font-size: 1em;
+}
+
+#kpi-data .highcharts-dashboards-component {
+    cursor: default;
+}
+
+#kpi-data .highcharts-dashboards-component-title::before {
+    content: url(data:image/svg+xml;base64,PHN2ZyB3aWR0aD0iMzIiIGhlaWdodD0iMzMiIHZpZXdCb3g9IjAgMCAzMiAzMyIgZmlsbD0ibm9uZSIgeG1sbnM9Imh0dHA6Ly93d3cudzMub3JnLzIwMDAvc3ZnIj4KPHBhdGggZD0iTTE0LjczNTIgMzEuODU0NEM1LjQyMTIzIDE4LjY4OTQgMy42OTIzOCAxNy4zMzgzIDMuNjkyMzggMTIuNUMzLjY5MjM4IDUuODcyNTYgOS4yMDI3IDAuNSAxNi4wMDAxIDAuNUMyMi43OTc0IDAuNSAyOC4zMDc4IDUuODcyNTYgMjguMzA3OCAxMi41QzI4LjMwNzggMTcuMzM4MyAyNi41Nzg5IDE4LjY4OTQgMTcuMjY0OSAzMS44NTQ0QzE2LjY1MzcgMzIuNzE1MiAxNS4zNDY0IDMyLjcxNTIgMTQuNzM1MiAzMS44NTQ0Wk0xNi4wMDAxIDE3LjVDMTguODMyMyAxNy41IDIxLjEyODMgMTUuMjYxNCAyMS4xMjgzIDEyLjVDMjEuMTI4MyA5LjczODU2IDE4LjgzMjMgNy41IDE2LjAwMDEgNy41QzEzLjE2NzggNy41IDEwLjg3MTkgOS43Mzg1NiAxMC44NzE5IDEyLjVDMTAuODcxOSAxNS4yNjE0IDEzLjE2NzggMTcuNSAxNi4wMDAxIDE3LjVaIiBmaWxsPSIjOUE5MkNFIi8+Cjwvc3ZnPgo=);
```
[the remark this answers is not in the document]
jomar-honsi: Will do.

samples/dashboards/demo/live-weather/demo.html
sebastianbochan: I mean that the reference to the `timeline.js` module is unnecessary :)

jomar-honsi:
> What worries me, it's a lot of custom code. Our purpose was to create a simplified demo as much as possible.

We can do away with the JSON format in the "Cities" series, probably an overkill. This would save some code. It would also be possible to throw away more code but the cost would be duplication of data (e.g. constants like 'hPa' in several places).

karolkolodziej (approved):
Looks good! 
The only small hiccup is that it takes a lot of time to reload the data if you click on the map marker.

jomar-honsi:
> Looks good! The only small hiccup is that it takes a lot of time to reload the data if you click on the map marker.

Perhaps the weather data is reloaded every time the marker is clicked?

samples/dashboards/demo/live-weather/demo.html
jomar-honsi: > I mean that the reference to the `timeline.js` module is unnecessary :)

That makes sense! I thought it would be extremely heavy-handed to remove the entire series :-)

samples/dashboards/demo/live-weather/demo.css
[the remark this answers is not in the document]
jomar-honsi: Fixed now.

samples/dashboards/demo/live-weather/demo.css
jomar-honsi: Done

stitot (commented):
Looks good, but some UX comments:

1. It's not clear for me (consider me a stupid one) what the KPIs are. Averages? Then I believe they should state that, perhaps also with dates/day. "Average temperature next 24H" or something along those lines.
2. What's the min/max value in the gauges? Is gauge the correct component to use?
3. The purple color indicates that something is selected. Makes sense in Map and KPIs, but the Forecast is only "selected by proxy". Something is off there. Perhaps just skip bgcolor on that one?
4. Possible to let selected pin in map be purple, and not entire map? Should then also be purple on hover, as KPIs are.
5. Possible to move the axis description to the title, e.g. "Temperature forecast for Ulaanbatar". Could then be removed from axis. 
6. Legend could also be moved to the title/top, and then removed.
7. Clicking legend produces strange results
8. Elevation KPI appears as clickable as the rest, but is not (obviously ;-). Do we need it?

jomar-honsi:
> Looks good, but some UX comments:
> 
> 1. It's not clear for me (consider me a stupid one) what the KPIs are. Averages? Then I believe they should state that, perhaps also with dates/day. "Average temperature next 24H" or something along those lines.
> 2. What's the min/max value in the gauges? Is gauge the correct component to use?
> 3. The purple color indicates that something is selected. Makes sense in Map and KPIs, but the Forecast is only "selected by proxy". Something is off there. Perhaps just skip bgcolor on that one?
> 4. Possible to let selected pin in map be purple, and not entire map? Should then also be purple on hover, as KPIs are.
> 5. Possible to move the axis description to the title, e.g. "Temperature forecast for Ulaanbatar". Could then be removed from axis.
> 6. Legend could also be moved to the title/top, and then removed.
> 7. Clicking legend produces strange results
> 8. Elevation KPI appears as clickable as the rest, but is not (obviously ;-). Do we need it?
> 
> Suggestion:
> 
> Current elevation KPI could be a HTML component at the top, spanning three remaining KPI. Content could be name of city and elevation, and perhaps additional relevant metadata.

Many good suggestions and most would also apply to the existing climate demo. The bug (#7) must be fixed, most are easy to fix, #2 requires more effort.

1. I agree, it is not immediately clear what the data mean (also in climate demo). Easy to implement.
2. We could consider replacing the gauges: in "classic" meteorology the temperature would be indicated by a column chart with embedded ticks, the pressure by something similar to the HC [Gauge Series](https://www.highcharts.com/demo/highcharts/gauge-speedometer) that is visual very similar to a physical barometer. Humidity could also look like the barometer.
3. Agree: easy to skip bgcolor.
4. I don't think this can be done in the configuration. but should be possible using CSS (xxxx:hover). Should be easy to fix unless extra code is required.
5. Easy fix.
6. Will investigate.
7. Bug: **must be fixed**.
8. It is distinguished by not using a cursor pointer. Could make it appear less "clickable" by giving it a slightly grey bgcolor.

TorsteinHonsi:
Nice work! While we're at it, I was thinking
* Did we consider adding precipitation and wind? Or if we're short on space, maybe replace pressure and humidity with precipitation and wind? In most real-world scenarios, those are more relevant to readers. And it would allow us to show off more chart types - change the main chart to columns for the precipitation and area+windbarb for the wind.
* When I click the KPIs, there is a delay before the dashboard components are updated. I assume we're loading new data from YR asynchronously? Then it would be good UX practice to present some kind of loading indicator to the user. Highcharts has the [showLoading](https://api.highcharts.com/class-reference/Highcharts.Chart#showLoading) method, but maybe the KPI component or the whole dashboard has some other built-in method to use?

jomar-honsi:
> Nice work! While we're at it, I was thinking
> 
> * Did we consider adding precipitation and wind? Or if we're short on space, maybe replace pressure and humidity with precipitation and wind? In most real-world scenarios, those are more relevant to readers. And it would allow us to show off more chart types - change the main chart to columns for the precipitation and area+windbarb for the wind.
> * When I click the KPIs, there is a delay before the dashboard components are updated. I assume we're loading new data from YR asynchronously? Then it would be good UX practice to present some kind of loading indicator to the user. Highcharts has the [showLoading](https://api.highcharts.com/class-reference/Highcharts.Chart#showLoading) method, but maybe the KPI component or the whole dashboard has some other built-in method to use?

Good input! It also fits with input from @stitot to somewhat reshape the KPI component. On input from @sebastianbochan I have just changed the map to North America as a way to make this demo easier to distiguish from the `climate ` demo. This actually saves some space that we could use to display an extra parameter.

Vazonik (changes requested):
Nice! A few more comments below.

jomar-honsi:
New PR created for latest updsates.

PR #20449, opened 2024-01-16
bugfix/20433-scatter-boost-without-min-max

### LINKED ISSUE #20433: Regression: Boosted scatter doesn't render when y-axis extremes not set
#### Expected behaviour
Extremes should be computed from the data.

#### Actual behaviour
When `yAxis.min` or `yAxis.max` are not explicitly set, the series fails to render.

https://jsfiddle.net/highcharts/1905aobL/

#### Product version
v11.3.0

bre1470:
Fixed #20433, boosted scatter chart did not work without axis min/max-options.

~~Fixed #20433; Scatter did not boost without axis min/max-options.~~


marvin19 (approved):

ts/Extensions/Boost/BoostSeries.ts
```diff
@@ -954,11 +958,29 @@ function scatterProcessData(
...
+            yDataMax = Math.max(yDataMax, y);
+            yDataMin = Math.min(yDataMin, y);
         } else {
             cropped = true;
         }
     }
 
+    if (!isNumber(xExtremes.max) || !isNumber(xExtremes.min)) {
+        xAxis.setOptions({
+            max: xAxis.options.max || xDataMax,
```
KacperMadej, line 972: If the check were to be done before the calculation of the xDataMax then it should make the boost run even faster when the axis extremes are set. Should we have some performance tests similar to e.g. the ones for the Stock?

KacperMadej: (preferably in a different task but that would help with comments like mine here)

bre1470: Performance tests are a good idea, especially for boost.

samples/unit-tests/boost/enter-exit/demo.html
```diff
@@ -2,8 +2,4 @@
 <script src="https://code.highcharts.com/modules/boost-canvas.js"></script>
 <script src="https://code.highcharts.com/modules/boost.js"></script>
 
-
-<div id="qunit"></div>
-<div id="qunit-fixture"></div>
-
```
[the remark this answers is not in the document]
bre1470: I did not saw any purpose in the `demo.js` and `karma-setup.js`, and the other boost-tests lack it. Do you know more about it?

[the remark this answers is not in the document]
hubertkozik: This divs contains the UI for tests. Below is ss with added divs. In my opinion, we should keep it.
image

hubertkozik (approved):
OP demo works like a charm now! Approving, but left comments in unresolved conversations.

KacperMadej (approved):

ts/Extensions/Boost/BoostSeries.ts
```diff
@@ -937,11 +937,17 @@ function scatterProcessData(
     // Filter unsorted scatter data for ranges
     const processedData: Array<PointOptions> = [],
         processedXData: Array<number> = [],
-        processedYData: Array<number> = [];
+        processedYData: Array<number> = [],
+        xRangeNeeded = !(isNumber(xExtremes.max) || isNumber(xExtremes.min)),
+        yRangeNeeded = !(isNumber(yExtremes.max) || isNumber(yExtremes.min));
```
TorsteinHonsi, line 942: I think we should look for `..Extremes.dataMax` and `dataMin` instead of the derived `min` and `max`

ts/Extensions/Boost/BoostSeries.ts
```diff
@@ -954,11 +960,32 @@ function scatterProcessData(
...
+                yDataMax = Math.max(yDataMax, y);
+                yDataMin = Math.min(yDataMin, y);
+            }
         } else {
             cropped = true;
         }
     }
 
+    if (xRangeNeeded) {
+        xAxis.setOptions({
```
TorsteinHonsi, line 977: Is there any reason to run `setOptions`? It is cheaper to just set the properties directly, and as far as I can see this works. Also taking into consideration that what we are really looking for is `dataMin` and `dataMax`:

```ts
    if (xRangeNeeded) {
        xAxis.dataMax = xDataMax;
        xAxis.dataMin = xDataMin;
    }
```

TorsteinHonsi (changes requested):
Thanks! In addition to the inline comments, please take a look at [Writing content for the changelog](https://github.com/highcharts/highcharts/tree/master/changelog#writing-content-for-the-changelog), rule no. 5.

ts/Extensions/Boost/BoostSeries.ts
bre1470: I had issues with zoom in/out and boost enter/exit not working as expected.

ts/Extensions/Boost/BoostSeries.ts
```diff
@@ -957,11 +963,28 @@ function scatterProcessData(
...
+                yDataMin = Math.min(yDataMin, y);
+            }
         } else {
             cropped = true;
         }
     }
 
+    if (xRangeNeeded) {
+        xAxis.options.max = xAxis.options.max || xDataMax;
+        xAxis.options.max = xAxis.options.min || xDataMin;
```
TorsteinHonsi, line 981: I think this fails if `options.min` or `max` is 0. So we want a nullish test:

`xAxis.options.max ??= xDataMax;`

TorsteinHonsi (approved):

PR #20457, opened 2024-01-17
demo/stock-black-and-white

kamil-musialowski:
Added black and white Stock demo, with timeframe switch buttons.


samples/stock/demo/stock-design-black-and-white/demo.html
```diff
@@ -0,0 +1,6 @@
+<script src="https://code.highcharts.com/stock/highstock.js"></script>
+<script src="https://code.highcharts.com/modules/price-indicator.js"></script>
```
pawelfus, line 2: The path should be `stock/modules`

samples/stock/demo/stock-design-black-and-white/demo.js
```diff
@@ -0,0 +1,191 @@
...
+            borderWidth: 0,
+            backgroundColor: 'transparent',
+            formatter() {
+                const {
+                    open,
+                    high,
+                    low,
+                    close
+                } = this.point;
+                return `${this.series.name} ${open} ${high} ${low} ${close}`;
```
pawelfus, line 137: Please use `format` instead of `formatter`

pawelfus (changes requested):
Just minor polishing

samples/stock/demo/stock-design-black-and-white/demo.js
```diff
@@ -0,0 +1,191 @@
...
+        chart: {
+            plotBorderWidth: 1,
+            plotBorderColor: '#000000',
+            marginRight: 30
+        },
+        title: {
+            text: ''
+        },
+        rangeSelector: {
+            buttons: [{
```
TorsteinHonsi, line 22: Can we add a [title](https://api.highcharts.com/highstock/rangeSelector.buttons.title) to the buttons? So that the native tooltip and assistive tech will pick it up.

kamil-musialowski: @TorsteinHonsi Yes, of course! What should be the content of such title? Simple 'M15' or more descriptive like '15 minutes'?

kamil-musialowski: Got it, thank you

bm64 (approved):

pawelfus (approved):
Approved, thanks!

pawellysy (approved):
LGTM

samples/stock/demo/stock-design-black-and-white/demo.details
```diff
@@ -0,0 +1,11 @@
+---
+name: Black and white Stock chart
+authors:
+  - Kamil Musialowski
+js_wrap: b
+tags:
+  - Highcharts Stock demo
+categories:
+  - General:
```
pawelfus, line 9: As agreed on: move demo to a new category, called "Designs"

pawelfus (changes requested):

pawelfus (approved):
Thanks!

samples/stock/demo/stock-design-black-and-white/demo.css
TorsteinHonsi, line 1: As with the other design PR - do we need to repeat the word `stock` in the file path? All demos are stock, so nothing special for these new ones. The common denominator is that they are about design.
Skjermbilde 2024-01-30 kl  10 38 14

TorsteinHonsi (changes requested):

TorsteinHonsi (changes requested):
Thanks!

Like in the other design PR, I think we should separate out the part of the config that is pure styling (colors, border radius, line widths etc) and put it in a `setOptions` call. Otherwise it will be hard for implementers that want to continue building from this demo, because the chart config itself is intermixed with style.

kamil-musialowski:
@TorsteinHonsi Moved styling to `setOptions`, if there is anything else that should be moved there, please let me know :)

samples/stock/demo/black-and-white/demo.js
```diff
@@ -0,0 +1,218 @@
...
+                distance: 4,
+                formatter() {
+                    const isAboveD1 =
+                        this.chart.series[0].currentDataGrouping.unitRange >=
+                        86400000;
+
+                    if (this.isFirst || isAboveD1) {
+                        return this.chart.time.dateFormat('%d %b %Y', this.value);
+                    }
+                    return this.chart.time.dateFormat('%d %b %k:%M', this.value);
```
TorsteinHonsi, line 203: I believe the original sketch used time format `%e %b %k:%M`? The difference is the zero-padding on the date.

samples/stock/demo/black-and-white/demo.js
```diff
@@ -0,0 +1,218 @@
...
+                snap: false,
+                label: {
+                    enabled: true,
+                    format: '{value:%d %b %k:%M}'
+                }
+            },
+            startOnTick: true,
+            labels: {
+                distance: 4,
+                formatter() {
```
TorsteinHonsi, line 195: Did you consider using [dateTimeLabelFormats](https://api.highcharts.com/highcharts/xAxis.dateTimeLabelFormats)? It should be able to cover the case with `isAboveD1`, but probably not the `isFirst` case. You might be able to hack `isFirst` through a format string though.

kamil-musialowski: I didn't consider it to be honest due to the `isFirst` check. But I guess, we would prefer that approach instead of callbacks, right?

Could you point me in a right direction with this `isFirst` hack? I don't see how I could do it honestly.

marvin19 (approved):

marvin19 (changes requested):
- add a non-visual chart title ( `lang.accessibility.defaultChartTitle` )
- add a [linkedDescription](https://api.highcharts.com/highcharts/accessibility.linkedDescription)
- Add a title to yAxis, can be visually hidden

samples/stock/demo/black-and-white/demo.js
TorsteinHonsi: Looking closer I don't think it's easy, given that we're checking the current data grouping. We'd end up with custom code anyway, so let's just keep the formatter.

TorsteinHonsi (approved):

samples/stock/demo/black-and-white/demo.js
kamil-musialowski: Cool, thanks!

kamil-musialowski:
Removing @oysteinmoseng from reviewers, we agreed that Torstein's final review is enough.

marvin19 (approved):

PR #20461, opened 2024-01-17
bugfix/20273-special-font

### LINKED ISSUE #20273: Fonts in exported images render as "Lucida Grande" when font-family starts with a number
#### Expected behaviour
Charts render "72-Regular" font for export as an SVG/JPEG/PNG when fontFamily in export config is "72-Regular"

#### Actual behaviour
Charts render "Lucida Grande", "Lucida Sans Unicode" font for export as an SVG/JPEG/PNG when fontFamily in export config starts with a number such as "72-Regular"

#### Product version
Highcharts 10.3.11

magdalena-gut:
Fixed #20273, the `chart.style` documentation updated.

### DIFF
--- a/samples/highcharts/chart/style-special-font/demo.css
+++ b/samples/highcharts/chart/style-special-font/demo.css
@@ -1,0 +1,20 @@
+#container {
+    height: 500px;
+    min-width: 310px;
+    max-width: 800px;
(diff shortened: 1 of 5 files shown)


TorsteinHonsi (approved):
Thanks!

PR #20466, opened 2024-01-18
Fixed #20426, regression when updating a series with zones

### LINKED ISSUE #20426: Zones API is broken in 11.13.0
#### Expected behaviour
The zones API allows you to set the colour of your bars based on the data. For example the following code produces light green bars if the bar has a value below 0 or dark if above:

`zones: [
          {
            value: 0,
            color: '#B6D727',
          },
          { color: '#43B65A' },
        ],`

#### Actual behaviour
In v11.13.0 this now does not work and all bars have a random colour. If I downgrade back to v11.12.0 it works correctly again.

#### Product version

Highcharts v11.13.0

TorsteinHonsi:
Fixed #20426, a regression causing `series.zones` to be lost after `chart.update`.

### DIFF
--- a/samples/unit-tests/series/zones/demo.js
+++ b/samples/unit-tests/series/zones/demo.js
@@ -267,6 +267,19 @@ QUnit.test('Zones and column presentational props (#6234)', assert => {
         '40,30',
         'Zones dash array'
     );
+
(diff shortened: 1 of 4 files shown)


jakubSzuminski (approved):
Everything works as it used to: https://jsfiddle.net/9v6t4acz/ so LGTM!

PR #20467, opened 2024-01-18
bugfix/20166-ticks-visibility-outside-axis

### LINKED ISSUE #20166: Plot-background and x/y axis groups are misaligned for column series

hubertkozik:
Fixed #20166, ticks and gridlines shouldn't be visible outside the axis.

---
  - https://app.asana.com/0/0/1206008636612601


MarkusBarstad (approved):
Awesome work! 

Do you know why this does not break animation? Is it because we here check for whether the tick is new, or is that check there so that we only have to check for this once?

hubertkozik:
Yes, 

> check for whether the tick is new

 is the key, because if the tick already exists then it probably should be re-rendered to new position, new style etc.

bre1470 (approved):
Nice fix! Looks good. 👍

TorsteinHonsi (approved):
Thanks, looks good! I managed to squeeze a few more bytes out, after CI checks I'll merge.

KacperMadej (approved):

PR #20474, opened 2024-01-19
Added initial language pack bundle

goransle:


goransle:
@pawelfus would appreciate it if you could take a look on the RangeSelector changes. 

As an aside I noticed the rangeselector buttons do not expand to the text, added as issue https://github.com/highcharts/highcharts/issues/20545

i18n/lang.json
```diff
@@ -0,0 +1,538 @@
+{
+    "language": "en-US",
+    "chartTitle": "Chart title",
+    "pieSliceName": "Slice",
+    "seriesName": "Series {add series.index 1}",
+    "yAxisTitle": "Values",
+    "rangeSelector": {
+        "monthText": "{count}m",
```
pawelfus, line 8: Q: Should we also include `week`, `day`, `hour`, `minute` and `second`? Those were not defined in the default buttons of RangeSelector, however, those are quite popular in intraday charts (eg [this](https://www.highcharts.com/demo/stock/intraday-breaks) or [this](https://www.highcharts.com/demo/stock/data-grouping))

ts/Stock/RangeSelector/RangeSelector.ts
```diff
@@ -2153,32 +2196,21 @@ extend(RangeSelector.prototype, {
...
         type: 'year',
-        count: 1,
-        text: '1y',
-        title: 'View 1 year'
+        count: 1
     }, {
         type: 'all',
-        text: 'All',
-        title: 'View all'
+        text: 'All'
```
pawelfus, line 2213: It seems `allText` is also defined, shouldn't we use it?

ts/Stock/RangeSelector/RangeSelector.ts
```diff
@@ -440,7 +442,49 @@ class RangeSelector {
...
+            },
+            ytd: {
+                title: chart.options.lang.rangeSelector.YTDTitle,
+                text: chart.options.lang.rangeSelector.YTDText
+            },
+            all: {
+                title: chart.options.lang.rangeSelector.allTitle,
+                text: chart.options.lang.rangeSelector.allText
+            }
+        };
```
pawelfus, line 465: Do we need to create this mapping? Perhaps if we replace `YTDText` (and title counterpart) with `ytdTitle` then a few lines below, it could be simply:
```js
opt.title = chart.options.lang.rangeSelector[`${opt.type}Title`];
```

pawelfus (commented):
Just a few questions, see below

ts/Core/Chart/Chart.ts
```diff
@@ -1186,8 +1186,20 @@ class Chart {
...
+
+            // If title text is unset, and not explicitly unset,
+            // default to lang option
+            if (
+                name === 'title' &&
+                chart.userOptions.title !== null &&
+                !Object.hasOwnProperty.call(options, 'text')
+            ) {
+                text = chart.options.lang.chartTitle;
+            }
```
TorsteinHonsi, line 1199: In general we want to avoid checking the `userOptions` as far as possible, and rely on the merged `options` being accurate. As far as I can see we can make this a little bit cheaper and at the same time avoid checking the user options:

```ts
            // If title text is unset, and not explicitly unset, default to lang
            // option
            const text = options.text ?? (
                name === 'title' && !('text' in options) ?
                    chart.options.lang.chartTitle :
                    void 0 // or null?
            );
```

TorsteinHonsi (changes requested):
Thanks, great work!

In addition to some inline comments, I think while we're working on this we add the relevant information to the [Internationalization article](https://github.com/highcharts/highcharts/blob/095fc94c68e89728f6574efb87addfbe5f985126/docs/advanced-chart-features/internationalization.md#L5), and add a good sample. Ideally the sample would showcase all supported languages through a dropdown or some other kind of language selector.

pawelfus (changes requested):
It seems I can't test it fully:

image


I tried `npx gulp lang-build` but it didn't help :( Are JSON i18n files not part of this PR?

goransle:
> I tried `npx gulp lang-build` but it didn't help :( Are JSON i18n files not part of this PR?

For the JSON files you'll currently need this branch of utils https://github.com/highcharts/highcharts-utils/pull/86

pawelfus:
Extra note:
It seems hints are not translated:
Zrzut ekranu 2024-02-29 o 11 41 20
(tested with: https://jsfiddle.net/BlackLabel/cdLkz9s7/2/ )

pawelfus:
And two extra questions (tested with https://jsfiddle.net/BlackLabel/91mwu05k/3/ ):

Q1: Should point keys in Tooltip also be translated? For example open-high-low-close? See:

image


Q2: Should we translate indicators too? And the names of the algorithms? The latter I'm afraid is not supported..

image

goransle:
> Extra note: It seems browser tooltips/hints are not translated [...]
Seems the rangeselector texts needs to be set on the initial render to work https://jsfiddle.net/s9e12a4g/5/ 🤔

ts/Core/Chart/Chart.ts
```diff
@@ -1205,8 +1205,20 @@ class Chart {
...
+            const text = options.text ?? (
+                key === 'title' &&
+                    // If the title option is set to null,
+                    // it will be an empty object at this point
+                    Object.keys(options).length &&
+                    !('text' in options) ?
+                    chart.options.lang.chartTitle :
+                    void 0
+            );
+
```
TorsteinHonsi, line 1219: Rethinking this, I _think_ we can avoid this whole complex condition by just adding this at the end of the `setOptions` function:

```ts
    if (options.lang?.chartTitle) {
        defaultOptions.title.text = options.lang.chartTitle;
    }
```

It would be shorter, but also way easier to grasp and to maintain.

Unfortunately the `yAxis.title.text` doesn't seem to be that simple, because the default y-axis options are a bit buried. But it may be possible also there, by importing the appropriate object into the Defaults.ts.

i18n/lang.json
```diff
@@ -0,0 +1,543 @@
...
+    ],
+    "weekdays": [
+        "Sunday",
+        "Monday",
+        "Tuesday",
+        "Wednesday",
+        "Thursday",
+        "Friday",
+        "Saturday"
+    ],
```
TorsteinHonsi, line 67: Months, shortMonths and weekdays are now undefined by default, because those strings are picked from the `Intl` object, based on `lang.locale`. The same applies to `decimalPoint` and `thousandsSep`, which means those should be removed from the i18n JSON files.

docs/advanced-chart-features/internationalization.md
```diff
@@ -29,8 +29,53 @@ Since v12, date and number formatting can be applied through the [lang.locale](h
...
+```
+
+Note that the language options have to be set before initializing the chart with `Highcharts.chart()`.
+
+<iframe
+    style="width: 100%; height: 450px; border: none;"
+    src="https://www.highcharts.com/samples/embed/highcharts/demo/i18n"
+    allow="fullscreen"
+></iframe>
+
```
TorsteinHonsi, line 81: Should we add advice how how users can add their own language packs? In short:
1. Ask ChatGPT or Copilot for a translation of the English original
2. Proofread
3. Save and run `gulp lang-build`

samples/highcharts/demo/i18n/demo.js
TorsteinHonsi, line 1: Currently I cannot run this in the utils. The [PR in the utils](https://github.com/highcharts/highcharts-utils/pull/86) is merged into the master. The Highcharts object is not defined. The same with the range selector demo. If I remove the language script, it works. Could it be an issue with webpack building the language scripts?

samples/highcharts/demo/i18n/demo.js
```diff
@@ -0,0 +1,79 @@
...
+                render: function () {
+                    setTimeout(() => this.hideLoading(), 1000);
+                }
+
+            }
+        },
+        xAxis: {
+            type: 'datetime',
+            dateTimeLabelFormats: {
+                month: '%B %Y'
```
TorsteinHonsi, line 18:
```suggestion
                month: '%[BY]'
```

samples/stock/rangeselector/i18n/demo.details
```diff
@@ -0,0 +1,7 @@
+---
+ name: Highcharts Demo
+ authors:
+   - Torstein Hønsi
```
TorsteinHonsi, line 4: Did I? 👀

tools/gulptasks/lang-build.js
```diff
@@ -0,0 +1,117 @@
+/*
+ * Copyright (C) Highsoft AS
+ */
+
+const gulp = require('gulp');
+const log = require('../libs/log');
+
+/**
+ * Crete language ts files from JSON
```
TorsteinHonsi, line 9:
```suggestion
 * Create language ts files from JSON
```

tools/gulptasks/lang-build.js
```diff
@@ -0,0 +1,117 @@
...
+* @module highcharts/i18n/${lang}
+* @requires highcharts
+*
+* ${lang} language pack
+*
+* (c) 2009-2024 Torstein Honsi
+*
+* License: www.highcharts.com/license
+*
+* **Do not edit this file!** This file is generated using the 'gulp i18n-bundle' task.
```
TorsteinHonsi, line 62:
```suggestion
* **Do not edit this file!** This file is generated using the 'gulp lang-build' task.
```

tools/gulptasks/lang-build.js
```diff
@@ -0,0 +1,117 @@
...
+
+
+    for (const langFile of langFiles) {
+        const jsonContent = require(resolve(LANG_DIR, langFile));
+
+        // Consistency checks
+        assert.equal(
+            baseKeys.length,
+            Object.keys(jsonContent).length,
+            `Keys in ${langFile} do not match keys in lang.json`
```
TorsteinHonsi, line 82:
```suggestion
            `Keys in ${langFile} should match keys in lang.json`
```

tools/gulptasks/lang-build.js
```diff
@@ -0,0 +1,117 @@
...
+        );
+
+        assert.deepEqual(
+            getNestedKeys(langBase).sort(),
+            getNestedKeys(jsonContent).sort(),
+            `Keys in ${langFile} do not match keys in lang.json`
```
TorsteinHonsi, line 88:
```suggestion
            `Keys in ${langFile} should match keys in lang.json`
```

ts/Core/Defaults.ts
```diff
@@ -126,6 +126,19 @@ const defaultOptions: DefaultOptions = {
      * ```
      */
     lang: {
+        stockOpen: 'Open',
+        stockHigh: 'High',
+        stockLow: 'Low',
+        stockClose: 'Close',
```
TorsteinHonsi, line 132: Those are specific to Stock and shouldn't weigh down the core.

ts/Core/Defaults.ts
```diff
@@ -256,6 +276,30 @@ const defaultOptions: DefaultOptions = {
...
+        rangeSelector: {
+            allText: 'All',
+            allTitle: 'View all',
+            monthText: '{count}m',
+            monthTitle: 'View {count} {#eq count 1}month{else}months{/eq}',
+            yearText: '{count}y',
+            yearTitle: 'View {count} {#eq count 1}year{else}years{/eq}',
+            ytdText: 'YTD',
+            ytdTitle: 'View year to date'
+        },
```
TorsteinHonsi, line 301: Also specific to Stock

TorsteinHonsi (changes requested):
Thanks again! 

- Should this PR be marked Ready for review?
- There also seems to be a CI failed test related to bread crumbs.

ts/Extensions/DataGrouping/DataGrouping.ts
```diff
@@ -148,11 +148,16 @@ function onTooltipHeaderFormatter(
...
         e.text = format(
             formatString, {
-                point: extend(point, { key: formattedKey }),
-                series: series
+                point: extend(
+                    point,
+                    { key: format(formattedKey, { lang }) }
+                ),
+                series,
+                lang
```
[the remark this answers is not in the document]
goransle: It's for the cases where `lang` is in the option string, as with `lang.weekFrom` in `ts/Extensions/DataGrouping/DataGroupingDefaults.ts`.

samples/highcharts/demo/i18n/demo.js
goransle: Seems it only runs when "Compile on demand" is enabled. Will have to look into why

ts/Core/Defaults.ts
TorsteinHonsi: See https://github.com/highcharts/highcharts/blob/4e64984bbde6fe6c825708fe57d6d98f1e8fb59e/ts/Stock/RangeSelector/RangeSelectorDefaults.ts#L44 for an example of extending `lang` in a module.

ts/Core/Defaults.ts
TorsteinHonsi: See [RangeSelectorDefaults](https://github.com/highcharts/highcharts/blob/4e64984bbde6fe6c825708fe57d6d98f1e8fb59e/ts/Stock/RangeSelector/RangeSelectorDefaults.ts#L44) for an example of extending `lang` in a module.

ts/Core/Chart/StockChart.ts
```diff
@@ -240,8 +238,16 @@ class StockChart extends Chart {
         // Avoid doing these twice
         userOptions.xAxis = userOptions.yAxis = void 0;
 
+        const langOptions = {
+            stockOpen: 'Open',
+            stockHigh: 'High',
+            stockLow: 'Low',
+            stockClose: 'Close'
+        };
+
```
TorsteinHonsi, line 247: 1. I think these language strings should be defined in the HLCSeries file rather than StockChart. (O)HLCSeries and candlestick are standalone series types that can also be applied to a vanilla Chart, so technically the lang options are more tied to the series type than the constructor.
2. The second problem is that is actually seems like it doesn't work: https://jsfiddle.net/highcharts/c7zaLto4/ . The tooltip still shows the English terms.

ts/Core/Defaults.ts
TorsteinHonsi: Still valid as of January 20 - these four keys should be removed from the core.

ts/Core/Defaults.ts
```diff
@@ -2873,6 +2905,13 @@ function setOptions(
         });
     }
 
+    if (options.lang?.chartTitle) {
```
TorsteinHonsi, line 2908: Kan we skip this condition and just do `text: options.lang?.chartTitle`. Because if it's undefined then there's no harm done.

TorsteinHonsi (changes requested):

ts/Core/Defaults.ts
goransle: Not quite with the current setup it seems, as undefined is disabled. We could do this, but I suspect the `if` is cheaper:

```ts
merge(true, defaultOptions.title, {
     text: options.lang?.chartTitle ?? defaultOptions.title?.text
});
```

TorsteinHonsi (approved):
🎉

TorsteinHonsi:
Now pending review from @pawelfus

ts/Stock/RangeSelector/RangeSelector.ts
```diff
@@ -442,7 +444,13 @@ class RangeSelector {
             options = (
                 chart.options.rangeSelector as RangeSelectorOptions
             ),
-            buttonOptions = options.buttons,
+            langOptions = chart.options.lang,
+            buttonOptions = (
+                options.buttons || this.defaultButtons.map(
```
pawelfus, line 449: Q: Where is `this.defaultButtons` array defined? I can't find it. Tested also with GUI demo:
image

ts/Stock/RangeSelector/RangeSelector.ts
```diff
@@ -1450,11 +1473,25 @@ class RangeSelector {
...
+        const langOptions = this.chart.options.lang;
 
         this.buttonOptions.forEach((
             rangeOptions: RangeSelectorButtonOptions,
             i: number
         ): void => {
+            if (rangeOptions.type && langOptions.rangeSelector) {
+                rangeOptions.text ??= langOptions.rangeSelector[`${rangeOptions.type}Text`];
+                rangeOptions.title ??= langOptions.rangeSelector[`${rangeOptions.type}Title`];
+            }
```
pawelfus, line 1485: Does this part seem redundant? Or is it a fix for the missing title translation I reported some time ago? I mean, in lines 473-775 (`init()` method) we have exactly the same logic.

pawelfus (changes requested):
Amazing work! Just two minor things in the Stock part.

Extra note:
Tested creating langpacks, and I wish we get more detailed info, eg which key is missing 😅 File was generated by chatgpt, so now, no idea what is missing..
> AssertionError [ERR_ASSERTION]: Keys in pl-PL.json should match keys in lang.json

A simple change in lang-build:
```ts
        const nestedBaseKeys = getNestedKeys(langBase).sort();
        const translationKeys = getNestedKeys(jsonContent).sort();

        for (let i = 0; i < nestedBaseKeys.length; i++) {
            if (nestedBaseKeys[i] !== translationKeys[i]) {
                console.log(`Key mismatch in ${langFile}: ${nestedBaseKeys[i]} !== ${translationKeys[i]}`);
            }
        }

        assert.deepEqual(
            nestedBaseKeys,
            translationKeys,
            `Keys in ${langFile} should match keys in lang.json`
        );
```
Gives promising output, thoughts? Perhaps limit to max 10 mismatch infos?
```
[15:59:19] Created /Users/pawelfus/Projects/highcharts/ts/masters/i18n/fr-FR.src.ts
[15:59:19] Created /Users/pawelfus/Projects/highcharts/ts/masters/i18n/nb-NO.src.ts
Key mismatch in pl-PL.json: accessibility.screenReaderSection.annotations.descriptionSinglePoint !== accessibility.screenReaderSection.annotations.descriptionSingle
[15:59:19] 'lang-build' errored after 28 ms
[15:59:19] AssertionError [ERR_ASSERTION]: Keys in pl-PL.json should match keys in lang.json
    at langBuild (/Users/pawelfus/Projects/highcharts/tools/gulptasks/lang-build.js:99:16)
```

ts/Stock/RangeSelector/RangeSelector.ts
goransle: IIRC there was a refactor here, likely missed when merging in. Will remove the remaining `defaultButtons` references

ts/Stock/RangeSelector/RangeSelector.ts
goransle: Seems so 👍

ts/Stock/RangeSelector/RangeSelector.ts
```diff
@@ -1450,11 +1467,18 @@ class RangeSelector {
...
             rangeOptions: RangeSelectorButtonOptions,
             i: number
         ): void => {
+            rangeOptions.text = format(rangeOptions.text, {
+                count: rangeOptions.count || 1
+            });
+
+            rangeOptions.title = format(rangeOptions.title, {
+                count: rangeOptions.count || 1
+            });
```
pawelfus, line 1480: Q: The same as previously: Do we need this? In `init()` we set these options with already formatted values, this seems redundant to format already formatted text.

pawelfus (commented):
One question remaining, not sure if we need to change or keep as is.

pawelfus (approved):
Thanks!

PR #20475, opened 2024-01-19
Skip test for troublesome study

TorsteinHonsi:
It dug too greedily and too deep


TorsteinHonsi:
Better solution in https://github.com/highcharts/highcharts/pull/20449

